Issue #800, opened 2026-06-16
UI: crit/dodge/block combat log + display


ginderjeremiah:
Part of spike **#178** (player crit/dodge/block) — area **D** in [`docs/spikes/178-player-crit-dodge-block.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/178-player-crit-dodge-block.md). Frontend-only. Depends on **#798**.

## Goal

Surface crit/dodge/block to the player in combat and on the stat/skill surfaces.

## Scope

- **Combat log** (frontend-only, per the logs policy) — lines for a critical hit, a dodge, and a block, driven off the `crit`/`dodged`/`blocked` flags `battleStep` adds to `SkillActivation` in #798. Reuse the `Damage`/`EnemyDefeated` log types, or add a log type if independent toggling warrants it.
- **Display** — show crit/dodge/block on the skill tooltip / battler stat surfaces. The attribute-breakdown screen self-selects them once they have non-combat contributors (per #528's "has a non-combat modifier" rule), which the #178 derivations provide.
- **Multiplier-style display** — `CriticalDamage` (base `1.5`, read directly) renders as a multiplier (`1.5×`), the same treatment the CooldownRecovery rebase introduces in #797.

Icon art for these attributes is tracked in the deferred follow-ups (#528 deferred it "until they are real").

ginderjeremiah:
Picking this up. The combat-log lines (crit/dodge/block) and the attribute-breakdown surfacing are in PR #833. One bullet needs your call before I implement it:

**The `CriticalDamage` "multiplier-style display" (`1.5×`).** The issue says to give it *"the same treatment the CooldownRecovery rebase introduces in #797"* — but #797 actually shipped `CooldownRecovery` as **`109%`** (`IsPercentage: true`), and `game-design.md` documents it as `109%`, not `1.09×`. `CriticalDamage` is likewise `IsPercentage: true` today, so it already renders **`150%`**.

So there are two ways to read this, and they conflict:

1. **Keep the `%` convention #797 settled** → `CriticalDamage` stays `150%`, no work needed, everything stays consistent. (But `150%` for a `×1.5` crit multiplier is arguably confusing.)
[...]

Issue #799, opened 2026-06-16
Sourcing: derive crit/dodge/block from core attributes + gear authoring


ginderjeremiah:
Part of spike **#178** (player crit/dodge/block) — area **C** in [`docs/spikes/178-player-crit-dodge-block.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/178-player-crit-dodge-block.md). Depends on **#798** (so the values do something).

## Goal

Feed crit/dodge/block from the core attributes so the mechanic is live from raw allocations, and confirm gear/item-mods can grant them.

## Scope

- **Derivation rows** on `StaticAttributeModifiers` (decimal convention; **strawman — tune during balancing**):

  | Attribute | Derivation |
  | --- | --- |
  | `CriticalChance` | `0.002·Dexterity + 0.001·Luck` |
  | `CriticalDamage` | base `1.5` + `0.0025·Luck` (read directly, like CooldownRecovery) |
  | `DodgeChance` | `0.001·Agility` |
  | `BlockChance` | `0.002·Endurance` |
  | `BlockReduction` | base `2` + `0.5·Endurance` |

  `CriticalDamage` carries a base multiplier of `1.5` so a crit is worth something before gear. Rows mirror to the client via the `STATIC_ATTRIBUTE_MODIFIERS` codegen automatically.
- **Gear authoring** — verify items/item-mods can carry crit/dodge/block modifiers (they already accept arbitrary `EAttribute` values); confirm the admin attribute pickers expose them now that they're "real" rather than unused.
- **Tests** — unit-test the derived composition (each attribute resolves from its core inputs; `CriticalDamage` base present).

Coefficients are a conservative starting point and expected to be tuned.

Issue #797, opened 2026-06-16
Foundation: decimal percentage convention + CooldownRecovery base-1 multiplier rebase


ginderjeremiah:
Part of spike **#178** (player crit/dodge/block) — area **A** in [`docs/spikes/178-player-crit-dodge-block.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/178-player-crit-dodge-block.md). The foundation the crit/dodge/block work builds on; **no new mechanic**.

## Goal

Settle the percentage/multiplier attribute storage convention #528 deferred, and rebase `CooldownRecovery` to a base-`1` multiplier so multiplicative modifiers behave intuitively.

## Scope

- **Decimal convention** — percentage attributes are stored as decimal fractions and consumed **without transformation** (`0.05` = 5%, compared directly against a `[0,1)` RNG draw). Multiplier-style attributes (`CooldownRecovery`, and later `CriticalDamage`) are read directly with a base ≥ 1.
- **CooldownRecovery rebase** (behaviour-preserving):
  - `StaticAttributeModifiers`: add a `BaseValue` of `1.0` for `CooldownRecovery` and rescale the derived coefficients `÷100` → `+0.004·Agility + 0.001·Dexterity` (AGI 20, DEX 10 → `1.09`, identical to today).
  - `Battler.GetCooldownMultiplier()` (BE) and `cooldownMultiplier()` (FE, `battle-formulas.ts`) become `return attributes[CooldownRecovery]` — drop the `1 + x/100` transform. Audit every consumer (the shared display surfaces — skills page, skill tooltips — go through the same function and follow automatically).
- **Parity vectors** — non-buff CDR scenarios keep their expected `totalMs` (value unchanged); re-author the CDR-*buff* scenarios' effect amounts (a "double speed" buff becomes `+1.0` additive or a `×2` multiplicative on the base `1`) on **both** sides (`BattleSimulatorParityTests` ↔ `battle-simulation-parity.test.ts`).
- **Display** — `CooldownRecovery` is now a multiplier centred at `1`; revisit its Attribute display metadata (`IsPercentage`/`Decimals`) so it renders sensibly (e.g. `1.09×` rather than `109%`).

## Parity note

The CDR formula executes on both sides; keep the mirrored parity matrices in lockstep. Because the `STATIC_ATTRIBUTE_MODIFIERS` frontend table is codegen'd from `StaticAttributeModifiers.All`, the rebased rows mirror to the client automatically.

Issue #336, opened 2026-06-10
Deferred skill-effect extensions (post-#180 tracking)


ginderjeremiah:
Tracking issue for the extensions the #180 skill-effects spike **deliberately deferred** (see [`docs/spikes/180-skill-effects.md`](../blob/main/docs/spikes/180-skill-effects.md) → "Deferred extensions"). The V1 design (#331–#335) leaves an explicit seam for each; this issue keeps them on the list so they aren't forgotten. Each item should be split into its own issue (with design discussion where flagged) when it's actually picked up.

- **Proc chance (`ChanceToApply`)** — chance-based effect application. Must land **with/after the seeded-RNG work (#178)** and requires defining a global RNG draw order across crit/dodge/effect rolls so frontend/backend consume the `BattleSeed` stream identically. *Needs design.*
- **Hit-count durations & every-Nth-hit triggers** — the spike issue's "after x hits" idea. Add a duration-type discriminator (e.g. `EEffectDurationType { Time, HitsTaken }`) and/or trigger fields to `SkillEffect`; needs hit-event bookkeeping on the `Battler` and new parity vectors.
- ~~**Stacking policies** — V1 is refresh-only. If that proves too limiting, make the policy authored per effect (refresh / stack with cap), with stack-count UI on the active-effect chips.~~ - Done
- ~~**Per-skill DoT attribution** — V1 books DoT damage into the aggregate stats only; attributing it to the sourcing skill's `SkillStats` requires tracking effect provenance through the DoT phase.~~ - Not planned
- **Dispel/cleanse and instant-heal effect kinds** — effect kinds beyond timed attribute modifiers (an effect "kind" discriminator on `SkillEffect`).
- **Content authoring** — once the mechanic ships, author effect-bearing skills (and revisit enemy loadouts / challenge rewards) so effects are actually encountered in play. Content task in the admin tools, not code.
- **Card minigame synergy** — #259 expects skills to map onto cards; effect-bearing skills are the interesting design space there.

PR #830, opened 2026-06-17
Battle runtime: seeded RNG draw order + player crit/dodge/block (#798)

### LINKED ISSUE #798: Battle runtime: seeded RNG draw order + player crit/dodge/block
Part of spike **#178** (player crit/dodge/block) — area **B** in [`docs/spikes/178-player-crit-dodge-block.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/178-player-crit-dodge-block.md). **The parity-critical core** — the first live consumer of the seeded battle RNG. Depends on **#797** (the convention/CDR math).

## Goal

Let the **player** crit (on its own attacks), dodge, and block (on incoming attacks). Enemies do not — gated on the acting battler, leaving a clean seam for later enemy parity.

## Scope

- **Seed threading** — construct one `Mulberry32` from `BattleSeed`, seeded once at battle start and advanced in lockstep on both sides:
  - Backend: `PlayerState.BattleSeed → BattleService.SimulateBattle → BattleSimulator → BattleContext`.
  - Frontend: `BattleEngine.reset` seeds from `enemyInstance.seed`; the headless `BattleSimulator` from the same value; both pass it into `battleStep`.
- **RNG draw order** (a pure function of the skill-fire sequence, outcome-independent):
  - Each **player** skill fire → **1 crit draw** (before damage is dealt).
  - If the enemy survives, each **enemy** skill fire → **2 draws (dodge, then block)**, taken **unconditionally** (even when the hit is dodged), so the per-tick draw count is `playerFires×1 + enemyFires×2` and never branches on a roll result.
- **Math** (player-only, gated on who acts):
  - **Crit:** `c < CriticalChance` ⇒ raw damage `× CriticalDamage` (read directly, base `1.5`) **before** Defense subtraction.
  - **Dodge:** `d < DodgeChance` ⇒ hit deals `0`.
  - **Block:** `b < BlockChance` (non-dodged) ⇒ `max(raw − Defense − BlockReduction, 0)`.
  - `TakeDamage` gains an optional block-reduction parameter; `SkillActivation` (FE) gains `crit`/`dodged`/`blocked` flags for logging (the headless simulator ignores them; the backend needs no flags in V1, but the **rolls and damage results must match**).
- **Parity vectors** (both sides, `BattleSimulatorParityTests` ↔ `battle-simulation-parity.test.ts`) — use forced `1`/`0` chances so outcomes are deterministic without a real probability: a forced crit multiplies pre-Defense; a forced dodge negates; a forced block reduces flatly; draw-order alignment over a multi-skill exchange; and a forced-`1` chance on the **enemy** is ignored (enemies never crit/dodge/block).

## Parity note

This is the riskiest surface in the spike: both simulators must seed identically and draw in the identical, outcome-independent order. #336's deferred proc chance later slots its own draw into this order.

ginderjeremiah:
Closes #798. Implements area **B** of spike #178 — the parity-critical core and the **first live consumer** of the seeded battle RNG. Depends on #797 (merged).

### Seed threading (one `Mulberry32`, seeded once, advanced in lockstep)
- **Backend:** `PlayerState.BattleSeed → BattleService.SimulateBattle → BattleSimulator → BattleContext`.
- **Frontend:** `BattleEngine.reset` seeds from `enemyInstance.seed`; the headless `BattleSimulator` from its own seed; both pass the rng into `battleStep`.

### RNG draw order (a pure function of the skill-fire sequence)
Each **player** skill fire → **1 crit draw** (before damage). If the enemy survives, each **enemy** skill fire → **2 draws (dodge, then block)**, taken **unconditionally** (even on a dodge), so the per-tick count is `playerFires×1 + enemyFires×2` and never branches on a roll result. Pure-buff fires still draw. This is the seam #336's deferred proc chance later slots into.

### Math — player-only, gated on the acting battler (enemies never crit/dodge/block)
- **Crit:** `c < CriticalChance` ⇒ raw `× CriticalDamage` (read directly) **before** Defense subtraction.
- **Dodge:** `d < DodgeChance` ⇒ hit deals `0`.
- **Block:** `b < BlockChance` (non-dodged) ⇒ `max(raw − Defense − BlockReduction, 0)`.
- `TakeDamage` / `applyDefense` gain an optional block-reduction parameter; the FE `SkillActivation` gains `crit`/`dodged`/`blocked` flags for the combat log (consumed in #800). The headless simulator and backend ignore the flags, but the **rolls and damage results match** across both sides.

### Inert until sourced
No `StaticAttributeModifiers` row feeds crit/dodge/block yet, so every chance is `0` from raw allocations and the feature does nothing in real play until the derivations land in **#799** — exactly as the spike scopes it.

## Tests
5 new **mirrored** parity scenarios (`BattleSimulatorParityTests` ↔ `battle-simulation-parity.test.ts`), using forced `1`/`0` chances so outcomes are deterministic without a real probability:
- `forcedCrit` — crit multiplies pre-Defense (kill on hit 3 vs 6)
- `forcedDodge` — a forced dodge negates a one-shot, flipping loss → win
- `forcedBlock` — a flat block reduction flips loss → win
- `drawOrderMultiSkill` — two player + two enemy skills firing the same ticks keep the draw stream aligned
- `enemyForcedChanceIgnored` — every chance forced to `1` on the enemy is ignored

Plus focused draw-count + crit/dodge/block math unit tests on **both** sides (the draw-count test pins `1 draw per player fire / 2 per enemy fire` by comparing the stream position, independent of the seed).

### Verification (run locally)
- Backend: `Game.Core.Tests` 571 ✓, `Game.Application.Tests` 272 ✓ (integration + round-trip replay), `Game.Api.Tests` 498 ✓, `Game.Infrastructure.Tests` 37 ✓.
- Frontend: `npm run lint` ✓ (+ `svelte-check` clean), `npm run test:unit:ci` 1899 ✓.

## Docs
Updated `docs/backend-battle.md` (the seed note now has a live consumer + a new seeded crit/dodge/block runtime section), `docs/game-design.md` (crit/dodge/block as player mechanics), and `docs/frontend.md` (the seeded RNG as a parity surface).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GNfGqxeGRyByJeqCte7PnL)_


Game.Core/Battle/BattleSimulator.cs
```diff
@@ -4,18 +4,22 @@ public class BattleSimulator
     {
         private Battler PlayerBattler { get; set; }
         private Battler EnemyBattler { get; set; }
+        private readonly uint _seed;
 
-        public BattleSimulator(Battler playerBattler, Battler enemyBattler)
+        public BattleSimulator(Battler playerBattler, Battler enemyBattler, uint seed = 0)
```
ginderjeremiah: Nit (non-blocking): the `seed = 0` default makes a parity-critical input optional, and the FE `BattleSimulator` constructor has the same default. It's harmless today (the one production caller passes the real seed, and every chance is `0` so the stream is inert), but a future production caller that forgets the seed would silently simulate against a seed-0 stream and diverge from the client's replay. Note that `battleStep` itself made `rng` **required** rather than defaulted — the safer choice for a parity surface. Consider doing the same here and letting the test call sites pass an explicit `0`, so the default can't mask a real divergence later. Happy to leave it as-is given the test ergonomics, just flagging the asymmetry.

---

ginderjeremiah (commented):
## Review: Battle runtime — seeded RNG draw order + player crit/dodge/block (#798)

**Verdict:** ✅ Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating event from me) — the parity-critical surface is implemented correctly, the draw order is decoupled from roll outcomes as the spike requires, and the change is thoroughly mirrored on both sides. No blocking findings.

I verified the things most likely to bite on a parity surface:

- **Draw order matches tick-for-tick.** `BattleSimulator.Simulate` fires all player skills (each → 1 crit draw in `DamageTarget`) then, only if the enemy survives, all enemy skills (each → 2 draws, dodge-then-block, unconditional) — exactly mirroring `battleStep`'s player-then-enemy order and its `if (!enemy.isDead)` gate. Per-tick count is `playerFires×1 + enemyFires×2` and never branches on a roll.
- **The two `Mulberry32` ports are identical** and already pinned by `Mulberry32ParityTests` ↔ `mulberry32-parity.test.ts`. Combined with the forced-`1/0` math scenarios and the draw-count test, this transitively covers a real-probability draw too (same seed → same sequence → same comparison → same outcome).
- **Gating is on the acting battler** — enemies never crit/dodge/block, and `enemyForcedChanceIgnored` proves it on both sides.
- **`?? 0` seed fallback is sound** — `SetActiveBattle`/`ClearBattle` set and clear `BattleSeed` and `Snapshot` together, so the non-null-snapshot guard guarantees a non-null seed; the `?? 0` mirrors the existing `?? 1`/`?? []` defensive fallbacks.
- **No runtime hazards** — `AttributeCollection` returns `0` for unallocated attributes (no `KeyNotFoundException`), and there's no `StaticAttributeModifiers` row for crit/dodge/block, so the feature is genuinely inert and the test allocations resolve to exact values.

### Should-consider (non-blocking)

- **Skill stats vs. crit.** `BattleSkill.Update` records `RecordSkillUse(Skill.Id, damage)` with the **pre-crit, pre-Defense** raw damage, while `DamageTarget` accrues `PlayerDamageDealt`/`HighestPlayerAttack` from the **post-crit, post-Defense** actual. So a crit now shows up in `HighestPlayerAttack` but not in `SkillStats.TotalDamage`/`HighestSingleAttack`. Skill stats already diverged from actual on Defense, so this may be the intended "raw skill contribution" semantics — but the crit widens the gap. Worth a conscious decision; if the per-skill stat is meant to reflect realized damage, that's a small follow-up.

### Nits

- The `seed = 0` default on both `BattleSimulator` constructors makes a parity-critical input optional where `battleStep` (correctly) made `rng` required — see the inline note. Optional.

### Test coverage

Strong, and it follows the project's battle-parity rule precisely:
- 5 mirrored parity scenarios (`forcedCrit`/`forcedDodge`/`forcedBlock`/`drawOrderMultiSkill`/`enemyForcedChanceIgnored`) exist row-for-row in `BattleSimulatorParityTests` ↔ `battle-simulation-parity.test.ts` with the same hardcoded expectations.
- Focused crit/dodge/block math + "enemy never crits" + a seed-independent draw-count test on **both** sides (`BattleContextTests` ↔ `battle-step.test.ts`), plus the `applyDefense` block-reduction case.
- The draw-count test (comparing stream position against a reference advanced by exactly 3 draws) is a nice way to pin the per-fire count independent of the seed.

A direct real-probability parity scenario (outcome driven by an actual draw value rather than a forced `1/0`) isn't included — not required given the three guarantees compose, but it'd be cheap belt-and-suspenders once #799 sources real chances.

### What's good

- Draw order designed as a pure function of the fire sequence — the explicit "both draws always taken, even on a dodge" is exactly the outcome-independence the spike calls for, and leaves the clean seam for #336's deferred proc draw.
- Crit multiplying **before** Defense (read directly like `CooldownRecovery`) is consistent with the attribute-convention doc.
- Docs are accurate and well-scoped: the seed note gains its live consumer, a new seeded-runtime section lands in `backend-battle.md`, and `game-design.md`/`frontend.md` are updated without over-documenting.

---

Game.Core/Battle/BattleSimulator.cs
ginderjeremiah: Good call — addressed in 4facff5. Made `seed` a required parameter on both the backend and frontend `BattleSimulator` constructors (matching `battleStep`'s required `rng`); the test call sites now pass an explicit seed (`0` where the scenario is RNG-independent, the shared `ParitySeed`/`PARITY_SEED` in the parity suites). The one production caller, `BattleService.SimulateBattle`, already threads the real seed, so this just removes the latent foot-gun for a future caller.

---

ginderjeremiah:
Thanks for the review! Both non-blocking items handled:

- **Seed default (nit):** addressed in 4facff5 — `seed` is now required on both `BattleSimulator` constructors, matching `battleStep`'s required `rng`. (Replied on the inline thread.)
- **Skill stats vs. crit (should-consider):** conscious decision — keep the current raw skill-contribution semantics for now. The per-skill `SkillStats` already recorded the pre-Defense raw contribution; crit just widens the existing gap against the realized battle-level stats. Realized per-skill crit attribution is statistics work, which the spike defers to #801 — I've added a note there so the realized-vs-raw decision is made when that item is split out.

The direct real-probability parity scenario is similarly a nice-to-have for once #799 sources real chances; noted for then.

---

PR #832, opened 2026-06-17
Match frontend skill-damage float grouping to the backend (#802)

### LINKED ISSUE #802: Frontend skill-damage formula uses a different float grouping than the backend (parity divergence)
## Problem

The shared scalar damage formula is grouped differently on the two sides, breaking the bit-for-bit parity the anti-cheat replay depends on.

**Backend** — `Game.Core/Battle/BattleSkill.cs:56-64` accumulates the multiplier bonus from `0.0` and adds `BaseDamage` **last**:

```csharp
var multiplierBonus = 0.0;
for (...) { multiplierBonus += attr * amount; }
return Skill.BaseDamage + multiplierBonus;   // base + (m1 + m2 + …)
```

It carries an explicit comment that this exact ordering is *"required for frontend/backend battle parity … bit-for-bit identical"* because floating-point addition is not associative.

**Frontend** — `UI/src/lib/battle/battle-formulas.ts:24-30` adds `baseDamage` **first**, then accumulates:

```ts
let damage = skill.baseDamage;
for (const mult of skill.damageMultipliers) { damage += contributionValue(mult, attributes); }
return damage;                                 // ((base + m1) + m2)
```

So the two compute `base + (m1 + m2)` vs `((base + m1) + m2)`. These are identical for 0 or 1 multiplier, but for a skill with **≥2 damage multipliers** they can differ by a ULP, and at an exact HP/Defense boundary that sub-ULP difference can flip a kill — a live-vs-replay desync.

## Why it isn't already caught

`battle-simulation-parity.test.ts` only asserts coarse outcomes (`victory`/`playerDied`/`totalMs`), never per-hit damage, so a single-tick damage difference that doesn't change the gross outcome (or only changes it at a rare boundary) passes.

## Suggested fix

- Reorder `calculateSkillDamage` to accumulate the multiplier bonus separately and add `baseDamage` last, matching the backend grouping exactly.
- Add a per-hit-damage parity assertion (a multi-multiplier skill) to the mirrored parity matrices so this can't silently drift again.

### Labels
bug, claude, scope: small

ginderjeremiah:
## Summary

Fixes the frontend/backend float-grouping divergence in the shared skill-damage formula (#802).

`calculateSkillDamage` (`UI/src/lib/battle/battle-formulas.ts`) added `baseDamage` **first** and then accumulated each multiplier contribution — computing `((base + m1) + m2)`. The backend (`BattleSkill.CalculateDamage`) accumulates the multiplier bonus from `0.0` and adds `BaseDamage` **last** — `base + (m1 + m2)`, with an explicit comment that this exact ordering is required for bit-for-bit parity.

Floating-point addition is not associative, so for a skill with **two or more** damage multipliers the two groupings can differ by a ULP, and at an exact HP/Defense boundary that sub-ULP difference can flip a kill — a live-vs-replay desync the anti-cheat replay depends on not happening.

## How it addresses the issue

- **Reordered `calculateSkillDamage`** to accumulate the multiplier bonus separately from zero and add `baseDamage` last, mirroring the backend grouping exactly. Identical for 0 or 1 multiplier (so existing scenarios are unaffected); now bit-for-bit for ≥2.
- **Added a mirrored per-hit-damage parity test** on both sides — `battle-formulas.test.ts` and `BattleSkillTests.cs` — driven by ULP-sensitive inputs (two contributions each below `baseDamage`'s ULP that vanish when added one at a time but survive when summed first). The correct grouping lifts the result above `baseDamage`; the naive grouping returns it unchanged. This pins the exact ordering rather than only the coarse battle outcomes the simulation-parity matrix asserts (which is why the existing suite never caught it).

The simulation-parity matrix asserts whole-battle `BattleResult`s (victory/playerDied/totalMs), which can't carry a per-hit value, so the guard lives as a dedicated mirrored test pair next to each formula, each cross-referencing the other.

## Test plan

- [x] `dotnet test Game.Core.Tests` — 577 passed (incl. new `CalculateDamage_FloatGrouping_MatchesBackend`).
- [x] `vitest run battle-formulas.test.ts battle-simulation-parity.test.ts` — 40 passed (incl. new grouping test).

Closes #802

https://claude.ai/code/session_01MCTFjZnYhN71UME1JTkgfs

---
_Generated by [Claude Code](https://claude.ai/code/session_01MCTFjZnYhN71UME1JTkgfs)_

### DIFF
--- a/Game.Core.Tests/Battle/BattleSkillTests.cs
+++ b/Game.Core.Tests/Battle/BattleSkillTests.cs
@@ -96,6 +96,64 @@ namespace Game.Core.Tests.Battle
         }
 
         [Fact]
+        public void CalculateDamage_FloatGrouping_MatchesBackend()
+        {
+            // Per-hit-damage parity guard (mirrors battle-formulas.test.ts "groups the multiplier sum like the
+            // backend"). Floating-point addition is not associative, so `base + (m1 + m2)` and a naive
+            // `(base + m1) + m2` can differ by a ULP at a kill boundary — a live-vs-replay desync (#802). Two
+            // contributions each below baseDamage's ULP vanish if added one at a time but survive when summed
+            // first, so the correct grouping lifts the result above baseDamage and the wrong one returns it
+            // unchanged — pinning the exact ordering, not just a coarse outcome.
+            const double tiny = 1e-16; // below the ULP of 1.0, so a single contribution is lost when added to base
+            var multipliers = new List<DamageMultiplier>
+            {
+                new() { Attribute = EAttribute.Strength, Amount = tiny },
+                new() { Attribute = EAttribute.Agility,  Amount = tiny },
+            };
+            var skill = MakeSkill(cooldownMs: 1000, baseDamage: 1, multipliers: multipliers);
+            var battleSkill = new BattleSkill(skill);
+
+            var statAllocations = new List<StatAllocation>
+            {
+                new() { Attribute = EAttribute.Strength,  Amount = 1 },
+                new() { Attribute = EAttribute.Agility,   Amount = 1 },
+                new() { Attribute = EAttribute.Endurance, Amount = 50 },
+                new() { Attribute = EAttribute.Intellect, Amount = 0 },
+                new() { Attribute = EAttribute.Dexterity, Amount = 0 },
+                new() { Attribute = EAttribute.Luck,      Amount = 0 },
+            };
+            var attacker = new Battler(new Player
+            {
+                Id = 0,
+                Name = "t",
+                Level = 1,
+                Exp = 0,
+                CurrentZoneId = 0,
+                StatPoints = new PlayerStatPoints
+                { StatAllocations = statAllocations, StatPointsGained = 52, StatPointsUsed = 52 },
+                Inventory = new Inventory(),
+                SelectedSkills = [],
+                Skills = [],
+                LogPreferences = [],
+            });
+            var defender = MakeBattler(strength: 0);
+            var context = new BattleContext(attacker, defender, timeDelta: 0, new Mulberry32(0));
+
+            const double c1 = 1 * tiny;
+            const double c2 = 1 * tiny;
+            const double correct = 1 + (c1 + c2); // base + (m1 + m2)
+            const double naive = 1 + c1 + c2;     // ((base + m1) + m2)
+            // Sanity: the inputs are ULP-sensitive — the two groupings genuinely differ.
+            Assert.NotEqual(correct, naive);
+            Assert.Equal(1.0, naive);
+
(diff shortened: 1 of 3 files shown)


ginderjeremiah:
## Automated review — PR #832 (fixes #802)

**Verdict: approve-with-nits** · **CI: green** (7/7 check runs succeeded, incl. `all-checks-passed`; the legacy commit-status API reports `pending`/`total_count:0` only because nothing uses it — the check runs are the real signal.)

### Correctness — verified
- The reordered frontend `calculateSkillDamage` now accumulates `multiplierBonus` from `0` and adds `baseDamage` **last**, producing `base + (m1 + m2 + …)` — bit-for-bit identical to the backend `BattleSkill.CalculateDamage` grouping (`Skill.BaseDamage + multiplierBonus`). Confirmed against `Game.Core/Battle/BattleSkill.cs:56-64`. Identical for 0/1 multiplier, so existing scenarios are unaffected; correct for ≥2.
- The mirrored test pair is genuinely ULP-sensitive and equivalent on both sides. With `tiny = 1e-16` and `baseDamage = 1` (ULP of 1.0 ≈ 2.22e-16): a single contribution added to base rounds away (`1 + 1e-16 == 1.0`), so the naive `((base + m1) + m2)` collapses to `baseDamage`; summing the two contributions first yields `2e-16`, which crosses the rounding threshold so `base + (m1 + m2)` lifts to `1 + 1 ULP`. Both suites assert `naive == baseDamage`, `correct != naive`, and `damage == correct` — a real guard on the exact ordering, not a coarse outcome. Exact-equality assertions (no tolerance) are the right call here.

### Battle-parity test requirement — satisfied
Mirrored as `BattleSkillTests.CalculateDamage_FloatGrouping_MatchesBackend` ↔ `battle-formulas.test.ts "groups the multiplier sum like the backend"`, with identical inputs/expectations and each cross-referencing the other. Living next to each formula (rather than in the whole-battle simulation-parity matrix, which only carries `victory/playerDied/totalMs` and can't express a per-hit value) is the right placement and matches the reasoning in #802.

### Quality
Braces present, no dead code, DRY. The expanded doc comment on `calculateSkillDamage` is multi-line but appropriately so (no one-line rule here). The backend comment already documented this invariant; the frontend now mirrors it.

### Nits (non-blocking)
- Backend test builds a full `Player`/`PlayerStatPoints` inline because `MakeBattler` only sets a single attribute; a small multi-allocation `MakeBattler` overload would tidy this and similar future tests. Minor, optional.
- `c1 = 1 * tiny` / `c2 = 1 * tiny` keep the literal `* 1` to mirror "attribute value 1 × multiplier"; self-documenting but slightly redundant.

No follow-up issues required.

— Automated review by Claude Code

---

PR #834, opened 2026-06-17
Make AttributeDistribution & Challenge immutable; discovery-based immutability guard (#803)

### LINKED ISSUE #803: Shared reference-data domain models AttributeDistribution & Challenge are mutable; immutability guard is hand-maintained
## Problem

`docs/backend.md` → _Reference Data_ states the invariant that gameplay reference-data domain models are *"structurally immutable value objects — `init`-only properties and `IReadOnlyList<>` collections, enforced by the compiler and guarded by a reflection test — so a consumer cannot grow or reassign the shared graph and corrupt the cache for every player."* Two shared models violate it:

- **`Game.Core/Attributes/AttributeDistribution.cs:13-23`** exposes public `set;` on `AttributeId`, `BaseAmount`, `AmountPerLevel`. The same `IReadOnlyList<AttributeDistribution>` instance is shared by reference across every `Enemy` produced from a cached `EnemyTemplate` (`EnemyTemplate.ToEnemy(level)` does not clone the distributions).
- **`Game.Core/Progress/Challenge.cs:8-20`** exposes public `set;` on every property; `Challenges.All()` (`Game.DataAccess/Repositories/Challenges.cs:9-14`) hands the cached snapshot back directly ("callers treat it as read-only") and it is fed into `PlayerProgress.EvaluateChallenges` on every battle.

Both are latent today (nothing mutates them), but any future code that sets a field on a shared distribution/challenge would silently corrupt state for **every concurrent battle in the process**.

Compounding it: the reflection guard `Game.Core.Tests/Items/ReferenceDataImmutabilityTests.cs:22-30` enumerates the covered types **by hand** (`Item, ItemMod, ItemModSlot, Skill, SkillEffect, AttributeModifier`) and omits `AttributeDistribution`, `Challenge`, `EnemyTemplate`, and `Enemy` — so new shared models silently escape the guard.

## Suggested fix

- Change the offending `set;` to `init;` on `AttributeDistribution` and `Challenge` (and audit `EnemyTemplate`/`Enemy`).
- Make the immutability guard **discover** the reference-data model types by reflection (e.g. everything returned by the gameplay reference repos / under a namespace) rather than a hand-maintained list, so coverage can't lag behind new models.

### Labels

ginderjeremiah:
## Summary

Closes the two shared reference-data models that violated the documented "structurally immutable value object" invariant (`docs/backend.md` → Reference Data), and reworks the immutability guard so coverage can't silently lag behind new models again (#803).

### What changed

- **`AttributeDistribution`** (`Game.Core/Attributes/AttributeDistribution.cs`) — `AttributeId` / `BaseAmount` / `AmountPerLevel` changed from `set;` → `init;`. The same `IReadOnlyList<AttributeDistribution>` instance is shared by reference across every `Enemy` produced from a cached `EnemyTemplate` (`ToEnemy` does not clone), so a stray write would have corrupted the shared graph for every concurrent battle.
- **`Challenge`** (`Game.Core/Progress/Challenge.cs`) — every property changed from `set;` → `init;`. `Challenges.All()` hands the cached snapshot back directly into `PlayerProgress.EvaluateChallenges` on every battle.
- **`EnemyTemplate` / `Enemy` audit** — both already expose only `init`-only public properties (`Enemy`'s only mutable state is the private per-encounter `_battleSkills` field, which is by design and not public). No change needed; both are now also covered by the guard.
- **`ReferenceDataImmutabilityTests`** — replaced the hand-maintained type list with **graph discovery**: a breadth-first walk of the public-property graph out from the cached aggregate roots (Item, ItemMod, Skill, EnemyTemplate, Enemy, Zone, Challenge), collecting every reachable `Game.Core` class/struct (unwrapping `IReadOnlyList<T>` / `Nullable<T>` generic args). Nested value objects — exactly how `AttributeDistribution` slipped through before — and any new field on an existing model are now picked up automatically. Added a `Discovery_CoversTheOriginalReferenceDataModels` sanity test so a refactor that severs the graph (and thus the coverage) is caught.

### How it addresses the issue

Both bullets of the suggested fix are done: the offending `set;` are now `init;` (with `EnemyTemplate`/`Enemy` audited), and the guard discovers the reference-data model types by reflection rather than a hand-maintained list.

### Design note

The discovery still has a small, stable list of *aggregate roots* (the cached snapshot element types — which the docs already enumerate). It can't derive them from the reference repos because those live in `Game.DataAccess` (internal) and `Game.Core.Tests` references only `Game.Core`; a namespace-based scan is also unworkable because `Game.Core.Progress` mixes the immutable `Challenge` with the mutable `PlayerChallenge`/`PlayerProgress`. The roots-plus-graph-walk approach directly fixes the actual failure mode (nested models escaping) while staying contained to the test project with no production marker types.

## Test plan

- [x] `dotnet test --no-build --no-restore` — all backend suites green (Core 595, Application 281, Api 502, Infrastructure 37).

Closes #803

https://claude.ai/code/session_012TpvUrnno1WL37DvUCTSbR

---
_Generated by [Claude Code](https://claude.ai/code/session_012TpvUrnno1WL37DvUCTSbR)_

### DIFF
--- a/Game.Core.Tests/Items/ReferenceDataImmutabilityTests.cs
+++ b/Game.Core.Tests/Items/ReferenceDataImmutabilityTests.cs
@@ -1,32 +1,56 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using Game.Core.Attributes;
 using Game.Core.Attributes.Modifiers;
+using Game.Core.Enemies;
 using Game.Core.Items;
+using Game.Core.Progress;
 using Game.Core.Skills;
+using Game.Core.Zones;
 using Xunit;
 
 namespace Game.Core.Tests.Items
 {
     /// <summary>
-    /// Structural-immutability guard for the shared, cached reference-data domain models (#547) and the
-    /// <see cref="AttributeModifier"/> element type held inside their read-only collections (#603).
-    /// <c>GetItem</c>/<c>GetItemMod</c>/<c>GetSkill</c> hand the same pre-materialized instance to every
-    /// player straight from the reference cache, so these types must stay structurally immutable — every
-    /// property init-only and every collection an <see cref="IReadOnlyList{T}"/> — or a stray mutation would
-    /// silently corrupt the cache for every player. The compiler enforces "you can't mutate it"; this test
-    /// enforces "you can't quietly drop the immutability" (flip an <c>init</c> back to <c>set</c>, or a
-    /// read-only collection back to a mutable one), which the compiler alone would let pass.
+    /// Structural-immutability guard for the shared, cached reference-data domain models (#547) and every
+    /// value object reachable through their graphs (e.g. the <c>AttributeModifier</c>/<c>AttributeDistribution</c>
+    /// element types held inside their read-only collections, #603). The reference repos hand the same
+    /// pre-materialized instance to every player straight from the reference cache, so these types must stay
+    /// structurally immutable — every property init-only and every collection an <see cref="IReadOnlyList{T}"/> —
+    /// or a stray mutation would silently corrupt the cache for every player. The compiler enforces "you can't
+    /// mutate it"; this test enforces "you can't quietly drop the immutability" (flip an <c>init</c> back to
+    /// <c>set</c>, or a read-only collection back to a mutable one), which the compiler alone would let pass.
+    ///
+    /// The covered set is <b>discovered</b> by walking the public-property graph out from the cached aggregate
+    /// roots rather than hand-listed (#803): a nested model (or a brand-new field on an existing one) is picked
+    /// up automatically, so coverage can't silently lag behind the models the way a maintained list did when
+    /// <see cref="AttributeDistribution"/> and <see cref="Challenge"/> were added.
     /// </summary>
     public class ReferenceDataImmutabilityTests
     {
+        /// <summary>
+        /// The cached reference-data aggregate roots — the gameplay domain models handed out from the
+        /// reference caches (items, item mods, skills, enemies, zones, challenges; see
+        /// <c>docs/backend.md → Reference Data</c>). Every other covered type is reached by walking these.
+        /// <see cref="Enemy"/> is included alongside <see cref="EnemyTemplate"/> because it reuses the
+        /// template's shared collections by reference, so it is part of the same shared graph.
+        /// </summary>
+        private static readonly Type[] ReferenceDataRoots =
+        [
+            typeof(Item),
+            typeof(ItemMod),
+            typeof(Skill),
+            typeof(EnemyTemplate),
+            typeof(Enemy),
+            typeof(Zone),
+            typeof(Challenge),
+        ];
+
+        private static readonly Assembly CoreAssembly = typeof(Item).Assembly;
+
         public static IEnumerable<object[]> SharedReferenceDataModels()
         {
-            yield return new object[] { typeof(Item) };
-            yield return new object[] { typeof(ItemMod) };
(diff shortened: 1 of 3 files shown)


ginderjeremiah:
## Code review — PR #834

**Verdict: approve-with-nits** ✅  ·  **CI: green** (all 7 check runs succeeded; `test-backend`, all three `test-e2e` browsers, and `all-checks-passed` passed — `test-frontend` correctly skipped as no frontend files changed. The `get_status` "pending" is only an empty legacy commit-status API; the check runs are authoritative.)

This is a clean, well-scoped fix that closes both bullets of #803.

### What was verified
- **`AttributeDistribution`** and **`Challenge`** — all `set;` → `init;`. Correct: `EnemyTemplate.ToEnemy` reuses the distribution list by reference across every produced `Enemy`, and `Challenges.All()` hands the cached snapshot straight into per-battle evaluation, so both were genuine (if latent) shared-graph hazards. Mandatory `required` semantics are preserved.
- **`EnemyTemplate` / `Enemy` audit** — confirmed: both expose only init-only / getter-only public properties. `Enemy._battleSkills` is a private per-encounter field (by design), and `Enemy.BattleSkills` is a getter-only `IReadOnlyList<Skill>`, so it passes both the init-only and read-only-collection checks. No change needed, and both are now covered.
- **Discovery test** — the BFS graph walk is correct and robust: it terminates via the `discovered.Add` visited-set on cycles/back-references, `ModelTypesIn` recursively unwraps `IReadOnlyList<T>` / `Nullable<T>` generic args, and `IsCoreModel` correctly scopes to the `Game.Core` assembly while excluding enums, `string`, and framework types. I traced every reachable node — the guard now genuinely covers types the old hand-list missed: `Tag`, `ChallengeType`, `StatisticType`, `DamageMultiplier`, `ItemModSlot`, `EnemyTemplate`, `Enemy`, plus the two fixed models — and all of them are init-only/getter-only, consistent with the green run.
- **`Discovery_CoversTheOriginalReferenceDataModels`** sanity test uses `Assert.Contains` (subset), which is the right call: it catches a severed graph without breaking when a new model is legitimately added later.

### Blocking
- None.

### Should-fix
- None.

### Nits (non-blocking)
1. `ModelTypesIn` returns only a constructed generic's *arguments*, never the generic type itself. That's exactly right for `IReadOnlyList<T>`/`Nullable<T>`, but it means a hypothetical generic `Game.Core` model (`Foo<T>`) would be silently skipped. No such type exists today, so this is latent only — a one-line comment noting the assumption would future-proof it.
2. The class-level XML doc still opens with the `GetItem/GetItemMod/GetSkill` phrasing; with coverage now spanning enemies, zones, and challenges, "the reference repos hand out…" (as used later in the same comment) is the more accurate framing. Cosmetic.

### Follow-ups
- None required. The change is self-contained and the doc invariant in `backend.md → Reference Data` already matches the new behavior.

### Summary
The PR does precisely what #803 asked: it closes the two `set;`-exposing shared reference-data models (`AttributeDistribution`, `Challenge`) to `init;`, audits `EnemyTemplate`/`Enemy`, and replaces the brittle hand-maintained immutability type list with a reflection-based BFS over the cached aggregate-root graph so nested value objects and new fields are picked up automatically — directly fixing the failure mode that let `AttributeDistribution` slip through. The discovery logic is correct, terminating, and now covers a meaningfully larger set, all of which already satisfy the invariant. Standards are met (immutable value objects, DDD isolation, DRY, mandatory braces, no `!`, no dead code, test coverage). CI is green. Approving with two cosmetic nits.

---
*🤖 Automated review by Claude Code (Opus 4.8).*

---

PR #884, opened 2026-06-17
Fix tooltip jumping to a fixed spot on mouse click (#880)

### LINKED ISSUE #880: Clicking on elements with tooltips (with mouse) causes tooltip to jump
Especially noticeable in the attribute breakdown page, but also appears for skills on the fight screen and other places.

Acceptance Criteria:
Tooltip can still be triggered via keyboard focus at a fixed location, but mouse clicks do not cause the tooltip to stop tracking the mouse position.

ginderjeremiah:
## Summary

Tooltip triggers open the global tooltip on both **hover** (anchored at the cursor, tracked as the mouse moves) and **focus** (anchored off the element's box, for keyboard users). Because a mouse click *also* focuses the element it hits, the focus handler re-anchored the already-cursor-tracking tooltip off the element's box — so the tooltip visibly **jumped** to a fixed location on click. This was most obvious on the attribute breakdown page, but happened anywhere a tooltip trigger is a focusable element (fight-screen skills, effect chips, zone-nav lock, gated skill rows, challenge rewards, point-buy rows, attribute chips).

## How it's fixed

A new shared helper `focusAnchor(event)` (`$stores/tooltip.svelte`) returns the element to anchor off **only for keyboard focus**, and `undefined` for a mouse/pointer click (whose hover handlers already own the cursor position). Every focus-driven tooltip handler now routes through it, so a click is a no-op for focus-anchoring and the tooltip keeps tracking the cursor.

Distinguishing keyboard vs. mouse focus is done by tracking input modality from document-level `keydown` / `pointerdown` (the `:focus-visible` polyfill strategy) rather than reading `:focus-visible` directly — that pseudo-class isn't reliably readable inside a focus event handler (and is unset during the handler in jsdom). Touch taps still surface the tooltip via the synthesized `mouseenter`, so touch behavior is unchanged.

Sites updated (all share one helper, so behavior stays consistent):
- `components/tooltip/attribute-hover.ts` — attribute breakdown rows, point-buy rows, attribute chips
- `fight/Skills.svelte` — per-skill combat tooltip
- `fight/ActiveEffectChips.svelte` — effect chips
- `fight/ZoneNav.svelte` — locked-zone gate tooltip
- `skills/SkillRow.svelte` — gated skill rows
- `challenges/RewardAffordance.svelte` — reward chips/tiles

## Acceptance criteria

- ✅ Tooltip can still be triggered via keyboard focus at a fixed location (off the element's box).
- ✅ Mouse clicks no longer cause the tooltip to stop tracking the mouse position.

## Test plan

- New unit tests for `focusAnchor` (keyboard → element, mouse/pointer → `undefined`, modality re-tracking, non-element target).
- Updated `attributeHover` tests: keyboard focus anchors off the box; a mouse-click focus is a no-op so the cursor anchor stands (regression test for #880).
- `npm run lint` (eslint + svelte-check): clean.
- `npm run test:unit:ci`: 200 files / 1983 tests pass.

Docs: added a concise note to the frontend Accessibility section so future tooltip triggers gate their focus handler through `focusAnchor`.

Closes #880

---
_Generated by [Claude Code](https://claude.ai/code/session_01KYKzFVz2YLavvjt1W6YHvf)_


ginderjeremiah (commented):
## Review: Fix tooltip jumping on mouse click (#880)

**Verdict:** Approve — correct, consistently applied fix with solid central + regression test coverage. _(Submitted as a `COMMENT` only because I'm the PR author and GitHub won't accept a gating verdict on one's own PR; the intended verdict is **Approve**.)_

### What's good

- **Right approach.** `focusAnchor` uses the standard `:focus-visible`-polyfill heuristic — document-level `keydown`/`pointerdown` in **capture** phase, so the modality is recorded before the focus the interaction triggers fires. Native order for a mouse click is `pointerdown` → `mousedown` → `focus`, so `focusViaKeyboard` is reliably `false` by the time the focus handler runs, and the prior `mouseenter`/`mousemove` cursor anchor stands. Both acceptance criteria hold (keyboard focus anchors off the box; a click no longer re-anchors).
- **SSR-safe** — guarded by `typeof document !== 'undefined'`, and the `true` default is the safe one (anchor off the box when modality is unknown).
- **Consistent.** All six re-anchoring focus sites route through the one shared helper, so behavior can't drift between them. I grepped the base branch and confirmed these are the only tooltip-on-focus re-anchor sites — the only other `onfocus` is `NumInput.svelte`, which just tracks an input's focus state. Each updated site still pairs focus with a hover handler, so a mouse user (whose click-focus is now a no-op) still gets the tooltip via `mouseenter`.
- `RewardAffordance.svelte`: moving `hovered = true` inside the `if (reward && anchor)` guard is fine — `onFocus` only fires on the `<button>` (rendered only when `reward` is truthy), and the mouse hover visual is already set by `onmouseenter`.
- Docs note added to the frontend Accessibility section, matching the project's requirement to record cross-cutting design decisions.

### Test coverage

Sufficient. The modality heuristic is unit-tested directly (`tooltip.test.ts` — keyboard→element, pointer→`undefined`, re-tracking, non-element target), with a `#880` regression in `attribute-hover.test.ts` (mouse-click focus is a no-op). Each modality-dependent test establishes its own modality first, so the module-global `focusViaKeyboard` doesn't create cross-test fragility (and vitest isolates module state per file). The per-component focus wiring (`Skills`, `ActiveEffectChips`, `SkillRow`, `RewardAffordance`) leans on this central coverage plus the `ZoneNav` mock rather than a test per component — reasonable here since the wiring is a trivial `const anchor = focusAnchor(ev); if (anchor) …` and the logic lives in one tested place.

- Input modality is a cross-cutting concern that currently lives in `tooltip.svelte.ts`. It's only consumed by tooltips today, so colocating is fine (YAGNI) — but if anything else ever needs "was this keyboard or pointer," consider lifting it into its own small module then.


---

Issue #801, opened 2026-06-16
Deferred crit/dodge/block extensions (post-#178 tracking)


ginderjeremiah:
Tracking issue for the extensions the **#178** crit/dodge/block spike **deliberately deferred** (see [`docs/spikes/178-player-crit-dodge-block.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/178-player-crit-dodge-block.md) → "Out of scope / deferred"). The V1 sub-issues (#797–#800) leave a seam for each; this keeps them on the list so they aren't forgotten. **Standalone, not a sub-issue** — closing the spike must not depend on them. Split each into its own issue (with design discussion where flagged) when actually picked up.

- **Statistics & challenges** — crit/dodge/block counts (`CriticalHits` / `AttacksDodged` / `AttacksBlocked` or similar) on `BattleStats`, the battle-completed resolve wiring into new `EStatisticType` values, and challenge-type hooks. V1 surfaces crit/dodge/block through the combat log only. *Needs design (which counts, per-enemy vs global).*
- ~~**Enemy crit/dodge/block** — V1 is player-only by decision (gated on the acting battler). Extending to enemies adds a crit draw on enemy attacks and dodge/block draws on player attacks — a change to the RNG draw order, safe because no mid-battle RNG state is persisted across versions. *Needs design.*~~ - Not going to do
- ~~**Chance caps / diminishing returns** — the raw chances are uncapped (parity works at any value, but a `≥1` dodge is invincibility). A design cap (e.g. a dodge ceiling) or DR curve is a balance overlay.~~ - Probably not going to do.
- **Derivation-coefficient tuning** — the #178 strawman coefficients (crit/dodge/block derivations) are a conservative starting point; revisit once the mechanic is encountered in play.
- ~~**Icon art** — generate crit/dodge/block icons via the `docs/icon-art.md` pipeline; #528 deferred these "until they are real," which they now are.~~ - Done

ginderjeremiah:
Surfaced in the #798 review ([PR #830](https://github.com/ginderjeremiah/GameServer/pull/830)): with player crit landing in V1, the per-skill `SkillStats.TotalDamage`/`HighestSingleAttack` record the **pre-crit** raw skill contribution (they already recorded pre-Defense), while the battle-level `HighestPlayerAttack`/`PlayerDamageDealt` reflect the **realized** post-crit, post-Defense damage — so a crit shows up in the battle-level stat but not the per-skill ones.

[...]

Issue #788, opened 2026-06-16
Add max battle time to UI


ginderjeremiah:
The max battle time is hardcoded to 2 minutes in the backend simulator.

We need to add a time limit indicator to the UI. Probably needs a design mockup first. Also need to add some sort of display indicator for enemy cooldown...

PR #885, opened 2026-06-17
Fight-screen combat feedback: XP bar, floating numbers, battle timer

ginderjeremiah:
Implements the **Fight Screen Final** design handoff (combat-feedback enhancements) on the real Mirror fight screen. The handoff bundle was fetched, its README/chat read, and the three shipping treatments built into the live components (not a copy of the prototype's structure).

Addresses the UI-indicator portion of #788 (the design handoff is the mockup that issue asked for).

### 1. Player XP bar (player card header)
- New shared **`XpBar`** component (`$components`) — a slim gold track filling toward the next level, with a clipped specular shimmer and a gold flash on level-up. Like `HpBar`, it's a specialised multi-layer bar rather than built on the generic `Bar`.
- The player card shows account `LV` + the bar; the enemy card keeps the plain `LV · n`.
- `PlayerManager.nextLevelThreshold` is now **public** so the fill reuses the same threshold the level-up logic uses (no re-deriving `level * EXP_PER_LEVEL` in the view).
- Built reusable so the Attributes screen can mirror it (right-anchored) later.

### 2. Floating combat numbers
- New **`CombatFloaters`** layer over each card (player, enemy, and boss), colour/label-coded by outcome — `CRIT` (gold), `DODGE` (neutral), `BLOCK` (green), and plain hits (accent outgoing / enemy hue incoming). Includes the placeholder icon square for the real combat icons coming later (see #801).
- Driven by a new **`onCombatFloat`** engine hook (alongside `onBattleStageChanged`): `BattleEngine.logicalUpdate` emits one `CombatFloatEvent` per skill activation, derived from the same flags as the combat-log outcome. The shared `battleStep`/parity surface is untouched, so FE/BE battle parity is unaffected.
- The layer is `aria-hidden` (the combat log stays the accessible record); it subscribes client-side via `onMount`.

### 3. 2:00 battle timer (above the VS badge)
- New **`BattleTimer`** component — a stacked mono readout of `battleEngine.timeElapsed` against `DEFAULT_MAX_BATTLE_MS`, with the track flipping to `--warning` in the final 20s and a `TIMEOUT · draw` state at the cap. It sits in a center column over the VS diamond, clear of the zone-nav/boss band.
- **Display-only:** the live engine does not yet enforce the 2-minute timeout as a draw. That mechanic (and its backend parity) is tracked in **#886**.

### Styling / docs
- Repurposed the orphaned `dmg-rise` keyframe for the "punchy" float motion; added `xp-shimmer` and `levelup-flash` keyframes. All colours use existing semantic tokens.
- Copied the handoff artifacts into `docs/designs/game/fight/` (`Fight Screen Final.dc.html`, `Fight Enhancements.dc.html`, `chat-enhancements.md`) and updated `designs.md`.
- Documented the three features under the Fight screen section of `frontend-screens.md`.

## Testing
- New unit tests: `XpBar`, `BattleTimer`, `CombatFloaters`, plus combat-float emission cases in the battle-engine suite; updated `BattlerCard`/`Fight` tests.
- Full unit suite green (**2010 passed**), `eslint` + `svelte-check` clean (0 errors / 0 warnings). Existing fight test-ids are preserved, so e2e selectors are unaffected.

## Follow-ups
- **#886** — enforce the timeout-as-draw battle mechanic (engine + backend parity) so the `TIMEOUT` state is a real outcome, not just a display cap.

https://claude.ai/code/session_01Pcr1KRnWnzS2go9AXqx5gc


UI/src/components/XpBar.svelte
```diff
@@ -0,0 +1,82 @@
...
+// `lastLevel` starts undefined so the first run only records the baseline (no flash on mount).
+let lastLevel: number | undefined;
+$effect(() => {
+	const current = level;
+	if (lastLevel !== undefined && current > lastLevel && fillEl) {
+		fillEl.style.animation = 'none';
+		void fillEl.offsetWidth;
+		fillEl.style.animation = 'levelup-flash 900ms ease-out';
+	}
+	lastLevel = current;
```
ginderjeremiah, line 50: Nit (optional): the level-up flash baseline logic here — "record on mount, flash only when the level rises" — is the one piece of real logic in `XpBar` that isn't covered by `XpBar.test.ts` (the suite covers the fill %, clamping and ARIA, but not the flash). It's a presentational side-effect so this is genuinely optional, but since the `lastLevel !== undefined` mount-guard is exactly the kind of thing a later edit could "simplify" and regress, a small test (mount at level N, re-render at N+1, assert `fillEl.style.animation` was set; and assert it's *not* set on first mount) would lock the intent in. Take it or leave it.

---

ginderjeremiah (commented):
## Review: Fight-screen combat feedback — XP bar, floating numbers, battle timer (UI portion of #788)

**Verdict:** Approve _(submitted as a `COMMENT` because GitHub won't let me gate my own PR — there are no blocking findings; this is an approval in intent)._

This is a clean, well-scoped, well-tested frontend change. I checked it against `frontend.md` and the battle-parity rules in `frontend.md`/`game-design.md`, and it holds up on every axis that matters here.

### Why it's safe on parity (the thing I scrutinised hardest)
- `notifyCombatFloat` fires inside `BattleEngine.logicalUpdate` (the live render loop) — **not** inside the shared `battleStep` or the headless `BattleSimulator` — so the tick-for-tick replay surface is genuinely untouched.
- `combatFloatEvent` is a pure mirror of `damageLogOutcome`: same inputs (`byPlayer/crit/dodged/blocked`), identical branch order, emitted in the same loop iteration as the log line. Outcome coding stays in lockstep with the combat log by construction.
- The floater spawn position uses `Math.random()`, not the seeded `Mulberry32`, so it can't perturb the RNG draw order. No backend test mirror is required for a FE-only presentational hook.

### Guideline adherence
- **Specialised bar, not a `Bar` copy:** `XpBar` staying standalone is the documented pattern — `Bar.svelte`'s own header says "multi-layer bars (`HpBar`) stay specialised," and the clipped-to-fill shimmer + level-up flash genuinely need that. Consistent with `HpBar`, not a violation of "extend, don't copy."
- **Theming:** every colour is a semantic token; I verified each referenced token (`--gold`, `--block-color`, `--enemy-accent`, `--accent-light`, `--warning`, `--text-secondary/-muted/-primary`, …) is actually defined in `+layout.svelte`. No hard-coded hex.
- **DRY:** making `PlayerManager.nextLevelThreshold` public so the bar reuses the level-up threshold (instead of re-deriving `level * EXP_PER_LEVEL` in the view) is the right call. `DEFAULT_MAX_BATTLE_MS` is consumed from codegen, not hand-rolled.
- **Repurposed `dmg-rise` is safe:** confirmed it was genuinely orphaned on `main` (zero consumers) and is now used only by `CombatFloaters`.
- **Boss/enemy floaters don't double up:** the enemy `BattlerCard` and `BossBattlerCard` are mutually exclusive (`{:else}`), so only one `side="enemy"` layer is ever mounted.

### Test coverage
Solid and to the project's bar. `XpBar`/`BattleTimer`/`CombatFloaters` each get their own unit suite (fill %, clamping, ARIA, clock formatting, warning-hue threshold, cap/timeout state, per-side spawn + outcome coding + removal), the battle-engine suite gains the four float-emission cases mirroring the log outcomes, and `BattlerCard`/`Fight` are updated for the new header + timer. One small gap noted inline (the level-up flash logic in `XpBar` is the only untested branch) — optional.

### Nits
- `XpBar.svelte:42-50` — level-up flash logic is untested (inline; optional).
- `CombatFloaters.svelte` — `size: crit ? 30 : labelOnly ? 20 : 21` is a nested ternary; fine as-is, minor readability only.

### Out of scope (no action needed)
- The `TIMEOUT · draw` readout is **display-only** — the live engine doesn't yet enforce the cap, so a true stalemate would pin the readout while the fight runs on. This is correctly called out in the PR body and tracked in #886; good follow-up hygiene.
- #788 also mentioned an enemy-cooldown indicator, which this PR intentionally doesn't cover. The PR scopes itself to "the UI-indicator portion," so #788 stays open for that piece — just flagging it isn't fully closed here.

Nice work — happy to see this merge once the nits are considered.

---

PR #899, opened 2026-06-18
Treat socket auth-retry budget exhaustion as terminal (#888)

### LINKED ISSUE #888: Socket keepalive ping reconnects after the auth-retry budget is exhausted, bypassing the bound
## Problem

The socket auth-rejection retry path is deliberately bounded (`MAX_SOCKET_AUTH_RETRIES`, 5 in a row) so a server that keeps rejecting a freshly-refreshed token "can't loop and burn rotating refresh tokens" (see `docs/frontend.md` → _Socket connection lifecycle_). But the keepalive ping can defeat that bound.

In `UI/src/lib/api/api-socket.ts`, when the budget is hit, `handleClose` logs and returns **without stopping the ping interval**:

```ts
if (this.socketAuthRetries >= MAX_SOCKET_AUTH_RETRIES) {
    console.warn(`Socket auth retry limit (${MAX_SOCKET_AUTH_RETRIES}) reached; not refreshing again.`);
    return;   // <-- stopPingInterval() / handleAuthFailure() never called
}
```

`attemptPing` only self-terminates when there is **no access token**:

```ts
public attemptPing() {
    if (!getAccessToken()) { this.stopPingInterval(); return; }
    void this.ensureSocket();   // <-- still fires every PING_INTERVAL_MS
    ...
}
```

Because the access token is still present (the *refresh* succeeded; it's the *handshake* that keeps being rejected), the 10s ping keeps calling `ensureSocket()` → `openSocket()`, which reconnects and pre-emptively refreshes. These reconnects come from the ping path, which never increments `socketAuthRetries`, so the "5 in a row" cap is bypassed entirely and the very unbounded refresh/reconnect loop the bound exists to prevent resumes — once the access token expires, each cycle's pre-emptive refresh burns a single-use refresh token.

## Suggested direction

Treat budget exhaustion as terminal: when the cap is hit, call `stopPingInterval()` (and most likely `handleAuthFailure()` to route the user to re-auth) rather than just returning — or gate `attemptPing`'s `ensureSocket()` call on the remaining auth-retry budget so the keepalive can't resurrect a socket the bound has given up on.

Add a unit test covering "budget exhausted → ping does not reconnect" (relates to the socket race-ordering gaps called out in #810).

### Labels
bug, claude, scope: small

ginderjeremiah:
## Summary

Fixes #888. The socket keepalive ping could bypass the `MAX_SOCKET_AUTH_RETRIES` bound after it was exhausted.

When the handshake keeps being auth-rejected, `handleClose` refreshes the token pair and reconnects, but only `MAX_SOCKET_AUTH_RETRIES` (5) times in a row. On exhaustion it previously just logged and `return`ed **without stopping the keepalive interval**:

```ts
if (this.socketAuthRetries >= MAX_SOCKET_AUTH_RETRIES) {
    console.warn(...);
    return;   // <-- interval left running
}
```

Because the *refresh* succeeded (it's the *handshake* that keeps being rejected), the access token is still present, so `attemptPing` never self-terminates and keeps calling `ensureSocket()` → `openSocket()` every 10s. Those reconnects come from the ping path, which never increments `socketAuthRetries`, so the "5 in a row" cap is bypassed entirely and the unbounded refresh/reconnect loop the bound exists to prevent resumes — burning a single-use refresh token each cycle once the access token expires.

## How it addresses the issue

Treats budget exhaustion as **terminal**: on the cap, call `stopPingInterval()` and `handleAuthFailure()` (route the user to re-auth), mirroring the existing refresh-failure branch and the documented "unrecoverable auth failure tears down the keepalive" lifecycle. With the interval cleared, the keepalive can no longer resurrect a socket and bypass the bound.

This is the issue's recommended primary direction (stop the ping + route to re-auth) rather than the alternative of gating `attemptPing` on the remaining budget — it's consistent with how the sibling unrecoverable-auth paths already behave.

## Changes

- `UI/src/lib/api/api-socket.ts` — on budget exhaustion, stop the keepalive and route to re-auth.
- `UI/src/tests/lib/api/api-socket.test.ts` — new test: budget exhausted → keepalive interval cleared, `handleAuthFailure` invoked, and advancing past the ping cadence opens no new socket.
- `docs/frontend.md` — note that budget exhaustion is terminal in the socket-lifecycle section.

## Test plan

- [x] `npm run test:unit` (`api-socket.test.ts`) — 36 passed (35 existing + 1 new).

Closes #888.

---
_Generated by [Claude Code](https://claude.ai/code/session_013DNM9w91Ljy9WjKdYk5BAG)_

### DIFF
--- a/UI/src/lib/api/api-socket.ts
+++ b/UI/src/lib/api/api-socket.ts
@@ -334,6 +334,13 @@ export class ApiSocket {
 
 		if (this.socketAuthRetries >= MAX_SOCKET_AUTH_RETRIES) {
 			console.warn(`Socket auth retry limit (${MAX_SOCKET_AUTH_RETRIES}) reached; not refreshing again.`);
+			// Budget exhausted: repeated freshly-refreshed tokens are still being rejected at the handshake,
+			// so the session is effectively unrecoverable. Stop the keepalive and route to re-auth (mirroring
+			// the refresh-failure branch below) — otherwise the ping would keep calling ensureSocket on the
+			// still-present access token, reconnecting via a path that never increments socketAuthRetries and
+			// so bypassing this very bound.
+			this.stopPingInterval();
+			handleAuthFailure();
 			return;
 		}
 
--- a/UI/src/tests/lib/api/api-socket.test.ts
+++ b/UI/src/tests/lib/api/api-socket.test.ts
@@ -674,6 +674,39 @@ describe('ApiSocket', () => {
 			warnSpy.mockRestore();
 		});
 
+		it('stops the keepalive and routes to re-auth when the retry budget is exhausted', async () => {
+			vi.useFakeTimers();
+			const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+			try {
+				setTokens({ accessToken: 'a', refreshToken: 'r' });
+				vi.mocked(refreshTokens).mockResolvedValue({ accessToken: 'a2', refreshToken: 'r2' });
+
+				apiSocket.sendSocketCommand('DefeatEnemy', { timestamp: 1 });
+				await flushMicrotasks();
+				expect(internals(apiSocket).pingIntervalId).not.toBeNull();
+
+				// Reject every reconnect pre-open so the budget is spent and the next close hits the cap.
+				for (let i = 0; i < 6; i++) {
+					rejectHandshake(lastWs());
+					await flushMicrotasks();
+				}
+
+				// Exhausting the budget is terminal: the keepalive is torn down and the user is routed to
+				// re-auth, rather than being left to silently reconnect.
+				expect(internals(apiSocket).pingIntervalId).toBeNull();
+				expect(handleAuthFailure).toHaveBeenCalled();
+
+				// With the interval cleared, the keepalive can no longer fire to resurrect a socket and
+				// bypass the bound: advancing past the ping cadence opens no new connection.
+				const socketsBefore = socketInstances.length;
+				await vi.advanceTimersByTimeAsync(30000);
+				expect(socketInstances.length).toBe(socketsBefore);
+			} finally {
+				vi.useRealTimers();
+				warnSpy.mockRestore();
+			}
+		});
+
 		it('refills the retry budget only after the connection stays open (stable), not on open', async () => {
 			vi.useFakeTimers();
 			try {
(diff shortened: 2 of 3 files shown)


ginderjeremiah:
## Review: treat socket auth-retry budget exhaustion as terminal (#888)

**Verdict: Approve** — the fix is correct, minimal, and consistent with the existing unrecoverable-auth lifecycle; CI is green. (Posting as a summary comment rather than a formal review since I'm the PR author.)

### Correctness
The fix is right. On hitting `socketAuthRetries >= MAX_SOCKET_AUTH_RETRIES`, `handleClose` now calls `stopPingInterval()` + `handleAuthFailure()` before returning, so the keepalive can no longer fire `attemptPing → ensureSocket → openSocket` on the still-present access token via a path that never increments `socketAuthRetries`. That was exactly the bound-bypass described in #888.

This is the issue's recommended primary direction (stop the ping + route to re-auth) and it correctly mirrors the two sibling unrecoverable-auth branches — `openSocket`'s pre-emptive-refresh-failure branch and `handleClose`'s refresh-returns-null branch both already do the same `stopPingInterval()` + `handleAuthFailure()` pair. Choosing this over gating `attemptPing` on the remaining budget keeps the terminal-teardown behavior in one place rather than splitting the bound's enforcement across two methods; good call.

### Test coverage
Adequate and well-targeted. The new test in `UI/src/tests/lib/api/api-socket.test.ts` drives 6 pre-open handshake rejections so rejections 1–5 spend the budget and the 6th lands on the cap branch — the off-by-one is correct (the check is `>=`, evaluated before the increment). It asserts the three things that matter: `pingIntervalId` is cleared, `handleAuthFailure` was invoked, and advancing past the ping cadence (30s) opens no new socket — the last assertion directly proves the bypass is closed, not just that a flag flipped. Structure mirrors the existing "stops refreshing after the retry limit" sibling test.

### Docs
`docs/frontend.md` socket-lifecycle bullet updated to state budget exhaustion is terminal. Accurate and in the right section.

### Nits
None blocking. The added comment is detailed but consistent with this file's deliberately heavy-comment house style, and it documents a non-obvious bypass — appropriate to keep.

### Merge-ordering note (re: #900)
This PR and #900 both edit the **same budget-exhausted branch** in `handleClose`. They're complementary, but whichever merges second will need conflict resolution on that branch. Flagging so the second merge is reviewed for a clean combination rather than a blind resolve.

### CI
All check runs green: `all-checks-passed`, `test-frontend`, and `test-e2e` (chromium/firefox/webkit) succeeded; `test-backend` skipped (no backend changes). The legacy combined-status API reads "pending" only because nothing populates it — the check runs are the real signal.

---

PR #900, opened 2026-06-18
Settle queued-but-unsent socket commands on a terminal close (#889)

### LINKED ISSUE #889: Socket commands queued during a connect that fails before `onopen` can hang forever
## Problem

`openSocket()` (`UI/src/lib/api/api-socket.ts`) resolves its promise as soon as `new WebSocket(url)` is constructed — i.e. while `readyState === CONNECTING`, before `onopen`. So the `await this.ensureSocket()` in `processCommandQueue` returns while the socket is not yet `OPEN`:

```ts
private async processCommandQueue() {
    await this.ensureSocket();
    if (socket && socket.readyState === socket.OPEN) {   // <-- false during CONNECTING
        // send loop + timeout arming live here
    }
}
```

The send loop (which is also where the per-request 30s timeout is **armed**) is skipped, and the actual flush relies entirely on `onStart` (the `onopen` handler). If the socket goes `CONNECTING → error → close` **without ever opening**, `onStart` never fires.

`handleClose` then calls `settleInFlightRequests`, which only settles requests that were already *sent* — the not-yet-sent `socketCommandQueue` is left untouched, and no timeout was ever armed for those commands. The auth-retry branch recovers the common case (it re-runs `processCommandQueue` after a refresh), but several close branches do **not** retry and leave queued commands stranded with no timeout:

- a `NORMAL_CLOSURE` received before the socket ever opened (returns after `stopPingInterval`),
- `!getRefreshToken()` (an anonymous / token-less caller),
- the auth-retry budget already exhausted.

In those cases the awaiting caller's `getResponse()` never resolves and never rejects — a permanent hang.

## Suggested direction

On a close/error for a socket that never opened, also settle (or bounded-retry) the queued-but-unsent `socketCommandQueue`, OR have `openSocket` await the `onopen`/`onerror` resolution so `connecting` reflects true readiness and `processCommandQueue` either sends or fails deterministically. Either way, ensure no code path leaves a queued command with neither a response nor an armed timeout.

Relates to the socket race-ordering test gaps in #810.

### Labels
bug, claude, scope: medium

ginderjeremiah:
## Summary

Fixes #889. A socket command queued while the connection is still `CONNECTING` (before `onopen`) could hang its caller forever if that connect failed before it ever opened.

`openSocket()` resolves as soon as `new WebSocket(url)` is constructed (`readyState === CONNECTING`), so `processCommandQueue`'s `socket.readyState === OPEN` guard is false and the command stays in `socketCommandQueue` — unsent, and with **no per-request timeout armed** (the timeout is armed only at send time). Flushing then relies entirely on `onStart` (`onopen`). If the socket goes `CONNECTING → error → close` without ever opening, `onStart` never fires, and `handleClose`'s `settleInFlightRequests` only settles already-**sent** requests. The auth-retry branch re-flushes the queue on reconnect, but the close branches that don't reconnect left the queued command stranded with neither a response nor a timeout — `getResponse()` never resolves or rejects.

## How it addresses the issue

Settle the queued-but-unsent commands via the existing resolve-with-error contract on every **terminal** close (one that won't reconnect and re-flush):

- normal closure received before the socket opened,
- auth-retry budget already exhausted,
- no refresh token to recover with,
- the reconnect refresh itself failed.

The reconnecting paths (the keepalive ping after a post-open drop, and the auth-retry after a successful refresh) still **keep** the queue so they can re-send it — so the common recovery case is unchanged. This is the issue's option A (settle on terminal close); I preferred it over option B (have `openSocket` await `onopen`/`onerror`) because it's targeted and low-risk, whereas reworking when `connecting` resolves would change the timing of the single-flight open / pre-emptive-refresh race the existing tests pin down.

I also extracted the duplicated `'Connection lost. Please try again.'` string into a `CONNECTION_LOST_ERROR` constant now that it has multiple call sites.

## Changes

- `UI/src/lib/api/api-socket.ts` — add `settleQueuedRequests`; call it in each terminal `handleClose` branch; extract `CONNECTION_LOST_ERROR`.
- `UI/src/tests/lib/api/api-socket.test.ts` — new `describe` covering all four terminal-close branches: a queued-but-unsent command settles with an error (no hang) for normal closure / budget exhausted / no refresh token / failed refresh.
- `docs/frontend.md` — note that the unsent queue is re-flushed only on a reconnecting close and settled on a terminal one.

## Test plan

- [x] `npx vitest run src/tests/lib/api/api-socket.test.ts` — 39 passed (35 existing + 4 new).

## Note for the reviewer

This touches `api-socket.ts`'s `handleClose`, the same method as the still-open PR #899 (#888, budget-exhaustion). The two changes are complementary (this settles the queue; #899 stops the keepalive + routes to re-auth on exhaustion) but both edit the budget-exhausted branch, so whichever merges second will need a small conflict resolution there.

Closes #889.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DmH9thX79YjkSB3ZX7oQne)_


ginderjeremiah:
## Review: settle queued-but-unsent socket commands on a terminal close (#889)

**Verdict: Approve (with one nit)** — the change is correct, well-scoped, and precisely matches the issue's enumeration of stranded close paths. CI is green (frontend, e2e chromium/firefox/webkit all pass; backend skipped — no backend changes; `all-checks-passed` succeeded).

### Correctness of the settling logic

The fix is sound. `settleQueuedRequests` drains the queue via `socketCommandQueue.shift()` and settles each with the resolve-with-error contract, mirroring the existing `settleInFlightRequests`. It is wired into exactly the four **terminal** branches of `handleClose` — normal closure, auth-retry budget exhausted, no refresh token, and a failed reconnect refresh — and the two **reconnecting** paths (`socketOpened`; successful refresh → `processCommandQueue`) deliberately keep the queue to re-send. That maps 1:1 onto the stranded paths called out in #889, and each terminal branch `return`s immediately after settling, so there's no path that both settles and later re-flushes the same command.

Choosing the issue's option A (settle on terminal close) over option B (await `onopen`/`onerror` in `openSocket`) is the right call and well-justified in the PR body — it avoids disturbing the single-flight-open / pre-emptive-refresh timing the existing tests pin down.

### Test coverage

Strong. Four new tests, one per terminal branch, each driving the exact CONNECTING-then-close state from the bug and asserting `response.error` is defined (i.e. no hang), plus the right collaborator expectations (`refreshTokens` / `handleAuthFailure` called or not). This satisfies the project's frontend testing rule (lib module changes covered under mirrored `tests/` path) and directly exercises the previously-stranded code paths.

### Nit (non-blocking)

- `UI/src/lib/api/api-socket.ts` — `settleQueuedRequests` and `settleInFlightRequests` are now near-identical drain-and-settle loops over the two collections. Not worth a refactor here (one drains an array, the other a Map and clears timers), but worth a mention.

### Cross-PR conflict (acknowledged in the PR body)

This edits the budget-exhausted branch of `handleClose`, the same branch PR #899 (#888) touches. The changes are complementary — this settles the queue, #899 stops the keepalive and routes to re-auth on exhaustion — but whichever merges second will need a small manual conflict resolution in that branch. Flagging so it isn't auto-merged blindly.

### What's good

- Comments are genuinely helpful: each branch explains *why* it settles vs. keeps the queue, which is the kind of context that prevents a future "simplification" from re-introducing the hang.
- Extracting `CONNECTION_LOST_ERROR` once it had multiple call sites is a clean DRY win.
- `docs/frontend.md` updated to document the re-flush-vs-settle distinction.

No blocking findings.

---

ginderjeremiah:
## Review — Request changes (rebase required)

### Verdict
**Request changes** — not on the merits (the change is correct and cleanly done) but because the sibling PR it warned about, **#899 (#888), has now merged into `main`**, so this branch no longer applies cleanly and needs a rebase + small conflict resolution before it can land.

### What the PR does
Fixes #889: a command queued while the socket is still `CONNECTING` (before `onopen`) sits unsent in `socketCommandQueue` with **no per-request timeout armed** (the timeout is armed only at send time). If the connect goes `CONNECTING → error → close` without ever opening, `onStart` never fires to flush it, and `settleInFlightRequests` only settles already-*sent* requests — so on a close branch that doesn't reconnect, the caller's `getResponse()` hangs forever.

The fix (issue's option A) adds `settleQueuedRequests`, which drains `socketCommandQueue` and settles each entry through the existing resolve-with-error contract, and calls it on every **terminal** `handleClose` branch: normal closure before open, auth-retry budget exhausted, no refresh token, and a failed reconnect refresh. The reconnecting paths (keepalive after a post-open drop, auth-retry after a successful refresh) deliberately keep the queue so they can re-send it. Also extracts the duplicated `'Connection lost. Please try again.'` string into `CONNECTION_LOST_ERROR`, and updates `docs/frontend.md`.

This is the right call over option B (awaiting `onopen`/`onerror` in `openSocket`), which would change the single-flight-open / pre-emptive-refresh timing the existing tests pin down.

### Code quality
- Settlement reuses the established resolve-with-error contract — consistent with `settleInFlightRequests`/`timeoutRequest`, so callers surface it via `response.error` as they already do. Good.
- `settleQueuedRequests` correctly drains via `shift()` and empties the queue; no risk of a re-flush picking up stale entries.
- Tests cover all four terminal branches with a realistic CONNECTING-socket setup (`queueOnConnectingSocket`), asserting both the settle-with-error and that no spurious refresh/auth-failure side effects fire. Solid coverage; matches the project's test-coverage expectations.
- Doc update is accurate and appropriately concise.
- No nits in the code itself.

### Must-fix (blocker) — #899 overlap is now a real conflict
This was authored against pre-#899 `main` (base `91ff50f4`). **#899 merged at 2026-06-18T07:37Z**, and its changes are now on `main` in the exact two spots this PR edits:

1. **Budget-exhausted branch** — `main` now has `stopPingInterval()` + `handleAuthFailure()` there (from #899). This PR's diff context still shows only the old `console.warn(...) + return` and inserts `settleQueuedRequests(...)` after the warn. The resolution: keep #899's `stopPingInterval()`/`handleAuthFailure()` **and** add `this.settleQueuedRequests(CONNECTION_LOST_ERROR);` before routing to re-auth.
2. **Refresh-failure branch comment** — #899 left the original comment; this PR rewords it. Minor, but it's a second overlapping hunk.

Please rebase onto current `main`, re-apply `settleQueuedRequests` in the (now terminal) budget-exhausted branch alongside #899's teardown, re-run `api-socket.test.ts` + lint, and re-request review. I'm holding the merge to avoid a tangled resolution.

Everything else looks good — once it's rebased cleanly this should be a quick approve.

---

Issue #915, opened 2026-06-18
AttributeAllocationsChangedHandler and ModAppliedHandler share the non-atomic-insert race fixed in #891


ginderjeremiah:
## Problem

While implementing #891 (making the write-behind upsert handlers idempotent against a concurrent double-apply), I noticed two more handlers in `Game.DataAccess/PlayerUpdates/Handlers/` carry the **same** non-atomic-insert race that #891 fixed, but were not enumerated in that issue:

- **`AttributeAllocationsChangedHandler`** — loads the player's `PlayerAttribute` rows, then `Add`s any missing `(PlayerId, AttributeId)`. Two concurrent applies of the same at-least-once event can both find a row missing and both insert it, throwing a unique-constraint violation on the `(PlayerId, AttributeId)` primary key.
- **`ModAppliedHandler`** — `ExecuteDeleteAsync` of the slot's existing mod, then `Add`s the new `AppliedMod`. Two concurrent applies can both delete (no-op) and both insert the same `(PlayerId, ItemId, ItemModSlotId)` primary key, throwing a unique-constraint violation.

As with the #891 handlers, this self-heals via the queue's retry policy (the second pass sees the row and converges), so it's not data loss — but it surfaces as logged errors and wasted retries, weakening the idempotency guarantee the write-behind durability design rests on (see `docs/backend-persistence.md`).

## Suggested direction

Apply the same treatment landed in #891: catch the Postgres unique-violation as a benign no-op (re-using the `UpsertConcurrency.IsUniqueViolation` helper added there), with the existing primary keys as the backing constraint. `ModAppliedHandler`'s delete-then-insert needs the catch around its insert; `AttributeAllocationsChangedHandler`'s mixed update/insert/remove batch can mirror `ProgressUpdatedHandler`'s clear-and-re-apply-once pattern so the conflicting insert resolves to an update on the second pass.

Add concurrency tests mirroring `PlayerUpdateHandlerIdempotencyIntegrationTests` (apply the same event through several scopes at once; assert no throw and a single converged row).

### Labels
bug, claude, tech debt, scope: small

PR #916, opened 2026-06-18
Make write-behind upsert handlers atomically idempotent (#891)

### LINKED ISSUE #891: Write-behind player/progress handlers aren't fully idempotent under concurrent re-apply (duplicate insert + `ToDictionary` throw)
## Problem

The write-behind design leans explicitly on its handlers being idempotent: the at-least-once read "is safe because the handlers are idempotent (existence checks, absolute updates, delete-then-insert)", and startup reclaim "at worst [applies] a still-live item twice" (`docs/backend-persistence.md`). The drain gate only serializes drains **within one instance**; cross-instance same-player ordering is best-effort, and the reclaim path can legitimately re-apply an item another instance is still processing.

But several handlers implement idempotency with a non-atomic **read-then-write**, so two concurrent applies of the same event can both pass the existence check and both insert:

- `Game.DataAccess/PlayerUpdates/Handlers/ItemUnlockedHandler.cs`, `SkillUnlockedHandler.cs`, `ModUnlockedHandler.cs` — `AnyAsync(...)` → `Add` → `SaveChangesAsync`.
- `LogPreferenceChangedHandler.cs` — `ExecuteUpdateAsync` → conditional `Add`.
- `ProgressUpdatedHandler.cs` — load touched rows → `Add` missing.

A concurrent double-apply throws a unique-constraint violation; the retry policy re-runs and converges (the second pass sees the row), so it self-heals — but it surfaces as logged errors and wasted retries, weakening the idempotency guarantee the durability design rests on.

Worse, `ProgressUpdatedHandler` builds its lookup with `ToDictionary`, which **throws on a duplicate key**:

```csharp
var byKey = existing.ToDictionary(ps => (ps.StatisticTypeId, ps.EntityId));
...
var byId = existing.ToDictionary(pc => pc.ChallengeId);
```

If a race ever lands two rows for the same `(player, statistic/entity)` or `(player, challenge)` key, this handler then **permanently** fails for that player — every retry throws and the event eventually dead-letters, turning a transient race into a stuck/poisoned player-progress stream (and progress gates zone navigation).

## Suggested direction

- Make the upsert handlers genuinely atomic — Postgres `INSERT ... ON CONFLICT DO NOTHING/UPDATE`, or catch the duplicate-key violation as a benign no-op — backed by the appropriate DB unique indexes (player+item, player+skill, player+mod-slot, player+statistic-type+entity, player+challenge).
- Defend `ProgressUpdatedHandler`'s grouping against duplicates (`GroupBy`/`ToLookup` taking the first) so a duplicate row can never poison the stream; a unique index makes the duplicate impossible in the first place.

While in these files, add `AsNoTracking()` to the pure existence checks (`AnyAsync`) — they run on the per-battle persistence hot path and never need tracking (matching `PlayerRepository.GetPlayerFromDb`).

### Labels
bug, claude, scope: medium

ginderjeremiah:
## Summary

Fixes #891. The write-behind durability design leans explicitly on its handlers being idempotent (the at-least-once queue read and startup reclaim "at worst apply a still-live item twice" — `docs/backend-persistence.md`), but several handlers implemented idempotency with a non-atomic **read-then-write**: two concurrent applies of the same event could both pass the existence check and both insert, throwing a unique-constraint violation. The queue's retry policy re-runs and converges, so it self-heals — but it surfaces as logged errors and wasted retries, weakening the guarantee the design rests on. `ProgressUpdatedHandler` additionally built its lookups with `ToDictionary`, which **throws on a duplicate key** and could permanently poison (dead-letter) a player's progress stream.

All the needed unique constraints already exist — composite PKs on `UnlockedItem (PlayerId, ItemId)`, `UnlockedMod (PlayerId, ItemModId)`, `PlayerSkill (PlayerId, SkillId)`, `LogPreference (PlayerId, LogTypeId)`, `PlayerChallenge (PlayerId, ChallengeId)`, plus the unique index on `PlayerStatistic (PlayerId, StatisticTypeId, EntityId)` — so **no migration is required**; the fix is making the handlers atomic against those constraints.

## How it addresses the issue

- **`ItemUnlocked` / `SkillUnlocked` / `ModUnlocked`** — keep the existence check as the fast path for the common re-apply (now `AsNoTracking`, matching `PlayerRepository.GetPlayerFromDb`) and absorb the rare concurrent double-insert as a benign no-op. The catch reuses a shared `UpsertConcurrency.IsUniqueViolation` helper, the same `PostgresErrorCodes.UniqueViolation` shape already used in `Repositories/Users.cs`.
- **`LogPreferenceChanged`** — keeps the absolute-update-first upsert; on a lost insert race it clears the failed insert and re-runs the absolute update so the event's value still lands on the now-existing row.
- **`ProgressUpdated`** — defends both grouping lookups with `GroupBy`-first (instead of `ToDictionary`) so a stray duplicate row can never throw, and clears-and-re-applies once on a unique violation so the conflicting insert resolves to an update on the second pass (a second failure propagates to the queue retry policy rather than looping).

I chose **catch-the-unique-violation** over raw `INSERT … ON CONFLICT` to stay consistent with the existing `Users.cs` concurrency pattern and avoid hand-written SQL with quoted identifiers; both are listed as acceptable in the issue.

## Changes

- `Game.DataAccess/PlayerUpdates/Handlers/UpsertConcurrency.cs` — new shared `IsUniqueViolation` helper.
- `ItemUnlockedHandler` / `SkillUnlockedHandler` / `ModUnlockedHandler` — `AsNoTracking` existence check + benign unique-violation catch.
- `LogPreferenceChangedHandler` — re-run the absolute update on a lost insert race.
- `ProgressUpdatedHandler` — `GroupBy`-first lookups + clear-and-re-apply-once on unique violation.
- `Game.Application.Tests/DataAccess/PlayerUpdateHandlerIdempotencyIntegrationTests.cs` — sequential re-apply + concurrent double-apply (8-way) tests per handler. I confirmed the concurrency tests are real regression guards: temporarily removing a handler's catch makes the corresponding test fail.
- `docs/backend-persistence.md` — documents that idempotency now holds under a concurrent cross-instance double-apply.

## Note for the reviewer

While in these files I found that **`AttributeAllocationsChangedHandler`** (load-then-`Add`) and **`ModAppliedHandler`** (delete-then-insert) carry the *same* non-atomic-insert race but weren't enumerated in #891. To keep this PR scoped to what the issue listed, I filed #915 to apply the identical treatment there rather than expanding scope here.

## Test plan

- [x] `dotnet test` (whole solution) — Core 630, Application 294 (incl. 8 new), Infrastructure 38, Api 509; 0 failures.
- [x] Verified each concurrency test fails with its handler's catch removed (the race fires reliably at 8-way parallelism).

Closes #891.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MMTvnHZhTjduX5jPGWgWuh)_


ginderjeremiah:
## Review: write-behind upsert handlers made atomically idempotent (#891)

**Verdict: Approve with nits.** The change does what #891 asked — every read-then-write handler now absorbs the concurrent double-insert as a benign no-op (or clear-and-re-apply), `ProgressUpdatedHandler` no longer throws on a duplicate key, and `AsNoTracking()` is on the hot-path existence checks. It stays consistent with the existing `Repositories/Users.cs` unique-violation pattern, needs no migration, and CI is green (`all-checks-passed` success; backend + all e2e browsers pass; frontend correctly skipped). No blocking findings.

### Correctness — verified
- `LogPreferenceChangedHandler`: catch path clears the tracker (dropping the failed insert so the per-request UoW commit filter won't re-attempt it) then re-runs the `ExecuteUpdateAsync` absolute update — which bypasses the tracker. Correct.
- `ProgressUpdatedHandler`: on unique violation, `ChangeTracker.Clear()` + one re-run reloads the now-existing rows as updates, so the second pass carries no conflicting insert; a second failure escapes to the queue retry policy rather than looping. Correct.
- `UpsertConcurrency.IsUniqueViolation` mirrors the `Users.cs` `PostgresException { SqlState: PostgresErrorCodes.UniqueViolation }` shape exactly.

### Should-fix / follow-up (non-blocking)
- **`PlayerStatistic` global stats and NULL `EntityId`.** The unique index is `(PlayerId, StatisticTypeId, EntityId)` with `EntityId int?` (`GameContext.cs:286`). Postgres treats NULLs as distinct in a default unique index, so for **global** statistics (`EntityId == null`) the index does *not* block two concurrent inserts of the same `(PlayerId, StatisticTypeId, NULL)` — no unique violation fires and a duplicate row can land. The `ProgressUpdatedHandler` concurrency guard leans on that violation, so for the null-entity case the true protection here is the new `GroupBy`-first defense + absolute updates, which keep the duplicate harmless and convergent rather than poisoning the stream (a strict improvement over the prior `ToDictionary` throw). Worth tightening the index to `NULLS NOT DISTINCT` (PG15+) so global stats get the same hard guarantee as entity-scoped ones — this is a pre-existing schema nuance, not introduced here, and out of scope for #891, so I'd track it as a small follow-up rather than expand this PR.
- The new idempotency integration test asserts a single row for the `EntityId = null` case; given the point above, that concurrent assertion is timing-dependent for the null case specifically. Once the index is tightened it becomes a hard guarantee.

### Test coverage
Solid. Per-handler sequential re-apply + 8-way concurrent double-apply integration tests, each apply through its own DI scope to mirror the synchronizer's per-event scope. The author confirmed each concurrency test fails with its handler's catch removed, so these are real regression guards, not smoke tests. Matches the project's integration-test rule for out-of-process (Postgres) interactions.

### What's good
- Tight scoping: `AttributeAllocationsChangedHandler` / `ModAppliedHandler` carry the same race but weren't in #891, so they were split to #915 instead of widening this PR — the right call.
- Doc update to `backend-persistence.md` accurately extends the idempotency guarantee to the concurrent cross-instance case.
- Comments explain the *why* (the non-atomic window, the converge-on-retry reasoning) without being verbose.


---

ginderjeremiah:
## Review: Approve ✅

### Summary
This PR makes the write-behind upsert handlers atomically idempotent, closing the gap between the durability design's stated reliance on idempotent handlers (at-least-once queue read + startup reclaim "at worst apply a still-live item twice") and the handlers' actual non-atomic read-then-write implementation. Two concurrent applies of the same event could both pass the existence check and both insert, throwing a unique-constraint violation that only self-healed via the retry policy (logged errors + wasted retries). `ProgressUpdatedHandler` additionally used `ToDictionary`, which throws on a duplicate key and could permanently dead-letter (poison) a player's progress stream.

The fix:
- **`ItemUnlocked` / `SkillUnlocked` / `ModUnlocked`** — keep the existence check as the fast path (now `AsNoTracking`, matching `PlayerRepository.GetPlayerFromDb`) and absorb the rare concurrent double-insert as a benign no-op via the shared `UpsertConcurrency.IsUniqueViolation` helper.
- **`LogPreferenceChanged`** — on a lost insert race, clears the tracker and re-runs the absolute `ExecuteUpdateAsync` so the event's value still lands on the now-existing row.
- **`ProgressUpdated`** — `GroupBy`-first lookups (so a stray duplicate row can never throw) plus clear-and-re-apply-once on a unique violation; a second failure falls through to the queue retry policy rather than looping.
- New `UpsertConcurrency.IsUniqueViolation` helper extracts exactly the `PostgresErrorCodes.UniqueViolation` shape already inlined in `Repositories/Users.cs`, keeping the two consistent. No migration — all backing constraints already exist.
- Docs (`backend-persistence.md`) updated to state idempotency now holds under a concurrent cross-instance double-apply.

### What I checked
- **Correctness** — the `ChangeTracker.Clear()`-before-re-apply pattern is correct in both `LogPreferenceChangedHandler` (the follow-up `ExecuteUpdateAsync` is a direct SQL update bypassing the tracker, so it lands on the now-existing row) and `ProgressUpdatedHandler` (the conflicting rows reload as updates on the second pass). The catch-the-unique-violation choice over raw `INSERT ... ON CONFLICT` is the right call for consistency with `Users.cs` and avoids hand-written quoted-identifier SQL — both were listed as acceptable in #891.
- **Domain isolation** — changes stay entirely in the data tier (`Game.DataAccess/PlayerUpdates/Handlers`); the helper is `internal`, no entity types leak upward.
- **Tests** — `PlayerUpdateHandlerIdempotencyIntegrationTests` adds sequential re-apply + 8-way concurrent double-apply per handler, each apply through its own DI scope/`GameContext` to mirror the synchronizer and genuinely provoke the race. Referenced infra (`ApplicationIntegrationTestBase.CancellationToken`/`CreateScope`, `TestDataSeeder` factories) all exist. PR author confirmed each concurrency test fails with its handler's catch removed — real regression guards, not vacuous.
- **CI** — all 7 checks green (`test-backend`, `test-e2e` chromium/webkit/firefox, `all-checks-passed`; `test-frontend` correctly skipped — backend-only change).
- **Follow-up** — #915 is filed and reasonable: it covers `AttributeAllocationsChangedHandler` and `ModAppliedHandler`, which carry the identical race but weren't enumerated in #891, keeping this PR correctly scoped.

### Verdict
**Approve.** Clean, well-scoped, faithful to the issue and the established concurrency pattern, properly tested, docs updated, out-of-scope work tracked. No nits, should-fixes, or must-fixes. Merging via squash.

---

PR #918, opened 2026-06-18
Fix CombatFloaters timer leak and duration drift (#893)

### LINKED ISSUE #893: CombatFloaters leaks per-float timers and mutates state after unmount
## Problem

`UI/src/routes/game/screens/fight/CombatFloaters.svelte` schedules a `setTimeout` per spawned floater to remove it from the reactive array, but the timer handles are never tracked or cleared:

```ts
setTimeout(() => { floaters = floaters.filter(...); }, DURATION_MS + 80);
```

The `onMount` cleanup only returns the `onCombatFloat` unsubscribe. The layer spawns many floaters per second across up to three combatant cards, so navigating away mid-fight leaves a burst of pending callbacks that fire after the component is destroyed, each writing the orphaned `$state` array — a timer leak and a write-after-unmount smell.

Separately, the floater lifetime is a magic number duplicated in two places that can silently desync: the JS `DURATION_MS = 1500` (cleanup) and the CSS `animation: dmg-rise 1500ms` — retuning one alone makes floaters vanish early or linger. There's also a placeholder empty `.floater-icon` box shipped on every combat number.

## Suggested direction

- Track the timer ids in a `Set` and clear them in the `onMount` cleanup alongside the unsubscribe — or drop per-float timers entirely and expire floats off a single age check in the existing frame loop.
- Drive the CSS animation duration from the JS constant via a CSS custom property so the two can't drift.
- Remove the placeholder icon box (or file a follow-up if it's a real upcoming feature).

### Labels
bug, claude, scope: small

ginderjeremiah:
## Summary

Fixes #893. `CombatFloaters.svelte` scheduled a `setTimeout` per spawned floater to remove it from the reactive array, but never tracked or cleared those handles — the `onMount` cleanup only returned the `onCombatFloat` unsubscribe. Since the layer spawns many floaters per second across up to three combatant cards, navigating away mid-fight left a burst of pending callbacks that fired after the component was destroyed, each writing the orphaned `$state` array (a timer leak + write-after-unmount). Separately, the floater lifetime was a magic number duplicated in two places that could silently desync: the JS `DURATION_MS = 1500` (removal) and the CSS `animation: dmg-rise 1500ms`.

## How it addresses the issue

- **Timer leak / write-after-unmount** — track each removal `setTimeout` handle in a `Set`; each timer removes its own handle when it fires (so the set stays bounded during a long fight), and the `onMount` cleanup now unsubscribes **and** clears every still-pending timer, so none fire after the component is gone. The `Set` is plain (not `SvelteSet`) because it's pure bookkeeping never read reactively — the same precedent as `ToastContainer`'s `removalTimers`, with the matching `eslint-disable` line.
- **Duration drift** — the `.floaters` container now sets a `--float-duration` custom property from `DURATION_MS`, and the `dmg-rise` animation consumes `var(--float-duration)`, so the JS removal timer and the CSS animation read one source and can't drift. The timer still adds its small grace margin (`DURATION_MS + 80`).

## Note for the reviewer

The issue's third bullet (remove the "placeholder empty `.floater-icon` box shipped on every combat number") looks **stale**: the icon is now a real, implemented feature — `FLOAT_ICON` maps crit/dodge/block to actual art and the template renders the `` only `{#if floater.icon}`, so a plain hit ships no icon box (the existing test asserts exactly this). I left the icon rendering untouched rather than remove a working feature. Happy to revisit if you intended something different here.

## Changes

- `UI/src/routes/game/screens/fight/CombatFloaters.svelte` — track + clear removal timers on unmount; drive the CSS animation duration from `DURATION_MS` via `--float-duration`.
- `UI/src/tests/.../CombatFloaters.test.ts` — assert the `--float-duration` custom property mirrors the JS constant, and that unmount unsubscribes and clears every pending removal timer (so no callback fires after destroy). The shared mock now exposes the unsubscribe spy.

## Test plan

- [x] `npx vitest run` for `CombatFloaters.test.ts` — 9 passed (7 existing + 2 new).

Closes #893.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JummTpxjGZ3hRCbETDNzYv)_

### DIFF
--- a/UI/src/routes/game/screens/fight/CombatFloaters.svelte
+++ b/UI/src/routes/game/screens/fight/CombatFloaters.svelte
@@ -2,7 +2,7 @@
      combat events and, for the events striking this card's side, spawns a short-lived number/label
      that pops and drifts up. Colour and label are coded by outcome (crit/dodge/block/hit). Purely
      presentational (aria-hidden) — the combat log is the accessible record of the same events. -->
-<div class="floaters" aria-hidden="true" data-testid={testId}>
+<div class="floaters" aria-hidden="true" data-testid={testId} style:--float-duration="{DURATION_MS}ms">
 	{#each floaters as floater (floater.id)}
 		<div
 			class="floater"
@@ -51,11 +51,16 @@ const FLOAT_ICON: Partial<Record<CombatFloatEvent['kind'], string>> = {
 	block: '/img/Block Reduction.png'
 };
 
-/** Matches the float animation length in `common.scss` (`dmg-rise`), with a small grace margin. */
+/** Drives both the JS removal timer and the CSS `dmg-rise` animation (via the `--float-duration`
+ *  custom property) so the two can't drift; the timer adds a small grace margin over the animation. */
 const DURATION_MS = 1500;
 
 let floaters = $state<Floater[]>([]);
 let nextId = 0;
+// Pending removal timers, tracked so they can be cleared on unmount (no write-after-destroy).
+// Plain Set — pure bookkeeping that's never read reactively (mirrors ToastContainer's removalTimers).
+// eslint-disable-next-line svelte/prefer-svelte-reactivity
+const removalTimers = new Set<ReturnType<typeof setTimeout>>();
 
 const colorFor = (event: CombatFloatEvent): string => {
 	switch (event.kind) {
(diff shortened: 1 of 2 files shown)


ginderjeremiah:
## Review: Approve ✅

### Summary
Fixes the two genuine defects in `CombatFloaters.svelte` from #893:

1. **Timer leak / write-after-unmount** — each removal `setTimeout` handle is now tracked in a plain `Set`, each timer self-deletes its handle when it fires (keeping the set bounded over a long fight), and the `onMount` cleanup unsubscribes **and** `clearTimeout`s every still-pending timer. No orphaned callback can write the destroyed component's `$state` array after navigation mid-fight.
2. **Duration drift** — the `.floaters` container exposes `--float-duration` from the JS `DURATION_MS`, and the `dmg-rise` animation consumes `var(--float-duration)`, so the JS removal timer and the CSS animation read a single source.

### Verification
- **Stale third bullet confirmed.** The issue's "remove the placeholder empty `.floater-icon` box" is indeed stale: the template renders the `` only `{#if floater.icon}`, and `icon` is populated solely from `FLOAT_ICON` (crit/dodge/block). A plain hit gets `icon: ''` and ships no icon box. Leaving the implemented icon feature untouched was the correct call.
- **Pattern consistency confirmed.** The non-reactive `Set` with `eslint-disable svelte/prefer-svelte-reactivity` mirrors `ToastContainer.svelte`'s `removalTimers` exactly (same disable line, same teardown loop) — appropriate use of an established precedent rather than reaching for `SvelteSet` for non-reactive bookkeeping.
- **CI green.** `all-checks-passed` succeeded; frontend, e2e (chromium/webkit/firefox), and detect-changes all passed. (`test-backend` correctly skipped — no backend changes.)

### Tests
The two new tests are well-targeted: one asserts `--float-duration` mirrors the JS constant (locking the anti-drift contract), the other asserts unmount unsubscribes once and clears both pending removal timers with no throw after advancing past the lifetime. The shared mock now exposes the unsubscribe spy. Good coverage for a scope:small fix.

### Nits / should-fixes / must-fixes
None. Clean, minimal, well-documented (the multi-line comments add genuine context), and consistent with project conventions.

Approving and merging via squash.

---

PR #948, opened 2026-06-19
Fix frontend a11y/interaction gaps (nested button, drag-only reorder, unnamed sliders, hidden focus)

### LINKED ISSUE #4: Fix write-behind caching implementation for player data
Currently, the write-behind caching done in `Game.DataAccess.DataProviderSynchronizer` is not wired up correctly. The pubsub events it relies on are not actually being published anywhere.

There are already domain events set up such that this probably just needs a new domain event handler to translate them to the pubsub events the `DataProviderSynchronizer` is expecting. That being said, I don't believe this to be a particularly great solution as it stands, so you should research the code and see if you can come up with a better solution. If not, then that's fine.

### LINKED ISSUE #913: Frontend a11y/interaction gaps: nested button, drag-only reorder, unnamed sliders, hidden focus
## Problem

Several accessibility/interaction gaps surfaced during a codebase health review. These are distinct from the hover-tooltip keyboard work in #825.

1. **Nested `<button>` inside the challenge type-card (invalid HTML).** `OverviewPane.svelte:45-66` makes the `type-card` a `<button onclick={onPick}>`, and inside it `RewardAffordance variant="chip"` (`RewardAffordance.svelte:4-22`) renders another real `<button>` (the reward tooltip trigger). Button-in-button is invalid HTML — browsers recover unpredictably, focus/activation is ambiguous, and screen readers announce a broken control. `next2` is non-null for any group with incomplete challenges, so this renders in normal use. **Fix:** make `type-card` a non-button container with a single interactive child, or render the inline reward non-interactively with its tooltip trigger hosted separately.

2. **Equipped-skill loadout reorder is drag-only — no keyboard or touch path.** `EquippedBand.svelte:22-129` reorders equipped skills (priority order is *battle-affecting*) exclusively via native HTML5 drag-and-drop, and the header advertises "drag to reorder." HTML5 DnD doesn't work on touch, and there's no keyboard alternative (the OverlayButton swap flow changes *which* skills are equipped but can't reorder existing slots). **Fix:** add keyboard reorder (e.g. Alt+Arrow on a focused card calling `view.reorder`) and/or explicit up/down buttons (which also serve touch). _(Larger item in this set — scope medium on its own.)_

3. **Sandbox range sliders have no accessible name (card-game).** The three `<input type="range">` (AGI/DEX/LCK) in `SandboxStrip.svelte:9-41` have no `<label for>` and no `aria-label` — the labels are adjacent `<span>`s with no programmatic association, so screen readers announce three unnamed sliders. **Fix:** add `aria-label` to each (or wrap in a `<label>`).

4. **Favorite star is focusable but invisible while focused.** `.fav-star` in `GridSlot.svelte:42-54,180-204` is a real `<button>` but `opacity: 0` unless mouse-hovered or favorited. A keyboard user tabbing the grid lands on every unfavorited item's star with zero visual feedback. **Fix:** reveal on focus, e.g. `.fav-star:focus-visible { opacity: 1 }` and/or `.grid-slot:focus-within .fav-star { opacity: .75 }`.

5. **Radar vertex has no keyboard refund/decrement.** `AttributesRadar.svelte:21-39,150-155` `handleKey` only handles Enter/Space → `view.inc(i)`; refund is drag-inward only and `onkeydown` is gated on `clickable` (false at max budget), so a keyboard user can never decrement via the radar — though its own `aria-label` says "drag to refund points." (The Stepper offers keyboard decrement separately, so it isn't wholly lost.) **Fix:** handle ArrowUp/Right → `inc`, ArrowDown/Left → `dec`; keep focusable whenever interactive.

ginderjeremiah:
## Summary

Closes #913. Addresses the five distinct accessibility/interaction gaps surfaced during the codebase health review (all frontend-only — no battle logic or backend touched).

| # | Gap | Fix |
|---|-----|-----|
| 1 | **Nested `<button>` in the challenge type-card** (the reward chip is a real button inside the card button — invalid HTML) | The `type-card` is now a presentational `<div>` with a full-bleed `OverlayButton` primary action (open the type), and the reward chip is raised above it so its own tooltip trigger stays interactive. This is the project's established *accessible-card pattern* (docs/frontend.md → Accessibility). |
| 2 | **Loadout reorder is drag-only** (HTML5 DnD is mouse-only; priority is battle-affecting) | Added explicit move-earlier/move-later buttons on each filled card — real `<button>`s that serve keyboard, touch, and mouse. They reveal on hover/`:focus-within`, sit above the overlay, and disable at the ends. Drag still works. |
| 3 | **Sandbox sliders have no accessible name** | Added `aria-label` (`Agility`/`Dexterity`/`Luck`) to each range input. |
| 4 | **Favorite star focusable but invisible while focused** | Reveal on `.fav-star:focus-visible` (opacity 1) and `.grid-slot:focus-within .fav-star` (opacity .75), in addition to the existing hover/favorited reveal. |
| 5 | **Radar vertex has no keyboard refund/decrement** | `handleKey` now maps Arrow up/right → `inc` and Arrow down/left → `dec`; vertices stay focusable (`tabindex=0`) whenever `interactive`, so a keyboard user can refund even at max budget (the affordance the aria-label already advertised). aria-labels updated to mention arrow keys. |

## How it addresses the issue

Each item maps 1:1 to the corresponding numbered point in #913, using the patterns the docs prescribe (presentational container + `OverlayButton`, real interactive elements over `role`-divs, focus-reveal over hover-only).

## Test plan

- **Unit/component tests** added or updated:
  - `OverviewPane.test.ts` (new) — type-card is a `<div>` with a `<button>` overlay; reward chip is never nested in another button; activation calls `onPick`.
  - `EquippedBand.test.ts` — move buttons present/labelled, disabled at ends, reorder via activation, hidden during swap.
  - `AttributesRadar.test.ts` — Arrow-key inc/dec, refund still works at max budget, focusable regardless of `canInc`, ignored when not interactive.
  - `SandboxStrip.test.ts` (new) — slider aria-labels + `setStat` wiring.
  - `Challenges.test.ts` — updated the two type-card click sites to target the new overlay button.
  - Fix #4 is CSS-only (`:focus-visible`/`:focus-within`); not unit-tested as jsdom doesn't resolve scoped-CSS computed opacity.
- `npm run lint` (ESLint + svelte-check): clean.
- `npm run test:unit:ci`: full suite green (2078 tests).

## Note for review

The move-button reorder (item 2) was chosen over an Alt+Arrow-on-card keyboard handler because explicit buttons cover **both** the keyboard and touch gaps the issue calls out with a single discoverable affordance, and keep the "real interactive elements" philosophy. Happy to revisit if you'd prefer a keyboard modifier path instead.

---
_Generated by [Claude Code](https://claude.ai/code/session_011t1Aud3HkbdiQ8ZLbuzcH1)_

### DIFF
--- a/UI/src/routes/game/screens/attributes/AttributesRadar.svelte
+++ b/UI/src/routes/game/screens/attributes/AttributesRadar.svelte
@@ -25,13 +25,13 @@
 			class:interactive
 			class:dragging={dragIndex === i}
 			role="button"
-			tabindex={clickable ? 0 : -1}
+			tabindex={interactive ? 0 : -1}
 			aria-label={clickable
-				? `Adjust ${attributeName(coreIds[i], staticData.attributes)} — drag to allocate or click to add a point`
-				: `Adjust ${attributeName(coreIds[i], staticData.attributes)} — drag to refund points`}
+				? `Adjust ${attributeName(coreIds[i], staticData.attributes)} — drag, or use arrow keys to allocate and refund points`
+				: `Adjust ${attributeName(coreIds[i], staticData.attributes)} — drag, or use arrow keys to refund points`}
 			onpointerdown={(e) => startDrag(e, i)}
 			onclick={() => onVertexClick(i)}
-			onkeydown={(e) => clickable && handleKey(e, i)}
+			onkeydown={(e) => interactive && handleKey(e, i)}
 		>
 			<circle cx={vertex[0]} cy={vertex[1]} r="11" fill="transparent" />
 			<circle cx={vertex[0]} cy={vertex[1]} r={big ? 4.5 : 3.5} class="dot" style="--c: {attributeColor(coreIds[i])}" />
@@ -148,9 +148,15 @@ const labels = $derived(
 );
 
 function handleKey(e: KeyboardEvent, i: number): void {
-	if (e.key === 'Enter' || e.key === ' ') {
+	// Enter/Space and Arrow-up/right allocate a point; Arrow-down/left refunds one. The view's
+	// inc/dec no-op past their bounds, so a keyboard user can refund even at max budget — the
+	// drag-inward affordance the aria-label advertised but the keyboard previously couldn't reach.
+	if (e.key === 'Enter' || e.key === ' ' || e.key === 'ArrowUp' || e.key === 'ArrowRight') {
 		e.preventDefault();
 		view.inc(i);
+	} else if (e.key === 'ArrowDown' || e.key === 'ArrowLeft') {
+		e.preventDefault();
+		view.dec(i);
 	}
 }
 
--- a/UI/src/routes/game/screens/card-game/loom/SandboxStrip.svelte
+++ b/UI/src/routes/game/screens/card-game/loom/SandboxStrip.svelte
@@ -10,6 +10,7 @@
 			type="range"
 			min="0"
 			max="300"
+			aria-label="Agility"
 			value={game.agi}
 			oninput={(e) => view.setStat('agi', +e.currentTarget.value)}
 		/>
@@ -23,6 +24,7 @@
 			type="range"
 			min="1"
 			max="500"
+			aria-label="Dexterity"
 			value={game.dex}
 			oninput={(e) => view.setStat('dex', +e.currentTarget.value)}
 		/>
@@ -36,6 +38,7 @@
 			type="range"
 			min="0"
 			max="60"
+			aria-label="Luck"
 			value={game.luck}
 			oninput={(e) => view.setStat('luck', +e.currentTarget.value)}
 		/>
--- a/UI/src/routes/game/screens/challenges/OverviewPane.svelte
+++ b/UI/src/routes/game/screens/challenges/OverviewPane.svelte
@@ -42,11 +42,11 @@
 					group.items.filter((c) => c.state !== 'done'),
 					'progress'
 				)[0]}
-				<button
-					class="type-card"
-					style:border-left="2px solid {tintColor(group.accent, 0.6)}"
-					onclick={() => onPick(group.typeId)}
-				>
+				<!-- Accessible card: a presentational container whose primary action is a full-bleed
+				     OverlayButton (open this challenge type), with the reward chip raised above it so
+				     its own tooltip trigger stays interactive — no nested <button>s. -->
+				<div class="type-card" style:border-left="2px solid {tintColor(group.accent, 0.6)}">
+					<OverlayButton label={`View ${group.label} challenges`} onActivate={() => onPick(group.typeId)} />
 					<div class="type-card-head">
 						<RingMeter {pct} accent={group.accent} size={34} stroke={3} done={complete}>
 							<TypeGlyph typeId={group.typeId} color={complete ? 'var(--success)' : group.accent} size={15} />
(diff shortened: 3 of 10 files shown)


ginderjeremiah:
## Code Review — Approve ✅

Reviewed against `CLAUDE.md` and `docs/frontend.md`. This is a clean, well-scoped frontend-only change. All five fixes map 1:1 to the numbered points in #913 and use the patterns the docs prescribe.

### What the PR does
| # | Gap (#913) | Fix | Verdict |
|---|------------|-----|---------|
| 1 | Nested `<button>` in challenge type-card | `type-card` → presentational `<div>` with a full-bleed `OverlayButton`; reward chip raised (`.raised` z-index 2) so its tooltip trigger stays interactive | Exactly the docs' *accessible-card pattern* (frontend.md → Accessibility). The `.raised` class is correctly gated on `next2` so the non-interactive "All unlocked" state stays flat and clicks fall through to the overlay — nice touch. |
| 2 | Loadout reorder is drag-only | Real move-earlier/move-later `<button>`s per filled card, disabled at the ends, raised above the overlay, revealed on `:hover`/`:focus-within` | Sound. See design note below. |
| 3 | Unnamed sandbox sliders | `aria-label` on each range input | Correct and minimal. |
| 4 | Favorite star invisible while focused | `.fav-star:focus-visible { opacity: 1 }` + `.grid-slot:focus-within .fav-star { opacity: .75 }` | Matches the exact remedy the issue suggested. |
| 5 | Radar vertex no keyboard refund | `handleKey` maps Arrow up/right → `inc`, down/left → `dec`; `tabindex`/`onkeydown` gate switched from `clickable` to `interactive` so vertices stay focusable at max budget | Correct. The view's `inc`/`dec` no-op past bounds, so refund-at-max-budget works without extra guards. |

### Design choice: move-buttons vs Alt+Arrow (item 2)
The author's choice is the **right** one per the docs. #913 item 2 calls out *both* a keyboard gap *and* a touch gap (HTML5 DnD is mouse-only). An Alt+Arrow-on-card handler would close only the keyboard half; explicit `<button>`s close both with a single discoverable affordance and uphold the "real interactive elements over `role`-divs" principle (frontend.md → Accessibility). They also slot naturally into the existing accessible-card layout as higher-z-index siblings of the overlay, mirroring the `.rm` remove button. No change requested.

### Quality checks
- **Patterns/consistency:** `OverlayButton` reused (not re-implemented); the new `.reorder`/`.mv` styling mirrors the existing `.rm` button. The radar `<g role="button">` div is the sanctioned exception explicitly named in frontend.md → Accessibility, so leaving it as a role-div is correct.
- **Theming:** All colours go through tokens (`--accent`, `--text-tertiary`, `--attr-*`) and `color-mix`; no hard-coded hex introduced. ✔
- **No dead code:** The `clickable` const remains genuinely load-bearing (aria-label wording + the click `+1` path via `onVertexClick`); only the focusability/keydown gate moved to `interactive`. ✔
- **Tests:** New/updated suites cover all four testable fixes — `OverviewPane.test.ts` (container is a `<div>`, overlay is a `<button>`, no `button button`, activation calls `onPick`), `EquippedBand.test.ts` (labels, end-disabling, reorder via activation, hidden during swap), `AttributesRadar.test.ts` (arrow inc/dec, refund at max budget, focusable regardless of `canInc`, no-op when not interactive), `SandboxStrip.test.ts` (aria-labels + `setStat` wiring), and the two `Challenges.test.ts` click sites retargeted to the overlay. Fix #4 being CSS-only and untested in jsdom is a reasonable, well-documented call.
- **Comments:** A few multi-line comments, all justifying non-obvious a11y/z-index decisions — appropriate, not excessive.

### CI
`all-checks-passed`, `test-frontend`, and `test-e2e` (chromium/firefox/webkit) all green; `test-backend` correctly skipped (frontend-only).

No nits or follow-ups. Solid work — merging.


---

PR #949, opened 2026-06-19
Reject malformed socket frames missing "name" with a structured error (#935)

### LINKED ISSUE #935: Socket command frame missing "name" throws an unobserved ArgumentNullException and hangs the client
## Problem

A structurally-valid JSON frame that omits `"name"` produces an **unobserved server exception** and leaves the client waiting out its full 30s request timeout.

In `SocketHandler.HandleMessage`:

```csharp
// Game.Api/Sockets/SocketHandler.cs:334-359
try
{
    var commandInfo = message.Deserialize<SocketCommandInfo>();
    if (commandInfo is not null)
    {
        if (_commandFactory.IsServerInitiatedOnly(commandInfo.Name)) // <-- Name can be null
        ...
    }
}
catch (JsonException)   // <-- only JsonException is caught
{
    _logger.LogWarning("Failed to deserialize socket command: {Message}", message);
}
```

`System.Text.Json` happily deserializes `{"id":1,"parameters":"..."}` into a non-null `SocketCommandInfo` whose `Name` is `null` (the parameterized ctor binds `null` for the omitted member). The very next line calls `IsServerInitiatedOnly(commandInfo.Name)`:

```csharp
// Game.Api/Services/SocketCommandFactory.cs:17-20
public bool IsServerInitiatedOnly(string commandName)
{
    return _serverInitiatedCommandNames.ContainsKey(commandName); // ConcurrentDictionary.ContainsKey(null) throws ArgumentNullException
}
```

`ConcurrentDictionary.ContainsKey(null)` throws `ArgumentNullException`. The surrounding `catch` only handles `JsonException`, so the exception escapes into `ReadLoop`'s generic catch — logged, the connection survives, but **the client gets no response for that command id** and hangs until its own per-request 30s backstop fires.

## Suggested fix

Guard for a missing/empty `Name` (and `Id`) and return the project's structured error response, consistent with the server-initiated rejection path immediately below it — e.g.:

```csharp
if (string.IsNullOrEmpty(commandInfo.Name))
{
    await _context.SendData(new ApiSocketResponse { Id = commandInfo.Id, Error = "Malformed command." });
    return;
}
```

(or make `IsServerInitiatedOnly` null-tolerant). A malformed-but-parseable frame should be rejected with an error response, not produce a swallowed server exception + client hang.

## Test gap

There are tests for server-initiated rejection and JSON deserialization (`SocketSerializationTests`, `ServerInitiatedCommandRejectionTests`) but none feeds a structurally-valid frame missing `name`. Add a unit test pinning the error-response behavior once fixed.

ginderjeremiah:
## Summary

Closes #935.

A structurally-valid JSON socket frame that omits `"name"` (e.g. `{"id":"1","parameters":"..."}`) deserialized into a non-null `SocketCommandInfo` whose `Name` was `null`. The very next line — `IsServerInitiatedOnly(commandInfo.Name)` → `ConcurrentDictionary.ContainsKey(null)` — threw an `ArgumentNullException`. Because `HandleMessage`'s `catch` only handled `JsonException`, the exception escaped into the read loop's generic catch: the connection survived but **the client got no response for that command id and hung until its own 30s backstop fired**.

## How it addresses the issue

`HandleMessage` now guards for a missing/empty `Name` *before* the command lookup and returns the project's structured error response (`"Malformed command."`), mirroring the server-initiated rejection path immediately below it. A malformed-but-parseable frame is now rejected with an error the client can react to, instead of producing a swallowed server exception + client hang.

- `Game.Api/Sockets/SocketHandler.cs` — added the empty-`Name` guard; logs a warning and sends `ApiSocketResponse { Id, Error = "Malformed command." }`. `HandleMessage` changed from `private` to `internal` for direct unit-test access (consistent with the already-`internal` `ExecuteServerCommand`, which is exercised the same way under the existing `InternalsVisibleTo("Game.Api.Tests")`).

## Test plan

- Added `HandleMessage_FrameMissingName_RejectsWithMalformedErrorInsteadOfThrowing` to `SocketCommandExecutionTests` — feeds a structurally-valid frame missing `name`, asserts the client receives a `"Malformed command."` error (echoing the id) and a warning is logged, and that no exception escapes. This fills the test gap the issue calls out.
- `dotnet build` clean (0 warnings).
- Full `Game.Api.Tests` suite green (525 tests).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GTLDxf9dT7fHv1sBZp7Bk3)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
@@ -37,6 +37,20 @@ namespace Game.Api.Tests.Unit
         }
 
         [Fact]
+        public async Task HandleMessage_FrameMissingName_RejectsWithMalformedErrorInsteadOfThrowing()
+        {
+            // A structurally-valid frame that omits "name" deserializes to a SocketCommandInfo with a null
+            // Name. It must be rejected with a structured error the client can react to, not throw an
+            // unobserved exception that leaves the client hanging on its request id (#935).
+            var (socket, handler) = CreateHandler(_ => null);
+
+            await handler.HandleMessage("{\"id\":\"c1\",\"parameters\":null}");
+
+            Assert.Contains(socket.SentMessages, m => m.Contains("Malformed command.") && m.Contains("c1"));
+            Assert.Contains(_logs.Entries, e => e.Level == LogLevel.Warning && e.Message.Contains("no name"));
+        }
+
+        [Fact]
         public async Task ExecuteCommand_CommandFaults_SurfacesInternalServerErrorToTheClient()
         {
             var (socket, handler) = CreateHandler(name => name == "Boom" ? new InvalidOperationException("boom") : null);
--- a/Game.Api/Sockets/SocketHandler.cs
+++ b/Game.Api/Sockets/SocketHandler.cs
@@ -318,7 +318,7 @@ namespace Game.Api.Sockets
             }
         }
 
-        private async Task HandleMessage(string message)
+        internal async Task HandleMessage(string message)
         {
             if (message is "pong")
             {
@@ -336,6 +336,20 @@ namespace Game.Api.Sockets
                 var commandInfo = message.Deserialize<SocketCommandInfo>();
                 if (commandInfo is not null)
                 {
+                    if (string.IsNullOrEmpty(commandInfo.Name))
+                    {
+                        // A structurally-valid frame can omit "name" (System.Text.Json binds it null), so reject
+                        // it with a structured error instead of letting the null reach the command lookup — which
+                        // would throw an unobserved exception and leave the client hanging on its request id.
+                        _logger.LogWarning("Received socket command frame with no name: {Message} on socket: {Id}", message, Id);
+                        await _context.SendData(new ApiSocketResponse
+                        {
+                            Id = commandInfo.Id,
+                            Error = "Malformed command."
+                        });
+                        return;
+                    }
+
                     if (_commandFactory.IsServerInitiatedOnly(commandInfo.Name))
                     {
                         // Server-initiated commands (e.g. ChallengeCompleted, SocketReplaced) are only valid


ginderjeremiah:
## Review: Approve ✅

Reviewed the full diff against the issue (#935) and the project standards.

### Summary
A structurally-valid socket frame omitting `"name"` deserialized into a non-null `SocketCommandInfo` with a `null` `Name`, which then hit `IsServerInitiatedOnly(commandInfo.Name)` → `ConcurrentDictionary.ContainsKey(null)` → `ArgumentNullException`. Since the surrounding `catch` only handled `JsonException`, the exception escaped to the read loop's generic catch: the connection survived but the client got no response and hung on its request id until its 30s backstop. This PR adds an early `string.IsNullOrEmpty(commandInfo.Name)` guard that returns the structured `"Malformed command."` error, mirroring the server-initiated rejection path directly below it.

### What I checked
- **Guard placement (correctness):** Correct. The guard sits *before* the `IsServerInitiatedOnly` lookup that was the throwing call, so the null can no longer reach the `ContainsKey`. `string.IsNullOrEmpty` covers both the omitted (`null`) and empty-string cases.
- **Mirrors the existing rejection path:** Yes — same `_context.SendData(new ApiSocketResponse { ... })` shape, echoes `Id` so the client can correlate and react. Omitting `Name` on the response is correct here (it's null/empty by definition), unlike the server-initiated path which has a real name to echo.
- **Null handling:** `SocketCommandInfo.Name` is declared non-nullable but System.Text.Json binds `null` into it (the root cause). The fix addresses the real deserialization behavior at the right layer (the API input-validation edge) rather than masking it — no null-forgiving operator used.
- **Logging:** A single `LogWarning` with the raw frame + socket id, consistent in level and tone with the adjacent server-initiated-rejection warning. Appropriate.
- **Visibility change (`private` → `internal`):** Justified and consistent — `ExecuteServerCommand`/`ExecuteCommand` are already `internal` and unit-tested the same way under the existing `InternalsVisibleTo("Game.Api.Tests")`. No broader exposure than the established pattern.
- **Test coverage:** `HandleMessage_FrameMissingName_RejectsWithMalformedErrorInsteadOfThrowing` feeds the exact malformed-but-parseable frame and asserts both the `"Malformed command."` error (echoing the id) and the warning log. Notably the stub factory does *not* override `IsServerInitiatedOnly`, so the base `ContainsKey(null)` would still throw — meaning the test genuinely proves the new guard short-circuits before that call. This closes the test gap the issue called out.
- **DRY / dead code:** No duplication or dead code introduced. Comments are concise and explain the non-obvious "valid JSON, null Name" trap.
- **CI:** Green (`test-backend` success, full suite reported at 525 tests; `all-checks-passed` succeeded).

### Nits / follow-ups
None. Minimal, targeted, well-tested fix that fully resolves the issue.

Approving and merging.

---

PR #959, opened 2026-06-19
Insert missing rows in reorder-sensitive write-behind handlers (#936)

### LINKED ISSUE #936: Write-behind handlers silently drop state when the target row hasn't been inserted yet (loadout / equip)
## Problem

Three of the order-sensitive player write-behind handlers only ever **mutate existing rows** — they never insert — so under cross-instance event reordering (which `backend-persistence.md` documents as best-effort, not guaranteed, and which the reclaim path can re-introduce) they silently no-op against a row set that doesn't yet contain the target.

### 1. `SelectedSkillsChangedHandler` drops a selection when the skill row doesn't exist yet

```csharp
// Game.DataAccess/PlayerUpdates/Handlers/SelectedSkillsChangedHandler.cs:16-40
var playerSkills = await context.PlayerSkills.Where(ps => ps.PlayerId == evt.PlayerId).ToListAsync();
// ... flips Selected/Order only on rows that already exist; never inserts
```

If a `SelectedSkillsChangedEvent` is applied before the `SkillUnlockedEvent` that creates the row, the just-unlocked skill is silently left `Selected = false`. The DB self-heals only when another loadout change later arrives. Within a single instance the per-socket lock makes this unlikely, but the write-behind queue applies on a different thread/instance with no such guarantee — and this is the one handler whose correctness actually contradicts the "reordering is a real hazard" assumption the rest of the design is built around.

### 2 & 3. `ItemEquippedHandler` / `ItemUnequippedHandler` leave the slot empty on a missing `UnlockedItem` row

```csharp
// Game.DataAccess/PlayerUpdates/Handlers/ItemEquippedHandler.cs:17-21
await context.UnlockedItems
    .Where(ui => ui.PlayerId == evt.PlayerId && (ui.ItemId == evt.ItemId || ui.EquipmentSlotId == evt.SlotId))
    .ExecuteUpdateAsync(...);
```

`ExecuteUpdateAsync` over a `Where` updates **zero rows (no error)** if the `UnlockedItem` for `evt.ItemId` isn't present. Equip presupposes unlock, so an `ItemUnlockedEvent` reordered after the equip leaves the slot empty until the next equip. `ItemUnequippedHandler.cs:11-14` has the same shape.

## Suggested direction

Either make these handlers resilient to the row not existing yet (reconcile against the unlocked set / insert the missing row, or assert via FK so the reclaim retries rather than silently swallowing), **or** explicitly document that they rely on unlock-before-equip / unlock-before-select ordering — which currently contradicts the best-effort-ordering statement elsewhere. While here, consider documenting in `backend-persistence.md` that the player and progress aggregates are ordered **independently**, not jointly (they share the queue but are separate enqueues).

## Test gap

`PlayerUpdateHandlerIdempotencyIntegrationTests` covers `ItemUnlocked`, `SkillUnlocked`, `ModUnlocked`, `ModApplied`, `ProgressUpdated`, `AttributeAllocationsChanged`, `LogPreferenceChanged` (sequential + concurrent) — but **`SelectedSkillsChanged`, `ItemEquipped`, and `ItemUnequipped` have no idempotency/convergence coverage at all**, despite being the most reorder-sensitive. Add convergence/concurrent cases plus the missing-row scenario above. Separately, `RedisQueue.ReclaimProcessingAsync`'s order-preservation claim (`RedisQueue.cs:80-96`, tail-first so oldest ends at head) is asserted to recover items but not to preserve order — add a test that reserves several items, simulates a crash, reclaims, and asserts head-to-tail order.

ginderjeremiah:
## Summary

Closes #936.

Three order-sensitive player write-behind handlers only ever **mutated existing rows**, so under best-effort cross-instance event reordering (which the reclaim path can re-introduce) they silently no-op'd against a row set that didn't yet contain the target — dropping a just-equipped skill or leaving a slot empty until a later edit self-healed the DB.

The fix follows the project's own established idempotency pattern (`AttributeAllocationsChangedHandler`'s load-then-upsert + unique-violation **retry-once**): the order-sensitive child handlers now **insert the missing row if the prerequisite unlock is reordered behind them**, rather than mutating only existing rows.

## Changes

| Handler | Change |
|---|---|
| `SelectedSkillsChangedHandler` | Inserts a `Selected` row for any loadout id with no existing `PlayerSkill`, instead of skipping it. Wrapped in the same clear-and-retry-once on a concurrent insert. (Its comment already *claimed* parity with the attribute-allocations handler that it didn't actually have.) |
| `ItemEquippedHandler` | Converted the single-statement `ExecuteUpdate` to a tracked load-then-upsert that inserts the equipped item's row when missing — clearing the prior slot occupant in the same transaction — and retries once on a concurrent insert. |
| `ItemUnequippedHandler` | **No functional change** — documented why none is needed (see below). |

### Why `ItemUnequipped` needs no insert

For unequip, a missing row already **is** the desired end state: "unequipped" is the absence of an equipped row. If the unequip is reordered ahead of the item's `ItemUnlockedEvent`, the missing-row update is a benign no-op and the later unlock inserts the row with a null (unequipped) slot — so the end state still converges. Inserting here would only create a phantom row. I added a clarifying comment and a convergence test rather than a spurious "fix".

### Stale comment corrected

`ItemEquippedHandler`'s old comment justified its single-statement design with *"the queue read is a destructive pop, so the event is not redelivered."* That's now stale — the queue **reserves** (at-least-once) and redelivers on a crash (`backend-persistence.md` → Queue processing resilience). Corrected inline since I was rewriting the handler anyway.

## Tests

Added to `PlayerUpdateHandlerIdempotencyIntegrationTests` (the missing coverage the issue called out — these three handlers previously had none):

- **SelectedSkillsChanged**: missing-row insert (reorder-before-unlock), twice-applied convergence to the ordered loadout, concurrent double-apply → one row.
- **ItemEquipped**: missing-row insert (reorder-before-unlock), twice-applied convergence (clears prior slot occupant), concurrent double-apply → one row.
- **ItemUnequipped**: twice-applied convergence, and missing-row-then-unlock convergence to unequipped.

## Notes / decisions

- **The `RedisQueue.ReclaimProcessingAsync` order-preservation test gap in the issue is already covered** by the existing `ReclaimProcessingAsync_RestoresUnacknowledgedItemsToHead_PreservingOrder` (reserves several, simulates a crash, reclaims, asserts head-to-tail order). No new test needed there.
- `ItemFavoriteChangedHandler` has the same mutate-only shape, but favoriting presupposes ownership and a lost-favorite is cosmetic and self-healing, so I left it out of scope (flagging it here rather than expanding the change).
- Documented in `backend-persistence.md` that the player and progress aggregates are ordered **independently** (separate enqueues sharing the queue) and that order-sensitive child handlers must insert-if-missing.

## Verification

- `dotnet build` (full solution): clean, 0 warnings.
- `dotnet test Game.Application.Tests` (incl. the new cases): **312 passed**.
- `dotnet format --verify-no-changes` on the changed files: clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_016FKBBCbYgTdU533EGNHREG)_


ginderjeremiah:
## Code Review — Verdict: Approve ✅

Reviewed the full diff against `backend.md` / `backend-persistence.md` and the established `AttributeAllocationsChangedHandler` idempotency pattern.

### Summary
The PR closes the silent-drop bug in #936: three reorder-sensitive player write-behind handlers previously mutated only existing rows, so a child-row event (`SelectedSkillsChanged` / `ItemEquipped`) reordered ahead of its prerequisite `SkillUnlockedEvent` / `ItemUnlockedEvent` no-op'd against a row set that didn't yet contain the target. The handlers now insert the missing row, wrapped in the project's load-then-upsert + clear-and-retry-once-on-unique-violation pattern. `ItemUnequipped` is correctly left unchanged with a documented rationale.

### What I verified
- **Pattern adherence:** `SelectedSkillsChangedHandler` and `ItemEquippedHandler` mirror `AttributeAllocationsChangedHandler` exactly — `try { ApplyAsync } catch (DbUpdateException ex) when (ex.IsUniqueViolation()) { ChangeTracker.Clear(); ApplyAsync }`. The `IsUniqueViolation()` helper is the shared extension already used across the get-or-create and write-behind paths.
- **`ItemEquippedHandler` correctness:** the load-then-upsert preserves the old `ExecuteUpdate` CASE semantics. The `else if (row.EquipmentSlotId == evt.SlotId)` branch correctly never treats the equipped item's own row as the slot occupant to clear, so the move-from-another-slot case still vacates the old slot via the single reassigned row. Single `SaveChanges` keeps it atomic.
- **`SelectedSkillsChangedHandler` correctness:** existing skill ids tracked via a `HashSet` sized to the row count; missing loadout ids inserted as `Selected = true` with their index as `Order`. Reset-then-rebuild semantics preserved.
- **`ItemUnequipped` reasoning:** sound — "unequipped" *is* the absence of an equipped row, so a missing-row no-op converges once the later unlock inserts a null-slot row. An insert here would only create a phantom row.
- **Stale-comment correction:** the old `ItemEquippedHandler` comment justified the single-statement design with "the queue read is a destructive pop, so the event is not redelivered" — now stale (the queue reserves, at-least-once). Correcting it inline while rewriting the handler is the right call and aligns with the doc.
- **Doc update:** the new `backend-persistence.md` paragraph (independent ordering of the player vs. progress aggregates; child handlers must insert-if-missing) is accurate and lands in the right section.
- **Tests:** comprehensive and fill exactly the gap the issue called out — missing-row insert, twice-applied convergence, and concurrent double-apply for both equip and select; plus convergence and reorder-converges-to-unequipped for unequip. The reclaim order-preservation gap from the issue is genuinely already covered by `ReclaimProcessingAsync_RestoresUnacknowledgedItemsToHead_PreservingOrder`, so no new test there is correct.
- **Standards:** no null-forgiving operator, mandatory braces honored, no dead code, DRY via the shared private `ApplyAsync`. CI all-checks-passed is green (backend + e2e).

### Follow-up (non-blocking)
The author flagged `ItemFavoriteChangedHandler` as having the same mutate-only shape but out of scope. I agree it's out of scope here, but it does leave one child-row handler as an unstated exception to the rule the new doc paragraph establishes. Filed as #962 (decide insert-if-missing vs. documented exception) so the handler set stays internally consistent — not a blocker for this PR.

No nits or should-fixes on the change itself. Merging.

— Code review

---

PR #963, opened 2026-06-19
Fix frontend optimistic-mutation races: serialize setFavorite, cancellable enemy backoff (#939)

### LINKED ISSUE #939: Frontend optimistic-mutation bugs: setFavorite not serialized; un-cancellable boss-transition backoff blocks controls
## Problem

Two real behavioral issues in the frontend engine surfaced during a codebase health review.

### 1. `setFavorite` is not serialized with the other optimistic inventory mutations

`UI/src/lib/engine/player/inventory-manager.ts:223-235`. `equipItem`/`unequipItem`/`applyMod`/`removeMod` all run through `serialize()` (chained off `lastOperation`) so their optimistic write + rollback baseline can't interleave with another in-flight item mutation. `setFavorite` mutates `item.favorite` and persists **outside** that chain. A favorite toggle that resolves between another op's snapshot and its rollback races the publish/rebuild, and the divergent failure policy (no rollback) is undocumented at the call site. **Fix:** route `setFavorite` through `serialize()` (keeping its no-rollback-on-error behavior inside the serialized closure) so its optimistic write can't interleave with an in-flight item mutation.

### 2. Un-cancellable retry backoff blocks boss-transition controls under flaky network

`UI/src/lib/engine/battle/enemy-manager.ts:100,112-153`. `getNewEnemy` retries with `await delay(result.data?.cooldown || NEW_ENEMY_RETRY_DELAY_MS)`, and `delay` (`UI/src/lib/common/functions.ts:6`) is a bare, **non-cancellable** `setTimeout` promise. `challengeBoss`/`retreatFromBoss` set `transitioning = true` and `await this.getNewEnemy()`. During a sustained server outage `getNewEnemy` sits in the retry loop, so `transitioning` stays `true` for the full backoff window and a player rapidly hitting Challenge/Retreat has their action silently dropped by the `if (this.transitioning) return;` guards. Not a crash — but the boss controls feel unresponsive/broken under flaky network. **Fix:** make the backoff cancellable (an abortable delay keyed off `started` / a transition token) so a stop/retreat short-circuits the wait.

### Minor (fold in if convenient)
- `grantExp` (`player-manager.ts:84-94`) loops `while (this.exp >= nextLevelThreshold)`; a one-line guard (`if (threshold <= 0) break;`) makes it robust against a hypothetical `EXP_PER_LEVEL` constant regression.

ginderjeremiah:
## Summary

Addresses the two behavioral bugs (plus the minor guard) called out in #939, all in the frontend engine.

### 1. `setFavorite` is now serialized with the other optimistic inventory mutations
`InventoryManager.setFavorite` previously mutated `item.favorite` and persisted **outside** the `serialize()` chain that `equipItem`/`unequipItem`/`applyMod`/`removeMod` run through. A favorite toggle resolving between another op's snapshot and its rollback could race the publish/rebuild. It now routes through `serialize()` so its optimistic write can't interleave with an in-flight item mutation — its low-stakes, no-rollback-on-error policy is preserved inside the serialized closure.

### 2. Un-cancellable retry backoff no longer blocks boss-transition controls
`EnemyManager.getNewEnemy`'s retry loop awaited a bare, non-cancellable `delay`. During a sustained outage `transitioning` stayed `true` for the full backoff window, so rapid Challenge/Retreat presses were silently dropped by the `if (this.transitioning) return;` guards, and `stop()` couldn't take effect until the delay elapsed.

The backoff is now cancellable, coordinated by a `fetchGeneration` counter + a `cancelBackoff` handle:
- **`stop()`** cuts the parked backoff short so the loop exits immediately on the cleared `started` flag.
- **A rapid second `challengeBoss`/`retreatFromBoss` press** (still dropped by its guard) interrupts the prior transition's parked backoff so it settles promptly — releasing the guard — instead of holding the controls for the full retry window. The press itself is dropped; the player presses again once freed.
- **A proceeding boss transition** also supersedes any in-flight idle-loop fetch (via the generation bump) so a stale retry can't later overwrite the boss being loaded.

The actual sleep still delegates to `delay`, so explicit cooldowns are honored.

### 3. Minor: `grantExp` level-up loop guard
`PlayerManager.grantExp` now bails the level-up loop on a non-positive level threshold, so a hypothetical `EXP_PER_LEVEL` constant regression can't spin it forever.

## How it addresses the issue
Each of the three points in #939 is implemented as specified. The cancellation design keeps the existing re-entrancy/anti-cheat guarantees (single in-flight fetch, no double-spawn) intact.

## Note on a design choice worth a look
The guard-path interrupt means a rapid second press is still *dropped* (the player presses again once the prior transition frees the guard) rather than the new press *superseding* the in-flight one. Superseding would be a larger behavioral change touching the careful `transitioning`/boss-victory re-entrancy handling, so I kept it out of scope — happy to revisit if you'd prefer the new press to take over.

One compound edge case I left as-is (and didn't want to expand scope for): if an idle fetch is mid-flight (`fetchingEnemy` true) when a *failing* `ChallengeBoss` falls back to `getNewEnemy`, the re-entrancy guard can briefly drop that fallback fetch; it recovers on the next stage change. Let me know if you'd like a follow-up issue for it.

## Test plan
- [x] `npm run lint` — clean (0 errors / 0 warnings)
- [x] `npm run test:unit:ci` — full suite green (2081 passed)
- [x] New unit tests:
  - `setFavorite` is serialized behind a pending equip (optimistic write can't interleave)
  - `stop()` short-circuits an in-flight retry backoff without waiting out the delay
  - a rapid second boss press frees a transition parked in its enemy-fetch backoff

No backend parity mirror is needed — these changes are frontend engine/manager logic, not the `battleStep` per-tick arithmetic that the parity suite guards.

Closes #939

---
_Generated by [Claude Code](https://claude.ai/code/session_01XEKaRQJwQajNDaKfbSP2Qa)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -50,6 +50,14 @@ export class EnemyManager {
 	/** Re-entrancy guard for getNewEnemy: overlapping stage handlers must not both spawn-and-notify an
 	 *  enemy (a double-spawn the backend replay would flag as cheating). */
 	private fetchingEnemy = false;
+	/** Bumped whenever the active enemy-fetch loop is superseded (a stop, or a transition taking over).
+	 *  The running getNewEnemy captures this value and exits once it no longer matches, so a stale loop
+	 *  under a sustained outage can't keep spinning — nor later clobber the new fight with an enemy
+	 *  nobody is waiting for. */
+	private fetchGeneration = 0;
+	/** Resolver that short-circuits an in-flight retry backoff so a stop / superseding transition need
+	 *  not wait out the full delay before the loop re-checks whether it should still be running. */
+	private cancelBackoff?: () => void;
 
 	public start() {
 		if (!this.started) {
@@ -63,6 +71,9 @@ export class EnemyManager {
 		if (this.started) {
 			this.started = false;
 			this.battleStageUnhook?.();
+			// Cut short an in-flight retry backoff so a getNewEnemy parked under a sustained outage exits
+			// at once on the cleared `started` flag rather than after the full backoff window.
+			this.interruptFetch();
 			this.returnToIdle();
 		}
 	}
@@ -76,11 +87,14 @@ export class EnemyManager {
 			return;
 		}
 		this.fetchingEnemy = true;
+		const generation = this.fetchGeneration;
 		try {
 			// Retry iteratively rather than via self-recursion: each attempt returns to a flat stack
-			// (a sustained outage no longer grows the async chain without bound) and `stop()` cancels
-			// the loop by flipping `started`.
-			while (this.started) {
+			// (a sustained outage no longer grows the async chain without bound). The loop ends as soon
+			// as `stop()` flips `started` or a transition supersedes this fetch (bumping the generation);
+			// because the retry backoff is cancellable, either takes effect immediately rather than after
+			// the full delay — so the controls don't feel frozen while a backoff is parked.
+			while (this.started && generation === this.fetchGeneration) {
 				const result = await apiSocket.sendSocketCommand('NewEnemy', {
 					newZoneId: playerManager.currentZone
 				});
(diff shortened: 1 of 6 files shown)


ginderjeremiah:
## Review: approve-with-nits

Reviewed against CLAUDE.md and `docs/frontend.md`. Pulled the diff, read the changed files locally, and ran the affected suites + lint.

### Verdict: **approve-with-nits** — merging.
All three points from #939 are implemented as specified, CI is green, and the nits are follow-ups rather than should-fix blockers.

### What's good
- **`setFavorite` serialization** — correctly wraps the whole body in `serialize()`, preserving the no-rollback-on-error policy *inside* the serialized closure. Signature change to `Promise<boolean>` is clean and the doc comment explains the rationale.
- **Cancellable backoff** — the `backoff` / `interruptFetch` / `fetchGeneration` design is sound. The `settled` idempotency flag and the "only clear the shared handle if it still points at this backoff" check correctly handle the race where the real `delay` timer fires after an early cancel. `stop()` and the rapid-press guard paths are well covered.
- **`grantExp` guard** — matches the issue's suggestion; refactored to a local `threshold` to avoid awkward double getter reads.
- **Tests** — new tests are focused and assert the right invariants (favorite queued behind a pending equip; `stop()` short-circuits a held backoff; a rapid second press frees a parked transition). Ran the four affected files locally: **100 passed**. Lint clean on the changed source files.
- Comment depth is appropriate; CLAUDE.md has no one-line-max rule and the explanatory comments earn their keep here.

### Nits / follow-ups (non-blocking)
1. **Narrow clobber window not covered by the generation check** → #969. The generation check gates only the `while` loop condition, not the enemy application right after `await sendSocketCommand('NewEnemy')`. If an idle fetch is superseded *while awaiting `NewEnemy`* (not parked in backoff) and that command then returns an enemy, the loop sets `currentEnemy` / fires `notifyNewEnemyLoaded` before re-checking generation — the same "stale fetch clobbers the boss" class via a narrower window. Low severity (tiny window; `transitioning`/`fetchingEnemy` guards mitigate).

### Out-of-scope items the PR flagged (tracked)
2. Rapid second press is dropped rather than superseding the in-flight transition → #970 (design decision).
3. Failing `ChallengeBoss` fallback fetch can be briefly dropped if an idle fetch is mid-flight → #971.

No backend parity mirror needed — this is engine/manager logic, not `battleStep` per-tick arithmetic.

Squash-merging now.

---

PR #964, opened 2026-06-19
Harden admin content-authoring write-path validation (#951)

### LINKED ISSUE #951: Admin content-authoring write path: duplicate-key & missing-existence validation gaps
## Problem

Several admin content-authoring persistence helpers/repos are missing the up-front validation that their siblings perform, so a malformed `Set*`/`Save*` payload reaches the DB as a duplicate-key/FK violation (a 500 at commit) or a silent 0-row update instead of a clean rejection. These are anti-cheat/robustness gaps on the Content Authoring write path.

### 1. `ChildCollectionReconciler.Reconcile` double-inserts on duplicate desired keys
`Game.DataAccess/Repositories/Admin/ChildCollectionReconciler.cs:39-60`

`existingKeys` is computed once and never updated; the insert loop iterates the raw `desired` collection (not a deduped set). Two desired entries with the **same key** (e.g. a client sends the same `ZoneId`/`AttributeId` twice to a `Set*` endpoint) both miss `existingKeys` → two `Insert` calls for the same PK → unique violation at commit. This backs every admin `Set*` collection setter.

**Fix:** dedupe `desired` by key before the insert/update phase, or reject a desired set containing duplicate keys up front.

### 2. `AttributeChangeSetProcessor.Apply` guards Edit/Delete but not Add
`Game.DataAccess/Repositories/Admin/AttributeChangeSetProcessor.cs:37-52`

`edit`/`delete` are membership-guarded against `existingKeys`, but `add` calls `entityStore.Insert(...)` unconditionally. An `Add` for an attribute the owner already has inserts a duplicate composite-PK row → unique violation at commit.

**Fix:** guard `add` against `existingKeys` (treat an Add of an already-present key as Update, or reject).

### 3. `AdminItems.SaveModSlots` skips the existence/membership validation every sibling does
`Game.DataAccess/Repositories/Admin/AdminItems.cs:76-93`

Unlike the other `Save*`/`Set*` methods it does not verify the owning `Item` exists or that an edited/deleted slot belongs to the item. An Edit of a non-existent slot id becomes a silent EF 0-row update; an Add with a bad `ItemId` FK-faults at commit.

**Fix:** add the same existence/membership pre-check the sibling methods use.

### 4. `AdminSkills.SetEffects` does an O(changes × effects) membership scan
`Game.DataAccess/Repositories/Admin/AdminSkills.cs` (`SetEffects`, ~:95/:111)

Edit/Delete handlers call `skill.SkillEffects.Any(se => se.Id == effect.Id)` per change — a linear scan per change, where every other guard in this layer precomputes a `HashSet` once.

**Fix:** precompute `skill.SkillEffects.Select(e => e.Id).ToHashSet()` once, or route through the shared `ChangeSetProcessor`.

## Related

The Edit-existence rejection guard is copy-pasted across seven admin repos (`AdminEnemies`/`AdminItems`/`AdminItemMods`/`AdminSkills`/`AdminZones`/`AdminChallenges`), which is exactly why #3 was able to omit it. Folding that guard into `ChangeSetProcessor` (predicate + resource name) so a repo *can't* omit it would prevent this class of bug. `AdminChallenges` also reimplements the zero-based-id bounds check inline instead of reusing `_challenges.ValidateChallengeId(...)` (which `AdminZones` already uses).

These paths are also untested (see the test-coverage-gaps issue).

ginderjeremiah:
## Summary

Closes the four duplicate-key / missing-existence validation gaps called out in #951 on the Content Authoring admin write path, so a malformed `Set*`/`Save*` payload is rejected cleanly (a 400-mapped `AdminSaveResult` or a guarded no-op) instead of reaching the DB as a duplicate-key/FK violation (a 500 at commit) or a silent 0-row update.

### 1. `ChildCollectionReconciler.Reconcile` rejects duplicate desired keys
`existingKeys` was computed once and the insert loop iterated the raw `desired` collection, so two desired entries with the same key both missed `existingKeys` → two `Insert` calls for the same PK (unique violation), and on an existing key two `Update` calls double-track the same entity (which EF also rejects). The reconciler now scans the desired set for duplicate keys up front and returns a graceful business failure (`"The submitted {resource} set contains duplicate entries."`) before staging anything. Its signature changed from `void` to `AdminSaveResult` plus a `resourceName`; the five call sites (enemy attribute distributions / skills / spawns, zone enemies, tag assignments) propagate the result. The tag path passes a `HashSet`, so duplicates are structurally impossible there and the guard never fires.

### 2. `AttributeChangeSetProcessor.Apply` guards the Add path
`add` previously called `Insert` unconditionally, so an Add of an attribute the owner already has duplicate-inserted a composite-PK row. Add is now a membership-guarded **upsert**: insert when the key is absent, update when present. It mutates the `existingKeys` set as it inserts, so a duplicate Add **within the same batch** routes the second occurrence to Update instead of a second Insert. Adds sort last (after Delete/Edit), so this can't disturb the edit/delete membership guards.

### 3. `AdminItems.SaveModSlots` adds existence/membership validation
It now rejects the batch up front (`NotFound("Item")`) if any **Add** targets a non-existent owning item (a bad `ItemId` would otherwise FK-fault at commit), and guards **Edit/Delete** by the slot's membership in its stated owning item — a slot the item doesn't have is reconciled away rather than becoming a silent EF 0-row update/delete. Membership uses a memoized per-item slot-id set so each guard is O(1).

### 4. `AdminSkills.SetEffects` precomputes the membership set
The Edit/Delete guards did an `O(changes × effects)` `.Any(...)` linear scan per change; they now share a single precomputed `HashSet` of the skill's effect ids, matching every other guard in this layer.

## How it addresses the issue
Each of the four numbered points in #951 is implemented as specified, choosing the "upsert / reject up front" options most consistent with the documented admin write-path contract (identity/owner saves reject up front; child edits/deletes of a missing member are guarded no-ops). Docs/XML comments for the affected helpers and `IAdminItems.SaveModSlots` were updated to match.

## Out of scope (follow-up)
The issue's "Related" section suggests folding the copy-pasted Edit-existence guard into `ChangeSetProcessor` (so a repo *can't* omit it, as #3 did) and having `AdminChallenges` reuse `_challenges.ValidateChallengeId(...)` instead of its inline bounds check. That's a cross-repo refactor touching seven repositories, so I've left it for a follow-up issue rather than expanding this PR's scope.

## Test plan
- [x] `dotnet build` — clean (0 warnings / 0 errors)
- [x] Full backend suite green: Core 620, Infrastructure 42, Application 325, Api 525 — 0 failures
- New unit tests: reconciler duplicate-key rejection; attribute Add-of-present upsert and duplicate-Add-in-batch
- New integration tests: `AdminItems` mod-slot add-of-unknown-item rejection, add/edit/delete-against-membership, non-member edit/delete no-ops, and attribute upsert; `AdminSkills` effect add/edit/delete, unknown-effect no-ops, multiplier upsert; enemy & zone duplicate-key rejection

Closes #951

---
_Generated by [Claude Code](https://claude.ai/code/session_01FAsAGk8mfmdkoyTn4tTxuT)_

### DIFF
--- a/Game.Abstractions/DataAccess/Admin/IAdminItems.cs
+++ b/Game.Abstractions/DataAccess/Admin/IAdminItems.cs
@@ -16,7 +16,9 @@ namespace Game.Abstractions.DataAccess.Admin
         /// <summary>Applies a change set to an item's attributes. Fails if the item does not exist.</summary>
         AdminSaveResult SetAttributes(AddEditAttributesData data);
 
-        /// <summary>Applies an Add/Edit/Delete change set to an item's mod slots.</summary>
+        /// <summary>Applies an Add/Edit/Delete change set to item mod slots. Fails (applying nothing) if an
+        /// add targets an item that does not exist; an edit/delete of a slot the named item does not have is
+        /// a guarded no-op.</summary>
         AdminSaveResult SaveModSlots(IReadOnlyList<Change<ItemModSlot>> changes);
 
         /// <summary>Replaces an item's tag associations. Fails if the item does not exist.</summary>
--- a/Game.Application.Tests/DataAccess/AdminEnemiesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminEnemiesIntegrationTests.cs
@@ -174,6 +174,39 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SetAttributeDistributions_DuplicateDesiredKeys_ReturnsFailure()
+        {
+            int enemyId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var enemy = await TestDataSeeder.CreateEnemyAsync(context);
+                enemyId = enemy.Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            // The same attribute named twice in the desired set would otherwise double-insert into a
+            // composite-PK violation at commit; it must reject up front instead.
+            var data = new SetEnemyAttributeDistributions
+            {
+                EnemyId = enemyId,
+                AttributeDistributions =
+                [
+                    new Contracts.AttributeDistribution { AttributeId = EAttribute.Intellect, BaseAmount = 1m, AmountPerLevel = 1m },
+                    new Contracts.AttributeDistribution { AttributeId = EAttribute.Intellect, BaseAmount = 2m, AmountPerLevel = 2m },
+                ],
+            };
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminEnemies>();
+
+            var result = admin.SetAttributeDistributions(data);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal("The submitted attribute distribution set contains duplicate entries.", result.ErrorMessage);
+        }
+
+        [Fact]
         public void SetAttributeDistributions_UnknownEnemy_ReturnsNotFound()
         {
             using var scope = CreateScope();
--- a/Game.Application.Tests/DataAccess/AdminItemsIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminItemsIntegrationTests.cs
@@ -1,0 +1,203 @@
+using Game.Abstractions.Contracts.Admin;
+using Game.Abstractions.DataAccess.Admin;
+using Game.Core;
+using Game.Infrastructure.Database;
+using Game.TestInfrastructure.Base;
+using Game.TestInfrastructure.Fixtures;
+using Game.TestInfrastructure.Helpers;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+using Contracts = Game.Abstractions.Contracts;
+using Entities = Game.Infrastructure.Entities;
+
+namespace Game.Application.Tests.DataAccess
+{
+    /// <summary>
+    /// Exercises <see cref="IAdminItems"/> write paths that guard malformed change sets before they reach the
+    /// database as a duplicate-key/FK violation or a silent 0-row update: the mod-slot existence/membership
+    /// validation and the attribute change-set upsert. Seeding, the admin write, and the assertion each use a
+    /// separate DI scope so the write runs against an empty change tracker, as a real admin call does.
+    /// </summary>
+    [Collection("Integration")]
+    public class AdminItemsIntegrationTests : ApplicationIntegrationTestBase
+    {
+        public AdminItemsIntegrationTests(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper)
+            : base(containers, testOutputHelper) { }
+
+        [Fact]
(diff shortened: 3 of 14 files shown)


ginderjeremiah:
## Review: approve

Reviewed against `CLAUDE.md` and `docs/backend.md` (+ the admin write-path contract). All four gaps from #951 are addressed exactly as specified, and the correctness of the upsert / reject-up-front logic checks out.

### Correctness
- **#1 `ChildCollectionReconciler`** — duplicate desired keys are scanned and rejected up front (`AdminSaveResult.Failure`) before anything is staged, so neither the double-`Insert` (unique violation) nor the double-`Update` (EF double-tracking) can reach commit. The `void`→`AdminSaveResult` + `resourceName` signature change is propagated cleanly through all five call sites. The tag path passes a `HashSet`, so the guard is structurally unreachable there — correctly documented as such.
- **#2 `AttributeChangeSetProcessor`** — Add is now a membership-guarded upsert via `existingKeys.Add(...)` (insert on first-seen, update otherwise). Verified safe: `EChangeType` orders Delete=2 → Edit=1 → Add=0 and `ChangeSetProcessor` applies `OrderByDescending`, so Adds run last and mutating `existingKeys` cannot disturb the Edit/Delete guards. The duplicate-Add-in-batch case routes the second occurrence to Update. `EntityStore.Update` sets `EntityState.Modified` on a fresh key-only stub regardless of key value, so the upsert emits a correct single-row UPDATE of the existing composite-PK row rather than a duplicate insert.
- **#3 `AdminItems.SaveModSlots`** — rejects the whole batch (`NotFound("Item")`) if any Add targets an unknown owning item (avoids the opaque FK-fault 500), and guards Edit/Delete by per-item memoized slot-id sets (O(1)), turning a non-member edit/delete into a guarded no-op instead of a silent EF 0-row update. Consistent with the documented contract (identity/owner saves reject up front; child edits/deletes of a missing member are guarded no-ops).
- **#4 `AdminSkills.SetEffects`** — replaces the `O(changes × effects)` `.Any(...)` scan with a single precomputed `HashSet`, matching every other guard in this layer.

### Tests
Strong coverage: reconciler duplicate-key rejection (no handlers invoked); attribute Add-of-present upsert + duplicate-Add-in-batch; `AdminItems` add-of-unknown-item rejection, add/edit/delete against membership, non-member edit/delete no-ops, attribute upsert; `AdminSkills` effect add/edit/delete, unknown-effect no-ops, multiplier upsert; enemy & zone duplicate-key rejection. Integration tests correctly use separate DI scopes for seed/write/assert so the write runs against an empty change tracker as a real admin call does.

### Standards / docs
- XML docs on `AttributeChangeSetProcessor`, `ChildCollectionReconciler`, and `IAdminItems.SaveModSlots` updated to match the new behavior.
- `if` blocks use braces throughout; no null-forgiving operator introduced; comments are appropriately concise.

### Out-of-scope follow-up
The cross-repo refactor this PR defers (folding the Edit-existence guard into `ChangeSetProcessor` so a repo can't omit it, and routing `AdminChallenges` through `ValidateChallengeId`) is already tracked in **#965** (`tech debt`, `claude`, `scope: medium`). No new issue needed.

### CI
Green — `test-backend`, all three `test-e2e` browsers, and the `all-checks-passed` gate all succeeded.

No nits or should-fix items. Merging via squash.

---

Issue #966, opened 2026-06-19
ArchiveUser/BanUser have no last-admin guard (concurrent mutual archive/ban can lock out all admins)


ginderjeremiah:
## Context

Follow-up from #952, which asked that the sibling `ArchiveUser`/`BanUser` last-admin guards "be reviewed together." On review they do **not** share `SetUserRoles`'s shape: `Users.SetUserTimestamp` (`Game.DataAccess/Repositories/Users.cs`) — the shared implementation behind both `ArchiveUser` and `BanUser` — only guards `AdminLockoutPolicy.IsSelfTarget`. It has **no last-admin/admin-count guard at all**.

## Problem

Because the actor can't target themselves and remains an active admin, a *single* archive/ban can never lock everyone out. But two concurrent actions can:

- Admins A and B are the only admins. A archives (or bans) B **while** B archives (or bans) A. Neither action checks the admin count, and there's no serialization, so both commit — leaving **zero active admins**. (`otherAdminsRemain` in `SetUserRoles` counts only non-archived accounts, so archiving an admin removes them from the usable-admin pool, exactly like a role removal does.)

This is the same lockout class #952 fixed for role removal, but it needs its own guard since `SetUserTimestamp` has none today.

## Suggested fix

- Decide the policy: block an archive/ban that would leave no active admin holding the Admin role (a new `AdminLockoutPolicy` rule, kept pure and unit-tested, mirroring `CheckRoleChange`).
- Apply it atomically with the same row-locked, in-tier commit pattern #952 introduced for `SetUserRoles` (lock the Admin role row `FOR UPDATE`, re-read the surviving-admin fact under the lock, commit in-tier) so the check-then-act can't be split by a concurrent action.
- Add a concurrency integration test firing two simultaneous mutual archives/bans and asserting at least one is rejected and at least one active admin survives.

## Severity

Low-frequency (requires two concurrent admin actions on each other) but a genuine correctness/lockout bug with the same security implication as #952 — a fully admin-less instance can't recover through the tooling.

PR #967, opened 2026-06-19
Fix SetUserRoles last-admin check-then-act race (#952)

### LINKED ISSUE #952: SetUserRoles last-admin protection has a check-then-act race (can demote the last admin)
## Problem

`Game.DataAccess/Repositories/Users.cs:132-179` (`SetUserRoles`) protects against demoting the final admin, but the check and the act are not atomic:

1. `otherAdminsRemain` is computed by a separate `_context.Users.AnyAsync(...)` query (`:161`).
2. `AdminLockoutPolicy.CheckRoleChange(...)` is evaluated against that snapshot (`:164`).
3. The role set is then mutated, and the actual persistence happens **later**, out of tier, via the MVC `CommitFilter`.

Two concurrent admin edits each demoting a *different* admin can both observe `otherAdminsRemain == true` and both commit, leaving the system with **zero admins** — exactly the state `AdminLockoutPolicy` exists to prevent. There is no transaction or row-lock spanning the check-then-act.

The sibling `ArchiveUser`/`BanUser` last-admin guards likely share the same shape and should be reviewed together.

## Suggested fix

- Run the policy check + mutation under a serializable transaction (or a `SELECT … FOR UPDATE` on the admin role rows) so the admin-count invariant holds across concurrent role changes.
- Add a concurrency integration test that fires two simultaneous demotions of distinct admins and asserts at least one is rejected with `LastAdmin`.

## Severity

Low-frequency (requires two concurrent admin actions) but a genuine correctness/lockout bug with security implications — a fully admin-less instance can't recover through the tooling.

ginderjeremiah:
## Summary

Fixes the check-then-act race in `Users.SetUserRoles` (#952). The last-admin lockout guard computed `otherAdminsRemain` in a separate query, evaluated `AdminLockoutPolicy` against that snapshot, and then deferred the actual persistence to the request's unit of work (the MVC `CommitFilter`). Two concurrent admin edits each demoting a *different* admin could both observe `otherAdminsRemain == true` and both commit, leaving the system with **zero admins** — exactly the state the policy exists to prevent.

## How it addresses the issue

Admin-role removal now applies as an **atomic check-then-act**:

1. Locks the Admin role row (`SELECT … FOR UPDATE`), serializing every admin-membership change through a single point — the lock is held until commit.
2. Re-reads the surviving-admin fact **under that lock**, so a demotion that committed first is reflected.
3. Hands those facts to `AdminLockoutPolicy.CheckRoleChange` (the decision stays a pure, unit-tested rule in `Game.Core` — the data tier only owns the locking/atomicity), and
4. Commits the mutation **in-tier** rather than deferring to the unit of work — the same exception [account creation already makes](https://github.com/ginderjeremiah/GameServer/blob/main/docs/backend.md) for its unique-index race — so the check and the act can't be split apart.

Two concurrent demotions now serialize: the first commits, the second re-reads under the lock, sees only its own target left as admin, and is rejected with `LastAdmin`. Non-removal changes (grants, edits to non-admins) keep the existing deferred unit-of-work path, since they carry no lockout hazard and `CheckRoleChange` returns `Allowed` for them regardless of the count.

## Test plan

- [x] `dotnet build` — clean (0 warnings / 0 errors)
- [x] Full backend suite green: Api 526, Core 620, Application 312, Infrastructure 42 — 0 failures
- New integration test `SetUserRoles_ConcurrentDemotionsOfDistinctAdmins_LeaveOneAdminStanding`: fires two simultaneous demotions of distinct admins (each on its own scope/connection) and asserts exactly one returns `Success`, one returns `LastAdmin`, and the DB is left with exactly one admin. Under the old deferred-commit behavior both demotions return `Success`, so the test pins the regression.
- Existing `AdminLockoutPolicy` unit tests and the other `SetUserRoles` cases are unchanged and still pass.

## Note on the siblings (reviewed, as the issue asked)

The issue suggested reviewing `ArchiveUser`/`BanUser` together. They don't share this shape — their implementation (`SetUserTimestamp`) has **no last-admin guard at all**, only a self-target check. A *single* archive/ban can't lock everyone out (the actor can't target themselves and stays an active admin), but two concurrent *mutual* archives/bans can leave zero active admins. Since closing that needs a new policy rule and is a distinct design decision, I've filed it as **#966** rather than expanding this PR's scope.

Closes #952

---
_Generated by [Claude Code](https://claude.ai/code/session_01GMekNaXZiceFNFjjbs3yDh)_


ginderjeremiah:
## Review: approve ✅

Reviewed against CLAUDE.md and `docs/backend.md` / `docs/backend-admin.md`. This is a clean, well-scoped fix for the check-then-act race in #952.

**Correctness**
- `SELECT 1 FROM "Roles" WHERE "Id" = {adminRoleId} FOR UPDATE` locks the single Admin role row inside the transaction, so every admin-removal path serializes through one point. The first demotion holds the lock to commit; the second blocks, then re-reads `otherAdminsRemain` under the lock and correctly sees the committed change → `LastAdmin`. Standard, correct serialization pattern.
- The rejection paths (`SelfAdminRemoval`/`LastAdmin`) return before any entity is mutated (`ApplyRequestedRoles` runs only after the switch), and `await using var transaction` rolls back the uncommitted transaction on dispose — clean, no dirty tracker left behind.
- No ambient-transaction collision: `CommitFilter` only flushes at request end and never opens a transaction, so `BeginTransactionAsync` is safe. After the in-tier `SaveChangesAsync`+commit, the trailing commit-filter save is a harmless no-op over an emptied tracker. The `CreateAccount` in-tier-commit precedent is accurately cited.

**DDD / isolation** — The decision stays a pure, unit-tested rule in `Game.Core` (`AdminLockoutPolicy`, unchanged); the data tier owns only the locking/atomicity. The `RoleEntity` alias keeps the entity seam internal to the data tier. Matches the project's onion-architecture guidance.

**Tests** — The new `SetUserRoles_ConcurrentDemotionsOfDistinctAdmins_LeaveOneAdminStanding` races two distinct demotions on separate scopes/connections and asserts exactly one `Success`, one `LastAdmin`, and exactly one surviving admin. Correctly an integration test (out-of-process DB locking isn't unit-testable per the testing guidelines), and it genuinely pins the regression.

**Docs** — `backend-admin.md` updated concisely and accurately.

**CI** — Green: `all-checks-passed`, `test-backend`, and all three e2e browsers passed; frontend skipped (no FE changes).

**Siblings / follow-up** — Correctly determined `ArchiveUser`/`BanUser` (`SetUserTimestamp`) don't share this shape (no last-admin guard at all) and filed the distinct concurrent-mutual-archive gap as #966 rather than expanding scope. Good call. #966 already carries the right labels (bug, claude, scope: medium).

**Nits (non-blocking)**
- The new EF calls (`BeginTransactionAsync`/`ExecuteSqlInterpolatedAsync`/`SaveChangesAsync`/`CommitAsync`) pass no `CancellationToken`, but `IUsers.SetUserRoles` carries none, so this is consistent with the existing contract — not a regression.

No should-fix items. Merging with squash.

---

Issue #976, opened 2026-06-19
Optional per-effect stacking cap (balance lever + unbounded-growth safeguard)


ginderjeremiah:
## Context

Follow-up to #740, which made skill effects **stack** (each application adds its own timed modifier; magnitudes sum and each expires independently) instead of refreshing. Per the decision on #740, stacking is currently **unbounded**.

## Motivation

Two reasons a per-effect cap may eventually be wanted:

1. **Balance.** A fast-cooldown skill that re-applies a long-duration effect can snowball without bound (e.g. a poison or self-buff ramping every fire). A cap (`MaxStacks`) would let content authors tune how strong stacking gets.
2. **Performance safeguard.** Each application adds a real `AttributeModifier` and invalidates the attribute cache. For *finite-duration* effects the concurrently-active count is bounded by `DurationMs / CooldownMs`, so realistic authored content stays cheap. But a **permanent-duration** effect on a fast skill would grow the modifier list every tick (O(n²) over a long battle). This is only pathological for authored content that pairs a battle-length duration with a short cooldown — not a concern today — but a cap would make it structurally impossible.

## Proposed approach (deferred design)

Add an authored, per-effect `MaxStacks` field to the `SkillEffect` reference data:

- Entity + EF migration; read contract + TS codegen; domain type; admin Workbench skill-editor field.
- Battle runtime (both `Game.Core/Battle` and `UI/src/lib/battle`, kept in parity): when applying an effect already at its cap, drop/replace the oldest application instead of adding a new one (exact policy — refresh-oldest vs. ignore — to be decided).
- Mirrored parity vectors for the capped behaviour.

This is effectively its own feature (the "Stacking policies" deferred extension in `docs/spikes/180-skill-effects.md`), so it should only be picked up if unbounded stacking proves too strong in balancing.

<sub>Originated while implementing #740.</sub>

Issue #180, opened 2026-06-08
Spike: Plan out skill effects


ginderjeremiah:
Add effects that can be triggered by skills. Will mostly come down to buffs/debuffs (timed and/or after x hits?) that increase or decrease attributes.

May need to be designed first. There will probably be some sort of indicator in the UI for skills that have effects.

PR #977, opened 2026-06-19
Stack skill effects instead of refreshing (#740)

### LINKED ISSUE #740: Add the ability to stack skill effects
If a skill effect has already been applied, allow it to be applied again with the summed effect for the duration of the new application.

For example,

1. I apply 10 DamageTakenPerSecond (DoT) to an enemy for 5s.
2. 2 seconds later, I apply 20 DoT for 2 seconds -> Enemy now has 30 DoT for 2 seconds.
3. 1 second later, I apply 10 DoT again for 5 seconds (from the same skill) -> Enemy now has 40 DoT for 5 seconds.

The UI will need to be updated to show all the sources of the skill effect and the individual amounts applied, along with the total. It should also show a number in the top right of the effect icon, which corresponds to the number of active applications.

ginderjeremiah:
## Summary

Implements #740: re-applying an already-active skill effect now **stacks** instead of refreshing. Each application adds its own timed attribute modifier — magnitudes sum (additive amounts add, multiplicative factors compound) and each application expires on its own schedule. This reverses the original *"refresh, not stack"* rule from spike #180 (its deferred "stacking policies" extension).

Because battle logic runs on **both** the frontend and backend (the anti-cheat replay depends on them agreeing tick-for-tick), the change is implemented identically in `Game.Core/Battle` and `UI/src/lib/battle`, with the parity matrices kept row-for-row aligned.

## How it addresses the issue

> If a skill effect has already been applied, allow it to be applied again with the summed effect for the duration of the new application.

- **Runtime:** `Battler.ApplyEffect` / `applyEffect` drop the refresh branch and always add a new modifier + active-effect entry. The frontend tags each application with a unique id so its modifier is matched on expiry (applications stack, so an effect id no longer maps to a single modifier). Independent per-application expiry already falls out of the existing bookkeeping, so a 10 dmg/s poison re-applied over a 5s one reads as 15 dmg/s until the first expires, then 10 — exactly the issue's example.
- **UI (as the issue calls out):** active-effect chips group every application of one authored effect into a single tile showing the **combined total** magnitude, a **count badge in the top-right** (number of active applications), and a sweep tracking the longest-remaining application. The chip tooltip lists **each application's individual amount and remaining time alongside the total**.
- Every application is now logged (there is no refresh to suppress).

## Test plan

- **Backend** (`Game.Core.Tests`, 621 tests): green. `dotnet build` of the full solution: 0 errors.
- **Frontend** (`UI`, 2116 unit tests): green. `npm run lint` (eslint + svelte-check): clean.
- Mirrored parity matrices updated identically on both sides:
  - The self-buff scenario now stacks (ends at 2000ms vs the old 2800ms); a new `dotStacksEachApplication` scenario pins same-effect DoT stacking (victory at 400ms vs 2000ms under refresh).
  - DoT/HoT *phase* scenarios that needed a constant effect now source it from a base attribute (same expected values, no stacking confound); mid-battle *buff* scenarios (CDR live-read, MaxHealth clamp, same-tick slot ordering) are recomputed under stacking with updated derivations.
  - Unit tests (`battler-effects`, `battle-step`, `BattlerEffectsTests`, `BattleDamageOverTimeTests`) flipped from refresh to stacking; new `combineEffectAmount` helper + chip/tooltip stacking tests added.
- The attribute-cache performance test's premise ("persistent re-application stays warm via refresh") no longer holds under stacking — updated to assert persistent effects ramp/stack while churning effects toggle; the timing gates still pass.

## Docs

- `docs/game-design.md` (Skill Effects rule + log note) and `docs/backend-battle.md` updated to describe stacking; `docs/spikes/180-skill-effects.md`'s deferred "stacking policies" note points here.

## Follow-up

Stacking is **unbounded** (per the decision on #740). An optional per-effect `MaxStacks` cap — as a balance lever and a safeguard against unbounded modifier growth for permanent-duration effects on fast skills — is tracked in #976.

Closes #740

https://claude.ai/code/session_01HG6ecMjoa46SUNktQErvnZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01HG6ecMjoa46SUNktQErvnZ)_


ginderjeremiah (commented):
## Review: Stack skill effects instead of refreshing (#740)

**Verdict:** Approve — implements the issue's stacking semantics correctly and identically on both battle engines, with mirrored test coverage and updated docs. (Submitted as a `COMMENT` because I'm the PR author; GitHub won't accept a gating verdict on one's own PR.)

### Correctness
- `Battler.ApplyEffect` / `applyEffect` drop the refresh branch and always add a fresh `AttributeModifier` + active-effect entry. Expiry matches by **modifier identity** (backend) / per-application **`applicationId`** (frontend), so stacked applications expire independently — exactly the issue's "15 dmg/s until the first expires, then 10" example. Verified the backend expiry pass (`Battler.cs:153`) removes by `Modifier` reference and the frontend keys `#effectModifiers` by `applicationId`.
- Parity holds: both sides append a new modifier at the end of its type-run on apply and remove by reference identity on expiry, so modifier ordering — the float-parity-critical surface called out in `docs/backend-battle.md` — stays consistent. The `selfStrengthBuffStacksEachFire` and `dotStacksEachApplication` scenarios exist row-for-row in both `BattleSimulatorParityTests.cs` and `battle-simulation-parity.test.ts`.
- `combineEffectAmount` (additive sum, multiplicative compound via `amount ** count`) mirrors how the real modifiers aggregate, so the chip total and tooltip headline match the battle math.

### UI (per the issue)
- Chips now group all applications of one authored effect into a single tile with a combined-total magnitude, a top-right count badge (shown only when >1), a sweep tracking the longest-remaining application, and a tooltip breakdown of each application's amount + remaining time. Grouping by `sourceId` is sound because same-effect stacks share the authored amount, so the per-row amounts are intentionally identical and only the remaining times differ.

### Test coverage
- Backend and frontend unit suites flipped from refresh→stacking in lockstep (`BattlerEffectsTests` / `battler-effects.test.ts`), including independent-expiry and "every application is logged/reported" cases. New `combineEffectAmount`, chip-grouping/count-badge, and tooltip-breakdown tests added. The performance test's refresh premise was correctly retargeted to stacking. Coverage meets the project's mirrored-battle-logic rule.

### Docs
- `game-design.md`, `backend-battle.md`, and the spike's deferred "stacking policies" note all updated to describe unbounded stacking. The unbounded-growth/balance follow-up is tracked in #976.

Clean, well-scoped change with no blocking findings, no should-fixes, and no nits worth holding it for. Merging once CI is green.

---

Issue #983, opened 2026-06-19
Ban is not enforced at login (GetUser filters only ArchivedAt, not BannedAt)


ginderjeremiah:
## Context

Surfaced while implementing #966 (last-admin guard for `ArchiveUser`/`BanUser`).

## Problem

A banned user can still log in. `Users.GetUser` (`Game.DataAccess/Repositories/Users.cs`) — the credential lookup behind the login flow — filters only on `ArchivedAt == null`:

```csharp
return _context.Users
    .Where(u => u.Username == username && u.ArchivedAt == null)
    .Select(ToCredentials)
    .FirstOrDefaultAsync();
```

There is **no `BannedAt` check anywhere** in the auth path (`AccountService`, `LoginController`, `GetUser`). Grepping `Banned`/`BannedAt` across `Game.Application`/`Game.Api`/`Game.Core` finds only the `AdminUser` contract field. The `User` entity even documents ban as "blocked," but nothing enforces it:

```
/// When set, the user has been banned. Unlike archiving, a banned user still occupies their
/// username (it is not freed for reuse) and remains visible in the admin roster.
```

So today banning a user reserves their username and flags them in the roster, but does **not** actually block them from authenticating — the ban has no teeth.

## Suggested fix

- Exclude banned users from the login credential lookup (or, better, fetch them and return a distinct "account is banned" result so the client can show an accurate message rather than a generic "invalid credentials"). Decide whether a banned user should be told they are banned or treated as non-existent.
- Add an integration test asserting a banned user cannot establish a session.

## Related: unify the last-admin "usable admin" predicate

#966 made the archive/ban last-admin guard treat a *usable* admin as one that is **neither archived nor banned** (so banning the last clean admin is rejected). `SetUserRoles`'s `otherAdminsRemain` check (from #952) still counts banned admins as remaining (`ArchivedAt == null` only). Once ban-at-login is enforced here, that guard should be tightened to the same "non-archived **and** non-banned" definition so a role removal can't strand the system with only banned admins. Ideally both guards share one predicate.

## Severity

Genuine correctness/security defect: an admin action intended to block access does nothing of the kind. Low blast radius today (ban is admin-only), but the feature is effectively non-functional.

PR #984, opened 2026-06-19
Add last-admin guard to ArchiveUser/BanUser (#966)

### LINKED ISSUE #966: ArchiveUser/BanUser have no last-admin guard (concurrent mutual archive/ban can lock out all admins)

ginderjeremiah:
## Summary

`ArchiveUser`/`BanUser` (both backed by `Users.SetUserTimestamp`) only guarded self-targeting — they had **no admin-count guard at all**. Because the actor can't target themselves, a *single* archive/ban can never empty the admin pool, but two **concurrent** actions can: admins A and B are the only admins, A archives/bans B *while* B archives/bans A, neither checks the count, both commit → **zero usable admins** and nobody can recover through the tooling.

This is the same lockout class #952 fixed for role removal, but `SetUserTimestamp` had none of that protection. This PR adds it, reusing the exact pattern #952 introduced.

Closes #966.

## How it addresses the issue

- **New pure policy rule.** `AdminLockoutPolicy.CheckUserAction(...)` mirrors `CheckRoleChange`: it takes plain facts (self-target, target-has-admin, whether another usable admin remains) and returns a new `UserActionProtection` enum. Kept free of data access and unit-tested.
- **Atomic check-then-act in the data tier.** An admin-targeted archive/ban now runs through `ApplyLifecycleActionToAdminAtomically`: begin a transaction, `SELECT … FOR UPDATE` the Admin role row (the same single serialization point `RemoveAdminRoleAtomically` locks), re-read the surviving-admin fact under that lock, let the policy decide, and commit **in-tier** (like `CreateAccount`) so the check and the act it gates can't be split by a concurrent action. Archiving/banning a non-admin keeps the cheap deferred-commit fast path.
- **Surviving-admin definition.** A *usable* admin must be **neither archived nor banned** — both actions take the target out of the recovery pool — so two mutual bans (or archives, or a mix) can leave at most one admin standing.
- **Controller + contract.** `UserActionStatus.LastAdmin` is mapped to "Cannot archive/ban the last remaining admin."; `IUsers` docs updated.

## Note for the reviewer (deliberate scoping)

To make the **ban** half of the issue's scenario a real lockout to prevent, the new guard treats a banned admin as not-usable. While implementing this I found that **ban is not actually enforced at login** — `Users.GetUser` filters only `ArchivedAt`, so a banned user can still authenticate today. That's a separate defect, filed as **#983**, which also tracks unifying `SetUserRoles`'s `otherAdminsRemain` check (still `ArchivedAt`-only, from #952) to the same "non-archived **and** non-banned" predicate once ban-at-login lands. I left `SetUserRoles` untouched here to keep this PR scoped to #966 and #952's shipped behavior intact.

## Test plan

- **Unit** (`AdminLockoutPolicyTests`): `CheckUserAction` self-target, non-admin target, admin-with-other-admins, and last-admin cases.
- **Integration** (`AdminUsersControllerTests`):
  - Archiving/banning another admin while admins remain → succeeds.
  - Archiving/banning the last admin (ghost actor, via repo) → `LastAdmin`, no mutation.
  - **Concurrent mutual archives** and **concurrent mutual bans** → exactly one `Success` + one `LastAdmin`, exactly one usable admin left in the DB.
- Full backend suite green locally: **1537 passed, 0 failed** (`dotnet build` + `dotnet test`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EpmjZ1C5g4NZaATn5hmEtk)_

### DIFF
--- a/Game.Abstractions/DataAccess/IUsers.cs
+++ b/Game.Abstractions/DataAccess/IUsers.cs
@@ -24,7 +24,8 @@ namespace Game.Abstractions.DataAccess
     /// <summary>
     /// Outcome of a single-user lifecycle action (<see cref="IUsers.ArchiveUser"/> /
     /// <see cref="IUsers.BanUser"/>), so the caller can surface the appropriate message. The data tier
-    /// rejects an admin targeting their own account to prevent a self-inflicted lockout.
+    /// enforces the admin-lockout rules: an admin cannot target their own account, nor take the last
+    /// usable admin out of circulation.
     /// </summary>
     public enum UserActionStatus
     {
@@ -33,6 +34,9 @@ namespace Game.Abstractions.DataAccess
 
         /// <summary>The acting admin attempted the action on their own account.</summary>
         SelfTarget,
+
+        /// <summary>The action would have taken the last usable admin out of circulation.</summary>
+        LastAdmin,
     }
 
     public interface IUsers
@@ -95,15 +99,18 @@ namespace Game.Abstractions.DataAccess
         Task<SetUserRolesStatus> SetUserRoles(int actingUserId, int targetUserId, IReadOnlyCollection<int> roleIds);
 
         /// <summary>
-        /// Archives (soft-deletes) the target user, freeing their username for reuse. Rejects an admin
-        /// archiving their own account (<see cref="UserActionStatus.SelfTarget"/>) and returns
-        /// <see cref="UserActionStatus.UserNotFound"/> if the user does not exist.
+        /// Archives (soft-deletes) the target user, freeing their username for reuse. Enforces the
+        /// admin-lockout rules: rejects an admin archiving their own account
+        /// (<see cref="UserActionStatus.SelfTarget"/>) and archiving the last usable admin
+        /// (<see cref="UserActionStatus.LastAdmin"/>). Returns <see cref="UserActionStatus.UserNotFound"/>
+        /// if the user does not exist.
         /// </summary>
         Task<UserActionStatus> ArchiveUser(int actingUserId, int targetUserId);
 
         /// <summary>
-        /// Bans the target user while keeping their username reserved. Rejects an admin banning their own
-        /// account (<see cref="UserActionStatus.SelfTarget"/>) and returns
+        /// Bans the target user while keeping their username reserved. Enforces the admin-lockout rules:
+        /// rejects an admin banning their own account (<see cref="UserActionStatus.SelfTarget"/>) and
+        /// banning the last usable admin (<see cref="UserActionStatus.LastAdmin"/>). Returns
         /// <see cref="UserActionStatus.UserNotFound"/> if the user does not exist.
         /// </summary>
         Task<UserActionStatus> BanUser(int actingUserId, int targetUserId);
--- a/Game.Api.Tests/Integration/AdminUsersControllerTests.cs
+++ b/Game.Api.Tests/Integration/AdminUsersControllerTests.cs
@@ -559,6 +559,155 @@ namespace Game.Api.Tests.Integration
         }
 
         [Fact]
+        public async Task ArchiveUser_ArchivingAnotherAdmin_WhileAdminsRemain_Succeeds()
+        {
+            var (authClient, _) = await SetupAdminClientWithIdAsync();
+            using var client = authClient;
+            int otherAdminId;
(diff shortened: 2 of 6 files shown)


ginderjeremiah (commented):
## Review: Last-admin guard for ArchiveUser/BanUser (#966)

**Verdict:** Approve (posted as a comment since I'm the PR author) — closes the concurrent-mutual-archive/ban lockout hole by reusing #952's row-locked, in-tier check-then-act pattern, cleanly and with strong test coverage.

This is a faithful, well-scoped extension of the pattern #952 introduced for `SetUserRoles`. The new `AdminLockoutPolicy.CheckUserAction` stays pure and free of data access in `Game.Core`, the atomic work sits in the data tier, and the controller only maps the status — so the DDD layering in `docs/backend.md` is respected. `ApplyLifecycleActionToAdminAtomically` mirrors `RemoveAdminRoleAtomically` line-for-line on the locking/commit shape (lock the Admin role row `FOR UPDATE`, re-read the surviving-admin fact under the lock, commit in-tier), and the cheap non-admin/self-target fast paths keep the common case off the transaction.

### What's good
- Correctly identifies *banned* admins as not-usable in the surviving-admin predicate, which is what makes the ban half of the scenario a real lockout to prevent.
- The deliberate scoping note is excellent: the `SetUserRoles` predicate still being `ArchivedAt`-only (and ban not being enforced at login) is a genuine pre-existing gap, correctly left untouched here and tracked in #983 rather than widened into this PR.
- Surviving-admin definition (`ArchivedAt == null && BannedAt == null && holds Admin`) is consistent between the production query and the `CountUsableAdminsAsync` test assertion.

### Test coverage
- **Unit** — `CheckUserAction` covers self-target (all four role/remain combos), non-admin target, admin-with-others, and last-admin.
- **Integration** — success paths for archiving/banning another admin while admins remain, last-admin rejection with no mutation (driven through the repo with a ghost actor), and the two concurrency tests the issue explicitly asked for (mutual archives / mutual bans → exactly one `Success` + one `LastAdmin`, exactly one usable admin left).

### Nits (non-blocking, no action required)
- `Users.cs:296` — `ApplyLifecycleActionToAdminAtomically` handles `UserActionProtection.SelfTarget`, but `SetUserTimestamp` already rejects self-target up front (`Users.cs:239`), so that switch arm is unreachable on this path. It's a harmless, defensible choice — passing the full facts to the comprehensive pure policy keeps the call honest — and it differs from the sibling only because `RemoveAdminRoleAtomically` doesn't pre-check self. Fine to leave as-is.

No blocking findings. Approving; will let CI finish before merging.

---

PR #985, opened 2026-06-19
Re-check fetch generation after NewEnemy await in EnemyManager (#969)

### LINKED ISSUE #969: EnemyManager: superseding generation check doesn't cover a successful NewEnemy mid-flight (narrow clobber window)
## Context
Follow-up from PR #963 (closes #939), which made `getNewEnemy`'s retry backoff cancellable and added a `fetchGeneration` counter so a stop / superseding transition ends a stale idle fetch loop.

## Observation
The generation check (`while (this.started && generation === this.fetchGeneration)`) only gates the **loop condition** — it runs at the top of each iteration, not after `await apiSocket.sendSocketCommand('NewEnemy', ...)`. The enemy is applied immediately after that await:

```ts
const result = await apiSocket.sendSocketCommand('NewEnemy', { ... });
if (result.data?.enemyInstance) {
    this.currentEnemy = result.data.enemyInstance;
    notifyNewEnemyLoaded(this.currentEnemy);
    return;
}
```

So if `challengeBoss` (or another `interruptFetch` caller) supersedes the idle fetch **while it is awaiting `NewEnemy`** — i.e. parked on the command, not on the cancellable `backoff` — and that command then resolves with an enemy, the loop sets `currentEnemy` and fires `notifyNewEnemyLoaded` *before* re-checking the generation. That is exactly the "a stale retry can't later overwrite the boss being loaded" class the generation bump in `challengeBoss` is meant to close, just via the narrower successful-response window rather than the retry/backoff window the PR fully closed.

## Why it's narrow / low severity
- The window is the exact moment between an `interruptFetch` and the in-flight `NewEnemy` resolving with an enemy.
- The `transitioning` / `fetchingEnemy` re-entrancy guards mitigate the practical impact, and the boss's own `ChallengeBoss` response normally lands first.
- No crash; worst case a momentary wrong enemy that the next stage change corrects.

## Suggested fix
Re-check the generation (and/or `started`) after the `NewEnemy` await, before applying the enemy — e.g. bail if `generation !== this.fetchGeneration || !this.started` rather than only on the loop condition.

## Acceptance
- A superseded idle fetch that receives a successful `NewEnemy` after being superseded does **not** apply the enemy / notify.
- Unit test covering the success-path supersession (analogous to the existing backoff/stop tests in `enemy-manager.test.ts`).

_Found during review of PR #963._

ginderjeremiah:
## Summary

Closes #969.

`EnemyManager.getNewEnemy`'s retry loop gated the active fetch only on the **loop condition** (`while (this.started && generation === this.fetchGeneration)`), which runs at the top of each iteration — not after the `await apiSocket.sendSocketCommand('NewEnemy', …)`. So if a stop or a superseding transition (e.g. `challengeBoss`, which bumps the generation via `interruptFetch`) landed while the loop was parked on the **in-flight `NewEnemy` command** (rather than on the cancellable backoff the PR #963 work already covered), and that command then resolved with an enemy, the loop would set `currentEnemy` and fire `notifyNewEnemyLoaded` *before* re-checking — clobbering the fight the supersession had already moved on to.

This is the narrow success-path window the generation bump in `challengeBoss` was meant to close, just via the successful-response path rather than the retry/backoff path.

## How it addresses the issue

Re-check `this.started` and the captured `generation` immediately after the `NewEnemy` await, before applying the enemy. A superseded success-path response is dropped (the loop returns) instead of spawning-and-notifying. This mirrors the protection the cancellable backoff already gives the retry window.

## Test plan

Added unit tests covering both halves of the post-await guard, each verified to **fail without the fix and pass with it**:

- `enemy-manager.test.ts`
  - stop() landing while parked on the in-flight `NewEnemy` request → the resolved enemy is not applied/notified (`started` re-check).
  - a generation bump (transition supersession) while parked on the request, manager still started → the resolved enemy is not applied/notified (generation re-check).
- `enemy-manager-boss.test.ts`
  - the exact issue scenario end-to-end: an idle fetch parked on its `NewEnemy` is superseded by a real `challengeBoss`, and when the stale idle `NewEnemy` later resolves it does **not** overwrite the loaded boss (`currentEnemy` stays the boss; only the boss was ever notified).

Verification run locally:
- `npx vitest run` on both enemy-manager test files → 37 passed.
- `npm run lint` (eslint + svelte-check) → clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AToj92dJ5WDTDAxLNq6WUk)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -98,6 +98,13 @@ export class EnemyManager {
 				const result = await apiSocket.sendSocketCommand('NewEnemy', {
 					newZoneId: playerManager.currentZone
 				});
+				// The loop condition only gates the top of each iteration, so a stop / superseding
+				// transition that lands while parked on the await above (not on the cancellable backoff)
+				// would otherwise still apply this response. Re-check before applying so a successful
+				// NewEnemy can't clobber the fight the supersession already moved on to.
+				if (!this.started || generation !== this.fetchGeneration) {
+					return;
+				}
 				if (result.data?.enemyInstance) {
 					this.currentEnemy = result.data.enemyInstance;
 					notifyNewEnemyLoaded(this.currentEnemy);
--- a/UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts
+++ b/UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts
@@ -176,6 +176,33 @@ describe('EnemyManager boss mode', () => {
 		expect(logMessage).toHaveBeenCalledWith(ELogType.Debug, 'There was an error challenging the boss: no boss');
 	});
 
+	it('does not let a superseded idle fetch clobber the boss when its NewEnemy resolves mid-challenge', async () => {
+		// The exact issue scenario: an idle getNewEnemy is parked on its in-flight NewEnemy when
+		// challengeBoss supersedes it (bumping the generation) and loads the boss. When the stale idle
+		// NewEnemy then resolves with an enemy, it must not overwrite the boss the supersession moved to.
+		let releaseNewEnemy!: (r: IApiSocketResponse<'NewEnemy'>) => void;
+		const newEnemyGate = new Promise<IApiSocketResponse<'NewEnemy'>>((resolve) => (releaseNewEnemy = resolve));
+		send.mockImplementation((name: string) =>
+			name === 'NewEnemy' ? newEnemyGate : Promise.resolve(challengeResponse)
+		);
+		const loaded: IEnemyInstance[] = [];
+		onNewEnemyLoaded((e) => loaded.push(e), false);
+
+		const idleFetch = manager.getNewEnemy();
+		await flush(); // park the idle fetch on the held NewEnemy request
+
+		await manager.challengeBoss(); // supersedes the idle fetch and loads the boss
+		expect(manager.mode).toBe('boss');
+		expect(manager.currentEnemy).toEqual(bossInstance);
+
+		releaseNewEnemy(newEnemyResponse); // the superseded idle fetch resolves with a normal enemy
+		await idleFetch;
+
+		// The stale idle enemy was dropped: the boss remains, and only it was ever notified.
+		expect(manager.currentEnemy).toEqual(bossInstance);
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: post-await generation/started re-check in `EnemyManager.getNewEnemy` (closes #969)

**Verdict:** Approve _(submitted as a comment since I'm the PR author — GitHub won't accept a self-approval)_ — the change is exactly the fix #969 specifies, minimal, correctly placed, and well-tested.

### Assessment

The guard lands in precisely the right spot — immediately after the `NewEnemy` await and **before** `currentEnemy`/`notifyNewEnemyLoaded` — and re-checks both `started` (the stop path) and `generation` (the transition/supersession path), which is what the issue's acceptance criteria call for. The `return` sits inside the `try`, so the `finally` still clears `fetchingEnemy`; no leak there.

A nice side benefit: bailing on the superseded success path also short-circuits the no-enemy/error branch's backoff for a fetch that's already been moved past, rather than logging + sleeping pointlessly. Behavior is consistent with the cancellable-backoff protection PR #963 added — the two windows (backoff vs. in-flight command) now both honor supersession.

The explanatory comment is a touch long at 4 lines, but this is a subtle concurrency window that a future edit could easily "simplify" away by trusting the loop condition — the context is justified, not over-documentation.

### Test coverage

Thorough and matches the project's testing rules:
- `enemy-manager.test.ts` — both halves of the guard isolated: `stop()` mid-request (the `started` re-check) and `interruptFetch()` mid-request with the manager still started (the generation re-check). The latter correctly asserts `started === true` so it's genuinely exercising the generation branch, not the stop branch.
- `enemy-manager-boss.test.ts` — the end-to-end issue scenario: an idle fetch parked on its held `NewEnemy` is superseded by a real `challengeBoss`, and the late idle resolution leaves the boss intact with only the boss ever notified.

All three use the established helpers/gating patterns (`flush`, manual-release promises, `onNewEnemyLoaded(..., false)`) and assert the right post-conditions. The PR notes each was verified to fail without the fix.

This is frontend-only fight-loop control flow (no battle-sim arithmetic), so there's no backend parity mirror to add.

### Note

CI is still in progress at review time (frontend + e2e jobs running). The change is approved on the merits; I'll hold the merge until the checks come back green.


---

PR #986, opened 2026-06-19
Fix dropped boss-fallback enemy fetch racing an in-flight idle fetch (#971)

### LINKED ISSUE #971: EnemyManager: failing ChallengeBoss fallback fetch can be dropped if an idle fetch is mid-flight
## Context
Follow-up from PR #963 (closes #939). The PR explicitly called out this compound edge case as left as-is to avoid expanding scope.

## Observation
If an idle fetch is mid-flight (`fetchingEnemy === true`) when a **failing** `ChallengeBoss` falls back to `getNewEnemy()` (the `else` branch in `challengeBoss`), the re-entrancy guard (`if (this.fetchingEnemy) return;`) can briefly drop that fallback fetch. The state recovers on the next stage change, so the player isn't permanently stranded, but there's a transient window where the fallback spawn doesn't happen as intended.

## Why it's low severity
- Self-recovers on the next `BattleStage` change.
- Only triggers on a `ChallengeBoss` failure that races a still-in-flight idle fetch.

## Suggested direction
Coordinate the boss-fallback `getNewEnemy` with the in-flight idle fetch (e.g. the boss path already bumps the generation via `interruptFetch`; consider having the fallback await/supersede the in-flight fetch rather than no-op against the re-entrancy guard), so the fallback spawn isn't silently dropped.

## Acceptance
- A failing `ChallengeBoss` reliably produces a fallback enemy even when an idle fetch was mid-flight.
- Unit test covering the fallback-during-in-flight-idle-fetch case.

_Raised during review of PR #963 (compound edge case the PR called out)._

ginderjeremiah:
## Summary

Fixes the compound edge case #963 called out and deferred: when an idle `getNewEnemy` is mid-flight (`fetchingEnemy === true`) and a **failing** `ChallengeBoss` falls back to `getNewEnemy()`, the re-entrancy guard (`if (this.fetchingEnemy) return;`) silently dropped that fallback fetch. The player was left without an enemy until the next `BattleStage` change recovered the state.

### Root cause
`challengeBoss` bumps the fetch generation (`interruptFetch`) before awaiting `ChallengeBoss`, superseding any in-flight idle fetch so it will abandon **without spawning**. But the superseded idle loop only clears `fetchingEnemy` in its `finally`, which can land *after* the failing `ChallengeBoss` resolves and reaches its fallback `await this.getNewEnemy()`. The fallback then hits the still-set re-entrancy guard and no-ops — so nobody spawns.

### Fix
`getNewEnemy` now distinguishes the two cases instead of blanket-dropping when a fetch is in flight:

- **Same generation** → a genuine re-entrant duplicate (two overlapping stage handlers). Still dropped, preserving the anti-cheat double-spawn protection.
- **Newer generation** → the running fetch has been superseded and won't spawn. Await its teardown (keeping a single fetch in flight), then fetch fresh so the fallback spawn is reliably produced.

The fetch loop is split into a public `getNewEnemy` entry and a private `runFetchLoop(generation)` that owns the loop and the `fetchingEnemy`/`fetchingGeneration` bookkeeping. This is the `await/supersede` direction the issue suggested.

## How it addresses the issue
- A failing `ChallengeBoss` now reliably produces a fallback enemy even when an idle fetch was mid-flight — meeting both acceptance criteria.
- The re-entrancy / single-in-flight guarantees the existing tests cover are preserved (same-generation re-entrant calls still drop; a superseded fetch still abandons without clobbering the new fight).

## Test plan
- [x] New unit test: a failing `ChallengeBoss` racing an in-flight idle fetch waits the superseded fetch out and spawns a fresh idle enemy (would drop on the old guard).
- [x] `npx vitest --run enemy-manager` — 38 passed (both boss + idle suites), including the re-entrancy and supersession cases.
- [x] `npm run lint` — clean (0 errors / 0 warnings).

No backend parity mirror needed — this is frontend `EnemyManager` orchestration, not the `battleStep` per-tick arithmetic the parity suite guards.

Closes #971

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGZQZvemLwN6PLiGRX6Joj)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -55,6 +55,14 @@ export class EnemyManager {
 	 *  under a sustained outage can't keep spinning — nor later clobber the new fight with an enemy
 	 *  nobody is waiting for. */
 	private fetchGeneration = 0;
+	/** The generation the in-flight fetch loop captured at its start. Lets a fresh getNewEnemy tell a
+	 *  genuine re-entrant duplicate (same generation — the running loop will still spawn, so drop) apart
+	 *  from a superseded loop (older generation — it will abandon without spawning, so wait it out then
+	 *  fetch fresh rather than be dropped by the re-entrancy guard). */
+	private fetchingGeneration = 0;
+	/** The in-flight fetch loop, so a superseding caller (e.g. a failing boss-challenge falling back to
+	 *  the idle loop) can await its teardown before fetching afresh — keeping a single fetch in flight. */
+	private inFlightFetch?: Promise<void>;
 	/** Resolver that short-circuits an in-flight retry backoff so a stop / superseding transition need
 	 *  not wait out the full delay before the loop re-checks whether it should still be running. */
 	private cancelBackoff?: () => void;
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Fix dropped boss-fallback enemy fetch racing an in-flight idle fetch (#971)

**Verdict:** Approve _(submitted as a comment because GitHub won't let me gate my own PR — there are no blocking or should-fix findings)_ — the fix correctly resolves the #971 compound edge case and preserves the existing anti-cheat / single-in-flight guarantees.

### Walkthrough against the issue

Traced the #971 scenario through the new code:

1. Idle `getNewEnemy()` (gen 0) parks on a held `NewEnemy` request — `fetchingEnemy = true`, `fetchingGeneration = 0`.
2. `challengeBoss()` calls `interruptFetch()` (gen → 1), `ChallengeBoss` fails, falls back to `getNewEnemy()`.
3. Fallback captures `generation = 1`, sees `fetchingEnemy === true` but `fetchingGeneration (0) !== 1`, so it awaits `inFlightFetch` instead of dropping.
4. The superseded idle loop resolves, fails its post-await generation re-check, returns **without spawning**, clears `fetchingEnemy`.
5. Fallback passes the `started && generation === fetchGeneration` re-check and runs a fresh fetch loop → spawns. ✅

Both acceptance criteria are met, and the re-entrancy drop for the genuine same-generation duplicate (anti-cheat) is preserved.

### Correctness notes I verified

- **No `inFlightFetch` stale-read window.** `runFetchLoop` sets `fetchingEnemy = true` synchronously and `getNewEnemy` assigns `inFlightFetch = run` with no intervening `await`, so whenever `fetchingEnemy` is true, `inFlightFetch` points at the live run. The `await this.inFlightFetch` path is only reached under `fetchingEnemy === true`.
- **No double-spawn regression.** A superseded (older-generation) loop always returns before `notifyNewEnemyLoaded` (the synchronous check-then-notify can't be interleaved by `interruptFetch`), so only the fallback spawns.
- **Post-await guards are correct.** Re-checking `!this.started || generation !== this.fetchGeneration` after the `await` covers a stop or a further supersession landing during the wait.
- **Concurrency into `getNewEnemy` is well controlled** — stage handlers bail on `transitioning`, and boss transitions are serialized by the `transitioning` guard, so the only entrants are the documented same-generation re-entrant case and the superseding fallback.

### Test coverage

- New unit test (`enemy-manager-boss.test.ts`) faithfully reproduces #971: it holds the first `NewEnemy`, fails `ChallengeBoss`, releases the idle fetch, and asserts `newEnemyCalls === 2` plus a fresh `normalInstance` spawn — i.e. it would fail under the old blanket-drop guard. Good, targeted regression test.
- No backend parity mirror needed: this is `EnemyManager` orchestration, not `battleStep` per-tick arithmetic. Correctly called out in the PR description and consistent with `frontend.md`'s parity scope.

### What's good

- Clean split into a public `getNewEnemy` entry and a private `runFetchLoop(generation)` that owns the loop bookkeeping — exactly the await/supersede direction the issue suggested, without weakening any existing guard.
- The generation-based distinction is the minimal correct fix and the comments explain the *why* (the two cases) well enough that a future edit is unlikely to naively collapse them back into a blanket drop.


---

PR #987, opened 2026-06-19
Enforce ban at login and unify usable-admin lockout predicate (#983)

### LINKED ISSUE #983: Ban is not enforced at login (GetUser filters only ArchivedAt, not BannedAt)

ginderjeremiah:
## Summary

Closes #983.

A banned user could still log in. `Users.GetUser` — the credential lookup behind login — filtered only on `ArchivedAt == null`, and **no part of the auth path checked `BannedAt`**. So banning a user reserved their username and flagged the roster, but the ban had no teeth: the user could still authenticate.

This PR enforces the ban in the login flow and unifies the related last-admin "usable admin" predicate.

### Ban enforced at login
- `AccountCredentials` now carries an `IsBanned` flag, projected in SQL from `BannedAt != null` (the entity stays an implementation detail of the data tier).
- `AccountService.Login` rejects a banned account with a distinct `LoginStatus.Banned`, which `LoginController` maps to a clear *"This account has been banned."* message.
- **The ban is disclosed only after the password verifies.** This was the one real design fork the issue flagged ("told they are banned or treated as non-existent"). Checking after verification means:
  - the legitimate owner gets an accurate message instead of a misleading "invalid credentials";
  - an anonymous probe can't enumerate ban status (a wrong password still returns the generic invalid-credentials result) — and ban-status would otherwise be more sensitive to leak than the username-taken signal account creation already exposes;
  - a banned login short-circuits before loading the player or re-hashing the credential.

### Unify the last-admin "usable admin" predicate
The issue noted that `SetUserRoles`'s `otherAdminsRemain` check still counted **banned** admins as remaining (`ArchivedAt == null` only), so a role removal could strand the system with only banned admins (who, now that ban is enforced, can't log in to recover it). Extracted a shared `OtherUsableAdminsRemainAsync` (non-archived **and** non-banned) used by **both** the role-removal and archive/ban guards, and renamed `AdminLockoutPolicy.CheckRoleChange`'s parameter to match the `CheckUserAction` "usable admin" wording.

## Test plan
- [x] **Application layer** (`AccountServiceIntegrationTests`): a banned account with correct credentials returns `LoginStatus.Banned`; a banned account with a wrong password still returns `InvalidCredentials` (no ban-status leak).
- [x] **API layer** (`LoginControllerTests`): a banned login is rejected (`400` + error, no tokens/data) and **establishes no session** — the assertion the issue explicitly asked for.
- [x] **Data tier** (`AdminUsersControllerTests`): removing the last clean admin's role is rejected as `LastAdmin` when the only other admin is banned (previously would have wrongly succeeded).
- [x] Policy unit tests + named-arg call sites updated for the renamed parameter.
- [x] Full backend suite green (1,541 tests).

## Docs
- `docs/backend-admin.md`: ban is now enforced at login (after credential verification); the last-admin guards share one "usable admin" (non-archived, non-banned) definition.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GPAuGCShbwJ52asYA83oQW)_


ginderjeremiah (commented):
## Review: Enforce ban at login & unify the usable-admin lockout predicate (#983)

**Verdict:** Approve — closes the security defect from #983 cleanly, with correct layering and thorough cross-layer tests. (Posted as a comment because I'm the PR author and GitHub won't accept a self-gating verdict.)

This is a well-scoped fix. Both halves of the issue are addressed and the code holds to the project's DDD/layering rules in `docs/backend-admin.md`.

### Correctness
- The ban is enforced in `AccountService.Login` *after* the password verifies (`Game.Application/Services/AccountService.cs:79`), which is the right call: the owner gets an accurate message, an anonymous probe can't enumerate ban status, and the short-circuit happens before loading the player or re-hashing — exactly the ordering the issue asked us to decide on. Session is only created on the success path (`LoginController.cs:41`), so a rejected banned login establishes no session.
- `LoginErrorMessage` keeps its exhaustive `switch` with the throwing default (`LoginController.cs:159`), so the new `Banned` status is a build-time gap to fill, not a silent fall-through. Good.

### Layering / DDD
- `IsBanned` is projected in SQL via `ToCredentials` (`Users.cs:26`); the `User` entity never escapes the data tier, and `AccountCredentials` stays an internal read contract. Consistent with the documented "reads speak contracts, not entities" rule.
- Extracting `OtherUsableAdminsRemainAsync` (`Users.cs:308`) and routing both the role-removal and archive/ban guards through it is the right de-duplication — it's the unification the issue explicitly wanted, and the helper is correctly invoked under the `FOR UPDATE` Admin-role lock in both atomic methods so the read stays serialized with the mutation. The renamed `otherUsableAdminsRemain` parameter on `AdminLockoutPolicy.CheckRoleChange` now matches `CheckUserAction`'s wording.

### Test coverage
Strong and well-targeted, at every layer the change touches:
- Application: banned + correct creds → `Banned`; banned + wrong password → `InvalidCredentials` (the no-leak assertion).
- API: banned login → 400, no tokens/data, and `ISessionStore.GetSession` is null — the "establishes no session" assertion #983 called for.
- Data tier: removing the last clean admin's role is rejected as `LastAdmin` when the only other admin is banned, with a post-condition that the rejected change didn't mutate roles.
- Policy unit tests + call sites updated for the renamed parameter.

No battle logic is involved, so no frontend mirror is required.

### Docs
`docs/backend-admin.md` is updated to reflect ban-at-login and the single "usable admin" (non-archived, non-banned) definition shared by both guards.

No blocking findings, no should-fixes, no nits. Nicely done. Will merge once CI is green.

---

PR #988, opened 2026-06-19
Harden SocketManagerService registration/teardown robustness (#909)

### LINKED ISSUE #909: SocketManagerService registration/teardown robustness cleanups
## Problem

A few related robustness gaps in `Game.Api/Services/SocketManagerService.cs`, all low-likelihood but worth hardening on the connection lifecycle:

1. **`SocketReplacedInfo` is sent to the old socket before the new socket's registration can fail.** In `RegisterSocket`, `EmitSocketCommand(new SocketReplacedInfo(), oldSocketId)` (line ~53) runs **before** `RegisterSocketCommandListener` / `_socketRegistry.Register` — the steps that can throw (e.g. a transient Redis `SubscribeAsync` failure). On a throw, `RollbackRegistration` correctly undoes the *new* registration, but the *old* socket has already been told to disconnect — leaving the player with a closing old connection and a rolled-back new one (no working connection) on a transient blip. **Fix:** subscribe + register first; emit `SocketReplacedInfo` only once the new socket is guaranteed established.

2. **`GetSocketCommandProcessor` can hot-spin on a persistent dequeue fault.** On any exception from `GetNextAsync` (lines ~163-203) the loop logs and `continue`s with no backoff and no consecutive-failure ceiling. A sustained fault (Redis down, or a throw *before* the pop) becomes a tight loop hammering Redis and the log for the life of the connection. **Fix:** add a small backoff on consecutive failures and a threshold after which the processor breaks (the socket re-establishes on reconnect).

3. **`UnRegisterSocket` clean-teardown path is not fault-guarded.** Unlike `RollbackRegistration` (which guards each step), the normal teardown (lines ~125-130) runs three awaited ops unguarded. If `UnRegisterSocketCommandListener` throws, `CompareAndDelete` never runs, so the presence key survives its full TTL reporting a ghost session and `HasActiveSocket` lies until expiry. **Fix:** apply the same best-effort per-step guards used in `RollbackRegistration`, ensuring the presence-key release always runs.

ginderjeremiah:
## Summary

Closes #909. Three low-likelihood-but-worth-hardening robustness gaps on the socket connection lifecycle in `Game.Api/Services/SocketManagerService.cs`.

### 1. `SocketReplaced` is now sent only after the new socket is established
`RegisterSocket` previously notified the old socket (`EmitSocketCommand(SocketReplacedInfo)`) **before** the steps that can throw (Redis `Subscribe`, registry `Register`). On a transient failure the new registration rolled back but the old socket had already been told to disconnect — leaving the player with a closing old connection and no working new one.

Now the subscribe + register happen first; the replace-notification is emitted only once the new socket is guaranteed established. The notify is **best-effort** (guarded + logged): if it fails, the old socket is left to its own inactivity-watchdog teardown rather than tearing down the live new connection over a notification the old socket no longer needs.

### 2. The pub/sub command processor no longer hot-spins on a persistent dequeue fault
`GetSocketCommandProcessor` previously logged-and-`continue`d on any `GetNextAsync` exception with no backoff or ceiling, so a sustained fault (Redis down, or a throw before the pop) became a tight loop hammering Redis and the log for the life of the connection.

It now applies a backoff scaled by the consecutive-failure streak and gives up past a ceiling (`MaxConsecutiveDequeueFailures = 5`); the socket re-establishes its subscription on reconnect. The streak resets on any successful dequeue.

### 3. `UnRegisterSocket` clean-teardown is now fault-guarded
The normal teardown ran three awaited ops unguarded, so if `UnRegisterSocketCommandListener` threw, the `CompareAndDelete` presence-key release never ran — leaving the key to survive its full TTL reporting a ghost session (`HasActiveSocket` would lie until expiry).

Each step is now best-effort guarded, mirroring the rollback path. Because that made the clean-teardown and rollback paths identical, both now share a single `TeardownSocketRegistration(context, reason)` helper (the old `RollbackRegistration` is folded in), keeping the teardown logic DRY.

## How it addresses the issue
Each numbered fix maps directly to the three points in the issue.

## Test plan
- [x] New unit tests in `SocketCommandProcessorTests`:
  - `Processor_PersistentDequeueFault_AbandonsAfterCeilingInsteadOfHotSpinning` — a persistently-throwing queue makes the processor give up after exactly the ceiling (not loop forever) and log the abandonment.
  - `UnRegisterSocket_UnsubscribeThrows_StillReleasesPresenceKey` — a throwing unsubscribe is swallowed and the presence key is still compare-and-deleted on the socket's own id.
- [x] Existing socket processor / rollback / presence-lifecycle tests still pass (the `RegisterSocket` reorder + best-effort emit and the shared teardown helper).
- [x] Full `Game.Api.Tests` suite green (534 passed).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EvJqinF9BdBxR6MByKskXP)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
@@ -143,6 +143,54 @@ namespace Game.Api.Tests.Unit
             Assert.Contains("42", release.Key);
         }
 
+        [Fact]
+        public async Task Processor_PersistentDequeueFault_AbandonsAfterCeilingInsteadOfHotSpinning()
+        {
+            var commands = new CapturingCommandFactory(throwOn: _ => null);
+            var (processor, _) = BuildProcessor(commands);
+
+            // Every dequeue throws (Redis down, or a throw before the pop). Without a ceiling the loop would
+            // hot-spin for the life of the connection, hammering Redis and the log (#909); it must back off
+            // and give up after a bounded number of consecutive failures, returning rather than looping.
+            var queue = new AlwaysThrowingQueue(new InvalidOperationException("dequeue boom"));
+
+            await processor(queue);
+
+            // The processor stopped after exactly the consecutive-failure ceiling rather than spinning, and
+            // it logged the abandonment so the give-up is observable.
+            Assert.Equal(SocketManagerService.MaxConsecutiveDequeueFailures, queue.Attempts);
+            Assert.Contains(_logs.Entries, e => e.Level == LogLevel.Error && e.Message.Contains("Abandoning"));
+        }
+
+        [Fact]
+        public async Task UnRegisterSocket_UnsubscribeThrows_StillReleasesPresenceKey()
+        {
+            // The clean-disconnect teardown runs the same guarded best-effort steps as a rollback: even when
+            // the unsubscribe throws, the presence-key release must still run so the key can't survive its
+            // full TTL reporting a ghost session that makes HasActiveSocket lie until expiry (#909).
+            var cache = new RecordingCacheService();
+            var pubSub = new ThrowingUnsubscribePubSubService(new InvalidOperationException("unsubscribe boom"));
+            var scopeFactory = _provider.GetRequiredService<IServiceScopeFactory>();
+            var registry = new SocketConnectionRegistry(new NoOpHostLifetime(), NullLogger<SocketConnectionRegistry>.Instance);
+            var manager = new SocketManagerService(
+                pubSub, cache, new CapturingCommandFactory(_ => null), scopeFactory, _loggerFactory, registry);
+
+            var socket = new FakeWebSocket(sendDuration: TimeSpan.Zero);
+            var session = new SessionService(new NoOpSessionStore());
+            session.CreateSession(userId: 1, playerId: 77);
+            var context = await manager.RegisterSocket(socket, session);
+
+            // The unsubscribe fault during teardown must be swallowed, not propagated...
+            await manager.UnRegisterSocket(context);
+
+            // ...and the presence key was still released via a compare-and-delete keyed on the socket's own
+            // id (so a newer owner's key would be left intact), on the player's presence key.
+            var release = Assert.Single(cache.CompareAndDeletes);
+            Assert.Contains("77", release.Key);
+            Assert.Equal(context.SocketId, release.DeleteIfValue);
+            Assert.Contains(_logs.Entries, e => e.Level == LogLevel.Error && e.Message.Contains("unsubscribe"));
+        }
+
         private (Func<IPubSubQueue, Task> Processor, CapturingPubSubService PubSub) BuildProcessor(CapturingCommandFactory commandFactory)
         {
             var capturingPubSub = new CapturingPubSubService();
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Harden `SocketManagerService` registration/teardown robustness (#909)

**Verdict:** Approve — all three robustness gaps from the issue are addressed correctly, the change is well-scoped to the API socket-lifecycle layer, and each fix is backed by a targeted test. (Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR; the intended verdict is **Approve**.)

### Verification against the issue

- **(1) `SocketReplacedInfo` ordering** — The notify now runs *after* `RegisterSocketCommandListener` + `_socketRegistry.Register`, and is best-effort (guarded + logged). This is strictly better than the old behavior: a transient fault in the registration steps no longer tears down the old socket over a now-irrelevant notification, and a failure to notify the old socket no longer rolls back the live new connection. The old socket falls back to its inactivity watchdog. Correct.
- **(2) Dequeue hot-spin** — Linear backoff scaled by the streak (`100ms × n`) with a ceiling of 5, streak reset on any successful dequeue. Caps a sustained fault at ~1s / 5 attempts per processor invocation instead of an unbounded tight loop. Consistent with the existing `break`-on-empty behavior, and the subscription remains for re-invocation. Correct.
- **(3) `UnRegisterSocket` guarding** — Folding the clean-teardown and rollback paths into the shared `TeardownSocketRegistration(context, reason)` is a clean DRY win and guarantees the compare-and-delete presence release runs even if the unsubscribe throws, closing the ghost-session window. Step order matches the original. Correct.

### Test coverage

Sufficient and on-target:
- `Processor_PersistentDequeueFault_AbandonsAfterCeilingInsteadOfHotSpinning` asserts the exact ceiling (`Attempts == MaxConsecutiveDequeueFailures`) and the abandonment log — pins the "give up, don't spin" contract.
- `UnRegisterSocket_UnsubscribeThrows_StillReleasesPresenceKey` asserts the presence key is still compare-and-deleted on the socket's own id after a thrown unsubscribe — pins the most important guarantee of fix (3).
- Existing processor / rollback / presence-lifecycle tests cover the `RegisterSocket` reorder and shared-helper refactor. No frontend mirror needed (API infra only, not battle logic).

- The new ceiling test incurs ~1s of real wall-clock delay (100+200+300+400ms) since the backoff isn't behind a time abstraction. This matches the established pattern in this codebase (no `TimeProvider`; real `Task.Delay` is used throughout the existing socket/integration tests), so it's consistent rather than a defect — just noting it as the one place the suite pays a small real-time cost.
- The production backoff `Task.Delay` (`SocketManagerService.cs:221`) isn't passed a cancellation token, unlike `SocketHandler`'s inactivity delay. The bound is ≤400ms and the processor signature carries no token today, so the impact is negligible; worth a glance only if the processor ever gains cooperative-cancellation plumbing.

### What's good

Genuinely tidy hardening: the reordering reasoning is sound, the shared teardown helper removes duplication while keeping the per-step best-effort guards, the `reason` parameter keeps the failure logs distinguishable between the rollback and clean-disconnect paths, and the thorough comments explain *why* each guard exists. Tests assert the precise contracts rather than just "it ran."

CI is still in progress, so I'm leaving the merge to the pipeline going green rather than merging now.

---

Issue #941, opened 2026-06-18
Battle parity & engine test-matrix gaps (fractional values, refresh-after-expiry, render catch-up, boss re-entrancy)


ginderjeremiah:
## Problem

The FE/BE battle parity surface is in good shape and densely covered, but the parity matrix and engine tests have a few gaps clustered exactly where a future divergence is most likely to hide. Both suites (`Game.Core.Tests/Battle/BattleSimulatorParityTests.cs` and `UI/src/tests/.../battle-simulation-parity.test.ts`) must stay row-for-row aligned, so add these to **both**.

### Parity matrix gaps

1. **No scenario uses fractional enemy attribute distributions.** `AttributeDistribution` (`Game.Core/Attributes/AttributeDistribution.cs:32-41`) computes `(double)(BaseAmount + AmountPerLevel * level)` from authored `decimal`s and serializes the doubles to the client (`EnemyInstance.FromSource`), which re-derives them. **Every** parity scenario uses integer amounts. A fractional `AmountPerLevel` (e.g. `2.5`) is a realistic authored value and the most plausible place a JSON float-roundtrip divergence surfaces an off-by-one kill-tick. Add one mirrored scenario with a fractional per-level distribution and a tight kill margin.

2. **No effect apply → expire → re-apply scenario.** The backend keys expiry to an absolute `_elapsedMs` clock (`Game.Core/Battle/Battler.cs:121-184`); the FE decrements `remainingMs` (`UI/src/lib/battle/battler.ts:138-194`). The matrix covers refresh-while-active and single-shot expiry, but not the lapse-then-re-apply cycle, where the two bookkeeping models are most likely to drift by a tick. Add a scenario where a periodic skill's effect lapses for a tick and is re-applied. (See also the related code-comment note below.)

3. **Crit/dodge/block scenarios only force chances to 1 or 0.** That validates draw ordering + math but never validates that a real `[0,1)` `Mulberry32` draw compared against a **fractional** chance lands identically on both sides. `Mulberry32ParityTests` pins the RNG values so risk is low, but add one scenario with a fractional chance (e.g. `CriticalChance ≈ 0.5`) and a fixed seed asserting the crit/no-crit sequence matches.

### Engine test gaps (frontend)

[...]

PR #989, opened 2026-06-20
Close battle parity & engine test-matrix gaps (#941)

ginderjeremiah:
## Summary

Closes the FE/BE battle parity-matrix and engine-test gaps called out in #941 — the spots clustered exactly where a future divergence is most likely to hide — and fixes the one real defect the review surfaced.

Every new parity scenario is added **row-for-row to both** suites (`Game.Core.Tests/Battle/BattleSimulatorParityTests.cs` ↔ `UI/src/tests/lib/battle/battle-simulation-parity.test.ts`) so the two simulators stay pinned tick-for-tick.

## Parity matrix (3 new mirrored scenarios)

| Scenario | What it pins |
| --- | --- |
| `fractionalEnemyDistribution` | Enemy Strength built from a **fractional per-level** `AttributeDistribution` (`2.5`/level at level 3 → `7.5` → `MaxHealth 87.5`) — every other row uses an integer `BaseAmount` at level 1. A **tight kill margin** (cum. damage `87` against a `87.5` max) means a float roundtrip that dropped the `.5` would change the kill tick (1600 → 1200ms). The backend `MakeEnemy` helper gains a `level` + per-level-distribution parameter; a mirrored focused test pins the `87.5` max on both sides. |
| `effectExpiresThenReapplies` | A periodic effect whose duration (80ms = 2 ticks) is shorter than its cooldown (120ms = 3 ticks), so it **fully lapses then re-applies** — the apply→expire→re-apply cycle where the backend's absolute `ExpiresAtMs` clock and the frontend's decrementing `remainingMs` are most likely to drift by a tick. |
| `fractionalCritChance` | A real `0.5` `CriticalChance` against the fixed parity seed, so each player fire's crit depends on the **actual `Mulberry32` `[0,1)` draw** (not a forced 1/0). The 6th draw being a crit is decisive for the kill tick (2400 vs 2800ms). |

## Engine fix + tests

- **`render-engine.ts`: floor `logicalDelta` at 0.** The logical engine's `tickSizeX5` tab-background catch-up branch advances its clock by the discarded excess, which can momentarily push `logicEngine.time` past the render clock, making `logicalDelta` negative and driving the render-only charge/effect interpolation **backwards for a frame** (`updateRenderCooldowns`/`updateRenderEffects` cap with `Math.min` but had no `Math.max(0, …)` floor). Purely cosmetic — logical state and battle parity are unaffected. Clamping at the source is DRY (covers both render-cooldown and render-effect consumers). Adds a render-engine clamp test and a logical-engine catch-up-branch test.
- **`enemy-manager-boss.test.ts`: the missing re-entrancy boundary.** `resolveBossVictory` re-checks `bossLoopActive` after each `await`; the `claimVictory` and challenge-reload boundaries were already covered, but the **victory-overlay-delay** boundary was not. Added a test asserting an auto-fight re-challenge is abandoned when a retreat lands during the overlay.

## Code-comment hardening

Documented on the frontend `Battler.advanceEffects` that the decrementing `remainingMs` and the backend's absolute `ExpiresAtMs` are **value-equal only under the current fixed tick size** and must stay in lockstep (the cheap option offered alongside scenario 2).

## Already-covered gap

The `CombatFloaters` unmount-during-pending-timer cleanup (#941 item 6) already has an explicit regression test (`clears pending removal timers and unsubscribes on unmount`) that passes, so no change was needed.

## Test plan

- [x] `dotnet build` + `dotnet test Game.Core.Tests` — **637 passed** (incl. the new parity scenarios and pin test)
- [x] `vitest run` over the affected battle/engine suites + CombatFloaters — **431 passed**
- [x] Verified the three scenarios' `totalMs` independently and confirmed FE/BE agree

https://claude.ai/code/session_018zJFxXdDY2H5ayE5kRNVxJ

---
_Generated by [Claude Code](https://claude.ai/code/session_018zJFxXdDY2H5ayE5kRNVxJ)_

### DIFF
--- a/Game.Core.Tests/Battle/BattleSimulatorParityTests.cs
+++ b/Game.Core.Tests/Battle/BattleSimulatorParityTests.cs
@@ -171,6 +171,28 @@ namespace Game.Core.Tests.Battle
                     ExpectedPlayerDied: false,
                     ExpectedTotalMs: 7840),
 
+                // Fractional enemy attribute distribution (#941): the only scenario whose enemy attribute is
+                // built from a FRACTIONAL per-level AttributeDistribution (BaseAmount 0 + 2.5/level at level 3
+                // → Strength 7.5) rather than an integer BaseAmount at level 1 — the production path
+                // EnemyInstance.FromSource serializes the resulting double to the client, which re-derives it.
+                // Strength feeds only MaxHealth here: 50 + 5×7.5 = 87.5, a fractional max with a TIGHT kill margin.
+                //   Player: skill baseDamage 31, no multiplier, cooldown 400 → 31−2 Def = 29/hit every 10 ticks.
+                //   Enemy:  MaxHealth 87.5, Def 2, no skills (the player never dies).
+                //   Cumulative 29,58,87 (the 87.5 survives the 87), 116 → dies on hit 4 at tick 160 → 1600ms. Had
+                //   the .5 been lost to a float roundtrip (MaxHealth 87) the enemy would die a hit earlier at 1200.
+                ["fractionalEnemyDistribution"] = new ParityScenario(
+                    Player: () => MakeBattler(
+                        strength: 0, endurance: 0,
+                        skills: [MakeSkill(1, baseDamage: 31, cooldownMs: 400)]),
+                    Enemy: () => MakeEnemy(
+                        strength: 0, endurance: 0,
+                        skills: [],
+                        level: 3,
+                        perLevelDistributions: [(EAttribute.Strength, 0m, 2.5m)]),
+                    ExpectedVictory: true,
+                    ExpectedPlayerDied: false,
+                    ExpectedTotalMs: 1600),
+
                 // A self Strength buff: the hit that CARRIES the effect uses the pre-effect attributes, and
                 // only subsequent hits see the boost (which also raises damage via the Strength→damage path)
                 // — and re-applying it every fire now STACKS (each fire adds another +10), so the buff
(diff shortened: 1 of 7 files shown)


ginderjeremiah (commented):
## Review: Close battle parity & engine test-matrix gaps (#941)

**Verdict:** Approve — closes every #941 item against both parity suites, the one real defect (negative `logicalDelta`) is fixed at the right layer, and the change is fully and correctly tested. (Posted as a comment since the review identity matches the PR author; treat it as an approval.)

This is a tidy, high-quality tech-debt PR. I traced the engine fix and the test scaffolding rather than just the diff:

### Engine fix — correct and at the right layer
`render-engine.ts:45` clamps `logicalDelta` to ≥ 0 at the source instead of inside `updateRenderCooldowns`/`updateRenderEffects`. I confirmed `logicalDelta` is consumed only by `battle-engine.ts:308-314`'s render-only interpolation (the `onRenderUpdate` hook at line 100; the line-185 hook ignores it), so no consumer ever wants the negative value — clamping once at the source is the DRY choice and covers both the render-cooldown and render-effect paths. The matching `logical-engine.test.ts` catch-up test (`engine.time` reaches 1000 while only 5 ticks fire) pins exactly the condition that produced the negative delta, so the two tests bracket the bug from both ends.

### Parity matrix — mirrored row-for-row
All three new scenarios (`fractionalEnemyDistribution`, `effectExpiresThenReapplies`, `fractionalCritChance`) are added to both `BattleSimulatorParityTests.cs` and `battle-simulation-parity.test.ts` with identical expected `totalMs`, satisfying the project's hard rule that battle-logic scenarios stay aligned across both ports. The `MakeEnemy` helper change adds `level`/`perLevelDistributions` as optional params with backward-compatible defaults, so existing callers are untouched. The FE side hardcodes the post-serialization `Strength 7.5` (it consumes the wire double rather than re-deriving from per-level distributions) while the BE builds it via the `2.5/level @ level 3` distribution — this is the correct mirror of the production path, and the paired pin tests (`Parity_FractionalEnemyDistribution_DerivesFractionalMaxHealth` ↔ the FE `derives ... MaxHealth` test) assert `87.5` on both sides so a wire-roundtrip divergence is reported directly rather than only as an off-by-one kill tick.

### Boss re-entrancy + comment hardening
The new `enemy-manager-boss.test.ts` case covers the previously-untested third `resolveBossVictory` await boundary (retreat landing during the victory-overlay delay), completing the set called out in #941 item 5. The `battler.ts` `advanceEffects` comment documenting the FE `remainingMs` ↔ BE `ExpiresAtMs` value-equality-only-under-fixed-tick seam is the cheap hardening the issue asked for, and item 6 (`CombatFloaters` cleanup) was correctly identified as already covered.

### Test coverage
Excellent. Every code change has a dedicated regression test, the parity additions are mirrored, and the scenario math is documented inline so the expected ticks are auditable.

No blocking items, no should-fixes, no nits. I'll hold the merge until CI goes green.

---

PR #901, opened 2026-06-18
Graceful shutdown + cancellation plumbing for DataProviderSynchronizer (#890)

ginderjeremiah:
## Summary

Fixes #890. `DataProviderSynchronizer` — the hosted service that drains the player write-behind queue — was the one drain coordinator that did nothing on shutdown (`StopAsync` returned `Task.CompletedTask`), unlike `SocketConnectionRegistry` and `ReferenceCacheSynchronizer`. As a result:

- a host stop mid-drain could cut off an in-flight `DrainQueueAsync` / `SaveChangesAsync` abruptly instead of letting it reach a clean boundary; and
- the `cancellationToken` passed to `StartAsync` was never threaded into the startup reclaim/drain, so a large boot-time reclaim/drain couldn't be cancelled if the host was told to stop during boot.

This was never a data-loss bug (the reserve/acknowledge design reclaims in-flight items on the next boot), but it undercut the orderly drain-on-shutdown the rest of the app deliberately invests in and contradicted the documented cancellation-plumbing invariant.

## How it addresses the issue

Mirrors the graceful drain pattern already used by `SocketConnectionRegistry`:

- **`StopAsync`** signals a stopping `CancellationTokenSource`, then awaits the in-flight drain by acquiring the existing `_drainGate` within a **bounded timeout** (5s default, sized under the host's 30s shutdown timeout, configurable for tests). On timeout it logs a warning; anything left queued — or reserved but not yet acknowledged — is reclaimed and re-applied on the next startup, so the bound loses nothing.
- **The drain loops check the token between items**, so a stop ends after the current item is acknowledged (a clean boundary) rather than mid-apply. A late pub/sub wake during shutdown becomes a no-op.
- **The startup token is threaded** through a linked token source into the reclaim/drain path, so a stop requested mid-boot unwinds the startup work too. The queue ops themselves aren't cancelable (the `IPubSubQueue` interface takes no token — see note below), so the token is checked at each boundary between them.
- Implements `IDisposable` to dispose the stopping CTS, consistent with the sibling coordinators.

## Note for the reviewer

[...]

Issue #890, opened 2026-06-17
Player write-behind synchronizer skips graceful shutdown (`StopAsync` no-op, no cancellation plumbing)


ginderjeremiah:
## Problem

`DataProviderSynchronizer` is the hosted service that drains the player write-behind queue, but unlike the other drain coordinators in the codebase (`SocketConnectionRegistry`, `ReferenceCacheSynchronizer`) it does nothing on shutdown:

```csharp
// Game.DataAccess/DataProviderSynchronizer.cs
public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
```

When the host stops mid-drain, the reserve/acknowledge design (Redis `LMOVE` onto the processing list) means a *reserved* item isn't lost — it's reclaimed on the next boot — so this is not a data-loss bug. But:

- an in-flight `DrainQueueAsync` / `SaveChangesAsync` can be cut off abruptly instead of being awaited to a clean boundary, undercutting the orderly drain-on-shutdown the rest of the app deliberately invests in (`docs/backend-persistence.md` and the graceful socket drain in `backend.md`);
- the `cancellationToken` passed to `StartAsync` is never threaded into `InitSubscriber()`, `queue.ReclaimProcessingAsync()`, or `ProcessQueue(queue)`, so a large startup reclaim/drain can't be cancelled if the host is told to stop during boot. This contradicts the documented invariant that cancellation tokens are plumbed down through the application services and the data tier.

## Suggested direction

Track the active drain `Task` (and/or a stopping `CancellationTokenSource`), and in `StopAsync` flag stopping + `await` the in-flight drain within a bounded timeout, mirroring `SocketConnectionRegistry`'s drain-await. Thread the `StartAsync` cancellation token through the startup reclaim/drain path (and into the queue ops where the Redis wrapper honours it).

### Labels
bug, claude, scope: medium

PR #990, opened 2026-06-20
Thread CancellationToken into IPubSubQueue read/reserve ops (#921)

### LINKED ISSUE #921: Thread CancellationToken into IPubSubQueue read/reserve ops for cooperative cancellation
## Background

Follow-up from #890 / #901 (graceful shutdown + cancellation plumbing for `DataProviderSynchronizer`).

#890 suggested threading the `StartAsync` cancellation token "into the queue ops where the Redis wrapper honours it." PR #901 threaded the token through the loops the synchronizer controls and checks it at each **boundary between** queue ops, but did **not** make the queue ops themselves cancelable, because no `IPubSubQueue` method currently accepts a `CancellationToken`:

- `ReserveNextAsync` / `AcknowledgeAsync` / `ReclaimProcessingAsync` / `GetLengthAsync` / `GetNextAsync` / `PeekAsync` / `RemoveAsync` / `AddToQueueAsync` (and the `Subscribe`/`UnSubscribe` ops, which deliberately keep tokens off).

This is sufficient for prompt unwind today (the synchronizer's individual queue ops are short Redis round-trips), so it is a refinement, not a defect.

## Scope

Add `CancellationToken` overloads (or default-parameter tokens) to the read/reserve/reclaim `IPubSubQueue` ops and honor them in the Redis implementation (the Redis client is already wrapped to cooperatively honor a token elsewhere — see the per-command timeout plumbing in `backend.md`), then thread the synchronizer's startup/stopping token into those calls so a wedged Redis round-trip during a boot-time reclaim/drain unwinds promptly instead of only at the next boundary.

Consider whether the change is worth the cross-cutting churn across the interface + Redis implementation + all callers, or whether the boundary-check approach is the right long-term design. If the latter, close this with a short note rather than implementing.

## Acceptance

- Either: read/reserve/reclaim `IPubSubQueue` ops accept and honor a `CancellationToken`, threaded from `DataProviderSynchronizer`'s startup/stopping token; or a documented decision to keep the boundary-check approach.
- Existing graceful-shutdown tests still pass.

Relates to #890, #901.

ginderjeremiah:
## Summary

Closes #921 (follow-up to #890 / #901). Makes the `IPubSubQueue` read/reserve/reclaim ops cancelable so a **wedged Redis round-trip during a boot-time reclaim/drain unwinds promptly** instead of only at the next item boundary — closing the gap PR #901 left open and fulfilling the documented invariant that *"the Redis client is wrapped to honour [the per-command budget] cooperatively"* (`backend.md`). `RedisQueue` was the one Redis wrapper that didn't.

I chose to **implement** rather than close-with-a-note because:
- The cooperative-cancellation pattern is already established and documented for `RedisService`/`RedisPubSubService`; the queue being the lone exception was an inconsistency, not a deliberate design.
- Default-parameter tokens keep every existing caller compiling — only `DataProviderSynchronizer` threads a real token — so the "cross-cutting churn" the issue worried about is contained.

## How it addresses the issue

**Interface + Redis implementation**
- Added `CancellationToken cancellationToken = default` to the async `IPubSubQueue` ops (sync ops left tokenless, mirroring `RedisService`).
- `RedisQueue` now honours the token cooperatively: pure reads (`GetLengthAsync`/`PeekAsync`) throw up front then `WaitAsync`; mutating ops (pop/reserve/acknowledge/reclaim/remove/push) take a write path that attaches a fault-logging continuation so an abandoned write that later faults is surfaced, not silently lost. `ReserveNextAsync`/`ReclaimProcessingAsync` also check the budget up front / each pass so a stop unwinds them promptly.

**DRY cleanup**
- The cooperative-cancellation logic was already duplicated in `RedisService` and `RedisPubSubService`; implementing the queue naively would have made it three copies. Extracted it into a shared `RedisCommandBudget` helper and migrated all three wrappers onto it. `RedisPubSubService`'s publish path now delegates to the token-aware queue ops, so its own `ObserveWrite` is removed (no dead code).

**Synchronizer threading**
- `DataProviderSynchronizer` threads its startup/stopping token into the `ReclaimProcessingAsync` and `ReserveNextAsync` calls. The resulting prompt-unwind `OperationCanceledException` is treated as a clean stop (the pub/sub `BackgroundWorker` loop logs every escaping exception as an error, so this must be swallowed). Once an item is **reserved**, it is still processed and acknowledged **without** the token, so an in-flight apply always finishes cleanly at the boundary — preserving the existing drain-to-boundary semantics.

## Test plan
- [x] `Game.Application.Tests` — 332 passed (includes the existing graceful-shutdown tests, unchanged).
- [x] `Game.Infrastructure.Tests` — 42 passed.
- [x] `Game.Api.Tests` — 532 passed (socket-processor doubles updated for the new signatures).
- [x] New tests:
  - `RedisQueueIntegrationTests.ReadReserveReclaimOps_WhenTokenAlreadyCancelled_ThrowOperationCanceled` — the read/reserve/reclaim ops honour an already-cancelled budget and leave the queue untouched.
  - `DataProviderSynchronizerTests.StopAsync_WedgedReserveDuringDrain_UnwindsPromptlyViaThreadedToken` — a wedged reserve unwinds at once via the threaded token (well under the generous drain timeout) and never surfaces as an error.

## Notes
- The socket-command drain (`SocketManagerService.GetSocketCommandProcessor`) continues to use the default (None) token, so its behaviour is unchanged; threading shutdown cancellation there is outside this issue's scope and already covered by the graceful socket drain.

Relates to #890, #901.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V4dM9FaXQHbHyN7jh3TP5V)_


ginderjeremiah (commented):
## Review: Thread `CancellationToken` into `IPubSubQueue` read/reserve ops (#921)

**Verdict:** Request changes (posted as a comment since I'm the PR author) — the change itself is clean and correct, but `test-backend` is red: the merge with current `main` no longer compiles.

### Blocking

- **CI `test-backend` fails to build (11× CS0535).** `main` has advanced 2 commits since this branch's merge-base and added a third `IPubSubQueue` test double — `AlwaysThrowingQueue` — in `Game.Api.Tests/Unit/SocketCommandProcessorTests.cs:305` (part of a +93-line addition to that file). This branch updated the two doubles it knew about (`FakeQueue`, `RecordingQueue`) but `AlwaysThrowingQueue` doesn't exist on the branch yet, so it never got the new signatures. The git merge is *textually* clean, so this surfaces only as a semantic break at compile time: the merged `AlwaysThrowingQueue` still implements the old tokenless `IPubSubQueue`, and CI reports `'AlwaysThrowingQueue' does not implement interface member 'IPubSubQueue.GetNextAsync(CancellationToken)'` (and the other 10 ops).

  **Fix:** merge/rebase `main` into the branch, then add `CancellationToken cancellationToken = default` to every async method on `AlwaysThrowingQueue` (the same mechanical update already applied to `FakeQueue`/`RecordingQueue`). Please re-verify the full build after the merge rather than just that one class — `main`'s update to `SocketManagerService` and the other doubles came in alongside it, so confirm `test-backend` is green on the updated branch.

### What's good

This is a well-scoped, well-reasoned change and I have no other findings:

- **`RedisCommandBudget` extraction is the right call.** Implementing the queue naively would have made three copies of the cooperative-cancellation logic; pulling `Read`/`Write` into one helper and migrating `RedisService` + `RedisPubSubService` onto it removes the duplication and the now-dead `ObserveWrite` methods, with no behavior change. The Read-throws-up-front vs Write-attaches-fault-continuation split is preserved and clearly documented.
- **The read/write classification in `RedisQueue` is correct.** Destructive ops (`GetNextAsync`/LPOP, `ReserveNextAsync`/LMOVE, acknowledge/reclaim/remove/push) take the write path so an abandoned mutation that later faults is logged not lost; pure `GetLengthAsync`/`PeekAsync` take the read path. The up-front `ThrowIfCancellationRequested()` in `ReserveNextAsync`/`ReclaimProcessingAsync` (before the Redis command is issued) is exactly what makes a cancelled budget leave the queue untouched, and the integration test proves the seed item survives.
- **Synchronizer threading preserves the drain-to-boundary semantics.** The token is threaded into reserve/reclaim only; once an item is reserved it's processed and acknowledged *without* the token, so an in-flight apply always finishes cleanly. The `catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)` guards (both in `StartAsync` and around `DrainQueueAsync` in `ProcessQueue`) correctly swallow only our own cancellation so the pub/sub worker loop doesn't log a clean stop as an error.
- Docs (`backend-persistence.md`) updated to match.

### Test coverage

Good for the paths that matter: the integration test (`ReadReserveReclaimOps_WhenTokenAlreadyCancelled_ThrowOperationCanceled`) covers the up-front cancellation contract for reserve/reclaim/length/peek and asserts the queue is untouched, and `StopAsync_WedgedReserveDuringDrain_UnwindsPromptlyViaThreadedToken` proves a wedged reserve unwinds via the threaded token well under the drain timeout without surfacing an error. The `RedisCommandBudget.Write` fault-logging continuation remains indirectly exercised only (as it was before this refactor) — not a regression, fine to leave, though a small focused test for it would be a reasonable follow-up if you want it covered directly.

Once the branch is updated against `main` and `test-backend` goes green, this is an approve.

---

ginderjeremiah:
Thanks for the review — both points addressed:

- **Build fix:** merged `main` into the branch and gave `AlwaysThrowingQueue` the same default-parameter `CancellationToken` update as the other doubles. Re-verified the full build after the merge (clean, 0 warnings) and re-ran the suites that the merge touched: `test-backend`-equivalent locally is green — Api 536, Application 334, Core 633, Infrastructure 46.
- **Optional follow-up (now included):** added `RedisCommandBudgetTests` covering the helper directly — `Read` throwing up front on an already-cancelled budget without awaiting the command, and `Write`'s fault-logging continuation surfacing an abandoned command that later faults (while a late success logs nothing). Each transition is driven via a `TaskCompletionSource` so the continuation path is deterministic rather than raced against real Redis.

Pushed as `616a59c`; CI should re-run on the updated branch.

---

PR #991, opened 2026-06-20
Small backend correctness/hygiene cleanups (#937)

### LINKED ISSUE #937: Small backend correctness/hygiene cleanups (immutability, pub/sub overload, HTTP rehydration)
## Problem

A grab-bag of small, mostly-latent backend correctness/hygiene items surfaced during a codebase health review. None is an active user-facing bug; each is a fragility or guideline mismatch worth tightening.

1. **`AttributeDistribution` isn't `required`/`sealed` like its sibling immutable reference models.** `Game.Core/Attributes/AttributeDistribution.cs:15-25` declares `AttributeId`/`BaseAmount`/`AmountPerLevel` as `init`-only but **not** `required`, unlike `Item`/`ItemMod`/`Skill`/`SkillEffect`/`EnemyTemplate` (all `required` + `sealed`). A distribution built without `AttributeId` silently defaults to the enum-0 attribute and zero amounts. It's part of the shared cached `EnemyTemplate` graph, so it should match its peers. (Adjacent to the #896 immutability work but a distinct, un-addressed member.)

2. **`RedisPubSubService` wake-vs-enqueue `Publish` overload is a footgun.** `Game.DataAccess/Repositories/Admin/PlayerUpdateDeadLetters.cs:76` calls `_pubsub.Publish(Constants.PUBSUB_PLAYER_CHANNEL, "")` — a bare channel wake. It resolves correctly today, but the two-arg `(channel, message)` and three-arg `(channel, queueName, queueData)` overloads are easy to confuse, and a future edit adding a third argument would silently switch overloads and enqueue `""` as a poison message onto the player queue. **Fix:** a dedicated `Wake(channel)` method that removes the overload ambiguity.

3. **`RehydrateSession` writes the session cache on the HTTP `Status`/`ActiveSession` path.** `Game.Api/Services/SessionInitializer.cs:41-51` (via `LoginController.Status`/`ActiveSession`) writes the session cache off the HTTP thread on a cache miss. `backend.md` is emphatic that player-state writes belong on the socket precisely because HTTP runs concurrently with the command loop. The early `HasPlayerSession` return mitigates clobbering a *populated* session, and `ActiveSession` is documented to run before a socket exists — so the practical race window is narrow. Still: this is an **undocumented exception** to "writes only on the socket." Either document it explicitly (and confirm `RehydrateSession`/`SessionStore.Update` can't clobber a concurrently-establishing socket session), or have `EnsureSessionLoaded` rehydrate without persisting an empty `PlayerState`. A focused unit test on `SessionInitializer`'s three-way branch (hit → no-op, miss → rehydrate, no-player → graceful skip) is also missing.

4. **`DataProviderSynchronizer.ProcessMessage` retry loop has no terminal action after the loop.** `Game.Infrastructure/.../DataProviderSynchronizer.cs:234-268` dead-letters inside the final `catch`, but if the attempt arithmetic were ever changed control could fall off the loop end returning normally without dead-lettering — a silent drop, in the one place a player write must never be silently lost. Add a defensive post-loop `throw`/dead-letter so the terminal path is unmistakable.

ginderjeremiah:
## Summary

Closes #937. Four small, mostly-latent backend correctness/hygiene items from a codebase health review. None was an active user-facing bug; each tightens a fragility or guideline mismatch.

### 1. `AttributeDistribution` is now `required` + `sealed`
`AttributeId`/`BaseAmount`/`AmountPerLevel` are now `required` (and the class `sealed`), so a distribution can never be silently built with the default enum-0 attribute and zero amounts. It's part of the shared cached `EnemyTemplate` graph and now matches the `required` shape of its sibling reference models.

> **Note on the issue's premise:** the issue states `Item`/`ItemMod`/`Skill`/`SkillEffect`/`EnemyTemplate` are "all `required` + `sealed`". In `Game.Core` they are all `required`, but only `EnemyTemplate` and the value-object element `AttributeModifier` are actually `sealed` — `Item`/`ItemMod`/`Skill`/`SkillEffect` are not. I matched the verified-consistent `required` and additionally sealed `AttributeDistribution` because it's a leaf immutable value object (mirroring its element-type peer `AttributeModifier` and its aggregate `EnemyTemplate`). The reflection immutability guard only checks init-only/read-only, so `required`/`sealed` don't affect it.

### 2. Dedicated `IPubSubService.Wake(channel)` for the bare channel wake
`PlayerUpdateDeadLetters` previously woke the synchronizer via `Publish(channel, "")` — the two-arg `(channel, message)` overload. That resolved correctly but is a footgun: a future edit adding a third argument would silently switch to the `(channel, queueName, queueData)` overload and enqueue `""` as a poison message. A dedicated `Wake(channel)` removes the ambiguity. The two internal wake sites in `RedisPubSubService` (`Publish(channel, queueName, ...)` and `PublishBatch`) now route through it too, consolidating the bare-wake concept in one place.

The two-arg `Publish(channel, message)` overload stays — it's still legitimately used by `ReferenceCacheSynchronizer` to publish a real message (the instance id) for cross-instance reference-data invalidation.

### 3. HTTP rehydration no longer writes the session cache
`RehydrateSession` (reached from `Status`/`ActiveSession` via `SessionInitializer.EnsureSessionLoaded`) previously persisted an empty `PlayerState` off the HTTP thread on a cache miss. `backend.md` is emphatic that player-state writes belong on the socket. Rehydration now re-binds the request's `PlayerState` **in memory only**; the session cache is (re)populated only by login (which establishes the binding before any socket exists) and by the socket's write-behind path.

This removes the narrow-but-real clobber window (HTTP read-miss → a socket concurrently establishes+saves → the HTTP empty-write overwrites it). Login already primes the cache, so this rehydration path only fired after an eviction or on a genuinely uncached request anyway. Documented in `backend.md → Authentication`.

### 4. Defensive terminal guard in `DataProviderSynchronizer.ProcessMessage`
The retry loop dead-letters inside the final `catch`, but if the attempt arithmetic were ever changed, control could fall off the loop end and return normally — a silent drop, in the one place a player write must never be silently lost. Added a defensive post-loop error-log + dead-letter so the terminal path is unmistakable and preserves the write even if the bounds change.

## Testing notes
- **Item 4** is a defensive guard that is unreachable under any valid configuration: `RetryPolicy` rejects `MaxAttempts < 1`, so the loop always runs at least one attempt and every terminal branch returns. Per `docs/backend.md → Testing Guidelines`, such defensive guards are accepted as measure-only and cannot be exercised without an invalid policy, so no unit test targets it.
- **Item 3 — `SessionInitializer`'s three-way branch.** The issue asks for a focused *unit* test. `SessionInitializer` orchestrates two out-of-process-coupled collaborators — `SessionService` (cache) and `AccountService` (DB, a 7-dependency service with no fakes; it is integration-tested against real Postgres/Redis). Per the project's classical-school testing guidelines, such dependency interactions are covered by integration tests rather than doubles. The three branches are already exercised end-to-end through the real controller:
  - **hit → no-op:** `Login_IssuedAccessToken_AuthenticatesProtectedEndpoint` (login primes the cache; the subsequent `Status` hits it)
  - **miss → rehydrate:** `Status_/ActiveSession_ValidTokenWithNoSessionCache_Rehydrates...` (updated to assert the in-memory-only, no-persist behavior)
  - **no-player → graceful skip:** `Status_AuthenticatedButPlayerNotLoadable_Returns404WithError`
  - The `SessionService`-level change is unit-tested directly (`RehydrateSession_EstablishesSessionInMemoryWithoutWritingTheCache`, plus a new `CreateSession_PrimesTheCache`).

## Test plan
- [x] `Game.Core.Tests` — 633 passed
- [x] `Game.Api.Tests` (unit + integration) — 533 passed
- [x] `Game.Application.Tests` (incl. pub/sub + synchronizer) — 330 passed
- [x] `Game.Infrastructure.Tests` — 42 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01VSAUoR1bLYHJ58cFjwsqBq)_


ginderjeremiah (commented):
## Review: Four small backend correctness/hygiene cleanups (#937)

**Verdict:** Approve on the merits — _but cannot be merged yet_: the branch has a real conflict with `main` that must be resolved first (details below). Posting as `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR.

All four items are faithful to the issue, well-scoped, and well-documented. I verified each against the codebase and the docs:

- **Item 1 — `AttributeDistribution` `required` + `sealed`:** every construction site uses an object initializer setting all three properties (incl. the production `Game.DataAccess/Mapping/EnemyMapper.cs:54`), so `required` is satisfied everywhere and the build stays clean. Honest correction of the issue's premise in the description (siblings are `required` but not all `sealed`) — sealing this leaf value object to match `AttributeModifier`/`EnemyTemplate` is the right call.
- **Item 2 — dedicated `Wake(channel)`:** removes the wake-vs-enqueue overload footgun and consolidates the two internal `RedisPubSubService` wake sites + `PlayerUpdateDeadLetters` through it. Confirmed the one remaining two-arg `Publish(channel, message)` call (`ReferenceCacheSynchronizer.cs:52`, the real instance-id message) is correctly left untouched.
- **Item 3 — in-memory-only rehydration:** correctly aligns the HTTP `Status`/`ActiveSession` path with `backend.md`'s "player-state writes belong on the socket," closing the narrow read-miss clobber window. `CreateSession`/`RehydrateSession` deliberately diverge (one primes the cache, one doesn't), so inlining the former `EstablishSession` helper into both is justified rather than dead duplication.
- **Item 4 — defensive terminal guard:** clean and unmistakable; confirmed `backend.md` (line 23) accepts such defensive guards as measure-only, so the absence of a dedicated unit test is per-guidelines.

### Blocking (merge-readiness)

- **Conflict with `main` in `Game.Infrastructure/PubSub/Redis/RedisPubSubService.cs`** — the PR is `mergeable_state: dirty`. Since this branch was cut, #990 landed on `main` and threaded `CancellationToken` directly into the queue ops (`AddToQueueAsync(queueData, cancellationToken)` / `AddRangeToQueueAsync(values, cancellationToken)`) on the exact lines this PR rewrites — this branch still carries the older `ObserveWrite(...)` wrapper that no longer exists on `main`. Rebase/merge `main` in and resolve by **keeping main's `CancellationToken`-threaded queue writes** and **routing the wake through the new `Wake(channel)`** (the consolidation still applies cleanly on top of #990). CI has not run yet and won't be meaningful until this is resolved.

### Test coverage

- Good. New `SessionService` unit tests (`RehydrateSession_EstablishesSessionInMemoryWithoutWritingTheCache`, `CreateSession_PrimesTheCache`) pin the divergent cache behavior directly; the integration tests are updated to assert the no-persist rehydration (`AssertSessionNotEstablishedAsync`). Item 3's `SessionInitializer` three-way branch is reasonably argued as integration-covered under the classical-school guidelines. No battle logic touched, so no frontend/backend parity obligation.

### What's good

- Honest, precise commit/PR narrative that corrects the issue where it was slightly wrong rather than blindly implementing it. The new `Wake` XML doc and the `backend.md` rehydration paragraph are exactly the kind of "why" context that prevents a future edit from re-introducing each footgun. Nicely surgical for a four-item grab-bag.

Once the conflict is resolved and CI is green, this is good to merge.

---

PR #993, opened 2026-06-20
Add level/XP progress to the Attributes screen (#177)

### LINKED ISSUE #177: Visualize player level progress
- Yellow bar under player health bar/card in the Fight screen
- Add a yellow exp bar with current/total somewhere in the Attributes screen

ginderjeremiah:
## Summary

Completes the second half of [#177](https://github.com/ginderjeremiah/GameServer/issues/177) — the fight-screen XP bar was already shipped; this adds level/XP progress to the **Attributes** (build) screen.

A new `LevelMeter` sits beside the points `BudgetMeter` in the budget row and renders the player's **level** plus a visible **current / total XP** readout over the shared gold `XpBar` primitive. Grouping the two meters reads as one progression dashboard — XP → level → stat points to spend — and `LevelMeter` mirrors `BudgetMeter`'s label/value layout (just gold instead of the points accent).

## Design notes

- **Reuses the existing `XpBar`** primitive (the same gold, shimmering, level-up-flashing bar used on the fight screen) rather than hand-rolling a new bar — per the "extend an existing primitive" guideline.
- **Reads `playerManager` directly** (level/exp/threshold), like the fight-screen `BattlerCard`, since XP progression is global player state and unrelated to the screen's point-buy `AttributesView` model.
- All colours go through existing CSS variables (`--gold`, `--text-tertiary`, `--text-muted`); no hard-coded hex.

## Testing

- New unit test in `Attributes.test.ts` renders the real screen and asserts the XP bar's `progressbar` role, `aria-valuenow`/`aria-valuemax`, the 40% fill, and the visible level + current/total readout.
- `vitest` (14 passed), `svelte-check` (0 errors), ESLint, and Prettier all pass.

> Note: a live in-browser walkthrough wasn't done — the Attributes screen only renders behind the full backend stack (Postgres + Redis + API + auth + a websocket-loaded session), which the Vite dev server can't exercise on its own. Verification is via the unit tests above plus reuse of the already-shipped `XpBar`/`BudgetMeter` surfaces.

Closes #177

### DIFF
--- a/UI/src/routes/game/screens/attributes/LevelMeter.svelte
+++ b/UI/src/routes/game/screens/attributes/LevelMeter.svelte
@@ -1,0 +1,68 @@
+<!-- Character progression meter for the Attributes screen: the player's level plus a visible
+     current/total XP readout over the shared gold XpBar. Sits beside the points BudgetMeter and
+     mirrors its label/value layout, so the two read as one progression dashboard (XP → level →
+     stat points to spend). Reads the player manager directly, like the fight-screen card. -->
+<div class="level-meter">
+	<div class="row">
+		<span class="label">Level {playerManager.level}</span>
+		<span class="value">
+			{formatNum(playerManager.exp)}<span class="xp-total"
+				>&nbsp;/&nbsp;{formatNum(playerManager.nextLevelThreshold)} XP</span
+			>
+		</span>
+	</div>
+	<XpBar
+		level={playerManager.level}
+		exp={playerManager.exp}
+		nextLevelThreshold={playerManager.nextLevelThreshold}
+		ariaLabel="Experience progress"
+		testId="attributes-xp-bar"
+	/>
+</div>
+
+<script lang="ts">
+import { formatNum } from '$lib/common';
+import { playerManager } from '$lib/engine';
+import { XpBar } from '$components';
+</script>
+
+<style lang="scss">
+.level-meter {
+	display: flex;
+	flex-direction: column;
+	gap: 7px;
+	min-width: 210px;
+}
+
+.row {
+	display: flex;
+	align-items: baseline;
+	justify-content: space-between;
+	gap: 12px;
+	white-space: nowrap;
+}
+
+.label {
+	font-family: var(--mono);
+	font-size: 9.5px;
+	letter-spacing: 1.4px;
+	text-transform: uppercase;
+	color: var(--text-tertiary);
+}
+
+.value {
+	font-family: var(--sans);
+	font-size: 18px;
+	font-weight: 600;
+	letter-spacing: -0.3px;
+	white-space: nowrap;
+	color: var(--gold);
+	text-shadow: 0 0 12px color-mix(in srgb, var(--gold) 40%, transparent);
+}
+
+.xp-total {
+	font-size: 11px;
+	font-weight: 400;
+	color: var(--text-muted);
+}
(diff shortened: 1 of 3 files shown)


UI/src/routes/game/screens/attributes/LevelMeter.svelte
```diff
@@ -0,0 +1,68 @@
...
+
+.value {
+	font-family: var(--sans);
+	font-size: 18px;
+	font-weight: 600;
+	letter-spacing: -0.3px;
+	white-space: nowrap;
+	color: var(--gold);
+	text-shadow: 0 0 12px color-mix(in srgb, var(--gold) 40%, transparent);
+}
```
ginderjeremiah, line 61: **Should-fix (DRY):** this readout block — the markup (`.row` > `.label` + `.value` > `xp-total`) and these `.row`/`.label`/`.value` style rules — is essentially a verbatim copy of `BudgetMeter.svelte`'s readout. The two only differ in the accent color (`--gold` vs `--accent`) and the depleted state. `CLAUDE.md` asks us to "prefer writing DRY code whenever possible," and now that there are two parallel "label + value + bar" meters this duplication is worth removing.

A couple of light options that keep the two visually in sync:
- Extract a small shared `MeterReadout` component that takes the label text and value content (e.g. via snippets) plus an accent token, used by both meters; or
- Pull the shared `.row`/`.label`/`.value` rules into an SCSS placeholder/mixin in the styles layer.

Not blocking — but please address it or consciously decline before merge.

---

ginderjeremiah (commented):
## Review: Add level/XP progress to the Attributes screen (#177)

**Verdict:** Approve (intent) — solid, well-scoped change; one DRY should-fix below. Posted as a comment because GitHub won't accept a gating verdict from the PR author's own account.

This completes the Attributes half of #177: a `LevelMeter` beside the points `BudgetMeter` renders the player's level + a current/total XP readout over the shared gold `XpBar`. The approach is right — it reuses the existing `XpBar` primitive rather than hand-rolling a bar, and reads `playerManager` directly exactly like the fight-screen `BattlerCard` does, so the two surfaces stay consistent.

### Should-fix

- `LevelMeter.svelte:37-61` — the readout markup + `.row`/`.label`/`.value` SCSS is a near-verbatim copy of `BudgetMeter.svelte`. With two parallel meters now, this duplication is worth extracting (shared component or SCSS placeholder) per `CLAUDE.md`'s DRY guidance. Details inline. Not blocking.

### Nits

- `.meters` allows shrinking (`min-width: 0`), but both meters keep `min-width: 210px`, so the row can't actually compress below ~452px + hint. Matches the pre-existing `BudgetMeter` width, so no regression — just noting it if narrow viewports ever become a concern.

### Test coverage

- Good: the new test renders the real screen and asserts the `progressbar` role, `aria-valuenow`/`aria-valuemax`, the 40% fill, and the visible level + current/total readout — meaningful behavior, not just "it renders." Mirrors the existing test style in the file. This is frontend-only UI, so no backend/battle-logic mirroring is required.

### What's good

- Correct reuse of the `XpBar` primitive and the direct-`playerManager` read pattern from `BattlerCard`.
- All colors go through theme tokens; `--gold` intentionally matches the bar's own token rather than the themeable `--accent`, which is the right call for an XP surface.
- Component stays small and the grouping into one "progression dashboard" reads well.

---

PR #994, opened 2026-06-20
Stack skill effects with a single shared, resetting expiration (#992)

### LINKED ISSUE #992: Skill effects do not stack as intended
Despite #740 being absolutely clear on how skill effects are supposed to stack, they were not implemented correctly.

Any new instance of a skill effect should stack and use the duration of the new application. There should NOT be independent expirations for portions of the stacked effect.

ginderjeremiah:
## Summary

Closes #992. Re-applying a skill effect was implemented (in #740) as a *second timed modifier with its own independent expiry*, but #740 actually requires the stacked effect to **share a single expiration that resets to the newest application's duration**. The two models coincide at the *instant* of application (the magnitudes match), which is what masked the bug — the divergence only shows up in how the stack expires.

This restores the intended behavior from #740's worked example:

> apply 10 DoT for 5s → re-apply 20 DoT for 2s → **30 DoT for 2s** → re-apply 10 DoT for 5s → **40 DoT for 5s**

## What counts as "the same effect" (design decision)

`#740`'s text ("a skill effect … applied again") and its worked example pointed in slightly different directions, so I confirmed the grouping with the issue owner: the shared expiration is keyed **by target attribute**. All active timed modifiers on the same attribute on a battler share one expiry that resets to the latest application — exactly reproducing the worked example (where the short 2s re-application cuts the older 5s one short, and the later 5s one extends the whole stack).

## Changes

**Runtime (`Game.Core/Battle/Battler.cs` ↔ `UI/src/lib/battle/battler.ts`, kept parity-identical)**
- Applying an effect still adds its own modifier, so magnitudes **sum** (additive add, multiplicative compound).
- All active applications on the **same attribute** now share **one expiry**: applying any effect on that attribute resets every active application of it to the new duration — extending a longer-lived application or cutting a shorter one short. The whole stack expires together; there are **no independent per-portion expirations**. Different attributes remain independent.

**UI**
- Active-effect chips now group by `(attribute, modifier type)` so the chip's single shared countdown matches the new expiry, the count spans every contributing application, and the combined total is computed from each application's **actual** amount (correct even when caster-scaling makes stacked amounts differ — the old `combineEffectAmount(amount × count)` shortcut assumed uniform amounts).
- The chip tooltip breaks down each application's amount and source skill; the shared remaining is the header countdown pill. Removed the now-dead `combineEffectAmount` helper.

**Tests & docs**
- Mirrored unit tests flipped from independent expiry to shared-reset (extend, cut-short, and cross-attribute independence) on **both** the backend and frontend suites.
- New mirrored parity scenario `dotReapplyExtendsSharedExpiration` pins the shared-reset behavior — it resolves at **280ms** under the fix and would have resolved at **400ms** under the old independent-expiry model, so it's a real regression guard.
- `game-design.md` / `backend-battle.md` document the corrected rule; the backend performance-test comments are updated (a re-applied persistent effect now never lapses, accumulating rather than holding a bounded window).

## Test plan
- [x] `dotnet build Game.sln` — clean (0 warnings, 0 errors)
- [x] `Game.Core.Tests` — 640 passed (incl. the new effect/parity/performance scenarios)
- [x] Frontend `npm run lint` (eslint + svelte-check) — clean
- [x] Frontend `npm run test:unit:ci` — 2136 passed (209 files)

## Notes / minor follow-up
- A single attribute carrying **both** additive and multiplicative timed effects at once (rare, authored) now renders as two chips that share one countdown — so re-applying the additive also resets the multiplicative chip's sweep. This is a faithful consequence of the per-attribute shared expiry and a very rare case; flagging it here rather than filing an issue.
- Unbounded stacking is unchanged and still tracked by the existing follow-up #976 (per-effect cap).

---
_Generated by [Claude Code](https://claude.ai/code/session_01K77TUdhdhtwzJyBy8GUYqh)_


UI/src/routes/game/screens/fight/ActiveEffectChips.svelte
```diff
@@ -74,105 +73,108 @@ type Props = {
...
-	}
-	for (const group of groups) {
-		group.totalAmount = combineEffectAmount(group.modifierType, group.stackAmount, group.count);
-		group.applications.sort((a, b) => b.remainingMs - a.remainingMs);
+		// Additive amounts sum; multiplicative factors compound — matching the battle math, and read from
+		// each application's own amount (they can differ when caster-scaling shifts between fires).
+		group.totalAmount =
+			effect.modifierType === EModifierType.Multiplicative
+				? group.totalAmount * effect.amount
+				: group.totalAmount + effect.amount;
```
ginderjeremiah, line 122: **Should-fix (test coverage):** This block is where the two genuinely *new* chip behaviors live, and neither is exercised by `ActiveEffectChips.test.ts` (which wasn't touched):

1. **Grouping now keys on `(attribute, modifierType)`** instead of `sourceId`, so two *different* source skills hitting the same attribute+type collapse into a single chip — a real semantic change from the old per-`sourceId` grouping. Every existing test uses either one source or distinct attributes, so the collapse is untested.
2. **`totalAmount` is summed from each application's own `amount`** — the headline fix over the removed `combineEffectAmount(amount × count)` shortcut, whose whole point was handling caster-scaled applications with *differing* amounts. Yet every existing chip test stacks uniform `amount: 5`s, so the differing-amount path the change exists to fix has no regression guard here. (`AttributeTooltip.test.ts` covers the differing-amount *display* in isolation, but not this grouping/sum logic.)

The logic itself is correct — I traced it and the existing suite still passes. But per CLAUDE.md ("thoroughly tested... including edge cases") this newly-introduced behavior should have its own coverage: a test that stacks two applications with different amounts (and ideally different `sourceId`s) on the same attribute+type, asserting one chip with the summed total. Cheap to add given the existing fixtures.

---

ginderjeremiah (commented):
## Review: Stack skill effects with a single shared, resetting per-attribute expiration (#992 / #740)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating verdict from me) — the fix is correct, parity-faithful, and well-documented; one should-fix on UI test coverage and one rare-case nit, neither blocking.

I traced the worked example from #740 against the new model and it holds exactly (`apply 10/5s → +20/2s ⇒ 30/2s → +10/5s ⇒ 40/5s`), and the runtime change is a clean, minimal reset of the shared expiry keyed by target attribute. The backend (`ExpiresAtMs` clock) and frontend (`remainingMs` countdown) stay value-equal under the fixed tick size, and the new mirrored parity scenario `dotReapplyExtendsSharedExpiration` is a genuine regression guard (280ms under the fix vs 400ms under the old independent-expiry bug — math checks out: stacks accumulate 10/20/40/60/90/120 against the 120-HP enemy, death on tick 7).

### Should-fix
- **`ActiveEffectChips.svelte` — the two new chip behaviors lack component-level tests** (see inline comment): grouping now collapses different source skills on the same `(attribute, modifierType)` into one chip, and `totalAmount` is summed from each application's actual amount (the fix the removed `combineEffectAmount` shortcut couldn't do). The existing suite still passes but only ever uses uniform amounts / single sources, so neither new path has a regression guard. Add a differing-amount stack test.

### Nits
- **Rare cross-type denominator skew.** The shared expiry resets per *attribute*, but chips group per `(attribute, modifierType)`. When an attribute carries both an additive and a multiplicative timed effect, re-applying one resets the other chip's `renderRemainingMs` to the new duration while that chip's `durationMs` denominator stays its own authored duration — so its radial sweep jumps to a fraction rather than full on a reset. You flagged the shared-countdown coupling in the PR notes; this denominator detail is the visible consequence. Display-only and genuinely rare (requires authored additive+multiplicative on one attribute), so fine to leave — worth a one-line code comment near `remainingSweep` so a later edit doesn't "fix" it the wrong way.

### Test coverage
- Backend `BattlerEffectsTests`: shared-reset extend, cut-short, and cross-attribute independence all added; the MaxHealth-clamp test correctly refactored to put the two effects on *different* attributes (since same-attribute would now share expiry). Good.
- Frontend `battler-effects.test.ts`: mirrors the backend scenarios 1:1, including the stacked-view test asserting the older application is reset back to full. Good.
- Parity matrix: new scenario mirrored identically on both sides. Good.
- `AttributeTooltip.test.ts` updated to differing amounts + per-source rows; `skill-effect-display.test.ts` drops the now-dead `combineEffectAmount` tests cleanly.
- Gap: the `ActiveEffectChips` grouping/sum logic (see should-fix).

### What's good
- Genuinely parity-disciplined: backend and frontend runtime changes, docs, and the parity scenario all moved together, and the divergence-prone apply→expire→re-apply cycle is pinned on both sides.
- Docs (`game-design.md`, `backend-battle.md`) and the performance-test comments were all corrected to the new rule rather than left stale — including the subtle "persistent effects now never lapse / accumulate" consequence.
- Dead `combineEffectAmount` helper and its tests removed rather than left dangling; the tooltip context shape was reshaped (`stackAmount`/per-row `remainingMs` → per-row `amount`/`sourceName`) to match the new single-shared-expiry reality instead of being patched around.


---

PR #1002, opened 2026-06-20
Cache zero-based-id save fixup metadata; confirm two read-path overheads (#938)

### LINKED ISSUE #938: Backend hot-path overhead: per-save zero-based-id fixup, per-battle progress reload, per-call catalog re-map
## Problem

Three "verify-then-optimize" overhead items on the busiest backend write/read paths (an idle game resolves a battle per cooldown per connected player). Each may be acceptable; flagging for a deliberate measure-and-decide.

1. **`GameContext.SaveChangesAsync` runs the zero-based-id fixup unconditionally on every save.** `Game.Infrastructure/Database/GameContext.cs:538-573` iterates `ChangeTracker.Entries()` and does reflection-y metadata walks (`GetContainingForeignKeys`, `IsAssignableTo`) on **every** `SaveChangesAsync`, including the hot write-behind handler saves that touch 1–2 rows. **Fix:** short-circuit when there are no `Added` entities with temporary keys, and/or cache the per-entity-type "has a zero-based-identity FK" decision rather than re-deriving it from metadata each save.

2. **`BattleStatisticsEventHandler` reloads the progress aggregate on every battle completion.** `Game.Application/Events/BattleStatisticsEventHandler.cs:21` calls `_progressRepo.Load(...)` per `BattleCompletedEvent` — i.e. per idle battle tick resolution. It's documented cache-first (likely a warm Redis read), but it's still a round-trip per battle on a path that already holds the player aggregate in memory. **Fix / decision:** confirm whether the progress aggregate can be co-located on the same connection-scoped session as the player to avoid the per-battle load, or accept it as intentional given the separate write-behind key and note why.

3. **Reference `All()` projections re-map the full catalog per call.** `Game.DataAccess/Repositories/Enemies.cs:17-20` (`All()` does `Snapshot.Enemies.Select(EnemyMapper.ToContract).ToList()` fresh each call); the peer reference repos do the same. The version-hash path is memoized per snapshot, so this is likely only hit on full catalog reads — but **verify the loading-screen / admin read path isn't calling `All()` per connect**. If it is, projecting once per snapshot swap (mirroring the existing hash memo) removes a per-connect O(N) map.

All three are "confirm it's actually on the hot path, then fix or document." None is a correctness issue.

ginderjeremiah:
## Summary

Addresses the three "verify-then-optimize" overhead items in #938. Each was traced to confirm whether it is genuinely on the per-battle hot path, then **fixed** (item 1) or **measured, confirmed cold, and documented** (items 2 & 3).

### 1. `GameContext.SaveChangesAsync` zero-based-id fixup — fixed

The fixup ran on **every** save by walking each `ChangeTracker` entry's properties and re-deriving PK/FK metadata via reflection (`GetContainingForeignKeys`, `IsAssignableTo`, `IsPrimaryKey`, `IsForeignKey`) — paid on the hot write-behind saves that touch 1–2 rows per battle tick per connected player.

Which property on a given entity type is a zero-based PK (or a non-nullable `int` FK to a zero-based principal) is **fixed by the model**, so it's now derived **once per model and cached**. The per-save loop looks the entity type up in that cache, skips irrelevant types outright, and only flips the temporary key on the precomputed targets.

**Behavior is unchanged.** I deliberately did *not* take the issue's "short-circuit when there are no `Added` entities" suggestion: the PK branch fires on a **Modified** record-0 edit (editing the first row of a zero-based set), so gating on `Added` would have broken it. The metadata-caching approach keeps both branches intact:
- **PK branch** — a `Modified` zero-based record whose `Id == 0` (the identity seed, indistinguishable from `default(int)`), which EF would mark temporary and resolve against the wrong row.
- **FK branch** — a non-nullable `int` FK pointing at record 0 of a store-generated zero-based principal (e.g. `EnemySkill.EnemyId`, `PlayerChallenge.ChallengeId`), which EF likewise marks temporary.

### 2. `BattleStatisticsEventHandler` per-battle progress load — confirmed intentional, documented

The handler reloads the progress aggregate per `BattleCompletedEvent` (a cache-first, typically warm Redis read). Co-locating it on the connection-scoped session like the player aggregate is **not cleanly achievable**:
- Progress is a **separate write-behind aggregate** specifically to keep it off the player blob (which is re-serialized in full on every save). Folding it back in reintroduces exactly that bloat.
- The handler lives in `Game.Application` and only sees the player aggregate; the connection-scoped holder (`SessionService`) lives in the presentation layer (`Game.Api`) and is reached via a **new DI scope per command**. Co-locating would invert that layer dependency.

So the per-battle load is accepted over those costs. Documented in the handler and `docs/backend-persistence.md`.

### 3. Reference `All()` projection — confirmed cold, documented

The loading screen issues `GetReferenceDataVersions` **first** and fetches a set only when its (memoized-per-snapshot) version hash differs from its cached copy (`reference-data.ts` → `hydrateFromCacheIfCurrent` / `hydrateAllFromCache`). So the full `All()` projection runs on a **content change or a first/cleared cache — not per connect**. The hot per-connect path (the version hash) is already memoized, so a per-snapshot projection memo would optimize a path that isn't hot. Documented on `AbstractReferenceDataCommand`. The only other `All()` callers are admin paths (`TagsController`, `AdminChallenges`), also not per-connect.

## How it addresses the issue

All three items asked to "confirm it's actually on the hot path, then fix or document." Item 1 was on the hot path and is optimized with no behavior change; items 2 & 3 were confirmed cold and the deliberate decisions are now recorded so a future health review doesn't re-flag them.

## Test plan

- **New** `Game.Infrastructure.Tests/GameContextZeroBasedIdentityFixupTests.cs` (6 in-process unit tests): the per-entity-type fixup-plan derivation (PK target, FK targets, exclusion of unrelated types) plus the FK fixup forcing the real `0` and the `Added`-row key being left store-generated. (The model builds without a DB connection, so these stay in-process per the project's lightweight-`Game.Infrastructure.Tests` rule; the end-to-end save path is covered by the integration suites below.)
- `Game.Infrastructure.Tests` — **52 passed**
- `Game.Application.Tests` (admin saves, account creation, write-behind, progress repo, idempotency, battle service — the integration coverage of the real save path) — **334 passed**
- `Game.Api.Tests` — **537 passed**

A note on coverage for the PK branch: in EF Core 10 a `Modified` zero-based record-0 key isn't marked temporary at fixup time (the fixup runs before `base.SaveChangesAsync`), so it isn't reproducible as a pure in-process behavior test. The branch is preserved exactly (a guarded no-op when the key isn't temporary) and its end-to-end behavior rides on the admin integration tests. I'll open a small follow-up to confirm whether that branch is still reachable in EF Core 10 and remove it if it's provably dead.

Closes #938

---
_Generated by [Claude Code](https://claude.ai/code/session_016o9Ab9xSwr7FoHjjx2WfBF)_


Game.Application/Events/BattleStatisticsEventHandler.cs
```diff
@@ -18,6 +18,14 @@ ISkills skills
...
         public async Task HandleAsync(BattleCompletedEvent domainEvent, CancellationToken cancellationToken = default)
         {
+            // Progress is loaded per battle (a cache-first, typically warm Redis read) rather than co-located
+            // in memory with the player for the connection's lifetime. That is deliberate: progress is a
+            // separate write-behind aggregate under its own key precisely to keep it off the player blob, which
+            // is re-serialized in full on every save and would otherwise grow with the stat set (see
+            // docs/backend-persistence.md). Co-locating it would either fold it back into that blob (the bloat
+            // the split avoids) or require a connection-scoped holder reachable from this application-layer
+            // handler — which only sees the player aggregate, not the presentation-layer session — so the
+            // per-battle load is accepted over inverting that dependency to save one warm cache read.
```
ginderjeremiah: Nit: this comment is nearly a verbatim copy of the new paragraph in `docs/backend-persistence.md`. Two literal copies of the same rationale are a drift risk — when one is edited the other silently goes stale, which is exactly what `CLAUDE.md`'s "avoid over-documenting" guidance warns against. Since the decision now lives in the persistence doc (the right home for cross-cutting "why" narrative), consider trimming this to a short pointer — e.g. *"Progress is loaded per battle rather than co-located with the player; see docs/backend-persistence.md for why."* — and letting the doc carry the full reasoning. Take it or leave it.

---

ginderjeremiah (commented):
## Review: zero-based-id save-fixup caching + documenting two confirmed-cold read paths (#938)

**Verdict:** Approve (posted as a comment because I'm the PR author, so GitHub won't accept a gating event from me) — the hot-path fixup is optimized with behavior preserved, and items 2 & 3 are correctly confirmed cold and documented. One nit, no blockers.

Note: I am **not merging**. CI is still in flight (`test-backend`, `test-e2e (firefox)`, `test-e2e (webkit)` were in progress at review time) and there's an open nit, so I'm leaving the merge to a green run.

### What's good
- **Correct, behavior-preserving optimization.** The per-save `ChangeTracker` walk no longer re-derives PK/FK metadata via reflection on every save; it's derived once per `IModel` and cached. Keying the cache on `Model` is the right call — EF returns a process-stable model singleton, so the dictionary effectively holds one entry and never grows unboundedly.
- **Semantics are preserved on both branches.** `ForceZero`'s `IsTemporary` guard reproduces the original "only flip a temporary key" behavior; the PK branch keeps its `State is not Added && Id: 0` gate, and the FK branch still targets only non-nullable `int` FKs to zero-based principals (nullable FKs stay excluded, which is correct — a `0` in a nullable FK isn't read as unset). The early `fixups.Count == 0` short-circuit is a nice touch.
- **Good judgment on the issue's suggestion.** Declining the "short-circuit when there are no `Added` entities" idea is correct — it would have broken the `Modified` record-0 PK branch. The PR explains this clearly.
- **Items 2 & 3 are genuinely verified, not hand-waved.** The progress-reload and `All()`-projection analyses trace the actual call paths (version-hash memoization gating the full projection; the layer boundary preventing co-location) and land on documented, defensible decisions.

### Nits
- `BattleStatisticsEventHandler.cs:21-28` — the in-handler comment duplicates the new `docs/backend-persistence.md` paragraph nearly verbatim (drift risk). Suggest trimming to a short pointer to the doc. (inline)

### Test coverage
- `GameContextZeroBasedIdentityFixupTests` covers the derivation thoroughly: PK identification, FK-to-zero-based-principal identification, exclusion of unrelated types, the FK force-to-zero, the no-op on a concrete key, and the Added-row key being left store-generated. Appropriately in-process per the lightweight `Game.Infrastructure.Tests` rule.
- The one path not directly unit-tested — forcing a *temporary* `Modified` record-0 PK — is honestly called out as not reproducible in-process under EF Core 10 (the fixup runs before `base.SaveChangesAsync`, before EF marks it temporary), preserved as a guarded no-op, and covered end-to-end by the admin integration suites. That's an acceptable, well-reasoned gap.

### Follow-up
- The PR promises a small follow-up to confirm whether the `Modified` record-0 PK branch is still reachable under EF Core 10 (and remove it if provably dead). Please open that issue (tag `tech debt`, `claude`, `scope: small`) so the dead-branch question is tracked.

---

ginderjeremiah:
Thanks for the review. Addressed the nit in `463fd48` — trimmed the in-handler comment to a short pointer and left the full rationale in `docs/backend-persistence.md`.

The follow-up issue for the `Modified` record-0 PK branch is already open: #1003 (tagged `tech debt`, `claude`, `scope: small`).

---

ginderjeremiah:
## Review — approve (with optional nits)

I reviewed this against #938, the diff, the docs, and the local checkout.

**Verdict: approve.** Correct, behavior-preserving, thread-safe, well-tested for the testable branches, and scoped exactly to the issue. CI is fully green (`test-backend`, all three `test-e2e` browsers, `detect-changes`, `all-checks-passed`; `test-frontend` correctly skipped — no frontend changes).

### What I verified
- **Behavior is genuinely unchanged.** Walked both branches old-vs-new. PK branch: old required `temporary + IsPrimaryKey + IZeroBasedIdentityEntity + Id==0 + !Added`; new requires the precomputed `Id` PK target + `!Added + Id==0`, with `ForceZero` re-guarding `IsTemporary`. FK branch: old required `temporary + int + IsForeignKey + zero-based principal`; new precomputes "int FK → zero-based principal" and `ForceZero` re-guards `IsTemporary`. Equivalent in both cases.
- **Slightly more correct than before:** the old PK branch picked an arbitrary temporary PK property via `FirstOrDefault(IsPrimaryKey())`; the new code targets the property literally named `Id`. No real entity is affected (composite-key tables aren't `IZeroBasedIdentityEntity`), so it's a no-op improvement, not a regression.
- **Thread-safety is sound.** EF caches the compiled `IModel` per provider+options, so the `static readonly ConcurrentDictionary<IModel, …>` holds effectively one immutable entry — not an unbounded leak. The `GetOrAdd` factory is pure/idempotent, so a benign double-build under a race is harmless. The fixups mutate only `this` context's `ChangeTracker`.
- **Good judgment declining the issue's "short-circuit on no `Added` entities" suggestion** — the PK branch fires on a *Modified* record-0 edit, so that gate would have silently broken editing the first row of a zero-based set. That suggestion was a latent bug; correct call to skip it.
- Items 2 & 3 confirmed cold and documented (handler + `docs/backend-persistence.md`; `AbstractReferenceDataCommand`). Guideline adherence is clean — braces everywhere, no `!` introduced, `internal` persistence detail stays in `Game.Infrastructure`, tests placed per the in-process `Game.Infrastructure.Tests` rule.

### Follow-up
The PK-branch reachability follow-up you committed to is already tracked in **#1003** — good. I'd reinforce one point there: the follow-up should land an **integration test that edits record 0 and asserts the `UPDATE` hits the right row** *before* deleting the branch, rather than removing it on a hunch — that test is the only thing that would catch a silent EF-version behavior change in either direction.

### Optional nits (non-blocking, take or leave)
- `GameContext.cs` — the "derived once per model" doc comment is accurate, but a reader unfamiliar with EF's model caching might read it as "per context instance." A half-sentence noting EF shares the `IModel` across contexts would preempt that.
- `BuildZeroBasedFixups` being `internal static` purely for testability is fine given `InternalsVisibleTo`; a one-word "test-exposed" note is optional.

Not auto-merging only because of the optional nits above — happy to merge once you've glanced at them (or confirm you'd rather leave them and I'll merge).

---

ginderjeremiah:
## Review summary — ✅ Approve

Reviewed against the project's standards (DDD/layering, DRY, mandatory braces, null-handling, test coverage, docs), with extra attention to caching correctness, thread-safety, and the "behavior unchanged" claim.

**What it does:** Resolves the three "verify-then-optimize" items in #938.
- **Item 1 (fixed):** `GameContext.SaveChangesAsync` previously re-derived zero-based PK/FK metadata via reflection on *every* save by walking each `ChangeTracker` entry — paid on the hot write-behind path. Since the set of zero-based PKs / non-nullable `int` FKs-to-zero-based-principals is fixed by the EF model, it's now derived **once per `IModel`** into a `static readonly ConcurrentDictionary` and the per-save loop just looks up the entity type and flips only the precomputed temporary keys.
- **Items 2 & 3 (confirmed cold, documented):** the per-battle progress reload in `BattleStatisticsEventHandler` and the reference `All()` projection were traced, confirmed off the hot per-connect path, and the deliberate decisions recorded so a future health review doesn't re-flag them.

**Correctness / thread-safety (verified line-by-line):**
- Both selection predicates are reproduced exactly: PK = `IZeroBasedIdentityEntity` with an `Id` PK; FK = non-nullable `int` (`int?` like `BossEnemyId`/`UnlockChallengeId` stay excluded, matching the old code) whose FK principal is an `IZeroBasedIdentityEntity`. `ForceZero`'s `IsTemporary` guard preserves the original "only flip a temporary key" semantics on both branches.
- Thread-safe: EF caches one immutable `IModel` per provider+options, so the static dictionary holds effectively one entry (no unbounded growth); `GetOrAdd`'s factory is pure/idempotent (benign double-build under a race); the cached plan is read-only after build and the only mutation touches the per-context `ChangeTracker`.
- "Behavior unchanged" holds. Targeting the property named `Id` instead of `FirstOrDefault(IsPrimaryKey())` is a no-op tightening for every implementer (all single-column `Id` PKs). The author correctly declined the issue's "short-circuit when no `Added` entities" suggestion — it would have broken the `Modified` record-0 PK branch.

**CI:** Green — `all-checks-passed`, `test-backend`, `test-e2e (chromium/firefox/webkit)`, `detect-changes` all succeeded; frontend checks correctly skipped.

**Findings:** No blocking or should-fix issues. Optional only: a half-sentence in the `GameContext` comment clarifying that EF shares the `IModel` singleton across contexts would preempt a "per-context-instance?" misread. The handler-comment-duplication nit is already trimmed at HEAD (463fd48).

**Follow-up:** The PK-branch-reachability question (is the `Modified` record-0 branch still reachable in EF Core 10?) is already tracked in #1003. Recommend #1003 also land an integration test that edits record 0 and asserts the `UPDATE` targets the right row before removing the guard, so a silent EF-version behavior change is caught either way rather than removed on a hunch.

Merging.

---

PR #1004, opened 2026-06-20
Game.Core progress/player correctness & hygiene cleanups (#912)

### LINKED ISSUE #912: Game.Core progress/player correctness &amp; hygiene cleanups
## Problem

A grab-bag of small, mostly-latent correctness/hygiene issues in `Game.Core` (player + progress aggregates), surfaced during a codebase health review. None is an active user-facing bug today; each is a fragility or guideline mismatch worth tightening. (Distinct from the model-hygiene items already tracked in #896.)

1. **`ZonesCleared` first-clear check reads a magic `0` instead of row presence.** `PlayerProgress.cs:78-82` uses `if (GetStatisticValue(EStatisticType.ZonesCleared, zoneId) == 0)`, the exact pattern the documented invariant forbids ("'no data yet' is the *absence* of the row, never a magic 0"). The rest of the file keys off `TryGetStatisticValue`. Works today only because the per-zone row never legitimately holds 0 — a latent double-count trap. **Fix:** branch on `!TryGetStatisticValue(..., out _)`.

2. **`EvaluateChallenges` persists empty (zero-progress, incomplete) challenge rows on first relevance.** `PlayerProgress.cs:131-146` adds a newly-relevant challenge to `_dirtyChallenges` unconditionally because `isNew` is true — even when `UpdateChallengeProgress` made no actual change. This persists information-free rows via write-behind and partially undoes the `RelevantTo` reverse-index optimization. **Fix:** mark dirty only when something actually changed (drop the `isNew ||` term, or defer inserting the `PlayerChallenge` until it has non-zero progress or is completed).

3. **`NewPlayerFactory` hardcodes the attribute count, decoupled from the `EAttribute` core set.** `NewPlayerFactory.cs:15,40-42` uses `AttributeCount = 6` + `Enumerable.Range(0, 6).Select(id => (EAttribute)id)`, assuming the first six enum values are exactly the allocatable primaries. If a seventh primary is added, new players get no allocation row for it, and `PlayerStatPoints.TryUpdateAttributes` rejects updates targeting an attribute with no row — permanently blocking allocation into the new stat. **Fix:** derive the seed set from the `EAttribute` values whose `EAttributeType` is `Primary`.

4. **Idempotent unlocks raise domain events on the no-op path.** `Inventory.UnlockItem`/`UnlockMod` early-return on a duplicate and `Player.UnlockSkill` checks for an existing skill, yet `Player` raises `ItemUnlockedEvent`/`SkillUnlockedEvent` unconditionally (`Player.cs:75-79, 96-104`). Re-granting an already-owned reward produces redundant write-behind work and a redundant "unlocked!" client notification. **Fix:** have the inventory unlock methods return whether a change occurred and raise the event only when it did; move the `UnlockSkill` event inside the `if`.

5. **`EquipmentSlot` / `UnlockedItemSlot` expose `ItemId` and `Item` as independently-settable public properties.** `EquipmentSlot.cs:28,33` keeps them consistent only via the private `Set`/`Clear` helpers; the snapshot path reads `slot.ItemId` (`BattleSnapshot.cs:44`) while live attribute composition reads `slot.Item` (`Inventory.cs:62`). Production paths use `Set`/`Clear`, but the dual public setters are a footgun on the parity surface (a desync would silently fail legitimate victories). **Fix:** make `Item` the single source of truth and derive `ItemId` from it (`public int? ItemId => Item?.Id;`), or make both setters private.

6. **Multi-player selection is duplicated and unreachable.** `AccountService.cs:81,114-118` hardcodes `PlayerIds[0]` in both `Login` and `ResolveSelectedPlayerId`. The model exposes a plural `PlayerIds` but there's no selection mechanism, so players 2..N are unreachable and the rule is duplicated. **Fix:** either document/constrain the single-player invariant, or factor the selection into one method and add a real selection path. (Low urgency — flagging the duplication + latent assumption.)

ginderjeremiah:
## Summary

Addresses the grab-bag of latent correctness/hygiene issues in `Game.Core`'s player + progress aggregates from #912. None was an active user-facing bug; each was a fragility or guideline mismatch.

1. **`ZonesCleared` first-clear keys off row presence, not a magic 0.** `PlayerProgress` now branches on `!TryGetStatisticValue(...)` (the documented invariant) instead of `GetStatisticValue(...) == 0`, removing the latent double-count trap.
2. **`EvaluateChallenges` no longer persists info-free challenge rows.** A newly-relevant challenge is committed to the aggregate (and marked dirty) only once it actually gains progress or completes — so a zero-progress, incomplete row no longer hits the write-behind queue *or* the cache snapshot, restoring the `RelevantTo` reverse-index scoping.
3. **`NewPlayerFactory` derives the allocation seed from the core attribute set.** Replaces the hardcoded `AttributeCount = 6` with `Enum.GetValues<EAttribute>().Where(Attribute.IsCore)`, so adding a seventh core attribute automatically grants new players its allocation row (without one, `PlayerStatPoints` permanently rejects allocations into it).
4. **Idempotent unlocks raise no event on the no-op path.** `Inventory.UnlockItem`/`UnlockMod` now return whether a change occurred, and `Player` raises `ItemUnlocked`/`ModUnlocked`/`SkillUnlocked` only when something actually unlocked — no redundant write-behind work or client "unlocked!" notice on a re-grant.
5. **`EquipmentSlot`/`UnlockedItemSlot` make `Item` the single source of truth.** `ItemId` is now derived (`Item?.Id` / `Item.Id`) rather than an independently-settable property, so the two can never desync on the battle-parity surface.
6. **`AccountService` factors the duplicated player selection.** `Login` and `ResolveSelectedPlayerId` now share one `SelectPlayerId` helper instead of each hardcoding `PlayerIds[0]`.

## Notes / decisions

- **Item 3 sources from the core set, not the display `Primary` taxonomy.** The issue suggested deriving from `EAttributeType.Primary`, but `Enums.cs` and `backend.md` deliberately keep the display taxonomy *distinct* from the core/allocatable power-calc invariant. The set that actually governs which attributes get an allocation row (and which `PlayerStatPoints.TryUpdateAttributes` accepts) is `Attribute.IsCore`, so I seeded from that — the more correct source.
- **Item 6 (low urgency) was treated as the DRY/document option, not a new multi-player selection feature.** The single-player-per-account invariant is documented on the new `SelectPlayerId` helper, which is the single seam a real selection mechanism would grow from. A real multi-player selection path remains future work and is out of scope here.

## Test plan

- [x] `dotnet build` clean (0 warnings)
- [x] `Game.Core.Tests` — 649 passed
- [x] `Game.Application.Tests` — 334 passed
- [x] `Game.Api.Tests` — 537 passed

New coverage: no-op unlock events (item 4), derived `ItemId` on both slot types (item 5), presence-gated first clear on a loaded aggregate (item 1), and the no-info-free-row challenge path (item 2). Items 3 & 6 are pinned by the updated `NewPlayerFactoryTests` and the existing `AccountServiceIntegrationTests`. No battle math changed, so no frontend parity changes; nothing is `[ClientMirrored]`, so no codegen drift.

Closes #912

---
_Generated by [Claude Code](https://claude.ai/code/session_01XyETdxoik4drm1QMNzP83M)_

### DIFF
--- a/Game.Application/Services/AccountService.cs
+++ b/Game.Application/Services/AccountService.cs
@@ -81,12 +81,13 @@ namespace Game.Application.Services
                 return AccountLoginResult.Failed(LoginStatus.Banned);
             }
 
-            if (account.PlayerIds.Count == 0)
+            var selectedPlayerId = SelectPlayerId(account.PlayerIds);
+            if (selectedPlayerId is null)
             {
                 return AccountLoginResult.Failed(LoginStatus.NoPlayer);
             }
 
-            var player = await _playerRepo.GetPlayer(account.PlayerIds[0]);
+            var player = await _playerRepo.GetPlayer(selectedPlayerId.Value);
             if (player is null)
             {
                 return AccountLoginResult.Failed(LoginStatus.PlayerDataNotFound);
@@ -122,6 +123,17 @@ namespace Game.Application.Services
         public async Task<int?> ResolveSelectedPlayerId(int userId)
         {
             var playerIds = await _users.GetPlayerIds(userId);
+            return SelectPlayerId(playerIds);
+        }
+
+        /// <summary>
+        /// Selects which player an account's session binds to. The game is single-player per account today,
+        /// so this is always the first (and only) player; centralizing it here keeps the login and
+        /// session-rehydration paths from diverging and gives a future player-selection mechanism a single
+        /// seam to grow from. Returns <see langword="null"/> when the account has no player.
+        /// </summary>
+        private static int? SelectPlayerId(IReadOnlyList<int> playerIds)
+        {
             return playerIds.Count > 0 ? playerIds[0] : null;
         }
 
--- a/Game.Core.Tests/Players/InventoryTests.cs
+++ b/Game.Core.Tests/Players/InventoryTests.cs
@@ -33,6 +33,16 @@ namespace Game.Core.Tests.Players
             Assert.Single(inventory.UnlockedItems);
         }
 
+        [Fact]
+        public void UnlockItem_ReportsWhetherItWasNewlyUnlocked()
+        {
+            var inventory = new Inventory();
+            var item = MakeItem(42);
+
+            Assert.True(inventory.UnlockItem(item));
+            Assert.False(inventory.UnlockItem(item));
+        }
+
         // ── UnlockMod ───────────────────────────────────────────────────────
 
         [Fact]
@@ -54,6 +64,15 @@ namespace Game.Core.Tests.Players
             inventory.UnlockMod(7);
 
             Assert.Single(inventory.UnlockedMods);
+        }
+
+        [Fact]
+        public void UnlockMod_ReportsWhetherItWasNewlyUnlocked()
+        {
+            var inventory = new Inventory();
+
+            Assert.True(inventory.UnlockMod(7));
+            Assert.False(inventory.UnlockMod(7));
         }
 
         // ── TryEquipItem ────────────────────────────────────────────────────
@@ -604,6 +623,30 @@ namespace Game.Core.Tests.Players
             Assert.Equal(10, Assert.Single(slot.AppliedMods).ItemModId);
             // The rebuilt index is live, so a lookup-driven operation against the restored item resolves it.
             Assert.True(restored.TryEquipItem(3, EEquipmentSlot.AccessorySlot));
+        }
+
+        // ── Slot ItemId is derived from Item (single source of truth) ────────
+
+        [Fact]
+        public void EquipmentSlot_ItemId_IsDerivedFromItem()
+        {
+            // ItemId can never desync from Item: it is computed from it rather than stored independently.
+            var slot = new EquipmentSlot(EEquipmentSlot.WeaponSlot);
+            Assert.Null(slot.ItemId);
+
+            slot.Set(MakeItem(7, EItemCategory.Weapon));
+            Assert.Equal(7, slot.ItemId);
+
+            slot.Clear();
+            Assert.Null(slot.ItemId);
+        }
+
+        [Fact]
+        public void UnlockedItemSlot_ItemId_IsDerivedFromItem()
+        {
+            var slot = new UnlockedItemSlot { Item = MakeItem(9), AppliedMods = [] };
+
+            Assert.Equal(9, slot.ItemId);
         }
 
         // ── Helpers ──────────────────────────────────────────────────────────
--- a/Game.Core.Tests/Players/NewPlayerFactoryTests.cs
+++ b/Game.Core.Tests/Players/NewPlayerFactoryTests.cs
@@ -1,5 +1,6 @@
 using Game.Core.Players;
 using Xunit;
+using CoreAttribute = Game.Core.Attributes.Attribute;
 
 namespace Game.Core.Tests.Players
 {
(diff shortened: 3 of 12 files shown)


ginderjeremiah (commented):
## Review: `Game.Core` progress/player correctness & hygiene cleanups (#912)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't accept a gating approval from me; treating this as an approve). No blocking findings.

I checked all six items against the issue, the production read/write paths, and the backend DDD/layering rules. Each is addressed correctly and the reasoning on the two judgment calls holds up.

### Item-by-item

1. **ZonesCleared first-clear** — `!TryGetStatisticValue(...)` is the documented absence-of-row invariant; the rest of the file already keys off presence, so this removes the lone magic-`0` outlier. ✓
2. **EvaluateChallenges** — Deferring the `_challenges` insert until the row actually changes is correct. I confirmed the completion path is unaffected: a row that completes always trips the `Progress`/`Completed` change check, so it's committed before the `completed.Add(...)` at the bottom. A re-relevant-but-no-progress challenge simply re-evaluates next battle via `RelevantTo`, which is the intended behavior. ✓
3. **NewPlayerFactory** — Good call sourcing from `Attribute.IsCore` rather than the issue's suggested `EAttributeType.Primary`: `IsCore` is the set that actually governs which attributes get an allocation row (and which `PlayerStatPoints` accepts), and `Attribute.cs` deliberately keeps the display taxonomy distinct. Using `Enum.GetValues().Where(IsCore)` also preserves deterministic enum order — better than iterating the `HashSet`-backed `CoreAttributes`. ✓
4. **Idempotent unlocks** — Inventory methods now return whether a change occurred and `Player` gates the event on it; `UnlockSkill`'s event moved inside the `if`. ✓
5. **Derived `ItemId`** — Verified `Item` is now the single source of truth on both slot types. The only two producers that set `ItemId` independently (`Inventory.UnlockItem` and `PlayerMapper.ToCore`) were updated; equipment slots go through `Set`/`Clear`, and `BattleSnapshot.FromPlayer` reads the derived `ItemId` cleanly via `SelectNotNull`. No stale `ItemId =` assignments remain on the core types. ✓
6. **SelectPlayerId** — Sensible DRY of the duplicated `PlayerIds[0]`, with the single-player invariant documented on the one seam a future selection path would grow from. Correctly scoped as the document/factor option, not a new feature. ✓

### Test coverage

Solid and matches the project's rules — each behavioral change has a focused test: no-op unlock events (items 4), derived `ItemId` on both slot types (item 5), the presence-gated first-clear on a *loaded* aggregate exercising the load-then-mutate path (item 1), and the no-info-free-row challenge path asserting empty `DirtyChallenges` and `ChallengeProgress` (item 2). Items 3 & 6 are pinned by the updated `NewPlayerFactoryTests` (now asserting the derived core set) and existing integration tests. No battle math changed, so the frontend/backend parity suites don't need mirrored additions.

- `NewPlayerFactory` could read marginally more intention-revealing as `Attribute.CoreAttributes.Select(...)` since that's the named concept — but the current `Enum.GetValues().Where(IsCore)` form is justified by its deterministic ordering, so this is a wash. Fine as-is.

### What's good

Tightly scoped to the issue, no gold-plating, every change lands in `Game.Core` with mapping kept in `Game.DataAccess` — layering is clean. The judgment call on item 3 (core set vs. display `Primary`) is the more correct read of the invariant and is well-documented in both the code comment and the PR description.

Not merging yet — backend/e2e CI is still running. Once it's green this is good to go from my side.

---

ginderjeremiah:
## Review — approve

Reviewed against #912, the diff, `backend.md`, and the local checkout. All six items map cleanly to the issue and the implementation is clean, well-scoped, and idiomatic.

### What I verified
- **Item 1** (`PlayerProgress` first-clear) now branches on `!TryGetStatisticValue(...)` (row presence) instead of `== 0`, matching the documented "absence is the no-data state" invariant.
- **Item 2** (`EvaluateChallenges`) defers committing a newly-relevant `PlayerChallenge` until it gains progress/completes. I confirmed the `DirtyChallenges` invariant still holds — `_dirtyChallenges.Add` is only reachable after the `_challenges[id]` insert, so the later `_challenges[id]` lookup can't throw `KeyNotFoundException`.
- **Item 3** (`NewPlayerFactory`) seeds from `Attribute.IsCore` rather than the issue's suggested `EAttributeType.Primary` — the more correct source, since `IsCore` is the exact set `PlayerStatPoints.TryUpdateAttributes` validates. Confirmed the core set is enum values 0–5, so `Enum.GetValues<EAttribute>().Where(IsCore)` is behavior-preserving today and order-equivalent to the old `Range(0,6)`.
- **Item 4** unlock no-op events: `Inventory.UnlockItem`/`UnlockMod` return a real change flag (`HashSet<int>.Add`'s bool) and events fire only on an actual unlock.
- **Item 5** `ItemId => Item?.Id` derived on both slot types — confirmed no stray `ItemId =` initializers remain in `Game.Core`/`Game.DataAccess` (the `GameContext`/migration hits are the unrelated EF entity).
- **Item 6** `SelectPlayerId` DRY helper with the single-player invariant documented as the future selection seam.
- Built `Game.Core` (0 warnings); `Game.Core.Tests` 649 passed locally, matching the PR.

### Quality
DDD layering respected (changes stay in `Game.Core` + orchestration-only `AccountService` + the `PlayerMapper` initializer). No null-forgiving operators, braces on all control blocks, no dead code left behind (the `AttributeCount` const and `ItemId` setters are fully removed incl. their `PlayerMapper` consumer). New tests cover every behavioral change. No battle math touched, so frontend/backend parity is unaffected and there's no codegen drift.

**Verdict: approve** — no blocking issues, no should-fix items, no follow-up issues needed (item 6's real multi-player path is appropriately deferred and documented). CI is fully green. Proceeding to merge.

---

PR #1005, opened 2026-06-20
Make IApiSocketResponse.data optional so the error-path guard is type-enforced

### LINKED ISSUE #940: Make IApiSocketResponse.data optional so the error-path guard is type-enforced
## Problem

The socket transport signals **every** failure (server error, dropped connection, per-request timeout) by **resolving** with a response carrying an `error` field and **no `data`** — callers are expected to inspect `response.error` and optional-chain `response.data?.x`. But the response type declares `data` as **non-optional**:

```ts
// UI/src/lib/api/api-socket-request.ts:40-44 — settleWithError fabricates a data-less response
return { ... } as IApiSocketResponse<T>;   // <-- cast required because data is mandatory
```

So `settleWithError` (`api-socket.ts:51`) must use an `as IApiSocketResponse<T>` assertion to produce a legitimately data-less error response, and **every** caller has to remember to optional-chain `response.data`. Several callers do this correctly (`enemy-manager.ts:87,100,276,303`), but a missing `?.` throws only on the error path — which tests rarely exercise — making it an easy latent runtime-throw trap.

## Suggested fix

Make `data` optional in `IApiSocketResponse` (`data?: …`). The type system then enforces the very guard the docs describe: the cast in `settleWithError` disappears, and any caller dereferencing `response.data` without checking `error`/optional-chaining becomes a compile error. This is squarely a CLAUDE.md guideline fix ("avoid the null-forgiving operator / unsafe casts; handle nullability properly") and removes a whole class of error-path runtime throws.

ginderjeremiah:
## Summary

The socket transport signals **every** failure (server error, dropped connection, per-request timeout) by **resolving** the promise with a response carrying an `error` field and **no `data`** — callers are expected to inspect `response.error` and optional-chain `response.data?.x`. But `IApiSocketResponse.data` was declared **non-optional**, so:

- `settleWithError` had to use an `as IApiSocketResponse<T>` assertion to fabricate a legitimately data-less error response, and
- any caller dereferencing `response.data` without first checking `error` (or optional-chaining) compiled fine, throwing only on the rarely-tested error path — a latent runtime-throw trap.

This makes `data` optional (`data?: …`), so the type system now enforces the very guard the docs describe.

## How it addresses the issue (#940)

- **The cast is gone.** `settleWithError` constructs the data-less error response directly, no `as` assertion.
- **The guard is now type-enforced.** Every caller that dereferences `response.data` must guard `error` / optional-chain. Audited all socket-response consumers:
  - `enemy-manager.ts`, `engine.ts`, `attributes-view.svelte.ts` already optional-chain or guard `!response.data` — no change needed, and they now type-check correctly.
  - `fetchSocketData` previously dereferenced `response.data` after only checking `error`; the two fields aren't coupled in the type, so it now also surfaces a data-less **non-error** reply (a protocol violation) as a throw rather than returning `undefined` as if it were valid data.
  - `+page.svelte` and `session.ts` `.data` derefs are on **HTTP** `ApiRequest` responses, not the socket type — unaffected.

No consumer code changes were required; this is purely the type tightening plus the `fetchSocketData` completeness guard. This is squarely the CLAUDE.md guideline (avoid the null-forgiving operator / unsafe casts; handle nullability properly).

## Test plan

- [x] `npm run lint` (eslint + `svelte-check`) — 0 errors / 0 warnings across 981 files
- [x] `npm run test:unit:ci` — **2183 passed** (214 files)
- [x] New test: `fetchSocketData` throws when a non-error response carries no data (covers the new completeness guard). Existing `api-socket-request` tests already pin `response.data === undefined` on the error path.

No battle logic touched, so no parity changes; nothing is codegen'd here.

Closes #940

---
_Generated by [Claude Code](https://claude.ai/code/session_01AyXCQnQtvz1aKxvCBw6h3y)_

### DIFF
--- a/UI/src/lib/api/api-socket-request.ts
+++ b/UI/src/lib/api/api-socket-request.ts
@@ -38,9 +38,9 @@ export class ApiSocketRequest<T extends ApiSocketCommand | void = void> {
 	 *  drops before the server answered. Mirrors the shape of a server-reported error so callers handle it
 	 *  through the same `response.error` path they already use, rather than a separate reject channel. */
 	public settleWithError(error: string) {
-		// An error response carries no `data` (matching how the server signals errors), so the cast bridges
-		// the type's non-optional `data` field that callers already guard with optional chaining.
-		this.promiseResolver({ id: this.id, name: this.commandName, error } as IApiSocketResponse<T>);
+		// An error response carries no `data`, matching how the server signals errors so callers fall
+		// through the same `response.error` guard rather than reading a partial payload.
+		this.promiseResolver({ id: this.id, name: this.commandName, error });
 	}
 
 	public async getResponse() {
--- a/UI/src/lib/api/api-socket.ts
+++ b/UI/src/lib/api/api-socket.ts
@@ -48,7 +48,9 @@ export interface IApiSocketResponse<T extends ApiSocketCommand | void = void> {
 	id: string;
 	name: T;
 	error?: string;
-	data: T extends ApiSocketCommand ? ApiSocketResponseTypes[T] : never;
+	// Optional because every failure (server error, dropped connection, timeout) resolves with an
+	// `error` and no `data`; the type then forces callers to guard `error`/optional-chain `data`.
+	data?: T extends ApiSocketCommand ? ApiSocketResponseTypes[T] : never;
 }
 
 let socket: WebSocket;
@@ -417,8 +419,10 @@ export async function fetchSocketData<T extends ApiSocketCommandNoRequest>(
 	command: T
 ): Promise<ApiSocketResponseTypes[T]> {
 	const response = await apiSocket.sendSocketCommand(command);
-	if (response.error) {
-		throw new Error(response.error);
+	// A failure resolves with an `error` and no `data`; a successful reply always carries data, so a
+	// missing payload on a non-error response is itself a protocol error worth surfacing as a throw.
+	if (response.error || response.data === undefined) {
+		throw new Error(response.error ?? 'The server returned no data.');
 	}
 	return response.data;
 }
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: Make `IApiSocketResponse.data` optional so the error-path guard is type-enforced (#940)

**Verdict:** Approve — clean, minimal type tightening that does exactly what the issue asked, with no findings. _(Posted as a comment because I'm the PR author; the intended verdict is **Approve**.)_

The change makes `IApiSocketResponse.data` optional, which is squarely correct given the documented socket lifecycle (`frontend.md` → _Socket request lifecycle_): every failure resolves with an `error` and no `data`. The type now enforces the guard the docs describe, the `as IApiSocketResponse` cast in `settleWithError` is gone (aligning with CLAUDE.md's "avoid unsafe casts / handle nullability properly"), and no consumer code needed changing.

I independently re-ran the consumer audit and it holds:
- `enemy-manager.ts` optional-chains (`result.data?.enemyInstance`); `engine.ts` optional-chains and guards `!data`; `attributes-view.svelte.ts` guards `response.error || !response.data`.
- `inventory-manager.ts` mutations only inspect `response.error` and never dereference `.data` — unaffected.
- `session.ts:47` and `+page.svelte` derefs are on HTTP `ApiRequest` responses, not the socket type — correctly excluded.

Nice catch on the `fetchSocketData` completeness guard: it isn't merely defensive — once `data` is optional, the early throw on `response.data === undefined` is what narrows the return to `T` instead of `T | undefined`, so it's load-bearing for the type as well as surfacing a data-less non-error reply (a protocol violation) as a throw rather than silently returning `undefined`.

### Test coverage
Adequate. New unit test covers the non-error-without-data protocol case (`fetchSocketData` throws "The server returned no data."), and the existing `api-socket-request` tests already pin `response.data === undefined` on the error path. No battle logic touched, so no parity mirroring required.

### What's good
- Tight, well-scoped fix with accurate, non-redundant comments explaining *why* `data` is optional.
- The `fetchSocketData` guard turns a latent "undefined-as-valid-data" path into an explicit throw.
- Thorough consumer audit in the description that matches reality.


---

PR #1006, opened 2026-06-20
Split non-test verifications into separate CI jobs (#943)

### LINKED ISSUE #943: Split out new CI job for non-test verifications
Split out a new CI job for running verification steps for frontend/backend.

- Verify typescript codegen
- Run dotnet formatter
- Run lint (ESLint + svelte check)

This is mostly to speed up the test-backend job, which usually runs the longest.

ginderjeremiah:
## Summary

Splits the non-test verification steps out of the heavy test jobs into their own jobs, as requested in #943, so the test jobs run leaner and the verifications run in parallel.

- **New `verify-backend` job** — runs the TypeScript codegen-drift check and `dotnet format --verify-no-changes`, moved out of `test-backend`. Both checks need the build, so this job does its own `dotnet build` and runs concurrently with `test-backend` rather than chaining off it.
- **New `verify-frontend` job** — runs `npm run lint` (ESLint + `svelte-check`, since the repo's `lint` script is `eslint . && npm run check`), moved out of `test-frontend`.
- **`test-backend`** is left with just build + tests/coverage — its longest path — shortening its wall-clock by the codegen + format time.
- **`test-frontend`** is left with just `npm ci` + the unit-test/coverage run.
- **`all-checks-passed`** now lists the two new jobs in `needs`, so they stay blocking for branch protection.

Both new jobs reuse the existing `detect-changes` gating (backend vs. frontend) and the same toolchain/cache setup as their test counterparts, so a frontend-only PR still won't spin up .NET and vice-versa.

## How it addresses the issue

#943 asks to split a CI job for the non-test verifications (codegen, dotnet formatter, ESLint + svelte check) "mostly to speed up the test-backend job, which usually runs the longest." Moving codegen + format off the `test-backend` critical path is the main win; they now run in parallel in `verify-backend` instead of in series before the tests.

## Design note for reviewer

The issue says "a new CI job" (singular). I implemented it as **two parallel jobs** (`verify-backend` / `verify-frontend`) rather than one combined job, because:
- It serves the stated goal (wall-clock speed) better — a single combined job would serialize the backend build and the frontend lint, whereas two jobs run them concurrently.
- It mirrors the existing `test-backend` / `test-frontend` split and reuses the per-area `detect-changes` gating cleanly (no per-step conditionals), so each side stays independently skippable.

Happy to consolidate into a single job if you'd prefer that — it's a small change either way.

There is a deliberate trade-off: `verify-backend` builds the solution again (in parallel with `test-backend`'s build) since codegen + format require the binaries. This costs an extra parallel build of runner minutes in exchange for the shorter critical path, consistent with the issue's "speed up test-backend" intent and the repo's existing preference for simple parallelism over build-artifact sharing (see the `test-e2e` notes in `docs/infrastructure.md`).

## Test plan

- [x] Workflow YAML parses and the job graph is correct (`detect-changes` → `verify-backend`/`verify-frontend`/`test-backend`/`test-frontend`/`test-e2e` → `all-checks-passed`).
- [x] Every moved step uses the exact command from the previously-passing workflow; no command text changed.
- [ ] CI on this PR exercises all jobs end-to-end (the workflow change touches `.github/workflows/**`, which both change-detection filters include, so backend, frontend, and e2e all run).

Closes #943

---
_Generated by [Claude Code](https://claude.ai/code/session_016CFH93e4bwuQ63XrNfL6jc)_

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -62,7 +62,10 @@ jobs:
             - 'UI/**'
             - '.github/workflows/**'
 
-  test-backend:
+  # Non-test backend verifications (codegen drift + formatter), split out of test-backend so that
+  # job is left with just build + tests/coverage (its longest path). Both checks need the build, so
+  # this job builds independently and runs in parallel with test-backend rather than chaining off it.
+  verify-backend:
     needs: [detect-changes]
     # always()/!cancelled() lets this evaluate even when detect-changes was skipped (push events).
     if: |
@@ -73,8 +76,6 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v6
-      with:
-        fetch-depth: 0
 
     - name: Install .NET Core
       uses: actions/setup-dotnet@v5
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Split non-test verifications into separate CI jobs (#943)

**Verdict:** Approve (intended) — submitted as a comment because I'm the PR author, so GitHub won't accept a gating review on my own PR. One should-fix (docs), no blockers.

The change cleanly does what #943 asks: codegen-drift + `dotnet format` move out of `test-backend` into a new `verify-backend`, ESLint/`svelte-check` move out of `test-frontend` into `verify-frontend`, and both run in parallel so the test jobs are left on their longest path. I verified nothing was dropped in the move — every step from the old combined jobs is preserved on one side or the other — and the live CI run on this PR confirms the graph: `verify-backend` and `verify-frontend` both ran and passed, and `all-checks-passed` lists them in `needs` so they stay blocking for branch protection.

The two-jobs-instead-of-one and the redundant `verify-backend` build are reasonable, well-justified calls: parallelism serves the stated wall-clock goal better than one serialized job, and the extra build trades free CI minutes for a shorter critical path — the same trade-off `test-e2e` already makes. Dropping `fetch-depth: 0` from `verify-backend` is a correct optimization, since the codegen-drift `git diff` and the formatter both operate on the working tree and need no history (`test-backend` keeps it, unchanged).

### Should-fix

- `docs/infrastructure.md` isn't updated to reflect the new jobs. The redundant-build trade-off is exactly the kind of "why" the docs capture — the parallel `test-e2e` build is already documented there for the same reason, so for consistency (and per CLAUDE.md's "document important design decisions") a short note on the `verify-backend`/`verify-frontend` split and its deliberate extra parallel build would be worth adding. Not a blocker; happy for it to be a quick follow-up commit or a follow-up issue.

### Test coverage

- N/A — a GitHub Actions workflow can't be meaningfully unit-tested, and the PR exercises itself: this run drove all jobs (backend, frontend, e2e) end-to-end. No test gap.

### What's good

- Faithful 1:1 move — no command text changed, so behavior is identical and only the job topology shifts.
- `detect-changes` gating reused per-area, so a frontend-only PR still won't spin up .NET and vice-versa; the `push`-event `always() && !cancelled()` escape hatch is preserved on both new jobs.
- Comments on the new jobs match the file's existing style and explain the non-obvious bits (independent build, parallelism).


---

ginderjeremiah:
Addressed the should-fix in 4a5ada5 — added a short "CI job split" section to `docs/infrastructure.md` covering the `verify-backend`/`verify-frontend` split and the deliberate extra parallel build, mirroring the existing `test-e2e` build trade-off note.

Heads-up: that commit only touches `docs/`, which doesn't match the backend/frontend `detect-changes` path filters, so the verify/test jobs will skip on this push (skipped → green for `all-checks-passed`). The workflow itself was already exercised end-to-end on the previous commit.

---

Issue #910, opened 2026-06-18
X-Forwarded-For is trusted unconditionally for login-tracking IP (spoofable audit data)


ginderjeremiah:
## Problem

`LoginTrackingMiddleware.ResolveIpAddress` returns the first `X-Forwarded-For` entry whenever the header is present, with no trusted-proxy check:

```csharp
// Game.Api/Middleware/LoginTrackingMiddleware.cs:51-60
```

Any client can set `X-Forwarded-For` directly, so the recorded "connection IP" — a security/audit surface (login history, device tracking) — is fully attacker-controlled when the app is reachable without a proxy or behind a misconfigured one. An attacker can forge the IP attributed to their logins.

## Suggested fix

Use ASP.NET Core's `UseForwardedHeaders` with a configured `KnownProxies` / `KnownNetworks` allowlist and read `HttpContext.Connection.RemoteIpAddress`, rather than hand-parsing the header. Only honor `X-Forwarded-For` from known proxies; fall back to the socket peer address otherwise.

_Found during a codebase health review._

PR #1009, opened 2026-06-20
Add per-IP rate limiting to anonymous auth endpoints (#950)

### LINKED ISSUE #950: Add rate limiting / brute-force protection to the anonymous auth endpoints
## Problem

The anonymous auth endpoints have **no rate limiting or brute-force protection**. There is no `AddRateLimiter`/`UseRateLimiter` anywhere in the project (confirmed by grep across `Game.Api`).

- `Game.Api/Controllers/LoginController.cs:31` (`Login`), `:52` (`Refresh`), `:62` (`CreateAccount`) are all `[AllowAnonymous]` and unmetered.
- An attacker can brute-force passwords, spray refresh tokens, or flood the user table via `CreateAccount` at full request rate.
- Notably, PBKDF2 at 600k iterations (`Pbkdf2PasswordHasher`) *amplifies* the login DoS surface — every guess costs the server real CPU, so an unmetered `Login` is both a credential-stuffing and a resource-exhaustion vector.
- The single-use, 48h refresh token is the highest-value brute-force target, and `Refresh` has no edge throttle.

## Suggested fix

- Wire ASP.NET Core's built-in rate limiter (`AddRateLimiter` + `UseRateLimiter`) with a partitioned fixed/sliding window keyed by client IP — resolved the same trusted-proxy way the login-IP tracking already does (see commit 30782ba, "Trust X-Forwarded-For only from configured proxies").
- Apply tighter limits on `Login`/`Refresh` than on the rest of the surface.
- Consider an account-level lockout/backoff after N consecutive failed logins.

## Notes

The limits should be configuration-bound (deployment-specific), consistent with how CORS origins are handled. This is a security-hardening gap, not a defect in existing behavior.

ginderjeremiah:
## Summary

The anonymous auth endpoints — `Login`, `Refresh`, `CreateAccount` (all `[AllowAnonymous]` on `LoginController`) — were completely unmetered (`#950`). That left three live abuse vectors open: credential stuffing against `Login`, refresh-token brute force against `Refresh`, and — because PBKDF2 runs at 600k iterations — a CPU resource-exhaustion DoS where every login/create guess costs the server real work.

This wires ASP.NET Core's built-in rate limiter (`AddRateLimiter` / `UseRateLimiter`) with a **per-client-IP fixed-window** policy applied to those three endpoints.

### How it works

- **One shared `"auth"` policy** across the three endpoints, so the limit is a **combined per-IP budget** an attacker can't multiply by spreading across endpoints. Applied via `[EnableRateLimiting(RateLimitingOptions.AuthPolicy)]`.
- **Partition key = the trusted-proxy-correct client IP.** I extracted the IP resolution that already lived in `LoginTrackingMiddleware` into a shared `ClientIp.Resolve` helper and use it for both, so a spoofed `X-Forwarded-For` from a direct client can't shard an attacker across partitions (consistent with #910 — only a configured trusted proxy's forwarded header is honoured).
- **429 returns the project's standard `{ errorMessage }` envelope** (via `OnRejected`) plus a `Retry-After` header, instead of the framework's empty 429 body — matching every other failure response shape the client already handles.
- **Config-bound with safe defaults.** `RateLimiting:Auth:{PermitLimit,WindowSeconds}` defaults to 10 requests / 60s. Unlike the CORS origins (which fail fast when unset), rate limiting carries **safe positive defaults** so an unconfigured deployment is *still protected* — config only tunes the numbers — and the limits are validated as positive on start so an accidental `0` fails fast rather than silently disabling the guard.

### Pipeline placement

`UseRateLimiter` runs **after** `UseCors` (so a throttled browser request still carries the CORS headers it needs to read the 429 body) and **after** the forwarded-headers middleware (so the partition keys off the corrected client IP).

## Scope note / decision

The issue also suggested *considering* an account-level lockout after N consecutive failed logins. I deliberately scoped that **out** of this PR and opened follow-up **#1010**: account-level lockout carries its own lockout-DoS tradeoff (an attacker can lock a victim out by spamming failed logins for their username) and needs a design decision, whereas per-IP limiting is the safe, unambiguous first line that directly closes the stated gap. Happy to fold it in here instead if preferred.

## Test plan

- [x] `dotnet build Game.Api` — clean (0 warnings)
- [x] **New** `Game.Api.Tests/Unit/RateLimitingOptionsTests.cs` — defaults, partial-bind, and the validate-on-start path (exercises the real `AddAuthRateLimiter` registration, not a re-declaration).
- [x] **New** `Game.Api.Tests/Integration/AuthRateLimitingTests.cs` — throttle-after-limit returns 429 + `Retry-After` + envelope; the shared per-IP budget spans `Login`→`CreateAccount`; a single request under the limit is untouched.
- [x] Retargeted the IP-resolution unit test to the new shared `ClientIp.Resolve` (renamed `LoginTrackingIpResolutionTests` → `ClientIpResolutionTests`); no behavior change, no dead code left behind.
- [x] `Game.Api.Tests` — **544 passed** (was 537; the shared test factory pins a high limit so existing suites aren't throttled, and the dedicated rate-limit tests override it with a small limit).

Docs updated: new "Auth endpoint rate limiting" subsection under Authentication in `docs/backend.md`.

Closes #950

### DIFF
--- a/Game.Api.Tests/Integration/AuthRateLimitingTests.cs
+++ b/Game.Api.Tests/Integration/AuthRateLimitingTests.cs
@@ -1,0 +1,98 @@
+using Game.Api.Models.Common;
+using Game.TestInfrastructure.Base;
+using Game.TestInfrastructure.Fixtures;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using System.Net;
+using System.Net.Http.Json;
+using Xunit;
+
+namespace Game.Api.Tests.Integration
+{
+    /// <summary>
+    /// Covers the per-client-IP auth rate limiter (#950): the anonymous auth endpoints are throttled with
+    /// the project's standard error envelope once the configured limit is exceeded, the limit is a shared
+    /// budget across those endpoints, and a request under the limit is unaffected. The factory pins the
+    /// limit to a small value so the throttle can be exercised deterministically (the in-memory TestServer
+    /// has no socket peer, so every request shares the "unknown" partition).
+    /// </summary>
+    [Collection("Integration")]
+    public class AuthRateLimitingTests : ApiIntegrationTestBase
+    {
+        private const int PermitLimit = 3;
+
+        public AuthRateLimitingTests(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper)
+            : base(containers, testOutputHelper) { }
+
+        protected override GameServerFactory CreateFactory(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper)
+        {
+            return new RateLimitedFactory(containers, testOutputHelper);
+        }
+
+        [Fact]
+        public async Task Login_BeyondPermitLimit_IsThrottledWithEnvelope()
+        {
+            var creds = new { Username = "nobody", Password = "wrong" };
+
+            // The first PermitLimit attempts run the endpoint (rejected as bad credentials, not throttled).
+            for (var i = 0; i < PermitLimit; i++)
+            {
+                var allowed = await Client.PostAsJsonAsync("/api/Login", creds, CancellationToken);
+                Assert.NotEqual(HttpStatusCode.TooManyRequests, allowed.StatusCode);
+            }
+
+            // The next attempt is throttled before the endpoint runs.
+            var throttled = await Client.PostAsJsonAsync("/api/Login", creds, CancellationToken);
+            Assert.Equal(HttpStatusCode.TooManyRequests, throttled.StatusCode);
+            Assert.True(throttled.Headers.Contains("Retry-After"));
+
+            var body = await throttled.Content.ReadFromJsonAsync<ApiResponse>(CancellationToken);
+            Assert.NotNull(body);
+            Assert.False(string.IsNullOrEmpty(body.ErrorMessage));
+        }
+
+        [Fact]
+        public async Task AuthEndpoints_ShareOnePerIpBudget()
+        {
+            var creds = new { Username = "nobody", Password = "wrong" };
+
+            // Exhaust the budget via Login.
+            for (var i = 0; i < PermitLimit; i++)
+            {
+                await Client.PostAsJsonAsync("/api/Login", creds, CancellationToken);
+            }
+
+            // A sibling auth endpoint draws from the same per-IP partition, so it is already throttled.
+            var createAccount = await Client.PostAsJsonAsync("/api/Login/CreateAccount", creds, CancellationToken);
+            Assert.Equal(HttpStatusCode.TooManyRequests, createAccount.StatusCode);
+        }
+
+        [Fact]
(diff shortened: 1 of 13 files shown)


ginderjeremiah (commented):
## Review: Per-IP rate limiting for the anonymous auth endpoints (#950)

**Verdict:** Approve — closes the stated abuse vectors (credential stuffing, refresh brute-force, PBKDF2 resource exhaustion) cleanly, with the right layering, config story, and test coverage. (Submitted as a `COMMENT` only because GitHub treats me as the PR author and won't accept a gating self-approval; the intent is an approval.)

### What's good
- **Correct layering.** This is purely an API-layer concern (HTTP throttling, IP partitioning, 429 envelope) and it stays in `Game.Api` — no domain logic leaks into `Game.Core`.
- **DRY IP resolution.** Extracting `ResolveIpAddress` into a shared `ClientIp.Resolve` (`Game.Api/Http/ClientIp.cs`) and pointing both `LoginTrackingMiddleware` and the limiter partition at it is the right move — the trusted-proxy semantics from #910 stay in one place, and the renamed `ClientIpResolutionTests` keep covering them with no dead code left behind.
- **Pipeline placement is reasoned and correct.** `UseRateLimiter` after `UseCors` (so a throttled browser can read the 429 body) and after forwarded-headers (so the partition keys off the corrected client IP) — and the comments at `Startup.cs:198-200` capture the *why*, not just the *what*.
- **Safe-by-default config.** Positive defaults (10/60s) that protect an unconfigured deployment, validated `> 0` on start so an accidental `0` fails fast rather than silently disabling the guard — a deliberate and well-documented contrast with the fail-fast CORS origins.
- **Standard error envelope.** `OnRejected` returns the project's `{ errorMessage }` shape + `Retry-After` (invariant-culture formatted) instead of the framework's empty 429 — consistent with every other failure the client already handles.
- **Scope discipline.** Account-level lockout correctly deferred to #1010 with the lockout-DoS tradeoff called out, rather than bundled in.

### Test coverage
Sufficient and aligned with the project's rules. Unit tests cover the options defaults, partial bind, and both validation paths (exercising the real `AddAuthRateLimiter` registration, not a re-declaration); integration tests cover throttle-returns-429-with-envelope-and-Retry-After, the shared per-IP budget spanning `Login`→`CreateAccount`, and an under-limit request passing through. The `TestServer`-has-no-socket-peer insight that makes the throttle deterministic is a nice touch, and pinning a high limit suite-wide while the dedicated tests override it keeps the rest of the suite unthrottled. No frontend/battle-logic surface, so no mirrored tests needed.

### Notes (non-blocking, no action required)
- IP-based limiting inevitably groups callers behind CGNAT/corporate NAT into one budget; 10/60s across the three endpoints is conservative enough that this is a fine first line, and the issue explicitly asked for per-IP. Worth keeping in mind if false-positive throttling is ever reported.
- Defaults are now stated in both `RateLimitingOptions` and `appsettings.json`; the latter usefully documents the knob, so this is acceptable duplication rather than something to fix.

Clean work — approving.

---

PR #1011, opened 2026-06-20
Harden socket handler layer (#953)

### LINKED ISSUE #953: Socket handler robustness hardening (read-loop spin guard, watchdog memory visibility, handshake error envelope, server-push notice logging)
## Problem

A cluster of small robustness gaps in the WebSocket handler layer. None is a hot-path bug today, but each is a latent reliability/observability hole in an area the rest of the code treats with care.

### 1. `ReadLoop` can hot-spin on a persistent non-cancellation exception
`Game.Api/Sockets/SocketHandler.cs:282-319`

The catch-all (`:305`) logs (only when `State is Open`) and the `do…while (_context.State is Open)` loops again. Most receive failures abort the socket (state leaves `Open`, loop exits), but an exception that does **not** transition `WebSocketState` (e.g. an OOM/decoding error while assembling the message) leaves state `Open` and the loop becomes a tight CPU spin that also floods the log — and the inactivity watchdog can't help because `_lastResponse` is only updated on a *successful* read.

**Fix:** treat a thrown receive as terminal — after a non-drain exception, break and close (or add a small consecutive-failure counter that closes after a few attempts).

### 2. `_lastResponse` is shared across loops without a memory-visibility guarantee
`Game.Api/Sockets/SocketHandler.cs:49` (field), written at `:292` (read loop), read at `:264` (inactivity loop)

`private DateTime _lastResponse` is written by the read-loop task and read by the inactivity-loop task with no `volatile`/`Interlocked`/lock. `DateTime` is a 16-byte struct so the read isn't even guaranteed atomic (torn read possible), and there's no happens-before edge. The class is otherwise meticulous about concurrency (two documented semaphores), so this is an inconsistency.

**Fix:** store the timestamp as `long` ticks via `Interlocked`/`Volatile`, or guard with a lock.

### 3. The socket-handshake failure path returns a bare 500 with no body
`Game.Api/Middleware/SocketInterceptorMiddleware.cs:27-32`

When `TryLoadPlayer` fails, the middleware sets `StatusCode = 500` with no body — inconsistent with the project's contract that every failure returns the `{ errorMessage }` envelope. It's also arguably the wrong status (authenticated-but-unloadable is closer to 404/409 than 500). The socket isn't upgraded yet, so a body can still be written (or it can throw a typed exception and let `ExceptionHandlingMiddleware` shape it).

### 4. A failed server-push re-sync notice is silent
`Game.Api/Services/SocketManagerService.cs:212-234`

`EscalateFailedServerCommand` dead-letters the faulted push and dispatches a `ServerCommandFailed` notice (`:230`), but the notice dispatch's own outcome is ignored. If the notice *also* faults (e.g. the socket just closed), there's no log — so for `ChallengeCompleted` (which gates zone navigation) a player whose re-sync cue was lost diverges with no operator signal.

**Fix:** capture and warn on the notice-dispatch outcome ("client may not have received re-sync cue for {Command}").

### 5. DRY: the error-frame send is hand-rolled three times
`Game.Api/Sockets/SocketHandler.cs:119-124`, `:186-191`, `:344-349`

The same `new ApiSocketResponse { Id, Name, Error = … }` + `SendData` triple appears three times with only the message differing. Extract a `SendErrorAsync(SocketCommandInfo info, string error)` helper (also simplifies the #940 `Data`-optional change).

ginderjeremiah:
## Summary

Closes #953 — a cluster of small robustness/observability gaps in the WebSocket handler layer. None was a hot-path bug, but each was a latent reliability hole in an area the rest of the code treats with care.

### 1. `ReadLoop` no longer hot-spins on a persistent non-cancellation receive failure
`Game.Api/Sockets/SocketHandler.cs`

A thrown receive is now treated as **terminal**: after a non-drain exception the loop breaks and the post-loop close fires. Most receive failures abort the socket (state leaves `Open`, the `do…while` would exit anyway), but a failure that does **not** transition `WebSocketState` (e.g. an OOM/decoding error while assembling the message) previously left the loop a tight CPU spin that also flooded the log — and the watchdog couldn't help because the activity timestamp only advances on a *successful* read. The post-loop close now covers both `CloseReceived` (complete the client handshake) and a still-`Open` terminal break (close so teardown completes); `Close` is a no-op on an already-closed/aborted socket.

### 2. The inactivity-watchdog timestamp now has a memory-visibility guarantee
The `DateTime _lastResponse` field — written by the read loop, read by the inactivity loop on a separate thread with no synchronization — is replaced by a `long _lastResponseTicks` accessed via `Interlocked.Exchange`/`Interlocked.Read`. This gives the cross-thread happens-before edge and avoids the torn read possible on the 16-byte `DateTime` struct, matching the class's otherwise-meticulous concurrency discipline.

### 3. The socket-handshake failure path returns the `{ errorMessage }` envelope
`Game.Api/Middleware/SocketInterceptorMiddleware.cs`

When an authenticated session's player can't be loaded, the middleware previously set a bare `500` with no body. The socket isn't upgraded yet, so a body can still be written: it now returns the project's `ApiResponse.Error` envelope with a **404** (the player resource is absent — a genuine load *error* throws and is shaped by `ExceptionHandlingMiddleware` into a 500 as before).

### 4. A failed server-push re-sync notice is no longer silent
`Game.Api/Services/SocketManagerService.cs`

`EscalateFailedServerCommand` dispatches a `ServerCommandFailed` re-sync notice but ignored its outcome. The dispatch outcome is now captured and, when it's not `Succeeded`, warned on ("the client may not have received the re-sync cue for {Command}") — so a player whose gating `ChallengeCompleted` cue was lost (e.g. the socket closed mid-dispatch) leaves an operator signal instead of diverging silently.

### 5. DRY: the error-frame send is consolidated
The hand-rolled `new ApiSocketResponse { Id, Name, Error = … }` + `SendData` triple (which appeared across the fault, timeout, malformed-frame, and server-initiated-rejection paths) is extracted into a single `SendErrorAsync(SocketCommandInfo, string)` helper, so every failure path emits a consistent envelope.

## How it addresses the issue

Each of the five listed gaps is closed. Items 1, 2, and 5 are in `SocketHandler`; item 3 in the handshake middleware; item 4 in the pub/sub escalation path.

## Test plan

- **New** `Game.Api.Tests/Unit/SocketReadLoopTests.cs` — asserts the read loop reads exactly once and closes on a terminal receive fault (no spin), and that a client close frame still completes the handshake without an error.
- **New** test in `SocketCommandProcessorTests` — asserts a failed re-sync notice logs the "may not have received the re-sync cue" warning.
- The error-frame consolidation (#5) is covered by the existing fault/timeout/malformed-frame assertions, which continue to pass unchanged.
- `Game.Api.Tests` — **547 passed**; full solution builds with **0 warnings**.

## Notes for review

- **Item 3 scope:** the issue flagged the `500` path. The sibling `401` (unauthenticated) and `400` (non-WebSocket request) branches in the same middleware also return no body; I left those as-is to keep the change scoped, but they're a candidate for the same envelope treatment if desired (happy to fold in or file a follow-up).
- **No dedicated automated test for item 3:** the handshake "authenticated but player-not-loadable" branch is hard to reach deterministically in an integration test (it needs an authenticated session whose `SelectedPlayerId` resolves to a non-existent player), and per `docs/backend.md` the `Game.Api` middleware adapters are measure-only for coverage. The change is a small, self-contained response shaping. Flagging it here per the pick-up-issue guidance rather than leaving it unmentioned.

Closes #953

---
_Generated by [Claude Code](https://claude.ai/code/session_012CfAe2rvKViDnmEMfkwQT1)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
@@ -83,6 +83,25 @@ namespace Game.Api.Tests.Unit
         }
 
         [Fact]
+        public async Task Processor_ResyncNoticeAlsoFails_WarnsThatTheClientMayHaveMissedTheCue()
+        {
+            // The original push faults and is escalated; if the ServerCommandFailed re-sync notice itself does
+            // not get through (here it also faults), the client never receives the re-sync cue for a gating
+            // command and would diverge silently — so the lost cue is surfaced at warning (#953), not ignored.
+            var commands = new CapturingCommandFactory(throwOn: name =>
+                name is "ChallengeCompleted" or "ServerCommandFailed" ? new InvalidOperationException("boom") : null);
+            var (processor, _) = BuildProcessor(commands);
+
+            var queue = new FakeQueue(new SocketCommandInfo("ChallengeCompleted") { Id = "push-1" });
+
+            await processor(queue);
+
+            Assert.Contains(_logs.Entries, e => e.Level == LogLevel.Warning
+                && e.Message.Contains("may not have received the re-sync cue")
+                && e.Message.Contains("ChallengeCompleted"));
+        }
+
+        [Fact]
         public async Task Processor_DequeueFaultThenValidCommand_QueueKeepsDrainingAndProcessorDoesNotThrow()
         {
             var commands = new CapturingCommandFactory(throwOn: _ => null);
--- a/Game.Api.Tests/Unit/SocketReadLoopTests.cs
+++ b/Game.Api.Tests/Unit/SocketReadLoopTests.cs
@@ -1,0 +1,182 @@
+using Game.Abstractions.DataAccess;
+using Game.Api.Services;
+using Game.Api.Sockets;
+using Game.Api.Sockets.Commands;
+using Game.Application;
+using Game.Core.Players;
+using Game.TestInfrastructure.Helpers;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System.Net.WebSockets;
+using Xunit;
+
+namespace Game.Api.Tests.Unit
+{
+    /// <summary>
+    /// Verifies the read-loop's terminal-fault handling (#953): a thrown receive that does not transition the
+    /// WebSocket state must stop the loop and close the socket rather than tight-spinning on the same failure,
+    /// while a client-initiated close frame still completes the closing handshake. Scripted <see cref="WebSocket"/>
+    /// stand-ins keep these as plain unit tests.
+    /// </summary>
+    public sealed class SocketReadLoopTests : IDisposable
+    {
+        private static readonly TimeSpan WaitTimeout = TimeSpan.FromSeconds(5);
+
+        private readonly ServiceProvider _provider;
+        private readonly IServiceScopeFactory _scopeFactory;
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly CapturingLoggerProvider _logs = new();
+
+        public SocketReadLoopTests()
+        {
+            _provider = new ServiceCollection()
+                .AddScoped<IUnitOfWork, NoOpUnitOfWork>()
+                .BuildServiceProvider();
+            _scopeFactory = _provider.GetRequiredService<IServiceScopeFactory>();
+            _loggerFactory = LoggerFactory.Create(b => b.AddProvider(_logs).SetMinimumLevel(LogLevel.Trace));
+        }
+
+        [Fact]
+        public async Task ReadLoop_ReceiveThrowsWhileSocketStaysOpen_StopsAndClosesInsteadOfSpinning()
+        {
+            // A receive failure that does NOT transition WebSocketState (the socket stays Open) would loop the
+            // read loop forever on the same throw — a tight CPU spin that floods the log. The fix treats a
+            // thrown receive as terminal: the loop reads exactly once, then closes the socket so teardown can
+            // complete.
+            var socket = new ScriptedReadWebSocket();
+            socket.QueueThrow(new WebSocketException("decode failure that leaves the socket open"));
+            var (context, handler) = CreateHandler(socket);
+
+            using var inactivityStop = new CancellationTokenSource();
+            handler.Listen(hostStopping: inactivityStop.Token);
+
+            await context.WaitSocketClosed().WaitAsync(WaitTimeout, TestContext.Current.CancellationToken);
+            inactivityStop.Cancel();
+            await handler.Completion.WaitAsync(WaitTimeout, TestContext.Current.CancellationToken);
+
+            // The loop did not spin: it read once, then broke. And it closed the still-open socket gracefully.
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: Harden socket handler layer (#953)

**Verdict: Approve** — all five hardening items are implemented correctly, scoped tightly, and well-tested. (Posted as a `COMMENT` because GitHub won't accept a gating verdict from the PR's own author; the substantive call is approve.)

Each item maps cleanly to the issue and the surrounding code holds up under inspection:

1. **ReadLoop terminal break** (`Game.Api/Sockets/SocketHandler.cs:299-324`) — the catch-all now breaks instead of re-looping, and the post-loop close was widened to `Open or CloseReceived` so a still-`Open` terminal break closes the socket. Verified against `SocketContext.Close`, which only emits a frame from `Open` and is a safe no-op otherwise, so the `CloseReceived` path settles teardown without a redundant frame — exactly what the tests assert.
2. **Watchdog memory visibility** (`SocketHandler.cs:53, :258, :286`) — `long _lastResponseTicks` via `Interlocked.Exchange`/`Read` gives the cross-thread happens-before edge and dodges the torn 16-byte `DateTime` read. Comparison rewritten correctly in ticks.
3. **Handshake envelope** (`SocketInterceptorMiddleware.cs:35-36`) — `404` + `ApiResponse.Error` is the right shape; the body is writable pre-upgrade, and a genuine load fault still throws into `ExceptionHandlingMiddleware`. The `404` (resource-absent) vs `500` distinction is sound.
4. **Re-sync notice outcome** (`SocketManagerService.cs:274-278`) — the dispatch outcome is captured and warned on when not `Succeeded`, and the existing `ServerCommandFailed`-name guard still prevents an escalation loop.
5. **`SendErrorAsync` helper** (`SocketHandler.cs:378-386`) — consolidates the four hand-rolled error frames into one consistent envelope. Clean DRY win.

### Test coverage

Strong. `SocketReadLoopTests` drives both the terminal-fault (reads once, closes, logs, no spin) and clean client-close (handshake completes, no error) paths with a deterministic scripted `WebSocket` — a nice classical-style unit test with no unmanaged dependency. The new `SocketCommandProcessorTests` case asserts the lost-cue warning. Items 1, 2, 4, 5 are well covered; item 3's absence of an automated test is reasonably justified (hard-to-reach branch, and `Game.Api` middleware adapters are measure-only per `docs/backend.md`).

### Follow-up (non-blocking)

The sibling `401`/`400` short-circuits in the same middleware (`SocketInterceptorMiddleware.cs:22, :26`) still return a bare status with no body, which is inconsistent with the same envelope contract item #3 restores. The author flagged this and scoped it out deliberately — fair call. Tracked in a follow-up issue so it isn't lost; not a blocker for this PR.

### What's good

The comments explaining *why* each change matters (the non-state-transitioning receive failure, the torn-read rationale, the `Open or CloseReceived` widening) are exactly the kind of context that keeps a later "simplification" from quietly reintroducing the spin. Tight, well-reasoned change.

Will merge once CI goes green.

---

Issue #558, opened 2026-06-14
Plumb CancellationToken through the data-access layer for cooperative socket-command cancellation


ginderjeremiah:
## Context

Follow-up to #483 (PR #557). That work added a per-command timeout to `SocketHandler.ExecuteCommand`: each command runs under a `CancellationTokenSource` budget, the token is threaded through `AbstractSocketCommand.ExecuteAsync`, and a command that ignores the token is bounded by a `Task.WaitAsync` backstop (the abandoned task keeps running until it settles, holding the command lock so the next command can't race it).

The token is currently threaded only to the **command boundary**. It is **not** plumbed below that into the data-access layer — `ICacheService` and `IPubSubService` take no `CancellationToken`, and neither do `IUnitOfWork.CommitAsync`, the repositories, or the application services (`PlayerService`, `BattleService`). So a command genuinely wedged inside a token-ignoring dependency only unwinds via that dependency's own timeout (Redis 5s sync timeout, Npgsql command timeout); the per-command budget's cooperative-cancellation half is currently a no-op for those paths.

## Why it matters

With cooperative cancellation reaching the actual I/O, a wedged command unwinds *promptly* on the budget instead of being abandoned (the `Task.WaitAsync` backstop, which holds the command lock until the stuck dependency clears on its own). That makes the timeout fully effective and removes the residual "later commands queue behind a stuck dependency until it clears" trade-off documented in `docs/backend.md`.

## Suggested direction

- Add an optional `CancellationToken` (defaulting to `default`/`CancellationToken.None`) to the async members of `ICacheService` and `IPubSubService`, and forward it to the StackExchange.Redis calls (note SE.Redis honours tokens only partially).
- Add a token to `IUnitOfWork.CommitAsync` and forward it to EF Core's `SaveChangesAsync(ct)` (Npgsql honours it).
- Thread the token from `AbstractSocketCommand.ExecuteAsync` down through the application services and repositories the player-mutating commands use (`LoadPlayer`, `EndBattleVictory`/`EndBattleLoss`, `TryUpdateAttributes`, etc.).
- Keep the `Task.WaitAsync` backstop in `ExecuteCommand` for any path that still can't observe the token.

This is a broad, cross-cutting signature change touching most async members in the data tier, hence `scope: large`.

PR #1013, opened 2026-06-20
Redis infrastructure hygiene & robustness cleanups (#954)

### LINKED ISSUE #954: Redis infrastructure hygiene & robustness cleanups
## Problem

A set of small robustness/hygiene items in the Redis cache + pub/sub infrastructure. Independent, mostly self-contained.

### 1. Worker-backed `Subscribe` with `id == null` leaks the `BackgroundWorker`
`Game.Infrastructure/PubSub/Redis/RedisPubSubService.cs:155-181` (`SubscribeWithWorker`)

When `id` is null the guard `if (id is not null && !_handles.TryAdd(...))` is skipped, so on success the created `BackgroundWorker` (with its `RegisteredWaitHandle` + `AutoResetEvent`) is **never stored in `_handles`** — it can't be `Dispose()`d and can't be targeted by `UnSubscribe(channel, id)`. All current callers pass a non-null id, so this is latent, but the API silently leaks OS handles for an id-less worker subscription.

**Fix:** require a non-null id for the worker-backed overloads (or track id-less workers separately so they can be torn down).

### 2. `GetSet(key, value, expiry)` Lua path crashes on a null value
`Game.Infrastructure/Cache/Redis/RedisService.cs:71-80`

The inline Lua does `redis.call('set', KEYS[1], ARGV[1])`; a null `value` makes the server-side SET error, whereas the non-script setters treat null as a delete. The impl parameter is `string?` while `ICacheService` declares the value non-null, so the null contract is unenforced and the two paths disagree.

**Fix:** align interface/impl nullability and pick one null semantic (guard+throw, or branch to delete to match the other setters).

### 3. `IDatabase`/`ISubscriber` resolved per call on the hot path
`RedisService.cs:12` (`Redis => Multiplexer.GetDatabase()`); `RedisPubSubService.cs:21-22`

The multiplexer is fixed for the instance lifetime; cache the `IDatabase`/`ISubscriber` once in the ctor instead of calling `GetDatabase()`/`GetSubscriber()` on every cache get/set.

### 4. `RedisQueue.ReclaimProcessingAsync` is O(N) sequential round-trips on startup recovery
`Game.Infrastructure/PubSub/Redis/RedisQueue.cs:80-96`

One `ListMoveAsync` per orphaned item, serially, on the crash-recovery path. A single Lua `rpoplpush` loop reclaims everything in one round-trip.

### 5. No graceful shutdown disposal of the connection multiplexers
`Game.Infrastructure/Redis/RedisMultiplexerFactory.cs:15`

Multiplexers live in a static dictionary for process lifetime, disposed only via `ResetForTesting()`. There's no `IHostApplicationLifetime`/`IAsyncDisposable` stop hook, so graceful shutdown never closes the connections cleanly. Depending on the concrete `ConnectionMultiplexer` (vs `IConnectionMultiplexer`) also hampers unit testing.

### 6. `RetryPolicy` permits `BaseDelay == 0` → busy retry loop
`Game.DataAccess/RetryPolicy.cs:24,63-66`

The ctor rejects only a negative base delay; a zero base makes `DelayAfterAttempt` always return `TimeSpan.Zero`, so the message-retry loop spins with no backoff. Forbid `BaseDelay == Zero` (or document immediate-retry semantics) and unit-test the saturation/zero-base paths.

### 7. DRY: `ObserveWrite` cancellation-budget continuation duplicated
`RedisService.cs:167-182` and `RedisPubSubService.cs:82-97` — the same fault-logging continuation (generic + non-generic) is copy-pasted, only the log category differs. Extract a shared internal helper.

ginderjeremiah:
## Summary

Works through the independent Redis cache + pub/sub robustness items in #954. Each was implemented or, where the suggested change conflicts with a deliberate existing decision, addressed the way the issue offered as the alternative — with the reasoning called out below.

### 1. Worker-backed `Subscribe` with a null id leaked the `BackgroundWorker` — fixed
The two queue/worker `Subscribe` overloads now **require a non-null id** (`IPubSubService` + `RedisPubSubService`). A worker registered under no id was never stored in `_handles`, so it could never be `Dispose()`d or targeted by `UnSubscribe(channel, id)` — a leaked OS wait handle. The id is validated **before** the worker is created so the guard path can't leak it either.

The issue stated "all current callers pass a non-null id", but `DataProviderSynchronizer.InitSubscriber` actually subscribed **without** one — exactly the latent leak. It now uses a per-instance `InstanceId` and `UnSubscribe`s it on stop, mirroring `ReferenceCacheSynchronizer`. The plain-action overload (which creates no worker) keeps its optional id.

### 2. `GetSet(key, value, expiry)` Lua path crashed on a null value — fixed
The impl parameter was `string?` while `ICacheService` declares the value non-null, and the inline Lua `SET` would error server-side on a nil `ARGV`. Aligned the impl to the **non-null** contract (`string value`), so a nil can't reach the script. Unlike the other setters there is deliberately no null-means-delete path here — writing a TTL implies writing a value — which is now documented on the method.

### 3. `IDatabase`/`ISubscriber` resolved per call on the hot path — fixed
Both `RedisService` and `RedisPubSubService` now resolve the database/subscriber handle **once in the ctor** instead of calling `GetDatabase()`/`GetSubscriber()` on every cache get/set and publish.

### 4. `ReclaimProcessingAsync` O(N) sequential round-trips — fixed
The crash-recovery reclaim now drains the whole processing list back onto the queue head in **one server-side Lua `LMOVE` loop** (tail→head, preserving original order) rather than one `ListMove` round-trip per orphaned item. It's atomic, so readers never see a half-reclaimed state; the in-flight processing list is bounded so the script stays short. An already-cancelled budget is still honoured up front.

### 5. No graceful shutdown disposal of the multiplexers — fixed
A new `RedisConnectionLifetime` hosted service disposes the shared, process-wide multiplexers on graceful host stop (via `RedisMultiplexerFactory.DisposeAllAsync`, which closes each connection and clears the cache). It's registered ahead of the Redis-consuming hosted services so it **stops last** — connections are torn down only after the write-behind/reference synchronizers and socket registry have drained. The factory and services now depend on **`IConnectionMultiplexer`** rather than the concrete type.

### 6. `RetryPolicy` allows `BaseDelay == 0` — documented, not forbidden
Forbidding a zero base delay would break a **deliberate** existing decision: `Constructor_ZeroBaseDelay_IsAllowed` pins it, and the synchronizer/performance tests pass `TimeSpan.Zero` to drive retries without real waits. The retries are already bounded by `MaxAttempts`, so this is a finite back-to-back retry, not an unbounded busy loop. Took the issue's "or document immediate-retry semantics" path and documented it on `RetryPolicy`; the saturation/zero-base math is already unit-tested.

### 7. DRY: duplicated `ObserveWrite` continuation — already resolved
The copy-pasted fault-logging continuation the issue describes was already extracted into `RedisCommandBudget.ObserveFault` (#558). `RedisService`/`RedisQueue` now keep only thin per-class `ObserveWrite` wrappers that differ solely by their fault message, and `RedisPubSubService` no longer has such a continuation at all. No change needed.

## Tests
- **New** `RedisConnectionLifetimeTests` (stop disposes, start doesn't) and `RedisMultiplexerFactoryTests.DisposeAllAsync_*` (disposes every entry + clears the cache, empty no-op) — both in-process via the generic seam, no live connection.
- **New** null-id worker-subscribe guard test in `RedisPubSubSubscribeRollbackTests`.
- **New** `ReclaimProcessingAsync_DrainsTheWholeProcessingListInOneCall_PreservingOrder` integration test.
- Full local runs: Game.Infrastructure.Tests **51**, Game.Application.Tests **335**, Game.Api.Tests **547**, Game.Core.Tests **649** — all green; `dotnet build` clean (0 warnings).

## Note
Item 5's switch to `IConnectionMultiplexer` is a clean dependency-on-abstractions improvement, though under this project's classical/integration testing stance the services hold no unit-testable logic to mock — its concrete payoff here is letting the disposal seam stay testable without a live connection.

Closes #954

---
_Generated by [Claude Code](https://claude.ai/code/session_01FE4GRZqyt6ZhqLC5BQx62Q)_

### DIFF
--- a/Game.Abstractions/Infrastructure/IPubSubService.cs
+++ b/Game.Abstractions/Infrastructure/IPubSubService.cs
@@ -26,8 +26,13 @@
         /// </summary>
         public Task Wake(string channel);
         public Task Subscribe(string channel, Action<(string message, string channel)> action, string? id = null);
-        public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string? id = null);
-        public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string? id = null);
+
+        // The worker-backed overloads require a non-null id: each spins up a BackgroundWorker (holding an OS wait
+        // handle) that can only be disposed via UnSubscribe(channel, id), so an id-less worker subscription could
+        // never be torn down and would leak its handle (#954). The plain-action overload above creates no worker,
+        // so its id stays optional.
+        public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string id);
+        public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string id);
         public Task UnSubscribe(string channel);
         public Task UnSubscribe(string channel, string id);
 
--- a/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
@@ -354,13 +354,13 @@ namespace Game.Api.Tests.Unit
             // The dead-letter queue the escalation writes a faulted server command to (the only GetQueue use).
             public RecordingQueue DeadLetterQueue { get; } = new();
 
-            public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string? id = null)
+            public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string id)
             {
                 CapturedProcessor = queue => action((queue, channel));
                 return Task.CompletedTask;
             }
 
-            public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string? id = null) => Task.CompletedTask;
+            public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string id) => Task.CompletedTask;
             public Task Subscribe(string channel, Action<(string message, string channel)> action, string? id = null) => Task.CompletedTask;
             public Task Publish(string channel, string message, CancellationToken cancellationToken = default) => Task.CompletedTask;
             public Task Publish(string channel, string queueName, string queueData, CancellationToken cancellationToken = default) => Task.CompletedTask;
@@ -430,9 +430,9 @@ namespace Game.Api.Tests.Unit
         /// <summary>A pub/sub whose <c>Subscribe</c> throws, to drive the registration-rollback path.</summary>
         private sealed class ThrowingSubscribePubSubService(Exception toThrow) : IPubSubService
         {
-            public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string? id = null) => throw toThrow;
+            public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string id) => throw toThrow;
 
-            public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string? id = null) => throw new NotSupportedException();
+            public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string id) => throw new NotSupportedException();
             public Task Subscribe(string channel, Action<(string message, string channel)> action, string? id = null) => throw new NotSupportedException();
             public Task Publish(string channel, string message, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task Publish(string channel, string queueName, string queueData, CancellationToken cancellationToken = default) => throw new NotSupportedException();
@@ -448,10 +448,10 @@ namespace Game.Api.Tests.Unit
         /// guarded-teardown path where the presence-key release must still run after an unsubscribe fault.</summary>
         private sealed class ThrowingUnsubscribePubSubService(Exception toThrow) : IPubSubService
         {
-            public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string? id = null) => Task.CompletedTask;
+            public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string id) => Task.CompletedTask;
             public Task UnSubscribe(string channel, string id) => throw toThrow;
 
-            public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string? id = null) => throw new NotSupportedException();
+            public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string id) => throw new NotSupportedException();
             public Task Subscribe(string channel, Action<(string message, string channel)> action, string? id = null) => throw new NotSupportedException();
             public Task Publish(string channel, string message, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task Publish(string channel, string queueName, string queueData, CancellationToken cancellationToken = default) => throw new NotSupportedException();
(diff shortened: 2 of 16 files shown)


ginderjeremiah (commented):
## Review: Redis infrastructure hygiene & robustness cleanups (#954)

**Verdict:** Approve (posted as a comment since I'm the PR author identity) — every item in #954 is either correctly implemented or, where it conflicted with a deliberate existing decision, addressed via the alternative the issue itself offered, with sound reasoning. No blocking findings.

I traced each of the seven items against the code:

1. **Null-id worker leak** — the worker-backed overloads now require a non-null `id`, validated *before* the `BackgroundWorker` is created (`RedisPubSubService.cs`), so the guard path can't leak either. `SubscribeWithWorker` drops its now-dead `id is not null` branches. `DataProviderSynchronizer` was the actual latent offender (it subscribed with no id) and now uses a per-instance `InstanceId` torn down in `StopAsync`, mirroring `ReferenceCacheSynchronizer`. `UnSubscribe(channel, id)` is a tolerant no-op when the id isn't registered, so a failed start can't make `StopAsync` throw. ✔️
2. **`GetSet(…, expiry)` null crash** — impl signature aligned to the non-null `ICacheService` contract, so a nil `ARGV` can't reach the Lua `SET`. ✔️
3. **Per-call handle resolution** — `IDatabase`/`ISubscriber` resolved once in the ctor. ✔️
4. **`ReclaimProcessingAsync` O(N)** — replaced with one atomic Lua `LMOVE RIGHT→LEFT` loop. I verified the ordering: items reserved head→tail land back at the head oldest-first, matching the prior per-item `ListMove`, and `lmove` on an empty list returns Lua `false` so the loop terminates. Up-front cancellation check preserved. New integration test covers multi-item order + idempotent second reclaim. ✔️
5. **Graceful multiplexer disposal** — new `RedisConnectionLifetime` hosted service, registered via `TryAddEnumerable` from both `AddCache`/`AddPubSub` (deduped) *ahead* of the Redis-consuming hosted services. I confirmed the ordering: `AddDataAccess` calls `AddCache`/`AddPubSub` before registering `DataProviderSynchronizer`/`ReferenceCacheSynchronizer`, and `SocketConnectionRegistry` is registered later still in `Startup`, so the lifetime stops last and connections drain after their consumers. Switch to `IConnectionMultiplexer` is a clean dependency-on-abstractions move and keeps the disposal seam testable. ✔️
6. **`RetryPolicy` zero base delay** — kept (a pinned, deliberate test affordance, bounded by `MaxAttempts`) and documented as immediate-retry semantics, the issue's offered alternative. ✔️
7. **DRY `ObserveWrite`** — already resolved by the `RedisCommandBudget.ObserveFault` extraction; no change needed. ✔️

### Nits / follow-up (non-blocking, pre-existing)

- The **non-expiry** `GetSet` has the same nullability mismatch this PR fixed for the expiry overload: `RedisService.GetSet(string key, string? value, …)` vs the interface's non-null `GetSet(string key, string value, …)`, and the generic `GetSet<T>` passes `value?.Serialize()` (possibly null) into it. Out of scope for this PR (#954 item 2 was specifically the Lua/expiry path), but worth a small consistency follow-up. I'll open a tracking issue.

### Test coverage

Good and proportionate to the change: `RedisConnectionLifetimeTests` (stop disposes / start doesn't), `RedisMultiplexerFactoryTests.DisposeAllAsync_*` (disposes every entry + clears, empty no-op), the null-id worker-subscribe guard test (both overloads, asserts it never connects), and the new multi-item reclaim ordering integration test. The disposal/lifetime seams are exercised in-process without a live connection, consistent with the project's classical testing stance.

### What's good

Careful, well-scoped work: each item traced back to the issue, the two places where the issue's primary suggestion conflicted with a deliberate decision were handled via the documented alternative rather than forced, and the reasoning is captured in tight code comments rather than the PR description alone. The hosted-service stop-ordering argument in particular is exactly the kind of thing that's easy to get subtly wrong and is both correct here and explained at the registration site.

---

PR #1016, opened 2026-06-20
Assert id==index contiguity when building reference snapshots (#955)

### LINKED ISSUE #955: Add a build-time id==index contiguity assertion for reference-data snapshots
## Problem

The whole reference-data scheme rests on the invariant `Id == listIndex` (id seeded from 0, id doubles as array index on the hottest gameplay path). The docs state this contiguity is *guaranteed by retire-not-delete* — and the admin repos do reject top-level hard deletes — but **nothing actually asserts the invariant when a snapshot is built**.

`Game.DataAccess/Repositories/ReferenceSnapshotExtensions.cs:17-36` and every holder (e.g. `EnemiesCacheHolder.cs:38`, `ItemsCacheHolder.cs:34`) only `OrderBy(x => x.Id)`, which guarantees sort order, **not** contiguity from 0. A seed/migration gap (the one way a gap could still appear, since hard-delete is blocked) would make `list[5]` silently resolve the wrong record on every enemy/skill/item lookup — silent corruption rather than a loud failure. `EnemyMapper.ToTemplate`'s `allSkills[es.SkillId]` inherits the same fragility.

## Suggested fix

- After `ToListAsync()`, assert `list[i].Id == i` for all `i` and throw from the build so a bad reload **fails the build-then-swap** (keeping the prior good snapshot in place) instead of serving corruption. Centralize this in one helper since all six holders need it.

## Related (orphaned player references)

`Game.DataAccess/Mapping/PlayerMapper.cs:84,95,130` resolve `GetItem`/`GetItemMod`/`GetSkill` against the in-memory catalogs with no missing-id guard, so if a referenced id ever went missing (again, a migration/seed mistake — retirement keeps the slot) the whole player aggregate fails to load. Worth deciding a policy (skip the orphaned ownership row with a warning vs. guarantee referential integrity) once the contiguity assertion above is in place. Defensive — both only trigger on a content-data mistake the retirement model is designed to prevent.

ginderjeremiah:
## Summary

The reference-data scheme rests on the invariant that a static reference record's `Id` doubles as its index into the cached list (ids seeded from 0). Contiguity is guaranteed *structurally* by retire-not-delete — and the admin repos reject top-level hard deletes — but **nothing asserted the invariant when a snapshot was built**. Every holder only `OrderBy(x => x.Id)`, which guarantees sort order, **not** contiguity from 0. A seed/migration gap (the one remaining way a gap could appear) would make `list[5]` silently resolve the wrong record on every lookup above the gap — silent corruption rather than a loud failure.

## How it addresses the issue

- Adds a centralized `AssertZeroBasedContiguity` helper to `ReferenceSnapshotExtensions` (next to the existing `Lookup`/`GetById` index helpers):
  - an **entity overload** over the shared `IZeroBasedIdentityEntity` contract (used by the five entity-backed holders via covariance — no per-set id selector needed), and
  - a generic **id-selector overload** for the challenge holder, which caches the projected domain `Challenge` rather than the EF entity.
- Calls it in **all six** holders right after the list is materialized. A violation throws `InvalidOperationException`, so:
  - on an **admin reload** the build-then-swap leaves the prior good snapshot in place (matching the existing fail-safe behavior pinned by `ReferenceCacheHolderTests`), and
  - on the **eager startup load** the corruption surfaces as a boot failure (the documented "a database problem surfaces as a boot failure" contract).
- The **enemies holder** also asserts the separate `skills` list it maps templates against — `EnemyMapper.ToTemplate` resolves an enemy's skills via `allSkills[SkillId]`, which the issue flags as inheriting the same fragility.
- Documents the backstop concisely in `docs/backend.md` → Reference Data.

## Not in this PR (follow-up)

The issue's "Related (orphaned player references)" section asks to decide a policy for `PlayerMapper`'s catalog resolution (skip the orphaned ownership row with a warning vs. guarantee referential integrity) **once the contiguity assertion is in place**. That's a genuine design fork, so I've opened a follow-up issue for it rather than bundling a decision here.

## Test plan

- New unit tests in `ReferenceSnapshotExtensionsTests` cover the contiguous case, empty list, off-by-one (seeded from 1), a mid-list gap (asserting the message names the failing index and id), and the entity overload reading `Id` off the contract.
- `Game.Application.Tests` — **340 passed** (includes the reference-cache holder/synchronizer integration tests that load the real seeded data through the new assertion, confirming no false-positive).

Closes #955

---
_Generated by [Claude Code](https://claude.ai/code/session_01R8EYgsfLENSSBtpn2TQaHc)_

### DIFF
--- a/Game.DataAccess/Repositories/ReferenceSnapshotExtensions.cs
+++ b/Game.DataAccess/Repositories/ReferenceSnapshotExtensions.cs
@@ -39,7 +39,7 @@ namespace Game.DataAccess.Repositories
 
         /// <summary>
         /// Asserts the zero-based-id contiguity invariant the snapshots rely on: the record at index
-        /// <c>i</c> must have <c>Id == i</c>, for every <c>i</c>. <see cref="OrderBy{T}"/>-by-id guarantees
+        /// <c>i</c> must have <c>Id == i</c>, for every <c>i</c>. <c>OrderBy</c>-by-id guarantees
         /// sort order but not that the ids run 0..n-1 with no gaps; a seed/migration gap (the one way a gap
         /// could appear, since hard-delete is blocked) would make every index lookup above the gap silently
         /// resolve the wrong record. Throwing here fails the build-then-swap so the prior good snapshot stays
--- a/Game.DataAccess/Repositories/ReferenceSnapshotExtensions.cs
+++ b/Game.DataAccess/Repositories/ReferenceSnapshotExtensions.cs
@@ -1,3 +1,5 @@
+using Game.Infrastructure.Entities;
+
 namespace Game.DataAccess.Repositories
 {
     /// <summary>
@@ -33,6 +35,40 @@ namespace Game.DataAccess.Repositories
             }
 
             return snapshot[id];
+        }
+
+        /// <summary>
+        /// Asserts the zero-based-id contiguity invariant the snapshots rely on: the record at index
+        /// <c>i</c> must have <c>Id == i</c>, for every <c>i</c>. <see cref="OrderBy{T}"/>-by-id guarantees
+        /// sort order but not that the ids run 0..n-1 with no gaps; a seed/migration gap (the one way a gap
+        /// could appear, since hard-delete is blocked) would make every index lookup above the gap silently
+        /// resolve the wrong record. Throwing here fails the build-then-swap so the prior good snapshot stays
+        /// in place (and a startup load surfaces the corruption as a boot failure) rather than serving
+        /// silently mis-resolved data.
+        /// </summary>
+        public static void AssertZeroBasedContiguity<T>(this IReadOnlyList<T> snapshot, Func<T, int> idSelector, string setName)
+        {
+            for (var i = 0; i < snapshot.Count; i++)
+            {
+                var id = idSelector(snapshot[i]);
+                if (id != i)
+                {
+                    throw new InvalidOperationException(
+                        $"Reference set '{setName}' violates the zero-based-id contiguity invariant: the record at " +
+                        $"index {i} has Id {id} (expected {i}). Ids must be seeded from 0 and contiguous; a gap " +
+                        "indicates a seed/migration mistake (top-level reference records are retired, never deleted).");
+                }
+            }
+        }
+
+        /// <summary>
+        /// Entity overload of <see cref="AssertZeroBasedContiguity{T}(IReadOnlyList{T}, Func{T, int}, string)"/>
+        /// for the cached entity lists, reading <c>Id</c> off the shared
+        /// <see cref="IZeroBasedIdentityEntity"/> contract so no per-set id selector is needed.
+        /// </summary>
+        public static void AssertZeroBasedContiguity(this IReadOnlyList<IZeroBasedIdentityEntity> snapshot, string setName)
+        {
+            snapshot.AssertZeroBasedContiguity(entity => entity.Id, setName);
         }
     }
 }
--- a/Game.Application.Tests/DataAccess/ReferenceSnapshotExtensionsTests.cs
+++ b/Game.Application.Tests/DataAccess/ReferenceSnapshotExtensionsTests.cs
@@ -1,4 +1,5 @@
 using Game.DataAccess.Repositories;
+using Game.Infrastructure.Entities;
 using Xunit;
 
 namespace Game.Application.Tests.DataAccess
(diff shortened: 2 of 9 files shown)


Game.DataAccess/Repositories/ReferenceSnapshotExtensions.cs
```diff
@@ -34,5 +36,39 @@ public static T GetById<T>(this IReadOnlyList<T> snapshot, int id, string setNam
 
             return snapshot[id];
         }
+
+        /// <summary>
+        /// Asserts the zero-based-id contiguity invariant the snapshots rely on: the record at index
+        /// <c>i</c> must have <c>Id == i</c>, for every <c>i</c>. <see cref="OrderBy{T}"/>-by-id guarantees
```
ginderjeremiah: Nit (non-blocking): `<see cref="OrderBy{T}"/>` is a dangling cref — there's no `OrderBy` method in scope for it to resolve to (the holders use LINQ's `Enumerable.OrderBy`). It's harmless today since no project sets `GenerateDocumentationFile`/`TreatWarningsAsErrors`, but it would surface as CS1574 if doc generation is ever enabled. Consider plain `<c>OrderBy</c>` instead.

---

ginderjeremiah (commented):
## Review: Build-time `Id == index` contiguity assertion for reference snapshots (#955)

**Verdict:** Approve *(submitted as a comment — I'm the PR author, so GitHub won't accept a gating verdict from me).* Clean, focused, well-tested change that does exactly what the issue asked.

This directly closes the gap #955 identified: every holder only `OrderBy(x => x.Id)`, which guarantees sort order but not zero-based contiguity, leaving a seed/migration gap to silently mis-resolve every index lookup above it. The PR adds a centralized `AssertZeroBasedContiguity` and wires it into all six holders right after materialization, so a corrupt set fails the build-then-swap (prior good snapshot stays) or surfaces as a boot failure on the eager load — matching the documented contracts.

### What's good
- **Right abstraction.** The entity overload reads `Id` off the shared `IZeroBasedIdentityEntity` contract and leans on `IReadOnlyList<out T>` covariance, so the five entity-backed holders pass their lists with no per-set selector; the generic id-selector overload is correctly reserved for the challenge holder, which caches the projected `Game.Core.Challenge` domain object rather than the EF entity. DRY — the entity overload just delegates to the generic one.
- **Placement & layering.** Lives in the data-access layer where snapshots are built; no domain logic leaks. The enemies holder also asserts the separate `skills` list it maps against via `allSkills[SkillId]`, which is the exact secondary fragility the issue flagged.
- **Docs.** `docs/backend.md` → Reference Data is updated concisely to describe the backstop without over-documenting.
- **Follow-up handled correctly.** The orphaned-player-reference policy decision is a genuine design fork and was split into #1017 (labeled `tech debt` / `claude` / `scope: small`) rather than bundled here. Good call.

### Test coverage
Solid for the scope. `ReferenceSnapshotExtensionsTests` covers contiguous, empty, off-by-one (seeded from 1), a mid-list gap, and the entity overload reading `Id` off the contract — and asserts the exception message names the failing index and id. The real seeded data flows through the new assertion via the existing holder/synchronizer integration tests, confirming no false-positive on real content. No new domain/battle logic, so nothing to mirror on the frontend.

### Nits
- `ReferenceSnapshotExtensions.cs:42` — `<see cref="OrderBy{T}"/>` is a dangling doc cref (inline comment). Purely cosmetic; harmless today.

Nothing blocking. Once CI goes green this is good to merge.

---

ginderjeremiah:
## Review summary — ✅ Approve

Reviewed against the project's standards (DDD/layering, DRY, mandatory braces, null-handling, test coverage, docs).

**What it does:** Adds a centralized `AssertZeroBasedContiguity` backstop to `ReferenceSnapshotExtensions` (an `IZeroBasedIdentityEntity` overload + a generic id-selector overload) and invokes it in all six reference-cache holders right after the list is materialized, plus on the enemies holder's separate `skills` list. This turns a previously *silent* corruption (a seed/migration gap making `list[5]` resolve the wrong record on every lookup above the gap) into a loud `InvalidOperationException` — which preserves the prior good snapshot on an admin reload and surfaces as a boot failure on the eager startup load, matching the documented contracts. Concisely documented in `docs/backend.md`.

**Verification beyond the diff:**
- No false-positive risk on retired content: the main `enemies` query and the other holders don't filter `RetiredAt`, so retired records keep their id slots and contiguity holds. Retirement is excluded only from the *derived* spawn tables, which aren't index-addressed. The 340-passing integration suite loads real seeded data through the new assertion, confirming no false-positive.
- Layering respected — the entity overload stays inside the data tier (`Game.DataAccess` references `Game.Infrastructure` with `PrivateAssets="Compile"`); the helper is `internal`; no domain logic leaks.
- DRY: entity overload delegates to the generic one; the generic overload is correctly reserved for the challenge holder (which caches the projected `Challenge`, not the EF entity).
- Braces everywhere, no null-forgiving operator introduced.
- Test coverage is strong for the scope: contiguous, empty, off-by-one (seeded from 1), mid-list gap (asserting the message names the failing index/id), and the entity overload.

**CI:** Green — `all-checks-passed`, `test-backend`, `verify-backend`, `test-e2e (chromium/firefox/webkit)`, `detect-changes` all succeeded; frontend checks correctly skipped (no FE changes).

**Findings:** No blocking or should-fix issues. The one doc-cref nit from the author's self-review is already fixed in 472ce7e.

**Follow-ups:** None to create — the one genuine design fork (orphaned `PlayerMapper` catalog references: skip-with-warning vs. enforce referential integrity) was correctly split into #1017 rather than bundled here, which is what the issue asked for.

Merging.

---

PR #1019, opened 2026-06-20
SocketInterceptorMiddleware: return { errorMessage } envelope on the 401/400 short-circuits

### LINKED ISSUE #1012: SocketInterceptorMiddleware: return the { errorMessage } envelope on the 401/400 short-circuits
## Background

Follow-up from #953 / PR #1011, which fixed the handshake `500` path in `Game.Api/Middleware/SocketInterceptorMiddleware.cs` to return a `404` with the project's `ApiResponse.Error` (`{ errorMessage }`) envelope.

The same middleware has two sibling short-circuits that still set a bare status code with no body:

- `Game.Api/Middleware/SocketInterceptorMiddleware.cs:22` — `401 Unauthorized` when the session is unauthenticated.
- `Game.Api/Middleware/SocketInterceptorMiddleware.cs:26` — `400 BadRequest` when the request to `/socket` isn't a WebSocket upgrade.

These are out of scope for #953 (which only flagged the `500`), and the PR author deliberately scoped them out to keep the change focused — but they're inconsistent with the project's contract that every failure returns the `{ errorMessage }` envelope.

## Suggested change

Write the `ApiResponse.Error(...)` envelope on the `401` and `400` branches as well, matching the `404` branch. Confirm a body is writable at these points (the socket isn't upgraded yet, so it should be) and that no existing client relies on an empty body for these statuses.

## Notes

- `tech debt` / `scope: small` — response shaping only, no user-facing behavior change beyond the added body.

ginderjeremiah:
Closes #1012

## Summary

Follow-up to #953 / #1011, which fixed the handshake `500` path in `SocketInterceptorMiddleware` to return the project's `{ errorMessage }` envelope. The same middleware had two sibling short-circuits that still set a bare status code with no body:

- `401 Unauthorized` when the session is unauthenticated.
- `400 BadRequest` when the request to `/socket` isn't a WebSocket upgrade.

Both branches now write the `ApiResponse.Error(...)` envelope via `WriteAsJsonAsync`, using the exact same approach as the `404` branch from #1011. A body is writable at these points because the socket hasn't been upgraded yet. Clear, case-specific messages are provided for each (authentication required vs. WebSocket upgrade required). No status codes or control flow changed — this is response-shaping only.

## How I tested

- Built the full solution with `-warnaserror`: 0 warnings, 0 errors.
- Added two integration tests to `SocketConnectionTests` mirroring the existing handshake coverage, asserting each short-circuit now returns the `{ errorMessage }` body:
  - `Socket_Unauthenticated_ReturnsErrorEnvelope` — plain HTTP GET to `/socket` with no token returns 401 + envelope.
  - `Socket_AuthenticatedButNotWebSocketUpgrade_ReturnsErrorEnvelope` — authenticated non-upgrade GET returns 400 + envelope.
- Ran the targeted integration suite `--filter-class "*SocketConnectionTests"`: 7/7 passed (held the shared-infra lock briefly).
- Ran the full `Game.Api.Tests` unit suite: 136/136 passed.

## Notes / follow-ups

- #1011 fixed the `404` branch but did not add a dedicated integration test asserting its `{ errorMessage }` body; the two tests added here establish that pattern for the 401/400 branches. A matching test for the 404 (player-not-loadable) branch could be added later for symmetry, but it's outside this issue's scope.
- The middleware sets the status code directly, so the `ApiResponse` `ErrorCategory` (which only steers MVC results via `ErrorStatusFilter`) is intentionally irrelevant here — matching how the 404 branch passes only the message.

---

### DIFF
--- a/Game.Api.Tests/Integration/SocketConnectionTests.cs
+++ b/Game.Api.Tests/Integration/SocketConnectionTests.cs
@@ -1,4 +1,5 @@
 using Game.Api;
+using Game.Api.Models.Common;
 using Game.Infrastructure.Database;
 using Game.TestInfrastructure.Fixtures;
 using Game.TestInfrastructure.Helpers;
@@ -77,6 +78,35 @@ namespace Game.Api.Tests.Integration
 
             await Assert.ThrowsAnyAsync<Exception>(
                 () => wsClient.ConnectAsync(new Uri("ws://localhost/socket"), CancellationToken));
+        }
+
+        [Fact]
+        public async Task Socket_Unauthenticated_ReturnsErrorEnvelope()
+        {
+            // A plain HTTP request to /socket with no token short-circuits to 401 before any upgrade, so the
+            // body carries the project's { errorMessage } envelope rather than being empty.
+            var response = await Client.GetAsync("/socket", CancellationToken);
+
+            Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            var body = await response.Content.ReadFromJsonAsync<ApiResponse>(CancellationToken);
+            Assert.NotNull(body);
+            Assert.False(string.IsNullOrWhiteSpace(body.ErrorMessage));
+        }
+
+        [Fact]
+        public async Task Socket_AuthenticatedButNotWebSocketUpgrade_ReturnsErrorEnvelope()
+        {
+            // An authenticated but non-upgrade request to /socket short-circuits to 400 before any upgrade, so
+            // the body carries the { errorMessage } envelope rather than being empty.
+            var (userId, playerId) = await SeedAsync();
+            var authedClient = CreateAuthenticatedClient(userId, playerId);
+
+            var response = await authedClient.GetAsync("/socket", CancellationToken);
+
+            Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
+            var body = await response.Content.ReadFromJsonAsync<ApiResponse>(CancellationToken);
+            Assert.NotNull(body);
+            Assert.False(string.IsNullOrWhiteSpace(body.ErrorMessage));
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: return the `{ errorMessage }` envelope on the socket 401/400 short-circuits (closes #1012)

**Verdict:** Approve _(submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author)_ — the change does exactly what #1012 asks, mirrors the established #1011 `404` pattern, and is correctly tested.

### What's good
- **Faithful to the pattern.** Both new branches use `ApiResponse.Error(...)` + `WriteAsJsonAsync` with `context.RequestAborted`, identical to the `404` (player-not-loadable) branch directly below — no new approach invented, status codes and control flow untouched (`SocketInterceptorMiddleware.cs:24` / `:30`).
- **Correct overload choice.** Using the single-arg `ApiResponse.Error(message)` is right here: the middleware sets `StatusCode` directly, and `ErrorCategory` only steers MVC results via `ErrorStatusFilter`, so it's genuinely irrelevant on this path. The PR notes this explicitly — good call-out.
- **Body is safely writable.** The socket isn't upgraded at either branch, so writing a JSON body is valid; the comments explain this without over-narrating.
- **Messages are clear and case-specific** (authentication-required vs. WebSocket-upgrade-required), which is the whole point of the envelope.

### Test coverage
- Two integration tests added to `SocketConnectionTests`, mirroring the existing handshake coverage: `Socket_Unauthenticated_ReturnsErrorEnvelope` (plain GET → 401 + envelope) and `Socket_AuthenticatedButNotWebSocketUpgrade_ReturnsErrorEnvelope` (authed non-upgrade GET → 400 + envelope). Each asserts the status code and a non-blank `ErrorMessage` — asserting presence rather than exact text is the right level of brittleness here. Adequate for response-shaping; matches the project's integration-test guidance for out-of-process interactions.

### Nits
- None.

### Follow-ups (out of scope, already noted in the PR)
- The `404` (player-not-loadable) branch from #1011 still lacks a dedicated envelope test; a matching one could be added later for symmetry. Not blocking this change.


---

PR #1021, opened 2026-06-20
Frontend small perf/cleanups (per-frame skill lookup, inventory .find→Map.get, item-tooltip test, KindTag fallback)

### LINKED ISSUE #957: Frontend small perf/cleanups (per-frame skill lookup, inventory .find→Map.get, item-tooltip test, KindTag fallback)
## Problem

The Svelte 5 frontend is in strong health overall (theming discipline, lifecycle teardown, and view-model test coverage all check out). A few small, isolated cleanups remain.

### 1. O(n·m) skill-by-effect-id lookup runs every render frame while an effect tooltip is open
`UI/src/routes/game/screens/fight/ActiveEffectChips.svelte:81-82` (consumed by the `shownEffectContext` `$derived` at `:85-95`)

`sourceSkillName` does `staticData.skills?.find(skill => skill?.effects.some(e => e.id === sourceId))` — O(skills × effects). It's recomputed inside `shownEffectContext`, which depends on `shownEffect.renderRemainingMs` (interpolated every frame), so the full skill catalogue is rescanned each frame while an active-effect chip tooltip is hovered.

**Fix:** build a `Map<effectId, skill>` once (module-level memo keyed off the `staticData.skills` reference) and `.get(sourceId)`. The skill→effect ownership is static reference data.

### 2. Inventory item resolution uses `.find()` when an O(1) `itemId`-keyed Map exists
`UI/src/routes/game/screens/inventory/inventory-view.svelte.ts:76-78` (`selected`) and `:80-82` (`dragItem`)

Both derivations do `this.items.find(i => i.itemId === id)`, but `InventoryManager.unlockedItems` is a public `Map<number, Item>` keyed by `itemId` (`inventory-manager.ts:37`) that the manager itself resolves through everywhere. Low-frequency (selection/drag) but contradicts the manager's own O(1) access convention.

**Fix:** `inventoryManager.unlockedItems.get(this.selectedId) ?? null` (and the same for `dragItemId`); keep `items` as the source for list/count derivations.

### 3. `item-tooltip.svelte.ts` is the one view-model controller without a dedicated test
`UI/src/routes/game/screens/inventory/item-tooltip.svelte.ts`

Every other screen/component controller has a matching test. This shared controller (`createItemTooltip` show/move/hide + context plumbing) does not. Add a small test asserting `show` sets `item` + positions + shows, `move` repositions without clearing `item`, and `hide` clears `item`.

### 4. `KindTag` silently renders an empty pill for an unmapped `EAttributeType`
`UI/src/.../KindTag.svelte:20`

Falls back to `''` for an unmapped value, rendering an invisible empty pill rather than failing loud. The taxonomy is currently complete so it never triggers; a `?? 'Unknown'` fallback (or a test asserting all `EAttributeType` values are mapped) would surface a future enum addition.

> Note: `statistics-view.svelte.ts:206-208` also uses a per-row `.find` over the projected entity list (O(rows × entities)); that likely falls within #914's "Stats" scope, so it's not duplicated here.

ginderjeremiah:
Closes #957

Four small, isolated frontend cleanups.

## 1. Per-frame skill-by-effect-id lookup (`ActiveEffectChips.svelte`)
`sourceSkillName` did `staticData.skills?.find(s => s?.effects.some(e => e.id === sourceId))` — O(skills × effects) — inside the `shownEffectContext` `$derived`, which depends on a per-frame interpolated value, so the whole skill catalogue was rescanned every frame while an effect-chip tooltip was hovered. Replaced with an `effectId → skill name` `Map` memoised off the `staticData.skills` reference (skill→effect ownership is static reference data, so the index is rebuilt only when that catalogue changes); resolution is now `.get(sourceId)`.

## 2. Inventory selection/drag resolution (`inventory-view.svelte.ts`)
`selected` and `dragItem` did `this.items.find(i => i.itemId === id)`. `InventoryManager.unlockedItems` is a public `itemId`-keyed `Map` the manager resolves through everywhere, so both now use `inventoryManager.unlockedItems.get(id) ?? null`. `items` remains the source for the list/count/visible derivations.

## 3. `createItemTooltip` controller test (new)
`item-tooltip.svelte.ts` was the one view-model controller without a dedicated test. Added `item-tooltip.test.ts` (plus a small `ItemTooltipControllerFixture.svelte` so the controller's `onDestroy` cleanup runs in a real component lifecycle, mirroring the tooltip-store tests) asserting: `show` sets the item, positions the panel, and reveals it; `move` repositions without clearing the item; `hide` clears the item and conceals the panel; and register/unregister + `describedById` wiring.

## 4. `KindTag` fallback (`KindTag.svelte`)
An unmapped `EAttributeType` fell back to `''`, rendering an invisible empty pill. Changed the fallback to a visible `'Unknown'` so a future enum addition surfaces loudly, and added a test asserting every `EAttributeType` value maps to a display group (a durable guard against the taxonomy falling out of sync).

## Verification
- `vitest run` on all affected suites (inventory, attribute-breakdown, fight) — all green; the existing `ActiveEffectChips.test.ts` (which exercises `sourceSkillName`) still passes unchanged, confirming item 1's output is identical.
- `npm run check` (svelte-check): 0 errors / 0 warnings.
- `eslint .`: clean.

---

### DIFF
--- a/UI/src/routes/game/screens/inventory/inventory-view.svelte.ts
+++ b/UI/src/routes/game/screens/inventory/inventory-view.svelte.ts
@@ -75,6 +75,8 @@ export class InventoryView {
 
 	// Resolve selection/drag through the manager's itemId-keyed Map (its own O(1) access convention)
 	// rather than a linear scan of the list; `items` stays the source for list/count derivations.
+	// NOTE: unlockedItems is a plain (non-reactive) Map, so these re-resolve only on id changes, not on
+	// set changes — fine while items are add-only; if a removal path lands, clear the ids on removal.
 	readonly selected = $derived(
 		this.selectedId != null ? (inventoryManager.unlockedItems.get(this.selectedId) ?? null) : null
 	);
--- a/UI/src/routes/game/screens/inventory/inventory-view.svelte.ts
+++ b/UI/src/routes/game/screens/inventory/inventory-view.svelte.ts
@@ -73,12 +73,14 @@ export class InventoryView {
 		return map;
 	});
 
+	// Resolve selection/drag through the manager's itemId-keyed Map (its own O(1) access convention)
+	// rather than a linear scan of the list; `items` stays the source for list/count derivations.
 	readonly selected = $derived(
-		this.selectedId != null ? (this.items.find((i) => i.itemId === this.selectedId) ?? null) : null
+		this.selectedId != null ? (inventoryManager.unlockedItems.get(this.selectedId) ?? null) : null
 	);
 
 	readonly dragItem = $derived(
-		this.dragItemId != null ? (this.items.find((i) => i.itemId === this.dragItemId) ?? null) : null
+		this.dragItemId != null ? (inventoryManager.unlockedItems.get(this.dragItemId) ?? null) : null
 	);
 
 	readonly counts = $derived.by(() => {
--- a/UI/src/routes/game/screens/attribute-breakdown/KindTag.svelte
+++ b/UI/src/routes/game/screens/attribute-breakdown/KindTag.svelte
@@ -17,7 +17,9 @@ interface Props {
 
 let { type }: Props = $props();
 
-const label = $derived(ATTRIBUTE_TYPE_GROUPS.find((g) => g.type === type)?.label ?? '');
+// Falls back to a visible "Unknown" rather than an empty pill so an unmapped EAttributeType (e.g. a
+// future enum addition the taxonomy hasn't caught up to) surfaces loudly instead of rendering blank.
+const label = $derived(ATTRIBUTE_TYPE_GROUPS.find((g) => g.type === type)?.label ?? 'Unknown');
 </script>
 
 <style lang="scss">
--- a/UI/src/routes/game/screens/fight/ActiveEffectChips.svelte
+++ b/UI/src/routes/game/screens/fight/ActiveEffectChips.svelte
@@ -137,10 +137,32 @@ const tip = createAttributeTooltip(() => tooltip);
 let shownKey = $state<string>();
 const shownGroup = $derived(shownKey != null ? effectGroups.find((g) => g.key === shownKey) : undefined);
 
-// The skill an active effect came from: its `sourceId` is the authored skill-effect id, so find the
-// skill that owns it. Display-only (never touches battle math); undefined for a retired/unknown skill.
-const sourceSkillName = (sourceId: number): string | undefined =>
-	staticData.skills?.find((skill) => skill?.effects.some((e) => e.id === sourceId))?.name;
+// The skill an active effect came from: its `sourceId` is the authored skill-effect id, so look up
+// the skill that owns it. Display-only (never touches battle math); undefined for a retired/unknown
+// skill. Skill→effect ownership is static reference data, so the effectId→name index is memoised off
+// the `staticData.skills` reference and rebuilt only when that catalogue changes — rather than
+// rescanning every skill's effects per call (this resolves inside the per-frame `shownEffectContext`).
+let skillNameByEffectIdCache: { skills: unknown; map: Map<number, string> } | undefined;
+const skillNameByEffectId = (): Map<number, string> => {
+	const skills = staticData.skills;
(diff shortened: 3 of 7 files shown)


UI/src/routes/game/screens/inventory/inventory-view.svelte.ts
```diff
@@ -73,12 +73,14 @@ export class InventoryView {
...
+	// rather than a linear scan of the list; `items` stays the source for list/count derivations.
 	readonly selected = $derived(
-		this.selectedId != null ? (this.items.find((i) => i.itemId === this.selectedId) ?? null) : null
+		this.selectedId != null ? (inventoryManager.unlockedItems.get(this.selectedId) ?? null) : null
 	);
 
 	readonly dragItem = $derived(
-		this.dragItemId != null ? (this.items.find((i) => i.itemId === this.dragItemId) ?? null) : null
+		this.dragItemId != null ? (inventoryManager.unlockedItems.get(this.dragItemId) ?? null) : null
 	);
```
ginderjeremiah, line 86: Nit (dormant, non-blocking): `inventoryManager.unlockedItems` is a plain `Map` (not a `SvelteMap` / reassigned `$state`), so these `$derived`s only re-run when `selectedId`/`dragItemId` change — not when the item *set* changes. The previous `this.items.find(...)` read the statify'd reactive array, so it would re-resolve to `null` if the selected item left the set. Today this is harmless: items are only ever added (`publish()` on add) and never removed, and in-place field edits stay reactive through the statify'd item objects this returns. But if an item-removal path is ever added, these derivations would silently hold a stale removed item. Worth a one-line note here, or clearing `selectedId`/`dragItemId` on removal when that path lands.

---

ginderjeremiah (commented):
## Review: Four small frontend perf/cleanups (closes #957)

**Verdict:** Approve _(submitted as a comment — I'm the PR author, so GitHub won't accept a gating approval from me)_ — all four items land exactly as the issue specified, with good test coverage and clear rationale. The single finding below is a dormant nit, not a blocker.

### Item-by-item
1. **Per-frame skill lookup (`ActiveEffectChips.svelte`)** — the O(skills×effects) `.find` that ran inside the per-frame `shownEffectContext` `$derived` is now an `effectId→name` `Map` memoised off the `staticData.skills` reference. Correctly rebuilt only when the catalogue reference changes; the `svelte/prefer-svelte-reactivity` disable is justified (a non-reactive memo that never drives rendering). Resolution drops to `.get(sourceId)` with identical output, and the existing `ActiveEffectChips.test.ts` still passes — good evidence of behavioral equivalence.
2. **Inventory `.find → Map.get` (`inventory-view.svelte.ts`)** — `selected`/`dragItem` now resolve through the manager's `itemId`-keyed `unlockedItems` Map, matching the manager's own O(1) access convention, with `items` kept as the source for list/count derivations. One dormant reactivity caveat noted inline.
3. **`createItemTooltip` test (new)** — fills the one missing view-model controller test. The `ItemTooltipControllerFixture.svelte` to run the `onDestroy` cleanup in a real lifecycle mirrors the established tooltip-store test pattern; show/move/hide, register/unregister, and `describedById` wiring are all asserted.
4. **`KindTag` fallback** — `'' → 'Unknown'` so an unmapped `EAttributeType` surfaces loudly, backed by a taxonomy-coverage test asserting every numeric `EAttributeType` value maps to a group (a durable guard, which is the more valuable half of the fix).

### Nits
- `inventory-view.svelte.ts:78-84` — `selected`/`dragItem` now read a non-reactive plain `Map`, so they won't re-resolve on set-level changes (only on `selectedId`/`dragItemId` changes). Harmless today (no item-removal path; field edits stay reactive via statify'd items), but a latent footgun if removal is ever added. See inline.

### Test coverage
Meets the project's frontend testing rules: new controller has a dedicated test mirroring its source path, the KindTag concern is guarded by an enum-coverage test, and the inventory test mock was updated to keep `unlockedItems` in sync with `sampleItems`. No battle logic touched, so no frontend/backend mirroring needed.

### What's good
Tightly scoped, each change traces directly to the issue, and the inline comments explain the *why* (per-frame cost, static reference data, fail-loud intent) without over-explaining. Nicely done.

---

PR #1022, opened 2026-06-20
Fail loudly on orphaned player references in PlayerMapper (#1017)

### LINKED ISSUE #1017: Decide a missing-reference policy for PlayerMapper's catalog resolution
Follow-up to #955 (split out so the contiguity assertion lands first).

## Background

With the build-time `Id == index` contiguity assertion now in place for the reference snapshots (#955 / #1016), the remaining defensive gap is on the **player-load** side.

`Game.DataAccess/Mapping/PlayerMapper.cs` resolves owned references against the in-memory catalogs (`GetItem` / `GetItemMod` / `GetSkill`) with **no missing-id guard**. The retirement model keeps a slot resolvable forever, so this only triggers on a content-data mistake (a migration/seed error that actually removed a referenced id). But when it does, the failure mode is severe: a single orphaned ownership row makes the **whole player aggregate fail to load**, locking that player out entirely.

## Decision to make

Pick a policy for an unresolvable owned reference and apply it consistently across the three resolution points:

1. **Skip-with-warning** — drop the orphaned ownership row, log a warning (and ideally surface it for content review), and let the player load. Resilient, but silently discards player-owned data.
2. **Guarantee referential integrity** — keep the hard failure but make it loud/diagnosable (which player, which id, which catalog), and rely on the contiguity assertion + retire-not-delete to ensure it never fires in practice.

Option 1 is the more player-friendly default; option 2 leans on the invariants the rest of the system already enforces. Worth a quick design call before implementing.

## Acceptance

- Policy decided and documented (brief note in `docs/backend.md` if it sets an expectation).
- All three `PlayerMapper` resolution points handle a missing id per the chosen policy.
- Unit tests cover the orphaned-reference path for each of item / item-mod / skill ownership.

Labels: tech debt, claude, scope: small

ginderjeremiah:
Closes #1017

## Chosen policy

Per the owner's decision, this implements **loud fail-fast / guarantee referential integrity** (option 2 in the issue), not skip-with-warning. An unresolvable owned reference keeps the hard failure — the player-owned row is **never** silently dropped — but the failure is now made loud and diagnosable.

## Summary

`PlayerMapper.ToCore` resolved a player's owned items, item mods, and skills against the in-memory catalogs (`GetItem` / `GetItemMod` / `GetSkill`) with no missing-id guard. A single unresolvable owned reference made the whole player aggregate fail to load with an opaque error, locking that player out.

- Added a small shared `ResolveOrThrow` helper in `PlayerMapper`, applied at all three resolution points, that rethrows a catalog miss as a new sealed `OrphanedReferenceException`.
- The exception message names the **player id**, the **catalog** (item / item mod / skill), and the **missing id**, and preserves the originating catalog failure as the inner exception — so a content-data mistake (a seed/migration that removed a referenced id) is immediately diagnosable from logs instead of surfacing as an opaque load failure.
- This leans on the invariants the rest of the system already enforces (the `Id == index` contiguity assertion from #955/#1016 plus retire-not-delete), so in practice it should never fire — but when it does, it fails loudly with an actionable message.
- Documented the policy concisely in `docs/backend.md` (Reference Data section), since it sets an expectation about player-load failure behavior.

## How tested

- Added three unit tests covering the orphaned-reference path for item, item-mod, and skill ownership, asserting the right exception type (`OrphanedReferenceException`) and that the message contains the player id, catalog label, and missing id (plus a preserved inner exception). Mirrors the existing `PlayerMapperTests` patterns.
- `Game.Application.Tests` filtered to `PlayerMapperTests`: 7/7 passing (4 existing + 3 new).
- Built `Game.DataAccess` and `Game.Application.Tests` with `-warnaserror`: 0 warnings, 0 errors.
- The new tests are pure unit tests (no out-of-process dependency), so no integration lock was needed.

---


Game.DataAccess/Mapping/PlayerMapper.cs
```diff
@@ -167,5 +167,23 @@ public static Player ToCore(EntityPlayer entity, IItems items, IItemMods itemMod
...
+        /// id. We never silently drop the player-owned row; instead the aggregate fails loudly so a content-data
+        /// mistake (a removed referenced id) is obvious from logs rather than surfacing as an opaque load failure.
+        /// </summary>
+        private static T ResolveOrThrow<T>(Func<int, T> resolve, int referenceId, int playerId, string catalogName)
+        {
+            try
+            {
+                return resolve(referenceId);
+            }
+            catch (Exception ex)
```
ginderjeremiah: **Should-fix:** this catches the broad `Exception`, but the documented catalog contract for an unresolvable id is specifically `ArgumentOutOfRangeException` (`ReferenceSnapshotExtensions.GetById` throws exactly that, and your test double faithfully mirrors it). Catching everything means any *unrelated* failure inside `resolve` — a future NRE/bug in `GetItem`, etc. — gets re-thrown as an `OrphanedReferenceException` whose message asserts a definite cause ("a content-data mistake — a referenced id was removed"). That's an actively misleading diagnosis for a bug that has nothing to do with content data, and it's the kind of footgun the loud-fail-fast policy is meant to avoid.

Suggest narrowing to `catch (ArgumentOutOfRangeException ex)` so genuine orphaned-reference misses are wrapped with the diagnosable message while any other exception propagates unwrapped (and the orphaned-reference message stays truthful). The inner exception is still preserved either way.

---

ginderjeremiah (commented):
## Review: Fail loudly on orphaned player references in `PlayerMapper` (#1017)

**Intended verdict:** Approve with one should-fix — _(posted as a comment because GitHub won't let me gate my own PR)._ The change correctly implements the owner-chosen policy (option 2: guarantee referential integrity / loud fail-fast) and satisfies every acceptance criterion in #1017. One non-blocking footgun in the exception handling is worth tightening before merge.

### Should-fix
- `PlayerMapper.cs:183` — `ResolveOrThrow` catches the broad `Exception` and re-throws it as `OrphanedReferenceException` with a message that asserts a specific cause ("a content-data mistake — a referenced id was removed"). The catalog's actual missing-id contract is `ArgumentOutOfRangeException` (`ReferenceSnapshotExtensions.GetById`), so an unrelated failure would get a misleading "orphaned reference" diagnosis. Narrow the catch to `ArgumentOutOfRangeException` (details inline).

### Test coverage
- Solid and exactly to the acceptance criteria: all three resolution points (item / item mod / skill) are covered, asserting the `OrphanedReferenceException` type, that the message names the player id + catalog label + missing id, and that the originating exception is preserved as `InnerException`. The `InMemoryCatalog` test double faithfully mirrors the real catalog's `ArgumentOutOfRangeException`, so the path under test matches production. No gaps. (Note: if you narrow the catch per the should-fix, the tests already align since the double throws AOORE.)

### What's good
- Clean DRY application of a single `ResolveOrThrow` helper across all three points rather than three copies.
- `OrphanedReferenceException` is `internal sealed` and stays inside `Game.DataAccess`, respecting the layering — good.
- Policy is documented concisely in `docs/backend.md` alongside the existing contiguity/retirement invariants it leans on, which is exactly where a reader would look.
- Diagnosable message (player id, catalog, missing id) plus preserved inner exception turns a previously opaque player-lockout into an actionable log line.

Not merging on this pass: there's an outstanding should-fix, and CI (`test-backend`, `verify-backend`, e2e) was still in progress at review time.

---

PR #1024, opened 2026-06-20
Align background reference-cache reload with the request-time path

### LINKED ISSUE #1008: Background reference-cache reload sweeps serially and re-reloads already-succeeded caches on retry
## Summary

The two entry points that reload the in-memory reference caches after an admin write have diverged. The request-time `AdminCacheReloadFilter` reloads all holders **concurrently** (with a comment explaining why), but the cross-instance background path in `CoalescingReferenceCacheReloader` reloads them **serially** and, on a mid-sweep failure, re-reloads the holders that already succeeded. Both are correctness-safe today (no holder depends on another's reload order), so this is an efficiency/consistency cleanup, not a bug.

## Detail

**(a) Serial vs. concurrent divergence.** The request-time filter (`Game.Api/Filters/AdminCacheReloadFilter.cs:52-57`) reloads concurrently and documents the rationale:

```csharp
// Reload all holders concurrently rather than back-to-back ... no holder depends on another's
// reload order, so a serial pass only adds latency and could leave some sets fresh and some
// stale if one query failed mid-list.
await Task.WhenAll(caches.Select(cache => cache.ReloadAsync()));
```

The background reloader (`Game.DataAccess/CoalescingReferenceCacheReloader.cs:67-72`) does the opposite:

```csharp
foreach (var cache in caches)
{
    await cache.ReloadAsync(cancellationToken);
}
```

So the two paths have different latency and partial-failure behaviour for the same logical operation (N sequential DB round-trips on the background path vs. one concurrent fan-out on the request path).

**(b) Retry re-reloads succeeded caches.** The retry loop (`CoalescingReferenceCacheReloader.cs:63-88`) wraps the entire `foreach`, so a failure on the k-th cache re-runs `ReloadAsync` (full re-query + snapshot rebuild + atomic swap) for caches `1..k-1` again on the next attempt. Each individual reload is already a build-then-swap that leaves the previous snapshot in place on failure, so re-reloading a succeeded cache is wasted work, not a correctness problem.

## Suggested fix

Reload via `Task.WhenAll` in `ReloadAllAsync` (with the retry wrapping the whole sweep) for parity with the request-time filter — the simplest change and it closes the latency gap. Optionally, track the still-pending set across attempts so a succeeded holder isn't rebuilt again on retry. Confirm the concurrency is safe for `EnemiesCacheHolder` (it re-queries the skills graph directly from its own scoped context rather than reading another holder's snapshot, so there is no cross-holder ordering dependency — see its reload comment).

## Notes

Adjacent to but distinct from #356 (cache-reload design spike) and #938 (per-call hot-path overhead). A small related observation worth bearing in mind while in this code: `EnemiesCacheHolder` fully re-materializes the Skills graph on every enemy reload for correctness (avoiding a cross-holder stale read) — sound as-is, but a candidate for revisiting only if reloads ever become hot.

ginderjeremiah:
Closes #1008

## Summary

The two entry points that reload the in-memory reference caches after an admin write had diverged. The request-time `AdminCacheReloadFilter` reloads all holders concurrently via `Task.WhenAll`, while the cross-instance background `CoalescingReferenceCacheReloader` reloaded them serially in a `foreach` and, on a mid-sweep failure, re-reloaded the already-succeeded holders on the next retry. Both paths were correctness-safe (no holder depends on another's reload order); this is an efficiency/consistency cleanup, not a bug fix.

This change aligns the background reloader with the request-time path:

- **Concurrent reload.** `ReloadAllAsync` now reloads all pending holders concurrently via `Task.WhenAll` (extracted into `ReloadConcurrentlyAsync`), matching the filter — closing the latency gap (N sequential DB round-trips becomes one concurrent fan-out) and the comment rationale is kept consistent with `AdminCacheReloadFilter`'s.
- **Retry only the unsucceeded holders.** The retry loop now tracks the still-pending set across attempts, so a failure on one holder no longer re-runs the build-then-swap for holders that already succeeded (wasted work — each succeeded reload already published its fresh snapshot).
- **Preserved semantics.** Retry/coalescing behaviour and cancellation-token propagation are unchanged: a cancellation rethrows promptly so the loop unwinds, and `Task.WhenAll` is awaited fully before inspecting results so one failure doesn't abandon the others mid-flight. A holder that throws synchronously is captured as a faulted task (`StartReload`) so it is retried uniformly rather than escaping the await.

No documentation change: the change is a narrowly-scoped efficiency cleanup and the documented cache-reload architecture (build-then-swap, retry-with-backoff, readers never disturbed) is unchanged.

## How I tested

- `dotnet build Game.sln -warnaserror` — full solution builds with **0 warnings, 0 errors**.
- Added two unit tests to `CoalescingReferenceCacheReloaderTests` and verified each genuinely guards the new behaviour (each fails against the prior implementation):
  - `RunAsync_Sweep_ReloadsAllHoldersConcurrently` — holders block at a shared gate that only opens once all have entered, so it completes iff the reloads run concurrently (a serial sweep deadlocks/times out — confirmed).
  - `RunAsync_FailureMidSweep_RetriesOnlyTheFailedHolderAndSkipsSucceededOnes` — asserts the succeeded holders are attempted exactly once and only the failed holder is retried (fails under the old retry-the-whole-sweep behaviour — confirmed).
- `dotnet test Game.Application.Tests` — full suite green: **342 passed, 0 failed** (the 8 reloader tests included).

Integration tests were not run for this change — it touches only in-process logic with no Redis/Postgres dependency, and the reloader's pub/sub glue is covered by existing `ReferenceCacheSynchronizerTests`.

---

### DIFF
--- a/Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs
+++ b/Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs
@@ -65,6 +65,47 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task RunAsync_Sweep_ReloadsAllHoldersConcurrently()
+        {
+            // Each holder blocks until every holder has entered its reload: a serial sweep would deadlock here
+            // (the second holder never enters until the first returns) and the gate would time out, so the
+            // gate opening proves the reloads ran concurrently rather than back-to-back.
+            var gate = new ConcurrentEntryGate(participants: 3);
+            var holders = new[] { new GatedCache(gate), new GatedCache(gate), new GatedCache(gate) };
+            var policy = new ReferenceCacheReloadPolicy(TimeSpan.Zero, maxAttempts: 1, baseDelay: TimeSpan.Zero);
+            await using var harness = new Harness(policy, holders);
+
+            harness.Reloader.NotifyChanged();
+
+            await WaitUntilAsync(() => holders.All(h => h.CompletedReloads == 1), "every holder to reload concurrently");
+        }
+
+        [Fact]
+        public async Task RunAsync_FailureMidSweep_RetriesOnlyTheFailedHolderAndSkipsSucceededOnes()
+        {
+            // The middle holder fails its first attempt; the others succeed. The retry must re-run only the
+            // failed holder, never the already-succeeded ones (re-reloading a succeeded build-then-swap is
+            // wasted work).
+            var first = new RecordingCache();
+            var failing = new RecordingCache { FailFirstAttempts = 1 };
+            var last = new RecordingCache();
+            var policy = new ReferenceCacheReloadPolicy(TimeSpan.Zero, maxAttempts: 3, baseDelay: TimeSpan.Zero);
+            await using var harness = new Harness(policy, first, failing, last);
+
+            harness.Reloader.NotifyChanged();
+
+            await WaitUntilAsync(() => failing.CompletedReloads == 1, "the failed holder to succeed on retry");
+            // The succeeded holders ran exactly once; only the failed holder was attempted again.
+            Assert.Equal(1, first.Attempts);
+            Assert.Equal(1, last.Attempts);
+            Assert.Equal(2, failing.Attempts);
+            Assert.Equal(1, first.CompletedReloads);
+            Assert.Equal(1, last.CompletedReloads);
+            Assert.Single(harness.Logs.Entries, e => e.Level == LogLevel.Warning);
+            Assert.DoesNotContain(harness.Logs.Entries, e => e.Level == LogLevel.Error);
+        }
+
+        [Fact]
         public async Task RunAsync_TransientFailure_RetriesAndCompletesTheSweep()
         {
             var cache = new RecordingCache { FailFirstAttempts = 1 };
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Align background reference-cache reload with the request-time path (#1008)

**Verdict:** Approve — does exactly what #1008 asked for, the concurrency/retry semantics are correct, and the new tests genuinely guard the new behaviour. (Posted as a `COMMENT` because I'm the PR author and GitHub won't accept a self-`APPROVE`; treat this as an approval.)

This is a tidy, well-scoped efficiency/consistency cleanup. I worked through the concurrency carefully and it holds up:

- `Task.WhenAll` is awaited to completion before the catch inspects results, so by the time `ReloadConcurrentlyAsync` walks `reloads[i].IsCompletedSuccessfully`, every task is guaranteed terminal — the still-pending computation is reliable, and one fault never abandons the others mid-flight.
- Cancellation rethrows promptly (`when (cancellationToken.IsCancellationRequested)`), matching the existing `RunAsync_CancelledMidSweep_*` test, so the loop still unwinds on shutdown.
- `StartReload` folding a synchronous throw into a faulted task is the right call here — it keeps a misbehaving holder lined up by index and retried uniformly instead of escaping the await. (Worth noting this makes the background path slightly *more* robust than `AdminCacheReloadFilter`, which has no such guard — fine, since the filter doesn't retry, so I'd leave it as-is.)
- Edge cases check out: empty `caches` → `Task.WhenAll([])` succeeds → returns immediately; final-attempt error path logs once and exits cleanly.

No doc change needed — `backend.md`'s cache-reload section ("a failed background sweep is retried with backoff and never disturbs readers") still describes the behaviour exactly.

### Test coverage

Solid and exactly to the project's classical/in-memory style for this class:
- `RunAsync_Sweep_ReloadsAllHoldersConcurrently` — the shared `ConcurrentEntryGate` proves concurrency by construction (a serial sweep deadlocks and times out). Good design.
- `RunAsync_FailureMidSweep_RetriesOnlyTheFailedHolderAndSkipsSucceededOnes` — pins the retry-only-pending invariant precisely (succeeded holders attempted once, failed holder twice, exactly one warning, no error).

No frontend/battle-logic surface, so no mirrored-test obligation. Integration tests correctly skipped — this is in-process logic with no Redis/Postgres dependency.

### What's good

Extracting `ReloadConcurrentlyAsync`/`ReloadFailure` keeps the retry loop readable, the rationale comment now mirrors the filter's wording (the consistency #1008 was about), and the partial-retry tracking removes real wasted work without changing the coalescing/backoff contract.

No blocking items, no should-fixes, no nits. Will merge once CI is green.

---

PR #1025, opened 2026-06-20
Reject duplicate-key admin change sets gracefully (#1007)

### LINKED ISSUE #1007: Duplicate-key admin change sets throw an opaque EF 500 instead of a graceful failure
## Summary

The admin Content-Authoring write path has an inconsistency in how it defends against a malformed change set that names the same key more than once. `ChildCollectionReconciler` explicitly guards this case and returns a graceful business failure, but the sibling `ChangeSetProcessor` / `AttributeChangeSetProcessor` paths do **not**, so a duplicate-keyed batch double-tracks an entity key in EF and surfaces as an unhandled exception (HTTP 500) rather than a 400 with an actionable message.

## Detail

`ChildCollectionReconciler.Reconcile` (`Game.DataAccess/Repositories/Admin/ChildCollectionReconciler.cs:46-53`) rejects a desired set containing duplicate keys up front:

```csharp
// A full desired set must not name the same child twice ... double-tracks the same entity as
// Modified, which EF rejects. The set is malformed input, so reject the whole write up front.
var desiredKeys = new HashSet<TKey>();
foreach (var item in desired)
{
    if (!desiredKeys.Add(desiredKey(item)))
    {
        return AdminSaveResult.Failure($"The submitted {resourceName} set contains duplicate entries.");
    }
}
```

The change-set processors carry no equivalent guard:

- `ChangeSetProcessor.Apply` (`ChangeSetProcessor.cs`) dispatches each `Change<T>` straight to its `add`/`edit`/`delete` handler with no de-duplication.
- `AttributeChangeSetProcessor.Apply` (`AttributeChangeSetProcessor.cs:44-68`) de-dups *Adds* via `existingKeys.Add(...)` (a repeated Add routes to Update), but:
  - **Two `Add`s of the same key** → one `entityStore.Insert(...)` then one `entityStore.Update(...)` on two distinct CLR instances with the same key → EF *"another instance with the same key value is already being tracked."*
  - **Two `Edit`s of the same existing key** → both pass the `existingKeys.Contains(...)` check → two `entityStore.Update(...)` calls on distinct instances → same EF tracking conflict.
  - **Two `Delete`s of the same key** → two `entityStore.Delete(...)` calls → same conflict.

The same gap applies to any other consumer of `ChangeSetProcessor` that maps each change to an entity-store op (e.g. the skill-effects setter in `AdminSkills`). Because the exception escapes mid-batch, the failure is an opaque 500 instead of the unified `AdminSaveResult.Failure` → 400 the reconciler produces, and the message gives the author nothing to act on.

## Likelihood / severity

This requires a malformed client delta — the well-behaved Workbench client computes its change set against a freshly-read baseline and won't normally produce duplicate keys. So there is no user-facing gameplay effect today. It's flagged as a **robustness/consistency** gap: the project already established (in the reconciler) that a duplicate-key set is malformed input that should fail gracefully, and the change-set processors should hold to the same standard rather than 500-ing.

## Suggested fix

Add an up-front duplicate-key rejection to `ChangeSetProcessor.Apply` (or a shared helper it calls), returning `AdminSaveResult.Failure(...)` when a key appears more than once in the batch — mirroring `ChildCollectionReconciler`. This needs a `Func<T, TKey>` key selector supplied by the caller (attribute change sets key on `AttributeId`, skill effects on `Id`, the identity-level batches on the primary id). Deletes already sort first and nothing is applied before the loop completes for the identity-level case, but for safety the dedup should run before any handler is invoked so the batch stays atomic.

Add a unit/integration test covering a duplicate-keyed Add/Edit/Delete batch asserting a graceful failure result rather than a thrown exception.

## Relationship to #965

Distinct from #965 (fold the admin Edit-*existence* guard into `ChangeSetProcessor`): that addresses an Edit of a record that doesn't exist; this addresses two changes naming the *same* key within one batch. Both touch the same processor, so they could reasonably be tackled together.

ginderjeremiah:
Closes #1007

## Summary

The admin Content-Authoring write path was inconsistent in defending against a malformed change set that names the same key more than once. `ChildCollectionReconciler` rejected a duplicate-keyed desired set up front, but the sibling `ChangeSetProcessor` and `AttributeChangeSetProcessor` did not — so a duplicate-keyed batch double-tracked an entity key in EF (*"another instance with the same key value is already being tracked"*) and surfaced as an opaque HTTP 500 instead of a graceful 400 with an actionable message.

This adds a consistent up-front duplicate-key guard that returns `AdminSaveResult.Failure(...)`, matching the reconciler's pattern and message style.

## What changed

- **`ChangeSetProcessor.Apply`** gains an optional `key` selector + `resourceName`. When `key` is supplied, the batch is rejected before any handler runs if a key appears more than once among the **value-tracked Edit/Delete** changes. Adds are deliberately excluded: an insert's wire id is a store-generated sentinel (e.g. identity records post `id: -1`, skill effects `id: 0`), so two distinct Adds share that sentinel and would falsely collide — but they never conflict in EF because each insert gets a fresh generated key.
- **`AttributeChangeSetProcessor.Apply`** dedups the **whole batch** (including Adds) and now returns the `AdminSaveResult`. An attribute key (`AttributeId`) is a genuine composite-PK key for every change type — including an Add — so any repeated key across the batch is malformed.
- A small shared helper (`ChangeSetProcessor.HasDuplicateKey` / `DuplicateFailure`) keeps the dedup logic in one place (DRY).
- All callers updated to pass the key selector + resource name: identity-level saves (enemies, items, item mods, skills, zones, challenges, tags), the skill-effects and mod-slot setters, and the three attribute setters (which now return the processor result).

The well-behaved Workbench client computes its delta against a freshly-read baseline and won't normally produce duplicate keys, so there's no user-facing gameplay effect — this is a robustness/consistency fix holding the change-set processors to the standard the reconciler already set.

## How I tested

- `dotnet build Game.sln -warnaserror` — clean, **0 warnings, 0 errors**.
- **Unit tests** (`ChangeSetProcessorTests`, `AttributeChangeSetProcessorTests`): added coverage for duplicate Add, duplicate Edit, duplicate Delete, and cross-type (Edit+Delete) combos for both processors, asserting a graceful `Failure` (no thrown exception) and that nothing is applied; plus the Add-sentinel exclusion and distinct-key success paths. All 24 pass.
- **Integration tests** (held the shared-infra file lock briefly via a class filter): added `SaveEnemies_DuplicateEditKeysInBatch_ReturnsFailureWithoutThrowing` (identity-level `ChangeSetProcessor`) and `SetAttributes_DuplicateKeyInBatch_ReturnsFailureWithoutThrowing` (attribute processor), proving the graceful failure against a real Postgres without throwing. The 2 touched classes (13 tests) pass.

## Follow-ups / concerns

- `HasDuplicateKey` keys on `Func<T, object>`, which boxes int/enum keys into a `HashSet<object>`. Boxed value types compare by value, so this is correct; the small allocation is negligible for admin batches (not a hot path) and keeps the API from sprouting a `TKey` generic everywhere.
- With the up-front guard, the attribute processor's Add-upsert `else → Update` branch is now only reachable for a legitimate Add of an attribute the owner *already has* (covered by `SetAttributes_AddOfAlreadyPresentAttribute_Upserts`), never an intra-batch duplicate.

---


Game.DataAccess/Repositories/Admin/ChangeSetProcessor.cs
```diff
@@ -25,15 +25,32 @@ namespace Game.DataAccess.Repositories.Admin
...
             Action<T> add,
             Action<T> edit,
-            Action<T>? delete = null) where T : IModel
+            Action<T>? delete = null,
+            Func<T, object>? key = null,
+            string? resourceName = null) where T : IModel
         {
+            // Only Edit/Delete are value-tracked (Adds get a store-generated key), so they alone can collide.
+            if (key is not null
+                && HasDuplicateKey(changes, c => c.ChangeType != EChangeType.Add, key))
```
ginderjeremiah, line 49: Nit (consistency, non-blocking): with `key` supplied, `changes` is now enumerated twice — once here in `HasDuplicateKey` and again in the `OrderByDescending` below. `ChangeSetProcessor.Apply` still takes `IEnumerable<Change<T>>`, whereas `ChildCollectionReconciler` deliberately tightened its params to `IReadOnlyCollection` for exactly this reason (its comment: "a deferred/lazy source would re-run — and could yield a different sequence the second time"). Every current caller passes a materialized `IReadOnlyList`, so there's no bug today, but for consistency with the reconciler's documented rationale consider tightening this parameter to `IReadOnlyCollection<Change<T>>` (or `IReadOnlyList`). Take it or leave it.

---

ginderjeremiah (commented):
## Review: Reject duplicate-key admin change sets gracefully (#1007)

**Verdict:** Approve (posted as a comment because GitHub won't let me formally approve my own PR) — solves the issue cleanly, mirrors the existing `ChildCollectionReconciler` pattern, and is thoroughly tested on both sides. Only one non-blocking nit.

The change closes the consistency gap the issue describes: `ChildCollectionReconciler` already rejected duplicate-keyed desired sets up front, but `ChangeSetProcessor`/`AttributeChangeSetProcessor` would double-track an entity key in EF and surface an opaque 500. The new up-front guard returns a graceful `AdminSaveResult.Failure(...)` instead, matching the reconciler's message style.

A few things I verified rather than took on faith:

- **The Add exclusion is correct for the identity-level processor.** An insert's wire id is a store-generated sentinel (`-1`/`0`), so two distinct Adds share that sentinel and would falsely collide — but they never conflict in EF because each gets a fresh generated key. Excluding Adds (`c.ChangeType != EChangeType.Add`) is the right call, and the `Apply_DuplicateAddKey_IsNotRejected` test pins it.
- **The attribute processor correctly dedups the *whole* batch including Adds.** `AttributeId` is a genuine composite-PK key for an Add, so a repeated key there really would double-track. This also fixes a latent bug: the old "duplicate Add upserts (insert-then-update)" path would have thrown the same EF "already tracked" error in real EF — the old unit test only passed because `RecordingEntityStore` doesn't enforce tracking. Flipping that test to `RejectsGracefullyWithoutApplying` is the right behavior change.
- **No dead code left behind.** The attribute `add` handler keeps its upsert branch, but it's still reachable for the legitimate case (an Add of an attribute the owner *already has*); the now-unreachable "added earlier in this same batch" path and its stale comment were correctly removed.
- **All callers updated** — every `ChangeSetProcessor.Apply`/`AttributeChangeSetProcessor.Apply` call site now passes the `key` selector + `resourceName` (and the three attribute setters now return the processor result instead of a hardcoded `Success`).

### Nits

- `ChangeSetProcessor.cs:49` — with `key` supplied, `changes` is enumerated twice (`HasDuplicateKey` + `OrderByDescending`); the reconciler tightened its params to `IReadOnlyCollection` for exactly this reason. No bug today (all callers pass materialized lists), purely a consistency suggestion. See inline.
- The `Func<T, object>` key selector boxes int/enum keys into a `HashSet<object>` — already noted in your "Follow-ups / concerns". Negligible on admin write paths; fine to leave or fold into a follow-up. Not worth its own issue IMO.

### Test coverage

Solid and matches the project's testing rules. Unit tests cover duplicate Edit/Delete/cross-type for `ChangeSetProcessor`, the Add-exclusion and distinct-key success paths, the no-key-selector passthrough, and the full Add/Edit/Delete/cross-type matrix for `AttributeChangeSetProcessor`. Integration tests exercise both processors against real Postgres (`SaveEnemies_DuplicateEditKeysInBatch_...`, `SetAttributes_DuplicateKeyInBatch_...`), proving the graceful failure without a thrown exception. No frontend/battle-logic surface, so no mirrored tests required.

### What's good

DRY shared helper (`HasDuplicateKey`/`DuplicateFailure`) keeps the guard in one place; the `backend-admin.md` doc was updated to record the new standard; and the XML-doc rationale on both processors is genuinely useful for explaining *why* Adds are treated differently between the two. Nice, surgical fix.

I'm not auto-merging: CI is still running and there's an open nit. Once CI is green, this is good to merge at your discretion.

---

PR #1026, opened 2026-06-20
Align ICacheService value-setter nullability with impl behaviour

### LINKED ISSUE #1014: Align nullability of the non-expiry GetSet overload with ICacheService
## Problem

#954 / #1013 fixed the value-nullability mismatch on the **expiry** `GetSet` overload (`ICacheService` declares the value non-null, but the impl took `string?` and could feed a nil into the inline Lua `SET`). The **non-expiry** overload still has the same class of mismatch:

- `Game.Abstractions/Infrastructure/ICacheService.cs:14` — `Task<string?> GetSet(string key, string value, ...)` (value **non-null**)
- `Game.Infrastructure/Cache/Redis/RedisService.cs:52` — `public async Task<string?> GetSet(string key, string? value, ...)` (value **nullable**), routing to `Redis.StringGetSetAsync(key, value)`
- `RedisService.cs:58-62` — the generic `GetSet<T>` passes `value?.Serialize()` (possibly null) into that overload.

So the interface's non-null contract is unenforced and a null can flow into `StringGetSetAsync`. This is lower-risk than the expiry path (it's a single `GETSET`, not a Lua script), but the impl/interface disagree and the generic relies on the nullable behaviour.

## Fix

Pick one null semantic for the non-expiry `GetSet`, the same way the expiry overload was resolved:

- Align the impl signature to the interface's **non-null** `string value` and decide what the generic `GetSet<T>` should do when `value` is null (e.g. route to a delete to match the other setters, or guard+throw), **or**
- If a null value is genuinely meant to delete here, make that explicit on both the interface and the impl and document it.

Whichever is chosen, make the interface and impl nullability agree and add/adjust a unit test pinning the null path.

## Notes

- Pre-existing; surfaced while reviewing #1013. Out of scope for that PR (which was scoped to the expiry overload per #954 item 2).
- No user-facing behaviour change expected — purely a contract/consistency cleanup.

### LINKED ISSUE #1015: Align non-expiry GetSet/Set nullability with the ICacheService contract
## Problem

Follow-up from #954 / PR #1013. That PR fixed the nullability mismatch on the **expiry** overload (`GetSet(key, value, expiry)`) by aligning the impl to the non-null `ICacheService` contract. The **non-expiry** overload — and the other value-accepting setters — still carry the same latent mismatch:

- `ICacheService` declares the value **non-null**: `GetSet(string key, string value, …)`, `Set(string key, string value, …)`, `SetAndForget(string key, string value, …)`.
- The `RedisService` impls accept `string? value`, and the generic overloads pass a possibly-null value into them, e.g. `GetSet<T>` calls `GetSet(key, value?.Serialize(), …)` and `Set<T>` calls `Set(key, value?.Serialize(), …)`.

So the declared contract says non-null while the impls (and the generic callers) actually rely on null-as-delete. The contract is unenforced and the interface understates nullability.

## Suggested direction

Pick one and apply it consistently across the value-accepting setters:

1. **Make the contract honest** — declare `string?` on the interface for the setters whose impls already treat null as delete (`GetSet` non-expiry, `Set`, `SetAndForget`), and document the null-means-delete semantic. This matches the de-facto behavior and the generic overloads. (Ripples to the `ICacheService` test stubs in `SocketCommandProcessorTests`.)
2. Or **enforce non-null** the way the expiry overload now does, and stop the generic overloads from passing null.

Option 1 is likely the lower-friction, more truthful fix since the null-as-delete behavior is already depended upon.

## Notes

- Purely a consistency/correctness-of-contract cleanup; no user-facing behavior change.
- Out of scope for #954, which targeted specifically the Lua/expiry crash path.

ginderjeremiah:
Closes #1015
Closes #1014

#1014 is a duplicate of #1015 (the same non-expiry `GetSet` nullability mismatch), folded into this single PR.

## Summary

Finishes the contract-honesty work #1013 started for the expiry `GetSet` overload by aligning the rest of the value-accepting setters on `ICacheService` with what the `RedisService` impl actually does. The setters split into two camps, and the interface now states each truthfully:

- **`Set` and `SetAndForget`** (plain and expiry overloads): declared **`string?`**, documenting the de-facto **null-means-delete** semantic their generic `<T>` overloads already rely on (a null `T` serializes to a null string, which Redis's `StringSet` treats as a delete). The impls already accepted `string?` and deleted on null, so only the interface and the test stubs changed.
- **Non-expiry `GetSet`**: kept **non-null**. Per the issue's scope guard I verified against the impl before changing the signature — and the underlying `StringGetSet` **rejects** a null value (`ArgumentException: "A null value is not valid in this context"`) rather than deleting it. So there is no real null-as-delete path here; the honest contract is non-null, matching how #1013 resolved the expiry overload. The impl signature and the generic `GetSet<T>` were aligned to non-null accordingly.
- The deliberately-non-null **expiry `GetSet`** overload from #1013 was left untouched.

This is a contract/consistency cleanup with **no user-facing behaviour change**. A brief note recording the two-camp setter semantic was added to `docs/backend.md`.

### Note on the original premise

The issues assumed the non-expiry `GetSet` also relied on null-as-delete. It does not — `StringGetSet` throws on null — so applying "make the contract honest" correctly meant leaving `GetSet` non-null (and fixing its generic overload to stop passing a possibly-null value), while only `Set`/`SetAndForget` became nullable.

## How tested

- Full solution build: **0 warnings, 0 errors** (nullable analysis clean).
- `Game.Api.Tests` `SocketCommandProcessorTests` (the stub consumers): **9/9 pass**.
- `Game.Application.Tests` `RedisServiceIntegrationTests` against real Redis: **7/7 pass** — added integration tests pinning the null path: `Set`/`Set`+expiry/`SetAndForget`/`SetAndForget`+expiry each delete the key on a null value, and the non-expiry `GetSet` stores-and-returns-prior with a non-null value. (Integration was the right level here: per `docs/backend.md`, `RedisService` is a thin out-of-process adapter exercised through the DI-resolved interface, not mocked, and null-as-delete is a genuine Redis behaviour.)

---


Game.Infrastructure/Cache/Redis/RedisService.cs
```diff
@@ -49,15 +49,17 @@ public RedisService(IConnectionMultiplexer multiplexer, ILogger<RedisService> lo
...
+            // Read-and-set is a write (it stores the new value), so it routes through ObserveWrite too. The value
+            // is required (non-null): the underlying GETSET has no null-means-delete path — it rejects a null
+            // value — so unlike Set/SetAndForget this overload matches ICacheService's non-null contract (#1015).
             return await ObserveWrite(Redis.StringGetSetAsync(key, value), cancellationToken);
         }
 
         public async Task<T?> GetSet<T>(string key, T value, CancellationToken cancellationToken = default)
         {
-            var val = await GetSet(key, value?.Serialize(), cancellationToken);
+            var val = await GetSet(key, value.Serialize(), cancellationToken);
```
ginderjeremiah, line 64: Nit (non-blocking): with the contract now non-null, `value.Serialize()` is right for the happy path, but since `T` is unconstrained the compiler won't stop a caller from passing a null reference here. When that happens, `JsonSerializer.Serialize<T>(null)` yields the literal string `"null"`, which gets stored as the value (and round-trips back to `default(T)` via `Deserialize`). That's a quiet divergence from the sibling `Set<T>`/`SetAndForget<T>`, where a null is `null`-means-delete. It's a contract violation either way so I wouldn't add a guard, but a one-line comment noting that null `T` is unsupported here (unlike `Set<T>`) would save a future reader the double-take. Not a blocker.

---

ginderjeremiah (commented):
## Review: Align `ICacheService` value-setter nullability with the impl (closes #1015 / #1014)

**Verdict:** Approve _(submitted as a comment because I'm the PR author, so GitHub won't take a gating verdict from me)_ — the contract is now honest, the split between the two setter camps is correct and well-documented, and the integration tests pin the behaviour that matters.

The standout here is that the PR didn't just rubber-stamp the issue's suggested direction. The issue assumed the non-expiry `GetSet` also relied on null-as-delete; the author verified against the impl, found that `StringGetSet` *rejects* null rather than deleting, and resolved it the truthful way — `Set`/`SetAndForget` become `string?` (null-means-delete, matching their generic overloads), `GetSet` stays non-null. The interface, the `RedisService` impl comments, `docs/backend.md`, and the test stubs all line up with that two-camp model, and the deliberately-non-null expiry `GetSet` from #1013 was correctly left alone.

### Nits

- `RedisService.cs:62` — the generic `GetSet<T>` now calls `value.Serialize()`; for an unconstrained `T`, a null reference silently serializes to the literal `"null"` rather than deleting (diverging from `Set<T>`/`SetAndForget<T>`). It's a contract violation to pass null here either way, so no guard needed, but a one-line comment would spare a future reader the double-take. (inline)

### Test coverage

Good. New integration tests cover the null-deletes-key path for `Set`, `Set`+expiry, `SetAndForget`, `SetAndForget`+expiry, plus the non-null store-and-return for the non-expiry `GetSet` — exercised through the DI-resolved interface against real Redis, which is the correct level per `docs/backend.md` (RedisService is a thin out-of-process adapter, null-as-delete is genuine Redis behaviour, not domain logic). The fire-and-forget tests correctly poll-until-deleted rather than racing a fixed delay. No battle logic is touched, so the dual-suite consistency rule doesn't apply. The only thing not pinned is the "GetSet rejects null" throw, but since the typed contract is now non-null there's no way to drive that path through the interface, so that's fine.

### What's good

- Correctly challenged the issue's premise instead of following it blindly, and documented the reasoning in the PR body and the impl comments.
- Clear, accurate XML docs on each setter spelling out the null semantic per camp.
- Concise `docs/backend.md` note recording the two-camp contract — exactly the kind of non-obvious design decision the docs are meant to capture.

Not merging on my end: I have an outstanding nit, and `test-backend` / `test-e2e (webkit)` were still in progress at review time. Once they're green this is good to go.

---

PR #1028, opened 2026-06-20
Extract a shared insert-if-missing helper for the write-behind unlock handlers (DRY)

### LINKED ISSUE #961: Extract a shared insert-if-missing helper for the write-behind unlock handlers (DRY)
## Problem

The recent run of "make write-behind upsert handlers atomically idempotent" fixes (#891/#916, #915/#924) left the same algorithm copy-pasted across several handlers: `AsNoTracking` existence check → return-if-exists → `Add` → `SaveChanges` → swallow the unique-violation. It appears verbatim in:

- `Game.DataAccess/PlayerUpdates/Handlers/ItemUnlockedHandler.cs`
- `Game.DataAccess/PlayerUpdates/Handlers/ModUnlockedHandler.cs`
- `Game.DataAccess/PlayerUpdates/Handlers/SkillUnlockedHandler.cs`
- and the catch blocks of `ModAppliedHandler` / `LogPreferenceChangedHandler`

differing only in entity type, key predicate, and the factory. The `GroupBy(...).ToDictionary(g => g.First())` dedup is likewise duplicated across `AttributeAllocationsChangedHandler` and `ProgressUpdatedHandler`.

Because the idempotency contract is the whole point of these handlers, having it expressed five times means a future fix (or a subtle divergence) has to be made in five places, and the contract can't be unit-tested in one spot.

## Suggested fix

Extract a single `InsertIfMissingAsync<TEntity>(ctx, existsPredicate, factory, ct)` (and a small `ToFirstByKey` helper for the dedup), collapsing the copies to one and making the insert-race idempotency contract testable once. Keep it narrowly scoped — this is a targeted DRY pass, not a broader data-tier refactor.

## Notes

There is broader structural duplication in this layer too (the six reference cache holders are near-identical; the mappers repeat per-field projections; the admin Edit-existence guard is copy-pasted across seven repos — that last one is called out in the admin-write-path issue). Those are lower-value and the systems are still evolving, so they're noted here only for context rather than bundled into this change.

ginderjeremiah:
Closes #961

## Summary

The "make write-behind upsert handlers atomically idempotent" fixes left the same `AsNoTracking` existence check → return-if-exists → `Add` → `SaveChanges` → swallow-unique-violation algorithm copy-pasted across the unlock handlers, and the `GroupBy(...).First()` dedup duplicated across the batched upsert handlers. This is a targeted DRY pass that factors both out so the idempotency contract lives — and is tested — in one place.

Two narrowly-scoped helpers now live on a new shared `internal static` class, `PlayerUpdateHandlerExtensions`, in the `Handlers` folder (kept inside the data tier so it never surfaces above `Game.DataAccess`, consistent with the entity-isolation conventions in `docs/backend.md`):

- **`InsertIfMissingAsync(context, existsPredicate, factory)`** — an extension on `GameContext` that encapsulates the full insert-race idempotency contract. It resolves the `DbSet<TEntity>` via `context.Set<TEntity>()`, so the entity type is inferred from the predicate/factory and no `DbSet` needs to be threaded through.
- **`ToFirstByKey(keySelector)`** — the group-by-first dedup, returning a `Dictionary<TKey, TSource>`.

## Handlers collapsed

- **`ItemUnlockedHandler`, `ModUnlockedHandler`, `SkillUnlockedHandler`** now route through `InsertIfMissingAsync` (the skill handler keeps its `Selected = false, Order = 0` note in its factory).
- **`AttributeAllocationsChangedHandler`** and **`ProgressUpdatedHandler`** (its statistics, challenges, and tuple-keyed lookups) now use `ToFirstByKey`.

`ModAppliedHandler` and `LogPreferenceChangedHandler` are deliberately **left as-is**: their shapes differ from insert-if-missing (delete-then-insert; update-first-then-insert-then-re-update), and the only thing they share is the `catch ... when (ex.IsUniqueViolation())` clause — which is already factored into the `IsUniqueViolation()` extension. Folding them into `InsertIfMissingAsync` would distort the contract, so they were left out to keep the change narrowly scoped.

## How the idempotency contract is preserved

- The unique-violation detection is **unchanged** — every call site still keys off the existing `DbUpdateExceptionExtensions.IsUniqueViolation()` (`InnerException is PostgresException { SqlState: PostgresErrorCodes.UniqueViolation }`), so the EF/Npgsql signature is identical across all handlers. The helper reuses that same extension verbatim.
- The collapsed handlers reproduce the original sequence exactly: `AsNoTracking().AnyAsync(...)` → early return → `Add` → `SaveChangesAsync()` → swallow the unique violation.

## Testing

- **`InsertIfMissingIntegrationTests`** (new) proves the contract once at the helper level against a real DB: row-missing inserts; already-present short-circuits without re-running the factory; and a concurrent double-apply (parallelism 8, each on its own DI scope/`GameContext`) converges to a single row without throwing — exercising the unique-violation catch.
- **`ToFirstByKeyTests`** (new, unit) covers indexing, duplicate-key-keeps-first (the reason for group-by-first over `ToDictionary`), empty source, and composite keys.
- The existing **`PlayerUpdateHandlerIdempotencyIntegrationTests`** (25 tests) still pass unchanged, confirming each refactored handler behaves identically.
- Full solution builds clean with `-warnaserror` (0 warnings).

Test results: helper integration tests 3/3 pass, `ToFirstByKey` unit tests 4/4 pass, existing handler idempotency suite 25/25 pass.

## Note / follow-up

The handler path does not currently thread a `CancellationToken` (the `IPlayerUpdateHandler.HandleAsync` signature and the `SaveChangesAsync()` calls take none), so the issue's suggested `ct` parameter was intentionally omitted to stay scoped to a DRY pass — adding one would mean changing the interface and dispatcher, a broader refactor. Worth a follow-up if cancellation should be plumbed through the synchronizer's drain handlers.

---

### DIFF
--- a/Game.Application.Tests/DataAccess/InsertIfMissingIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/InsertIfMissingIntegrationTests.cs
@@ -1,0 +1,135 @@
+using Game.DataAccess.PlayerUpdates.Handlers;
+using Game.Infrastructure.Database;
+using Game.Infrastructure.Entities;
+using Game.TestInfrastructure.Base;
+using Game.TestInfrastructure.Fixtures;
+using Game.TestInfrastructure.Helpers;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Game.Application.Tests.DataAccess
+{
+    /// <summary>
+    /// Verifies the shared <see cref="PlayerUpdateHandlerExtensions.InsertIfMissingAsync"/> contract once, at
+    /// the helper level: the existence-check-then-insert is idempotent under the queue's at-least-once read,
+    /// converging to a single row whether re-applied sequentially or as a concurrent double-apply that provokes
+    /// the unique-violation race the catch absorbs. The collapsed unlock handlers all route through this helper,
+    /// so their per-handler idempotency tests (see <see cref="PlayerUpdateHandlerIdempotencyIntegrationTests"/>)
+    /// exercise the same path and pin the wiring. Exercised against <see cref="UnlockedItem"/> as a
+    /// representative entity carrying a (player, item) unique key.
+    /// </summary>
+    [Collection("Integration")]
+    public class InsertIfMissingIntegrationTests : ApplicationIntegrationTestBase
+    {
+        // Enough concurrent inserts that several pass the existence check before any insert commits, so the
+        // unique-violation catch path is exercised rather than only the fast existence-check no-op.
+        private const int Parallelism = 8;
+
+        public InsertIfMissingIntegrationTests(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper)
+            : base(containers, testOutputHelper) { }
+
+        [Fact]
+        public async Task RowMissing_InsertsTheRow()
+        {
+            var playerId = await SeedPlayerAsync();
+            var itemId = await SeedItemAsync();
+
+            await InsertUnlockedItemAsync(playerId, itemId);
+
+            Assert.Equal(1, await CountUnlockedItemsAsync(playerId, itemId));
+        }
+
+        [Fact]
+        public async Task RowAlreadyPresent_DoesNotInsertOrInvokeFactoryTwice()
+        {
+            var playerId = await SeedPlayerAsync();
+            var itemId = await SeedItemAsync();
+
+            await InsertUnlockedItemAsync(playerId, itemId);
+
+            // The second call must short-circuit on the existence check and never run the factory.
+            var factoryInvoked = false;
+            using (var scope = CreateScope())
+            {
+                var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+                await context.InsertIfMissingAsync(
+                    (UnlockedItem ui) => ui.PlayerId == playerId && ui.ItemId == itemId,
+                    () =>
+                    {
+                        factoryInvoked = true;
+                        return new UnlockedItem { PlayerId = playerId, ItemId = itemId };
+                    });
+            }
+
+            Assert.False(factoryInvoked);
+            Assert.Equal(1, await CountUnlockedItemsAsync(playerId, itemId));
+        }
+
+        [Fact]
+        public async Task AppliedConcurrently_InsertsOneRowWithoutThrowing()
+        {
+            var playerId = await SeedPlayerAsync();
+            var itemId = await SeedItemAsync();
+
+            // Each insert runs through its own scope/context, mirroring the synchronizer's per-event scope, so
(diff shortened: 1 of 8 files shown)


ginderjeremiah (commented):
## Review: Extract a shared insert-if-missing helper for the write-behind unlock handlers (#961)

**Verdict:** Approve — clean, correctly-scoped DRY pass; the idempotency contract is reproduced verbatim and now tested in one place. _(Posted as a comment because I'm the PR author and GitHub won't accept a self-approval; treat this as an approve.)_

### What's good

- **The scoping is exactly right.** I checked every handler that still references `IsUniqueViolation()`. `ItemEquippedHandler`, `ItemFavoriteChangedHandler`, and `SelectedSkillsChangedHandler` use the load-then-upsert + catch-and-retry-once shape (mirroring `AttributeAllocationsChangedHandler`), and `ModAppliedHandler`/`LogPreferenceChangedHandler` are delete-then-insert / update-then-insert. None of them fit the insert-if-missing contract, so folding them in would have distorted it — leaving them out was the correct call, and the only thing they genuinely share (`catch ... when (IsUniqueViolation())`) is already factored.
- **The contract is preserved byte-for-byte.** `InsertIfMissingAsync` reproduces `AsNoTracking().AnyAsync(predicate)` → early return → `Add(factory())` → `SaveChangesAsync()` → swallow on `IsUniqueViolation()`, and resolves the `DbSet` via `context.Set<TEntity>()` so the entity type is inferred from the predicate/factory. Passing the predicate as `Expression<Func<TEntity,bool>>` keeps it translating to SQL; `ToFirstByKey` taking a plain `Func` is correct since it runs over already-materialized lists.
- **Layering is respected.** The helper is `internal static` inside `Game.DataAccess/.../Handlers`, so the entity seam never surfaces above the data tier — consistent with the entity-isolation conventions in `docs/backend.md`.

### Test coverage

Meets the project's testing rules well. The DB-touching helper is covered by an integration test (`InsertIfMissingIntegrationTests`) — row-missing insert, already-present short-circuit that also asserts the factory is **not** re-invoked, and a parallelism-8 concurrent double-apply each on its own DI scope/`GameContext` that exercises the unique-violation catch and converges to one row. The pure `ToFirstByKey` is covered classically as a unit test (indexing, duplicate-key-keeps-first, empty, composite keys). The existing 25-test handler idempotency suite continues to pin per-handler behaviour through the collapsed path. No battle-logic parity concern (data tier only).

### Follow-up

- The `CancellationToken` the author deliberately deferred is tracked in **#1029** — framed as a design decision (the write-behind drain may legitimately *not* want to cancel saves on shutdown) rather than a mechanical param add, so nothing is lost.

CI is green (`all-checks-passed`, `test-backend`, `verify-backend`, e2e ×3; frontend skipped). Merging.

---

PR #1030, opened 2026-06-20
Confirm and pin the zero-based-id save fixup PK branch (#1003)

### LINKED ISSUE #1003: Confirm whether the Modified-record-0 PK branch of the zero-based-id save fixup is still reachable in EF Core 10
EDIT: **Thoroughly** investigate cases where an update may target or be related to a record with id 0 for reference data. If there is any scenario that would cause issues -> Keep the code and denote why it is there. Do NOT only investigate cases that are actively used by the code base. I'm 99% sure this was required for how the codebase interacted with EF Core in the past, and I doubt it is something that was changed in v10. Even if we aren't currently triggering the branch with how we interact with EF, I don't want to remove it if it is still technically needed for complete correctness.

## Context

Follow-up from #938 / PR #1002, which optimized `GameContext.SaveChangesAsync`'s zero-based-identity fixup (caching the per-entity-type PK/FK targets instead of re-walking metadata per save). The optimization preserved both branches exactly:

- **FK branch** — a non-nullable `int` FK pointing at record 0 of a store-generated zero-based principal. Confirmed reachable: an in-process unit test shows EF marks such an FK temporary on `Add`, and the fixup forces the real `0`.
- **PK branch** — a `Modified` zero-based record whose `Id == 0` (the identity seed). The fixup forces the temporary key back to `0`.

## The question

While writing the unit tests, I found that in **EF Core 10**, attaching a zero-based record with `Id == 0` as `Modified` does **not** mark its key temporary at the point the fixup runs (the fixup runs before `base.SaveChangesAsync`, where `DetectChanges` would run). So the PK branch could be **defensive / no longer reachable** — but I did not want to remove it as part of a performance-only change, so it was preserved as a guarded no-op (`ForceZero` only acts when the key is actually temporary).

## Proposed work

1. Add an **integration test** that edits record 0 of a zero-based reference set (e.g. enemy/item/challenge id 0) via the admin path and asserts the `UPDATE` targets the right row.
2. With that test in place, determine whether the PK branch ever fires:
   - Temporarily remove the PK branch and run the integration suite.
   - If nothing fails, the branch is dead in EF Core 10 → remove it (and trim the now-unused `KeyProperty`/`IZeroBasedIdentityEntity { Id: 0 }` handling) and keep the integration test as the regression guard.
   - If something fails, the branch is needed → keep it and add the failing scenario as explicit coverage.

No user-facing change either way; this is purely confirming reachability and removing dead code if so.

_Found while implementing #938._

ginderjeremiah:
Closes #1003

## Conclusion: KEEP the PK branch

Per the owner's binding guidance, I defaulted to keeping the code and investigated thoroughly — including scenarios the codebase does not currently exercise. The PK branch is retained as a defensive guard, with a clear comment explaining exactly why, plus an integration test pinning the record-0 edit behavior. I did **not** remove anything.

## Investigation

The fixup's PK branch forces the literal `0` back onto a non-`Added` record's key when EF marked it temporary (having read `Id == 0`, the identity seed, as an unset store-generated value). The question was whether that precondition (non-`Added` entry + temporary PK) is reachable in EF Core 10.

I empirically probed EF Core 10.0.9 (throwaway in-process test, since the change tracker runs without a DB connection; removed before commit) across every way a reference-data record with `Id == 0` can be touched or related to:

| Scenario | Result |
|---|---|
| `Entry.State = Modified`, `Id == 0` (the admin path via `EntityStore.Update`) | key **NOT temporary**; PK branch no-ops |
| `ctx.Update(entity)`, `Id == 0` | EF infers state **Added** (not Modified) + temporary key — PK branch excludes `Added` |
| `ctx.Attach(entity)`, `Id == 0` | EF makes it **Added** + temporary key — same as above |
| `ctx.Add(entity)`, `Id == 0` | **Added** + temporary (intended; left alone so the real id is store-generated) |
| Force `IsTemporary = true` on a `Modified` record's key | **throws** "part of a key … cannot be modified" |
| `Add` (temp key) then flip to `Modified` / `Deleted` / `Unchanged` | **throws** "has a temporary value while attempting to change the entity's state" |
| Delete record 0 (`Unchanged → Remove`, `Attach → Remove`, `Update → Remove`) | key non-temporary (literal 0) |

**Key EF Core 10 invariant:** a property that is part of a primary key cannot be temporary while the entry is `Modified`, `Deleted`, or `Unchanged`. EF rejects both *creating* that state (setting `IsTemporary`) and *transitioning into* it (flip out of `Added`) — and it throws at the state-change point, before `SaveChanges`/the fixup runs. The only state where `Id == 0` yields a temporary key is `Added`, which the PK branch deliberately excludes (`entry.State is not EntityState.Added`).

So the PK branch is **unreachable as we interact with EF today**, and `ForceZero` is a guaranteed no-op for any non-`Added` entry regardless. The historical (pre-#1002) implementation had the identical precondition (`entry.State is not EntityState.Added` + only acting on already-temporary properties), so this is not a regression introduced by the caching optimization.

I also ran the issue's "Proposed work" probe directly: with the PK branch disabled, the new integration test (and the rest of `AdminEnemiesIntegrationTests`) still passes — empirically confirming the admin record-0 edit path does not depend on it. The FK branch, by contrast, is genuinely live (an `int` FK referencing record 0 of a store-generated principal is marked temporary on `Add`).

## Why keep it anyway

The branch costs nothing at runtime (a guarded no-op) and protects a correctness-critical invariant: a record-0 edit must never silently `UPDATE` the wrong row. I can't exercise older EF versions here to disprove the owner's recollection that it was once required, and I won't remove a correctness guard on the strength of a single EF-version probe. Keeping it is the conservative, low-risk choice the owner asked for.

## Changes

- **`GameContext.cs`** — replaced the PK-branch comment with a thorough explanation of why it exists, that it is currently an unreachable no-op in EF Core 10, and that it is retained deliberately as a defensive guard (with a pointer to the pinning test).
- **`AdminEnemiesIntegrationTests.cs`** — new `SaveEnemies_EditsRecordZero_UpdatesTheCorrectRow`: seeds two enemies into a `RESTART IDENTITY`-truncated DB (so they land at ids 0 and 1), edits record 0 through the real admin path + unit of work, and asserts the `UPDATE` changed only row 0 while row 1 is untouched (proving the edit wasn't mis-resolved by a default/temporary key).
- **`GameContextZeroBasedIdentityFixupTests.cs`** — corrected the in-process unit test's comment, which previously implied EF marks the Modified record-0 key temporary; it does not.
- **`docs/backend-admin.md`** — extended the existing zero-based-identity gotcha note to cover the `GameContext` save fixup (FK branch live, PK branch a retained defensive no-op).

## Testing

- Full solution build: **0 warnings, 0 errors**.
- `GameContextZeroBasedIdentityFixupTests` (in-process unit): **6/6 pass**.
- `AdminEnemiesIntegrationTests` (integration, against shared Postgres + Redis under the cross-process lock): **8/8 pass**, including the new test.
- Verification: with the PK branch disabled, the new integration test still passes — confirming it is a no-op for the admin path in EF Core 10. Branch restored before commit.

## Follow-ups / concerns

- None blocking. If the owner is confident the branch was never needed and wants to slim the surface, a future change could drop the PK branch plus the now-only-FK `KeyProperty` handling — but that should be a deliberate decision, not bundled into this conservative-by-design change. I recommend leaving it as-is.

---


ginderjeremiah (commented):
## Review: Confirm & pin the zero-based-id save fixup PK branch (#1003)

**Verdict:** Approve _(submitted as a comment because I'm the PR author, so GitHub won't accept a gating event from me)_ — the change does exactly what the issue's binding guidance asks: thoroughly investigate, **keep** the guard, document precisely why, and pin record-0 edit behavior with a test. No blocking findings, no should-fixes.

### On the central decision
The issue's top edit is binding ("Keep the code and denote why… I don't want to remove it if it is still technically needed for complete correctness"), which overrides the original "Proposed work" remove-if-dead path. The PR honors that: it retains the PK branch as a guarded no-op, and the new `GameContext.cs` comment is honest that the branch is currently unreachable in EF Core 10 while justifying why it stays. Good call keeping the conservative line rather than trimming a correctness guard on the strength of a single EF-version probe.

### Test coverage
- `SaveEnemies_EditsRecordZero_UpdatesTheCorrectRow` is the right shape: separate seed/write/assert scopes (mirrors the per-request change tracker), asserts the seeded ids land at 0 and 1 (self-validating against `RESTART IDENTITY`), and the **neighbor** assertion is what actually proves the `UPDATE` was scoped to row 0 and not broadened by a mis-resolved key. Helper signatures (`TestDataSeeder.CreateEnemyAsync`, base-class scope helpers) line up.
- Worth being clear-eyed (the PR already is): since the PK branch is a no-op on the admin path in EF Core 10, this test guards the end-to-end record-0 edit *outcome*, not the branch firing. That's still the valuable thing to pin. The FK branch retains its existing in-process coverage.
- The `GameContextZeroBasedIdentityFixupTests` comment correction is accurate — the prior wording implied EF marks the Modified record-0 key temporary, which the investigation disproved.

### Nits
- None worth holding the PR for. The expanded `GameContext.cs` comment is long (~11 lines), but this is exactly the "extra context is critical" case CLAUDE.md carves out — it explains why a future reader shouldn't delete a branch that looks dead — so I'd keep it as-is.

### What's good
- The empirical EF-Core-10 matrix in the PR description (every state transition into/out of a temporary key) is genuinely thorough and gives future readers a real answer instead of a guess.
- Docs (`backend-admin.md`) updated to match: FK branch live, PK branch a retained defensive no-op.

I'll merge once CI goes green (verify-backend already passed; test-backend and the e2e jobs are still running).

---

PR #1036, opened 2026-06-20
Fold the admin Edit-existence guard into ChangeSetProcessor (#965)

### LINKED ISSUE #965: Fold the admin Edit-existence guard into ChangeSetProcessor (DRY)
## Background

Follow-up to #951 / PR #964. While fixing the duplicate-key/missing-existence gaps on the admin Content Authoring write path, the root cause of gap #3 (`AdminItems.SaveModSlots` omitting the existence check its siblings perform) was that the **Edit-existence rejection guard is copy-pasted across the admin repos**. Because each repo re-implements it inline, a new write path can simply forget it — which is exactly what happened.

## Proposal

Fold the identity-level Edit-existence rejection into `ChangeSetProcessor.Apply` so a repo *can't* omit it:

- Add an optional existence predicate + resource name to `ChangeSetProcessor` (the part that varies per repo: `_items.LookupItem(id) is null`, `_skills.LookupSkill(id) is null`, the zero-based-id bounds check, etc.). When supplied, the processor validates every `Edit` (and applicable `Add`/owner) up front and returns `AdminSaveResult.NotFound(resource)` before staging anything, mirroring the current per-repo up-front validation.
- Migrate `AdminItems.SaveItems`, `AdminItemMods.SaveItemMods`, `AdminSkills.SaveSkills`, `AdminEnemies.SaveEnemies`, `AdminZones.SaveZones`, and `AdminChallenges.SaveChallenges` onto it, deleting the duplicated `changes.Any(c => c.ChangeType == EChangeType.Edit && ...)` blocks.
- While there: `AdminChallenges.SaveChallenges` reimplements the zero-based-id bounds check inline (`c.Item.Id < 0 || c.Item.Id >= challengeCount`) instead of reusing `_challenges.ValidateChallengeId(...)`, which `AdminZones` already uses. Route it through the shared validator.

## Notes

- Keep the distinction documented in `docs/backend-admin.md`: identity/owner saves **reject** a missing edit up front; child-collection edits/deletes of a missing *member* remain guarded **no-ops**. This refactor only consolidates the former.
- Pure tech debt — no user-facing behavior change; the existing admin tests should continue to pass unchanged.

### Labels
tech debt, claude, scope: medium

ginderjeremiah:
## Summary

The identity-level "an Edit must target an existing record" rejection was **copy-pasted across every admin save path**, so a new write path could simply forget it — which is exactly what happened with `SaveModSlots` in #951. This folds that guard into `ChangeSetProcessor` itself so a save *can't* omit it.

`ChangeSetProcessor.Apply` gains an optional `editExists` predicate. When supplied, the processor validates every `Edit` up front and returns `AdminSaveResult.NotFound(resource)` **before staging anything** (checked ahead of the duplicate-key guard to preserve the prior per-repo precedence). Each save now supplies only the part that varies — the record-set's existence check — instead of re-implementing the loop.

## How it addresses the issue

- **Migrated** `SaveItems`, `SaveItemMods`, `SaveSkills`, `SaveEnemies`, `SaveZones`, and `SaveChallenges` onto the processor, deleting the duplicated `changes.Any(c => c.ChangeType == EChangeType.Edit && ...)` blocks (and the equivalent loop line in `SaveZones`).
- **`AdminChallenges`** now routes through the shared `IChallenges.ValidateChallengeId(...)` instead of reimplementing the zero-based-id bounds check inline (matching what `AdminZones` already did).
- **`AdminZones`** keeps its record-set-specific boss/unlock FK validation (those checks also apply to `Add`s, so they can't fold into the generic guard); only the zone Edit-existence check moves to the processor.
- **Docs:** `docs/backend-admin.md` updated to record that the Edit-existence rejection now lives in the change-set processor, while child-collection edits/deletes of a missing *member* remain guarded no-ops — the distinction the issue asked to keep documented.

### Note for the reviewer
One deliberate, minor behavior nuance in `SaveZones`: the boss/unlock FK pre-pass now runs before the processor's edit-existence check, so a doubly-invalid Edit (a *missing* zone that *also* carries an invalid boss reference) now surfaces the boss-reference failure instead of "Zone not found." Both are valid rejections of invalid input and no test pins the precedence; flagging it in case the ordering matters.

## Test plan

- [x] `ChangeSetProcessor` unit tests for the new `editExists` guard: rejects a missing edit (`NotFound`, nothing staged), allows a present edit, only gates `Edit`s (an `Add` still applies), no-predicate skips the check, and not-found takes precedence over the duplicate-key guard.
- [x] New `SaveZones_EditUnknownZone_ReturnsNotFound` integration test pinning the riskiest split.
- [x] Existing admin integration tests (`AdminChallenges`, `AdminItemMods`, etc.) pass **unchanged**, confirming no user-facing change.
- [x] Full backend build clean (0 warnings) and the entire test suite green (Infrastructure 57, Application 393, Core 649, Api 549).

Closes #965

---
_Generated by [Claude Code](https://claude.ai/code/session_01XEYCqvCJT4vF6URkB1Akfb)_


ginderjeremiah (commented):
## Review: Fold the admin Edit-existence guard into `ChangeSetProcessor` (#965)

**Verdict:** Approve (posted as a comment since I'm the PR author) — clean, faithful tech-debt consolidation that does exactly what the issue asked, with thorough test coverage and no behavior change.

I verified the migration preserves the prior behavior precisely:
- `ChangeSetProcessor.NotFoundResource` capitalizes the lowercase `resourceName`, so `"item mod"` → `"Item mod not found."` etc. — byte-for-byte the same copy the deleted per-repo `AdminSaveResult.NotFound("Item mod")` calls produced (`AdminSaveResult.NotFound` ⇒ `"{entity} not found."`).
- `AdminChallenges` now routing through `IChallenges.ValidateChallengeId` (`id >= 0 && id < Count`) is equivalent to the deleted inline `id < 0 || id >= challengeCount` reject, and matches what `AdminZones` already used — exactly the dedup the issue called for. No dead `challengeCount` left behind.
- The `editExists` guard runs ahead of the duplicate-key guard, preserving the per-repo existence-before-dedup precedence (pinned by `Apply_MissingEditAndDuplicateKey_NotFoundTakesPrecedence`).
- `editExists` gates only `Edit`s, so `Add`s (new identities) are untouched, and `AdminZones` correctly keeps its record-set-specific boss/unlock FK validation inline since those also apply to `Add`s.

### Test coverage
- The five new `ChangeSetProcessor` unit tests cover every branch of the guard: missing edit → `NotFound` with nothing staged (batch stays atomic), present edit applies, only `Edit`s gated, no-predicate skips the check, and not-found-before-duplicate precedence. The capitalization helper is exercised indirectly here (`"Widget not found."`) and in the integration test.
- `SaveZones_EditUnknownZone_ReturnsNotFound` pins the riskiest split — the one save where the guard now sits behind the FK pre-pass.
- The other migrated repos leaning on their existing (unchanged-passing) integration tests is appropriate for a pure refactor. Backend-only, no battle logic, so no frontend mirroring needed.

### Notes (non-blocking)
- The `SaveZones` precedence nuance you flagged (a missing zone that *also* carries an invalid boss ref now surfaces the boss error instead of "Zone not found") is fine — both are valid rejections of input no real client produces, the batch is rejected with nothing staged either way, and no test pins the ordering. Documenting it in the PR body was the right call.

Docs (`backend-admin.md`) are updated to record that the Edit-existence rejection now lives in the processor while child-collection member edits/deletes stay guarded no-ops — the distinction the issue asked to preserve.

Nicely done — I'll merge once CI is green.


---

Issue #1044, opened 2026-06-20
Offline rewards: async simulation progress streaming (deferred follow-up)


ginderjeremiah:
Sub-issue of #879 (offline time tracking & rewards spike). **Deferred follow-up — depends on #1042 (inline orchestration).** Not `claude`-tagged: pick up only on a measured need.

See the spike design doc: `docs/spikes/879-offline-rewards.md` (decision 2 & the deferred follow-up note).

## Goal

If the inline offline simulation (#1042) causes a latency or thread-pool problem at peak login (e.g. a post-deploy thundering herd), replace it with a background simulation that streams incremental progress to the client, per the original issue idea: a command kicks off the sim and returns an initial estimate, then the server pushes progress updates as it runs.

## Why deferred

The inline run is bounded by the 10h cap + 5s cooldown to ~a few seconds worst-case, comfortably under the 25s command timeout, so V1 ships inline. This is gated on a real measured need and warrants its own design pass.

## Design considerations (when picked up)

- Reuse the server-push pattern (the `ChallengeCompleted` → `SocketManagerService.EmitSocketCommand` → Redis backplane path) for progress + completion pushes.
- Background-task lifecycle and cancellation (player disconnects mid-sim).
- Off-loop player-state mutation must respect the per-socket command lock (the lost-update guard) — the background task can't freely mutate the cached player concurrently with the read loop.
- Partial-result persistence / resumability if the sim is interrupted.

Issue #1043, opened 2026-06-20
Offline rewards: frontend welcome-back gate & summary


ginderjeremiah:
Sub-issue of #879 (offline time tracking & rewards spike). **Depends on #1042 (orchestration & socket command).**

See the spike design doc: `docs/spikes/879-offline-rewards.md` (decision 2 & the flow section).

## Goal

A "welcome back" gate between the loading screen and the game shell that runs the offline-progress check and shows the player what they earned while away, before the idle loop starts.

## Scope

- After the socket connects and reference data loads, issue `GetOfflineProgress` **before** starting the idle loop (so simulated and live battles never overlap).
- If the summary has content, show a gate screen: away duration, loop mode (idle zone vs boss), battles won/lost/drawn, total exp, levels & stat points gained, and challenges completed (with the items/mods/skills they unlocked). Then re-sync authoritative state (player, progress, inventory, skills) and enter the game.
- If the summary is empty (away < threshold), pass straight through to the game with no gate.
- Fits in the login → loading → game route flow (`UI/src/routes`); follow the loading-screen/gate patterns.

## Notes

- The frontend half of the mode-sync ships in #1040; this issue consumes the resulting summary.
- Document the new screen in `docs/frontend-screens.md`.

## Tests

- Unit: the gate renders the summary fields; an empty summary skips the gate.
- The idle loop does not start until the gate is dismissed.

Issue #1042, opened 2026-06-20
Offline rewards: reward application, challenge consolidation, orchestration & socket command


ginderjeremiah:
Sub-issue of #879 (offline time tracking & rewards spike). **Depends on #1039 (away-time foundation), #1040 (loop-mode persistence), #1041 (core sim engine).**

See the spike design doc: `docs/spikes/879-offline-rewards.md` (decisions 2, 6, 7, and the flow section).

## Goal

Wire the core simulator (#1041) into the live application: compute the away time, run the sim, apply all rewards to the player, and expose it as a socket command returning a welcome-back summary. Runs **inline** within the command for V1 (the few-seconds worst case fits under the 25s command timeout); async progress streaming is the deferred follow-up.

## Scope

Orchestration (application layer — e.g. `BattleService.SimulateOfflineProgress` or a new `OfflineRewardService`):
- Compute away time = `now − Player.LastActivity` (#1039). If `< 5 min`, set `LastActivity = now` and return an empty result.
- Resolve any stale in-flight battle first (a mid-battle disconnect), then read the persisted loop mode + zone (#1040), fetch reference data, and build the resolver funcs.
- Call the #1041 simulator (away clamped to the 10h cap).
- **Apply exp per simulated victory** via the existing `Player.GrantExp` (decision 6): each grant is small so the per-grant `MaxExpPerGrant` clamp never truncates a legitimate haul, the level-up loop runs correctly, and the per-event envelopes buffer into the one write-behind batch. The net level gain is summarized, not surfaced per level (`PlayerLeveledUpEvent` has no consumer).
- **Consolidate statistics + challenges** (decision 7): accumulate the stat deltas across all battles, apply them, evaluate the affected challenges **once**, unlock their rewards, and collect the completions for the summary. **Suppress the live per-challenge `ChallengeCompleted` push** for the offline window — the summary is the notification and the client re-syncs on leaving the gate. Factor the per-battle stat-update logic so the live `BattleCompletedEvent` handler and this offline batch **share** it (DRY) rather than duplicating.
- Set `LastActivity = now` (idempotent re-claim), persist once.
- Short-circuit a run that yields no progress over its first batch of battles (CPU-waste guard); observe the command cancellation token in the loop.

Socket command + API:
[...]

Issue #1041, opened 2026-06-20
Offline rewards: core offline simulation engine (Game.Core)


ginderjeremiah:
Sub-issue of #879 (offline time tracking & rewards spike). No dependencies (pure domain; consumed by the orchestration sub-issue).

See the spike design doc: `docs/spikes/879-offline-rewards.md` (decisions 1 & 3, and the performance section).

## Goal

A pure `Game.Core` domain service that replays a player's missed idle/boss battles over an away period and returns the accumulated results, without touching persistence or the application layer.

## Why full simulation is feasible (from the spike)

Offline, a player's combat power and reward distribution are **stationary** — combat reads only allocations + equipment (neither changeable offline), level is unused in combat math, and enemies scale with the zone, not the player. The 5s post-battle cooldown throttles battle count, so a full 10h replay costs only ~a few seconds of CPU. So we simulate fully (exact), not sample-and-extrapolate.

## Scope

- A domain service (e.g. `OfflineProgressSimulator`) taking: the player `BattleSnapshot`, the loop mode (idle/boss) + zone, the away budget (ms), the cap (10h), the cooldown (`BattleService.PostBattleCooldown`, 5s), and resolver funcs (enemy/item/mod/skill) — mirroring how `BattleFactory`/`BattleSnapshot` already take resolvers to keep `Game.Core` free of data-access refs.
- Loop the mode's battle type for the whole budget:
  - **Idle** → `BattleFactory.CreateBattleEnemy` (random enemy/level/loadout) in `CurrentZoneId`.
  - **Boss** → `BattleFactory.CreateBossEnemy` (deterministic) for the boss of the current zone. Continues through losses and draws (decision 3 — no stop, no idle fallback).
- Each iteration: fresh seed, `BattleSimulator.Simulate()`, consume `result.TotalMs + cooldown` from the budget, stop at budget/cap exhaustion.
- Accumulate per-battle results: win/loss/draw counts, total/per-victory exp (`DefeatRewards`), per-enemy kill counts, and the battle-stat deltas needed by the statistics consolidation.
- Return a structured `OfflineProgressResult`; **do not mutate the player or persist** (that is the orchestration sub-issue).
- Make the seed/RNG source injectable so tests are deterministic.

## Tests (heavy — this is the core logic)

- Idle loop runs to the cap; budget accounting (`duration + cooldown`) is correct.
- Boss loop runs the whole budget through a mix of wins/losses/draws.
[...]

PR #1046, opened 2026-06-20
Document offline rewards spike (#879)

ginderjeremiah:
Closes the research phase of #879 (offline time tracking & rewards). Adds `docs/spikes/879-offline-rewards.md` recording the decided design and splitting the work into implementation sub-issues. **Docs only — no code changes.**

## Decided direction
- **Full server-side simulation**, not sample-and-extrapolate — replay every missed battle, capped at 10h, triggered above 5 min away.
- Run **inline within one socket command** behind a "welcome back" gate at login (async progress streaming deferred to a follow-up).
- Respect the active loop: **idle-farming or auto-challenge-boss**. Each just loops its battle type for the whole away budget; boss farming continues through losses/draws (no present player to turn auto-fight off).

## Why it's feasible (the key insight)
Offline, a player's combat power and reward distribution are **stationary**: combat reads only stat allocations + equipment (neither changeable offline), `Battler.Level` is unused in combat math, and enemies scale with the zone, not the player. With the 5s cooldown throttling battle count, a full 10h replay costs only ~a few seconds of CPU and is exact — so it reuses `BattleSimulator`/`BattleFactory`/`DefeatRewards` directly.

## Sub-issues (linked under #879)
- #1039 — away-time tracking foundation (`LastActivity`) *(small)*
- #1040 — persist active idle-loop mode (idle vs auto-boss) *(medium)*
- #1041 — core offline simulation engine, `Game.Core` *(large)*
- #1042 — reward application + challenge consolidation + orchestration + socket command *(large; depends on #1039/#1040/#1041)*
- #1043 — frontend welcome-back gate & summary *(medium; depends on #1042)*
- #1044 — *(deferred)* async simulation progress streaming *(depends on #1042)*

### DIFF
--- a/docs/spikes/879-offline-rewards.md
+++ b/docs/spikes/879-offline-rewards.md
@@ -1,0 +1,85 @@
+# Spike #879 — Offline time tracking & rewards
+
+- **Spike issue:** [#879](https://github.com/ginderjeremiah/GameServer/issues/879)
+- **Status:** Research complete; direction decided with the project owner; split into implementation sub-issues [#1039–#1044](#implementation-issues). Not yet implemented.
+
+## Goal
+
+When a player returns after being away, credit them for the progress their idle loop *would* have made while disconnected. Track when the player was last active, and on the next login simulate the battles they missed (capped), then present a "welcome back" summary before they re-enter the game. The simulation must respect what the player's idle loop was actually doing at disconnect — idle-farming a zone **or** auto-challenging a zone's boss.
+
+## The core enabling insight — offline battles are *stationary*
+
+The whole feasibility argument rests on one fact established from the code: **a player's combat power and reward never change while they are offline.**
+
+- A player's battle attributes come **only** from stat allocations + equipped items/mods (`BattleSnapshot.GetModifiers`, composed identically to the live `Player.GetAllModifiers`). Neither can change offline — allocating stat points and equipping gear are socket commands the absent player never sends.
+- The player's **level is not used in combat math at all** — `Battler` stores `Level` but no combat path reads it; all math reads the attribute collection. Level only feeds the *threshold* for the next level-up.
+- The exp reward is a function of **enemy power vs. the player's snapshot power** (`DefeatRewards`), and enemies scale with the **zone's** level range (`Zone.RollEncounterLevel`), independent of the player's level.
+
+So every offline idle battle is an i.i.d. draw from the *same* distribution (random enemy, random level within the zone band, random enemy loadout, fresh RNG seed). The player accumulates exp, levels, and (unallocated) stat points, but none of that shifts the per-battle outcome or reward distribution. This is what makes a faithful full simulation both correct and bounded.
+
+## Performance — full simulation is cheap, the cooldown is the throttle
+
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: Document offline-rewards spike (#879)

**Verdict:** Approve *(submitted as a `COMMENT` since I authored the PR and GitHub won't accept a self-approval — but this is a clean approve).*

Docs-only PR recording the decided direction for the offline-rewards spike and splitting the work into linked sub-issues. It meets the project's spike requirements in `CLAUDE.md`: results are documented in `docs/spikes/`, the implementation work is split into sub-issues #1039–#1044, and they're linked under the parent spike #879.

The value of this doc is that it commits to a non-obvious design (full server-side replay over sample-and-extrapolate) on the strength of a specific "offline battles are stationary" argument. I spot-checked every load-bearing claim that argument rests on against the code, and they all hold:

### What's good
- **Stationarity argument is accurate.** Verified against the code: `Battler.Level` is stored but never read by any combat path (all math goes through the attribute/modifier collection in `BattleContext`); `BattleSnapshot.GetModifiers` composes the same sources as the live `Player.GetAllModifiers` (stat allocations + equipped item/mod modifiers); and enemy power comes from `Zone.RollEncounterLevel` over the zone's band, independent of player level. The conclusion that per-battle outcome/reward distribution is fixed for the whole offline window is sound.
- **Feasibility constants are correct.** `SocketHandler.DefaultCommandTimeout = 25s`, `BattleService.PostBattleCooldown = 5s`, `GameConstants.DefaultMaxBattleMs = 2 min`, `MsPerTick = 40`, and `ServerGameConstants.MaxExpPerGrant = 100k` all match the doc — so the "inline within one socket command" feasibility case checks out.
- **"What it builds on" section is accurate.** Confirmed: `BattleFactory.CreateBattleEnemy`/`CreateBossEnemy` build battle-ready enemies from resolver funcs; idle kills grant exp only (`DefeatRewards`), with items/mods/skills unlocked exclusively via `Player.CompleteChallenge`; `Player.LastActivity` is not updated during play and `PlayerState` holds only cooldown/battle-session fields; the `EnemyManager.autoFight` flag is frontend-only and never sent to the backend; and each `CompleteChallenge` raises one `ChallengeCompletedEvent` → one socket push.
- **Decisions are well-reasoned and the footguns are pre-empted.** The CPU-waste guard (all-draw run repeatedly re-triggered by reconnecting), the cancellation-token requirement to avoid the 25s timeout, double-collection prevention via resetting `LastActivity` on claim, and the boss-farming-through-losses balance lever are all called out as notes/follow-ups — exactly the kind of thing that should be flagged before implementation rather than discovered during it.
- **Sub-issue decomposition is clean** with an explicit dependency order (#1039/#1040/#1041 independent → #1042 → #1043), correct scope tags, and the async-streaming idea correctly deferred (#1044, un-`claude`-tagged) behind a measured need.

No blocking findings, no should-fixes, no nits. The doc itself notes the docs that need updating when the feature lands (`game-design.md`, `backend.md`/`backend-battle.md`, `frontend-screens.md`), which is the right place for that to live.

Approving and merging.

---

PR #1053, opened 2026-06-20
Add per-account login backoff after consecutive failures (#1010)

### LINKED ISSUE #1010: Consider account-level login lockout/backoff after N consecutive failures
## Context

Follow-up from #950 / PR #1009, which added **per-client-IP** rate limiting to the anonymous auth endpoints (login, refresh, account creation). That closes the IP-rate-based credential-stuffing / refresh-brute / PBKDF2-DoS surface.

The original issue also suggested *considering* an **account-level lockout/backoff after N consecutive failed logins** as defense-in-depth against a slow, distributed (botnet) guess against a single account that stays under any single IP's rate limit. That was deliberately scoped out of PR #1009 because it needs a design decision rather than being a drop-in.

## The design tension

A naive per-account lockout is itself a **denial-of-service vector**: an attacker who knows a victim's username can lock them out at will by spamming failed logins. Any implementation has to weigh that, e.g.:

- **Exponential backoff** (delay grows per consecutive failure) rather than a hard lockout, so a legit user is slowed but never fully locked out by an attacker.
- **Soft lockout with side channels** — temporary lock plus an email/owner-notify unlock path (we don't currently have email).
- **Track failures per (account × IP)** rather than per account, so one attacker IP can't lock the account for everyone — though this weakens the distributed-guess protection the feature is meant to add.

## Proposed work

1. Decide the model (recommend exponential backoff keyed on the account, with a low cap, over a hard lockout) — get sign-off on the approach first given the DoS tradeoff.
2. Track consecutive-failure state in Redis (auth already depends on Redis, consistent with refresh-token storage), reset on a successful login.
3. Surface a clear client-facing message distinct from "invalid credentials" when backed off, and add unit + integration coverage (including the reset-on-success and the attacker-can't-hard-lock paths).

## Notes

Lower urgency than the IP-rate gap that #950/#1009 closed; this is the next layer of defense-in-depth, not an open hole.

ginderjeremiah:
## Summary

Closes #1010. Adds a **per-account exponential login backoff** layered on top of the per-client-IP auth rate limiter from #950/#1009, as defence-in-depth against a **slow, distributed (botnet)** credential guess against a single account that stays under any single IP's rate limit.

The design fork the issue called out — **per-account** vs **per-(account × IP)** keying — was signed off as **per-account** before implementation. Per-(account × IP) would give each distributed bot its own counter and add little over the IP limiter, which is exactly the gap this feature is meant to close.

## How it works

- `Login` checks a per-account failure streak (tracked in Redis, keyed by a SHA-256 hash of the case-sensitive username — the same exact-match semantics login uses) **before** any DB hit or PBKDF2 work.
- The first `FailureThreshold` consecutive failures are free; after that each further failure delays the next permitted attempt by a **doubling** window (`BaseDelaySeconds × 2ⁿ`) up to a **low cap** (`MaxDelaySeconds`).
- It is a **bounded backoff, never a hard lockout** — a legitimate user is delayed by at most the cap — so an attacker who knows a username can only briefly *slow* the owner, not lock them out.
- The streak is **reset on a successful credential verification** (a correct password isn't a brute-force guess), before any later ban / no-player rejection.
- A backed-off attempt is rejected before credentials are touched (so spamming during an active window neither extends it nor spends server CPU) and surfaces as a **429 with `Retry-After`** plus a client-facing message **distinct from invalid-credentials**.
- Parameters are **config-bound** (`LoginBackoff` section) with safe defaults and validated as sane on start, mirroring the IP rate limiter.

## Design / layering

- **Pure policy** (`LoginBackoffPolicy`) holds the backoff arithmetic and is unit-tested in isolation (the current instant is supplied).
- A thin **`LoginBackoffGuard`** wraps the Redis-backed `ILoginBackoffStore` (impl in the data tier, mirroring `RefreshTokenStore`) and an injected `TimeProvider`, so the time-based logic stays deterministically testable.
- The 429 reuses the existing `ApiErrorCategory` → `ErrorStatusFilter` status mapping (new `TooManyRequests` → 429); the controller sets `Retry-After`. No frontend change is needed — `ApiErrorCategory`/`LoginStatus` are not on the wire, and the client already surfaces the `{ errorMessage }` envelope on a 429 (as it does for the #1009 rate-limit 429).

## Test plan

- ✅ `LoginBackoffPolicyTests` — unit tests for the threshold, exponential doubling, the cap, retention TTL, and the active-window check.
- ✅ `AccountServiceIntegrationTests` (new cases) — backs off after consecutive failures; the window **elapses and allows retry** (proves it's a slowdown, not a hard lockout); a **successful login resets** the streak.
- ✅ `AuthBackoffTests` — API-level: a 429 with `Retry-After` and a message distinct from the 400 invalid-credentials response.
- ✅ Full backend suite green (1,563 tests); existing `AuthRateLimitingTests` / `LoginControllerTests` unaffected.

## Docs

- `docs/backend.md` — added a concise "Per-account login backoff" subsection under auth.

https://claude.ai/code/session_01M1PzhRTo2PmZ5qENUJB1e2

---
_Generated by [Claude Code](https://claude.ai/code/session_01M1PzhRTo2PmZ5qENUJB1e2)_


Game.Application/Auth/LoginBackoffGuard.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// Records a failed login attempt, incrementing the account's consecutive-failure count and extending
+        /// its backoff window per the policy. Only ever called for attempts that passed
+        /// <see cref="GetActiveBackoff"/>, so spamming during an active window neither extends it nor spends
+        /// server credential work — the gate rejects those attempts before they reach here.
+        /// </summary>
+        public async Task RegisterFailure(string username, CancellationToken cancellationToken = default)
+        {
+            var current = await _store.Get(username, cancellationToken);
+            var (next, retention) = _policy.RegisterFailure(current, _timeProvider.GetUtcNow());
+            await _store.Set(username, next, retention, cancellationToken);
```
ginderjeremiah, line 42: **Should-fix (follow-up filed):** this GET → compute → SET is a non-atomic read-modify-write. Two concurrent failed attempts for the same account can read the same `FailureCount`, compute the same next state, and the later `Set` overwrites the earlier — concurrent failures lose increments and the streak climbs slower than the attempt rate.

That's worth calling out because the feature's stated threat model is a *slow, distributed (botnet)* guess on one account, which is exactly where concurrent failures land. It's still defence-in-depth on top of the per-IP limiter, so not a blocker — tracked as a follow-up in #1054 (atomic `INCR`/Lua + a concurrency test). Fine to merge this as-is and harden there.

---

ginderjeremiah (commented):
## Review: per-account exponential login backoff (#1010)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating vote on my own PR) — solid, well-layered defence-in-depth that matches the signed-off per-account design; one non-blocking should-fix already filed as a follow-up.

This cleanly implements what #1010 asked for and respects the sign-off in the issue thread (per-account keying, exponential backoff with a low cap, never a hard lockout). Layering is right: the pure arithmetic lives in `LoginBackoffPolicy` (Application, unit-tested in isolation with an injected instant), the Redis read-modify-write + clock sit in a thin `LoginBackoffGuard`, and `ILoginBackoffStore`/its Redis impl mirror the existing `RefreshTokenStore`. The 429 reuses the existing `ApiErrorCategory` → `ErrorStatusFilter` mapping rather than special-casing the controller.

A few correctness points I checked and am satisfied with:
- **No user enumeration / bypass:** `RegisterFailure` runs for both the unknown-account and wrong-password branches, and the backoff key (SHA-256 of the case-sensitive username) uses the same exact-match semantics as the login lookup (`u.Username == username`), so the key can't be sidestepped in a way login itself isn't.
- **Streak reset ordering:** reset happens immediately after credential verification, *before* the ban / no-player / player-data branches — correct, since those aren't credential failures.
- **TTL is a true sliding window:** each `Set` replaces the entry with `max(window, delay)` retention, so an in-effect lock is never dropped early and a quiet streak self-expires.
- **Cap arithmetic:** computing in `double` and clamping before the `TimeSpan` conversion avoids overflow on a long streak; the unit tests pin the doubling, the cap, and the saturation case.

### Should-fix

- `LoginBackoffGuard.RegisterFailure` (`Game.Application/Auth/LoginBackoffGuard.cs:40-42`) — non-atomic read-modify-write; concurrent failures can lose increments. Matters because concurrency is the feature's threat model, but it's defence-in-depth on the per-IP limiter so not a blocker. Filed as **#1054** (atomic `INCR`/Lua + concurrency test). Fine to merge and harden there.

### Nits

- Per failed attempt there are two Redis GETs (one in `GetActiveBackoff`, discarded, then another in `RegisterFailure`) plus the SET, and every *successful* login issues an unconditional `Clear` DELETE even with no streak. All cheap, but if #1054 reworks the guard it'd be natural to thread the already-read state through and skip the no-op delete.

### Test coverage

Meets the project's bar and the issue's explicit asks: `LoginBackoffPolicyTests` covers the threshold, doubling, cap, retention, and active-window check; `AccountServiceIntegrationTests` covers back-off-after-failures, **window-elapses-allows-retry** (proves slowdown, not lockout), and **reset-on-success**; `AuthBackoffTests` covers the HTTP edge (429 + `Retry-After` + a message distinct from the 400). No frontend/battle-logic surface is touched, so no mirrored frontend tests are owed. The one gap — concurrent-attempt behavior — is exactly what #1054 will add.

### What's good

Genuinely careful security work: gating before any DB/PBKDF2 cost, bounded backoff to avoid the self-inflicted DoS the issue warned about, config-bound with start-time validation, hashed keys, and concise docs added to `backend.md`. The XML doc comments are a bit long but they're carrying real design rationale, which is fair here.

CI is still running; I'm leaving the merge to a green pipeline plus a call on the #1054 follow-up rather than merging over an open should-fix.

---

PR #1056, opened 2026-06-20
Document deliberate no-CancellationToken choice for write-behind drain handlers

### LINKED ISSUE #1029: Decide whether to thread a CancellationToken through the write-behind player-update drain handlers
## Context

Follow-up from #961 / #1028. The new `PlayerUpdateHandlerExtensions.InsertIfMissingAsync` helper — and the `IPlayerUpdateHandler.HandleAsync` handlers that call it — currently thread **no** `CancellationToken`; their `SaveChangesAsync()` calls run without one. The original issue (#961) suggested an `InsertIfMissingAsync(ctx, predicate, factory, ct)` signature, but it was intentionally left out of the DRY pass to avoid widening scope to the handler interface and dispatcher.

`docs/backend.md` notes that the per-socket-command cancellation token is plumbed through the application services and the data tier. That path is the *socket command* path, though — these write-behind drain handlers run off a background service, so they're a separate question.

## The actual decision to make

It is **not obvious** that plumbing a token here is even desirable:

- The write-behind drain exists to *flush* player state to Postgres. The graceful-shutdown drain (see `docs/backend.md` → "Caching, Pub/Sub & write-behind persistence" / `backend-persistence.md`) wants in-flight saves to **complete**, not abort — cancelling `SaveChangesAsync` on the shutdown token could drop a write that the queue then has to redeliver.
- On the other hand, a per-drain-operation timeout (distinct from the shutdown token) could be worth having so a wedged save unwinds rather than stalling the drain loop.

So this is a design question, not a mechanical "add a `ct` param": decide whether the drain handlers should honour cancellation at all, and if so *which* token (operation timeout vs. host shutdown) and what the desired behaviour on cancel is. If the answer is "yes", thread it through `IPlayerUpdateHandler.HandleAsync`, the dispatcher, `InsertIfMissingAsync`, and the `SaveChangesAsync()` calls. If "no", a short doc note recording the deliberate choice closes it.

## Scope

Small. One interface + dispatcher + the handlers, or a doc note if we decide against it.

ginderjeremiah:
## Summary

Resolves the design question in #1029 (a follow-up from #961 / #1028): should a `CancellationToken` be threaded through the write-behind player-update drain handlers? **Decision: no** — and this PR records that deliberate choice rather than changing behaviour.

## How the decision was reached

The synchronizer is already designed so that once an item is *reserved* off the queue, it is processed and acknowledged **without** the stopping token, so an in-flight apply finishes cleanly (`DataProviderSynchronizer.DrainQueueAsync`, and the existing note in `docs/backend-persistence.md` → "Shutdown drains to a clean boundary"). Two findings settle the question:

- **Threading the *shutdown* token would regress behaviour.** If `SaveChangesAsync` were cancelled mid-save, the `OperationCanceledException` would be caught by `ProcessMessage`'s generic retry/dead-letter path — so a stop landing during a save would **dead-letter a perfectly good write** instead of letting it complete.
- **A separate *per-operation* timeout is redundant.** `SaveChangesAsync` already runs under Npgsql's command timeout (the DbContext uses `UseNpgsql(connectionString)` with no override, i.e. the 30s default in `GameContextFactory`). A genuinely wedged save already throws → is retried (3× / 200ms backoff per `PlayerUpdateRetryPolicy`) → dead-lettered, and host shutdown is independently bounded by the 5s drain timeout. Threading a token through `IPlayerUpdateHandler.HandleAsync`, the dispatcher, and `InsertIfMissingAsync` would widen the interface for no behavioural gain — exactly the scope #961/#1028 deliberately avoided.

## Changes

Documentation only — no behavioural change:

- **`docs/backend-persistence.md`** — a concise note under the shutdown section recording the deliberate no-token choice and why both alternatives (shutdown token, per-op timeout) were rejected.
- **`IPlayerUpdateHandler`** — a short `<remarks>` on the interface explaining the signature carries no token by design.
- **`InsertIfMissingAsync`** (`PlayerUpdateHandlerExtensions`) — a one-line pointer to the interface rationale, since #961 originally proposed an `InsertIfMissingAsync(..., ct)` signature.

## Test plan

- [x] `dotnet build Game.DataAccess -warnaserror` → **0 warnings, 0 errors** (doc `<see cref>` references resolve).
- No new tests: this is a documentation/comment-only change with no logic or behavioural change to cover.

Closes #1029

---
_Generated by [Claude Code](https://claude.ai/code/session_01DAqjYJ2wjXw23qW3MR9hQQ)_


ginderjeremiah (commented):
## Review: Document the deliberate no-`CancellationToken` choice for write-behind drain handlers (#1029)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the recorded decision is correct, the rationale is accurate against the code, and the scope is exactly right for a "decide and document" issue.

I verified every technical claim in the doc/comments against the source rather than taking them on faith:

- **"Once reserved, an item is processed and acknowledged without the stopping token."** Confirmed — `DataProviderSynchronizer.DrainQueueAsync` (lines 200–210) threads the token into `ReserveNextAsync` only; `ProcessMessage` and `AcknowledgeAsync` take none.
- **"Threading the shutdown token would dead-letter a good write."** Confirmed — a cancel mid-save surfaces as `OperationCanceledException`, which `ProcessMessage`'s generic `catch (Exception)` arms (DataProviderSynchronizer.cs:286–298) would retry then dead-letter. The regression the note describes is real, so rejecting the shutdown token is the right call.
- **"A per-operation timeout is redundant."** Confirmed — `GameContextFactory.cs:23` calls `UseNpgsql(connectionString)` with no `CommandTimeout` override, i.e. Npgsql's 30s default, and a wedged save is unwound by `PlayerUpdateRetryPolicy.Default` (3× / 200ms) → dead-letter, with host shutdown independently bounded by the 5s drain timeout. A second timeout would widen `IPlayerUpdateHandler.HandleAsync` + the dispatcher + `InsertIfMissingAsync` for no behavioural gain — the surface #961/#1028 deliberately kept narrow.

### Strengths

- The decision is the correct one and is argued from the actual failure mode, not hand-waving.
- Documentation is layered sensibly without duplication bloat: the full rationale lives in `backend-persistence.md`, the interface `<remarks>` carries a condensed version pointing at the doc, and `InsertIfMissingAsync` is just a one-line pointer back to the interface (where #961 originally proposed the `ct` signature).
- `<remarks>`/`<see cref>` placement is valid XML doc and `dotnet build -warnaserror` confirms the crefs resolve.

### Test coverage

None required, and correctly so — this is a comment/doc-only change with no logic or behavioural delta. Per the project's testing rules there's nothing new to cover.

### Notes (non-blocking)

- The `<remarks>` on `IPlayerUpdateHandler` runs three lines. That's above the usual 1–2 line guidance, but it's exactly the "critical context" exception — it's recording a tracked design decision — so it's appropriate as-is, not a finding.

Clean, well-scoped, and accurate. Holding the merge only until the remaining CI jobs (`test-backend`, `test-e2e` firefox/webkit) finish — `verify-backend` and `test-e2e (chromium)` are already green.

---

Issue #958, opened 2026-06-19
Backend test-coverage gaps: write-behind handlers, admin item/mod/skill/challenge repos, reference-cache reload


ginderjeremiah:
## Problem

Three pockets of backend logic with non-trivial correctness consequences are exercised only indirectly (or not at all), so a regression wouldn't fail a test.

### 1. Six write-behind player-update handlers are untested
`Game.Application.Tests/DataAccess/PlayerUpdateHandlerIdempotencyIntegrationTests.cs` covers some handlers but has **zero** coverage for `ItemEquippedHandler`, `ItemUnequippedHandler`, `ItemFavoriteChangedHandler`, `SelectedSkillsChangedHandler`, `PlayerCoreUpdatedHandler`, and `ModRemovedHandler` — including the two subtlest (the CASE-driven equip upsert and the delete-then-rebuild loadout handler). No test exercises same-player event ordering.

Given the recent run of write-behind idempotency fixes (#915/#924, #891/#916) and the open #936, these are exactly the handlers most worth pinning.

**Fix:** idempotency + concurrent-apply tests for the six, plus an ordering test asserting convergence (or documenting a known divergence).

### 2. Admin repos AdminItems / AdminItemMods / AdminSkills / AdminChallenges are untested
Only `AdminEnemies`/`AdminZones`/`AdminTags` + the generic processors have dedicated tests under `Game.Application.Tests/DataAccess/`. The exact edge cases in the admin-write validation issue (duplicate-key handling, membership-guarded edit/delete, `SaveModSlots`/`SetEffects`) are untested.

**Fix:** integration tests covering not-found rejection, membership-guarded edit/delete, duplicate-key handling, and the `SaveModSlots`/`SetEffects` paths.

### 3. Reference-cache reload/holder invariants are untested
`CoalescingReferenceCacheReloader` (debounce/coalesce + retry/give-up) and the `Id == index` build invariant (see the contiguity-assertion issue) have no unit coverage.

**Fix:** fake-`IReloadableReferenceCache` tests for coalescing N→1, retry-then-succeed, retry-exhausted-keeps-snapshot; and a holder test asserting `Current[i].Id == i` after a real build.

### 4. (Coverage gating) socket/filter logic is measure-only despite holding real branching
[...]

PR #1060, opened 2026-06-20
Gate Game.Api socket/services/filters as a coverage sub-area (#1034)

### LINKED ISSUE #1034: Carve Game.Api socket/services/filters into a gated coverage sub-area
## Background

Split out from #958 (item 4). The other three coverage gaps in #958 are now closed (write-behind handler tests, admin item-mod/challenge repo tests, and the already-present reference-cache reloader/holder tests). This issue tracks the remaining item, which is a distinct **build-infrastructure** change rather than a "write a missing test" task.

## Problem

`build/coverage-floors.json` marks the whole `Game.Api` assembly **measure-only**, justified as "thin adapters." That is accurate for the controllers and DTOs, but **not** for the socket layer (`SocketCommandProcessor`/`SocketManagerService` timeout/lock handoff, drain, presence rollback, server-push escalation) or `AdminRoleAuthorizationFilter` / `ErrorStatusFilter`. These are security- and data-loss-sensitive branches exercised by integration tests but **not gated**, so a regression can drop their coverage with no build failure.

## Why it's separate

The gate (`build/coverage-gate.ps1`) reads ReportGenerator's `Summary.json` at **assembly** granularity (`summary.coverage.assemblies`). Gating a *namespace subset* of `Game.Api` requires:

- a ReportGenerator namespace/class filter (a second filtered report, or parsing class-level entries) so the socket/services/filters namespaces are measured as their own bucket;
- an extended `coverage-floors.json` schema to express a sub-assembly (namespace) floor while keeping the rest of `Game.Api` measure-only;
- gate-script changes to read and enforce that namespace floor.

It also needs a design decision on exactly which namespaces to gate and what floors to seed (at current).

## Proposed work

- Carve the socket/services/filters namespaces into a **gated** sub-area, seeded at current, leaving controllers/DTOs measure-only — matching the project's "strict floors where logic concentrates" principle (`docs/infrastructure.md` → Code coverage).
- Update `docs/infrastructure.md` to document the new sub-area gate.

### Labels
tech debt, claude, scope: medium</body>
<parameter name="labels">["tech debt", "claude", "scope: medium"]

ginderjeremiah:
Closes #1034 (split from #958, item 4).

## Problem

`build/coverage-floors.json` marks the whole `Game.Api` assembly **measure-only** ("thin adapters"). That's accurate for the controllers and DTOs, but **not** for the socket layer, the request filters (`AdminRoleAuthorizationFilter` / `ErrorStatusFilter`), or the services — security- and data-loss-sensitive branches exercised by integration tests but **not gated**, so a regression could drop their coverage with no build failure.

## Change

Carve those namespaces into a **gated sub-area**, seeded at-current, leaving controllers/DTOs measure-only — the project's "strict floors where logic concentrates" principle, applied below assembly granularity.

- **`build/coverage-floors.json`** — new `gatedNamespaces` section expressing a namespace-subset floor for an otherwise measure-only assembly. Seeds `Game.Api.Sockets` + `Game.Api.Services` + `Game.Api.Filters` at **line 94 / branch 92** (current: **94.8% / 92.2%**).
- **`build/coverage-gate.ps1`** — aggregates the sub-area's coverage from the report's per-class entries (`classesinassembly`), summing covered/coverable lines and covered/total branches for the listed namespaces. No second ReportGenerator pass needed: the per-class counts sum exactly to the assembly totals, so the namespace bucket is computed on the same basis as the assembly gate. Same ratchet/floor semantics, exit-code contract, and table as the assembly gate; a namespace rename that empties the bucket fails loudly (`MISSING`) rather than passing silently on a 0/0.
- **`docs/infrastructure.md`** — documents the sub-area gate under *Code coverage*.
- **`.github/workflows/run-tests.yml`** — adds `build/**` to the `detect-changes` backend path filter (see below).

## Minor related fix: CI path filter

While verifying, I found the `detect-changes` backend filter omitted `build/**`, so a change to the coverage gate itself (`coverage.ps1` / `coverage-gate.ps1` / `coverage-floors.json`) would **skip** the `test-backend` job that runs it — the gate could regress unnoticed. Added `build/**` to the filter. A nice side effect: because the filter also matches `.github/workflows/**`, this PR's workflow edit flips `backend: true`, so the backend coverage job runs on this very PR and actually exercises the new gate logic.

## How it addresses the issue

The issue called for: (1) a way to measure the socket/services/filters namespaces as their own bucket — done via per-class aggregation, no second filtered report; (2) an extended floors schema for a sub-assembly floor while the rest of `Game.Api` stays measure-only — the `gatedNamespaces` section; (3) gate-script changes to enforce it; and (4) a design decision on which namespaces + what floors — `Game.Api.Sockets`/`Game.Api.Services`/`Game.Api.Filters`, seeded at the current integer floor.

## Test plan

No PowerShell test harness exists in the repo (the gate is verified by running it), so behaviour was confirmed manually against a full merged-coverage run (all suites green: Core 649, Application 404, Api 551, Infrastructure 57):

- **Pass at current** — `coverage-gate.ps1` exits 0: *"3 gated assemblies, 1 gated namespaces"*.
- **Floor breach** — raising the namespace floor to 99/99 exits 1 with `Game.Api (sockets/services/filters) line 94.8% < floor 99%` + the branch breach.
- **Bucket emptied** (namespace renamed/removed) — exits 1 with a loud `matched no covered classes` message instead of silently passing.
- **Aggregate cross-checked** — the script's 1165/1229 lines (94.8%) and 212/230 branches (92.2%) match an independent recomputation, and the per-class sums reproduce the `Game.Api` assembly totals exactly (no double-counting).

## Notes / follow-up

- The gate's summary table uses `Out-String` at the host's default width; in a non-TTY shell (width 0) it renders empty, while CI/console widths print it fine. This is pre-existing (the assembly rows have always behaved this way) and untouched here — a one-line `-Width` would make it robust in non-TTY contexts too. Happy to fold that in here or file a small follow-up if wanted.


ginderjeremiah (commented):
## Review: Gate `Game.Api` socket/services/filters as a coverage sub-area (#1034)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't accept a gating event from me). This cleanly delivers all four things the issue asked for: a namespace-bucket measurement, an extended floors schema, the gate-script enforcement, and a documented namespace/floor decision — with no second ReportGenerator pass.

I verified the logic against the full files, not just the diff:

- **Prefix matching is correct and collision-safe.** `Get-NamespaceCoverage` matches on `StartsWith("$ns.")` with the trailing dot, so `Game.Api.Sockets` includes sub-namespaces (`Game.Api.Sockets.Internal.*`) but won't falsely match a sibling like `Game.Api.SocketsExtra`. The first-match `break` means a class is never double-counted.
- **Aggregation basis is sound.** Summing per-class `coveredlines`/`coverablelines`/`coveredbranches`/`totalbranches` over the in-scope classes reproduces the assembly totals on the same basis as the assembly gate, so the namespace bucket is consistent with the existing rows. No double-report: `Game.Api` still appears as a measure-only row alongside the gated slice, which is the right presentation.
- **Empty-bucket safety.** A rename/removal that drops `CoverableLines` to 0 fails loudly as `MISSING` rather than passing on a 0/0 — exactly the footgun this kind of namespace gate invites, and it's handled.
- **Ratchet seeding is consistent.** Integer floors (line 94 / branch 92) seeded below current (94.8% / 92.2%), matching how the existing gated assemblies are seeded.
- **CI wiring.** Adding `build/**` to the backend path filter (`run-tests.yml`) is necessary so the gate actually validates on build-only changes — correctly placed in the `backend` filter.
- **Docs** are in the right place and at the right altitude (the "strict floors where logic concentrates" principle applied below assembly granularity).

### Nits

- `coverage-gate.ps1:118` — the `$null -ne $ns.LinePct` guard is unreachable as written (the `CoverableLines -eq 0` check above already `continue`s, so `LinePct` is always set here). Harmless and defensive; fine to leave.

### Test coverage

No PowerShell test harness exists in the repo, so the gate script (like its existing assembly-gate logic) isn't unit-tested — consistent with the current setup and not a regression. Manual verification in the PR description (pass-at-current, floor-breach, emptied-bucket, and an independent recompute of the 1165/1229 line and 212/230 branch counts) covers the meaningful branches. Acceptable for build infra here.

### Follow-up (optional, pre-existing)

- The summary table's `Out-String` (`coverage-gate.ps1:139`) renders empty under a width-0/non-TTY host. Pre-existing and cosmetic only — failures print separately via the `$failures` list, so the gate's correctness is unaffected. A one-line `-Width` would harden it; worth a small follow-up issue rather than blocking this PR.

### What's good

Genuinely tidy solution to an awkward problem — gating a logic-heavy slice of a measure-only assembly without a second filtered report, and failing loud instead of silent when the bucket empties. Documentation and seeding are exactly in line with the project's coverage conventions.

Not merging yet: `test-backend` (which runs the gate) is still in progress.

---

Issue #1050, opened 2026-06-20
Retirement: take retired challenges and zones out of circulation (challenge completion + zone navigation)


ginderjeremiah:
Follow-up to #296. This is one of three scoped sub-issues that resolve the deferred per-entity retirement edge cases. It implements the two cases where "retired" currently fails to mean **out of circulation**, applying the established principle:

> Retirement removes content from circulation (new exposure/acquisition) but never retroactively revokes what a player has already obtained or the access they already have.

## A. Retired challenges stop being completable by new players

Today `PlayerProgress.EvaluateChallenges` (`Game.Core/Progress/PlayerProgress.cs:121`) only skips *already-completed* challenges — it never checks `Challenge.RetiredAt`. So a player who has not yet completed a since-retired challenge can still accrue progress, complete it, and unlock its reward. This is inconsistent with the enemy precedent (a retired enemy is dropped from random spawns), where "retired" already means out of circulation.

**Decision (per #296 discussion): retired = out of circulation for challenges too.**

- In `EvaluateChallenges`, skip a challenge that is **retired AND not already completed**. A challenge a player has already completed keeps its reward/unlock (an existing reference — never revoked). Nobody new can complete a retired challenge.
- Frontend: hide retired-but-incomplete challenges from the challenges screen (an already-completed retired challenge stays visible as completed). Keep the reference-data lookup sites unchanged — retired challenges must remain resolvable by id.
- Note the downstream consequence: with this rule, retiring a challenge that is a zone's `UnlockChallengeId` permanently locks new players out of that zone. That is surfaced as a warning in the referenced-by sub-issue (below), not blocked here.

## B. Retired zones are not navigable; an occupant is lazily relocated

Zone retirement currently has no effect on navigation or on a player idling in the zone. Decision (per #296): **a retired zone leaves circulation (is not navigable), and a player whose current zone is no longer viable is lazily relocated** rather than left stranded or hard-kicked.

- **Frontend:** exclude retired zones from navigable zones — `isZoneUnlocked`/zone-progression (`UI/src/lib/common/zone-progression.ts:18`) and the nav arrows (`UI/src/routes/game/screens/fight/ZoneNav.svelte`) treat a retired zone like a locked neighbour (not navigable).
[...]

Issue #296, opened 2026-06-09
Define per-entity retirement edge-case semantics (zones, challenges, in-session)


ginderjeremiah:
Follow-up to #289 (retire lifecycle for reference data).

#289 established the retire lifecycle and the one concrete circulation rule that was localized and clearly correct (a retired enemy is excluded from random spawn rolls). Several per-entity edge cases were intentionally deferred because they need design decisions rather than a mechanical fix:

## Edge cases to resolve

- **Player currently idling in a zone that gets retired.** Should the player be moved out / blocked from re-entering, or left until they navigate away? (Retirement currently has no effect on a zone the player is already in.)
- **In-progress challenge whose target entity or reward becomes retired.** Does it still complete and grant? (Today it does — the reward id still resolves.) Is that the desired behaviour?
- **Retiring a challenge that is a zone's `UnlockChallengeId`.** A retired-but-completed challenge still unlocks the zone (id resolves). Should retiring a gate-challenge be warned/blocked?
- **Retiring an enemy that is a zone's authored boss (`BossEnemyId`).** The boss still resolves and is fightable. Confirm that's intended or surface a warning.
- **"Referenced by …" warning on retire.** Surface what currently references a record before retiring it (e.g. "this enemy is the boss of Zone X / appears in N challenges").

## Notes

- All of these are safe today (every id still resolves), so this is about *intended product behaviour*, not correctness. See `docs/game-design.md` → Content Lifecycle and `docs/backend.md` → Retiring reference data.

ginderjeremiah:
Adding a specific edge case surfaced in the #298 review to this issue's scope:

**Last active enemy in a live zone gets retired.** If an admin retires *every* enemy assigned to an active (non-retired) zone, that zone drops out of `_zoneEnemyTables`, and the next idle `GetRandomEnemy(zoneId)` throws `InvalidOperationException("Zone {zoneId} has no enemies to spawn.")` — players idling there hit an exception on their next encounter. It's admin-only and not a regression (hard-deleting all of a zone's enemies did the same), but it should fail at authoring time rather than at spawn time. Preferred fix: a "this is the last active enemy in live zone X" guard/warning when retiring (and/or the same check when retiring a zone's authored boss). Fits alongside the other authoring-time guards in this issue.

---
_Generated by [Claude Code](https://claude.ai/code)_

ginderjeremiah:
[...]

PR #1063, opened 2026-06-20
Block retiring the last active enemy of a live zone (#1051)

### LINKED ISSUE #1051: Retirement: prevent (and gracefully recover from) a zone left with no spawnable enemies
Follow-up to #296 (surfaced in the #298 review). One of three scoped sub-issues for the deferred retirement edge cases. This is the only case that produces an actual runtime exception, so it is the one **hard block** (the rest are warnings — see the referenced-by sub-issue).

## Problem

If an admin retires **every** active enemy assigned to a live (non-retired) zone, that zone drops out of `_zoneEnemyTables` (`Game.DataAccess/Repositories/Caching/EnemiesCacheHolder.cs:73-83` filters retired enemies), and the next idle encounter throws:

```
InvalidOperationException("Zone {zoneId} has no enemies to spawn.")   // Game.DataAccess/Repositories/Enemies.cs:58
```

— so a player idling there hits an exception on their next `GetRandomEnemy`. It is admin-only and not a regression (hard-deleting all of a zone's enemies did the same), but per #296 it should **fail at authoring time, not at spawn time**.

## Decision (per #296): warn + one hard block

Two complementary layers:

1. **Authoring-time hard block (backend-enforced).** Reject retiring an enemy when it is the **last active spawn enemy of a live (non-retired) zone**. This must live in the backend admin save path (`AdminEnemies` + `ChangeSetProcessor`/`AdminSaveResult`) because a tampered/scripted admin client could otherwise bypass a frontend check. The check: for each zone the retiring enemy spawns in, if the zone is not retired and the enemy is the only non-retired enemy in that zone's spawn table, reject with a clear message ("Enemy X is the last active spawn in live Zone Y; retire the zone first or add another enemy"). Workflow stays clean: retire the *zone* first, then its enemies.
   - `AdminSaveResult` is binary success/failure today (`Game.Abstractions/DataAccess/Admin/AdminSaveResult.cs`); this is a `Failure`, so no new result shape is needed here.
2. **Runtime safety net (defense in depth).** Reuse the "zone not viable" lazy-relocation from the companion behavior sub-issue (retired OR no spawn table) so that even if a zone ends up with no spawnable enemies — e.g. a player grandfathered/reconnecting into an already-retired zone, which the live-zone block deliberately does not cover — the next encounter relocates the player to a viable zone instead of throwing.

## Acceptance

- Retiring the last active spawn enemy of a **live** zone is rejected at save time with a descriptive error; the same retire is allowed once the zone itself is retired.
- A player whose current zone has no spawnable enemies is relocated (never hits the `GetRandomEnemy` throw).
- Integration test for the authoring rejection (API round-trip) and the "allowed after zone retired" path; a test that the runtime path no longer throws for an empty/retired zone.
- Update `docs/backend.md` → Retiring reference data with the authoring guard.

## Dependency

Builds on the lazy-relocation predicate introduced in the behavior sub-issue; can be implemented after it (or share the predicate).

ginderjeremiah:
## Summary

Closes #1051 (sub-issue of #296). Prevents an admin from authoring a live zone into a no-spawnable-enemies state that throws at spawn time, and confirms the runtime safety net handles the case it can't.

If an admin retired **every** active enemy of a live (non-retired) zone, that zone dropped out of `_zoneEnemyTables` and the next idle `GetRandomEnemy(zoneId)` threw `InvalidOperationException("Zone {zoneId} has no enemies to spawn.")` — a player idling there hit an exception on their next encounter. Per #296 this should fail at **authoring time, not spawn time**.

## How it addresses the issue

**1. Authoring-time hard block (backend-enforced).** `AdminEnemies.SaveEnemies` now rejects retiring an enemy that is the last active spawn of a live zone, before delegating to the change-set processor — mirroring how `AdminZones` validates its boss/unlock references up front. It reports through the existing `AdminSaveResult.Failure` (a 400, no new result shape), with a descriptive message:

> Retiring 'Lone Goblin' would leave live zone 'Forest' with no spawnable enemies. Retire the zone first, or keep at least one active enemy spawning in it.

The check is computed from a post-save retirement view (combining the batch's edits with cached state) so it is correct for multi-enemy batches, and it only fires when *this save* empties a currently-viable live zone. A missing/already-retired zone is skipped (left to the runtime net), and a malformed duplicate-key batch is deferred to the processor's existing guard. Living in the backend save path means a scripted/tampered admin client can't bypass it. Workflow stays clean: **retire the zone first, then its enemies.**

**2. Runtime safety net (defense in depth).** Already delivered by #1050's lazy relocation (`EnsureViableZone` / `IsZoneViable` over `IZones.IsZoneRetired` + `IEnemies.HasSpawnableEnemies`). This PR adds a test pinning the exact #1051 manifestation — a live zone whose only spawn enemy is *retired* (so the table empties) relocates instead of throwing — covering the grandfathered/reconnect case the live-zone block deliberately doesn't.

## Test plan

All backend tests green (`dotnet build` + full `dotnet test`: 1676 passed).

- **Authoring rejection — API round-trip** (`AdminToolsControllerTests`): POST retire of a live zone's last enemy → 400 + message, enemy stays active.
- **Repo-level** (`AdminEnemiesIntegrationTests`): rejection (no staging) · allowed once the zone is retired · allowed when a sibling enemy remains active · allowed when the enemy spawns in no zone.
- **Runtime** (`BattleServiceIntegrationTests`): a live zone whose only spawn enemy was retired relocates the player to the nearest viable zone (no throw).
- Existing duplicate-batch / retire / delete-rejection tests still pass (the guard defers malformed batches to the processor).

## Docs

`docs/backend.md` → *Retiring reference data* now documents the authoring guard alongside the existing runtime-relocation note.

https://claude.ai/code/session_0184e6yrqbixVq64NMYbFf5r

---
_Generated by [Claude Code](https://claude.ai/code/session_0184e6yrqbixVq64NMYbFf5r)_


ginderjeremiah (commented):
## Review: Block retiring the last active enemy of a live zone (#1051)

**Verdict:** Approve — _(posted as a comment because I'm the PR author; GitHub won't accept a self-gating approval)_. The change satisfies every acceptance criterion in #1051, follows the established admin-save validation pattern, and is well tested. No blocking findings.

### What's good

- **Right placement, right pattern.** The authoring guard lives in `AdminEnemies.SaveEnemies` and runs up front before delegating to `ChangeSetProcessor.Apply`, mirroring how `AdminZones.SaveZones` validates its boss/unlock references. This keeps it backend-enforced (a tampered admin client can't bypass it), exactly as #1051 requires, and reuses the existing `AdminSaveResult.Failure` → 400 path with no new result shape.
- **Viability check matches runtime semantics.** `FindLiveZoneLeftEmpty` treats a zone as viable when it has any non-retired spawner (`AdminEnemies.cs:204-205`), which is precisely how `BuildZoneEnemyTables` (`EnemiesCacheHolder`) builds `_zoneEnemyTables` — filter `RetiredAt is null`, group by zone, weight-independent. So the authoring projection and the runtime `HasSpawnableEnemies` won't diverge.
- **Edge cases handled correctly.** The post-save projection (`WillBeRetired`) combines the batch's edits with cached state, so a batch that retires the only active spawner *and* reinstates a previously-retired sibling in the same zone is correctly allowed; a duplicate-key batch is deferred to the processor's shared guard rather than throwing here; and missing/already-retired zones are skipped to the runtime safety net. The `wasViable && !staysViable` gate ensures only a save that *itself* empties a currently-viable live zone is blamed.
- **Clean, well-documented code.** Always-braced blocks, no null-forgiving operator, concise explanatory comments, and a focused one-bullet docs update in `docs/backend.md` → Retiring reference data.

### Test coverage

Thorough and aligned with the issue's acceptance list:
- API round-trip rejection + enemy stays active (`AdminToolsControllerTests`).
- Repo-level matrix (`AdminEnemiesIntegrationTests`): rejected with no staging · allowed after the zone is retired · allowed when a sibling stays active · allowed when the enemy spawns in no zone.
- Runtime safety net (`BattleServiceIntegrationTests`): a live zone whose only spawner was retired relocates instead of throwing — pinning the exact #1051 trigger.

### Nits

- **Optional test for the multi-enemy / reinstatement projection.** The PR description (and the code) claim correctness for multi-enemy batches and the post-save reinstatement case, and the logic genuinely handles them — but neither has a dedicated test. A small case like "retire the only active spawner while reinstating a previously-retired sibling in the same zone → allowed" (or "two enemies both stranding the same zone → one rejection") would pin behavior that's currently only covered indirectly. Not blocking.

CI is still running; I'll hold off on merging until it's green.

---

ginderjeremiah:
Thanks for the review. Addressed the optional nit in `8fbeb82` — added two repo-level tests in `AdminEnemiesIntegrationTests`:

- `SaveEnemies_RetiringLastActiveEnemyWhileReinstatingSibling_Succeeds` — retire the only active spawner while reinstating a previously-retired sibling in the same zone → allowed (the post-save projection sees the reinstated sibling).
- `SaveEnemies_BatchRetiringAllActiveEnemiesOfLiveZone_ReturnsFailure` — two enemies that individually aren't the last spawner but together strand the zone → one rejection.

Also refactored the test helper into `EnemyEditChange(id, name, retiredAt)` so retire and reinstate edits share one builder. Full backend suite stays green locally.

---

Issue #1040, opened 2026-06-20
Offline rewards: persist active idle-loop mode (idle vs auto-challenge-boss)


ginderjeremiah:
Sub-issue of #879 (offline time tracking & rewards spike). No hard dependencies (pairs with the away-time foundation; consumed by the orchestration sub-issue).

See the spike design doc: `docs/spikes/879-offline-rewards.md` (decisions 3 & 5).

## Goal

Persist what the player's idle loop was doing — idle-farming a zone vs. auto-challenging a zone's boss — so the offline simulation at next login can continue the correct loop. Today the auto-challenge-boss state is a **frontend-only, in-memory `EnemyManager.autoFight` flag** that resets on loss/draw/stop and is never sent to the backend.

## Scope

Backend:
- Persist the active loop mode (idle / boss) on the durable `Player` aggregate — synergizing with `CurrentZoneId`, which already persists the idle location — so it is available at next login regardless of how long the player was away (up to the cap). Persist via the existing write-behind path.
- A new lightweight socket command (e.g. `SetAutoChallengeBoss(enabled, zoneId)`) to sync the mode from the frontend. Auto-discovered by `SocketCommandFactory`; validates as anti-cheat (zone exists/unlocked) like the other player mutations.
- Mirror the online auto-fight-off as a backstop: when the backend records a dedicated-boss **loss or draw** (`BattleLost` / abandon resolving as a draw), reset the persisted mode to idle, so the persisted mode tracks the player's actual present-loop even if a frontend sync is missed.

Frontend:
- Call the sync command when `AutoFightToggle` changes and when the boss loop returns to idle (`EnemyManager.returnToIdle`), so the persisted mode mirrors the live `autoFight` state.
- Regenerate TS types for the new command.

## Notes

- This is the data the offline sim reads to pick idle vs boss (decision 3). Offline does **not** switch modes mid-run; it loops the persisted mode's battle type for the whole budget.

## Tests

- Unit: the aggregate mutation sets/clears mode.
- Integration: the command persists and round-trips; anti-cheat rejects an invalid/locked zone.
- Frontend: toggling auto-fight / returning to idle issues the sync command.

Issue #879, opened 2026-06-17
Add offline time tracking/rewards


ginderjeremiah:
Track last enemy defeated -> next login time, simulate rewards if > 5 minutes time spent away.

Do a sample of 10 (tunable?) battles in the zone the player was idling in, and extrapolate the number of battles (and victory/defeat odds) -> average rewards.

Alternatively, we could attempt to simulate all the battles... Based on performance tests, I would estimate a reasonable worst-case for battle simulation time to be 10ms per 1 minute of simulation. Depending on the length of the battles, that could be a significantly higher ratio of away time due to enemy spawn cooldowns. Ignoring cooldowns, that would lead to a worst-case scenario of 3 seconds of simulation time per 5 hours of away time. This might actually be reasonable, but the login flow would need to have an "away time" gate before loading into the game. Ideally, the UI would also have a way to show the progress of the simulation as it occurs on the backend, which might require something like: socket command to start simulation -> server responds with initial time calculation response, and then asynchronously sends additional socket command responses back down as the simulation progresses.

Limit to 10 hours of away time if fully simulating, otherwise don't set a limit.

Need to respect the auto-challenge boss setting.

ginderjeremiah:
Spike complete. Full design recorded in [`docs/spikes/879-offline-rewards.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/879-offline-rewards.md).

**Decided direction:** full server-side simulation (not sampling), run inline behind a "welcome back" gate at login, capped at 10h, triggered above 5 min away. Both idle-farming and auto-challenge-boss loops are simulated — each just loops its battle type for the whole away budget (boss farming continues through losses/draws, since there's no present player to turn auto-fight off).

The key enabler from the code: offline, a player's combat power and reward distribution are **stationary** (combat reads only allocations + equipment, level is unused in combat math, enemies scale with the zone) and the 5s cooldown throttles battle count — so a full 10h replay costs only ~a few seconds of CPU and is exact.

**Sub-issues (linked above):**
- #1039 — away-time tracking foundation (`LastActivity`) *(scope: small)*
- #1040 — persist active idle-loop mode (idle vs auto-boss) *(scope: medium)*
- #1041 — core offline simulation engine, `Game.Core` *(scope: large)*
- #1042 — reward application + challenge consolidation + orchestration + socket command *(scope: large; depends on #1039/#1040/#1041)*
[...]

PR #1064, opened 2026-06-20
Add Player.LastActivity away-time tracking foundation (#1039)

### LINKED ISSUE #1039: Offline rewards: away-time tracking foundation (Player.LastActivity)
Sub-issue of #879 (offline time tracking & rewards spike). Foundation — no dependencies.

See the spike design doc: `docs/spikes/879-offline-rewards.md` (decision 4).

## Goal

Maintain `Player.LastActivity` as a real "last active" timestamp so the offline-rewards flow can compute how long a player was away (`now − LastActivity`), server-side. Today the column exists on the `Player` DB entity but is set only at player creation and read only for the admin player summary — nothing updates it during play, and it is not on the core `Player` aggregate at all.

## Scope

- Add `LastActivity` to the core `Game.Core/Players/Player.cs` aggregate (it currently lives only on the EF entity + DB).
- Stamp it on battle completion — the issue's "last enemy defeated" anchor — so it is current at any disconnect. Battle-end already calls `SavePlayer`, so this rides the existing save.
- Provide a way to set it to "now" when offline progress is computed at login (consumed by the orchestration sub-issue), so the next away period starts fresh and a re-claim is a no-op.
- Persist it through the existing write-behind path (an absolute write; mirror the existing core-field handler in `Game.DataAccess/PlayerUpdates/Handlers`).
- Map it on load (`PlayerMapper`) so it is available on the in-memory aggregate without an extra query (Redis cache is source of truth, 48h TTL; DB fallback on miss).

## Notes

- Away time is computed **server-side only**, never client-supplied.
- Anchoring on battle completion slightly over-counts away time for a session that connected but never battled — acceptable (no reward was earned then, and the 10h cap bounds it). Stamping on clean disconnect/logout too is a possible later refinement.

## Tests

- Unit: the aggregate stamps `LastActivity` on the relevant mutation(s); the setter updates it.
- Integration: the value round-trips through the write-behind persistence + reload.

ginderjeremiah:
## Summary

Closes #1039 (foundation sub-issue of #879, offline rewards). Maintains `Player.LastActivity` as a real "last active" timestamp so the offline-rewards flow can compute how long a player was away (`now − LastActivity`) **server-side**. Today the column exists on the `Player` DB entity but is set only at player creation and read only for the admin summary — nothing updates it during play, and it isn't on the core aggregate at all.

No user-facing effect on its own; this is the anchor every other offline-rewards piece (#1040–#1042) reads.

## How it addresses the issue

- **Core aggregate.** Adds `LastActivity` to `Game.Core/Players/Player.cs` (previously only on the EF entity + DB).
- **Stamped on battle completion.** New `StampActivity(DateTime timestamp)` sets the field and raises `PlayerCoreUpdatedEvent`. `RecordBattleCompleted` now takes a server timestamp and calls it, so **every** completion path — win, loss, and abandon (incl. won-abandon) — advances the anchor. A loss otherwise raises no core update, so this is what carries its anchor onto the Players row. `StampActivity` is also the reusable "set to now" entry point the offline-claim orchestration (#1042) will call at login (resetting it makes a re-claim a no-op).
- **Persisted via the existing write-behind path.** `LastActivity` rides `PlayerCoreUpdatedEvent` → `PlayerCoreUpdatedHandler` (absolute write), so it persists on the save battle-end already performs — no new event type on the hot path beyond the loss case that previously wrote nothing core.
- **Mapped on load.** `PlayerMapper.ToCore` reads `entity.LastActivity` onto the in-memory aggregate (Redis cache is source of truth; this is the DB-fallback path). The aggregate also serializes into the Redis cache automatically.
- **Anchored at creation.** `PlayerMapper.ToEntity` now stamps `LastActivity = UtcNow` for a brand-new player so its first login computes a fresh (near-zero) away period rather than an enormous one from the default `DateTime` — a small correctness fix, since the creation mapper wasn't setting it.

## Notes / design

- Away time is computed **server-side only**, never client-supplied — the timestamp comes from `BattleService`'s server clock (matching its existing `DateTime.UtcNow` usage), passed into the domain so the aggregate stays pure/testable.
- On a **victory**, `GrantExp` already raises a core update; the stamp raises a second carrying the new `LastActivity`. The write-behind handlers are idempotent absolute writes and the final value converges correctly — the extra update is on the async/batched queue, off the request path. Accepted for the robustness of a single completion method that no path can forget; the loss path inherently needed a new core update regardless.
- No frontend changes: `LastActivity` is server-only, not part of any battle-parity surface, and the domain event isn't code-generated.

## Test plan

Full backend suite green (`dotnet build` + `dotnet test`: 2073 passed, 0 failed).

- **Unit** (`PlayerTests`): `StampActivity` sets the field and raises a `PlayerCoreUpdatedEvent` carrying it; `RecordBattleCompleted` stamps `LastActivity` and raises the core update (loss path).
- **Mapping** (`PlayerMapperTests`): `ToCore` maps `LastActivity` from the entity.
- **Integration** (`PlayerUpdateHandlerIdempotencyIntegrationTests`): `PlayerCoreUpdatedEvent` round-trips `LastActivity` through the write-behind handler to the DB and converges to the latest value on re-apply.
- Updated the existing `PlayerCoreUpdatedEvent` constructions in `DataProviderSynchronizerTests` and the `PlayerBuilder` for the new field.

## Docs

No docs change in this PR: the foundation has no user-facing effect, and the offline-rewards architecture (including this anchor) is captured in `docs/spikes/879-offline-rewards.md`, to be promoted to `docs/backend.md` when the feature itself lands (per the spike's documentation note).

Closes #1039

---
_Generated by [Claude Code](https://claude.ai/code/session_01GjQifDXv9TQzibE6LZrEAn)_

### DIFF
--- a/Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs
+++ b/Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs
@@ -44,7 +44,8 @@ namespace Game.Application.Tests.DataAccess
                 Exp: 1234,
                 CurrentZoneId: 0,
                 StatPointsGained: 100,
-                StatPointsUsed: 100);
+                StatPointsUsed: 100,
+                LastActivity: DateTime.UtcNow);
 
             var logger = new CapturingLogger<DataProviderSynchronizer>();
             var pubsub = scope.ServiceProvider.GetRequiredService<IPubSubService>();
@@ -86,8 +87,8 @@ namespace Game.Application.Tests.DataAccess
             var logger = new CapturingLogger<DataProviderSynchronizer>();
             var synchronizer = new DataProviderSynchronizer(brokenServices, pubsub, logger, TestRetryPolicy);
 
-            var firstEvent = new PlayerCoreUpdatedEvent(1, 2, 3, 0, 100, 100);
-            var secondEvent = new PlayerCoreUpdatedEvent(2, 3, 4, 0, 100, 100);
+            var firstEvent = new PlayerCoreUpdatedEvent(1, 2, 3, 0, 100, 100, DateTime.UtcNow);
+            var secondEvent = new PlayerCoreUpdatedEvent(2, 3, 4, 0, 100, 100, DateTime.UtcNow);
             var queue = new InMemoryPubSubQueue(Serialize(firstEvent), Serialize(secondEvent));
 
             await synchronizer.ProcessQueue(queue);
@@ -120,7 +121,8 @@ namespace Game.Application.Tests.DataAccess
                 Exp: 4321,
                 CurrentZoneId: 0,
                 StatPointsGained: 100,
-                StatPointsUsed: 100);
+                StatPointsUsed: 100,
+                LastActivity: DateTime.UtcNow);
 
             var pubsub = scope.ServiceProvider.GetRequiredService<IPubSubService>();
             var logger = new CapturingLogger<DataProviderSynchronizer>();
@@ -719,7 +721,8 @@ namespace Game.Application.Tests.DataAccess
                 Exp: 4242,
                 CurrentZoneId: 0,
                 StatPointsGained: 100,
-                StatPointsUsed: 100);
+                StatPointsUsed: 100,
+                LastActivity: DateTime.UtcNow);
 
             var realPubSub = scope.ServiceProvider.GetRequiredService<IPubSubService>();
             var queue = realPubSub.GetQueue(Constants.PUBSUB_PLAYER_QUEUE);
@@ -838,7 +841,8 @@ namespace Game.Application.Tests.DataAccess
                 Exp: 5678,
                 CurrentZoneId: 0,
                 StatPointsGained: 100,
-                StatPointsUsed: 100);
+                StatPointsUsed: 100,
+                LastActivity: DateTime.UtcNow);
 
             var realPubSub = scope.ServiceProvider.GetRequiredService<IPubSubService>();
 
@@ -885,7 +889,8 @@ namespace Game.Application.Tests.DataAccess
                 Exp: 9876,
                 CurrentZoneId: 0,
                 StatPointsGained: 100,
-                StatPointsUsed: 100);
+                StatPointsUsed: 100,
+                LastActivity: DateTime.UtcNow);
 
             var realPubSub = scope.ServiceProvider.GetRequiredService<IPubSubService>();
             var queue = realPubSub.GetQueue(Constants.PUBSUB_PLAYER_QUEUE);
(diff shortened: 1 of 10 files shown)


ginderjeremiah (commented):
## Review: `Player.LastActivity` away-time tracking foundation (#1039)

**Verdict:** Approve — clean, well-scoped foundation that matches the issue and the #879 spike (decision 4); no blocking findings. _(Submitted as a `COMMENT` because GitHub won't accept a self-approval on my own PR; the intended verdict is Approve.)_

### Correctness review

- The aggregate change is the right shape: `LastActivity` lives on `Game.Core/Players/Player.cs`, mutated through a domain method (`StampActivity`) that raises `PlayerCoreUpdatedEvent`, so it rides the existing write-behind path with no new hot-path event type. Domain stays pure — the server clock is passed in from `BattleService`, never client-supplied (matches the spike's server-side-only constraint).
- Every completion path advances the anchor: `RecordBattleCompleted` now stamps, so win, loss, and won-abandon all converge. The loss path previously raised no core update, so routing the stamp through it is exactly what carries the anchor onto the row.
- The new **double** `PlayerCoreUpdatedEvent` on a victory (GrantExp raises one with the old `LastActivity`, then `StampActivity` raises a second with the new one) is safe: `SavePlayer` drains the batch FIFO and the handler does absolute `SetProperty` writes, so the later event wins; and the Redis cache — the source of truth — is written from the in-memory aggregate that already holds the final value. The redundant DB write is off the request path and acknowledged in the PR description. Reasonable tradeoff for a single completion method no path can forget.
- No migration needed — the `LastActivity` column already exists on the EF entity (non-nullable `DateTime`). `ToCore` now maps it, `ToEntity` anchors it at creation, and the handler persists it. All consistent.

### Test coverage

Good and matches the issue's required coverage:
- Unit (`PlayerTests`): `StampActivity` and the loss-path `RecordBattleCompleted` both set the field and raise the carrying event.
- Mapping (`PlayerMapperTests`): `ToCore` maps `LastActivity`.
- Integration (`PlayerUpdateHandlerIdempotencyIntegrationTests`): round-trips through the write-behind handler and converges to the latest value on re-apply.

### Minor (non-blocking, no action needed)

- The creation-time stamp in `PlayerMapper.ToEntity` (`LastActivity = DateTime.UtcNow`) is untested, but it's a trivial `UtcNow` assignment with no clock seam — testing it deterministically would mean injecting a clock into the static mapper, which would be over-engineering for this foundation. Measure-only is fine here.

### What's good

- Domain isolation is exactly right (no logic leaked into `BattleService`; it only supplies the timestamp).
- The XML docs on `StampActivity` and the `ToEntity` comment explain the *why* (re-claim no-op at login, fresh-vs-default away period) without over-documenting.
- Correctly defers docs to when the feature lands, per the spike's documentation note — nothing user-facing ships here.


---

PR #1066, opened 2026-06-21
Persist active idle-loop mode (idle vs auto-challenge-boss) (#1040)

### LINKED ISSUE #1040: Offline rewards: persist active idle-loop mode (idle vs auto-challenge-boss)

ginderjeremiah:
## Summary

Persists what the player's idle loop is doing — idle-farming a zone vs. auto-challenging that zone's boss — so the offline-rewards simulation (spike [#879](https://github.com/ginderjeremiah/GameServer/issues/879), decisions 3 & 5) can resume the correct loop at next login. Today the auto-challenge-boss state is a frontend-only, in-memory `EnemyManager.autoFight` flag that resets on loss/draw/stop and is never sent to the backend.

This is groundwork: nothing consumes the persisted mode yet (the offline sim lands in #1041/#1042), so there is no user-facing change.

## How it addresses the issue

**Backend**
- Persists the active loop mode on the durable `Player` aggregate via a boolean `AutoChallengeBoss`. The boss farmed is **always the current zone's boss** — the loop never targets a separate zone — so the mode is a flag, not a stored zone: `false` = idle-farming `CurrentZoneId`, `true` = auto-challenging its boss. It rides the existing core write-behind path (`PlayerCoreUpdatedEvent` → handler `ExecuteUpdate`) and a new boolean EF column/migration.
- New lightweight `SetAutoChallengeBoss(enabled)` socket command, auto-discovered by `SocketCommandFactory`. Enabling validates the player's own `CurrentZoneId` as anti-cheat exactly like a boss challenge (in circulation, unlocked, has a dedicated boss) by reusing `BattleService`'s existing zone/unlock checks; disabling always succeeds. The farmed zone is never client-supplied.
- Backstop mirroring the online auto-fight-off: when the backend records a dedicated-boss **loss or draw** (`BattleLost` / an abandon resolving as a draw), the persisted mode resets to idle — folded into `RecordBattleCompleted` and carried by the single core-updated event it already raises.

**Frontend**
- Syncs the mode from the live `autoFight` state when `AutoFightToggle` changes and when the boss loop returns to idle (`EnemyManager.returnToIdle`); only the enabled flag is sent (no zone). Teardown (`stop()`) deliberately does **not** sync, so a disconnecting boss-farmer's persisted intent isn't clobbered to idle right before disconnect.
- Regenerated TS client types for the new command.

## Design notes
- The boss zone is implicitly `CurrentZoneId` (kept aligned by design), so a boolean captures the mode with no redundant/divergent zone field.
- **Deferred to [#1067](https://github.com/ginderjeremiah/GameServer/issues/1067):** the pre-armed-while-idle semantic (auto-fight can be toggled on from `BossTrigger` while the loop is still idle, persisting boss intent before any challenge) and a sync-dedup nit. Harmless until a consumer reads the field; a code comment at the sync site flags it.
- **Re-entry reconciliation is out of scope** (a returning boss-farmer's live `autoFight` resets to idle while the persisted mode stays boss until the next sync) — per the spike, the welcome-back gate (#1043) owns re-syncing authoritative state.

## Test plan
- [x] **Unit (domain):** `SetAutoChallengeBoss` sets/clears the flag and raises a core-updated event carrying it; the `RecordBattleCompleted` backstop resets on boss loss/draw, keeps boss mode on a boss win, never touches it on an idle loss, and rides a single core-updated event.
- [x] **Integration:** the command persists and round-trips through the write-behind cache; anti-cheat rejects a locked / bossless / retired current zone; the boolean column round-trips both directions through the `PlayerCoreUpdated` write-behind handler (with an intermediate assertion proving the set direction landed); the socket command covers enable / disable / bossless-zone-rejection.
- [x] **Mapper:** `ToCore` maps both `false` and `true`.
- [x] **Frontend:** toggling auto-fight on/off and returning to idle (retreat / boss loss / boss draw) issue the sync command; teardown does not.
- [x] Full backend suites + frontend (2254) green; `dotnet format --verify-no-changes`, `npm run lint`, and `svelte-check` clean; codegen regenerated (no drift).

Closes #1040


UI/src/lib/engine/battle/enemy-manager.ts
```diff
@@ -286,13 +288,33 @@ export class EnemyManager {
 	/** Toggle auto-fight: when on, a boss victory immediately re-challenges the boss. */
 	public setAutoFight(on: boolean) {
 		this.autoFight = on;
+		this.syncAutoChallengeBoss(on);
```
ginderjeremiah, line 295: **Should-fix (semantic divergence — design call worth making explicit):** the sync mirrors `autoFight` (the player's *intent*), but `autoFight` can be toggled on while the live loop is still **idle**. `BossTrigger.svelte` renders the `AutoFightToggle` in the not-engaged state, so a player can pre-arm auto-fight *before* pressing Challenge — at which point `setAutoFight(true)` persists `AutoChallengeBossZoneId = currentZone` even though the loop is still idle-farming.

Today this is harmless (no consumer reads the field, and re-entry reconciliation is deferred to #1043). But once the offline sim (#1041/#1042) lands and loops the persisted mode for the whole budget, a player who pre-armed auto-fight and then walked away *without ever challenging* would be resumed as a boss-farmer instead of an idle-farmer.

Worth a conscious decision: either gate boss-enable on actually being in the boss loop (`mode === 'boss'`), or affirm that "auto-fight armed" is the intended persisted semantic — and file a follow-up if deferring, so it isn't lost before the consumer arrives.

---

UI/src/lib/engine/battle/enemy-manager.ts
```diff
@@ -88,7 +88,9 @@ export class EnemyManager {
 			// Cut short an in-flight retry backoff so a getNewEnemy parked under a sustained outage exits
 			// at once on the cleared `started` flag rather than after the full backoff window.
 			this.interruptFetch();
-			this.returnToIdle();
+			// Teardown (screen unmount / session end), not a user intent change, so don't sync the persisted
+			// loop mode — clobbering it to idle here would lose a disconnecting boss-farmer's mode.
+			this.returnToIdle(false);
```
ginderjeremiah, line 93: **Should-fix (missing test for a deliberate branch):** the `sync` parameter on `returnToIdle` exists *solely* so teardown here passes `false` and doesn't clobber a disconnecting boss-farmer's persisted mode. That's exactly the kind of subtle, intentional construct a later refactor could naively "simplify" back to `returnToIdle()` — silently reintroducing the clobber. The test plan claims "teardown does not [sync]", but no test pins it (the `stop()` test at `enemy-manager-boss.test.ts:317` doesn't assert it). The existing `stop()` test could add a one-line `expect(send).not.toHaveBeenCalledWith('SetAutoChallengeBoss', expect.anything())` to lock the behavior in.

---

UI/src/lib/engine/battle/enemy-manager.ts
```diff
@@ -286,13 +288,33 @@ export class EnemyManager {
...
+	 * the current zone, off ⇒ idle. Fire-and-forget — anti-cheat validation is the server's, and a transient
+	 * failure only leaves the persisted mode briefly stale, corrected by the next sync or the backend's
+	 * boss-loss/draw backstop.
+	 */
+	private syncAutoChallengeBoss(enabled: boolean) {
+		void apiSocket.sendSocketCommand('SetAutoChallengeBoss', {
+			enabled,
+			zoneId: playerManager.currentZone
+		});
 	}
```
ginderjeremiah, line 320: **Nit:** every `returnToIdle()` path (retreat, boss loss, boss draw, single-victory handoff, failed-challenge fallback) fires `SetAutoChallengeBoss(false)` unconditionally — even when the persisted mode is already idle. On a boss loss it also duplicates the backend's own `BattleLost` backstop. It's fire-and-forget and bounded to boss transitions, so not a real problem, but tracking the last-synced state and only sending on an actual change would cut the redundant socket traffic.

---

ginderjeremiah (commented):
## Review: Persist active idle-loop mode (idle vs auto-challenge-boss) — closes #1040

**Verdict:** Approve (posted as a comment since I'm the PR author — GitHub won't take a gating verdict on one's own PR) — no blocking findings. Two should-fixes and a couple of nits to address or consciously defer before merge.

This is clean groundwork. The single nullable `Player.AutoChallengeBossZoneId` elegantly encodes both *which mode* and *which boss zone* in one field, ruling out an invalid "boss mode with no zone" state. The layering is correct: the domain mutation and the loss/draw backstop live in `Game.Core`, anti-cheat orchestration sits in `BattleService` reusing the existing zone/unlock checks, and the socket command is a thin delegator. It rides the existing core write-behind path rather than inventing a new one. No consumer reads the field yet, so there is no user-facing change and even the semantic question below is harmless today.

### Should-fix
- **Sync mirrors intent, not the active loop** — `enemy-manager.ts:289`. `setAutoFight(true)` persists boss mode, but `BossTrigger.svelte` exposes the auto-fight toggle while the loop is still *idle* (pre-arm before Challenge). A player who pre-arms and walks away without challenging would be resumed as a boss-farmer once the offline sim (#1041/#1042) reads this. Decide whether to gate boss-enable on `mode === 'boss'` or affirm pre-armed intent is the intended semantic — file a follow-up if deferring. (inline)
- **No test pins the deliberate teardown no-sync branch** — `enemy-manager.ts:93`. The `sync` param on `returnToIdle` exists only so `stop()` passes `false`; nothing locks that in, so a later refactor could regress it silently. One assertion on the existing `stop()` test covers it. (inline)

### Nits
- Redundant `SetAutoChallengeBoss(false)` on every `returnToIdle()` even when already idle (and duplicating the backend backstop on a boss loss) — `enemy-manager.ts:313`. Fire-and-forget, low priority. (inline)
- `PlayerUpdateHandlerIdempotencyIntegrationTests.cs:519`'s `Assert.Null(row.AutoChallengeBossZoneId)` would pass even if the set→null write were dropped (null is the column default), so it doesn't actually prove the round-trip its comment claims. The set direction *is* covered by `BattleServiceIntegrationTests.SetAutoChallengeBoss_EnableValidBossZone_PersistsZoneAndRoundTrips`, so this is cosmetic — just tighten the comment or assert the intermediate set.

### Test coverage
Strong on the backend. Domain: set/clear + event payload, and the `RecordBattleCompleted` backstop across boss loss / boss draw / boss win (kept) / idle loss (untouched) / single-event guarantee. Integration: persist + both round-trip directions through the write-behind cache, plus every anti-cheat rejection path (locked, bossless, retired, out-of-range). Mapper covers both null and set. Frontend covers toggle-on, toggle-off, retreat, boss loss, boss draw. The two gaps are the teardown no-sync assertion (should-fix above) and the pre-armed-idle case — both frontend.

### What's good
- Reuses the existing core-updated event and write-behind handler instead of a parallel path; the backstop is folded into the single event `StampActivity` already raises (and a test asserts exactly that).
- Anti-cheat for enable mirrors `StartBossBattle` precisely, with the range check ordered before `GetDomainZone` so a tampered id is rejected rather than throwing.
- Codegen regenerated (not hand-edited), and `docs/backend-battle.md` updated with a concise design note.

---

ginderjeremiah:
Thanks for the review — addressed in 7279aea, plus the coverage-gate fix.

**CI fix:** `test-backend` failed only on the coverage gate (`Game.Api sockets/services/filters` line 93.2% < 94%, branch 90.5% < 92%) — every test passed. The new `SetAutoChallengeBoss` command class was untested at the Api layer. Added `SetAutoChallengeBossSocketTests` (enable-valid / disable / enable-invalid) covering both branches of `ExecuteAsync`. (`all-checks-passed` was the same root cause.)

**Should-fix — teardown no-sync test:** added a dedicated test asserting `stop()` issues no `SetAutoChallengeBoss`, pinning the `returnToIdle(false)` branch against a naive future "simplification".

**Nit — idempotency test:** now asserts the intermediate boss value between the two applies, so the set→null round-trip is actually proven (the final null alone is the column default).

**Should-fix — pre-armed-while-idle semantic, and Nit — redundant syncs:** deferred to **#1067**. #1040's stated scope is "mirror the live `autoFight` state", both are harmless until a consumer reads the field, and the pre-arm gating decision is entangled with the dedup's stale-initial-state subtlety and #1043's login reconciliation — so they're best settled together before #1041/#1042 land. Added a code comment at the sync site flagging the semantic and linking #1067.

Re-verified locally: full backend suites + frontend (now 2254) green, lint and `dotnet format --verify-no-changes` clean.

---

Game.Core/Players/Player.cs
```diff
@@ -28,6 +28,15 @@ public class Player : AggregateRoot
...
         public required DateTime LastActivity { get; set; }
 
+        /// <summary>
+        /// The zone whose dedicated boss the player's idle loop is auto-challenging, or <c>null</c> when the
+        /// loop is idle-farming. Persisted alongside <see cref="CurrentZoneId"/> (which carries the idle
+        /// location) so the offline-rewards simulation can resume the correct loop at next login: idle when
+        /// null, boss-farming this zone otherwise. Synced from the frontend's live auto-fight state via
+        /// <see cref="SetAutoChallengeBoss"/>.
+        /// </summary>
+        public required int? AutoChallengeBossZoneId { get; set; }
```
ginderjeremiah: I tried to reframe the issue to avoid this, but this is not ideal. There is no reason to allow a separate zone ID for challenging bosses vs. idling. They should be kept aligned to the same zone.

ginderjeremiah: Done in 954a391. Replaced the nullable `AutoChallengeBossZoneId` with a boolean `Player.AutoChallengeBoss` — the boss farmed is always the current zone's boss, so the zone is implicitly `CurrentZoneId` and nothing separate is stored. The socket command now takes a bare bool (no client-supplied zone) and validates the player's own `CurrentZoneId` on enable, which also tightens anti-cheat. EF migration reauthored as a boolean column, codegen regenerated, and all tests updated (the invalid-client-zone test is dropped as moot; bossless/locked/retired current-zone rejections remain).

---

Issue #1073, opened 2026-06-21
Spike: recover idle time lost to background-tab throttling (mid-session catch-up)


ginderjeremiah:
Follow-up surfaced by the #922 spike. See `docs/spikes/922-multiple-players-per-account.md`.

## Problem

The idle loop is client-driven via `setInterval` in `logical-engine.ts`, and over-budget time is **discarded** by the `tickSizeX5` clamp (each invocation simulates at most 200 ms of battle; the excess advances the clock without simulating it). Browsers throttle hidden-tab timers (>=1 s, ~once/minute after sustained hiding), so a backgrounded character silently loses idle progress. There is **no `visibilitychange` handling today** — only the render loop's `requestAnimationFrame` pauses when hidden. This affects a **single** character today, independent of multiple-players-per-account; the #922 spike surfaced it.

## Key enabling fact — the in-flight battle is reconstructable, not opaque

A battle outcome is a **pure deterministic function of inputs the server already persists** in `PlayerState` — `ActiveEnemyId` / `ActiveEnemyLevel` / `ActiveEnemySkillIds`, `BattleSeed`, `Snapshot`, `BattleStartTime`, `BattleZoneId`, `IsBossBattle` — plus elapsed time. So "resume a partial battle" needs **no mid-battle state serialization** (HP / cooldowns / effects): reconstruct the battler from the snapshot and **replay from the seed to elapsed time `T`**. That is exactly what the engine does tick-by-tick and what the server already does to anti-cheat-replay a completed battle.

## Mechanism — resolve, credit, fast-forward the remainder

On resume (compute `elapsed = now - BattleStartTime`, and the wider away window from `LastActivity`):

1. **Resolve the in-flight battle.** Replay it from the seed; it concludes at its fixed simulated duration `D`:
   - if `elapsed >= D` → it completed: credit its reward, advance the clock past `D` + cooldown, and continue;
   - if `elapsed < D` → still in progress: hand it to the frontend with its current offset and **fast-forward** the client to `elapsed`, then continue live. Done.
2. **Credit whole intervening battles** for any remaining window via the #879 whole-battle simulator (#1042), until the remaining time is less than one battle + cooldown.
3. **Carry the trailing remainder forward, don't drop it.** Set up the next battle (fresh enemy + seed) as the active battle with its start offset so elapsed-into-it equals the remainder; the client fetches it and fast-forwards to the offset, then continues live.

[...]

PR #1076, opened 2026-06-21
Offline rewards: reward application, challenge consolidation, orchestration & socket command (#1042)

### LINKED ISSUE #1042: Offline rewards: reward application, challenge consolidation, orchestration & socket command

ginderjeremiah:
## Summary

Wires the core offline simulator (#1041) into the live application (spike [#879](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/879-offline-rewards.md)): on return, compute the away time from `Player.LastActivity` (#1039), replay the missed idle/boss battles, apply the accumulated rewards, and expose it as the `GetOfflineProgress` socket command returning a welcome-back summary. Runs inline within the command — the 10h cap keeps the worst case well under the command timeout, and the cancellation token is threaded into the simulation loop.

## How it addresses the issue

`BattleService.SimulateOfflineProgress` (application orchestration):
- Computes away time = `now − LastActivity`. Below the **5-minute floor** it re-anchors `LastActivity` and returns an empty summary (so an immediate re-claim is a no-op); otherwise it proceeds.
- **Resolves any stale in-flight battle first** (a mid-battle disconnect) exactly like an abandon, so its outcome is credited once rather than re-abandoned by the first live `StartBattle` after the gate.
- Resumes the **persisted loop mode** (`Player.AutoChallengeBoss`, #1040) in the player's current zone — boss mode requires a still-challengeable boss, else it falls back to idle in the nearest viable zone (the same lazy relocation the live idle loop uses).
- Calls the #1041 simulator (away clamped to the **10h cap**).
- **Applies exp per simulated victory** via the new `Player.GrantOfflineExp`, which runs each grant through the same per-grant clamp + level-up loop as `GrantExp` but raises a **single** core-update for the whole window (so a thousand-victory window doesn't flood the write-behind queue with one envelope per kill — decision 6).
- **Consolidates statistics + challenges** (decision 7): each battle feeds the same `PlayerProgress.RecordBattleCompleted` the live handler uses; the affected challenges are evaluated **once** at the end and their rewards unlocked, with the live per-challenge push **suppressed** (`Player.CompleteChallenge(..., notify: false)`). The new shared `ChallengeRewardService` factors that evaluate-and-unlock step out of `BattleStatisticsEventHandler` so the live and offline paths share it (DRY) rather than duplicating it.
- Re-anchors `LastActivity = now` and persists once.

Plus the `GetOfflineProgress` socket command (auto-discovered), the `OfflineProgressModel` summary DTO, and regenerated TS types.

### Notes on judgement calls
- **CPU-waste guard.** The merged simulator runs the full budget by design, so the "no progress over the first batch" guard is added as an **opt-in** simulator parameter (`StalemateCutoffBattles`, default off so existing tests are unaffected) that the orchestration sets. I scoped it to the genuinely expensive case — an **opening batch of pure draws** (no win *and* no loss), i.e. an unwinnable-and-unloseable stalemate that would otherwise simulate the whole budget of maximum-duration draws for nothing. Losses are cheap (quick battles) and still credited; any win or loss in the opening batch disables the guard; the 10h cap bounds every run regardless. Happy to widen this to "no wins" if you'd prefer the broader reading of the issue.
- **Push suppression** is verified at the domain level (`Player.CompleteChallenge(notify: false)` raises no `ChallengeCompletedEvent`); the `ChallengeCompletedNotifier` itself lives in `Game.Api` and isn't exercised in the application-layer integration tests.
- The away window includes the (≤2 min) in-flight battle's wall-clock even though that battle is credited separately — the precise mid-session catch-up is the deferred spike #1073, so this V1 keeps it simple as the issue describes.

## Test plan
- **Unit (`Game.Core.Tests`):** `Player.GrantOfflineExp` (batches grants, one core update, per-grant-not-sum clamp, one level-up event per level); `CompleteChallenge(notify: false)` unlocks rewards but raises no completed event; simulator stalemate cutoff stops an all-draw opening batch and does **not** fire on wins/losses.
- **Integration (`Game.Application.Tests`):** multi-battle away applies exp/levels/stat points and re-anchors `LastActivity`; sub-threshold is a no-op; a second immediate call is a no-op; a kill-count challenge crossed mid-window completes once and unlocks its reward (and appears in the summary); a stale in-flight battle is resolved before simulating; boss mode runs the boss loop.
- Updated the three `BattleStatisticsEventHandler` constructions for the new shared-service constructor.
- Full suites green: **1679** backend tests, **2279** frontend tests; `npm run lint` clean; TS regenerated via `dotnet run --project Game.Api -- codegen`.

## Docs
- `docs/game-design.md` — new "Offline Rewards" section.
- `docs/backend-battle.md` — the offline simulation path (authoritative simulator, mode/zone resolution, stale-battle resolution, consolidated/shared reward application, CPU-waste guard).

Closes #1042

---
_Generated by [Claude Code](https://claude.ai/code/session_01EcLYuyMYtEqDAy9WHjxFK6)_


Game.Application/Services/BattleService.cs
```diff
@@ -243,6 +261,187 @@ public async Task<bool> EndBattleLoss(Player player, PlayerState state, Cancella
...
+
+            var progress = await _progressRepo.Load(player, cancellationToken);
+
+            var victoryExpRewards = new List<int>();
+            IReadOnlyCollection<(EStatisticType Type, int? EntityId)> touchedStatistics = [];
+            foreach (var battle in result.Battles)
+            {
+                // RecordBattleCompleted accumulates each battle's stat deltas into the single progress aggregate
+                // and returns the cumulative set of touched rows, so the final call carries the whole window.
+                touchedStatistics = progress.RecordBattleCompleted(
```
ginderjeremiah: **Should-fix (test coverage / subtle coupling).** This consolidation is correct, but only because `PlayerProgress.RecordBattleCompleted` returns `[.. _dirtyStatistics]` and `_dirtyStatistics` is an *aggregate-level* HashSet that accumulates across the loop (the progress aggregate is loaded once for the whole window), so the final call genuinely carries every touched row. The risk is that this contradicts the method's own doc comment ("On a freshly loaded aggregate ... those keys are exactly *this battle's* mutated statistics"), which frames it as per-battle — a future refactor that resets the dirty set per call would silently break offline challenge completion.

The existing offline tests don't guard against that regression: they're all-win, all-loss, or boss-win, where every battle touches the same statistics, so the final call's set equals the full set regardless of accumulation. The case where accumulation actually matters — a *mixed-outcome* window where a tracked statistic is touched only in a non-final battle (e.g. a `BattlesWon`/`EnemiesKilled` challenge crossed by early wins, with the final battle a loss) — is untested. Consider adding a mixed-outcome integration test that crosses a win-tracking challenge and ends on a loss, or at minimum a note on `RecordBattleCompleted` that the offline path relies on the accumulating dirty set.

---

Game.Application/Services/BattleService.cs
```diff
@@ -243,6 +261,187 @@ public async Task<bool> EndBattleLoss(Player player, PlayerState state, Cancella
...
+            var parameters = BuildSimulationParameters(player, mode, zone, awayMs);
+
+            var result = _offlineSimulator.Simulate(parameters, cancellationToken);
+
+            var levelBefore = player.Level;
+            var statPointsBefore = player.StatPoints.StatPointsGained;
+            var completedChallenges = await ApplyOfflineRewards(player, result, cancellationToken);
+
+            // Re-anchor and persist the player (exp/levels/unlocks) once. The single core-updated event raised
+            // here carries the final absolute state, so it persists the batched exp and levels too.
```
ginderjeremiah: **Nit.** This says "*The single* core-updated event raised here", but the window actually raises two `PlayerCoreUpdatedEvent`s — one from `GrantOfflineExp` in `ApplyOfflineRewards`, and this one from `StampActivity`. Harmless (write-behind keeps the final absolute state, and two is nowhere near the per-kill flood decision 6 avoids), but the wording reads as if exp/levels ride on *this* event when they were already raised earlier. Worth tweaking to "a single core update for the whole exp batch (here, the re-anchor) ..." or similar.

---

ginderjeremiah (commented):
## Review: Offline rewards orchestration, reward application, challenge consolidation & `GetOfflineProgress` socket command (#1042)

**Verdict:** Approve *(submitted as a `COMMENT` because GitHub won't accept a gating review from the PR author — treat this as an approve)* — the orchestration faithfully implements spike #879 decisions 2/6/7, layering is clean, and coverage is strong. One should-fix on a test gap and a doc-comment nit, neither blocking.

This wires the #1041 simulator into the live app exactly as the issue scopes it: server-side away time off `LastActivity`, stale in-flight battle settled first, persisted loop-mode/zone resolution with idle fallback, capped simulation, consolidated stat/challenge application with the per-challenge push suppressed, single re-anchor + persist. The judgement calls (opt-in stalemate cutoff scoped to opening pure-draw batches; ≤2-min in-flight battle counted in the away wall-clock but credited separately, deferring precise catch-up to #1073) are reasonable and well-documented.

### Architecture & layering
- DDD boundaries hold: the pure simulator stays in `Game.Core` (no persistence/app deps), exp-batching and notify-suppression live on the `Player` aggregate, and the new `ChallengeRewardService` factors the evaluate-resolve-complete step so the live `BattleStatisticsEventHandler` and the offline batch share it (DRY) rather than duplicating — directly satisfying decision 7. The `GrantExp`→`ApplyExp` extraction keeps the clamp + level-up arithmetic from drifting between the live and batched paths.
- DI lifetimes are right: `OfflineProgressSimulator` singleton (stateless), `ChallengeRewardService`/`BattleService` scoped.
- The socket command is a thin seam — resolves player/state from the session, delegates, persists state. No cross-player access surface.

### Should-fix
- **Untested reliance on cumulative dirty statistics** — `BattleService.cs:419-421`. Correct in production, but the only guard against a future per-call-reset of `_dirtyStatistics` is a mixed-outcome test that doesn't exist yet (all current offline tests touch identical stats every battle). See inline.

### Nits
- **"single core-updated event" comment** is slightly inaccurate — the window raises two (`GrantOfflineExp` + `StampActivity`). `BattleService.cs:320-321`, inline.

### Test coverage
Strong and aligned with the guidelines. Unit: `GrantOfflineExp` (batched grant, one core update, per-grant-not-sum clamp, one level-up event/level), `CompleteChallenge(notify:false)` (unlocks but no completed event), and the stalemate cutoff (fires on all-draw opening, not on wins/losses). Integration: multi-battle away applying exp/levels/stat points + re-anchor, sub-threshold no-op, second-immediate no-op, mid-window challenge completing once + unlocking, stale in-flight resolution, boss-mode loop. No frontend tests needed — the change is backend + regenerated codegen types only; the welcome-back gate is #1043. The one gap is the mixed-outcome challenge case noted above.

### What's good
- The stationarity-driven "materialize player modifiers once" and the cancellation-token threading are exactly the right calls for an inline-within-timeout simulation.
- Idle/boss resolution degrades gracefully (retired/locked/bossless boss target → idle in the nearest viable zone, mirroring the live loop) and the summary reports what *actually* ran (`result.IsBossBattle`/`ZoneId`), not what was requested.
- Thorough, accurate doc comments throughout, plus the `game-design.md` / `backend-battle.md` updates the spike called for.

---

PR #1077, opened 2026-06-21
Make per-account login backoff increment race-safe (#1054)

### LINKED ISSUE #1054: Make per-account login backoff increment atomic (race-safe)
## Context

Follow-up from #1010 / PR #1053, which added the per-account exponential login backoff. The backoff counter is maintained by a read-modify-write in `LoginBackoffGuard.RegisterFailure`:

```csharp
var current = await _store.Get(username, cancellationToken);
var (next, retention) = _policy.RegisterFailure(current, _timeProvider.GetUtcNow());
await _store.Set(username, next, retention, cancellationToken);
```

This GET → compute → SET sequence is **not atomic**. Two concurrent failed attempts for the same account can both read the same `FailureCount`, both compute the same next state, and the later SET overwrites the earlier one — so concurrent failures can lose increments and the streak climbs slower than the attempt rate.

## Why it matters

The threat model this feature targets is a **slow, distributed (botnet) guess** against one account — i.e. exactly the case where multiple failed attempts can land concurrently. The lost-update race partially undercuts the control's core purpose, letting a sufficiently concurrent attacker keep the effective count (and thus the backoff) lower than it should be. It's defence-in-depth layered on the per-IP limiter, so this is a robustness gap rather than an open hole, which is why it's a follow-up rather than a blocker.

## Proposed work

- Make the increment atomic, e.g. a Redis `INCR` for the failure count (with the lock-until/expiry derived from it) or a small Lua script that does the read-modify-write server-side.
- Add an integration test that fires concurrent failed attempts and asserts the count reflects all of them.
- Keep the pure `LoginBackoffPolicy` arithmetic unit-tested as-is; only the guard/store interaction changes.

Related: #1010

ginderjeremiah:
## Summary

Follow-up from #1010 / PR #1053. `LoginBackoffGuard.RegisterFailure` recorded a failure with a plain `GET → compute → SET`:

```csharp
var current = await _store.Get(username, ct);
var (next, retention) = _policy.RegisterFailure(current, now);
await _store.Set(username, next, retention, ct);
```

That sequence is **not atomic**: two concurrent failed attempts for the same account can both read the same `FailureCount`, both compute the same next state, and the later `SET` overwrites the earlier — losing increments so the streak climbs slower than the attempt rate. Since this control's threat model is a **slow, distributed (concurrent) guess** against one account, the lost-update race partially undercuts its purpose.

## How it addresses the issue

Makes the increment atomic via **optimistic concurrency**, keeping the pure `LoginBackoffPolicy` arithmetic — and its unit tests — **unchanged**; only the guard/store interaction changes (as the issue scopes it):

- **`ICacheService.CompareAndSet`** — a server-side Lua compare-and-set that writes value + TTL only if the stored value is still unchanged (or, with a `null` expected, only if the key is absent). The conditional-write counterpart to the existing `CompareAndDelete`, and atomic for the same reason.
- **`ILoginBackoffStore.Set` → `TryUpdate`** — an atomic compare-and-set against the prior state, returning whether the swap won.
- **`LoginBackoffGuard.RegisterFailure`** — re-reads, recomputes, and retries whenever a concurrent failure wins the swap. Each iteration observes the now-higher count, so the count reflects **every** failure. The loop is lock-free (some writer commits each round), so it makes progress under contention rather than spinning.

The compare-and-set matches the exact stored payload by re-serialising the expected state with the same (deterministic) serializer that wrote it — the store only ever writes via that serializer, so `Serialize(Deserialize(stored)) == stored` is guaranteed for this record.

## Test plan

- **New integration tests (`Game.Application.Tests/Auth/LoginBackoffGuardIntegrationTests.cs`)** against real Redis:
  - 50 concurrent `RegisterFailure` calls for one account → stored `FailureCount == 50` (the race the fix closes; a plain read-modify-write would fall short).
  - Sequential increments exercise both compare-and-set branches (absent-key first failure, then unchanged-value swaps).
- `LoginBackoffPolicyTests` unchanged — the arithmetic is untouched.
- Two `ICacheService` test fakes get the new (throwing) member.
- Full suites green locally: **Game.Application.Tests** 427, **Game.Api.Tests** 555, **Game.Infrastructure.Tests** 57; `dotnet build` clean (0 warnings).

## Docs

- `docs/backend.md` — the per-account-backoff note now describes the race-safe compare-and-set retry (relevant because the threat model is concurrent).

## Notes / judgement calls

- **Optimistic CAS over an atomic `INCR`.** The issue suggested either `INCR` or a Lua read-modify-write. A bare `INCR` only makes the *count* atomic and would force the derived `LockedUntil`/TTL out into a second racy write (or duplicate the exponential delay arithmetic into Lua, which would orphan the unit-tested policy). The CAS-retry keeps the policy the single, untouched home of the arithmetic and makes the *whole* state update atomic — the tightest fit for "keep the policy arithmetic as-is; only the guard/store interaction changes." Happy to switch to `INCR` if you prefer.

Closes #1054

---
_Generated by [Claude Code](https://claude.ai/code/session_014fXPfoWLMniUqnG4F8PneE)_


Game.Application/Auth/LoginBackoffGuard.cs
```diff
@@ -37,9 +38,21 @@ public class LoginBackoffGuard(
...
+            // it makes progress under contention rather than spinning indefinitely.
+            while (true)
+            {
+                var current = await _store.Get(username, cancellationToken);
+                var (next, retention) = _policy.RegisterFailure(current, _timeProvider.GetUtcNow());
+                if (await _store.TryUpdate(username, current, next, retention, cancellationToken))
+                {
+                    return;
+                }
+            }
```
ginderjeremiah, line 55: Nit (non-blocking): correctness here rests entirely on the invariant that `Serialize(Deserialize(stored)) == stored` for `LoginBackoffState`. That invariant does hold today — STJ round-trips `int` + `DateTimeOffset` idempotently and the camelCase/record-declaration ordering is deterministic — and `TryUpdate`'s comment captures the assumption well.

The thing to be aware of: if that invariant is ever broken (a future field with a culture-sensitive or non-idempotent converter, a custom `JsonConverter`, or another writer touching this key via a different serializer), `TryUpdate` would return `false` on every iteration and this `while (true)` becomes an unbounded hot loop hammering Redis on the login path — i.e. a self-inflicted spin in the exact control meant to resist abuse.

A hard retry cap isn't obviously the right fix (on the cap you'd have to choose between dropping the increment, which regresses this PR, or throwing, which turns a login failure into a 500), so I'd lean toward leaving the loop as-is rather than bounding it. Flagging mainly so the serialization-determinism coupling is a conscious, documented constraint on `LoginBackoffState` going forward. Take it or leave it.

---

ginderjeremiah (commented):
## Review: race-safe per-account login backoff increment (closes #1054)

**Verdict:** Approve *(submitted as a comment — I'm the PR author, so GitHub won't accept a gating review from me)*. This is a clean, correct fix that matches the issue's scope. **Not merging:** CI checks haven't reported yet (status is pending) and there's one optional nit below.

The change replaces the non-atomic `GET → compute → SET` in `RegisterFailure` with an optimistic compare-and-set retry, keeping the pure `LoginBackoffPolicy` arithmetic (and its unit tests) untouched. That's exactly what #1054 asked for, and the CAS-over-`INCR` judgement call is well reasoned — `INCR` would only make the count atomic and force the derived `LockedUntil`/TTL into a second racy write or duplicate the exponential arithmetic into Lua, orphaning the unit-tested policy. CAS keeps the whole state update atomic with the policy as the single home of the math.

### Correctness (verified)
- The Lua branches are sound: the null-expected branch asserts key absence (`exists == 1 → 0`), the value branch swaps only on an unchanged match, and TTL is written alongside the value via `PX` so the key never lingers without one. A key that expires between `Get` and `TryUpdate` falls through to a retry that re-reads `null` and takes the absent-key branch — handled correctly.
- The `PX` expiry can't be ≤ 0: retention is `Max(FailureWindowSeconds, delay)`, and `FailureWindowSeconds` is validated sane on startup.
- The retry loop genuinely makes progress under legitimate contention (some writer commits each round), and the old `ILoginBackoffStore.Set` is fully removed — no dead code or stale callers left behind.
- DDD layering holds: `CompareAndSet` in Infrastructure, `TryUpdate` in the DataAccess store, orchestration in the Application guard, arithmetic in the pure policy. No domain logic leaked.

### Test coverage
Adequate and follows the project's rules. The 50-concurrent integration test is deterministic in its assertion (`count == 50` only passes if every increment lands), which is the right way to pin the race. The sequential test deterministically exercises both CAS branches (absent-key first failure, then unchanged-value swaps). Omitting a dedicated `RedisService.CompareAndSet` integration test is consistent with the existing pattern — `CompareAndDelete` has no direct test either and is likewise covered through its caller.

### Nits
- One non-blocking observation on the `while (true)` loop's coupling to serialization determinism — see the inline comment. No change required.

### What's good
Tight, well-documented change; the comments explain the *why* (the lost-update race and the deterministic round-trip the CAS depends on) without over-explaining, and `docs/backend.md` was updated to reflect the concurrent threat model. Nicely scoped follow-up to #1053.

---

Game.Application/Auth/LoginBackoffGuard.cs
ginderjeremiah: Took it — documented the constraint on the `LoginBackoffState` record itself (d4aabde) so a future field-adder sees it at the type that bears it, rather than only in `TryUpdate`'s comment. Agreed on leaving the loop unbounded: a cap would force a choice between dropping the increment (regressing this fix) or a 500 on login, both worse than the spin it'd guard against — and the spin is now a documented, avoidable invariant rather than a silent trap.

---

PR #1078, opened 2026-06-21
Gate persisted boss mode on actual boss-loop engagement (#1067)

### LINKED ISSUE #1067: Refine frontend auto-challenge-boss mode sync (pre-armed-idle semantic + dedup)
Follow-up from #1040 (PR #1066). Both points are harmless today — no consumer reads `Player.AutoChallengeBoss` yet — but should be resolved before the offline-rewards simulation (#1041/#1042) loops the persisted mode for the whole away budget.

## 1. Pre-armed-while-idle semantic (should-fix)

`EnemyManager.setAutoFight(true)` persists boss mode (`AutoChallengeBoss = true`, with the boss zone implicitly being `CurrentZoneId`), mirroring the player's auto-fight *intent* per #1040's stated scope ("mirror the live `autoFight` state"). But `BossTrigger.svelte` renders the `AutoFightToggle` in the **not-engaged** state, so a player can pre-arm auto-fight *before* pressing Challenge — at which point the persisted mode flips to boss even though the live loop is still **idle-farming**.

Once the offline sim reads this field, a player who pre-armed auto-fight and then walked away *without ever challenging* would be resumed as a boss-farmer instead of an idle-farmer.

**Decide:** either
- gate boss-enable on actually being in the boss loop — sync `enabled = autoFight && mode === 'boss'`, which also requires syncing on `challengeBoss()` success so a pre-armed player who *does* challenge still persists boss; or
- affirm that "auto-fight armed" is the intended persisted semantic and document it.

Related: since the boss is always the current zone's boss (a `bool`, not a stored zone), consider whether changing the idle zone (`ChangeZone` / lazy relocation) while `AutoChallengeBoss` is `true` should reset it to idle — otherwise the flag would silently retarget the new (possibly bossless) zone. This only matters in combination with the pre-arm case above.

## 2. Dedup redundant syncs (nit)

Every `returnToIdle()` path (retreat, boss loss, boss draw, single-victory handoff, failed-challenge fallback) fires `SetAutoChallengeBoss(false)` unconditionally — even when the persisted mode is already idle, and on a boss loss it duplicates the backend's own `BattleLost` backstop. It's fire-and-forget and bounded to boss transitions, so not a real problem, but tracking the last-synced state and only sending on an actual change would cut the redundant socket traffic. (Note the stale-initial-state subtlety: the live `autoFight` starts `false` at session start even if the persisted mode is boss — see point 3.)

## 3. Tie-in

Login re-entry reconciliation (a returning boss-farmer's live `autoFight` resets to idle while the persisted mode stays boss until the next sync) is already deferred to the welcome-back gate (#1043), which "includes the frontend half of #1040's mode-sync." Whatever is decided here should be consistent with that.

ginderjeremiah:
## Summary

Follow-up from #1040 (PR #1066). Resolves the two points raised in #1067 about the frontend auto-challenge-boss mode sync. Both were harmless today (no consumer reads `Player.AutoChallengeBoss` yet) but needed settling before the offline-rewards sim (#1042) loops the persisted mode for the whole away budget.

### 1. Pre-armed-while-idle semantic (the should-fix) — **Option A**

`AutoFightToggle` is rendered on the **not-engaged** `BossTrigger`, so a player can pre-arm auto-fight *before* pressing Challenge. The previous `setAutoFight` persisted boss mode on that intent alone, so a player who pre-armed and then walked away **without ever challenging** would be resumed by the offline sim as a boss-farmer instead of an idle-farmer.

Per #1040's stated goal — persist *what the loop is actually doing* — I took **option A**: gate the boss-persist on actual engagement rather than intent.

- `setAutoFight(on)` now syncs `on && mode === 'boss'`, so pre-arming while idle-farming persists **idle**, not boss.
- `challengeBoss()` re-syncs on a successful engage, so a pre-armed player who *does* challenge still persists boss (and a one-off, auto-fight-off challenge stays idle-persisted — a single victory hands straight back to the idle loop).

**On the related zone-retarget question** (changing the idle zone while `AutoChallengeBoss` is true silently retargeting the new zone's boss): the issue notes this *"only matters in combination with the pre-arm case above."* Option A eliminates that combination — boss is never persisted while idle-farming — so no separate zone-reset is required. (A deliberate mid-fight zone navigation during an active auto-fight boss farm retargets the boss the player is *actively* fighting, which matches intent and is pre-existing; the backend's `CurrentZoneId` isn't updated mid-engagement anyway.)

### 2. Dedup redundant syncs (the nit)

Every `returnToIdle()` path (retreat, boss loss/draw, single-victory handoff, failed-challenge fallback) fired `SetAutoChallengeBoss(false)` unconditionally — even when the persisted mode was already idle. `syncAutoChallengeBoss` now tracks the last value actually sent and drops a no-op, cutting the redundant socket traffic. The tracker starts `undefined` so the first authoritative state always sends.

### 3. Tie-in (#1043)

Point 3 (a returning boss-farmer's live `autoFight` resets to idle while the persisted mode stays boss until the next sync) is **left to the welcome-back gate (#1043)**, which the spike already says "includes the frontend half of #1040's mode-sync." The dedup's stale-initial subtlety is documented in code pointing at #1043 rather than reconciled here, to avoid overlapping that work.

## How it addresses the issue

`UI/src/lib/engine/battle/enemy-manager.ts`:
- `setAutoFight` → `syncAutoChallengeBoss(on && this.mode === 'boss')`.
- `challengeBoss` success branch → `syncAutoChallengeBoss(this.autoFight)` (mode is `'boss'` there).
- `syncAutoChallengeBoss` → deduped against `lastSyncedAutoChallengeBoss`.

No backend change: the `SetAutoChallengeBoss` command already just stores the bool, and the anti-cheat validation (current zone has a boss) is satisfied whenever boss is now persisted.

## Test plan

`UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts` — reworked/added to cover option A + dedup:
- pre-arm while idle does **not** persist boss;
- toggling auto-fight on while engaged persists boss;
- pre-arm **then** challenge persists boss;
- retreat/loss/draw from a boss-farm persist idle;
- a redundant return-to-idle (one-off challenge) does **not** re-emit an idle command.

- `npm run lint` clean (eslint + svelte-check, 0 errors).
- Full frontend unit suite green: **2282 passed** (215 files).

Frontend-only; no battle-step logic touched, so backend parity is unaffected.

Closes #1067

---
_Generated by [Claude Code](https://claude.ai/code/session_01X3HypTpYeAr5cWJ46wyU2V)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -72,6 +72,10 @@ export class EnemyManager {
 	/** Resolver that short-circuits an in-flight retry backoff so a stop / superseding transition need
 	 *  not wait out the full delay before the loop re-checks whether it should still be running. */
 	private cancelBackoff?: () => void;
+	/** The last `AutoChallengeBoss` value synced to the backend, so {@link syncAutoChallengeBoss} drops
+	 *  redundant re-syncs (e.g. a returnToIdle when the persisted mode is already idle). `undefined` until
+	 *  the first sync, so the first authoritative state is always sent. */
+	private lastSyncedAutoChallengeBoss?: boolean;
 
 	public start() {
 		if (!this.started) {
@@ -248,6 +252,10 @@ export class EnemyManager {
 			if (result.data?.enemyInstance) {
 				this.currentEnemy = result.data.enemyInstance;
 				notifyNewEnemyLoaded(this.currentEnemy);
+				// Now genuinely in the boss loop, so persist boss mode iff auto-fight is armed — a pre-armed
+				// toggle (set while still idle-farming) only becomes real boss-farming here. A one-off challenge
+				// (auto-fight off) stays idle-persisted: a single victory hands straight back to the idle loop.
+				this.syncAutoChallengeBoss(this.autoFight);
 			} else {
 				if (result.error) {
 					logMessage(ELogType.Debug, 'There was an error challenging the boss: ' + result.error);
@@ -288,11 +296,11 @@ export class EnemyManager {
 	/** Toggle auto-fight: when on, a boss victory immediately re-challenges the boss. */
 	public setAutoFight(on: boolean) {
 		this.autoFight = on;
-		// Persist the player's auto-fight intent (#1040's "mirror the live autoFight state" semantic). Note
-		// auto-fight can be pre-armed from BossTrigger while the loop is still idle, so this persists boss
-		// mode on intent rather than active engagement; whether to gate on mode==='boss' is tracked in #1067
-		// (must be settled before the offline-sim consumer in #1041/#1042 reads the field).
-		this.syncAutoChallengeBoss(on);
+		// Persist boss mode only when the loop is *actually* boss-farming: auto-fight on AND already engaged in
+		// the boss loop. Auto-fight can be pre-armed from BossTrigger while the loop is still idle-farming —
+		// that is intent, not engagement, and must not persist boss (the offline sim would otherwise resume a
+		// never-challenged player as a boss-farmer). challengeBoss() re-syncs once a pre-armed player engages.
+		this.syncAutoChallengeBoss(on && this.mode === 'boss');
 	}
 
 	private returnToIdle(sync = true) {
@@ -314,8 +322,17 @@ export class EnemyManager {
 	 * player's `CurrentZoneId`. Fire-and-forget — anti-cheat validation is the server's, and a transient
 	 * failure only leaves the persisted mode briefly stale, corrected by the next sync or the backend's
 	 * boss-loss/draw backstop.
+	 *
+	 * Deduped against the last value actually sent so redundant transitions — e.g. every returnToIdle path
+	 * firing when the persisted mode is already idle — don't emit a no-op socket command. The live `autoFight`
+	 * starts false each session even when the persisted mode is boss; reconciling that stale-initial state on
+	 * login is the welcome-back gate's job (#1043).
 	 */
 	private syncAutoChallengeBoss(enabled: boolean) {
+		if (enabled === this.lastSyncedAutoChallengeBoss) {
+			return;
+		}
+		this.lastSyncedAutoChallengeBoss = enabled;
 		void apiSocket.sendSocketCommand('SetAutoChallengeBoss', enabled);
 	}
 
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Gate persisted boss mode on actual boss-loop engagement (#1067)

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't accept a formal approve on my own PR)*. The change implements #1067's Option A correctly and is well-tested. One small test-coverage gap noted below, not a blocker.

I traced the three behavioral changes against the full state machine in `enemy-manager.ts` and walked every existing boss-sync test through the new dedup tracker — all remain consistent.

### What's good
- **Faithful to Option A.** `setAutoFight(on)` → `syncAutoChallengeBoss(on && this.mode === 'boss')` correctly treats a `BossTrigger` pre-arm (mode still `idle`) as intent rather than engagement, and `challengeBoss()`'s success branch re-syncs `this.autoFight` *after* the supersession guard (line 249), so a superseded challenge never persists boss. The boss-victory auto-refarm loop also benefits — each re-challenge is now deduped instead of (previously) not syncing at all; net behavior is unchanged and quieter.
- **Dedup is sound.** `undefined`-initial means the first authoritative state always sends; `stop()` correctly passes `returnToIdle(false)` and leaves the tracker untouched so a disconnecting boss-farmer's persisted mode isn't clobbered.
- **Scope discipline.** The zone-retarget sub-question and the stale-`autoFight`-on-login reconciliation are both correctly left to #1043, with code comments pointing there rather than half-solving it. Logic-only, no battle-step simulation touched, so backend parity is genuinely unaffected.

### Nits
- **Missing test for toggling auto-fight *off* while engaged in the boss loop.** Under Option A, `setAutoFight(false)` during an active boss fight (mode `boss`, previously synced `true`) now persists **idle** — a meaningful new behavior (the one-off-mid-farm handoff) that isn't directly covered. The existing "persists idle mode ... when auto-fight is toggled off" test fires from a fresh idle state, where `false && mode==='boss'` is `false` regardless of the mode gate, so it wouldn't catch a regression in the gate itself. Worth one test: `challengeBoss()` → `setAutoFight(true)` → `setAutoFight(false)` asserts `SetAutoChallengeBoss` `false` is sent. Optional — take it or leave it.

### Test coverage
Good. New/reworked tests cover pre-arm-while-idle (no boss persist), toggle-on-while-engaged, pre-arm-then-challenge, retreat/loss/draw persisting idle, and the dedup no-op on a one-off return-to-idle. Test file mirrors the source path per the frontend testing rules. Only the toggle-off-while-engaged path (above) is uncovered.

CI is still in progress (frontend unit + e2e); `verify-frontend` is green. Not merging here since I'm the author and there's an open nit — leaving it to you to optionally fold in the extra test before merge.

---

Issue #1070, opened 2026-06-21
Multiple players per account: player-select & create-character screen (frontend)


ginderjeremiah:
Sub-issue of #922. See the spike design doc: `docs/spikes/922-multiple-players-per-account.md`. Depends on #1068 (auth seam) and #1069 (character creation).

## Goal

A character-select screen in the login flow.

## What to build

- New screen between the login page and `/loading`: after credentials verify, show the account's `playerSummaries`; selecting one calls `SelectPlayer` and proceeds to the loading screen with the returned player.
- "Create character" affordance on the same screen (name input → `CreatePlayer` → refresh list), respecting the per-account cap.
- Move the session-takeover (`ActiveSession`) confirmation to fire **after** selection.
- Re-initialize `playerManager` from the selected player's data.

## Docs

Update `docs/frontend-screens.md` with the new screen.

Issue #1069, opened 2026-06-21
Multiple players per account: character creation & per-account cap


ginderjeremiah:
Sub-issue of #922. See the spike design doc: `docs/spikes/922-multiple-players-per-account.md`. Depends on #1068 (player-selection seam — shared player DTOs / select flow).

## Goal

Let an account create additional characters beyond the one made at signup.

## What to build

- `CreatePlayer(name)` authenticated endpoint/command that builds a new player via the existing `NewPlayerFactory` and attaches it to the account.
- A **per-account character cap**, config-bound (proposed default: 6), enforced server-side as anti-cheat.
- **Per-player names**: today a player's name is the account username (`NewPlayerFactory.Create(username)`). Each character now needs its own user-supplied name. Names need not be globally unique (no competitive surface in the game); validate length/charset consistent with the existing 20-char `Player.Name` limit.

## Out of scope

Deleting/retiring a character — deferred to a later issue.

## Tests

Unit-test cap enforcement and name validation; integration-test creation + attach to the account.

Issue #922, opened 2026-06-18
Add functionality to allow tying multiple players to an account.


ginderjeremiah:
Need a Player select screen somewhere in the login flow. Also need to decide if the game sessions should continue to be restricted to one per User or if they should be restricted to one per player.

---

**Spike resolved** — see the design doc: `docs/spikes/922-multiple-players-per-account.md`.

**Decision:** one active session per account (one character live at a time), with **lossless in-app switching** via the offline-rewards simulator (#879). The selected player is carried as a **token claim**. Genuine per-player parallel progression (server-side idle) is deferred — client-driven idle loops can't progress in throttled background tabs, so true parallelism would require the server to become the authoritative idle driver.

### Implementation issues (dependency order)
- [ ] #1068 — Player-selection seam & auth flow (backend)
- [ ] #1069 — Character creation & per-account cap (backend)
- [ ] #1070 — Player-select & create-character screen (frontend)
- [ ] #1071 — Lossless character switch (backend) — also depends on #1042
- [ ] #1072 — In-app character switcher (frontend)

### Follow-ups (non-blocking)
- [ ] #1073 — Spike: recover idle time lost to background-tab throttling (mid-session catch-up)
- [ ] #1074 — Notify users to allowlist the game as always-active

PR #1084, opened 2026-06-21
Split login around player selection with a selected-player token claim (#1068)

### LINKED ISSUE #1068: Multiple players per account: player-selection seam & auth flow
Sub-issue of #922 (multiple players per account spike). See the spike design doc: `docs/spikes/922-multiple-players-per-account.md`.

## Goal

Turn the existing single-player login binding into an explicit player-selection step so an account with multiple players can choose which one to enter as. The DB is already one-to-many (`Player.UserId` is a non-unique FK) and `AccountService.SelectPlayerId` is the documented seam for this.

## What to build

- Split `Login` so it verifies credentials and returns `{ tokens, playerSummaries }` (id, name, level, current zone) instead of binding + loading the first player. No session binding yet.
- New authenticated `SelectPlayer(playerId)` endpoint/command that:
  - validates the player belongs to the authenticated account (anti-cheat),
  - establishes the session binding (`PlayerState`),
  - **rotates the token pair to carry a `playerId` claim** — the selected-player anchor (spike decision).
- Rework `SelectPlayerId` / `ResolveSelectedPlayerId` to read the selected player from the **token claim**, not default to `playerIds[0]`. This fixes the rehydration bug where an evicted session would silently snap back to the first character.
- Add an `Id` to the player DTOs (`PlayerData`, plus a lean `PlayerSummary` for the list).
- Move the `Login/ActiveSession` session-takeover check to run **after** selection (it is a per-player presence check).

## Notes

- Keeps the one-active-session-per-account model (Option A++ from the spike): one character live at a time.
- The socket handshake reads its player from the token claim, making the binding explicit and stateless (consistent with "the token is the sole authority").

## Tests

Unit-test the ownership validation and the claim-based resolution; integration-test the login → select → socket-handshake path.

ginderjeremiah:
## Summary

The foundation for **multiple characters per account** (spike #922). Turns the single-step login into an explicit **player-selection** flow anchored on a token claim:

- **`Login`** verifies credentials and returns `{ tokens, playerSummaries }` (id, name, level, current zone) — it **binds no player**.
- **`SelectPlayer(playerId, refreshToken)`** (new, authenticated) validates the player belongs to the account (anti-cheat), loads it, binds the session, and **rotates the token pair** to carry the chosen player as a `player_id` claim.
- **Session rehydration & the socket handshake** now resolve the bound player from that claim — **authoritative over a mismatched cache** — instead of defaulting to the account's first character. This fixes the real correctness bug the spike calls out: an evicted session silently snapping a multi-player account back to its first character.

## How it addresses the issue (#1068 checklist)

- ✅ Split `Login` → `{ tokens, playerSummaries }`, no bind.
- ✅ New `SelectPlayer` (ownership-validated, binds session, rotates the token with a `player_id` claim).
- ✅ Resolve the selected player from the **claim**, not `playerIds[0]` (removed the now-dead `ResolveSelectedPlayerId` / `SelectPlayerId` seam and `AccountCredentials.PlayerIds`).
- ✅ Added `Id` to `PlayerData`; extended the lean `PlayerSummary` contract with `CurrentZoneId`.
- ✅ Moved the `Login/ActiveSession` takeover check to run **after** selection (per-player presence check).

## Key design decisions (worth a look)

- **Token-claim mechanics.** The `player_id` claim rides the access token **and** the `RefreshTokenData` record, so the selection survives a refresh. `SelectPlayer` **consumes + rotates** the caller's refresh token (upholding the single-use rotation invariant; roles carried from the consumed token) and consumes it **only after** the ownership/load checks pass, so a rejected selection leaves the caller's token intact.
- **Login no longer rejects an account with no characters** — it returns an empty summary list (character creation is #1069). `LoginStatus.NoPlayer` / `PlayerDataNotFound` moved to a new `SelectPlayerStatus`.
- **`PlayerSummary` reuse.** Extended the existing Identity contract (shared User aggregate) rather than adding a near-duplicate — DRY, and it already carried id/name/level.

## ⚠️ Frontend scope note (stopgap)

The real **player-select screen is #1070**. To keep the app working and CI green after the `LoginResult` shape change, this PR adds a **minimal stopgap**: after login the client **auto-selects the first character** via `SelectPlayer`, then proceeds exactly as before. #1070 replaces the auto-select with the actual screen. Documented inline in `+page.svelte`.

## Test plan

- **Backend (all green):** Api **560**, Application **427**, Core **685**, Infrastructure **57**.
  - New: `SelectPlayer` ownership/rotation/invalid-token (unit + controller), `player_id` claim emission/absence, `Login` returns summaries (single + multi-player + empty), token-claim rehydration.
  - Test helpers now mint **post-selection** tokens; `LoginAndBuildClientAsync` goes through `SelectPlayer`.
- **Frontend (all green):** **2280** unit tests, `npm run lint` (eslint + svelte-check) clean, coverage floors pass. Updated the login-submit flow tests for Login → SelectPlayer.
- `dotnet format --verify-no-changes` clean; TS client regenerated (no codegen drift).

## Docs

- `docs/backend.md` — Authentication (selected-player token claim, claim-authoritative rehydration) and Account/login orchestration (split login + `SelectPlayer`).

Closes #1068

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bvq6R3PTh6hdRMk1uo1Vs5)_


Game.Application.Tests/Services/AccountServiceIntegrationTests.cs
```diff
@@ -328,32 +351,70 @@ public async Task Login_BannedUserWithWrongPassword_ReturnsInvalidCredentials()
...
-            var user = await TestDataSeeder.CreateUserAsync(context, "rehydrateuser", "pass");
+            var user = await TestDataSeeder.CreateUserAsync(context, "selectuser", "pass");
+            var skill = await TestDataSeeder.CreateSkillAsync(context);
             var player = await TestDataSeeder.CreatePlayerAsync(context, user.Id);
+            await TestDataSeeder.LinkSkillToPlayerAsync(context, player.Id, skill.Id);
+            await ReloadReferenceCachesAsync();
 
             var accountService = CreateAccountService(scope.ServiceProvider);
+            var login = await accountService.Login("selectuser", "pass");
+            Assert.True(login.Success);
```
ginderjeremiah, line 367: **Blocking — breaks the backend build (CI red).** `login.Tokens` is `AuthTokenPair?`, and nothing here asserts `login.Success` first, so the compiler can't narrow it via `[MemberNotNullWhen(true, nameof(Tokens), …)]`. Under the CI's `dotnet build -warnaserror` this is `CS8602: Dereference of a possibly null reference` and fails `test-backend`/`verify-backend`. Add `Assert.True(login.Success);` (or otherwise guard) before reading `login.Tokens.RefreshToken`. The same fix is needed at line 394.

Note this contradicts the PR description's "Backend (all green)" claim — please re-run the suite locally with `-warnaserror` before re-requesting review.

---

Game.Application.Tests/Services/AccountServiceIntegrationTests.cs
```diff
@@ -328,32 +351,70 @@ public async Task Login_BannedUserWithWrongPassword_ReturnsInvalidCredentials()
...
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+            var owner = await TestDataSeeder.CreateUserAsync(context, "owner", "pass");
+            var attacker = await TestDataSeeder.CreateUserAsync(context, "attacker", "pass");
+            var ownerPlayer = await TestDataSeeder.CreatePlayerAsync(context, owner.Id);
+
+            var accountService = CreateAccountService(scope.ServiceProvider);
+            var login = await accountService.Login("attacker", "pass");
+            Assert.True(login.Success);
+
```
ginderjeremiah, line 394: **Blocking — same `CS8602` as line 367.** `login.Tokens.RefreshToken` dereferences a possibly-null `AuthTokenPair?` without an `Assert.True(login.Success)` to narrow it, so `-warnaserror` fails the build here too.

---

ginderjeremiah (commented):
## Review: Split login around player selection with a selected-player token claim (#1068)

**Verdict:** Request changes — the backend build is failing on CI (`test-backend` + `verify-backend` red), so this isn't mergeable as-is. _(Submitted as a comment because I'm the PR author; the intent is "request changes".)_

The design itself is solid and faithfully implements the spike (#922) and issue (#1068): login splits into verify → `{ tokens, playerSummaries }` with no binding; `SelectPlayer` validates ownership, consumes-after-validate to keep a rejected token intact, and rotates the pair with a `player_id` claim; rehydration is now claim-authoritative (the genuine fix for the evicted-session-snaps-to-first-character bug). Layering is respected (orchestration in `AccountService`, the controller stays a thin adapter wiring the request-scoped session), the docs are updated well, and the dead `ResolveSelectedPlayerId`/`SelectPlayerId`/`AccountCredentials.PlayerIds` seam is fully removed. The only thing standing between this and an approve is the build break.

### Blocking

- **Backend build fails under `-warnaserror`** — `Game.Application.Tests/Services/AccountServiceIntegrationTests.cs:367` and `:394` dereference `login.Tokens` (an `AuthTokenPair?`) without first asserting `login.Success`, so `[MemberNotNullWhen]` can't narrow it → `CS8602`, which CI promotes to an error. This is what's reding `test-backend`/`verify-backend`. Fix: `Assert.True(login.Success);` before reading `login.Tokens` in both tests (inline comments mark the spots). Please also re-validate the PR's "Backend (all green)" / `dotnet format` claims — they don't match the current CI run.

### Nits / follow-ups (non-blocking)

- `Users.GetPlayerSummaries` duplicates the `PlayerSummary` projection in `ToAdminUser`. The mirror comments are a fine stopgap, but the two will drift; a shared projection expression (or a follow-up to extract one) would be cleaner.
- No test covers the `SelectPlayerStatus.PlayerDataNotFound` branch (ownership passes but `GetPlayer` returns null → 404). It's an edge/race, but worth a unit test for completeness.

### Test coverage

Strong overall — `SelectPlayer` ownership/rotation/invalid-token (unit + controller), `player_id` claim emission/absence, multi-player and empty-account login, and claim-based rehydration are all exercised, and the helpers were updated to mint post-selection tokens and route through `SelectPlayer`. The frontend login→select pipeline (happy path, takeover-after-select, select failure) is covered too. Only gaps are the `PlayerDataNotFound` branch noted above. (Once the build is fixed, these should run.)

### What's good

- The consume-refresh-token-**after**-validation ordering (ownership + load before `Consume`) is the right call and is explicitly tested for token-intact-on-rejection.
- Claim-authoritative rehydration correctly re-binds a cache left on a *different* character, not just an evicted one — directly addressing the spike's correctness bug.
- Reusing/extending the existing `PlayerSummary` contract instead of adding a near-duplicate, and the thorough doc updates in `backend.md`.

---

ginderjeremiah:
Thanks for the thorough review.

**Blocking build break — fixed in b1d683d.** Added `Assert.True(login.Success);` before reading `login.Tokens` at both `:367` and `:394`. You're right about the green-claim discrepancy: my local validation grep filtered the output to errors only, so the two `CS8602` *warnings* (which only CI promotes to errors via `-warnaserror`) slipped through while the local build still "succeeded". I've since re-run a full `dotnet build` with **0 warnings / 0 errors**, and `verify-backend` is now green on the latest run. My mistake on the claim — corrected.

**Nits:**

- **`GetPlayerSummaries` / `ToAdminUser` projection duplication** — I left the commented mirror deliberately. EF can't translate a referenced `Expression<Func<…>>` when it's nested inside another expression tree (the `u.Players.Select(…)` in `ToAdminUser`), so a shared expression would only DRY the top-level `GetPlayerSummaries` selector and `ToAdminUser` would still need its own inline copy — i.e. the duplication doesn't actually go away without LINQKit-style machinery. For a 5-field projection I judged that not worth a dependency; the mirror comments flag the coupling. Happy to extract it (or file a follow-up) if you'd prefer.

- **`SelectPlayerStatus.PlayerDataNotFound` branch test** — I left this uncovered intentionally. It's only reachable as a TOCTOU race: ownership passes (`GetPlayerIds` contains the id) but `GetPlayer` then returns null, which requires the player row to vanish *between* those two reads inside one `SelectPlayer` call. With the project's classical/no-mocks integration tests (real Postgres, no seam to stub the repo) there's no deterministic way to interleave that, and per the testing guidelines such defensive race-guards are accepted as measure-only. (Note a genuinely *unloadable* owned player throws under the fail-fast missing-reference policy rather than returning null, so `null` really is just the delete-race.) If you'd rather, I can add an `IPlayerRepository` test double purely for this branch, but that would break the no-mocks convention for one defensive line.

The remaining test jobs are running; I'll keep watching and confirm once it's fully green.

---

PR #1086, opened 2026-06-21
Notify users to allowlist the game as always-active on background time-loss (#1074)

### LINKED ISSUE #1074: Notify users to allowlist the game as always-active when idle time is lost to tab backgrounding
Follow-up surfaced by the #922 spike. See `docs/spikes/922-multiple-players-per-account.md`. Interim mitigation for the background-time-loss problem researched in #1073.

## Context

The client-driven idle loop (`logical-engine.ts`) is throttled when the browser backgrounds the tab, and the engine **discards** over-budget time (the `tickSizeX5` clamp), so a backgrounded character silently loses idle progress. It is not recovered by offline rewards because the socket stays alive (the character isn't "away").

## Goal

Detect meaningful background time-loss and nudge the user to add the game to their browser's "never sleep / always active" list, so the foreground idle rate continues while the tab is hidden.

## What to build

- Detect when significant idle time was lost to backgrounding (e.g. a large gap between elapsed wall-time and processed ticks after a hidden period).
- Show a one-time, dismissible notice with browser-specific guidance: Chrome/Edge Memory Saver "sites that can never sleep"; note Firefox/Safari equivalents where they exist (and say so plainly where they don't).

## Notes

This is a mitigation, not the fix. The robust fix (recover the lost time via a mid-session elapsed-time catch-up) is researched in the spike #1073.

ginderjeremiah:
## Summary

Interim mitigation for the background-tab time-loss problem (follow-up surfaced by the #922 spike; the robust fix is researched in spike #1073).

The client-driven idle loop (`logical-engine.ts`) is throttled when the browser backgrounds the tab, so the engine's catch-up cap (`tickSizeX5`) **discards** the over-budget time. A backgrounded character silently loses idle progress, and offline rewards never recover it because the socket stays alive (the character isn't "away"). This PR detects a meaningful loss and nudges the user — once — to add the game to their browser's always-active list so the foreground rate keeps running while the tab is hidden.

## How it addresses the issue

- **`LogicalEngine` reports discarded idle time** via a new `onIdleTimeLost` hook, fired from the existing catch-up-cap branch (the one place over-budget time is dropped) — no behavioural change to the loop itself.
- **`BackgroundThrottleMonitor`** (`UI/src/lib/engine/background-throttle-notice.ts`) accumulates lost time **only while the tab is hidden** (so a foreground main-thread stall, which the cap can also clip, doesn't trigger it). On return to the foreground, if the hidden period cost more than the threshold (default **60s** of lost idle), it shows a **one-time, dismissible** warning toast and persists a `localStorage` flag so it never nags again.
- **Browser-specific guidance:** Chrome → Settings → Performance "Always keep these sites active"; Edge → System and performance "Never put these sites to sleep"; Firefox and Safari have no per-site equivalent, so the notice says so plainly and falls back to keeping the tab focused; unknown UAs get generic guidance.
- Wired into the engine lifecycle alongside the logical engine (`engine.ts`), torn down on stop/`SocketReplaced`.

## Notes / judgement calls

- **Toast over modal.** Used the existing sticky, dismissible `toastWarning` (a non-blocking nudge) rather than a blocking modal that would interrupt the moment the user returns to the tab.
- **Per-hidden-period detection.** The accumulator resets each time the tab hides, so the threshold is measured per hidden period (matching the issue's "after a hidden period"); the persisted flag makes it one-time across the whole account/browser.
- **Threshold** is a single named constant (`LOST_TIME_NOTICE_THRESHOLD_MS`) for easy tuning.

## Test plan

- **`background-throttle-notice.test.ts`** (new): notice fires past the threshold (one-time, dismissible, `duration: 0`); below-threshold no-op; foreground loss ignored; each hidden period measured independently; at-most-once via the persisted flag; stop unsubscribes; idempotent start; stop-before-start no-op; graceful degradation when `localStorage` is unavailable; SSR (no `document`) no-op. Plus `backgroundThrottleGuidance` per-browser (Edge-before-Chrome ordering, Firefox, Safari, unknown, no-navigator).
- **`logical-engine.test.ts`**: `onIdleTimeLost` reports the discarded ms on a long stall and stays silent within the cap.
- Updated `engine.test.ts`'s `logical-engine` mock for the new export.
- Full frontend suite green: **2298** unit tests; `npm run lint` clean (eslint + svelte-check); coverage floors pass. Frontend-only, no battle-step logic touched (backend parity unaffected).

Closes #1074

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qmn37Gvbb9dRmfND1rytwz)_

### DIFF
--- a/UI/src/lib/engine/background-throttle-notice.ts
+++ b/UI/src/lib/engine/background-throttle-notice.ts
@@ -1,0 +1,99 @@
+import { onIdleTimeLost } from './logical-engine';
+import { toastWarning } from '$stores';
+import { safeLocalStorage } from '$lib/common/local-storage';
+import type { Action } from '$lib/common';
+
+// A backgrounded tab is throttled by the browser, so the logical engine discards the idle time it
+// can't process within its catch-up cap (see logical-engine `update`). Once a single hidden period
+// has silently cost the player more than this much idle progress, nudge them — once, ever — to
+// allowlist the game as always-active so the foreground rate keeps running while the tab is hidden.
+const LOST_TIME_NOTICE_THRESHOLD_MS = 60_000;
+const NOTICE_SHOWN_KEY = 'gs:bgThrottleNoticeShown';
+
+const BASE_MESSAGE = 'Idle progress pauses while this tab runs in the background.';
+
+/**
+ * The notice text, tailored to the current browser. Chromium browsers expose a per-site
+ * always-active list; Firefox and Safari have no per-site equivalent, so we say so plainly and
+ * fall back to keeping the tab in the foreground.
+ */
+export const backgroundThrottleGuidance = (): string => {
+	const ua = typeof navigator !== 'undefined' ? navigator.userAgent : '';
+	// Edge's UA also contains "Chrome", so it must be matched first.
+	if (/Edg\//.test(ua)) {
+		return `${BASE_MESSAGE} To keep progressing, open Edge Settings → System and performance and add this site to “Never put these sites to sleep.”`;
+	}
+	if (/Chrome\//.test(ua)) {
+		return `${BASE_MESSAGE} To keep progressing, open Chrome Settings → Performance and add this site to “Always keep these sites active.”`;
+	}
+	if (/Firefox\//.test(ua)) {
+		return `${BASE_MESSAGE} Firefox has no per-site always-active setting — keep the game in a focused window to keep progressing.`;
+	}
+	if (/Safari\//.test(ua)) {
+		return `${BASE_MESSAGE} Safari has no per-site always-active setting — keep the game tab in the foreground to keep progressing.`;
+	}
+	return `${BASE_MESSAGE} Check your browser’s performance settings for a “keep sites active” option, or keep the game tab in the foreground.`;
+};
+
+/**
+ * Watches for meaningful idle time lost to background-tab throttling and shows a one-time,
+ * dismissible notice nudging the user to allowlist the game as always-active. Tied to the engine
+ * lifecycle (started/stopped alongside the logical engine in `engine.ts`).
+ */
+export class BackgroundThrottleMonitor {
+	private unhookIdleTimeLost?: Action;
+	private lostWhileHiddenMs = 0;
+	private running = false;
+
+	public start() {
+		if (this.running || typeof document === 'undefined') {
+			return;
+		}
+		this.running = true;
+		this.lostWhileHiddenMs = 0;
+		this.unhookIdleTimeLost = onIdleTimeLost((lostMs) => this.recordLostTime(lostMs));
+		document.addEventListener('visibilitychange', this.handleVisibilityChange);
+	}
+
+	public stop() {
+		if (!this.running) {
+			return;
+		}
+		this.running = false;
+		this.lostWhileHiddenMs = 0;
+		this.unhookIdleTimeLost?.();
+		this.unhookIdleTimeLost = undefined;
+		document.removeEventListener('visibilitychange', this.handleVisibilityChange);
+	}
+
+	// Only count time discarded while the tab is hidden: the catch-up cap can also clip a foreground
+	// main-thread stall, but the backgrounding case is the one this notice addresses.
+	private recordLostTime(lostMs: number) {
+		if (document.hidden) {
+			this.lostWhileHiddenMs += lostMs;
+		}
+	}
+
+	private handleVisibilityChange = () => {
+		if (document.hidden) {
+			// Entering a new hidden period — measure it on its own.
+			this.lostWhileHiddenMs = 0;
+			return;
+		}
+		// Back in the foreground: if the hidden period cost a meaningful amount of idle progress, nudge.
+		if (this.lostWhileHiddenMs >= LOST_TIME_NOTICE_THRESHOLD_MS) {
+			this.maybeShowNotice();
+		}
+		this.lostWhileHiddenMs = 0;
+	};
+
+	private maybeShowNotice() {
+		const storage = safeLocalStorage();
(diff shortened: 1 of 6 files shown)


UI/src/lib/engine/background-throttle-notice.ts
```diff
@@ -0,0 +1,99 @@
...
+import { toastWarning } from '$stores';
+import { safeLocalStorage } from '$lib/common/local-storage';
+import type { Action } from '$lib/common';
+
+// A backgrounded tab is throttled by the browser, so the logical engine discards the idle time it
+// can't process within its catch-up cap (see logical-engine `update`). Once a single hidden period
+// has silently cost the player more than this much idle progress, nudge them — once, ever — to
+// allowlist the game as always-active so the foreground rate keeps running while the tab is hidden.
+const LOST_TIME_NOTICE_THRESHOLD_MS = 60_000;
+const NOTICE_SHOWN_KEY = 'gs:bgThrottleNoticeShown';
```
ginderjeremiah: **Should-fix (consistency):** every other persisted key in the app uses the `gameserver.` namespace — `gameserver.auth-tokens` (token store), `gameserver.device-fingerprint`, `gameserver.refdata.*`. This one uses a different `gs:` prefix, so it won't visually group with the others when inspecting storage. Suggest `gameserver.bgThrottleNoticeShown` to match the established convention.

---

UI/src/lib/engine/background-throttle-notice.ts
```diff
@@ -0,0 +1,99 @@
...
+		this.unhookIdleTimeLost = undefined;
+		document.removeEventListener('visibilitychange', this.handleVisibilityChange);
+	}
+
+	// Only count time discarded while the tab is hidden: the catch-up cap can also clip a foreground
+	// main-thread stall, but the backgrounding case is the one this notice addresses.
+	private recordLostTime(lostMs: number) {
+		if (document.hidden) {
+			this.lostWhileHiddenMs += lostMs;
+		}
```
ginderjeremiah, line 86: **Should-fix (detection gap worth documenting):** this `document.hidden` guard means the notice only fires when the hidden tab still gets *throttled* timer fires while hidden (Chrome's intensive throttling, ~1/min after 5 min hidden) — those polls run with `document.hidden === true` and accumulate here. But a tab the browser fully **freezes/discards** (Memory Saver — the exact audience this notice targets) runs *no* timer while hidden; its one big catch-up poll fires on resume, *after* `visibilitychange` has already flipped `document.hidden` to `false`, so this guard drops it and `handleVisibilityChange` has already evaluated a zero accumulator → no notice.

It's a fair limitation for an interim mitigation, but at minimum it deserves a comment so a future reader doesn't assume it catches all background loss. If you want it to cover the freeze case, attribute the first lost-time event right after a hidden→visible transition to the hidden period that just ended (e.g. a short-lived "expect catch-up" flag set in `handleVisibilityChange`), rather than discarding it as a foreground stall.

---

ginderjeremiah (commented):
## Review: Notify users to allowlist the game as always-active on background time-loss (#1074)

**Verdict:** Approve (posted as a comment since I'm the PR author) — solid, well-tested interim mitigation that matches the issue's intent. Two should-fixes below, neither blocking.

This is the right shape for the issue: an additive `onIdleTimeLost` hook on the one branch in `LogicalEngine.update` that already discards over-budget time (no behavioural change to the loop, no battle-step arithmetic touched, so backend parity is unaffected), plus a lifecycle-managed `BackgroundThrottleMonitor` that fires a one-time, dismissible sticky toast (`duration: 0`). Lifecycle wiring is complete — started in `startLogicEngine`, and torn down in both `stopGame` (which `handleSocketReplaced` calls) and the `onDestroy`. Not statifying the monitor is correct (no reactive state), and the explicit comment saying so is good.

### Should-fix

- **`background-throttle-notice.ts:11`** — localStorage key uses a `gs:` prefix; every other persisted key uses `gameserver.`. Inline comment with suggested rename.
- **`background-throttle-notice.ts:71-74`** — the `document.hidden` guard makes detection depend on the tab getting throttled (not frozen) timer fires while hidden. A fully frozen/discarded tab — the Memory Saver audience this notice targets — never accumulates loss, because its single catch-up poll arrives after `visibilitychange`. Works in the common intensive-throttling regime; worth documenting the limitation (or handling the post-resume catch-up). Details inline.

### Nits

- `background-throttle-notice.test.ts` redefines `document.hidden` via `Object.defineProperty` in `beforeEach` but `afterEach` only calls `vi.unstubAllGlobals()`, which doesn't undo a `defineProperty`. Harmless within this file, but the redefinition leaks past it. Minor.

### Test coverage

Strong and follows the project's conventions — test mirrors source path (`tests/lib/engine/...`), classical style with minimal doubles. Covers threshold/below-threshold, foreground-stall ignore, per-hidden-period independence, at-most-once via the persisted flag, idempotent start, stop-before-start, localStorage-unavailable degradation, SSR no-op, and per-browser guidance (incl. the Edge-before-Chrome ordering). The mirrored `onIdleTimeLost` assertions were added to `logical-engine.test.ts` and the `engine.test.ts` mock updated. No battle-step logic changed, so no backend mirror is required. One gap: no test exercises the freeze-case behaviour described above (expected, since it's the limitation itself).

### What's good

- Catch-up branch reused as the single source of "lost time" rather than re-deriving it — the hook fires exactly where the time is dropped.
- Persist-before-show ordering so the one-time guarantee holds even if the toast is dismissed instantly.
- Graceful degradation when `localStorage` is unavailable, and SSR-safe (`typeof document` guard).
- Browser-specific guidance with the correct Edge-before-Chrome UA ordering, and honest messaging for Firefox/Safari where no per-site setting exists.

---

PR #1088, opened 2026-06-21
Restore codegen unmapped-struct safety net for Guid/TimeSpan (#1081)

### LINKED ISSUE #1081: Codegen: `IsStructThatNeedsInterface` defeats its own unmapped-type safety net for Guid/TimeSpan (+ minor codegen cleanups)
## Primary issue — the unmapped-type guard is unreachable for the types it names

`CodeGenTypeFormatter.GetTypeText` has a deliberate guard that throws on an unmapped type so a broken client surfaces at generation time instead of silently:

```csharp
// Game.Api/CodeGen/CodeGenTypeFormatter.cs:84-96
else if (type.NeedsInterface()) { ... }
else
{
    // A type with no TypeScript mapping that NeedsInterface() also rejects (e.g. Guid,
    // TimeSpan, byte, char) would otherwise silently emit a reference to an interface ...
    throw new InvalidOperationException(...);
}
```

But `NeedsInterface()` does **not** reject `Guid`/`TimeSpan`. `IsStructThatNeedsInterface` returns `true` for any non-primitive struct except `DateTime`/`decimal`:

```csharp
// Game.Api/CodeGen/Extensions.cs:91-94
return type.IsValueType && !type.IsPrimitive && type != typeof(DateTime) && type != typeof(decimal);
```

`Guid`, `TimeSpan`, `DateTimeOffset` are non-primitive structs, so `NeedsInterface()` returns `true` and they take the `else if` branch — `ApiInterfaceWriter` emits a nonsense `IGuid`/`ITimeSpan` interface from the struct's public properties — instead of reaching the `throw`. (`byte`/`char` *are* primitives, so they correctly fall through to the throw; the comment is only wrong about `Guid`/`TimeSpan`.)

**Latent** — no DTO uses these types today — but the safety net is provably broken for exactly the types it was written to catch. The first DTO to use a `Guid`/`TimeSpan` would get a silently-broken client.

**Fix:** make `IsStructThatNeedsInterface` an allow-list of structs actually intended to mirror, or explicitly exclude `Guid`/`TimeSpan`/`DateTimeOffset` so they fall through to the `throw`.

## Minor codegen cleanups (bundle)

Low-severity, codegen-only (so runtime perf is irrelevant — flagged for robustness/consistency):

1. **Culture-sensitive route check** — `ControllerMetadataExtractor.cs:51-55` uses `route.StartsWith(apiPrefix)` with the default culture-sensitive comparison; should be `StringComparison.Ordinal` for a structural match.
2. **`IsGet` defaults to `true`** — `ControllerMetadataExtractor.cs:61`: `methodAtt?.HttpMethods?.Contains("GET") ?? true` treats an action with no explicit HTTP-method attribute as GET. Confirm intended; an unannotated POST action would generate the wrong client verb.
3. **Interpolated string to structured logger** — `ApiCodeGenerator.cs:24`: `_logger.LogDebug($"Beginning code generation.")` should be a plain template (CA2254; inconsistent with the structured templates around it).
4. **Multi-enumeration of deferred LINQ** — `ApiInterfaceWriter.cs:42-82`: the lazy `GroupBy` (`interfaceDataGroups`) is enumerated in the `foreach` and again inside `GetExportsText`, re-running the reflection-heavy `SelectMany(GetAllUsedDescriptors)` + `DistinctBy` each pass. Materialize once with `.ToList()`.

ginderjeremiah:
## Summary

`CodeGenTypeFormatter.GetTypeText` has a deliberate `throw` for any type with no TypeScript mapping, so a broken client surfaces at generation time. But `NeedsInterface()` did **not** reject `Guid`/`TimeSpan`/`DateTimeOffset`: `IsStructThatNeedsInterface` was a deny-list (`IsValueType && !IsPrimitive && != DateTime && != decimal`), so those non-primitive BCL structs took the `NeedsInterface` branch and `ApiInterfaceWriter` would emit a nonsense `IGuid`/`ITimeSpan` interface instead of reaching the `throw`. Latent today (no DTO uses them), but the safety net was provably broken for exactly the types it names.

### Primary fix — allow-list instead of deny-list

`IsStructThatNeedsInterface` is now an explicit **allow-list** of structs intended to mirror (currently empty — every DTO in the codebase is a class, and no struct is mirrored today). This is strictly more robust than excluding the three named structs: it makes the safety net catch **every** unmapped struct (Guid, TimeSpan, DateTimeOffset, and future ones like `DateOnly`/`Half`/`Int128`), aligning with the project's loud fail-fast policy. `DateTime`/`decimal` are mapped to primitives directly in `GetTypeText`, so they never reach here. Confirmed no codegen drift — regenerating the TS client produces no changes.

### Bundled minor cleanups

1. **Culture-sensitive route check** — `route.StartsWith(apiPrefix)` now uses `StringComparison.Ordinal`.
2. **Interpolated log template** — `LogDebug($"Beginning code generation.")` → plain template (CA2254).
3. **Multi-enumeration of deferred LINQ** — `interfaceDataGroups` is now materialized with `.ToList()`, so the reflection-heavy `SelectMany(GetAllUsedDescriptors)` + `DistinctBy` walk runs once instead of again inside `GetExportsText`.
4. **`IsGet` default** — left as-is (every action in the codebase carries an explicit HTTP-method attribute), with a comment documenting that an unannotated action falls back to GET. This was a "confirm intended" item, not a defect.

## Test plan

- **New tests:** `NeedsInterface` returns `false` for `Guid`/`TimeSpan`/`DateTimeOffset`/`List<Guid>`; `GetTypeText` throws (with the type name in the message) for `Guid`/`TimeSpan`/`DateTimeOffset` properties — mirroring the existing `byte`/`char` throw tests. New `ModelWithUnmappedStruct` fixture.
- Full **Game.Api.Tests** suite green: **562** passed (0 failed).
- `dotnet build` clean (0 warnings); `dotnet format --verify-no-changes` clean on `Game.Api` and `Game.Api.Tests`.
- TS client regenerated — **no drift**.

Closes #1081

---
_Generated by [Claude Code](https://claude.ai/code/session_011Lpb9tnUbsaxraR3hWEw9U)_

### DIFF
--- a/Game.Api.Tests/CodeGen/CodeGenExtensionsTests.cs
+++ b/Game.Api.Tests/CodeGen/CodeGenExtensionsTests.cs
@@ -131,6 +131,32 @@ namespace Game.Api.Tests.CodeGen
         }
 
         [Fact]
+        public void NeedsInterface_Guid_ReturnsFalse()
+        {
+            // Guid is a non-primitive struct with no TypeScript mapping; it must not generate an interface.
+            Assert.False(typeof(Guid).NeedsInterface());
+        }
+
+        [Fact]
+        public void NeedsInterface_TimeSpan_ReturnsFalse()
+        {
+            Assert.False(typeof(TimeSpan).NeedsInterface());
+        }
+
+        [Fact]
+        public void NeedsInterface_DateTimeOffset_ReturnsFalse()
+        {
+            Assert.False(typeof(DateTimeOffset).NeedsInterface());
+        }
+
+        [Fact]
+        public void NeedsInterface_ListOfGuid_ReturnsFalse()
+        {
+            // The element type is the unmapped struct, so the enumerable must not generate an interface.
+            Assert.False(typeof(List<Guid>).NeedsInterface());
+        }
+
+        [Fact]
         public void NeedsInterface_ListOfClass_ReturnsTrue()
         {
             // For enumerables, checks the generic argument
--- a/Game.Api.Tests/CodeGen/CodeGenTypeFormatterTests.cs
+++ b/Game.Api.Tests/CodeGen/CodeGenTypeFormatterTests.cs
@@ -304,6 +304,32 @@ namespace Game.Api.Tests.CodeGen
             var ex = Assert.Throws<InvalidOperationException>(() => CodeGenTypeFormatter.GetTypeText(descriptor));
             Assert.Contains("Char", ex.Message);
         }
+
+        [Fact]
+        public void GetTypeText_UnmappedGuid_Throws()
+        {
+            // Guid is a non-primitive struct with no TypeScript mapping. Before #1081 NeedsInterface let it
+            // through and the formatter emitted a nonsense IGuid; it must now throw at generation time.
+            var descriptor = GetPropertyDescriptor<ModelWithUnmappedStruct>("GuidValue");
+            var ex = Assert.Throws<InvalidOperationException>(() => CodeGenTypeFormatter.GetTypeText(descriptor));
+            Assert.Contains("Guid", ex.Message);
+        }
+
+        [Fact]
+        public void GetTypeText_UnmappedTimeSpan_Throws()
+        {
+            var descriptor = GetPropertyDescriptor<ModelWithUnmappedStruct>("TimeSpanValue");
+            var ex = Assert.Throws<InvalidOperationException>(() => CodeGenTypeFormatter.GetTypeText(descriptor));
+            Assert.Contains("TimeSpan", ex.Message);
+        }
+
+        [Fact]
+        public void GetTypeText_UnmappedDateTimeOffset_Throws()
+        {
+            var descriptor = GetPropertyDescriptor<ModelWithUnmappedStruct>("DateTimeOffsetValue");
+            var ex = Assert.Throws<InvalidOperationException>(() => CodeGenTypeFormatter.GetTypeText(descriptor));
+            Assert.Contains("DateTimeOffset", ex.Message);
+        }
     }
 
     public class GenericHolder
--- a/Game.Api.Tests/CodeGen/TestFixtures.cs
+++ b/Game.Api.Tests/CodeGen/TestFixtures.cs
@@ -92,6 +92,17 @@ namespace Game.Api.Tests.CodeGen
         public char CharValue { get; set; }
     }
 
+    // A model whose property types are non-primitive BCL structs (Guid, TimeSpan, DateTimeOffset) with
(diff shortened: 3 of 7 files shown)


ginderjeremiah (commented):
## Review: Restore codegen unmapped-struct safety net for Guid/TimeSpan (#1081)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the fix correctly closes the safety-net hole, the cleanups all match what #1081 asked for, and coverage is solid. No blocking or should-fix findings.

### Why the fix is correct
- `IsStructThatNeedsInterface` now gates on an empty `MirroredStructs` allow-list, so **every** value-type struct returns `false` from `NeedsInterface()` and falls through to `GetTypeText`'s `throw` (`CodeGenTypeFormatter.cs:88-97`). This is strictly more robust than the old deny-list — it catches `Guid`/`TimeSpan`/`DateTimeOffset` plus any future BCL struct (`DateOnly`, `Half`, `Int128`, …) rather than only the three named ones.
- The mapped primitives are handled before the struct branch is ever reached: `decimal`/`int`/etc. → `"number"`, `DateTime` → `"string"`, enums via `type.IsEnum`. So the empty allow-list can't regress them. Confirmed against `GetTypeText` and `NeedsInterface`.
- The `type.IsValueType &&` guard is technically redundant against an allow-list that will only ever hold structs, but it's a reasonable defensive guard that documents intent — not worth changing.

### Cleanups — all match the issue
- `StringComparison.Ordinal` on the route prefix check — correct structural match.
- `LogDebug` interpolated → plain template — resolves CA2254, consistent with surrounding structured templates.
- `.ToList()` on `interfaceDataGroups` — materializes once; the grouping is enumerated by the `foreach` and again in `GetExportsText`, so this avoids re-running the reflection-heavy `SelectMany(GetAllUsedDescriptors)` + `DistinctBy` walk.
- `IsGet` left as-is with a documenting comment — appropriate; it was a "confirm intended" item, not a defect.

### Test coverage
Good. New `NeedsInterface` false-cases for `Guid`/`TimeSpan`/`DateTimeOffset`/`List<Guid>` and `GetTypeText` throw-cases (asserting the type name is in the message) mirror the existing `byte`/`char` throw tests exactly, with a dedicated `ModelWithUnmappedStruct` fixture. This is codegen, not domain or battle logic, so no cross-suite mirroring applies.

### What's good
Clear problem framing, a genuinely more-robust fix than the minimum, and the comments explain the non-obvious "why" (allow-list vs deny-list, why the set is empty, why `DateTime`/`decimal` never reach the struct branch) — exactly the context a future maintainer needs to avoid reintroducing the bug.

I'll hold the merge until CI is fully green (`test-backend` and the e2e jobs are still running).

---

PR #1087, opened 2026-06-21
Floor damage-over-time at zero to prevent uncapped heal (#1080)

ginderjeremiah:
## Summary

`Battler.ApplyDamageOverTime` applied the per-tick DoT unconditionally with **no floor**, unlike its two sibling methods that *do* guard:
- `TakeDamage` floors at zero.
- `ApplyHealOverTime` caps the heal at `MaxHealth`.

So a **negative** `DamageTakenPerSecond` — reachable purely through authored skill-effect content (effects can apply arbitrary additive/multiplicative magnitudes to any attribute via the admin Workbench) — would make `CurrentHealth -= damage` *add* health with **no `MaxHealth` cap** (an unbounded heal) and book negative values into `DamageDealt`/`DamageTaken` (and thus challenge progress). The frontend mirrored the same unclamped subtraction, so both simulators agreed — and both were wrong in the same way.

Not currently exploitable (no authored effect produces a negative DoT today), so this is hardening rather than an active defect.

## How it addresses the issue

Clamps DoT to `>= 0` before applying, mirroring `TakeDamage`, on **both** sides (a parity surface — must stay in lockstep):

- `Game.Core/Battle/Battler.cs` — `ApplyDamageOverTime`
- `UI/src/lib/battle/battler.ts` — `applyDamageOverTime`

A heal must instead go through the capped `ApplyHealOverTime` channel rather than silently bypassing the `MaxHealth` cap.

## Test plan

Added mirrored unit/parity tests for a negative-magnitude DoT (battler starts below full health to prove no heal occurs):

- `Game.Core.Tests/Battle/BattleDamageOverTimeTests.cs` — `ApplyDamageOverTime_NegativePerSecondAttribute_IsFlooredAtZero_NoHeal`
- `UI/src/tests/lib/battle/battler-dot.test.ts` — `floors a negative DamageTakenPerSecond at zero (no heal past the cap)`

Verified green locally:
- Backend: `BattleDamageOverTimeTests` — 9 passed.
- Frontend: `battler-dot.test.ts` — 6 passed; `npm run lint` clean (0 errors/warnings).

Closes #1080

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01HmroRyuR5sc4bqR7BRnZfa)_

Issue #1080, opened 2026-06-21
Damage-over-time path doesn't clamp a negative `DamageTakenPerSecond`, allowing an uncapped heal (FE + BE parity)


ginderjeremiah:
## Problem

`Battler.ApplyDamageOverTime` applies the per-tick DoT unconditionally with no floor:

```csharp
// Game.Core/Battle/Battler.cs:88-93
public double ApplyDamageOverTime(int ms)
{
    var damage = _attributes[DamageTakenPerSecond] * ms / 1000.0;
    CurrentHealth -= damage;          // no `damage > 0` clamp
    return damage;
}
```

This is asymmetric with two sibling methods that *do* guard:
- `TakeDamage` floors at zero (`Game.Core/Battle/Battler.cs:75-81`).
- `ApplyHealOverTime` caps the heal at `MaxHealth` (`Battler.cs:100-111`).

If a skill effect or gear ever yields a **negative** `DamageTakenPerSecond` (skill effects can apply arbitrary additive/multiplicative magnitudes to any attribute via the admin Workbench), `CurrentHealth -= damage` *adds* health with **no `MaxHealth` cap** — an unbounded heal through authored content — and books negative values into `DamageDealt`/`DamageTaken` (and thus challenge progress).

The frontend mirrors the same unclamped subtraction (`UI/src/lib/battle/battler.ts:155-156`), so the two simulators currently agree — but both are wrong in the same way.

## Why it matters

It's an unbounded-heal vector reachable purely through content authoring, and the asymmetry with the already-capped `TakeDamage`/`ApplyHealOverTime` is surprising. Not currently exploitable (no authored effect produces a negative DoT today), so this is hardening rather than an active defect.

## Suggested fix

Clamp DoT to `>= 0` before applying, mirroring `TakeDamage`, on **both** sides (parity surface — must stay in lockstep and get a mirrored parity vector):

- `Game.Core/Battle/Battler.cs` `ApplyDamageOverTime`
- `UI/src/lib/battle/battler.ts` `applyDamageOverTime`

If a negative DoT is ever intended to mean "heal", it should instead be routed through the capped HoT channel rather than silently bypassing the `MaxHealth` cap. Add mirrored unit/parity tests for a negative-magnitude DoT effect.

_Found during a general code-health review._

PR #1089, opened 2026-06-21
Revert DoT zero-floor from #1087

ginderjeremiah:
## Summary

Reverts PR #1087 ("Floor damage-over-time at zero to prevent uncapped heal"), which clamped `ApplyDamageOverTime` / `applyDamageOverTime` at zero on both the backend and frontend.

There is no reason to restrict damage-over-time here. The zero-floor on normal `TakeDamage` exists only because mitigation effects (e.g. Defense) shouldn't be able to turn incoming damage into a heal. Damage-over-time has no such mitigation step, so a negative `DamageTakenPerSecond` from authored skill-effect content is a legitimate heal channel and should be allowed to apply.

## Changes

Restores the original unclamped subtraction and the original doc comments, and removes the parity tests that PR #1087 added:

- `Game.Core/Battle/Battler.cs` — `ApplyDamageOverTime`
- `UI/src/lib/battle/battler.ts` — `applyDamageOverTime`
- `Game.Core.Tests/Battle/BattleDamageOverTimeTests.cs` — removed `ApplyDamageOverTime_NegativePerSecondAttribute_IsFlooredAtZero_NoHeal`
- `UI/src/tests/lib/battle/battler-dot.test.ts` — removed the matching negative-DoT parity test

## Test plan

- Backend: `Game.Core.Tests` — 685 passed.
- Frontend: `battler-dot.test.ts` — 5 passed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JfNLbK36Cm7uvwscG5qrNY)_

### DIFF
--- a/Game.Core/Battle/Battler.cs
+++ b/Game.Core/Battle/Battler.cs
@@ -83,15 +83,11 @@ namespace Game.Core.Battle
         /// <summary>
         /// Applies one tick of damage-over-time from <see cref="DamageTakenPerSecond"/> (authored per second,
         /// scaled to <paramref name="ms"/>). Unlike <see cref="TakeDamage"/> it <b>bypasses Defense</b>, and
-        /// returns the damage dealt so the caller can attribute it to the battle statistics. Floored at zero
-        /// (like <see cref="TakeDamage"/>) so a negative <see cref="DamageTakenPerSecond"/> can't heal past the
-        /// <see cref="MaxHealth"/> cap or book negative statistics — a heal must go through the capped
-        /// <see cref="ApplyHealOverTime"/> channel instead.
+        /// returns the damage dealt so the caller can attribute it to the battle statistics.
         /// </summary>
         public double ApplyDamageOverTime(int ms)
         {
             var damage = _attributes[DamageTakenPerSecond] * ms / 1000.0;
-            damage = damage > 0 ? damage : 0;
             CurrentHealth -= damage;
             return damage;
         }
--- a/Game.Core/Battle/Battler.cs
+++ b/Game.Core/Battle/Battler.cs
@@ -85,6 +85,13 @@ namespace Game.Core.Battle
         /// scaled to <paramref name="ms"/>). Unlike <see cref="TakeDamage"/> it <b>bypasses Defense</b>, and
         /// returns the damage dealt so the caller can attribute it to the battle statistics.
         /// </summary>
+        /// <remarks>
+        /// Intentionally <b>not</b> floored at zero, unlike <see cref="TakeDamage"/>. That floor exists only so
+        /// Defense mitigation can't drive net damage below zero and turn a hit into a heal; DoT has no mitigation
+        /// step, so a tick is negative only if a negative <see cref="DamageTakenPerSecond"/> is deliberately
+        /// authored — and a floor wouldn't prevent that, it would just silently rewrite the authored value to
+        /// zero. Authored healing belongs in the capped <see cref="ApplyHealOverTime"/> channel instead.
+        /// </remarks>
         public double ApplyDamageOverTime(int ms)
         {
             var damage = _attributes[DamageTakenPerSecond] * ms / 1000.0;
--- a/UI/src/lib/battle/battler.ts
+++ b/UI/src/lib/battle/battler.ts
@@ -150,12 +150,9 @@ export class Battler {
 	}
 
 	/** Applies one tick of damage-over-time from DamageTakenPerSecond (authored per second, scaled to
-	 *  `timeDelta`). Unlike {@link takeDamage} it BYPASSES Defense; returns the damage dealt. Floored at zero
-	 *  (like {@link takeDamage}) so a negative DamageTakenPerSecond can't heal past the MaxHealth cap — a heal
-	 *  must go through the capped {@link applyHealOverTime} channel instead. */
+	 *  `timeDelta`). Unlike {@link takeDamage} it BYPASSES Defense; returns the damage dealt. */
 	public applyDamageOverTime(timeDelta: number) {
-		const rawDamage = (this.attributes.getValue(EAttribute.DamageTakenPerSecond) * timeDelta) / 1000;
-		const damage = rawDamage > 0 ? rawDamage : 0;
+		const damage = (this.attributes.getValue(EAttribute.DamageTakenPerSecond) * timeDelta) / 1000;
 		this.currentHealth -= damage;
 		this.isDead = this.currentHealth <= 0;
 		return damage;
--- a/UI/src/lib/battle/battler.ts
+++ b/UI/src/lib/battle/battler.ts
@@ -150,7 +150,13 @@ export class Battler {
 	}
 
 	/** Applies one tick of damage-over-time from DamageTakenPerSecond (authored per second, scaled to
-	 *  `timeDelta`). Unlike {@link takeDamage} it BYPASSES Defense; returns the damage dealt. */
+	 *  `timeDelta`). Unlike {@link takeDamage} it BYPASSES Defense; returns the damage dealt.
+	 *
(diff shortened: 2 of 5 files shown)


Game.Core/Battle/Battler.cs
```diff
@@ -83,15 +83,11 @@ public double TakeDamage(double rawDamage, double blockReduction = 0)
...
-        /// <see cref="ApplyHealOverTime"/> channel instead.
+        /// returns the damage dealt so the caller can attribute it to the battle statistics.
         /// </summary>
         public double ApplyDamageOverTime(int ms)
         {
             var damage = _attributes[DamageTakenPerSecond] * ms / 1000.0;
-            damage = damage > 0 ? damage : 0;
             CurrentHealth -= damage;
             return damage;
         }
```
ginderjeremiah, line 100: Removing the floor reintroduces the exact two problems #1080 documented and #1087 fixed — and the PR's stated rationale doesn't engage with either of them:

1. **Uncapped heal past `MaxHealth`.** With the floor gone, a negative `DamageTakenPerSecond` makes `CurrentHealth -= damage` *add* health with no cap. Everywhere else this class holds the invariant `CurrentHealth <= MaxHealth`: `ApplyHealOverTime` caps the heal, and `ClampHealthToMaxHealth` re-clamps on every effect/MaxHealth change. So a negative-DoT heal pushes `CurrentHealth` above `MaxHealth`, where it then gets *silently snapped back down* on the next `ClampHealthToMaxHealth` (effect expiry or a MaxHealth shift). That's not stable overheal — it's an invariant violation with surprising, order-dependent behavior. `game-design.md` also states heal is "capped at MaxHealth."

2. **Negative statistics.** In `BattleContext.ResolveDamageOverTime`, `Stats.PlayerDamageDealt += ApplyDamageOverTime(...)` (and `PlayerDamageTaken`) will *decrease* on a negative tick, corrupting cumulative damage stats and challenge progress.

The PR says the floor "exists only because mitigation effects shouldn't turn damage into a heal." That's the rationale for `TakeDamage`'s floor, not this one — the DoT floor was about the MaxHealth cap and statistics (see #1080), so the justification rebuts a problem that wasn't the reason here.

If healing through authored content is genuinely the goal, the existing mechanism is the **capped** `HealthRegenPerSecond` / `ApplyHealOverTime` channel. A negative DoT as a heal is both redundant with it and strictly stronger (uncapped), which is a balance foot-gun. If you still want this behavior, it needs: (a) a `MaxHealth` cap on the heal path, (b) a decision on the negative-stats booking, and (c) a `game-design.md` update — not just removing the guard.

---

UI/src/lib/battle/battler.ts
```diff
@@ -150,12 +150,9 @@ export class Battler {
...
 
 	/** Applies one tick of damage-over-time from DamageTakenPerSecond (authored per second, scaled to
-	 *  `timeDelta`). Unlike {@link takeDamage} it BYPASSES Defense; returns the damage dealt. Floored at zero
-	 *  (like {@link takeDamage}) so a negative DamageTakenPerSecond can't heal past the MaxHealth cap — a heal
-	 *  must go through the capped {@link applyHealOverTime} channel instead. */
+	 *  `timeDelta`). Unlike {@link takeDamage} it BYPASSES Defense; returns the damage dealt. */
 	public applyDamageOverTime(timeDelta: number) {
-		const rawDamage = (this.attributes.getValue(EAttribute.DamageTakenPerSecond) * timeDelta) / 1000;
-		const damage = rawDamage > 0 ? rawDamage : 0;
+		const damage = (this.attributes.getValue(EAttribute.DamageTakenPerSecond) * timeDelta) / 1000;
```
ginderjeremiah, line 161: Same concern on the frontend parity surface: a negative `DamageTakenPerSecond` heals uncapped past `MaxHealth` here too. The two sides stay in lockstep (good), but both reintroduce the #1080 behavior. See the backend comment for the full rationale — whatever the resolution, it must land on both sides with mirrored tests.

---

ginderjeremiah (commented):
## Review: Revert of #1087 (DoT zero-floor), restoring unclamped `ApplyDamageOverTime` / `applyDamageOverTime`

**Verdict:** Request changes — the revert reintroduces the uncapped-heal and negative-statistics behavior that issue #1080 documented and #1087 deliberately fixed, and the stated rationale doesn't address either of those reasons. (Posting as `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR — but the intended verdict is *request changes*; I have not merged.)

The change itself is clean and correctly symmetric across both parity surfaces — the problem is the behavior it restores, not how it's done.

### Blocking
- **Uncapped heal past `MaxHealth`** — `Game.Core/Battle/Battler.cs:88` / `UI/src/lib/battle/battler.ts:155`. A negative `DamageTakenPerSecond` (reachable purely through authored Workbench content) makes `CurrentHealth -= damage` add health with no cap. This breaks the `CurrentHealth <= MaxHealth` invariant the rest of the class upholds (`ApplyHealOverTime` caps; `ClampHealthToMaxHealth` re-clamps on every effect/MaxHealth change). The over-cap health is then silently snapped back down at the next clamp point, so it's an unstable, order-dependent invariant violation — not an intentional overheal. `game-design.md` also documents heal as "capped at MaxHealth."
- **Negative statistics** — `BattleContext.ResolveDamageOverTime` does `Stats.PlayerDamageDealt += ApplyDamageOverTime(...)` (and `PlayerDamageTaken`). A negative tick *decreases* these cumulative stats and challenge progress, which is nonsensical for "damage dealt/taken."

### Why the rationale doesn't hold
The PR argues the floor "exists only because mitigation effects shouldn't turn damage into a heal," and that DoT has no mitigation step. That's the rationale for `TakeDamage`'s floor; this floor was about the `MaxHealth` cap and statistics (see #1080). #1080 explicitly anticipated the "negative DoT = heal" case: *"it should instead be routed through the capped HoT channel rather than silently bypassing the MaxHealth cap."* The capped `HealthRegenPerSecond` / `ApplyHealOverTime` channel already exists for authored healing, so a negative DoT is both redundant with it and strictly stronger (uncapped) — a balance foot-gun.

### Test coverage
- The revert removes the mirrored negative-DoT parity tests (`BattleDamageOverTimeTests.cs`, `battler-dot.test.ts`) **without any replacement**, leaving the negative-DoT branch unspecified on both sides. Per the project's testing rules, battle-logic edge cases must be thoroughly covered and mirrored FE/BE. If the new behavior is genuinely intended, it needs *positive* mirrored tests asserting exactly what a negative DoT does (heal amount, cap behavior, stat booking), not just deletion of the guard test.

### If allowing negative DoT really is the goal
Then this should be a deliberate, documented mechanic rather than a bare revert:
1. Decide and enforce the `MaxHealth` cap on the heal path (don't let `CurrentHealth` exceed `MaxHealth`).
2. Decide the statistics behavior (don't book negative `DamageDealt`/`DamageTaken`/challenge progress).
3. Update `game-design.md` for the DoT/HoT section.
4. Add mirrored FE/BE tests for the chosen behavior.

As written, none of these are addressed, so I'd keep #1087's clamp until the design intent is settled.

---

ginderjeremiah:
Keeping the revert as-is after weighing the review.

The floor guards a state — a negative `DamageTakenPerSecond` — that is only reachable through trusted, authored Workbench content. Unlike `TakeDamage` (whose floor exists precisely because Defense mitigation can drive *net* damage below zero), DoT has no mitigation step, so a tick can't be pushed negative by gameplay. The clamp also doesn't prevent a negative value from being authored; it silently rewrites it to `0`, which makes an authored negative DoT behave *unexpectedly* rather than addressing anything. Since no authored content produces a negative DoT and that remains the author's responsibility to control, the floor is hardening against a non-issue and is being removed.

If we ever do want authored healing, the capped `HealthRegenPerSecond` / `ApplyHealOverTime` channel already exists for that — a separate, intentional decision from this revert.

CI is green (all 9 checks passed).

---

PR #1094, opened 2026-06-21
Character creation: per-account cap & per-player names (#1069)

### LINKED ISSUE #1069: Multiple players per account: character creation & per-account cap

### LINKED ISSUE #1095: RequestLogging tests assert exact "Request Start" but middleware now logs "Request Start {Path}"
## Problem

Three `RequestLogging` tests fail on `main` (confirmed against a clean tree, independent of any feature branch):

- `Game.Api.Tests/Integration/RequestLoggingTests.cs` → `UnauthenticatedRequest_LogsRequestStartWithNullUserId`
- `Game.Api.Tests/Integration/RequestLoggingTests.cs` → `AuthenticatedRequest_LogsUserIdInScope`
- `Game.Api.Tests/Unit/RequestLoggingMiddlewareTests.cs` → `ThrowingPipeline_StillLogsRequestEnded_AndRethrows`

They assert the captured log message equals `"Request Start"` exactly, but commit `d4ed2e8` ("Log path in request start message") changed the middleware to `LogInformation("Request Start {Path}", ...)`. `CapturingLogger` stores the rendered message (`"Request Start /api/Login/Status"`), so the exact-equality predicate matches nothing and `Assert.Single` throws.

## Resolution

Fixed in **#1094** (the trivial test-only change was bundled there since it was blocking that PR's `test-backend` CI): the three assertions now use `.StartsWith("Request Start")`, consistent with the sibling "Request Ended" assertions. No production change.

ginderjeremiah:
## Summary

Implements #1069 (sub-issue of #922): let an authenticated account create additional characters beyond the one made at signup, with a server-side cap and per-character names.

- **`CreatePlayer(name)` endpoint** on `LoginController` (HTTP, part of the pre-game character-select flow — binds no session, so the caller's selected character is unchanged). Orchestrated by `AccountService.CreatePlayer`, which builds the player via the existing `NewPlayerFactory` and attaches it to the account. Returns the new character's `PlayerSummary` so the client can add it to the select list.
- **Config-bound per-account cap** (`PlayerCreation:MaxPlayersPerAccount`, default 6) enforced server-side as anti-cheat. Validated as positive on start (safe default like the login backoff). The count check and insert run under a lock on the owning user row and commit in-tier (`Users.CreatePlayer`), so two concurrent creations for the same account can't both slip past the cap.
- **Per-player names** validated by a new `PlayerName` domain rule: trimmed, non-blank, ≤ the existing 20-char `Player.Name` limit, no control characters. Names need not be globally unique (no competitive surface).
- Regenerated the TypeScript API client and documented the flow in `docs/backend.md`.

Delete/retire of a character is out of scope (deferred, per the issue).

## How it addresses the issue

| Issue requirement | Implementation |
| --- | --- |
| `CreatePlayer(name)` via `NewPlayerFactory`, attached to the account | `AccountService.CreatePlayer` → `IUsers.CreatePlayer` |
| Config-bound per-account cap (default 6), server-enforced | `PlayerCreationOptions` + atomic cap check in `Users.CreatePlayer` |
| Per-player user-supplied names, length/charset validated | `PlayerName.TryNormalize` (domain) |

## Design notes

- The cap is enforced atomically in the data tier (row lock + count + insert in one transaction, mirroring the existing admin-lockout pattern) rather than a naive count-then-insert, so the anti-cheat cap holds against an account firing parallel `CreatePlayer` requests.
- The data tier reports a `CreatePlayerOutcome` (Success / CapReached / UserNotFound); the application layer adds `InvalidName` and exposes a `CreatePlayerStatus`. The two enums are deliberately named distinctly (following `LoginStatus` vs `SetUserRolesStatus`) to avoid a cross-layer name collision.

## Test plan

- [x] `PlayerNameTests` — unit coverage of valid names, trimming, blank/too-long/control-character rejection, boundary at 20 chars.
- [x] `AccountServiceIntegrationTests` — create attaches to account & builds the starter graph, additional character doesn't disturb existing ones, invalid name rejected (creates nothing), at-cap rejected, and a concurrent at-cap-boundary race never exceeds the cap.
- [x] `LoginControllerTests` — authenticated valid name (200 + persisted), unauthenticated (401), invalid name (400, creates nothing), at-configured-cap (400) verifying the default-6 binding end-to-end.
- [x] Full backend suite green. While verifying, found 3 **pre-existing** `RequestLogging*` failures unrelated to this feature (introduced by `d4ed2e8`, which changed the log message to `"Request Start {Path}"` without updating the exact-match assertions). Bundled the trivial test-only fix here since it was blocking `test-backend` CI — see the second commit; closes #1095.
- [x] Frontend `npm run lint` (eslint + svelte-check) clean after regenerating the client.

Closes #1069


Game.Abstractions/DataAccess/IUsers.cs
```diff
@@ -51,6 +85,16 @@ public interface IUsers
         /// </summary>
         Task<bool> CreateAccount(NewAccount account, NewPlayer player);
 
+        /// <summary>
+        /// Creates an additional character on an existing account from the <paramref name="player"/>
+        /// blueprint, enforcing the per-account cap (<paramref name="maxPlayersPerAccount"/>) as anti-cheat.
+        /// The count check and insert run under a lock on the owning user row and commit in-tier, so two
+        /// concurrent creations for the same account can't both slip past the cap. Returns
+        /// <see cref="CreatePlayerStatus.CapReached"/> when the account is already at the cap and
+        /// <see cref="CreatePlayerStatus.UserNotFound"/> when no active account matches.
```
ginderjeremiah: **Should-fix:** these `<see cref="CreatePlayerStatus..."/>` references won't resolve from `Game.Abstractions` — `CreatePlayerStatus` lives in `Game.Application.Services`, and this project only references `Game.Core`. This method also returns `CreatePlayerResult`/`CreatePlayerOutcome`, not `CreatePlayerStatus`, so the references are conceptually wrong too. Point them at `CreatePlayerOutcome.CapReached` / `CreatePlayerOutcome.UserNotFound`. (The `<c>CreatePlayerStatus</c>` mention up at line 45 is fine — that's plain text contrasting the two layers, not a cref.)

---

Game.DataAccess/Repositories/Users.cs
```diff
@@ -86,6 +86,49 @@ public async Task<bool> CreateAccount(NewAccount account, NewPlayer player)
...
+            await transaction.CommitAsync();
+
+            return CreatePlayerResult.Created(new PlayerSummary
+            {
+                Id = entity.Id,
+                Name = entity.Name,
+                Level = entity.Level,
+                CurrentZoneId = entity.CurrentZoneId,
+                LastActivity = entity.LastActivity,
+            });
```
ginderjeremiah: **Nit (DRY):** this is now the third hand-written copy of the `PlayerSummary` projection (`ToAdminUser` and `GetPlayerSummaries` already carry "keep the two in step" comments). The other two are EF `Expression`s that must translate to SQL, but this one builds from a tracked entity in memory and could use a small shared `PlayerMapper.ToSummary(EntityPlayer)` helper — consistent with mapping living in `Game.DataAccess/Mapping`, and it removes one more place to forget a new field. Worth a follow-up at least.

---

ginderjeremiah (commented):
## Review: Character creation — per-account cap & per-player names (#1069)

**Verdict:** Approve (posted as a comment since I authored the PR) — solves #1069 cleanly with sound layering and strong tests; the findings below are non-blocking.

The change does exactly what the issue asks: a `CreatePlayer(name)` flow that builds via `NewPlayerFactory`, attaches to the account, enforces a config-bound cap server-side, and gives each character its own validated name. Layering is right — name validation is a `Game.Core` domain rule (`PlayerName`), `AccountService` only orchestrates, the cap + persistence live in the data tier, and the controller just does HTTP/auth.

### Should-fix
- `Game.Abstractions/DataAccess/IUsers.cs:93-94` — broken `<see cref="CreatePlayerStatus..."/>` references (wrong layer + wrong enum). Inline comment has the fix.

### Nits
- `Game.DataAccess/Repositories/Users.cs:122-129` — third copy of the `PlayerSummary` projection; consider a shared `PlayerMapper.ToSummary`. Inline comment.
- `Game.DataAccess/Repositories/Users.cs:89-120` — unlike `CreateAccount`, the new `CreatePlayer` doesn't clear the change tracker on an unexpected `SaveChangesAsync` failure. There's no *expected* unique violation for a player insert, so this is low-risk, but if the save throws, the transaction rolls back at the DB while the `Added` entity stays tracked and the per-request unit-of-work would re-attempt it. A `try/catch` that clears the tracker (or a comment noting why it's unnecessary) would mirror the sibling method.

### Test coverage
Strong and aligned with the project's rules:
- `PlayerNameTests` — thorough domain unit coverage (valid, trim, blank, too-long, boundary at 20, control chars).
- `AccountServiceIntegrationTests` — attach-to-account + starter graph, additional char doesn't disturb existing, invalid name creates nothing, at-cap rejected, and a concurrent at-cap-boundary race that asserts the cap holds (good — that's the whole point of the row lock).
- `LoginControllerTests` — 200/401/400 paths and the default-6 cap wired end-to-end.

Cap enforcement is covered by integration rather than pure unit tests, which is correct here since the guarantee lives in the DB transaction/row lock — no logic-bearing class took on an out-of-process dependency unnecessarily.

### What's good
- The atomic `SELECT … FOR UPDATE` + count + insert under one transaction is the right call over a naive count-then-insert, and it mirrors the existing account-creation pattern. The race test proves it.
- Distinct `CreatePlayerOutcome` (data tier) vs `CreatePlayerStatus` (application) enums avoid a cross-layer name collision and keep name-validation failures out of the data tier.
- Config-bound cap with a safe positive default and fail-fast `ValidateOnStart`, consistent with the login-backoff options.
- `docs/backend.md` updated to describe the new flow.

Note: CI was still running at review time, and there are no auto-merge-eligible conditions here (open should-fix). I'll leave merging until the doc-cref fix lands and CI is green.

---

PR #1096, opened 2026-06-21
Add character-select &amp; create-character screen (#1070)

### LINKED ISSUE #1070: Multiple players per account: player-select & create-character screen (frontend)

ginderjeremiah:
## Summary

Implements the frontend **character-select & create-character screen** (#1070), the third sub-issue of the multiple-players-per-account spike (#922). Both backend prerequisites — the player-selection seam (#1068) and character creation (#1069) — are merged, so this wires the new screen onto them.

A new `/select` route sits between the login page and the loading screen:

1. `Login` returns the account's `playerSummaries` (it already no longer auto-binds the first character). The login page now hands those off and navigates to `/select` instead of auto-selecting.
2. The select screen lists the account's characters, offers a **create-character** affordance, and on selection calls `SelectPlayer` — binding the session and rotating the token to carry the chosen `playerId` — then runs the per-player active-session takeover check before entering the game.

## How it addresses the issue

- **New screen between login and loading** → `src/routes/select/+page.svelte` (`/select` route).
- **Wire `SelectPlayer`** → on card click; stores the rotated token pair and seeds `playerManager` from the loaded player before handing off to `/loading`.
- **Create-character affordance, respecting the per-account cap** → `CreatePlayer`, appending the returned summary. Name validation mirrors the backend `PlayerName` rule (trimmed 1–20 chars, no control chars) for fast feedback; the backend re-validates and enforces the config-bound cap as anti-cheat, and the screen surfaces its cap-reached / invalid-name errors.
- **Relocate the takeover confirmation** → `confirmSessionTakeover` moved out of the login submit path into the select flow (it's a per-player presence check, post-selection).
- **Re-initialize `playerManager` from the selected player** → in the select page's `enterWorld`.
- **Docs** → `docs/frontend-screens.md` gains a "Character-select screen" section.

## Design notes

- **Summaries are handed off, not re-fetched.** There's no "list my players" endpoint (the list is only returned by the credentialed `Login`), so the login page stashes the summaries in a one-shot, non-reactive module store and the select page takes them on mount.
- **`/select` is a boot-only route.** A pre-selection token (no `playerId` claim) makes `Login/Status` fail, so a refresh on `/select` cleanly falls back to login via the boot gate — no backend change needed. Added `/select` to `BOOT_ONLY_ROUTES` so a fully-restored session never rests there.
- **Injected-deps view-model.** `PlayerSelectView` owns list / busy / create-form state; the API calls, token storage, takeover confirm, and world entry are injected for unit-testability — the same reactive-view-model split the other screens use.
- Zone names aren't available pre-load, so cards show level + a relative "last played" age rather than the current zone.

## Test plan

- [x] `npm run lint` (eslint + svelte-check) — clean, 0 errors.
- [x] `npm run test:unit` (vitest) — **2346 passed** (223 files), including new suites: `player-select-view`, `select-page`, `player-name`, `last-played`, the rewritten `login-submit`, and the extended `boot-redirect`.
- [x] Updated e2e helpers to route through the select screen; added `e2e-tests/select.test.ts` (signup → select, create a second character, enter). **Note:** the live backend isn't running in this routine's container, so the Playwright suite wasn't executed locally — it runs in CI.

## Note for review

The per-account cap value isn't exposed to the frontend, so creation can't be proactively disabled at the cap — the screen surfaces the backend's cap-reached error instead. If proactive disabling is wanted, exposing the cap (a small backend addition) would be a reasonable follow-up.

Closes #1070

https://claude.ai/code/session_01D3zDaR28a2nTp68dAoTHhQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01D3zDaR28a2nTp68dAoTHhQ)_


UI/src/routes/+page.svelte
```diff
@@ -198,36 +191,18 @@ const handleSubmit = async () => {
...
-		if (!(await confirmSessionTakeover())) {
-			submitting = false;
-			return;
-		}
-		// Fire-and-forget: report this device's capabilities now that we're authenticated.
-		void reportDeviceInfo();
-		enterWorld(selected.data.player);
+		success = true;
+		playerSelectHandoff.set(summaries);
+		goto(resolve('/select'));
```
ginderjeremiah, line 207: **Blocking** — this navigation to `/select` is bounced straight back to `/` by the root layout's safety-net guard, which is why the entire e2e suite is red (`toHaveURL('/select')` → "24 × unexpected value `http://localhost:4173/`").

The guard in `UI/src/routes/+layout.svelte` (not in this diff) is:

```js
$effect(() => {
	if (booted && !playerManager.name && page.url.pathname !== '/' && !redirecting) {
		redirecting = true;
		goto(resolve('/'));
	}
});
```

`/select` is authenticated but intentionally player-less — `playerManager` isn't initialized until `enterWorld` runs *after* selection. So the moment we land on `/select`, `booted && !playerManager.name && pathname !== '/'` is true and the effect redirects to `/`. The old flow didn't trip this because it called `playerManager.initialize(...)` *before* navigating to `/loading`.

This needs the safety-net effect (and ideally that "any non-`/` route implies a live player" invariant) to exempt `/select` the same way `BOOT_ONLY_ROUTES` does for the boot gate — e.g. skip the guard on the pre-selection boot routes. The unit suites pass only because they mock the layout/guard away; please add an e2e (or a layout test) that actually covers landing on `/select` so this can't regress silently.

---

ginderjeremiah (commented):
## Review: Character-select & create-character screen (#1070)

**Verdict:** Request changes — the feature's own critical path is broken; the entire Playwright suite (all 3 browsers) fails deterministically because landing on `/select` is bounced back to the login page. _(Posted as a comment since I'm the PR author and GitHub won't accept a gating verdict on one's own PR.)_

The design, layering, and unit-test coverage here are genuinely strong — this is one root-cause guard interaction away from being mergeable.

### Blocking

- **`/select` is bounced to `/` by the layout safety-net guard — the whole e2e suite is red.** (`UI/src/routes/+page.svelte:205`, root cause in `UI/src/routes/+layout.svelte`, not in this diff.) The boot gate's reactive safety-net effect redirects to `/` whenever `booted && !playerManager.name && pathname !== '/'`. The new `/select` route is authenticated but deliberately player-less (the player isn't initialized until `enterWorld`, *after* selection), so the effect fires the instant we navigate there and sends the user back to login. CI confirms it: every e2e test (admin, navigation, game-flow, challenge-boss, and the new `select.test.ts`) fails on `await expect(page).toHaveURL('/select')` with "24 × unexpected value `http://localhost:4173/`". The boot gate's `onMount` path was correctly taught about `/select` via `BOOT_ONLY_ROUTES`, but the *reactive* guard at the bottom of the layout wasn't — it still assumes any non-`/` route implies a live player. The fix is to exempt the pre-selection boot routes from that guard (the same set the boot gate already special-cases). See the inline comment for the exact effect.

### Should-fix

- **A real layout/e2e test must cover landing on `/select`.** Every new unit suite passes because they mock the layout and its guard away (`select-page.test.ts` mocks `$lib/engine`, navigation, etc.), so the one interaction that actually breaks the feature is entirely untested. The Playwright suite *would* have caught this, but per the PR description it wasn't run locally ("the live backend isn't running in this routine's container"). Beyond fixing the guard, please add coverage that exercises the real boot-gate/guard with an empty `playerManager` on `/select` so this can't silently regress.

### Nits

- `UI/src/routes/select/player-select-view.svelte.ts:64-67` — the comment says a declined takeover leaves "the freshly-bound session for the user to retry or pick another character." In production the injected `confirmTakeover` is `confirmSessionTakeover`, which calls `logout()` (full-page redirect to `/`) on decline — so the user is sent back to login, not left on `/select` to retry. The view-model's *local* behavior matches the comment, but the wired behavior doesn't; worth a word so a future reader isn't misled.
- `UI/src/routes/+page.svelte:203` — `success = true` is now effectively dead: navigation to `/select` happens immediately, so the success status line is never observed (the old 600ms delay that surfaced it is gone). Either drop it or keep it intentionally; harmless either way since `submitting` stays true to hold the form disabled through the nav.

### Test coverage

- Unit coverage of the new pure/logic units is thorough and well-targeted: `player-select-view` (select success/error/declined-takeover/busy-guard, create success/invalid/cap-reached/trim), `player-name` (length, control-char, trimming), `last-played` (all buckets, future clamp, unparseable), the rewritten `login-submit`, and the extended `boot-redirect`. The classical, double-light style matches the project's guidance.
- Gap noted above: nothing exercises the actual layout guard against a player-less `/select`, which is exactly the interaction that's broken. That's the missing test, not the unit ones.

### What's good

- Clean reactive-view-model split with injected deps — orchestration is unit-testable without real I/O, consistent with the other screens and the established raw-`$state`-in-`.svelte.ts` pattern (statify is reserved for the heavier domain objects).
- All styling flows through theme tokens / `color-mix` over semantic vars — no hard-coded colors; real `<button>` elements with `:focus-visible` rings throughout, in line with the a11y rules in `frontend.md`.
- Client-side `player-name` validation is correctly framed as fast-feedback-only that mirrors the backend `PlayerName` rule (incl. a faithful `char.IsControl` port), with the backend re-validating as anti-cheat. Docs (`frontend-screens.md`) and the boot-only-route handling are properly updated. The one-shot non-reactive handoff is a reasonable answer to there being no "list my players" endpoint.

---

ginderjeremiah:
Fixed in `958382c`. Same root cause you identified — I'd traced it from the e2e failures in parallel.

**Blocking:** Extracted a pure `shouldReturnToLogin(hasPlayer, pathname)` predicate that exempts the pre-player auth routes (`/` and `/select`) and wired it into the layout's safety-net `$effect`. `/loading` deliberately stays guarded (a character is selected and `playerManager` initialized before reaching it), so it's a distinct, narrower set than `BOOT_ONLY_ROUTES`.

**Missing test:** Added a layout component test that drives the real boot-gate/guard with an empty `playerManager` on `/select` — asserts it is *not* bounced, and that the guard is still live (losing the player on `/game` still returns to login). Plus predicate unit tests in `boot-redirect.test.ts`.

**Nits:** Corrected the view-model comment about a declined takeover (the wired `confirmSessionTakeover` logs out / redirects to login rather than resting on `/select`), and documented the intentional brief success status on the login → select transition.

Full unit suite green (2350 passed); lint/svelte-check clean. The Playwright suite will re-run on this push.

---

Issue #1072, opened 2026-06-21
Multiple players per account: in-app character switcher (frontend)


ginderjeremiah:
Sub-issue of #922. See the spike design doc: `docs/spikes/922-multiple-players-per-account.md`. Depends on #1071 (lossless switch backend) and #1070 (select screen).

## Goal

Switch characters from within the game without re-logging-in.

## What to build

- A switch affordance in-game that runs the switch flow: tear down the current socket/session, trigger the backend lossless switch (#1071), show the entered character's catch-up summary (reuse the #879 welcome-back summary UI), and re-initialize the game for the new character.
- Reuse the player-select UI (#1070) for picking the target character.

## Notes

The departed character's idle progress is credited server-side (#1071); the summary surfaces what it earned while parked.

## Docs

Update `docs/frontend-screens.md`.

PR #1097, opened 2026-06-21
Add lossless character switch backend (#1071)

### LINKED ISSUE #1071: Multiple players per account: lossless character switch (backend)
Sub-issue of #922. See the spike design doc: `docs/spikes/922-multiple-players-per-account.md`. Depends on #1068 (auth seam) and **#1042** (offline reward application/orchestration — the elapsed-time simulator).

## Goal

Switching the active character must not lose the departed character's idle progress.

## Why

The client-driven idle loop only runs at full rate in the foreground, and switching tears down the current session/socket. To make a switch lossless, credit the departed character via the same elapsed-time simulation used for offline rewards (#879 / #1042), anchored on `Player.LastActivity` (#1039).

## What to build

- On switch: resolve any in-flight battle, run the offline/elapsed-time simulation for the **departed** character from its `LastActivity`, apply the rewards, persist, then bind + load the newly selected character (re-using the `SelectPlayer` path from #1068).
- **Drop the >=5-minute floor for a deliberate switch** — the 5-min floor in #879 is for login-time offline rewards; a deliberate switch should credit any elapsed time.

## Notes

- This makes one-live-at-a-time **outcome-equivalent** to parallel progression (the departed character is fully credited for every second until you return) — the core rationale in the spike.
- Reuses the #1042 simulator; **no new partial-battle simulation** — the sim runs whole battles and the new character resumes with a fresh one.

## Tests

Unit/integration-test that a switch credits the departed character's elapsed time and cleanly binds the new one.

ginderjeremiah:
## Summary

Implements the **backend lossless character switch** (#1071), the fourth sub-issue of the multiple-players-per-account spike (#922). Both prerequisites are merged — the player-selection seam (#1068) and the offline-rewards orchestration/simulator (#1042) — so this layers the switch on top of them.

A logged-in account can change its live character without re-logging in. Switching **credits the departed character** for its elapsed idle time so it loses no progress, making one-live-at-a-time outcome-equivalent to parallel progression (the spike's core rationale).

## How it works

New `SwitchPlayer(playerId, refreshToken)` HTTP flow on `LoginController`:

1. **Validate the target up front** (`AccountService.OwnsPlayer`, read-only, no token rotation) so a rejected/tampering switch never mutates the departed character.
2. **Credit the departed character** — the one the caller's token currently binds — via the offline-rewards simulator: resolve its in-flight battle, replay the elapsed window, apply the rewards, re-anchor `LastActivity`, persist.
3. **Bind + load the new character** through the existing `SelectPlayer` path (re-validates ownership, rotates the token to carry the new `player_id` claim), then establish the session.

The entered character's own welcome-back catch-up runs through the normal offline-rewards path on reconnect (unchanged), so this PR only adds the departed-character settle.

### Key pieces

- **`BattleService.SimulateSwitchProgress`** — the same elapsed-time replay as the login welcome-back path (`SimulateOfflineProgress`) but with the **≥5-minute floor dropped**, since a deliberate switch should credit *any* elapsed time (per the issue). The floor is the only difference, so both now share one private `SimulateProgress(..., minimumAway, ...)` body — they can't drift.
- **`AccountService.OwnsPlayer`** — read-only ownership check that consumes no token.
- Regenerated the TypeScript client map (the new endpoint reuses the existing `SelectPlayerRequest`/`SelectPlayerResult`, so only the route map changed).

## Design notes

- **Why credit over HTTP, off the socket.** The switch tears the game socket down first (frontend #1072), so there's no concurrent battle loop on the departed character — the player-state write is safe over HTTP, mirroring how login binds before any socket exists. This contract (socket down first) is documented in the method/endpoint comments.
- **In-flight battle isn't lost.** `PlayerState` (the in-flight battle) is cache-only and would be discarded when the session is rebound to the new character; crediting first settles it. The departed character's durable rewards persist via the write-behind player cache (keyed by playerId), independent of the session rebind.
- **Ordering.** Ownership is validated before the credit so a doomed switch is a clean no-op on the departed character. The credit deliberately drops the floor, which means it runs the full sim even for a tiny window — gating on ownership first also avoids running it on a tampering path.

## Test plan

- [x] `dotnet build` — clean (0 warnings).
- [x] `Game.Application.Tests` — **445 passed**, incl. new `SimulateSwitchProgress` integration tests (sub-threshold away still credits + re-anchors; stale in-flight battle resolved).
- [x] `Game.Api.Tests` — **576 passed**, incl. new `LoginControllerTests`: switch credits the departed character (it levels from the simulated window) and binds the target with a rotated token; unowned target rejected; unauthenticated → 401.
- [x] `npm run lint` (eslint + svelte-check) — clean, 0 errors.

## Follow-up

This is the backend half; the in-game switcher UI (tear down socket → call `SwitchPlayer` → reconnect and show the entered character's welcome-back summary) is the already-tracked frontend sub-issue **#1072**, which depends on this and #1070.

Closes #1071

https://claude.ai/code/session_013CLt983K9YBznATWbtDFZ6

---
_Generated by [Claude Code](https://claude.ai/code/session_013CLt983K9YBznATWbtDFZ6)_


Game.Api/Controllers/LoginController.cs
```diff
@@ -91,6 +93,74 @@ public async Task<ApiResponse<SelectPlayerResult>> SelectPlayer([FromBody] Selec
...
+            // Validate the switch target up front (read-only, no token rotation) so a rejected switch never
+            // credits or mutates the departed character. SelectPlayer below re-validates as the authority.
+            if (!await _accountService.OwnsPlayer(_sessionService.UserId, request.PlayerId))
+            {
+                return ApiResponse.Error(SelectPlayerErrorMessage(SelectPlayerStatus.NotOwned), SelectPlayerErrorCategory(SelectPlayerStatus.NotOwned));
+            }
+
+            // Settle the departed character before binding the new one, so its idle progress and in-flight
+            // battle are credited rather than discarded when the session is rebound below.
+            await CreditDepartedCharacter(request.PlayerId, HttpContext.RequestAborted);
```
ginderjeremiah, line 124: Nit / consider: the credit runs before `SelectPlayer` validates the refresh token, so the "doomed switch is a clean no-op on the departed character" rationale only holds for the *unowned-target* case (gated above). A switch with a valid access token but an invalid/foreign `RefreshToken` still credits and re-anchors the departed character, then fails at `SelectPlayer` with `InvalidToken`. It's benign in practice — re-anchoring `LastActivity` makes a subsequent retry near-idempotent, and the credit is the character's legitimate progress — but it's a small gap between the stated design and the behavior. Either validate the token before crediting, or tighten the comment/design note to scope the "no-op" guarantee to the ownership check.

---

docs/game-design.md
```diff
@@ -145,6 +145,10 @@ When a player returns after being away, they are credited for the progress their
...
 - **A welcome-back gate** shows the player what they earned (away duration, loop mode, battles won/lost/drawn, exp, levels & stat points gained, and the challenges completed with their unlocks) before the live idle loop starts, so simulated and live battles never overlap. A sub-threshold or empty return skips the gate. Re-anchoring `LastActivity` on claim makes an immediate re-claim a no-op, so rewards can't be double-collected by reconnecting.
 
+## Multiple characters per account
+
+An account can own several characters, but **only one is live at a time** (spike [#922](./spikes/922-multiple-players-per-account.md)). Switching to another character is **lossless**: the character being left behind is credited for all its elapsed idle time via the same offline-rewards simulator (with the 5-minute floor dropped, since a deliberate switch should lose nothing), and the entered character runs its own welcome-back catch-up. A parked character therefore loses no progress while you play another — one-at-a-time is outcome-equivalent to running them in parallel, just without rendering two battles on screen at once.
```
ginderjeremiah: Nit (doc precision): "credited for all its elapsed idle time" / "loses no progress" / "outcome-equivalent to running them in parallel" slightly overstates it — the shared simulator still caps each window at `MaximumOfflineSimulation` (10h). Since `LastActivity` re-anchors on each switch, nothing is lost as long as a character is switched to/returned to within the cap, but a single park longer than 10h does drop time beyond the cap. Same caveat already applies to the login welcome-back path, so this isn't introduced here — just worth a brief "up to the offline simulation cap" qualifier so the claim stays accurate.

---

ginderjeremiah (commented):
## Review: Backend lossless character switch (#1071, sub-issue of spike #922)

**Verdict: Approve** *(posted as a `COMMENT` because GitHub won't let me gate my own PR)* — solves the issue cleanly, layering correctly on the #1068 selection seam and the #1042 offline simulator. Only minor nits below; none blocking.

This is a well-scoped, well-documented change. The standout is the refactor of `SimulateOfflineProgress`/`SimulateSwitchProgress` into a shared `SimulateProgress(..., minimumAway, ...)` body — the away floor is genuinely the *only* difference, so collapsing them to one path is exactly right and removes any chance of the two drifting (which the battle-consistency rules care about). Ownership is validated up-front via the new read-only `OwnsPlayer` before any mutation, the departed-character credit is settled before the session rebind discards its cache-only state, and the socket-down-first contract is documented at every relevant seam.

DDD layering holds: domain replay stays in `BattleService`, the ownership check is a thin read in `AccountService`, and the controller only orchestrates session-binding (a presentation concern, consistent with the existing `SelectPlayer`/`Status` methods). Docs (`backend.md`, `game-design.md`) and the regenerated TS client map are all updated.

### Should-fix / Nits
- **`LoginController.cs:122`** — the departed-character credit runs before `SelectPlayer` validates the refresh token, so the "doomed switch is a clean no-op" rationale only covers the unowned-target case, not an invalid/foreign refresh token. Benign (re-anchoring makes a retry near-idempotent), but worth scoping the comment or validating the token earlier.
- **`game-design.md:150`** — "loses no progress" / "outcome-equivalent to parallel" slightly overstates it given the simulator's 10h `MaximumOfflineSimulation` cap. A single park >10h still drops the excess. A brief "up to the offline simulation cap" qualifier keeps the claim precise (same caveat already applies to the login path).

### Test coverage
Solid. Api integration covers 401 (unauthenticated), unowned-target rejection, and the full happy path (departed credited → levels up, target bound, token rotated, session store + Status reflect the target). Application integration covers the dropped-floor sub-threshold credit + re-anchor and the stale in-flight battle being resolved. Two nice-to-haves, not blockers:
- The unowned-rejection test asserts `BadRequest` but doesn't assert the departed character was left *uncredited* — that's the actual anti-cheat guarantee being claimed.
- The `CreditDepartedCharacter` no-op branches (same-character switch, pre-selection token, `departed is null`) aren't exercised at the controller level.

### Note
No frontend battle-logic mirror is needed here — this adds an orchestration entry point that reuses the existing simulator rather than changing battle math; the UI half is the already-tracked #1072.

---

PR #1099, opened 2026-06-21
Frontend minor hardening: tooltip listener lifetime, frozen static modifiers, shared outcome classifier

### LINKED ISSUE #1082: Frontend minor hardening: tooltip listener lifetime, shared static-modifier template, duplicated outcome branching
Low-severity hardening/cleanup notes from a general frontend code-health review. None is an active defect; bundled to avoid issue noise. Pick off opportunistically.

### 1. Module-level tooltip listeners are registered but never removed
`UI/src/stores/tooltip.svelte.ts:47-51` registers two capture-phase `document.addEventListener('keydown'/'pointerdown', …)` at module evaluation with no removal. Benign in production (true module singleton, app-lifetime), but it leaks a listener per module re-evaluation under HMR/tests and is the "added at import, never cleaned" pattern the codebase otherwise avoids (cf. the careful keepalive/interval teardown in `api-socket.ts`). Either add a short comment justifying the singleton lifetime, or scope the registration behind a once-guarded init.

### 2. `STATIC_ATTRIBUTE_MODIFIERS` references are shared across every `BattleAttributes`
`UI/src/lib/battle/battle-attributes.ts:70` builds `#modifiers = [...base, ...STATIC_ATTRIBUTE_MODIFIERS]`. The spread copies the array but not the modifier objects, so every battler shares the same static-modifier object references; `removeModifier` (`:82-90`) matches by reference. Safe today (effect modifiers are freshly allocated, and nothing removes a static modifier), but a sharp edge: any future code that removes/mutates a static modifier on one battler would corrupt the shared template for all. Make the invariant explicit — `Object.freeze`/`readonly` the template, or clone per instance.

### 3. Duplicated outcome-branching in the battle engine
`UI/src/lib/engine/battle/battle-engine.ts:327-360`: `damageLogOutcome` and `combatFloatEvent` encode the same `byPlayer/crit/dodged/blocked → outcome` decision twice (the comments acknowledge they "mirror" each other). Small and well-commented, but the kind of parallel mapping that drifts. Derive both the log `LogOutcome` and the float `kind` from one shared classifier.

ginderjeremiah:
## Summary

Addresses the three low-severity frontend code-health items bundled in #1082.

### 1. Module-level tooltip listeners registered but never removed
`stores/tooltip.svelte.ts` registered two capture-phase `document` listeners at module evaluation with no teardown — benign in production (true app-lifetime singleton) but leaking a pair per re-evaluation under HMR/repeated test imports. The modality state and its listeners now hang off a global singleton keyed by a stable `Symbol.for`, so a re-evaluation reuses the existing registration instead of stacking a second, unremovable pair. The listeners are still app-lifetime by design (now documented as such); the shared-state object lets them keep updating the live modality even across a re-eval, so `focusAnchor` behaviour is unchanged.

### 2. `STATIC_ATTRIBUTE_MODIFIERS` shared across every `BattleAttributes`
Each `BattleAttributes` spreads these shared modifier object references into its own list, and `removeModifier` matches by reference — safe today, but a future mutation of a static modifier on one battler would corrupt the shared template for all. The table and its entries are now `Object.freeze`d, enforcing the immutability invariant at runtime as a complement to the union's existing `readonly` fields. Per-instance cloning was rejected — it would defeat the shared-template optimization for no benefit over a freeze.

### 3. Duplicated outcome-branching in the battle engine
`battle-engine.ts` encoded the same `byPlayer/crit/dodged/blocked → outcome` decision twice (`damageLogOutcome` for the log, `combatFloatEvent` for the float). `combatFloatEvent` now derives its event from the resolved `LogOutcome`, so the single `damageLogOutcome` classifier is the sole source of that decision and the two surfaces can't drift. Pure FE refactor (these surfaces are frontend-only — the headless simulator emits neither), no battle-parity impact.

## Test plan
- [x] Added a freeze-invariant assertion to `attribute-modifier.test.ts`.
- [x] Existing `battle-engine` crit/dodge/block log + combat-float tests cover the refactor (behaviour preserved).
- [x] `npm run test:unit:ci` — 224 files / 2357 tests pass.

## Note for review
Item 2's mutation vector was already blocked at compile time by the union's `readonly` fields; the freeze adds the runtime guarantee the issue asked for (it does not change the existing type-level contract). Flagging in case you'd prefer to rely on the type-level guard alone.

Closes #1082

---
_Generated by [Claude Code](https://claude.ai/code/session_01HMhmQJUTucAoNZi7q1uKYj)_

### DIFF
--- a/UI/src/lib/battle/attribute-modifier.ts
+++ b/UI/src/lib/battle/attribute-modifier.ts
@@ -49,5 +49,12 @@ export type AttributeModifier = DerivedAttributeModifier | BaseAttributeModifier
  *  of, generated from `Game.Core.Attributes.Modifiers.StaticAttributeModifiers.All`
  *  and kept in the same order the backend applies them. The explicit type both
  *  documents the contract and verifies, at compile time, that the generated table
- *  conforms to the {@link AttributeModifier} union. */
-export const STATIC_ATTRIBUTE_MODIFIERS: readonly AttributeModifier[] = GENERATED_STATIC_ATTRIBUTE_MODIFIERS;
+ *  conforms to the {@link AttributeModifier} union.
+ *
+ *  Each `BattleAttributes` spreads these into its own modifier list, so every battler
+ *  shares the same modifier object references. The objects (and the table) are frozen
+ *  to enforce that shared-template invariant at runtime — a complement to the union's
+ *  `readonly` fields — so a stray mutation on one battler can't corrupt every other. */
+export const STATIC_ATTRIBUTE_MODIFIERS: readonly AttributeModifier[] = Object.freeze(
+	GENERATED_STATIC_ATTRIBUTE_MODIFIERS.map((modifier) => Object.freeze(modifier))
+);
--- a/UI/src/lib/engine/battle/battle-engine.ts
+++ b/UI/src/lib/engine/battle/battle-engine.ts
@@ -212,7 +212,7 @@ export class BattleEngine {
 			)) {
 				const outcome = damageLogOutcome(byPlayer, crit, dodged, blocked);
 				logMessage(ELogType.Damage, this.damageLogMessage(skill.name, damage, outcome), outcome);
-				notifyCombatFloat(combatFloatEvent(byPlayer, crit, dodged, blocked, damage));
+				notifyCombatFloat(combatFloatEvent(outcome, damage));
 			}
 			this.logEffectApplications();
 			this.accumulateEffectDamage(timeDelta);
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Frontend minor hardening (#1082) — tooltip listener lifetime, frozen static modifiers, shared outcome classifier

**Verdict:** Approve _(submitted as a comment since I authored the PR; intent is approve)_ — all three low-severity items from #1082 are addressed faithfully, with no behavior change and adequate coverage.

### Findings

None blocking, should-fix, or nit. Verified each item against the issue's intent:

1. **Tooltip listeners** — the modality state and its capture-phase listeners now hang off a `Symbol.for`-keyed `globalThis` singleton, guarded so a re-eval reuses the existing registration instead of stacking a second unremovable pair. Old listeners and the new `focusAnchor` both reference the same shared `FocusModality` object, so the live modality keeps updating across a re-eval and `focusAnchor` behavior is unchanged. The singleton lifetime is now documented inline, satisfying the issue's "comment or once-guard" ask (it does both). SSR and client don't share `globalThis`, so there's no cross-context leakage.
2. **Frozen static modifiers** — table and entries are `Object.freeze`d. The entries are flat scalar records, so the shallow freeze fully enforces the shared-template invariant `removeModifier`'s reference-matching relies on; it's a clean runtime complement to the union's `readonly` fields. Declining per-instance cloning was the right call.
3. **Shared outcome classifier** — `combatFloatEvent` now derives from the resolved `LogOutcome` rather than re-branching the raw flags, so `damageLogOutcome` is the single source of that decision and the log/float surfaces can't drift. Worth noting: the `switch` has no `default` and `LogOutcome` is exactly its five cases, so a future variant becomes a compile error — a nice exhaustiveness guard. Frontend-only surfaces, no parity impact, as stated.

### Test coverage

Appropriate for the scope. A freeze-invariant assertion was added for item 2; item 3 is a pure refactor covered by the existing crit/dodge/block log + combat-float tests; item 1 is a structural change with no observable behavior change, so leaning on existing `focusAnchor` coverage is reasonable.

### What's good

Tight, well-scoped tech-debt cleanup — each change makes an implicit invariant explicit (runtime freeze, single classifier, once-guarded singleton) without over-engineering, and the reasoning is documented where a future reader will need it.


---

PR #1105, opened 2026-06-21
Show actual event time in combat log (#1104)

### LINKED ISSUE #1104: Show correct event time in logs
The logs currently do not reflect the actual time that logged events take place.

ginderjeremiah:
## Summary

Fixes #1104 — the combat-log panel's per-row time had no relation to when the event actually occurred.

Each `LogRow` rendered `formatLogTime(log.id)`, and `formatLogTime` treated the entry's **monotonic id** as a clock (`id / 60` "seconds"). Since the id increments once per logged event — not once per second — the displayed time was a fabricated counter, not the real event time.

## How it addresses the issue

- Added a `timestamp` (epoch ms) field to `LogMessage`, set to `Date.now()` in `addLog` — the single funnel every real log entry passes through.
- Rewrote `formatLogTime` to format an actual timestamp as local 24-hour **HH:MM:SS**, and `LogRow` now renders `log.timestamp`.
- Widened the time column (38px → 54px) to fit the longer value.
- Updated the Options screen's live-preview sample feed (which builds `LogMessage`s directly) to stamp its synthetic entries the same way.

### Note on the chosen format

I went with **wall-clock time-of-day (HH:MM:SS)** as the most literal reading of "the actual time that logged events take place". Combat events arrive several per second, so second-level granularity is needed (HH:MM would collapse many entries onto one value). If you'd prefer an *elapsed*-since-session clock instead (closer to the old MM:SS shape, but backed by real timestamps), that's a small follow-up — happy to switch.

## Test plan

- [x] `npm run test:unit:ci` — 2354 passed
- Rewrote the `formatLogTime` unit suite for timestamp formatting (timezone-independent: each instant is built from and read back via local-time components).
- Added a `logs` store test asserting each entry is stamped with the wall-clock time it was logged (fake timers).
- Updated `LogPanel` / `log-kind` tests to satisfy the new required field.

This is a frontend-only change; no backend or battle-parity surfaces are touched.

https://claude.ai/code/session_01JMqupUun3WnZ7pFuqy9DF3

---
_Generated by [Claude Code](https://claude.ai/code/session_01JMqupUun3WnZ7pFuqy9DF3)_

### DIFF
--- a/UI/src/components/log-panel/LogRow.svelte
+++ b/UI/src/components/log-panel/LogRow.svelte
@@ -10,7 +10,7 @@
 	style:box-shadow={isLatest ? `inset 2px 0 0 ${kind.color}` : undefined}
 	style:--row-height="{rowHeight}px"
 >
-	<span class="row-time" class:latest={isLatest}>{formatLogTime(log.id)}</span>
+	<span class="row-time" class:latest={isLatest}>{formatLogTime(log.timestamp)}</span>
 
 	<div
 		class="row-chip"
@@ -63,7 +63,7 @@ const rowOpacity = $derived(isLatest ? 1 : Math.max(0.4, 0.9 - index * 0.1));
 	font-family: var(--mono);
 	font-size: 10.5px;
 	letter-spacing: 0.4px;
-	min-width: 38px;
+	min-width: 54px;
 	color: var(--text-muted);
 
 	&.latest {
--- a/UI/src/components/log-panel/log-kind.ts
+++ b/UI/src/components/log-panel/log-kind.ts
@@ -83,10 +83,12 @@ export function logKind(log: LogMessage): LogKind {
 	}
 }
 
-/** Treats the monotonically-increasing log id as an elapsed-time clock. */
-export function formatLogTime(id: number): string {
-	const totalSeconds = Math.floor(id / 60);
-	const minutes = Math.floor(totalSeconds / 60);
-	const seconds = totalSeconds % 60;
-	return `${String(minutes % 100).padStart(2, '0')}:${String(seconds).padStart(2, '0')}`;
+/** Formats an entry's wall-clock timestamp (epoch ms) as local 24-hour HH:MM:SS, so the panel
+ *  shows the actual time each event occurred. */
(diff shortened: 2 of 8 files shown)


ginderjeremiah (commented):
## Review: Show actual event time in combat log (fixes #1104)

**Verdict:** Approve *(posted as a comment because I'm the PR author and GitHub won't accept a gating verdict on one's own PR).*

The root cause is correctly diagnosed: `formatLogTime` was treating the monotonic `id` (`id / 60`) as a clock, so the displayed value was a fabricated counter, not a real time. The fix adds a real `timestamp` (epoch ms) stamped at `Date.now()` in `addLog` — the single funnel every genuine entry passes through (`logMessage` → `addLog`) — and formats it as local 24-hour HH:MM:SS. That's the most literal reading of the issue, and second-level granularity is the right call given multiple combat events per second.

Every `LogMessage` construction site is accounted for (`addLog`, `LivePreview.makeEvent`, and the tests), the field is required so TypeScript/`svelte-check` guards against any future omission, and the change is frontend-only with no battle-parity surface touched.

### Test coverage
Solid and appropriate:
- `formatLogTime` suite fully rewritten and made timezone-independent (each instant built from and read back via local-time components) — good attention to the runner-timezone trap.
- New `logs` store test asserts the wall-clock stamp using fake timers.
- `LogPanel` / `log-kind` tests updated to satisfy the new required field.
- `LogRow` is a non-interactive display component and the formatting logic is tested directly, so no separate component test is needed (consistent with `frontend.md`'s testing guidance).

### Nits
- None blocking. The time-column `min-width` bump (38px → 54px) is a hardcoded layout dimension, but it sits alongside the existing hardcoded px values in that block and isn't a color/theme token, so it's consistent with the surrounding code.

### What's good
- Minimal, well-targeted diff that fixes the actual bug rather than papering over the symptom.
- Clear, right-sized comments on the new field and `formatLogTime`.
- The format-choice tradeoff (wall-clock vs. elapsed-session) is called out explicitly in the PR description — wall-clock time-of-day is the better default here.

Nice work. Will merge once CI goes green.

---

PR #1109, opened 2026-06-21
Hide new-enemy fetch latency by bundling the next battle (#1092)

### LINKED ISSUE #1092: Hide new-enemy fetch latency in the idle loop (overlap with post-battle cooldown)
## Problem

After a victory the idle loop in `EnemyManager.watchIdleStage` (`UI/src/lib/engine/battle/enemy-manager.ts`) runs strictly sequentially:

1. `claimVictory()` — `DefeatEnemy` round trip
2. `startLoading(cooldown)` — wait out the ~5s post-battle cooldown
3. `getNewEnemy()` — `NewEnemy` round trip, only **after** the cooldown completes

Because `getNewEnemy()` is awaited *after* the cooldown, its round-trip latency is pure dead time the player pays on **every** idle cycle (the boss-loss and draw paths follow the same `startLoading → getNewEnemy` order, so they're affected too).

This is the new-enemy analog of the victory-cooldown latency compensation just added in the server-clock victory work: the post-battle cooldown is now latency-independent, but the subsequent enemy fetch is not. Impact scales with RTT and compounds over a long idle farm. (The dev API runs as a Railway container, so non-trivial RTT is the norm, not the exception.)

## Idea

Overlap the new-enemy fetch with the cooldown wait so the enemy is ready the instant the cooldown elapses — the ~5s cooldown comfortably hides typical latency. Two broad approaches:

1. **Client-side overlap** — start `getNewEnemy()` concurrently with `startLoading(cooldown)` and begin the next battle once both resolve.
2. **Server-bundled** — return the next enemy (+ seed) in the `DefeatEnemy` response so there's no separate `NewEnemy` request at all.

## Anti-cheat wrinkle (must be addressed)

`getNewEnemy()` → server `StartBattle` sets `BattleStartTime = server now`. Fetching during the cooldown sets it ~5s before the client actually begins fighting. With the new server-clock victory validation (`BattleService.EndBattleVictory`):

- The elapsed-time anti-cheat (`now - BattleStartTime >= TotalMs - tick`) still passes (more time has elapsed) — fine.
- **But** the *next* victory's cooldown is anchored to `BattleStartTime + TotalMs`; a too-early `BattleStartTime` would shrink or zero that cooldown.

So the server must anchor `BattleStartTime` to when the fight *actually* starts, not when the enemy is provisioned. Cleanest fit: the next fight deterministically begins at `battleCompletedAt + PostBattleCooldown`, so the server can stamp the next `BattleStartTime` to that scheduled time when bundling/prefetching. This needs a short design pass before implementation.

## Acceptance

- New-enemy fetch latency no longer adds dead time between battles in the idle loop (and the boss-loss / draw paths).
- The post-battle cooldown and the elapsed-time anti-cheat stay correct under the prefetched/bundled flow (no shrunken/zeroed cooldowns, no false rejections).
- FE/BE battle parity and existing tests remain green.

## Related

Server-clock-only victory validation + completion-anchored cooldown — recent change to `BattleService.EndBattleVictory` / `DefeatEnemy` (removed the client-supplied timestamp; kept `clientTotalMs` as a diagnostic).

ginderjeremiah:
## Summary

After a victory the idle loop ran strictly sequentially — `claimVictory()` → wait out the ~5s cooldown → **then** `getNewEnemy()` — so the `NewEnemy` round-trip latency was pure dead time paid on *every* idle cycle (and the boss-loss path too). On the Railway dev API that RTT is non-trivial and compounds over a long farm.

This bundles the **next** idle battle into the battle-end response so the client can begin it the instant the cooldown elapses, with no separate request — the cooldown comfortably hides the (already-paid) fetch.

### How it works
- An idle victory (`DefeatEnemy`) and a boss loss (`BattleLost`) now set up the next idle battle immediately via the new `BattleService.PrepareNextIdleBattle` and return it (`nextEnemy` + `nextZoneId`) in their response.
- The prefetched battle's `BattleStartTime` is anchored to the **scheduled cooldown expiry** (`EnemyCooldown`), not `now`. The cooldown is deterministic, so the next fight begins exactly when it elapses; anchoring there keeps the server-clock elapsed-time check correct (latency only ever *delays* the claim past the scheduled completion) and the *following* cooldown correctly sized — anchoring to `now` would back-date the start by the whole cooldown and shrink (or zero) it, which is the anti-cheat wrinkle the issue called out.
- The client holds the bundled enemy across the cooldown and presents it **only while the zone and the player's battle-relevant state are unchanged** since it was prepared. A zone or build change during the cooldown falls back to a fresh `NewEnemy` (re-syncing the new zone and taking a fresh snapshot that matches what the client now simulates).

### Parity (the subtle bit)
The bundled player snapshot is frozen at the previous battle's end, while the frontend derives its player battler live at fight time. They match **iff** nothing battle-relevant changed during the cooldown. The fallback above guarantees that: unchanged ⇒ bundle (state equal ⇒ parity holds); changed ⇒ fresh fetch (post-cooldown snapshot ⇒ parity holds). So there are no false victory rejections.

The boss-**victory** path is intentionally excluded — it paces itself via the Zone-Cleared overlay and re-challenge, so a bundled idle battle there would just be discarded.

## Addresses the acceptance criteria
- ✅ New-enemy fetch latency no longer adds dead time after the cooldown (idle-victory and boss-loss paths). Idle draws/defeats have no cooldown, so there is nothing to overlap there (their fetch latency is inherent, not added by ordering).
- ✅ Cooldown + elapsed-time anti-cheat stay correct under the prefetched flow (scheduled-start anchoring — no shrunken/zeroed cooldowns, no false rejections).
- ✅ FE/BE parity preserved via the zone/build-change fallback; all existing tests stay green.

## Test plan
- **Backend** (`dotnet test`): Core `711` ✓, Application `448` ✓ (incl. new `PrepareNextIdleBattle_AnchorsBattleStartToScheduledCooldownExpiry`), API `578` ✓ (incl. augmented `DefeatEnemy`/`BattleLost` socket tests asserting the bundled `NextEnemy`).
- **Frontend** (`npm run lint` + `npm run test:unit:ci`): lint/typecheck clean, `2362` tests ✓ — including new cases for presenting the bundle with no round-trip, and falling back on a zone change / build change / boss handoff during the cooldown.
- TypeScript API types regenerated via `codegen` (no drift).

## Notes
- `docs/backend-battle.md` updated with the bundled-prefetch + scheduled-start anchoring decision.

Closes #1092

https://claude.ai/code/session_017jzn5FhwPJ51ci5dXiNwDk

---
_Generated by [Claude Code](https://claude.ai/code/session_017jzn5FhwPJ51ci5dXiNwDk)_


Game.Application/Services/BattleService.cs
```diff
@@ -110,6 +114,23 @@ public async Task<BattleStartResult> StartBattle(Player player, PlayerState stat
...
+        /// scheduled time (a) keeps the elapsed-time victory check correct — network latency only ever delays
+        /// the claim past the scheduled completion — and (b) keeps the <em>following</em> cooldown correctly
+        /// sized: anchoring to now would back-date the start by the whole cooldown and shrink (or zero) it.
+        /// The result rides the battle-end response so the client can begin the next fight the instant the
+        /// cooldown elapses, without a separate <c>NewEnemy</c> round-trip. Call only after the battle-end
+        /// method has set the cooldown and cleared the resolved battle.
+        /// </summary>
+        public Task<BattleStartResult> PrepareNextIdleBattle(Player player, PlayerState state, CancellationToken cancellationToken = default)
+        {
+            return StartBattle(player, state, player.CurrentZoneId, scheduledStartTime: state.EnemyCooldown, cancellationToken: cancellationToken);
```
ginderjeremiah, line 131: **Blocking** — the prepared battle is left *active* on the server, so any path that supersedes it before it is actually fought will abandon-**resolve** it and record a phantom outcome for a fight that never happened.

After `PrepareNextIdleBattle`, `state` holds an active battle whose `BattleStartTime` is the scheduled cooldown expiry (`EnemyCooldown`). If the client doesn't go on to fight it — a **zone change**, a **build change**, or a **boss challenge** during the cooldown (all paths this PR adds and frontend-tests) — it instead calls `NewEnemy`/`ChallengeBoss`, which hits `if (state.HasActiveBattle) AbandonBattle(...)` in `StartBattle`/`StartBossBattle`.

The trouble is the timing: the client only fetches *after* waiting out the cooldown, so on arrival `now ≈ BattleStartTime + RTT`. `AbandonBattle` then sees `elapsedMs ≈ RTT > 0` (the `elapsedMs <= 0` clean-clear guard doesn't fire) and re-simulates the never-fought prepared battle for ~RTT ms:
- the common case (neither dies in ~RTT) is recorded as an **incomplete abandon → `BattlesAbandoned`**, inflating a tracked statistic that feeds challenges;
- an over-levelled player can resolve it as a **phantom victory** (free exp + win/kill credit), and an under-levelled one as a **phantom loss**.

Pre-PR there was no active battle during the cooldown, so superseding it was a clean no-op — this is a new regression introduced by keeping the prepared battle live. The scheduled-start anchoring (correct for the happy path) is exactly what makes `elapsedMs` land at ~RTT instead of ~0 here, so the two features interact badly.

Suggest discarding a not-yet-claimed prepared battle **without** recording an outcome when it's superseded (e.g. a clean clear for a scheduled battle the client never reported starting), and add a backend integration test for `NewEnemy`/`ChallengeBoss` issued after `PrepareNextIdleBattle` asserting no `BattlesAbandoned`/win/loss is recorded.

---

UI/src/lib/engine/battle/enemy-manager.ts
```diff
@@ -392,10 +417,64 @@ export class EnemyManager {
...
+	}
+
+	/**
+	 * Presents a prefetched battle as the live enemy (no round-trip). Re-checks the stop / transition guard
+	 * first because this path bypasses {@link getNewEnemy}'s fetch-generation guard: a stop or a
+	 * challenge/retreat that landed during the cooldown owns the next fight, so a stale prefetched idle enemy
+	 * must not clobber it.
+	 */
+	private presentPreparedBattle(prepared: PreparedBattle) {
+		if (!this.started || this.transitioning || this.mode !== 'idle') {
```
ginderjeremiah: **Should-fix** — `presentPreparedBattle` bypasses the `fetchGeneration`/`fetchingEnemy` re-entrancy guard that `getNewEnemy` relies on to prevent a double-spawn, re-checking only `started`/`transitioning`/`mode`.

If a `getNewEnemy` is ever in flight concurrently — the same "overlapping stage handlers" race the guard was built for (per `getNewEnemy`'s own comment: "an idle victory racing an Idle/Defeated change") — both can spawn-and-notify: one *fetched* enemy and one *prepared* enemy. The client's displayed enemy would then diverge from the server's active battle, and the next `DefeatEnemy` would be rejected. There's also the server-side interaction with the blocking finding: the fallback `getNewEnemy` abandons the prepared battle, so a racing present-then-fetch leaves client and server pointing at different enemies.

I couldn't construct a definite production trigger (`resume()` — the only internal `Idle` re-emit during a parked cooldown — isn't wired to anything external), so flagging as should-fix rather than blocking. Consider having this path participate in the fetch-generation guard (or add a short comment justifying why concurrent `getNewEnemy` can't occur here).

---

ginderjeremiah (commented):
## Review: Hide new-enemy fetch latency by bundling the next battle (#1092)

**Verdict:** Request changes — the prefetched battle is left active on the server, so the zone-change / build-change / boss-challenge fallbacks this PR adds will abandon-resolve a never-fought battle and record a phantom outcome.

> Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR — treat the verdict above as the real call.

The core idea is sound and the design write-up is excellent: scheduled-start anchoring keeps the elapsed-time check and the *following* cooldown correct, and the reference-equality `playerBattleStateMatches` fallback is a clean way to guarantee FE/BE parity (it also correctly captures the post-exp-grant level on both sides). One real problem keeps it from merging.

### Blocking
- **Phantom abandon of the prepared battle on every fallback** — `BattleService.cs:131` (`PrepareNextIdleBattle`) leaves an active battle on `state` with `BattleStartTime = EnemyCooldown`. When the client *doesn't* fight it (zone change, build change, or boss challenge during the cooldown — all supported, frontend-tested paths), it calls `NewEnemy`/`ChallengeBoss`, which abandons that battle. Because the client only fetches *after* the cooldown, `now ≈ BattleStartTime + RTT`, so `AbandonBattle` re-simulates the never-fought battle for ~RTT ms and records it — usually `BattlesAbandoned`, occasionally a phantom win (free exp) or loss. Pre-PR the cooldown had no active battle, so this is a new regression. Discard a not-yet-claimed prepared battle without recording an outcome, and cover it with a backend integration test. (Full detail inline.)

### Should-fix
- **`presentPreparedBattle` bypasses the re-entrancy guard** — `enemy-manager.ts:471` only re-checks `started`/`transitioning`/`mode`, not the `fetchGeneration`/`fetchingEnemy` guard `getNewEnemy` uses to prevent a double-spawn. If a concurrent `getNewEnemy` ever lands (the "overlapping stage handlers" race the guard exists for), the client could spawn both a fetched and a prepared enemy and desync from the server. I couldn't find a definite production trigger, so it's a should-fix — but it weakens an invariant the file deliberately maintains. (Inline.)

### Test coverage
- Frontend coverage is strong: present-the-bundle (victory + boss-loss), zone-change fallback, build-change fallback, and boss-challenge-during-cooldown are all exercised.
- Backend adds the anchoring test (`PrepareNextIdleBattle_AnchorsBattleStartToScheduledCooldownExpiry`) and augments the socket tests. The gap is the server side of the blocking finding — there is no integration test for `NewEnemy`/`ChallengeBoss` issued *after* `PrepareNextIdleBattle`, which is exactly where the phantom abandon hides. The frontend tests mock the socket, so they can't catch it.
- No battle-simulation logic changed, so the parity matrix needs no new mirrored cases — correct.

### What's good
- The scheduled-start anchoring rationale (and its `backend-battle.md` entry) is clear and the elapsed-time/anti-cheat reasoning holds up.
- Parity is handled carefully: the bundle is used only while zone + battle-relevant state are unchanged, with a fresh-fetch fallback otherwise; capturing player state *after* `grantExp` on both sides keeps a mid-cooldown level-up parity-safe.
- `EnemyInstance.FromSource` reuse keeps the bundled projection identical to the `NewEnemy` one — no drift.
- Boss-victory exclusion (it paces itself) is the right call and is guarded by capturing `wasBossBattle` before the battle is cleared.

---

ginderjeremiah:
Both review findings are addressed in 996d128.

**Blocking — phantom abandon.** Implemented per your steer (send the client's elapsed battle time on simulation-triggering requests rather than trusting wall-clock for the abandon window):
- `NewEnemy`/`ChallengeBoss` now carry `clientBattleMs` — how long the client actually simulated the battle being superseded. An unfought prefetched battle reports `0`; a genuinely-fought loss/draw reports its real duration.
- `AbandonBattle` bounds its replay by `clientBattleMs`, clamped to `[0, server-elapsed wall-clock]` then `DefaultMaxBattleMs`. So a never-fought prefetch re-simulates 0 ms → `elapsedMs <= 0` → clean clear, no outcome. The wall-clock cap is retained, so a won-abandon still can't be claimed faster than real time allows (under-reporting only ever resolves to an *earlier*, not-yet-won state — no reward exploit).

A note on why I didn't take the literal "discard on supersede" route: the supersede path alone can't distinguish a *fought-then-lost/drawn* prepared battle from a *never-fought* one — both arrive via the next `NewEnemy`, and idle losses/draws are recorded *by* that abandon, so blanket-discarding would have dropped real losses/draws that happen right after a win. Bounding by the client-reported duration is what cleanly separates them (0 ⇒ unfought ⇒ nothing; D ⇒ fought ⇒ recorded), which is exactly your suggestion.

Backend test added (`StartBattle_SupersedingAnUnfoughtPreparedBattle_RecordsNoOutcome`): after `PrepareNextIdleBattle`, superseding with `clientBattleMs: 0` despite a ~10-minute wall-clock gap records no `BattlesAbandoned`/loss/win and starts a fresh battle. `NewEnemy`/`ChallengeBoss` share the one `AbandonBattle` path, so it covers both.

**Should-fix — re-entrancy guard.** The prepared battle is now presented *inside* `getNewEnemy`'s guarded `runFetchLoop` (via an optional `prepared` arg) instead of a separate `presentPreparedBattle` entry point, so it shares the `fetchingEnemy`/`fetchGeneration` re-entrancy + supersession guards — a concurrent fetch can't double-spawn, and a transition that bumps the generation skips the present. `presentPreparedBattle`/`spawnNextIdleEnemy` are removed.

All green locally: Core 711, Application 449 (+1 new), API 578, frontend 2362, lint/typecheck clean; `backend-battle.md` updated.

---

PR #1110, opened 2026-06-21
Backend minor cleanups: ApplyExp threshold, EnsureViableZone, reconciler DRY (#1083)

### LINKED ISSUE #1083: Backend minor cleanups: GrantExp threshold dedup/fail-fast, EnsureViableZone double-resolve, Min↔AtMost invariant, reconciler DRY
Low-severity domain / data-access cleanup notes from a general backend code-health review. No active defects; bundled to avoid issue noise.

### 1. `Player.GrantExp` — recomputed threshold + a guard that masks a corrupt invariant
`Game.Core/Players/Player.cs:86-100`. The level-up threshold `Math.Max(1, Level) * GameConstants.ExpPerLevel` is computed twice per loop iteration (the `while` condition and the body subtraction) — hoist to one local per iteration (DRY). Separately, `Level` is `required` and every construction path yields `Level >= 1`, so the `Math.Max(1, Level)` guard only papers over a `Level == 0` corrupt aggregate instead of failing loudly. Prefer an explicit `Level >= 1` assertion at the top (consistent with the loud fail-fast policy used in e.g. `BattleSnapshot.FromPlayer`), then drop the `Math.Max`.

### 2. `EnsureViableZone` resolves the domain zone twice per candidate
`Game.Application/Services/BattleService.cs:330-353`. The `OrderBy(...).FirstOrDefault(z => IsZoneViable(z.Id) && _zones.GetDomainZone(z.Id).IsUnlocked(...))` calls `GetDomainZone` per candidate after `IsZoneViable` already resolved/validated the zone. Cold path (relocation only), so low priority, but filtering before ordering or resolving the domain zone once per candidate is cleaner.

### 3. Pin the `Min` → `AtMost` invariant
`Game.Core/Progress/ChallengeType.cs:26-34` derives `GoalComparison` purely from `AggregationKind == Min`. Correct today (only `FastestVictory` is `Min`, only `TimeTrial` maps to it), but it silently couples *every* future `Min`-backed statistic to an `AtMost` goal — a `Min` stat that should be "reach at least N" would be inverted. Add a test/assertion pinning the intended 1:1 `Min`↔`AtMost` mapping, or document the invariant inline.

### 4. Reconciler duplicate-message bypasses the shared helper (DRY)
`Game.DataAccess/Repositories/Admin/ChildCollectionReconciler.cs:51` hand-builds `"The submitted {resourceName} set contains duplicate entries."` while `ChangeSetProcessor.DuplicateFailure<T>` exists to centralize that copy (and `AttributeChangeSetProcessor` already uses it). Route the reconciler through the shared helper so the wording can't drift.

ginderjeremiah:
## Summary

Addresses the four low-severity backend code-health notes bundled in #1083. No behaviour change beyond a stricter fail-fast and a unified rejection message.

### 1. `Player.ApplyExp` — recomputed threshold + a guard masking a corrupt invariant
`Game.Core/Players/Player.cs`. The level-up threshold `Math.Max(1, Level) * ExpPerLevel` was computed twice per loop iteration; it's now hoisted to one `levelThreshold` local (recomputed once per level gained). The `Math.Max(1, Level)` guard only papered over a corrupt `Level == 0` aggregate, so it's replaced with an explicit `Level < 1` fail-fast at the top, consistent with the loud-fail policy in `BattleSnapshot.FromPlayer`.

### 2. `EnsureViableZone` resolved the domain zone per candidate inside the predicate
`Game.Application/Services/BattleService.cs`. The relocation scan now filters to viable candidates **before** ordering, then resolves each domain zone once via a lazy `Select` (so `FirstOrDefault` stops resolving at the first unlocked match) instead of re-resolving inside the predicate. Cold path (relocation only); purely a readability/redundancy cleanup.

### 3. Pin the `Min` → `AtMost` invariant
`Game.Core/Progress/ChallengeType.cs`. The existing tests pin the *derivation rule* (`AtMost` iff `Min`-aggregated), but that rule is self-referential and can't catch a future `Min`-backed statistic whose challenge should still accumulate "at least". Added `EveryChallengeTypeMapsToItsIntendedGoalComparison`, an exhaustive per-type expected-direction map guarded so a newly-added `EChallengeType` forces a deliberate declaration before it passes.

### 4. Reconciler duplicate-message bypassed the shared helper (DRY)
`Game.DataAccess/Repositories/Admin/ChildCollectionReconciler.cs`. The hand-built duplicate message now routes through `ChangeSetProcessor.DuplicateFailure`, so the wording is centralized and can't drift. This unifies the copy from `"... {resource} set contains duplicate entries."` to `"... {resource} change set contains duplicate entries."`; the three reconciler-path test assertions were updated to match.

## Test plan
- `Game.Core.Tests` — 713 passed (incl. new `GrantExp_CorruptZeroLevel_FailsFastInsteadOfSpinning` and `EveryChallengeTypeMapsToItsIntendedGoalComparison`).
- `Game.Application.Tests` — 455 passed (incl. the integration tests whose duplicate-message assertions were updated).

## Notes
- Backend-only; no frontend or battle-parity surface touched.

Closes #1083

---
_Generated by [Claude Code](https://claude.ai/code/session_01Se2SsXiddsQjTLaWaJ8znB)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminEnemiesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminEnemiesIntegrationTests.cs
@@ -268,7 +268,7 @@ namespace Game.Application.Tests.DataAccess
             var result = admin.SetAttributeDistributions(data);
 
             Assert.False(result.Succeeded);
-            Assert.Equal("The submitted attribute distribution set contains duplicate entries.", result.ErrorMessage);
+            Assert.Equal("The submitted attribute distribution change set contains duplicate entries.", result.ErrorMessage);
         }
 
         [Fact]
--- a/Game.Application.Tests/DataAccess/AdminZonesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminZonesIntegrationTests.cs
@@ -111,7 +111,7 @@ namespace Game.Application.Tests.DataAccess
             var result = admin.SetEnemies(data);
 
             Assert.False(result.Succeeded);
-            Assert.Equal("The submitted enemy set contains duplicate entries.", result.ErrorMessage);
+            Assert.Equal("The submitted enemy change set contains duplicate entries.", result.ErrorMessage);
         }
 
         [Fact]
--- a/Game.Application.Tests/DataAccess/ChildCollectionReconcilerTests.cs
+++ b/Game.Application.Tests/DataAccess/ChildCollectionReconcilerTests.cs
@@ -67,7 +67,7 @@ namespace Game.Application.Tests.DataAccess
             var result = Reconcile(existing, desired, recorder);
 
             Assert.False(result.Succeeded);
-            Assert.Equal("The submitted child set contains duplicate entries.", result.ErrorMessage);
+            Assert.Equal("The submitted child change set contains duplicate entries.", result.ErrorMessage);
             Assert.Empty(recorder.Deleted);
             Assert.Empty(recorder.Updated);
             Assert.Empty(recorder.Inserted);
--- a/Game.Application/Services/BattleService.cs
+++ b/Game.Application/Services/BattleService.cs
@@ -610,10 +610,13 @@ namespace Game.Application.Services
             }
 
             var completedChallengeIds = await _progressRepo.GetCompletedChallengeIds(player.Id, cancellationToken);
+            // Filter to viable candidates before ordering, then resolve each domain zone once (lazily, so the
+            // resolve stops at the first unlocked match) rather than re-resolving it inside the predicate.
             var destination = _zones.All()
+                .Where(zone => IsZoneViable(zone.Id))
                 .OrderBy(zone => zone.Order)
-                .FirstOrDefault(zone =>
-                    IsZoneViable(zone.Id) && _zones.GetDomainZone(zone.Id).IsUnlocked(completedChallengeIds));
+                .Select(zone => _zones.GetDomainZone(zone.Id))
+                .FirstOrDefault(zone => zone.IsUnlocked(completedChallengeIds));
             var newZoneId = destination?.Id ?? NewPlayerFactory.StartingZoneId;
 
             if (newZoneId != player.CurrentZoneId)
--- a/Game.Core.Tests/Players/PlayerTests.cs
+++ b/Game.Core.Tests/Players/PlayerTests.cs
@@ -170,6 +170,16 @@ namespace Game.Core.Tests.Players
             Assert.Empty(player.DomainEvents.OfType<PlayerLeveledUpEvent>());
         }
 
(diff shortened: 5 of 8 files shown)


ginderjeremiah (commented):
## Review: Backend minor cleanups addressing the four notes in #1083

**Verdict:** Approve _(submitted as a comment since I'm the PR author and GitHub won't accept a self-gating verdict)_ — all four cleanups land exactly as the issue scoped them, with no behaviour change beyond the intended fail-fast and the unified rejection message, and each is covered by tests.

I verified all four items against the issue and the surrounding code:

1. **`Player.ApplyExp` threshold dedup + fail-fast** (`Game.Core/Players/Player.cs:118`) — threshold hoisted to a single `levelThreshold` local (recomputed once per level gained), and the `Math.Max(1, Level)` guard replaced with an explicit `Level < 1` `InvalidOperationException`. Matches the loud-fail policy the issue cites, and the ordering (guard before any mutation) is correct.
2. **`EnsureViableZone` double-resolve** (`Game.Application/Services/BattleService.cs:615`) — `Where(IsZoneViable).OrderBy(Order).Select(GetDomainZone).FirstOrDefault(IsUnlocked)` is behaviour-equivalent to the old predicate: the domain `Zone.Id` equals the reference-data id passed to `GetDomainZone`, so `destination?.Id` resolves the same zone, and the lazy `Select` preserves the short-circuit so resolution still stops at the first unlocked match.
3. **`Min`↔`AtMost` invariant** (`ChallengeTypeTests.cs`) — the new exhaustive per-type expected-direction map is guarded by `Enum.GetValues<EChallengeType>()`, so a newly-added type forces a deliberate declaration. I confirmed all 8 enum values are present and the directions are correct (including `LevelReached` → `AtLeast` via the null-statistic path).
4. **Reconciler DRY** (`ChildCollectionReconciler.cs:53`) — now routes through `ChangeSetProcessor.DuplicateFailure`. I grepped the whole test suite: every duplicate-message assertion now reads "change set", with no stragglers on the old "set" wording — the copy is genuinely unified across all processors now.

### Test coverage
- New `GrantExp_CorruptZeroLevel_FailsFastInsteadOfSpinning` covers the fail-fast; new `EveryChallengeTypeMapsToItsIntendedGoalComparison` pins the concrete mapping; the three reconciler-path duplicate-message assertions were updated to match.
- `EnsureViableZone` gets no new test, which is fine — it's a behaviour-preserving readability refactor on a cold path already exercised by the existing relocation tests.

### What's good
Tightly scoped, each change traceable to its note in #1083, comments explain the "why" without over-narrating, and the DRY change is the kind that actually prevents future drift rather than just shuffling code. CI is green.

No blocking findings, no should-fixes, no nits. Merging.

---

PR #1149, opened 2026-06-22
Fix re-entrant async worker loop corrupting IsRunning (#1136)

### LINKED ISSUE #1136: BackgroundWorker async loop can run re-entrantly and corrupt IsRunning

## Problem

`Game.Infrastructure/BackgroundWorker.cs` registers its wait with `executeOnlyOnce: false`:

```csharp
private RegisteredWaitHandle RegisterWorker(WaitOrTimerCallback callback)
{
    return ThreadPool.RegisterWaitForSingleObject(_resetEvent, callback, null, -1, false);
}
```

The thread pool re-arms the wait as soon as the callback **returns**. For the **async** loop the callback is `async (state, timedOut) => { await loopAction(); ...; IsRunning = false; }` (`CreateAsyncWorkerLoop`, lines 176-192), which returns to the pool at the **first incomplete `await`**, not at task completion. So:

1. A second `Start()` (`_resetEvent.Set()`) that arrives while a previous async run is still in flight fires a **second concurrent** `loopAction()`.
2. The first run's trailing `IsRunning = false` (line 189) can clobber the second run's `IsRunning = true`.

The careful "publish `IsRunning = true` before `Set()`" ordering documented at lines 96-99 only protects the **sync** loop (`CreateWorkerLoop`), whose callback doesn't return until `loopAction()` completes. The async path is the case it doesn't cover.

The pub/sub wake path calls `worker.Start()` on every inbound message, so back-to-back wakes during a drain are the normal operating mode — making overlapping async runs realistically reachable, not a remote edge case.

## Impact

- Two concurrent drain passes over the same write-behind / cache-reload queue.
- `IsRunning` no longer reliably reflects whether a run is active.

## Suggested fix

For the async path, re-arm manually after the awaited task completes (register `executeOnlyOnce: true` and re-register in a `finally`), or guard re-entry with an interlocked in-progress flag that coalesces overlapping signals into a single trailing run.

## Tests

Add a concurrency test that signals the async worker repeatedly while a run is in flight and asserts non-overlapping execution + a stable terminal `IsRunning`.

ginderjeremiah:
## Summary

Fixes **#1136**: `BackgroundWorker`'s **async** worker loop could run re-entrantly and corrupt `IsRunning`.

The callback is registered with `executeOnlyOnce: false`, so the thread pool re-arms the wait as soon as the callback **returns** — which for an `async` callback is at its **first incomplete `await`**, not at task completion. So a `Start()` arriving while a previous async run was still in flight fired a **second concurrent** `loopAction()`, and the first run's trailing `IsRunning = false` could clobber the second run's `IsRunning = true`. The pub/sub wake path calls `Start()` on every inbound message, so back-to-back wakes during a drain make this realistically reachable, not a remote edge case.

The careful "publish `IsRunning = true` before `Set()`" ordering only protects the **sync** loop, whose callback doesn't return until `loopAction()` completes — the async path is the case it didn't cover.

## Fix

Gate the async loop with a lock-guarded `_runActive` / `_runPending` pair (the issue's "interlocked in-progress flag that coalesces overlapping signals into a single trailing run" option):

- A callback that finds a run already active records `_runPending` and bows out, so **two `loopAction()` calls never overlap**.
- The active run consumes `_runPending` in a `do/while`, so **any number of mid-run signals coalesce into exactly one trailing pass** (no lost wakes, no unbounded re-runs).
- `IsRunning` is set `true` on claim and reset `false` only at the gated exit, both under the lock, so a finishing run can't clobber a fresh run's state.

The lock is held only for the brief state transitions, **never across the awaited `loopAction`**. The **sync path is untouched** (it isn't vulnerable — its callback returns only after `loopAction()` completes). I chose this over re-registering with `executeOnlyOnce: true` to avoid per-run wait re-registration on the hot pub/sub wake path and the extra coordination it would need with `Kill()`/`Dispose()`.

## How it addresses the issue

- No overlapping async runs; `IsRunning` reaches a stable terminal value.
- Two concurrent drain passes over the same write-behind / cache-reload queue can no longer happen.

## Test plan

- [x] New concurrency test `AsyncWorker_SignaledRepeatedlyDuringRun_NeverOverlapsAndSettlesNotRunning`: holds the first async run suspended mid-`await`, fires 20 `Start()`s while it's in flight, then asserts **max observed concurrency == 1**, that the coalesced trailing run still executed (signals not lost), and a stable terminal `IsRunning == false`.
- [x] **Verified the test reproduces the bug**: reverting the production fix makes the new test fail (concurrent runs drive max-concurrency past 1); with the fix it passes.
- [x] Ran the new test 5× to check for flakiness — green every time.
- [x] Full `Game.Infrastructure.Tests` suite green (58 passed), `dotnet build` clean (0 warnings / 0 errors).

Closes #1136

---
_Generated by [Claude Code](https://claude.ai/code/session_016KuUMwx6QYptACoLmeBV7A)_

### DIFF
--- a/Game.Infrastructure.Tests/BackgroundWorkerTests.cs
+++ b/Game.Infrastructure.Tests/BackgroundWorkerTests.cs
@@ -339,6 +339,77 @@ namespace Game.Infrastructure.Tests
             }
         }
 
+        [Fact]
+        public void AsyncWorker_SignaledRepeatedlyDuringRun_NeverOverlapsAndSettlesNotRunning()
+        {
+            // Reproduces #1136: an async callback returns to the thread pool at its first incomplete await, which
+            // re-arms the wait mid-run. Pre-fix, a Start() during an in-flight async run launched a *second concurrent*
+            // loopAction and the finishing run's trailing IsRunning = false could clobber the new run's true. The fix
+            // gates the async loop to one run at a time and coalesces mid-run signals into a single trailing pass.
+            var active = 0;
+            var maxActive = 0;
+            var totalRuns = 0;
+            using var firstRunStarted = new ManualResetEventSlim(false);
+            var releaseFirstRun = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            async Task Action()
+            {
+                var now = Interlocked.Increment(ref active);
+                InterlockedMax(ref maxActive, now);
+                Interlocked.Increment(ref totalRuns);
+
+                // Hold only the first run suspended mid-await so the test can fire more signals while it is in flight;
+                // every later (coalesced) run runs straight through.
+                if (!firstRunStarted.IsSet)
+                {
+                    firstRunStarted.Set();
+                    await releaseFirstRun.Task.WaitAsync(WaitTimeout);
+                }
+
+                Interlocked.Decrement(ref active);
+            }
+
+            var worker = new BackgroundWorker(NullLogger<BackgroundWorker>.Instance, Action);
+
+            worker.Start();
+            Assert.True(firstRunStarted.Wait(WaitTimeout, TestContext.Current.CancellationToken), "The first async run did not start.");
+            Assert.True(worker.IsRunning);
+
+            // Signal repeatedly while the first run is suspended at its await. Pre-fix each re-armed wait fired a
+            // concurrent run (driving maxActive past 1); post-fix they collapse into one trailing pass.
+            for (var i = 0; i < 20; i++)
+            {
+                worker.Start();
+            }
+
+            releaseFirstRun.SetResult();
+            WaitUntil(() => !worker.IsRunning && Volatile.Read(ref active) == 0);
+
+            // Two loopAction calls never ran at once.
+            Assert.Equal(1, Volatile.Read(ref maxActive));
+            // The mid-run signals were not lost — the coalesced trailing run executed.
+            Assert.True(Volatile.Read(ref totalRuns) >= 2, "The coalesced trailing run did not execute.");
+            // The worker settles in a stable, non-running terminal state.
+            Assert.False(worker.IsRunning);
+
+            worker.Kill();
+        }
+
+        // Atomically raises target to value when value is greater, used to detect any overlapping run.
+        private static void InterlockedMax(ref int target, int value)
+        {
+            int current;
+            do
+            {
+                current = Volatile.Read(ref target);
+                if (value <= current)
+                {
+                    return;
+                }
+            }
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Fix re-entrant async worker loop corrupting `IsRunning` (#1136)

**Verdict:** Approve _(submitted as a comment — GitHub won't accept a formal approval from the PR author)_ — the fix correctly closes the async re-entrancy described in #1136, is minimal and well-documented, and the new test genuinely reproduces the bug.

### Correctness

- The lock-guarded `_runActive` / `_runPending` pair is the right shape for the problem. A callback that fires mid-run (the thread pool re-arming the wait at the first incomplete `await`) records `_runPending` and bows out, so two `loopAction()` invocations can never overlap; the `do/while` then consumes `_runPending` and coalesces any number of mid-run signals into exactly one trailing pass — no lost wakes, no unbounded re-runs.
- `IsRunning` transitions (`true` on claim, `false` only at the gated exit) both happen under `_runLock`, and a claim requires `_runActive == false` — which is only published by the prior run's gated reset under the same lock. So a fresh run's `IsRunning = true` always happens-after a finishing run's `IsRunning = false`; the wedged-`false` clobber from the issue can't occur. The reaffirm-on-claim comment captures this reasoning accurately.
- The lock is only held across the brief state transitions, never across the awaited `loopAction`, and there's no nested lock acquisition between `_disposeLock` and `_runLock`, so no deadlock or contention on the hot wake path.
- Sync path is correctly left untouched — its callback returns only after `loopAction()` completes, so it was never vulnerable. The choice to gate rather than re-register with `executeOnlyOnce: true` avoids per-run wait re-registration on the pub/sub wake path; well-justified in the PR body.

### Test coverage

- `AsyncWorker_SignaledRepeatedlyDuringRun_NeverOverlapsAndSettlesNotRunning` is a solid reproduction: it suspends the first run mid-`await`, fires 20 `Start()`s while in flight, and asserts max observed concurrency == 1 (overlap), `totalRuns >= 2` (signals not lost to coalescing), and a stable terminal `IsRunning == false`. Because all signals are fired and drained during the suspension window before release, the assertions aren't subject to a late-claim TOCTOU.
- Consistent with the file's conventions and `docs/backend.md` classical testing guidance — deterministic signal-based synchronization (no sleeps), observable-behavior assertions, real `NullLogger`. The author verified the test fails when the fix is reverted and ran it 5× for flakiness.

### What's good

- The extended comments on the async loop and the gate fields are appropriately detailed for tricky concurrency code and make the happens-before reasoning easy to follow without over-explaining.

No blocking issues, should-fixes, or nits. CI is green (frontend jobs correctly skipped). Merging.

---

PR #1150, opened 2026-06-22
Make challenge type→statistic resolution total and fail-loud (#1137)

### LINKED ISSUE #1137: Unmapped EChallengeType silently never progresses — add a fail-loud guard

## Problem

`Challenge.UpdateChallengeProgress` (`Game.Core/Progress/Challenge.cs:24-36`) only advances progress when the challenge type has a backing statistic **or** is the special-cased `LevelReached`:

```csharp
if (Type.StatisticType is not null) { ... }
else if (Type.Id is EChallengeType.LevelReached) { ... }
```

`ChallengeType.GetStatisticType` (`Game.Core/Progress/ChallengeType.cs:36-49`) maps 7 of the 8 `EChallengeType` values and falls through to `_ => null` (which is also the legitimate result for `LevelReached`, handled specially above).

If a new `EChallengeType` is ever added **without** both (a) a `GetStatisticType` arm and (b) a branch in `UpdateChallengeProgress`, a challenge of that type will **silently never progress** — no exception, no completion, its reward permanently unreachable. That contradicts the project's fail-loud convention: every other enum dispatch in the domain (`StatisticType.GetEntityType`, `Attribute.GetDescription`, …) throws `ArgumentOutOfRangeException` on the default arm.

## Suggested fix

Make the type→statistic resolution total and fail-loud: add an explicit `LevelReached => null` arm and throw `ArgumentOutOfRangeException` on the default in `GetStatisticType` (or assert in `UpdateChallengeProgress` that a type with no statistic is `LevelReached`). Then a new challenge type can't be wired up half-way without a loud failure.

## Tests

A test asserting every `EChallengeType` either maps to a statistic or is explicitly handled (no silent no-op).

scope: small

ginderjeremiah:
## Summary

Fixes **#1137**. A new `EChallengeType` added without both (a) a `GetStatisticType` arm and (b) a branch in `UpdateChallengeProgress` would have its challenges **silently never progress** — no exception, no completion, its reward permanently unreachable. That contradicted the project's fail-loud convention for enum dispatch. This closes both halves of that gap.

## Changes

- **`ChallengeType.GetStatisticType`** — added an explicit `EChallengeType.LevelReached => null` arm and replaced the catch-all `_ => null` with a `throw new ArgumentOutOfRangeException`. The type→statistic resolution is now total: a newly-added type with no mapping arm throws in the `ChallengeType` constructor, which runs eagerly at cache build (so it surfaces as a boot failure, not a silent runtime no-op).
- **`Challenge.UpdateChallengeProgress`** — added a final `else` that throws for any statistic-less type that isn't explicitly handled (currently only `LevelReached`). This closes the second half: a new type given a `null` statistic mapping but no progress branch would otherwise still silently no-op.

## How it addresses the issue

Both required wiring points are now fail-loud, so a challenge type can no longer be wired up half-way without a loud, diagnosable failure naming the offending type — rather than shipping a challenge whose reward is permanently unreachable.

## Tests

- `ChallengeTypeTests.EveryChallengeTypeEitherMapsToAStatisticOrIsExplicitlyHandled` — asserts every `EChallengeType` either carries a backing statistic or is the deliberate statistic-less accumulator (`LevelReached`), with no silent fall-through.
- `ChallengeTests.UpdateChallengeProgress_EveryChallengeType_AdvancesProgress_NeverSilentlyNoOps` — drives `UpdateChallengeProgress` for every type with satisfying data and asserts each completes, so a half-wired new type trips the fail-loud guard instead of silently never progressing.

## Test plan

- [x] `dotnet build Game.Core` — clean (0 warnings / 0 errors).
- [x] Full `Game.Core.Tests` suite green (**729 passed, 0 failed**).

Closes #1137

---
_Generated by [Claude Code](https://claude.ai/code/session_017t3HkdPiaQ1tSogauqfJrU)_

### DIFF
--- a/Game.Core.Tests/Progress/ChallengeTests.cs
+++ b/Game.Core.Tests/Progress/ChallengeTests.cs
@@ -66,6 +66,31 @@ namespace Game.Core.Tests.Progress
             Assert.False(playerChallenge.Completed);
         }
 
+        [Fact]
+        public void UpdateChallengeProgress_EveryChallengeType_AdvancesProgress_NeverSilentlyNoOps()
+        {
+            // Each challenge type must actually move progress when its tracked value is satisfied — a
+            // statistic-backed type from its statistic, LevelReached from the player's level. A type that
+            // is neither would fall through the fail-loud guard and throw, so this fails on a half-wired
+            // new type rather than letting it silently never progress.
+            var player = new PlayerBuilder().WithLevel(50).Build();
+
+            foreach (var type in Enum.GetValues<EChallengeType>())
+            {
+                var challenge = MakeChallenge(type, goal: 1);
+                var playerChallenge = new PlayerChallenge(challenge, progress: 0m, completed: false);
+                var statisticType = challenge.Type.StatisticType;
+                var statistics = statisticType is not null
+                    ? new[] { new PlayerStatistic { Type = statisticType.Id, EntityId = null, Value = 1m } }
+                    : [];
+                var progress = MakeProgress(player, statistics);
+
+                challenge.UpdateChallengeProgress(playerChallenge, progress);
+
+                Assert.True(playerChallenge.Completed, $"Challenge type {type} did not progress.");
+            }
+        }
+
         private static Challenge MakeChallenge(EChallengeType type, decimal goal, int? targetEntityId = null) => new()
         {
             Id = 0,
--- a/Game.Core.Tests/Progress/ChallengeTypeTests.cs
+++ b/Game.Core.Tests/Progress/ChallengeTypeTests.cs
@@ -135,5 +135,28 @@ namespace Game.Core.Tests.Progress
                 Assert.Equal(type, challengeType.Id);
             }
         }
+
+        [Fact]
+        public void EveryChallengeTypeEitherMapsToAStatisticOrIsExplicitlyHandled()
+        {
+            // The type→statistic resolution is total and fail-loud: every challenge type must either
+            // carry a backing statistic or be a deliberately statistic-less accumulator (LevelReached).
+            // A newly-added type with no GetStatisticType arm throws in the constructor rather than
+            // silently resolving to "no statistic" and never progressing.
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Make challenge type→statistic resolution total and fail-loud (#1137)

**Verdict:** Approve — implements exactly the fix #1137 called for, closing both wiring gaps with the project's standard fail-loud enum dispatch, and backs it with guard tests. (Posted as a comment because GitHub won't accept a gating verdict from the PR author; the intended verdict is **Approve**.)

This is a tight, well-scoped change. Both halves of the silent-no-op gap are now closed:

- `ChallengeType.GetStatisticType` replaces the catch-all `_ => null` with an explicit `LevelReached => null` arm plus a throwing default, so an unmapped type fails loudly in the constructor — which runs eagerly at cache build, surfacing as a boot failure rather than a runtime no-op. This matches the convention in `StatisticType.GetEntityType` / `Attribute.GetDescription`.
- `UpdateChallengeProgress` adds the throwing `else`, closing the second half (a future type given a `=> null` mapping but no progress branch). It's effectively defensive given current types — construction throws first for any unmapped type — but it's the right guard for the scenario the issue describes.

The comments are clear about *why* each guard exists, which is exactly the kind of context that keeps a later edit from naively "simplifying" the explicit `LevelReached => null` arm back into the catch-all.

### Test coverage

Solid and on-point. `ChallengeTypeTests.EveryChallengeTypeEitherMapsToAStatisticOrIsExplicitlyHandled` locks the type→statistic totality (only `LevelReached` is statistic-less), and `ChallengeTests.UpdateChallengeProgress_EveryChallengeType_AdvancesProgress_NeverSilentlyNoOps` drives every type to completion so a half-wired addition trips a guard instead of silently never progressing. The goal=1 / value=1 setup correctly completes both `AtLeast` and the `AtMost` (TimeTrial/FastestVictory) comparison. No frontend mirror needed — this is backend reference-data wiring, not battle logic.

### Nits

None.

Nice work — this is exactly the fail-loud guard the issue asked for. Will merge once CI is green.

---

PR #1151, opened 2026-06-22
Fix write-behind slot data-integrity gaps (#1138)

### LINKED ISSUE #1138: Write-behind handler data-integrity gaps: ModAppliedHandler drops a colliding distinct mod; equipment slot has no DB uniqueness
Found during a code-health review. Two related gaps in the player-update write-behind handlers, both around concurrent / reordered apply correctness. The persistence layer is explicitly designed to tolerate cross-instance reorder and at-least-once double-apply, so these are genuine gaps rather than hypotheticals — though both are bounded today by single-connection-per-player and (see #BackgroundWorker) sequential drains.

## 1. `ModAppliedHandler` can silently drop a *different* mod on a slot collision

`Game.DataAccess/PlayerUpdates/Handlers/ModAppliedHandler.cs:28-35` deletes the slot row, inserts the new one, then absorbs a unique violation as a no-op:

```csharp
catch (DbUpdateException ex) when (ex.IsUniqueViolation())
{
    // A concurrent apply inserted the same (player, item, slot) first; the row exists, so this is a no-op.
}
```

The `AppliedMod` PK is `(PlayerId, ItemId, ItemModSlotId)` — it **excludes `ItemModId`**. The no-op justification ("both applies carry the identical ItemModId") only holds for a double-apply of the *same* event. Two genuinely **different** `ModAppliedEvent`s targeting the same slot (concurrent or reordered) collide on the slot PK, and the loser's distinct `ItemModId` is **dropped terminally** — unlike the sibling `ItemEquippedHandler`, which on a unique violation does `ChangeTracker.Clear()` and re-runs the delete-then-insert (last-writer-wins).

**Fix:** mirror `ItemEquippedHandler` — on unique violation, `ChangeTracker.Clear()` and re-run, so last-writer-wins matches the absolute-upsert intent.

## 2. Missing DB uniqueness on `(PlayerId, EquipmentSlotId)`

`UnlockedItem` is keyed only on `(PlayerId, ItemId)` (`GameContext.cs:441-444`); "one item per slot" is enforced purely in app code via the load-clear-save in `ItemEquippedHandler.ApplyAsync` (`:37-69`), which is not atomic across instances. Two different equip events for the same slot applied concurrently can each load before the other commits and both succeed, leaving the slot **doubly occupied** with no constraint to catch it.

**Fix:** add a partial unique index on `(PlayerId, EquipmentSlotId) WHERE EquipmentSlotId IS NOT NULL` and handle the violation with the same clear-and-re-apply the handler family already uses.

## Tests

Integration tests applying two distinct mods to one slot, and two distinct equips into one slot, asserting convergence to a single last-writer-wins row.

scope: medium

ginderjeremiah:
## Summary

Closes #1138. Two related correctness gaps in the player-update write-behind handlers, both around concurrent/reordered apply on a **slotted child row**.

### 1. `ModAppliedHandler` dropped a colliding *different* mod
The `AppliedMod` PK `(PlayerId, ItemId, ItemModSlotId)` **excludes `ItemModId`**, so two genuinely different `ModAppliedEvent`s targeting the same slot collide on the slot key. The old `catch` absorbed the loser's unique violation as a no-op — justified only for an *identical* double-apply — so the loser's distinct mod was **dropped terminally**.

**Fix:** on the violation the slot row already exists, so settle it with a **server-side last-writer-wins `ExecuteUpdate`** to the latest mod. An update can't re-violate the PK, so it converges without looping or propagating a second failure (robust even under the suite's 8× concurrent double-apply), and an identical double-apply is a harmless self-write.

### 2. Missing DB uniqueness on `(PlayerId, EquipmentSlotId)`
"One item per slot" was enforced purely in app code via `ItemEquippedHandler`'s load-clear-save, which isn't atomic across instances, so two concurrent equips into one slot could each load before the other commits and both succeed — leaving the slot **doubly occupied** with no constraint to catch it.

**Fix:** add a **partial unique index** on `(PlayerId, EquipmentSlotId) WHERE EquipmentSlotId IS NOT NULL` (migration `AddUnlockedItemSlotUniqueIndex`).

## Note: `ItemEquippedHandler` reworked beyond the literal ask

The issue suggested handling the new violation "with the same clear-and-re-apply the handler family already uses." Adding the index surfaced that the existing **load-clear-save** is fragile under a concurrent retry: the prior occupant is materialized into a snapshot that a concurrent commit can tear, so the in-memory clear is silently skipped and the upsert collides again. I therefore changed `ApplyAsync` to **vacate the destination slot with a single server-side `ExecuteUpdate` before placing the item** — mirroring `ModAppliedHandler`'s proven `ExecuteDelete`-then-write — and kept the once-only retry on the rare cross-instance race.

Tradeoff: the clear and the upsert are now **two commits** rather than one transaction, so a crash between them leaves the slot momentarily empty; the queue's reserve/acknowledge read redelivers and converges, so there is no lost write (the same tradeoff `ModAppliedHandler` already accepts). Happy to revisit if you'd rather keep a single-transaction shape.

## Tests

All deterministic and stable (the genuine 2-way concurrent distinct-slot race is a non-production scenario — per-player drains are sequential — and reproducing it in-process triggers a connection-pool snapshot artifact, so it isn't asserted directly; the existing same-event concurrent tests already cover each handler's retry branch):

- `UnlockedItems_DuplicateSlotOccupant_RejectedByUniqueIndex` — the index rejects a second occupant of an already-filled slot (mirrors the existing `PlayerStatistics` index guard).
- `UnlockedItems_MultipleUnequippedItems_AllowedByPartialFilter` — the partial filter still lets many unequipped (null-slot) rows coexist.
- `ItemEquipped_IntoSlotOccupiedByHigherIdItem_ClearsOccupantWithoutColliding` — equipping into an occupied slot clears the prior occupant regardless of row-id order (guards against an EF command-ordering regression the index could otherwise expose).

### Test plan
- [x] `dotnet build` clean (0 warnings / 0 errors).
- [x] Full backend suite green — Core 729, Infrastructure 58, Application 468, Api 580; 0 failed.
- [x] `PlayerUpdateHandlerIdempotencyIntegrationTests` run 5× consecutively, stable (31/31 each).

## Docs
- `docs/backend-persistence.md` — documents that a no-op is the wrong resolution where a row's natural key is narrower than its payload (mods settle last-writer-wins), and the new equipment-slot DB invariant + clear-then-write equip path.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HNHhFKwtrdcERUMSAoJXwp)_


Game.Infrastructure/Database/GameContext.cs
```diff
@@ -441,6 +441,14 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<UnlockedItem>(entity =>
             {
                 entity.HasKey(ui => new { ui.PlayerId, ui.ItemId });
+
+                // One item per equipment slot, enforced at the DB level so two concurrent equips into the same
+                // slot can't both succeed — the load-clear-save in ItemEquippedHandler isn't atomic across
+                // instances. The partial filter lets the many unequipped (null-slot) rows coexist, and declaring
+                // the index lets EF order the clear of a slot's prior occupant before the new occupant takes it.
```
ginderjeremiah: Nit (comment accuracy): this rationale describes the *old* handler shape. `ItemEquippedHandler` no longer does a "load-clear-save" — it now vacates the slot with an explicit server-side `ExecuteUpdate` and then upserts in a separate commit. Likewise "declaring the index lets EF order the clear of a slot's prior occupant before the new occupant takes it" no longer applies: the clear is now its own statement that commits before the placing `SaveChanges`, so there's no single-batch UPDATE-ordering for EF to arrange (that's exactly the hazard the dedicated `ItemEquipped_IntoSlotOccupiedByHigherIdItem_ClearsOccupantWithoutColliding` test guards against). Suggest rewording to "...the load-clear-save the handler *used* to do isn't atomic across instances" and dropping the EF-ordering sentence, so a future reader isn't misled about why the clear is split out.

---

ginderjeremiah (commented):
## Review: Fix write-behind slot data-integrity gaps (#1138)

**Verdict:** Approve (intended) — submitted as `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on my own PR. No blocking findings; one comment-accuracy nit and one acknowledged test-coverage note.

Both gaps from #1138 are addressed correctly and the reasoning is sound:

- **`ModAppliedHandler`** now resolves the slot-PK violation with a server-side last-writer-wins `ExecuteUpdate` instead of swallowing it as a no-op. Since the `AppliedMod` PK `(player, item, slot)` excludes `ItemModId`, the old no-op terminally dropped a *different* colliding mod; the update converges without re-violating the PK or looping. Correct.
- **`(PlayerId, EquipmentSlotId)` uniqueness** is now a partial unique index (`WHERE EquipmentSlotId IS NOT NULL`). The partial filter is the right call here — it sidesteps the `NULLS NOT DISTINCT` requirement that nullable unique columns otherwise need (per `backend-persistence.md`), since null-slot rows are excluded from the index entirely.
- **`ItemEquippedHandler` rework** (clear-then-write via a dedicated `ExecuteUpdate`, retry once) is a justified deviation from the issue's literal "clear-and-re-apply" ask: it avoids materializing the prior occupant into a tearable snapshot and mirrors `ModAppliedHandler`'s proven shape. The two-commit tradeoff (momentary empty slot on a crash between statements, recovered by reserve/acknowledge redelivery) is consistent with the rest of the handler family and is documented.

I traced the equip paths (re-apply, item moving between slots, reordered-unlock insert, concurrent-winner retry) and they all converge to the intended last-writer-wins state.

### Nits
- `GameContext.cs:446-448` — index-rationale comment still describes the old "load-clear-save" handler and an EF batch-ordering behavior the final code no longer relies on. Inline comment with suggested wording. (Pure comment accuracy — not blocking.)

### Test coverage
Equip/index side is well covered, including a deterministic guard for the command-ordering hazard the index could expose (`ItemEquipped_IntoSlotOccupiedByHigherIdItem_ClearsOccupantWithoutColliding`), the index rejecting a second occupant, and the partial filter still allowing many null-slot rows.

One gap worth naming explicitly: the **distinct-mod last-writer-wins resolution in `ModAppliedHandler`'s catch branch** — the literal core of gap #1 — has no test that distinguishes the new behavior from the old no-op. `ModApplied_AppliedConcurrently_...` exercises the catch branch but only with an *identical* mod id (a self-write), and `ModApplied_AppliedTwice_ConvergesToOneRowWithLatestMod` covers the last-writer outcome via the non-catch (delete-then-insert) path. So a regression of the catch branch back to a no-op would not be caught. The PR documents that deterministically interleaving an insert between the delete and `SaveChanges` (the only way to force the catch with a distinct mod) triggers a connection-pool snapshot artifact in-process, which is a fair reason — flagging it as an accepted, documented gap rather than a blocker, since the *outcome* (two distinct mods on one slot → latest wins) is already asserted.

### What's good
Thorough, accurate persistence docs update; the partial-index `NULLS NOT DISTINCT` reasoning is exactly right; the deviation from the issue's literal ask is well-justified and explicitly surfaced for review rather than buried.

Not merging despite the green CI and approve verdict, per the workflow's rule that a clean auto-merge requires zero nits — the comment-accuracy nit is left for you to fold in (or wave off) first.

---

PR #1154, opened 2026-06-22
Fix torn read in Zones reference repo across build-then-swap (#1139)

### LINKED ISSUE #1139: Zones reference repo can tear across a build-then-swap (validate-then-index on two snapshot reads)

## Problem

`Game.DataAccess/Repositories/Zones.cs:11` re-reads the volatile snapshot on every access:

```csharp
private IReadOnlyList<Zone> Entities => holder.Current;
```

The lookups (`GetZone`/`GetDomainZone`/`LookupZone`/`IsZoneRetired`) call `ValidateZoneId(zoneId)` (which reads `Entities.Count`, **read #1**) and then `Entities[zoneId]` (**read #2**). If a build-then-swap publishes a *shorter* snapshot between the two reads, validation passes against the old list and the index throws `IndexOutOfRangeException` against the new one.

`Enemies.cs` deliberately avoids exactly this by capturing `var snapshot = Snapshot;` once per method; `Zones` is the lone reference repo that doesn't.

## Impact

Low probability (snapshots rarely shrink, and the swap window is tiny), but it's a real torn-read that would surface as a spurious exception on a hot zone lookup during an admin-triggered reload.

## Suggested fix

Capture `var entities = holder.Current;` once per method and run both the bounds check and the index against that local, matching the `Enemies` pattern.

scope: small

ginderjeremiah:
## Summary

Fixes the torn read in `Game.DataAccess/Repositories/Zones.cs` flagged in #1139. Each lookup (`GetZone` / `GetDomainZone` / `LookupZone` / `IsZoneRetired`) called `ValidateZoneId(zoneId)` — which read `Entities.Count` (read #1 of the volatile snapshot) — and then indexed `Entities[zoneId]` (read #2). A build-then-swap publishing a *shorter* snapshot between the two reads could pass validation against the old list and throw `IndexOutOfRangeException` against the new one on a hot zone lookup during an admin-triggered reload.

## How it addresses the issue

- Each lookup now captures `holder.Current` **once** into a local and runs both the bounds check and the index against that same list — the exact snapshot-capture pattern `Enemies` already uses (`var snapshot = Snapshot;`).
- Extracted a private `IsValidZoneId(IReadOnlyList<Zone> entities, int zoneId)` helper so the bounds check operates on the captured local. The public `ValidateZoneId(int zoneId)` (on `IZones`, called by `Enemies`/`BattleService`) is preserved and delegates to the helper — it was already a single volatile read, so it never teared on its own.

No behavior change for valid or invalid ids; this is correctness hardening for the concurrent reload window.

## Test plan

- [x] `dotnet build Game.DataAccess` — clean, 0 warnings.
- [x] `ZonesIntegrationTests` (14 tests) green — covers `GetZone`/`GetDomainZone`/`LookupZone`/`IsZoneRetired` valid + invalid-id paths.

## Note on test coverage

I did not add a dedicated torn-read regression test. Reproducing the race deterministically requires injecting a snapshot swap *between* the two reads, but `ZonesCacheHolder` is `internal sealed` with a private, non-injectable `_current`, and `Zones` depends on the concrete holder — so a deterministic test would mean adding production seams that exist only for the test. This matches how the identical `Enemies` capture pattern is covered (integration tests pin behavior; the capture itself is verified by reasoning). Happy to add a fakeable seam + unit test if preferred.

Closes #1139

---
_Generated by [Claude Code](https://claude.ai/code/session_01DmdJuwGS3R2JXmg39S4Xcw)_

### DIFF
--- a/Game.DataAccess/Repositories/Zones.cs
+++ b/Game.DataAccess/Repositories/Zones.cs
@@ -20,33 +20,44 @@ namespace Game.DataAccess.Repositories
 
         public Contracts.Zone GetZone(int zoneId)
         {
-            return ValidateZoneId(zoneId)
-                ? ZoneMapper.ToContract(Entities[zoneId])
+            // Capture the volatile snapshot once so the bounds check and the index read the same list;
+            // a build-then-swap publishing a shorter snapshot between the two reads would otherwise tear.
+            var entities = holder.Current;
+            return IsValidZoneId(entities, zoneId)
+                ? ZoneMapper.ToContract(entities[zoneId])
                 : throw new ArgumentOutOfRangeException(nameof(zoneId));
         }
 
         public Core.Zones.Zone GetDomainZone(int zoneId)
         {
-            return ValidateZoneId(zoneId)
-                ? ZoneMapper.ToCore(Entities[zoneId])
+            var entities = holder.Current;
+            return IsValidZoneId(entities, zoneId)
+                ? ZoneMapper.ToCore(entities[zoneId])
                 : throw new ArgumentOutOfRangeException(nameof(zoneId));
         }
 
         public Zone? LookupZone(int zoneId)
         {
-            return ValidateZoneId(zoneId) ? Entities[zoneId] : null;
+            var entities = holder.Current;
+            return IsValidZoneId(entities, zoneId) ? entities[zoneId] : null;
         }
 
         public bool IsZoneRetired(int zoneId)
         {
-            return ValidateZoneId(zoneId)
-                ? Entities[zoneId].RetiredAt is not null
+            var entities = holder.Current;
+            return IsValidZoneId(entities, zoneId)
+                ? entities[zoneId].RetiredAt is not null
                 : throw new ArgumentOutOfRangeException(nameof(zoneId));
         }
 
         public bool ValidateZoneId(int zoneId)
         {
-            return zoneId >= 0 && zoneId < Entities.Count;
+            return IsValidZoneId(holder.Current, zoneId);
+        }
+
+        private static bool IsValidZoneId(IReadOnlyList<Zone> entities, int zoneId)
+        {
+            return zoneId >= 0 && zoneId < entities.Count;
         }
     }
 }


ginderjeremiah (commented):
## Review: Fix torn read in Zones reference repo across build-then-swap (#1139)

**Verdict:** Approve — the fix correctly resolves the torn read by capturing the volatile snapshot once per lookup, exactly as the issue prescribes. (Submitted as a comment because GitHub treats me as the PR author and won't accept a gating verdict on my own PR.)

The change replaces the per-access `Entities => holder.Current` re-reads in each lookup with a single `var entities = holder.Current;` capture, then runs both the bounds check (`IsValidZoneId`) and the index against that same local. I verified `ReferenceCacheHolder.Current` is a `Volatile.Read` over a reference field swapped atomically via `Volatile.Write`, so capturing once is sufficient to guarantee the validate and the index see the same snapshot. This matches the `Enemies` capture pattern (`var snapshot = Snapshot;`) and closes the build-then-swap window described in #1139. No behavior change for valid/invalid ids.

### What's good

- Minimal, targeted fix that aligns `Zones` with the established `Enemies` snapshot-capture pattern.
- Extracting `IsValidZoneId(IReadOnlyList<Zone>, int)` keeps the bounds check DRY while preserving the public `ValidateZoneId(int)` on `IZones` (used by `Enemies`/`BattleService`), which delegates to the helper — that path was already a single volatile read so it never teared.
- Comment density is right: one concise explanatory comment on the first method rather than repeated on each.

### Test coverage

The PR adds no dedicated torn-read regression test, and I agree with that call. Deterministically reproducing the race requires injecting a snapshot swap *between* the two reads, which would mean adding production seams (a fakeable holder) that exist only for the test — for a change that produces no observable behavior difference on valid/invalid ids. The existing `ZonesIntegrationTests` pin the valid + invalid-id behavior of all four lookups, and this mirrors how the identical `Enemies` capture is covered. Acceptable for a small correctness-hardening fix.

No blocking issues, no should-fixes, no nits.


---

PR #1156, opened 2026-06-22
Clear Loading.svelte reveal-delay timer on unmount (#1140)

### LINKED ISSUE #1140: Loading.svelte: uncleared setTimeout writes $state after unmount

## Problem

`UI/src/components/Loading.svelte:34-39` arms a reveal-delay timer in `onMount` and never clears it (the `onMount` callback returns no cleanup):

```js
onMount(() => {
    waitingOnDelay = !!delay;
    if (delay) {
        setTimeout(() => (waitingOnDelay = false), delay);
    }
});
```

If a `Loading` with a `delay` unmounts before the delay elapses — common, since a spinner that resolves before its reveal delay is exactly the case the delay exists for — the timer still fires and writes `waitingOnDelay` on a destroyed component. Every other timer in the codebase (CombatFloaters, ToastContainer, LivePreview) is tracked and cleared; this is the one that isn't.

## Fix

```js
onMount(() => {
    waitingOnDelay = !!delay;
    if (!delay) return;
    const t = setTimeout(() => (waitingOnDelay = false), delay);
    return () => clearTimeout(t);
});
```

scope: small

ginderjeremiah:
## Summary

`Loading.svelte` armed a reveal-delay `setTimeout` in `onMount` but never cleared it. When a `Loading` with a `delay` unmounts before the delay elapses — exactly the case the delay exists for, a spinner that resolves before its reveal delay — the timer still fired and wrote `$state` (`waitingOnDelay`) on a destroyed component. Every other timer in the codebase (CombatFloaters, ToastContainer, LivePreview) is tracked and cleared; this was the lone exception.

## How it addresses the issue

`onMount` now returns a cleanup function that clears the timer, so the pending write can never run after the component is destroyed:

```js
onMount(() => {
    waitingOnDelay = !!delay;
    if (!delay) {
        return;
    }
    const timer = setTimeout(() => (waitingOnDelay = false), delay);
    return () => clearTimeout(timer);
});
```

## Test plan

- [x] Added a test asserting the reveal-delay timer is cleared on unmount.
- [x] Added a test asserting no timer is armed when no `delay` is set.
- [x] `npx vitest --run Loading` — all green.

Closes #1140

---
_Generated by [Claude Code](https://claude.ai/code/session_01ATWg3pvxYe5A4T4uQnBMYD)_

### DIFF
--- a/UI/src/components/Loading.svelte
+++ b/UI/src/components/Loading.svelte
@@ -33,9 +33,11 @@ let waitingOnDelay = $state(true);
 
 onMount(() => {
 	waitingOnDelay = !!delay;
-	if (delay) {
-		setTimeout(() => (waitingOnDelay = false), delay);
+	if (!delay) {
+		return;
 	}
+	const timer = setTimeout(() => (waitingOnDelay = false), delay);
+	return () => clearTimeout(timer);
 });
 
 const offset = Math.floor(Math.random() * 360);
--- a/UI/src/tests/components/Loading.test.ts
+++ b/UI/src/tests/components/Loading.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, afterEach } from 'vitest';
+import { describe, it, expect, afterEach, vi } from 'vitest';
 import { render, cleanup } from '@testing-library/svelte';
 import Loading from '$components/Loading.svelte';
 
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Clear `Loading.svelte` reveal-delay timer on unmount (#1140)

**Verdict:** Approve _(submitted as a `COMMENT` because GitHub won't accept a gating verdict on one's own PR — treat this as an approval)_ — the change is exactly the fix #1140 called for, correctly implemented and tested.

The bug is real: `onMount` armed a `setTimeout` and returned no cleanup, so a `Loading` with a `delay` that unmounted before the delay elapsed (the very case the delay exists for) would still fire and write `$state` on a destroyed component. The fix returns a cleanup that clears the timer, bringing this in line with every other timer in the codebase (CombatFloaters, ToastContainer, LivePreview).

### What's good
- Implementation matches the issue's proposed fix and correctly uses braces on the early-return `if` per the `CLAUDE.md` house rule (rather than the brace-less form in the issue snippet).
- Tests live in the right place — `UI/src/tests/components/Loading.test.ts` mirroring the source path, per `docs/frontend.md`.
- Both branches are covered: timer cleared on unmount, and no timer armed when `delay` is unset.

### Test coverage
Adequate for a fix this size. One optional thought (not blocking): the unmount test asserts `clearTimeout` was *called* rather than called with the specific timer id; a behavioral variant using fake timers (advance past the delay after unmount, assert no further `$state` write) would be slightly more robust against false positives. The current spy-based test is direct and fine to keep.

No blocking or should-fix findings. Will merge once CI goes green.


---

Issue #1161, opened 2026-06-22
Home-tier falloff routing + absolute-falloff proficiency XP accrual


ginderjeremiah:
Part of the **Path refinement** to the proficiency system (spike #982 → [Paths](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/982-proficiency-system.md#paths-refinement)).

**Reworks the merged XP accrual from #1116** so a path-contribution **routes to the path's current tier** and **falls off with tier distance**, with the falloff applied as an **absolute** multiplier — the fixed pie becomes a *ceiling*, so coasting on stale skills genuinely trains slower (not just relatively).

### What changes

- **Routing.** At battle completion each *represented* path resolves to its **frontier tier** (the lowest un-maxed, unlocked proficiency in the path); the contribution lands there.
- **Falloff weight.** A fired skill's pull on the frontier tier is `tierWeight × Weight × falloff(frontierOrdinal − HomeTier)`, with `falloff(0) = 1` decaying per the path's `FalloffBase`. Distance is always ≥ 0 (a tier's native skills are gated behind that tier opening, so an older skill never trains a tier *below* where it was acquired).
- **Absolute-falloff split** (the deliberate economic change). Split the fixed pie across represented paths by their **falloff-free attention** (`Σ tierWeight × Weight` — how much of the loadout points at the path), then scale each path's slice by its **on-tier efficiency** (the contribution-weighted average falloff). The un-earned remainder is **not minted**. This evaporation is intentional (the cost of coasting) and is distinct from the maxed-tier slice-drop already in place.
- Rework `ProficiencyXpCalculator.Split` (separate *attention* from *efficiency*; keep it pure/testable) and `ProficiencyRewardService.BuildContributions` (resolve skill → path → frontier tier + falloff).
- **Live + offline parity.** Same loadout over the away window → identical falloff; still computed off the 40 ms tick loop, so no battle-parity surface is added.

### Worked check (strawman `falloff(1) = 0.3`)

- Solo path, on-tier loadout → full pie. Solo path coasting on 1-step-stale skills → ~`0.3 ×` pie (absolute slowdown), the rest evaporates.
- A freshly-opened tier with its **native seed skill** (home tier = that tier, falloff 1) trains at full pace; the same tier on inherited lower-tier skills only trains at the discount — the pull to acquire the new tier's natives. Never walled, just slower.

### Acceptance

[...]

Issue #1132, opened 2026-06-22
Admin Workbench editor for proficiencies


ginderjeremiah:
EDIT: This will go through a standard design phase and may result in different requirements than what is described below.

Split from #1114, which delivered the backend foundation + admin endpoints for the proficiency system (spike #982). Build the admin Workbench editor UI, consuming the admin endpoints already shipped in #1114.

See `docs/spikes/982-proficiency-system.md` -> area A.

- Add `UI/src/routes/admin/workbench/entities/proficiency.ts` (`EntityConfig`) with sections: Identity (name, description, icon, maxLevel, baseXp, xpGrowth, startsUnlocked, seed-skill picker), per-level modifiers, per-level reward skills, prerequisites, and skill contributions.
- Register it in `entities/index.ts` under a `progression` group.
- Wire `persist()` to the shipped endpoints: `AdminTools/AddEditProficiencies`, `SetProficiencyModifiers`, `SetProficiencyRewards`, `SetProficiencyPrerequisites`, `SetProficiencyContributions`.
- Deliver the proficiency reference data to the workbench reference store for the pickers.

**Depends on:** #1114.

ginderjeremiah:
**Scope grows with the Path refinement** (spike #982 → [Paths](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/982-proficiency-system.md#paths-refinement); backend lands in #1160).

This editor now needs to cover **two** entities in the `progression` group:
- **Path** editor — identity (name, description, `FalloffBase`) + an ordered-tiers view + a **skill-contributions** sub-editor with a **home-tier** column (contributions moved off the proficiency onto the path).
- **Proficiency** editor — identity now includes **path membership** (`PathId` + `PathOrdinal`); the old contributions section is **gone** (it lives on the Path); level modifiers / rewards / cross-path prerequisites stay.

The backend contracts (`Path`, reworked `SkillPathContribution`, `Proficiency.PathId/PathOrdinal`, `SetPathContributionsData`) and codegen'd TS interfaces ship in #1160.

ginderjeremiah:
**Add: tier-ordinal uniqueness validation** (surfaced reviewing #1163).

The backend enforces `(PathId, PathOrdinal)` uniqueness via a DB unique index, but `AdminProficiencies.SaveProficiencies` does **not** pre-check it — a tier collision currently surfaces as a raw unique-violation (500) rather than a graceful `AdminSaveResult.Failure`.

[...]

Issue #1116, opened 2026-06-22
Proficiency XP accrual + leveling on battle completion (live + offline)


ginderjeremiah:
Part of spike #982 (proficiency system). See `docs/spikes/982-proficiency-system.md` -> area C. **The core mechanic.**

**Depends on:** areas A + B.

Compute proficiency XP at battle completion and apply leveling.

- Fixed-pie total per won battle, scaled by the same difficulty curve as DefeatRewards (anti-grind).
- Split across proficiencies "represented" (a contributing skill -- selected OR innate item skill -- fired in the won battle), each slice proportional to sum(skillTierWeight x contributionWeight). Consumes BattleStats.SkillStats.
- Apply to PlayerProficiency with multi-level handling vs XpCurve.
- Wire live (BattleStatisticsEventHandler) and offline consolidation (constant loadout over the window).
- Tier weight flat (=1) until #979 lands (see the deferred tier-weighting tracker).
- Emit a per-proficiency result payload (xp gained / new level / milestones crossed) for the client push.

Tests: pie-split + difficulty scaling, multi-level gain, offline==live accrual.

ginderjeremiah:
From the #1114 review ([#1133](https://github.com/ginderjeremiah/GameServer/pull/1133)): when the leveling/curve interpretation lands here, add backend bounds validation on authored proficiency `Level` (e.g. `1 ≤ Level ≤ MaxLevel`) in the admin `SetProficiencyModifiers`/`SetProficiencyRewards` setters. #1114 ships those setters without it, so a modifier/reward authored at level 0 or beyond the cap is currently accepted silently (the cap isn't interpreted until this issue).

Issue #1114, opened 2026-06-22
Proficiency reference-data model + skill contributions + admin authoring


ginderjeremiah:
Part of spike #982 (proficiency system). See `docs/spikes/982-proficiency-system.md` -> area A. **Foundation (no dependency).**

Add the `Proficiency` authored reference-data entity and the `SkillProficiency` contribution join.

- `Proficiency`: Id, Name, Description, MaxLevel (~10), XpCurve, Prerequisites (proficiency ids to max), PerLevelModifiers (attribute modifiers applied x level), Milestones ({Level, Modifiers, RewardSkillId?}), SeedSkillId? (tree-opened nodes), StartsUnlocked (root flag), RetiredAt.
- `SkillProficiency` join: SkillId, ProficiencyId, Weight (multi-contribution).
- DbContext + migration; cache holder + snapshot + VersionKey; entity<->core<->contract mappers; reader repo + DI (mirror SkillsCacheHolder).
- Admin Workbench proficiency editor (Identity / Levels & milestones / Contributions) + add/edit endpoints; add a `progression` group.

Tests: mapping + contiguity unit tests; cache-reload + admin-CRUD integration. Strawman caps/curves/numbers per the spike doc (tunable).

PR #1163, opened 2026-06-22
Add Path system to proficiency reference data (#1160)

### LINKED ISSUE #1160: Path system: reference-data model, contribution-join rework, admin authoring
Part of the **Path refinement** to the proficiency system (spike #982 → [Paths](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/982-proficiency-system.md#paths-refinement)).

Introduces the **path** — a linear sequence of proficiencies (Fire Magic → Inferno Magic) treated as a first-class unit, so a skill ties to the *sequence* once instead of to every node in it. This **reworks the merged foundation from #1114**.

### What changes

- **New `Path` entity** (authored reference data): `Id`, `Name`, `Description`, `FalloffBase` (the per-tier contribution decay), `RetiredAt`. Ordered membership is carried on the proficiencies via `PathOrdinal`.
- **`Proficiency` gains `PathId` + `PathOrdinal`** (its tier within the path). Each tier keeps its own identity, bonuses, milestones, and seed skill — tiers are **distinct nodes, not renamed bands**, so `PlayerProficiency` per-proficiency progress (#1115) is unchanged.
- **Contribution join reworked**: `SkillProficiency (SkillId, ProficiencyId, Weight)` → **`SkillPathContribution (SkillId, PathId, HomeTier, Weight)`**. A skill feeds a *path* at a home tier; multiple rows now mean a genuine cross-**path** (multi-school) skill. Within-path spread is the falloff curve (handled in the accrual issue), not extra rows.
- **Prerequisites reduced to cross-path gateways.** Within-path order (`PathOrdinal`) makes the linear edges implicit; `ProficiencyPrerequisite` rows remain only for gateways (lava = advanced-fire + advanced-earth → first tier of the lava path). A standalone proficiency is just a path of length 1.
- DbContext config + **migration** (replace the #1114 shapes — dev DB, no back-compat needed).
- Cache holder + snapshot + `VersionKey`; entity↔core↔contract mappers; reader repository + the reverse **skill → path contributions** index the accrual consumes.
- **Admin Workbench**: a **path editor** (identity + ordered tiers + `FalloffBase`) alongside the revised proficiency editor; the contribution editor now targets a path + home tier.

### Acceptance

- Authoring a path with ordered tiers, a skill→path contribution with a home tier, and a cross-path gateway round-trips through the admin API and the reference cache.
- The reverse skill→contribution index exposes `(PathId, HomeTier, Weight)` for the accrual path.
- Unit tests (mappers, cache snapshot/versioning, immutable reference instances) and admin integration tests (add/edit path, contribution, gateway), mirroring the existing proficiency test suites.

**Depends on:** nothing new (extends merged #1114). **Blocks:** the home-tier falloff accrual rework.

ginderjeremiah:
## Summary

Implements the backend of the **Path system** — the [Paths refinement](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/982-proficiency-system.md#paths-refinement) to the proficiency system (spike #982). A **path** is an ordered sequence of proficiencies that share an identity arc (Fire Magic → Inferno Magic) and is the first-class unit a skill contributes to, so a skill ties to the *sequence* once instead of to every node.

Scope is **backend only**, matching how #1114 actually shipped: the Workbench authoring UI is tracked by #1132 (commented there), and the home-tier-falloff XP routing by #1161.

## How it addresses #1160

- **New `Path`** entity/contract, cached as reference data (0-based id, retirement). `Proficiency` gains a required `PathId` + `PathOrdinal` with a unique `(PathId, PathOrdinal)` tier index. Within-path order is implicit, so `ProficiencyPrerequisite` is now reserved for **cross-path gateways**.
- **Contribution join reworked**: `SkillProficiency` → `SkillPathContribution (SkillId, PathId, HomeTier, Weight)`, hung off the `Path`.
- **Admin**: new `AdminPaths` (identity save + a contributions setter that validates each home tier resolves to a real tier of the path); the proficiency identity save now takes path membership and rejects an unknown path; the contributions endpoint moved from the proficiency controller to a new `AdminPathsController`.
- **Accrual shim**: the reverse skill→contribution index resolves each path contribution to the proficiency at its home tier, preserving the `{ProficiencyId, Weight}` shape the merged XP accrual (#1116) reads — so this PR doesn't touch the accrual and stays green. #1161 replaces the shim with the path-routed shape and applies the falloff.
- EF migration (`AddPathSystem`) and **regenerated** TypeScript contracts (`IPath`, `ISkillPathContribution`, `IProficiency.pathId/pathOrdinal`, `ISetPathContributionsData`, `AddEditPaths`/`SetPathContributions`).

A `Path` type alias (`using Path = …`) is used in the four files that import the entity/contract namespace, mirroring the existing `using Attribute = …` pattern, to avoid the `System.IO.Path` collision.

## Test plan

- [x] `dotnet build Game.sln` — clean (0 warnings).
- [x] Backend tests: **Core 774**, **Application 512**, **Infrastructure 58**, **Api 594** — all pass (incl. the Api codegen-drift tests and the Testcontainers integration suite, which applies the new migration).
- [x] New/updated tests: `PathMapper` mapping; `ProficienciesIntegrationTests` (path-routed contribution resolves back via the shim, `AllPaths`); `AdminProficienciesIntegrationTests` (unknown-path rejection, path-seeded saves); new **`AdminPathsIntegrationTests`** (identity save, unknown-skill / bad-home-tier rejection, reconcile).
- [x] Frontend: `npm run lint` (eslint + svelte-check, 0 errors) and `npm run test:unit:ci` pass.

## Follow-ups

- `docs/game-design.md` / `docs/backend.md` prose for the path model + absolute falloff is deferred to #1161, when the mechanic goes live (noted in the spike).
- Workbench path + proficiency editors: #1132.

Closes #1160


Game.Application.Tests/DataAccess/ProficienciesIntegrationTests.cs
```diff
@@ -86,11 +92,16 @@ public async Task GetProficiency_AssemblesLevels_AndContributionIndexIsExposed()
...
+            // The contribution targets the path at the proficiency's tier; the reverse index resolves it back
+            // to the home-tier proficiency (the shim the accrual reads until #1161).
             var contributions = proficiencies.ContributionsForSkill(skillId);
             Assert.Equal(proficiencyId, contributions.Single().ProficiencyId);
             Assert.Equal(1.5d, contributions.Single().Weight);
 
             Assert.Contains(proficiencies.AllProficiencies(), p => p.Id == proficiencyId && p.Name == "Blades");
+            Assert.Contains(
+                proficiencies.AllPaths(),
+                p => p.Name == "Fire" && p.Contributions.Any(c => c.SkillId == skillId && c.HomeTier == 0));
```
ginderjeremiah, line 104: **Should-fix (test coverage):** This is the only test exercising the reverse-index shim — the one genuinely new piece of logic in this PR (`proficiencyIdByTier[(path.Id, c.HomeTier)]` in `ProficienciesCacheHolder`) — but it does so with a **single-tier path whose contribution is homed at tier 0**. At ordinal 0 the resolution `(path, HomeTier)` → proficiency is indistinguishable from "the only/first proficiency of the path", so a regression that ignored `HomeTier` (e.g. resolved to `path.Proficiencies[0]`) would still pass.

Please add a case with a **multi-tier path** where a contribution homed at tier ≥ 1 resolves back to the tier-≥1 proficiency (a different id than tier 0). That pins the shim's actual contract and is what the project's testing rigor calls for on novel logic.

---

Game.DataAccess/Repositories/Admin/AdminProficiencies.cs
```diff
@@ -221,14 +201,29 @@ private static Entities.ProficiencyLevelReward ToRewardEntity(int proficiencyId,
...
+                }
+
+                if (_proficiencies.LookupPath(change.Item.PathId) is null)
+                {
+                    return AdminSaveResult.Failure($"Path {change.Item.PathId} does not exist.");
+                }
+
+                if (change.Item.PathOrdinal < 0)
+                {
+                    return AdminSaveResult.Failure("A proficiency's path ordinal (tier) cannot be negative.");
```
ginderjeremiah, line 222: **Nit / follow-up:** `FindPathViolation` cleanly pre-checks path-exists and ordinal ≥ 0, but not the `(PathId, PathOrdinal)` **uniqueness** the new index enforces. Two added/edited proficiencies colliding on the same tier (or colliding with an existing tier of that path) will slip past this check and surface as a raw DB unique-violation — an opaque 500 — rather than the graceful `AdminSaveResult.Failure` the comment above ("a named check fails the whole set cleanly before anything is staged") establishes as the pattern for the FK.

Not blocking — the Workbench editor (#1132) will likely guard ordinals client-side, and it's authoring-only — but a named pre-check here would match the existing rationale and avoid the 500. Fine to defer to #1132 if you'd rather own it there.

---

ginderjeremiah (commented):
## Review: Path system backend (#1160) — Path reference entity, contribution-join rework, admin authoring

**Intended verdict: Approve** (posted as a comment because I'm the PR author, so GitHub won't accept a gating event on my own PR). No blocking findings — one should-fix on test coverage and two notes below. Because there's an outstanding should-fix, I'm not auto-merging; happy to merge once the multi-tier shim test lands (or if you decide to defer it as a tracked follow-up).

The change matches the issue's scope precisely: `Path` as zero-based retired-able reference data, `Proficiency` gains `PathId`/`PathOrdinal` with a unique `(PathId, PathOrdinal)` tier index, the join moves from `SkillProficiency` to `SkillPathContribution (SkillId, PathId, HomeTier, Weight)`, prerequisites narrow to cross-path gateways, and the accrual is left untouched behind a reverse-index shim that resolves each path contribution back to its home-tier proficiency. Layering and the docs' reference-data rules are respected throughout.

### Should-fix
- **Shim tier-resolution is under-tested** — `Game.Application.Tests/DataAccess/ProficienciesIntegrationTests.cs:103`. The novel logic (`proficiencyIdByTier[(path.Id, c.HomeTier)]`) is only exercised by a single-tier path homed at tier 0, where it's indistinguishable from "first proficiency of the path". Add a multi-tier case where a contribution homed at tier ≥ 1 resolves to the tier-≥1 proficiency (inline detail on the file).

### Notes (non-blocking)
- **Duplicate `(PathId, PathOrdinal)` isn't pre-validated** in `AdminProficiencies.FindPathViolation` (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs:222`) — a tier collision surfaces as a raw DB unique-violation (500) rather than a graceful `AdminSaveResult.Failure`. Authoring-only; fine to own in #1132.
- **Migration isn't apply-safe onto a non-empty `Proficiencies` table** — every existing row defaults to `PathId=0, PathOrdinal=0`, which both violates the new FK (no `Path` 0 exists) and collides on the unique index for ≥2 rows. This is the explicitly-accepted "dev DB, no back-compat — reseed" decision from #1160, so it's expected, not a defect — flagging only so a reseed is on the radar when applying locally.

### Test coverage
Solid for the scope: `PathMapper` round-trip (incl. `HomeTier`), unknown-path rejection on proficiency save, the new `AdminPaths` suite (identity save, unknown-skill and bad-home-tier rejection, reconcile-against-home-tier), `AllPaths`, and the shared-instance/contiguity coverage. `TestDataSeeder` correctly re-homes existing accrual links at the proficiency's tier so the existing accrual integration tests keep passing through the shim. The one gap is the multi-tier shim case above. No frontend tests needed — the TS changes are pure regenerated codegen, and no battle-parity surface is touched (accrual stays off the tick loop).

### What's good
- Clean DDD: entity-free admin contract surface (`IAdminPaths`), mapper in `Game.DataAccess/Mapping`, contract reads only; the entity seam stays internal.
- The shim is the right call — it isolates #1161 and keeps the accrual untouched/green while reshaping the schema.
- `(PathId, PathOrdinal)` unique index bakes the tier invariant into the schema and makes the home-tier → proficiency resolution unambiguous (the cache build asserts contiguity and would fail loud on a bad authored tier — consistent with the reference-data philosophy).
- `using Path = …` alias mirrors the existing `Attribute` pattern; no dead code left behind (old `SkillProficiency`/`SetProficiencyContributions` types fully removed, remaining references are only historical migration designers + the spike doc).
- Migration is reversible and the `RetiredAt` + retire-only `SavePaths` (no delete handler) correctly inherit the reference-data lifecycle.

---

ginderjeremiah:
Thanks for the review. Addressed:

**Should-fix — multi-tier shim test:** added in baadaa22. `ContributionsForSkill_ResolvesToTheHomeTierProficiency_NotTierZero` builds a two-tier path and a contribution homed at tier 1, asserting it resolves to the tier-1 proficiency (and *not* tier 0) — so a regression that ignored `HomeTier` (e.g. `path.Proficiencies[0]`) now fails.

**Note — `(PathId, PathOrdinal)` uniqueness pre-check:** deferring to #1132 (tracked there). A naive per-change occupant check would false-reject a valid tier **reorder/swap** — the cache reflects pre-edit state, so moving A→tier 1 while B vacates tier 1 in the same batch reads as a collision. Doing it right needs the full desired per-path layout, which the editor sends; the unique index is the correctness backstop until then.

**Note — migration onto a non-empty `Proficiencies` table:** as you noted, this is the accepted "dev DB, reseed" decision from #1160 — no change.

PR #1164, opened 2026-06-22
Backend minor cleanups: allocation nits, dead defensive code, conventions (#1143)

### LINKED ISSUE #1143: Backend minor cleanups (allocation nits, convention inconsistencies, dead defensive code)
Found during a code-health review — a basket of small, low-risk cleanups (cf. the prior "Backend minor cleanups" #1083/#1110). None is individually worth a PR; together they're worth one sweep.

## Allocation / perf nits
- **`AggregateRoot.DomainEvents`** (`Game.Core/AggregateRoot.cs:17`) returns `_domainEvents.AsReadOnly()`, allocating a fresh `ReadOnlyCollection<T>` on **every** get; the dispatcher reads it multiple times per drain (`DomainEventDispatcher`). Cache the wrapper once (or expose `Count` + a single materialization).
- **`Roles.GetRoles()`** (`Game.DataAccess/Repositories/Roles.cs:12-21`) re-enumerates `Enum.GetValues<ERole>()` and rebuilds `Role` instances (with reflection-backed `ToString()`) on every call for intrinsic enum data. Precompute as a `static readonly` list.
- **`Zones.GetDomainZone`** (`Game.DataAccess/Repositories/Zones.cs:28-33`) maps a fresh `CoreZone` per call via `ZoneMapper.ToCore`, on the per-battle setup path (`BattleService.cs:92,143,214`). Every other gameplay read (`Items`/`ItemMods`/`Skills`/`Enemies`) hands back shared pre-materialized instances per the documented convention. Pre-materialize `CoreZone`s into the zones snapshot and return shared instances. (Coordinate with the Zones torn-read fix.)

## Convention / dead code
- **`EnemyMapper.ToTemplate`** (`Game.DataAccess/Mapping/EnemyMapper.cs:64-67`) filters out-of-range skill ids with `.Where(es => es.SkillId >= 0 && es.SkillId < allSkills.Count)`. Contiguity is asserted before templates build and `SkillId` is an FK into that set, so the filter can never exclude anything valid — and if it ever did, it would **silently drop** a skill from an enemy loadout, contradicting the loud-fail policy the player-load path enforces (`OrphanedReferenceException`). Index directly, or throw on an out-of-range id.
- **`Player.UnlockSkill`** (`Game.Core/Players/Player.cs:173-180`) does an O(n) `Skills.Any(s => s.Id == skill.Id)` scan, inconsistent with the deliberately O(1) `UnlockItem`/`UnlockMod` paths (whose comments call out the unlocked set being read on the battle-start hot path). Skills grow far slower than items, so either mirror the id-keyed approach or add a one-line comment that the linear scan is acceptable for the expected small skill count.
- **`Game.DataAccess/Constants.cs:3`** is a pure const-bag declared as an instantiable `internal class`; make it `internal static class`.

scope: small

ginderjeremiah:
## Summary

Addresses the basket of small, low-risk backend cleanups tracked in #1143. Each is individually minor; bundled into one sweep.

### Allocation / perf nits
- **`AggregateRoot.DomainEvents`** — was `_domainEvents.AsReadOnly()`, allocating a fresh `ReadOnlyCollection` on every get (the dispatcher reads `Count` then `ToArray` per drain iteration). Now the wrapper is built once in the constructor; `ReadOnlyCollection` is a *live view* over the backing list, so it still reflects every `RaiseEvent`/`ClearEvents`.
- **`Roles.GetRoles()`** — was re-enumerating `Enum.GetValues()` and rebuilding `Role`s (with reflection-backed `ToString()`) on every call for intrinsic enum data. Now precomputed once as a `static readonly Role[]`; `GetRoles()` returns a fresh list copy so the shared set stays immutable to callers.
- **`Zones.GetDomainZone`** — was mapping a fresh `CoreZone` per call on the per-battle setup path, against the documented gameplay-read convention (every other read hands back shared pre-materialized instances). Introduces a `ZoneSnapshot` record (mirroring `EnemySnapshot`) bundling the entity list + pre-materialized `CoreZone`s so they swap atomically, and `GetDomainZone` now returns the shared instance. The `CoreZone` domain model is immutable (`init`-only, no mutable collections), so sharing is safe.

### Convention / dead code
- **`EnemyMapper.ToTemplate`** — removed the `.Where(es => es.SkillId >= 0 && es.SkillId < allSkills.Count)` filter that would *silently drop* a skill from an enemy loadout. Contiguity is asserted before templates build and `SkillId` is an FK into that set, so the filter could never validly exclude anything; a genuinely out-of-range id is now a loud index throw, matching the player-load loud-fail policy.
- **`Player.UnlockSkill`** — kept the linear unlocked-skill scan but documented why it's acceptable (skills grow far slower than items and aren't on the battle-start hot path the unlocked-item set is). Restructuring `Skills` into an id-keyed collection would ripple into mapping/serialization, out of scope for a minor cleanup.
- **`Game.DataAccess/Constants.cs`** — pure const-bag is now `internal static class`.

### Out of this PR
The Zones item in #1143 noted "coordinate with the Zones torn-read fix" — that fix already landed (each read captures `holder.Current` once), and this change keeps it: every method captures the snapshot once before the bounds-check + index.

## Test plan
- `Game.Core.Tests` — 761 passed.
- `Game.Application.Tests` (integration) — 508 passed, including a new `GetDomainZone_ReturnsSharedInstanceAndCorrectData` shared-instance test added alongside the existing reference-read cases.
- `Game.Api.Tests` — 603 passed (covers `GetRoles` via `AdminUsersControllerTests`).

Closes #1143

https://claude.ai/code/session_018oWmT6gggtbLSrEu8tQjJr

---
_Generated by [Claude Code](https://claude.ai/code/session_018oWmT6gggtbLSrEu8tQjJr)_

### DIFF
--- a/Game.DataAccess/Mapping/EnemyMapper.cs
+++ b/Game.DataAccess/Mapping/EnemyMapper.cs
@@ -2,6 +2,7 @@ using Game.Core;
 using Game.Core.Attributes;
 using Game.Core.Enemies;
 using Contracts = Game.Abstractions.Contracts;
+using CoreSkill = Game.Core.Skills.Skill;
 using EntityEnemy = Game.Infrastructure.Entities.Enemy;
 using EntitySkill = Game.Infrastructure.Entities.Skill;
 
@@ -57,16 +58,30 @@ namespace Game.DataAccess.Mapping
                         BaseAmount = ad.BaseAmount,
                         AmountPerLevel = ad.AmountPerLevel,
                     }).ToList(),
-                // The cached skill list is the zero-based, contiguous-id reference set (docs/backend.md
-                // → Reference Data), so a skill resolves by direct index instead of a per-call dictionary.
-                // A persisted EnemySkill.SkillId is FK-guaranteed in range against the contiguity-asserted
-                // skill set, so the index always resolves; an out-of-range id would be content-data
-                // corruption and throws loudly here rather than silently dropping the skill (the loud-fail
-                // policy the player-load path enforces, vs. the prior skip-if-missing filter).
                 AvailableSkills = entity.EnemySkills
-                    .Select(es => SkillMapper.ToCore(allSkills[es.SkillId]))
+                    .Select(es => ResolveSkill(entity, es.SkillId, allSkills))
                     .ToList(),
             };
+        }
+
+        /// <summary>
+        /// Resolves an enemy's available skill by direct index into the cached, zero-based, contiguous-id
+        /// skill set (docs/backend.md → Reference Data). A persisted <c>EnemySkill.SkillId</c> is
+        /// FK-guaranteed in range against the (contiguity-asserted) skill set, so this always resolves; an
+        /// out-of-range id can only be content-data corruption and fails loudly with a diagnosable message
+        /// naming the enemy and offending id — mirroring the player-load loud-fail policy rather than the
+        /// prior filter that silently dropped the skill from the loadout.
+        /// </summary>
+        private static CoreSkill ResolveSkill(EntityEnemy enemy, int skillId, IReadOnlyList<EntitySkill> allSkills)
+        {
+            if (skillId < 0 || skillId >= allSkills.Count)
+            {
+                throw new InvalidOperationException(
+                    $"Enemy {enemy.Id} ('{enemy.Name}') references a skill with Id {skillId} that does not " +
+                    "resolve against the skill catalog (a content-data mistake — a referenced id is out of range).");
+            }
+
+            return SkillMapper.ToCore(allSkills[skillId]);
         }
     }
 }
--- a/Game.DataAccess/Mapping/EnemyMapper.cs
+++ b/Game.DataAccess/Mapping/EnemyMapper.cs
@@ -59,10 +59,11 @@ namespace Game.DataAccess.Mapping
                     }).ToList(),
                 // The cached skill list is the zero-based, contiguous-id reference set (docs/backend.md
                 // → Reference Data), so a skill resolves by direct index instead of a per-call dictionary.
-                // The bounds check skips a malformed out-of-range ref, matching the prior skip-if-missing
-                // behaviour (a persisted EnemySkill.SkillId is FK-guaranteed in range, so it is defensive).
+                // A persisted EnemySkill.SkillId is FK-guaranteed in range against the contiguity-asserted
+                // skill set, so the index always resolves; an out-of-range id would be content-data
+                // corruption and throws loudly here rather than silently dropping the skill (the loud-fail
+                // policy the player-load path enforces, vs. the prior skip-if-missing filter).
                 AvailableSkills = entity.EnemySkills
-                    .Where(es => es.SkillId >= 0 && es.SkillId < allSkills.Count)
                     .Select(es => SkillMapper.ToCore(allSkills[es.SkillId]))
                     .ToList(),
             };
--- a/Game.Application.Tests/DataAccess/ReferenceDataSharedInstanceTests.cs
+++ b/Game.Application.Tests/DataAccess/ReferenceDataSharedInstanceTests.cs
@@ -181,6 +181,31 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task GetDomainZone_ReturnsSharedInstanceAndCorrectData()
+        {
+            int zoneId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                zoneId = (await TestDataSeeder.CreateZoneAsync(context, "Shared Zone", levelMin: 4, levelMax: 7)).Id;
+            }
+
+            await ReloadReferenceCachesAsync();
+
+            using var scope = CreateScope();
+            var zones = scope.ServiceProvider.GetRequiredService<IZones>();
+
+            var first = zones.GetDomainZone(zoneId);
+            var second = zones.GetDomainZone(zoneId);
+
+            // The optimization: repeated reads hand back the snapshot's pre-materialized instance, not a fresh map.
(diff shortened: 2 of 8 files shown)


Game.DataAccess/Mapping/EnemyMapper.cs
```diff
@@ -59,10 +59,11 @@ public static EnemyTemplate ToTemplate(EntityEnemy entity, IReadOnlyList<EntityS
...
                 // → Reference Data), so a skill resolves by direct index instead of a per-call dictionary.
-                // The bounds check skips a malformed out-of-range ref, matching the prior skip-if-missing
-                // behaviour (a persisted EnemySkill.SkillId is FK-guaranteed in range, so it is defensive).
+                // A persisted EnemySkill.SkillId is FK-guaranteed in range against the contiguity-asserted
+                // skill set, so the index always resolves; an out-of-range id would be content-data
+                // corruption and throws loudly here rather than silently dropping the skill (the loud-fail
+                // policy the player-load path enforces, vs. the prior skip-if-missing filter).
                 AvailableSkills = entity.EnemySkills
-                    .Where(es => es.SkillId >= 0 && es.SkillId < allSkills.Count)
                     .Select(es => SkillMapper.ToCore(allSkills[es.SkillId]))
```
ginderjeremiah: Nit (non-blocking): the comment frames this as "the loud-fail policy the player-load path enforces," but those two paths don't actually fail the same way. The player-load policy (per `docs/backend.md` → Reference Data) throws a *diagnosable* exception naming the player id, catalog, and missing id; `allSkills[es.SkillId]` here throws a bare `IndexOutOfRangeException` ("Index was outside the bounds of the array") with no enemy id or skill id in the message. Removing the silent-drop filter is the right call, but if you want it to truly match the cited policy, consider a bounds check that throws a contextual exception naming the enemy + offending `SkillId`. Otherwise, soften the comment to avoid implying parity with the player-load diagnostics. Either way, optional — this only fires on impossible (FK + contiguity-asserted) content corruption.

---

ginderjeremiah (commented):
## Review: Backend minor cleanups — allocation nits, dead defensive code, conventions (closes #1143)

**Verdict:** Approve (posted as a comment because I'm the same GitHub account as the PR author, so GitHub won't accept a gating review here) — every item in #1143 is addressed correctly and each change lands squarely on the documented conventions. Only one optional nit.

This is a tidy, well-scoped sweep. I verified each change against `docs/backend.md` and the surrounding code:

- **`AggregateRoot.DomainEvents`** — `ReadOnlyCollection` is a live view over `_domainEvents`, so building it once in the ctor still reflects every `RaiseEvent`/`ClearEvents`. Confirmed against `DomainEventDispatcher.DispatchAsync`, which reads `.Count` then `.ToArray()` once per drain iteration — exactly the repeated-get the cached wrapper removes. Correct.
- **`Roles.GetRoles()`** — precomputed `static readonly Role[]`, returning `[.. _roles]` per call so the shared set stays immutable to callers. Matches the intrinsic-reference-data convention. Correct.
- **`Zones.GetDomainZone` / `ZoneSnapshot`** — the new `ZoneSnapshot(Entities, CoreZones)` record mirrors `EnemySnapshot` faithfully: both halves are built in `BuildSnapshotAsync` and published in one atomic swap, so a reader can never see a new entity list against stale/null domain models. `GetDomainZone` now hands back the shared pre-materialized `CoreZone`, satisfying the "gameplay reads return shared, pre-materialized instances" rule in `docs/backend.md`. `Game.Core.Zones.Zone` is `init`-only with no mutable collections, so sharing is safe. The per-call single-snapshot capture preserves the torn-read fix. Correct.
- **`EnemyMapper.ToTemplate`** — dropping the `.Where(...)` filter is right: `EnemiesCacheHolder` asserts skill contiguity and `SkillId` is an FK into that set, so the filter could only ever silently drop a valid skill. Now it indexes directly. See the inline nit on the diagnostic framing.
- **`Player.UnlockSkill`** / **`Constants`** — comment-only justification and `internal static class` respectively; both match the issue's asks.

### Nits
- `Game.DataAccess/Mapping/EnemyMapper.cs:67` — the new behavior throws a bare `IndexOutOfRangeException` on an out-of-range `SkillId`, which doesn't carry the enemy/skill diagnostic context the cited player-load loud-fail policy does. Optional; details inline.

### Test coverage
- New `GetDomainZone_ReturnsSharedInstanceAndCorrectData` integration test asserts the shared-instance optimization (`Assert.Same`) and correct data, alongside the existing reference-read cases — good, and consistent with how the enemy/item shared-instance reads are covered.
- `Roles.GetRoles` is exercised via `AdminUsersControllerTests` per the test plan.
- The remaining changes are structural (cached wrapper, `internal static`, comment) and adequately covered by existing suites. The only thing not directly tested is the new `EnemyMapper` throw-on-out-of-range path, but since that only fires on impossible (FK + contiguity-asserted) content corruption, a dedicated test would be gold-plating — fine to skip. Coverage is sufficient.

### What's good
- Each change is justified inline with a reference to the convention it satisfies, the `ZoneSnapshot` atomic-publish reasoning is spelled out clearly, and the scope stays disciplined (the `Player.Skills` id-keying and the Zones torn-read coordination are both correctly deferred/noted rather than scope-crept).

CI is still running at review time; this approval is contingent on it going green. Not auto-merging given the open nit — happy to merge once you've decided on the nit and CI is green.

---

Game.DataAccess/Mapping/EnemyMapper.cs
ginderjeremiah: Good call — went with the contextual-throw option rather than softening the comment. `ToTemplate` now resolves each skill through a bounds-checked `ResolveSkill` helper that throws an `InvalidOperationException` naming the enemy (id + name) and the offending `SkillId`, so a content-data corruption is diagnosable from logs the same way the player-load path is. Pushed in aaa1636.

---

PR #1178, opened 2026-06-23
Make next-battle prefetch best-effort so a win can't be double-credited (#1169)

### LINKED ISSUE #1169: Battle win persisted before session save — a next-battle prefetch failure can strand the resolved PlayerState (double-credit window)
Found during a code-health review. A narrow ordering gap in the victory commands that can leave the in-memory session state inconsistent with the durably-credited battle, opening a re-credit window on reconnect.

## Problem

In `DefeatEnemy.HandleExecuteAsync` (`Game.Api/Sockets/Commands/DefeatEnemy.cs:38-56`):

1. `EndBattleVictory` already (a) granted exp + recorded the result through the Player write-behind aggregate, and (b) cleared the active battle and set the cooldown on the in-memory `PlayerState`.
2. `PrepareNextIdleBattle(...)` then runs to prefetch/bundle the next idle enemy.
3. **Only afterward** is `context.Session.SavePlayerState()` called.

If step 2 throws (e.g. an unexpected zone-resolution failure after `EnsureViableZone`), the win is durably persisted but the cleared-battle/cooldown `PlayerState` is never written to the session cache. On reconnect the stale session still shows the resolved battle as active, and the first `StartBattle` will `AbandonBattle` it again — re-crediting a battle that was already paid out. Double-credit is exactly the class the battle anti-cheat is otherwise careful to prevent.

The window is small (the prefetch rarely throws after `EnsureViableZone`), but the failure mode is durable double-credit.

## Fix

Persist the post-resolution `PlayerState` (battle-cleared + cooldown) **before** the prefetch, or treat the prefetch as best-effort (catch + log, still return without the bundled next enemy) so a prefetch failure can't strand the resolved state. The client already round-trips `NewEnemy` when no next enemy is bundled.

Check the sibling victory/loss paths for the same ordering — `BattleLost.cs` and the boss-victory branch should be audited together.

scope: small

ginderjeremiah:
## Summary

Closes #1169.

`DefeatEnemy` and `BattleLost` durably credit the resolved battle (exp + statistics/challenges, via the player write-behind aggregate) and clear the in-flight battle on the **in-memory** `PlayerState`, but only persist that cleared `PlayerState` to the session cache **after** prefetching the next idle battle. If the prefetch threw — e.g. an unexpected zone-resolution failure after `EnsureViableZone` — the win/loss was durably persisted while the cleared-battle/cooldown `PlayerState` was never written to the session. On reconnect the stale session still showed the resolved battle as active, so the first `StartBattle` would `AbandonBattle` it again and **re-credit an already-paid-out battle** — durable double-credit, exactly the class the battle anti-cheat otherwise prevents.

## How it's fixed

- Added `BattleService.TryPrepareNextIdleBattle` — a best-effort wrapper around `PrepareNextIdleBattle` that catches any prefetch failure, logs it, and returns `null` instead of throwing.
- `DefeatEnemy` and `BattleLost` now call it, so `context.Session.SavePlayerState()` **always** runs after the resolved (battle-cleared + cooldown) state. A prefetch failure can no longer strand that state; the response simply carries no bundled `NextEnemy`/`NextZoneId` and the client round-trips `NewEnemy` (already the contract — both fields are nullable and documented as "null when no next battle was prepared").

## Sibling-path audit (as the issue asked)

- **`BattleLost`** — same ordering bug (a re-abandon re-records the loss completion / re-counts challenges). Fixed the same way.
- **Boss-victory branch** (`DefeatEnemy`, `wasBossBattle`) — skips the prefetch entirely, so it was never exposed to this window. Left unchanged.
- **`ChallengeBoss`** — only *starts* a battle (no prior win to credit), so no double-credit window. Unchanged.

Why best-effort over "save before the prefetch": `SavePlayerState` is a fire-and-forget Redis write, and the happy-path save after the prefetch must persist the *prefetched* battle for its later anti-cheat replay. Saving both before and after would double the session write on the idle hot path; the best-effort catch keeps a single write on the happy path and only degrades (an extra `NewEnemy` round-trip) on the rare failure.

## Test plan

- Added `BattleServiceIntegrationTests.TryPrepareNextIdleBattle_PrefetchFails_SwallowsAndLeavesResolvedStateCleared`: wins/resolves a battle, forces an out-of-range `CurrentZoneId` so the prefetch's zone resolution throws, and asserts the wrapper swallows it (`null`) and leaves the resolved state cleared — so the command's `SavePlayerState` persists the credited outcome rather than a stale active battle.
- `BattleServiceIntegrationTests` (38) green, including the existing prefetch/abandon coverage.
- `DefeatEnemySocketTests`, `BattleLostSocketTests`, `NewEnemySocketTests` green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Krz1ghABJZVNiCVjCZthdd)_

### DIFF
--- a/Game.Api/Sockets/Commands/BattleLost.cs
+++ b/Game.Api/Sockets/Commands/BattleLost.cs
@@ -38,7 +38,11 @@ namespace Game.Api.Sockets.Commands
 
                 // A boss loss returns to the idle farm, so prefetch and bundle the next idle battle — letting
                 // the client begin it the instant the post-loss cooldown elapses, hiding the NewEnemy round-trip.
-                var next = await _battleService.PrepareNextIdleBattle(player, state, cancellationToken);
+                // The prefetch is best-effort: the loss is already durably recorded and its battle-cleared
+                // PlayerState is saved below regardless, so a prefetch failure must not strand that resolved
+                // state (which a reconnect would re-abandon, re-recording the completion). The client
+                // round-trips NewEnemy when none is bundled.
+                var next = await _battleService.TryPrepareNextIdleBattle(player, state, cancellationToken);
 
                 context.Session.SavePlayerState();
 
@@ -46,8 +50,8 @@ namespace Game.Api.Sockets.Commands
                 return Success(new BattleLostResponse
                 {
                     Cooldown = (state.EnemyCooldown - now).TotalMilliseconds,
-                    NextEnemy = EnemyInstance.FromSource(next),
-                    NextZoneId = player.CurrentZoneId,
+                    NextEnemy = next is not null ? EnemyInstance.FromSource(next) : null,
+                    NextZoneId = next is not null ? player.CurrentZoneId : null,
                 });
             }
             else
--- a/Game.Api/Sockets/Commands/DefeatEnemy.cs
+++ b/Game.Api/Sockets/Commands/DefeatEnemy.cs
@@ -43,14 +43,20 @@ namespace Game.Api.Sockets.Commands
                     player.Id, rewards.ExpReward);
 
                 // Prefetch and bundle the next idle battle so the client can begin it the instant the
-                // post-battle cooldown elapses, hiding the NewEnemy round-trip under the cooldown.
+                // post-battle cooldown elapses, hiding the NewEnemy round-trip under the cooldown. The prefetch
+                // is best-effort: the victory is already durably credited and its battle-cleared PlayerState is
+                // saved below regardless, so a prefetch failure must not strand that resolved state (which a
+                // reconnect would re-abandon and re-credit). The client round-trips NewEnemy when none is bundled.
                 EnemyInstance? nextEnemy = null;
                 int? nextZoneId = null;
                 if (!wasBossBattle)
                 {
-                    var next = await _battleService.PrepareNextIdleBattle(player, state, cancellationToken);
-                    nextEnemy = EnemyInstance.FromSource(next);
-                    nextZoneId = player.CurrentZoneId;
+                    var next = await _battleService.TryPrepareNextIdleBattle(player, state, cancellationToken);
+                    if (next is not null)
+                    {
+                        nextEnemy = EnemyInstance.FromSource(next);
+                        nextZoneId = player.CurrentZoneId;
+                    }
                 }
 
                 context.Session.SavePlayerState();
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: best-effort next-battle prefetch so a win/loss can't be double-credited (#1169)

**Verdict:** Approve _(submitted as a comment — I'm the PR author, so GitHub won't accept a gating approve on my own PR)_ — the fix matches the issue's recommended approach, audits the sibling paths correctly, and the contract change is client-safe.

### Correctness

- The ordering bug is resolved exactly as the issue suggested: `TryPrepareNextIdleBattle` catches + logs and returns `null`, so `context.Session.SavePlayerState()` now always runs against the resolved (battle-cleared + cooldown) state. A prefetch throw can no longer strand it, closing the durable double-credit window on reconnect.
- Sibling audit is right: `BattleLost` had the identical bug and is fixed the same way; the boss-victory branch (`wasBossBattle`) skips the prefetch entirely so it was never exposed; `ChallengeBoss` only starts a battle (no prior credit), so neither needed changing.
- Contract change is safe. `BattleLost` now returns nullable `NextZoneId`/`NextEnemy` where it previously always set `NextZoneId`. The frontend (`enemy-manager.ts` `capturePreparedBattle`) already returns `undefined` when `nextEnemy` is falsy and round-trips `NewEnemy` — so a null bundle degrades gracefully, exactly as `DefeatEnemy`'s boss branch already did.

### Design

- Best-effort over "save before the prefetch" is the correct call and well-justified in the PR/doc-comment: `SavePlayerState` is a fire-and-forget Redis write and the happy-path save must persist the *prefetched* battle for its later anti-cheat replay, so saving twice would double the session write on the idle hot path. Single happy-path write, graceful degradation only on the rare failure.
- Layering is clean — the best-effort orchestration lives in the `Game.Application` service and the commands just call it.

### Test coverage

- `TryPrepareNextIdleBattle_PrefetchFails_SwallowsAndLeavesResolvedStateCleared` exercises a genuine failure path: an out-of-range `CurrentZoneId` makes `EnsureViableZone` pass it through (out-of-range is treated as corruption, not relocation) and the downstream `GetDomainZone(999)` throws *after* `EndBattleVictory` has credited the win. It asserts the swallow (`null`) and that the resolved state stays cleared — the precise invariant the command's `SavePlayerState` depends on. Adequate for a small-scope fix; the command-level mapping is a thin null-check covered by the existing socket suites.

### Note (non-blocking)

- The broad `catch (Exception)` will also swallow `OperationCanceledException` from the token. That's defensible — even arguably desirable — here, since the entire point is to never let the prefetch abort the path to saving the resolved state, and it's documented. No change requested; just flagging the intent so a later edit doesn't "tighten" it to rethrow on cancellation and silently reopen the stranding window.

Clean, well-scoped, well-documented fix. Approving; will merge once CI is green.

---

PR #1179, opened 2026-06-23
Serialize skills-loadout commits and derive equipped from the player manager

### LINKED ISSUE #1171: Skills loadout: serialize optimistic commits + re-sync `equipped` from the player manager
Found during a code-health review. Two related correctness gaps in the Skills loadout screen — one a real user-facing data-loss race, one latent.

## 1. `commit` has no single-flight guard — overlapping edits can silently revert (bug)

`SkillsView.commit` (`UI/src/routes/game/screens/skills/skills-view.svelte.ts:437-447`) is fire-and-forget (`void this.commit(next)` from `toggle`/`swapInto`/`reorder`). It captures `previous = this.equipped`, optimistically applies `next`, then awaits `SetSelectedSkills`. If a second edit lands before the first resolves, the two commits overlap: the first's `previous` snapshot is now stale, so if that first persist **fails**, its rollback restores a baseline captured *before* the second (successful) edit — discarding the second change on both `this.equipped` and `playerManager.setSelectedSkills`, desyncing the manager and the live battle.

This is the exact race the inventory manager already solved with its `serialize()` mutation queue (`UI/src/lib/engine/inventory-manager.ts:325`), and it contradicts the documented "one atomic set-loadout write" intent. Rapid equip/reorder clicks on a flaky connection trigger it.

**Fix:** serialize loadout commits (chain on a `lastOperation` promise like the inventory manager), or guard with an in-flight flag that queues only the latest edit.

## 2. `equipped` never re-syncs from the player manager after construction (latent)

`equipped` is seeded once from `playerManager.selectedSkills` in the constructor (`:144-154`); `syncFromPlayer()` is never called again and no `$effect` re-invokes it. The catalogue, `unlockedIds`, and `innateSkills` *are* reactive, but `equipped` is not. Today the screen is the only loadout writer while open, so it holds — but any external loadout change (a future auto-equip-on-unlock, server reconciliation, the `ChallengeCompleted` skill-unlock push) leaves the screen showing a stale loadout while the battle uses the new one, and the innate-grant duplicate flag misreads.

**Fix:** reconcile `equipped` from a `$derived` baseline when there's no pending optimistic edit.

scope: small

ginderjeremiah:
## Summary

Fixes the two correctness gaps in the Skills loadout screen reported in #1171.

### 1. Overlapping optimistic commits could silently revert a later edit (bug)

`SkillsView.commit` was fire-and-forget: it captured a per-call `previous = this.equipped` snapshot, optimistically applied the new loadout, then awaited `SetSelectedSkills`. If a second edit landed before the first persist resolved, the first commit's `previous` was already stale — so if that first persist **failed**, its rollback restored a baseline captured *before* the second (successful) edit, discarding it on both the view and the player manager.

Commits are now **serialized** on a `lastCommit` promise chain (the same pattern the inventory manager's `serialize()` already uses), so persists apply in edit order and rollback baselines never interleave. A failed persist rolls back **only when it was the latest edit issued** (tracked via a monotonic `commitSeq`), and it reverts to the last *server-confirmed* loadout (`committed`) rather than a stale per-call snapshot. An earlier failure superseded by a later successful edit no longer reverts or toasts.

### 2. `equipped` never re-synced from the player manager after construction (latent)

`equipped` was seeded once in the constructor and never reconciled, so an external loadout change (a future auto-equip-on-unlock, server reconciliation, the `ChallengeCompleted` skill-unlock push) left the screen showing a stale loadout. It's now a `$derived` that returns the pending optimistic edit while a commit is in flight, otherwise the player manager's committed loadout — so external changes are reflected once no edit is pending. Because the production `playerManager` is statified, this propagates reactively while idle.

## Notes

- The optimistic write to `equipped`/`playerManager` stays **synchronous**, so rapid edits read fresh state and queue rather than racing. Only the socket dispatch now lands on a microtask (consistent with the inventory manager's serialized mutations) — a couple of existing tests were updated to `await` the dispatch accordingly.
- Frontend-only; this is loadout persistence, not battle-tick logic, so no backend parity mirror is required.

## Test plan

- [x] New unit tests: an earlier overlapping commit failing keeps the later successful edit (no spurious revert/toast); the latest edit failing reverts to the last *confirmed* loadout (not the original); the screen reconciles with the manager once no edit is pending.
- [x] Existing skills-screen tests updated for the serialized (microtask) dispatch.
- [x] `npm run check` (svelte-check) — 0 errors.
- [x] `eslint .` — clean.
- [x] Full unit suite — 2438 passing.

Closes #1171

---
_Generated by [Claude Code](https://claude.ai/code/session_016rnXMdaNprh7BPJiGwh8Vt)_

### DIFF
--- a/UI/src/routes/game/screens/skills/skills-view.svelte.ts
+++ b/UI/src/routes/game/screens/skills/skills-view.svelte.ts
@@ -457,7 +457,7 @@ export class SkillsView {
 		this.pendingLoadout = next;
 		playerManager.setSelectedSkills(next);
 		const seq = ++this.commitSeq;
-		this.lastCommit = this.lastCommit.then(async () => {
+		const result = this.lastCommit.then(async () => {
 			const response = await apiSocket.sendSocketCommand('SetSelectedSkills', next);
 			const isLatest = seq === this.commitSeq;
 			if (response.error) {
@@ -473,5 +473,8 @@ export class SkillsView {
 				}
 			}
 		});
+		// Keep the queue tail always-fulfilled: a rejecting callback would otherwise make every later
+		// commit skip its persist and raise an unhandled rejection (mirrors the inventory manager's serialize).
+		this.lastCommit = result.catch(() => undefined);
 	}
 }
--- a/UI/src/routes/game/screens/skills/skills-view.svelte.ts
+++ b/UI/src/routes/game/screens/skills/skills-view.svelte.ts
@@ -125,9 +125,16 @@ export function zoneSpawnLevel(zone: IZone): number {
 export class SkillsView {
 	/** The inspected skill (rail/band selection). */
 	selectedId = $state<number>(-1);
-	/** Working equipped loadout, ordered by priority. Optimistically updated, then
-	 *  persisted via {@link commit}; reverted if the persist fails. */
-	equipped = $state<number[]>([]);
+	/** Pending optimistic loadout while a commit is in flight (or after the latest commit failed without
+	 *  a newer edit); null when the screen mirrors the player manager. {@link equipped} reads through it. */
+	private pendingLoadout = $state<number[] | null>(null);
+	/** Last loadout the server confirmed — the rollback target when the latest commit fails. */
+	private committed: number[] = [];
+	/** Tail of the optimistic-commit queue; each commit chains off it so persists apply in edit order
+	 *  and rollback baselines never interleave (mirrors the inventory manager's serialized mutations). */
+	private lastCommit: Promise<unknown> = Promise.resolve();
+	/** Monotonic edit counter; a failed persist rolls back only when it was the latest edit issued. */
+	private commitSeq = 0;
 	search = $state('');
 	sort = $state<SkillSort>('dps');
 	filterAttributes = $state<EAttribute[]>([]);
@@ -145,13 +152,19 @@ export class SkillsView {
 		this.syncFromPlayer();
 	}
 
-	/** (Re)seed the working loadout + selection from the player manager. */
+	/** (Re)seed the committed baseline + inspector selection from the player manager. */
 	syncFromPlayer(): void {
-		this.equipped = [...playerManager.selectedSkills];
+		this.committed = [...playerManager.selectedSkills];
+		this.pendingLoadout = null;
 		if (this.selectedId < 0) {
 			this.selectedId = this.equipped[0] ?? staticData.skills?.[0]?.id ?? -1;
 		}
 	}
+
+	/** Working equipped loadout, ordered by priority: the pending optimistic edit while a commit is in
+	 *  flight, otherwise the player manager's committed loadout — so an external loadout change (e.g. a
+	 *  challenge skill unlock or server reconciliation) is reflected once no edit is pending. */
+	readonly equipped = $derived(this.pendingLoadout ?? [...playerManager.selectedSkills]);
 
 	/** The loadout cap (number of equip slots) — the single generated game constant. */
 	readonly cap = MAX_SELECTED_SKILLS;
@@ -368,9 +381,9 @@ export class SkillsView {
 			return;
 		}
 		if (this.isEquipped(id)) {
-			void this.commit(this.equipped.filter((x) => x !== id));
+			this.commit(this.equipped.filter((x) => x !== id));
 		} else if (this.equipped.length < this.cap) {
-			void this.commit([...this.equipped, id]);
+			this.commit([...this.equipped, id]);
 		} else {
 			this.pendingSwap = id;
 		}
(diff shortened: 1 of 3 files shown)


UI/src/routes/game/screens/skills/skills-view.svelte.ts
```diff
@@ -433,16 +446,32 @@ export class SkillsView {
...
+		this.pendingLoadout = next;
 		playerManager.setSelectedSkills(next);
-		const response = await apiSocket.sendSocketCommand('SetSelectedSkills', next);
-		if (response.error) {
-			this.equipped = previous;
-			playerManager.setSelectedSkills(previous);
-			toastError('Your loadout could not be saved. Please try again.');
-		}
+		const seq = ++this.commitSeq;
+		this.lastCommit = this.lastCommit.then(async () => {
```
ginderjeremiah: Should-fix (non-blocking): this inline chain drops the rejection guard the inventory manager's `serialize` deliberately keeps. There, the tail is stored as `this.lastOperation = result.catch(() => undefined)` and the previous promise is consumed with `.then(operation, operation)`, with a comment noting it stops "one operation's rejection from breaking the chain or surfacing as an unhandled rejection on the shared promise."

Here, `this.lastCommit = this.lastCommit.then(async () => { ... })` chains only the fulfilled path. If any callback ever rejects, `lastCommit` becomes a rejected promise, every subsequent `commit` skips its `.then` callback (so loadout edits silently stop persisting for the rest of the screen's life) and an unhandled rejection fires. In practice `sendSocketCommand`/`getResponse` is documented to never reject (errors come back via `response.error`), so this is latent — but the only post-`await` code (`playerManager.setSelectedSkills`, `toastError`) isn't guaranteed throw-free, and you're explicitly mirroring a pattern whose whole point includes this guard.

Two options, either is fine: append `.catch(() => undefined)` when assigning `this.lastCommit` (cheap, local), or extract the inventory manager's `serialize` into a shared helper and route both through it (DRY — there are now two hand-rolled copies of the same serialized-optimistic-mutation pattern). The latter is arguably a small follow-up rather than this PR's scope.

---

ginderjeremiah (commented):
## Review: serialize Skills-loadout commits + derive `equipped` from the player manager (#1171)

**Verdict: Approve** (posted as a comment because the review account is also the PR author, so GitHub won't accept a gating event) — both correctness gaps from #1171 are fixed cleanly and the new logic holds up under the interleavings that matter. One non-blocking should-fix below.

I traced the concurrency logic against the scenarios in the issue:

- **Serialization / ordering** — each `commit` chains its persist off `lastCommit`, so `SetSelectedSkills` is dispatched in edit order rather than racing. The optimistic write to `pendingLoadout`/`playerManager` stays synchronous, so rapid edits read fresh state and queue. ✓
- **Rollback discipline** — `commitSeq` ensures only the *latest* edit's failure rolls back, and it reverts to `committed` (last server-confirmed) rather than a stale per-call snapshot. Walked A-fails/B-succeeds, A-succeeds/B-fails, and both-fail by hand; the manager always ends on the latest confirmed loadout and a superseded earlier failure neither reverts nor toasts. ✓
- **Reconcile** — `equipped` is a `$derived` reading `pendingLoadout ?? manager`. When the latest persist settles, `pendingLoadout = null` (a reactive `$state` write) re-triggers the derived, which re-reads the manager — so an external change is picked up once no edit is pending. The note that production propagation while idle relies on the statified manager is accurate.

### Should-fix (non-blocking)
- `skills-view.svelte.ts:460` — the inline `lastCommit` chain omits the rejection guard the inventory manager's `serialize` deliberately keeps (`.then(operation, operation)` + `lastOperation = result.catch(() => undefined)`). Latent today because the socket transport never rejects, but it's a real footgun on a shared promise and diverges from the very pattern this mirrors. Add a `.catch(() => undefined)` on the tail, or extract a shared `serialize` helper (DRY) as a small follow-up. Details inline.

### Test coverage
Solid and matches the project's frontend rules (unit tests mirroring the source path). The three new cases pin exactly the behaviors that were broken/latent — overlapping-failure-then-success keeps the winning edit with no spurious toast, latest-failure reverts to the *confirmed* loadout, and the screen reconciles once the queue drains. Existing tests were correctly updated to `await` the now-microtask dispatch, and `EquippedBand` switched off assigning the now-readonly `equipped` to driving `toggle(2)`. Frontend-only loadout persistence, so no backend battle-parity mirror is required. ✓

### What's good
Clear severity-aware rollback model, honest comments (including the production-vs-test reconcile caveat), and tests that target the actual races instead of just happy paths. Nice, contained fix.

---

UI/src/routes/game/screens/skills/skills-view.svelte.ts
ginderjeremiah: Fixed in 34e38e3 — the tail is now stored as `this.lastCommit = result.catch(() => undefined)`, so a rejecting callback can't poison the queue or raise an unhandled rejection, matching the inventory manager's `serialize`.

Filed the DRY extraction (a shared `serialize` helper for both call sites) as a follow-up: #1183.

---

PR #1180, opened 2026-06-23
Fix Admin Workbench editor correctness issues (index-keyed rows, stale store, NumInput coercion, diffing)

### LINKED ISSUE #1172: Admin Workbench editor correctness: index-keyed table rows, stale store on partial save failure, NumInput coercion
Found during a code-health review — a basket of correctness issues in the entity-driven admin Workbench. The two marked **bug** are data-integrity issues; the rest are latent/minor. Grouped since they live in the same save/edit layer.

## 1. `SectionTable` keys rows by array index — dirty edges & input state corrupt on mid-list delete (bug)

`UI/src/routes/admin/workbench/components/SectionTable.svelte:22-46` renders `{#each rows as row, i (i)}` and matches the baseline positionally (`isNewRow = i >= baseRows.length`, `baseRow = baseRows?.[i]`). `removeRow(i)` does `list.splice(i, 1)`, shifting every later row's index. After a non-last removal:

- each shifted row is compared against a *different* baseline row, so clean rows flash "modified" and changed rows can read "clean";
- because the `{#each}` key is the positional index, Svelte reuses the `<tr>`/`NumInput` DOM nodes across what are now logically different rows, so `NumInput`'s local `inputText` `$state` can bind to the wrong row.

Affects skill effects, item-mod slots, and the attribute tables. **Fix:** key by a stable per-row identity (`row.id` for surrogate-id collections, `row.attributeId`/`row.zoneId` for attribute-keyed ones) and diff the baseline by matching that identity rather than by index.

## 2. A child-saver failure after the primary batch leaves the store stale → duplicate re-adds on retry (bug)

`UI/src/routes/admin/workbench/save-helpers.ts:48-78` + `entity-store.svelte.ts:159-177`: `persistEntity` posts the identity Add/Edit/Delete batch, then runs child savers sequentially. If a child saver throws (e.g. `SetEnemySpawns` 500s) **after** the primary batch already committed the new identity rows, the exception propagates to `EntityStore.save`, which catches and toasts — but never updates `this.base`/`this.items`. The added records are still `added` locally while the server already has them, so the next Save re-Adds duplicates (the k-th-add → k-th-new-id zip now resolves against stale `existingIds`).

**Fix:** on any failure inside `save()`, force a `refresh()` to re-seed the baseline from server truth before surfacing the error, so a retry diffs against what actually persisted.

## 3. `NumInput` coerces partial/empty input to `0` straight into the store (minor)

`UI/src/routes/admin/workbench/components/NumInput.svelte:61-64`: clearing a numeric field to retype it emits `onChange(0)` on every intermediate token (`''`, `'-'`, `'.'`), which `setCell`/`patch` writes immediately to the draft record. Navigating away mid-edit persists `0` rather than the original value, and an empty field is indistinguishable from an intentional `0`. Display is fine (the `text` derived shows local `inputText` while focused); only the stored value transiently degrades. **Fix:** emit only on a fully-parseable value; surface empty as a validation state rather than coercing.

## 4. Structural comparison via `JSON.stringify` is sensitive to `null`-vs-absent & key order (minor/latent)

`entity-store.svelte.ts:8`, `save-helpers.ts:4`: `recordsEqual`/`listsEqual` are `JSON.stringify(a) === JSON.stringify(b)`, which drops `undefined`-valued keys and is key-order-sensitive. If the server ever returns `null` for an unset optional where the locally-constructed record leaves the key absent, a freshly-loaded untouched record reads as `modified` (phantom dirty dot). Works today because the round-trip shapes align, but fragile. **Fix:** normalize to a canonical shape (or compare field-by-field) before diffing.

scope: medium

ginderjeremiah:
## Summary

Closes #1172. Fixes the basket of correctness issues in the entity-driven admin Workbench save/edit layer — the two **bug** items (data-integrity) plus the two **minor/latent** ones, since they all live in the same diff/save layer.

### 1. `SectionTable` keyed & diffed rows by array index (bug)

Rows were rendered with `{#each rows as row, i (i)}` and matched to their baseline positionally. After a non-last `removeRow`, every later row's index shifted, so clean rows flashed "modified" and the index-based key reused `<tr>`/`NumInput` DOM state across logically different rows.

**Fix:** rows are now keyed and matched to their baseline by a **stable per-row identity**. A new `rowKey` on `TableSectionConfig` names the identity field — `id` for surrogate-id collections, the natural key (`attributeId`/`enemyId`/`zoneId`) for the rest. New surrogate rows get a **unique negative client id** so the key is collision-free; the persist layer normalises it back to `0` on the Add payload.

### 2. Stale store on a partial save failure → duplicate re-adds (bug)

If a child saver threw after the identity batch committed, `save()` toasted but never updated `base`/`items`, so a retry re-Added duplicates. **Fix:** `save()` now re-seeds from server truth via `refresh()` on failure (keeping local edits only if the recovery refresh itself fails).

### 3. `NumInput` coerced partial/empty input to `0` (minor)

Clearing a field emitted `onChange(0)`, persisting `0` on navigate-away. **Fix:** only commit a fully-parseable number; empty/in-progress leaves the stored value untouched (blur restores it).

### 4. `JSON.stringify` structural comparison (minor/latent)

`recordsEqual`/`listsEqual` were key-order-sensitive and treated `null`-vs-absent optionals as different (phantom dirty dots). **Fix:** a shared `canonicalEqual` (keys sorted, null/undefined/absent collapsed) backs both.

## Test plan

- [x] New/updated unit tests for all four fixes (SectionTable identity diffing + unique negative ids, NumInput no-coercion, EntityStore refresh-on-failure, canonicalEqual + negative-id normalisation).
- [x] Full unit suite — 2445 passing.

Frontend-only; no battle-tick logic touched, so no backend parity mirror required.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GDcquLVLDeK5F37WkQUQCz)_


UI/src/routes/admin/workbench/entity-store.svelte.ts
```diff
@@ -168,8 +169,19 @@ export class EntityStore<T extends Identified> {
...
 		} catch (ex) {
-			// Without this, a failed persist reset only the saving flag and the
-			// user was left with unsaved edits and no indication anything broke.
+			// A partial failure (e.g. the identity Add/Edit batch committed but a child saver then
+			// threw) leaves our baseline behind the server: the persisted adds are still local
+			// "added" records, so a naive retry would re-Add duplicates. Re-seed from server truth
+			// so a retry diffs against what actually persisted. If even the refresh fails, keep the
+			// local edits rather than blanking the screen — the toast below still flags the failure.
+			try {
+				const fresh = await this.config.refresh();
```
ginderjeremiah: **Should-fix:** This re-seeds from server truth on *every* failure, including a **total** failure where nothing committed (e.g. `postPrimary` itself 500s/times out before any write). In that case `refresh()` returns the unchanged server state and silently discards **all** the user's unsaved edits — a regression from the prior behavior, which kept the edits intact so the user could just retry (per the old comment this block replaced).

The re-seed is only correct when the primary batch actually committed (issue #1172's scenario: primary Add/Edit succeeded, a child saver then threw). For the more common transient-failure-on-the-primary-call path, blowing away the in-progress edits behind a toast is a worse outcome than the duplicate-add bug this fixes.

Suggest distinguishing the two: e.g. have `persistEntity` own the refresh-reseed only when it knows the primary committed (catch the child-saver failure, reseed, rethrow), and let a pre-commit `postPrimary` failure propagate without reseeding so `save()` leaves the edits in place. Not blocking — data stays consistent either way and the error is surfaced — but worth addressing or consciously deciding the edit-loss is acceptable.

---

ginderjeremiah (commented):
## Review: Admin Workbench editor correctness fixes (closes #1172)

**Verdict:** Approve *(submitted as a comment — GitHub won't accept a gating verdict on one's own PR)* — all four issues are correctly addressed with thorough, well-targeted tests. One should-fix on the save-failure recovery path is worth resolving or consciously declining; it doesn't block since data stays consistent.

### Should-fix
- `entity-store.svelte.ts:177-184` — `save()` re-seeds via `refresh()` on **any** failure, which discards all unsaved edits on a *total* failure where nothing committed (e.g. `postPrimary` 500s before any write). The re-seed is only needed when the primary batch actually committed (issue #1172's partial-failure case). Inline comment has a suggested split. Not blocking — consistency is preserved and the error is toasted.

### Nits
- `SectionTable.svelte:46` — `dirty={!isNewRow && !!baseRow && !recordsEqual(...)}`: `!isNewRow` already implies `baseRow !== undefined` (since `isNewRow = baseRow === undefined`), so the `!!baseRow` guard is redundant. Harmless, but a later reader may puzzle over it.
- `save-helpers.ts:229` — `modSlotChanges` normalizes a negative client id to `0` exactly like `skillEffectChanges`, but only the latter has a direct normalization test. The shared `surrogateIdChanges` path is covered, so this is a minor gap, not a hole.

### Test coverage
Strong and matches the project's rules:
- SectionTable: identity-vs-index diffing after a mid-list delete, and unique negative ids across multiple unsaved rows.
- NumInput: no coercion-to-0 on clear, and in-progress text preserved.
- EntityStore: refresh-on-failure re-seeding, plus the refresh-also-fails fallback keeping edits.
- save-helpers: `canonicalEqual` (key order, null/undefined/absent, 0-vs-absent, nested), and negative-id normalization.

Frontend-only with no battle-tick logic touched, so no backend parity mirror is required — correct call.

### What's good
- The stable-identity `rowKey` design is clean and consistently applied across all eight table sections, with the natural key for relation/attribute tables and a collision-free unique-negative id for surrogate collections; the persist layer normalizing `id <= 0` back to `0` keeps the wire contract intact.
- `canonicalEqual` cleanly retires the brittle `JSON.stringify` compare and kills the phantom-dirty class of bugs (null-vs-absent, key order) in one shared helper.
- Comments explain the *why* (index-shift corruption, partial-save staleness) without over-narrating — genuinely useful.

---

ginderjeremiah:
Addressed the should-fix in d88c205: the save-failure recovery no longer discards edits on a total failure.

- `persistEntity` now distinguishes **pre-commit** from **post-commit** failures via a `committed` flag, rethrowing post-commit failures as a new `PersistFailedError` and letting pre-commit failures (e.g. `postPrimary` throwing before any write) propagate raw.
- `EntityStore.save` re-seeds from server truth **only** on `PersistFailedError`; a pre-commit failure leaves the user's edits in place for a clean retry.

Also folded in the two nits: dropped the redundant `!!baseRow` guard in `SectionTable`, and added a direct negative-id-normalization test for `modSlotChanges`.

New tests cover both failure paths (pre-commit propagates raw / keeps edits; post-commit wraps + re-seeds, incl. the refresh-also-fails fallback). Full unit suite green (2450).

---

PR #1185, opened 2026-06-23
Game.Api polish: unknown-command rejection, stale comments, Retry-After rounding (#1144)

### LINKED ISSUE #1144: Game.Api polish: unknown-command fault classification, stale endpoint comments, Retry-After rounding
Found during a code-health review — low-severity correctness/cleanup polish in the API layer. The socket serialization, auth, and drain paths were reviewed and found sound; these are the rough edges.

## 1. Unknown client command name is treated as an internal fault
`Game.Api/Sockets/SocketHandler.cs:228-235` (`HandleMessage`) validates that a command name is non-empty and not server-initiated, but does **not** check that it is *known*. An unknown name flows into `ExecuteCommand` → `SocketCommandFactory.CreateCommand`, which throws `InvalidOperationException("No socket command generator found for: ...")` (`SocketCommandFactory.cs:38`). That is classified as `Faulted`, **logged at `LogError` with the exception**, and returned to the client as `"Internal Server Error"`. A client sending a stale/garbage command name (realistic across deploys) thus produces error-level log noise and a 500-shaped response for what is really a bad-request — inconsistent with the structured `"Malformed command."` / `"Command cannot be invoked by the client."` rejections sitting right beside it.

**Fix:** add an `IsKnownCommand(name)` guard (the factory already has the dictionary) and reject with a structured `"Unknown command."` instead of letting it reach the fault path.

## 2. Stale code comments referencing removed HTTP endpoints
`Game.Api/Sockets/Commands/GetEnemies.cs:7-8`, `GetAttributes.cs:6-7`, the other `Get*` reference commands, and `GetPlayerStatistics.cs:11-12` say things like "WebSocket equivalent of the `GET /api/Enemies` endpoint". Those controllers no longer exist (only `Login`, `Tags`, and the `Admin*` set remain) — the HTTP→socket migration is complete. Reword to "serves the reference-data set over the socket" (or drop the "equivalent of GET /api/X" phrasing).

## 3. Inconsistent `Retry-After` rounding between the two 429 paths
The IP rate limiter emits `((int)retryAfter.TotalSeconds)` (truncates — a 0.4s remaining window emits `Retry-After: 0`, telling a client it may retry immediately when it can't) at `RateLimiting/RateLimiterServiceCollectionExtensions.cs:63-64`, while the per-account login backoff uses `Math.Ceiling` at `Controllers/LoginController.cs:48-49`. Use `Math.Ceiling` in both (and consider a shared 429-envelope helper).

## 4. `SocketReplaced` self-sends then closes (latent footgun)
`Game.Api/Sockets/Commands/SocketReplaced.cs:9-14` calls `SendData(Success())` and `Close(...)` itself, then returns `new()`; `RunCommandUnderLock` (`SocketHandler.cs:194`) then unconditionally `SendData(response)` on that empty response. It's a no-op only because `Close()` already moved the socket out of `Open`. This breaks the otherwise-strict "the command runner owns the single send" invariant — a reorder (close after self-send) would emit a spurious empty frame. Prefer returning `Success()` and letting the runner send, then closing; or document why the self-send-then-close is required and the second send is intentionally suppressed.

scope: small

ginderjeremiah:
Closes #1144.

Addresses the four low-severity Game.Api polish items from the code-health review.

## 1. Unknown client command name → bad request, not internal fault

`SocketHandler.HandleMessage` validated that a command name was non-empty and not server-initiated, but not that it was *known*. An unknown name (a stale/garbage name, realistic across deploys) flowed into `SocketCommandFactory.CreateCommand`, which threw `InvalidOperationException` — classified as `Faulted`, **logged at `LogError` with the exception**, and returned to the client as `"Internal Server Error"`.

Adds `SocketCommandFactory.IsKnownCommand` (an O(1) lookup against the existing generator dictionary) and a guard in `HandleMessage` that rejects with a structured `"Unknown command."`, consistent with the `"Malformed command."` / `"Command cannot be invoked by the client."` rejections beside it.

## 2. Stale doc comments referencing removed HTTP endpoints

The `Get*` reference-data commands (and the two player-progress reads) said things like *"WebSocket equivalent of the `GET /api/Enemies` endpoint"*. Those controllers no longer exist — only `Login`, `Tags`, and the `Admin*` set remain, so the HTTP→socket migration is complete. Reworded to *"Serves the … reference-data set over the socket"* and dropped the `GET /api/X` references from the player-progress and base-class comments too.

## 3. Consistent `Retry-After` rounding

The IP rate limiter emitted `(int)retryAfter.TotalSeconds` (truncates — a 0.4 s remaining window emits `Retry-After: 0`, telling a client it may retry immediately when it can't), while the per-account login backoff used `Math.Ceiling`. Switched the IP limiter to `Math.Ceiling` so both 429 paths round up.

> A shared 429-envelope/`Retry-After` helper (noted as a "consider" in the issue) was left out: the two call sites differ enough in shape (rejection middleware vs. controller) that extracting one for a single formatting line wasn't clearly a win. Easy follow-up if a third 429 path appears.

## 4. `SocketReplaced` self-send-then-close (latent footgun)

`SocketReplaced.ExecuteAsync` self-sends `Success()` then closes, returning `new()` — which `RunCommandUnderLock` then unconditionally sends, a no-op only because `Close()` already moved the socket out of `Open`. The self-send-then-close is actually **required**: the client's `SocketReplaced` handler reacts to the success frame, which must arrive *before* the close frame, and the runner sends only after the command returns (too late). Documented that, and changed the return to `Success()` (not an empty frame) so that even a future reorder leaving the socket `Open` would emit a valid success frame rather than a spurious empty one.

## Test plan

- [x] `dotnet test` full backend suite — green (Core 802, Infrastructure 58, Application 528, Api 609).
- [x] New unit coverage: `HandleMessage` unknown-command rejection (structured error, no fault-path error log / "Internal Server Error"), and `IsKnownCommand` true/false for registered/unregistered names.

No frontend changes — the reworded comments are XML doc summaries that don't affect the TypeScript codegen.

https://claude.ai/code/session_01RRr3jdTfzajDgMYeUPspsr

---
_Generated by [Claude Code](https://claude.ai/code/session_01RRr3jdTfzajDgMYeUPspsr)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
@@ -51,6 +51,24 @@ namespace Game.Api.Tests.Unit
         }
 
         [Fact]
+        public async Task HandleMessage_UnknownCommandName_RejectsWithUnknownCommandErrorInsteadOfFaulting()
+        {
+            // A frame naming a command with no registered generator (a stale/garbage name) is a bad request,
+            // not an internal fault: it must be rejected with a structured "Unknown command." error rather than
+            // reaching the command lookup, whose throw would be logged at error and surfaced as an
+            // "Internal Server Error".
+            var (socket, handler) = CreateHandler(_ => null);
+
+            await handler.HandleMessage("{\"id\":\"c1\",\"name\":\"Unknown\",\"parameters\":null}");
+
+            Assert.Contains(socket.SentMessages, m => m.Contains("Unknown command.") && m.Contains("c1"));
+            Assert.Contains(_logs.Entries, e => e.Level == LogLevel.Warning && e.Message.Contains("unknown socket command"));
+            // It never reached the fault path: no error-level log, no "Internal Server Error" frame.
+            Assert.DoesNotContain(_logs.Entries, e => e.Level == LogLevel.Error);
+            Assert.DoesNotContain(socket.SentMessages, m => m.Contains("Internal Server Error"));
+        }
+
+        [Fact]
         public async Task ExecuteCommand_CommandFaults_SurfacesInternalServerErrorToTheClient()
         {
             var (socket, handler) = CreateHandler(name => name == "Boom" ? new InvalidOperationException("boom") : null);
@@ -138,6 +156,10 @@ namespace Game.Api.Tests.Unit
             {
                 return new StubCommand(commandInfo.Name, throwOn(commandInfo.Name)) { Id = commandInfo.Id };
             }
+
+            // Every name is treated as registered except the explicit "Unknown" sentinel, so the
+            // HandleMessage unknown-command guard can be exercised without populating the static registry.
+            public override bool IsKnownCommand(string commandName) => commandName != "Unknown";
         }
 
         private sealed class StubCommand(string name, Exception? toThrow) : AbstractSocketCommand
--- a/Game.Api.Tests/Unit/SocketCommandFactoryTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandFactoryTests.cs
@@ -48,5 +48,26 @@ namespace Game.Api.Tests.Unit
 
             Assert.False(factory.IsServerInitiatedOnly(commandName));
         }
+
+        [Theory]
+        [InlineData("GetZones")]
+        [InlineData("DefeatEnemy")]
+        [InlineData("SocketReplaced")]
+        public void IsKnownCommand_ReturnsTrue_ForRegisteredCommands(string commandName)
+        {
+            var factory = new SocketCommandFactory();
+
+            Assert.True(factory.IsKnownCommand(commandName));
+        }
+
+        [Theory]
+        [InlineData("NonExistentCommand")]
+        [InlineData("")]
+        public void IsKnownCommand_ReturnsFalse_ForUnregisteredCommands(string commandName)
+        {
+            var factory = new SocketCommandFactory();
+
+            Assert.False(factory.IsKnownCommand(commandName));
+        }
     }
 }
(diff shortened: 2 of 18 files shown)


ginderjeremiah (commented):
## Review: Game.Api polish — unknown-command rejection, stale comments, Retry-After rounding (#1144)

**Verdict:** Approve _(posted as a comment because I'm the PR author; GitHub won't accept a gating approval on your own PR)_ — all four items from #1144 are addressed correctly, well-tested, and CI is fully green.

I checked each change against the issue and the actual code:

1. **Unknown-command guard** (`SocketHandler.cs:369`) — correctly placed *after* the server-initiated check. `IsKnownCommand` returns true for server-initiated commands too (they're in the same generator dictionary), but ordering means those are already rejected first, so the guard only catches genuinely unregistered names. It cleanly stops a stale/garbage name from reaching `CreateCommand`'s throw → `Faulted` → error log → "Internal Server Error". The rejection shape matches the sibling `Malformed`/`server-initiated` rejections beside it.
2. **`IsKnownCommand`** (`SocketCommandFactory.cs:29`) — O(1) lookup against the existing dictionary, `virtual` so the test double controls known-ness without touching the static registry. Consistent with `IsServerInitiatedOnly` next to it.
3. **Stale comments** — all `GET /api/X` references removed; rewording is accurate (those controllers are gone, only `Login`/`Tags`/`Admin*` remain). XML-doc-only, no codegen impact.
4. **`Retry-After` rounding** (`RateLimiterServiceCollectionExtensions.cs`) — `Math.Ceiling` now matches the login backoff; fixes the truncate-to-0 footgun.
5. **`SocketReplaced`** — returns `Success()` instead of `new()`, with a clear explanation of why the self-send-then-close is required and the runner's second send is an intentional no-op. The defensive framing is correct.

### Test coverage
Solid. `HandleMessage` unknown-command path asserts the structured error, the warning log, and crucially the *absence* of the fault path (no error log, no "Internal Server Error"). `IsKnownCommand` covers registered/unregistered/empty names. The longer comments on these changes document genuine non-obvious invariants, so they're warranted rather than noise.

### What's good
Clean, surgically-scoped tech-debt PR that closes every item in the issue with proper test coverage. Deferring the shared 429-envelope helper was the right call — the issue listed it as a "consider", and the two call sites differ enough that extracting one for a single format line isn't a win until a third 429 path exists. No follow-up issue needed for it.

Merging.

---

PR #1147, opened 2026-06-22
Add skill rarity (field + admin authoring + display) (#1128)

ginderjeremiah:
## Summary

Implements **#1128** (spike #979): give skills an `ERarity` tier (Common→Mythic) **matching items and item mods**, authored in the admin Workbench and surfaced wherever skills display. The full vertical slice ships together — the field is useless without its display.

Rarity is **authoring + display only in V1** (the battle simulation never reads it), but it is a **reserved logic hook**: it lands on the lean `Game.Core.Skills.Skill` model so the future proficiency-XP accrual (#982 / #1123) can read the per-skill tier weight server-side at battle completion.

## Backend

- **Entity** `Skill` gains `int RarityId` + a `virtual Rarity` navigation (FK inferred by convention, like `Item`).
- **Migration** `AddSkillRarity` adds the column `NOT NULL, defaultValue: 1` (Common) — the migration only backfills existing rows; the model carries no default, so EF always writes the authored value on insert.
- **Contract** gains `ERarity RarityId`; **lean core model** gains `required ERarity Rarity` (the deliberate divergence from the `ESkillAcquisition` precedent — acquisition stayed off the core model because the battle never reads it; rarity *will* be read there).
- **`SkillMapper`** maps rarity to the contract **and** the core model; **`AdminSkills.SaveSkills`** persists it on add and edit.
- Tests: `SkillRarityVersioningTests` (re-tiering changes the skills content hash), extended `SkillMapperTests` (rarity round-trips to contract + core), and a new `AdminSkillsIntegrationTests` case (save/edit persists rarity). `TestDataSeeder` seeds a valid rarity (the new FK target).
- Docs: `docs/backend.md` (the placement + why) and `docs/game-design.md` (Skills section).

## Frontend

- **`ISkill` regenerated** from the contract via codegen (gains `rarityId`).
- **Workbench `skill.ts`**: `rarityId` in `newItem`, a Rarity `select` in Identity, and a rarity list badge — copying `item.ts`.
- **`challenge-unlocks.ts` / `challenges-view.svelte.ts`**: skill rewards now tint, sort, **and glow** by their real tier — dropped the neutral-accent fallback in the resolver and the `kind === 'skill' ? null` glow guard in the view-model (`glow` is now always a tier token, so `RewardIcon`'s dead null-fallback was removed too).
- **`SkillRewardTooltip.svelte`**: shell/sealed accent from the rarity hue (mirroring the item/mod reward tooltips).
[...]

Issue #1128, opened 2026-06-22
Add skill rarity (field + admin authoring + display)


ginderjeremiah:
Implementation of spike [#979](https://github.com/ginderjeremiah/GameServer/issues/979). Full design: `docs/spikes/979-skill-rarity.md`.

Give skills an `ERarity` tier (Common→Mythic) **matching items and item mods**, authored in the Workbench and surfaced wherever skills display. The full vertical slice ships in one issue (the field is useless without its display; no cross-entity validation, no battle assembly, no parity matrix).

## Decisions (from the spike)

- **Rarity is authoring + display in V1** — the battle simulation never reads it; no gameplay effect ships here.
- **It is a reserved logic hook**, not purely cosmetic: rarity is the **tier weight** the proficiency XP accrual consumes (#982 decision 4 / #1116, applied by #1123 — flat `1` until then). So the field is positioned where that calc reads it.
- **Placement: `Skill` entity + contract + the lean `Game.Core.Skills.Skill`** (mapped in both `ToContract` and `ToCore`). This is a **deliberate divergence from the `ESkillAcquisition` precedent** (#1100), which stayed off the core model because the battle never reads provenance. Rarity *will* be read at battle completion — proficiency XP is backend-authoritative there (#982 decision 9) and looks a skill up by id via `ISkills.GetSkill(id)`, which returns the lean core model, so `GetSkill(id).Rarity` is the natural tier-weight lookup. It also brings skills into parity with `Game.Core.Items.Item.Rarity`.
- **Zero battle-parity surface:** rarity is not read in the seeded 40 ms tick simulation, so the **frontend battle/skill model gains no rarity field** and no parity vectors change. The client gets rarity for display from the `ISkill` reference contract.
- **Existing skills backfill to `Common`** in the migration; the model carries no default, so EF always writes the authored value on insert.

## Backend

- [ ] `Game.Infrastructure/Entities/Skill.cs`: add `int RarityId` + `virtual Rarity Rarity` navigation (FK inferred by convention, like `Item` — no `GameContext` relationship config needed; the `Rarities` seed supplies the FK targets).
- [ ] Migration `AddSkillRarity`: add `RarityId` column, `NOT NULL`, `defaultValue: 1` (Common). Pattern: `20260621221535_AddSkillAcquisition`.
- [ ] `Game.Abstractions/Contracts/Skill.cs`: add `ERarity RarityId`.
[...]

PR #1194, opened 2026-06-24
Validate authored intrinsic-enum FK references on admin reference saves (#1148)

### LINKED ISSUE #1148: Validate skill RarityId on admin save (BadRequest instead of FK 500)
Follow-up from #1128 / PR #1147 review.

## Problem

`AdminSkills.SaveSkills` persists `RarityId` straight through to the entity. If a client sends a missing/`0` `RarityId` (or any value with no matching `Rarities` row), the only thing catching it is the `FK_Skills_Rarities_RarityId` foreign key, which surfaces as an opaque **500 Internal Server Error** rather than a clean validation rejection.

`ERarity` has no `0` member (`Common = 1`), so `(ERarity)0` is not a valid tier. This is the exact gap that produced the failing test fixed in PR #1147 (inline payloads omitting `RarityId` 500'd against the FK).

Not a live exploit — real clients always send a rarity (`newItem` defaults to `Common`; the Workbench binds the rarity select) — but it's an unguarded edge.

## Proposal

Validate the rarity in `AdminSkills.SaveSkills` (and return `AdminSaveResult.Failure` → `BadRequest`) the same way the existence check is done for edits — reject a `RarityId` that isn't a defined `ERarity` before the insert/update is staged, so a bad payload gets a clean message instead of a 500.

Consider whether a generic "validate authored FK references" helper is warranted (acquisition is a free `[Flags]` int and needs no such guard, but other reference FKs might).

## Notes
- Surfaced during the #1128 skill-rarity review.
- Small, self-contained, no user-facing behavior change for valid input.

ginderjeremiah:
## Summary

Closes #1148 — a follow-up from the #1128 / PR #1147 skill-rarity review.

`AdminSkills.SaveSkills` persisted `RarityId` straight through to the entity, so a missing/`0` `RarityId` (or any value with no matching `Rarities` row) was caught only by the `FK_Skills_Rarities_RarityId` foreign key and surfaced as an opaque **500** instead of a clean validation rejection. `ERarity` has no `0` member (`Common = 1`), so `(ERarity)0` is not a valid tier — the exact gap behind the failing test fixed in PR #1147.

## What changed

- **New shared guard `ReferenceFieldValidation.FindUndefinedEnum`** (`Game.DataAccess/Repositories/Admin/`): rejects any add/edit change whose authored single-value enum FK isn't a defined member, returning `AdminSaveResult.Failure` up front before anything is staged. The controllers already map `Failure` → `BadRequest`, so a bad payload now gets a clean 400 + message instead of a 500. Deletes are skipped (they target an existing row by id).
- **Wired into all three reference-data save paths** for every intrinsic single-value enum FK they persist:
  - `AdminSkills.SaveSkills` → `RarityId`
  - `AdminItems.SaveItems` → `RarityId`, `ItemCategoryId`
  - `AdminItemMods.SaveItemMods` → `RarityId`, `ItemModTypeId`

## On the "generic helper" question (and scope beyond skills)

The issue asked to consider whether a generic *"validate authored FK references"* helper is warranted. It is: `RarityId` is straight-through on **skills, items, and item mods**, and `ItemCategoryId` / `ItemModTypeId` are the identical FK-backed-enum gap (all three enums start at `1`, so a `0`/missing payload 500s the FK). Rather than fix skills alone and leave three near-identical latent 500s, I extracted one small generic guard and applied it to every such FK on the paths I was already touching. **Flagging this as the deliberate scope decision** — the issue named skills, and I extended to the parallel cases; happy to trim back to skill-rarity only if you'd prefer.

`[Flags]` fields (skill `Acquisition`) are deliberately **not** covered — a free combination of members is valid, and `Enum.IsDefined` would wrongly reject composite values. This matches the issue's note that acquisition needs no such guard.

## Test plan

- New pure unit tests `ReferenceFieldValidationTests` (6) — valid passes, undefined add/edit rejected, delete skipped, first-violation order, empty set.
- New integration wiring tests confirming each DI-resolved save path rejects before the FK hit: skill rarity, item rarity, item category, item-mod rarity, item-mod type.
- Full `Game.Application.Tests` suite green (539 passed).

## Notes / possible follow-up

This covers the intrinsic single-value enum FKs on the reference *save* paths. Other authored FK references (e.g. relationship setters that already validate membership their own way) are unaffected. If a broader "validate every authored FK" sweep is wanted it can be a separate tech-debt task — this PR scopes to the reported rarity gap and its immediate siblings.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CMGbKifLjtXdoVqhkaeh8v)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminItemModsIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminItemModsIntegrationTests.cs
@@ -42,6 +42,36 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public void SaveItemMods_UndefinedRarity_ReturnsFailureWithoutPersisting()
+        {
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminItemMods>();
+
+            var result = admin.SaveItemMods(
+            [
+                new Change<Contracts.ItemMod> { ChangeType = EChangeType.Add, Item = NewItemMod(name: "Bad Rarity", rarity: (ERarity)0) },
+            ]);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal("0 is not a valid item mod rarity.", result.ErrorMessage);
+        }
+
+        [Fact]
+        public void SaveItemMods_UndefinedType_ReturnsFailureWithoutPersisting()
+        {
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminItemMods>();
+
+            var result = admin.SaveItemMods(
+            [
+                new Change<Contracts.ItemMod> { ChangeType = EChangeType.Add, Item = NewItemMod(name: "Bad Type", type: (EItemModType)0) },
+            ]);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal("0 is not a valid item mod type.", result.ErrorMessage);
+        }
+
+        [Fact]
         public async Task SaveItemMods_AddAndEdit_PersistAndUpdateInPlace()
         {
             int modId;
--- a/Game.Application.Tests/DataAccess/AdminItemsIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminItemsIntegrationTests.cs
@@ -89,6 +89,36 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public void SaveItems_UndefinedRarity_ReturnsFailureWithoutPersisting()
+        {
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminItems>();
+
+            var item = NewItem(name: "Bad Rarity Item");
+            item.RarityId = (ERarity)0;
+
+            var result = admin.SaveItems([new Change<Contracts.Item> { ChangeType = EChangeType.Add, Item = item }]);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal("0 is not a valid item rarity.", result.ErrorMessage);
+        }
+
+        [Fact]
+        public void SaveItems_UndefinedCategory_ReturnsFailureWithoutPersisting()
+        {
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminItems>();
+
+            var item = NewItem(name: "Bad Category Item");
+            item.ItemCategoryId = (EItemCategory)0;
+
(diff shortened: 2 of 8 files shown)


ginderjeremiah:
## Review: Validate authored intrinsic-enum FK references on admin reference saves (#1148)

**Verdict: Approve** — clean, correctly-scoped fix for the reported FK-500 gap, with good test coverage and no blocking findings.

### What it does well
- The root cause is correctly diagnosed: `RarityId`/`ItemCategoryId`/`ItemModTypeId` are strongly-typed enums on the contracts (`Skill`/`Item`/`ItemMod`) persisted straight through as FKs into enum-seeded reference tables, and all three enums start at `1` (verified in `Game.Core/Enums.cs`), so a missing/`0` payload had no backing row and 500'd on the FK. `ReferenceFieldValidation.FindUndefinedEnum` rejects it up front as an `AdminSaveResult.Failure` → 400.
- Generic helper is the right call. The issue explicitly asked whether a shared "validate authored FK references" guard was warranted; extending past skill-rarity to the three parallel FKs avoids leaving near-identical latent 500s, and the scope decision is flagged transparently in the PR body. The `[Flags]` exclusion (acquisition) is correct — `Enum.IsDefined` would wrongly reject composite values — and is documented in the XML remarks.
- Guard runs before anything is staged, leaving the batch atomic, consistent with the existing `ChangeSetProcessor` up-front validation pattern. Deletes are correctly skipped (they target an existing row by id).
- Type constraints are sound: `where T : IModel, TEnum : struct, Enum` matches `Change<T>` and resolves to the non-boxing generic `Enum.IsDefined<TEnum>`; `Convert.ToInt32(value)` is valid for the int-backed enums and yields a clear message.
- Layering is respected — the guard lives in `Game.DataAccess/Repositories/Admin` alongside the admin repos and speaks contracts/`AdminSaveResult`, no entity leakage, no domain logic in the API layer.

### Test coverage
Solid and follows the project's rules. `ReferenceFieldValidationTests` (6 pure unit tests) cover valid pass, undefined add/edit rejection, delete-skipped, first-violation ordering, and empty set. Integration tests exercise each DI-resolved save path (skill rarity, item rarity, item category, item-mod rarity, item-mod type) confirming rejection before the FK hit. No frontend mirror needed — this is server-side admin authoring, not battle-parity logic.

### Notes (non-blocking)
- The `[Flags]`-exclusion is enforced by convention (the caller must not pass a flags enum), not by the type system. The clear XML remark is the right mitigation; no change needed.

CI is fully green (`all-checks-passed`; frontend jobs skipped as expected for a backend-only change) and the branch merges cleanly against current `main`. Merging via squash.

---

PR #1197, opened 2026-06-24
Guard switch-away credit against a live departed-character socket (#1170)

### LINKED ISSUE #1170: SwitchPlayer credits the departed character without verifying its socket is closed (off-lock RMW race)
Found during a code-health review. The lossless character-switch credit relies entirely on client cooperation, with no server-side guard against the lost-update race the per-socket command lock otherwise prevents.

## Problem

`SwitchPlayer` → `SimulateSwitchProgress` (`Game.Api/Controllers/LoginController.cs:106-164`, `Game.Application/Services/BattleService.cs:374-377`) credits the departed character's elapsed idle time by a read-modify-write against its `PlayerState` and Player aggregate, **over HTTP**. Correctness depends on the documented assumption that the client tears down its game socket before calling this.

There is no server check that the departed player's socket is actually gone before crediting. A misbehaving/malicious client that keeps its socket open while calling `SwitchPlayer` would run the offline-credit RMW concurrently with the live battle loop's saves — the exact lost-update race the per-socket command lock exists to prevent, except this path runs outside that lock.

## Fix

Before crediting, assert the departed player has no active socket (reuse `SocketManagerService.HasActiveSocket`) and reject/skip the credit if one is live. At minimum, document it as a known anti-cheat gap. The target's ownership is already validated up front, so this just closes the departed side.

scope: small

ginderjeremiah:
## Summary

Closes #1170.

`SwitchPlayer` → `CreditDepartedCharacter` (`Game.Api/Controllers/LoginController.cs`) credits the departed character's elapsed idle time with a **read-modify-write against its aggregate, over HTTP, off that character's battle loop**. Correctness depended entirely on the documented assumption that the client tears its game socket down before switching — there was no server-side check. A misbehaving/malicious client that kept its socket open while calling `SwitchPlayer` would run the offline-credit RMW concurrently with the live battle loop's saves: the exact lost-update race the per-socket command lock exists to prevent, except this path runs outside that lock.

## Fix

Before crediting, assert the departed character has no live socket via the existing `SocketManagerService.HasActiveSocket`. If one is live, **skip the credit (and log a warning)** — its battle loop is already serializing its own saves under the per-socket command lock, so the live loop owns them. The switch itself still proceeds (binding the new character is independent and legitimate); only the off-lock credit is skipped. The target's ownership is already validated up front, so this just closes the departed side as the issue described.

- `LoginController` gains an `ILogger<LoginController>` so the skip is observable (it signals a client that didn't tear its socket down).
- XML docs on `SwitchPlayer`/`CreditDepartedCharacter` updated to note the server-side guard.

## Why skip rather than reject the whole switch

The departed character's progress is being saved by its own live loop, so skipping the HTTP credit loses nothing — it avoids the double-write race. Rejecting the entire switch would be more disruptive for no correctness benefit. This matches the issue's "reject/skip the credit if one is live".

## Tests

- New integration test `SwitchPlayer_DepartedCharacterStillHasLiveSocket_SkipsCreditButStillSwitches`: registers a live socket for the departed character, then switches. Asserts the switch still binds the target and rotates the token, but the departed character is **not** credited (its level does not advance). Verified it **fails without the guard** (the departed character levels up) and **passes with it** — a real regression guard, not a tautology.
- New `AssertPlayerNotCreditedAsync` helper (the inverse of the existing `AssertPlayerLeveledAboveAsync`).

## Test plan

- `Game.Api.Tests` — 610 passed (full suite, incl. the new test).
- Negative check: temporarily disabling the guard makes the new test fail; restored and re-confirmed green.

## Docs

Updated the lossless-switching paragraph in `docs/backend.md` to note the socket-teardown expectation is now enforced server-side (credit skipped + logged when the departed socket is still live).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gg29Mk5kZLXYifKWdBq5Nk)_


ginderjeremiah:
## Review: Guard switch-away credit against a live departed-character socket (#1170)

**Verdict:** Approve — closes the off-lock RMW race exactly as the issue prescribed, with a real regression test and docs updated. CI is green.

### Correctness
- `CreditDepartedCharacter` now checks `SocketManagerService.HasActiveSocket(departedPlayerId)` before the credit and skips (logging a warning) when the departed character still holds a live socket. This is the precise fix #1170 asked for: the live battle loop already serializes its saves under the per-socket command lock, so skipping the off-lock HTTP credit avoids the lost-update race without losing progress. The check is correctly keyed on the *departed* player id, and the guard sits before `EnsureSessionLoaded`/`LoadPlayer`, so it also avoids needless work on the skip path.
- "Skip the credit, still bind the target" is the right call — the switch is independent of the departed-side settle, and ownership is validated up front, so nothing is mutated on a rejected switch.

### Layering / conventions
- The change lives entirely in the API orchestration layer (`LoginController`), which is the correct home for this HTTP-flow concurrency concern — no domain logic leaks into the controller. `ILogger` injection follows the existing constructor-injection pattern. All `if` blocks are braced per CLAUDE.md.

### Test coverage
- New integration test `SwitchPlayer_DepartedCharacterStillHasLiveSocket_SkipsCreditButStillSwitches` is the appropriate test type (exercises real Redis socket presence) and is a genuine regression guard — confirmed to fail without the guard and pass with it. It asserts both halves: the switch still binds the target + rotates the token, and the departed character is not credited.
- `AssertPlayerNotCreditedAsync` polls a 2s window to let any write-behind credit surface before asserting absence — sound, mirroring the positive credit test's window. Adds ~2s runtime by design, which is the inherent cost of asserting a negative; acceptable.

### Docs
- `docs/backend.md` lossless-switching paragraph updated to note the server-side enforcement, per CLAUDE.md's documentation requirement.

### Nits
- None blocking. Comments are on the verbose side but justified given how subtle the race is.

No blocking or should-fix findings. (Note: I am the PR author, so this posts as a comment rather than a formal GitHub approval, but the verdict is approve.)

---

Issue #1191, opened 2026-06-24
Frontend small correctness follow-ups (deep-link payload, skills-view commit baseline)


ginderjeremiah:
Found during a code-health review — two small, currently-untracked frontend correctness nuances. Individually minor; grouped here so they're tracked.

## 1. Cross-screen deep-link payload is dropped when the target screen is already active

`navigation.requestScreen(key, payload)` (`UI/src/stores/navigation.svelte.ts:25-29`) hands its one-shot payload to the destination via `consumePayload()` **on mount**. The game shell's `handleNavigate` (`UI/src/routes/game/+page.svelte:93-117`) only sets `currentScreen = key` — so if the requested screen is *already* the active one, the component does not remount, `onMount`/`consumePayload` never runs, and the payload is silently lost.

The one documented caller (Statistics → Codex dossier) is always cross-screen, so it works today. But any within-target deep-link, or a future caller re-targeting the open screen, will no-op. Either consume the payload reactively in the shell and push it to the active screen, or document the same-screen limitation explicitly at `requestScreen`.

## 2. Skills-view `committed` baseline is advanced by superseded (non-latest) commits

`SkillsView.commit()` (`UI/src/routes/game/screens/skills/skills-view.svelte.ts:456-479`) serializes persists and only rolls back the **latest** edit on failure (an `isLatest` guard), but on success it sets `this.committed = next` for **every** commit, including superseded ones. Because commits are serialized in order, the final value resolves correctly today — but mid-sequence the rollback target (`committed`) momentarily reflects an intermediate loadout rather than the last *confirmed-and-still-current* one, so a later failure would roll back to that intermediate state instead of the true last-good loadout.

Tighten `committed` to advance only for the latest in-flight edit (mirroring the rollback's `isLatest` guard) so the baseline matches the intended "last loadout the server confirmed" semantics. Distinct from the shared-`serialize` refactor (#1183) — that's about extracting the queue mechanics; this is a baseline-advancement correctness nuance — though both touch this method, so they could be done together.

tech debt, claude, scope: small

ginderjeremiah:
Picked this up in #1204.

**Part 1** is fixed: a deep-link to the already-active screen now forces a remount (keyed on a `screenNonce`, gated by `navigation.hasPendingPayload`) so the target still consumes its one-shot payload. Cross-screen requests are unchanged.

[...]

PR #1200, opened 2026-06-24
Extract shared SerializedQueue for optimistic-mutation queues (#1183)

### LINKED ISSUE #1183: Extract a shared `serialize` helper for optimistic-mutation queues (inventory + skills loadout)
Follow-up from PR #1179 (review of #1171).

## Problem

There are now two hand-rolled copies of the same "serialized optimistic mutation" pattern — chaining each mutation off a `lastOperation`/`lastCommit` promise so rollback baselines never interleave, with a `.catch(() => undefined)` tail keeping the shared promise always-fulfilled:

- `InventoryManager.serialize` (`UI/src/lib/engine/player/inventory-manager.ts:325`) — generic `serialize<T>(operation): Promise<T>`, used by equip/unequip/applyMod/removeMod/setFavorite.
- `SkillsView.commit` (`UI/src/routes/game/screens/skills/skills-view.svelte.ts`) — an inline chain with the same `result` + `this.lastCommit = result.catch(() => undefined)` tail.

The chaining/guard mechanics are identical; only the per-operation body and rollback model differ (inventory uses field snapshots and returns a promise; the skills view uses a `commitSeq` + last-confirmed `committed` baseline and returns void).

## Proposal

Extract the queue mechanics into a small shared helper (e.g. a `SerializedQueue`/`serialize` utility in `$lib/common` or `$lib/engine`) that owns the `lastOperation` promise and the always-fulfilled tail, and route both call sites through it. Each caller keeps its own optimistic-apply + rollback policy inside the operation closure.

## Notes
- Surfaced in the #1179 review: <https://github.com/ginderjeremiah/GameServer/pull/1179#discussion_r3462864439>.
- No user-facing behavior change — pure DRY/maintainability cleanup.
- Small and self-contained; the two consumers are the only current users of the pattern.

ginderjeremiah:
## Summary

Both `InventoryManager` and `SkillsView` had grown a hand-rolled copy of the same "serialized optimistic mutation" pattern — chaining each mutation off a shared `lastOperation`/`lastCommit` promise so rollback baselines never interleave, with a `.catch(() => undefined)` tail keeping the shared promise always-fulfilled. The chaining/guard mechanics were identical; only the per-operation body and rollback model differed.

This extracts the queue mechanics into a small **`SerializedQueue`** helper in `$lib/common` and routes both call sites through it. Each caller keeps its own optimistic-apply + rollback policy inside the operation closure — only the ordering and the always-fulfilled tail move into the helper.

## How it addresses the issue

- New `UI/src/lib/common/serialized-queue.ts` — `SerializedQueue` owns the `tail` promise, the `tail.then(op, op)` chaining, and the `.catch(() => undefined)` tail, exposing a single `run(operation)`.
- `InventoryManager.serialize` now delegates to a `SerializedQueue` instance (its public surface and per-mutation rollback closures are unchanged).
- `SkillsView.commit` now enqueues through the same helper; the `commitSeq`/`isLatest` baseline-advancement logic stays in the closure.
- No user-facing behavior change — pure DRY/maintainability cleanup, exactly the two consumers the issue named.

The `tail.then(op, op)` form (run the next operation regardless of whether the previous fulfilled or rejected) preserves the inventory manager's original `.then(operation, operation)` semantics; the skills view used a single-arg `.then`, but since the tail is always-fulfilled the two were already equivalent in practice.

## Test plan

- [x] New `serialized-queue.test.ts` covers: each op's result returns to its own caller; an op doesn't start until the previous settles; a rejecting op rejects to its caller but later ops still run; enqueue order is preserved.
- [x] Existing `skills-view` and `inventory-manager`/player tests pass unchanged (178 tests across the affected suites).

## Note

The `#1191` part-2 baseline-advancement correctness nuance also lives in `SkillsView.commit` and is left untouched here, as that issue intends — this PR only relocates the queue mechanics.

Closes #1183

---
_Generated by [Claude Code](https://claude.ai/code/session_01CVAcVb3j7bigQ94Ng2tRLH)_

### DIFF
--- a/UI/src/lib/common/index.ts
+++ b/UI/src/lib/common/index.ts
@@ -12,4 +12,5 @@ export * from './item-display';
 export * from './types';
 export * from './hooks';
 export * from './event-measuring';
+export * from './serialized-queue';
 export * from './statify.svelte';
--- a/UI/src/lib/common/serialized-queue.ts
+++ b/UI/src/lib/common/serialized-queue.ts
@@ -1,0 +1,27 @@
+/**
+ * Serializes async operations onto a single chain so each one runs only after the previous has fully
+ * settled — its persist and any rollback included. Overlapping optimistic mutations would otherwise
+ * interleave their rollback baselines: one operation's rollback could restore a snapshot taken before
+ * another's optimistic change, silently discarding it and leaving local state diverged from the server.
+ * Chaining queues the operations instead, which is better UX than dropping the later one — the same
+ * "collapse concurrency" spirit as auth.ts's single-flight refresh.
+ *
+ * Each caller keeps its own optimistic-apply + rollback policy inside the operation closure; the queue
+ * only owns the ordering and the always-fulfilled tail.
+ */
+export class SerializedQueue {
+	/** Tail of the queue; each enqueued operation chains off it. Kept always-fulfilled (see {@link run}). */
+	private tail: Promise<unknown> = Promise.resolve();
+
+	/**
+	 * Queues `operation` to run after every previously-enqueued operation has settled, returning a promise
+	 * for its result (rejecting if `operation` itself rejects). `operation` runs regardless of whether the
+	 * previous one fulfilled or rejected. The internal tail swallows that rejection so one operation's
+	 * failure can't break the chain or surface as an unhandled rejection on the shared promise.
+	 */
+	run<T>(operation: () => Promise<T>): Promise<T> {
+		const result = this.tail.then(operation, operation);
+		this.tail = result.catch(() => undefined);
+		return result;
+	}
+}
--- a/UI/src/lib/engine/player/inventory-manager.ts
+++ b/UI/src/lib/engine/player/inventory-manager.ts
@@ -2,6 +2,7 @@ import { IInventoryItem, IBattlerAttribute, ELogType, EItemCategory, EEquipmentS
 import { playerManager } from '$lib/engine';
 import { BattleAttributes, Item, newItem, newItemMod } from '$lib/battle';
 import { logMessage } from '$lib/engine/log';
+import { SerializedQueue } from '$lib/common';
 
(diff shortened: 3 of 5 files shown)


ginderjeremiah (commented):
## Review: Extract shared `SerializedQueue` for optimistic-mutation queues (#1183)

**Verdict:** Approve (submitted as a comment since I'm the PR author) — faithfully extracts the duplicated queue mechanics, no behavior change, well tested.

This does exactly what #1183 asked: the identical `lastOperation`/`lastCommit` chaining + always-fulfilled `.catch(() => undefined)` tail is lifted into a single `SerializedQueue.run`, and both `InventoryManager.serialize` and `SkillsView.commit` route through it while keeping their own optimistic-apply/rollback policy in the closure. Placement in `$lib/common` and the mirrored test path both match the project conventions.

The one real semantic question — `tail.then(op, op)` (two-arg) replacing `SkillsView`'s original single-arg `.then` — is correctly reasoned in the PR description: because the tail is always kept fulfilled, the rejection handler can never fire, so the two forms are equivalent. The returned promise is discarded at the skills call site but is still handled internally via the tail's `.catch`, so no unhandled rejection is introduced.

### Test coverage
- New `serialized-queue.test.ts` covers the four behaviors that matter: per-caller result routing, no-start-until-previous-settles ordering, a rejecting op rejecting to its own caller while later ops still run, and enqueue order preservation. Adequate for a pure-mechanics helper.
- Existing inventory/skills suites exercise the consumers unchanged; frontend + e2e checks are green (backend jobs correctly skipped — no backend changes).

### Nits
- None blocking. The class/`run` doc comments are on the verbose side and partially overlap the retained `InventoryManager.serialize` comment, but the subject (interleaving rollback baselines) is genuinely subtle and the extra context is justified, so I'd leave it.

Clean, well-scoped cleanup. No follow-ups needed.

---

PR #1201, opened 2026-06-24
Lazily build BattleAttributes display projections (#1190)

### LINKED ISSUE #1190: BattleAttributes.recompute() rebuilds display-only projections on the per-effect battle hot path
Found during a code-health review. The live battlers' `BattleAttributes` rebuilds two **display-only** projections on every mid-battle modifier change, even though nothing in the combat loop reads them.

## Problem

`recompute()` (`UI/src/lib/battle/battle-attributes.ts:106-120`) runs on **every** `setData`/`addModifier`/`removeModifier`. Beyond the `attributeValues` array the combat actually consumes, it always rebuilds:

- `attributeMap` — a `.map` over all ~25 attributes, each calling `attributeName(id, staticData.attributes)` (a reference-data `.find` scan per attribute), and
- `nonZeroAttributeMap` — a `.filter` over that.

During a live battle these are pure waste: every effect application does `removeModifier` + `addModifier` (a swap → two full recomputes) per fire, and `advanceEffects` calls `removeModifier` per expiry — for **both** battlers, potentially every 40 ms tick. None of `attributeMap`/`nonZeroAttributeMap` is consumed during combat; they back the inventory/breakdown/tooltip surfaces only. So each effect tick rebuilds ~25-entry arrays plus ~25 `.find` scans for no consumer.

The #811 memoization work covered equipment/attribute *derivation* and the per-spawn re-derive skip, but not this per-modifier display reprojection.

## Fix

Compute the display projections **lazily on read** (the `getAttributeMap`/`getAttributeCount` accessors already exist as the read seam — back them with a memo invalidated on `recompute` rather than eagerly rebuilt), or otherwise skip the projection rebuild when only `attributeValues` is needed. The combat hot path then only touches `attributeValues`; the projections rebuild once, the next time a UI surface reads them.

## Notes

- No behavior change — pure perf. Most impactful with effect-bearing loadouts (the common case for an idle RPG built around skill effects); zero cost when no effects are active.
- The live `BattleAttributes` is `statify`-reactive, so each eager rebuild also reassigns reactive state unnecessarily.

ginderjeremiah:
## Summary

`BattleAttributes.recompute()` ran on every `setData`/`addModifier`/`removeModifier` — i.e. on every mid-battle effect application/expiry, for **both** battlers, up to each tick — and eagerly rebuilt two **display-only** projections (`attributeMap` and `nonZeroAttributeMap`) that the combat loop never reads. Each rebuild did a ~25-entry `.map`, a per-attribute `attributeName` reference-data `.find` scan, and a `.filter`, purely to back the inventory/breakdown/tooltip surfaces.

This builds those projections **lazily on first read** instead, memoised and invalidated on the next recompute, so the per-modifier hot path only touches `attributeValues` (the array combat actually consumes). The projections rebuild once, the next time a UI surface reads them.

Closes #1190.

## How it addresses the issue

The issue asked to compute the display projections lazily on read (the `getAttributeMap`/`getAttributeCount` accessors are already the read seam), backed by a memo invalidated on `recompute` rather than eagerly rebuilt.

- `recompute()` now reassigns only `attributeValues` and nulls the two projection caches.
- A new `#ensureProjections()` builds + memoises the projections on first read.
- The caches live in **non-reactive `#` fields**. That matters for the `statify` reactivity contract documented on the class: the lazy build's write is *not* a `$state` write, so it's legal mid-derivation (the same reason the modifier list is `#`), while `#ensureProjections()` reads the reactive `attributeValues` so `$derived` tooltip/inventory consumers keep tracking changes.

No behavior change — pure perf. Most impactful with effect-bearing loadouts (the common case for an idle RPG built around skill effects); zero cost when no effects are active.

## Test plan

- [x] Existing memoization + rebuild-on-change tests still pass (`getAttributeMap` returns a stable reference, rebuilds after a modifier change).
- [x] Added a test asserting the projection is **not** rebuilt on the modifier hot path (spying on the `staticData.attributes` reference-set access) until the next read, and that the rebuilt map reflects the accumulated changes.
- [x] Full battle unit suite (203 tests) + inventory/challenges consumer suites green.
- [x] `npm run lint` (eslint + svelte-check) clean — 0 errors, 0 warnings.

> Note: this is a frontend-only display-projection change with no effect on battle outcomes, so no backend parity mirror is required.

https://claude.ai/code/session_01W9kwaudyVThMjoFRaSZMDi

---
_Generated by [Claude Code](https://claude.ai/code/session_01W9kwaudyVThMjoFRaSZMDi)_

### DIFF
--- a/UI/src/lib/battle/battle-attributes.ts
+++ b/UI/src/lib/battle/battle-attributes.ts
@@ -125,6 +125,8 @@ export class BattleAttributes {
 	 *  `attributeValues` keeps `$derived` consumers tracking changes; the projections themselves cache
 	 *  in non-reactive `#` fields, so this lazy write is legal mid-derivation. */
 	#ensureProjections(): { all: AttributeEntry[]; nonZero: AttributeEntry[] } {
+		// Must read attributeValues unconditionally (not inside the `if`): on a cache hit this read is
+		// the only thing registering the $derived dependency, so consumers re-derive on recompute.
 		const values = this.attributeValues;
 		let all = this.#attributeMap;
 		let nonZero = this.#nonZeroAttributeMap;
--- a/UI/src/lib/battle/battle-attributes.ts
+++ b/UI/src/lib/battle/battle-attributes.ts
@@ -25,21 +25,25 @@ export interface AttributeEntry {
  * path the breakdown screen uses — only when the modifier set changes, and memoised between changes.
  *
  * Two reactivity rules make this safe once an instance is made reactive (`statify`):
- * - The recompute is **eager** (on each `setData`/`addModifier`/`removeModifier`), so
- *   {@link getValue}/{@link getAttributeMap} are pure reads. Those reads happen inside Svelte
- *   `$derived` (e.g. a battler card's MaxHealth bar); a lazy recompute would be an illegal
- *   mid-derivation `$state` write (`state_unsafe_mutation`).
- * - The modifier list and the `calcDerived` flag are **private (`#`) fields**, invisible to
- *   `statify`, so they stay non-reactive. That keeps `removeModifier`'s reference identity intact
- *   (a reactive array would deep-proxy its elements, so the stored modifier would no longer be
- *   `===` the reference the caller holds). Only the reactive fields — what the UI reads — are
- *   reassigned on each recompute.
+ * - The per-attribute totals recompute is **eager** (on each `setData`/`addModifier`/`removeModifier`),
+ *   so {@link getValue} is a pure read. That read happens inside Svelte `$derived` (e.g. a battler
+ *   card's MaxHealth bar); a lazy total recompute would be an illegal mid-derivation `$state` write
+ *   (`state_unsafe_mutation`).
+ * - The modifier list, the `calcDerived` flag, and the memoised display projections are **private
+ *   (`#`) fields**, invisible to `statify`, so they stay non-reactive. For the modifier list that
+ *   keeps `removeModifier`'s reference identity intact (a reactive array would deep-proxy its
+ *   elements, so the stored modifier would no longer be `===` the reference the caller holds). Only
+ *   the reactive `attributeValues` — what combat reads — is reassigned on each recompute.
  *
- * The named display projections ({@link getAttributeMap}/{@link getAttributeCount}) are memoised
- * alongside the totals (they depend only on the values), so every `$derived` tooltip/inventory
- * consumer reads a shared cached array instead of rebuilding its own per call site. Names resolve
- * through the documented {@link attributeName} convention against the live `Attributes` reference
- * set, the single source other display surfaces use.
+ * The named display projections ({@link getAttributeMap}/{@link getAttributeCount}) are nothing the
+ * combat loop consumes — they back the inventory/breakdown/tooltip surfaces only — so they are built
+ * **lazily on first read** and memoised, then invalidated on the next recompute. This keeps the
+ * per-modifier hot path (every effect application/expiry, for both battlers, up to each tick) off the
+ * per-attribute `.map` + `attributeName` reference scans; the projections rebuild once, the next time
+ * a UI surface reads them. Caching them in non-reactive `#` fields makes the lazy build's write legal
+ * mid-derivation (it is not a `$state` write), while a read of the reactive `attributeValues` keeps
+ * the `$derived` consumers tracking changes. Names resolve through the documented {@link attributeName}
+ * convention against the live `Attributes` reference set, the single source other display surfaces use.
  */
 export class BattleAttributes {
 	/** Every modifier composing this set, in the order the backend applies them. */
@@ -48,10 +52,10 @@ export class BattleAttributes {
 	#calcDerived = true;
 	/** The memoised per-attribute totals, recomputed only when the modifier set changes. */
 	private attributeValues: number[] = new Array<number>(attributesMaxId + 1).fill(0);
-	/** Memoised display projection of every attribute (including zeroes), recomputed alongside the totals. */
-	private attributeMap: AttributeEntry[] = [];
-	/** Memoised display projection of only the non-zero attributes. */
-	private nonZeroAttributeMap: AttributeEntry[] = [];
+	/** Lazily-built display projection of every attribute (including zeroes); null when invalidated. */
+	#attributeMap: AttributeEntry[] | null = null;
+	/** Lazily-built display projection of only the non-zero attributes; null when invalidated. */
+	#nonZeroAttributeMap: AttributeEntry[] | null = null;
 
 	constructor(attList: IBattlerAttribute[] = [], calcDerivedStats: boolean = true) {
 		this.setData(attList, calcDerivedStats);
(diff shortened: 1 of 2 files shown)


UI/src/lib/battle/battle-attributes.ts
```diff
@@ -107,16 +111,35 @@ export class BattleAttributes {
...
+	/** The count of non-zero attributes, for consumers that only need the size. */
+	public getAttributeCount = (): number => this.#ensureProjections().nonZero.length;
 
-	/** Recomputes the per-attribute totals and their named projections from the current modifiers.
-	 *  Called only when the modifier set changes; reads then return the memoised results. */
+	/** Builds and memoises the named display projections on demand. A read of the reactive
+	 *  `attributeValues` keeps `$derived` consumers tracking changes; the projections themselves cache
+	 *  in non-reactive `#` fields, so this lazy write is legal mid-derivation. */
+	#ensureProjections(): { all: AttributeEntry[]; nonZero: AttributeEntry[] } {
+		const values = this.attributeValues;
```
ginderjeremiah, line 130: Should-fix (maintainability hardening, non-blocking): this read is load-bearing for reactivity and must run on **every** call, including cache hits — it's what registers the `$derived` dependency on the reactive `attributeValues` so consumers re-derive when the totals change. But `values` is only *used* inside the `if (all === null …)` block below, so the obvious "tidy-up" is to move `const values = this.attributeValues;` into that block. That refactor would silently break tracking on the cache-hit path: a `$derived` reading `getAttributeMap()` on a cache hit would no longer depend on `attributeValues` and would go stale until an unrelated re-render — and no unit test would catch it (reactive tracking isn't exercised in isolation).

The method-level doc already explains *why* the read exists, but it doesn't pin it to this line's placement. A one-line inline note here would guard the exact failure mode, e.g.:

```ts
// Must read attributeValues unconditionally (not inside the if): on a cache hit this is
// the only thing registering the $derived dependency, so consumers re-derive on recompute.
const values = this.attributeValues;
```

---

ginderjeremiah (commented):
## Review: Lazily build `BattleAttributes` display projections (#1190)

**Verdict: Approve** — _(submitted as a comment because I'm the PR author; GitHub won't accept a self-approval as a gating event)_. The change does exactly what #1190 asked, the reactivity contract is correct and well-documented, and the test directly proves the hot-path win. One non-blocking hardening suggestion inline.

### Correctness — the subtle part holds up
The risk in this change is the `statify` reactivity contract, and it's handled correctly:
- `#ensureProjections()` reads `this.attributeValues` **unconditionally** at the top, so a `$derived` consumer reading `getAttributeMap()` registers the dependency on the reactive `attributeValues` even on a cache hit — it re-derives when the totals change.
- `recompute()` nulls both caches in the same step it reassigns `attributeValues`, so the invariant "cache non-null ⟹ built from the current `attributeValues`" always holds (no path mutates values without invalidating, or vice-versa).
- Moving the caches from `private` (statify-reactive) to `#` (non-reactive, per `statify.svelte.ts:18`) makes the lazy build's write legal mid-derivation — it isn't a `$state` write, so no `state_unsafe_mutation`.

I checked the consumers (`ItemTooltip.svelte:74`, `ItemDrawer.svelte:59` via `$derived`, plus the one-shot `inventory-view.svelte.ts:114`): all read through `getAttributeMap()`, so they all get the lazy build and stay correct.

### Should-fix
- `battle-attributes.ts:128` — the unconditional `const values = this.attributeValues;` is load-bearing for reactivity but only *used* inside the `if` block below, making it an easy target for a naive "move it into the block" simplification that would silently break `$derived` tracking on cache hits. Worth an inline comment pinning it in place. Details inline. Non-blocking.

### Test coverage
- Solid for a unit of this class. The new test asserts the projection is **not** rebuilt on the add/remove/add hot path (spying the `staticData.attributes` reference scan) and **is** rebuilt lazily on the next read, reflecting accumulated changes — directly proving the issue's win. Existing stable-reference and rebuild-after-change tests are retained.
- Gap (acknowledged, not blocking): the reactive-tracking guarantee itself — that a `$derived` re-runs on an `attributeValues` change even on a cache hit — isn't directly exercised, since that needs a Svelte reactive harness rather than a plain unit test. The inline hardening comment is the more practical guard against a regression there.

### What's good
- The class doc comment is updated thoroughly and accurately — it explains *why* the projections are non-reactive `#` caches and *why* the `attributeValues` read keeps consumers tracking. Exactly the kind of context this subtle construct needs.
- Correctly skipping a backend parity mirror: this touches only the display projection, not `attributeValues` (the parity-relevant array), so battle outcomes are unaffected. Right call.
- Tight, well-targeted change that resolves the hot-path waste #1190 identified with no behavior change.

---

UI/src/lib/battle/battle-attributes.ts
ginderjeremiah: Addressed in 3922075 — added the exact inline comment pinning the unconditional `attributeValues` read in place.

---

ginderjeremiah:
## Code Review — Approve

Clean, well-scoped perf fix that does exactly what #1190 asked: the per-modifier hot path (`recompute`) no longer rebuilds the display projections — it now only reassigns `attributeValues` and nulls the two caches, deferring the `.map` + per-attribute `attributeName` scans until a UI surface actually reads them.

### Reactivity contract — verified correct
The central claim holds up against `statify` (`UI/src/lib/common/statify.svelte.ts`):
- `statify` only walks **enumerable public** fields (`for...in`), so the JS-private `#attributeMap`/`#nonZeroAttributeMap` are invisible to it and stay non-reactive. Writing them in `#ensureProjections()` is **not** a `$state` write, hence legal mid-derivation. ✔
- `attributeValues` is declared TS-`private` (still enumerable) → it *is* statified/reactive, and `#ensureProjections()` reads it **unconditionally** (outside the `if`), so on a cache hit the read still registers the `$derived` dependency and consumers re-derive on the next `recompute`. This is the crux of correctness and it's both implemented right and called out in a comment. ✔

All three consumers (`ItemDrawer.svelte:59`, `ItemTooltip.svelte:74` — both `$derived` — and the throwaway construction in `inventory-view.svelte.ts:114`) read inside reactive/one-shot contexts, so no staleness risk.

### Strengths
- Memoization + invalidation are correct: nulled on `recompute`, rebuilt once on next read.
- Strong test coverage — the new test spies on the `staticData.attributes` getter to **prove** no projection rebuild across an add/remove/add hot-path sequence, then asserts the next read reflects the accumulated state. Existing stable-reference and rebuild-on-change tests retained.
- Comments are load-bearing and accurate (the doc-comment rewrite correctly re-narrates the eager-vs-lazy split).

### Nits (non-blocking)
- `battle-attributes.ts:141` — `all.filter((entry) => entry.value != 0)` uses loose `!=`; the test uses strict `!== 0`. Values are always numbers so behavior is identical, but strict `!==` would be more consistent and idiomatic.
- `battle-attributes.ts:138` — `all === null || nonZero === null` is mildly redundant since both are always nulled together in `recompute`; harmless defensive check.

### CI
All checks green (`test-frontend`, `verify-frontend`, e2e chromium/firefox/webkit all pass; backend jobs correctly skipped for a frontend-only diff).

No must-fixes. Approving.

---

Issue #982, opened 2026-06-19
Add proficiency system


ginderjeremiah:
> **Spike complete — design decided.** Full write-up: [`docs/spikes/982-proficiency-system.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/982-proficiency-system.md).
>
> **Reframe:** proficiency is a mastery *layer* over world-sourced skills (starter / item / milestone), **not** a competing skill faucet. It pays out attribute bonuses, new skills, and access to deeper nodes in a tree (DAG). Victory-based fixed-pie, difficulty-scaled XP; multi-contribution skills; depth-scaled power + the loadout/gear slots make going wide costly. Challenges stop granting skills.

### Implementation sub-issues (dependency order)

1. #1114 — reference-data model + skill contributions + admin authoring *(foundation)*
2. #1115 — PlayerProficiency persistence *(→ #1114)*
3. #1116 — XP accrual + leveling, live + offline *(→ #1114, #1115)*
4. #1117 — bonuses → attribute pipeline (parity) *(→ #1114, #1115)*
5. #1118 — milestone & unlock effects + open triggers *(→ #1114, #1115, #1116)*
6. #1119 — client reference + player data delivery *(→ #1114, #1115)*
7. #1120 — client Proficiencies tree screen *(→ #1119)*
8. #1121 — client XP / level-up / milestone feedback *(→ #1116, #1118, #1119)*
9. #1122 — remove skill rewards from challenges + re-home *(→ #1118)*

### Deferred / standalone trackers

- #1123 — tier/rarity-weighted XP (after #979)
- #1124 — gear-gating by proficiency (fast-follow)
- #1125 — spike: skill synthesis
- #1126 — spike: class system
- #1127 — author the initial proficiency tree (content)

---

### Original framing (superseded by the spike — kept for history)

The proficiency system (at least in its current form) may require significant changes to the core game systems.

The idea is that the proficiency system will replace challenges as a way of unlocking skills. Under the hood, it may use the statistics system (and thus be very similar to challenges), but the idea is a Skyrim-esque progression. Skills will need to be categorized into Proficiencies (maybe extending the tags feature?), and then unlocking skills will be gated by experience in the given Proficiency. Experience would be gained by using skills related to that Proficiency.

PR #1202, opened 2026-06-24
Freeze retired proficiency paths in XP accrual; dedup new-player seed skills against actual starter set (#1176)

### LINKED ISSUE #1176: Proficiency backend follow-ups: retired proficiencies/paths still accrue, NewPlayerFactory seed dedup is positional
Found during a code-health review of the proficiency system (#982). Two smaller backend gaps grouped together.

## 1. Retired proficiencies/paths still accrue XP and grant skills (design gap)

The proficiency snapshot loads all proficiencies and paths regardless of `RetiredAt` (`Game.DataAccess/Repositories/Caching/ProficienciesCacheHolder.cs:40-47`, `73-77`), and the core `Proficiency`/`Path` models don't carry `RetiredAt` at all. So `ProficiencyRewardService.BuildContributions`/`AccrueAndApply` keep routing XP into, leveling, and granting milestone skills from a retired proficiency.

This is the outlier vs. the rest of the retirement model:
- `EnemiesCacheHolder` excludes retired enemies from random selection,
- `PlayerProgress.EvaluateChallenges` stops progressing retired challenges,
- `AccountService.RootSeedSkillIds` already filters `RetiredAt is null`.

The spike lists `RetiredAt` on both `Path` and `Proficiency` as authored fields but never says retirement halts accrual. **Decision needed:** if retirement should freeze a track (matching the challenge/enemy precedent), the snapshot/accrual must honor it (and the core models need the flag, or the holder must filter); if proficiencies are deliberately different, document why.

## 2. `NewPlayerFactory` seed-skill dedup is a positional assumption (minor/refactor)

`Game.Core/Players/NewPlayerFactory.cs:37-45` filters root seed skills with `Where(id => id >= StarterSkillCount)` to avoid duplicating starter skills, assuming starter skills are exactly ids `0..StarterSkillCount-1`. This holds under the 0-based-identity convention, but couples seed dedup to a positional assumption rather than the actual starter id set: if starter id assignment ever changes, a root seed skill with id `< StarterSkillCount` that *isn't* a starter would be silently dropped. Consider deduping against the actual starter id set.

scope: small

ginderjeremiah:
Closes #1176.

Two backend proficiency follow-ups from the #982 code-health review.

## Part 1 — Retired paths now freeze (no accrual, no leveling, no skill grants)

Per the decision confirmed on the issue: **freeze a retired track**, keep retirement **out of the Core domain**, and **filter in the contributions-by-skill mapping**.

The battle XP accrual reads exactly one reference structure to decide what to train: the skill→contribution reverse index (`ContributionsForSkill`), built in `ProficienciesCacheHolder`. That is the single routing choke point — `BuildContributions` only ever routes XP into paths a fired skill contributes to. Excluding retired paths from that index there means:

- no skill routes XP into a retired path → the track **accrues nothing**,
- nothing levels → no **milestone skill grants**,
- no tier maxes → no **successor opening**.

This matches the existing retirement precedents (`EnemiesCacheHolder` drops retired spawns, `PlayerProgress.EvaluateChallenges` skips retired challenges, `AccountService.RootSeedSkillIds` filters `RetiredAt is null`) and the documented rule that retirement is *"a catalogue/authoring concern only — not carried on the lean battle models."* The core `Proficiency`/`Path` models stay retirement-agnostic; the filter lives entirely in the cache holder.

**Granularity:** the **path** is the freeze unit. A path's tiers progress as a contiguous-ordinal frontier walk (`Path.Frontier`/`NextTier`), so retiring an individual *mid-path* proficiency is structurally incoherent (it would soft-lock the walk). "Freeze a track" therefore means retiring the path.

**Already-accrued state:** untouched — the filter only stops *further* routing, so existing levels and granted skills remain (matching "already-earned references are never revoked").

**Parity:** none required. Proficiency XP is computed server-side at battle completion, off the frontend/backend parity surface (`docs/backend.md` "Proficiency XP accrual"). Only the eventual attribute *bonus* is parity-sensitive, and it is unaffected.

## Part 2 — `NewPlayerFactory` seed-skill dedup is no longer positional

`Create` dropped root seed skills with `Where(id => id >= StarterSkillCount)`, assuming starter skills are exactly ids `0..StarterSkillCount-1`. It now dedups against the **actual starter id set** derived from the starter skills it just built, so a future change to starter id assignment can't silently drop a non-starter root seed skill with a low id. Behavior is identical under today's 0-based starter ids; this is a robustness refactor.

## Testing

- Full `Game.Core.Tests` (810) and `Game.Application.Tests` (552) suites pass.
- New integration test `RetiredPath_FreezesTheTrack_AccruesNothingAndGrantsNoSkill`: a skill contributing to a retired path (whose tier would otherwise max and grant a cap milestone reward on one victory) accrues nothing and grants no skill.
- Existing `NewPlayerFactory` dedup tests still pass (behavior-preserving refactor).

## Docs

`docs/backend.md` retirement lifecycle section now documents path-retirement freeze and that a frozen track's already-accrued progress is preserved.

## Follow-up

Filed #1203 for a related edge discovered here (out of this issue's stated accrual scope): a retired path's tier-0 can still be *opened* as a cross-path gateway when its live prerequisites max, granting its seed skill even though the track can never be trained.

https://claude.ai/code/session_015Ri5QiLQbxdR831FHFMzdp


ginderjeremiah (commented):
## Review: Freeze retired proficiency paths in XP accrual; non-positional new-player seed dedup (#1176)

**Verdict:** Approve — both parts are correct, minimal, and well-placed; no blocking issues. (Submitting as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on my own PR — the intended verdict is **Approve**.)

I verified the central claim that `ContributionsBySkill` is the single routing choke point: accrual flows exclusively through `ProficiencyRewardService.BuildContributions` → `IProficiencies.ContributionsForSkill`, and that accessor is consumed nowhere else but accrual and its own DataAccess tests. Filtering retired paths out of the reverse index in the holder therefore freezes the track end-to-end (no XP routed → no leveling → no milestone grants → no successor opening) without touching `paths`/`corePaths`/`entities`, so id-contiguity (`AssertZeroBasedContiguity`) and `GetById` resolution of already-accrued state are preserved. Keeping the core `Proficiency`/`Path` models retirement-agnostic and filtering only in the cache holder is the right call and matches the documented retirement model and the three existing precedents.

### Should-fix
- _None._

### Nits / suggestions (optional)
- **Direct change-site test for the freeze** — the freeze is covered end-to-end by the new Application-layer test, which is the most important coverage. Consider *also* adding a case to `Game.Application.Tests/DataAccess/ProficienciesIntegrationTests.cs` asserting `ContributionsForSkill(skillId)` returns empty for a skill that only contributes to a retired path. That pins the behavior at the exact line you changed (`ProficienciesCacheHolder.cs:104`) alongside the existing `ContributionsForSkill_*` cases, so a future refactor of the reverse index can't silently regress it.
- **Part 2 robustness isn't pinned by a test** — the dedup refactor is behavior-preserving under today's 0-based starter ids, so the existing tests still pass, but nothing exercises the *new* property (a non-starter root seed skill with a low id is no longer dropped). That's hard to test today because `starterSkills` is hardcoded as `Enumerable.Range(0, StarterSkillCount)`, so there's no seam to inject a non-positional starter set — fine to leave as-is, just flagging that the robustness win is currently unverified. Worth a quick comment or a follow-up if starter-id assignment is ever expected to change for real.

### Test coverage
- **Part 1:** Solid. `RetiredPath_FreezesTheTrack_AccruesNothingAndGrantsNoSkill` is a clean integration test, and its `(cap 1, cost 1, level-1 reward)` setup is the exact mirror of the live-path baselines (`LivePath_RaisesTheEventCarryingGrantedSkillsAndOpenedTiers`, `MaxingATier_OpensAndSeedsTheNextTierInItsPath`) — so the only varied input is `RetiredAt`, which makes the contrast meaningful. Asserting empty `results` *and* the single owned skill closes both the accrual and the grant.
- **Part 2:** Relies on existing dedup tests (behavior-preserving) — acceptable; see the nit above.
- **Parity:** Correctly N/A. `docs/backend.md` ("Proficiency XP accrual") confirms XP is computed server-side at battle completion, off the parity surface; only the eventual attribute *bonus* is parity-sensitive and it's untouched.

### What's good
- Surgical, single-line filter at exactly the right layer; the reasoning for path-as-freeze-unit (contiguous frontier walk would soft-lock on a mid-path retire) is sound and well-documented in the code comment and `backend.md`.
- Excellent follow-up hygiene: you discovered the residual cross-path-gateway edge (a retired tier-0 can still be opened via `OpenSuccessors`/`DependentsOf` when live prerequisites max), correctly scoped it *out* of #1176, and filed #1203 with a concrete fix and an open design question. That's the honest call rather than quietly half-fixing the freeze.
- `docs/backend.md` retirement section updated accurately, including the "already-accrued progress is preserved" nuance.


---

ginderjeremiah:
## Code Review — Approve

Verdict: **Approve.** A clean, well-reasoned pair of proficiency follow-ups that lands #1176 exactly as scoped. Both changes are minimal, the choke-point reasoning is sound, the domain stays isolated, and the discovered edge is correctly carved out as #1203 rather than smuggled in. CI green (`all-checks-passed` ✅; `verify-backend`, `test-backend`, all `test-e2e` browsers pass — the legacy combined-status "pending" is just an unused status API, the check-runs are authoritative).

### Strengths

- **Choke-point is correctly identified.** `ContributionsForSkill` is genuinely the single routing entry for *accrual* — `BuildContributions` only routes XP into paths a fired skill contributes to, so filtering retired paths out of the reverse index in `ProficienciesCacheHolder.BuildSnapshotAsync` cleanly gives "accrues nothing → levels nothing → grants no milestone skill" with one `.Where(path => path.RetiredAt is null)`. Verified against `ProficiencyRewardService.AccrueAndApply`.
- **Domain isolation preserved.** Core `Path`/`Proficiency` carry no `RetiredAt` (verified); the freeze lives entirely in the data-tier cache holder, matching the documented rule that retirement is a catalogue/authoring concern not carried on lean models. Consistent with the cited precedents (`EnemiesCacheHolder`, `EvaluateChallenges`, `RootSeedSkillIds`).
- **Path-as-freeze-unit reasoning is correct.** `Path.Frontier`/`NextTier` is a contiguous-ordinal walk, so retiring a mid-path tier would soft-lock it — retiring the path is the coherent granularity. Good call.
- **Parity correctly assessed as N/A.** Proficiency XP is computed server-side off the battle-parity surface (docs/backend.md + game-design.md §93); only the attribute *bonus* is parity-sensitive and it's untouched. No frontend change needed.
- **#1203 is the right move.** `OpenSuccessors` → `DependentsOf`/`DependentsByProficiency` is a genuine *second* skill-grant path the contributions filter does not cover (a retired tier-0 opened as a cross-path gateway still grants its `SeedSkillId`). Verified the gap is real and out of #1176's stated accrual scope — filing it rather than expanding scope is correct.
- **Test coverage is appropriate.** Domain-adjacent integration test at the index level (`ContributionsForSkill_ReturnsEmpty_WhenSkillOnlyFeedsARetiredPath`) plus the end-to-end accrual assertion (`RetiredPath_FreezesTheTrack_AccruesNothingAndGrantsNoSkill`) covers both the mechanism and the observable behavior. Docs updated in the right section.

### Nits (non-blocking)

- **`NewPlayerFactory.cs`** — the new comment says the dedup "stays correct if starter id assignment ever changes," but `starterSkills` is still built from `Enumerable.Range(0, StarterSkillCount)` in the same method, so starter ids *are* still positionally `0..N-1`. The refactor correctly removes the positional assumption from the **dedup** (it now keys off the actual built set), which is the real robustness win — but the starter ids themselves remain positional, so the "if assignment ever changes" framing slightly overstates it. Also `StarterSkillCount`'s own XML doc still reads "skill ids 0..N-1". Worth a one-line tweak to keep the comment honest about what's positional vs. not.
- **`ProficiencyRewardServiceTests.cs:~218`** — `Assert.Empty(results)` + `Assert.DoesNotContain(...)` + `Assert.Single(player.Skills)` is a thorough triple-assert; the inline comment is clear. No change needed, just noting the `Single` assertion is the load-bearing one (it's what would catch a stray gateway grant) — fine as-is.

No must-fixes. The dedup refactor is behavior-preserving today and the freeze logic is correct.

---

PR #1204, opened 2026-06-24
Fix dropped deep-link payload on same-screen navigation; guard skills-view rollback baseline (#1191)

### LINKED ISSUE #1191: Frontend small correctness follow-ups (deep-link payload, skills-view commit baseline)

ginderjeremiah:
Closes #1191.

Two small frontend correctness follow-ups from the code-health review. **Part 1 is a fix; Part 2 is a reasoned decline with a guard test — please confirm the Part 2 stance.**

## Part 1 — same-screen deep-link payload no longer dropped ✅

`navigation.requestScreen(key, payload)` hands its one-shot payload to the target via `consumePayload()` **on mount**. The shell's `handleNavigate` only sets `currentScreen = key`, so a deep-link to the screen that's **already active** never remounts the component, and `onMount`/`consumePayload` never runs — the payload is silently lost.

The shell now keys the active screen on a `screenNonce` it bumps when a request targets the already-active screen **and** a payload is queued, forcing a remount so the target re-consumes the handoff:

- `navigation` exposes `hasPendingPayload` (set when `requestScreen` is called with a payload, cleared on `consumePayload`).
- A pure `requiresRemount(requested, active, hasPendingPayload)` helper encodes the decision (unit-tested).
- The shell wraps the screen in `{#key screenNonce}` and bumps the nonce in the navigation effect for a same-screen payload-bearing request.

A **cross-screen** switch is unaffected — the dynamic `<CurrentScreen />` remounts itself when the component changes, and the nonce only bumps for the same-screen case, so no spurious remount of the outgoing screen. The one current caller (Statistics → Codex) is always cross-screen, so today's behavior is unchanged; this closes the latent gap for any within-target deep-link.

## Part 2 — `committed` baseline: investigated, **did not** apply the proposed change ⚠️

The issue proposed tightening `SkillsView.commit()` so `committed` advances **only for the latest** in-flight edit (mirroring the rollback's `isLatest` guard). I worked through it and believe that change would **introduce a rollback bug**, so I left the logic as-is and added a guard test + clarifying comment instead.

Persists are serialized on `SerializedQueue`, so when several edits overlap they apply strictly in order. Consider two rapid edits where the **earlier** persist succeeds and the **latest** then fails:

- `toggle(0)`: `[0,1,2] → [1,2]` — **succeeds**, so the server now holds `[1,2]`.
- `toggle(3)`: `[1,2] → [1,2,3]` — latest, **fails**.

The latest's failure must roll back to the server's actual current state, `[1,2]`. The current code reaches that because every success advances `committed`, so by the time the latest fails `committed === [1,2]`. The proposed "advance only for the latest" never advances `committed` for the earlier success, so it would roll back to the pre-sequence `[0,1,2]` — **past a server-confirmed loadout**, diverging local state from the server.

In other words, under serialization `committed` advancing on every success is precisely what keeps the rollback baseline equal to "the last loadout the server confirmed" — which is the semantics the issue is after. I added a regression test (`rolls back to an earlier overlapping commit that succeeded when the latest fails`) that passes today and would fail under the proposed change, plus a comment documenting the invariant.

If you still want the change after this, I'm happy to revisit — but I'd want to first see a scenario where the current behavior actually rolls back to the wrong loadout. Flagging on the issue too.

## Testing

- `npm run lint` (eslint + svelte-check) — clean, 0 errors / 0 warnings.
- `npm run test:unit:ci` — full suite green (233 files / 2484 tests), including the new `navigation` store tests, the `requiresRemount` tests, and the Part 2 rollback regression test.

## Docs

`docs/frontend.md` → _Cross-screen navigation_ now notes the same-screen deep-link remount guard.

https://claude.ai/code/session_01RdDDAmoYX8PibUtU2jFF72

---
_Generated by [Claude Code](https://claude.ai/code/session_01RdDDAmoYX8PibUtU2jFF72)_


ginderjeremiah:
## Code review — PR #1204

**Verdict: ✅ Approve (with one optional nit).** Both parts are correct, well-tested, and well-documented. CI is fully green (`all-checks-passed`, frontend unit + verify + all three e2e browsers).

### Part 2 decline — I agree, the reasoning is sound ✔️

I traced the argument against the actual `SerializedQueue` (`UI/src/lib/common/serialized-queue.ts:22-26`): `run` chains each operation off `this.tail`, so an earlier persist — *including its synchronous `committed = next` advance* — fully settles before the later one starts. That serialization is exactly what the Part 2 argument leans on, and it holds.

Walking the scenario in `commit` (`skills-view.svelte.ts:456-479`):
- `toggle(0)` succeeds → `committed = [1,2]` (server now holds `[1,2]`).
- `toggle(3)` is latest, fails → rolls back to `committed`, which is correctly `[1,2]` — the server's true current state.

Under the proposed "advance `committed` only when `isLatest`", `toggle(0)`'s success would *not* advance the baseline (its `seq !== commitSeq` by then), so the `toggle(3)` failure would roll back to the pre-sequence `[0,1,2]` — past a server-confirmed loadout, diverging local from server. So advancing on **every** serialized success is precisely what keeps `committed` == "last loadout the server confirmed". Declining the issue's proposal is the right call, and the new regression test (`skills-view.test.ts:463-476`) is a real guard that would fail under the proposed change. The clarifying comment at `skills-view.svelte.ts:470-473` documents the invariant well.

### Part 1 — correct and cleanly factored ✔️

- The decision is extracted into a pure, unit-tested `requiresRemount` (`navigation.svelte.ts:17-19`) rather than inlined — good.
- Effect ordering in `+page.svelte:118-128` is right: it captures `remount` from `hasPendingPayload` *before* `handleNavigate`, then bumps `screenNonce`, then `clear()`. `clear()` only nulls `requestedScreen` and leaves `payloadPending` intact, so the forced remount's `onMount`/`consumePayload` still finds the payload.
- Cross-screen case is correctly excluded (`requiresRemount` returns false), so the dynamic `<CurrentScreen>` remounts on its own and the nonce never bumps spuriously.
- `null` is correctly treated as a present payload (`payload !== undefined`), and the test at `navigation.test.ts:46-50` locks that in.
- Doc update in `docs/frontend.md` → *Cross-screen navigation* is accurate and concise.

### Nits (optional)

- **`payloadPending` is a plain `let`, not `$state`** (`navigation.svelte.ts:12`), yet `hasPendingPayload` is read inside the shell's `$effect`. This is *fine today* because the effect is driven by `requestedScreen` (which is `$state`) and `requestScreen` sets both synchronously — so the flag is always current when the effect reads it. But it's a subtle coupling; a one-line comment noting that the flag is only ever read in lockstep with the reactive `requestedScreen` would save a future reader the same trace I just did. Not blocking.

### No must-fixes, no follow-up issues needed.

Nice work resisting the issue's suggested change and instead proving why the current behavior is correct — that's the right instinct, and the guard test makes the invariant explicit. Closing #1191 is appropriate.

---

PR #1205, opened 2026-06-24
Harden data-access & caching edges (#1173)

### LINKED ISSUE #1173: Data access & caching robustness follow-ups
Found during a code-health review — a basket of small robustness gaps in the data-access and caching layer. The core patterns (build-then-swap snapshots, write-behind idempotency, drain serialization) were reviewed and found sound; these are the edges. Grouped as in #1146.

## Correctness

- **`PlayerUpdateEventDispatcher` keys handlers by the *unqualified* type name** (`Game.DataAccess/PlayerUpdates/PlayerUpdateEventDispatcher.cs:37-55`; publisher `PlayerPersistencePublisher.cs:52` uses `GetType().Name`). Two persisted event types sharing a simple name across namespaces would silently overwrite each other in the `ConcurrentDictionary` and route one type's writes to the wrong handler with no error. Latent (all current names are unique). Either use `FullName` or assert uniqueness at registration (throw on a re-register to a different handler). Note: the key couples to the serialized envelope `Type`, so this is a wire-format change (fine pre-release).

## Convention

- **Snapshot read-once idiom.** Several reference-repo methods read `holder.Current` (`Snapshot`) more than once per logical operation, or mix one holder's snapshot with another's (`Game.DataAccess/Repositories/Enemies.cs` — `GetRandomEnemyId` reads the enemy snapshot then the zones holder; `All()` etc.). A build-then-swap between reads means one call can mix old+new snapshots. Harm is benign today (worst case: a just-retired enemy spawns once), but it's a latent foot-gun. `GetRandomDomainEnemy` already models the safe idiom (`var snapshot = Snapshot;` once at the top). Apply consistently across `Enemies`/`Items`/`ItemMods`/`Proficiencies` and note the convention in `backend.md`.

## Minor / fold-in

- **`AdminCacheReloadFilter` reload is unbounded and broadcasts before the local reload** (`Game.Api/Filters/AdminCacheReloadFilter.cs:45-57`). `Task.WhenAll(... ReloadAsync())` has no timeout, so a wedged DB connection holds the admin request open indefinitely; and `NotifyChangedAsync` runs *before* the local reload, so a local-reload failure leaves the serving instance as the *only* stale one (the inverse of the intended read-your-writes). Add a bounded timeout and a comment acknowledging the asymmetry.
- **`ReferenceDataVersioning` comment overstates `ConditionalWeakTable.GetValue` guarantees** (`Game.Api/Sockets/Commands/ReferenceDataVersioning.cs:19-20`): the factory may run more than once under concurrent first-touch (behaviorally harmless — the hash is idempotent). Correct the comment.
- **`EnemiesCacheHolder` re-maps every skill per (enemy, skill) pair at reload** (`Game.DataAccess/Repositories/Caching/EnemiesCacheHolder.cs:47-61`, `EnemyMapper.ToTemplate`). Map each distinct skill once into a `Dictionary<int, CoreSkill>` and share within the build. Reload-time only (low impact), easy win.
- **`ProgressUpdatedHandler` statistics upsert loads the `typeIds × entityIds` cross-product** (`Game.DataAccess/PlayerUpdates/Handlers/ProgressUpdatedHandler.cs:40-49`) — a superset of the changed pairs, on the hot drain path. Correct but over-fetches; if profiling shows it, switch to an exact-pair `INSERT ... ON CONFLICT` upsert. Future optimization, not urgent.

---

> **Edit:** dropped an earlier bullet about the fire-and-forget `SetAndForget`/`ExpireAndForget` cache writes (`RedisService.cs`) swallowing failures. The fire-and-forget choice is deliberate to keep the cache write off the player-mutation latency path; adding observability would mean dropping `CommandFlags.FireAndForget` (awaiting adds latency, defeating the purpose). The live session reads/writes the in-memory aggregate, not the cache, so a dropped write is invisible to the session and overwritten by the next `SavePlayer` — the only exposure is a rehydration falling in the failed-write→next-write gap, and durability is unaffected (the event-queue write is awaited). Too narrow and self-healing to track.

scope: small

ginderjeremiah:
Closes #1173.

A basket of small robustness/consistency follow-ups from the data-access & caching code-health review. Each is independently low-risk; grouped per the issue.

## Correctness — `PlayerUpdateEventDispatcher` unqualified-name collision
The dispatch key is the **unqualified** type name the persisted envelope carries (`GetType().Name`). Two persisted event types sharing a simple name across namespaces would silently overwrite each other in the `ConcurrentDictionary` and misroute one type's writes to the other's handler with no error.

Rather than switch to `FullName` (a wire-format change that would orphan in-flight queued envelopes), registration now **fails loud** on a collision: `Register<TEvent>` throws `InvalidOperationException` if a *different* type is registered under an already-registered simple name. Re-registering the **same** type stays idempotent (the normal startup path). This converts the latent silent-overwrite into a startup failure and keeps the wire format unchanged.

## Convention — snapshot read-once idiom
`Items`/`ItemMods`/`Proficiencies` read `holder.Current` through a property getter that re-reads on each access. Since a build-then-swap replaces the snapshot reference atomically, a method touching the snapshot more than once could mix an old and a new snapshot. Each repo now exposes a single `Snapshot => holder.Current` accessor and reads every field off it once per operation (matching `Enemies`, which already modeled the safe idiom). Documented in `docs/backend.md`.

## Minor / fold-in
- **`AdminCacheReloadFilter`**: the awaited local reload is now bounded by a 30s timeout (`Task.WhenAll(...).WaitAsync`) so a wedged DB connection can't hold the admin request open indefinitely; on timeout a `TimeoutException` surfaces on the admin response (the write persisted; retry re-runs the reload). Added a comment documenting the accepted broadcast-before-reload asymmetry (a local-reload failure after a successful broadcast leaves the serving instance as the only stale one until its next write/notification).
- **`ReferenceDataVersioning`**: corrected the comment overstating `ConditionalWeakTable.GetValue` — the factory may run more than once under concurrent first-touch (only one result is published); harmless here because the hash is a pure, idempotent function of the snapshot.
- **`EnemiesCacheHolder`/`EnemyMapper`**: a build-scoped `Dictionary<int, CoreSkill>` is shared across all templates so each distinct skill is mapped once, not once per `(enemy, skill)` pair. Safe because reference-data skills are immutable value objects (already shared by id elsewhere). The loud out-of-range loud-fail (naming the enemy + id) is preserved on the first touch of each skill.

## Intentionally not done
The `ProgressUpdatedHandler` `typeIds × entityIds` cross-product bullet is left as-is — the issue marks it a profiling-gated future optimization ("Future optimization, not urgent"), and the existing comment already documents the over-fetch rationale.

## Testing
- Full backend suite green: `Game.Core.Tests` (810), `Game.Infrastructure.Tests` (59), `Game.Application.Tests` (553), `Game.Api.Tests` (611).
- New `PlayerUpdateEventDispatcher` unit tests: a different type with a colliding simple name throws; re-registering the same type is idempotent.
- The enemy skill-sharing change is behavior-preserving (reload-time only) and covered by the existing `EnemiesIntegrationTests`.

## Docs
`docs/backend.md` → build-then-swap section now documents the reload timeout and the snapshot read-once idiom.

https://claude.ai/code/session_01CM6ERcwnSu6gK9fqjCsKB6

---
_Generated by [Claude Code](https://claude.ai/code/session_01CM6ERcwnSu6gK9fqjCsKB6)_


ginderjeremiah:
## Code review — Approve

A tight, well-scoped robustness pass that lands exactly the basket from #1173. Each change is independently low-risk, faithful to the documented architecture, and the one user-visible behavior change (collision → loud fail) is correct and tested. CI is fully green (backend build/test, e2e chromium/firefox/webkit, `all-checks-passed`).

### Strengths
- **`PlayerUpdateEventDispatcher` collision guard** is the right call: keeping the wire key as the unqualified name (no orphaned in-flight envelopes) and converting the latent silent-overwrite into a startup `InvalidOperationException` via `AddOrUpdate`, while staying idempotent for the same type. The exception message is genuinely diagnosable (names both `FullName`s and the fix). Tests cover both the colliding-different-type throw and the same-type idempotency.
- **Snapshot read-once idiom** is applied consistently across `Items`/`ItemMods`/`Proficiencies`, each now binding a single `Snapshot => holder.Current` accessor and reading every field off it — matching the safe idiom `Enemies` already modeled. Documented in `docs/backend.md` so the convention is enforceable going forward.
- **`AdminCacheReloadFilter`** 30s `WaitAsync` bound is correct (deliberately not tied to the request token since the write has committed), and the accepted broadcast-before-reload asymmetry is now documented in-code rather than left implicit.
- **`EnemyMapper`/`EnemiesCacheHolder`** build-scoped skill cache is behavior-preserving (reload-time only) and the loud out-of-range fail is preserved — a cache hit can only be a previously range-validated id, so corruption still throws on first touch.
- The `ReferenceDataVersioning` comment correction (factory may run >1× under concurrent first-touch, harmless because the hash is idempotent) accurately reflects `ConditionalWeakTable.GetValue` semantics.
- Good judgment leaving the `ProgressUpdatedHandler` cross-product alone — the issue marks it a profiling-gated future optimization and the existing comment already documents the over-fetch.

### Nits (non-blocking)
- `PlayerUpdateEventDispatcherTests`: the static `_handlers` registry is process-wide, so the new collision test relies on `NamespaceA.Collide`/`NamespaceB.Collide` being unique simple names not registered elsewhere — which holds today. Worth a one-line note that the test depends on the simple name being unused process-wide (the existing `FakeUpdateEvent` registrations already share this implicit coupling, so it's pre-existing, not introduced here).
- `Proficiencies.LookupProficiencyByTier` does a linear `FirstOrDefault` scan per call off the snapshot — untouched by this PR and out of scope, just flagging it as a latent item if it ever lands on a hot path.

Neither blocks merge. Nice clean follow-up. 

— Reviewed against `docs/backend.md` (Reference Data, build-then-swap, entity isolation) and CLAUDE.md.

---

PR #1206, opened 2026-06-24
Backend small cleanups: dead field, enum-count coupling, dup lookup, dead local (#1192)

### LINKED ISSUE #1192: Backend small cleanups: dead/denormalized fields, brittle enum-count coupling, ChallengeBoss client zone
Found during a code-health review — a basket of small, low-risk backend cleanups across `Game.Core` and `Game.Api`. Grouped as in #1146/#1173.

## Cleanups

- **`SkillStats.SkillId` is a dead, denormalized field** (`Game.Core/Battle/BattleStats.cs:15`, written at `Game.Core/Battle/BattleContext.cs:156`). It is set on creation but never read by production code — both consumers (`PlayerProgress.RecordBattleCompleted`, `ProficiencyRewardService`) key off the `Dictionary<int, SkillStats>` key, not this field. It only duplicates the dictionary key (and can silently disagree with it) and is read only in tests. Remove it (or drop the dictionary in favour of a list keyed on it).

- **`Inventory.EquipSlots` conflates "max enum value" with "slot count"** (`Game.Core/Players/Inventories/Inventory.cs:11`, used at `:223`). `EquipSlots = (int)Enum.GetValues<EEquipmentSlot>().Max()` plus a `Range(0, EquipSlots + 1)` fix-up is correct only because `EEquipmentSlot` is contiguous and 0-based; the max-as-count coupling and the `+ 1` are brittle against any future renumbering. Derive the slot set directly from `Enum.GetValues<EEquipmentSlot>()` (or at minimum add a comment pinning the assumption).

- **Duplicated slot lookup in `Player.TryUnequipItem`** (`Game.Core/Players/Player.cs:305-321` + `Game.Core/Players/Inventories/Inventory.cs:112`). The player scans `EquipmentSlots.FirstOrDefault(s => s.Value == slot)` to capture the unequipped item id, then `Inventory.TryUnequipItem(slot)` performs the identical scan again. Tiny collection (not a perf issue) — have the inventory method return the unequipped item id so the caller doesn't re-derive it.

- **Dead local in `SocketHandler.ExecuteServerCommand`** (`Game.Api/Sockets/SocketHandler.cs:137`): `var now = Stopwatch.GetTimestamp();` is assigned and never read (the real timing lives in `RunCommandUnderLock`'s `finally`; the sibling `ExecuteCommand` has no such line). Remove it.

## Confirm intended (not clearly a defect)

- **`ChallengeBoss` accepts a client-supplied `ZoneId`** (`Game.Api/Sockets/Commands/ChallengeBoss.cs:30`: `Parameters.ZoneId ?? player.CurrentZoneId`). The auto-loop sibling `SetAutoChallengeBoss` deliberately ignores any client zone and validates `player.CurrentZoneId` as anti-cheat, but the interactive command lets the client name any zone. `StartBossBattle` still gates on unlocked + not-retired (`BattleService.cs:177-188`), so it can't reach locked content — a player can only challenge a boss of a zone they've **already unlocked** but aren't standing in. This may well be intended (farm a previously-cleared boss without walking back), but it's an undocumented asymmetry with the auto-loop. Decide: keep it (and note it as intended), or drop the parameter and use `player.CurrentZoneId`.

ginderjeremiah:
## Summary

Addresses the basket of small, low-risk backend cleanups in #1192.

### Clear cleanups (done)

- **Dead `SkillStats.SkillId`** — removed the denormalized field on `Game.Core/Battle/BattleStats.cs`. It duplicated the `Dictionary<int, SkillStats>` key and was never read by production code (only tests). `BattleContext.RecordSkillUse` now constructs `new SkillStats()`; the tests that set/read the field were updated.
- **`Inventory.EquipSlots` enum-count coupling** — `NewEquippedList` now derives one slot per `EEquipmentSlot` value directly from `Enum.GetValues<EEquipmentSlot>()`, dropping the brittle `(int)…Max()` + `Range(0, EquipSlots + 1)` pattern that was correct only while the enum stays contiguous and 0-based. The `EquipSlots` static field is gone.
- **Duplicated slot lookup in `Player.TryUnequipItem`** — `Inventory.TryUnequipItem` now returns the unequipped item id (`int?`) instead of `bool`, so the player no longer re-scans `EquipmentSlots` to capture the id for the `ItemUnequippedEvent`.
- **Dead local in `SocketHandler.ExecuteServerCommand`** — removed the unused `var now = Stopwatch.GetTimestamp();` (the real timing lives in `RunCommandUnderLock`'s `finally`).

### Confirm-intended item — `ChallengeBoss` client zone

This one is a product decision, not a defect. `ChallengeBoss` accepts a client-supplied `ZoneId` (`Parameters.ZoneId ?? player.CurrentZoneId`) while the auto-loop sibling `SetAutoChallengeBoss` always uses the current zone. Since `StartBossBattle` still gates on the zone being **unlocked and not retired**, the parameter can't reach locked content — a client can only name a zone it has already unlocked (e.g. to re-farm a previously-cleared boss without walking back). That reads as a deliberate, player-favouring choice consistent with the project's other documented asymmetries, so I took the **non-behavior-changing** option: kept it and documented the asymmetry with a code comment.

👉 If you'd rather **drop the parameter** and force `player.CurrentZoneId` instead, that's a one-line change — happy to flip it. Flagging for your call.

## Test plan

- [x] `Game.Core.Tests` — 810 passed
- [x] `Game.Application.Tests` — 555 passed (Postgres/Redis integration)
- [x] `Game.Api.Tests` — 611 passed (integration)

No frontend changes (backend-only); the frontend battle code has no per-skill `SkillStats`/`skillId` equivalent, so removing the dead field has no parity impact.

Closes #1192

https://claude.ai/code/session_0194RnT6n3kY8PKpswJv9fpd

---
_Generated by [Claude Code](https://claude.ai/code/session_0194RnT6n3kY8PKpswJv9fpd)_

### DIFF
--- a/Game.Api/Sockets/Commands/ChallengeBoss.cs
+++ b/Game.Api/Sockets/Commands/ChallengeBoss.cs
@@ -27,6 +27,11 @@ namespace Game.Api.Sockets.Commands
         {
             var state = context.Session.PlayerState;
             var player = context.Session.Player;
+            // Deliberate asymmetry with the auto-loop sibling SetAutoChallengeBoss (which always uses the
+            // current zone): the interactive challenge lets the client name a zone so a player can re-farm a
+            // previously-cleared boss without walking back. This can't reach locked content — StartBossBattle
+            // still gates on the zone being unlocked and not retired — so a client can only name a zone it has
+            // already unlocked.
             var zoneId = Parameters.ZoneId ?? player.CurrentZoneId;
 
             var result = await _battleService.StartBossBattle(player, state, zoneId, Parameters.ClientBattleMs, cancellationToken);
--- a/Game.Api/Sockets/SocketHandler.cs
+++ b/Game.Api/Sockets/SocketHandler.cs
@@ -134,7 +134,6 @@ namespace Game.Api.Sockets
         /// </summary>
         internal async Task<SocketCommandOutcome> ExecuteServerCommand(SocketCommandInfo commandInfo)
         {
-            var now = Stopwatch.GetTimestamp();
             _logger.LogTrace("Executing server-initiated command: {CommandInfo} on socket: {Id}", commandInfo, Id);
             var (outcome, fault) = await RunCommandUnderLock(commandInfo);
             if (outcome is SocketCommandOutcome.Faulted)
--- a/Game.Application.Tests/Events/BattleStatisticsEventHandlerTests.cs
+++ b/Game.Application.Tests/Events/BattleStatisticsEventHandlerTests.cs
@@ -197,7 +197,7 @@ namespace Game.Application.Tests.Events
             // The skill fired once in the won battle, so its proficiency is represented and (being the only one)
             // takes the whole pie; below the first threshold it accrues without leveling.
             var stats = new BattleStats();
-            stats.SkillStats[skill.Id] = new SkillStats { SkillId = skill.Id, Uses = 1 };
+            stats.SkillStats[skill.Id] = new SkillStats { Uses = 1 };
             await handler.HandleAsync(VictoryEvent(loadedPlayer, loadedEnemy, stats, difficultyMultiplier: 1.0), CancellationToken);
 
             var progressRepo = scope.ServiceProvider.GetRequiredService<IPlayerProgressRepository>();
@@ -248,7 +248,7 @@ namespace Game.Application.Tests.Events
             Assert.NotNull(loadedEnemy);
 
             var stats = new BattleStats();
-            stats.SkillStats[skill.Id] = new SkillStats { SkillId = skill.Id, Uses = 1 };
+            stats.SkillStats[skill.Id] = new SkillStats { Uses = 1 };
             await MakeHandler(scope).HandleAsync(
                 VictoryEvent(loadedPlayer, loadedEnemy, stats, difficultyMultiplier: 1.0), CancellationToken);
 
@@ -323,7 +323,7 @@ namespace Game.Application.Tests.Events
             Assert.NotNull(loadedEnemy);
 
             var stats = new BattleStats();
-            stats.SkillStats[skill.Id] = new SkillStats { SkillId = skill.Id, Uses = 1 };
+            stats.SkillStats[skill.Id] = new SkillStats { Uses = 1 };
             const double multiplier = 1.5;
 
             var liveLoaded = await playerRepo.GetPlayer(livePlayer.Id);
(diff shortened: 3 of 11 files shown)


ginderjeremiah (commented):
## Review: Backend small cleanups for #1192 (dead field, enum-count coupling, dup lookup, dead local)

**Verdict:** Approve _(posted as a comment since I'm the PR author — GitHub won't accept a gating self-review)_ — all four cleanups are correct and verified against the codebase; the `ChallengeBoss` call is a sound, documented product decision. One optional test nit.

I verified each change against the actual code rather than the description:

### What's good

- **Dead `SkillStats.SkillId` removed** — confirmed both production consumers (`Game.Core/Progress/PlayerProgress.cs:118`, `Game.Application/Services/ProficiencyRewardService.cs:175`) destructure `(skillId, skillStats)` off the `Dictionary` key and never read the field. No remaining references anywhere outside the (updated) tests. Clean removal that also kills the silent key/field divergence risk.
- **`NewEquippedList` derives from `Enum.GetValues<EEquipmentSlot>()`** — verified nothing indexes `EquipmentSlots` positionally (every access is `FirstOrDefault(s => s.Value == …)` / `SelectNotNull`), and `EEquipmentSlot` is 0-based contiguous, so this is behavior-preserving today and genuinely robust against future renumbering. Good explanatory comment.
- **`Inventory.TryUnequipItem` → `int?`** — the only caller is `Player.TryUnequipItem` (which still returns `bool`, so `PlayerService.cs:29` is unaffected), and the duplicate `EquipmentSlots` scan is gone. The `is not int itemId` pattern reads well and the new XML doc documents the null-on-empty contract.
- **Dead local removed** — `Stopwatch` / `using System.Diagnostics` are still used at `SocketHandler.cs:159` and `:213`, so no orphaned import left behind.
- **`ChallengeBoss` zone asymmetry** — I traced `StartBossBattle` (`BattleService.cs:186`): it no-ops on `IsZoneRetired || !IsZoneUnlocked`, so a client-supplied `ZoneId` provably can't reach locked/retired content — only an already-unlocked zone. Keeping the parameter (re-farm a cleared boss without walking back) is consistent with the sibling's documented anti-cheat note, and the added comment accurately captures the constraint. Reasonable to keep as-is.

### Test coverage

Adequate. Signature/field changes are mirrored in the suites — `InventoryTests` covers both the `Assert.Equal(1, result)` and `Assert.Null(result)` paths, `PlayerTests` covers the `ItemUnequippedEvent` path, and the `SkillStats { Uses = 1 }` updates are consistent across `Game.Core.Tests` / `Game.Application.Tests`.

### Nits

- **(optional) Lock in the `NewEquippedList` invariant** — the refactor's whole point is "one slot per enum value, regardless of numbering," but no test asserts that directly. The existing tests only prove `AccessorySlot` exists via `.First(...)`. A tiny test (`EquipmentSlots.Count == Enum.GetValues<EEquipmentSlot>().Length` and one slot per value) would guard the new robustness guarantee against a future enum change. Non-blocking.

No frontend parity concern — confirmed the dead field has no `skillId` counterpart in the frontend battle code, as the description states.


---

PR #1207, opened 2026-06-24
Exclude retired-path proficiencies from the gateway-opening index (#1203)

### LINKED ISSUE #1203: Retired proficiency path can still be opened (and seed-skill granted) as a cross-path gateway
Discovered while implementing #1176 (PR #1202).

## Context

#1202 freezes a retired path for **XP accrual** by excluding it from the skill→contribution reverse index in `ProficienciesCacheHolder` — so no skill routes XP into a retired track, and it accrues/levels/grants nothing. That fully covers the accrual path the issue scoped (`BuildContributions` / `AccrueAndApply`).

## The gap

The reverse index is *not* the only structure that grants skills from a proficiency. `ProficiencyRewardService.OpenSuccessors` opens nodes when a tier maxes, and one branch opens **cross-path gateways** via the `DependentsByProficiency` index (also built in `ProficienciesCacheHolder`):

```
foreach (var gatedId in _proficiencies.DependentsOf(maxed.Id))
{
    if (AllPrerequisitesMaxed(gatedId, progress))
    {
        Open(gatedId, player, opened, openedIds); // grants the gated tier's SeedSkillId
    }
}
```

If a **retired** path's tier-0 is authored as a cross-path gateway whose prerequisites live in **live** paths, then maxing those live prerequisites opens the retired gateway and grants its `SeedSkillId` — even though the retired track can never be trained (its contributions are filtered out). The player ends up with a seed skill for a dead track.

This is out of the accrual scope #1176 called out, but it's the same class of issue ("granting skills from a retired track") and makes the freeze internally inconsistent.

## Proposed fix

Mirror the contributions filter: exclude proficiencies on retired paths from the `DependentsByProficiency` values built in `ProficienciesCacheHolder`, so a retired track is never opened as a gateway. Entirely in the holder, no Core change. Add an integration test (a retired gateway whose live prerequisites max must not grant its seed skill).

Open design question to confirm first: should a retired proficiency still be allowed to act as a *prerequisite* gating live gateways? If a path is retired it can never max, so any live gateway depending on it would soft-lock — but that is arguably an authoring mistake (retiring a path that gates a live one) rather than something the accrual code should paper over.

scope: small

ginderjeremiah:
## Summary

Closes #1203.

#1202 froze retired proficiency paths for **XP accrual** by excluding them from the skill→contribution reverse index in `ProficienciesCacheHolder`. But the gap #1203 calls out is that the reverse index is *not* the only structure that grants skills from a proficiency: `ProficiencyRewardService.OpenSuccessors` also opens **cross-path gateways** via the `DependentsByProficiency` prerequisite index. A retired path's tier-0 authored as a gateway whose prerequisites live in **live** paths would still be opened — and its `SeedSkillId` granted — when those live prerequisites max, even though the retired track can never be trained. That left the freeze internally inconsistent.

## Change

Mirror the existing contributions filter, entirely in the cache holder (no Core change):

- `ProficienciesCacheHolder` now derives a `retiredPathIds` set once (paths are loaded up-front) and **excludes proficiencies on retired paths from the gated side of `DependentsByProficiency`**. A retired tier is therefore never returned by `DependentsOf`, so it's never opened as a cross-path gateway and never grants its seed skill.
- The existing `contributionsBySkill` retired filter now reuses the same `retiredPathIds` set (was a separate `RetiredAt is null` predicate).
- `docs/backend.md` retirement lifecycle section now notes the gateway-side freeze alongside the accrual freeze.

The key side of the index (a retired proficiency acting as a *prerequisite*) is intentionally left alone — see the design note below.

## Design note (the issue's open question)

> Should a retired proficiency still be allowed to act as a *prerequisite* gating live gateways?

I left this as-is. A retired path can never max (its contributions are filtered out), so a live gateway depending on a retired prerequisite would simply never open — a soft-lock. But per #1202's decision that retirement is an **authoring concern**, retiring a path that gates a live one is an authoring mistake, not something the open logic should paper over. The open trigger (`OpenSuccessors`) only ever runs when a proficiency *maxes*, and a retired (frozen) proficiency never maxes, so a retired proficiency is never even used as a `maxed` key in practice. Happy to add a guard or an admin-save validation if you'd prefer to catch that authoring case explicitly — flagging for your call.

## Test plan

- [x] `Game.Application.Tests` — 556 passed (Postgres/Redis integration), including the new `MaxingThePrerequisitesOfARetiredGateway_DoesNotOpenOrSeedIt` (a retired gateway whose live prerequisites both max grants no seed skill — fails without the fix, passes with it)

No frontend changes: the proficiency open/gateway logic is computed server-side at battle completion, off the frontend/backend parity surface (`docs/backend.md` "Proficiency XP accrual"), so no parity mirror is required — same as #1202.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RMfeV62V616Vra3dsCeph6)_


ginderjeremiah (commented):
## Review: Exclude retired-path proficiencies from the gateway-opening index (#1203)

**Verdict:** Approve _(posted as a comment since I'm the PR author — GitHub won't accept a gating event on one's own PR)_ — the fix closes the last accrual-vs-open inconsistency from #1202 at the right choke point, mirrors the established contributions filter, stays entirely in the cache holder with no Core change, and ships a true regression test plus a docs update.

### Correctness

- The filter `entities.Where(p => !retiredPathIds.Contains(p.PathId))` drops retired-path proficiencies from the **gated** side of `DependentsByProficiency`, so `DependentsOf(maxed)` never returns a retired gateway and `OpenSuccessors` never opens or seeds it. Verified `DependentsOf` (`Proficiencies.cs:70`) reads only from this snapshot and its sole consumer is `ProficiencyRewardService.OpenSuccessors` — so the blast radius is exactly the cross-path gateway branch and nothing else.
- The **within-path** successor branch (`OpenSuccessors` → `NextTier`) needs no equivalent guard: it only fires when a tier of that path maxes, and a retired path accrues no XP (the `contributionsBySkill` freeze), so its tiers never max. Leaving that branch untouched is correct, not an oversight.
- Reusing the single `retiredPathIds` set for both `contributionsBySkill` and `dependentsByProficiency` (replacing the old standalone `RetiredAt is null` predicate) is a nice DRY-up, and hoisting the `paths` load above the prerequisite-graph build is behavior-preserving (`AssertZeroBasedContiguity("Paths")` still runs).

### Design — the issue's open question (retired proficiency as a *prerequisite*)

I agree with leaving the runtime open logic alone. Per #1202, retirement is an authoring concern, and `OpenSuccessors` only runs on a max — a frozen proficiency never maxes, so a retired node is never a `maxed` key. Papering over a retired-prereq-gating-a-live-gateway soft-lock in the open path would be the wrong layer.

The clean place to catch that authoring mistake is an **admin-save validation** — rejecting retiring a path whose tiers are authored as prerequisites of a live gateway — which mirrors the existing precedent already documented two bullets down in `backend.md` ("Authoring guard — no spawnable enemies in a live zone"). I'd suggest a small follow-up issue for that rather than expanding this PR's scope. Answering your flagged call: no guard needed in *this* change.

### Test coverage

- `MaxingThePrerequisitesOfARetiredGateway_DoesNotOpenOrSeedIt` is structured as the exact negative of `MaxingTheLastPrerequisite_OpensAndSeedsTheGatewayTier` — same seeding, gateway moved onto a retired path — so the positive control proves the mechanism and the new test isolates the retirement suppression. Integration level is correct here (the holder reads from `GameContext`; this matches the `RetiredPath_FreezesTheTrack` precedent). No frontend/parity mirror required — proficiency open logic is server-side at battle completion, off the parity surface.
- _Optional nit:_ the test asserts only that the seed skill isn't granted. Since `AccrueAsync` already returns `results`, you could add `Assert.Equal(1, Assert.Single(results).NewLevel)` to prove prereq B genuinely maxed — making the test self-contained (suppression is specifically the gateway open, not a dead accrual) rather than leaning on the sibling control. Take it or leave it.

### What's good

Tight, single-choke-point fix that makes the freeze internally consistent without touching the domain, a faithful regression test, and a concise docs update that keeps the retirement lifecycle section accurate. Clean work.

---

PR #1211, opened 2026-06-24
Add Aetheric conlang font for proficiency "words of power"

ginderjeremiah:
## What

Adds a self-hosted constructed-script display font, **Aetheric**, used to render proficiency identities as decorative **"words of power"**, plus the component + theming hook to use it.

Aetheric specimen

## Why

The proficiency system ([game-design.md → Proficiency](docs/game-design.md)) frames each path/tier as a mastery identity (Fire Magic → Inferno Magic). A bespoke arcane script gives those identities flavour without inventing readable copy. This PR lands the **asset + plumbing**; wiring it into a proficiency screen is left for when that UI is built.

## What's in it

- **`Aetheric` font** — `UI/static/fonts/Aetheric.woff2` (~5 KB). Generated from parametric glyph definitions in `UI/scripts/conlang/` (`generate_font.py` + `render_specimen.py` + `README.md` + a committed reference specimen), **not** hand-drawn as a binary — re-run the script to change the script.
  - Intentionally **abstract / non-Latin** (words must not read as English), **baseline-grounded** with a deliberate few descenders (`g j p y`), **varied glyph bases** (stem+bow, twin-stem bridges, stem+twigs, leaning stems, bowls) for rhythm, **in-script numerals**, and `th sh ch ng ph kh` ligatures.
- **Theming hook** — `@font-face` in `common.scss` and a `--conlang` token (falls back to `--sans` so a failed load degrades to legible Latin).
- **`WordOfPower` component** — keeps the romanization as the DOM text so it stays accessible and selectable (or announces an `aria-label` when given a readable name), restyling only the glyphs. Unit-tested (11 cases).
- **Docs** — `docs/frontend.md` records the decision: decorative flavour only, **never readable copy**.

## Design iteration

The aesthetic was chosen collaboratively: an initial angular/Futhark direction was rejected, then a flowing direction was selected and refined over several passes to ground every glyph to the baseline and strip out forms that still read as their Latin counterparts.

## Validation

- `npm run check` (svelte-check): 0 errors / 0 warnings
- `vitest` (WordOfPower): 11/11 pass
- eslint + prettier clean on touched files
- woff2 round-trips through fontTools (51 glyphs, `liga` GSUB present)

## Follow-ups (not in this PR)

- Use `WordOfPower` in the proficiency UI once that screen exists.
- Optional: true Bézier outlines instead of the overlapping-quad stroke approximation (only matters at very large display sizes).

https://claude.ai/code/session_01N6YXuczgoVKAVnC856Pj4F

---
_Generated by [Claude Code](https://claude.ai/code/session_01N6YXuczgoVKAVnC856Pj4F)_


docs/frontend.md
```diff
@@ -105,6 +105,10 @@ A few cross-cutting conventions follow from that rule:
...
 
+## Conlang script ("words of power")
+
+Proficiency identities are surfaced as **words of power** drawn in `Aetheric`, a self-hosted display font (`static/fonts/Aetheric.woff2`, `@font-face` in `common.scss`, exposed as the `--conlang` token). It is **decorative flavour, never readable copy** — the glyphs are intentionally abstract and non-Latin, so never use it for anything the player must actually read. Render through the `WordOfPower` component, which keeps the romanization as the DOM text (so it stays accessible and selectable) and only restyles the glyphs, falling back to `--sans` if the font fails to load. The font is **generated** from parametric glyph definitions in `UI/scripts/conlang` (re-run the script to change the script) rather than hand-edited as a binary.
```
ginderjeremiah, line 110: This documents the *implementation* of the decision in the frontend doc, which is right — but the "words of power" concept itself (proficiency identities surfaced as a decorative arcane script) is a **player-facing game-design decision**, and it appears nowhere in `docs/game-design.md`. I grepped the docs: `words of power` / `conlang` / `Aetheric` exist only in this PR. Per CLAUDE.md, important design decisions belong in the appropriate doc, and game-design.md is the source of truth for game concepts — the Proficiency section there frames tiers as readable identities (Fire Magic → Inferno Magic) with no mention of an alien script.

So this should land in the Proficiency section of game-design.md (even a sentence), not just here. More importantly, since the concept is net-new and not previously agreed, it's worth confirming it's actually wanted before committing the toolchain + asset to surface it. Not a code blocker — a "is this the direction" check.

---

UI/src/components/WordOfPower.svelte
```diff
@@ -0,0 +1,61 @@
...
+
+let { text, label, size, glow = false, title, class: klass = '' }: Props = $props();
+
+let fontSize = $derived(typeof size === 'number' ? `${size}px` : size);
+</script>
+
+<style lang="scss">
+.word-of-power {
+	// Decorative conlang face; colour is inherited so the theme/consumer controls it.
+	font-family: var(--conlang, 'Aetheric');
```
ginderjeremiah: Nit: `--conlang` is already defined in `+layout.svelte` as `'Aetheric', var(--sans)`, so the local `'Aetheric'` fallback here only kicks in when the token is entirely undefined (component rendered outside the layout `:root`) — and in that case it falls back to *just* `Aetheric` with no Latin `--sans` safety net, the opposite of the "degrades to legible romanization" guarantee the component comment promises. Minor since in-app the token is always defined; if you want the guarantee to hold unconditionally, `var(--conlang, var(--sans))` would be the more faithful fallback. Take it or leave it.

---

ginderjeremiah (commented):
## Review: Aetheric conlang font + `WordOfPower` component for proficiency "words of power"

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating event from me). The code is clean, isolated, accessible, and well-tested. My one real reservation is a **product/design question, not a code defect**: this lands speculative infrastructure for a concept that isn't in the design docs yet and has no consumer. Worth settling that before merge, but nothing here is broken or unsafe.

### Should-fix / confirm before merge
- **The "words of power" concept isn't in `docs/game-design.md`** (`docs/frontend.md:110`). It's a net-new, player-facing flavour decision — proficiency identities rendered in an invented arcane script — and it appears in *no* doc outside this PR. The Proficiency section currently frames tiers as readable identities (Fire Magic → Inferno Magic). If the project is adopting this, the concept belongs in game-design.md; and since it wasn't previously agreed, it's worth an explicit "yes, this is the direction" before committing the toolchain + asset.
- **Speculative / unused.** Nothing renders `WordOfPower` yet — by the PR's own admission, wiring waits for a proficiency screen that doesn't exist. The infra is well-isolated and low-risk, so this isn't a blocker, but landing a component + font + 357 lines of Python ahead of any consumer carries real bit-rot risk: when the proficiency UI is actually built it may want something different. Reasonable to hold until there's a consumer, or merge knowing it's a bet.

### Nits
- **New Python toolchain in a Svelte/C# repo.** `UI/scripts/` previously held one JS helper; this adds Python as an asset-gen dependency. It's dev-only, self-contained, and the generated-not-hand-drawn approach is consistent with the icon-art pipeline philosophy — fine — but there's no pinned `requirements.txt`; the README leans on ambient `pip install fonttools brotli pillow`. A pinned manifest would make regeneration reproducible.
- **`specimen.png` is a committed generated artifact** that can drift from the script. Intentional as install-free documentation, just flagging the staleness vector.
- `WordOfPower.svelte:52` — the local `'Aetheric'` font-family fallback is slightly weaker than the component's stated "degrades to legible Latin" guarantee when the `--conlang` token is undefined; `var(--conlang, var(--sans))` would be more faithful. Take it or leave it.

### Test coverage
- `WordOfPower` is thoroughly unit-tested (11 cases under `tests/components/`, mirroring source path per the frontend testing rule): DOM-text-as-romanization, role/aria-label presence-and-absence, label-hides-glyphs, title default/override, numeric-vs-string sizing, and glow/class passthrough. Good coverage of the component's actual logic.
- The Python generators are untested — acceptable; they're build tooling, not app domain logic, and outside the project's testing rules. No battle-logic/parity surface is touched, so no frontend/backend mirroring obligation.

### What's good
- **Accessibility is the standout.** Keeping the romanization as the real DOM text (and switching to `role="img"` + `aria-label` + `aria-hidden` glyphs only when a readable name is supplied) is exactly right — the alien script stays decorative without ever trapping content a screen reader or selection needs.
- **Theming is clean** — `--conlang` token sits with the other font tokens in `+layout.svelte`, colour inherits via `currentColor`, the glow derives from `color-mix(... currentColor ...)` with no hard-coded hex, and `font-display: swap` + the `--sans` fallback chain degrade gracefully. Matches the styling doc.
- **The asset is regenerable, not an opaque binary** — parametric glyph source + README + committed specimen means the script is the source of truth, in the spirit of the project's generated-asset approach.

---

PR #1242, opened 2026-06-25
Isolate login tracking on its own DI scope (#1230)

### LINKED ISSUE #1230: LoginTrackingMiddleware: a swallowed tracking failure can ride along on (or poison) an unrelated request's commit
## Problem

`LoginTrackingMiddleware` records connection tracking on every authenticated HTTP request as **best-effort**, swallowing any exception so it "never affects the response":

```csharp
// Game.Api/Middleware/LoginTrackingMiddleware.cs:25-44
try
{
    await trackingService.RecordConnection(...);
}
catch (Exception ex)
{
    _logger.LogWarning(ex, "Failed to record login tracking for user {UserId}.", ...);
}
await _next(context);
```

`RecordConnection` flows down to `UserLogins` (`Game.DataAccess/Repositories/UserLogins.cs`), which **shares the request-scoped `GameContext`** and owns its own commit via `SaveWithConflictRetry`. That retry loop only recovers from **unique violations**:

```csharp
// UserLogins.cs:167-182
for (var attempt = 1; ; attempt++)
{
    await buildChanges();                 // Adds Device / BrowserInfo / UserLogin to the tracker
    try { await _context.SaveChangesAsync(ct); return; }
    catch (DbUpdateException ex) when (attempt < MaxSaveAttempts && ex.IsUniqueViolation())
    { _context.ChangeTracker.Clear(); }   // only cleared on a unique violation
}
```

If `SaveChangesAsync` throws for any **other** reason (a non-unique `DbUpdateException`, a transient connection fault, etc.), `buildChanges()` has already enqueued `Added` entities and the change tracker is **not** cleared. The middleware swallows the exception, then `_next(context)` runs and the global `CommitFilter` calls `IUnitOfWork.CommitAsync` on the **same** `GameContext`. Two outcomes, both wrong:

1. The leftover tracking inserts are silently persisted as a side effect of an unrelated request's commit, or
2. The same insert fails again and turns a best-effort tracking failure into a **500 on an unrelated action**.

## Why it matters

The whole point of the swallow is isolation ("never affects the response"), but sharing the scoped context defeats it: a tracking failure can leak into or break the real request's unit of work. Low likelihood (it needs a non-unique-violation save failure), but it undermines the documented guarantee.

## Suggested fix

Make tracking truly isolated. Either:
- Clear the change tracker in the middleware's `catch` (and after a successful `RecordConnection`, since it self-commits) so nothing it touched can ride the later commit; or, more robustly,
- Run `RecordConnection`/`SaveDeviceInfo` on their **own DI scope / `GameContext`** (these operations already commit themselves, like account creation), so their failure is structurally independent of the request's unit of work.

## Verification

Confirmed by reading the middleware, `LoginTrackingService`, `UserLogins`, and `CommitFilter`. `RecordConnection` shares the request `GameContext` and only the unique-violation path clears the tracker.

ginderjeremiah:
## Summary

`LoginTrackingMiddleware` recorded connection tracking on the **request-scoped `GameContext`** and swallowed any failure to keep tracking best-effort ("never affects the response"). But sharing the request's context defeated that isolation:

- `RecordConnection` → `UserLogins.SaveWithConflictRetry` **self-commits**, and its retry loop only clears the change tracker on a **unique violation**.
- For any *other* save failure (a non-unique `DbUpdateException`, a transient connection fault), the `Added` tracking entities stay queued on the shared context.
- The middleware swallows the exception, then `_next` runs the real action and the per-action `CommitFilter` calls `CommitAsync` on the **same** context — so the leftover tracking inserts either silently persist as a side effect of an unrelated request, or fail again and turn a best-effort tracking failure into a **500 on an unrelated action**.

I verified the issue's claim end to end (middleware → `LoginTrackingService` → `UserLogins` → `CommitFilter`); it's accurate.

## How it addresses the issue

Takes the issue's **more robust** option: run the tracking work on its **own DI scope / `GameContext`** (these operations already self-commit, like account creation), so a tracking save failure is **structurally independent** of the request's unit of work. The scope is disposed regardless of outcome, discarding anything it queued — no leftover state can ride a later commit.

The middleware reads the request's user identity from the still request-scoped `SessionService`; only the tracking repository graph (`LoginTrackingService` → `IUserLogins` → `GameContext`) is resolved from the fresh scope.

### Notes
- **Doc fix:** the `IUserLogins` summary claimed "Mutations are queued on the change tracker; the caller commits the unit of work" — inaccurate, since both operations self-commit. Corrected to match reality (and the existing impl comment in `UserLogins`).
- **`DeviceInfo` controller path reviewed, left as-is:** `SaveDeviceInfo` also self-commits on the request context, but it runs inside a controller action that does **not** swallow — a non-unique failure throws out of the action, `CommitFilter` skips the commit, and the request 500s on its *own* work (correct). There's no ride-along onto an unrelated request, so it needs no change.
- **Scope:** backend-only. No battle-parity surface, no codegen, no frontend.

## Test plan

- [x] `dotnet build` clean (0 warnings); `dotnet format --verify-no-changes` clean on the changed file.
- [x] New `LoginTrackingIsolationTests` — a fault-injecting `IUserLogins` double reproduces the issue's failure mode (queue a tracking insert, then a **non-unique FK-violation** save failure) and asserts the unrelated `Status` request still returns 200 and **nothing leaks** to the DB. **Confirmed it's a genuine regression guard:** it fails (500) against the pre-fix shared-context middleware and passes against the fix.
- [x] Existing `LoginTrackingTests` (6) and `ForwardedHeadersTrustTests` (7) green; application-layer `UserLogins`/`LoginTracking` tests (4) green.

Closes #1230

---
_Generated by [Claude Code](https://claude.ai/code/session_01F4LiHEgYctZ1hEXPJXCSM3)_

### DIFF
--- a/Game.Abstractions/DataAccess/IUserLogins.cs
+++ b/Game.Abstractions/DataAccess/IUserLogins.cs
@@ -4,7 +4,8 @@ namespace Game.Abstractions.DataAccess
     /// Persists lightweight user-connection tracking: a per-(user, IP, device) <c>UserLogin</c> whose
     /// last-connection timestamp is continually refreshed, the <c>Device</c> (deduplicated by fingerprint)
     /// the login references, and the <c>BrowserInfo</c> (deduplicated by user-agent) the device reports.
-    /// Mutations are queued on the change tracker; the caller commits the unit of work.
+    /// Each operation <b>owns its own commit</b> (like account creation) because it runs outside the
+    /// per-action commit filter and must retry its build-and-save on a concurrent-insert unique violation.
     /// </summary>
     public interface IUserLogins
     {
--- a/Game.Api.Tests/Integration/LoginTrackingIsolationTests.cs
+++ b/Game.Api.Tests/Integration/LoginTrackingIsolationTests.cs
@@ -1,0 +1,131 @@
+using Game.Abstractions.DataAccess;
+using Game.Api.Http;
+using Game.Infrastructure.Database;
+using Game.Infrastructure.Entities;
+using Game.TestInfrastructure.Base;
+using Game.TestInfrastructure.Fixtures;
+using Game.TestInfrastructure.Helpers;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using System.Net;
+using Xunit;
+
+namespace Game.Api.Tests.Integration
+{
+    /// <summary>
+    /// Pins the isolation guarantee of <c>LoginTrackingMiddleware</c> (#1230): because connection tracking
+    /// self-commits and its failures are swallowed, it must run on its own <c>GameContext</c> — otherwise a
+    /// non-unique save failure leaves queued inserts on the request's shared context that the per-action
+    /// commit filter then re-flushes (a 500 on an unrelated request) or silently persists. The fault double
+    /// reproduces that failure mode; the test asserts the unrelated request is unaffected and nothing leaks.
+    /// </summary>
+    [Collection("Integration")]
+    public class LoginTrackingIsolationTests : ApiIntegrationTestBase
+    {
+        private const string UserAgent = "TestAgent/1.0 (LoginTrackingIsolationTests)";
+        private const string Fingerprint = "fp-isolation";
+
+        public LoginTrackingIsolationTests(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper)
+            : base(containers, testOutputHelper) { }
+
+        protected override GameServerFactory CreateFactory(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper)
+        {
+            return new FaultInjectingFactory(containers, testOutputHelper);
+        }
+
+        [Fact]
+        public async Task TrackingSaveFailure_DoesNotBreakTheRequest_OrPersistAnything()
+        {
+            // Arrange — a real, logged-in user whose first authenticated request triggers tracking, which the
+            // injected double makes fail with a non-unique (FK-violation) save error.
+            var userId = await SeedUserAsync("isouser", "isopass");
+            using var authClient = await LoginWithDeviceAsync("isouser", "isopass");
+
+            // Act — an unrelated, read-only request. Its commit must not inherit the failed tracking inserts.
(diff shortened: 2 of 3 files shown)


ginderjeremiah:
## Review: Approve

Verified the fix end to end against the source.

**Issue is real and accurately diagnosed.** `UserLogins.SaveWithConflictRetry` self-commits (`_context.SaveChangesAsync`) and only clears the change tracker `when (... ex.IsUniqueViolation())`. Any other `DbUpdateException`/transient fault leaves `Added` tracking entities queued on the request-shared `GameContext`, which the swallowing middleware then carries into the per-action `CommitFilter` (`IUnitOfWork.CommitAsync` on the same context) — silently persisting tracking inserts on an unrelated request, or 500-ing it.

**Scope isolation is correct.**
- `scopeFactory.CreateAsyncScope()` produces a fresh scope; `LoginTrackingService` (Scoped) → `IUserLogins`/`UserLogins` (Scoped) → `GameContext` (Scoped) are all resolved from it, so the tracking self-commit lands on a `GameContext` independent of the request's. The `CommitFilter` commits the request's unit of work, now untouched by tracking.
- `SessionService` stays request-scoped (read from the middleware DI parameter), correctly supplying `UserId`/`Authenticated`; only the tracking repo graph moves to the fresh scope. Right split.
- `await using var scope` disposes whatever the outcome (success, swallowed failure), discarding anything queued; it also releases the tracking context before `_next` runs the downstream pipeline. Good.
- This mirrors the existing `SocketInterceptorMiddleware` scope pattern, so it's consistent with established precedent.

**DeviceInfo left as-is** is the right call — it self-commits inside a controller action that doesn't swallow, so a non-unique failure 500s on its *own* work with no ride-along.

**Quality:** DDD layering preserved (middleware → application `LoginTrackingService` → `IUserLogins`); braces everywhere; nullable client-hint params handled; no null-forgiving operator. The `IUserLogins` summary doc correction matches reality.

**Regression guard:** `LoginTrackingIsolationTests` injects a fault double producing the *non-unique* (FK-violation) failure — the exact uncovered branch — and asserts the unrelated `/Login/Status` returns 200 with zero leaked `UserLogin` rows. Confirmed to fail pre-fix / pass post-fix; uses the standard integration fixtures cleanly.

CI green (backend + e2e success; frontend skipped, backend-only). No nits/should-fixes/must-fixes. Merging via squash.

---

PR #1243, opened 2026-06-25
Align frontend loadout dedup with backend Distinct (#1231)

### LINKED ISSUE #1231: Battle parity: skill-loadout dedup rules diverge between backend OrderSkillIds and frontend fillSkills
## Problem

The shared loadout-assembly rule is documented as parity-critical (it must field the same skills on both simulators), but the two implementations dedupe **different input sets**.

Backend de-duplicates the **entire** concatenation, so a duplicate *within* the selected skills is also collapsed:

```csharp
// Game.Core/Battle/BattleLoadout.cs:22
return selectedSkillIds.Concat(grantedSkillIds).Distinct();
```

Frontend only builds the `seen` set from the selected skills and dedupes the **granted** skills against it — the selected skills are mapped directly with no intra-selected dedup:

```ts
// UI/src/lib/battle/battler.ts:353-366
const skills = battlerData.selectedSkills.map((id) => new Skill(skillData[id], this)); // no dedup here
const seen = new Set(battlerData.selectedSkills);
for (const skillId of grantedSkillIds) {
    if (!seen.has(skillId)) { seen.add(skillId); skills.push(new Skill(skillData[skillId], this)); }
}
```

So if `selectedSkills` ever contained the same id twice (e.g. `[5, 5, 3]`), the **backend fields it once** and the **frontend fields it twice** — that skill fires an extra time per tick on the client, the two simulators diverge, and the backend replay would reject a legitimate victory.

## Why it matters

It is currently unreachable: `SetSelectedSkills` validates "no duplicate ids" as anti-cheat, so a clean loadout can't contain a dup. But the two sides having **different dedup rules for the same input** is a latent parity footgun — a future loadout change (or a different grant path) that lets a dup through would silently break battle parity, which is the worst class of bug here.

## Suggested fix

Make the two rules identical, or assert uniqueness of `selectedSkills` at the snapshot boundary so the invariant is enforced in one place. Cheapest: have `fillSkills` seed `seen` as it maps the selected skills and skip dups there too, mirroring `Distinct()`.

## Verification

Confirmed by reading both implementations. The padding-to-`MAX_SELECTED_SKILLS` difference is **not** a problem (the `undefined` slots are skipped and don't affect firing order) — only the dedup asymmetry is.

ginderjeremiah:
## Summary

The shared battle-loadout rule must field the same skills on both simulators (it's parity-critical — the backend replays the client's battle as anti-cheat), but the two implementations deduped **different input sets**:

- **Backend** `BattleLoadout.OrderSkillIds` runs a single `Distinct()` over the **whole** `selected.Concat(granted)` sequence — so a duplicate *within* the selected skills is also collapsed.
- **Frontend** `fillSkills` only built its `seen` set from the selected skills and deduped the **granted** skills against it; the selected skills were mapped directly with **no intra-selected dedup**.

So a `selectedSkills` containing the same id twice (e.g. `[5, 5, 3]`) would be fielded **once** by the backend and **twice** by the frontend — that skill fires an extra time per tick on the client, the simulators diverge, and the backend replay would reject a legitimate victory.

I verified the issue's claim by reading both implementations. It is currently **unreachable** — `SetSelectedSkills` anti-cheat rejects duplicate ids — but two sides having different dedup rules for the same input is a latent parity footgun: a future loadout/grant change that let a dup through would silently break battle parity, the worst class of bug here. The issue's note that the padding-to-cap difference is *not* a problem (undefined slots are skipped) is also correct — only the dedup asymmetry needed fixing.

## How it addresses the issue

Took the issue's cheapest suggested fix: rebuild `fillSkills` around a single `seen` set seeded **as it maps the selected skills** (skipping intra-selected dups), then pad to the loadout cap (preserving the fight screen's fixed slots), then append the granted skills deduped against the same set. This mirrors the backend's single `Distinct()` over the full loadout exactly. No backend code change was needed — it was already correct; the frontend was the diverging side.

## Test plan

- [x] **Parity matrix mirrored** — added a `duplicateSelectedSkill` scenario to **both** `battle-simulation-parity.test.ts` and `BattleSimulatorParityTests.cs`, same `2000ms` expected result (a skill selected twice fields once at 20/hit → dies hit 5; un-deduped 40/tick would kill at 1200ms, so the scenario fails against the pre-fix frontend and passes after).
- [x] Backend `BattleSnapshotTests.ToBattler_DuplicateSelectedSkill_AppearsOnce` unit test, alongside the existing granted-dedupe unit tests.
- [x] `dotnet build Game.sln` clean (0 warnings); `dotnet format --verify-no-changes` clean on changed files.
- [x] Backend `Game.Core.Tests` green (813); frontend `npm run lint` clean; full frontend unit suite **2497/2497** pass.

Closes #1231

https://claude.ai/code/session_014f6yrW43z15owAMGaER1ES

---
_Generated by [Claude Code](https://claude.ai/code/session_014f6yrW43z15owAMGaER1ES)_

### DIFF
--- a/Game.Core.Tests/Battle/BattleSimulatorParityTests.cs
+++ b/Game.Core.Tests/Battle/BattleSimulatorParityTests.cs
@@ -704,6 +704,21 @@ namespace Game.Core.Tests.Battle
                     ExpectedPlayerDied: false,
                     ExpectedTotalMs: 2000),
 
+                // The SAME skill selected twice is de-duplicated WITHIN the selected loadout (first wins): the
+                // player fields S1 once (22−2 = 20/hit, cooldown 400). The 100-HP enemy dies on hit 5 at tick 50
+                // → 2000ms. (Without intra-selected dedupe the two copies — 40/tick — would kill on hit 3 at
+                // 1200ms.) Pins the single Distinct() over the full loadout that the frontend mirrors.
+                ["duplicateSelectedSkill"] = new ParityScenario(
+                    Player: () => MakeBattlerWithGrants(
+                        strength: 10,
+                        selectedSkillIds: [1, 1],
+                        grants: [],
+                        skillPool: [MakeSkill(1, baseDamage: 22, cooldownMs: 400)]),
+                    Enemy: () => MakeEnemy(strength: 10, endurance: 0, skills: []),
+                    ExpectedVictory: true,
+                    ExpectedPlayerDied: false,
+                    ExpectedTotalMs: 2000),
+
                 // A granted skill carrying an EFFECT works exactly as a selected one would: the item grants a
                 // skill whose self +10 Strength buff stacks each fire, ramping its own damage. Str×1.0 raw less
                 // 2 Def deals 8,18,28,38,48,58 (Str climbing 10→60); cumulative 8,26,54,92,140,198 drops the
--- a/Game.Core.Tests/Battle/BattleSnapshotTests.cs
+++ b/Game.Core.Tests/Battle/BattleSnapshotTests.cs
@@ -319,6 +319,26 @@ namespace Game.Core.Tests.Battle
         }
 
         [Fact]
+        public void ToBattler_DuplicateSelectedSkill_AppearsOnce()
+        {
+            var snapshot = new BattleSnapshot
+            {
+                Level = 1,
+                StatAllocations = [],
+                EquippedItems = [],
+                SkillIds = [3, 3],
+            };
+
+            var battler = snapshot.ToBattler(
+                ThrowItem,
+                ThrowMod,
+                SkillResolver(MakeSkill(3)));
+
+            // The same skill selected twice is de-duplicated within the selected loadout (first wins).
+            Assert.Equal([3], battler.Skills.Select(s => s.Skill.Id));
+        }
+
+        [Fact]
         public void ToBattler_TwoItemsGrantingSameSkill_AppearsOnce()
         {
             var snapshot = new BattleSnapshot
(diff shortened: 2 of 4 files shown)


ginderjeremiah:
## Review: Approve

Verified the fix against both implementations and the battle-parity rule.

**Correctness** — The rewritten `fillSkills` now seeds a single `seen` set as it maps the selected skills (skipping intra-selected dups), pads to `MAX_SELECTED_SKILLS`, then dedupes grants against the same set. That is a first-occurrence-wins dedup over the whole `selected + granted` sequence, exactly mirroring the backend's `selectedSkillIds.Concat(grantedSkillIds).Distinct()` in `BattleLoadout.OrderSkillIds`. The asymmetry described in #1231 (selected mapped with no intra-selected dedup) is gone. No backend change was needed — it was already the correct side.

**Parity tests mirrored (battle-logic rule)** — The new `duplicateSelectedSkill` scenario is added to **both** `battle-simulation-parity.test.ts` and `BattleSimulatorParityTests.cs` with equivalent setup (Str 10, one skill at 22 dmg / 400ms cooldown selected twice, enemy Str 10 / End 0) and the identical `2000ms` expected result. Math confirmed: 22−2 = 20/hit, 5 hits clear 100 HP, 5th lands at tick 50 → 2000ms; un-deduped 40/tick would kill at 1200ms, so the scenario genuinely fails against the pre-fix frontend. Backend also gains the `ToBattler_DuplicateSelectedSkill_AppearsOnce` unit test alongside the existing granted-dedupe cases.

**Quality** — DRY (one dedup pass instead of two paths), all blocks braced, no null-forgiving operators, doc comment updated and concise. The choice to make both rules identical (rather than the issue's alternate "assert uniqueness at the snapshot boundary") is the right call — it keeps the shared `OrderSkillIds` the single source of truth.

CI green across all checks (backend, frontend, e2e on all three browsers, `all-checks-passed`). No nits or follow-ups. Merging via squash.

---

PR #1244, opened 2026-06-25
Guard zero-cooldown skill display against NaN/Infinity (#1232)

### LINKED ISSUE #1232: Skill cooldown display produces NaN/Infinity for a zero-cooldown skill
## Problem

The fight-screen skill display divides by the cooldown without guarding against a zero cooldown:

```ts
// UI/src/routes/game/screens/fight/SkillTooltip.svelte:92-95
const adjustedCd = $derived((skill?.cooldownMs ?? 0) / 1000 / cdMultiplier);
...
const cooldownProgress = $derived(isReady ? 100 : Math.max(0, ((adjustedCd - remainingCd) / adjustedCd) * 100));
```

If `cooldownMs === 0`, `adjustedCd` is `0` and `cooldownProgress` becomes `(0 / 0) * 100 → NaN`, which renders as an invalid `width: NaN%`. The same shape exists in `Skills.svelte` (`fight/Skills.svelte:109-112`, DPS / charge fraction) where a zero cooldown yields `Infinity` DPS and a permanently-not-ready charge state.

The sibling pure-metrics code already guards a zero denominator — but only for the multiplier, not the base cooldown:

```ts
// UI/src/routes/game/screens/skills/skills-view.svelte.ts:207
cooldown: cdMultiplier > 0 ? skill.cooldownMs / 1000 / cdMultiplier : skill.cooldownMs / 1000,
```

The inconsistency is the signal that the zero case isn't being handled uniformly.

## Why it matters

Low likelihood — it requires an authored skill with `cooldownMs === 0`. But authoring such a skill (e.g. a passive-ish or instant skill) would render a broken cooldown bar / NaN width and a never-ready charge indicator, and nothing currently prevents authoring a `0` cooldown. Cheap to harden.

## Suggested fix

Guard the denominators in `SkillTooltip.svelte` and `Skills.svelte` (treat a zero/near-zero `adjustedCd` as "always ready", `cooldownProgress = 100`). If a zero cooldown is never intended, also reject it in the admin skill editor / backend authoring validation.

## Verification

Confirmed by reading the three files. The divide is unguarded in the display path; only `cdMultiplier` is guarded in the metrics path.

ginderjeremiah:
## Summary

A skill authored with `cooldownMs === 0` fires every tick — the battle's `chargeTime >= cooldownMs` is always true (`battler.ts:129`), so it's "always ready". But two fight-screen display surfaces divided by the cooldown without guarding it:

- **`SkillTooltip` DPS** — `dps={total / adjustedCd}` yielded **`Infinity`** for a zero adjusted cooldown.
- **`Skills.svelte` slot charge** — `fraction = renderChargeTime / cooldownMs` yielded **`0 / 0 = NaN`**, rendering a broken `CooldownOverlay` (`sweep: NaN deg`) and a permanently-not-ready slot (`NaN >= 0.999` is false).

The loadout screen (`skills-view.svelte.ts`) already handled this gracefully via its `damagePerSecond` / `cdMultiplier > 0` guards; the fight screen was the diverging side.

## Verifying the issue's claims

I confirmed the two defects above, but the issue's third claim — that `SkillTooltip`'s **`cooldownProgress`** produces `NaN` — is **not** reachable in the current code. A zero `cooldownMs` makes `remainingCd === 0`, so `isReady` is true and the `isReady ? 100 : …` branch short-circuits **before** the `(… / adjustedCd)` divide ever runs. The genuine defects were the DPS metric and the slot charge fraction. The fix still routes `cooldownProgress` through the guarded helper, so the value stays correct regardless.

## How it addresses the issue

- New pure, unit-tested module `fight/skill-cooldown.ts`:
  - `chargeProjection(renderChargeTime, cooldownMs)` — slot overlay sweep/ready; a non-positive cooldown is fully charged (always ready).
  - `cooldownReadout(cooldownMs, renderChargeTime, cdMultiplier)` — tooltip cooldown/remaining/ready/progress; a non-positive adjusted cooldown reads ready at full fill, and it mirrors the loadout screen's `cdMultiplier > 0` fallback so a non-positive recovery multiplier can't reintroduce a divide-by-zero.
- Promoted `damagePerSecond` from `skills-view.svelte.ts` to `$lib/common` (it's a pure display helper) and reused it for the tooltip's DPS, so both surfaces share the single zero-cooldown guard instead of duplicating it (DRY) — and it removes a cross-screen import smell.
- `Skills.svelte` and `SkillTooltip.svelte` now consume the shared helpers; the now-unused inline `chargeState` and `Skill` import were removed.

## Scope note

I did **not** add backend/admin authoring validation to reject a `0` cooldown. The battle engine handles a zero cooldown deterministically (fires every tick), so it's an extreme-but-legitimate authored value, not a defect to forbid — forbidding it is a balance/authoring decision worth its own discussion, not part of a display-hardening fix. Happy to add it as a follow-up if a minimum cooldown is desired.

## Test plan

- [x] New `skill-cooldown.test.ts` — covers normal, full-charge, and the **zero-cooldown** edge for both helpers, plus the non-positive `cdMultiplier` fallback.
- [x] `damagePerSecond` zero-cooldown assertion preserved (moved import to `$lib/common`).
- [x] `npm run lint` (ESLint + svelte-check) clean — 0 errors, 0 warnings.
- [x] Full unit suite **2503/2503** pass; coverage thresholds satisfied.

Closes #1232

https://claude.ai/code/session_01V8FsxioKtasQmq8sdmkCH9

---
_Generated by [Claude Code](https://claude.ai/code/session_01V8FsxioKtasQmq8sdmkCH9)_

### DIFF
--- a/UI/src/lib/common/functions.ts
+++ b/UI/src/lib/common/functions.ts
@@ -3,6 +3,12 @@ export function formatNum(num: number): string {
 	return '' + parseFloat(num.toFixed(2));
 }
 
+/** Damage per second for a damage value over a cooldown in seconds. A non-positive cooldown yields 0
+ *  (rather than Infinity/NaN), so a zero-cooldown skill reads as 0 DPS consistently across surfaces. */
+export function damagePerSecond(damage: number, cooldown: number): number {
+	return cooldown > 0 ? damage / cooldown : 0;
+}
+
 export async function delay(delay: number) {
 	return new Promise<void>((res) => {
 		setTimeout(res, delay);
--- a/UI/src/routes/game/screens/fight/SkillTooltip.svelte
+++ b/UI/src/routes/game/screens/fight/SkillTooltip.svelte
@@ -18,7 +18,7 @@
 	</TooltipSection>
 
 	<TooltipSection label="Tempo" last={effectLines.length === 0}>
-		<TempoMetrics cooldown={adjustedCd} dps={total / adjustedCd} />
+		<TempoMetrics cooldown={adjustedCd} dps={damagePerSecond(total, adjustedCd)} />
 	</TooltipSection>
 
 	{#if effectLines.length > 0}
@@ -31,7 +31,14 @@
 <script lang="ts">
 import { EAttribute } from '$lib/api';
 import { applyDefense, expectedCritMultiplier, scaledEffectAmount, skillContributions, type Skill } from '$lib/battle';
-import { attributeIsHarmful, attributeName, describeEffect, formatAttributeValue, rarityColor } from '$lib/common';
+import {
+	attributeIsHarmful,
+	attributeName,
+	damagePerSecond,
+	describeEffect,
+	formatAttributeValue,
+	rarityColor
+} from '$lib/common';
 import { battleEngine } from '$lib/engine';
 import { staticData } from '$stores';
 import TooltipShell from '$components/tooltip/TooltipShell.svelte';
@@ -41,6 +48,7 @@ import CooldownPill from './skill-tooltip/CooldownPill.svelte';
 import DamageBreakdown from './skill-tooltip/DamageBreakdown.svelte';
 import EffectLines from './skill-tooltip/EffectLines.svelte';
 import TempoMetrics from './skill-tooltip/TempoMetrics.svelte';
+import { cooldownReadout } from './skill-cooldown';
 
 export const getBaseNode = () => container;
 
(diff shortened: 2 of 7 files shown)


ginderjeremiah:
## Review: Approve

Clean, well-scoped fix for the zero-cooldown display defect (#1232). Verified the diff against the codebase and the project's standards.

**Correctness**
- Confirmed both genuine defects: `SkillTooltip` DPS (`total / adjustedCd` → `Infinity`) and `Skills.svelte` slot charge (`renderChargeTime / cooldownMs` → `0/0 = NaN`, driving a broken `CooldownOverlay` sweep and a never-ready slot).
- Good critical verification of the issue's third claim: `cooldownProgress` was indeed **not** reachable as NaN — a zero `cooldownMs` makes `remainingCd === 0`, so `isReady` short-circuits the `isReady ? 100 : …` branch before the divide. Routing it through the guarded helper anyway is the right call.
- The new helpers preserve prior behavior exactly (including the `renderChargeTime / cdMultiplier` term), so no regression on the normal path.

**Standards**
- DRY: promoting `damagePerSecond` to `$lib/common` removes the cross-screen import smell and gives the tooltip and loadout screen a single zero-cooldown guard. Extracting the fight-screen math into a pure, DOM-free `skill-cooldown.ts` matches the "keep logic testable" guidance.
- Battle parity: this touches only display surfaces, not the `battleStep` / `battle-formulas` arithmetic, so no frontend/backend mirroring is required — correctly left untouched.
- Tests cover normal, full-charge, zero-cooldown, and the non-positive `cdMultiplier` fallback for both helpers; the moved `damagePerSecond` assertion is preserved. No missing braces; comments are appropriately detailed, not excessive.

**Scope call**
Agree with declining to add backend/admin validation rejecting a `0` cooldown — the engine handles it deterministically (fires every tick), so it's an authored-value/balance decision, not part of a display-hardening fix.

CI is green (frontend unit + e2e across all browsers; backend correctly skipped — no backend changes). No nits, should-fixes, or must-fixes. Merging via squash.

---

PR #1245, opened 2026-06-25
Harden ItemEquippedHandler retry against re-violating the slot unique index (#1233)

### LINKED ISSUE #1233: ItemEquippedHandler retry can re-violate the slot unique index (does not actually mirror ModAppliedHandler)
## Problem

`ItemEquippedHandler`'s comment claims it "Mirrors `ModAppliedHandler`", but the two retry strategies differ in a way that matters:

```csharp
// Game.DataAccess/PlayerUpdates/Handlers/ItemEquippedHandler.cs:17-25
try { await ApplyAsync(evt); }
catch (DbUpdateException ex) when (ex.IsUniqueViolation())
{
    context.ChangeTracker.Clear();
    await ApplyAsync(evt);   // re-runs the SAME insert-capable path
}
```

`ApplyAsync` does a vacate (`ExecuteUpdate`) **then** a `FirstOrDefault` + `Add`/update + `SaveChanges`. On the retry it re-runs that insert-capable path, which can violate the `(PlayerId, EquipmentSlotId)` unique index again if the slot was re-taken between the retry's vacate and its save. A second violation then propagates to the queue retry policy and ultimately **dead-letters a legitimate equip**.

Contrast `ModAppliedHandler`, which was deliberately hardened so its retry **cannot** re-violate — it switches to a server-side last-writer-wins `ExecuteUpdate`:

```csharp
// Game.DataAccess/PlayerUpdates/Handlers/ModAppliedHandler.cs:37-39
await context.AppliedMods.Where(...)
    .ExecuteUpdateAsync(s => s.SetProperty(am => am.ItemModId, evt.ItemModId)); // can't re-violate
```

## Why it matters

Low likelihood in practice — the per-socket command lock and single-active-connection serialize one player's events, and the write-behind queue is drained sequentially, so concurrent writers to the same slot are rare (cross-instance edge only). But the comment overstates the guarantee, and the asymmetry is real: under contention this handler can dead-letter an equip the mod handler would converge.

## Suggested fix

Either harden the retry to a non-re-violating `ExecuteUpdate` that reassigns the item's slot (mirroring `ModAppliedHandler` as the comment claims), or correct the comment to state the retry can still re-violate and rely on the queue's dead-letter path.

## Verification

Confirmed by reading both handlers.

ginderjeremiah:
## Summary

`ItemEquippedHandler` retried a `(PlayerId, EquipmentSlotId)` unique violation by clearing the change tracker and re-running the **insert-capable** `ApplyAsync` exactly once. Because `ApplyAsync` does a server-side vacate (`ExecuteUpdate`) **then** a `FirstOrDefault` + `Add`/update + `SaveChanges`, the retry could re-trip the same unique index if a writer re-occupied the destination slot between the retry's vacate and its save — ultimately dead-lettering a legitimate equip under sustained contention.

## Verifying the issue's claims

I confirmed the issue's core claim by reading both handlers: the retry **can** re-violate, and the `// (Mirrors ModAppliedHandler.)` comment overstated the guarantee.

But the issue's **first suggested fix rests on a false premise.** It proposes "a non-re-violating `ExecuteUpdate` that reassigns the item's slot (mirroring `ModAppliedHandler`)". `ModAppliedHandler`'s retry can't re-violate only because its key (`PlayerId, ItemId, ItemModSlotId`) **includes the slot** — the conflicting row *is* the row it owns, so a single-row `ExecuteUpdate` settling a non-key column (`ItemModId`) converges. An **equip is structurally different**: it must **evict a *different* item** from the destination slot, so vacate-then-place is inherently *two* writes with an unavoidable race window. No single `ExecuteUpdate` can mirror `ModAppliedHandler` here; truly closing the window would require serializing per-player slot writes (e.g. a Postgres advisory lock), which has no precedent in the codebase and is out of scope for a low-likelihood cross-instance edge that the per-player serialization + queue backstop already make rare.

## How it addresses the issue

Took the robust, in-scope path that mirrors a **real** codebase precedent — `UserLogins.SaveWithConflictRetry`:

- **Bounded retry loop** (`MaxSaveAttempts = 3`) replacing the one-shot retry. Each pass re-vacates whoever won the race and re-places this item, so transient cross-instance re-occupation converges **in-handler** instead of immediately burning the queue's coarser per-event attempt budget — directly reducing the "dead-letter a legitimate equip" risk the issue raises. A still-failing final attempt propagates to the queue retry/dead-letter backstop (at-least-once + idempotent handlers, so nothing is lost) rather than looping.
- **Corrected the misleading comment** — dropped the false `ModAppliedHandler` retry-parity claim and documented the real (weaker) semantics: the retry here genuinely *can* re-violate, why it converges anyway, and the queue backstop. (The `ApplyAsync`-internal "mirrors `ModAppliedHandler`'s clear-then-write" note stays — the *forward* vacate-then-write path genuinely does mirror it; only the *retry* claim was wrong.)

## Test plan

- [x] New integration test `ItemEquipped_DifferentItemsRacingIntoSameSlot_ConvergesToASingleOccupant` — races two **different** items into one slot from independent scopes (the cross-item eviction contention this handler's retry could re-violate, previously untested — the existing concurrent test races the *same* item / insert idempotency). Asserts the slot never ends doubly occupied and converges to a single occupant. A throw past the in-handler bound is swallowed (the queue's redelivery cue) and the at-least-once redelivery settles the final state deterministically.
- [x] `dotnet build Game.DataAccess` clean (0 warnings); `dotnet format --verify-no-changes` clean on changed files.
- [x] `PlayerUpdateHandlerIdempotencyIntegrationTests` — **34/34** pass.

## Scope note

I deliberately did **not** add a per-player advisory lock to fully close the race window. Given the existing per-player serialization (per-socket command lock, single-active-connection, sequential queue drain) the residual re-violation is a rare cross-instance edge, and the bounded loop + queue dead-letter backstop handle it without adding a hot-path lock and a new concurrency primitive to the codebase. Happy to file a follow-up if a hard cross-instance serialization guarantee is wanted.

Closes #1233

---
_Generated by [Claude Code](https://claude.ai/code/session_01GZieJV1D1MmpAGQ7uMY3Ax)_

### DIFF
--- a/Game.Application.Tests/DataAccess/PlayerUpdateHandlerIdempotencyIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/PlayerUpdateHandlerIdempotencyIntegrationTests.cs
@@ -441,6 +441,45 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task ItemEquipped_DifferentItemsRacingIntoSameSlot_ConvergesToASingleOccupant()
+        {
+            // Two *different* items race into the same slot from independent scopes — the cross-instance case the
+            // (player, slot) unique index guards, and the one ItemEquippedHandler's retry can re-violate (an equip
+            // evicts a different item, so vacate-then-place is two writes with a race window, unlike
+            // ModAppliedHandler's single-row settle). The bounded in-handler retry converges transient
+            // re-occupation; whatever still loses the race past that bound surfaces to the queue, which redelivers.
+            // Regardless of interleaving the slot must never end doubly occupied, and the at-least-once redelivery
+            // (applied in order below) deterministically settles the later item in with the earlier unequipped.
+            var playerId = await SeedPlayerAsync();
+            var firstItemId = await SeedItemAsync();
+            var secondItemId = await SeedItemAsync();
+            using (var scope = CreateScope())
+            {
+                var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+                await TestDataSeeder.LinkItemToPlayerAsync(context, playerId, firstItemId);
+                await TestDataSeeder.LinkItemToPlayerAsync(context, playerId, secondItemId);
+            }
+
+            var slot = (int)EEquipmentSlot.HelmSlot;
+            await ApplyRacingThenRedeliverAsync(
+                new ItemEquippedEvent(playerId, firstItemId, slot),
+                new ItemEquippedEvent(playerId, secondItemId, slot));
+
+            using var verifyScope = CreateScope();
+            var verifyContext = verifyScope.ServiceProvider.GetRequiredService<GameContext>();
+            // The unique index guarantees at most one occupant; the redelivery applied both, so exactly one item
+            // holds the slot — the later one wins — and the other is unequipped. Never two occupants, never a lost row.
+            var occupant = Assert.Single(await verifyContext.UnlockedItems.AsNoTracking()
+                .Where(ui => ui.PlayerId == playerId && ui.EquipmentSlotId == slot)
+                .ToListAsync(CancellationToken));
+            Assert.Equal(secondItemId, occupant.ItemId);
+            var displaced = Assert.Single(await verifyContext.UnlockedItems.AsNoTracking()
+                .Where(ui => ui.PlayerId == playerId && ui.ItemId == firstItemId)
+                .ToListAsync(CancellationToken));
+            Assert.Null(displaced.EquipmentSlotId);
+        }
+
+        [Fact]
         public async Task ItemUnequipped_AppliedTwice_ClearsSlotAndConverges()
         {
             var playerId = await SeedPlayerAsync();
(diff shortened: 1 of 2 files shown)


ginderjeremiah:
## Review: Approve

Verified the diff against the actual handlers and the cited precedent. The PR's analysis is correct on all counts.

**Verification**
- Confirmed `ItemEquippedHandler.ApplyAsync` is vacate (`ExecuteUpdate`) then `FirstOrDefault` + add/update + `SaveChanges` — two non-atomic writes, so the old one-shot retry genuinely *could* re-trip the `(PlayerId, EquipmentSlotId)` index. The old `(Mirrors ModAppliedHandler.)` retry claim was indeed overstated.
- Confirmed the issue's first suggested fix rests on a false premise: `ModAppliedHandler`'s key (`PlayerId, ItemId, ItemModSlotId`) *includes the slot*, so its conflict row is the row it owns and a single-row `ExecuteUpdate` converges. An equip must evict a *different* item, so it is structurally vacate-then-place — no single `ExecuteUpdate` can mirror it. Taking the comment-correction + bounded-retry path (the issue's stated alternative) is the right call.
- The new loop mirrors the real `UserLogins.SaveWithConflictRetry` precedent exactly: same `for (attempt = 1; ; attempt++)` shape, same `MaxSaveAttempts = 3`, same `attempt < MaxSaveAttempts && ex.IsUniqueViolation()` guard. On the final attempt the violation correctly propagates to the queue's retry/dead-letter backstop rather than looping.
- Corrected comment is accurate, and the `ApplyAsync`-internal forward-path "mirrors clear-then-write" note was correctly preserved.

**Code quality**
- Braces present everywhere; no null-forgiving operator; clean and DRY.
- Backend-only, no battle logic touched — battle-parity mirroring N/A.

**Tests**
- New integration test `ItemEquipped_DifferentItemsRacingIntoSameSlot_ConvergesToASingleOccupant` covers the previously-untested cross-item eviction race (the existing concurrent test only raced the *same* item / insert idempotency). The `ApplyRacingThenRedeliverAsync` helper models the queue redelivery backstop appropriately and asserts the slot never ends doubly occupied.

**Nit (non-blocking, no follow-up filed):** the bounded-retry loop now exists in two places (`UserLogins.SaveWithConflictRetry` and here). Extraction into a shared helper is plausible future cleanup, but the two live in different concerns (repository vs. handler) and the loop is small, so a shared abstraction would be speculative today. Not worth blocking or tracking.

CI is green. Merging.

---

Issue #1226, opened 2026-06-25
Class system H: author the V1 class roster (content)


ginderjeremiah:
**Depends on:** #1220, #1221, #1222, #1223, #1224, #1225, #1127

Part of the class-system spike #1126 — **area H**. See `docs/spikes/1126-class-system.md`.

Author the V1 class roster (content/design — follows the tooling).

## Scope
- Design and author three V1 classes (strawman): **Warrior** (Sword/Axe, STR/END, a martial path), **Mage** (Staff, INT, an elemental path), **Ranger** (Bow, AGI/DEX, a marksman path).
- For each: starter kit (weapon + secondary skill + shared punch), starting `AttributeDistributions` (base spread + growth vector), signature passive, and word-of-power.
- Tune the auto/free split and passive magnitudes (strawman → balanced).

## Notes
- Depends on the class system (areas A–F) **and** on proficiency content (#1127) so the kit's skills have real paths to feed.

## Tests
- Content validation; gives areas D/E real data to exercise their parity vectors against.

Issue #1225, opened 2026-06-25
Class system F: client class delivery + picker on create-character screen


ginderjeremiah:
**Depends on:** #1220, #1221

Part of the class-system spike #1126 — **area F**. See `docs/spikes/1126-class-system.md`.

Deliver class reference data to the client and let players pick a class when creating a character.

## Scope
- `GetClasses` reference socket command (version-keyed) → `staticData` slot + local cache; codegen contract; reference-data registration.
- Class **picker** on the create-character screen (`UI/src/routes/select/CreateCharacter.svelte`, from #922): preview each class's kit (skills, weapon, stat fingerprint, signature passive, word-of-power); wire the chosen `classId` into the create call.
- Render the class conlang label via the existing `WordOfPower` component (flat decorative label — no decipher).

## Docs
- `docs/frontend-screens.md`: document the class picker on the create-character screen.

## Tests
- View-model/store unit tests for class delivery + selection; create-character screen tests for the picker and that `classId` is sent.

ginderjeremiah:
Heads-up from #1254 (area D): the **live player battler's class locked base is already wired** there, so this issue does **not** need to re-do it. #1254 delivers the player's own class attribute distributions on the `PlayerData` payload (`LockedBaseDistribution`) and `PlayerManager` composes them into the battler — the live attributes already match the backend snapshot.

What remains for this issue is unchanged in spirit:
- the `GetClasses` reference socket command → `staticData` slot + local cache + reference-data registration (for the picker/codex/full class info — distinct from the per-player fingerprint #1254 delivers);
- the class **picker** on the create-character screen;
- and (from area D's leftover) the attributes screen rendering **locked-base vs free-pool distinctly** — it needs the player's `lockedBaseDistribution` (already on `PlayerData`) plus the class reference data this issue delivers.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1223, opened 2026-06-25
Class system D: Hybrid attribute model (locked base + free pool)


ginderjeremiah:
**Depends on:** #1221

Part of the class-system spike #1126 — **area D**. See `docs/spikes/1126-class-system.md`.

The hybrid attribute model — **locked base + free pool** — which also *is* the attribute floor.

## Scope
- Assemble the class `AttributeDistributions` into the battler as a level-scaled, **non-reallocatable** locked base (`BaseAmount + AmountPerLevel × level`), on **both** frontend and backend battler assembly. Source-tag it (`Class` / reuse `AttributeDistribution`) for the breakdown screen.
- Split the per-level stat-point grant so only the reduced **free pool** (strawman ~30%) flows through `PlayerStatPoints.TryUpdateAttributes`. Persist only the free-pool allocations; the locked base is recomputed from `(class, level)`, never stored.
- **Offline simulator (#879):** re-derive the locked base as levels cross the away window (power is a deterministic function of level — no stationarity violation, since the player still cannot *reallocate* while away).
- Frontend attributes screen: render locked-base vs free-pool distinctly; the free-pool budget drives the allocation UI.

## Notes
- This delivers the floor structurally (you cannot strip the locked base), so no separate floor-validation rule is needed.

## Tests
- Unit tests for the locked-base derivation and the free-pool budget math; **frontend↔backend parity vectors** for a leveled locked base; offline-sim test that the locked base grows correctly across simulated level-ups.

Issue #1220, opened 2026-06-25
Class system A: Class reference-data model + admin authoring


ginderjeremiah:
Part of the class-system spike #1126 — **area A**. See `docs/spikes/1126-class-system.md`.

Foundation: the authored `Class` catalogue and its admin authoring, following the established reference-data pattern (skills/items/proficiencies/paths).

## Scope
- `Class` EF entity + DbContext config + migration. Fields (strawman): `Id`, `Name`, `Description`, `RetiredAt`, conlang `Word` (decorative label), `StarterSkillIds[]`, `StarterEquipment[]` (`{ ItemId, EEquipmentSlot }`), `AttributeDistributions[]` (`{ Attribute, BaseAmount, AmountPerLevel }` — the locked base, same shape as enemy distributions), `SignaturePassive` (`{ Attribute, Amount, ScalingAttribute?, ScalingAmount, ModifierType }`).
- Cache holder + `Volatile` snapshot + `VersionKey`; entity↔core↔contract mappers; reader repository + DI registration.
- Admin Workbench **class editor** in the **Progression** nav group (Identity / Starter kit / Attributes / Passive sections), child savers per the `persistEntity` pattern; add/edit endpoints.
- Content lifecycle: retire (not delete) per the reference-data convention; authoring pickers exclude retired records from new references.

## Notes
- Authored content is created at runtime through the Workbench, not seeded in migrations.
- No hard external dependency; real content (area H) needs proficiency content (#1127), but the model + tooling stand alone.

## Tests
- Mapper round-trip (entity↔core↔contract) unit tests; admin add/edit integration tests; cache snapshot/version tests. Mirror the proficiency/path equivalents.

PR #1246, opened 2026-06-25
Add Player.ClassId seam + class-driven character creation (#1221)

### LINKED ISSUE #1221: Class system B: Player.ClassId seam + class-driven character creation
**Depends on:** #1220

Part of the class-system spike #1126 — **area B**. See `docs/spikes/1126-class-system.md`.

Make class the parameter that drives character creation, and add the persistent `Player.ClassId` seam.

## Scope
- Add non-nullable `Player.ClassId` (+ migration).
- Rework `NewPlayerFactory.Create` to take a resolved `Class`: produce the class's starter skills (selected) and starting attribute allocation from class data, retiring the hardcoded `StarterSkillCount` / `StartingAttributeAmount` constants.
- Thread a validated `classId` through `AccountService.CreatePlayer` and `CreateAccount` (reject unknown/retired class).
- **Retire `StartsUnlocked`** on `Proficiency` and `AccountService.RootSeedSkillIds()` — a class's roots now emerge from its kit via derived openness (first won battle). Update the proficiency model/contract/mapper/admin editor + tests accordingly.
- **Existing-character reset**: delete all existing characters; every new character requires a class (no grandfathering — pre-release).

## Notes
- The shared path-less "punch" fallback skill is content listed in every class's `StarterSkillIds`; verify it earns no proficiency XP (no path contribution).

## Docs
- Add a **Classes** section to `docs/game-design.md`; refresh #982's stale `StartsUnlocked` mentions; correct #922's "not yet implemented" status note.

## Tests
- `NewPlayerFactory` unit tests for class-driven kit + allocation; creation orchestration integration tests (valid/invalid class); regression that removing `StartsUnlocked` still opens kit paths on first contribution.

ginderjeremiah:
Implements **area B** of the class-system spike (#1126): class becomes the parameter that drives character creation, the persistent `Player.ClassId` seam is added, and the `StartsUnlocked` stopgap is retired. Depends on #1220 (area A — `Class` reference data), which is merged.

Closes #1221

## What changed

**`Player.ClassId` seam**
- Added non-nullable `ClassId` to the domain `Player`, the EF entity, the `NewPlayer` blueprint, and the lean cache model — threaded through `PlayerMapper`, `PlayerCacheMapper`, and the load projection so a loaded player carries its class (the readable seam areas D/E build on). Modelled as a plain `int` column, consistent with `CurrentZoneId` and the zero-based reference-cache convention (no EF FK navigation).

**Class-driven creation**
- `NewPlayerFactory.Create` now takes a resolved `Game.Core.Classes.Class`: starter skills come from the kit (all selected, deduped by id so the shared "punch" can appear in every kit), and the starting attribute allocation is sourced from the class's `AttributeDistributions`. Retired `StarterSkillCount` / `StartingAttributeAmount`.
- `AccountService.CreateAccount` and `CreatePlayer` take a `classId`, resolve it via `IClasses` and **reject an unknown or retired class** (new `InvalidClass` outcome). Added a dedicated `CreateAccountRequest` (kept separate from the shared `LoginCredentials`) and `classId` on `CreatePlayerRequest`; the TS client was regenerated.

**`StartsUnlocked` retirement**
- Removed the flag from the proficiency entity / contract / core model / mapper / admin editor and removed `AccountService.RootSeedSkillIds()`. A class's starting roots now **emerge from its kit** via the existing derived-openness rule (a kit skill opens its path the first time it accrues XP). Openness never read `StartsUnlocked` (only the creation-time root seeding did), so this is behaviour-neutral outside creation. The migration drops the column.

**Existing-character reset**
- The migration truncates `Players` (CASCADE) before adding the required `ClassId` — no grandfathering, per the pre-release decision. Accounts (`Users`) are kept; they simply have no characters until one is created with a class.

**Docs**
- New **Classes** section in `game-design.md`; a class-driven-creation + `ClassId`-seam note in `backend.md`; refreshed the stale `StartsUnlocked` openness wording in `game-design.md` / `backend.md` and the #982 spike's strawman model.

## Two decisions worth a look

1. **B/D attribute boundary.** The starting allocation is sourced from each core attribute's `BaseAmount` in the class distribution (level-1 base), seeding a row for every core attribute (0 where the class doesn't invest) so `PlayerStatPoints` can still allocate into it. The full hybrid model — the level-scaled, non-reallocatable *locked base* (not stored) plus the reduced *free pool* — is explicitly **area D (#1223)**; this PR only retires the hardcoded constant and makes the base spread class-driven.
2. **Transitional creation state.** With the class **picker (#1225)** and **authored content (#1226)** still to come, creation is wired end-to-end but not yet usable from the client. The frontend call sites send a placeholder `classId: 0` with a `TODO(#1225)`, kept type-correct so everything compiles and tests pass. This is the intended consequence of the spike's B→F→H ordering.

## Test plan

- **Backend:** `dotnet build` + full suite green — Core 819, Application 580, Api 614, Infrastructure 60. `dotnet format --verify-no-changes` clean.
  - `NewPlayerFactory` unit tests rewritten for the class-driven kit + base spread.
  - Creation orchestration integration tests for valid **and** invalid/retired class (`AccountService` + `LoginController`).
  - Explicit regression: a fresh player's kit skill opens its path on first contribution **without** `StartsUnlocked` (`ProficiencyRewardServiceTests`).
  - `ClassId` round-trips through the player cache mapper.
- **Frontend:** `npm run lint` clean (ESLint + svelte-check), `npm run test:unit:ci` green (2504 tests); codegen drift check satisfied (regenerated `auth.ts` / `api-type-map.ts` / `contracts.ts`).

https://claude.ai/code/session_01MESVtwPCYD4CrNFDvn96jN

---
_Generated by [Claude Code](https://claude.ai/code/session_01MESVtwPCYD4CrNFDvn96jN)_


Game.Core/Players/NewPlayerFactory.cs
```diff
@@ -1,69 +1,64 @@
...
-                {
-                    SkillId = id,
-                    Selected = false,
-                    Order = StarterSkillCount + index,
-                });
+
+            // The class's base attribute spread, keyed by attribute. This is the level-1 base only; the
+            // level-scaled, non-reallocatable locked base and the reduced free pool are introduced in #1223.
+            var baseByAttribute = @class.AttributeDistributions
+                .ToDictionary(distribution => distribution.AttributeId, distribution => (double)distribution.BaseAmount);
```
ginderjeremiah, line 38: Nit / follow-up (non-blocking): this factory trusts `class.AttributeDistributions` to be well-formed in two ways that aren't guarded:

1. **Duplicate-attribute distribution → `ToDictionary` throws.** If a class authors two rows for the same `AttributeId`, this throws `ArgumentException` at character creation, surfacing as a 500.
2. **Non-core distribution → silently dropped.** Only core attributes are read downstream (`Where(Attribute.IsCore)` + `TryGetValue`), so a distribution authored against a derived attribute (e.g. MaxHealth) is ignored without warning.

Both are really *class-authoring* validation concerns, not factory bugs — the factory's fail-loud-on-duplicate and core-only seeding are both defensible. But there's currently no admin-side guard that a class's `AttributeDistributions` are core-attribute-only and unique per attribute (the same anti-tamper shape the skill/enemy authoring guards use). Worth a follow-up issue (tech-debt / scope: small) so a malformed class is rejected at save time rather than at creation. Not a blocker for this PR.

---

ginderjeremiah (commented):
## Review: Add `Player.ClassId` seam + class-driven character creation (#1221, spike #1126 area B)

**Verdict:** Approve _(submitted as a comment — I authored this PR, so GitHub won't take a gating review from me; the real call is approve)_ — clean DDD layering, faithful reference-data conventions, comprehensive tests on both tiers, and thorough doc updates. No blocking findings.

### What's good
- **Layering is correct.** Class resolution + validation lives in the application layer (`AccountService.ResolveCreatableClass`), the blueprint is built in the domain (`NewPlayerFactory` taking a resolved `Game.Core.Classes.Class`), and entity mapping stays in the data tier. The application layer never encodes the defaults or touches entities — exactly what `backend.md`'s account/login-orchestration note prescribes.
- **Reference-data conventions respected.** Retirement is checked off the contract's `RetiredAt` (`_classes.All()`), the lean gameplay model comes from `GetClass`, and a live-contract-but-missing-lean-model is treated as a corrupt cache and thrown loudly rather than papered over — consistent with the documented fail-fast policy.
- **Exhaustive status mapping.** `InvalidClass` is threaded through both `CreateAccount` and `CreatePlayer` and handled in every controller `switch` (message + error category), preserving the "a new status is a build-time gap, not a silent success" property.
- **`StartsUnlocked` retirement is complete and behaviour-justified.** Removed from entity / contract / core model / mapper / admin repo; no production references remain; `contracts.ts` regenerated cleanly (no stale `startsUnlocked`); and the doc rewrites in `game-design.md` / `backend.md` / the #982 spike correctly reframe roots as emergent-from-kit. The explicit regression test (`FreshKitSkill_OpensItsPathOnFirstContribution_WithoutStartsUnlocked`) pins the behaviour-neutrality claim.
- **Validation parity preserved.** The new `CreateAccountRequest` carries the same `[Length(1,25)]` / `[Length(1,256)]` constraints as `LoginCredentials`; splitting it out so the class only rides creation (not every login) is the right call.
- **Dedup + base-spread logic is correct** (`[5,9,5]` → skills `[5,9]`, orders `[0,1]`; `BaseAmount` only, `AmountPerLevel` deferred to #1223), and a row is seeded for every core attribute so `PlayerStatPoints` can allocate into a zero-investment stat.

### Heads-up — transitional state (intended, worth a conscious ack)
The frontend sends a placeholder `classId: 0` (`TODO(#1225)`), and account/character creation now **rejects** an unknown/retired class. Combined with the migration's `TRUNCATE Players … CASCADE`, this branch leaves the game **unplayable until classes are authored (#1226) and the picker ships (#1225)**: existing characters are wiped, and no new character can be created because no class id 0 exists yet. This is explicitly the intended consequence of the spike's B→F→H ordering and the pre-release no-grandfathering decision, and authored content can't be migration-seeded without violating the intrinsic-vs-authored split — so this is the right shape, not a defect. Flagging only to confirm #1225/#1226 are tracked as the unblockers and that nothing (e.g. a staging deploy) depends on signup working in the interim.

### Should-fix / follow-up (non-blocking)
- **Class-authoring guard for `AttributeDistributions`.** `NewPlayerFactory` trusts the distributions to be unique-per-attribute (`ToDictionary` throws otherwise) and core-only (non-core rows are silently dropped). The factory's behaviour is defensible, but there's no admin-side guard rejecting a malformed class at save time — worth a follow-up (tech-debt / scope: small), mirroring the skill/enemy authoring anti-tamper guards. (Inline note on `NewPlayerFactory.cs`.)

### Nits
- No `CreatePlayer_RetiredClass` test mirroring `CreateAccount_RetiredClass`. The resolver is shared so the path is covered, but a symmetric test would be tidy.
- Pre-existing, out of scope: the domain `AttributeDistribution` XML doc still describes itself purely as part of the `EnemyTemplate` graph though it's now reused by `Class` (introduced in area A / #1220) — refresh opportunistically.

### Test coverage
Solid on both tiers and following the guidelines:
- **Backend:** `NewPlayerFactory` unit tests rewritten for the class-driven kit (all-selected, authored order, first-wins dedup) and the base spread (per-core-attribute rows, `AmountPerLevel` exclusion); `AccountService` integration tests for valid + unknown (both entry points) + retired class, asserting nothing is created on rejection; `ClassId` round-trips through `PlayerCacheMapper` both directions; the `StartsUnlocked`-retirement regression. No new frontend↔backend battle-parity surface is introduced (ClassId/passive/locked-base aren't read by the simulation here), so no mirrored battle test is required.
- **Frontend:** all three touched call sites (`+page.svelte`, `CharacterSwitcher`, `select/+page`) have their unit tests updated to assert the `classId: 0` payload.

---

ginderjeremiah:
Thanks for the review. Addressed the two nits and filed the follow-up:

- **Symmetric retired-class test** — added `CreatePlayer_RetiredClass_ReturnsInvalidClass` mirroring the `CreateAccount` one.
- **`AttributeDistribution` doc** — refreshed to note its reuse by `Class` (#1126), not just `EnemyTemplate`.
- **Class-authoring guard for `AttributeDistributions`** (core-only + unique-per-attribute, rejected at save time) — tracked as follow-up #1247 (tech debt / scope: small), since it belongs in the Content Authoring admin path rather than this seam.

On the transitional-state heads-up: confirmed. The unblockers are tracked — **#1225** (class picker) and **#1226** (V1 roster content) — and this is the intended B→F→H ordering with the pre-release no-grandfathering reset. Nothing depends on signup working in the interim (the game isn't publicly available).

Enabling auto-merge (squash) so it lands once the re-run is green.

---

Issue #1253, opened 2026-06-25
Offline: should the class locked base grow across the away window (vs. the stationary model)?


ginderjeremiah:
Follow-up surfaced while implementing class-system area D (#1223, spike #1126).

## Background

Area D added the class **locked base** — a level-scaled attribute fingerprint (`BaseAmount + AmountPerLevel × level`) assembled into the battler on both sides. Its acceptance criteria asked for the offline simulator to *"re-derive the locked base as levels cross the away window"* and a *"test that the locked base grows correctly across simulated level-ups."*

## What shipped in #1223, and why it deviates

The locked base is **frozen at the snapshot's window-start level**, not grown mid-window. This is a deliberate deviation, because growing only the locked base is **internally inconsistent with the existing stationary offline model (#879)**:

- The offline simulator (`OfflineProgressSimulator`) replays the whole away window from **one frozen `BattleSnapshot`**: the battler's `Level`, proficiency levels, gear, and stat allocations are *all* frozen at window-start, and exp is applied **after** the simulation (`GrantOfflineExp`). The simulator has no notion of leveling mid-loop.
- The locked base is `f(class, level)`. With the level frozen, it freezes too — which is the *only* treatment consistent with the rest of the snapshot. Growing only the locked base would produce a battler reporting level *N* while wielding a level *N+k* fingerprint.
- Truly modeling mid-window growth means un-freezing the **whole** battler's level (and then the leveling-threshold math would need extracting into a shared helper to keep live/offline parity, and `DefeatRewards` would re-measure power per level for a power→reward feedback loop) — i.e. a rework of #879's stationarity, not area D's job.

So #1223 ships the locked base frozen-at-snapshot-level (the offline path gets it for free via the snapshot), with an offline test (`Simulate_AppliesClassLockedBase_FromTheFrozenSnapshotLevel`) pinning *that* behavior. `docs/backend.md` documents the freeze.

## The question for this issue

Do we want true mid-window level progression for offline runs (all systems — level, locked base, enemy scaling implications, reward feedback), or is the stationary model the intended behavior indefinitely? The former is a deliberate #879 design decision with real complexity/risk; the latter means this is a no-op and the deviation above stands as documented.

Serious design considerations (offline stationarity), so no `claude` tag — needs an owner call before any implementation.

Related: #1223, #1126, #879.

ginderjeremiah:
[...]

Issue #1126, opened 2026-06-22
Spike: class system


ginderjeremiah:
Deferred spike spun off from #982 (proficiency system) -- tracking. See `docs/spikes/982-proficiency-system.md` (deferred; decision 3).

Research a class system as an overlay on the proficiency tree: in-class branches cheaper/stronger, possible class-locked nodes, class-determined starting roots. Strong synergy with multiple-characters-per-account (#922) -- each character a different specialization.

We would not be grandfathering this in; all characters will be deleted, and any new characters will require a class.

Classes will determine your starter kit:
- Starter skill (or maybe multiple)
- Starter weapon (Staff, Bow, Sword, Axe, etc)

Other things to consider
- Starting attributes? (Should an attribute floor be implemented based on these?)
- Reduce the amount of custom-allocated stat points and introduce class-based attribute distributions? (This would matter for away-time simulation)
- Allow class levels/advancement? (Potentially huge system with lots of overlap with proficiencies, probably not worthwhile).

Related: #982, #922.

Issue #1119, opened 2026-06-22
Client: proficiency reference + player data delivery


ginderjeremiah:
Part of spike #982 (proficiency system). See `docs/spikes/982-proficiency-system.md` -> area F.

**Depends on:** areas A + B (contracts).

Deliver the proficiency catalogue + player progress to the client.

- `GetProficiencies` (reference, version-keyed) and `GetPlayerProficiencies` (player) socket commands; codegen contracts.
- `staticData` slot + a `playerProficiencies` store; reference-data registration + local cache.

Mirror the skills/challenges delivery plumbing.

Issue #1117, opened 2026-06-22
Proficiency bonuses -> attribute pipeline (parity)


ginderjeremiah:
Part of spike #982 (proficiency system). See `docs/spikes/982-proficiency-system.md` -> area E. **The only frontend<->backend parity surface.**

**Depends on:** areas A + B. Can run parallel to area C.

Apply proficiency level bonuses through the attribute modifier pipeline.

- New `EAttributeModifierSource.Proficiency`.
- At battler assembly, generate modifiers from each proficiency Level: PerLevelModifiers x level + reached-milestone modifiers; feed into AttributeCollection (BE) and BattleAttributes (FE).
- Ship player proficiency levels to the client as player state; bonuses baked into the battle snapshot at battle start.

Tests: mirrored FE/BE parity vectors for a leveled proficiency's bonus (and a milestone bonus).

PR #1254, opened 2026-06-25
Class system D: hybrid attribute model (locked base + free pool) (#1223)

### LINKED ISSUE #1223: Class system D: Hybrid attribute model (locked base + free pool)

ginderjeremiah:
## Summary

Implements **area D** of the class-system spike ([#1126](https://github.com/ginderjeremiah/GameServer/issues/1126)) — the **hybrid attribute model**: a level-scaled, non-reallocatable **class locked base** (the attribute fingerprint / floor) plus a reduced **free pool** of manually-allocated stat points. Builds on the merged areas A–C (#1220/#1221/#1222).

## How it addresses the issue

**Backend — locked base in the battler (both the simulation and the reward power measurement):**
- `BattleSnapshot` captures `ClassId` and composes the class's `AttributeDistributions` as a locked base (`BaseAmount + AmountPerLevel × level` — the same shape and `EAttributeModifierSource.AttributeDistribution` source an enemy uses), derived from `(class, level)` and **never stored**.
- `BattleService` threads a fail-loud class resolver into `ToBattler`/`GetModifiers`, so the locked base flows into both the anti-cheat replay **and** `DefeatRewards` (a class-boosted player is measured at its true power).
- **Composition order is fixed and mirrored:** stat allocations → gear → **class locked base** → proficiency bonuses → static engine modifiers (float addition isn't associative, so position is part of the parity contract).

**Backend — the free-pool split:**
- `GameConstants.StatPointsPerLevel` reduced (6 → 2, strawman) — the locked base now supplies the bulk of attribute growth. (`[ClientMirrored]`; `STAT_POINTS_PER_LEVEL` regenerated.)
- `NewPlayerFactory` seeds the free-pool allocations **empty** (was the class base spread) — the locked base provides the starting spread, so seeding it too would double-count.

**Frontend — the parity surface, wired live on both sides:**
- New shared `UI/src/lib/battle/class-modifiers.ts` (`classLockedBaseModifiers`) — the exact mirror of the backend composition, with bit-exact parity vectors on **both** sides (`ClassLockedBaseParityTests` / `class-locked-base-parity.test.ts`), including a derived-attribute (`MaxHealth`) scenario to pin the additive ordering.
- The player's class distributions ride `PlayerData` (`LockedBaseDistribution`, resolved from the class catalogue in the login projection); `PlayerManager` memoises `classLockedBaseModifiers(distribution, level)` into the live battler (recomputed on level-up), composed in the same `additionalModifiers` slot as the proficiency bonuses.

## Scope notes

1. **The live frontend wiring is included, not deferred.** My first pass deferred it to #1225 on the proficiency #1117→#1119 precedent — but that precedent only held because proficiency was *empty* until content existed. **Every character has a class with a locked base**, so the live battler must compose it or the anti-cheat replay diverges (the e2e suite caught exactly this on the first push). The fix delivers the player's own fingerprint via `PlayerData`; the `GetClasses` **reference** delivery + the create-character picker + the attributes-screen locked/free render remain **#1225's** scope (the live battler needs none of them).

2. **Offline: the locked base is frozen at the snapshot's window-start level**, not grown mid-window. The issue's AC asked for growth, but the offline simulator (#879) is strictly *stationary* (level/proficiency/gear/allocations all frozen, exp applied after) — growing only the locked base would be internally inconsistent, and truly modeling it means un-freezing the whole battler's level (a #879 redesign). Filed [#1253](https://github.com/ginderjeremiah/GameServer/issues/1253) for that design question.

## Test plan

- [x] **Backend** — build + `dotnet format --verify-no-changes` clean. `Game.Core.Tests` (828: parity vectors, `BattleSnapshot` locked-base, offline, free-pool), `Game.Application.Tests` (583: end-to-end reward test — the locked base lifts measured power 100→200, reward 200 vs the un-boosted 800), `Game.Api.Tests` (615: `PlayerData` locked-base mapping), `Game.Infrastructure.Tests` (60) green.
- [x] **Frontend** — `npm run lint` clean; `npm run test:unit:ci` 2554 passed, coverage floors hold (new `class-locked-base-parity`, `PlayerManager.battleLockedBaseModifiers` memoisation/recompute, battle-engine composition).
- [ ] **e2e** — re-running on CI after the live-wiring fix (the first push failed e2e on the parity break above; the fix restores it — verified by the trace + the parity/integration tests).

Closes #1223


UI/src/lib/battle/class-modifiers.ts
```diff
@@ -0,0 +1,32 @@
...
+ *  surface the class system adds (spike #1126 area D). The produced modifiers are fed into the player battler
+ *  at assembly — alongside the proficiency bonuses, before the static engine modifiers — once the class
+ *  reference data and the player's class id reach the client (#1225). */
+export function classLockedBaseModifiers(
+	distributions: readonly ClassAttributeDistribution[],
+	level: number
+): AttributeModifier[] {
+	return distributions.map((distribution) => ({
+		attribute: distribution.attributeId,
+		amount: distribution.baseAmount + distribution.amountPerLevel * level,
```
ginderjeremiah, line 30: **Should-fix (parity hole the contract doesn't cover).** This recomputes `baseAmount + amountPerLevel × level` in IEEE-754 **double**, but the backend's `AttributeDistribution.GetDistributionModifier` does `(double)(BaseAmount + AmountPerLevel × level)` where both operands are **`decimal`** — the multiply/add happen in decimal, then cast. For a distribution with a non-zero fractional `amountPerLevel` (or a fractional `baseAmount` plus a non-zero level term), the two results can differ in the last bits, and the replay asserts bit-exactly (`toBe` / `Assert.Equal`, no tolerance). That would silently flag a legitimate player on the anti-cheat path.

The parity vectors deliberately dodge this (integer base/per-level, or fractional only with per-level 0), and the only record of the constraint is in the test comments. It can't manifest until class reference delivery + content (#1225) wires real distributions, so it's not breaking anything today — but there's no guard or follow-up tracking it. Before fractional class content lands, either:
- validate that a class's `AttributeDistribution`s are integer-valued (enforced where class content is authored), or
- compute the class locked base in `double` on the backend too so both sides share one arithmetic (class-specific, since enemies share `GetDistributionModifier` and don't recompute on the client), or
- give this one surface a tolerance.

At minimum, hoist the integer-only constraint out of the test comments into `docs/backend.md` / a comment here, and file a follow-up.

---

Game.Api/Controllers/LoginController.cs
```diff
@@ -33,8 +36,25 @@ public class LoginController(
...
         private readonly BattleService _battleService = battleService;
+        private readonly IClasses _classes = classes;
         private readonly ILogger<LoginController> _logger = logger;
 
+        // Projects the player to its wire DTO, resolving its class's attribute distributions (the locked-base
+        // fingerprint #1126 area D) so the client battler composes the same locked base the backend snapshot
+        // does. An unresolvable class yields an empty fingerprint rather than throwing — the player still loads.
+        private PlayerData BuildPlayerData(Player player)
+        {
+            var lockedBaseDistribution = (_classes.GetClass(player.ClassId)?.AttributeDistributions ?? [])
```
ginderjeremiah: **Should-fix (inconsistent failure policy).** This silently degrades to an **empty** fingerprint on an unresolvable class ("the player still loads"), but `BattleService.ResolveClass` and `AccountService.ResolveCreatableClass` both **throw** on the same condition (the latter: *"a miss here is a corrupt cache, not a bad request"*). A player's `ClassId` is validated at creation, so a login miss is exactly that corrupt-cache case — and the lenient path produces a *worse* outcome than failing fast: the client composes no locked base (simulates a weaker player) while the battle path throws on the same id, so the player can log in but every battle 500s on submit.

The method comment claims this applies "the player-load missing-reference policy," but the actual sibling policy here is fail-loud. Recommend reconciling — throw at login too (consistent with the other two call sites) — or, if login leniency is intentional, document why this one path tolerates the corruption the others reject.

---

ginderjeremiah (commented):
## Review: Class system D — hybrid attribute model (locked base + free pool), #1223

**Verdict:** Approve _(submitted as a comment — I'm the PR author, so GitHub won't take a gating event from me; no blocking findings)._ Clean layering, honest scope handling, and thorough mirrored tests. The findings below are should-fixes/nits, none gate the merge.

> Note: GitHub's PR diff is against the merge-base (30 files / +804). The branch is behind `main`, so a raw `main..head` diff shows phantom deletions of the proficiency-screen rework that this PR didn't touch — ignore those.

### Should-fix

- **Decimal-vs-double parity hole the "bit-exact" contract doesn't cover** — `class-modifiers.ts:28` recomputes the locked base in IEEE-754 double, while the backend (`AttributeDistribution.GetDistributionModifier`) computes it in `decimal` then casts. A non-zero fractional `amountPerLevel` (or fractional base + non-zero level) can diverge in the last bits, and the replay compares with no tolerance → a legitimate player gets flagged. The parity vectors deliberately avoid this (integer-only / per-level 0) and the constraint lives only in test comments. Can't manifest until class content/#1225 wires real distributions, but there's no guard or follow-up. Add a guard (integer-only validation, or compute the class base in double on both sides) + a doc note before fractional class content lands. (inline on `class-modifiers.ts:28`)
- **Inconsistent class-resolution failure policy** — `LoginController.cs` `BuildPlayerData` returns an empty fingerprint on an unresolvable class, but `BattleService.ResolveClass` and `AccountService.ResolveCreatableClass` both throw ("a miss is a corrupt cache"). Since `ClassId` is validated at creation, the lenient login path just produces a subtler broken state (loads fine, every battle 500s) and contradicts the method's own "player-load missing-reference policy" comment. Reconcile to fail-loud, or document why login alone is lenient. (inline on `LoginController.cs:47`)

### Test coverage

Strong overall — backend snapshot composition (incl. the `ClassId = 0` resolver edge for 0-based reference data), the fail-loud "class captured but no resolver" guard, offline frozen-base, the `NewPlayerFactory` empty-seeding, `PlayerData` mapping, and a convincing end-to-end integration test (locked base lifts measured power → reward 200 vs the un-boosted 800). Both parity suites mirror row-for-row, and `maxHealthDerivedAdditive` cleverly pins additive ordering past the 10th decimal.

One gap: the PR adds `lockedBaseModifiers === this.lastLockedBaseModifiers` to the battle engine's per-spawn rebuild gate, but no test exercises it — the mock's `battleLockedBaseModifiers` is a fixed `[]` that's never reassigned, so unlike proficiency (which has a dedicated "re-derives when the proficiency modifiers change" test) there's no locked-base equivalent, and the engine's `[...lockedBaseModifiers, ...proficiencyModifiers]` order at the `reset` seam is never asserted with both lists non-empty. Add the mirrored change-detection test + an order assertion with both non-empty.

### Nits

- `Game.Abstractions/Contracts/AttributeDistribution.cs` doc comment still reads "Read contract for an **enemy's** … distribution" — it's now reused for the class locked base too; widen it. (file not in this diff — follow-up.)
- Branch is behind `main` (lacks the proficiency-screen rework); needs an update/rebase before merge, but that's mechanical.

### On the deliberate AC deviations (both reasonable)

- **Offline base frozen at window-start level**, not grown across the window as the issue AC asked. Correct call: the offline simulator is stationary by design (level/gear/allocations all frozen, exp applied after), so growing *only* the locked base would be internally inconsistent. Documented, with #1253 filed for the redesign question.
- **Live frontend wiring included, not deferred** to #1225. Right — unlike the (empty-until-content) proficiency precedent, every character has a class, so the live battler must compose the locked base or the replay diverges. The minimal `PlayerData.LockedBaseDistribution` delivery is the correct surface for that.

### What's good

- The locked base is derived from `(class, level)` and never stored — it re-tunes for free when growth vectors change and structurally can't be stripped, which delivers the "floor" with no separate validation rule.
- Composition order made an explicit, mirrored, documented parity contract rather than an accident.
- Domain logic stays in `Game.Core` (`BattleSnapshot`), `BattleService` only threads resolvers, the API layer only maps — layering is clean throughout.

---

ginderjeremiah:
Thanks for the thorough review — all findings addressed in `0391318` (and merged `main`):

**Should-fix 1 — decimal-vs-double parity hole:** fixed at the shared source rather than papered over. `AttributeDistribution.GetDistributionModifier` now evaluates `BaseAmount + AmountPerLevel × level` in `double` (casting each `decimal` operand first) instead of decimal-then-cast, so it's bit-identical to the frontend's IEEE-754 recompute for *any* authored value — covering the **enemy** distribution (also client-recomputed) as well as the class locked base. Added a `fractionalPerLevel` parity vector to both suites — `0.1 × 3`, which the old path produced as `0.3` vs the correct double `0.30000000000000004` — so the case that would've diverged is now pinned green. Documented in `docs/backend.md` + the module/test comments (out of the test-comment-only constraint).

**Should-fix 2 — inconsistent failure policy:** `LoginController.BuildPlayerData` now **fails loud** on an unresolvable class, matching `BattleService.ResolveClass` / `AccountService`. That surfaced one fixture (`TagsControllerTests`) that seeded its player's class directly without a cache reload — added the reload (mirroring the battle-test pattern), so the fail-loud path is exercised correctly.

**Test gap:** added the locked-base change-detection case (reassign → re-derive) and an explicit order assertion that `reset` receives `[...lockedBase, ...proficiency]` with **both** non-empty.

**Nits:** widened the `AttributeDistribution` contract doc (now reused for the class fingerprint); branch updated with `main` (proficiency-screen rework).

Left as-is per your sign-off: offline base frozen at window-start (#1253 tracks the redesign question) and the live-wiring-not-deferred call. Full backend (Core 828 / Application 583 / Api 615) + frontend (2556, coverage held) + lint + format green locally; CI re-running.

---

Issue #1256, opened 2026-06-25
Class selection for the signup-path first character (CreateAccount)


ginderjeremiah:
Follow-up from #1225 (class system area F — spike #1126).

#1225 (PR #1257) delivered the class picker on the **login→select** create-character screen and the in-game character switcher, sourcing the options from the new bespoke `Login/CharacterCreationData` HTTP endpoint (authenticated, reachable pre-player-selection). That covers every create-character surface **except one**: the **signup** flow.

## The remaining gap

`UI/src/routes/+page.svelte` → `Login/CreateAccount` creates the account's *first* character and still sends a placeholder `classId: 0`. Signup runs **before any token exists** (`CreateAccount` is `[AllowAnonymous]`), so it can't call the authenticated `CharacterCreationData` endpoint to populate a picker. So a brand-new account's first character is always class 0, with no way to choose (the select-screen picker only creates *additional* characters — the account already has its first one after signup).

## The design fork

~~1. **Pick a class during signup.** Make the class options reachable pre-auth — e.g. an anonymous variant of `CharacterCreationData` (the class catalogue is low-sensitivity content) — and add the `ClassPicker` to the signup form. Crowds the auth form, and exposes the class list to unauthenticated callers (a minor info-disclosure call).~~
2. **Defer first-character creation to the select screen.** Have `CreateAccount` create the account only (no character), so the user always lands on the (empty) select screen and creates their first character through the existing, working picker. Cleaner UX and a single class-selection surface, but a backend change — `Player.ClassId` is non-nullable and `CreateAccount` currently builds the first player.

Option 2 keeps one class-selection path and avoids duplicating the picker into the auth screen; option 1 is smaller but spreads the surface. Owner's call.

## Notes

- The `ClassPicker` component, `class-summary` helpers, the `CharacterCreationData` endpoint + `CreatableClass` contract, and the `classId` wiring all landed in #1257 and are reusable.
- Placeholder lives at `UI/src/routes/+page.svelte` (the `Login/CreateAccount` post sends `classId: 0`).
- Not usable end-to-end until authored class content exists (#1226).

PR #1257, opened 2026-06-25
Client class delivery + class picker on create-character screen (#1225)

### LINKED ISSUE #1225: Class system F: client class delivery + picker on create-character screen

ginderjeremiah:
## Summary

Implements **area F** of the class-system spike ([#1126](https://github.com/ginderjeremiah/GameServer/issues/1126)) — deliver the authored classes to the game client and let players pick a class when creating a character. Builds on the merged areas A–C (#1220/#1221/#1222). Backend `GetClasses` + `IClass` already existed from area A; `CreatePlayer`/`CreateAccount` already accept `classId`.

## The constraint, and the chosen design

Character creation happens on the **login→select** screen, which runs on a *pre-selection* token — and the socket (and all the reference data it serves, including the class catalogue) requires a *selected* player (`SocketInterceptorMiddleware` 404s otherwise). So reference data can't reach the create-character screen.

Per @ginderjeremiah's call, the class options are delivered over a **bespoke HTTP endpoint** rather than the reference-data socket, kept distinct from the `staticData` class catalogue:

- **`Login/CharacterCreationData`** (new GET, authenticated but **not** player-bound) returns the creatable classes (active only) as a self-contained view model — `CreatableClass` with the signature passive, attribute fingerprint, and **starter skill / equipment names resolved server-side**. Pre-resolving the names means the picker needs no other reference set, so the kit preview is complete even before reference data loads. New contracts are codegen'd to TS; the projection is `AccountService.GetCreatableClasses`.

## Frontend

- **`ClassPicker`** is presentational (takes the options as a prop) — a radiogroup of the creatable classes plus a kit preview (fingerprint, signature passive, starter skills, weapon-first equipment) and the decorative word of power (`WordOfPower`). The view-model loads the options from the endpoint and defaults the selection to the first class; the chosen `classId` threads into `CreatePlayer`.
- The same picker drives both the **select screen** and the **in-game character switcher** (one HTTP source, one component).
- `staticData.classes` registration is **retained** as the distinct reference catalogue (separate from this endpoint), for the imminent live-battler wiring (areas D/E) — that's what the issue's "GetClasses → staticData slot" deliverable feeds.

## Scope note

The **signup** flow's first character (`Login/CreateAccount`, which runs *before* any token exists) still can't reach the authenticated endpoint, so it sends a placeholder class. Closing that is a separate UX decision (anonymous class delivery vs. deferring first-character creation to the select screen) tracked in [#1256](https://github.com/ginderjeremiah/GameServer/issues/1256). Not usable end-to-end until authored class content lands (#1226).

## History on this branch

The first two commits took a wrong turn — fetching the catalogue over the socket on the select screen, which can't work pre-selection and broke `SelectPlayer` across the e2e suite (the earlier CI failures). The third commit is the bespoke-endpoint design above; the e2e regression is resolved.

## Test plan

- [x] **Backend** — `dotnet build` clean; `Game.Api.Tests` (616) incl. new `CharacterCreationData` integration tests (resolved kit names + retired-class exclusion + unauthenticated rejection); `Game.Application.Tests` (582) green.
- [x] **Frontend** — `npm run lint` (eslint + svelte-check) clean; `npm run test:unit:ci` **2599 passed**, coverage floors hold. Updated `ClassPicker` / `class-summary` / view-model / select-page / character-switcher tests for the props-driven picker + HTTP source.
- Docs: bespoke-endpoint notes in `backend.md` and `frontend-screens.md`.

Closes #1225


ginderjeremiah:
Heads up: the e2e failures on the first commit were a real regression I've now fixed, and it forced a scope change worth your eyes before review.

**What happened:** I'd fetched the class catalogue over the socket on the login→select screen. But that screen runs on a **pre-selection token**, and the socket handshake requires a *selected* player (`SocketInterceptorMiddleware.TryLoadPlayer` → `/socket` 404, *"valid token with no selected player; session not established"*). So the fetch couldn't succeed — and worse, opening the socket there set off an auth-retry/refresh loop that consumed the single-use refresh token and broke `SelectPlayer` across the whole e2e suite.

**The fix (2nd commit):** the picker now activates **only where the class catalogue is loaded** (`staticData.classes`) — i.e. the in-game character switcher, which runs after world entry. On the pre-selection select/signup screens it stays hidden and creation falls back to the placeholder class, as before. All unit tests + lint are green again and the e2e regression is removed.

**The decision I need from you:** completing the select-screen picker (the issue's headline target) requires delivering reference data *before* a player is selected, which the current architecture doesn't allow. The two options — (A) let the socket serve reference commands pre-selection, or (B) an HTTP/login-bootstrap payload — are both architecturally significant and (B) cuts against the documented "reference reads go over the socket" convention, so I didn't want to pick one unilaterally. Written up in #1256. Once you point at a direction I'll finish the select/signup picker on top of this PR.

I left this PR **not** auto-closing #1225 for that reason.

---

UI/src/routes/select/ClassPicker.svelte
```diff
@@ -0,0 +1,219 @@
+<!-- Class picker for the create-character flow: a radiogroup of the creatable classes plus a preview of
+     the chosen class's kit (attribute fingerprint, signature passive, starter skills, weapon-first
+     equipment) and its decorative word of power. The class data — with starter skill/item names already
+     resolved — is supplied by the host from the bespoke `Login/CharacterCreationData` payload, so this
+     component is presentational and needs no reference data of its own (only attribute display metadata,
+     which degrades to enum names before the reference sets load). -->
+<div class="class-picker" data-testid="class-picker">
+	{#if classes.length === 0}
+		<p class="picker-empty" data-testid="class-picker-empty">No classes are available yet.</p>
```
ginderjeremiah: Nit: this empty-state branch is unreachable in practice. The only consumer — `CreateCharacter.svelte` — already guards the picker with `{#if classes.length > 0}`, so `ClassPicker` is never rendered with an empty `classes`. It also quietly contradicts the documented UX (`frontend-screens.md`: the picker *stays hidden* when no classes load, rather than showing a message). As-is, this branch and its `class-picker-empty` unit test exercise code the app can't reach.

Consider picking one place for the show/hide decision: either drop this empty branch and keep the host guard, or move the decision into the component (render nothing when empty) and drop the host guard. Right now it's handled in two spots with inconsistent intent.

---

Game.Api.Tests/Integration/LoginControllerTests.cs
```diff
@@ -1075,5 +1075,62 @@ public async Task Login_NonAdminUser_DoesNotGrantAdminAccess()
...
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            var result = await response.Content
+                .ReadFromJsonAsync<ApiEnumerableResponse<Game.Abstractions.Contracts.CreatableClass>>(CancellationToken);
+            Assert.NotNull(result);
+            Assert.Null(result.ErrorMessage);
+            Assert.NotNull(result.Data);
+
+            var creatable = Assert.Single(result.Data, c => c.Id == active.Id);
+            Assert.Equal("Warrior", creatable.Name);
```
ginderjeremiah, line 1115: Should-fix (test coverage): this is the only backend test that exercises the real `Class → CreatableClass` projection, but the assertions cover just id/name, the resolved skill/item names, the equipment slot, and retired exclusion. The signature-passive fields (`PassiveAttributeId`/`PassiveAmount`/`PassiveScalingAttributeId`/`PassiveScalingAmount`/`PassiveModifierType`) and `AttributeDistributions` are copied 1:1 in `ToCreatableClass` but go unverified anywhere on the backend — the test even seeds `attributeDistributions: [(EAttribute.Strength, 10m, 1m)]` (line 1091) yet never asserts they come back. Per the project's testing rules, add a couple of assertions for the passive fields and the fingerprint so a future mis-wiring of those mappings is caught.

---

ginderjeremiah (commented):
## Review: Client class delivery + class picker on the create-character screen (#1225, area F of spike #1126)

**Verdict:** Approve *(submitted as a comment — I'm the PR author, so GitHub won't take a gating event from me; no blocking findings)* — the bespoke-endpoint design is sound, well-justified, and well-tested; only a should-fix test gap and one nit.

The core constraint is handled cleanly: character creation runs pre-player-selection where the socket (and the reference data it serves) is unreachable, so the creatable classes are delivered over a purpose-built authenticated-but-not-player-bound `Login/CharacterCreationData` endpoint with starter skill/item **names resolved server-side**, keeping the picker self-contained and distinct from the `staticData` `GetClasses` catalogue. The same presentational `ClassPicker` drives both the select screen and the in-game switcher off one HTTP source. Auth, layering, and cache-safety all check out.

### Should-fix
- `Game.Api.Tests/Integration/LoginControllerTests.cs:1114` — the one backend test covering the real `Class → CreatableClass` projection asserts only id/name, resolved kit names, equipment slot, and retired exclusion. The signature-passive fields and `AttributeDistributions` (copied 1:1 in `AccountService.ToCreatableClass`) are never asserted anywhere on the backend — the test even seeds `attributeDistributions` but doesn't check them back. Add assertions for the passive + fingerprint.

### Nits
- `UI/src/routes/select/ClassPicker.svelte:8` — the `picker-empty` empty-state branch is unreachable: the only consumer guards with `{#if classes.length > 0}`, and the documented UX is hide-on-empty, not a message. Pick one place for the show/hide decision (drop the branch, or move the decision into the component and drop the host guard).

### Test coverage
- **Frontend — strong.** `class-summary.test.ts` thoroughly covers `weaponFirst` (ordering + non-mutation) and `passiveSummary` (flat/multiplicative/scaled/zero-scale/enum-fallback); `ClassPicker.test.ts` covers empty/options/select/kit-preview; `player-select-view.test.ts` covers class load + default, null-on-none, `classId` wiring, null-class fallback, and `selectClass` error-clear; the select-page and character-switcher tests were correctly updated to route the new HTTP source (and to spread the real `$lib/api` so the picker's enum-reading helpers survive the mock), and the reference-data/loading/static-data tests were updated for the new `classes` set.
- **Backend.** New `CharacterCreationData` integration tests cover resolved kit names, retired-class exclusion, and unauthenticated rejection. Gap noted above (passive/fingerprint projection unverified). No mirrored battle-logic change here, so no cross-suite parity concern.

### What's good
- Auth matches the established `Players`/`Status`/`ActiveSession` pattern (manual `_sessionService.Authenticated` check → `Unauthorized`), and the unauthenticated test confirms it.
- `ToCreatableClass` copies reference data into fresh lists (`[.. …]`) rather than mutating the cached `AsNoTracking` graphs — consistent with the reference-data gotchas in `backend.md`.
- Name resolution via `GetSkill`/`GetItem` → `GetById(...)` fails loud on a corrupt cache rather than silently dropping names, matching the gameplay `Get*` idiom; the rationale is captured in the method doc-comment.
- Display logic lives in a store-free `class-summary.ts` (attributes passed in), so it unit-tests without the static-data store and degrades to enum names before reference data loads — good separation.
- `ClassPicker` uses theme tokens (`--accent`, `--panel`, `--border-subtle`, `color-mix` over `--accent`) with a proper radiogroup ARIA shape; no hardcoded colors, and reduced-motion is handled by the global `common.scss` collapse.
- Scope is honest and well-documented: the signup-flow placeholder class is called out and tracked in #1256, and both `backend.md` and `frontend-screens.md` are updated to explain the HTTP-vs-socket split.

---

ginderjeremiah:
Both addressed in `6be4e43`:

- **Should-fix (passive/fingerprint coverage):** `CharacterCreationData_ReturnsCreatableClassesWithResolvedKitNames_ExcludingRetired` now asserts the five signature-passive fields and the seeded `AttributeDistributions` come back through the `Class → CreatableClass` projection, so a future mis-wiring of those 1:1 copies is caught.
- **Nit (two show/hide spots):** moved the decision into `ClassPicker` — it renders nothing when `classes` is empty — and dropped the duplicate `{#if classes.length > 0}` guard in `CreateCharacter`. Removed the unreachable empty-message branch + its style; the unit test now asserts the picker renders nothing on empty.

Backend (616) + frontend unit (2599) + lint all green locally; CI re-running.

---

Issue #811, opened 2026-06-16
Memoize equipmentStats / avoid full player attribute re-derivation on every enemy spawn


ginderjeremiah:
## Problem

Two related instances of "expensive getter behind a hot path / hot derivation":

1. **Every idle enemy spawn** calls `BattleEngine.reset` → `player.reset(playerManager, inventoryManager.equipmentStats)` (`UI/src/lib/engine/battle/battle-engine.ts:92-102`). `equipmentStats` (`UI/src/lib/engine/player/inventory-manager.ts:92-103`) re-flattens all equipped items + their mods into a new array on each access, and `Battler.reset` → `attributes.setData(...)` (`UI/src/lib/battle/battler.ts:203-224`) rebuilds the modifier graph and recomputes derived attributes from scratch — even though the player's attributes didn't change between spawns. In a fast idle farm this repeats every cooldown.

2. The display path has the same shape: `selectedSkills` (`player-manager.ts:31-36`, `filter().sort().map()`) and `equipmentStats` are read inside `$derived` chains in `skills-view.svelte.ts:167,188` and `inventory-view.svelte.ts:110` that also construct a `new BattleAttributes(...)`, so an unrelated reactive change re-runs the whole flatten+derive.

## Suggested direction

Memoize the manager-side getters (`equipmentStats`, `selectedSkills`), invalidated on the equip/mod/loadout mutations the managers already centralize, and only re-derive player battle attributes when the loadout/equipment actually changed between spawns.

### Labels
tech debt, claude, scope: medium

ginderjeremiah:
Note from a codebase health review: when scoping this, make sure the fix covers **both halves** of the per-spawn cost, not just `equipmentStats`.

`BattleEngine.reset` → `player.reset(playerManager, inventoryManager.equipmentStats)` runs on **every** idle enemy spawn (`UI/src/lib/engine/battle/battle-engine.ts:131`). Memoizing `equipmentStats` alone still leaves `Battler.reset` → `attributes.setData(...)` rebuilding the modifier graph and re-deriving all player battle attributes from scratch each spawn — even though the player's loadout/equipment didn't change between farm enemies. The higher-value half is skipping the **player attribute re-derive** when equipment/loadout is unchanged between spawns; the `equipmentStats` memo is necessary but not sufficient on its own.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1260, opened 2026-06-25
Class signature passive → attribute pipeline (#1224)

### LINKED ISSUE #1224: Class system E: Class signature passive -> attribute pipeline (parity)
**Depends on:** #1220, #1221

Part of the class-system spike #1126 — **area E**. See `docs/spikes/1126-class-system.md`.

The class signature passive — durable combat identity through the attribute pipeline.

## Scope
- Add `EAttributeModifierSource.Class`.
- At battler assembly (both sides), generate the class's signature-passive modifier — flat or **attribute-scaled** (`Amount + scalingAttribute × ScalingAmount`, mirroring skill-effect scaling).
- Attribute-breakdown screen surfaces the `Class` source.

## Notes
- V1 passives read only snapshot state. Conditional/"stance" passives that read live battle state are V2 (#1219).

## Tests
- Unit tests for flat + attribute-scaled passive resolution; **frontend↔backend parity vectors** covering a class passive (incl. an attribute-scaled one).

ginderjeremiah:
## Summary

Implements **area E** of the class-system spike ([#1126](https://github.com/ginderjeremiah/GameServer/issues/1126)) — the class **signature passive**, the durable combat-identity bonus fed through the existing attribute-modifier pipeline. Builds directly on area D's class→battler plumbing (#1254), exactly as the deferral note on #1224 anticipated ("post-D this becomes a small, well-scoped addition").

A signature passive is **flat** (`Amount`) or **attribute-scaled** (`Amount + ScalingAmount × value(ScalingAttribute)`, mirroring how a skill effect scales off its caster). It resolves to a single `AttributeModifier` and is composed into the battler on **both** sides — a third frontend↔backend battle-parity surface.

## Design

- **`EAttributeModifierSource.Class`** added (codegen mirrors it to TS).
- **`ClassSignaturePassive.GetModifier(resolveScalingValue)`** resolves the passive to one `Class`-sourced modifier. The `Amount + ScalingAmount × value` math runs in `double` (each authored `decimal` cast first), bit-identical to the frontend's IEEE-754 recompute — the no-tolerance anti-cheat replay demands it.
- **Composed LAST at assembly** — after the free pool, locked base, proficiency bonuses, **and the static engine modifiers** — so an attribute-scaled passive reads its scaling attribute's *fully-resolved* value (the snapshot state a V1 passive sees, so it never depends on itself) and lands in the identical per-attribute apply order on both sides (float addition isn't associative). Backend: `BattleSnapshot.ToBattler` adds it after building the `AttributeCollection`. Frontend: the passive rides `PlayerData.SignaturePassive` → `PlayerManager` → the battle engine adds it after building the battler (a data-less #811 re-arm keeps it, so it's composed only on a full rebuild).

## Scope decisions (flagged for review)

- **Excluded from the `DefeatRewards` power measurement.** That heuristic already sums only flat additive *core* contributions (it omits the static derived modifiers); the passive's scaled term is an engine-derived value of the same kind, and the coarse ±20% difficulty band makes a small identity bonus immaterial. The battle-sim parity path (the actual deliverable) does include it. Noted in `backend.md`.
- **Breakdown inspector: the `Class` source is registered, not yet rendered.** The read-only attribute-breakdown screen's `buildPlayerModifiers` doesn't surface the locked base or proficiency bonuses either (area D / proficiency client work left them out), and surfacing *only* the passive would mis-scale it. So this registers `Class` in the breakdown's `source-display` metadata (label + hue) — the same pre-registered-but-not-surfaced state `Proficiency`/`AttributeDistribution` are in — and defers wiring those battle-assembly modifiers into the inspector as a cohesive follow-up: **#1261**.

## Test plan

- [x] **Backend** — `dotnet build` clean; `Game.Core.Tests` (841), `Game.Api.Tests` (621), `Game.Application.Tests` (583) all green. New: `ClassSignaturePassiveParityTests`, `ClassSignaturePassiveTests` (domain), `BattleSnapshotTests` passive-composition cases, `PlayerDataMappingTests` passive projection.
- [x] **Frontend** — `npm run lint` (eslint + svelte-check) clean; `npm run test:unit:ci` **2638 passed**, coverage floors hold. New: `class-signature-passive-parity.test.ts` (row-for-row mirror of the backend vectors: flat, core-scaled, derived-scaled, self-scaling, multiplicative, fractional-scaling).
- [x] **Codegen** — regenerated; drift check clean (`EAttributeModifierSource.Class`, `ISignaturePassive`, `IPlayerData.signaturePassive`).

The parity suites assert **bit-exactly** and are mirrored row-for-row across the two languages, including a fractional-scaling case that decimal-then-cast would diverge on.

Closes #1224


Game.Core/Battle/BattleSnapshot.cs
```diff
@@ -117,11 +117,41 @@ public Battler ToBattler(
             Func<int, Proficiency>? resolveProficiency = null, Func<int, Class>? resolveClass = null)
         {
             var attributes = new AttributeCollection(GetModifiers(resolveItem, resolveMod, resolveProficiency, resolveClass));
+            ComposeSignaturePassive(attributes, resolveClass);
```
ginderjeremiah: **Should-fix (design — confirm the exclusion):** the passive is composed here in `ToBattler`, *not* in `GetModifiers`, so it's excluded from the `DefeatRewards` power measurement (which sums additive **core** modifiers off `GetModifiers`). The PR rationale — "scaled term is engine-derived like the static derived modifiers the power read already omits" — is sound for the *scaled* / derived-attribute case, but it doesn't cover a **flat-additive passive on a core attribute** (e.g. the `flatAdditiveCore` vector, Strength +4). That is exactly the kind of contribution `SumCoreAttributes` is designed to count, and the sibling class-identity bonus (the locked base) *is* counted via `GetModifiers`. So a flat-core signature passive is treated inconsistently with the locked base, and the only thing keeping it immaterial is an unenforced assumption about authored magnitudes — a class designed around a large flat-core signature (a plausible "combat identity") would under-report the player's power and inflate rewards.

Not blocking — this is a reward-scaling heuristic with no parity/correctness impact (the actual sim includes the passive on both sides), and it's consciously documented. But worth either folding the flat additive-core term into the measurement (for consistency with the locked base) or tightening the rationale to address the flat-core case specifically.

---

UI/src/lib/engine/battle/battle-engine.ts
```diff
@@ -198,6 +198,15 @@ export class BattleEngine {
...
 		]);
+		// Compose the class signature passive LAST — after the locked base, proficiency bonuses, and the static
+		// engine modifiers the rebuild just assembled — so an attribute-scaled passive reads the fully-resolved
+		// value of its scaling attribute (snapshot state, like a skill effect reads its caster) and lands in the
+		// same per-attribute apply order the backend's BattleSnapshot.ToBattler appends it in; float addition is
+		// not associative, so the anti-cheat replay depends on that order matching. The data-less re-arm above
+		// keeps this modifier (setData isn't re-run), so it is composed only here, on a full rebuild.
+		this.player.attributes.addModifier(
+			playerManager.battleSignaturePassiveModifier((attribute) => this.player.attributes.getValue(attribute))
+		);
```
ginderjeremiah, line 209: **Should-fix (test coverage):** the backend covers this exact composition seam directly — `BattleSnapshotTests.ToBattler_ComposesFlatClassSignaturePassive` / `...AttributeScaled...OffAssembledValue` assert the real `ToBattler` wiring. The frontend has no equivalent: the parity suite exercises the `BattleAttributes` primitive (`setData` + `addModifier`), and `battle-engine.test.ts` stubs `battleSignaturePassiveModifier` to a no-op. So the engine-level behavior here is untested — in particular the subtle claim in the comment above that the **data-less re-arm preserves this modifier** (because `setData` isn't re-run), so it's composed only on a full rebuild and not lost or double-applied across `#811` re-arms. That's a frontend-specific risk a later `resetPlayer` refactor could silently break.

Consider a focused `BattleEngine` test: a full rebuild composes a non-trivial passive onto the player battler, and a subsequent data-less re-arm preserves it exactly once. (Acknowledged this mirrors a pre-existing gap — the area-D locked base is also untested at this seam — but the re-arm-preservation behavior is the riskier of the two.)

---

UI/src/routes/game/screens/attribute-breakdown/source-display.ts
```diff
@@ -23,7 +23,11 @@ const SOURCE_KEY: Record<EAttributeModifierSource, string> = {
 	[EAttributeModifierSource.SkillEffect]: 'derived',
 	// Proficiency is a permanent progression bonus; it is not yet surfaced in the player breakdown (that
 	// lands with the proficiency client work), so it falls back to the stat-points hue until then.
-	[EAttributeModifierSource.Proficiency]: 'points'
+	[EAttributeModifierSource.Proficiency]: 'points',
+	// Class is the signature-passive bonus composed at battler assembly (#1126 area E); like the locked base
+	// and proficiency bonuses it is not yet surfaced in the player breakdown, so it gets its own hue ready for
+	// when those battle-assembly modifiers are wired in (follow-up).
+	[EAttributeModifierSource.Class]: 'class'
```
ginderjeremiah: **Nit:** `Class` gets a dedicated hue (`'class'`, plus the new `$source-class` / `--source-class` tokens), while the sibling not-yet-surfaced source `Proficiency` (line 26) falls back to the `points` hue. Both are in the same "registered but not rendered until a follow-up" state, so the asymmetry is a little surprising — and registering a brand-new theme token for a source nothing renders yet is mildly speculative ahead of #1261. Harmless either way; just flagging the inconsistency in case you'd rather both fall back until the breakdown actually surfaces them.

---

ginderjeremiah (commented):
## Review: Class signature passive → attribute pipeline (#1224, spike #1126 area E)

**Verdict:** Approve _(submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author — the intended verdict is **Approve**)_ — no blocking findings; the change is correct, parity is preserved on both sides, and the production paths are wired through. A few should-fix design/test considerations and one nit, all non-blocking.

I verified the core risk surfaces:
- **Production wiring is sound.** Both `ToBattler` call sites (`BattleService.cs:796`, `OfflineProgressSimulator.cs:102`) pass `ResolveClass`, so `ComposeSignaturePassive`'s fail-loud guard won't trip the live/offline paths. `Class.SignaturePassive` is `required`, so the `LoginController` projection can't NRE.
- **Composition order parity holds.** Backend appends the passive after `AttributeCollection` assembly (free pool → gear → locked base → proficiency → statics); the frontend adds it after `Battler.reset`'s `setData`, reading the fully-assembled scaling value. The data-less `#811` re-arm (`Battler.reset()` with no args) skips `setData`, so the previously-added passive persists — the comment's claim checks out against `battle-attributes.ts` / `battler.ts`.
- **Enum exhaustiveness.** Both `Record<EAttributeModifierSource, string>` maps in `source-display.ts` were updated; no backend switch is exhaustive over the source enum, so `Class = 9` breaks nothing.

### Should-fix
- **`DefeatRewards` excludes a flat-core passive** — `BattleSnapshot.cs:120`. The passive is composed in `ToBattler`, not `GetModifiers`, so the power measurement misses it. The documented rationale covers the scaled/derived case but not a flat-additive passive on a *core* attribute, which is exactly what `SumCoreAttributes` counts and is inconsistent with how the locked base is treated. No parity/correctness impact; either fold the flat core term in or tighten the rationale. (inline)
- **Frontend engine-seam test gap** — `battle-engine.ts:207`. The backend tests `ToBattler` composition directly; the frontend stubs the wiring to a no-op and only tests the `BattleAttributes` primitive via parity. The subtle re-arm-preservation behavior is untested. (inline)

### Nits
- **`source-display.ts:30`** — `Class` gets a dedicated hue/token while the equally-unsurfaced `Proficiency` falls back to `points`; minor asymmetry / slightly speculative ahead of #1261. (inline)
- `resolveClass(classId)` is now resolved twice per assembly (locked base in `GetModifiers`, passive in `ComposeSignaturePassive`). It's an in-memory lookup, so the cost is nil and the separation is cleaner — noting only for completeness, no change needed.

### Test coverage
Strong on the math/parity side. The mirrored vector tables (`ClassSignaturePassiveParityTests` ↔ `class-signature-passive-parity.test.ts`) are genuinely well-designed: the `fractionalScaling` case pins `double` (not decimal-then-cast) arithmetic bit-exactly, `selfScaling` confirms the passive reads the pre-passive snapshot value (never itself), and `scaledOffDerived` confirms it reads the fully-assembled scaling source. Domain resolution is isolated in `ClassSignaturePassiveTests`, and backend `ToBattler` composition is covered in `BattleSnapshotTests`. The one gap is the frontend engine seam noted above.

### What's good
- Fail-loud resolver guard in `ComposeSignaturePassive` mirrors the locked-base guard — no silent attribute drop that would fail legitimate anti-cheat replays.
- The "compose last, after statics" rule is implemented consistently on both sides with clear rationale, and the float-associativity reasoning is correct.
- Clean DDD layering (contract in `Game.Abstractions`, resolution math in `Game.Core`), theming via tokens, thorough doc-comments, and a concise `backend.md` update that records the scope decisions.

---

ginderjeremiah:
Thanks for the review — addressed all three in `359078b`:

1. **`DefeatRewards` flat-core passive (should-fix).** Folded it in rather than tightening the rationale — you're right that treating the passive differently from the locked base is a real inconsistency, not just a doc gap. New `BattleSnapshot.GetModifiersWithSignaturePassive` resolves the passive against a transient collection (same assembly `ToBattler` uses) and appends it to the modifier list the power measurement reads, so a flat (or core-additive-scaled) signature passive now counts toward power exactly like the locked base. Both reward sites (`RecordVictory`, offline) use it; composition is shared via a `ResolveSignaturePassive` helper. `SumCoreAttributes` still naturally omits the derived-attribute / multiplicative cases, as it does for every source. `backend.md` updated; added `GetModifiersWithSignaturePassive` unit tests.

2. **Frontend engine-seam test (should-fix).** Added a `BattleEngine` test that a full rebuild composes a non-trivial passive (+7 Strength) onto the player battler exactly once (→ 57, not 64 or 50), and a subsequent data-less `#811` re-arm preserves it — pinning the re-arm-preservation claim the backend can't cover.

3. **`source-display` hue nit.** Dropped the speculative `$source-class` / `--source-class` tokens; the unsurfaced `Class` source now falls back to the stat-points hue like `Proficiency`, so the two registered-but-not-rendered sources are symmetric until #1261 wires the battle-assembly modifiers into the breakdown.

Local: backend `Game.Core` 843 / `Game.Application` 583, frontend 2639, lint + codegen-drift clean.

---

PR #1263, opened 2026-06-26
Keep selection/status out of derived list projections (#1238)

### LINKED ISSUE #1238: Selection state baked into derived list projections forces a full re-map/re-sort on every click
## Problem

Several list view-models fold the current **selection** into the derived row projection, so selecting a row re-runs the entire `map → filter → sort` pipeline:

```ts
// UI/src/routes/game/screens/codex/codex-view.svelte.ts:358
enemyRows = ... e => ({ ...e, selected: e.id === this.selectedEnemyId })
// same pattern at :505 (zoneRows) and :597 (skillRows)
```

Reading `selectedEnemyId` inside the `$derived` makes the whole projection a dependency of the selection, so each click rebuilds and re-sorts the full list.

The admin Workbench has a related shape: per-record `status`/`entityWarnings` are recomputed on every keystroke in three places (`Workbench.svelte:71-75`, `WorkbenchList.svelte:24-25`), and `store.status` runs a `JSON.stringify`-based `recordsEqual` per record.

## Why it matters

Minor today (lists are small), but it scales with content: as the codex/admin catalogues grow, a click or keystroke does O(n·log n) work it doesn't need to. The fix is also a readability win.

## Suggested fix

- Keep `selected` out of the projection; apply it in the component (`class:selected={row.id === view.selectedEnemyId}`) so selection doesn't invalidate the sorted list.
- For the Workbench, memoize a per-record `{ status, warnings }` map off the store rather than recomputing it in three consumers.

## Verification

Confirmed by reading the cited view-models and Workbench components.

ginderjeremiah:
## Summary

Stops two list-projection pipelines from re-running on input that doesn't change the underlying data, per #1238.

### Codex — selection out of the row projections
`enemyRows` / `zoneRows` / `skillRows` folded the current selection into each row (`selected: e.id === this.selectedEnemyId`). Reading the selection id inside the `$derived` made the whole `map → filter → sort` pipeline a dependency of the selection, so every click rebuilt **and re-sorted** the full list.

- Dropped the `selected` flag from `EnemyRowVM` / `ZoneRowVM` / `SkillRowVM` and their projections.
- Selection now applies in the components: `class:selected={row.id === view.selectedEnemyId}` (and the zone/skill equivalents). The highlight reads the same raw `selectedXId` as before, so behaviour is unchanged — only the sorted list no longer invalidates on selection.

### Workbench — memoised per-record status + warnings
`store.status(record)` runs a `JSON.stringify`-based structural diff (`canonicalEqual`) per record, and `entityWarnings(...)` was recomputed in three consumers (page summary, save-bar counts, list rows). On every search keystroke the `{#each}` re-ran the diff for each rendered row.

- Added `recordStates` (a `$derived` `{ status, warnings }` map keyed by id) and a `stateOf(record)` accessor on `EntityStore`. The memo recomputes only when records / baseline / pending-deletes change, so a search keystroke or selection reuses the cached map.
- `counts`, `Workbench.svelte` (`liveItems` / `flagged`), and `WorkbenchList.svelte` (rows + `liveCount`) all read the memo instead of re-diffing. `status` stays public for the single-record consumers (`WorkbenchDetail`, `ChallengeRewardSection`).

Both changes are also a readability win and scale with content size (O(n·log n) per click → cached), as the issue notes.

## How it addresses the issue
Implements both suggested fixes verbatim: keep `selected` out of the codex projections and apply it in the component; memoise a per-record `{ status, warnings }` map off the store rather than recomputing it in three consumers.

## Test plan
- `npm run lint` (eslint + svelte-check) — clean.
- `npm run test:unit:ci` — 2659 passing.
- Codex: updated the view-model selection test to assert `selectedSkillId` (the contract the component now reads), and added a `Codex.test.ts` test pinning that the `selected` class is applied to the head row by default and moves to the clicked row.
- Workbench: added `entity-store` tests for `recordStates` / `stateOf` (status + required-field warnings, the absent-id fallback, and counts deriving from the memo).

Closes #1238

---
_Generated by [Claude Code](https://claude.ai/code/session_01QAkhL31JyW64Pc77ARURSN)_

### DIFF
--- a/UI/src/routes/admin/workbench/Workbench.svelte
+++ b/UI/src/routes/admin/workbench/Workbench.svelte
@@ -39,7 +39,6 @@ import { Loading } from '$components';
 import './workbench.scss';
 import type { EntityConfig, Identified } from './entities/types';
 import { EntityStore } from './entity-store.svelte';
-import { entityWarnings } from './validation';
 import WorkbenchIcon from './WorkbenchIcon.svelte';
 import WorkbenchList from './components/WorkbenchList.svelte';
 import WorkbenchDetail from './components/WorkbenchDetail.svelte';
@@ -70,9 +69,12 @@ const selected = $derived(store ? (store.items.find((it) => it.id === selId) ?? 
 const selectedBaseline = $derived(store && selected ? store.baselineOf(selected.id) : undefined);
 const liveItems = $derived.by(() => {
 	const s = store;
-	return s ? s.items.filter((it) => s.status(it) !== 'deleted') : [];
+	return s ? s.items.filter((it) => s.stateOf(it).status !== 'deleted') : [];
 });
-const flagged = $derived(liveItems.filter((it) => entityWarnings(entity, it).length > 0).length);
+const flagged = $derived.by(() => {
+	const s = store;
+	return s ? liveItems.filter((it) => s.stateOf(it).warnings.length > 0).length : 0;
+});
 
 // If the selected record was removed (e.g. deleted then saved), fall back to the first.
 $effect(() => {
--- a/UI/src/routes/admin/workbench/components/WorkbenchList.svelte
+++ b/UI/src/routes/admin/workbench/components/WorkbenchList.svelte
@@ -21,8 +21,9 @@
 
 	<div class="list-scroll">
 		{#each filtered as record (record.id)}
-			{@const status = store.status(record)}
-			{@const warns = entityWarnings(entity, record)}
+			{@const state = store.stateOf(record)}
+			{@const status = state.status}
+			{@const warns = state.warnings}
 			{@const badge = entity.listBadge?.(record)}
 			{@const edge =
 				status === 'added'
@@ -81,7 +82,6 @@
 <script lang="ts">
 import type { EntityConfig, Identified } from '../entities/types';
 import type { EntityStore } from '../entity-store.svelte';
-import { entityWarnings } from '../validation';
 import WorkbenchIcon from '../WorkbenchIcon.svelte';
 import WarnTriangle from './WarnTriangle.svelte';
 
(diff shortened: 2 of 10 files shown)


ginderjeremiah:
## Review — Approve

Clean, well-scoped performance refactor that implements both fixes from #1238 exactly as suggested. Verified against `docs/frontend.md` (runes / derived-reactivity patterns) and the test-coverage rules.

### Correctness
**Codex projections.** Dropping `selected` from `EnemyRowVM` / `ZoneRowVM` / `SkillRowVM` and applying it in the components via `class:selected={row.id === view.selectedXId}` is correct. The highlight reads the same public `$state` selection ids (`selectedEnemyId` / `selectedZoneId` / `selectedSkillId`) on the same `CodexView` instance the rows come from, so selection still reacts — but the `map → filter → sort` pipeline no longer takes the selection as a dependency and stops re-sorting on every click. The per-row `class:selected` re-eval is a cheap O(n) comparison, which is the intended win.

**Workbench memo.** The `recordStates` `$derived.by` reads `this.items`, `this.status(record)` (which reads `this.deleted` and `baseMap` ← `this.base`), and `entityWarnings(this.config, record)` — `config` is set once in the constructor and is non-reactive. So the memo's dependency set is exactly records / baseline / pending-deletes, as the doc comment claims; a search keystroke or selection change touches none of them and reuses the cached map. Deleted records remain in `items` (only added to the `deleted` set), so the map covers all rows including deleted, and `counts` deriving from `Object.values(recordStates)` is equivalent to the old per-item loop. The `stateOf` inline fallback for an unknown id is a sound defensive path.

No derived reactivity is broken; `status` is correctly kept public for the single-record consumers.

### Tests
Coverage is appropriate for the change:
- `entity-store.test.ts`: status + required-field warnings, the absent-id fallback, and `counts` deriving from the memo.
- `Codex.test.ts`: pins that the `selected` class starts on the head row and moves to the clicked row (the new component-applied contract).
- `codex-view.test.ts`: selection test updated to assert `selectedSkillId` / `selectedSkill` (the contract the component now reads) rather than a projection flag.

### CI
Green. `all-checks-passed`, `verify-frontend`, `test-frontend`, and e2e (chromium / firefox / webkit) all succeeded; backend jobs correctly skipped (frontend-only diff). The legacy commit-status API reads "pending" only because no classic statuses are posted — the check-runs are authoritative.

### Nits (non-blocking)
- None blocking. Optional: `stateOf`'s fallback duplicates the `{ status, warnings }` construction inline; minor and arguably clearer as-is, so fine to leave.

Verdict: **Approve / safe to merge.** No blockers, no should-fixes.

---

PR #1264, opened 2026-06-26
Reuse Bar and a shared focus-trap instead of re-implementing them (#1239)

### LINKED ISSUE #1239: Frontend: reuse Bar and a shared focus-trap instead of re-implementing them
`frontend.md` calls out extending shared primitives over copying them. A few surfaces re-implement primitives that already exist:

## Bar re-implementations
- `UI/src/routes/game/screens/fight/BattleTimer.svelte` and `fight/EnemyCooldown.svelte` hand-roll near-identical `.track` / `.track-fill` percentage-width markup + styles. `Bar.svelte`'s `presentational` mode + `--bar-*` tokens cover both (the warning-color swap is a `--bar-fill` override). `MiniBar` already reuses `Bar` correctly as the precedent.
- `UI/src/routes/game/screens/attributes/BudgetMeter.svelte` re-implements the clamped, divide-guarded percentage fill `Bar` provides (sibling `MiniBar` reuses it). `attributes/TheoryRow.svelte` is more custom (delta cursor) — lower priority, could use `Bar`'s overlay slot.

## Focus-trap duplication (+ latent bug)
`Modal*` and `Popover.svelte:58-89` carry a near-identical `trapTab` / focus-capture / restore / scroll-lock implementation. Both query **only** `button:not([disabled])` for focusables — a Modal or Popover containing a link, input, or `[tabindex]` element would not trap or anchor focus onto it. No live bug (both current consumers are button-only), but it's a latent trap for any future content.

**Suggested:** extract a shared focus-trap action/util and widen the focusable selector; migrate `BattleTimer`/`EnemyCooldown`/`BudgetMeter` onto `Bar`.

Minor cleanups noted in passing (fold in or split as you prefer):
- `UI/src/components/nav-menu/types.ts` — `INavMenuItem` is dead (no consumer; `NavSidebar` uses a local `ScreenDef`); delete it and the `export *` re-export.
- `UI/src/routes/admin/workbench/components/challenge/RewardPicker.svelte` + `ChallengeRewardSection.svelte` — inline dismissable overlay with no Escape/focus management; the app standardizes this via `Popover.svelte`.

## Verification
Confirmed by reading the cited components and `Bar.svelte` / `Popover.svelte`.

ginderjeremiah:
## Summary

Extends shared primitives over copies (per `frontend.md`'s "extend a primitive rather than copy it" guidance), addressing #1239.

### Shared focus-trap action (+ latent-bug fix)
`ModalHost` and `Popover` carried a near-identical `trapTab` / focus-capture+restore / scroll-lock implementation, and both queried **only** `button:not([disabled])` for focusables — so a modal/popover containing a link, input, or `[tabindex]` element would not trap or anchor onto it.

- Added `UI/src/components/focus-trap.ts`: a `focusTrap` Svelte action owning Tab-trap, Escape→dismiss, body scroll-lock (via a caller-supplied class), and focus capture-on-mount / restore-on-destroy.
- The focusable selector is **widened** to `a[href], button:not([disabled]), input:not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"])` — fixing the latent button-only trap bug noted in the issue.
- `ModalHost` and `Popover` now use `use:focusTrap` and drop their duplicated handlers. Per-open **initial focus** stays component-specific (the modal must aim at its kind-appropriate action and re-aim per queued modal; the popover takes the first focusable), reusing the exported `FOCUSABLE_SELECTOR`.

### Bar re-implementations
`BattleTimer`, `EnemyCooldown`, and `BudgetMeter` hand-rolled `.track` / `.track-fill` percentage-width markup + styles that `Bar`'s `presentational` mode + `--bar-*` tokens already cover (the warning-hue swap is a `--bar-fill` override; the glow is `--bar-fill-shadow`). All three now compose `Bar`, matching the `MiniBar` precedent.

### Minor cleanup
- Deleted the dead `INavMenuItem` type (no consumer) and its `export *` re-export; removed the now-empty `nav-menu/types.ts`.

## How it addresses the issue
Implements the issue's primary asks verbatim — migrate `BattleTimer`/`EnemyCooldown`/`BudgetMeter` onto `Bar`, and extract a shared focus-trap with a widened selector — plus the dead-`INavMenuItem` cleanup folded in.

## Scope notes / deferred
- `TheoryRow` (delta-cursor, marked lower-priority in the issue) is left as-is.
- The `RewardPicker` / `ChallengeRewardSection` inline-overlay → `Popover` migration is **deferred to follow-up #1265**: that overlay is an in-flow expanding panel, and moving it to `Popover` (a centered backdrop dialog) is a UX change rather than a pure refactor, so it warrants its own decision.

## Test plan
- `npm run lint` (eslint + svelte-check) — clean.
- `npm run test:unit:ci` — 250 files / 2663 tests passing.
- Added `tests/components/focus-trap.test.ts`: Tab wrapping, escaped-focus recapture, the widened selector (links/inputs/`[tabindex]` trapped; disabled/`tabindex="-1"` skipped), Escape→onEscape, scroll-lock add/release, focus restore on destroy.
- Updated `BattleTimer` / `EnemyCooldown` tests for the Bar-based markup (`.bar-fill`, warning hue via the `--bar-fill` token). The existing `ModalHost`/`Popover` suites pass unchanged, pinning the refactor's behavioral parity.

Closes #1239

### DIFF
--- a/UI/src/components/ModalHost.svelte
+++ b/UI/src/components/ModalHost.svelte
@@ -4,12 +4,15 @@
 	a focus trap, focus capture+restore, and a body scroll lock. Mounted once in the root layout
 	alongside the toast container.
 -->
-<svelte:window onkeydown={onKeydown} />
-
 {#if active}
 	<div class="modal-layer">
 		<button class="backdrop" type="button" tabindex="-1" aria-label="Dismiss dialog" onclick={dismiss}></button>
-		<div class="shell" tabindex="-1" bind:this={shell}>
+		<div
+			class="shell"
+			tabindex="-1"
+			bind:this={shell}
+			use:focusTrap={{ onEscape: dismiss, scrollLockClass: 'modal-open' }}
+		>
 			<!-- Key on the modal id so a queued modal remounts and replays the entrance animation
 			     rather than just swapping its props in place. -->
 			{#key active.id}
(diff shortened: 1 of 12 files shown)


ginderjeremiah:
## Review — Approve

Clean, faithful refactor that does exactly what #1239 asked and nothing more. Verified the focus-trap extraction, the widened selector, the Bar migrations, and the dead-type removal against the codebase and the project's "extend a primitive rather than copy it" guidance.

### Focus-trap extraction — correct
- `focusTrap` (`components/focus-trap.ts`) faithfully reproduces the prior Tab-trap / Escape→dismiss / scroll-lock / capture-on-mount + restore-on-destroy behavior. The structural change — registering the `window` keydown listener only while the action is mounted (behind the overlays' `{#if}`) rather than a permanently-mounted `<svelte:window>` gated by `if (!active)` — is cleaner and equivalent.
- Focus-restore ordering is preserved: the action captures `document.activeElement` at init (which runs before the consumer's `$effect` moves focus inward), so restore still targets the pre-open element. `??=`-style "capture once" is no longer needed because the action's lifetime *is* the open lifetime.
- Widened `FOCUSABLE_SELECTOR` is the standard interactive-element set (`a[href]`, non-disabled `button`/`input`/`select`/`textarea`, `[tabindex]:not([tabindex="-1"])`) and correctly fixes the latent button-only trap. `document` access at init is browser-only (actions never run in SSR), so no hydration concern.
- The component-specific initial-focus step is rightly kept out of the action (modal aims kind-aware + re-aims per queued modal; popover takes first focusable) and reuses the exported selector. `frontend-overlays.md` updated to document the shared action.

### Bar migrations — faithful
- `BattleTimer`, `EnemyCooldown`, `BudgetMeter` now compose `Bar` in `presentational` mode, matching the `MiniBar` precedent. Confirmed against `Bar.svelte`: every token used (`--bar-height/-radius/-track-bg/-fill/-fill-shadow/-transition`) exists and the overrides reproduce the original height/radius/track/fill/glow/transition exactly. Warning hue rides `--bar-fill` and the BudgetMeter glow rides `--bar-fill-shadow` as intended.

### Cleanup — confirmed
- `INavMenuItem` has no consumers (only the deleted file referenced it); the `export *` re-export is dropped and `components/index.ts`'s `NavSidebar` export is unaffected.

### Tests
- New `focus-trap.test.ts` covers Tab wrap, escaped-focus recapture, the widened selector (links/inputs/`[tabindex]` trapped; disabled + `tabindex="-1"` skipped), Escape→onEscape, scroll-lock add/release, and focus restore on destroy. `BattleTimer`/`EnemyCooldown` tests updated to the `.bar-fill` markup and the `--bar-fill` token assertions. Existing `ModalHost`/`Popover` suites pass unchanged, pinning behavioral parity.

### Deferral
- Deferring the `RewardPicker`/`ChallengeRewardSection` migration to #1265 is reasonable: it's an in-flow expanding panel, and moving it to the centered-backdrop `Popover` is a UX decision, not a pure refactor. #1265 captures both options (lighter-weight inline `focusTrap` vs. full `Popover`) and is correctly linked. `TheoryRow` left as-is per the issue's lower-priority note.

### CI
verify-frontend, test-frontend, all three e2e browsers (chromium/firefox/webkit), and `all-checks-passed` are green. Backend jobs skipped (no backend changes). The PR's `mergeable_state: blocked` is the branch-protection gate, not a failing check.

**No blockers, no should-fixes, no nits.** Recommend merge.

---

PR #1278, opened 2026-06-26
Surface battle-assembly modifiers in the attribute-breakdown inspector

### LINKED ISSUE #1261: Surface battle-assembly modifiers (locked base, proficiency, class passive) in the attribute-breakdown inspector
Follow-up from #1224 (PR #1260, class-system spike #1126 area E), and the analogous gaps left by area D (#1254) and the proficiency client work.

## Problem

The read-only **attribute-breakdown inspector** (`UI/src/routes/game/screens/attribute-breakdown/attribute-breakdown-view.svelte.ts`, `buildPlayerModifiers`) assembles the player's modifier set from **allocations + equipped gear + static engine modifiers only**. It does **not** include the battle-assembly modifiers that the *actual* battler now carries:

- the class **locked base** (`EAttributeModifierSource.AttributeDistribution`, area D),
- the **proficiency** bonuses (`EAttributeModifierSource.Proficiency`), and
- the class **signature passive** (`EAttributeModifierSource.Class`, area E / #1224).

So the inspector — whose header comment promises it "can never disagree with the totals the simulation actually uses" — is now a slightly **stale mirror**: it under-reports any attribute those three contribute to. The source-display metadata for all three sources is already registered (`source-display.ts`), with `Proficiency`/`AttributeDistribution` explicitly noted as "not yet surfaced," and #1224 adds `Class` in the same pre-registered state.

## Why these go in together

The class signature passive is **attribute-scaled** off the resolved value of its scaling attribute. Surfacing the passive *without* the locked base (and proficiency) in the inspector's modifier set would compute its scaled amount against a **different** scaling-attribute value than the battle does — i.e. the inspector would show a passive amount that disagrees with the simulation. So the passive can only be surfaced *correctly* once the locked base (and proficiency, for completeness) are in the same set. They're one cohesive change, not three.

## Suggested approach

- Extend `buildPlayerModifiers` to also compose: the locked base (`classLockedBaseModifiers(playerManager.lockedBaseDistribution, level)`), the proficiency bonuses (`playerProficiencies.battleModifiers`), and the class signature passive (`classSignaturePassiveModifier`, resolved against the assembled set — a two-pass within the builder, since the scaled term needs the resolved scaling value).
- Keep the apply order identical to the battler (`BattleSnapshot.ToBattler` / the battle engine): free pool → gear → locked base → proficiency → statics → passive (last).
- Confirm the inspector totals match `BattleAttributes` for a scenario exercising all three (the inspector already claims this invariant; add a test that pins it).
- The `Class`/`AttributeDistribution`/`Proficiency` hues + labels already exist in `source-display.ts`; just wire the lines through.

## Notes

- Pure read-only inspector change; no battle-logic or parity-surface change (the battler already composes all three correctly — this only makes the *inspector* faithful again).
- Surfaced during #1224.

ginderjeremiah:
## Summary

The read-only **attribute-breakdown inspector** assembled the player's modifier set from **allocations + equipped gear + static engine modifiers only**. It did **not** include the battle-assembly modifiers the *actual* battler now carries:

- the class **locked base** (`AttributeDistribution`, area D),
- the **proficiency** bonuses (`Proficiency`, #982 area E), and
- the class **signature passive** (`Class`, #1224 / area E).

So the inspector — whose header promises it "can never disagree with the totals the simulation actually uses" — had become a **stale mirror**, under-reporting any attribute those three contribute to.

## How it addresses the issue (#1261)

- **`buildPlayerModifiers` now composes the same set the live player battler does** (`BattleEngine.resetPlayer` → backend `BattleSnapshot.ToBattler`), in the identical apply order: allocated stat points → gear → **class locked base** → **proficiency** → engine statics → the **class signature passive last**.
- The passive is **attribute-scaled off the resolved value of its scaling attribute**, so it's composed via a **lazy two-pass** within the builder (build the set, then add the passive reading the assembled scaling value — like a skill effect reads its caster), matching where the battler appends it. Float addition isn't associative, so this order is load-bearing for parity.
- Non-contributing (`+0` / `×1`) modifiers are skipped, so the **flat no-op signature-passive default** can't spuriously self-select its attribute into the inspector.
- **Display wiring:** the three sources are added to `SOURCE_ORDER` (a source absent there is silently dropped from the by-source view), given contribution-line labels (`Class base` / `Proficiency` / `Signature passive`), and — since they previously fell back to existing hues — given **distinct `--source-distribution` / `--source-proficiency` / `--source-class` theme tokens** so the stacked bar and legend stay disambiguated (three sources no longer share the stat-points hue). The prior `source-display.ts` comments anticipated minting these "until #1261 wires those modifiers in."

Pure read-only inspector + theming change — no battle-logic or parity-surface change (the battler already composes all three correctly; this only makes the *inspector* faithful again).

## Test plan

- [x] New **parity test** pinning `view.computedFor(attr).total` to `BattleAttributes` (the live battle path, used as an independent oracle) for a scenario exercising all three sources, including a scaled passive.
- [x] Unit coverage for the apply order (gear → locked base → proficiency → statics), the scaled-passive resolution + last-position, the no-op skip, the by-source grouping surfacing the new sources, and the new source labels.
- [x] Updated the `AttributeBreakdown` component test's mocks for the new manager/store surfaces.
- [x] `npm run test:unit:ci` — full suite green.

## Notes

- No `docs` change warranted: the inspector↔`BattleAttributes` parity invariant and the locked-base/proficiency/passive assembly order are already documented in `frontend.md`; this restores faithfulness to those documented patterns rather than establishing a new one.
- Per-proficiency provenance on the line labels (e.g. the proficiency name) was intentionally **not** added — the inspector consumes the parity-guaranteed `playerProficiencies.battleModifiers` array directly (no provenance) to avoid duplicating the flattening logic and risking drift from the parity source; lines are labelled by source group instead. A follow-up could thread proficiency names through if desired.

Closes #1261

https://claude.ai/code/session_01QhtosZvDYMbKb4fCW9AC4q

---
_Generated by [Claude Code](https://claude.ai/code/session_01QhtosZvDYMbKb4fCW9AC4q)_

### DIFF
--- a/UI/src/lib/battle/attribute-collection.ts
+++ b/UI/src/lib/battle/attribute-collection.ts
@@ -141,14 +141,19 @@ export interface SourceGroup<T extends AttributeModifier = AttributeModifier> {
 	lines: AppliedModifier<T>[];
 }
 
-/** The display order sources are grouped/stacked in (base → points → gear →
- *  mods → derived), matching the backend's `EAttributeModifierSource` intent. */
+/** The display order sources are grouped/stacked in (base → class locked base → points → proficiency →
+ *  gear → mods → derived → class signature passive), matching the backend's `EAttributeModifierSource`
+ *  intent. A source absent here is dropped from the by-source view, so every player-facing source must
+ *  be listed. */
 export const SOURCE_ORDER: readonly EAttributeModifierSource[] = [
 	EAttributeModifierSource.BaseValue,
+	EAttributeModifierSource.AttributeDistribution,
 	EAttributeModifierSource.PlayerStatPoints,
+	EAttributeModifierSource.Proficiency,
 	EAttributeModifierSource.Item,
 	EAttributeModifierSource.ItemMod,
-	EAttributeModifierSource.Derived
+	EAttributeModifierSource.Derived,
+	EAttributeModifierSource.Class
 ];
 
 /** A computed attribute's additive lines bucketed by source (in {@link SOURCE_ORDER}),
--- a/UI/src/routes/+layout.svelte
+++ b/UI/src/routes/+layout.svelte
@@ -263,9 +263,12 @@ $effect(() => {
 	// the breakdown). Consumed via the helpers in the attribute-breakdown screen's
 	// source-display.
 	--source-base: #{colors.$source-base};
+	--source-distribution: #{colors.$source-distribution};
 	--source-points: #{colors.$source-points};
+	--source-proficiency: #{colors.$source-proficiency};
 	--source-item: #{colors.$source-item};
 	--source-mod: #{colors.$source-mod};
 	--source-derived: #{colors.$source-derived};
+	--source-class: #{colors.$source-class};
 }
 </style>
--- a/UI/src/routes/game/screens/attribute-breakdown/attribute-breakdown-view.svelte.ts
+++ b/UI/src/routes/game/screens/attribute-breakdown/attribute-breakdown-view.svelte.ts
@@ -31,7 +31,7 @@ import {
 } from '$lib/battle';
 import { attributeName } from '$lib/common';
 import { playerManager, inventoryManager } from '$lib/engine';
-import { staticData } from '$stores';
+import { staticData, playerProficiencies } from '$stores';
 
 /** An {@link AttributeModifier} carrying display-only provenance so the
  *  breakdown can label each contribution (which item/mod it came from). The core
@@ -150,7 +150,9 @@ export function fmtSigned(n: number, dec?: number, pct = false): string {
 
 /** Shared shape of a contribution-line label: both the full and the short variants resolve a derived
  *  line to its source attribute and any gear line to the item/mod name, differing only in the fixed
- *  phrasing used for the base-value and stat-point sources (`base`/`statPoints`). */
+ *  phrasing used for the base-value and stat-point sources (`base`/`statPoints`). The class
+ *  battle-assembly sources (locked base / proficiency / signature passive) have no per-line provenance
+ *  to surface, so each gets a fixed phrase shared by both variants. */
 function contributionLabel(line: AppliedModifier<LabeledModifier>, base: string, statPoints: string): string {
 	switch (line.source) {
 		case EAttributeModifierSource.Derived:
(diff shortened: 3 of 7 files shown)


ginderjeremiah:
## Review: Surface battle-assembly modifiers (locked base, proficiency, class passive) in the attribute-breakdown inspector (#1261)

**Verdict: Approve** — the inspector now composes the identical modifier set, in the identical apply order, as the live player battler; parity is pinned by a genuine independent oracle and the change is a faithful read-only restoration of an already-documented invariant.

### What it does
Extends `buildPlayerModifiers` to compose the three battle-assembly sources the actual battler carries but the inspector was omitting — the class locked base (`AttributeDistribution`), proficiency bonuses (`Proficiency`), and the class signature passive (`Class`) — in the battler's order (allocations → gear → locked base → proficiency → engine statics → signature passive last). The passive is resolved via a lazy two-pass against the assembled set, mirroring how the battler appends it after build. Adds distinct `--source-distribution` / `--source-proficiency` / `--source-class` theme tokens, source-display labels (`Class base` / `Proficiency` / `Signature passive`), and `SOURCE_ORDER` entries.

### Verification
- **Apply-order parity confirmed against the engine.** `BattleAttributes.setData` composes `[...base, ...additionalModifiers, ...STATIC_ATTRIBUTE_MODIFIERS]` and `BattleEngine.resetPlayer` passes `[...lockedBase, ...proficiency]` as the additional modifiers, then `addModifier(signaturePassive)` last. `buildPlayerModifiers` reproduces exactly this sequence. Since float addition isn't associative, this order is load-bearing — and it matches.
- **No-op skip is total-preserving.** `contributes` only drops `+0` additive / `×1` multiplicative modifiers, which cannot change any total, so skipping them keeps inspector totals identical to the battler while avoiding empty rows and the flat no-op passive spuriously self-selecting its attribute. The passive's scaling read resolves against the same value either way (the skipped entries are no-ops by definition).
- **Docs correctly unchanged.** The inspector↔`BattleAttributes` parity invariant is already in `frontend.md` ("the same `computeAttributes` path the breakdown screen uses"); this restores faithfulness rather than introducing a new pattern.

### Test coverage
Strong and exactly what #1261 asked for:
- An independent-oracle **parity test** that builds a `BattleAttributes` from the same inputs (the live battle path) and asserts `view.computedFor(attr).total` matches `battler.getValue(attr)` across a scenario exercising all three sources plus a scaled passive.
- Apply-order test (gear → locked base → proficiency → statics), scaled-passive last-position + resolution test, no-op-skip test, by-source grouping test, and the new source-label assertions.
- `AttributeBreakdown.test.ts` mocks updated for the new manager/store surfaces.

### What's good
- Faithful, minimal, read-only change with the load-bearing ordering reasoning captured in comments.
- Theming done properly through layout custom properties + distinct, mutually-disambiguated source hues, no hardcoded colors in the component.
- The lazy two-pass resolution is a clean mirror of the battler's "build then resolve passive against caster" pattern.

No blocking, should-fix, or nit findings. CI green (`all-checks-passed` success; backend jobs correctly skipped for a frontend-only change). Merging.

---

PR #1281, opened 2026-06-26
Defer first-character creation to the select screen (#1256)

### LINKED ISSUE #1256: Class selection for the signup-path first character (CreateAccount)

ginderjeremiah:
## Summary

Closes the last create-character surface left without class selection: the **signup** path. Per the issue's chosen design (**Option 2** — Option 1 was struck through), signup no longer creates the account's first character with a placeholder `classId: 0`. Instead `CreateAccount` creates the account only, and a freshly signed-up account lands on the (empty) character-select screen and creates its first character through the **same class picker** as any additional one. Class selection becomes a single surface rather than being duplicated into the pre-auth signup form (which couldn't reach the authenticated `CharacterCreationData` endpoint anyway).

## Backend

- `CreateAccountRequest` drops `ClassId`; `AccountService.CreateAccount` drops the `classId` param, the class resolution, and the `NewPlayerFactory` call — it now creates the account only.
- `CreateAccountStatus.InvalidClass` removed (no class is validated at signup) and the controller's exhaustive switch updated.
- `IUsers.CreateAccount` / `Users.CreateAccount` drop the `NewPlayer` blueprint and insert only the user row (the active-username uniqueness guard is unchanged).
- Regenerated the TypeScript client — `ICreateAccountRequest` no longer carries `classId`.

## Frontend

- The login page drops `classId: 0` and navigates to `/select` even for an account with no characters (an empty character list is now a valid post-signup state, so the old "This account has no characters." error is gone).
- `PlayerSelectView` gains an opt-in `openCreateWhenEmpty` option. The `/select` screen sets it so the create form **auto-opens** for a brand-new account and the heading adapts to *"Create your character."* The in-game character switcher (whose list can also be empty for a single-character account) is unaffected — it does not opt in.

## How it addresses the issue

A new account's first character is now chosen with a real class through the existing, working `ClassPicker`, so there is no more placeholder `classId: 0`. The class picker lives on exactly one surface (the select screen), used identically for the first character and any additional ones.

## Test plan

- [x] **Backend** — `dotnet build` clean (0 warnings). Full suites green: `Game.Application.Tests` (588), `Game.Api.Tests` (624), `Game.Core.Tests` (843), `Game.Infrastructure.Tests` (60). The full-blueprint and starter-equipment-with-granted-skill coverage that previously rode `CreateAccount` was **moved onto `CreatePlayer`** (the path that now builds every character); the obsolete signup class-validation tests were removed.
- [x] **Frontend** — `npm run lint` (eslint + svelte-check) clean; `npm run test:unit:ci` **2675 passed**, coverage floors hold. Added `PlayerSelectView` tests for `openCreateWhenEmpty`; updated the login-submit and select view-model tests.
- [x] **Codegen** — regenerated; the only drift is the removed `classId`.
- [x] **E2E (Playwright)** — **not run locally** (needs the full dockerized API stack to be brought up + seeded). The shared `e2e-tests/helpers.ts` was updated so a freshly signed-up account creates its first character before selecting it (new `createFirstCharacter` helper), and `select.test.ts` was rewritten for the empty-then-create flow. `e2e-seed.sql` comments updated (class 0 / skills 0-2 are still needed — now for the select-screen create flow rather than signup).
- Docs: updated the account-orchestration and post-signup-character sections in `backend.md`, and the create-character note in `frontend-screens.md`.

Closes #1256

---
_Generated by [Claude Code](https://claude.ai/code/session_017BJG4oKMGijRz2A7G3ojbd)_


UI/src/routes/select/+page.svelte
```diff
@@ -82,7 +88,10 @@ onMount(() => {
 
 	view = new PlayerSelectView(
 		{ selectPlayer, createPlayer, confirmTakeover: confirmSessionTakeover, enterWorld, loadCreationData },
-		summaries
+		summaries,
+		// A freshly signed-up account arrives here with no characters; open the create form so the first
+		// character creation (the only class-selection surface now — #1256) is front and centre.
+		{ openCreateWhenEmpty: true }
```
ginderjeremiah, line 93: **Should-fix (test coverage):** This `openCreateWhenEmpty: true` wiring — and the empty-account flow it enables — has no page-level test. `select-page.test.ts` still only exercises a non-empty handoff (redirect-on-`null`, render cards, select, create-additional). The PR's central new page behavior is untested here: an empty `[]` handoff should render the *auto-opened* create form with the "Create your character." heading and **not** be treated as "no handoff."

The view-model option is unit-tested (`player-select-view.test.ts`) and the login side hands off `[]` (`login-submit.test.ts`), but the receiving end isn't. It matters because correctness hinges on `if (!summaries)` (line 84) distinguishing `[]` (valid → render) from `null` (redirect) — a fragile distinction a later "simplification" to `!summaries?.length` would silently break, taking the whole new-account flow with it and nothing to catch it.

Please add a `takeMock.mockReturnValue([])` case asserting: no redirect to `/`, the create form is open (`new-name-input` visible), and the heading reads "Create your character."

---

ginderjeremiah (commented):
## Review: Defer first-character creation to the select screen (#1256)

**Verdict:** Approve — clean Option-2 implementation; layering, exhaustive-switch discipline, and the doc updates are all solid. One should-fix on test coverage and one pre-existing follow-up, neither blocking. _(Posted as a comment because I'm the PR author — GitHub won't take a gating vote from me; the call is Approve.)_

### Should-fix
- **Page-level coverage for the new empty-account flow is missing** — `UI/src/tests/routes/select/select-page.test.ts`. The PR's central new page behavior (an empty `[]` handoff renders the auto-opened create form + "Create your character." heading and is **not** redirected as "no handoff") has no page-level test. Covered indirectly by the view-model and login-submit suites, but the receiving end and the fragile `[]`-vs-`null` distinction in `+page.svelte:84` aren't. Detail + suggested test in the inline comment on `+page.svelte:94`.

### Nits / follow-ups
- **(pre-existing) The `?? 0` class fallback now backs first-character creation too** — `UI/src/routes/select/+page.svelte:53`. `createPlayer` coerces a null class to `classId: 0`, and `CreateCharacter.svelte` gates submit only on name validity, not on the catalogue having loaded — so if `CharacterCreationData` fails to load the picker is hidden but creation still posts `classId: 0`, the exact placeholder #1256 set out to remove. It's pre-existing and intentional (documented in `frontend-screens.md`), and the backend now rejects an invalid class (`CreatePlayer` → `InvalidClass`), so the worst case is a clear error rather than a silent mis-class — *unless* class 0 happens to be live. This PR doesn't introduce it, but every account now hits this path, so it's worth a follow-up: gate submission on a loaded class, or surface a "classes unavailable, try again" state when the picker is required.

### Test coverage
- **Backend — good.** The full-blueprint and starter-equipment-with-granted-skill coverage was *moved* onto `CreatePlayer` (not dropped), and the removed `CreateAccount_{Unknown,Retired}Class` tests are already mirrored by existing `CreatePlayer_{Unknown,Retired}Class` tests (`AccountServiceIntegrationTests.cs:637,658`) — no net loss. `CreateAccount_NewUsername_InsertsUserWithNoCharacter` asserts the new no-character invariant; the concurrency / duplicate / archived-reuse paths are all updated for the new signature.
- **Frontend.** The `openCreateWhenEmpty` matrix (on+empty, on+non-empty, off+empty) is covered in `player-select-view.test.ts`, and `login-submit.test.ts` is updated to assert the empty-list handoff + navigate. The remaining gap is the page-level empty path noted above.
- **No dead code.** `ResolveCreatableClass` / `NewPlayerFactory` / `_classes` all still serve `CreatePlayer`/`GetCreatableClasses`; `PlayerMapper` still backs `Users.CreatePlayer`. The `CreateAccountStatus` enum and the controller's exhaustive switch stay consistent after dropping `InvalidClass`.

### What's good
- Genuine Option-2 execution: `CreateAccount` is account-only now, domain defaults stay in `NewPlayerFactory` (Core), the app layer orchestrates, and the controller stays a thin total map — class selection is a single surface, as the issue intended.
- `openCreateWhenEmpty` as an opt-in is the right call — it correctly leaves the in-game switcher (also possibly empty for a one-character account) out of forced-create mode, with a clear comment explaining why.
- Reactive heading/subtitle that flips once the first character exists is a nice touch.
- Thorough, accurate doc updates across `backend.md`, `frontend-screens.md`, and the `e2e-seed.sql` comments, and the e2e `createFirstCharacter` helper keeps the suite honest about the new flow.

---

UI/src/routes/select/+page.svelte
ginderjeremiah: Addressed in d282e51.

- **Should-fix (page-level test):** added a `takeMock.mockReturnValue([])` case to `select-page.test.ts` asserting the empty handoff renders the auto-opened create form with the "Create your character." heading, no `player-card`, and **no** redirect to `/` — pinning the `[]`-vs-`null` distinction at the receiving end.
- **Nit (`?? 0` fallback):** per your follow-up direction, fixed inline rather than deferring — a class is now mandatory. `PlayerSelectView.create` blocks creation (with an inline error) when none is selected, both hosts drop the `?? 0`, and the create form gates the Create button on a selected class. When the catalogue fails to load, the form now shows a loading hint → unavailable + retry state (`classesLoading` + `retryLoadClasses`) instead of silently posting class 0. Covered by new view-model and page-level tests.

---

PR #1283, opened 2026-06-26
Split immutable Codex row projections from interaction state (#1270)

### LINKED ISSUE #1270: Codex row projections rebuild the full immutable projection (incl. an O(enemies×zones) boss lookup) on every search keystroke
## Problem

`codex-view.svelte.ts`'s `enemyRows` is a single `$derived.by` (`UI/src/routes/game/screens/codex/codex-view.svelte.ts:339`) that reads `this.filteredEnemies`, `this.search`, `this.sort`, **and** `this.selectedEnemyId` (via the `selected` field), then maps the entire filtered catalogue before filtering/sorting:

```ts
return this.filteredEnemies
  .map((e) => {
    const range = levelRange(e, zones);
    const zoneNames = e.isBoss
      ? [zones.find((z) => z?.bossEnemyId === e.id)?.name ?? '']   // O(zones) per boss
      : e.spawns.map((sp) => zones[sp.zoneId]?.name ?? '');
    return { id, name, band: formatBand(range), zoneCount, skillCount, searchText: [...].join(' ').toLowerCase(), selected: e.id === this.selectedEnemyId };
  })
  .filter((row) => matchesEnemySearch(row, this.search))
  .sort(sortEnemyRows(this.sort));
```

None of the projected fields except `selected` depend on `search` or `sort`, yet the whole `.map` — including `levelRange`, the per-boss `zones.find` (O(enemies × zones) across the catalogue), and the `searchText` join+lowercase — re-runs on **every search keystroke and every sort change**. The parallel zone-row and skill-row deriveds in the same file share the shape.

## Why it matters

Search is keystroke-driven, so the heavy projection (especially the boss `zones.find`) runs on each keypress and scales with catalogue size as content grows.

## Suggested fix

Split into two deriveds:
- a stable `enemyProjections` derived from only `staticData.enemies`/`zones` — the immutable per-enemy view-model incl. `searchText`/`band`/`zoneCount`;
- a lightweight `enemyRows` that only `.filter(search).sort(sort)` over it.

Apply the same split to the zone/skill row deriveds.

## Relationship to #1238

The `selected` field baked into this projection is the selection-on-click dimension already tracked by #1238. This issue is the **additional** search/sort-keystroke recompute of the immutable fields, which persists even after the `selected` extraction; a proper split addresses both, so the two are best fixed together.

Found during a general codebase health review.

ginderjeremiah:
## Summary

Closes #1270.

`CodexView` rebuilt heavy, immutable row view-models every time cheap, frequently-changing interaction state moved:

- **`enemyRows`** mapped the whole filtered catalogue — `levelRange`, the per-boss `zones.find` (O(enemies×zones)) and the `searchText` join/lowercase — on **every search keystroke, sort and filter change**, even though none of those projected fields depend on search/sort.
- **`zoneRows`** recomputed the O(zones×enemies) `spawnCount` alongside each zone's live `status`, which flips **continuously during idle play** (zones clear, gates open).

This splits each derived in two:

- a stable **projection** derived from reference data only (`enemyProjections`, `zoneProjections`) — the immutable per-row fields built once and memoised until reference data changes;
- a thin **row** derived that layers the interaction state on top — `filter`/`search`/`sort` for enemies, the cleared/locked `status` overlay for zones.

`filteredEnemies` is folded into the new `enemyRows` and removed (it had no other consumer). `ZoneRowVM` now extends a new `ZoneProjectionVM` (immutable fields) with the `status` overlay.

## Note on the `selected` field

The issue body shows `enemyRows` baking in a `selected: e.id === this.selectedEnemyId` field. That field is **already gone** — it was extracted by #1238 (which the issue acknowledges as the related selection-recompute concern). So this PR only had to address the remaining search/sort-keystroke recompute, which is what the split does.

## Note on `skillRows` (deliberately not split)

The issue asks to "apply the same split to the zone/skill row deriveds." I split zones but **left `skillRows` as-is on purpose**: it depends solely on reference data (`skillsCatalogue` + `enemies`) — the Skills tab has no search/sort and the row carries no live overlay (selection is read separately, kept out of the projection). So `skillRows` is already effectively computed once per reference-data load; a projection split would just add an identity derived with no runtime benefit and more code. Happy to add it for symmetry if you'd prefer, but it seemed like dead-weight to introduce now.

## Test plan

- [x] `npm run check` — 0 errors / 0 warnings
- [x] `eslint` on the changed files — clean
- [x] `vitest` codex suite — 105 passed (incl. 3 new tests):
  - `enemyProjections` projects every enemy independent of filter/search/sort
  - `zoneProjections` projects the immutable fields without `status`, in authored order
  - `zoneProjections` stays referentially stable when a zone is cleared / its gate opens
- [x] All existing `enemyRows` / `zoneRows` / `skillRows` behaviour tests still pass unchanged (output contract preserved)

---
_Generated by [Claude Code](https://claude.ai/code/session_01E3dfQ4tgcojwbyS2ux3r3B)_

### DIFF
--- a/UI/src/routes/game/screens/codex/codex-view.svelte.ts
+++ b/UI/src/routes/game/screens/codex/codex-view.svelte.ts
@@ -135,14 +135,18 @@ export interface EnemyChallengeVM {
 	completed: boolean;
 }
 
-export interface ZoneRowVM {
+/** The immutable per-zone fields (everything the rail row needs except the live progression status). */
+export interface ZoneProjectionVM {
 	id: number;
 	name: string;
 	band: string;
-	status: ZoneStatus;
 	/** Enemies (excluding the boss) that spawn in this zone. */
 	spawnCount: number;
 	hasBoss: boolean;
+}
+
+export interface ZoneRowVM extends ZoneProjectionVM {
+	status: ZoneStatus;
 }
 
 export interface ZoneBossVM {
@@ -311,10 +315,6 @@ export class CodexView {
 
 	readonly enemies = $derived(liveEnemies(staticData.enemies));
 
-	readonly filteredEnemies = $derived(
-		this.enemies.filter((e) => this.filter === 'all' || (this.filter === 'boss' ? e.isBoss : !e.isBoss))
-	);
-
 	/** Top-level tabs with live counts + section accents. */
 	readonly tabs = $derived.by<CodexTabVM[]>(() => {
 		const counts: Record<CodexTab, number> = {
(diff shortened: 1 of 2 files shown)


ginderjeremiah:
## Review — Approve with a nit

Verified every claim against the branch (`claude/wonderful-dirac-dz7czs`). The split is correct and faithful to #1270. CI is fully green (frontend, e2e on all three browsers, `all-checks-passed`).

### What I verified

- **Projection memoization is correct.** `enemyProjections` / `zoneProjections` read only reference-data deriveds (`this.enemies` / `this.zones` / `staticData.zones`) — none of which depend on `search` / `sort` / `filter` or the zone `status` inputs. Under Svelte 5 `$derived` semantics those projections therefore won't recompute when interaction state moves, which is exactly the recompute #1270 was about. (`codex-view.svelte.ts:337`, `:501`)
- **Row layering preserves behavior.** `enemyRows` applies the same `filter → matchesEnemySearch → sortEnemyRows` chain as before (`:362`); `zoneRows` overlays `status` by index zip (`:515`). The zip is safe — `zoneProjections` and `zones` both derive from `this.zones`, so identical order and length.
- **`filteredEnemies` removal is clean** — no dangling references on the branch; the filter predicate moved inline into `enemyRows`.
- **Not splitting `skillRows` is justified.** It reads only `skillsCatalogue` + `enemies` (reference data) and carries no live/interaction overlay — selection is read separately via `selectedSkillId`. A projection split would be an identity derived with no runtime benefit. Agree with leaving it.
- **`selected` field note checks out** — the projection carries no `selected` field; `selectedEnemy` is a separate derived (extracted by #1238). PR note is accurate.
- **No dead code; theming/DRY respected** — accents still go through the `rarityColor` helper, no hard-coded colours, comments concise.

### Findings

**Blocking:** none.

**Should-fix:** none.

**Nit (deferred to a follow-up, non-blocking):**
- The projection stability test (`codex-view.test.ts`, `zoneProjections` *"stays stable when a zone is cleared or its gate opens"*) compares **two fresh `CodexView` instances** by value (`toEqual`). That proves value-independence but not the memoization itself — a single instance's projection not *re-running* (same reference) when interaction state mutates. The runtime guarantee the split exists to provide isn't asserted. Same gap for `enemyProjections`. Severity low (memoization is correct by construction); tracked in **#1288** to strengthen the regression guard with a single-instance `toBe` assertion across an interaction-state mutation.

Verdict: **Approve.** The one nit is captured as a follow-up (#1288), so merging.

— Review by Claude

---

PR #1286, opened 2026-06-26
Document direct-2-cycle-only guarantee of the derived-modifier circular guard (#1271)

### LINKED ISSUE #1271: AttributeCollection circular-derived detection only catches direct 2-cycles, not its advertised general contract
## Problem

`AttributeCollection.AddModifierWithoutCacheInvalidation` (`Game.Core/Attributes/AttributeCollection.cs:118-127`) guards against circular derived modifiers by throwing `AttributeCircularDerivedModifierException` only when `node.DerivedNodes?.Contains(sourceNode) == true`:

```csharp
if (modifier.Source is EAttributeModifierSource.Derived)
{
    var sourceNode = GetOrCreateNode((int)modifier.DerivedSource);
    if (node.DerivedNodes?.Contains(sourceNode) == true)   // only the direct A<->B cycle
    {
        throw new AttributeCircularDerivedModifierException(modifier);
    }
    sourceNode.GetOrCreateDerivedNodes().Add(node);
}
```

This catches only a **length-2 cycle** (A derives from B while B already derives from A). A longer derived chain (A→B→C→A) is not detected and would recurse unbounded in `GetAttributeValue`/`SetCachedValue`, ending in a `StackOverflow`. The exception type's name advertises general circular-dependency protection it does not actually provide.

## Why it matters

Today derived modifiers come **only** from the hardcoded, acyclic `StaticAttributeModifiers`, so a cycle is unreachable in practice — this is not a live bug. But it's a latent crash vector the moment derived modifiers become authorable (e.g. gear/skill-effect-sourced derived attributes), and the class currently over-promises its guarantees.

## Suggested fix

- narrow the exception/comment to document the actual guarantee (only direct cycles are detected, justified by static-only acyclic derived modifiers), so the limitation is explicit; or

Given derived modifiers are static-only today, the documentation is likely sufficient until authorable derived modifiers are on the roadmap.

Found during a general codebase health review.

ginderjeremiah:
## Summary

Closes #1271.

`AttributeCircularDerivedModifierException` and the guard in
`AttributeCollection.AddModifierWithoutCacheInvalidation` (`Game.Core/Attributes/AttributeCollection.cs`)
advertised general circular-dependency protection, but the check —
`node.DerivedNodes?.Contains(sourceNode) == true` — only catches a **direct (length-2)** cycle
(A derives from B while the new modifier makes B derive back from A). A longer derived chain
(A→B→C→A) is not detected and would recurse unbounded in `GetAttributeValue`/`SetCachedValue`,
ending in a stack overflow.

As the issue notes, this is **not a live bug**: the only derived modifiers today come from the
hardcoded, acyclic `StaticAttributeModifiers`, so a longer cycle is unreachable in practice. The
problem is purely that the names/comments over-promise a guarantee the code doesn't provide.

Per the issue's recommendation (documentation is sufficient until authorable derived modifiers are
on the roadmap — implementing full cycle detection now would be premature), this PR narrows the docs
to state the **actual** guarantee:

- **Exception XML doc** — describes the direct-2-cycle scope and adds a `<remarks>` covering the
  justification (static-only acyclic derived modifiers) and the upgrade trigger (if derived
  modifiers become authorable, this must become a full cycle check or a longer cycle stack-overflows).
- **Inline guard comment** — states that only a direct cycle is detected, why that's acceptable, and
  points to the exception for the full rationale.

No behavior change.

## Test plan

- [x] `dotnet build Game.Core/Game.Core.csproj` — clean, 0 warnings (the new `<see cref>` references resolve)
- [x] `dotnet test Game.Core.Tests` — 843 passed, including the existing direct-cycle throw test (`Constructor_DuplicateDerivedModifierSameSourceTarget_Throws`)

No new tests were added: this is a doc-only change with no new domain logic, and the existing test
already pins the detected (direct) cycle behavior.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YXVxiynD9ax6eExcNnjt8M)_

### DIFF
--- a/Game.Core/Attributes/AttributeCircularDerivedModifierException.cs
+++ b/Game.Core/Attributes/AttributeCircularDerivedModifierException.cs
@@ -3,8 +3,18 @@
 namespace Game.Core.Attributes
 {
     /// <summary>
-    /// An exception that is thrown when adding an <see cref="AttributeModifier"/> that causes a circular dependency.
+    /// Thrown when adding a derived <see cref="AttributeModifier"/> would create a <b>direct</b>
+    /// (length-2) cycle between two attributes — the target already derives from the modifier's
+    /// source while the new modifier makes that source derive back from the target.
     /// </summary>
+    /// <remarks>
+    /// This guard intentionally detects only direct 2-cycles, not longer derived chains
+    /// (e.g. A→B→C→A). That is sufficient because the only derived modifiers today come from the
+    /// static, acyclic <see cref="StaticAttributeModifiers"/>, so a longer cycle is unreachable in
+    /// practice. If derived modifiers ever become authorable (e.g. gear- or skill-effect-sourced),
+    /// this must be upgraded to a full cycle check — an undetected longer cycle would otherwise
+    /// recurse unbounded during attribute evaluation and end in a stack overflow.
+    /// </remarks>
     public class AttributeCircularDerivedModifierException : Exception
     {
         /// <summary>
--- a/Game.Core/Attributes/AttributeCollection.cs
+++ b/Game.Core/Attributes/AttributeCollection.cs
@@ -118,6 +118,11 @@ namespace Game.Core.Attributes
             if (modifier.Source is EAttributeModifierSource.Derived)
             {
                 var sourceNode = GetOrCreateNode((int)modifier.DerivedSource);
+                // Detects only a direct (length-2) cycle: the target already derives from this
+                // modifier's source. Longer derived chains (A→B→C→A) are not detected — acceptable
+                // only because derived modifiers come solely from the static, acyclic
+                // StaticAttributeModifiers. See AttributeCircularDerivedModifierException for the
+                // full rationale and the upgrade trigger if derived modifiers become authorable.
                 if (node.DerivedNodes?.Contains(sourceNode) == true)
                 {
                     throw new AttributeCircularDerivedModifierException(modifier);


ginderjeremiah:
## Review: Approve

Doc-only change that narrows `AttributeCircularDerivedModifierException` and the guard comment in `AttributeCollection.AddModifierWithoutCacheInvalidation` to state the *actual* guarantee. I verified the claims against the code rather than taking the description at face value.

### Verification
- **Guard genuinely catches only a direct (length-2) cycle.** `node.DerivedNodes?.Contains(sourceNode) == true` tests only whether the target already *directly* derives from this modifier's source; a longer chain (A→B→C→A) is not in `node.DerivedNodes` directly and slips through, recursing unbounded in `GetAttributeValue`/`SetCachedValue` → stack overflow. The new docs describe this precisely — no over- or under-claim.
- **The static-only justification holds.** The only production-code uses of `EAttributeModifierSource.Derived` are the consuming logic (`AttributeCollection`, `AttributeModifier`), the codegen writer, and the source itself, `StaticAttributeModifiers`. Every other hit is a test or the generated frontend mirror. The authored static set is acyclic (derived targets all derive from non-derived core attributes), so a longer cycle is genuinely unreachable today.
- **`<see cref="StaticAttributeModifiers"/>` resolves** — same `Game.Core.Attributes.Modifiers` family, and the PR reports a clean build with 0 warnings.
- **Behavior-neutral** — only XML doc + one inline comment added; the guard logic is untouched. Existing direct-cycle throw test still pins the detected behavior.
- **CI green** — `all-checks-passed`, backend test/verify, and all e2e (chromium/firefox/webkit) succeeded; frontend jobs correctly skipped (no FE changes).

### Findings
- **Blocking:** none
- **Should-fix:** none
- **Nits:** none. The new comment/remarks run a few lines, but they encode a real latent-crash rationale plus a concrete upgrade trigger — appropriate, not verbose, per the project's "concise but no hard one-line cap" guidance.

This is exactly the documentation-over-premature-implementation path #1271 recommended, and it correctly leaves the full-cycle-detection work for when derived modifiers become authorable.

*Review by Claude (Opus 4.8)*

---

PR #1284, opened 2026-06-26
Admin Workbench editor for paths &amp; proficiencies (#1132)

ginderjeremiah:
## Summary

Implements the Admin Workbench editor for the proficiency/path progression system (#1132), faithful to the `docs/designs/admin/proficiency-admin` mockup. Because the design authors **two cross-referenced reference catalogues** (paths and their proficiency tiers) through one master→detail with a path→tier drill-down, it doesn't fit the generic single-entity `EntityConfig`. Following the established escape hatch, it's a **bespoke surface that still lives inside the Progression Workbench group** (the new **Paths** tool), branched into the admin page like the Ops tools but reusing the diff/save helpers, the reference store, and the change-state tokens/primitives.

The deferred dependencies all shipped: backend foundation (#1114), Path rework (#1160), and conlang fields (#1215).

## What's included

**Frontend — `UI/src/routes/admin/workbench/progression/`**
- **`ProgressionStore`** controller: manages both catalogues with baseline diff tracking, selection/drill/tab state, tier reorder, retire/reinstate, and a **combined save** that persists paths first, resolves the new path ids, then the proficiencies that FK to them, then each child collection (contributions → modifiers → rewards → cross-path prerequisites). Partial-failure re-seed semantics mirror `EntityStore`.
- **Path editor** — Identity (name, falloff, description); **Tiers** spine with accessible ↑/↓ reorder + add-tier + drill; **Contributions** (skill / home-tier / weight) with a live falloff preview.
- **Proficiency (tier) editor** — Identity · Conlang with **words-of-power decipher preview** (undeciphered → pronunciation at `ceil(maxLevel/2)` → translation at `maxLevel`); **XP Curve** with a derived cost chart; **Milestones** level timeline + per-level payout authoring (attribute modifiers + optional reward skill); **Gateways** (seed skill + cross-path prerequisites, root tier only).
- Reference store loads `GetPaths`/`GetProficiencies` and exposes the pickers (skill, player-flagged seed/reward skill, prerequisite proficiency, home-tier).
- Registered as the **Paths** tool in the Progression nav group.

**Backend — `AdminProficiencies.SaveProficiencies`**
- Pre-checks `(PathId, PathOrdinal)` uniqueness against the **full prospective per-path tier layout** (not a per-change cache probe, which would false-reject a valid reorder/swap), returning a named `AdminSaveResult.Failure` instead of a raw DB unique-violation 500. The DB unique index stays the backstop. (Issue comment.)

## Design decisions
[...]

PR #1293, opened 2026-06-26
Add cross-path Map view to the admin progression editor (#1285)

### LINKED ISSUE #1285: Admin progression editor: cross-path Map view
Follow-up to #1132 / #1284 (the Admin Workbench editor for paths & proficiencies).

The progression editor (`UI/src/routes/admin/workbench/progression`) shipped the **List** authoring surface (path list → tier drill-down, all tabs). The mockup (`docs/designs/admin/proficiency-admin/Proficiency Workbench - Hifi.dc.html`) also includes a **Map view**: a cross-path DAG visualization that lays each path out as a column of tier nodes and draws **SVG bezier gateway edges** between cross-path prerequisites, with a **List/Map toggle** in the page header and click-through from a node to its tier in the List editor.

This was deferred because it is pure navigation/visualization — the authoring tool is fully functional without it (gateways are authored in the tier **Gateways** tab).

### Scope
- Add a List/Map segmented toggle to the `Progression.svelte` header.
- A `ProgressionMap.svelte` rendering paths as columns / tiers as nodes, with measured SVG edges for `prerequisiteIds` (see the mockup's `measureEdges`/`mapEdgeDefs` for the approach), gated/starter node styling, and click-to-open into the List editor (reuse `ProgressionStore.selectPath`/`drillTier`).
- It reads from the same `ProgressionStore` (no new data plumbing).

### Notes
- Keep it read/navigate only — all editing stays in the List surface.
- Accessibility: nodes are real `<button>`s; the SVG edges are decorative (`aria-hidden`).

Labels: this is additive UI with no backend work.

ginderjeremiah:
## Summary

Implements the deferred **Map view** for the Progression workbench tool (the **Paths** surface), faithful to the `docs/designs/admin/proficiency-admin` mockup. The List authoring surface shipped in #1132 / #1284; this adds the cross-path DAG visualization the mockup also describes.

- **List/Map segmented toggle** in the `Progression.svelte` header (mirrors the mockup's segmented control; `--text-on-accent` on the active segment).
- **`ProgressionMap.svelte`** — each path is a column of its tiers (ascending by ordinal), tiers are real `<button>` nodes, and cross-path gateway prerequisites are drawn as **measured SVG bezier edges** between the prerequisite node's right edge and the gated node's left edge. Gated tiers get the dashed-accent treatment + a `· gated ✦` mark; the within-path order shows as solid connectors; retired records render muted.
- **Click-through into the List editor** — a node calls `ProgressionStore.selectPath` + `drillTier`; a column header calls `selectPath` — then flips the surface back to List. No editing happens in the Map (navigation/visualization only).
- It reads from the same `ProgressionStore` — **no new data plumbing and no backend work**.

## How it addresses the issue

All three asks from #1285 are covered: the List/Map toggle in the header, the `ProgressionMap.svelte` rendering paths-as-columns / tiers-as-nodes with measured `prerequisiteIds` edges, gated/starter node styling, and click-to-open into the List editor reusing the existing store navigation.

## Design notes

- **Edge logic is extracted as pure functions** (`progression-map.ts`: `mapColumns` / `mapEdgeDefs` / `bezierPath`) so the DAG projection and path geometry are unit-testable — jsdom does no layout, so the measurement itself can't be asserted via the DOM, but everything that decides *what* to draw can.
- **Measurement** runs in an `$effect` (re-measuring after the columns change, via `tick()`) and on container resize via a `ResizeObserver`, querying the laid-out `[data-node]` elements — the mockup's `measureEdges`/`mapEdgeDefs` approach, adapted to Svelte 5.
- **Accessibility**: nodes and the column headers are real `<button>`s; the SVG edge layer is decorative (`aria-hidden="true"`), per the issue's note.
- A prerequisite that references an absent tier still yields an edge def; the measurer drops it when the node element can't be found (matches the mockup), so an in-progress authoring state can't throw.
- No `docs` change: the progression editor and the entity-driven Workbench are already documented in `frontend.md`; the Map view is an additive sub-surface of that already-documented tool, not a new cross-cutting pattern.

## Test plan

- [x] `vitest` progression suite — `progression-map.ts` helpers (columns ordering, gated/retired flags, edge defs incl. multi-prereq and absent-tier, bezier geometry) and a `ProgressionMap.svelte` component test (column/node rendering, `data-node` ids + connectors, gated styling, node→drill and header→open navigation, aria-hidden edge layer, empty states).
- [x] Full unit suite — **2740 passed**, coverage floors (incl. `src/routes/**`) hold.

Closes #1285

---
_Generated by [Claude Code](https://claude.ai/code/session_018pRFsnBcAY1RzhNRLVV5DV)_


UI/src/routes/admin/workbench/progression/ProgressionMap.svelte
```diff
@@ -0,0 +1,248 @@
+<div class="map" bind:this={container} data-testid="progression-map">
+	<svg class="edges" aria-hidden="true" style="width:{svgW}px;height:{svgH}px">
+		<defs>
+			<marker id="prog-gateway-arrow" markerWidth="8" markerHeight="8" refX="6" refY="3" orient="auto">
+				<path d="M0,0 L6,3 L0,6 Z" fill="var(--accent)" />
+			</marker>
+		</defs>
+		{#each edges as d (d)}
```
ginderjeremiah: **Should-fix (low-likelihood crash):** this keyed block uses the geometry string itself as the key. `mapEdgeDefs` emits one edge per `prerequisiteId` with no dedup, so if a proficiency's `prerequisiteIds` ever carries a duplicate id, two edges measure to the same `d` and Svelte throws `each_key_duplicate`, taking down the whole Map view. The authoring path (`addPrerequisite`) dedupes, but nothing guarantees loaded backend data does. Since `edges` is fully recomputed on every measure, keying on the index is strictly safer with no downside:

```svelte
{#each edges as d, i (i)}
```

---

ginderjeremiah (commented):
## Review: Cross-path Map view for the progression editor (#1285)

**Verdict: Approve** — clean, faithful to the issue and mockup, and well-tested; no blocking findings. _(Submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author — treat this as an approve.)_

The change delivers all three asks from #1285: the List/Map header toggle, `ProgressionMap.svelte` rendering paths-as-columns / tiers-as-nodes with measured `prerequisiteIds` edges and gated/starter styling, and click-through into the List editor reusing `selectPath`/`drillTier`. The DAG projection and geometry are extracted as pure, unit-tested functions; the component reads straight off `ProgressionStore` with no new data plumbing, matching the bespoke-progression-surface pattern documented in `frontend.md`.

### Should-fix
- **Keyed `{#each edges as d (d)}` on the geometry string** — `ProgressionMap.svelte:8`. Identical `d` values (e.g. a duplicate id in a proficiency's `prerequisiteIds`, which `mapEdgeDefs` doesn't dedup) would throw `each_key_duplicate` and crash the Map. Key on the index instead — see inline comment.
- **The new List/Map toggle is untested.** `Progression.svelte` gained interactive UI (the `view` state + surface swap, and the `onNavigate` callback that flips back to List), but there's no `Progression.test.ts` — the existing tests only cover the store and helpers. `frontend.md` asks for unit tests on touched components with UI interaction. A small test asserting that clicking **Map** renders `ProgressionMap`, **List** renders `ProgressionList`, and that a node/header navigation flips the surface back would close the gap. Not blocking (the substantive logic — `ProgressionMap` + the pure helpers — is thoroughly covered, and the toggle is trivial glue over the tested store), but worth adding.

### Nits
- **"Cross-path" is slightly loose.** `GatewaysEditor` lets an author pick *any* proficiency as a prerequisite (it only excludes self and existing picks), so `prerequisiteIds` aren't guaranteed cross-path. A within-path prereq would draw a bezier looping back into the same column. This is pre-existing data-model looseness (the Gateways tab calls them "cross-path gateways" too), not introduced here — fine to leave, or tighten both surfaces in a follow-up.
- `color-mix(in srgb, var(--accent) 7%, transparent)` for the gated-node tint is theme-overridable and matches existing usage (`GatewaysEditor`, the save-bar buttons), so it's consistent — but `frontend.md` names `tintColor(color, alpha)` as the canonical helper for this. Take it or leave it.

### Test coverage
- **Well covered:** `progression-map.ts` helpers (column ordering/independence, gated + retired flags, edge defs incl. multi-prereq and absent-tier, bezier geometry) and a `ProgressionMap.svelte` component test (column/node rendering, `data-node` ids + connectors, gated node/header styling, node→drill and header→open navigation with `onNavigate`, aria-hidden edge layer, both empty states).
- **Untested (acceptable):** the actual edge measurement — jsdom does no layout, and the PR correctly notes this; `bezierPath` covers the geometry in isolation.
- **Gap:** the `Progression.svelte` toggle (see Should-fix).

### What's good
- Pure functions (`mapColumns` / `mapEdgeDefs` / `bezierPath` / `tierNodeId`) cleanly separate the testable DAG projection from the DOM measurement — exactly the right seam for a layout-dependent view.
- Accessibility is handled per `frontend.md`: nodes and column headers are real `<button>`s, the SVG layer is `aria-hidden`, and the segmented toggle uses `role="group"` + `aria-pressed`.
- All colours go through theme tokens; retired paths/tiers render muted rather than being dropped, consistent with the "never silently drop a retired reference" convention.
- Graceful handling of an in-progress authoring state: a prerequisite to an absent tier still yields an edge def and is dropped at measure time rather than throwing.


---

ginderjeremiah:
Addressed the keyed-each should-fix (now keyed on index). On the toggle-test should-fix: I added one, but it broke the `src/routes/**` coverage gate and I've since removed it — flagging the tradeoff rather than forcing it.

A `Progression.svelte` component test has to render the full surface, which transitively imports the untested detail-pane subtree (`PathDetail`/`TierDetail` → the gateways/milestones/xp-curve editors). Since `coverage.all` is false, that pulls those low-coverage files into the routes aggregate and drops it from 93.5% → 91%, under the 93% floor. Isolating the toggle would mean stubbing all four child components just to instrument the trivial glue.

Given the reviewer flagged this as non-blocking and `ProgressionMap.test.ts` already covers the only non-trivial behavior (the `onNavigate` flip-back on node/header click), I opted to leave the glue untested rather than regress the gate. If we want it covered later, the clean path is lifting the detail-pane components' own coverage first (worth a follow-up), after which a full-surface toggle test stops dragging the aggregate.

---

PR #1307, opened 2026-06-27
Standardize admin RewardPicker overlay on Popover (#1265)

### LINKED ISSUE #1265: Standardize the admin RewardPicker inline overlay on Popover
Follow-up split out of #1239 (PR #1264).

## Problem

`UI/src/routes/admin/workbench/components/challenge/RewardPicker.svelte` (opened from `ChallengeRewardSection.svelte`) is a dismissable overlay with **no Escape handling and no focus management** — it's an in-flow panel that expands below the reward slots, dismissed only via its close button or by re-toggling the slot. The app otherwise standardizes dismissable overlays on `components/Popover.svelte`, which owns backdrop/Escape dismissal, the shared `focusTrap` (Tab-trap + focus capture/restore), and a scroll lock.

## Why it wasn't folded into #1239

`Popover` renders a **centered backdrop dialog** over its nearest positioned ancestor. The RewardPicker today renders **inline** in the section flow (the list expands in place under the two reward slots). Moving it onto `Popover` relocates it to a centered modal-style overlay — a **UX change**, not a pure refactor — so it deserves its own decision:

- Keep it inline but add Escape + focus management (a lighter-weight, in-flow focus-trap usage of the shared `focusTrap` action), **or**
- Adopt `Popover` fully and accept the centered-overlay presentation.

## Notes

- The shared `focusTrap` action (`components/focus-trap.ts`) landed in #1264 and is reusable for the lighter-weight option.
- Same pattern likely applies to any other inline admin pickers; worth a quick scan when picked up.

Surfaced during #1239.

ginderjeremiah:
## Summary

Moves the admin challenge **reward picker** (`RewardPicker.svelte`, opened from `ChallengeRewardSection.svelte`) from a hand-rolled in-flow overlay — dismissable only via its close button or by re-toggling the slot, with **no Escape handling and no focus management** — onto the shared `components/Popover.svelte` primitive. It now gets backdrop/Escape dismissal, the shared `focusTrap` (Tab-trap + focus capture/restore), and a body scroll lock for free, closing the accessibility gap described in #1265.

Per the owner's decision on the issue (comment: _"Let's go with option 2 for now and fully adopt `Popover`"_), this is the **full `Popover` adoption** — relocating the picker from an in-flow panel below the reward slots to a centered overlay — rather than the lighter-weight in-flow `focusTrap` option.

## How it addresses the issue

- **`ChallengeRewardSection.svelte`** now wraps the conditional `RewardPicker` in a single `<Popover open={open !== null} onClose={…} label={…}>`. The `open` state machine (`'item' | 'mod' | null`) is unchanged, so toggling a slot still opens/closes the right picker and a pick/record-switch still collapses it.
- **`workbench.scss`** — `.ch-picker` swaps its in-flow `margin-top: 16px` for a fixed centered-modal width (`width: 720px; max-width: 100%`); the border/shadow card styling already suited an overlay.
- **No extra positioning wrapper needed.** `Popover` overlays its nearest positioned ancestor, which here is the existing `.admin-shell` (`position: relative; overflow: hidden`) — every intermediate container (`.detail-pane`, `.detail-body`, `.workbench`, `.admin-workspace`) is unpositioned, so the picker centers over the full admin workspace and `.detail-body`'s `overflow-y: auto` doesn't clip it.

## Scan for other inline admin pickers

The issue noted this pattern may apply elsewhere. I scanned the admin tree: `RewardPicker` was the only hand-rolled dismissable overlay. `WorkbenchDetail` and `DeadLetterConsole` already use the modal helpers, and `ScopeField` is a plain select field, not an overlay. Nothing else needs converting.

## Test plan

- [x] `npm run test:unit:ci` — full suite green (2743 tests).
- [x] Added integration tests to `ChallengeRewardSection.test.ts` covering the two new Popover-owned dismissal paths (Escape closes the picker without mutating the reward; backdrop click closes it). Existing picker open/pick/clear/record-switch tests continue to pass unchanged.

Closes #1265

---
_Generated by [Claude Code](https://claude.ai/code/session_01S83q39142zrEBCqApXLdGw)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/challenge/ChallengeRewardSection.svelte
+++ b/UI/src/routes/admin/workbench/components/challenge/ChallengeRewardSection.svelte
@@ -46,25 +46,31 @@
 		/>
 	</div>
 
-	{#if open === 'item'}
-		<RewardPicker
-			kind="item"
-			records={itemPickRecords}
-			currentId={challenge.rewardItemId}
-			claimed={claimedItems}
-			onPick={setItem}
-			onClose={() => (open = null)}
-		/>
-	{:else if open === 'mod'}
-		<RewardPicker
-			kind="mod"
-			records={modPickRecords}
-			currentId={challenge.rewardItemModId}
-			claimed={claimedMods}
-			onPick={setMod}
-			onClose={() => (open = null)}
-		/>
-	{/if}
+	<Popover
+		open={open !== null}
+		onClose={() => (open = null)}
+		label={open === 'mod' ? 'Select item mod reward' : 'Select item reward'}
+	>
+		{#if open === 'item'}
+			<RewardPicker
+				kind="item"
+				records={itemPickRecords}
+				currentId={challenge.rewardItemId}
+				claimed={claimedItems}
+				onPick={setItem}
+				onClose={() => (open = null)}
+			/>
+		{:else if open === 'mod'}
+			<RewardPicker
+				kind="mod"
+				records={modPickRecords}
+				currentId={challenge.rewardItemModId}
+				claimed={claimedMods}
+				onPick={setMod}
+				onClose={() => (open = null)}
+			/>
+		{/if}
+	</Popover>
 
 	{#if none}
 		<div class="ch-reward-warn">
@@ -76,6 +82,7 @@
 
 <script lang="ts">
 import { ERarity, type IChallenge } from '$lib/api';
+import { Popover } from '$components';
 import { reference } from '../../reference.svelte';
 import type { EntityStore } from '../../entity-store.svelte';
 import type { Identified } from '../../entities/types';
--- a/UI/src/routes/admin/workbench/workbench.scss
+++ b/UI/src/routes/admin/workbench/workbench.scss
@@ -1085,12 +1085,14 @@
 		margin-top: 5px;
 		font-style: italic;
 	}
+	// Centered overlay card (mounted in a Popover), so it owns a fixed width rather
+	// than the in-flow spacing it previously needed below the reward slots.
 	.ch-picker {
-		margin-top: 16px;
+		width: 720px;
+		max-width: 100%;
 		border: 1px solid color-mix(in srgb, var(--accent) 30%, transparent);
 		border-radius: 8px;
 		overflow: hidden;
-		max-width: 760px;
 		box-shadow: 0 12px 36px color-mix(in srgb, var(--black) 35%, transparent);
 	}
 	.ch-picker-bar {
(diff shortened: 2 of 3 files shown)


ginderjeremiah:
## Code review — Verdict: approve ✅

Reviewed against `CLAUDE.md` and `docs/frontend.md` (overlay systems, accessibility, styling/theming, testing rules).

### Summary
Clean, well-scoped refactor that fully adopts the shared `Popover` primitive for the admin challenge `RewardPicker`, exactly per the owner's decision on #1265 (_"option 2 ... fully adopt Popover"_). It closes the stated a11y gap (Escape + focus trap + capture/restore + scroll lock) by delegating to the standardized overlay rather than hand-rolling, which is precisely the pattern the docs prescribe ("Prefer these over hand-rolling a backdrop").

### Verified claims
- **Popover positioning is correct.** `Popover` is `position: absolute; inset: 0` and overlays its nearest *positioned* ancestor. I traced the DOM chain — `.body-inner` / `.detail-body` / `.detail-pane` (`WorkbenchDetail.svelte`) and `.admin-workspace` (`+page.svelte`) are all unpositioned; `.admin-shell` is `position: relative; overflow: hidden`. So the picker centers over the full workspace, and `.detail-body`'s `overflow-y: auto` does not clip it (the layer anchors to `.admin-shell`, escaping the scroll container). As described.
- **Scan for other inline pickers holds up.** `RewardPicker` was the only hand-rolled dismissable overlay in the admin tree; `WorkbenchDetail`/`DeadLetterConsole` use the modal helpers and `ScopeField` is a plain select. Nothing else to convert.
- **Tests cover the new paths.** The two added integration tests exercise the Popover-owned dismissals (Escape and backdrop click) and assert the reward is *not* mutated on dismissal — the correct semantic. Existing open/pick/clear/record-switch tests are untouched and still pass.
- **`Popover` is exported from the `$components` barrel** (used by the import). Import wiring is correct.

### Quality
- Theming preserved — `.ch-picker` swaps in-flow `margin-top` for a fixed centered-modal width using CSS variables / `color-mix`; no hard-coded colours introduced.
- No new null-forgiving (`!`) operators; braces conventions respected; the one added scss comment is appropriately brief.
- a11y is strictly improved: `Popover` adds `role="dialog"`, `aria-modal`, focus trap, and reduced-motion-friendly chrome where the old in-flow panel had none.

### Minor / non-blocking observations (no action needed)
- `RewardPicker`'s `<input autofocus>` is now redundant with `Popover`'s on-open focus `$effect` (both target the search input). Not a conflict — `Popover`'s explicit focus is actually more reliable than `autofocus` across conditional re-renders — so leaving it is fine.

### CI
All required checks green: `all-checks-passed` ✅, `test-frontend` ✅, `verify-frontend` ✅, `test-e2e` (chromium/firefox/webkit) ✅. Backend jobs correctly skipped (frontend-only change).

No nits or should-fixes. Merging via squash.

---

PR #1310, opened 2026-06-27
Collect dictionary key-type imports in CodeGen (#1304)

### LINKED ISSUE #1304: CodeGen omits the import for a dictionary's key type (latent FE build break)
## Problem

The TypeScript client codegen emits a `Record<key, value>` for a dictionary wire model but only imports the **value** type, never the **key** type.

In `Game.Api/CodeGen/CodeGenTypeFormatter.cs`:

- `GetTypeText` (the dictionary branch, ~lines 71-77) renders `Record<keyText, valueText>`, where `keyText = GetTypeText(GenericArgumentDescriptors[0])` — so a key that is an enum (or any type needing an interface) emits that type's **name** into the generated file.
- `GetImportText` (the dictionary branch, ~lines 20-22) returns `GetImportText(descriptor.GenericArgumentDescriptors[1])` — the value descriptor **only**. The key descriptor (`[0]`) is never walked, so its import is never collected.

If a wire model ever uses a `Dictionary<SomeEnum, X>` (or a dict keyed by a type that `NeedsInterface()`), the generated file would reference `SomeEnum` without importing it, breaking the frontend build. The safety `throw` in `GetTypeText` for unmapped types would **not** fire here, because an enum key *is* mappable — it just isn't imported.

## Why it's latent (not an active defect)

No current wire model uses a `Dictionary<,>` — `grep` over `Game.Api/Models` and `Game.Abstractions/Contracts` finds none (the only dictionaries are infrastructure-internal). So nothing is broken today; this is a trap that would surface silently the first time someone adds a dictionary-with-non-primitive-key to the wire surface.

## Suggested fix

In `GetImportText`'s dictionary branch, collect the imports for **both** generic argument descriptors (key and value), e.g. concatenating `GetImportText([0])` and `GetImportText([1])` (the caller already `Distinct()`s and filters nulls, so a primitive key contributing `null` is harmless). A small unit test with a `Dictionary<SomeEnum, X>` wire model would pin it.

ginderjeremiah:
## Summary

The TypeScript client codegen renders a `Record<K, V>` for a dictionary wire model but only ever collected the **value** type's import, never the **key**'s. A dictionary keyed by an enum (or any type that `NeedsInterface()`) would emit a `Record<SomeEnum, …>` referencing `SomeEnum` without importing it, silently breaking the frontend build. This is **latent** today — no wire model uses a `Dictionary<,>` — so nothing is broken now; this closes the trap.

Closes #1304.

## How it addresses the issue

The issue suggested concatenating `GetImportText([0])` and `GetImportText([1])` in the dictionary branch. That can't be done literally: the single-descriptor `GetImportText` returns one `string?` and is reused as a **key selector** by `ApiInterfaceWriter`'s `DistinctBy`/`ExceptBy`, so concatenating two import names into one string would produce a garbage dedup key (`"SomeEnumISimpleModel"`). The author's underlying intent — *"the caller already `Distinct()`s and filters nulls"* — points at the collection path, so the fix lives there:

- New **`GetImportTexts(descriptor) : IEnumerable<string>`** — the recursive import collector. It walks the enumerable element and, for a dictionary, **both** type arguments (`Concat` of key and value), returning each leaf's enum/interface import.
- The plural `GetImportText(IEnumerable<…>)` now flows through `SelectMany(GetImportTexts)` instead of `SelectNotNull(GetImportText)`, so a dictionary's key import is collected.
- The single-descriptor **`GetImportText(descriptor) : string?`** is kept unchanged as the single-identity key selector the interface writer's de-duplication relies on, with a doc comment clarifying the split.
- A shared private `GetLeafImportText` keeps the enum/interface leaf logic DRY between the two.

Because no current wire model uses `Dictionary<,>`, the generated output is **unchanged today** — for every non-dictionary descriptor the two paths produce identical results.

## Test plan

- [x] `dotnet build Game.Api.Tests` — clean (0 warnings / 0 errors)
- [x] **New unit tests** in `CodeGenTypeFormatterTests` for `GetImportTexts` (class, primitive→empty, list element, dict with primitive key+value→empty, dict with class value, **dict with enum key → key import included**, dict with enum key + class value → both), plus an end-to-end assertion through the plural `GetImportText` that an enum-keyed dictionary emits the key in the import statement. Two `Dictionary<TestEnum, …>` fixtures added to `ModelWithDictionary`.
- [x] `Game.Api.Tests` full suite — **633 passed**, 0 failed.

## Notes / follow-up

While verifying the issue I found two **parallel, entangled** latent gaps in the same dictionary-on-the-wire path that are out of scope here (fixing one in isolation surfaces the other): `Extensions.NeedsInterface` has the same key-vs-value asymmetry, and `ApiInterfaceWriter` would try to generate a bogus `IDictionary` interface for a `Dictionary<_, IFoo>` descriptor. Filed as a follow-up issue (#1311) rather than expanding this PR.

### DIFF
--- a/Game.Api.Tests/CodeGen/CodeGenTypeFormatterTests.cs
+++ b/Game.Api.Tests/CodeGen/CodeGenTypeFormatterTests.cs
@@ -288,6 +288,67 @@ namespace Game.Api.Tests.CodeGen
         }
 
         [Fact]
+        public void GetImportTexts_Class_ReturnsSingleImport()
+        {
+            var descriptor = GetPropertyDescriptor<NestedModel>("Child");
+            Assert.Equal(["ISimpleModel"], CodeGenTypeFormatter.GetImportTexts(descriptor));
+        }
+
+        [Fact]
+        public void GetImportTexts_Primitive_ReturnsEmpty()
+        {
+            var descriptor = GetPropertyDescriptor<SimpleModel>("Name");
+            Assert.Empty(CodeGenTypeFormatter.GetImportTexts(descriptor));
+        }
+
+        [Fact]
+        public void GetImportTexts_ListOfClass_ReturnsElementImport()
+        {
+            var descriptor = GetPropertyDescriptor<ModelWithList>("Items");
+            Assert.Equal(["ISimpleModel"], CodeGenTypeFormatter.GetImportTexts(descriptor));
+        }
+
+        [Fact]
+        public void GetImportTexts_DictionaryWithPrimitiveKeyAndValue_ReturnsEmpty()
+        {
+            var descriptor = GetPropertyDescriptor<ModelWithDictionary>("StringToInt");
+            Assert.Empty(CodeGenTypeFormatter.GetImportTexts(descriptor));
+        }
+
+        [Fact]
+        public void GetImportTexts_DictionaryWithClassValue_ReturnsValueImport()
+        {
+            var descriptor = GetPropertyDescriptor<ModelWithDictionary>("StringToClass");
+            Assert.Equal(["ISimpleModel"], CodeGenTypeFormatter.GetImportTexts(descriptor));
+        }
+
+        [Fact]
+        public void GetImportTexts_DictionaryWithEnumKey_IncludesKeyImport()
+        {
+            // The key type is an enum that needs importing; the value is a primitive. The old value-only
+            // collection dropped the key, emitting a Record<TestEnum, number> with no TestEnum import.
+            var descriptor = GetPropertyDescriptor<ModelWithDictionary>("EnumToInt");
+            Assert.Equal(["TestEnum"], CodeGenTypeFormatter.GetImportTexts(descriptor));
+        }
+
+        [Fact]
+        public void GetImportTexts_DictionaryWithEnumKeyAndClassValue_IncludesBothImports()
+        {
+            var descriptor = GetPropertyDescriptor<ModelWithDictionary>("EnumToClass");
+            Assert.Equal(["TestEnum", "ISimpleModel"], CodeGenTypeFormatter.GetImportTexts(descriptor));
+        }
+
+        [Fact]
+        public void GetImportText_Multiple_DictionaryWithEnumKey_ImportsKeyType()
+        {
+            // End-to-end through the import-statement formatter: the enum key must appear in the emitted
+            // import list so the generated Record<TestEnum, number> reference resolves.
+            var descriptors = new[] { GetPropertyDescriptor<ModelWithDictionary>("EnumToInt") };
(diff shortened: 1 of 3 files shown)


ginderjeremiah:
## Verdict: Approve ✅

Clean, well-scoped fix for the latent dictionary key-import omission (#1304). Reviewed the diff against the project standards and the surrounding codegen — no blocking, should-fix, or nit findings.

**Correctness**
- The `GetImportText`/`GetImportTexts` split is the right call. The single-descriptor `GetImportText(descriptor) : string?` is left behavior-unchanged (dict → value import only) and remains the stable identity key for `ApiInterfaceWriter`'s `DistinctBy`/`ExceptBy` (lines 40, 56, 60) — confirmed those call sites still pass the single-descriptor overload, so dedup keys are unaffected. The issue's literal suggestion (concatenating `GetImportText([0]) + [1]` into one string) would have corrupted that identity key; the PR correctly rejected it and moved the dual-walk to a new collection path instead.
- The new `GetImportTexts` plural collector walks key + value for dictionaries (`Concat`), which mirrors `GetTypeText`'s `Record<keyText, valueText>` rendering exactly — both arguments are named in the output, so both must be imported. Key-then-value order is consistent with the rendered type and the tests.
- `GetImportText(IEnumerable<…>)` now flows through `SelectMany(GetImportTexts)`; `Distinct().OrderBy()` are preserved, so dedup/sort of the final import list is unchanged.

**DRY / quality**
- `GetLeafImportText` cleanly factors the enum/I-prefixed-interface leaf logic shared by both paths — no duplication.
- No null-forgiving operator; the leaf null case is handled with `leaf is null ? [] : [leaf]` and naturally flattened by `SelectMany`. All control blocks use braces. No dead code (`SelectNotNull` is still used by `ApiInterfaceWriter`).
- Doc comments correctly flag the load-bearing distinction between the two overloads.

**Tests** — genuinely cover the target case: enum-key-only → `["TestEnum"]` (the exact dropped-import bug), enum-key + class-value → `["TestEnum", "ISimpleModel"]`, plus negative cases (primitive key+value → empty) that pin against over-collection, and an end-to-end assertion through the plural formatter that the emitted import statement contains `TestEnum`. Two fixtures added to `ModelWithDictionary`.

**Scope** — correctly minimal. The parallel `NeedsInterface` key asymmetry and the bogus `IDictionary`-interface gap are genuinely entangled and rightly deferred to #1311 (which exists and reads accurately) rather than expanded here.

**CI** — all required checks green (`test-backend`, `verify-backend`, e2e chromium/firefox/webkit, `all-checks-passed`). Frontend jobs skipped correctly via path filtering (no FE files touched).

Merging.

---

Issue #1301, opened 2026-06-27
Skill synthesis client: feedback (toasts + combat log)


ginderjeremiah:
Part of the skill-synthesis spike #1125 (**area F**). See `docs/spikes/1125-skill-synthesis.md`. Depends on #1296, #1298.

Player feedback for synthesis, mirroring the milestone/challenge feedback path.

### Scope
- On the `SynthesizeSkill` server push (#1296): update the relevant stores, ~~**toast the synthesized-skill unlock**, and add a combat-log / event line.~~ - Sythesizing is player-driven, so there is no need to add toast/combat-log notifications.
- A **"new recipe available"** toast when a previously hint-only recipe becomes creatable (derived client-side from the availability state in #1298 / #1300 — e.g. on acquiring the last input or crossing a `Condition` threshold).

### Dependencies
#1296, #1298.

### Tests
Store update on push; toast/log emission for both the unlock and the new-recipe-available cases.

_Labels: enhancement, claude, scope: small._

ginderjeremiah:
Flagging a blocker before picking this up: the formal **Dependencies** (#1296, #1298) are both closed, but after the strikethrough the remaining scope is just the **"new recipe available" toast** that fires when a recipe transitions hint-only → creatable. That transition is derived from the per-recipe availability/reveal state (creatable-now / owned-but-gated / hint-only), which is built in **#1300** (the Synthesis screen, Area E) — and #1300 is still **open** (scope: large).

So this issue is effectively blocked on #1300: implementing the toast now would mean duplicating #1300's availability-state derivation, then reconciling it when #1300 lands. Recommend either (a) gating this on #1300, or (b) confirming the availability/transition derivation should live in a shared store consumed by both — in which case it'd be cleaner to fold this small toast into #1300.

Skipping for now and picking up #1323 (damage-types foundation) instead. Happy to take this once #1300 is in or the derivation home is decided.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1295, opened 2026-06-27
Skill synthesis: SkillRecipe reference-data model + Synthesis acquisition flag + admin authoring


ginderjeremiah:
Part of the skill-synthesis spike #1125 (**area A — foundation**). See `docs/spikes/1125-skill-synthesis.md`.

Add the authored `SkillRecipe` reference data that synthesis combines owned skills through, plus the new acquisition channel and backend authoring. The model is shaped **general** (inputs/conditions as child collections — the "combination engine, skills-first" decision) but only the skill→skill path is wired in V1.

### Scope
- **`SkillRecipe` entity** (zero-based identity, retire-not-delete): `Id`, `ResultSkillId`, child `Inputs` (skill ids, ≥1), child `Conditions` (`{ ProficiencyId, MinLevel }`), `RetiredAt`. Inputs/conditions as child collections so the `{ inputs, conditions, outputs }` shape generalizes later; V1 is a single result skill.
- **`ESkillAcquisition.Synthesis = 8`** (`Game.Core/Enums.cs`) — authoring intent that a skill is allowed to be a synthesis result.
- **DbContext config + migration.**
- **Cache holder + snapshot + `VersionKey`** following `SkillsCacheHolder`; a cache-time **reverse index** (skill id → recipes it is an input to) to drive the client hints (area D/E); zero-based contiguity assert.
- **entity↔core↔contract mappers**; reader repository + DI (`AllSkillRecipes()`, `GetSkillRecipe(id)`).
- **Authoring validation:** result must be `Synthesis`-flagged; inputs and result resolve to non-retired skills; inputs do not include the result; **acyclicity + reachability** over the recipe graph (a result may not be its own ancestor through chained recipes — the lighter analogue of the gateway cycle check being removed in the area-C issue).
- **Admin add/edit endpoints** (`Change<SkillRecipe>` + `ChangeSetProcessor`, child savers for inputs/conditions) with `[ReloadReferenceCaches]`.

### Dependencies
None — the proficiency system it references is already merged.

### Tests
Mapper round-trips, cache snapshot + reverse index, the acyclicity/reachability validation, the `Synthesis`-flag authoring guard.

_Labels: enhancement, claude, scope: large._

Issue #1125, opened 2026-06-22
Spike: skill synthesis (combine skills)


ginderjeremiah:
Deferred spike spun off from #982 (proficiency system) -- tracking. See `docs/spikes/982-proficiency-system.md` (deferred; decision 8).

Research a player-driven skill-combination / crafting feature (e.g. fire + earth -> lava; fuse two sword techniques into a stronger one). Beyond being a content pillar, it would give the player agency in the proficiency tree-seed gap and could supersede the seed-skill exception (decision 8 of #982). This would also remove the need for "gateway" proficiencies as you would be able to synthesize what is currently their "seed skill" and then unlock them like every other proficiency. This is the final step in the goal for proficiency progression to simply fall out naturally based on other progression mechanics.

Own spike. Related: #982.

ginderjeremiah:
Ideation complete — written up in [`docs/spikes/1125-skill-synthesis.md`](../blob/main/docs/spikes/1125-skill-synthesis.md).

**Settled direction:** an authored, deterministic **combination engine, skills-first** — a `SkillRecipe` (specific input skill ids + optional proficiency-level conditions → a pre-authored result skill), non-consumptive and one-time, added as a fourth skill-acquisition channel (`ESkillAcquisition.Synthesis`). It **supersedes #982 decision 8**: proficiency seed skills and cross-path gateways are removed, and every path opens the one uniform way (a contributing skill accrues XP), with synthesis bootstrapping the cross-path lines (`fire + earth → lava`). Discovery is a conservative hinted reveal; the engine's defining law is **non-consumptive, deterministic, possession/mastery-gated — never a consumed-resource economy**. Cross-type extensions (item-forging, imbuing, reagents, …) were priced and deferred as demand-gated, not roadmapped.

**Implementation sub-issues:**
- #1295 — `SkillRecipe` model + `Synthesis` flag + admin authoring (foundation)
- #1296 — `SynthesizeSkill` command + execution (anti-cheat)
- #1297 — Remove proficiency seed skills + cross-path gateways
- #1298 — Client: recipe reference-data delivery
- #1299 — Admin Workbench: synthesis recipe editor
- #1300 — Client: Synthesis screen (greenfield)
- #1301 — Client: synthesis feedback

PR #1314, opened 2026-06-27
Add SynthesizeSkill command + execution (#1296)

### LINKED ISSUE #1296: Skill synthesis: SynthesizeSkill command + execution (anti-cheat)
Part of the skill-synthesis spike #1125 (**area B**). See `docs/spikes/1125-skill-synthesis.md`. Depends on #1295.

Add the player-initiated synthesis action — the only way a `SkillRecipe` is executed.

### Scope
- **`SynthesizeSkill(recipeId)` socket command.**
- **Backend validation (authoritative — anti-cheat):** the recipe exists and is not retired; the player **owns every input skill as an unlocked skill** (`Player.Skills`) — innate item-granted skills are deliberately excluded (you can't equip-to-synthesize then unequip); every `Condition` is met against the player's current proficiency levels; the result is `Synthesis`-flagged.
- **On success:** `Player.UnlockSkill(result)` (idempotent — re-synthesis is a no-op) and emit a result payload for the client push (mirrors the skill-unlock / milestone push).
- **Non-consumptive:** inputs are never removed.

### Dependencies
#1295.

### Tests
Unit on each validation branch (missing input, unmet condition, retired recipe, non-`Synthesis` result, item-grant input rejected); integration (command → unlock + push); idempotency of a repeat synthesize.

_Labels: enhancement, claude, scope: medium._

ginderjeremiah:
## Summary

Implements the player-initiated skill-synthesis command (spike #1125, **area B**): the player combines owned input skills into a recipe's result skill. Builds on the `SkillRecipe` reference-data foundation merged in #1295.

The recipe is identified by its id; **all validation is server-side and authoritative (anti-cheat)** — a tampered client controls only the id, so every fact is re-checked against the live recipe and player state.

## How it addresses the issue

- **`SynthesizeSkill(recipeId)` socket command** — a player mutation, so it lives on the socket alongside the other player writes (serialized with the battle loop).
- **Authoritative validation** in `Player.TrySynthesizeSkill` (domain), orchestrated by a new `SynthesisService`:
  - the recipe exists and is **not retired** (`ISkillRecipes.ValidateRecipeId` bounds-guards a tampered id so it's rejected cleanly rather than throwing a 500);
  - the player owns **every input as an *unlocked* skill** (`Player.Skills`) — innate item-granted skills are derived at battle assembly and never live in the unlocked set, so they're excluded by construction (no equip-to-synthesize-then-unequip, decision 6);
  - **every proficiency condition is met** against the player's current levels (read from the `PlayerProgress` aggregate, like the gear gate; a missing proficiency counts as level 0).
- **On success:** idempotent `Player.UnlockSkill(result)` (re-synthesis is a no-op) and the result skill id is returned in the command response.
- **Non-consumptive:** inputs are never removed.

### Note on the `Synthesis`-flag check

The issue lists "the result is `Synthesis`-flagged" as a command-level validation branch. I deliberately did **not** re-check it at command time:

- The result is **already guaranteed** `Synthesis`-flagged by the `AdminSkillRecipes` authoring guard (#1295) — it's a reference-data invariant, **not a player-tamperable surface** (the player only sends a recipe id).
- The lean `Game.Core.Skills.Skill` battle model deliberately carries **no** acquisition flag (`docs/backend.md` → skill acquisition flags), so re-checking it would force a gameplay command to read the contract-layer flag and contradict that documented decision.

So that branch stays an **authoring-time** concern (covered by #1295's `AdminSkillRecipes` tests). I've documented this reasoning in `backend.md`. Happy to add a defensive check (or push the flag onto the core model, à la `Rarity`) if you'd prefer it as defense-in-depth — flagging it for a decision.

### Delivery vs. server push

The issue mentions "emit a result payload for the client push." Since synthesis is player-initiated, the command **response** is the delivery — there is no separate Redis-backplane server push. This matches the later edit to #1301 ("Synthesizing is player-driven, so there is no need to add toast/combat-log notifications").

## Test plan

- [x] **Domain unit tests** (`PlayerTests`) — every branch: success, non-consumptive inputs, retired recipe, missing input, innate-item-grant input rejected, unmet/met proficiency condition, missing proficiency = level 0, idempotent repeat synthesize.
- [x] **Integration tests** (`SynthesizeSkillSocketTests`) — command → unlock wiring, repeat-synthesize idempotency, missing-input rejection with no state change.
- [x] Full backend suite (2197 tests) green; `dotnet format --verify-no-changes` clean.
- [x] Regenerated TypeScript client (no drift); `npm run lint` (ESLint + svelte-check) clean.

Closes #1296

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nfrrb4GpMLLJu9Nhv4PNZd)_


ginderjeremiah:
## Automated Code Review — PR #1314 (SynthesizeSkill command + execution, #1296)

**Verdict: Approve with nits** ✅

Clean, well-scoped implementation of spike #1125 area B. Domain logic sits in `Player.TrySynthesizeSkill`, orchestration in `SynthesisService`, the command is a thin socket adapter, and validation is server-side authoritative throughout. All 9 CI checks are green (backend, frontend, 3× e2e, verify, codegen-drift). Doc and TypeScript client are updated. Per-branch test coverage is genuinely thorough.

### CI status
Green. All check runs pass on `1f94ecf`. (The `get_status` "pending" is only the absence of legacy commit statuses — `total_count: 0` — not a failure; the GitHub Checks API is authoritative here.)

### On the `Synthesis`-flag decision (called out explicitly)
**The author's decision not to re-check the flag at command time is correct for this command.** I verified the anti-cheat surface end-to-end:

- The player input is a single recipe **id**; the recipe→result mapping is server-side reference data the client cannot influence.
- `AdminSkillRecipes.SaveSkillRecipes` → `FindResultSkillViolation` rejects any recipe whose result is not `Synthesis`-flagged (and non-existent/retired) **before it can enter the catalogue**, so every reachable result is flagged by construction. This is a true authoring-time invariant, not a player-tamperable fact.
- Re-checking at command time would force a gameplay command to read the contract-layer acquisition flag (or push it onto the lean `Game.Core.Skills.Skill` battle model, which deliberately omits provenance) — contradicting the documented skill-acquisition-flag decision. The chosen approach mirrors the existing item-`GrantedSkillId` / enemy-skill-set pattern, which also validate the flag only at authoring.

So this is **not** a coverage gap — it is the documented intent-vs-reality boundary applied consistently. The one residual is that it trusts the authoring guard is the only write path (a direct DB seed could insert a non-flagged result), but that same trust model applies project-wide (starter equipment, starter skills), so it's acceptable. No defense-in-depth check needed; if you ever push the flag onto the core model for another reason (à la `Rarity`), revisit then.

I also confirmed `ValidateRecipeId` is genuinely load-bearing, not redundant: `GetSkillRecipe` → `GetById` throws `ArgumentOutOfRangeException` on an out-of-range id, so the up-front bounds guard converts a tampered id into a clean rejection instead of a 500. Good catch by the author.

### Blocking
None.

### Should-fix
None.

### Nits (non-blocking)
1. **`SynthesisService` duplicates the guard/save shape** that `PlayerService` centralizes in its private `SaveIf` helper. Because synthesis returns `int?` (not `bool`) and `SaveIf` is private to `PlayerService`, it can't be reused as-is — so the manual `if (!Try...) return null; await SavePlayer; return id;` is acceptable. Flagging only as a future DRY opportunity if a shared save-on-success helper is ever extracted to a common application-service base.
2. **Generic rejection message** (`"Unable to synthesize skill."`) is the right call for anti-cheat (no leak of which branch tripped) — noting only that it means a legitimate client can't distinguish "missing input" from "unmet proficiency." Fine for now; the client mirrors the rule for the affordance anyway.
3. The integration test's `InnateItemGrantedInputNotInUnlockedSet` case stands in a not-unlocked skill for an item-grant rather than actually equipping a granting item. The domain comment is honest about this, and the real exclusion is structural (innate grants never enter `Player.Skills`), so the substitution is reasonable — but a follow-up that equips an item with a `GrantedSkillId` and asserts it's still rejected would close the loop more convincingly.

### Notes
- Doc update to `backend.md` (skill-recipes section) is accurate and appropriately concise.
- Non-consumptive + idempotent-unlock semantics are correctly tested on both the domain and integration tiers.
- Braces, no null-forgiving operator, domain isolation (no entities in `Game.Api`/`Game.Application`) all hold.

Recommendation: merge after a glance at nit #1/#3 (neither blocks). Not merging as part of this review.

—
*Automated review by Claude (Opus 4.8). Not a substitute for human judgment on the flag-check design decision, which is flagged above for an explicit sign-off.*

---

PR #1315, opened 2026-06-27
Deliver skill recipes to client as reference data (#1298)

ginderjeremiah:
## Summary

Implements the client reference-data delivery for skill synthesis (spike #1125, **area D**). It builds on the `SkillRecipe` reference-data foundation merged in #1295, which already provides the cache holder, contract projection (`ISkillRecipes.AllSkillRecipes()`), and version key — so this change is the thin delivery layer that puts that set on the wire and into the client's cached static data.

## How it addresses the issue

- **`GetSkillRecipes` socket command** — a new `AbstractReferenceDataCommand<SkillRecipe>` mirroring `GetSkills`, served over the authenticated socket and version-keyed off `ISkillRecipes.VersionKey`. The reference-data assembly scan picks it up automatically, so it's included in `GetReferenceDataVersions` (the per-set content hash) with no extra wiring.
- **Codegen contracts** — regenerated the TypeScript client. `ISkillRecipe`/`ISkillRecipeCondition` already existed (the admin HTTP authoring path emits them), so the only generated delta is the `'GetSkillRecipes': ISkillRecipe[]` socket-response-map entry.
- **`staticData` slot** — a `skillRecipes` slot added to the static-data store (getter/setter + `loaded` set), and a `skillRecipes` row in `REFERENCE_DATA`, so the loading screen and the silent session-resume path version-check, download-on-change, and locally cache the set exactly like every other reference set.
- **No separate player-data command** — per the issue, recipe availability + hint state is derived client-side from this set plus the player's already-delivered owned skills and proficiency levels. That derivation lands in the synthesis screen / feedback sub-issues (#1300, #1301), not here.

The admin Workbench reference loader (`reference.svelte.ts`) is deliberately left untouched — wiring recipes into the authoring picker set is the recipe-editor sub-issue (#1299), not this delivery.

## Test plan

- [x] **Backend integration** (`ReferenceDataSocketTests`) — new `GetSkillRecipes_ReturnsSeededSkillRecipesWithInputsAndConditions` (self-contained seed → command returns the recipe with its inputs/conditions/result), and the `GetReferenceDataVersions` exact-command-set assertion extended to include `GetSkillRecipes`. Full class green (16 tests). Added a `TestDataSeeder.CreateSkillRecipeAsync` helper.
- [x] **Frontend unit** — `static-data`, `reference-data` (version-key cache invalidation / hydrate), and `loading-view` set lists extended to cover `skillRecipes`/`GetSkillRecipes`. Full suite green (2755 tests).
[...]

Issue #1298, opened 2026-06-27
Skill synthesis client: recipe reference-data delivery


ginderjeremiah:
Part of the skill-synthesis spike #1125 (**area D**). See `docs/spikes/1125-skill-synthesis.md`. Depends on #1295.

Deliver synthesis recipes to the client as version-keyed reference data.

### Scope
- **`GetSkillRecipes` socket command** (reference, version-keyed) + codegen contracts.
- **`staticData` slot** + reference-data registration + version-keyed local cache (mirrors `GetSkills`).
- **No separate player-data command:** recipe **availability + hint state** is derived client-side from the delivered recipes + the player's owned skills + proficiency levels (all already on the client).

### Dependencies
#1295.

### Tests
Contract codegen, store hydration, version-key cache invalidation.

_Labels: enhancement, claude, scope: medium._

PR #1319, opened 2026-06-27
Add admin Workbench synthesis recipe editor (#1299)

### LINKED ISSUE #1299: Admin Workbench: synthesis recipe editor
Part of the skill-synthesis spike #1125 (**area A — admin UI**). See `docs/spikes/1125-skill-synthesis.md`. Depends on #1295.

The admin Workbench editor for authoring `SkillRecipe`s (split out of the area-A foundation, mirroring how the proficiency editor #1132 followed the proficiency foundation).

### Scope
- A **recipe editor**: result-skill picker, input-skill list, and proficiency-`Condition` rows (`ProficiencyId` + `MinLevel`), following the `persistEntity` pattern with child savers (inputs, conditions), under the relevant Workbench group.
- **Retired-record exclusion** in the pickers (result/input skill pickers exclude retired skills; an existing authored reference stays visible).
- Surface the authoring-validation errors from #1295 (non-`Synthesis` result, cyclic/unreachable recipe).

### Dependencies
#1295 (endpoints + contracts).

### Tests
Editor save / child-saver behavior per the existing Workbench editor test patterns.

_Labels: enhancement, claude, scope: medium._

ginderjeremiah:
## Summary

Implements the admin Workbench editor for authoring `SkillRecipe`s (skill-synthesis spike #1125, **area A — admin UI**). It builds on the area-A foundation (#1295, which shipped the cache holder/reader + the `AddEditSkillRecipes` / `SetSkillRecipeInputs` / `SetSkillRecipeConditions` authoring endpoints and contracts) and on #1298's `GetSkillRecipes` reference delivery (PR #1315, now on `main`) — the functional prerequisite this issue was earlier deferred on.

It follows the entity-driven Workbench pattern (`EntityConfig` + `persistEntity` with child savers), mirroring how the class editor saves its child collections and how the proficiency editor followed its foundation.

## How it addresses the issue

- **Recipe editor** (`entities/skill-recipe.ts`), registered under the **Progression** Workbench group:
  - **Result-skill picker** — a select restricted to `Synthesis`-flagged skills.
  - **Input-skill list** — a chips section over `inputSkillIds`.
  - **Proficiency-`Condition` rows** — a table section of `ProficiencyId` + `MinLevel`.
  - Persisted via `persistEntity`: the identity save (`AddEditSkillRecipes`) carries only the result skill (+ retirement), and the inputs/conditions go through their own setters as child savers, so a child-only edit never needlessly hits the identity endpoint and an untouched collection is skipped.
- **Retired-record exclusion in the pickers** — the result and input pickers omit retired skills from new authoring (a hard block), while an already-authored value stays visible (marked `· retired`) and editable; the result picker is additionally restricted to `Synthesis`-flagged skills.
- **Surfaced authoring-validation errors (#1295)** — the cheap, local rules are flagged inline as warnings: a result that isn't a live `Synthesis` skill, no inputs, an input equal to the result, and a condition level below 1. The authoritative graph checks (acyclicity/reachability), the input-equals-result and proficiency-range checks remain backend-enforced and surface as a save-failure toast via the existing store error path.

### Two small generic additions

A recipe has **no `name`** of its own (it's identified by its result skill) — a first for the generic Workbench, which rendered/searched on `record.name`. Rather than store a stale derived name:

- Added an optional `EntityConfig.title(rec)` hook (consistent with the existing `headline`/`dirtyKeys` opt-in hooks) that drives the list/detail title and search live, falling back to `name`/`blankName`.
- Closed a latent gap where a `fields` section's `warn` predicate was silently ignored — it now surfaces (after the per-field required warns), which is how the result section flags a non-`Synthesis`/retired result. No existing `fields` section set `warn`, so this is a no-op for every other entity.

Documented the `title` hook concisely in `docs/frontend.md` (Workbench section).

## Test plan

- [x] **Entity unit tests** (`entities/skill-recipe.test.ts`) — `newItem` defaults (and the no-`Synthesis`-skill fallback), the derived title/headline/meta, the conditions `newRow`, every warn predicate, the picker retired/Synthesis filtering, and the persist path: inputs-only and conditions-only changes skip the identity Edit and the untouched sibling setter; a result change sends an identity Edit with child collections stripped; a new recipe resolves its real id before the child savers run.
- [x] **Validation test** — a `fields` section now surfaces its section-level `warn` after the field warns.
- [x] Full Workbench suite green (`vitest run src/tests/routes/admin/workbench` — 37 files, 395 tests).
- [x] `npm run lint` (ESLint + svelte-check) clean.

Closes #1299

---
_Generated by [Claude Code](https://claude.ai/code/session_01UqAwj4YACXXFhp4722264Q)_


ginderjeremiah (commented):
## Review: Admin Workbench synthesis recipe editor (#1299, spike #1125 area A)

**Verdict:** Approve _(submitted as a comment because I'm the PR author — GitHub won't take a gating event on one's own PR)_ — no blocking findings. The editor is correctly scoped to frontend authoring, follows the established `persistEntity`/`EntityConfig` pattern, and matches the spike's design decisions.

I verified the change against `docs/frontend.md` (Workbench section) and the spike doc `docs/spikes/1125-skill-synthesis.md`:

- **Authoring rules match the design.** Result picker restricted to live `Synthesis`-flagged skills (decision 4); inputs are any non-retired skill with a "no inputs" warn (data model: inputs ≥ 1, "non-retired inputs/result"); conditions are `{proficiencyId, minLevel}` with a min-level≥1 warn (decision 5). The authoritative checks (acyclicity/reachability, input-equals-result, proficiency range) are correctly left to the backend and surface via the existing save-failure toast rather than being re-implemented client-side.
- **Persist path is right.** `toPrimaryDto` strips the child collections, so the identity diff in `persistEntity` only fires `AddEditSkillRecipes` when the result/retirement actually changes; a child-only edit goes through the input/condition setters and skips the identity endpoint. The retired/non-`Synthesis` current value stays visible in the picker because `FieldControl.svelte:80` passes the field's current value into `options(current)`, and `synthesisResultSkillOptions(keep)` keeps it.
- **The two generic additions are minimal and justified.** The `title(rec)` hook is a clean opt-in consistent with the existing `headline`/`dirtyKeys` hooks, and the four nameless-entity display sites (list row + search, detail title, retire-confirm body) are all updated — I checked there are no other `record.name` render sites that a nameless entity would blank (the remaining ones in `UsageSection`/`RewardPicker` are tag/challenge-specific). The `sectionWarnings` fix correctly surfaces a `fields`-section `warn` after the per-field warns, and it's a genuine no-op for every existing entity (none set `warn` on a `fields` section).

### Should-fix

- _None blocking._

### Nits

- **No component-level test for the nameless-entity path.** `entities/skill-recipe.test.ts` thoroughly covers the config transforms and persist behavior, and `validation.test.ts` covers the new `fields`-section warn. But the `title`-driven rendering/search in `WorkbenchList.svelte` (the new `displayName` helper now backing the search filter) and `WorkbenchDetail.svelte` only has the existing named-entity tests. Worth a small `WorkbenchList` test that a `title`-only entity renders its derived title and filters by it, to lock in that search uses `displayName` rather than `name`. Low risk since the derivation itself is unit-tested and the component change is a thin fallback.
- **`metaOf` double-cast** (`(s as unknown as { baseDamage: number })`) is copied verbatim from `enemy.ts:129`/`class.ts:134`, so it's consistent and not introduced here — but it's a latent type-safety gap in the shared `ChipsSectionConfig.catalogue` entry type. Reasonable follow-up (widen the catalogue entry type to carry meta fields) rather than anything for this PR.

### Test coverage

Meets the project's frontend testing rules: unit tests mirror the source path under `tests/`, cover `newItem` defaults (including the no-`Synthesis`-skill `-1` fallback), the derived title/headline/meta, picker retired/`Synthesis` filtering, every warn predicate, and all four persist permutations (inputs-only, conditions-only, result-change with stripped children, and a new add resolving its real id before the child savers run). The only gap is the minor component-render/search test noted above. No backend or battle-logic surface here, so no mirrored-test obligation.

### What's good

Tight adherence to the existing Workbench architecture — the editor is a pure config object plus two small, well-documented generic hooks, no forking of the shared components. The persist wiring (identity DTO carrying only the result + retirement, children via their own setters) is exactly the class-editor pattern, and the retired-value-stays-visible handling is consistent with the other reference pickers. Docs updated concisely. Clean work.

---

Issue #1320, opened 2026-06-27
Introduce damage types to skills/effects


ginderjeremiah:
Right now, damage is just an amount with no discriminator applied to it. 10 damage via a fireball skill vs 10 damage from a sword skill is represented the same way. That worked well in the beginning, but with the introduction of proficiencies, there became more motivation for being able to buff only certain skills. For example, a fire magic proficiency doesn't want to just buff intellect, because then ALL magic skills will become more powerful as a result. It would be advantageous to introduce damage types (for example, fire/water/ice/electric/physical, etc) so proficiencies can isolate their buffs to the type of magic that is actually relevant. This also has the added bonus of increasing strategy, by allowing certain enemies or bosses to have increased (or decreased) resistance to certain kinds of damage.

In this spike, we should evaluate the costs and benefits of introducing damage types and also explore the potential of such a system. If we do choose to implement it in some form, a spike document should be created and sub-issues attached to this parent issue.

ginderjeremiah:
## Spike outcome

Design decided — full writeup in [`docs/spikes/1320-damage-types.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/1320-damage-types.md).

**Model:** damage types ride the attribute pipeline. A skill carries one leaf `DamageType`; per-type/per-category **amplification** (attacker) and **resistance** (defender) are percentage attributes resolved through a static `applies(type) → keys` map, so proficiencies/items/mods/effects/class passives all gain typed buffs with no new plumbing.

**Taxonomy:** leaf types `Physical, Fire, Water, Earth, Wind, Bleed, Poison, Burn`; cross-cutting categories `Elemental` and `DoT` (Physical is a leaf, not cross-cutting; no `Magical`). Resistances/amps are **additive within a side**, applied as separate multipliers across sides, **unclamped** so vulnerability and absorption fall out for free. `% resist` first, flat `Defense` last; crit still punches through. DoT is typed via per-type per-second accumulators. Bidirectional (player & enemy), crit/dodge/block stay player-only.

**Reduce-to-today guarantee:** with no typed content, every added multiply is ×1.0 (exact), so existing battles and the parity matrix are undisturbed.

### Sub-issues
- #1323 — Foundation: `EDamageType`, amp/resist attributes, `applies()` map *(medium)*
- #1324 — Direct-hit typing: `Skill.DamageType` + damage pipeline + parity *(medium, depends on #1323)*
[...]

PR #1351, opened 2026-06-29
Workbench: searchable, group-by-type attribute picker (#1327)

### LINKED ISSUE #1327: Workbench: searchable, group-by-type attribute picker
Part of #1320. See `docs/spikes/1320-damage-types.md` (Area E).

The attribute picker is a flat native `<select>` of every enum value, used in 5 authoring contexts (skill damage multipliers, skill effects, item mods, enemy distributions, class stat locks). With the damage-type attributes (and future content) the enum grows well past what a flat dropdown can handle.

## Scope
- Replace the picker with a searchable control that **groups by damage type** (and the existing primary/secondary/status grouping).
- Apply it across all 5 authoring contexts.

## Tests / acceptance
- Picker filters by search text and groups correctly; selecting still writes the attribute id.
- All 5 authoring contexts use the new control.

**Depends on:** foundation (A) — soft; valuable as soon as the new attributes exist.
**Scope:** small

ginderjeremiah:
## Summary

The admin workbench's attribute picker was a flat native `<select>` of every `EAttribute`. The #1320 damage-type work pushed that enum well past 30 values (the amplification/resistance family plus the typed DoT accumulators), so a flat dropdown is no longer usable. This replaces it with a reusable **searchable, group-by-type** control (spike #1320, Area E).

Options are grouped by the reference display taxonomy (**Primary / Secondary / Status**), with the **Affinity** family further split into one sub-group per **damage type** — reusing the same grouping the attribute-breakdown screen already applies, so the two surfaces stay consistent. Selecting still writes the attribute id.

## How it addresses the issue

- **`AttributePicker.svelte`** — a trigger that mimics the native select look, opening a non-blocking `Popover` (focus-trap / Escape / backdrop dismissal for free, and it escapes table-cell clipping) hosting a search box + the grouped, filtered options. Damage-type sub-group headers carry the type's accent + glyph. Self-styled so it works both in the `.workbench` root and the progression editor's `.prog` root.
- **`attribute-groups.ts`** — pure grouping + search-filter logic (taxonomy bands, per-damage-type Affinity sub-groups, a leading headerless group for the optional "None" sentinel, and a flat fallback when the reference set isn't loaded).
- **New `type: 'attribute'` field/column kind** wired into `FieldControl`, `TableCell`, and the `SectionTable` unique-column (Add-disable) detection. It reuses the same `options` provider as `select`.
- **Every attribute picker routed through it** — beyond the 5 contexts named in the issue, this covers all attribute selection so none is left as a flat dropdown: skill damage multipliers, skill effects (attribute **and** scaling attribute), item bonuses, item-mod bonuses, enemy distributions, class stat distributions, the class signature passive (attribute + optional scaling), and the progression milestone editor.
- **`GetAttributes` added to the workbench reference loader** so the by-type grouping has its metadata when the admin is entered directly (without the game's loading screen).

## Test plan

- [x] `attribute-groups.test.ts` — taxonomy ordering, per-damage-type Affinity split, display-order sort, "None" sentinel handling, fallback, and search filtering.
- [x] `AttributePicker.test.ts` — trigger shows current value; opens a grouped/searchable panel; selecting calls `onChange` and closes; Enter selects the first match; a value taken by a sibling row is disabled but the current value never is.
- [x] `reference-load.test.ts` updated for the added `GetAttributes` socket call (transport split now 10 socket / 2 HTTP).
- [x] `npm run test:unit` — 2979 passed.

Frontend-only change; no backend or battle-parity surface touched.

Closes #1327

---
_Generated by [Claude Code](https://claude.ai/code/session_01E3ium21fKkwCFh56NCBdR4)_


UI/src/routes/admin/workbench/components/AttributePicker.svelte
```diff
@@ -0,0 +1,328 @@
...
+-->
+<div class="attr-pick">
+	<div class="attr-trigger-wrap" style:width="100%">
+		<button
+			type="button"
+			class="attr-trigger"
+			class:dirty
+			class:invalid
+			aria-label={ariaLabel}
+			aria-haspopup="dialog"
```
ginderjeremiah, line 20: Nit (optional): the trigger declares `aria-haspopup="dialog"` but not `aria-expanded`. Adding `aria-expanded={open}` would let assistive tech announce the open/closed state of the popover — a small win that fits this project's a11y conventions (frontend.md → Accessibility). Take it or leave it.

---

ginderjeremiah (commented):
## Review: searchable, group-by-type attribute picker (closes #1327)

**Verdict:** Approve — clean, well-scoped, well-tested frontend-only change that fully satisfies the issue and goes beyond the 5 named contexts to route *every* attribute picker through the new control. (Submitted as a `COMMENT` only because I'm the PR author and GitHub won't accept a self-approval; the intended verdict is **Approve**.)

Grouping logic is correctly extracted into a pure, unit-tested `attribute-groups.ts`; the component composes the existing `Popover` primitive (so it inherits focus-trap / Escape / backdrop dismissal and autofocuses the search box) rather than hand-rolling a backdrop; damage-type colour/name/glyph all go through the `$lib/common/damage-type-display` helpers exactly as the styling guidelines require; and the new `type: 'attribute'` field/column kind reuses the same `options` provider as `select`, keeping the diff small. Disabled-value semantics (`v !== value`) match the existing `select` behavior, and the `SectionTable` Add-disable detection was correctly extended to the new kind.

### Nits
- `AttributePicker.svelte:20` — trigger has `aria-haspopup="dialog"` but no `aria-expanded`; adding `aria-expanded={open}` is a small a11y improvement. Optional.

### Test coverage
- Strong. `attribute-groups.test.ts` covers taxonomy ordering, per-damage-type Affinity split, displayOrder sort (with deliberately scrambled input), the "None" sentinel, the no-reference fallback, and search filtering. `AttributePicker.test.ts` covers trigger text, open/grouped/searchable panel, select-and-close, Enter-to-pick-first-match, and the sibling-disabled-but-current-allowed rule. `reference-load.test.ts` was updated for the added `GetAttributes` socket call (10 socket / 2 HTTP). This matches the project's frontend testing rules; no backend or battle-parity surface is touched.

### What's good
- The grouping logic was pulled out as pure functions instead of living in the component — directly testable and consistent with the attribute-breakdown grouping it mirrors.
- `frontend.md` was updated with the new "attribute selection uses a searchable, group-by-type picker, not a flat `<select>`" decision, so the pattern is documented for future pickers.
- Sensible robustness: untyped-Affinity fallback, headerless sentinel group, and a flat fallback when the reference set isn't loaded.

Not auto-merging: CI is still in progress (frontend unit + firefox/webkit e2e at review time) and there's an open nit — happy to leave the call (and the optional `aria-expanded` tweak) to you.

---

Issue #1341, opened 2026-06-28
Proficiency: admin authoring (ActivityKey) + docs update


ginderjeremiah:
Part of spike **#1318** — full design in [`docs/spikes/1318-proficiency-progression-avenues.md`](../blob/main/docs/spikes/1318-proficiency-progression-avenues.md).

## What

Update authoring and documentation for the reworked model.

- **Admin Workbench** — the proficiency/path editor moves from skill→path contributions to a single **`ActivityKey`** binding per path (damage type / category / event picker). Remove the `FalloffBase` and `HomeTier` controls and the skill-contribution section.
- **Docs** — refresh the **Proficiency** section of `docs/game-design.md` for the effect-based `pie × activity/power` accrual, `ActivityKey` routing, the activity-key set (offense / elemental / weapon / DoT / resist / crit / dodge / heal / reflect), the two-book model, the removal of falloff/home-tier/rarity-weight, and the weapon-type taxonomy addition. Touch the damage-types section for the weapon leaves under the `Physical` category.

## Touches
The Workbench proficiency/path editor + its save endpoints; `docs/game-design.md`.

## Notes
- **Depends on:** the accrual engine rework (model fields) and the weapon-type leaves (taxonomy) being settled.

Issue #1336, opened 2026-06-28
Proficiency accrual rework — effect-based XP (activity ÷ power) + ActivityKey routing


ginderjeremiah:
Part of spike **#1318** — full design in [`docs/spikes/1318-proficiency-progression-avenues.md`](../blob/main/docs/spikes/1318-proficiency-progression-avenues.md). Reworks the shipped representation + falloff accrual (#1116 / #1160 / #1161).

## What

Replace the representation/attention accrual with the unified **effect-based** model:

```
XP_proficiency = basePie × clamp( activity ÷ totalAttributes )   // victory-only, then routed to the frontier tier
```

- **Power-normalization subsumes the difficulty multiplier** — `activity / totalAttributes` *is* the continuous difficulty ratio; the clamp mirrors `ServerGameConstants.MaxExpRewardMultiplier`. Do **not** also multiply by the banded `DifficultyMultiplier` (power would double-count → strip-power exploit).
- **No representation floor** — a path earns strictly in proportion to its real contribution; commitment (build + loadout) is the only lever.
- **Routing by `ActivityKey`** — replace the `SkillPathContribution` skill→path join with a single `ActivityKey` per path (a damage type, category, or combat event). A battle quantity trains the path(s) its key resolves to via the (extended) `applies()` map. Offense consumes existing per-skill damage (`BattleStats.SkillStats.TotalDamage`) grouped by the skill's resolved damage type.
- **Remove** `FalloffBase`, `HomeTier`, and rarity/tier weight from accrual (rarity reverts to display-only).
- Wire both the **live** path (`BattleStatisticsEventHandler`) and the **offline** consolidation (`OfflineProgressSimulator`) — they must accrue identically.

## Touches
`ProficiencyXpCalculator`, `ProficiencyRewardService.BuildContributions`, the `Path`/`Proficiency` reference model + DbContext + migration, mappers/contracts, the live + offline accrual paths. Rework the existing accrual unit/integration tests.

## Notes
- **No new parity surface** — accrual stays server-side at battle completion, off the tick loop. Only the attribute *bonus* (unchanged) is parity-sensitive.
- **Depends on:** #1320 direct-hit typing (skills must carry a resolved damage type at the recording site).

Issue #1330, opened 2026-06-27
Spike: rework flat mitigation — Defense armor curve + replace Block with reflection


ginderjeremiah:
## Problem

Flat damage reduction — `Defense`, and Block's flat `BlockReduction` — is **scale-dependent**: a fixed value is overpowering against small hits and worthless against big ones, with a degenerate immunity breakpoint at the `0` floor. The classic "useless or too powerful" feel. Block compounds this by gating flat reduction behind **RNG**, which is anti-tank (tanks want *low* variance, not more of it), and it overlaps both `Defense` (the same reduce-damage operation) and `Dodge` (which already owns the high-variance defensive / evasion niche).

Surfaced during the damage-types spike (#1320); split out as its own spike.

## Direction (decided in discussion)

1. **Defense → diminishing-curve %.** Convert the bulk stat to a percentage via `defReduction = Armor / (Armor + K·attackerLevel)`, applied as a multiplier rather than a subtraction. Gives **linear effective-HP** (each point worth a constant % EHP — never useless) while the % asymptotes below 100% (never immune, no breakpoint); `K` scaling on the **attacker's** level keeps it relevant at high zones. Applies to both battlers.

2. **Remove Block as damage reduction** — redundant with Defense, and RNG defense works against the tank role.

3. **Repurpose the slot into deterministic damage reflection.** Return a % of damage taken to the attacker. **Deterministic** (no proc — tanks want reliability) and **authored-only** (granted by gear/proficiency/class, not derived from a core attribute), so it's a deliberate build identity rather than a stat tax, and player/enemy gating is moot (enemy reflection becomes a free boss-design seam). It gives tanks a real **kill condition** in a game that otherwise draws on the 2-minute timeout, and self-scales against the hard-hitters you'd tank. Scoped to **direct hits** (not DoT); reflect bypasses the attacker's own mitigation. *Retaliation* (a counter scaling off your own stats) was the considered alternative; reflection chosen for its self-scaling elegance and single-knob simplicity.

End-state mitigation becomes fully multiplicative — `dealt × (1 − Σ typeResist) × (1 − defReduction) × (reflection applied to attacker)` — removing the flat-subtraction floor entirely (the damage-types V1 absorption branch then disappears, since nothing flat remains to floor).

## Relationship to #1320

[...]

Issue #1318, opened 2026-06-27
Add other avenues for progressing proficiencies


ginderjeremiah:
Right now, proficiency XP is derived purely from skill contributions and pie-splitting a total that is based on fight victory XP.

Tying proficiency progression directly to skills limits the potential usage of proficiencies. The following proficiency paths, for example, would benefit from having other avenues of progression:

- Fire/Water/Poison/Elemental (etc.) damage reduction (Proficiency xp increases by _taking_ damage)
- Bleed/poison/burn damage (Proficiency xp increases by applying effects or dealing damage through effects).
- Crit/dodge/block (Proficiency xp increases when these actions are performed.

This spike is to evaluate the costs and benefits of adding these additional avenues. If we determine that this is a worthwhile endeavor, then a spike should be created to plan out the introduction of this feature and sub-issues attached to this parent issue.

ginderjeremiah:
## Spike concluded — yes, worthwhile; replace representation rather than bolt on avenues

Full write-up: [`docs/spikes/1318-proficiency-progression-avenues.md`](../blob/main/docs/spikes/1318-proficiency-progression-avenues.md).

**Conclusion:** there is *one* avenue — every proficiency is a mastery track for an **activity key**, trained by `pie × (activity ÷ total attributes)` on victory — and "skills fired" was only ever a special case. The shipped representation + falloff accrual is replaced; the issue's three examples (elemental/DoT resistance, DoT mastery, crit/dodge) become instances of the same formula and routing rule.

Key decisions: power-normalization subsumes the difficulty multiplier; no representation floor; magnitude carried consistently by `activity/power` (pre-mitigation exposure, no self-throttle); routing by a single `ActivityKey` per path (skill→path join, home-tier, and **falloff** removed); rarity weight dropped from accrual; healing/reflection in scope, pure-utility out. Stays server-side at completion (no parity surface), victory-only. Gated on the damage-types pipeline (#1320).

**Implementation sub-issues:** #1336 (accrual engine) · #1337 (BattleStats) · #1338 (resist & DoT bindings) · #1339 (proc/heal/reflect bindings) · #1340 (weapon-type leaves) · #1341 (admin + docs).

**Follow-up spikes** (standalone): #1342 — restrict skill usability by equipped weapon · #1343 — multi-typed damage.

PR #1355, opened 2026-06-29
Wire proc/heal proficiency activity bindings (#1339)

### LINKED ISSUE #1339: Proficiency: proc / heal / reflect activity bindings
Part of spike **#1318** — full design in [`docs/spikes/1318-proficiency-progression-avenues.md`](../blob/main/docs/spikes/1318-proficiency-progression-avenues.md).

## What

Add the **event-keyed** activity bindings — proficiencies trained by combat events rather than typed damage:

- **Precision** ← crit damage (`pie × critDamage / totalAttributes`). Damage-type-neutral (single global track; crit damage is not typed at the recording site).
- **Evasion** ← dodged damage (`pie × dodgedDamage / totalAttributes`). Single global track.
- **Restoration** ← healing done (`PlayerDamageHealed`).
- **Retribution** ← reflected damage (the #1330 `DamageReflection` output).

All reuse existing `BattleStats` aggregates (crit damage, dodged damage, healing) except reflected damage, which the mitigation rework produces.

## Touches
Activity-key resolution for events, the four bindings, and tests.

## Notes
- **Depends on:** the #1318 accrual engine; the **Retribution** binding additionally depends on #1330 (deterministic reflection). Precision/Evasion/Restoration can land without #1330.

ginderjeremiah:
## Summary

Wires the **event-keyed** proficiency activity bindings from spike #1318 onto the now-shipped effect-based accrual engine (#1336):

- **Precision** ← crit damage (`stats.CriticalDamageDealt`)
- **Evasion** ← dodged damage (`stats.DamageDodged`)
- **Restoration** ← healing done (`stats.PlayerDamageHealed`)

These are damage-type-neutral combat magnitudes, so each maps straight to a single `EActivityKey` (`Crit`/`Dodge`/`Heal`) in `ProficiencyRewardService.BuildActivities` — no `applies()` routing, no per-type split. The existing routing loop then routes each key's activity to the frontier tier of every path bound to it, exactly as it does for the typed offense book.

## How it addresses the issue

#1339 asked for the four event bindings. Three are delivered here. **Retribution** (reflected damage) is deliberately left **inert**: nothing produces a reflected-damage figure until the mitigation rework **#1330** lands, which the issue itself calls out ("Precision/Evasion/Restoration can land without #1330"). The `EActivityKey.Reflect` member and its doc note remain, so a Retribution-keyed path simply accrues nothing until #1330 wires it.

Because both the live path (`BattleStatisticsEventHandler`) and the offline batch (`BattleService`) call the same `AccrueAndApply` with a `BattleStats` produced by the same simulator, the new avenues accrue identically on both paths — the "offline == live" invariant holds with no extra wiring.

## Test plan

- [x] `dotnet build` — clean
- [x] `Game.Application.Tests` — 636 passed (integration, DB-backed reference data)
- [x] `Game.Core.Tests` — 1000 passed

New tests in `ProficiencyRewardServiceTests`:
- `CritDamage_TrainsThePrecisionPath`
- `DodgedDamage_TrainsTheEvasionPath`
- `HealingDone_TrainsTheRestorationPath`
- `ReflectedDamage_IsNotYetWired_SoARetributionPathTrainsNothing` (inertness guard for Retribution / #1330)
- `OffenseAndEventAxes_TrainInParallel_WithoutDilution` (spike decision 4: axes don't share a pie)

## Notes / follow-ups

- **Docs were intentionally not touched.** The spike assigns the `docs/game-design.md` Proficiency rewrite (effect-based accrual, the full activity-key set, two-book model) to the dedicated tooling/docs sub-issue **#1341**, so folding it in here would duplicate that scope.
- The `Retribution` binding remains tracked by #1339's relationship to **#1330**; this PR does not close out that avenue.

Closes #1339

---
_Generated by [Claude Code](https://claude.ai/code/session_01QpoeubmyhbkU6WVeD6cp4Z)_

### DIFF
--- a/Game.Application.Tests/Services/ProficiencyRewardServiceTests.cs
+++ b/Game.Application.Tests/Services/ProficiencyRewardServiceTests.cs
@@ -359,6 +359,25 @@ namespace Game.Application.Tests.Services
         }
 
         [Fact]
+        public async Task ZeroEventActivity_TrainsNothing()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+
+            // The AddEvent amount > 0 guard: a battle with no crit damage produces no Crit activity, so a
+            // Crit-keyed path is not even routed to (no zero-activity slice reaches the calculator).
+            var path = await TestDataSeeder.CreatePathAsync(context, name: "Precision", activityKey: EActivityKey.Crit);
+            await TestDataSeeder.CreateProficiencyAsync(
+                context, name: "Precision", maxLevel: 10, baseXp: 1m, xpGrowth: 1m, pathId: path.Id, pathOrdinal: 0);
+            var playerId = await SeedPlayerAsync(context);
+            await ReferenceCacheReloader.ReloadAllAsync(scope.ServiceProvider);
+
+            var (_, accrual) = await AccrueStatsAsync(scope, playerId, new BattleStats { CriticalDamageDealt = 0 });
+
+            Assert.Empty(accrual.Results);
+        }
+
+        [Fact]
         public async Task OffenseAndEventAxes_TrainInParallel_WithoutDilution()
         {
             using var scope = CreateScope();
--- a/Game.Application.Tests/Services/ProficiencyRewardServiceTests.cs
+++ b/Game.Application.Tests/Services/ProficiencyRewardServiceTests.cs
@@ -270,6 +270,126 @@ namespace Game.Application.Tests.Services
             Assert.Single(player.Skills);
         }
 
+        [Fact]
+        public async Task CritDamage_TrainsThePrecisionPath()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+
+            // A path keyed on the Crit event trains from the battle's crit damage — no skill contribution and no
+            // damage-type routing involved (Precision is a single global, type-neutral track; spike #1318).
+            var path = await TestDataSeeder.CreatePathAsync(context, name: "Precision", activityKey: EActivityKey.Crit);
+            var tier = await TestDataSeeder.CreateProficiencyAsync(
+                context, name: "Precision", maxLevel: 10, baseXp: 1m, xpGrowth: 1m, pathId: path.Id, pathOrdinal: 0);
+            var playerId = await SeedPlayerAsync(context);
+            await ReferenceCacheReloader.ReloadAllAsync(scope.ServiceProvider);
+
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Wire proc/heal proficiency activity bindings (#1339)

**Verdict:** Approve _(posted as a comment — I'm the PR author, so GitHub won't accept a self-gating Approve; treat this as an approval)_ — the change cleanly wires three of the four event bindings on the existing accrual engine, with Retribution correctly deferred.

### Scope check
The PR delivers Precision ← `CriticalDamageDealt`, Evasion ← `DamageDodged`, and Restoration ← `PlayerDamageHealed`, and deliberately leaves Retribution (Reflect) inert. This matches #1339 exactly: the issue itself notes "Precision/Evasion/Restoration can land without #1330" and ties Retribution to the reflection rework. Verified against `BattleContext.cs:146/170/103` that the three `BattleStats` fields carry the intended semantics (post-Defense crit damage, post-Defense avoided hit, healing done), so the bindings read the right magnitudes.

### Design
- `AddEvent` maps each event magnitude straight to a single global `EActivityKey` with no `applies()` routing — correct, since these are damage-type-neutral and the event keys never collide with the damage-type keys already in `activityByKey`. The `amount > 0` guard keeps event-free battles from producing zero-activity entries.
- The "axes don't share a pie" property (spike #1318 decision 4) holds because `ProficiencyXpCalculator.Split` claims `pie × clamp(activity/power)` per path independently — so crit damage training Precision while its typed damage also trains the offense path is intended, not double-spend.
- No frontend mirror is required: proficiency accrual is server-side reward logic, not battle-simulation logic.

### Docs
Skipping the `docs/game-design.md` Proficiency rewrite here is the right call — #1341 explicitly owns that rewrite (and the admin authoring change). The inline doc comments on `ProficiencyRewardService` and the `EActivityKey` enum were updated to match the new wiring, which keeps the source-level docs honest.

### Test coverage
Good — each binding has a positive test, Reflect has an inertness guard, and `OffenseAndEventAxes_TrainInParallel_WithoutDilution` proves the no-shared-pie property across axes. Integration-style (DB-backed reference data) is appropriate here since `BuildActivities` resolves paths/skills from reference data; the pure math stays unit-tested in `ProficiencyXpCalculator`.

### Nits
- The `amount > 0` guard in `AddEvent` isn't directly exercised. `ReflectedDamage_IsNotYetWired...` sets crit/dodge/heal all positive (so it only proves Reflect is unwired), and the offense-only tests never seed an event-keyed path. A small test seeding a Crit-keyed path with `CriticalDamageDealt = 0` and asserting it trains nothing would lock in the guard. Optional — take it or leave it.

### What's good
Tight, well-scoped change that reuses the existing routing loop rather than adding a parallel code path, accurate doc updates, and a thoughtful negative test for the deferred Retribution avenue.

---

Issue #1334, opened 2026-06-27
Mitigation rework: UI / combat log + reflection authoring


ginderjeremiah:
Part of #1330. See `docs/spikes/1330-mitigation-rework.md` (Area C).

Surface the reworked mitigation to the player and confirm authoring. Frontend-led.

## Scope
- Combat-log line for **reflected damage**; drop the block line.
- Attribute-breakdown screen: surface `Armor` and `DamageReflection`; remove the block stats.
- Verify gear / mod / proficiency / class authoring can grant `DamageReflection` (authored-only — needs an authoring path, no derivation).

## Tests / acceptance
- Combat-log renders reflected-damage lines; no block lines remain.
- Breakdown shows `Armor` / `DamageReflection`; block stats gone.

**Depends on:** Area A, Area B.
**Scope:** small

Issue #1333, opened 2026-06-27
Mitigation rework: remove Block, add deterministic damage reflection


ginderjeremiah:
Part of #1330. See `docs/spikes/1330-mitigation-rework.md` (Area B).

Retire Block (redundant with Defense; RNG defense is anti-tank) and repurpose the slot into deterministic, authored-only damage reflection. Done together so the defensive slot is never empty (no intermediate state where tanks are strictly worse).

## Scope — remove Block
- Retire `BlockChance` / `BlockReduction` (enum entries, `StaticAttributeModifiers` derivations, the `TakeDamage` block param, `SkillActivation.blocked` flag, combat-log block line).
- Remove the **block draw** from the #178 seeded RNG order — an enemy attack now draws **dodge only** (1 draw, not 2).

## Scope — add reflection
- Add `DamageReflection` (%) — **authored-only** (no core-attribute derivation; grantable by gear/mod/proficiency/class).
- On a **direct hit**, after mitigation, deal `net × defender.DamageReflection` to the **attacker**, **bypassing the attacker's mitigation**. Deterministic (no proc). Scoped to direct hits — DoT is not reflected.

## Tests / acceptance (parity-critical)
- New parity vectors: reflection on a direct hit, reflection ignores DoT, and the **simplified draw order** (enemy attack = 1 dodge draw; confirm crit/dodge still align).
- Vectors confirming block no longer affects any outcome.

**Depends on:** Area A (shared mitigation path / `TakeDamage`). Coordinate with #1320 — whichever of this / #1324 lands second removes the flat-floor absorption branch.
**Scope:** medium

PR #1356, opened 2026-06-29
Mitigation rework A: Defense → diminishing Toughness curve (#1332)

### LINKED ISSUE #1332: Mitigation rework: Defense → diminishing armor curve
Part of #1330. See `docs/spikes/1330-mitigation-rework.md` (Area A).

Replace flat `Defense` subtraction with a diminishing-curve percentage.

## Scope
- Mitigation uses `defReduction = Armor / (Armor + K·attackerLevel)` applied as a **multiplier**, on both sides (`Battler.TakeDamage` ↔ `applyDefense`). Thread the **attacker's level** into the mitigation call.
- Rework the `Defense`/`Armor` derivation in `StaticAttributeModifiers` (Endurance → Armor; mirrors to the client via codegen). Pick a strawman `K` so a level-appropriate investment sits in a sensible band (~30–60%), stable as content scales.
- Applies to both battlers (the curve uses the attacker's level).

## Notes
- Linear effective-HP (constant % EHP per Armor point), asymptotes <100% (no immunity, no breakpoint).
- Coordinate with #1320: this is the multiplicative mitigation step; if the damage-types resistance step (#1324) has landed, the curve composes after it. The flat-floor / absorption branch is removed once no flat subtraction remains (see Area B / #1320).

## Tests / acceptance (parity-critical)
- New/updated parity vectors for the curve on both sides; re-author the scenarios that asserted flat-Defense outcomes.
- Unit tests for the curve (EHP-linearity sanity, asymptote <100%, `K·level` scaling).

**Depends on:** — (foundation of this spike).
**Scope:** medium

ginderjeremiah:
## Summary

Implements **Area A** of the flat-mitigation rework spike (#1330). Replaces the scale-dependent flat `Defense` subtraction with a **diminishing-returns percentage** applied as a multiplier on both battlers:

```
net = dealt × (1 − Σ typeResist)                              // damage-type resistance (unchanged)
    × (1 − Toughness / (Toughness + K·attackerLevel))         // the new Toughness curve
    − blockReduction                                          // floored at 0
```

The old flat step had `reduction ÷ hit-size` value — overpowering against small hits, worthless against big ones, with a degenerate immunity breakpoint at the 0 floor. The curve fixes this: **effective HP is linear in Toughness** (each point is worth a constant % EHP, never useless) while the % reduction itself diminishes and **asymptotes below 100%** (no immunity, no breakpoint). `K` scales on the **attacker's** level so a level-appropriate investment stays in a sensible band as both Toughness and incoming damage grow with content.

## Changes

- **Stat rename + derivation.** `EAttribute.Defense` → `Toughness`, derived from **Endurance only** (`Toughness = 2·Endurance`). Agility no longer feeds it — its defensive identity is Dodge. ("Armor" was a working term; settled on Toughness since the stat is pure bulk, not equipment.)
- **Curve constant.** New `GameConstants.ToughnessMitigationConstant` (`K = 20`, client-mirrored), scaling the denominator on the attacker's level.
- **Attacker level threaded** into the mitigation call on both sides — `Battler.TakeDamage`/`ComputeNetDamage` ↔ `mitigateDamage`/`toughnessMitigatedDamage` — sourced from the active attacker (`BattleContext` / `battleStep`).
- **EF migration** renames the seeded attribute; **codegen regenerated** (`enums.ts`, `game-constants.ts`, `attribute-modifiers.ts`).
- **Tests re-authored.** Every parity vector and unit test that asserted flat-Defense outcomes updated on **both** sides with identical scenarios/results; added curve coverage (EHP-linearity, asymptote < 100%, K·level scaling). Scenarios where the curve fundamentally changed behaviour (e.g. boss full-loadout, high-mitigation trickle, the old "Defense clamps to zero" case) were redesigned rather than patched.
- **Docs.** `game-design.md` and `backend-battle.md` updated for the curve and attacker-level threading.

## Out of scope

Block removal (#1333) and the UI/combat-log/reflection surface (#1334) are **not** in this PR — `Block` still functions here. A dedicated `Toughness` icon is deferred to #1334 (it currently reuses the former `Defense` art).

## Testing

- Backend: `Game.Core.Tests` 1005, `Game.Application.Tests` 632, `Game.Infrastructure.Tests` 60, `Game.Api.Tests` 646 (includes the migration integration + codegen-drift checks) — all pass.
- Frontend: 2970 unit tests pass; `eslint` and `svelte-check` clean.
- Codegen produces no drift.

Closes #1332

---
_Generated by [Claude Code](https://claude.ai/code/session_01EmEWZ8j7WQePQ1JVhLmc8g)_


UI/src/lib/battle/battle-formulas.ts
```diff
@@ -43,10 +44,22 @@ export function skillContributions(skill: ISkill, attributes: BattleAttributes):
...
+ *  bit-for-bit for battle parity. */
+export function toughnessMitigatedDamage(
+	rawDamage: number,
+	toughness: number,
+	attackerLevel: number,
+	blockReduction = 0
+): number {
+	const scaled = TOUGHNESS_MITIGATION_CONSTANT * attackerLevel;
+	const toughnessReduction = toughness / (toughness + scaled);
+	return Math.max(rawDamage * (1 - toughnessReduction) - blockReduction, 0);
```
ginderjeremiah: Nit (optional, latent — not reachable today): when `toughness == 0` **and** `attackerLevel == 0`, `scaled` is 0 and `toughness / (toughness + scaled)` is `0/0 = NaN` on both sides. The two sides then diverge: C#'s `ComputeNetDamage` returns via `net > 0 ? net : 0`, so `NaN` falls to `0`; here `Math.max(NaN - blockReduction, 0)` returns `NaN` (`Math.max` propagates `NaN`). That's a bit-for-bit parity break in the exact spot the doc calls "must match bit-for-bit."

It isn't reachable in real play — battler `Level` is always ≥ 1, so `scaled ≥ K` and the `toughness == 0` branch is a clean `0/K = 0`. Given how hard this PR (rightly) leans on parity, consider either a short comment pinning the `level ≥ 1` invariant here, or guarding the `toughness == 0` case explicitly so the two implementations can't drift if a level-0 battler ever sneaks in. Non-blocking.

---

ginderjeremiah (commented):
## Review: Mitigation rework A — `Defense` → diminishing `Toughness` curve (#1332, Area A of spike #1330)

**Verdict: Approve** *(submitted as a comment because GitHub won't let me gate my own PR — there are no blocking or should-fix findings; one optional nit inline.)*

This cleanly delivers exactly Area A of the spike: the flat `Defense` subtraction becomes the multiplier `1 − Toughness / (Toughness + K·attackerLevel)`, derivation moves to Endurance-only, the attacker's level is threaded through `TakeDamage`/`ComputeNetDamage` ↔ `takeDamage`/`mitigateDamage`, and Block stays in scope while #1333/#1334 are correctly deferred. The acceptance criteria from #1332 (EHP-linearity, asymptote < 100%, K·level scaling, re-authored flat-Defense scenarios) are all met and the math checks out: EHP linear in Toughness, reduction asymptotes below 100% (no immunity/breakpoint), and the per-attacker-level denominator keeps the band stable as content scales.

### Parity
Strong. The frontend `toughnessMitigatedDamage`/`mitigateDamage` mirror the backend expression exactly (`mitigated × (1 − reduction) − blockReduction`, floored at 0), `K = 20` is shared via codegen (`game-constants.ts` ← `GameConstants`), and the unit scenarios are mirrored value-for-value on both sides (same Toughness/level inputs, same EHP-linearity and asymptote assertions). Resistance-then-Toughness ordering, the absorption (resistance > 1) heal path, and the DoT "bypasses mitigation" rule are all preserved consistently across both simulators. The parity matrices (`BattleSimulatorParityTests` / `battle-simulation-parity.test.ts`) and the curve unit tests are re-authored rather than patched, which is the right call where behavior fundamentally changed.

### Nits
- `battle-formulas.ts:62` — a latent, **non-reachable** parity divergence when `toughness == 0` *and* `attackerLevel == 0` (`0/0 = NaN`; C# resolves to 0 via its `> 0` ternary, JS's `Math.max(NaN, …)` returns `NaN`). Guarded by `Level ≥ 1` in real play; flagged only because the doc holds this exact spot to bit-for-bit parity. Inline detail above — optional.

### Test coverage
Meets the project's bar. Backend: dedicated curve cases (`TakeDamage_AppliesToughnessMitigationCurve`, `…NoToughness_LeavesHitUnreduced`, `…ToughnessNeverFullyMitigates`, K·level scaling, `…EffectiveHpIsLinearInToughness`, resistance-before-Toughness) plus the regenerated parity vectors. Frontend mirrors each scenario in `battler.test.ts` and `battle-formulas.test.ts`. The migration's integration + codegen-drift checks are covered per the test plan. No untested domain logic spotted.

### What's good
- Faithful, minimal-surface implementation that does *only* Area A and resists scope creep into Block/UI.
- Docs (`game-design.md`, `backend-battle.md`, spike status) are updated accurately and concisely — the new "Mitigation: the Toughness curve" section is exactly the right altitude.
- The `Toughness` art deliberately reuses the former `Defense` icon with a comment pointing at #1334, rather than silently — good signposting.
- Comment hygiene: every "Defense"→"mitigation/Toughness" doc-comment was updated in lockstep, including the subtle DoT-floor rationale.

Not merging yet: the **Run tests** CI workflow hasn't completed on this head (status is still `pending`). Once it's green this is good to merge as-is — the inline nit is optional and need not block.

---

UI/src/lib/battle/battle-formulas.ts
ginderjeremiah: Addressed in 74f77a8. Rather than guard the `toughness == 0` case or just pin the invariant in a comment, I swapped the clamp to the backend's exact `net > 0 ? net : 0` ternary (dropping `Math.max`). The two operators agree for every reachable input, but only the ternary floors a `NaN` to `0` — so this makes the function bit-for-bit with `ComputeNetDamage` on *all* inputs, including the unreachable 0/0, with no special-case branch. Left a comment pinning the `Level ≥ 1` rationale too.

---

PR #1377, opened 2026-06-29
Drive battle loading cooldown off the logical clock (#1366)

### LINKED ISSUE #1366: Battle loading cooldown runs on the render (rAF) clock — parks the farm loop in a backgrounded tab

## Observation

The post-victory loading/cooldown countdown is driven by the **render** loop, not the catch-up-capable logical loop:

- `UI/src/lib/engine/battle/battle-engine.ts:263-284` (`startLoading`) hooks `onRenderUpdate` and subtracts the **render** delta from `loadingTime` each frame, resolving the returned promise when it hits zero.
- The idle farm loop awaits that promise before spawning the next enemy (`getNewEnemy`).

`requestAnimationFrame` is throttled/suspended while a tab is backgrounded (the render loop's own `rAF` pauses), so during the loading stage the countdown **freezes** and the whole farm loop parks in `Loading` until the tab is refocused. The logical loop has explicit catch-up for over-budget time, but the loading countdown does not — it's tied to `rAF`.

## Why it matters

A backgrounded character that is between battles (in the loading window) stalls there rather than continuing to accrue idle progress, and idle time spent in `Loading` while hidden is lost. This is a concrete, independently-fixable contributor distinct from the broader tick-clamp analysis in the background-throttle spike.

## Suggested direction

Drive the loading countdown off the logical clock (which already handles catch-up and lost-time accounting) instead of the render delta, so the cooldown advances consistently whether or not the tab is focused.

## Related

- #1073 (spike: recover idle time lost to background-tab throttling) — this is a narrower, separately-fixable defect; could be folded into that spike's scope, but does not require its full resume machinery.

_Labels: bug, claude, scope: small._

ginderjeremiah:
## Summary

Fixes #1366. The post-victory loading/cooldown countdown was driven by the **render (rAF) loop** rather than the catch-up-capable logical loop:

- `BattleEngine.startLoading` hooked `onRenderUpdate` and subtracted the **render** delta from `loadingTime` each frame.
- `requestAnimationFrame` is throttled/suspended while a tab is backgrounded, so during the loading window the countdown **froze** and the idle farm loop parked in the `Loading` stage until the tab was refocused — losing idle progress accrued between battles.

## How it addresses the issue

`startLoading` now drives the countdown off `onLogicalUpdate` (the `setInterval`-based logical clock), which keeps running — throttled, with its existing over-budget catch-up — when the tab is hidden. The cooldown now advances consistently whether or not the tab is focused, so the farm loop no longer parks in `Loading`.

The related comments (`finishLoading` doc, the `stop()` unconditional-cancel rationale) were updated from "render" to "logical" to match.

### Scope note

This is the narrow, independently-fixable defect from #1366. Time lost *beyond* the logical loop's `tickSizeX5` catch-up cap (the broader background-throttle accounting) remains the concern of spike #1073 and is out of scope here.

## Test plan

- [x] Updated the `startLoading` unit tests to exercise the **logical** hook (delta-only signature) and renamed the countdown-down test accordingly.
- [x] `battle-engine.test.ts` — 48 passed.
- [x] Full frontend unit suite — 268 files / 3012 tests passed.

Frontend-only change; no backend surface touched. Not parity-sensitive (the loading countdown is outside `battleStep`).

Closes #1366.

https://claude.ai/code/session_01BWi6ZNc17JPsscUZ2LaU1z

---
_Generated by [Claude Code](https://claude.ai/code/session_01BWi6ZNc17JPsscUZ2LaU1z)_

### DIFF
--- a/UI/src/lib/engine/battle/battle-engine.ts
+++ b/UI/src/lib/engine/battle/battle-engine.ts
@@ -261,7 +261,7 @@ export class BattleEngine {
 	}
 
 	public startLoading(loadingTime: number) {
-		// Cancel any countdown still in flight so re-invoking can't leak the previous render hook.
+		// Cancel any countdown still in flight so re-invoking can't leak the previous logical hook.
 		this.finishLoading?.();
 		this.loadingTime = loadingTime;
 		this.loadingTotal = loadingTime;
--- a/UI/src/lib/engine/battle/battle-engine.ts
+++ b/UI/src/lib/engine/battle/battle-engine.ts
@@ -88,7 +88,7 @@ export class BattleEngine {
 	private logicalUnhook?: Action;
 	private renderUnhook?: Action;
 	private enemyLoadedUnhook?: Action;
-	/** Tears down the in-flight post-victory loading countdown (removes its render hook and resolves the
+	/** Tears down the in-flight post-victory loading countdown (removes its logical hook and resolves the
 	 *  awaited promise); undefined when no countdown is active. */
 	private finishLoading?: Action;
 
@@ -143,8 +143,8 @@ export class BattleEngine {
 			// no progress on their own. Resetting here lets a return from the admin screen resume cleanly.
 			this.setBattleStage(Idle);
 		}
-		// A loading countdown is driven by the render engine independently of `running`, so cancel it
-		// unconditionally — otherwise its render hook leaks and the awaiting getNewEnemy path hangs.
+		// A loading countdown is driven by the logical engine independently of `running`, so cancel it
+		// unconditionally — otherwise its logical hook leaks and the awaiting getNewEnemy path hangs.
 		this.finishLoading?.();
 	}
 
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Drive the battle loading cooldown off the logical clock (#1366)

**Intended verdict:** Approve (posted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author). One trivial nit, non-blocking.

The fix is correct and well-scoped. Switching the post-victory cooldown from `onRenderUpdate` to `onLogicalUpdate` is exactly the right mechanism: the logical loop (`setInterval`-driven, with the `tickSizeX5` catch-up) keeps advancing in a backgrounded tab where `rAF` is suspended, so the farm loop no longer parks in `Loading`.

A few correctness points I checked and confirmed:

- **Delta units match.** `notifyLogicalUpdate` always fires with `tickSize` (40ms), in the same millisecond unit as `loadingTime`, so `loadingTime -= delta` is sound. During catch-up, multiple `tickSize` ticks fire in one poll and drain the cooldown correctly before `finishLoading` unhooks.
- **Dropping the `false` arg is fine.** The removed second argument was `cleanupOnDestroy`, which defaults to `false` — so behavior is unchanged. Leaving it off is also the documented-correct choice here, since `startLoading` is reached outside component init and manages its own teardown via `unhook`.
- **Not parity-sensitive.** The loading window sits outside `battleStep`, so there's no backend mirror to keep in sync — correctly a frontend-only change.

### Nits

- `battle-engine.ts:264` — the comment `"…can't leak the previous render hook."` still says "render"; it should read "logical hook" to match the other comments this PR updated. It's just outside the diff hunk so I couldn't anchor it inline. Trivial consistency miss.

### Test coverage

Solid. The `startLoading` suite was correctly migrated to the logical hook (delta-only signature, `logicalUpdateCallbacks`), and the existing mock already exposed `onLogicalUpdate`. The four scenarios — countdown-to-zero, stop mid-loading, reset mid-loading, and re-invoke-while-pending — all still assert the right hook-leak invariants against the new clock. No additional coverage needed.

### What's good

Tight, minimal diff that fixes the real defect without reaching into the broader background-throttle accounting (correctly deferred to spike #1073). The accompanying comment updates explain *why* the logical clock is used, which is the part a future reader needs.


---

Issue #1189, opened 2026-06-24
Proficiency attribute-bonus modifier ordering diverges FE/BE (latent battle-parity bug)


ginderjeremiah:
Found during a code-health review. A **latent parity landmine** in the proficiency attribute-bonus pipeline (#982 area E): the frontend and backend insert the proficiency modifiers at a *different position relative to the static modifiers*, so the additive accumulation order differs. It is dormant only because no proficiency content is authored yet; it activates the moment an additive proficiency bonus on a derived attribute ships.

## The divergence

Both sides feed the same modifiers into one composition path, but in a different order:

- **Backend** — `BattleSnapshot.GetModifiers` (`Game.Core/Battle/BattleSnapshot.cs:135-156`) concatenates the proficiency modifiers onto the end of the base set (stat allocations + item attrs), and `AttributeCollection` (`Game.Core/Attributes/AttributeCollection.cs:23-31`) then appends `StaticAttributeModifiers.All` **last**. Net additive order on a node: `base → proficiency → static`.
- **Frontend** — `BattleAttributes.setData` builds `this.#modifiers = [...base, ...STATIC_ATTRIBUTE_MODIFIERS]` (`UI/src/lib/battle/battle-attributes.ts:70`), and `Battler.reset` then **appends** the proficiency `additionalModifiers` via `addModifier` (`UI/src/lib/battle/battler.ts:323-325`). Net additive order: `base → static → proficiency`.

For an **additive** proficiency modifier landing on an attribute that *also* carries a **static additive base** — MaxHealth, CooldownRecovery (base 1), CriticalDamage (base 1.5), BlockReduction (base 2), Defense — the two sides sum the same additive terms in a different order. Float addition is non-associative, so `((0 + Δprof) + base)` and `((0 + base) + Δprof)` can differ in the last bit, producing a ULP-level attribute difference that can cascade across ticks into a different damage number, a shifted kill tick, and ultimately a **failed anti-cheat replay**.

## Why the parity tests don't catch it

`ProficiencyBonusParityTests` / `proficiency-bonus-parity.test.ts` deliberately assert only on **core** attributes, which have no static modifiers — so the one case that diverges (an additive proficiency bonus on a *derived* attribute) is outside their matrix.

## Fix

[...]

PR #1383, opened 2026-06-29
Make frontend BattleAttributes recompute incrementally (#1364)

### LINKED ISSUE #1364: Frontend battle: per-modifier full attribute recompute on the hot path

## Observation

The frontend `BattleAttributes` rebuilds the **entire** per-attribute total array from scratch every time a single modifier is added or removed, whereas the backend caches per-attribute and invalidates lazily.

- `UI/src/lib/battle/battle-attributes.ts:145` (`recompute()`) allocates a fresh `number[]` of size `attributesMaxId + 1` and re-runs `computeAttributes(this.#modifiers)` over **all** modifiers on every `addModifier`/`removeModifier`. With derived attributes enabled it also allocates a `Map`. This runs per effect application and per effect expiry, for both battlers.
- The backend equivalent (`AttributeCollectionNode`) caches per-attribute values and invalidates only the changed attribute and its derived dependents — the docs explicitly describe the backend side as having "no per-tick allocations."

A second, related instance: `UI/src/lib/engine/battle/battle-engine.ts` (`damageLogMessage`, ~lines 303-312) calls `classifyResist(resistanceTotal(...))` on **every** player/enemy fire, re-deriving the defender's resistance from the attribute graph each time rather than reading a memoized per-battle value. This is correctly kept out of `battleStep` for parity, but it re-walks modifiers on the hottest path.

## Why it matters

This is the frontend's main per-tick allocation/CPU cost in heavy effect/DoT/buff-stacking battles. It is **not** a parity concern (the backend math is independently verified) and the live game runs one battle at a time, so impact is modest today — but it is a real, bounded optimization target if combat perf ever shows up in profiling, and it is the one place the FE diverges from the BE's "cheap incremental recompute" model.

## Suggested direction

- Targeted recompute: invalidate/recompute only the changed attribute plus its derived dependents, mirroring the backend's lazy cascade, instead of rebuilding the whole array.
- For `resistanceTotal` in the damage-log path, memoize the per-type resistance per battler and invalidate on modifier change.

Not parity-sensitive (display/perf only). _Labels: tech debt, claude, scope: medium._

ginderjeremiah:
## Summary

Closes #1364.

The frontend `BattleAttributes` previously rebuilt **every** attribute total from the full modifier list on each `addModifier`/`removeModifier`, running `computeAttributes` — which also allocates a breakdown-line object per modifier — over **all** modifiers on the per-effect hot path. The backend equivalent (`AttributeCollectionNode`) caches per attribute and invalidates only the changed attribute and its derived dependents. This change mirrors that lazy cascade on the frontend.

### What changed

- **`battle-attributes.ts`** — modifiers are now bucketed per attribute, with a `source → dependents` map (the same shape as `AttributeCollectionNode.DerivedNodes`). `addModifier`/`removeModifier` recompute **only the changed attribute and the attributes that transitively derive from it**, instead of rebuilding the whole array from the full modifier list. So a battler with many stacked DoT/buff modifiers pays per change only for the attributes that actually moved.
- **`attribute-collection.ts`** — new exported `foldAttributeValue(modifiers, valueOf)` kernel: the value-only counterpart of the `computeAttributes` per-attribute reduction the breakdown screen uses. `BattleAttributes` folds each single attribute through it, keeping the two in lockstep.
- The totals array is still **reassigned** on each recompute, so the documented reactivity contract and the lazy display-projection memoization are unchanged. The win is dropping the `O(all modifiers)` walk + per-modifier line allocation, not the small fixed-size array copy. Not a FE/BE parity change — display/perf only.

### Note on the issue's second point (`resistanceTotal`)

The issue also flags `classifyResist(resistanceTotal(...))` in the damage-log path as "re-deriving the defender's resistance from the attribute graph" each fire. On inspection this isn't actually a graph re-walk: `resistanceTotal` sums a few `getValue` calls, and `getValue` is already an `O(1)` read of the memoised totals array — it does not re-fold modifiers. The only avoidable work is `resistanceAttributes(type)` allocating a small fixed array per call; memoising the per-type resistance per battler would add invalidation complexity on every modifier change for negligible gain, so I deliberately left it. Happy to reconsider if profiling ever shows it.

## How it addresses the issue

Targets exactly the issue's primary ask — "invalidate/recompute only the changed attribute plus its derived dependents, mirroring the backend's lazy cascade, instead of rebuilding the whole array" — while preserving the eager-read reactivity rules that keep `getValue` a pure `$derived` read.

## Test plan

- [x] `npm run test:unit:ci` — full suite green, `src/lib/battle/**` coverage thresholds met
- [x] New **incremental-vs-full equivalence** test: every `addModifier`/`removeModifier` step is compared against a fresh `computeAttributes` over the same modifier set (10-decimal tolerance), exercising derived cascades (Strength/Endurance → MaxHealth, Luck → Critical*).
- [x] New derived-cascade hook/unhook coverage (link kept while another modifier still derives from the source; unhooked on the last removal) and circular-derived guard tests for both `foldAttributeValue`/`computeAttributes` and the incremental path.
- [x] Existing battle parity + reactive-read (`statify`) suites still green (329 battle tests).

https://claude.ai/code/session_01U29F8WqvhLuNWquPujvHo9

---
_Generated by [Claude Code](https://claude.ai/code/session_01U29F8WqvhLuNWquPujvHo9)_


ginderjeremiah (commented):
## Review: Incremental `BattleAttributes` recompute (#1364)

**Verdict:** Approve (posted as a comment — I'm the PR author by GitHub identity, so a gating review isn't accepted) — **one should-fix** before merge: a stale doc line. No blocking issues.

The change does exactly what the issue's primary ask describes: `addModifier`/`removeModifier` now recompute only the changed attribute plus its transitive derived dependents, via a `source → dependents` cascade that mirrors the backend `AttributeCollectionNode`/`DerivedNodes`. I verified the named backend counterparts (`AddModifierWithoutCacheInvalidation`, `UnhookDerivedLink`, the `DerivedNodes` cascade) exist and match the comments. Correctness looks solid:

- The per-attribute fold in `foldAttributeValue` is operation-identical to `computeAttributes`' inner reduction (additives in bucket order, then multiplicatives), and an unchanged derived source is read from its cached value — which is itself the deterministic result of that same fold. So the incremental path stays **bit-identical** to a full recompute, not just close — which matters because the FE/BE parity (anti-cheat replay, #1189) rests on it. The equivalence test guards this at every add/remove step.
- The `inProgress` guard in `#dirtyResolver` faithfully reproduces `computeAttributes`' circular-derived fallback, and the `dirty` guard in `#collectDependents` terminates and dedupes shared dependents.
- `#unlinkDerived` correctly keeps the cascade link alive while another modifier on the same attribute still derives from the source — matching the backend, and covered by a test.
- Display-only mode (`calcDerived === false`) keeps the old raw-sum semantics, and no dead code is left behind (`recompute()` fully replaced).

I also agree with the deliberate decision to **not** memoize `resistanceTotal`/`classifyResist` in the damage-log path. The issue framed it as "re-deriving from the attribute graph," but `resistanceTotal` only sums a few `getValue` calls and `getValue` is now an O(1) array read — so it isn't a graph re-walk, and the rationale in the PR body is correct.

### Should-fix

- **`docs/frontend.md:37` is now stale.** It still says `BattleAttributes` composes totals "through the same `computeAttributes` path the breakdown screen uses (memoised, recomputed only on add/remove)." After this PR it folds through the value-only `foldAttributeValue` kernel and recomputes **incrementally** (changed attribute + derived dependents), not via `computeAttributes`. CLAUDE.md treats the docs as the source of truth and asks that they be updated when a design like this changes — please refresh that paragraph (the updated class-level doc comment and test header are good source text). Couldn't anchor this inline since `frontend.md` isn't in the diff.

### Test coverage

Thorough and in the right place (`tests/` mirroring source):
- `foldAttributeValue` unit tests (additive→multiplicative order, derived scaling, unsupported-type throw).
- The incremental-vs-full equivalence test exercising real cascades (Strength/Endurance → MaxHealth, Endurance → Toughness, Luck → Critical*) across both add and remove — the key invariant.
- Derived-link hook/unhook across add/remove, circular-chain termination (both kernels), and the absent-bucket removal no-op.

The `toBeCloseTo(…, 10)` tolerance is consistent with the file's existing parity convention (`toBeCloseTo(…, 6)`), so no concern there.

### What's good

Clean mirror of the backend's lazy-cascade model, the reactivity contract (eager recompute + non-reactive `#` buckets preserving `removeModifier` reference identity) is preserved and well-explained, and `setData` now routes through the same kernel so there's a single value path. Nicely scoped and well-tested.

Not merging on this pass: the doc should-fix needs addressing first, and frontend CI is still in progress.

---

PR #1361, opened 2026-06-29
Synthesis screen: recipe-graph ("Web") visualization (#1346)

ginderjeremiah:
## Summary

Adds the alternate **"Web" view** to the Synthesis screen (spike #1125 area E) alongside the shipped **Bench** view. A **Bench / Web** toggle switches between the existing list → bench → dossier layout and a **pan/zoom recipe-graph canvas** — input skills → fusion nodes → result skills — over the same result dossier. Deeper recipes (a synthesized result feeding another recipe) are drawn as edges, surfacing the "recipe-graph chaining is the depth source" the spike calls for (decision 11).

## How it addresses the issue

| Issue ask | Change |
| --- | --- |
| A **Bench / Web** view toggle | `SynthesisView.mode` (`'bench'`/`'web'`) + a segmented toggle in the toolbar. |
| A graph canvas of the discovered recipe DAG with pan + zoom | `SynthesisGraph.svelte` paints SVG edges + positioned nodes and wires pointer-drag pan and wheel zoom; `SynthesisGraphNode.svelte` renders the input / fusion / result node kinds. |
| Reuse `buildSynthesis` reveal/gating state for node styling | The graph is derived from the **same** `RecipeView[]` the list uses (`SynthesisView.graph`), so node/edge styling reads each recipe's `ready`/`gated`/`hinted`/`done` state — the two views can never diverge. |
| Selecting a node drives the existing inspector/dossier | Clicking a fusion/result node calls the existing `select(recipeId)`, driving the shared `selectedRecipeId` + `ResultDossier` (leaf input nodes are inert). |
| Respect the conservative hinted reveal | Hidden recipes are already absent (dropped by `buildSynthesis`); a hinted recipe's fusion/result are **masked** (keyed by recipe id, not skill, so no identity leaks) and only its *owned* inputs become nodes. |
| Pure layout/edge math in a framework-free module | `synthesis-graph.ts` (`layoutSynthesisGraph` + `zoomAt`/`clampScale`), unit-tested without rendering, mirroring the Bench view's `synthesis.ts` split. |

## Notable design decisions

- **Layered-DAG layout.** The mock used hardcoded coordinates; this derives a real layout — longest-path layering (leaf inputs → fusion → result, deepening with each chain) with a predecessor-barycenter in-layer ordering to reduce crossings, normalized to a padded, centred extent. Each *skill* is a single node, so a result reused as an input is one node with onward edges (the chaining).
[...]

Issue #1346, opened 2026-06-29
Synthesis screen: recipe-graph ("Web") visualization


ginderjeremiah:
Follow-up to #1300 (PR #1345), part of the skill-synthesis spike #1125 (area E).

The Synthesis screen landed with the **Bench** view (recipe list → bench → result dossier), which fully satisfies the issue's functional scope. The Hi-Fi design mock (`docs/designs/game/skill-synthesis`) also includes an alternate **"Web" view**: a pan/zoom recipe-graph canvas that visualizes the synthesis chain — input skills → fusion nodes → result skills, with deeper recipes (a synthesized result feeding another recipe) drawn as edges. The spike calls recipe-graph chaining "the depth source," so a graph explorer is a natural way to surface that depth.

### Scope
- A **Bench / Web** view toggle on the Synthesis screen.
- A graph canvas rendering the discovered recipe DAG (reusing the same `buildSynthesis` reveal/gating state for node styling: ready / gated / hinted / done), with pan + zoom.
- Selecting a node drives the existing inspector/dossier (reuse `RecipeView` + `ResultDossier`).
- Respect the conservative hinted reveal — a hinted result node stays masked (no leaked identity), hidden recipes are absent.

### Notes
- The synthesize **cascade animation** from the mock is intentionally *not* part of this — it overlaps the synthesis-feedback sub-issue #1301.
- Pure layout/edge math should live in a framework-free module (unit-tested without rendering), mirroring the Bench view's `synthesis.ts` split.

_Labels: enhancement, claude, scope: medium._

PR #1397, opened 2026-06-30
Enforce one skill = one producing recipe at recipe save (#1362)

### LINKED ISSUE #1362: Synthesis Web graph: alternate recipes producing the same skill collapse to one node
Follow-up surfaced during review of the Synthesis "Web" recipe-graph view (#1346 / PR #1361).

## Observation

In `synthesis-graph.ts`, the graph keys each *revealed* result by its **skill id** (`s{skillId}`), which is what enables recipe-graph chaining (a synthesized result reused as an input is one shared node). The trade-off: if **two non-retired recipes produce the same result skill**, `producedBy` keeps only the last one and `ensureSkillNode` returns a single `result` node tied to that last producer. Consequences in the Web view only:

- Selecting that result node drives only the **last** producer's dossier.
- The other recipe's `fusion → result` edge points at the same shared node (so both fusions appear to feed one result).

The masked (hinted) path is unaffected — hinted results are keyed per-recipe (`m{recipeId}`), so no collapse there. The Bench list is also unaffected (it renders one row per recipe).

## Decision needed

Is "multiple recipes producing the same result skill" a real authoring possibility, or is **one skill = one producing recipe** an intended invariant?

- If it's not allowed → consider an **authoring/validation guard** (admin Workbench) rejecting a duplicate result-skill across recipes, and this graph behavior is then correct by construction.
- If it *is* allowed → the Web view should disambiguate (e.g. a distinct node per producing recipe, or a node that lets the player pick which producer's dossier to view).

No user-facing impact today (proficiency/recipe content is still unauthored, #1127), so this is low priority — but worth pinning the invariant before recipe content lands.

Related: #1346, #1125 (decision 11 — recipe-graph chaining), PR #1361.

ginderjeremiah:
## Summary

Pins the authoring invariant **one skill = one producing (non-retired) `SkillRecipe`**, the owner-confirmed direction on #1362.

The synthesis "Web" graph (`synthesis-graph.ts`) keys each revealed result by its **skill id** (`s{skillId}`) — that's what enables recipe-graph chaining (a synthesized result reused as an input is one shared node). The trade-off, surfaced in review of #1346/PR #1361: if two non-retired recipes produce the same result skill, they collapse to a single node — only the last producer drives the dossier, and both fusions appear to feed one result.

Rather than make the Web view disambiguate, the owner confirmed enforcing the simpler invariant. This PR adds the authoritative guard at the recipe save.

## How it addresses the issue

- `AdminSkillRecipes.SaveSkillRecipes` now rejects (via `AdminSaveResult.Failure`, anti-tamper, before anything is staged) a change set that would leave **two non-retired recipes producing the same result skill** — sitting alongside the existing Synthesis-flag and acyclicity guards.
- The check builds the **prospective** live-result map: live recipes, with each `Edit` re-pointing/retiring its recipe and each `Add` contributing a new producer (Adds keyed by a batch-local sentinel since their ids are store-generated), then flags any result claimed more than once.
- **Retired recipes are inert** and don't count — retiring a producer in the same batch frees its result for a new recipe.
- With the invariant enforced, the `synthesis-graph.ts` collapse behavior is **correct by construction** — no frontend changes needed. A frontend advisory warning was intentionally not added: the recipe editor's own design (per its doc comment) defers cross-recipe **graph-level** checks like acyclicity to the backend save-failure toast, and duplicate-result is the same class of check.

## Documentation

`docs/backend.md` → Skill recipes: records the new invariant next to the acyclic-graph guard.

## Test plan

Added integration tests to `AdminSkillRecipesIntegrationTests` (all green; full suite 660 passing):

- Add producing a result an existing live recipe already produces → failure
- Two Adds producing the same result in one batch → failure
- Edit re-pointing a recipe's result to one another live recipe produces → failure
- Add duplicating a **retired** producer's result → succeeds (retired is inert)
- Retire the producer + Add the same result in one batch → succeeds

Closes #1362

---
_Generated by [Claude Code](https://claude.ai/code/session_01Khafwm5iWLavpthaiNubUa)_


Game.DataAccess/Repositories/Admin/AdminSkillRecipes.cs
```diff
@@ -198,6 +206,50 @@ private static Entities.SkillRecipeCondition ToConditionEntity(int recipeId, Con
...
+                {
+                    resultByRecipe[--addKey] = change.Item.ResultSkillId;
+                }
+                else if (change.ChangeType == EChangeType.Edit)
+                {
+                    if (change.Item.RetiredAt is not null)
+                    {
+                        resultByRecipe.Remove(change.Item.Id);
+                    }
+                    else if (_recipes.LookupSkillRecipe(change.Item.Id) is not null)
```
ginderjeremiah, line 234: Nit (test coverage): this `else if` branch is the only path not directly exercised by the added tests. `EditResultToOneAnotherLiveRecipeProduces` covers re-pointing an *already-live* recipe into a collision, but the *un-retire* sub-case — an `Edit` with `RetiredAt is null` on a currently-retired recipe whose result collides with a live producer — takes this same branch from a different starting state (the recipe isn't in `resultByRecipe` yet, so it's re-added here). The logic is sound and `LookupSkillRecipe` does resolve retired ids, so this is low risk, but a dedicated test (`UnRetireResultIntoCollision_ReturnsFailure`) would lock the branch in. Not blocking.

---

ginderjeremiah (commented):
## Review: Enforce one skill = one producing recipe at recipe save (#1362)

**Verdict:** Approve — implements exactly the owner-confirmed invariant (one skill = one producing, non-retired recipe) at the authoritative save boundary, with correct, well-tested logic. _(Submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author — read it as an approve.)_

The guard lands in the right place: alongside the existing Synthesis-flag and acyclicity checks in `AdminSkillRecipes.SaveSkillRecipes`, rejecting via `AdminSaveResult.Failure` before anything is staged. Confirming the invariant with the owner before building (rather than picking a direction unilaterally) was the right call given it was a *Decision needed* issue.

### Strengths

- **Prospective-map construction is correct and order-independent.** Building the live-result map and folding each `Edit` (re-point / retire / un-retire) and `Add` into it before checking for a collision handles the same-batch cases cleanly — e.g. retire-producer-and-add-same-result in one batch succeeds regardless of change order. Verified against `LookupSkillRecipe` (resolves retired ids) and `AllSkillRecipeEntities` (full snapshot).
- **Careful detail on zero-based ids.** The descending negative sentinel for `Add` keys (`--addKey` starting from 0 → -1, -2, …) deliberately avoids colliding with real zero-based recipe ids and with other Adds in the batch. The comment explains why, which a later "simplification" might otherwise miss.
- **No cache mutation.** The check projects into a fresh dictionary rather than touching the cached `AsNoTracking` entity graph — consistent with the reference-data gotchas in `backend.md`.
- **Correct-by-construction follow-through.** Leaving `synthesis-graph.ts` untouched is the right scope: with the invariant enforced at authoring, the skill-id keying collapse can no longer occur, so no Web-view disambiguation is needed. Deferring the graph-level check to the backend save-failure toast (rather than a frontend advisory) matches how acyclicity is already handled.
- **Docs updated concisely** — one sentence appended to the skill-recipes paragraph in `backend.md`, next to the acyclic-graph guard. Right altitude, not over-documented.

### Test coverage

Good coverage of the main paths via integration tests (consistent with how the sibling cycle/result-skill guards are tested): live-collision, in-batch duplicate, edit-into-collision, retired-producer-is-inert, and retire-and-re-add-in-one-batch. Error-message assertions match the produced text.

### Nits

- One branch — un-retiring a retired recipe directly into a collision (`AdminSkillRecipes.cs:234`) — isn't directly exercised; the existing edit-collision test enters that branch from the already-live state instead. Low risk, optional test suggested inline. Not blocking.

Nothing here blocks merge. CI was still running (test-backend / e2e in progress) at review time, so I'm leaving the merge to once it's green.

---

Issue #1342, opened 2026-06-28
Spike — Restrict skill usability by equipped weapon


ginderjeremiah:
Surfaced during the proficiency-progression-avenues spike (**#1318**, [`docs/spikes/1318-proficiency-progression-avenues.md`](../blob/main/docs/spikes/1318-proficiency-progression-avenues.md)) while settling **static weapon-typed martial skills**. With weapon types as damage-type leaves and martial skills statically typed to a weapon, a coherence question opens up that the project has never pinned down: how do **skills, the equipped weapon, and the loadout** relate?

## Explore
- Whether a weapon-typed skill should be **usable only with its matching weapon** (or weapon class) equipped — selection-time gating vs activation-time gating vs no gating (purely cosmetic mismatch).
- Interaction with the weapon's **granted signature skill**, the 4-skill loadout cap, and item-granted skills.
- **Weapon-swap behaviour** — auto-unequip mismatched skills, grey them out, or allow them?
- Whether elemental/caster skills are **weapon-agnostic** (a staff is not a "weapon damage" source) and how hybrid kits read.
- How gating interacts with **weapon masteries** (training Swordsmanship implies wielding a sword) and with the proficiency loadout opportunity cost.
- UX + authoring implications.

## Output
Documented directions and sub-issues if pursued. **Not a blocker for #1318** — static weapon typing functions without usability gating; this decides whether to add it.

ginderjeremiah:
Ideation complete — resolved by the spike doc in #1376 (merged): [`docs/spikes/1342-restrict-skill-usability-by-weapon.md`](../blob/main/docs/spikes/1342-restrict-skill-usability-by-weapon.md). Decision: pursue a non-destructive, setup-time weapon-skill usability gate. Implementation is tracked in the linked sub-issues #1372 (prerequisite) → #1373 → #1374, with #1375 trailing for docs. Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1399, opened 2026-06-30
Add BattleEngine vs BattleSimulator parity smoke test (#1365)

### LINKED ISSUE #1365: Battle parity suite only exercises the headless BattleSimulator, not the live BattleEngine

## Observation

The frontend battle-parity suite drives the production headless `BattleSimulator`, which is the right call — but the **live** `BattleEngine.logicalUpdate` (the render-driven loop the player actually runs) is not directly under the parity matrix. Both consumers call the shared `battleStep` (`UI/src/lib/battle/battle-step.ts`), and `logical-engine.ts` guarantees fixed `tickSize` deltas, so today the live loop matches the simulator (and therefore the backend) **by construction**.

## Why it matters

The match holds only as long as no per-tick logic is added to `BattleEngine.logicalUpdate` *outside* `battleStep`. Any future tick-level behavior added directly to the live loop (rather than to the shared stepper) would silently escape the parity matrix and could diverge from the backend replay without any test failing. This is a latent risk, not a current bug.

## Suggested direction

Add a smoke test that drives `BattleEngine` over one or two representative parity scenarios and asserts the same outcome as the headless `BattleSimulator` for the same seed/inputs — cheap insurance that the two consumers of `battleStep` can't drift.

ginderjeremiah:
## Summary

Closes #1365.

The frontend battle-parity suite (`battle-simulation-parity.test.ts`) drives only the headless `BattleSimulator`. The **live** `BattleEngine.logicalUpdate` — the render-driven loop the player actually runs — shares the same `battleStep`, so today it matches the simulator (and the backend replay) *by construction*. That match holds only as long as no per-tick logic is added to `logicalUpdate` **outside** `battleStep`; any future tick-level behavior added directly to the live loop would silently escape the parity matrix and could diverge from the backend without failing a test.

This PR adds a smoke test that closes that gap: it runs the **live `BattleEngine`** over a representative slice of scenarios and asserts its terminal outcome (victory/defeat/draw + `totalMs`) equals what the `BattleSimulator` produces for the **same battlers and seed**. The expected result is derived from the simulator itself (not hardcoded), so the test directly pins "the two consumers of `battleStep` can't drift," with a cross-check against the documented matrix value as a transcription guard.

## How it addresses the issue

- New file `UI/src/tests/lib/engine/battle/battle-engine-parity.test.ts`.
- Configures the engine's mocked derivation inputs (player manager / inventory / proficiency seams held to no-ops) so the engine derives battlers identical to ones built directly for the simulator from the same registered skills, raw allocations, level, and seed.
- Drives the engine with fixed `tickSize` logical updates until the battle leaves `Active`, mirroring the live loop, then compares the outcome to `new BattleSimulator(...).simulate()`.
- Covers all three terminal outcomes, reusing rows from the existing simulator matrix:
  - `drawOrderDodgeOnlyAlignsCrit` — seeded fractional crit (depends on the exact RNG stream) with the enemy firing back → victory.
  - `dotRegenNetNegativeKillsAfterHeal` — net-negative DoT/HoT, no direct hits → player defeat.
  - `highToughnessTrickle` — both sides out-tank the trickle → timeout draw.

## Test plan

- [x] `npx vitest run src/tests/lib/engine/battle/battle-engine-parity.test.ts` — 3/3 pass.
- [x] `npx vitest run src/tests/lib/engine/battle/ src/tests/lib/battle/` — 489/489 pass (no regressions).
- [x] `npm run check` — 0 errors, 0 warnings.
- [x] `npx eslint` on the new file — clean.

## Notes

- This is test-only; no production code changes.
- The scenarios deliberately use weapon-agnostic (`Physical`) player skills and an ungated player so the engine-derived battlers match the simulator battlers exactly; the weapon-match gate (#1342) is exercised separately in the simulator matrix.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FdFASEMRSbRCFhTxJM3aqs)_

### DIFF
--- a/UI/src/tests/lib/engine/battle/battle-engine-parity.test.ts
+++ b/UI/src/tests/lib/engine/battle/battle-engine-parity.test.ts
@@ -1,0 +1,255 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { EAttribute, type IBattlerAttribute, type IEnemy, type ISkill } from '$lib/api';
+import { DEFAULT_MAX_BATTLE_MS } from '$lib/api/types/game-constants';
+
+// Both consumers of the shared `battleStep` — the live, render-driven `BattleEngine.logicalUpdate` and the
+// headless `BattleSimulator` — must produce the same outcome for the same inputs and seed, since the backend
+// anti-cheat replay drives the simulator path. The existing battle-simulation-parity suite only exercises the
+// simulator; this smoke test runs a few representative scenarios through the LIVE engine and asserts it agrees
+// with the simulator, so any per-tick logic added to the live loop OUTSIDE `battleStep` (which would silently
+// escape the parity matrix) fails here instead (#1365).
+
+// The shared skill registry the mocked `staticData.skills` returns; both the engine's player/enemy battlers and
+// the comparison simulator battlers resolve their skills out of it by id, exactly as the live game does.
+const mockSkills: ISkill[] = [];
+const mockEnemies: IEnemy[] = [];
+const mockAttributes: { id: number; name: string }[] = [{ id: 0, name: 'Strength' }];
+
+// Captured logical-update hook — driving the engine means invoking this with fixed `tickSize` deltas.
+type DeltaCallback = (delta: number) => void;
+type EnemyLoadedCallback = (enemy: import('$lib/api').IEnemyInstance) => void;
+let logicalUpdateCallbacks: DeltaCallback[] = [];
+let enemyLoadedCallbacks: EnemyLoadedCallback[] = [];
+
+// The player's derivation inputs (#811): attributes / loadout / level, plus the class + proficiency modifier
+// seams. Held to no-ops here so the engine's player battler matches a simulator battler built from the raw
+// allocations alone — a flat additive-0 signature passive leaves every attribute untouched (raw enum values:
+// type 1 = Additive, source 9 = Class), and the locked-base / proficiency modifier lists are empty.
+const mockPlayerManager = {
+	name: 'TestPlayer',
+	level: 1,
+	selectedSkills: [] as number[],
+	attributes: [] as IBattlerAttribute[],
+	battleLockedBaseModifiers: [] as unknown[],
+	battleSignaturePassiveModifier: () => ({ attribute: 0, amount: 0, type: 1, source: 9 })
+};
+// The weapon-match gate's equipped weapon type is left undefined so the player is ungated, matching the
+// comparison simulator battlers (which field their full loadout); the scenarios use weapon-agnostic skills.
+const mockInventoryManager = {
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: BattleEngine vs BattleSimulator parity smoke test (#1365)

**Verdict:** Approve _(submitted as a comment since I'm the PR author; GitHub won't accept a gating review on one's own PR)_ — the change closes exactly the latent gap #1365 described, with no blocking, should-fix, or nit findings.

### What it does well

- **Tests the right consumer.** The existing parity matrix only drives the headless `BattleSimulator`; this runs the **live** `BattleEngine.logicalUpdate` over representative scenarios and asserts its terminal outcome equals the simulator's for the same battlers/seed — so per-tick logic added to the live loop *outside* `battleStep` can no longer silently escape parity.
- **Reference outcome is derived, not hardcoded.** `simResult` comes from `new BattleSimulator(...).simulate()`, and the documented matrix value is kept only as a `toEqual(scenario.expected)` transcription cross-check. The real assertion (`runEngine(engine)` ≡ `simResult`) is what pins the two consumers together.
- **Covers all three terminal outcomes** — a seeded fractional-crit victory with the enemy firing back (`drawOrderDodgeOnlyAlignsCrit`), a net-negative DoT/HoT defeat (`dotRegenNetNegativeKillsAfterHeal`), and a timeout draw (`highToughnessTrickle`) — each matching its row in `battle-simulation-parity.test.ts` exactly.

### Correctness checks I verified

- **Timing parity holds.** The engine accumulates `timeElapsed` *before* `battleStep` and checks the cap *after*; the simulator increments `totalMs` at the loop tail and checks at the head. Both run the same number of `battleStep` calls before declaring a draw and both report `totalMs = DEFAULT_MAX_BATTLE_MS`. `runEngine`'s `maxTicks` guard mirrors the simulator's cap and surfaces a non-terminating battle as a thrown error rather than hanging.
- **Mock seams are sound.** `tickSize: 40` matches the real `MS_PER_TICK` (and the BattleSimulator reads the same mocked value, so both stay consistent); `battleSignaturePassiveModifier` returns an additive-0 `Class`-source modifier (`EModifierType.Additive = 1`, `EAttributeModifierSource.Class = 9`) that is a true no-op, so the engine-derived player battler is equivalent to the simulator one built from raw allocations; the ungated `equippedWeaponType: undefined` + weapon-agnostic skills keep the weapon-match gate (#1342) out of scope, as the PR notes.

### Test coverage

Appropriate for the issue — a smoke test, as #1365 explicitly requested, exercising the live loop's victory/defeat/draw branches. No production code changed, so no further coverage is owed. Per `frontend.md`, the test lives under `tests/` mirroring the source path.

Nice, targeted insurance against a real future foot-gun. Merging.


---

PR #1400, opened 2026-06-30
Cancel orphaned reference-cache reloads on timeout (#1367)

### LINKED ISSUE #1367: AdminCacheReloadFilter: reload queries are uncancellable and orphaned on timeout

## Observation

`Game.Api/Filters/AdminCacheReloadFilter.cs:72`:

```csharp
await Task.WhenAll(caches.Select(cache => cache.ReloadAsync())).WaitAsync(ReloadTimeout);
```

`ReloadAsync()` is invoked with **no cancellation token**, then merely bounded by `WaitAsync(ReloadTimeout)`. The surrounding comment deliberately decouples this from the request's cancellation token (correct — the write has committed, so the caches must refresh even if the client disconnected). The gap is the timeout path: `WaitAsync` only abandons the *await*. On timeout, every holder's underlying reload query keeps running to completion against its own scoped context, each holding its `_reloadGate` semaphore until it finishes. The admin sees a `TimeoutException` while a swap may still land afterward.

## Why it matters

A genuinely wedged DB connection means N reload queries continue in the background holding their per-holder gates after the admin request has already failed — a resource leak on the (rare) timeout path, plus a confusing "it timed out but the data updated anyway" outcome. `ReferenceCacheHolder.ReloadAsync` already accepts a token, so the fix is small.

## Suggested direction

Pass a **linked timeout token** (a `CancellationTokenSource(ReloadTimeout)`, independent of the request token) into `ReloadAsync(...)` so the queries actually cancel on timeout (Npgsql honors it) instead of being orphaned, while preserving the deliberate "not tied to the request token" behavior.

ginderjeremiah:
## Summary

`AdminCacheReloadFilter` bounded its post-write reload with `WaitAsync(ReloadTimeout)` but invoked `ReloadAsync()` with **no cancellation token**. `WaitAsync` only abandons the *await* — on timeout every holder's reload query kept running to completion against its own scoped context, each holding its `_reloadGate` semaphore until it finished. The admin saw a `TimeoutException` while a swap could still land afterward (a resource leak on the rare timeout path, plus a confusing "it timed out but the data updated anyway" outcome).

## How it addresses the issue

- Introduce a standalone, timeout-linked `CancellationTokenSource(ReloadTimeout)` — deliberately **independent of the request token** (the committed write must still land even if the client disconnected, preserving read-your-writes) — and pass its token into `ReloadAsync(...)`.
- The token now both **bounds the await** and **drives cancellation**, so on timeout the underlying queries and per-holder gates are actually released (`ReferenceCacheHolder.ReloadAsync` already threads the token to the gate wait and the EF/Npgsql query) instead of being orphaned in the background.
- The cancellation is re-surfaced as the same `TimeoutException` the admin response already expected, so the externally observed behavior is unchanged.
- The 30s timeout became an optional constructor parameter (defaults to 30s, resolved by DI since the type isn't registered) purely so the timeout path is testable without the full wait.

## Test plan

- [x] `dotnet build Game.Api.Tests`
- [x] Full `Game.Api.Tests` suite green (652 passed)
- [x] New unit tests on the filter:
  - `PassesACancellableToken_ToEachReload` — the reload runs under a cancellable token, not `CancellationToken.None`.
  - `ThrowsTimeoutAndCancelsTheReload_WhenAReloadWedges` — a wedged reload is cancelled on timeout (verified deterministically via a cancellation callback) and surfaces as a `TimeoutException`, rather than hanging or orphaning the query.
  - Existing broadcast/reload/skip-on-fault tests still pass.

Closes #1367

---
_Generated by [Claude Code](https://claude.ai/code/session_01S9PanivAfMBKT5xb14CfNM)_

### DIFF
--- a/Game.Api.Tests/Unit/AdminCacheReloadFilterTests.cs
+++ b/Game.Api.Tests/Unit/AdminCacheReloadFilterTests.cs
@@ -13,10 +13,29 @@ namespace Game.Api.Tests.Unit
     /// <summary>
     /// Covers the post-admin-write reload gate: on a successful write the filter broadcasts the change and
     /// reloads every cache; a broadcast failure is swallowed so the local read-your-writes reload still runs;
-    /// and a faulted action skips both the broadcast and the reload entirely.
+    /// a faulted action skips both the broadcast and the reload entirely; and a wedged reload is cancelled and
+    /// surfaced as a <see cref="TimeoutException"/> rather than orphaned.
     /// </summary>
     public class AdminCacheReloadFilterTests
     {
+        private static ActionExecutingContext BuildExecutingContext()
+        {
+            var httpContext = new DefaultHttpContext();
+            var actionContext = new ActionContext(httpContext, new RouteData(), new ActionDescriptor());
+            return new ActionExecutingContext(actionContext, [], new Dictionary<string, object?>(), controller: new object());
+        }
+
+        private static ActionExecutedContext BuildExecutedContext(Exception? exception, bool exceptionHandled)
+        {
+            var httpContext = new DefaultHttpContext();
+            var actionContext = new ActionContext(httpContext, new RouteData(), new ActionDescriptor());
+            return new ActionExecutedContext(actionContext, [], controller: new object())
+            {
+                Exception = exception!,
+                ExceptionHandled = exceptionHandled,
+            };
+        }
+
         private static async Task<(int notified, int reloaded)> RunAsync(
             Exception? exception, bool exceptionHandled, bool broadcastThrows = false)
         {
@@ -24,16 +43,8 @@ namespace Game.Api.Tests.Unit
             var notifier = new RecordingNotifier(broadcastThrows);
             var filter = new AdminCacheReloadFilter([cache], notifier, NullLogger<AdminCacheReloadFilter>.Instance);
 
-            var httpContext = new DefaultHttpContext();
-            var actionContext = new ActionContext(httpContext, new RouteData(), new ActionDescriptor());
-            var executing = new ActionExecutingContext(actionContext, [], new Dictionary<string, object?>(), controller: new object());
-            var executed = new ActionExecutedContext(actionContext, [], controller: new object())
-            {
-                Exception = exception!,
-                ExceptionHandled = exceptionHandled,
-            };
-
-            await filter.OnActionExecutionAsync(executing, () => Task.FromResult(executed));
+            var executed = BuildExecutedContext(exception, exceptionHandled);
+            await filter.OnActionExecutionAsync(BuildExecutingContext(), () => Task.FromResult(executed));
 
             return (notifier.NotifyCount, cache.ReloadCount);
         }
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Cancel orphaned reference-cache reloads on timeout (#1367)

**Verdict:** Approve — the fix matches the issue's suggested direction exactly, the cancellation genuinely propagates, and the new timeout path is covered by a deterministic test. (Posted as a `COMMENT` only because GitHub won't accept a gating verdict from the PR author; treat it as an approval.)

This correctly closes the gap described in #1367. `WaitAsync(timeout)` only abandoned the *await*, so on timeout each holder's reload query kept running against its own scoped context, holding its `_reloadGate` semaphore until it completed. Passing a standalone `CancellationTokenSource(_reloadTimeout)` token into `ReloadAsync(...)` fixes the actual leak.

I verified the propagation rather than taking the description's word for it: `ReferenceCacheHolder.ReloadAsync` threads the token into both `_reloadGate.WaitAsync(cancellationToken)` and `BuildSnapshotAsync(context, cancellationToken)` (which drives the EF/Npgsql query), so on timeout the gate and the query are genuinely released — not orphaned.

A few design points worth confirming, all of which check out:

- **Belt-and-suspenders await bound.** Keeping `.WaitAsync(timeoutSource.Token)` in addition to passing the token into each reload is deliberate and correct — it still bounds the admin request even if some future `IReloadableReferenceCache` implementation ignored its token. Good defensive design.
- **Exception filter is precise.** `catch (OperationCanceledException) when (timeoutSource.IsCancellationRequested)` only re-maps to `TimeoutException` when *our* timeout actually fired; a genuine DB failure still propagates as its original exception. Externally observed behavior (a `TimeoutException` on the admin response) is unchanged.
- **No DI footgun from the new optional param.** The filter is registered via `AddScoped<AdminCacheReloadFilter>()` and applied through `ReloadReferenceCachesAttribute : ServiceFilterAttribute`. The built-in DI container honors the `TimeSpan? reloadTimeout = null` default for the unregistered parameter, so production resolution still gets the 30s default. The XML comment is honest that the seam exists purely for test timing.

### Test coverage

Solid and follows the project's classical (Detroit) style with hand-written test doubles rather than a mocking framework:
- `PassesACancellableToken_ToEachReload` asserts the cache receives a token where `CanBeCanceled` is true (vs. `CancellationToken.None`).
- `ThrowsTimeoutAndCancelsTheReload_WhenAReloadWedges` exercises the real timeout path deterministically — it asserts the `TimeoutException` surfaces *and* awaits a `TaskCompletionSource` signalled from the token's cancellation callback, confirming the wedged reload was actually released rather than just abandoned. Avoiding a racy boolean flag here is the right call.
- The `BuildExecutingContext` / `BuildExecutedContext` extraction is a clean DRY refactor reused across old and new tests.

### What's good

Tightly scoped to the issue, the in-code comments were updated to match the new behavior (no stale "abandoned" wording left behind), and the testability seam is the minimal idiomatic choice rather than dragging in a clock abstraction. No blocking findings, no should-fixes, no nits.

Holding the merge only until the in-progress backend/e2e CI checks go green.

---

PR #1401, opened 2026-06-30
Load PlayerProgress once in the offline-progress path (#1368)

### LINKED ISSUE #1368: Offline-progress path re-reads the same Progress_{playerId} Redis key 3-4 times

## Observation

A single `GetOfflineProgress` command issues several independent reads of the **same** `Progress_{playerId}` cache key in `Game.Application/Services/BattleService.cs` (`SimulateProgress`, ~lines 411-464 and callees):

- `CaptureProficiencyLevels` → `GetProficiencies`
- `ResolveOfflineLoop` → `IsZoneUnlocked` → `GetCompletedChallengeIds`, and/or `EnsureViableZone` → `GetCompletedChallengeIds`
- `ApplyOfflineRewards` → `progressRepo.Load`

Each `GetCachedProgress` is a full Redis round-trip (plus a fire-and-forget `ExpireAndForget`). They are warm cache hits, so this is **not** a correctness bug — just 3-4 serial round-trips where a single `Load` up front (threading the aggregate down) would do.

## Why it matters

This path runs once per reconnect, so the impact is modest, but it's avoidable serial latency on a login-time path. Note the live `StartBattle` hot path is **not** affected — it reads progress once per battle and the zone-gate checks short-circuit before reading on the common case.

## Suggested direction

Load `PlayerProgress` once at the top of `SimulateProgress` and pass it down to the callees instead of each re-reading the key.

ginderjeremiah:
## Summary

`GetOfflineProgress` (the offline-rewards / character-switch path) re-read the same `Progress_{playerId}` Redis key 3–4 times within a single command:

- `CaptureProficiencyLevels` → `GetProficiencies`
- `ResolveOfflineLoop` → `IsZoneUnlocked` / `EnsureViableZone` → `GetCompletedChallengeIds`
- `ApplyOfflineRewards` → `Load`

Each is a full Redis round-trip (plus a fire-and-forget `ExpireAndForget`). They're warm cache hits, so this was never a correctness bug — just avoidable serial latency on a login-time path.

This loads the `PlayerProgress` aggregate **once** at the top of `SimulateProgress` and threads it down, collapsing the 3–4 reads into one.

## How it addresses the issue

- `SimulateProgress` now does a single `_progressRepo.Load` and derives everything from it: completed-challenge ids (loop/zone gating) via the new `PlayerProgress.CompletedChallengeIds()`, proficiency levels (battle snapshot) via a shared `ToProficiencyLevels` projection, and statistics (reward application) by passing the same aggregate into `ApplyOfflineRewards` instead of having it reload.
- **Ordering preserved:** the load happens *after* `ResolveStaleBattle`. That step settles the disconnected battle, which fires `BattleCompletedEvent` → `BattleStatisticsEventHandler` synchronously (via the awaited dispatcher in `SavePlayer`), mutating the progress cache. Loading before it would snapshot stale progress and overwrite the cache on save, dropping the settled battle's stats. The comment documents this.
- **Live hot paths untouched:** `StartBattle` / `StartBossBattle` / `SetAutoChallengeBoss` keep reading via the existing accessors. `EnsureViableZone` retains a lazy-read overload that only hits Redis on an actual relocation, so the idle loop never pays the read on the common case; a new overload takes the pre-resolved set for the offline caller.
- A shared `ToProficiencyLevels` projection keeps the live and offline proficiency-snapshot derivations from drifting.

## Test plan

- **New unit tests** for `PlayerProgress.CompletedChallengeIds()` (returns only completed challenges; empty when none).
- **Existing offline integration tests** (`BattleServiceOfflineProgressIntegrationTests`, 10 tests) pass unchanged — they cover multi-battle exp/levels, below-threshold no-op, second-call no-op, mid-window challenge completion + reward unlock, proficiency accrual, and zone resolution, which are exactly the behaviors this refactor must preserve.
- `BattleServiceIntegrationTests` + `BattleStatisticsEventHandlerTests` (53 tests) and the full `Game.Core.Tests` suite (1111 tests) pass.
- `dotnet build` clean (0 warnings, 0 errors).

### Note for the reviewer

No regression test pins the *read count* itself: the integration test base (`ApplicationIntegrationTestBase`) builds its provider with no service-override seam, so wrapping `ICacheService` with a counting decorator would mean retrofitting shared infrastructure — out of scope for this small change. Behavioral preservation is covered by the existing integration suite. Happy to add a cache-spy seam as a follow-up if wanted.

Closes #1368

---
_Generated by [Claude Code](https://claude.ai/code/session_01FmHbn28nokHJWgCNN9VXtb)_

### DIFF
--- a/Game.Application/Services/BattleService.cs
+++ b/Game.Application/Services/BattleService.cs
@@ -429,15 +429,22 @@ namespace Game.Application.Services
             // (here) rather than being re-abandoned by the first live StartBattle after the gate.
             await ResolveStaleBattle(player, state, cancellationToken);
 
-            var (mode, zone) = await ResolveOfflineLoop(player, cancellationToken);
-            var proficiencyLevels = await CaptureProficiencyLevels(player.Id, cancellationToken);
+            // Load the progress aggregate once for the whole offline pass. Its completed challenges (loop/zone
+            // gating), proficiency levels (battle snapshot), and statistics (reward application) all read the
+            // same Progress_{playerId} cache key, so threading one aggregate down replaces 3-4 serial round-trips.
+            // Loaded after ResolveStaleBattle so it reflects the settled stale battle's progress mutation.
+            var progress = await _progressRepo.Load(player, cancellationToken);
+            var completedChallengeIds = progress.CompletedChallengeIds();
+
+            var (mode, zone) = await ResolveOfflineLoop(player, completedChallengeIds, cancellationToken);
+            var proficiencyLevels = ToProficiencyLevels(progress.Proficiencies);
             var parameters = BuildSimulationParameters(player, mode, zone, awayMs, proficiencyLevels);
 
             var result = _offlineSimulator.Simulate(parameters, cancellationToken);
 
             var levelBefore = player.Level;
             var statPointsBefore = player.StatPoints.StatPointsGained;
-            var rewards = await ApplyOfflineRewards(player, result, cancellationToken);
+            var rewards = await ApplyOfflineRewards(player, progress, result, cancellationToken);
 
             // Re-anchor the away clock and persist the player (exp/levels/unlocks) in one save. The exp batch
             // already raised its own single core update in ApplyOfflineRewards; this re-anchor raises one more.
@@ -467,7 +474,8 @@ namespace Game.Application.Services
         // when the persisted flag is set and the current zone still has a challengeable boss (in circulation,
         // unlocked, boss authored); otherwise the loop falls back to idle in the nearest viable zone (the same
         // lazy-relocation the live idle loop uses, so a retired/empty current zone never stalls the sim).
-        private async Task<(OfflineLoopMode Mode, CoreZone Zone)> ResolveOfflineLoop(Player player, CancellationToken cancellationToken)
+        private async Task<(OfflineLoopMode Mode, CoreZone Zone)> ResolveOfflineLoop(
+            Player player, IReadOnlySet<int> completedChallengeIds, CancellationToken cancellationToken)
         {
             var currentZoneId = player.CurrentZoneId;
             if (player.AutoChallengeBoss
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Load `PlayerProgress` once in the offline-progress path (#1368)

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't accept a gating approval on my own PR)* — the refactor collapses the 3–4 serial `Progress_{playerId}` reads into one with behavior-preserving equivalence on every threaded path, the new domain method is unit-tested, and the existing offline integration suite pins the preserved behavior.

### Correctness — verified equivalent on each path

- **Completed-challenge gating.** `PlayerProgress.CompletedChallengeIds()` (`_challenges.Values.Where(c => c.Completed).Select(c => c.Challenge.Id)`) yields the same set as the repo's `GetCompletedChallengeIds` (`cached.Challenges.Where(c => c.Completed).Select(c => c.ChallengeId)`) — both derive from the same `GetCachedProgress`, and `ToCoreChallenge` preserves the id. The boss-unlock check `bossZone.IsUnlocked(completedChallengeIds)` is equivalent to the old `IsZoneUnlocked` because `Zone.IsUnlocked` returns `true` for an ungated (`UnlockChallengeId is null`) zone.
- **Proficiency snapshot.** `ToProficiencyLevels(progress.Proficiencies)` projects the same `ToCoreProficiency`-mapped rows the old `CaptureProficiencyLevels` → `GetProficiencies` did; the list is materialized before `ApplyOfflineRewards` mutates `progress`, so the frozen-power window invariant holds.
- **Ordering preserved.** Loading after `ResolveStaleBattle` correctly captures the settled battle's synchronous progress mutation (`BattleStatisticsEventHandler`), and nothing writes the progress cache between `Load` and `ApplyOfflineRewards`, so the shared aggregate is behaviorally identical to the old fresh load (dirty sets empty on entry, same as before).
- **Live hot path untouched.** The live `EnsureViableZone` overload keeps its `IsZoneViable` short-circuit *before* reading `GetCompletedChallengeIds`, so the idle loop still pays only the cheap viability check on the common case. The new overload re-checks viability to stand alone for the offline caller — a cheap in-memory re-check, the right way to keep it self-contained.

### Layering / guidelines

`CompletedChallengeIds()` lands on the `Game.Core` domain aggregate (correct home for the rule), `BattleService` stays orchestration-only, and the shared `ToProficiencyLevels` projection keeps the live and offline snapshot derivations from drifting — all in line with `backend.md`.

### Test coverage

- New unit tests for `CompletedChallengeIds()` (completed-only filtering; empty when none) cover the added domain logic.
- Behavioral preservation rides the existing `BattleServiceOfflineProgressIntegrationTests` (multi-battle exp/levels, mid-window challenge completion + unlock, proficiency accrual, zone resolution) — exactly the paths this refactor threads through.
- The author's note about not pinning the read-count itself is reasonable: a counting `ICacheService` decorator would mean retrofitting the shared integration-test base, fairly out of scope for a small refactor. No follow-up needed.

### What's good

Tight, well-reasoned change that does exactly what the issue asked with no behavioral risk. The comments documenting the load-ordering constraint (after `ResolveStaleBattle`) and the live-vs-offline overload split are the kind of context a later edit could easily break without them — appropriately placed.


---

Issue #1390, opened 2026-06-30
Spike — Content authoring pipeline: source-controlled content + progression-graph lint


ginderjeremiah:
Now that content authoring is starting in earnest, this spike scopes the **tooling and process** around authoring — *not* the content itself. The admin **Workbench** already covers all 9 reference entities well (add/edit/retire, validation warnings, cross-ref pickers), so catalogue *values* are solved. The gaps are everything around them.

## Context / problem
- Content (`Skill`, `Item`, `ItemMod`, `Enemy`, `Zone`, `Challenge`, `Class`, `Path`/`Proficiency`, `SkillRecipe`) lives **only in Postgres**, authored at runtime. There is no source-controlled snapshot, so a content change can't be diffed, reviewed, reproduced in a fresh dev/test/CI DB, or rolled back — and a fresh database boots with **zero** content.
- Content is a heavily cross-referenced **graph** (zone → gated-by → challenge → rewards → item → grants → skill → trains → path; class starter kits; synthesis recipes; proficiency-milestone skill grants). Save-time invariants are enforced **per entity**, but nothing checks the **whole graph** for reachability — the dangling end of a broken reference lives on the *other* record.
- Design *intent* (why a piece exists, the macro progression arc) has nowhere to live; the Workbench holds values, not rationale.

## Central question — authoring model (strawman to validate)
Decide between:
- **(A) DB-primary + one-way export** — Workbench stays canonical; a deterministic exporter mirrors reference data to source-controlled JSON used only as a seed (dev/test/CI/recovery). JSON is *generated, never hand-edited.*
- **(B) Files-primary** — JSON in the repo is the source of truth; DB is a build artifact seeded from files.

Strawman lean is **(A)**: the Workbench's pickers/validation (the safety net that makes a cross-referenced graph authorable) and the 0-based contiguous-id + retire-don't-delete contract both live in the DB write path; (A) buys ~90% of the git-native benefit (diff/review/reproduce/rollback/CI-lint) for ~10% of the cost. The spike must confirm the export diffs are clean and stable enough to make (B) unnecessary.

## Explore
- **Serialization shape & stability** — per-entity files vs per-zone "content packs"; deterministic ordering + canonical formatting (id order, stable float formatting, child-collection ordering) so diffs are minimal and review-friendly. This is the make-or-break for (A).
[...]

PR #1409, opened 2026-06-30
Add 'Home' zone — a no-combat sanctuary (#1394)

### LINKED ISSUE #1394: Add 'Home' zone where no enemies spawn (and never will)
A zone to sit in when you don't want to be idly battling enemies. I'm tempted to make this a 'fake' zone where, even if the player moves here, they are still counted as being in the zone they last idled in (for offline calculation purposes).

ginderjeremiah:
## Summary

Adds a special **Home** zone the player can idle in without battling — a place to step away from the idle farm (manage a build, read, etc.) without progress pressure.

Per the issue's lean ("a 'fake' zone where… they are still counted as being in the zone they last idled in"), and after confirming two user-facing decisions, Home is modelled as a **fake zone**:

- It **never spawns enemies** (and never will — authoring rejects a boss or any spawn-table membership on it).
- It **never becomes the player's persisted `CurrentZoneId`** — the battle-start zone-change anti-cheat refuses a move into it.

Because the persisted zone always stays the player's last *real* combat zone, **offline rewards need no special case**: an away window replays that last zone exactly as if the player had never stepped Home. This also means Home is the one zone naturally exempt from the existing "empty zone → relocate" logic, since the player is never actually there server-side.

Confirmed with the issue author before implementing:
- **Persistence:** live-only — reconnecting resumes combat in the last real zone (no persisted "at home" state).
- **Access:** Home sits leftmost in the existing zone nav (always open).

## How it addresses the issue

- A navigable zone where no enemies spawn (and never will), enforced structurally via authoring guards.
- Entering Home halts the live idle loop and shows a resting state instead of fighting.
- Offline calculation continues to credit the zone the player last idled in, with zero change to the offline math.

## Changes

**Backend**
- `Zone.IsHome` flag on the entity + read contract + EF config (default `false`, with a migration); the battle never reads it, so it's resolved via the new `IZones.IsHomeZone(id)` (mirroring `IsZoneRetired`), not carried on the lean `Game.Core` zone.
- `BattleService.StartBattle` zone-change anti-cheat refuses a move into a Home zone (the "fake zone" invariant the offline path relies on).
- Authoring guards: a Home zone can't declare a boss (`AdminZones.SaveZones`), and neither `AdminZones.SetEnemies` nor `AdminEnemies.SetSpawns` can populate a Home zone's spawn table.

**Frontend**
- `IZone.isHome` (regenerated client types). Home renders leftmost and always-open in `ZoneNav`, labelled **"Sanctuary"** — combat zones stay numbered from 1.
- `EnemyManager.navigateToZone` halts the idle loop and stops the live battle (resting panel via new `HomeRest.svelte`) when entering Home, and resumes the loop on leaving; `runFetchLoop` no longer sends `NewEnemy` while at Home.
- The battle clock (`BattleEngine.timeElapsed`) now advances only while a battle is **Active** — a latent fix so it can't drift while parked between battles / at Home and corrupt a later abandon's reported duration. Added `BattleEngine.rest()` to stop the live fight cleanly.
- Admin Workbench zone editor gains an **IsHome (Sanctuary)** toggle, and its "no enemies spawn here" warning is suppressed for a Home zone.

**Docs**
- Documented the Home-zone design + invariant in `docs/game-design.md` and `docs/backend.md`.

## Test plan

- **Backend** (`dotnet test Game.sln`): all green — Core 1111, Infrastructure 60, Application 667, Api 652. New tests: `IsHomeZone` repo behaviour, `StartBattle` refuses a Home zone-change, and the three authoring guards (boss-on-Home, `SetEnemies`/`SetSpawns` targeting Home).
- **Frontend** (`npm run lint` + `npx vitest --run`): lint/type-check clean; 3115 unit tests pass. New tests: Home loop-halt + `navigateToZone` in `enemy-manager`, `rest()` + Active-only clock in `battle-engine`, Home labelling/numbering/navigation in `ZoneNav`, and the resting panel in `Fight`.

## Follow-up note

The Home zone is **authored content** (like every zone — a fresh DB boots empty), identified by the `IsHome` flag and surfaced for authoring in the Workbench. It's seeded in `e2e-seed.sql` for tests; a production Home zone would be authored through the admin tools (consistent with the content-authoring pipeline being scoped in #1390).

Closes #1394

---
_Generated by [Claude Code](https://claude.ai/code/session_018nszzfLUw48eyjqRQXiPmn)_


ginderjeremiah (commented):
## Review: Add a no-combat "Home" sanctuary zone (closes #1394)

**Verdict:** Approve _(submitted as a comment — I'm the PR author, so GitHub won't accept a gating event from me; this is a genuine approval)_ — the implementation matches the issue's "fake zone" lean, is cleanly layered, and is thoroughly tested on both tiers with green CI.

### What's good

- **Sound "fake zone" model.** Modelling Home as a zone that never becomes the persisted `CurrentZoneId` (the `StartBattle` zone-change anti-cheat refuses the move) is what lets offline rewards need *zero* special-casing — the away window just replays the last real combat zone. This is the right call and it's documented well in both `game-design.md` and `backend.md`.
- **Correct DDD layering.** `IsHome` rides the entity + read contract only and is resolved via `IZones.IsHomeZone` (mirroring `IsZoneRetired`); `ZoneMapper.ToCore` correctly omits it so the lean battle `Game.Core` zone stays untouched. The battle never reads it — exactly the documented authoring-metadata pattern.
- **Authoring guards in both directions.** Rejecting a boss on a Home zone (`AdminZones.SaveZones`) and a spawn-table membership from both `AdminZones.SetEnemies` and `AdminEnemies.SetSpawns` is the anti-tamper rigor this codebase expects. The `LookupZone(...) is { IsHome: true }` pattern also degrades safely on an out-of-range id.
- **The `timeElapsed` latent fix is a real improvement.** Advancing the battle clock only while `stage === Active` stops it drifting up during the post-death/Loading window before the next `NewEnemy`, so the abandon re-simulation bound (`clientBattleMs`) reflects only real combat time. Parity is preserved — the timeout check still evaluates against this tick's freshly-accumulated value inside the `Active` branch, and this clock is a frontend-only live-loop concern, not part of the deterministic shared simulation.
- **Careful loop-halt wiring.** `enterHome`/`runFetchLoop`'s home-check correctly reuse the existing generation/supersession guards, so the various races (entering Home mid-cooldown, leaving Home, a prepared battle that must be discarded) all resolve without a double-spawn or a fight rendering at Home. Leaving Home reports `clientBattleMs: 0`, forfeiting the orphaned fight as a clean abandon.
- **Frontend theming + organization.** `HomeRest.svelte` is a small, focused component using theme tokens (`color-mix`, `--accent-light`, `--mono`, `--text-primary`) rather than hardcoded colors, and `ZoneNav`'s `combatZoneNum` numbers combat zones independently of Home's nav position (robust even if Home weren't leftmost).

### Test coverage

Strong on both tiers and meets the project's rules:
- **Backend:** `IsHomeZone` repo behaviour (valid id, invalid-id throw, contract field), the `StartBattle` Home-move refusal, and all three authoring guards (boss-on-Home, `SetEnemies`/`SetSpawns` targeting Home).
- **Frontend:** Active-only clock + idle freeze + `rest()` in `battle-engine`, the Home loop-halt + `navigateToZone` in/out of Home in `enemy-manager`, the resting panel in `Fight`, and `Sanctuary` labelling / combat-zone numbering / left-arrow navigation in `ZoneNav`.
- The remaining test-file churn is just the required `isHome: false` fixture additions now that `IZone.isHome` is non-optional.

No FE/BE battle-parity gap: the only battle-logic touch (the `timeElapsed` gate) is a live-loop concern outside the deterministic simulation, so it needs no mirrored backend test.

### Nits (non-blocking)

- **No enforcement of the "single Home zone" the design states** — `AdminZones.SaveZones` would accept a second `IsHome` zone, yielding a confusing double-"Sanctuary" nav. Nothing breaks (numbering and offline rewards both handle it), but a save-time guard would be consistent with how every other authoring invariant here is backend-enforced. Filed as follow-up **#1410** (scope: small).
- Minor: `atHome` is derived two slightly different ways (`ZoneNav` via the ordered-zones lookup, `Fight` via the direct id-indexed `staticData.zones[id]`). Both are correct; just an inconsistency to be aware of.

---

PR #1412, opened 2026-06-30
Gate proficiency XP on a gateway's prerequisites being maxed (#1411)

### LINKED ISSUE #1411: Enforce that umbrella proficiencies that have not been unlocked (due to prereqs) cannot be contributed to
Enforce that proficiencies that have not been unlocked (due to prereqs) cannot be contributed to.

For example, the "elemental" umbrella (currently fire, water, wind, earth) will not be available at the start. Instead, it will be unlocked only after reaching advanced levels of all the leaf damage types under it. In that case, the leaf skills should not contribute towards it until it is unlocked.

ginderjeremiah:
Closes #1411.

## Problem

A path's first tier may be a cross-path **gateway** whose prerequisites must all be maxed before the path unlocks (e.g. the *Elemental* umbrella, gated behind advanced levels of its leaf damage types). The battle XP accrual routes a battle's activity to every path bound to its activity key via the path's **frontier** tier — but `Path.Frontier` returns the lowest *un-maxed* tier even when that tier is a still-locked gateway.

As a result, an umbrella category banked XP from leaf activity before it was unlocked: a fire hit routes to both `Fire` and `Elemental` (`applies(Fire) = [Fire, Elemental]`), so the leaf type trained the locked `Elemental` path. The resulting `PlayerProficiency` row also leaked the gateway to the client, which derives a gateway's visibility from the presence of a progress row.

## Fix

In `ProficiencyRewardService.BuildActivities`, route a key's activity to a frontier tier **only when that tier is unlocked** — i.e. all of its prerequisites are maxed. This reuses the existing `AllPrerequisitesMaxed` gateway-unlock test (its comment is generalized now that it has a second caller).

- **Within-path tiers are unaffected** — they carry no authored prerequisites (order is implicit in `PathOrdinal`, already enforced by the frontier scan), so the check is trivially true for them.
- **Both accrual paths are covered** — the live (`BattleStatisticsEventHandler`) and offline (`BattleService` consolidation) paths both run through `BuildActivities`, so the offline window begins training an umbrella only after a prior battle in the window maxes its last prerequisite.
- **The client self-corrects** — with no progress row written for a locked gateway, its `openedIds`-based visibility no longer fires until the prerequisites are genuinely maxed. No frontend change needed.

## Tests

Two integration tests added to `ProficiencyRewardServiceTests`:
- `LockedUmbrella_AccruesNoLeafActivity_UntilItsPrerequisitesAreMaxed` — Fire activity trains the Fire leaf path but **not** the locked Elemental umbrella (regression guard; fails without the fix).
- `UnlockedUmbrella_AccruesFromLeafActivity_OnceItsPrerequisitesAreMaxed` — with the prerequisite maxed, the same Fire activity trains the Fire path *and* the Elemental umbrella.

Full `Game.Application.Tests` (669) and `Game.Core.Tests` proficiency suites pass.

## Docs

Updated the "Opening tiers and milestones" section of `docs/game-design.md` to call out the gateway exception: a gateway tier accrues no XP until every prerequisite is maxed, so leaf activity does not bank progress onto a still-locked umbrella.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fz9Te6nFNymw6e4GpB2SRe)_


ginderjeremiah (commented):
## Review: Gate proficiency XP routing on a gateway's prerequisites being maxed (#1411)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't accept a gating event) — correct, minimal, well-tested fix with verified FE/BE consistency.

The one-line gate in `BuildActivities` (`ProficiencyRewardService.cs:221`) is the right place and the right shape. Routing a key's activity to a frontier tier only when `AllPrerequisitesMaxed(frontier)` holds precisely closes the leak: a locked umbrella (e.g. *Elemental*, fed by `applies(Fire) = [Fire, Elemental]`) no longer banks leaf activity, so no `PlayerProficiency` row is written and the gateway stays locked until its prerequisites max.

Things I checked and confirmed:

- **Reuse over duplication.** The fix leans on the existing `AllPrerequisitesMaxed` (its comment generalized for the second caller) rather than a new gateway-unlock predicate. Good DRY.
- **Both accrual paths covered.** `AccrueAndApply` → `BuildActivities` is the shared route for the live (`BattleStatisticsEventHandler`) and offline (`BattleService`) paths, so the offline window self-corrects too.
- **No false negatives for pure roots / within-path tiers.** Those carry no prerequisites, so `AllPrerequisitesMaxed` is trivially true — a leaf root like Fire still opens by training, unaffected.
- **FE/BE consistency holds, no frontend change needed.** `derivePathSpine` in `proficiencies-lexicon.ts` reveals a gateway root via `openedIds.has(prof.id) || prereqsMaxed`. The bug was that the leaked progress row made `openedIds.has` fire early; with no row written, the gateway is drawn only once prereqs max. The existing `proficiencies-view.test.ts` "Lava gated on Inferno maxed" case already pins this derivation. I also checked the deeper-tier-with-prerequisites edge case — both sides converge on banking nothing until the gate clears, so there's no divergence.

### Test coverage
Two focused integration tests against live DB-backed reference data, matching the sibling-test pattern (`CreateKeyedTierAsync` + `AccrueStatsAsync`): `LockedUmbrella_AccruesNoLeafActivity_…` is a genuine regression guard (fails without the fix), and `UnlockedUmbrella_AccruesFromLeafActivity_…` confirms the parallel case once the prerequisite is maxed. Asserting on `accrual.Results` is the right proxy for "a row was written." No mirrored frontend test is required — proficiency XP accrual is deliberately off the battle-parity surface (per `backend.md`), and the FE openness derivation it depends on is unchanged and already covered.

### Docs
The `game-design.md` "Opening tiers and milestones" update correctly carves out the gateway exception. The sentence runs long, but it's accurate and not worth churning.

CI is fully green. No blocking items, no should-fixes, no nits. Merging.

---

PR #1431, opened 2026-07-01
Seed a fresh DB from the content export; retire e2e-seed.sql (#1419)

### LINKED ISSUE #1419: Seed a fresh DB from the content export (dev/test/CI/recovery); retire e2e-seed.sql
Part of the content-authoring-pipeline spike #1390 — decision 3. See [`docs/spikes/1390-content-authoring-pipeline.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/1390-content-authoring-pipeline.md).

**Depends on #1418** (defines the on-disk export format this consumes).

A freshly-migrated database has **zero** static content (only the intrinsic `HasData` enum sets). This adds a seeder that reconstructs the full static content graph from the source-controlled export.

## Scope
- A **dedicated bulk seeder** (decision 3, option ii) that writes entities directly from the export:
  - **explicit ids** (identity insert) to preserve the 0-based contiguity and the `Id == index` snapshot assertion.
  - preserves `RetiredAt` exactly.
  - inserts in FK/dependency order (referenced sets before referencing sets, parents before children).
  - coexists with the intrinsic `HasData` seeds already present post-migration.
  - warms/reloads the reference caches once at the end.
- Wire it into dev / test / CI / recovery startup paths (a fresh DB gets a real content baseline).
- **Retire `e2e-seed.sql`** once the seeder covers that slice (the punch fallback + class-kit skills 0–3, one zone with enemies, a boss). Update the e2e/CI workflow accordingly.

## Notes / open
- Chosen over replaying through the admin repos: the split child endpoints + awaited-reload filter + whole-graph cross-ref validation can't satisfy forward references mid-import (chicken-and-egg). A fresh-DB seed is a trusted whole-graph op needing correctness/speed/id-fidelity, not the incremental anti-tamper path.
- Decide the trigger surface (env-gated startup seed vs. an explicit CLI/test fixture) and whether prod is ever seeded this way (recovery).
- This also hands the offline-rewards (#879) and battle test fixtures real content instead of bespoke rows.

## Tests
- Integration: seed a fresh (migrated) DB from the export, then assert the reference caches build cleanly (`Id == index` holds) and a representative cross-ref (zone → boss → enemy, skill → portions) resolves.
- The e2e suite still reaches the game after `e2e-seed.sql` is removed.

ginderjeremiah:
## Summary

Implements the reverse direction of the content-authoring pipeline (spike #1390, decision 3): a JSON-driven seeder that reconstructs the whole static reference-data graph on a freshly-migrated database from the source-controlled export (`content/*.json`), so dev / CI / recovery get a real content baseline instead of an empty catalogue. Closes #1419.

## How it addresses the issue

- **`IContentSeeder`** (`Game.DataAccess/Content`) — a **dedicated bulk seeder**, not a replay through the admin repos (per decision 3, option ii). It:
  - writes rows with their **explicit ids** — a metadata-driven `EntityRowInserter` builds the INSERTs from the EF model, so it preserves id `0` (which EF's normal insert path treats as store-generated for a zero-based identity column) and can't drift from the schema;
  - preserves `RetiredAt`, inserts in **FK/dependency order** (referenced sets before referencing ones, parents before children), and advances each identity sequence past its seeded max;
  - is **idempotent** — no-ops when the DB already holds content (Skills present), so it never clobbers an authored dev DB;
  - coexists with the intrinsic `HasData` seeds already present post-migration.
- **Reverse `ToEntity` mappers** for the 10 static sets (Skill, Item, ItemMod, Enemy, Challenge, Zone, Class, Path, Proficiency, SkillRecipe) alongside the existing `ToContract`/`ToCore`.
- **`IContentImportReader`** (`Game.Application/Content`) deserializes the export through the exporter's canonical JSON options; the files are copied next to the app so the seeder resolves them from `AppContext.BaseDirectory/content` uniformly in a local run and the container.
- **Startup wiring** — seeds after migrating and before the eager reference-cache load (so the caches build from the seed, no restart needed), gated on the new `DataAccessOptions.SeedContentOnStartup`.
- **Retires `e2e-seed.sql`** — the e2e Docker stack now seeds itself at boot (`SeedContentOnStartup=true`), dropping the old post-boot `psql` seed + API restart. The e2e-only admin-role fixture (not content) is split out to `e2e-admin-provision.sql` and applied by CI.

## Design decisions (the issue's open questions)

- **Trigger surface:** env-gated startup seed — always on in Development, opt-in elsewhere via `SeedContentOnStartup`, idempotent so it's safe to leave enabled. Chosen over an explicit CLI/fixture because the Docker/e2e stack needs a populated DB at boot and this subsumes the old seed+restart dance. Integration tests call the seeder directly.
- **Production / recovery:** off by default; a recovery is an explicit opt-in (enable the flag against a fresh/empty DB). The idempotency guard keeps that safe.

## Test plan

- **New `ContentSeederTests`** (integration): seed a fresh migrated DB from the export and assert the caches build cleanly (`Id == index` holds), cross-refs resolve (zone→boss→enemy, skill→portions, class kit→skills), ids stay 0-based, and re-seeding is a no-op.
- **`ContentExportDriftGuardTests`** now seeds via the seeder → re-exports → asserts byte-equality: a round-trip that also catches seeder/exporter asymmetry.
- Full backend suite green locally (2528 tests across Core/Infrastructure/Application/Api); `dotnet format --verify-no-changes` clean.

## Follow-up

- The export includes item/mod **tag ids** but tags themselves aren't among the 10 exported sets, so seeding tag-join rows assumes the referenced `Tag` rows already exist. The committed content has no items/tags, so nothing is stranded today — worth a follow-up if/when tag content lands (either export tags or make the seeder tolerant).

https://claude.ai/code/session_01JL7Kig4mztVxn3o3yNkmv2

---
_Generated by [Claude Code](https://claude.ai/code/session_01JL7Kig4mztVxn3o3yNkmv2)_


Game.DataAccess/Content/ContentSeeder.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            // Advance every seeded identity table's sequence past its highest explicit id.
+            await EntityRowInserter.AdvanceIdentitySequencesAsync(
+                _context,
+                [
+                    typeof(EntityPath), typeof(EntitySkill), typeof(EntitySkillEffect), typeof(EntityProficiency),
+                    typeof(EntityItemMod), typeof(EntityItem), typeof(EntityItemModSlot), typeof(EntityEnemy),
+                    typeof(EntityChallenge), typeof(EntityZone), typeof(EntityClass), typeof(EntitySkillRecipe),
+                ],
+                cancellationToken);
```
ginderjeremiah: Nit (maintainability, non-blocking): this explicit type list is a hand-maintained parallel to the `Insert(...)` calls above. It's correct today — `SkillEffect` and `ItemModSlot` are the only surrogate-keyed child entities and both are here; everything else is composite-keyed — but a future child entity added with its own `Id` identity would need to be remembered here or its sequence silently isn't advanced, letting a later admin insert collide.

Since `AdvanceIdentitySequencesAsync` already guards (`key.Properties.Count != 1` / `ValueGenerated != OnAdd` are skipped), you could pass the full set of inserted entity types and let the guard filter down to the identity-keyed ones, removing the second list to keep in sync. Cheap robustness for a fresh-DB one-shot.

---

Game.Api/Startup.cs
```diff
@@ -199,10 +207,19 @@ public static async Task Main(string[] args)
...
                 await migrator.Migrate();
+
+                // Seed the static content into a freshly-migrated database (no-op when it already has content),
+                // before the reference caches load below so their eager build picks the seed up.
+                if (seedContentOnStartup)
+                {
+                    var reader = migrationScope.ServiceProvider.GetRequiredService<IContentImportReader>();
+                    var seeder = migrationScope.ServiceProvider.GetRequiredService<IContentSeeder>();
+                    await seeder.SeedAsync(reader.ReadDefault());
+                }
```
ginderjeremiah, line 222: Nit (non-blocking): the seed only runs inside the `IsDevelopment() || migrateOnStartup` block, so `seedContentOnStartup` computed from `SeedContentOnStartup` alone is a silent no-op when `MigrateOnStartup` isn't also set — the recovery scenario the option is meant for. It's defensible (seeding needs a migrated schema, and an unmigrated DB fails at the cache load anyway) and the comment at 176-177 calls it out, but a recovery operator enabling only `SeedContentOnStartup` gets nothing with no hint why. Consider logging a warning when `seedContentOnStartup` is set but the migration gate is off, so the misconfiguration is visible rather than silent.

---

ginderjeremiah (commented):
## Review: JSON-driven content seeder that reconstructs a fresh DB from the source-controlled export; retires `e2e-seed.sql` (#1419)

**Verdict:** Approve — solves the issue cleanly, respects the layering and the reference-data id gotchas, and is backed by real tests. (Posted as a comment because I'm the PR author; GitHub won't take a gating verdict on your own PR. Two nits below, neither blocking.)

I reviewed the diff against `docs/backend.md` (entity isolation, reference-data 0-based ids, retirement, content export). The change matches the design already documented at backend.md → "Content export / Seeding a fresh DB from the export."

### What's good
- **Correct layering.** `IContentSeeder` is implemented in `Game.DataAccess` (it touches EF entities) while `IContentImportReader` lives in `Game.Application` and speaks only read contracts — the entity seam stays inside the data tier, honoring the `Game.Api`/`Game.Application` entity-invisibility rule.
- **The 0-based-id problem is handled head-on.** `EntityRowInserter` builds the INSERT from EF model metadata (table/column names can't drift from the schema) and parameterizes values, so it preserves explicit id `0` that EF's normal path would let the store generate. `SeedAsync_PreservesExplicitZeroBasedIds` pins it.
- **FK/dependency insert order is carefully reasoned** (paths → skills → proficiencies → mods → items → enemies → challenges → zones → zone-enemy joins → classes → recipes), with self-referencing sets (proficiency prerequisites, recipe inputs) inserted after their parent set. Sequence advancement uses `setval(MAX)` with the `HAVING COUNT(*)>0` guard so empty sets keep their natural start.
- **Idempotency guard** (Skills present → no-op) means an authored dev DB is never clobbered, and Development re-runs are safe.
- **Strong test coverage.** `ContentSeederTests` covers fresh-seed cache build, cross-refs (zone→boss→enemy, class kit→skills, skill→portions), 0-based ids, and re-seed no-op. `ContentExportDriftGuardTests` now round-trips seed→re-export→byte-equality, which additionally catches any `ToEntity`/`ToContract` mapper asymmetry — a good backstop given the 10 new reverse mappers.
- **Infra simplification is real.** The e2e stack seeds itself at boot; the seed+`restart api` dance is gone, and only the admin-role fixture (correctly identified as test-harness infra, not content) is applied out-of-band. Docs, README, docker-compose, and the CI `detect-changes` path filter (`content/**`) are all updated consistently.

### Nits (take-it-or-leave-it)
- `Game.DataAccess/Content/ContentSeeder.cs:113` — the identity-sequence type list is a hand-maintained parallel to the insert calls; correct today, but could be derived from the inserted set since `AdvanceIdentitySequencesAsync` already filters non-identity keys. (inline)
- `Game.Api/Startup.cs:217` — `SeedContentOnStartup` alone is a silent no-op without `MigrateOnStartup`; consider logging when it's set but the migration gate is off. (inline)

### Test coverage
Adequate and in line with the project's rules: integration tests for the out-of-process (Postgres) interaction, plus the round-trip drift guard. No frontend/battle logic touched, so no parity-mirror obligations. The reverse mappers are pure data projections exercised end-to-end by the drift guard.

### Follow-up (already noted in the PR description, non-blocking)
The export carries item/mod **tag ids** but `Tag` isn't among the 10 exported sets, so `itemTags`/`itemModTags` assume the referenced `Tag` rows pre-exist. Nothing is stranded today (no item/tag content committed). Worth a tracked follow-up (either export tags or make the seeder tolerant) before tag-bearing content lands — I'll open one so it isn't lost.

---

ginderjeremiah:
## Review — Seed a fresh DB from the content export (#1419)

**Verdict: Approve (with one non-blocking nit).** This is a clean, well-reasoned implementation that faithfully closes #1419. CI is green (verify/test backend + frontend, all three e2e browsers, all-checks-passed). I verified the PR's three headline claims against the actual code, not just the description.

### What it does well
- **Idempotency guard is correct.** `SeedAsync` no-ops when `Skills` already has rows — Skills is genuinely the foundational set (punch fallback, class kit, enemy/proficiency grants all reference it), so its presence is a sound "already populated" marker. Covered by `SeedAsync_AlreadyPopulated_IsANoOp`.
- **0-based id preservation is real, not just claimed.** `EntityRowInserter` builds a parameterized multi-row INSERT from EF model metadata and reads values straight off the CLR entity, so it writes explicit ids including `0` (which EF's normal `Add`/`SaveChanges` would treat as store-generated on a zero-based identity column). Table/column names come from the model so the SQL can't drift from the schema; all values are parameterized (no injection surface). Pinned by `SeedAsync_PreservesExplicitZeroBasedIds` and, transitively, by the cache-build `Id == index` assertion in `SeedAsync_FreshDatabase_...`.
- **Dependency ordering is FK-correct.** I traced every set and child/join table: paths/skills first → proficiencies (reward skills) → mods/items (gate on proficiencies, grant skills) → challenges (reward items/mods) → zones (boss enemy + unlock challenge, both already present) → `ZoneEnemies` after both zones and enemies → classes (skills+items) → recipes (skills+proficiencies). Prerequisites/join rows land after their parents. Whole seed is wrapped in a single transaction, so a partial seed can't leave a torn catalogue.
- **Sequence advancement is thoughtful.** `AdvanceIdentitySequencesAsync` filters to single-int identity keys (skips composite-key joins) and uses `HAVING COUNT(*) > 0` to leave an empty table's sequence at its natural start — which correctly dodges the "setval to 0 on a non-zero-based sequence" trap. The 2nd commit deriving the inserted-type set from what was actually inserted (vs. a hand-maintained parallel list) is a good robustness fix.
- **Startup wiring** seeds after migrate / before the eager cache load (no restart), and the `else if (seedContentOnStartup)` branch logs a warning when seed is enabled but migration is off — surfacing the recovery misconfiguration instead of a silent no-op. Nice touch.
- **e2e-seed.sql retirement is complete and consistent.** The admin-role trigger is extracted verbatim to `e2e-admin-provision.sql`, and the CI workflow, both compose files, README, and docs (backend.md + infrastructure.md) are all updated coherently. The drift guard is now a genuine round-trip (seed export → re-export → byte-equal), which additionally catches seeder/exporter asymmetry.
- No null-forgiving operators, no `TODO`/`HACK`, no dead code left behind, domain isolation respected (entities stay inside the data tier; the seeder speaks contracts in, entities internally). Reverse `ToEntity` mappers are symmetric with `ToContract` and correctly carry `RetiredAt` and explicit parent ids on child rows.

### Nit (non-blocking)
- The noted **tag-id follow-up** is accurate and genuinely dormant today (`content/items.json` and `content/item-mods.json` are both `[]`, so no `ItemTag`/`ItemModTag` join rows are seeded and nothing references a `Tag` that doesn't exist). Worth tracking before item/tag content lands, because the seeder builds tag-join rows straight from contract tag-ids and assumes the referenced `Tag` rows already exist — but tags aren't among the 10 exported sets, so a fresh DB would have none. Options: export tags as an 11th set, or make the seeder tolerant/validating. No change needed in this PR.

### Suggested follow-up to track
- **Seed tag content or guard tag-join FKs** — the export carries item/mod tag-ids but not the `Tag` catalogue itself, so seeding tag-join rows on a fresh DB assumes the `Tag` rows exist. Harmless now (no tagged content), but a latent FK failure once tags/items are authored. Labels: `enhancement`, `claude`, `scope: small`.

Overall this is solid work — approving. Nice job on the defensive touches (transaction, empty-table sequence handling, the migration-gate warning).

— Reviewed by Claude

---

ginderjeremiah:
### Merge coordination note

Verdict stands at **approve-with-nits, CI green** — no blocking issues. I've **held off auto-merging** because of coupling with the sibling content-pipeline PRs from spike #1390:

- **#1434 (progression-graph lint) is now merged**, and all three PRs touch `docs/backend.md`, so expect a trivial rebase conflict there.
- **#1436 (authoring `DesignerNotes`) is the real coupling.** #1436 adds a `required DesignerNotes` to every static contract and threads it through `ToContract`. This PR adds the reverse `ToEntity` mappers plus a round-trip drift guard (seed the export → re-export → assert byte-equality). Whichever of the two lands second must thread `DesignerNotes` through the **`ToEntity`** mappers here, or the seeder drops the field and the round-trip guard breaks as soon as a non-empty note is authored (it only passes today because everything backfills to `""`). Because that's a real code change that should be reviewed rather than force-merged, I didn't auto-merge.

**Suggested sequence:** merge #1436 first, then rebase this branch and add `DesignerNotes` to the 10 `ToEntity` mappers before merging.

The tag-id gap you flagged in the description is filed as follow-up **#1440** (export tags as an 11th set, or make the seeder validate tag FKs) — confirmed dormant today since `items.json` / `item-mods.json` are empty. Ready to merge once the sequencing above is handled.

---

ginderjeremiah:
## Review by Claude

### Summary
Implements the reverse of the content-authoring pipeline (spike #1390, decision 3): a JSON-driven bulk seeder that reconstructs the whole static reference graph on a freshly-migrated DB from `content/*.json`, retires `e2e-seed.sql`, and splits the e2e admin-role trigger into `e2e-admin-provision.sql`. Scope, wiring, and docs match the linked issue #1419 faithfully.

### Assessment
This is a well-executed, cohesive change. Highlights I verified:

- **Explicit-id fidelity.** `EntityRowInserter` builds parameterized multi-row INSERTs from EF model metadata and reads values straight off the CLR entity, so id `0` is written as a literal `0` rather than being treated as store-generated. The `AdvanceIdentitySequencesAsync` fix-up correctly filters to single-int identity keys, skips composite-key join tables, and the `HAVING COUNT(*) > 0` guard avoids a `setval(..., 0)` against a non-zero-based sequence. Clean.
- **Security (raw SQL).** No injection surface: table/column names come exclusively from EF metadata (trusted), and every non-null value is parameterized via positional `{N}` placeholders passed to `ExecuteSqlRawAsync`; `NULL` is the only inlined literal. Values are never string-interpolated. Good.
- **Dependency ordering.** The insert order is correct against every FK I traced — paths/skills first, then proficiencies (reward skills), item-mods/items (gate on proficiencies, grant skills), challenges (reward items/mods), zones (boss enemies + unlock challenges), spawn-table joins after both zones and enemies exist, classes (skills+items), recipes (skills+proficiencies). Sequence advancement is derived from the actually-inserted types rather than a hand-maintained list — a nice touch that can't drift.
- **Idempotency.** Guarded on `Skills` presence (the foundational set) inside a transaction; the round-trip proven by tests.
- **Reverse `ToEntity` mappers** cover all 10 sets with their child graphs; casts (`(int)`/`(int?)`) line up with the entity nullability I spot-checked (`ScalingAttributeId` non-nullable both ends).
- **Startup wiring** seeds after migrate and before the eager cache load, and the `SeedContentOnStartup && !MigrateOnStartup` misconfiguration now logs a warning instead of silently no-op'ing — good defensive follow-through from the second commit.
- **Test coverage** is strong: fresh-seed cross-ref resolution + `Id==index`, explicit 0-based id preservation, re-seed no-op, and the drift guard is now a genuine seed→re-export→byte-equality round-trip that also catches seeder/exporter asymmetry.
- **CI fixture split** is clean and the docs (backend.md, infrastructure.md, README, compose files, e2e helpers) are all updated consistently with no stale `e2e-seed.sql` references left behind.

**CI:** green — all 9 checks pass (verify/test backend + frontend, all three e2e browsers, all-checks-passed).

### Nits / non-blocking
- **Tag-join rows assume `Tag` rows pre-exist** (already flagged by the author). `content/items.json` and `item-mods.json` are both `[]` today, so nothing is stranded, but the seeder would emit `ItemTag`/`ItemModTag` rows referencing a `Tag` catalogue that isn't among the 10 seeded sets. Worth a tracked follow-up before tag content lands — either export tags as an 11th set or make the join insertion tolerant. Fine to defer.
- **Minor:** `EntityRowInserter` throws on any shadow-property key (no CLR accessor). That's the right fail-loud stance for a trusted seeder, but it's an implicit coupling to "no static entity uses shadow properties" — a one-line note or a targeted test would pin the assumption. Optional.

### Verdict
**Approve** — clean, correct, well-tested, and faithful to #1419. The tag-join follow-up is the only item worth tracking, and it's non-blocking given the empty item/mod slice today.


---

PR #1445, opened 2026-07-01
Fix flaky AsyncWorker_SignaledRepeatedlyDuringRun infrastructure test

ginderjeremiah:
## Summary

Fixes the intermittent CI failure in `BackgroundWorkerTests.AsyncWorker_SignaledRepeatedlyDuringRun_NeverOverlapsAndSettlesNotRunning`, which failed with:

> The coalesced trailing run did not execute.

The failure is a **test-timing bug, not a production defect**. `BackgroundWorker`'s async gate (added for #1136) is correct: no signal is lost and no overlap occurs.

## Root cause

The test suspends the first async run mid-`await`, fires 20 `Start()` signals while it is suspended, releases the run, then waited for the worker to settle with a **single sample**:

```csharp
WaitUntil(() => !worker.IsRunning && Volatile.Read(ref active) == 0);
Assert.True(Volatile.Read(ref totalRuns) >= 2, "The coalesced trailing run did not execute.");
```

The trailing run is delivered one of two ways, both correct:
- **coalesced** — a burst callback sets `_runPending` before the first run's final lock, so the loop makes an extra pass; or
- **fresh** — the burst callback runs *after* the first run already exited, so it starts a new run.

Under CI load the first run's continuation can complete before the burst callbacks are dispatched. The worker then transiently reaches `!IsRunning && active == 0` while `totalRuns` is still `1`, and the fresh trailing run happens *after* that window. `WaitUntil` returns in the gap, and `Assert.True(totalRuns >= 2)` fires with the reported message.

A related, narrower race exists on the terminal `Assert.False(worker.IsRunning)`: the `AutoResetEvent` can retain one leftover signal from the burst that dispatches a late pass just after the worker first looks idle.

## Fix

Wait for a **stable** non-running terminal state with the trailing run already observed, instead of sampling once:

```csharp
WaitUntilStablyNotRunning(worker, () => Volatile.Read(ref active) == 0 && Volatile.Read(ref totalRuns) >= 2);
```

`WaitUntilStablyNotRunning` requires `!IsRunning` plus the supplied condition to hold across several consecutive polls, so neither the transient between-runs gap nor a leftover pass can be mistaken for the settled state. All three original assertions are kept and are now race-free (`totalRuns` is monotonic and the state is stable when they run).

Only the test changed — `BackgroundWorker.cs` is untouched.

## Verification

- Built a stress harness that reproduces the exact interleaving (forcing the first-run continuation to win the race). The **old** condition would have flaked in **410/500** worst-case iterations; the **new** wait passed **1000/1000** (500 worst-case + 500 realistic).
- The fixed test passed **50/50** runs, including **20/20 under full CPU load**.
- Full `Game.Infrastructure.Tests` suite: **60/60 passing**.

---
_Generated by [Claude Code](https://claude.ai/code/session_016WVZoUYcBr6WjgdYW1VaSw)_

### DIFF
--- a/Game.Infrastructure.Tests/BackgroundWorkerTests.cs
+++ b/Game.Infrastructure.Tests/BackgroundWorkerTests.cs
@@ -383,7 +383,12 @@ namespace Game.Infrastructure.Tests
             }
 
             releaseFirstRun.SetResult();
-            WaitUntil(() => !worker.IsRunning && Volatile.Read(ref active) == 0);
+            // The mid-run signals must produce at least one trailing run — coalesced into the in-flight run, or
+            // dispatched as a fresh run once it exits — and the worker must then settle for good. Wait for a *stable*
+            // non-running terminal state with that trailing run observed: sampling once could otherwise land in the
+            // brief gap after the first run exits but before the coalesced/leftover pass runs, reading totalRuns == 1
+            // or a transient IsRunning.
+            WaitUntilStablyNotRunning(worker, () => Volatile.Read(ref active) == 0 && Volatile.Read(ref totalRuns) >= 2);
 
             // Two loopAction calls never ran at once.
             Assert.Equal(1, Volatile.Read(ref maxActive));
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: Fix flaky `AsyncWorker_SignaledRepeatedlyDuringRun` infrastructure test (#1445)

**Verdict:** Approve — test-only change with a correct root-cause diagnosis and a sound fix; no production code touched. (Posted as `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR.)

The diagnosis is right: the old `WaitUntil(() => !worker.IsRunning && active == 0)` sampled the settled state once and could return in the brief gap after the first run exits but before the coalesced/leftover trailing pass runs — reading `totalRuns == 1` or a transient `IsRunning`. `BackgroundWorker.cs` itself is untouched and correct.

The `WaitUntilStablyNotRunning` helper is the right shape:
- It requires `!IsRunning && active == 0 && totalRuns >= 2` to hold across 10 consecutive polls, which rejects both the between-runs gap and a leftover late pass.
- No false-timeout risk: `totalRuns` is monotonic and `active`/`IsRunning` reach a permanent terminal state once settled, so the stable condition is guaranteed to hold within `WaitTimeout`.
- All three original assertions (`maxActive == 1`, `totalRuns >= 2`, `!IsRunning`) are preserved and are now race-free.

### Test coverage
- Test-only PR; it strengthens an existing test rather than adding behavior. Classical style is preserved (real `NullLogger`, observable-behavior assertions), consistent with `docs/backend.md`.

### What's good
- Braces on all blocks, clear naming, and `Assert.Fail` on timeout consistent with the existing `WaitUntil`.
- The multi-line comments earn their length — they capture non-obvious concurrency rationale that a future edit could otherwise naively "simplify" and reintroduce the flake.

CI was still in progress at review time; will merge once checks are green.

---

Issue #1444, opened 2026-07-01
Content seeder: handle Tag catalogue for item/mod tag-join rows


ginderjeremiah:
## Context

Follow-up from PR #1431 (JSON-driven content seeder, issue #1419, spike #1390).

The content export includes item/mod **tag ids** (`ItemTag` / `ItemModTag` join rows), but `Tag` itself is **not** one of the 10 exported/seeded static sets. The seeder's join-row insertion therefore assumes the referenced `Tag` rows already exist in the target DB.

This is **harmless today** because the committed content has no items or item-mods (`content/items.json` and `content/item-mods.json` are both `[]`), so no tag-join rows are seeded and nothing is stranded. But once item/mod content with tags lands, seeding a fresh DB would produce dangling FK references (or fail the insert).

## Proposed work

Before tag-carrying item/mod content is authored, do one of:

1. Export `Tag` as an 11th content set and seed it ahead of items/item-mods in FK order; or
2. Make the seeder's tag-join insertion tolerant of a missing/partial `Tag` catalogue (skip + warn, or validate up front with a clear error).

Option (1) is the more complete fix and keeps the seeder strict; (2) is a smaller guard. Recommend (1) if/when tags become real content.

## Notes
- No user-facing behavior change; purely seeding-path correctness for future content.
- Surfaced during review of #1431 and already noted in that PR's description.

_Follow-up filed during automated PR review._

ginderjeremiah:
Closing as a duplicate of #1432 (see #1433 for the full explanation) — already fixed and merged in `54f0904` (PR #1446).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1440, opened 2026-07-01
Seed tag content or guard tag-join FKs in the content seeder


ginderjeremiah:
## Context

Follow-up from PR #1431 (JSON-driven content seeder that reconstructs the static reference graph on a fresh DB from the content export; closes #1419; spike #1390).

The content export carries item/mod **tag ids** (`Item.Tags`, `ItemMod.Tags`), but tags themselves are **not** among the 10 exported static sets. The seeder therefore builds `ItemTag` / `ItemModTag` join rows assuming the referenced `Tag` rows already exist on a freshly-migrated database.

Today this is **dormant and harmless** — the committed `content/items.json` and `content/item-mods.json` are both empty (`[]`), so no tag-join rows are seeded and no `Tag` FK is referenced. But it's a latent FK failure the moment tagged item/mod content is authored and exported.

## The follow-up

Before tagged content lands, close the gap by one of:

1. **Export tags as an 11th static set** and seed them ahead of items/mods (referenced-before-referencing order), or
2. Make the seeder **tolerant / validating** — verify referenced `Tag` ids exist (or are seeded intrinsically via `HasData`) and fail with a clear message if not.

Option 1 is the more complete fix and keeps the export self-contained.

Labels: enhancement, claude, scope: small

ginderjeremiah:
Closing as a duplicate of #1432 (see #1433 for the full explanation) — already fixed and merged in `54f0904` (PR #1446).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1433, opened 2026-07-01
Content seeder: handle Tag content (tags aren't among the exported sets)


ginderjeremiah:
Follow-up from #1419 (PR #1431), part of the content-authoring-pipeline spike #1390.

## Problem

The content seeder (`ContentSeeder`) builds `ItemTag` / `ItemModTag` join rows straight from the item/mod read contracts' `Tags` id lists:

```csharp
var itemTags = content.Items
    .SelectMany(item => item.Tags.Select(tagId => new EntityItemTag { ItemId = item.Id, TagId = tagId }))
    .ToList();
```

But `Tag` is **not** one of the 10 exported static sets (tags carry their own non-zero-based identity and are resolved by lookup, not by index — see backend.md → "Tags are the deliberate exception"). So seeding a tag-join row assumes the referenced `Tag` rows already exist in the fresh DB. They don't — a freshly-migrated DB has no tags.

Today nothing is stranded: the committed `content/*.json` has no items or item-mods (`items.json` / `item-mods.json` are `[]`), so no tag-join rows are ever built. The gap only bites once tag-bearing item/mod content is authored and exported.

## Options

1. **Export tags as an 11th set** and seed them first (before items/mods), advancing/handling their non-zero-based identity appropriately. Keeps the export a complete snapshot.
2. **Make the seeder tolerant** — resolve/create the referenced `Tag` rows on the fly, or skip unknown tag ids with a loud diagnostic.

Option 1 is the more faithful "the export reproduces the whole graph" direction and matches how the other sets work; option 2 is lighter but leaves the export incomplete. Lean toward (1) unless there's a reason tags should stay out of the export.

## Acceptance
- Seeding a fresh DB from an export that includes tag-bearing items/mods reconstructs the tag catalogue and the join rows with cross-refs resolving.
- The round-trip drift guard covers tag content.

## Notes
- Blocked in practice only by the arrival of tag-bearing content, but worth landing before that so the first tagged item doesn't fail the seed.

ginderjeremiah:
[...]

PR #1446, opened 2026-07-01
Export tags as the 11th content set so the seeder can resolve tag joins (#1432)

### LINKED ISSUE #1432: Content seeder: tags aren't part of the export, so item/mod tag joins assume pre-existing Tag rows
Follow-up from #1419 (PR #1431).

The content export (#1418) serializes an item's/mod's **tag ids** (`Item.Tags`, `ItemMod.Tags`), but `Tag` entities are **not** one of the 10 exported static sets — they carry their own (non-zero-based) identity and are hard-deleted, so they were deliberately left out of the export. The JSON-driven content seeder therefore inserts `ItemTag`/`ItemModTag` join rows straight from the contract tag ids, which assumes the referenced `Tag` rows already exist in the target database.

Today this is dormant: the committed `content/*.json` has no items/mods (and no tags), so no join rows are produced and nothing is stranded. But once tag content is authored, seeding a fresh DB would FK-fail (the join references a `Tag` that a fresh DB doesn't have), or silently omit tags if we make it tolerant.

## Options
- **Export tags too** — add `Tag` (and `TagCategory` if needed) to the export + seeder as an 11th set. Tags aren't zero-based/contiguous, so the seeder's explicit-id path already works for them, but the drift-guard/lint and cache assumptions for tags need checking.
- **Make the seeder tolerant** — skip a tag-join whose `Tag` id is absent, with a loud warning (weaker; loses data).

Prefer exporting tags so the fresh-DB seed is complete and the drift guard covers them.

## Acceptance
- A fresh DB seeded from an export that includes tagged items/mods resolves the tag joins.
- The export + drift guard cover tags (or a documented decision on why they stay out).

ginderjeremiah:
## Summary

The content export serialized item/mod **tag ids** (`Item.Tags`, `ItemMod.Tags`) but never exported the `Tag` catalogue itself, so `ContentSeeder` built `ItemTag` / `ItemModTag` join rows assuming the referenced `Tag` rows already existed on a freshly-migrated database. They don't — I verified that only `TagCategory` is intrinsic (enum-seeded via `HasData`); `Tag` is authored, **non-zero-based** content resolved by lookup, so a fresh DB has zero tags. The gap was dormant only because the committed `content/items.json` / `content/item-mods.json` are empty; the first tagged item/mod would FK-fail the seed.

This implements **option 1** from the issue (the direction all four duplicate issues recommend and the more complete "export reproduces the whole graph" fix): export `Tag` as an 11th content set.

## How it addresses the issue

- **Contract + reader:** `Tags` added to `ContentImport` and a `tags.json` read into it.
- **Exporter:** `ContentExporter` now sources tags straight from the DB (they aren't held in an in-memory reference cache), so `ExportAll` becomes `ExportAllAsync`. Its only caller is the drift-guard test. `tags.json` is emitted before `item-mods.json`/`items.json` (referenced-before-referencing order).
- **Serializer:** a `Canonicalize` overload orders tags by id (stable, total order despite the non-zero-based identity).
- **Seeder:** the `Tag` catalogue is inserted ahead of the item/mod tag-join rows. The existing `EntityRowInserter.AdvanceIdentitySequencesAsync` already advances `Tag`'s non-zero-based sequence past its seeded max, so a later admin-authored tag won't collide.
- **`TagCategory` stays intrinsic** (enum/`HasData`), so per the intrinsic-vs-static rule in `backend.md` it is *not* exported — only `Tag` is authored content.
- Committed a `content/tags.json` (`[]` today) so the reader and drift guard resolve it.
- Updated `docs/backend.md` (content-export section) to list `tags` and note its non-cached / non-zero-based nature.

## Design note

The one real fork in the issue (export tags vs. a tolerant seeder) is resolved toward **option 1**, matching the recommendation repeated across the duplicate issues. The only implementation decision I made independently is excluding `TagCategory` from the export because it is intrinsic — flagging it here in case the intent was to export it too.

## Test plan

- [x] `dotnet test Game.Application.Tests` — all 773 pass, including the content-export **drift-guard round-trip** (seed → re-export → byte-equality, now covering `tags.json`).
- [x] New `SeedAsync_TaggedItemMod_SeedsTagCatalogueAndResolvesJoin` — a tagged mod now seeds and the join resolves (this FK-failed before the fix).
- [x] New `Canonicalize_Tags_OrderTopLevelById` — tag id ordering.

Closes #1432. Also resolves the duplicate follow-ups #1433, #1440, and #1444 (same gap).

---
_Generated by [Claude Code](https://claude.ai/code/session_01BT2NwWFjwuMB8Tc5rasFtc)_


ginderjeremiah (commented):
## Review: Export `Tag` as the 11th content set so the seeder can resolve tag joins (#1432)

**Verdict:** Approve — the fix is correct, consistently threaded through every layer of the content-export pipeline, well-tested, and documented. No blocking findings.

_(Posted as a `COMMENT` because GitHub won't accept an approving verdict on one's own PR — the substantive call is Approve.)_

I verified the change end-to-end rather than trusting the description:

- **Whole-graph consistency.** `tags` is threaded through `ContentImport`, `ContentImportReader`, `ContentExporter`, `ContentExportSerializer.Canonicalize`, the new `TagMapper`, and `ContentSeeder` — no half-wired set. `TagMapper.ToEntity` maps all three `Contracts.Tag` fields (`Id`/`Name`/`TagCategoryId`), so nothing is dropped in the round-trip.
- **Seeder ordering is correct.** `Tag` is inserted after proficiencies and before item-mods/items, so both `ItemModTag` and `ItemTag` join rows land after the catalogue exists (`ContentSeeder.cs:81`, join inserts at `:85`/`:90`). The referenced `TagCategory` is intrinsic (migration-seeded), so it's present on a fresh DB.
- **Non-zero-based identity handled.** `Tag` flows through the `Insert<T>` helper, so it's in `insertedTypes` and `AdvanceIdentitySequencesAsync` advances its sequence past the seeded max — a later admin-authored tag won't collide. `Canonicalize` orders by id (stable total order despite the non-index identity), matching the reader/drift-guard expectations.
- **`ExportAll` → `ExportAllAsync`.** The only caller (the drift-guard test) is updated; confirmed via search. Materializing `_tags.All()` (a projection to a non-entity contract, so untracked) into a list before assembling the otherwise-synchronous export is a clean way to keep the DB read out of the cached path.
- **Layering / DDD.** Mapping stays in `Game.DataAccess/Mapping`; the exporter reads contracts via the repo; no domain logic leaks. `ITags`/`IContentExporter` are both DI-registered, so the new dependency resolves.
- **Design fork resolved sensibly.** Option 1 (export tags), excluding intrinsic `TagCategory` — matches the intrinsic-vs-static rule in `backend.md`, and the author flagged the one independent call in the description.

### Test coverage
- `SeedAsync_TaggedItemMod_SeedsTagCatalogueAndResolvesJoin` pins the actual bug (join FK-failed pre-fix); `Canonicalize_Tags_OrderTopLevelById` pins ordering; the drift-guard round-trip now covers `tags.json`. Good coverage for a change this size.

### Nits (optional)
- The **seed** direction is now tested with a tagged mod, but the **export** direction isn't asserted against a non-empty tag set — the drift-guard round-trip only exercises it with the empty committed `tags.json`. Since the bug was a seed/export asymmetry, a symmetric "export a DB containing tags → tags.json content" test would round it out. `Canonicalize` + the round-trip structure cover the mechanism adequately, so this is take-it-or-leave-it.

### What's good
- Documentation (`backend.md`) updated in the same PR, including the non-cached/non-zero-based nature of tags.
- The design note calling out the `TagCategory` exclusion decision, and the code comments explaining *why* tags are seeded ahead of the join rows, are exactly the right level of context.


---

Issue #1448, opened 2026-07-01
Crit tally: full-hit → normalized-marginal (delivery-archetype reference implementation)


ginderjeremiah:
**Reference implementation for the delivery-archetype tally shape** — the worked example of the *normalized-marginal* tally, exactly as #1425 was the worked example of the *opt-in-multiplicative* shape. Full reasoning: [`docs/spikes/1398-utility-in-proficiency-system.md`](docs/spikes/1398-utility-in-proficiency-system.md) → *Follow-on refinement — the overlay tally shape*. Part of #1398.

## Problem

Crit currently books the **full** post-mitigation hit to the Precision signal:

```csharp
Stats.CriticalDamageDealt += totalNet;   // BattleContext.DamageTarget — the whole crit hit, not the bonus
```

That is a **full-hit** tally, not the **marginal** crit *bonus*. Because it folds in the base damage that would have landed on a normal hit, its magnitude is only weakly proportional to crit-damage investment — a `1.55×` and a `1.9×` crit train Precision at nearly the same rate. The accrual "double-book" into the type path is **harmless** (accrual is non-zero-sum — each path independently claims `pie × clamp(activity ÷ power)`), so this is about **strength-proportionality**, not double-counting.

## Change — migrate Precision to the normalized marginal

Per the four resolutions in the spike (§ *Follow-on refinement — the overlay tally shape*):

- **Marginal, not full-hit.** Tally the crit *bonus* (the extra over a non-crit hit), measured against a **fixed baseline** — add-one-in, `N(crit only) − N₀` — **never** leave-one-out from a fully-amplified state, so it still composes correctly once Hex/Sunder/Momentum are also amplifying a hit (leave-one-out would double-count the multiplicative synergy into every overlay).
- **Absolute marginal, normalized on the investment.** Tally the absolute enabled damage `D × (m−1) × mitigation`, and apply the concave-saturating normalization `φ` to the **investment** `(m−1)` — not to an enemy-dependent amplification ratio.
- **Parity-critical.** The tally is a side-channel to the health math, but the crit/dodge/reflect tallies are mirrored in `battler.ts` for the stats display, so this must be implemented **identically frontend↔backend** with matching test scenarios (the battle-parity rule). Cover: a `0`-chance build (never crits, trains nothing), a minimally-invested crit, and a heavily-invested crit (verifying the saturation).

## Open knobs — decide as part of this issue

[...]

Issue #1429, opened 2026-07-01
Delivery archetype: Sunder (anti-mitigation)


ginderjeremiah:
Delivery archetype from spike #1398 (see [`docs/spikes/1398-utility-in-proficiency-system.md`](docs/spikes/1398-utility-in-proficiency-system.md) → *Follow-on refinement*, and `docs/content-design.md` §4). A **committed, ready-to-implement** engine feature — it follows the opt-in-multiplicative template (#1425) and the **normalized-marginal tally** whose worked reference is the crit migration #1448.

## Fantasy

The **anti-tank** offense: **lower** the enemy's mitigation rather than punching through it. Strongest exactly where crit and DoT are weakest — bulky, high-`Toughness` enemies.

## Shape (per the template)

- **Enabler (opt-in):** a sunder effect — a debuff applying a **negative `Toughness`** (and/or resistance) modifier to the opponent. Expressible on today's effect system.
- **Technique path:** *Sunder* — multiplicative, inert without a sunder enabler.
- **Tallied magnitude:** the **mitigation bypassed** — the normalized **marginal** extra damage that landed because the enemy's mitigation was reduced, vs. the un-sundered baseline (the shared overlay tally shape, below).

## Attribution — the shared overlay tally shape (settled)

Sunder is the **same computation as Hex** (#1427), just on `Toughness`/resist instead of resistance — and `ComputeNetDamage` already composes resistance × toughness, so one mechanism serves both. It uses the normalized-marginal tally settled in [spike #1398 → *Follow-on refinement*](docs/spikes/1398-utility-in-proficiency-system.md#follow-on-refinement--the-overlay-tally-shape-normalized-marginal) and implemented first as the crit reference #1448:

- **Marginal, not full-hit** — the extra vs. an un-sundered hit. *(Correction to the original framing: the "double-counts against the type paths, the trap crit had" reasoning was wrong — the double-book is harmless under the non-zero-sum accrual; marginal is needed for strength-proportionality.)*
- **Add-one-in against a fixed baseline**, not leave-one-out — so Sunder's tally doesn't inflate because you also crit/hex.
- **Absolute damage enabled, normalized on the debuff strength** — so Sunder trains consistently regardless of the enemy's base mitigation.

The `Toughness`/resist debuff mechanic itself exists; the tally copies #1448.

## Dependencies / sequencing

[...]

Issue #1398, opened 2026-06-30
Determine if utility should be worked into the proficiency system


ginderjeremiah:
For example, a skill that exists only to buff CDR does not contribute towards any proficiency.

The open questions are:
- Should it contribute to a proficiency path?
- What benefits would said proficiency path provide?
- How is the contribution codified?

ginderjeremiah:
Ideation complete. Full writeup: [`docs/spikes/1398-utility-in-proficiency-system.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/1398-utility-in-proficiency-system.md).

**Outcome:** "utility" was the wrong unit of analysis. A proficiency path is earned by a **committed** build identity — reached through an **opt-in enabler** with a real opportunity cost, measured as *excess over a floor that is naturally `0` when uncommitted* (expectation-0). "Universal vs committed" is the gate; a buff/debuff is just the delivery *mechanism*.

- **The filed CDR-buff example → no path.** It is a universal amplifier; its payoff is the paths it amplifies (and, for cadence, farm throughput). Universal residue (flat CDR, all-damage%, raw EHP) must be authored as conditional/tradeoff effects, never free value.
- **Delivery archetypes follow one template** — flat enabler (items/skills, no attribute base) + multiplicative Technique path + tallied magnitude. Crit is reworked to this shape as the worked example; DoT/Reflection already fit; Momentum/Hex/Sunder are candidate content; Cull needs an engine conditional.
- **Attribute-concentration** ("all-in STR → STR synergies") is **attribute-scaling content**, not a proficiency track — the taxonomy stays type/delivery-based.
- **Frequency/CDR proficiency** is deferred, reachable only by severing CDR's attribute sourcing (its `1.0` base blocks the opt-in trick) — folded into the attribute spike.

**Issues created:**
- #1425 — Crit rework: opt-in multiplicative crit chance (the worked example)
- #1426 — Spike: attribute identity-sourcing (what DEX/LUK/AGI become once identities are opt-in)

**Docs updated:** `docs/content-design.md` (§4 delivery-archetype/commitment rule, §5 secondary-root rule, parking-lot entry resolved). `docs/game-design.md` crit section updates land with #1425.

Ready to close once reviewed — sub-issues track the remaining work.

ginderjeremiah:
Candidate delivery-archetype issues also created (deferred — built when the content arc reaches them, per the roster in content-design.md §4):
[...]

PR #1452, opened 2026-07-01
Add Hex delivery archetype: vulnerability-enabled damage tally (#1427)

### LINKED ISSUE #1427: Delivery archetype: Hex (vulnerability / debuffer)
Delivery archetype from spike #1398 (see [`docs/spikes/1398-utility-in-proficiency-system.md`](docs/spikes/1398-utility-in-proficiency-system.md) → *Follow-on refinement*, and `docs/content-design.md` §4). A **committed, ready-to-implement** engine feature — it follows the opt-in-multiplicative template (#1425) and the **normalized-marginal tally** whose worked reference is the crit migration #1448.

## Fantasy

The **debuffer**: apply **vulnerability** (raise the enemy's damage-taken) so everything after hits harder — a curse-caster / enabler identity. This is also the home for **debuff-utility** skills: a debuff that lowers enemy resistance has a real, measurable downstream magnitude.

## Shape (per the template)

- **Enabler (opt-in):** a vulnerability effect — a debuff applying **negative resistance** to the opponent (already supported: "negative resistance — vulnerability — falls out for free").
- **Technique path:** *Hex* — multiplicative, inert without a vulnerability enabler.
- **Tallied magnitude:** the **damage enabled by applied vulnerability** — the normalized **marginal** extra damage vs. the un-hexed baseline (the shared overlay tally shape, below).

## Attribution — the shared overlay tally shape (settled)

This was previously flagged as a Hex-specific "design crux." It is **not** Hex-specific — it's the same normalized-marginal tally every overlay archetype uses (crit included), settled in [spike #1398 → *Follow-on refinement*](docs/spikes/1398-utility-in-proficiency-system.md#follow-on-refinement--the-overlay-tally-shape-normalized-marginal) and implemented first as the crit reference #1448. Apply that rule here:

- **Marginal, not full-hit** — the extra vs. an un-hexed hit. *(Correction to the original framing: the old "double-counts against the type paths — the exact trap crit's full-hit tally had" reasoning was wrong. The double-book is **harmless** under the non-zero-sum accrual; marginal is needed for **strength-proportionality** — so a token debuff doesn't train Hex as fast as a committed one.)*
- **Add-one-in against a fixed baseline** — measure Hex's contribution against the vanilla hit, not leave-one-out from a fully-amplified state, so Hex's tally doesn't balloon just because you also crit.
- **Absolute damage enabled (`D × v`), normalized on the debuff strength `v`** — a resist debuff's *multiplier* is hyperbolic in enemy resistance, but the absolute enabled damage is flat (`D × v`, the base resist cancels), so Hex trains **the same against a soft or a resistant enemy** — no resist-farming exploit.

The vulnerability mechanic itself already exists (negative resistance — sampled live for DoT, applied to direct hits); the tally copies #1448.

## Dependencies / sequencing

- **Depends on the crit reference #1448** landing the normalized-marginal tally pattern, so this copies a proven implementation rather than re-deriving attribution.
- **Not gated on content** — building the mechanic does not require a zone. **Content placement** (which zone/challenge introduces a vulnerability enabler, and its authored strength) follows the arc as a separate authoring step, carrying one tuning watch-point: a debuff strong enough to zero a resist wall trivializes a resist-gated boss.

## Notes

- Resolves the debuff-utility half of the original #1398 question: a debuff is a *committed* archetype (opt-in enabler), so it earns a path — unlike a universal buff.
- **Shares its attribution mechanism with Sunder (#1429)** — the identical computation on `Toughness`/resist. Pick them up together, or Hex-then-Sunder.

Part of #1398.

ginderjeremiah:
## Summary

Implements the **Hex** delivery archetype (spike #1398) — the second overlay proficiency to adopt the normalized-marginal tally shape the crit reference (#1448) established. Hex trains on the extra damage a **player-applied vulnerability** (a negative-resistance debuff on the opponent) let through.

This is the engine feature only — a committed, ready-to-implement mechanic per the issue thread. **Content placement** (which zone/challenge introduces a vulnerability enabler, and its authored strength) is a separate authoring step and is not part of this PR.

## How it addresses the issue

Per the settled overlay tally shape, each hit/tick books the marginal damage the applied vulnerability enabled, normalized on the debuff strength:

- **Enabler:** a negative-resistance debuff on the opponent — already supported by the effect system, no new mechanic.
- **`EActivityKey.Hex`** (append-only, ordinal 30) + **`BattleStats.HexBonusDealt`**, wired into `ProficiencyRewardService`'s event accrual alongside crit/dodge/heal/reflect.
- **Tally:** `marginal / (1 + v)`, the *same* `/(1 + investment)` saturation the crit bonus uses — since `marginal/(1+(m−1)) ≡ baseline·φ(m−1)`, the two overlays share one rule. The marginal is `dealt × v × toughnessFactor` (direct) / `preMitigation × v` (DoT), so it is **flat in the enemy's base resistance** (no resist-farming) and proportional to the debuff strength. Covers **both** direct hits and the player's DoT on the enemy.
- Measured on the **pre-crit** hit so Hex and crit compose without either overlay inflating the other.

## Design note — the piece #1448 didn't need (please sanity-check)

The issue frames the tally as "copies #1448." That holds for the tally *shape*, but crit's investment `m` is a clean attacker-side scalar, whereas Hex's investment `v` is folded into the **defender's** single combined resistance modifier — indistinguishable from innate resistance once applied. To isolate the *player-applied* vulnerability I added one new piece: each `Battler` **snapshots its innate (pre-effect) resistance once at construction**. Durable resistance (base/gear/mods/proficiency/class) is already composed by then, so any live resistance below that baseline is exactly what an opponent debuffed; `v = innate − live` (clamped ≥ 0). This sidesteps tracking the folding/stacking/expiry of effects and works for free with them.

Design calls I made (all defensible, flagging for review):
- **Innate-below-baseline heuristic** attributes any resistance below the construction baseline to the player. An enemy *self-buffing* its resistance up reads as `v ≤ 0` (no credit, correct); a nonsensical enemy self-debuff would over-credit — no content does this, and it's documented.
- **`ComputeNetDamage` refactored** to share a resistance-parameterized core with the Hex counterfactual — byte-identical, so battle parity is untouched.
- Hex shares its mechanism with **Sunder (#1429)** (same computation on `Toughness`/resist); this PR sets up the pattern Sunder copies.

## Parity

Backend-only, like the crit bonus (the frontend simulator computes no battle stats). The innate snapshot and the two extra `ApplyDamageOverTime` recorders are backend-only side channels; `ComputeNetDamage` stays byte-identical, so the seeded health math and the parity matrix are unaffected.

## Test plan

- [x] `Game.Core.Tests` — 1132 pass. New: no-vulnerability, innate-vs-applied vulnerability (innate weakness trains nothing), normalized marginal, **flat-in-base-resistance** (soft vs resistant enemy bank the same), saturation ceiling, crit+hex independent bonuses, enemy-attacking books nothing, and DoT via `ResolveDamageOverTime`.
- [x] `Game.Application.Tests` — 780 pass. New: `HexBonus_TrainsTheHexPath` + `NoHexBonus_TrainsNoHex` (accrual routing).
- [x] Frontend `npm run lint` — 0 errors; progression-helpers test updated for the new event-key label (21 pass).
- [x] `enums.ts` regenerated via codegen (not hand-edited).

Closes #1427

https://claude.ai/code/session_01SuQABFrV3BCHKknqj8gAMY

---
_Generated by [Claude Code](https://claude.ai/code/session_01SuQABFrV3BCHKknqj8gAMY)_


Game.Core/Attributes/DamageTypes.cs
```diff
@@ -129,9 +129,19 @@ public readonly record struct KeyInfo(
...
+        // (Hex vulnerability isolation, #1427); the amplification-only weapon keys drop out (null Resistance).
+        private static readonly IReadOnlyList<EAttribute> ResistanceAttributeList =
+            KeyInfos.Select(info => info.Resistance).OfType<EAttribute>().ToArray();
+
         /// <summary>The leaf damage types, in enum order.</summary>
         public static IReadOnlyList<EDamageType> LeafTypes { get; } = Enum.GetValues<EDamageType>();
 
+        /// <summary>Every distinct resistance attribute (one per non-weapon key), in canonical key order — the set
+        /// a battler snapshots at construction for the Hex innate-resistance baseline (#1427).</summary>
+        public static IReadOnlyList<EAttribute> ResistanceAttributesAll => ResistanceAttributeList;
```
ginderjeremiah: Nit: the private `ResistanceAttributeList` field plus the public `ResistanceAttributesAll => ...` passthrough is one member more than needed — this can collapse to a single get-only auto-property with an initializer, matching the sibling `LeafTypes`/`WeaponLeaves` on this same class:

```csharp
public static IReadOnlyList<EAttribute> ResistanceAttributesAll { get; } =
    KeyInfos.Select(info => info.Resistance).OfType<EAttribute>().ToArray();
```

Not blocking, purely a consistency/DRY tidy-up.

---

Game.Core/Battle/Battler.cs
```diff
@@ -129,6 +164,42 @@ public double ComputeNetDamage(double dealt, EDamageType damageType, int attacke
...
+            }
+
+            var vulnerability = innateResistance - liveResistance;
+            if (vulnerability <= 0)
+            {
+                return 0;
+            }
+
+            var marginal = NetDamageAfterResistance(dealt, liveResistance, attackerLevel)
+                - NetDamageAfterResistance(dealt, innateResistance, attackerLevel);
```
ginderjeremiah: Nit / edge-case note: in the normal region (both `live` and `innate` nets positive) this marginal is exactly `dealt × v × toughnessFactor`, so the "flat in base resistance" property holds. But if the enemy's **innate** resistance is > 1 (authored absorption), `NetDamageAfterResistance(dealt, innateResistance, …)` returns a *negative* net (a heal, toughness not applied) while the live net is positive — so the marginal folds in the avoided heal and the flat-in-base-resistance guarantee no longer holds there. That's arguably the correct "damage the vulnerability enabled" (it turned a heal into a hit), and no content authors enemy absorption today, so I'm not treating it as a bug — but the crossover isn't covered by a test. Worth either a pinning test or a one-line note that the flat-in-base-resistance claim is scoped to the non-absorbing region.

---

ginderjeremiah (commented):
## Review: Hex delivery archetype — vulnerability-enabled damage tally (#1427)

**Verdict:** Approve (submitted as a `COMMENT` because I'm the PR author, so GitHub won't take a gating event from me — the real call is approve). Only minor nits below, none blocking.

This lands the second overlay archetype cleanly on the normalized-marginal tally shape the crit reference (#1448) established, and it faithfully implements the *settled* direction from the issue thread — engine-only, marginal (not full-hit), normalized on the debuff strength `v`, flat in the enemy's base resistance, measured pre-crit, covering both direct hits and the player's DoT. I verified the design decisions the PR flagged and they hold up.

### Correctness (verified)
- **Attribution isolation is sound.** The construction-time innate-resistance snapshot correctly isolates *player-applied* vulnerability: durable resistance (base/gear/mods/proficiency/class) is fully composed before the `Battler` ctor runs, and only in-battle skill effects move resistance afterward, so `v = innate − live` is exactly the opponent-applied debuff. Effect expiry falls out for free (live returns to innate → `v = 0`). The over-credit case (enemy self-debuffing resistance) is correctly documented as no-content.
- **No lookup can throw.** `ResistanceAttributesAll` (the snapshot key set) is every non-weapon key's `Resistance`, which is a guaranteed superset of anything `ResistanceAttributes(type)` returns (including the `Elemental`/`Dot` category keys and the weapon-leaf → `PhysicalResistance` fold), so the `_innateResistance[...]` reads are always present.
- **Marginal shape matches the docs.** Direct hit: `NetAfterResistance(live) − NetAfterResistance(innate)`, `/(1 + v)`; DoT: `preMitigation × v / (1 + v)` (toughness correctly bypassed). Both reduce to `dealt × v × toughnessFactor / (1+v)` in the linear region — flat in base resistance, as claimed.
- **Composes with crit correctly.** Hex reads the pre-crit `dealt`; crit reads its own baseline. The `DamageTarget_CritAndHex_BookIndependentBonuses` test pins that neither overlay inflates the other.
- **Player-offense-only** is enforced on both channels (direct hits gated on `_isPlayerActive`; DoT recorder supplied only for the enemy's DoT), and pinned by tests.

### Guideline compliance
- **DDD/layering:** all logic stays in `Game.Core`; `ProficiencyRewardService` only routes the new `HexBonusDealt` through the existing type-neutral event path — no orchestration logic leaked. Good.
- **Parity:** genuinely backend-only. `ComputeNetDamage`'s extraction into `NetDamageAfterResistance` is byte-identical, health math is untouched, and the frontend simulator computes no battle stats — so no FE battle mirror is required, consistent with the crit-bonus precedent. The only FE changes are the regenerated `enums.ts` (codegen, not hand-edited) and the workbench label + its test.
- **Enum:** `Hex = 30` is correctly append-only; persisted ordinals untouched.
- **Docs:** `backend-battle.md` and `game-design.md` both updated to describe the new signal accurately.

### Test coverage
Strong. `BattleContextTests` covers no-vulnerability, innate-vs-applied (innate weakness trains nothing), normalized marginal, flat-in-base-resistance (soft vs resistant bank the same), the `/(1+v)` saturation ceiling, crit+hex independence, enemy-attacking, and DoT with/without vulnerability. `ProficiencyRewardServiceTests` covers the accrual routing both ways. The one gap is the absorption crossover (innate resistance > 1) — see the inline note; low priority given no such content exists.

### Nits (non-blocking)
- `DamageTypes.ResistanceAttributeList` field + `ResistanceAttributesAll` passthrough property can collapse to one auto-property (inline).
- Absorption-crossover edge in `HexBonusForHit` is untested and the "flat in base resistance" claim is scoped to the non-absorbing region (inline).
- Minor: the player's own `_innateResistance` snapshot is always built but never read (Hex only credits the enemy target). Trivial once-per-battle cost; fine to leave for simplicity, just noting it.

### What's good
The innate-snapshot approach is an elegant way to sidestep tracking effect folding/stacking/expiry, and the shared `/(1 + investment)` saturation genuinely unifying crit and Hex under one rule (`marginal/(1+(m−1)) ≡ baseline·φ(m−1)`) is a nice result. Thorough tests and clear, well-scoped documentation throughout.

---

Game.Core/Battle/Battler.cs
```diff
@@ -45,6 +55,23 @@ public Battler(AttributeCollection attributes, IEnumerable<Skill> skills, int le
             CurrentHealth = _attributes[MaxHealth];
             Skills = skills.Select(s => new BattleSkill(s)).ToList();
             Level = level;
+            _innateResistance = SnapshotInnateResistance(attributes);
+        }
+
+        // Captures each resistance attribute's composed value once, before any timed effect can lower it — the
+        // baseline the Hex tally measures an applied vulnerability against (#1427). Durable resistance (base,
+        // gear, mods, proficiency, class) is already composed at this point; only in-battle skill effects move
+        // resistance afterward, so a live value below its snapshot is exactly the opponent-applied vulnerability.
```
ginderjeremiah: Does this still work if enemies can buff their resistance mid-battle?

ginderjeremiah: Yes — it stays safe, because the isolation reads the *net* live resistance against the construction-time innate snapshot rather than tracking modifier sources:

- **Enemy self-buffs resistance, no player debuff:** live sits *above* the innate baseline, so `v = innate − live < 0` → clamped to 0 → no Hex credit. An enemy hardening itself never trains the player's Hex.
- **Enemy self-buff + player debuff on the same resistance:** live ends up below innate by the *net* amount, and Hex credits that net (the extra damage that actually landed vs. the innate hit), not the player's gross debuff — the enemy's buff first had to be overcome.

So no crash and no over-credit; the only behavioural nuance is that a simultaneous enemy buff would credit the *net* reduction rather than the player's gross contribution. No enemy content buffs its own resistance today, and the innate-snapshot approach handles the hypothetical for free.

Pinned both cases with tests (`DamageTarget_EnemySelfBuffsResistance_NoPlayerDebuff_BooksNoHexBonus`, `DamageTarget_EnemySelfBuffThenPlayerDebuff_CreditsOnlyTheNetReductionBelowInnate`) and documented the interaction on `SnapshotInnateResistance`, including the note that if enemy resistance buffs ever become content *and* gross-contribution attribution is wanted, it would need to track the player-contributed resistance delta specifically (source-tagged) instead of the innate diff — happy to file that as a follow-up if you'd like. Fixed in fb8c533.

---

ginderjeremiah: The point I'm trying to get at is that an enemy who buffs their resistance causes players to not be able to train their hex proficiency as well. It seems like you are acknowledging that in your second bullet. Even if the _net_ is zero, a player's debuff is still doing work that should result in a contribution.

ginderjeremiah: Good call — you're right, the net-below-innate attribution shortchanges the player's debuff for work it genuinely did. Switched to crediting the **gross** contribution in `72428bb`.

Instead of diffing against a construction-time baseline, the attacker-applied resistance delta is now tracked directly: an opponent-applied additive resistance debuff accrues its signed amount onto the target's effect stack (`ApplyEffect(tracksVulnerability)`), so `v = AppliedVulnerability(type)` is exactly the resistance the debuff removed — independent of the target's base resistance **and** of any resistance the enemy buffs on itself. The marginal is now measured against the same hit *without the player's debuff* (live resistance raised back by `v`), still `marginal / (1 + v)`.

So for your example (enemy self-buffs +0.5, player debuffs −0.8): `v = 0.8` (the player's gross), and the tally credits `N(−0.3) − N(0.5) = 24`, ÷ 1.8 = **13.33** — the full value of the player's debuff, not the net 0.3. The enemy hardening itself no longer eats the player's Hex training.

Bonus: this is cleaner than the snapshot — the contribution rides the effect stack's shared expiry (clears for free when the debuff lapses) and never touches the combined modifier or health math, so it stays backend-only with no parity surface. Removed the innate snapshot and `ResistanceAttributesAll`; added `DamageTypes.IsResistanceAttribute` for the enabler gate. Updated the enemy-buff test to assert the gross contribution and added a vulnerability-expiry test. Full Core suite (1136) + accrual tests green.

---

Issue #1461, opened 2026-07-02
Toughness mitigation curve has an unguarded pole at Toughness = −K·attackerLevel


ginderjeremiah:
## Context

Surfaced while implementing Sunder (#1429, PR TBD) — the first feature that makes an opponent-applied `Toughness` debuff a first-class, designed mechanic. `Toughness` was previously only ever positive (derived `2 × Endurance`, no existing content debuffs it), so this was latent and unreachable; Sunder makes it directly reachable by content design.

## Problem

The shared, parity-critical Toughness mitigation curve —

```
toughnessReduction = Toughness / (Toughness + K·attackerLevel)
```

(`Battler.NetDamageAfterMitigation` in `Game.Core/Battle/Battler.cs`, mirrored byte-for-byte by `toughnessMitigatedDamage` in `UI/src/lib/battle/battle-formulas.ts`) is unclamped below `0` and has a genuine pole at `Toughness = −K·attackerLevel` (`K = GameConstants.ToughnessMitigationConstant`, currently `20`):

- For `Toughness` in `(−K·attackerLevel, +∞)` the curve is well-behaved and monotonically increasing from `−∞` toward `1`.
- As `Toughness → (−K·attackerLevel)⁺` the reduction → `−∞` (a hit gets amplified without bound).
- Crossing the pole (`Toughness < −K·attackerLevel`) flips the reduction to `> 1`, which can turn what should be "no mitigation" into a **net heal** on a positive hit — clearly not an intended outcome of a mitigation debuff.

A large enough single Toughness-debuffing effect (author error, or a deliberately abusive build once Sunder enablers exist in content) can push a target's live `Toughness` past this pole.

## Suggested fix

Floor the curve's `Toughness` input at `0` before computing the reduction (`effectiveToughness = Math.Max(toughness, 0)`), consistent with the "anti-mitigation" design intent — a debuff should be able to strip mitigation down to `0%`, never past it into bonus amplification or a heal. This needs to be applied **identically** on both simulators (`Battler.cs` and `battle-formulas.ts`) plus a `Mulberry32`/`BattleSimulatorParityTests`-style parity test covering the boundary, since this is parity-critical shared math.

## Why not fixed in #1429 directly

[...]

Issue #1430, opened 2026-07-01
Delivery archetype: Cull (execute / finisher) — needs damage-calc conditional


ginderjeremiah:
Delivery archetype from spike #1398 (see [`docs/spikes/1398-utility-in-proficiency-system.md`](docs/spikes/1398-utility-in-proficiency-system.md) → *Follow-on refinement*, and `docs/content-design.md` §4). A **committed, ready-to-implement** engine feature that follows the opt-in-multiplicative template (#1425) and the **normalized-marginal tally** (worked reference: the crit migration #1448). **It is the one archetype that also needs new damage-calc engine logic.**

## Fantasy

The **finisher**: bonus damage scaled to the enemy's **missing HP** (or gated on a low-HP threshold) — the last portion of a boss melts. Great against high-HP bosses, wasted on trash. A distinct target profile from crit / DoT / Momentum.

## Shape (per the template)

- **Enabler (opt-in):** an execute effect on gear/skills granting HP-conditional bonus damage — inert (`0`) without it.
- **Technique path:** *Cull* — multiplicative, inert without an execute enabler.
- **Tallied magnitude:** the **execute bonus dealt** (the HP-conditional bonus portion) — an overlay, so it uses the shared tally shape (below).

## Attribution — the shared overlay tally shape (settled)

The execute bonus is an outgoing overlay, so it uses the same normalized-marginal tally as crit/Hex/Sunder/Momentum, settled in [spike #1398 → *Follow-on refinement*](docs/spikes/1398-utility-in-proficiency-system.md#follow-on-refinement--the-overlay-tally-shape-normalized-marginal) and implemented first as the crit reference #1448: the **marginal** execute bonus, **add-one-in** against the no-execute baseline, as **absolute** enabled damage with the concave normalization on the bonus magnitude.

## Engine work (why this one is flagged separately)

- The damage calc has **no "scale by target's current/missing HP" conditional** today. Adding it is **parity-critical**: it runs in the seeded battle simulation, so it must be implemented identically frontend↔backend with matching test scenarios (the battle-parity rule).
- Everything else (path, enabler, tally, activity key) follows the template + the #1448 tally shape.

## Dependencies / sequencing

- **Depends on the crit reference #1448** for the tally pattern, **and** carries its own new damage-calc conditional (above) — so it is the heaviest of the four, but still ready to pick up once #1448 lands.
[...]

Issue #1460, opened 2026-07-02
Remove null-forgiving operator from Battler stack-scan helpers


ginderjeremiah:
`Battler.StackMomentumContribution` (#1459) and its pre-existing sibling `Battler.StackVulnerabilityContribution` (#1427) both open with `foreach (var stack in _attributeStacks!)`, relying on the caller (`AppliedMomentum` / `AppliedVulnerability`) to have null-guarded `_attributeStacks` first.

`CLAUDE.md` explicitly says to avoid the null-forgiving operator (`!`) and handle nullability properly. The compiler can't see the caller's guard across the method boundary, hence the `!`. Both helpers should be restructured to avoid it — e.g. an early `if (_attributeStacks is null) { return 0; }` inside each helper, or having the caller pass the already-verified non-null collection.

No user-facing effect — purely a maintainability cleanup covering both sibling scan helpers.

Surfaced during review of #1459.

ginderjeremiah:
Note from the #1462 (Sunder) review: once that lands there will be **three** of these near-identical `_attributeStacks` scan helpers — `StackVulnerabilityContribution`, `StackMomentumContribution`, and `StackSunderContribution` — backed by three parallel `*Contribution` fields on `AttributeEffectStack`. Beyond removing the `!`, the cleanest fix is now to consolidate all three into a single parameterized helper (e.g. `StackContribution(Func<AttributeEffectStack, decimal> selector)`), which also naturally resolves the null-guarding this issue is about. Worth folding into the same change rather than fixing the three helpers individually.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1462, opened 2026-07-02
Add Sunder delivery archetype: mitigation-bypassed damage tally (#1429)

### LINKED ISSUE #1429: Delivery archetype: Sunder (anti-mitigation)

ginderjeremiah:
## Summary

Implements the **Sunder** delivery archetype (spike #1398) — an overlay proficiency, and the anti-mitigation counterpart to Hex (#1427). Sunder trains on a player-applied `Toughness` debuff against an enemy.

This is the engine feature only — a committed, ready-to-implement mechanic per the issue thread. **Content placement** (which zone/challenge introduces a Sunder enabler) is a separate authoring step and is not part of this PR.

## How it addresses the issue

- **Enabler:** an opponent-targeted additive `Toughness` debuff — already expressible on the effect system, no new mechanic.
- **`EActivityKey.Sunder`** (append-only, ordinal 33 — after Cull #1430, which landed on `main` while this was in flight) + **`BattleStats.SunderBonusDealt`**, wired into `ProficiencyRewardService`'s event accrual alongside crit/Hex/Momentum/Cull.
- **Applied-Sunder tracking:** `Battler.AppliedSunder()` + a new `AttributeEffectStack.SunderContribution` field, set via `ApplyEffect(tracksSunder: true)` — the same shared-expiry, no-parity-surface side channel Hex's `VulnerabilityContribution` and Momentum's `MomentumContribution` use.

## Design: a proxy, not a literal marginal (settled after two review rounds)

The other overlays (crit, Hex, Momentum) tally a **literal before/after marginal** — the actual extra damage a hit dealt because of the debuff/buff, measured through the real damage-calc curve. That works cleanly for Hex because its debuff (resistance) is a flat, linear percentage: the target's own base resistance cancels out of the difference algebraically, so the marginal is genuinely flat in the target's stats (`dealt × v × toughnessFactor`).

Sunder's debuff is on `Toughness`, which is a **nonlinear, diminishing-returns curve**. I went through two designs trying to compute a literal marginal through it and normalize it to be enemy-independent — first normalizing on the curve's percentage-points removed, then (after review) normalizing on raw Toughness points scaled by `K·attackerLevel`. Both still had the same root problem: **the marginal itself**, not just its normalization, is inescapably a function of the target's own baseline Toughness and resistance when computed through a nonlinear curve — there's no algebraic cancellation available the way there is for Hex. That meant the same Sunder investment still trained faster or slower depending on which enemy it landed on, and cost two extra curve evaluations per hit to compute.

**Final design:** Sunder books `dealt × φ(investment)` directly — `dealt` is the same attacker-amplified, pre-crit, pre-mitigation value the other overlays already receive; `φ(a) = a/(1+a)`; the investment is the raw Toughness debuff scaled by `K·attackerLevel` (the curve's own characteristic magnitude). This is a **designed proxy**, not a measured before/after difference — but it's the only way to get genuine enemy-independence here, and it's cheaper than the marginal-based version (no counterfactual curve evaluation at all). New test `DamageTarget_SunderBonus_IsIndependentOfEnemyStats` pins that the same investment trains identically against a soft vs. a resistant, heavily-invested enemy, even though the real damage dealt differs sharply between them.

**Direct-hit only, unlike Hex:** DoT bypasses the Toughness curve entirely, so a Toughness debuff cannot affect it — there is no DoT counterpart to this tally.

## The Toughness-floor tangent (#1461 → #1463 → #1478)

My own follow-up #1461 (filed from this PR, flagging that the unclamped Toughness curve has a pole at `Toughness = −K·attackerLevel`) got picked up and merged separately as #1463, which **floored Toughness at 0** in the live damage math. That fix was wrong: driving an enemy's Toughness below 0 via a strong-enough debuff should amplify damage — that's the correct extension of "anti-mitigation" past "no mitigation," not a bug. Per direct correction, I've filed **#1478** to revert #1463 (intentionally left off the `claude` label — not scoped for this PR). This PR's rebase onto current `main` inherits #1463's floor as-is for the real damage math; Sunder's tally no longer reads the Toughness curve at all, so it's unaffected either way once #1478 lands.

## Test plan

- [x] `dotnet test` — all 4 backend suites green (Game.Core.Tests 1185, Game.Infrastructure.Tests 60, Game.Application.Tests 798, Game.Api.Tests 652)
- [x] `npm run lint` (eslint + svelte-check) — 0 errors
- [x] `npm run test:unit:ci` — 275 files / 3165 tests passing
- [x] Frontend codegen re-run (`dotnet run --project Game.Api -- codegen`) and diffed clean
- [x] `Game.Core.Tests/Battle/BattleContextTests.cs` coverage: no-debuff, dealt-scaled-by-investment, partial debuff (proportionality), saturation toward the full dealt amount, enemy-stat independence, crit+Sunder independence, expiry, enemy-attacking isolation, and a DoT no-op check
- [x] `Game.Application.Tests/Services/ProficiencyRewardServiceTests.cs` coverage: `SunderBonus_TrainsTheSunderPath` + `NoSunderBonus_TrainsNoSunder`
- [x] `docs/backend-battle.md` and `docs/game-design.md` updated for the mechanic
- [x] Also fixed a latent gap in `progression-helpers.ts`: `EActivityKey.Cull` (#1430) had no admin-Workbench label, so it would've been mis-grouped under "Damage dealt" instead of "Combat events" — added alongside Sunder's own label.

Closes #1429

---
_Generated by Claude Code_


Game.Core/Battle/Battler.cs
```diff
@@ -222,6 +230,88 @@ private double StackVulnerabilityContribution(EAttribute attribute)
...
+            var baselineToughness = liveToughness + sunder;
+            var marginal = NetDamageAfterMitigation(dealt, liveResistance, liveToughness, attackerLevel)
+                - NetDamageAfterMitigation(dealt, liveResistance, baselineToughness, attackerLevel);
+            if (marginal <= 0)
+            {
+                return 0;
+            }
+
+            var scaled = GameConstants.ToughnessMitigationConstant * attackerLevel;
+            var investment = baselineToughness / (baselineToughness + scaled) - liveToughness / (liveToughness + scaled);
```
ginderjeremiah: **Should-fix (DRY):** this line re-derives the Toughness-reduction curve `t / (t + scaled)` inline — twice — while `NetDamageAfterMitigation` (line 144) computes the same `toughness / (toughness + scaled)` expression. That's now three copies of the same parity-critical curve. CLAUDE.md calls out DRY explicitly, but the bigger reason is #1461: its fix floors the curve's input (`Math.Max(toughness, 0)`) in `NetDamageAfterMitigation`, which the `marginal` reads through — but this `investment` line computes the curve independently, so the floor would silently *not* apply here, leaving the two halves of the tally reading different curves.

Suggest extracting a small `private double ToughnessReduction(double toughness, int attackerLevel)` used by both `NetDamageAfterMitigation` and this line, so #1461's floor lands in one place and can't diverge. Not a blocker — the current behavior is correct and well-tested — but worth doing before #1461 rather than after.

---

ginderjeremiah (commented):
## Review: Sunder delivery archetype — mitigation-bypassed damage tally (#1429)

**Verdict:** Approve _(submitted as a comment — I'm the PR author, so GitHub won't take a formal APPROVE from me)_ — the mechanic is correct, faithfully mirrors the settled Hex/Momentum overlay pattern, is thoroughly tested, and the one genuinely novel design call (normalizing on mitigation-percentage-points rather than raw Toughness points) is the right one and well-justified.

I have no blocking findings. There's one DRY should-fix and a couple of nits below; none need to gate the merge, so I'm leaving them for you to take or decline rather than merging over an open thread.

### Design review

- **The Hex-divergence is correct.** Toughness feeds a diminishing-returns curve, not a flat percentage, so raw points removed aren't a comparable investment unit across enemies. Normalizing φ on the mitigation-percentage-points removed (curve reduction at baseline − at live) keeps the investment bounded like Hex's `v` and preserves the "weak still trains proportionally" goal. Good call, and correctly flagged for scrutiny.
- **Direct-hit-only is right.** DoT bypasses the Toughness curve entirely, so there's no marginal to credit — and `ResolveDamageOverTime_AppliedSunder_BooksNoSunderBonus` pins it. This is the correct asymmetry vs. Hex (which does have a DoT counterpart).
- **The `NetDamageAfterResistance` → `NetDamageAfterMitigation` refactor is a clean parameter extraction**, byte-identical to the former inline body, so battle parity is preserved.
- **The pole/inversion edge is handled responsibly.** The `marginal <= 0` and `investment > 0` guards keep the tally from manufacturing bogus credit in the curve's unclamped region, and the underlying live-damage fix is correctly scoped out to #1461 (well-written, correctly tagged `bug` / `scope: small`) rather than smuggled into a tally-only PR.

### Frontend parity — verified, no mirror needed

Battle logic normally must be mirrored in `battle-step.ts`, so I checked this explicitly: `grep` for any overlay tally (`HexBonus`/`MomentumBonus`/`SunderBonus`/`BonusForHit`) across `UI/src` returns nothing, and `backend-battle.md` documents these tallies as backend-only (the frontend computes no battle stats; the value ships from the server). Since the shared Toughness curve itself is unchanged, there's correctly no `battle-step.ts` change to make.

### Should-fix

- `Game.Core/Battle/Battler.cs:276` — the `investment` line re-derives the Toughness curve `t/(t+scaled)` inline (twice), duplicating `NetDamageAfterMitigation`. Extract a shared `ToughnessReduction(...)` helper so #1461's floor lands in one place and the two halves of the tally can't diverge. (Inline comment with detail.)

### Nits

- `Game.Core/Battle/Battler.cs` — the new `StackSunderContribution` null-checks internally and avoids the `!` operator, which is *better* than its sibling `StackVulnerabilityContribution` (line 222) that uses `_attributeStacks!`. Worth a small follow-up (tech debt) to bring the Hex helper in line with CLAUDE.md's "avoid `!`" rule — the new code sets the better precedent.
- `UI/src/tests/.../progression-helpers.test.ts` — the picker test asserts explicit labels for `Crit`/`Hex`/`Momentum` but not `Sunder`. The "groups every key" test already guarantees Sunder is present and grouped, so this is purely for symmetry; an `activityKeyLabel(EActivityKey.Sunder) === 'Mitigation damage enabled'` assertion would match the existing pattern.

### Test coverage

Strong and mirrors the established overlay pattern. `BattleContextTests` covers no-debuff, innate-low-Toughness (untracked), the normalized marginal, curve-inversion past zero (pinned to actual not idealized behavior), saturation under heavy investment, crit+Sunder independence, expiry, enemy-attacking isolation, and the DoT no-op. `ProficiencyRewardServiceTests` covers both the trains/doesn't-train paths. No EF migration is needed — `BattleStats` is an in-memory type, not a persisted entity.

### What's good

Exceptionally clear doc comments and PR writeup; the divergence from Hex is called out honestly and reasoned through, the pole edge case was discovered, defensively guarded, and filed as its own parity-scoped issue rather than hidden. This is a model example of copying an established pattern while thinking critically about where it shouldn't be copied verbatim.

---

Game.Core/Battle/Battler.cs
ginderjeremiah: Done in f0c75a4 — extracted `ToughnessReduction(toughness, attackerLevel)`, used by both `NetDamageAfterMitigation` and the Sunder `investment` line. Good catch on the #1461 divergence risk.

---

ginderjeremiah:
Addressed the should-fix (f0c75a4: extracted `ToughnessReduction` so #1461's eventual fix can't diverge between the two halves of the tally) and the frontend nit (added the explicit `Sunder` label assertion in `progression-helpers.test.ts`, matching the Crit/Hex/Momentum pattern).

The `!`-operator nit on `StackVulnerabilityContribution` is already tracked by #1460 (filed during the Hex/Momentum review), so I didn't duplicate it here — leaving that cleanup to that issue.

---

ginderjeremiah:
## Code review — Sunder delivery archetype (#1429)

**Verdict: approve on code quality** — but **not merging yet**: the branch now conflicts with `main` and needs a rebase (details below).

### Summary
Implements **Sunder**, the fourth overlay proficiency to adopt the normalized-marginal tally shape, faithfully copying the Precision/Hex/Momentum pattern: `EActivityKey.Sunder = 32` (append-only) + codegen'd `enums.ts` mirror, `BattleStats.SunderBonusDealt` wired into `ProficiencyRewardService`, `Battler.SunderBonusForHit` + `AppliedSunder()` + `AttributeEffectStack.SunderContribution` via `ApplyEffect(tracksSunder)`, and the `NetDamageAfterResistance` → `Toughness`-parameterized `NetDamageAfterMitigation` + extracted `ToughnessReduction` refactor. Correctly solves the issue; docs updated consistently with the sibling overlays.

### Battle parity
Backend-only training signal — no frontend parity surface (the client simulator computes no battle stats), same as crit/Hex/Momentum, so no frontend mirror is expected. The one shared-math touchpoint — the `NetDamageAfterMitigation`/`ToughnessReduction` extraction — was byte-identical to the prior inline body **at the time this branch was cut**, so the live curve was unchanged and `battle-formulas.ts` needed no edit. See the rebase note below, though.

### Design divergence (normalizing on mitigation percentage-points) — sound
Using *mitigation-percentage-points removed* rather than raw Toughness points as the `φ` investment is the correct call: raw points feed a diminishing-returns curve and have no bounded, cross-enemy-comparable scale, so `φ` would saturate almost immediately and defeat strength-proportionality. The percentage-points form stays bounded in `[0,1)` like Hex's `v`, `ToughnessReduction` is monotonic on `(−K·level, ∞)` so `sunder > 0 ⇒ investment ≥ 0`, and holding resistance constant across both counterfactual calls correctly yields zero credit for an already-absorbed hit.

### Test coverage — strong
`BattleContextTests` covers no-debuff, innate-low-Toughness (untracked), normalized marginal, curve inversion past 0 Toughness, heavy-investment saturation, crit+Sunder independence, expiry, enemy-attacking isolation, and the DoT no-op; `ProficiencyRewardServiceTests` adds the positive/negative accrual paths. Nullable handling is clean and there's no dead code.

### ⚠️ Blocking merge — rebase required, and it's not purely mechanical
Since this branch was cut, **#1463 ("Floor Toughness at 0 in the mitigation curve") merged to `main`** — that's the fix for the exact edge case this PR filed as #1461 (now resolved). #1463 modifies `Battler.NetDamageAfterMitigation`, so the branch is in conflict (`git merge main` → conflict in `Game.Core/Battle/Battler.cs`). Resolving it needs judgment, not just conflict markers:

1. **Fold the #1463 Toughness floor into the extracted `ToughnessReduction`/`NetDamageAfterMitigation`** so the live curve matches `main` (and stays in parity with `battle-formulas.ts`, which #1463 also floored).
2. **Reconcile the sub-pole guard in `SunderBonusForHit` and the "curve inversion past 0 Toughness" test** — with Toughness now clamped at 0, the pole/inversion region is no longer reachable through the live curve, so that guard is likely dead and the test's "pinned to the actual, not idealized, curve" expectation needs updating to the clamped behavior.
3. **Update the docs design note** ("unclamped below 0 … pole … can flip mitigation into bonus amplification") — that scenario is no longer content-reachable post-#1463, so the rationale for the defensive guard should be revised or removed.

Once rebased and green again, this is good to merge. Flagging rather than resolving it myself since it's parity-critical shared math on a branch I shouldn't be pushing to.

### Follow-up (already tracked)
The optional DRY nit — `StackSunderContribution` / `StackVulnerabilityContribution` / the Momentum equivalent are now three near-identical `_attributeStacks` scans with three parallel `*Contribution` fields — overlaps with the existing #1460 (`!`-operator cleanup on these same helpers). I've noted the triplication there rather than filing a duplicate; a single parameterized `StackContribution` helper would resolve both. No new issue needed.

---

ginderjeremiah:
## Automated code review — Sunder delivery archetype (#1429)

**Verdict: request changes** — the implementation is sound and CI was green when it ran, but the branch has gone **stale against `main` and now conflicts non-mechanically**. Not mergeable as-is.

### Summary
Faithful fourth copy of the normalized-marginal overlay tally (Precision #1448 / Hex #1427 / Momentum #1428): `EActivityKey.Sunder = 32` (append-only) + `enums.ts` codegen mirror + workbench label, `BattleStats.SunderBonusDealt` wired into `ProficiencyRewardService`, `Battler.SunderBonusForHit`/`AppliedSunder()`/`AttributeEffectStack.SunderContribution` via `ApplyEffect(tracksSunder)`, and the `NetDamageAfterResistance` → Toughness-parameterized `NetDamageAfterMitigation` + extracted `ToughnessReduction` refactor. Docs (`backend-battle.md`, `game-design.md`) updated consistently with the siblings.

### What's good
- **Design divergence is the right call.** Normalizing `φ`'s investment on *mitigation-percentage-points removed* rather than raw Toughness points is correct — Toughness feeds a diminishing-returns curve with no bounded cross-enemy scale, so raw points would saturate `φ` instantly and break strength-proportionality. The percentage-points form stays bounded in `[0,1)` like Hex's `v`; `ToughnessReduction` is monotonic on `(−K·level, ∞)` so `sunder > 0 ⇒ investment ≥ 0`, and holding resistance constant across both counterfactual calls correctly yields zero credit for an already-absorbed hit.
- **Refactor was byte-identical** to the prior inline body *at branch-cut time*, so live curve unchanged and no `battle-formulas.ts` edit was needed — correct for a backend-only tally (client computes no `BattleStats`).
- **Guards are adequate for the tally.** `marginal <= 0 → 0` and `investment > 0` together catch the past-pole net-heal region, so no bogus credit is minted even where the underlying curve misbehaves.
- **Test coverage is strong.** No-debuff, innate-low-Toughness (untracked), normalized marginal, curve inversion, heavy-investment saturation, crit+Sunder independence, expiry, enemy-attacking isolation, DoT no-op; plus positive/negative accrual in `ProficiencyRewardServiceTests`. Nullable handling clean, braces fine, no dead code.
- **CI:** all 9 checks green on `f0c75a4` (backend/frontend/e2e×3/verify).

### ⚠️ Blocking — non-mechanical rebase required
Since this branch was cut (`ca52c8f`), `main` merged **#1463 "Floor Toughness at 0 in the mitigation curve"** (the #1461 fix — now `Math.Max(_attributes[Toughness], 0)`), **#1460** (the stack-scan helpers collapsed into one parameterized `StackContribution(attribute, selector)`), and **#1466 (Cull)**. `git merge-tree origin/main` conflicts in **every file this PR touches** (`Battler.cs`, `BattleContext.cs`, `BattleStats.cs`, `ProficiencyRewardService.cs`, `Enums.cs`, both test files, `enums.ts`, both docs). Resolution needs judgment, not just markers:

1. **Fold the #1463 Toughness floor into `ToughnessReduction`/`NetDamageAfterMitigation`** so the live curve matches `main` and stays in parity with the (now-floored) `battle-formulas.ts`.
2. **The "curve inversion past 0 Toughness" test will fail post-floor** — with Toughness floored, live Toughness `−10 → 0`, so `DamageTarget_SunderBeyondZeroToughness_CurveInvertsPastZero` would compute 30 dmg / tally 10, not the asserted 60 / 18. Update it to the clamped behavior (or repurpose it as a "floor holds mitigation at 0%" test).
3. **`SunderBonusForHit`'s sub-pole guard is now largely dead** against the live curve — keep it as defense-in-depth but revise the doc note, and **update the design note** in `backend-battle.md`/`SunderBonusForHit` XML that still describes the "unclamped below 0 / pole / net heal" scenario as reachable (it no longer is post-#1463).
4. **Adopt `main`'s `StackContribution(Toughness, stack => stack.SunderContribution)`** instead of adding the new parallel `StackSunderContribution` scan (which #1460 already eliminated for the other two).

### Follow-ups
None to file. #1461 is already resolved by #1463; the DRY/null-forgiving triplication is already handled by #1460 (merged). Both fold into the rebase above.

Rebase, re-green, and this is good to merge.

_— Automated review (Claude Code)_

---

PR #1472, opened 2026-07-02
Remove null-forgiving operator from Battler stack-scan helpers

### LINKED ISSUE #1460: Remove null-forgiving operator from Battler stack-scan helpers

ginderjeremiah:
## Summary

Implements #1460: `Battler.StackVulnerabilityContribution` and `Battler.StackMomentumContribution` both opened with `foreach (var stack in _attributeStacks!)`, relying on their callers (`AppliedVulnerability` / `AppliedMomentum`) having already null-guarded `_attributeStacks`. `CLAUDE.md` calls out avoiding the null-forgiving operator (`!`) in favor of proper nullability handling, so both are restructured to guard internally instead.

## Changes

- **`Battler.cs`**: consolidated the two near-identical scan helpers into a single `StackContribution(EAttribute, Func<AttributeEffectStack, double>)` helper with its own `_attributeStacks is null` guard, rather than fixing each one individually. `AppliedVulnerability`/`AppliedMomentum` now each just pass a selector for their own contribution field.
- Also fixed the same pattern in `GetOrCreateStack` (`foreach (var stack in _attributeStacks!)`), which had the identical caller-guards-first-then-`!`-inside issue via `ApplyEffect`'s `_attributeStacks ??= [];`. Moved the lazy-init into `GetOrCreateStack` itself so the guard travels with the usage, and dropped the now-redundant line from `ApplyEffect`. Small enough to fold into the same change rather than filing a separate issue.

## Why consolidate instead of two local guards

A [review comment on the issue](https://github.com/ginderjeremiah/GameServer/issues/1460#issuecomment-4863336401) noted that once the Sunder overlay (#1462, still open) lands, there will be a third near-identical helper (`StackSunderContribution`) and suggested folding all three into one parameterized helper at that point. Rather than wait on an unmerged PR, I built the parameterized helper now — it already takes a selector, so wiring in a third `Func` for Sunder later is a one-line addition, no further refactor needed.

## Test plan

- [x] `dotnet build` (whole solution) — clean, 0 warnings/errors
- [x] `dotnet test` (whole solution) — all 4 backend suites green (Game.Core.Tests 1156, Game.Infrastructure.Tests 60, Game.Application.Tests 784, Game.Api.Tests 652)
- No behavior change (pure internal refactor), no new tests needed — existing `Battler`/`BattleContext` coverage (Hex/Momentum scenarios) exercises these paths unchanged.
- No frontend changes — this is backend-only domain code with no battle-parity surface.

Closes #1460

---
_Generated by [Claude Code](https://claude.ai/code/session_01DM1LSr2oETXWewkA6CSXvS)_

### DIFF
--- a/Game.Core/Battle/Battler.cs
+++ b/Game.Core/Battle/Battler.cs
@@ -203,27 +203,12 @@ namespace Game.Core.Battle
             var resistanceAttributes = DamageTypes.ResistanceAttributes(damageType);
             for (var i = 0; i < resistanceAttributes.Count; i++)
             {
-                contribution += StackVulnerabilityContribution(resistanceAttributes[i]);
+                contribution += StackContribution(resistanceAttributes[i], stack => stack.VulnerabilityContribution);
             }
 
             // Contributions are the signed resistance deltas the opponent applied (negative for a debuff); the
             // vulnerability is their reduction, so negate and clamp — an opponent that only raised resistance is 0.
             return contribution < 0 ? -contribution : 0;
-        }
-
-        // The vulnerability-tracked resistance delta the opponent has applied to one attribute, or 0 when no
-        // effect stack for it is active. A linear scan over the affected-attribute count, like GetOrCreateStack.
-        private double StackVulnerabilityContribution(EAttribute attribute)
-        {
-            foreach (var stack in _attributeStacks!)
-            {
-                if (stack.Attribute == attribute)
-                {
-                    return stack.VulnerabilityContribution;
-                }
-            }
-
-            return 0;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: Remove null-forgiving operator from `Battler` stack-scan helpers (#1460)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change removes every `_attributeStacks!` usage exactly as the issue asked, with no behavior change and a genuinely cleaner shape.

I verified the diff against the head of the branch:

- No `_attributeStacks!` remains anywhere in `Battler.cs`. Every access is now null-guarded — the two `Applied*` readers early-return, the new `StackContribution` guards internally, `GetOrCreateStack` lazy-inits, and `ProcessExpirations` already guarded.
- `GetOrCreateStack` is the *only* caller of the lazy-init, and it's the first thing `ApplyEffect` calls, so moving `_attributeStacks ??= []` out of `ApplyEffect` and into `GetOrCreateStack` is behaviorally identical — the list is always initialized before any use.
- The consolidation into `StackContribution(EAttribute, Func<AttributeEffectStack, double>)` matches the direction agreed in the issue comment for the eventual third (Sunder) helper. The selector lambdas capture nothing, so the compiler caches them as static delegates — no per-call allocation in the battle path.

### Test coverage

Adequate. This is a pure internal refactor with no behavior change, and the touched paths (`AppliedVulnerability` / `AppliedMomentum`) are exercised by existing `BattleContextTests` Hex/Momentum scenarios. No new tests warranted.

### What's good

- Solving all three sibling helpers (including the `GetOrCreateStack` case that wasn't strictly in the issue) in one parameterized helper is the right call — it collapses the duplication and makes the future Sunder addition a one-liner, without waiting on the unmerged #1462.
- The updated doc comments accurately describe the shared helper and the moved lazy-init.

The outer `_attributeStacks is null` guards in `AppliedVulnerability`/`AppliedMomentum` are now technically redundant with `StackContribution`'s internal guard, but they're a worthwhile short-circuit (they skip the attribute lookup and loop entirely), so leaving them is correct — not a finding.

I'll merge once CI goes green.

---

PR #1463, opened 2026-07-02
Floor Toughness at 0 in the mitigation curve (#1461)

ginderjeremiah:
## Summary

Fixes an unguarded pole in the shared, parity-critical Toughness mitigation curve. The curve —

```
toughnessReduction = Toughness / (Toughness + K·attackerLevel)
```

(`Battler.NetDamageAfterResistance` in `Game.Core/Battle/Battler.cs`, mirrored by `toughnessMitigatedDamage` in `UI/src/lib/battle/battle-formulas.ts`) has a pole at `Toughness = −K·attackerLevel`. `Toughness` was previously always ≥ 0 (derived `2 × Endurance`, no content debuffed it), so the pole was unreachable — but a `Toughness` debuff (e.g. the in-flight Sunder archetype, #1429) makes it directly reachable, and crossing the pole flips mitigation from a reduction into amplification or an outright net heal on a positive hit.

## How it addresses the issue

Per the issue's suggested fix, both curves now floor their `Toughness` input at `0` before computing the reduction — a debuff can strip mitigation down to `0%`, never past it:

- `Battler.NetDamageAfterResistance` (`Game.Core/Battle/Battler.cs`): `var toughness = Math.Max(_attributes[Toughness], 0);`
- `toughnessMitigatedDamage` (`UI/src/lib/battle/battle-formulas.ts`): `const effectiveToughness = Math.max(toughness, 0);`

Doc comments on both functions, `docs/backend-battle.md`, and `Enums.cs`'s `Toughness` XML doc are updated to describe the floor. The floor applies **only** to this curve's `Toughness` input — it does not clamp the raw `Toughness` attribute value itself, which is fine to be negative elsewhere (attribute breakdowns, etc.).

## Test plan

- [x] `dotnet build` — clean, 0 warnings/errors
- [x] `dotnet test` — all 4 backend suites green (Game.Core.Tests 1147, Game.Infrastructure.Tests 60, Game.Application.Tests 782, Game.Api.Tests 652)
- [x] `dotnet format --verify-no-changes` — clean
- [x] `npm run lint` (eslint + svelte-check) — 0 errors
- [x] `npm run test:unit:ci` — 275 files / 3154 tests passing
- [x] New backend unit tests in `Game.Core.Tests/Battle/BattlerTests.cs`: negative Toughness behaves like zero, Toughness exactly at the pole doesn't divide by zero, Toughness well past the pole never amplifies or heals
[...]

PR #1483, opened 2026-07-02
Revert "Floor Toughness at 0 in the mitigation curve" (#1478)

### LINKED ISSUE #1478: Revert "Floor Toughness at 0 in the mitigation curve" (#1463) — amplification past 0 is intended
## Context

#1461 (filed by me while implementing Sunder, #1429/PR #1462) flagged that the shared Toughness mitigation curve — `Toughness / (Toughness + K·attackerLevel)` — has an unguarded pole at `Toughness = −K·attackerLevel`, and suggested flooring `Toughness` at `0` as the fix. That suggestion was wrong, and #1463 (commit `320a9c1`) implemented it anyway, flooring `Toughness` at `0` in `Battler.ComputeNetDamage`/`NetDamageAfterMitigation` and the mirrored `battle-formulas.ts` `toughnessMitigatedDamage` on both simulators.

## Problem

Per direct correction: **if a debuff (e.g. Sunder) drives an enemy's `Toughness` below `0`, it makes sense for damage to be amplified** — that's the natural, intended extension of "anti-mitigation" past "no mitigation," not a bug. #1463's floor prevents this entirely, capping every debuff's effect at "0% mitigation" and silently discarding any amplification a sufficiently strong Toughness debuff should produce. This is a live behavior change to the shared, parity-critical damage formula, not a narrow fix.

The **actual** hazard #1461 correctly identified is narrower: the curve has a genuine pole exactly at `Toughness = −K·attackerLevel`, where the denominator vanishes and the reduction jumps from `−∞` to `+∞` (crossing into a regime where a positive hit can flip into a net heal). That pole is real and worth keeping in mind, but flooring at `0` was the wrong way to close it — it also closes off the intended amplification region between `0` and the pole.

## Proposed work

- Revert #1463 (`320a9c1`) on both simulators (`Game.Core/Battle/Battler.cs` and `UI/src/lib/battle/battle-formulas.ts`), restoring the unclamped curve and removing its floor-specific tests/doc language.
- Leave the actual pole (`Toughness = −K·attackerLevel`) undocumented-but-real for now unless/until it's judged worth guarding — reaching it requires an extreme single debuff (thousands of Toughness points at typical levels), well outside any currently-authored content.

## Notes

Surfaced while implementing Sunder (#1429) — its tally already defensively returns `0` credit rather than manufacturing bogus positive credit in the pole-crossing region, so the tally itself isn't at risk here; this issue is only about restoring the correct live-damage behavior for `0 < Toughness` amplification.

ginderjeremiah:
## Summary

- Cleanly reverts #1463 (`320a9c1`), which floored `Toughness` at `0` in the shared mitigation curve on both simulators.
- Restores the unclamped curve: `Toughness / (Toughness + K·attackerLevel)`, applied directly in `Battler.ComputeNetDamage` (Game.Core) and `toughnessMitigatedDamage` (`battle-formulas.ts`).
- Removes the floor-specific tests and doc language that #1463 added (`BattlerTests.cs`, `battle-formulas.test.ts`, `docs/backend-battle.md`).

## How it addresses the issue

Per the issue: if a debuff (e.g. Sunder) drives an enemy's `Toughness` below `0`, damage should **amplify** — that's the intended extension of "anti-mitigation" past "no mitigation," not a bug. #1463's floor prevented this entirely, capping every debuff's effect at 0% mitigation. This revert restores the intended amplification behavior for `0 < Toughness`.

The genuine hazard #1461 originally flagged — an unguarded pole at `Toughness = −K·attackerLevel` where the curve inverts into a net heal — is left unguarded for now, per the issue's explicit direction: reaching it requires an extreme single debuff (thousands of Toughness points at typical levels), well outside any currently-authored content.

## Test plan

- [x] `dotnet build Game.sln` — clean, 0 warnings.
- [x] `dotnet test --no-build --no-restore` — all backend suites green (Core 1185, Infrastructure 60, Api 652, Application 798).
- [x] `npm run test:unit:ci` — 3166 passed (275 files).

Closes #1478.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YWA27puKJXtyWabZQQUjba)_


ginderjeremiah (commented):
## Review: revert of #1463's Toughness floor (closes #1478)

**Verdict: Approve** — *(submitted as a `COMMENT` because I'm the PR author; GitHub won't accept a gating verdict on your own PR).* Clean, symmetric, parity-preserving revert that correctly implements the owner-directed issue. One non-blocking should-fix, filed as a follow-up.

### What it does, and whether it's right

The change restores the unclamped mitigation curve `Toughness / (Toughness + K·attackerLevel)` on both simulators (`Battler.ComputeNetDamage` / `NetDamageAfterResistance` and `toughnessMitigatedDamage`), removing the `Math.Max(Toughness, 0)` / `Math.max(toughness, 0)` floor #1463 added. That re-enables damage **amplification** for a debuff-driven negative `Toughness`, which is exactly the intended behavior per #1478 and what Sunder (#1462) depends on. I verified:

- The original #1463 (`320a9c1`) touched exactly 5 files; this PR inverts each one — no partial revert.
- No leftover floor code remains in either battle formula source (the surviving `Math.max` hits are unrelated UI slider clamps in `skills-view.svelte.ts`).
- Both simulators now compute the identical unclamped expression — battle parity holds.
- Tests and doc language for the floored behavior are removed symmetrically on both sides.
- Leaving the genuine pole at `Toughness = −K·attackerLevel` (and the past-pole heal region) unguarded is a deliberate, documented decision in #1478 — reaching it needs thousands of Toughness points of debuff at typical levels, outside authored content. Correct to defer.

### Should-fix (non-blocking)

- **Missing regression test for the restored amplification.** The revert removes the tests that pinned the old floored behavior but adds nothing to pin the *intended* one, so there's now no test asserting a negative `Toughness` amplifies a hit through the curve — the whole point of this change, and behavior Sunder relies on. Since a floor was just removed here, this region is a prime candidate for a future edit to silently re-clamp. Recommend a mirrored test on both simulators (e.g. `toughnessMitigatedDamage(40, -10, 1)` → `80` and the equivalent `Battler` case), per the battle-parity rule. Filed as **#1484** so it doesn't gate this revert.

### Test coverage

Removals are symmetric across `BattlerTests.cs` and `battle-formulas.test.ts`, keeping the suites consistent. The gap is the absence of a *new* amplification test noted above (→ #1484); existing amplification coverage is attacker-side (`AmplifyDamage`) and negative-*resistance*, neither of which exercises the Toughness curve on negative input.

### What's good

Proper `git revert` with a commit message that captures *why* the floor was the wrong fix, a thorough PR description, and the doc (`backend-battle.md`) kept in lockstep with the code. Both simulators moved together — no parity drift.

_Not merging: CI is still mid-run (e2e / test-backend / test-frontend in progress) and the should-fix above is better tracked as #1484 than folded into this revert._

---

ginderjeremiah:
Thanks for the review — folded the should-fix in (`fab09d7`) since it was small and directly strengthens this revert: added the mirrored `toughnessMitigatedDamage(40, -10, 1)` / `Battler.TakeDamage` amplification test on both simulators, matching the region #1484 called out. Closes #1484 too.

---

Issue #1481, opened 2026-07-02
Unify overlay proficiency tallies on post-mitigation net × φ(investment)


ginderjeremiah:
Settled with the project owner (2026-07-02 design discussion), following on the Sunder implementation (#1429 / #1462). This **supersedes spike #1398's follow-on resolutions 2–3** (fixed-baseline add-one-in, enemy-independence via absolute marginal) — the spike doc gets an addendum as part of this issue.

## Problem

The overlay tallies that remained on the literal-marginal / net-baseline shape all read the target's **live** mitigation, so the debuff archetypes landing (Hex #1427, Sunder #1429) silently contaminated them first-order:

- **Crit** (`BattleContext.DamageTarget`): `baselineNet × φ(m−1)` — `baselineNet` runs `ComputeNetDamage` at live (debuffed) mitigation. A 0.4 vulnerability at 30% enemy resistance inflates the baseline by (1−0.3+0.4)/(1−0.3) ≈ **1.57×** — "your crit training balloons because you also Hex," the exact anti-goal of resolution 2's add-one-in rule. **Momentum** and **Cull** compute their marginals through the same live curve and inherit both debuffs' inflation.
- **Hex** (`Battler.HexBonusForHit`): the counterfactual marginal reduces to `dealt × (1−T) × v/(1+v)` where `(1−T)` is the **live** Toughness-curve factor — so it is not flat in enemy Toughness (contradicting resolution 3's enemy-independence goal and game-design.md's own "flat" claim), and a Sunder debuff on the same target inflates it.

These violations were latent when the crit reference (#1448) landed — nothing moved mitigation mid-battle back then.

## Decision: HP-bounded share claims

Every overlay books, on hits where its enabler is active:

**`hitNet × φ(investment)`** — the hit's **post-mitigation net** (the value the health math already computed), scaled by the saturating `φ(a) = a/(1+a)` on the player's own investment.

| Overlay | Tally | Investment |
| --- | --- | --- |
| Crit | `hitNet × φ(m−1)` | crit-damage multiplier over 1 |
| Hex | `hitNet × φ(v)` (DoT ticks: post-resistance tick × φ(v)) | applied vulnerability (`AppliedVulnerability`) |
| Momentum | `hitNet × φ(r)` | applied ramp (`AppliedMomentum`) |
| Sunder | `hitNet × φ(s/(K·L))` | applied Toughness debuff, K·attackerLevel-normalized as today |
[...]

PR #1485, opened 2026-07-02
Cap proficiency-booked damage at health actually removed (#1482)

### LINKED ISSUE #1482: Cap proficiency-booked damage at health actually removed (overkill inflates offense-book XP)
Found while reviewing the proficiency tally bases (2026-07-02 design discussion; companion to the overlay-tally unification #1481).

## Problem

`Battler.TakeDamage` (Game.Core/Battle/Battler.cs) subtracts the full net from `CurrentHealth` with no floor and returns it uncapped; `BattleContext.DamageTarget` books that full net into `BattleStats.TypedDamageDealt` — so the offense book counts **overkill**.

Effect: one-shotting a trivial enemy books the player's entire hit, giving every kill a power-proportional activity floor. That cuts against the documented anti-grind invariant — "power-normalization is the difficulty curve … a trivial enemy lets you do little relative to your power" (game-design.md §Proficiency XP) — and makes gray-enemy speed-farming a type-XP route bounded only by kill cadence rather than by the enemy's authored HP. Once the overlay tallies also move onto the post-mitigation net basis (companion issue), the same uncapped tail feeds every book, so the fix pays off everywhere at once.

## Fix

Cap the **booked** (proficiency-facing) amount at the health actually removed. Do **not** change the health math itself — `CurrentHealth` going negative is how death resolves, and the damage calc is parity-critical; this is a stats-recording change only.

Design decisions to make in-implementation:

- The whole-hit statistics (`HighestPlayerAttack`, `CriticalDamageDealt`, per-skill totals) arguably should keep the full net — "biggest hit" showing overkill is good feedback. Probably cap only what feeds `TypedDamageDealt` (and, post-companion-issue, the overlay tallies).
- Note `TypedDamageDealt` also drives the Statistics screen's per-damage-type breakdown (#1473/#1480) — capping makes the displayed numbers reflect damage that actually landed on health, which reads as more accurate, but confirm nothing else assumes the uncapped sum.
- Check the DoT path (`ApplyDamageOverTime`): whether a killing tick books its full accumulated tick or should be capped the same way.
- Multi-portion hits: portions are applied sequentially, so the cap must distribute across portions in application order (later portions of a killing swing book 0 once health is exhausted).

## Tests

Backend: killing hit books only remaining health per type; multi-portion killing swing distributes the cap in portion order; non-lethal hits unchanged; DoT killing tick per the decision above; `HighestPlayerAttack` retains full net if that's the call.

ginderjeremiah:
## Summary

Caps the proficiency offense book's bookings at the health a hit actually removed (#1482). Overkill — the tail of a killing blow past the target's remaining health — no longer counts as training activity, so the offense book's per-battle total is bounded by the enemy's health pool and one-shotting trivial enemies no longer mints a power-proportional activity floor per kill (the anti-grind property `activity ÷ power` was designed around).

This is a **stats-recording change only**: the health math, death resolution, reflection, and the whole parity surface are untouched (`battle-step.ts` unchanged — the typed books are backend-only side channels).

## How it addresses the issue

- **Direct hits** (`BattleContext.DamageTarget`): each portion books `min(net, healthBeforePortion)` via the new `Battler.HealthRemoved` rule, so the cap distributes across a killing swing's portions in the fixed portion order — later portions of a killing swing book 0.
- **DoT ticks** (`Battler.ApplyDamageOverTime`): the damage-dealt recorder books capped the same way across the fixed accumulator order, tracking the running remaining health (a negative tick — authored absorption — restores it, matching the uncapped DoT-heal rule).
- **Whole-hit statistics keep the full net** (the issue's open call, resolved as proposed): `PlayerDamageDealt`, `HighestPlayerAttack`, `CriticalDamageDealt`, and per-skill totals deliberately include overkill — "biggest hit" is real feedback; it just isn't training activity. Absorption's negative bookings pass through unchanged.
- **`KillsByDamageType`** (majority-type attribution) now reads the capped book — if anything more accurate, since the majority reflects health actually removed.
- Docs updated where the offense book is described: `game-design.md` §Proficiency XP, `backend-battle.md` (typed-books bullet), plus the XML docs on `BattleStats`/`BattleContext`/`Battler` and the `ProficiencyRewardService` avenue comment.

## Test plan

- New: killing direct hit books only remaining health while health math and whole-hit stats stay uncapped; multi-portion killing swing caps in portion order (second portion books 0); absorbed hit's negative net books unchanged; DoT killing tick books only remaining health; multi-type DoT killing tick caps in accumulator order; DoT absorption restores bookable health for later types in the fixed order.
- Adjusted: three `MultiTypedDamageProficiencyTests` scenarios and two `BattleContextTests` scenarios incidentally overkilled their 50-HP target — their intent is split/accumulation ratios, so their targets got bigger health pools rather than weakened assertions.
- Full backend suite green locally: **2704 passed, 0 failed** (Core 1194, Application 798, Api 652, Infrastructure 60).

## Notes

- Companion to #1481 (overlay tallies moving to `net × φ(investment)`): this cap is what makes that issue's "per-battle basis is bounded by enemy HP" property exact. #1481 will build on this branch.
- The Retribution signal (`PlayerReflectedDamageDealt`) is deliberately not capped here: reflection scales with the *enemy's* attack, so the power-proportional-floor exploit shape doesn't apply, and its basis is #1481's concern.

Closes #1482


ginderjeremiah (commented):
## Review: Cap proficiency-booked damage at health actually removed (closes #1482)

**Verdict:** Approve _(posted as a comment — reviewer identity matches the PR author, so GitHub won't accept a gating event; the intended verdict is approve)_ — the change does exactly what the issue asks, cleanly scoped to the stats-recording layer with the health/parity surface untouched, and the tests cover every design decision the issue called out.

I verified the logic in full:

- `Battler.HealthRemoved(damage, healthBefore) = Min(damage, Max(0, healthBefore))` caps positive overkill at remaining health and passes a negative (absorption heal) through unchanged — matching the "booking-only, never cap the health math" requirement.
- **Direct-hit path** captures `healthBefore` per portion inside the loop; since `TakeDamage` mutates `CurrentHealth` each portion, the re-read distributes the cap across a killing swing in the fixed portion order (later portions book 0). Correct.
- **DoT path** can't re-read `CurrentHealth` (it's subtracted once after the loop via `CurrentHealth -= dot`), so it tracks a local `bookableHealth`, decrementing by the **uncapped** `tickDamage` — correctly mirroring the uncapped health math, including a negative (heal) tick restoring bookable health for later types in the accumulator order. The two paths use different mechanisms because their health-mutation timing differs, and the comments explain why — good.
- Whole-hit aggregates (`PlayerDamageDealt`, `HighestPlayerAttack`, `CriticalDamageDealt`, per-skill totals) deliberately keep the full net, resolving the issue's open call as proposed.
- `battle-step.ts` correctly left untouched: the typed offense book is a backend-only side channel with no parity surface, so the battle-parity mirror rule doesn't apply here.

### Test coverage

Thorough and matches the issue's checklist: killing direct hit books only remaining health while whole-hit stats stay full; multi-portion killing swing caps in portion order (second portion books 0); absorbed hit's negative net passes through; DoT killing tick caps; multi-type DoT caps in accumulator order; DoT absorption restores bookable health for later types. The adjusted `MultiTypedDamageProficiency`/`BattleContext` scenarios were given larger HP pools rather than weakened assertions — the right call, since their intent is split/accumulation ratios, not the cap.

### What's good

- Correctly recognized that Retribution (`PlayerReflectedDamageDealt`) and the overlay tallies (Hex/Sunder/Momentum/Cull) are out of scope here and belong to #1481 — no over-reach.
- Docs updated in lockstep (`game-design.md` §Proficiency XP, `backend-battle.md`, and the XML docs on `BattleStats`/`BattleContext`/`Battler`), keeping the offense-book description accurate.
- Clean adherence to house rules: the `recordDamageDealt?.Invoke(...)` single-statement was properly expanded to a braced `if` block once it needed two statements; no null-forgiving operators introduced.

No blocking findings, no should-fixes, no nits. Once backend CI comes back green this is good to merge.


---

Issue #1421, opened 2026-06-30
Add authoring-only DesignerNotes to the 9 static reference entities


ginderjeremiah:
Part of the content-authoring-pipeline spike #1390 — decision 6. See [`docs/spikes/1390-content-authoring-pipeline.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/1390-content-authoring-pipeline.md).

Per-piece design intent (why a piece exists) has nowhere to live; the Workbench holds values, not rationale. Add an authoring-only notes field next to the piece.

## Scope
- Add `string DesignerNotes` to the 9 static reference entities (`Skill`, `Item`, `ItemMod`, `Enemy`, `Zone`, `Challenge`, `Class`, `Path`/`Proficiency`, `SkillRecipe`), following the established **authoring-metadata pattern** (same shape as `IconPath` / word-of-power):
  - lives on the **entity + contract only** — **never** on the lean `Game.Core` battle model, **never** sent to the game client.
  - surfaced/edited in the **Workbench**.
  - **is** included in the content export (#1418) — per-piece rationale is worth version-controlling.
  - backfills to `""`; bumps each set's version hash once (clients re-download the set once).
- Migration + entity→contract mapper updates + Workbench editor field for each set.

## Notes
- Macro arc lives in `docs/content-design.md` (already exists); this field is the home for **per-piece** rationale. Settled division of labor — see spike decision 7.
- Alternative homes (sidecar notes file, the content-design doc) rejected: rationale belongs next to the piece, in the authoring surface.

## Tests
- Mapper test: `DesignerNotes` round-trips entity↔contract and is absent from the lean Core model / client projection.
- Admin save persists the field; existing rows backfill to `""`.

PR #1436, opened 2026-07-01
Add authoring-only DesignerNotes to the static reference entities (#1421)

ginderjeremiah:
## Summary

Adds a free-text **`DesignerNotes`** field to every static reference entity — `Skill`, `Item`, `ItemMod`, `Enemy`, `Zone`, `Challenge`, `Class`, `Path`, `Proficiency`, `SkillRecipe` — so per-piece design intent (*why* a piece exists) has a home next to the piece, not just in the macro `docs/content-design.md`. Closes #1421 (spike #1390, decision 6).

It follows the established **authoring-metadata pattern** (same shape as `IconPath` / the word of power):
- lives on the **entity + read contract only**, never on the lean `Game.Core` battle models (the entity→core mappers ignore it);
- is included in the **content export** (per-piece rationale is worth version-controlling);
- backfills to `""` and bumps each set's version hash **once** (clients re-download the set once).

## How it addresses the issue

- **Entities + contracts** (10 each): `required string DesignerNotes`, mapped in every `ToContract`, persisted in every admin `SaveXxx` add/edit path, and configured in `GameContext` (`varchar(2000)`).
- **EF migration** `AddDesignerNotes`: adds the column to all 10 tables as `character varying(2000) NOT NULL DEFAULT ''`.
- **Content export**: the exporter reads the contracts, so the field flows in automatically — `content/*.json` regenerated (drift guard green).
- **TS client**: regenerated (`designerNotes` on all 10 interfaces).
- **Workbench**: a "Designer Notes" textarea added to all 8 config-driven editors and the bespoke Path/Proficiency progression editor (identity DTOs + new-record factories threaded through).
- **Docs**: documented under the reference-data authoring-metadata section in `docs/backend.md`.

## Note on the Challenge exception

Unlike the other 9 sets (entity → contract directly), the challenge read contract is projected from the **domain `Game.Core.Progress.Challenge`** model (`ChallengesCacheHolder` → `ChallengeContractMapper`). So the field is carried on that domain model too. It's still never a battle input — `Challenge` already carries display strings (`Name`/`Description`) and is not part of the simulation.

## Interpretation flagged for review

[...]

PR #1488, opened 2026-07-02
Redact DesignerNotes from non-admin reference-data socket responses

### LINKED ISSUE #1439: Admin-only contract projection for authoring-only reference metadata
## Context

Follow-up from PR #1436 (adds `DesignerNotes` authoring metadata to the static reference entities, closes #1421; spike #1390).

Issue #1421 states authoring-only metadata should be "never sent to the game client", but the accepted acceptance criterion also says the change "bumps each set's version hash once (clients re-download the set once)". Those are in tension because this repo has a **single read contract** shared by the game client and the admin Workbench, and the version hash is computed over exactly what the client receives. PR #1436 resolved this the way the `IconPath` / word-of-power precedent points — the notes ride the shared contract (so they bump the hash and satisfy the concrete criterion) and the client simply never renders them.

## The follow-up

Revisit whether authoring-only fields (`DesignerNotes`, and any future ones) should be excluded from the **client-facing** reference payload — e.g. via a dedicated admin-only projection or a serializer filter — while still being version-controlled in the content export and editable in the Workbench.

- Pro: reconciles the "never sent to the client" intent literally, and trims wire payload (up to 2000 chars per record shipped to every client purely to be discarded).
- Con: introduces a second projection / diverges from the current single-contract simplicity; would change how the version hash is computed for these fields.

Non-urgent — the current shared-contract approach is acceptable and matches existing precedent.

Labels: tech debt, claude, scope: medium

ginderjeremiah:
## Summary

Follow-up from #1436 / #1421: `DesignerNotes` (and every other authoring-only field) still rides the shared reference-data read contract, but the socket layer now redacts it before it reaches a non-admin connection, so it's never actually sent to a playing client — literally satisfying #1421's "never sent to the game client" intent without splitting into a second projection type.

- `SocketContext` now carries `IsAdmin`, read once at the WebSocket handshake from the connection's JWT role claim (the same source `AdminRoleAuthorizationFilter` reads for HTTP admin endpoints).
- `AbstractReferenceDataCommand<TModel>.HandleExecuteAsync` blanks `DesignerNotes` to `""` for a non-admin connection. The Workbench connects with the Admin role and still receives (and can edit) the real notes over this same `Get*` command — it has no separate read path today, so this had to stay role-aware rather than a blanket strip.
- `ComputeVersion()` is unchanged: it always hashes the genuine, unredacted data from a separate `GetReferenceData()` call, so authoring a note still bumps the set's version hash once for every client — the accepted #1436 criterion holds exactly as before. Only the wire payload differs by role.
- Added `IHasDesignerNotes`, a marker interface implemented by the 10 contracts that carry the field (`Skill`, `Item`, `ItemMod`, `Enemy`, `Zone`, `Class`, `Path`, `Proficiency`, `SkillRecipe`, `Challenge`), so the redaction is generic in the base command rather than a per-set override.
- Updated `docs/backend.md`'s "Designer notes" section to describe the new split.

I raised the design fork from #1439 with @ginderjeremiah first (a plain "strip the field" would have silently blanked the Workbench's Designer Notes editor and the version-hash change would have reversed the accepted #1436 criterion) — the agreed direction was to redact from the wire payload only, leaving the hash untouched.

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test` (full solution) — all green, including new coverage:
  - `Game.Api.Tests/Unit/ReferenceDataCommandAdminRedactionTests.cs` — admin vs. non-admin redaction, a model without `IHasDesignerNotes` passes through untouched, and `ComputeVersion()` is unaffected by redaction regardless of call order.
  - `Game.Api.Tests/Integration/ReferenceDataSocketTests.cs` — a real `GetSkills` socket round-trip redacts notes for a non-admin connection, preserves them for an admin connection, and the `GetReferenceDataVersions` hash for `GetSkills` matches across both.

Closes #1439

---
_Generated by [Claude Code](https://claude.ai/code/session_01G8vm2LKmCR8n7LPf2EQ9t2)_


ginderjeremiah (commented):
## Review: Redact `DesignerNotes` from non-admin reference-data socket responses (#1439)

**Verdict: Approve** — solves the follow-up cleanly, and the one load-bearing safety claim holds up under verification. _(Posted as a comment because GitHub won't accept a gating verdict on one's own PR; treat this as an approval.)_

### Verification of the key risk
The whole design rests on one claim: mutating `DesignerNotes = ""` during redaction can't corrupt the in-memory reference cache or bleed across connections. I checked it against every affected store rather than trusting the comment, and it genuinely holds:

- All 9 DB-backed sets return freshly-mapped contract instances per call (`[.. Snapshot.Entities.Select(...ToContract)]` in `Skills`, `Items`, `ItemMods`, `Enemies`, `Zones`, `Classes`, `Proficiencies`, `Paths`, `SkillRecipes`). The cache holds *entities*, not contracts, so the mutated objects are throwaway projections.
- `Challenge` is the one that returns a shared domain instance from `_challenges.All()`, but `GetChallenges` re-maps it through `ChallengeContractMapper.ToContract` — the `IHasDesignerNotes` marker is on the freshly-mapped *contract*, so it's safe too.
- `ComputeVersion` calls `GetReferenceData()` separately and is memoized against `VersionKey`, so the hash always reflects the genuine unredacted data regardless of call order — pinned by `ComputeVersion_AfterNonAdminRedaction_StillHashesTheUnredactedData` and the memoized-before-redaction test.
- `ContentExporter`/`ContentHealthService` hit the stores directly (not the socket command), so the export still carries real notes.

### Security
`SocketInterceptorMiddleware` reads `context.User.IsInRole(nameof(ERole.Admin))` — the exact check `AdminRoleAuthorizationFilter` uses for HTTP admin endpoints, off the cryptographically-validated JWT `role` claim, not the volatile session cache. Correct and consistent. `IsAdmin` being fixed for the connection lifetime is documented and matches every other handshake claim.

### Test coverage
Solid and at the right altitude:
- Unit (`ReferenceDataCommandAdminRedactionTests`) exercises the generic mechanism directly — admin vs non-admin redaction, a model *without* `IHasDesignerNotes` passing through untouched, and version-hash independence in both orderings. Testing the marker mechanism generically is the right call over repeating all 10 concrete sets.
- Integration (`ReferenceDataSocketTests`) covers a real `GetSkills` round-trip both ways plus `GetReferenceDataVersions` parity across an admin and non-admin connection.

### What's good
- The design fork against a *documented* pattern was raised with the owner before any code was written, and the accepted #1436 criterion (hash bumps once per client) is preserved exactly.
- `IHasDesignerNotes` keeps the redaction generic in the base command instead of a per-set override.
- `docs/backend.md` was updated to describe the new wire-vs-hash split.

No blocking or should-fix items. I'll merge once CI is green.

---

PR #1489, opened 2026-07-02
Replace Toughness curve's level normalization with a tuned constant (#1487)

### LINKED ISSUE #1487: Reconsider the attackerLevel factor in the Toughness mitigation denominator (level-normalization → constant)
## Summary

Reconsider the `K·attackerLevel` factor in the **denominator** of the Toughness mitigation curve and evaluate replacing it with a tuned **constant**. The level-normalization was introduced in spike #1330 to keep the mitigation band stable as content scales, but under a realistic progression lens it doesn't buy what it appears to, and it imposes a "scale Toughness proportionally or watch it decay" tax that a constant would not.

## Current behavior

The direct-hit mitigation curve reduces a hit by `Toughness / (Toughness + K·attackerLevel)`, with `K = ToughnessMitigationConstant = 20`:

- Backend: `Battler.NetDamageAfterResistance` — `Game.Core/Battle/Battler.cs:146`
- Frontend (parity mirror): `toughnessMitigatedDamage` / `mitigateDamage` — `UI/src/lib/battle/battle-formulas.ts:58`
- Constant: `Game.Core/GameConstants.cs:41`, mirrored to the client at `UI/src/lib/api/types/game-constants.ts:10` via the constants codegen.
- Toughness source: `Toughness = 2·Endurance` — `Game.Core/Attributes/Modifiers/StaticAttributeModifiers.cs:37`.

Design intent lives in `docs/game-design.md` ("Mitigation: the Toughness curve") and spike `docs/spikes/1330-mitigation-rework.md`.

## The problem (realistic-progression lens)

Assume the realistic case where **player level ≈ enemy level** (any fight that isn't trivially easy) and enemy damage, HP, and Toughness are all authored to scale with level via `BaseAmount + AmountPerLevel × level`.

Because the denominator is `Toughness + K·L`, **holding your mitigation % requires Toughness to grow proportionally to L**. A player who does *not* index heavily into Toughness carries Toughness that grows slower than `L`, so `Toughness / (K·L) → 0` and their mitigation erodes toward zero as they progress. The same content scaling that raises enemy damage (`D ∝ L`) also inflates the mitigation denominator, so a static/modest defensive investment is squeezed from both sides and **decays** — a build that had, say, 20% mitigation early quietly drops toward ~0%, sliding alive-time back to the no-defense baseline even though no Toughness was ever lost.

This is **not** runaway immunity. Removing the level factor does not make you invulnerable: with a constant denominator, mitigation % → 100% at high level but post-mitigation damage flattens to a *constant* (`≈ dC/t`) while HP keeps growing, so survivability grows ~linearly with level rather than exploding. Enemy damage's own level-scaling cancels the "explosion."

Two clarifications from the discussion that shaped this:
- Normalizing by **defender** level instead of attacker level fixes nothing — with player level ≈ enemy level the two are numerically equivalent; the only difference (level gaps) occurs in trivially-easy fights that don't matter.
- The level term's headline benefit — an auto-stable mitigation band — **only materializes if Toughness scales ∝ level**, which is exactly what a non-proportional (modest) investor doesn't do. For them it is a decay tax, not a stability guarantee.

## Proposed change

Replace `K·attackerLevel` with a tuned constant `C`, so mitigation becomes `Toughness / (Toughness + C)`:

- A fixed Toughness investment **retains its mitigation %** across all of progression instead of decaying.
- Per-item mitigation value becomes **legible** (`Toughness = C → 50%`), rather than depending on enemy level and the flat-vs-per-level split of your Toughness sources.

## Decision needed before implementing

This is a design-values call, not a correctness fix. The constant changes Toughness from a **flat band** stat (current: same % at every tier *for a proportional investor*, decaying for everyone else) into a **retained / compounding** stat (constant: investment holds its value; with proportional scaling, defense is weak early and strong late). Trade-offs to sign off on:

- **Weak early game**: a constant makes early-tier mitigation low unless co-tuned. Need to re-derive the Endurance→Toughness coefficient / `C` so early defense isn't paper.
- **High-tier direct-hit relevance**: at extreme tiers direct hits become small relative to HP. DoT bypasses Toughness entirely (resistance is a DoT's only mitigation), so it remains the pressure valve — but confirm this end-state is acceptable.

## Implementation notes

- **Parity-critical.** The expression must stay bit-for-bit identical between `Battler.cs` and `battle-formulas.ts`; re-baseline the mitigation parity vectors (`BattleSimulatorParityTests` ↔ `battle-simulation-parity.test.ts`) and the `TakeDamage_AppliesToughnessMitigationCurve` unit tests on both sides.
- **Sunder ripple.** `Battler.SunderBonusForHit` (`Game.Core/Battle/Battler.cs:265`, spike #1429) deliberately reuses `K·attackerLevel` as the curve's characteristic magnitude ("the same constant the live mitigation curve itself divides by"). If the curve drops the level factor, Sunder's normalization must be reconciled to stay coherent.
- **Constant codegen.** Update `GameConstants` (rename if it's no longer "per attacker level"); the client constant regenerates from it.
- **NaN edge case.** The current code documents an unreachable `0/0` when Toughness and attackerLevel are both 0; a constant denominator removes that entirely — clean up the related comments.
- **Docs.** Update the "Mitigation: the Toughness curve" section of `docs/game-design.md` and the relevant `backend-battle.md` notes; add a short addendum to spike #1330 recording the revision and rationale.

## Related

- Spike #1330 — introduced the `Toughness / (Toughness + K·attackerLevel)` curve.
- #1429 — Sunder delivery archetype, which keys off `K·attackerLevel`.
- #1461 — the Toughness=0 floor behavior on this curve.

ginderjeremiah:
## Summary

Replaces the `K·attackerLevel` factor in the Toughness mitigation denominator with a tuned constant, per #1487. The direct-hit curve is now `Toughness / (Toughness + C)` with `C = GameConstants.ToughnessMitigationConstant = 200` — the Toughness at which mitigation is exactly 50%, making per-item mitigation value legible.

**Design values signed off before implementing** (per the issue's "Decision needed" section):
- `C = 200` — early-game durability is carried by MaxHealth (which Endurance also provides), so the half-point sits at a mid-game Toughness rather than being co-tuned low.
- The `Toughness = 2·Endurance` derivation is unchanged (one knob at a time; both are strawmen for Workbench-era balancing).
- The high-tier end-state is accepted: post-mitigation direct-hit damage flattens toward a constant while HP grows (survivability vs direct hits grows ~linearly — no runaway immunity), and DoT (which bypasses Toughness) remains the late-game pressure valve.

## Changes

- **Curve** (parity-critical, bit-for-bit identical on both simulators): `Battler.ComputeNetDamage` ↔ `toughnessMitigatedDamage`. The client constant regenerates from `GameConstants` via codegen.
- **Sunder reconciliation** (#1429): `SunderBonusForHit` normalizes its investment by the same constant the live curve divides by — `φ(s / C)` instead of `φ(s / K·attackerLevel)`.
- **Dead parameter removal**: the now-unused `attackerLevel` threading is removed end-to-end (`TakeDamage`/`ComputeNetDamage`/`SunderBonusForHit` ↔ `takeDamage`/`mitigateDamage`/`toughnessMitigatedDamage`, plus the skills-view/tooltip display path). The documented unreachable `0/0` NaN edge case is gone with it; the unguarded pole (#1478/#1483) moves from `−K·attackerLevel` to `−C` and the comments/tests now pin that.
- **Docs**: `game-design.md` "Mitigation: the Toughness curve", `backend-battle.md` pipeline + Sunder notes, and a revision addendum in `docs/spikes/1330-mitigation-rework.md`.

## Test plan

- Re-baselined the Toughness unit tests with mirrored scenarios and expected results on both sides (`BattlerTests` ↔ `battler.test.ts` / `battle-formulas.test.ts`), anchored on the new half-point (END 100 → Toughness 200 → exactly 50%).
- Re-derived all Toughness-involving parity vectors by hand and confirmed by simulation (`BattleSimulatorParityTests` ↔ `battle-simulation-parity.test.ts`), preserving each scenario's intent — `highToughnessTrickle` still trickles (END scaled to 1000 Toughness), `critPunchesThroughTyped` keeps its clean halving (authored Toughness 200), the offline coin-flip boss keeps its exact 0.8 reduction (injected Toughness 80 → 800), and the Sunder tests keep their round φ values (debuffs scaled ×10).
- [x] Backend: `dotnet build` clean, `dotnet test` — 2,704 passed / 0 failed (Core, Infrastructure, Api, Application)
- [x] Frontend: `npm run lint` clean (eslint + svelte-check 0 errors), `npm run test:unit:ci` — 3,166 passed / 0 failed

Closes #1487.

https://claude.ai/code/session_01QA8njnX5bcPgCKCWtdzBGd

---
_Generated by [Claude Code](https://claude.ai/code/session_01QA8njnX5bcPgCKCWtdzBGd)_


UI/src/tests/routes/game/screens/skills/skills-view.test.ts
```diff
@@ -231,7 +231,7 @@ describe('pure helpers', () => {
 		expect([a, b].sort(sortMetrics('cd', 0)).map((m) => m.skill.id)).toEqual([1, 0]);
 		// dps with no toughness: a=100/2=50, b=30/1=30 → a first.
 		expect([b, a].sort(sortMetrics('dps', 0)).map((m) => m.skill.id)).toEqual([0, 1]);
-		// dps with toughness 80 (factor 20/100=0.2): a=100·0.2/2=10, b=30·0.2/1=6 → a first; damage a=20, b=6.
+		// dps with toughness 80 (factor 200/280≈0.714): a=100·0.714/2≈35.7, b=30·0.714/1≈21.4 → a first.
```
ginderjeremiah: Nit (pre-existing, carried forward): the assertion is a `'dmg'` sort, but this comment labels it "dps" and divides by cooldown (`/2`, `/1`) — that's the dps formula, not dmg. For the `'dmg'` sort the mitigated values are `a=100·0.714≈71.4` and `b=30·0.714≈21.4` (no cooldown division), and the old comment used to note the correct `damage a=20, b=6` tail which got dropped here. The ordering conclusion is unaffected, so this is just a comment-accuracy nit — since you were already retuning the numbers on this line, worth fixing the label/math to match the `'dmg'` sort while you're here.

---

ginderjeremiah (commented):
## Review: replace the Toughness curve's `K·attackerLevel` denominator with a tuned constant `C` (#1487)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change is correct, parity is preserved bit-for-bit, tests are mirrored on both sides, and the design decision was signed off in #1487. One trivial comment-accuracy nit, non-blocking.

### Blocking
- None.

### Should-fix
- None.

### Nits
- `UI/src/tests/routes/game/screens/skills/skills-view.test.ts:234` — the comment labels a `'dmg'` sort as "dps" and divides by cooldown; the ordering is unaffected but the label/math don't match the sort (inline comment). Pre-existing and slightly worsened here (the correct `damage a=20, b=6` tail was dropped).

### Test coverage
Strong. The mitigation math (`Battler.ComputeNetDamage` ↔ `toughnessMitigatedDamage`) and the removed `attackerLevel` threading are covered end-to-end:
- New half-point anchor test on **both** sides (`TakeDamage_ToughnessAtTheConstant_MitigatesExactlyHalf` ↔ `mitigates exactly half at the constant`), pinning `Toughness = C → 50%`.
- The obsolete `AttackerLevelScalesTheCurve` test is correctly removed on both sides rather than left asserting dead behavior.
- All Toughness-involving parity vectors were re-derived and the intent of each scenario preserved (`highToughnessTrickle` still trickles via scaled Endurance, `critPunchesThroughTyped`/`dotBypassesMitigation` re-anchored on the new half-point, the Sunder normalization moved to `φ(s / C)`). I spot-checked several by hand (e.g. `cooldownRecovery` 85·200/230 = 73.913/hit → 6 hits at ticks 28..168 → 6720ms; `multiSkill` first reaches 450 at tick 160 → 6400ms) and they line up, and the frontend `battle-simulation-parity.test.ts` expectations match the backend `BattleSimulatorParityTests` exactly.

### What's good
- **Clean dead-parameter removal.** `attackerLevel` is threaded out end-to-end (`TakeDamage`/`ComputeNetDamage`/`SunderBonusForHit` ↔ `takeDamage`/`mitigateDamage`/`toughnessMitigatedDamage`, plus the skills-view/tooltip display path) with no stragglers left behind — the only remaining mention is the intentional "replaced the former `K·attackerLevel`" note in `GameConstants`.
- **Parity discipline.** The live expression is a single constant divide on both sides, and the unreachable `0/0` NaN edge case documented under the old formula is correctly retired while the `−C` pole (#1478/#1483) is re-pinned in comments and tests.
- **Sunder reconciliation (#1429)** correctly re-normalizes by the same constant the live curve divides by, keeping the "dimensionless by the curve's own characteristic magnitude" invariant intact.
- **Docs.** `game-design.md`, `backend-battle.md`, the proficiency-routing Sunder note, and a well-written spike addendum in `1330-mitigation-rework.md` all updated consistently — old level-normalization language is gone, not just appended to.

CI: all critical checks green (test-backend, test-frontend, verify-backend, verify-frontend, e2e chromium/firefox); `test-e2e (webkit)` still running at review time.

---

Issue #1534, opened 2026-07-03
Surface combat power ratings in the UI (player + enemy)


ginderjeremiah:
Frontend follow-up from spike #1526 (design in `docs/spikes/1526-combat-rating-power-measure.md`, Decision 7). Blocked by #1531 and #1532 (the displayed number should mean what the reward system uses).

## What

Surface the combat power rating in the UI for **both** the player and the enemy. Display-only: the server computes ratings and sends values; the client never recomputes them (**no parity surface**).

- **Player rating** — on the attributes screen (and wherever the character summary fits). Key synergy with #1528's inert-stat signaling: allocating a point into an enabler-less amplifier visibly does not move the power number — the dead-stat lesson taught by the UI for free.
- **Enemy rating** — on the battle screen alongside the player's, with a simple matched/trivial cue derived from the ratio, making the anti-grind curve legible ("you've outgrown this zone") instead of mysterious. The enemy's rating is per-encounter (level roll + fielded skill draw), so it rides the battle-start payload.
- Delivery: enemy rating on the battle-start payload; player rating with player state refreshes (it changes on allocation/gear/level changes — pick the natural existing payloads rather than a dedicated endpoint if possible).

## Notes

- Rating values are unitless server numbers; presentation (rounding, iconography, naming — "Power") is this issue's call.
- A zone-preview power range (min/max over the spawn table) is a natural extension; in scope only if cheap.

Issue #1533, opened 2026-07-03
Calibration report: rating vs SumCoreAttributes across authored content; pick XP scale and pie


ginderjeremiah:
Supporting tooling for spike #1526 (design in `docs/spikes/1526-combat-rating-power-measure.md`). Blocked by #1531; its output picks the constants the consumer-swap issue needs.

## What

A one-off calibration report (console harness or report-style test in the vein of the content lint direction, spike #1390) that answers, before the reward swap lands:

1. **Per authored enemy**: `SumCoreAttributes` vs combat rating, side by side — flags enemies whose pricing shifts drastically (likely kit-vs-distribution mismatches the rating is now seeing correctly; cross-reference the #1529 lint).
2. **Per zone, per reference build**: where a normally-progressing player's rating places each zone's spawn table on the `r = enemyRating/playerRating` axis — verifying every zone's intended enemies sit near matched (not trivial, not unreachable) for the level band content-design.md targets.
3. **XP/hour and proficiency-XP/hour curves** across the zone arc for the reference builds under the new curve (`k × E × min(r,1)²`, `activity ÷ max(P,E)`) — the direct check that optimization and zone advancement both pay per-hour, and the input for choosing:
   - `k` (XP scale — keep payouts roughly continuous with today for a mid-arc reference point), and
   - the `ProficiencyXpPerVictory` retune (the rating's numeric scale differs from an attribute-point sum).

Reference builds: a small authored set (e.g. focused damage, tank/reflect, DoT, buff-centric) assembled through the normal `BattleSnapshot`/battler path so the rating sees exactly what battles see.

## Notes

- Read-only over reference data; no persistence changes. Output is a table/report a human reads during tuning, not a runtime system.
- Worth keeping runnable (not deleted after use): re-running it after content additions or reference-constant changes is the cheap way to spot pricing drift.

Issue #1532, opened 2026-07-03
Swap the power measure onto the combat rating and reshape the reward curve (enemy-authored bounty)


ginderjeremiah:
Second half of spike #1526 (design + rationale in `docs/spikes/1526-combat-rating-power-measure.md`): swap both power-measure consumers onto the combat rating (#1531) **and** reshape the reward curve. The two must land together — the honest measure and the curve only work as a pair (see spike Decision 4: under the old `ratio²`-band curve, an honest measure makes build optimization *lower* XP/hour).

Blocked by #1531. Constants (`k`, pie) come from the calibration report issue.

## XP reward: enemy-authored bounty, no upward premium

```
XP/kill = k × enemyRating × min(enemyRating / playerRating, 1)²
```

- The `±20%` band, the quadratic upward premium, and the `MaxExpRewardMultiplier` cap on the XP path all dissolve into this one continuous curve. `min(r,1)² ≤ 1`, so XP is bounded by `k × enemyRating`; `MaxExpPerGrant` stays as the tamper backstop.
- Properties (verified in the spike): XP/hour peaks at matched fights; optimizing against any at-or-above-matched enemy strictly raises XP/hour (same bounty, faster kills); anti-grind below matched is unchanged in spirit; sandbagging becomes structurally pointless (nothing pays more for being weak).
- `DefeatRewards` takes rated battlers (or precomputed ratings) instead of modifier lists; `DifficultyMultiplier`'s doc comment and semantics update; the `ToIntReward` overflow clamp stays.

## Proficiency accrual: max-normalization

```
XP_path = pie × pathActivity ÷ max(playerRating, enemyRating)
```

- Above your weight the claim rate is what a matched player would earn (no weakness premium); at or below, the existing treadmill is unchanged. The `4.0` clamp mirroring `MaxExpRewardMultiplier` retires — `activity ≤ enemyHP` bounds the claim naturally (decide whether to keep a generous safety clamp; record the choice).
- `ProficiencyXpPerVictory` retunes per the calibration report (the rating's numeric scale differs from an attribute-point sum).

## Threading

- **Live**: `BattleService.RecordVictory` rates the player from the frozen snapshot (`ToBattler`, same anti-tamper rationale as today) and the enemy from its **fielded** loadout (`enemy.BattleSkills` — spike Decision 5: rate the fight actually fought; the level roll already varies per-encounter XP).
[...]

Issue #1531, opened 2026-07-03
Combat rating: closed-form capability rating computed from the assembled battler


ginderjeremiah:
Implements the **combat rating** decided in spike #1526 (design + rationale in `docs/spikes/1526-combat-rating-power-measure.md`). This issue is the rating function only — the consumer swap (XP/proficiency) is a follow-up issue.

## What

A pure, server-only `CombatRating` in `Game.Core.Battle`: `double Rate(Battler)` computed from the assembled battler (attribute collection + fielded `BattleSkills` + `CounterSkill`) against fixed reference profiles. No frontend mirror — the rating is never simulated client-side (display values are sent, not recomputed).

```
Rating = √(OffenseRate × Survivability)
```

The geometric mean is exact, not aesthetic: for any matchup, `(enemyRating / playerRating)²` **is** the time-to-kill ratio (time to kill the enemy ÷ time to be killed), which is the outcome-relevant quantity in this engine.

## Offense rate (expected damage/sec vs the reference defense)

Sum over fielded skills of `expectedHit ÷ effectiveCooldownSeconds`, plus DoT, reflect, and riposte terms:

- `expectedHit`: per-portion `raw × weight ÷ Σweights × (1 + amp)` × Toughness curve at `RefToughness` × crit expectation `1 + skill.CriticalChance × CriticalChanceMultiplier × (CriticalDamage − 1)` × execute expectation `1 + ExecuteBonus × 0.5` (missing-HP fraction averages ~½ over a full kill). Raw damage, amplification, and cooldown multipliers MUST be read through the same `Battler`/`BattleSkill` methods the engine uses (`CalculateRawDamage`, `AmplifyDamage`, `GetCooldownMultiplier`, the check-site chance products) — that is the anti-drift mechanism.
- **DoT effects** (effects targeting the Bleed/Poison/Burn accumulators on `Opponent`): steady-state DPS is exactly `magnitude × (1 + frozen amp) × durationSec ÷ effectiveCooldownSec` (the shared-expiry stacking rules make this the true closed form). DoT bypasses the reference Toughness (its real property — with `RefToughness > 0` this prices the bypass premium).
- **Reflect**: `DamageReflection × net incoming DPS vs the reference offense` (reflection returns post-mitigation damage — yes, less armor means more reflect output; true to the engine).
- **Riposte** (player-side only, `CounterSkill != null`): `effectiveParryChance × RefAttackRate × expectedHit(counterSkill)`.

[...]

PR #1535, opened 2026-07-03
Document spike #1526: combat-rating power measure decisions

### LINKED ISSUE #1526: Spike: combat-rating power measure — replace SumCoreAttributes with a capability rating
Follow-on spike from #1426 (fix the power measurement instead of contorting content around it) — context in `docs/spikes/1426-attribute-identity-sourcing.md`.

## Problem

`DefeatRewards.SumCoreAttributes` counts *investment* (the sum of additive core-attribute modifiers), but its two consumers — the XP difficulty band and the proficiency accrual denominator (`activity ÷ power`) — want *capability*:

- **Player and enemy don't scale equally per point** (owner-flagged): a point of END buys HP + Toughness; a point of enemy LUK buys nothing (enemies don't crit/dodge/parry) yet prices the enemy's XP payout up.
- **Dead stats are taxed**: an inert AGI/LUK allocation raises the denominator, lowering the XP band and every path's accrual — the sting behind #1426's newbie-trap question.
- **Authored secondaries are invisible**: an enemy's `CooldownRecovery` / `DamageReflection` distributions and all multiplicative modifiers count `0`.
- **Min-maxed and scattered builds read identically** at equal point totals.

## Seed direction — a combat rating from the assembled battle snapshot

Measure what the battler can *do*, both sides through the same function:

- **Offense rate**: Σ over fielded skills of `expectedHit ÷ effectiveCooldown` against a fixed reference defense profile — folding in the damage multipliers, amplification, and the crit expectation (`1 + chance × mult × (critDamage − 1)`); DoT as `amount × duration ÷ cooldown`. Skills cycle independently in this engine (loadout order is only a same-tick tie-break), so sum-of-rates is the correct closed form, not an approximation.
- **Survivability**: effective HP against a reference offense profile — `MaxHealth ÷ (1 − mitigation)`, scaled by expected avoidance (dodge/parry), plus a regen term.
- **Rating = √(offense × survivability)** — the geometric mean is scale-symmetric and resists degenerate all-offense/all-defense reads.

Properties: dead stats become *neutral* (contribute ~0) instead of taxed; enemy kits price correctly (the ±20% band centers per matchup instead of per-zone tuning compensation); multiplicatives and authored secondaries count; **server-only** (both consumers are post-battle server computations — no FE parity surface); offline stays cheap (power is stationary per away window, so the rating computes once). Per-attribute marginals double as the exact dead-stat detector for the enemy content lint.

## To resolve in the spike

- The reference profiles (the only tuning constants) — where they live (`ServerGameConstants`?) and how they're set.
- Fidelity cuts for v1: timed effects/self-buffs (rate statically? ignore with a note?), reflection, execute, and parry-counter expected damage.
- Migration: both consumers swap denominators — the band-retuning fallout, and whether the `PlayerPower` sent to the client changes meaning.
- The maintenance contract: the rating is a maintained model of combat value — each new mechanic adds a term (recorded and accepted in #1426 as better than the alternatives).

## Alternatives already weighed in #1426 (record, don't re-tread)

- **Per-attribute weight table** — chases the meta; the same maintained-balance liability as the authored `expectation(power)` curves rejected in #1398.
- **Battle-outcome-derived difficulty** — pays players for slow, sandbagged fights.
- **Level-based power** — ignores build strength entirely and breaks the allocations-count invariant.

ginderjeremiah:
Closes #1526.

Records the spike's ideation outcome in `docs/spikes/1526-combat-rating-power-measure.md`:

- **The capability rating** (#1531): `√(offense × survivability)` from the assembled battler vs fixed reference profiles, both sides through one function — with the TTK-ratio semantic that makes the geometric mean exact. v1 already prices steady-state buffs/debuffs, the DoT Toughness-bypass premium (nonzero `RefToughness`), and resists vs a uniform incoming mix.
- **The reward-curve reshape** (#1532), co-decided during the spike after the discussion exposed that an honest measure under the old `ratio²` band makes build optimization *lower* XP/hour: XP becomes an enemy-authored bounty `k × E × min(r,1)²` (no upward premium — sandbagging dies structurally) and the proficiency accrual max-normalizes by `max(playerRating, enemyRating)`.
- Supporting sub-issues: calibration report (#1533), UI power display (#1534); the rating's marginals feed the #1529 lint.

### DIFF
--- a/docs/spikes/1526-combat-rating-power-measure.md
+++ b/docs/spikes/1526-combat-rating-power-measure.md
@@ -1,0 +1,65 @@
+# Spike #1526 — Combat-rating power measure: replace SumCoreAttributes with a capability rating
+
+- **Spike issue:** [#1526](https://github.com/ginderjeremiah/GameServer/issues/1526)
+- **Status:** Ideation complete; design decided. Implementation split into sub-issues (see [Implementation issues](#implementation-issues)).
+- **Related:** follow-on from [#1426](./1426-attribute-identity-sourcing.md) (Decision 6); reshapes the reward curve shipped with the effect-based accrual ([#1318](./1318-proficiency-progression-avenues.md)); the rating's marginals feed the enemy dead-stat lint ([#1529](https://github.com/ginderjeremiah/GameServer/issues/1529)).
+
+## The question, as filed
+
+`DefeatRewards.SumCoreAttributes` counts *investment* (sum of additive core-attribute modifiers) where its two consumers — the XP difficulty band and the proficiency accrual denominator — want *capability*: player and enemy don't scale equally per point, dead stats are taxed, authored secondaries and multiplicatives are invisible. Seed direction: a combat rating from the assembled battle snapshot.
+
+## What research sharpened
+
+- **The defect is slightly worse than filed**: the enemy's attribute sum is not just the band input but the **XP magnitude itself** (`ExpReward = enemyAttTotal × multiplier`), so swapping the measure rescales all payouts, not just band placement.
+- **`PlayerPower` never reaches the client** (server-side only; the API DTO omits it), so there was no display-meaning migration — instead the rating becomes a deliberate UI feature (Decision 7).
+- **Sequencing is independent of the #1426 severing issues** (#1523/#1524/#1525): the rating prices whatever the engine currently reads (e.g. enemy `CooldownRecovery` fires today and is priced; enemy `CriticalDamage` never fires and is neutral), and the opt-in pairs go dormant-neutral automatically once severed.
+- **The discussion exposed a second, co-equal defect in the reward curve itself** (Decision 4): under the `ratio²` band, an *honest* measure makes build optimization lower XP/hour — the old measure's unpriced channels were the only thing masking it.
+
+## Decisions
+
+Settled with the project owner during the spike:
+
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: spike #1526 outcome doc — combat-rating power measure

**Verdict:** Approve *(submitted as a comment because I'm the PR author; GitHub won't accept a gating self-review)* — a docs-only PR that records a completed spike per the `CLAUDE.md` spike rule, with every claim verified.

Verified against the spike rule ("document the final results… and reference the created issues, linked to the parent spike for tracking"):

- **Sub-issue linkage is real, not just asserted.** #1531, #1532, #1533, #1534 all exist, are open, and each carries a parent link to #1526 — the doc's "Created as sub-issues of #1526" is accurate.
- **The doc resolves every open question filed in #1526** — reference constants location (Decision 10 → `ServerGameConstants`), v1 fidelity cuts (Decisions 1–2: steady-state effects, execute-at-½, riposte; reflection detailed in #1531), the migration question incl. the correct finding that `PlayerPower` never reaches the client so there's no display-meaning migration (Decision 7 turns it into a UI feature instead), and the maintenance contract (Decision 9: exhaustiveness test + marginal helper).
- **All cross-referenced local docs exist** (`1426-…`, `1318-…`, `backend-attributes.md`, `content-design.md`, `game-design.md`), and the "Documentation to update on landing" section correctly defers the game-design/content-design/backend-attributes edits to #1532 rather than doing them here.
- **Scope is honest**: the research "What sharpened" section usefully flags that the defect was worse than filed (enemy attribute sum is the XP magnitude, not just the band input) and that the reward-curve reshape (Decision 4) is a co-equal defect surfaced during the spike — good spike hygiene.

### What's good

- The geometric-mean justification is tied to a concrete engine quantity (`(enemyRating/playerRating)²` = time-to-kill ratio) rather than left as aesthetic — that rationale carries straight into the #1531 issue body, so the implementer inherits the *why*.
- Alternatives are recorded-not-re-treaded (deferring to #1426) and the newly-rejected option (any upward `rⁿ` premium) is called out with its reasoning — exactly what a spike record is for.
- The implementation split is a clean dependency graph (#1531 → #1532, calibration #1533 feeding constants, UI #1534 last) with scope tags, so the follow-on work is directly actionable.

No blocking, should-fix, or nit findings. CI is green (test jobs correctly skipped for a docs-only change; `all-checks-passed` succeeded). Approving and merging.

---

PR #1536, opened 2026-07-03
Fix SavePlayer's write-behind batch losing events on a failed publish

### LINKED ISSUE #1494: SavePlayer's batch flush drains before publishing — a failed LPUSH permanently loses that save's persistence events
## Problem

`PlayerRepository.SavePlayer` (`Game.DataAccess/Repositories/PlayerRepository.cs:98`) flushes the scoped batch with:

```csharp
await _pubsub.PublishBatch(..., _updateBatch.Drain(), cancellationToken);
```

`PlayerUpdateBatch.Drain()` empties the buffer **before** the LPUSH is attempted. If `PublishBatch` throws (a transient Redis blip/timeout), the drained envelopes are discarded — no re-buffer, no retry, and the consume-side retry policy never sees them. The aggregate's events were already snapshotted and cleared by `DomainEventDispatcher.DispatchAsync` before this point, so nothing re-raises them.

## Failure scenario

The in-memory aggregate keeps the mutation, so the next save re-writes the full **cache blob** (Redis looks correct) — but one-shot child-row events (`ItemUnlockedEvent`, `SkillUnlockedEvent`, `ModUnlockedEvent`, equip changes) are never enqueued again, so Postgres never learns of them. When the player key ages out of Redis (48h idle TTL), the DB fallback reload is missing the unlock: **permanent player-data loss**.

A second wrinkle on the same failed flush: `RunFlushedCallbacks()` is skipped but `_onFlushed` is **not cleared** (`Game.DataAccess/PlayerUpdateBatch.cs`). The next *successful* `SavePlayer` in the scope then runs the stale progress `AdvanceCache` callback (captured snapshot) after a flush that does **not** contain its event — advancing the progress source-of-truth cache to a state that was never queued, the exact violation the documented publish-before-cache ordering (`docs/backend-persistence.md`) exists to prevent.

Note the progress repo defends exactly this class of failure (pinned by `Save_PublishFails_LeavesCacheUnadvanced` in `Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs`), but that defense doesn't cover the player aggregate because its working copy is the in-memory object, not the cache.

## Suggested fix

- Drain only after a successful publish (or catch-and-restore the drained envelopes into the batch so the connection's next save re-flushes them).
- Clear/invalidate `_onFlushed` when the flush that owns it fails.
- Test shape mirrors the existing progress-side publish-failure test.

Related (distinct) prior work: #769 covered the consume-side destructive pop; #412 covered a handler throwing mid-dispatch. This is the produce-side flush.

ginderjeremiah:
## Summary

`PlayerRepository.SavePlayer` flushed its scoped `PlayerUpdateBatch` by calling `_updateBatch.Drain()` *as the argument to* `PublishBatch(...)` — which empties the buffer synchronously before the LPUSH is even attempted. If `PublishBatch` throws (a transient Redis blip/timeout), the drained envelopes are gone with no way to recover them: the aggregate's events were already snapshotted and cleared by `DomainEventDispatcher.DispatchAsync` before this point, so nothing re-raises them. For one-shot child-row events (`ItemUnlockedEvent`, `SkillUnlockedEvent`, etc.) this is a permanent loss once the player's Redis key eventually ages out and a DB reload can't find the change.

A second, related bug: on a failed flush, `RunFlushedCallbacks()` was skipped but `_onFlushed` was never cleared. If a *later* successful `SavePlayer` in the same command scope flushed again (real — e.g. `StartBattle` can save the player once for a zone change and again for the battle result), it would run the stale progress `AdvanceCache` callback even though that flush never actually contained the earlier event — advancing the progress cache to a state that was never durably queued.

## Fix

Added `PlayerUpdateBatch.FlushAsync(IPubSubService pubsub, CancellationToken)`, which publishes the *live* buffer directly (safe — `PublishBatch` materializes it synchronously before its first `await`) and only drains it / runs `OnFlushed` callbacks once the publish has actually succeeded. On failure, both the buffered events and any deferred callbacks are left exactly as they were, so they're carried into the batch's next flush attempt in the same scope instead of being silently discarded.

`PlayerRepository.SavePlayer` and `PlayerProgressRepository.Save`'s standalone-flush branch both previously duplicated the buggy drain-then-publish pattern; they now share this one safe helper.

Documented the new contract in `docs/backend-persistence.md`.

## Test plan

- [x] New unit tests in `PlayerUpdateBatchTests` exercising `FlushAsync` success/failure via a hand-written fake `IPubSubService` (classical/state-based style, matching the project's testing conventions) — proves a failed publish preserves both the buffered event and the deferred callback for the next flush.
- [x] New integration test `PlayerWriteBehindTests.SavePlayer_PublishFails_PreservesTheEventForTheNextFlush` — a cancelled-token publish failure followed by a normal save proves the first save's event isn't lost.
- [x] Updated `PlayerProgressRepositoryIntegrationTests` to flush via the new helper, matching production code.
- [x] Full backend suite: `dotnet build` + `dotnet test` — 2742 tests passing.

Closes #1494

---
_Generated by [Claude Code](https://claude.ai/code/session_0151zVagcnzXsVuFeDRE1U7K)_


Game.DataAccess/PlayerUpdateBatch.cs
```diff
@@ -72,8 +74,24 @@ public IReadOnlyList<DomainEventEnvelope> Drain()
...
-        /// after its flush so a batched progress save's deferred cache advance lands only once its event is
+        /// Publishes the buffered batch to the write-behind queue and only then clears it and runs any
+        /// deferred <see cref="OnFlushed"/> callbacks. If the publish throws (a transient Redis blip/timeout),
+        /// the buffered envelopes and callbacks are left exactly as they were — nothing is drained or run —
+        /// so they are carried into the batch's next flush attempt instead of being silently discarded (#1494).
+        /// Passing the live buffer directly (rather than draining a snapshot first) is safe because
+        /// <see cref="Game.Abstractions.Infrastructure.IPubSubService.PublishBatch{T}"/> materializes it
+        /// synchronously before its first await.
+        /// </summary>
+        public async Task FlushAsync(IPubSubService pubsub, CancellationToken cancellationToken = default)
```
ginderjeremiah, line 70: `FlushAsync` now fully replaces the drain-then-publish flow at both former call sites (`PlayerRepository.SavePlayer` and `PlayerProgressRepository.Save`), and it clears via `_events.Clear()` directly rather than calling `Drain()`. That leaves `Drain()` (line 64) with no remaining production caller — its only references are the tests that exercise it directly. Per CLAUDE.md ("If you are removing or replacing functionality, make sure to remove any code that is no longer being used as well"), consider removing `Drain()` and its now-orphaned tests as part of this PR rather than leaving replaced code behind. (Should-fix, not blocking.)

---

ginderjeremiah (commented):
## Review: Fix `SavePlayer`'s write-behind batch losing events on a failed publish (closes #1494)

**Verdict:** Approve — with one should-fix (dead-code cleanup). Submitted as a `COMMENT` because I'm the PR author; treat the intent as an approve conditional on the cleanup below.

The fix is correct and squarely addresses both bugs described in #1494:

1. **Primary bug (event loss):** `FlushAsync` publishes the live buffer and only calls `_events.Clear()` **after** the publish awaits successfully, so a throwing `PublishBatch` leaves the envelopes buffered for the next flush in the same scope instead of discarding them. Passing the live `_events` reference is safe — `RedisPubSubService.PublishBatch` materializes it via `.ToArray()` (`RedisPubSubService.cs:73`) synchronously before its first `await`, and the `Clear()` runs after the full await completes regardless.
2. **Secondary bug (stale `_onFlushed`):** because a failed publish never reaches `RunFlushedCallbacks()`, the deferred progress cache-advance is neither run nor cleared — it now travels with its event to the next flush, so the callback and the event that justifies it stay paired. This preserves the documented publish-before-cache ordering.

Consolidating both former call sites onto the one `FlushAsync` helper is a good DRY improvement, and the accepted "occasional harmless duplicate over a silent loss" trade-off is consistent with the consume side's at-least-once/idempotent-handler contract.

### Should-fix

- **`Drain()` is now dead production code** — `Game.DataAccess/PlayerUpdateBatch.cs:64`. `FlushAsync` replaced both callers and clears via `_events.Clear()` directly, so nothing in production calls `Drain()` anymore; only its own tests do. CLAUDE.md requires removing replaced/unused code as part of the change — remove `Drain()` and its now-orphaned tests (`PlayerUpdateBatchTests` lines ~24/33/35/43), or state why it's being kept.

### Nits

- `RunFlushedCallbacks()` is now only called from within `FlushAsync` (production) — it could become `private`, but it's still exercised by direct unit tests, so leaving it public is defensible. Optional.

### Test coverage

Good and following the project's classical/state-based conventions:
- `FlushAsync_PublishSucceeds_...` and `FlushAsync_PublishThrows_...` cover both the happy path (buffer cleared, callback runs once) and the failure path (event + callback preserved for the next flush) with hand-written fakes — no mocking framework.
- Integration test `SavePlayer_PublishFails_PreservesTheEventForTheNextFlush` exercises the real Redis path via a pre-cancelled token, and correctly documents/handles the client-side-cancellation caveat (asserting `>= 2` rather than an exact untouched-queue state).
- Backend-only change (not battle logic), so no frontend mirror is required.
- The `Drain()` tests should be removed alongside `Drain()` per the should-fix above.

### What's good

- The `FlushAsync` XML doc and the `backend-persistence.md` update capture the *why* (buffer cleared only on success; duplicates accepted over silent loss) — exactly the kind of non-obvious design decision the docs should carry.
- Clean, minimal diff that fixes the root cause rather than papering over it.

Not merging yet: the should-fix is outstanding and CI is still in progress (`test-backend` and two `test-e2e` browsers running). Once `Drain()`/its tests are removed and CI is green, this is good to merge.

---

ginderjeremiah:

Reviewed against the code, not just the description — both core claims hold up. I confirmed that `RedisPubSubService.PublishBatch<T>` materializes its input via `.ToArray()` synchronously before its first `await`, so passing the live `_events` buffer and clearing it only after the publish succeeds is safe. I also reproduced the stale-`_onFlushed` reasoning against the pre-PR code: because `Drain()` emptied the buffer unconditionally and `RunFlushedCallbacks()` was skipped-but-not-cleared on a failed publish, a later same-scope save (real via `StartBattle`'s zone-change + battle-result double save) would have advanced the progress cache to a snapshot that was never queued — exactly the publish-before-cache violation the issue describes.

The fix is correct and nicely consolidating: `FlushAsync` publishes first and only then drains the buffer and runs/clears the deferred callbacks, so a transient Redis failure carries both the events and the callbacks into the next flush instead of silently discarding them. Both `PlayerRepository.SavePlayer` and `PlayerProgressRepository.Save`'s standalone branch now share the one helper, and `Drain()` is removed cleanly with no orphaned callers. Domain isolation is intact — `PlayerUpdateBatch` stays an internal data-tier type — and no null-forgiving operators or missing braces were introduced.

Test coverage is solid: the new unit tests pin both the buffered-event preservation and the callback preservation (`flushCount` stays 0 on failure, 1 after the next success), directly covering both bugs, and the integration test appropriately asserts "nothing permanently lost" given the client-side cancellation nondeterminism it documents. The `backend-persistence.md` update accurately captures the new contract and honestly frames the accepted duplicate-vs-loss trade-off. One minor non-blocking nit: the two `IPubSubService` test fakes each re-stub ~11 identical `NotSupportedException` members — I filed #1541 (tech-debt) to track sharing a base fake.

CI is green across backend and all e2e browsers. Merging.

---

PR #1537, opened 2026-07-03
Validate client-supplied zone ids before resolving in StartBattle/StartBossBattle

### LINKED ISSUE #1495: Client-supplied zone ids are resolved before validation in StartBattle/StartBossBattle — opaque 500s and a re-creditable stranded battle
## Problem

Both battle-start paths resolve a client-supplied zone id via `_zones.GetDomainZone(id)` **before** any bounds check; `Zones.GetDomainZone` throws `ArgumentOutOfRangeException` on an out-of-range id (`Game.DataAccess/Repositories/Zones.cs:31-40`):

- `BattleService.StartBattle` (`Game.Application/Services/BattleService.cs:89`) — reached from `NewEnemy` with an unvalidated `Parameters.NewZoneId`.
- `BattleService.StartBossBattle` (`BattleService.cs:181`) — reached from `ChallengeBoss.Parameters.ZoneId`.

Every sibling anti-cheat path rejects gracefully (`SetAutoChallengeBoss` shows the intended pattern: `ValidateZoneId` first); here a tampered client gets `"Internal Server Error"` plus error-level log noise on demand. The docs' claim that an invalid zone target is "ignored and the battle proceeds in the current zone" only holds for valid-but-locked/retired ids.

## The worse consequence on the `StartBattle` path

`StartBattle` runs `AbandonBattle` **first** (lines 75-78), which re-simulates and — if the replay scores a win — credits the battle via `RecordVictory` + `SavePlayer`, clearing the active battle only on the in-memory `PlayerState`. Then `GetDomainZone(9999)` throws → the command faults → `context.Session.SavePlayerState()` in `NewEnemy` never runs → the Redis session still holds the already-credited battle as active. On reconnect, the rehydrated session shows the same battle in flight, and the next resolve **credits it again** — the exact hazard `TryPrepareNextIdleBattle`'s docstring describes, here deliberately triggerable: win → bad-zone request → reconnect → repeat = exp/statistics duplication.

## Suggested fix

Guard with `_zones.ValidateZoneId(...)` before resolving in both paths: ignore the bogus target in `StartBattle` (proceed in the current zone, matching the documented behavior), return a graceful null/failure in `StartBossBattle`. Add anti-cheat tests for the out-of-range id on both commands.

ginderjeremiah:
## Summary

Both battle-start paths resolved a client-supplied zone id via `_zones.GetDomainZone(id)` **before** any bounds check. `Zones.GetDomainZone` throws `ArgumentOutOfRangeException` on an out-of-range id, so a tampered client could trigger an opaque "Internal Server Error" instead of a graceful no-op:

- `BattleService.StartBattle` — reached from `NewEnemy` with an unvalidated `Parameters.NewZoneId`.
- `BattleService.StartBossBattle` — reached from `ChallengeBoss.Parameters.ZoneId`.

On the `StartBattle` path this was also the worse hazard described in the issue: `StartBattle` runs `AbandonBattle` first, which can credit a win in-memory via `RecordVictory`/`SavePlayer`. If the subsequent `GetDomainZone` then throws on a bad `newZoneId`, the command faults before `SavePlayerState()` runs, so the Redis session still shows the already-credited battle as active — a reconnect then re-credits it (exp/statistics duplication).

## Fix

Guarded both paths with `IZones.ValidateZoneId` before resolving, matching the pattern already used by the sibling anti-cheat path `SetAutoChallengeBoss`:

- `StartBattle`: an out-of-range `newZoneId` is now ignored, same as a locked/retired/Home target — the battle proceeds in the player's current zone (matches the documented behavior).
- `StartBossBattle`: an out-of-range `zoneId` now returns `null` before touching the active battle, same as the existing bossless/locked/retired no-op checks — `ChallengeBoss` already handles a `null` result gracefully.

## Test plan

- [x] New integration test `StartBattle_TransitionToOutOfRangeZone_KeepsPlayerInCurrentZone` — mirrors the existing locked/retired-zone tests, asserting the player stays in their current zone and the battle still starts.
- [x] New integration test `StartBossBattle_OutOfRangeZone_ReturnsNullAndStartsNoBattle` — mirrors `StartBossBattle_RetiredZone_ReturnsNullAndStartsNoBattle`.
- [x] Full backend suite: `dotnet test --no-build --no-restore` — 2741 tests passing.

Closes #1495

---
_Generated by [Claude Code](https://claude.ai/code/session_01GKeM8Wkah9MUj64bJGihKp)_

### DIFF
--- a/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
@@ -1414,6 +1414,40 @@ namespace Game.Application.Tests.Services
         }
 
         [Fact]
+        public async Task StartBattle_TransitionToOutOfRangeZone_KeepsPlayerInCurrentZone()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+
+            var skill = await TestDataSeeder.CreateSkillAsync(context);
+            var enemy = await TestDataSeeder.CreateEnemyAsync(context);
+            await TestDataSeeder.LinkSkillToEnemyAsync(context, enemy.Id, skill.Id);
+            var zone = await TestDataSeeder.CreateZoneAsync(context);
+            await TestDataSeeder.LinkEnemyToZoneAsync(context, zone.Id, enemy.Id);
+
+            var user = await TestDataSeeder.CreateUserAsync(context);
+            var playerEntity = await TestDataSeeder.CreatePlayerAsync(context, user.Id, zoneId: zone.Id);
+            await TestDataSeeder.LinkSkillToPlayerAsync(context, playerEntity.Id, skill.Id);
+
+            await ReloadReferenceCachesAsync();
+
+            var playerRepo = scope.ServiceProvider.GetRequiredService<IPlayerRepository>();
+            var player = await playerRepo.GetPlayer(playerEntity.Id);
+            Assert.NotNull(player);
+
+            var battleService = scope.ServiceProvider.GetRequiredService<BattleService>();
+            var state = new PlayerState();
+
+            // A tampered request naming an out-of-range zone id is ignored — not thrown — exactly like a
+            // locked or retired target: the player stays put and the battle proceeds in their current zone.
+            var result = await battleService.StartBattle(player, state, zoneId: zone.Id, newZoneId: 999);
+
+            Assert.NotNull(result);
+            Assert.Equal(zone.Id, player.CurrentZoneId);
+            Assert.Equal(zone.Id, state.BattleZoneId);
+        }
+
+        [Fact]
         public async Task StartBattle_TransitionToHomeZone_KeepsPlayerInCurrentZone()
         {
             using var scope = CreateScope();
@@ -1596,6 +1630,35 @@ namespace Game.Application.Tests.Services
             var state = new PlayerState();
 
             var result = await battleService.StartBossBattle(player, state, zone.Id);
+
+            Assert.Null(result);
+            Assert.False(state.HasActiveBattle);
+        }
+
+        [Fact]
+        public async Task StartBossBattle_OutOfRangeZone_ReturnsNullAndStartsNoBattle()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+
+            var playerSkill = await TestDataSeeder.CreateSkillAsync(context, name: "PlayerSkill");
+            var zone = await TestDataSeeder.CreateZoneAsync(context);
+            var user = await TestDataSeeder.CreateUserAsync(context);
+            var playerEntity = await TestDataSeeder.CreatePlayerAsync(context, user.Id, zoneId: zone.Id);
+            await TestDataSeeder.LinkSkillToPlayerAsync(context, playerEntity.Id, playerSkill.Id);
+
+            await ReloadReferenceCachesAsync();
+
+            var playerRepo = scope.ServiceProvider.GetRequiredService<IPlayerRepository>();
+            var player = await playerRepo.GetPlayer(playerEntity.Id);
+            Assert.NotNull(player);
+
+            var battleService = scope.ServiceProvider.GetRequiredService<BattleService>();
+            var state = new PlayerState();
+
+            // A tampered ChallengeBoss request naming an out-of-range zone id is a graceful no-op — not an
+            // ArgumentOutOfRangeException surfacing as an opaque 500 — like the bossless/locked/retired cases.
+            var result = await battleService.StartBossBattle(player, state, 999);
 
             Assert.Null(result);
             Assert.False(state.HasActiveBattle);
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Validate client-supplied zone ids before resolving in StartBattle/StartBossBattle (#1495)

**Verdict:** Approve — correct, minimal, well-scoped fix that mirrors the established anti-cheat pattern. _(Formal event is `COMMENT` only because I'm the PR author; GitHub won't accept a gating verdict on one's own PR.)_

I traced both paths and the fix is sound:

- `IZones.ValidateZoneId` (`Zones.cs:69-77`) is exactly the `zoneId >= 0 && zoneId < Count` bounds check that `GetDomainZone` (`Zones.cs:31-40`) applies before throwing — so guarding with it genuinely prevents the `ArgumentOutOfRangeException`.
- **StartBattle** (`BattleService.cs:87`): folding `ValidateZoneId` into the zone-change condition short-circuits the whole block on a bad `newZoneId`, so `GetDomainZone` is never reached and the command completes normally. That's the crux of the issue's "worse consequence" — since the throw was the sole reason `SavePlayerState()` was skipped after `AbandonBattle` credited a win, eliminating the throw eliminates the stranded, re-creditable battle. Behavior (proceed in current zone) matches the documented contract.
- **StartBossBattle** (`BattleService.cs:181-187`): guard placed correctly *before* `AbandonBattle` and `GetDomainZone`, returning `null` like the sibling bossless/retired/locked no-ops — and `ChallengeBoss` already handled `null`, so no caller contract changes.
- The `zoneId` param at `BattleService.cs:105` is not a second hole: `NewEnemy.cs:35` always passes `player.CurrentZoneId`, and `EnsureViableZone` already validates. The only client-controlled input on this path is `NewZoneId`, which is now guarded.
- The shape matches the `SetAutoChallengeBoss` sibling (`ValidateZoneId` → `IsZoneRetired` → resolve) precisely, so the codebase stays internally consistent.

No DDD/layering concerns — this is orchestration-layer anti-cheat using a data-access abstraction, no domain logic leaking. The multi-line comments earn their length (they encode non-obvious anti-cheat invariants).

### Nits
- Both new tests start from a fresh `PlayerState` with no active battle, so they prove the graceful-ignore behavior but not the headline hazard the issue leads with — an *active* battle whose `AbandonBattle` win-credit was being stranded by the throw. Existing abandon tests + this no-throw test cover it transitively, and the added tests are exactly what the issue asked for, so this is optional: a single test that starts an active battle, wins on abandon, then requests an out-of-range `newZoneId` and asserts the credit is persisted (not re-creditable on a reload) would pin the regression more directly. Take it or leave it.

### Test coverage
- Adequate. `StartBattle_TransitionToOutOfRangeZone_KeepsPlayerInCurrentZone` and `StartBossBattle_OutOfRangeZone_ReturnsNullAndStartsNoBattle` mirror the existing locked/retired sibling tests one-for-one, which is precisely the coverage #1495 requested. No frontend mirror needed — zone-id anti-cheat is server-authoritative, not deterministic battle simulation.

### What's good
- Tight, surgical fix that reuses the existing pattern rather than inventing a new guard. Comment updates keep the "gated on being in range, unlocked, in circulation, and a combat zone" invariant honest. Clear PR writeup that correctly identifies the deeper stranded-battle consequence.

I'm holding off on auto-merging: CI is still running (`test-backend` and `test-e2e (webkit)` in progress at review time) and the note above is a small outstanding item for you to accept or wave off. Once CI is green, this is good to merge as-is.

---

ginderjeremiah:

Reviewed and verified against the actual code — this is a clean, correctly-scoped fix. Both battle-start paths did resolve a client-supplied zone id via `GetDomainZone` before any bounds check, and `Zones.GetDomainZone` throws `ArgumentOutOfRangeException` on an out-of-range id, so the "opaque 500" and the stranded-battle double-credit described in #1495 are both real. I confirmed the duplicate-exp mechanism specifically: `StartBattle` runs `AbandonBattle` first, which on a won replay credits exp via `RecordVictory` + `SavePlayer` and clears the battle only on the in-memory `PlayerState`; a throw at the subsequent `GetDomainZone` then faults the command before the socket handler persists the cleared session, so a reconnect re-abandons and re-credits the same win.

The `IZones.ValidateZoneId` guards resolve this correctly and idiomatically — they mirror the sibling anti-cheat pattern already used by `SetAutoChallengeBoss`, `IsZoneViable`, and `ResolveOfflineLoop`. `StartBattle` now falls through to the current-zone battle (matching the documented locked/retired behavior), and `StartBossBattle` returns `null` before touching the active battle, which `ChallengeBoss` already handles. Domain isolation, brace/style guidelines, and the no-null-forgiving rule are all respected.

Test coverage mirrors the existing locked/retired pairs. Two minor, non-blocking notes: (1) the flagship `...CreditsVictoryWithoutThrowing` test asserts no-throw + exp credited + battle-still-active, a solid proxy but not a direct "credited exactly once" assertion — that harm lives at the session-persistence layer above `BattleService`, so it's reasonably out of scope for a service-level test; (2) the `!ValidateZoneId` + retired/unlock preamble now recurs across ~5 call sites. I filed #1540 (tech-debt) to track optionally extracting a shared guard.

CI is green (backend, e2e, verify all pass; frontend jobs correctly skipped). Merging.

---

PR #1538, opened 2026-07-03
Settle WaitSocketClosed on an abrupt (Aborted) disconnect

### LINKED ISSUE #1496: Abrupt client disconnect never settles WaitSocketClosed — socket registration, pub/sub subscription, and middleware task leak until shutdown
## Problem

`SocketContext`'s `_socketClosedSource` is settled **only** inside `SocketContext.Close()` (`Game.Api/Sockets/SocketContext.cs`). The read loop's terminal step (`Game.Api/Sockets/SocketHandler.cs:336-339`) is:

```csharp
if (_context.State is Open or CloseReceived)
{
    await _context.Close(ESocketCloseReason.Finished);
}
```

On a hard disconnect (TCP reset, killed browser, network drop) `ReceiveAsync` throws and the WebSocket transitions to **Aborted** — the guard is false, `Close` is never called, and the TCS never settles. `SocketInterceptorMiddleware` (`Game.Api/Middleware/SocketInterceptorMiddleware.cs:50-53`) then awaits `WaitSocketClosed()` forever: `UnRegisterSocket` never runs, leaking the Redis pub/sub subscription, the socket-registry entry, and the parked middleware task **per abrupt disconnect for the process lifetime** (only a shutdown drain's explicit `ShutdownAsync` → `Close` releases them).

The comment right above the guard even says "Close is a no-op on an already-closed/aborted socket" — but the guard prevents it from being *called* on Aborted. `SocketReadLoopTests` covers only the throw-while-still-Open case (where the guard does fire).

## Suggested fix

Call `Close` unconditionally after the loop (it already re-checks state internally), or settle the closed-TCS whenever the read loop exits. Add a read-loop test for the receive-throws-with-Aborted-state case asserting `WaitSocketClosed` completes.

Related history: #909 hardened registration/teardown ordering but did not cover this abort path.

ginderjeremiah:
## Summary

`SocketHandler.ReadLoop`'s terminal step only called `SocketContext.Close()` when the socket was still in the `Open` or `CloseReceived` state:

```csharp
if (_context.State is Open or CloseReceived)
{
    await _context.Close(ESocketCloseReason.Finished);
}
```

On a hard disconnect (TCP reset, killed browser, network drop), `ReceiveAsync` throws and the WebSocket transitions itself straight to `Aborted` — the guard above is then false, `Close()` is never called, and `SocketContext`'s `_socketClosedSource` TCS never settles. `SocketInterceptorMiddleware` awaits `WaitSocketClosed()` right after registering the socket, so it hangs forever on an abrupt disconnect: `UnRegisterSocket` never runs, leaking the Redis pub/sub subscription, the socket-registry entry, and the parked middleware task for the process lifetime (only an explicit shutdown drain releases them).

## Fix

Call `Close()` unconditionally after the read loop instead of gating it on state. `Close()` already re-checks `_socket.State` internally and only sends a close frame when still `Open` — its `finally` unconditionally settles the `WaitSocketClosed` TCS regardless of state — so this is a no-op send (but a real settle) on an already-closed/aborted socket, and unchanged behavior for the Open/CloseReceived cases the old guard handled.

## Test plan

- [x] New unit test `ReadLoop_ReceiveThrowsWithSocketAborted_StillSettlesWaitSocketClosed` in `SocketReadLoopTests` — scripts a receive that throws *and* transitions the socket to `Aborted` (distinct from the existing still-Open throw case), and asserts `WaitSocketClosed()` still completes.
- [x] Full backend suite: `dotnet test --no-build --no-restore` — 2740 tests passing.

Closes #1496

---
_Generated by [Claude Code](https://claude.ai/code/session_01GSuq6aJBLUxm23RGpTo2q9)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketReadLoopTests.cs
+++ b/Game.Api.Tests/Unit/SocketReadLoopTests.cs
@@ -63,6 +63,31 @@ namespace Game.Api.Tests.Unit
         }
 
         [Fact]
+        public async Task ReadLoop_ReceiveThrowsWithSocketAborted_StillSettlesWaitSocketClosed()
+        {
+            // A hard disconnect (TCP reset, killed browser) throws from ReceiveAsync AND transitions the
+            // WebSocket to Aborted — unlike the still-Open throw above. The read loop must still call Close()
+            // so WaitSocketClosed settles (issue #1496): otherwise SocketInterceptorMiddleware awaits it
+            // forever and the socket registration, pub/sub subscription, and middleware task leak.
+            var socket = new ScriptedReadWebSocket();
+            socket.QueueThrow(new WebSocketException("hard disconnect"), resultingState: WebSocketState.Aborted);
+            var (context, handler) = CreateHandler(socket);
+
+            using var inactivityStop = new CancellationTokenSource();
+            handler.Listen(hostStopping: inactivityStop.Token);
+
+            await context.WaitSocketClosed().WaitAsync(WaitTimeout, TestContext.Current.CancellationToken);
+            inactivityStop.Cancel();
+            await handler.Completion.WaitAsync(WaitTimeout, TestContext.Current.CancellationToken);
+
+            // Close() re-checks state and does not send a close frame on an already-Aborted socket, but
+            // WaitSocketClosed above still had to complete for this assertion to be reached.
+            Assert.Equal(1, socket.ReceiveAttempts);
+            Assert.False(socket.CloseAsyncCalled);
+            Assert.Equal(WebSocketState.Aborted, socket.State);
+        }
+
+        [Fact]
         public async Task ReadLoop_ClientCloseFrame_CompletesTheClosingHandshake()
         {
             // A client-initiated close still drives the read loop out of the CloseReceived state and completes
@@ -154,7 +179,16 @@ namespace Game.Api.Tests.Unit
             /// <summary>Completes when the loop calls receive with no scripted input left — i.e. it looped back to read again.</summary>
             public Task IdleReadReached => _idleReadReached.Task;
 
-            public void QueueThrow(Exception toThrow) => _throwOnReceive = toThrow;
+            /// <param name="resultingState">
+            /// The state the socket transitions to as part of the throw. Defaults to <see cref="WebSocketState.Open"/>
+            /// (a throw that does NOT transition state — the terminal-fault case); pass <see cref="WebSocketState.Aborted"/>
+            /// to script a hard disconnect, where the underlying WebSocket transitions itself before the read loop observes it.
+            /// </param>
+            public void QueueThrow(Exception toThrow, WebSocketState resultingState = WebSocketState.Open)
+            {
+                _throwOnReceive = toThrow;
+                _state = resultingState;
+            }
             public void QueueClose() => _closeFrameQueued = true;
             public void QueueMessage(string message) => _messageQueued = Encoding.UTF8.GetBytes(message);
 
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Settle `WaitSocketClosed` on an abrupt (Aborted) disconnect (#1496)

**Verdict:** Approve — the fix correctly closes the leak, and the reasoning behind making the `Close()` call unconditional holds up against the source. (Submitted as a comment since I'm the PR author; treat it as an approval.)

### Analysis

The root cause and fix check out against `SocketContext.Close` (`Game.Api/Sockets/SocketContext.cs:155-184`): the `CloseAsync` send is guarded on `_socket.State is Open` (re-checked under the send lock), while the `finally` unconditionally calls `_socketClosedSource.TrySetResult`. So dropping the `if (_context.State is Open or CloseReceived)` guard and calling `Close()` unconditionally settles the TCS on an `Aborted` socket with **no** spurious close frame, and leaves the `Open`/`CloseReceived` paths behaving exactly as before. That's the minimal, correct fix for the leak.

I also confirmed this doesn't regress the client-initiated close path: `Close()` never emitted a frame on `CloseReceived` (only `Open`), so the existing `ReadLoop_ClientCloseFrame_CompletesTheClosingHandshake` behavior is unchanged.

### Nits

- `SocketReadLoopTests.cs:165` — the `ScriptedReadWebSocket` class doc still says "A throw leaves `State` Open — exactly the case that would spin the loop." That's now only the *default* for `QueueThrow`; the new overload can leave the socket `Aborted`. Minor staleness worth a one-line tweak while you're here. (Not on a diff line, hence here rather than inline.)

### Test coverage

Good. The new `ReadLoop_ReceiveThrowsWithSocketAborted_StillSettlesWaitSocketClosed` mirrors the existing terminal-fault test and pins the exact gap from #1496 — reaching its asserts *is* the proof that `WaitSocketClosed` settled — while also asserting no close frame is sent on `Aborted`. The `QueueThrow(resultingState:)` refactor keeps the still-Open and clean-close tests intact. Setting the socket state to `Aborted` at queue time (rather than during the throwing receive) is a fair simplification for a unit-level stand-in and is called out in the param doc.

### What's good

Tightly scoped one-line behavior change with a clear, accurate comment explaining *why* the call is unconditional, and a test that would actually catch a regression. Exactly the shape a fix like this should take.


---

ginderjeremiah:

I verified the bug against the actual code rather than the description: `SocketContext.Close()` settles `_socketClosedSource` in its `finally` regardless of state, but only sends a close frame while `Open` — so the old `if (_context.State is Open or CloseReceived)` guard in `ReadLoop` genuinely skipped the call on an `Aborted` socket, leaving `WaitSocketClosed()` unsettled and `SocketInterceptorMiddleware` parked forever without ever reaching `UnRegisterSocket`. The leaked Redis subscription, registry entry, and middleware task are all real.

Calling `Close()` unconditionally is the right fix and is safe: on a non-`Open` socket it does only the idempotent `TrySetResult` settle (no `_sendLock` wait, no `CloseAsync`), and it's harmless on paths that already closed the socket since the second settle no-ops. Behavior for the `Open`/`CloseReceived` cases the old guard handled is unchanged.

The new test is a genuine regression guard — it scripts a throw that also transitions the socket to `Aborted` and would hang on the bounded `WaitSocketClosed` await under the old code — and the accompanying test-double changes (the `resultingState` overload, updated docs, removal of the now-inaccurate inline comment) are clean. CI is green across backend and all three e2e browsers.

One purely optional nit, non-blocking: the new test could also assert the *absence* of an error log to pin the `if (_context.State is Open)` false branch symmetrically with the still-Open test. Not worth tracking as an issue.

Merging.

---

PR #1539, opened 2026-07-03
Fix socket presence refresh unable to resurrect a lost key

### LINKED ISSUE #1497: Socket presence refresh can't resurrect a lost key (ExpireAndForget no-ops on missing keys) — pushes dropped, takeover/switch guards defeated
## Problem

`SocketManagerService.RefreshSocketPresence` (`Game.Api/Services/SocketManagerService.cs:139-142`) refreshes the presence claim with `_cache.ExpireAndForget(...)`, which is a **no-op on a missing key** (`RedisService.cs:81-84`, `KeyExpire`). Once the presence key is gone while the socket is still live, no inbound message ever re-creates it.

Two concrete ways it goes missing under a live socket:

1. **TTL lapse** — any &gt;30s inbound stall (e.g. background-tab timer throttling; the socket itself survives because the inactivity close is 60s) expires the key; every subsequent heartbeat no-ops forever.
2. **Rollback of a superseding registration** — `RegisterSocket`'s `GetSet` displaces the live old socket's claim; if a later registration step fails, the rollback `CompareAndDelete`s the *new* id, leaving the displaced-but-still-live old socket with no key (and no `SocketReplaced` push, which is only emitted on success).

## Consequences for the rest of that connection's life

- Every server push (`ChallengeCompleted`, `ProficiencyXpGained`) resolves no target socket and is dropped with only a warning.
- `Login/ActiveSession` reports no active session — the takeover warning lies.
- `SwitchPlayer`'s `HasActiveSocket` guard is defeated, so the switch-away credit runs concurrently with the live battle loop — the lost-update race the guard exists to prevent.
- The next handshake's `GetSet` returns null, so the still-live old socket is never told to close → **two concurrent live sockets for one player**.

## Suggested fix

Make the refresh a conditional re-claim rather than a bare expire — e.g. `SET NX` + expire, or a `CompareAndSet` keyed on the socket's own id — so a live socket can restore its claim without clobbering a newer registration. Test: expire the key manually, deliver a heartbeat, assert the claim is restored to the same socket id.

ginderjeremiah:
## Summary

`SocketManagerService.RefreshSocketPresence` refreshed the per-player presence claim with `ICacheService.ExpireAndForget`, a bare TTL bump that **no-ops on a missing key**. Once the presence key was gone while the socket was still live, no inbound heartbeat could ever recreate it:

- **TTL lapse** — any inbound stall past the 30s presence TTL (the socket itself survives up to the 60s inactivity close) expires the key; every later heartbeat no-ops forever.
- **Rollback of a superseding registration** — `RegisterSocket`'s `GetSet` displaces a live old socket's claim; if the new registration's later step fails, the rollback compare-and-deletes the *new* id, leaving the displaced-but-still-live old socket with no key at all.

With the key permanently gone, every server push (`ChallengeCompleted`, `ProficiencyXpGained`) silently dropped, `Login/ActiveSession` reported no active session, `SwitchPlayer`'s `HasActiveSocket` guard was defeated, and the next real handshake's `GetSet` returned null — leaving two concurrent live sockets for one player.

## Fix

Added `ICacheService.ReclaimAndForget(key, ownerValue, expiry)` — a fire-and-forget "SET NX + expire" in one Lua script:

- If the key is currently **unset**, claims it as `ownerValue` with a fresh TTL (the resurrection path `ExpireAndForget` lacked).
- If the key is **already set**, it just extends the TTL of whatever value is there — matching `ExpireAndForget`'s existing "don't-care-who-asked" refresh behavior, so a stale caller still can never overwrite a *different* (newer) claim's value.

`RefreshSocketPresence` now calls this with the socket's own id instead of blind-expiring, so a live socket can restore its own claim after either failure mode above, while a genuine takeover is still never clobbered.

## Test plan

- [x] New unit-level integration tests on `RedisServiceIntegrationTests` covering all three `ReclaimAndForget` branches (missing key claims, owned key refreshes TTL without changing value, someone-else's key is left untouched).
- [x] New integration test `SocketManagerServiceTests.RefreshSocketPresence_AfterItsKeyLapsesWhileStillLive_ResurrectsItsOwnClaim` — deletes the presence key under a live socket and asserts the next heartbeat restores it under the same socket id.
- [x] Existing `RefreshSocketPresence_FromConnectionThatWasTakenOver_ProlongsCurrentKeyWithoutResurrectingOldSocketId` still passes unchanged — a stale connection's heartbeat still can't resurrect its own id over a newer claim.
- [x] Full backend suite: `dotnet test --no-build --no-restore` — 2743 tests passing.

Closes #1497

---
_Generated by [Claude Code](https://claude.ai/code/session_01WXHscu11ZMdbMdxjuHk192)_

### DIFF
--- a/Game.Abstractions/Infrastructure/ICacheService.cs
+++ b/Game.Abstractions/Infrastructure/ICacheService.cs
@@ -56,5 +56,14 @@
         /// update is never silently lost. The optimistic-concurrency counterpart to <see cref="CompareAndDelete"/>.
         /// </summary>
         public Task<bool> CompareAndSet(string key, string? expectedValue, string newValue, TimeSpan expiry, CancellationToken cancellationToken = default);
+        /// <summary>
+        /// Fire-and-forget "SET NX + expire": if <paramref name="key"/> is currently unset, claims it as
+        /// <paramref name="ownerValue"/> with <paramref name="expiry"/> as its TTL; otherwise just extends the
+        /// TTL of whatever value is already there (matching <see cref="ExpireAndForget"/>'s existing
+        /// don't-care-who-asked refresh — it never overwrites a value that isn't <paramref name="ownerValue"/>).
+        /// Unlike <see cref="ExpireAndForget"/> (a bare TTL bump that no-ops on a missing key), this can
+        /// resurrect a claim that expired or was rolled back out from under a still-live owner.
+        /// </summary>
+        public void ReclaimAndForget(string key, string ownerValue, TimeSpan expiry);
     }
 }
--- a/Game.Api.Tests/Integration/SocketManagerServiceTests.cs
+++ b/Game.Api.Tests/Integration/SocketManagerServiceTests.cs
@@ -146,6 +146,35 @@ namespace Game.Api.Tests.Integration
         }
 
         [Fact]
+        public async Task RefreshSocketPresence_AfterItsKeyLapsesWhileStillLive_ResurrectsItsOwnClaim()
+        {
+            var (userId, playerId) = await SeedAndLoginAsync();
+            var key = PresenceKey(playerId);
+
+            await using var socketA = new TestSocketClient();
+            await socketA.ConnectAsync(Factory.Server.CreateWebSocketClient(), userId);
+            await socketA.SendCommandAsync<object>("GetStatisticTypes");
+            var idA = await ReadPresenceValueAsync(playerId);
+            Assert.NotNull(idA);
+
+            // Simulate the presence key lapsing out from under a still-live socket — an inbound stall past the
+            // 30s TTL, or a rolled-back superseding registration — by deleting it directly (#1497).
+            using (var scope = CreateScope())
+            {
+                var cache = scope.ServiceProvider.GetRequiredService<ICacheService>();
+                await cache.Delete(key);
+            }
+            Assert.Null(await ReadPresenceValueAsync(playerId));
+
+            // The next heartbeat on the still-live socket must resurrect its own claim rather than leaving the
+            // key permanently gone (a bare ExpireAndForget is a no-op on a missing key and never recovers).
+            await socketA.SendCommandAsync<object>("GetStatisticTypes");
+
+            Assert.True(await WaitUntilAsync(async () => await ReadPresenceValueAsync(playerId) == idA),
+                "Expected the live socket's next heartbeat to resurrect its own presence claim.");
+        }
+
+        [Fact]
         public async Task HasActiveSocket_AfterPresenceTtlExpires_ReturnsFalse()
         {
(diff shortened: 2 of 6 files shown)


ginderjeremiah (commented):
## Review: Resurrect a lapsed socket-presence key so heartbeats can restore a live socket's claim (closes #1497)

**Verdict:** Approve — correctly replaces the no-op `ExpireAndForget` refresh with a conditional re-claim that resurrects a lost key without ever clobbering a newer takeover; thoroughly tested and well-documented. (Submitted as `COMMENT` because I'm the PR author and GitHub won't accept a gating self-approval.)

### Correctness

The `ReclaimAndForget` Lua script is atomic (Redis runs it under a single lock), so the `exists`→`set`/`pexpire` sequence has no internal race:
- **Missing key** → claims it as `ownerValue` with a fresh TTL — the resurrection path `ExpireAndForget` structurally lacked.
- **Existing key (any value)** → extends the TTL only, never touching the value — preserving the exact "don't-care-who-asked" refresh `ExpireAndForget` already provided.

The no-clobber guarantee holds under the concurrent cases that matter, because `RefreshSocketPresence` only ever fires from a live socket's own read loop:
- A stale/being-replaced socket heartbeating during the takeover-close window hits the `else` branch (key holds the newer id) and can't steal it back.
- A resurrection racing a fresh `GetSet` registration converges either way — whichever runs first, the other is a benign TTL bump. The resurrected value is the socket's own `Guid` id, identical to what `GetSet` writes and what teardown's `CompareAndDelete` keys on, so the whole presence lifecycle stays consistent.

Argument types are right (`PX` in ms for both `set` and `pexpire`), and `FireAndForget` keeps it off the hot inbound path exactly as `ExpireAndForget` was. The `else`-branch behavior is unchanged from the prior implementation, so there's no regression for the already-tested takeover case.

### Test coverage

Strong and matches the project's rules for an out-of-process (Redis) dependency:
- `RedisServiceIntegrationTests` covers all three script branches — missing-key claim, owned-key TTL refresh without value change, and someone-else's-key left untouched.
- `SocketManagerServiceTests.RefreshSocketPresence_AfterItsKeyLapsesWhileStillLive_ResurrectsItsOwnClaim` proves the end-to-end resurrection through a real heartbeat.
- The existing takeover-safety test is preserved unchanged, guarding the no-clobber path.

### What's good

- The three-branch fix precisely matches the issue's suggested approach and the design is airtight against the takeover race the guard exists to protect.
- Comments and XML docs clearly explain *why* (resurrection vs. no-clobber) without over-documenting.
- Follows existing infra patterns cleanly — Lua mirroring `CompareAndDelete`/`CompareAndSet`, fire-and-forget mirroring `ExpireAndForget`, and both `NoOpCacheService` stubs updated consistently. `ExpireAndForget` remains in use by the repositories, so nothing is left dead.

---

Issue #1542, opened 2026-07-03
Extend the AdminTools Ops page to inspect/replay the socket command dead-letter queue


ginderjeremiah:
### Context

Follow-up to #1498. The socket push pipeline's dead-letter queue (`SocketCommandDeadLetterQueue`, written by `SocketManagerService.EscalateFailedServerCommand`) was write-only — nothing read it, unlike the player write-behind dead-letter queue, which got both depth alerting (`DataProviderSynchronizer`) and an admin Ops inspect/replay page (#794, `PlayerUpdateDeadLetters` / `IPlayerUpdateDeadLetters`).

#1498 added the lighter half of that gap: the escalation path now logs the queue's depth on every dead-letter (`SocketManagerService.EscalateFailedServerCommand`), so an accumulating backlog is at least visible in logs/alerting. It deliberately did **not** build the inspect/replay Ops surface — that's a distinct, non-trivial feature (new admin contract types, a new endpoint, a new classification helper, and a Workbench/Ops UI panel), out of scope for a hygiene bug-fix PR.

### Scope

Mirror `PlayerUpdateDeadLetters` (`Game.DataAccess/Repositories/Admin/PlayerUpdateDeadLetters.cs`) for the socket command queue:

- A non-destructive **peek** that classifies each entry (malformed / known-command / unknown-command) similar to `EDeadLetterReason`, using `SocketCommandInfo`'s shape (`Name`, `Id`, `Parameters`) rather than `DomainEventEnvelope`.
- A guarded **replay** that re-enqueues selected (or all) entries — note replay here needs to target a specific socket's live queue (`SocketQueue_{socketId}`) if the player has since reconnected, or simply re-dispatch immediately via `SocketManagerService.EmitSocketCommand`, since the socket id embedded in the original push context is not preserved on the dead-lettered payload today. Decide during implementation whether the payload needs a player/socket id added to make replay meaningful (today `SocketCommandInfo` carries no addressing info once dequeued).
- Wire it into the existing shared `/api/AdminTools/*` Ops surface alongside the player dead-letter page.

### Acceptance

- An operator can inspect the socket command dead-letter queue's entries from the admin Workbench.
- An operator can replay a selected (or all) entries without needing direct Redis access.
- Unit tests cover the classification and replay logic; integration tests cover the admin endpoint(s) against the real queue.

Issue #794, opened 2026-06-16
Add a guarded inspection/replay consumer for the player write-behind dead-letter queue


ginderjeremiah:
## Background

Follow-up to #769, which made the player write-behind queue read at-least-once (reserve → process → acknowledge, with startup reclaim of orphaned in-flight items) and started **surfacing the dead-letter queue's depth** so poison messages no longer pile up invisibly — a warning is logged whenever the dead-letter depth grows past the last reported value (and on the first drain after startup, catching a standing backlog).

That covered the *minimum* the parent issue asked for. The **optional** part — a guarded inspection/replay path — was deliberately deferred to keep #769 focused.

## Problem

`PUBSUB_PLAYER_DEAD_LETTER_QUEUE` is still a passive holding list: nothing reads it. Once a message is dead-lettered (malformed payload, unknown event type, or retries exhausted), the only recourse is manual Redis surgery. There is no surface to:

- **inspect** what is sitting in the dead-letter queue (type, player, raw payload, why it failed), or
- **replay** a message back onto the main queue once the underlying cause is fixed (e.g. a transient outage that outlasted the retry budget, or a handler bug that has since been patched).

## Suggested direction

- A small **admin-only** inspection surface (count + peek of the dead-letter entries) — read-only, no destructive pop. Note that a destructive pop for the dead-letter read would have the same at-most-once crash window #769 fixed for the main queue, so a replay path should reserve/acknowledge (or otherwise move atomically) rather than `LPOP`.
- A **guarded replay** action that moves selected (or all) dead-letter entries back onto the player update queue and wakes the channel, relying on the existing idempotent handlers. Replaying a genuinely-poison message (malformed / unknown type) will just re-fail and re-dead-letter, so the surface should make the failure reason visible to avoid pointless replays.
- Decide where this lives — most naturally the existing admin tooling (Content/Ops), consistent with how the challenge-completion server-push dead-letter path is described in `docs/backend-persistence.md`.

### Labels
claude, enhancement, scope: medium

Issue #671, opened 2026-06-14
Escalate persistently-failing server-initiated socket commands (dead-letter / bounded retry)


ginderjeremiah:
Parent: #656

## Context

#656 established and documented the *current* contract for server-initiated (pub/sub) socket command failures: `SocketManagerService.GetSocketCommandProcessor` logs a failure and keeps draining the queue, distinguishing a lifetime `OperationCanceledException` (logged at debug as a teardown) from a genuine fault (logged at error). It deliberately did **not** add client surfacing, dead-lettering, or bounded retry — that escalation work is this issue.

The gap: a genuinely persistent fault (e.g. a poisoned `ChallengeCompleted` payload that always throws) is silently re-logged on every redelivery with no escalation, and the affected client never learns a server-pushed command failed. This is inconsistent with the "process the rest, but escalate the stuck one" resilience the write-behind consumer (`DataProviderSynchronizer` retry-then-dead-letter) already follows.

## Options to evaluate

1. **Surface-to-client** — push a typed error/ack back to the affected socket so the client can react (re-fetch authoritative state) instead of silently diverging. Note the push commands (`ChallengeCompleted`, `SocketReplaced`) are server-initiated only and carry no client request type today, so this needs a one-way "server error notice" channel.
2. **Bounded retry** — retry the failed push a small number of times (with backoff) before giving up, for transient faults.
3. **Dead-letter** — after N failures, move the poisoned payload to a dead-letter queue (mirroring `DataProviderSynchronizer`) so it stops re-logging on every redelivery and can be inspected, rather than looping forever.

These are not mutually exclusive: transient vs. persistent faults likely want different treatment (retry the former, dead-letter the latter; surface either to the client).

## Important secondary finding (from #656 investigation)

While implementing #656 I empirically confirmed that the processor's `catch` is essentially **unreachable for command-execution faults**: `SocketHandler.ExecuteCommand` has its own outer `catch (Exception)` that absorbs every command fault — *including a non-timeout `OperationCanceledException`* — turning it into an `"Internal Server Error"` response and logging it at error there. So:

- For a server push, that error response goes nowhere (no client awaits it), and the fault is logged inside `ExecuteCommand`, not the processor.
- A **lifetime cancellation is mislabeled** as `"Internal Server Error"` inside `ExecuteCommand` (its `catch (OperationCanceledException) when (cts.IsCancellationRequested)` guard only matches the per-command *timeout*, so a non-timeout cancel falls through to the generic catch).

[...]

PR #1543, opened 2026-07-03
Fix socket push pipeline hygiene gaps

### LINKED ISSUE #1498: Socket push pipeline hygiene: orphaned queue keys, write-only dead-letter queue, undelivered sends reported as Succeeded, malformed-frame misclassification
A basket of related gaps in the server-push/socket-command pipeline, none individually large but together eroding the documented "never silently drop" contract. All verified by reading the code end-to-end.

## 1. Per-socket command queues are never cleaned up — orphaned TTL-less Redis keys accumulate forever

`EmitSocketCommand(playerId)` RPUSHes onto `SocketQueue_{socketId}` (`Game.Api/Services/SocketManagerService.cs:152-168`). `TeardownSocketRegistration` (`:96-118`) deletes the presence key but never the queue key, and queue keys carry **no TTL**. Any push that races a disconnect (presence key not yet released/expired), lands after `UnSubscribe`, or arrives after the processor abandoned its loop on 5 consecutive dequeue failures strands a permanent key. Over months of pushes racing disconnects this grows without bound. Fix: delete the queue key in teardown, and/or EXPIRE queue keys on push.

## 2. The socket dead-letter queue is write-only

`EscalateFailedServerCommand` writes `SocketCommandDeadLetterQueue` (`SocketManagerService.cs:256-266`), but nothing anywhere reads that key — no depth warning, no Ops surface. Contrast the player write-behind DLQ, which got both depth alerting (`DataProviderSynchronizer.cs:227-236`) and an admin Ops inspect/replay page (#794). Poisoned server pushes pile up invisibly forever. Either surface it (depth log + extend the AdminTools Ops page) or give it a retention TTL and document it as best-effort.

## 3. A failed WebSocket send is reported as `Succeeded`

`SocketContext.SendData` returns `false` on a non-Open socket or send failure, but `RunCommandUnderLock` (`Game.Api/Sockets/SocketHandler.cs:193`) ignores the result and reports `Succeeded`. A server push executed just as the socket closes is silently undelivered — no dead-letter, no log — and `EscalateFailedServerCommand`'s notice-outcome check, whose comment explicitly claims to catch "the socket just closed", can never observe that case. Impact is softened by full re-sync on reconnect, but the outcome should carry the send result (e.g. a `NotDelivered` outcome) or at least log it.

## 4. Malformed client frames are misclassified

The handler gives structured rejections for absent/unknown/server-only command names, but (a) a JSON-invalid frame gets only a log — the client waits out its 30s request timeout with no error frame (`SocketHandler.cs:393-396`); (b) malformed `Parameters` (a `SetParameters` throw inside `RunCommand`) is classified as a genuine fault → client-controlled error-level logs + `"Internal Server Error"` instead of a "Malformed parameters" rejection (`AbstractSocketCommand.cs:43-56`, `SocketCommandFactory.cs:39-52`).

Related history: #656/#671 established the escalation path this basket hardens the edges of; #909 covered registration/teardown ordering.

ginderjeremiah:
## Summary

Addresses the four related gaps in the server-push/socket-command pipeline described in #1498:

1. **Orphaned per-socket command queues.** `SocketQueue_{socketId}` (a plain Redis list) never carried a TTL and was never deleted on teardown, so a push racing a disconnect or arriving after the processor abandoned its drain loop could strand a permanent key.
   - `TeardownSocketRegistration` now deletes the socket's queue key outright (best-effort, guarded like the other teardown steps) — the common graceful-disconnect path frees it immediately.
   - `EmitSocketCommand` now refreshes a backstop `SocketQueueTtl` (1h) on every push, so any path that skips teardown (a crash, a hard disconnect) still bounds the key's lifetime rather than leaving it permanent.

2. **Write-only socket dead-letter queue.** `EscalateFailedServerCommand` now logs the queue's depth on every escalation, giving operators visibility into an accumulating backlog. A full inspect/replay admin Ops surface (mirroring the player write-behind DLQ's page from #794) is a distinct, non-trivial feature — filed as a follow-up: #1542.

3. **A failed WebSocket send reported as `Succeeded`.** `SocketContext.SendData`'s return value was previously ignored after a successful command execution. `SocketCommandOutcome` gains a `NotDelivered` case; both `ExecuteCommand` and `ExecuteServerCommand` now log it instead of silently treating an undelivered response as success.

4. **Malformed frames misclassified.**
   - A JSON-invalid inbound frame carries no request id to correlate a structured rejection to, so it previously left the client to wait out its 30s request timeout undiagnosed. It now closes the socket (`ESocketCloseReason.MalformedFrame` → `WebSocketCloseStatus.InvalidPayloadData`), mirroring the existing oversized-frame path — the client's close handler surfaces the drop and reconnects promptly.
   - A command with malformed/missing `Parameters` (a `SetParameters` throw) previously fell into the generic fault path (error-level log + "Internal Server Error"). It's now classified as `SocketCommandOutcome.MalformedParameters` — a bad request, not a server fault — and rejected with `"Malformed parameters."` at warning level for the client-invoked path. A malformed *server-initiated* push (server-authored, so a genuine bug) still escalates the same as a fault.

## Notes on the issue

Two follow-up history references (#656, #671, #909) that #1498 says "this basket hardens the edges of" are already merged/closed — confirmed before starting, so nothing here was blocked.

## Test plan

- [x] `dotnet build Game.sln`
- [x] `dotnet test Game.sln` — full backend suite (2758 tests) passes, including new/updated coverage for:
  - queue-key deletion on teardown and TTL refresh on push (`SocketCommandProcessorTests`)
  - `NotDelivered` / `MalformedParameters` outcome classification on both `ExecuteCommand` and `ExecuteServerCommand` (`SocketCommandExecutionTests`)
  - the malformed-frame close path (`SocketCommandExecutionTests`, `SocketCloseReasonTests`)
- [x] Updated `docs/backend.md` → "Challenge-completion notifications (server push)" to reflect the new fault classification and DLQ depth logging.

Closes #1498

---
_Generated by [Claude Code](https://claude.ai/code/session_01U9HbFifcAL3rvDNdpWp23g)_


Game.Api/Sockets/SocketHandler.cs
```diff
@@ -227,7 +257,20 @@ internal async Task<SocketCommandOutcome> ExecuteServerCommand(SocketCommandInfo
...
+            AbstractSocketCommand command;
+            try
+            {
+                // CreateCommand binds Parameters (SetParameters), which throws on malformed/missing JSON. That
+                // is a bad request, not a server fault, so it is wrapped and classified distinctly rather than
+                // falling into the generic fault path below (which would log at error and reply "Internal
+                // Server Error").
+                command = _commandFactory.CreateCommand(commandInfo, scope);
+            }
+            catch (Exception ex) when (ex is JsonException or ArgumentNullException)
```
ginderjeremiah: Nit: the `try` wraps the whole `CreateCommand` call — DI construction (`generator(...)`) **and** the `SetParameters` binding — but the `catch` classifies *any* `JsonException`/`ArgumentNullException` as `MalformedParameters` (a client bad-request). Only the `SetParameters` deserialize is actually a bad-request; a `JsonException`/`ArgumentNullException` thrown while constructing the command (e.g. an injected dependency's ctor) would be misclassified as client input error and warn-logged instead of surfacing as a genuine fault.

The realistic surface today is `SetParameters` (commands are thin, deps injected), so this is near-zero risk and non-blocking. But it reads as a heuristic on exception *type* rather than on *where* it came from. Cleaner options: narrow the `try` to just the `command.SetParameters(...)` call, or throw `MalformedSocketCommandParametersException` directly from the base `SetParameters` overrides (`AbstractSocketCommand.cs:43-56`), where the malformed-params semantics are actually known — that removes the guess entirely. Take it or leave it.

---

ginderjeremiah (commented):
## Review: Socket push pipeline hygiene gaps (#1498)

**Verdict:** Approve *(posted as a comment — I authored the PR, so GitHub won't take a gating vote from me)* — all four gaps are addressed correctly and the change is well-tested; the one nit below is non-blocking.

This is a tidy, well-scoped fix for the four related gaps in #1498. I verified the correctness-critical assumptions rather than taking the description on faith:

- **Orphaned queue keys** — `TeardownSocketRegistration` now deletes the per-socket queue key (plain delete, correctly justified vs. the shared presence key's compare-and-delete), and `EmitSocketCommand` refreshes a 1h backstop TTL on every push. Nice DRY touch folding the `playerId` overload into the `socketId` one so both paths get the TTL.
- **Write-only DLQ** — depth is now logged on every escalation; the full inspect/replay Ops surface is correctly split out as follow-up #1542 rather than scope-crept here.
- **Undelivered send reported as `Succeeded`** — `SendData`'s `bool` result is now threaded into a distinct `NotDelivered` outcome (confirmed `SendData` returns `Task<bool>` and returns `false` on a non-Open socket), logged but not escalated, which is the right call since the payload itself ran fine.
- **Malformed frames** — unparseable JSON now closes with `MalformedFrame` → `InvalidPayloadData`, mirroring the oversized-frame path; malformed `Parameters` is classified as `MalformedParameters` and rejected as a bad request. Confirmed the `JsonException or ArgumentNullException` catch filter matches what `SetParameters`/`Deserialize` actually throw (null → `ArgumentNullException`, bad JSON → `JsonException`).

### Nits
- `SocketHandler.cs:261-272` — the exception-classification `catch` wraps the whole `CreateCommand` (DI construction + binding) but only `SetParameters` is genuinely a bad-request; a `JsonException`/`ArgumentNullException` from construction would be misclassified. Near-zero risk today, but throwing the typed exception from `SetParameters` itself would make it precise. Inline detail above.

### Test coverage
Thorough and matches the project's rules. Both `ExecuteCommand` and `ExecuteServerCommand` paths are covered for each new outcome (`MalformedParameters`, `NotDelivered`), the malformed-frame close path is tested, and `SocketCloseReasonTests` extends every parameterized case to `MalformedFrame`. `SocketCommandProcessorTests` covers queue-key deletion on teardown, the TTL refresh on emit, and the DLQ depth read. No frontend/battle surface touched, so no parity mirror needed.

### What's good
Comment quality is high and the outcome taxonomy (`Succeeded`/`NotDelivered`/`TimedOut`/`Faulted`/`MalformedParameters`/`TornDown`) is a clean way to keep each caller's fault policy explicit. `docs/backend.md` is updated to match. Solid work.

*Note: CI (`test-backend`, `test-e2e (webkit)`) was still running at review time — holding the merge on green + the nit.*

---

ginderjeremiah:
## Review: Fix socket push pipeline hygiene gaps

**Verdict: Approve (with minor nits).** CI is green (backend, e2e all pass; frontend jobs correctly skipped). All four gaps from #1498 are addressed cleanly, the classification logic is correct, and coverage is thorough. Nits below are optional and non-blocking.

### Strengths

- **Correct classification at the right seam.** Moving the malformed-params detection into `AbstractSocketCommand.DeserializeParameters` — throwing a typed `MalformedSocketCommandParametersException` *at the deserialize site* rather than having `SocketHandler` guess from a caught `JsonException`/`ArgumentNullException` — is the right call. It can't misclassify an unrelated `JsonException` thrown during a command's DI construction, and the new `AbstractSocketCommandTests` pins exactly that intent.
- **`NotDelivered` is correctly wired.** `SocketContext.SendData` already returned `false` on a non-Open socket / send failure / pre-frame cancellation; threading that through as `delivered ? Succeeded : NotDelivered` closes the "reported as Succeeded" gap and, as a bonus, makes `EscalateFailedServerCommand`'s notice-outcome check (`noticeOutcome is not Succeeded`) actually observable for the first time.
- **Server-vs-client fault policy is right.** A client bad-request (`MalformedParameters`) is warned + rejected with `"Malformed parameters."`; a *server-initiated* malformed push is logged at Error and escalated the same as a fault (payload is server-authored → genuine bug). `NotDelivered` is deliberately not escalated (dead-lettering a fine payload wouldn't help; re-sync on reconnect covers it) — consistent with the documented contract.
- **Queue-key hygiene is sound and bounded.** Graceful teardown `Delete`s the key outright (safe as a plain delete since a per-socket queue is never contended, unlike the compare-and-delete presence key); the 1h `SocketQueueTtl` refreshed on every push is a genuine backstop for crash/hard-disconnect paths. A push racing a disconnect re-creates the key, but the TTL bounds it rather than stranding it permanently — no re-strand race. Nice DRY refactor folding the `playerId` overload through the `socketId` one so both get the refresh.
- **Malformed-frame close mirrors the oversized-frame path** (`Close(MalformedFrame)` → `InvalidPayloadData`), so a JSON-invalid frame with no correlatable request id surfaces to the client's close handler instead of stranding it for the 30s request timeout.
- Docs updated, follow-up #1542 filed for the DLQ inspect/replay Ops surface (correctly scoped out), and the `#656/#671/#909` history references were verified as already-merged before starting. Good diligence.

1. **Depth read is coupled into the dead-letter `try`.** In `EscalateFailedServerCommand`, `GetLengthAsync()` runs inside the same `try` as `AddToQueueAsync()`. If the enqueue succeeds but the (purely informational) depth read then hits a transient Redis blip, the `catch` logs `"Failed to dead-letter a faulted server-initiated command"` at Error — misreporting a dead-letter that actually succeeded. Consider reading the depth in its own guarded step (or `try`/swallow), so an informational read can't masquerade as an escalation failure.
2. **No assertion on the logged depth value.** `RecordingCacheService`/the queue stub now returns a length, and a test asserts `"Dead-lettered"` is logged, but nothing asserts the depth actually appears in the escalation message. A one-line assertion would pin the new `{Depth}` field against silent regression.
3. **Last-push TTL residual (accept-as-is).** `ExpireAndForget` is fire-and-forget, so a hard crash immediately after a `Publish` whose `EXPIRE` was dropped could still leave a TTL-less key for the *final* push to a dead socket. This is the narrow best-effort gap the issue explicitly allowed ("and/or EXPIRE queue keys on push"), and graceful teardown's `Delete` covers the common path — noting only for completeness, not asking for a change.

Clean, well-documented, well-tested change that tightens the "never silently drop" contract. Good to merge once the nits are considered.

---

PR #1544, opened 2026-07-03
Add up-front reference validation to AdminChallenges.SaveChallenges

### LINKED ISSUE #1499: AdminChallenges.SaveChallenges skips the up-front reference validation every sibling admin save has — bad FKs surface as 500s
## Problem

`AdminChallenges.SaveChallenges` (`Game.DataAccess/Repositories/Admin/AdminChallenges.cs:22-56`) persists challenges without validating their references first. The Challenge entity carries real FKs — `ChallengeTypeId` → ChallengeTypes, `RewardItemId` → Items, `RewardItemModId` → ItemMods — so a bad reward id or an unmapped type id surfaces as an FK `PostgresException` 500 at commit. That is exactly the failure class `ReferenceFieldValidation` and the per-repo guards convert to clean `AdminSaveResult.Failure` 400s everywhere else (`AdminItems` validates rewards/gates/enum FKs; `AdminZones` validates boss/unlock refs; `AdminSkills`/`AdminItemMods` validate enum FKs).

`TargetEntityId` is likewise unvalidated at save time — only the post-hoc content lint / Content Health view flags e.g. a `KillsByDamageType` challenge with no target (#1455's checker covers committed content, not the save).

## Suggested fix

Add the established guards to the challenge save, matching the sibling pattern (anti-tamper — a frontend picker filter alone is insufficient):

- `FindUndefinedEnum(ChallengeTypeId)`
- reward item / item-mod existence (non-retired for new references)
- target-dimension validation appropriate to the challenge type (e.g. a damage-type-scoped challenge's target must be a defined `EDamageTypeKey`)

Plus unit tests mirroring the sibling repos' validation coverage.

ginderjeremiah:
## Summary

`AdminChallenges.SaveChallenges` (`Game.DataAccess/Repositories/Admin/AdminChallenges.cs`) persisted challenges without validating their references first, so a bad `ChallengeTypeId`, `RewardItemId`/`RewardItemModId`, or `TargetEntityId` would 500 on an FK `PostgresException` at commit instead of surfacing a clean `AdminSaveResult.Failure` 400 — the guard every sibling admin save (`AdminItems`, `AdminItemMods`, `AdminZones`) already has.

This adds three up-front guards, mirroring the established sibling pattern:

1. **`ChallengeTypeId` enum guard** — `ReferenceFieldValidation.FindUndefinedEnum`, same as the item rarity/category/mod-type checks.
2. **Reward existence + retirement guard** — a reward that *newly* references an item/item-mod (an Add, or an Edit that actually changes the reward) must exist and be non-retired. An edit's **unchanged** reward is left alone: per `docs/backend.md`, a reward resolves by id forever, so a reward retired *after* being authored must not permanently block later, unrelated edits to that challenge (mirrors the content-graph lint, which treats a retired-but-still-referenced reward as a `Warning`, not an `Error`).
3. **Target-dimension guard** — `TargetEntityId` must resolve against the dimension the challenge type derives (enemy/zone/skill existence for those types, or a defined `EDamageTypeKey` for `KillsByDamageType`), and a `KillsByDamageType` challenge must carry a target at all (it can never track progress otherwise — same rule the content-graph lint already enforces post-hoc). `None`-dimension types (e.g. `LevelReached`) carry no target to validate.

### Note on the issue

The issue's suggested fix said the reward check should be "non-retired for new references." I read that as requiring the diff-against-previous-state behavior in guard (2) above (see `AdminPaths.SavePaths`'s `FindRetiredPathGatingLiveGateway` for existing precedent of this "only guard the transition" pattern) rather than a blanket retirement check on every save — a blanket check would make it impossible to ever edit a challenge again once its reward is retired, which contradicts the lenient design already encoded in the content-graph lint. Flagging this interpretation in case it's not what was intended.

Target existence checks intentionally do **not** reject a *retired* target (only a nonexistent one) — same reasoning, and consistent with how `AdminItems`/`AdminZones` validate their own record-set-specific references (existence only, no retirement check, except where retirement itself would strand live content).

## Test plan

- [x] `dotnet build Game.sln`
- [x] `dotnet test Game.sln` — full backend suite (2758 tests) passes, including new coverage in `AdminChallengesIntegrationTests`:
  - undefined `ChallengeTypeId`
  - missing / retired reward item and item-mod (as a new reference)
  - an edit that keeps an already-retired reward unchanged still succeeds
  - `KillsByDamageType` with no target / an undefined damage-type key
  - missing enemy/zone/skill target
  - a valid target + reward round-trip

Closes #1499

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vv7PSBR3akFD7LicPJk1Uk)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminChallengesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminChallengesIntegrationTests.cs
@@ -212,6 +212,32 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SaveChallenges_NewRewardItemModRetired_ReturnsFailure()
+        {
+            int itemModId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var itemMod = await TestDataSeeder.CreateItemModAsync(context, name: "Old Prefix");
+                itemMod.RetiredAt = new DateTime(2026, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+                await context.SaveChangesAsync(CancellationToken);
+                itemModId = itemMod.Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminChallenges>();
+
+            var result = admin.SaveChallenges(
+            [
+                new Change<Contracts.Challenge> { ChangeType = EChangeType.Add, Item = NewChallenge(name: "Retired Reward Mod", rewardItemModId: itemModId) },
+            ]);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal("Reward item mod 'Old Prefix' is retired and cannot be newly assigned as a challenge reward.", result.ErrorMessage);
+        }
+
+        [Fact]
         public async Task SaveChallenges_EditKeepingRetiredRewardUnchanged_Succeeds()
         {
             // A reward resolves by id forever: an item retired after being authored as a challenge's reward
@@ -337,6 +363,65 @@ namespace Game.Application.Tests.DataAccess
 
             Assert.False(result.Succeeded);
             Assert.Equal("Target skill 99999 does not exist.", result.ErrorMessage);
+        }
+
+        [Fact]
+        public async Task SaveChallenges_TargetingRetiredEnemy_Succeeds()
+        {
+            // Unlike a reward, a retired target is tolerated: it can't fault the runtime the way a dangling id
+            // would, and only risks an eventually-uncompletable challenge, which the content-graph lint already
+            // flags post-hoc as a warning rather than a save-time rejection.
+            int enemyId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var enemy = await TestDataSeeder.CreateEnemyAsync(context, name: "Ancient Dragon");
+                enemy.RetiredAt = new DateTime(2026, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+                await context.SaveChangesAsync(CancellationToken);
+                enemyId = enemy.Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            var changes = new List<Change<Contracts.Challenge>>
+            {
+                new()
+                {
+                    ChangeType = EChangeType.Add,
+                    Item = NewChallenge(name: "Dragon Slayer", challengeTypeId: EChallengeType.EnemiesKilled, targetEntityId: enemyId),
+                },
+            };
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminChallenges>();
+            Assert.True(admin.SaveChallenges(changes).Succeeded);
+        }
+
+        [Fact]
+        public async Task SaveChallenges_ValidDamageTypeTarget_PersistsTargetEntityId()
+        {
+            var changes = new List<Change<Contracts.Challenge>>
+            {
+                new()
+                {
+                    ChangeType = EChangeType.Add,
+                    Item = NewChallenge(name: "Fire Hunter", challengeTypeId: EChallengeType.KillsByDamageType,
+                        targetEntityId: (int)EDamageTypeKey.Fire),
+                },
+            };
+
+            using (var writeScope = CreateScope())
+            {
+                var admin = writeScope.ServiceProvider.GetRequiredService<IAdminChallenges>();
+                Assert.True(admin.SaveChallenges(changes).Succeeded);
+                await writeScope.ServiceProvider.GetRequiredService<IUnitOfWork>().CommitAsync();
+            }
+
+            using (var assertScope = CreateScope())
+            {
+                var context = assertScope.ServiceProvider.GetRequiredService<GameContext>();
+                var created = await context.Challenges.AsNoTracking().SingleAsync(c => c.Name == "Fire Hunter", CancellationToken);
+                Assert.Equal((int)EDamageTypeKey.Fire, created.TargetEntityId);
+            }
         }
 
         [Fact]
--- a/Game.Application.Tests/DataAccess/AdminChallengesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminChallengesIntegrationTests.cs
@@ -140,15 +140,259 @@ namespace Game.Application.Tests.DataAccess
             Assert.Equal("The submitted challenge change set contains duplicate entries.", result.ErrorMessage);
         }
 
+        [Fact]
+        public void SaveChallenges_UndefinedChallengeType_ReturnsFailure()
+        {
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminChallenges>();
+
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: up-front reference validation for `AdminChallenges.SaveChallenges` (#1499)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating vote from me) — the change is correct, faithfully mirrors the sibling admin-save pattern, and is well tested. Only minor nits below.

I verified the core logic rather than trusting the description:

- **Dimension derivation matches the lint.** `new ChallengeType(item.ChallengeTypeId).StatisticType?.EntityType ?? None` resolves to the same `EEntityType` the content-graph lint reads off `challenge.EntityType` (confirmed via `ChallengeType.GetStatisticType` → `StatisticType.GetEntityType`, e.g. `KillsByDamageType → DamageType`, `EnemiesKilled/BattlesWon/TimeTrial → Enemy`, `ZonesCleared → Zone`, `SkillsUsed/DamageDealt → Skill`). Deriving from the type id rather than the contract's `EntityType` projection is the right anti-tamper choice and consistent with the class's own doc comment.
- **`GetChallenge` is correctly guarded.** It throws `ArgumentOutOfRangeException` on an out-of-range id, so gating the `previous` lookup behind `ValidateChallengeId` in `FindRewardViolation` is necessary, not defensive noise.
- **Retirement semantics line up with the caches.** The entity snapshots include retired records (the retired-reward test proves `LookupItem` returns a retired item with `RetiredAt` set), so target existence checks tolerate retirement (return non-null) exactly as the doc comment claims, and the reward retirement check can actually observe `RetiredAt`.
- **DI resolves.** All five injected caches (`IItemEntityCache`, `IItemModEntityCache`, `IEnemyEntityCache`, `IZoneEntityCache`, `ISkillEntityCache`) are registered in `ServiceCollectionExtensions`.
- **Diff-against-previous reward logic is sound** for Add (previous null → validated), unchanged Edit (skipped), reward added/changed on Edit (validated), and reward cleared (skipped).

### Nits

- **Reward retirement vs. target retirement is asymmetric.** A *newly assigned* retired reward is rejected, but a *newly assigned* retired target (e.g. an enemy retired-but-still-in-cache) is accepted — arguably the worse authoring mistake, since it yields an uncompletable challenge. The asymmetry follows the issue text and the lint's existence-vs-retirement split, so it's defensible; flagging it only so it's a conscious call rather than an accident. A one-line note on `FindTargetViolation` explaining why targets tolerate retirement while rewards don't would prevent a future "make these consistent" edit.
- `Enum.IsDefined(typeof(EDamageTypeKey), targetId)` uses the non-generic overload to match the lint verbatim; `ReferenceFieldValidation` elsewhere uses the generic `Enum.IsDefined(value)`. Harmless, just an inconsistency in style.

### Test coverage

Strong and mirrors the sibling repos' coverage well: undefined type, missing reward item/mod, newly-retired reward item, the key "edit keeping an already-retired reward unchanged still succeeds" nuance, `KillsByDamageType` with no target / undefined target, missing enemy/zone/skill targets, and a valid target+reward round-trip. Small gaps worth a follow-up test or two (not blocking):

- No test for a newly-retired reward *item-mod* (only the item path is covered, though the code is symmetric).
- Nothing asserts the *positive* side of the tolerated-retired-target claim — i.e. that targeting a retired-but-existing enemy/zone/skill is accepted. Without it, a future tightening of target validation could silently regress the documented intent.
- The damage-type happy path (a valid `EDamageTypeKey` target) isn't exercised on save; the round-trip test uses `EnemiesKilled`.

### What's good

Clean, DRY, and consistent with the established sibling saves — the `Find…Violation` helpers, up-front rejection ordering, and fresh-entity writes all match the house pattern. The doc comments explaining *why* only changed rewards are checked and why targets tolerate retirement are genuinely useful. The PR description also correctly flags its interpretation of the issue's "non-retired for new references" wording, which matches what I verified in the code.

---

ginderjeremiah:
## Review: Add up-front reference validation to `AdminChallenges.SaveChallenges`

**Verdict: Approve (with minor nits).** This is a clean, faithful implementation of the guard pattern the issue asked for. CI is green on `c77d97e`.

### What was verified

- **Mirrors the sibling pattern.** The three guards line up with `AdminItems` / `AdminZones` / `AdminItemMods`: `ReferenceFieldValidation.FindUndefinedEnum` for the `ChallengeTypeId` enum FK, and existence+retirement guards shaped like `AdminItems.FindProficiencyGateViolation`. Up-front rejection before anything is staged, returning `AdminSaveResult.Failure` (400) instead of an FK `PostgresException` (500). Correct.
- **Reward "guard only the transition" logic is sound.** Diffing the changed reward against `_challenges.GetChallenge(item.Id)` and skipping an unchanged reward matches the documented "a reward resolves by id forever" rule and the precedent in `AdminPaths.FindRetiredPathGatingLiveGateway`. An Add always validates (previous is null); an Edit that adds a second reward field while keeping the first validates only the new one. No duplicate-key throw hazard here because the guard iterates rather than building a per-id dictionary (the processor still rejects the duplicate batch afterward).
- **Target-dimension guard covers all challenge types.** Deriving the dimension from `ChallengeType(...).StatisticType?.EntityType` reaches every reachable `EEntityType` (Enemy/Zone/Skill/DamageType, with None correctly carrying no target). All 9 `EChallengeType` members are mapped in `GetStatisticType`, so `new ChallengeType(...)` can't throw for a defined type, and `FindUndefinedEnum` rejects undefined values first — no unhandled-throw path.
- **The `KillsByDamageType`-must-have-a-target rule and the `EDamageTypeKey` membership check mirror `ProgressionGraphChecker.CheckChallenges` exactly.** `EDamageTypeKey` is int-backed, so `(EDamageTypeKey)targetId` can't wrap a large tampered value into a valid key.
- **Retired-inclusive lookups confirmed.** The retirement tests pass, proving the injected entity caches resolve retired items/enemies (existence check sees the retired row), consistent with the reward-rejects-retired / target-tolerates-retired split.
- No null-forgiving operators; domain isolation intact (entity caches stay inside the data tier, no entity leaks up); test coverage is comprehensive (undefined type, missing/retired reward item + mod, unchanged-retired-reward edit succeeds, KillsByDamageType no-target / bad key, missing enemy/zone/skill target, retired-target succeeds, and two persistence round-trips).

### Nits / optional follow-ups (non-blocking)

1. **`BossesDefeated` boss-only target isn't enforced at save.** `StatisticType.BossOnly` is true for `BossesDefeated`, and the editor's target picker restricts to bosses, but the guard only checks enemy *existence*, so a tampered client could author a `BossesDefeated` challenge targeting a non-boss enemy — an eventually-uncompletable challenge. This is consistent with the PR's stated tolerance for uncompletable-but-not-runtime-faulting targets, and `CheckChallenges` doesn't catch it either, so it's out of scope here. Worth a small follow-up to tighten *both* the save guard and the lint to reject a non-boss target for a boss-only type.
2. Purely cosmetic: `FindTargetViolation` uses the generic `Enum.IsDefined((EDamageTypeKey)targetId)` while the lint uses the `typeof` overload — both correct; the generic form is fine to keep.

The interpretation flagged in the PR description (transition-diff rather than a blanket retirement check) is the right call and matches the content-graph lint's lenient Warning-vs-Error treatment of a retired-but-referenced reward.

Recommend merge once an approving review satisfies branch protection (`mergeable_state` is currently `blocked` on required review, not on CI).

---

ginderjeremiah:
Merged. The optional boss-only-target tightening noted in the review is tracked as follow-up #1547.

---

PR #1545, opened 2026-07-03
Fix Skills screen's incomplete battle-attribute composition

### LINKED ISSUE #1500: Skills screen computes battle numbers from an incomplete attribute set (no class locked base, proficiency bonuses, or signature passive)
## Problem

`skills-view.svelte.ts:189-191` builds the screen's attribute set as:

```ts
new BattleAttributes([...playerManager.attributes, ...inventoryManager.equipmentStats], true)
```

with a comment claiming it is "the same composition the battle uses". It no longer is: the live battler (`$lib/engine/battle/battle-engine.ts:204-257`) additionally composes the **class locked base** (`playerManager.battleLockedBaseModifiers`), the **proficiency bonuses** (`playerProficiencies.battleModifiers`), and the **class signature passive**. Since every character has a class, the screen's raw damage, DPS/burst, cooldown, crit numbers, and the Compare-vs mitigation results all understate the live battle — and visibly disagree with the in-fight SkillTooltip (which reads the live battler). The attribute-breakdown screen was updated when the class (#1126) and proficiency (#982) systems landed; this screen wasn't. The unit tests mock only attributes+equipmentStats, pinning the stale composition.

This breaks the documented frontend.md invariant that display surfaces consume the same composition the simulation uses — the *formulas* are shared, but the attribute *inputs* drifted.

## Root cause / fix direction

There are now **three hand-rolled copies** of the player-battler input composition (battle engine, attribute-breakdown screen, skills screen), and the third rotted. Extract one shared "assemble player battle attribute modifiers" builder in `$lib` and have all three consume it, so a fourth copy can't drift the same way. Update the skills-view tests to cover the full composition.

## Related smaller gap on the same screen

`metricsById` (`skills-view.svelte.ts:181,204-219`) is built only over **unlocked** skills, but innate item-granted skills are deliberately not unlocked — so the typical innate card (`InnateBand.svelte:22-23`) silently omits the dmg/cd/dps block the band is documented to share. (The only stats test uses a granted skill that happens to also be unlocked.) Include granted skill ids when building metrics.

ginderjeremiah:
## Summary

`skills-view.svelte.ts:189-191` built the screen's attribute set as `new BattleAttributes([...playerManager.attributes, ...inventoryManager.equipmentStats], true)`, with a comment claiming it's "the same composition the battle uses." It no longer was: the live battler (`battle-engine.ts`'s `resetPlayer`) additionally composes the **class locked base**, the **proficiency bonuses**, and the **class signature passive**. Since every character has a class, the screen's raw damage, DPS/burst, cooldown, and crit numbers all understated the live battle and visibly disagreed with the in-fight `SkillTooltip`.

There were three drifting hand-rolled copies of this composition: the battle engine, the attribute-breakdown screen (already correct), and the skills screen (missing three of five pieces). This PR extracts one shared builder and has all three consume it:

- New `composePlayerBattleAttributes`/`playerBattleModifiers` — pure functions in `UI/src/lib/battle/player-battle-composition.ts` — assemble the player's live `BattleAttributes`: allocation + equipped gear, the class locked base, the proficiency bonuses, and the class signature passive (added last, resolved against the fully-assembled set), mirroring the backend's `BattleSnapshot.ToBattler`.
- `battle-engine.ts`'s `resetPlayer` now builds its `additionalModifiers` array via `playerBattleModifiers(...)` instead of an inline spread.
- `skills-view.svelte.ts`'s `battleAttributes` now calls `composePlayerBattleAttributes(...)`.
- `attribute-breakdown-view.svelte.ts`'s `buildPlayerModifiers()` now sources its locked-base + proficiency section from `playerBattleModifiers(...)`.

The functions are pure (inputs passed in) rather than reading the live stores/singletons directly, matching this codebase's existing convention for shared display-battle helpers (`sortMetrics`, `buildGroups`, etc.) and keeping them unit-testable independent of the engine.

### Related smaller gap (from the same issue)

`metricsById` was built only over the player's **unlocked** skills, but an innate item-granted skill is deliberately never unlocked — so a typical innate card (`InnateBand.svelte`) silently lost its dmg/cd/dps block. `metricsById` now covers the union of unlocked skills and `inventoryManager.grantedSkillIds`.

## Test plan

- [x] Added coverage in `skills-view.test.ts` for the locked-base, proficiency, and signature-passive contributions folding into `rawDamage`, and for `metric()` resolving a granted-but-unlocked skill.
- [x] Updated the `$stores`/`$lib/engine` mocks in `EquippedBand.test.ts`, `InnateBand.test.ts`, `SkillCardStats.test.ts`, and `Skills.test.ts` to supply the new `battleLockedBaseModifiers`/`battleSignaturePassiveModifier`/`playerProficiencies.battleModifiers`/`grantedSkillIds` fields the composition now reads.
- [x] `npx vitest --run` — full frontend suite (3192 tests) passes.
- [x] `npx eslint .` and `npx svelte-check` — clean.
- [x] Updated `docs/frontend.md`'s "scalar formulas are shared with the display surfaces" note to call out the new shared composition builder.

Closes #1500

---
_Generated by [Claude Code](https://claude.ai/code/session_01ENT42japmAaGr6oJdYn2rC)_


ginderjeremiah (commented):
## Review: Fix Skills screen's incomplete battle-attribute composition (#1500)

**Verdict:** Approve _(posted as a `COMMENT` because I'm the PR author and GitHub won't accept a self-gating verdict)_ — the extracted shared builder faithfully reproduces the live battler's composition, the innate-metrics gap is closed, and coverage is solid.

### What I verified

- **Composition parity is genuinely preserved.** `composePlayerBattleAttributes` calls `setData([...attributes, ...equipmentStats], true, playerBattleModifiers(lockedBase, proficiency))`, and `setData` with `calcDerived=true` appends `STATIC_ATTRIBUTE_MODIFIERS` internally (`battle-attributes.ts:100`), then the signature passive is `addModifier`'d last against the assembled set. That is bit-for-bit the order `resetPlayer` produces (`battle-engine.ts:242-258`: `player.reset(..., playerBattleModifiers(...))` → `setData` → signature passive last), so the additive accumulation the anti-cheat replay depends on matches. The earlier worry — that the display omits the static modifiers and would resolve an attribute-scaled passive against a different base than the live battler — doesn't hold, precisely because `setData(true)` folds statics in before the passive resolves.
- **`metricsById` fix is correct.** Iterating the union of `catalogue` ids and `inventoryManager.grantedSkillIds`, then resolving `skills[id]` (valid under the documented 0-based reference-data ids, `backend.md`) with a graceful `continue` on a missing/retired id, closes the innate-card gap without disturbing the unlocked catalogue.
- **attribute-breakdown-view refactor is behavior-identical** — the single `playerBattleModifiers(...)` loop preserves the prior `[lockedBase, proficiency]` order.
- The `(resolve) => playerManager.battleSignaturePassiveModifier(resolve)` wrapper correctly preserves `this`; passing the method bare would have risked an unbound call.

### Test coverage

Good. `skills-view.test.ts` exercises each folded component into `rawDamage` (locked base, proficiency, signature passive) and the granted-but-unlocked `metric()` path; the four band/screen mock updates keep the new inputs consistent. Battle logic itself is unchanged (this DRYs an already-mirrored FE composition), so no backend mirror test is owed.

### Nits (non-blocking)

- The new pure module `player-battle-composition.ts` has no dedicated unit-test file — its contract (compose order, and the signature passive resolving against the fully-assembled set) is only exercised transitively through the view tests. A small focused test on `composePlayerBattleAttributes`/`playerBattleModifiers` would document that contract directly and guard it independently of the view. Optional.

### What's good

Extracting one builder and having all three consumers read off it is exactly the fix the issue called for, and the doc comments + `frontend.md` update make the "shared inputs, not just shared formulas" invariant explicit so a fourth copy can't quietly drift again.

_Note: `test-frontend` and the e2e jobs were still in progress at review time; `verify-frontend` (eslint/svelte-check) is green._


---

ginderjeremiah:
## Review — approve (with nits)

Reviewed against the frontend parity docs and the backend `BattleSnapshot.ToBattler`. This is a clean, well-scoped fix that resolves the real bug in #1500. **Verdict: approve.** CI is green (`test-frontend`, `verify-frontend`, all e2e browsers pass; backend jobs correctly skipped for a FE-only change).

### Parity verification (the load-bearing claim)
I checked the new composition against the backend line-by-line and it matches bit-for-bit:

- Backend `GetModifiers`: stat allocations → equipped gear → class locked base → proficiency bonuses; `AttributeCollection` appends `StaticAttributeModifiers`; then `ToBattler` calls `AddModifier(passive)` **last**, resolved against the fully-assembled set via `attributes.GetAttributeValue`.
- New `composePlayerBattleAttributes`: `setData([...attributes, ...equipmentStats], true, [...lockedBase, ...proficiency])` (which appends `STATIC_ATTRIBUTE_MODIFIERS`) → `addModifier(resolveSignaturePassive(attr => attrs.getValue(attr)))`. Same order, signature passive resolved against the fully-resolved scaling attribute.
- `Battler.reset` (the engine path) uses `[...battlerData.attributes, ...additionalAtttributes]` + `playerBattleModifiers(...)` + its own `addModifier(passive)` afterward — bit-identical to the helper. Confirmed the engine and the helper agree.

The signature-passive-LAST requirement is genuinely satisfied and directly tested (`Luck = resolvedEndurance × 2` case), and there's an equivalence test asserting the helper matches the manual `setData` + `addModifier` production path.

### Strengths
- Fixes the actual understatement bug; all three surfaces now agree with the in-fight `SkillTooltip`.
- Pure over its inputs (no store/singleton coupling), matching the existing display-shared-helper convention — cleanly unit-testable.
- The `metricsById` gap is fixed correctly: `metricsSkillIds` is the union of unlocked catalogue skills + `grantedSkillIds`, resolved by index per the zero-based-id lookup convention, with a proper `if (!skill) continue` guard.
- Strong test coverage — dedicated `player-battle-composition.test.ts` plus skills-view tests for locked-base / proficiency / signature-passive folding and a granted-but-unlocked skill.
- No null-forgiving (`!`) operators introduced; the now-unused `BattleAttributes` import was correctly dropped from `skills-view`; `docs/frontend.md` updated.

### Nits (non-blocking)
1. **The dedup is partial for the passive step.** `playerBattleModifiers` (locked base + proficiency pairing) is now genuinely shared across all three consumers, but the *signature-passive-added-last* step is only shared via `composePlayerBattleAttributes` (skills screen). The battle engine still applies it inline (`addModifier` after `Battler.reset`, needed for its re-arm optimization) and the breakdown screen applies it inline via `computeAttributes` (needed for labeled-modifier provenance). So the PR's "a fourth copy can't drift" holds for the base-modifier pairing but the passive application still lives in ~3 places. This is largely unavoidable given the engine/breakdown structural differences and the parity suite guards a drift, but a one-line comment on the engine's inline `addModifier` pointing at the shared helper as the canonical order would help the next reader. Optional follow-up at most.
2. **Trivial:** `composePlayerBattleAttributes` does `new BattleAttributes()` (whose constructor already runs `setData([], true)`) then immediately `setData(...)` again — one redundant empty init. Harmless; the constructor can't take `additionalModifiers` so the current shape is reasonable.

### Merge recommendation
Ship it. No blocking issues. The nits are optional cleanup, not merge-gating.

---

ginderjeremiah:
Merged. The optional passive-last consolidation/annotation noted in the review is tracked as follow-up #1549.

---

PR #1546, opened 2026-07-03
Reconcile DefeatEnemy's authoritative post-victory state on the client

### LINKED ISSUE #1501: DefeatEnemy's authoritative newLevel/newExp/statPointsGained are discarded by the client — a lost response leaves exp/level diverged all session
## Problem

The backend deliberately returns post-grant authoritative state (`NewLevel`, `NewExp`, `StatPointsGained`, `StatPointsUsed`) on every victory (`BattleService.EndBattleVictory`, surfaced on the `DefeatEnemy` response), but the client's `claimVictory` (`UI/src/lib/engine/battle/enemy-manager.ts:626-652`) reads only `rewards.expReward` and re-derives level-ups locally via `playerManager.grantExp` (`player-manager.ts:181-199`, unclamped — the server clamps to `MaxExpPerGrant`). On a response error it only logs.

## Failure scenario

The server processes `DefeatEnemy` and credits exp, but the response is lost (socket drop mid-claim, or the 30s request timeout — the transport settles the request with `error` and deliberately never re-sends). The client skips `grantExp`, and there is no re-sync until re-login/welcome-back (`refreshPlayer` is only called from the welcome-back gate). Because the class locked base is `baseAmount + amountPerLevel × level`, a missed level-up makes the FE battler **weaker than every subsequent server snapshot**: battle durations diverge (spamming the server's battle-duration divergence warning meant for real parity bugs), and edge battles get misclassified client-side — the client can abandon/lose fights the server replay would score as wins, forfeiting rewards.

## Suggested fix

Nearly free: assign the returned authoritative fields in `claimVictory` instead of (or as a reconciliation after) the local `grantExp`, and call `refreshPlayer()` when the claim settles with an error. Add a test for the error-settled claim path.

ginderjeremiah:
## Summary

The backend returns post-grant authoritative state (`NewLevel`, `NewExp`, `StatPointsGained`, `StatPointsUsed`) on every victory (`BattleService.EndBattleVictory`, surfaced on the `DefeatEnemy` response), but `claimVictory` (`UI/src/lib/engine/battle/enemy-manager.ts`) only read `rewards.expReward` and re-derived level-ups locally via `playerManager.grantExp` — the server's clamp to `MaxExpPerGrant` and any other server-side authority were silently discarded. On a response error it only logged, with no reconciliation at all.

### Failure scenario this fixes

The transport settles a lost/timed-out `DefeatEnemy` as an `error` response even when the command actually succeeded server-side (per the documented socket request lifecycle — commands aren't blind-resent since they aren't idempotent). Previously, the client would skip `grantExp` entirely on that error and never resync, leaving exp/level diverged from the server for the rest of the session — causing battle-duration divergence and misclassified edge battles (a fight the server replay would score as a win could be scored a loss client-side).

## Changes

- `PlayerManager.applyVictoryRewards(rewards)` — grants exp locally via the existing `grantExp` (preserving the "Earned X exp." / level-up log messages), then reconciles `level`/`exp`/`statPointsGained`/`statPointsUsed` onto the server's authoritative response fields, since the backend's `MaxExpPerGrant` clamp is a case `grantExp` deliberately doesn't mirror.
- `EnemyManager.claimVictory` now calls `applyVictoryRewards` instead of `grantExp` directly, and calls `refreshPlayer()` (the same Login/Status resync the welcome-back gate already uses) when the claim settles with an `error` — so a lost/timed-out response no longer leaves the client silently diverged.

## A note on the issue

Existing `grantExp`/`levelUp` unit and progression-parity tests are untouched and remain meaningful — `applyVictoryRewards` calls through `grantExp` for its log-message UX before reconciling, so the parity coverage of the exp/level formula itself still guards against drift in the common (non-clamped) case; the reconciliation step is the backstop for the clamp/lost-response cases the issue describes.

## Test plan

- [x] `npx vitest --run` — full frontend suite (3190 tests) passes, including new `applyVictoryRewards` coverage in `player-manager.test.ts` and updated `enemy-manager-boss.test.ts` coverage asserting `refreshPlayer()` fires only on an errored claim.
- [x] `npx eslint .` and `npx svelte-check` — clean.

Closes #1501

---
_Generated by Claude Code_

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qa6FmUrcaqwU7NnfHwwsGK)_

### DIFF
--- a/UI/src/lib/engine/player/player-manager.ts
+++ b/UI/src/lib/engine/player/player-manager.ts
@@ -207,6 +207,11 @@ export class PlayerManager implements IPlayerData {
 	 * server-side — always win over the local recompute.
 	 */
 	public applyVictoryRewards(rewards: IDefeatRewards) {
+		// grantExp is invoked only for its log-message side effect (the "Earned X exp." / level-up lines);
+		// its own level/exp recompute is discarded below. `expReward` is the server's pre-clamp raw reward,
+		// so in the (effectively unreachable in normal play) case of a single grant exceeding
+		// MaxExpPerGrant, this can over-level and log phantom level-up lines before the reconcile below
+		// corrects level/exp back down to the server's actual post-clamp state.
 		this.grantExp(rewards.expReward);
 		this.level = rewards.newLevel;
 		this.exp = rewards.newExp;
--- a/UI/src/lib/engine/player/player-manager.ts
+++ b/UI/src/lib/engine/player/player-manager.ts
@@ -2,6 +2,7 @@ import {
 	ELogType,
 	IAttributeDistribution,
 	IBattlerAttribute,
+	IDefeatRewards,
 	IInventoryData,
 	ILogPreference,
 	IPlayerData,
@@ -196,6 +197,21 @@ export class PlayerManager implements IPlayerData {
 		this.statPointsGained += STAT_POINTS_PER_LEVEL;
 		logMessage(ELogType.LevelUp, 'Congratulations, you leveled up!');
 		logMessage(ELogType.LevelUp, `You are now level ${this.level}.`);
+	}
+
+	/**
+	 * Applies a battle victory's exp grant, then reconciles onto the server's authoritative post-grant
+	 * state. `grantExp` re-derives level/exp/statPointsGained locally for its log messages, but the
+	 * backend additionally clamps a single grant to `MaxExpPerGrant` (an anti-cheat backstop the frontend
+	 * deliberately doesn't mirror) and is the source of truth regardless, so `rewards`' fields — computed
+	 * server-side — always win over the local recompute.
+	 */
+	public applyVictoryRewards(rewards: IDefeatRewards) {
+		this.grantExp(rewards.expReward);
+		this.level = rewards.newLevel;
+		this.exp = rewards.newExp;
+		this.statPointsGained = rewards.statPointsGained;
+		this.statPointsUsed = rewards.statPointsUsed;
 	}
 }
 
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -3,6 +3,7 @@ import { Action, createHook, delay, isZoneUnlocked, navigableZones, nextZoneByOr
 import { staticData, statistics, playerChallenges } from '$stores';
 import { battleEngine, BattleStage, onBattleStageChanged, playerManager, type PlayerBattleState } from '../';
 import { logMessage } from '../log';
+import { refreshPlayer } from '../session';
 
 const newEnemyLoadedHook = createHook<[IEnemyInstance]>();
 const notifyNewEnemyLoaded = newEnemyLoadedHook.notify;
@@ -638,9 +639,16 @@ export class EnemyManager {
 			if (enemyName) {
 				logMessage(ELogType.EnemyDefeated, enemyName + ' was defeated!');
 			}
-			playerManager.grantExp(defeatResponse.data.rewards.expReward);
+			playerManager.applyVictoryRewards(defeatResponse.data.rewards);
 		} else {
 			logMessage(ELogType.Debug, 'There was an error defeating the enemy: ' + defeatResponse.error);
+			if (defeatResponse.error) {
+				// The transport settles a lost/timed-out response as an error even when the command may have
+				// actually succeeded server-side (see docs/frontend.md's socket request lifecycle) — resync the
+				// authoritative player state rather than leaving exp/level silently diverged for the rest of
+				// the session.
+				await refreshPlayer();
+			}
 		}
 		// Guard `data` for a possible error response (absent `data`), now that this is the shared
 		// victory path for both the idle and boss loops.
--- a/UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts
+++ b/UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts
@@ -19,7 +19,8 @@ const h = vi.hoisted(() => ({
 		playerBattleStateMatches: vi.fn(() => true)
 	},
 	BattleStage: { Idle: 0, Active: 1, Victorious: 2, Defeated: 3, Loading: 4, Paused: 5, Drawn: 6 },
-	playerManager: { currentZone: 3, grantExp: vi.fn() },
+	playerManager: { currentZone: 3, applyVictoryRewards: vi.fn() },
+	refreshPlayer: vi.fn(() => Promise.resolve()),
 	staticData: {
 		enemies: [{ id: 0, name: 'Catacomb Lich', isBoss: true }],
 		zones: undefined as IZone[] | undefined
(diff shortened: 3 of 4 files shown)


UI/src/lib/engine/player/player-manager.ts
```diff
@@ -197,6 +198,21 @@ export class PlayerManager implements IPlayerData {
...
+
+	/**
+	 * Applies a battle victory's exp grant, then reconciles onto the server's authoritative post-grant
+	 * state. `grantExp` re-derives level/exp/statPointsGained locally for its log messages, but the
+	 * backend additionally clamps a single grant to `MaxExpPerGrant` (an anti-cheat backstop the frontend
+	 * deliberately doesn't mirror) and is the source of truth regardless, so `rewards`' fields — computed
+	 * server-side — always win over the local recompute.
+	 */
+	public applyVictoryRewards(rewards: IDefeatRewards) {
+		this.grantExp(rewards.expReward);
```
ginderjeremiah, line 215: Nit (edge-case, non-blocking): `rewards.expReward` here is the server's **pre-clamp** raw reward — `BattleService` sets `DefeatResult.ExpReward = rewards.ExpReward` (the raw `enemyAttTotal * DifficultyMultiplier`), while `newLevel`/`newExp` reflect the *post*-clamp state after `Player.ApplyExp` clamps to `MaxExpPerGrant`. So in the pathological case where a single grant exceeds `MaxExpPerGrant` (100k), the unclamped local `grantExp` over-levels and emits phantom "Congratulations, you leveled up!" / "You are now level N" log lines just before the reconciliation below corrects `level`/`exp` back down. The persisted state ends up correct, but the log briefly lies.

In normal play this is effectively unreachable (the ×4 `MaxExpRewardMultiplier` cap keeps per-battle exp far below 100k), so I'd leave it as-is rather than clamp on the client — just flagging that the second reconcile test (`expReward: 1_000_000`) silently exercises exactly this phantom-log path without asserting on it. A one-line note here that `grantExp` is invoked for its log side-effect and may over-log in the clamp edge would save a future reader the trace.

---

ginderjeremiah (commented):
## Review: Reconcile `DefeatEnemy`'s authoritative post-victory state on the client (#1501)

**Verdict:** Approve — correctly closes both halves of the bug (discarded authoritative state + no resync on a lost/rejected claim), scoped tightly, and well tested. _(Submitted as a `COMMENT` because GitHub won't accept a gating verdict on one's own PR; the intended verdict is **Approve**.)_

### Correctness
- `applyVictoryRewards` grants exp locally (preserving the "Earned X exp." / level-up log UX) and then reconciles `level`/`exp`/`statPointsGained`/`statPointsUsed` onto the server's authoritative fields — the right fix for the discarded-`newLevel`/`newExp` bug.
- The resync is well-placed: the backend returns `ErrorWithData` on a *rejected* claim (`DefeatEnemy.cs`), so `error` is set for both the lost/timed-out response **and** the server-rejected-divergence case — `refreshPlayer()` covers both. And the success path always populates `Rewards`, so the `!error && data?.rewards` guard can't fall through to a misleading "error … undefined" log. `refreshPlayer` → `loadPlayer` → `playerManager.initialize`, so the reconciled aggregate fully re-seats.
- Best-effort resync (a failed `loadPlayer` leaves in-memory state intact and retries on the next errored claim) is the right failure posture.

### Nits
- `player-manager.ts:210` — phantom level-up log lines in the (essentially unreachable) `MaxExpPerGrant`-clamp edge, since `expReward` is the pre-clamp raw value. Inline detail there; leave as-is, optionally a one-line comment. Non-blocking.

### Test coverage
- `player-manager.test.ts`: covers the normal grant-and-adopt path and the clamp-drift reconciliation — good.
- `enemy-manager-boss.test.ts`: asserts `applyVictoryRewards` is called with the rewards, `refreshPlayer` fires exactly once on an errored claim and **not** on success. `claimVictory` is the shared idle+boss path, so this one suite covers both loops. Adequate for the change.

### What's good
- Extracting the reconciliation into `applyVictoryRewards` (rather than inlining assignments in `claimVictory`) keeps the player-state authority in `PlayerManager` and mirrors the backend's `EndBattleVictory` return shape. Comments are accurate and appropriately scoped, and the honest "note on the issue" about why the existing `grantExp` parity tests stay meaningful is exactly the right call-out.

---

ginderjeremiah:
## Review: approve-with-nits

Reviewed the diff against the linked issue #1501 and the backend source. The fix is correct, well-scoped, and the claims in the description hold up against the actual code.

### Verified
- **Backend authority is real.** `BattleService.EndBattleVictory` returns `NewLevel`/`NewExp`/`StatPointsGained`/`StatPointsUsed` read from the player *after* `player.GrantExp(...)`, and `Player.GrantExp` clamps to `[0, MaxExpPerGrant]`. The `ExpReward` on the response is the pre-clamp raw reward, so the FE's `grantExp(rewards.expReward)` + reconcile-to-authoritative pattern is the right shape, and the doc comment's description of the clamp edge case is accurate.
- **The reconciliation is correct.** `applyVictoryRewards` overwrites `level`/`exp`/`statPointsGained`/`statPointsUsed` with the server fields after `grantExp`, so the common (non-clamped) path is a no-op overwrite and the clamped path corrects back down. `IDefeatRewards` carries exactly these fields.
- **Error-path resync is robust.** `refreshPlayer()` → `loadPlayer()` is fully `try/catch`-guarded and never throws, so the awaited call in the error branch can't reject. It resyncs over HTTP (`Login/Status`), which is the right call given the failure this targets is a dropped/timed-out *socket* response — a separate channel that can still succeed. Gating the resync on `defeatResponse.error` (rather than the whole `else`) correctly avoids a needless refetch in the anomalous no-error/no-rewards case.
- **No dead code / good isolation.** `grantExp` remains used (internally by `applyVictoryRewards`); the domain logic is encapsulated in `PlayerManager` and `EnemyManager` just delegates. No null-forgiving operator or other flagged patterns.
- **Tests are adequate.** New `applyVictoryRewards` coverage (normal grant + clamped-drift reconcile) and the boss-suite updates asserting `refreshPlayer` fires exactly once on an errored claim and never on success. `claimVictory` is the shared idle/boss path, so the boss suite exercises both.
- **CI green** — `all-checks-passed` success; frontend + all three e2e browsers pass (backend jobs skipped as expected for a FE-only change).

### Nits (non-blocking)
1. **`applyVictoryRewards` does work then discards it.** It calls `grantExp` purely for the "Earned X exp."/level-up log side effects, then overwrites its `level`/`exp` computation. It's documented, but in the clamped edge case it emits phantom level-up log lines before correcting. A cleaner design would set the authoritative fields and log the *actual applied* delta directly — worth considering but low value given the case is effectively unreachable in normal play.
2. **This PR newly writes `statPointsUsed` from the victory path.** The attributes screen updates `statPointsUsed` with a *relative* `+= netSpent` after its own `UpdatePlayerStats` resolves. Under ordered single-socket delivery + per-player serialization this doesn't misbehave, but the relative-increment pattern is inherently fragile against an absolute reconcile. Suggest a small follow-up to make the attributes-screen update adopt the server-authoritative value absolutely (matching this PR's pattern) rather than incrementing.
3. Trivial: the error log reads `"...error defeating the enemy: undefined"` in the no-error/no-rewards branch. Pre-existing wording, Debug-level; not worth churning here.

### Verdict
Approve with nits. The core fix is correct and the reconciliation/resync behavior matches the documented socket request lifecycle. None of the nits block merge; item 2 is the only one worth a follow-up issue.

---

ginderjeremiah:
Merged. The `statPointsUsed` relative-reconcile hardening noted in the review is tracked as follow-up #1548.

---

PR #1551, opened 2026-07-03
Share the signature-passive-last step between the battle engine and composePlayerBattleAttributes

### LINKED ISSUE #1549: Consolidate/annotate the signature-passive-last application shared across battle engine, skills, and attribute-breakdown
Follow-up from PR #1545 (which extracted the shared `composePlayerBattleAttributes`/`playerBattleModifiers` battle-attribute builder).

## Problem

PR #1545 successfully shares the base-modifier pairing (locked base + proficiency) across all three surfaces via `playerBattleModifiers`. However, the *signature-passive-added-last* step (resolved against the fully-assembled attribute set) is only shared through `composePlayerBattleAttributes` (used by the skills screen). The battle engine and the attribute-breakdown screen each still apply the passive inline:

- the engine for its re-arm optimization,
- the breakdown for labeled-modifier provenance.

So the PR's "a fourth copy can't drift" guarantee only fully holds for the base-modifier pairing, not the passive-last step. The parity suite guards against drift, but the inline copies are easy to miss when editing.

## Suggested work

- At minimum, add a short comment on the engine's and breakdown's inline `addModifier(passive)` pointing at the canonical `composePlayerBattleAttributes` helper as the source of truth for ordering.
- If practical, factor out a small shared "apply signature passive against assembled set" helper the inline paths can reuse without losing their respective optimizations/labeling.

Low priority — cleanup only, no user-facing effect.

ginderjeremiah:
## Summary

Follow-up to #1545. The signature-passive-added-last step existed as three copies: inside `composePlayerBattleAttributes`, inline in `BattleEngine.resetPlayer`, and inline in the attribute-breakdown modifier builder.

- Extracted `applySignaturePassive` into `$lib/battle/player-battle-composition` — the single "resolve the passive against the just-assembled set, add it last" step. `composePlayerBattleAttributes` and the engine's full-rebuild path (which can't consume the whole composer because of the data-less re-arm optimization, #811) now both call it, so the ordering is structural for those two surfaces rather than comment-guarded.
- The attribute-breakdown copy stays inline by design — it needs the labeled modifier itself, resolved lazily over its `computeAttributes` fold — and its comment now points at `applySignaturePassive` as the canonical ordering to keep in lockstep.
- No behavior change; the parity suites pass unchanged.

## Testing

- `vitest run` on `player-battle-composition.test.ts` (incl. two new `applySignaturePassive` cases: flat passive, and attribute-scaled passive resolved against the supplied set), `class-signature-passive-parity.test.ts`, `battle-engine-parity.test.ts`, `battle-engine.test.ts`, `battle-simulation-parity.test.ts`, `attribute-breakdown-view.test.ts`, `skills-view.test.ts` — all passing.
- `npm run check` (svelte-check): 0 errors / 0 warnings; eslint + prettier clean on touched files.
- Backend untouched (the backend composition lives in `BattleSnapshot.ToBattler` and is unchanged), so no mirrored backend scenario is required — the FE/BE parity scenarios themselves are unmodified.

Closes #1549

### DIFF
--- a/UI/src/lib/battle/player-battle-composition.ts
+++ b/UI/src/lib/battle/player-battle-composition.ts
@@ -25,12 +25,26 @@ export function playerBattleModifiers(
 	return [...lockedBaseModifiers, ...proficiencyModifiers];
 }
 
+/** The composition's LAST step — adds the class signature passive to `attrs`, resolved against that same
+ *  just-assembled set so an attribute-scaled passive reads the final value of its scaling attribute
+ *  (snapshot state, like a skill effect reads its caster). Callers must apply it after every other
+ *  modifier — float addition is not associative, so the anti-cheat replay depends on this apply order
+ *  matching the backend `BattleSnapshot.ToBattler`. `resolveSignaturePassive` is
+ *  `PlayerManager.battleSignaturePassiveModifier`, threaded through so the module stays pure. The
+ *  attribute-breakdown screen mirrors this step inline (it needs the labeled modifier itself, resolved
+ *  over its `computeAttributes` fold); keep that copy in lockstep with this one. */
+export function applySignaturePassive(
+	attrs: BattleAttributes,
+	resolveSignaturePassive: (resolveScalingValue: (attribute: EAttribute) => number) => AttributeModifier
+): void {
+	attrs.addModifier(resolveSignaturePassive((attribute) => attrs.getValue(attribute)));
+}
+
 /** Builds the player's full live {@link BattleAttributes}: allocation + equipped gear, the class locked
- *  base and proficiency bonuses, and the class signature passive composed last against the resolved set.
- *  Any surface displaying live battle numbers must read off this rather than a partial hand-rolled copy,
- *  so a composition change can't silently desync one surface from what the player actually fights with.
- *  `resolveSignaturePassive` is `PlayerManager.battleSignaturePassiveModifier`, threaded through so this
- *  stays pure — it is invoked with a resolver over the just-assembled set, as the live battler does. */
+ *  base and proficiency bonuses, and the class signature passive composed last against the resolved set
+ *  ({@link applySignaturePassive}). Any surface displaying live battle numbers must read off this rather
+ *  than a partial hand-rolled copy, so a composition change can't silently desync one surface from what
+ *  the player actually fights with. */
 export function composePlayerBattleAttributes(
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Share the signature-passive-last step between the battle engine and `composePlayerBattleAttributes` (closes #1549)

**Verdict:** Approve — clean, behavior-preserving refactor that directly resolves the issue; no blocking findings, no nits. (Posting as a `COMMENT` since GitHub won't accept a gating verdict on my own PR.)

Issue #1549 asked, at minimum, for comments pointing the engine and breakdown inline passive-applications at a canonical source of truth, and — if practical — to factor out a shared helper. This does the stronger option correctly: `applySignaturePassive` becomes the single "resolve the passive against the just-assembled set, add it last" step, and both `composePlayerBattleAttributes` and the engine's full-rebuild path call it, so the ordering is structural rather than comment-guarded for those two surfaces.

### Correctness

- The extracted helper is exactly equivalent to the three prior inline copies — `attrs.addModifier(resolveSignaturePassive((attribute) => attrs.getValue(attribute)))` — so the append order the anti-cheat replay depends on is preserved.
- The engine call site wraps the resolver as `(resolve) => playerManager.battleSignaturePassiveModifier(resolve)`, matching the existing `skills-view.svelte.ts:197` idiom — no `this`-binding footgun, and consistent with its sibling call sites.
- The re-arm optimization is respected: the passive is still applied only on a full rebuild (the data-less re-arm keeps the modifier since `setData` isn't re-run), so no behavior change there.

### Design

Leaving the attribute-breakdown copy inline is the right call — it needs the labeled modifier itself, resolved lazily over its `computeAttributes` fold — and its comment now names `applySignaturePassive` as the canonical ordering to keep in lockstep. That's the correct boundary; forcing it through the shared helper would have lost the labeling/laziness.

### Test coverage

Adequate. Two new dedicated unit tests for `applySignaturePassive` (flat passive; attribute-scaled passive resolved against the supplied set), and the engine and compose paths remain covered by the unchanged FE/BE parity suites, which is what guards the append-order-matches-backend invariant. Backend composition (`BattleSnapshot.ToBattler`) is untouched, so no mirrored backend scenario is required.

### What's good

The doc comments carry their weight: the helper's comment captures *why* the order matters (float addition isn't associative, replay parity) and explicitly flags the one remaining inline mirror to keep in lockstep — exactly the drift risk #1549 was about.


---

PR #1553, opened 2026-07-03
Measure the tooltip after render so first-show edge flips work

### LINKED ISSUE #1504: Global tooltip measures itself at 0×0 while display:none — keyboard-focus tooltips clip offscreen at viewport edges
## Problem

`UI/src/components/Tooltip.svelte:26-50` — the `style` derived reads `container.offsetWidth/offsetHeight` to decide the left/right and top/bottom flip, but on the evaluation that first makes the panel visible the container is still `display: none`, so both measure **0** and the flip logic never triggers.

Mouse hover self-heals because continuous `mousemove` repositioning re-evaluates with the panel now visible. Keyboard focus (`tooltip-hover.ts:46-51`) sets the position exactly **once**, so near the right/bottom viewport edges the ~280px panel overflows offscreen — precisely the focus-anchored accessibility path the docs standardize (`aria-describedby` + focus-reachable triggers).

jsdom reports `offsetWidth === 0` universally, which is why the existing clamp tests can't catch this.

## Suggested fix

Measure after the panel is actually rendered (e.g. flush/`tick()` then position, or measure a visibility-hidden clone/pre-layout state), or re-evaluate the position once post-render for the focus path. A Playwright check near a viewport edge would pin it (jsdom can't).

ginderjeremiah:
## Summary

`Tooltip.svelte` computed its positioning in a `$derived.by` bound to the `style` attribute. A derived evaluates **before** Svelte writes the attribute to the DOM, so on the evaluation that first makes the panel visible it read `offsetWidth`/`offsetHeight` while the container was still `display: none` — both measured 0x0 and the viewport-edge flip logic never fired. Mouse hover self-healed because continuous `mousemove` repositioning re-evaluated with the panel now visible; keyboard focus (`tooltip-hover.ts`) positions exactly once, so near the right/bottom edges the ~280px panel clipped offscreen — precisely the focus-anchored accessibility path.

Positioning is now an `$effect` that owns the container's inline styles: it applies `display: block` to the DOM first, measures the real box, then writes the anchor side per axis, clearing the stale opposite side on a flip. Hiding clears the inline display so the scoped `display: none` takes back over. Content injection (the sibling `$effect`) is declared first, so the panel is measured with its content in place on first show.

## Tests

- `UI/src/tests/components/Tooltip.test.ts` — all pre-existing scenarios kept passing unchanged; new `first-show measurement` block stubs `offsetWidth`/`offsetHeight` via `Object.defineProperty` to emulate real layout (0 until `display: block` is actually applied — jsdom otherwise always reports 0, which is why the existing clamp tests couldn't catch this). Verified the two new flip/reposition tests **fail against the old implementation** and pass with the fix.
- Ran the full `src/tests/components` + `src/tests/stores` suites (45 files, 367 tests) — all pass.
- `npm run check` (svelte-check) and eslint on the changed files — clean.
- Skipped the optional Playwright e2e pin: it requires the full backend stack running and the jsdom layout-emulation stub already pins the render-then-measure ordering the fix rests on.

No backend changes.

Closes #1504

### DIFF
--- a/UI/src/components/Tooltip.svelte
+++ b/UI/src/components/Tooltip.svelte
@@ -1,4 +1,4 @@
-<div id={tooltipElementId(id)} role="tooltip" class="tooltip-container" {style} bind:this={container}>
+<div id={tooltipElementId(id)} role="tooltip" class="tooltip-container" bind:this={container}>
 	<!-- The tooltip content is rendered directly into the container, so it is not present in the markup here. -->
 </div>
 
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Measure the tooltip after render so first-show edge flips work (#1504)

**Verdict:** Approve — the fix targets the exact root cause, the reasoning is sound, and the tests pin the behavior the old code couldn't catch. (Posted as a comment because GitHub won't let me formally approve my own PR; treat this as an approval.)

### Why the fix is correct

The bug was that `$derived.by` evaluates *before* Svelte writes the DOM, so `offsetWidth`/`offsetHeight` were read while the panel was still `display: none` (0×0) and the edge-flip never fired. Moving positioning into an `$effect` that imperatively sets `display: block` on the live node **and then** measures in the same synchronous body is the right shape: the reflow forced by reading `offsetWidth` now reflects the visible box. I confirmed the significance against `tooltip-hover.ts:46-51` — the focus handler calls `show()` exactly once with no follow-up movement, so first-show measurement is the *only* positioning the keyboard path gets, while mouse hover self-heals through continuous `onMove`.

A couple of subtleties that are handled correctly and worth noting for future maintainers:
- The content-injection `$effect` is declared before the positioning `$effect`, so on first show the panel is measured with its content already in place (effects run in declaration order). Good that the ordering is intentional.
- On hide, clearing the inline `display` (rather than setting `none`) correctly hands control back to the scoped `display: none`, and each show reconciles all four sides (left/right, top/bottom), so no stale anchor leaks across repositions.

### Test coverage

Strong and well-aimed. The `stubLayout` helper emulates the one thing jsdom gets wrong (0×0 until `display: block`), which is precisely why the pre-existing clamp tests couldn't catch this. The new `first-show measurement` block covers the flip on first show, the fits-in-viewport anchor, stale-side clearing on reposition, and inline-display clearing on hide. Verifying the new tests fail against the old implementation is exactly the right diligence.

Skipping the optional Playwright pin is a reasonable call: e2e is reserved for critical user flows per the testing guidelines, and the jsdom layout stub already pins the render-then-measure ordering the fix depends on.

### What's good

Tight, root-cause fix with no scope creep, correct Svelte 5 `$effect` semantics, and comments that explain the non-obvious reactivity timing rather than restating the code. Nicely done.

---

Issue #1516, opened 2026-07-03
Split BattleService: extract the offline/switch replay orchestration (940 lines, 13 constructor deps, three responsibilities)


ginderjeremiah:
## Problem

`Game.Application/Services/BattleService.cs` has grown to ~940 lines with 13 constructor dependencies, hosting three separable responsibilities:

1. **Live battle lifecycle** — start/boss/victory/loss/abandon.
2. **Offline/switch replay pipeline** — `SimulateProgress`, `ApplyOfflineRewards`, `ResolveOfflineLoop`, `BuildSimulationParameters`, the accumulator plumbing.
3. **Zone viability/relocation** — `IsZoneViable`, both `EnsureViableZone` overloads, `IsZoneUnlocked`.

The code is coherent today, but it is the one application-layer file trending toward unwieldy, and each new offline/consolidation feature (welcome-back additions, proficiency accumulators) lands here.

## Suggested direction

The offline-rewards orchestration is the natural extraction (an `OfflineRewardService` was even the name floated in #1042's scoping) — it touches the live paths only through `ResolveStaleBattle`. Zone viability could follow into a small helper if the split leaves it awkward. Pure refactor, no behavior change; the existing integration tests should pin the seams.

Not urgent — fold into the next substantial touch of the offline path rather than doing it standalone if preferred.

_Found during a codebase health review._

PR #1554, opened 2026-07-03
Extract the shared challengeable-boss zone gate in BattleService

### LINKED ISSUE #1540: Extract shared zone-validation preamble in BattleService
## Context

Follow-up from PR #1537 (validating client-supplied zone ids before resolving in `StartBattle`/`StartBossBattle`).

The `!ValidateZoneId(...) → no-op/return` guard, combined with the retired/locked/unlock preamble, now recurs across roughly five call sites in `BattleService` (`StartBattle`, `StartBossBattle`, `SetAutoChallengeBoss`, `IsZoneViable`, `ResolveOfflineLoop`). Each has different follow-on logic, so this was intentionally not blocking on #1537, but the repeated validation preamble is a mild DRY smell worth addressing.

## Suggested work

Evaluate extracting a small shared guard/helper that encapsulates the "resolve a client-supplied zone id, treating out-of-range/locked/retired as a no-op" pattern, so the individual call sites express only their divergent follow-on behavior. Keep it lightweight — do not force sites with genuinely different semantics into a shared abstraction if it hurts readability.

No user-facing effect expected.

ginderjeremiah:
## Summary

Follow-up to #1537: the "resolve a client-supplied zone id, treating out-of-range / retired / bossless / locked as a no-op" preamble recurred across three `BattleService` call sites. This extracts it as a small `ResolveChallengeableBossZone` gate:

- **`ResolveBossZone(zoneId)`** — the shared synchronous core: in range (checked before `GetDomainZone`, which throws on an out-of-range id), in circulation (not retired), and a dedicated boss authored.
- **`ResolveChallengeableBossZone(playerId, zoneId, ct)`** — live-path overload (`StartBossBattle`, `SetAutoChallengeBoss`); the unlock check goes through `IsZoneUnlocked`, still deferred behind the cheap checks so a bossless/retired target never pays the completion read.
- **`ResolveChallengeableBossZone(zoneId, completedChallengeIds)`** — set-taking overload for `ResolveOfflineLoop`, which already holds the loaded progress aggregate's completed-challenge set (no re-read of the progress cache key).

`StartBossBattle` also now reuses `BossEnemyResolver` instead of duplicating the boss-enemy resolving lambda inline, and the resolver's defensive throw message is generalised since it is no longer offline-only.

Deliberately **not** unified, per the issue's guidance: `StartBattle`'s zone-change guard (Home-zone check, ignore-and-continue semantics) and `IsZoneViable`/`EnsureViableZone` (spawn-table presence, relocation semantics) have genuinely different follow-on behavior. Kept conservative and local given #1516 plans to split this file later.

No user-facing effect — behavior-preserving; the check reordering inside `StartBossBattle` (retired before bossless) changes no outcome and adds no reads.

## Test notes

- All 59 tests in `BattleServiceIntegrationTests` + `BattleServiceOfflineProgressIntegrationTests` pass (55 pre-existing unchanged, 4 new).
- New edge coverage for the gate through its public paths:
  - `SimulateOfflineProgress_BossModeInBosslessZone_FallsBackToIdleInSameZone`
  - `SimulateOfflineProgress_BossModeInRetiredZone_FallsBackToIdleInViableZone` (asserts the relocation too)
  - `SimulateOfflineProgress_BossModeInLockedZone_FallsBackToIdleInSameZone`
  - `SetAutoChallengeBoss_OutOfRangeCurrentZone_ReturnsFalseAndLeavesModeIdle`
  (Out-of-range/retired/locked/bossless were already pinned for `StartBossBattle`, and retired/locked/bossless for `SetAutoChallengeBoss`.)
- Socket-level suites (`SetAutoChallengeBossSocketTests`, `ChallengeBoss*`) pass.
- No battle-simulation logic touched (server-side orchestration only), so no mirrored frontend scenarios are required.

Closes #1540

### DIFF
--- a/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
@@ -1861,5 +1861,33 @@ namespace Game.Application.Tests.Services
             Assert.False(success);
             Assert.False(player.AutoChallengeBoss);
         }
+
+        [Fact]
+        public async Task SetAutoChallengeBoss_OutOfRangeCurrentZone_ReturnsFalseAndLeavesModeIdle()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+
+            var zone = await TestDataSeeder.CreateZoneAsync(context);
+            var user = await TestDataSeeder.CreateUserAsync(context);
+            var playerEntity = await TestDataSeeder.CreatePlayerAsync(context, user.Id, zoneId: zone.Id);
+
+            await ReloadReferenceCachesAsync();
+
+            var playerRepo = scope.ServiceProvider.GetRequiredService<IPlayerRepository>();
+            var player = await playerRepo.GetPlayer(playerEntity.Id);
+            Assert.NotNull(player);
+
+            // A stale/corrupt CurrentZoneId (e.g. after a content reseed) must be a graceful rejection, not an
+            // ArgumentOutOfRangeException from resolving the domain zone.
+            player.ChangeZone(999);
+
+            var battleService = scope.ServiceProvider.GetRequiredService<BattleService>();
+
+            var success = await battleService.SetAutoChallengeBoss(player, enabled: true);
+
+            Assert.False(success);
+            Assert.False(player.AutoChallengeBoss);
+        }
     }
 }
--- a/Game.Application.Tests/Services/BattleServiceOfflineProgressIntegrationTests.cs
+++ b/Game.Application.Tests/Services/BattleServiceOfflineProgressIntegrationTests.cs
@@ -304,6 +304,108 @@ namespace Game.Application.Tests.Services
         }
 
         [Fact]
+        public async Task SimulateOfflineProgress_BossModeInBosslessZone_FallsBackToIdleInSameZone()
+        {
+            using var scope = CreateScope();
+            var setup = await SeedWinningScenarioAsync(scope);
+
+            var (player, state) = await LoadAsync(scope, setup.PlayerId);
+
+            // The persisted mode is boss-farming, but the current zone has no dedicated boss (e.g. it was
+            // unauthored since the mode was enabled): the loop must fall back to idle rather than stall.
+            player.SetAutoChallengeBoss(true);
+            player.LastActivity = DateTime.UtcNow.AddMinutes(-30);
+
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: Extract the shared challengeable-boss zone gate in `BattleService` (#1540)

**Verdict:** Approve — a correct, behavior-preserving DRY extraction that honors the issue's "don't over-unify" caution and lands the edge-case tests the old inline preambles lacked. (Posted as `COMMENT` because I'm the PR author and GitHub won't accept a self-`APPROVE`; the intent is Approve.)

### Behavior preservation (verified against all three call sites)

- `ResolveBossZone` short-circuits `!ValidateZoneId(zoneId) || IsZoneRetired(zoneId)` before `GetDomainZone`, so neither `IsZoneRetired` nor `GetDomainZone` (which throws on an out-of-range id) is ever reached with an invalid id — matching the old guards exactly.
- The `StartBossBattle` reordering (retired-before-bossless) is among pure cached-reference-data predicates that all fall through to `return null` with no side effects, and the only potentially-DB-touching check (`IsZoneUnlocked`) stays last and still gated behind boss-authored. So it's genuinely outcome- and read-count-neutral, as the description claims.
- `ResolveOfflineLoop` and `SetAutoChallengeBoss` reduce to the same predicate chain they had inline; the offline overload correctly reuses the already-loaded completed-challenge set rather than re-reading the progress cache key.

### Layering (docs/backend.md — DDD)

The gate orchestrates reference-data lookups (`_zones`) and delegates the actual unlock *rule* to the domain (`zone.IsUnlocked`) — domain logic stays in `Game.Core`, orchestration in the application layer. No mutation of cached zone graphs. `BossEnemyResolver` reuse removes the last duplicated inline lambda, and its defensive throw is now correctly unreachable for both callers.

### Scope judgment

Leaving `StartBattle`'s zone-change guard and `IsZoneViable`/`EnsureViableZone` un-unified is the right call — they have genuinely different follow-on behavior (Home-zone / relocation / spawn-table semantics), exactly the case #1540 warned against forcing into a shared abstraction.

### Test coverage

Good. The four new integration tests close the gaps the old inline code left: out-of-range `SetAutoChallengeBoss`, and the bossless / retired / locked offline-loop fallbacks — with the retired case also asserting the idle relocation. This is server-side orchestration (not battle-simulation logic), so no mirrored frontend parity scenario is required, correctly noted.

Nothing to change.


---

PR #1555, opened 2026-07-03
Skip zero-delta Sum statistics in PlayerProgress

### LINKED ISSUE #1515: RecordBattleCompleted creates and re-dirties zero-valued statistic rows every battle (hot-path churn + spurious challenge re-evaluation)
## Problem

`PlayerProgress.RecordBattleCompleted` (`Game.Core/Progress/PlayerProgress.cs:57-72,127-128`) unconditionally records the crit/dodge/parry/heal family (`CriticalHits`, `CriticalDamageDealt`, `AttacksDodged`, `DamageDodged`, `AttacksParried`, `DamageParried`, `CounterDamageDealt`, `DamageHealed`, `SkillsUsed`). `Record → Increment → GetOrCreate` (`:351-367`) creates the row and adds it to `_dirtyStatistics` even when the delta is **0**, so a player who never crits/dodges/parries still:

- gets those rows created on their first battle and persisted as dirty absolute writes after **every** battle thereafter, and
- has those keys included in the touched-statistics set, so every challenge keyed to e.g. `(CriticalHits, null)` is re-evaluated after every battle (`ChallengeIndex.cs:59-78`) despite nothing moving.

This is the per-player battle-completion hot path.

## Suggested fix

Skip a zero-delta `Sum` record (or skip the dirty-mark when the value didn't change).

**Caveat to preserve:** row *presence* is the domain's "has data" signal (`TryGetStatisticValue`; docs: a statistic's "no data yet" state is the absence of its row). Scope the fix to Sum-aggregated stats where `0 + 0` genuinely carries no information — the Min/Max first-write semantics (`:328-349`) must keep their created-row behavior (a legitimate first-write `0` minimum must still create the row).

ginderjeremiah:
## Summary

`RecordBattleCompleted` unconditionally recorded the crit/dodge/parry/heal family even when a battle reported 0, so `Record → Increment → GetOrCreate` created the row, re-dirtied it as an absolute write after every battle, and added its key to the touched-statistics set that drives `ChallengeIndex.RelevantTo` — spuriously re-evaluating every challenge keyed to a never-moved statistic on the per-player battle-completion hot path.

**Fix:** `Increment` (the Sum mutator) now returns early on a zero amount — no row creation (row absence stays the documented "no data" signal), no dirty-mark, no touched key.

**Scope deliberately preserved:** Min/Max first-write semantics are untouched. A first recorded 0 (an instant `FastestVictory`, a battle with no player attack for `HighestSingleAttackDamage`) is a genuine value and still creates and persists its row — pinned by a dedicated test. This is behavior-neutral for challenge evaluation: only Min-backed challenges use the `AtMost` / row-presence path, and Sum-backed (`AtLeast`) challenges read an absent row and a zero row identically.

The touched-set test whose comment asserted "a touched stat can legitimately be dirty with value 0" was rewritten as the new behavior's pin, plus new tests covering: the full zero-delta family creating no rows/dirty/touched keys, a zero delta on an existing loaded row staying clean, non-zero deltas still recording, the Min/Max zero-first-write asymmetry, and a zero-damage utility skill (uses recorded, per-skill DamageDealt absent).

Statistics recording is backend-only (no battle-parity surface), so no frontend mirror is required. Also documented the zero-delta skip in `docs/backend.md`'s player-progress section.

## Test notes

- `Game.Core.Tests` full suite: 1226 passed.
- `Game.Application.Tests` (Testcontainers): `PlayerProgressRepositoryIntegrationTests`, `BattleStatisticsEventHandlerTests`, `PlayerUpdateHandlerIdempotencyIntegrationTests` (55 passed), `BattleServiceIntegrationTests` (45 passed), `BattleServiceOfflineProgressIntegrationTests` + `OfflineProgressSummaryTests` (13 passed).

Closes #1515


ginderjeremiah (commented):
## Review: Skip zero-delta Sum statistics in `PlayerProgress` (#1515)

**Verdict:** Approve — correct, minimally-scoped fix for the hot-path churn; behavior-neutral for challenge evaluation and thoroughly tested. _(Submitted as a comment because GitHub won't accept a gating verdict on one's own PR; treat this as an approval.)_

The one-line guard in `Increment` is exactly the right seam. I verified the correctness argument end to end rather than taking it on faith:

- `Increment` is reached **only** from the `Sum` arm of `Record` (`PlayerProgress.cs:307-309`), so the skip cannot touch Min/Max first-write semantics — a legitimate first `0` (instant `FastestVictory`, no-attack `HighestSingleAttackDamage`) still creates and persists its row.
- The row-presence–sensitive challenge path (`AtMost`) is produced **only** when `AggregationKind == Min` (`ChallengeType.cs:31-33`), and the sole Min stat (`FastestVictory`) records via `SetMin`, never `Increment`. So no `AtMost` challenge is affected.
- For `AtLeast` (every Sum-backed) challenge, `UpdateProgress` ignores `hasData` and reads an absent row and a present zero identically (`PlayerChallenge.cs:49-57`) — the skip is genuinely behavior-neutral, not just usually so.
- The only Sum stat gated on row presence is `ZonesCleared` (`:100`), which only ever records `1` — never a `0` the skip could swallow.

Write-behind ↔ cache stays consistent (a never-created row is absent from both), and the offline loop accumulates correctly once a real delta arrives.

### Test coverage

Strong and targeted, matching the project's Detroit-style domain-testing rules:

- The zero-delta family creating no row / no dirty-mark / no touched key, and a zero delta over an existing loaded row staying clean.
- Non-zero deltas still recording, dirtying, and touching (guards against the skip swallowing real progress).
- The deliberate Min/Max zero-first-write asymmetry pinned explicitly.
- The zero-damage utility-skill case (use count records, per-skill `DamageDealt` stays absent, its `Max` twin keeps first-write semantics).
- The reused `MarksOnlyTheTouchedStatisticsDirty` test's stale "a touched stat can legitimately be dirty with value 0" comment is now correctly re-pinned as the new invariant.

Backend-only statistics persistence with no battle-simulation surface, so no frontend parity mirror is required — correctly called out.

### Nits

None. The guard comment is clear and the `docs/backend.md` invariant is updated to match. Clean change.

I'll merge once CI goes green.

---

PR #1556, opened 2026-07-03
Fix blank glyphs and tighten the glyph typing that hid them

### LINKED ISSUE #1502: Missing glyph entries render blank icons (Challenges KillsByDamageType, admin Classes 'gauge') — tighten the glyph typing that hid them
Two missing-glyph bugs, both hidden by weak typing:

## 1. Challenges Type Rail: `KillsByDamageType` renders a blank glyph everywhere

`UI/src/routes/game/screens/challenges/TypeGlyph.svelte:33-42` — `GLYPH_PATH` covers challenge types 0–8 (TimeTrial special-cased), but `EChallengeType.KillsByDamageType = 9` has no entry, so `<path d={undefined}>` renders an empty SVG in `RailButton`, `TypeHero` (including its 150px watermark), and the Overview ring meters. Everything else supports the type (`CHALLENGE_TYPE_META` and the `--challenge-kills-by-damage-type` token exist); only the glyph was missed. The `Record<number, string>` typing is why nothing caught it — key it on `EChallengeType` exhaustively (`Record<EChallengeType, string>`) so a future enum member fails typecheck. (Distinct from #1475, which is about the Statistics screen card.)

## 2. Admin sidebar: the Classes tool icon renders blank

`UI/src/routes/admin/workbench/entities/class.ts:28` declares `glyph: 'gauge'` (a valid `WorkbenchIconKind`), and `nav.ts:40` force-casts it to `AdminGlyphKind`, which has no `gauge` member — `AdminGlyph.svelte`'s if/else chain renders an empty `<svg>`. Add the gauge glyph and remove the `as` cast so TS catches future drift.

ginderjeremiah:
## Summary

Fixes both missing-glyph bugs from #1502 and hardens the typing so this class of drift fails typecheck instead of rendering blank icons.

**Challenges Type Rail — `KillsByDamageType` glyph.** `TypeGlyph.svelte`'s `GLYPH_PATH` was `Record<number, string>` and had no entry for `EChallengeType.KillsByDamageType`, so the rail button, hero banner/watermark, and Overview ring meters all drew an empty SVG. Added a typed-kill glyph (elemental droplet with a kill X, echoing the EnemiesKilled X) and re-keyed the record as `Record<Exclude<EChallengeType, EChallengeType.TimeTrial>, string>` — exhaustive over the enum (TimeTrial keeps its inline clock), matching the `CHALLENGE_TYPE_META` precedent, so a future challenge type without a glyph is a typecheck error.

**Admin sidebar — Classes tool icon.** `AdminGlyph.svelte` had no `gauge` branch, but `class.ts` declares `glyph: 'gauge'` and `nav.ts` force-cast it into `AdminGlyphKind`. Added the gauge icon (arc + needle + filled hub, in the file's two-tone style, matching WorkbenchIcon's gauge), removed the `as AdminGlyphKind` cast, and reconciled the two unions at the seam: `EntityConfig.glyph` is now `EntityGlyphKind = Extract<WorkbenchIconKind, AdminGlyphKind>`, since an entity's headline glyph renders through both components (sidebar + detail header). All existing entity glyphs are members of the intersection, so no other icons were missing. `AdminGlyphKind` is now derived from a runtime `ADMIN_GLYPH_KINDS` const so tests can prove every declared kind actually draws geometry.

## Tests

- New `TypeGlyph.test.ts`: renders every `EChallengeType` member (enumerated from the enum, so new members join automatically) and asserts non-empty path geometry; TimeTrial clock and color/size props covered.
- New `AdminGlyph.test.ts`: renders every `ADMIN_GLYPH_KINDS` kind and asserts visible geometry (the exact failure mode of the bug), plus the active-stroke behavior.
- `nav.test.ts`: every admin tool's glyph must be drawable by `AdminGlyph`, and the Classes tool carries `gauge`.
- Ran: `vitest` over `src/tests/routes/admin` + `src/tests/routes/game/screens/challenges` (61 files, 588 tests, all green), `npm run check` (svelte-check, 0 errors/warnings), eslint on changed files (clean). Frontend-only change; no backend impact.

Closes #1502

### DIFF
--- a/UI/src/routes/admin/AdminGlyph.svelte
+++ b/UI/src/routes/admin/AdminGlyph.svelte
@@ -30,6 +30,10 @@
 	{:else if kind === 'tag'}
 		<path d="M2.5 2.5h5L13 8l-5 5-5.5-5.5z" stroke-linejoin="round" />
 		<circle cx="5" cy="5" r="0.9" fill={stroke} stroke="none" />
+	{:else if kind === 'gauge'}
+		<path d="M2.8 11.5a5.2 5.2 0 0110.4 0" stroke-linecap="round" />
+		<path d="M8 11.3l2.7-3" stroke-linecap="round" />
+		<circle cx="8" cy="11.3" r="1" fill={stroke} stroke="none" />
 	{:else if kind === 'back'}
 		<path d="M7 3.5L2.5 8 7 12.5M2.5 8H13" stroke-linecap="round" stroke-linejoin="round" />
 	{:else if kind === 'trophy'}
@@ -43,19 +47,24 @@
 </svg>
 
 <script lang="ts" module>
-export type AdminGlyphKind =
-	| 'skull'
-	| 'bars'
-	| 'rune'
-	| 'box'
-	| 'bolt'
-	| 'multiply'
-	| 'map'
-	| 'pin'
-	| 'tag'
-	| 'back'
-	| 'trophy'
-	| 'inbox';
+/** Runtime list of every glyph kind (each must have a branch above) — lets tests assert none render blank. */
+export const ADMIN_GLYPH_KINDS = [
+	'skull',
+	'bars',
+	'rune',
+	'box',
+	'bolt',
+	'multiply',
+	'map',
+	'pin',
+	'tag',
+	'gauge',
+	'back',
+	'trophy',
+	'inbox'
+] as const;
+
+export type AdminGlyphKind = (typeof ADMIN_GLYPH_KINDS)[number];
 </script>
 
 <script lang="ts">
--- a/UI/src/routes/admin/workbench/entities/types.ts
+++ b/UI/src/routes/admin/workbench/entities/types.ts
@@ -1,4 +1,11 @@
+import type { AdminGlyphKind } from '../../AdminGlyph.svelte';
 import type { WorkbenchIconKind } from '../WorkbenchIcon.svelte';
+
+/**
+ * An entity's headline glyph is drawn by both the sidebar ({@link AdminGlyphKind}) and the
+ * detail header ({@link WorkbenchIconKind}), so it must be a member of both glyph sets.
+ */
+export type EntityGlyphKind = Extract<WorkbenchIconKind, AdminGlyphKind>;
 
 /** Every workbench record is keyed by a numeric id (negative while unsaved). */
 export interface Identified {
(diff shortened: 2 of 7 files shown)


ginderjeremiah (commented):
## Review: Fix blank glyphs (`KillsByDamageType`, admin `gauge`) and tighten glyph typing — closes #1502

**Verdict:** Approve — both bugs from #1502 are fixed, and the typing is hardened so this class of drift becomes a typecheck error instead of a silently-blank icon. (Submitted as a comment because I'm the PR author; the intended verdict is approve.)

Both root causes are addressed exactly as the issue diagnosed, and the fixes go one step further by making the failure mode impossible to reintroduce quietly:

- **`TypeGlyph.svelte`** — `GLYPH_PATH` is now `Record<Exclude<EChallengeType, EChallengeType.TimeTrial>, string>`, so a future enum member without a glyph fails typecheck. The `{:else}` branch narrows TimeTrial away before indexing the record, and TimeTrial keeps its inline clock. The added `KillsByDamageType` path echoes the `EnemiesKilled` kill-X motif — a good visual fit.
- **`AdminGlyph.svelte` / `entities/types.ts` / `nav.ts`** — the `gauge` branch is added, the `as AdminGlyphKind` cast is removed, and `EntityConfig.glyph` becomes `EntityGlyphKind = Extract<WorkbenchIconKind, AdminGlyphKind>`. This is the correct reconciliation: an entity's headline glyph renders through both the sidebar and the detail header, so requiring membership in both sets is exactly right. I confirmed every top-level `EntityConfig.glyph` (gauge/box/trophy/map/skull/tag/rune/bolt) is in the intersection; the `target`/`gift` glyphs live on sub-structures typed against `WorkbenchIconKind` directly, so they're correctly unaffected. The gauge geometry matches `WorkbenchIcon`'s gauge (arc + needle) with a filled hub in AdminGlyph's two-tone style.
- Deriving `AdminGlyphKind` from the runtime `ADMIN_GLYPH_KINDS` const is a nice touch — it lets the test suite assert every declared kind actually draws geometry, which is the precise failure mode of this bug.

### Test coverage

Solid and aimed squarely at the defect:
- `TypeGlyph.test.ts` enumerates every `EChallengeType` member from the enum (new types join automatically) and asserts non-empty `d` geometry, plus TimeTrial's clock and the color/size props.
- `AdminGlyph.test.ts` renders every `ADMIN_GLYPH_KINDS` kind and asserts visible geometry (the blank-SVG failure mode), plus the active-stroke behavior.
- `nav.test.ts` pins that every tool's glyph is drawable by `AdminGlyph` and that Classes carries `gauge`.

Frontend-only change; no backend or battle-parity surface, so no mirrored backend tests are required.

### What's good

The fix doesn't just patch the two glyphs — it closes the typing gap that let them slip through, so the regression can't silently recur. No blocking findings, no should-fixes, no nits.

---

PR #1558, opened 2026-07-03
Share a NotSupportedException IPubSubService test fake base

### LINKED ISSUE #1541: Share a NotSupportedException IPubSubService test fake base
## Context

Follow-up from PR #1536 (fixing `SavePlayer`'s write-behind batch losing events on a failed publish).

The two test doubles introduced in `PlayerUpdateBatchTests.cs` — `RecordingPubSubService` and `ThrowingPubSubService` — each re-stub the ~11 `IPubSubService` members they don't exercise as `NotSupportedException`, duplicating ~20 lines. This is a minor DRY smell against the project's DRY guideline; it was non-blocking for #1536.

## Suggested work

Introduce a shared abstract `NotSupportedPubSubService` base (or equivalent) that stubs every member as `NotSupportedException`, and have the test fakes derive from it and override only `PublishBatch`. This removes the duplicated stubs and makes each fake's intent (record vs. throw) obvious.

No user-facing effect expected.

ginderjeremiah:
## Summary

Extracts the duplicated `IPubSubService` test-fake stubs (#1541) into a single abstract base, `NotSupportedPubSubService`, in `Game.TestInfrastructure/Helpers`. Every member is `virtual` and throws `NotSupportedException`, so each fake now overrides only the members it means to exercise — its intent (record vs. throw vs. single-queue) is obvious at a glance, and any accidental call to an unexercised member fails loudly instead of silently no-oping.

Converted fakes:

- **PlayerUpdateBatchTests** — `RecordingPubSubService` and `ThrowingPubSubService` override only `PublishBatch` (the issue's original target).
- **DataProviderSynchronizerTests** — `ThrowingPubSubService` overrides only the three `Subscribe` overloads; `SingleQueuePubSubService` overrides `GetQueue`, the `Subscribe` overloads, and the id-scoped `UnSubscribe` that `StopAsync` issues. This also tightens both fakes: members the synchronizer never touches were silent `Task.CompletedTask` no-ops before and now throw. `SubscribeSuppressingPubSubService` is deliberately untouched — it delegates to a real inner service.
- **SocketCommandProcessorTests** (`Game.Api.Tests`) — the same duplication existed here, which is why the base lives in the shared `Game.TestInfrastructure` (the `FakeWebSocket`/`CapturingLogger` precedent) rather than `Game.Application.Tests`. `ThrowingSubscribePubSubService` and `ThrowingUnsubscribePubSubService` keep only their exercised members. `CapturingPubSubService` is left as a direct implementation: it deliberately no-ops nearly every member, so deriving would save nothing and would mislabel its intent.

Net effect: -79 lines of copy-pasted stubs across three test files for one 24-line shared base. Test-only change, no user-facing effect.

## Test notes

- `dotnet build` of `Game.Application.Tests` and `Game.Api.Tests`: clean, 0 warnings.
- `dotnet test` (existing suites are the coverage, no new tests):
  - `Game.Application.Tests` filtered to `PlayerUpdateBatchTests` + `DataProviderSynchronizerTests` (Testcontainers-backed): 33/33 passed.
  - `Game.Api.Tests` filtered to `SocketCommandProcessorTests`: 10/10 passed.

Closes #1541

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
@@ -447,39 +447,22 @@ namespace Game.Api.Tests.Unit
             public void ReclaimAndForget(string key, string ownerValue, TimeSpan expiry) => throw new NotSupportedException();
         }
 
-        /// <summary>A pub/sub whose <c>Subscribe</c> throws, to drive the registration-rollback path.</summary>
-        private sealed class ThrowingSubscribePubSubService(Exception toThrow) : IPubSubService
+        /// <summary>A pub/sub whose <c>Subscribe</c> throws, to drive the registration-rollback path
+        /// (whose best-effort unsubscribe steps are tolerated as no-ops).</summary>
+        private sealed class ThrowingSubscribePubSubService(Exception toThrow) : NotSupportedPubSubService
         {
-            public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string id) => throw toThrow;
+            public override Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string id) => throw toThrow;
 
-            public Task Subscribe(string channel, string queueName, Action<(IPubSubQueue queue, string channel)> action, string id) => throw new NotSupportedException();
-            public Task Subscribe(string channel, Action<(string message, string channel)> action, string? id = null) => throw new NotSupportedException();
-            public Task Publish(string channel, string message, CancellationToken cancellationToken = default) => throw new NotSupportedException();
-            public Task Publish(string channel, string queueName, string queueData, CancellationToken cancellationToken = default) => throw new NotSupportedException();
-            public Task Publish<T>(string channel, string queueName, T queueData, CancellationToken cancellationToken = default) => throw new NotSupportedException();
-            public Task PublishBatch<T>(string channel, string queueName, IEnumerable<T> queueData, CancellationToken cancellationToken = default) => throw new NotSupportedException();
-            public Task Wake(string channel) => throw new NotSupportedException();
-            public Task UnSubscribe(string channel) => Task.CompletedTask;
-            public Task UnSubscribe(string channel, string id) => Task.CompletedTask;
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Share a `NotSupportedException` `IPubSubService` test-fake base (#1541)

**Verdict:** Approve — a clean, correctly-scoped DRY refactor that does exactly what #1541 asked, with no behavior change and no findings. (Posted as a `COMMENT` because I'm the PR author and GitHub won't accept a self-`APPROVE`; the intent is Approve.)

### What's good

- **Complete interface coverage.** The base stubs all 11 `IPubSubService` members (Publish×3, `PublishBatch`, `Wake`, Subscribe×3, UnSubscribe×2, `GetQueue`), so a derived fake never silently falls through to the real interface. `abstract` correctly prevents direct instantiation, and the generic `PublishBatch<T>` overrides cleanly.
- **"Fail loudly" is the right default.** Throwing `NotSupportedException` on unexercised members converts the previous silent `Task.CompletedTask` no-ops into loud failures — this actually *tightens* `DataProviderSynchronizerTests`' fakes, not just deduplicates them. Each fake now reads as its intent (record / throw / single-queue).
- **Placement is justified.** Living in `Game.TestInfrastructure/Helpers` (the `FakeWebSocket`/`CapturingLogger` precedent) is what lets `Game.Api.Tests` share it — verified that project references `Game.TestInfrastructure` and already imports the `Helpers` namespace, and the `public` base is reachable regardless of the `InternalsVisibleTo`.
- **The exclusions are correct, not laziness.** `SubscribeSuppressingPubSubService` delegates to a real inner service and `CapturingPubSubService` deliberately no-ops nearly everything — deriving either would save nothing and misrepresent intent. Good call leaving them as direct implementations.

### Test coverage

Appropriate. This is a behavior-preserving test-only refactor, so the existing suites (`PlayerUpdateBatchTests`, `DataProviderSynchronizerTests`, `SocketCommandProcessorTests`) *are* the coverage — no new tests warranted. The stated runs (33/33, 10/10) exercise every converted fake.

Net -79 duplicated lines for a 24-line shared base, with a strictly stronger failure mode. Nothing to change.


---

PR #1560, opened 2026-07-03
Fix combat log entrance animation and per-entry scroll nudge

### LINKED ISSUE #1510: Combat log: entrance animation self-cancels before paint; multi-entry ticks under-nudge the pinned scroll
## Problem

`UI/src/components/log-panel/LogPanel.svelte:38,104,115-125`:

1. **The 360ms slide-in never plays.** `animate` derives from `newestId !== seenId`, but the pinning `$effect` sets `seenId = newestId` in the same flush — the animation class is removed before paint. Rows keyed by id would animate natively; only the `atTop` gate is actually needed.
2. **Back-scrolled readers get shifted on bursty ticks.** The effect nudges `scrollTop += rowHeight` once per flush, but one battle tick often appends several entries — a reader scrolled back through history is shifted by N−1 rows per burst instead of staying visually anchored.

Both are minor visual defects but sit on the panel every player watches during idle battle.

ginderjeremiah:
## Summary

Two fixes to the combat log panel (`UI/src/components/log-panel`):

**1. Entrance animation self-cancelled before paint.** `animate` derived from `newestId !== seenId`, but the pinning `$effect` set `seenId = newestId` in the same flush, stripping the class before the browser ever painted it. The animation is now a mount-time effect: each keyed row (`log.id`) samples its `animate` prop once at creation, gated only by `atTop`. A first-render guard (`ready`, flipped in `onMount`) keeps a panel mounting over existing entries from flash-animating every row, and the mount-time sampling means scrolling back to the top can't retroactively replay a row's entrance. A side effect of the simpler gating: every row of a multi-entry burst animates in (previously only index 0), so the stack slides down smoothly instead of jumping N−1 rows.

**2. Back-scrolled readers were under-nudged on bursty ticks.** The effect nudged `scrollTop += rowHeight` once per flush, but one battle tick often appends several entries. The nudge now multiplies by the new-entry count computed from the monotonic log-id delta — clamped to zero because `resetLogs()` restarts the counter — and deliberately ignores array-length changes so the store's 40-entry tail-pop can't skew the math.

`LogRow`'s other consumer (the Options screen's `LivePreview`) already passes a mount-time-stable `animate` value, so its behavior is unchanged.

## Test notes

- `LogPanel.test.ts` now routes the `$stores` mock to the real logs store so tests drive reactive updates through `addLog`/`resetLogs`; `scrollTop` is property-stubbed since jsdom has no layout (per the issue triage, jsdom can't assert real animation playback, so the tests cover the animate-class gating and the nudge arithmetic).
- New cases: no animation on first render, burst rows animate while pinned, no animation while back-scrolled, no retroactive replay on scroll-to-top, pinned reader stays at top across a burst, N-entry burst nudges N×rowHeight, id-delta (not length) math at the 40-entry cap, and clamping across `resetLogs()`.
- Ran `vitest` over `src/tests/components` + `src/tests/routes/game` (1377 tests) plus the logs-store suite: all pass. `svelte-check`: 0 errors/warnings. ESLint on touched files: clean.
- Frontend-only visual fix — no battle-logic arithmetic changed, so no mirrored backend scenarios apply.

Closes #1510

### DIFF
--- a/UI/src/components/log-panel/LogPanel.svelte
+++ b/UI/src/components/log-panel/LogPanel.svelte
@@ -33,9 +33,9 @@
 		<span class="log-count">{eventCount} events</span>
 	</div>
 
-	<div class="log-body" bind:this={scrollEl} onscroll={onScroll}>
+	<div class="log-body" data-testid="log-body" bind:this={scrollEl} onscroll={onScroll}>
 		{#each allLogs as log, i (log.id)}
-			<LogRow {log} index={i} isLatest={i === 0} animate={i === 0 && isNew && atTop} {rowHeight} />
+			<LogRow {log} index={i} isLatest={i === 0} animate={ready && atTop} {rowHeight} />
 		{/each}
 		{#if allLogs.length === 0}
 			<div class="log-empty">No combat activity yet.</div>
@@ -98,10 +98,11 @@ let scrollEl: HTMLDivElement | undefined;
 // `atTop` is read inside the effect via `untrack` so scroll changes don't
 // re-run the pinning effect — we only want it to react to new entries.
 let atTop = $state(true);
-// Tracks the id we've already processed, so the entrance animation plays only
-// for a freshly-arrived top row — not every time the list re-renders.
-let seenId = $state(0);
-const isNew = $derived(newestId !== seenId);
+// False during the first render so rows already in the store mount without the
+// entrance animation; keyed rows mounted afterwards are genuinely new entries.
+let ready = $state(false);
+// Newest id the pinning effect has already positioned for (plain — nothing reacts to it).
+let seenId = 0;
 
 const onScroll = () => {
 	if (scrollEl) {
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Fix combat log entrance animation and per-entry scroll nudge (#1510)

**Verdict:** Approve — both defects from #1510 are correctly fixed, the approach is sound, and the test coverage is thorough. (Posted as a `COMMENT` because GitHub won't accept a self-approval on my own PR; treat this as an approve.)

I read the full head versions of `LogPanel.svelte`, `LogRow.svelte`, and `logs.svelte.ts` and traced both fixes:

**1. Entrance animation.** The root cause was `animate = newestId !== seenId` being stripped in the same flush the pinning effect set `seenId = newestId`, so the class never survived to paint. Moving the entrance to a mount-time sample (`LogRow`'s `const playEntrance = animate` with the `state_referenced_locally` ignore) is the right fix: rows are keyed by `log.id`, so each new row mounts fresh and samples `ready && atTop` exactly once — later prop changes can't cancel or replay it. The `ready` first-render guard correctly suppresses flash-animation of rows already in the store, and because existing rows never re-sample, scrolling back to the top can't retroactively animate an off-screen arrival. Keeping the class applied permanently is fine — a CSS keyframe plays once per element creation and won't loop.

**2. Scroll nudge.** `el.scrollTop += rowHeight * Math.max(0, id - seenId)` correctly nudges by the *new-entry count* (ids are monotonic via `++lastId`), fixing the N−1 under-nudge on bursty ticks. The `Math.max(0, …)` clamp is genuinely necessary: `resetLogs()` restarts `lastId`, so `newestId - seenId` goes negative for one flush, and without the clamp the view would jump backwards. I verified the transient reset→add ordering resolves correctly, and that the delta is immune to the 40-entry tail-pop (it keys off id, not `allLogs.length`).

**Naming the side effect:** every row of a burst now animates (not just index 0). The PR calls this out and it's a reasonable improvement — the stack slides in smoothly — so no concern, just noting it's intentional.

### Test coverage

Strong and well-matched to the two defects. Routing the `$stores` mock to the real `logs` store so tests drive `addLog`/`resetLogs` reactively is the right call, and stubbing `scrollTop` (jsdom has no layout) is a reasonable, documented workaround. The eight new cases cover both branches precisely: no-animate-on-first-render, burst-animates-while-pinned, no-animate-while-back-scrolled, no-retroactive-replay, pin-stays-at-top across a burst, N-per-entry nudge, count-not-length at the store cap, and the `resetLogs` clamp. `LogRow`'s changed behavior is exercised through the rendered `.manifest-row` classes in `LogPanel.test.ts` rather than a standalone suite, which is adequate here since the meaningful behavior is LogPanel's orchestration. Frontend-only visual fix with no battle-logic surface, so no backend mirror applies — correctly noted.

### What's good

- Correct root-cause diagnosis on the same-flush self-cancel, fixed structurally (mount-time sample + keyed rows) rather than patched.
- The monotonic-id delta is a clean way to count per-flush arrivals that's inherently robust to the tail-pop, with the clamp rationale documented where it's easy to miss.

Nothing blocking, should-fix, or nit. Good to merge once CI is green.

---

PR #1561, opened 2026-07-03
Fix Synthesis Web view input handling: recenter-on-push, wheel preventDefault, pointer capture

### LINKED ISSUE #1505: Synthesis Web view: viewport recenters on every ProficiencyXpGained push; wheel-zoom preventDefault is a no-op; pointer capture likely eats node clicks
Three interaction bugs in `UI/src/routes/game/screens/synthesis/SynthesisGraph.svelte`:

## 1. Recenter stomps the user's pan/zoom every few seconds while idling

The recenter `$effect` (`:140-144`) reads `layout.width/height`, but `layout` is a new object identity per derivation: every won battle's XP push reassigns the proficiencies array → new `recipes` → new `layoutSynthesisGraph` result → the effect re-fires → `recenter()` resets pan/zoom to origin. The graph is effectively un-pannable while the idle loop is running. Fix: derive width/height as primitive `$derived` values so the effect fires only on a real extent change.

## 2. Wheel-zoom `preventDefault()` is a no-op

Svelte 5 attaches `onwheel` as passive by default (`:15,126-135`), so `event.preventDefault()` is ignored — ancestors still scroll during zoom and the console warns per tick. The codebase already solves this correctly in `routes/loading/ManifestWindow.svelte:71` (manual `addEventListener` with `{ passive: false }`); apply the same here.

## 3. Pointer capture on every press likely retargets clicks away from node buttons

`onPointerDown` (`:103-108`) calls `setPointerCapture` on the container for **every** press with no `button === 0` guard. Per pointer-capture semantics the subsequent click retargets to the container, which likely kills mouse clicks on graph node buttons and the reset button (keyboard activation masks it). Capture only when a drag actually starts (movement threshold) or release capture on click-without-drag, and guard on the primary button.

ginderjeremiah:
## Summary

Fixes the three interaction bugs in `UI/src/routes/game/screens/synthesis/SynthesisGraph.svelte` from #1505:

1. **Recenter no longer stomps pan/zoom on every XP push.** The recenter `$effect` read `layout.width`/`layout.height` off the `layout` prop, so every `ProficiencyXpGained` push (which re-derives `SynthesisView.graph` as a new object identity) re-fired it and reset the viewport. The extent is now exposed as primitive `$derived` values (`layoutWidth`/`layoutHeight`) that the effect and `recenter()` key on — a same-extent re-derivation leaves the user's pan/zoom alone, while a real extent change (recipe set changed) still recenters.
2. **Wheel zoom uses a manual non-passive listener.** The template `onwheel` is replaced with an `addEventListener('wheel', onWheel, { passive: false })` in `onMount` (with cleanup), the same pattern as `ManifestWindow.svelte`. One honest note on the issue's premise: the currently installed Svelte (5.56.3) no longer forces `wheel` into its passive-event list (`PASSIVE_EVENTS` is just `touchstart`/`touchmove` now), so `preventDefault` was not an active no-op on this version. Earlier Svelte 5 releases *did* force it passive, and the zoom depends on `preventDefault` stopping ancestor scroll, so the explicit `passive: false` registration pins that contract rather than leaning on a framework default that has already flipped once.
3. **Pointer capture is guarded and deferred.** `onPointerDown` ignores non-primary buttons and no longer calls `setPointerCapture` on the bare press. Capture is taken only once the pointer travels past a small threshold (`isDragPastThreshold`, 4px Euclidean, extracted as a pure helper in `synthesis-graph.ts`), so a plain click reaches the graph-node and reset-view buttons instead of being retargeted to the canvas; once a drag really starts, capture keeps the pan alive off-element and the post-drag click retargets away from the node (no accidental select after a pan). The first pan step applies the accumulated pre-threshold delta so no travel is lost.

## Tests

- `synthesis-graph.test.ts`: unit tests for `isDragPastThreshold` (sub-threshold jitter, exact threshold, Euclidean diagonals, direction-agnostic). Existing layout/zoom tests unchanged and passing.
- `SynthesisGraph.test.ts` (new component suite): recenter on mount; pan preserved across a same-extent layout identity change (the XP-push shape); recenter still fires on a real extent change; plain press + sub-threshold jitter takes no capture and the node click still selects; capture only after the threshold; non-primary-button presses ignored; wheel zooms through the manually registered listener and cancels the event (`fireEvent.wheel` returning `false` proves `preventDefault` ran).
- Whether the listener is *non-passive* is not assertable in jsdom (jsdom ignores passive semantics), so that specific property is deliberately untested rather than covered by a vacuous test.
- Ran: `npx vitest run src/tests/routes/game/screens/synthesis` (6 files, 64 tests, all passing), `npm run check` (0 errors/warnings), eslint + prettier on the changed files. Frontend-only change — no backend/battle-logic surface touched.

Closes #1505

### DIFF
--- a/UI/src/routes/game/screens/synthesis/SynthesisGraph.svelte
+++ b/UI/src/routes/game/screens/synthesis/SynthesisGraph.svelte
@@ -12,7 +12,6 @@
 	onpointermove={onPointerMove}
 	onpointerup={endDrag}
 	onpointerleave={endDrag}
-	onwheel={onWheel}
 	role="application"
 	aria-label="Recipe graph — drag to pan, scroll to zoom"
 >
@@ -43,8 +42,9 @@
 </div>
 
 <script lang="ts">
+import { onMount } from 'svelte';
 import { recipeStateAccent } from './synthesis';
-import { type SynthesisGraphLayout, type Viewport, zoomAt } from './synthesis-graph';
+import { isDragPastThreshold, type SynthesisGraphLayout, type Viewport, zoomAt } from './synthesis-graph';
 import SynthesisGraphNode from './SynthesisGraphNode.svelte';
 
 type Props = {
@@ -58,8 +58,12 @@ const { layout, selectedRecipeId, onSelect }: Props = $props();
 let container = $state<HTMLDivElement>();
 let vp = $state<Viewport>({ x: 0, y: 0, scale: 1 });
 
-// Drag-to-pan state. The pointer is captured so a drag that leaves the node still pans the canvas.
+// Drag-to-pan state. The pointer is captured only once the drag threshold is crossed — capturing on the
+// bare press would retarget the tail click away from the node/reset buttons and eat their activation.
 let dragging = false;
+let captured = false;
+let startX = 0;
+let startY = 0;
 let lastX = 0;
 let lastY = 0;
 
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Fix Synthesis graph input handling — recenter, wheel, pointer capture (#1505)

**Verdict:** Approve _(submitted as a Comment — I'm the PR author, so GitHub won't accept a gating event from me; intended verdict is **Approve**)_ — all three interaction bugs are fixed correctly, the tricky Svelte-reactivity fix is empirically pinned, and the test suite is thorough. CI is green.

### Correctness — each bug addressed
- **Recenter no longer stomps pan/zoom.** Exposing the extent as primitive `$derived` (`layoutWidth`/`layoutHeight`) and keying the `$effect` on those is the right fix: a Svelte 5 derived short-circuits propagation when it recomputes to an `===`-equal value, so an XP-push re-derivation with the same extent never re-fires the effect, while a real extent change still recenters. The "preserves the user pan when the layout is re-derived with the same extent" test locks this at runtime, so it can't silently regress if the reactivity model shifts.
- **Wheel `preventDefault` is now effective.** Moving to `addEventListener('wheel', onWheel, { passive: false })` in `onMount` with cleanup mirrors the existing `ManifestWindow.svelte:71` precedent exactly (verified). The honest note in the description — current Svelte 5.56.3 no longer forces `wheel` passive, but earlier releases did — is a fair reason to pin the contract explicitly rather than lean on a framework default that has already flipped once.
- **Pointer capture no longer eats clicks.** The `event.button !== 0` guard plus deferring `setPointerCapture` until the 4px threshold is crossed means a plain press (or sub-threshold jitter) reaches the node/reset buttons, while a real pan still captures off-element. Applying the accumulated pre-threshold delta on the first pan step (lastX/Y still at the press point) avoids losing travel; the resulting ≤4px catch-up on threshold crossing is imperceptible and documented.

### Code quality
- Clean extraction of `isDragPastThreshold` into the framework-free `synthesis-graph.ts` keeps the threshold math unit-testable, matching the module's existing pure-helper pattern. No dead code (`onWheel` is still used by the manual listener), imports narrowed, all branches braced.
- Reactive logic stays in the component via `$derived`/`$effect` rather than a new store — aligned with the frontend guidance.

### Test coverage
Meets the project's rule for a component with UI interaction: the new `SynthesisGraph.test.ts` covers recenter-on-mount, pan-preserved-across-same-extent, recenter-on-real-extent-change, no-capture-on-plain-press-with-node-select, capture-only-past-threshold, non-primary-button ignored, and wheel-zoom-via-listener (asserting `preventDefault` ran). `isDragPastThreshold` gets its own unit tests (sub-threshold, exact, Euclidean diagonal, direction-agnostic). Correctly acknowledges that jsdom can't assert listener passive-ness and leaves that browser-only property untested rather than writing a vacuous assertion. No battle-parity surface, so no backend mirror required.

### What's good
Precise, well-scoped fix for all three bugs with a genuinely subtle reactivity fix (primitive-keyed effect) that's both explained and runtime-tested. Good judgment on the passive-listener contract and on the click-vs-drag threshold ergonomics.

No blocking findings, no should-fixes, no nits. CI is green, so I'll merge.

---

PR #1562, opened 2026-07-03
Chain a fresh fetch when a forced player-progress load hits an in-flight request

### LINKED ISSUE #1508: Player-progress stores: load(force) silently no-ops when a non-forced load is already in flight
## Problem

All three player-progress stores share the pattern `if (!inFlight) { inFlight = fetch...; } await inFlight;`:

- `UI/src/stores/challenges.svelte.ts:41-51`
- `UI/src/stores/proficiencies.svelte.ts:100-112`
- `UI/src/stores/statistics.svelte.ts:41-51`

A **forced** reload issued while a fetch is in flight just awaits the *old* fetch and never refetches — `force` doesn't actually mean "get fresh data".

## Consumers that rely on force meaning fresh

- `resolveBossVictory` calls `playerChallenges.load(true)` to detect whether the clear flipped the next zone's gate (`enemy-manager.ts:567`). If the boot-time load is still in flight (fast first boss kill on a slow connection), the check reads pre-clear data and the "Next zone unlocked" line is wrongly suppressed.
- `handleServerCommandFailed` (`engine.ts:94-98`) force-reloads to **recover from divergence** after a failed server push, and can likewise be satisfied by a stale in-flight response — leaving the client diverged despite the recovery protocol firing.

## Suggested fix

When `force` is passed and a fetch is in flight, chain a fresh fetch after the current one resolves (or track whether the in-flight request predates the force and refetch if so). Test: start a slow load, call `load(true)` mid-flight, assert a second fetch is issued.

ginderjeremiah:
## Summary

All three player-progress stores (challenges, statistics, proficiencies) shared the pattern `if (!inFlight) { inFlight = fetch(); } await inFlight;` — so a `load(force = true)` issued while a non-forced fetch was in flight just awaited the stale promise and never refetched. Consumers rely on force meaning fresh: `resolveBossVictory` reads post-clear gate state after `playerChallenges.load(true)`, and `handleServerCommandFailed` force-reloads to recover from divergence after a failed server push.

Rather than patching the three verbatim copies, the coalescing logic is extracted into a shared `CoalescedLoader` (`UI/src/lib/common/coalesced-loader.ts`):

- Concurrent non-forced loads still coalesce onto a single request.
- A force issued mid-flight queues **one** fresh fetch behind the stale in-flight one and resolves only after it completes; concurrent forces coalesce onto that same queued fetch, and a force arriving while the fresh fetch is already running chains another.
- `reset()` abandons any queued fetch (epoch guard), so a discarded session can't issue a refetch post-logout, and a stale fetch settling can't clobber a post-reset in-flight load.

The helper is deliberately **not** re-exported from the `$lib/common` barrel — the barrel transitively imports the stores (`enemy-attributes` → `battle-attributes` → `$stores`), so a barrel import from a store is circular. It's imported by path, matching the existing `local-storage` precedent.

## Test notes

- New `coalesced-loader.test.ts` covers the coalescing/force/chaining/reset matrix with manually-settled fetches (11 tests).
- Each store suite gains a mid-flight-force test (deferred `fetchSocketData` mock) asserting a second fetch fires after the stale one settles and its data wins.
- Ran: the 4 affected suites (45 tests), plus all engine + consumer screen suites (63 files, 857 tests) — all green. `npm run check` (svelte-check): 0 errors/warnings. ESLint on changed files: clean.
- Frontend-only change; no battle logic touched, so no mirrored backend scenarios required.

Closes #1508

### DIFF
--- a/UI/src/lib/common/coalesced-loader.ts
+++ b/UI/src/lib/common/coalesced-loader.ts
@@ -1,0 +1,74 @@
+/**
+ * Coalesces a store's concurrent `load()` calls onto a single in-flight fetch while keeping `force`
+ * honest: a forced load issued mid-flight queues one fresh fetch behind the stale in-flight one
+ * (concurrent forces share it) rather than resolving with data requested before the call. The
+ * player-progress stores rely on that guarantee for post-victory gate checks and for the
+ * divergence-recovery reload after a failed server push.
+ *
+ * Deliberately not re-exported from the `$lib/common` barrel: the stores consume it, and the barrel
+ * transitively imports the stores (via enemy-attributes → battle-attributes), so a barrel import
+ * from a store would be circular. Import it by path, like `local-storage`.
+ */
+export class CoalescedLoader {
+	/** Tail of the current load pipeline; concurrent callers await it instead of re-fetching. */
+	private inFlight: Promise<void> | undefined;
+	/** Whether a fresh (post-force) fetch is already queued behind the in-flight one. */
+	private freshQueued = false;
+	/** Bumped by {@link reset} so a queued fresh fetch from a discarded session never fires. */
+	private epoch = 0;
+
+	/** `fetchFn` must never reject — the stores' fetchers catch and record errors internally. */
+	constructor(
+		private readonly fetchFn: () => Promise<void>,
+		private readonly isLoaded: () => boolean
+	) {}
+
+	/**
+	 * Fetches via `fetchFn`. Idempotent — no-ops once loaded unless `force`, which always resolves
+	 * with data from a fetch issued no earlier than this call.
+	 */
+	async load(force = false): Promise<void> {
+		if (this.isLoaded() && !force) {
+			return;
+		}
+		if (!this.inFlight) {
+			await this.track(this.fetchFn());
+			return;
+		}
+		if (force && !this.freshQueued) {
+			// The in-flight fetch predates this force, so its response may be stale — queue one
+			// fresh fetch behind it; later forces coalesce onto that queued fetch.
+			this.freshQueued = true;
+			const chainEpoch = this.epoch;
+			this.track(
+				this.inFlight.then(() => {
+					if (this.epoch !== chainEpoch) {
+						return;
+					}
+					this.freshQueued = false;
+					return this.fetchFn();
+				})
+			);
+		}
+		await this.inFlight;
+	}
+
+	/** Forget any in-flight or queued fetch (e.g. on logout / session replacement). */
+	reset(): void {
+		this.inFlight = undefined;
+		this.freshQueued = false;
+		this.epoch += 1;
+	}
+
+	/** Publishes `pipeline` as the awaitable in-flight load, clearing it once it settles unless a
+	 *  chained fetch has replaced it in the meantime. */
+	private track(pipeline: Promise<void>): Promise<void> {
+		const tracked = pipeline.finally(() => {
+			if (this.inFlight === tracked) {
+				this.inFlight = undefined;
+			}
+		});
+		this.inFlight = tracked;
+		return tracked;
+	}
+}
--- a/UI/src/stores/challenges.svelte.ts
+++ b/UI/src/stores/challenges.svelte.ts
@@ -5,12 +5,11 @@
    challenge is completed). */
 
 import { fetchSocketData, type IPlayerChallenge } from '$lib/api';
+import { CoalescedLoader } from '$lib/common/coalesced-loader';
 
 let challenges = $state<IPlayerChallenge[]>([]);
 let loaded = $state(false);
 let error = $state(false);
-/** Coalesces concurrent callers (game boot + screen mount) onto one request. */
-let inFlight: Promise<void> | undefined;
 
 const fetchChallenges = async () => {
 	try {
@@ -24,6 +23,10 @@ const fetchChallenges = async () => {
 		error = true;
 	}
 };
+
+/** Coalesces concurrent callers (game boot + screen mount) onto one request; a forced load
+ *  issued mid-flight chains a fresh fetch so it never resolves with stale data. */
+const loader = new CoalescedLoader(fetchChallenges, () => loaded);
 
 export const playerChallenges = {
 	get all() {
@@ -39,15 +42,7 @@ export const playerChallenges = {
 	/** Fetch the player's challenge progress. Idempotent — only hits the network the first
 	 *  time unless `force` re-fetches the latest values (e.g. on the challenges screen). */
 	async load(force = false) {
-		if (loaded && !force) {
-			return;
-		}
-		if (!inFlight) {
-			inFlight = fetchChallenges().finally(() => {
-				inFlight = undefined;
-			});
-		}
-		await inFlight;
+		await loader.load(force);
 	},
 
 	/** Whether the player has completed the given challenge. */
@@ -75,6 +70,6 @@ export const playerChallenges = {
 		challenges = [];
 		loaded = false;
 		error = false;
-		inFlight = undefined;
+		loader.reset();
 	}
 };
--- a/UI/src/stores/proficiencies.svelte.ts
+++ b/UI/src/stores/proficiencies.svelte.ts
@@ -7,13 +7,12 @@
 import { fetchSocketData, type IPlayerProficiency, type IProficiencyXpGainedModel } from '$lib/api';
 import { proficiencyModifiers } from '$lib/battle/proficiency-modifiers';
 import type { AttributeModifier } from '$lib/battle/attribute-modifier';
+import { CoalescedLoader } from '$lib/common/coalesced-loader';
 import { staticData } from './static-data.svelte';
 
 let proficiencies = $state<IPlayerProficiency[]>([]);
 let loaded = $state(false);
 let error = $state(false);
-/** Coalesces concurrent callers (game boot + screen mount) onto one request. */
-let inFlight: Promise<void> | undefined;
 
 const fetchProficiencies = async () => {
 	try {
@@ -27,6 +26,10 @@ const fetchProficiencies = async () => {
 		error = true;
 	}
 };
+
+/** Coalesces concurrent callers (game boot + screen mount) onto one request; a forced load
+ *  issued mid-flight chains a fresh fetch so it never resolves with stale data. */
+const loader = new CoalescedLoader(fetchProficiencies, () => loaded);
 
 /**
  * The proficiency attribute bonuses for the player's current levels, composed against the proficiency
@@ -100,15 +103,7 @@ export const playerProficiencies = {
 	/** Fetch the player's proficiency progress. Idempotent — only hits the network the first time
 	 *  unless `force` re-fetches the latest values. */
 	async load(force = false) {
-		if (loaded && !force) {
-			return;
-		}
-		if (!inFlight) {
-			inFlight = fetchProficiencies().finally(() => {
-				inFlight = undefined;
-			});
-		}
-		await inFlight;
+		await loader.load(force);
 	},
 
 	/** Reset to the unloaded state (e.g. on logout / session replacement). */
@@ -116,6 +111,6 @@ export const playerProficiencies = {
 		proficiencies = [];
 		loaded = false;
 		error = false;
-		inFlight = undefined;
+		loader.reset();
 	}
 };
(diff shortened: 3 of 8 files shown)


ginderjeremiah (commented):
## Review: Chain a fresh fetch when a forced player-progress load hits an in-flight request (#1508)

**Verdict:** Approve — the fix correctly makes `force` mean "fresh," de-duplicates the three verbatim store copies behind a well-tested helper, and introduces no foot-guns. (Posted as a `COMMENT` because GitHub won't accept a self-approval on my own PR; treat this as an approve.)

I traced the `CoalescedLoader` state machine against each scenario and it holds up:

- **Coalescing** — concurrent non-forced loads share the single `inFlight` tail; a mid-flight non-forced load never queues a second fetch.
- **Force honesty** — a mid-flight force queues exactly one fresh fetch *sequentially behind* the stale one via `inFlight.then(...)`, so the fresh fetch is the last write and a late-settling stale response can't clobber it. The forced caller awaits the chained pipeline, so it resolves only after the fresh fetch completes — exactly what `resolveBossVictory` and `handleServerCommandFailed` need.
- **Concurrent forces** — the `freshQueued` flag collapses them onto the one queued fetch; a force arriving while the queued fetch is already running correctly chains another.
- **`reset()` safety** — the `epoch` guard neutralizes an already-queued fetch from a discarded session, and `track`'s identity check (`if (this.inFlight === tracked)`) stops a settling stale fetch from nulling out a post-reset in-flight load. Both edges are pinned by tests.

Braces are on every conditional, no null-forgiving operator, and the by-path import (not the `$lib/common` barrel) is the right call given the documented store→barrel circular-import cycle, matching the `local-storage` precedent.

### Test coverage

Excellent. `coalesced-loader.test.ts` drives manually-settled fetches through the full coalesce/force/chain/reset matrix including the retry-when-unloaded path, and each of the three store suites gains a mirrored mid-flight-force test (deferred mock) asserting the second fetch fires after the stale one settles and its data wins. This is backend-agnostic frontend state; no battle-parity mirror applies.

### What's good

- Extracting the shared helper instead of patching three copies, with the coalescing invariants captured as a documented contract (`fetchFn must never reject`) and the sequential-chaining rationale explained where it's easy to miss.
- The `epoch` + `track`-identity double guard against post-reset races is a genuinely subtle correctness detail that's both implemented and tested.

Nothing blocking or should-fix. Good to merge once the frontend/e2e CI jobs finish green.

---

ginderjeremiah (commented):
## Review: Chain a fresh fetch when a forced player-progress load hits an in-flight request (#1508)

**Verdict:** Approve — correctly fixes the `load(force)` staleness bug, DRYs the three verbatim copies into one well-tested helper, and handles the edge cases (reset, chaining, coalescing) the issue's naive "chain a fetch" suggestion would have missed. (Submitted as `COMMENT` only because I'm the PR author; GitHub won't accept a self-`APPROVE`.)

### What's good

- **The state machine is correct.** I traced every path in `CoalescedLoader`: first load, concurrent non-forced coalescing, mid-flight force (queues one fresh fetch behind the stale one and resolves only after it), concurrent forces coalescing onto that queued fetch, a force arriving while the fresh fetch is already running (chains another), and the `reset()` epoch guard. The `track()` `finally` correctly avoids clobbering `inFlight` after a chain or reset has replaced it (`this.inFlight === tracked`), and the `chainEpoch` capture cleanly abandons a queued fetch belonging to a discarded session.
- **Right call to extract rather than patch three copies.** DRY per CLAUDE.md, and it means the fix can't drift across the three stores.
- **The circular-import reasoning is sound and documented at the import site** — the `$lib/common` barrel transitively pulls in the stores, so importing by path (matching the `local-storage` precedent) is the correct avoidance, not an arbitrary style choice.
- **Test coverage is thorough and matches the project's rules.** The new `coalesced-loader.test.ts` unit suite mirrors the source path and drives the full coalesce/force/chain/reset matrix with manually-settled fetches (deterministic, no timing flakiness beyond an explicit microtask `flush`). Each of the three store suites gains the mid-flight-force test the issue asked for, asserting the second fetch fires and its data wins. Frontend-only, no battle-logic surface, so no mirrored backend scenarios are required.
- Scope is complete: the three player-progress stores were the only instances of this `if (!inFlight) { inFlight = fetch(); }` pattern; the other `inFlight` reference (`api-socket.ts`) is unrelated per-request timeout tracking with no `force` concept.

### Notes (non-blocking)

- The correctness of the chain relies on `fetchFn` never rejecting (if the in-flight promise rejected, `.then()`'s fulfilment handler wouldn't run and `freshQueued` would latch `true`). This is a real invariant, but it's explicitly documented on both the constructor and the class doc comment, and the stores' fetchers do catch internally — so it's a well-guarded contract, not a latent bug.

No blocking, should-fix, or nit findings. Clean change.

---

ginderjeremiah (commented):
## Review: Chain a fresh fetch when a forced player-progress load hits an in-flight request (#1508)

**Verdict:** Approve _(submitted as a Comment — I'm the PR author, so GitHub won't accept a gating event from me; the intended verdict is **Approve**)_ — the extraction fixes the real `force`-no-ops-mid-flight bug across all three stores, the state machine is correct under every concurrency ordering, and the test matrix is thorough.

### Correctness
Traced `CoalescedLoader` through every path — first load, coalesced non-forced loads, mid-flight force, concurrent forces, a force arriving while the queued fresh fetch is already running, and both `reset()` cases — and it holds:
- **Chaining, not concurrent issue, is the load-bearing choice.** The fresh fetch fires only after the stale one settles (`this.inFlight.then(...)`), so the fresh response is always the last write and can't be clobbered by the stale one. This is exactly what the store tests assert (`all`/`stats` equal the fresh data), and it's the property a concurrent re-issue would *not* guarantee.
- **The `inFlight === tracked` guard in `track()`** correctly prevents a settling stale fetch from nulling out a newer pipeline that already replaced it — verified against the post-`reset()` "new load then late stale settle" ordering.
- **The `epoch` guard** cleanly abandons a queued fresh fetch from a discarded session, and `reset()` bumping it before the chained `.then` runs means a post-logout refetch can't fire. Both pinned by tests.

### Design
- Keeping the helper **out of the `$lib/common` barrel** is correct and necessary — I confirmed the cycle is real (`index.ts` → `enemy-attributes` → `battle-attributes` → `$stores`), and the by-path import mirrors the documented `local-storage` precedent. Good call to document the reasoning in the class comment so a future contributor doesn't "tidy" it into the barrel and reintroduce the cycle.
- `fetchFn must never reject` is a real contract the class leans on (no `.catch` on the chain); it's satisfied because the stores' fetchers try/catch internally, and it's documented on both the class and the constructor. Fine as-is.

### Test coverage
Meets the project's lib-module rule (`tests/lib/common/coalesced-loader.test.ts` mirrors the source). 11 unit tests cover the full coalesce/force/chain/reset matrix with manually-settled fetches, and each of the three store suites gains a mid-flight-force integration test driving the real store through a deferred mock. Frontend-only, no battle-parity surface, so no backend mirror is required — correctly noted.

### Note (non-blocking, by-design)
When a store is **already loaded with no fetch in flight** and two forced loads land in the *same* tick, the first takes the direct-fetch path (no `freshQueued`) and the second queues a redundant chained fetch behind it, even though the first fetch would already satisfy it. This is a correctness-preserving conservative choice (it errs toward fresh, never stale) and the real consumers never issue two concurrent forces on the same store (`handleServerCommandFailed` forces two *different* stores; `resolveBossVictory` forces once), so the extra round trip can't occur in practice. Not worth complicating the state machine to shave.

### What's good
Clean DRY extraction that deletes three verbatim copies of the coalescing block, fixes a genuine staleness bug, and encodes the tricky "fresh must win the last write" and "abandon on session replacement" invariants explicitly with tests to match. Well-documented design decisions.

I'll hold the merge until CI (`test-frontend` + the e2e matrix) reports green.

---

ginderjeremiah (commented):
## Re-review: CI came back red — this supersedes my earlier Approve

**Verdict:** Request changes _(as a Comment — I'm the PR author, so GitHub won't accept a gating event; intended verdict is **Request changes** until CI is green)_ — the code is still correct and well-tested, but `test-frontend` **failed on the coverage gate** and it's this PR's doing, so it can't merge as-is.

### Blocking — `src/stores/**` branch-coverage gate

`test-frontend` ([run 28667332549](https://github.com/ginderjeremiah/GameServer/actions/runs/28667332549/job/85022265135)) exited 1 on:

```
ERROR: Coverage for branches (89.65%) does not meet "src/stores/**" threshold (90%)
```

Root cause is this change, even though the uncovered branch itself is pre-existing:
- The old inline coalescing in each store carried several **covered** branches (`if (loaded && !force)`, `if (!inFlight)`). This PR moves that logic into `CoalescedLoader` under `lib/common/`, so those covered branches leave the `src/stores/**` denominator.
- With them gone, the store folder's remaining **pre-existing** uncovered branch — `statistics.svelte.ts:57`, the `markZoneCleared` already-cleared early-return (`if (this.isZoneCleared(zoneId)) { return; }`) — now represents a bigger fraction, dropping the aggregate from ≥90% to 89.65% (`statistics.svelte.ts` branch coverage is 78.57%).

It's deterministic (coverage, not a flake) so a re-run won't clear it. The fix belongs in this PR:
- **Preferred:** add the missing `statistics` test — call `markZoneCleared(zoneId)` on an already-cleared zone and assert it no-ops (doesn't duplicate/re-add the row). That covers line 57's true branch, is a genuine gap under the project's "all store logic tested" bar, and lifts the folder back over 90%. (`challenges.svelte.ts:17` / `proficiencies.svelte.ts:20` are the analogous event branches but those files already sit at 90.9% / 92.3%; statistics is the one dragging the aggregate down.)
- Please re-run coverage locally (`npm run test` with the coverage gate) after adding it to confirm `src/stores/**` clears 90%, rather than adding exactly one branch and landing on the knife's edge.

### Everything else still stands (from the prior review)
The `CoalescedLoader` state machine is correct across every ordering, the barrel-exclusion is justified and matches the `local-storage` precedent, and the loader's own test matrix (11 cases) plus the three store mid-flight-force tests are thorough. None of that changes — this is purely the coverage gate.

I'll hold the merge until `test-frontend` is green.

---

ginderjeremiah (commented):
## Re-review: CI ran after the last approval and the store coverage gate is red

**Verdict: Request changes** — do not merge. _(Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating event from me; the intended verdict is **Request changes**.)_ The code change itself is still correct and I stand by the earlier trace of `CoalescedLoader`. But the three prior approvals all conditioned merge on green CI, and CI has since reported **failure** on `test-frontend`, which fails the `all-checks-passed` gate. This must be resolved before merge.

### Since the last review

- ✅ Correctness / design / test-logic findings — unchanged and still clean (coalescing, force-honesty via sequential chaining, `reset()` epoch + `track()` identity guards). No regressions in the diff.
- 🔁 **New blocker surfaced by CI:** `test-frontend` fails the coverage ratchet:
  ```
  ERROR: Coverage for branches (89.65%) does not meet "src/stores/**" threshold (90%)
  ```

### Blocking — restore `src/stores/**` branch coverage above 90%

Root cause (verified, not a flake): this PR moved the `if (loaded && !force)` / `if (!inFlight)` branches out of each store's `load()` and into `CoalescedLoader`. Those branches were exercised by the existing store tests, so they counted as *covered* branches under `src/stores/**`. `CoalescedLoader` lives in `UI/src/lib/common/`, **outside** the `src/stores/**` glob, so those covered branches left the store denominator entirely. That shrank the denominator and tipped a **pre-existing, untested** branch over the edge:

- `statistics.svelte.ts` `markZoneCleared` — the `if (this.isZoneCleared(zoneId)) { return; }` early-return (line 56–57) is not covered. `statistics.test.ts` covers `markZoneCleared` on an empty store (adds a row) and on an uncleared row (promotes it), but never on an **already-cleared** zone. `statistics.svelte.ts` is the drag at 78.57% branch, pulling the `stores` aggregate to 89.65%.

The uncovered branch is pre-existing code the PR didn't touch, but the PR is what pushed the gate red, so it has to be closed here (in this PR) to merge — not deferred to a follow-up.

**Suggested fix:** add a small `statistics.test.ts` case asserting `markZoneCleared(zoneId)` is a no-op when the zone is already cleared (calling it on a `value: 1` row and asserting the row isn't duplicated / value unchanged). That covers the exposed branch and should lift the `stores` aggregate back over 90%. Please re-run `vitest run --coverage` locally to confirm `src/stores/**` clears the threshold before pushing — if it's still marginally short, cover one more cheap store branch (e.g. `statistics` line 57 is the cheapest).

Everything else stands. Once coverage is restored and `test-frontend` (and thus `all-checks-passed`) goes green, this is an approve.

---

PR #1563, opened 2026-07-03
Fix a11y gaps: conditional AttributeChip tab stops, focus-trapped inventory item drawer

### LINKED ISSUE #1511: A11y: AttributeChip creates dead tab stops on tooltip-less surfaces; inventory item drawer hand-rolls a backdrop with no Escape/focus handling
Two accessibility regressions against the documented patterns:

## 1. `AttributeChip` is unconditionally `tabindex="0"`

`UI/src/components/AttributeChip.svelte:9,37` — the tab stop exists to fire the tooltip focus action, but `ClassPicker.svelte:37` renders up to six chips on the character-select/switcher surfaces, which have **no tooltip controller in context** — so keyboard users tab through six inert stops (inside the switcher's focus trap, no less). Fix: `tabindex={attrTip ? 0 : undefined}` so the stop exists only when it surfaces something.

## 2. Inventory item drawer: hand-rolled blocking backdrop

`UI/src/routes/game/screens/inventory/Inventory.svelte:21-35` — dims and pointer-blocks the whole screen behind a click-to-close-only backdrop: no Escape, no focus capture/trap/restore. This contradicts the documented "compose Popover / reuse focusTrap, don't hand-roll a backdrop" rule; the Skills screen's `SortFilterModal` shows the compliant shape.

ginderjeremiah:
## Summary

- **`AttributeChip`**: `tabindex` is now conditional (`attrTip ? 0 : undefined`) — the chip takes a tab stop only when a screen publishes an attribute-tooltip controller via `setAttributeTooltip`. On controller-less surfaces (the class picker on character select and the character switcher) the chips are purely presentational, so keyboard users no longer tab through up to six inert stops. `aria-describedby` was already omitted without a controller; the `svelte-ignore` stays (the compiler warns on the dynamic expression too) with the justification in the component comment.
- **Inventory item drawer**: replaces the click-only backdrop behavior with the documented overlay chrome by applying the shared `focusTrap` action to the drawer's `{#if}`-mounted content — Escape dismissal, Tab trap, focus capture on open (first focusable, mirroring `Popover`), and focus restore on close — plus `role="dialog"`/`aria-modal`/`aria-label`. Literal `Popover` composition would have broken the slide-over transform (the shell must stay mounted while closed), so this uses the sanctioned direct-`focusTrap` shape; documented in `docs/frontend-overlays.md`.
- **`focusTrap` stacking**: the drawer introduces the first real stacked-trap case (its weapon-swap confirm modal opens over the trapped drawer). Without a guard, both traps ran on every keydown — Escape double-dismissed and the outer trap stole focus from the modal on Tab. The action now keeps a mount-order stack and only the top-most trap handles keydown; benefits `ModalHost`/`Popover` layering generally.

## Test notes

- New/updated vitest coverage: `AttributeChip.test.ts` (no tab stop / no `aria-describedby` without a controller), `ClassPicker.test.ts` (renders the real chip now; asserts the fingerprint chips stay out of the tab order), `Inventory.test.ts` (drawer opens as a labelled dialog, captures focus, closes on Escape, restores focus, traps Tab), `focus-trap.test.ts` (stacked traps: top-most owns Tab/Escape, outer resumes on unmount).
- Ran: full `src/tests/routes` (160 files / 1704 tests), `src/tests/components` + inventory/attributes/attribute-breakdown suites (605 tests), all passing; `npm run check` (svelte-check, 0 errors/warnings) and eslint on the changed files.
- Frontend-only change; no backend or battle-logic surface touched.

Closes #1511


ginderjeremiah (commented):
## Review: A11y fixes — conditional AttributeChip tab stops + focus-trapped inventory item drawer (#1511)

**Verdict:** Approve — both issue items are fixed correctly against the documented a11y/overlay patterns, coverage is thorough, and no foot-guns are introduced. (Posted as a `COMMENT` because GitHub won't accept a self-approval on my own PR; treat this as an approve.)

I traced each change against the issue and the frontend docs:

- **`AttributeChip` conditional tab stop** — `tabindex={attrTip ? 0 : undefined}` is exactly the issue's fix: the stop exists only when a screen publishes a controller via `setAttributeTooltip`, so the class-picker/switcher fingerprint chips (no controller in context) stop being dead tab stops. `aria-describedby` was already omitted for a controller-less chip — `describedByTooltip` removes the attribute for an `undefined` id (`describedby-tooltip.ts:16-17`) — so the no-`aria-describedby` assertion holds. The `svelte-ignore a11y_no_noninteractive_tabindex` is still warranted (the compiler warns on the dynamic expression too) and the justification is in the component comment.
- **Inventory drawer → shared `focusTrap`** — replaces the click-only backdrop with the documented overlay chrome (Escape dismiss, Tab trap, focus restore, `role="dialog"`/`aria-modal`/`aria-label`). Applying the action to the `{#if}`-mounted *content* rather than composing `Popover` is the right call given the shell must stay mounted for the slide transform; the first-focusable capture `$effect` mirrors `Popover`, and the action captures `restoreFocus` on mount before the effect moves focus inward (verified by the focus-restore test). Documented in `frontend-overlays.md`.
- **`focusTrap` stacking guard** — the confirm modal at `inventory-view.svelte.ts:178` (weapon swap) genuinely opens over the trapped drawer, so this is a real stacked-trap case, not a speculative one. The mount-order `trapStack` + top-most-only keydown guard is a clean, general fix (benefits `ModalHost`/`Popover` layering too) and is a no-op for the single-trap path, so no regression.

Braces on every conditional, no null-forgiving operator, and the docs are updated where the patterns live.

### Test coverage

Solid and matches the project's rules — every touched surface gets a test:
- `AttributeChip.test.ts`: no tab stop / no `aria-describedby` without a controller, presentational render intact.
- `ClassPicker.test.ts`: now renders the *real* chip (not the stub) and asserts the fingerprint chips stay out of the tab order on the controller-less surface.
- `Inventory.test.ts`: drawer opens as a labelled dialog, captures focus, closes on Escape, restores focus, traps Tab.
- `focus-trap.test.ts`: stacked traps — top-most owns Tab/Escape, outer resumes on inner unmount.

No battle-logic surface touched, so no FE/BE parity mirror applies.

### What's good

- Reusing the shared `focusTrap` rather than hand-rolling drawer chrome, and turning the stacking requirement into a general primitive-level guard instead of a drawer-local hack.
- The controller-less-surface reasoning (a focusable stop that surfaces nothing is a dead stop) is captured in the component comment where a later edit is most likely to naively re-add the unconditional `tabindex`.

Nothing blocking or should-fix. Will merge once the firefox/webkit e2e jobs finish green (frontend, verify, and chromium e2e are already passing).

---

PR #1564, opened 2026-07-03
Fix save-window state bugs on Options/Attributes and adopt server-authoritative statPointsUsed

### LINKED ISSUE #1506: Save-window state bugs on Options and Attributes screens (mid-flight edits lost / stat-point budget stale)
Two screens mishandle state that changes while a save is in flight (or in the background):

## 1. Options: a toggle flipped while a save is in flight is marked clean but never persisted

`UI/src/routes/game/screens/options/options-view.svelte.ts:201-223` — `save()` snapshots `changedPreferences` at send time, but on success runs `applyToPlayer(this.draft)` and `baseline = { ...this.draft }` from the draft *as of response time*. Toggles stay enabled while `saving`, so a mid-flight flip is folded into the baseline as clean without ever being sent — the user sees "Preferences saved" while the server holds the old value, and the dirty tracking can never resurface it. Fix: merge only the actually-sent entries into the baseline (leaving later edits dirty), or disable the toggles while a save is in flight.

## 2. Attributes: the stat-point budget goes stale when the player levels up on-screen

`UI/src/routes/game/screens/attributes/attributes-view.svelte.ts:194-228` — `savedAvailable` is `$state` seeded in the constructor and after save, so stat points granted by a background level-up don't appear until remount — while the adjacent `LevelMeter` reads `playerManager.level/exp` live, so the player watches the level tick up with "points to spend" frozen. Safe to make it a `$derived` over the manager (`statPointsUsed` only changes through this screen's save). Related smaller window in the same file (`:324-346`): the steppers/radar stay active during the awaited save, and the post-save `syncFromPlayer()` silently wipes allocations made in that window — same fix family as the Options item.

### LINKED ISSUE #1548: Attributes screen should adopt server-authoritative statPointsUsed absolutely, not via relative += netSpent
Follow-up from PR #1546 (which made the victory path reconcile `statPointsUsed` from the server's authoritative `DefeatEnemy` response).

## Problem

PR #1546 newly writes `statPointsUsed` absolutely from the victory reconcile path. However, the attributes screen (`UI/src/routes/game/screens/attributes/attributes-view.svelte.ts`) still updates `statPointsUsed` with a *relative* `+= netSpent`. A relative-increment pattern is inherently fragile against an absolute reconcile happening concurrently.

It is not demonstrably buggy today under ordered single-socket delivery plus per-player serialization, but it is worth hardening so the stat-spend path aligns with the authoritative-reconcile pattern #1546 establishes.

## Suggested work

- Have the attributes screen adopt the server-authoritative `statPointsUsed` absolutely rather than via relative increment.
- Add coverage asserting the absolute adoption.

Low priority.

ginderjeremiah:
## Summary

Fixes the two in-flight-save state bugs from #1506 and hardens the stat-spend path per #1548, following the authoritative-reconcile pattern established by #1546.

**Options — mid-flight toggle no longer lost.** `save()` snapshotted the diff at send time but baselined the *response-time* draft, so a toggle flipped while the request was in flight was folded into the baseline as clean without ever reaching the server. It now merges only the entries actually sent into the baseline, and `applyToPlayer` likewise applies only the sent values (merged over the player's current preferences), so a mid-flight flip stays dirty and saveable.

**Attributes — live stat-point pool + preserved mid-save edits.** `savedAvailable` was `$state` seeded at mount/save, so a background level-up's new points stayed frozen until remount while the adjacent LevelMeter ticked up live; it is now `$derived` over the reactive player manager. The post-save `syncFromPlayer()` also silently wiped allocations made during the awaited save; the committed baseline now advances by exactly the deltas captured at send time, leaving mid-flight edits pending.

**Attributes — absolute `statPointsUsed` adoption.** The `UpdatePlayerStats` socket response is extended to `UpdatePlayerStatsResponse { Attributes, StatPointsUsed }` (regenerated client types included), returned on the rejection (`ErrorWithData`) path too. The client adopts the server's allocations and `StatPointsUsed` absolutely from any returned payload, replacing the relative `+= netSpent` that was fragile against the absolute victory reconcile from #1546.

## Test notes

- `UI`: `vitest run src/tests/routes/game/screens/attributes src/tests/routes/game/screens/options` — 156/156 pass, including new coverage for mid-flight toggle/allocation preservation, the live (derived) pool reacting to a background level-up, absolute `statPointsUsed` adoption, and the rejection-path reconcile. The attributes-view test's player-manager mock is now statified so the derived pool is exercised with real reactivity. `npm run check` (svelte-check) and eslint on changed files: clean.
- Backend: `dotnet test Game.Api.Tests -- --filter-class "*UpdatePlayerStatsSocketTests"` — 3/3 pass (now asserting `StatPointsUsed` on both the success and rejection paths); CodeGen namespace suite 230/230; `dotnet format --verify-no-changes` on changed files: clean. TypeScript codegen re-run, so the committed types match the drift check.

Closes #1506
Closes #1548


UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
```diff
@@ -327,21 +323,28 @@ export class AttributesView {
...
-		// Persist to the player manager so battles and other screens use the new
-		// allocation, then re-seed the baseline to clear the dirty state.
-		playerManager.attributes = response.data;
-		playerManager.statPointsUsed += netSpent;
-		this.syncFromPlayer();
+		// Advance the baseline by exactly the deltas that were sent, preserving any
+		// draft edits made while the save was in flight instead of wiping them (#1506).
+		this.committed = this.committed.map(
+			(value, i) => value + (updates.find((u) => u.attributeId === CORE_ATTRIBUTES[i])?.amount ?? 0)
+		);
```
ginderjeremiah, line 347: **Nit (non-blocking, take-it-or-leave-it).** A few lines up (line 334) the success path adopts the server's authoritative allocations onto `playerManager.attributes`, but here the baseline is re-derived by independent delta math (`committed_old + sent deltas`) rather than from that same authoritative payload. Under the documented per-player socket serialization the two always agree, so this is correct today — but it's two derivations of the same post-save state, and a later edit could let them drift (e.g. if the server ever normalized an allocation on save, `committed` would silently disagree with `playerManager.attributes`).

Since you've already written the authoritative allocations onto the manager, you can seed the baseline from that single source and still preserve mid-flight edits — re-seeding `committed` alone doesn't touch `draft`:

```ts
this.committed = CORE_ATTRIBUTES.map(readAllocation);
```

Same mid-flight-preservation behavior, one source of truth. Not blocking — the current code is correct and well-tested.

---

ginderjeremiah (commented):
## Review: Fix save-window state bugs on Options/Attributes + adopt server-authoritative `statPointsUsed` (#1506, #1548)

**Verdict: Approve** — _(submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a self-`APPROVE`; the intended verdict is **Approve**.)_ Both in-flight-save bugs from #1506 are fixed correctly, #1548's absolute-adoption hardening is done cleanly following the #1546 pattern, and CI is fully green. One non-blocking nit inline; not holding merge on it, but flagging so you can decide before it lands.

### Correctness — traced each fix

- **Options mid-flight toggle (#1506.1).** `changed` is snapshotted at send time, and both the baseline advance and `applyToPlayer` now key off *only the sent entries* (merged over the player's current prefs), so a toggle flipped while the request is in flight stays dirty and unapplied. The flip-back-mid-flight case resolves correctly too (server holds the sent value; the flip reads as a fresh pending change). Both paths are pinned by the two new tests.
- **Attributes live pool (#1506.2).** Converting `savedAvailable` to `$derived(playerManager.statPointsGained − statPointsUsed)` is the right call — `applyVictoryRewards` (player-manager.ts:218-219, from #1546) can now change `statPointsUsed`/`statPointsGained` in the background, so a `$state` snapshot was genuinely stale. Verified the derived pool can only grow mid-edit under normal flow, so it can't retroactively push the draft over budget.
- **Attributes mid-save edits (#1506.2).** Dropping `syncFromPlayer()` (which did `draft = [...committed]`) in favor of advancing `committed` by the sent deltas preserves allocations made during the awaited save. Confirmed no dangling `syncFromPlayer` callers remain (the only refs are in the unrelated `skills-view`).
- **Absolute `statPointsUsed` (#1548).** Adopting `response.data.statPointsUsed` absolutely from any payload — including the `ErrorWithData` rejection path — replaces the fragile `+= netSpent`. Backend returns the authoritative post-command state on both outcomes (unchanged on rejection), and the client reconciles onto it while keeping the draft dirty for retry.

### Design / layering

- Backend: `UpdatePlayerStatsResponse` is an API-layer response DTO (`IModel`, `required` members) built from the domain `player.StatPoints` in the socket command — the same presentation-mapping shape the pre-existing inline `BattlerAttribute.From(...)` already used, so no new DDD/layering violation. Returning the authoritative state on the error path is a sound, explicit contract.
- Frontend: statifying the test's `PlayerManager` mock (class instance, not object literal) to exercise the `$derived` pool with real reactivity is the correct pattern per frontend.md, and mirrors the real statified manager. Generated types (`IUpdatePlayerStatsResponse`, the socket type map) are codegen output, regenerated rather than hand-edited.

### Test coverage

Thorough and matches the project's rules. Frontend: mid-flight toggle preservation + flip-back (Options), mid-flight allocation preservation, the live derived pool reacting to a background level-up, absolute-adoption (#1548), and the rejection-path reconcile (Attributes) — all mirroring the source path under `tests/`. Backend: `UpdatePlayerStatsSocketTests` now asserts `StatPointsUsed` on the success and both rejection paths (Testcontainers integration). No battle-logic surface, so no frontend/backend parity mirror applies.

### Nits

- One inline on `attributes-view.svelte.ts:345` — the success-path baseline is re-derived by delta math while the authoritative allocations are already adopted onto the manager two lines above; seeding `committed` from that single source would preserve the same mid-flight behavior with one source of truth. Non-blocking.

### What's good

Genuinely clean fixes for three related save-window defects, each with a focused test that would fail on the old behavior. The authoritative-reconcile pattern is now consistent across the victory and stat-spend paths, and the rejection path carrying unchanged authoritative state is a nice touch that makes the client's reconcile uniform.

---

PR #1566, opened 2026-07-03
Fix Codex correctness gaps: retired challenges, retired deep-links, swallowed load errors

### LINKED ISSUE #1507: Codex correctness: retired uncompleted challenges shown in dossiers, deep-links to retired records open the wrong dossier, load failures swallowed
Three related gaps in the Codex screen (`UI/src/routes/game/screens/codex/`):

## 1. Enemy dossiers show retired uncompleted challenges the Challenges screen deliberately hides

`codex-view.svelte.ts:463-487` filters enemy-scoped challenges with no `retiredAt` check, while `challenges-view.svelte.ts:299-305` hides retired-unless-completed (the documented retirement rule). A retired incomplete challenge shows as `sealed`/`N/goal` forever in the dossier — and its presence alone adds the Challenges sub-tab. Apply the same retired-unless-completed filter.

## 2. Deep-links/cross-links to a retired record silently open the wrong (first-in-list) dossier

The dossier resolvers filter retired records and then fall back to head-of-list (`codex-view.svelte.ts:375-382, 779-794`), but link producers still emit links to retired entities: Statistics rows (`stats/statistics-view.svelte.ts:147-154, 358-366`), enemy skill-pool rows (`codex-view.svelte.ts:404-418`), and the zone boss card (`:542-549`). Clicking such a row lands on an unrelated dossier with no feedback. Either resolve an explicitly-requested retired id (retired ids stay resolvable by design — that's the retirement contract) or make those rows non-interactive; resolving the id seems more consistent with "already-earned references never break".

## 3. Codex swallows load failures the other screens toast

`Codex.svelte:36-43` never calls `toastError` and never reads `playerChallenges.error`, so a failed challenge-progress load renders as authoritative "Locked/Sealed/zero progress" — exactly the error-collapsed-into-friendly-state the shared rule (frontend-screens.md) forbids. Statistics and Challenges both do this correctly; mirror them.

ginderjeremiah:
## Summary

Fixes three related correctness gaps in the Codex screen (`UI/src/routes/game/screens/codex/`) described in #1507:

1. **Enemy dossiers showed retired-but-uncompleted challenges forever.** `codex-view.svelte.ts`'s enemy-scoped `challenges` derivation had no `retiredAt` check, unlike the Challenges screen's documented retired-unless-completed rule. A retired incomplete challenge now hides from the dossier (still shows if the player already completed it before it retired, since that reward was already earned).
2. **Deep-links/cross-links into a retired record silently opened the wrong dossier.** `selectedEnemy`/`selectedZone`/`selectedSkill` (and the constructor's deep-link resolvers) resolved an id against the *live-filtered* catalogue, so a retired-record reference (a Statistics row, an enemy's skill-pool row, a zone boss card) fell back to the head of the list instead of opening the actual record. These now resolve an explicitly-requested id against the full catalogue first (retirement keeps a record's slot resolvable by id, per `backend.md`'s Reference Data convention), falling back to the head of the live list only when there's no explicit id.
3. **Codex swallowed challenge-progress load failures.** `Codex.svelte` never read `playerChallenges.error` or called `toastError`, so a failed load rendered every enemy-scoped challenge as authoritative "sealed"/zero-progress — the exact error-collapsed-into-friendly-state pattern `frontend-screens.md` forbids elsewhere (Statistics/Challenges both toast on this). Codex now toasts on failure, and the enemy dossier's Challenges sub-tab shows an inline "could not be loaded" note instead of the (potentially fake) progress list.

## Scope note

The zone-unlock "sealed" indicator and the zone rail's locked/unlocked status also read `playerChallenges` and would show a stale-but-safe (fail-closed) value during the same transient failure. I left those as-is rather than introducing a tri-state status, since that's a larger structural change than this bug fix calls for — the toast at least surfaces that the state may be stale. Happy to open a follow-up if a tri-state is wanted there.

## Test plan

- [x] Added/updated unit tests in `codex-view.svelte.test.ts` (retired-challenge filtering; retired-record resolution for enemy/zone/skill via deep-link payload and `selectEnemy`/`selectZone`/`selectSkill`)
- [x] Added component tests in `Codex.test.ts` (challenge-load-error toast + inline note; boss-card cross-link into a retired enemy's dossier)
- [x] `npx vitest run` — 281 files / 3295 tests pass

Closes #1507

---
_Generated by [Claude Code](https://claude.ai/code/session_011NAsQBz4CjmUMPtnTmJ7fJ)_

### DIFF
--- a/UI/src/routes/game/screens/codex/Codex.svelte
+++ b/UI/src/routes/game/screens/codex/Codex.svelte
@@ -23,7 +23,7 @@
 
 <script lang="ts">
 import { onMount } from 'svelte';
-import { navigation, playerChallenges, statistics } from '$stores';
+import { navigation, playerChallenges, statistics, toastError } from '$stores';
 import { CodexView, type CodexNavPayload } from './codex-view.svelte';
 import CodexTabBar from './CodexTabBar.svelte';
 import EnemiesTab from './EnemiesTab.svelte';
@@ -40,6 +40,12 @@ onMount(async () => {
 	view.stats = statistics.stats;
 	view.statsError = statistics.error;
 	view.statsLoading = false;
+	view.challengesError = playerChallenges.error;
+	if (playerChallenges.error) {
+		// Don't conflate a failed load with genuine sealed/zero-progress challenges — a dropped fetch
+		// would otherwise render every enemy-scoped challenge as if it had no progress.
+		toastError('Your challenge progress could not be loaded. Please try again later.');
+	}
 });
 </script>
 
--- a/UI/src/routes/game/screens/codex/EnemyChallengesPanel.svelte
+++ b/UI/src/routes/game/screens/codex/EnemyChallengesPanel.svelte
@@ -1,17 +1,21 @@
 <!-- Challenges sub-tab: the challenges scoped to this enemy, with the player's progress. -->
 <div class="challenges">
 	<div class="label">Related challenges · {view.challenges.length}</div>
-	<div class="list">
-		{#each view.challenges as challenge (challenge.id)}
-			<div class="challenge" style:--ch-accent={challenge.accent}>
-				<div class="text">
-					<div class="name">{challenge.name}</div>
-					<div class="type">{challenge.typeLabel}</div>
+	{#if view.challengesError}
+		<div class="note">Your challenge progress could not be loaded.</div>
+	{:else}
+		<div class="list">
+			{#each view.challenges as challenge (challenge.id)}
+				<div class="challenge" style:--ch-accent={challenge.accent}>
+					<div class="text">
+						<div class="name">{challenge.name}</div>
+						<div class="type">{challenge.typeLabel}</div>
+					</div>
+					<span class="prog" class:done={challenge.completed}>{challenge.progressText}</span>
 				</div>
-				<span class="prog" class:done={challenge.completed}>{challenge.progressText}</span>
-			</div>
-		{/each}
-	</div>
+			{/each}
+		</div>
+	{/if}
 </div>
 
 <script lang="ts">
@@ -81,5 +85,12 @@ let { view }: Props = $props();
 	&.done {
 		color: var(--success);
 	}
+}
+
+.note {
+	font-size: 12px;
+	color: var(--text-tertiary);
+	font-style: italic;
+	line-height: 1.5;
 }
 </style>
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: Codex retired-record correctness + swallowed challenge-load errors (Closes #1507)

**Verdict:** Approve _(submitted as a `COMMENT` because GitHub won't accept a gating review from the PR author; treat this as an approval)_ — all three gaps from #1507 are addressed correctly, the change moves the code **toward** the documented conventions rather than away, and the tests are thorough.

### Why this is correct

- **Retired-challenge filter (gap 1)** — `codex-view.svelte.ts:481` is a faithful mirror of the Challenges screen's rule (`challenges-view.svelte.ts:303`): `c.retiredAt == null || progressById.get(c.id)?.completed`, same `== null` handling, same completed-stays-visible intent. Good that the source-of-truth rule is duplicated verbatim rather than paraphrased.
- **Explicit-id resolution (gap 2)** — switching `selectedEnemy`/`selectedZone`/`selectedSkill` and the `resolve*` helpers from `.find(e => e.id === id)` to `staticData.<set>[id]` is not just a bug fix, it aligns with `frontend.md` → _Reference Data_ ("index by id for the zero-based-id catalogues … don't mix in `.find(e => e.id === id)`"). Retirement preserves the Id-as-index slot, so a retired id stays resolvable while the live-filtered lists still drive display. Correct on both counts.
- **Swallowed load errors (gap 3)** — `Codex.svelte:43-48` reads `playerChallenges.error` only after `await`-ing the load, and the store's `error` flag is authoritative (set true on failure, reset on success), so the toast + inline note can't misfire. Matches the fail-loud pattern Statistics/Challenges already use.

- `codex-view.svelte.ts:387` — the `selectedEnemy` no-explicit-id fallback still uses `this.enemies.find((e) => e.id === firstVisible.id)`. It's pre-existing (this PR didn't introduce it) and safe since `firstVisible` is always live, but given the PR's whole theme is adopting index-by-id, it'd be consistent to resolve it as `staticData.enemies[firstVisible.id]` too. Fine to leave for a follow-up.
- `EnemyChallengesPanel.svelte:3` — the `Related challenges · {view.challenges.length}` count still renders above the error note on a failed load. Minor; the authored count is arguably still meaningful, and the note makes the state clear.

### Test coverage

Solid and at the right levels: view-model unit tests for the retired-challenge filter and retired-record resolution (deep-link payload **and** `selectEnemy`/`selectZone`/`selectSkill`) in `codex-view.svelte.test.ts`; component tests for the load-error toast + inline note (asserting no fake progress renders) and the retired-boss cross-link in `Codex.test.ts`. No battle logic touched, so no FE/BE parity mirror needed.

### What's good

Tight, well-scoped fix. The scope note on deliberately leaving the zone-unlock "sealed" indicator fail-closed (rather than expanding into a tri-state) is exactly the right call for a bug fix, and the inline comments explain the retirement-contract reasoning without over-documenting.


---

PR #1567, opened 2026-07-03
Fix Workbench editor state bugs: phantom dirty, save-selection jump, dirty-token color

### LINKED ISSUE #1509: Workbench editor state bugs: phantom dirty from undefined≠null, post-save selection jump to first record, dirty-dot styled with --warning
Three verified state/styling bugs in the admin Workbench:

## 1. Phantom per-field dirty state from `undefined !== null`

`ScopeField.svelte:71` and `ChallengeRewardSection.svelte:119-120` compare optional values with raw `!==`, but server baselines carry `null` while local edits write `undefined` — the field shows permanently dirty while the save bar (via null-collapsing `canonicalEqual`) simultaneously says "No unsaved changes". `FieldControl.svelte:146` already uses the right primitive (`recordsEqual`); these call sites should too.

## 2. After saving a newly-added record, selection jumps to the first record

`Workbench.svelte:68,80-84` — `selId` still holds the temporary negative id after the save re-seeds from the server; `find(...) ?? items[0]` then yanks the admin off the record they just created. `persistEntity` (`save-helpers.ts:110-118`) computes the local→persisted id map and **discards it**; the progression editor follows its remap correctly (`progression-store.svelte.ts:436-441`) — the generic path should do the same.

## 3. Dirty state styled with the validation token (documented dot/triangle separation)

`ScopeField.svelte:20` styles a *dirty* scope with the validation `--warning` label class (and renders no DirtyDot); `RewardSlot.svelte:60-68`'s dirty dot is `var(--warning)` instead of `--change-modified`. frontend.md is explicit that dirty (`--change-*` + dot) and validation (`--warning` + triangle) never overlap. Nit while there: the progression `DetailHeader.svelte:43` tab-dot lacks the `sr-only` "Unsaved changes" text its Workbench counterpart has.

ginderjeremiah:
## Summary

Fixes the three verified admin-Workbench state/styling bugs from #1509:

1. **Phantom per-field dirty state from `undefined !== null`.** `ScopeField.svelte` and `ChallengeRewardSection.svelte` compared optional fields (`targetEntityId`, `rewardItemId`, `rewardItemModId`) with raw `!==`, but a server baseline serializes an unset optional as `null` while a local clear (`setGlobal`, `setItem(undefined)`) leaves the field simply absent — so the field showed dirty while the save bar (via the null-collapsing `canonicalEqual`/`recordsEqual`) simultaneously said "No unsaved changes". Both now use `recordsEqual`, matching `FieldControl.svelte`'s existing convention.

2. **Post-save selection jump to the first record.** `Workbench.svelte`'s `selId` still held a just-saved record's temporary negative id after the save re-seeded from the server, so `items.find(...)` failed and the fallback yanked the admin to the first record instead of the one they just created. `persistEntity` computed this local→persisted id map internally and discarded it — the progression editor already followed its own copy of this remap correctly (`resolveNewIds`/`resolveId` in `progression-helpers.ts`).

   Rather than threading the map through every entity's `persist()` return type, I moved `resolveNewIds`/`resolveId` into the shared `save-helpers.ts` (deduping the two identical copies), had `persistEntity` use it, and added `EntityStore.lastIdMap` — populated from the same diff/fresh-list data the store already has after every save. `Workbench.svelte`'s selection-tracking effect now consults `lastIdMap` before falling back to the head of the list.

3. **Dirty state styled with the validation token.** `ScopeField.svelte` styled a dirty scope with the `--warning` label class (and rendered no `DirtyDot`), and `RewardSlot.svelte`'s dirty dot used `var(--warning)` instead of `--change-modified` — frontend.md is explicit that dirty (`--change-*` + dot) and validation (`--warning` + triangle) never overlap. Fixed both. Also picked up the noted nit: the progression `DetailHeader.svelte` tab-dot was missing the `sr-only` "Unsaved changes" text its Workbench counterpart has.

## Test plan

- [x] Added regression tests for the phantom-dirty case (local `undefined` vs. baseline `null`) in `ScopeField.test.ts` and `ChallengeRewardSection.test.ts`
- [x] Added `EntityStore` tests for `lastIdMap` (maps a new record's temp id to its persisted id; replaced wholesale each save)
- [x] Added a `Workbench.svelte` integration test driving a full add → save → verify-selection-follows-remap flow
- [x] Added `DetailHeader.test.ts` covering the sr-only dirty-tab text
- [x] Moved the `resolveNewIds`/`resolveId` unit tests from `progression-helpers.test.ts` to `save-helpers.test.ts` alongside their new home
- [x] `npx vitest run` — 282 files / 3294 tests pass

Closes #1509

---
_Generated by [Claude Code](https://claude.ai/code/session_01QYksGvCL3BNud4NRbHUqLG)_

### DIFF
--- a/UI/src/routes/admin/workbench/Workbench.svelte
+++ b/UI/src/routes/admin/workbench/Workbench.svelte
@@ -76,9 +76,17 @@ const flagged = $derived.by(() => {
 	return s ? liveItems.filter((it) => s.stateOf(it).warnings.length > 0).length : 0;
 });
 
-// If the selected record was removed (e.g. deleted then saved), fall back to the first.
+// After a save, a selected-but-unsaved record's temporary negative id no longer matches any
+// item — follow it to its persisted id first, and only fall back to the first record (e.g.
+// the selection was actually deleted then saved) when no remap exists for it.
 $effect(() => {
-	if (store && selected && selected.id !== selId) {
+	if (!store) {
+		return;
+	}
+	const persistedId = store.lastIdMap.get(selId);
+	if (persistedId !== undefined) {
+		selId = persistedId;
+	} else if (selected && selected.id !== selId) {
 		selId = selected.id;
 	}
 });
--- a/UI/src/routes/admin/workbench/components/challenge/ChallengeRewardSection.svelte
+++ b/UI/src/routes/admin/workbench/components/challenge/ChallengeRewardSection.svelte
@@ -85,6 +85,7 @@ import { ERarity, type IChallenge } from '$lib/api';
 import { Popover } from '$components';
 import { reference } from '../../reference.svelte';
 import type { EntityStore } from '../../entity-store.svelte';
+import { recordsEqual } from '../../entity-store.svelte';
 import type { Identified } from '../../entities/types';
 import { claimedItemMap, claimedModMap } from '../../entities/challenge-helpers';
 import WorkbenchIcon from '../../WorkbenchIcon.svelte';
@@ -116,8 +117,8 @@ const liveChallenges = $derived(store.items.filter((it) => store.status(it) !== 
 const claimedItems = $derived(claimedItemMap(liveChallenges, challenge.id));
 const claimedMods = $derived(claimedModMap(liveChallenges, challenge.id));
 
-const itemDirty = $derived(base ? challenge.rewardItemId !== base.rewardItemId : false);
-const modDirty = $derived(base ? challenge.rewardItemModId !== base.rewardItemModId : false);
+const itemDirty = $derived(base ? !recordsEqual(challenge.rewardItemId, base.rewardItemId) : false);
+const modDirty = $derived(base ? !recordsEqual(challenge.rewardItemModId, base.rewardItemModId) : false);
 const none = $derived(challenge.rewardItemId == null && challenge.rewardItemModId == null);
 
 // Retired records drop out of the pick list (can't be newly granted) unless they're the
--- a/UI/src/routes/admin/workbench/components/challenge/RewardSlot.svelte
+++ b/UI/src/routes/admin/workbench/components/challenge/RewardSlot.svelte
@@ -63,8 +63,8 @@ const { kind, label, valueId, name, sub, color, retired = false, dirty, open, on
 	height: 5px;
 	margin-left: 7px;
 	border-radius: 50%;
-	background: var(--warning);
-	box-shadow: 0 0 5px var(--warning);
+	background: var(--change-modified);
+	box-shadow: 0 0 5px var(--change-modified);
 }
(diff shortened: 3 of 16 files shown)


ginderjeremiah (commented):
## Review: Workbench editor state bugs (phantom dirty, save-selection jump, dirty-token color) — closes #1509

**Verdict:** Approve _(submitted as a `COMMENT` since I'm the PR author and GitHub won't accept a self-`APPROVE`)_ — all three bugs from #1509 are fixed correctly, cleanly, and with matching tests.

### Correctness

- **Phantom dirty (`undefined ≠ null`)** — `ScopeField` and `ChallengeRewardSection` now compare through `recordsEqual`/`canonicalEqual`, whose `canonicalize` collapses `undefined`/`null`/absent to one state. I verified the scalar edge: `recordsEqual(undefined, null)` → `true` (no phantom dirty) while `recordsEqual(0, null)` → `false`, so a real target/reward id of `0` (a valid zero-based id) stays correctly distinct from unset. This aligns the field-level dirty check with the null-collapsing save bar, matching `FieldControl.svelte`'s convention as the issue asked.
- **Post-save selection jump** — the `lastIdMap` remap is sound. New records get monotonically-decreasing negative ids and persisted ids are non-negative, so a stale map entry (`-1 → 3`) can never collide with a future `selId`; the effect converges after re-running once (the persisted id isn't a map key). The genuine delete-then-save case still falls back to the head of the list, preserving the old behavior. Good call moving `resolveNewIds`/`resolveId` into `save-helpers` to dedupe the two identical copies rather than threading the map through every `persist()` return type — confirmed no dangling importers remain.
- **Dirty styled with the validation token** — `ScopeField` now renders `DirtyDot` (dropping the `--warning` `.lbl` class), `RewardSlot`'s dot uses `--change-modified`, and the progression `DetailHeader` tab-dot picks up the `sr-only` "Unsaved changes" text. This restores the dot/`--change-*` vs. triangle/`--warning` separation `frontend.md` requires, and the a11y text keeps the change state from being conveyed by color/shape alone.

### Test coverage

Thorough and matched to the fixes: phantom-dirty vs. genuinely-dirty cases in both `ScopeField.test.ts` and `ChallengeRewardSection.test.ts`; `EntityStore.lastIdMap` (temp→persisted mapping + wholesale replacement on a later no-add save); a full add → save → selection-follows-remap integration test in `Workbench.test.ts`; the `DetailHeader` sr-only assertion; and the `resolveNewIds`/`resolveId` unit tests moved alongside their new home in `save-helpers.test.ts`.

### What's good

Tight scope, each change traces directly to a documented rule, and the refactor leaves the code DRYer than it found it. The scope note deferring a Codex-style tri-state is out of scope here and correctly left alone. CI is green. Merging.


---

PR #1568, opened 2026-07-03
Backend dead-code & stale-comment sweep

### LINKED ISSUE #1514: Backend dead-code & stale-comment sweep (unused cache/pub-sub surface, test-only modifier path documented as live, misc)
Verified-unused production code and actively-misleading stale comments, grouped per the established cleanup-basket pattern (#937, #1192). All usages verified by repo-wide reference search.

## Dead cache/queue/pub-sub surface (only tests/fakes reference them)

- `ICacheService.SetNotExists` (`Game.Abstractions/Infrastructure/ICacheService.cs:46`, `RedisService.cs:96-99`) — also the **only setter that writes a key with no TTL**, violating the every-key-carries-a-TTL invariant; worth deleting before someone reaches for it.
- `ICacheService.Expire` awaited overload (`ICacheService.cs:33`, `RedisService.cs:76-79`) — only `ExpireAndForget` is used.
- `IPubSubQueue` sync methods `GetNext()/GetNext<T>()/AddToQueue/AddToQueue<T>` (`IPubSubQueue.cs:5-6,50-51`, `RedisQueue.cs:35-49,149-158`).
- `IPubSubService.UnSubscribe(string channel)` no-id overload (`RedisPubSubService.cs:175-178`) — unused **and** hazardous: it unsubscribes every handler on the channel, bypassing id-scoped worker disposal.
- `IPubSubService.Subscribe(channel, queueName, Action<...>, id)` sync-action worker overload (`RedisPubSubService.cs:117-129`) — all production subscribers use the `Func<Task>` overload.

## Dead repository/domain members

- `Enemies.GetRandomEnemy` + `IEnemyEntityCache.GetRandomEnemy` (`Enemies.cs:27-31`) — no caller.
- `EntityStore.Track` + `IEntityStore.Track` (`EntityStore.cs:44-47`) — no caller.
- `Zones.GetZone` + `IZones.GetZone` (`Zones.cs:23-29`) — referenced only by its own integration tests.
- `SkillRecipes.RecipesForInputSkill` + the snapshot's `RecipeIdsByInputSkill` reverse index (`SkillRecipes.cs:35-40`, `SkillRecipesCacheHolder.cs:25,60-68`) — rebuilt on every cache reload, never consumed (the client's hinted reveal reads full recipes via `GetSkillRecipes`). Caveat: possibly a deliberate seam from spike #1125 — confirm before deleting.
- `IModelToSource<TSource>` and `IMappedModel<TModel,TEntity>` (`Game.Api/Models/IModel.cs:8-16`) — zero implementations solution-wide.

## Test-only modifier path documented as the live path (footgun)

`Player.GetAllModifiers()` / `Player.GetAttributes()` / `Inventory.GetEquippedAttributeModifiers()` (`Player.cs:455-464`, `Inventory.cs:58-78`) are called only by `BattlerFactory` and unit tests — every production composition routes through `BattleSnapshot.GetModifiers*`. The danger: `GetAllModifiers` composes a **lesser** set (no class locked base, no proficiency bonuses, no signature passive), so a future production caller silently gets pre-class-system attributes. Move the chain into `Game.Core.TestInfrastructure` (or clearly mark test-only), and fix the three stale doc comments that still call it the live path: `BattleSnapshot.cs:109`, `DefeatRewards.cs:79`, `Item.cs:73-77`. Same pattern in miniature: `AttributeCollection.AllModifiers()` (`AttributeCollection.cs:105`).

## Stale model descriptions

- `ServerGameConstants.ProficiencyXpPerVictory` XML doc (`ServerGameConstants.cs:30-38`) still describes the retired pre-#1318 model ("scaled by the DefeatRewards difficulty curve … split across the proficiencies … diversifying dilutes"). All three claims are now the opposite (independent overlapping claims, difficulty multiplier deliberately **not** applied, multi-axis mints more). Anyone tuning the constant from its own doc reasons from the wrong model.
- `BattleService.RecordVictory` comment (`BattleService.cs:658-664`) claims `EndBattleVictory` "validates the *client-supplied* claimed timestamp"; the code is deliberately server-clock-only.

## Misc

- `GameContext` is hand-wired per scope, rebuilding `DbContextOptions` on every resolution — including the scope-per-event drain path (`ServiceCollectionExtensions.cs:22-30`, `GameContextFactory.cs:9-34`). No correctness issue (EF's internal provider is cached), but the standard `AddDbContext`/`AddDbContextPool` shape builds options once and enables pooling on the drain hot path.

ginderjeremiah:
## Summary

Implements #1514 — a backend dead-code & stale-comment cleanup basket (same pattern as #937/#1192). I independently re-verified every claim in the issue via repo-wide reference search before deleting anything (per CLAUDE.md's "don't trust the issue, verify" guidance) and found one discrepancy, noted below.

### Dead cache/queue/pub-sub surface removed
- `ICacheService.SetNotExists` — also the only setter that wrote a key with no TTL.
- `ICacheService.Expire` (awaited overload) — only `ExpireAndForget` was used.
- `IPubSubQueue`'s sync `GetNext()`/`GetNext<T>()`/`AddToQueue(string)`/`AddToQueue<T>()`.
- `IPubSubService.Subscribe(channel, queueName, Action<...>, id)` (the sync-Action worker overload) — no production caller (both real call sites use the `Func` overload).
- `IPubSubService.UnSubscribe(string channel)` (no-id overload) — **the issue claimed this was unused, but it wasn't**: `AdminCacheReloadBroadcastTests.cs` called it for cleanup. I switched that test to the safer id-scoped `Subscribe`/`UnSubscribe(channel, id)` pattern every other test in the suite already uses, then removed the hazardous unsubscribe-every-handler-on-the-channel method.

### Dead repository/domain members removed
- `Enemies.GetRandomEnemy` + `IEnemyEntityCache.GetRandomEnemy` — its dedicated integration tests (invalid zone, empty zone, retired-exclusion) were ported onto `GetRandomDomainEnemy`, which shares the same underlying zone-validation/spawn-table logic, so that edge-case coverage isn't lost.
- `EntityStore.Track` + `IEntityStore.Track`.
- `Zones.GetZone` + `IZones.GetZone` — its field-mapping assertions were redundant with `ZoneMapperTests` (already covers the full contract, including `IsHome`), so those tests were removed rather than ported; the bounds-check behavior stays pinned via `GetDomainZone`'s tests (same shared validation helper).
- `SkillRecipes.RecipesForInputSkill` + the cache's `RecipeIdsByInputSkill` reverse index — the issue flagged this as a possible deliberate seam from spike #1125, so I checked: the spike's design doc says the reverse index was meant to drive the client's hinted reveal, but the shipped implementation (`GetSkillRecipes` command) sends the full recipe set and the client derives hints itself — the reverse index was never actually wired to anything. Confirmed dead.
- `IModelToSource<TSource>` / `IMappedModel<TModel, TEntity>` — zero implementations solution-wide (`IModelFromSource`, which many models do implement, is untouched).

### Test-only modifier path (footgun) fixed
`Player.GetAllModifiers()`/`GetAttributes()` and `Inventory.GetEquippedAttributeModifiers()` were called only by `BattlerFactory` (test infra) and unit tests — every production battler composition goes through `BattleSnapshot.GetModifiers*`, which additionally layers the class locked base, proficiency bonuses, and signature passive. Moved the chain into `Game.Core.TestInfrastructure` (`PlayerAttributeTestExtensions`) as extension methods so a future production caller can no longer reach for the lesser composition by accident. `AttributeCollection.AllModifiers()` has the same shape but needs the collection's private per-attribute node storage, so it couldn't move without adding new public surface for a test-only accessor — left in place with a doc comment flagging it as test-only instead.

Fixed the stale doc comments this uncovered/it referenced:
- `BattleSnapshot.cs`, `DefeatRewards.cs`, `Item.cs` — no longer describe the moved methods as "the live path."
- `ServerGameConstants.ProficiencyXpPerVictory` — was still describing the retired pre-#1318 band/dilution model; now describes the actual independent-overlapping-claims, no-difficulty-multiplier, multi-axis-mints-more model.
- `BattleService.RecordVictory` — claimed `EndBattleVictory` validates a *client-supplied* timestamp; it's actually server-clock-only (matches the accurate comment a few lines above it in the same file).

### Misc
`GameContext`'s DI registration switched from a per-resolution factory (rebuilding `DbContextOptions` from scratch on every scope, including the write-behind drain loop's scope-per-event path) to `AddDbContextPool`, which builds options once at startup and pools `GameContext` instances. `GameContext` has no per-scope state beyond `DbContextOptions`, so it's safe to pool. `GameContextFactory.GetGameContext` (used directly by a couple of unit tests) is unchanged in behavior — the options-building logic was extracted into a shared `Configure` method both paths call.

## Test plan
- [x] `dotnet test --no-build --no-restore` — full suite green (2791 tests: Core 1231, Infrastructure 59, Application 824, Api 677), including the integration suites that exercise `GameContext` through the new pooled DI registration
- [x] Repo-wide reference search on every deleted member before removal to confirm no production caller

Closes #1514

---
_Generated by [Claude Code](https://claude.ai/code/session_01UpjjF8hpvJkG26UxR9RJCA)_

### DIFF
--- a/Game.Abstractions/DataAccess/ISkillRecipes.cs
+++ b/Game.Abstractions/DataAccess/ISkillRecipes.cs
@@ -5,8 +5,9 @@ namespace Game.Abstractions.DataAccess
 {
     /// <summary>
     /// Read access to the cached skill-synthesis recipe catalogue (spike #1125): the contract projection the
-    /// client and admin Workbench read, the lean domain model the synthesis command validates against, and the
-    /// input-skill → recipe-ids reverse index that drives the client's hinted reveal.
+    /// client and admin Workbench read, and the lean domain model the synthesis command validates against. The
+    /// client derives its own hinted-reveal state from the full recipe set (<see cref="AllSkillRecipes"/>) plus
+    /// owned skills/proficiencies rather than a server-computed reverse index.
     /// </summary>
     public interface ISkillRecipes
     {
@@ -19,10 +20,6 @@ namespace Game.Abstractions.DataAccess
         public bool ValidateRecipeId(int recipeId);
 
         public CoreSkillRecipe GetSkillRecipe(int recipeId);
-
-        /// <summary>The (non-retired) recipes the given skill is an input to, or empty if none — the reverse
-        /// index the client hint state consumes when the player owns the skill.</summary>
-        public IReadOnlyList<int> RecipesForInputSkill(int skillId);
 
         /// <inheritdoc cref="IItems.VersionKey"/>
         public object VersionKey { get; }
--- a/Game.Abstractions/DataAccess/IZones.cs
+++ b/Game.Abstractions/DataAccess/IZones.cs
@@ -6,8 +6,6 @@ namespace Game.Abstractions.DataAccess
     public interface IZones
     {
         public List<Contracts.Zone> All();
-        // Returns the read contract for a single zone; throws if the id is out of range.
-        public Contracts.Zone GetZone(int zoneId);
 
         /// <summary>Returns the lean gameplay <see cref="CoreZone"/> domain model for a single zone (battle
         /// setup); throws if the id is out of range.</summary>
--- a/Game.Abstractions/Infrastructure/ICacheService.cs
+++ b/Game.Abstractions/Infrastructure/ICacheService.cs
@@ -16,9 +16,8 @@
         /// as its TTL and returns the previous value (null if the key was unset). The value is required
         /// (non-null): writing a TTL implies writing a value, so unlike <see cref="Set(string, string?,
         /// TimeSpan, CancellationToken)"/> there is no null-means-delete path. Writing the value and its expiry
-        /// in a single operation (rather than a separate set followed by
-        /// <see cref="Expire(string, TimeSpan, CancellationToken)"/>) means a fault between the two can never
-        /// leave the key lingering without a TTL.
+        /// in a single operation (rather than a separate set followed by a TTL reset) means a fault between the
+        /// two can never leave the key lingering without a TTL.
         /// </summary>
         public Task<string?> GetSet(string key, string value, TimeSpan expiry, CancellationToken cancellationToken = default);
         /// <summary>
@@ -28,8 +27,6 @@
         /// </summary>
         public Task Set(string key, string? value, TimeSpan expiry, CancellationToken cancellationToken = default);
         public Task Set<T>(string key, T value, TimeSpan expiry, CancellationToken cancellationToken = default);
-        /// <summary>Resets the time-to-live on an existing key. A no-op if the key does not exist.</summary>
-        public Task Expire(string key, TimeSpan expiry, CancellationToken cancellationToken = default);
         /// <summary>
         /// Resets the time-to-live on an existing key without awaiting the result (fire-and-forget).
         /// A no-op if the key does not exist. Lets a hot read path slide a sliding-expiration TTL
@@ -43,7 +40,6 @@
         /// </summary>
         public void SetAndForget(string key, string? value, TimeSpan expiry);
         public void SetAndForget<T>(string key, T value, TimeSpan expiry);
-        public Task SetNotExists(string key, string value, CancellationToken cancellationToken = default);
         public Task Delete(string key, CancellationToken cancellationToken = default);
         public void DeleteAndForget(string key);
         public Task CompareAndDelete(string key, string deleteIfValue, CancellationToken cancellationToken = default);
--- a/Game.Abstractions/Infrastructure/IPubSubQueue.cs
+++ b/Game.Abstractions/Infrastructure/IPubSubQueue.cs
@@ -2,8 +2,6 @@
 {
     public interface IPubSubQueue
     {
-        public string? GetNext();
-        public T? GetNext<T>();
         public Task<string?> GetNextAsync(CancellationToken cancellationToken = default);
         public Task<T?> GetNextAsync<T>(CancellationToken cancellationToken = default);
 
@@ -47,8 +45,6 @@
         /// </summary>
         public Task<bool> RemoveAsync(string value, CancellationToken cancellationToken = default);
 
-        public void AddToQueue(string value);
-        public void AddToQueue<T>(T value);
         public Task AddToQueueAsync(string value, CancellationToken cancellationToken = default);
         public Task AddToQueueAsync<T>(T value, CancellationToken cancellationToken = default);
 
--- a/Game.Abstractions/Infrastructure/IPubSubService.cs
+++ b/Game.Abstractions/Infrastructure/IPubSubService.cs
@@ -27,13 +27,11 @@
         public Task Wake(string channel);
         public Task Subscribe(string channel, Action<(string message, string channel)> action, string? id = null);
 
-        // The worker-backed overloads require a non-null id: each spins up a BackgroundWorker (holding an OS wait
+        // The worker-backed overload requires a non-null id: it spins up a BackgroundWorker (holding an OS wait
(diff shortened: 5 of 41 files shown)


ginderjeremiah (commented):
## Review: Backend dead-code & stale-comment sweep (#1514)

**Verdict:** Approve — clean, correctly-scoped cleanup; every deletion is justified and no test coverage is genuinely lost. _(Submitted as a comment because GitHub won't accept a gating verdict from the PR author; the intent is Approve.)_

I independently re-checked the claims rather than taking the PR/issue at face value, and they hold up:

### What's good
- **Verified deletions, not assumed ones.** The one member the issue flagged as "confirm before deleting" — `SkillRecipes.RecipesForInputSkill` + the `RecipeIdsByInputSkill` reverse index — was traced through spike #1125's design vs. the shipped `GetSkillRecipes` command and confirmed never wired. The right call, and documented.
- **No coverage lost on the deletions.** `Enemies.GetRandomEnemy`'s edge-case tests (invalid zone → `ArgumentOutOfRange`, empty zone → `InvalidOperation`, retired-exclusion) were retargeted onto `GetRandomDomainEnemy`, which shares the same `GetRandomEnemyId` zone-validation/spawn logic, so the coverage is preserved through the shared helper. `Zones.GetZone`'s field-mapping assertions are genuinely redundant with `ZoneMapperTests` — I confirmed `ToContract_RoundTripsAllScalarFields` and `ToContract_PreservesNullOptionalFieldsAndHomeFlag` both assert `IsHome` (true and false), so removing the integration-level `GetZone(...).IsHome` checks loses nothing.
- **The `UnSubscribe(channel)` discrepancy was caught.** The issue claimed it unused; it wasn't (`AdminCacheReloadBroadcastTests` used it). Switching that test to the id-scoped `Subscribe`/`UnSubscribe(channel, id)` pattern before removing the hazardous unsubscribe-every-handler overload is the correct fix, not a workaround.
- **Test-only footgun properly sealed.** Moving `Player.GetAllModifiers`/`GetAttributes` and `Inventory.GetEquippedAttributeModifiers` into `Game.Core.TestInfrastructure` as extension methods means a future production caller can no longer accidentally reach the lesser (pre-class-system) composition. Verified the extensions only touch public surface (`Player.StatPoints`, `Inventory.EquipmentSlots`/`GetUnlockedItem`, `Item.GetAttributeModifiers`), so the cross-assembly move is sound. Leaving `AttributeCollection.AllModifiers()` in place with a test-only doc comment (it needs private node storage) is a reasonable pragmatic exception.
- **Comment corrections are factually right** — the `BattleService.RecordVictory` anti-cheat note (server-clock-only, not client-timestamp validation) and the `ProficiencyXpPerVictory` doc (independent overlapping claims, no difficulty multiplier, multi-axis mints more) now match the actual post-#1318 behavior.

### On the one behavioral change (`AddDbContextPool`)
Switching `AddGameContext` from a per-resolution factory to `AddDbContextPool` is the only non-deletion change, so I checked it against the pooling constraints: `GameContext` has the required single `(DbContextOptions<GameContext>)` constructor, no instance fields or per-scope injected state, no `OnConfiguring` override (config flows through options), and its only `ChangeTracker` use (`ApplyZeroBasedIdentityFixups`) runs per-save against state the pool resets on return. Safe to pool, and the integration suites exercise it through the pooled registration. The `GameContextFactory.Configure` extraction keeps `GetGameContext` behavior-identical.

### Test coverage
Appropriate for a cleanup PR — no new domain logic is introduced (the new extension class is test infrastructure, exercised by `BattlerFactory` and the unit tests that consume it), and the DI/factory change is covered by the existing integration suites running through the new pooled path.

Nice, careful work. Merging once CI is green.

---

Issue #1529, opened 2026-07-03
Content lint: flag enemy attribute distributions the enemy's kit can't consume


ginderjeremiah:
Sub-issue of #1426. Part of the #1390 graph-lint family.

## Problem

After the #1426 severings, an enemy's AGI/LUK distribution points do nothing (enemies don't crit/dodge/parry, and nothing derives from those attributes on the enemy side) — yet they still count into `SumCoreAttributes`, inflating the enemy's XP payout without adding threat. The same applies to any anchor attribute (STR/DEX/INT) that no skill in the enemy's pool scales off.

## Changes

Add a content-health lint (warning severity): for each enemy, flag any attribute distribution where nothing in the enemy's kit consumes the attribute:

- **Consumed** = a static derivation targets it (END → Toughness/MaxHealth, STR → MaxHealth), a skill in the enemy's pool has a `DamageMultiplier` on it, or a skill effect scales off it.
- **AGI/LUK**: always flagged for enemies until an enemy-side consumer exists (enemy proc parity is a recorded later seam in the spike doc, not current mechanics).

## Notes

- A heuristic consumed-by map is fine for v1; the combat-rating spike (#1526) computes per-attribute marginals that would make this exact — share the model when it lands.
- Add the authoring-guidance line alongside: fast enemies author `CooldownRecovery`/`CooldownBonus` directly (enemy distributions have no core-only validation).

ginderjeremiah:
Heads-up from spike #1526: the combat rating (#1531) ships a per-attribute marginal helper (finite-difference on the rating) that is exactly the dead-stat detector this lint needs — a distribution point with ~zero rating marginal is dead weight for the kit. Once #1531 lands, prefer building this lint on that helper rather than hand-maintaining a consumed-attributes map; it stays correct automatically as mechanics (and the rating's terms) evolve.

ginderjeremiah:
[...]

PR #1574, opened 2026-07-03
Combat rating: closed-form capability rating computed from the assembled battler (#1531)

### LINKED ISSUE #1531: Combat rating: closed-form capability rating computed from the assembled battler

ginderjeremiah:
Closes #1531 (implements the rating decided in spike #1526 — full reasoning in `docs/spikes/1526-combat-rating-power-measure.md`).

## What

A pure, server-only `CombatRating` in `Game.Core.Battle`: `Rate(Battler, bool isPlayer)` computed from the assembled battler (attributes + fielded `BattleSkills` + `CounterSkill`) against fixed reference profiles in `ServerGameConstants` (`RefToughness`, `RefIncomingTypeMix`, `RefAttackRate`, `RefDps`, `RefFightDuration` — all strawman values pending the calibration report, #1533).

```
Rating = √(OffenseRate × Survivability)
```

- **Offense rate**: for each fielded skill, `expectedHit ÷ effectiveCooldownSeconds` — per-portion weighted split, amplification, the Toughness curve (against a reference defense), crit expectation (player-only), execute expectation — plus DoT effects (exact steady-state closed form: `magnitude × frozen amp × duration ÷ cooldown`), a reflect term (`DamageReflection × RefDps`), and a riposte term (player-only, `effectiveParryChance × RefAttackRate × expectedHit(counterSkill)`).
- **Survivability**: `(MaxHealth + regenRate × RefFightDuration) ÷ mitigated-and-avoided fraction`, where avoid (`parry + (1-parry)×dodge`) is player-only.
- **Steady-state effect pricing**: every fielded skill effect (buffs/debuffs) is folded into an ephemeral snapshot at its uptime-weighted average magnitude (`(Amount + casterAttr×ScalingAmount) × min(duration, RefFightDuration/2) ÷ effectiveCooldown`) — self-targeted effects fold into a caster snapshot, opponent-targeted non-DoT effects (Hex/Sunder-style debuffs) fold into a reference-defense snapshot — so offense/survivability compute once, order-free, with no per-tick simulation.
- **Engine-asymmetry mirroring**: enemies never crit/dodge/parry in this engine, so those terms (and riposte) are skipped entirely for `isPlayer: false`, even when an enemy skill carries an authored `CriticalChance` — the same asymmetry the live damage pipeline enforces.
- **Maintenance contract**: `CombatRating.Classify(EAttribute)` is an exhaustive classification (Offense / Survivability / Both / NeutralByDesign / AsymmetryGated) over every attribute, backed by a test enumerating the whole enum — a new attribute fails the build until its rating term is decided.
- **Marginal helper**: `CombatRating.Marginal(Battler, bool, EAttribute, delta)` — a finite-difference per-attribute contribution, backed by a new `Battler.CloneWithAttributeDelta` (full cascade re-derivation, so bumping a core attribute re-derives everything that hangs off it). Doubles as the dead-stat detector the enemy content lint (#1529) will need.

## Anti-drift

Raw damage, amplification, mitigation, and cooldown multipliers are read through the exact same methods the live engine uses — `BattleSkill.CalculateRawDamage`, `Battler.AmplifyDamage`, `Battler.ComputeNetDamage`, `Battler.GetCooldownMultiplier` — rather than reimplementing the formulas, so the rating cannot silently diverge from the real damage pipeline.

## Out of scope

- Swapping `DefeatRewards`/proficiency onto the rating and the reward-curve reshape (#1532).
- The calibration report that picks final constant values (#1533).
- UI surfacing (#1534).
- Documentation updates (`game-design.md`, `content-design.md`, `backend-attributes.md`) — these ride #1532 per the spike doc, not this issue.

## Testing

- New `Game.Core.Tests/Battle/CombatRatingTests.cs`: the `Classify` exhaustiveness contract, degenerate guards (a skill-less battler rates a finite positive value), survivability (MaxHealth/Toughness raise the rating), offense (a damaging skill, `DamageReflection`, and a zero-BaseDamage DoT effect all raise the rating), the enemy asymmetry gate (an authored `CriticalChance`/riposte is priced for a player but not an enemy), and `Marginal` (Luck marginals to exactly 0 with no fielded crit/parry enabler; Agility and Endurance always marginal positive).
- Full backend suite: `Game.Core.Tests` (1295), `Game.Infrastructure.Tests` (59), `Game.Application.Tests` (824), `Game.Api.Tests` (677) — all pass.
- No frontend changes: this is explicitly server-only with no client-side simulation (display values will be sent, not recomputed — #1534), so there is no parity surface and no codegen impact (`ServerGameConstants` is deliberately not `[ClientMirrored]`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01831iuoTxRFoNW75ySfx6vQ)_

### DIFF
--- a/Game.Core/Battle/CombatRating.cs
+++ b/Game.Core/Battle/CombatRating.cs
@@ -55,14 +55,6 @@ namespace Game.Core.Battle
         // zero-survivability battler never divides a downstream consumer (e.g. a rating ratio) by zero.
         private const double Epsilon = 1e-6;
 
-        // The composed value of every attribute when built from no extra modifiers — i.e. just
-        // StaticAttributeModifiers.All with every core attribute at 0. Since every static entry is Additive
-        // (no Multiplicative terms) and no core attributes are seeded into the effective-caster snapshot below,
-        // this is exactly the amount that snapshot's automatic re-application of the static modifiers already
-        // contributes — subtracting it out lets a single explicit BaseValue term set the snapshot's attribute
-        // to the battler's real composed value with no double-counting.
-        private static readonly IReadOnlyDictionary<EAttribute, double> ZeroBaseline = BuildZeroBaseline();
-
         /// <summary>
         /// The rating for <paramref name="battler"/> as assembled — <c>√(OffenseRate × Survivability)</c>
         /// against the fixed reference profiles in <see cref="ServerGameConstants"/>. Pass
@@ -190,8 +182,10 @@ namespace Game.Core.Battle
 
                     // The exact DoT steady-state closed form (spike #1526 Decision 1): magnitude × frozen
                     // amplification × duration ÷ effective cooldown. DoT bypasses the reference Toughness
-                    // (its real property), so no mitigation term applies here.
-                    var magnitude = effect.Amount + effectiveCaster.GetAttributeValue(effect.ScalingAttributeId) * effect.ScalingAmount;
+                    // (its real property), so no mitigation term applies here. The scaling attribute is read
+                    // off the original battler, not effectiveCaster — mirroring FoldedEffectMagnitude's "the
+                    // caster's ORIGINAL attribute" rule, so the two DoT-adjacent reads agree.
+                    var magnitude = effect.Amount + battler.GetAttributeValue(effect.ScalingAttributeId) * effect.ScalingAmount;
                     var ampedMagnitude = effectiveCaster.AmplifyDamage(magnitude, dotType);
                     var durationSec = effect.DurationMs / 1000.0;
                     total += ampedMagnitude * durationSec / effectiveCooldownSec;
@@ -295,24 +289,40 @@ namespace Game.Core.Battle
 
         // A battler equal to the assembled one but with every fielded Self-targeted effect folded in at its
         // uptime-weighted average magnitude — so offense/survivability compute once from one order-free state
-        // rather than simulating the battle. Every other (non-effect) attribute is copied at its exact composed
-        // value.
+        // rather than simulating the battle. Every other attribute is copied at its exact composed value,
+        // INCLUDING the core attributes: BattleSkill.CalculateRawDamage sums DamageMultipliers over core
+        // attributes directly (not through a derived static), so a core-stripped snapshot would zero out every
+        // core-scaled skill's damage. Seeding the real cores lets a core-derived attribute (CooldownRecovery,
+        // Toughness, MaxHealth, the crit/parry/dodge multipliers) re-derive naturally from them via the fresh
+        // AttributeCollection's automatic re-application of StaticAttributeModifiers — so each non-core
+        // attribute's explicit delta below is computed against a "cores-only" baseline (the value the real
+        // cores alone would produce) rather than an all-zero one, crediting only the extra (gear/proficiency/
+        // effect) contribution on top and avoiding double-counting the core-derived share.
         private static Battler BuildEffectiveCaster(Battler battler)
         {
-            var modifiers = new List<AttributeModifier>();
+            var coreModifiers = Attributes.Attribute.CoreAttributes
+                .Select(a => new AttributeModifier
+                {
+                    Attribute = a,
+                    Amount = battler.GetAttributeValue(a),
+                    Type = EModifierType.Additive,
+                    Source = EAttributeModifierSource.BaseValue,
+                })
+                .ToList();
+            var coreOnlyBaseline = new AttributeCollection(coreModifiers);
+
+            var modifiers = new List<AttributeModifier>(coreModifiers);
             foreach (var attribute in Enum.GetValues<EAttribute>())
             {
                 if (Attributes.Attribute.IsCore(attribute))
                 {
-                    // Core attributes are never read directly by the rating — only the attributes they feed
-                    // (via StaticAttributeModifiers) are — so they are simply omitted from the copy.
-                    continue;
+                    continue; // already seeded directly above
                 }
 
                 modifiers.Add(new AttributeModifier
                 {
                     Attribute = attribute,
-                    Amount = battler.GetAttributeValue(attribute) - ZeroBaseline[attribute],
+                    Amount = battler.GetAttributeValue(attribute) - coreOnlyBaseline[attribute],
                     Type = EModifierType.Additive,
                     Source = EAttributeModifierSource.BaseValue,
                 });
@@ -411,12 +421,6 @@ namespace Game.Core.Battle
             }
 
             return (EModifierType.Additive, rawMagnitude * uptime);
-        }
-
-        private static IReadOnlyDictionary<EAttribute, double> BuildZeroBaseline()
-        {
-            var collection = new AttributeCollection([]);
-            return Enum.GetValues<EAttribute>().ToDictionary(a => a, a => collection[a]);
         }
     }
 }
--- a/Game.Core/Battle/CombatRating.cs
+++ b/Game.Core/Battle/CombatRating.cs
@@ -1,0 +1,422 @@
+using Game.Core.Attributes;
+using Game.Core.Attributes.Modifiers;
+using Game.Core.Skills;
+using static Game.Core.EAttribute;
+
+namespace Game.Core.Battle
+{
+    /// <summary>
+    /// The combat-rating classification of an <see cref="EAttribute"/> — the maintenance contract from spike
+    /// #1526 Decision 9: every attribute is assigned one via <see cref="CombatRating.Classify"/>, backed by an
+    /// exhaustiveness test, so a newly added attribute fails the build until its rating term is decided rather
+    /// than being silently unpriced.
+    /// </summary>
+    public enum ECombatRatingClassification
+    {
+        /// <summary>Feeds <see cref="CombatRating"/>'s offense rate (amplification, ExecuteBonus, …).</summary>
+        Offense,
+
+        /// <summary>Feeds <see cref="CombatRating"/>'s survivability (MaxHealth, resistances, Toughness, regen, …).</summary>
+        Survivability,
+
+        /// <summary>Feeds both sides (e.g. a core attribute that derives into both an offense and a defense term).</summary>
+        Both,
+
+        /// <summary>Deliberately unpriced — inert by design (obsolete) or read structurally rather than as a scalar term.</summary>
+        NeutralByDesign,
+
+        /// <summary>Priced only for a player battler; skipped entirely for an enemy (crit/dodge/parry).</summary>
+        AsymmetryGated,
+    }
+
+    /// <summary>
+    /// The closed-form capability rating decided in spike #1526
+    /// (<c>docs/spikes/1526-combat-rating-power-measure.md</c>): <c>Rate = √(OffenseRate × Survivability)</c>,
+    /// computed from an assembled <see cref="Battler"/> against fixed reference profiles
+    /// (<see cref="ServerGameConstants"/>). The geometric mean is exact, not aesthetic — for any matchup,
+    /// <c>(enemyRating / playerRating)²</c> <b>is</b> the time-to-kill ratio in this engine.
+    /// <para>
+    /// Server-only: this is never simulated client-side (display values are sent, not recomputed), so it
+    /// carries no frontend/backend parity surface. Reuses the exact <see cref="Battler"/>/<see cref="BattleSkill"/>
+    /// methods the engine itself uses (<see cref="BattleSkill.CalculateRawDamage"/>, <see cref="Battler.AmplifyDamage"/>,
+    /// <see cref="Battler.ComputeNetDamage"/>, <see cref="Battler.GetCooldownMultiplier"/>) — the anti-drift
+    /// mechanism that keeps the rating from silently diverging from the real damage pipeline.
+    /// </para>
+    /// <para>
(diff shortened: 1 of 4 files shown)


Game.Core/Battle/CombatRating.cs
```diff
@@ -0,0 +1,422 @@
...
+        {
+            var modifiers = new List<AttributeModifier>();
+            foreach (var attribute in Enum.GetValues<EAttribute>())
+            {
+                if (Attributes.Attribute.IsCore(attribute))
+                {
+                    // Core attributes are never read directly by the rating — only the attributes they feed
+                    // (via StaticAttributeModifiers) are — so they are simply omitted from the copy.
+                    continue;
+                }
```
ginderjeremiah, line 320: **Blocking — this omission zeroes out the core-attribute scaling of every skill's damage.**

The comment's premise ("core attributes are never read directly by the rating") is not true. Two of the reused engine reads take a core attribute *directly*, not through a derived static:

- `BattleSkill.CalculateRawDamage` (line 220) sums `caster.GetAttributeValue(multiplier.Attribute) × multiplier.Amount` over the skill's `DamageMultipliers`, and those multipliers are core attributes — every seeded damage skill scales off one (`Punch/Strike/Cleave` → Strength ×1.0, `Focus` → Intellect ×1.0).
- The DoT term (line 194) reads `effect.ScalingAttributeId`, which the live engine (`BattleContext.ApplySkillEffect`) scales off the caster and is likewise a core attribute in general.

Because `effectiveCaster` has no core attributes, both reads see `0`. So a skill's raw damage collapses to `BaseDamage` alone and the entire Strength/Intellect contribution to offense is dropped — the rating materially under-prices offense for essentially all real content (and `Marginal(_, Strength, …)` can never see a damage build's primary stat). The existing offense tests pass only because they use `DamageMultipliers = []` and `ScalingAmount = 0`.

The fix needs `effectiveCaster` to expose core attributes at their real composed values without double-counting the derived statics they feed. One option: seed a first collection from the real core values (which auto-derives the statics), then for each non-core attribute add `real − firstCollection[attribute]` as the extra (gear/effect) contribution on top — so core reads real and derived reads real with no double count. Whatever the shape, add tests with a core-scaled `DamageMultiplier` and a nonzero-`ScalingAmount` DoT so this is pinned.

---

Game.Core/Battle/CombatRating.cs
```diff
@@ -0,0 +1,422 @@
...
+
+            return total;
+        }
+
+        // The expected post-mitigation damage of one fire of <paramref name="skill"/>, amplified by
+        // <paramref name="effectiveCaster"/> and mitigated by <paramref name="referenceDefense"/> — shared by
+        // a skill's own direct hit and the parry counter's phantom fire (both are "one hit of a skill").
+        private static double ExpectedDirectHit(Skill skill, bool isPlayer, Battler effectiveCaster, Battler referenceDefense)
+        {
+            var raw = BattleSkill.CalculateRawDamage(skill, effectiveCaster);
```
ginderjeremiah, line 214: `CalculateRawDamage(skill, effectiveCaster)` — since `effectiveCaster` drops all core attributes (see `BuildEffectiveCaster`), the `DamageMultipliers` sum evaluates to `0` and raw damage collapses to `BaseDamage`. This is where the blocking core-attribute omission manifests for direct hits. (Same root cause flagged in `BuildEffectiveCaster`.)

---

Game.Core/Battle/CombatRating.cs
```diff
@@ -0,0 +1,422 @@
...
+
+                    if (DamageTypes.DotTypeForAccumulator(effect.AttributeId) is not EDamageType dotType)
+                    {
+                        continue;
+                    }
+
+                    // The exact DoT steady-state closed form (spike #1526 Decision 1): magnitude × frozen
+                    // amplification × duration ÷ effective cooldown. DoT bypasses the reference Toughness
+                    // (its real property), so no mitigation term applies here.
+                    var magnitude = effect.Amount + effectiveCaster.GetAttributeValue(effect.ScalingAttributeId) * effect.ScalingAmount;
```
ginderjeremiah: This reads the DoT scaling attribute off `effectiveCaster`, whereas `FoldedEffectMagnitude` deliberately reads it off the original `battler` (with the explicit "reads the caster's ORIGINAL attribute" rationale). Two DoT-adjacent code paths disagree on which battler to scale from, and `effectiveCaster` additionally zeroes core attributes — so a DoT scaling off a core attribute is priced at `effect.Amount` only. Seeded content currently has `ScalingAmount = 0`, so it's latent today, but it's the same defect as the direct-hit case above; it should read the same battler `FoldedEffectMagnitude` does once core attributes are exposed.

---

ginderjeremiah (commented):
## Review: Combat rating — closed-form capability rating from the assembled battler (#1531)

**Verdict: Request changes** — submitted as a comment because I'm the PR author and GitHub won't let me gate my own PR; treat this as request-changes. There's one blocking correctness defect that makes the offense side of the rating wrong for essentially all real content.

The overall design is faithful to spike #1526 and the issue: the geometric-mean shape, the steady-state effect folding, the reference profiles, the enemy asymmetry gating, and the enforced `Classify` exhaustiveness contract are all well-executed, and the anti-drift intent (reuse the engine's own `Battler`/`BattleSkill` methods) is exactly right. The bug is that the anti-drift reuse is fed a battler that has been stripped of its core attributes.

### Blocking

- **`BuildEffectiveCaster` omits core attributes, but the reused engine reads take them directly — so all core-attribute damage scaling is priced as 0.** `CombatRating.cs:305-310` skips every core attribute, on the premise that "core attributes are never read directly by the rating." That premise is false: `CalculateRawDamage` (`:220`) sums `DamageMultipliers` over core attributes, and the DoT term (`:194`) reads `effect.ScalingAttributeId` (a core attribute in general). Every seeded damage skill scales off a core stat (`Punch/Strike/Cleave` → Strength, `Focus` → Intellect), so with `effectiveCaster` core-zeroed, each skill's raw damage collapses to `BaseDamage` and the whole Strength/Intellect contribution to offense vanishes. The rating materially under-prices offense, and `Marginal(_, Strength, …)` can't see a damage build's primary stat. The offense unit tests pass only because they use empty `DamageMultipliers` and `ScalingAmount = 0`. Details + a suggested fix direction are on the inline comment at `:305`.

### Should-fix

- **Test coverage gap that hid the blocker.** The offense/DoT/marginal suites never field a skill with a core-scaled `DamageMultiplier` or a nonzero-`ScalingAmount` DoT. Add both — a Strength-scaled skill whose rating rises with Strength, and a `Marginal(_, Strength, …) > 0` for a damage build — so the core-attribute path is pinned and can't silently regress again.

### Nits

- `Classify_IsDefinedForEveryAttribute` asserts `Enum.IsDefined(Classify(a))`, but `Classify` can only ever return a defined member or throw — so the assertion is near-tautological and the real contract it enforces is "doesn't throw for any member." That's fine as-is (the throw still fails the build), just noting the assert body is somewhat cosmetic.

### Test coverage

Structure and breadth are good — degenerate guards, survivability, reflect, DoT-with-zero-base-damage, the enemy crit/riposte asymmetry, and the `Classify` exhaustiveness contract are all covered, and the server-only no-parity-surface rationale is correct (no frontend mirror needed). The one real gap is the core-attribute scaling path called out above, which is exactly what let the blocking bug through.

### What's good

- The `ZeroBaseline` subtract-the-static-contribution trick is a clean, correct way to set a folded attribute to its real composed value given the statics are all additive.
- `CloneWithAttributeDelta` correctly filters the by-reference static instances and `SkillEffect` modifiers before rebuilding, so the finite-difference marginal re-derives the full cascade with no double-count.
- Reusing `ComputeNetDamage`/`AmplifyDamage`/`GetCooldownMultiplier` for anti-drift, and mirroring the engine's crit/dodge/parry/riposte asymmetry in one documented place, is the right call and matches the spike.
- The maintenance contract (exhaustiveness test failing the build on a new unclassified attribute) is a genuinely good guardrail.


---

Issue #1525, opened 2026-07-03
Derive CriticalChanceMultiplier and ParryChanceMultiplier from Luck (proc-payoff amplifier)


ginderjeremiah:
Sub-issue of #1426 (LUK is the proc-payoff amplifier) — full reasoning in `docs/spikes/1426-attribute-identity-sourcing.md`.

## Changes

- Add `Derived` entries in `StaticAttributeModifiers`: `Luck → CriticalChanceMultiplier` and `Luck → ParryChanceMultiplier` (strawman coefficients, e.g. `0.002` each, tuned during balancing).
- **Keep** `CriticalDamage`'s `1.5` base and `0.0025·Luck` derivation — LUK both raises proc frequency (the multipliers) and spikes the payoff (crit damage), and the `1.5` floor remains the "welcome to crit" payoff that gets the identity off the ground.

Opt-in is preserved by construction: both new targets are base-`1` multipliers scaling `0`-based enablers (a skill's authored `CriticalChance`; the authored-only `ParryChance`), so `0 × mult = 0` and LUK stays inert until a proc enabler is fielded — dormant, not dead.

## Notes

- No new enum members — no migration. The frontend statics table is codegen'd; regenerate.
- Deliberately **no** DEX → `CriticalChanceMultiplier` derivation (#1426): DEX stays a pure magnitude anchor; a crit hook would double-dip (damage scaling + crit scaling) and dominate STR for any crit-committed physical build.
- Tests: derivation/parity scenarios; a LUK build with no crit/parry enabler must behave identically to a `0`-LUK build.

Issue #1524, opened 2026-07-03
Sever CDR's attribute sourcing; add the CooldownBonus × CooldownBonusMultiplier cadence pair


ginderjeremiah:
Sub-issue of #1426 (CDR's incidental attribute sourcing is severed; cadence becomes a committed, opt-in channel) — full reasoning in `docs/spikes/1426-attribute-identity-sourcing.md`. Unblocks the Frequency proficiency path (follow-up sub-issue of #1426).

## Goal

`CooldownRecovery`'s functional `1.0` base blocks the multiplicative-of-zero trick (#1398 Decision 7), so de-universalizing cadence means severing the attribute derivations outright and rebuilding the committed channel as the fourth instance of the enabler × multiplier pair (crit, parry, dodge, now cadence).

## Changes

- **Remove the `0.004·Agility` and `0.001·Dexterity` → `CooldownRecovery` derivations.** The attribute keeps its `1.0` base and stays the directly-read multiplier for authored/effect modifiers.
- **Add `CooldownBonus`** — new `EAttribute`, base `0`, authored-only (items / skill effects). The committed cadence enabler.
- **Add `CooldownBonusMultiplier`** — new `EAttribute`, base `1.0` additive plus a `Derived` entry from `Agility` (strawman coefficient, e.g. `0.002`) — AGI retains its cadence identity in opt-in form, and the Frequency path grants its bonuses here.
- **The charge site reads `CooldownRecovery + CooldownBonus × CooldownBonusMultiplier`** as the effective rate (product computed at consumption, like crit/parry/dodge).

Why the pair instead of path bonuses on `CooldownRecovery` itself: a flat path reward would be permanent universal value that survives abandoning the build, and a multiplicative one would scale the functional `1.0` base and lift the uncommitted. `bonus × mult` idles at `0` the moment the enabler is unequipped.

## Notes

- Two new enum members ⇒ **EF migration required** (HasData-seeded intrinsic enums; startup project `Game.Infrastructure`).
- The frontend statics table is codegen'd; the effective-rate read is parity-critical (tick loop) — mirror FE/BE with matching tests (a no-enabler build charges at exactly the base rate regardless of AGI; the bonus scales with the multiplier; a timed `CooldownBonus` effect is amplified).
[...]

Issue #1523, opened 2026-07-03
Dodge rework: opt-in dodge — authored-only DodgeChance × AGI-derived DodgeChanceMultiplier


ginderjeremiah:
Sub-issue of #1426 (dodge goes opt-in on the parry template) — full reasoning in `docs/spikes/1426-attribute-identity-sourcing.md`.

## Goal

Make dodge a committed, opt-in identity like crit (#1425) and parry (#1457): flat `DodgeChance` becomes authored-only (items / item mods / skill effects), and `Agility` amplifies the committed base instead of granting free evasion.

## Changes

- **Remove the `0.001·Agility` → `DodgeChance` derivation** in `StaticAttributeModifiers`. `DodgeChance` keeps no base, so an un-enabled build sits at exactly `0` (and a class locked base that grows AGI no longer opens the Evasion path for free).
- **Add `DodgeChanceMultiplier`** — a new `EAttribute` (Secondary, percentage-rendered): base `1.0` additive plus a `Derived` entry from `Agility` (coefficient strawman, e.g. `0.002`, tuned during balancing). Exactly mirrors `ParryChanceMultiplier` (#1457).
- **The dodge check reads `DodgeChance × DodgeChanceMultiplier`** at check time, like parry's `ParryChance × ParryChanceMultiplier` (the attribute pipeline can't express an attribute product; the consumption site computes it). Parry-before-dodge ordering is unchanged.
- **Evasion path bonuses target the multiplier** (authoring note — path content lives in the Workbench), so path investment scales a committed base and idles for anyone who abandons the build (`0 × mult = 0`).

## Notes

- New enum member ⇒ **EF migration required** (intrinsic enums are HasData-seeded; startup project `Game.Infrastructure`) or all integration tests fail on `PendingModelChangesWarning`.
- The frontend `STATIC_ATTRIBUTE_MODIFIERS` table is codegen'd — regenerate rather than hand-mirror.
- Parity-critical: the check-time product runs in the tick loop on both sides — mirror FE/BE with matching test scenarios (a `0`-chance build never dodges regardless of AGI; the multiplier scales an authored base; parry precedence unchanged).
- Content sequencing: the Evasion row's intro becomes "the first dodge-granting item" — that enabler must be reachable early (content-design §4 rollout row updated by the spike).
- Pre-release: existing AGI allocations simply stop granting dodge — no grandfathering.
- Art: the new attribute will want an icon (follow-up like #1490).

Issue #1426, opened 2026-07-01
Spike: attribute identity-sourcing — what DEX/LUK/AGI do when combat identities become opt-in


ginderjeremiah:
EDIT: for AGI, we should go opt-in. Make it a multiplier towards dodge, and make items the main source of flat dodge chance.

Follow-on spike from #1398 (see [`docs/spikes/1398-utility-in-proficiency-system.md`](docs/spikes/1398-utility-in-proficiency-system.md), Decision 8).

## Context

#1398 concluded that combat *identities* (crit, and potentially CDR/dodge) should be **opt-in enablers** (expectation-0), not attribute derivations, so they are committed rather than universal. Making crit opt-in (#1425) empties `Dexterity`/`Luck` of the crit chance they currently source, and the same logic pressures `Agility` (dodge / CDR). This spike decides what those attributes become.

## Questions

- **Per stat: multiplicative-amplifier vs direct removal.** Multiplicative-of-zero only de-universalizes a stat that is already `0`-based. Owner's seed direction:
  - **Crit** (`Dexterity`/`Luck`) → **multiplicative** (crit chance is `0`-based; `Luck` can also stay on crit *damage*).
  - **CDR** (`Agility`/`Dexterity`) → **direct removal** — CDR's functional `1.0` base blocks the multiplicative trick (a multiplicative `Agility` still lifts everyone above `1.0`), so it must be severed outright if cadence is to become a commitment.
  - **Dodge** (`Agility`) → middle case: dodge chance is `0`-based like crit (so multiplicative is available), but dodge is *substitutive* and, once CDR is stripped from `Agility`, `Agility` collapses to pure dodge — arguably narrow enough to be a genuine commitment already. Decide whether it stays additive-AGI or goes opt-in.
- **Do these remain allocatable attributes at all**, or become gear/skill-only? Once they only matter with a matching enabler, allocating free-pool points into them is a **newbie trap** (dead stats). Weigh a leaner magnitude-only set (STR/INT/END generalist baseline + identity attributes as specialist amplifiers) against keeping the six with sequencing safeguards.
- **Frequency / CDR proficiency** rides here: it is reachable via the opt-in route *iff* CDR sourcing is severed (per #1398). If pursued, note it is a *representational* signal (build-state — accurate, but flat-per-victory) that does not self-supply a difficulty signal, so it needs explicit difficulty scaling.

## Out of this spike

[...]

PR #1576, opened 2026-07-04
Attributes screen: signal inert AGI/LUK when no matching enabler is fielded

### LINKED ISSUE #1528: Attributes screen: signal inert identity attributes (no matching enabler fielded)
Sub-issue of #1426 (all six attributes stay allocatable; the newbie trap is handled by signaling, not by removing stats). Depends on the amplifier reworks (#1523, #1524, #1525) for the final semantics.

## Goal

After #1426, AGI and LUK are pure amplifiers — inert until a matching enabler is fielded. The free pool is freely refundable, so a misallocation only costs *noticing it*; this issue makes it noticeable at the point of allocation.

## Changes

Frontend-only. On the attributes screen, an identity attribute whose current build fields no matching enabler gets an inline hint (exact copy/affordance per the screen's design conventions):

- **AGI** is inert unless any fielded source grants `DodgeChance > 0` or `CooldownBonus > 0`.
- **LUK** is inert unless any fielded skill has an authored `CriticalChance > 0` or any fielded source grants `ParryChance > 0`.

"Fielded" = the assembled battle loadout (selected + granted skills post weapon-match gate, equipped gear/mods) — the same state the battler composes from, so the hint never contradicts the simulation.

## Notes

- Read-only signal — no allocation gating (paternalism was rejected in the spike; dormant points are a legitimate pre-commitment).
- Consider the inverse affordance too: when an enabler *is* fielded, show what the amplifier currently multiplies (e.g. "×1.24 on your 5% dodge") so the commitment is legible.

ginderjeremiah:
## Summary

Sub-issue of #1426, unblocked now that the amplifier reworks (#1523, #1524, #1525) have landed. AGI and LUK are pure amplifiers: their `*Multiplier` targets scale a `0`-based enabler (`DodgeChance`/`CooldownBonus` for AGI; a fielded skill's own authored `criticalChance` or `ParryChance` for LUK), so a build with no matching enabler fielded gets nothing from points invested there — dormant, not dead. This adds a read-only inline hint on the Attributes screen (both Guided and Theorycraft modes) signaling that state, per #1528.

- `attributes-view.svelte.ts`: adds `isAttributeInert`/`hasFieldedEnabler` (pure, unit-tested) and `inertHint`, plus an `AttributesView.isInert(i)` / `fieldedSkills` derived that assembles the player's live fielded skill set — selected loadout + equipped-gear grants, weapon-match gated (`isSkillDormant`) — and equipped gear/mod static stat lines (`inventoryManager.equipmentStats`), mirroring what the battler actually fields so the hint can never contradict the simulation.
- `GuidedRow.svelte` / `TheoryRow.svelte`: render a `⊘ Dormant — …` note (styled like the Skills screen's existing `DormantNote.svelte` off-weapon note) when the row's attribute is currently inert.
- No allocation gating — the free-pool points stay a legitimate pre-commitment, per the spike's Decision 5.

**Scope note:** left the "inverse affordance" suggestion in the issue (showing what the multiplier currently amplifies, e.g. "×1.24 on your 5% dodge") out of scope — it was called out as a "consider," and would require surfacing the new `*Multiplier` attributes in `DERIVED_STATS`/the per-point-yield panel, which is a separate, larger change. Happy to follow up if wanted.

## Test plan

- [x] `dotnet build` — not applicable, frontend-only change
- [x] `npm run test:unit:ci` — full suite passes (287 files / 3376 tests), including new coverage in `attributes-view.test.ts` for `isAttributeInert` across gear-static, skill-effect, and skill-authored-crit enabler sources, the weapon-match gate, and equipped-gear-granted skills.

Closes #1528

---
_Generated by [Claude Code](https://claude.ai/code/session_01KhjEB9d89nZx55uif6embt)_

### DIFF
--- a/UI/src/routes/game/screens/attributes/GuidedRow.svelte
+++ b/UI/src/routes/game/screens/attributes/GuidedRow.svelte
@@ -28,6 +28,9 @@
 				{/each}
 			{/if}
 		</div>
+		{#if inert}
+			<span class="dormant-note">⊘ {hint}</span>
+		{/if}
 	</div>
 
 	<Delta from={view.savedValues[i]} to={view.values[i]} />
@@ -43,7 +46,7 @@ import { tooltipHover } from '$components/tooltip/tooltip-hover';
 import { describedByTooltip } from '$components/tooltip/describedby-tooltip';
 import Delta from './Delta.svelte';
 import Stepper from './Stepper.svelte';
-import { CORE_ATTRIBUTES, DERIVED_STATS, feedsFor, type AttributesView } from './attributes-view.svelte';
+import { CORE_ATTRIBUTES, DERIVED_STATS, feedsFor, inertHint, type AttributesView } from './attributes-view.svelte';
 
 interface Props {
 	i: number;
@@ -57,6 +60,10 @@ const color = $derived(attributeColor(id));
 const code = $derived(attributeCode(id, staticData.attributes));
 const name = $derived(attributeName(id, staticData.attributes));
 const feeds = $derived(feedsFor(i));
+// Read-only "dormant, not dead" signal for the amplifier attributes (AGI/LUK, spike #1426 Decision 5) —
+// no matching enabler is fielded, so the current build gets nothing from points spent here.
+const inert = $derived(view.isInert(i));
+const hint = $derived(inertHint(id));
 
 // Hover explainer for the attribute, driven through the screen-level controller.
 const attrTip = getAttributeTooltip();
@@ -150,5 +157,17 @@ const changed = $derived(
 		color: var(--text-muted);
 		border-style: dashed;
 	}
+}
+
+// Neutral muted tone matching the Skills screen's off-weapon dormancy note (DormantNote.svelte) —
+// --warning is reserved for validation, and this is a read-only signal, not a problem to fix.
+.dormant-note {
+	display: block;
+	margin-top: 4px;
+	font-family: var(--mono);
+	font-size: 8.5px;
+	letter-spacing: 0.5px;
+	text-transform: uppercase;
+	color: var(--text-tertiary);
 }
 </style>
--- a/UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
+++ b/UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
@@ -176,7 +176,10 @@ export function inertHint(id: EAttribute): string | undefined {
 /** Whether any fielded source — an equipped item/mod's static stat line, or a fielded skill's own
  *  self-targeted effect — authors a nonzero base amount of `attributeId`. Only the *authored* amount is
  *  checked (an effect's own base `amount`, not its live in-combat scaled value), matching the "committed
- *  enabler" framing the crit/dodge/parry/cadence template already uses. */
+ *  enabler" framing the crit/dodge/parry/cadence template already uses. This is exactly the source set
+ *  `StaticAttributeModifiers`' "granted by items/item-mods/skill-effects" convention documents for
+ *  DodgeChance/CooldownBonus/ParryChance — a future source authoring one of them outside that convention
+ *  (e.g. a class/proficiency payout) would need a matching scan added here. */
 function hasFieldedEnabler(
 	attributeId: EAttribute,
 	equipmentStats: readonly IBattlerAttribute[],
--- a/UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
+++ b/UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
@@ -22,11 +22,18 @@
    freely — any attribute can be decremented back to 0, refunding its points to
    the pool — so there is no respec cost or locked baseline here. */
 
-import { apiSocket, EAttribute, type IAttributeUpdate, type IBattlerAttribute } from '$lib/api';
-import { BattleAttributes } from '$lib/battle';
+import {
+	apiSocket,
+	EAttribute,
+	ESkillEffectTarget,
+	type IAttributeUpdate,
+	type IBattlerAttribute,
+	type ISkill
+} from '$lib/api';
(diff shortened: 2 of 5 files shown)


UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
```diff
@@ -147,6 +154,84 @@ export function contributorsFor(derivedId: EAttribute): EAttribute[] {
...
+/** Whether any fielded source — an equipped item/mod's static stat line, or a fielded skill's own
+ *  self-targeted effect — authors a nonzero base amount of `attributeId`. Only the *authored* amount is
+ *  checked (an effect's own base `amount`, not its live in-combat scaled value), matching the "committed
+ *  enabler" framing the crit/dodge/parry/cadence template already uses. */
+function hasFieldedEnabler(
+	attributeId: EAttribute,
+	equipmentStats: readonly IBattlerAttribute[],
+	fieldedSkills: readonly ISkill[]
+): boolean {
+	if (equipmentStats.some((a) => a.attributeId === attributeId && a.amount > 0)) {
```
ginderjeremiah, line 188: Nit (robustness): this enumerates the enabler sources as `equipmentStats` (items + item-mods) plus fielded-skill self-effects, which is exactly the set `StaticAttributeModifiers.cs` documents — DodgeChance/CooldownBonus "granted by items/(item-mods/)skill-effects", ParryChance "granted by stance effects/items" — with proficiency/class/AGI/LUK feeding the `*Multiplier` instead. So this is correct today.

The coupling is implicit, though: if a class locked base, proficiency payout, or signature passive ever authored a *raw* enabler (against that documented convention), this scan would miss it and the hint would silently disagree with the sim — the exact "hint contradicts the simulation" failure the feature is meant to avoid. A one-line comment tying this source list back to `StaticAttributeModifiers`' "granted by …" list would make the invariant explicit for a future editor. Optional.

---

UI/src/routes/game/screens/attributes/GuidedRow.svelte
```diff
@@ -151,4 +158,16 @@ const changed = $derived(
 		border-style: dashed;
 	}
 }
+
+// Neutral muted tone matching the Skills screen's off-weapon dormancy note (DormantNote.svelte) —
+// --warning is reserved for validation, and this is a read-only signal, not a problem to fix.
+.dormant-note {
```
ginderjeremiah, line 164: Nit (DRY): this `.dormant-note` block is now the third near-copy of the same `⊘` mono/uppercase/`--text-tertiary` treatment — `skills/DormantNote.svelte` and `TheoryRow.svelte` carry the other two, diverging only in `font-size`/margins. Per the frontend guideline to extend a primitive rather than copy it, consider a tiny presentational note component (`⊘ {text}` + the shared style, with size/margin as a prop or modifier class) that all three render, so the "dormant" visual language has one home. Minor — fine as a follow-up.

---

ginderjeremiah (commented):
## Review: Attributes screen inert-AGI/LUK signal (closes #1528)

**Verdict:** Approve — correct, faithful to the documented amplifier/enabler model, well-tested, and consistent with the frontend theming/organization conventions. (Submitted as a comment since I'm the PR author; the intent is an approval.) Two nits below, neither blocking.

I verified the crux the issue's own triage note flagged — that the "Dormant — no enabler fielded" copy would be *factually wrong* if shipped before the amplifier reworks. Post-#1523/#1524/#1525 the codegen'd `STATIC_ATTRIBUTE_MODIFIERS` confirms AGI derives **only** `DodgeChanceMultiplier`/`CooldownBonusMultiplier` and LUK **only** `CriticalChanceMultiplier`/`ParryChanceMultiplier` (+ a direct `CriticalDamage` term that is itself inert without a crit). The four enablers are 0-based and, per `StaticAttributeModifiers.cs`, granted exclusively by items/item-mods/skill-effects — precisely the set `hasFieldedEnabler` scans (`equipmentStats` + fielded-skill self-effects). Crit is `skill.criticalChance × CriticalChanceMultiplier` at fire time with no battler-level crit attribute, so keying the LUK-crit branch off a fielded skill's authored `criticalChance` (rather than an attribute scan) is exactly right. The loadout assembly in `fieldedPlayerSkills` mirrors `Battler.fillSkills` (selected + granted, de-duped first-wins, weapon-match-gated via `isSkillDormant`).

### Nits
- `attributes-view.svelte.ts:185` — the enabler-source list (items/mods/skill-effects) is correct but implicitly coupled to `StaticAttributeModifiers`' "granted by …" convention; a one-line comment tying it back there would harden it against a future class/proficiency/passive that authors a raw enabler. Optional.
- `GuidedRow.svelte:164` — the `.dormant-note` styling is now the third near-copy (with `skills/DormantNote.svelte` and `TheoryRow.svelte`); a small shared presentational note component would DRY up the "⊘ dormant" visual language. Fine as a follow-up.

### Test coverage
Strong. `attributes-view.test.ts` covers `isAttributeInert` across every enabler source (gear dodge/cooldown, zero-amount and unrelated-attribute negatives, skill self-effect dodge, opponent-targeted exclusion, LUK authored-crit, gear parry, zero-crit), `inertHint`, and `AttributesView.isInert` end-to-end (defaults, non-amplifier always-false, gear enabler, selected crit skill, unlocked-but-not-fielded exclusion, weapon-gate dormancy exclusion, gear-granted skill inclusion). Frontend-only display logic, so no FE/BE battle-parity mirror is required. CI is green.

### What's good
- Correctly treats LUK's direct `CriticalDamage` derivation as still-inert-without-a-crit rather than letting it leak a false "active" signal — a subtle trap handled right.
- De-dup + weapon-gate assembly mirrors the real battler, so the hint can't drift from what actually fires.
- Read-only signal with no allocation gating, honoring the spike's Decision 5; `--text-tertiary` (not `--warning`) keeps it an advisory, not a validation error.
- The out-of-scope "inverse affordance" was correctly deferred and called out.

---

Issue #1578, opened 2026-07-04
Consolidate enemy→Battler construction onto Enemy.ToBattler()


ginderjeremiah:
### Context

While building the combat-rating calibration report (#1533), I added `Enemy.ToBattler()` (`Game.Core/Enemies/Enemy.cs`) — a one-line factory wrapping `new Battler(new AttributeCollection(GetAttributeModifiers()), BattleSkills, Level)` — because the calibration report needed to build an enemy's battler from its fielded loadout.

That exact expression was already duplicated in two production call sites before this change:

- `Game.Application/Services/BattleService.cs` (~line 511)
- `Game.Core/Battle/Offline/OfflineProgressSimulator.cs` (~line 113)

plus several test fixtures (`TypedEnemyCombatTests`, `BattleSimulatorParityTests`, `BattleSimulatorTests`, `OfflineProgressSimulatorTests`).

### Ask

Migrate `BattleService.cs` and `OfflineProgressSimulator.cs` (and the test duplicates, where it's a clean win) to call the new `Enemy.ToBattler()` instead of re-deriving the same construction inline. This is a pure DRY cleanup with no behavior change — `ToBattler()` is byte-for-byte the same expression.

### Notes

- No behavior/output change expected — this is purely reuse, not a rewrite.
- Worth double-checking whether any of the test-file duplicates are asserting on the construction itself (in which case leave them, or update them to call the new helper if it doesn't weaken the test).

ginderjeremiah:
While triaging this issue for pickup: it asks to migrate `BattleService.cs`/`OfflineProgressSimulator.cs` to call `Enemy.ToBattler()`, but that method doesn't exist on `main` yet — it was added by PR #1579 (implementing #1533), which is still open/unmerged. Unassigning myself and treating this as blocked on #1579 landing rather than duplicating that method here ahead of merge. Will revisit once #1579 is merged.

---
_Generated by [Claude Code](https://claude.ai/code)_

ginderjeremiah:
Resolved by #1603 (merged). Both production `Enemy.ToBattler()` call sites this issue targeted — `BattleService.SimulateBattle` and `OfflineProgressSimulator.SimulateBattle` — were migrated onto `Enemy.ToBattler()` there (needed anyway to source the fielded-loadout combat rating). Closing as completed. The optional test-fixture dedup this issue also mentioned was left out of scope; reopen or file a fresh `tech debt` issue if it's still wanted.

---
[...]

PR #1579, opened 2026-07-04
Combat rating calibration report: old-vs-new pricing, zone placement, k/pie recommendation

### LINKED ISSUE #1533: Calibration report: rating vs SumCoreAttributes across authored content; pick XP scale and pie

ginderjeremiah:
## Summary

Implements #1533 (supporting tooling for spike #1526, Decision 10): a re-runnable calibration report that compares the outgoing `DefeatRewards.SumCoreAttributes` power measure against the incoming `CombatRating` (#1531) across authored content, and recommends the `k`/proficiency-pie constants the #1532 consumer swap needs.

- **`CombatRatingCalibrator`** (`Game.Application/Content/Calibration`) — pure over already-resolved domain objects:
  - **Enemy pricing**: old vs. new measure per enemy, with a `RelativeShift` (each enemy's share of the total pricing before/after the swap) that flags enemies whose pricing moves disproportionately relative to their peers — e.g. a LUK-heavy enemy's dead-stat inflation disappearing under the new rating.
  - **Zone placement**: for each zone, at sampled levels across its idle range, where a reference build's rating places the zone's spawn-weighted enemy rating on the `r = enemy/player` axis — under both the old and new measure, side by side.
  - **Recommended constants**: `k` and the proficiency pie, solved by anchoring on the zone-placement sample the *current* game already treats as matched (its old-measure ratio closest to 1), so the new bounty curve and max-normalized accrual pay the same as today's curve at that mid-arc reference point (spike #1526 Decision 10's "keep payouts roughly continuous with today" goal).
  - **Reward curve**: XP/kill and XP/hour across the zone arc under the recommended `k`, with win rate and battle duration measured by actually running `BattleSimulator` (not a formula-derived approximation) — the ground truth for "time per kill."
- **`Enemy.ToBattler()`** (`Game.Core/Enemies/Enemy.cs`) — extracted the `new Battler(new AttributeCollection(GetAttributeModifiers()), BattleSkills, Level)` construction (already duplicated in `BattleService.cs` and `OfflineProgressSimulator.cs`) so the calibration report reuses the same construction rather than a third copy. Filed #1578 to migrate the two existing production call sites (left untouched here — both are on the live battle-parity / offline-replay paths, out of scope for a calibration-tooling PR).
- **`DefeatRewards.SumCoreAttributes`** made `public` (was `private`) so the report reads the exact same "old measure" production code uses, rather than re-deriving it.
- **Docs**: added a short "Combat-rating calibration report" note to `backend.md` alongside the existing progression-graph-lint section, since it's the same content-tooling family.

## Notes on today's content

Current authored content (`content/*.json`) is a minimal placeholder seed — one class (flat, non-level-scaling attribute distribution), four skills (no DoT/reflect/buff effects), zero items/mods/proficiencies. The reference builds in the integration test can therefore only vary by core-attribute allocation (offense vs. survivability) rather than the richer tank/reflect/DoT/buff-centric archetypes the spike anticipates — the report is generic over whatever content is loaded and will pick up that variety automatically once it's authored. Running the report against this content also surfaced a genuine (pre-existing, unrelated to this PR) finding: because the seed class's attributes don't scale with level, enemies badly outscale the player at higher levels in the second zone — exactly the kind of pricing/placement issue this tool exists to catch. Not fixed here since it's a content-authoring/tuning concern, not a code defect.

This PR does **not** wire `k`/pie into `ServerGameConstants` or change any reward math — per #1533's scope, that's #1532's job. The report is a tool a human runs and reads during tuning.

## Test plan

- [x] `dotnet test` — full suite passes (2892 tests, including the new unit + integration coverage)
- [x] New unit tests (`CombatRatingCalibratorTests`) cover the pure math: enemy pricing/share math, zone-weighting, anchor selection, k/pie derivation, reward-curve determinism — no DB.
- [x] New integration test (`CombatRatingCalibrationIntegrationTests`) seeds the committed content export, wires real `ISkills`/`IEnemies`/`IZones`/`IClasses`, runs the full report, and logs it — asserts structural sanity (non-empty, positive ratings/constants) rather than pinning specific numbers, since the report's job is to inform tuning, not gate it.

Closes #1533

---
_Generated by [Claude Code](https://claude.ai/code/session_01EVBpEF39d6DY4vaTVtvm6e)_


Game.Application/Content/Calibration/CombatRatingCalibrator.cs
```diff
@@ -0,0 +1,249 @@
...
+                throw new InvalidOperationException(
+                    $"The calibration anchor ({anchorRow.ZoneName} L{anchorRow.Level}, {anchorRow.BuildName}) has a "
+                    + "non-positive old power measure — a degenerate reference build, not a value to divide by.");
+            }
+
+            // The old curve pays exactly the enemy's old measure at a matched ratio (DifficultyMultiplier == 1),
+            // so that is the "today" XP the new curve is anchored to reproduce at this sample.
+            var xpUnderOldCurveAtAnchor = anchorRow.SpawnTableOldMeasure;
+            var clampedNewRatio = Math.Min(anchorRow.NewRatio, 1.0);
+            var xpScaleK = xpUnderOldCurveAtAnchor / (anchorRow.SpawnTableNewRating * clampedNewRatio * clampedNewRatio);
```
ginderjeremiah: **Should-fix (accuracy):** `xpUnderOldCurveAtAnchor` assumes the anchor sits at a matched ratio (`DifficultyMultiplier == 1`), but the anchor is only the sample *closest* to 1 — nothing guarantees it's inside the ±20% band. The production old curve is `enemyOldMeasure × DifficultyMultiplier`, where the multiplier is `1` only for `0.8 ≤ ratio ≤ 1.2` and otherwise `ratio²` (clamped at `MaxExpRewardMultiplier`) — see `DefeatRewards.GetDifficultyMultiplier`. If the best anchor lands at, say, `oldRatio = 0.5`, today's real payout is `0.25 × measure`, so using the bare `SpawnTableOldMeasure` over-estimates `k` by 4× — silently, since the derivation never checks. This only bites when no near-matched sample exists, which is precisely the mis-calibrated content this tool exists to surface, so the `k` output is least trustworthy exactly when it matters most. Fold the actual old multiplier into the anchor XP (apply the same band/clamp `DifficultyMultiplier` uses to `anchorRow.OldRatio`) so `k` is correct regardless of where the closest sample lands. The unit test `RecommendConstants_XpScaleK_ReproducesTodaysXpAtTheAnchor` only exercises the `ratio = 1.0` case, so it wouldn't catch this.

---

Game.Application/Content/Calibration/CombatRatingCalibrator.cs
```diff
@@ -0,0 +1,249 @@
...
+        public static RecommendedConstants RecommendConstants(
+            IReadOnlyList<ZonePlacementRow> zonePlacement,
+            double currentProficiencyPie = ServerGameConstants.ProficiencyXpPerVictory)
+        {
+            if (zonePlacement.Count == 0)
+            {
+                throw new ArgumentException("Cannot recommend constants with no zone-placement samples.", nameof(zonePlacement));
+            }
+
+            var anchorRow = zonePlacement.MinBy(row => Math.Abs(Math.Log(row.OldRatio)))!;
```
ginderjeremiah: **Nit:** the null-forgiving `!` runs against `CLAUDE.md`'s "AVOID using the null-forgiving operator" rule. The `Count == 0` guard above does make `MinBy` non-null here, but the intent reads more clearly (and stays `!`-free) by folding the empty check into the selection, e.g. capture the result and throw on `null`:
```csharp
var anchorRow = zonePlacement.MinBy(row => Math.Abs(Math.Log(row.OldRatio)));
if (anchorRow is null)
{
    throw new ArgumentException("Cannot recommend constants with no zone-placement samples.", nameof(zonePlacement));
}
```
That also lets you drop the separate `Count == 0` guard.

Minor related edge case: `Math.Log(row.OldRatio)` is `-∞` when `OldRatio == 0` (an enemy population with zero old core-sum) and `NaN` propagates oddly through `MinBy`. Not worth guarding for real content, but noting it since the surrounding code is otherwise careful about degenerate inputs (the `PlayerOldMeasure <= 0` guard just below).

---

ginderjeremiah (commented):
## Review: Combat-rating calibration report (#1533)

**Verdict:** Approve *(submitted as a comment — GitHub won't accept a gating verdict from the PR author)* — solid, well-tested tuning tooling with clean scope discipline. One should-fix on the `k`-derivation accuracy and a couple of nits; none block, so I'm not merging on your behalf so they can be addressed first.

The change delivers exactly what #1533 asks for: a re-runnable report comparing `SumCoreAttributes` vs `CombatRating` per enemy, zone placement on the `r = enemy/player` axis under both measures, and recommended `k`/pie constants for the #1532 swap — with XP/hour driven by real `BattleSimulator` runs rather than a formula approximation, as the issue specified. It stays read-only and touches no runtime reward math, matching the issue's "report a human reads, not a runtime system" framing.

I verified the supporting production changes against the codebase:
- `Enemy.ToBattler()` is a **faithful** extraction — `BattleService.cs:511` and `OfflineProgressSimulator.cs:114` both construct the battler identically (`new Battler(new AttributeCollection(enemy.GetAttributeModifiers()), enemy.BattleSkills, enemy.Level)`), so leaving those two live-path sites for follow-up #1578 is a defensible scope call.
- The reference builds' `OldMeasure` uses `GetModifiersWithSignaturePassive`, which is exactly what production feeds `DefeatRewards` for player power (`BattleService.cs:423`) — the "old measure" the report shows is the real one, not a re-derivation.
- Widening `SumCoreAttributes` and `Battler` construction to read the same code paths (rather than duplicating them) is the right instinct for a calibration tool whose whole value is fidelity to production.

### Should-fix

- **`CombatRatingCalibrator.cs:128-132`** — the `k` derivation treats the anchor as band-matched (`DifficultyMultiplier == 1`) but never verifies it; the anchor is only the *closest*-to-1 sample. If no sample lands inside the ±20% band, `k` is silently miscalibrated (a `ratio = 0.5` anchor over-estimates `k` by 4×). Fold the actual old multiplier into the anchor XP. Details inline.

### Nits

- **`CombatRatingCalibrator.cs:120`** — null-forgiving `!` runs against `CLAUDE.md`'s "avoid `!`" rule; foldable into the empty-check. Plus a minor `Math.Log(0)` degenerate-input note. Inline.

### Test coverage

Strong and appropriately layered. Unit tests (`CombatRatingCalibratorTests`) cover the pure math — shares summing to 1, the dead-stat relative-shift, spawn-weighting, anchor selection, `k`/pie reproduction, and reward-curve determinism — DB-free and classical style, matching the project's testing guidance. The integration test wires the real seeded catalog and asserts structural invariants rather than pinning numbers, which is the right call for a report whose job is to inform tuning, not gate it. One gap worth closing alongside the should-fix: `RecommendConstants_XpScaleK_ReproducesTodaysXpAtTheAnchor` only exercises `ratio = 1.0`, so it can't catch the off-band `k` error above — add a case where the anchor sits outside the ±20% band. No frontend mirror is needed here: this is backend content tooling, not battle-parity logic.

### What's good

- Genuinely careful fidelity to production (shared measure, shared battler construction, real simulator for time-per-kill) — the tool measures what battles actually do.
- Thorough XML docs that explain *why* (share-based `RelativeShift`, stateful-`Battler`-so-rebuild-per-call, deterministic hashed seeds) without over-narrating.
- Honest, well-scoped PR description: the placeholder-content caveats and the deliberately-deferred #1578 migration are called out rather than glossed.

---

ginderjeremiah:
## Review — Combat rating calibration report

Verdict: **Approve, with one thing to address before merge.** Clean, well-tested content tooling that delivers #1533 as specified, and the placement in `Game.Application/Content/Calibration` is consistent with the existing content-tooling family (progression-graph lint, content exporter, content-health).

**Verified against the code (not just the description):**
- The calibration math is sound. `k = (E_old × mult_old) / (E_new × min(r_new,1)²)` correctly reproduces today's XP payout at the anchor, folding in the *real* `DifficultyMultiplier` rather than assuming a matched anchor; `pie_new = pie_old × (P_new / P_old)` preserves accrual share across the measure-scale change; and log-space anchor selection (`Abs(Log(ratio))`) is symmetric. Unit tests pin these.
- `Enemy.ToBattler()` is byte-for-byte identical to the two duplicated production sites (`BattleService`, `OfflineProgressSimulator`), so it's behavior-preserving, and leaving those two live sites for #1578 is a reasonable scope call for a tooling PR.
- The integration test's "old measure" uses `GetModifiersWithSignaturePassive` — the same set production `DefeatRewards` reads — so the old-vs-new comparison is faithful, not a re-derivation.

**Should-fix (before merge):** `SimulateMatchups` seeds battles via `HashCode.Combine(zone.ZoneId, level, build.Name, spawn.EnemyId, i)`. Both `HashCode.Combine` **and** `string.GetHashCode()` (on `build.Name`) are randomized per-process in .NET by default, so the "deterministic / reproducible across runs against unchanged content" claim only holds *within a single process*. A fresh run yields different win-rates and battle durations in the XP/hour curve, which undercuts the tool's stated use for drift detection — and `ComputeRewardCurve_IsDeterministicAcrossRuns` passes only because it runs twice in one process, giving false confidence. Note the recommended *constants* (the actual #1532 deliverable) are unaffected, since they don't touch the simulation — this only wobbles the illustrative per-hour figures. Fix is small: derive the seed from a stable hash (arithmetic-combine the integer ids; use the build index instead of `build.Name`), or soften the claim. Fine to land here or as a small tracked follow-up.

**Nits (non-blocking):**
- `PriceEnemies` divides by `oldTotal`/`newTotal` with no zero guard — unreachable with real content (the `RecommendConstants` path is properly guarded), but worth a defensive check.
- The `PlayerOldMeasure <= 0` guard branch in `RecommendConstants` isn't unit-pinned.
- Making both `DefeatRewards.SumCoreAttributes` and `GetDifficultyMultiplier` public widens a `Game.Core` domain surface for a tooling consumer — both are pure/stateless so it's acceptable, just worth being conscious of.

CI is fully green. Holding the merge pending the seed-determinism decision (fix in-PR or a tracked small follow-up) — everything else is solid.

---

ginderjeremiah:
## Review: Approve

Clean, well-scoped tooling PR. Verified against the diff and the surrounding code (`CombatRating`, `DefeatRewards`, `Enemy`, the seeded-content integration path).

**Architecture / isolation** — `CombatRatingCalibrator` in `Game.Application/Content/Calibration` sits with the existing content-tooling family (`IContentExporter`, `IProgressionGraphChecker`), stays pure over already-resolved domain objects, and pushes all data-access wiring into the integration test. Correct placement and DDD isolation.

**Visibility changes** — making `DefeatRewards.SumCoreAttributes` and `GetDifficultyMultiplier` public is the right call, not a domain leak. They are pure, side-effect-free static measures; the calibrator lives in a separate assembly, so the alternative is `InternalsVisibleTo` to a tooling/test consumer, which is uglier than exposing a pure function. More to the point, the tool's whole purpose is comparing against the *exact* production "old measure" — re-deriving it would defeat the calibration. Folding the real `DifficultyMultiplier` into the anchor (rather than assuming a matched ratio-1 anchor) is a genuine correctness fix; the off-band-anchor test pins it.

**`Enemy.ToBattler()`** — clean extraction of a construction that was already duplicated across `BattleService` and `OfflineProgressSimulator`, well documented. Deferring the two production call-site migrations to #1578 is the right scope boundary — keeping the live battle-parity / offline-replay paths untouched in a tooling PR.

**Tests** — unit tests pin the pure math meaningfully (share-sum-to-one, dead-LUK relative-shift, spawn weighting, log-space anchor selection, k reproduction, off-band multiplier fold-in, pie rescale, determinism). The integration test asserting structural sanity rather than pinned numbers is exactly right for a tuning report — pinning the output would make it a brittle gate on content that is meant to change.

**Minor observations (non-blocking, no action needed):**
- `PriceEnemies` / `RecommendConstants` don't guard a degenerate all-zero old-measure population (NaN/Infinity shares). Consistent with the code's stated philosophy — the `RecommendConstants` comment already calls this out as not worth guarding for real content, and I agree.
- The pre-existing "seed class attributes don't scale with level, so enemies outscale the player in the second zone" finding is a content-tuning concern on placeholder seed data (exactly what this tool exists to surface), correctly left to content authoring rather than fixed here.

CI green (backend, e2e chromium/firefox/webkit, `all-checks-passed` all success; frontend jobs correctly skipped). Docs note in `backend.md` is concise and in the right section. Approving and merging.

---

PR #1580, opened 2026-07-04
Content lint: flag enemy attribute distributions the enemy's kit can't consume

### LINKED ISSUE #1529: Content lint: flag enemy attribute distributions the enemy's kit can't consume

ginderjeremiah:
## Summary

Implements #1529: a new `EnemyInertAttribute` warning on the progression-graph lint (`ProgressionGraphChecker`, part of the #1390 graph-lint family). For each live enemy, flags a core-attribute (STR/END/INT/AGI/DEX/LUK) distribution point that nothing in the enemy's fielded kit consumes — dead weight that still counts into `DefeatRewards.SumCoreAttributes`, inflating the enemy's XP payout without adding threat.

**Consumed** = a static derivation targets the attribute (Endurance → Toughness/MaxHealth, Strength → MaxHealth — always live, no enabler needed), a pooled live skill's `DamageMultipliers` scales off it, or a pooled live skill effect's `ScalingAttributeId` does (with a nonzero `ScalingAmount`).

Only live skills in the enemy's pool count as consumers (a retired skill is out of circulation), and a zero-valued distribution entry (`BaseAmount == 0 && AmountPerLevel == 0`) is skipped rather than flagged.

## A note on the AGI/LUK rule (worth a second look)

The issue's own text says "AGI/LUK: always flagged for enemies" as a blanket rule. I verified the premise against the current engine (post the #1523/#1524/#1525 severings, which had landed on `main` since the issue was filed — the issue had actually been triage-deferred pending them, and that condition has now cleared) and found a wrinkle worth calling out:

- **Luck**'s only static derivations (`CriticalChanceMultiplier`, `ParryChanceMultiplier`, `CriticalDamage`) are all genuinely inert for enemies — `BattleContext.DamageTarget` never rolls crit/dodge/parry for a non-player battler, confirmed by reading the code. So "always flagged" is exactly right for Luck.
- **Agility** is a little different: it also derives `CooldownBonusMultiplier`, and unlike the crit/dodge/parry family, `Battler.GetCooldownMultiplier()` applies symmetrically to *any* battler (`CombatRating.Classify` even calls this out explicitly: "Symmetric (enemies cycle too), so not asymmetry-gated"). So if an enemy directly authors a `CooldownBonus` distribution point (the design doc for enemies explicitly allows this — "fast enemies author CooldownRecovery/CooldownBonus directly"), its Agility investment *would* have a real, non-dead effect via that multiplier.

Rather than blindly flagging Agility/Luck unconditionally, I implemented it as: **excepted from the static-derivation consumption path** (since their static derivations are the enemy-inert opt-in-multiplier family), but still eligible via the general kit-consumer check (a `DamageMultiplier` or effect scaling authored directly on Agility/Luck). In today's authored content this produces the same practical result as "always flagged" (no such skill exists yet), but it's more correct if/when an enemy skill is ever authored to scale off Agility/Luck directly, and it avoids hard-coding an exception that would need revisiting the moment enemy `CooldownBonus` authoring becomes common. Flagging this explicitly in case the blanket "always" reading was intentional and I've over-thought it — happy to simplify if so.

## Notes

- The heuristic is deliberately v1, per the issue's own scope note: the combat-rating spike (#1526) shipped `CombatRating.Marginal` (#1531), whose doc comment literally says it "doubles as the dead-stat detector the enemy content lint (#1529) needs." I didn't build on it here because it requires assembling a full `Battler` (attributes + fielded skills + a representative level) from the lightweight `Contracts` DTOs this lint otherwise never needs — a bigger dependency/architecture change than a "scope: small" issue calls for. Filed #1581 to track migrating this check onto `CombatRating.Marginal` for exact per-enemy detection once that's worth the lift.
- Verified against the committed `content/*.json`: all authored enemies only distribute Strength/Endurance today, so this check produces no new findings there — `ContentGraphLintTests` stays green with no `AcceptedWarnings` changes needed.
- Added a short mention of the new warning example to `docs/backend.md`'s progression-graph-lint section.

## Test plan

- [x] `dotnet test` — full suite passes (2889 tests, no regressions)
- [x] New unit tests (`ProgressionGraphCheckerTests`) cover: STR/END always consumed, AGI/LUK/INT flagged with no kit consumer, AGI/LUK consumed via a pooled `DamageMultiplier`/effect, a zero-`ScalingAmount` effect not counting as a consumer, a retired pooled skill not counting as a consumer, a zero-valued distribution entry not flagged, and a retired enemy not checked at all.

Closes #1529


Game.Application/Content/ProgressionGraphChecker.cs
```diff
@@ -194,6 +197,69 @@ private void CheckEnemies()
...
+                }
+            }
+
+            /// <summary>
+            /// Consumed = a static derivation targets it (Endurance/Strength feed the always-live
+            /// Toughness/MaxHealth), a pooled skill's <c>DamageMultipliers</c> scales off it, or a pooled skill
+            /// effect's <c>ScalingAttributeId</c> does. Agility and Luck are excepted from the static-derivation
+            /// half: their only derivations (the crit/dodge/parry-multiplier family) gate behind mechanics the
+            /// engine never rolls for an enemy (<c>BattleContext.DamageTarget</c>'s parry/dodge/crit draws are
+            /// player-only), so — until enemy proc parity exists — they're consumed only by a direct kit hit.
```
ginderjeremiah: **Should-fix (comment accuracy):** This says Agility and Luck's "only derivations (the crit/dodge/parry-multiplier family)" gate behind player-only mechanics — but that's only fully true for **Luck** (`CriticalChanceMultiplier`/`ParryChanceMultiplier`/`CriticalDamage`, all enemy-inert per `BattleContext.DamageTarget`). **Agility** also derives `CooldownBonusMultiplier` (`StaticAttributeModifiers.cs:85-86`), which is *not* in that family and *is* applied symmetrically to enemies. Its actual justification for being excepted is the one your PR description spells out and #1581 records: `CooldownBonusMultiplier` only does anything when the enemy also authors a `CooldownBonus` distribution point (0 × mult = 0), so Agility *alone* is still dead weight. The behavior is right and matches the issue's "always flag AGI/LUK" intent — but a future maintainer reading just this comment would wrongly conclude Agility has no enemy-live derivation at all. Suggest tightening it to name `CooldownBonusMultiplier` as the one symmetric-but-enabler-gated derivation rather than lumping Agility into the never-rolled family.

---

Game.Application/Content/ProgressionGraphChecker.cs
```diff
@@ -194,6 +197,69 @@ private void CheckEnemies()
...
+            /// </summary>
+            private static bool IsConsumedByKit(EAttribute attribute, IReadOnlyList<Contracts.Skill> kit)
+            {
+                if (attribute != EAttribute.Agility && attribute != EAttribute.Luck && HasStaticDerivation(attribute))
+                {
+                    return true;
+                }
+
+                return kit.Any(skill => skill.DamageMultipliers.Any(m => m.AttributeId == attribute)
```
ginderjeremiah: **Nit:** The effect-scaling arm guards `e.ScalingAmount != 0` (a zero amount means the field is just unset, correctly covered by a test), but the `DamageMultipliers` arm matches on `AttributeId` alone with no analogous `m.Multiplier != 0` guard. A `Multiplier == 0` entry is equally inert, so for symmetry with the effect check it'd be marginally more robust to guard it too. Low priority — a zero-valued `DamageMultiplier` entry is unlikely in practice since such an entry exists only to declare a multiplier, so take it or leave it.

---

ginderjeremiah (commented):
## Review: Enemy dead-stat content lint — `EnemyInertAttribute` warning (#1529)

**Verdict:** Approve *(submitted as a comment — GitHub won't accept a gating verdict from the PR author)* — the change does exactly what #1529 asks, is thoroughly tested, and stays cleanly within the existing `ProgressionGraphChecker` patterns. Two non-blocking findings (a comment-accuracy should-fix and a small nit); I'm not merging on your behalf so the comment fix can land first.

The lint flags a live enemy's core-attribute distribution point that nothing in its kit consumes — dead weight that still counts into `DefeatRewards.SumCoreAttributes` and inflates XP without adding threat. It slots in as one more `Run()` check, reuses `Live(...)`/`Warn(...)`, and holds to the house rules (braced blocks, no `!`, concise comments).

I verified the load-bearing correctness claims against the engine:
- **Enemies genuinely never crit/dodge/parry.** `BattleContext.DamageTarget` (`BattleContext.cs:214-232`) rolls the crit draw only on the player-active branch (`ResolvePlayerHit`, `_activeBattler`), and the parry/dodge draws only on the enemy-active branch against `_targetBattler` — the *player* defending. So an enemy's Luck (crit/parry multipliers) and Agility (dodge multiplier) derivations are never read → excepting them from the static-derivation path is sound.
- **The heuristic reads the right sources.** `HasStaticDerivation` deriving the consumed set from `StaticAttributeModifiers.All` `Derived`/`DerivedSource` entries (rather than hard-coding STR/END) is exactly what the issue's triage note asked for, and stays correct across the #1523/#1524/#1525 severings.
- **Zero-amount / retired handling is right:** zero-valued distribution entries and retired enemies/skills are all correctly skipped, each with a test.

### Should-fix

- **`ProgressionGraphChecker.cs:245-247`** — the inline justification for excepting Agility/Luck overstates the case: it's fully accurate for Luck, but Agility *also* derives `CooldownBonusMultiplier`, which is symmetric (enemies cycle too) and isn't in the crit/dodge/parry family. The exception is still correct because that multiplier only bites when the enemy co-authors a `CooldownBonus` (0 × mult = 0), so Agility alone is still dead — but the comment as written would mislead a future reader. Details inline. (This is behavior-correct and matches the issue's "always flag AGI/LUK" directive; it's purely the comment.)

### Nit

- **`ProgressionGraphChecker.cs:256`** — the `DamageMultipliers` consumer check has no `Multiplier != 0` guard to mirror the effect arm's `ScalingAmount != 0`. Marginal; a zero-valued multiplier entry is unlikely in practice. Inline.

### Test coverage

Strong and well-targeted (`ProgressionGraphCheckerTests`): STR/END consumed via static derivation, AGI/LUK/INT flagged with no kit consumer, AGI/LUK rescued by a pooled `DamageMultiplier`/effect scaling, a zero-`ScalingAmount` effect *not* counting as a consumer, a retired pooled skill not excusing the point, a zero-valued distribution skipped, and a retired enemy not checked. The `Skill`/`Enemy` builder extensions are clean and backward-compatible. No frontend mirror needed — this is backend content tooling, not battle-parity logic. The one edge not directly tested (Agility rescued by an effect vs. a `DamageMultiplier`) is covered symmetrically by the LUK effect-scaling case, so no gap worth blocking on.

### Follow-up

- #1581 (already filed) tracks migrating this heuristic onto `CombatRating.Marginal` for exact per-attribute detection once the #1533 Battler-construction infra lands — and it explicitly captures the Agility/`CooldownBonus` edge above, which is the right home for it. Good scope discipline keeping that out of a "scope: small" issue.

### What's good

- Careful engine verification behind the AGI/LUK exception rather than blindly implementing the issue's blanket "always flag" rule, with the nuance surfaced honestly in the PR description for a second look.
- Deriving the consumed set from `StaticAttributeModifiers.All` so the lint can't drift from the engine's actual derivations.
- Honest, well-scoped PR: the v1-heuristic framing, the deferred #1581 migration, and the "no new findings on committed content" verification are all called out.


---

ginderjeremiah:
Pushed ee8cde0 addressing both findings: tightened the Agility/Luck comment to correctly attribute Agility's exception to `CooldownBonusMultiplier`'s enabler-gating rather than lumping it in with Luck's never-rolled derivations, and added the `Multiplier != 0` guard for symmetry with the effect-scaling arm (plus a test for it). Full suite + format still green.

---

ginderjeremiah:
## Review — Content lint: enemy inert-attribute warning

Verdict: **Approve.** I verified the implementation against the engine rather than trusting the description, and it holds up.

**Correctness (verified against the code):**
- `CoreAttributes` is exactly `{STR, END, INT, AGI, DEX, LUK}`, and the STR/END → MaxHealth/Toughness static derivations are always-live, so treating STR/END as never-flagged is correct — and is the *more* correct reading than the issue's naive "STR/DEX/INT need a skill" phrasing (it matches the issue's own detailed Consumed list).
- Enemies never crit/dodge/parry — confirmed in `BattleContext.DamageTarget` (the crit draw is gated on the active battler being the player, and parry/dodge read the player as target). So Luck's whole derivation family and Agility's dodge multiplier really are inert for enemies.
- The zero-coefficient guards (`Multiplier != 0`, `ScalingAmount != 0`) are load-bearing precisely because `SkillEffect.ScalingAttributeId` is a non-nullable `EAttribute` defaulting to `Strength` (enum 0) — without them an unset effect would falsely "consume" Strength. Both zero cases are covered by tests. The 11 new unit tests hit every branch. Clean code (braces everywhere, no null-forgiving operator, DRY), and the `backend.md` note is accurate.

**On the AGI/LUK nuance you flagged:** your reasoning is right — Agility's `CooldownBonusMultiplier` is symmetric (`Battler.GetCooldownMultiplier()` applies it to any battler, and `CombatRating` even annotates "enemies cycle too"), so an enemy that co-authors a non-zero `CooldownBonus` distribution point *and* invests Agility would genuinely gain threat, yet this heuristic would still flag that Agility as inert. That's a real (if currently untriggered) false positive. The good news is it's already captured deliberately: **#1581 explicitly documents this exact case** and notes the `CombatRating.Marginal` migration resolves it automatically. So this is a tracked, intentional deferral rather than a latent gap — no further action needed on it here.

**Nits (non-blocking):**
- Test comment at the Agility case labels its derivations "the crit/dodge/parry-multiplier family" — that's actually Luck's family; Agility derives DodgeChanceMultiplier + the symmetric CooldownBonusMultiplier. Worth correcting since it glosses over the exact case #1581 tracks.
- The `attribute != Agility && attribute != Luck` exception is a name-based proxy for "enemy-inert opt-in multiplier vs. symmetric derivation" and will need revisiting if the derivation map shifts — acceptable for a v1 heuristic, and #1581 supersedes it anyway.

Holding the merge only so those two trivial nits can be tidied (or explicitly waved off) — the substance is correct, tested, and CI is green. Happy to merge on your say-so.

---

ginderjeremiah:
## Review — `EnemyInertAttribute` lint (#1529)

Verdict: **Approve with one nit.** I verified the combat reasoning against the engine directly rather than the description, and the substance holds up.

### Correctness (verified against source)
- `Attribute.CoreAttributes` is exactly `{STR, END, INT, AGI, DEX, LUK}`, and `MaxHealth ← Strength/Endurance` + `Toughness ← Endurance` are always-live static derivations — so treating STR/END as never-flagged is right, and matches the issue's own detailed Consumed list (a better reading than its looser "STR/DEX/INT need a skill" phrasing).
- Deriving `HasStaticDerivation` off `StaticAttributeModifiers.All` rather than hardcoding a list is the right call — single source of truth, so it can't drift if the derivation map changes.
- Enemies never crit/dodge/parry: confirmed in `DamageTarget` — the crit draw is gated on the player being the active battler, and parry/dodge read the player-defender's stats. So Luck's entire derivation family (Critical/Parry) and Agility's `DodgeChanceMultiplier` really are inert for enemies.
- The zero-coefficient guards (`Multiplier != 0`, `ScalingAmount != 0`) are load-bearing: `SkillEffect.ScalingAttributeId` is a non-nullable `EAttribute` defaulting to `Strength` (enum 0), so without the guard an unset effect would falsely "consume" Strength. Both zero cases are tested.

### On the AGI/LUK deviation from the issue
Your reasoning is correct, and I confirmed it: `Battler.GetCooldownMultiplier()` returns `CooldownRecovery + CooldownBonus × CooldownBonusMultiplier` with **no player gating** — it applies to any battler. So Agility's `CooldownBonusMultiplier` derivation is genuinely symmetric, unlike Luck's crit/parry family. An enemy that co-authors a nonzero `CooldownBonus` distribution *and* invests Agility would gain real threat, yet this heuristic would still flag that Agility as inert — a true (currently untriggered) false positive. That's fine: it's excepted from the static-derivation path and left to the direct-kit-consumer check, no enemy authors such a kit today, and **#1581 explicitly documents this exact case** and the `CombatRating.Marginal` migration that resolves it automatically. Deviating from the issue's blanket "always flag AGI/LUK" is the more-correct choice, not an over-think.

Domain isolation is respected — the check stays on the lightweight `Contracts` DTOs + `Game.Core` static tables and deliberately avoids pulling `Game.Core.Battle`/`Battler` construction into the lint, with #1581 tracking that as the real architecture step. 11 new tests cover every branch; DRY (reuses `Live`/`Warn`/`_skills`), braces everywhere, no null-forgiving operator; the `backend.md` note is accurate.

### Nit (non-blocking, but worth tidying before merge)
- The test comment on `Enemy_DistributesAgilityWithNoKitConsumer_IsWarning` says Agility's derivations "feed the crit/dodge/parry-multiplier family, which the engine never rolls for an enemy." That's inaccurate — Agility derives `DodgeChanceMultiplier` + the **symmetric** `CooldownBonusMultiplier` (not crit/parry), and `CooldownBonusMultiplier` *is* applied to enemies. It's exactly the subtlety the production comment in `IsConsumedByKit` (correctly, after commit 2) and #1581 capture, so the test comment now contradicts them. A one-line fix.

CI is green (all-checks-passed). Substance is correct, tested, and merge-ready modulo that stale test comment — I'd fix it (or wave it off) and then this is a clean merge. Holding the merge only for that.


---

PR #1582, opened 2026-07-04
Extend AdminTools Ops page to inspect/replay the socket command dead-letter queue

### LINKED ISSUE #1542: Extend the AdminTools Ops page to inspect/replay the socket command dead-letter queue

ginderjeremiah:
## Summary

Implements #1542 (follow-up to #1498): mirrors the existing player write-behind dead-letter Ops surface (#794) for the socket command dead-letter queue, which previously only had its depth logged on escalation.

- **`SocketCommandDeadLetters`** (`Game.Api/Services/Admin`) — a guarded inspect/replay service over `Constants.PUBSUB_SOCKET_DEAD_LETTER_QUEUE`, reusing the **existing** `DeadLetterEntry`/`DeadLetterInspection`/`DeadLetterReplayResult`/`ReplayDeadLettersData`/`EDeadLetterReason` contracts (widened their doc comments to cover both queues) rather than introducing near-duplicate types. It lives in `Game.Api` rather than the data tier because replay dispatches through `SocketManagerService.EmitSocketCommand` and classification reads `SocketCommandFactory` — both Api-tier types.
- **Addressing decision (the issue's open question):** a bare `SocketCommandInfo` carries no player/socket id once dequeued, and the specific socket a command originally failed on may no longer exist by the time an operator replays it. `EscalateFailedServerCommand` (`SocketManagerService`) now wraps the dead-lettered payload in a `SocketCommandDeadLetterEnvelope` carrying the **player id** it was addressed to. Replay redelivers via `EmitSocketCommand(command, playerId)`, which resolves whatever socket is **currently** live for that player — the natural fix for reconnects, since targeting the original (possibly stale) socket id wouldn't work anyway.
- **Never silently drops an entry:** if the player has no live socket right now (or the envelope doesn't parse), the entry is left queued rather than counted as replayed — there's no second queue to "reserve onto" first the way the player write-behind replay does, so redelivery only proceeds once it can actually succeed. An operator can retry later once the player reconnects.
- **Controller:** two new actions (`GetSocketCommandDeadLetters`/`ReplaySocketCommandDeadLetters`) added to the existing `AdminDeadLettersController`, same `/api/AdminTools/*` route family and Admin-role gate.
- **Frontend:** the existing `DeadLetterConsole.svelte`/`DeadLetterConsoleState` are parameterized by a `variant` (`'player-update' | 'socket-command'`) rather than duplicated — both queues share the same backend contracts, so only the routes and a little copy text differ. New "Socket Dead Letters" entry in the admin Ops nav group.

## Test plan

- [x] `dotnet test` — full suite passes (2889 tests, no regressions)
- [x] New backend tests: `SocketCommandDeadLettersIntegrationTests` (classification, replay against a real live-socket presence key, and the "no live socket → left queued" case) and new cases on `AdminDeadLettersControllerTests` (Admin-role gate, classify, replay-all over HTTP)
- [x] `npm run test:unit:ci` — full suite passes (3379 tests, no regressions), including new variant coverage for `DeadLetterConsoleState`/`DeadLetterConsole`

## Docs

- `docs/backend-persistence.md` — new paragraph on the socket queue's replay-by-player-id design, distinct from the player queue's re-enqueue.
- `docs/backend.md` — updated the challenge-completion notifications section to note the socket dead-letter queue now has an inspect/replay Ops surface.

Closes #1542

---
_Generated by [Claude Code](https://claude.ai/code/session_01KWPVcWwe8e6UL32LZvHxm9)_


Game.Api/Services/Admin/SocketCommandDeadLetters.cs
```diff
@@ -0,0 +1,139 @@
...
+            var replayed = 0;
+            foreach (var payload in targets)
+            {
+                var envelope = TryDeserialize(payload);
+                if (envelope is null || !await _socketManager.HasActiveSocket(envelope.PlayerId))
+                {
+                    continue;
+                }
+
+                await _socketManager.EmitSocketCommand(envelope.Command, envelope.PlayerId);
```
ginderjeremiah: **Should-fix:** replay never skips `UnknownEventType` entries, so a poison entry gets emitted to the live client and dropped from the queue.

The only guard here is `envelope is null` (which catches `Malformed`) and the live-socket check. A *parseable* command whose name is no longer server-initiated — e.g. a command dead-lettered, then renamed/removed in a later deploy, then replayed — falls through to `EmitSocketCommand` and then `RemoveAsync`. That ships a stale command the client won't recognize **and** silently drops the entry from the dead-letter queue.

That contradicts two things this PR states explicitly:
- the class doc's "left queued rather than silently discarded" / "never lost" guarantee, and
- the UI copy (`DeadLetterConsole.svelte`): *"malformed and unknown command type entries are poison and will return to the queue immediately."*

Unlike the player-update queue, "return to the queue" is emergent there because a poison entry round-trips through the synchronizer and re-dead-letters — the socket queue has no such re-processing, so replaying poison just delivers it and removes it. It doesn't come back.

Suggest gating the emit on the same classification `Classify` already computes, e.g.:

```csharp
if (envelope is null
    || !_commandFactory.IsServerInitiatedOnly(envelope.Command.Name)
    || !await _socketManager.HasActiveSocket(envelope.PlayerId))
{
    continue;
}
```

so `UnknownEventType` entries stay queued the way `Malformed` ones already do.

---

Game.Api/Services/Admin/SocketCommandDeadLetters.cs
```diff
@@ -0,0 +1,139 @@
...
+                : EDeadLetterReason.UnknownEventType;
+            return entry;
+        }
+
+        private static SocketCommandDeadLetterEnvelope? TryDeserialize(string raw)
+        {
+            try
+            {
+                var envelope = raw.Deserialize<SocketCommandDeadLetterEnvelope>();
+                return string.IsNullOrEmpty(envelope?.Command.Name) ? null : envelope;
```
ginderjeremiah: **Nit (robustness):** `envelope?.Command.Name` can throw an uncaught `NullReferenceException`. `Command` is `required`, but `required` only means the key must be *present* — a payload of `{"playerId":5,"command":null}` deserializes fine (required satisfied) with `Command == null`, so `envelope.Command.Name` NREs. The `catch` only handles `JsonException`, so this escapes and aborts the entire `InspectAsync`, failing the whole page rather than classifying the one bad entry as malformed.

Only reachable via manual Redis tampering (the escalation path always writes a real command), so low severity — but the sibling `PlayerUpdateDeadLetters.Classify` defends against exactly this (`envelope is null || string.IsNullOrEmpty(envelope.Type)`). A `envelope?.Command is null` guard here would match that parity.

---

ginderjeremiah (commented):
## Review: Socket command dead-letter inspect/replay Ops surface (#1542)

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't take a gating review from me)* — no blocking findings; one should-fix and two nits below. CI is green across the board.

This is a clean, well-scoped implementation that mirrors the player write-behind dead-letter surface (#794) without duplicating it. The addressing decision (envelope carries the **player id**, replay resolves whatever socket is currently live via `EmitSocketCommand(command, playerId)`) is the right call and well justified — targeting the original socket id would be pointless after a reconnect. Contracts, the frontend console, and the docs are all reused/updated thoughtfully.

### Should-fix

- **`SocketCommandDeadLetters.cs:65-70` — poison (`UnknownEventType`) entries are replayed and dropped, not left queued.** The replay guard only skips `Malformed` (via `envelope is null`) and no-live-socket entries. A parseable-but-unregistered command (dead-lettered, then renamed/removed in a later deploy) gets emitted to the live client and removed from the queue — silently dropped from a queue whose whole job is preservation. This contradicts both the class doc's "never lost" guarantee and the UI copy that promises "unknown command type entries are poison and will return to the queue immediately." Unlike the player queue, there's no re-processing round-trip that makes poison come back. Gate the emit on `IsServerInitiatedOnly` (the same classification `Classify` already computes). See inline for a suggested guard.

### Nits

- **`SocketCommandDeadLetters.cs:131` — uncaught `NullReferenceException` on `{"command":null}`.** `required` only guarantees the key is present, not non-null, so `envelope?.Command.Name` can NRE and escape the `JsonException`-only catch, aborting the whole `InspectAsync`. Only reachable via Redis tampering, but the sibling `PlayerUpdateDeadLetters.Classify` guards this. (inline)
- **`SocketCommandDeadLetters.cs:65 & 70` — TOCTOU + redundant socket lookup.** `HasActiveSocket(playerId)` and `EmitSocketCommand(command, playerId)` each resolve the socket independently; if the player disconnects between them, the emit no-ops (it swallows the no-socket case and returns `void`) but `RemoveAsync` still runs — another silent drop. The window is tiny and socket delivery is best-effort anyway, so impact is low. Having `EmitSocketCommand(SocketCommandInfo, int)` report whether it actually delivered, and gating the remove on that, would both close the race and let you drop the now-redundant `HasActiveSocket` pre-check. Optional.

### Test coverage

Strong. Backend integration tests cover classification (malformed / unknown / replayable), player-id derivation, the `max`-vs-full-depth split, replay-all/replay-selected against a real live-socket presence key, the "no live socket → left queued" case, and — nicely — the security case that a fabricated payload can't be injected onto a live socket. Controller tests cover the Admin-role gate and the HTTP shape. Frontend adds variant coverage for both the console and its state. The one gap that maps to the should-fix: there's no test asserting an `UnknownEventType` entry is *left queued* on replay — worth adding alongside the fix.

### What's good

- Reusing the `DeadLetterEntry`/`Inspection`/`ReplayResult` contracts and parameterizing `DeadLetterConsole` by `variant` rather than forking a near-identical component.
- The snapshot-validated selection (`FilterToSnapshot`, multiplicity-aware) closing off arbitrary-payload injection, with a dedicated test proving it.
- Placing the service in `Game.Api` (not the data tier) is correctly reasoned — it depends on `SocketManagerService`/`SocketCommandFactory`, both Api-tier — and the docs capture the replay-by-player-id design distinctly from the player queue's re-enqueue.

---

ginderjeremiah:
Fixed in a1d537a:

- **Should-fix:** `ReplayAsync` now gates on `IsServerInitiatedOnly` (the same classification `Classify` computes) before emitting, so an `UnknownEventType` entry stays queued instead of being delivered and dropped.
- **Nit (NRE):** `TryDeserialize` now guards `envelope?.Command is null` explicitly, so a `{"command":null}` payload classifies as `Malformed` instead of escaping `InspectAsync` uncaught.
- **Nit (TOCTOU, optional):** took this one too — `SocketManagerService.EmitSocketCommand` now returns `Task<bool>` (whether a live socket was found), so replay gates its queue removal on one lookup instead of the separate `HasActiveSocket` + `EmitSocketCommand` pair. Non-breaking for the existing callers (`Task<bool>` derives from `Task`).

Added integration tests for both the should-fix (`ReplayAllAsync_LeavesAnUnknownCommandEntryQueued_EvenWithALiveSocket`) and the NRE nit (`InspectAsync_ClassifiesANullCommandEnvelopeAsMalformed_WithoutThrowing`). Full backend suite (2891 tests) and `dotnet format --verify-no-changes` are green.

---

ginderjeremiah:
## Review: Approve

Reviewed against CLAUDE.md and the backend/persistence docs. This is a clean, well-reasoned extension of the existing dead-letter Ops surface. Verified the diff locally at `a1d537a`.

**Verdict: Approve.** CI green (9/9). Merging.

### What I verified
- **DRY / contract reuse:** `DeadLetterEntry`/`DeadLetterInspection`/`DeadLetterReplayResult`/`ReplayDeadLettersData`/`EDeadLetterReason` are reused with only doc-comment widening — no near-duplicate types. The frontend `DeadLetterConsole` is parameterized by a `variant` prop rather than forked, and `DeadLetterConsoleState` routes on a `QUEUE_ROUTES` map. Good.
- **Tier placement:** `SocketCommandDeadLetters` correctly lives in `Game.Api` (not the data tier like `PlayerUpdateDeadLetters`) because replay dispatches through `SocketManagerService.EmitSocketCommand` and classification reads `SocketCommandFactory` — both Api-tier. The asymmetry with the player queue is documented in `backend-persistence.md`.
- **Addressing decision:** `SocketCommandDeadLetterEnvelope` carries the `PlayerId`, and replay redelivers via `EmitSocketCommand(command, playerId)` → resolves whatever socket is *currently* live. This is the right call — targeting the original (stale) socket id would be pointless across a reconnect. Co-located with `SocketCommandInfo` in `AbstractSocketCommand.cs`, a sensible placement.
- **"Left queued, never dropped":** Confirmed in code — poison (malformed / unknown-command) and no-live-socket entries hit `continue` and stay on the queue; removal is gated on `EmitSocketCommand` returning true. Because `RemoveAsync` uses `ListRemoveAsync(..., count: 1)`, duplicate-payload multiplicity is handled correctly on both the all and snapshot-filtered selected paths.
- **Security:** Admin-role gated via `ServiceFilter`; replay validates the requested payloads against a live queue snapshot (`FilterToSnapshot`), so a fabricated/stale selection can't inject an arbitrary command onto a player's socket. Explicitly covered by `ReplaySelectedAsync_IgnoresPayloadsNotOnTheQueue_...` and the controller's non-admin `Forbidden` test.
- **Null-command guard:** `TryDeserialize` correctly handles `{"command":null}` (which passes `required`) as Malformed rather than NRE-ing — pinned by a test.
- **Test coverage:** Integration tests exercise classification, replay against a real presence key, no-live-socket, unknown-command-with-live-socket, snapshot-injection guard, and null-command. Controller tests cover the Admin gate, classification, and replay-all over HTTP. Frontend adds variant coverage for both the state class and the component. Presence-key / socket-queue-name formats in the tests match production `CurrentSocketKey`/`SocketQueueName`.
- **Docs:** `backend-persistence.md` and `backend.md` accurately updated (the latter's stale "no inspect/replay Ops surface yet" line is corrected); the cross-reference anchor resolves.

### Minor observations (non-blocking, no action needed)
- `EmitSocketCommand(commandInfo, socketId)` now returns `true` unconditionally — it signals "a socket was found to publish to," not delivery. Clearly documented, and existing callers ignore the result, so this is fine.
- Replay emits (fire-and-forget) then removes from the dead-letter queue, so a crash in that window can duplicate an entry (re-sync on reconnect absorbs it) but never lose it — consistent with the player queue's guarantee.

Solid work.


---

Issue #1598, opened 2026-07-04
Suppress/bound presentational side effects while the tab is hidden


ginderjeremiah:
Follow-up to #1594 (worker tick source), part of spike #1073.

Once the simulation runs at full rate while the tab is hidden, its per-tick presentational side effects run too — while `requestAnimationFrame` is suspended and page timers are throttled:

- **Combat floaters**: `CombatFloaters.svelte` removes each floater via `setTimeout(DURATION_MS + 80)`. Hidden-tab timer throttling batches those to ~once/minute, so up to a minute of floater DOM accumulates at a time while battles keep resolving in the background. They are purely presentational (`aria-hidden`); simplest fix is to skip `spawn()` while `document.hidden`.
- **Audit other per-battle-event UI producers** for the same pattern — e.g. toast duration timers on victory-driven toasts (challenge completions, proficiency milestones). The boss-victory overlay `delay(BOSS_VICTORY_OVERLAY_MS)` stretching under throttling just slows hidden boss-farming slightly — acceptable, no change.
- The combat log is already capped (`UI/src/stores/logs.svelte.ts` `maxLogEntries`) — no change needed.

No user-visible behavior change (the tab is hidden throughout); this is DOM-churn/battery hygiene so a long-hidden tab doesn't stutter on refocus.

Issue #1597, opened 2026-07-04
Client: resume a battle already in progress (replay-to-offset)


ginderjeremiah:
Part of spike #1073 — see `docs/spikes/1073-idle-time-recovery.md`. Client consumer of the "active battle + elapsed offset" payload #1595/#1596 produce.

## Problem

`BattleEngine.reset` (`UI/src/lib/engine/battle/battle-engine.ts:194-207`) always starts a battle at `timeElapsed = 0`. A battle the server hands back mid-flight (a still-in-progress stale battle from #1595, or the backdated trailing-remainder battle from #1596) must instead be reconstructed and fast-forwarded to the server-supplied offset.

## Fix

A resume entry point on `BattleEngine`: reset from the enemy instance/seed exactly as today, then run the shared `battleStep` in a headless loop up to the offset with the presentational side effects suppressed (per-hit `logMessage`, `notifyCombatFloat`, effect-application logging, DoT/HoT flush — the side effects `logicalUpdate` emits at `battle-engine.ts:323-383`), then continue live from the caught-up state. The replay must run the same victory/defeat/draw checks `logicalUpdate` runs so a battle that concludes within the offset resolves correctly (defensive — the server resolves concluded battles before handing anything back, but the tick boundary makes it possible).

Performance is a non-issue: the offset is bounded by `DefaultMaxBattleMs` (2 min) = at most ~3000 `battleStep` ticks.

## Architecture note

This makes the resume path a **third consumer of `battleStep`**, beyond the closed live-engine/parity-simulator pair `docs/frontend.md` currently documents. Accepted in the spike (owner sign-off, 2026-07-04) — update `docs/frontend.md`'s battle-parity section ("who drives `battleStep`") when this lands.

## Explicitly out of scope (spike decision)

A client-measured-gap fast-forward driven by `onIdleTimeLost` (the original design's §1). The worker tick source (#1594) covers throttling; a freeze-thaw short enough that the socket survives is an accepted, small loss that funnels into the reconnect path. If it proves meaningful in practice, wiring `onIdleTimeLost` to this primitive is a cheap follow-up — the hard part (this issue) already exists.

## Tests

- FE unit tests: replay-to-offset yields battler state identical to an uninterrupted live run to the same tick (state parity), and emits no log/float events during the replay.
- Battle-logic scenarios mirrored in the backend suite per the FE/BE parity testing rule where applicable.

Issue #1596, opened 2026-07-04
Offline catch-up: carry the trailing remainder as a backdated next battle


ginderjeremiah:
Part of spike #1073 — see `docs/spikes/1073-idle-time-recovery.md`. Closes the boundary-slice loss #879 accepted as a known limitation (`docs/spikes/879-offline-rewards.md`).

## Problem

`OfflineProgressSimulator.Simulate` (`Game.Core/Battle/Offline/OfflineProgressSimulator.cs:28-85`) credits whole battles until the remaining away budget is under one battle + cooldown, then drops the remainder on the floor — the live loop starts the next battle fresh at zero, so up to ~2 minutes of idle time evaporates on every welcome-back.

## Fix

After the crediting loop, set up the **next** battle (fresh enemy + seed via the existing spawn path) as the active battle backdated by the remainder: `SetActiveBattle(..., startTime: now - remainderMs)` — the overload already exists (`Game.Core/Entities/PlayerState.cs:55`). The player resumes elapsed-into-it by exactly the remainder; nothing is dropped.

Cooldown handling: the remainder is served against the post-battle cooldown first — if it falls entirely within the cooldown window, the next battle starts with the residual cooldown rather than a backdated fight; if it spans the cooldown, the excess becomes the new battle's elapsed offset.

No changes to the crediting loop's internals — this only touches what happens at its trailing edge (mirroring #1595, which fixes the leading edge).

## Depends on

- #1595 defines the "active battle + elapsed offset" client delivery contract this reuses.

## Tests

Unit tests for the remainder math (remainder < cooldown, remainder spanning cooldown + partial battle, remainder of zero) and integration coverage of `OfflineProgressService.SimulateProgress` handing the offset forward.

Issue #1595, opened 2026-07-04
Offline catch-up: hand back a still-in-progress stale battle instead of booking a draw


ginderjeremiah:
Part of spike #1073 — see `docs/spikes/1073-idle-time-recovery.md`.

## Problem

`BattleService.AbandonBattle` (`Game.Application/Services/BattleService.cs:338-390`; `ResolveStaleBattle` delegates to it) replays a stale in-flight battle from its seed, capped at `DefaultMaxBattleMs` (2 min). When the replay does not conclude, it is booked as a draw (no rewards) and the battle is cleared. But a non-concluding replay whose real elapsed time (`now - BattleStartTime`) is still **under** the 2-minute cap is not a draw — it is a battle still in progress, and discarding it loses the player's progress and restarts the fight from scratch.

## Fix

In the non-concluding branch: when `elapsed < DefaultMaxBattleMs`, keep the battle active — same enemy/seed/snapshot, `BattleStartTime` unchanged — and surface it to the client as in-progress with its elapsed offset instead of clearing it. Only a replay that reaches the cap with both battlers alive is a genuine draw.

Anti-cheat needs no new surface: every elapsed-time check already derives from `now - state.BattleStartTime` (the victory tolerance check at `BattleService.cs:268-286`, the cooldown gate), and the server itself owns `BattleStartTime`.

## Delivery seam (shared contract — settle it here)

The client must receive "active battle + elapsed offset". Natural options: `NewEnemy` returns the still-active battle (with offset) rather than abandoning it, and/or `GetOfflineProgress` carries it for the welcome-back gate. Define the payload contract in this issue; the client replay-to-offset issue consumes it.

## Tests

Integration tests around `ResolveStaleBattle`/`AbandonBattle` covering the three partitions: replay concludes (existing behavior), still-in-progress (new hand-back branch), true draw at the cap.

PR #1599, opened 2026-07-04
Spike #1073: settled direction — worker tick source + server-authoritative resume

### LINKED ISSUE #879: Add offline time tracking/rewards

### LINKED ISSUE #1073: Spike: recover idle time lost to background-tab throttling (mid-session catch-up)

ginderjeremiah:
## Summary

Final write-up for spike #1073 with the direction settled with the owner (2026-07-04). Docs-only: one new file, `docs/spikes/1073-idle-time-recovery.md`.

Supersedes #1584 (closed unmerged). Relative to that write-up, the owner review changed the client-side mechanism: the primary fix for the throttled-but-connected case is now a **Web Worker tick source** for `LogicalEngine` (worker timers and message delivery aren't throttled on hidden pages), replacing the clamp-driven `onIdleTimeLost` fast-forward, which is explicitly dropped as an accepted small loss (freeze-thaw with a surviving socket funnels into the reconnect path). The server-side design carried over unchanged: backdate `PlayerState.BattleStartTime` to hand back still-in-progress stale battles and to carry the offline loop's trailing remainder, which makes "already in progress" fall out of the existing anti-cheat model for free.

The flagged architectural call from #1584 — a third `battleStep` consumer — is now settled as accepted: it's unavoidable in any design that resumes a battle at a non-zero offset (the server-resume path needs it regardless of which client mechanism wins), and it reuses the parity-tested stepper. `docs/frontend.md`'s battle-parity section gets updated when that issue lands, not in this PR.

## Implementation issues filed (linked to #1073)

- #1594 — worker tick source for `LogicalEngine` (small)
- #1595 — hand back a still-in-progress stale battle instead of booking a draw; defines the battle+offset payload contract (medium)
- #1596 — carry the offline loop's trailing remainder as a backdated next battle; closes #879's accepted boundary-slice loss (medium)
- #1597 — client replay-to-offset, the third `battleStep` consumer (medium)
- #1598 — hidden-tab presentational side-effect hygiene, follow-up to #1594 (small)

## Test plan

- N/A — documentation only.

Closes #1073.

---

### DIFF
--- a/docs/spikes/1073-idle-time-recovery.md
+++ b/docs/spikes/1073-idle-time-recovery.md
@@ -1,0 +1,74 @@
+# Spike #1073 — Recover idle time lost to background-tab throttling
+
+- **Spike issue:** [#1073](https://github.com/ginderjeremiah/GameServer/issues/1073)
+- **Status:** Direction settled with the project owner (2026-07-04); implementation issues filed (see table below). Supersedes the earlier write-up in PR [#1584](https://github.com/ginderjeremiah/GameServer/pull/1584) (closed unmerged) — the owner review replaced its client-side mechanism (clamp-driven fast-forward) with a Web Worker tick source; the server-side design carried over unchanged.
+- **Related:** surfaced by [#922](./922-multiple-players-per-account.md) (Decision 6); its sibling mitigation [#1074](https://github.com/ginderjeremiah/GameServer/issues/1074) (allowlist nudge) already shipped; reuses the [#879](./879-offline-rewards.md) offline simulator (#1042).
+
+## The problem, and what's changed since the issue was filed
+
+The idle loop is client-driven: `LogicalEngine.logicLoop` (`UI/src/lib/engine/logical-engine.ts:53-58`) polls every 10ms via `window.setInterval` and computes `timeDelta` from `performance.now()`. `update()` (`:60-75`) clamps any single poll to at most `tickSizeX5` (200ms) of *simulated* battle progress — the excess is added to the engine's wall-clock but never fed into a battle tick. Browsers throttle hidden-tab page timers (≥1s, ~once/minute after sustained hiding), so a backgrounded character's simulated battle drifts behind real time indefinitely.
+
+Two claims in the original issue needed correcting during research:
+
+- **"No `visibilitychange` handling today" is stale.** #1074 shipped since: `BackgroundThrottleMonitor` (`UI/src/lib/engine/background-throttle-notice.ts`) listens for `visibilitychange` and accumulates lost time via the `onIdleTimeLost` hook (emitted by the clamp, `logical-engine.ts:65`) — but purely to show a one-time allowlist-nudge toast. It never touches the engine or the server.
+- **#1042 (offline rewards) already implements most of the issue's proposed mechanism.** `OfflineProgressService.SimulateProgress` (`Game.Application/Services/OfflineProgressService.cs:100-160`) resolves the stale in-flight battle first (replay from `BattleSeed`/`Snapshot`, capped at `DefaultMaxBattleMs` = 2 min) and then runs the whole-battle catch-up loop. The two genuine gaps are at its edges: a non-concluding replay is booked as a draw and discarded rather than handed back as "still in progress" (`BattleService.AbandonBattle`, `Game.Application/Services/BattleService.cs:338-390`), and the whole-battle loop (`Game.Core/Battle/Offline/OfflineProgressSimulator.cs:28-85`) drops its trailing remainder instead of carrying it into the next battle's starting offset.
+
+## Decision 1 — a Web Worker tick source is the primary client-side fix (#1594)
+
+Replace `LogicalEngine`'s `window.setInterval` with a dedicated module worker that posts a message every polling interval; the main thread drives `logicLoop` from `onmessage`. Dedicated-worker timers are not throttled in Chromium/Firefox, and `message` delivery to a hidden page is not throttled either — so the engine keeps ticking at full rate while hidden and the clamp stops firing in the common case. This changes the common case from "catch up on return" to "never fall behind": hidden-tab completions, challenge pushes, and proficiency XP report through the existing flow in real time, and the hidden-but-connected dead zone (socket alive, so `LastActivity` stays fresh and #1042 never triggers) closes entirely. It is the programmatic equivalent of the browser-settings allowlisting the #1074 toast asks users for.
+
+What a worker **cannot** cover — and why the server-side work below is needed regardless:
+
+- **Page freeze / tab discard** (Chrome Memory Saver, battery saver) stops the whole page *including its workers* — and Memory Saver is exactly the audience #1074's notice targets.
+- **Safari** suspends hidden tabs wholesale (fully on iOS), workers included.
+- **Sleep, browser close, hard refresh, socket reconnect** — client state is simply gone.
+
+**Rejected: moving the battle engine itself into the worker.** `BattleEngine` reads statified Svelte `$state` managers (`inventoryManager`, `playerManager`, `playerProficiencies`, `staticData`) and emits `logMessage`/`notifyCombatFloat` per tick; the `Battler` objects are rendered directly by components. Off-threading that means serializing the build in, streaming events out, and a thread hop between the sim and the socket — a large refactor that buys nothing, because the worker dies with the page in every case above anyway.
+
+Two consequences, handled as their own issues:
+
+- **`BackgroundThrottleMonitor` self-gates and needs no change**: where the worker trick works, `onIdleTimeLost` stops firing while hidden, so the toast naturally only appears where it still helps. The clamp and `onIdleTimeLost` stay as the safety valve and instrumentation.
+- **Presentational side effects now run while hidden** (#1598): floater-removal `setTimeout`s are throttled to ~1/min batches while battles keep resolving; floats should simply not spawn while `document.hidden` (they are `aria-hidden` presentational). The combat log is already capped.
+
+## Decision 2 — server-authoritative resume via backdated `BattleStartTime` (#1595, #1596)
+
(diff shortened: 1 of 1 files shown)


docs/spikes/1073-idle-time-recovery.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Decision 2 — server-authoritative resume via backdated `BattleStartTime` (#1595, #1596)
+
+When client state is gone (freeze/discard, reconnect, refresh), the fix extends #1042's two lossy edges rather than adding a parallel mechanism:
+
+- **Hand back a still-in-progress stale battle (#1595).** In `AbandonBattle`'s non-concluding branch, when `now - BattleStartTime` is still under the 2-minute cap, keep the battle active (same enemy/seed/snapshot, `BattleStartTime` unchanged) and surface it to the client with its elapsed offset instead of booking a draw. Only a replay that reaches the cap with both battlers alive is a genuine draw.
+- **Carry the offline loop's trailing remainder (#1596).** After the whole-battle crediting loop, set up the next battle backdated by the remainder via the existing `PlayerState.SetActiveBattle(..., startTime:)` overload (`Game.Core/Entities/PlayerState.cs:55`), serving the post-battle cooldown out of the remainder first. This closes the boundary-slice loss #879 accepted as a known limitation.
```
ginderjeremiah, line 38: Nit (path): `SetActiveBattle` with the `startTime:` parameter is correct at **line 55**, but the file lives at `Game.Core/Players/PlayerState.cs`, not `Game.Core/Entities/PlayerState.cs`. Worth fixing so the implementer of #1596 lands on the right file. (Minor sibling nit on line 14: `AbandonBattle` actually starts at `BattleService.cs:344`, not `338` — the `338-390` range is close enough that it's optional.)

---

ginderjeremiah (commented):
## Review: Spike #1073 settlement — worker tick source + server-authoritative resume (docs-only)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the write-up is technically accurate, the direction is sound, and the implementation issues are filed and scoped. **Not merging**, per policy that spike results are not merged; this stays open for the owner to close alongside #1073.

I spot-checked the load-bearing code references against the current tree and they hold up:

- `logical-engine.ts` — `logicLoop` `:53-58`, `update` `:60-75`, the `tickSizeX5` clamp emitting `onIdleTimeLost` at `:65`, 10ms poll. ✅ Matches, and the "excess advances the wall-clock but is never fed into a tick" description is exactly what the clamp does.
- `OfflineProgressSimulator.Simulate` `:28` — the loop does `remainingMs -= result.TotalMs + CooldownMs` and returns on the next guard, so the trailing partial battle genuinely is dropped. ✅ The gap the write-up names is real.
- **The central architectural claim is verified.** `BattleService.cs:268-286` computes `battleCompletedAt = state.BattleStartTime.AddMilliseconds(result.TotalMs)` and rejects a victory only when `now < battleCompletedAt - tolerance` — server-clock only, rejecting *early* completions. So backdating `BattleStartTime` really does make "already in progress" fall out of the existing anti-cheat model for free, and the freeze-thaw *late*-completion case is accepted rather than rejected — both consequences the write-up leans on. ✅
- `PlayerState.SetActiveBattle(..., startTime:)` exists (line 55), and both docs-to-update (`docs/frontend.md`, `docs/backend-battle.md`) exist. ✅

### What's good

- **Diligence in correcting the original issue.** Flagging that "no `visibilitychange` handling today" is stale (#1074 shipped `BackgroundThrottleMonitor`) and that #1042 already implements most of the proposed mechanism — reducing the remaining work to two well-defined edges — is exactly the kind of verification CLAUDE.md asks for (don't trust the issue's own claims). This materially shrank the scope.
- **The reframe from "catch up on return" to "never fall behind"** via the worker tick source is the right call, and the honest enumeration of what a worker *cannot* cover (page freeze/discard, Safari, sleep/close/refresh/reconnect) is what justifies keeping the server-authoritative path — no hand-waving.
- **Rejected alternatives are documented with reasons** (moving `BattleEngine` into the worker; client-measured-gap fast-forward), and the third-`battleStep`-consumer architectural call is surfaced explicitly with owner sign-off rather than buried.
- Issues #1594–#1598 are filed, scoped (small/medium), linked to #1073, and carry dependency ordering (#1596 depends on #1595's payload contract; #1597 consumes it). The doc format matches the other `docs/spikes/*` write-ups.

### Nits (non-blocking)

- `Game.Core/Entities/PlayerState.cs:55` → should be `Game.Core/Players/PlayerState.cs:55` (line 38). Left inline.
- `AbandonBattle` cited as `BattleService.cs:338-390`; the method actually starts at `:344`. Trivial, optional (line 14).

### Test coverage

N/A — documentation only, no code paths introduced. Appropriate for a spike write-up.

Nice work — this is a well-reasoned settlement that leaves five cleanly-scoped issues in its wake. Fix the path nit at your convenience; it doesn't gate the spike.

---

PR #1607, opened 2026-07-04
Migrate the enemy dead-stat lint onto CombatRating.Marginal (#1581)

### LINKED ISSUE #1581: Migrate the enemy dead-stat lint (#1529) onto CombatRating.Marginal for exact detection
### Context

#1529 (PR #1580) added an `EnemyInertAttribute` progression-graph-lint warning that flags a live enemy's core-attribute distribution point nothing in its kit consumes. It uses a v1 heuristic map (static-derivation check + pooled `DamageMultiplier`/effect-scaling check), per the issue's own scope note.

The combat-rating spike (#1526) shipped exactly the tool that would make this exact: `CombatRating.Marginal` (`Game.Core/Battle/CombatRating.cs`, from #1531), whose doc comment says it "doubles as the dead-stat detector the enemy content lint (#1529) needs: an attribute whose fielded kit has no matching enabler ... marginals to exactly 0."

### Why not done in #1580

`CombatRating.Marginal` operates on a `Battler` (an `AttributeCollection` + fielded `BattleSkills`, at a specific level — attribute distributions are level-scaled), but `ProgressionGraphChecker`/`ContentGraph` work entirely off the lightweight `Contracts` read-DTOs and never assemble a `Battler`. Bringing that construction into the lint (deciding what level to rate at, wiring `Game.Core.Battle` into `Game.Application.Content`) is a real architecture step, not a small follow-up to a "scope: small" issue.

### Ask

Once the calibration work (#1533, PR #1579) lands and there's an established pattern for building a representative `Battler` from `Contracts.Enemy` at a chosen level outside of live battle, revisit `CheckEnemyAttributeConsumption` in `ProgressionGraphChecker.cs`: replace the static-derivation/DamageMultiplier/effect-scaling heuristic with an actual `CombatRating.Marginal(battler, isPlayer: false, attribute)` check per core attribute, flagging when the marginal is ~0. This would also correctly resolve the Agility/CooldownBonus edge case noted in #1580 (Agility's `CooldownBonusMultiplier` derivation is only alive if the enemy also authors `CooldownBonus` — the marginal captures that automatically, where the current heuristic has to special-case it).

### Notes

- Not urgent — the current heuristic already matches all authored content and the issue's intended v1 scope.
- Worth cross-referencing the calibration report (#1533/#1579) since it uses similar Battler-construction infrastructure and could share a helper.

ginderjeremiah:
## Summary

Implements #1581: replaces the v1 heuristic in `ProgressionGraphChecker.CheckEnemyAttributeConsumption` (static-derivation check + pooled `DamageMultiplier`/effect-scaling check, from #1580) with an exact per-attribute check via spike #1526's `CombatRating.Marginal` (`Game.Core/Battle/CombatRating.cs`), now that the calibration report (#1533, merged as PR #1579) established the pattern for assembling a representative `Battler` from reference data at a chosen level outside of live battle.

- **Level resolution**: `Contracts.Enemy` carries no level of its own, so `ResolveRatingLevel` rates a live enemy at the arc midpoint of the first live zone it spawns in, or that zone's fixed `BossLevel` if it's the dedicated boss there — the same placement precedent `CombatRatingCalibrationIntegrationTests.BuildPricedEnemyList` uses. An enemy with no live placement has no representative encounter to rate and is skipped (mirroring the calibration report's own omission of unplaced enemies), rather than being checked at a made-up level.
- **Battler assembly**: `BuildRatingBattler`/`ToRatingSkill` build a `Battler` directly from `Contracts.Enemy`/`Contracts.Skill` (every live skill in the enemy's `SkillPool` — the check asks whether anything in the *kit* consumes the attribute, not a random per-encounter draw, matching the prior heuristic's kit). This mirrors `Enemy.ToBattler()`, but can't reuse it or `Game.DataAccess.Mapping.SkillMapper`: `Game.Application` has no reach into the data tier's entity↔domain mapping, and `Game.Core` itself has no dependency on `Game.Abstractions.Contracts` to host a shared mapping instead. So this is a small, local Contracts→Core mapping inside `ProgressionGraphChecker`, scoped to what this check needs.
- **Dead-stat threshold**: attributes whose finite-difference `Marginal` is under `1e-6` are flagged as inert — matching the tolerance `CombatRatingTests` already uses to assert an exactly-dead marginal (`Assert.Equal(0, marginal, 6)`).
- The old `IsConsumedByKit`/`HasStaticDerivation` heuristic helpers are removed — no remaining callers.

## Judgment calls (flagged per this project's convention)

1. **Where the Contracts→Core.Skills.Skill mapping lives.** Since `Game.Core` has zero project references and `Game.Application` can't reach `Game.DataAccess`, this mapping has no natural shared home — I kept it as a private, `ProgressionGraphChecker`-local helper rather than introducing a new shared abstraction for a single caller. Worth revisiting if a second consumer of "Battler from Contracts" ever appears outside the data tier.
2. **Test fixture fix required for real math.** `Enemy_DistributesLuckConsumedByPooledEffectScaling_ProducesNoFinding`'s skill effect never set `Target`/`DurationMs`, which the old structural heuristic didn't care about but which made the effect invisible to `CombatRating` (no target ⇒ folded into neither the effective-caster nor reference-defense pass; zero duration ⇒ zero uptime). Fixed the fixture to set `Target = Self` and a nonzero `DurationMs` so it's a genuine self-buff, per the project's convention of only changing test fixtures when the change reflects reality — the old assertion (not flagged) still holds, now for the right reason.
3. Added one new test (`Enemy_NotPlacedInAnyLiveZone_ProducesNoFinding`) covering the new "no live placement → skip" branch, since it's new domain logic without a prior heuristic equivalent.

## Test plan

- [x] `dotnet test --no-build --no-restore` — full suite passes (2916 tests: Core 1317, Infrastructure 59, Application 851, Api 689), including the updated `ProgressionGraphCheckerTests` (all prior `EnemyInertAttribute` cases still pass under the real rating math) and `ContentGraphLintTests` (the committed `content/*.json` stays lint-clean under the exact check)

Closes #1581

---
_Generated by [Claude Code](https://claude.ai/code/session_01S4jY9p8PH1sVo7PpWJAMkV)_

### DIFF
--- a/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
+++ b/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
@@ -291,6 +291,24 @@ namespace Game.Application.Tests.Content
         }
 
         [Fact]
+        public void Enemy_DistributesAgilityWithCoAuthoredCooldownBonus_ProducesNoFinding()
+        {
+            // Agility has no direct kit consumer here, but co-authoring CooldownBonus makes its derived
+            // CooldownBonusMultiplier live (CooldownBonus × CooldownBonusMultiplier feeds the cooldown rate),
+            // so Agility is no longer dead weight — the exact case #1581 moved off the heuristic for.
+            var graph = HealthyGraph() with
+            {
+                Enemies =
+                [
+                    Enemy(0, skillPool: [2], spawns: [(0, 1), (1, 1)],
+                        attributeDistribution: [(EAttribute.Agility, 5, 1), (EAttribute.CooldownBonus, 5, 1)]),
+                    Enemy(1, isBoss: true),
+                ],
+            };
+            Assert.DoesNotContain(_checker.Check(graph), f => f.Check == "EnemyInertAttribute");
+        }
+
+        [Fact]
         public void Enemy_DistributesLuckConsumedByPooledEffectScaling_ProducesNoFinding()
         {
             var graph = HealthyGraph() with
--- a/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
+++ b/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
@@ -360,6 +360,20 @@ namespace Game.Application.Tests.Content
             Assert.DoesNotContain(_checker.Check(graph), f => f.Check == "EnemyInertAttribute");
         }
 
+        [Fact]
+        public void Enemy_NotPlacedInAnyLiveZone_ProducesNoFinding()
+        {
+            // No spawn table entry and no dedicated-boss slot means no representative level to rate at, so the
+            // check is skipped rather than rating at a made-up level.
+            var graph = HealthyGraph() with
+            {
+                Enemies = HealthyGraph().Enemies
+                    .Append(Enemy(9, skillPool: [2], spawns: [], attributeDistribution: [(EAttribute.Luck, 5, 1)]))
+                    .ToList(),
+            };
+            Assert.DoesNotContain(_checker.Check(graph), f => f.Check == "EnemyInertAttribute");
+        }
+
         // --- Classes ----------------------------------------------------------------------------------
 
         [Fact]
@@ -927,7 +941,14 @@ namespace Game.Application.Tests.Content
                 .Select(m => new Contracts.AttributeMultiplier { AttributeId = m.attributeId, Multiplier = m.multiplier })
                 .ToList(),
                 Effects = (effectScaling ?? [])
-                .Select(e => new Contracts.SkillEffect { AttributeId = EAttribute.MaxHealth, ScalingAttributeId = e.attributeId, ScalingAmount = e.scalingAmount })
+                .Select(e => new Contracts.SkillEffect
+                {
+                    Target = ESkillEffectTarget.Self,
+                    AttributeId = EAttribute.MaxHealth,
+                    DurationMs = 1000,
+                    ScalingAttributeId = e.attributeId,
+                    ScalingAmount = e.scalingAmount,
+                })
                 .ToList(),
                 DamagePortions = [new Contracts.SkillDamagePortion { Type = primaryType, Weight = 1 }],
                 Description = "",
--- a/Game.Application/Content/ProgressionGraphChecker.cs
+++ b/Game.Application/Content/ProgressionGraphChecker.cs
@@ -1,6 +1,7 @@
 using Game.Core;
 using Game.Core.Attributes;
 using Game.Core.Attributes.Modifiers;
+using Game.Core.Battle;
 using Game.Core.Progress;
 using Game.Core.Skills;
 using Contracts = Game.Abstractions.Contracts;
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Migrate the enemy dead-stat lint onto `CombatRating.Marginal` (#1581)

**Intended verdict: Approve** — this is a clean, well-scoped migration that does exactly what #1581 asked, with strong test coverage and no blocking findings. (Submitted as a `COMMENT` because GitHub won't accept a gating verdict on one's own PR; treat this as an approve modulo the should-fix below.)

The core swap is sound: replacing the `IsConsumedByKit`/`HasStaticDerivation` heuristic with `CombatRating.Marginal(battler, isPlayer: false, attr) ≈ 0` is the exact detector the spike promised, and it correctly subsumes the special cases the old heuristic had to hand-code — Luck/crit being asymmetry-gated (skipped for enemies) and the Agility→`CooldownBonusMultiplier` derivation only mattering when the enemy co-authors `CooldownBonus` all fall out of the marginal automatically. Rating at a single representative level is defensible: dead/alive is a structural property of the kit, level-invariant for these smooth derivations, so one placement is sufficient. The `ResolveRatingLevel` skip for unplaced enemies is the right call over rating at a made-up level, and it mirrors the calibration report's own omission.

### Should-fix

- **Test coverage — the motivating edge case isn't pinned.** #1581 explicitly cites the Agility/`CooldownBonus` case as the reason to move off the heuristic ("the marginal captures that automatically, where the current heuristic has to special-case it"), but no test locks it in: there's no case where an enemy distributes **both** `Agility` and `CooldownBonus` and asserts Agility is *not* flagged (the `CooldownBonusMultiplier` derivation making it live). That's precisely the behavior the migration improves, and without a test a future refactor of the rating could silently regress it back to a false positive. The general mechanism is otherwise very well covered (Strength/Endurance derivation, Agility/Luck/Intellect no-consumer, consumed-by-DamageMultiplier, consumed-by-effect-scaling, zero-scaling/zero-multiplier, retired skill/enemy, zero-amount, and the new unplaced-enemy skip) — this is the one gap. Worth adding to the PR rather than deferring.

### Notes (non-blocking)

- **Local `Contracts`→`Core` mapping (`BuildRatingBattler`/`ToRatingSkill`).** Duplicating what `SkillMapper` does is a real smell, but the judgment call is correct and well-documented: `Game.Core` has no reference to `Game.Abstractions.Contracts` and `Game.Application` can't reach `Game.DataAccess.Mapping`, so there's no shared home for a single caller. Fine as-is; revisit if a second "Battler from Contracts" consumer appears (a follow-up, not this PR).
- **Fixture fix in the shared `Skill(...)` helper** (adding `Target = Self` + `DurationMs`) is legitimate — a scaling effect with no target/duration is genuinely invisible to `CombatRating`, so the old assertion was passing for the wrong reason. Only two `effectScaling` consumers exist and both remain correct (the Luck-consumed case now registers; the zero-scaling case stays inert).

### CI

Backend-only change, so the frontend jobs are correctly skipped; `verify-backend` and all three e2e browsers are green. `test-backend` was still in progress at review time — worth confirming it lands green before merge.

Not merging on this review: the should-fix (add the Agility/`CooldownBonus` test) should be handled in-PR, and backend CI wasn't finished. Once that test is in and `test-backend` is green, this is good to merge.

---

ginderjeremiah:
## Review: Migrate the enemy dead-stat lint onto `CombatRating.Marginal` (#1581)

**Verdict: Approve with nits** — the migration is correct, the claims in the description hold up against the code, and CI is green. One test-coverage gap is worth a quick follow-up but doesn't block merge.

_(Posting as a comment rather than a gating review since I'm also the PR author.)_

### Claims verified against the code
- **Domain-isolation forcing a local mapping** — confirmed. `Game.Application.csproj` references only `Game.Core` + `Game.Abstractions`; `Game.Core.csproj` has zero project references. So neither layer can host a shared `Contracts→Core` mapping, and `Game.DataAccess.Mapping.SkillMapper` is genuinely unreachable from here. The private `ProgressionGraphChecker`-local `BuildRatingBattler`/`ToRatingSkill` is the right call for a single consumer, and the judgment-call note is accurate.
- **Signatures/mappings** — `CombatRating.Marginal(Battler, isPlayer, attribute)` and `new Battler(AttributeCollection, skills, level)` match. `ToRatingSkill` sets every `required` prop on `Game.Core` `Skill`/`SkillEffect`; decimal→double casts match `AttributeDistribution.GetDistributionModifier`'s double-first arithmetic.
- **Fixture fix (judgment call #2)** — genuinely required, not cosmetic. `CombatRating.BuildEffectiveCaster` only folds effects with `Target == Self`, and `FoldedEffectMagnitude` gives zero uptime when `DurationMs == 0`. `ESkillEffectTarget.Self = 1`, so the old default-`0` target left the effect invisible to both the effective-caster and reference-defense passes. Setting `Target = Self` + `DurationMs = 1000` makes it a real self-buff — the assertion now holds for the right reason.

### What's good
- The Agility + `CooldownBonus` edge case #1581 explicitly called out is now handled automatically by the marginal (no special-casing) and pinned by a new test.
- Dead heuristic helpers (`IsConsumedByKit`/`HasStaticDerivation`) removed cleanly, no orphan callers.
- Positive (finding) and negative (no-finding) cases all carry real spawn placement, so they exercise the marginal-is-~0 path rather than passing by accident.
- The `1e-6` threshold is tied to the tolerance `CombatRatingTests` already uses, and the doc comments explain the *why* well.

### Should-fix (follow-up, non-blocking)
- **Boss-level branch of `ResolveRatingLevel` is uncovered.** In `HealthyGraph` no zone sets `BossEnemyId`, and `Enemy(1, isBoss: true)` is never referenced as a zone's boss — so `zone.BossEnemyId == enemy.Id ⇒ return zone.BossLevel` is never exercised by any test. This is meaningful for real content (a dedicated boss placed only via `BossEnemyId`, not a spawn table, is rated *only* through this branch). Add a test with a boss-placed enemy carrying a core distribution (one live, one dead) to pin it. Not blocking: the spawn and no-placement branches are well covered and the committed-content lint passes in CI.

### Nits (optional)
- `CombatRating.Marginal` recomputes its `Rate(battler)` baseline on every call, so per enemy the baseline is recomputed up to 6× (once per core attribute). Negligible for an offline lint; only worth touching if a batched-marginal API ever lands.
- `ResolveRatingLevel` precedence: within a zone, a spawn entry wins over the boss slot, so an enemy that is both a zone's boss *and* in its spawn table rates at the arc midpoint, not `BossLevel`. Almost certainly intended (it does spawn there) — a one-line note would guard against a later "simplification."

### CI
All checks green — `test-backend`, `verify-backend`, `test-e2e` (chromium/firefox/webkit), `all-checks-passed` all succeeded; frontend jobs correctly skipped (backend-only change).

— Review by Claude

---

Issue #1591, opened 2026-07-04
Lesson reference data: entity, Workbench authoring, graph lint coverage


ginderjeremiah:
Part of #1392 (revised direction — see `docs/spikes/1392-ui-tutorials.md`). Tutorial lessons are **Workbench-authored reference data**, not hardcoded frontend strings, so the progression-graph lint can verify the content arc's taught-by-blurb lane and copy iterates through the established DB-primary export/seed pipeline (#1390).

## Scope

- **`Lesson` entity (+ ordered `LessonStep`s)**: key/name, trigger (type: screen-visit or mechanic-event, plus target — a screen key or mechanic-event enum member), the screen its tour plays on, per-step text and optional `anchorKey` (a string the frontend maps to a UI anchor target), and display ordering for the Help screen.
- **New intrinsic mechanic-event enum** (initial members: first crit landed, first dodge, first cooldown recharge — matching the detectors in `UI/src/lib/common/content-events.ts`). Note: intrinsic enums are `HasData`-seeded by `GameContext`, so this needs an EF migration (startup project `Game.Infrastructure`) or all integration tests fail on `PendingModelChangesWarning`.
- **Reference-data cache + seed/export**: ride the existing in-memory reference cache and the #1390 export/seed pipeline like any other content entity.
- **Workbench CRUD** for lessons and their steps, following the existing admin add/edit/update patterns.
- **Graph lint rules**: every taught-by-blurb candidate named in `docs/content-design.md` §2 has a live lesson; every lesson trigger references an existing screen key / enum member; no lesson is orphaned from a trigger.

## Out of scope

- Per-player lesson state (#1588) and client trigger evaluation (#1587).
- Frontend validation that `anchorKey`s resolve to real UI anchors — that's a UI-side test owned by the tour component issue.

## Tests

Standard reference-data entity coverage plus unit tests for each lint rule (present/missing/retired-reference cases).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1589, opened 2026-07-04
Build the Help screen (tutorial reading room)


ginderjeremiah:
Part of #1392. **Updated 2026-07-04** to match the revised direction (see `docs/spikes/1392-ui-tutorials.md`): the Help screen is the tutorial **reading room**, not just a replay list of dismissed toasts.

## Scope

- Build the reserved `help` screen (`screen-defs.ts`, `key: 'help', built: false` today) listing every lesson with its per-player state (locked / unread / read, from #1588).
- Unread lessons emphasized (this is where the unread indicator from #1587 leads); selecting a lesson navigates to its host screen and plays its tour (#1592) — same open flow as #1587, marking it read.
- Read lessons are replayable the same way. Locked lessons display greyed with name only (a light teaser, no step content) — revisit if this spoils content-arc surprises.

## Depends on

#1587, #1588, #1592.

## Tests

Component tests: list rendering per state, unread emphasis, replay flow invokes the shared open flow.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1588, opened 2026-07-04
Lesson state persistence (server-side PlayerLessons)


ginderjeremiah:
Part of #1392. **Rescoped 2026-07-04** — the direction was revised after review (see `docs/spikes/1392-ui-tutorials.md`); this issue previously specified `localStorage` one-shot dismissal, which is superseded: lesson state is account-scoped, so it lives server-side like other player state (device-scoped `localStorage` re-prompts everything on every new device).

## Scope

- **`PlayerLesson` state**: one row per player+lesson — absent row = locked, `unlockedAt` (mechanic trigger fired), `readAt` (tour completed). EF entity + migration.
- **Socket commands**: unlock-lesson (client-detected trigger — trusted, since nothing is rewarded and a dishonest client can only show itself tutorials early) and mark-lesson-read. Include lesson states in the player load payload.
- **Client store** exposing lesson states to the trigger/flow layer (#1587), replacing any notion of `localStorage` flags.

## Depends on

#1591 (Lesson entity).

## Tests

Integration tests for both commands and load-payload inclusion; unit tests for state transitions (idempotent unlock, read-before-unlock rejected or normalized).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1587, opened 2026-07-04
Tutorial trigger evaluation + unread lesson flow (client)


ginderjeremiah:
Part of #1392. **Rescoped 2026-07-04** — the direction was revised after review (see `docs/spikes/1392-ui-tutorials.md`); this issue previously described a hardcoded lesson registry delivering via toasts, which is superseded. Lessons are now reference data (#1591) with a per-player `locked → unread → read` lifecycle (#1588), presented as anchored tours (#1592).

## Scope

- **Screen-anchored triggers**: hook the screen-activation point in `routes/game/+page.svelte` (screens are registry state, not routes — no `afterNavigate`). First activation of a screen with an associated locked lesson plays its tour immediately (the player just navigated there), then marks it read.
- **Mechanic-anchored triggers**: wire the content-events detectors (#1586, `UI/src/lib/common/content-events.ts`) into the engine loop; when one fires for a locked lesson, report the unlock (#1588 command) so the lesson becomes **unread**. No display at trigger time — the player may be AFK.
- **Unread indicator**: persistent badge on the Help nav entry / HUD affordance, visually distinct from the toast channel (toasts are not used for lessons). Opening an unread lesson navigates to its host screen via `requestScreen` and plays its tour (#1592), marking it read on completion.
- Queue-always for mechanic lessons is the deliberate v1 behavior; play-immediately-if-on-screen is a deferred refinement.

## Depends on

#1591 (lesson data), #1588 (state persistence), #1592 (tour component).

## Tests

Unit tests for trigger evaluation (first-activation semantics, locked/unread/read transitions, no re-fire on already-read) and the open→navigate→play→mark-read flow.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1609, opened 2026-07-04
Anchored tutorial tour component (coach-marks) (#1592)

### LINKED ISSUE #1592: Anchored tutorial tour component (coach-marks)
Part of #1392 (revised direction — see `docs/spikes/1392-ui-tutorials.md`). Tutorial lessons present as an **anchored popover tour** (coach-marks pointing at real controls), not toasts — toasts are the reward/unlock noise channel and are missable while idle.

## Scope

- **Tour player component** in the existing overlay family: highlights a target element and positions a callout beside it with the step's text, next/back/done controls, and a step counter. Reuse the `focus-trap` chrome (`Escape`, focus capture/restore, scroll lock) that `Popover.svelte`/`ModalHost` share. Note `Popover.svelte` itself is a container overlay (positions over its nearest positioned ancestor, no target anchoring) — the tour needs new anchored positioning, it is not a straight reuse.
- **Anchor registration**: a Svelte action (e.g. `use:tutorialAnchor={'skill-bar'}`) registering DOM targets by key, so lesson steps reference stable `anchorKey` strings rather than selectors. A step whose anchor is missing/unregistered degrades to a centered, unanchored callout rather than failing.
- Driven by the `Lesson`/`LessonStep` reference-data shape from #1591 (the component can be built against the typed shape before that lands).

## Tests

- Component tests: step navigation, anchored vs. degraded positioning, dismissal restores focus.
- A test asserting every `anchorKey` referenced by seeded lessons resolves to a registered anchor — the frontend half of the lint coverage in #1591 (the backend lint can't see the DOM).

---

ginderjeremiah:
## Summary

Implements #1592 (spike #1392, revised direction): the anchored popover "coach-marks" tour player, built ahead of #1591 (the `Lesson`/`LessonStep` reference-data entity) landing — the issue explicitly calls this out as buildable against a typed shape before that entity exists.

- **`TourPlayer.svelte`** (`UI/src/components/tour/`): highlights a registered anchor target with a spotlight cutout (an oversized `box-shadow` trick — no `clip-path` mask needed) and positions a callout beside it with the step's text, a 1-based step counter, and Back/Next/Done controls. Edge-aware positioning (flips above the anchor if there's no room below, clamps to the viewport) recomputes on window resize. Degrades to a centered, unanchored callout when the current step has no `anchorKey` or the key isn't (yet) registered.
- **`tutorial-anchor.ts`**: the `tutorialAnchor` Svelte action (`use:tutorialAnchor={'skill-bar'}`) registers a DOM element under a stable string key in a reactive `SvelteMap`; `getTutorialAnchor(key)` resolves it. The destroy guard only clears a key's slot if the destroying element still owns it — the same overlapping-mount/unmount hazard the existing tooltip registry's id-keyed map documents, here guarded by identity instead since only one element should claim a given anchor key at a time.
- **`tour-types.ts`**: a local `TourStep` (`{ text, anchorKey? }`) shape mirroring the future `LessonStep`, so lesson steps map onto it directly (`ordinal` is just array position) once #1591 lands.
- Reuses the shared `focusTrap` action (Escape, Tab-trap, scroll-lock, focus capture/restore) from the `Popover`/`ModalHost` overlay family, per `docs/frontend-overlays.md`.

## Scope notes / judgment calls

- **Not wired up anywhere yet.** This issue is scoped to the component + anchor action only; the trigger-evaluation/open-flow layer that will actually invoke it with real lesson content is #1587 (blocked on #1591, which is still open as of this PR), and the Help screen consumer is #1589. `TourPlayer` is a fully controlled component (`open`/`steps`/`onDismiss`/`onComplete`) so either can drive it once they land.
- **`onDismiss` vs `onComplete` are separate callbacks**, mirroring `Modal`'s `onConfirm`/`onCancel` split rather than a single dismissal callback — early exit (Escape/backdrop/Skip) and finishing the last step via Done are different signals for the lesson's read state (#1588), and that distinction belongs to the caller, not this component.
- **The anchor-resolution test is fixture-based, not seeded-lesson-based**, since `content/lessons.json` doesn't exist yet (#1591 unmerged). `TourPlayer.test.ts` registers a plain `tutorialAnchor` fixture and asserts the anchored/centered branches; the real "every anchorKey referenced by seeded lessons resolves to a registered anchor" lint test the issue asks for needs actual lesson content and is left for whichever of #1587/#1589/#1605 first has real seeded steps to test against.
- No global store/host (à la `modal.svelte.ts`/`ModalHost`) is added — how tours queue and get triggered is a design decision that belongs to #1587 ("trigger evaluation"), not this issue.

## Test plan

- [x] New unit tests: `TourPlayer.test.ts` (rendering, step navigation, Back-disabled-on-first-step, Done-replaces-Next-on-last-step, anchored vs. centered positioning, Escape/backdrop/Skip dismissal, focus restore, reset-to-first-step on reopen) and `tutorial-anchor.test.ts` (register/resolve, unregister, key-change via `update`, no-op on unchanged key, destroy-guard against a second claimant)
- [x] `npm run test:unit:ci` — full suite passes (290 files / 3416 tests); `components/tour` coverage (97.32% lines / 79.62% branches / 96% functions / 96.29% statements) clears the `src/components/**` floor

Closes #1592

---
_Generated by [Claude Code](https://claude.ai/code/session_01SfUnvKTqDuGLRGuP4usgj3)_

### DIFF
--- a/UI/src/components/tour/TourPlayer.svelte
+++ b/UI/src/components/tour/TourPlayer.svelte
@@ -1,0 +1,252 @@
+<!--
+	Anchored tutorial tour ("coach-marks"): highlights a registered {@link tutorialAnchor} target and
+	positions a callout beside it with the current step's text and next/back/done controls. Distinct
+	from `Popover` (a container overlay with no target anchoring) — this positions relative to a real
+	on-screen control, and degrades to a centered, unanchored callout when the step names no anchor or
+	the anchor isn't (yet) registered, rather than failing.
+
+	A controlled component: the caller owns `open`/`steps` and is told about early dismissal
+	(`onDismiss` — Escape, backdrop, Skip) versus finishing the last step (`onComplete`) separately, so
+	it can decide what each means for the lesson's read state.
+-->
+{#if open && currentStep}
+	<div class="tour-layer">
+		<button class="tour-backdrop" type="button" tabindex="-1" aria-label="Dismiss tour" onclick={onDismiss}></button>
+
+		{#if spotlightRect}
+			<div
+				class="tour-spotlight"
+				style:top="{spotlightRect.top}px"
+				style:left="{spotlightRect.left}px"
+				style:width="{spotlightRect.width}px"
+				style:height="{spotlightRect.height}px"
+			></div>
+		{/if}
+
+		<div
+			class="tour-callout"
+			class:tour-callout--centered={!calloutPos}
+			role="dialog"
+			aria-modal="true"
+			aria-label={label}
+			data-tour-position={calloutPos ? 'anchored' : 'centered'}
+			tabindex="-1"
+			bind:this={shell}
+			style:top={calloutPos ? `${calloutPos.top}px` : undefined}
+			style:left={calloutPos ? `${calloutPos.left}px` : undefined}
+			use:focusTrap={{ onEscape: onDismiss, scrollLockClass: 'tour-open' }}
+		>
+			<div class="tour-callout__body" bind:this={calloutEl} aria-live="polite">
+				<p class="tour-callout__step">{stepIndex + 1} of {steps.length}</p>
+				<p class="tour-callout__text">{currentStep.text}</p>
+			</div>
+			<div class="tour-callout__controls">
+				<button class="tour-callout__skip" type="button" onclick={onDismiss}>Skip</button>
+				<div class="tour-callout__nav">
+					<button type="button" disabled={isFirst} onclick={goBack}>Back</button>
+					{#if isLast}
+						<button type="button" onclick={onComplete}>Done</button>
+					{:else}
+						<button type="button" onclick={goNext}>Next</button>
+					{/if}
+				</div>
+			</div>
+		</div>
+	</div>
+{/if}
+
+<script lang="ts">
+import { focusTrap, FOCUSABLE_SELECTOR } from '../focus-trap';
+import { getTutorialAnchor } from './tutorial-anchor';
+import type { TourStep } from './tour-types';
+
+interface Props {
+	/** Whether the tour is shown. */
+	open: boolean;
+	/** The tour's ordered steps. Resets to the first step whenever `open` transitions to `true`. */
+	steps: TourStep[];
+	/** Accessible name for the dialog (e.g. the lesson's display name). */
+	label: string;
+	/** Called on early dismissal — Escape, backdrop click, or the Skip control. */
+	onDismiss: () => void;
+	/** Called when the player finishes the tour via Done on the last step. */
+	onComplete: () => void;
+}
+
+const { open, steps, label, onDismiss, onComplete }: Props = $props();
+
+let stepIndex = $state(0);
+let shell = $state<HTMLElement | null>(null);
+let calloutEl = $state<HTMLElement | null>(null);
+let calloutPos = $state<{ top: number; left: number } | null>(null);
+let spotlightRect = $state<{ top: number; left: number; width: number; height: number } | null>(null);
+
+// A fresh open always starts from the first step; steps changing while already open (not a supported
+// use — reopen per lesson instead) deliberately leaves the current position alone.
+$effect(() => {
+	if (open) {
+		stepIndex = 0;
+	}
+});
+
+const currentStep = $derived(steps[stepIndex] as TourStep | undefined);
+const isFirst = $derived(stepIndex === 0);
+const isLast = $derived(stepIndex === steps.length - 1);
+const anchorEl = $derived(currentStep?.anchorKey ? getTutorialAnchor(currentStep.anchorKey) : undefined);
+
+const goBack = () => {
+	if (!isFirst) {
+		stepIndex -= 1;
+	}
+};
+const goNext = () => {
+	if (!isLast) {
+		stepIndex += 1;
+	}
+};
+
+// Recomputed on resize so a viewport change while the tour is open doesn't leave it pointing at a
+// stale rect (getBoundingClientRect is a snapshot, not itself reactive).
+let resizeTick = $state(0);
+$effect(() => {
+	if (!open) {
+		return;
+	}
+	const onResize = () => (resizeTick += 1);
+	window.addEventListener('resize', onResize);
+	return () => window.removeEventListener('resize', onResize);
+});
+
+const GAP = 12;
+const EDGE_PADDING = 8;
+const SPOTLIGHT_PADDING = 6;
+
+$effect(() => {
+	void resizeTick;
+	if (!open || !anchorEl) {
+		spotlightRect = null;
+		calloutPos = null;
+		return;
+	}
+
+	const rect = anchorEl.getBoundingClientRect();
+	spotlightRect = {
+		top: rect.top - SPOTLIGHT_PADDING,
+		left: rect.left - SPOTLIGHT_PADDING,
+		width: rect.width + SPOTLIGHT_PADDING * 2,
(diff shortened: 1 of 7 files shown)


UI/src/components/tour/TourPlayer.svelte
```diff
@@ -0,0 +1,252 @@
...
+		left: rect.left - SPOTLIGHT_PADDING,
+		width: rect.width + SPOTLIGHT_PADDING * 2,
+		height: rect.height + SPOTLIGHT_PADDING * 2
+	};
+
+	if (!calloutEl) {
+		return;
+	}
+	const calloutWidth = calloutEl.offsetWidth;
+	const calloutHeight = calloutEl.offsetHeight;
```
ginderjeremiah: Should-fix (minor): this measures `calloutEl` (`.tour-callout__body`), the inner element, but the positioned box is `.tour-callout` (`shell`), which adds `padding: 16px` + `1px` border on every side. So `calloutWidth`/`calloutHeight` under-report the real callout box by ~34px in each dimension — horizontal centering is off by ~17px, and the edge clamps let the callout extend ~34px past `EDGE_PADDING` at the right/bottom.

Measuring `shell` instead fixes both, and since `calloutEl` is used only for measurement here, that binding can then be dropped entirely (the `aria-live` region doesn't need a ref). Not blocking — degradation is graceful — but worth tightening.

---

ginderjeremiah (commented):
## Review: Anchored tutorial tour component (coach-marks) — #1592

**Verdict:** Approve *(submitted as a comment — I'm the PR author, so GitHub won't accept a gating verdict on my own PR)* — no blocking findings; one minor should-fix on the positioning math, and one deferred test I agree is out of scope until #1591 lands.

This lands cleanly against the issue's scope and the overlay/theming guidelines. The component is a fully controlled overlay, reuses the shared `focusTrap` chrome (Escape / Tab-trap / scroll-lock / focus restore) exactly as `docs/frontend-overlays.md` prescribes rather than re-rolling it, and matches the `Popover` scroll-lock-class-plus-`:global(body.…)` pattern. Anchors are keyed by stable string (not selector) and degrade to a centered callout when missing — as the issue requires. The `onDismiss`/`onComplete` split is the right call and mirrors `Modal`'s `onConfirm`/`onCancel`.

### Should-fix

- `TourPlayer.svelte:143-144` — positioning measures the inner `.tour-callout__body` (`calloutEl`) instead of the positioned `.tour-callout` shell, under-reporting the box by its padding + border (~34px/side). Graceful degradation, so not blocking; details inline.

### Nits

- The on-open/per-step focus `$effect` is a verbatim copy of `Popover`'s first-focusable idiom. Not worth extracting for four lines, but if a third overlay ever repeats it, a shared helper would be the DRY move. (Note only — no change requested.)

### Test coverage

- `TourPlayer.test.ts` and `tutorial-anchor.test.ts` are thorough: step navigation, Back-disabled-on-first, Done-replaces-Next-on-last, anchored vs. centered branches, all three dismissal paths, focus restore, reset-on-reopen, and the destroy-guard against a second claimant. Good coverage of the branches this component owns.
- The issue's "every `anchorKey` referenced by seeded lessons resolves to a registered anchor" test is **deliberately deferred** — `content/lessons.json` doesn't exist until #1591 merges, so there's nothing to assert against yet. Agree with the call; just make sure it actually gets written once real seeded steps exist (it's the frontend half of #1591's lint coverage, so it shouldn't quietly fall through the cracks between #1587/#1589/#1605).

### What's good

- Theming is fully token-driven (`--accent`, `--tooltip-bg`, `--border-light`, `--text-*`, `color-mix` over `--black`) — no hard-coded colour, per `frontend.md`.
- The `box-shadow` spotlight-cutout trick (dimming everything *except* the anchor box, no `clip-path` mask) is a nice, well-commented touch.
- Anchor destroy-guard (only clear the slot if this node still owns it) correctly handles the overlapping mount/unmount hazard, and it's explicitly tested.
- Accessibility is handled properly: real `<button>` elements, `role="dialog"`/`aria-modal`, `aria-live="polite"` on the step body, and keyboard-focus management via the shared trap.

Approving over the should-fix; feel free to fold the positioning tweak in before merge or take it as a quick follow-up. Holding off on merging myself since CI (frontend + e2e) is still running and the should-fix is open.

---

ginderjeremiah:
## Re-review: Anchored tutorial tour component (coach-marks) — #1592

**Verdict: Approve (approve-with-nits)** — the one should-fix from the previous review is resolved, and CI is green. No blocking findings.

### Since the last review

- ✅ **Positioning measured the wrong box** — addressed in `75177079`. The effect now sizes off `shell` (bound to `.tour-callout`, padding + border included) via `shell.offsetWidth/offsetHeight`, and the now-redundant `calloutEl` ref was removed. This was the only should-fix outstanding.

### Verification against project standards

- **Overlay chrome reuse is correct** — genuinely imports the shared `focusTrap` + `FOCUSABLE_SELECTOR` from `components/focus-trap.ts` (Escape / Tab-trap / scroll-lock / focus capture+restore), exactly as `docs/frontend-overlays.md` prescribes, with only the first-focusable initial focus left component-specific. Not a re-roll of the trap.
- **Theming is fully token-driven** — `--accent`, `--tooltip-bg`, `--border-light`, `--text-*`, and `color-mix` over `--black`; no hardcoded colours, per `frontend.md`. `z-index: 60` slots cleanly between Popover (40) and Toast (100) / Modal (200).
- **Scope is honest** — controlled component, no global store/host, `onDismiss`/`onComplete` split, local `TourStep` mirroring the future `LessonStep`. All the deferrals (#1587 trigger layer, #1589 Help consumer, the seeded-lesson anchor lint) are correctly punted with the entity (#1591) still unmerged.
- **Anchor destroy-guard** (only clear the slot if this node still owns the key) is the right analogue of the tooltip registry's overlapping mount/unmount hazard, and it's explicitly tested.

### Nits (non-blocking)

- `TourPlayer.svelte` — the per-step focus `$effect` (`void stepIndex; … (target ?? shell).focus()`) re-runs on every `stepIndex` change and moves focus to the *first* focusable in the callout, which is the **Skip** button. So a keyboard user pressing Enter on **Next** is bounced to Skip each step and must Tab back to advance again. `aria-live="polite"` already announces the new step text, so the forced refocus isn't needed for a11y. Consider only refocusing on open (not per step), or targeting the nav button rather than Skip. Minor UX polish — fine as a follow-up.

### Test coverage

- `TourPlayer.test.ts` and `tutorial-anchor.test.ts` cover the branches this component owns: step navigation, Back-disabled-on-first, Done-replaces-Next-on-last, anchored vs. centered positioning, all three dismissal paths, focus restore, reset-on-reopen, and the destroy-guard against a second claimant. Adequate per the frontend testing rules.
- The "every seeded `anchorKey` resolves to a registered anchor" lint test is correctly deferred until `content/lessons.json` exists (#1591) — just make sure it lands with whichever of #1587/#1589/#1605 first has real seeded steps, rather than slipping through the cracks.

### What's good

- The `box-shadow` spotlight-cutout trick (dim everything *except* the anchor box, no `clip-path` mask) is a clean, well-commented touch, and edge-aware flip/clamp positioning recomputed on resize is handled properly.
- Accessibility is solid: real `<button>` elements, `role="dialog"` / `aria-modal`, `aria-live="polite"` step body, keyboard focus via the shared trap.

Approving over the single nit. CI is green (verify-frontend, test-frontend, all three e2e browsers pass; backend jobs skipped — no backend changes).

---
_Review by Claude_

---

PR #1610, opened 2026-07-04
Drive the logical engine from a Web Worker tick source (#1594)

### LINKED ISSUE #1594: Drive the logical engine from a Web Worker tick source (background-tab throttling)
Part of spike #1073 — see `docs/spikes/1073-idle-time-recovery.md` for the full design and the owner-settled direction (worker-first).

## Problem

`LogicalEngine.logicLoop` is driven by `window.setInterval` (`UI/src/lib/engine/logical-engine.ts:37`). Browsers throttle hidden-tab page timers (>=1s, ~once/minute after sustained hiding), so poll deltas blow past the `tickSizeX5` catch-up clamp and the excess simulation time is discarded — a hidden character falls behind wall-clock indefinitely.

## Fix

Replace the page-timer tick source with a dedicated module worker that posts a message every `pollingIntervalMs`; drive `logicLoop` from `onmessage` on the main thread. Dedicated-worker timers are not throttled in Chromium/Firefox, and `message` delivery to a hidden page is not throttled either — so the engine keeps ticking at full rate while hidden and the clamp simply stops firing in the common case.

## Scope / notes

- Vite/SvelteKit worker: `new Worker(new URL('./tick-worker.ts', import.meta.url), { type: 'module' })`. Guard for environments without `Worker` (SSR, jsdom unit tests) by falling back to the current `window.setInterval` path — this doubles as the test seam.
- `logicLoop`/`update` stay unchanged: deltas are still measured via `performance.now()`, and the clamp + `onIdleTimeLost` remain as the safety valve / instrumentation for what a worker cannot cover (page freeze/discard, Safari tab suspension, the thaw catch-up poll).
- `BackgroundThrottleMonitor` (#1074) needs **no change** and self-gates: where the worker trick works, `onIdleTimeLost` stops firing while hidden, so the allowlist toast naturally only appears in the environments where it still helps.
- `stop()` must terminate the worker.
- Tests: existing logical-engine tests keep covering `update()`/the clamp; add coverage for tick-source selection via the fallback seam.

Out of scope: hidden-tab presentational side effects (follow-up issue, see spike doc), server-side resume (separate issues under #1073).

ginderjeremiah:
## Summary

Implements #1594 (spike #1073, Decision 1): replaces `LogicalEngine`'s `window.setInterval` tick source with a dedicated module worker.

- **`tick-source.ts`** (new): `createTickSource(onTick)` returns a `{ stop() }` handle. When `Worker` is available it spins up `tick-worker.ts` as a module worker and drives `onTick` from `worker.onmessage`; when it isn't (SSR, jsdom unit tests) it falls back to `window.setInterval` — the fallback is also the unit-test seam the issue asked for.
- **`tick-worker.ts`** (new): the worker entry point — `setInterval(() => postMessage(null), pollingIntervalMs)`. Dedicated-worker timers aren't throttled in Chromium/Firefox, and `postMessage` delivery to a hidden page isn't throttled either, so the engine keeps ticking at full rate while the tab is backgrounded instead of falling behind and hitting the `tickSizeX5` catch-up clamp.
- **`logical-engine.ts`**: `start()`/`stop()` now delegate to a `TickSource` instead of managing a raw interval handle directly. `logicLoop`/`update` are unchanged — deltas are still measured via `performance.now()`, and the catch-up clamp + `onIdleTimeLost` remain as the safety valve/instrumentation for what a worker can't cover (page freeze/discard, Safari tab suspension), per the spike's Decision 1.
- `BackgroundThrottleMonitor` (#1074) needs no change — it self-gates on `onIdleTimeLost`, which naturally stops firing while hidden wherever the worker trick works.

## Test plan

- [x] New `tick-source.test.ts`: covers both branches — Worker unavailable (fallback to `window.setInterval`, verified via `vi.useFakeTimers`) and Worker available (mocked `Worker` class asserting construction args, tick delivery via `onmessage`, and `terminate()` on stop).
- [x] Existing `logical-engine.test.ts` passes unchanged — jsdom has no global `Worker`, so it continues to exercise the fallback path.
- [x] `npm run test:unit:ci` — full suite passes, coverage thresholds met.
- [x] `npm run build` — confirms Vite's worker plugin picks up `new Worker(new URL('./tick-worker.ts', import.meta.url))` and emits a dedicated worker chunk (`_app/immutable/workers/tick-worker-*.js`).

Closes #1594

---
_Generated by [Claude Code](https://claude.ai/code/session_01TNkTVtHjz7kqRtZR16J2Wc)_

### DIFF
--- a/UI/src/lib/engine/tick-source.ts
+++ b/UI/src/lib/engine/tick-source.ts
@@ -13,6 +13,7 @@ export function createTickSource(onTick: () => void): TickSource {
 	if (typeof Worker !== 'undefined') {
 		const worker = new Worker(new URL('./tick-worker.ts', import.meta.url), { type: 'module' });
 		worker.onmessage = () => onTick();
+		worker.onerror = (ev) => console.error('The logical engine tick worker failed', ev);
 		return {
 			stop: () => worker.terminate()
 		};
--- a/UI/src/lib/engine/tick-source.ts
+++ b/UI/src/lib/engine/tick-source.ts
@@ -1,0 +1,25 @@
+// Dedicated-worker timers are not throttled when the tab is hidden (unlike page timers), so
+// driving the logical engine's polling loop from a worker is what keeps a backgrounded character's
+// simulation ticking at full rate instead of falling behind wall-clock. Environments without
+// `Worker` (SSR, jsdom unit tests) fall back to `window.setInterval` — that fallback also doubles
+// as the unit-test seam, since `LogicalEngine` itself has no environment branching.
+export const pollingIntervalMs = 10;
+
+export interface TickSource {
+	stop(): void;
+}
+
+export function createTickSource(onTick: () => void): TickSource {
+	if (typeof Worker !== 'undefined') {
+		const worker = new Worker(new URL('./tick-worker.ts', import.meta.url), { type: 'module' });
+		worker.onmessage = () => onTick();
+		return {
+			stop: () => worker.terminate()
+		};
+	}
+
+	const handle = window.setInterval(onTick, pollingIntervalMs);
+	return {
+		stop: () => window.clearInterval(handle)
+	};
+}
--- a/UI/src/lib/engine/logical-engine.ts
+++ b/UI/src/lib/engine/logical-engine.ts
@@ -1,11 +1,11 @@
 import { createHook, getEventCounter } from '$lib/common';
 import { MS_PER_TICK } from '$lib/api/types/game-constants';
+import { createTickSource, type TickSource } from './tick-source';
 
 // The logical tick size in ms — generated from the backend GameConstants so it can never drift
 // from the simulation the server replays (battle parity depends on it).
 export const tickSize = MS_PER_TICK;
 const tickSizeX5 = tickSize * 5;
-const pollingIntervalMs = 10;
 
 const logicalUpdateHook = createHook<[number]>();
 const notifyLogicalUpdate = logicalUpdateHook.notify;
@@ -24,26 +24,26 @@ export class LogicalEngine {
 	private lastTime = 0;
 	private timeBank = 0;
 	private countTick = getEventCounter((t) => (this.tickRate = Math.round(t)));
-	private tickHandle = 0;
+	private tickSource?: TickSource;
 
 	public start() {
-		if (this.tickHandle) {
+		if (this.tickSource) {
 			return;
 		}
 
 		this.lastTime = performance.now();
 		this.time = this.lastTime;
 
-		this.tickHandle = window.setInterval(() => this.logicLoop(), pollingIntervalMs);
+		this.tickSource = createTickSource(() => this.logicLoop());
 	}
 
 	public stop() {
-		if (!this.tickHandle) {
+		if (!this.tickSource) {
 			return;
 		}
 
-		window.clearInterval(this.tickHandle);
-		this.tickHandle = 0;
+		this.tickSource.stop();
+		this.tickSource = undefined;
 		this.lastTime = 0;
 		this.timeBank = 0;
 		this.time = 0;
--- a/UI/src/lib/engine/tick-worker.ts
+++ b/UI/src/lib/engine/tick-worker.ts
@@ -1,0 +1,6 @@
+// Dedicated module worker driving `LogicalEngine`'s tick source (see `tick-source.ts`). A worker's
+// own timers keep running at full rate while the page is hidden, and posting a message to a hidden
+// page is not throttled either — so this is the whole fix for background-tab throttling.
+import { pollingIntervalMs } from './tick-source';
+
+setInterval(() => postMessage(null), pollingIntervalMs);
--- a/UI/src/tests/lib/engine/tick-source.test.ts
+++ b/UI/src/tests/lib/engine/tick-source.test.ts
@@ -1,0 +1,57 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+
+import { createTickSource, pollingIntervalMs } from '$lib/engine/tick-source';
+
+afterEach(() => {
+	vi.unstubAllGlobals();
+	vi.useRealTimers();
+});
+
+describe('createTickSource', () => {
+	it('falls back to window.setInterval when Worker is unavailable (the jsdom/SSR case)', () => {
+		vi.stubGlobal('Worker', undefined);
+		vi.useFakeTimers();
+
+		const onTick = vi.fn();
+		const source = createTickSource(onTick);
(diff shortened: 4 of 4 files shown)


UI/src/lib/engine/tick-source.ts
```diff
@@ -0,0 +1,25 @@
...
+export const pollingIntervalMs = 10;
+
+export interface TickSource {
+	stop(): void;
+}
+
+export function createTickSource(onTick: () => void): TickSource {
+	if (typeof Worker !== 'undefined') {
+		const worker = new Worker(new URL('./tick-worker.ts', import.meta.url), { type: 'module' });
+		worker.onmessage = () => onTick();
```
ginderjeremiah, line 15: Nit (non-blocking): the worker path sets `onmessage` but no `onerror`. If the worker ever fails to load or throws on startup, ticks silently stop and the logical engine freezes with no fallback — the `typeof Worker !== 'undefined'` guard only covers *absence* of `Worker`, not a failed instantiation. Given this engine drives battle simulation and reward accrual, a silent freeze is a worse failure mode than a page timer.

The practical risk is low (the worker is a same-origin Vite-emitted chunk and its body is a single `setInterval(postMessage)` that can't realistically throw), so I'm not blocking on it. But `api-socket.ts:140` already wires `socket.onerror` on its message source, so there's a codebase precedent — consider at least a `worker.onerror` that logs, so a load failure is diagnosable rather than a mystery stall.

---

ginderjeremiah (commented):
## Review: Drive the logical engine from a Web Worker tick source (#1594)

**Verdict:** Approve *(posted as a comment since I'm the PR author — GitHub won't accept a gating approve on one's own PR)* — the change faithfully implements the spike's worker-first direction, is well-scoped, and is tested on both branches. One non-blocking nit.

This is a clean, minimal implementation. Extracting a `TickSource` abstraction and having `LogicalEngine` delegate `start()`/`stop()` to it keeps the engine free of environment branching (the issue's explicit intent), and `logicLoop`/`update` are untouched so the `tickSizeX5` clamp + `onIdleTimeLost` remain as the safety valve for what a worker can't cover (page freeze/discard, Safari suspension) — exactly as the issue specifies. All old references (`tickHandle`, the in-file `pollingIntervalMs`) are fully removed, no dead code left behind.

### Nits

- `tick-source.ts:15` — worker path has no `onerror`; a failed worker load would silently freeze the engine. Low practical risk, but there's a codebase precedent (`api-socket.ts:140`) for handling it. See inline comment.

### Test coverage

Solid and matches the issue's ask ("add coverage for tick-source selection via the fallback seam"):
- `tick-source.test.ts` covers **both** branches — the `Worker`-undefined fallback (via `vi.useFakeTimers`, asserting ticks fire and stop cleanly) and the `Worker`-available path (mock `Worker` asserting construction args `{ type: 'module' }`, the `tick-worker` URL, tick delivery via `onmessage`, and `terminate()` on stop).
- Existing `logical-engine.test.ts` keeps exercising `update()`/the clamp through the fallback path (jsdom has no global `Worker`), so that coverage is preserved.
- `tick-worker.ts` itself is untested, which is fine — it's a single `setInterval(postMessage)` line with no meaningful jsdom-testable surface, and the real-browser e2e suite exercises the actual worker (chromium/firefox green).

### What's good

- The worker-bundle URL pattern (`new Worker(new URL('./tick-worker.ts', import.meta.url), { type: 'module' })`) is the documented Vite/SvelteKit approach, and the PR verified the emitted worker chunk via `npm run build`.
- Sharing `pollingIntervalMs` between the worker and the fallback (rather than duplicating the `10`) is the right DRY call — the two intervals must stay in lockstep.
- Comments explain *why* (worker timers + message delivery aren't throttled while hidden) without over-explaining — good level of context for a non-obvious browser trick.

---

ginderjeremiah:
## Review: Drive the logical engine from a Web Worker tick source (#1594)

**Verdict: Approve** — faithfully implements spike #1073 Decision 1 and issue #1594, cleanly scoped, tested on both branches, CI green. No blocking findings.

I verified the diff against the issue/spike intent and `docs/frontend.md` rather than the PR description alone.

### What's good

- **Correct architectural shape.** Extracting a `TickSource` abstraction and having `LogicalEngine.start()/stop()` delegate to it keeps the engine free of environment branching — exactly the issue's stated intent ("the fallback also doubles as the test seam, since `LogicalEngine` itself has no environment branching"). `logicLoop`/`update` are untouched, so the `tickSizeX5` clamp + `onIdleTimeLost` correctly remain as the safety valve/instrumentation for what a worker can't cover (freeze/discard, Safari suspension), matching the spike.
- **No dead code.** The old `tickHandle` field and the in-file `pollingIntervalMs` are fully removed; `pollingIntervalMs` now has a single home in `tick-source.ts`, shared by both the worker and the fallback so the two intervals can't drift.
- **Worker bundling is the documented Vite/SvelteKit pattern** (`new Worker(new URL('./tick-worker.ts', import.meta.url), { type: 'module' })`), and the PR verified the emitted worker chunk via `npm run build`.
- **`onerror` handler is present** (`tick-source.ts`) — a worker-load failure is logged rather than silently stalling. (Noting this because an earlier self-review flagged its absence; it's addressed in the current head.)
- Comments explain the non-obvious *why* (worker timers + `postMessage` delivery aren't throttled while hidden) at a good level of detail.

### Nits (non-blocking)

- `tick-source.ts` — on a worker failure, `onerror` logs but the engine does not fall back to `setInterval`, so ticking stops entirely (foreground included) if the worker ever fails to load/run. Practical risk is very low (same-origin bundled asset, exercised by the real-browser e2e suite), so a follow-up rather than a change here — but if you want defense-in-depth, `onerror` could tear down and re-create via the `setInterval` path.

### Test coverage

Meets the project's frontend testing rules and the issue's explicit ask ("add coverage for tick-source selection via the fallback seam"):

- `tests/lib/engine/tick-source.test.ts` mirrors the source path and covers **both** branches — the `Worker`-undefined fallback (fake timers assert ticks fire and stop cleanly) and the `Worker`-available path (mock `Worker` asserts `{ type: 'module' }`, the `tick-worker` URL, tick delivery via `onmessage`, `terminate()` on stop), plus the `onerror` logging path.
- Existing `logical-engine.test.ts` keeps exercising `update()`/the clamp through the fallback (jsdom has no global `Worker`), so that coverage is preserved.
- `tick-worker.ts` itself is untested — acceptable: a single `setInterval(postMessage)` line with no jsdom-testable surface, exercised by the green chromium/firefox/webkit e2e runs.
- Not battle logic, so no backend mirror is required.

---
*Review by Claude*

---

PR #1606, opened 2026-07-04
Lesson reference data: entity, Workbench authoring, graph lint coverage (#1591)

ginderjeremiah:
## Summary

Implements #1591 (spike #1392, dependency-order-first issue: `#1591 → (#1588, #1592) → #1587 → #1589`): the `Lesson`/`LessonStep` reference-data entity, riding the standard content-authoring pipeline like every other authored set.

- **EF entity + migration**: `Lesson` (Key, Name, TriggerType, ScreenKey, TriggerMechanicEvent, Ordinal, DesignerNotes, RetiredAt — zero-based identity) + ordered child `LessonStep` (LessonId, Ordinal, Text, AnchorKey). `Key` carries a unique index — it's the stable slug the lint matches content-design's taught-by-blurb candidates against, distinct from the freely-reworded display `Name`.
- **Contracts + mapping**: `Lesson`/`LessonStep` read/authoring contracts, `LessonMapper` (entity ↔ contract), steps assembled ordered by `Ordinal`.
- **Reference cache + repo**: `LessonsCacheHolder`/`Lessons` (`ILessons`/`ILessonEntityCache`), following the simple entities-only snapshot shape (no derived index — lessons aren't consumed by battle logic).
- **Admin CRUD**: `IAdminLessons` — retire-only identity save (`SaveLessons`, with a trigger-consistency guard and a Key-collision guard mirroring `AdminProficiencies`' tier-ordinal check) + `SetSteps` (`ChildCollectionReconciler` keyed by `Ordinal`). `AdminLessonsController` at `/api/AdminTools/*`.
- **Client reference data**: `GetLessons` socket command, wired into both the game's `REFERENCE_DATA` manifest (so `staticData.loaded` doesn't regress for every player) and the Workbench's own reference loader.
- **Content export/seed**: `lessons.json` (currently `[]`), `ContentExporter`/`ContentSeeder`/`ContentImportReader`/`ContentGraph` all updated.
- **Progression-graph lint**: a `CheckLessons` pass — trigger structural consistency (mechanic-event lesson names an event XOR screen-visit lesson doesn't dangle one) and a `LessonCoverage` warning when a content-design.md §2 taught-by-blurb candidate (`idle-loop-basics`, `crit-dodge-variance`, `cooldown-charging`) has no live lesson.
[...]

PR #1603, opened 2026-07-04
Swap XP reward/proficiency accrual onto CombatRating; reshape reward curve (#1532)

ginderjeremiah:
## Summary

Implements #1532 (spike #1526 Decisions 4–6), the consumer swap now that the combat rating (#1531) and its calibration report (#1533) have landed:

- **XP reward — enemy-authored bounty, no upward premium.** `DefeatRewards` now rates the battle-simulated player `Battler` and the enemy's *fielded* loadout via `CombatRating.Rate` instead of summing core attributes. The reward is `ExpReward = floor(k × EnemyRating × min(EnemyRating ÷ PlayerRating, 1)²)` — the old ±20% flat band, the upward quadratic premium, and the `MaxExpRewardMultiplier` cap all dissolve into this one continuous, self-capping curve. At or above a matched fight the reward saturates at the enemy's bounty (punching up no longer pays extra); below it, the reward drops off quadratically (anti-grind, unchanged in spirit).
- **Proficiency accrual — max-normalized.** `XP_path = pie × pathActivity ÷ max(PlayerRating, EnemyRating)`. Above your weight the claim rate is what a matched player would earn; at/below, the treadmill is unchanged. The old clamp (`ServerGameConstants.MaxExpRewardMultiplier`) is **dropped entirely** rather than kept as a smaller safety clamp — a path's activity is already bounded upstream by the enemy's health pool (#1482), so there's no unbounded-payout risk left to guard against, and the redesign's whole point is that the rating no longer needs an exploit-proofing cap to stay honest.
- **Threading.** `BattleService.RecordVictory` rates the player from the frozen snapshot and the enemy from its fielded `BattleSkills` (already selected before `RecordVictory` runs). `OfflineProgressSimulator` rates the player once per stationary away window and the enemy per encounter. `BattleStats.PlayerRating`, `BattleCompletedEvent.PlayerRating`/`EnemyRating`, and `OfflineBattleOutcome.PlayerRating`/`EnemyRating` carry the ratings (renamed from `PlayerPower`); `Player.RecordBattleCompleted` threads both.
- **Constants.** `ServerGameConstants.XpScaleK` (new) and the retuned `ProficiencyXpPerVictory` come from running `CombatRatingCalibrator`'s integration test against the current seeded content — anchored at the zone/level/build sample the old curve already treated as matched (Ashen Wastes L8, Offense build), so the swap pays continuously with the old curve at that reference point.
[...]

PR #1611, opened 2026-07-04
Hand back a still-in-progress stale battle instead of booking a draw (#1595)

### LINKED ISSUE #1595: Offline catch-up: hand back a still-in-progress stale battle instead of booking a draw

ginderjeremiah:
## Summary

Implements #1595 (spike #1073, Decision 2): `BattleService.AbandonBattle` (which backs `ResolveStaleBattle`, and is also invoked when `StartBattle`/`StartBossBattle` supersede an active battle) previously booked *any* non-concluding replay as a draw/loss, discarding real progress whenever real elapsed time since `BattleStartTime` hadn't actually reached the 2-minute cap. Now:

- **Three-way split in `AbandonBattle`'s non-concluding branch:** a win still resolves as before; a loss (`PlayerDied`) or a genuine draw (both alive **and** `elapsedMs >= DefaultMaxBattleMs`) is still resolved and cleared exactly as before; but when neither battler died and elapsed time is **under** the cap, the battle is genuinely still in progress — it is left completely untouched (same enemy/seed/snapshot, `BattleStartTime` unchanged, nothing cleared or persisted) and handed back as a `BattleStartResult` with a new `ElapsedOffsetMs` (the real elapsed ms since `BattleStartTime`).
- **`AbandonBattle`/`ResolveStaleBattle` now return `BattleStartResult?`** — non-null only for the still-in-progress handoff.
- **`StartBattle` (`NewEnemy`)** returns the handoff instead of starting a fresh battle when the existing one is still in progress — the delivery point the spike names explicitly.
- **`StartBossBattle` (`ChallengeBoss`) deliberately keeps overriding** an in-progress battle (discarding the handoff) since challenging the boss is an explicit different action, not "give me my battle back" — this matches the existing, already-tested behavior (`StartBossBattle_ActiveBattleInProgress_AbandonsItBeforeStarting`).
- **`OfflineProgressService.SimulateProgress`** (backing `GetOfflineProgress` and the character-switch credit) now checks `ResolveStaleBattle`'s handoff: when still in progress, it returns early with a new `OfflineProgressSummary.ActiveBattle`/`StillInProgress(...)` — no away-window battles are simulated (there's no settled outcome or leftover budget to spend), and `Player.LastActivity` is deliberately left untouched so the away clock keeps counting against the original disconnect.
- **Wire contract:** `EnemyInstance` (shared by `NewEnemy`/`ChallengeBoss`) gains `ElapsedOffsetMs`, and `OfflineProgressModel` gains `ActiveBattle` — both non-null only for a still-in-progress hand-back. TypeScript client regenerated (`dotnet run --project Game.Api -- codegen`).

This defines the "battle + elapsed offset" delivery contract the linked client issue (#1597, replay-to-offset) will consume; that work is out of scope here.

## Judgment call

The issue left `ChallengeBoss`'s behavior implicit. I kept it always overriding an in-progress idle battle (rather than also handing back), since it's a deliberate player action distinct from "give me a new enemy," and flipping it would have broken the existing `StartBossBattle_ActiveBattleInProgress_AbandonsItBeforeStarting` test's intent. Flagging in case that's worth reconsidering later.

## Tests

- `BattleServiceIntegrationTests`: rewrote `StartBattle_AbandoningAnUnresolvedBattle_RecordsAbandonedNotLost` (the old scenario is now exactly the still-in-progress case, so its assertions were inverted) into `StartBattle_AbandoningAStillInProgressBattle_HandsItBackWithoutRecordingAnOutcome`; added `ResolveStaleBattle_Concludes_ReturnsNullAndCreditsTheWin` and `ResolveStaleBattle_StillInProgressAndGenuineDrawAtCap` covering the three partitions the issue asks for (concludes / still-in-progress / genuine draw at cap) directly against `ResolveStaleBattle`.
- `OfflineProgressServiceIntegrationTests`: added `SimulateSwitchProgress_StaleBattleStillInProgress_HandsItBackWithoutSimulatingAwayWindow` (the realistically reachable path, since the character-switch credit has no away-time floor — under the 5-minute login floor this branch is essentially unreachable in practice).

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test` — full suite passes (2919 tests: Core 1317, Infrastructure 59, Application 856, Api 689 — 2 new + old test rewritten)
- [x] `dotnet run --project Game.Api -- codegen` — regenerated `enemies.ts`/`progress.ts`
- [x] `npm run test:unit:ci` — full suite passes (3400 tests)

Closes #1595

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ti8NVsx3PTm3uazw8mcfQA)_


Game.Application/Services/BattleService.cs
```diff
@@ -379,6 +396,20 @@ private async Task AbandonBattle(Player player, PlayerState state, int? clientBa
...
+                // Neither battler died and the replay fell short of the cap: this is not a stalemate, it is a
+                // battle genuinely still in progress (#1595) — not enough real server time has elapsed since
+                // BattleStartTime for it to have concluded. Leave it active (same enemy/seed/snapshot,
+                // BattleStartTime unchanged) and hand it back with its elapsed offset instead of booking a
+                // phantom draw. Only a replay that reaches the cap with both battlers alive is a genuine draw.
+                return new BattleStartResult
+                {
+                    Enemy = enemy,
+                    Seed = state.BattleSeed ?? 0,
+                    ElapsedOffsetMs = wallClockMs,
```
ginderjeremiah: **Should-fix (subtle, low practical impact):** the still-in-progress partition is keyed on `elapsedMs`, but `ElapsedOffsetMs` (and the issue's own definition of the partition — "real elapsed time `now - BattleStartTime`") is `wallClockMs`. These two clocks diverge only when a `clientBattleMs` is reported and is smaller than the real wall-clock, i.e. the **live `StartBattle` abandon path** (the `NewEnemy` production call site), never the `ResolveStaleBattle`/offline path where `clientBattleMs` is null and `elapsedMs == wallClockMs`.

In that live path, if `wallClockMs >= DefaultMaxBattleMs` while the reported `clientBattleMs` is small, this branch hands the battle back as "still in progress" with an `ElapsedOffsetMs` past the cap — even though real elapsed time says it's a genuine draw. It's unreachable in honest play (a connected, actively-fighting client's `clientBattleMs` tracks wall-clock) and confers no cheat advantage (a draw pays nothing either), so this isn't a blocker — but the two-clock split is a latent trap: raising the cap, or the client path becoming primary, would surface it as a visible bug.

Two clean options: key the guard on `wallClockMs` (matches the issue and the offset), or clamp `ElapsedOffsetMs` to the validated/simulated window. If the current `elapsedMs` keying is deliberate (it's consistent with `simulateMs = min(elapsedMs, cap)` — you only replayed what the client claims it fought), a one-line comment noting why the guard and the offset use different clocks would keep the next reader from "fixing" one to match the other.

---

ginderjeremiah (commented):
## Review: hand back a still-in-progress stale battle instead of booking a draw (#1595)

**Intended verdict:** Approve _(submitted as a comment — GitHub won't let me formally approve my own PR)_ — the three-way partition in `AbandonBattle` is correct, `PlayerState` is genuinely left untouched on handoff, the wire contract is complete, and the tests cover the partitions the issue asks for. One should-fix and one test-coverage nit, neither blocking.

### Should-fix
- `BattleService.cs:399-410` — the still-in-progress guard keys on `elapsedMs` while `ElapsedOffsetMs` and the issue's own definition key on `wallClockMs`. They diverge only in the live `StartBattle`+`clientBattleMs` path; low practical impact but a latent two-clock trap. Details inline.

### Nits
- **Test coverage:** every new/rewritten test exercises the handoff branch either through `ResolveStaleBattle` (no `clientBattleMs`) or through `StartBattle` without passing `clientBattleMs`. The actual production `NewEnemy` call site passes `clientBattleMs`, so the `clientBattleMs`-reported handoff path — exactly where the clock subtlety above lives — is untested. A case that sets `clientBattleMs` on an in-progress `StartBattle` abandon would close that gap.

### Test coverage
- Backend integration coverage is at the right level (application-layer orchestration over Postgres/Redis; the underlying `Game.Core` simulator is unchanged). The three issue-mandated partitions are covered directly against `ResolveStaleBattle` (concludes / still-in-progress / genuine draw at cap), the rewritten `StartBattle` test asserts the no-outcome, untouched-state handoff, and `SimulateSwitchProgress_StaleBattleStillInProgress_...` covers the offline seam — correctly placed on the switch path, since the login path's 5-minute floor plus `BattleStartTime ≤ LastActivity` makes the handoff unreachable there. Frontend changes are pure codegen output (`enemies.ts`/`progress.ts`), correctly regenerated rather than hand-edited, with no logic to test until the #1597 client consumer lands.

### What's good
- Clean three-way split with the anti-cheat timeline preserved (offset derives from server-owned `BattleStartTime`; no new trusted surface). The `ChallengeBoss` "always override, discard the handoff" decision is well-reasoned and safe (nothing was cleared/persisted for an in-progress abandon), and flagged for reconsideration rather than silently chosen. `LastActivity` left untouched on handoff is the right call and is asserted. Comments are appropriately scoped to the non-obvious timing/anti-cheat reasoning, and the shared-contract seam (`ElapsedOffsetMs` on the one `EnemyInstance` projection, `ActiveBattle` on the offline model) is defined once and reused. Docs updated in `backend-battle.md`.

---

ginderjeremiah:
## Review from Claude — Verdict: Approve with nits

Reviewed against #1595 / spike #1073 (Decision 2), with #1597 (client replay-to-offset) confirmed out of scope. Checked the branch code directly, not just the diff. **CI is fully green** (all 9 checks, incl. `test-backend`, `test-frontend`, all three `test-e2e` browsers).

### Verification of the PR's key claims — all confirmed
- **(a) Three-way split in `AbandonBattle`'s non-concluding branch** — correct: `Victory` pays out; the new `else if (!result.PlayerDied && wallClockMs < DefaultMaxBattleMs)` hands back a still-in-progress battle; the final `else` books a loss (`PlayerDied`) or a genuine draw at the cap. All partitions reachable and correctly separated.
- **(b) Still-in-progress path leaves state untouched** — confirmed: the branch `return`s the `BattleStartResult` *before* `state.ClearBattle()` and `SavePlayer`, so enemy/seed/snapshot/`BattleStartTime` are untouched and nothing is persisted; `StartBattle` returns the handoff early without further mutation.
- **(c) `LastActivity` untouched on the offline still-in-progress path** — confirmed: the early `return StillInProgress(...)` in `SimulateProgress` sits ahead of the `player.StampActivity(now)` / `SavePlayer` re-anchor.
- **(d) `ChallengeBoss` keeps overriding** — confirmed: `StartBossBattle` discards `AbandonBattle`'s return and overwrites state via `SetActiveBattle`. Sound judgment call, and safe because nothing was cleared/persisted for the discarded handoff.

### Strengths
- The commit-2 fix to gate the classification on `wallClockMs` (real server elapsed) rather than the client-bounded `elapsedMs` is exactly right — it closes the inverse exploit where an under-reported `clientBattleMs` could keep an already-expired battle "in progress" with an offset past its own cap. Nicely pinned by `StartBattle_ReportedClientBattleMsUnderCapButRealElapsedPastCap_RecordsGenuineDrawNotHandoff`.
- Test coverage hits all three issue partitions directly against `ResolveStaleBattle` plus the divergent-clocks regression and the switch-path offline handoff.
- No battle-parity surface touched — the change is battle *lifecycle/handoff* in the application layer; `ElapsedOffsetMs` is plain wall-clock, and the parity-sensitive replay-to-offset is correctly deferred to #1597. `docs/backend-battle.md` updated accordingly. No mirrored FE test required for this PR.
- Wire contract (`EnemyInstance.ElapsedOffsetMs`, `OfflineProgressModel.ActiveBattle`) is additive/optional with regenerated TS client; no entity leakage across layers.

### Nits / should-fixes (none block merge — all follow-up candidates)
1. **Stale XML doc (should-fix).** `SimulateSwitchProgress`'s summary still reads "Resolves the in-flight battle, applies the rewards, re-anchors `LastActivity`, and persists" — no longer true on the still-in-progress hand-back (which deliberately does none of those). Worth a one-line tweak.
2. **Missing targeted `ChallengeBoss` test (should-fix / follow-up).** The discard-the-handoff path (boss override of an *under-cap, still-in-progress* idle battle) is only exercised indirectly by the pre-existing `StartBossBattle_ActiveBattleInProgress_AbandonsItBeforeStarting`, which resolves a concluded/expired battle and so doesn't actually enter the new non-null branch. A test that starts an under-cap non-concluding battle and asserts the boss battle cleanly overrides it would pin the documented judgment call.
3. **Interim empty welcome-back gate (nit, for #1597).** `HasProgress` now returns true for a still-in-progress-only summary, but today's client doesn't consume `ActiveBattle`, so it could briefly render a welcome-back gate showing zero of everything. The window is extremely narrow (character-switch path only — the login path is unreachable under the 5-min floor, as the PR notes), and #1597 resolves the rendering. Acceptable as a forward contract; flagging so it isn't forgotten when #1597 lands.
4. **`Seed = state.BattleSeed ?? 0` (nit).** Defensive fallback; harmless since `HasActiveBattle` implies a seed, but a `?? 0` would silently desync a client replay if that invariant ever broke.
5. **Observation (pre-existing, not introduced here).** Battle-outcome classification lives in `BattleService` (Application) rather than `Game.Core`; the new branch follows that established local pattern. If it's ever worth realigning to the "battle decisions are domain decisions" guidance, that's a tech-debt item for the whole `AbandonBattle`, not this PR.

Clean, well-reasoned, well-tested change that correctly settles the shared contract for #1597. Recommend merging after (optionally) tidying nits 1–2.


---

ginderjeremiah:
Follow-up on my review: pushed a maintenance commit before merging.

- **Rebased onto current `main`** — it had advanced past this PR's base (the reward-math swap onto `CombatRating` #1532/#1603, the lessons entity #1606, and the dead-stat lint #1607 all landed today). The rebase was textually clean, but the reward change is a **semantic** overlap: it swapped `BattleService`/`OfflineProgressService` reward math onto `CombatRating`, which this PR also modifies.
- That surfaced one real breakage: `ResolveStaleBattle_Concludes_ReturnsNullAndCreditsTheWin` began failing because its enemy skill (`1` dmg) is now capability-trivial under the rating-based reward, so the anti-grind curve floored the bounty to `0` and `player.Exp > expBefore` no longer held. Rebalanced it the same way #1603 rebalanced its sibling reward tests — raised the enemy skill's authored damage so its `CombatRating` is comparable to the player's, keeping the cooldown far longer than the one-shot fight so it never lands and the win outcome is unchanged.
- **Fixed the stale XML doc** flagged in the review: `SimulateSwitchProgress`'s summary no longer claims it always resolves/credits/persists — it now describes the still-in-progress hand-back (battle, `LastActivity`, and persistence left untouched).

Validated locally against the running containers: full backend suite green (Core 1303, Infrastructure 59, Application 877, Api 690 — 0 failures), `dotnet format --verify-no-changes` clean solution-wide.

On the "missing targeted `ChallengeBoss` test" note from the review — I checked and it's already covered: `StartBossBattle_ActiveBattleInProgress_AbandonsItBeforeStarting` starts a fresh idle battle (both alive, ~0 ms elapsed) and immediately challenges the boss, which is exactly the still-in-progress branch, and asserts the boss overrides the handoff. So no new test needed. Will merge once CI is green.

---

PR #1615, opened 2026-07-05
Fix TourPlayer stealing keyboard focus to Skip on step change

### LINKED ISSUE #1613: TourPlayer: per-step focus effect steals keyboard focus to the Skip button
### Context

Follow-up from PR #1609 (issue #1592), the anchored coach-marks `TourPlayer.svelte`.

### Gap

The per-step focus `$effect` re-runs on every `stepIndex` change and moves focus to the **first** focusable element in the callout — which is the **Skip** button. A keyboard user who advances the tour by pressing Enter on **Next** is bounced back to Skip on every step transition, which is a poor and slightly hazardous keyboard flow (repeated Enter could land on Skip).

`aria-live="polite"` already announces step-content changes to screen readers, so the forced refocus on each step isn't needed for announcement.

### Ask

Don't force-refocus the first focusable on every step change. Either keep focus where it is across step transitions (focus is already trapped within the callout by the shared `focusTrap`), or move it to a more sensible target such as the primary action (Next/Done) rather than Skip. Capture/restore of focus on open/close should stay as-is. Add/adjust a test asserting that advancing steps does not pull focus onto Skip.

### Note

This component is not yet wired into any live flow (consumers land with #1587/#1589), so there's no user-facing impact today — but worth fixing before it's mounted for real.

ginderjeremiah:
## Summary

`TourPlayer.svelte`'s per-step focus `$effect` re-ran on every `stepIndex` change and moved focus to the **first** focusable element in the callout, which is the **Skip** button. A keyboard user advancing the tour by pressing Enter on **Next** was bounced back to **Skip** on every step transition — a poor and slightly hazardous keyboard flow.

## Fix

- Bound a `primaryButton` ref to whichever of **Next**/**Done** is currently rendered and retarget the focus effect at it instead of "first focusable via `FOCUSABLE_SELECTOR`".
- Because the same `Next` button DOM node persists across `Next` → `Next` transitions (Svelte only creates a new node when the `{#if isLast}` branch flips to `Done`), the effect no longer needlessly re-fires on every step — focus naturally stays wherever the user last put it (e.g. on `Next`, or on `Back` if they clicked it).
- The effect still re-anchors focus when `Next` swaps for `Done` (a genuinely new DOM node that would otherwise lose focus), and still sets initial focus on open — now onto the primary action rather than Skip.
- Removed the now-unused `FOCUSABLE_SELECTOR` import.
- Capture/restore of focus on open/close (owned by the shared `focusTrap`) is unchanged.

## Test plan

- [x] Added unit tests: focuses `Next` on open (not `Skip`), does not steal focus back on step advance, and moves focus onto `Done` when it swaps in for `Next`.
- [x] `npx vitest run src/tests/components/tour/TourPlayer.test.ts` — 13/13 passing
- [x] `npm run lint` — clean
- [x] `npm run test:unit:ci` — full suite, 3434/3434 passing

Closes #1613

---
_Generated by [Claude Code](https://claude.ai/code/session_01XePrcm1mua9jfX3YyfzoRP)_

### DIFF
--- a/UI/src/components/tour/TourPlayer.svelte
+++ b/UI/src/components/tour/TourPlayer.svelte
@@ -45,9 +45,9 @@
 				<div class="tour-callout__nav">
 					<button type="button" disabled={isFirst} onclick={goBack}>Back</button>
 					{#if isLast}
-						<button type="button" onclick={onComplete}>Done</button>
+						<button type="button" bind:this={primaryButton} onclick={onComplete}>Done</button>
 					{:else}
-						<button type="button" onclick={goNext}>Next</button>
+						<button type="button" bind:this={primaryButton} onclick={goNext}>Next</button>
 					{/if}
 				</div>
 			</div>
@@ -56,7 +56,7 @@
 {/if}
 
 <script lang="ts">
-import { focusTrap, FOCUSABLE_SELECTOR } from '../focus-trap';
+import { focusTrap } from '../focus-trap';
 import { getTutorialAnchor } from './tutorial-anchor';
 import type { TourStep } from './tour-types';
 
@@ -77,6 +77,7 @@ const { open, steps, label, onDismiss, onComplete }: Props = $props();
 
 let stepIndex = $state(0);
 let shell = $state<HTMLElement | null>(null);
+let primaryButton = $state<HTMLButtonElement | null>(null);
 let calloutPos = $state<{ top: number; left: number } | null>(null);
 let spotlightRect = $state<{ top: number; left: number; width: number; height: number } | null>(null);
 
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: TourPlayer per-step focus no longer steals focus to Skip (#1613)

**Verdict:** Approve — the fix correctly resolves the issue with clean reasoning and adequate tests. (Submitted as a comment since I'm the PR author; GitHub won't accept a gating verdict on one's own PR.)

The change retargets the focus `$effect` from "first focusable in the callout" (Skip) to a `bind:this` ref on the primary action (Next/Done). The key insight is sound: because Svelte only creates a new DOM node when the `{#if isLast}` branch flips, `primaryButton` reassigns only on the Next↔Done swap — so the effect no longer re-fires on every `stepIndex` change and focus is left wherever the user last put it. It also drops the `void stepIndex` dependency, so repositioning/resize no longer steals focus either. Verified the reactivity: when closed, `open` short-circuits before `primaryButton` is read so it isn't a stale dependency; on reopen the effect re-runs and re-anchors onto Next.

### Test coverage
- Solid. Three new tests cover the exact scenarios that matter: initial focus on Next (not Skip), no focus-steal-back to Skip on advance, and re-anchor onto Done when it swaps in. These map directly to the issue's ask. Frontend-only change, so backend/parity suites aren't in play.

- Going *Back* off the last step swaps Done→Next, which reassigns `primaryButton` and moves focus onto Next. This is a benign focus jump (never onto Skip, so the issue's concern still holds) and is arguably the desired "re-anchor when the nav control changes" behavior — just noting it as the one case where focus isn't left untouched.

### What's good
- The 5-line comment explaining the node-persistence subtlety is exactly the kind of documentation this deserves — it's a genuine footgun a later "simplification" could break, and the comment guards against that. Clean removal of the now-unused `FOCUSABLE_SELECTOR` import, no null-forgiving operators, correct nullable handling.


---

Issue #1586, opened 2026-07-04
Content-events derivation layer (shared before/after-diff for mechanic occurrence)


ginderjeremiah:
Part of spike #1392 (design in `docs/spikes/1392-ui-tutorials.md`, area A). Content-events scoping resolved via PR #1585 review: this lands as part of the #1392 tutorial work, not a standalone dependency.

## What

A shared before/after-diff derivation module for detecting "a mechanic just occurred for the first time" (first crit landed, first dodge occurred, cooldown visibly recharged) and "something just unlocked" (proficiency/skill/item/challenge), consolidating the existing pattern already duplicated three times in the frontend:

- `UI/src/lib/common/challenge-unlocks.ts` (`zonesUnlockedBy`, `resolveUnlockReward`)
- `UI/src/lib/common/synthesis-feedback.ts` (`isRecipeCreatable`/`creatableRecipeIds`)
- `UI/src/lib/common/proficiency-feedback.ts` (level-up/milestone feedback)

Each is a pure function comparing player/battle state before and after a change and reporting what became newly true. This issue extracts/generalizes that pattern into one module the tutorial mechanic-anchored triggers (#1392) can consume, and that #1395 (notification badges) should also consume once it's picked up, rather than re-deriving "did X just unlock" a fourth time.

## Scope

- Pure, unit-testable diff functions — no UI, no persistence.
- Cover at minimum: first crit, first dodge, first cooldown-charge-visible, proficiency level-up/milestone, skill unlocked, item unlocked, challenge completed.
- Existing call sites (`challenge-unlocks.ts` etc.) migrate to the shared module where it's a clean win; leave call sites that would meaningfully change behavior/tests untouched and note why.

## Tests

Unit tests per detector (each before/after pair correctly reports "no change" vs. "newly true"), mirroring the existing `*-feedback.ts` test suites.

Related: #1392, #1395.

Issue #1392, opened 2026-06-30
Add UI tutorials


ginderjeremiah:
Plan out UI tutorials for each page and develop a robust system for displaying tutorials.

---

## Cross-reference: content-design teaching lanes

The content plan ([docs/content-design.md](../blob/main/docs/content-design.md) §2) now treats **tutorial blurbs as one of two teaching lanes** (the other being zone design): a mechanic a player can grasp from a single explanation (crit/dodge variance, cooldown charging, the idle loop) is taught by a **blurb** instead of spending a whole zone on it. That makes this system a **dependency of the content arc**, not just UX polish — the planned progression assumes blurbs exist to carry the lessons that don't warrant a zone.

Questions to fold into the spike:
- **Trigger model** — what fires a blurb? First visit to a page, first time a mechanic actually occurs in battle, reaching a zone/level, completing a challenge?
- **Authoring** — is tutorial content **reference data** (authored like other content, so it rides the [#1390](https://github.com/ginderjeremiah/GameServer/issues/1390) export/seed + Workbench), or hardcoded? This is the fork that decides where it lives.
- **Per-page vs per-mechanic** — the issue frames "tutorials for each page," but the content arc also wants **per-mechanic** blurbs not bound to a single page. Reconcile the two axes.
- **Dismissal / replay** — one-shot vs revisitable from a help menu (and how dismissal state persists per player).

ginderjeremiah:
Posted a research + recommended design spike doc: #1585 (`docs/spikes/1392-ui-tutorials.md`). Docs-only, doesn't close this issue — holding off on filing the candidate implementation issues until the direction gets a look.

**tl;dr:** no tutorial system exists today, but three adjacent pieces of infra make this cheaper than it looks — a config-driven screen registry with an already-reserved-but-unbuilt `help` screen, a one-shot dismissible-notice precedent (`BackgroundThrottleMonitor`), and a recurring "diff what changed → toast" pattern (`challenge-unlocks.ts`/`synthesis-feedback.ts`/`proficiency-feedback.ts`) that's a natural home for mechanic-occurrence triggers.

Recommended resolution to the four open questions:
1. **Trigger model** — every lesson is screen-anchored or mechanic-anchored; both feed the same toast pipeline, so "per-page vs. per-mechanic" isn't a fork to pick between.
2. **Authoring** — hardcode the blurb copy (mirrors the one existing precedent), but reference existing content ids for the trigger condition — no new Workbench entity.
[...]

PR #1619, opened 2026-07-05
Tutorial trigger evaluation + unread lesson flow (client)

### LINKED ISSUE #1587: Tutorial trigger evaluation + unread lesson flow (client)

ginderjeremiah:
## Summary

Delivers the client-side trigger evaluation and unread-lesson flow for spike #1392's tutorial system, wiring the two lesson trigger kinds (from the already-merged #1591/#1588/#1592) into the live game:

- **Screen-anchored triggers** — `evaluateScreenTrigger` (new `UI/src/lib/engine/tutorials.ts`) is hooked at the single screen-activation point in `routes/game/+page.svelte` (screens are registry state, not routes). The first activation of a screen with a locked screen-anchored lesson plays its tour immediately. Gated on the welcome-back gate having passed (`entered`), so a lesson can't fire against the default screen before the player has "arrived."
- **Mechanic-anchored triggers** — `evaluateMechanicTriggers` is wired into `BattleEngine.logicalUpdate`'s per-tick activation loop (live loop only — never the headless parity simulator, since this is a client-detection/presentation concern like the existing combat floaters). A locked mechanic lesson whose content-events detector (`isFirstCrit`/`isFirstDodge`/`isFirstCooldownRecharge`, #1586) fires this tick is unlocked — queued as **unread** rather than displayed immediately, since the player may be AFK.
- **Unread indicator** — a persistent badge on the sidebar's Help nav entry (`NavSidebar.svelte`), derived from `playerManager.lessons`, visually distinct from the "wip" badge and the toast channel.
- **Shared open flow** — `openLesson`/`closeTutorialTour` (navigate to the host screen via the `navigation` store → play the tour via a new `tutorialTour` store → mark the lesson read on completion *or* early dismissal, since Skip means "got it," not "remind me later"). This is exported for #1589 (Help reading-room screen) to reuse for manually opening/replaying a lesson.

A new `TourPlayer` host lives in `routes/game/+page.svelte`, reading the new `stores/tutorial-tour.svelte.ts` state.

## Judgment call to flag

`isLocked` gating (a lesson absent from `playerManager.lessons`) does double duty as the "was this true before" flag the content-events detectors expect — since `unlockLesson`/`markLessonRead` are themselves idempotent, once a lesson unlocks it naturally drops out of consideration on every later tick, so no separate "ever seen" flag is needed. Flagging this since it's a slightly different shape than the content-events module's own doc comment ("callers own snapshotting and any one-shot/ever-seen state") suggests.

The Help nav badge is visual-only for now (matching the existing "wip" badge's expanded-rail-only visibility) — clicking through to a *specific* unread lesson from a list is #1589's job (the reading room), which doesn't exist yet. `openLesson` is ready for it to call.

## Test plan

- [x] `npm run lint` (eslint + `svelte-check`) — clean
- [x] `npm run test:unit:ci` — full suite green (3459/3459), including new/updated coverage:
  - `tests/lib/engine/tutorials.test.ts` — trigger evaluation (first-activation semantics, locked/unread/read gating, no re-fire once unlocked/read, retired-lesson exclusion) and the open→navigate→play→mark-read flow
  - `tests/lib/engine/battle/battle-engine.test.ts` — the live loop wires tick activations into `evaluateMechanicTriggers` only when something actually fired
  - `tests/components/nav-menu/NavSidebar.test.ts` — unread badge visibility/count, superseding the wip badge
- [x] `npm run build` — production build succeeds
- [x] Docs updated: `docs/frontend.md` (new "Tutorials" section) and `docs/content-design.md` (dropped the "still landing" caveats now that #1587 lands alongside the already-merged #1591/#1588/#1592 — only #1589's reading room remains)

Closes #1587

---
_Generated by [Claude Code](https://claude.ai/code/session_01UtSwKyeUSmtfkK6T5ZAREf)_


UI/src/lib/engine/tutorials.ts
```diff
@@ -0,0 +1,89 @@
...
+ * passing the newly-activated screen's key — the single activation point cross-screen navigation
+ * already uses (`routes/game/+page.svelte`; screens are registry state, not routes). The first
+ * activation of a screen with a locked screen-anchored lesson plays its tour immediately, since the
+ * player just navigated there.
+ */
+export function evaluateScreenTrigger(screenKey: string) {
+	const lesson = liveLessons().find(
+		(candidate) => candidate.triggerType === ELessonTriggerType.ScreenVisit && candidate.screenKey === screenKey
+	);
+	if (lesson && isLocked(lesson.id)) {
```
ginderjeremiah: **Should-fix (latent footgun):** `.find` matches the first screen-anchored lesson for the screen *regardless of lock state*, then checks `isLocked` on only that one. If two `ScreenVisit` lessons target the same `screenKey`, this silently strips the second: on the first visit lesson A plays and is marked read; on every later visit `.find` still returns A (now read → `isLocked` false), so lesson B never triggers even though it's locked.

Note `evaluateMechanicTriggers` below iterates *all* lessons and handles this correctly — so the two evaluators diverge on the same "multiple lessons for one anchor" case. `ILesson` carries an `ordinal` and nothing constrains `screenKey` to be unique, so this is reachable content, not hypothetical.

Fold the lock check into the predicate so it picks the first *locked* match and stays consistent with the mechanic path:

```ts
const lesson = liveLessons().find(
    (candidate) =>
        candidate.triggerType === ELessonTriggerType.ScreenVisit &&
        candidate.screenKey === screenKey &&
        isLocked(candidate.id)
);
if (lesson) {
    openLesson(lesson);
}
```

If the content model deliberately guarantees one screen-anchored lesson per screen, a one-line comment stating that invariant here would be enough instead — but the defensive predicate is cheap and removes the divergence.

---

ginderjeremiah (commented):
## Review: Client-side tutorial trigger evaluation + unread-lesson flow (closes #1587)

**Verdict:** Approve — one non-blocking should-fix, otherwise a clean, well-tested landing of the client half of spike #1392's tutorial system.

_(Submitted as a `COMMENT` because I'm also the PR author, so GitHub won't accept a gating verdict from me — the intended call is Approve.)_

The change wires the two trigger kinds into the live game exactly as #1587 scopes them: screen-anchored lessons play immediately at the single screen-activation point, mechanic-anchored lessons queue as **unread** (never displayed at trigger time, since the player may be AFK), and the unread count surfaces as a persistent Help-nav badge distinct from the toast channel. Behavior matches the issue's acceptance criteria and the queue-always-for-mechanic v1 decision.

I traced the two reactivity paths that looked most likely to bite and both are sound:
- **Re-fire of an open screen tour** (e.g. a first-crit mechanic unlock mutates `playerManager.lessons` mid-tour, re-running the screen effect): `openLesson` re-runs but `open = !!activeLesson` stays `true`, so `TourPlayer`'s `stepIndex = 0` reset effect doesn't re-run and the player's position is preserved. `requestScreen` to the already-active screen is a harmless no-op.
- **No effect loop**: the screen-trigger effect and the navigation effect don't share tracked reads that would ping-pong, and once a lesson is marked read `isLocked` returns false so it can't re-open.

The `battleStep(...)` refactor to a `const activations` is safe — it returns a real `SkillActivation[]`, so `activations.length` and the double read (for-of + guard) are both fine; and the mechanic evaluation is correctly confined to the live loop, never the headless parity simulator, so there's no battle-parity surface here.

### Should-fix
- `evaluateScreenTrigger` — `tutorials.ts:46` — `.find` doesn't filter by lock state, so a second `ScreenVisit` lesson on the same `screenKey` is silently unreachable. Diverges from the mechanic evaluator, which iterates all lessons. Details + suggested one-line fix inline. Non-blocking (only manifests with multiple screen lessons per screen), but worth closing before it becomes a content author's confusing bug.

### Test coverage
Strong and matches the project's rules. `tutorials.test.ts` covers the full trigger matrix (first-activation, locked/unread/read gating, no re-fire once unlocked/read, retired exclusion, per-event specificity) with a stubbed player-manager that mirrors the real idempotent unlock/mark-read semantics rather than a hand-waved spy. `battle-engine.test.ts` asserts the live-loop wiring fires only on ticks with activations; `NavSidebar.test.ts` covers badge visibility/count and superseding the wip badge. The untested piece is the thin `$effect` in the game shell `+page.svelte` (the `welcome.phase === 'entered'` gate), but that's consistent with the repo's practice of not unit-testing the shell and pushing the real logic into `tutorials.ts` (which is covered).

### What's good
- The badge reuses the theme tokens (`--accent`, `--text-on-accent`) and mirrors the sibling `wip-badge`'s transition exactly, so reduced-motion is handled by the existing global rule — no hardcoded colors or unguarded motion.
- Clean separation: presentation-agnostic `content-events` detectors → `tutorials.ts` orchestration → a tiny reactive `tutorial-tour` store → shared `TourPlayer` host. `openLesson` is genuinely reusable by #1589's reading room.
- The judgment call flagged in the description (using `isLocked` absence as the "was this true before" flag) is correct here: `unlockLesson`/`markLessonRead` are idempotent, so an unlocked lesson naturally drops out of later ticks — no separate ever-seen flag needed.
- Docs updated well: the new `frontend.md` Tutorials section and the `content-design.md` status trim both reflect reality now that #1587 lands alongside the merged #1591/#1588/#1592.

---

ginderjeremiah:
## Code review — Tutorial trigger evaluation + unread lesson flow (#1587, spike #1392)

**Verdict: Approve with nits.** A clean, well-scoped client-side implementation. Not a spike result — a normal feature PR. Parity discipline is respected and clearly commented.

**CI: green** — all checks passed on `5b96f41` (`verify-frontend`, `test-frontend`, all three e2e browsers; backend checks correctly skipped for this frontend-only change).

### Verified against the code (not just the description)
- `battleStep` returns `SkillActivation[]`, so `const activations = battleStep(...)` + the `if (activations.length > 0)` guard is correct (not a generator that would break `.length`). ✅
- Mechanic evaluation is wired **only** into `BattleEngine.logicalUpdate` (the live loop), never the headless `BattleSimulator` — the parity surface is correctly untouched. ✅
- `playerManager` is `statify`'d, so the `$derived` badge count and the reactive screen-trigger `$effect` behave as intended; badge CSS tokens (`--accent`, `--text-on-accent`, `--mono`) all exist and are used semantically. ✅
- The self-flagged judgment call (passing literal `false` to the content-events detectors and letting `isLocked` supply the one-shot "ever-seen" gate) is sound — `unlockLesson`/`markLessonRead` are genuinely idempotent, so an unlocked lesson naturally drops out of consideration. ✅
- Docs (`docs/frontend.md` Tutorials section, `docs/content-design.md` caveats dropped) match the shipped behavior.

### Findings
- **[should-fix / discuss] Unread badge is invisible in the default collapsed rail.** `.unread-badge` has `opacity: 0` unless `.show` (bound to `expanded`), and the sidebar starts collapsed — so an idle/AFK player with a queued unread lesson sees nothing until they hover/pin the rail. Issue #1587 explicitly asks for a "persistent badge / HUD affordance," and unlike the `wip` badge (marks "not built yet"), an unread lesson is *actionable*. Either add a minimal always-visible collapsed-state dot, or explicitly defer the persistent affordance to the #1589 reading-room work — worth a deliberate decision rather than leaving it as an accident of mirroring the `wip` badge.
- **[nit] `evaluateScreenTrigger` uses `.find`** → one screen-visit lesson per `screenKey`, in `staticData.lessons` order. If two ever target the same screen, the second is silently unreachable (and, since screen lessons skip `unlockLesson`, never surfaces as unread). `ILesson.ordinal` exists — sort by it or document the one-per-screen constraint. Fine to defer.
- **[nit] Screen-trigger `$effect` re-runs on every `playerManager.lessons` mutation**, re-calling `openLesson(sameLesson)` during battle unlocks. Benign today (relies on `TourPlayer` only resetting on an `open` false→true transition + lesson reference stability) — noting for awareness, no change needed.
- **[nit] No test for the page-level `welcome.phase === 'entered'` gate** — the direct trigger evaluators and battle-engine wiring are covered; only the one-line glue guard is not. Low value, acceptable.

**Holding the merge** to the repo owner: the collapsed-rail badge visibility is the one item worth an explicit call (fix here vs. defer to #1589) before this goes in — the rest is solid.

---

UI/src/lib/engine/tutorials.ts
ginderjeremiah: Addressed in ab23285 — the `isLocked(candidate.id)` check is now folded into the `.find` predicate, so `evaluateScreenTrigger` picks the first *locked* screen-anchored lesson and stays consistent with `evaluateMechanicTriggers`. A regression test (`finds a second locked lesson on the same screen even when an earlier one is already read`) covers the multiple-lessons-per-screen case. Resolving.

---

ginderjeremiah (commented):
## Re-review: the one should-fix from the last round is addressed

**Verdict:** Approve — the sole should-fix (`evaluateScreenTrigger` not filtering by lock state) is resolved in ab23285 with a regression test; nothing new introduced.

_(Submitted as a `COMMENT` because I'm also the PR author, so GitHub won't accept a gating verdict from me — the intended call is Approve.)_

### Since the last review
- ✅ **`evaluateScreenTrigger` skipping a second locked lesson on the same screen** — addressed in ab23285. The `isLocked(candidate.id)` check is now folded into the `.find` predicate, so it picks the first *locked* `ScreenVisit` lesson and matches `evaluateMechanicTriggers`' multiple-lessons-per-anchor handling. Covered by the new `finds a second locked lesson on the same screen even when an earlier one is already read` test in `tutorials.test.ts`.

### New findings
- None. The delta is a single focused predicate change plus its test; I re-checked the mechanic path (`mechanicFired` passes `false` as the "before" flag and leans on `isLocked` gating for the "first ever" semantics — sound, since `unlockLesson`/`markLessonRead` are idempotent) and the content-events detector signatures still line up.

CI is green (frontend unit, verify, and all three e2e browsers pass; backend jobs correctly skipped — no backend changes). Clean to merge.

---

Issue #1601, opened 2026-07-04
Offline sim: mid-window level progression — grow the locked base and reward power across the away window


ginderjeremiah:
Resolves the design question in #1253 (owner decision, 2026-07-04): the offline simulation must be **accurate** — offline play must not be better *or* worse than the live idle loop — so everything that grows **automatically** in live play grows across the away window, while everything requiring player action stays frozen.

## Background

The #879 stationarity argument rested on "the player's level is not used in combat math at all," which the class locked base (#1223, `BaseAmount + AmountPerLevel × level`) ended. Worse, the frozen model **suspends the anti-grind treadmill offline**: the `DefeatRewards` difficulty multiplier, win rate, and battle speed all hold at their window-start values for up to 10 hours, while a live player's growing power would decay the multiplier and speed up kills. The distortion cuts both ways depending on matchup — exactly the "offline as optimal playstyle" shape we don't want. Full analysis in #1253.

## Scope

- Extract the level-up threshold/clamp loop from `Player.ApplyExp` (`Game.Core/Players/Player.cs:136-144`) into a shared helper used by both `Player` and the offline simulator. Semantics must not change — the curve is already pinned by `PlayerProgressionParityTests` (frontend parity via the codegen'd `ExpPerLevel`).
- `OfflineProgressSimulator`: track cumulative exp + current level in-loop. On each level-up (between battles — matching live, where exp lands on battle completion), re-derive the battler at the new level and re-materialize the `GetModifiersWithSignaturePassive` modifier set (currently built once at `OfflineProgressSimulator.cs:39`) so the locked base, signature passive, and `DefeatRewards` power measurement all grow — the treadmill multiplier decays offline exactly as live.
- `Player.GrantOfflineExp` stays the single authority for applying exp post-loop. Add a test pinning that the applied end state (level / exp remainder / stat points) matches the simulator's in-loop accounting.
- **Stays frozen (correctly):** stat allocations, gear, loadout — player-action state. Proficiency levels also stay frozen in this issue; un-freezing them is the committed follow-up (see #1253).
- Replace `Simulate_AppliesClassLockedBase_FromTheFrozenSnapshotLevel` (`OfflineProgressSimulatorTests.cs:415`) with growth tests: the locked base grows across simulated level-ups (the original #1223 acceptance criterion); a threshold matchup where mid-window leveling flips losses into wins; exp-multiplier decay as the player outgrows the zone; boss-mode growth.
[...]

PR #1620, opened 2026-07-05
Offline catch-up: carry the trailing remainder as a backdated next battle

### LINKED ISSUE #1596: Offline catch-up: carry the trailing remainder as a backdated next battle

ginderjeremiah:
## Summary

Implements #1596 (spike #1073) — the trailing edge of the offline catch-up simulation, mirroring #1595's fix to the leading edge.

`OfflineProgressSimulator.Simulate` always credits a whole battle+cooldown cycle once it starts one, so the budget-exhausting cycle typically runs past the real away-window boundary. Previously that overshoot was silently discarded — the live loop just started the next battle fresh at zero elapsed, so up to ~2 minutes of idle time evaporated on every welcome-back.

- **`OfflineProgressResult.RemainderMs`** — the simulator now reports that trailing overshoot (0 when nothing was simulated, or when the stalemate CPU-waste-guard cutoff stopped the loop early — that's genuine unspent budget, not an overshoot, and is left alone).
- **`OfflineProgressService`** serves the remainder against the post-battle cooldown first:
  - Within the cooldown → just sets `PlayerState`'s residual cooldown; the live `NewEnemy` cooldown gate naturally waits it out.
  - Spanning the cooldown → the excess becomes a fresh next battle's elapsed offset via a new `BattleService.SetupBackdatedBattle` helper (mirrors `StartBattle`/`StartBossBattle`'s enemy construction but sets `PlayerState.BattleStartTime` backdated by the excess instead of anchoring to now).
- The resulting active battle is surfaced on the welcome-back summary's `ActiveBattle` the same way a still-in-progress stale-battle hand-back (#1595) is, so the client can resume it via replay-to-offset (#1597) with no extra round trip. No new wire fields were needed — `BattleStartResult`/`EnemyInstance.ElapsedOffsetMs` already carry this generically.

## Test plan

- [x] Core unit tests (`OfflineProgressSimulatorTests`, `OfflineProgressResultTests`): remainder equals the overshoot past budget, remainder is zero when the budget divides evenly, zero when nothing was simulated, zero when the stalemate cutoff stops the loop early.
- [x] Application integration tests (`OfflineProgressServiceIntegrationTests`): remainder within cooldown → residual cooldown, no active battle; remainder spanning cooldown → backdated active battle + summary `ActiveBattle`; remainder exactly zero → neither.
- [x] `BattleServiceIntegrationTests`: new coverage for `SetupBackdatedBattle` (idle, boss, and zero-offset).
- [x] Updated the one existing test whose assertion assumed no battle could ever be left active after an offline pass (`SimulateOfflineProgress_StaleInFlightBattle_IsResolvedBeforeSimulating`) — that assumption is no longer true by design once a trailing-remainder battle can legitimately follow.
- [x] `dotnet test` — full suite green (Core 1316, Infrastructure 59, Application 883, Api 696).
- [x] Updated `docs/backend-battle.md`'s offline-rewards section.

Closes #1596

---
_Generated by [Claude Code](https://claude.ai/code/session_01T6ewXPWxEdaWyU8G8R21Af)_


Game.Application/Services/OfflineProgressService.cs
```diff
@@ -221,6 +230,34 @@ private OfflineSimulationParameters BuildSimulationParameters(
...
+
+            if (result.RemainderMs <= parameters.CooldownMs)
+            {
+                state.SetCooldown(now.AddMilliseconds(result.RemainderMs));
+                return null;
+            }
+
+            var elapsedOffsetMs = (int)(result.RemainderMs - parameters.CooldownMs);
+            return _battleService.SetupBackdatedBattle(
+                state, mode, zone, parameters.Snapshot, parameters.ResolveEnemy, now, elapsedOffsetMs);
```
ginderjeremiah: **Blocking — the remainder carried here double-counts the last battle when the boundary lands mid-battle.**

The crediting loop credits `ceil(away/step)` battles: it always simulates the battle the away boundary falls *inside* and credits it as a fully-completed outcome (see `OfflineProgressSimulatorTests.cs:81-83` — `4·step + 1` → **5** battles, and `Simulate_RemainderMs_EqualsOvershootPastTheAwayBudget` pins `RemainderMs = step - 1`). So `RemainderMs` is the overshoot of an *already-credited* final battle past the real boundary, not leftover unspent budget.

That splits into two regimes:

- `RemainderMs <= CooldownMs` (boundary in the last battle's post-win cooldown): the battle really did finish, and `SetCooldown(now + RemainderMs)` is correct. ✅
- `RemainderMs > CooldownMs` (boundary *inside* the last battle — the common case, since a battle is much longer than the cooldown): the last battle did **not** complete, yet the loop already credited it as a win. This branch then spawns a *fresh, additional* battle backdated by `RemainderMs - CooldownMs`. Two errors compound:
  1. **Phantom win** — one extra battle's exp/levels/challenge/proficiency progress is credited for a battle that only partially elapsed.
  2. **Inverted offset** — the true elapsed into the in-progress battle is `step - RemainderMs`, but this backdates by `battleDuration - (step - RemainderMs)`, the *complement*. A battle the player barely started at logout is handed back nearly finished (and vice versa).

The correct shape mirrors the leading-edge sibling (#1595): the battle straddling the boundary **is** the last simulated one — hand *it* back at its true elapsed offset (`step - RemainderMs`, which is `≤ battleDuration` in this regime) and do **not** credit it as a win; the residual-cooldown branch already handles the case where the boundary sits in a completed battle's cooldown. As written, a returning player who was ~1ms into a fresh battle is credited a phantom win and shown that battle nearly complete.

Worth calling out per CLAUDE.md's "verify the issue's claims": the issue frames this as `floor`-crediting that *drops* the remainder ("up to ~2 minutes evaporates"), but the real loop `ceil`-credits and *over*-counts — so the premise the fix was built on is inverted.

---

ginderjeremiah (commented):
## Review: Offline catch-up — carry the trailing remainder as a backdated next battle (closes #1596)

**Verdict: Request changes** (posted as a comment since I'm the PR author) — the remainder math double-counts the boundary battle whenever the away window ends mid-battle, over-crediting a phantom win and handing the resumed battle back at an inverted elapsed offset.

The scaffolding around this is genuinely clean: `SetupBackdatedBattle` faithfully mirrors `StartBattle`/`StartBossBattle`'s enemy construction, threads the offset through the existing `BattleStartResult`/`EnemyInstance.ElapsedOffsetMs` contract with no new wire fields, and reuses the #1595 hand-back surface. The `RemainderMs` plumbing on `OfflineProgressResult` and the `stoppedByStalemateCutoff` guard are both correct. But the value being carried forward is wrong.

### Blocking

- **The carried remainder double-counts the last credited battle** — `OfflineProgressService.cs:250-258` (details inline). The crediting loop credits `ceil(away/step)` battles, so the battle the away boundary lands *inside* is already counted as a completed win. `RemainderMs` is that battle's overshoot past the real boundary — not leftover budget. When the boundary falls inside the battle (the common case, since battle ≫ cooldown), `ApplyTrailingRemainder` both leaves the phantom win credited *and* spawns a fresh extra battle at the complement of the true elapsed offset. Net effect on a mid-battle return: one extra battle's rewards, plus a resume position that's inverted (barely-started → shown nearly-done). The residual-cooldown branch (`RemainderMs <= CooldownMs`) is correct; only the spanning branch is broken.

  The fix should follow the leading-edge sibling (#1595): the boundary battle *is* the last simulated one — hand it back at its true offset (`step - RemainderMs`) and don't credit it as a win. That likely means the simulator distinguishing "the last battle completed within budget" (credit it, carry only a residual cooldown) from "the last battle was still in progress at the boundary" (don't credit it, hand it back active at its real elapsed), rather than deriving both from a single `ceil`-loop overshoot.

### Test coverage

The new tests are thorough in structure but assert the *current* behavior rather than physical correctness, so they don't catch the above:

- `OfflineProgressSimulatorTests` bakes in the `ceil` count (`4·step + 1` → 5 battles) and `RemainderMs = step - 1`, i.e. the overshoot semantics themselves.
- `SimulateOfflineProgress_RemainderSpansCooldown_SetsUpBackdatedActiveBattle` asserts `BattlesWon = 100` (the `ceil` over-credit) and only that the backdated elapsed is loosely `InRange(1, battleMs + 2000)` — which passes for the inverted offset too.

Once the logic is corrected, these need to assert the invariant directly: a player away for `N·step + partialBattle` returns with exactly `N` credited wins and an active battle whose elapsed equals `partialBattle` (not its complement, and not `N+1` wins). The `RemainderWithinCooldown` and `RemainderExactlyZero` cases are correct and should stay.

### What's good

- No new wire fields — reusing `BattleStartResult.ElapsedOffsetMs` and surfacing on the summary's `ActiveBattle` exactly like the #1595 hand-back is the right call, and the `OfflineProgressModel`/`OfflineProgressSummary` doc updates describe the contract well.
- The `stoppedByStalemateCutoff` guard correctly excludes the CPU-waste cutoff from the remainder (a positive `remainingMs` there is unspent budget, not an overshoot), with focused unit coverage.
- The switch path was verified safe: `LoginController.cs:219` discards the returned summary, and the state mutation is discarded with the session — so the backdated battle never leaks onto a character being switched away from.
- `SetupBackdatedBattle`'s XML doc and the `backend-battle.md` addition are clear about intent.

Happy to be wrong here if I've mis-read the loop's termination — but as I read it, `while (remainingMs > 0)` credits the battle *before* decrementing, so the boundary battle is always fully credited, which is what makes the carried overshoot a double-count.

---

Game.Application/Services/OfflineProgressService.cs
ginderjeremiah: Confirmed — verified with concrete numbers (away=3000ms, battle=10000ms, cooldown=5000ms: the old loop credited a full win and computed a 7000ms backdate when the true elapsed was 3000ms). Fixed in 85f474c:

- The simulator now checks each candidate battle against the *remaining* budget before crediting it. A battle whose own duration fits is a real completed outcome (credited normally); one that doesn't fit is never credited — it's carried forward as `OfflineProgressResult.PendingBattle` (the exact enemy/seed already drawn and simulated, with its true elapsed-so-far offset, i.e. `remainingMs` at the point the attempt started).
- `RemainderMs` is now only ever the residual cooldown left after a genuinely completed battle (always ≤ `CooldownMs`), and is mutually exclusive with `PendingBattle`.
- `BattleService.SetupBackdatedBattle` (which built a fresh enemy) is replaced by `HandBackPendingBattle`, which hands back that exact pending battle instead of spawning a new one — matching the #1595 leading-edge shape you described.
- Rewrote the tests that baked in the old `ceil`-and-credit-regardless semantics (`Simulate_IdleLoop_RunsUntilBudgetExhausted_...`, the removed `Simulate_RemainderMs_EqualsOvershootPastTheAwayBudget`, the app-layer "spans cooldown" test) and added direct coverage for `PendingBattle` at both the Core and Application layers, including a test asserting the boundary battle is never both credited and pending.

Full suite green (Core 1320, Infrastructure 59, Application 883, Api 696), `dotnet format --verify-no-changes` clean.

---

ginderjeremiah (commented):
## Re-review: the double-count blocker from the first review is resolved

**Verdict: Approve** (posted as a comment since I'm the PR author) — the blocking finding from the first review is fully addressed, the code at HEAD (`92e9b16`) is correct, and CI is green. No new findings.

### Since the last review

- ✅ **Double-counted win + inverted elapsed offset at the boundary** (the first review's sole blocker) — fixed in `85f474c`. Verified against the current code, not just the commit message:
  - `OfflineProgressSimulator.Simulate` now checks each candidate battle against the *remaining* budget **before** crediting it (`if (result.TotalMs > remainingMs)`). A battle that fits is credited normally; one that doesn't is carried forward uncredited as `OfflineProgressResult.PendingBattle` with `ElapsedOffsetMs = (int)remainingMs` — the *true* elapsed-so-far, not the complement. This flips the canonical `4·step + 1` case from crediting 5 wins to crediting 4 + one pending, which is the physically correct answer.
  - `RemainderMs` is now only ever the residual cooldown after a genuinely completed battle (`Math.Max(0, -remainingMs)`, structurally `≤ CooldownMs` since a battle whose own duration didn't fit is never credited), and is mutually exclusive with `PendingBattle`.
  - `SetupBackdatedBattle` (which spawned a *fresh* enemy) is replaced by `HandBackPendingBattle`, which hands back the exact enemy/seed already simulated — matching the #1595 leading-edge shape. Verified `SetActiveBattle` signature, and that the enemy carries `BattleSkills` (idle via `SelectBattleSkills`, boss via `SelectAllBattleSkills`), so the persisted `ActiveEnemySkillIds` make the client replay deterministic.
- ✅ **Tests rewritten to assert the invariant, not the old semantics** — the simulator tests now assert `count = 4` + `PendingBattle` at the true offset; a dedicated `Simulate_PendingBattle_IsNeverAlsoCreditedAsAWin` pins the no-double-count guarantee; the app-layer `BoundaryFallsMidBattle` test asserts exactly `creditedBattles` wins and an active battle backdated by the true elapsed. The prior review thread is already resolved.
- ✅ **CI-flaky timing assertions** (`2762fe24`, `92e9b16`) — the mid-battle/residual-cooldown assertions were reworked from exact-equality into ranges bounded by before/after timestamps, with a slower dedicated battle scenario widening the margin around the intentional mid-battle boundary. Sound: the service reads its own `now` internally, so exact equality against a test-side anchor was never safe.

### Verified correct at HEAD

- **`now` consistency** — offset/backdate math and the away budget both derive from the single `now` captured in `SimulateProgress`; `StampActivity(now)` re-anchors to the same instant.
- **State is persisted** — `GetOfflineProgress` calls `SavePlayerState()` unconditionally after the sim, so the login-path trailing-remainder mutation (cooldown or active battle) durably lands. The switch path (`LoginController.cs:219`) intentionally discards the departed character's `PlayerState`, so the mutation is dropped there — no regression vs. prior behavior, and documented.
- **int cast is safe** — a pending battle only occurs when `remainingMs < result.TotalMs ≤ 2-minute battle cap`, well within `int`.

### Note (not blocking)

- This PR deliberately diverges from #1596's stated "no changes to the crediting loop's internals" — the issue assumed you could tack on a fresh next battle without touching the loop, which is precisely what produced the double-count. Changing the loop to credit only battles that fit the remaining budget is the correct call, so the deviation is an improvement, not a defect.

### What's good

- Reusing `BattleStartResult.ElapsedOffsetMs` / the summary's `ActiveBattle` with no new wire fields, mirroring the #1595 hand-back, keeps the client contract uniform.
- `PendingBattle` vs `RemainderMs` as mutually-exclusive carriers is a clean way to distinguish "boundary inside a battle" from "boundary inside a completed battle's cooldown", and both the XML docs and the `backend-battle.md` addition describe the contract clearly.

---

ginderjeremiah:
Resolved the merge conflict with `main`. #1622 (the offline-sim level-growth work) landed first and touched the same simulator files, so I merged `main` into this branch and reconciled the two additive changes:

- **`OfflineProgressResult`** — kept both feature sets: `RemainderMs` + `PendingBattle` (#1596) and `EndingLevel` + `EndingExp` (#1601). The constructor now takes all four (`remainderMs`, `pendingBattle`, `endingLevel`, `endingExp`), all optional.
- **`OfflineProgressSimulator.Simulate`** — the two changes compose cleanly: the per-iteration seed draw + "does this battle fit the remaining budget? if not, carry it forward uncredited as a `PendingBattle`" gate (#1596) runs *before* the in-loop exp/level growth crediting (#1601). `SimulateBattle` now takes both the mid-window-grown `workingSnapshot` and the caller-drawn `seed`. The return threads remainder/pending + ending level/exp.

**One test needed updating** for the cross-PR interaction: `Simulate_NonZeroStartingExp_CarriesIntoTheFirstLevelThreshold` (from #1622) sized its budget with `awayMs: 1` on the old assumption that any started battle is credited. Under this PR's trailing-remainder semantics, a battle whose own duration exceeds the remaining budget is now carried forward as an uncredited `PendingBattle` instead — so `awayMs: 1` yields zero credited wins, not one. I resized it to probe the deterministic (fixed-seed) per-battle duration and budget for exactly one completed battle, preserving the test's original intent (that banked `StartingExp` carries a single win's reward across the level threshold).

Verified locally: `dotnet build` clean, `Game.Core.Tests` (1325) and `Game.Application.Tests` (884) green, `dotnet format --verify-no-changes` clean. Waiting on CI to re-run before merging.

---

ginderjeremiah:
CI is now green. The first `test-backend` run failed the coverage gate on `Game.DataAccess` (92.9% < 93% floor), but I verified this is **pre-existing boundary flakiness in `main`, not caused by this merge**:

- `Game.DataAccess` line coverage is **5024/5402 = 93.0026%** on both `main` (`0d0bbda`) and this merge commit — the merge is coverage-neutral to that assembly (it only touches `Game.Core` offline files).
- The floor is seeded at exactly the current value, leaving **under one line of margin**. `Game.DataAccess` is partly exercised via DB integration tests with ~1 line of run-to-run nondeterminism, so any run can dip to 5023 (→ 92.9%) and fail — `main`'s own CI only passes on runs that land at 5024.

Re-running the failed jobs (identical tree) came back green, confirming the flake. Filed **#1627** (tech debt) to give the `Game.DataAccess` floor real margin / fix the ratchet's zero-margin seeding so this stops randomly blocking unrelated PRs.

Merging now.

---

Issue #1602, opened 2026-07-04
Offline sim: mid-window proficiency growth — run accrual in-loop and move the math into Game.Core


ginderjeremiah:
**Depends on:** #1601. Part of the #1253 decision (offline must model everything that grows automatically in live play).

In live play, proficiency levels and their attribute payouts grow battle-to-battle — battle N's accrual applies to battle N+1's battler automatically. The offline path accrues proficiency XP per battle but only **after** the simulation (`OfflineProgressService.ApplyOfflineRewards`, `Game.Application/Services/OfflineProgressService.cs:254`), so the whole window fights with window-start proficiency bonuses and none of the mid-window milestone attribute payouts. Left as-is after #1601, this is a residual, smaller version of the same distortion: defensive/resist accrual in particular reads a power denominator that never grows.

## Scope

- Move the proficiency accrual domain math out of `ProficiencyRewardService` (`Game.Application/Services/ProficiencyRewardService.cs`) into `Game.Core` — it is domain logic and belongs there regardless (owner-agreed in #1253). Swap its `IProficiencies`/`ISkills` repo dependencies for the resolver-func pattern the offline simulator already uses (`ResolveProficiency` etc.); the application service becomes a thin adapter for the live `BattleStatisticsEventHandler` path (persistence orchestration and client push/notify stay application-side).
- Run accrual in-loop in `OfflineProgressSimulator`: after each victory, accrue with the **current** power, resolve frontier routing / level-ups / milestone payouts, and feed the updated proficiency levels into the next battle's battler assembly (alongside #1601's level growth).
- Milestone skill *grants* still don't auto-slot — the loadout stays frozen (player-action state). Milestone and per-level attribute bonuses take effect from the next battle, matching live.
- Persistence and notification stay consolidated at end-of-window as today; the applied end state must match the in-loop accounting (pin with a test, same shape as #1601's exp pin).

## Tests

- Same-sequence parity: the in-loop offline accrual and the live per-battle path produce identical proficiency state for an identical battle sequence.
- Mid-window proficiency growth changes later battle outcomes (a milestone attribute payout flips a threshold matchup).
- Defensive/resist accrual reflects the growing power denominator rather than window-start power.

Related: #1253, #1601, #982, #1318.

PR #1622, opened 2026-07-05
Offline sim: grow the locked base and reward power across the away window

### LINKED ISSUE #1601: Offline sim: mid-window level progression — grow the locked base and reward power across the away window

ginderjeremiah:
## Summary

Implements #1601 (owner decision in #1253, closed): the offline simulation must be **accurate** — offline play must not be better *or* worse than the live idle loop — so everything that grows automatically in live play grows across the away window, while everything requiring player action stays frozen.

- **`Game.Core.Players.ExpProgression`** — a new shared static helper extracting the level-up threshold/clamp loop from `Player.ApplyExp` (`level * GameConstants.ExpPerLevel` thresholds, clamped to `ServerGameConstants.MaxExpPerGrant`). `Player.ApplyExp` now delegates to it; semantics are unchanged (pinned by the existing `PlayerProgressionParityTests`/`PlayerTests`).
- **`OfflineProgressSimulator`** now tracks cumulative exp and level in-loop (`OfflineSimulationParameters.StartingExp` + the snapshot's captured level), via a private mutable-level clone of the caller's `BattleSnapshot`. On each victory, exp is applied through `ExpProgression.ApplyExp` — between battles, matching when it lands live — and a level crossed mid-window re-derives the battler `DefeatRewards` measures, so the class locked base, signature passive, and reward power all grow, and the anti-grind treadmill's difficulty multiplier decays offline exactly as it would live.
- **`Player.GrantOfflineExp` stays the single authority** for actually applying the window's exp to the aggregate, post-loop. `OfflineProgressResult.EndingLevel`/`EndingExp` expose the simulator's own in-loop accounting, pinned by a new test to agree exactly with what `GrantOfflineExp` produces from the same reward sequence.
- **Stays frozen (correctly):** stat allocations, gear, loadout — player-action state. Proficiency levels also stay frozen in this issue; un-freezing them (running their accrual in-loop too) is the committed follow-up, #1602.
- Replaced `Simulate_AppliesClassLockedBase_FromTheFrozenSnapshotLevel` (no longer true) with growth tests: the locked base/rating growing across simulated level-ups, a threshold matchup where mid-window leveling flips losses into wins, difficulty-multiplier decay as the player outgrows the zone, and boss-mode growth.
- Docs: corrected `docs/game-design.md`'s offline-rewards stationarity rationale and the proficiency-accrual section's framing, and updated `docs/backend-battle.md`'s offline section — plus the stationarity comments in the simulator sources (`OfflineProgressSimulator`, `OfflineSimulationParameters`, `OfflineBattleOutcome`, `OfflineProgressService`).

## Test plan

- [x] New/updated Core unit tests (`OfflineProgressSimulatorTests`): mid-window level-ups grow the locked base and measured rating; a threshold matchup flips from losses to wins as the player levels; the difficulty ratio decays as the player outgrows a zone-static enemy; boss-mode growth; the simulator's `EndingLevel`/`EndingExp` match what `Player.GrantOfflineExp` applies from the identical reward sequence.
- [x] `dotnet test` — full suite green (Core 1313, Infrastructure 59, Application 877, Api 696), including the existing `OfflineProgressServiceIntegrationTests` (unaffected, since those fixtures' classes carry no level-scaled locked base).

Closes #1601

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wa2FyZDA4FGr4ncLTYZmTk)_


ginderjeremiah (commented):
## Review: Offline sim mid-window level growth (#1601)

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't accept a gating approval from me; there are no blockers)* — the change is correct, cleanly layered, and well-tested, with one non-blocking test-coverage gap worth closing.

I reviewed the exp-application logic closely since that's where a divergence between the live and offline paths would hide, and it holds up:

- **`ExpProgression` extraction is semantics-preserving.** `Player.ApplyExp` now delegates but keeps identical behavior: the per-level `PlayerLeveledUpEvent` carries `startLevel + levelsIn`, which equals the original post-increment `Level`, and `StatPointsGained` still accumulates in the same order before each event fires. The clamp (`[0, MaxExpPerGrant]`) and `level * ExpPerLevel` threshold loop are moved verbatim.
- **Simulator order of operations mirrors live.** The winning battle's own `DefeatRewards` is measured from the pre-level-up rating battler, *then* exp is applied between battles and the battler rebuilt for subsequent ones — so a mid-window level-up grows the locked base/signature passive/reward power for the *next* fight, not the one that earned it. Both `SimulateBattle` and `playerBattlerForRating` derive from the same `workingSnapshot.Level`, so the fight and its rating always agree.
- **Single authority preserved.** `Player.GrantOfflineExp` still applies the window's exp to the aggregate; the stored `OfflineBattleOutcome.ExpReward` is the raw reward, clamped identically by both the in-loop `ExpProgression.ApplyExp` and `GrantOfflineExp`, so `EndingLevel`/`EndingExp` genuinely pin the two paths together.
- **Layering is clean** — `ExpProgression` and the simulator stay in `Game.Core`; `OfflineProgressService` only sets `StartingExp = player.Exp`. And this is correctly backend-only: `docs/backend-battle.md` establishes the offline sim as parity-irrelevant (no client replay), so there's no frontend mirror to add.
- `BattleSnapshot.Level` was already `{ get; set; }` on a mutable class, so the mutable-level clone introduces no new immutability regression; sharing every other field by reference is fine since only `Level` changes.

### Should-fix

- **`StartingExp` is never exercised with a non-zero value.** Every growth and parity test starts from `exp = 0` on both sides, so the parity test would still pass even if the simulator ignored `StartingExp` and always started from 0 — the one behavior the field exists for ("a player already partway to their next level doesn't get an extra free battle's grace") isn't pinned. This is a test that belongs in the PR, not a follow-up: add a case seeding a non-zero `StartingExp` where a small first reward crosses a threshold *only* because of the carried-in exp, and start the `GrantOfflineExp` parity player from the same non-zero exp so they must still agree.

### Nits

- `BattleSnapshot.ProficiencyLevels`' XML doc still reads "the offline window fights at a stationary power (spike #982 decision 7)." That's now only true for the still-frozen inputs (proficiency levels, gear, allocations), not level — the claim is technically fine scoped to proficiencies, but it's the one spot the PR's otherwise-thorough stationarity reframing missed. Minor wording tweak.

### Test coverage

Solid: locked-base growth raising measured rating, a threshold matchup flipping losses→wins one-way, difficulty-multiplier decay as the player outgrows a zone-static enemy, boss-mode growth, and the `EndingLevel`/`EndingExp` ↔ `GrantOfflineExp` parity pin. The `FreeFarmEnemy` fixture (never lands a hit, still mints exp via continuous-DPS rating) is a clean way to isolate growth from win/loss confounds. The only gap is the non-zero `StartingExp` case above.

### What's good

Careful, single-authority design that keeps the live and offline exp math on one shared loop; the comments explain the *why* (order of operations, why the rating battler is safe to reuse between level-ups) without over-narrating; and the docs (`game-design.md`, `backend-battle.md`, and the source stationarity comments) were all updated to match the new "grows what grows automatically, freezes what needs player action" model rather than left stale.


---

ginderjeremiah (commented):
## Re-review: `edee79f5` addresses both findings from the last round

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't accept a gating approval from me)* — both items from the previous review are resolved, and the small delta (a test + a doc comment) introduces no new concerns. CI is green.

### Since the last review

- ✅ **Should-fix: `StartingExp` never exercised non-zero** — addressed by the new `Simulate_NonZeroStartingExp_CarriesIntoTheFirstLevelThreshold` test (`OfflineProgressSimulatorTests.cs`). It nails exactly the behavior the field exists for: an identical single win crosses the level-1 threshold *only* when pre-window exp is banked (`startingExp: 100 - firstWinExp` → level 2, exp 0) and not from a fresh 0 (stays level 1). It anchors `firstWinExp` from the from-zero run rather than hardcoding it, and holds the `GrantOfflineExp` parity pin from the same non-zero starting exp — so the field can no longer be silently ignored without a failure.
- ✅ **Nit: stale `BattleSnapshot.ProficiencyLevels` stationarity doc** — corrected. It now distinguishes the still-frozen proficiency levels from the mid-window-growing `Level` (`#1601`), and points at the `#1602` un-freeze follow-up, instead of describing the whole snapshot as stationary.

### New findings

None. The delta is test-and-docs only; the growth/exp logic reviewed last round is unchanged.

### What's good

The added test is self-consistent (derives `firstWinExp` from the from-zero run instead of assuming it) and isolates the carried-in-exp effect cleanly via the `FreeFarmEnemy` fixture. Single-authority exp math (shared `ExpProgression` loop, `GrantOfflineExp` as the post-loop authority) remains intact.

---

PR #1624, opened 2026-07-05
Workbench: add a text column type; wire lesson tour steps (#1604)

### LINKED ISSUE #1604: Workbench: editable table columns need a text-input cell type (blocks lesson-step authoring)
### Context

#1591 built the `Lesson`/`LessonStep` reference-data entity, its cache, admin CRUD (`IAdminLessons.SetSteps`), and the content export/seed/lint plumbing. The Workbench `EntityConfig` for lessons (`UI/src/routes/admin/workbench/entities/lesson.ts`) only covers the identity fields (key, name, trigger, host screen, ordinal, designer notes) — the ordered tour **steps** (`{ ordinal, text, anchorKey }`) have no Workbench editor yet, even though the backend `SetLessonSteps` endpoint is fully built and tested.

### Why it wasn't done in #1591

The generic Workbench `table` section (`TableSectionConfig`/`TableCell.svelte`/`SectionTable.svelte`) is typed and rendered for **numeric-only** row values: `ColumnConfig.type` is `'select' | 'attribute' | 'number' | 'share'`, `TableCell`'s `onChange` takes `(value: number) => void`, and every row is `Record<string, number>`. A `LessonStep`'s `text` (the tutorial callout copy) and optional `anchorKey` are free-form strings, which this shared, widely-reused component doesn't support. Retrofitting it (e.g. `Record<string, number | string>` throughout `TableSectionConfig`, `TableCell`, the diff helpers) is a real framework change touching every existing table section (skill effects, damage portions, zone enemies, item mod slots, …) — too risky to bundle into a content-plumbing PR.

### Ask

Add a `'text'` (and maybe `'textarea'` for longer copy) column type to the generic table-section framework:
- `ColumnConfig` gains the new `type` variant.
- `TableCell.svelte` renders a text input (or textarea) for it, with the same dirty-dot styling as the other cell types.
- The row-value type across `TableSectionConfig`/`TableCell`/`SectionTable` widens from `Record<string, number>` to `Record<string, number | string>` (or similar), audited against every existing table section to confirm none break.

Then wire `LessonStep`'s `text`/`anchorKey` into `lesson.ts`'s `sections` as a `table` section (`itemsKey: 'steps'`, `rowKey: 'ordinal'`), with a `childSavers` entry calling `AdminTools/SetLessonSteps`. Write the ordinal-keyed diff helper (mirroring `damagePortionChanges` in `UI/src/routes/admin/workbench/save-helpers.ts`, keyed by `ordinal` instead of `type`) alongside this section — an earlier draft added it ahead of its consumer and it was removed as dead code per review, so add it together with the section that calls it.

### Notes

- The backend surface (`IAdminLessons.SetSteps`, `SetLessonStepsData`, the `ChildCollectionReconciler`-based reconcile) is complete and integration-tested; this is frontend-only.
- Until this lands, lesson steps can only be authored via a direct `POST AdminTools/SetLessonSteps` call (or a future content-JSON hand-seed), not through the Workbench UI.

ginderjeremiah:
## Summary

Implements #1604: the generic Workbench `table` section (`TableSectionConfig`/`TableCell.svelte`/`SectionTable.svelte`) was numeric-cell-only, so a `LessonStep`'s free-text tour copy (`text`/`anchorKey`) had no Workbench editor even though the backend `SetLessonSteps` endpoint (from #1591) was already built and integration-tested.

- **`ColumnConfig` gains a `'text'` type.** `TableCell.svelte` renders a plain text `<input>` for it, with the same dirty-dot styling as the number/select cell types.
- **The row-value type widens** from `Record<string, number>` to `Record<string, number | string>` across `TableSectionConfig`/`TableCell`/`SectionTable` (and `reference.svelte.ts`'s `enemySpawnShareTotal`, which types against the same shape). Audited against every existing table section (skill effects/portions/multipliers, zone spawns, item mod slots, class equipment/attributes, skill recipe conditions) — none break, confirmed by the full existing test suite staying green.
- **`lesson.ts` gets a new "Tour Steps" table section** (`itemsKey: 'steps'`, `rowKey: 'ordinal'`) with `Step` (number), `Callout Text` (text), and `Anchor Key` (text) columns, plus a `warn` when a lesson has no steps yet. A `childSaver` posts the step set to `AdminTools/SetLessonSteps` when changed, normalising a blank anchor key back to `undefined` for the wire.

## Judgment call

The issue asked for an ordinal-keyed diff helper mirroring `damagePortionChanges` (Add/Edit/Delete changes). I didn't add one: `AdminTools/SetLessonSteps` takes the **full desired step set** reconciled server-side by `ChildCollectionReconciler` (keyed by `Ordinal`), not an Add/Edit/Delete diff — matching `SetZoneEnemies`'s full-list contract, not `SetSkillPortions`'s diff contract. The childSaver follows the `SetZoneEnemies` pattern instead (`childChanged` gates the call, then the full `record.steps` is sent). Writing a diff-changes helper here would have no consumer, reproducing exactly the "added ahead of its consumer, removed as dead code" problem the issue itself flags for an earlier draft.

## Verification

- `dotnet build`/tests untouched — this is frontend-only, per the issue.
- `npm run lint` (ESLint + svelte-check) — clean.
- `npm run test:unit:ci` — full suite green (3449/3449), including new coverage in `TableCell.test.ts` (text column renders, binds, fires `onChange`, dirty styling) and `lesson.test.ts` (steps section `warn`/`newRow`, and the `SetLessonSteps` childSaver posting the full normalised step set only when steps changed).
- `npm run build` — production build succeeds.
- **End-to-end, live**: ran the API natively against fresh Postgres/Redis containers plus the SvelteKit dev server, signed up as an e2e admin, opened the Lessons entity in the Workbench, added a tour step with callout text and an anchor key through the new table, saved, and confirmed both columns round-tripped correctly into the `LessonSteps` table.

## Test plan

- [x] Unit tests for the new `text` column type (render, value binding, `onChange`, dirty state)
- [x] Unit tests for the lesson steps section (`warn`, `newRow` ordinal sequencing) and its `SetLessonSteps` childSaver (posts on change, normalises blank anchor key, skips when unchanged)
- [x] Full frontend test suite green
- [x] Lint/typecheck clean
- [x] Production build succeeds
- [x] Manually verified in a live dev stack (signup → admin → Lessons → add step → save → confirmed in Postgres)

Closes #1604

---
_Generated by [Claude Code](https://claude.ai/code/session_01D8tvhd2hbbv6qAWxrxCgyn)_


UI/src/routes/admin/workbench/entities/lesson.ts
```diff
@@ -109,6 +107,43 @@ export const lessonEntity: EntityConfig<WorkbenchLesson> = {
...
+			emptyIcon: 'map',
+			emptyTitle: 'No tour steps',
+			emptySub: 'This lesson’s tour has nothing to show yet.',
+			newRow: (l) => ({
+				ordinal: l.steps.length ? Math.max(...l.steps.map((s) => s.ordinal)) + 1 : 0,
+				text: '',
+				anchorKey: ''
+			}),
+			columns: [
+				{ key: 'ordinal', label: 'Step', type: 'number', align: 'r', width: 90 },
```
ginderjeremiah, line 147: **Should-fix — editable `ordinal` is the row identity, with no uniqueness guard.** `ordinal` is simultaneously the section's `rowKey` (so it keys `SectionTable`'s `{#each rows (row[rowKey])}` and `baseByKey`) *and* an author-editable `number` column here. Nothing stops an author typing two steps to the same ordinal, and the consequences are real on both ends:

- **Frontend:** a Svelte 5 keyed `{#each}` throws at runtime on duplicate keys, so a collision crashes the detail panel while editing.
- **Backend:** `AdminLessons.SetSteps` reconciles via `ChildCollectionReconciler` keyed by `Ordinal` (`existingKey`/`desiredKey` both `s => s.Ordinal`), so two rows sharing an ordinal collapse/collide there too.

This is the first table section to use a plain editable numeric column as its `rowKey` — every other section keys off a `unique` select/attribute (zone enemies, skill portions, class attributes) or an auto-assigned surrogate `id` (mod slots, skill effects), none of which an author can collide by hand. The `newRow` next-ordinal logic keeps the *default* flow safe, so only manual editing trips it — but it's trivially reachable. Suggest guarding it: at minimum a section `warn` for duplicate ordinals so the validation triangle surfaces it before save.

---

UI/src/routes/admin/workbench/entities/lesson.ts
```diff
@@ -109,6 +107,43 @@ export const lessonEntity: EntityConfig<WorkbenchLesson> = {
...
 				}
 			]
+		},
+		{
+			key: 'steps',
+			label: 'Tour Steps',
+			glyph: 'map',
+			desc: 'The ordered coach-mark callouts this lesson’s tour plays',
+			count: (l) => l.steps.length,
+			warn: (l) => (l.steps.length ? null : 'No tour steps'),
```
ginderjeremiah: **Nit — no validation that a step's callout `text` is non-empty.** The `warn` only fires when the lesson has *zero* steps; a step with a blank `Callout Text` saves fine (`ToStepEntity` sets `Text = step.Text` with no guard) and renders an empty coach-mark. Since a text column can't declare `required` the way a field can, this section `warn` is the natural place to also catch `l.steps.some((s) => !s.text.trim())`. Cheap to fold in alongside the duplicate-ordinal check.

---

ginderjeremiah (commented):
## Review: Workbench text column type + lesson tour-step authoring (#1604)

**Intended verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating review from me) — the change does what #1604 asked, is well-tested, and the one design departure from the issue is correct. Two non-blocking items below, plus a CI note that isn't this PR's fault.

The PR adds a `text` column type to the generic Workbench table framework, widens the row-value type from `Record<string, number>` to `Record<string, number | string>` across `TableSectionConfig`/`TableCell`/`SectionTable` (and `reference.svelte.ts`), and wires a "Tour Steps" table section into `lesson.ts` with a `childSaver` posting the full step set to `AdminTools/SetLessonSteps`. That matches the issue's ask.

### Should-fix
- **`ordinal` is the row identity *and* a freely-editable numeric column, with no uniqueness guard** — `lesson.ts:131`. A hand-entered duplicate ordinal crashes the detail panel (Svelte keyed `{#each}` rejects duplicate keys) and collides in the backend `ChildCollectionReconciler` (keyed by `Ordinal`). First section to key off an editable plain number rather than a `unique` select or surrogate id. Details inline — a section `warn` for duplicate ordinals is the cheap fix.

### Nits
- **No validation that a step's callout `text` is non-empty** — `lesson.ts:117`. A blank step saves and renders an empty coach-mark; fold a `steps.some((s) => !s.text.trim())` check into the section `warn`.

### Good judgment
- **Skipping the ordinal-keyed diff helper the issue asked for is the right call, and I verified it.** `AdminLessons.SetSteps` reconciles the *full* desired set via `ChildCollectionReconciler.Reconcile(existingKey: s => s.Ordinal, desiredKey: s => s.Ordinal, …)` — the `SetZoneEnemies` full-set contract, not the `SetSkillPortions` diff contract. A `damagePortionChanges`-style helper would have no consumer, reproducing exactly the "added ahead of its consumer, removed as dead code" problem the issue itself flags. Following `SetZoneEnemies` instead is correct.
- Type-widening audit holds up: the `Number(...)` coercions on share/weight columns are safe no-ops for the numeric weights those columns carry, and the `as number` casts land only on columns whose values are always numeric ids. Removing the now-stale "steps aren't editable yet" comment is good hygiene.

### Test coverage
Good. `TableCell.test.ts` covers the new text cell (render, empty-value fallback, `onChange` string, dirty styling); `lesson.test.ts` covers the steps section (`warn`, `newRow` ordinal sequencing) and the childSaver (posts the full set, normalises blank `anchorKey` → absent, skips when unchanged). The only gap is a test for the duplicate-ordinal guard once it's added.

### ⚠️ CI note — the `verify-frontend` red is a pre-existing `main` breakage, not this PR
`verify-frontend` fails on `svelte-check`, but the two errors are in `src/tests/lib/engine/battle/battle-engine.test.ts:510` / `:521` — a file this PR doesn't touch. On `main` (`0d0bbda`), codegen made `IEnemyInstance.enemyRating` a required field and most mocks in that test were updated to set `enemyRating: 100`, but the two `evaluateMechanicTriggers` mocks were missed. So `verify-frontend` is red on `main` itself; this PR just inherits it through the CI merge. It needs fixing on `main` (add `enemyRating` to those two mocks), after which a rebase here should go green. I'll file a bug for that separately. **Not merging** while CI is red and the should-fix is open.

---

ginderjeremiah:
Addressed both review items:

- **Should-fix (duplicate ordinals)**: the steps `warn` now also flags two steps sharing an ordinal.
- **Nit (blank callout text)**: `warn` now also flags a step with blank/whitespace-only text.
- Added tests for both, plus a test for the `count` callback (closing the coverage gap that tripped `src/routes/**`'s functions threshold in CI).
- Fixed the pre-existing `main` regression you flagged (`battle-engine.test.ts` mocks missing the now-required `enemyRating`) by merging `main` and patching the two missed mocks, so this PR's merge carries the fix forward too.

CI was pending as of the last push (`79cf727`).

---

PR #1649, opened 2026-07-05
BattleLost: add elapsed-time anti-cheat gate (#1630)

### LINKED ISSUE #1630: BattleLost has no elapsed-time anti-cheat: a loss can be claimed instantly, farming statistics/challenges at an accelerated rate
## Problem

The three battle-end paths gate timing differently, and the loss path is the only one with **no** timing control:

- `EndBattleVictory` rejects a claim before `BattleStartTime + replayMs − 1 tick` of *server* time has elapsed (`Game.Application/Services/BattleService.cs:287-296`).
- `AbandonBattle` caps its re-simulation at server-measured wall-clock elapsed time.
- `EndBattleLoss` (`BattleService.cs:319-340`, reached from `Game.Api/Sockets/Commands/BattleLost.cs`) replays the battle to its conclusion (`TryResolveActiveBattle` with no max) and, if the replay is a loss, books it **immediately** — a tampered client can send `BattleLost` the instant the battle starts.

## Exploit

The full simulated battle's statistics are recorded regardless of timing — `PlayerProgress.RecordBattleCompleted` books `DamageDealt` (global and per-skill), `DamageTaken`, `DamageHealed`, crit/dodge/parry tallies, `CounterDamageDealt`, `TotalBattleTime`, `EnemiesEncountered`, `BattlesLost`, `PlayerDeaths` — and `BattleStatisticsEventHandler` then evaluates every challenge those statistics track. Pick a slow losing matchup (e.g. a tanky build that survives ~120s while dealing damage), claim the loss at t=0, eat the 5s post-battle cooldown, repeat: statistic/challenge accrual runs at up to ~24× the legitimate rate. This violates the documented rule that every player-mutating socket command validates as anti-cheat.

## Fix direction

Mirror the victory gate (reject when `now < BattleStartTime + result.TotalMs − tolerance`), or replay capped at server-elapsed wall-clock like `AbandonBattle` and only book a loss when the player actually died within the observed window (handing back "still in progress" per the #1595 shape otherwise). The two options interact with the #1595 hand-back model, so pick deliberately.

Note: `EndBattleLoss_ValidLoss_ReturnsTrueAndClearsState` (`Game.Application.Tests/Services/BattleServiceIntegrationTests.cs`) claims the loss immediately after `StartBattle` — it pins the lenient behavior incidentally and needs a backdated `BattleStartTime` under the fix.

## Tests

Integration tests mirroring the victory gate's rejection test: a too-early `BattleLost` claim is refused with no statistics recorded; a legitimately-timed loss still books.

ginderjeremiah:
## Summary

`EndBattleVictory` and `AbandonBattle` both gate their outcome on server-measured elapsed time, but `EndBattleLoss` (reached from the `BattleLost` socket command) didn't — it replayed the battle to conclusion and booked the loss's full statistics (damage, crits, kills, `TotalBattleTime`, …) immediately, with no check that real server time had actually passed. A tampered client could pick a slow losing matchup, claim `BattleLost` the instant the battle started, and repeat — farming statistic/challenge accrual at an accelerated rate.

- Extracted the victory path's elapsed-time check (`now < battleStart + replayMs - tolerance`) into a shared `ClaimedBeforeBattleCouldFinish` helper in `Game.Application/Services/BattleService.cs`, and applied the same gate to `EndBattleLoss`. `TryResolveActiveBattle` already fully replays the loss to conclusion (capped at `DefaultMaxBattleMs`), so `result.TotalMs` is the real duration the claimed outcome took — the same shape of check the victory path already runs.
- Backdated `BattleServiceIntegrationTests.EndBattleLoss_ValidLoss_ReturnsTrueAndClearsState`'s battle start (per the issue's note that this test incidentally pinned the lenient behavior) and added `EndBattleLoss_ClaimedBeforeBattleCouldFinish_ReturnsFalse`.
- Added the socket-level mirror in `BattleLostSocketTests`: backdated `BattleLost_ValidLoss_Succeeds`'s battle start and added `BattleLost_ClaimedBeforeBattleCouldFinish_ReturnsError`.

### Design note (flagging per issue's "pick deliberately")

The issue offered two fix directions: mirror the victory gate, or cap the replay itself at wall-clock elapsed time (like `AbandonBattle`) and hand back "still in progress" per the #1595 shape when it hasn't concluded. I went with the first — it's a much smaller, lower-risk change that reuses the exact pattern already proven in `EndBattleVictory`, and deliberately avoids entangling the `BattleLost` claim path with the `#1595`/`#1596` hand-back model, which the issue itself flagged as an interaction to be careful about. The tradeoff: a genuinely-drawn battle (both alive at the `DefaultMaxBattleMs` cap) now requires waiting out the full ~2 minutes before `BattleLost` will accept it, same as a real loss would — this seems like the correct anti-cheat behavior rather than a regression, but flagging it since it wasn't explicitly a design goal.

## Test plan

- [x] `dotnet build` — clean.
- [x] `dotnet test --no-build --no-restore` (full backend suite: `Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Api.Tests`, `Game.Application.Tests`) — all passing (2,970 tests, 0 failures).
- [x] New/updated tests specifically cover: a too-early `BattleLost`/`EndBattleLoss` claim is rejected with the battle left active and no state persisted; a legitimately-timed loss still books, at both the service and socket layers.

Closes #1630

---
_Generated by [Claude Code](https://claude.ai/code/session_01K2JaoHB7T67XDn2U26NQ3i)_


Game.Api.Tests/Integration/BattleLostSocketTests.cs
```diff
@@ -109,15 +121,40 @@ public async Task BattleLost_ValidLoss_Succeeds()
...
+                state.SetActiveBattle(
+                    state.ActiveEnemyId!.Value,
+                    state.ActiveEnemyLevel!.Value,
+                    state.ActiveEnemySkillIds!,
+                    state.BattleSeed!.Value,
+                    startTime: DateTime.UtcNow.AddMinutes(-30),
+                    state.Snapshot!,
+                    zoneId: state.BattleZoneId ?? 0,
+                    isBossBattle: state.IsBossBattle
+                );
```
ginderjeremiah: Nit: this reconstructs the entire active battle just to backdate the start time, which forces six null-forgiving `!` operators (CLAUDE.md asks to avoid `!` wherever possible). `BattleStartTime` has a public setter — the sibling `BattleServiceIntegrationTests` sets it directly (`state.BattleStartTime = DateTime.UtcNow.AddMinutes(-10);`). Since `SetPlayerState` already loads the persisted battle intact and re-saves it, you can collapse the whole callback to:

```csharp
await SetPlayerState(userId, state =>
{
    state.BattleStartTime = DateTime.UtcNow.AddMinutes(-30);
});
```

Same effect, no `!`, and it matches how the service-layer test backdates. Non-blocking.

---

ginderjeremiah (commented):
## Review: BattleLost elapsed-time anti-cheat gate (#1630)

**Verdict:** Approve (intended) — the core fix is correct, DRY, and well-tested at both the service and socket layers. Posting as a `COMMENT` because GitHub won't accept a gating verdict from the PR author; there are no blocking findings. I have **not** merged: one light should-fix (doc accuracy) and one nit are outstanding, and CI is still in progress.

The change does exactly what the issue asked: it extracts the victory path's server-clock elapsed-time check into a shared `ClaimedBeforeBattleCouldFinish` helper and applies the identical gate to `EndBattleLoss`, before `RecordBattleCompleted` (and the statistics/challenge accrual it drives) can run. I verified the reasoning holds: `TryResolveActiveBattle` for the loss path passes `maxMs = null`, which `BattleSimulator.Simulate` clamps to `DefaultMaxBattleMs`, so `result.TotalMs` is the real (capped) duration the claimed outcome took — the PR description and inline comment are accurate. Choosing to mirror the victory gate rather than entangle the claim path with the #1595/#1596 hand-back model is the right, lower-risk call. Server-authoritative anti-cheat with no simulation change, so no frontend/parity mirror is required.

### Should-fix

- **Doc accuracy — `docs/backend-battle.md`.** Bullet 20 documents the *victory*-end elapsed-time gate in detail; the loss path being the ungated outlier is exactly what this bug was. Now that `EndBattleLoss` shares the same gate, add a line so the documented anti-cheat surface stays accurate (the docs are the source of truth per CLAUDE.md, and leaving loss undocumented invites the same gap to reopen). Not in the diff, so noting it here rather than inline.

### Nits

- `BattleLostSocketTests.SetPlayerState` reconstructs the whole active battle via `SetActiveBattle(...)` with six `!` operators just to backdate the start time — collapsible to a single `state.BattleStartTime = ...` assignment (inline comment). Minor.

### Test coverage

Solid and matches the issue's ask. Both layers get a rejection test plus an updated valid-loss test:
- Service: `EndBattleLoss_ClaimedBeforeBattleCouldFinish_ReturnsFalse` asserts the claim is refused **and** the battle is left active (a good proxy for "nothing booked/persisted", since the gate returns before `RecordBattleCompleted`); `EndBattleLoss_ValidLoss_ReturnsTrueAndClearsState` backdated so it still exercises the accept path.
- Socket: `BattleLost_ClaimedBeforeBattleCouldFinish_ReturnsError` and the backdated `BattleLost_ValidLoss_Succeeds`. The connect → start → disconnect → backdate-out-of-band → reconnect dance in the valid-loss test is the right way to get the persisted (not in-memory) start time in front of the command.

### What's good

- The shared helper is the correct DRY move and keeps the two claim paths from drifting; the `out battleCompletedAt` mirrors the victory path exactly.
- Thorough, accurate inline comments explaining the exploit and why `result.TotalMs` is trustworthy here.
- The design note in the PR body flags the drawn-battle wait-out tradeoff honestly — and it's the correct anti-cheat behavior, not a regression.

---

PR #1650, opened 2026-07-05
startGame: don't opt into onDestroy cleanup outside component init (#1631)

### LINKED ISSUE #1631: startGame's push listeners call onDestroy outside component init — ChallengeCompleted/ProficiencyXpGained/ServerCommandFailed pushes are silently dropped every session
## Problem

`startGame` (`UI/src/lib/engine/engine.ts:92-95`) passes `cleanupOnDestroy: true` to all four `apiSocket.listenCommand` calls, which routes to Svelte's `onDestroy` inside `createHook.onNotified` (`UI/src/lib/common/hooks.ts:72-74`). But `startGame` never runs during component initialisation: it is invoked via `welcome.enter()` either in the async continuation of `welcome.run()` (after `await fetchProgress()` — the no-reward pass-through) or from the WelcomeBackGate button's DOM click handler. In both contexts Svelte 5's `onDestroy` throws `lifecycle_outside_component`.

`onNotified`'s own doc comment warns about exactly this ("ONLY pass `true` from within component initialization"), and `+page.svelte:77-79` documents the same constraint for its own `onDestroy(stopEngines)` registration.

## Effect (every game boot)

`listenCommand('SocketReplaced', …, true)` pushes its tracker (hooks.ts:70) **then** throws at `onDestroy(unhook)`; the exception aborts `startGame` before lines 93-95 and is swallowed by `enterGame`'s try/catch in `routes/game/+page.svelte` (console: "Failed to start game"). Net effect for the whole session:

- **`ChallengeCompleted`, `ProficiencyXpGained`, and `ServerCommandFailed` pushes are never registered** — no challenge-completion toast/zone-unlock/reward-unlock without a reload, proficiency XP/levels (which feed the live battler's attribute bonuses) never update client-side, and the dead-letter resync path is dead.
- `SocketReplaced` still fires (registered before the throw) but `socketReplacedUnhook` is never assigned, so `stopGame` can't unhook it and handlers stack across re-logins in one tab.
- The engines start before line 92, so the game *looks* healthy.

## Why tests don't catch it

Closed #814 made hook cleanup opt-in because of this constraint; a later commit opted `startGame` back in with `true`, and #875 updated `engine.test.ts` by mocking `onDestroy` away — so the unit suite can't see the throw.

## Fix

Drop the `true` flags in `startGame` and rely on the already-captured unhooks invoked in `stopGame` (they're all assigned and called there today). Consider a dev-time guard in `onNotified` so a non-component `cleanupOnDestroy` fails loudly instead of depending on Svelte's error. Add a regression test that runs `startGame` without a component context (no `onDestroy` mock) and asserts all four listeners register.

ginderjeremiah:
## Summary

`startGame` (`UI/src/lib/engine/engine.ts:92-95`) passed `cleanupOnDestroy: true` to all four `apiSocket.listenCommand` calls, which routes to Svelte's `onDestroy` inside `createHook.onNotified` (`UI/src/lib/common/hooks.ts:72-74`). But `startGame` never runs during component initialization — it's invoked via `welcome.enter()` either in the async continuation of `welcome.run()` (after `await fetchProgress()`) or from the `WelcomeBackGate` button's DOM click handler. In both contexts Svelte 5's `onDestroy` throws `lifecycle_outside_component`.

Because `+page.svelte`'s `enterGame` wraps `startGame()` in a try/catch that only logs ("Failed to start game"), the whole call silently aborted on the first throw (right after `SocketReplaced` registers, since it's listed first). Net effect on every game boot:

- `ChallengeCompleted`, `ProficiencyXpGained`, and `ServerCommandFailed` pushes were never registered — no challenge-completion toast/zone-unlock/reward-unlock without a reload, proficiency XP/levels never updated client-side, and the dead-letter resync path was dead.
- `SocketReplaced` still registered (it's called before the throw), but its unhook was never assigned, so `stopGame` couldn't unhook it and handlers stacked across re-logins in one tab.
- The engines start before the throw, so the game *looked* healthy.

## Fix

Dropped the `true` flag from all four `listenCommand` calls in `startGame`. Teardown already relies on the captured unhooks (`socketReplacedUnhook`, etc.), which `stopGame` invokes unconditionally — so no cleanup behavior is lost.

### Note on why the existing unit suite didn't catch this

`UI/src/tests/lib/engine/engine.test.ts` mocks both `svelte`'s `onDestroy` (as a no-op) and `apiSocket.listenCommand` (as a bare spy), so the real throw was invisible to it. Fixed the assertions that pinned the buggy `true` argument, and added a regression test that routes the mocked `listenCommand` through the real `createHook` from `hooks.ts` so a re-introduced `cleanupOnDestroy: true` would call the real `onDestroy` and be caught. The underlying "onDestroy isn't called by default" contract is already thoroughly covered in `hooks.test.ts`.

## Test plan

- [x] `npm run test:unit:ci` (full suite) — 294 files / 3491 tests passing, including the two new regression tests in `engine.test.ts`.

Closes #1631

---
_Generated by [Claude Code](https://claude.ai/code/session_01MHLkmV8DcrvXYXx79GicxC)_

### DIFF
--- a/UI/src/lib/engine/engine.ts
+++ b/UI/src/lib/engine/engine.ts
@@ -89,10 +89,13 @@ export const startGame = () => {
 		startLogicEngine();
 		startRenderEngine();
 		startBattleEngine();
-		socketReplacedUnhook = apiSocket.listenCommand('SocketReplaced', handleSocketReplaced, true);
-		challengeCompletedUnhook = apiSocket.listenCommand('ChallengeCompleted', handleChallengeCompleted, true);
-		proficiencyXpGainedUnhook = apiSocket.listenCommand('ProficiencyXpGained', handleProficiencyXpGained, true);
-		serverCommandFailedUnhook = apiSocket.listenCommand('ServerCommandFailed', handleServerCommandFailed, true);
+		// cleanupOnDestroy must stay off: startGame runs outside component init (from welcome.run()'s
+		// async continuation or the WelcomeBackGate click handler), where Svelte's onDestroy throws.
+		// Teardown instead relies on the unhooks captured here, invoked unconditionally by stopGame.
+		socketReplacedUnhook = apiSocket.listenCommand('SocketReplaced', handleSocketReplaced);
+		challengeCompletedUnhook = apiSocket.listenCommand('ChallengeCompleted', handleChallengeCompleted);
+		proficiencyXpGainedUnhook = apiSocket.listenCommand('ProficiencyXpGained', handleProficiencyXpGained);
+		serverCommandFailedUnhook = apiSocket.listenCommand('ServerCommandFailed', handleServerCommandFailed);
 	}
 };
 
--- a/UI/src/tests/lib/engine/engine.test.ts
+++ b/UI/src/tests/lib/engine/engine.test.ts
@@ -162,6 +162,8 @@ import {
 	inventoryManager
 } from '$lib/engine/engine';
 import { activeModal, clearModals, confirmActiveModal } from '$stores/modal.svelte';
+import { createHook } from '$lib/common/hooks';
+import { onDestroy } from 'svelte';
 import {
 	EItemCategory,
 	ELogType,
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: drop `cleanupOnDestroy` from `startGame`'s push listeners (#1631)

**Verdict:** Approve — the fix is correct, minimal, and preserves all teardown behavior; the root-cause analysis holds up against the code. (Submitted as a comment since I'm the PR author; treat this as an approval.)

I verified the fix end-to-end against the code rather than just the description:

- `apiSocket.listenCommand`'s `cleanupOnDestroy` defaults to `false` (`api-socket.ts:171`) and forwards to `createHook.onNotified`, which only calls `onDestroy(unhook)` when the flag is set (`hooks.ts:72-74`). Dropping the `true` args is exactly the right lever — the listeners still register and return their unhooks.
- No teardown is lost: `stopGame` invokes all four captured unhooks unconditionally (`engine.ts:322-325`), independent of Svelte's lifecycle. So the change removes the throwing `onDestroy` path without dropping cleanup.
- The added inline comment (`engine.ts:92-94`) accurately records *why* the flag must stay off, which is the real guard against a future re-introduction. It matches the contract documented on `onNotified` and `listenCommand`.

### Test coverage

Good — meets the project's testing rules for the touched module:
- The pinned assertions that baked in the buggy `true` arg were corrected.
- The `does not opt into onDestroy cleanup` test asserts each `listenCommand` call has exactly two args.
- The stronger regression test routes the mocked `listenCommand` through the *real* `createHook`, so a re-introduced `cleanupOnDestroy: true` would hit the real `onDestroy` and fail the test — closing the exact blind spot (`onDestroy` mocked to a no-op) that let this bug ship. The one-shot `mockImplementationOnce` ordering is explained and correct (`Object.values` insertion order matches `startGame`'s fixed call order).

### Note on the issue's optional dev-guard suggestion

#1631 suggested *considering* a dev-time guard in `onNotified` so a non-component `cleanupOnDestroy` fails loudly. I looked at this and judged it not worth adding here: Svelte 5's `onDestroy` already throws `lifecycle_outside_component`, and the real amplifier of this bug was `enterGame`'s swallowing try/catch — which would swallow a clearer guard error just the same. The added real-`createHook` regression test is the more valuable protection and covers the actual regression path directly. Not filing a follow-up.

### What's good

Tight, well-scoped bugfix with an accurate root-cause writeup, a comment that pins the constraint at the call site, and a regression test deliberately built to defeat the mock that hid the original bug.

---

PR #1651, opened 2026-07-05
Write-behind: reload the session's Player after a flush failure outlives its scope (#1632)

### LINKED ISSUE #1632: Write-behind: a failed final flush permanently loses one-shot player events (Redis/Postgres divergence that reverts unlocks on cache eviction)
## Problem

`SavePlayer`'s event dispatch buffers the aggregate's domain events into the **scoped** `PlayerUpdateBatch` (`Game.DataAccess/PlayerUpdateBatch.cs`), and the #1494 fix deliberately keeps them buffered when `FlushAsync`'s publish throws so they carry into "the batch's next flush attempt". But the batch is per-DI-scope and `SocketHandler.RunCommand` creates a scope per command (`Game.Api/Sockets/SocketHandler.cs:259`) — so when the failed flush is the command's **last** flush, the buffered envelopes die with the scope.

The events were already drained off the aggregate by the dispatch, and the in-memory `PlayerState` (socket-lifetime) keeps the mutation — so the *next* command's successful save writes the full `PlayerCacheModel` blob to Redis while the one-shot event (`ItemUnlockedEvent`, `SkillUnlockedEvent`, `ModUnlockedEvent`, `LessonUnlockedEvent`, …) never reaches the queue and is never re-raised.

## Effect

Silent, permanent Redis/Postgres divergence: the source-of-truth blob holds an unlock whose relational row never lands. DB-sourced reads (character-select `PlayerSummary`, admin views) stay stale, and when the `Player_{id}` key ages out (48h idle) or Redis data is lost, the player's unlock **silently reverts**. Low frequency (needs a Redis blip on a command's final flush) but high impact, and the queue side's "never silently drops an event" guarantee (`docs/backend-persistence.md`) has no publish-side counterpart.

## Fix directions (pick one)

- On flush failure, invalidate the in-memory aggregate / force a reload from the un-advanced cache blob so a retry regenerates the events (fails the command loudly, converges on retry).
- Or park unflushed envelopes in a connection-lifetime holder that the next command's flush drains (extends the #1494 retry across scopes).

Either way, document the chosen residual in `docs/backend-persistence.md` alongside the existing failure-mode notes.

## Tests

Integration test: fail the publish on a command that raised a one-shot event, run a subsequent successful save, then drain the queue and assert the event's row landed (or the command failed loudly and the retry re-raised it) — i.e. no path where the blob holds the unlock but Postgres never sees it.

ginderjeremiah:
## Summary

`SavePlayer`'s buffered domain events are only cleared once `PlayerUpdateBatch.FlushAsync` actually succeeds (#1494), so a transient flush failure keeps them for a retry — but that retry only works within the *same* command's DI scope. `SocketHandler` opens a fresh scope per command, so when the failing flush is a command's *last* save, the buffered envelopes die with the scope. Meanwhile the aggregate's mutation already happened before the flush ran, and the connection holds that same in-memory `Player` for its whole lifetime — so the stuck mutation would silently ride along and eventually reach Redis via some later, unrelated command's save, with its own queued event (and thus the matching Postgres row) permanently lost. Redis and Postgres diverge with no error ever surfaced past the one command that already legitimately failed.

- `PlayerRepository.SavePlayer` now wraps a non-cancellation flush failure in a new `PlayerPersistenceFlushFailedException` (`Game.Abstractions/DataAccess`), leaving the existing cancellation behavior (and the #1494 same-scope retry) untouched.
- `SessionService` gained a `PlayerNeedsReload` marker: `MarkPlayerNeedsReload()` sets it, `SetPlayer`/`ClearSession` clear it.
- `SocketHandler` marks the session when a command faults with that exception, and reloads the session's `Player` from the repository before the *next* command runs whenever the marker is set — discarding the failed command's in-memory mutations and converging back onto the last successfully-persisted state.

### Design note (this issue offered two directions — flagging per this repo's convention)

The issue offered two fixes: (1) force a reload on flush failure (fails loudly, converges on retry), or (2) park the unflushed envelopes in a connection-lifetime holder that a later command's flush drains (extending the #1494 retry across scopes). I went with (1). It's the simpler, more contained change — it reuses the existing connection-lifetime `SessionService`/`Player` split rather than requiring new plumbing to bridge the per-command DI scope back to a connection-scoped envelope buffer across the `Game.Api`/`Game.DataAccess` boundary. The trade-off: the failed command's *other* mutations (not just the one-shot event) are also discarded from memory, not just retried — but the client already receives a genuine command failure for every direct `SavePlayer` call site in the codebase.

One swallowed path is a partial exception: `BattleService.TryPrepareNextIdleBattle` catches broadly around a prefetch, so it can absorb this exception without the marker ever getting set. That's accepted (documented at the swallow) because the only `SavePlayer` reachable there — `ZoneResolutionService.EnsureViableZone`'s relocation save, invoked unconditionally from `StartBattle` — only ever raises a self-healing `PlayerCoreUpdatedEvent` (zone change): the next core mutation re-raises it, and the next battle re-runs `EnsureViableZone` and re-relocates if it's still lost. No one-shot unlock event reaches this path today; if a future change routes one through it, the swallow would silently reopen the #1632 hole and needs revisiting.

## Test plan

- [x] `dotnet test` (full backend suite: `Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Api.Tests`, `Game.Application.Tests`) — all passing (2,973 tests, 0 failures).
- [x] New coverage:
  - `PlayerWriteBehindTests.SavePlayer_PublishFailsForANonCancellationReason_ThrowsPlayerPersistenceFlushFailedException` — a non-cancellation publish failure is wrapped distinctly, and pins that the cache blob is left un-advanced (a re-read after the throw still sees the pre-mutation zone) — the invariant the reload's correctness rests on.
  - `SocketCommandExecutionTests.ExecuteCommand_PlayerPersistenceFlushFails_ReloadsPlayerBeforeTheNextCommand` — the session reloads the player from the repository before the next command, and the marker clears.
  - `SessionServiceTests` — `MarkPlayerNeedsReload`/`SetPlayer`/`ClearSession` marker interactions.
- [x] `docs/backend-persistence.md` updated with the new residual/fix under "The write-behind path".

Closes #1632


Game.Application.Tests/Services/PlayerWriteBehindTests.cs
```diff
@@ -111,6 +114,51 @@ public async Task SavePlayer_PublishFails_PreservesTheEventForTheNextFlush()
...
+                new ThrowingPubSubService(),
+                scope.ServiceProvider.GetRequiredService<IDomainEventDispatcher>(),
+                scope.ServiceProvider.GetRequiredService<PlayerUpdateBatch>(),
+                scope.ServiceProvider.GetRequiredService<IItems>(),
+                scope.ServiceProvider.GetRequiredService<IItemMods>(),
+                scope.ServiceProvider.GetRequiredService<ISkills>());
+
+            player.ChangeZone(1);
+            var ex = await Assert.ThrowsAsync<PlayerPersistenceFlushFailedException>(() => throwingRepo.SavePlayer(player));
+            Assert.IsType<InvalidOperationException>(ex.InnerException);
```
ginderjeremiah, line 147: **Should-fix (test coverage):** This test pins that the flush failure is *wrapped*, but not the invariant the whole fix rests on — that a failed flush leaves the persisted state at the last-good value so the later reload converges correctly. The reload's correctness depends entirely on `SavePlayer` writing the cache blob *after* the flush (line 122 of `PlayerRepository`), so a failed flush never advances the blob; nothing here pins that ordering. Consider adding, after the throw, a re-read asserting the un-advanced state — e.g. `Assert.Equal(originalZoneId, (await playerRepo.GetPlayer(playerEntity.Id))!.CurrentZoneId)` — so a future reorder that advanced the cache before flushing (silently reintroducing the #1632 divergence) would fail the build. This is exactly the "no path where the blob holds the [mutation] but Postgres never sees it" assertion the issue's Tests section asked for; the `SocketCommandExecutionTests` reload test uses a fake repo, so it can't cover it.

---

ginderjeremiah (commented):
## Review: write-behind reload after a flush failure outlives its scope (#1632)

**Verdict: Approve** — the fix correctly closes the severe permanent-loss class the issue targets, is layered cleanly, and is well-tested. Two should-fixes below; neither blocks. (Submitted as a comment since I'm the PR author — treat it as an approve.)

I traced the fix end-to-end and it holds:
- The reload targets `Session.Player` — the read-modify-write aggregate `SavePlayer` actually serializes — not the `PlayerState` the issue text loosely named. Correctly verified against the code rather than followed blindly.
- `SavePlayer` writes the cache blob (`PlayerRepository:122`) *after* the flush, so a failed flush leaves the blob at the last-good state and the cache-first `GetPlayer` reload converges to pre-mutation state. The mechanism is sound.
- The marker only takes effect if the exception propagates out of the command to `RunCommandUnderLock`. I checked the unlock-bearing save paths (`DefeatEnemy`/`BattleLost`/`GetOfflineProgress` → `OfflineProgressService`, which has no catch blocks; the commands only catch parameter-deserialization pre-execution) — they all propagate, so the marker is set on every one-shot-unlock save. Good.

### Should-fix

- **The PR's `TryPrepareNextIdleBattle` justification is inaccurate.** The description says that swallow "never actually reaches `SavePlayer` in practice since it always targets the player's own current zone." That covers `StartBattle`'s zone-change save (gated on `newZoneId != CurrentZoneId`), but `StartBattle` also calls `ZoneResolutionService.EnsureViableZone` unconditionally (`BattleService.cs:92`), which calls `SavePlayer` when the current zone has gone non-viable (`ZoneResolutionService.cs:83`). So a flush failure *can* be reached under that broad `catch (Exception)` and be swallowed with the marker unset. It's **not a blocker**: that path only ever carries a `PlayerCoreUpdatedEvent` (zone change), which is re-raised on the next core mutation and re-relocated on the next battle — self-healing, not the permanent one-shot-unlock loss #1632 is about. But please correct the reasoning and add a short code comment at that swallow noting *why* it's acceptable (no one-shot unlock reaches the prefetch save; the core event self-heals), so a future edit that routes an unlock through that path doesn't silently reopen the hole.

- **Test coverage — pin the load-bearing invariant.** See the inline comment on `PlayerWriteBehindTests`: the integration test asserts the wrapping but not that the failed save left the persisted state un-advanced (the reload unit test uses a fake repo, so it can't). A re-read assertion after the throw would pin the exact no-divergence guarantee the issue's Tests section requested and catch a future reorder of the flush/cache-write.

### Nits

- The `when (ex is not OperationCanceledException)` exclusion in `SavePlayer` (correctly keeping a teardown cancellation from being mistaken for a flush failure) has no direct negative-branch test. Likely acceptable as measure-only per the testing guidance, but worth a line if convenient.
- If the reload's `GetPlayer` returns `null` (`SocketHandler.cs:275`), the current command proceeds against the stale in-memory player and the marker persists to retry next command. Fine for a pathological missing-player case; noting it only because the silent fall-through isn't obvious.

### What's good

- Clean layering: the new exception lives in `Game.Abstractions/DataAccess`, the marker is a presentation-layer session concern, and the reload is orchestration in the socket handler where the `Player` lifecycle already lives — no domain logic leaked. The design note documenting why direction (1) was chosen over the connection-lifetime envelope holder is exactly the deliberate trade-off call the repo asks for, and `backend-persistence.md` is updated with the residual. Solid, well-scoped fix.


---

ginderjeremiah:
Pushed fixes for both should-fix items:

- Corrected the `TryPrepareNextIdleBattle` rationale (in the description and as a code comment on the swallow) — you're right that `ZoneResolutionService.EnsureViableZone`'s relocation save can reach this catch, not just `StartBattle`'s zone-change save. Documented why it's still acceptable (self-healing `PlayerCoreUpdatedEvent`, no one-shot unlock) and flagged it as needing revisiting if that ever changes.
- Added the re-read assertion to `PlayerWriteBehindTests` pinning that a failed flush leaves the cache blob un-advanced.

Also added the two nit comments (cancellation-branch coverage pointer, null-reload fallback note). Full suite still green (2,973 tests).

---

ginderjeremiah:
## Review: Approve

Reviewed the full diff, linked issue #1632, and the write-behind docs against the actual code paths. This is a sound, well-scoped fix and the reasoning holds up under scrutiny.

### Correctness verification
- **Cancellation / #1494 retry untouched.** The `when (ex is not OperationCanceledException)` guard keeps cancellations propagating unwrapped, and the wrap happens *after* `FlushAsync` throws without touching `PlayerUpdateBatch` state — so buffered envelopes still survive for a same-scope retry. `SavePlayer_PublishFails_PreservesTheEventForTheNextFlush` (pre-cancelled token → `OperationCanceledException`) still exercises exactly that and would fail if the wrap leaked into the cancellation path.
- **Reload discards the stuck mutation and converges.** The cache blob is only advanced *after* the flush succeeds, so a failed flush leaves Redis holding the pre-mutation state; the reload re-reads that last-persisted state onto the connection-lifetime `Player`. The new `PlayerWriteBehindTests` pins the un-advanced-blob invariant the reload rests on, and `SocketCommandExecutionTests` pins the reload-before-next-command + marker-clear end to end.
- **Marker set in the shared `RunCommandUnderLock` catch**, so both the client (`ExecuteCommand`) and pub/sub (`ExecuteServerCommand`) paths mark. Commands serialize under `_commandLock` and `SessionService` is connection-lifetime, so there's no race on the marker; `SetPlayer`/`ClearSession` clear it correctly.
- **BattleService swallow reasoning is correct.** `PrepareNextIdleBattle` passes `player.CurrentZoneId` as the target, so `StartBattle`'s explicit zone-change save (guarded by `newZoneId != CurrentZoneId`) is skipped — the *only* `SavePlayer` reachable through the prefetch is `EnsureViableZone`'s relocation save, which raises just the self-healing `PlayerCoreUpdatedEvent`. The documented caveat about a future one-shot event reopening the hole is the right guard rail.
- **Approach (1) over (2) is the right call** — it reuses the existing session/`Player` split instead of plumbing a connection-lifetime envelope buffer across the `Game.Api`/`Game.DataAccess` boundary, and "discard the failed command's mutations" is consistent with the client already receiving a genuine command failure.

Code quality is clean: exception lives in the correct abstraction layer, no null-forgiving operators, braces everywhere, docs updated concisely, no dead code.

### CI
All checks green (`test-backend`, `verify-backend`, e2e chromium/webkit/firefox, `all-checks-passed`). Frontend jobs skipped (no FE changes).

### Non-blocking follow-ups
1. **(tech debt, small)** The reload marker is only set on the synchronous fault path. A command abandoned on the 25s timeout whose flush then fails with a *non-cancellation* error faults in the `ReleaseCommandLockWhenSettled` continuation, which logs but never marks — theoretically reopening #1632. It's near-impossible in practice (the timeout cancels the token, so a racing flush almost always throws `OperationCanceledException`, not the wrapped type), but marking in that fault continuation would close it fully.
2. **(test, small)** No test pins the documented null-reload branch (`GetPlayer` returns null → marker stays set, retried next command). A small test would lock in that pathological-case behavior.

Neither blocks merge.

---

ginderjeremiah:
Two minor follow-ups filed before merge, both non-blocking and near-unreachable in practice:
- #1663 — reload marker not set on the timed-out-command abandoned-flush fault path (in practice that path faults with `OperationCanceledException`, which is deliberately left unwrapped).
- #1664 — add a test pinning the null-`GetPlayer` reload branch in `SocketHandler`.

Merging.

---

PR #1653, opened 2026-07-05
Workbench: only commit persistEntity after a child saver actually writes (#1634)

### LINKED ISSUE #1634: Workbench persistEntity misclassifies pre-commit failures after no-op child savers, discarding unsaved edits meant to be kept for retry
## Problem

In `persistEntity` (`UI/src/routes/admin/workbench/save-helpers.ts:139-145`), `committed = true` flips unconditionally after **every** child saver returns — but savers no-op internally when their collection is unchanged (e.g. `entities/item.ts` guards each POST with `if (changes.length)` / `childChanged`). `committed` gates whether a later failure is treated as recoverable (keep edits, retry) or post-commit (re-seed from server truth).

## Scenario

Edit only an item's **tags**: the identity DTO is unchanged (child collections are blanked in `toPrimaryDto`), so `postPrimary` never runs; the attribute and mod-slot savers no-op yet still flip `committed`; then the tags saver (`SetTagsForItem`) fails on a transient error. The catch sees `committed === true`, throws `PersistFailedError`, and `EntityStore.save()` (`entity-store.svelte.ts:208-223`) re-seeds from the server — **silently discarding the tag edit** that the design explicitly intends to keep for a clean retry (per the comments at `save-helpers.ts:121-122` and `entity-store.svelte.ts:208-213`).

## Fix

Have child savers report whether they actually wrote (return a boolean, only then flip `committed`), or hoist the changed-collection check into `persistEntity` so no-op savers never mark the record committed.

## Tests

Unit test: a record whose only change is a child collection, whose saver throws before writing → the store keeps the pending edit rather than re-seeding.

ginderjeremiah:
## Summary

`persistEntity` (`UI/src/routes/admin/workbench/save-helpers.ts`) flipped `committed = true` unconditionally after **every** child saver ran, even when the saver no-op'd because its collection was unchanged. `committed` gates whether a later failure is treated as recoverable (keep the user's pending edit for retry) or post-commit (re-seed from server truth, discarding the edit). A record whose only real change lives in a *later* child collection — with earlier no-op savers running first — would get its edit silently discarded on a pre-commit failure, because the no-ops had already (incorrectly) marked the record committed.

- `ChildSaver<T>` now returns `Promise<boolean>` — whether it actually wrote — instead of `Promise<void>`.
- `persistEntity` only ORs `committed` with a saver's return value, so a run of no-op savers no longer marks the record committed.
- Updated every `childSavers` definition (`class.ts`, `enemy.ts`, `item-mod.ts`, `item.ts`, `lesson.ts`, `skill-recipe.ts`, `skill.ts`, `zone.ts`) to return `true`/`false` based on whether it actually posted a change, matching their existing no-op guards (`changes.length` / `childChanged(...)`).

## Test plan

- [x] Added a unit test (`save-helpers.test.ts`): a record with only a child-collection edit, where an earlier no-op saver runs before a saver that fails *before* writing — the failure now propagates raw (not wrapped as `PersistFailedError`), so the caller keeps the pending edit.
- [x] Updated the existing `persistEntity` tests' inline child savers to return a boolean per the new `ChildSaver` contract.
- [x] `npm run test:unit:ci` (full suite) — 295 files / 3496 tests passing.

Closes #1634

---
_Generated by [Claude Code](https://claude.ai/code/session_013L6Nw9Jo7eexkoFpc7CVN7)_

### DIFF
--- a/UI/src/routes/admin/workbench/entities/class.ts
+++ b/UI/src/routes/admin/workbench/entities/class.ts
@@ -244,28 +244,34 @@ export const classEntity: EntityConfig<WorkbenchClass> = {
 			refresh,
 			childSavers: [
 				async (id, record, baseline) => {
-					if (childChanged(record.starterSkillIds, baseline?.starterSkillIds)) {
-						await ApiRequest.post('AdminTools/SetClassStarterSkills', {
-							classId: id,
-							skillIds: record.starterSkillIds
-						});
+					if (!childChanged(record.starterSkillIds, baseline?.starterSkillIds)) {
+						return false;
 					}
+					await ApiRequest.post('AdminTools/SetClassStarterSkills', {
+						classId: id,
+						skillIds: record.starterSkillIds
+					});
+					return true;
 				},
 				async (id, record, baseline) => {
-					if (childChanged(record.starterEquipment, baseline?.starterEquipment)) {
-						await ApiRequest.post('AdminTools/SetClassStarterEquipment', {
-							classId: id,
-							equipment: record.starterEquipment
-						});
+					if (!childChanged(record.starterEquipment, baseline?.starterEquipment)) {
+						return false;
 					}
+					await ApiRequest.post('AdminTools/SetClassStarterEquipment', {
+						classId: id,
+						equipment: record.starterEquipment
+					});
+					return true;
 				},
 				async (id, record, baseline) => {
-					if (childChanged(record.attributeDistributions, baseline?.attributeDistributions)) {
-						await ApiRequest.post('AdminTools/SetClassAttributeDistributions', {
-							classId: id,
-							attributeDistributions: record.attributeDistributions
-						});
+					if (!childChanged(record.attributeDistributions, baseline?.attributeDistributions)) {
+						return false;
 					}
+					await ApiRequest.post('AdminTools/SetClassAttributeDistributions', {
+						classId: id,
+						attributeDistributions: record.attributeDistributions
+					});
+					return true;
 				}
 			]
 		})
--- a/UI/src/routes/admin/workbench/entities/enemy.ts
+++ b/UI/src/routes/admin/workbench/entities/enemy.ts
@@ -196,22 +196,28 @@ export const enemyEntity: EntityConfig<WorkbenchEnemy> = {
 			refresh,
 			childSavers: [
 				async (id, record, baseline) => {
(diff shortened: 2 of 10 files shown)


ginderjeremiah (commented):
## Review: Workbench — only mark a record committed after a child saver actually writes (#1634)

**Verdict:** Approve — correctly fixes the misclassification; well-scoped and well-tested. _(Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on my own PR — the real call is Approve.)_

The `committed` flag is the pivot of the whole recoverable-vs-re-seed decision, and the fix targets it precisely: `ChildSaver` now reports whether it wrote, and `persistEntity` does `committed ||= wrote` instead of flipping unconditionally. I traced the original issue scenario (tags-only edit) through the new code — no primary write, both preceding savers `return false`, the tags saver throws before its POST → `committed` stays `false` → the raw error propagates and the caller keeps the edit for retry. The `||=` also correctly latches `committed` once *any* saver writes, so a later no-op can't un-commit it. Correct.

### Nits (optional)
- The `if (!changed) { return false; } await ApiRequest.post(...); return true;` shape now repeats across ~18 savers in the 8 entity files. A small helper — e.g. `guardedSave(changed: boolean, post: () => Promise<unknown>): Promise<boolean>` — would collapse each saver to one expression and remove the boilerplate (`CLAUDE.md`: prefer DRY). It's pre-existing structure this PR only minimally extended, so it's fine as a follow-up rather than something to fold in here.

### Test coverage
- Solid. The new `save-helpers.test.ts` case pins the exact defect: a modified record whose only change is a *later* child collection, with an earlier no-op saver ahead of a saver that throws before writing → the failure propagates **raw** (not wrapped as `PersistFailedError`), which is the observable proof the edit is kept. Existing `persistEntity` tests were updated to the new boolean contract, and the per-entity `persist` tests still run through the real `persistEntity`, so a saver returning the wrong boolean would surface there. No gap worth blocking on — the per-entity changes are mechanical mirrors of their existing guards.

### What's good
- The fix matches the issue's recommended direction and the PR body/commit history make the reasoning easy to follow.
- The `ChildSaver` contract change is documented right at the type (`// Returns whether it actually wrote — a no-op … must report false`), so the next author sees the invariant at the point they'd need it.
- Consistent application across all 8 entity configs with no missed saver.


---

PR #1655, opened 2026-07-05
SocketReplaced: stop misclassifying its self-delivered push as NotDelivered (#1636)

### LINKED ISSUE #1636: Every session takeover logs a spurious "Failed to deliver a server-initiated socket command" warning (SocketReplaced misclassified as NotDelivered)
## Problem

`SocketReplaced.ExecuteAsync` (`Game.Api/Sockets/Commands/SocketReplaced.cs`) deliberately owns its own send-then-close (the success frame must reach the client **before** the close frame). The command runner then unconditionally sends the returned response; `SocketContext.SendData` returns `false` because the socket is no longer Open, so `RunCommandUnderLock` (`Game.Api/Sockets/SocketHandler.cs:217-218`) classifies the outcome `NotDelivered`, and `ExecuteServerCommand` (`SocketHandler.cs:159-165`) warn-logs "Failed to deliver a server-initiated socket command: SocketReplaced … the client was likely disconnecting" — for a push that actually succeeded.

Every legitimate takeover (second tab/device — the normal single-active-connection flow) emits this false-failure warning, polluting warning-level logs and diluting the real signal `NotDelivered` exists to give. The command's comment covers the no-op send but not the outcome classification, and no test pins it.

## Fix

Let a command report that it delivered its own response — e.g. a "self-delivering" marker on the command the runner checks before classifying, or map `NotDelivered` → `Succeeded` when the socket's close reason is `SocketReplaced`. Pin with a test asserting a takeover logs no delivery warning.

ginderjeremiah:
## Summary

`SocketReplaced.ExecuteAsync` (`Game.Api/Sockets/Commands/SocketReplaced.cs`) deliberately owns its own send-then-close, since the success frame must reach the client **before** the close frame. But `RunCommandUnderLock` (`Game.Api/Sockets/SocketHandler.cs`) always ran its own follow-up `SendData` afterward — which returns `false` because the socket is no longer `Open` — so every legitimate session takeover got classified `NotDelivered` and warn-logged as "Failed to deliver a server-initiated socket command", even though the push had actually succeeded.

- Added `ISelfDeliveringCommand` (`Game.Api/Sockets/Commands/ISelfDeliveringCommand.cs`), a marker interface following the existing `IServerInitiatedCommand` pattern, for a command that sends its own response instead of letting the runner send it.
- `SocketReplaced` now implements it.
- `RunCommandUnderLock`/`RunCommand` in `SocketHandler` check whether the executed command is self-delivering; if so, the runner skips its own follow-up send and reports `SocketCommandOutcome.Succeeded` directly instead of running a guaranteed-to-fail send and misclassifying the result.

## Test plan

- [x] `dotnet test` (full backend suite: `Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Api.Tests`, `Game.Application.Tests`) — all passing (2,970 tests, 0 failures).
- [x] New coverage in `SocketCommandExecutionTests`:
  - `ExecuteServerCommand_SelfDeliveringCommand_ReturnsSucceededWithoutASecondSendOrDeliveryWarning` — a self-delivering command (mirroring `SocketReplaced`'s send-then-close shape) reports `Succeeded` with no second send and no "Failed to deliver" warning.
  - `ExecuteCommand_SelfDeliveringCommand_DoesNotAttemptASecondSend` — same assertion via the client-initiated path.

Closes #1636

---
_Generated by [Claude Code](https://claude.ai/code/session_01J38oCCtE7GCu9YtmFK3Z5i)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
@@ -1,4 +1,5 @@
 using Game.Abstractions.DataAccess;
+using Game.Api;
 using Game.Api.Models.Common;
 using Game.Api.Services;
 using Game.Api.Sockets;
@@ -193,6 +194,32 @@ namespace Game.Api.Tests.Unit
         }
 
         [Fact]
+        public async Task ExecuteServerCommand_SelfDeliveringCommand_ReturnsSucceededWithoutASecondSendOrDeliveryWarning()
+        {
+            // Mirrors SocketReplaced: the command sends its own response and closes the socket before
+            // returning. The runner must not attempt (and then misclassify as NotDelivered) a follow-up send
+            // on the now-closed socket (#1636).
+            var (socket, handler) = CreateHandler(_ => null, selfDelivering: name => name == "SelfDelivers");
+
+            var outcome = await handler.ExecuteServerCommand(new SocketCommandInfo("SelfDelivers") { Id = "c1" });
+
+            Assert.Equal(SocketCommandOutcome.Succeeded, outcome);
+            Assert.Single(socket.SentMessages, m => m.Contains("c1"));
+            Assert.DoesNotContain(_logs.Entries, e => e.Message.Contains("Failed to deliver"));
+        }
+
+        [Fact]
+        public async Task ExecuteCommand_SelfDeliveringCommand_DoesNotAttemptASecondSend()
+        {
+            var (socket, handler) = CreateHandler(_ => null, selfDelivering: name => name == "SelfDelivers");
+
+            await handler.ExecuteCommand(new SocketCommandInfo("SelfDelivers") { Id = "c1" });
+
+            Assert.Single(socket.SentMessages, m => m.Contains("c1"));
+            Assert.DoesNotContain(_logs.Entries, e => e.Message.Contains("Failed to deliver"));
+        }
+
+        [Fact]
         public async Task HandleMessage_UnparseableJson_ClosesTheSocketInsteadOfLeavingTheClientToTimeOut()
         {
             // An invalid-JSON frame carries no request id to correlate a structured rejection to, so closing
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: SocketReplaced self-delivered push no longer misclassified as NotDelivered (#1636)

**Verdict:** Approve — a clean, correctly-scoped fix that stops every session takeover from warn-logging a spurious "Failed to deliver" for a push that actually succeeded. (Posted as a comment because I'm the PR author; the intended verdict is Approve.)

### What's good

- **Right design choice.** The `ISelfDeliveringCommand` marker mirrors the existing `IServerInitiatedCommand` pattern and lets the runner decide generically, rather than the issue's alternative of hard-coding a `close reason == SocketReplaced` check into `RunCommandUnderLock`. That keeps the runner decoupled from any one command's close semantics and leaves the door open for future self-delivering commands.
- **Correct ordering.** The self-delivered short-circuit (`SocketHandler.cs:214`) sits *after* the timeout/cancellation handling, so a wedged self-delivering command still classifies as `TimedOut` — the fix only suppresses the follow-up send on the success path, exactly where the false `NotDelivered` came from.
- **Clean threading.** `RunCommand` now returns `(Response, SelfDelivered)` and the tuple flows through `RunCommandUnderLock` without disturbing the timeout/lock-handoff logic. The `SocketReplaced` comment was updated to point at the marker instead of the stale "runner's no-op send" description.

### Test coverage

Sufficient and on-point for a scope:small bug fix:
- `ExecuteServerCommand_SelfDeliveringCommand_ReturnsSucceededWithoutASecondSendOrDeliveryWarning` covers the real server-initiated path (the actual `SocketReplaced` shape) — asserts `Succeeded`, a single send, and no "Failed to deliver" warning, which is precisely what the issue asked to pin.
- `ExecuteCommand_SelfDeliveringCommand_DoesNotAttemptASecondSend` covers the client-initiated runner path for the same mechanism.
- The `SelfDeliveringStubCommand` faithfully mirrors `SocketReplaced`'s send-then-close-then-return-Success shape.

### Notes (non-blocking)

- One behavioral consequence worth being explicit about: for a self-delivering command the runner now reports `Succeeded` unconditionally, even if the command's own internal `SendData` returned false (client already gone). For `SocketReplaced` that's correct — the connection is being replaced, so a genuine delivery failure to the outgoing client is moot and is covered by re-sync on reconnect anyway. No change needed; just flagging the semantics for the next reader.
- Neither `IServerInitiatedCommand` nor `ISelfDeliveringCommand` is documented in `docs/`, which is consistent — no doc update warranted here.

No blocking, should-fix, or nit findings. Will merge once CI (backend + e2e still running) goes green.

---

PR #1656, opened 2026-07-06
Exclude dormant skills from Skills header combined dps/burst totals (#1637)

### LINKED ISSUE #1637: Skills header "eff. dps / eff. burst" totals count dormant (off-weapon) skills that can't fire
## Problem

`combinedEffectiveDps`/`combinedEffectiveBurst` (`UI/src/routes/game/screens/skills/skills-view.svelte.ts:357-360`, rendered in the `Skills.svelte` header) reduce over the raw `this.equipped` loadout with no `dormant()` gate — while the same screen suppresses per-card stats for dormant skills (`EquippedBand.svelte` swaps `SkillCardStats` for `DormantNote`) precisely because a weapon-mismatched skill is not fielded in battle (#1342).

Unequip your sword with three sword skills equipped and the header keeps summing their DPS — an output the battle can never produce, under a heading that reads as "the loadout you fight with".

## Fix

Exclude dormant skills from both totals (reuse the existing `dormant()` predicate against `metricsById`).

## Secondary design question (resolve in the same pass)

Fielded **innate** (item-granted) skills *do* fire but are excluded from the totals today. Decide whether the header should reflect the true fielded set (selected-minus-dormant **plus** granted) or stay a selected-loadout-only readout — and label it accordingly.

## Tests

Unit tests on the view-model: a dormant skill contributes 0 to both totals; totals restore when the matching weapon returns.

ginderjeremiah:
## Summary

`combinedEffectiveDps`/`combinedEffectiveBurst` (`UI/src/routes/game/screens/skills/skills-view.svelte.ts`) reduced over the raw equipped loadout with no `dormant()` gate, while the same screen already suppresses per-card stats for a dormant (weapon-mismatched) skill via `EquippedBand.svelte` (#1342). Unequipping the matching weapon left the header's "eff. dps"/"eff. burst" totals still summing damage from skills the battle can never fire.

## Fix

Added a private `fieldedEquipped` derived value that filters the equipped loadout through the existing `dormant()` predicate against `metricsById`, and reduced both totals over it instead of the raw `equipped` list.

## Design-question note (from the issue's "resolve in the same pass" ask)

The issue also raised whether fielded **innate** (item-granted) skills — which do fire in battle but are excluded from these totals today — should be folded in, versus keeping the header a selected-loadout-only readout. I kept it scoped to the selected loadout (the minimal fix for the reported bug: an unfireable output shown as fireable) rather than also changing what's counted, since that's a separate, non-trivial scope/labeling decision. Left a code comment flagging it for whoever revisits.

## Test plan

- [x] `npm run test:unit:ci` (full suite) — 295 files / 3498 tests passing, including new coverage:
  - A dormant equipped skill contributes 0 to both `combinedEffectiveDps`/`combinedEffectiveBurst`.
  - The skill's contribution is restored once the matching weapon is equipped.

Closes #1637

---
_Generated by [Claude Code](https://claude.ai/code/session_01JJfVmoze5j61yYABJKXDRf)_

### DIFF
--- a/UI/src/routes/game/screens/skills/skills-view.svelte.ts
+++ b/UI/src/routes/game/screens/skills/skills-view.svelte.ts
@@ -355,8 +355,8 @@ export class SkillsView {
 
 	/** Equipped skills the battle can actually fire — excludes a weapon-mismatched skill dormant under
 	 *  the equipped weapon (#1342), so the combined totals below never sum an output the battle can't
-	 *  produce. Innate (item-granted) skills also fire but are a selected-loadout-only readout here;
-	 *  see #1637 for the open question of whether they belong in these totals. */
+	 *  produce. Innate (item-granted) skills also fire but are deliberately excluded — the totals stay
+	 *  a selected-loadout-only readout; see #1657 for the open question of folding them in. */
 	private readonly fieldedEquipped = $derived(
 		this.equipped.filter((id) => {
 			const skill = this.metricsById[id]?.skill;
--- a/UI/src/routes/game/screens/skills/skills-view.svelte.ts
+++ b/UI/src/routes/game/screens/skills/skills-view.svelte.ts
@@ -353,11 +353,22 @@ export class SkillsView {
 	/** Metrics for the inspected skill, falling back to the first listed skill. */
 	readonly selected = $derived(this.metricsById[this.selectedId] ?? this.railList[0]);
 
+	/** Equipped skills the battle can actually fire — excludes a weapon-mismatched skill dormant under
+	 *  the equipped weapon (#1342), so the combined totals below never sum an output the battle can't
+	 *  produce. Innate (item-granted) skills also fire but are a selected-loadout-only readout here;
+	 *  see #1637 for the open question of whether they belong in these totals. */
+	private readonly fieldedEquipped = $derived(
+		this.equipped.filter((id) => {
+			const skill = this.metricsById[id]?.skill;
+			return skill != null && !this.dormant(skill);
+		})
+	);
+
 	/** Combined effective DPS of the equipped loadout vs the current Toughness. */
-	readonly combinedEffectiveDps = $derived(this.equipped.reduce((sum, id) => sum + this.effectiveDps(id), 0));
+	readonly combinedEffectiveDps = $derived(this.fieldedEquipped.reduce((sum, id) => sum + this.effectiveDps(id), 0));
 
 	/** Combined effective single-hit burst of the equipped loadout vs the current Toughness. */
-	readonly combinedEffectiveBurst = $derived(this.equipped.reduce((sum, id) => sum + this.effective(id), 0));
+	readonly combinedEffectiveBurst = $derived(this.fieldedEquipped.reduce((sum, id) => sum + this.effective(id), 0));
 
 	/* ── per-skill effective reads (Toughness-aware) ─────────────────────────── */
 
--- a/UI/src/tests/routes/game/screens/skills/skills-view.test.ts
+++ b/UI/src/tests/routes/game/screens/skills/skills-view.test.ts
@@ -718,6 +718,44 @@ describe('SkillsView — critical hits fold into effective damage', () => {
 	});
 });
 
+/* The header's combined effective totals must never sum an output the battle can't produce (#1637):
+   a dormant (weapon-mismatched) equipped skill is excluded, mirroring the per-card suppression
+   `EquippedBand.svelte` already applies. */
+describe('SkillsView — combined effective totals exclude dormant skills (#1637)', () => {
+	const swordSkill = skill({
+		id: 6,
+		name: 'Golf',
+		baseDamage: 20,
+		cooldownMs: 1000,
+		damagePortions: [{ type: EDamageType.Sword, weight: 1 }]
+	});
+
+	beforeEach(() => {
+		staticData.skills = [...SKILLS, swordSkill];
+		mockPlayerManager.unlockedSkills = [
+			{ skillId: 0, selected: true, order: 0 },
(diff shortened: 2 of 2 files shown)


UI/src/routes/game/screens/skills/skills-view.svelte.ts
```diff
@@ -353,11 +353,22 @@ export class SkillsView {
 	/** Metrics for the inspected skill, falling back to the first listed skill. */
 	readonly selected = $derived(this.metricsById[this.selectedId] ?? this.railList[0]);
 
+	/** Equipped skills the battle can actually fire — excludes a weapon-mismatched skill dormant under
+	 *  the equipped weapon (#1342), so the combined totals below never sum an output the battle can't
+	 *  produce. Innate (item-granted) skills also fire but are a selected-loadout-only readout here;
+	 *  see #1637 for the open question of whether they belong in these totals. */
```
ginderjeremiah: Nit: this comment points at `#1637` as the home of "the open question of whether they belong in these totals" — but this PR closes #1637, so once merged the reference dangles at a closed issue for a supposedly-open question. The decision here (header stays a selected-loadout-only readout) is a legitimate resolution of the issue's secondary design question, so either reword to state that decision without pointing at the closed issue, or file a dedicated follow-up issue for the innate-inclusion question and reference *that*. Not blocking.

---

ginderjeremiah (commented):
## Review: exclude dormant (off-weapon) skills from the Skills header combined effective totals (#1637)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating event from me) — the reported bug is fixed correctly and minimally, with the right tests.

The fix is on target: `fieldedEquipped` filters the equipped loadout through the existing `dormant()` predicate (which delegates to the shared `isSkillDormant` in `$lib/battle/loadout.ts`), so the header totals now reuse the *same* weapon-match rule the per-card suppression uses (`EquippedBand.svelte`, #1342) rather than a display-only copy — consistent with the frontend battle-formula-sharing guideline. Both totals reduce over the filtered set. No regression for the null-metrics case (a metrics-less equipped id contributed 0 before and is now simply filtered out).

The issue's secondary design question (should fielded innate/item-granted skills count?) is answered by keeping the header a selected-loadout-only readout — a defensible, in-scope decision, documented in the code comment.

### Nits
- `skills-view.svelte.ts:358` — the new comment points at `#1637` for "the open question of whether [innate skills] belong in these totals," but this PR closes #1637, leaving a dangling reference to a closed issue. Reword to state the decision, or file a dedicated follow-up issue for the innate-inclusion question and point at that. (Inline comment left.)

### Test coverage
- Adequate and matches what the issue asked for: a dormant equipped skill contributes 0 to both `combinedEffectiveDps`/`combinedEffectiveBurst`, and its contribution is restored once the matching weapon is equipped. Both assert against the exact expected sums (skill 6 excluded/included). Placed correctly under `tests/` mirroring the source.

### What's good
- Minimal, well-scoped fix that eliminates a display/battle divergence by reusing the canonical predicate instead of re-deriving the weapon-match rule; clear naming (`fieldedEquipped`) and a comment that explains the *why*.


---

UI/src/routes/game/screens/skills/skills-view.svelte.ts
ginderjeremiah: Filed #1657 for the innate-inclusion question and reworded the comment to state the decision and point there instead (d8f7a77).

---

ginderjeremiah:
## Review: Approve ✅

**Summary.** Clean, minimal, well-scoped fix for #1637. A new private `fieldedEquipped` derived filters the equipped loadout through the existing `dormant()` predicate, and both `combinedEffectiveDps`/`combinedEffectiveBurst` reduce over it instead of the raw `equipped` list. The header no longer sums an output the battle can never fire.

**Correctness — verified against the source, not just the description:**
- `dormant()` delegates to the shared `isSkillDormant` in `$lib/battle/loadout.ts` — the same rule `Battler.fillSkills` uses to field skills. `EquippedBand.svelte` gates per-card stats with the very same `view.dormant(...)` call, so the header and cards can't diverge. ✔
- `fieldedEquipped` reads the skill from `metricsById[id]?.skill` (the same objects `equippedRail` is built from) and guards with `skill != null && !this.dormant(skill)` — no null-forgiving operator, correct nullable handling. ✔
- Numerically safe: excluded/missing ids would have contributed 0 anyway (`rawDamage ?? 0`), so the only behavioral change is dropping dormant skills — exactly the intent. ✔
- Tests cover both directions (dormant → 0 contribution; matching weapon equipped → contribution restored) and assert against `effective`/`effectiveDps` sums with `toBeCloseTo`. Braces/style conform to project rules. ✔

**Deferred design question is handled correctly.** Scoping out the "should fielded innate (item-granted) skills be folded in?" question is the right call — it's a separate labeling/scope decision, not part of this bug (an *unfireable* output shown as fireable). It's captured in follow-up #1657 and flagged with an in-code comment pointing there. No reason to widen this PR.

**Nits (non-blocking):**
- Minor DRY: `equippedRail` already maps `equipped → metrics (non-null)`. `fieldedEquipped` re-derives the same map to produce ids. Not worth changing here (the reduces key off `id`), but a future cleanup could have the totals reduce over a single dormant-filtered metrics list.

**CI:** all-checks-passed green (frontend/e2e all pass; backend jobs correctly skipped for a UI-only change).

No blocking issues.

---

ginderjeremiah:
Follow-up filed before merge: #1662 — minor DRY cleanup for the combined totals re-deriving the equipped→metrics map that `equippedRail` already builds. Non-blocking. (The deferred innate-skills design question is already tracked in #1657.) Merging.

---

PR #1658, opened 2026-07-06
Guard openLesson against zero-step lessons wedging the tutorial tour (#1638)

### LINKED ISSUE #1638: A zero-step lesson wedges the tutorial-tour flow and re-triggers on every screen visit
## Problem

A lesson with zero steps is saveable — the Workbench only *warns* ("No tour steps", `entities/lesson.ts:121-124`; warnings never block saves). If such a screen-anchored lesson goes live:

- `evaluateScreenTrigger` (`UI/src/lib/engine/tutorials.ts`) plays it via `openLesson`.
- `TourPlayer.svelte` renders nothing (`{#if open && currentStep}` — `currentStep` is undefined).
- `closeTutorialTour` — the only thing that marks a lesson read — is reachable solely through TourPlayer's callbacks, which never fire.

The lesson stays locked and invisibly "active", re-firing on every activation of its host screen; a mechanic-anchored zero-step lesson wedges the same way when opened from the unread queue. Timely because #1605 is about to author the first real lessons.

## Fix

Guard the open path: `openLesson`/`evaluateScreenTrigger` skip (or immediately mark read) a lesson with no steps, so bad content degrades to a no-op instead of a stuck tour. Optionally also have `tutorialTour.play` refuse an empty-step lesson defensively.

## Tests

Unit tests in `tutorials.test.ts`: a zero-step screen-anchored lesson neither plays a tour nor re-triggers repeatedly (and/or is marked read once).

ginderjeremiah:
## Summary

A lesson with zero steps is saveable in the Workbench (only a warning, never a blocking validation), and if such a screen-anchored lesson goes live it wedges the tutorial-tour flow:

- `TourPlayer.svelte` renders nothing when `currentStep` is `undefined` (`{#if open && currentStep}`), so its `onDismiss`/`onComplete` callbacks never fire.
- `closeTutorialTour` — the only thing that marks a lesson read — is only reachable through those callbacks.
- The lesson stays permanently "active" and, for a screen-anchored lesson, re-triggers on every activation of its host screen (`evaluateScreenTrigger` keeps finding it since it's still locked). A mechanic-anchored zero-step lesson wedges the same way once opened from the unread queue.

## Fix

`openLesson` (`UI/src/lib/engine/tutorials.ts`) now guards on `lesson.steps.length === 0`: instead of navigating and playing a tour that can never close, it calls `playerManager.markLessonRead` directly and returns — a no-op that also unlocks/marks the lesson so it stops being reconsidered by `evaluateScreenTrigger` on later visits.

Since `openLesson` is the single production call site for `tutorialTour.play` (verified via search), gating there covers both the screen-anchored auto-trigger and any manual open of a queued unread lesson (e.g. from the Help nav badge). I did not add the issue's optional secondary guard directly in `tutorialTour.play` — it isn't needed given the single call site, and `closeTutorialTour`'s own tests rely on calling `tutorialTour.play` directly with a lesson that doesn't otherwise matter for that test's purpose.

## Tests

Added to `UI/src/tests/lib/engine/tutorials.test.ts`:
- `openLesson` marks a zero-step lesson read immediately instead of opening a tour (no navigation, no active tour).
- `evaluateScreenTrigger` does not wedge on a zero-step screen-anchored lesson, and does not re-fire it on a later revisit.

`makeLesson`'s default now includes one step (previously `[]`) since most existing tests assert the tour actually opens; the two new tests explicitly override `steps: []`.

## Test plan

- [x] `npm run test:unit:ci` (full suite) — 295 files / 3498 tests passing

Closes #1638

---
_Generated by [Claude Code](https://claude.ai/code/session_01EEKYo77b8oY3QYoQkPdpkj)_

### DIFF
--- a/UI/src/lib/engine/tutorials.ts
+++ b/UI/src/lib/engine/tutorials.ts
@@ -15,8 +15,17 @@ const isLocked = (lessonId: number): boolean => !playerManager.lessons.some((les
  * Opens `lesson`'s tour: navigates to its host screen (a no-op if already there) and plays its
  * steps. This is the single open→navigate→play flow shared by the screen-anchored auto-trigger
  * below and manually opening a queued unread lesson (e.g. from the Help nav badge, #1589).
+ *
+ * A lesson with no steps can't render `TourPlayer` (it has nothing to show as `currentStep`), so
+ * its dismiss/complete callbacks — the only path that marks a lesson read — would never fire.
+ * Rather than wedge as a permanently "active" lesson that keeps re-triggering, treat it as a no-op
+ * and mark it read immediately so bad content degrades gracefully instead of getting stuck.
  */
 export function openLesson(lesson: ILesson) {
+	if (lesson.steps.length === 0) {
+		playerManager.markLessonRead(lesson.id);
+		return;
+	}
 	navigation.requestScreen(lesson.screenKey);
 	tutorialTour.play(lesson);
 }
--- a/UI/src/tests/lib/engine/tutorials.test.ts
+++ b/UI/src/tests/lib/engine/tutorials.test.ts
@@ -45,7 +45,7 @@ const makeLesson = (overrides: Partial<ILesson> = {}): ILesson => ({
 	screenKey: 'fight',
 	ordinal: 0,
 	designerNotes: '',
-	steps: [],
+	steps: [{ ordinal: 0, text: 'Step text' }],
 	...overrides
 });
 
@@ -80,6 +80,16 @@ describe('openLesson', () => {
 		expect(navigation.requestedScreen).toBe('skills');
 		// $state wraps the assigned lesson in a reactive proxy, so compare by value, not identity.
 		expect(tutorialTour.activeLesson).toEqual(lesson);
+	});
+
+	it('marks a zero-step lesson read immediately instead of opening a stuck tour (#1638)', () => {
+		const lesson = makeLesson({ id: 8, steps: [] });
+
+		openLesson(lesson);
+
+		expect(mockMarkLessonRead).toHaveBeenCalledWith(8);
+		expect(tutorialTour.activeLesson).toBeNull();
+		expect(navigation.requestedScreen).toBeNull();
 	});
 });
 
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Guard `openLesson` against zero-step lessons wedging the tutorial tour (#1638)

**Verdict:** Approve (posted as a comment since GitHub won't accept a self-approval on my own PR) — the fix is minimal, correct, and lands the guard at the right single choke point, with faithful test coverage.

I traced the fix against the issue and the surrounding code:

- The guard lives in `openLesson` (`UI/src/lib/engine/tutorials.ts:25`), the sole production call site of `tutorialTour.play` and the shared entry for both the screen-anchored auto-trigger (`evaluateScreenTrigger`) and the manual Help-badge open. Placing it there is the DRY choice — better than duplicating the check in `evaluateScreenTrigger`, and it covers the mechanic-anchored-from-unread-queue case too.
- `markLessonRead(lesson.id)` on a never-unlocked lesson is safe and already the established pattern: screen-anchored lessons go locked → `markLessonRead` with no prior `unlockLesson` in the normal flow (`closeTutorialTour`), and `PlayerManager.markLessonRead` explicitly pushes a fresh unlocked+read entry when no entry exists (`player-manager.ts:186-198`, documented there). So no new server-command path is introduced.
- Convergence holds: once marked read, `isLocked(candidate.id)` returns false, so `evaluateScreenTrigger` stops reconsidering it — no re-fire on revisit. The second test pins exactly this.
- `ILesson.steps` is a non-nullable `ILessonStep[]` (`contracts.ts:163`), so `steps.length === 0` can't throw.

### Test coverage
Solid. Two new tests in `tutorials.test.ts` cover the two wedge paths from the issue (immediate mark-read on manual open; no wedge + no re-fire on screen revisit), and the player-manager stub faithfully mirrors the real idempotent/absent-tolerant behavior rather than a hand-waved spy. The `makeLesson` default gaining one step is the right call given most existing tests assert the tour opens; the two new tests override `steps: []` explicitly. This is frontend-only, non-battle logic, so no backend mirror is required.

### On the skipped optional guard
Not adding the defensive `steps.length === 0` check inside `tutorialTour.play` is justified: single production call site, and `closeTutorialTour`'s tests drive `play` directly. Agreed — leaving it out avoids gold-plating.

### What's good
Tight, well-scoped fix with a clear explanatory comment on the guard, and the graceful-degradation framing (bad content → no-op, not a stuck tour) matches the issue's intent exactly.


---

PR #1659, opened 2026-07-06
Refresh InventoryManager on mid-session player resync (#1639)

### LINKED ISSUE #1639: Mid-session player resync never refreshes InventoryManager — an outage-window item/mod unlock stays invisible until reload
## Problem

The lost-response recovery path (`claimVictory` → `refreshPlayer()` on a `DefeatEnemy` transport error, `UI/src/lib/engine/battle/enemy-manager.ts:645-652`) re-pulls `Login/Status` and re-initializes `playerManager` (`UI/src/lib/engine/session.ts` → `loadPlayer`) — level, exp, unlocked skills. But it never re-runs `inventoryManager.initialize()`, which runs only in `startGame` (`UI/src/lib/engine/engine.ts:78`). So an item/mod unlocked server-side during the outage window stays invisible until a full page reload.

The same gap underlies `handleServerCommandFailed`'s comment ("any reward/skill it carried still surfaces on the next natural load of the inventory/skills stores") — no such natural load exists mid-session for the inventory. Skills recover via the `playerManager` refresh; items/mods don't.

**Scenario:** the connection blips exactly as a boss kill completes a zone-gating challenge with an item reward — exp/level resync, the reward item doesn't appear.

## Fix

Have the resync path also refresh the inventory (re-run `inventoryManager.initialize()` from the refreshed player payload, preserving the by-reference re-derive contract), or extract a shared "resync player + inventory" helper both `startGame` and `refreshPlayer` use.

## Tests

Unit test on the recovery path asserting the inventory reflects an item unlocked between battle start and the resync.

ginderjeremiah:
## Summary

The lost-response recovery path (`claimVictory` → `refreshPlayer()` on a `DefeatEnemy` transport error, `UI/src/lib/engine/battle/enemy-manager.ts`) re-pulls `Login/Status` and re-initializes `playerManager` (level, exp, unlocked skills), but never re-ran `inventoryManager.initialize()` — which otherwise only runs once, in `startGame`. An item/mod unlocked server-side during the outage window stayed invisible until a full page reload.

**Scenario:** the connection blips exactly as a boss kill completes a zone-gating challenge with an item reward — exp/level resync happens, but the reward item doesn't appear until reload.

## Fix

`enemy-manager.ts` now imports `inventoryManager` from the engine barrel (alongside `playerManager`, which it already pulled from there) and calls `inventoryManager.initialize()` right after `refreshPlayer()` in the resync branch — re-deriving the inventory from the freshly-loaded `playerManager.inventoryData`, mirroring what `startGame` does on the initial load.

## Test plan

- [x] Unit tests (`enemy-manager-boss.test.ts`): a successful `DefeatEnemy` claim does **not** trigger an inventory re-init (no resync needed); a failed/errored `DefeatEnemy` claim triggers both `refreshPlayer` and `inventoryManager.initialize`.
- [x] `npm run test:unit:ci` (full suite) — 3498 tests passing.

Closes #1639

---
_Generated by [Claude Code](https://claude.ai/code/session_01NHF2ZN22DKY8Qd38hkMSpe)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -1,7 +1,14 @@
 import { apiSocket, ELogType, IEnemyInstance } from '$lib/api';
 import { Action, createHook, delay, isZoneUnlocked, navigableZones, nextZoneByOrder } from '$lib/common';
 import { staticData, statistics, playerChallenges } from '$stores';
-import { battleEngine, BattleStage, onBattleStageChanged, playerManager, type PlayerBattleState } from '../';
+import {
+	battleEngine,
+	BattleStage,
+	inventoryManager,
+	onBattleStageChanged,
+	playerManager,
+	type PlayerBattleState
+} from '../';
 import { logMessage } from '../log';
 import { refreshPlayer } from '../session';
 
@@ -646,8 +653,11 @@ export class EnemyManager {
 				// The transport settles a lost/timed-out response as an error even when the command may have
 				// actually succeeded server-side (see docs/frontend.md's socket request lifecycle) — resync the
 				// authoritative player state rather than leaving exp/level silently diverged for the rest of
-				// the session.
+				// the session. refreshPlayer only re-initializes playerManager, so the inventory (an item/mod
+				// reward the lost response may have carried) is re-derived from it separately, mirroring what
+				// startGame does on the initial load.
 				await refreshPlayer();
+				inventoryManager.initialize();
 			}
 		}
 		// Guard `data` for a possible error response (absent `data`), now that this is the shared
--- a/UI/src/lib/engine/session.ts
+++ b/UI/src/lib/engine/session.ts
@@ -69,8 +69,13 @@ async function restorePlayer(): Promise<boolean> {
 /**
  * Re-pulls the authoritative player aggregate (Login/Status) and re-initializes the player manager.
  * Used by the welcome-back gate to refresh the offline-reward-updated state — level, exp, stat points,
- * unlocked skills, inventory — before the game engine builds the live battler from it. Best-effort: a
- * failed refresh leaves the existing in-memory player intact rather than stranding the player at the gate.
+ * unlocked skills, raw inventory data — before the game engine builds the live battler from it. Best-effort:
+ * a failed refresh leaves the existing in-memory player intact rather than stranding the player at the gate.
+ *
+ * This only re-initializes {@link playerManager}, not the derived `InventoryManager` — a caller that needs
+ * the player's unlocked items/mods to reflect the refresh (rather than just the raw payload) must also
+ * re-run `inventoryManager.initialize()` itself, as the mid-session resync in `EnemyManager.claimVictory`
+ * does (the welcome-back gate gets this for free from the `startGame` call that follows it).
  */
 export async function refreshPlayer(): Promise<void> {
 	await loadPlayer();
--- a/UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts
+++ b/UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts
@@ -20,6 +20,7 @@ const h = vi.hoisted(() => ({
 	},
 	BattleStage: { Idle: 0, Active: 1, Victorious: 2, Defeated: 3, Loading: 4, Paused: 5, Drawn: 6 },
 	playerManager: { currentZone: 3, applyVictoryRewards: vi.fn() },
+	inventoryManager: { initialize: vi.fn() },
 	refreshPlayer: vi.fn(() => Promise.resolve()),
 	staticData: {
 		enemies: [{ id: 0, name: 'Catacomb Lich', isBoss: true }],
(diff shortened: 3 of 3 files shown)


ginderjeremiah (commented):
## Review: Refresh InventoryManager on mid-session player resync (#1639)

**Verdict:** Approve — the fix precisely closes the gap described in the issue, mirrors the initial-load path, and is well-tested. _(Submitted as a comment since I'm the PR author; treat it as an approval.)_

Traced the change end to end and it holds up:

- `refreshPlayer()` → `loadPlayer()` → `playerManager.initialize(...)` refreshes only `playerManager` (including the raw `inventoryData` payload), but the **derived** `InventoryManager` is rebuilt solely in `startGame`. The resync path re-ran the former without the latter, so an outage-window item/mod unlock stayed invisible until reload. Adding `inventoryManager.initialize()` right after the awaited `refreshPlayer()` re-derives it from the freshly-loaded `playerManager.inventoryData` — exactly what `startGame` does.
- Ordering is correct: `refreshPlayer()` is awaited before `initialize()`, and `initialize()` is idempotent (it clears `unlockedItems`/`unlockedMods`/`equippedSlots` and rebuilds from `playerManager.inventoryData`), so re-running it mid-session is safe. It also reassigns the memoised `equipmentStats`/granted-skill caches, so the next per-spawn battle reset re-derives the player by-reference as intended (per `frontend.md`'s battle-parity reset notes) — the resync converges cleanly rather than leaving stale battler inputs.
- Scope is complete: `refreshPlayer` has exactly one caller (this resync path), so there's no sibling site left with the same gap. The `handleServerCommandFailed` case the issue alludes to is a separate channel (challenge/proficiency reloads) and out of scope here.
- `inventoryManager` is correctly pulled from the `$lib/engine` barrel alongside `playerManager`, which the file already imported from there.

- The new test asserts `inventoryManager.initialize` **was called** rather than that the inventory actually surfaces a newly-unlocked item (the literal ask in the issue's "Tests" section). Because `inventoryManager` is fully mocked in this suite, the call assertion is a reasonable proxy and the re-derivation itself is already covered by `inventory-manager`'s own tests — so this is fine as-is, just noting the slight divergence from the issue's wording.
- Pre-existing (not touched here): `refreshPlayer`'s JSDoc in `session.ts:71-74` still lists "inventory" among what it refreshes. It refreshes the raw `inventoryData` payload but not the derived `InventoryManager` — which is precisely the assumption that caused this bug. Worth tightening the wording in a future pass so a reader doesn't re-make the same assumption; not worth blocking this PR.

### Test coverage

Adequate and following the guidelines. Both branches of the shared `claimVictory` path are pinned: a successful `DefeatEnemy` claim triggers neither `refreshPlayer` nor `inventoryManager.initialize` (no resync needed), and an errored claim triggers both exactly once. The mock/hoist wiring is added cleanly alongside the existing `refreshPlayer` fixture.

### What's good

Minimal, surgical change that stays inside the existing resync seam rather than adding new plumbing. The expanded call-site comment and the test comments make the "refresh the raw payload, then re-derive the inventory" contract obvious to the next reader — good defense against a future edit dropping the second call.

---

ginderjeremiah:
## Code review — Approve

**Verdict: Approve.** The fix is correct, minimal, and well-targeted at #1639.

**What it does / correctness:** On the lost-response recovery path in `EnemyManager.claimVictory` (errored `DefeatEnemy`), the PR now calls `inventoryManager.initialize()` right after `await refreshPlayer()`, so an item/mod unlocked server-side during the outage window becomes visible without a full reload. I verified the claims against the code rather than the description:

- `inventoryManager.initialize()` is fully idempotent — it clears `unlockedItems`/`unlockedMods`, resets `equippedSlots` to a fresh array, then re-derives everything from `playerManager.inventoryData`. Safe to re-run mid-session.
- Ordering is correct: `refreshPlayer()` → `loadPlayer()` → `playerManager.initialize()` reassigns `inventoryData`, and the `await` guarantees that completes before `initialize()` reads it.
- The barrel import resolves: `inventoryManager` is the `statify(new InventoryManager())` singleton exported from `engine.ts`, re-exported through `../`. The call is synchronous (`void`), matching how `startGame` invokes it.
- Tests genuinely cover both branches (no re-init on a successful claim; both `refreshPlayer` and `inventoryManager.initialize` on an errored claim), and the mock wiring is consistent with the file's existing `vi.hoisted` style.
- CI is green (`all-checks-passed` success; frontend + all e2e browsers passing; backend jobs skipped as expected).

**Nits (non-blocking):**
- The new test asserts `initialize()` is *called*, whereas #1639's test note asked to assert the inventory actually *reflects* a newly-unlocked item. Given this file fully mocks the engine barrel, a true re-derive assertion isn't feasible here, so the call-level assertion is a reasonable compromise — just noting the slight divergence from the issue's wording.
- Minor DRY: `startGame` and the resync branch now both do "refresh player + re-init inventory," with the welcome-back gate implicitly relying on `startGame` following `refreshPlayer`. The PR documents this coupling clearly in the `refreshPlayer` JSDoc, so it's acceptable; a shared "resync player + inventory" helper (the issue's alternative) would remove the footgun if this pattern spreads.

**Out of scope (possible follow-up):** the same latent gap exists on the `handleServerCommandFailed` dead-letter path — a dead-lettered `ChallengeCompleted` push carrying a reward item reloads only `playerChallenges`, never the inventory, and its comment still claims the reward "surfaces on the next natural load of the inventory stores" (which #1639 itself flagged as untrue mid-session). Not this PR's job, but worth tracking separately.

---

ginderjeremiah:
Follow-up filed before merge: #1661 — the analogous inventory-refresh gap on the dead-lettered `ChallengeCompleted` push path (`handleServerCommandFailed`), which reloads only `playerChallenges` and would leave a challenge reward item invisible mid-session. Out of scope for this PR. Merging.

---

Issue #1625, opened 2026-07-05
verify-frontend red on main: two battle-engine.test.ts mocks miss the required IEnemyInstance.enemyRating


ginderjeremiah:
### Summary

`verify-frontend` (`npm run lint` → `svelte-check`) fails on `main` (`0d0bbda`). Codegen made `IEnemyInstance.enemyRating` a **required** field (`UI/src/lib/api/types/interfaces/enemies.ts`), and most mocks in `UI/src/tests/lib/engine/battle/battle-engine.test.ts` were updated to set `enemyRating: 100` — but two were missed, so `svelte-check` reports:

```
src/tests/lib/engine/battle/battle-engine.test.ts:510  Property 'enemyRating' is missing in type '{ id; level; seed; selectedSkills; attributes }' but required in type 'IEnemyInstance'.
src/tests/lib/engine/battle/battle-engine.test.ts:521  (same)
```

Both are the `evaluateMechanicTriggers` tests:

- `wires the tick activations into evaluateMechanicTriggers when a skill fires (#1587)`
- `does not call evaluateMechanicTriggers on a tick with no activations (nothing off cooldown yet)`

Each calls `enemyLoadedCallbacks[0]({ id: 1, level: 1, seed: 0, selectedSkills: [0], attributes: [] })` with no `enemyRating`.

### Impact

`verify-frontend` is red on `main` itself, so every open PR inherits the failure through the CI merge (surfaced while reviewing #1624, which touches none of these files). It blocks the merge queue.

### Fix

Add `enemyRating: 100` (matching the sibling mocks) to those two `enemyLoadedCallbacks` payloads. Trivial, no logic change.

ginderjeremiah:
Picked this up, but it's already resolved — no PR needed here.

The two `enemyLoadedCallbacks[0](...)` calls this issue flags (originally missing `enemyRating` on `main` @ `0d0bbda`) were fixed as a side-effect of the very next merge, PR #1624 (commit `2be5fc4`), whose second sub-commit is literally titled "Fix battle-engine.test.ts mocks missing required enemyRating (main regression)". Both call sites (`battle-engine.test.ts:510` and `:521`) now carry `enemyRating: 100`, matching the sibling mocks.

Verified against current `main` HEAD (`2be5fc4`): `npm run check` (svelte-check) reports `0 ERRORS 0 WARNINGS`.

[...]

PR #1665, opened 2026-07-06
Fix OverflowException in PlayerStatPoints.TryUpdateAttributes (#1641)

### LINKED ISSUE #1641: PlayerStatPoints.TryUpdateAttributes can throw OverflowException on a crafted payload instead of rejecting it
## Problem

`var changedPoints = matchedUpdates.Values.Sum(match => match.Update.Amount);` (`Game.Core/Players/PlayerStatPoints.cs:53`) — LINQ's `int` `Sum` uses checked arithmetic, and `Amount` is a raw client-controlled `int` (`UpdatePlayerStats` socket command → `PlayerService.TryUpdateAttributes`) with no upstream clamp. Two updates of ~1.5 billion each (targeting valid distinct core-attribute rows, so they pass the matching checks that run first) overflow the sum and throw. `availablePoints - changedPoints` on the next line can overflow the same way.

Every other invalid payload honors the documented all-or-nothing anti-cheat contract by returning `false`; this one surfaces as a generic command failure via the socket layer's catch-all plus a logged server exception. Single-update extremes are already rejected by the later comparisons — only the multi-update aggregation is unguarded.

## Fix

Use a `long` accumulator (and compare against the available points in `long`), or clamp per-update magnitudes before summing — either way return `false` per the contract.

## Tests

Unit test: two near-`int.MaxValue` updates on valid rows → `false`, no mutation, no throw.

ginderjeremiah:
## Summary

- `PlayerStatPoints.TryUpdateAttributes` (`Game.Core/Players/PlayerStatPoints.cs`) summed the batch of stat-point updates with `Enumerable.Sum` over `int`, which uses **checked** arithmetic. A crafted payload with two updates of ~1.5 billion each — each individually valid enough to pass the earlier attribute-matching checks — overflowed the sum and threw `OverflowException` instead of being rejected via the documented all-or-nothing anti-cheat contract.
- Switched the aggregation (`changedPoints`) and the available-points comparison to `long`, so the sum can't overflow and an out-of-range payload is rejected the normal way (`return false`, no mutation). The per-item `Allocation.Amount + Update.Amount >= 0` check was also widened to `long` for consistency, though that check was not the one throwing (plain `+` doesn't use checked arithmetic).
- `StatPointsUsed += (int)changedPoints` stays safe: by that point the code has already verified `changedPoints <= availablePoints`, and `availablePoints` is derived from existing `int` fields, so the cast back can't lose information.

## Test plan

- Added `TryUpdateAttributes_NearIntMaxValueUpdates_RejectsWithoutOverflowing`: two updates of 1.5 billion each → `false`, no mutation, no throw.
- `dotnet build` — 0 warnings, 0 errors.
- `dotnet test --no-build --no-restore` — full backend suite green (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests all passed).

Closes #1641

## Note

While picking up work in this backlog sweep, I also found #1625 (`verify-frontend red on main`) was already fixed by a prior, unrelated merge (#1624) — verified against current `main` and closed it as completed; no PR needed there.

---
_Generated by [Claude Code](https://claude.ai/code/session_015naWqXZpi1vfwtSY5jr2RW)_

### DIFF
--- a/Game.Core.Tests/Players/PlayerStatPointsTests.cs
+++ b/Game.Core.Tests/Players/PlayerStatPointsTests.cs
@@ -126,6 +126,24 @@ namespace Game.Core.Tests.Players
         }
 
         [Fact]
+        public void TryUpdateAttributes_NearIntMaxValueUpdates_RejectsWithoutOverflowing()
+        {
+            // Two updates whose sum overflows int (checked LINQ Sum would throw OverflowException)
+            // must still be rejected via the normal all-or-nothing contract, not crash.
+            var stats = MakeStats(gained: 10, used: 0);
+
+            var result = stats.TryUpdateAttributes([
+                new Update(EAttribute.Strength, 1_500_000_000),
+                new Update(EAttribute.Endurance, 1_500_000_000),
+            ]);
+
+            Assert.False(result);
+            Assert.Equal(0, stats.StatPointsUsed);
+            Assert.Equal(0, stats.StatAllocations.First(a => a.Attribute == EAttribute.Strength).Amount);
+            Assert.Equal(0, stats.StatAllocations.First(a => a.Attribute == EAttribute.Endurance).Amount);
+        }
+
+        [Fact]
         public void TryUpdateAttributes_UnknownAttribute_RejectsWithoutMutating()
         {
             // An update targeting an attribute the player has no allocation row for is rejected rather
--- a/Game.Core/Players/PlayerStatPoints.cs
+++ b/Game.Core/Players/PlayerStatPoints.cs
@@ -50,12 +50,15 @@ namespace Game.Core.Players
                 }
             }
 
-            var changedPoints = matchedUpdates.Values.Sum(match => match.Update.Amount);
-            var availablePoints = StatPointsGained - StatPointsUsed;
+            // long accumulator: matchedUpdates.Values.Sum for int uses checked arithmetic and would
+            // throw OverflowException on a crafted payload (e.g. two near-int.MaxValue updates)
+            // instead of honoring the all-or-nothing reject contract.
+            var changedPoints = matchedUpdates.Values.Sum(match => (long)match.Update.Amount);
+            var availablePoints = (long)StatPointsGained - StatPointsUsed;
             if (availablePoints - changedPoints >= 0
-                && matchedUpdates.Values.All(match => match.Allocation.Amount + match.Update.Amount >= 0))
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Fix OverflowException in `PlayerStatPoints.TryUpdateAttributes` (#1641)

**Verdict:** Approve — the fix correctly resolves the overflow via `long` aggregation while preserving the all-or-nothing reject contract, with a test that reproduces the exact crafted payload from the issue. (Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a self-approval; the intended verdict is **Approve**.)

### Correctness

The fix is sound and complete. Walking the arithmetic:

- `changedPoints` sums `(long)Update.Amount` over the matched rows — with at most a handful of distinct core attributes, each an `int`, the sum can't approach `long` overflow. The `OverflowException` from `int` `Sum`'s checked arithmetic is gone.
- `availablePoints = (long)StatPointsGained - StatPointsUsed` and `availablePoints - changedPoints >= 0` are now fully in `long`, so the second overflow site the issue flagged is closed too.
- Widening the per-item `Allocation.Amount + (long)Update.Amount >= 0` guard to `long` is a genuine improvement, not just cosmetic: the old `int + int` could silently wrap (unchecked) and mis-evaluate the guard. Good catch to fold it in.
- `StatPointsUsed += (int)changedPoints` is safe in both directions: the passed `availablePoints - changedPoints >= 0` check bounds `changedPoints <= StatPointsGained <= int.MaxValue`, and the per-item `>= 0` guard bounds it below by `-StatPointsUsed >= -int.MaxValue`. It can't truncate. The PR description reasons through the upper bound; the lower bound holds for the same per-item reason.

### Test coverage

- `TryUpdateAttributes_NearIntMaxValueUpdates_RejectsWithoutOverflowing` reproduces exactly the issue's scenario (two ~1.5B updates on valid distinct rows) and asserts `false`, no mutation on either row, and no throw. That matches the issue's requested test precisely.
- Backend-only change to stat-point allocation — no battle logic, so no frontend mirror is required.

### What's good

- Minimal, surgical fix that stays within the domain layer (`Game.Core`) where it belongs.
- The 3-line comment on the `long` accumulator is justified: it explains *why* `int` can't be used, which is exactly the context a future "simplification" back to `int` would need to avoid reintroducing the bug.

No blocking findings, no should-fixes, no nits. Clean fix.


---

PR #1666, opened 2026-07-06
Cap DoT-absorption healing at MaxHealth for consistency (#1642)

### LINKED ISSUE #1642: DoT absorption heals past MaxHealth, contradicting the no-overheal invariant — decide cap vs document (parity-sensitive)
## Problem

The engine's two other healing channels cap at MaxHealth with an explicit "no overheal/shield concept" rationale: direct-hit absorption in `TakeDamage` (`Game.Core/Battle/Battler.cs:357-373`) and `ApplyHealOverTime`. But `ApplyDamageOverTime` does not — `CurrentHealth -= dot` with no clamp (`Battler.cs:509`), so a DoT-afflicted battler with type resistance above 1 (absorption) gets a negative tick that raises `CurrentHealth` **above** MaxHealth, banked until the next effect apply/expiry incidentally re-clamps via `ClampHealthToMaxHealth`. Mirrored in `UI/src/lib/battle/battler.ts:211-222`.

Tests pin the uncapped negative tick (`BattleDamageOverTimeTests` asserts a below-max case) but never the above-max case, so the contradiction between the channels is unadjudicated. game-design.md documents absorption "falling out for free" for DoT but also that HoT is "capped at MaxHealth" — the invariant is inconsistent across channels.

## Ask

Decide: cap the DoT-absorption heal at MaxHealth like the other two channels, or bless transient overheal and document it. If capping, the change is **parity-critical** — it must land in `Battler.cs` and `battler.ts` simultaneously with mirrored test scenarios per the battle-parity rule (including the incidental `ClampHealthToMaxHealth` interaction).

ginderjeremiah:
## Summary

`Battler.TakeDamage`'s direct-hit absorption (resistance driving net damage negative) and `Battler.ApplyHealOverTime` both cap the resulting heal at `MaxHealth`, with an explicit "no overheal/shield concept" rationale in their doc comments. `Battler.ApplyDamageOverTime`'s absorption channel (a DoT tick with type-resistance above `1`) did not — a battler could be healed above `MaxHealth`, banked until the next incidental `ClampHealthToMaxHealth` call (an effect apply/expiry).

## Decision

The issue asked to decide: cap it, or bless the transient overheal and document it. I chose to **cap**, for consistency with the other two channels' explicit invariant, since both existing doc comments and code already establish "no overheal/shield concept" as the game's rule. This is a narrower fix than it might first appear:

- Each DoT type's own tick is still **not floored at zero** — a deliberately authored negative accumulator or a resistance above `1` still produces a negative per-type tick, and the per-type statistics recorders (`recordDamageDealt`, `recordMitigated`, `recordExposure`, `recordHexBonus`) still see that **uncapped** value, so the #1482 overkill-booking behavior and existing pinned tests are unaffected.
- Only the **aggregate** health change the call realizes across all typed ticks is capped at the remaining room to `MaxHealth`, mirroring `TakeDamage`'s exact pattern (compute room, cap the heal, adjust the returned value to match reality).

## Changes

- `Game.Core/Battle/Battler.cs`: `ApplyDamageOverTime` caps the aggregate negative `dot` at the remaining room to `MaxHealth` before applying it to `CurrentHealth`; updated doc comments.
- `UI/src/lib/battle/battler.ts`: mirrored fix in `applyDamageOverTime`, per the battle-parity rule (guarded against returning `-0`, since JS's strict equality distinguishes it from `0` unlike C#'s `double`).
- Tests added in both `Game.Core.Tests/Battle/BattleDamageOverTimeTests.cs` and `UI/src/tests/lib/battle/battler-dot.test.ts`: capping at zero room (already at `MaxHealth`) and capping at partial room, plus a backend assertion that per-type booking stays uncapped even when the aggregate is capped.
- `docs/backend-battle.md` and `docs/game-design.md`: updated the DoT phase description, which previously stated absorption "heals uncapped," to describe the new capped-aggregate/uncapped-per-type behavior.

## Test plan

- [x] `dotnet test` (full solution: Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests) — all passing
- [x] `npm run test:unit:ci` — 295 files / 3503 tests passing

Closes #1642

---
_Generated by [Claude Code](https://claude.ai/code/session_01WqkVD69xh2DzToDEoUGtHU)_


ginderjeremiah (commented):
## Review: Cap DoT-absorption healing at MaxHealth (#1642)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating event from me) — the fix resolves the adjudicated inconsistency cleanly, on both engines, with mirrored tests and docs. No blocking or should-fix items.

### What the change does well

- **Solves the stated problem and makes the right call.** #1642 asked to *decide* cap-vs-document; capping is the defensible choice given both sibling channels (`TakeDamage` absorption, `ApplyHealOverTime`) already establish "no overheal/shield concept" as the invariant in code and docs. The fix ends the cross-channel contradiction rather than papering over it.
- **Correctly scoped to the aggregate.** The cap is applied to the summed `dot` *after* the type loop, so the per-type recorders (`recordDamageDealt`/`recordMitigated`/`recordExposure`/`recordHexBonus`) still see the uncapped tick — the #1482 overkill-booking behavior and the pinned parity tests are untouched. The `AtFullHealth_CapsHealAtZero` test explicitly asserts `dealtByType[Bleed] == -80` while `dealt == 0`, which pins exactly this distinction.
- **Mirrors the existing pattern precisely.** `Battler.cs:512-522` is a line-for-line analog of `TakeDamage`'s absorption cap (`Battler.cs:360-369`) — compute room, cap the heal, adjust the returned value to reality — so there's no new shape for a maintainer to reason about.
- **Parity is respected.** `battler.ts:222-228` mirrors the backend, and the `ClampHealthToMaxHealth` interaction the issue flagged is now moot by construction (health can no longer exceed max in this channel). I traced the consumers (`BattleContext.ResolveDamageOverTime` ↔ `battle-step.ts:260-275`): the capped return flows into `PlayerDamageTaken`/`PlayerDamageDealt`/`PlayerDamageHealed`, and this is strictly more correct than before (the healed stat is now bounded).
- **Docs updated in the right places** — `backend-battle.md` and `game-design.md` both had the old "heals uncapped" wording, and both are corrected to the capped-aggregate/uncapped-per-type description. No stale references remain.

### Test coverage

Meets the project's testing bar. Zero-room and partial-room cases are added on **both** sides (`BattleDamageOverTimeTests.cs` and `battler-dot.test.ts`) per the battle-parity rule, plus the backend assertion that per-type booking stays uncapped under the aggregate cap. This is the right set of scenarios for the change.

### Nits (optional, no functional impact)

- `Battler.cs:521` returns `dot = -heal`, which yields `-0.0` in the zero-room case, whereas the frontend deliberately normalizes to `0` (`battler.ts:227`) to satisfy JS strict equality. In C# `-0.0 == 0.0` and every downstream use is numeric (`Stats.PlayerDamageTaken += -0.0` is a no-op), so this is purely cosmetic asymmetry between the two engines rather than a defect — noting it only for completeness. No change needed unless you want the two returns bit-identical.

Holding off on merging only because `test-backend` was still in progress at review time (every other check — all e2e browsers, frontend, and both verify jobs — is green).


---

ginderjeremiah:
## Code review — Approve ✅

Reviewed against `CLAUDE.md`, `docs/game-design.md`, `docs/backend-battle.md`, and the frontend/backend battle-parity rule.

**What this does:** Caps the net heal from DoT absorption (a DoT tick whose type-resistance exceeds `1`, driving the tick negative) at the battler's remaining room to `MaxHealth`. This closes the one healing channel — `ApplyDamageOverTime` — that previously let `CurrentHealth` bank above `MaxHealth` until an incidental `ClampHealthToMaxHealth` on the next effect apply/expiry. Issue #1642 offered "cap or bless+document"; capping is the right call, since `TakeDamage` and `ApplyHealOverTime` already establish the explicit "no overheal/shield concept" invariant.

**Verification:**
- **C#/TS parity confirmed** — both compute `room = MaxHealth − CurrentHealth`, `heal = clamp(min(−dot, room), 0)`, then re-derive `dot`, and apply the same `CurrentHealth -= dot`. Behaviorally identical; matches the existing `TakeDamage` shape verbatim.
- **Per-type booking stays uncapped** — the cap is applied only to the aggregate `dot` after the loop, so `recordDamageDealt`/`recordMitigated`/`recordExposure`/`recordHexBonus` still see the uncapped per-type tick. This preserves the #1482 overkill behavior, and the backend test explicitly asserts `dealtByType[Bleed] == -80` while the aggregate returns `0` — exactly the right invariant to pin.
- **The `-0` guard is genuinely necessary** on the JS side (`Object.is(-0, 0)` is false, so `expect(dealt).toBe(0)` would fail without it); mirrored on the C# side so the returned value is bit-identical across sides.
- **Test scenarios match across both suites** (zero-room and partial-room), with identical stats and expected values; arithmetic checks out.
- **Docs updated accurately** in both `backend-battle.md` and `game-design.md`.
- **CI green.**

**Nit (non-blocking):** the cap-to-room block now appears in three channels (`TakeDamage`, `ApplyDamageOverTime`, `ApplyHealOverTime`) on both sides. A small shared helper would prevent future drift. This spans pre-existing code beyond this PR, so I've filed it as a follow-up rather than expanding scope here: **#1671** (`tech debt`, `scope: small`), which also notes the optional parity-matrix row.

Correct, minimal, parity-clean, well-tested. Merging.

---

PR #1667, opened 2026-07-06
Escape ILIKE wildcards in admin user search

### LINKED ISSUE #1643: Admin user search doesn't escape ILIKE wildcards
## Problem

`FilteredUsers` (`Game.DataAccess/Repositories/Users.cs:395-400`) builds `$"%{search}%"` from raw admin input, so `%` and `_` are interpreted as wildcards: searching `%` matches every user, and a literal `%`/`_`/`\` in a username or player name can't be searched for. Parameterized (no injection risk), admin-only surface — but the results are wrong.

## Fix

Escape `\`, `%`, and `_` in the search term before building the pattern (and pass the escape character consistently, e.g. via `ILIKE ... ESCAPE '\'` semantics EF's `ILike` overload supports).

## Tests

Integration test: a username containing a literal `%`/`_` is findable, and a bare `%` search doesn't return everything.

ginderjeremiah:
## Summary

`FilteredUsers` (`Game.DataAccess/Repositories/Users.cs`) built its `ILIKE` search pattern directly from raw admin input (`$"%{search}%"`), so `%` and `_` in the search term were interpreted as `ILIKE` wildcards rather than matched literally: a bare `%` search matched every user, and a username/player name containing a literal `%`, `_`, or `\` couldn't be searched for. Parameterized (no injection risk), admin-only surface — but the results were wrong.

## Fix

- Added `EscapeLikePattern`, which escapes `\`, `%`, and `_` in the raw search term (backslash first, so it doesn't double-escape the characters it introduces).
- Passed the escape character through to `EF.Functions.ILike(matchExpression, pattern, escapeCharacter)` for both the username and player-name predicates, so the escaped sequences are honored by Postgres.

## Tests

Added two integration tests to `AdminUsersControllerTests`:
- A username containing a literal `%` is found by that exact term and doesn't cause every user to match.
- A username containing a literal `_` is found by that exact term, without matching a similarly-named user that merely has some other character in that position.

## Test plan

- [x] `dotnet build` (whole solution) — 0 warnings, 0 errors
- [x] `dotnet test` (whole solution, against the containerized Postgres/Redis) — 2980 tests passing
- [x] New tests specifically verified passing (`Game.Api.Tests -- --filter-class "*AdminUsersControllerTests"`)

Closes #1643

---
_Generated by [Claude Code](https://claude.ai/code/session_016E17wLuMc1kxnbDbaQkqJ8)_

### DIFF
--- a/Game.Api.Tests/Integration/AdminUsersControllerTests.cs
+++ b/Game.Api.Tests/Integration/AdminUsersControllerTests.cs
@@ -172,6 +172,23 @@ namespace Game.Api.Tests.Integration
         }
 
         [Fact]
+        public async Task GetUsers_SearchTermWithLiteralBackslash_MatchesOnlyThatUser()
+        {
+            using var authClient = await SetupAdminClientAsync();
+            using (var scope = CreateScope())
+            {
+                var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+                await TestDataSeeder.CreateUserAsync(context, @"back\slash", "pw");
+                await TestDataSeeder.CreateUserAsync(context, "someoneelse", "pw");
+            }
+
+            // The escape character itself must be escaped first, or a literal "\" in the search term
+            // would corrupt the pattern's own escaping of "%"/"_" (or fail the query outright).
+            var results = await GetUsersAsync(authClient, "?search=back%5Cslash");
+            Assert.Equal(new[] { @"back\slash" }, results.Users.Select(u => u.Username));
+        }
+
+        [Fact]
         public async Task GetUsers_FilterByRole_ReturnsOnlyUsersInRole()
         {
             using var authClient = await SetupAdminClientAsync();
--- a/Game.Api.Tests/Integration/AdminUsersControllerTests.cs
+++ b/Game.Api.Tests/Integration/AdminUsersControllerTests.cs
@@ -139,6 +139,39 @@ namespace Game.Api.Tests.Integration
         }
 
         [Fact]
+        public async Task GetUsers_SearchTermWithLiteralPercent_MatchesOnlyThatUserAndDoesNotMatchEveryone()
+        {
+            using var authClient = await SetupAdminClientAsync();
+            using (var scope = CreateScope())
+            {
+                var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+                await TestDataSeeder.CreateUserAsync(context, "100%legit", "pw");
+                await TestDataSeeder.CreateUserAsync(context, "someoneelse", "pw");
+            }
+
+            // A literal "%" in the search term must not act as an ILIKE wildcard: it should find only
+            // the user whose name actually contains it, not every user (as an unescaped "%...%" would).
+            var results = await GetUsersAsync(authClient, "?search=100%25legit");
+            Assert.Equal(new[] { "100%legit" }, results.Users.Select(u => u.Username));
+        }
+
+        [Fact]
+        public async Task GetUsers_SearchTermWithLiteralUnderscore_MatchesOnlyThatUser()
+        {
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Escape ILIKE wildcards in admin user search (#1643)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating event from me) — the fix resolves the wrong-results bug cleanly and correctly, with the escape character threaded through both predicates and integration tests matching the issue's acceptance criteria.

### What the change does well

- **Solves the stated problem.** `EscapeLikePattern` escapes `\`, `%`, and `_`, and — critically — escapes the backslash *first*, so it doesn't double-escape the `\` sequences it then introduces for `%`/`_`. The escaped pattern is passed to `EF.Functions.ILike(..., LikeEscapeCharacter)` for both the username and player-name predicates, so Postgres honors the `ESCAPE '\'` semantics. A bare `%` search now matches literally instead of everyone.
- **Correctly scoped.** The outer `%…%` wildcards in `$"%{EscapeLikePattern(search)}%"` are preserved (still a substring match), while only the user-supplied inner text is escaped. That's exactly the intended behavior.
- **Right layer.** This is a query-construction concern living in the `Users` repository — no domain logic leaks in, consistent with `backend.md`'s data-access boundaries. Verified this is the only `ILike`/`LIKE` usage in production code (`Users.cs`), so there's no sibling site carrying the same bug and no DRY follow-up.
- **DRY / house rules.** The escape char is a single `const`, reused for both the pattern build and both predicates; comments are concise; no unbraced blocks or `!` introduced.

### Test coverage

Meets the bar for this change. Integration tests (the correct kind, since this exercises real Postgres `ILIKE` behavior) cover the two acceptance criteria: a literal `%` finds only that user and does not match everyone, and a literal `_` doesn't behave as a single-char wildcard. Good use of `%25` URL-encoding to drive the literal `%` through the query string.

### Nits (optional — take or leave)

- No test exercises a literal `\` in the search term. Since the escape function's correctness hinges on the ordering (`\` replaced before `%`/`_`), a "username containing a literal backslash is findable by that exact term" case would guard against an accidental reorder regression. Truly optional — the two wildcard chars are the practical concern and are covered.

Not merging yet: `test-backend` (the relevant suite) and `test-e2e (webkit)` were still in progress at review time — everything else is green or appropriately skipped.

---

ginderjeremiah:
## Code review — Approve ✅

Reviewed against `CLAUDE.md` and `docs/backend.md`.

**What this does:** `FilteredUsers` built its ILIKE pattern (`%{search}%`) from raw admin input, so `%`/`_` in a search term acted as wildcards — a bare `%` matched every user and names containing a literal `%`/`_`/`\` were unsearchable. The fix adds a private `EscapeLikePattern` helper and passes the escape character through the three-arg `EF.Functions.ILike(col, pattern, escape)` overload for both the username and player-name predicates.

**Verification:**
- Escape ordering is correct — `\` is doubled first, then `%`/`_`, so the introduced escape chars aren't re-escaped.
- Escape char is plumbed through to the `ESCAPE`-clause overload. (Postgres already defaults to `\`, so this is belt-and-suspenders, but being explicit is the right call.)
- Tests are sound: each URL-encodes correctly and asserts both the positive match and a negative decoy, which is exactly what pins the bug.
- Domain isolation is respected — LIKE-pattern escaping is a persistence concern and lives correctly as a private helper in the data tier. This is the only `EF.Functions.Like/ILike` call site in the codebase, so the fix is complete with no duplicated bug elsewhere.

**Nit (non-blocking, description only):** the summary says "two integration tests" but the diff adds three (`%`, `_`, and `\`). The backslash test is a good addition — it's what proves the escape-first ordering. No code change needed.

Clean, minimal, correct. CI green. Merging.

---

PR #1672, opened 2026-07-06
Offline sim: run proficiency accrual in-loop, move the math into Game.Core

### LINKED ISSUE #1602: Offline sim: mid-window proficiency growth — run accrual in-loop and move the math into Game.Core

ginderjeremiah:
## Summary

Implements #1602 (depends on #1601, merged): the offline simulation must model everything that grows automatically in live play, including proficiency levels and their attribute payouts, not just player level.

- **`Game.Core.Proficiencies.ProficiencyAccrual`** — a new pure domain function extracted from the application layer's `ProficiencyRewardService`: the frontier routing, XP claim (`ProficiencyXpCalculator`), leveling, and milestone/open triggers, driven entirely by a small `ProficiencyCatalog` resolver-func bundle plus read/write funcs for a proficiency's current `(level, xp)` — no `IProficiencies`/`ISkills`/`PlayerProgress`/`Player` dependency. Milestone reward-skill grants are *reported* (`ProficiencyXpResult.GrantedSkillIds`), not applied, keeping the function free of any aggregate coupling.
- **`ProficiencyRewardService` is now a thin adapter** over `ProficiencyAccrual` for the live path: it bundles `IProficiencies` into the resolvers, writes through the `PlayerProgress` aggregate, grants the reported reward skills (`GrantRewardSkills`, now also reused by the offline path), and raises the live client push.
- **`OfflineProgressSimulator` runs the accrual in-loop**: immediately after each victory, against its own working copy of the snapshot's captured proficiency levels (deep-copied per run so growth never leaks onto the caller's frozen snapshot). A milestone crossed early in the away window is therefore already baked into every later simulated battle's attribute bonuses — unlike before, where the whole window fought at window-start proficiency levels because the accrual only ran after the simulation had already concluded. The rating battler measuring `DefeatRewards` (and through it every path's own `ratingDenominator`) is re-derived whenever a proficiency level-up actually changes an attribute modifier, exactly like a player level-up already does — so the "power denominator" other paths' accrual normalizes by grows with proficiency too, not just level.
- **`OfflineProgressService.ApplyOfflineRewards`** no longer re-runs the accrual per battle against the real aggregate; it folds each battle's already-computed result (`ProficiencyGainAccumulator`, unchanged) and applies the window's final per-proficiency level/XP plus skill grants once, through the same `GrantRewardSkills` step the live path uses.
- `BattleSnapshot.ProficiencyLevelSnapshot` gained an `Xp` field (residual XP within the captured level) so the offline in-loop growth has a real starting point, mirroring how `StartingExp` seeds level growth.
- Docs updated: `docs/backend.md`'s proficiency-accrual section, `docs/backend-battle.md`'s offline-rewards section, and `docs/game-design.md`'s proficiency/offline sections — all previously noted #1602 as a pending follow-up.

## Note on scope

This issue's stated scope also proposed swapping `ProficiencyRewardService`'s `IProficiencies`/`ISkills` dependencies for the resolver-func pattern *entirely*. I kept the app-layer service's constructor on `IProficiencies`/`ISkills` (DI-friendly, unchanged public surface) and had it build the `ProficiencyCatalog` bundle internally — the resolver-func pattern applies to the new `Game.Core` domain function itself, which is the part that actually needed to be data-access-free to run inside the simulator. This seemed like the right reading of the intent (keep `Game.Core` free of `IProficiencies`) without forcing an awkward DI change on the application-layer adapter.

## Test plan

- [x] `Game.Core.Proficiencies.ProficiencyAccrual` — new unit tests (`Game.Core.Tests/Proficiencies/ProficiencyAccrualTests.cs`): milestone crossing + granted-skill reporting, bonus-only milestone, opening the next tier, cross-path gateway open/non-open, retired-path-equivalent (absent from catalog) accrues nothing, locked/unlocked umbrella gating, resist-key training from exposure, non-positive rating denominator and trivial-activity rounding both bank nothing.
- [x] `OfflineProgressSimulatorTests` — new tests per the issue's explicit asks:
  - Same-sequence parity: replaying the simulator's own recorded battle stats through `ProficiencyAccrual.Accrue` one battle at a time (mirroring the live per-battle path) lands on the identical final state.
  - Mid-window proficiency growth flips a threshold matchup from losses to wins (mirroring the #1601 level-growth flip test).
  - Resist accrual shrinks as an unrelated offense path's payout grows the player's rating mid-window — proving the "power denominator" is no longer frozen at window-start power.
  - All existing offline-simulator tests updated for the new required `OfflineSimulationParameters` fields and pass unchanged otherwise.
- [x] Existing DB-backed `ProficiencyRewardServiceTests`/`OfflineProgressServiceIntegrationTests` pass unchanged — confirms the thin-adapter refactor and the rewritten offline apply step are wire-compatible end-to-end (including a DB-reload assertion that the persisted final state matches the folded summary).
- [x] `dotnet test` — full suite green: Game.Core.Tests 1344, Game.Application.Tests 886, Game.Api.Tests 710, Game.Infrastructure.Tests 59.

Closes #1602

---
_Generated by Claude Sonnet 5_

---
_Generated by [Claude Code](https://claude.ai/code/session_01SK1qgoFNaJxJtYdjpqNULq)_


ginderjeremiah (commented):
## Review: Offline sim — run proficiency accrual in-loop, move the math into `Game.Core` (#1602)

**Verdict:** Approve (posted as a comment since I'm the PR author) — solves the issue exactly as scoped, the DDD layering is clean, the persisted-state invariant is preserved and pinned by tests, and coverage hits all three issue-mandated scenarios.

Reviewed the full diff against `CLAUDE.md`, `docs/backend.md`, `docs/backend-battle.md`, and `docs/game-design.md`.

### Correctness / design

- **Domain math lands in `Game.Core`.** `ProficiencyAccrual` is a pure function driven entirely by the `ProficiencyCatalog` resolver bundle plus `levelOf`/`xpOf`/`setProgress` funcs — no `IProficiencies`/`ISkills`/`PlayerProgress`/`Player` coupling. That's the right home for domain logic per the backend DDD guidelines, and it's what makes the in-loop simulator use possible. `ProficiencyRewardService` correctly collapses to a thin application-layer adapter (bundles `IProficiencies` into the catalog, writes through the aggregate, grants the reported reward skills, raises the push).
- **The offline/live parity invariant holds.** The simulator accrues in-loop against its own deep-copied working state; `OfflineProgressService` folds each battle's already-computed result and writes the window's final absolute level/XP once via `SetProficiencyProgress` (absolute, so no double counting) plus a single `GrantRewardSkills`. Because `ProficiencyGainAccumulator` keeps the last touch per proficiency and the accrual only emits a result when it actually wrote, the folded final state provably equals the simulator's own working state — and the existing DB-backed `OfflineProgressServiceIntegrationTests.SimulateOfflineProgress_AccruesProficiencyXpFromFiredSkills` reloads from Postgres and asserts `stored.Level/Xp == gain.NewLevel/NewXp`, so the refactor is verified wire-compatible end-to-end.
- **Rating denominator grows mid-window.** Re-deriving the rating battler on `progression.LevelsGained > 0 || proficiencyLeveledUp` mirrors the #1601 level-growth handling, and `result.Stats.PlayerRating = rewards.PlayerRating` matches what live `BattleService` does for the statistics-recording path — so the denominator (`max(PlayerRating, EnemyRating)`, #1526 Decision 5) is sourced identically to `BattleStatisticsEventHandler`.
- **The `BattleSnapshot.Xp` addition + `CloneForGrowth` deep-copy** are the right call — the working proficiency levels have to grow without leaking onto the caller's frozen snapshot, and the rationale is well documented at the copy site.
- **Scope deviation is sound.** Keeping `ProficiencyRewardService`'s constructor on `IProficiencies`/`ISkills` (building the catalog internally) rather than forcing the resolver-func pattern onto the DI-visible application service is the correct reading of the issue's intent — the part that *needed* to be data-access-free is the new `Game.Core` function, and it is.

### Test coverage

- New `ProficiencyAccrualTests` — classical-style, no doubles, covering frontier leveling, trivial-activity rounding, milestone skill reporting without side effects, within-path and cross-path gateway opening, umbrella gating, resist-from-exposure, and retired-path-absent — a thorough pure-domain suite.
- New `OfflineProgressSimulatorTests` cover all three issue-mandated scenarios: same-sequence live-vs-offline parity, a mid-window milestone flipping a threshold matchup from losses to wins, and resist accrual shrinking as the offense payout grows the shared rating denominator. The flakiness reasoning on the flip test (requiring ~20 free wins so an unlucky roll ordering can't produce a false negative) is a nice touch.
- Existing DB-backed integration/reward-service tests pass unchanged, confirming the thin-adapter + fold-and-apply refactor stays behavior-compatible.

### Docs

All three docs updated correctly and the stale "#1602 is a committed follow-up" notes removed. `backend.md`'s split of the live-adapter vs. offline-direct call paths is clear and accurate.

### What's good

Genuinely clean extraction: the layering is textbook, the invariant is both argued in comments and pinned by a DB-reload test, and the dense rationale comments earn their length by capturing subtle offline/live parity invariants a later edit could easily break. Nice work.


---

PR #1676, opened 2026-07-06
TickSource: fall back to setInterval when the worker fails to start

### LINKED ISSUE #1612: TickSource: fall back to setInterval when the worker fails to start
### Context

Follow-up from PR #1610 (issue #1594), which introduced `createTickSource` in `UI/.../tick-source.ts`. When a `Worker` is available it drives the logical-engine tick from a dedicated module worker; when `Worker` is undefined (SSR/jsdom) it falls back to `window.setInterval`.

### Gap

The worker's `onerror` handler currently only **logs** — it does not fall back to `setInterval`. If the worker ever fails to construct or load (e.g. the bundled worker chunk 404s, a CSP blocks it, or the worker throws on init), ticking would stop entirely, including while the tab is foregrounded, with no recovery.

The practical risk is low: the worker is a same-origin, Vite-emitted chunk and the real-browser e2e suites exercise it successfully. But a silent, unrecoverable stall of the game's tick loop is a bad failure mode to leave latent.

### Ask

On worker `onerror` (at least during the initial construction/first-message window), tear down the failed worker and transparently fall back to the `window.setInterval` tick source, so a worker failure degrades to the existing fallback rather than halting ticks. Add a unit test covering the fallback-on-error path (the existing test already mocks the `Worker` class, so an erroring mock is straightforward).

ginderjeremiah:
## Summary

- `createTickSource`'s worker `onerror` handler (`UI/src/lib/engine/tick-source.ts`) previously only logged the failure — if the dedicated tick worker ever failed to construct or load, the logical engine's tick loop would stall silently and permanently, even in a foregrounded tab.
- `onerror` now tears down the failed worker (`terminate()`) and transparently starts the existing `window.setInterval` fallback, so a worker failure degrades gracefully instead of halting ticks. `stop()` cleans up whichever of the two (worker or fallback interval) ended up active.

## Test plan

- [x] Added a unit test (`UI/src/tests/lib/engine/tick-source.test.ts`) that fires the mocked worker's `onerror`, asserts the failed worker is torn down, and asserts ticks continue via the `setInterval` fallback (and stop cleanly on `source.stop()`).
- [x] `npm run lint` — 0 errors, 0 warnings
- [x] `npm run test:unit:ci` — 3506/3506 passing

Closes #1612

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lgmaa7k8xbneY3nU8h36yd)_

### DIFF
--- a/UI/src/lib/engine/tick-source.ts
+++ b/UI/src/lib/engine/tick-source.ts
@@ -11,11 +11,26 @@ export interface TickSource {
 
 export function createTickSource(onTick: () => void): TickSource {
 	if (typeof Worker !== 'undefined') {
-		const worker = new Worker(new URL('./tick-worker.ts', import.meta.url), { type: 'module' });
+		let worker: Worker | null = new Worker(new URL('./tick-worker.ts', import.meta.url), {
+			type: 'module'
+		});
+		let fallbackHandle: number | null = null;
+
 		worker.onmessage = () => onTick();
-		worker.onerror = (ev) => console.error('The logical engine tick worker failed', ev);
+		worker.onerror = (ev) => {
+			console.error('The logical engine tick worker failed, falling back to setInterval', ev);
+			worker?.terminate();
+			worker = null;
+			fallbackHandle = window.setInterval(onTick, pollingIntervalMs);
+		};
+
 		return {
-			stop: () => worker.terminate()
+			stop: () => {
+				worker?.terminate();
+				if (fallbackHandle !== null) {
+					window.clearInterval(fallbackHandle);
+				}
+			}
 		};
 	}
 
--- a/UI/src/lib/engine/tick-source.ts
+++ b/UI/src/lib/engine/tick-source.ts
@@ -18,8 +18,11 @@ export function createTickSource(onTick: () => void): TickSource {
 
 		worker.onmessage = () => onTick();
 		worker.onerror = (ev) => {
+			if (worker === null) {
+				return;
+			}
 			console.error('The logical engine tick worker failed, falling back to setInterval', ev);
-			worker?.terminate();
+			worker.terminate();
 			worker = null;
 			fallbackHandle = window.setInterval(onTick, pollingIntervalMs);
 		};
@@ -27,6 +30,7 @@ export function createTickSource(onTick: () => void): TickSource {
 		return {
 			stop: () => {
 				worker?.terminate();
+				worker = null;
 				if (fallbackHandle !== null) {
 					window.clearInterval(fallbackHandle);
 				}
--- a/UI/src/tests/lib/engine/tick-source.test.ts
+++ b/UI/src/tests/lib/engine/tick-source.test.ts
@@ -72,4 +72,25 @@ describe('createTickSource', () => {
 		expect(errorSpy).toHaveBeenCalledTimes(1);
 		errorSpy.mockRestore();
 	});
+
+	it('falls back to window.setInterval when the worker fails to start', () => {
+		vi.stubGlobal('Worker', MockWorker);
+		vi.spyOn(console, 'error').mockImplementation(() => {});
+		vi.useFakeTimers();
+
+		const onTick = vi.fn();
+		const source = createTickSource(onTick);
+		const worker = MockWorker.instances[0];
+
+		worker.onerror?.(new ErrorEvent('error', { message: 'boom' }));
+		expect(worker.terminate).toHaveBeenCalledTimes(1);
+
+		vi.advanceTimersByTime(pollingIntervalMs * 3);
+		expect(onTick.mock.calls.length).toBeGreaterThanOrEqual(2);
+
+		const callCount = onTick.mock.calls.length;
+		source.stop();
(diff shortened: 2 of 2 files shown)


UI/src/lib/engine/tick-source.ts
```diff
@@ -11,11 +11,26 @@ export interface TickSource {
...
+		});
+		let fallbackHandle: number | null = null;
+
 		worker.onmessage = () => onTick();
-		worker.onerror = (ev) => console.error('The logical engine tick worker failed', ev);
+		worker.onerror = (ev) => {
+			console.error('The logical engine tick worker failed, falling back to setInterval', ev);
+			worker?.terminate();
+			worker = null;
+			fallbackHandle = window.setInterval(onTick, pollingIntervalMs);
```
ginderjeremiah, line 27: Nit: `onerror` isn't idempotent. If it ever fired twice, `worker?.terminate()` no-ops the second time (already null) but `window.setInterval` runs again — leaking the first interval, since `stop()` only clears the most recent `fallbackHandle`. In practice `terminate()` makes a repeat fire unlikely, so this is minor, but a cheap guard (`if (worker === null) return;` at the top of the handler) makes the fallback path strictly single-shot and would also cover a stray error arriving after `stop()`. Take it or leave it.

---

ginderjeremiah (commented):
## Review: worker-failure fallback for the logical-engine tick source (#1612)

**Verdict:** Approve — the change solves the issue exactly and is well-tested. (Submitted as a comment because I'm the PR author, so GitHub won't accept a gating APPROVE from me; the intent is Approve.)

The `onerror` handler now tears down the failed worker and transparently starts the `window.setInterval` fallback, and `stop()` cleans up whichever of worker/fallback ended up active. This matches the issue's ask and removes the silent, unrecoverable tick-stall failure mode.

### Nits

- `UI/src/lib/engine/tick-source.ts:20` — `onerror` isn't idempotent; a second error would leak a second interval. Minor (`terminate()` makes a repeat fire unlikely). Details inline — take it or leave it.

### Test coverage

- Solid. The new test drives the mocked worker's `onerror`, asserts the failed worker is torn down, confirms ticks continue via `setInterval`, and confirms `stop()` halts them — mirroring the source path under `tests/lib/engine/`. The pre-existing "logs a diagnosable error" test still holds (one `console.error`).

### What's good

- Tight, well-scoped change that degrades gracefully rather than adding complexity. Nullable `worker` + `fallbackHandle` bookkeeping is clean, and `stop()` correctly covers both paths.

I'm leaving this approved-in-intent without merging since there's an outstanding nit — your call on whether to fold in the idempotency guard first.

---

ginderjeremiah:
## Review: worker-failure fallback for `createTickSource` (Closes #1612)

**Verdict: Approve** — the change correctly implements the graceful degradation asked for in #1612, and the guarded state machine handles the race conditions cleanly. CI is green (frontend unit, all e2e browsers, verify-frontend all pass; backend jobs skipped as expected for a frontend-only change).

### Correctness (verified against the issue's ask)

- **Fallback on worker failure** — `onerror` now tears down the failed worker (`terminate()`), nulls the reference, and starts the existing `window.setInterval` source. This is exactly the "degrade to the existing fallback rather than halt ticks" behavior #1612 requested. ✅
- **Fallback can't be started twice** — the `if (worker === null) return;` guard at the top of `onerror` is the key defensive move. A `Worker` can fire `onerror` more than once; the guard ensures only the first error spins up an interval, so we never leak a second `setInterval`. ✅
- **No "start fallback after stop" leak** — because `stop()` sets `worker = null`, a late `onerror` firing after `stop()` also hits the null-guard and returns without creating an orphaned interval. Nice that the single guard covers both the repeated-error and the stop-then-error orderings. ✅
- **`stop()` cleans up whichever is active** — `worker?.terminate()` is a no-op when the fallback is active, and the `fallbackHandle !== null` branch clears the interval when it isn't. Both the worker-active and fallback-active states are cleaned up correctly. ✅
- No double-tick concern: `terminate()` stops further `onmessage` delivery, so the worker and the interval never both drive `onTick`.

### Test coverage

Good. The new test (`UI/src/tests/lib/engine/tick-source.test.ts`) genuinely exercises the fallback path end to end: it fires the mocked worker's `onerror`, asserts `terminate()` was called, advances fake timers to confirm ticks continue via `setInterval`, then calls `source.stop()` and confirms ticks cease. Test lives at the mirrored path and follows the existing `MockWorker` pattern. Frontend-only infra change, so no backend mirror is required.

### Nits (non-blocking, optional)

- `stop()` doesn't null `fallbackHandle` after clearing it. Harmless today — a second `stop()` would just call `clearInterval` on a stale handle, which is safe — but nulling it would make the terminal state fully idempotent if you want the symmetry with the `worker = null` line.
- Minor duplication: the fallback (`window.setInterval(onTick, pollingIntervalMs)` + `clearInterval`) now exists in both the `onerror` branch and the no-`Worker` branch. Not worth extracting now — a shared helper would arguably be over-engineering for two call sites — but noting it in case the logic grows.

Clean, well-scoped small PR that closes a real latent failure mode. Good to merge.

---

PR #1677, opened 2026-07-06
ProgressionGraphChecker: cover the boss-level branch of ResolveRatingLevel

### LINKED ISSUE #1614: ProgressionGraphChecker: cover the boss-level branch of ResolveRatingLevel + minor cleanups
### Context

Follow-up from PR #1607 (issue #1581), which migrated `CheckEnemyAttributeConsumption` onto `CombatRating.Marginal` and added `ResolveRatingLevel`.

### Gaps

1. **Untested boss-level branch (main item).** `ResolveRatingLevel` returns `zone.BossLevel` when the enemy is a zone's dedicated boss (placed via `BossEnemyId`, not a spawn table). In the current test graph no zone sets `BossEnemyId`, and no boss-placed enemy carries a core attribute distribution, so this branch is never exercised. It is functionally meaningful for real content — a boss placed only via `BossEnemyId` is rated *exclusively* through this path. Add a test with a boss-placed enemy that has a nonzero core distribution and assert the dead-stat check runs at the zone's `BossLevel`.

2. **(Optional) `CombatRating.Marginal` recomputes its baseline `Rate` on every call.** Per enemy the baseline is recomputed up to 6× (once per attribute). Negligible for an offline lint, but a memoized baseline would be cheap to add if this ever runs hot.

3. **(Optional) Precedence comment.** `ResolveRatingLevel` intentionally gives a zone's spawn-table placement precedence over its boss slot; a one-line comment would guard that intent against a later "simplification".

### Ask

Add the boss-branch test (item 1); items 2–3 are optional polish.

ginderjeremiah:
## Summary

Follow-up from PR #1607 (issue #1581), which migrated `CheckEnemyAttributeConsumption` onto `CombatRating.Marginal` and added `ResolveRatingLevel`. `ResolveRatingLevel` returns `zone.BossLevel` when an enemy is placed only via a zone's dedicated `BossEnemyId` (not a spawn table), but no test in the existing graph exercised that branch — no zone set `BossEnemyId` on an enemy with a nonzero core attribute distribution.

- Added `Enemy_PlacedOnlyAsZoneBoss_IsRatedAtZoneBossLevel`: places an enemy exclusively via a zone's `BossEnemyId` (no spawn-table entry) with a dead (unconsumed) Agility distribution, and asserts the `EnemyInertAttribute` warning still fires — proving the boss-level branch actually supplies a representative level rather than the enemy being skipped as unplaced.
- Extended the test file's `Zone` fixture helper with an optional `bossLevel` parameter so the test can pin a distinct `BossLevel`.
- Added a one-line comment on `ResolveRatingLevel` documenting the (already-implemented, now more discoverable) precedence: a zone's spawn-table placement is checked before its boss slot, so an enemy that's both spawn-tabled and the dedicated boss of the same zone rates at the zone's level-range midpoint, not `BossLevel`.

Per the issue, item 2 (memoizing `CombatRating.Marginal`'s baseline `Rate`) was left out — it's explicitly called out as optional/negligible for this offline lint and out of scope for a test-coverage fix.

## Test plan

- [x] `dotnet test Game.sln --no-build --no-restore` — 3000/3000 passing (`Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Application.Tests` incl. the new test, `Game.Api.Tests`)

No frontend changes.

Closes #1614

---
_Generated by [Claude Code](https://claude.ai/code/session_01XLPXmHiqQcKaubLxtAxF96)_

### DIFF
--- a/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
+++ b/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
@@ -392,6 +392,24 @@ namespace Game.Application.Tests.Content
             Assert.DoesNotContain(_checker.Check(graph), f => f.Check == "EnemyInertAttribute");
         }
 
+        [Fact]
+        public void Enemy_PlacedOnlyAsZoneBoss_IsRatedAtZoneBossLevel()
+        {
+            // Enemy 9 carries no spawn-table entry at all — its only live placement is zone 1's dedicated boss
+            // slot — so a representative level can only come from ResolveRatingLevel's boss-level branch
+            // (zone.BossLevel). Its kit (skill 2, same as the sibling AGI-dead-stat tests above) doesn't consume
+            // Agility, so a live rating (i.e. the branch actually firing rather than being skipped as unplaced)
+            // must still surface the dead-stat warning.
+            var graph = HealthyGraph() with
+            {
+                Zones = [Zone(0), Zone(1, unlockChallengeId: 0, bossEnemyId: 9, bossLevel: 5), Zone(2, isHome: true)],
+                Enemies = HealthyGraph().Enemies
+                    .Append(Enemy(9, isBoss: true, skillPool: [2], spawns: [], attributeDistribution: [(EAttribute.Agility, 5, 1)]))
+                    .ToList(),
+            };
+            AssertHasFinding(graph, "EnemyInertAttribute", ContentGraphSeverity.Warning, "Enemy", 9);
+        }
+
         // --- Classes ----------------------------------------------------------------------------------
 
         [Fact]
@@ -1075,6 +1093,7 @@ namespace Game.Application.Tests.Content
             int id,
             bool isHome = false,
             int? bossEnemyId = null,
+            int bossLevel = 1,
             int? unlockChallengeId = null,
             DateTime? retiredAt = null) => new()
             {
@@ -1085,7 +1104,7 @@ namespace Game.Application.Tests.Content
                 LevelMin = 1,
                 LevelMax = 10,
                 BossEnemyId = bossEnemyId,
-                BossLevel = 1,
+                BossLevel = bossLevel,
                 UnlockChallengeId = unlockChallengeId,
                 IsHome = isHome,
                 DesignerNotes = "",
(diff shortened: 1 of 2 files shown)


ginderjeremiah:
## Review: test coverage for the boss-level branch of `ResolveRatingLevel` (closes #1614)

**Verdict:** Approve — the new test genuinely exercises the previously-uncovered boss-level branch, asserts something meaningful, and the scope matches the issue (item 1 done; items 2–3 correctly treated as optional).

### Verification

- **Right branch is hit.** `Enemy(9, ..., spawns: [])` has no spawn-table entry, and its only live placement is Zone 1's dedicated `bossEnemyId: 9`. In `ResolveRatingLevel`, `enemy.Spawns.Any(...)` is false for every live zone, so the only way a non-null level is produced is the `zone.BossEnemyId == enemy.Id → zone.BossLevel` branch. Before this PR no zone in the graph set `BossEnemyId` on an enemy with a core distribution, so this path was never covered — that gap is now closed.
- **Assertion is meaningful.** `AssertHasFinding` requires an `EnemyInertAttribute`/`Warning` finding on `Enemy 9`. The dead (unconsumed) Agility distribution only surfaces that warning if the enemy was rated at all — i.e. if the boss branch supplied a level rather than the enemy being skipped as unplaced. Good inspection-based proxy given the lint exposes no direct level value.
- **Precedence comment** lands correctly inside `ResolveRatingLevel` just ahead of the spawn check, and accurately documents the deliberate spawn-before-boss ordering — exactly what item 3 asked for as a guard against a later "simplification."
- **Fixture change** (`bossLevel` param, default `1`) is backward-compatible for every existing caller.

### Nits (non-blocking)

- The test proves the branch *fires* but doesn't assert the resolved level actually equals `BossLevel`. That value isn't observable through this lint's output and isn't load-bearing here (Agility is dead at any level), so the warning-fires proxy is a reasonable call — just noting the coverage is "branch executed," not "returned value verified."

### CI

All backend and e2e checks green (`test-backend`, `verify-backend`, `test-e2e` chromium/firefox/webkit, `all-checks-passed`). Frontend jobs skipped as expected (no FE changes). The `get_status` "pending" is just the empty legacy commit-status API, not a real check.

### What's good

Tight, well-scoped test-coverage fix with a clear intent-documenting comment. The explanatory comment on the new test is genuinely helpful for a non-obvious "warning-fires-therefore-branch-ran" assertion. Correctly deferred the optional memoization (item 2) as out of scope.

_Reviewed via the review-pr workflow._

---

PR #1617, opened 2026-07-05
Surface combat power ratings in the UI (player + enemy)

ginderjeremiah:
## Summary

Delivers spike #1526 Decision 7 (the last sub-issue of that spike): both the player's and the enemy's combat rating (`CombatRating.Rate`) are now computed server-side and surfaced in the UI, display-only — the client never recomputes them (no parity surface), matching how the reward curve (#1531/#1532) already consumes the same measure.

**Backend**
- `EnemyInstance.EnemyRating` — added to the battle-start payload (`NewEnemy`/`ChallengeBoss`/the prefetched next-battle), computed from the enemy's already-selected fielded `BattleSkills` loadout.
- `BattleService.RatePlayer(Player, CancellationToken)` — a new public helper that rebuilds a fresh `BattleSnapshot` from the player's *current* live state (not a frozen battle snapshot) and rates it, for display purposes.
- The player's rating is threaded onto the **natural existing player-state payloads** rather than a dedicated endpoint, per the issue's own guidance:
  - `PlayerData.PlayerRating` — login / `Status` / `SelectPlayer` / `SwitchPlayer`.
  - `UpdatePlayerStatsResponse.PlayerRating` — recomputed after a stat reallocation (the key synergy with #1528's inert-stat signal: a point spent on an enabler-less amplifier visibly doesn't move this number).
  - `DefeatRewards.PlayerRating` (the wire model) / `DefeatResult.PlayerRating` — threaded through from the reward computation `RecordVictory` already does, so it updates immediately on a level-up with no extra cost.

**Frontend**
- `PowerMeter.svelte` — a new "Power" readout on the Attributes screen, beside the budget/level meters.
- `BattlerCard.svelte` / `BossBattlerCard.svelte` — both battlers' ratings on the Fight screen; the enemy's carries a matched/trivial cue (`rating-cue.ts`) derived from the ratio, reusing existing semantic color tokens, so the anti-grind curve reads as "you've outgrown this zone" rather than a bare number.
- `player-manager.ts` reconciles `playerRating` on load, stat-allocation save, and victory — the same reconcile pattern the existing fields already use.

## Judgment call worth flagging

[...]

PR #1679, opened 2026-07-06
Mark PlayerData.PlayerRating as required for consistency

### LINKED ISSUE #1623: Mark PlayerData.PlayerRating as required for consistency
Follow-up from PR #1617 (review nit).

`PlayerData.PlayerRating` (`Game.Api/Models/Player/PlayerData.cs`) is a `double` that is not marked `required`, while every other property on `PlayerData` is. Because it's a value type it silently defaults to `0` if a future construction path forgets to set it, rather than failing at compile time.

`FromPlayer` always sets it today, so this is a no-op behavior change — purely a consistency/safety improvement to bring it in line with its siblings.

### Acceptance
- `PlayerData.PlayerRating` is marked `required`.
- Build stays green (all construction sites already set it).

ginderjeremiah:
## Summary

Follow-up from PR #1617 (review nit). `PlayerData.PlayerRating` (`Game.Api/Models/Player/PlayerData.cs`) was a `double` that was not marked `required`, while every other property on `PlayerData` is. Because it's a value type it would silently default to `0` if a future construction path forgot to set it, rather than failing at compile time.

- Marked `PlayerRating` `required`.
- Verified `FromPlayer` (the only construction site solution-wide) already sets it, so this is a no-op behavior change — purely a consistency/safety improvement to bring it in line with its siblings.

## Test plan

- [x] `dotnet test Game.sln --no-build --no-restore` — 3000/3000 passing

No frontend changes.

Closes #1623

---
_Generated by [Claude Code](https://claude.ai/code/session_017XbsyXWPto4xFyveDd4mf4)_

### DIFF
--- a/Game.Api/Models/Player/PlayerData.cs
+++ b/Game.Api/Models/Player/PlayerData.cs
@@ -42,7 +42,7 @@ namespace Game.Api.Models.Player
         /// (#1528). Recomputed fresh from current state (not a stored battle snapshot), display-only, never
         /// recomputed client-side (no parity surface).
         /// </summary>
-        public double PlayerRating { get; set; }
+        public required double PlayerRating { get; set; }
 
         public static PlayerData FromPlayer(
             CorePlayer player,


Game.Api/Models/Player/PlayerData.cs
```diff
@@ -42,7 +42,7 @@ public class PlayerData : IModel
         /// (#1528). Recomputed fresh from current state (not a stored battle snapshot), display-only, never
         /// recomputed client-side (no parity surface).
         /// </summary>
-        public double PlayerRating { get; set; }
+        public required double PlayerRating { get; set; }
```
ginderjeremiah, line 45: Nit (non-blocking): the change is correct, but the "in line with its siblings" premise in the issue/PR isn't quite right. `PlayerData`'s other value-type properties — `Id`, `Level`, `Exp`, `CurrentZone`, `StatPointsGained`, `StatPointsUsed` — are *not* marked `required` either, and they carry the exact same "silently defaults to 0" footgun this PR is guarding against. So after this change `PlayerRating` becomes the *only* `required` value type on the model, which is arguably a new inconsistency rather than the uniformity the description implies.

The safety win for a display-only field is still reasonable, so I'm fine landing this as-is. But it'd be worth a follow-up to decide the rule intentionally: either mark all these always-set value types `required` (uniform compile-time safety), or accept that value types stay unmarked and treat this one as a deliberate exception. Since `FromPlayer` is the sole construction site, either direction stays green.

---

ginderjeremiah (commented):
## Review: Mark `PlayerData.PlayerRating` as `required` (issue #1623)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't accept a formal approval from me) — a correct, verified no-op safety improvement; one non-blocking nit noted below.

I verified the change against the codebase:
- `new PlayerData` appears in exactly one place solution-wide — `FromPlayer` (`PlayerData.cs:65`), which sets `PlayerRating`. So the build stays green and this is a genuine no-op behavior change, matching the issue's acceptance criteria.
- `required` is a compile-time initializer constraint and doesn't affect JSON serialization, so the frontend `IPlayerData` contract is unaffected.

### Nits
- `PlayerData.cs:45` — The issue/PR premise that "every other property on `PlayerData` is required" is inaccurate: the value-type properties (`Id`, `Level`, `Exp`, `CurrentZone`, `StatPointsGained`, `StatPointsUsed`) are **not** marked `required`, and they share the same "silently defaults to `0`" footgun. This change makes `PlayerRating` the only `required` value type on the model. Not a blocker — the safety gain is real — but a follow-up to decide the rule intentionally (mark all always-set value types `required`, or accept value types stay unmarked) would resolve the inconsistency cleanly.

### Test coverage
No new tests, which is appropriate — this is a compile-time annotation on a response model with no runtime logic to exercise, and the existing 3000-test suite covering `FromPlayer`'s construction path already guards the (unchanged) behavior.

### What's good
Tight, well-scoped tech-debt change with the sole construction site verified and an accurate no-op test plan. The XML doc on the property remains accurate.

---

Game.Api/Models/Player/PlayerData.cs
ginderjeremiah: Good catch — filed #1680 to decide the rule intentionally rather than expand this PR's scope. Landing as-is per your note.

---

ginderjeremiah:
## Review: Mark `PlayerData.PlayerRating` as `required` (closes #1623)

**Verdict: Approve.** Correct, minimal, and safe consistency change.

### Verification
- **Only construction site confirmed.** The single `new PlayerData` in the whole solution is inside `FromPlayer` itself (`Game.Api/Models/Player/PlayerData.cs:65`), and it already assigns `PlayerRating = playerRating` (line 139). Since `required` is satisfied by object-initializer assignment, the build stays green and there is no behavior change — the no-op claim holds. Every other call path (`LoginController`, the API test bases) routes through `FromPlayer`, so nothing else needs touching.
- **CI is green** — `all-checks-passed` succeeded; `test-backend`, `verify-backend`, and all three `test-e2e` browsers pass; frontend jobs correctly skipped (no frontend changes).
- Coverage is already in place: `PlayerDataMappingTests.FromPlayer_ProjectsThePlayerRating` asserts the value flows through, plus the integration tests (`LoginControllerTests`, `InventorySocketTests`, `UpdatePlayerStatsSocketTests`). No new tests warranted for a modifier-only change.

### Nit (non-blocking, optional follow-up)
- The PR/issue framing says "every other property on `PlayerData` is marked `required`," but that's only true of the reference-type/collection members. The value-type `int` siblings — `Id`, `Level`, `Exp`, `CurrentZone`, `StatPointsGained`, `StatPointsUsed` — are *not* `required` and carry the exact same silent-default-to-0 footgun this PR is closing for `PlayerRating`. Marking `PlayerRating` required is still a clear net improvement, but if the goal is real construction-time safety across the model, those ints deserve the same treatment. Could be a small tech-debt follow-up rather than expanding this PR's scope.

### What's good
- Tightly scoped, well-justified one-line change with a clear commit rationale, and the existing XML doc on the property already explains what the value is. No dead code, no scope creep.

Recommend merge.

---

ginderjeremiah:
Follow-up filed as #1683 for the review nit — the value-type `int` siblings on `PlayerData` (`Id`, `Level`, `Exp`, `CurrentZone`, `StatPointsGained`, `StatPointsUsed`) share the same silent-default-to-0 risk and aren't yet `required`. Out of scope for this PR, so tracked separately. Merging this one.

---

PR #1682, opened 2026-07-06
Store player progress cache as a Redis hash so saves write only dirty rows

### LINKED ISSUE #1635: Progress cache round-trips the full blob per battle; payload grows O(total statistic rows) forever
## Problem

Every battle completion (`BattleStatisticsEventHandler`) loads and saves player progress through `PlayerProgressRepository` (`Game.DataAccess/Repositories/PlayerProgressRepository.cs:104-105, 125-141`), which deserializes the **entire** `Progress_{id}` blob and re-serializes/rewrites the **entire** snapshot — even though a battle touches ~10-20 rows.

`ProgressUpdatedHandler`'s own comment notes a long-lived account holds one statistic row per enemy/skill (plus kills-by-damage-type, per-zone entries, proficiency rows, …) — hundreds of rows — so per-battle Redis traffic and JSON allocation grow linearly with account age, on the hottest per-player path in the game. The docs accept the per-battle *reload*; the O(N) full-blob **write** reintroduces the bloat concern that justified splitting progress out of the player blob in the first place.

## Direction

Store progress as a Redis hash and write only the dirty rows (`HSET` — the `ProgressUpdated` event already carries exactly those), keeping the absolute-value/idempotent contract. Alternatively keep the blob but split it per kind. Needs a small design pass on the key shape + migration of existing cached blobs (or a TTL-based cutover, since Postgres holds the durable copy).

## Tests

Integration tests covering read-your-writes across the new shape and idempotent re-apply; a sanity check that a battle's save only touches the dirty rows.

ginderjeremiah:
## Summary

Every battle completion (`BattleStatisticsEventHandler`) loaded and saved player progress through `PlayerProgressRepository`, which deserialized the **entire** `Progress_{id}` cache blob and re-serialized/rewrote the **entire** snapshot on every save — even though a battle only touches ~10-20 rows. A long-lived account holds one statistic row per enemy/skill (plus kills-by-damage-type, per-zone entries, proficiency rows, …) — hundreds of rows — so per-battle Redis traffic and JSON allocation grew linearly with account age, on the hottest per-player path in the game.

## Direction taken

Per the issue's suggested direction, progress is now cached as a **Redis hash** rather than a serialized blob, with each row (statistic, challenge, proficiency) stored as its own hash field keyed by its natural identity (statistic type + entity id / challenge id / proficiency id):

- `Save`'s cache advance now `HSET`s only the rows the existing write-behind dirty-tracking (`PlayerProgress.DirtyStatistics`/`DirtyChallenges`/`DirtyProficiencies`) already identified as changed — the exact same dirty set the durable DB write-behind event already persists, now applied to the cache write too — instead of re-serializing every row the player has ever earned.
- A load reads the whole hash in one round trip (`HGETALL`), guarded by an atomic `TYPE` check in the same Lua script so a hash with no fields yet is never mistaken for a genuine cache hit.
- **No explicit migration step or key-format bump.** A key still holding the prior string-blob representation is self-healing: the same `TYPE` check treats a non-hash value as a miss and clears it, falling through to the existing DB-reload path — Postgres remains the durable copy, so there's nothing to migrate.
- Two new `ICacheService` primitives back this: `HashGetAllIfExists` (atomic type-check + `HGETALL`, one round trip) and `HashSetAndForget` (atomic multi-field `HSET` + TTL refresh via a Lua script, mirroring the existing `GetSet`/`ReclaimAndForget` pattern so a hash is never left without its TTL refreshed).

Dirty-tracking correctness was verified against `Game.Core.Progress.PlayerProgress`: every mutation path (`Increment`/`SetMax`/`SetMin` via `GetOrCreate`, `SetProficiencyProgress`, challenge progress updates) already marks the row dirty, so the cache's dirty-only write is exactly as complete as the existing DB write-behind event's dirty-only persist.

## Tests

- `RedisServiceIntegrationTests`: round-trip, TTL refresh, partial-overwrite (untouched fields survive), no-op on empty fields, and the WRONGTYPE self-heal for the two new `ICacheService` primitives.
- `PlayerProgressRepositoryIntegrationTests`: added `Save_OnlyWritesTheRowsThatChanged_LeavingEarlierCachedRowsInTheHashUntouched`, directly inspecting the raw Redis hash to pin the O(dirty) write claim — the issue's own "sanity check that a battle's save only touches the dirty rows" ask. All pre-existing tests (cache-first reads, write-behind event contents, publish-before-cache ordering, shared-batch-flush deferral) pass unchanged since they assert repository behavior, not the underlying cache representation.
- The durable DB write-behind path (`ProgressUpdatedEvent` / `DataProviderSynchronizer`) is untouched by this PR — only the Redis cache representation changed — so its existing idempotent-re-apply coverage still applies.

- [x] `dotnet test Game.sln --no-build` — 3006/3006 passing
- No frontend changes.

## Docs

Updated `docs/backend-persistence.md`'s progress-cache section to describe the hash representation and the self-healing type-check.

Closes #1635

---
_Generated by [Claude Code](https://claude.ai/code/session_01E1TvBLExjTuoVB37jokEJh)_


ginderjeremiah (commented):
## Review: Store player progress cache as a Redis hash so saves write only dirty rows (#1635)

**Verdict:** Approve — solves the O(total-rows) per-battle write cleanly by reusing the existing dirty-tracking, with atomic Lua primitives consistent with the surrounding cache layer and thorough integration coverage. (Posted as a COMMENT since I'm the PR author; GitHub won't accept a gating APPROVE on one's own PR.)

The change is well-targeted: the write-behind DB event already carried exactly the dirty rows (`changed`), and the cache advance now `HSET`s that same set instead of re-serializing the full snapshot — so both durable and cache writes are now O(dirty) and share one source of truth. A load reconstructs the whole aggregate from the accumulated hash fields in one `HGETALL`.

### Correctness — verified
- **Field keys are stable and collision-free.** `S_{type}_{entityId|n}` / `C_{id}` / `P_{id}` — distinct prefixes across kinds, and within statistics the fixed two-underscore/three-token format means each `(type, entity)` maps to a unique field, so a later `HSET` on the same row overwrites rather than duplicates. Prefixes double as the read discriminator in `FromHashFields`, so no separate kind tracking is needed.
- **The atomic TYPE-then-HGETALL guard is the crux and is handled right.** Without it, `HGETALL` on a missing key returns an empty array indistinguishable from a genuine empty-hash "hit", which would mask the DB reload. The single-script `type` check returns Lua `false` (→ RESP null → `null`) for a missing key, and the same check self-heals a stale string-blob (`del` + treat as miss), so the old representation needs no explicit migration since Postgres is durable.
- **Deferred cache advance still snapshots correctly.** `fields` is materialized synchronously off `changed`, so the `OnFlushed(AdvanceCache)` closure on the shared-batch path carries this save's state, preserving publish-before-cache.
- **Absolute-value/idempotent contract preserved** — each row's field holds its absolute value and a re-save overwrites the same field.
- **`HashSetAndForget` bundles all field writes + the `PEXPIRE` into one script** (mirroring `GetSet`/`ReclaimAndForget`), so a hash is never left without its TTL refreshed, and the empty-fields no-op correctly avoids an erroring empty `HSET`.

### Test coverage
Strong and directly matched to the issue's asks:
- `RedisServiceIntegrationTests` — round-trip + TTL, partial-overwrite (untouched fields survive), empty-fields no-op, and the WRONGTYPE self-heal.
- `PlayerProgressRepositoryIntegrationTests.Save_OnlyWritesTheRowsThatChanged...` inspects the raw Redis hash to pin the O(dirty) write claim — the issue's "sanity check that a battle's save only touches the dirty rows".
- Pre-existing repository tests (cache-first reads, write-behind contents, publish-before-cache ordering) pass unchanged since they assert repository behavior, not the cache representation. No frontend/battle-logic surface, so no mirrored FE tests needed.

### Notes (non-blocking, no action required)
- The brand-new-player-with-zero-rows path re-reads the DB on each progress read until the first row is earned (empty `ToHashFields` → no-op `HSET`, key never created). Documented in-code and self-limiting to the very first battle — a fine trade-off.
- `FromHashFields` silently skips a field whose JSON deserializes to null. This only happens on genuine corruption and matches the tolerant `Deserialize` behavior used elsewhere in the cache layer, so it's reasonable as-is.
- Each row's serialized value redundantly re-encodes its identity (already in the field key), but that keeps deserialization key-parsing-free — a sensible simplicity trade.

Docs updated (`docs/backend-persistence.md`) per the design-decision requirement. Nice work — I'll merge once CI goes green.

---

PR #1687, opened 2026-07-07
Skills header: include fielded innate skills in combined dps/burst totals

### LINKED ISSUE #1657: Decide whether fielded innate (item-granted) skills belong in the Skills header's combined dps/burst totals
### Context

#1637 fixed `combinedEffectiveDps`/`combinedEffectiveBurst` (`UI/src/routes/game/screens/skills/skills-view.svelte.ts`) counting a dormant (weapon-mismatched) equipped skill that can't actually fire. That fix kept the header a **selected-loadout-only** readout — it deliberately did not change what's counted, only excluded the unfireable dormant case.

### Open question

Fielded **innate** (item-granted) skills — surfaced in `InnateBand.svelte` via `SkillsView.innateSkills` — *do* fire in battle (unless themselves dormant or a duplicate of an already-selected skill) but are excluded from `combinedEffectiveDps`/`combinedEffectiveBurst` today. Decide whether the header should:

- Stay a selected-loadout-only readout (current behavior) — arguably fine since the header sits directly above the selected-loadout rail, not the innate band, and could just be labeled more precisely (e.g. "loadout dps" or similar) if it isn't already unambiguous.
- Or reflect the true fielded set (selected-minus-dormant **plus** non-duplicate, non-dormant granted skills), matching what the battle actually produces.

If the latter, reuse `innateSkills`' existing `duplicate`/dormant-eligibility logic rather than re-deriving it, and update/extend the unit tests covering `combinedEffectiveDps`/`combinedEffectiveBurst`.

_Surfaced during #1656's review of the #1637 fix._

ginderjeremiah:
## Summary

#1657 asked whether the Skills header's `combinedEffectiveDps`/`combinedEffectiveBurst` totals should stay a selected-loadout-only readout, or reflect the true fielded set (loadout + fielded item-granted "innate" skills), matching what the battle actually produces.

**Decision made in this PR:** reflect the true fielded set. A player looking at "combined dps/burst" reasonably expects it to represent everything currently contributing in battle — and innate skills already fire under the same rules as the loadout (unless dormant or a duplicate). Leaving them out understated the real number rather than making the header more honest, so option 2 (fold them in) is the better default. If this reading is wrong, happy to revisit — flagging it here since #1657 framed it as an open design call rather than a defect.

## Changes

- `UI/src/routes/game/screens/skills/skills-view.svelte.ts`: renamed the private `fieldedEquipped` derivation to `fieldedSkillIds` and extended it to also include `innateSkills` entries that are fielded — i.e. not a duplicate of an already-counted skill and not themselves dormant under the equipped weapon. Reuses `innateSkills`' own `duplicate`/`dormant` logic (the same rule `InnateBand.svelte` already applies), so the two views of "what's fielded" can't drift apart.
- `combinedEffectiveDps`/`combinedEffectiveBurst` now sum over `fieldedSkillIds` instead of the loadout-only list.
- No header relabeling needed — the existing "eff. dps"/"eff. burst" labels in `Skills.svelte` were already generic enough to cover the corrected totals.

## Tests

Added to `UI/src/tests/routes/game/screens/skills/skills-view.test.ts`:
- A granted (innate) skill's contribution is added once, alongside the selected loadout.
- A grant that duplicates an already-selected skill is not double-counted.
- A granted skill that is itself dormant under the equipped weapon is excluded.

## Test plan

- [x] `npm run test:unit:ci` (full suite) — all tests passing

Closes #1657

---
_Generated by [Claude Code](https://claude.ai/code/session_01JwmY28bxJJ3iMBQzhGH46h)_

### DIFF
--- a/UI/src/routes/game/screens/skills/skills-view.svelte.ts
+++ b/UI/src/routes/game/screens/skills/skills-view.svelte.ts
@@ -353,22 +353,24 @@ export class SkillsView {
 	/** Metrics for the inspected skill, falling back to the first listed skill. */
 	readonly selected = $derived(this.metricsById[this.selectedId] ?? this.railList[0]);
 
-	/** Equipped skills the battle can actually fire — excludes a weapon-mismatched skill dormant under
-	 *  the equipped weapon (#1342), so the combined totals below never sum an output the battle can't
-	 *  produce. Innate (item-granted) skills also fire but are deliberately excluded — the totals stay
-	 *  a selected-loadout-only readout; see #1657 for the open question of folding them in. */
-	private readonly fieldedEquipped = $derived(
-		this.equipped.filter((id) => {
+	/** Every skill id the battle will actually fire right now: equipped skills that aren't dormant under
+	 *  the current weapon (#1342), plus item-granted innate skills that are fielded — not a duplicate
+	 *  already counted via the loadout or an earlier grant, and not themselves dormant. Reuses
+	 *  {@link innateSkills}'s own duplicate/dormant rule so this can't drift from what the innate band
+	 *  already flags (#1657). */
+	private readonly fieldedSkillIds = $derived([
+		...this.equipped.filter((id) => {
 			const skill = this.metricsById[id]?.skill;
 			return skill != null && !this.dormant(skill);
-		})
-	);
+		}),
+		...this.innateSkills.filter((i) => !i.duplicate && !this.dormant(i.skill)).map((i) => i.skill.id)
+	]);
 
-	/** Combined effective DPS of the equipped loadout vs the current Toughness. */
-	readonly combinedEffectiveDps = $derived(this.fieldedEquipped.reduce((sum, id) => sum + this.effectiveDps(id), 0));
+	/** Combined effective DPS of every fielded skill (loadout + innate) vs the current Toughness. */
+	readonly combinedEffectiveDps = $derived(this.fieldedSkillIds.reduce((sum, id) => sum + this.effectiveDps(id), 0));
 
-	/** Combined effective single-hit burst of the equipped loadout vs the current Toughness. */
-	readonly combinedEffectiveBurst = $derived(this.fieldedEquipped.reduce((sum, id) => sum + this.effective(id), 0));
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Skills header combined dps/burst totals now reflect fielded innate skills (#1657)

**Verdict:** Approve — clean, correctly-scoped resolution of the open design question in #1657, with the right decision and matching tests. (Posting as `COMMENT` only because I'm the PR author; GitHub won't accept a gating `APPROVE` from the author — the intended verdict is approve.)

The decision (fold fielded innate skills into the totals rather than keeping them loadout-only) is the right call: the header reads as "combined dps/burst" of what's fighting, and innate grants already fire under the same rules, so excluding them silently understated the real number.

### Correctness

- `fieldedSkillIds` composition is sound. The equipped branch keeps the existing non-dormant filter (#1342); the innate branch reuses `innateSkills`' own `duplicate` flag and the shared `dormant()` method. Because `innateSkills` seeds its `seen` set from `equipped` and then each prior grant, `!i.duplicate` correctly excludes both loadout-overlap and grant-vs-grant overlap — no double counting.
- Metrics resolve for granted ids: `effective`/`effectiveDps` read `metricsById`, and `metricsSkillIds` already folds in `inventoryManager.grantedSkillIds`, so innate contributions aren't silently zero.
- The "can't drift" claim holds concretely: `InnateBand.svelte` renders a `SkillCardStats` block only for `!duplicate && !dormant` grants — exactly the innate set summed here — so a card showing a dps/dmg block is precisely one that contributes to the total.

### Test coverage

Adequate for this change. The three added cases cover the meaningful branches: contribution added once alongside the loadout, duplicate-of-selected not double-counted, and dormant grant excluded. This is UI-side display logic (not battle simulation), so no backend/frontend mirrored-suite requirement applies.

### What's good

- Reuses the existing `duplicate`/`dormant` logic rather than re-deriving it, exactly as #1657 requested — keeps the header and the innate band structurally consistent.
- The rename `fieldedEquipped` → `fieldedSkillIds` accurately reflects the widened meaning; the updated doc comment is clear about what's counted and why.
- Correctly declined the optional relabeling — the existing generic "eff. dps"/"eff. burst" labels already fit.


---

PR #1688, opened 2026-07-07
Reuse equippedRail's equipped→metrics association in fieldedSkillIds (#1662)

### LINKED ISSUE #1662: Skills header combined totals re-derive the equipped→metrics map already built by equippedRail
Follow-up (minor DRY cleanup) from review of #1656.

## Observation

In `UI/src/routes/game/screens/skills/skills-view.svelte.ts`, the new `fieldedEquipped` / `combinedEffectiveDps` / `combinedEffectiveBurst` derivations re-walk the equipped loadout and re-associate each skill with its entry in `metricsById`. The `equippedRail` derived value already builds essentially the same `equipped → metrics` association for the per-card rendering.

Re-deriving it a second time is a small redundancy — no user-facing effect, just a maintainability nit (two places that must stay in sync on how equipped skills map to their metrics).

## Suggested cleanup

Factor the equipped→metrics pairing into a single derived value that both `equippedRail` and the combined totals consume.

Non-blocking; surfaced during PR review of #1656.

ginderjeremiah:
## Summary

Minor DRY cleanup flagged during review of #1656/#1657.

`fieldedSkillIds` (`UI/src/routes/game/screens/skills/skills-view.svelte.ts`) re-walked the equipped loadout and re-indexed into `metricsById` just to check dormancy — an equipped→metrics association `equippedRail` already builds for the per-card rendering. `fieldedSkillIds` now filters `equippedRail` directly instead of re-deriving the pairing from `this.equipped` + `metricsById`.

Note: the code has moved on somewhat since #1662 was filed (the #1657 fix reshaped `fieldedSkillIds` from the described `fieldedEquipped` shape into a plain id list), so the redundancy is smaller than originally described, but the core ask — factor the equipped→metrics pairing into a single derived value both `equippedRail` and the combined totals consume — still applied and is now done.

No behavior change: `equippedRail` already filters out ids with no `metricsById` entry the same way the old inline check did, so the resulting `fieldedSkillIds` set is identical.

## Test plan

- [x] `npm run test:unit:ci` (full suite) — 3509/3509 tests passing, including the existing `skills-view.test.ts` coverage of `fieldedSkillIds`/combined-totals behavior.

Closes #1662

---
_Generated by [Claude Code](https://claude.ai/code/session_01EXLbepnBAK1GsUEfkXwxJg)_

### DIFF
--- a/UI/src/routes/game/screens/skills/skills-view.svelte.ts
+++ b/UI/src/routes/game/screens/skills/skills-view.svelte.ts
@@ -357,12 +357,10 @@ export class SkillsView {
 	 *  the current weapon (#1342), plus item-granted innate skills that are fielded — not a duplicate
 	 *  already counted via the loadout or an earlier grant, and not themselves dormant. Reuses
 	 *  {@link innateSkills}'s own duplicate/dormant rule so this can't drift from what the innate band
-	 *  already flags (#1657). */
+	 *  already flags (#1657), and {@link equippedRail}'s equipped→metrics association rather than
+	 *  re-walking `equipped` against `metricsById` a second time (#1662). */
 	private readonly fieldedSkillIds = $derived([
-		...this.equipped.filter((id) => {
-			const skill = this.metricsById[id]?.skill;
-			return skill != null && !this.dormant(skill);
-		}),
+		...this.equippedRail.filter((m) => !this.dormant(m.skill)).map((m) => m.skill.id),
 		...this.innateSkills.filter((i) => !i.duplicate && !this.dormant(i.skill)).map((i) => i.skill.id)
 	]);
 


ginderjeremiah (commented):
## Review: Reuse `equippedRail`'s equipped→metrics association in `fieldedSkillIds` (#1662)

**Verdict:** Approve — a correct, behavior-preserving DRY cleanup with no findings. (Posted as a comment since I'm the PR author; the intended verdict is **Approve**.)

I verified the "no behavior change" claim by hand and it holds on all three axes:

- **Filter equivalence.** Old: `equipped.filter(id => metricsById[id]?.skill != null && !dormant(skill))`. New: `equippedRail.filter(m => !dormant(m.skill))`, where `equippedRail = equipped.map(id => metricsById[id]).filter(non-null)`. A `metricsById` entry is only ever constructed *with* a skill (`byId[skill.id] = { skill, … }`), so `metricsById[id] != null` ⟺ `metricsById[id]?.skill != null` — the two guards select exactly the same equipped ids, then apply the identical `!dormant(...)` check.
- **Returned value.** Old yields the equipped `id`; new yields `m.skill.id`. These are equal because `metricsById` is keyed by `skill.id` over 0-based reference data (`skills[id].id === id`), so `m.skill.id === id`. Equipped ids are always unlocked ⟹ in the catalogue ⟹ present in `metricsById`, and a would-be missing entry is dropped identically by both forms.
- **Order & duplicates.** `equippedRail` walks `this.equipped` in loadout order without de-duplicating, so ordering and any (theoretical) duplicate ids are preserved exactly as before.

The doc comment is updated to cite both the #1657 innate-band rule and the new #1662 `equippedRail` reuse, which reads clearly and keeps the "single source of the equipped→metrics pairing" intent visible.

### Test coverage
Adequate — no new tests needed for a pure refactor with identical output, and the existing `skills-view.test.ts` already guards the exact paths this touches: dormant-equipped exclusion from the combined totals (`#1637` block, lines ~724–747), combined DPS/burst over the equipped loadout (lines ~600–603), and fielded-innate inclusion/dormant-grant exclusion (`#1657` block, lines ~762–799). The full suite is green (3509/3509) per the PR's test plan.

### What's good
- Correctly recognizes that `equippedRail` already *is* the equipped→metrics association #1662 asked to centralize, so the fix is a one-line reuse rather than a new derived value — minimal surface, maximal DRY.
- Honest scoping note in the PR body: the redundancy shrank after #1657 reshaped `fieldedSkillIds`, and the description reflects that rather than overselling.

Nothing to change. Once `test-frontend` and the webkit e2e job finish green I'll merge.

---

PR #1689, opened 2026-07-07
Thread a boss flag through the battle hand-back payload (#1647)

### LINKED ISSUE #1647: A resumed boss battle (#1595/#1596 hand-back) always routes through the idle loop, not the boss loop
### Context

#1595/#1596 hand back a battle already in progress via a generic `BattleStartResult`/`EnemyInstance` — no field on it says whether the underlying `PlayerState.IsBossBattle` is set. #1597 (client replay-to-offset) wires this payload into `EnemyManager` so a resumed battle is presented directly via `notifyNewEnemyLoaded` on session start (`EnemyManager.start(activeBattle)`), or (for the sub-5-minute quick-reconnect case) via the ordinary `NewEnemy` response.

### Gap

`EnemyManager.mode` always starts `'idle'` for a fresh session (or stays whatever it already was for the live in-session `NewEnemy` abandon-and-resume case, which is also always `'idle'` — a boss fight is only ever entered via the separate `ChallengeBoss` command, never via `NewEnemy`). If the battle handed back was actually a boss fight (`PlayerState.IsBossBattle == true`), the frontend has no way to know that, so it stays routed through `watchIdleStage` instead of `watchBossStage`.

Server-side crediting is unaffected — `DefeatEnemy`/`AbandonBattle` key off the authoritative `PlayerState.IsBossBattle`, not which command the client happens to call, so exp/zone-clear/challenge crediting for the resumed boss fight is correct either way. The gap is purely client-side presentation and loop control:

- No Zone-Cleared overlay / "next zone unlocked" flow on a resumed boss victory (`resolveBossVictory`'s pacing never runs).
- A resumed boss loss never sends `BattleLost` (the idle path sends nothing explicit on a loss/draw), so any boss-specific bookkeeping that command does is skipped for this one fight.
- An auto-farming boss player (`autoChallengeBoss` persisted) whose fight is resumed this way won't automatically re-challenge after it concludes, since the loop treats it as an idle-mode conclusion.

### Ask

Decide whether this is worth fixing, and if so, thread a boss flag through the hand-back payload (`BattleStartResult` → `EnemyInstance`/`IEnemyInstance`, mirroring how #1616 proposes extending a currently-thin response) so `EnemyManager` can set `mode = 'boss'` before presenting a resumed boss battle. Likely small in isolation, but touches the same payload contract #1595–#1597 established, so it's worth reviewing on its own rather than silently folding in.

Surfaced while implementing #1597 (client replay-to-offset) — not fixed there since it's a pre-existing gap in #1595/#1596's generic hand-back design, not something #1597 introduced.

ginderjeremiah:
## Summary

`BattleStartResult`/`EnemyInstance` (the payload shared by `NewEnemy`, `ChallengeBoss`, and the `GetOfflineProgress` welcome-back summary) carried no field indicating whether a handed-back battle was a dedicated-boss fight. `PlayerState.IsBossBattle` is the authoritative server-side truth, but the client had no way to see it, so `EnemyManager` always defaulted a resumed/reconnected battle to `mode = 'idle'` — even when it was genuinely a boss fight still in progress. That skipped:
- The Zone-Cleared overlay / "next zone unlocked" flow on a resumed boss victory
- Sending `BattleLost` on a resumed boss loss
- Auto-rechallenge for an `autoChallengeBoss` farmer whose fight resumed this way

Server-side crediting was already correct (it keys off `PlayerState.IsBossBattle`, not which command the client called) — this was purely a client-side presentation/loop-control gap.

## Changes

- **Backend**: added `IsBossBattle` to `BattleStartResult` (`Game.Application/Services/BattleService.cs`), set at all four construction sites (fresh idle spawn, fresh boss spawn, the offline-progress hand-back `HandBackPendingBattle`, and the still-in-progress `AbandonBattle` hand-back — the exact path this issue is about). Threaded it into `EnemyInstance`'s `FromSource` projection, the single source shared by every consumer.
- **Codegen**: regenerated `UI/src/lib/api/types/interfaces/enemies.ts` (`dotnet run --project Game.Api -- codegen`) to add `isBossBattle: boolean` to `IEnemyInstance`.
- **Frontend**: `EnemyManager` now sets `this.mode` from `isBossBattle` wherever a server-handed-back battle is presented — both the session-start `start(activeBattle)` path (#1595/#1596 welcome-back hand-back) and the ordinary `NewEnemy` fetch-loop response (the sub-5-minute quick-reconnect case named in the issue, where the still-in-progress battle comes back through a normal fetch rather than the welcome-back summary).

## Test plan

- [x] Backend: new/extended unit + integration tests — `EnemyInstanceMappingTests.FromSource_ProjectsIsBossBattle`, a new `StartBattle_AbandoningAStillInProgressBossBattle_HandsItBackWithBossFlagSet` integration test, and assertions added to the existing idle/boss hand-back and prefetch tests.
- [x] Frontend: new/extended vitest tests in `enemy-manager.test.ts` covering both the session-start resume and the ordinary-`NewEnemy` reconnect path adopting boss mode.
- [x] `dotnet build` + full backend test suite (3007 tests) pass.
- [x] `npm run lint` (ESLint + svelte-check) and full frontend test suite (3522 tests) pass.
- [x] `dotnet format --verify-no-changes` passes.

Closes #1647

---
_Generated by [Claude Code](https://claude.ai/code/session_01BL3ovSP3Vd4tfPzNbw6Zk6)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -112,12 +112,17 @@ export class EnemyManager {
 	 * remainder) — presented directly rather than through the normal fetch, since the idle loop's first
 	 * `NewEnemy` would otherwise report 0ms fought and abandon it with no outcome (#1597). Absent, the loop
 	 * starts exactly as before: the initial stage (`Idle`) drives a fresh fetch.
+	 *
+	 * A handed-back battle's `isBossBattle` (the authoritative `PlayerState.IsBossBattle`) routes the mode
+	 * accordingly (#1647), so a resumed boss fight is presented through the boss loop instead of always
+	 * defaulting to idle.
 	 */
 	public start(activeBattle?: IEnemyInstance) {
 		if (!this.started) {
 			this.started = true;
 			this.battleStageUnhook = onBattleStageChanged((stage) => this.watchBattleStage(stage));
 			if (activeBattle) {
+				this.mode = activeBattle.isBossBattle ? 'boss' : 'idle';
 				this.currentEnemy = activeBattle;
 				notifyNewEnemyLoaded(activeBattle);
 			} else {
@@ -266,6 +271,11 @@ export class EnemyManager {
 					if (result.data.zoneId != null && result.data.zoneId !== playerManager.currentZone) {
 						playerManager.currentZone = result.data.zoneId;
 					}
+					// A still-in-progress battle handed back on this NewEnemy (e.g. a sub-5-minute reconnect
+					// the welcome-back gate didn't resume directly) may be a boss fight the authoritative
+					// PlayerState never actually abandoned — trust its isBossBattle flag over this fetch
+					// loop's own idle-only assumption so the client routes into the boss loop (#1647).
+					this.mode = result.data.enemyInstance.isBossBattle ? 'boss' : 'idle';
 					this.currentEnemy = result.data.enemyInstance;
 					notifyNewEnemyLoaded(this.currentEnemy);
 					return;
--- a/Game.Api.Tests/Unit/EnemyInstanceMappingTests.cs
+++ b/Game.Api.Tests/Unit/EnemyInstanceMappingTests.cs
@@ -27,6 +27,20 @@ namespace Game.Api.Tests.Unit
             Assert.Equal([20, 10], model.SelectedSkills);
         }
 
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public void FromSource_ProjectsIsBossBattle(bool isBossBattle)
+        {
+            var enemy = MakeEnemy();
+            enemy.SetBattleSkills([]);
+            var result = new BattleStartResult { Enemy = enemy, Seed = 1u, IsBossBattle = isBossBattle };
+
+            var model = EnemyInstanceModel.FromSource(result);
+
+            Assert.Equal(isBossBattle, model.IsBossBattle);
+        }
+
         [Fact]
         public void FromSource_ProjectsEnemyRating_FromTheFieldedBattleSkills()
         {
--- a/Game.Api/Models/Enemies/EnemyInstance.cs
+++ b/Game.Api/Models/Enemies/EnemyInstance.cs
@@ -27,6 +27,13 @@ namespace Game.Api.Models.Enemies
         public double EnemyRating { get; set; }
 
         /// <summary>
+        /// Whether this is (or, for a hand-back, was) a dedicated-boss fight rather than an idle-zone spawn
+        /// (#1647) — lets the client route a resumed hand-back into the boss loop instead of always
+        /// defaulting to idle.
+        /// </summary>
+        public bool IsBossBattle { get; set; }
+
+        /// <summary>
         /// Projects a battle-start result onto the wire model. The single source of truth for the
         /// enemy-instance projection shared by the <c>NewEnemy</c> and <c>ChallengeBoss</c> socket commands,
         /// keeping the two from drifting (#492).
@@ -44,6 +51,7 @@ namespace Game.Api.Models.Enemies
                     .Select(modifier => BattlerAttribute.From(modifier.Attribute, modifier.Amount)),
                 ElapsedOffsetMs = source.ElapsedOffsetMs,
                 EnemyRating = CombatRating.Rate(enemy.ToBattler(), isPlayer: false),
+                IsBossBattle = source.IsBossBattle,
             };
         }
     }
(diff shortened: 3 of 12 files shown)


UI/src/lib/engine/battle/enemy-manager.ts
```diff
@@ -266,6 +271,11 @@ export class EnemyManager {
 					if (result.data.zoneId != null && result.data.zoneId !== playerManager.currentZone) {
 						playerManager.currentZone = result.data.zoneId;
 					}
+					// A still-in-progress battle handed back on this NewEnemy (e.g. a sub-5-minute reconnect
+					// the welcome-back gate didn't resume directly) may be a boss fight the authoritative
+					// PlayerState never actually abandoned — trust its isBossBattle flag over this fetch
+					// loop's own idle-only assumption so the client routes into the boss loop (#1647).
+					this.mode = result.data.enemyInstance.isBossBattle ? 'boss' : 'idle';
```
ginderjeremiah, line 278: Non-blocking observation (correct as written): this unconditional mode-set also fires on the **retreat** path. `retreatFromBoss()` → `returnToIdle()` → `getNewEnemy()` sends `NewEnemy`, and if the boss fight is genuinely still in progress the backend hands *that same boss battle* back with `IsBossBattle=true` (the #1595 AbandonBattle hand-back), so `mode` flips right back to `'boss'`.

I worked through whether that's a regression and concluded it isn't — it's arguably an improvement:
- Pre-PR the same hand-back presented the boss enemy in `idle` mode (the exact #1647 bug class), so retreat-mid-fight never cleanly abandoned the boss either — it just fought it through the idle loop.
- Post-PR the client honestly reflects the authoritative `PlayerState.IsBossBattle` (still a boss battle server-side) and the in-progress boss gets the correct victory ceremony. `returnToIdle()` has already cleared `autoFight`, so farming still stops.

The underlying limitation — that retreat can't *instantly* abandon a still-in-progress boss because it only sends `NewEnemy` (unlike `ChallengeBoss`, which deliberately discards the handoff) — is pre-existing and orthogonal to this PR. Worth a follow-up if we want retreat to force-abandon symmetrically, but nothing to change here. One thing the test suite doesn't cover: the retreat test mocks `NewEnemy` to return a non-boss enemy, so it never exercises this still-in-progress-boss hand-back on retreat.

---

ginderjeremiah (commented):
## Review: Thread a boss flag through the battle hand-back payload (#1647)

**Verdict:** Approve *(submitted as a comment since I'm the PR author — GitHub won't accept a gating self-review)*. The change is correct, minimal, and well-tested; the one CI failure is an unrelated flake (see below).

The PR does exactly what #1647 asks: it threads the authoritative `PlayerState.IsBossBattle` through `BattleStartResult` → `EnemyInstance`/`IEnemyInstance` so `EnemyManager` can route a resumed hand-back into the boss loop instead of always defaulting to idle. `isBossBattle` is presentation/loop-control metadata computed off the tick loop, not part of the deterministic simulation, so it correctly does **not** touch the frontend↔backend battle-parity surface.

### Correctness
- All four `BattleStartResult` construction sites set the flag explicitly (fresh idle → `false`, fresh boss → `true`, offline hand-back → `isBossBattle`, still-in-progress abandon hand-back → `state.IsBossBattle`). Setting it explicitly everywhere rather than leaning on the `bool` default is the right call. `EnemyInstance.FromSource` — the single shared projection — threads it through, so every consumer (`NewEnemy`/`ChallengeBoss`/welcome-back) gets it uniformly.
- Frontend mode routing is consistent across both presentation paths (`start(activeBattle)` and the `NewEnemy` fetch-loop response). I verified the unconditional mode-set can't regress an active boss-farm loop: `getNewEnemy` is only ever entered from idle contexts (every caller runs `returnToIdle()` first, or is the initial idle fetch), and boss re-challenge goes through `challengeBoss()`, which never hits this line.

### Nits / observations
- One inline note on `enemy-manager.ts:278` about the retreat interaction — non-blocking, arguably an improvement over the prior behavior, with a pre-existing follow-up worth considering separately.

### Test coverage
Solid on both sides. Backend adds a unit theory (`FromSource_ProjectsIsBossBattle`), a focused integration test (`StartBattle_AbandoningAStillInProgressBossBattle_HandsItBackWithBossFlagSet`, mirroring its idle sibling with infinite-cooldown skills so the re-sim resolves as neither win nor death), plus flag assertions on the existing idle/boss hand-back and prefetch tests. Frontend adds explicit tests for both new branches (start-path boss resume, `NewEnemy` boss hand-back) and pins the idle defaults. I ran `BattleServiceIntegrationTests` locally — all 55 pass, including the new one.

### CI
`test-backend` is red on **1 failing test out of 894 in `Game.Application.Tests`** (the integration suite), and the failure detail isn't in stdout. This PR's production changes are purely additive and can't affect an unrelated integration test; every test it touched passes locally. This is a flaky integration test, not a real failure — re-running the backend job to confirm before merge.


---

PR #1692, opened 2026-07-07
Set reload marker when an abandoned command's late flush fails

### LINKED ISSUE #1663: Write-behind: reload marker not set when a timed-out command's abandoned flush fails
Follow-up from review of #1651.

## Observation

#1651 sets the `PlayerNeedsReload` marker in the synchronous fault path of `RunCommandUnderLock`. A command that is abandoned on the 25s timeout, whose flush *later* fails with a non-cancellation error, faults inside `ReleaseCommandLockWhenSettled`'s continuation — which only logs and never marks the session for reload. In theory this reopens the #1632 divergence window on that path.

## Why it's low priority

Near-impossible in practice: a timeout cancels the command token, so the abandoned flush fails with `OperationCanceledException`, which #1651 deliberately leaves unwrapped (not `PlayerPersistenceFlushFailedException`). It would take a non-cancellation flush failure racing a just-timed-out command to hit this.

## Suggested action

Either set the reload marker (or otherwise converge the in-memory `Player`) in the timeout/continuation fault path as well, or document explicitly why the cancellation semantics make it unreachable.

Non-blocking; surfaced during PR review of #1651.

ginderjeremiah:
## Summary

#1651 set the `PlayerNeedsReload` marker in `RunCommandUnderLock`'s synchronous fault path when a command's `SavePlayer` flush fails with `PlayerPersistenceFlushFailedException`. It missed a second fault path: a command abandoned on the 25s per-command timeout, whose flush *later* fails for a non-cancellation reason, faults inside `ReleaseCommandLockWhenSettled`'s continuation instead — which only logged the fault and never marked the session for reload. In theory this reopened the #1632 divergence window (Redis/Postgres diverging silently) on that specific path.

This was flagged as low-priority/near-impossible in practice (a timeout cancels the command's token, so the abandoned flush normally fails with `OperationCanceledException`, which is deliberately left unwrapped — it would take a non-cancellation flush failure racing a just-timed-out command to hit this), but the fix is small and closes the gap for good rather than leaving it undocumented.

## Changes

- **`Game.Api/Sockets/SocketHandler.cs`**: `ReleaseCommandLockWhenSettled`'s continuation now also marks the session for reload (`_context.Session.MarkPlayerNeedsReload()`) when the abandoned command's fault is/contains a `PlayerPersistenceFlushFailedException`, mirroring the existing synchronous fault-path behavior in `RunCommandUnderLock`. The existing "log a genuine (non-cancellation) fault" behavior is unchanged.

## Test plan

- [x] New test: `ExecuteCommand_AbandonedCommandsFlushFailsAfterTheTimeoutResponse_ReloadsPlayerBeforeTheNextCommand` (`Game.Api.Tests/Unit/SocketCommandExecutionTests.cs`) — a command wedged past the timeout (ignoring the cancellation token) whose flush fails after the timeout response was already sent still causes the next command to reload the player from the repository, mirroring the existing synchronous-path test's shape.
- [x] `dotnet test --no-build --no-restore` — full backend suite (3005 tests) passes.

Closes #1663

---
_Generated by [Claude Code](https://claude.ai/code/session_0177hYHxk8rkX3f1U7ZonUAD)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
@@ -271,6 +271,49 @@ namespace Game.Api.Tests.Unit
             Assert.Same(reloadedPlayer, session.Player);
         }
 
+        [Fact]
+        public async Task ExecuteCommand_AbandonedCommandsFlushFailsAfterTheTimeoutResponse_ReloadsPlayerBeforeTheNextCommand()
+        {
+            // A command abandoned on the per-command timeout budget whose flush later fails for a
+            // non-cancellation reason faults inside ReleaseCommandLockWhenSettled's continuation, not
+            // RunCommandUnderLock's synchronous fault path — it must still mark the session for reload so the
+            // #1632 divergence window doesn't reopen on this path too (#1663).
+            var reloadedPlayer = new PlayerBuilder().WithId(1).Build();
+            var playerRepo = new FakePlayerRepository(reloadedPlayer);
+            using var provider = new ServiceCollection()
+                .AddScoped<IUnitOfWork, NoOpUnitOfWork>()
+                .AddScoped<IPlayerRepository>(_ => playerRepo)
+                .BuildServiceProvider();
+
+            var socket = new FakeWebSocket(sendDuration: TimeSpan.Zero);
+            var session = new SessionService(new NoOpSessionStore());
+            session.CreateSession(userId: 1, playerId: 1);
+            var context = new SocketContext(socket, playerId: 1, session, isAdmin: false, _loggerFactory.CreateLogger<SocketContext>());
+            var release = new TaskCompletionSource();
+            var handler = new SocketHandler(
+                context,
+                new WedgedThenStubCommandFactory(release.Task, new PlayerPersistenceFlushFailedException(new InvalidOperationException("late blip"))),
+                provider.GetRequiredService<IServiceScopeFactory>(),
+                _loggerFactory.CreateLogger<SocketHandler>(),
+                () => { },
+                commandTimeout: TimeSpan.FromMilliseconds(10));
+
+            await handler.ExecuteCommand(new SocketCommandInfo("Wedged") { Id = "c1" });
+            Assert.Contains(socket.SentMessages, m => m.Contains("Command timed out.") && m.Contains("c1"));
+            Assert.False(session.PlayerNeedsReload);
+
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: close the reload-marker gap on the abandoned-command continuation fault path (#1663)

**Verdict:** Approve — the fix correctly mirrors the synchronous fault path into `ReleaseCommandLockWhenSettled`'s continuation, closing the last path that could reopen the #1632 Redis/Postgres divergence window; it's well-tested and well-documented. (Filed as a `COMMENT` because GitHub won't accept a gating verdict from the PR author — this is an approve.)

### Correctness

- The mirror is faithful and lands in the right place. The synchronous path (`RunCommandUnderLock`, `SocketHandler.cs:248`) catches a single `PlayerPersistenceFlushFailedException`; the continuation instead inspects `task.Exception.InnerExceptions` (`SocketHandler.cs:319`), which is the correct shape for a faulted `Task`'s `AggregateException` and matches the pre-existing logging check right below it.
- The two fault paths are mutually exclusive — a command either settles within the timeout budget (synchronous catch) or is abandoned on the timeout and settles later in the continuation — so there's no risk of double-marking.
- Ordering is correct: `MarkPlayerNeedsReload()` is set inside the continuation *before* `_commandLock.Release()`, so the next command's `WaitAsync` acquire observes the marker (the release/acquire is the memory barrier). The test documents and exercises exactly this handoff.
- A genuine flush failure now both marks reload and logs (both `InnerExceptions.Any` predicates are true), which is the intended behavior.

### Test coverage

- `ExecuteCommand_AbandonedCommandsFlushFailsAfterTheTimeoutResponse_ReloadsPlayerBeforeTheNextCommand` is a solid integration-style test that drives a real `SocketHandler` with a command wedged past the timeout (ignoring the cancellation token) whose flush fails afterward, and asserts the full contract: timeout response sent, marker not set prematurely, then reload-on-next-command, marker cleared, player replaced, and the fault logged. It mirrors the existing synchronous-path test's shape, consistent with the project's testing guidelines. This is backend-only write-behind/reload logic (not battle logic), so no frontend mirror is required.

### What's good

- Small, surgical fix that closes a real (if near-unreachable) gap rather than leaving it merely documented. The comment on the new branch explains *why* the marker is set here and ties it back to #1663/#1632 without over-explaining. Layering is clean — this is API-layer socket serialization, no domain logic leaks in.

---

Issue #854, opened 2026-06-17
Flaky test in Game.Application.Tests fails intermittently on CI (passes locally)


ginderjeremiah:
EDIT: The failed test from 27675635297 and 27676030110 was GetSession_WhenTokenAlreadyCancelled_ThrowsOperationCanceled.

## Problem

A single test in `Game.Application.Tests` fails intermittently on the GitHub Actions `test-backend` job while consistently passing locally. This surfaced on PR #850 (a `Game.Core` battle-stats change that does not touch the Application layer):

- CI run [27675635297](https://github.com/ginderjeremiah/GameServer/actions/runs/27675635297): `Game.Application.Tests` — Failed: 1, Passed: 283, Total: 284
- CI run [27676030110](https://github.com/ginderjeremiah/GameServer/actions/runs/27676030110): `Game.Application.Tests` — Failed: 1, Passed: 283, Total: 284
- Locally (same commit, reused Postgres/Redis containers): 284/284 passed on **4 consecutive runs**.

Every other assembly (Core 578, Infrastructure 37, Api 502) passed on both CI runs. The Microsoft.Testing.Platform console output does not name the failing test — the detail is only in the per-run `Game.Application.Tests_net10.0_x64.log` artifact (`backend-test-results`), which would confirm the exact test.

## Likely culprits

The pattern (passes locally, fails on a loaded CI runner) points to a timing-sensitive integration test using **fixed-delay negative assertions** — wait a fixed `Task.Delay`, then assert that no further work happened. These are inherently fragile under CI scheduling jitter. Strongest candidates:

- `Game.Application.Tests/DataAccess/ReferenceCacheSynchronizerTests.cs:96` — `await Task.Delay(policy.DebounceWindow * 10)` then `Assert.Same(snapshotBeforeOwnMessage, holder.Current)`.
- `Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs:36-37` and `:63-64` — `await Task.Delay(policy.DebounceWindow * N)` then `Assert.Equal(N, ...CompletedReloads/EnteredReloads)`.

## Suggested fix

1. Pull the `backend-test-results` artifact from one of the failed runs to confirm the exact failing test.
[...]

PR #1693, opened 2026-07-07
Add eager cancellation check to RedisCommandBudget.Write

### LINKED ISSUE #1691: RedisCommandBudget.Write lacks the eager cancellation check Read got in #854 (same race, write path)
### Context

Surfaced while babysitting CI on PR #1689 (issue #1647, unrelated feature). `test-backend` failed identically on 3 consecutive CI runs of the same commit — `PlayerWriteBehindTests.SavePlayer_PublishFails_PreservesTheEventForTheNextFlush` (`Game.Application.Tests/Services/PlayerWriteBehindTests.cs:80-115`) failed to observe an `OperationCanceledException` from an already-cancelled `SavePlayer` call, while 6 separate local runs (including one replicating CI's exact `dotnet-coverage`-wrapped invocation) all passed 894/894. PR #1689's diff never touches Redis/data-access code, so this is a pre-existing, unrelated flake.

### Root cause

This is the *exact same bug class* as #854 (fixed in `f749442a`), just on the write side instead of read.

#854's fix added an eager check to `RedisCommandBudget.Read` (`Game.Infrastructure/Redis/RedisCommandBudget.cs`):

```csharp
public static async Task<T> Read<T>(Task<T> command, CancellationToken cancellationToken)
{
    // Honour an already-cancelled budget before awaiting: WaitAsync alone is racy because it returns the
    // inner task unchanged when that task is already complete (it checks IsCompleted before the token), so
    // a command that finished first would silently swallow the cancellation.
    cancellationToken.ThrowIfCancellationRequested();
    try { return await command.WaitAsync(cancellationToken); }
    ...
}
```

**`RedisCommandBudget.Write<T>` and `Write` (void) never got the same fix** — they still go straight to `await command.WaitAsync(cancellationToken)` with no eager pre-check. Two `RedisQueue` methods worked around this locally with their own inline `cancellationToken.ThrowIfCancellationRequested()` (`ReserveNextAsync`, `ReclaimProcessingAsync` — both call out "rather than relying on WaitAsync's racy already-complete check" in their comments), but every other write path — notably `RedisQueue.AddToQueueAsync` (the path `PlayerRepository.SavePlayer`'s write-behind publish uses), `AcknowledgeAsync`, `RemoveAsync`, `GetNextAsync`, and everything in `RedisService` that routes through `ObserveWrite` (`Delete`, `Set`, `GetSet`, `GetDelete`, `CompareAndDelete`, `CompareAndSet`) — still has the race.

If the underlying Redis command (e.g. the `RPUSH` behind `AddToQueueAsync`) completes before `WaitAsync` observes the token, the cancellation is silently swallowed and an already-cancelled caller's write "succeeds" instead of throwing — exactly what `PlayerWriteBehindTests`' test expects **not** to happen. CI's Redis round-trip timing apparently wins that race consistently on the hosted runner; it hasn't reproduced locally.

### Suggested fix

Add the same eager `cancellationToken.ThrowIfCancellationRequested();` to `RedisCommandBudget.Write<T>` and `Write` (void), mirroring `Read`. This fixes every write path in one shared place rather than requiring each call site to duplicate the inline check (as `ReserveNextAsync`/`ReclaimProcessingAsync` currently do — those two inline checks become redundant once the shared helper has it, and could be removed for DRY, though leaving them is harmless).

### Why it matters

Same as #854: this test is on `main` and will intermittently fail `test-backend` for unrelated PRs, forcing reviewers to judge red CI as "probably fine" — which is exactly the trust erosion #854 already called out once.

Labels: bug, tech debt, claude, scope: small.

ginderjeremiah:
## Summary

`RedisCommandBudget.Read` was fixed in #854 to check `cancellationToken.ThrowIfCancellationRequested()` eagerly before awaiting the underlying Redis command, because `Task.WaitAsync` is racy: it returns the inner task unchanged when that task is already complete (checking `IsCompleted` before the token), so a command that finishes first can silently swallow an already-cancelled budget.

`RedisCommandBudget.Write<T>` and `Write` (void) never got the same fix, leaving every write path that routes through them (`RedisQueue.AddToQueueAsync`, `AcknowledgeAsync`, `RemoveAsync`, `GetNextAsync`, and `RedisService`'s `ObserveWrite`-backed methods) exposed to the same race. This was surfaced by `PlayerWriteBehindTests.SavePlayer_PublishFails_PreservesTheEventForTheNextFlush` failing intermittently on CI (hosted-runner Redis round-trip timing apparently wins the race consistently there) while passing reliably locally.

## Changes

- Added the same eager `cancellationToken.ThrowIfCancellationRequested()` to both `RedisCommandBudget.Write<T>` and `Write` (void), mirroring `Read`.
- Added two unit tests (`Write_WhenBudgetAlreadyCancelled_ThrowsWithoutAwaitingTheCommand` and the void-overload equivalent) that pin the fix the same way the existing `Read_WhenBudgetAlreadyCancelled_ThrowsWithoutAwaitingTheCommand` test does — using a never-completing `TaskCompletionSource` to prove the already-cancelled budget is honored without awaiting the command.

Left the redundant inline `ThrowIfCancellationRequested()` checks in `RedisQueue.ReserveNextAsync`/`ReclaimProcessingAsync` in place — they're harmless now that the shared helper covers it, and removing them wasn't necessary to fix the bug.

## Test plan

- [x] `dotnet build Game.sln` — succeeds
- [x] `dotnet test Game.sln --no-build` — all 3010 backend tests pass (`Game.Core.Tests` 1341, `Game.Infrastructure.Tests` 61, `Game.Application.Tests` 894 including the previously-flaky `PlayerWriteBehindTests`, `Game.Api.Tests` 714)
- [x] New tests specifically cover the fixed race for both `Write` overloads

Closes #1691

---
_Generated by [Claude Code](https://claude.ai/code/session_01GRRR1yXPNbzFeDJDMfoauG)_

### DIFF
--- a/Game.Infrastructure.Tests/RedisCommandBudgetTests.cs
+++ b/Game.Infrastructure.Tests/RedisCommandBudgetTests.cs
@@ -50,6 +50,34 @@ namespace Game.Infrastructure.Tests
         }
 
         [Fact]
+        public async Task Write_WhenBudgetAlreadyCancelled_ThrowsWithoutAwaitingTheCommand()
+        {
+            // A command that never completes on its own: if Write awaited it the test would hang, so completing
+            // promptly proves the already-cancelled budget is honoured up front (the racy-WaitAsync guard).
+            var command = new TaskCompletionSource<int>();
+            using var cts = new CancellationTokenSource();
+            var logger = new CapturingLogger();
+            await cts.CancelAsync();
+
+            await Assert.ThrowsAnyAsync<OperationCanceledException>(() => RedisCommandBudget.Write(command.Task, cts.Token, logger, "ignored"));
+            Assert.False(command.Task.IsCompleted);
+            Assert.Empty(logger.Entries);
+        }
+
+        [Fact]
+        public async Task Write_Void_WhenBudgetAlreadyCancelled_ThrowsWithoutAwaitingTheCommand()
+        {
+            var command = new TaskCompletionSource();
+            using var cts = new CancellationTokenSource();
+            var logger = new CapturingLogger();
+            await cts.CancelAsync();
+
+            await Assert.ThrowsAnyAsync<OperationCanceledException>(() => RedisCommandBudget.Write(command.Task, cts.Token, logger, "ignored"));
+            Assert.False(command.Task.IsCompleted);
+            Assert.Empty(logger.Entries);
+        }
+
+        [Fact]
         public async Task Write_AwaitCancelledThenCommandFaults_LogsTheAbandonedFault()
         {
             var command = new TaskCompletionSource<int>();
--- a/Game.Infrastructure/Redis/RedisCommandBudget.cs
+++ b/Game.Infrastructure/Redis/RedisCommandBudget.cs
@@ -41,6 +41,9 @@ namespace Game.Infrastructure.Redis
 
         public static async Task<T> Write<T>(Task<T> command, CancellationToken cancellationToken, ILogger logger, string faultMessage)
         {
+            // Same racy-WaitAsync guard as Read: honour an already-cancelled budget before awaiting, so a command
+            // that finished first can't silently swallow the cancellation.
+            cancellationToken.ThrowIfCancellationRequested();
             try
             {
                 return await command.WaitAsync(cancellationToken);
@@ -54,6 +57,9 @@ namespace Game.Infrastructure.Redis
 
         public static async Task Write(Task command, CancellationToken cancellationToken, ILogger logger, string faultMessage)
         {
+            // Same racy-WaitAsync guard as Read: honour an already-cancelled budget before awaiting, so a command
+            // that finished first can't silently swallow the cancellation.
+            cancellationToken.ThrowIfCancellationRequested();
             try
             {
                 await command.WaitAsync(cancellationToken);


ginderjeremiah (commented):
## Review: eager cancellation check for `RedisCommandBudget.Write` (#1691)

**Verdict:** Approve — the fix is correct, minimal, exactly mirrors the `Read` path, and closes the same racy-`WaitAsync` gap that surfaced as the intermittent `PlayerWriteBehindTests` failure. (Submitted as a comment only because I'm the PR author; GitHub won't accept a self-approval.)

### What's good

- The eager `cancellationToken.ThrowIfCancellationRequested()` is placed identically in both `Write<T>` and `Write` (void), before the `try`/`WaitAsync`, with a comment that ties it back to `Read`'s rationale (`RedisCommandBudget.cs:44-46,60-62`). One shared fix covers every write path (`RedisQueue` mutations + `RedisService.ObserveWrite`) rather than duplicating the check per call site — the right call.
- The two new tests mirror the existing `Read_WhenBudgetAlreadyCancelled…` test faithfully (never-completing `TaskCompletionSource`, assert `OperationCanceledException`, assert the command was never awaited) and cover both overloads. The added `Assert.Empty(logger.Entries)` is a good extra assertion the `Read` test can't make — it pins that the up-front cancel path logs nothing, distinguishing it from the abandoned-fault path.

### Notes (no action needed)

- The redundant inline `ThrowIfCancellationRequested()` checks in `RedisQueue.ReserveNextAsync`/`ReclaimProcessingAsync` are now covered by the shared helper. Leaving them is the right choice: they double as call-site intent documentation ("a requested stop must not start draining another item" / "must not start a reclaim"), so this isn't dead code. The one line worth tidying eventually is the "rather than relying on WaitAsync's racy already-complete check" clause in `ReserveNextAsync`'s comment, which is now the helper's job — but that's cosmetic and not worth a follow-up.

### Test coverage

Sufficient. Domain/logic behavior for both `Write` overloads' already-cancelled path is now unit-tested classically with a deterministic `TaskCompletionSource` (no real Redis round-trip), consistent with the rest of the file. No integration test is needed since the fix is pure in-process cancellation ordering.

Will merge once CI is green.

---

PR #1694, opened 2026-07-07
Test: pin the null-reload branch in SocketHandler player reload path

### LINKED ISSUE #1664: Test: pin the null-reload branch in SocketHandler player reload path
Follow-up from review of #1651.

## Gap

The new reload-before-next-command logic in `SocketHandler` (from #1651) has a branch where the repository's `GetPlayer` returns `null` — in which case the marker should remain set so the reload is retried on the following command. This branch currently has no test coverage.

## Suggested action

Add a unit test covering: marker set → next command runs → `GetPlayer` returns null → session `Player` is not replaced and `PlayerNeedsReload` stays set (retried on the subsequent command).

Non-blocking; surfaced during PR review of #1651.

ginderjeremiah:
## Summary

Follow-up from review of #1651 (issue #1632's write-behind reload path). The reload-before-next-command logic in `SocketHandler.RunCommand` has a branch where the repository's `GetPlayer` returns `null` (the player row is gone) — in that case the marker must stay set so the reload retries on the next command, and the in-memory `Player` must be left untouched rather than replaced with `null`. This branch had no test coverage.

## Changes

- Added `SocketCommandExecutionTests.ExecuteCommand_ReloadFindsNoPlayerRow_KeepsStalePlayerAndRetriesReloadOnNextCommand`, covering: marker set → next command runs → `GetPlayer` returns `null` → session `Player` is not replaced and `PlayerNeedsReload` stays set → a subsequent command retries the reload, this time succeeding, and the session converges.
- Generalized the existing `FakePlayerRepository` test double to accept a sequence of responses (`params Player?[]`) instead of a single fixed player, so it can return `null` once and then a real player on the next call. Existing single-player call sites are unaffected — passing one player keeps returning that same player on every call, matching the prior behavior exactly.

## Test plan

- [x] `dotnet test` (full backend suite: `Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Api.Tests`, `Game.Application.Tests`) — all passing (2,973 tests, 0 failures).

Closes #1664

---
_Generated by [Claude Code](https://claude.ai/code/session_01ADbtGW4zvyxJ9VvdWCkhuZ)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
@@ -314,6 +314,48 @@ namespace Game.Api.Tests.Unit
             Assert.Contains(_logs.Entries, e => e.Level == LogLevel.Error && e.Message.Contains("Abandoned socket command faulted"));
         }
 
+        [Fact]
+        public async Task ExecuteCommand_ReloadFindsNoPlayerRow_KeepsStalePlayerAndRetriesReloadOnNextCommand()
+        {
+            // A reload that finds the player row gone (a pathological case — see RunCommand's comment) must
+            // not clobber the in-memory Player with null, and must leave the marker set so the reload is
+            // retried before every subsequent command until it actually finds the row (#1632).
+            var originalPlayer = new PlayerBuilder().WithId(1).Build();
+            var reloadedPlayer = new PlayerBuilder().WithId(1).Build();
+            var playerRepo = new FakePlayerRepository(null, reloadedPlayer);
+            using var provider = new ServiceCollection()
+                .AddScoped<IUnitOfWork, NoOpUnitOfWork>()
+                .AddScoped<IPlayerRepository>(_ => playerRepo)
+                .BuildServiceProvider();
+
+            var socket = new FakeWebSocket(sendDuration: TimeSpan.Zero);
+            var session = new SessionService(new NoOpSessionStore());
+            session.CreateSession(userId: 1, playerId: 1);
+            session.SetPlayer(originalPlayer);
+            var context = new SocketContext(socket, playerId: 1, session, isAdmin: false, _loggerFactory.CreateLogger<SocketContext>());
+            var handler = new SocketHandler(
+                context,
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: adds test coverage for the null-reload branch in `SocketHandler`'s player reload path (closes #1664)

**Verdict:** Approve _(submitted as a comment since I'm the PR author — GitHub won't accept a gating verdict on one's own PR)_ — the new test faithfully pins the exact branch #1664 called out, is backwards-compatible, and follows the file's existing classical-style patterns.

I traced the test against `SocketHandler.RunCommand` (`Game.Api/Sockets/SocketHandler.cs:280-291`) and `SessionService` (`SetPlayer`/`MarkPlayerNeedsReload`), and it exercises the branch correctly:

- **c1 "Boom"** → `PlayerPersistenceFlushFailedException` → `MarkPlayerNeedsReload()` sets the marker.
- **c2 "Ok"** → marker set → `GetPlayer` returns `null` → the `reloadedPlayer is not null` guard skips `SetPlayer`, so the stale in-memory `Player` survives and the marker stays set (call count 1).
- **c3 "Ok"** → marker still set → `GetPlayer` returns the real row → `SetPlayer` clears the marker and the session converges (call count 2).

The three assertions on `GetPlayerCallCount`, `Assert.Same(...)` reference identity, and `PlayerNeedsReload` are exactly the right observations for this branch.

### Test coverage
- Covers the previously-untested null-then-recovers reload sequence, including the retry-on-every-subsequent-command behavior — matching the issue's suggested action precisely.
- `FakePlayerRepository` was generalized to `params Player?[]` with `responses[Math.Min(GetPlayerCallCount, responses.Length - 1)]`, which keeps single-response callers (lines 247, 282) returning the same player every call. Verified backwards-compatible.
- Classical (Detroit) style with minimal test doubles, consistent with `docs/backend.md` testing guidance and the rest of the file.

### What's good
- Tightly scoped, single-file, test-only change that does exactly what #1664 asked — no scope creep.
- The `FakePlayerRepository` XML-doc comment clearly explains the sequence semantics and why single-response callers are unaffected, which is genuinely useful given the behavior change.

No blocking findings, should-fixes, or nits. I'll merge once CI is green.

---

Issue #1661, opened 2026-07-06
Dead-lettered ChallengeCompleted push never refreshes inventory for a reward item


ginderjeremiah:
Follow-up from review of #1659 (which fixed the analogous gap on the `claimVictory` → `refreshPlayer()` resync path for #1639).

## Problem

`handleServerCommandFailed`'s dead-letter path for a `ChallengeCompleted` push reloads only `playerChallenges`, never the inventory stores. If the dead-lettered challenge completion carried an **item/mod reward**, that reward stays invisible mid-session until a full page reload — the same class of bug #1639/#1659 fixed, just on a different path.

The existing code comment there even claims the reward "surfaces on the next natural load of the inventory stores," which #1639 established is false for a mid-session player who never triggers such a load.

## Suggested fix

Mirror the #1659 fix: after re-pulling the affected player state on a dead-lettered `ChallengeCompleted`, re-run `inventoryManager.initialize()` (idempotent — clears and re-derives from `playerManager.inventoryData`) so a reward item unlocked by the challenge appears immediately.

Consider whether a small shared "resync player + inventory" helper is warranted if this pattern spreads further.

Surfaced during PR review; out of scope for #1659.

PR #1695, opened 2026-07-07
Extract shared resyncPlayerAndInventory() helper

### LINKED ISSUE #1669: Extract a shared "resync player + inventory" helper for the refreshPlayer + inventoryManager.initialize pattern
Follow-up from review of #1668 (which resolved #1661).

## Problem

`refreshPlayer()` (`UI/src/lib/engine/session.ts`) deliberately only re-initializes `playerManager`, **not** the derived `InventoryManager` — its own doc comment documents this footgun and instructs callers to remember to also run `inventoryManager.initialize()` themselves. That coupling is now duplicated at two call sites:

- `EnemyManager.claimVictory` (`UI/src/lib/engine/battle/enemy-manager.ts`)
- `handleServerCommandFailed`'s `ChallengeCompleted` path (`UI/src/lib/engine/engine.ts`, added in #1668)

#1661 explicitly anticipated this: "Consider whether a small shared 'resync player + inventory' helper is warranted if this pattern spreads further." It has now spread to a second site.

## Suggested fix

Introduce a small shared helper (e.g. `resyncPlayerAndInventory()` in `session.ts`) that does `await refreshPlayer(); inventoryManager.initialize();`, and route both call sites through it. This removes the duplicated two-step and eliminates the "must remember the second call" footgun so a future third caller can't silently forget it. Update `refreshPlayer`'s doc comment to point at the helper.

Note the one call-shape difference to preserve: `claimVictory` awaits sequentially, while `handleServerCommandFailed` (a synchronous void event handler) uses `void ...then(...)` — the helper returns a `Promise` so each caller can `await` or `void` it as appropriate.

No user-facing behavior change.

### LINKED ISSUE #1670: Extract a shared resyncPlayerAndInventory() helper (refreshPlayer + inventoryManager.initialize now duplicated at 2 sites)
### Context

Follow-up from review of #1668 (issue #1661).

`refreshPlayer()` followed by `inventoryManager.initialize()` is the resync pattern for "the authoritative player aggregate changed underneath us and the derived inventory needs to catch up." It's now duplicated at two call sites:

- `EnemyManager.claimVictory` (`UI/src/lib/engine/battle/enemy-manager.ts`) — resyncs after a lost/timed-out `DefeatEnemy` response that may have actually succeeded server-side.
- `handleServerCommandFailed` (`UI/src/lib/engine/engine.ts`) — resyncs after a dead-lettered `ChallengeCompleted` push (#1661/#1668).

`refreshPlayer`'s own doc comment (`UI/src/lib/engine/session.ts`) already flags that a caller needing the derived inventory to reflect the refresh must remember the second step — exactly the footgun that's now bitten twice.

### Ask

Extract a small shared helper, e.g. `resyncPlayerAndInventory()` (in `session.ts` alongside `refreshPlayer`, or wherever fits best), that performs both steps, and have both call sites use it. Purely mechanical, no behavior change.

_Surfaced during review of #1668._

ginderjeremiah:
## Summary

`refreshPlayer()` followed by `inventoryManager.initialize()` was the "resync the authoritative player state and re-derive the inventory" pattern used at two call sites:

- `EnemyManager.claimVictory` (`UI/src/lib/engine/battle/enemy-manager.ts`) — resyncs after a lost/timed-out `DefeatEnemy` response that may have actually succeeded server-side.
- `handleServerCommandFailed`'s `ChallengeCompleted` path (`UI/src/lib/engine/engine.ts`)

`refreshPlayer`'s own doc comment already flagged that a caller needing the derived inventory to reflect the refresh must remember the second step — exactly the footgun #1661 anticipated and #1669/#1670 both surfaced once it spread to a second site.

This PR extracts a shared `resyncPlayerAndInventory()` helper and routes both call sites through it, so a future third caller can't silently forget the inventory step. No user-facing behavior change.

## Note on the two linked issues

#1669 and #1670 are near-duplicate follow-ups filed 14 seconds apart requesting the same fix. This PR resolves both; I'll close #1670 as a duplicate of #1669 once this merges.

## Design note

The issue suggested putting the helper in `session.ts` alongside `refreshPlayer`. I put it in `engine.ts` instead: `refreshPlayer` lives in `session.ts`, but the `inventoryManager` singleton instance is constructed and owned by `engine.ts`, which already imports `refreshPlayer` from `session.ts`. Putting the helper in `session.ts` would have required `session.ts` to import back from `engine.ts`, creating a circular module dependency. `engine.ts` avoids that while keeping both call sites just as simple (they already import from the `$lib/engine` barrel, which re-exports `engine.ts`). Updated `refreshPlayer`'s doc comment to point at the new helper instead of describing the manual two-step.

## Test plan

- [x] `npm run lint` (eslint + svelte-check) — passes
- [x] `npm run test:unit:ci` — 295 test files / 3522 tests pass
- [x] Updated `enemy-manager-boss.test.ts` mocks/assertions to match the new single-call shape

Closes #1669
Closes #1670

---
_Generated by [Claude Code](https://claude.ai/code/session_01MzqdugyeLYf7Udkz1VBVYH)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -4,13 +4,12 @@ import { staticData, statistics, playerChallenges } from '$stores';
 import {
 	battleEngine,
 	BattleStage,
-	inventoryManager,
 	onBattleStageChanged,
 	playerManager,
+	resyncPlayerAndInventory,
 	type PlayerBattleState
 } from '../';
 import { logMessage } from '../log';
-import { refreshPlayer } from '../session';
 
 const newEnemyLoadedHook = createHook<[IEnemyInstance]>();
 const notifyNewEnemyLoaded = newEnemyLoadedHook.notify;
@@ -674,12 +673,10 @@ export class EnemyManager {
 			if (defeatResponse.error) {
 				// The transport settles a lost/timed-out response as an error even when the command may have
 				// actually succeeded server-side (see docs/frontend.md's socket request lifecycle) — resync the
-				// authoritative player state rather than leaving exp/level silently diverged for the rest of
-				// the session. refreshPlayer only re-initializes playerManager, so the inventory (an item/mod
-				// reward the lost response may have carried) is re-derived from it separately, mirroring what
-				// startGame does on the initial load.
-				await refreshPlayer();
-				inventoryManager.initialize();
+				// authoritative player state (and the inventory derived from it, in case the lost response
+				// carried an item/mod reward) rather than leaving exp/level silently diverged for the rest of
+				// the session.
+				await resyncPlayerAndInventory();
 			}
 		}
 		// Guard `data` for a possible error response (absent `data`), now that this is the shared
--- a/UI/src/lib/engine/engine.ts
+++ b/UI/src/lib/engine/engine.ts
@@ -44,6 +44,18 @@ import {
 export const inventoryManager = statify(new InventoryManager());
 export const enemyManager = statify(new EnemyManager());
 
+/**
+ * Re-pulls the authoritative player aggregate and re-derives the inventory from it — the shared resync
+ * used whenever a lost/dead-lettered command may have actually succeeded server-side and left exp/level/
+ * inventory silently diverged. See {@link refreshPlayer}'s doc comment for why the two steps don't happen
+ * together automatically. Lives here (rather than alongside `refreshPlayer` in `session.ts`) because it
+ * needs the `inventoryManager` instance owned by this module.
+ */
+export async function resyncPlayerAndInventory(): Promise<void> {
+	await refreshPlayer();
+	inventoryManager.initialize();
+}
+
 export const logicEngine = statify(new LogicalEngine());
 export const renderEngine = statify(new RenderEngine());
 export const battleEngine = statify(new BattleEngine());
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: extract shared `resyncPlayerAndInventory()` helper (#1669 / #1670)

**Verdict:** Approve — clean, correct mechanical refactor that eliminates the `refreshPlayer()` + `inventoryManager.initialize()` footgun exactly as the two follow-up issues asked, with a well-justified placement decision. No blocking findings, should-fixes, or nits. _(Submitted as a `COMMENT` because GitHub won't accept an `APPROVE` from the PR author; the intended verdict is approve.)_

### What's good
- **The placement deviation is the right call and is documented.** Both issues suggested `session.ts`, but the `inventoryManager` singleton is owned by `engine.ts`, which already imports `refreshPlayer` from `session.ts`. Putting the helper in `session.ts` would force `session.ts` to import back from `engine.ts` — a new circular dependency. I confirmed `session.ts` currently imports nothing from `engine.ts`, so the reasoning holds; `engine.ts` is the correct home.
- **The barrel re-export is verified.** `index.ts` does `export * from './engine.ts'`, so `enemy-manager.ts` resolving `resyncPlayerAndInventory` via `'../'` works, and it's only referenced inside a method (not at module-init), so it composes with the pre-existing engine↔enemy-manager barrel cycle without an init-order hazard.
- **Clean removal** — both the `inventoryManager` and `refreshPlayer` imports are gone from `enemy-manager.ts` with no leftover references, and the `void`/`await` call shapes at the two sites are preserved (the one call-shape difference #1669 flagged).
- **Comments are correctly consolidated**, not just moved: the two-step explanation collapses into one place, and `refreshPlayer`'s doc now points callers at the helper instead of describing the manual dance.

### Test coverage
Adequate. `enemy-manager-boss.test.ts` is updated to the new single-call shape — the "successful claim needs no resync" and "errored claim resyncs" assertions now target `resyncPlayerAndInventory` instead of asserting on both underlying calls, which is the correct level of abstraction for a behavior-preserving extraction. No new behavior was introduced, so no new test surface is required.

Nothing to change here.

---

PR #1712, opened 2026-07-07
Guard resolveBossLoss against a superseding transition after its awaits

### LINKED ISSUE #1696: resolveBossLoss doesn't re-check for a superseding transition after its awaits — a re-challenge during the post-loss cooldown clobbers the new boss fight with the stale prepared idle enemy
## Problem

`enemy-manager.ts`'s three boss resolutions handle mid-resolution supersession inconsistently. `watchIdleStage` re-checks `transitioning`/`mode` after its awaits (`enemy-manager.ts:517-519`), and `resolveBossVictory` re-checks `bossLoopActive` after every await (`:580-582`, `:599-601`, `:609-611`). **`resolveBossLoss` (`:621-639`) has no such guard**: it awaits `BattleLost`, then `battleEngine.startLoading(cooldown)`, then unconditionally calls `getNewEnemy(prepared)`.

## Failure scenario (traced)

1. Player loses to the boss → `resolveBossLoss` records the loss, captures the server-prefetched idle battle, and parks on `startLoading(cooldown)`.
2. During the cooldown the player presses **Challenge** again — explicitly supported (`challengeBoss` handles `stage === Loading` with `clientBattleMs: 0`, `:370`). The challenge transitions to boss mode, bumps `fetchGeneration`, and its engine reset resolves the parked `startLoading` early via `finishLoading`.
3. `resolveBossLoss`'s continuation now runs `getNewEnemy(prepared)`. Because this is a **fresh call**, it captures the *post-bump* generation (`:196`), so the generation guard that protects in-flight fetches doesn't help. The prepared battle is still valid (same zone, unchanged player state), so `runFetchLoop` presents it **without a round-trip** (`:237-241`) — the stale idle enemy replaces the just-started boss fight.

Server-side the active battle is the boss; the client is now simulating an idle enemy the server never started. The next `DefeatEnemy` replay diverges (anti-cheat rejection). The same hole applies to a Retreat pressed while `BattleLost` is in flight.

## Fix shape

Mirror the sibling guards: re-check `transitioning`/`mode === 'boss'`-departure (or capture a generation up front and compare) after the `BattleLost` await and again after `startLoading`, before calling `getNewEnemy`. `resolveBossDraw` needs no change (no awaits before its transition).

## Test gap

`enemy-manager-boss.test.ts` has no scenario for a control press landing inside the loss resolution — add one alongside the fix.

ginderjeremiah:
## Summary

`enemy-manager.ts`'s three boss resolutions handled mid-resolution supersession inconsistently. `resolveBossVictory` re-checks `bossLoopActive` after every await, but `resolveBossLoss` had no equivalent guard — it awaited `BattleLost`, then `battleEngine.startLoading(cooldown)`, then unconditionally called `getNewEnemy(prepared)`.

**Failure scenario:** a player loses to the boss and `resolveBossLoss` parks on the post-loss cooldown. If the player presses **Challenge** again during that cooldown, the challenge transitions into boss mode and its engine reset resolves the parked `startLoading` early. `resolveBossLoss`'s continuation then ran `getNewEnemy(prepared)` with a *fresh* generation capture, so the existing fetch-generation guard didn't catch the supersession — the stale prepared idle enemy replaced the boss fight the challenge had just started. The same hole applied to a Retreat pressed while `BattleLost` was still in flight.

## Fix

`resolveBossLoss` couldn't reuse `bossLoopActive` (which requires `mode === 'boss'`) because it deliberately leaves boss mode partway through, via `returnToIdle()`. Instead:

- Snapshot `transitionGeneration` at the top of `resolveBossLoss`, before any awaits.
- Added a `transitionCurrent(generation)` guard (parallel to `bossLoopActive`, but generation-based) that re-checks after the `BattleLost` await and again after `startLoading`, before calling `getNewEnemy`.
- If a challenge/retreat bumped the generation in the meantime, the resolution abandons rather than clobbering the transition that took over.

## Test plan

- [x] Added a test pinning that a retreat landing while `BattleLost` is in flight wins — the loss resolution's own (stale) response no longer re-fetches over the idle fight the retreat already started.
- [x] Added a test reproducing the exact traced scenario: a challenge landing during the post-loss cooldown takes over the boss loop, and the stale prepared idle enemy from the bundled `BattleLost` response is never presented.
- [x] `npx vitest run src/tests/lib/engine/battle/enemy-manager-boss.test.ts` — 51/51 passed
- [x] `npm run test:unit:ci` — 295 files / 3524 tests passed
- [x] `npm run lint` — clean (eslint + svelte-check, 0 errors/warnings)

No backend changes — this is a purely frontend client-side race in the transition/generation bookkeeping; the backend's authoritative state and anti-cheat replay are unaffected.

Closes #1696

---
_Generated by [Claude Code](https://claude.ai/code/session_01V9RxkTLdD7L2igYQDQXxqW)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -487,6 +487,14 @@ export class EnemyManager {
 		return this.started && this.mode === 'boss' && !this.transitioning;
 	}
 
+	/** Whether `generation` (a {@link transitionGeneration} snapshot) is still current — false once a stop
+	 *  or a superseding challenge/retreat (which bumps the generation via {@link beginTransition}) has moved
+	 *  on. Unlike {@link bossLoopActive}, this doesn't require `mode === 'boss'`: boss-loss resolution leaves
+	 *  boss mode itself partway through (via returnToIdle), so it re-checks this after each await instead. */
+	private transitionCurrent(generation: number): boolean {
+		return this.started && generation === this.transitionGeneration;
+	}
+
 	private async watchBattleStage(stage: BattleStage) {
 		// While swapping the active battle, ignore the outgoing battle's resolving stage changes.
 		if (this.transitioning) {
@@ -619,9 +627,19 @@ export class EnemyManager {
 	}
 
 	private async resolveBossLoss() {
+		// Snapshot the transition generation up front: a challenge/retreat pressed during either await below
+		// bumps it (via beginTransition), and its result can resolve the parked startLoading below early
+		// (finishLoading) — so this resolution must not then hand a stale prepared idle battle to a fresh
+		// getNewEnemy call and clobber the fight the supersession already started (#1696).
+		const generation = this.transitionGeneration;
 		// Record the loss explicitly (turning auto-fight off) and drop back to the boss-available
 		// state — the normal idle farm loop — honoring the post-loss cooldown.
 		const lostResponse = await apiSocket.sendSocketCommand('BattleLost');
+		// A retreat/challenge superseded this resolution while BattleLost was in flight; the superseding
+		// transition already owns the loop's state, so abandon rather than clobber it below.
+		if (!this.transitionCurrent(generation)) {
+			return;
+		}
 		if (lostResponse.error) {
 			logMessage(ELogType.Debug, 'There was an error recording the boss loss: ' + lostResponse.error);
 		}
@@ -634,6 +652,11 @@ export class EnemyManager {
 		const cooldown = lostResponse.data?.cooldown ?? 0;
 		if (cooldown > 0) {
 			await battleEngine.startLoading(cooldown);
+			// Re-check after the cooldown: a challenge pressed during it can resolve this parked startLoading
+			// early, so a fresh getNewEnemy below must not present the stale prepared battle over the new fight.
+			if (!this.transitionCurrent(generation)) {
+				return;
+			}
 		}
 		await this.getNewEnemy(prepared);
 	}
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Guard `resolveBossLoss` against a superseding transition after its awaits (#1696)

**Verdict:** Approve — a minimal, correct fix that closes the last unguarded boss-resolution path, consistent with the sibling guards, and covered by both traced scenarios. _(Submitted as a comment because GitHub won't accept an approving event on one's own PR; the call is Approve.)_

The change makes `resolveBossLoss` snapshot `transitionGeneration` before any await and re-check `transitionCurrent(generation)` after both the `BattleLost` and `startLoading` awaits, so a challenge/retreat that supersedes the resolution abandons rather than clobbering the fight the takeover already started.

The load-bearing detail is that the resolution's own `returnToIdle()` (called between the two guards) does **not** touch `transitionGeneration` — only `beginTransition()` does. So the up-front snapshot stays valid through the normal loss path, and the second guard trips *only* on a genuine superseding transition. This is why a generation snapshot is the right tool here and `bossLoopActive` isn't (the resolution deliberately leaves boss mode via `returnToIdle` partway through) — the new `transitionCurrent` doc comment captures exactly that.

### What's good

- All three awaits are accounted for: the two intermediate awaits are guarded, `getNewEnemy` self-guards via `fetchGeneration`, and `resolveBossDraw` correctly needs no change (no awaits before its transition).
- `transitionCurrent` is a clean, well-named helper that parallels `bossLoopActive` rather than duplicating ad-hoc checks.
- The cooldown-guard is correctly placed inside `if (cooldown > 0)` — with `cooldown === 0` there is no await between the first guard and `getNewEnemy`, so no supersession window exists there.

### Test coverage

Both new tests reproduce the exact failure modes from the issue and assert the right invariants:

- The retreat-during-`BattleLost` test asserts exactly one `NewEnemy` send, which proves the loss resolution bailed before re-fetching over the retreat's idle enemy (not merely that the final state looks right).
- The re-challenge-during-cooldown test gates `startLoading`, fires the challenge, and asserts the boss fight stands and the stale prepared idle enemy is never presented.

No mirrored backend test is needed: this is client-side transition/generation bookkeeping, not battle-simulation math (the backend is authoritative and has no equivalent client race), and the PR correctly makes no backend changes.

Nice, surgical fix.

---

PR #1713, opened 2026-07-07
Disable zone navigation while a dedicated-boss fight is engaged

### LINKED ISSUE #1698: Zone navigation stays active during an engaged boss fight — mislabeled boss UI, vanishing Retreat control, and the wrong zone optimistically marked cleared
## Problem

The Fight screen renders `ZoneNav` unconditionally (`UI/src/routes/game/screens/fight/Fight.svelte:6-16`) and `EnemyManager.navigateToZone` (`enemy-manager.ts:154-167`) has no boss-mode handling — it just reassigns `playerManager.currentZone` while `mode` stays `'boss'` and the fight continues. All the boss-view derivations (`bossName`/`zoneName`/`hasBoss`/`cleared` in `boss/boss-view.svelte.ts:13-44`) read the *live* current zone, and `resolveBossVictory` snapshots `clearedZoneId = playerManager.currentZone` at **resolution** time (`enemy-manager.ts:575`), not challenge time.

## Consequences (all traced)

- Navigating mid-fight relabels the boss bar and Zone-Cleared overlay to the *new* zone's boss/zone while the old boss is still being fought.
- Navigating to a **bossless** zone unrenders `BossAffordanceSlot` entirely — taking the **Retreat button** with it, so the player is stuck in the fight until it concludes on its own.
- On victory after a mid-fight navigation, `statistics.markZoneCleared(currentZone)` optimistically seals the **wrong zone** for the rest of the session (statistics load once at boot), and the "next zone unlocked" check evaluates the wrong gate.
- With auto-fight on, the re-challenge sends the new zone's id and engages a different zone's boss.

## Suggested direction

Cheapest coherent fix: disable/hide the zone-nav arrows while `enemyManager.mode === 'boss'` and a fight is engaged (retreat is the sanctioned way out — it already returns to idle before navigating). Alternatively, snapshot the boss's zone at **challenge** time and derive the boss UI + clear/unlock checks from that snapshot. Either way, add a test pinning that a mid-boss-fight navigation can't strand the Retreat control or mis-clear a zone.

Related but distinct: #1690 covers retreat's abandon semantics, not zone-nav during the fight.

ginderjeremiah:
## Problem

`Fight.svelte` renders `ZoneNav` unconditionally, and `EnemyManager.navigateToZone` had no boss-mode handling — it just reassigned `playerManager.currentZone` while `mode` stayed `'boss'` and the fight kept running. All the boss-view derivations (`bossName`/`zoneName`/`hasBoss`/`cleared` in `boss-view.svelte.ts`) read the *live* current zone, and `resolveBossVictory` snapshots `clearedZoneId` at resolution time, not challenge time. So navigating mid-fight could:

- Relabel the boss bar and Zone-Cleared overlay to the new zone's boss while the old boss was still being fought.
- Unrender `BossAffordanceSlot` entirely when navigating to a bossless zone — taking the Retreat button with it and stranding the player until the fight concluded on its own.
- Mark the wrong zone as cleared on victory after a mid-fight navigation, and evaluate the wrong "next zone unlocked" gate.
- With auto-fight on, re-challenge against a different zone's boss.

## Fix

Went with the issue's suggested "cheapest coherent fix": the zone-nav arrows now hard-disable (native `disabled`, non-focusable) whenever `enemyManager.mode === 'boss'`, mirroring the reactive pattern `BossView.engaged` already uses. Retreat remains the sanctioned way out of a boss fight — it already returns to idle before any navigation is possible. `goToZone` also short-circuits on `bossEngaged` as defense in depth, matching the existing `isZoneUnlocked` guard.

This removes the whole class of consequences described in the issue at the root — since navigation is impossible while a boss fight is engaged, the mid-fight zone-reassignment that caused them can no longer happen.

## Test plan

- [x] Added `ZoneNav.test.ts` coverage: both arrows hard-disable with a boss-specific `aria-label` while `mode === 'boss'`; clicking them (even bypassing the disabled attribute) doesn't navigate; navigation stays enabled while idle-farming.
- [x] `npm run test:unit:ci` — 3527/3527 passing.

Closes #1698

---
_Generated by [Claude Code](https://claude.ai/code/session_01KtJ9iuGabQfduhUzBcGoXa)_

### DIFF
--- a/UI/src/routes/game/screens/fight/ZoneNav.svelte
+++ b/UI/src/routes/game/screens/fight/ZoneNav.svelte
@@ -11,7 +11,11 @@
 			class:locked={leftLocked}
 			disabled={leftAbsent}
 			aria-disabled={leftLocked || undefined}
-			aria-label={leftLocked ? 'Previous zone locked' : 'Previous zone'}
+			aria-label={bossEngaged
+				? 'Zone navigation locked during boss fight'
+				: leftLocked
+					? 'Previous zone locked'
+					: 'Previous zone'}
 			onclick={handleClickLeft}
 			onfocus={(e) => focusLock(leftLocked, prevZone?.unlockChallengeId, e)}
 			onblur={hideLock}
@@ -40,7 +44,11 @@
 			class:locked={rightLocked}
 			disabled={rightAbsent}
 			aria-disabled={rightLocked || undefined}
-			aria-label={rightLocked ? 'Next zone locked' : 'Next zone'}
+			aria-label={bossEngaged
+				? 'Zone navigation locked during boss fight'
+				: rightLocked
+					? 'Next zone locked'
+					: 'Next zone'}
 			onclick={handleClickRight}
 			onfocus={(e) => focusLock(rightLocked, nextZone?.unlockChallengeId, e)}
 			onblur={hideLock}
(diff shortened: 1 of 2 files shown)


ginderjeremiah:
## Review — APPROVE WITH NITS

Reviewed against issue #1698. The fix is **correct and complete** for the described bug class: hard-disabling the `ZoneNav` arrows while `enemyManager.mode === 'boss'` (plus a `!bossEngaged` short-circuit in `goToZone`) makes the mid-fight `currentZone` reassignment impossible, which removes the whole class at the root — no more relabeled boss bar, stranded Retreat control, mis-cleared zone, or auto-refight of the wrong boss. Choosing the issue's "cheapest coherent fix" over snapshotting the boss's zone was the right call.

### Correctness checks
- **Reactivity is real:** `enemyManager` is `statify(...)`, so `$derived(enemyManager.mode === 'boss')` reacts — mirrors the established `BossView.engaged` pattern.
- **No transition gap:** `challengeBoss()` sets `mode='boss'` synchronously before awaiting, and `retreatFromBoss()` calls `returnToIdle()` synchronously before awaiting, so nav disables/re-enables on the exact frame. Pre-armed auto-fight stays `idle` until the boss is actually engaged, so nav is correctly still enabled during idle-farming.
- **Auto-fight covered:** mode stays `'boss'` across victory→re-challenge and the Zone-Cleared overlay.
- **The "bypass the disabled button" test is meaningful:** `fireEvent.click` does dispatch through to the handler on a `disabled` button, so removing the `goToZone` guard would actually make that test fail — the defense-in-depth guard is genuinely under test, not vacuous.

CI: `test-frontend`, `verify-frontend`, and `test-e2e` (chromium/firefox/webkit) all green; backend jobs correctly skipped.

### Should-fix in this PR (readability/DRY — all low severity)
1. **Misleading names:** `leftAbsent`/`rightAbsent` (`ZoneNav.svelte:135-136`) are now `true` when a neighbour *is present* but a boss is engaged, so the name no longer describes the condition. Rename to `leftDisabled`/`rightDisabled`.
2. **Duplicated nested ternary:** the three-way `bossEngaged ? … : locked ? … : …` aria-label expression is copy-pasted at `ZoneNav.svelte:14-18` and `:43-47`. Per CLAUDE.md's DRY guidance, extract a tiny `arrowLabel(locked, dir)` helper instead of two copies.
3. **Lock tooltip still fires during a boss fight:** the hover handlers live on the wrapping `<span class="zone-btn-wrap">`, not the button, so if a boss is engaged *and* a neighbour is locked, hovering still pops the challenge-gate tooltip — mildly incongruent with the "locked during boss fight" label. Optionally gate `showLock` on `!bossEngaged`.

These all sit in the code this PR added and are quick tidy-ups, so I'd rather see them here than as a follow-up. No follow-up issues needed. Ping me once they're in and I'll re-review and merge.

---

PR #1714, opened 2026-07-07
Fix multi-byte UTF-8 characters corrupting on 4KB receive boundary

### LINKED ISSUE #1699: SocketContext.ReadMessage decodes UTF-8 per receive-chunk — a multi-byte character straddling the 4KB buffer boundary corrupts the frame
## Problem

`SocketContext.ReadMessage` appends `Encoding.UTF8.GetString(_buffer, 0, result.Count)` per `ReceiveAsync` fill (`Game.Api/Sockets/SocketContext.cs:134`, read loop `:115-152`). `ReceiveAsync` returns at most 4096 bytes per call regardless of client framing and can split a multi-byte UTF-8 code point across two fills; each half then decodes to U+FFFD. The corrupted JSON fails to deserialize and the socket is closed as `MalformedFrame`.

The read path deliberately supports multi-chunk messages up to ~4 MB (`MAX_FRAMES_PER_MESSAGE = 1024`), so the large-message support that exists is deterministically broken for any non-ASCII payload over 4096 bytes. Today's inbound commands are small and effectively ASCII, so this is latent — but any future command carrying user text (names, notes, chat) trips it.

## Fix

Hold a stateful `System.Text.Decoder` (`Encoding.UTF8.GetDecoder()`) across chunks of one message instead of calling `GetString` per fill. The existing multi-frame reassembly test (`SocketContextTests.ReadMessage_ReassemblesMultiFrameTextMessage`, ASCII-only, `Game.Api.Tests/Unit/SocketContextTests.cs:154`) should gain a case with a multi-byte character straddling the buffer boundary.

ginderjeremiah:
## Problem

`SocketContext.ReadMessage` decoded each `ReceiveAsync` fill independently via `Encoding.UTF8.GetString(_buffer, 0, result.Count)`. `ReceiveAsync` returns at most 4096 bytes per call regardless of client framing, so a multi-byte UTF-8 code point could be split across two fills — each half then decoded to U+FFFD (replacement character), corrupting the reassembled JSON and causing the socket to close as `MalformedFrame`.

The read path deliberately supports multi-chunk messages up to ~4 MB (`MAX_FRAMES_PER_MESSAGE = 1024`), so this large-message support was deterministically broken for any non-ASCII payload over 4096 bytes.

## Fix

`ReadMessage` now holds a stateful `System.Text.Decoder` (`Encoding.UTF8.GetDecoder()`) across all the fills of a single message, instead of calling `GetString` per fill. The decoder carries an incomplete trailing byte sequence over to the next fill, so a code point split by the buffer boundary decodes correctly. A fixed `_charBuffer` (sized via `Encoding.UTF8.GetMaxCharCount(MAX_MESSAGE_SIZE)`) avoids a per-fill allocation, mirroring the existing `_buffer` field.

The decoder is created fresh per `ReadMessage` call (i.e. per message), so no state leaks between messages.

## Test plan

- [x] Added `ReadMessage_MultiByteCharacterSplitAcrossFrames_DecodesCorrectly` to `SocketContextTests`, which splits a 4-byte emoji's UTF-8 sequence mid-character across two receive fills and asserts correct reassembly. Required adding a `QueueBytes` helper to the `ScriptedWebSocket` test double so a test can queue exact byte-boundary splits instead of always splitting on a string boundary.
- [x] `dotnet build` — clean.
- [x] `dotnet test` (full solution) — 3012/3012 passing.

Closes #1699

---
_Generated by [Claude Code](https://claude.ai/code/session_01GqEpptQTi5oDi9ydyDokA6)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketContextTests.cs
+++ b/Game.Api.Tests/Unit/SocketContextTests.cs
@@ -166,6 +166,26 @@ namespace Game.Api.Tests.Unit
         }
 
         [Fact]
+        public async Task ReadMessage_MultiByteCharacterSplitAcrossFrames_DecodesCorrectly()
+        {
+            var cancellationToken = TestContext.Current.CancellationToken;
+            var socket = new ScriptedWebSocket();
+            var context = CreateContext(socket);
+
+            // "hel😀lo" — the emoji's 4-byte UTF-8 sequence (F0 9F 98 80) is split mid-character across two
+            // receive fills, reproducing a code point straddling the 4KB buffer boundary (#1699).
+            var fullBytes = Encoding.UTF8.GetBytes("hel😀lo");
+            var firstChunk = fullBytes[..5];
+            var secondChunk = fullBytes[5..];
+            socket.QueueBytes(firstChunk, endOfMessage: false);
+            socket.QueueBytes(secondChunk, endOfMessage: true);
+
+            var message = await context.ReadMessage(cancellationToken).WaitAsync(WaitTimeout, cancellationToken);
+
+            Assert.Equal("hel😀lo", message);
+        }
+
+        [Fact]
         public async Task ReadMessage_CloseFrame_StopsReadingAndReturnsEmpty()
         {
             var cancellationToken = TestContext.Current.CancellationToken;
@@ -230,6 +250,9 @@ namespace Game.Api.Tests.Unit
 
             public void QueueData(string text, bool endOfMessage)
                 => _receives.Enqueue(new ReceiveStep(Encoding.UTF8.GetBytes(text), WebSocketMessageType.Text, endOfMessage));
+
+            public void QueueBytes(byte[] payload, bool endOfMessage)
+                => _receives.Enqueue(new ReceiveStep(payload, WebSocketMessageType.Text, endOfMessage));
 
             public void QueueClose()
                 => _receives.Enqueue(new ReceiveStep([], WebSocketMessageType.Close, EndOfMessage: true));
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Fix multi-byte UTF-8 corruption on the 4KB receive boundary (#1699)

**Verdict:** Approve (submitted as a comment since I'm the PR author) — the fix is the minimal, correct solution the issue asked for, and it's well-tested.

The root cause and fix are both handled correctly:

- A stateful `Encoding.UTF8.GetDecoder()` held across the fills of one message is exactly the right mechanism — it buffers an incomplete trailing byte sequence and completes it on the next `GetChars`, eliminating the per-fill U+FFFD corruption.
- The decoder is created fresh inside `ReadMessage`, so no decoder state leaks between messages.
- `_charBuffer` is sized via `GetMaxCharCount(MAX_MESSAGE_SIZE)` (byteCount + 1 for UTF-8), which is the correct worst case for a single fill including a carried-over char — the field comment accurately explains the +1.
- Sharing `_charBuffer`/`_buffer` as instance fields is safe here: `ReadMessage` runs serially on the read loop, the same assumption the pre-existing `_buffer` already relied on. No new concurrency surface.
- Correctly scoped as backend-only — this is socket transport, not battle logic, and the browser WebSocket API decodes UTF-8 natively, so no frontend mirror is needed.

### Test coverage

`ReadMessage_MultiByteCharacterSplitAcrossFrames_DecodesCorrectly` reproduces the exact failure: the emoji's 4-byte sequence (`F0 9F 98 80`) is split 2/2 across two receive fills via the new `QueueBytes` helper, and the reassembled string is asserted intact. This is the case the issue explicitly requested and it directly exercises the boundary bug. Adequate.

### What's good

Tight, well-scoped fix with rationale comments in exactly the spots a future reader would otherwise be tempted to "simplify" back into the broken per-fill `GetString` — good defensive documentation of a subtle bug.

No blocking, should-fix, or nit findings.

---

PR #1715, opened 2026-07-07
Fix /socket block to use dynamic-DNS re-resolution

### LINKED ISSUE #1700: nginx /socket block uses a literal proxy_pass, bypassing the dynamic-DNS re-resolution the /api block deliberately uses
## Problem

The `/api/` and `/` blocks in `nginx.conf` set the upstream in a **variable** specifically so nginx re-resolves it through the `resolver … valid=10s` directive and survives a backend restart (see the comment at `nginx.conf:16-20`, same pattern at `:48-51`). The `/socket` block (`nginx.conf:29-31`) instead uses a literal `proxy_pass "http://gameserver.railway.internal:8080"`.

With a literal upstream, nginx resolves the name **once at config load** (the `resolver` directive does not apply) and never re-resolves. Consequences:

- After a backend redeploy changes the internal IP, **all WebSocket traffic — the primary game transport — stays pinned to the stale IP until nginx itself restarts**, while REST recovers within 10s.
- nginx fails to start entirely if the name is unresolvable at boot.

## Fix

One line: use the same variable indirection as the `/api/` block (`set $dotnet_backend …; proxy_pass $dotnet_backend;`) in the `/socket` location.

ginderjeremiah:
## Problem

The `/api/` and `/` blocks in `nginx.conf` set their upstream in a **variable** specifically so nginx re-resolves it through the `resolver … valid=10s` directive and survives a backend restart. The `/socket` block instead used a literal `proxy_pass "http://gameserver.railway.internal:8080"`.

With a literal upstream, nginx resolves the name once at config load and never re-resolves — so after a backend redeploy changes the internal IP, all WebSocket traffic (the primary game transport) would stay pinned to the stale IP until nginx itself restarted, while REST recovered within 10s.

## Fix

Applied the same variable-indirection pattern already used by the `/api/` block: `set $dotnet_backend "…"; proxy_pass $dotnet_backend;` in the `/socket` location.

## Test plan

- [x] Validated the resulting config with a real `nginx:alpine` container: mounted `nginx.conf` as the config template, ran `envsubst` + `nginx -t` — syntax OK, config test successful.
- [x] No application code paths affected; this is an infra config-only change with no unit/integration test surface.

Closes #1700

---
_Generated by [Claude Code](https://claude.ai/code/session_01DWJZZbgpnQ7GSoF26xNDVU)_

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -27,8 +27,11 @@ server {
 
 		# 2. Dedicated WebSocket rerouting path
     location /socket {
-        proxy_pass "http://gameserver.railway.internal:8080";
-        
+        # Use a variable for the hostname to prevent Nginx from crashing if a service is restarting
+        set $dotnet_backend "http://gameserver.railway.internal:8080";
+
+        proxy_pass $dotnet_backend;
+
         proxy_http_version 1.1;
 				proxy_set_header Host $host;
 				proxy_set_header X-Real-IP $remote_addr;


ginderjeremiah:
## Review — APPROVE ✅

Reviewed the change against the issue (#1700) and the surrounding `nginx.conf`.

**What it does:** Converts the `/socket` block from a literal `proxy_pass "http://gameserver.railway.internal:8080";` to the variable-indirection pattern (`set $dotnet_backend "…"; proxy_pass $dotnet_backend;`) already used by `/api/`. With a literal upstream, nginx resolves the name once at config load and ignores the `resolver … valid=10s` directive, so after a Railway redeploy WebSocket traffic (the primary game transport) stays pinned to the stale IP until nginx restarts. Routing through a variable forces runtime re-resolution, bringing the socket path's recovery in line with REST (~10s). Correct and complete fix for #1700.

**Correctness checks:**
- `resolver` is in http/top-level scope, so it applies to this block — the indirection actually takes effect.
- URI handling is unchanged: neither the old nor new `proxy_pass` has a URI part, so nginx forwards the full normalized request URI in both cases. Nothing dropped, no path needed — exact parity with the `/api/` block it mirrors.
- WebSocket-specific config preserved: `proxy_http_version 1.1`, `Upgrade`/`Connection $connection_upgrade`, and the 3600s timeouts are all untouched.
- No other literal `proxy_pass` remains — the only three upstreams are `/api/` (already a variable), `/socket` (fixed here), and `/` (already a variable, distinct `$node_backend`).

**Nits (non-blocking, pre-existing — not introduced here):** the file mixes tab/space indentation across location blocks, and `set $dotnet_backend …` is now duplicated in `/api/` and `/socket` (hoisting to `server` scope is possible but per-location `set` is idiomatic). The added lines correctly match the local style. Not worth a follow-up.

CI: path-filtered jobs correctly skipped; `detect-changes` and `all-checks-passed` green. Validated with a real `nginx -t` per the PR's test plan.

Merging.

---

PR #1716, opened 2026-07-07
FE test: assert each Lesson's screenKey resolves to a real ScreenDef.key

### LINKED ISSUE #1673: FE test: assert each Lesson's screenKey resolves to a real ScreenDef.key
### Context

Follow-up from #1605 (authoring the three taught-by-blurb lessons: `idle-loop-basics`, `crit-dodge-variance`, `cooldown-charging`).

`Lesson.ScreenKey` is a plain string, not an FK — screens are a frontend-only registry (`UI/src/routes/game/screens/screen-defs.ts`'s `GAME_SCREENS`) with no backend representation, so `ProgressionGraphChecker.CheckLessons` (`Game.Application/Content/ProgressionGraphChecker.cs`) deliberately does not validate that a lesson's `screenKey` matches a real screen — its comment says this is intentionally left to a frontend-only test, mirroring how the `anchorKey`-resolves-to-a-registered-anchor check is FE-only (#1592).

No such test exists yet. Today a typo'd or stale `screenKey` (e.g. after a screen is renamed/removed) would silently degrade — nothing would catch it in CI.

### Ask

Add a light Vitest test that loads the committed `content/lessons.json` (or the seeded reference-data fixture, whichever the FE test suite already has a pattern for reading committed content from) and asserts every live lesson's `screenKey` matches a `key` in `GAME_SCREENS`. Was surfaced as an explicit judgment call in #1605 rather than bundled, since there's no established pattern yet for a frontend test to read `content/*.json` and it may be a small piece of new test infrastructure rather than a one-line assertion.

ginderjeremiah:
## Summary

`Lesson.ScreenKey` is a plain string, not an FK — screens are a frontend-only registry (`GAME_SCREENS` in `UI/src/routes/game/screens/screen-defs.ts`) with no backend representation. The backend's `ProgressionGraphChecker.CheckLessons` deliberately skips validating `screenKey` against real screens, leaving that check to the frontend — but no such frontend test existed yet, so a typo'd or stale `screenKey` (e.g. after a screen rename/removal) would silently degrade at runtime with nothing catching it in CI.

## Change

Added a test in `UI/src/tests/routes/game/screens/screen-defs.test.ts` (the existing home for `GAME_SCREENS` invariant tests) that:

- Imports the committed `content/lessons.json` export directly (`resolveJsonModule` is already enabled in `tsconfig.json`).
- Filters to live (non-retired) lessons.
- Asserts every live lesson's `screenKey` matches a `key` in `GAME_SCREENS`.

This is new (small) test infrastructure — no frontend test previously read `content/*.json` directly — since the issue flagged there wasn't yet an established pattern for it.

I verified the test actually catches a regression: temporarily corrupting `content/lessons.json` with a bogus `screenKey` made the new test fail with a clear message, then reverted.

## Test plan

- [x] `npm run test:unit:ci` — 3525/3525 passing (295/295 test files).
- [x] Manually verified the assertion fails on a corrupted `screenKey`, then restored the original content file.

Closes #1673

---
_Generated by Claude Code_

---
_Generated by [Claude Code](https://claude.ai/code/session_011xvRaLUmmJc9SrxbvUUG4t)_

### DIFF
--- a/UI/src/tests/routes/game/screens/screen-defs.test.ts
+++ b/UI/src/tests/routes/game/screens/screen-defs.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import { GAME_SCREENS, visibleScreens, type ScreenDef } from '$routes/game/screens/screen-defs';
 import { ERole } from '$lib/api';
+import lessonsContent from '../../../../../../content/lessons.json';
 
 const screens: ScreenDef[] = [
 	{ key: 'fight', label: 'Fight', group: 'combat', built: true },
@@ -64,5 +65,20 @@ describe('GAME_SCREENS', () => {
 		const wipWithComponent = GAME_SCREENS.filter((s) => !s.built && s.component).map((s) => s.key);
 
 		expect(wipWithComponent).toEqual([]);
+	});
+});
+
+// `Lesson.screenKey` is a plain string, not an FK — screens are a frontend-only registry with no
+// backend representation, so the backend progression-graph lint deliberately leaves this check to
+// the frontend (see ProgressionGraphChecker.CheckLessons and #1673).
+describe('committed lesson content (content/lessons.json)', () => {
+	it('gives every live lesson a screenKey that resolves to a real ScreenDef.key', () => {
+		const screenKeys = new Set(GAME_SCREENS.map((s) => s.key));
+		const badScreenKeys = lessonsContent
+			.filter((lesson) => !lesson.retiredAt)
+			.filter((lesson) => !screenKeys.has(lesson.screenKey))
+			.map((lesson) => `${lesson.key} -> "${lesson.screenKey}"`);
+
+		expect(badScreenKeys).toEqual([]);
 	});
 });


ginderjeremiah:
## Review — APPROVE ✅

Reviewed against issue #1673.

**What it does:** Adds a single Vitest to `UI/src/tests/routes/game/screens/screen-defs.test.ts` that imports the committed `content/lessons.json`, filters to live (non-retired) lessons, and asserts every lesson's `screenKey` matches a `key` in `GAME_SCREENS`. This closes the exact gap #1673 describes: `ProgressionGraphChecker.CheckLessons` intentionally defers `screenKey` validation to the frontend, and no frontend test existed — so a typo'd or stale key would degrade silently at runtime. The fixture has 3 live lessons (`screenKey: "fight"`), so the test exercises real data and would fail on a bad key; the author verified this manually by corrupting the file. Minimal, correct, lands in the conventional location, clear failure message.

**CI:** green — `test-frontend`, `verify-frontend`, and `test-e2e` (chromium/firefox/webkit) all pass; backend jobs correctly skipped.

**Nits (non-blocking):** the assertion passes vacuously if the live-lesson set is ever empty, and the import uses a deep six-level relative path outside `UI/`. Both are minor and the issue explicitly acknowledged there was no established content-fixture pattern yet, so this is acceptable as the first instance. Filed as follow-up #1719 (test hardening + reusable import pattern, plus confirming the sibling `anchorKey` check from #1592).

Follow-up tracked, so merging.

---

PR #1717, opened 2026-07-07
Wire real tutorialAnchor targets into the Fight screen

ginderjeremiah:
## Summary

#1592 built the coach-mark tour mechanism (`TourPlayer.svelte`, the `use:tutorialAnchor` action and its anchor registry), but no screen component anywhere in `UI/src` registered an anchor. As a result, every step in the three lessons authored in #1605 used `anchorKey: null` and degraded to a centered, unanchored callout instead of pointing at the concrete UI it describes.

## Change

Registered `use:tutorialAnchor` on the three Fight-screen elements the issue calls out, keyed per side (`{key}-player` / `{key}-enemy`) so both the player and enemy `BattlerCard`/`Skills` instances can register safely without one clobbering the other's slot:

- `BattlerCard.svelte` — the `.hp-bar-slot` div → `fight-hp-bar-player` / `fight-hp-bar-enemy`
- `Skills.svelte` — the `.skills-row` div → `fight-skill-bar-player` / `fight-skill-bar-enemy`
- `CombatFloaters.svelte` — the `.floaters` overlay div → `fight-combat-log-player` / `fight-combat-log-enemy`

Then pointed the relevant lesson steps in `content/lessons.json` at these keys:

- `idle-loop-basics` step 2 ("the bars above each combatant track HP") → `fight-hp-bar-player`
- `crit-dodge-variance` step 1 (the crit) → `fight-combat-log-enemy` — a crit-capable skill exists in every starter kit, so the player's own hit landing the crit is the reliable path (per the lesson's own design notes), and that lands on the enemy card.
- `crit-dodge-variance` step 2 (the dodge) → `fight-combat-log-player` — "the same goes on defense" refers to an incoming attack against the player missing.
- `cooldown-charging` step 1 → `fight-skill-bar-player`

A few steps stay `anchorKey: null` on purpose — they're general/scene-setting statements (the intro welcome, "you don't have to watch every fight", "cooldown speed compounds with your other stats") that don't point at one specific control, so the centered-callout fallback is the right treatment for them.

**Judgment call worth flagging:** picking *which* side's element (player vs. enemy) each step anchors on isn't spelled out in the issue — I inferred it from the lesson text and design notes as described above. If that reads wrong once seen in the actual tour, it's a one-line `anchorKey` change in `content/lessons.json`, not a structural one.

[...]

PR #1721, opened 2026-07-08
Fix flaky PlayerProgressRepository cache-hash integration test (#1718)

### LINKED ISSUE #1718: Flaky integration test: PlayerProgressRepositoryIntegrationTests.Save_OnlyWritesTheRowsThatChanged_LeavingEarlierCachedRowsInTheHashUntouched
## Summary

`Game.Application.Tests.DataAccess.PlayerProgressRepositoryIntegrationTests.Save_OnlyWritesTheRowsThatChanged_LeavingEarlierCachedRowsInTheHashUntouched` (`Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs:336`) fails intermittently, turning `test-backend` (and the `all-checks-passed` gate) red on PRs that don't touch backend code at all.

Observed on PR #1717 (a frontend-/content-only change): `test-backend` failed on the initial run and again on a `rerun_failed_jobs`, then the same test **passed on a subsequent local run of the identical commit** — a textbook non-deterministic failure.

## Reproduction

On the PR head commit (no backend changes), running `Game.Application.Tests` repeatedly:

- Run 1 → `Failed: 1, Passed: 893` — this test fails.
- Run 2 → `Failed: 0, Passed: 894` — all green.

Same binary, same DB/Redis containers, no code change between runs.

## Failing assertion

```
Expected the battle-completion save to touch more than one statistic row.
   at ...PlayerProgressRepositoryIntegrationTests.Save_OnlyWritesTheRowsThatChanged_LeavingEarlierCachedRowsInTheHashUntouched() ... :line 336
```

```csharp
await repo.Save(progress); // first save: creates the hash with several statistic fields
...
var fieldsAfterFirstSave = await redis.HashGetAllAsync(hashKey);
Assert.True(fieldsAfterFirstSave.Length > 1, "Expected the battle-completion save to touch more than one statistic row.");
```

The first `Save` intermittently leaves the `Progress_{playerId}` Redis hash with ≤1 field when the test reads it back.

## Likely cause (to confirm)

The sibling test `Save_DuringAPlayerSave_DefersToTheSharedBatchFlush_InsteadOfItsOwnRoundTrip` shows this repository can **defer a save into a shared batch flush** rather than doing its own round-trip (#1635). If the first `Save` here is ever coalesced into a deferred/async flush, the immediate `HashGetAllAsync` reads the hash before the HSET has fully landed — producing a partial (or empty) field set and the `> 1` assertion failing. The test likely needs to await/observe the actual flush completion before reading Redis, rather than assuming `Save` has synchronously written the hash.

Worth verifying whether cross-test Redis state (the setup flush vs. the per-test `Progress_{playerId}` key) also contributes.

## Impact

No user-facing effect — this is a test-suite defect. But it blocks CI non-deterministically on unrelated PRs, so it should be fixed rather than papered over with reruns.

Tags: bug, tech debt, claude, scope: small

ginderjeremiah:
## Summary

`PlayerProgressRepositoryIntegrationTests.Save_OnlyWritesTheRowsThatChanged_LeavingEarlierCachedRowsInTheHashUntouched` failed intermittently, turning `test-backend` red on unrelated PRs (observed on #1717).

## Root cause

`PlayerProgressRepository.Save` advances the Redis cache via `ICacheService.HashSetAndForget`, which (`Game.Infrastructure/Cache/Redis/RedisService.cs:167-189`) dispatches its HSET Lua script with `CommandFlags.FireAndForget` — a deliberate perf choice to keep it off the hot inbound-message path. That means `Save` can return before the write has actually landed on the Redis server.

The test read the `Progress_{playerId}` hash on a **separate** Redis connection immediately after `Save` returned. Most of the time the fire-and-forget write had already landed by then, but occasionally the test's read raced it and observed a partial (≤1 field) hash, failing the `> 1` assertion.

The issue's own hypothesis pointed at the write-behind batch flush; that's not quite it (this test's `Save` calls take the "standalone" branch, which already awaits the batch flush before calling `AdvanceCache()`) — the race is specifically in the fire-and-forget `HSET` itself, not the batch/event-queue mechanism.

This isn't a new problem class: `RedisServiceIntegrationTests.cs` already has a `WaitForHashFieldCountAsync` polling helper for exercising `HashSetAndForget`, for exactly this reason. This PR applies the same pattern here instead of reinventing it.

## Fix

- Added a `WaitForHashFieldCountAsync` helper (mirroring the existing one in `RedisServiceIntegrationTests.cs`, adapted to the raw `IDatabase` this test file uses) that polls the hash until the expected field count appears, with a 5s timeout.
- Replaced both immediate `redis.HashGetAllAsync(hashKey)` reads in the flaky test with calls to the new helper.

No production code changed — `HashSetAndForget`'s fire-and-forget behavior is intentional and correct; the test just needs to account for it, like its sibling tests already do.

## Note on the issue

I noticed issue #1718 has a comment from an unaffiliated external account linking to a `.zip` "fix" — I did not download or use that file. This fix was derived independently from reading the actual repository/cache code.

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test Game.Application.Tests` (full suite, 894 tests) — ran 3x back-to-back, 0 failures each run (previously ~50% flake rate per the issue's own repro)

Closes #1718

---
_Generated by [Claude Code](https://claude.ai/code/session_01DB1DF6BedWwqGggh7t6iNi)_

### DIFF
--- a/Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs
@@ -332,7 +332,10 @@ namespace Game.Application.Tests.DataAccess
                 await repo.Save(progress); // first save: creates the hash with several statistic fields
             }
 
-            var fieldsAfterFirstSave = await redis.HashGetAllAsync(hashKey);
+            // The cache advance is a fire-and-forget HSET (Save returns once the write is dispatched, not once
+            // it lands), so poll rather than reading immediately — otherwise this races the write and sees a
+            // partial (or empty) hash (#1718).
+            var fieldsAfterFirstSave = await WaitForHashFieldCountAsync(redis, hashKey, minCount: 2);
             Assert.True(fieldsAfterFirstSave.Length > 1, "Expected the battle-completion save to touch more than one statistic row.");
 
             using (var scope = CreateScope())
@@ -346,7 +349,7 @@ namespace Game.Application.Tests.DataAccess
             // The second save's HSET only wrote the one new proficiency field — every field the first save
             // wrote is still present, proving a save's cache write is scoped to its own dirty rows rather than
             // re-serializing (and re-sending) the whole cached snapshot (#1635).
-            var fieldsAfterSecondSave = await redis.HashGetAllAsync(hashKey);
+            var fieldsAfterSecondSave = await WaitForHashFieldCountAsync(redis, hashKey, minCount: fieldsAfterFirstSave.Length + 1);
             Assert.Equal(fieldsAfterFirstSave.Length + 1, fieldsAfterSecondSave.Length);
             foreach (var field in fieldsAfterFirstSave)
             {
@@ -385,6 +388,27 @@ namespace Game.Application.Tests.DataAccess
         {
             var options = ConfigurationOptions.Parse(Containers.PubSubConnectionString);
             return await ConnectionMultiplexer.ConnectAsync(options);
+        }
+
+        // HashSetAndForget dispatches its HSET with CommandFlags.FireAndForget (RedisService.cs), so it can
+        // return before the write lands on the server — a read on a separate connection right after a Save can
+        // race it. Poll until the expected field count shows up rather than asserting on a single read (#1718).
+        private static async Task<HashEntry[]> WaitForHashFieldCountAsync(IDatabase redis, string hashKey, int minCount, int timeoutMs = 5000)
+        {
+            var deadline = DateTime.UtcNow.AddMilliseconds(timeoutMs);
+            HashEntry[] fields;
+            do
+            {
+                fields = await redis.HashGetAllAsync(hashKey);
+                if (fields.Length >= minCount)
+                {
+                    return fields;
+                }
+
+                await Task.Delay(25);
+            } while (DateTime.UtcNow < deadline);
+
+            return fields;
         }
 
         private static async Task<ProgressUpdatedEvent> DequeueProgressEvent(IDatabase redis)


ginderjeremiah (commented):
## Review: Fix flaky `PlayerProgressRepository` cache-hash integration test (#1718)

**Verdict:** Approve _(posted as a COMMENT since I'm the PR author and GitHub won't accept a self-approval)_ — correct root-cause and the minimal, right-shaped fix.

I verified the reasoning against the code and it holds up:

- `HashSetAndForget` genuinely dispatches its Lua script with `CommandFlags.FireAndForget` (`Game.Infrastructure/Cache/Redis/RedisService.cs:187`), so `Save` can return before the write lands — the race is real and, as the PR notes, is in the fire-and-forget HSET itself, not the batch-flush path the issue hypothesized. Good catch correcting the issue's premise.
- `PlayerProgressRepository.AdvanceCache` writes the whole dirty set through a **single** `HashSetAndForget` call, and that script `hset`s every field atomically. So the hash goes 0 → full-set in one shot — there's no risk of `WaitForHashFieldCountAsync(..., minCount: 2)` capturing a partial first-save set and skewing the `fieldsAfterFirstSave.Length + 1` assertion on the second read. The fix is robust, not just usually-lucky.
- Both call sites poll *then* assert, so a genuine timeout still surfaces as the original meaningful assertion failure rather than a silent pass — the helper returning-on-timeout (instead of `Assert.Fail`) is fine here.
- Mirrors the established `WaitForHashFieldCountAsync` pattern already in `RedisServiceIntegrationTests.cs` rather than inventing a new mechanism.

### Nits

- The suite now has ~4 "poll until condition or timeout" helpers across the integration tests (`WaitForHashFieldCountAsync` in two files, `WaitUntilValueEqualsAsync`, etc.). Not worth blocking this fix — the two hash helpers operate on different abstractions (`ICacheService` vs raw `IDatabase`) — but a small shared test utility would remove the duplication. Filing as a follow-up so it isn't lost.

### Test coverage

Test-only change; no production code touched. The fix makes the existing assertions deterministic without weakening them. Appropriate.

### What's good

Thorough, honest write-up: it corrects the issue's stated hypothesis, cites exact line numbers, and explicitly notes the untrusted `.zip` comment on the issue was not used. Exactly the right amount of change for a flaky-test fix.

---

PR #1724, opened 2026-07-08
Mark PlayerData's always-set value-type properties as required

### LINKED ISSUE #1680: Decide whether PlayerData's always-set value-type properties should be marked required
Follow-up from review of #1679 (issue #1623).

## Context

#1679 marked `PlayerData.PlayerRating` (`Game.Api/Models/Player/PlayerData.cs:45`) `required` to close the "silently defaults to 0 if a future construction path forgets to set it" gap, on the premise that every other property on `PlayerData` was already `required`.

That premise doesn't fully hold: several value-type properties are *not* marked `required` and share the exact same footgun —

- `Id`
- `Level`
- `Exp`
- `CurrentZone`
- `StatPointsGained`
- `StatPointsUsed`

All of them are always set by the sole construction site (`PlayerData.FromPlayer`), same as `PlayerRating` was.

## Ask

Decide the rule intentionally, rather than leaving `PlayerRating` as the only required value type on the model:

- Mark all always-set value-type properties `required` for consistency (mechanical change, mirrors #1679's fix — verify `FromPlayer` still sets all of them, which it does today), **or**
- Explicitly accept that value types stay unmarked as a deliberate convention (e.g. reserve `required` for reference types where the default is `null`/uninitialized-collection risk) and note that reasoning somewhere so it doesn't get flagged again.

Low priority / non-blocking — surfaced as a nit during #1679's review, not a defect.

ginderjeremiah:
## Summary

Follow-up from review of #1679. That PR marked `PlayerData.PlayerRating` `required` on the premise that every other property on `PlayerData` was already `required`, but that premise didn't fully hold: `Id`, `Level`, `Exp`, `CurrentZone`, `StatPointsGained`, and `StatPointsUsed` were left unmarked despite sharing the exact same footgun (a value type silently defaulting to `0` if a future construction path forgets to set it).

## Decision

The issue asked to pick a rule intentionally: mark all always-set value-type properties `required` for consistency, or explicitly adopt a convention that value types stay unmarked.

I went with **marking them all `required`**. The rest of the codebase already establishes this as the standing convention — `required` on always-set value types (not just reference types) appears across 46 files (`Game.Core`, `Game.Api`, `Game.DataAccess`, etc.), and `PlayerRating` itself (#1679) is a direct precedent on this same model. Carving out a value-type exception here would be inconsistent with the rest of the domain, not a real simplification.

## Changes

- `Game.Api/Models/Player/PlayerData.cs`: marked `Id`, `Level`, `Exp`, `CurrentZone`, `StatPointsGained`, `StatPointsUsed` as `required`.
- Verified `PlayerData.FromPlayer` — the sole construction site solution-wide (confirmed via `grep -rn "new PlayerData"`) — already sets all six unconditionally, so this is a no-op behavior change.

## Test plan

- [x] `dotnet test Game.sln --no-build --no-restore` — 3012/3012 passing

No frontend changes.

Closes #1680

---
_Generated by [Claude Code](https://claude.ai/code/session_01F5jAg1irJJoGviFwfTtkow)_

### DIFF
--- a/Game.Api/Models/Player/PlayerData.cs
+++ b/Game.Api/Models/Player/PlayerData.cs
@@ -6,15 +6,15 @@ namespace Game.Api.Models.Player
 {
     public class PlayerData : IModel
     {
-        public int Id { get; set; }
+        public required int Id { get; set; }
         public required string Name { get; set; }
-        public int Level { get; set; }
-        public int Exp { get; set; }
+        public required int Level { get; set; }
+        public required int Exp { get; set; }
         public required List<BattlerAttribute> Attributes { get; set; }
         public required List<UnlockedSkill> UnlockedSkills { get; set; }
-        public int CurrentZone { get; set; }
-        public int StatPointsGained { get; set; }
-        public int StatPointsUsed { get; set; }
+        public required int CurrentZone { get; set; }
+        public required int StatPointsGained { get; set; }
+        public required int StatPointsUsed { get; set; }
         public required List<LogPreference> LogPreferences { get; set; }
         public required List<PlayerLesson> Lessons { get; set; }
         public required InventoryData InventoryData { get; set; }


ginderjeremiah (commented):
## Review: mark `PlayerData`'s always-set value-type properties `required` (closes #1680)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't accept a gating review from me) — the change is a correct, well-justified resolution of the #1680 nit with no findings.

I verified the PR's central claim independently:

- `new PlayerData` has exactly one construction site solution-wide — `PlayerData.FromPlayer` (`Game.Api/Models/Player/PlayerData.cs:65`).
- All six newly-`required` properties are set unconditionally in that initializer: `Id` (67), `Level` (69), `Exp` (70), `CurrentZone` (71), `StatPointsGained` (72), `StatPointsUsed` (73). So this is a genuine no-op behavior change that only tightens the compile-time contract.

The decision is the right one. #1680 explicitly asked to pick a rule intentionally, and marking these `required` matches the standing convention (`PlayerRating` from #1679 on this same model, plus `required` value types across the domain). Carving out a value-type exception would have been the inconsistent choice.

### Test coverage

None required, and none is missing. This is a compile-time contract change with no runtime logic — the build itself enforces the `required` members, and the project's testing rules target domain logic, of which there is none here.

### What's good

Correctly scoped to the one file the issue named, verified `FromPlayer` still satisfies every new `required` before relying on the no-op claim, and the PR body documents the reasoning behind the convention choice so it won't get re-flagged.

---

PR #1725, opened 2026-07-08
Let retreatFromBoss force-abandon a still-in-progress boss battle

### LINKED ISSUE #1690: retreatFromBoss() can't force-abandon a still-in-progress boss battle (unlike ChallengeBoss)
### Context

Follow-up from review of #1689 (issue #1647).

#1689 made `EnemyManager` set `mode` from the authoritative `isBossBattle` flag whenever a server-handed-back battle is presented — including the ordinary `NewEnemy` response (`enemy-manager.ts`'s `runFetchLoop`). This surfaced a pre-existing asymmetry between the two ways a player can leave a boss fight:

- `challengeBoss()` sends the `ChallengeBoss` command, which the backend deliberately treats as "abandon whatever is running and start the boss fight" (`BattleService.StartBossBattle` discards any in-progress handoff via `AbandonBattle`, no hand-back).
- `retreatFromBoss()` calls `returnToIdle()` then `getNewEnemy()`, which sends the ordinary `NewEnemy` command. Server-side, `StartBattle`'s `AbandonBattle` path only discards a battle that has actually concluded (or hit the 2-minute cap) — a boss fight still genuinely in progress is hidden back **unchanged**, `IsBossBattle` included.

Post-#1689 the client now honestly reflects that: pressing "retreat" while the boss fight is still active flips `mode` right back to `'boss'` (the same fight resumes, correctly presented). This is arguably better than the pre-#1689 behavior (the same unconcluded fight silently continued under `idle` chrome), but it means retreat still cannot **instantly** abandon a genuinely-ongoing boss fight the way `ChallengeBoss` can — the player has to wait for the fight to reach a real conclusion (win/loss/draw/cap) before `NewEnemy` will hand back a fresh idle spawn instead.

### Ask

Decide whether retreat should force-abandon a still-in-progress boss fight symmetrically with `ChallengeBoss` (e.g. by having the frontend route retreat through a similar "discard, don't hand back" server command/parameter), or whether the current wait-it-out behavior is acceptable given a boss fight is capped at `DefaultMaxBattleMs` (2 minutes) anyway.

### Related test gap

`enemy-manager-boss.test.ts`'s retreat test currently mocks `NewEnemy` to return a non-boss enemy, so it never exercises the still-in-progress-boss hand-back path on retreat. If the behavior above is intentionally kept as-is, consider adding a test pinning it (retreat while boss fight is still active → `mode` stays/returns to `'boss'`).

Non-blocking; surfaced during review of #1689.

ginderjeremiah:
## Summary

Closes #1690.

`retreatFromBoss()` couldn't instantly leave a still-in-progress boss fight, unlike `challengeBoss()`:

- `challengeBoss()` → `ChallengeBoss` → `BattleService.StartBossBattle` unconditionally discards any in-progress battle (even a genuinely-ongoing one) and proceeds — a deliberate override.
- `retreatFromBoss()` → `NewEnemy` → `BattleService.StartBattle` only discards a battle that's actually *concluded*; a still-active fight (boss or idle) is handed back **unchanged**, `IsBossBattle` included — so post-#1689, pressing retreat while the boss fight is still going flips the client's `mode` right back to `'boss'` and the same fight resumes.

## Decision

The issue asked to decide between force-abandon symmetry with `ChallengeBoss`, or keeping the current wait-it-out behavior (bounded by the 2-minute battle cap anyway). I went with **symmetry**: a player choosing to retreat should be able to leave immediately, the same way choosing to challenge the boss already discards whatever's running. `game-design.md`'s Battle Outcomes section already frames "retreating from a boss" as a walk-away action, which reads as the intended UX.

## Changes

- **Backend**: `BattleService.StartBattle` gains an opt-in `forceAbandon` parameter — when the existing handoff is still in progress, `forceAbandon: true` discards it and starts fresh instead of handing it back, mirroring exactly what `StartBossBattle` already does unconditionally for `ChallengeBoss`. Threaded through a new `NewEnemyRequest.ForceAbandon` field (codegen'd to the frontend). Ordinary `NewEnemy` calls (idle loop, zone-switch, reconnect) default to `false` and keep the existing hand-back behavior — this closes the "retreat" gap specifically, not the general hand-back invariant `#1595` relies on.
- **Frontend**: `enemy-manager.ts`'s `getNewEnemy`/`runFetchLoop` thread an optional `forceAbandon` flag into the `NewEnemy` socket call; `retreatFromBoss()` passes `true`.

Nothing was cleared or persisted for a still-in-progress abandon either way, so proceeding to overwrite it in `StartBattle` is safe — same reasoning `StartBossBattle`'s existing comment already documents.

## Note on scope

This closes the gap only for `retreatFromBoss`. It intentionally does **not** touch the general `NewEnemy` zone-switch-mid-battle hand-back (which still waits out an in-progress fight) — that's the `#1595` stale-battle-recovery invariant, out of scope here.

## Test plan

- [x] Backend: new `BattleServiceIntegrationTests.StartBattle_ForceAbandonAStillInProgressBossBattle_DiscardsItAndStartsAFreshIdleBattle`, mirroring the existing hand-back-pinning sibling test but asserting the force-discard instead.
- [x] Frontend: new `enemy-manager-boss.test.ts` case asserting `retreatFromBoss()` sends `forceAbandon: true`; updated exact-object `NewEnemy` call assertions across the existing suites for the new required field.
- [x] `dotnet build` + full backend test suite (`Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Application.Tests`, `Game.Api.Tests`) — all passing.
- [x] Frontend `npm run lint` (ESLint + svelte-check) and `npm run test:unit:ci` (3532 tests) — all passing.
- [x] Codegen re-run confirms no drift beyond the intended `INewEnemyRequest.forceAbandon` addition.

---
_Generated by [Claude Code](https://claude.ai/code/session_012m43zqNvRRnpmbV6Fgp2p1)_

### DIFF
--- a/Game.Api/Models/Enemies/NewEnemyRequest.cs
+++ b/Game.Api/Models/Enemies/NewEnemyRequest.cs
@@ -9,5 +9,10 @@
         /// server-prefetched next battle discarded by a zone/build change during the cooldown — reports 0,
         /// so the abandon records no outcome. Null when not reported (the server falls back to wall-clock).</summary>
         public int? ClientBattleMs { get; set; }
+
+        /// <summary>When true, a still-in-progress battle is discarded rather than handed back unchanged —
+        /// the same override <c>ChallengeBoss</c> always applies. Lets a retreat leave a still-active boss
+        /// fight immediately instead of waiting for it to reach a real conclusion (#1690).</summary>
+        public bool ForceAbandon { get; set; }
     }
 }
--- a/Game.Api/Sockets/Commands/NewEnemy.cs
+++ b/Game.Api/Sockets/Commands/NewEnemy.cs
@@ -32,7 +32,7 @@ namespace Game.Api.Sockets.Commands
 
             var player = context.Session.Player;
 
-            var result = await _battleService.StartBattle(player, state, player.CurrentZoneId, Parameters.NewZoneId, clientBattleMs: Parameters.ClientBattleMs, cancellationToken: cancellationToken);
+            var result = await _battleService.StartBattle(player, state, player.CurrentZoneId, Parameters.NewZoneId, clientBattleMs: Parameters.ClientBattleMs, forceAbandon: Parameters.ForceAbandon, cancellationToken: cancellationToken);
 
             context.Session.SavePlayerState();
 
--- a/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
@@ -1298,6 +1298,58 @@ namespace Game.Application.Tests.Services
         }
 
         [Fact]
+        public async Task StartBattle_ForceAbandonAStillInProgressBossBattle_DiscardsItAndStartsAFreshIdleBattle()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+
+            // Mirrors StartBattle_AbandoningAStillInProgressBossBattle_HandsItBackWithBossFlagSet, but this
+            // NewEnemy request sets forceAbandon (retreat, #1690): the still-in-progress boss fight must be
+            // discarded instead of handed back, and a fresh idle battle started in its place — the same
+            // override StartBossBattle already applies unconditionally for ChallengeBoss.
+            var boss = await TestDataSeeder.CreateEnemyAsync(context, "Zone Boss", isBoss: true);
+            var bossSkill = await TestDataSeeder.CreateSkillAsync(context, "SlowBossSwipe", baseDamage: 1m, cooldownMs: 100_000_000);
+            await TestDataSeeder.LinkSkillToEnemyAsync(context, boss.Id, bossSkill.Id);
+            var idleEnemy = await TestDataSeeder.CreateEnemyAsync(context, "Idle Critter");
+            var idleSkill = await TestDataSeeder.CreateSkillAsync(context, "Nibble");
+            await TestDataSeeder.LinkSkillToEnemyAsync(context, idleEnemy.Id, idleSkill.Id);
+            var zone = await TestDataSeeder.CreateZoneAsync(context, levelMin: 1, levelMax: 1, bossEnemyId: boss.Id, bossLevel: 1);
+            await TestDataSeeder.LinkEnemyToZoneAsync(context, zone.Id, idleEnemy.Id);
+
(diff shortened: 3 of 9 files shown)


ginderjeremiah (commented):
## Review: force-abandon on retreat (#1690)

**Verdict:** Approve — the `forceAbandon` opt-in is correct, minimal, and makes retreat symmetric with `ChallengeBoss` without touching the `#1595` hand-back invariant that ordinary `NewEnemy` calls rely on. (Posted as a comment since I'm the PR author; intent is approve.)

I verified the change against the code rather than the description:

- `AbandonBattle`'s still-in-progress branch returns the handoff **without** calling `state.ClearBattle()` or `SavePlayer` (`BattleService.cs:459-470`), so falling through to overwrite the in-memory state with a fresh idle spawn is genuinely safe — exactly the reasoning `StartBossBattle` already relies on for its unconditional discard. The relocated comment documents this accurately.
- The gate `handoff is not null && !forceAbandon` leaves every other path unchanged: a *concluded* battle still resolves/credits/persists (so retreating off an unclaimed win doesn't drop it), and default (`forceAbandon: false`) calls keep the `#1595` hand-back verbatim.
- Frontend flow lands where intended: `retreatFromBoss` → `returnToIdle()` + `getNewEnemy(undefined, true)`; server returns `isBossBattle: false`, so `runFetchLoop` resolves `mode` to idle. Retreat leaves immediately instead of resuming the fight.

### Test coverage

Adequate and mirrored across both sides, per the battle-consistency rule:
- Backend: `StartBattle_ForceAbandonAStillInProgressBossBattle_DiscardsItAndStartsAFreshIdleBattle` pins the discard by mirroring the sibling hand-back test and asserting the opposite outcome (fresh idle enemy, `IsBossBattle` cleared).
- Frontend: new `enemy-manager-boss.test.ts` case asserts `retreatFromBoss()` sends `forceAbandon: true`; the exact-object `NewEnemy` assertions across the existing suites were correctly updated for the new required field.

### What's good

- Scope discipline is exactly right — the note explicitly scopes this to `retreatFromBoss` and leaves the general zone-switch-mid-battle hand-back (`#1595`) untouched.
- The `forceAbandon` field is the only `NewEnemy` sender in production code (`enemy-manager.ts`), which always sets it, so codegen'ing it as a required boolean is compiler-safe rather than a footgun.
- Comments are appropriately detailed for the subtlety involved without over-explaining.

---

Issue #1739, opened 2026-07-08
Batch consecutive same-player write-behind events onto one SaveChangesAsync


ginderjeremiah:
## Context

Follow-up to #1701, split out to keep that PR reviewable. #1701 relieves the write-behind drain's throughput ceiling by applying **distinct players'** events concurrently (bounded, per-player ordering preserved via ordering lanes) within one drain pass. It deliberately does **not** touch the other half of #1701's suggested direction — collapsing several **same-player** events onto a single `SaveChangesAsync` — because that requires a materially larger, riskier change.

## Why this is separate (not just an extension)

Today each `IPlayerUpdateHandler<TEvent>.HandleAsync` owns its own `SaveChangesAsync` call, and several handlers (`ProgressUpdatedHandler`, the equip/mod handlers per `docs/backend-persistence.md` → Queue processing resilience) have their **own** unique-violation retry/conflict-resolution logic wrapped tightly around that call. Moving to one shared `SaveChangesAsync` per same-player batch means:

- A batch-level retry-on-unique-violation has to re-run *all* handlers in the batch (safe, since they're idempotent), not just the one that collided.
- Failure isolation changes: today a poison event is dead-lettered alone, leaving batch-mates untouched. With a single shared `SaveChangesAsync`, a genuine (non-transient) failure from *one* event in a batch would fail the whole batch's save. To preserve the existing "never wrongly dead-letter a good sibling" guarantee, a failed batch save would need to degrade to a per-event fallback path (retry the batch's items one at a time) rather than dead-lettering the whole batch.
- The equip/mod handlers' bespoke "vacate destination then place" / "server-side update to latest" conflict resolution would need to be reconciled with a batch-level save rather than each owning its own attempt.

None of that is undoable, but it's a distinct design/review unit from #1701's cross-player concurrency change, and touches every one of the ~12 persistence handlers rather than just the drain loop.

## Suggested direction

- Group consecutive same-player envelopes (already naturally grouped by #1701's per-player ordering lanes) into one DI scope/`GameContext`.
- Stage each handler's changes without calling `SaveChangesAsync` itself; call it once after the whole same-player batch's handlers have run.
- On a unique-violation, clear the tracker and re-run the batch (idempotent) before retrying the save once.
- On any other failure, fall back to re-applying the batch's items one at a time (preserving today's per-event dead-letter isolation) rather than dead-lettering the whole batch.
[...]

Issue #578, opened 2026-06-14
Concurrent write-behind queue drains can apply a player's events out of order


ginderjeremiah:
## Problem

`BackgroundWorker` registers its wait with `ThreadPool.RegisterWaitForSingleObject(..., executeOnlyOnce: false)` (`Game.Infrastructure/BackgroundWorker.cs:132`), and `Start()` (`BackgroundWorker.cs:71-95`) signals `_resetEvent.Set()` **without checking `IsRunning`**. Each Redis pub/sub wake calls `worker.Start()` for the `DataProviderSynchronizer` drain.

With `executeOnlyOnce: false`, the thread pool re-arms the wait independently of callback completion, so if a second wake arrives while `ProcessQueue` (`Game.DataAccess/DataProviderSynchronizer.cs:44-54`) is still draining, a **second `ProcessQueue` can run concurrently** on another thread-pool thread. Both drain the same queue via atomic `LPOP`, so two events for the *same* player can be popped and applied in parallel / out of order.

Several handlers are order-sensitive — `HandleItemEquipped` (clear-then-set a slot), `HandleItemUnequipped`, `HandleSelectedSkillsChanged` (delete-then-rebuild). An equip immediately followed by an unequip (or two slot changes) could land in the wrong final order and persist stale state to the DB.

## Impact

A rare-but-real reordered-update class for write-behind persistence. The Redis cache remains source of truth so it self-heals on the next save, but the DB can diverge until then, and a crash in that window loses the correction. The socket's per-player command serialization does **not** extend to this independent consumer.

## Suggested fix

Serialize drains. Simplest options:
- Have `BackgroundWorker.Start()` skip the `Set()` (or no-op) when already running — the drain loop already pulls the whole queue, so a skipped wake is harmless; or
- Guard `ProcessQueue` with a `SemaphoreSlim(1,1)` using `Wait(0)`/skip-if-busy.

Add an integration test that enqueues ordered same-player events and asserts final DB state under a forced-concurrent drain.

Issue #1702, opened 2026-07-07
Reclaim stranded write-behind processing items opportunistically, not only at process startup


ginderjeremiah:
## Problem

`ReclaimProcessingAsync` runs only in `DataProviderSynchronizer.StartAsync` (`Game.DataAccess/DataProviderSynchronizer.cs:90`). If `ProcessMessage` itself throws — the escape paths are the dead-letter `AddToQueueAsync` faulting (a Redis blip at exactly the wrong moment) or `AcknowledgeAsync` faulting after a durable apply (`:246-314`) — the reserved item sits on `PlayerUpdateQueue:processing` and is never retried when Redis recovers; it waits for *some* instance to restart. In a stable single-instance deployment that delay is unbounded.

Nothing is lost and re-apply is idempotent, but a genuine player write can lag invisibly for days.

## Suggested hardening

Reclaim opportunistically when a drain pass finds the main queue empty but the processing list non-empty past a grace period (safe under the documented idempotent double-apply contract — the same reasoning that makes the startup reclaim safe). Keep the startup reclaim as-is.

_Found during a codebase health review._

PR #1740, opened 2026-07-08
Apply write-behind drain events for distinct players concurrently

### LINKED ISSUE #1701: Write-behind drain applies one event per DB round-trip, serialized per instance — a scalability ceiling on DB convergence
## Problem

Each drained write-behind event gets its own DI scope, its own handler queries, and its own `SaveChangesAsync` (`Game.DataAccess/DataProviderSynchronizer.cs:188-219`, `:320-325`; e.g. `ProgressUpdatedHandler` runs 3 SELECTs + a save per event), and the `_drainGate` serializes the whole drain within an instance (deliberately, for same-player ordering).

Every battle completion of every connected player enqueues 1-2 events, so single-instance convergence throughput is bounded at roughly `1 / (DB RTT × queries-per-event)` events/sec. Under load the queue backlog (and DB lag behind Redis) grows, eroding the documented "TTL ≫ max queue-drain time" invariant that eviction safety rests on. Fleet-wide it partially self-scales (all instances race the shared queue via `LMOVE`), which is why this is a ceiling rather than a present defect.

## Suggested direction

The ordering constraint is only **per player** (cross-instance ordering is already best-effort per `docs/backend-persistence.md`), so batching preserves the contract while removing most round-trips:

- group consecutive events for the same player into one scope/`SaveChanges`, and/or
- bounded parallelism across **distinct** players within a drain pass.

Re-apply idempotency (already pinned by `PlayerUpdateHandlerIdempotencyIntegrationTests`) keeps this safe. Benchmark before/after with a realistic event mix; no user-facing behavior change.

ginderjeremiah:
## Summary

Closes #1701.

Each drained write-behind event today gets its own DI scope, its own queries, and its own `SaveChangesAsync`, and the whole drain is serialized to one item at a time — bounding single-instance convergence throughput at roughly `1 / (DB RTT × queries-per-event)` events/sec.

This PR relieves that ceiling by letting **distinct players'** events apply concurrently within one drain pass:

- Reservation (`ReserveNextAsync`) stays strictly sequential — one atomic move off the queue at a time, in order.
- Applying an already-reserved item (its handler queries, `SaveChangesAsync`, acknowledge) no longer blocks the next reservation.
- Each reserved item is routed onto a per-player ordering lane, keyed by a player id read generically off its JSON payload — reusing (via a new shared `PlayerUpdateEnvelopeReader`) the exact same-style peek the dead-letter inspector already used to classify entries, so no envelope schema change was needed.
- An item only starts once the previous item on its **own player's** lane has fully applied and acknowledged, so same-player ordering (e.g. equip-then-unequip) is exactly as strict as before.
- Concurrency is bounded by a configurable cap (default 4, via a new optional `DataProviderSynchronizer` constructor parameter, mirroring the existing `drainTimeout` pattern).
- The concurrency slot is acquired **before** reserving, not just before applying, so a backlog dominated by one hot player can't let the drain reserve unboundedly far ahead of what has actually converged — a stop still ends at a bounded number of in-flight items, and `StopAsync`'s "wait for a clean boundary" guarantee is preserved (an in-flight item is always awaited before the drain gate is released, even when a stop cancels the reserve loop mid-wait).

## Scope note (see the issue and the follow-up)

The issue's suggested direction also floated batching **consecutive same-player** events onto one shared `SaveChangesAsync`. I deliberately left that out of this PR — it requires restructuring per-event failure isolation across all ~12 persistence handlers (several have their own bespoke unique-violation/conflict-resolution logic wrapped tightly around their own `SaveChangesAsync` call), which is a distinct, larger design/review unit. Filed as #1739.

## Changes

- `Game.DataAccess/PlayerUpdates/PlayerUpdateEnvelopeReader.cs` (new): shared best-effort `playerId` peek off a raw queued envelope/payload.
- `Game.DataAccess/Repositories/Admin/PlayerUpdateDeadLetters.cs`: refactored to reuse the shared reader instead of its own near-identical private method (no behavior change).
- `Game.DataAccess/DataProviderSynchronizer.cs`: `DrainQueueAsync` reworked for bounded, per-player-ordered concurrency; new `maxConcurrentDrainItems` constructor parameter.
- `docs/backend-persistence.md`: documents the new concurrency model.
- Tests: three new cases (cross-player concurrency actually overlaps; same-player ordering holds even with another player's event interleaved; reservation is bounded — not unbounded — ahead of a stuck item), plus one existing shutdown-boundary test updated to pin `maxConcurrentDrainItems: 1` (its assertion is specifically about single-item boundary granularity, which this PR intentionally relaxes for concurrency > 1).

## Test plan

- [x] `dotnet test` — full solution suite passes (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests): 3016/3016 passing, including the existing `#578` concurrent-drain-ordering regression test and the new concurrency tests.

Closes #1701

---
_Generated by [Claude Code](https://claude.ai/code/session_01WfPEee9oG34yF6LLEMsHKR)_


Game.DataAccess/DataProviderSynchronizer.cs
```diff
@@ -192,23 +210,64 @@ private async Task DrainQueueAsync(IPubSubQueue queue, CancellationToken cancell
...
+            // there. Same-player items still apply strictly in order: each is chained onto a per-player "lane"
+            // (playerLanes) that only starts an item once the previous same-player item has fully applied and
+            // acknowledged. The concurrency gate is acquired before reserving, not just before processing, so
+            // reservation itself is bounded by the same budget — a stop still ends at a bounded (not unbounded)
+            // number of in-flight items, reclaimed on the next startup if the drain timeout is exceeded. Once an
+            // item is reserved its apply and acknowledge run without the token so the in-flight write finishes
+            // cleanly — only the dead-time retry backoff between failed attempts honors the token (a stop during
+            // it abandons the retry, and the reserved item is reclaimed and re-applied on the next startup).
+            var playerLanes = new Dictionary<int, Task>();
+            var inFlight = new List<Task>();
```
ginderjeremiah, line 253: Should-fix (non-blocking): `inFlight` and `playerLanes` both grow to O(pass size) and are never pruned — every started item's `Task` (plus its async state machine) is retained until the whole pass ends, even after it has completed. In steady state a pass drains only a handful of events so this is nothing, but the ceiling this PR targets is precisely the *large-backlog* case (a post-outage reclaim of a big queue), where retention flips from O(1) to O(n) for the pass. It's transient and GC-reclaimed at pass end, so it doesn't threaten correctness, but given the issue's motivation it's worth bounding: periodically drop completed tasks from `inFlight`, and evict a `playerLanes` entry once its task has completed and no successor was chained. Fine to defer to a follow-up if you'd rather keep this PR tight — just calling it out so it isn't lost.

---

Game.DataAccess/DataProviderSynchronizer.cs
```diff
@@ -192,23 +210,64 @@ private async Task DrainQueueAsync(IPubSubQueue queue, CancellationToken cancell
...
+
+                    if (next is null)
+                    {
+                        concurrencyGate.Release();
+                        break;
+                    }
 
-                await ProcessMessage(next, deadLetterQueue, cancellationToken);
-                await queue.AcknowledgeAsync(next);
+                    var playerId = PlayerUpdateEnvelopeReader.TryReadPlayerId(next) ?? UnknownPlayerLane;
```
ginderjeremiah, line 316: Nit: each item's envelope JSON is now parsed twice — once here (`TryReadPlayerId` → deserialize envelope + `JsonDocument.Parse` payload) for lane routing, and again inside `ProcessMessage`'s `message.Deserialize<DomainEventEnvelope>()`. The DB round-trip dwarfs a JSON parse so this is negligible, and it's a touch ironic in a throughput PR. Not worth restructuring `ProcessMessage`'s signature to thread the already-parsed envelope through — just noting it.

---

ginderjeremiah (commented):
## Review: bounded per-player concurrent apply in the write-behind drain (#1701)

**Verdict:** Approve *(posted as a comment — GitHub won't accept a gating verdict on my own PR; there are no blocking findings)*.

This is a clean, well-scoped take on #1701. It relieves the one-round-trip-per-event ceiling by parallelizing the *apply* phase across distinct players while keeping reservation strictly sequential, and it preserves same-player ordering via per-player lanes rather than weakening the contract. The reasoning is sound throughout:

- **Ordering is genuinely preserved.** Each same-player item chains onto the previous lane task and only starts after it has fully applied *and* acknowledged, so equip→unequip stays strict. The bounded gate is acquired *before* reserving, so reservation can't run unboundedly ahead of what's converged — the `StopAsync` "clean boundary" guarantee survives because the `finally` awaits every started item (including one still mid-apply when a stop unwinds the reserve loop). Deadlock-free too: a lane's head item never awaits anything, so a hot player's chain always makes progress and releases its slots.
- **Slot accounting is correct** on every path — the `catch`/`Release` around `ReserveNextAsync`, the null/break release, and the per-item `finally` release all line up, and the `using` semaphore is disposed only after `Task.WhenAll` has awaited every releaser.
- **Lane-key safety:** routing malformed/unidentifiable items onto a single serial `UnknownPlayerLane` (`int.MinValue`, which no real player id can be) is a nice touch — it can't reorder against a genuine player.

The `PlayerUpdateEnvelopeReader` extraction (deduping the dead-letter inspector's near-identical peek) is the right call, lands in the correct layer (`Game.DataAccess`), and is documented as best-effort with the authoritative parse left to each caller. Docs (`backend-persistence.md`) are updated accurately, and the same-player-batch `SaveChangesAsync` half is correctly split out to #1739 with a clear rationale rather than being crammed in here.

### Should-fix

- `DataProviderSynchronizer.cs:227` — `inFlight`/`playerLanes` retention is O(pass size) with no pruning. Non-blocking and GC-reclaimed at pass end, but it's exactly the large-backlog case (#1701's motivation) where it matters. Bound it, or defer to a follow-up. See inline.

### Nits

- `DataProviderSynchronizer.cs:253` — envelope JSON is parsed twice per item (lane routing + `ProcessMessage`). Negligible next to the DB round-trip; not worth reworking `ProcessMessage`'s signature over. See inline.

### Test coverage

Solid and targeted at the new behavior: cross-player apply actually overlaps (`MaxObservedConcurrency`), same-player ordering holds with an interleaved other player, and reservation is bounded (not unbounded) ahead of a gated item. The shutdown-boundary test is correctly pinned to `maxConcurrentDrainItems: 1` since its assertion is about single-item granularity that this PR intentionally relaxes. Backend-only persistence infra — no battle-logic mirror needed on the frontend. No coverage gaps worth blocking on.

### What's good

Genuinely careful concurrency work: the bounded-reservation reasoning, the stop-at-a-clean-boundary preservation, and the deadlock-freedom of the lane chaining are all correct and, importantly, *explained* in the comments and doc so a later edit is less likely to naively break them.

I'll hold off on merging: the should-fix is worth a conscious decision (fix here vs. explicit follow-up), and CI is still running.

---

Game.DataAccess/DataProviderSynchronizer.cs
ginderjeremiah: Fixed in cae6298: periodic `CompactCompletedItems` sweep (triggered once `inFlight` crosses `max(concurrency * 8, 32)`) drops finished tasks/lane entries, keeping retention close to the concurrency budget for a large-backlog pass instead of O(pass size). Added a test (`ProcessQueue_LargeBacklogAcrossManyPlayers_CompactsInFlightTrackingWithoutLosingOrdering`) with 40 events across 20 players — all equips enqueued before all unequips, so by the time the later unequips are processed many earlier players' lane entries have already been swept — verifying ordering survives compaction.

---

ginderjeremiah:
CI's `test-backend` failure (run 28926367226) was a genuine race, not incidental flakiness: `StopAsync_DrainInFlight_StopsAtCleanBoundaryAndAwaitsIt` intermittently saw `ReservedCount == 2` instead of `1`.

Root cause: `SemaphoreSlim.WaitAsync(CancellationToken)` can win a race against its own token — if a slot is released at roughly the same moment the token is cancelled, the wait can complete by acquiring the slot instead of throwing `OperationCanceledException`. Under CI's heavier, resource-constrained load (this repo's own `docs/backend-persistence.md` already notes CI runs all test assemblies concurrently under the coverage profiler on a 2-core runner) this race window widens enough to bite; it never reproduced locally in isolation.

I reproduced it reliably locally by pinning the whole-solution test run to 2 CPUs (`taskset -c 0,1 dotnet test Game.sln`) — failed in 2 of 3 runs. Fixed in b919d11 by re-checking `cancellationToken.IsCancellationRequested` immediately after `WaitAsync` succeeds, releasing and breaking out if the race was lost — closing it deterministically rather than leaving it best-effort. Verified with 10 consecutive clean runs under the same 2-CPU constraint, plus a full unconstrained solution run (3017/3017 passing).

---

ginderjeremiah:
## Review: bounded per-player-lane drain concurrency

**Verdict: comment (approve-leaning) — one low-severity should-fix, otherwise solid.** CI is green (backend build/test, verify-backend, e2e chromium/firefox/webkit all pass; frontend jobs correctly skipped; `all-checks-passed` success).

I verified the concurrency design against the diff rather than trusting the description, and the core claims hold up:

- **Per-player lane keying is correct.** Each reserved item chains onto `playerLanes[playerId]` via a captured `previous` task and then replaces the lane entry, forming a proper linear per-player chain. `playerId` falls back to `UnknownPlayerLane = int.MinValue`, which a real (positive) player id can never collide with, so malformed items serialize safely on one shared lane.
- **Ordering guarantee holds.** Reservation is strictly sequential and in-order; same-player items `await previous` before applying, so equip→unequip stays ordered while distinct players overlap. `playerLanes`/`inFlight` are mutated only on the single reserve-loop thread (item tasks never touch them), so the lock-free dictionary use is safe.
- **Bounded reservation is correct.** The semaphore is acquired *before* `ReserveNextAsync` and released exactly once per acquire on every path (cancel re-check, reserve-throw, null, and the task's `finally`) — no double-release, no leak. Because a chained item holds its slot while awaiting `previous`, a hot single player genuinely bounds reserve-ahead to the budget (the `ReservesAtMostMaxConcurrentItemsAheadOfAGatedItem` test pins this). No deadlock: the globally-oldest in-flight item always bottoms out on `Task.CompletedTask`, so there is always a runnable item.
- **StopAsync clean boundary preserved.** The `finally { await Task.WhenAll(inFlight) }` runs inside the `_drainGate`-held region, so the gate isn't released (and `StopAsync` can't return) until every started item finishes — including one still applying when a stop cancels the reserve loop mid-slot-wait. The `WaitAsync`-vs-token race re-check is a correct, deliberate touch.

**Should-fix (low severity): `CompactCompletedItems` sweeps faulted/canceled tasks, not just successful ones.** `inFlight.RemoveAll(task => task.IsCompleted)` (and the lane sweep) treat `Faulted`/`Canceled` as done. If an item task faults (e.g. a Redis `AcknowledgeAsync`/dead-letter-enqueue throw, or an OCE from retry backoff on stop) and then compaction fires before the final `WhenAll`, that faulted task is dropped from `inFlight` and its exception is never observed — an unobserved-task-exception that in the serial code would have surfaced and aborted the drain. Secondarily, evicting a faulted lane entry lets a later same-player item chain onto `Task.CompletedTask` instead of rethrowing, which can micro-reorder in that same narrow window. It's genuinely low-probability (needs a handler/Redis fault *and* a 32+ item backlog in one pass) and data-safe (unacked items are reclaimed on next startup, handlers are idempotent), but the fix is a one-liner: use `IsCompletedSuccessfully` as the predicate for both collections. That keeps faulted tasks in `inFlight` so `WhenAll` observes them and preserves the chain's abort semantics. Worth doing in this PR rather than deferring.

**Nits (non-blocking):**
- Each item is JSON-deserialized twice — once in `PlayerUpdateEnvelopeReader.TryReadPlayerId` for routing, once again in `ProcessMessage`. Acceptable cost for keeping routing decoupled from the handlers; just noting it.
- `maxConcurrentDrainItems` is a constructor default (4), not wired to config — consistent with the existing `drainTimeout` pattern, so fine.
- PR body says "three new cases" but the diff adds four (`DifferentPlayers`, `SamePlayerInterleaved`, `StopDuringDrain`, `LargeBacklog`). Doc-only.

**On the #1739 deferral:** reasonable. Batching consecutive same-player events onto one `SaveChangesAsync` really does require unwinding the per-event unique-violation/conflict handling wrapped around each of the ~12 handlers' own saves — a distinct, larger, higher-risk unit. Splitting it out is the right call; this PR delivers the cross-player parallelism cleanly on its own.

The shared `PlayerUpdateEnvelopeReader` refactor (dead-letter inspector reusing it, no behavior change) is clean and the `docs/backend-persistence.md` update accurately describes the new model. No new follow-up issues filed — the one substantive finding above should be fixed in-PR, not deferred.

---

ginderjeremiah:
## Automated review — concurrent write-behind drain (#1701)

**Verdict: Request changes** (integration blocker) — the concurrency design itself is correct and well-tested; the blocker is a stale base that drops a shipped resilience feature.
**CI: green** (`all-checks-passed` ✓, backend+e2e ✓, frontend skipped — backend-only).

### Scope
`DataProviderSynchronizer.DrainQueueAsync` reworked for bounded, per-player-ordered concurrency; new shared `PlayerUpdateEnvelopeReader`; dead-letter inspector refactored to reuse it; docs updated; 4 new tests + 1 pinned to `maxConcurrentDrainItems: 1`.

### Blocking

**1. Branch predates #1702 — the rewrite drops the opportunistic stranded-processing reclaim.**
PR base `d13006e` is an ancestor of current `main`, and `main` has since gained #1702's opportunistic reclaim (`TryReclaimStrandedProcessingAsync`, `_reclaimGracePeriod`, `_timeProvider`, and two extra constructor params). This PR's head (`DataProviderSynchronizer.cs`) has none of them, and the new reserve loop's empty-queue branch is a bare `if (next is null) { concurrencyGate.Release(); break; }` — the pre-#1702 shape. Rebasing onto `main` will conflict in both `DrainQueueAsync` and the constructor, and the reclaim must be re-integrated *into the new concurrent loop* (its `next is null` → reclaim → `continue` path has to re-enter the concurrent reserve loop correctly). This interaction is nontrivial and currently unreviewed.

It's not cosmetic: this PR **introduces a new way to strand a reserved item** — a same-player successor whose predecessor faults on `AcknowledgeAsync` (Redis blip) will observe that fault through `await previous` in `ProcessReservedItemAsync`, skip its own apply, and sit reserved-but-unacked. Without #1702's opportunistic reclaim that item waits for a full process restart to recover instead of being swept within the grace period. Please rebase onto `main`, re-integrate the reclaim, and add a test for the predecessor-fault lane path.

### Non-blocking / nits

**2. Exception masking in the drain `finally`.** If the reserve loop throws a non-`OperationCanceledException` (e.g. `ReserveNextAsync` faults) *and* `Task.WhenAll(inFlight)` in the `finally` also throws, the original exception is superseded by the finally's. Low severity (rare, and both surface via logging in practice), but worth a defensive await.

**3. Double envelope deserialize per item.** `PlayerUpdateEnvelopeReader.TryReadPlayerId(next)` parses the envelope, then `ProcessMessage` parses it again. Fine given the goal is DB-bound throughput, but if profiling ever matters, thread the parsed envelope through.

**4. Test gap.** No case covers the same-player-successor-after-predecessor-acknowledge-fault path (nor a StopAsync-with-concurrency>1 awaiting multiple in-flight applies). Adds value alongside the #1702 re-integration above.

### What's correct (verified)
- **Same-player ordering holds.** Each item chains onto `playerLanes[playerId]`; the lane always holds the most-recent task, and "latest task completed ⇒ all earlier same-player tasks completed", which is what makes `CompactCompletedItems` safe to evict a completed lane without losing a later item's chain. Equip→unequip stays strict.
- **No unbounded growth / leak.** `playerLanes`/`inFlight` are per-pass locals, compacted past `max(cap*8, 32)`, and fully drained by the `finally` `WhenAll`.
- **Slot-before-reserve is correct and balanced.** Every acquire has a matching release on all paths — the post-`WaitAsync` cancellation re-check, the `ReserveNextAsync` catch, the null-break, and the item's `finally`. This genuinely bounds reserve-ahead to the cap (the gated-item test pins it at 2).
- **StopAsync clean boundary preserved.** The `finally` awaits every started item before the drain gate is released, so no apply is cut off mid-`SaveChanges`.
- **No deadlock.** Lane chains are linear DAGs rooted at `Task.CompletedTask`; slots are released without re-acquisition, so the root always makes progress.
- **Unknown-player lane** (`int.MinValue`) safely serializes malformed items; a real id can't collide.
- Reusing the dead-letter peek via `PlayerUpdateEnvelopeReader` is a clean DRY consolidation with no behavior change.

Concurrency logic: sound. Please rebase to reconcile with #1702 before merge.

_— Automated review (Claude opus-4-8). Not merged._

---

ginderjeremiah:
## Review — concurrent per-player-lane write-behind drain (#1701)

**Verdict: Request changes.** The concurrency design is correct and well-tested, but the PR is built on a stale base that no longer merges and would regress a shipped resilience feature. **Not merged.**
**CI: green** on the head commit (`all-checks-passed` ✓, `test-backend` / `verify-backend` / e2e chromium+firefox+webkit ✓; frontend jobs correctly skipped) — but that run was against the stale base, so it does not reflect the post-rebase reality.

### Blocking

**1. Stale base — merge conflict + drops #1702's opportunistic reclaim.** GitHub reports `mergeable_state: dirty`, and I confirmed why via git ancestry: this PR's base `d13006e` predates #1702 (`7dff6b8`, merged onto `main` ~10h after the base). Current `main` carries `TryReclaimStrandedProcessingAsync`, `_reclaimGracePeriod`, `_timeProvider`, `GetProcessingCountAsync`, and two extra constructor params — none of which exist on this branch. The reworked `DrainQueueAsync` reverts the empty-queue branch to the pre-#1702 shape (`if (next is null) { concurrencyGate.Release(); break; }`), with no reclaim. A rebase onto `main` will conflict in both `DrainQueueAsync` and the constructor, and the reclaim must be **re-integrated into the new concurrent reserve loop** (its `next is null → reclaim → continue` must re-enter the concurrent loop, releasing/re-acquiring the slot correctly). That interaction is nontrivial and currently unreviewed.

This is not cosmetic, because this PR **introduces a new stranding vector** that #1702 is exactly what recovers from: when a predecessor item's `AcknowledgeAsync` faults (a Redis blip), its same-player successor's `await previous` in `ProcessReservedItemAsync` rethrows, so the successor skips its own apply and sits reserved-but-unacked. On `main` #1702 sweeps that within the grace period; on this branch it waits for a full process restart. Rebase, re-integrate the reclaim, and add a test for the predecessor-fault lane path.

### Should-fix (in this PR, low severity)

**2. `CompactCompletedItems` evicts faulted/canceled tasks, not just successful ones.** `inFlight.RemoveAll(task => task.IsCompleted)` and the lane sweep treat `Faulted`/`Canceled` as done. If an item task faults (a Redis `AcknowledgeAsync` throw, an OCE from retry backoff on stop, or a propagated `await previous` fault) and compaction fires before the final `WhenAll`, that task is dropped from `inFlight` and its exception is never observed — an unobserved-task-exception that the serial code would have surfaced up through `ProcessQueue`. Secondarily, evicting a faulted lane lets a later same-player item chain onto `Task.CompletedTask` instead of the faulted predecessor. Low probability (needs a fault *and* a 32+ item single-pass backlog), data-safe (unacked items reclaim on next startup, handlers idempotent), but the fix is a one-liner: predicate on `IsCompletedSuccessfully` for both collections. This was raised in an earlier review on this PR and is still unaddressed.

### Nits (non-blocking)

- **Exception masking in the drain `finally`.** If the reserve loop throws a non-OCE *and* `Task.WhenAll(inFlight)` in the `finally` also throws, the original is superseded. Rare, both surface via logging in practice.
- **Double envelope deserialize per item** — `TryReadPlayerId` parses, then `ProcessMessage` parses again. Acceptable given the goal is DB-bound throughput; thread the parsed envelope through only if profiling ever shows it.
- PR body says "three new tests" but the diff adds four.

### Verified correct (independent read of the diff)
- **Same-player ordering** — each reserved item chains onto `playerLanes[playerId]` via a captured `previous` and replaces the lane entry; strictly linear per player, so equip→unequip stays ordered while distinct players overlap.
- **Lock-free collection use is safe** — `playerLanes`/`inFlight` are per-pass locals mutated only on the single reserve-loop thread; item tasks never touch them.
- **Bounded reservation** — the slot is acquired *before* `ReserveNextAsync` and released exactly once on every path (cancel re-check, reserve-throw, null-break, item `finally`); a chained item holds its slot while awaiting `previous`, so a hot single player genuinely bounds reserve-ahead to the budget.
- **No deadlock** — lane chains are linear, rooted at `Task.CompletedTask`; slots release without re-acquisition, so the oldest in-flight item always makes progress.
- **StopAsync clean boundary preserved** — `finally { await Task.WhenAll(inFlight) }` runs inside the `_drainGate`-held region, so the gate isn't released (and `StopAsync` can't return) until every started item finishes, including one still applying when a stop cancels the reserve loop mid-slot-wait. The `WaitAsync`-vs-token re-check is a correct, deliberate touch.
- **`UnknownPlayerLane = int.MinValue`** safely serializes malformed items on one lane; a real player id can never collide.
- The `PlayerUpdateEnvelopeReader` extraction (dead-letter inspector reusing it) is a clean, behavior-preserving DRY consolidation, and the `docs/backend-persistence.md` update accurately describes the model.

### Test coverage
The three headline tests do prove what they claim — `DifferentPlayersEvents_ApplyConcurrently` asserts real overlap (`MaxObservedConcurrency == 2`), `SamePlayerEventsInterleavedWithOtherPlayers` pins ordering under interleave, and `StopDuringDrain_ReservesAtMostMaxConcurrentItemsAheadOfAGatedItem` deterministically pins bounded reserve-ahead; `LargeBacklog…Compacts…` exercises compaction. **Gaps:** no test for the predecessor-fault lane path (item 1/2 above), nor a `StopAsync` with concurrency > 1 awaiting multiple in-flight applies. Add these alongside the #1702 re-integration.

### Follow-ups
None filed — the substantive findings (rebase + reclaim re-integration, the `IsCompletedSuccessfully` fix, the missing tests) all belong **in this PR**, not deferred. Same-player `SaveChangesAsync` batching is already tracked as #1739.

Concurrency logic: sound. Please rebase onto `main` to reconcile with #1702, fix the compaction predicate, and add the fault-path test before merge.

_— Automated review (Claude opus-4-8). Not merged; base does not merge cleanly._

---

ginderjeremiah:
## Review — concurrent per-player-lane write-behind drain (#1701)

**Verdict: request changes.** The concurrency design is correct and genuinely well-tested; the blocker is a stale base that drops a shipped resilience feature and won't merge cleanly. **CI: green** on the head commit (`all-checks-passed`, `test-backend`/`verify-backend`, e2e chromium+firefox+webkit; frontend jobs correctly skipped) — but that run predates the divergence below, so it doesn't reflect post-rebase reality.

I verified the design against the diff independently rather than trusting the description.

### Blocking

**1. Stale base drops #1702's opportunistic reclaim.** Confirmed via git ancestry: base `d13006e` is an ancestor of current `main` but does **not** contain #1702 (`7dff6b8`, merged ~10h after the base). `main` carries `TryReclaimStrandedProcessingAsync`, `_reclaimGracePeriod`, `_timeProvider`, `GetProcessingCountAsync`, and two extra ctor params — none on this branch. The reworked `DrainQueueAsync` reverts the empty-queue branch to the pre-#1702 shape (`if (next is null) { concurrencyGate.Release(); break; }`), with no reclaim. A rebase will conflict in both `DrainQueueAsync` and the constructor, and the reclaim must be re-integrated **into the new concurrent loop** — its `next is null → reclaim → continue` path has to re-enter the reserve loop releasing/re-acquiring the slot correctly. That interaction is nontrivial and currently unreviewed.

Not cosmetic: this PR **introduces a new stranding vector** that #1702 is precisely what recovers. When a predecessor item's `AcknowledgeAsync` faults (Redis blip), its same-player successor's `await previous` in `ProcessReservedItemAsync` rethrows, so the successor skips its own apply and sits reserved-but-unacked. On `main`, #1702 sweeps that within the grace period; on this branch it waits for a full process restart. Rebase, re-integrate the reclaim, and add a test for the predecessor-fault lane path.

### Should-fix (in this PR, low severity)

**2. `CompactCompletedItems` evicts faulted/canceled tasks, not just successful ones.** `inFlight.RemoveAll(task => task.IsCompleted)` and the lane sweep treat `Faulted`/`Canceled` as done. If an item task faults (a Redis `AcknowledgeAsync`/dead-letter-enqueue throw, or a propagated `await previous` fault) and compaction fires before the final `WhenAll`, that task is dropped from `inFlight` and its exception is never observed — whereas in a sub-threshold pass the same fault *does* surface up through `ProcessQueue` (inconsistent behavior depending on backlog size). Secondarily, evicting a faulted lane lets a later same-player item chain onto `Task.CompletedTask` instead of rethrowing. Low probability (needs a fault *and* a 32+ item single-pass backlog), data-safe (unacked items reclaim on restart, handlers idempotent), but the fix is a one-liner: predicate on `IsCompletedSuccessfully` for both collections.

### Nits (non-blocking)

- **Exception masking in the drain `finally`:** if the reserve loop throws a non-OCE *and* `Task.WhenAll(inFlight)` in the `finally` also throws, the original is superseded. Rare; both log in practice.
- **Double envelope deserialize per item** — `TryReadPlayerId` parses, then `ProcessMessage` parses again. Acceptable given the goal is DB-bound throughput; thread the parsed envelope through only if profiling shows it.
- PR "Changes" says "three new cases"; the diff adds four (`DifferentPlayers`, `SamePlayerInterleaved`, `StopDuringDrain`, `LargeBacklog`).

### Verified correct (independent read)

- **playerId routing is robust.** Every persisted event carries a `playerId` — including `ProgressUpdatedEvent` (`PlayerId`), so progress events (half the per-battle volume) route to their player's lane rather than piling onto `UnknownPlayerLane`. Malformed/unknown → `int.MinValue`, which a real (positive) id can't collide with, so poison items serialize safely on one shared lane. Reuses the dead-letter inspector's exact `playerId` peek via `PlayerUpdateEnvelopeReader` — a clean, behavior-preserving DRY consolidation.
- **Same-player ordering holds.** Each reserved item chains onto `playerLanes[playerId]` via a captured `previous` and replaces the lane entry — strictly linear per player; equip→unequip stays ordered while distinct players overlap. `playerLanes`/`inFlight` are per-pass locals mutated only on the single reserve-loop thread (item tasks never touch them), so the lock-free dictionary use is safe. "Latest lane task completed ⇒ all earlier same-player tasks completed" is what makes compaction safe to evict a completed lane without losing a later item's chain.
- **Bounded reservation.** Slot acquired *before* `ReserveNextAsync` and released exactly once on every path (post-`WaitAsync` cancel re-check, reserve-throw catch, null-break, item `finally`) — no double-release, no leak. A chained item holds its slot while awaiting `previous`, so a hot single player genuinely bounds reserve-ahead to the budget (`ReservesAtMostMaxConcurrentItemsAheadOfAGatedItem` pins this at 2).
- **No deadlock.** Lane chains are linear DAGs rooted at `Task.CompletedTask`; slots release without re-acquisition, so the oldest in-flight item always makes progress.
- **StopAsync clean boundary preserved.** `finally { await Task.WhenAll(inFlight) }` runs inside the `_drainGate`-held region, so the gate isn't released (and `StopAsync` can't return) until every started item finishes — including one still applying when a stop cancels the reserve loop mid-slot-wait. The `WaitAsync`-vs-token re-check is a correct, deliberate touch.

### Test coverage
Headline tests prove what they claim (real overlap via `MaxObservedConcurrency == 2`, ordering under interleave, deterministic bounded reserve-ahead, mid-pass compaction). **Gaps:** no predecessor-fault lane path (item 1/2), nor a `StopAsync` with concurrency > 1 awaiting multiple in-flight applies. Add these alongside the reclaim re-integration.

### On the #1739 deferral
Reasonable — batching consecutive same-player events onto one `SaveChangesAsync` needs the per-handler unique-violation/conflict logic unwound across ~12 handlers, a distinct larger unit. This PR delivers the cross-player parallelism cleanly on its own.

Concurrency logic: sound. Please rebase onto `main` to reconcile with #1702, fix the compaction predicate, and add the fault-path test before merge. Not merged.

---
_Automated review (Claude opus-4-8)._

---

ginderjeremiah:
## Review

Bounded, per-player-ordered concurrency for the write-behind drain. I reviewed the concurrency design closely and verified claims against the code.

### Concurrency correctness (the core of this PR) — looks solid

- **Same-player ordering** is preserved: each reserved item chains onto `playerLanes[playerId]` and awaits that predecessor (through its `AcknowledgeAsync`) before calling `ProcessMessage`, so equip-then-unequip on one lane can never reorder. Malformed/unknown-player items collapse onto the shared `UnknownPlayerLane` (`int.MinValue`), which a real player id can never collide with — order-safe.
- **Reservation stays strictly sequential** (one `ReserveNextAsync` at a time in the single reserve loop), and the concurrency slot is acquired *before* reserving, so the drain can't run unboundedly ahead of what has converged. Bound holds.
- **Slot accounting is exception-safe:** every acquired slot is released on exactly one path — the cancel re-check, the `ReserveNextAsync` catch, the `next is null` break, and `ProcessReservedItemAsync`'s `finally`. No leaks on any path I traced.
- **No deadlock:** a chained item waits only on a predecessor that already holds its own slot and is progressing, so waiters never starve the budget.
- **StopAsync clean boundary preserved:** the `finally` awaits `Task.WhenAll(inFlight)`, so an in-flight apply is always completed before the drain gate is released, even when the stop cancels the reserve loop mid-wait. The `WaitAsync`-vs-token race re-check is a nice touch.
- **Lane dictionary is single-threaded** (only the reserve loop and `CompactCompletedItems` touch it) — no races. Compaction only evicts a lane whose *current* task is complete, so a later same-player item still chains correctly rather than onto a stale entry.

The shared `PlayerUpdateEnvelopeReader` reuse (dead-letter classifier + drain) is clean DRY, no null-forgiving, braces everywhere, old private method removed. The three new concurrency tests plus the large-backlog compaction test are meaningful and deterministic (the acknowledge-delay overlap and the gated-reserve-count tests avoid timing guesses). The `maxConcurrentDrainItems: 1` pin on the existing shutdown-boundary test is correctly justified.

### Blocking

1. **The branch is stale and will not merge/compile as-is.** It branched from `d13006e`, but `#1702`/`#1745` (opportunistic reclaim of stranded processing items) has since landed on `main` and rewrites the *same* `DrainQueueAsync` loop and constructor signature. Concretely:
   - `git merge-tree` reports a **content conflict** in `DataProviderSynchronizer.cs` (both rewrote the reserve loop; both added a 6th constructor parameter — `maxConcurrentDrainItems` here vs `reclaimGracePeriod`/`timeProvider` on main).
   - `IPubSubQueue` gained `GetProcessingCountAsync` on main; the new `ConcurrentAcknowledgeTrackingQueue` test double doesn't implement it, so the test project **won't compile** after rebase.
   - This is **not a mechanical rebase**: the reclaim path runs `TryReclaimStrandedProcessingAsync` when `ReserveNextAsync` returns null, keying off `GetProcessingCountAsync`. Under this PR's model, reserved-but-not-yet-acked in-flight items legitimately sit on the processing list, so the reclaim's notion of "stranded" now overlaps with "actively converging concurrently." The grace-period/first-sighting logic likely still covers it, but that interaction must be reasoned through and re-tested, not merged blind.

   Please rebase onto current `main` and re-review the reclaim + concurrency interaction; green CI here predates `#1745`.

### Nits (non-blocking)

- `CompactCompletedItems` sweeps *faulted* tasks out of `inFlight` too (`IsCompleted` is true for a faulted task), so if `AcknowledgeAsync` throws a non-cancellation error mid-large-pass, that exception is dropped rather than surfaced by the final `WhenAll`. No data loss (the item stays on the processing list and is reclaimed later), and `ProcessMessage` is non-throwing by design, so this is minor — worth a short comment noting the intent.
- Each reserved item is now deserialized ~twice extra (envelope + inner payload in the reader, then again in `ProcessMessage`). Negligible against the DB round-trips this PR is relieving; fine to leave.

### Scope note

Agree with deferring same-player `SaveChanges` batching to `#1739` — it needs the per-handler failure-isolation redesign and is a separate review unit. No new follow-up issue needed.

### CI

Green on the last run (`test-backend`, `verify-backend`, e2e all pass; frontend skipped) — but that run is against the pre-`#1745` base, so it does not reflect the merged state.

**Verdict: request changes** — the design is correct and I'd approve it on its own merits, but it must be rebased onto current `main` (non-mechanically) and re-verified before merge.

---

ginderjeremiah:
## Review: bounded per-player concurrent write-behind drain (#1701)

Reviewed the diff against the issue and `docs/backend-persistence.md`. The core design is sound and matches the docs: reservation stays strictly sequential, apply+ack runs with a `SemaphoreSlim`-bounded budget, per-player lanes chain each item onto its predecessor, and the slot-before-reserve ordering plus the `finally { await Task.WhenAll(inFlight) }` correctly preserve the `StopAsync` clean-boundary guarantee. The `PlayerUpdateEnvelopeReader` DRY extraction is a genuine improvement, and I confirmed `ProgressUpdatedEvent` carries a (camelCase) `playerId`, so the issue's most expensive handler (3 SELECTs + save) actually lands on a concurrent lane rather than the serial unknown-lane. CI is green (backend, e2e all pass). Code-quality-wise it's clean: always-braces, no null-forgiving, no dead code.

Two issues before merge.

### BLOCKING

**1. PR base is stale and its `DrainQueueAsync` rework drops #1702's opportunistic stranded-processing reclaim.**
This PR branched from `d13006e` (2026-07-08 01:36). #1702 — "Reclaim stranded write-behind processing items opportunistically" (`7dff6b8`, merged as #1745) — landed ~10h later (11:35) and is now on `main`. That change added a `next is null → TryReclaimStrandedProcessingAsync(...) → continue` branch (plus `_reclaimGracePeriod` / `_timeProvider` / `_strandedProcessingObservedAt`) to `DrainQueueAsync`. This PR's reworked loop replaces the whole method and its `next is null` path is just `concurrencyGate.Release(); break;` — the opportunistic reclaim is gone. A merge will either conflict on this method or, if resolved toward this PR, silently regress a merged feature. Rebase onto current `main` and re-integrate the opportunistic reclaim into the new concurrent loop (the `next is null` → reclaim → `continue` needs to keep the slot accounting correct — release-then-reclaim-then-reacquire, or reclaim before the WaitAsync).

### SHOULD-FIX

**2. `CompactCompletedItems` evicts on `task.IsCompleted`, not `IsCompletedSuccessfully` — a narrow but real same-player ordering hole.**
The compaction sweep (and `inFlight.RemoveAll(task => task.IsCompleted)`) treats a *faulted* lane task the same as a successful one. A non-cancellation fault is reachable: `ProcessReservedItemAsync` faults if `queue.AcknowledgeAsync(message)` (or a dead-letter write inside `ProcessMessage`) hits a Redis blip *after* the apply has durably landed. In that case the item is applied-but-not-acknowledged and stays on the processing list for reclaim. If the backlog is past the compaction threshold (default 32) and a *later same-player* item is reserved after that faulted lane was compacted away, the successor chains onto `Task.CompletedTask` instead of the faulted predecessor, applies immediately, and the reclaimed predecessor then re-applies *after* it — reordering e.g. equip→unequip. Without compaction this can't happen, because the successor's `await previous` would observe the fault and skip itself (staying reserved for in-order reclaim). So compaction specifically weakens the "same-player ordering is exactly as strict as before" claim. Secondary symptom: removing the faulted task from `inFlight` before `Task.WhenAll` swallows its exception. Fix: evict only `IsCompletedSuccessfully` lanes/tasks; leave faulted (and canceled) ones in place so the lane dependency and the exception are both preserved.

**3. Test gap for the above.** `ProcessQueue_LargeBacklog...CompactsInFlightTracking` exercises compaction only with successful tasks; the same-player-ordering test uses 3 items (well under threshold) and no faults. Add a case that faults a lane's ack/apply and then reserves a later same-player item past the compaction threshold.

### FOLLOW-UP / nit
- PR body says "three new cases," but four new `[Fact]`s were added (the large-backlog compaction test is the fourth). Cosmetic.

### Verdict: request-changes
Design and per-player ordering under the happy path are correct and well-documented, but (1) the PR must be rebased so it doesn't regress the now-merged #1702 reclaim, and (2) compaction should key off `IsCompletedSuccessfully` to hold the ordering guarantee the PR advertises on the fault path.

---

ginderjeremiah:
## Review: Request changes

I reviewed this carefully as the concurrency-critical PR it is. **The concurrency design is correct** — this is good work — but the branch is out of date with `main` in a way that conflicts non-trivially, so it can't merge as-is. Details below.

### Concurrency correctness (verified — all good)

- **Sequential, ordered reservation:** `ReserveNextAsync` is only ever called from the single drain loop, one await at a time. ✓
- **Same-player ordering:** each item chains onto `playerLanes[playerId]`, and `ProcessReservedItemAsync` does `await previous` (which includes the predecessor's own `await previous → ProcessMessage → AcknowledgeAsync`) before applying. So item N+1 genuinely waits for N to fully apply *and* ack. ✓
- **Slot-before-reserve bounds look-ahead:** `concurrencyGate.WaitAsync` precedes `ReserveNextAsync`, and the slot is held for the item's whole lifetime including its wait on `previous`. A hot single player therefore consumes up to `max` slots even though only one is applying — that's the intended cost that keeps reservation from running unboundedly ahead of convergence. ✓
- **No deadlock / no slot leak:** every acquisition path has exactly one matching release (cancel-recheck, reserve-throw catch, null-break, or `ProcessReservedItemAsync`'s `finally`) — no path both releases and hands the slot to the task. A lane's predecessor is always reserved earlier, so it already holds its slot or is done; the chain is strictly backward, so some head item is always progressing. ✓
- **Clean shutdown boundary:** the `finally { Task.WhenAll(inFlight) }` awaits every started item even when a stop throws `OperationCanceledException` out of `WaitAsync` mid-wait, so `StopAsync` can't release the drain gate while an apply is in flight. The `WaitAsync`-wins-the-race re-check is a nice touch. ✓
- **Generic `playerId` peek is safe:** missing/unparseable id → `UnknownPlayerLane` (`int.MinValue`, never a real id), collapsing only the rare malformed items onto one serial lane (order-safe, throughput-only cost). I confirmed both real hot-path events (`PlayerCoreUpdatedEvent`, `ProgressUpdatedEvent`) serialize `PlayerId` as camelCase `playerId`, so they route onto proper lanes and actually benefit. ✓
- **Tests** cover the four things that matter: cross-player overlap (`MaxObservedConcurrency`), same-player ordering with an interleaved other player, bounded reservation ahead of a gated item, plus compaction under a large backlog. The shutdown-boundary test is correctly pinned to `maxConcurrentDrainItems: 1`. ✓

### Blocking: branch is stale and conflicts with `main` (#1745)

This PR branched **before** #1745 (opportunistic reclaim of stranded processing items) merged, and it rewrote the exact `DrainQueueAsync` loop #1745 also touched. Merging into current `main` produces a real content conflict in `DataProviderSynchronizer.cs`.

The reconciliation is **not just textual**: #1745 added a `TryReclaimStrandedProcessingAsync` call on the `next is null` path, whose trigger is *"main queue empty but the processing list still non-empty."* With bounded concurrency, when this pass hits `next is null` there may still be items **this pass** reserved and is legitimately applying — they're on the processing list, so `GetProcessingCountAsync` counts them. The 30s grace period largely masks it, but the interaction needs a deliberate decision (e.g. only consider opportunistic reclaim after `await Task.WhenAll(inFlight)`, once nothing this pass reserved is still in flight). Please rebase onto `main` and reconcile the two features explicitly — this warrants a re-review of the merged loop.

### Should-fix during the rebase (non-blocking on its own)

**Compaction can drop faulted/cancelled task exceptions unobserved.** `CompactCompletedItems` does `inFlight.RemoveAll(t => t.IsCompleted)` — a *faulted* task is completed, so it's removed before the terminal `Task.WhenAll(inFlight)` can observe it. `ProcessReservedItemAsync` faults on a Redis blip in `AcknowledgeAsync`/the dead-letter write, or an `OperationCanceledException` from the retry backoff during shutdown. Those become unobserved task exceptions (rare, and not a data-loss bug — the items stay reserved and are reclaimed/re-applied idempotently). Since you're reworking this area for the rebase anyway, consider observing exceptions as you compact (or aggregating them) so a fault is surfaced rather than swallowed, consistent with the "surface, don't silently drop" intent.

### Minor (optional)

- No concurrency-`> 1` test for the stop-mid-`WaitAsync` path specifically; the `maxConcurrentDrainItems: 1` boundary test plus the shared code path cover it, but a `> 1` variant asserting all in-flight items are awaited would harden it.

### Scope note

Leaving the consecutive-same-player `SaveChanges` batching out (filed as #1739) is the right call — it's a separate failure-isolation redesign. Good judgment.

Net: solid, correct concurrency work. **Request changes purely to rebase onto `main` and consciously reconcile with #1745** (plus the compaction/unobserved-exception tidy-up while you're in there). Happy to re-review the reconciled loop promptly.

---

ginderjeremiah:
## Review summary — Request changes ⚠️

The concurrency design here is genuinely strong — I'd approve it on its own merits. I verified the core claims against the diff and they hold:

- **Same-player ordering is preserved.** Each reserved item chains onto `playerLanes[playerId]` via a captured `previous` task and `await previous` (which transitively includes the predecessor's `ProcessMessage` + `AcknowledgeAsync`) before applying, so equip→unequip can't reorder while distinct players overlap.
- **Reserve-ahead is genuinely bounded** — `concurrencyGate.WaitAsync` precedes `ReserveNextAsync` and the slot is held across the wait on `previous`, so a hot single player consumes slots even while only one item applies. Pinned by the new bounded-reserve test.
- **StopAsync's clean boundary holds** — `finally { await Task.WhenAll(inFlight) }` runs inside the `_drainGate`-held region, so a stop can't return until in-flight applies finish.
- The `PlayerUpdateEnvelopeReader` extraction is a clean DRY consolidation, and routing actually works (camelCase serialization + `PlayerId` on the hot-path events), so the benefit is real. Deferring same-player batching to #1739 is the right call.

However, it **cannot merge as-is** — the issues below need to be fixed **in this PR** (not deferred):

**1. (Blocking) Stale base — won't compile after rebase, and drops the merged #1745 reclaim.** Current `main` carries #1745's opportunistic stranded-processing reclaim (`TryReclaimStrandedProcessingAsync`, `_reclaimGracePeriod`, `_timeProvider`, two extra ctor params, `IPubSubQueue.GetProcessingCountAsync`). This branch's base predates it and its reworked `DrainQueueAsync` reverts the empty-queue branch to the pre-#1745 shape.
   - Constructor conflict: main's added param vs this PR's `maxConcurrentDrainItems`.
   - The test double `ConcurrentAcknowledgeTrackingQueue` doesn't implement `GetProcessingCountAsync`, so the test project won't compile after rebase.
   - **Non-mechanical reconciliation:** #1745's reclaim triggers on "main queue empty but processing list non-empty." Under bounded concurrency, when the reserve loop hits `next is null` there may still be items from *this pass* legitimately applying (and thus on the processing list, counted as "stranded"). The interaction needs to be reasoned through explicitly (e.g. only consider reclaim after `await Task.WhenAll(inFlight)`), with correct release-then-reclaim-then-reacquire slot accounting. This PR also *introduces* a new stranding vector (a same-player successor whose predecessor faults on `AcknowledgeAsync` sits reserved-but-unacked) that #1745's reclaim is exactly what recovers — so re-integrating it isn't optional.

**2. (Should-fix, in this PR) `CompactCompletedItems` evicts on `IsCompleted`, not `IsCompletedSuccessfully`.** `ProcessReservedItemAsync` can fault non-cancellation (`AddToQueueAsync`/`AcknowledgeAsync` are Redis calls — a blip throws after the apply may have durably landed). If a faulted task is compacted (backlog ≥ `max(cap*8, 32)`) before the terminal `WhenAll`: (a) its exception is never observed (inconsistent with the sub-threshold path), and (b) its lane entry is evicted, so a later same-player item chains onto `Task.CompletedTask` and applies immediately while the reclaimed predecessor re-applies afterward — reordering equip→unequip. Low-probability and data-safe, but it directly weakens the advertised "same-player ordering exactly as strict as before." One-line fix: predicate both sweeps on `IsCompletedSuccessfully` so faulted/canceled tasks stay in place, preserving both the lane dependency and the exception.

**3. (Should-fix) Test gap for the fault path.** No test faults a lane's apply/ack then reserves a later same-player item past the compaction threshold, and none exercises `StopAsync` with concurrency > 1 awaiting multiple in-flight applies. Add these alongside 2 / the reclaim re-integration.

Nits (non-blocking): a non-OCE reserve-loop throw can be masked if the `finally` `WhenAll` also throws (both log in practice); each item is deserialized twice (routing peek + `ProcessMessage`) — acceptable for a DB-bound path; PR body says "three new cases" but four were added.

Please rebase onto current `main`, re-integrate #1745's reclaim into the concurrent loop (reconciling the stranded-vs-actively-converging overlap and slot accounting), fix the compaction predicate, implement `GetProcessingCountAsync` on the test double, add the fault-path + `StopAsync`-concurrency tests, and re-run CI (the current green run is stale). Happy to re-review the reconciled loop.

_Reviewed by Claude._

---

ginderjeremiah:
## Review — bounded per-player concurrent write-behind drain (#1701)

**Verdict: request changes.** The concurrency design is correct and genuinely well-crafted — I'd approve it on its own merits — but the branch is stale in a way that both fails to merge and would regress a shipped resilience feature, and a low-severity ordering hole on the fault path remains unaddressed. **Not merged.** Every finding below was verified independently against the diff and via git; none is new relative to the prior reviews, and none has been fixed on the current head (`b919d11`).

### CI
Green on the head commit (`all-checks-passed`, `test-backend`, `verify-backend`, e2e chromium/firefox/webkit; frontend jobs correctly skipped — backend-only). **But that run predates #1745 on `main`, so it does not reflect post-rebase reality** (the test project won't even compile after rebase — see below).

### Concurrency correctness (verified — sound)
- **Same-player ordering holds.** Each reserved item chains onto `playerLanes[playerId]` via a captured `previous` and replaces the lane entry; `ProcessReservedItemAsync` does `await previous` (transitively the predecessor's `ProcessMessage` + `AcknowledgeAsync`) before applying — so equip→unequip can't reorder while distinct players overlap. Reservation stays strictly sequential (one `ReserveNextAsync` per await in the single loop).
- **Reserve-ahead is genuinely bounded.** The slot is acquired *before* `ReserveNextAsync` and held across the wait on `previous`, so a hot single player consumes up to the budget even while only one item applies. Every acquire has exactly one matching release on every path (post-`WaitAsync` cancel re-check, reserve-throw catch, null-break, item `finally`) — no leak, no double-release. The `WaitAsync`-wins-the-race re-check is a correct, deliberate touch.
- **No deadlock.** Lane chains are linear, rooted at `Task.CompletedTask`; slots release without re-acquisition, so the oldest in-flight item always progresses.
- **StopAsync clean boundary preserved.** `finally { await Task.WhenAll(inFlight) }` runs inside the `_drainGate`-held region, so the gate isn't released (and `StopAsync` can't return) until every started item finishes — including one still applying when a stop cancels the reserve loop mid-slot-wait.
- **`playerLanes`/`inFlight` are per-pass locals touched only on the single reserve-loop thread** (item tasks never mutate them) — lock-free use is safe. Item tasks share only `concurrencyGate`, the queues, and per-call DI scopes/`GameContext`s — no shared mutable state.
- **playerId routing degrades safely.** Missing/malformed id → `UnknownPlayerLane` (`int.MinValue`), which a real positive id can never collide with, collapsing only rare poison items onto one serial lane (order-safe, throughput-only cost). I confirmed `PlayerCoreUpdatedEvent` and `ProgressUpdatedEvent` both serialize camelCase `playerId`, so the hot-path events actually land on real lanes and benefit. The `PlayerUpdateEnvelopeReader` extraction is a clean, behavior-preserving DRY consolidation.

Code quality is clean: always-braces, no null-forgiving, old private method removed, docs accurately updated. Deferring same-player `SaveChanges` batching to #1739 is the right call — that's a genuine per-handler failure-isolation redesign across ~12 handlers, correctly scoped and labeled as its own unit.

### Blocking

**1. Stale base — won't merge, won't compile after rebase, and drops #1745's opportunistic reclaim.** Verified via git: base `d13006e` is an ancestor of current `main` but does **not** contain #1745 (`7dff6b8`); the PR head does not contain it either. `git merge-tree origin/main b919d11` reports content conflicts in **both** `DataProviderSynchronizer.cs` and `docs/backend-persistence.md`. Concretely:
   - `main` now carries `TryReclaimStrandedProcessingAsync`, `_reclaimGracePeriod`, `_timeProvider`, `_strandedProcessingObservedAt`, and `IPubSubQueue.GetProcessingCountAsync`. This branch's reworked `DrainQueueAsync` reverts the `next is null` path to the pre-#1745 shape (`concurrencyGate.Release(); break;`) — the reclaim is gone.
   - Constructor conflict: main's added params vs this PR's `maxConcurrentDrainItems`.
   - `IPubSubQueue.GetProcessingCountAsync` is now a required interface member; the PR's new test double `ConcurrentAcknowledgeTrackingQueue` doesn't implement it, so **the test project won't compile after rebase**.
   - **The reconciliation is non-mechanical.** #1745's reclaim triggers on "main queue empty but processing list non-empty." Under bounded concurrency, when the reserve loop hits `next is null` there may still be items *this pass* legitimately applying — they sit on the processing list and `GetProcessingCountAsync` counts them, overlapping "stranded" with "actively converging concurrently." The 30s grace period largely masks it, but the interaction needs a deliberate decision (e.g. only consider opportunistic reclaim after `await Task.WhenAll(inFlight)`, with correct release-then-reclaim-then-reacquire slot accounting).
   - This isn't optional cleanup: **this PR introduces a new stranding vector** #1745 is exactly what recovers — a same-player successor whose predecessor faults on `AcknowledgeAsync` observes that fault through `await previous`, skips its own apply, and sits reserved-but-unacked. Without the reclaim it waits for a full process restart instead of being swept within the grace period.

### Should-fix (in this PR)

**2. `CompactCompletedItems` evicts on `IsCompleted`, not `IsCompletedSuccessfully`.** `ProcessReservedItemAsync` can fault non-cancellation (`AcknowledgeAsync` / the dead-letter `AddToQueueAsync` are Redis calls — a blip throws after the apply may have durably landed). If a faulted task is compacted (backlog ≥ `max(cap*8, 32)`) before the terminal `WhenAll`: (a) its exception is never observed — inconsistent with the sub-threshold path, where the same fault *does* surface up through `ProcessQueue`; and (b) its lane entry is evicted, so a later same-player item chains onto `Task.CompletedTask` and applies immediately while the reclaimed predecessor re-applies afterward — reordering equip→unequip. Low-probability and data-safe (unacked items reclaim, handlers idempotent), but it directly weakens the advertised "same-player ordering exactly as strict as before." One-line fix: predicate both sweeps on `IsCompletedSuccessfully` so faulted/canceled tasks stay in place, preserving both the lane dependency and the exception.

**3. Test gap for the fault path.** No test faults a lane's apply/ack then reserves a later same-player item past the compaction threshold, and none exercises `StopAsync` with concurrency > 1 awaiting multiple in-flight applies. Add these alongside 1/2.

### Nits (non-blocking)
- **Exception masking in the drain `finally`:** a non-OCE reserve-loop throw can be superseded if `Task.WhenAll(inFlight)` in the `finally` also throws. Rare; both log in practice.
- **Double deserialize per item** — routing peek in `TryReadPlayerId`, then again in `ProcessMessage`. Acceptable for a DB-bound path; thread the parsed envelope through only if profiling ever shows it.
- PR body says "three new cases"; the diff adds four (`DifferentPlayers`, `SamePlayerInterleaved`, `StopDuringDrain`, `LargeBacklog`).

### Follow-ups
None filed — all substantive findings belong **in this PR**, and same-player `SaveChanges` batching is already correctly tracked as #1739 (verified: open, well-scoped, `tech debt`/`claude`/`scope: medium`).

Please rebase onto current `main`, re-integrate #1745's reclaim into the concurrent loop (reconciling the stranded-vs-actively-converging overlap and slot accounting), fix the compaction predicate to `IsCompletedSuccessfully`, implement `GetProcessingCountAsync` on the test double, add the fault-path + `StopAsync`-concurrency>1 tests, and re-run CI. The core concurrency logic is sound — happy to re-review the reconciled loop promptly.

_Automated review (Claude opus-4-8). Not merged._

---

ginderjeremiah:
## Review: Bounded per-player concurrent drain (#1701)

Strong PR — the design is sound and unusually well-documented. I verified the core concurrency properties and they hold:

- **No deadlock** — lane dependencies follow reservation order (a DAG), so the oldest in-flight task never waits on a later one and progress is always guaranteed, even under same-player saturation.
- **Same-player ordering** is strictly preserved via the per-player lane chaining; **cross-player** items genuinely overlap only in apply/acknowledge.
- **Slot accounting** is exactly one release per acquire across every exit path, and the `WaitAsync`-wins-its-own-token re-check is a nice deterministic touch.
- **`StopAsync` boundary** holds — `DrainQueueAsync`'s `finally` awaits all in-flight items before the drain gate is released.

Scoping the same-player `SaveChangesAsync` batching out to #1739 is the right call.

### One should-fix before merge (concurrency/observability)

`CompactCompletedItems` uses `task.IsCompleted`, which is also `true` for **faulted**/**canceled** tasks. Since compaction runs mid-pass (threshold 32), a task that faults before the final `Task.WhenAll(inFlight)` — e.g. `AcknowledgeAsync` or the dead-letter `AddToQueueAsync` throwing on a transient Redis blip — is removed from `inFlight` and its exception becomes an unobserved (silently swallowed) task exception. Without compaction it would propagate to the pub/sub worker and be logged. No data is lost (the un-acknowledged item is reclaimed), but the error surface on the exact large-backlog path this PR targets is lost.

Fix is one line:
```csharp
inFlight.RemoveAll(task => task.IsCompletedSuccessfully);
```
(same predicate for the lane sweep). A test injecting an acknowledge/dead-letter-write failure during a ≥32-item pass and asserting it's surfaced would pin it — the current large-backlog test injects no faults.

### Nits (non-blocking)
- Each item is now JSON-parsed twice (envelope in `PlayerUpdateEnvelopeReader`, then again in `ProcessMessage`/`HandleEvent`). Acceptable to avoid an envelope schema change, but it's on the hot path.
- On shutdown, a full budget of same-player in-flight items is awaited sequentially (chained), so worst-case drain-timeout wait scales with the chain rather than a single apply — bounded and safe, just slightly more than "one item boundary."

Docs, DDD/domain isolation, and test coverage are all in good shape. CI is green (backend + e2e jobs pass; frontend correctly skipped). Happy to approve once the compaction predicate is tightened.

---

ginderjeremiah:
Addressed all three blockers from the review rounds:

- **Reconciled with #1745's opportunistic reclaim** (merged `main` into the branch — non-mechanical, as flagged): `DrainQueueAsync` now settles its own in-flight items *before* consulting `TryReclaimStrandedProcessingAsync` when the queue runs dry. An item mid-apply legitimately sits on the processing list, so counting it would start (or trip) the stranded clock for work that isn't stranded; the barrier costs nothing since the queue is empty at that point. The settle also resets the lanes, so a reclaimed item — including one whose own apply just faulted — re-runs on a fresh lane instead of chaining onto its predecessor's dead lane and re-faulting before ever reaching `ProcessMessage`.
- **`CompactCompletedItems` now evicts only `IsCompletedSuccessfully` tasks.** A faulted (or canceled) lane head stays in the map so later same-player items chain onto it and fault in turn — staying reserved for the reclaim — rather than starting a fresh lane and applying *ahead* of the failed item's eventual re-apply. Verified the new regression test fails against the old `IsCompleted` predicate.
- **Fault-path + multi-item-stop tests added:** `ProcessQueue_ItemFaultsPastCompactionThreshold_LaterSamePlayerItemStaysBlockedForReclaim` (faults an ack, pushes 34 other-player events past the compaction threshold, and pins that the later same-player unequip never applies and both items stay reserved) and `StopAsync_MultipleInFlightApplies_AwaitsEveryOneBeforeReleasing` (two players parked mid-acknowledge under a 2-slot budget; the stop must hold the drain gate until both finish).

Also fixed along the way: the drain-exit settle logs faulted items once and never throws (so it can't mask an exception already unwinding the `finally`), and the `InMemoryPubSubQueue` test helper is now lock-synchronized — its unsynchronized `LinkedList`s corrupted under the genuinely-concurrent acknowledges this PR introduces (it flaked exactly once in a local full-suite run).

Full backend suite: 3096/3096 passing, `dotnet format` clean.

---

PR #1741, opened 2026-07-08
A wedged WebSocket send now falls back to Abort() instead of defeating the watchdog/drain

### LINKED ISSUE #1726: A wedged WebSocket send defeats the inactivity watchdog and the shutdown drain — no Abort() fallback
## Problem

A client that stops reading (TCP zero-window — malicious, or a stalled network) can wedge a `SendAsync` mid-frame while holding the per-socket send lock, and nothing ever escalates to `WebSocket.Abort()`:

- Chunk sends deliberately use `CancellationToken.None` (`Game.Api/Sockets/SocketContext.cs:94-99`), and the response/error sends pass no token (`Game.Api/Sockets/SocketHandler.cs:226`, `:490`).
- `SocketContext.Close` waits on `_sendLock` with the caller's token (`SocketContext.cs:175`); the watchdog calls `Close` with the default token (`SocketHandler.cs:355`). So when the read loop is blocked inside `ExecuteCommand → SendData`, `_lastResponseTicks` stops advancing, the 60s inactivity watchdog fires — and its `Close(Inactivity)` blocks forever on the same `_sendLock`. The close TCS never settles, `SocketInterceptorMiddleware` never unregisters, and the handler + player aggregate + connection leak.
- The graceful shutdown drain is affected too: `_drainCts` only aborts `ReadMessage` (`SocketHandler.cs:365`), not a wedged send, so `SocketConnectionRegistry.DrainAsync`'s post-timeout `await drain` (`Game.Api/Services/SocketConnectionRegistry.cs:145`) can block until the host's force-kill — contradicting the comment that the drain "can't throw / completes promptly". Kestrel enforces no write timeout on upgraded connections.

In production nginx's default `send_timeout` (60s) eventually aborts the chain so it self-heals in ~1–2 minutes; in dev/direct exposure it is unbounded.

## Fix shape

- Bound the watchdog/terminal `Close` with a short CTS and fall back to `WebSocket.Abort()` when it can't acquire the send lock / complete in time.
- Same fallback in `ShutdownAsync` when the drain deadline fires.

## Related test gap

`InactivityTimeout`/`InactivityPollInterval` are hardcoded statics (`SocketHandler.cs:46-49`), so no test pins that an idle socket is actually closed with `ESocketCloseReason.Inactivity` (SocketReadLoopTests only cover the hostStopping path). Make them injectable like `commandTimeout` and add a unit test — that is also where this fix would be pinned.

ginderjeremiah:
## Summary

Closes #1726.

A client that stops reading (TCP zero-window — malicious, or a stalled network) can wedge a `SendAsync` mid-frame while holding `SocketContext`'s per-socket send lock, since chunk sends deliberately run under `CancellationToken.None` once a frame has begun (to avoid interleaving a half-sent frame into the wedged one). Nothing previously escalated to `WebSocket.Abort()`, so:

- The inactivity watchdog's terminal `Close(Inactivity)` call passes no cancellation token at all, so its wait for the send lock never timed out — the socket, handler, player aggregate, and connection leaked indefinitely.
- The graceful shutdown drain fared only slightly better: `_drainCts` cancelling `Close`'s *wait* for the lock did stop that wait, but never unblocked the wedged `SendAsync` itself (it ignores its own token), so `SocketConnectionRegistry.DrainAsync`'s post-timeout `await drain` could still hang until the host force-killed the process — contradicting the "can't throw / completes promptly" comment on that path.

## Fix

`SocketContext.Close` now applies its own short internal timeout (`_closeAbortTimeout`, default 5s) on top of whatever cancellation token it's given, both while waiting for the send lock and while running the close handshake itself. If either doesn't complete in time, it falls back to `WebSocket.Abort()`, which forces the wedged send to fault and release the lock. This covers both callers:

- The watchdog's `Close(Inactivity)` (no caller token) is now bounded by the internal timeout alone.
- The shutdown drain's `Close(ServerShuttingDown, drainDeadline)` is bounded by whichever fires first — the drain deadline or the internal timeout — and either now triggers the abort fallback instead of just giving up on the wait.

Since `Close` no longer throws `OperationCanceledException` in any path, `SocketHandler.ShutdownAsync`'s now-dead `catch (OperationCanceledException)` was removed.

## Related test gap (from the issue)

`InactivityTimeout`/`InactivityPollInterval` were hardcoded statics with no test pinning that an idle socket is actually closed with `ESocketCloseReason.Inactivity`. They're now constructor-injectable (mirroring `commandTimeout`), and a new fast, deterministic test exercises the real watchdog-close path (plus a guard that an *active* socket is never closed).

## Tests

- `Game.Api.Tests/Unit/SocketContextTests.cs`: two new tests pinning the `Close` → `Abort()` fallback — one with no caller token (mirrors the watchdog), one where the caller's token itself cancels mid-wait (mirrors the shutdown drain).
- `Game.Api.Tests/Unit/SocketInactivityTests.cs` (new): pins that an idle socket is closed with `ESocketCloseReason.Inactivity` once the (injected, short) inactivity timeout elapses, and that an active socket is never closed by the watchdog.
- `Game.Api.Tests/Integration/SocketDrainTests.cs`: a new integration test reproduces the reported bug directly — a stalled "pong" send wedges the send lock during a live drain — and asserts `DrainAsync` still completes within its bounded window instead of hanging.

## Test plan

- [x] `dotnet build -warnaserror` — solution builds clean, 0 warnings.
- [x] `dotnet test` — full solution suite passes (Game.Core.Tests, Game.Infrastructure.Tests, Game.Api.Tests, Game.Application.Tests): 3018/3018 passing, including all new and existing socket tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PYtkWLK6EW6S9acc8AtRyp)_


Game.Api.Tests/Unit/SocketInactivityTests.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            // A socket that keeps receiving well within the timeout must never be closed by the watchdog —
+            // guards the fix above against a regression that closes on a fixed schedule instead of tracking
+            // last-activity.
+            var cancellationToken = TestContext.Current.CancellationToken;
+            var socket = new IdleWebSocket();
+            var session = new SessionService(new NoOpSessionStore());
+            var context = new SocketContext(socket, playerId: 1, session, isAdmin: false, _loggerFactory.CreateLogger<SocketContext>());
+            var handler = new SocketHandler(context, new StubCommandFactory(), _scopeFactory, _loggerFactory.CreateLogger<SocketHandler>(),
+                () => { }, inactivityTimeout: TimeSpan.FromMilliseconds(150), inactivityPollInterval: TimeSpan.FromMilliseconds(20));
```
ginderjeremiah: Nit (optional): the active-socket test leans on a 150ms timeout with pings delivered every 50ms — a ~3x margin. It's real-timer based, so under a heavily loaded CI runner a single `Task.Delay(50)` overshooting past 150ms could let the watchdog fire and flake this. The 3x headroom makes it low-risk, but widening the ratio (e.g. a larger inactivity timeout relative to the delivery cadence) would harden it further. Not blocking.

---

ginderjeremiah (commented):
## Review: `SocketContext.Close` now falls back to `Abort()` so a wedged send can't defeat the watchdog/drain (#1726)

**Verdict:** Approve — the fix correctly bounds every `Close` path and adds an `Abort()` escape hatch for the wedged-send case, backed by well-targeted tests and a docs entry. _(Posting as `COMMENT` only because I'm the PR author; GitHub won't accept an approving event on one's own PR.)_

I traced each affected path against the issue and it holds up:

- **Watchdog** (`Close(Inactivity)`, no caller token) — now bounded by `_closeAbortTimeout` alone. Notably this also bounds the `CloseAsync` handshake, which previously ran token-less and could itself hang on a peer that never acks. ✅
- **Shutdown drain** (`Close(ServerShuttingDown, drainDeadline)`) — bounded by whichever of the caller deadline or the internal timeout fires first; on cancel it aborts instead of propagating OCE, so the now-removed `catch (OperationCanceledException)` in `ShutdownAsync` is genuinely dead. ✅
- **Lock accounting** — a cancelled wait leaves the lock unacquired (not released); an acquired lock is released in `finally` even after a `CloseAsync`→abort. `_socketClosedSource.TrySetResult` in the outer `finally` still settles waiters on every path, including abort. ✅

- `Game.Api.Tests/Unit/SocketInactivityTests.cs:77` — real-timer margin in the active-socket test (see inline). Low-risk, take it or leave it.

### Test coverage

Strong and matched to the issue:
- Two `SocketContextTests` cases pin the `Close`→`Abort()` fallback for both triggers — no caller token (watchdog) and a caller token cancelling mid-wait (drain).
- New `SocketInactivityTests` closes the gap the issue called out: an idle socket is closed with `ESocketCloseReason.Inactivity`, and an active socket is never closed by the watchdog (the injectable timeouts make this fast/deterministic).
- New integration test reproduces the reported bug directly — a stalled pong send wedges the lock during a live drain — and asserts `DrainAsync` completes within its bounded window.

Hand-written `WebSocket` stand-ins with no mocking framework, consistent with the project's classical testing approach. Backend-only change, so no frontend/battle-parity mirror is required.

### What's good

- The `boundedCts` linking the caller token with the internal abort timeout is the right shape — one mechanism covers both callers cleanly.
- Comments explain *why* the chunk sends ignore cancellation and why abort is the only escape, which is exactly the kind of non-obvious reasoning worth documenting; the `backend.md` entry captures the design decision at the right altitude.

Not merging from here: CI is still running and there's an outstanding (optional) nit, so I'll leave the merge to a clean, green state.

---

ginderjeremiah:
## Review — Approve

**Verdict: Approve.** The fix is correct, well-scoped, and matches the diagnosis in #1726 (which I verified against the actual code rather than taking on faith). CI is green (`all-checks-passed` success; `test-backend`, `verify-backend`, and all e2e jobs passing).

### What it does
`SocketContext.Close` now layers a short internal timeout (`_closeAbortTimeout`, default 5s) on top of the caller's token, applied both to the `_sendLock.WaitAsync` and to the `CloseAsync` handshake. If either can't complete in time, it falls back to `WebSocket.Abort()`, which faults the wedged `SendAsync` (the mid-frame chunk sends that deliberately run under `CancellationToken.None`) and releases the send lock. This unblocks the two callers the issue identified: the inactivity watchdog's tokenless `Close(Inactivity)` (previously an unbounded leak) and the shutdown drain's `Close(ServerShuttingDown, drainDeadline)`. The now-dead `catch (OperationCanceledException)` in `ShutdownAsync` is correctly removed, and `InactivityTimeout`/`InactivityPollInterval` become constructor-injectable so the watchdog-close path can be pinned deterministically.

### Correctness — verified the load-bearing parts
- **Semaphore permit accounting is right in every path.** On the lock-wait-timeout path, `lockAcquired = false` ensures `Close` does *not* release a lock it never took — the wedged `SendData` releases it itself in its own `finally` when `Abort()` faults its send. On the CloseAsync-timeout path, `Close` did hold the lock and releases it exactly once. No double-release / `SemaphoreFullException`, no leaked permit. This is the subtlest part of the change and it's handled correctly.
- **Idempotent re-close holds.** A second `Close` (e.g. the read loop's terminal `Close(Finished)`) sees a non-Open state and no-ops apart from the `TrySetResult`. The single-close-frame guard (state re-check inside the lock) is preserved.
- **Drain hang is genuinely fixed.** Confirmed the reported failure mode: with a send wedged, the read loop is parked inside `HandleMessage → SendData`, not in `ReadMessage`, so `_drainCts` aborting the receive never unblocked it and `await drain` hung. The abort fallback faults the send, the loop unwinds, `Completion` settles. The 5s internal bound and the 5s drain bound run concurrently (both start when `ShutdownAsync → Close` is invoked), so the effective worst case is ~5s, not additive.
- **Tests are meaningful, not tautological.** The two `SocketContextTests` cover the tokenless-watchdog and caller-token-cancels paths; the integration test reproduces the actual bug (stalled pong send during a live drain); the new `SocketInactivityTests` pin both the idle-close and the active-never-closed guard. Close-reason mapping (`Inactivity → PolicyViolation`) matches the assertion.

### Non-blocking observations (no changes required)
- `InactivityCheckerLoop`'s terminal `await _context.Close(Inactivity)` (SocketHandler.cs:355) sits outside the loop's try/catch, so a non-OCE throw from `Close` (e.g. a `WebSocketException` bubbling out of `CloseAsync`) would fault the watchdog task. This is **pre-existing** (the old `Close` could throw the same) and actually *less* likely now that OCE is swallowed, so it's out of scope here — flagging only for awareness.
- `Close` nesting is fairly deep now (try → if → try/catch → if → try/catch → finally). Readable and each level is justified; not worth churning.

No follow-up issues created — nothing here rises to genuine out-of-scope work.


---

PR #1742, opened 2026-07-08
Fix stale countdown ring on cross-type shared-expiry reset

### LINKED ISSUE #1731: Effect chips show a stale countdown ring when a shared-expiry reset crosses modifier types
## Problem

In `UI/src/lib/battle/battler.ts:312-317`, the shared-expiry reset loop updates `remainingMs`/`renderRemainingMs` on **every** effect view of the attribute, but `durationMs` is updated only on the matched `(attribute, type)` view at line 292.

Scenario: an additive buff (5000ms) and a multiplicative effect (2000ms) are active on the same attribute. Re-applying the 2000ms effect resets the additive view to `remainingMs = 2000` while its `durationMs` stays 5000, so the chip's countdown ring (`renderRemainingMs / durationMs`, `ActiveEffectChips.svelte:155`) renders 40% full immediately — or pins at 100% in the longer-refresh direction.

Display-only: the parity-relevant expiry bookkeeping is correct; only the ring fraction is wrong.

## Fix

One line — also update `durationMs` in the reset loop. Add/extend a battler unit test asserting sibling views adopt the new duration on a cross-type reset.

ginderjeremiah:
## Summary

Closes #1731.

In `UI/src/lib/battle/battler.ts`'s `applyEffect`, the shared-expiry reset loop (the block that resets every effect view on an attribute to the newest application's remaining time) updated `remainingMs`/`renderRemainingMs` on every sibling view, but only updated `durationMs` on the one view matching the applied effect's `(attribute, modifierType)`.

Scenario: an additive buff (5000ms) and a multiplicative effect (2000ms) are both active on the same attribute. Re-applying the 2000ms multiplicative effect resets the additive view's `remainingMs` to 2000 while its `durationMs` stays pinned at 5000 — so `ActiveEffectChips.svelte`'s countdown ring (`renderRemainingMs / durationMs`) renders against the wrong denominator (40% full immediately, or pinned near 100% depending on direction).

This is display-only — the parity-relevant expiry bookkeeping (which effects actually expire when) was already correct; only the ring fraction was wrong.

## Fix

One line: the reset loop now also updates `durationMs` on every sibling view, matching what it already does for `remainingMs`/`renderRemainingMs`.

## Tests

Added a case to `UI/src/tests/lib/battle/battler-effects.test.ts` pinning that a shared-expiry reset crossing modifier types updates the sibling view's `durationMs` (not just `remainingMs`), reproducing the additive/multiplicative scenario from the issue.

## Test plan

- [x] `npm run test:unit:ci` — full frontend suite passes (295 files, 3533 tests).
- [ ] Backend unaffected (frontend-only display bug) — no backend verification needed.

Closes #1731

---
_Generated by [Claude Code](https://claude.ai/code/session_011kCQgnRxcQDky5FGRZcGiy)_

### DIFF
--- a/UI/src/lib/battle/battler.ts
+++ b/UI/src/lib/battle/battler.ts
@@ -311,6 +311,7 @@ export class Battler {
 		// on the same tick (see advanceEffects).
 		for (const v of this.activeEffects) {
 			if (v.attribute === effect.attributeId) {
+				v.durationMs = effect.durationMs;
 				v.remainingMs = effect.durationMs;
 				v.renderRemainingMs = effect.durationMs;
 			}
--- a/UI/src/lib/battle/battler.ts
+++ b/UI/src/lib/battle/battler.ts
@@ -289,7 +289,6 @@ export class Battler {
 		if (view) {
 			view.totalAmount = combinedAmount;
 			view.count++;
-			view.durationMs = effect.durationMs;
 			foldSourceContribution(view.sources, effect.id, amount, isMultiplicative);
 		} else {
 			view = {
--- a/UI/src/tests/lib/battle/battler-effects.test.ts
+++ b/UI/src/tests/lib/battle/battler-effects.test.ts
@@ -310,6 +310,23 @@ describe('Battler skill-effect bookkeeping', () => {
 		expect(battler.activeEffects[0].sources).toEqual([{ sourceId: 1, amount: 10, count: 2 }]);
 	});
 
+	it('adopts the new duration on a sibling view when a shared-expiry reset crosses modifier types', () => {
+		const battler = makeBattler();
+
+		battler.applyEffect(effect(1, EAttribute.Strength, EModifierType.Additive, 5, 5000));
+		battler.applyEffect(effect(2, EAttribute.Strength, EModifierType.Multiplicative, 2, 5000));
+		expect(battler.activeEffects).toHaveLength(2);
+
+		// Re-applying the multiplicative effect at a shorter duration resets the shared remaining for
+		// both views; the additive sibling's durationMs must move with it, not stay pinned at 5000, or
+		// its countdown ring renders against the wrong denominator.
+		battler.applyEffect(effect(2, EAttribute.Strength, EModifierType.Multiplicative, 2, 2000));
+
+		const additiveView = battler.activeEffects.find((v) => v.modifierType === EModifierType.Additive);
+		expect(additiveView?.durationMs).toBe(2000);
+		expect(additiveView?.remainingMs).toBe(2000);
+	});
+
 	it('folds different source effects on one attribute+type into one view with a per-source breakdown', () => {
 		const battler = makeBattler();
 


ginderjeremiah (commented):
## Review: One-line fix for stale effect-chip countdown ring on cross-type shared-expiry reset (#1731)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the fix is correct, minimal, and matches the issue's diagnosis exactly.

I verified the bug end-to-end against the code: `ActiveEffectChips.svelte:155` renders the ring as `renderRemainingMs / durationMs`, and the shared-expiry reset loop (`battler.ts:312-317`) was resetting `remainingMs`/`renderRemainingMs` on every sibling view of the attribute while leaving each sibling's `durationMs` pinned to its own original application — so a sibling's ring rendered against the wrong denominator. Adding `v.durationMs = effect.durationMs` to the loop is the right fix. This is display-only; the parity-relevant expiry bookkeeping (`advanceEffects` / backend `AdvanceEffects`) is untouched, so no mirrored backend test is required.

### Nits

- `battler.ts:292` (`view.durationMs = effect.durationMs;`, outside the diff so noted here rather than inline) is now redundant with the loop you just extended. The matched `view` is always in `this.activeEffects` with `attribute === effect.attributeId`, so the loop at 312-318 now also sets its `durationMs`. It's harmless — just dead now. Optional to drop it (or leave a one-liner noting the loop covers the matched view too); either way not worth blocking on.

### Test coverage

- The added case in `battler-effects.test.ts` reproduces the exact additive(5000ms)/multiplicative(2000ms) scenario from the issue and pins the additive sibling's `durationMs` (not just `remainingMs`) adopting 2000 on the cross-type reset. That's precisely the regression, and it would have failed before the fix. Good, targeted coverage for a display-only bug.

### What's good

- Tight, correctly-scoped fix with a test that pins the actual denominator regression rather than just the shared `remainingMs` that was already correct. The PR description and issue framing are accurate.

CI is still running (frontend/e2e in progress); backend jobs correctly skipped for a frontend-only change. I'm not merging: the auto-merge criteria here require no outstanding nits and green CI, and I've left one (trivial) nit above — your call whether to fold in the `battler.ts:292` cleanup or dismiss it.

---

ginderjeremiah:
## Review: Approve ✅

**Verdict: approve.** Clean, correctly-scoped one-line bugfix with a test that actually reproduces the defect. CI green (frontend, e2e all browsers, verify-frontend all pass; backend jobs correctly skipped as no-change).

### What it does
Fixes the stale countdown ring reported in #1731. In `applyEffect`, the shared-expiry reset loop already synced `remainingMs`/`renderRemainingMs` across every sibling view on an attribute but left `durationMs` pinned on non-matched (cross-type) views. Since `ActiveEffectChips.svelte:155` renders the ring as `renderRemainingMs / durationMs`, a cross-type re-application produced a wrong denominator. The PR moves the `durationMs` assignment out of the matched-view branch and into the reset loop so all same-attribute views stay consistent.

### Verification
- **Correctness confirmed.** Removing `view.durationMs = effect.durationMs` from the `if (view)` branch is safe: the matched view has the same `attribute`, so the reset loop still updates it — no regression for the matched view, and siblings are now covered too.
- **Display-only claim holds.** `advanceEffects` keys expiry off `remainingMs` (`remainingMs -= timeDelta`, `<= 0`), never `durationMs`. `durationMs` is consumed only by the chip ring. So backend parity is genuinely unaffected and no backend/parity-matrix change is warranted.
- **Test is a true reproduction.** Without the fix, the additive sibling's `durationMs` stays at 5000 after the multiplicative re-application at 2000ms, so the added assertion (`durationMs === 2000`) would fail. Good regression guard.

### Nits / follow-ups
None blocking. No follow-up issues needed — scope is exactly right and there's no leftover dead code. (The `else`-branch view initializer still sets all three timing fields before the loop overwrites them, but the object must be fully constructed before being pushed, so that's not removable redundancy.)

Nice tight fix.

---
_Reviewed by Claude Code_

---

PR #1743, opened 2026-07-08
Fix synchronous Worker construction throw bypassing the setInterval fallback

### LINKED ISSUE #1732: createTickSource: a synchronous Worker construction throw bypasses the setInterval fallback and kills the game loop
## Problem

#1676 added a `setInterval` fallback when the tick worker fails — but only via the async `onerror` event (`UI/src/lib/engine/tick-source.ts:14`). A `new Worker(...)` blocked by CSP (`worker-src`) throws **synchronously** in Chromium, which would propagate out of `LogicalEngine.start()` → `startGame` and leave the whole game loop dead with no fallback — exactly the failure #1676 set out to survive.

## Fix

Wrap the worker construction in try/catch and fall into the existing `setInterval` branch. Pairs with a test gap: `tick-source.test.ts` covers only the `onerror` path, so add a case where the `Worker` constructor throws.

ginderjeremiah:
## Summary

Closes #1732.

#1676 added a `setInterval` fallback for when the logical engine's tick worker fails — but only via the async `onerror` event (`UI/src/lib/engine/tick-source.ts`). A `new Worker(...)` blocked by a CSP `worker-src` restriction throws **synchronously** in Chromium, which would propagate out of `createTickSource` → `LogicalEngine.start()` → `startGame` and leave the whole game loop dead with no fallback — exactly the failure #1676 set out to survive.

## Fix

`createTickSource` now wraps the `new Worker(...)` construction in a try/catch. On a synchronous construction failure, it logs the error and falls through to the same `window.setInterval` branch used both for environments without `Worker` and for the existing async `onerror` fallback — no new code path, just reusing what's already there.

## Incidental test fix

While adding a test for the new path, I found the existing `'falls back to window.setInterval when the worker fails to start'` test never restored its `console.error` spy (`errorSpy.mockRestore()` was missing). Because Vitest's `spyOn` on an already-spied method reuses the same mock, that stale spy's call history leaked into later tests in the file and caused a flaky assertion in my new test. Restored it there, matching the pattern already used by the sibling `'logs a diagnosable error...'` test.

## Tests

- `UI/src/tests/lib/engine/tick-source.test.ts`: added a case where the `Worker` constructor itself throws (a `DOMException`, mirroring a CSP rejection), asserting the error is logged once and the source falls back to ticking via `setInterval`.
- Fixed the pre-existing spy leak noted above.

## Test plan

- [x] `npm run test:unit:ci` — full frontend suite passes (295 files, 3533 tests).

---
_Generated by [Claude Code](https://claude.ai/code/session_019YDguG6Xy52pKsqS36fejb)_

### DIFF
--- a/UI/src/lib/engine/tick-source.ts
+++ b/UI/src/lib/engine/tick-source.ts
@@ -10,19 +10,31 @@ export interface TickSource {
 }
 
 export function createTickSource(onTick: () => void): TickSource {
+	let worker: Worker | null = null;
+
 	if (typeof Worker !== 'undefined') {
-		let worker: Worker | null = new Worker(new URL('./tick-worker.ts', import.meta.url), {
-			type: 'module'
-		});
+		try {
+			worker = new Worker(new URL('./tick-worker.ts', import.meta.url), {
+				type: 'module'
+			});
+		} catch (err) {
+			// A CSP worker-src restriction throws synchronously from the constructor itself,
+			// unlike a worker script failure, which only ever surfaces via the async onerror below.
+			console.error('The logical engine tick worker failed to start, falling back to setInterval', err);
+		}
+	}
+
+	if (worker !== null) {
+		const activeWorker = worker;
 		let fallbackHandle: number | null = null;
 
-		worker.onmessage = () => onTick();
-		worker.onerror = (ev) => {
+		activeWorker.onmessage = () => onTick();
+		activeWorker.onerror = (ev) => {
 			if (worker === null) {
 				return;
 			}
 			console.error('The logical engine tick worker failed, falling back to setInterval', ev);
-			worker.terminate();
+			activeWorker.terminate();
 			worker = null;
 			fallbackHandle = window.setInterval(onTick, pollingIntervalMs);
 		};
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: synchronous Worker-construction throw now falls into the setInterval fallback (closes #1732)

**Verdict:** Approve — the fix does exactly what #1732 asked, the failover logic is correct across every path, and it's covered by a focused new test. (Posting as `COMMENT` since I'm the PR author; the intended verdict is **Approve**.)

### What's good

- **Minimal and DRY.** Hoisting `worker` to `null` and wrapping only the `new Worker(...)` in try/catch means a synchronous CSP `worker-src` rejection falls through to the *same* final `window.setInterval` branch already used for the no-`Worker` (SSR/jsdom) case — no new code path, exactly as the issue suggested.
- **Failover logic holds under scrutiny.** `activeWorker` gives the `onmessage`/`onerror` closures a stable non-null capture while `worker` stays the mutable "still live" flag; the `worker === null` guard in `onerror` still prevents a double failover; and `stop()` behaves correctly in all three end states (worker active, async-failed-over, construction-failed).
- **Incidental spy-leak fix is real.** Adding the missing `errorSpy.mockRestore()` to the pre-existing `'falls back... when the worker fails to start'` test closes a genuine cross-test leak (Vitest reuses the same mock when re-spying an already-spied method) and matches the sibling test's pattern.

### Test coverage

- New case `'falls back to window.setInterval when the Worker constructor throws synchronously'` throws a `DOMException` from the constructor (mirroring a CSP rejection) and asserts the error is logged once, ticks continue via `setInterval`, and stop cleans up. This directly covers the gap #1732 called out.
- Not battle logic, so no frontend/backend mirroring applies.

No blocking items, should-fixes, or nits.

---

PR #1744, opened 2026-07-08
Fix skills loadout rollback clobbering an externally-resynced loadout

### LINKED ISSUE #1735: Skills loadout rollback can restore a stale baseline over an externally-resynced loadout
## Problem

In `UI/src/routes/game/screens/skills/skills-view.svelte.ts`, the rollback baseline `committed` is seeded only in the constructor (`:157`) and advanced only by this view's own successful commits (`:517-521`); `syncFromPlayer()` has no other caller. But `playerManager.initialize` (via `resyncPlayerAndInventory`, fired by the `ServerCommandFailed` push — `UI/src/lib/engine/engine.ts:130-135`) can replace the authoritative loadout mid-session while the screen is open.

Scenario: screen mounts with loadout A → a dead-lettered-command resync sets the manager to authoritative loadout B → player edits → `SetSelectedSkills` fails → rollback writes `playerManager.setSelectedSkills(this.committed)` = A, clobbering B; client and server now disagree until reload. The view's `equipped` getter deliberately supports external changes (a test pins that), but the rollback baseline never absorbs them.

Low frequency (needs a resync **and** a failed commit in one screen session), but the failure is silent divergence.

## Fix shape

Fold external manager changes into `committed` while no commit is pending (e.g. in the idle branch of `equipped`), or roll back to the manager's pre-optimistic value instead of the constructor-seeded baseline. Add a test: external resync during an open screen → failed commit → rollback lands on the resynced loadout.

ginderjeremiah:
## Summary

Closes #1735.

In `UI/src/routes/game/screens/skills/skills-view.svelte.ts`, the rollback baseline `committed` was only ever seeded once at construction (via `syncFromPlayer`) and advanced only by this view's own successful commits. But the authoritative loadout can be replaced mid-session by an external resync (e.g. `resyncPlayerAndInventory` fired off a `ServerCommandFailed` push) — and nothing ever folded that external change into `committed`.

Scenario: screen mounts with loadout A → a dead-lettered-command resync sets the manager to authoritative loadout B while no edit is pending → player edits → the commit fails → rollback writes `committed` (still A, or a stale intermediate value) over B, clobbering the resynced state and silently diverging client and server until reload.

## Fix

`equipped` now re-seeds `committed` from `playerManager.selectedSkills` whenever no edit is pending (the idle branch), instead of only at construction:

```ts
readonly equipped = $derived.by(() => {
	if (this.pendingLoadout !== null) {
		return this.pendingLoadout;
	}
	this.committed = [...playerManager.selectedSkills];
	return this.committed;
});
```

Every edit path (`toggle`, `swapInto`, `reorder`) already reads `this.equipped` before calling `commit`, so `committed` is guaranteed current at the moment a new edit starts. This also let the now-redundant `syncFromPlayer` method be folded away — it was only ever called from the constructor and its `committed` seeding is now handled by `equipped` itself.

## Tests

Added a case to `UI/src/tests/routes/game/screens/skills/skills-view.test.ts` reproducing the scenario: a commit succeeds, an external resync lands and settles, then a *further* edit fails — asserting the rollback lands on the resynced loadout rather than the pre-resync baseline. Verified this test fails against the pre-fix code (rolls back to the stale baseline) and passes with the fix.

## Test plan

- [x] `npm run test:unit:ci` — full frontend suite passes (295 files, 3534 tests).

---
_Generated by [Claude Code](https://claude.ai/code/session_018ZLpYVSQ3pKd71G9FTj6P4)_

### DIFF
--- a/UI/src/routes/game/screens/skills/skills-view.svelte.ts
+++ b/UI/src/routes/game/screens/skills/skills-view.svelte.ts
@@ -150,22 +150,23 @@ export class SkillsView {
 	dragIndex = $state<number | null>(null);
 
 	constructor() {
-		this.syncFromPlayer();
-	}
-
-	/** (Re)seed the committed baseline + inspector selection from the player manager. */
-	syncFromPlayer(): void {
-		this.committed = [...playerManager.selectedSkills];
-		this.pendingLoadout = null;
-		if (this.selectedId < 0) {
-			this.selectedId = this.equipped[0] ?? staticData.skills?.[0]?.id ?? -1;
-		}
+		// Reading `equipped` seeds `committed` via its idle branch below.
+		this.selectedId = this.equipped[0] ?? staticData.skills?.[0]?.id ?? -1;
 	}
 
 	/** Working equipped loadout, ordered by priority: the pending optimistic edit while a commit is in
 	 *  flight, otherwise the player manager's committed loadout — so an external loadout change (e.g. a
-	 *  challenge skill unlock or server reconciliation) is reflected once no edit is pending. */
-	readonly equipped = $derived(this.pendingLoadout ?? [...playerManager.selectedSkills]);
+	 *  challenge skill unlock or server reconciliation) is reflected once no edit is pending. While idle
+	 *  (no edit pending), this also re-seeds {@link committed} from the manager, so an external resync
+	 *  becomes the rollback target for the next edit rather than a stale pre-resync baseline — every edit
+	 *  path reads {@link equipped} before calling {@link commit}, so `committed` is always current by then. */
+	readonly equipped = $derived.by(() => {
+		if (this.pendingLoadout !== null) {
+			return this.pendingLoadout;
+		}
+		this.committed = [...playerManager.selectedSkills];
+		return this.committed;
+	});
 
 	/** The loadout cap (number of equip slots) — the single generated game constant. */
 	readonly cap = MAX_SELECTED_SKILLS;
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Fix skills loadout rollback clobbering an externally-resynced loadout (closes #1735)

**Verdict:** Approve (submitted as a comment since I authored the PR) — the fix is correct, minimal, well-documented, and covered by a faithful regression test. No blocking findings, no should-fixes.

### Why it's correct

- `committed` (`skills-view.svelte.ts:133`) is a **plain, non-reactive field**, so re-seeding it inside `equipped`'s `$derived.by` idle branch is an ordinary assignment — it does not trip Svelte's `state_unsafe_mutation`, and it schedules no reactive invalidation, so there's no loop risk.
- The load-bearing invariant holds: every edit entry point (`toggle`, `swapInto`, `reorder`) reads `this.equipped` before calling `commit`, and reading the derived re-runs it whenever `playerManager.selectedSkills` changed since the last read — so `committed` is always the current authoritative loadout at the moment an edit starts. Once a commit is in flight (`pendingLoadout !== null`) the idle branch is skipped, so `committed` only advances via the success path, exactly as before. Traced the new test end-to-end: it rolls back to the resynced `[3]`, not the stale `[1, 2]`.

### Design note (not a finding)

Using a derived side-effect rather than an `$effect` to sync `committed` is the right call here, and worth calling out so a future reader doesn't "correct" it: an `$effect` runs async after render, so `committed` wouldn't be guaranteed current at the synchronous moment an edit fires — reintroducing the very race #1735 is about. The thorough doc comment on `equipped` already captures this.

### Cleanup

Folding away `syncFromPlayer` is correct — its `selectedId < 0` guard was always true at construction (the `$state` default is `-1`), and `pendingLoadout`/`committed` are already at their needed values there. No dead code left behind, and no existing test referenced the removed method.

### Test coverage

Good — the added case in `skills-view.test.ts` reproduces the exact issue scenario (in-flight commit → external resync settles → later edit fails → rollback target) and was verified to fail pre-fix. CI is fully green (frontend unit + verify, all three e2e browsers; backend jobs correctly skipped).

Merging.

---

PR #1745, opened 2026-07-08
Reclaim stranded write-behind processing items opportunistically

### LINKED ISSUE #1702: Reclaim stranded write-behind processing items opportunistically, not only at process startup

ginderjeremiah:
## Summary

Closes #1702.

`ReclaimProcessingAsync` previously only ran in `DataProviderSynchronizer.StartAsync`. If `ProcessMessage` itself throws after a durable apply — the dead-letter `AddToQueueAsync` faulting, or `AcknowledgeAsync` faulting (a Redis blip at exactly the wrong moment) — the reserved item was stranded on `PlayerUpdateQueue:processing` and never retried until *some* instance restarted. Nothing was lost (re-apply is idempotent) but a genuine player write could lag invisibly for a long time in a stable, long-running single-instance deployment.

## Changes

- **`DataProviderSynchronizer.DrainQueueAsync`**: when a drain pass finds the main queue empty, it now also checks the processing list. If it's been observed non-empty for at least a grace period (default 30s), it opportunistically reclaims it and keeps draining — recovering the stranded item on the very next drain pass instead of waiting for a restart. The grace period exists only to avoid needlessly racing an item another live instance may still be genuinely applying (reclaiming early would still be safe under the queue's existing idempotent at-least-once contract, just wasteful); the "first sighting" always just starts the clock rather than reclaiming immediately.
- **`IPubSubQueue.GetProcessingCountAsync`** (new): a non-mutating read of the processing-list depth, implemented for `RedisQueue` (`ListLengthAsync` on the processing key) and every test fake that implements the interface.
- **`DataProviderSynchronizer`** now takes an optional `TimeProvider` (defaulting to `TimeProvider.System`, mirroring the existing `drainTimeout` optional-parameter pattern) so the grace period is deterministically testable.
- `docs/backend-persistence.md` updated to describe the opportunistic reclaim alongside the existing startup reclaim.

## Note on scope

I kept the grace period simple: an instance-local timer that starts on first sighting a non-empty processing list (with the main queue drained) and fires once it's stayed non-empty for 30s straight, resetting whenever the processing list is seen empty. It doesn't track *which* items are stranded vs. freshly reserved, so in a busy multi-instance deployment where the processing list is rarely fully empty, the clock could in principle keep restarting and delay a genuinely-stuck single item. Flagging this here rather than over-engineering it up front — the existing idempotent-apply contract makes an occasional early reclaim harmless, so I'd rather see if this simple version is sufficient in practice before adding e.g. per-item timestamps.

## Test plan

- [x] `dotnet test` — full solution suite passes (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests): 3021/3021 passing.
- New tests:
  - `RedisQueueIntegrationTests.GetProcessingCountAsync_ReflectsReservedButNotYetAcknowledgedItems` (+ cancellation coverage)
  - `DataProviderSynchronizerTests.ProcessQueue_ProcessingListStrandedWithinGracePeriod_DoesNotReclaimYet`
  - `DataProviderSynchronizerTests.ProcessQueue_ProcessingListStrandedPastGracePeriod_ReclaimsAndAppliesTheEvent`

---
_Generated by [Claude Code](https://claude.ai/code/session_01NkqwLcSzYtxsGJPBkMVBpb)_


Game.DataAccess/DataProviderSynchronizer.cs
```diff
@@ -218,6 +251,49 @@ private async Task DrainQueueAsync(IPubSubQueue queue, CancellationToken cancell
...
+                return false;
+            }
+
+            var reclaimed = await queue.ReclaimProcessingAsync(cancellationToken);
+            _strandedProcessingObservedAt = null;
+            if (reclaimed > 0)
+            {
+                _logger.LogInformation(
+                    "Opportunistically reclaimed {Count} player update(s) stranded on the processing list for over {GracePeriod} on queue '{Queue}'.",
+                    reclaimed, _reclaimGracePeriod, Constants.PUBSUB_PLAYER_QUEUE);
```
ginderjeremiah, line 291: Nit (non-blocking, tracked as follow-up #1746): this log asserts the item was "stranded ... for over {GracePeriod}", which is accurate in the current single-instance model. In a busy multi-instance deployment the shared processing list rarely reads empty, so the grace clock never resets and this fires against perfectly healthy in-flight items belonging to other instances — safe under the idempotent contract, but the "stranded" wording would misreport healthy churn and could trip false alarms. No change needed here given the deferral you called out in the PR description; just flagging where the heuristic's simplification surfaces.

---

ginderjeremiah (commented):
## Review: opportunistic reclaim of stranded write-behind processing items (#1702)

**Verdict:** Approve (posted as a comment because GitHub won't accept a gating self-review; treat this as an approval) — correct, safe under the documented at-least-once contract, and well-tested. Nothing blocking.

I verified the change end-to-end rather than just that code was added:

- `TryReclaimStrandedProcessingAsync` runs only when the main queue is drained, and only from `DrainQueueAsync` → `ProcessQueue` under `_drainGate`, so `_strandedProcessingObservedAt` is genuinely touched only on the serialized drain path — the "needs no lock" claim holds (same rationale as `_lastReportedDeadLetterDepth`).
- The count read (`GetProcessingCountAsync` → `LLEN` via the read path) is non-mutating, correctly distinct from `ReclaimProcessingAsync`.
- The reclaim-races-a-live-apply case is safe: `AcknowledgeAsync` is `LREM count 1` and no-ops when another run already reclaimed the item, so the live instance's later ack doesn't fault or re-strand.
- Grace period vs. a healthy apply: default retry policy is 3 attempts × 200 ms exponential (~600 ms worst-case backoff) + apply time, comfortably under 30 s — the "comfortably longer than a healthy apply" comment is accurate.
- The `continue`-then-re-reserve loop re-applies and acknowledges the reclaimed item, and the clock resets on both an empty processing list and a completed reclaim — no re-reclaim loop, no infinite loop (`ProcessMessage` always terminally acks/dead-letters).

### Should-fix

- None.

### Nits

- `DataProviderSynchronizer.cs:289-291` — the "stranded ... for over {GracePeriod}" INFO log is accurate in the current single-instance model but would misreport healthy in-flight items as stranded in a busy multi-instance deployment. Non-blocking; see follow-up below.

### Follow-up (not blocking, already acknowledged in the PR description)

- The grace clock is a single instance-local timer that doesn't distinguish stranded vs. freshly-reserved items. In an effectively single-instance deployment this never fires spuriously (the processing list reads 0 when a healthy drain finishes). In a busy multi-instance future it would reclaim healthy in-flight items ~every 30 s — safe (idempotent) but wasteful and log-noisy. You explicitly deferred the per-item-timestamp refinement; I filed **#1746** (tech debt) to track revisiting it when multi-instance write-behind draining is enabled.

### Test coverage

Solid and matches the project's rules:
- Integration: `GetProcessingCountAsync_ReflectsReservedButNotYetAcknowledgedItems` (reserve/ack transitions) + the already-cancelled-budget cancellation case.
- Unit: within-grace (starts the clock, doesn't reclaim) and past-grace (reclaims + applies + verifies persisted state + processing list ends empty) — including the deterministic `FakeTimeProvider` to cross the grace boundary without wall-clock delays.
- Every `IPubSubQueue` fake updated to implement the new interface member; backend compile (`verify-backend`) is green.

### What's good

- The `TimeProvider` injection mirrors the existing optional-parameter pattern (`drainTimeout`) and keeps the grace logic deterministically testable without touching production defaults.
- Honest, well-scoped PR description that surfaces the heuristic's limitation up front rather than hiding it.
- `docs/backend-persistence.md` updated concisely and accurately alongside the code.

---

PR #1747, opened 2026-07-08
Clear PlayerProgress dirty tracking after a successful save

### LINKED ISSUE #1703: PlayerProgress dirty sets are never cleared after a successful save — close the double-publish sharp edge
## Problem

`PlayerProgressRepository.Save` reads `DirtyStatistics`/`DirtyChallenges`/`DirtyProficiencies` (`Game.DataAccess/Repositories/PlayerProgressRepository.cs:72-124`) but nothing ever resets them (`Game.Core/Progress/PlayerProgress.cs:18-20` — no clear API). Saving the same aggregate twice re-enqueues and re-HSETs every row dirtied since load.

Today every caller loads-mutates-saves exactly once per aggregate (`BattleStatisticsEventHandler` per battle, `OfflineProgressService` once), so this is harmless — but it's an invisible contract a future caller (e.g. a multi-step command that saves progress twice) would silently violate with duplicate publishes.

## Fix

Add an `AcceptChanges()`-style method that clears the dirty sets, called by `Save` after the envelope is buffered (`Save` already captures `changed` up front, so this is safe including in the deferred-flush path). Add a unit test pinning save-twice → second save publishes nothing.

ginderjeremiah:
## Summary

Closes #1703.

`PlayerProgressRepository.Save` reads `DirtyStatistics`/`DirtyChallenges`/`DirtyProficiencies` off the aggregate to build the write-behind event and cache HSET, but nothing ever cleared those dirty sets. Every current caller loads-mutates-saves an aggregate exactly once, so this was harmless in practice — but it was an invisible contract: a future caller that saved the same aggregate twice (e.g. a multi-step command) would silently re-enqueue and re-persist every row dirtied since load, double-publishing the `ProgressUpdated` event.

## Changes

- `Game.Core/Progress/PlayerProgress.cs`: added `AcceptChanges()`, which clears all three dirty sets (statistics, challenges, proficiencies).
- `Game.DataAccess/Repositories/PlayerProgressRepository.cs`: `Save` now calls `progress.AcceptChanges()` right after the write-behind envelope is buffered. This is safe even in the deferred-cache-advance path, since `Save` already snapshots the changed rows (`changed`/`fields`) before touching the dirty sets.
- Tests:
  - `Game.Core.Tests/Progress/PlayerProgressTests.cs`: two new unit tests — `AcceptChanges` clears all three dirty sets, and a subsequent mutation after `AcceptChanges` dirties only the new change (the prior save's rows don't resurface).
  - `Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs`: new integration test pinning that saving the same aggregate a second time (no further mutation) enqueues nothing onto the write-behind queue.

## Test plan

- [x] `dotnet test` — full solution suite passes (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests): 3024/3024 passing, including the new tests above.

Closes #1703

---
_Generated by [Claude Code](https://claude.ai/code/session_01NoQsGMscnAvtjcAi7AoPCH)_

### DIFF
--- a/Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs
@@ -85,6 +85,29 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task Save_SavingTheSameAggregateAgainWithoutFurtherMutation_EnqueuesNothing()
+        {
+            var playerId = await SeedPlayerAsync();
+            using var multiplexer = await ConnectRedisAsync();
+            var redis = multiplexer.GetDatabase();
+
+            using var scope = CreateScope();
+            var repo = scope.ServiceProvider.GetRequiredService<IPlayerProgressRepository>();
+            var progress = new PlayerProgress(MakeDomainPlayer(playerId), [], [], []);
+            progress.RecordBattleCompleted(MakeEnemy(), victory: true, playerDied: false, totalMs: 1000,
+                new BattleStats(), isBossBattle: false, zoneId: 0);
+
+            await repo.Save(progress);
+            Assert.Equal(1, await redis.ListLengthAsync(Constants.PUBSUB_PLAYER_QUEUE));
+
+            // Saving the exact same aggregate a second time, with no new mutation, must not re-enqueue the
+            // rows the first save already persisted — Save clears the dirty tracking once its own envelope
+            // is buffered (the double-publish sharp edge a future multi-step caller could otherwise hit).
+            await repo.Save(progress);
+            Assert.Equal(1, await redis.ListLengthAsync(Constants.PUBSUB_PLAYER_QUEUE));
+        }
+
+        [Fact]
         public async Task Save_WritesCache_SoASubsequentReadIsServedWithoutTheDatabase()
         {
             var playerId = await SeedPlayerAsync();
--- a/Game.Core.Tests/Progress/PlayerProgressTests.cs
+++ b/Game.Core.Tests/Progress/PlayerProgressTests.cs
@@ -845,6 +845,38 @@ namespace Game.Core.Tests.Progress
         }
 
         [Fact]
+        public void AcceptChanges_ClearsAllThreeDirtySets()
+        {
+            var progress = MakeProgress();
+            progress.RecordBattleCompleted(MakeEnemy(), victory: true, playerDied: false, totalMs: 1000,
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Clear `PlayerProgress` dirty tracking after a successful save (#1703)

**Verdict:** Approve — correct, minimal, and well-tested fix for the double-publish sharp edge; the `AcceptChanges()` placement is provably safe on both save paths. (Submitted as a `COMMENT` because GitHub won't accept a self-approval on my own PR; the intent is Approve.)

### Correctness

The one subtlety worth confirming is whether clearing the dirty sets at `PlayerProgressRepository.cs:103` corrupts the **deferred** cache-advance path (`OnFlushed(AdvanceCache)`), which runs after `Save` returns. It doesn't:

- `changed` is fully materialized at `PlayerProgressRepository.cs:76` (`ToCached(...)` projects each dirty row into fresh `Cached*` instances via `.ToList()`), so it's decoupled from the live `HashSet` dirty sets before anything is cleared.
- Both consumers — the write-behind envelope (`:99`) and `fields`/`AdvanceCache` (`:109-110`) — read from `changed`, never from `DirtyStatistics`/`DirtyChallenges`/`DirtyProficiencies`. So `AcceptChanges()` at `:103` can't affect either the enqueued event or the deferred HSET.

That matches the issue's claim that `Save` already captures `changed` up front, and I verified it rather than taking it on faith.

### What's good

- `AcceptChanges()` follows the well-known .NET naming convention (`DataSet.AcceptChanges` / `ChangeTracker.AcceptAllChanges`), so the intent reads immediately, and the XML-doc comment is concise and accurate.
- The inline comment at `:101-102` explains *why* clearing there is safe (envelope already captured), which is exactly the kind of non-obvious invariant worth a comment.
- Domain logic (dirty-set clearing) stays in `Game.Core`; the repository only orchestrates. Consistent with the DDD layering in `docs/backend.md`.

### Test coverage

- Unit (`PlayerProgressTests`): `AcceptChanges_ClearsAllThreeDirtySets` covers all three sets, and `AcceptChanges_ThenMutatingAgain_DirtiesOnlyTheNewMutation` pins the important follow-on invariant — that a post-`AcceptChanges` mutation dirties *only* the new row and the prior save's rows don't resurface.
- Integration (`PlayerProgressRepositoryIntegrationTests`): `Save_SavingTheSameAggregateAgainWithoutFurtherMutation_EnqueuesNothing` pins the exact end-to-end behavior the issue asked for (save-twice → second save enqueues nothing), which is the real regression guard.

No blocking findings, should-fixes, or nits. Nice, tightly-scoped tech-debt fix.


---

PR #1748, opened 2026-07-08
Thread an injected timestamp through challenge completion instead of DateTime.UtcNow

### LINKED ISSUE #1704: PlayerChallenge.UpdateProgress stamps CompletedAt from ambient DateTime.UtcNow instead of an injected timestamp
## Problem

The domain otherwise injects timestamps (`Player.StampActivity(timestamp)`, `UnlockLesson(lessonId, timestamp)`, `RecordBattleCompleted(..., timestamp)` — `Game.Core/Players/Player.cs:428,447,472,499`), but challenge completion reads the wall clock inline: two `DateTime.UtcNow` sites in `Game.Core/Progress/PlayerChallenge.cs:44,58`.

Consequences:

- Offline-window challenge completions are stamped at **claim** time rather than the simulated time within the away window.
- The completion timestamp is untestable/non-deterministic, unlike every other domain write path.

## Fix

Thread the timestamp through `EvaluateChallenges` → `UpdateChallengeProgress` → `UpdateProgress` like the other write paths, passing the simulated time from the offline path and now from the live path. Mechanical; update the affected unit tests to pin the injected value.

ginderjeremiah:
## Summary

`PlayerChallenge.UpdateProgress` stamped `CompletedAt` from the ambient `DateTime.UtcNow` (`Game.Core/Progress/PlayerChallenge.cs:44,58`), unlike every other domain write path, which injects its timestamp (`Player.StampActivity(timestamp)`, `UnlockLesson(lessonId, timestamp)`, `RecordBattleCompleted(..., timestamp)`).

This meant:
- Offline-window challenge completions were stamped at **claim** time rather than the simulated time within the away window — most visibly wrong when a long absence is clamped by `OfflineProgressService.MaximumOfflineSimulation` (10h), where "now" can be hours later than the window the completion was actually simulated within.
- The completion timestamp was untestable/non-deterministic.

## Changes

Threaded a `DateTime timestamp` parameter through the whole chain:

- `PlayerChallenge.UpdateProgress(value, hasData, timestamp)` — stamps `CompletedAt` from the injected value.
- `Challenge.UpdateChallengeProgress(playerChallenge, playerProgress, timestamp)`
- `PlayerProgress.EvaluateChallenges(challenges, timestamp)`
- `ChallengeRewardService.EvaluateAndApply(progress, touchedStatistics, player, timestamp, notify)`

Callers:
- **Live path** (`BattleStatisticsEventHandler`): passes `DateTime.UtcNow` — no behavior change, just now an explicit value instead of an ambient read.
- **Offline path** (`OfflineProgressService.SimulateProgress`): passes the simulated window's end (`player.LastActivity + cappedAwayMs`, captured before anything can re-anchor `LastActivity`) instead of the claim-time `now` — a real behavior fix whenever the away duration exceeds the simulation cap.

## Test plan

- [x] `dotnet test` — full solution suite passes (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests): 3025/3025 passing.
- Updated `PlayerChallengeTests` / `ChallengeTests` / `PlayerProgressTests` to pin an injected timestamp and assert `CompletedAt` equals it (previously only asserted `NotNull`).
- Added `OfflineProgressServiceIntegrationTests.SimulateOfflineProgress_AwayLongerThanSimulationCap_StampsChallengeCompletedAtWithinCappedWindow`, which sets `LastActivity` 20h in the past (beyond the 10h cap) and asserts the completed challenge's `CompletedAt` lands at the capped window's end rather than near "now".

Closes #1704

---
_Generated by [Claude Code](https://claude.ai/code/session_01SG2ifraFVy9xT5Ye9ucDuV)_

### DIFF
--- a/Game.Application.Tests/Services/OfflineProgressServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/OfflineProgressServiceIntegrationTests.cs
@@ -134,6 +134,42 @@ namespace Game.Application.Tests.Services
         }
 
         [Fact]
+        public async Task SimulateOfflineProgress_AwayLongerThanSimulationCap_StampsChallengeCompletedAtWithinCappedWindow()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+
+            var setup = await SeedWinningScenarioAsync(scope, reload: false);
+            var item = await TestDataSeeder.CreateItemAsync(context);
+            var challenge = await TestDataSeeder.CreateChallengeAsync(
+                context, challengeTypeId: EChallengeType.EnemiesKilled, progressGoal: 3m, rewardItemId: item.Id);
+            await ReloadReferenceCachesAsync();
+
+            var (player, state) = await LoadAsync(scope, setup.PlayerId);
+            var offlineProgressService = scope.ServiceProvider.GetRequiredService<OfflineProgressService>();
+            var progressRepo = scope.ServiceProvider.GetRequiredService<IPlayerProgressRepository>();
+
+            // Well beyond MaximumOfflineSimulation (10h): the simulated window is clamped to the cap, so a
+            // challenge completed within it must be stamped at the capped window's end, not at this claim
+            // (which is ~10h later than the window it was actually simulated within).
+            var lastActivity = DateTime.UtcNow.AddHours(-20);
+            player.LastActivity = lastActivity;
+
+            var summary = await offlineProgressService.SimulateOfflineProgress(player, state, CancellationToken);
+
+            Assert.Single(summary.CompletedChallenges, c => c.ChallengeId == challenge.Id);
+            var challenges = await progressRepo.GetChallenges(setup.PlayerId);
+            var playerChallenge = Assert.Single(challenges, c => c.Challenge.Id == challenge.Id);
+            Assert.NotNull(playerChallenge.CompletedAt);
+
+            var expectedWindowEnd = lastActivity + OfflineProgressService.MaximumOfflineSimulation;
+            Assert.True(
+                Math.Abs((playerChallenge.CompletedAt!.Value - expectedWindowEnd).TotalSeconds) < 5,
+                $"Expected CompletedAt near the capped window end {expectedWindowEnd:O}, got {playerChallenge.CompletedAt:O}.");
+            Assert.True(DateTime.UtcNow - playerChallenge.CompletedAt!.Value > TimeSpan.FromHours(9));
+        }
+
+        [Fact]
         public async Task SimulateOfflineProgress_MixedWinLossWindow_CompletesAWinTrackingChallengeCrossedByEarlyWins()
         {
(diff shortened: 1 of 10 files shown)


ginderjeremiah:
## Automated review — approve

**Verdict:** Approve (clean, well-scoped tech-debt fix for #1704)
**CI:** Green — `all-checks-passed` success; `verify-backend`, `test-backend`, and all three `test-e2e` browsers pass. Frontend jobs correctly skipped (no FE changes).

### Scope
Threads an injected `DateTime timestamp` through `EvaluateAndApply` → `EvaluateChallenges` → `UpdateChallengeProgress` → `UpdateProgress`, replacing the two ambient `DateTime.UtcNow` reads in `PlayerChallenge`. 10 files, +107/-55, purely mechanical plumbing plus the offline behavior fix.

### Verification (claims check out)
- **Offline timestamp captured before re-anchor** — confirmed. `awayWindowEnd = player.LastActivity + cappedAwayMs` is computed at the top of `SimulateProgress`, before either `player.StampActivity(now)` re-anchor (early-return path and the final save). No path re-anchors `LastActivity` ahead of the capture, so the stamp is genuinely the simulated window's end.
- **Live path behavior-preserving** — confirmed. `BattleStatisticsEventHandler` now passes an explicit `DateTime.UtcNow`, semantically identical to the old inline read. For a non-capped offline window `awayWindowEnd == now`, so no regression there either; the fix only bites when the away duration exceeds the 10h cap, exactly as intended.
- **No missed callers** — the four changed signatures have only two production callers (`BattleStatisticsEventHandler`, `OfflineProgressService.ApplyOfflineRewards`), both updated; all inner domain callers and every test updated. Challenge evaluation is server-authoritative (client never evaluates challenges), so this is off the FE/BE battle-parity surface — no mirrored frontend change needed.
- **`CompletedAt` now deterministic/testable** — yes; domain tests pin the injected value (`Assert.Equal(Timestamp, ...)`) instead of the old `NotNull`, and the `KeepsOriginalCompletedAt` test now meaningfully asserts the first timestamp survives a later re-evaluation with a different timestamp.

### Findings
**Blocking:** none.

**Nits (non-blocking):**
- The offline batch stamps *all* completions in the pass at `awayWindowEnd`. A challenge that crossed its goal mid-window is stamped at the window's end rather than its true crossing time — an approximation, but strictly tighter than the old claim-time `now` and always within the simulated window, which is the best available without per-battle completion tracking. Worth a one-line code comment if not already obvious to future readers.
- Test code uses `CompletedAt!.Value` (null-forgiving) after `Assert.NotNull`. Common in tests but nominally against the project's avoid-`!` guideline.

_Rationale:_ Faithfully implements #1704, the real offline fix is correct and the timestamp capture is safely ordered ahead of every `LastActivity` re-anchor, and the new integration test plus the tightened domain assertions lock in the deterministic behavior. No follow-up issues warranted.

_— Automated review (Claude Opus 4.8). Do not merge on my say-so alone._

---

PR #1749, opened 2026-07-08
Converge frontend effect expiry onto the backend's absolute-clock model

### LINKED ISSUE #1705: Effect-expiry bookkeeping is structurally different FE vs BE (per-tick remainingMs decrement vs absolute clock) — converge on the absolute model
## Problem

The backend tracks effect expiry against an absolute `_elapsedMs` clock (`Game.Core/Battle/Battler.cs:33,635-671`); the frontend decrements a per-tick `remainingMs` (`UI/src/lib/battle/battler.ts:313-347`). The two are value-equal **only under the fixed 40 ms tick** — the frontend file's own parity note says so (`battler.ts:317-322`).

This is not a defect today (the mirrored apply→expire→re-apply parity scenarios pin it), but it is the **single place the two simulators implement different algorithms rather than mirrored ones** — on the riskiest parity surface. Any future variable-tick or fast-forward work would make it drift silently.

## Suggested fix

Convert the frontend to the backend's absolute-expiry model (keep a derived remaining-ms only where the display needs it), removing the class of risk rather than relying on the note.

## Constraints

- MUST preserve exact frontend/backend battle parity (`docs/backend-battle.md`) — the mirrored parity matrix and the effect-stacking unit tests must pass unchanged on both sides.
- This is a structural refactor with zero behavior change; if any parity vector shifts, the change is wrong.

ginderjeremiah:
## Summary

The frontend's `Battler.advanceEffects` decremented each active effect's `remainingMs` by `timeDelta` every tick, while the backend's `Battler.AdvanceEffects` keys expiry to an absolute per-battler `_elapsedMs` clock. The two were only value-equal under the fixed 40ms tick — the frontend's own parity note said so — making this the one place the two simulators implemented different algorithms rather than mirrored ones, on the riskiest parity surface (attribute-graph effect application).

This PR converts the frontend to the backend's model, per the issue's suggested fix, with zero intended behavior change.

## Changes

- `UI/src/lib/battle/battler.ts`:
  - Added a private `#elapsedMs` clock, advanced unconditionally each `advanceEffects` call — mirroring the backend's `_elapsedMs += ms` ahead of its early-return check, so an effect applied on a later tick still computes its absolute expiry correctly.
  - Added a private `#effectExpiry` map keyed by attribute (mirroring `AttributeEffectStack.ExpiresAtMs`, which is likewise shared across the additive/multiplicative modifier types on one attribute). `applyEffect` sets it to `elapsedMs + durationMs` on (re-)application.
  - `advanceEffects` now compares each active view's stored expiry against the clock instead of decrementing `remainingMs`; `remainingMs` is derived from `expiresAtMs - elapsedMs` for display, so the public `ActiveEffectView` shape (and every existing consumer — the active-effect chips, the attribute tooltip) is unchanged.
  - `reset()` clears the new clock and expiry map. The backend constructs a fresh `Battler` (and so a fresh `_elapsedMs`) per battle; the frontend reuses one `Battler` instance across battles (`#811`), so the clock has to be rewound explicitly there.
- `docs/backend-battle.md`: updated the skill-effects section, which explicitly documented the old "frontend stores a decrementing `remainingMs` instead" caveat, to describe the now-shared absolute-clock model.
- Two test comments (`UI/src/tests/lib/battle/battle-simulation-parity.test.ts`, `Game.Core.Tests/Battle/BattleSimulatorParityTests.cs`) that referenced "the frontend's decrementing remainingMs" as the likeliest drift point were updated to describe the shared model instead — no scenario values changed.

## Why zero behavior change

`remainingMs` at tick *k* after application was previously computed as `durationMs - k·timeDelta` via repeated subtraction; it's now computed as `expiresAtMs - elapsedMs = (elapsedMs_apply + durationMs) - (elapsedMs_apply + k·timeDelta) = durationMs - k·timeDelta` via one subtraction. Since `durationMs` and `timeDelta` are always integer-millisecond values well within double-precision's exact-integer range, both forms produce bit-identical results — there's no rounding-error seam between the two.

## Test plan

- [x] `npm run test:unit:ci` (UI) — 3535/3535 passing, including the full battle-simulation parity matrix (in particular `effectExpiresThenReapplies`, the apply→expire→re-apply scenario called out as the riskiest drift point) and the effect-stacking unit tests in `battler-effects.test.ts` — all pass **unchanged**.
- [x] `npm run lint` (UI) — clean, 0 errors/warnings (eslint + svelte-check).
- [x] `dotnet test` — full solution suite passes: 3024/3024 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests).

Closes #1705

---
_Generated by [Claude Code](https://claude.ai/code/session_01XRxpG7j6GjhARout1DLu77)_


ginderjeremiah (commented):
## Review: Converge frontend effect expiry onto the backend's absolute-clock model (#1705)

**Verdict:** Approve (submitted as a comment since I'm the PR author, so GitHub won't take a gating event on my own PR) — a clean, genuinely behavior-preserving refactor that removes the last place the two simulators used different algorithms rather than mirrored ones.

### Correctness / parity

Verified the change is algebraically identical to the old model under the fixed tick, and now structurally mirrors `Battler.AdvanceEffects`:

- Surviving view: `remainingMs = expiresAtMs − elapsedMs = (elapsedMs_apply + durationMs) − (elapsedMs_apply + k·timeDelta) = durationMs − k·timeDelta` — bit-for-bit the old repeated-subtraction result (integer-ms values, exact in double precision).
- Expiry condition: `expiresAtMs ≤ elapsedMs ⇔ durationMs − k·timeDelta ≤ 0` — the old `remainingMs ≤ 0`, same tick.
- Clock advances unconditionally ahead of the empty-effects early return, matching the backend's `_elapsedMs += ms`, so an effect applied several ticks in still dates its expiry correctly.
- Shared per-attribute expiry: when the stack lapses, the first sibling view deletes the `#effectExpiry` entry and the second falls through the `?? elapsedMs` guard to expire on the same tick — equivalent to the backend removing one `_attributeStacks` entry (both modifiers) at once.
- `reset()` rewinds `#elapsedMs` and clears `#effectExpiry`; correct given the FE reuses one `Battler` across battles (#811) where the backend builds a fresh one. `#elapsedMs` isn't read anywhere absolute, so this is clean rather than load-bearing — good hygiene regardless.
- `renderRemainingMs` is untouched here and still interpolated off `remainingMs` in `updateRenderEffects`, so the countdown ring (and #1742's fix) are unaffected.

### Test coverage

Right strategy for a zero-behavior-change refactor: the mirrored parity matrix (including `effectExpiresThenReapplies`, the apply→expire→re-apply drift point, which applies effects at later ticks and pins death timing) and the effect-stacking unit tests pass **unchanged** — that's the proof the algorithm didn't shift. Only two test comments were reworded to describe the shared model. No new scenario needed.

### What's good

- The parity note that used to say "value-equal only under the fixed tick" is now a real invariant, not a caveat — exactly what the issue asked for.
- `#effectExpiry`/`#elapsedMs` are kept off the reactive `activeEffects` views as private bookkeeping, and the public `ActiveEffectView` shape is unchanged, so no consumer is touched.
- `docs/backend-battle.md` updated to drop the "frontend decrements a separate remainingMs" caveat.

No blocking items, should-fixes, or nits. Merging once CI is green.

---

PR #1750, opened 2026-07-08
Split Player.RecordBattleCompleted into a mandatory-rating victory overload

### LINKED ISSUE #1706: Player.RecordBattleCompleted rating parameters default to 0, silently disabling proficiency accrual — make victory callers pass them explicitly
## Problem

`Player.RecordBattleCompleted`'s `playerRating`/`enemyRating` parameters default to `0` (`Game.Core/Players/Player.cs:466-478`) — the method's own comment acknowledges the hazard: a victory caller that forgets them accrues **zero proficiency XP with no error** (`ProficiencyXpCalculator.cs:43-46` treats a non-positive denominator as no claim).

Both current victory paths thread the ratings correctly; the defaults exist only for loss/abandon convenience. That's a commented footgun where a compile-time guarantee is available.

## Fix

Either split into a victory overload that **requires** the ratings, or drop the defaults and have loss/abandon paths pass `0` explicitly. ~5 lines of churn plus test updates; no behavior change.

ginderjeremiah:
## Summary

`Player.RecordBattleCompleted`'s `playerRating`/`enemyRating` parameters defaulted to `0` (`Game.Core/Players/Player.cs:466-478`). The method's own comment already flagged the hazard: a victory caller that forgot to pass them would accrue **zero proficiency XP with no error** (`ProficiencyXpCalculator.cs` treats a non-positive denominator as no claim). Both current victory paths threaded the ratings correctly, so this was a commented footgun rather than a live bug — but a compile-time guarantee was available and wasn't used.

## Changes

- `Game.Core/Players/Player.cs`: split the single defaultable-parameter method into two, both delegating to a shared private `RecordBattleOutcome`:
  - `RecordBattleCompleted(enemy, result, isBossBattle, zoneId, timestamp)` — for loss/draw outcomes. No rating parameters exist on this overload at all (not just defaulted), so there's nothing to forget.
  - `RecordBattleVictory(enemy, result, isBossBattle, zoneId, timestamp, playerRating, enemyRating)` — for victories. The ratings are required parameters, so a future victory caller that forgets them fails to compile instead of silently zeroing out proficiency accrual.
- `Game.Application/Services/BattleService.cs`: `RecordVictory` (the sole path that always has real ratings — both the explicit win and the won-abandon path route through it) now calls `player.RecordBattleVictory(...)` instead of the old combined method.
- `Game.Core.Tests/Players/PlayerTests.cs`: added coverage that was previously missing at the `Player` level — `RecordBattleCompleted` carries zero ratings onto the event, and `RecordBattleVictory` threads the given ratings onto the event and still stamps activity/raises `PlayerCoreUpdatedEvent` the same way.

No behavior change for either existing call path — the loss/abandon call sites (`BattleService.cs:358,475`) were never reachable for a `Victory` result (both are gated on `!result.Victory` upstream), so they simply drop the two rating arguments they never passed.

## Test plan

- [x] `dotnet test --no-build --no-restore` — full solution suite passes: 3027/3027 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests).

Closes #1706

---
_Generated by [Claude Code](https://claude.ai/code/session_01EkY1kKWURWzEy4mGfkALxJ)_

### DIFF
--- a/Game.Application/Services/BattleService.cs
+++ b/Game.Application/Services/BattleService.cs
@@ -524,7 +524,7 @@ namespace Game.Application.Services
             // Thread both combat ratings onto the battle-completed event so the progress handler can normalize
             // each path's activity by max(playerRating, enemyRating) for the effect-based proficiency accrual
             // (spike #1526 Decision 5) — the same snapshot-measured ratings the exp reward above used.
-            player.RecordBattleCompleted(
+            player.RecordBattleVictory(
                 enemy, result, state.IsBossBattle, state.BattleZoneId ?? player.CurrentZoneId, timestamp,
                 rewards.PlayerRating, rewards.EnemyRating);
 
--- a/Game.Core.Tests/Players/PlayerTests.cs
+++ b/Game.Core.Tests/Players/PlayerTests.cs
@@ -896,6 +896,56 @@ namespace Game.Core.Tests.Players
             Assert.Equal(timestamp, coreUpdated.LastActivity);
         }
 
+        [Fact]
+        public void RecordBattleCompleted_CarriesZeroRatingsOntoEvent()
+        {
+            // No rating parameters exist on this overload at all — a loss/draw never claims proficiency XP.
+            var player = MakePlayer();
+            var result = new BattleResult(Victory: false, PlayerDied: true, TotalMs: 1000, Stats: new BattleStats());
+
+            player.RecordBattleCompleted(MakeEnemy(), result, isBossBattle: false, zoneId: 3, timestamp: DateTime.UtcNow);
+
+            var evt = Assert.Single(player.DomainEvents.OfType<BattleCompletedEvent>());
+            Assert.Equal(0, evt.PlayerRating);
+            Assert.Equal(0, evt.EnemyRating);
+        }
+
+        // ── RecordBattleVictory ──────────────────────────────────────────────
+
+        [Fact]
+        public void RecordBattleVictory_ThreadsCombatRatingsOntoEvent()
+        {
+            var player = MakePlayer();
+            var enemy = MakeEnemy(id: 5);
+            var result = new BattleResult(Victory: true, PlayerDied: false, TotalMs: 3200, Stats: new BattleStats());
+
+            player.RecordBattleVictory(
+                enemy, result, isBossBattle: true, zoneId: 7, timestamp: DateTime.UtcNow,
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: Split `Player.RecordBattleCompleted` into a mandatory-rating victory overload (closes #1706)

**Verdict:** Approve — the change implements exactly what the issue asked for (option 1: a victory overload that *requires* the ratings), turns the commented footgun into a compile-time guarantee, and preserves behavior on every existing path. (Submitted as a comment because GitHub won't accept a gating verdict on one's own PR; the intended verdict is **approve**.)

I verified the change end-to-end rather than trusting the description:

- `RecordBattleOutcome` (private) is the single implementation; `RecordBattleCompleted` delegates with `0/0`, `RecordBattleVictory` delegates with the caller's ratings. `Player.cs:485-503`
- Both loss/draw call sites now use the parameterless overload and are genuinely non-victory-gated: `BattleService.cs:358` is the `EndBattleLoss` handler, `:475` is the genuine-draw branch (the still-in-progress case returns early above it). They never passed ratings before, so dropping the two default args is a no-op for them.
- The sole victory path, `RecordVictory` (`BattleService.cs:527`), now calls `RecordBattleVictory` with the snapshot-measured ratings — and it's the only route for both the explicit win and the won-abandon path, so no victory caller is left on the completed overload.
- The boss loss/draw `AutoChallengeBoss = false` backstop still lives in the shared `RecordBattleOutcome`, so it fires identically for both overloads.

### Test coverage

Solid and appropriately targeted. `PlayerTests` gains the previously-missing `Player`-level coverage: `RecordBattleCompleted` carries `0/0` onto the event, `RecordBattleVictory` threads the given ratings, and the victory path still stamps `LastActivity` / raises `PlayerCoreUpdatedEvent`. This is a backend recording change (not battle-simulation logic), so no frontend mirror is required.

### What's good

- Chose the compile-time-safe split over the "pass 0 explicitly" alternative — the stronger of the two options the issue offered.
- Comments were rewritten to describe the *new* contract (why the split exists) rather than narrating the change, matching the docs' guidance on keeping design rationale, not history.
- Genuinely zero behavior change, and the diff stays within the ~5-line scope the issue estimated.

Nothing blocking and no nits. Will merge once the backend/e2e checks finish green.

---

PR #1751, opened 2026-07-08
Untrack lesson-state reads in the screen-activation effect (#1709)

### LINKED ISSUE #1709: Screen-anchored tutorial $effect over-tracks lesson state — untrack the lesson reads and add game-shell component test coverage
## Problem

The screen-activation `$effect` in the game shell (`UI/src/routes/game/+page.svelte:144-148`) calls `evaluateScreenTrigger`, which synchronously reads `staticData.lessons` and `playerManager.lessons` via `isLocked` (`UI/src/lib/engine/tutorials.ts:8-12,54-64`; `player-manager.ts:173-179`). The effect therefore tracks those signals: **any** lesson-state mutation — a mechanic-anchored `unlockLesson` push from the battle tick, or `markLessonRead` on tour close — re-runs the screen-trigger evaluation, not just a screen activation.

Today the worst outcome is masked: `tutorialTour.play(sameLessonRef)` is a same-reference `$state` write (no signal), so a mid-tour re-open is a visual no-op. But the effect silently re-invokes `openLesson`/`requestScreen` mid-tour, and the moment two ScreenVisit lessons target the same screen, the second will auto-play the instant the first is marked read — bypassing the "player just navigated here" framing. With lesson content growing (#1605, #1675), this will surface.

## Fix

`untrack()` the lesson reads inside the effect (or guard on a last-evaluated-screen key), so the effect re-runs only on actual screen changes.

## Related test gap

`routes/game/+page.svelte` has no component test at all — the navigation-intent effect (same-screen `screenNonce` remount so a deep-link payload is consumed), the `welcome.phase === 'entered'` gating of `evaluateScreenTrigger`, and `handleNavigate`'s action keys are only covered indirectly. A regression (e.g. dropping the nonce bump) would silently break the Statistics→Codex deep-link. One component test covering the shell wiring plus this fix is good value — do them together.

ginderjeremiah:
## Summary

The screen-activation `$effect` in the game shell (`UI/src/routes/game/+page.svelte`) called `evaluateScreenTrigger` directly inside its body, so the effect tracked *everything that function reads* — including `staticData.lessons` and `playerManager.lessons` (via `isLocked`), not just the screen key. Any lesson-state mutation elsewhere — a mechanic-anchored `unlockLesson` push from the battle tick, or `markLessonRead` on tour close — re-ran the effect and re-evaluated the screen trigger, even though nothing about "which screen is active" had changed.

Today the worst-case outcome was masked because `tutorialTour.play(sameLessonRef)` is a same-reference `$state` write (no signal), but the effect was still silently re-invoking `openLesson`/`requestScreen` mid-tour on every unrelated lesson-state write, and the issue calls out that the moment two `ScreenVisit` lessons target the same screen this becomes visibly wrong (the second would auto-play the instant the first is marked read).

## Fix

`UI/src/routes/game/+page.svelte`: the effect now reads `welcome.phase` and `currentScreen` normally (so those two still drive re-evaluation), then calls `evaluateScreenTrigger` inside Svelte's `untrack()` so the function's *internal* reads of `staticData.lessons`/`playerManager.lessons` are no longer tracked by this effect.

## Tests

`routes/game/+page.svelte` had no component test at all before this PR. Added `UI/src/tests/routes/game/page.test.ts` (with stub components under `UI/src/tests/routes/game/page/`) covering:

- The `welcome.phase === 'entered'` gating of the screen trigger, **and the fix itself** — a 0-step lesson causes `openLesson` to call `playerManager.markLessonRead` from *inside* `evaluateScreenTrigger`, a self-triggered `playerManager.lessons` write that is exactly the regression this issue describes; the test asserts `evaluateScreenTrigger` stays at a call count of 1 across that write and a further unrelated lesson-state write.
- The navigation-intent effect's same-screen `screenNonce` remount (so a deep-link payload is consumed) vs. an ordinary cross-screen switch (no forced remount).
- `handleNavigate`'s action keys (`admin` navigates, `quit` confirms, `switch` opens/closes the character switcher).
- The sidebar `bind:pinned` two-way binding, the `PlaceholderScreen` fallback for an unbuilt screen, and the shared `TourPlayer`/`closeTutorialTour` wiring.

This intentionally uses the **real** `evaluateScreenTrigger` (wrapped as a spy via `importOriginal`), not a mock — a fully-mocked version would hide the exact bug/fix, since the bug lives in which reads the effect tracks.

## Test plan

- [x] `npm run test:unit:ci` — 296 files / 3542 tests pass, coverage gate (`src/routes/**` ratchet floor) passes

## Note on scope

The issue's linked `+page.svelte:144-148` line numbers reflect the file as originally reported; the fix targets the same block (now at a slightly different offset from unrelated changes since). No other unclear points — this was a well-scoped, unambiguous fix.

Closes #1709

---
_Generated by [Claude Code](https://claude.ai/code/session_012iGQuexYYFeJybWrRTTd3w)_

### DIFF
--- a/UI/src/routes/game/+page.svelte
+++ b/UI/src/routes/game/+page.svelte
@@ -47,7 +47,7 @@ import { apiSocket, getRoles } from '$lib/api';
 import { browser } from '$app/environment';
 import { goto } from '$app/navigation';
 import { resolve } from '$app/paths';
-import { onDestroy, onMount } from 'svelte';
+import { onDestroy, onMount, untrack } from 'svelte';
 import { confirmQuit } from './game-actions';
 import { WelcomeBackView } from './welcome-back/welcome-back-view.svelte';
 import WelcomeBackGate from './welcome-back/WelcomeBackGate.svelte';
@@ -140,10 +140,16 @@ $effect(() => {
 
 // Screen-anchored tutorial trigger (#1587): the single screen-activation point cross-screen
 // navigation uses. Gated on `entered` so a lesson can't fire against the default screen before the
-// welcome-back gate passes (the player hasn't "navigated" anywhere yet at that point).
+// welcome-back gate passes (the player hasn't "navigated" anywhere yet at that point). Only
+// `welcome.phase` and `currentScreen` should drive a re-evaluation — `evaluateScreenTrigger` also
+// reads `staticData.lessons`/`playerManager.lessons` (#1709), and those must NOT be tracked here or
+// any lesson-state mutation (e.g. a mechanic-anchored unlock, or marking a lesson read) would re-run
+// this effect and re-open/re-queue a lesson mid-tour instead of only on an actual screen change.
 $effect(() => {
-	if (welcome.phase === 'entered') {
-		evaluateScreenTrigger(currentScreen);
+	const phase = welcome.phase;
+	const screenKey = currentScreen;
+	if (phase === 'entered') {
+		untrack(() => evaluateScreenTrigger(screenKey));
 	}
 });
 
--- a/UI/src/tests/routes/game/page.test.ts
+++ b/UI/src/tests/routes/game/page.test.ts
@@ -1,0 +1,233 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { render, cleanup, screen, fireEvent, waitFor } from '@testing-library/svelte';
+import { ELessonTriggerType } from '$lib/api';
+import type { ILesson } from '$lib/api';
+import { staticData } from '$stores/static-data.svelte';
+import { navigation, requiresRemount } from '$stores/navigation.svelte';
+import { tutorialTour } from '$stores/tutorial-tour.svelte';
+import { playerManager } from '$lib/engine/player/player-manager';
+import { mountTracker } from './page/screen-mount-tracker';
+import FightScreenStub from './page/FightScreenStub.svelte';
+import StatsScreenStub from './page/StatsScreenStub.svelte';
+import NavSidebarStub from './page/NavSidebarStub.svelte';
+import LogPanelStub from './page/LogPanelStub.svelte';
+import TourPlayerStub from './page/TourPlayerStub.svelte';
+import CharacterSwitcherStub from './page/CharacterSwitcherStub.svelte';
+import BootSplashStub from './page/BootSplashStub.svelte';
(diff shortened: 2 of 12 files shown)


ginderjeremiah (commented):
## Review: Untrack lesson-state reads in the screen-activation effect (#1709)

**Verdict:** Approve — the fix is exactly what the issue calls for, minimally scoped, and the new test exercises the real regression rather than a mock. (Submitted as a `COMMENT` only because I'm the PR author; GitHub won't accept a gating verdict on one's own PR.)

### Correctness

The screen-activation `$effect` now reads `welcome.phase` and `currentScreen` at the top (so those two still drive re-evaluation) and calls `evaluateScreenTrigger` inside `untrack()`, so the function's internal reads of `staticData.lessons`/`playerManager.lessons` (through `isLocked`) are no longer tracked. This is the `untrack()` option the issue recommended, done idiomatically — the tracked reads are correctly hoisted outside the `untrack` callback rather than left inside it. Behavior on an actual screen change or welcome-gate transition is unchanged; only the spurious lesson-mutation re-runs are eliminated.

### Test coverage

Strong, and it closes the "related test gap" the issue flagged. `+page.svelte` had no component test before; the new `page.test.ts` covers the welcome-gate `entered` gating, the same-screen `screenNonce` remount vs. cross-screen switch, `handleNavigate`'s action keys, the `bind:pinned` two-way binding, the `PlaceholderScreen` fallback, and the `TourPlayer`/`closeTutorialTour` wiring. Test location mirrors the source path per `frontend.md`.

The regression test itself is the highlight: it spies on the **real** `evaluateScreenTrigger` (via `importOriginal`) rather than mocking it, so the assertion `toHaveBeenCalledTimes(1)` genuinely fails on the pre-fix code — the 0-step lesson's self-triggered `markLessonRead` write would re-run the effect and re-invoke the trigger. Wrapping the spy around the real function is the correct call here, since the bug lives in *which reads the effect tracks*; a full mock would hide it. The stub components are minimal and each documents its purpose.

### What's good

- The inline comment on the effect explains *why* the reads must not be tracked (mid-tour re-open, and the two-same-screen-lesson case), which protects against a future edit naively folding the `untrack` back in.
- Uses `untrack` from `svelte` rather than reaching for a new store or a manual last-evaluated-key guard — consistent with the project's reactivity approach.

No blocking items, should-fixes, or nits. I'll merge once CI is green.


---

PR #1752, opened 2026-07-08
Make statify's idempotency marker non-enumerable

### LINKED ISSUE #1710: statify writes an enumerable __statifyPerformed marker onto every statified instance
## Problem

`statify`'s idempotency flag is set as a plain enumerable property (`UI/src/lib/common/statify.svelte.ts:29`), so `__statifyPerformed` shows up in `Object.keys`/spread/`JSON.stringify` of any statified object — every manager, battler, and inventory `Item`.

Nothing currently serializes a statified object onto the wire (checked the socket command call sites — all send primitives/fresh DTOs), but an `Item` accidentally passed into a command payload would leak the marker and its accessor-backed fields.

## Fix

One line: define the marker with `Object.defineProperty(..., { enumerable: false })` or key it with a `Symbol`. Add/adjust a statify unit test asserting the marker doesn't appear in `Object.keys` of a statified instance.

ginderjeremiah:
## Summary

`statify`'s idempotency flag was set as a plain enumerable property (`UI/src/lib/common/statify.svelte.ts:29`), so `__statifyPerformed` showed up in `Object.keys`/spread/`JSON.stringify` of any statified object — every manager, battler, and inventory `Item`. No current call site serializes a statified object onto the wire, but a statified instance accidentally passed into a command payload would have leaked the marker.

## Changes

- `Object.defineProperty(statified, '__statifyPerformed', { value: true, enumerable: false })` instead of a plain assignment.
- Removed the now-unreachable `if (property === '__statifyPerformed') continue;` guard in the `for...in` loop — a non-enumerable property never appears there, so the check was dead code.
- Added a test asserting the marker doesn't appear in `Object.keys` or `JSON.stringify` of a statified instance.

## Test plan

- [x] `npm run lint` — clean, 0 errors/warnings.
- [x] `npm run test:unit:ci` — full suite passes: 3543/3543 tests across 296 files, including the new assertion.

Closes #1710

---
_Generated by [Claude Code](https://claude.ai/code/session_01QBzddcUESL78FxvaYAriyy)_

### DIFF
--- a/UI/src/lib/common/statify.svelte.ts
+++ b/UI/src/lib/common/statify.svelte.ts
@@ -26,13 +26,9 @@ export const statify = <T>(state: T) => {
 		return state;
 	}
 
-	statified.__statifyPerformed = true;
+	Object.defineProperty(statified, '__statifyPerformed', { value: true, enumerable: false });
 
 	for (const property in state) {
-		if (property === '__statifyPerformed') {
-			continue;
-		}
-
 		const data = statified[property];
 		const dataType = typeof data;
 		if (dataType === 'function') {
--- a/UI/src/tests/lib/common/statify.svelte.test.ts
+++ b/UI/src/tests/lib/common/statify.svelte.test.ts
@@ -39,6 +39,13 @@ describe('statify', () => {
 		expect(statify(outer)).toBe(outer);
 	});
 
+	it('stamps the idempotency guard as non-enumerable', () => {
+		const outer = statify(new Outer());
+
+		expect(Object.keys(outer)).not.toContain('__statifyPerformed');
+		expect(JSON.stringify(outer)).not.toContain('__statifyPerformed');
+	});
+
 	it('recursively statifies nested class instances and array elements', () => {
 		const outer = statify(new Outer());
 


ginderjeremiah (commented):
## Review: Make `statify`'s idempotency marker non-enumerable (#1710)

**Verdict:** Approve — the fix matches the issue's prescribed remedy exactly, is correct, and cleanly removes the now-dead guard. (Submitted as a `COMMENT` because GitHub won't accept an `APPROVE` from the PR author; treat this as an approval.)

### What it does

`__statifyPerformed` is now stamped via `Object.defineProperty(..., { enumerable: false })` instead of a plain assignment, so it no longer surfaces in `Object.keys`/spread/`JSON.stringify` of any statified manager, battler, or inventory `Item` — closing the marker-leak hazard #1710 describes.

### Correctness / maintainability

- The removed `for...in` guard (`if (property === '__statifyPerformed') continue;`) was only ever needed because the marker was enumerable. A non-enumerable property never appears in `for...in`, so its removal is correct dead-code cleanup, per the CLAUDE.md rule against leaving dead code behind.
- Both remaining consumers read the marker by **direct property access** — the re-entry short-circuit (`statify.svelte.ts:25`) and the test's `wasStatified` helper — which works unchanged on a non-enumerable property, so idempotency is preserved.
- The `defineProperty` default of `writable:false`/`configurable:false` is safe here: the guard at line 25 short-circuits on re-entry, so `defineProperty` is only ever invoked once per instance and the marker is never reassigned or deleted anywhere in the codebase.

### Test coverage

Good. The new `stamps the idempotency guard as non-enumerable` case pins both `Object.keys` and `JSON.stringify` exclusion — directly covering the regression — and the pre-existing idempotency/reactivity tests still exercise the marker via direct access, so nothing lost coverage. Mirrors the source under `tests/lib/common/`, per the frontend testing guideline.

### What's good

Minimal, surgical change that also tidies up the dead guard rather than leaving it. Clear PR write-up that correctly notes no call site currently serializes a statified object, framing this as closing a latent footgun.


---

PR #1754, opened 2026-07-09
Extract shared PollingHelper for integration-test polling loops

### LINKED ISSUE #1723: Consolidate duplicated "poll until condition or timeout" helpers in integration tests
## Summary

The backend integration test suite has accumulated several near-identical "poll a Redis read until a condition holds, or fail on a timeout" helpers, driven by `HashSetAndForget`'s fire-and-forget writes. As of #1721 there are at least:

- `WaitForHashFieldCountAsync` in `Game.Application.Tests/DataAccess/RedisServiceIntegrationTests.cs:278` (operates on `ICacheService`, exact count, `Assert.Fail` on timeout)
- `WaitForHashFieldCountAsync` in `Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs` (operates on raw `IDatabase`, `>= minCount`, returns last read on timeout)
- `WaitUntilValueEqualsAsync` in `RedisServiceIntegrationTests.cs`

They differ in the abstraction they poll and in timeout behavior, which is why #1721 duplicated rather than reused — a reasonable call for that fix. But the underlying pattern (poll every 25ms until a predicate holds or a deadline passes) is the same everywhere.

## Suggestion

Extract a single generic polling helper into a shared integration-test utility, e.g.:

```csharp
static Task<T> PollUntilAsync<T>(Func<Task<T>> read, Func<T, bool> satisfied, int timeoutMs = 5000)
```

and express the hash/value waiters in terms of it. This removes the duplication while letting each call site keep its own predicate and its own "assert vs. return" decision.

## Notes

- No user-facing effect; purely test-suite maintainability.
- Low risk, self-contained.

ginderjeremiah:
## Summary

The backend integration test suite had accumulated several near-identical "poll a Redis read until a condition holds, or fail on a timeout" loops, driven by `HashSetAndForget`/`ReclaimAndForget`'s fire-and-forget writes racing a read taken on a separate connection right after (#1718). Per the issue, they differed in the abstraction polled and in timeout behavior (assert-and-fail vs. return-last-read), but the underlying loop — poll every 25ms until a predicate holds or a deadline passes — was identical everywhere.

## Changes

- **`Game.TestInfrastructure/Helpers/PollingHelper.cs`** (new): a single generic `PollUntilAsync<T>(Func<Task<T>> read, Func<T, bool> satisfied, int timeoutMs = 5000)`. It always returns the last read value — even on timeout — so each call site keeps its own pass/fail decision (`Assert.Fail`, `Assert.NotNull`, or just returning the value), matching the issue's suggested shape.
- **`RedisServiceIntegrationTests.cs`**: expressed both named waiters (`WaitForHashFieldCountAsync` on `ICacheService`, `WaitUntilValueEqualsAsync`) in terms of the helper, and additionally converted three inline ad-hoc polling loops in the same file (TTL-refresh wait, two hash-field waits) that had the identical pattern but weren't separately named — the issue's own notes call out that the pattern is "the same everywhere," and these were a mechanical fit for the same fix.
- **`PlayerProgressRepositoryIntegrationTests.cs`**: expressed its `WaitForHashFieldCountAsync` (on raw `IDatabase`, `>= minCount`, returns last read on timeout) in terms of the helper.

No behavior change — purely test-suite maintainability, as called out in the issue.

## Test plan

- [x] `dotnet test --no-build --no-restore` — full solution suite passes: 3027/3027 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), including both modified integration test classes.

Closes #1723

---
_Generated by [Claude Code](https://claude.ai/code/session_012TRToBYehkQe2fjbvPFujA)_

### DIFF
--- a/Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs
@@ -416,22 +416,9 @@ namespace Game.Application.Tests.DataAccess
         // HashSetAndForget dispatches its HSET with CommandFlags.FireAndForget (RedisService.cs), so it can
         // return before the write lands on the server — a read on a separate connection right after a Save can
         // race it. Poll until the expected field count shows up rather than asserting on a single read (#1718).
-        private static async Task<HashEntry[]> WaitForHashFieldCountAsync(IDatabase redis, string hashKey, int minCount, int timeoutMs = 5000)
+        private static Task<HashEntry[]> WaitForHashFieldCountAsync(IDatabase redis, string hashKey, int minCount, int timeoutMs = 5000)
         {
-            var deadline = DateTime.UtcNow.AddMilliseconds(timeoutMs);
-            HashEntry[] fields;
-            do
-            {
-                fields = await redis.HashGetAllAsync(hashKey);
-                if (fields.Length >= minCount)
-                {
-                    return fields;
-                }
-
-                await Task.Delay(25);
-            } while (DateTime.UtcNow < deadline);
-
-            return fields;
+            return PollingHelper.PollUntilAsync(() => redis.HashGetAllAsync(hashKey), fields => fields.Length >= minCount, timeoutMs);
         }
 
         private static async Task<ProgressUpdatedEvent> DequeueProgressEvent(IDatabase redis)
--- a/Game.Application.Tests/DataAccess/RedisServiceIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/RedisServiceIntegrationTests.cs
@@ -1,6 +1,7 @@
 using Game.Abstractions.Infrastructure;
 using Game.TestInfrastructure.Base;
 using Game.TestInfrastructure.Fixtures;
+using Game.TestInfrastructure.Helpers;
 using Microsoft.Extensions.DependencyInjection;
 using StackExchange.Redis;
 using Xunit;
@@ -136,18 +137,7 @@ namespace Game.Application.Tests.DataAccess
 
             cache.ReclaimAndForget(key, "owner-a", TimeSpan.FromSeconds(30));
 
-            var deadline = DateTime.UtcNow.AddSeconds(5);
-            TimeSpan? ttl = null;
-            while (DateTime.UtcNow < deadline)
-            {
-                ttl = await ReadTtlAsync(key);
-                if (ttl > TimeSpan.FromSeconds(10))
-                {
-                    break;
-                }
-
-                await Task.Delay(25);
-            }
+            var ttl = await PollingHelper.PollUntilAsync(() => ReadTtlAsync(key), t => t > TimeSpan.FromSeconds(10));
 
             Assert.True(ttl > TimeSpan.FromSeconds(10), $"Expected the TTL to be extended past its 2s seed but was {ttl}.");
             Assert.Equal("owner-a", await cache.Get(key));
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: Extract shared `PollingHelper` for integration-test polling loops (closes #1723)

**Intended verdict:** Approve — clean, faithful extraction that does exactly what the issue asked, with no behavior change. _(Submitted as a comment because GitHub won't accept a gating verdict from the PR author.)_

Reviewed the new helper and all four+ refactored call sites against the originals. This is a textbook tech-debt consolidation and I have no blocking or should-fix findings.

### What's good

- **`PollingHelper.PollUntilAsync`** lands exactly on the signature the issue suggested (`Func<Task<T>> read, Func<T, bool> satisfied, int timeoutMs`), and returns the last read value so each call site keeps its own assert-vs-return decision — the whole point of #1723. Placement (`Game.TestInfrastructure/Helpers/`, block-scoped namespace) matches the surrounding static utilities and both test projects already reference it.
- **Behavior is preserved at every call site.** Predicates map 1:1 onto the originals (`>= minCount`, `is not null`, `?.GetValueOrDefault("a") == "9"`, `?.Count == count`), the assert-on-timeout waiter still `Assert.Fail`s, and the nullable-`TimeSpan` comparison keeps its `null > 10s == false` semantics. The `do/while` guarantees at least one read, matching the originals in practice.
- Going beyond the three helpers the issue named to also fold the three inline ad-hoc loops in `RedisServiceIntegrationTests` was the right call — they were the identical pattern, and leaving them would have half-solved the duplication.
- No dead code left behind; the unneeded `async` state machines on the now-delegating waiters were correctly dropped.

### Test coverage

Adequate. This is test infrastructure exercised transitively by the two integration-test classes that consume it, so a dedicated unit test for a timing/delay-based poller would be low-value and slow — reasonable to omit. `WaitUntilValueEqualsAsync` now returns based on the last polled value instead of issuing an extra fresh read after timeout; functionally equivalent (marginally cleaner), not a regression.

Nice, self-contained cleanup. CI is still running — once it's green this is good to merge.

---

ginderjeremiah:
## Automated review — approve with nits

**Verdict:** Approve (with minor, non-blocking nits)
**CI:** Green — `test-backend`, `verify-backend`, all three `test-e2e` browsers, and `all-checks-passed` succeeded; frontend jobs correctly skipped (no frontend changes).

### Scope
Test-infrastructure-only. Adds a single generic `PollingHelper.PollUntilAsync<T>` and re-expresses the three named waiters called out in #1723 (`WaitForHashFieldCountAsync` on `IDatabase`, `WaitForHashFieldCountAsync` on `ICacheService`, `WaitUntilValueEqualsAsync`) plus three previously-inline ad-hoc loops in `RedisServiceIntegrationTests` in terms of it. Net -29 lines. Matches the issue's suggested `PollUntilAsync(read, satisfied, timeoutMs)` shape exactly, including the "return last read even on timeout, leave the assert/return decision to the caller" contract.

### Correctness of the extraction
The helper's `do { read; check; delay } while (now < deadline)` preserves the "always read at least once, return the last read on timeout" semantics of the original `do/while` version, and is a faithful superset of the `while`-guarded call sites (all have a 5s deadline, so they always execute ≥1 read). Each converted site keeps its original pass/fail decision:
- `IDatabase` count waiter → returns last read (unchanged).
- `ICacheService` count waiter → re-checks `fields?.Count != count` and `Assert.Fail`s (faithful; the re-check re-evaluates the predicate, which is inherent to the return-last-value design).
- Inline TTL / hash-field loops → return the value, assertions unchanged.

No behavior-affecting change found. Design of the generic helper is clean, correctly typed (`T = TimeSpan?` / nullable dictionary flow through the predicates fine), and well-documented.

### Nits (non-blocking)
1. **`WaitUntilValueEqualsAsync` drops a redundant extra read.** The original did one final fresh `await cache.Get(key)` *after* the loop on timeout; the new version returns `value == expected` from the last polled read. This is a (beneficial) micro-behavior change rather than strictly "no behavior change" — negligible, since the helper already polls to the deadline. Fine to keep as-is.
2. **`DateTime.UtcNow` rather than a monotonic `Stopwatch`** for the deadline — carried over verbatim from the originals, so not introduced here, but a `Stopwatch` would be marginally more correct against clock adjustments.

### Optional follow-up (out of scope for this PR / issue)
The suite still has ~a dozen other `while (DateTime.UtcNow < deadline) { … Task.Delay }` polling loops (e.g. `LoginControllerTests`, `SocketManagerServiceTests`, `SessionCacheEvictionTests`, `PlayerCacheEvictionTests`, `BackgroundWorkerTests`). They poll different subjects (socket/cache/worker state, not Redis fire-and-forget) and were rightly outside #1723's scope, but now that a genuinely generic `PollUntilAsync` exists, a tech-debt follow-up could migrate the mechanical ones for consistency.

Ship it.

_Automated review by Claude (Opus 4.8). Not merged._

---

ginderjeremiah:
Filed the follow-up to migrate the remaining polling loops as #1789. Nothing blocking in this PR — merging.

---

Issue #1708, opened 2026-07-07
Admin cache-reload amplification: every admin endpoint call rebuilds all reference sets, and one Workbench save spans several calls


ginderjeremiah:
## Problem

`AdminCacheReloadFilter` awaits `ReloadAsync` on **every** registered `IReloadableReferenceCache` (~11 sets, each a full DB re-query + snapshot rebuild) plus a cross-instance broadcast, per successful admin write (`Game.Api/Filters/AdminCacheReloadFilter.cs:70-110`). Because the Workbench persists a record as identity-save + one call per relationship setter (documented in `docs/backend-admin.md`), a single "Save" in the UI multiplies into N × (all-sets reload + broadcast + version-hash memo invalidation for every set).

This is admin-only and deliberately simple, so it is **not a defect** — filing it as the known optimization target in this area, gated on Workbench saves actually feeling slow as the catalog grows.

## Cheap levers when picked up

- Reload only the affected set(s) per controller (the filter could take the set from the attribute).
- And/or debounce the awaited reload across the endpoint calls of one logical Workbench save.

Constraint: preserve read-your-writes for the Workbench (the awaited local reload after the last write) and the cross-instance invalidation broadcast.

_Found during a codebase health review._

PR #1755, opened 2026-07-09
Dedupe LoginTrackingMiddleware's DB upsert per (user, IP, device)

### LINKED ISSUE #1727: LoginTrackingMiddleware runs an inline multi-query DB upsert on every authenticated HTTP request
## Problem

`Game.Api/Middleware/LoginTrackingMiddleware.cs:41` awaits `RecordConnection` **before** `_next`, and the implementation (`Game.DataAccess/Repositories/UserLogins.cs:23-72`) runs 2–3 SELECTs plus a `SaveChanges` per request. The frontend attaches the device-fingerprint header on every authenticated request, so:

- the login flow (Login → Players → SelectPlayer → ActiveSession → Status → DeviceInfo) pays this serially on each hop, and
- every admin Workbench call pays it too — on top of the awaited cache reload each admin save already pays (#1708).

## Suggested fix

Either dedupe per `(user, ip, fingerprint)` with a short in-memory TTL memo (the connection row is idempotent — repeat hits add no information), or run the recording after the response completes; the isolation scope already exists. Preserve the current behavior of recording a genuinely new device/IP promptly.

ginderjeremiah:
## Summary

`LoginTrackingMiddleware` awaits `RecordConnection` (2-3 SELECTs + a `SaveChanges`, `Game.DataAccess/Repositories/UserLogins.cs:23-72`) **before** `_next(context)` on every authenticated HTTP request. Since the frontend attaches the device-fingerprint header on every authenticated request, this serially delays:

- each hop of the login flow (Login → Players → SelectPlayer → ActiveSession → Status → DeviceInfo), and
- every admin Workbench call, on top of the awaited cache reload each admin save already pays (#1708).

## Fix

The connection row is idempotent — repeat hits from the same `(user, IP, device)` add no information — so the middleware now dedupes with a short-lived in-memory memo (`IMemoryCache`, 5-minute window, instance-local):

- A cache miss runs the DB upsert as before, and only sets the memo **after** a successful save — so a failed attempt is never memoized as done and is retried on the very next request.
- A cache hit skips the DB round-trip entirely.
- A genuinely new `(user, IP, device)` combination always misses the memo on its first request, so it is still recorded promptly.

Instance-local scope is fine here: per `docs/backend.md`, a player session sticks to one server instance for its lifetime, so the memo doesn't need to be shared across instances (Redis) — worst case on an instance switch is one extra DB write, not a correctness issue.

## Changes

- `Game.Api/Middleware/LoginTrackingMiddleware.cs`: adds the `IMemoryCache`-backed dedupe memo, keyed by `(UserId, IpAddress, Fingerprint)`.
- `Game.Api/Startup.cs`: registers `IMemoryCache` via `AddMemoryCache()`.
- `Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs` (new): covers the skip conditions (unauthenticated, no fingerprint header), that a new combination records and calls `next`, that a repeat request within the window skips the DB call, that a different IP/device/user each records separately, and that a save failure is swallowed but not memoized (so it's retried).

## Test plan

- [x] `dotnet test` — full solution suite passes: 3035/3035 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), including the existing `LoginTrackingIsolationTests` integration test (still passes unmodified — that test uses a fresh device/IP per run, so it isn't affected by the dedupe memo) and the new middleware unit tests.

Closes #1727

---
_Generated by [Claude Code](https://claude.ai/code/session_01WniWr4qbSdrsXWubjza3Zc)_

### DIFF
--- a/Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
+++ b/Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
@@ -1,0 +1,220 @@
+using Game.Abstractions.DataAccess;
+using Game.Api.Http;
+using Game.Api.Middleware;
+using Game.Api.Services;
+using Game.Application.Services;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging.Abstractions;
+using System.Net;
+using Xunit;
+
+namespace Game.Api.Tests.Unit
+{
+    /// <summary>
+    /// Covers <see cref="LoginTrackingMiddleware"/>'s request-skip conditions and its per-(user, IP, device)
+    /// dedupe memo (#1727): a genuinely new combination is always recorded, but repeat requests from the same
+    /// combination within the dedupe window skip the multi-query DB upsert entirely, and a failed attempt is
+    /// never memoized as done so it is retried on the next request.
+    /// </summary>
+    public class LoginTrackingMiddlewareTests
+    {
+        private const string Fingerprint = "fp-abc";
+        private const string UserAgent = "TestAgent/1.0";
+        private const string Ip = "203.0.113.7";
+
+        [Fact]
+        public async Task Unauthenticated_SkipsTrackingAndCallsNext()
+        {
+            var (middleware, userLogins, nextCalled) = CreateMiddleware();
+            var session = new SessionService(new NoOpSessionStore());
+            var context = CreateContext(Ip, Fingerprint, UserAgent);
+
+            await middleware.InvokeAsync(context, session, ScopeFactory(userLogins), new MemoryCache(new MemoryCacheOptions()));
+
+            Assert.Equal(0, userLogins.RecordConnectionCallCount);
+            Assert.True(nextCalled());
+        }
+
+        [Fact]
+        public async Task Authenticated_NoFingerprintHeader_SkipsTrackingAndCallsNext()
+        {
+            var (middleware, userLogins, nextCalled) = CreateMiddleware();
+            var session = AuthenticatedSession(userId: 5);
+            var context = CreateContext(Ip, fingerprint: null, UserAgent);
+
+            await middleware.InvokeAsync(context, session, ScopeFactory(userLogins), new MemoryCache(new MemoryCacheOptions()));
+
+            Assert.Equal(0, userLogins.RecordConnectionCallCount);
+            Assert.True(nextCalled());
+        }
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Dedupe LoginTrackingMiddleware's DB upsert per (user, IP, device) (closes #1727)

**Verdict:** Approve — the change matches the issue's suggested fix exactly, preserves the middleware's isolation and failure semantics, and is thoroughly tested. (Submitted as a `COMMENT` because GitHub won't accept an `APPROVE` from the PR author; the real verdict is approve.)

### Why this is correct
- **Idempotency holds.** `UserLogins.RecordConnection` only refreshes a `LastConnection` audit/analytics timestamp (get-or-create by unique index). The single-active-session presence check is a separate Redis TTL (`docs/backend.md` → *Single active connection*), so a 5-minute dedupe window doesn't affect session liveness — only bounds the audit timestamp's staleness to ≤5 min under continuous activity, which the issue owner explicitly endorsed.
- **Isolation preserved.** The dedicated DI scope / self-commit path is untouched; the memo just gates entry to it.
- **Failure semantics are right and tested.** `cache.Set` runs only *after* a successful `RecordConnection`, so a swallowed save failure is never memoized and retries on the next request. The `LoginTrackingIsolationTests` integration test still holds (its fault double always throws, so it never memoizes).
- **Cache-key correctness.** `TrackingCacheKey` is a `record struct` — value equality/hash survive the boxing `IMemoryCache` does, so equal tuples collapse to one entry. Growth is bounded by distinct `(user, IP, device)` tuples per 5-min window and self-expires.
- **Concurrent first-hit** from the same new tuple can double-miss and both reach the DB, but `RecordConnection`'s unique-index + retry makes that harmless — the memo is a best-effort optimization, not a lock.
- `ClientIp.Resolve` now runs on every authenticated request before the cache check, but the fuller `ClientHints.FromHeaders` still runs only on a miss — the right split.

### Test coverage
Comprehensive and aligned with the project's testing rules (classical style, DB dependency faked since it's out-of-process): skip-when-unauthenticated, skip-when-no-fingerprint, new-tuple records + calls next, repeat-within-window skips the DB round-trip, different IP/device/user each record separately, and the failure-not-memoized-so-retried path. Registration wired via `AddMemoryCache()` in `Startup.cs`.

### What's good
Clean, well-scoped fix with clear rationale documented right where it lives (the XML doc on the middleware and the `Startup.cs` comment explaining why instance-local is sufficient). Nicely done.

Holding the merge only until `test-backend` finishes (all other checks are green/skipped).

---

PR #1756, opened 2026-07-09
Route server-push socket commands through the shared DeserializeParameters helper

### LINKED ISSUE #1728: Server-push socket commands bypass the shared DeserializeParameters helper (wrong fault classification + triplicated boilerplate)
## Problem

The three server-push commands override `SetParameters` with raw `Deserialize` + `ArgumentNullException` instead of the base `DeserializeParameters&lt;T&gt;` helper (`Game.Api/Sockets/AbstractSocketCommand.cs:25`):

- `Game.Api/Sockets/Commands/ChallengeCompleted.cs:22`
- `Game.Api/Sockets/Commands/ProficiencyXpGained.cs:22`
- `Game.Api/Sockets/Commands/ServerCommandFailed.cs:20`

A malformed payload therefore classifies as `Faulted` rather than `MalformedParameters`, contradicting the comment at `Game.Api/Sockets/SocketHandler.cs:293-296` which claims every `SetParameters` implementation throws `MalformedSocketCommandParametersException`. No behavioral difference today (both outcomes escalate identically for server pushes, and clients can't invoke these commands), but it's triplicated boilerplate plus a null-recheck in each `HandleExecuteAsync` that the helper eliminates.

## Fix

Route the three overrides through `DeserializeParameters&lt;T&gt;`, drop the per-command null-rechecks, and the `SocketHandler` comment becomes true again.

ginderjeremiah:
## Summary

`ChallengeCompleted`, `ProficiencyXpGained`, and `ServerCommandFailed` each overrode `SetParameters` with a raw `Deserialize<T>` call plus a manual `ArgumentNullException`, instead of going through the shared `DeserializeParameters<T>` helper on `AbstractSocketCommand` (`Game.Api/Sockets/Commands/AbstractSocketCommand.cs:25`). That meant a malformed payload on these three commands classified as `Faulted` rather than `MalformedParameters`, contradicting the `SocketHandler` comment claiming every `SetParameters` implementation throws `MalformedSocketCommandParametersException` — plus each command carried its own null-recheck in `HandleExecuteAsync` that the helper already makes unnecessary.

## Changes

All three server-push commands echo their payload straight back as response data (same model in and out), so they're a natural fit for the existing `AbstractSocketCommand<TReturn, TParams>` base (used elsewhere for typed request/response commands) with `TReturn == TParams`:

- `ChallengeCompleted`, `ProficiencyXpGained`, `ServerCommandFailed`: now extend `AbstractSocketCommand<TModel, TModel>` instead of `AbstractSocketCommandWithResponseData<TModel>`. This drops the bespoke `SetParameters` override, the backing nullable field, and the `HandleExecuteAsync` null-recheck in each — `Parameters` is guaranteed bound before `ExecuteAsync` runs (`SocketCommandFactory.CreateCommand` always calls `SetParameters` right after construction).
- A malformed/missing payload on any of the three now throws `MalformedSocketCommandParametersException` (wrapping the original `JsonException`/`ArgumentNullException`) like every other command, so the `SocketHandler` comment is accurate again.

## Tests

- `Game.Api.Tests/Unit/ProficiencyXpGainedCommandTests.cs`: updated the null-payload test to assert `MalformedSocketCommandParametersException` (wrapping `ArgumentNullException`) instead of a bare `ArgumentNullException`. Removed `Execute_WithoutParameters_Throws` since the scenario it covered (executing before parameters are bound) no longer exists once the null-recheck is removed — no other command in the codebase tests that path either, since the factory always binds parameters first.
- No behavior change for the happy path or for existing integration tests (`ChallengeCompletedSocketTests`, `ServerInitiatedCommandRejectionTests`) — both still pass unchanged.

## Test plan

- [x] `dotnet test --no-build --no-restore` — full solution suite passes: 3034/3034.

Closes #1728

---
_Generated by [Claude Code](https://claude.ai/code/session_017L5P7BEMmS9JA6XuErsbMf)_

### DIFF
--- a/Game.Api/Sockets/Commands/ChallengeCompleted.cs
+++ b/Game.Api/Sockets/Commands/ChallengeCompleted.cs
@@ -8,30 +8,15 @@ namespace Game.Api.Sockets.Commands
     /// A server-initiated push (never sent by the client): the instance holding the player's socket
     /// receives the emitted <see cref="ChallengeCompletedInfo"/> and forwards its payload to the client.
     /// The payload arrives as the command parameters and is echoed straight back as the response data, so
-    /// the client's <c>ChallengeCompleted</c> listener can unlock the rewards locally. It carries no
-    /// request type (the parameters are not a typed <c>Parameters</c> property) since the client only
-    /// listens for it.
+    /// the client's <c>ChallengeCompleted</c> listener can unlock the rewards locally.
     /// </summary>
-    public class ChallengeCompleted : AbstractSocketCommandWithResponseData<ChallengeCompletedModel>, IServerInitiatedCommand
+    public class ChallengeCompleted : AbstractSocketCommand<ChallengeCompletedModel, ChallengeCompletedModel>, IServerInitiatedCommand
     {
-        private ChallengeCompletedModel? _model;
-
         public override string Name { get; set; } = nameof(ChallengeCompleted);
-
-        public override void SetParameters(string? parameters)
-        {
-            _model = parameters.Deserialize<ChallengeCompletedModel>()
-                ?? throw new ArgumentNullException(nameof(parameters));
-        }
 
         public override Task<ApiSocketResponse<ChallengeCompletedModel>> HandleExecuteAsync(SocketContext context, CancellationToken cancellationToken)
         {
-            if (_model is null)
-            {
-                throw new InvalidOperationException("ChallengeCompleted executed without parameters.");
-            }
-
-            return Task.FromResult(Success(_model));
+            return Task.FromResult(Success(Parameters));
         }
     }
 
--- a/Game.Api.Tests/CodeGen/SocketCommandMetadataTests.cs
+++ b/Game.Api.Tests/CodeGen/SocketCommandMetadataTests.cs
@@ -75,6 +75,18 @@ namespace Game.Api.Tests.CodeGen
             Assert.Null(metadata.ResponseDescriptor);
             Assert.Null(metadata.ParameterDescriptor);
         }
+
+        [Fact]
+        public void Constructor_ServerInitiatedCommand_OmitsParameterDescriptor()
+        {
+            // A server-initiated command may still bind Parameters from a typed base (to reuse
+            // DeserializeParameters<T>'s malformed-payload classification), but the client only ever listens
+            // for it and can never send it, so it must not surface in ApiSocketRequestTypes.
+            var metadata = new SocketCommandMetadata(typeof(TestSocketCommandServerInitiatedFull));
+
+            Assert.NotNull(metadata.ResponseDescriptor);
+            Assert.Null(metadata.ParameterDescriptor);
+        }
     }
 
     public class SocketParamModel
@@ -147,6 +159,16 @@ namespace Game.Api.Tests.CodeGen
         public override Task<ApiSocketResponse> ExecuteAsync(SocketContext context, CancellationToken cancellationToken)
         {
             return Task.FromResult(Success());
+        }
+    }
+
+    public class TestSocketCommandServerInitiatedFull : AbstractSocketCommand<SimpleModel, SocketParamModel>, IServerInitiatedCommand
+    {
+        public override string Name { get; set; } = "TestServerInitiatedFull";
+
+        public override Task<ApiSocketResponse<SimpleModel>> HandleExecuteAsync(SocketContext context, CancellationToken cancellationToken)
+        {
+            return Task.FromResult(Success(new SimpleModel()));
         }
     }
 }
--- a/Game.Api.Tests/Unit/ProficiencyXpGainedCommandTests.cs
+++ b/Game.Api.Tests/Unit/ProficiencyXpGainedCommandTests.cs
@@ -6,12 +6,18 @@ namespace Game.Api.Tests.Unit
 {
     /// <summary>
     /// Unit coverage for the server-initiated <see cref="ProficiencyXpGained"/> push command: it echoes the
-    /// emitted payload straight back as its response data (the client listens for it), and rejects a missing
-    /// payload on both the set-parameters and execute paths. The handler ignores the socket context, so these
-    /// run as plain unit tests without a live socket — mirroring the other echo push commands.
+    /// emitted payload straight back as its response data (the client listens for it), and classifies a missing
+    /// payload as <see cref="MalformedSocketCommandParametersException"/> like every other command's
+    /// <c>SetParameters</c>. The handler ignores the socket context, so these run as plain unit tests without a
+    /// live socket — mirroring the other echo push commands.
     /// </summary>
     public class ProficiencyXpGainedCommandTests
     {
+        private static ProficiencyXpGained NewCommand()
+        {
+            return new ProficiencyXpGained { Parameters = new ProficiencyXpGainedModel { Proficiencies = [], Opened = [] } };
+        }
+
         [Fact]
         public async Task SetParameters_ThenExecute_EchoesThePayloadBack()
         {
@@ -32,7 +38,7 @@ namespace Game.Api.Tests.Unit
                 Opened = [new ProficiencyOpenedModel { ProficiencyId = 4 }],
             };
 
-            var command = new ProficiencyXpGained();
+            var command = NewCommand();
             // The emitted info carries the serialized payload exactly as the notifier would send it.
             command.SetParameters(new ProficiencyXpGainedInfo(model).Parameters);
 
@@ -51,18 +57,13 @@ namespace Game.Api.Tests.Unit
         }
 
         [Fact]
-        public void SetParameters_NullPayload_Throws()
+        public void SetParameters_NullPayload_ThrowsMalformedSocketCommandParametersException()
         {
-            var command = new ProficiencyXpGained();
-            Assert.Throws<ArgumentNullException>(() => command.SetParameters(null));
-        }
+            var command = NewCommand();
 
-        [Fact]
-        public async Task Execute_WithoutParameters_Throws()
-        {
-            var command = new ProficiencyXpGained();
-            await Assert.ThrowsAsync<InvalidOperationException>(
-                () => command.HandleExecuteAsync(context: null!, CancellationToken.None));
+            var ex = Assert.Throws<MalformedSocketCommandParametersException>(() => command.SetParameters(null));
+
+            Assert.IsType<ArgumentNullException>(ex.InnerException);
         }
     }
 }
--- a/Game.Api/CodeGen/Data/SocketCommandMetadata.cs
+++ b/Game.Api/CodeGen/Data/SocketCommandMetadata.cs
@@ -30,8 +30,14 @@ namespace Game.Api.CodeGen.Data
                 }
             }
 
-            if (socketCommand.GetClosedGenericBase(typeof(AbstractSocketCommandWithParams<>)) is not null
-                || socketCommand.GetClosedGenericBase(typeof(AbstractSocketCommand<,>)) is not null)
+            // A server-initiated command (IServerInitiatedCommand) may bind its Parameters from a typed base
+            // purely to reuse DeserializeParameters<T>'s malformed-payload classification, but the client only
+            // ever listens for it and can never send it (SocketCommandFactory.IsServerInitiatedOnly rejects an
+            // inbound attempt). It must never surface in ApiSocketRequestTypes/ApiSocketCommandWithRequest,
+            // which the frontend uses to type its client->server sends.
+            if (!socketCommand.IsAssignableTo(typeof(IServerInitiatedCommand))
+                && (socketCommand.GetClosedGenericBase(typeof(AbstractSocketCommandWithParams<>)) is not null
+                    || socketCommand.GetClosedGenericBase(typeof(AbstractSocketCommand<,>)) is not null))
             {
                 var property = socketCommand.GetProperty(nameof(AbstractSocketCommandWithParams<>.Parameters));
                 if (property is not null)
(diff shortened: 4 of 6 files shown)


Game.Api/Sockets/Commands/ChallengeCompleted.cs
```diff
@@ -8,30 +8,15 @@ namespace Game.Api.Sockets.Commands
     /// A server-initiated push (never sent by the client): the instance holding the player's socket
     /// receives the emitted <see cref="ChallengeCompletedInfo"/> and forwards its payload to the client.
     /// The payload arrives as the command parameters and is echoed straight back as the response data, so
-    /// the client's <c>ChallengeCompleted</c> listener can unlock the rewards locally. It carries no
-    /// request type (the parameters are not a typed <c>Parameters</c> property) since the client only
-    /// listens for it.
+    /// the client's <c>ChallengeCompleted</c> listener can unlock the rewards locally.
     /// </summary>
-    public class ChallengeCompleted : AbstractSocketCommandWithResponseData<ChallengeCompletedModel>, IServerInitiatedCommand
+    public class ChallengeCompleted : AbstractSocketCommand<ChallengeCompletedModel, ChallengeCompletedModel>, IServerInitiatedCommand
```
ginderjeremiah, line 13: **Blocking** — switching the base to `AbstractSocketCommand<TReturn, TParams>` changes the generated client contract, not just the deserialize path. `SocketCommandMetadata` populates a `ParameterDescriptor` for anything whose base chain includes `AbstractSocketCommand<,>` (`Game.Api/CodeGen/Data/SocketCommandMetadata.cs:33`), and `SocketMapWriter` emits every command with a non-null `ParameterDescriptor` into `ApiSocketRequestTypes` / `ApiSocketCommandWithRequest` (`SocketMapWriter.cs:44`). These three were previously excluded (they extended `AbstractSocketCommandWithResponseData<T>` directly), so:

1. **CI will fail.** The committed `UI/src/lib/api/types/api-socket-type-map.ts` doesn't list `ChallengeCompleted`/`ProficiencyXpGained`/`ServerCommandFailed` under `ApiSocketRequestTypes`; after this change codegen would add them, so the "Verify TypeScript codegen is up to date" step (`run-tests.yml:100`, `git diff --cached --exit-code`) breaks. The generated file wasn't regenerated here.
2. **Semantic regression.** Regenerating is the *wrong* fix: it would move these server-push-only commands into `ApiSocketCommandWithRequest`, which is exactly the union the frontend's `sendSocketCommand`/`fetchSocketData` use to type client→server sends (`UI/src/lib/api/api-socket.ts:148`). That leaks server-initiated commands into the client's send surface — the precise thing the comments you removed ("carries no request type ... since the client only listens for it") were preserving.

The same applies identically to `ProficiencyXpGained.cs` and `ServerCommandFailed.cs`.

Two ways forward (see summary): keep them on `AbstractSocketCommandWithResponseData<T>` and just route their existing `SetParameters` override through `DeserializeParameters<T>` (minimal, matches issue #1728's ask), or fix the root cause by excluding `IServerInitiatedCommand` from the request map in codegen and regenerate.

---

ginderjeremiah (commented):
## Review: Route server-push socket commands through the shared `DeserializeParameters` helper (#1728)

**Verdict:** Request changes — the refactor is directionally right, but the chosen base-class switch has an unintended side effect on the generated client contract that will fail CI and regress the client's send-command type surface. _(Submitted as a COMMENT because I'm the PR author and GitHub won't accept a gating verdict on one's own PR; treat this as request-changes.)_

The goal is good and the fault-classification reasoning is sound — the boilerplate and the `SocketHandler` comment drift are real. The problem is only in *how* the boilerplate was removed.

### Blocking

- **The base-class change alters codegen output; the generated file is stale and CI will fail.** Moving `ChallengeCompleted`, `ProficiencyXpGained`, and `ServerCommandFailed` from `AbstractSocketCommandWithResponseData<T>` to `AbstractSocketCommand<TReturn, TParams>` gives them a `Parameters` property, which `SocketCommandMetadata` (`Game.Api/CodeGen/Data/SocketCommandMetadata.cs:33`) turns into a `ParameterDescriptor`, which `SocketMapWriter` (`Game.Api/CodeGen/Writers/SocketMapWriter.cs:44`) emits into `ApiSocketRequestTypes`. The committed `UI/src/lib/api/types/api-socket-type-map.ts` does not list these three under `ApiSocketRequestTypes`, so the **"Verify TypeScript codegen is up to date"** step (`.github/workflows/run-tests.yml:100`, `git diff --cached --exit-code`) will break. The PR didn't regenerate the codegen output.
- **Semantic regression — server-push-only commands leak into the client send surface.** Simply regenerating is the wrong fix. `ApiSocketRequestTypes` / `ApiSocketCommandWithRequest` is the union the frontend uses to type client→server sends (`sendSocketCommand`'s `WithRequest` overload at `UI/src/lib/api/api-socket.ts:148`, and `api-socket-request.ts`). These three are `IServerInitiatedCommand` — the client only ever *listens* for them. Adding them to the request map contradicts their design intent, captured verbatim in the comments this PR deletes ("It carries no request type ... since the client only listens for it"). Note the codegen has no `IServerInitiatedCommand` filter today (`ApiCodeGenerator.cs:29-30`), so nothing else prevents this leak.

### Suggested fix (author's call)

Either:
1. **Minimal, matches the issue's ask** — keep all three on `AbstractSocketCommandWithResponseData<T>` and just route their existing `SetParameters` override through `DeserializeParameters<T>` (issue #1728 literally says "Route the three overrides through `DeserializeParameters<T>`"). This fixes fault classification and drops the raw `Deserialize` + bespoke `ArgumentNullException` with zero codegen impact. You'd keep a backing field + a null-check in `HandleExecuteAsync` (less boilerplate removed, but no contract change), or
2. **Root-cause fix** — exclude `IServerInitiatedCommand` types from the request map in the codegen (they shouldn't ever appear as client requests), keep the cleaner `AbstractSocketCommand<,>` base, and **regenerate** `api-socket-type-map.ts`. This is the better long-term shape since it also protects any future server-push command with a payload. If you take this route, it's worth a one-line follow-up note in `docs/backend.md` (or a code comment near the codegen filter) so the exclusion isn't "simplified" away later.

My lean is option 2 — it addresses the actual defect (codegen treating a server-push payload as a client request) rather than working around it — but option 1 is a legitimate minimal-scope choice.

### Test coverage

- The `ProficiencyXpGainedCommandTests` changes are correct and well-reasoned: asserting the wrapped `MalformedSocketCommandParametersException` (inner `ArgumentNullException`), and removing `Execute_WithoutParameters_Throws` is justified once `Parameters` is `required` and always bound by the factory. No concern here.
- Whichever fix you land on, re-verify the null-payload path still classifies as `MalformedSocketCommandParametersException` (it does under both options).

### What's good

- Correct diagnosis of the fault-misclassification and the stale `SocketHandler` comment.
- The happy-path echo behavior and the `Success(Parameters)` simplification read cleanly.
- Solid, honest test reasoning in the PR description.

---

ginderjeremiah:
Confirmed both blocking points — verified the codegen path (`SocketCommandMetadata.cs:33`/`SocketMapWriter.cs:44`) and the CI job log directly; the diff it tried to write matched exactly what was flagged.

Went with option 2 (root-cause): `SocketCommandMetadata` now excludes `ParameterDescriptor` extraction for any `IServerInitiatedCommand` type, with a comment at the guard explaining why (so it isn't "simplified" away later). Regenerated `api-socket-type-map.ts` — no diff, confirming these three were never in `ApiSocketRequestTypes` and still aren't. Added a unit test (`SocketCommandMetadataTests.Constructor_ServerInitiatedCommand_OmitsParameterDescriptor`) pinning the exclusion for a command that has both a response and parameter generic base.

Full suite still green (3035/3035) and `dotnet build -warnaserror` clean. Pushed in 1f9f129.

---

ginderjeremiah:
## Automated review — approve with nits

**Verdict:** approve-with-nits (one pre-merge gate to clear on CI, no code changes required)

**Scope:** Routes the three server-push commands (`ChallengeCompleted`, `ProficiencyXpGained`, `ServerCommandFailed`) through the shared `DeserializeParameters<T>` helper by switching them from `AbstractSocketCommandWithResponseData<T>` to `AbstractSocketCommand<T, T>`, dropping the bespoke `SetParameters` overrides, backing fields, and `HandleExecuteAsync` null-rechecks. Matches issue #1728 exactly.

### Correctness — verified
- A malformed/missing payload on all three now throws `MalformedSocketCommandParametersException` (wrapping the `JsonException`/`ArgumentNullException`), so it classifies as `MalformedParameters` rather than `Faulted`, and the `SocketHandler` comment (~L293) is true again.
- Construction path is sound: `SocketCommandFactory` builds via `Expression.New` (bypasses `required`-member enforcement at runtime) and always calls `SetParameters` before `ExecuteAsync`, mirroring the existing `AbstractSocketCommandWithParams<T>` pattern — the `required Parameters` property is never observed unbound in the live path.
- **Good catch in commit 2:** giving these commands a `Parameters` property would otherwise have leaked them into the generated `ApiSocketRequestTypes` (the client→server send union). The `IServerInitiatedCommand` guard added to `SocketCommandMetadata` suppresses `ParameterDescriptor` while keeping `ResponseDescriptor`, so the generated client is unchanged. This is the load-bearing part of the change and it's handled correctly, with a dedicated unit test.

### Tests
- Reasonable and proportionate. The null-payload test now asserts the wrapped exception; `Execute_WithoutParameters_Throws` was correctly removed (that state is unreachable once the null-recheck is gone). The metadata guard has direct coverage. The malformed-classification behavior now lives in the shared base, so testing it once via `ProficiencyXpGained` is sufficient — no need to triplicate it.

### Findings by severity
- **Should-fix (gate, not code) — CI never validated this commit.** The latest `Run tests` run on head `1f9f129` concluded **failure**: `detect-changes` was *cancelled* (~15 min, infra), which **skipped** every `test-*`/`verify-*` suite, including the codegen-drift check that would confirm the regenerated client. The `all-checks-passed` gate went green only because skipped ≠ failed. Re-run CI to a real green (tests + codegen drift actually executing) before merge — the author's local `3034/3034` is reassuring but unverified here.
- **Nit (out of scope):** the `all-checks-passed` gate treating suites skipped-due-to-a-cancelled-`detect-changes` as success is a CI weakness worth a separate hardening follow-up.

No blocking code issues. Clean, well-scoped, DRY, and it removes the triplicated boilerplate the issue targeted.

—
_Automated review by Claude (Opus 4.8). Not a merge approval; CI must be re-run green first._

---

ginderjeremiah:
Re-triggered the failed jobs on run #2195 so the test suites + codegen-drift guard actually execute against the head SHA (they were skipped last time because `detect-changes` was cancelled, not because the code failed). Will merge once CI is genuinely green. Filed the CI-hardening gap (skipped-because-cancelled masquerading as green) as #1790.

---

ginderjeremiah:
## Review — approve (merging)

Independent verification against the real code. Verdict: **approve**. CI is now genuinely green on head `1f9f129` and #1728 is addressed exactly.

### Verified against source
- **Payload echo is behaviorally identical.** `AbstractSocketCommand<TReturn, TParams>` derives from `AbstractSocketCommandWithResponseData<TReturn>` and adds `required TParams Parameters` bound via `DeserializeParameters<TParams>`. With `TReturn == TParams`, `HandleExecuteAsync` returning `Success(Parameters)` produces the same `ApiSocketResponse<T>` the old `Success(_model)` did — request/response is the same model in and out, no regression.
- **Removed null-rechecks are safe.** `SocketCommandFactory.CreateCommand` builds the command via `Expression.New` (which bypasses `required`-member enforcement at runtime) and always calls `SetParameters` immediately after construction, before the caller (`SocketHandler.RunCommand`) invokes `ExecuteAsync`. `Parameters` is therefore never observed unbound on the live path, and the old `InvalidOperationException` guard was dead once `SetParameters` is guaranteed. So dropping `Execute_WithoutParameters_Throws` drops zero meaningful coverage — it pinned an unreachable state, and no other command tests that path.
- **Fault classification fixed.** A malformed/missing payload on all three now throws `MalformedSocketCommandParametersException` (wrapping the `JsonException`/`ArgumentNullException`) via the shared helper, so it classifies as `MalformedParameters` not `Faulted`, and the `SocketHandler` comment (~L293) is true again.
- **The load-bearing companion change is correct.** Giving these commands a `Parameters` property would otherwise have made `SocketCommandMetadata` emit a `ParameterDescriptor`, leaking them into the generated `ApiSocketRequestTypes` (the client→server send union) — but the client can never send them. The `!IsAssignableTo(IServerInitiatedCommand)` guard suppresses `ParameterDescriptor` while keeping `ResponseDescriptor`, so the generated client is unchanged. Well-commented and pinned by the new `Constructor_ServerInitiatedCommand_OmitsParameterDescriptor` test.
- **Consistency.** The only other `IServerInitiatedCommand` with a distinct pattern is `SocketReplaced`, which is parameterless (bare `AbstractSocketCommand`) and correctly untouched. No other command was left on the old boilerplate.

### Quality
Clean, DRY (removes triplicated `SetParameters` + backing field + null-recheck), correct nullability (no `!`), no dead code left behind, `{}` blocks respected. Tests are proportionate — the malformed-classification behavior now lives in the shared base, so exercising it once via `ProficiencyXpGained` is sufficient.

The earlier "CI never validated this commit" gate is now cleared: the re-run of the workflow on `1f9f129` executed `detect-changes` → `test-backend`/`verify-backend`/`test-e2e` all green (frontend suites correctly skipped, backend-only diff). The CI-hardening gap (skipped-because-cancelled masquerading as green) is already tracked in #1790, so no new follow-up is needed.

No blocking issues and no code changes required. Merging via squash.

—
_Automated review by Claude (Opus 4.8)._

---

PR #1793, opened 2026-07-09
Bound SocketContext.SendData with an abort fallback

### LINKED ISSUE #1760: A wedged send with live inbound traffic wedges the connection (and its Player) permanently — gap left by #1741
## Problem

#1741 bounded a wedged send inside `Close()` with an `Abort()` fallback, but the ordinary response send path has no bound and the watchdog can't catch it:

- `Game.Api/Sockets/SocketContext.cs:98-113` — chunk sends deliberately run under `CancellationToken.None`; a peer with a zero TCP receive window blocks `SendAsync` indefinitely while holding the send lock (the `finally` releasing it never runs).
- `Game.Api/Sockets/SocketHandler.cs:226` — the response send (`await _context.SendData(result.Response)`) is awaited **after** the per-command `WaitAsync` budget, unbounded, while holding the per-socket command lock.
- The inactivity watchdog (`SocketHandler.cs:341`) keys only off **inbound** activity — a client that keeps sending pings while never reading keeps the presence key fresh and the watchdog dormant forever.

## Failure scenario

A tampered client opens a socket, stalls its read side, keeps heartbeating. The first response wedges the send lock; the command lock is never released; the handler, DI scope, and in-memory `Player` are pinned until process shutdown. Socket takeover doesn't reap the old wedged handler (its `SocketReplaced` push queues behind the held command lock), so repeated handshakes accumulate wedged handlers — a memory-pressure vector from one account. (An honest suspended tab stops pinging too and is covered by the existing watchdog, hence this needs a deliberately stalled-reading peer.)

## Fix shape

Track outbound-send health in the watchdog (e.g. also close/abort when a send has been in flight past a bound), or bound the post-command `SendData` and fall back to `Abort()` the same way `Close` now does.

ginderjeremiah:
## Summary

`SocketContext.Close` already self-bounds and falls back to `WebSocket.Abort()` (#1741) when the send lock is wedged, but the ordinary response send path (`SocketContext.SendData`) had no such bound, and the inactivity watchdog can't catch it either since it only tracks inbound activity.

A peer that stalls its read side (a zero TCP receive window) while continuing to send inbound traffic (heartbeats) wedges the very first response send. That send holds `SocketContext`'s send lock forever — and, worse, `SocketHandler.RunCommandUnderLock` awaits that send *after* its own per-command timeout budget has already elapsed, still holding the per-socket command lock. So a single wedged send permanently pins the command lock, the handler, its DI scope, and the in-memory `Player` until process shutdown — a memory-pressure vector from one tampered client, since socket takeover can't reap the wedged handler either (its `SocketReplaced` push queues behind the same held command lock).

## Changes

- `Game.Api/Sockets/SocketContext.cs`: `SendData` now bounds the actual chunk send with a new `_sendAbortTimeout` (default 5s, mirroring `_closeAbortTimeout`'s default and configurable via a new optional constructor parameter). On timeout it logs a warning and calls `Abort()` — the same fallback `Close()` already uses — and hands the send lock's release off to a continuation once the abandoned send actually settles (mirroring `SocketHandler.ReleaseCommandLockWhenSettled`'s pattern), so a still in-flight frame can't be interleaved into by the next writer.
- `Game.TestInfrastructure/Helpers/FakeWebSocket.cs`: `Abort()` now actually unblocks a gated send (faulting it), matching real `WebSocket` abort semantics, so the fix has direct end-to-end test coverage through `SocketHandler` rather than only through a bespoke test-only socket.
- `docs/backend.md`: extends the existing `Close`-self-bounding design note to cover `SendData`'s matching fallback.

## Tests

- `Game.Api.Tests/Unit/SocketContextTests.cs`: new `SendData_WedgedSend_AbortsInsteadOfHangingForever` pins that a wedged send (reusing the existing `WedgeableSendWebSocket` fake) times out and aborts rather than hanging forever.
- `Game.Api.Tests/Unit/SocketCommandExecutionTests.cs`: new `ExecuteCommand_ResponseSendWedges_AbortsAndReleasesTheCommandLockForTheNextCommand` drives the real `SocketHandler` → `SocketContext` path end-to-end and proves a second command can run after the first's response send wedges — the actual permanent-pin scenario the issue describes.

## Test plan

- [x] `dotnet test` — full solution suite passes: 3038/3038 (Game.Core.Tests, Game.Application.Tests, Game.Api.Tests, Game.Infrastructure.Tests).

Closes #1760

---
_Generated by [Claude Code](https://claude.ai/code/session_011tBEtgkj416fXuWYbkaCnG)_


ginderjeremiah (commented):
## Review: bound `SocketContext.SendData` with an `Abort()` fallback (closes #1760)

**Verdict:** Approve — closes the exact gap #1741 left on the ordinary response-send path; the fix is correct, faithfully mirrors the established `Close()` / `ReleaseCommandLockWhenSettled` patterns, is well-documented, and is covered by both a unit and a faithful end-to-end test.

_(Posted as a `COMMENT` because GitHub won't accept a gating verdict on one's own PR — the intended verdict is **Approve**.)_

### Correctness

Traced the `_sendLock` lifecycle across all three `SendData` exit paths — it's released exactly once on each, with no double-release or leak:

- **Success** → released at the tail (`SocketContext.cs:144`).
- **Timeout** → `Abort()`, then the release is handed to a `ContinueWith` continuation that fires once the abandoned send settles (`SocketContext.cs:125-133`); the method returns at `:135` before reaching the tail release, so no double-release. Once aborted, the socket is `Aborted`, so a subsequent `SendData` short-circuits at `:89` and `Close` skips its send — the wedge can't recur.
- **Other fault** → `WaitAsync` only rethrows once `sendTask` has settled, so releasing at `:140` is safe.

`sendTask` starts outside the `try`, but since `SendChunks` is `async` any synchronous throw is captured in the returned task and surfaces through `WaitAsync` into the generic catch — no unobserved escape.

### Test coverage

- `SocketContextTests.SendData_WedgedSend_AbortsInsteadOfHangingForever` pins the fix in isolation on `SocketContext`.
- `SocketCommandExecutionTests.ExecuteCommand_ResponseSendWedges_...` drives the real `SocketHandler` → `SocketContext` path. Worth calling out that `StubCommandFactory(_ => null)` makes the command **succeed**, so the wedge lands on the success-path response send at `SocketHandler.cs:226` — awaited *under the command lock* — which is precisely the permanent-pin scenario #1760 describes, not an approximation. It then proves a second command runs, confirming the command lock was freed.
- `FakeWebSocket` now models real `Abort()` semantics (racing the gate against an abort CTS), and the `_activeSends--` decrement correctly moved into a `finally` so an aborted send no longer records as completed.

### What's good

- The fix stays inside `SendData` itself, so it self-bounds regardless of whether the caller holds the command lock — the right layer for the fix.
- Comments and the `backend.md` update accurately explain *why* cancellation alone can't unblock a wedged frame and why the lock is handed off rather than released inline.
- Consistent with the existing constructor-injection-with-default pattern used for `closeAbortTimeout` and the command timeout.

Nothing to change. Will merge once `test-backend` finishes green.

---

PR #1794, opened 2026-07-09
Guard the progress cache advance against resurrecting a partial hash

### LINKED ISSUE #1761: Dirty-row progress cache advance can resurrect a partial hash if the Redis key vanishes between load and save
## Problem

Since #1682 the player-progress cache is a Redis hash advanced by writing only dirty rows:

- `Game.DataAccess/Repositories/PlayerProgressRepository.cs:109-127` — `Save` advances the cache by `HSET`ting only the dirty rows.
- `Game.Infrastructure/Cache/Redis/RedisService.cs:167-189` — `HashSetAndForget` **creates** the key when absent.

If `Progress_{playerId}` disappears between the battle handler's load and the save's advance (Redis eviction under memory pressure — `docs/backend-persistence.md` explicitly blesses `volatile-*` policies — or an operator delete), the advance **recreates the hash holding only that battle's dirty rows**. The cache is the source of truth and a present key never falls through to the DB, so an active player's every other statistic/challenge/proficiency row is silently shadowed for as long as they stay active (the sliding TTL never lapses): regressed stats, possibly re-completable challenges.

The pre-#1682 whole-blob write did not have this failure mode — the aggregate always rewrote every row it held.

## Fix shape

Make `Save`'s advance an "HSET only if the key still exists" script (the miss path already rebuilds the full set from the DB); keep `HashSetAndForget`'s create-if-missing behavior for the miss-reload path. Integration test: delete the key mid-flight and assert the next read rebuilds from the DB rather than serving the partial hash.

ginderjeremiah:
## Summary

Since #1682 the player-progress cache is a Redis hash advanced by `HSET`ting only the rows a save actually dirtied (`Game.DataAccess/Repositories/PlayerProgressRepository.cs`). That advance only ever holds a *partial* view of the player's progress — this save's dirty rows — but it wrote unconditionally via `HashSetAndForget`, which **creates** the key if it's missing.

If the `Progress_{playerId}` Redis key disappeared between a battle handler's load and its save's advance (eviction under memory pressure — `docs/backend-persistence.md` explicitly blesses `volatile-*` policies — or an operator delete), the advance would recreate the hash holding **only that battle's dirty rows**. Since the cache is the source of truth and a present key never falls through to the DB, every other statistic/challenge/proficiency row the player held would be silently shadowed for as long as they stayed active (the sliding TTL never lapses) — regressed stats, possibly re-completable challenges.

## Changes

- `Game.Abstractions/Infrastructure/ICacheService.cs` / `Game.Infrastructure/Cache/Redis/RedisService.cs`: new `HashSetIfExistsAndForget` — same atomic HSET-fields-plus-refresh-TTL script as `HashSetAndForget`, but a no-op (fields untouched, TTL untouched) if the key doesn't currently exist.
- `Game.DataAccess/Repositories/PlayerProgressRepository.cs`:
  - `Save`'s cache advance now routes through `HashSetIfExistsAndForget` instead of `HashSetAndForget`, so it can never resurrect a key that vanished since the preceding load.
  - `GetCachedProgress`'s miss-reload path is the one place still allowed to *create* the key, since it just rebuilt the full row set from the DB. A brand-new player with zero rows yet leaves nothing to `HSET` (a hash can't exist with zero fields), so that path now writes a small presence-marker field (`_`) instead — otherwise that player's first-ever save (whose dirty rows *are* their entire state, and would otherwise be safe to create from) could never establish the key, since the guarded advance would see it missing and refuse to create it. `FromHashFields` already ignores any field outside the `S_`/`C_`/`P_` prefixes, so the marker is invisible to every read.
- `docs/backend-persistence.md`: documents the conditional advance and the presence-marker mechanism.

## Tests

- `Game.Application.Tests/DataAccess/RedisServiceIntegrationTests.cs`: new cases for `HashSetIfExistsAndForget` — no-ops on a missing key, overwrites only named fields (and refreshes TTL) on an existing hash, and no-ops with no fields.
- `Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs`: new `Save_CacheKeyVanishesBetweenLoadAndSave_DoesNotResurrectAPartialHash` — seeds a player with a pre-existing statistic, loads (establishing the cache), deletes the Redis key mid-flight to simulate the race, then saves a new battle-completion mutation. Asserts the advance does **not** recreate the key, and that a subsequent read correctly falls through to the DB and serves the original value rather than a shadowed/regressed one.
- `Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs`: updated `Save_DuringAPlayerSave_DefersToTheSharedBatchFlush_InsteadOfItsOwnRoundTrip` to `Load` before mutating (mirroring the real `BattleStatisticsEventHandler` flow) instead of constructing a `PlayerProgress` directly — the latter skipped the step that establishes the cache key in the first place, which isn't representative of any real call path (every production `Save` is preceded by a `Load` on the same object).
- `Game.Api.Tests/Unit/SocketCommandProcessorTests.cs`: added the new interface member to the two hand-rolled `ICacheService` test doubles.

## Test plan

- [x] `dotnet test` — full solution suite passes: 3042/3042 (Game.Core.Tests, Game.Application.Tests, Game.Api.Tests, Game.Infrastructure.Tests).

Closes #1761

---
_Generated by [Claude Code](https://claude.ai/code/session_01LTL4QtNynLZrgWBQiPofe4)_


Game.Infrastructure/Cache/Redis/RedisService.cs
```diff
@@ -188,6 +188,34 @@ public void HashSetAndForget(string key, IReadOnlyDictionary<string, string> fie
...
+            }
+
+            var argv = new RedisValue[1 + fields.Count * 2];
+            argv[0] = (long)expiry.TotalMilliseconds;
+            var i = 1;
+            foreach (var (field, value) in fields)
+            {
+                argv[i++] = field;
+                argv[i++] = value;
+            }
```
ginderjeremiah: Nit (optional, DRY): this argv-building block is now verbatim identical to the one in `HashSetAndForget` (lines 174–181) — the `[expiry, then field/value pairs]` layout is a shared contract of both scripts. Could hoist it into a small `private static RedisValue[] BuildHashArgv(IReadOnlyDictionary<string, string> fields, TimeSpan expiry)` so the two methods reduce to the early-return + `ScriptEvaluate` with their respective scripts. Not blocking — the two scripts differ, so keeping the methods separate is right; only the argv assembly is duplicated.

---

ginderjeremiah (commented):
## Review: guard the progress cache advance against resurrecting a partial hash (#1761)

**Verdict:** Approve (submitted as a `COMMENT` because I'm the PR author, so GitHub won't accept a gating event — the real call is approve) — the change correctly and safely closes #1761, matches the issue's suggested fix shape, and is thoroughly tested. One optional DRY nit only.

### What it does & why it's right
- `HashSetIfExistsAndForget` reuses the exact atomic HSET-fields-plus-`pexpire` script as `HashSetAndForget`, wrapped in an `exists` guard, so a vanished key is never resurrected from a save's partial (dirty-only) view. Correct and atomic.
- Routing `Save`'s advance through the guarded variant while leaving `GetCachedProgress`'s miss-reload as the one place allowed to *create* the key is exactly the right split — the reload just rebuilt the authoritative full row set from the DB, the save never has more than a partial view.
- The presence-marker (`_`) for a brand-new player's empty reload is a clean solution to "a Redis hash can't exist with zero fields": it keeps the key present so that player's first real save can advance it, and `FromHashFields` already ignores any field outside the `S_`/`C_`/`P_` prefixes, so it's invisible to every read.

### Correctness verification
- The design leans on the invariant that **every `Save` is preceded by a `Load`** (which establishes the key) — I confirmed both production call sites hold it: `BattleStatisticsEventHandler` (load → save) and `OfflineProgressService` (load at :141 → save at :340). No other production `Save` path exists.
- The one narrow behavioral change worth naming (not a defect): pre-#1682, a fresh-player save whose dirty rows happen to *be* the player's entire state would recreate an evicted key from that full state; now that path no-ops and relies on DB convergence + the next miss-reload instead. That's the correct trade-off — the `Save` site can't distinguish a full-state save from a partial one, so refusing to ever create from it is what actually forecloses #1761 — and the residual (dropped write + read-before-drain) is the already-accepted exposure documented in `backend-persistence.md`.

### Test coverage
Solid and appropriately targeted:
- `RedisServiceIntegrationTests`: the new method's full contract — no-op on missing key, overwrite-only-named-fields + TTL refresh on an existing hash, no-op with no fields.
- `PlayerProgressRepositoryIntegrationTests.Save_CacheKeyVanishesBetweenLoadAndSave_...`: directly reproduces #1761 (delete mid-flight) and asserts both that the key isn't resurrected and that the read falls through to the DB serving the un-shadowed value.
- The presence-marker path is transitively but genuinely covered by the existing fresh-player tests (`Save_WritesCache_...`, `Save_OnlyWritesTheRowsThatChanged_...`): they Load an empty player then Save and assert a cache-served read — they'd fail if the marker were dropped and the advance no-op'd.
- The `Save_During...` test change (Load-before-mutate) makes the fixture representative of the real path rather than testing behavior no production caller exercises. Good call.

### Nits
- One optional DRY nit inline on `RedisService.cs` — the argv-building block is duplicated verbatim between the two hash-set methods.

### Docs
- `docs/backend-persistence.md` is updated to describe both the conditional advance and the presence-marker mechanism. In line with the project's documentation expectations.

Not auto-merging: CI is still pending (only `detect-changes` has been scheduled, currently queued) and there's the open nit — leaving the merge decision to a green pipeline.

---

ginderjeremiah:
## Review — Approve

Verified the diff against the actual `PlayerProgressRepository`, `RedisService`, `ICacheService`, and `docs/backend-persistence.md`. The fix matches the shape #1761 prescribed and is correct.

**What it does:** Adds `HashSetIfExistsAndForget` — the same atomic HSET-fields-plus-`pexpire` Lua script as `HashSetAndForget`, wrapped in an `if exists(KEYS[1])` guard so a partial (dirty-only) view can never recreate an evicted/deleted key. `Save`'s cache advance now routes through it; the miss-reload path in `GetCachedProgress` remains the only place allowed to *create* the key (it just rebuilt the full row set from the DB), and a zero-row reload writes a `_` presence marker so a brand-new player's first save can still land on an existing key.

**Correctness:**
- The guard closes the resurrection window: if the key vanishes between load and save, the advance no-ops and the next per-battle reload falls through to the DB (stale-but-consistent while the write-behind lags, converging — never a partial shadow). Verified the deferred `OnFlushed(AdvanceCache)` path uses the guarded call too, so the in-flight-player-save hot path is covered identically.
- Presence marker is invisible to reads: `FromHashFields` only dispatches on `S_`/`C_`/`P_` prefixes, so `_` is silently dropped. A hash holding only the marker is a genuine cache hit that returns empty progress rather than re-missing every read — a minor behavioral improvement, consistent with the existing empty-hash type-check.
- Ordering holds because the marker write and the guarded advance are pipelined in-order on the shared multiplexer, so the `exists` check sees the marker the preceding empty reload wrote. This rests on the same in-order fire-and-forget assumption the write-behind model already depends on.
- `BuildHashArgv` extraction is a clean DRY refactor of the shared argv layout; no dead code, no domain logic leaking into the data tier.

**Presence-marker design:** Acceptable, not hacky. Given the O(dirty) constraint from #1635, `Save` genuinely can't know whether its dirty rows are the player's entire state, so *something* must establish the key from the authoritative full-state path — the marker is the minimal way to do that when that full state is empty. It's well-documented at both the field declaration and the write site. The one smell is that it leans on `FromHashFields`' prefix filtering as an implicit contract, and the `_` field lingers in the hash permanently once real rows arrive (harmless — never read, TTL refreshed normally). Not worth changing.

**Tests:** Good. `RedisServiceIntegrationTests` covers the three branches (missing-key no-op, existing-hash field overwrite + TTL refresh, no-fields no-op). The new `Save_CacheKeyVanishesBetweenLoadAndSave_DoesNotResurrectAPartialHash` reproduces #1761 directly and asserts DB fall-through serves the original value. Tightening `Save_During…` to `Load` before mutating better mirrors the real `BattleStatisticsEventHandler` flow. Worth noting the brand-new-player presence-marker path is regression-covered by the pre-existing `Save_WritesCache_SoASubsequentReadIsServedWithoutTheDatabase` (it would fail if the marker were removed), just not named for it.

**Nits (non-blocking):**
- The two Lua scripts duplicate the `for i=2,#ARGV,2 do hset … end + pexpire` fragment. Composing them would hurt readability more than it helps — leave as-is.
- Consider a follow-up test explicitly named for the brand-new-player first-save-creates-the-key path, to make the marker's purpose self-evident rather than relying on an unrelated test to catch its regression.

**CI:** Green — `test-backend`, `verify-backend`, all three `test-e2e` browsers, and `all-checks-passed` succeeded on `c73ef60`; frontend jobs correctly skipped (backend-only).

No must-fix items. LGTM.

---

PR #1795, opened 2026-07-09
Distinguish a definitive refresh rejection from a retryable failure

ginderjeremiah:
## Summary

`UI/src/lib/api/auth.ts`'s `performRefresh` returned `null` for both a network throw and any `!response.ok`, and `refreshTokens` cleared the stored token pair on any `null` — treating a transient failure exactly like a definitive "the refresh token is spent/revoked" rejection.

The backend's `Login/Refresh` (`Game.Api/Controllers/LoginController.cs`) only has one failure path — an invalid/expired refresh token, returned as **400** (not 401, since `ApiResponse.Error` defaults to `ApiErrorCategory.BadRequest`). A network blip, a `5xx`, or a `429` from the auth rate limiter is not proof the refresh token is dead, but was logging the user out anyway.

## Failure scenario this fixes

Wi-Fi blips → the keepalive reconnect's handshake fails pre-open → `handleClose` takes the auth-retry branch → `refreshTokens()` → `fetch` throws → tokens cleared + redirect to login. An idle game whose normal usage is long-lived background tabs hit this in ordinary conditions.

## Changes

- `auth.ts`: `refreshTokens`/`ensureValidAccessToken` now return a discriminated outcome — `success` / `rejected` / `retryable` — instead of `StoredTokens | null`. Only `rejected` (the refresh endpoint's 400, or no refresh token stored at all) clears tokens; `retryable` (network error, non-400 non-2xx, or a malformed 2xx body) leaves the stored pair intact.
- `api-request.ts`: the 401-retry path only calls `handleAuthFailure()` on a `rejected` outcome; a retryable refresh just returns the original 401 to the caller (the existing `isRetry` guard still caps this at one refresh attempt).
- `api-socket.ts`: both the pre-emptive open-time refresh and the handshake-auth-retry close handler now only route to `handleAuthFailure()` on `rejected`. On `retryable` they leave the queue and tokens intact and bail out — the keepalive ping's next `ensureSocket()` call retries the connect rather than forcing a logout.
- `docs/frontend.md`: documents the `rejected`-vs-`retryable` contract under Authentication.

## Tests

- `auth.test.ts`: new cases for a 400 rejection, a network throw, a 5xx, and a 429, pinning that only the 400 case clears storage — this also closes the "no network-throw scenario at all" gap called out in #1771.
[...]

Issue #1763, opened 2026-07-09
A transient network failure during token refresh logs the user out (tokens cleared on any refresh failure)


ginderjeremiah:
## Problem

`UI/src/lib/api/auth.ts:30-52` — `performRefresh` returns `null` both for a network throw and for any `!response.ok` (including 5xx), and `refreshTokens` (`auth.ts:69-80`) calls `clearTokens()` on any null. The code cannot distinguish "refresh token spent/revoked" (session is genuinely dead) from "server unreachable / 503" (retryable). Both socket paths treat a null refresh as terminal → `handleAuthFailure()` (`UI/src/lib/api/api-socket.ts:118-133`, `:357-392`).

## Failure scenario

Wi-Fi blips → the keepalive reconnect's handshake fails pre-open (`socketOpened = false`, code 1006) → `handleClose` takes the auth-retry branch → `refreshTokens()` → fetch throws → `clearTokens()` + redirect to login. An idle-game session dies on a network blip, and the still-valid refresh token is destroyed locally. For a game whose normal usage is long-lived background tabs, this fires in ordinary conditions.

## Fix shape

Make `performRefresh` return a discriminated result (`rejected` vs `retryable`); clear tokens only on a definitive 401/403, and let the socket's bounded retry/keepalive machinery handle retryable failures. Pin the chosen semantics with tests — `tests/lib/api/auth.test.ts` currently has no network-throw scenario at all (see also the companion test-gap issue).

_Found during a codebase health review._

PR #1796, opened 2026-07-09
Adopt another tab's rotated refresh token instead of clearing on a cross-tab race

### LINKED ISSUE #1764: Cross-tab refresh race can burn the single-use refresh token and log out every tab
## Problem

The documented "concurrent callers share a single request so the single-use refresh token is only consumed once" guarantee (`UI/src/lib/api/auth.ts:28` — `inFlightRefresh` is a module variable) only holds **within one tab**, while the token pair lives in shared `localStorage` (`UI/src/lib/api/token-store.ts:14-43`). Two tabs are a supported state — `SocketReplaced` exists precisely for that.

## Failure scenario

Two tabs refresh near-simultaneously; both present the same single-use refresh token. The loser gets `!ok`, and `clearTokens()` nukes the winner's freshly-rotated pair from shared storage too — both tabs logged out.

## Fix shape

Serialize cross-tab via `navigator.locks` (with a fallback), and/or on failure re-read storage: if the stored refresh token differs from the one presented, another tab already rotated — adopt the stored pair instead of clearing. The re-read guard also softens the transient-failure logout (see companion issue on refresh-failure semantics).

ginderjeremiah:
## Summary

`refreshTokens()`'s "concurrent callers share a single request so the single-use refresh token is only consumed once" guarantee (`UI/src/lib/api/auth.ts` — `inFlightRefresh` is a module variable) only holds **within one tab**, while the token pair lives in shared `localStorage` (`UI/src/lib/api/token-store.ts`). Two tabs are a supported state — `SocketReplaced` exists precisely for that.

Two tabs refresh near-simultaneously; both present the same single-use refresh token. The loser gets `!response.ok` from the backend (the token was already consumed by the winner), and `refreshTokens()` previously called `clearTokens()` unconditionally on any failure — nuking the winner's freshly-rotated pair from shared storage too. Both tabs end up logged out.

## Changes

- `auth.ts`: on a failed refresh, re-read storage before clearing. If it now holds a *different* refresh token than the one just presented, another tab already won the race and rotated it — adopt the stored pair instead of clearing. Only when storage still holds the same token we presented (no other tab touched it) is the session actually treated as dead.
- `docs/frontend.md`: documents the cross-tab re-read guard under Authentication.

## Note for reviewers

The issue's fix shape offered two options ("Serialize cross-tab via `navigator.locks` … and/or on failure re-read storage"). I went with the storage re-read alone — it directly and completely closes the described race without a browser-API fallback to maintain, and matches the existing `StoredTokens | null` contract with no signature changes. Happy to layer `navigator.locks` on top if you'd prefer belt-and-suspenders, but it isn't needed for correctness here.

This only addresses the cross-tab race. #1763 (transient-failure-logs-out-the-user) is a separate, already-in-review fix (#1795) that changes what counts as a "failure" in the first place (network/5xx vs. a definitive 400 rejection) — this PR doesn't depend on it and should merge cleanly either way, since it only touches the failure branch's clear-vs-adopt decision, not whether a given response counts as failure.

## Tests

- `auth.test.ts`: new case simulating a losing tab's refresh failing after a winning tab has already rotated storage — asserts the loser adopts the winner's pair rather than clearing it.

## Test plan

- [x] `npx vitest run` (`npm run test:unit:ci`) — full suite passes: 3547/3547 (including the new case above).

Closes #1764

---
_Generated by [Claude Code](https://claude.ai/code/session_01R1HGHnb7fM1eP5Mx4diENH)_


UI/src/lib/api/auth.ts
```diff
@@ -70,10 +77,16 @@ export const refreshTokens = (): Promise<StoredTokens | null> => {
...
 				setTokens(tokens);
-			} else {
-				clearTokens();
+				return tokens;
 			}
-			return tokens;
+
+			const current = getRefreshToken();
+			if (current !== null && current !== refreshToken) {
+				return getTokens();
```
ginderjeremiah: Nit (optional): this re-reads storage twice — `getRefreshToken()` then `getTokens()` each call `getTokens()` internally and re-parse the same localStorage entry. Reading once is a touch DRY-er and makes the "adopt exactly the pair we compared against" intent explicit:

```ts
const current = getTokens();
if (current !== null && current.refreshToken !== refreshToken) {
    return current;
}
```

No correctness difference (this whole block runs synchronously, so the two reads can't observe different values) — purely a small cleanup. Take it or leave it.

---

ginderjeremiah (commented):
## Review: cross-tab refresh race no longer logs out every tab (#1764)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the fix correctly closes the described race, is well-tested, and matches the issue's intent. One optional nit below; no blockers.

The change is tightly scoped and correct. On a failed refresh, `refreshTokens` now re-reads the stored refresh token: if it differs from the one just presented, another tab already won the race and rotated the pair, so we adopt the live pair instead of `clearTokens()`-ing a session that's still alive elsewhere. The adopted pair carries the winner's fresh access token, and because the loser never presents that token to the backend, there's no double-consumption of the single-use token. The `inFlightRefresh` per-tab collapsing still holds, so concurrent in-tab callers all receive the adopted pair.

I confirmed the scoping note in the PR body: this only touches the clear-vs-adopt decision on the failure branch, so it's orthogonal to #1795 (what counts as a failure) and the two should merge independently in either order.

### Nits

- `auth.ts:83-85` — the re-read calls `getRefreshToken()` then `getTokens()`, re-parsing the same localStorage entry twice. A single `getTokens()` read is marginally DRY-er and lets you return the exact pair you compared against. Purely optional (see inline comment).

### Test coverage

Solid and follows the frontend testing rules (unit test mirrors the source path under `tests/lib/api/`). Both failure branches are pinned:
- `auth.test.ts:52` — same token still in storage → clears and returns null (session genuinely dead).
- `auth.test.ts:65` — a winning tab rotates storage mid-flight → loser adopts the winner's pair rather than clearing.
- The existing concurrent-collapse case still passes.

### What's good

- The failure scenario is reasoned about at the right altitude — the fix distinguishes "token spent" from "another tab rotated it," which is the actual root cause.
- `docs/frontend.md` is updated with the cross-tab re-read guard under Authentication, keeping the docs the source of truth.
- Good judgment declining the `navigator.locks` half of the issue's suggested fix shape — the storage re-read fully closes the race without a browser-API fallback to maintain.

CI is green on the frontend + e2e jobs (backend jobs correctly skipped for a frontend-only change); `test-e2e (webkit)` was still running at review time.

---

ginderjeremiah:
## Review — cross-tab refresh re-read guard

**Verdict: approve** (net improvement, correct for the common case; one residual-window caveat below).

### What it does
On a failed refresh, instead of unconditionally `clearTokens()`, it re-reads `localStorage`: if the stored refresh token now differs from the one this tab presented, another tab won the race and rotated it, so the stored pair is adopted (and returned to the caller, letting the in-flight request retry with a valid token) rather than clearing a session that's still alive. Only when storage still holds the presented token is the session treated as dead.

### Correctness
- Logic is sound. `refreshToken` is captured before the request, and the `current.refreshToken !== refreshToken` comparison correctly distinguishes the winner's rotated token from the presented one — single-use rotation guarantees a *new* token, so a genuine "no other tab touched it" case always compares equal. `current === null` (a tab that logged out) correctly falls through to clear. The read-then-clear is synchronous (no `await` between `getTokens()` and `clearTokens()`), so no other tab's JS can interleave *within* that block.
- The adopt-and-return path is a nice touch: the losing tab's pending request succeeds on the adopted access token instead of bouncing.

### Residual race — the PR summary's "directly and completely closes the described race" is overstated
The guard closes the race only when the winner has *already persisted* `setTokens()` before the loser re-reads. But the two tabs process their HTTP responses independently, so this ordering is still possible:
1. Loser's rejection arrives first; it re-reads storage, still sees the old token (winner hasn't persisted yet), `clearTokens()`, returns `null`.
2. Winner's success arrives; `setTokens(new)`.
3. Loser's caller sees `null` → `handleAuthFailure()` → `clearTokens()` again (both `api-request.ts:113` and `api-socket.ts`), wiping the winner's fresh pair.

So in that timing sub-window both tabs can still end up logged out. It's a much narrower window than before (previously *any* failure cleared unconditionally), and this is a strict improvement — but it isn't a complete close. The genuinely complete fix is the `navigator.locks` serialization the issue offered (only one tab refreshes; others await the rotated result), which the author consciously deferred. Given the disclosed scope I'm fine merging the re-read as-is, but I'd (a) soften the "completely closes" wording, and (b) consider layering `navigator.locks` (or a re-read guard inside `handleAuthFailure` before its own `clearTokens()`) as a follow-up to close the tail. Not blocking.

### Tests
`auth.test.ts` covers the adopt path (winner rotates during the in-flight fetch → loser adopts, storage = winner). The clear path stays covered by the pre-existing failure test just above it. Adequate. Nice-to-have (not required): a case asserting that when storage still holds the *presented* token, it clears — to lock the negative branch against regressions independent of the older test.

### DRY / dead code
Clean. Restructure into an early success `return` reads well; `getTokens` import is used; no leftovers.

### Relationship to #1795 (issue #1763)
Orthogonal semantically — this PR only touches the clear-vs-adopt decision *inside* the failure branch, while #1795 changes what counts as a failure (transient network/5xx vs. definitive rejection). They compose fine. Heads-up on merge order: both edit the same `.then` failure block in `auth.ts`, so expect a **textual** conflict on the second-merged PR even though there's no semantic conflict.

### CI
Green — `all-checks-passed` success; `test-frontend`, `verify-frontend`, and all three e2e browsers pass; backend jobs correctly skipped (frontend-only change). Head `b24ba14`.

---

ginderjeremiah:
Follow-up filed for the residual race window (the second `clearTokens()` in `handleAuthFailure` isn't covered by this PR's re-read guard) plus the negative-branch test: #1829. This PR is a strict improvement over the prior unconditional clear, so merging now.

---

PR #1797, opened 2026-07-09
Detect a half-open socket via missed pongs and close it to trigger reconnect

### LINKED ISSUE #1765: Half-open (dead but readyState OPEN) socket is never detected — pings are sent but pongs are never checked
## Problem

`UI/src/lib/api/api-socket.ts:177-190` — `attemptPing` sends keepalive pings but nothing verifies a pong ever arrives; `timeoutRequest` (`:304-316`) settles the timed-out request's promise but leaves the "open" socket alone.

## Failure scenario

After laptop sleep/wake or a NAT drop, TCP can sit half-open for minutes: `ensureSocket` no-ops (readyState is still OPEN), pings vanish into the void, every command eats the full 30s timeout, and nothing recovers until the OS kills the connection. For an idle game, sleep/wake is the normal usage pattern — the keepalive only re-arms a socket the browser has *noticed* dropped.

## Fix shape

Track pong receipt; after N consecutive missed pongs — or on a request timeout — call `socket.close()` so the existing close/reconnect machinery takes over.

ginderjeremiah:
## Summary

`UI/src/lib/api/api-socket.ts` — `attemptPing` sent keepalive pings every 10s but nothing verified a pong ever came back; `timeoutRequest` settles a timed-out request's promise but leaves the "open" socket alone.

After laptop sleep/wake or a NAT drop, TCP can sit half-open for minutes: `ensureSocket` no-ops because `readyState` is still `OPEN`, pings vanish into the void, every command eats the full 30s request timeout, and nothing recovers until the OS eventually kills the connection. For an idle game, sleep/wake is the normal usage pattern.

## Changes

- `api-socket.ts`: each ping now tracks whether its pong arrived (`awaitingPong`) and counts consecutive misses (`missedPongs`). After `MAX_MISSED_PONGS` (3) consecutive unanswered pings, the socket is closed via `socket.close()` so the existing `handleClose` reconnect machinery takes over, instead of leaving every command to eat its own timeout. A pong received resets the streak; opening a fresh socket resets both.
- `docs/frontend.md`: documents the half-open detection under _Socket connection lifecycle_.

## Tests

- `api-socket.test.ts`: new `half-open detection (missed pongs)` suite —
  - closes the socket once `MAX_MISSED_PONGS` consecutive pings go unanswered
  - does not close the socket while every ping keeps getting answered
  - a pong resets the missed-pong streak, so an isolated slow round trip doesn't close the socket

## Test plan

- [x] `npm run test:unit:ci` (`vitest run`) — full suite passes: 3549/3549 (including the 3 new cases above).

Closes #1765

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tr62h52F66LVG4YMnUJAYX)_


UI/src/lib/api/api-socket.ts
```diff
@@ -184,7 +196,22 @@ export class ApiSocket {
...
+				this.missedPongs++;
+				if (this.missedPongs >= MAX_MISSED_PONGS) {
+					// readyState alone can't tell a half-open socket from a live one; MAX_MISSED_PONGS
+					// consecutive silent pings is the signal. Close it so handleClose's existing
+					// reconnect machinery takes over, rather than leaving every command to eat the full
+					// per-request timeout until the OS notices the connection is dead.
+					console.warn(`Socket missed ${this.missedPongs} consecutive pongs; closing the half-open connection.`);
+					this.awaitingPong = false;
+					this.missedPongs = 0;
+					socket.close();
```
ginderjeremiah, line 212: **Should-fix (add a comment):** the bare `socket.close()` here is load-bearing — it *must* stay code-less. `handleClose` branches on `ev.code === NORMAL_CLOSURE` (1000): a code-less close surfaces as 1005/1006, so it falls through to the `socketOpened` branch and lets the keepalive reconnect. If a future edit "tidies" this to `socket.close(NORMAL_CLOSURE)` (mirroring `disconnect()`), `handleClose` would take the normal-closure branch, call `stopPingInterval()`, and the socket would *not* auto-reconnect until the next user-initiated command — silently defeating the whole point of this detection. Worth a one-line comment noting the code-less close is deliberate so `handleClose` treats it as a reconnectable drop.

---

ginderjeremiah (commented):
## Review: Detect a half-open socket via missed pongs and close it to trigger reconnect (closes #1765)

**Verdict:** Approve _(posting as `COMMENT` because I'm the PR author — GitHub won't accept a formal approve on one's own PR)_ — the core detection logic is correct, well-commented, and well-tested, and the docs are updated. The items below are a should-fix comment and a scoping question, none blocking.

The change cleanly solves the described failure: an idle socket that goes half-open after sleep/wake is now detected within a bounded window (~3 unanswered ping cadences) and closed so the existing reconnect machinery takes over, rather than every command silently eating its 30s timeout forever. The `awaitingPong`/`missedPongs` state is reset at every correct point (ping send, pong receive, half-open close, and `openSocket`), and it correctly ignores the server-initiated `ping`→`pong` echo path — only a real pong to *our* ping clears the streak.

### Should-fix

- `api-socket.ts:209` — the bare `socket.close()` is deliberately code-less so `handleClose` treats it as a reconnectable drop (1005/1006), not the `NORMAL_CLOSURE` terminal branch that stops the keepalive. That's subtle and undocumented at the call site; a future "tidy up" to `socket.close(NORMAL_CLOSURE)` would silently disable auto-reconnect. Add a one-line comment. (Inline.)

### Nits / discussion

- **Request-timeout trigger not implemented (intentional?).** #1765's fix shape offered "after N consecutive missed pongs — *or on a request timeout* — call `socket.close()`." This PR implements only the pong path and leaves `timeoutRequest` settling just the one request. That's a defensible scope: the pong path is always-on and fully covers the idle sleep/wake scenario the issue centers on. But closing on a request timeout would meaningfully speed recovery in the *active* case — if the socket goes half-open right after commands are sent, each currently waits out its own full 30s before anything closes the socket, whereas closing on the first timeout would settle the rest via `CONNECTION_LOST` and reconnect immediately. Worth confirming this was a conscious scope decision; if so it's a reasonable follow-up, not something to hold this PR for.
- No test asserts that a reconnect actually follows the half-open close (the stated purpose). The three new cases cover the close/no-close/reset decision well; the reconnect-after-close leg leans on the existing reconnect tests. Optional to add.

### Test coverage

Good. The new `half-open detection (missed pongs)` suite covers the three core outcomes — closes after the threshold of unanswered pings, never closes while pongs keep arriving, and a late pong resets the streak so an isolated slow round trip is tolerated. Uses fake timers with clear cadence reasoning. Frontend-only change with no battle-logic surface, so no backend mirror is required.

### What's good

- The always-on, idle-safe detection is the right primary mechanism for this failure mode (the issue's scenario is a socket dropped while the tab sits idle).
- Comments explain *why* (readyState can't distinguish half-open) rather than restating the code, and `docs/frontend.md` gets an accurate bullet under _Socket connection lifecycle_.
- Reuses the existing close/reconnect path instead of adding a parallel recovery mechanism.

---

ginderjeremiah:
## Review: half-open socket detection (closes #1765)

**Verdict: approve** (posting as a comment). The change is correct, the state machine resets at every right point, and the tests exercise the three meaningful outcomes. CI is green (all-checks-passed; frontend/e2e/verify all success, backend appropriately skipped). Nothing blocking.

### Correctness

Traced the `awaitingPong`/`missedPongs` state machine end to end:

- Reset points are all covered — `openSocket` (fresh socket), pong receipt, ping send, and the half-open close. A pong for *our* ping clears the streak; the server-initiated `ping`→`pong` echo path correctly does **not** clear it.
- The threshold math is right: first ping arms `awaitingPong`, then 3 consecutive unanswered ticks (`missedPongs` 1→2→3) trip the close at `MAX_MISSED_PONGS`, so detection is bounded to ~30s of silence — no false positive under normal latency since a pong lands well within the 10s cadence and clears `awaitingPong` before the next tick.
- The code-less `socket.close()` is the right call to route through `handleClose`'s reconnect branch (1005/1006) rather than the `NORMAL_CLOSURE` terminal branch, and the inline comment now documents exactly that so a future "tidy-up" to `close(NORMAL_CLOSURE)` won't silently disable auto-reconnect. (Note this addresses the should-fix from the earlier self-review — it's present at head.)
- On reconnect after the half-open close, `openSocket` re-zeroes the tracking, so no stale streak carries over. No race in `attemptPing`: the fire-and-forget `ensureSocket()` only reopens when the current socket reads `CLOSED`, and the synchronous `readyState === OPEN` guard still refers to the old socket during the async reopen, so no ping is sent against a not-yet-created socket.

### Nits (non-blocking)

- The 2-line `awaitingPong = false; missedPongs = 0;` reset appears in three places; a small `resetPongTracking()` helper would DRY it and make the reset points self-documenting. The close-path reset is also redundant with `openSocket`'s reset (harmless, defensive).
- No test asserts a reconnect actually *follows* the half-open close — the three cases cover the close/no-close/reset decision well and lean on the existing reconnect suite for the follow-through. Optional to add.

### Scope note

#1765's fix shape offered "after N missed pongs — *or on a request timeout* — call `socket.close()`." This PR implements only the pong path and leaves `timeoutRequest` settling just its one request. That's a defensible scope call: the always-on pong path fully covers the idle sleep/wake scenario the issue centers on, and in the active case each in-flight command is still bounded to ~30s. Closing on the first request timeout would speed recovery in the active case slightly — reasonable as a follow-up, not a reason to hold this.

Good, focused change that reuses the existing reconnect machinery instead of adding a parallel recovery path, with an accurate `docs/frontend.md` update.

---

ginderjeremiah:
Follow-up filed for the issue's deferred second trigger (close the socket on a request timeout for faster half-open recovery in the active/commands-in-flight case), plus the `resetPongTracking()` DRY nit: #1828. The idle half-open case this PR fixes is complete on its own, so merging now.

---

Issue #1518, opened 2026-07-03
Character switcher: takeover check runs after the irreversible commit — declining destroys a healthy session


ginderjeremiah:
## Problem

`UI/src/routes/select/player-select-view.svelte.ts:126-137` — the switch flow runs `selectPlayer` **first** (switcher path: engines stopped, socket disconnected, token rotated to the new character) and only then checks `confirmTakeover` for an active session elsewhere. Declining the takeover at that point calls `logout()`, dumping the player to the login screen — a healthy session destroyed by saying "no".

## Why the ordering is currently forced

`Login/ActiveSession` checks the presence of the player named in the **token's** claim, so the client can't ask "does character X have a live session?" before rotating the token to X.

## Suggested direction

Give the presence check a target: either a `playerId` parameter on `ActiveSession` (validated against the authenticated account's characters, same anti-cheat shape as `SelectPlayer`) or a dedicated per-character presence endpoint. Then the switcher can ask first and only commit the switch after the player confirms. Declining becomes a true no-op (stay on the current character).

Needs a small API design call (parameterized presence check), so flagging rather than prescribing.

_Found during a codebase health review._

PR #1799, opened 2026-07-09
Await ensureValidAccessToken before reading the refresh token in CharacterSwitcher

### LINKED ISSUE #1767: CharacterSwitcher captures the refresh token before a request that can rotate it
## Problem

`UI/src/routes/game/CharacterSwitcher.svelte:67-77` captures `getRefreshToken()` and then posts `Login/SwitchPlayer` with it — but `SwitchPlayer` is not anonymous, so `ApiRequest.execute` runs `ensureValidAccessToken()` first (`api-request.ts:97-103`), which can rotate the token pair when the access token is inside the 30s expiry leeway. The captured refresh token is then already consumed.

## Failure scenario

Switch pressed while the access token is within the leeway window → pre-emptive refresh invalidates the captured refresh token → backend rejects the switch → `reloadToBoot()` on what should have been a clean character switch. Narrow window, but purely timing-dependent from the player's perspective. (Distinct from #1518's takeover-ordering problem.)

## Fix shape

`await ensureValidAccessToken()` first and read the refresh token after it settles — or have the backend source the refresh token server-side for `SwitchPlayer` instead of round-tripping it through the client.

ginderjeremiah:
## Summary

`UI/src/routes/game/CharacterSwitcher.svelte`'s `switchPlayer` captured `getRefreshToken()` and then posted `Login/SwitchPlayer` with it — but `SwitchPlayer` is not anonymous, so `ApiRequest.execute` runs `ensureValidAccessToken()` first (`api-request.ts`), which can pre-emptively rotate the token pair when the access token is inside the 30s expiry leeway. The refresh token captured before that call could already be consumed by the time the request actually went out.

Switch pressed while the access token is within the leeway window → the pre-emptive refresh inside `ApiRequest.execute` rotates the token pair → the backend rejects `SwitchPlayer` with the now-stale refresh token → `reloadToBoot()` fires on what should have been a clean character switch. Narrow, purely timing-dependent window from the player's perspective (distinct from #1518's takeover-ordering problem).

## Changes

- `CharacterSwitcher.svelte`: `switchPlayer` now `await`s `ensureValidAccessToken()` before reading the refresh token, so any pre-emptive refresh settles first and the token read is the one that's actually sent. (Went with the issue's first suggested fix shape rather than moving the refresh token server-side for `SwitchPlayer` — that's a bigger backend contract change for a narrow client-side race.)

## Tests

- `character-switcher.test.ts`: new case that simulates `ensureValidAccessToken` rotating the stored refresh token while it resolves (mirroring `ApiRequest`'s real pre-emptive-refresh behavior), and asserts `SwitchPlayer` is posted with the rotated token rather than the stale one captured before the await.

## Test plan

- [x] `npm run test:unit:ci` (`vitest run`) — full suite passes, including the new case above.

Closes #1767

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q1FcKXjKigD1XCpAAMwoLe)_

### DIFF
--- a/UI/src/routes/game/CharacterSwitcher.svelte
+++ b/UI/src/routes/game/CharacterSwitcher.svelte
@@ -39,7 +39,7 @@
 </Popover>
 
 <script lang="ts">
-import { ApiRequest, apiSocket, getRefreshToken, setTokens } from '$lib/api';
+import { ApiRequest, apiSocket, ensureValidAccessToken, getRefreshToken, setTokens } from '$lib/api';
 import { playerManager, stopEngines } from '$lib/engine';
 import Popover from '$components/Popover.svelte';
 import { confirmSessionTakeover } from '../login/session-takeover';
@@ -65,6 +65,10 @@ let view = $state<PlayerSelectView | null>(null);
 // credit + rebind through SwitchPlayer. Every terminal outcome ends in a full-page navigation, so the
 // brief window with a torn-down game is never interactive.
 const switchPlayer: PlayerSelectDeps['selectPlayer'] = async (playerId) => {
+	// SwitchPlayer is authenticated, so posting it would otherwise let ApiRequest.execute's own
+	// pre-emptive ensureValidAccessToken() rotate the refresh token *after* it had already been read
+	// below — settling that refresh here first means the token we read is the one that's actually sent.
+	await ensureValidAccessToken();
 	const refreshToken = getRefreshToken();
 	if (!refreshToken) {
 		reloadToBoot();
--- a/UI/src/tests/routes/game/character-switcher.test.ts
+++ b/UI/src/tests/routes/game/character-switcher.test.ts
@@ -5,16 +5,25 @@ import { render, fireEvent, cleanup, screen, waitFor } from '@testing-library/sv
 // switch/create), and the socket teardown, engine stop, token storage, and takeover are mocked so the
 // overlay can be driven without real I/O. window.location is replaced so the commit/recovery reloads are
 // observable instead of throwing in jsdom.
-const { getMock, postMock, disconnectMock, stopEnginesMock, getRefreshTokenMock, setTokensMock, confirmTakeoverMock } =
-	vi.hoisted(() => ({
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Await `ensureValidAccessToken` before reading the refresh token in `CharacterSwitcher` (closes #1767)

**Verdict:** Approve _(submitted as a comment since I authored the PR)_ — the fix correctly closes the timing race with a minimal, well-scoped change and a targeted regression test.

### Analysis

The race is real and the fix is correct:

- `refreshTokens` → `performRefresh` returns a **new** pair (rotated refresh token included) and `setTokens` persists it, so a pre-emptive `ensureValidAccessToken()` inside `ApiRequest.execute` consumes the refresh token that `switchPlayer` had already captured — exactly the failure #1767 describes.
- Awaiting `ensureValidAccessToken()` first settles any leeway-window refresh before the read, so `getRefreshToken()` returns the token that's actually sent. On the second call inside `execute`, the access token is freshly minted (~15 min out), so the leeway check is false and it won't rotate again — the captured token stays valid end-to-end.
- Ignoring the return value is fine: a failed refresh clears tokens, so `getRefreshToken()` returns `null` and the existing `reloadToBoot()` guard handles it.

Went with the issue's client-side fix shape over the server-side-sourced-token alternative, which is the right call for a narrow client race — the backend contract change would be a much larger unit for no additional correctness.

### Test coverage

- `character-switcher.test.ts` adds a focused case that mirrors the race: `ensureValidAccessToken` rotates the stored refresh token as it settles, and the test asserts `SwitchPlayer` posts the rotated token, not the stale one. This pins the ordering that the fix depends on. Frontend-only change, not battle logic, so no backend mirror is required.

### What's good

- The 3-line comment explains *why* the await is there — precisely the kind of context that stops a later edit from "simplifying" it away and silently reintroducing the race.
- Minimal surface, no collateral changes, clean regression coverage.

No blocking findings, no should-fixes, no nits. Merging once CI is green.

---

PR #1800, opened 2026-07-09
Guard Workbench/progression store mutators against an in-flight save

### LINKED ISSUE #1768: Workbench: edits made during an in-flight save are silently discarded
## Problem

`UI/src/routes/admin/workbench/entity-store.svelte.ts:112-122` — `patch()` has no `saving` guard, and `save()` (`:195-207`) wholesale-replaces both `items` **and** `base` from the server response after the multi-round-trip persist (`save-helpers.ts:112-166`). The detail pane's inputs stay live while only the Save/Discard buttons disable (`WorkbenchDetail.svelte`). The progression editor has the same pattern (`progression/progression-store.svelte.ts`).

## Failure scenario

Admin hits Save and keeps typing; the save lands; the keystrokes vanish — and because the baseline was replaced too, the record reads "clean", so there's no dirty indicator hinting anything was lost.

## Fix shape

Lock the detail pane while `saving` (reuse the deleted-record `.locked` treatment) or guard the mutators on `this.saving`; apply the same to the progression editor.

ginderjeremiah:
## Summary

`EntityStore.patch()` (`UI/src/routes/admin/workbench/entity-store.svelte.ts`) had no `saving` guard, and `save()` wholesale-replaces both `items` and `base` from the server response after the persist completes. The detail pane's inputs stayed live during a save — only the Save/Discard buttons disabled — so an admin who kept editing while a save was in flight would have those edits silently overwritten by the post-save baseline replacement, with no dirty indicator hinting anything was lost. The progression editor (`workbench/progression/progression-store.svelte.ts`) has the identical pattern via `patchPath`/`patchProf`.

Admin hits Save and keeps typing (or clicks Reset/Remove/Retire/Add on the same record); the save lands; the keystrokes/action vanish because the store's baseline is replaced from the server response, so the record reads "clean" with no trace anything was lost.

## Changes

- `entity-store.svelte.ts`: every mutator that touches `items`/`deleted` (`patch`, `addItem`, `removeItem`, `restoreItem`, `resetItem`) now no-ops while `this.saving` is true. `setRetired` is covered transitively since it routes through `patch`.
- `progression-store.svelte.ts`: same treatment for `patchPath`, `patchProf`, `addPath`, `addTier`, `reorderTiers`, `removePath`, `removeTier`. Every other mutator (`resetPath`/`resetProf`/`retirePath`/`retireProf`/the milestone-modifier, reward, and prerequisite editors) routes through `patchPath`/`patchProf`, so they're covered transitively too.
- `WorkbenchDetail.svelte`, `PathDetail.svelte`, `TierDetail.svelte`: the existing "locked" dim/pointer-events-none treatment (previously only for a deleted/retired record) now also applies while `store.saving` is true, so the pane visibly communicates it's locked during a save rather than just silently rejecting input.

I went with guarding the store mutators directly (one of the two fix shapes the issue proposed) rather than only relying on the CSS lock, since the header action buttons (Reset/Retire/Remove/Restore) sit outside the locked pane region in both `WorkbenchDetail` and the progression editor's `DetailHeader` — a CSS-only lock wouldn't have stopped those from racing a save. Guarding at the store level closes that gap without needing to thread a `saving` prop through every button.

## Test plan

- [x] `npx vitest run` (`npm run test:unit:ci`) — full suite passes: 3550/3550, including new cases pinning that every mutator no-ops while `saving` is true, for both `EntityStore` and `ProgressionStore`.

Closes #1768

---
_Generated by [Claude Code](https://claude.ai/code/session_01CktKeMigSXpZwzDXqb4wP7)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
+++ b/UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
@@ -77,7 +77,7 @@
 		</div>
 
 		<div class="detail-body">
-			<div class="body-inner" class:locked={status === 'deleted'}>
+			<div class="body-inner" class:locked={status === 'deleted' || store.saving}>
 				<div class="sec-title">
 					{curSection.label}{#if curSection.desc}<span class="sub">— {curSection.desc}</span>{/if}<span class="ln"
 					></span>
--- a/UI/src/routes/admin/workbench/entity-store.svelte.ts
+++ b/UI/src/routes/admin/workbench/entity-store.svelte.ts
@@ -109,7 +109,15 @@ export class EntityStore<T extends Identified> {
 		return { added, modified, deleted, total: added + modified + deleted };
 	});
 
+	/**
+	 * Edits made while a save is in flight would either be silently overwritten by that save's
+	 * post-persist baseline replacement or would land as a "clean" record whose dirty indicator
+	 * never fires — so every mutator no-ops while {@link saving} is true rather than risk either.
+	 */
 	patch(id: number, mutate: (draft: T) => void) {
+		if (this.saving) {
+			return;
+		}
 		this.items = this.items.map((record) => {
 			if (record.id !== id) {
 				return record;
@@ -122,6 +130,9 @@ export class EntityStore<T extends Identified> {
 	}
 
 	addItem(): number {
+		if (this.saving) {
+			return this.items[0]?.id ?? 0;
+		}
 		const id = this.nextId--;
 		this.items = [this.config.newItem(id), ...this.items];
 		this.saved = false;
@@ -129,6 +140,9 @@ export class EntityStore<T extends Identified> {
 	}
 
 	removeItem(id: number) {
+		if (this.saving) {
+			return;
+		}
 		if (this.baseMap[id] === undefined) {
 			// Never-saved record: just drop it.
 			this.items = this.items.filter((record) => record.id !== id);
@@ -139,6 +153,9 @@ export class EntityStore<T extends Identified> {
 	}
 
 	restoreItem(id: number) {
+		if (this.saving) {
+			return;
+		}
 		this.deleted.delete(id);
 		this.saved = false;
 	}
@@ -163,6 +180,9 @@ export class EntityStore<T extends Identified> {
 	}
 
 	resetItem(id: number) {
+		if (this.saving) {
+			return;
+		}
 		const baseline = this.baseMap[id];
 		if (baseline) {
 			this.items = this.items.map((record) => (record.id === id ? clone(baseline) : record));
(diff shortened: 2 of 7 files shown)


ginderjeremiah (commented):
## Review: Guard Workbench / progression store mutators against an in-flight save (#1768)

**Verdict:** Approve — the fix directly and completely closes the race from #1768 (edits landing during an in-flight save being silently overwritten by the post-persist baseline replacement), on both stores, with mirrored tests. (Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating self-review; the real call is Approve.)

### Why this is the right shape

Guarding the store mutators rather than relying on a CSS lock is the correct choice: the header actions (Add / Reset / Retire / Remove / Restore) live *outside* the locked detail-pane region, so a pane-only `.locked` treatment wouldn't stop them from racing a save. Closing the gap at the store level covers every entry point without threading a `saving` prop through each button, and the CSS `.locked` addition is kept as the visible signal that the pane is inert during a save.

### Correctness

- **`EntityStore`** — every mutator that touches `items`/`deleted` (`patch`, `addItem`, `removeItem`, `restoreItem`, `resetItem`) no-ops while `saving`; `setRetired` is covered transitively via `patch`. `save()` mutates `items`/`base` directly (not through the guarded mutators), so there's no self-lock.
- **`ProgressionStore`** — `patchPath`/`patchProf` plus the structural mutators (`addPath`, `addTier`, `reorderTiers`, `removePath`, `removeTier`) are guarded; every reset/retire/milestone-modifier/reward/prerequisite editor routes through `patchPath`/`patchProf`, so all are covered transitively. `save()`/`reseed()`/`setData()` write state directly, so no self-lock there either.
- Navigation/selection methods are intentionally left live during a save — they don't mutate record state, so browsing while a save is in flight stays fine.
- `addItem`/`addTier`/`addPath`'s during-save early returns hand back a valid existing id (or `0`) instead of minting a negative id; the only caller (`Workbench.svelte:108`) uses it purely for selection, so the worst case is selection nudging to an existing record — no crash, no orphaned negative id.

### Test coverage

Meets the project's frontend rule — the touched lib modules carry mirrored unit tests: `entity-store.test.ts` pins `patch`/`addItem`/`removeItem`/`restoreItem`/`resetItem` all no-op while `saving`, and `progression-store.test.ts` pins `patchPath`/`patchProf`/`addPath`/`addTier`/`reorderTiers`/`removePath`/`removeTier`. The three `.svelte` edits are a single-token `|| store.saving` on an existing `class:locked` binding — a trivial reactive derivation that doesn't warrant its own component test.

### What's good

Tight, DRY fix that leans on the existing transitive routing so only the base mutators needed guarding; the doc-comments on `patch`/`patchPath` explain *why* the no-op exists, which is exactly the kind of "a later edit might naively delete this guard" context worth keeping. No doc update needed — this is a localized behavioral fix, not an architectural decision.

If CI comes back green I'll merge this.

---

PR #1801, opened 2026-07-09
Batch proficiency prerequisite saves to fix order-dependent false cycle rejection

### LINKED ISSUE #1775: Progression editor: prerequisite child-saves are order-dependent — an acyclic gateway swap can be falsely rejected as a cycle, post-commit
## Problem

`UI/src/routes/admin/workbench/progression/progression-store.svelte.ts:415-431` posts `SetProficiencyPrerequisites` per tier in catalogue order, while the backend cycle check (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs:178-181,314-328`) builds the prospective graph from every *other* tier's **currently persisted** edges plus the one being set.

If one save reverses a gateway (tier A drops its prereq on B while B gains a prereq on A) and B's set posts first, the backend sees the still-persisted `A→B` plus the new `B→A` and rejects with a cycle error — for an edit whose end state is acyclic and valid. Because the identity batches have already committed, the store's catch re-seeds from server truth (`progression-store.svelte.ts:442-448`) and the operator's remaining unsaved edits are wiped.

## Fix shape

Order the prerequisite savers so shrunken sets post before grown ones (compare against baseline), or add a batch prerequisites endpoint validated against the final graph (preferred — removes the order sensitivity entirely).

ginderjeremiah:
## Summary

`UI/src/routes/admin/workbench/progression/progression-store.svelte.ts` posted `SetProficiencyPrerequisites` once per changed tier, in catalogue order. The backend's cycle check (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs`) built its prospective graph from every *other* proficiency's **currently persisted** edges plus the one being set.

If one save reverses a gateway (tier A drops its prerequisite on B while B gains one on A) and B's post happens to land before A's, the backend saw the still-persisted `A→B` edge plus the new `B→A` edge and rejected the whole save as a cycle — for an edit whose final state (`B→A` only) is acyclic and valid. Because the identity batches (path/proficiency Add/Edit) had already committed by that point, the store's catch handler re-seeded from server truth and silently wiped the operator's remaining unsaved edits.

## Changes

Per the issue's preferred fix shape ("add a batch prerequisites endpoint validated against the final graph — removes the order sensitivity entirely"):

- **`SetProficiencyPrerequisitesData`** is now submitted as a batch (`List<SetProficiencyPrerequisitesData>`, one entry per changed proficiency) rather than one at a time.
- **`IAdminProficiencies.SetPrerequisites`** / `AdminProficiencies.SetPrerequisites` now validate the *combined* prospective graph — every changed proficiency's desired edges plus every other proficiency's existing ones — in one pass, then reconcile all of them, before anything commits. New up-front guards: a batch can't name the same proficiency twice, and a single change's own prerequisite list can't contain a duplicate id (previously that duplicate check only ran once staging for that entry had already begun, which would have been unsafe for a batch — an earlier, valid entry's inserts/deletes would already be staged for the commit filter to persist alongside a later entry's rejected save).
- **`AdminProficienciesController.SetProficiencyPrerequisites`** now accepts the list payload; TypeScript client types regenerated via the standalone codegen (`api-type-map.ts`).
- **`ProgressionStore.save()`** now collects every changed tier's desired prerequisite set and posts them in one call instead of looping per tier.
- **`docs/backend-admin.md`**: documents this as the one full-desired-set setter that batches across owners, and why.

## Test plan

- [x] `dotnet test` — full solution suite passes: 3041/3041 (added `SetPrerequisites_BatchNamesTheSameProficiencyTwice_ReturnsFailure` and `SetPrerequisites_BatchedGatewaySwap_AcceptsTheCombinedAcyclicResult`, the latter reproducing the exact false-cycle scenario from the issue; updated the two existing prerequisite tests to the new batch signature).
- [x] `npm run test:unit:ci` — full suite passes: 3551/3551 (added a store-level test asserting the gateway swap posts as a single combined batch call).

Closes #1775

---
_Generated by [Claude Code](https://claude.ai/code/session_017Gx3Wd5iM4w35hGgdcotpN)_


ginderjeremiah (commented):
## Review: Batch proficiency prerequisite saves to fix order-dependent false cycle rejection (#1775)

**Verdict:** Approve — clean, correct implementation of the issue's preferred fix shape (a batch endpoint validated against the final combined graph), which removes the order sensitivity at the root rather than papering over it with a save-ordering heuristic. _(Submitted as a comment because GitHub won't accept a gating approval on one's own PR — the call is approve.)_

The reasoning holds up end-to-end:

- **Validation is fully front-loaded, staging is last.** All checks — batch-level duplicate-owner, per-change existence, self-reference, duplicate-desired-ids, and the combined-graph cycle detection — run before the reconcile loop stages a single entity-store op. That's exactly what the "validate before staging" contract in `backend-admin.md` requires, so a later invalid entry can never leave an earlier valid entry's inserts/deletes staged for the commit filter to persist. The added up-front `Distinct()` guard (over relying on the reconciler's own late duplicate check) is the right call, and the comment at `AdminProficiencies.cs:177-180` explains precisely why.
- **The combined prospective graph is built correctly** — `FindPrerequisiteCycle` overrides only the named nodes and keeps every other proficiency's persisted edges, so a multi-proficiency gateway swap is judged by its true final state while a real cycle (including one formed against an unchanged node's edges) is still caught.
- **`ToDictionary(c => c.Id, …)` is safe** because the duplicate-owner guard rejects repeated ids first; and each reconcile touches only its own `ProficiencyId` rows, so no cross-entry PK collision.
- **Frontend** collects every changed tier's desired set and posts once; step renumbering (5→6→7) is coherent and the `committed` flag is set on the batch post.

### Test coverage
Solid. Backend adds the exact false-cycle reproduction (`SetPrerequisites_BatchedGatewaySwap_AcceptsTheCombinedAcyclicResult`) and the duplicate-owner guard; the two existing prerequisite tests were migrated to the batch signature; the frontend adds a store-level test pinning the gateway swap to a single combined call. This matches the project's testing rules for the touched surface.

### What's good
Thoughtful front-loading of validation with a clear rationale in the comments, docs updated (`backend-admin.md` calls out this as the one full-desired-set setter that batches across owners, and why), and the fix targets the actual root cause named in #1775.

---

PR #1802, opened 2026-07-10
Post safely-resolvable gateway prerequisite removals before path retire

### LINKED ISSUE #1776: Progression editor: "retire path + remove the offending gateway" in one save can never succeed
## Problem

`AddEditPaths` is always the first call in the progression save sequence (`UI/src/routes/admin/workbench/progression/progression-store.svelte.ts:371-378`), and `FindRetiredPathGatingLiveGateway` (`Game.DataAccess/Repositories/Admin/AdminPaths.cs:20-30`) validates the retire against the **pre-save** prerequisite edges. So retiring a path whose tier gates a live gateway *and* removing that gateway in the same save is always rejected on the first POST — even though the combined end state is valid. The UI's retire-confirm flow actively invites the one-shot fix.

It fails pre-commit (edits kept, only a toast), but the operator is silently forced into an unexplained two-save dance.

## Fix shape

Post prerequisite removals before path identity changes when a retire is pending, or surface the guard's requirement in the retire dialog ("remove/save the gateway prerequisites first").

ginderjeremiah:
## Summary

`AddEditPaths` was always the first HTTP call in the progression editor's combined save (`UI/src/routes/admin/workbench/progression/progression-store.svelte.ts`), and the backend's `FindRetiredPathGatingLiveGateway` guard (`Game.DataAccess/Repositories/Admin/AdminPaths.cs`) validates a path retire against the **pre-save** prerequisite graph. So retiring a path whose tier gates a live gateway *and* removing that gateway's prerequisite in the same save was always rejected on the first POST — even though the combined end state is valid. The UI's retire-confirm flow actively invites exactly this one-shot fix, so the operator was silently forced into an unexplained two-save dance.

Path A has a tier that's a prerequisite of a live gateway on Path B. An operator retires Path A and, in the same editor session, removes A's tier from B's gateway prerequisites — intending to unblock the retire. Because the save posted `AddEditPaths` before `SetProficiencyPrerequisites`, the retire guard saw the still-live edge and rejected the whole save (edits kept, just a toast) — the operator had to save the prerequisite removal first, then retire in a second save.

## Changes

- `progression-store.svelte.ts`: `save()` now splits cross-path prerequisite changes into two groups. Changes that name only already-persisted tiers (both the gateway and every prerequisite id are already real, non-negative ids) need no id remap, so they're posted **first**, before path identities — clearing the guard's block before the retire is evaluated. Changes that name a brand-new tier (added in the same save, either as the gateway or as a prerequisite) still need the id map resolved from the path/proficiency identity saves, so they're deferred and posted last, exactly as before. The "combined batch, one call" cycle-safety property (documented in the existing code comment) is preserved within each group.
- `docs/frontend-admin.md`: updates the progression editor's save-orchestration description to reflect the new ordering.

## Note for reviewers

The fix is scoped to the common case described in the issue (retiring an *existing* path while removing an *existing* gateway's prerequisite — the only shape the backend guard can ever actually reject, since it only sees pre-save/cached state and can't complain about a gateway added in the same save). A change that spans a brand-new tier still posts in its original, later position, so I didn't touch that path's correctness guarantees at all.

## Tests

- `progression-store.test.ts`: two new cases —
  - retiring a path and removing the now-offending gateway prerequisite in one save succeeds, and asserts the prerequisite call is posted before the path-identity call (the fix's actual mechanism, not just the outcome).
  - retiring a path that still gates a live gateway (prerequisite left untouched) is still rejected, pinning that the existing guard behavior is unchanged.
  - Both new tests exercise a small fake-backend simulation of the real `FindRetiredPathGatingLiveGateway` guard, since the previous fake backend didn't model it at all (it would have silently accepted an invalid retire either way).

## Test plan

- [x] `npm run test:unit:ci` (`vitest run`) — full suite passes: 3553/3553 (including the 2 new cases above).

Closes #1776

---
_Generated by [Claude Code](https://claude.ai/code/session_0199fZ8nZbtKdqomqM2Zd18r)_


UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
```diff
@@ -401,7 +401,30 @@ export class ProgressionStore {
...
+				({ prof, prerequisiteIds }) => prof.id >= 0 && prerequisiteIds.every((id) => id >= 0)
+			);
+			if (resolvableNow.length) {
+				const changes: ISetProficiencyPrerequisitesData[] = resolvableNow.map(({ prof, prerequisiteIds }) => ({
+					id: prof.id,
+					prerequisiteIds
+				}));
+				await ApiRequest.post('AdminTools/SetProficiencyPrerequisites', changes);
+				committed = true;
+			}
```
ginderjeremiah: This split narrows the "one combined batch" cycle-safety invariant across the two groups, and I think it can now false-reject a save that previously succeeded.

The backend `SetPrerequisites` (`AdminProficiencies.cs:194-201`) runs cycle detection per POST against the *prospective* graph = every **unchanged** proficiency's existing edges + every **changed** proficiency's desired set. Because step 1 and step 7 are now separate POSTs, each validates independently.

Concrete regression — reverse a persisted edge while one side also touches a brand-new tier, in one save:
- Baseline: tier A requires tier B (edge B→A). Both A and B persisted.
- Edit A's prereqs to `[B removed, C added]` where C is a brand-new tier → A's change has a negative id in its set → deferred to **step 7**.
- Edit B's prereqs to `[A]` → all ids persisted → **resolvableNow**, posts in **step 1**.

At step 1, only B is "changed", so A contributes its *existing* edge (A still requires B) and B contributes its desired (B requires A) → transient A↔B cycle → the whole save is rejected with a cycle error. Under the old single-batch ordering both changes landed together, the combined graph had no cycle, and the save succeeded.

So this trades the #1776 false-rejection for a rarer one in the same defect class (and regresses a previously-valid save). Two options I'd consider:
1. **Gate the early split on a pending retire.** The early post only exists to clear the retire guard; if no path in this save is being retired, fall through to the original all-at-end single batch and the cycle invariant is fully preserved for every non-retire save (the overwhelming majority). That shrinks the exposure to only saves that retire *and* reverse a persisted edge *and* add a new tier.
2. At minimum, document the residual limitation explicitly in the step-1/step-7 comments so a future reader knows the cross-group cycle case is knowingly unguarded.

I'd lean on option 1 — it's a small, low-risk narrowing.

---

UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
```diff
@@ -462,19 +480,25 @@ export class ProgressionStore {
...
 			}
 
-			// 6. Cross-path prerequisites, all changed tiers in one call (see above).
+			// 7. Every remaining (not-yet-posted, step 1) cross-path prerequisite change — naming a brand-new
+			// tier as the gateway or as a prerequisite — collected into one combined batch: the backend
+			// validates a batch against its final combined graph, so a gateway swap spanning two tiers (one
+			// drops an edge while the other gains the reverse) can't be false-rejected as a cycle depending on
+			// which tier's post happens to land first.
+			const prerequisiteChanges: ISetProficiencyPrerequisitesData[] = allPrerequisiteChanges
+				.filter((change) => !resolvableNow.includes(change))
```
ginderjeremiah: Nit: `!resolvableNow.includes(change)` relies on object-reference identity — it works only because `resolvableNow` was `.filter`ed out of the same `allPrerequisiteChanges` array. That's a subtle coupling a later refactor (e.g. remapping/copying the objects) could silently break. A `Set(resolvableNow)` or a boolean flag on each change would make the partition explicit and O(1). Minor.

---

ginderjeremiah (commented):
## Review: split progression-save prerequisite posting so a retire + persisted-gateway-prereq removal succeeds in one save (#1776)

**Intended verdict:** Request changes — the fix for the common case is sound, but the split reintroduces a transient-cycle false-rejection (in the same defect class as #1776) for a rarer combined edit that used to succeed. Posting as a comment since I'm the PR author and GitHub won't accept a gating verdict on my own PR.

The core approach is right: the retire guard (`AdminPaths.FindRetiredPathGatingLiveGateway`) validates against the pre-save/live graph, so posting the persisted-only prerequisite removals **before** `AddEditPaths` is the correct way to let the retire clear in one shot. The partition (persisted-only → early, references-a-new-tier → deferred with id remap) is a reasonable way to keep the id-remap contract intact.

### Should-fix (my one real concern)

- **Cross-group cycle safety is no longer preserved — see inline at `progression-store.svelte.ts:415`.** The old design put *all* prerequisite changes in a single POST specifically so the backend validates the combined final graph and can't false-reject a mid-swap transient cycle. Splitting into two POSTs (step 1 vs step 7) validates each independently. A save that reverses a persisted A↔B edge while one side also gains a brand-new-tier prereq now splits across the two groups and gets rejected on the step-1 POST with a cycle error — a save the old ordering accepted. I confirmed the mechanism against `AdminProficiencies.SetPrerequisites` (`:194-201`). Suggested fix: gate the early split on whether a retire is actually pending, so every non-retire save keeps the original single-batch ordering and full cycle invariant. Details + alternatives inline.

### Nits

- `progression-store.svelte.ts:490-491` — the step-7 `!resolvableNow.includes(change)` partition relies on object-reference identity; a `Set` or explicit flag would be less fragile. Inline.

### Test coverage

Good, and better than the outcome-only bar: both new cases model the real `FindRetiredPathGatingLiveGateway` guard rather than asserting it away, and the success case asserts the *ordering* (prereq POST before the path POST) — the fix's actual mechanism, not just the happy result. The negative case pins that an untouched-prereq retire is still rejected. One gap: nothing covers the cross-group cycle regression above — if you take the retire-gating fix, a test that a non-retire edge-reversal + new-tier save still posts as one batch would lock the invariant in.

### What's good

- Correct read of *why* the original bug happened (guard validates the pre-save graph) and a targeted fix rather than reordering everything blindly.
- The deferred-group id-remap path and the "combined batch" cycle comment for brand-new tiers are preserved verbatim — the fix is scoped, not a rewrite.
- Docs (`frontend-admin.md`) updated to match the new orchestration.

---

UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
ginderjeremiah: Good catch, fixed in fed80b0. Gated the early split on `hasPendingRetire` (a path in this save transitioning to retired) — when no retire is pending, `resolvableNow` is always empty and every prerequisite change falls through to the single combined batch at step 7, so the full cycle-safety invariant holds for the overwhelming majority of saves. Added a test reproducing your exact scenario (persisted-edge reversal + a brand-new tier, no retire) asserting it still posts as one batch.

---

UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
ginderjeremiah: Fixed in fed80b0 — swapped `.includes()` for `new Set(resolvableNow)` + `.has()`.

---

ginderjeremiah:
## Review — approve (non-blocking nits)

**What it does:** Splits the progression save's cross-path prerequisite writes into two groups *only when the save also newly-retires a path*. The subset whose gateway id and every prerequisite id are already persisted (`>= 0`, no id remap needed) posts first — before `AddEditPaths` — so the offending edge is gone before the backend's `FindRetiredPathGatingLiveGateway` guard evaluates the retire. Everything else defers to the original combined batch at the end. A non-retiring save keeps the entire batch in one call, exactly as before.

**Correctness — solid.** Verified against the real code, not just the issue text:
- The issue's premise holds: `AdminPaths.FindRetiredPathGatingLiveGateway` (`Game.DataAccess/Repositories/Admin/AdminPaths.cs:61-117`) validates the retire against the cached (`_proficiencies`) pre-save graph, and `AddEditPaths` was the first POST — so retire + same-save removal was always rejected on the first call. The fix targets the only shape the guard can actually reject (an *existing* path retired while an *existing* gateway's prerequisite is dropped).
- The `resolvableNow` predicate (`prof.id >= 0 && prerequisiteIds.every((id) => id >= 0)`) correctly restricts the early batch to changes needing no id map; anything referencing a brand-new tier stays in its original later position, so the id-remap ordering is untouched.
- `hasPendingRetire` (`record.retiredAt != null && baseline.retiredAt == null`) matches the backend's live→retired "newly retired" semantics — a brand-new or already-retired path can't trigger it. Correct.
- **Cycle-safety is preserved in the common case, not just narrowed everywhere.** When a retire has no new tiers, *all* prereq changes are resolvable-now → they still post as a single early batch (step 7 ends up empty), so full combined-graph validation is retained. The invariant only narrows when a retire, a persisted-edge reversal, *and* a brand-new tier coincide — a genuinely rare intersection, and the code comment + `frontend-admin.md` document it precisely.
- The load-bearing assumption is backend cache coherence between the two requests: the guard must see the removed edge on the subsequent `AddEditPaths`. That's the exact fix the issue proposed, and the test's fake guard models it faithfully (it reads `serverProfs`, which the early `SetProficiencyPrerequisites` has already mutated).

**Tests — adequate.** Good call adding a real `FindRetiredPathGatingLiveGateway` simulation to the fake backend (the prior fake would have silently accepted an invalid retire). The three cases cover the meaningful axes: retire+removal succeeds *and* asserts prereq-before-path ordering (the mechanism, not just the outcome); retire without removal still rejects; non-retiring swap-with-new-tier stays one batch.

**Nits (non-blocking):**
- `const resolvedEarly = new Set(resolvableNow)` needs an eslint-disable for `svelte/prefer-svelte-reactivity`. Since `resolvableNow` holds the same object references and is tiny, `allPrerequisiteChanges.filter((c) => !resolvableNow.includes(c))` drops both the Set and the suppression for slightly cleaner code.
- `[...profDiff.added, ...profDiff.modified.map((m) => m.record)]` is now spread twice (building `allPrerequisiteChanges` and again in the step-6 modifier/reward loop); hoisting to one local would be marginally DRYer.

**Follow-up (optional):** consider a test pinning the *documented residual* — retire + persisted-edge reversal + brand-new tier producing the narrowed false-reject — so the known limitation is asserted rather than only described in prose.

**CI:** green — `test-frontend`, `verify-frontend`, all three e2e browsers, and `all-checks-passed` succeeded on `fed80b0`; backend jobs correctly skipped (frontend-only change).

No dead code left behind, docs updated accurately, and the change is scoped exactly to the failure it fixes.

---

PR #1803, opened 2026-07-10
Reject shrinking a proficiency's MaxLevel below a persisted payout

### LINKED ISSUE #1777: Shrinking a proficiency's maxLevel below persisted payout levels saves silently, leaving out-of-range payouts on the server
## Problem

The level-range rule is enforced only on the `Set*` child endpoints (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs:92,125`), not on the identity Edit (`:58-77`), and the editor's out-of-range warning is advisory (`UI/src/routes/admin/workbench/progression/progression-helpers.ts:178-189`).

If an operator lowers `maxLevel` from 10 to 5 while a persisted modifier/reward sits at level 7 and doesn't touch the collections, `childChanged` is false so `SetProficiencyModifiers`/`Rewards` are never posted (`progression-store.svelte.ts:418-424`): the identity Edit commits cap 5 and the server now holds payouts the level-range rule would reject — with no error anywhere. (If the operator *does* also touch the collection, the identity commits first and the child call then rejects — a post-commit failure that re-seeds away other edits.)

## Fix shape

Validate `MaxLevel` against existing child levels in `SaveProficiencies` (backend, preferred), and/or make the out-of-range warning block save in the editor.

ginderjeremiah:
## Summary

The level-range rule for a proficiency's per-level modifiers/rewards was enforced only on the `Set*` child endpoints (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs` — `SetModifiers`/`SetRewards`), not on the identity `Edit` (`SaveProficiencies`). The Workbench's out-of-range warning is advisory only.

An operator lowers a proficiency's `MaxLevel` from 10 to 5 while a persisted modifier/reward sits at level 7, without touching the modifier/reward collections in the same save. Since `childChanged` is `false`, the Workbench never posts `SetProficiencyModifiers`/`SetProficiencyRewards` — only the identity `Edit` goes out, and it committed unconditionally. The cap saved as 5 while the server kept a payout at level 7 that could never fire, with no error anywhere.

## Changes

- `Game.DataAccess/Repositories/Admin/AdminProficiencies.cs`: `SaveProficiencies` now rejects an identity `Edit` whose new `MaxLevel` falls below the highest level of a modifier or reward the proficiency **already has persisted** (checked against the currently-cached child rows, mirroring the scope of the existing `FindLevelOutOfRange` check on the child setters). The rejection fires before anything stages, consistent with this repository's other identity-level guards (path/ordinal, retirement).
- `docs/backend-admin.md`: documents the new guard alongside the other identity-level validation notes.

## Note for reviewers

The issue's fix shape offered two options ("Validate MaxLevel against existing child levels in SaveProficiencies (backend, preferred), and/or make the out-of-range warning block save in the editor"). I implemented the backend guard only — it's the preferred fix and it closes the actual defect (a silent save with no error). The Workbench's advisory warning already flags an out-of-range level as a triangle before save; I left it advisory rather than blocking, matching the existing pattern for the enemy/zone and path/gateway retirement guards documented in `docs/backend-admin.md` ("the Workbench shows the same rule as a validation warning, but the backend is authoritative").

One side effect worth calling out: if an operator *also* edits the modifier/reward collections in the same save (e.g. removing the offending level-7 row) intending a combined fix, the identity `Edit` is still posted first per the Workbench's save order, so it will now see the *pre-save* child rows and reject up front rather than the previous post-commit rejection from the child setter. This is a stricter but safer failure mode (nothing commits at all, vs. a partial commit that could require a re-seed) — the two-call ordering dance itself (post child removals before the identity edit) is out of scope for this issue and not something the issue asked for.

## Tests

- `Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs`: three new cases — shrinking below a persisted modifier's level is rejected, shrinking below a persisted reward's level is rejected, and shrinking to a cap still above every persisted payout is accepted.

## Test plan

- [x] `dotnet test --no-build --no-restore` — full solution suite passes: 3044/3044 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), including the 3 new cases above.

Closes #1777

---
_Generated by [Claude Code](https://claude.ai/code/session_01NZAj4rFN4rPeGWR4L27yHW)_


ginderjeremiah (commented):
## Review: backend guard rejecting a proficiency `MaxLevel` shrink below a persisted payout (#1777)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating event on my own PR) — the fix closes the actual defect cleanly, is well-tested, and follows the existing identity-guard pattern. No blocking findings.

The change correctly targets the real hole from #1777: the level-range rule lived only on the `Set*` child setters, so a lone identity `Edit` that lowered the cap stranded an out-of-range payout silently. `FindShrunkenMaxLevelViolation` runs up front (before staging), mirrors `FindLevelOutOfRange`'s scope (checks the currently-persisted child rows, `> MaxLevel` so a payout exactly at the cap stays valid), and the empty-collection case is handled by the nullable `.Max()` idiom. It reads only the cached graph (no mutation), consistent with `backend.md`.

### Should-fix (deferred — filed as #1804)
- One-save regression: because the Workbench posts `AddEditProficiencies` (identity) before `SetProficiencyModifiers/Rewards` (`progression-store.svelte.ts` steps 3 → 5), an operator who lowers the cap **and** removes the offending payout in the *same* save is now rejected up front, where before #1803 it succeeded (identity committed, then the child setter deleted the row against the new cap). Nothing invalid can be saved, so this isn't a blocker — but it forces a two-save dance for a valid end state. The proper fix is a frontend save-order reorder, exactly analogous to sibling PR #1802 for path/prerequisite removals; a backend-only guard genuinely can't see the same-request child changes. The PR body already calls this out and scopes it out, which I agree with — tracking as **#1804** so it isn't lost.

### Nits
- `Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs` — the "accepted" case uses a payout at level 3 with a cap of 5 (a gap). Consider also pinning the boundary (payout at level 5, cap 5 → accepted), since `> MaxLevel` vs `>= MaxLevel` is the exact line the guard walks. Optional.

### Test coverage
- Solid: three integration cases cover reject-below-modifier, reject-below-reward, and accept-above-every-payout, all against real persisted child rows via the cache reload — the right level for a data-tier guard. The `NewProficiency(maxLevel:)` param addition is backward-compatible. Only the boundary nit above is missing.

### What's good
- Tight scoping and an honest reviewer note about the two-endpoint ordering side effect. The guard is placed and shaped consistently with the sibling identity guards (`FindPathViolation`, `FindTierOrdinalCollision`), and `docs/backend-admin.md` is updated in the same spot as the other identity-level validation notes.

CI is still running (backend + e2e in progress) — not merging on this pass.

---

ginderjeremiah:
## Review: Approve

**What it does:** Adds `FindShrunkenMaxLevelViolation` to `SaveProficiencies` so an identity `Edit` that lowers `MaxLevel` below the level of an already-persisted modifier/reward is rejected up front, before anything stages. This closes the silent-save hole in #1777: when the operator doesn't touch the child collections, `childChanged` is false, no `SetModifiers`/`SetRewards` call is posted, and the identity edit previously committed a cap that stranded an out-of-range payout with no error.

**Correctness — verified against the code, not just the PR text:**
- The guard checks against the currently-cached child rows via `LookupProficiency(...).LevelModifiers`/`.LevelRewards` — the same source `SetModifiers`/`SetRewards` read, so those navigations are loaded (no `NotLoadedException` risk).
- Boundary matches `FindLevelOutOfRange`'s `level > MaxLevel`: a payout landing exactly on the new cap is still reachable and accepted. The `int?` + `Concat`/`Max` handles empty collections (returns null → skipped) cleanly, no null-forgiving.
- Runs only on `Edit` (Add has no persisted children; Delete/retire is irrelevant); returns on first violation across a batch; correctly ordered before `ChangeSetProcessor.Apply` so it rejects before staging, consistent with the path/ordinal and existence guards.
- Style is clean: braces everywhere, no `!`, nullable pattern-matching, XML doc explaining scope. This is the issue's explicitly preferred fix shape.

**Tests:** Four integration cases (not three, as the description says) — below-modifier reject, below-reward reject, above-every-payout accept, and exactly-on-cap accept. Good edge coverage, including the boundary.

**Behavior change worth calling out (not blocking):** The combined-edit flow now fails where it previously succeeded. Before this PR, lowering the cap *and* removing the offending payout in the same save worked — the identity edit committed, then the child setter accepted the reduced desired set. Now the identity edit is posted first and sees the *pre-save* child rows, so it rejects, and the Workbench never posts the child removal. The end state stays consistent (nothing commits), and the error message is clear, but the operator must now do it in two saves. The PR documents this and scopes the two-call ordering out; I agree it's out of scope for #1777, but it's a real workflow regression and deserves a follow-up rather than being left as an undocumented sharp edge in the editor.

**Nits:**
- PR description says "three new cases"; the diff has four (the exact-on-cap boundary test was added). Stale copy, not a code issue.
- DRY: the `level > cap` boundary is duplicated rather than shared with `FindLevelOutOfRange` (which keys off the entity's own `MaxLevel`, not the incoming new cap, so it can't be reused as-is). Fine as written; a small refactor to parameterize the cap could unify them if you feel strongly.

**CI:** Green — `all-checks-passed`, `test-backend`, `verify-backend`, and e2e all succeeded on `df7357d`. (`get_status` shows no legacy commit statuses; the check-runs API is the source of truth here.)

No must-fix items. Recommend a follow-up for the combined lower-cap + prune-payout ordering (either reorder the Workbench save to post child deletions before the identity edit, or promote the editor's advisory out-of-range warning to a blocking one — the "and/or editor" half of the issue's fix shape that this PR intentionally left undone).

---

ginderjeremiah:
Follow-up filed for the deferred combined-save regression (lower `MaxLevel` + prune the offending payout in one save now needs two saves): #1827. Merging this PR now — the actual defect from #1777 (silent out-of-range save) is closed and the deferral is data-safe.

---

PR #1806, opened 2026-07-10
Progression editor shell: fix infinite spinner on load failure and hidden save bar in Map view

### LINKED ISSUE #1779: Progression editor shell: initial load failure leaves an infinite spinner; Map view hides the save bar while changes are pending
Two small shell defects in the progression editor, same component pair:

1. **Infinite spinner on load failure.** `progression-store.svelte.ts:55-66` catches a failed `load()` into a transient toast and leaves `loaded = false`; `Progression.svelte:37-38` then renders `<Loading loading={true}>` forever with no retry affordance. The Ops consoles already handle this correctly (persistent error panel + Refresh button, `DeadLetterConsole.svelte:61-63`) — mirror that.
2. **Save bar only exists in List view.** `Progression.svelte:39-41` vs `:62-99` — the Save/Discard bar renders only in the `list` branch; in Map view the header still counts "N unsaved" but there's no way to save or discard without toggling views. No data risk (the `workbenchDirty` guard still protects tool switches) — move the save bar outside the view branch.

ginderjeremiah:
## Summary

Two small shell defects in the progression editor (`UI/src/routes/admin/workbench/progression/`):

1. **Infinite spinner on load failure.** `progression-store.svelte.ts`'s `load()` caught a failed fetch into a transient toast and left `loaded = false` forever — `Progression.svelte` then rendered the loading spinner with no retry affordance. The Ops consoles (`DeadLetterConsole.svelte`) already handle this correctly with a persistent error panel plus a Refresh button; this PR mirrors that pattern.
2. **Save bar only existed in List view.** The Save/Discard bar rendered only in the `list` branch of `Progression.svelte`; in Map view the header still counted "N unsaved" but there was no way to save or discard without toggling back to List view.

## Changes

- `progression-store.svelte.ts`: added an `error` state field. `load()` now clears it on each attempt and sets it (alongside the existing toast) on failure, instead of only toasting.
- `Progression.svelte`:
  - Template now checks `store.error` first and renders a persistent error panel (`role="alert"`, `data-testid="progression-error"`) with the error message and a Refresh button that re-invokes `store.load()`, before falling back to the spinner (only shown while genuinely still loading) or the loaded content.
  - The save bar moved outside the `view === 'map'` / list branch so it renders once loaded regardless of which view is active — Map view's `.map` element already has `flex: 1; min-height: 0`, so it sits correctly above the save bar in the existing flex-column shell.
  - Added `.prog-error` styling mirroring the Ops console's `.dl-error` treatment.

## Tests

- `progression-store.test.ts`: new cases — a failed initial load sets `store.error` (and still leaves `loaded` false), and a subsequent successful retry clears `store.error` and completes the load.
- `Progression.test.ts`: new cases — a failed load shows the persistent error panel (not the spinner) and recovers once "Refresh" is clicked; switching to Map view with a pending change keeps the save bar (and its save button) visible.

## Test plan

- [x] `npm run test:unit:ci` (`vitest run`) — full suite passes: 3557/3557 (including the 4 new cases above).

Closes #1779

---
_Generated by [Claude Code](https://claude.ai/code/session_0176tckGs4mxeJGSKPq2jidp)_

### DIFF
--- a/UI/src/routes/admin/workbench/progression/Progression.svelte
+++ b/UI/src/routes/admin/workbench/progression/Progression.svelte
@@ -34,30 +34,37 @@
 		</div>
 	</div>
 
-	{#if !store.loaded}
-		<div class="prog-loading"><Loading loading={true} delay={120} /></div>
-	{:else if view === 'map'}
-		<ProgressionMap {store} onNavigate={() => (view = 'list')} />
-	{:else}
-		<div class="prog-body">
-			<ProgressionList {store} />
-
-			<div class="detail-col">
-				{#if store.drilledTier}
-					<TierDetail {store} />
-				{:else if store.selectedPath}
-					<PathDetail {store} />
-				{:else}
-					<div class="empty-detail" data-testid="progression-empty">
-						<div class="glyph"><WorkbenchIcon kind="rune" size={22} /></div>
-						<div class="et">No path selected</div>
-						<button type="button" class="btn primary sm" onclick={() => store.addPath()}>
-							<WorkbenchIcon kind="plus" size={12} />New path
-						</button>
-					</div>
-				{/if}
-			</div>
+	{#if store.error}
+		<div class="prog-error" role="alert" data-testid="progression-error">
+			<p>{store.error}</p>
+			<button type="button" class="btn" onclick={() => store.load()}>Refresh</button>
 		</div>
+	{:else if !store.loaded}
+		<div class="prog-loading"><Loading loading={true} delay={120} /></div>
+	{:else}
+		{#if view === 'map'}
+			<ProgressionMap {store} onNavigate={() => (view = 'list')} />
+		{:else}
+			<div class="prog-body">
+				<ProgressionList {store} />
+
+				<div class="detail-col">
+					{#if store.drilledTier}
+						<TierDetail {store} />
+					{:else if store.selectedPath}
+						<PathDetail {store} />
+					{:else}
+						<div class="empty-detail" data-testid="progression-empty">
+							<div class="glyph"><WorkbenchIcon kind="rune" size={22} /></div>
+							<div class="et">No path selected</div>
+							<button type="button" class="btn primary sm" onclick={() => store.addPath()}>
+								<WorkbenchIcon kind="plus" size={12} />New path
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Progression editor shell — infinite spinner on load failure + hidden save bar in Map view (closes #1779)

**Verdict:** Approve — both shell defects from the issue are fixed correctly, the pattern reuse matches the established Ops-console precedent, and coverage is solid. (Submitted as a comment since I'm the PR author; treat it as an approval.)

Both defects are addressed exactly as the issue specified:

1. **Infinite spinner** — `progression-store.svelte.ts` now carries an `error` state, cleared at the start of every `load()` and set (alongside the existing toast) on failure. `Progression.svelte` checks `store.error` first and renders a persistent `role="alert"` panel with a Refresh button that re-invokes `store.load()`, falling through to the spinner only while genuinely still loading. This mirrors the `DeadLetterConsole.svelte` error-panel + `--error` token treatment faithfully.
2. **Hidden save bar** — the save bar moved out of the list-only branch so it renders once loaded regardless of view. I confirmed `ProgressionMap`'s `.map` root already has `flex: 1; min-height: 0`, so it sits correctly above the save bar in the flex-column shell as claimed.

### Test coverage

Good and targeted:
- `progression-store.test.ts` — failed initial load sets `error` and stays unloaded; a successful retry clears `error` and completes the load.
- `Progression.test.ts` — failed load shows the error panel (not the spinner) and recovers on Refresh; switching to Map view with a pending change keeps the save bar visible (directly exercises the fix, not just the outcome). The `ResizeObserver` stub for `ProgressionMap` on mount is a reasonable jsdom accommodation.

### What's good

- Reuses the existing Ops-console error affordance and `--error` theme token rather than inventing new colors — consistent with `frontend.md` theming guidance.
- The new `.prog-error` style is a distinct centered full-region layout (vs. the DeadLetterConsole inline banner), so a separate rule is justified — not a DRY miss.
- Scope is tight and matches the issue; no over-reach.

No blocking findings, no should-fixes, no nits. CI is green except `test-e2e (webkit)` still in progress at review time; will merge once it completes.

---

PR #1830, opened 2026-07-10
Key dead-letter rows by refresh generation, not queue position

### LINKED ISSUE #1780: Ops: DeadLetterRow expanded state leaks to a different entry across refreshes (rows keyed by queue position)
## Problem

`UI/src/routes/admin/ops/DeadLetterConsole.svelte:114` keys the rows by `entry.index` — the queue *position*, not an entry identity — and `DeadLetterRow.svelte:62` holds `expanded` locally. After a replay/refresh shifts the queue, the row component at position k is reused, so a payload panel the operator expanded stays open showing a **different** entry's payload.

Display-only confusion (selection is cleared in `load()`, and the backend snapshot filter makes stale replays harmless), but an operator inspecting dead letters is exactly who shouldn't be looking at the wrong payload.

## Fix shape

Collapse expansion on refresh, or key rows by an entry identity (payload hash + occurrence) instead of position.

ginderjeremiah:
## Summary

`UI/src/routes/admin/ops/DeadLetterConsole.svelte:114` keyed each row's `DeadLetterRow` component by `entry.index` — the queue *position*, not an entry identity — while `DeadLetterRow.svelte:62` holds `expanded` as local component state. After a replay/refresh shifts the queue, the row component at position `k` is reused for whatever entry now sits there, so a payload panel an operator had expanded stayed open showing a **different** entry's payload.

An operator expands the payload for the entry at position 0, then clicks Refresh (or replays some entries and the queue re-shuffles). A different entry now occupies position 0, but its row reuses the previous row's component state — the payload panel stays expanded, silently showing the wrong entry's payload to an operator who is specifically trying to inspect dead letters carefully.

## Changes

- `dead-letters.svelte.ts`: `DeadLetterConsoleState` gains a `generation` counter, bumped on every successful `load()`.
- `DeadLetterConsole.svelte`: the keyed `{#each}` now keys rows by `` `${queue.generation}-${entry.index}` `` instead of `entry.index` alone, so every refresh mounts fresh row components (collapsing any stale expansion) — consistent with how the console already clears the operator's selection on every `load()`.

I went with "collapse expansion on refresh" over the issue's alternative ("key by payload hash + occurrence") because the codebase already established that pattern for selection (see the comment on `DeadLetterConsoleState`'s selected set docblock, which deliberately avoids payload-based keying since duplicate payloads exist and complicate identity). Keying by payload + occurrence would also need special-casing duplicate payloads to stay unique per Svelte's keyed-each requirement, for comparatively little benefit — a plain "Refresh" is cheap enough that remounting up to 200 rows is a non-issue.

## Tests

- `dead-letters.test.ts`: new case pins that `generation` increments only on a successful `load()`, not a failed one.
- `DeadLetterConsole.test.ts`: new case expands a row, refreshes with a different entry at the same queue position, and asserts the payload panel is no longer shown (i.e., it collapsed rather than leaking).

## Test plan

- [x] `npm run lint` — clean.
- [x] `npx vitest run` (scoped: dead-letter suites) — 33/33 passing.
- [x] `npm run test:unit:ci` (full suite) — 3566/3566 passing.

Closes #1780

---
_Generated by [Claude Code](https://claude.ai/code/session_01TP8U98R4ZXcnEER7gGXn7i)_

### DIFF
--- a/UI/src/routes/admin/ops/DeadLetterConsole.svelte
+++ b/UI/src/routes/admin/ops/DeadLetterConsole.svelte
@@ -111,7 +111,7 @@
 					</tr>
 				</thead>
 				<tbody>
-					{#each queue.entries as entry (entry.index)}
+					{#each queue.entries as entry (`${queue.generation}-${entry.index}`)}
 						<DeadLetterRow
 							{entry}
 							selected={queue.isSelected(entry.index)}
--- a/UI/src/routes/admin/ops/dead-letters.svelte.ts
+++ b/UI/src/routes/admin/ops/dead-letters.svelte.ts
@@ -91,6 +91,13 @@ export class DeadLetterConsoleState {
 	error = $state<string | null>(null);
 	readonly selected = new SvelteSet<number>();
 
+	/**
+	 * Bumped on every successful {@link load}. `entry.index` is only a queue position, reused by whatever
+	 * entry now sits there, so the console keys rows by `generation` + `index` rather than `index` alone —
+	 * otherwise a row's local expanded-payload state would leak onto a different entry after a refresh.
+	 */
+	generation = $state(0);
+
 	private readonly routes: QueueRoutes;
 
 	constructor(variant: DeadLetterQueueVariant = 'player-update') {
@@ -160,6 +167,7 @@ export class DeadLetterConsoleState {
 			});
 			this.totalCount = inspection.totalCount;
 			this.entries = inspection.entries;
+			this.generation++;
 			this.clearSelection();
 			this.loaded = true;
 			return true;
--- a/UI/src/tests/routes/admin/ops/DeadLetterConsole.test.ts
+++ b/UI/src/tests/routes/admin/ops/DeadLetterConsole.test.ts
@@ -76,6 +76,22 @@ describe('DeadLetterConsole', () => {
 		expect(screen.queryByTestId('dl-empty')).toBeNull();
 	});
 
+	it('collapses an expanded row on refresh instead of leaking it onto whatever entry now sits at that position', async () => {
+		getMock.mockResolvedValueOnce(inspection([entry({ index: 0, rawPayload: 'old-payload' })], 1));
+		render(DeadLetterConsole);
+
+		await waitFor(() => expect(screen.getByTestId('dl-table')).toBeTruthy());
+		await fireEvent.click(screen.getByTestId('dl-row-expand'));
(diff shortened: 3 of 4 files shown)


ginderjeremiah (commented):
## Review: keying dead-letter rows by refresh generation instead of raw queue position (#1780)

**Verdict:** Approve — the fix is correct, minimal, well-tested, and consistent with the console's existing "reset per load" behavior. (Submitted as a `COMMENT` only because GitHub won't accept an `APPROVE` from the PR author; treat this as an approval.)

### Assessment

The root cause is accurately diagnosed: `{#each}` keyed by `entry.index` (a queue *position*) reuses the same `DeadLetterRow` component instance across refreshes, and `DeadLetterRow` holds `expanded` as local `$state` — so an expanded payload panel leaks onto whatever entry now occupies that position after a replay/refresh.

Keying by `` `${queue.generation}-${entry.index}` `` where `generation` bumps on every successful `load()` forces a fresh mount of every row on refresh, collapsing any stale expansion. Verified:

- `generation` is incremented only after the snapshot is successfully replaced (inside the `try`, before `clearSelection()`), so a failed load leaves the keys — and the currently-rendered rows — untouched. The new `dead-letters.test.ts` case pins exactly this.
- The composite key stays unique within a render: `generation` is constant across the render pass and `entry.index` is unique per position, so there's no risk of a duplicate-key error from Svelte's keyed each.
- This is the same "collapse on refresh" shape the issue offered as the simpler of its two options, and it mirrors the console already clearing `selected` on every `load()` — so identity is handled consistently across selection and row state rather than introducing a second, payload-hash-based identity scheme.

The trade-off (an unchanged row also remounts on a no-op Refresh, collapsing an expansion the operator may have wanted to keep) is real but minor, is explicitly acknowledged in the PR description, and is the cheaper/safer choice versus a payload-hash+occurrence identity that would have to special-case duplicate payloads. Good call.

The 3-line docblock on `generation` is warranted — it explains *why* the field exists and prevents a future edit from "simplifying" the key back to `entry.index` and silently reintroducing the bug.

### Test coverage

Meets the frontend testing rules: the interaction is covered at the component level (expand → refresh with a different entry at the same position → payload panel gone) and the state invariant is covered at the unit level (generation bumps on success, not on failure), both under `tests/` mirroring the source path. Nothing further needed for a change this size.

### What's good

Tight, correct fix with the reasoning captured in both the docblock and the PR description, and tests that assert the actual user-facing behavior rather than just the counter. No blocking items, should-fixes, or nits.


---

PR #1831, opened 2026-07-10
Fix KillsByDamageType challenge defaulting to a backend-rejected Global scope

### LINKED ISSUE #1781: Workbench: KillsByDamageType challenge authors as "Global" by default, which the backend hard-rejects — failing the whole challenge batch with no client warning
## Problem

Switching a challenge's objective type runs `deriveFromType`, which clears `targetEntityId` (`UI/src/routes/admin/workbench/entities/challenge-helpers.ts:162-168`), and `ScopeField.svelte:24` lands on Global — so the *default* state after picking "Kills By Damage Type" is exactly what the backend rejects (`Game.DataAccess/Repositories/Admin/AdminChallenges.cs:165-170`, "must target a damage-type key", rejecting the entire `AddEditChallenges` batch up front). The condition section's `warn` only checks `progressGoal` (`entities/challenge.ts:91`), so no warning triangle appears; the record reads valid until Save 400s with only a toast, taking any sibling adds/edits in the batch down with it.

Every other cheap, local authoring rule surfaces as a client warning (the documented pattern) — this one is the exception, on the default state.

## Fix shape

Add the rule to the condition section's `warn` (`challengeTypeId === EChallengeType.KillsByDamageType && targetEntityId == null`), or better, have `deriveFromType`/ScopeField default that type to Specific with the first damage-type key.

ginderjeremiah:
## Summary

Switching a Workbench challenge's objective type to "Kills By Damage Type" ran `deriveFromType`, which clears `targetEntityId` (`UI/src/routes/admin/workbench/entities/challenge-helpers.ts`), and `ScopeField.svelte` lands on **Global** whenever a target is unset — so the *default* state right after picking that type was exactly what the backend hard-rejects (`Game.DataAccess/Repositories/Admin/AdminChallenges.cs:165-170`, "must target a damage-type key"), failing the entire `AddEditChallenges` batch (including any sibling adds/edits) with only a toast and no warning triangle.

## Changes

- `challenge-helpers.ts`: `deriveFromType` now defaults the target to the first `EDamageTypeKey` (`Physical`) instead of `undefined` whenever the newly-selected type's statistic is `DamageType`-scoped, since that dimension has no global row (mirrors the backend's own reasoning: "the statistic writes only per-damage-type-key rows, no global row"). Every other entity dimension keeps defaulting to Global as before — this is metadata-driven off `entityType`, not hardcoded to `KillsByDamageType`, so it stays correct if another `DamageType`-scoped statistic is ever added.
- `challenge.ts`: the condition section's `warn` gains a matching check (`entityType === EEntityType.DamageType && targetEntityId == null`) as a backstop — an admin can still manually click "Global" in `ScopeField` after picking the type, and this closes that path with the same amber-triangle pattern the codebase already uses for every other cheap, local authoring rule (see `docs/frontend-admin.md` → "Validation lives where it occurs").

## Tests

- `challenge-helpers.test.ts`: new case pinning that `deriveFromType` defaults a `KillsByDamageType` switch to damage-type key `0` instead of clearing the target.
- `entities/challenge.test.ts`: new cases for the condition section's `warn` — non-positive goal, an untargeted `DamageType` statistic, a targeted one, and a non-`DamageType` statistic left untargeted (which is fine).

## Test plan

- [x] `npx vitest run` (full suite) — 296 files / 3571 tests passing.

Closes #1781

---
_Generated by [Claude Code](https://claude.ai/code/session_01XZ9ZgY1rWmESeEqztrnYfy)_

### DIFF
--- a/UI/src/routes/admin/workbench/entities/challenge-helpers.ts
+++ b/UI/src/routes/admin/workbench/entities/challenge-helpers.ts
@@ -1,6 +1,7 @@
 import {
 	EChallengeGoalComparison,
 	EChallengeType,
+	EDamageTypeKey,
 	EEntityType,
 	EStatisticType,
 	type IChallenge,
@@ -164,5 +165,8 @@ export function deriveFromType(c: IChallenge, types: IChallengeType[], typeId: E
 	c.challengeTypeId = typeId;
 	c.statisticType = stat?.id;
 	c.entityType = stat?.entityType ?? EEntityType.None;
-	c.targetEntityId = undefined;
+	// A DamageType-scoped statistic writes only per-damage-type-key rows (no global row), so
+	// Global is never a valid target for it — default to the first key rather than landing on
+	// the state the backend hard-rejects. Every other dimension defaults to Global as before.
+	c.targetEntityId = c.entityType === EEntityType.DamageType ? EDamageTypeKey.Physical : undefined;
 }
--- a/UI/src/routes/admin/workbench/entities/challenge.ts
+++ b/UI/src/routes/admin/workbench/entities/challenge.ts
@@ -86,9 +86,18 @@ export const challengeEntity: EntityConfig<IChallenge> = {
 			desc: 'What the player must do to complete it',
 			kind: 'challenge-condition',
 			// Statistic & entity are fully determined by the type, so they can't be
-			// inconsistent — the only condition-level slip is a non-positive goal.
+			// inconsistent — the remaining condition-level slips are a non-positive goal and a
+			// DamageType-scoped statistic left Global (it has no global row; the backend hard-rejects it).
 			dirtyKeys: ['challengeTypeId', 'statisticType', 'entityType', 'targetEntityId', 'progressGoal'],
-			warn: (c) => (!c.progressGoal || c.progressGoal <= 0 ? 'Goal must be greater than zero' : null)
+			warn: (c) => {
+				if (!c.progressGoal || c.progressGoal <= 0) {
+					return 'Goal must be greater than zero';
+				}
+				if (c.entityType === EEntityType.DamageType && c.targetEntityId == null) {
+					return 'Must target a damage type — this statistic has no global total';
+				}
+				return null;
+			}
 		},
 		{
 			key: 'reward',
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Fix KillsByDamageType challenge defaulting to a backend-rejected Global scope (#1781)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't accept a gating event from me) — the fix correctly removes the invalid default state and mirrors the backend's rejection with a client-side warning, following the documented admin-validation pattern.

Verified the change against the code, not just the diff:

- `deriveFromType` now defaults `targetEntityId` to `EDamageTypeKey.Physical` (which is `0`, the first key — `enums.ts:162`) for any `DamageType`-scoped statistic, so the default state after picking "Kills By Damage Type" is no longer the exact state `AdminChallenges.FindTargetViolation` hard-rejects (`AdminChallenges.cs:167-170`). Keying off `entityType === EEntityType.DamageType` (metadata-driven) rather than hardcoding `KillsByDamageType` is the right call — it stays correct if another `DamageType`-scoped statistic is ever added, and it reads from the same source of truth the rest of the helper does.
- The condition-section `warn` backstop matches the documented "Validation lives where it occurs" pattern (`docs/frontend-admin.md`) and correctly closes the one remaining path to the bad state — an admin manually clicking **Global** in `ScopeField` (`setGlobal` → `targetEntityId = undefined`). `setSpecific` lands on `options[0]?.value` (Physical), so toggling Specific is consistent with `deriveFromType`'s default.
- Note: the frontend `warn` generalizes to all `DamageType` scopes while the backend's null-check is narrowly `ChallengeTypeId == KillsByDamageType`. These align today (KillsByDamageType is the only DamageType-scoped statistic), and the generalization is the more-correct direction given "damage-type stats have no global row" — not a divergence to worry about.

### Test coverage

Sufficient and follows the project's frontend rules — unit tests mirror the source path, covering both changed pure functions:
- `challenge-helpers.test.ts`: pins the new `KillsByDamageType` default to key `0` (the `TYPES` fixture maps it to a `DamageType`-scoped statistic, so the assertion is real).
- `challenge.test.ts`: four `warn` cases — non-positive goal, untargeted `DamageType` (warns), targeted `DamageType` (passes, and `0 == null` is correctly false), and untargeted non-`DamageType` (passes).

### What's good

- Correctly identified that a valid default (not just a warning) is the better fix, so the common path no longer produces a broken record at all.
- The inline comments explain the *why* (no global row for damage-type stats) concisely, which is exactly the non-obvious context a future reader needs.

Frontend-only change; no backend or battle-logic mirror needed. Clean PR.

---

PR #1832, opened 2026-07-10
Filter class starter-equipment item picker by slot category

### LINKED ISSUE #1782: Workbench: class starter-equipment item picker is unfiltered by slot — backend rejects post-commit and the re-seed discards all unsaved edits
## Problem

- `UI/src/routes/admin/workbench/entities/class.ts:165-171` — a new starter-equipment row defaults to `itemId: reference.itemOptions()[0]` regardless of the auto-picked slot, and the Item column (`:181`) uses the unfiltered `reference.itemOptions`; the section has no `warn`.
- `Game.DataAccess/Repositories/Admin/AdminClasses.cs:229-248` — `FindStarterEquipmentViolation` rejects any item whose category doesn't match the slot.
- Because `SetClassStarterEquipment` is a **child saver**, the rejection lands *after* the identity batch commits: `save-helpers.ts:132-156` + `entity-store.svelte.ts:208-223` turn it into a `PersistFailedError` → `save()` re-seeds `items`/`base` from the server, silently discarding every unsaved local edit in the store behind a toast.

## Failure scenario

Admin renames a class, clicks "Add equipment" (Helm slot, first catalogue item is a weapon), saves → rename commits, starter-equipment call 400s, re-seed wipes the admin's other pending edits.

## Fix shape

Filter the Item column's options by the row's `equipmentSlot` → `EItemCategory` mapping and pick the first *category-matching* item in `newRow`; add a section `warn` for slot/category mismatch as a backstop. (The unsaved-edit-loss escalation on post-commit child failures is a general store behavior — worth noting on the fix, but the authoring-time guard removes this trigger.)

ginderjeremiah:
## Summary

`UI/src/routes/admin/workbench/entities/class.ts`'s starter-equipment row picker let an admin assign any item to any equipment slot, regardless of category. The backend's `FindStarterEquipmentViolation` (`Game.DataAccess/Repositories/Admin/AdminClasses.cs:229-248`) only rejects a slot/category mismatch when the `SetClassStarterEquipment` **child saver** runs — which is *after* the identity save has already committed. That turns the rejection into a `PersistFailedError`, and `EntityStore.save()`'s post-commit re-seed then discards every other unsaved edit in the store.

Admin renames a class, clicks "Add equipment" for the Helm slot (the first catalogue item happens to be a weapon), saves → the rename commits, the starter-equipment call 400s, and the re-seed silently wipes the admin's other pending edits.

## Changes

- `ColumnConfig.options` (`entities/types.ts`) now also receives the full row, not just the column's current value — mirroring the `shareTotal` column callback's existing `(row, rows, record)` shape. `TableCell.svelte` passes the row through at both call sites (`select` and `attribute`). This is additive/backward-compatible: existing zero/one-arg `options` callbacks are unaffected.
- `reference.svelte.ts`: adds `equipmentSlotCategory` (mirrors the backend's `EquipmentSlot.ItemCategory` switch), `itemOptionsForSlot` (an item picker narrowed to the category a slot accepts, keeping an already-assigned mismatched item visible the same way retired references stay visible elsewhere), and `itemCategoryOf`.
- `class.ts`: the starter-equipment table's Item column now filters its options by the row's `equipmentSlot`; `newRow` picks the first *category-matching* item for the auto-picked slot instead of the first catalogue item overall; the section gets a `warn` that flags any already-authored slot/category mismatch as a backstop.

## Not included

The general "a post-commit child-saver failure re-seeds the whole store, discarding unrelated unsaved edits" behavior is a store-wide design tradeoff, not specific to this entity — the issue itself scoped that out, and a related persistEntity commit-classification bug was already fixed separately in #1634. This PR removes the *trigger* for that path in the starter-equipment case rather than changing the store's re-seed behavior itself.

## Tests

- `reference.test.ts`: `equipmentSlotCategory` covers all six slots; `itemOptionsForSlot` narrows by category and keeps a mismatched current value visible; `itemCategoryOf` resolves an item's category.
- `class.test.ts`: `newRow` picks a category-matching item for the auto-picked slot (including the case where the first catalogue item is the wrong category); the Item column's `options` is filtered per row; the section `warn` is null when consistent and flags a mismatch otherwise.
- `TableCell.test.ts`: the select cell's `options` provider is called with both the current value and the full row.

## Test plan

- [x] `npm run lint` — clean.
- [x] `npx vitest run` (`test:unit:ci`) — full suite passes: 3581/3581.

Closes #1782

---
_Generated by [Claude Code](https://claude.ai/code/session_011ZGYPBuhYFo5zAFKSKSAxQ)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/TableCell.svelte
+++ b/UI/src/routes/admin/workbench/components/TableCell.svelte
@@ -2,7 +2,7 @@
 	<td style:min-width="{col.min ?? 160}px">
 		<div class="fld">
 			<select class="sel" class:dirty value={row[col.key] as number} onchange={(e) => onChange(+e.currentTarget.value)}>
-				{#each col.options?.(row[col.key] as number) ?? [] as option (option.value)}
+				{#each col.options?.(row[col.key] as number, row) ?? [] as option (option.value)}
 					<option value={option.value} disabled={taken.has(option.value) && option.value !== row[col.key]}>
 						{option.text}
 					</option>
@@ -16,7 +16,7 @@
 	<td style:min-width="{col.min ?? 160}px">
 		<AttributePicker
 			value={row[col.key] as number}
-			options={col.options?.(row[col.key] as number) ?? []}
+			options={col.options?.(row[col.key] as number, row) ?? []}
 			onChange={(v) => onChange(v)}
 			ariaLabel={col.label}
 			disabledValues={col.unique ? taken : undefined}
--- a/UI/src/routes/admin/workbench/entities/class.ts
+++ b/UI/src/routes/admin/workbench/entities/class.ts
@@ -1,4 +1,12 @@
-import { ApiRequest, EAttribute, EModifierType, ESkillAcquisition, fetchSocketData, type IClass } from '$lib/api';
+import {
+	ApiRequest,
+	EAttribute,
+	type EEquipmentSlot,
+	EModifierType,
+	ESkillAcquisition,
+	fetchSocketData,
+	type IClass
+} from '$lib/api';
 import { hasFlag } from '$lib/common';
 import { staticData } from '$stores';
 import { reference } from '../reference.svelte';
@@ -155,6 +163,12 @@ export const classEntity: EntityConfig<WorkbenchClass> = {
 			glyph: 'box',
 			desc: 'Items equipped at character creation',
 			count: (c) => c.starterEquipment.length,
+			warn: (c) => {
+				const mismatch = c.starterEquipment.find(
+					(e) => reference.itemCategoryOf(e.itemId) !== reference.equipmentSlotCategory(e.equipmentSlot)
+				);
+				return mismatch ? `Item doesn't match its equipment slot's category` : null;
+			},
 			kind: 'table',
 			itemsKey: 'starterEquipment',
 			rowKey: 'equipmentSlot',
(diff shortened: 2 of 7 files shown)


ginderjeremiah (commented):
## Review: Filter class starter-equipment item picker by slot category (#1782)

**Verdict:** Approve — the fix is correctly scoped to remove the authoring-time trigger, mirrors the backend faithfully, follows house conventions, and is well-tested. _(Posted as a comment since I'm the PR author; GitHub won't accept a gating verdict on one's own PR.)_

I traced the change against the issue, the backend it mirrors, and the frontend docs. It does what it claims and nothing more.

### What's good

- **Faithful backend mirror.** `equipmentSlotCategories` matches `EquipmentSlot.GetItemCategory` (`Game.Core/Players/Inventories/EquipmentSlot.cs:68-80`) slot-for-slot, and typing it as `Record<EEquipmentSlot, EItemCategory>` means adding a slot to the enum forces this map to be updated at compile time — a good guard against silent drift.
- **Honours house conventions.** `itemCategoryOf` uses the documented index-by-id lookup for the zero-based-id item catalogue (`frontend.md` → Reference Data), and `itemOptionsForSlot` reuses the existing retired-`keep` pattern so an already-assigned mismatched item stays visible instead of vanishing from the picker — consistent with how retired refs are handled elsewhere.
- **Backward-compatible signature change.** Widening `ColumnConfig.options` to `(current?, row?)` is additive; existing zero/one-arg callers are unaffected, and `TableCell` threads the row through at both the `select` and `attribute` call sites.
- **Right call on scope.** Removing the *trigger* rather than reworking the store-wide post-commit re-seed behaviour is the correct boundary — the issue scoped that out and it's a separate design unit.

### Test coverage

Solid and matches the project's mirror-the-source-path rule:
- `reference.test.ts` — the full slot→category map, category narrowing, mismatched-`keep` visibility, and `itemCategoryOf` (including the out-of-range `undefined`).
- `class.test.ts` — `newRow` picks a category-matching item (incl. the wrong-first-catalogue-item case), per-row option filtering, and both `warn` branches.
- `TableCell.test.ts` — the options provider now receives `(current, row)`.

### Nits

None blocking. One observation, not a change request: when a slot has no catalogue items of its category, `newRow` falls back to `itemId: 0`, which can be a real item of the wrong category (the `?? 0` fallback predates this PR). The section `warn` is the intended backstop and the picker will surface the mismatch, so this is acceptable as-is.


---

PR #1833, opened 2026-07-10
Fix lesson-step table ordinal collision and false-dirty Anchor Key

### LINKED ISSUE #1784: Workbench lesson-step table: ordinal collision breaks the keyed each; cleared optional Anchor Key sticks false-dirty
Two small defects in the lesson-step editing table:

1. **Ordinal collision breaks the keyed `{#each}` before the warning can help.** `SectionTable.svelte:22` keys rows by `row[rowKey]`, and the lesson steps table's rowKey is the author-editable `ordinal` (`entities/lesson.ts:117-133` — whose `warn` comment claims the duplicate "is caught here rather than left to blow up"). The warning is advisory: the moment `setCell` commits a duplicate ordinal, dev builds throw `each_key_duplicate` (editor crashes before the triangle renders) and prod builds mis-reconcile rows (edits/removes can hit the wrong row) until fixed. Fix: reject/swap on collision, or key step rows by a client-side surrogate id like the `id`-keyed tables.
2. **Cleared Anchor Key marks the step modified forever.** Text cells write back `''` (`TableCell.svelte:46-48`) while the server omits the optional `anchorKey`, and `canonicalize` (`save-helpers.ts:10-32`) collapses null/absent but not `''` — so typing into an existing step's Anchor Key then erasing it leaves a phantom dirty state and a redundant `SetLessonSteps` post. Fix: normalize `''` → absent on edit or in `canonicalize` for this column.

ginderjeremiah:
## Summary

Two small defects in the Workbench lesson-step editing table (`UI/src/routes/admin/workbench/entities/lesson.ts`'s "Tour Steps" section):

1. **Ordinal collision broke the keyed `{#each}`.** `SectionTable.svelte` keys rows by `row[rowKey]`, and the lesson steps table's `rowKey` is the author-editable `ordinal` — the only table in the Workbench where the identity column is also a freely-editable number (every other table's `rowKey` is either a surrogate id or a `unique` select/attribute the UI already keeps collision-free). The moment `setCell` committed a duplicate ordinal, dev builds threw `each_key_duplicate` and prod builds could mis-reconcile rows.
2. **A cleared optional Anchor Key stuck the row false-dirty.** Text cells wrote back `''`, the server omits the optional `anchorKey` when unset, and `canonicalize` (`save-helpers.ts`) collapses `null`/absent but not `''` — so clearing an existing step's Anchor Key left a phantom dirty state and triggered a redundant `SetLessonSteps` post.

## Changes

- `SectionTable.svelte`: `setCell` now detects when an edit to the row's identity column (`rowKey`) would collide with a sibling row, and swaps the sibling onto the vacated identity instead of allowing the duplicate. This is implemented generically (keyed on `key === rowKey`), not lesson-specific — it's a no-op for every other table today since their `rowKey` columns are already collision-proof by construction, but it protects any future editable-identity column the same way.
- `entities/types.ts`: new `ColumnConfig.optional` flag for text columns, and a `TableRow` type alias (`Record<string, number | string | undefined>`) replacing the repeated inline type across the table components — needed since an optional column's cleared value is now `undefined` rather than `''`.
- `TableCell.svelte`: a text column's `oninput` now stores `undefined` instead of `''` when `col.optional` is set and the field is cleared.
- `lesson.ts`: marks the Anchor Key column `optional: true`.
- `reference.svelte.ts`: `enemySpawnShareTotal`'s signature updated to the new shared `TableRow` type (mechanical, no behavior change).

## Tests

- `SectionTable.test.ts`: new `describe('SectionTable — editable-identity (rowKey) collections')` pinning that editing a row's ordinal into a sibling's swaps identities rather than duplicating the key.
- `TableCell.test.ts`: new cases confirming an `optional` text column's cleared input calls `onChange(undefined)`, while a non-optional column still calls `onChange('')`.
- `lesson.test.ts`: new case asserting the Anchor Key column carries `optional: true`.

## Test plan

- [x] `npx vitest run` (targeted + full suite) — 3585/3585 passing, including the new cases above.

Closes #1784

---
_Generated by [Claude Code](https://claude.ai/code/session_019vifMzWzFn6PSKhnq1FYoA)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/SectionTable.svelte
+++ b/UI/src/routes/admin/workbench/components/SectionTable.svelte
@@ -79,7 +79,7 @@
 import WorkbenchIcon from '../WorkbenchIcon.svelte';
 import type { EntityStore } from '../entity-store.svelte';
 import { recordsEqual } from '../entity-store.svelte';
-import { fieldsOf, type Identified, type TableSectionConfig } from '../entities/types';
+import { fieldsOf, type Identified, type TableRow, type TableSectionConfig } from '../entities/types';
 import EmptySection from './EmptySection.svelte';
 import TableCell from './TableCell.svelte';
 
@@ -94,13 +94,13 @@ const { section, record, baseline, store }: Props = $props();
 
 const itemsKey = $derived(section.itemsKey as string);
 const rowKey = $derived(section.rowKey);
-const rows = $derived((fieldsOf(record)[itemsKey] as Record<string, number | string>[]) ?? []);
-const baseRows = $derived(baseline ? (fieldsOf(baseline)[itemsKey] as Record<string, number | string>[]) : null);
+const rows = $derived((fieldsOf(record)[itemsKey] as TableRow[]) ?? []);
+const baseRows = $derived(baseline ? (fieldsOf(baseline)[itemsKey] as TableRow[]) : null);
 
 // Match each row to its baseline by stable identity (not array position), so a mid-list delete
 // shifting indices can no longer compare a row against the wrong baseline.
 const baseByKey = $derived.by(() => {
-	const map: Record<number, Record<string, number | string>> = {};
+	const map: Record<number, TableRow> = {};
 	for (const baseRow of baseRows ?? []) {
 		map[baseRow[rowKey] as number] = baseRow;
 	}
@@ -113,15 +113,15 @@ const noFree = $derived(
 	uniqueCol ? (uniqueCol.options?.() ?? []).every((o) => rows.some((r) => r[uniqueCol.key] === o.value)) : false
 );
 
-const mutateRows = (mutate: (list: Record<string, number | string>[]) => void) => {
+const mutateRows = (mutate: (list: TableRow[]) => void) => {
 	store.patch(record.id, (draft) => {
-		mutate(fieldsOf(draft)[itemsKey] as Record<string, number | string>[]);
+		mutate(fieldsOf(draft)[itemsKey] as TableRow[]);
 	});
 };
 
 const add = () =>
 	store.patch(record.id, (draft) => {
-		const list = fieldsOf(draft)[itemsKey] as Record<string, number | string>[];
+		const list = fieldsOf(draft)[itemsKey] as TableRow[];
 		const row = section.newRow(draft);
 		// Surrogate-id collections mark new rows with id 0; give each a unique negative id so its
 		// identity (and the {#each} key) stays stable and collision-free across several unsaved rows.
@@ -132,7 +132,19 @@ const add = () =>
 		list.push(row);
 	});
 const removeRow = (i: number) => mutateRows((list) => list.splice(i, 1));
-const setCell = (i: number, key: string, value: number | string) => mutateRows((list) => (list[i][key] = value));
+const setCell = (i: number, key: string, value: number | string | undefined) =>
+	mutateRows((list) => {
+		// Editing a row's identity column (e.g. a tour step's author-editable `ordinal`) into another
+		// row's value would duplicate the {#each} key — crashing dev builds and mis-reconciling prod
+		// ones. Swap the two rows' identities instead, so the edit still lands and both keys stay unique.
+		if (key === rowKey) {
+			const collision = list.find((r, ri) => ri !== i && r[rowKey] === value);
+			if (collision) {
+				collision[rowKey] = list[i][rowKey];
+			}
+		}
+		list[i][key] = value;
+	});
 </script>
 
 <style lang="scss">
--- a/UI/src/routes/admin/workbench/components/TableCell.svelte
+++ b/UI/src/routes/admin/workbench/components/TableCell.svelte
@@ -44,7 +44,7 @@
 				aria-label={col.label}
 				placeholder={col.placeholder}
 				value={(row[col.key] as string) ?? ''}
-				oninput={(e) => onChange(e.currentTarget.value)}
+				oninput={(e) => onChange(col.optional ? e.currentTarget.value || undefined : e.currentTarget.value)}
 			/>
 			{#if dirty}<DirtyDot />{/if}
 		</div>
@@ -59,7 +59,7 @@
 {/if}
 
 <script lang="ts">
-import type { ColumnConfig } from '../entities/types';
+import type { ColumnConfig, TableRow } from '../entities/types';
 import Bar from '$components/Bar.svelte';
 import NumInput from './NumInput.svelte';
 import SelectCaret from './SelectCaret.svelte';
@@ -68,12 +68,12 @@ import AttributePicker from './AttributePicker.svelte';
 
 interface Props {
 	col: ColumnConfig;
-	row: Record<string, number | string>;
+	row: TableRow;
 	idx: number;
-	rows: Record<string, number | string>[];
+	rows: TableRow[];
 	record: unknown;
 	dirty: boolean;
-	onChange: (value: number | string) => void;
+	onChange: (value: number | string | undefined) => void;
 }
 
 const { col, row, idx, rows, record, dirty, onChange }: Props = $props();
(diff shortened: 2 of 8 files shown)


UI/src/routes/admin/workbench/components/SectionTable.svelte
```diff
@@ -132,7 +132,19 @@ const add = () =>
...
+		// Editing a row's identity column (e.g. a tour step's author-editable `ordinal`) into another
+		// row's value would duplicate the {#each} key — crashing dev builds and mis-reconciling prod
+		// ones. Swap the two rows' identities instead, so the edit still lands and both keys stay unique.
+		if (key === rowKey) {
+			const collision = list.find((r, ri) => ri !== i && r[rowKey] === value);
+			if (collision) {
+				collision[rowKey] = list[i][rowKey];
+			}
+		}
+		list[i][key] = value;
```
ginderjeremiah, line 146: Nit (non-blocking): the swap preserves the array's ordering, so after editing row 0's `ordinal` from 0→1 the array is `[{ord:1,'First'}, {ord:0,'Second'}]` and the `{#each}` renders them in array order — briefly showing ordinal 1 above ordinal 0 until a save+refresh re-sorts them. Purely cosmetic and inherent to the swap approach (rejecting the edit would have the same "where did my number go" surprise), so I don't think it needs changing here — just flagging the behavior. If it ever feels off in practice, re-sorting the list by `rowKey` after a swap would keep the visual order stable.

---

ginderjeremiah (commented):
## Review: Fix lesson-step ordinal collision + false-dirty Anchor Key (#1784)

**Verdict:** Approve — both defects are fixed correctly and cleanly, with the collision handling implemented generically without changing behavior for any existing table. (Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR; the intended call is **Approve**.)

I traced both fixes against the code:

- **Ordinal collision** — `setCell` now swaps identities on a `key === rowKey` collision. I confirmed every other table's `rowKey` is either the non-editable surrogate `id` (`item.modSlots`, `skill.effects`) or a `unique: true` select/attribute column (`equipmentSlot`, `attributeId`, `type`, `zoneId`, `enemyId`, `proficiencyId`), all of which `TableCell`'s `taken` set already keeps collision-free. So the swap is a genuine no-op everywhere except the lesson `ordinal` case, exactly as the PR claims.
- **False-dirty Anchor Key** — `TableCell` emits `undefined` for a cleared `optional` text cell, and `canonicalize` (`save-helpers.ts:11`) already collapses `undefined`→absent. Since `recordsEqual` is `canonicalEqual`, both the per-cell dirty dot (`SectionTable.svelte:46`) and the `childChanged`→`listsEqual` change detection that gates the redundant `SetLessonSteps` post now read a cleared Anchor Key as unmodified. Correct root-cause fix.

### Nits (all optional, non-blocking)

- `entities/lesson.ts:144` — `newRow` still defaults `anchorKey: ''` rather than `undefined`. Harmless (a new row is always an Add and never diffed against a baseline), but slightly inconsistent with the new "unset optional = `undefined`" convention this PR establishes. Not in the diff, so noting it here rather than inline.
- `SectionTable.svelte` swap — cosmetic display-order note, left inline.
- Consider one test that pins the *actual reported symptom* end-to-end — that a step whose Anchor Key is cleared canonicalizes equal to a baseline omitting `anchorKey` (i.e. is not dirty). The new tests thoroughly cover the two mechanisms in isolation (`TableCell` emits `undefined`; the swap), but nothing asserts the composed false-dirty result that #1784's second bullet is about. `canonicalize`'s `undefined` handling is pre-existing, so the chain holds — this would just be belt-and-suspenders.

### Test coverage

Solid and matches the project's mirroring rule (tests live under `tests/` at the source path). `SectionTable.test.ts` pins the swap-not-duplicate behavior; `TableCell.test.ts` pins `undefined`-vs-`''` for optional vs non-optional; `lesson.test.ts` pins `optional: true` on the column. Frontend-only change, no backend/battle parity implications.

### What's good

- The collision handling is generic (keyed on `rowKey`) rather than lesson-specific, and I verified it's provably a no-op for the current tables — future-proofs any editable-identity column without over-fitting.
- Introducing the shared `TableRow` alias and threading it through the table components removes the repeated inline `Record<...>` type; the added `undefined` in the union is what makes the optional-cleared semantics type-honest.
- Both fixes target the actual root cause (keyed-`{#each}` identity, and `''`-vs-absent canonicalization) rather than papering over symptoms, and the code comments explain the *why* concisely.

CI is green (`all-checks-passed`, `test-frontend`, all three `test-e2e` browsers; backend jobs correctly skipped). Not auto-merging only because of the optional nits above — they're take-it-or-leave-it, so this is fine to merge as-is if you'd rather not address them.

---

ginderjeremiah:
## Review

**What this PR does**

Fixes the two defects from #1784 in the Workbench lesson-step table:

1. **Ordinal collision breaking the keyed `{#each}`.** `SectionTable.svelte` keys rows by `row[rowKey]`, and the lesson steps table's `rowKey` is the author-editable `ordinal`. `setCell` now detects when an edit to the identity column would collide with a sibling and swaps the sibling onto the vacated identity, keeping every key unique. Implemented generically (`key === rowKey`), so it's a no-op for every other table whose `rowKey` is a surrogate id or a `unique` select.
2. **Cleared optional Anchor Key sticking false-dirty.** A new `ColumnConfig.optional` flag makes a text column's cleared input write `undefined` instead of `''`; `canonicalize` already collapses `undefined`/absent, so the row reads unmodified against a server baseline that omits the field.

Also introduces a shared `TableRow` type alias replacing the repeated inline `Record<string, number | string>` across the table components (a nice DRY cleanup) — widened to include `undefined`.

**Claim verification (per CLAUDE.md, verified against the real code, not the issue text):**
- `SectionTable.svelte:22` keys the `{#each}` on `row[rowKey]`, `rowKey` is `ordinal`, and `ordinal` is a `type: 'number'` author-editable column (`lesson.ts:147`) — confirmed a duplicate would collide. ✔
- `canonicalize` (`save-helpers.ts:10-28`) collapses `null`/`undefined` but returns `''` as-is, and `recordsEqual`/`listsEqual` both alias `canonicalEqual` — confirmed a cleared-to-`''` cell reads dirty against an absent baseline. ✔
- The swap logic is sound: the uniqueness invariant guarantees at most one colliding sibling, and `NumInput` only commits fully-parseable numbers (empty/in-progress states don't fire `onChange`), so the swap only runs on real committed values. ✔

**Findings**

Blocking: none.

Nits:
- The `warn` comment/validation in `lesson.ts:117-133` ("a duplicate crashes the keyed `{#each}` ... so it's caught here rather than left to blow up") is now largely unreachable for the ordinal-duplicate case, since `setCell` prevents the duplicate from ever committing. The remaining `warn` checks (missing text, empty list) are still useful, so keeping it as defense-in-depth is fine — but the ordinal-collision wording is slightly stale and could be trimmed.
- Editing an identity column silently renumbers the swapped sibling (marking it dirty). This is the intended, documented behavior and strictly better than the previous crash; noting only for awareness.

Test coverage: good. New unit tests cover the identity swap (`SectionTable.test.ts`), the `optional` → `undefined` vs non-optional → `''` branch (`TableCell.test.ts`), and the Anchor Key `optional: true` config (`lesson.test.ts`). Standards check: braces on all control blocks, no null-forgiving operator introduced (the new tests even prefer `as HTMLInputElement` casts over `!`), no dead code, comments are appropriately scoped.

**CI:** Green. `all-checks-passed`, `test-frontend`, `verify-frontend`, and all three `test-e2e` browsers succeeded; backend jobs correctly skipped (frontend-only). The `pending` combined-status is just the absence of legacy status contexts, not a failure.

**Verdict: approve**


---

ginderjeremiah:
Addressed two of the three review nits in aeecb80:

- `newRow`'s `anchorKey` now defaults to `undefined` instead of `''`, matching the "unset optional = `undefined`" convention this PR establishes.
- Added a `SectionTable.test.ts` case pinning the composed end-to-end symptom: typing into then erasing an optional cell returns the row to a baseline-equal (not-dirty) state, rather than sticking on a stray `""`.

Left the swap-ordering cosmetic nit as-is, per your own note that it doesn't need changing.

Lint, typecheck, and the full frontend suite (3586/3586) pass with these changes.

---

PR #1835, opened 2026-07-10
Fix game route lifecycle: socket listener leak on remount and welcome-back gate race

### LINKED ISSUE #1807: Game route lifecycle: startGame's socket listeners leak on every remount, and the welcome-back gate can start the engines after unmount
Two related defects in the `/game` route lifecycle seam (`UI/src/lib/engine/engine.ts` + `UI/src/routes/game/+page.svelte`). They'd likely be fixed together.

## 1. `stopEngines` never unhooks the socket listeners `startGame` registers

`startGame` registers four `apiSocket.listenCommand` subscriptions (`SocketReplaced`, `ChallengeCompleted`, `ProficiencyXpGained`, `ServerCommandFailed`) and stores their unhooks in module-level vars (`engine.ts:70-73`, `114-117`). The only path that invokes those unhooks is `stopGame` (`engine.ts:343-346`), which runs solely on the `SocketReplaced` push. The game route's unmount cleanup is `stopEngines` (`+page.svelte:79`), which stops the loops but never touches the listeners.

**Failure scenario:** navigate `/game` → `/admin` → `/game` (any client-side round-trip; logout/character-switch use full reloads so they're unaffected). Each re-entry re-runs `welcome.run()` → `startGame`, which registers a **second** set of listeners and overwrites the module-level unhook vars — permanently orphaning the first set. Handlers then run N times per push: duplicated combat-log lines from `notifyProficiencyResult`, duplicated `resyncPlayerAndInventory` round-trips on a dead-lettered `ChallengeCompleted`, N stacked modals + `stopGame` calls on a session takeover, and unbounded listener growth over a long session.

`engine.test.ts` pins that `stopGame` unregisters, but nothing pins single-registration across repeated `startGame` calls or unmount cleanup.

## 2. The welcome-back gate has no cancellation — a navigation during the boot check starts the engines after unmount

`welcome.run()`'s async continuation calls `enterGame` → `startGame()` unconditionally (`+page.svelte:60-74`, `welcome-back-view.svelte.ts:39-65`); nothing checks whether the game page is still mounted, and `onDestroy(stopEngines)` has already run by then.

**Failure scenario:** user lands on `/game` (boot splash, `GetOfflineProgress` in flight — up to the 30s socket timeout on a bad connection), then navigates to `/admin` or presses back. The response arrives, `enterGame` runs, and the full game — logical/render loops, enemy-fetch loop, live battles, victory claims — runs headless behind the other screen. A later return to `/game` re-runs the gate against an *already running* live loop, violating the gate's own "simulated and live battles never overlap" invariant (and compounding defect 1's duplicate listeners).

## Suggested fix

- Unhook the four listeners in `stopEngines` (or guard `startGame` against double registration by invoking any existing unhooks first).
- Add a cancellation flag to the gate (set from the page's `onDestroy`, checked in `enter()` before starting engines).
- Pin both with tests: single-registration across repeated `startGame` calls, and no engine start after cancellation.

ginderjeremiah:
## Summary

Two related defects in the `/game` route lifecycle seam (`UI/src/lib/engine/engine.ts` + `UI/src/routes/game/+page.svelte`), per #1807.

**1. `stopEngines` never unhooked the socket listeners `startGame` registers.** The game route's unmount cleanup calls `stopEngines` directly (not `stopGame`, which is only invoked via a `SocketReplaced` push). Only `stopGame` unhooked the four listeners (`SocketReplaced`, `ChallengeCompleted`, `ProficiencyXpGained`, `ServerCommandFailed`), so a client-side round-trip like `/game` → `/admin` → `/game` re-ran `startGame` and registered a second set of listeners on top of the first — permanently orphaning the original set and causing every handler to fire once per stacked registration (duplicated combat-log lines, duplicated resync round-trips, N stacked modals on a session takeover).

**2. The welcome-back gate had no cancellation.** `welcome.run()`'s async continuation (waiting on `GetOfflineProgress`, up to a 30s socket timeout) calls `enterGame` → `startGame()` unconditionally, with no check for whether the game page is still mounted. Navigating away mid-check let the full game — logical/render loops, enemy-fetch loop, live battles — start running headless behind another screen, and a later return to `/game` would then run the gate again against an already-running live loop.

## Changes

- `engine.ts`: `stopEngines` now unhooks the four `startGame`-registered listeners (via a new `unhookSocketListeners` helper), since it's the route's actual unmount cleanup. `stopGame` no longer duplicates the unhook calls — it already calls `stopEngines`. `startGame` also defensively unhooks any listeners left behind by a prior call before registering its own, as extra insurance against a missed `stopEngines`.
- `welcome-back-view.svelte.ts`: `WelcomeBackView` gains a `cancelled` flag and a `cancel()` method. `run()` bails out after each `await` once cancelled (before reconciling mode / before showing the summary gate), and `enter()` is a no-op once cancelled.
- `+page.svelte`: the route's `onDestroy` now calls both `stopEngines()` and `welcome.cancel()`.

## Tests

- `engine.test.ts`: new cases pinning that `stopEngines` unhooks the four listeners, that it's idempotent/safe when `startGame` never ran, that a later `startGame` re-registers cleanly after `stopEngines`, and that a second `startGame` call (without an intervening stop) unhooks the first set before re-registering.
- `welcome-back-view.test.ts`: new cases covering cancellation before `fetchProgress` resolves, cancellation during the post-progress resync, and `enter()` as a no-op once cancelled (both from the fast-path and after the summary gate is already showing).

## Test plan

- [x] `npm run lint` — clean.
- [x] `npm run test:unit:ci` — full suite passes: 3595/3595 (including the new/updated cases above), coverage floors intact.

Closes #1807

---
_Generated by [Claude Code](https://claude.ai/code/session_01QHSpi5H7kUEP6jwtCsVTUk)_

### DIFF
--- a/UI/src/lib/engine/engine.ts
+++ b/UI/src/lib/engine/engine.ts
@@ -108,9 +108,13 @@ export const startGame = (activeBattle?: IEnemyInstance) => {
 		startLogicEngine();
 		startRenderEngine();
 		startBattleEngine(activeBattle);
+		// Unhook any listeners a prior startGame left behind (defensive against a missed stopEngines)
+		// before registering this call's own — otherwise a second startGame without an intervening
+		// unhook would leak the first set and double-fire every handler.
+		unhookSocketListeners();
 		// cleanupOnDestroy must stay off: startGame runs outside component init (from welcome.run()'s
 		// async continuation or the WelcomeBackGate click handler), where Svelte's onDestroy throws.
-		// Teardown instead relies on the unhooks captured here, invoked unconditionally by stopGame.
+		// Teardown instead relies on the unhooks captured here, invoked unconditionally by stopEngines.
 		socketReplacedUnhook = apiSocket.listenCommand('SocketReplaced', handleSocketReplaced);
 		challengeCompletedUnhook = apiSocket.listenCommand('ChallengeCompleted', handleChallengeCompleted);
 		proficiencyXpGainedUnhook = apiSocket.listenCommand('ProficiencyXpGained', handleProficiencyXpGained);
@@ -320,11 +324,28 @@ export const handleSocketReplaced = async () => {
 };
 
 /**
- * Stops the live game loops (logical, render, battle) and the background-throttle monitor. Shared by the
- * full {@link stopGame} teardown and the game route's own unmount cleanup, so navigating away from the
- * game stops the loops without tearing down the socket/stores a session-replacement must clear. Every
- * stop is idempotent, so this is safe to call even when the loops never started (e.g. the player left
- * during the welcome-back gate, before {@link startGame} ran).
+ * Unhooks the four `startGame`-registered socket listeners, if any are currently registered.
+ * Idempotent (each unhook no-ops once already removed) and safe to call when `startGame` never ran.
+ */
+const unhookSocketListeners = () => {
+	socketReplacedUnhook?.();
+	challengeCompletedUnhook?.();
+	proficiencyXpGainedUnhook?.();
+	serverCommandFailedUnhook?.();
+	socketReplacedUnhook = undefined;
+	challengeCompletedUnhook = undefined;
+	proficiencyXpGainedUnhook = undefined;
+	serverCommandFailedUnhook = undefined;
+};
+
+/**
+ * Stops the live game loops (logical, render, battle), the background-throttle monitor, and the four
+ * `startGame`-registered socket listeners. Shared by the full {@link stopGame} teardown and the game
+ * route's own unmount cleanup, so navigating away from the game fully tears down this run of the engines
+ * — including its listeners — without tearing down the socket/stores a session-replacement must clear.
+ * Every stop is idempotent, so this is safe to call even when the loops never started (e.g. the player
+ * left during the welcome-back gate, before {@link startGame} ran), and safe to call again on a route
+ * that never re-entered the game.
  */
 export const stopEngines = () => {
 	logicEngine.stop();
@@ -332,6 +353,7 @@ export const stopEngines = () => {
 	renderEngine.stop();
 	enemyManager.stop();
 	battleEngine.stop();
+	unhookSocketListeners();
 };
 
 const stopGame = () => {
@@ -340,10 +362,6 @@ const stopGame = () => {
 	playerChallenges.reset();
 	playerProficiencies.reset();
 	resetLogs();
-	socketReplacedUnhook?.();
-	challengeCompletedUnhook?.();
-	proficiencyXpGainedUnhook?.();
-	serverCommandFailedUnhook?.();
 	// SocketReplaced routes back to login client-side (no reload), so the socket singleton survives. Tear
 	// it down explicitly, otherwise the keepalive ping would silently reconnect and fight the session that
 	// just took over.
--- a/UI/src/routes/game/+page.svelte
+++ b/UI/src/routes/game/+page.svelte
@@ -76,7 +76,12 @@ const welcome = new WelcomeBackView({
 if (browser) {
 	// Cleanup is registered once at init (the gate may start the loops later, after an await, where
 	// onDestroy can't be called); stopEngines is idempotent, so it's safe even if the loops never ran.
-	onDestroy(stopEngines);
+	// welcome.cancel() guards against welcome.run()'s async continuation (still in flight on a slow
+	// GetOfflineProgress) starting the engines after this page has already unmounted.
+	onDestroy(() => {
+		stopEngines();
+		welcome.cancel();
+	});
 	onMount(() => {
 		void welcome.run();
 	});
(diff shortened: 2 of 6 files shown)


ginderjeremiah (commented):
## Review: Fix game-route socket-listener leak and welcome-back gate race (#1807)

**Verdict:** Approve — posted as a Comment because I'm the PR author and GitHub won't accept a gating review on your own PR. Both defects are fixed correctly and the change is well-tested. Two non-blocking items below; nothing blocks merge.

### What's good
- **Defect 1** is fixed cleanly. `unhookSocketListeners()` centralizes the four unhooks, `stopEngines` (the route's actual unmount cleanup) now calls it, and `stopGame` no longer duplicates them — it delegates to `stopEngines`, so the disconnect still happens after the unhooks, preserving the original order. The defensive unhook at the top of `startGame`'s registration block is good insurance against a missed stop.
- **Defect 2**'s cancellation is correct. `cancelled` is checked after each `await` in `run()`, `enter()` no-ops once cancelled, and both are wired into `onDestroy`. Worth noting the fast-path is safe too: there's no `await` between the post-`fetchProgress` cancel check and the `enter()` call, so an unmount can't interleave and start the engines there.
- Tests map directly onto the issue's suggested coverage: single-registration across repeated `startGame`, idempotent `stopEngines`, clean re-registration after a stop, and all three cancellation timings (before `fetchProgress`, mid-`resyncPlayer`, and `enter()` post-summary).

### Should-fix (recommend a follow-up issue — not a blocker for this PR)
- Scoping the four listeners to the game-engine lifecycle is the right call, but it has a consequence worth a conscious decision: navigating `/game → /admin` now unhooks them while the **socket itself stays alive** (the keepalive keeps it open; `stopEngines` doesn't `disconnect` — only `stopGame` does). So a `SocketReplaced` session-takeover push that arrives while the user is sitting on `/admin` is no longer handled at all — previously the leaked listeners masked this. `SocketReplaced` has no app-global handler; only `startGame` registers one. This is edge-case and pre-existing-adjacent, so it shouldn't block this fix, but it's worth a follow-up to decide whether session-replacement handling belongs at the root/layout level rather than game-scoped.

### Nits
- The new `onDestroy(() => { stopEngines(); welcome.cancel(); })` in `game/+page.svelte` is the exact seam that ties the two fixes together, and it isn't asserted in `page.test.ts` (which already mocks `stopEngines`). The underlying behaviors are fully pinned in the two view-model suites, so this is defense-in-depth only — but a one-line "unmount calls `stopEngines`, and a `run()` still in flight can't start the game after unmount" test would guard the wiring cheaply.

### Test coverage
Meets the project's frontend testing rules and the issue's own stated criteria. The engine listener lifecycle and the gate cancellation are thoroughly unit-tested Detroit-style with injected deps. Only the page-level `onDestroy` glue is untested (see nit).

---

ginderjeremiah:
Thanks for the review. Addressed both non-blocking items:

- Added a `page.test.ts` case pinning that unmount calls `stopEngines` and cancels an in-flight `run()` so a late `GetOfflineProgress` can't start the game afterward.
- Filed #1836 for the `SocketReplaced`-unhandled-outside-`/game` gap you flagged — it's a real behavior change (previously masked by the leak) and deciding where session-replacement handling should live is a separate design call, so I kept it out of this PR's scope.

---

ginderjeremiah:
## Review — Approve

Verdict: **Approve**. Clean, well-scoped fix that addresses both defects in #1807 exactly as specified. CI is green (frontend suites, e2e across all three browsers, verify-frontend all passing; backend jobs correctly skipped for a frontend-only change).

### What it does well
- **Correct relocation of the listener unhooks.** Moving the four `startGame`-registered unhooks out of `stopGame` and into the new `unhookSocketListeners` helper — now invoked by `stopEngines` — puts teardown on the path the route actually calls on unmount (`onDestroy`), which was the root cause of the leak. `stopGame` still gets them transitively via its `stopEngines()` call, so the session-replacement path is unchanged. No double-unhook, and the helper nulls the vars so repeat calls are genuinely idempotent.
- **Defensive unhook in `startGame`** before re-registering is a sensible belt-and-suspenders guard against any future path that starts without an intervening stop, and it's pinned by a test (double `startGame` unhooks the first set exactly once).
- **Cancellation gate is minimal and correct.** `cancelled` is a plain private field (not `$state`) — the right call, since it's only read in async control flow, never in rendered reactive state. `run()` bails after each `await` boundary (post-`fetchProgress`, post-`resyncPlayer`) and `enter()` no-ops once cancelled, closing both the "start engines headless after unmount" window and the fast-path/summary-gate entry points.
- **Test coverage is thorough** and maps directly to the failure scenarios: single-registration across remount, idempotent stop with no prior `startGame`, clean re-registration after stop, and cancellation at each await boundary plus both `enter()` no-op paths. The `+page.svelte` unmount test pins the end-to-end "GetOfflineProgress resolves after unmount cannot start the game" behavior.
- Doc comments updated to match (the `stopGame` → `stopEngines` reference in `startGame`'s comment is now accurate). No dead code, always-braces respected, no null-forgiving operator.

### Blocking
None.

### Should-fix (nits)
None blocking. The new JSDoc blocks on `stopEngines`/`cancel` are on the longer side, but they're consistent with this file's established heavily-annotated style and carry real rationale, so I'd leave them.

### Follow-up
None.


---

PR #1837, opened 2026-07-10
Guard InventoryManager rollback against a resync that superseded it

### LINKED ISSUE #1808: InventoryManager.initialize() bypasses the serialized-mutation queue — a mid-flight equip rollback can restore pre-resync item objects
## Problem

`InventoryManager`'s optimistic mutations are single-flight-serialized (`private queue = new SerializedQueue()`, `inventory-manager.ts:80-81`) precisely so rollback baselines can't interleave — the comment at `inventory-manager.ts:366-374` spells out the divergence class this prevents. But `initialize()` (`inventory-manager.ts:83-113`) rebuilds `unlockedItems`/`items`/`equippedSlots` with **fresh `Item` instances outside the queue**, and it is invoked mid-session by `resyncPlayerAndInventory` (`engine.ts:54-57`) after a lost `DefeatEnemy` response (`enemy-manager.ts:704-712`) or a dead-lettered `ChallengeCompleted` (`engine.ts:130-139`).

## Failure scenario

An `EquipItem` is in flight when a `DefeatEnemy` times out (both ride the same flaky socket):

1. The resync rebuilds the inventory with fresh `Item` objects.
2. The equip response resolves with `error`.
3. Its rollback restores the **pre-resync** `equippedSlots` array and mutates pre-resync `Item` objects that are no longer in `unlockedItems`.

The equipped rail and `equipmentStats` now point at orphaned objects and diverge from the item grid until the next full reload — exactly the divergence the serialization exists to prevent. Every ingredient is a normal flaky-connection event in an idle game.

## Suggested fix

Route `initialize()` through `this.queue.run(...)` so it serializes with in-flight mutations, or bump an epoch/generation counter on re-init that voids pending rollbacks. Add a test covering resync-during-in-flight-equip.

ginderjeremiah:
## Summary

`InventoryManager`'s optimistic mutations (equip/unequip/applyMod/removeMod) are single-flight-serialized through `SerializedQueue` precisely so rollback baselines can't interleave. But `initialize()` (`UI/src/lib/engine/player/inventory-manager.ts`) rebuilds `unlockedItems`/`equippedSlots` with fresh `Item` instances **outside** that queue, since a mid-session resync (`resyncPlayerAndInventory`) can't wait behind an in-flight mutation's persist.

An `EquipItem` is in flight when a `DefeatEnemy` response times out (both ride the same flaky socket):

1. `resyncPlayerAndInventory` calls `inventoryManager.initialize()`, rebuilding the inventory with fresh `Item` objects — the authoritative state may even show the equip already landed server-side (only the ack was lost).
2. The in-flight equip's response later resolves with `error` (the transport settles a lost/timed-out response as an error even when the command may have actually succeeded server-side).
3. Its rollback fires unconditionally, restoring the **pre-resync** `equippedSlots` array and mutating pre-resync `Item` objects that are no longer in `unlockedItems` — clobbering the resync's fresh, authoritative state with a stale snapshot and splicing orphaned objects back into live state.

The equipped rail and `equipmentStats` then diverge from the item grid until the next full reload.

## Changes

- `inventory-manager.ts`: added a `generation` counter bumped on every `initialize()`. Each of the four rollback-capable mutations (`equipItem`, `unequipItem`, `applyMod`, `removeMod`) now captures `this.generation` at snapshot time and routes its error-path rollback through a new `rollbackIfCurrent` helper, which skips the rollback (and the post-rollback stats refresh) entirely if a resync has superseded it since — leaving the resync's authoritative state untouched.

I kept `initialize()` itself synchronous rather than routing it through the queue (the issue's other suggested fix shape) — the existing test suite calls it synchronously and asserts immediately after in ~60 places, and queuing it would force those onto an async `await` for no behavioral benefit over the generation-counter guard.

## Tests

- `inventory-manager.test.ts`: new `resync during an in-flight mutation (#1808)` describe block:
  - a mid-flight equip whose persist fails after a resync superseded it does **not** roll back (verified this fails without the fix — a stale rollback would clobber the resync's "actually equipped" state).
  - a mid-flight equip still rolls back normally when no resync happened (regression guard on the existing behavior).
  - a mid-flight unequip whose persist fails after a resync superseded it does **not** roll back (verified this fails without the fix too — the same corruption class in the opposite direction).

## Test plan

- [x] `npm run test:unit:ci` — full suite passes: 3589/3589, coverage floors intact.

Closes #1808

---
_Generated by [Claude Code](https://claude.ai/code/session_01HXJEREDtmBVTKz2X4RGpZj)_

### DIFF
--- a/UI/src/lib/engine/player/inventory-manager.ts
+++ b/UI/src/lib/engine/player/inventory-manager.ts
@@ -80,7 +80,18 @@ export class InventoryManager {
 	/** Serializes the optimistic mutations so rollback baselines never interleave (see {@link serialize}). */
 	private queue = new SerializedQueue();
 
+	/**
+	 * Bumped on every {@link initialize}. `initialize()` runs outside the {@link queue} (a mid-session
+	 * resync can't wait behind an in-flight mutation's persist), so a mutation that started optimistically
+	 * applying before a resync — then has its persist fail afterward — must not roll back: its snapshot
+	 * predates the resync's fresh item instances, and restoring it would splice orphaned pre-resync objects
+	 * back into live state (#1808). Each mutation captures the generation at snapshot time and checks it
+	 * via {@link rollbackIfCurrent} before invoking its rollback.
+	 */
+	private generation = 0;
+
 	public initialize() {
+		this.generation++;
 		this.unlockedItems.clear();
 		this.unlockedMods.clear();
 		this.equippedSlots = new Array(6).fill(undefined);
@@ -185,6 +196,7 @@ export class InventoryManager {
 
 			// Apply optimistically (instant UI), then persist; a failed persist rolls the change back so
 			// the authoritative state can never diverge from what was actually saved.
+			const startGeneration = this.generation;
 			const affected = [item, this.equippedSlots[slotId]].filter((it): it is Item => it != null);
 			const rollback = combineSnapshots(
 				snapshotFields(this, 'equippedSlots'),
@@ -197,8 +209,7 @@ export class InventoryManager {
 				equipmentSlotId: slotId
 			});
 			if (response.error) {
-				rollback();
-				this.refreshEquipmentStats();
+				this.rollbackIfCurrent(startGeneration, rollback, () => this.refreshEquipmentStats());
 				return false;
 			}
 
@@ -220,6 +231,7 @@ export class InventoryManager {
 				return false;
 			}
 
+			const startGeneration = this.generation;
 			const rollback = combineSnapshots(
 				snapshotFields(this, 'equippedSlots'),
 				snapshotFields(item, 'equipped', 'equipmentSlotId')
@@ -236,8 +248,7 @@ export class InventoryManager {
 				equipmentSlotId: slotId
 			});
 			if (response.error) {
-				rollback();
-				this.refreshEquipmentStats();
+				this.rollbackIfCurrent(startGeneration, rollback, () => this.refreshEquipmentStats());
 				return false;
 			}
 
@@ -261,6 +272,7 @@ export class InventoryManager {
 				return false;
 			}
 
+			const startGeneration = this.generation;
 			const rollback = snapshotFields(item, 'appliedMods', 'totalAttributes');
 			item.appliedMods = [...item.appliedMods.filter((m) => m.itemModSlotId !== itemModSlotId), mod];
 			this.refreshItemAttributes(item);
@@ -272,8 +284,7 @@ export class InventoryManager {
 				itemModSlotId
 			});
 			if (response.error) {
-				rollback();
-				this.refreshEquipmentStatsIfEquipped(item);
+				this.rollbackIfCurrent(startGeneration, rollback, () => this.refreshEquipmentStatsIfEquipped(item));
 				return false;
 			}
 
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Guard InventoryManager rollback against a superseding resync (#1808)

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't accept a gating review from me; this is my genuine approve call).* The generation-counter guard correctly closes the resync-clobbers-rollback hole from #1808, and the reasoning holds up under the edge cases.

### Correctness analysis
Walked the fix against the failure modes it targets:

- `startGeneration` is captured **inside** each serialized closure, in the same synchronous block as the snapshot and optimistic apply — no `await` sits between the capture and the apply, so `initialize()` can't interleave there. The only window where the generation can move is across the `await sendSocketCommand(...)`, which is exactly the window `rollbackIfCurrent` guards.
- `generation` bumps **only** in `initialize()`, which is also the only place item objects are rebuilt — so `generation === startGeneration` is a sound proxy for "the snapshot's item objects are still the live ones." Skipping the rollback when it moved is correct in both directions: whether the resync's authoritative state shows the op as landed (ack lost) or not landed, the resync already carries the truth, so the stale snapshot must not be reapplied.
- Skipping the `after()` refresh alongside the rollback is right — refreshing off an orphaned `item` (applyMod/removeMod) or the pre-resync slot array would re-derive stale stats.

Choosing the generation counter over routing `initialize()` through the queue is the better of the two shapes the issue floated: a mid-session resync genuinely *shouldn't* wait behind an in-flight persist (the issue says as much), and the class-level comment documents the rationale well.

### Test coverage
Solid. The new `#1808` block covers the equip supersede-skip (with the "actually landed server-side" case), the unequip supersede-skip (opposite corruption direction), and a no-resync regression guard that the normal rollback still fires. Both branches of `rollbackIfCurrent` are exercised.

### Nits (non-blocking)
- The applyMod/removeMod call sites route through the same `rollbackIfCurrent` helper but only have their *normal* error-rollback tests (`applyMod:942`, `removeMod:1067`) — no resync-supersede case. Since all four sites share the one helper and equip/unequip already cover both of its branches, this is redundant-symmetry rather than a real gap; adding the two mirrored cases would just make the coverage uniform. Take it or leave it.

### What's good
- The class-level and helper doc comments make the invariant (`generation` bump ⟺ fresh item objects) explicit, which is exactly the kind of subtle timing contract a later "simplification" could otherwise break.
- The four call sites collapse to a single, uniform `rollbackIfCurrent(...)` shape — no divergence between them.

Not merging on this pass: `test-frontend` and the three `test-e2e` jobs were still in progress at review time (backend jobs correctly skipped for a frontend-only change). No follow-up issue needed — the one nit is optional and doesn't warrant tracking.

---

ginderjeremiah:
Thanks for the review. On the nit — I looked into mirroring the resync-supersede test for `applyMod`/`removeMod`, but it turns out not to be meaningful: unlike `equipItem`/`unequipItem`, whose rollback restores `this.equippedSlots` (the manager's own live field) directly, `applyMod`/`removeMod`'s rollback only mutates the orphaned pre-resync `item` object — it never touches the live `equippedSlots` array or the `unlockedItems` map. And `refreshEquipmentStatsIfEquipped`'s `after()` step always recomputes off the *current* `equippedSlots`, which already holds the fresh post-resync objects regardless of whether the rollback fired. So a mirrored test there would pass identically with or without the guard — a vacuous regression test. I kept the guard on those two call sites for uniformity/defense-in-depth (per your own "What's good" note), but left the test coverage as-is rather than pad the suite with a tautological case. Merging now that CI is green.

---

PR #1838, opened 2026-07-10
Re-seed Attributes/Options screen baselines on a mid-session resync

### LINKED ISSUE #1809: Attributes/Options screen baselines are constructor-seeded and go stale when the player aggregate is re-initialized mid-session
## Problem

`AttributesView` (`committed`/`draft`, `attributes-view.svelte.ts:292-297`, `469-471`) and `OptionsView` (`baseline`/`draft`, `options-view.svelte.ts:146-150`, `249-255`) snapshot `playerManager` state once at construction and never observe a mid-session `playerManager.initialize()` — the resync path (`refreshPlayer`/`loadPlayer` → `playerManager.initialize`) that runs after a lost `DefeatEnemy` response or a dead-lettered `ChallengeCompleted` reassigns `attributes`, `statPointsUsed`, and `logPreferences` underneath them.

## Failure scenario

Player parks on the Attributes screen (an idle game invites parking). An earlier `UpdatePlayerStats` response was lost (the save toasts an error, the baseline isn't advanced), and a later background resync pulls the server state where that save actually applied. The screen still shows the pre-resync allocations, `remaining` is computed against a mismatched baseline (`savedAvailable` updates live but `pendingSpent` uses the stale `committed`), and the next save sends deltas relative to a baseline the server no longer holds. Same shape for log preferences on Options.

## Suggested fix

Re-seed baseline + draft when the underlying manager state is replaced. `SkillsView.equipped` already solved exactly this pattern for the loadout (comparing against the manager state by reference in its `$derived` chain), so the fix has an in-repo template.

ginderjeremiah:
## Summary

`AttributesView` (`committed`/`draft`) and `OptionsView` (`baseline`/`draft`) snapshotted the player manager's state once at construction and never observed a mid-session `playerManager.initialize()` — the resync path that runs after a lost `DefeatEnemy`/`UpdatePlayerStats` response or a dead-lettered `ChallengeCompleted`.

Player parks on the Attributes (or Options) screen. An earlier save's response was lost (the save toasts an error, the baseline isn't advanced), and a later background resync pulls the server state where that save actually applied. Previously the screen kept showing the pre-resync allocations, `remaining`/dirty was computed against a baseline the server no longer held, and the next save would have sent deltas relative to a stale baseline.

## Changes

- `attributes-view.svelte.ts`: `committed` is now a live `$derived` read off `playerManager.attributes` instead of a `$state` field seeded once in the constructor. Unsaved edits are tracked as per-attribute deltas (`#pendingDeltas`) that stay valid only while `#pendingDeltasAttrs` (the manager `attributes` reference they were computed against) still matches the live one — `effectiveDeltas()` treats them as zero the instant it doesn't. An **external** resync (not this view's own `save`) is therefore silently dropped rather than layered onto a baseline it was never computed against; `save()` keeps the marker in lockstep with its own reassignment so it correctly preserves an edit made while the request was in flight (#1506).
- `options-view.svelte.ts`: the identical pattern for `baseline`/`draft`, using a sparse per-log-type override map (`#draftOverrides`) instead of numeric deltas.
- Test mocks for `playerManager` in `Attributes.test.ts` and `options-view.test.ts` now use a `statify`'d class instance (matching `attributes-view.test.ts`'s existing pattern) so the mock is reactive like the real (statified) `PlayerManager` — required for the live-baseline behavior to be exercised at all.
- New tests pin: following an idle resync, clearing a stale dirty state when a resync reveals a lost save actually applied, and discarding an in-progress unsaved edit on an unrelated external resync so a stale delta/diff is never sent.

## Note for reviewer

I considered a design where an edit unrelated to what a resync touched would survive the resync (only the touched attribute/log-type resets), but that reopens a double-counting risk: if the resync is itself the arrival of a previously-lost save, any surviving delta would be re-applied on top of a baseline that already includes it. I went with the simpler, safer rule instead: **any** external resync discards **all** unsaved edits outright, matching the issue's suggested fix ("re-seed baseline + draft"). This is a rare recovery-path edge case, so losing an in-progress unsaved allocation/toggle on a resync is an acceptable trade for correctness.

## Test plan

- [x] `npm run lint` — clean.
- [x] `npx vitest run` — full suite passes: 3595/3595.

Closes #1809

---
_Generated by [Claude Code](https://claude.ai/code/session_019fj4HMEo4bvGd4BM4duNab)_

### DIFF
--- a/UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
+++ b/UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
@@ -461,8 +461,14 @@ export class AttributesView {
 		this.saving = false;
 
 		// Snapshot the current effective deltas (any mid-flight edit included) before the
-		// reassignment below moves the resync marker out from under them.
+		// reassignment below moves the resync marker out from under them. Reference equality
+		// with `#pendingDeltas` tells whether they were still valid at this exact moment —
+		// `effectiveDeltas()` falls back to a *new* zero array (never `#pendingDeltas` itself)
+		// the instant an external resync lands mid-flight and invalidates them; in that case
+		// they're already gone and must not be "consumed" a second time below (which would
+		// otherwise subtract `sentDeltas` from zero and manufacture a phantom negative delta).
 		const currentDeltas = this.effectiveDeltas();
+		const deltasStillValid = currentDeltas === this.#pendingDeltas;
 
 		// Any returned payload is the server's authoritative post-command state (the
 		// rejection path returns it unchanged), so adopt the allocations and the spent
@@ -483,8 +489,11 @@ export class AttributesView {
 		}
 
 		// Consume exactly the deltas that were sent, preserving any edit made while the save was
-		// in flight instead of wiping it (#1506).
-		this.#pendingDeltas = this.#pendingDeltas.map((d, i) => d - sentDeltas[i]);
+		// in flight instead of wiping it (#1506) — unless an external resync already discarded
+		// the pending set mid-flight, in which case there's nothing left to consume.
+		if (deltasStillValid) {
+			this.#pendingDeltas = this.#pendingDeltas.map((d, i) => d - sentDeltas[i]);
+		}
 		this.flashSaved();
 	}
 
--- a/UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
+++ b/UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
@@ -268,17 +268,34 @@ export function derivedShortLabel(id: EAttribute): string {
 }
 
 /* ── reactive view-model ──────────────────────────────────────────────────
-   Owns the working allocation (`draft`) and the persisted baseline
-   (`committed`), derives the dirty set and the live/preview derived stats, and
+   Owns the persisted baseline (`committed`) and the working allocation
+   (`draft`), derives the dirty set and the live/preview derived stats, and
    persists only the per-attribute deltas (mirroring the backend's update
-   contract). */
+   contract).
+
+   `committed` is read live off the player manager rather than snapshotted once
+   at construction: a mid-session resync (`playerManager.initialize()`, #1809)
+   reassigns `attributes` wholesale, and the baseline must track it or the
+   screen computes `remaining`/deltas against a server state that no longer
+   exists. `draft` layers unsaved per-attribute deltas (`#pendingDeltas`) over
+   `committed`, but only while `#pendingDeltasAttrs` — the manager `attributes`
+   reference those deltas were computed against — still matches the live one;
+   {@link effectiveDeltas} treats them as zero the moment it doesn't. A
+   not-yet-saved edit can't be trusted to still apply cleanly against a
+   baseline it was never computed against, and an **external** resync (not this
+   view's own `save`, which keeps the marker in lockstep) is by definition the
+   fresher truth — so it silently drops the edit rather than layering it onto
+   the new baseline. `save` instead consumes exactly the deltas it sent,
+   preserving any edit made while that request was in flight (#1506). */
 export class AttributesView {
 	/** Active density mode; persisted to local storage per player. */
 	mode = $state<AttributeMode>('guided');
-	/** Working allocation per core attribute (index matches CORE_ATTRIBUTES). */
-	draft = $state<number[]>([]);
-	/** Last-saved allocation; the dirty set is `draft` minus this. */
-	committed = $state<number[]>([]);
+	/** Unsaved per-attribute deltas (index matches CORE_ATTRIBUTES), valid only against
+	 *  {@link #pendingDeltasAttrs} — see {@link effectiveDeltas}. */
+	#pendingDeltas = $state<number[]>(CORE_ATTRIBUTES.map(() => 0));
+	/** The `playerManager.attributes` reference {@link #pendingDeltas} was computed against; `null`
+	 *  never matches, so deltas read as zero until the first edit stamps it. */
+	#pendingDeltasAttrs = $state<IBattlerAttribute[] | null>(null);
 	/** Brief "Attributes saved" confirmation flash. */
 	saved = $state(false);
 	/** True while a save request is in flight (guards against double-submit). */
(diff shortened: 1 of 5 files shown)


UI/src/routes/game/screens/attributes/attributes-view.svelte.ts
```diff
@@ -420,29 +452,39 @@ export class AttributesView {
...
 		}
 
-		// Advance the baseline by exactly the deltas that were sent, preserving any
-		// draft edits made while the save was in flight instead of wiping them (#1506).
-		this.committed = this.committed.map(
-			(value, i) => value + (updates.find((u) => u.attributeId === CORE_ATTRIBUTES[i])?.amount ?? 0)
-		);
+		// Consume exactly the deltas that were sent, preserving any edit made while the save was
+		// in flight instead of wiping it (#1506).
+		this.#pendingDeltas = this.#pendingDeltas.map((d, i) => d - sentDeltas[i]);
```
ginderjeremiah: **Correctness (should-fix): this unconditional subtraction can produce spurious negative pending deltas when an external resync lands during an in-flight save.**

Consider this interleaving:
1. User allocates +3 STR → `#pendingDeltas = [+3,…]`, marker = `attributes` ref `R0`. `sentDeltas = [+3,…]`.
2. `save()` awaits `UpdatePlayerStats`. **During the await, an external resync** (`playerManager.initialize()` from a `ChallengeCompleted`/`DefeatEnemy` round-trip — exactly the path this PR targets) reassigns `playerManager.attributes` to `R1`.
3. `currentDeltas = effectiveDeltas()` at line 465 now runs with marker `R0` ≠ live `R1` → **returns all-zero**.
4. The command succeeded, so `response.data` is set: line 476 assigns `#pendingDeltas = currentDeltas` (`[0,…]`) and re-pins the marker to `R2` (the authoritative response).
5. This line then does `[0,…].map(d => d - sentDeltas[i])` = **`[-3,…]`**.

Result: after a *successful* save the screen is `dirty` with a phantom −3 STR edit (`committed` = server's post-command 8, `draft` = 5), and `changedUpdates` would send `{STR: -3}` — an unintended point refund — if the user hits Save again.

The sibling `OptionsView.save()` does **not** have this bug: it consumes via *delete-if-still-equal* (`options-view.svelte.ts:271-277`), so an emptied `currentOverrides` stays empty rather than going negative. The two siblings should behave consistently. A minimal fix is to only subtract when the pending set was still valid at consume time (i.e. skip the subtraction when `currentDeltas` was dropped to zero by the marker mismatch), mirroring the Options approach.

No test currently exercises this race — the `#1506` in-flight test (line 659) only does a *local* mid-flight `inc`, which never reassigns `playerManager.attributes`. A regression test that reassigns `attributes` while the save promise is pending would pin it.

---

ginderjeremiah (commented):
## Review: Re-seed Attributes/Options screen baselines on a mid-session resync (#1809)

**Verdict:** Request changes _(submitted as a comment — GitHub won't let me formally gate my own PR)_ — the core design is right and well-documented, but `AttributesView.save()` has a real correctness bug in the exact resync-race domain this PR targets, and its `OptionsView` sibling shows it's fixable.

The approach is sound: making `committed`/`baseline` live `$derived` reads off the reactive player manager, and validating unsaved edits by reference-identity against the `attributes`/`logPreferences` they were computed against, is the correct evolution of the constructor-snapshot pattern and matches the `SkillsView.equipped` template the issue pointed at. The three new idle/lost-save/external-resync tests on each side are the right scenarios.

### Should-fix (correctness)

- **`attributes-view.svelte.ts:487`** — the unconditional `#pendingDeltas.map((d,i) => d - sentDeltas[i])` produces spurious **negative** pending deltas when an external resync reassigns `playerManager.attributes` during an in-flight save that then succeeds: `currentDeltas` is dropped to zero by the marker mismatch (line 465), so the consume step computes `0 - sentDeltas = -sentDeltas`. The screen ends up `dirty` after a successful save and `changedUpdates` would send an unintended refund on the next Save. See the inline comment for the full interleaving and a suggested fix. The `OptionsView.save()` sibling (`options-view.svelte.ts:271-277`) handles the same race correctly via delete-if-still-equal, so this is an inconsistency between two siblings that should be reconciled.

### Test coverage

- Solid for the scenarios handled: idle resync, a resync revealing a lost save applied, and an external resync discarding an unsaved edit — mirrored on both views. The `statify`'d class-instance mock change is the right call and matches `attributes-view.test.ts`'s existing pattern.
- **Gap:** no test reassigns `playerManager.attributes`/`logPreferences` *while a save promise is pending*. The existing `#1506` in-flight test (`attributes-view.test.ts:659`) only does a local mid-flight `inc`, which never swaps the manager reference — so the race above slips through. A regression test that resolves the send after an external reassignment would pin it (and confirm the Options side stays clean).

### What's good

- The reference-identity marker (`#pendingDeltasAttrs`/`#draftOverridesPrefs`) with a `null` sentinel that "never matches until the first edit stamps it" is a clean, well-reasoned way to make an unsaved edit self-invalidating.
- The class-doc comments are longer than usual but genuinely earn their length here — the invalidation semantics are subtle and easy to "simplify" into a bug otherwise.
- Correctly keeps the `#1506` in-flight-edit and `#1548` authoritative-adopt behaviors intact rather than regressing them while reworking the state model.
- The design note in the PR body (deliberately dropping *all* unsaved edits on any external resync rather than a per-field merge, to avoid double-counting a lost-then-resynced save) is the right trade-off and well justified.

Once the `save()` consume-step is made resync-safe (and covered), this is good to merge.

---

ginderjeremiah:
Fixed in 9356e85 — `save()` now only subtracts `sentDeltas` when `effectiveDeltas()` returned the still-valid `#pendingDeltas` reference (not a fresh zero array from an invalidated marker), so a mid-flight external resync no longer produces a phantom negative delta. Added a regression test on each side that reassigns `attributes`/`logPreferences` while the save promise is pending, confirming `OptionsView` was already clean and `AttributesView` now is too.

---

ginderjeremiah:
## Review — Re-seed Attributes/Options baselines on mid-session resync

**Verdict: Approve.** This correctly closes #1809 and is a careful, well-tested implementation. CI is green (test-frontend, verify-frontend, all three e2e browsers, all-checks-passed — backend jobs correctly skipped for a frontend-only change).

### What it does well
- Replaces the constructor-snapshotted `committed`/`baseline` with a live `$derived` read off the player manager, so an external resync (`playerManager.initialize()`) is tracked instead of going stale. This mirrors the existing `SkillsView.equipped` live-derived + reference-marker pattern, so it's consistent with in-repo precedent rather than a new invention.
- The reference-equality marker (`#pendingDeltasAttrs` / `#draftOverridesPrefs`) is a clean way to invalidate an unsaved edit the instant the underlying manager state is reassigned, and `effectiveDeltas`/`effectiveOverrides` stay pure reads safe to call from `$derived`.
- I traced the `save()` flow through the tricky cases: (a) clean save with no mid-flight activity consumes exactly the sent deltas; (b) resync-mid-flight with no re-edit trips `deltasStillValid === false` and correctly skips the consume, avoiding the phantom negative delta; (c) a mid-flight edit is preserved (#1506). Even the resync-then-re-edit case stays self-consistent because the visible `draft` (committed + pending) is preserved across the committed jump. The error/rejection-with-data path re-pins and preserves the still-unconsumed deltas so retry works. The logic holds up.
- Test coverage is thorough and symmetric across both views (idle resync, lost-save reveal clears stale dirty, external-resync discards in-flight edit, phantom-negative/toggle-resurrection guards). The mock change to a `statify`'d class instance is the right call — a plain object literal wouldn't exercise the live-baseline reactivity at all.

### Concerns
**Blocking:** none.

**Should-fix (nits):** none blocking. The class-doc block comments are long, but they document a genuinely non-obvious reference-equality invariant, so the length is earned — not flagging.

**Follow-up (separate issue, optional):** Three view-models (`SkillsView`, `AttributesView`, `OptionsView`) now share the "baseline is live-derived off the player manager, never constructor-snapshotted, with a reference-equality marker for pending edits" pattern. A short note in `docs/frontend.md` capturing this as the expected pattern for player-manager-backed view-models would help prevent the constructor-snapshot regression from recurring in future screens. Labels: `tech debt`, `claude`, `scope: small`.

One correctness note worth double-checking outside this PR (not introduced here): the whole scheme relies on `playerManager.attributes`/`logPreferences` always being *reassigned by reference* (never mutated in place) on both save and resync. That's true today, but any future path that mutates those arrays in place would silently defeat the marker. Not a blocker for this PR.

---

PR #1842, opened 2026-07-10
Fail the coverage gate when a branch floor is configured but the report has no branch data

### LINKED ISSUE #1811: Backend coverage gate silently un-gates the branch dimension when the report carries no branch data
## Problem

In `build/coverage-gate.ps1`, `$branchOk` stays `$true` when `$a.branchcoverage` is `$null` (lines 61-64 for gated assemblies, 93-96 for gated namespaces; `Get-NamespaceCoverage` in `build/coverage-lib.ps1:26-27` returns a `$null` `BranchPct` when `totalBranches` is 0). So if the collector/report ever stops emitting branch data — a tool behavior change, a filter change — every configured branch ratchet passes silently.

This is asymmetric with the gate's own design principle: a missing gated assembly or an emptied namespace bucket fails loudly (`coverage-gate.ps1:55-58`, `87-90`), but a missing gated *dimension* does not.

## Suggested fix

When a branch floor is configured (`$floor.branch`/`$entry.branch` non-null) and the report has no branch figure, add a failure ("branch floor configured but no branch data in report") instead of passing.

ginderjeremiah:
## Summary

In `build/coverage-gate.ps1`, `$branchOk` defaulted to `$true` whenever `branchcoverage`/`BranchPct` was `$null` — for gated assemblies (lines 61-64) and for gated namespaces (lines 93-96, where `Get-NamespaceCoverage` in `build/coverage-lib.ps1` returns a `$null` `BranchPct` when `totalBranches` is 0). So if the collector/report ever stopped emitting branch data — a tool behavior change, a filter change — every configured branch ratchet would pass silently instead of enforcing.

This was asymmetric with the gate's own design principle: a missing gated assembly or an emptied namespace bucket fails loudly (lines 55-58, 87-90), but a missing gated *dimension* did not.

`dotnet-coverage`/ReportGenerator stops including branch data in a report (upstream behavior change, a filter tweak, etc.) — every assembly/namespace with a configured `branch` floor silently stops being gated on branches, and a real branch-coverage regression could ship undetected.

## Changes

- `build/coverage-gate.ps1`: when a branch floor is configured (`$floor.branch`/`$entry.branch` non-null) and the report has no branch figure, the gate now adds a failure ("branch floor configured but no branch data in report") instead of passing. Applies to both the gated-assembly loop and the gated-namespace loop.

## Test plan

`pwsh` isn't available in the CI test-runner sense here beyond running the script itself, so I verified behaviorally with `dotnet tool install --global PowerShell` and synthetic `Summary.json`/`coverage-floors.json` fixtures covering:
- [x] Branch floor configured, no branch data (`null`) → gate now **fails** with the new message (previously passed silently) — both assembly-level and namespace-level.
- [x] Branch floor configured, branch data present and above floor → **passes** (unaffected).
- [x] Branch floor configured, branch data present and below floor → **fails** with the original message (unaffected).
- [x] No branch floor configured at all, branch data `null` → **passes** (unaffected — the dimension simply isn't gated).
- [x] `dotnet test` — full solution suite passes: 3049/3049 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests).
- [x] `./build/coverage.ps1 -NoBuild` (the real script, with the fix in place) run end-to-end against the actual test suite — gate still passes cleanly against real coverage data, confirming the fix has no effect on the normal (branch-data-present) path.

Closes #1811

---
_Generated by [Claude Code](https://claude.ai/code/session_01SXqSwc2V1QoxscC7Aoum7q)_

### DIFF
--- a/build/coverage-gate.ps1
+++ b/build/coverage-gate.ps1
@@ -59,11 +59,16 @@ foreach ($name in $gatedNames) {
   }
   $lineOk = $a.coverage -ge $floor.line
   $branchOk = $true
-  if ($null -ne $floor.branch -and $null -ne $a.branchcoverage) {
-    $branchOk = $a.branchcoverage -ge $floor.branch
+  if ($null -ne $floor.branch) {
+    if ($null -eq $a.branchcoverage) {
+      $branchOk = $false
+      $failures.Add("$name branch floor configured ($($floor.branch)%) but no branch data in report")
+    } elseif ($a.branchcoverage -lt $floor.branch) {
+      $branchOk = $false
+      $failures.Add("$name branch $($a.branchcoverage)% < floor $($floor.branch)%")
+    }
   }
   if (-not $lineOk) { $failures.Add("$name line $($a.coverage)% < floor $($floor.line)%") }
-  if (-not $branchOk) { $failures.Add("$name branch $($a.branchcoverage)% < floor $($floor.branch)%") }
   $status = 'PASS'
   if (-not ($lineOk -and $branchOk)) { $status = 'FAIL' }
   $floorText = "L$($floor.line)"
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: coverage gate now fails when a branch floor is configured but the report has no branch data (#1811)

**Verdict:** Approve — the change fixes the exact asymmetry the issue describes, is correct in context, and leaves every other path unchanged. (Posted as a comment since GitHub won't accept a gating approval from the PR author.)

### Correctness
- Both loops — gated assemblies (`build/coverage-gate.ps1:62`) and gated namespaces (`:97`) — now add a loud failure when a branch floor is configured (`$floor.branch`/`$entry.branch` non-null) but the report's branch figure is `$null`, matching the gate's own "fail loudly on a missing gated thing" principle (missing assembly / emptied namespace already did).
- The other paths are untouched: no branch floor → `$branchOk` stays `$true` and the dimension isn't gated; floor met → no failure; floor breached → the original `branch X% < floor Y%` message, now added inline.
- The old trailing `if (-not $branchOk) { $failures.Add(...) }` line was correctly removed so the breach message isn't double-added, and `$branchOk` still feeds the `$status` computation. `Format-Pct` renders `n/a` for a null branch figure, so the summary table still prints cleanly on the failing path.

### Test coverage
These are `build/*.ps1` build-tooling scripts with no PowerShell test harness in the repo, so the C#/frontend testing rules in `CLAUDE.md` don't apply here. The behavioral verification with synthetic `Summary.json`/`coverage-floors.json` fixtures (all four branches: null-with-floor, present-above, present-below, no-floor) documented in the PR is the right level of rigor for this fix.

### What's good
- Tight, symmetric fix applied to both loops; failure messages are specific and actionable (they name the configured floor).
- Scope discipline — no drive-by changes, matches the issue's suggested shape precisely.

---

PR #1843, opened 2026-07-10
Fix createFirstCharacter e2e helper racing the async create-form mount

### LINKED ISSUE #1815: e2e createFirstCharacter races the auto-opening create form (flake source)
## Problem

`UI/e2e-tests/helpers.ts:65-68` does an instantaneous `if (!(await nameInput.isVisible()))` check against a create form that auto-opens **asynchronously** after the players list loads (`UI/src/routes/select/PlayerSelectPanel.svelte:26-51`). If the check evaluates false and the form then auto-opens, the `show-create` button no longer exists (it's the `{:else}` branch of the same conditional), so the fallback click times out the test.

Timing-dependent; Playwright retries currently absorb it, making it a standing background flake source rather than a hard failure.

## Suggested fix

Wait for either the name input or the show-create button (e.g. `nameInput.or(showCreateButton).waitFor()`), then branch on which one is present — instead of snapshotting visibility at one instant.

ginderjeremiah:
## Summary

`UI/e2e-tests/helpers.ts`'s `createFirstCharacter` did an instantaneous `nameInput.isVisible()` check immediately after `toHaveURL('/select')` resolved, to decide whether to click the `show-create` button. But `PlayerSelectPanel` (and its mutually-exclusive create-form-vs-`show-create` branch) mounts asynchronously after the URL changes — `+page.svelte`'s `onMount` constructs the `PlayerSelectView` (which synchronously opens the create form for an empty account) only after a render tick. If `isVisible()` ran in the window before that mount, it saw neither element and fell through to a `show-create` click — but by the time it executed, the create form was already open and `show-create` never renders (it's the `{:else}` branch), so the click timed out.

`createAccountAndLogin` → `createFirstCharacter` right after the signup submit click, while `/select` is still finishing its client-side mount. A standing background e2e flake source (Playwright's retries were absorbing it rather than it being a hard, reliably-reproducing failure).

## Changes

- `UI/e2e-tests/helpers.ts`: `createFirstCharacter` now waits for either `new-name-input` or `show-create` to actually exist (`nameInput.or(showCreateButton).waitFor()`) before checking which one is present, instead of taking an instantaneous snapshot.

## Test plan

- [x] `npm run lint` — clean.
- [x] `npx vitest run` (`test:unit:ci`) — 3607/3607 passing.
- [x] Ran the affected e2e spec (`select.test.ts`) against a natively-run API + the existing Postgres/Redis containers: 16/16 passes across 8 repeats under 4 parallel workers, confirming the fix holds under load.
- [ ] Backend unaffected (e2e-test-only change) — no backend logic changed.

Closes #1815

---
_Generated by [Claude Code](https://claude.ai/code/session_01NvEW94uTS2ECdCLCCxK9kP)_

### DIFF
--- a/UI/e2e-tests/helpers.ts
+++ b/UI/e2e-tests/helpers.ts
@@ -62,9 +62,15 @@ export async function selectFirstCharacter(page: Page) {
 export async function createFirstCharacter(page: Page, name = 'Hero') {
 	await expect(page).toHaveURL('/select', { timeout: 10000 });
 
+	// The panel (and its create-form-vs-"+ New character" branch) mounts asynchronously after the URL
+	// changes, so an immediate `isVisible()` snapshot can catch neither element rendered yet and
+	// wrongly fall through to a "show-create" click that never appears (the create form ends up open
+	// instead). Wait for either to actually exist before branching on which one it is.
 	const nameInput = page.getByTestId('new-name-input');
+	const showCreateButton = page.getByTestId('show-create');
+	await nameInput.or(showCreateButton).waitFor({ timeout: 10000 });
 	if (!(await nameInput.isVisible())) {
-		await page.getByTestId('show-create').click();
+		await showCreateButton.click();
 	}
 	// A class is required to create, so wait for the picker to load (it defaults to the first class)
 	// before submitting — the Create button stays disabled until a class is selected.


ginderjeremiah (commented):
## Review: e2e `createFirstCharacter` helper races the async-mounting create form (#1815)

**Verdict:** Approve *(posted as a comment — GitHub won't accept a gating self-review from the PR author)* — the change is a correct, minimal fix that exactly matches the issue's diagnosis and suggested shape.

### Analysis

The root cause is accurately identified: `PlayerSelectPanel.svelte` renders `new-name-input` (inside `CreateCharacter`) and `show-create` as the two mutually-exclusive branches of `{#if view.showCreate}` / `{:else}`, and the panel mounts a tick after `/select` resolves. The old instantaneous `isVisible()` snapshot could observe neither branch and fall through to a `show-create` click that races the auto-opening form — a click that then never lands because the form is already open.

The fix — `await nameInput.or(showCreateButton).waitFor()` before branching — is exactly right. After the wait, at least one branch is guaranteed visible, so the `!nameInput.isVisible()` check reliably resolves to the correct branch: if the form isn't open, `show-create` is necessarily present to click. Timeout is consistent with the surrounding `10000ms` waits.

### Test coverage

No test additions expected or needed — this *is* test-infrastructure code, and the fix is validated by the existing `select.test.ts` flows it supports (the PR notes 16/16 across repeated parallel runs). Correctly scoped as a frontend/e2e-only change: `verify-frontend` passed and the backend jobs skipped.

### What's good

- Fix matches the issue's suggested approach precisely, no over-engineering.
- The added comment clearly documents *why* the wait exists, which protects against a future "simplification" back into the racy snapshot.

Merging is gated only on the e2e/frontend CI jobs, which are still in progress at review time.

---

PR #1844, opened 2026-07-11
Fail fast on unparseable ForwardedHeaders proxy/network config entries

### LINKED ISSUE #1816: Unparseable trusted-proxy config entries silently invert the forwarded-headers trust model (trust-nothing becomes trust-everything)
## Problem

`ForwardedHeadersConfig.Apply` (`Game.Api/Forwarding/ForwardedHeadersConfig.cs:48-70`) silently drops any `KnownProxies`/`KnownNetworks` entry that fails `TryParse`, while `HasTrustedProxies` (line 39) counts the **raw config strings**. A config whose entries are all unparseable therefore enables the middleware (`Startup.cs` gates on `HasTrustedProxies`) with **empty** framework trust lists — and ASP.NET Core's `ForwardedHeadersMiddleware` skips its known-proxy check entirely when both lists are empty, trusting the header unconditionally. That's exactly the inversion the gate's own comment (#910) warns about.

## Failure scenario

An operator sets `KnownProxies: ["gameserver-proxy.railway.internal"]` (a hostname — plausible on Railway, whose private networking is DNS-based; see #1786) or puts a CIDR in `KnownProxies` instead of `KnownNetworks`. Result: `HasTrustedProxies == true` → `UseForwardedHeaders()` runs → every entry dropped → any direct client spoofs `X-Forwarded-For`, defeating the per-IP auth rate limiter, the login-backoff partition key, and the login-tracking audit IPs (`ClientIp.Resolve`).

Note the silent skip is currently *pinned as correct* by `Game.Api.Tests/Unit/ForwardedHeadersConfigTests.cs` (`Apply_SkipsUnparseableEntries`), so nothing fails today. Closed #910/#1236 covered adjacent aspects but not this inversion.

## Suggested fix

Fail fast at startup when a configured entry doesn't parse (`.Validate(...).ValidateOnStart()`, matching the project's fail-fast options pattern), and/or derive `HasTrustedProxies` from the **parsed** results so unparseable-only config keeps the middleware off. Update the test to pin the new behavior.

ginderjeremiah:
## Summary

`ForwardedHeadersConfig.Apply` (`Game.Api/Forwarding/ForwardedHeadersConfig.cs`) silently drops any `KnownProxies`/`KnownNetworks` entry that fails to parse, while `HasTrustedProxies` counts the **raw config strings** rather than the parsed results. A config whose entries are *all* unparseable therefore still reports `HasTrustedProxies == true` — enabling the forwarded-headers middleware (`Startup.cs` gates on it) — but with **empty** framework trust lists. ASP.NET Core's `ForwardedHeadersMiddleware` skips its known-proxy check entirely when both lists are empty, so it ends up trusting `X-Forwarded-For` unconditionally — the exact inversion the gate's own comment warns against.

An operator sets `KnownProxies: ["gateway.railway.internal"]` (a hostname — plausible on Railway, whose private networking is DNS-based) or puts a CIDR in `KnownProxies` instead of `KnownNetworks`. Every entry silently fails to parse and is dropped, yet the middleware still activates with an empty trust set — so any direct client can spoof `X-Forwarded-For`, defeating the per-IP auth rate limiter, the login-backoff partition key, and the login-tracking audit IPs (`ClientIp.Resolve`).

## Changes

- `Game.Api/Forwarding/ForwardedHeadersConfig.cs`: new `AllEntriesParse` property — true only when every configured `KnownProxies`/`KnownNetworks` entry actually parses.
- `Game.Api/Startup.cs`: the existing `AddOptions<ForwardedHeadersConfig>()` chain (which already validates `ForwardLimit` with `ValidateOnStart`) gets a second `.Validate(...)` rule requiring `AllEntriesParse`, matching the project's fail-fast options pattern — a misconfigured deployment now fails at boot instead of silently degrading its trust model.
- `docs/infrastructure.md`: documents the new fail-fast behavior under "Trusted reverse proxies".
- `ForwardedHeadersConfig.Apply` itself is unchanged (its skip-on-unparseable behavior is still exercised by the existing `Apply_SkipsUnparseableEntries` test) — with the new startup validation in place, an unparseable entry never actually reaches it in a real deployment; the skip is now a defense-in-depth backstop rather than the primary behavior. Updated the test class's doc comment to say so.

## Tests

`Game.Api.Tests/Unit/ForwardedHeadersConfigTests.cs`:
- New `AllEntriesParse` unit tests (all-parse/empty → true; a bad `KnownProxies` entry, or a bare IP in `KnownNetworks` → false).
- New `ValidateOnStart_*` tests that re-declare the same bind+validate+`ValidateOnStart` registration Startup uses (mirroring the existing `CorsOptionsTests` pattern) and assert an unparseable entry throws `OptionsValidationException`, while valid config (including the "nothing configured" default) still resolves.

## Test plan

- [x] `dotnet test` — full solution suite passes: 3055/3055 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Api.Tests, Game.Application.Tests), including the new/updated `ForwardedHeadersConfigTests` cases.

Closes #1816

---
_Generated by [Claude Code](https://claude.ai/code/session_01BVnazM9jQmLVyCursK5zp3)_


ginderjeremiah (commented):
## Review: Fail fast on unparseable ForwardedHeaders proxy/network config entries (#1816)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the fix correctly closes the trust-model inversion, matches the project's fail-fast options pattern, and is thoroughly tested. No blocking findings, no should-fixes, no nits.

### Why this is the right fix

The issue offered two shapes (fail-fast validation, or derive `HasTrustedProxies` from parsed results). Fail-fast is the stronger and more consistent choice:

- It sits alongside the adjacent `ForwardLimit` validation and the `CorsOptions` `ValidateOnStart` registration, so it reads as an established pattern rather than a one-off.
- It's strictly *stronger* than deriving `HasTrustedProxies` from parsed results: a **partially** unparseable config (`["10.0.0.1", "gateway.internal"]`) now fails at boot instead of silently dropping the bad entry and running with a subtly-wrong trust set. The derive-only approach would have let that through.
- With `ValidateOnStart` in place, no unparseable config can reach the `HasTrustedProxies` gate at `Startup.cs:262`, so the raw-string count and the parsed framework lists can never diverge there — the exact inversion the issue describes is now unreachable. Leaving `Apply`'s skip as a documented defense-in-depth backstop (rather than deriving `HasTrustedProxies`) is the correct call and avoids redundant parsing logic.

### Implementation notes (all fine)

- `AllEntriesParse` reuses the same `IPAddress.TryParse` / `System.Net.IPNetwork.TryParse` calls as `Apply`, including the same full-qualification on `IPNetwork` (which avoids the `Microsoft.AspNetCore.HttpOverrides.IPNetwork` ambiguity) — consistent with the existing code.
- The `Apply` doc comment and the test-class summary were updated to reflect that the silent skip is now a backstop, not the primary behavior — good bookkeeping.

### Test coverage

Solid and follows the project's testing rules:

- `AllEntriesParse` unit cases cover all-parse/empty → true, a bad `KnownProxies` hostname → false, and a bare IP in `KnownNetworks` → false (the "CIDR in the wrong list" mistake the issue calls out).
- The `ValidateOnStart_*` cases re-declare the exact bind + both `Validate` rules + `ValidateOnStart` registration Startup uses (mirroring the `CorsOptionsTests` pattern) and assert `OptionsValidationException` on a bad entry while valid/empty config still resolves. The one minor drift risk — the validation lambdas/messages are duplicated between the test and `Startup.cs` rather than shared — is inherent to this test shape and matches the existing `CorsOptionsTests` precedent, so it's not worth changing.

### What's good

Tight, well-scoped security fix: the root cause (raw-string count vs. parsed trust lists) is correctly identified, the fix is the more defensible of the two suggested directions, the reasoning is captured in both the code comments and `docs/infrastructure.md`, and the previously-`Apply_SkipsUnparseableEntries`-pinned behavior is re-framed rather than left contradictory.

I'll merge once CI goes green (backend/verify/e2e are still running at review time).

---

PR #1845, opened 2026-07-11
Preserve the cached in-flight battle when re-entering the same character

### LINKED ISSUE #1818: SelectPlayer/SwitchPlayer unconditionally reset the cached PlayerState — a credential re-login destroys the same character's resumable in-flight battle
## Problem

`SessionService.Session.CreateSession` (`Game.Api/Services/SessionService.cs:125-132`) always primes the cache with a **fresh** `PlayerState`, even when the cache already holds the *same* player's state — discarding the cache-only in-flight battle snapshot. `SessionStore.cs:19-24` documents that snapshot as reconstructable-nowhere-else, and sized the TTL to 48h precisely so "an active player's in-flight battle should not be dropped".

## Failure scenario

Quit mid-battle, then log back in with credentials (logout, new device, or any expired-refresh path): `Login → SelectPlayer` runs `CreateSession` *before* any socket connects, so by the time `GetOfflineProgress` runs, `state.HasActiveBattle` is false — the still-in-progress hand-back (#1595) / replay-to-offset (#1597) machinery never triggers and the interrupted battle is dropped without credit. For a &lt;5-minute absence it isn't even covered by the away simulation (which no-ops below the floor, `OfflineProgressService.cs:114-122`).

The same battle *is* resumed after a plain socket reconnect, so behavior is inconsistent purely based on how the client re-entered. `SessionServiceTests` pins only that `CreateSession` primes the cache, not this loss.

## Suggested fix

In `CreateSession` (or at its call sites), load the existing cached state first and keep it when its `PlayerId` already matches the selected player; only reset when actually changing characters. Add a test: credential re-login with a cached in-flight battle for the same player preserves it.

ginderjeremiah:
## Summary

`SessionService.CreateSession` (`Game.Api/Services/SessionService.cs`) always primed the session cache with a **fresh** `PlayerState`, discarding whatever was already cached — even when the cache already held state for the *same* player. `SessionStore`'s TTL is deliberately sized at the refresh-token lifetime specifically so "an active player's in-flight battle should not be dropped" (it's a cache-only snapshot, reconstructable nowhere else).

Quit mid-battle, then log back in with credentials (logout, new device, or any expired-refresh path): `Login → SelectPlayer` runs `CreateSession` before any socket connects, so by the time offline progress resolves, `HasActiveBattle` is already false — the interrupted battle is dropped without credit, even though the exact same battle *is* resumed fine after a plain socket reconnect. Behavior was inconsistent purely based on how the client re-entered.

## Changes

- `Game.Api/Services/SessionService.cs`: `CreateSession` now reads the existing cached session first. When it already holds state for the same `playerId`, that cached state (and its in-flight battle snapshot, if any) is kept instead of overwritten; a genuinely different (or absent) cached player still gets a fresh `PlayerState`. This required making `CreateSession` async (it needs the cache read), so all call sites were updated to await it.
- `Game.Api/Controllers/LoginController.cs`: `SelectPlayer`/`SwitchPlayer` now `await` the call.
- Test call sites across `Game.Api.Tests` updated mechanically for the new async signature (`ApiIntegrationTestBase.CreateAuthenticatedClient` also became async, with its handful of callers updated).

## Tests

- `SessionServiceTests.cs`: new cases — no existing cached session still primes fresh state; a cached session for the *same* player is preserved (including its in-flight battle); a cached session for a *different* player still resets (there's nothing to preserve in that cache slot for the new character, since it only ever holds one player's state per account).
- `LoginControllerTests.cs`: new integration test `SelectPlayer_ReEntersSameCharacterWithCachedInFlightBattle_PreservesTheBattle` — seeds an in-flight battle into the session cache, re-logs-in with credentials (not a socket reconnect) and re-selects the same character, and asserts the battle snapshot survives.

## Test plan

- [x] `dotnet test` — full solution suite passes: 3058/3058 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), including the new tests above.

Closes #1818

---
_Generated by [Claude Code](https://claude.ai/code/session_01XGWayhWzakCdhsG5WsKAHK)_

### DIFF
--- a/Game.Api.Tests/Integration/AdminCacheReloadBroadcastTests.cs
+++ b/Game.Api.Tests/Integration/AdminCacheReloadBroadcastTests.cs
@@ -34,7 +34,7 @@ namespace Game.Api.Tests.Integration
             var context = scope.ServiceProvider.GetRequiredService<GameContext>();
             var user = await TestDataSeeder.CreateUserAsync(context, "adminuser", "adminpass");
             var player = await TestDataSeeder.CreatePlayerAsync(context, user.Id);
-            using var authClient = CreateAuthenticatedClient(user.Id, player.Id, nameof(ERole.Admin));
+            using var authClient = await CreateAuthenticatedClient(user.Id, player.Id, nameof(ERole.Admin));
 
             var pubsub = scope.ServiceProvider.GetRequiredService<IPubSubService>();
             var received = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
--- a/Game.Api.Tests/Integration/AdminToolsControllerTests.cs
+++ b/Game.Api.Tests/Integration/AdminToolsControllerTests.cs
@@ -30,7 +30,7 @@ namespace Game.Api.Tests.Integration
             await TestDataSeeder.LinkSkillToPlayerAsync(context, player.Id, skill.Id);
 
             var roles = admin ? new[] { nameof(ERole.Admin) } : [];
-            var client = CreateAuthenticatedClient(user.Id, player.Id, roles);
+            var client = await CreateAuthenticatedClient(user.Id, player.Id, roles);
 
             // Tests seed their reference data directly before calling this, so reload the caches to mirror
             // the production invariant that they are warm before any admin action. The admin write path
--- a/Game.Api.Tests/Integration/AdminUsersControllerTests.cs
+++ b/Game.Api.Tests/Integration/AdminUsersControllerTests.cs
@@ -49,7 +49,7 @@ namespace Game.Api.Tests.Integration
                 await TestDataSeeder.AssignRoleToUserAsync(context, user.Id, ERole.Admin);
             }
 
-            return (CreateAuthenticatedClient(user.Id, player.Id, roles), user.Id);
+            return (await CreateAuthenticatedClient(user.Id, player.Id, roles), user.Id);
         }
 
         private static async Task<AdminUserSearchResults> GetUsersAsync(HttpClient client, string query = "")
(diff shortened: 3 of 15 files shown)


ginderjeremiah (commented):
## Review: Preserve the cached in-flight battle when re-entering the same character (closes #1818)

**Verdict:** Approve — the fix is correct, minimally scoped, and its coverage exceeds what the issue asked for. (Submitted as a `COMMENT` only because GitHub won't accept a self-approval on my own PR; the intended verdict is Approve.)

The change matches the issue's suggested shape exactly: `CreateSession` now reads the existing cached `PlayerState` and keeps it only when `existing.PlayerId == playerId`, otherwise falls back to a fresh state. The identity guard is the right discriminator — a genuine character switch (different `PlayerId`) still resets, so a departed character's battle can't bleed onto the new one, and switching *back* onto the already-bound character correctly preserves it.

### Layering / correctness
- The decision stays a simple identity comparison in the session/orchestration layer — no domain logic leaks into `SessionService`, consistent with `docs/backend.md`. `GetSession` slides the TTL on the read, then `Update` re-writes and slides again; both are cheap and land on the login flow, not the concurrent socket write path.
- `CancellationToken` is threaded through and both call sites pass `HttpContext.RequestAborted` — good, and there's even a dedicated `LoadPlayerState_ForwardsCancellationTokenToStore`-style propagation assertion.
- The async signature migration across call sites is mechanical and correct; the `.GetAwaiter().GetResult()` uses are confined to genuinely-synchronous test helpers (or match the immediately-adjacent existing style), and the test doubles complete synchronously so there's no deadlock risk.

### Test coverage
Thorough and directly on-point:
- Unit (`SessionServiceTests`): no-existing → fresh; same player → in-flight battle preserved; different player → reset. This is exactly the matrix that matters.
- Integration (`LoginControllerTests.SelectPlayer_ReEntersSameCharacterWithCachedInFlightBattle_PreservesTheBattle`): seeds an in-flight battle into the real session store, does a full credential re-login (not a socket reconnect), and asserts the snapshot survives — reproducing the exact #1818 scenario end-to-end.

### What's good
Clean, surgical fix with a genuinely useful XML-doc summary explaining the non-obvious preservation behavior so a later edit doesn't naively "simplify" the identity guard away. No blocking findings, no should-fixes, no nits.

I'll merge once the remaining CI jobs (`test-backend` and the firefox/webkit e2e runs) finish green.

---

PR #1846, opened 2026-07-11
Close the remaining #1632 bypass paths in the player-reload contract

### LINKED ISSUE #1819: Close the remaining #1632 bypass paths: a domain-event dispatch fault and a standalone progress-flush failure skip the player-reload contract
#1632 established the contract: when a save's persistence flush fails, the failure is wrapped in `PlayerPersistenceFlushFailedException` so `SocketHandler` (`SocketHandler.cs:244-253`) reloads the connection's in-memory `Player` instead of silently carrying un-persisted mutations forward. Two paths still bypass it:

## 1. A dispatch-handler fault in `SavePlayer` escapes unwrapped

`PlayerRepository.SavePlayer` (`Game.DataAccess/Repositories/PlayerRepository.cs:94-97`) runs `_dispatcher.DispatchAsync` **outside** the try/catch that wraps only `FlushAsync` (lines 105-116). `DomainEventDispatcher` snapshots and **clears events off the aggregate before handlers run**, and rethrows handler failures as `AggregateException` after the drain (`Game.Core/Events/DomainEventDispatcher.cs:48-61`, `104-111`).

**Scenario:** `DefeatEnemy` → `EndBattleVictory` → `SavePlayer` dispatches `BattleCompletedEvent`; `BattleStatisticsEventHandler` (progress Redis read/save) or an API-layer notifier hits a transient Redis/DB error. The victory's mutations (exp, milestone `SkillUnlocked`, challenge-reward `ItemUnlocked`) already happened in-memory and their envelopes were buffered — all lost when the command scope disposes. No reload marker is set, so the mutations silently ride along; the next command's save writes the full player blob to Redis (masking the divergence) while the DB never receives the one-shot unlock rows. When the Redis key ages out (~48h idle), the DB fallback rehydrates without the item/skill — silent, long-delayed player-data loss. Absolute-value fields self-heal; one-shot child rows do not.

## 2. The standalone progress flush isn't wrapped either

`PlayerProgressRepository.Save`'s standalone path (`PlayerProgressRepository.cs:120-127`) lets a `FlushAsync` failure propagate raw — note `AcceptChanges()` already ran by then. The offline/switch flow mutates the player (`GrantOfflineExp`, `GrantRewardSkills`, `CompleteChallenge` in `OfflineProgressService.ApplyOfflineRewards`) **before** `_progressRepo.Save`, and `StampActivity` + `SavePlayer` run after.

**Scenario:** Redis blips on the progress flush → `GetOfflineProgress` faults; `LastActivity` was never re-anchored and the connection's in-memory `Player` keeps the window's exp/levels. A same-connection re-claim re-simulates the same window onto the already-credited aggregate → **double exp/levels**, made durable by the next successful save's absolute-value events.

This also falsifies `docs/backend-persistence.md`'s claim that a standalone progress save "flushes its own event, so it is never stranded" — update that line with the fix.

## Suggested fix

Wrap both paths in `PlayerPersistenceFlushFailedException` (for the dispatch, either move it inside the same non-cancellation catch, or catch dispatch faults, still flush the already-buffered envelopes, then rethrow wrapped) so the existing reload machinery discards un-persisted in-memory state. Add tests at both seams — no current test covers a dispatch-handler fault (`PlayerWriteBehindTests` pins only the flush wrap).

ginderjeremiah:
## Summary

#1632 established the contract: when a save's persistence flush fails, the failure is wrapped in `PlayerPersistenceFlushFailedException` so `SocketHandler` reloads the connection's in-memory `Player` instead of silently carrying un-persisted mutations forward. Two paths still bypassed it:

1. **`PlayerRepository.SavePlayer`** ran `_dispatcher.DispatchAsync` outside the try/catch that wraps only `FlushAsync`. A handler fault (e.g. a transient Redis/DB error in `BattleStatisticsEventHandler` or a notifier) escaped unwrapped, and — because it happened before `FlushAsync` ran — the whole batch (including any envelopes other handlers had already buffered) was discarded when the command scope disposed, with no reload marker set.
2. **`PlayerProgressRepository.Save`**'s standalone path (the offline-rewards batch, reached *outside* a player save) let a `FlushAsync` failure propagate raw. A Redis blip there could leave `LastActivity` un-anchored while the in-memory `Player` already held the away-window's granted exp/levels, risking a same-connection re-claim double-crediting the same window.

## Changes

- `Game.DataAccess/Repositories/PlayerRepository.cs`: `SavePlayer` now catches a non-cancellation dispatch fault, still runs `FlushAsync` (so whatever the batch already buffered isn't lost), then rethrows the dispatch fault wrapped in `PlayerPersistenceFlushFailedException` — skipping the cache blob write, exactly like an existing flush failure does.
- `Game.DataAccess/Repositories/PlayerProgressRepository.cs`: the standalone-save branch wraps a non-cancellation `FlushAsync` failure the same way.
- `Game.Abstractions/DataAccess/PlayerPersistenceFlushFailedException.cs`: doc comment updated to reflect both new throw sites.
- `docs/backend-persistence.md`: updated the flush-failure paragraph and removed the now-false "a progress save outside a player save... is never stranded" claim (the exact line #1819 called out), describing the fixed behavior instead.

## Tests

- `Game.Application.Tests/Services/PlayerWriteBehindTests.cs`: new `SavePlayer_DispatchFaultsForANonCancellationReason_StillFlushesBufferedEnvelopes_AndThrowsPlayerPersistenceFlushFailedException`, using a stub dispatcher that buffers an envelope (standing in for a sibling handler that already succeeded) before throwing — pins that the envelope still reaches the queue and the cache blob write is skipped.
- `Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs`: new `Save_StandalonePublishFailsForANonCancellationReason_ThrowsPlayerPersistenceFlushFailedException`, pinning the wrap and that the cache isn't advanced past the failed flush.

## Test plan

- [x] `dotnet test` — full solution suite passes: 3060/3060 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Api.Tests, Game.Application.Tests), including the two new tests above.

Closes #1819

---
_Generated by Claude Code_

---
_Generated by [Claude Code](https://claude.ai/code/session_015tbPKdy88Vxxcu8ZSfpxHs)_


ginderjeremiah (commented):
## Review: Close the remaining #1632 bypass paths in the player-reload contract

**Verdict:** Approve (submitted as a comment since I'm the PR author — GitHub won't accept a self-approval) — both bypass paths are closed correctly, the fix stays within the established `PlayerPersistenceFlushFailedException` contract, docs are updated, and both seams are tested.

### What it does, verified against #1819
- **`PlayerRepository.SavePlayer`** — the dispatch fault is now captured (non-cancellation only), the flush still runs so sibling handlers' already-buffered envelopes aren't discarded, and the fault is then rethrown wrapped **before** the cache blob write. That ordering is the important part: the stale in-memory blob is not written, and `SocketHandler` gets the reload marker. Correct.
- **`PlayerProgressRepository.Save`** — the standalone path now wraps a non-cancellation `FlushAsync` failure the same way, closing the offline-rewards double-credit window. Cancellation still propagates unwrapped on both paths, preserving the cooperative-unwind semantics.
- The `catch ... when (ex is not OperationCanceledException)` guard is used consistently on all three sites, so a cooperative cancel never triggers a spurious reload.

### Test coverage
- `PlayerWriteBehindTests.SavePlayer_DispatchFaultsForANonCancellationReason_...` pins the subtle bit the issue asked for: the sibling envelope buffered before the fault still reaches the queue (length 1), the cache blob write is skipped (reread sees the pre-mutation zone), and the exception/inner types are asserted. Good seam test.
- `PlayerProgressRepositoryIntegrationTests.Save_StandalonePublishFailsForANonCancellationReason_...` pins the wrap, inner type, and that the cache didn't advance past the failed flush.
- Both mirror the issue's "add tests at both seams" ask; no battle logic touched, so no frontend mirror is required.

### Nits (non-blocking, take-or-leave)
- In `SavePlayer`, if the flush *also* fails after a captured dispatch fault, the flush's exception wins and `dispatchFault` is dropped entirely (not even chained). The observable outcome is identical — both force the same reload — so this is correct behavior, not a defect; just noting the dispatch fault is invisible to diagnostics in that rare double-failure case.

### What's good
- The fix is minimal and stays squarely inside the existing contract rather than inventing a new mechanism. The doc update in `backend-persistence.md` correctly removes the now-false "never stranded" claim that #1819 called out, and the inline comments accurately explain the non-obvious "capture, flush anyway, then rethrow" control flow — genuinely useful given how easy that ordering would be to naively "simplify" and break.


---

PR #1847, opened 2026-07-11
Fix sync BackgroundWorker loop running with IsRunning == false

### LINKED ISSUE #1822: Sync BackgroundWorker loop can run with IsRunning == false — fix the race or delete the production-unused sync path
## Problem

The sync worker loop (`Game.Infrastructure/BackgroundWorker.cs:168-184`) never re-asserts `IsRunning = true` when a run starts, so a `Start()` that lands between a finishing run's `loopAction()` and its trailing `IsRunning = false` (line 181) produces a follow-up run that executes entirely with `IsRunning == false`:

1. Run 1's action completes.
2. `Start()` sets `IsRunning = true` and `Set()`.
3. Run 1's tail writes `IsRunning = false`.
4. The re-armed wait fires run 2 — which runs with `IsRunning` false throughout.

The async loop closes exactly this window by reaffirming `IsRunning = true` under `_runLock` (lines 200-204, whose comment describes the same clobber); the sync loop got no equivalent.

## Impact

Production only constructs the async variant (`RedisPubSubService.cs:123`), so today's impact is a violated public contract plus latent flake in `BackgroundWorkerTests` — several tests gate on `WaitUntil(() => !worker.IsRunning)` before asserting invocation counts.

## Suggested fix

Either mirror the async loop's reaffirmation in the sync loop, or — given nothing in production uses it — delete the sync ctor/loop outright and simplify. `docs/backend-persistence.md`'s BackgroundWorker ordering guarantee should get a caveat or become unconditionally true as part of the fix.

ginderjeremiah:
## Summary

`ThreadPool.RegisterWaitForSingleObject` re-arms its wait independent of whether the callback delegate has finished running — not only for the async overload (where an incomplete `await` was the known trigger, fixed for #1136), but for the synchronous overload too. A `Start()` landing while a sync run is still executing could dispatch a genuinely concurrent second run; that new run's optimistic `IsRunning = true` then gets clobbered by the finishing run's trailing `IsRunning = false`, so the newly-dispatched run executes with `IsRunning` false throughout — violating `BackgroundWorker`'s own public contract (and the invariant several `BackgroundWorkerTests` gate on via `WaitUntil(() => !worker.IsRunning)`).

Production only constructs the async overload (`RedisPubSubService`), so this was latent rather than user-facing, per the issue.

## Fix shape

I initially tried the issue's first suggested shape — just reaffirming `IsRunning = true` at the top of the sync callback, mirroring the async loop's reaffirmation. A stress test proved that insufficient: the sync callback really can run concurrently (see below), so a bare reaffirm without mutual exclusion still raced.

I ended up going further than either literal option the issue offered ("mirror the reaffirmation" or "delete the sync path"): since the actual bug is that the sync loop never had the async loop's run-serialization gate (`_runLock`/`_runActive`/`_runPending`) at all, I unified both constructor overloads onto that one gated loop instead — the sync overload now wraps its delegate as `() => { action(); return Task.CompletedTask; }` and shares the exact same loop implementation. This fixes the race with already-hardened, already-tested logic and deletes the now-redundant duplicate loop, rather than adding a second bespoke concurrency mechanism. It also keeps the sync constructor's public API and the existing `BackgroundWorkerTests` (which exercise the sync/async overloads in parallel via `[InlineData]`) intact.

## Changes

- `Game.Infrastructure/BackgroundWorker.cs`: sync constructor now delegates to the `Func<Task>` constructor with a completed-task wrapper; `CreateWorkerLoop`/`CreateAsyncWorkerLoop` collapse into one `CreateWorkerLoop(Func<Task>)` shared by both overloads.
- `docs/backend-persistence.md`: documents that both overloads now share one gated loop and why.

## Tests

- `Game.Infrastructure.Tests/BackgroundWorkerTests.cs`: new `SyncWorker_StartRacingRunCompletion_NeverObservesNotRunningAtRunStart` — several threads spin `Start()` against a fast, non-blocking sync run and assert no run ever observed `IsRunning == false` at its start. Verified this reliably reproduces the bug against the pre-fix code (thousands of failing observations per run) and passes cleanly (3 consecutive runs, 0 failures) against the fix.
- All existing `[InlineData(false)]`/`[InlineData(true)]` sync/async paired tests in the file continue to pass unmodified.

## Test plan

- [x] `dotnet test` (no-build) — full solution suite passes: 3061/3061 (`Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Application.Tests`, `Game.Api.Tests`).
- [x] New race test run 3x consecutively post-fix with 0 failures; confirmed it fails against the pre-fix code (reproduces the bug).

Closes #1822

---
_Generated by [Claude Code](https://claude.ai/code/session_01N8pGEzwLh9ZZB926n143nZ)_


ginderjeremiah (commented):
## Review: Fix sync `BackgroundWorker` loop running with `IsRunning == false` (#1822)

**Verdict:** Approve — the fix is correct, addresses the actual race, and is well-tested. (Submitted as a comment because I'm the PR author; GitHub won't accept a gating approval on one's own PR.)

The change unifies both constructor overloads onto the single, already-hardened gated loop rather than taking either literal option the issue offered. I verified this is the right call:

- **The issue's premise was partly wrong, and the PR correctly caught it.** #1822 assumed the sync callback runs strictly serially ("returns only once `loopAction` completes"), so a bare `IsRunning = true` reaffirm would suffice. But `ThreadPool.RegisterWaitForSingleObject` re-arms its wait when the wait is *satisfied*, not when the callback *returns* — so a `Set()` landing mid-run can dispatch a genuinely concurrent second sync invocation on another pool thread. The stress test proves this. Mirroring the reaffirmation alone would still have raced; unifying onto the `_runLock`/`_runActive`/`_runPending` gate is the correct fix.
- **Behavior is preserved for the sync path.** The wrapper `() => { action(); return Task.CompletedTask; }` runs `action()` synchronously and the `await` on a completed task never suspends, so the sync delegate still executes inline on the callback thread as before — now with the gate. A synchronous throw propagates from the `loopAction()` invocation inside the `try`, so it's still caught and logged.
- **No dangling references** to the removed `CreateAsyncWorkerLoop` / `CreateWorkerLoop(Action)` — both were private and are fully collapsed into the one `CreateWorkerLoop(Func<Task>)`.

### Test coverage

- The new `SyncWorker_StartRacingRunCompletion_NeverObservesNotRunningAtRunStart` reproduces the race directly (4 starter threads × 3000+ invocations) and asserts `NotRunningAtStartCount == 0`; the PR notes it reliably fails pre-fix and passes post-fix. The existing `[InlineData(true/false)]` sync/async paired tests continue to cover both overloads unchanged. Coverage is appropriate for this infra change; no battle-logic mirroring applies.

### What's good

- Went past the issue's two literal suggestions to the root cause (the sync loop never had the gate at all), reusing battle-tested logic instead of adding a second bespoke concurrency mechanism, and deleting the redundant duplicate loop. `docs/backend-persistence.md` is updated to make the ordering guarantee unconditional across both overloads, exactly as the issue asked.

No blocking findings, no should-fixes, no nits. Will merge once CI is fully green (`test-backend` / `test-e2e (webkit)` still in progress).

---

Issue #1771, opened 2026-07-09
Test gaps on auth-critical request paths (refresh-fails-401, isRetry loop guard)


ginderjeremiah:
Targeted unit-test gaps on the frontend auth machinery, found while reviewing the refresh-failure semantics (#1763):

- `UI/src/tests/lib/api/api-request.test.ts`: no test for a 401 whose refresh **fails** (expected: `handleAuthFailure` + the 401 returned), and none for the `isRetry` guard (`api-request.ts:107`) — the single check preventing an infinite refresh/retry loop.
- `UI/src/tests/lib/api/auth.test.ts`: no network-throw scenario at all — the exact case whose semantics #1763 will decide; whatever is chosen should be pinned here.
- `UI/src/tests/lib/api/api-socket.test.ts:162-174` drives the pre-emptive-refresh-failure connect path but doesn't assert the queued command's promise settles (covered by #1766's fix).

_Found during a codebase health review._

ginderjeremiah:
Picked this up and found all three gaps already closed by later work — verifying rather than duplicating:

1. **`api-request.test.ts`: 401-whose-refresh-fails + `isRetry` guard** — covered by `'clears tokens and redirects when the refresh is definitively rejected (400)'` (line 235) and `'preserves tokens and does not redirect when the refresh is retryable (5xx)'` (line 255), both added in #1795. Both assertions explicitly pin `callsTo('/api/Login/Refresh')` to length 1, i.e. the `isRetry` guard.
2. **`auth.test.ts`: no network-throw scenario** — covered by `'keeps storage and returns retryable on a network error (fetch throws)'` (line 89), also from #1795.
3. **`api-socket.test.ts:162-174` queued-command-promise-settles gap** — covered by `'routes to auth failure (without opening) when the pre-emptive refresh is definitively rejected'` (line 165), which explicitly asserts `await expect(promise).resolves.toMatchObject({ error: expect.any(String) })`. This landed via #1766 as the issue body itself anticipated.

Ran all three files: 86/86 passing (`npx vitest run src/tests/lib/api/api-request.test.ts src/tests/lib/api/auth.test.ts src/tests/lib/api/api-socket.test.ts`).

No remaining gap to fill — closing as already resolved by #1795/#1766.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1867, opened 2026-07-11
Close the residual cross-tab refresh race in handleAuthFailure's callers

### LINKED ISSUE #1764: Cross-tab refresh race can burn the single-use refresh token and log out every tab

### LINKED ISSUE #1829: Residual cross-tab refresh-token race window remains in handleAuthFailure
## Context

Follow-up from PR #1796 (closed #1764), which made `refreshTokens()` re-read `localStorage` before clearing: on a failed refresh, if storage now holds a *different* refresh token than the one presented, the losing tab adopts the winner's rotated pair instead of clearing it. This is a strict improvement over the previous unconditional clear.

## Problem

A narrow residual window remains. If the losing tab re-reads storage **before** the winning tab persists its rotated pair (`setTokens`), it still sees the same token it presented, clears, and returns null. The caller's `handleAuthFailure()` (`api-request.ts` / `api-socket.ts`) then calls `clearTokens()` again — which can wipe the winner's freshly-rotated pair after it lands, logging both tabs out. The `refreshTokens()` re-read guard does not cover this second `clearTokens()` call site.

## Suggested fix (either)

- Add the same re-read guard inside `handleAuthFailure()` before its own `clearTokens()` — only clear if storage still holds a token consistent with a dead session.
- Or serialize the cross-tab refresh with `navigator.locks` (the other option offered in #1764) so only one tab presents the single-use token at a time, eliminating the race at the source.

## Also worth pinning (test nit from the review)

Add a test for the negative branch of #1796's guard — storage still holds the *presented* token ⇒ clears — independent of the existing race test.

## Note

PR #1796's description overstates the fix as "directly and completely closes the described race"; it is a strong partial mitigation. Worth softening if the wording is reused.

ginderjeremiah:
## Summary

PR #1796 (closed #1764) made `refreshTokens()` re-read `localStorage` before clearing: on a failed refresh, if storage now holds a *different* refresh token than the one just presented, the losing tab adopts the winner's rotated pair instead of clearing a session that's still alive elsewhere.

That guard only covers `refreshTokens()`'s own `clearTokens()` call. Every caller that treats a failed refresh as terminal — `ApiRequest.execute`'s 401 retry, `ApiSocket.openSocket`'s pre-emptive refresh, and `ApiSocket.handleClose`'s auth-retry reconnect — then calls `handleAuthFailure()`, which clears tokens **again**, unconditionally. If a winning tab's rotated pair lands in the gap between `refreshTokens()` returning and that second clear, `handleAuthFailure()` wipes it out, logging both tabs out even though a live session existed.

## Fix

Each of the three call sites now re-reads storage immediately before treating the refresh failure as terminal, and recovers using the fresh pair instead of tearing the session down when one is present:

- `api-request.ts`: if `getTokens()` is non-null after a failed refresh, retry the request once more (reusing the existing single-retry mechanism) instead of calling `handleAuthFailure()`.
- `api-socket.ts` `openSocket`: if the pre-emptive refresh returned no token but storage now holds one, use it to open the handshake instead of routing to auth failure.
- `api-socket.ts` `handleClose`: if the reconnect refresh returned no token but storage now holds one, reconnect (`processCommandQueue`) instead of routing to auth failure.

`handleAuthFailure()` itself is unchanged (still an unconditional clear+redirect) — the docblock now notes that callers are responsible for checking storage first, since it has no way to distinguish "genuinely dead" from "a concurrent tab just rotated in a fresh pair" on its own. The `socketAuthRetries`-exhausted branch (a different, legitimate case where the session is deemed unrecoverable due to repeated handshake rejection regardless of token freshness) is intentionally left unguarded — it's out of scope for this race.

Also updated `docs/frontend.md`'s Authentication section to describe the new caller-side guard.

## Tests

- `api-request.test.ts`: new case pinning the previously-untested "401 whose refresh is definitively rejected" path (tokens cleared, redirect to `/`) — this gap was also called out by #1771. New case for the residual race: `refreshTokens` resolves null but a winner's pair is present, so the request retries and succeeds, tokens are left intact, no redirect.
- `api-socket.test.ts`: new case for `openSocket`'s pre-emptive refresh recovering via a concurrent tab's rotated pair. New case for `handleClose`'s reconnect path recovering the same way. Updated the three existing "refresh fails" tests to have their mocks clear tokens as a side effect (matching what the real `refreshTokens`/`ensureValidAccessToken` actually do on a definitive failure) — without that, the mocks left stale tokens sitting in storage, which the new guard would (correctly, per its contract) mistake for a live session.

## Test plan

- [x] `npx vitest run` — full suite passes: 3611/3611 (including the new/updated cases above).

Closes #1829

---
_Generated by [Claude Code](https://claude.ai/code/session_017DgXAGQYZpAuKCnKDj5oBt)_


ginderjeremiah (commented):
## Review: Close the residual cross-tab refresh race in `handleAuthFailure`'s callers (#1829)

**Verdict:** Approve _(submitted as a comment — GitHub won't accept a gating verdict from the PR author; treat this as an approval)_ — the residual window described in #1829 is genuinely closed, the callers are all accounted for, and the tests faithfully model the real behavior.

### Analysis

The core insight is right: #1796's guard only covered `refreshTokens`'s own `clearTokens()`, leaving `handleAuthFailure()`'s second, unconditional clear as a live race window. This PR re-reads storage **as late as possible** at each terminal site — synchronously, immediately before the clear, with no `await` in between — so a winner's rotated pair that lands before the re-read is adopted, and one that lands after can't be wiped because the clear already ran. The only remaining gap is a synchronous one another tab cannot interleave. That's a real fix, not just a further narrowing.

All four `handleAuthFailure` call sites check out:
- `api-request.ts:116` — `getTokens() !== null` → retry (reuses the single-retry cap). ✓
- `api-socket.ts:145` (`openSocket`) — adopts `getTokens()?.accessToken`. ✓
- `api-socket.ts:423` (`handleClose`) — `tokens || getTokens() !== null` → reconnect. ✓
- `api-socket.ts:408` (retries exhausted) — deliberately unguarded, and correctly so: it only fires after MAX handshake rejections of the *shared* freshly-refreshed token, a state in which no live cross-tab session can exist.

Keeping the guard at the three call sites (rather than inside `handleAuthFailure`, per the issue's option 1) is a defensible choice — each site recovers differently, and it keeps the retries-exhausted branch giving up. The `handleAuthFailure` docblock now documents the caller contract clearly.

### Test coverage

Solid and honest:
- New recovery cases at all three guarded sites (`api-request` retry, `openSocket` adopt, `handleClose` reconnect).
- The definitive-rejection negative branch is pinned in `api-request.test.ts` (the issue's requested test); the `refreshTokens`-level negative branch was already covered in `auth.test.ts:52`.
- The three existing "refresh fails" mocks were correctly updated to clear storage as a side effect, matching what the real `refreshTokens`/`ensureValidAccessToken` do — without which they'd have left stale tokens that the new guard would (correctly) read as a live session.

### What's good

Careful reasoning about the exact ordering window, a documented and internally-consistent decision to leave the retries-exhausted branch unguarded, and docs updated in `frontend.md`. No blocking findings, no should-fixes, no nits.

---

PR #1868, opened 2026-07-11
Register SocketReplaced app-wide instead of scoping it to the game engine

### LINKED ISSUE #1836: SocketReplaced session-takeover push goes unhandled while browsing outside /game (e.g. /admin)
## Context

Follow-up to #1807 / PR #1835. That PR correctly fixed `stopEngines` (the `/game` route's actual unmount cleanup) to unhook the four `startGame`-registered socket listeners — `SocketReplaced`, `ChallengeCompleted`, `ProficiencyXpGained`, `ServerCommandFailed` — instead of leaking them across remounts.

## Problem

`SocketReplaced` (the server pushing "another session has started elsewhere, disconnect this one") has **no handler outside the game engine's lifecycle** — `handleSocketReplaced` in `UI/src/lib/engine/engine.ts` is the only place that ever calls `apiSocket.listenCommand('SocketReplaced', ...)`, and it's now correctly scoped to `startGame`/`stopEngines`.

So: navigate `/game` → `/admin` (or any other route). The game page unmounts, `stopEngines()` runs, and now correctly unhooks the `SocketReplaced` listener along with the other three. If a session takeover push arrives while the user is sitting on `/admin`, nothing is listening for it — no acknowledgement modal, no redirect to login, and `apiSocket.disconnect()` (which tears down the keepalive/socket) never runs. The stale session sits there until some other path happens to notice (e.g. the keepalive eventually detects a half-open socket, or an unrelated request 401s).

Before PR #1835, issue #1807's listener leak coincidentally masked this: the leaked `SocketReplaced` listener from the last `/game` visit stayed registered while browsing `/admin`, so a takeover was still (accidentally) caught. Fixing the leak correctly removed that accidental coverage.

## Suggested direction

Decide whether session-replacement handling belongs at the root/layout level — registered once for the whole authenticated app, independent of whether `/game`'s engines are running — rather than scoped to the game engine's lifecycle. That would close this gap cleanly (and stop the app-wide "another session took over" UX from coincidentally depending on which route happens to be mounted).

No user-facing effect today outside the edge case of a session takeover while parked on a non-`/game` route (most likely `/admin`, since that's the only other authenticated destination `handleNavigate` routes to).

_Found while reviewing PR #1835._

ginderjeremiah:
## Summary

`SocketReplaced` — the server pushing "another session has started elsewhere, disconnect this one" — was only ever listened for while the game engine's own lifecycle owned it: `startGame` registered it, `stopEngines` unhooked it. Navigating `/game` → `/admin` (or any other authenticated route) unmounts the game page, which correctly unhooks the listener along with the game-scoped ones (`ChallengeCompleted`, `ProficiencyXpGained`, `ServerCommandFailed`) — but that also silences the takeover push while the player is sitting on `/admin`. No acknowledgement modal, no redirect to login, and the keepalive socket never tears down; the stale session just sits there.

## Fix

Per the issue's suggested direction, session-replacement handling now belongs at the root/layout level: `apiSocket.listenCommand('SocketReplaced', handleSocketReplaced, true)` is registered once in `+layout.svelte`, right alongside the existing app-wide `onSocketError` subscription — independent of whether the game engines are running.

The other three `startGame`-registered pushes stay scoped to the game engine's lifecycle (unchanged), since they only matter while its stores are live.

`handleSocketReplaced`/`stopGame` needed no changes — `stopEngines()` (and the rest of `stopGame`'s teardown) is already documented as idempotent/safe to call even when the engines never started, so calling it from outside `/game` works as-is.

## Changes

- `UI/src/routes/+layout.svelte`: registers the app-wide `SocketReplaced` listener.
- `UI/src/lib/engine/engine.ts`: removes the `SocketReplaced` registration/unhook from `startGame`/`stopEngines` (now three game-scoped listeners, not four).
- `docs/frontend.md`: documents the app-wide vs. game-scoped listener split under *Socket connection lifecycle*.
- Tests updated/added in `engine.test.ts` (assert `startGame` no longer wires `SocketReplaced`) and `layout.svelte.test.ts` (assert the root layout registers it once, app-wide).

## Test plan

- [x] `npx vitest run` (full suite) — 3612/3612 passing.
- [x] `npm run build` + a manual SSR smoke request against the built handler — root layout renders (200) with the new registration wired in.

Closes #1836

---
_Generated by [Claude Code](https://claude.ai/code/session_01CRZGJHADw1Purf42tbPJvP)_


ginderjeremiah (commented):
## Review: Register `SocketReplaced` app-wide instead of scoping it to the game engine (#1836)

**Verdict:** Approve — clean, correctly-scoped fix that closes the takeover-push gap exactly as the issue's suggested direction proposed. (Posted as a comment since GitHub won't accept a gating self-review; the intended verdict is Approve.)

I verified the change against the code rather than the description:

- **Correct hook lifecycle.** `apiSocket.listenCommand('SocketReplaced', handleSocketReplaced, true)` is registered in the root layout's top-level script body — genuine Svelte component init — so passing `cleanupOnDestroy: true` is exactly what `listenCommand`'s contract requires (`api-socket.ts:184-188`: "only pass `true` from within Svelte component init"). It mirrors the `onSocketError(..., true)` registration directly above it, and the root layout lives for the whole session so cleanup only fires on app teardown. This avoids the `onDestroy`-throws-outside-component-init problem that forced `startGame` to capture manual unhooks.
- **Handler is safe to fire off-`/game`.** `handleSocketReplaced` → `stopGame` → `stopEngines`/`disconnect` are all documented and implemented as idempotent and safe when the engines never started (`engine.ts:339-367`), so a takeover push while parked on `/admin` tears down correctly. No double-registration now that `startGame` no longer wires it.
- **No coverage lost.** The deleted `startGame` tests only exercised the now-removed registration path; `handleSocketReplaced`'s behavior (stops engines, resets logs, shows the modal, navigates only after ack) is still fully covered by its own `describe` block (`engine.test.ts:232-267`). The new `layout.svelte.test.ts` case pins the app-wide registration with the `true` flag, and the `afterEach(stopEngines())` addition sensibly prevents module-level unhook state leaking between tests.
- **Docs updated.** The app-wide-vs-game-scoped split is documented under *Socket connection lifecycle* in `frontend.md`, per the project's documentation rule.

### Test coverage

Adequate. Both the removal (`startGame` no longer wires `SocketReplaced`) and the addition (root layout registers it once, app-wide, with `cleanupOnDestroy`) are asserted, and the pre-existing `handleSocketReplaced` behavior tests remain intact.

### What's good

Tight scope, the other three game-scoped listeners are correctly left alone, comments explain the *why* (component-init requirement, off-`/game` teardown) without over-explaining, and the fix matches the issue's analysis precisely.

Merging once CI is green (frontend/e2e jobs still finishing).

---

PR #1869, opened 2026-07-11
Mark the connection's player for reload on a command that settles via cancellation

### LINKED ISSUE #1849: Per-command timeout during SavePlayer drops buffered write-behind events without forcing the player reload
## Problem

The #1632/#1819 "flush failed → force in-memory `Player` reload" safety net deliberately carves cancellation out, but the carve-out is only sound for *teardown* cancellation, not the **per-command timeout** — and the two paths can't be told apart where the carve-out lives.

In `PlayerRepository.SavePlayer`:

- `PlayerRepository.cs:107` — the dispatch fault capture is `catch (Exception ex) when (ex is not OperationCanceledException)`, so a cancelled dispatch propagates OCE and **skips the flush entirely**.
- `PlayerRepository.cs:123` — the flush wrap has the same filter, so a cancelled flush propagates unwrapped instead of becoming `PlayerPersistenceFlushFailedException`.

And in `SocketHandler`:

- `ReleaseCommandLockWhenSettled` (`SocketHandler.cs:319-326`) — the abandoned-on-timeout continuation marks `PlayerNeedsReload` only for `PlayerPersistenceFlushFailedException`, and explicitly treats `OperationCanceledException` as "the expected" outcome. Nothing is marked.

The chain: `RunCommandUnderLock` cancels the per-command CTS at the 25s budget and that token is threaded into every `SavePlayer`. The domain-event dispatcher snapshots and **clears the events off the aggregate before handlers run** (`AggregateRoot`/`DomainEventDispatcher`), so once they're buffered into the **scoped** `PlayerUpdateBatch`, the batch is their only holder. `FlushAsync`'s preserve-on-failure only helps the *next flush in the same scope* — but `SocketHandler.RunCommand` opens a fresh DI scope per command, so when the cancelled save unwinds the command, the scope is disposed and the envelopes are gone. The mutated in-memory `Player` lives on, `PlayerNeedsReload` is never set, and the next command's save writes the full cache blob **including** the mutation — Redis (the rehydration source of truth) now permanently holds state whose durable event never reached the queue.

## Failure scenario

A Redis/DB stall >25s (exactly the transient-outage class the flush-failure reload was built for — the two are correlated) cancels a battle-completion command between mutation and enqueue. A challenge-reward `ItemUnlockedEvent`/`SkillUnlockedEvent`/`ModUnlockedEvent`/`LessonUnlockedEvent` raised in that save is **one-shot** — it is never re-raised by later saves (unlike the absolute `PlayerCoreUpdatedEvent`, which self-heals next tick). Postgres never gets the row; the cache blob has it. The divergence surfaces as the item/skill silently vanishing whenever the sliding cache key lapses/evicts and the player rehydrates from the DB.

Note: `PlayerWriteBehindTests.SavePlayer_PublishFails_PreservesTheEventForTheNextFlush` pins the current semantics, but it re-saves through the *same scope's* repo/batch — the case production never has after a timeout. `PlayerProgressRepository.Save` has the same carve-out (benign for stats, whose deltas recompute from the un-advanced cache, but the same hole applies to the offline-rewards save #1819 closed for non-cancellation faults).

## Suggested fix

Treat a cancellation that unwinds a command *after a save began* like a flush failure:

- Have the timeout/abandon path mark `PlayerNeedsReload` when the abandoned task settles with `OperationCanceledException` (marking on teardown too is harmless — the in-memory player dies with the connection and the blob never advanced), **or** track "batch has unflushed envelopes at scope disposal" and mark the reload from there.
- Update the pinning test to assert the reload marker rather than the unwrapped OCE.
- `SocketHandler.cs:229`'s bare `catch (OperationCanceledException)` also classifies a leaked dependency `TaskCanceledException` as teardown with no reload — worth folding into the same fix.

ginderjeremiah:
## Summary

`PlayerRepository.SavePlayer` deliberately lets a cancellation propagate unwrapped (not wrapped in `PlayerPersistenceFlushFailedException`) since a cooperative cancellation isn't itself a persistence fault. But the `#1632`/`#1819` "flush failed → force in-memory `Player` reload" safety net only fires off that exception type — so a per-command timeout (or any other cancellation) that unwinds *after* a save already mutated the aggregate never marked the connection for reload, leaving a mutation stuck in memory with no queued write-behind event to match it. A later command's unrelated save would then persist the full cache blob **including** the stranded mutation, permanently diverging Redis (the rehydration source of truth) from what was ever durably queued to Postgres — concretely, a one-shot challenge-reward unlock event (`ItemUnlockedEvent`/`SkillUnlockedEvent`/etc.) silently lost whenever the cache key later lapses/evicts and the player rehydrates from the DB.

## Root cause

An unhandled `OperationCanceledException` settles a `Task` as **Canceled**, not **Faulted** — regardless of which token it carries — so it never reaches the `PlayerPersistenceFlushFailedException`-only checks in `SocketHandler`:

- `ReleaseCommandLockWhenSettled`'s abandoned-timeout continuation only inspected `task.Exception` (populated only for `Faulted` tasks) and explicitly treated `OperationCanceledException` as "the expected" no-op outcome.
- `RunCommandUnderLock`'s outer `catch (OperationCanceledException ex)` ("teardown" path) handles a cancellation that leaks from a source *other* than the per-command budget (e.g. an internal dependency token) — it logged and returned `TornDown` with no reload marking either.

## Fix

Classify by the settled task's **status**, not by exception type:

- `ReleaseCommandLockWhenSettled` now marks `PlayerNeedsReload` whenever the abandoned task settles as `Canceled` (in addition to the existing `PlayerPersistenceFlushFailedException` check for genuine faults).
- `RunCommandUnderLock`'s teardown catch now marks `PlayerNeedsReload` too.

Marking is safe even when a cancellation happened before any mutation ever occurred — the reload just re-reads the already-current persisted state, so the added cost is one harmless extra `GetPlayer` before the next command runs. `PlayerRepository.SavePlayer` itself is untouched: its cancellation-propagates-unwrapped behavior (and the pinning test that asserts it) stays correct — the gap was purely in how `SocketHandler` classified the settled task, not in the repository's exception semantics. Updated the stale comment in `PlayerRepository.cs` that had implied cancellation never triggers a reload.

Documented the corrected contract in `docs/backend-persistence.md`.

## Tests

`Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs`:
- `CooperativelyCancellingCommand_MarksPlayerForReload` — a command that honors the per-command budget's token settles the abandoned task as `Canceled`; asserts the release continuation marks the session for reload.
- `CommandThatThrowsCancellationUnrelatedToTheBudget_IsTornDown_AndMarksPlayerForReload` — a cancellation unrelated to the per-command budget (simulating a leaked dependency token) is classified `TornDown` and still marks the session for reload.

## Test plan

- [x] `dotnet test` — full solution suite passes: 1347 + 62 + 912 + 743 = 3064/3064, including the two new cases above.

Closes #1849

---
_Generated by [Claude Code](https://claude.ai/code/session_01CttLvqwCGoA7EF4oWsTb7q)_


ginderjeremiah (commented):
## Review: force the in-memory `Player` reload when a command settles via cancellation mid-save (#1849)

**Verdict:** Approve — the fix correctly closes the stranded-mutation gap by classifying on the settled task's *status* rather than on exception type, and it's more complete than the issue's literal suggestion. _(Posted as a `COMMENT` because GitHub won't accept a self-approval on my own PR; treat this as an approve.)_

### Why this is right

The root cause is diagnosed accurately: an unhandled `OperationCanceledException` from an async method settles its `Task` as **Canceled**, not **Faulted**, *regardless of the token it carries* — so the old `PlayerPersistenceFlushFailedException`-only checks never fired for a cancellation that unwound a command mid-`SavePlayer`. All three settlement paths are now covered:

- **Abandoned/timed-out task** — `ReleaseCommandLockWhenSettled`'s new `if (task.IsCanceled)` branch marks reload. Correct: a Canceled task has `task.Exception == null`, so it would otherwise slip past the `else if (task.Exception is …)` fault branch entirely.
- **Synchronous leaked (non-budget) cancellation** — the outer teardown `catch (OperationCanceledException)` now marks reload (folding in the `SocketHandler.cs:229` gap the issue called out).
- **Synchronous flush fault** — unchanged.

### More complete than the issue's suggestion

By classifying at the socket layer on task status instead of patching each repository's `when (ex is not OperationCanceledException)` filter, this also covers the `PlayerProgressRepository.Save` carve-out the issue flagged — any OCE during a progress save within a command unwinds the same command task and is now caught. The repository's deliberate cancellation-propagates-unwrapped semantics stay intact (and its pinning test with them), so the fix is scoped exactly to the socket-layer classification where the gap actually lived. Good call.

### Secondary cleanup is sound

Dropping the `fault.InnerExceptions.Any(e => e is not OperationCanceledException)` log guard is safe: a single async method's task faults with exactly one exception, and a pure-OCE outcome now settles as Canceled (handled in the branch above), so any fault reaching the `else if` is genuine and worth logging unconditionally. No dead code left behind.

### Test coverage

Adequate and well-matched to the new paths:
- `CooperativelyCancellingCommand_MarksPlayerForReload` — exercises the abandoned-task continuation's `IsCanceled` branch (cooperative budget cancellation).
- `CommandThatThrowsCancellationUnrelatedToTheBudget_IsTornDown_AndMarksPlayerForReload` — exercises the synchronous teardown catch (a leaked non-budget cancellation), and asserts no client response is sent and only one scope is created.

Both build on the existing `GatedCommand`/`ScopesCreated` harness with no helper changes. Backend-only change; no frontend/battle-parity surface, so no mirrored FE tests needed.

### What's good

- The comments carry exactly the non-obvious async-cancellation semantics a future edit could naively "simplify" and reintroduce the bug — appropriately detailed, not excessive.
- `docs/backend-persistence.md` documents the corrected contract (classify-by-status, not by exception type) accurately, and the stale `PlayerRepository.cs` comment was fixed rather than left to rot.

No blocking findings, no should-fixes, no nits. I'll merge once CI is green.

---

PR #1871, opened 2026-07-11
Reject negative spawn weights in AdminZones.SetEnemies/AdminEnemies.SetSpawns

### LINKED ISSUE #1850: Admin spawn-table saves accept negative weights, which poison the enemy snapshot build (failed reloads fleet-wide, boot failure, Workbench can't self-repair)
## Problem

`ZoneEnemy.Weight`/`EnemySpawn.Weight` are plain `int`s and neither spawn setter validates them — `AdminZones.SetEnemies` (`AdminZones.cs:154-176`) and `AdminEnemies.SetSpawns` both stage the weight verbatim (the only weight guard in the tier is `AdminSkills`' damage-portion check). But the enemy snapshot build constructs a `ProbabilityTable` from them (`EnemiesCacheHolder.cs:88`), and `ProbabilityTable`'s constructor **throws `ArgumentException` on any negative weight**.

So a save carrying `Weight = -1` **commits** (no validation, no DB check constraint), and then the awaited post-commit reload throws.

## Blast radius

- The admin gets a 500 but the row persisted.
- Every `EnemiesCacheHolder.ReloadAsync` on every instance now fails — readers keep the stale snapshot; the background sweep retries into permanent `LogError`.
- Any process **restart fails boot** (the eager startup load throws).
- The Workbench **cannot fix itself**: the corrective `SetEnemies` diffs against the stale snapshot's `ZoneEnemies` (which lacks the bad row), so the "fix" stages as an *insert* that collides with the committed row's `(ZoneId, EnemyId)` PK — another 500. Recovery requires direct DB surgery.

This is exactly the "authoring live content into a permanently broken state" class the retirement guards exist to reject at save time, and it defeats the "failed reload leaves the old snapshot in place" resilience by making the failure permanent. (Zero weights are safe — the table treats all-zero as uniform.)

## Suggested fix

Reject `Weight < 1` (or `< 0` if zero-weight authoring is wanted) up front in both `AdminZones.SetEnemies` and `AdminEnemies.SetSpawns` via `AdminSaveResult.Failure`, mirroring the skill-portion guard. Add the rejection to both admin integration test suites (no test currently covers a non-positive weight).

ginderjeremiah:
## Summary

`ZoneEnemy.Weight`/`EnemySpawn.Weight` are plain `int`s and neither `AdminZones.SetEnemies` nor `AdminEnemies.SetSpawns` validated them before staging. The enemy snapshot build (`EnemiesCacheHolder.BuildZoneEnemyTables`) constructs a `ProbabilityTable<int>` straight from these weights, and `ProbabilityTable`'s constructor **throws `ArgumentException` on any negative weight**.

So a save carrying `Weight = -1` committed cleanly, and the awaited post-commit reload then threw — poisoning every instance's reload (readers stuck on the stale snapshot, background sweep permanently `LogError`s), failing boot on any restart (the eager startup load throws), and leaving the Workbench unable to self-repair (the corrective `SetEnemies` diffs against the stale snapshot, so its "fix" stages as an insert that collides with the already-committed row's PK).

## Changes

- `AdminZones.SetEnemies` and `AdminEnemies.SetSpawns` now reject any `Weight < 0` up front via `AdminSaveResult.Failure`, mirroring the existing `AdminSkills.SetPortions` anti-tamper guard (non-positive weight check). Zero stays allowed — `ProbabilityTable` already treats an all-zero weight set as uniform, so it's not a landmine.
- New integration tests in `AdminZonesIntegrationTests` and `AdminEnemiesIntegrationTests` cover the negative-weight rejection for both setters.

## Test plan

- [x] `dotnet test` — full solution suite passes: 3066/3066 (`Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Application.Tests`, `Game.Api.Tests`), including the two new negative-weight rejection tests.

Closes #1850

---
_Generated by [Claude Code](https://claude.ai/code/session_0184soDDPBrctZGurZPht9AL)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminEnemiesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminEnemiesIntegrationTests.cs
@@ -468,6 +468,37 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SetSpawns_NegativeWeight_ReturnsFailure()
+        {
+            int enemyId, zoneId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var enemy = await TestDataSeeder.CreateEnemyAsync(context);
+                var zone = await TestDataSeeder.CreateZoneAsync(context);
+                enemyId = enemy.Id;
+                zoneId = zone.Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            // A negative weight would otherwise commit and then throw inside ProbabilityTable's constructor
+            // when the enemy snapshot next rebuilds — reject it up front instead.
+            var data = new SetEnemySpawnsData
+            {
+                EnemyId = enemyId,
+                Spawns = [new Contracts.EnemySpawn { ZoneId = zoneId, Weight = -1 }],
+            };
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminEnemies>();
+
+            var result = admin.SetSpawns(data);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal("A spawn's weight cannot be negative.", result.ErrorMessage);
+        }
+
+        [Fact]
         public async Task SaveEnemies_RetiringLastActiveEnemyOfLiveZone_ReturnsFailure()
         {
             int enemyId;
--- a/Game.Application.Tests/DataAccess/AdminZonesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminZonesIntegrationTests.cs
@@ -115,6 +115,37 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SetEnemies_NegativeWeight_ReturnsFailure()
+        {
+            int zoneId, enemyId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var zone = await TestDataSeeder.CreateZoneAsync(context);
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Reject negative spawn weights in `AdminZones.SetEnemies` / `AdminEnemies.SetSpawns` (fixes #1850)

**Verdict:** Approve — correctly closes the anti-tamper gap the issue describes; posted as a `COMMENT` only because GitHub won't accept a gating verdict on my own PR.

Verified the fix against the code rather than the description:

- `ProbabilityTable`'s constructor does throw `ArgumentException` on `Weight < 0` (`Game.Core/Probability/ProbabilityTable.cs:41-44`), and an all-zero set falls back to uniform (`:46-47`) — so rejecting `< 0` while keeping `0` allowed is exactly right, and matches the issue's "or `< 0` if zero-weight authoring is wanted."
- The guard mirrors the existing `AdminSkills.SetPortions` weight check (`AdminSkills.cs:101`), deliberately relaxing `<= 0` to `< 0`. The comment explaining *why* (poisoned reloads, boot failure, Workbench can't self-repair) is the kind of non-obvious rationale worth keeping.
- Both setters reconcile against a full desired-state list (no delete markers), so `.Any(... Weight < 0)` over the whole set is the correct shape — no need to exclude deletes the way `SetPortions` does.

### Test coverage

- `AdminZonesIntegrationTests.SetEnemies_NegativeWeight_ReturnsFailure` and `AdminEnemiesIntegrationTests.SetSpawns_NegativeWeight_ReturnsFailure` cover both directions, asserting rejection + message — this closes the "no test covers a non-positive weight" gap the issue called out. Integration-level coverage is the right tier for these data-access classes per `backend.md`.

### What's good

- Tightly scoped fix that mirrors an established pattern, with clear rationale comments and symmetric tests on both authoring directions. Nothing to change.


---

PR #1872, opened 2026-07-11
Apply the post-battle cooldown to abandon-resolved battle outcomes

### LINKED ISSUE #1851: Abandon-resolved outcomes skip the post-battle cooldown — a tampered client can farm ~20-35% faster by never sending DefeatEnemy
## Problem

Both explicit battle-end paths set the 5s pacing cooldown — `EndBattleVictory` (`BattleService.cs:312`, `state.SetCooldown(battleCompletedAt + PostBattleCooldown)`) and `EndBattleLoss` (`BattleService.cs:360`) — but `AbandonBattle` (`BattleService.cs:420-485`) records a full outcome (win via `RecordVictory`, loss/draw via `RecordBattleCompleted`) without ever calling `SetCooldown` on any branch.

`NewEnemy`'s cooldown gate (`NewEnemy.cs:25`, `state.IsOnCooldown(now)`) runs *before* `StartBattle`, and the replacement battle is anchored to `now`.

## Failure scenario

A tampered client never sends `DefeatEnemy`. Instead it loops `NewEnemy { clientBattleMs: <elapsed> }`: each call re-simulates the active battle (server-clock-capped, so the win itself can't be claimed early), credits the **full victory** (exp, statistics, challenges, proficiency XP) through `RecordVictory`, and immediately starts the next battle with zero cooldown. The 5-second `PostBattleCooldown` is shaved off every kill — for typical short battles a ~20-35% idle-farm-rate amplification, permanently. The offline simulator deliberately paces with the same cooldown between simulated battles, so the live bypass is also inconsistent with the offline pacing model.

## Suggested fix

When `AbandonBattle` records an outcome (win, loss, or draw), set `state.SetCooldown(now + PostBattleCooldown)` before clearing, and have `StartBattle` anchor the replacement battle's `BattleStartTime` to the cooldown expiry (the same `scheduledStartTime` mechanism `PrepareNextIdleBattle` already uses) rather than `now`. The no-outcome branches (`elapsedMs <= 0`, the still-in-progress hand-back) must stay cooldown-free so the legitimate zone-switch/prefetch-supersede flow is unaffected.

ginderjeremiah:
## Summary

`BattleService.AbandonBattle` credits a resolved outcome (win, loss, or draw) exactly like the explicit `EndBattleVictory`/`EndBattleLoss` paths — but, unlike those two, never applied the shared `PostBattleCooldown`. A tampered client that never sends `DefeatEnemy` and instead only ever calls `NewEnemy` (which abandons the active battle server-side) got every kill's outcome credited in full while skipping the mandatory 5-second pacing cooldown between battles — a ~20-35% idle-farm-rate amplification for typical short battles.

## Changes

- `BattleService.AbandonBattle`: both outcome-bearing branches (won-abandon and lost/drawn-abandon — the still-in-progress hand-back is unaffected) now call `state.SetCooldown(now + PostBattleCooldown)` before clearing the battle, mirroring `EndBattleVictory`/`EndBattleLoss`.
- `BattleService.StartBattle`: when its own internal abandon call resolves an outcome (and thus just applied the cooldown above), the replacement battle it spawns in the same call is now anchored to that cooldown's expiry instead of `now` — otherwise the freshly-applied cooldown would be immediately bypassed by the very next battle starting right away, since `NewEnemy`'s cooldown gate only runs *before* `StartBattle` is called, not after an abandon inside it.
- `docs/backend-battle.md`: documents that an abandon-resolved outcome now applies the same cooldown as the explicit end paths, and that `StartBattle` anchors the replacement battle to it.

## Tests

Two new integration tests in `BattleServiceIntegrationTests.cs`:
- `StartBattle_AbandoningAWonBattle_SetsPostBattleCooldownAndAnchorsReplacementBattle` — a won-abandon sets the cooldown and the replacement battle's `BattleStartTime` is anchored to it.
- `StartBattle_AbandoningADrawnBattle_SetsPostBattleCooldown` — same coverage for a drawn (stalemate-timeout) abandon, so statistic/challenge farming via a looped drawn-abandon is closed too.

Existing abandon/cooldown/still-in-progress tests (`EndBattleVictory_CooldownAnchoredToBattleCompletion_NotServerClock`, `PrepareNextIdleBattle_AnchorsBattleStartToScheduledCooldownExpiry`, `StartBattle_AbandoningAStillInProgressBattle_HandsItBackWithoutRecordingAnOutcome`, etc.) still pass unchanged.

## Test plan

- [x] `dotnet test` — full solution suite passes: 3068/3068 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), including the two new regression tests above.
- [ ] Frontend unaffected (backend-only anti-cheat/pacing change) — no frontend verification needed.

Closes #1851

---
_Generated by [Claude Code](https://claude.ai/code/session_01HVchbGmSadtJ45CPvYtxoQ)_


ginderjeremiah (commented):
## Review: post-battle cooldown for abandon-resolved outcomes (#1851)

**Verdict:** Approve — the change closes the exact farming vector in #1851, mirrors the existing pacing/anchoring patterns, and is correctly scoped so the no-outcome flows stay untouched. (Posted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author; treat it as an approve.)

### Correctness

- The cooldown at `BattleService.cs:502` is reached by exactly the won branch and the lost/drew `else` branch. The still-in-progress hand-back returns early, and the `elapsedMs <= 0` / no-outcome branch clears-and-returns before it — so the legitimate zone-switch and prefetch-superseded (`clientBattleMs: 0`) flows correctly stay cooldown-free, as #1851 required.
- The `abandonedOutcomeCooldown` capture (`:82`) rests on a load-bearing invariant — only `NewEnemy` reaches `StartBattle` with an active battle, and it gates on `IsOnCooldown` first, so a future `EnemyCooldown` here can only be the one `AbandonBattle` just set. I confirmed `NewEnemy.cs:25` returns early on cooldown and that it's the sole non-test caller with an active battle (`PrepareNextIdleBattle` runs after `ClearBattle`). The invariant is documented inline, so a later caller that skips the cooldown gate would need to revisit this — the comment flags that adequately.
- The real enforcement is the future-anchored replacement `BattleStartTime`: even an immediate re-claim is rejected by `ClaimedBeforeBattleCouldFinish` until real time reaches the anchor + duration. This is server-authoritative, so no frontend change is needed — the "frontend unaffected" note is right (the frontend never replays abandon crediting).
- Anchoring the won-abandon cooldown to `now` (matching `EndBattleLoss`) rather than `EndBattleVictory`'s `battleCompletedAt` differs only by the abandon's network latency and errs toward a marginally longer cooldown — the correct, anti-farm direction. Fine.

### Test coverage

Adequate. The won test exercises the `if result.Victory` branch and the drawn test the `else` branch, so both structural paths that reach the new `SetCooldown` (`:502`) are covered, plus the replacement-battle anchoring assertion (`EnemyCooldown == BattleStartTime`, no `ElapsedOffsetMs`). The loss-abandon shares the `else` branch with the draw case and adds no new branch, and the existing still-in-progress / zero-elapsed tests keep the cooldown-free paths pinned.

### What's good

Tight, minimal fix that reuses the exact `scheduledStartTime` anchoring mechanism `PrepareNextIdleBattle` already established, keeps the two win paths from drifting, and documents the new pacing in `backend-battle.md`. The inline anti-cheat reasoning is genuinely useful context, not noise.

---

PR #1873, opened 2026-07-11
Clamp AbandonBattle's wall-clock elapsed-time cast to avoid int overflow

### LINKED ISSUE #1852: AbandonBattle's wall-clock int cast overflows for a stale battle older than ~24.9 days
## Problem

`BattleService.AbandonBattle` (`BattleService.cs:423`):

```csharp
var wallClockMs = (int)(now - state.BattleStartTime).TotalMilliseconds;
```

`TotalMilliseconds` exceeds `int.MaxValue` after ~24.9 days. The unchecked `double→int` conversion of an out-of-range value is undefined (in practice it saturates to `int.MinValue`), so `wallClockMs` becomes garbage, making `elapsedMs <= 0` clear the battle with **no outcome** — or feeding a bogus value into the still-in-progress classification at line 457.

## Failure scenario

A player disconnects mid-battle and returns after a month; `ResolveStaleBattle` (login/offline path, `clientBattleMs: null`) hits the overflow and the stale battle silently resolves as nothing instead of the draw/win it should be. The away-window simulation itself is unaffected. No crash, small progress loss — but the arithmetic is simply wrong past a threshold a real account can cross.

## Suggested fix

Compute in `double` and clamp before narrowing:

```csharp
var wallClockMs = (int)Math.Clamp((now - state.BattleStartTime).TotalMilliseconds, 0, int.MaxValue);
```

The very next line already caps the simulation window at `DefaultMaxBattleMs`, so clamping loses nothing.

ginderjeremiah:
## Summary

`BattleService.AbandonBattle` (`Game.Application/Services/BattleService.cs:440`) computed:

```csharp
var wallClockMs = (int)(now - state.BattleStartTime).TotalMilliseconds;
```

`TotalMilliseconds` exceeds `int.MaxValue` for a battle whose `BattleStartTime` is more than ~24.9 days in the past (e.g. a stale in-flight battle from a long-disconnected player, resolved via `ResolveStaleBattle`/offline login). Narrowing an out-of-range `double` to `int` in an unchecked context is unspecified behavior per the C# language spec — relying on it is fragile even though it isn't crashing today.

## What I verified vs. the issue's claim

The issue's suggested fix section asserted the cast "in practice ... saturates to `int.MinValue`", which would make `elapsedMs <= 0` clear the battle with no outcome. I checked this empirically on this repo's actual .NET 10 runtime: an out-of-range positive `double → int` cast currently saturates to `int.MaxValue`, not `int.MinValue`. Tracing that through the method, `int.MaxValue` still clears every downstream comparison (`Math.Min(elapsedMs, DefaultMaxBattleMs)`, `wallClockMs < DefaultMaxBattleMs`) the same way a correctly-computed huge value would — so the specific "silently clears with no outcome" symptom does not currently reproduce on this runtime.

That said, the underlying issue — relying on unspecified narrowing-conversion behavior instead of an explicit domain clamp — is real and worth fixing regardless of the current runtime's (undocumented-by-spec) saturation behavior. The fix is a one-line hardening with no behavior change today, and removes the fragility for other targets/runtimes.

## Changes

- `BattleService.AbandonBattle`: clamp the elapsed-`TotalMilliseconds` to `[0, int.MaxValue]` before narrowing to `int`, instead of an unchecked cast.
- New regression test `ResolveStaleBattle_BattleStartedOverIntOverflowThresholdAgo_StillCreditsTheWin`, mirroring the existing `ResolveStaleBattle_Concludes_ReturnsNullAndCreditsTheWin` but with `BattleStartTime` 30 days in the past (past the overflow threshold), pinning that the abandon still resolves and credits a definite win.

No documentation change — `docs/backend-battle.md`'s described contract (abandon resolves a definite outcome, capped at `DefaultMaxBattleMs`) already holds; this only hardens the internal computation, it doesn't change behavior.

## Test plan

- [x] `dotnet test` (full backend suite, no build/restore) — 3,069/3,069 passing (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests including the new test, Game.Api.Tests).

Closes #1852

---
_Generated by [Claude Code](https://claude.ai/code/session_01X66LrVVJ2oKVkZVT9dVWWB)_

### DIFF
--- a/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
@@ -1479,6 +1479,46 @@ namespace Game.Application.Tests.Services
         }
 
         [Fact]
+        public async Task ResolveStaleBattle_BattleStartedOverIntOverflowThresholdAgo_StillCreditsTheWin()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+
+            // Same one-shot setup as ResolveStaleBattle_Concludes_ReturnsNullAndCreditsTheWin, but the battle's
+            // wall-clock age (30 days) exceeds int.MaxValue ms (~24.9 days) — pinning that the abandon still
+            // resolves a definite outcome instead of the elapsed-time computation overflowing into garbage.
+            var playerSkill = await TestDataSeeder.CreateSkillAsync(context, "Smash", baseDamage: 1000m, cooldownMs: 500);
+            var enemy = await TestDataSeeder.CreateEnemyAsync(context);
+            var enemySkill = await TestDataSeeder.CreateSkillAsync(context, "Poke", baseDamage: 5000m, cooldownMs: 2000);
+            await TestDataSeeder.LinkSkillToEnemyAsync(context, enemy.Id, enemySkill.Id);
+            var zone = await TestDataSeeder.CreateZoneAsync(context, levelMin: 10, levelMax: 10);
+            await TestDataSeeder.LinkEnemyToZoneAsync(context, zone.Id, enemy.Id);
+
+            var user = await TestDataSeeder.CreateUserAsync(context);
+            var playerEntity = await TestDataSeeder.CreatePlayerAsync(context, user.Id, zoneId: zone.Id);
+            await TestDataSeeder.LinkSkillToPlayerAsync(context, playerEntity.Id, playerSkill.Id);
+
+            await ReloadReferenceCachesAsync();
+
+            var playerRepo = scope.ServiceProvider.GetRequiredService<IPlayerRepository>();
+            var player = await playerRepo.GetPlayer(playerEntity.Id);
+            Assert.NotNull(player);
+
+            var battleService = scope.ServiceProvider.GetRequiredService<BattleService>();
+            var state = new PlayerState();
+            var expBefore = player.Exp;
+
+            await battleService.StartBattle(player, state, zoneId: zone.Id);
+            state.BattleStartTime = DateTime.UtcNow.AddDays(-30);
+
+            var handoff = await battleService.ResolveStaleBattle(player, state);
+
+            Assert.Null(handoff);
+            Assert.False(state.HasActiveBattle);
+            Assert.True(player.Exp > expBefore);
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: harden `AbandonBattle`'s wall-clock elapsed-ms cast against int overflow (#1852)

**Verdict:** Approve — correct, minimal, well-tested fix; no blocking or should-fix findings. (Posted as a comment because GitHub won't accept a gating verdict from the PR author; the intended verdict is **Approve**.)

The change replaces an unchecked `(int)double` narrowing with `(int)Math.Clamp(…, 0, int.MaxValue)`. I traced `wallClockMs` through both of its consumers to confirm no behavior change today:

- `elapsedMs` (line 450) — subsequently capped at `DefaultMaxBattleMs`, so clamping the far-end to `int.MaxValue` loses nothing.
- The `wallClockMs < DefaultMaxBattleMs` "still in progress" gate (line 476) — a 30-day-old battle now yields `int.MaxValue`, which is correctly far above the cap, so it resolves as a genuine draw/win rather than being misclassified as in-progress. `ElapsedOffsetMs = wallClockMs` (line 491) is only reached on that branch, so it can never carry the overflow value.

I appreciate that the description empirically verified the issue's claim rather than taking it at face value — the issue asserted saturation to `int.MinValue` (which would have silently cleared the battle), but the actual .NET 10 runtime saturates to `int.MaxValue`, so the "resolves as nothing" symptom doesn't reproduce today. The fix is still worth making: relying on unspecified narrowing-conversion behavior is fragile regardless of the current runtime's undocumented saturation direction. Good, honest scoping.

### Test coverage

`ResolveStaleBattle_BattleStartedOverIntOverflowThresholdAgo_StillCreditsTheWin` mirrors the existing `ResolveStaleBattle_Concludes_ReturnsNullAndCreditsTheWin` with `BattleStartTime` pushed 30 days back (past the ~24.9-day `int.MaxValue`-ms threshold) and pins that the abandon still credits a definite win. That's the right regression to lock in, and it fails against the old unchecked cast on any runtime that saturates toward `int.MinValue`. The setup duplication with the sibling test matches the established pattern in this file — fine.

### What's good

- Tight, single-line hardening with an accurate explanatory comment.
- Correct call not to touch `docs/backend-battle.md` — the documented contract (abandon resolves a definite outcome capped at `DefaultMaxBattleMs`) is unchanged.

Merging once CI is green.

---

Issue #1874, opened 2026-07-11
Idempotency backstop for battle-lifecycle session saves (residual reconnect double-credit gap after #1853)


ginderjeremiah:
## Context

#1853 made the battle-lifecycle session save (`SessionService.SavePlayerStateAsync`, called by `DefeatEnemy`, `BattleLost`, `NewEnemy`, `ChallengeBoss`, `GetOfflineProgress`) an **awaited** Redis write instead of fire-and-forget (`ICacheService.SetAndForget`, which uses `CommandFlags.FireAndForget` and offers no way to observe a failure at all). That closes the common case: a normal Redis error now surfaces as a thrown exception on the command instead of a silently dropped write.

## Residual gap

A true crash (process kill, OOM, host failure) between the durable write-behind credit and the now-awaited session save can still strand the session showing the already-credited battle as active. On reconnect, the next battle-end (`StartBattle`'s abandon path) would re-abandon — and thus re-credit — that same battle. This window is now much narrower (bounded by one in-flight Redis round-trip instead of an entire unbounded fire-and-forget race), but it isn't eliminated.

## Suggested follow-up

#1853's issue body floated a cheap backstop: record the resolved battle's seed (or another stable identifier) on the `Player` aggregate when a victory/loss is durably credited, and have the abandon path refuse to re-credit a battle whose seed was already recorded. This would make the reconnect path idempotent regardless of the session cache's state, closing the gap structurally instead of narrowing it.

This is a distinct design decision (what identifier to key on, where it lives on the aggregate, how long to retain it) from the awaited-write fix, so it's split out rather than bundled into #1853's PR.

PR #1875, opened 2026-07-11
Await the battle-lifecycle session save instead of fire-and-forget

### LINKED ISSUE #1853: Battle-lifecycle session save is fire-and-forget — a lost "battle cleared" write reopens the reconnect double-credit path
## Problem

`SessionService.SavePlayerState()` → `SessionStore.Update` (`SessionStore.cs:48-51`) → `_cache.SetAndForget(...)`. The battle-end commands (`DefeatEnemy.cs:62`, `BattleLost.cs:47`, `NewEnemy.cs:37`) durably credit the outcome through the write-behind player save (queued, retried, dead-lettered), then persist the **cleared battle state** to the session cache with a fire-and-forget write that has no failure handling — no log, no retry, no reconciliation.

The codebase elsewhere treats "a reconnect re-abandons and re-credits an already-credited battle" as a defect worth structural defenses (`BattleService.TryPrepareNextIdleBattle`'s doc comment calls out exactly this). A dropped `SetAndForget` reopens that path: rewards persist, the session still shows the battle active, and the next connect resolves the stale (already-credited) battle as a fresh won-abandon → double exp/statistics/challenge credit for one battle.

## Relationship to #580

#580 (closed not-planned) accepted fire-and-forget for the *player aggregate* cache write because commands don't read the player back from the cache mid-session. This case is different: the session's `PlayerState` **is** read back — it is exactly what the reconnect/rehydration path consumes — so a lost write here has a concrete, player-visible consequence rather than a self-healing one.

## Suggested fix

Make the battle-lifecycle `SavePlayerState` an awaited write (the commands are async and already on the serialized per-socket path — one awaited Redis `SET` is cheap), or at minimum route failures through a logged callback so a lost session write is visible. A cheap idempotency backstop would be recording the resolved battle's seed on the player aggregate and refusing to re-credit the same seed twice.

ginderjeremiah:
## Summary

`SessionService.SavePlayerState()` → `ISessionStore.Update` → `ICacheService.SetAndForget` (`Game.DataAccess/Repositories/SessionStore.cs`) persisted the in-flight `PlayerState` (the session's active-battle snapshot) using Redis's `CommandFlags.FireAndForget` — a call that returns before the server even acknowledges it, with no way to observe a failure at all.

The battle-end commands (`DefeatEnemy.cs`, `BattleLost.cs`, `NewEnemy.cs`) durably credit a battle's outcome through the write-behind player save (queued, retried, dead-lettered), then persist the **cleared battle state** to the session cache with this fire-and-forget write. A dropped write there left the stale session still showing the already-credited battle as active — so on reconnect, `BattleService.TryPrepareNextIdleBattle`'s doc comment already calls out exactly this: `StartBattle`'s abandon path would re-abandon, and thus re-credit, the same battle. Rewards, statistics, and challenge progress would double up for one battle.

## Changes

- `ISessionStore`: added `UpdateAsync(PlayerState, int, CancellationToken)` — the awaited counterpart to the existing fire-and-forget `Update`.
- `SessionStore`: implements `UpdateAsync` via `ICacheService.Set` (awaited), alongside the existing `SetAndForget`-backed `Update` (still used by `CreateSession` at login, which isn't on the reconnect double-credit path).
- `SessionService.SavePlayerState()` → `SavePlayerStateAsync(CancellationToken)`, backed by `UpdateAsync`.
- All five callers of the old `SavePlayerState()` — `DefeatEnemy`, `BattleLost`, `NewEnemy`, `ChallengeBoss`, `GetOfflineProgress` — now `await` the async version (they're already async, on the serialized per-socket path, so one more awaited Redis round-trip doesn't change the concurrency shape).
- `docs/backend-persistence.md`: documents why this specific save is awaited while the player-aggregate cache write next to it deliberately stays fire-and-forget (the aggregate is never re-read mid-session; this session key is exactly what a reconnect reads back).

## Scope note

The issue also floated a "cheap idempotency backstop" — recording the resolved battle's seed on the player aggregate and refusing to re-credit a repeat. I left that out of this PR: it's a distinct design decision (what identifier to key on, where it lives, retention) and the awaited write already closes the common-case failure mode (any ordinary Redis error now surfaces as a thrown exception instead of vanishing). Filed as #1874 to track the residual gap (a true crash between the durable credit and the now-awaited session write).

## Test plan

- [x] `dotnet test --no-build --no-restore` — full solution suite passes: 3071/3071 (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), including two new `SessionServiceTests` cases pinning that `SavePlayerStateAsync` goes through the awaited store write (not the fire-and-forget one) and forwards its cancellation token.

Closes #1853

---
_Generated by [Claude Code](https://claude.ai/code/session_01DJ5ZLDMBwVVGbqNs64HKoj)_


ginderjeremiah (commented):
## Review: Await the battle-lifecycle session save (#1853)

**Verdict:** Approve — the fire-and-forget session write that reopened the reconnect double-credit path is now an awaited `SET`, correctly scoped to the exact save the reconnect path reads back. Clean, well-tested, and no findings. (Submitted as a comment because the review account is the PR author; treat this as an approval.)

The change matches the issue precisely: `SessionService.SavePlayerStateAsync` goes through the new `ISessionStore.UpdateAsync` → `ICacheService.Set` (awaited) instead of `SetAndForget`, so a dropped session write after a durably-credited battle now surfaces as a thrown exception rather than vanishing. All five callers (`DefeatEnemy`, `BattleLost`, `NewEnemy`, `ChallengeBoss`, `GetOfflineProgress`) were already async on the serialized per-socket path, so awaiting one Redis round-trip doesn't change the concurrency shape. The old sync `SavePlayerState()` is fully removed with no stragglers.

### Verified

- `Update` (fire-and-forget) is correctly retained only for `CreateSession` login priming — that path isn't the reconnect double-credit vector, and its doc comment on `UpdateAsync` makes the "which one to use" distinction explicit, so keeping both is a documented choice rather than a footgun.
- The `docs/backend-persistence.md` addition accurately explains why this save is awaited while the sibling player-aggregate write stays fire-and-forget (the session key is what a reconnect reads back; the aggregate never is).
- No frontend mirror needed — this is a pure persistence/infra change with no battle-calculation logic.

### Test coverage

Solid. The two new `SessionServiceTests` cases pin the important behaviors — that the save routes through the awaited store write (not the fire-and-forget one) and forwards its cancellation token — which is exactly the regression that matters here. The `NoOpSessionStore` test doubles across the middleware/socket suites were all updated for the new interface member.

### Follow-up

The residual crash-window gap (a process kill between the durable credit and the now-awaited session save) is explicitly out of scope and already tracked as #1874 (idempotency backstop keyed on the resolved battle's seed). Correctly split out — it's a distinct design decision, not something to bundle here.

### What's good

Tight, surgical fix that stays exactly on the issue's scope, with the awaited-vs-fire-and-forget rationale documented both in code and in the persistence doc so the next reader doesn't "simplify" it back to `SetAndForget`.


---

PR #1877, opened 2026-07-11
Abort Workbench child savers on an unresolved add id instead of falling back to it

### LINKED ISSUE #1855: Workbench child savers can POST against an unresolved negative id after the identity Add committed
## Problem

In `persistEntity` (`UI/src/routes/admin/workbench/save-helpers.ts:145`, same pattern in the progression store's child-saver targets):

```ts
const childTargets = [
    ...diff.added.map((record) => ({ id: idFor.get(record.id) ?? record.id, record, baseline: undefined })),
    ...
];
```

`idFor` comes from `resolveNewIds(fresh, ...)`. If the post-save refetch returns fewer new records than were added (a stale read that beat the server-side cache reload, or any refetch anomaly), the `?? record.id` fallback silently posts the child collections (skill portions/effects, item mod slots, proficiency modifiers) against the **local negative id** — *after* the identity Add already committed.

## Failure scenario

Identity Add lands → refetch momentarily stale → `SetSkillPortions { id: -1 }` posted. Best case the backend rejects it and the save surfaces as a confusing partial failure; worst case orphaned child rows.

## Suggested fix

If an added record has no resolved id, throw (treat as `PersistFailedError`, which already forces a re-seed from server truth) before running its child savers, instead of falling back to the negative id. Land with a test asserting child savers abort on an unresolved add id.

ginderjeremiah:
## Summary

`persistEntity` (`UI/src/routes/admin/workbench/save-helpers.ts`) resolves each added record's real persisted id positionally, from a post-save refetch (`resolveNewIds`). The lookup at the child-saver call site fell back to the local (negative) id whenever that lookup came up empty:

```ts
{ id: idFor.get(record.id) ?? record.id, record, baseline: undefined }
```

If the refetch returned fewer new records than were added (a stale read racing the server-side cache reload, or any other refetch anomaly), this silently posted the child-collection saves (skill portions/effects, item mod slots, proficiency modifiers) against the **local negative id** — after the identity Add had already committed.

## Changes

- `resolveId` (previously `idMap.get(id) ?? id`) now throws when given a local (negative) id that has no entry in the map. `0` still passes through unchanged, since this project's entity ids are 0-based — `0` is a legitimate persisted id, not a "new record" sentinel (only strictly-negative ids, from the workbench stores' `nextId` counters, mean "new/unsaved").
- `persistEntity`'s child-target resolution now calls `resolveId` instead of duplicating the fallback inline, so an unresolved add throws before any child saver runs. `persistEntity` already tracks whether anything committed, so this throw is automatically wrapped as `PersistFailedError` — the same path that already forces the caller to re-seed from server truth on any post-commit failure.
- `docs/frontend.md` is unaffected; no architectural decision here, just closing a silent-fallback gap in an existing documented contract (`PersistFailedError`'s re-seed-on-commit behavior).

## Bonus fix (same pattern, same file)

The issue's problem statement also flagged the progression store's proficiency/path save orchestration (`progression/progression-store.svelte.ts`) as having the same fallback pattern. It turned out those call sites already route through the shared `resolveId` helper, so hardening `resolveId` itself fixes both places with one change — no edits needed in `progression-store.svelte.ts`.

## Tests

- `save-helpers.test.ts`: new cases — `resolveId` throws on an unresolved negative id; `resolveId` passes `0` through unchanged (0-based ids); and a `persistEntity` case pinning that a stale refetch (fewer new records than added) aborts before any child saver runs, surfacing as `PersistFailedError`.

## Test plan

- [x] `npx vitest run` on the touched suites — all passing.
- [x] `npm run test:unit:ci` — full frontend suite passes (exit 0).

Closes #1855

---
_Generated by [Claude Code](https://claude.ai/code/session_019fpUWf9QKTkm9CdpDcmtyT)_

### DIFF
--- a/UI/src/routes/admin/workbench/save-helpers.ts
+++ b/UI/src/routes/admin/workbench/save-helpers.ts
@@ -58,8 +58,22 @@ export const resolveNewIds = (
 	return idFor;
 };
 
-/** Resolve a possibly-local id through the new-id map (returns the input when already persisted). */
-export const resolveId = (id: number, idMap: Map<number, number>): number => idMap.get(id) ?? id;
+/**
+ * Resolve a possibly-local id through the new-id map (passes through an already-persisted id
+ * unchanged — entity ids are 0-based, so `0` is a legitimate persisted id, not a sentinel). Throws
+ * if `id` is a local (negative) id absent from the map — the refetch that should have confirmed its
+ * add landed didn't, so the caller must not silently write against the stale local id.
+ */
+export const resolveId = (id: number, idMap: Map<number, number>): number => {
+	const resolved = idMap.get(id);
+	if (resolved !== undefined) {
+		return resolved;
+	}
+	if (id < 0) {
+		throw new Error(`No persisted id found for record (local id ${id})`);
+	}
+	return id;
+};
 
 /** Returns whether it actually wrote — a no-op (unchanged collection) must report `false`. */
 type ChildSaver<T> = (id: number, record: T, baseline: T | undefined) => Promise<boolean>;
@@ -142,7 +156,7 @@ export async function persistEntity<T extends Identified, D>(opts: PersistOption
 		const idFor = resolveNewIds(fresh, diff.existingIds, diff.added);
 
 		const childTargets: { id: number; record: T; baseline: T | undefined }[] = [
-			...diff.added.map((record) => ({ id: idFor.get(record.id) ?? record.id, record, baseline: undefined })),
+			...diff.added.map((record) => ({ id: resolveId(record.id, idFor), record, baseline: undefined })),
 			...diff.modified.map(({ record, baseline }) => ({ id: record.id, record, baseline }))
 		];
 
--- a/UI/src/tests/routes/admin/workbench/save-helpers.test.ts
+++ b/UI/src/tests/routes/admin/workbench/save-helpers.test.ts
@@ -156,6 +156,15 @@ describe('resolveNewIds & resolveId', () => {
 		expect(resolveId(-1, map)).toBe(7);
 		expect(resolveId(3, map)).toBe(3);
 	});
+
+	it('resolveId throws when a local (negative) id has no persisted match', () => {
+		const map = new Map([[-1, 7]]);
+		expect(() => resolveId(-2, map)).toThrow();
+	});
+
+	it('resolveId passes through 0 unchanged — entity ids are 0-based, so 0 is a real persisted id', () => {
+		expect(resolveId(0, new Map())).toBe(0);
+	});
 });
 
 interface Row extends Identified {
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: harden `resolveId` so Workbench child savers can't POST against an unresolved negative id (#1855)

**Verdict:** Approve — the fix closes the silent-fallback gap exactly at the contract boundary, correctly handles the 0-based-id subtlety, and is well tested. (Posted as a comment because GitHub won't accept a gating verdict from the PR author; this is a genuine approve.)

I traced the two things the correctness hinges on:

- **The `id < 0` sentinel is sound.** Both workbench `nextId` counters (`entity-store.svelte.ts:36`, `progression-store.svelte.ts:59`) start at `-1` and only decrement, so a new top-level entity is *always* strictly-negative — never `0`. That makes the `0`-passthrough and the negative-throw a clean partition, and it lines up with the project's 0-based reference-data id convention.
- **Hardening the shared `resolveId` doesn't regress the progression store.** It changes behavior for all call sites, so I checked each one. `selectedPathId`/`drilledTierId` (steps 9, `progression-store.svelte.ts:562-566`) are the only remap where a throw would be new — but a new path is always in `pathDiff.added`→`pathIdMap`, and a "tier" *is* a proficiency (`addTier` → `newProficiency` into `this.profs`), so a new tier is always in `profDiff.added`→`profIdMap`. The only way `resolveId` throws there is when a selection points to a record that genuinely didn't persist — precisely the anomaly this fix wants to surface, and it's inside the `committed`-tracked try so it re-seeds. No legitimate flow relied on the old silent `?? id` fallback.

The `.map()` that builds `childTargets` evaluates all `resolveId` calls eagerly, so an unresolved add throws *before* the child-saver loop — which is why the whole batch aborts rather than partially applying, matching the test's `childCalls === []` assertion. Reusing `resolveId` at the call site (instead of the inline `?? record.id`) is the right DRY move: one place now owns the contract.

### Test coverage

Sufficient and on-point:
- `resolveId` throws on an unresolved negative id; passes `0` through unchanged (pins the 0-based-id rule).
- `persistEntity`: a stale refetch returning fewer new records than added surfaces as `PersistFailedError` with **no** child saver run — verified the positional-resolution math (only `-1` resolves; `-2` is unresolved and throws). This is the exact failure scenario from #1855.

### What's good

Tightly scoped to the reported bug, the extended `resolveId` doc-comment earns its length by calling out the 0-vs-negative subtlety a future edit could easily get wrong, and the bonus finding (progression store already routes through `resolveId`, so no second edit needed) is correct — I confirmed it rather than taking the description's word for it.

No blocking items, no should-fixes, no nits. Approving; will merge once `test-e2e (webkit)` finishes (the other CI jobs are green; backend checks correctly skipped for a frontend-only change).

---

PR #1878, opened 2026-07-11
Flush partial-second DoT/HoT combat log on abandon-shaped battle ends

### LINKED ISSUE #1857: Sub-second DoT/HoT combat-log window is silently discarded on abandon-shaped battle ends (retreat / Home / zone swap)
## Problem

`BattleEngine` accumulates DoT/HoT combat-log lines per second and flushes the partial second only when a battle **concludes** (victory/defeat/draw — `flushEffectDamage()` at `battle-engine.ts:402/454`). The abandon-shaped ends never flush:

- `rest()` (`battle-engine.ts:181-185`) drops to Idle without flushing.
- `reset(enemyInstance)` calls `resetEffectDamage()` (`battle-engine.ts:204`), discarding whatever the current partial second accumulated.

## Failure scenario

A player retreats from a boss (or steps Home / swaps zones) mid-window — up to ~1s of DoT/HoT they actually took/dealt never reaches the combat log, contradicting the documented "any partial second flushed at battle end" contract (game-design.md → Logs) for these end paths. Cosmetic only — no parity or state impact.

## Suggested fix

Call `flushEffectDamage()` at the top of `rest()` and in `reset()` before `resetEffectDamage()`.

ginderjeremiah:
## Summary

`BattleEngine` accumulates DoT/HoT combat-log lines per second and only flushed the partial second when a battle **concluded** (victory/defeat/draw, via `flushEffectDamage()`). The abandon-shaped ends never flushed:

- `rest()` — parking in the no-combat Home zone — dropped to Idle without flushing.
- `reset(enemyInstance)` — arming the next battle (including after a retreat/zone swap) — called `resetEffectDamage()` directly, discarding whatever the current partial second had accumulated.

This contradicts the documented contract in `docs/game-design.md` → Logs: "each over-time channel is accumulated and emitted as a single line per second (with any partial second flushed at battle end)." A player retreating from a boss (or stepping Home / swapping zones) mid-window could lose up to ~1s of DoT/HoT that they actually took/dealt from the combat log. Cosmetic only — no parity or state impact (`flushEffectDamage()` only logs; it doesn't touch game state beyond what `resetEffectDamage()` already did).

## Changes

- `UI/src/lib/engine/battle/battle-engine.ts`
  - `rest()`: calls `flushEffectDamage()` before dropping to Idle.
  - `reset()`: replaced the direct `resetEffectDamage()` call with `flushEffectDamage()` (which logs any pending partial-second summary, then resets — same net state as before, plus the log line). Order preserved: still runs before `resetPlayer()`/`enemy.reset()`, so the flushed line correctly references the enemy being left behind.

## Tests

Added two cases in `UI/src/tests/lib/engine/battle/battle-engine.test.ts`:
- `rest()` flushes a partial-second DoT window (10 ticks / 400ms, under the 1s auto-flush threshold) before dropping to Idle.
- `reset()` flushes a partial-second DoT window from the abandoned battle before arming the next one.

## Test plan

- [x] `npx vitest run src/tests/lib/engine/battle/battle-engine.test.ts` — 66/66 passing (including the 2 new cases).
- [x] `npm run test:unit:ci` — full suite: 3621/3621 passing across 296 files.
- [ ] Backend unaffected (frontend-only, presentation-log change) — no backend verification needed.

Closes #1857

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kx1Vwe2heKv6HDFhw3Y519)_

### DIFF
--- a/UI/src/lib/engine/battle/battle-engine.ts
+++ b/UI/src/lib/engine/battle/battle-engine.ts
@@ -180,6 +180,7 @@ export class BattleEngine {
 	 */
 	public rest() {
 		this.finishLoading?.();
+		this.flushEffectDamage();
 		this.timeElapsed = 0;
 		this.setBattleStage(Idle);
 	}
@@ -201,7 +202,7 @@ export class BattleEngine {
 		// Seed the battle RNG from the enemy instance's seed (the same value the backend simulates against),
 		// so the crit/dodge/block draws stay in lockstep with the anti-cheat replay.
 		this.rng = new Mulberry32(enemyInstance.seed);
-		this.resetEffectDamage();
+		this.flushEffectDamage();
 		this.resetPlayer();
 		this.enemy.reset({ ...enemyInstance, ...enemyData[enemyInstance.id] });
 		// A non-null elapsedOffsetMs (#1595/#1596) means the server handed back a battle already in
--- a/UI/src/tests/lib/engine/battle/battle-engine.test.ts
+++ b/UI/src/tests/lib/engine/battle/battle-engine.test.ts
@@ -978,6 +978,75 @@ describe('BattleEngine', () => {
 			expect(effectCalls).toContainEqual([ELogType.SkillEffect, 'Goblin recovered 5 health.']);
 		});
 
+		it('flushes a partial-second DoT window when the player rests (retreat / Home / zone swap, #1857)', () => {
+			mockSkills[0].baseDamage = 0; // keep the focus on DoT — no skill kill
+			engine.start();
+			enemyLoadedCallbacks[0]({
+				id: 1,
+				level: 1,
+				seed: 0,
+				enemyRating: 100,
+				isBossBattle: false,
+				selectedSkills: [0],
+				attributes: [
+					{ attributeId: EAttribute.Endurance, amount: 200 }, // survive the partial window
+					{ attributeId: EAttribute.BleedDamagePerSecond, amount: 25 }
+				]
+			});
+
+			// 10 ticks of 40ms = 400ms — under the 1-second auto-flush threshold, so nothing has logged yet.
+			for (let i = 0; i < 10; i++) {
+				logicalUpdateCallbacks[0](40);
+			}
+			expect(logMessage).not.toHaveBeenCalledWith(ELogType.SkillEffect, expect.stringContaining('damage over time'));
+
+			engine.rest();
+
+			const dotCalls = vi
+				.mocked(logMessage)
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Flush partial-second DoT/HoT combat log on abandon-shaped battle ends (#1857)

**Verdict:** Approve — correct, minimal fix that matches the documented "any partial second flushed at battle end" contract; scope, tests, and docs all line up. (Submitted as a comment because I'm the PR author, so GitHub won't accept a gating approve from me.)

I verified the substance rather than just the diff:

- **Correctness of the `reset()` swap.** `flushEffectDamage()` (battle-engine.ts:461) ends by calling `resetEffectDamage()` (line 475), so replacing the direct `resetEffectDamage()` with `flushEffectDamage()` produces the exact same net state — zeroed channels and window — plus the pending log line. The PR's "same net state as before" claim holds.
- **Ordering.** In `reset()` the flush runs before `resetPlayer()`/`enemy.reset()`, so the emitted line still references the enemy being left behind — as intended. On the very first battle the channels are all 0, so the `> 0` guards emit nothing and `this.enemy.name` is never read against an unarmed enemy. Safe.
- **Scoping is right, not incomplete.** `pause()` correctly does *not* flush (it resumes the same battle mid-swap, so accumulation must continue), and `stop()` is a full loop teardown for leaving the game screen — the subsequent re-arm goes through `reset()`, which now flushes. Excluding both from this change is the correct call, matching the issue's `rest()`/`reset()` scope.
- **No parity concern.** Grepped `Game.Core`/`Game.Application` — there's no backend DoT/HoT combat-log accumulation; this is a frontend-only presentation channel, so the "no parity impact" note is accurate and no mirrored backend test is required.

### Test coverage

Both new abandon paths are covered — `rest()` and `reset()` each flush a sub-1s DoT window (400ms, under the 1s auto-flush threshold) and assert exactly one `"Goblin took 10 damage over time."` line (25 dmg/s × 0.4s = 10). Each test first asserts nothing logged pre-abandon, which properly pins the "partial second was otherwise silently dropped" behavior the issue describes. Good coverage for the change.

### What's good

Tight, well-reasoned fix: it reuses `flushEffectDamage`'s existing flush-then-reset semantics instead of adding a parallel code path, the comments explain the ordering intent, and the tests target exactly the two regressions from #1857.

CI is still finishing (frontend unit + e2e firefox/webkit in progress; chromium e2e and verify-frontend already green, backend correctly skipped). Will merge once it's green.

---

PR #1879, opened 2026-07-12
Memoize offline simulator's combat rating and battler materials

### LINKED ISSUE #1730: Offline simulator recomputes invariant combat ratings and re-composes the modifier list every simulated battle
## Problem

Per-battle constant work in the offline loop that only actually changes on a level-up:

- `Game.Core/Battle/Offline/OfflineProgressSimulator.cs:102` — each victory re-runs `new DefeatRewards(playerBattlerForRating, enemy)`, and `CombatRating.Rate` is not cheap (`BuildEffectiveCaster`/`BuildReferenceDefense`, `CombatRating.cs:306/365`, each materialize a full ~50-modifier set plus fresh `AttributeCollection`s). The rating battler itself is already rebuilt only on level transitions — but its rating is recomputed every battle. In Boss mode the enemy is fully deterministic, so `EnemyRating` is recomputed identically for every battle in the window.
- `OfflineProgressSimulator.cs:251-258` — `SimulateBattle` → `snapshot.ToBattler(...)` re-walks the whole resolver/LINQ modifier composition per battle although the composed list changes only on a player/proficiency level-up.

A 10-hour away window can credit thousands of battles, so this is a real constant-factor cost on the offline path (and on character switching, which reuses the simulator).

## Fix shape

Memoize `PlayerRating` alongside the existing `playerBattlerForRating` rebuild point (and the boss rating per run), and cache the composed base-modifier list between level-ups. No behavior change; the existing offline suite (in-loop growth, boundary math) pins correctness.

ginderjeremiah:
## Summary

`OfflineProgressSimulator` was doing constant per-battle work that only actually changes on a level-up or proficiency level-up:

- Every victory re-ran `new DefeatRewards(playerBattlerForRating, enemy)`, which calls `CombatRating.Rate` — not cheap, it materializes a full ~50-modifier attribute set twice (`BuildEffectiveCaster`/`BuildReferenceDefense`) — even though `playerBattlerForRating` itself was already only rebuilt on a level/proficiency level-up.
- In Boss mode, the enemy is fully deterministic (`CreateBossEnemy`: fixed level, full authored loadout, no random roll), so its rating was being recomputed identically on every single boss victory in a run.
- `SimulateBattle` called `snapshot.ToBattler(...)` every battle, re-walking the whole resolver/LINQ modifier composition (stat allocations, gear, class locked base, proficiency bonuses) even though everything except the level/proficiency-driven pieces is frozen for the entire run.

A 10-hour away window can credit thousands of battles, so this was a real constant-factor cost on the offline path (and on character switching, which reuses the simulator).

## Changes

- **`Game.Core/Battle/DefeatRewards.cs`**: adds a `DefeatRewards(double playerRating, double enemyRating)` overload so a caller with already-known ratings can skip re-deriving them. The existing `DefeatRewards(Battler, Enemy)` constructor now delegates to it.
- **`Game.Core/Battle/BattleSnapshot.cs`**: splits `ToBattler` into `GetBattlerMaterials` (the reusable, resolver/LINQ-composed ingredients — base attribute modifiers, resolved signature passive, fielded skills, counter skill) and a new `BattlerMaterials.Build()` (the cheap, always-fresh `AttributeCollection`/`Battler` construction a single-use battler needs every battle). `ToBattler` itself is now a thin wrapper (`GetBattlerMaterials(...).Build()`), so its existing callers and tests are unaffected.
- **`Game.Core/Battle/Offline/OfflineProgressSimulator.cs`**:
  - Caches `BattlerMaterials` and the player's `CombatRating.Rate` result, rebuilding both only when a level-up or proficiency level-up actually moves an attribute modifier (the same trigger the old rating-battler rebuild already used).
  - Memoizes the boss's rating for the whole run in Boss mode; still recomputed per victory in Idle mode, where each random encounter genuinely differs.
  - `SimulateBattle` now builds its per-battle `Battler` from the cached `BattlerMaterials` instead of re-walking `ToBattler`'s full composition.

No behavior change — this is a pure memoization refactor. The reward/rating math itself is untouched (same `CombatRating.Rate` calls, same `DefeatRewards` formula, same composition order for parity).

## Test plan

- [x] `dotnet test` — full backend suite passes: 3080/3080 (`Game.Core.Tests` 1347, `Game.Infrastructure.Tests` 62, `Game.Application.Tests` 917, `Game.Api.Tests` 754), including the existing offline-simulator suite that pins in-loop growth, boundary math, and rating stationarity between level-ups (`Simulate_MidWindowLevelUps_GrowClassLockedBaseAndPlayerRating`, `Simulate_BossMode_LevelUps_GrowPlayerRatingAcrossTheWindow`, `Simulate_PendingBattle_SnapshotReflectsMidWindowGrowth_NotTheWindowStartSnapshot`, etc.) and the `DefeatRewards`/`BattleSnapshot`/`CombatRating`/battle-simulator-parity suites.

Closes #1730

---
_Generated by [Claude Code](https://claude.ai/code/session_012uZkkYh7MDsSx9BtsYuTjC)_

### DIFF
--- a/Game.Core/Battle/Offline/OfflineProgressSimulator.cs
+++ b/Game.Core/Battle/Offline/OfflineProgressSimulator.cs
@@ -64,13 +64,21 @@ namespace Game.Core.Battle.Offline
             var catalog = new ProficiencyCatalog(
                 parameters.ResolveProficiency, parameters.ResolvePath, parameters.PathsForActivityKey, parameters.DependentsOf);
 
-            // The battler that measures each victory's exp reward is re-derived whenever a level-up grows the
-            // class locked base (and, through it, the signature passive) — rebuilding it eagerly here and again
-            // after every level-up, rather than lazily on next use, keeps the "current rating battler" a single
-            // invariant read at each DefeatRewards call below. It is never simulated (CombatRating.Rate only
-            // reads it), so reusing the same instance across battles between level-ups is safe even though a
-            // Battler is otherwise single-use.
-            var playerBattlerForRating = BuildRatingBattler(workingSnapshot, parameters);
+            // The materials that assemble the player's battler are re-derived whenever a level-up grows the
+            // class locked base (and, through it, the signature passive) or a proficiency level-up grows its
+            // bonuses — rebuilding them eagerly here and again after every such growth, rather than on every
+            // battle, is what lets SimulateBattle build a fresh single-use Battler per battle without re-walking
+            // the resolver/LINQ composition each time. playerRating is memoized alongside them: CombatRating.Rate
+            // is not cheap (it materializes a full attribute set twice, #1730), and it depends only on these same
+            // materials, so it is recomputed exactly when they are — not every battle.
+            var playerMaterials = BuildMaterials(workingSnapshot, parameters);
+            var playerRating = CombatRating.Rate(playerMaterials.Build(), isPlayer: true);
+
+            // The boss's own rating, memoized for the whole run: CreateBossEnemy is fully deterministic (fixed
+            // level, full authored loadout, no random roll), so in Boss mode every battle's enemy rates
+            // identically and re-deriving it per victory would be pure waste. Left null (and never consulted)
+            // in Idle mode, where each battle's random encounter genuinely differs.
+            double? bossEnemyRating = null;
 
             // Tracks whether any battle has produced progress (a win or a loss). A run that is nothing but
             // draws is a stalemate the player can neither win nor lose; the cutoff below stops it early.
@@ -84,7 +92,7 @@ namespace Game.Core.Battle.Offline
 
                 var enemy = BuildEnemy(parameters);
                 var seed = parameters.SeedSource();
-                var result = SimulateBattle(workingSnapshot, parameters, enemy, seed);
+                var result = SimulateBattle(playerMaterials, enemy, seed);
 
                 if (result.TotalMs > remainingMs)
                 {
@@ -97,10 +105,16 @@ namespace Game.Core.Battle.Offline
                 }
 
                 // Rewards are earned only on a victory, measured from the current (possibly mid-window-grown)
-                // rating battler like the live path. The same DefeatRewards yields both the exp and the combat
+                // player rating like the live path. The same DefeatRewards yields both the exp and the combat
                 // ratings the offline proficiency-XP accrual normalizes each path's activity by, so the two
-                // payouts share one evaluation.
-                var rewards = result.Victory ? new DefeatRewards(playerBattlerForRating, enemy) : null;
+                // payouts share one evaluation. The enemy rating reuses the memoized boss rating in Boss mode
+                // (populating it on the first victory) and is re-derived per victory in Idle mode, where each
+                // random encounter genuinely differs.
+                var rewards = result.Victory
+                    ? new DefeatRewards(playerRating, parameters.Mode == OfflineLoopMode.Boss
+                        ? bossEnemyRating ??= CombatRating.Rate(enemy.ToBattler(), isPlayer: false)
+                        : CombatRating.Rate(enemy.ToBattler(), isPlayer: false))
+                    : null;
                 var proficiencyGains = ProficiencyAccrualResult.Empty;
                 if (rewards is not null)
                 {
@@ -151,14 +165,16 @@ namespace Game.Core.Battle.Offline
                         workingSnapshot.Level = progression.Level;
                     }
 
-                    // Re-derive the rating battler whenever either growth vector actually moved an attribute
-                    // modifier: a player level-up (the locked base) or a proficiency level-up (its per-level/
-                    // milestone bonus) — never on XP alone, which changes nothing ModifiersForLevel reads. This
-                    // is what makes the reward-power measurement — and through it the proficiency accrual's own
-                    // rating denominator on the next battle — grow with proficiency, not just level.
+                    // Re-derive the materials (and the rating they measure) whenever either growth vector
+                    // actually moved an attribute modifier: a player level-up (the locked base) or a proficiency
+                    // level-up (its per-level/milestone bonus) — never on XP alone, which changes nothing
+                    // ModifiersForLevel reads. This is what makes the reward-power measurement — and through it
+                    // the proficiency accrual's own rating denominator on the next battle — grow with
+                    // proficiency, not just level.
                     if (progression.LevelsGained > 0 || proficiencyLeveledUp)
                     {
-                        playerBattlerForRating = BuildRatingBattler(workingSnapshot, parameters);
+                        playerMaterials = BuildMaterials(workingSnapshot, parameters);
+                        playerRating = CombatRating.Rate(playerMaterials.Build(), isPlayer: true);
                     }
                 }
 
@@ -218,11 +234,12 @@ namespace Game.Core.Battle.Offline
         };
 
         /// <summary>
-        /// Builds the battler <see cref="DefeatRewards"/> measures the player's power from — reused across
-        /// battles until the next level-up requires rebuilding it.
+        /// Builds the materials both the per-battle simulated battler and the <see cref="DefeatRewards"/> rating
+        /// battler assemble from — reused across battles until the next level-up or proficiency level-up
+        /// requires rebuilding them (#1730).
         /// </summary>
-        private static Battler BuildRatingBattler(BattleSnapshot snapshot, OfflineSimulationParameters parameters) =>
-            snapshot.ToBattler(
+        private static BattlerMaterials BuildMaterials(BattleSnapshot snapshot, OfflineSimulationParameters parameters) =>
+            snapshot.GetBattlerMaterials(
                 parameters.ResolveItem, parameters.ResolveMod, parameters.ResolveSkill,
                 parameters.ResolveProficiency, parameters.ResolveClass);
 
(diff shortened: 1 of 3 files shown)


Game.Core/Battle/Offline/OfflineProgressSimulator.cs
```diff
@@ -97,10 +105,16 @@ public OfflineProgressResult Simulate(OfflineSimulationParameters parameters, Ca
...
-                // payouts share one evaluation.
-                var rewards = result.Victory ? new DefeatRewards(playerBattlerForRating, enemy) : null;
+                // payouts share one evaluation. The enemy rating reuses the memoized boss rating in Boss mode
+                // (populating it on the first victory) and is re-derived per victory in Idle mode, where each
+                // random encounter genuinely differs.
+                var rewards = result.Victory
+                    ? new DefeatRewards(playerRating, parameters.Mode == OfflineLoopMode.Boss
+                        ? bossEnemyRating ??= CombatRating.Rate(enemy.ToBattler(), isPlayer: false)
+                        : CombatRating.Rate(enemy.ToBattler(), isPlayer: false))
+                    : null;
```
ginderjeremiah: Nit (readability): the boss-rating memoization happens via a `??=` compound-assignment side effect buried inside a nested ternary that's itself an argument to the `DefeatRewards(...)` constructor. It's correct and well-commented, but it's exactly the kind of "deliberate-looking construct a later edit might naively simplify and break" — a reader could easily miss that `bossEnemyRating ??= ...` writes back to the outer loop variable, or flatten the nested ternary and lose the cache. Consider lifting the enemy rating into an explicit local before the `new DefeatRewards(...)`, e.g.:

```csharp
double enemyRating;
if (parameters.Mode == OfflineLoopMode.Boss)
{
    enemyRating = bossEnemyRating ??= CombatRating.Rate(enemy.ToBattler(), isPlayer: false);
}
else
{
    enemyRating = CombatRating.Rate(enemy.ToBattler(), isPlayer: false);
}
var rewards = result.Victory ? new DefeatRewards(playerRating, enemyRating) : null;
```

Take it or leave it — the intent is clear enough with the comment.

---

ginderjeremiah (commented):
## Review: Memoize the offline simulator's combat rating and battler materials (#1730)

**Verdict:** Approve (posting as `COMMENT` since I'm the PR author, so GitHub won't accept a gating event on my own PR) — a clean, correct pure-memoization refactor; only nits, no blockers.

I verified the "no behavior change" claim by reading the composition paths end to end rather than trusting it:

- **Parity of `GetBattlerMaterials`/`Build` vs. the old `ToBattler`** — identical: same `GetModifiers()` base set, the signature passive is still resolved against a `new AttributeCollection(baseModifiers)` (i.e. base + static, since the ctor re-adds `StaticAttributeModifiers`) and added *last*, and `Build()` reconstructs the `AttributeCollection`/`Battler` the same way. `ToBattler` is now a thin `GetBattlerMaterials(...).Build()` wrapper, so its existing callers/tests still exercise the whole path.
- **Reuse safety of the cached materials** — sharing `baseModifiers`, `signaturePassive`, and `skills` across many *simulated* battlers is safe: `AttributeModifier` is immutable, the simulation only ever adds/removes its own fresh `SkillEffect` modifiers (never touches the shared list), `Skill` is read-only reference data (re-wrapped in a fresh `BattleSkill` per battler), and each `Build()` allocates a fresh `AttributeCollection`/`Battler`.
- **`DefeatRewards(double, double)` overload** — the `(Battler, Enemy)` ctor now delegates to it, so all four fields are set with byte-identical math; the existing `DefeatRewardsTests` cover it through the delegation.
- **Boss-rating memoization correctness** — confirmed against `BattleFactory.CreateBossEnemy`: it resolves at the fixed `Zone.BossLevel` (no level roll) and fields `SelectAllBattleSkills()` (full authored loadout, no random selection), so the *fielded* loadout `CombatRating.Rate` reads is deterministic for the whole run. Idle mode correctly leaves `bossEnemyRating` null and re-derives per encounter, where `CreateBattleEnemy` genuinely rolls level + skills.

### Nits

- `OfflineProgressSimulator.cs:113-117` — the boss-rating cache is a `??=` side effect inside a nested ternary passed as a ctor argument. Correct and commented, but a readability foot-gun a later "simplify" could break; suggest lifting the enemy rating into an explicit local (inline comment has a concrete form).

### Test coverage

Adequate — the refactor introduces no new *observable* behavior, and everything new is exercised transitively: `GetBattlerMaterials`/`Build` through the `ToBattler` wrapper (`BattleSnapshotTests`, parity suites), the new `DefeatRewards` overload through the delegating ctor (`DefeatRewardsTests`), and the memoization behavior through the offline suite's rating-stationarity-between-level-ups, boss-mode growth, and mid-window-growth tests. A direct assertion that the cached-materials path yields results identical to a fresh per-battle `ToBattler` (a memoization-parity test) would harden it against a future regression, but it's optional given the existing pins.

### What's good

- The `BattlerMaterials` split lands the reuse boundary exactly where it belongs — the invariant resolver/LINQ composition is separated from the always-fresh per-battle construction, and the doc comments make *why* it's safe to cache explicit.
- The rebuild trigger (`LevelsGained > 0 || proficiencyLeveledUp`) reuses the existing growth-detection point rather than inventing a parallel one, so the memoized `playerRating` and materials stay in lockstep with the class-locked-base/proficiency growth.
- Boss vs. idle enemy-rating handling correctly distinguishes the deterministic case from the genuinely-varying one, rather than blanket-caching.

Not merging: CI is still running (backend/e2e in progress) and there's an outstanding nit, so this stays a review rather than an auto-merge.

---

ginderjeremiah:
## Review: Approve

Reviewed the full diff against the baseline (`main`) source for `OfflineProgressSimulator`, `BattleSnapshot`, `DefeatRewards`, plus `AttributeCollection`, `CombatRating`, and `Battler`. This is a correct, clean pure-memoization refactor. **CI is green** on head `c00fd910` (test-backend, verify-backend, all e2e succeeded; frontend jobs correctly skipped for a backend-only change).

### The "no behavior change" claim holds — verified independently

**Cache-invalidation triggers are identical to the pre-PR logic.** `playerMaterials` + `playerRating` are rebuilt on exactly `progression.LevelsGained > 0 || proficiencyLeveledUp` — the same trigger the old `playerBattlerForRating` rebuild used. Because the materials now also feed the per-battle *simulated* battler, and the only in-loop `workingSnapshot` mutations that affect composition are `Level` and `ProficiencyLevels[].Level` (proficiency `Xp` is explicitly not read by `GetModifiers`/`ModifiersForLevel`), the cached materials can never go stale between rebuilds. Timing is preserved: battle N is simulated and rated from state as of the end of N-1's growth in both the old and new code.

**Shared-cache mutation safety — the real risk here — is sound.** `BattlerMaterials.Build()` reuses the cached `baseModifiers`/`signaturePassive`/`skills`/`counterSkill` across thousands of battles, but each `Build()` produces a fully independent, mutable `Battler`:
- `new AttributeCollection(baseModifiers)` *copies* modifiers into its own per-attribute node storage (it iterates, never retains the list); `AddModifier(signaturePassive)` and all in-battle effect add/remove operate on that per-collection storage, never the cached list.
- `Battler` wraps the shared `Skill` instances in *fresh* `BattleSkill` objects (`skills.Select(s => new BattleSkill(s)).ToList()`), so per-battle cooldown state is isolated; the cached `skills` list and the immutable reference-data `Skill`/`counterSkill` are read-only.
- The rating battler and the simulated battler come from separate `Build()` calls, so the simulation mutating one never touches the other.

**Boss-rating memoization is valid.** `CreateBossEnemy` is deterministic (fixed zone boss level, full authored loadout, no roll/shuffle) and enemy rating is independent of player growth, so it is genuinely invariant for the whole run — `bossEnemyRating ??=` on first victory is correct. Idle mode still recomputes per victory (`CombatRating.Rate(enemy.ToBattler(), ...)` in the else branch), which is right since each random encounter differs in level/loadout.

**Composition order / parity preserved.** `GetBattlerMaterials` composes base modifiers then resolves the signature passive against a transient `AttributeCollection(baseModifiers)` (same base-only view the old `ToBattler` resolved it against — identical to the established `GetModifiersWithSignaturePassive` pattern), and `Build()` re-adds it last via `AddModifier`. `ToBattler` is now `GetBattlerMaterials(...).Build()`, byte-identical.

**Delegated `DefeatRewards` ctor is safe** despite swapping the two `CombatRating.Rate` evaluations (player-first now vs enemy-first before): `Rate` is a pure function of its battler — no RNG, no shared mutable state, no input mutation — so order is irrelevant, and the reward formula is untouched.

### Standards
DDD/layering clean (all in `Game.Core`); no dead code (old `BuildRatingBattler`/`playerBattlerForRating` fully removed); no `!` operators introduced; all conditionals braced; DRY improved via the ctor delegation. Comment density matches the surrounding code.

### Nit (non-blocking)
The equivalence of the two new seams — `BattlerMaterials.Build()` ≡ `ToBattler(...)`, and `DefeatRewards(double, double)` ≡ `DefeatRewards(Battler, Enemy)` — is currently pinned only transitively (through the offline-sim suite and existing `ToBattler`/`DefeatRewards` tests). A couple of direct micro-tests would lock these extractions explicitly, but the existing coverage genuinely exercises both paths, so this is optional.

No follow-up issues warranted — nothing here is out of scope.

**Recommendation: mergeable-clean.**

---

PR #1880, opened 2026-07-12
Clamp replayToOffset's fast-forward to DEFAULT_MAX_BATTLE_MS

### LINKED ISSUE #1733: replayToOffset claims a DEFAULT_MAX_BATTLE_MS bound its loop doesn't enforce
## Problem

The doc comment on `BattleEngine.replayToOffset` (`UI/src/lib/engine/battle/battle-engine.ts:224-235`) says the fast-forward is "Bounded by `DEFAULT_MAX_BATTLE_MS` … so at most ~3000 ticks", but the loop runs to whatever `offsetMs` arrives — nothing clamps it.

Today only the trusted server supplies the offset, so this is defensive rather than a live defect. But a malformed hand-back (e.g. a backend regression returning a raw away-window offset) would synchronously run ~900k `battleStep` calls on the main thread, and could then declare a `Drawn` at a `timeElapsed` past the cap — a state the headless simulator can never reach.

## Fix

One line: `Math.min(offsetMs, DEFAULT_MAX_BATTLE_MS)` (making the comment true), plus a test pinning that an oversized offset stops at the cap.

ginderjeremiah:
## Summary

`BattleEngine.replayToOffset`'s doc comment (`UI/src/lib/engine/battle/battle-engine.ts`) claimed the fast-forward replay was "Bounded by `DEFAULT_MAX_BATTLE_MS` … so at most ~3000 ticks," but the loop actually ran to whatever `offsetMs` arrived — nothing clamped it.

Today only the trusted server supplies the offset (bounded server-side, matching the backend's own `BattleSimulator.Simulate`, which already bounds its tick loop by `limit`), so this was defensive rather than a live defect. But a malformed hand-back (e.g. a backend regression returning a raw away-window offset) would have synchronously run an unbounded number of `battleStep` calls on the main thread, and could then declare a `Drawn` at a `timeElapsed` past the cap — a state the headless parity simulator can never reach.

## Changes

- `battle-engine.ts`: `replayToOffset` now clamps `offsetMs` to `DEFAULT_MAX_BATTLE_MS` via `Math.min` before the replay loop, and the doc comment is updated to describe the clamp instead of merely asserting it.
- `battle-engine.test.ts`: new test pinning that an oversized offset (`DEFAULT_MAX_BATTLE_MS * 100`) stops the replay at the cap and resolves `Drawn`, mirroring the existing "offset lands exactly on the cap" test.

No behavior change for the trusted-server path — this only changes what happens with an out-of-contract offset.

## Test plan

- [x] `npx vitest run src/tests/lib/engine/battle/battle-engine.test.ts` — 67/67 passing, including the new clamp test.
- [x] `npm run test:unit:ci` — full frontend suite passes: 3622/3622.
- [ ] Backend unaffected (frontend-only defensive fix; the backend simulator already bounds its own loop) — no backend verification needed.

Closes #1733

---
_Generated by [Claude Code](https://claude.ai/code/session_01WBvpKPUox6evW8vuoxQ442)_

### DIFF
--- a/UI/src/lib/engine/battle/battle-engine.ts
+++ b/UI/src/lib/engine/battle/battle-engine.ts
@@ -222,12 +222,14 @@ export class BattleEngine {
 	 * backend `BattleSimulator.Simulate`'s loop exactly (tick death-checks, then the `totalMs - tickSize`
 	 * exit value) so the two stay in lockstep. Resolves into whichever stage {@link logicalUpdate} would
 	 * have reached by that point — defensive: the server only ever hands back an unconcluded, under-cap
-	 * battle (#1595/#1596), but the tick boundary makes replaying a conclusion possible. Bounded by
-	 * `DEFAULT_MAX_BATTLE_MS` (the server never reports a larger offset), so at most ~3000 ticks.
+	 * battle (#1595/#1596), but the tick boundary makes replaying a conclusion possible. Clamped to
+	 * `DEFAULT_MAX_BATTLE_MS` (the server never reports a larger offset, but a malformed hand-back
+	 * shouldn't run the main thread through an unbounded replay), so at most ~3000 ticks.
 	 */
 	private replayToOffset(offsetMs: number) {
+		const cappedOffsetMs = Math.min(offsetMs, DEFAULT_MAX_BATTLE_MS);
 		let totalMs = tickSize;
-		for (; totalMs <= offsetMs; totalMs += tickSize) {
+		for (; totalMs <= cappedOffsetMs; totalMs += tickSize) {
 			battleStep(this.player, this.enemy, tickSize, this.rng);
 			if (this.enemy.isDead || this.player.isDead) {
 				break;
--- a/UI/src/tests/lib/engine/battle/battle-engine.test.ts
+++ b/UI/src/tests/lib/engine/battle/battle-engine.test.ts
@@ -582,6 +582,27 @@ describe('BattleEngine', () => {
 			expect(engine.timeElapsed).toBe(DEFAULT_MAX_BATTLE_MS);
 			expect(logMessage).toHaveBeenCalledWith(ELogType.EnemyDefeated, expect.stringContaining('draw'));
 		});
+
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Clamp `replayToOffset`'s fast-forward to `DEFAULT_MAX_BATTLE_MS` (closes #1733)

**Verdict:** Approve _(submitted as a comment — GitHub won't accept a gating approval on my own PR)_ — a minimal, correct defensive fix that makes the doc comment's long-standing claim actually true.

The change does exactly what the issue asks: `replayToOffset` now clamps `offsetMs` to `DEFAULT_MAX_BATTLE_MS` before the replay loop, so an out-of-contract hand-back can't run an unbounded number of `battleStep` calls on the main thread or resolve a `Drawn` at a `timeElapsed` past the cap.

Verified the math holds: `DEFAULT_MAX_BATTLE_MS` (120000ms) is an exact multiple of `tickSize` (40ms) = 3000 ticks, so with an oversized offset the loop exits with `totalMs` one tick past the cap and `timeElapsed = totalMs - tickSize` lands exactly on `DEFAULT_MAX_BATTLE_MS` → `resolveConclusion()` returns `Drawn`. The new test pins precisely that.

### Parity
No FE/BE parity concern. The clamp only alters behavior for offsets the trusted server never emits (the backend's `BattleSimulator.Simulate` already bounds its own loop by `limit`), and the battle-step math itself is untouched — so the "mirror battle-logic tests on both sides" rule doesn't apply to this frontend-only guard.

### Test coverage
Adequate. The new test mirrors the existing "offset lands exactly on the cap" case with a grossly oversized offset (`DEFAULT_MAX_BATTLE_MS * 100`) and a true stalemate (`baseDamage = 0`), asserting the replay stops at the cap and resolves `Drawn`. Good use of an existing sibling test as the template.

### What's good
- Scope discipline — a genuinely one-line behavioral change plus a focused test, no gold-plating.
- The doc comment was updated to describe the clamp rather than just asserting the bound, closing the exact doc-vs-code drift #1733 flagged.

---

PR #1886, opened 2026-07-12
Battle tick-loop micro-optimizations: precompute amp/resist arrays; gate log formatting on enabled types

### LINKED ISSUE #1734: Battle tick-loop micro-optimizations: precompute amp/resist attribute arrays; skip log formatting for disabled log types
Two zero-parity-risk allocations/work reductions in the always-on 25Hz logical loop (which keeps running at full rate in hidden tabs, and is amplified by `replayToOffset`'s synchronous up-to-3000-tick fast-forward):

1. **Per-call array building in the typed-damage lookups.** `amplificationAttributes()`/`resistanceAttributes()` (`UI/src/lib/battle/damage-types.ts:74-90`) rebuild an array on every call — per portion per hit, per DoT type per tick, plus per activation in `classifyResist` (`battle-engine.ts:375`). The backend already precomputes exactly this (`DamageTypes.AmplificationByType`/`ResistanceByType`, `Game.Core/Attributes/DamageTypes.cs:218-227`), and the TS module already precomputes its other inverses (`KEY_BY_ATTRIBUTE`, `DOT_TYPE_BY_ACCUMULATOR`). Same values, same fixed order — no parity surface.

2. **Log-line strings are built before the enabled check.** `battle-engine.ts:376-381` builds `damageLogMessage` (plus `classifyResist`/`attributeName` work in `logEffectApplications`) for every activation; `logMessage` (`log.ts:23-27`) then discards it if the player disabled that log type. Gate on `playerManager.logTypeEnabled` before formatting — Damage is the busiest channel in a hidden tab.

ginderjeremiah:
## Summary

Two zero-parity-risk perf fixes in the always-on 25Hz logical battle loop (`UI/src/lib/engine/battle/battle-engine.ts`), which keeps running at full rate in hidden tabs and is amplified by `replayToOffset`'s synchronous up-to-3000-tick fast-forward:

1. **Per-call array building in the typed-damage lookups.** `amplificationAttributes()`/`resistanceAttributes()` (`UI/src/lib/battle/damage-types.ts`) rebuilt an array on every call — per portion per hit, per DoT type per tick, plus per activation in `classifyResist`. Both are now precomputed once into `ReadonlyMap<EDamageType, readonly EAttribute[]>` at module load, keyed by leaf damage type, mirroring the backend's already-precomputed `DamageTypes.AmplificationByType`/`ResistanceByType`. Same values, same fixed iteration order — no parity surface (the fold order downstream in `battle-formulas.ts` is unchanged).
2. **Log-line strings built before the enabled check.** `logMessage` (`log.ts`) already discards a message if the player disabled that log channel, but `battle-engine.ts` built `damageLogMessage`/`reflectLogMessage` (plus `classifyResist`/`resistanceTotal`) and `effectLogMessage` (plus `attributeName`/`attributeIsHarmful`) unconditionally before calling it — Damage is the busiest channel in a hidden tab. Both paths now check `playerManager.logTypeEnabled(...)` first and skip the formatting work entirely when disabled. The combat floaters (`notifyCombatFloat`) are untouched — they don't depend on the log text, only on the (still-computed) `outcome`/`damageType`.

No behavior change: `applies()`/`attributesForKey()` and their fixed iteration order are untouched, and the existing damage-types/battle-formulas/battle-simulation-parity suites (which pin exact values and order) all pass unchanged.

## Changes

- `UI/src/lib/battle/damage-types.ts`: `amplificationAttributes`/`resistanceAttributes` now read from precomputed maps instead of rebuilding arrays; return type widened to `readonly EAttribute[]` since the arrays are now shared/cached (callers only ever iterate them).
- `UI/src/lib/engine/battle/battle-engine.ts`: `logicalUpdate`'s damage/reflect log lines and `logEffectApplications`'s skill-effect log lines are now gated behind `playerManager.logTypeEnabled(...)`.
- Test mocks (`battle-engine.test.ts`, `battle-engine-parity.test.ts`) updated with a `logTypeEnabled` stub (defaulting to enabled, matching the real manager's `?? true` fallback), plus two new tests pinning that disabling a log channel skips `logMessage` for that channel while the combat float still fires.

## Test plan

- [x] `npx vitest run` (`test:unit:ci`, full suite with coverage) — 3624/3624 passing.

Closes #1734

---
_Generated by [Claude Code](https://claude.ai/code/session_016juTUTqQRBgiSZnnmVWuce)_

### DIFF
--- a/UI/src/lib/battle/damage-types.ts
+++ b/UI/src/lib/battle/damage-types.ts
@@ -69,24 +69,36 @@ export function attributesForKey(key: EDamageTypeKey): DamageTypeKeyAttributes {
 	return DAMAGE_TYPE_KEY_ATTRIBUTES[key];
 }
 
+// Amplification/resistance attribute lists per leaf type, precomputed once (mirrors the backend's
+// `DamageTypes.AmplificationByType`/`ResistanceByType`) so the per-hit/per-tick lookup in the damage
+// pipeline is a map read rather than rebuilding an array on every call.
+const AMPLIFICATION_ATTRIBUTES_BY_TYPE: ReadonlyMap<EDamageType, readonly EAttribute[]> = new Map(
+	Object.entries(DAMAGE_TYPE_APPLIES).map(([type, keys]) => [
+		Number(type) as EDamageType,
+		keys.map((key) => DAMAGE_TYPE_KEY_ATTRIBUTES[key].amplification)
+	])
+);
+
+const RESISTANCE_ATTRIBUTES_BY_TYPE: ReadonlyMap<EDamageType, readonly EAttribute[]> = new Map(
+	Object.entries(DAMAGE_TYPE_APPLIES).map(([type, keys]) => [
+		Number(type) as EDamageType,
+		keys
+			.map((key): EAttribute | null => DAMAGE_TYPE_KEY_ATTRIBUTES[key].resistance)
+			.filter((resistance): resistance is EAttribute => resistance !== null)
+	])
+);
+
 /** The attacker-side amplification attributes summed for a hit of the given leaf `type`, in fixed
  *  iteration order (per-hit lookup helper for the damage pipeline). */
-export function amplificationAttributes(type: EDamageType): EAttribute[] {
-	return applies(type).map((key) => DAMAGE_TYPE_KEY_ATTRIBUTES[key].amplification);
+export function amplificationAttributes(type: EDamageType): readonly EAttribute[] {
+	return AMPLIFICATION_ATTRIBUTES_BY_TYPE.get(type) ?? [];
 }
 
 /** The defender-side resistance attributes summed for a hit of the given leaf `type`, in fixed
  *  iteration order (per-hit lookup helper for the damage pipeline). Amplification-only weapon keys (#1340)
  *  contribute no resistance, so they drop out — a weapon hit resists via the shared `Physical` key only. */
-export function resistanceAttributes(type: EDamageType): EAttribute[] {
-	const resistances: EAttribute[] = [];
-	for (const key of applies(type)) {
-		const { resistance } = DAMAGE_TYPE_KEY_ATTRIBUTES[key];
-		if (resistance !== null) {
-			resistances.push(resistance);
-		}
-	}
-	return resistances;
+export function resistanceAttributes(type: EDamageType): readonly EAttribute[] {
+	return RESISTANCE_ATTRIBUTES_BY_TYPE.get(type) ?? [];
 }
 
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Battle tick-loop micro-optimizations (precompute amp/resist arrays; gate log formatting) — closes #1734

**Verdict:** Approve _(posted as a comment since I'm the PR author; GitHub won't accept a gating review on one's own PR)_ — both stated optimizations are implemented cleanly with no behavior or parity change, and the new gating behavior is covered by tests.

I verified the two things that actually matter for this change:

- **Parity preserved.** `AMPLIFICATION_ATTRIBUTES_BY_TYPE`/`RESISTANCE_ATTRIBUTES_BY_TYPE` are built from `DAMAGE_TYPE_APPLIES` with the same per-type `keys.map(...)` the old helpers used, so the values and the fixed intra-type iteration order are identical — and both consumers (`amplifiedDamage`, `resistanceTotal` in `battle-formulas.ts`) only fold via `for…of`, never mutate. The `EAttribute[] → readonly EAttribute[]` widening correctly reflects that the arrays are now shared/cached and enforces the no-mutation contract at compile time. Float fold order downstream is untouched.
- **Log gating is safe.** `logEffectApplications` only calls `logMessage`, so the early return when `SkillEffect` is disabled skips no state mutation (`accumulateEffectDamage` is a separate call in `logicalUpdate`). The `Damage`-channel gate keeps `notifyCombatFloat` and the reflect float outside the `if (damageLogEnabled)`, so floaters still fire regardless — the new test pins exactly that. `logTypeEnabled`'s `?? true` fallback matches the log writer's historical default, so a hidden/unknown channel behaves as before.

### Test coverage

Good. Two new `battle-engine.test.ts` cases pin the gating (Damage disabled → no damage log line but the hit still floats; SkillEffect disabled → no effect log line), and the parity mock stubs `logTypeEnabled: () => true`. The precompute is a pure refactor with identical values, already pinned by the existing damage-types/battle-formulas/battle-simulation-parity suites — no new test needed there. Backend needs no change: it already precomputes these maps, and no value/order changed, so FE/BE battle parity holds.

### Follow-up (not blocking)

The issue's later comment flagged a **third** per-tick candidate for "this pass" — `evaluateMechanicTriggers` (`tutorials.ts`) allocating/linear-scanning per firing tick even when all mechanic lessons are already unlocked. This PR correctly scoped to the two items in the issue body and doesn't touch it; I'll open a separate follow-up issue to track it rather than expand this PR's scope.

### What's good

Tight, well-scoped, and the comments explain the *why* (parity, hidden-tab hot path) without over-explaining. Mirroring the backend's already-precomputed structures is the right call.

---

PR #1902, opened 2026-07-12
Carry the abandon-incurred cooldown on NewEnemy's response

### LINKED ISSUE #1881: NewEnemy response never carries the abandon-incurred cooldown anchor — honest clients trip the anti-cheat gate after every idle loss/draw
## Problem

#1851 correctly made `AbandonBattle` apply the 5s post-battle cooldown when an abandon resolves an outcome, and `BattleService.StartBattle` anchors the replacement battle's `BattleStartTime` to that cooldown's expiry (`Game.Application/Services/BattleService.cs` — `abandonedOutcomeCooldown`). But the `NewEnemy` response contract was never updated: `NewEnemyModel` has a `Cooldown` field, yet the enemy-bearing response populates only `EnemyInstance` + `ZoneId` (`Game.Api/Sockets/Commands/NewEnemy.cs`). The client presents the new enemy immediately (`battle-engine.ts` — a fresh enemy with null `elapsedOffsetMs` goes `Active` right away), so after any idle loss/draw the client's battle timeline runs ~5s ahead of the server's.

## Failure scenario (traced end-to-end)

1. Idle loss → client calls `NewEnemy` with `clientBattleMs = timeElapsed` (the idle loop never sends `BattleLost`; that command is boss-only — `enemy-manager.ts:519-548`).
2. Server books the loss, sets `EnemyCooldown = now+5s`, anchors the replacement battle at that expiry — but returns the enemy with no cooldown/offset.
3. Client fights immediately; its honest `DefeatEnemy` claim arrives ~5s − rtt **before** the server-computed completion → rejected by the `ClaimedBeforeBattleCouldFinish` anti-cheat gate (warning logged), triggering a full `resyncPlayerAndInventory`.
4. The recovery `NewEnemy`'s abandon replay (capped at wall-clock = D−5s) hasn't concluded → the already-won battle is handed back at offset D−5s and the client refights the last 5 seconds.
5. In a losing streak (idling in a too-hard zone — a core idle-game state) the cycle repeats on every loss: extra round trips, hand-backs, refights, and warning-log noise that buries genuine tamper signals. Fast kills (duration &lt; 5s) get the whole battle refought.

The server-side anchor itself is pinned by `BattleServiceIntegrationTests.cs:1119-1161`; the missing piece is purely the response contract.

## Suggested fix

Populate `NewEnemyModel.Cooldown` (or an explicit "starts in" field) when the abandon resolved an outcome, and have the client wait it out before going `Active` — mirroring the existing `DefeatEnemyResponse` cooldown/prefetch flow. Add a mirrored client test.

Related: #1851 (closed) introduced the anchor; the boss-path variant of this pacing question is tracked separately.

ginderjeremiah:
## Summary

#1851 correctly made `AbandonBattle` apply the 5s post-battle cooldown when an abandon resolves an outcome, and `BattleService.StartBattle` anchors the replacement battle's `BattleStartTime` to that cooldown's expiry. But the `NewEnemy` response contract was never updated: `NewEnemyModel` already had a `Cooldown` field, but the enemy-bearing response only ever populated `EnemyInstance` + `ZoneId`. The client presented the new enemy immediately, so after any idle loss/draw the client's battle timeline ran ~5s ahead of the server's.

1. Idle loss → client calls `NewEnemy` (the idle loop never sends `BattleLost`; that's boss-only).
2. Server books the loss, sets `EnemyCooldown = now+5s`, anchors the replacement battle at that expiry — but returned the enemy with no cooldown/offset.
3. Client fought immediately; its honest `DefeatEnemy` claim arrived ~5s − rtt **before** the server-computed completion → rejected by the `ClaimedBeforeBattleCouldFinish` anti-cheat gate (warning logged), triggering a full `resyncPlayerAndInventory`.
4. In a losing streak (idling in a too-hard zone) this repeated on every loss: extra round trips, hand-backs, refights, and warning-log noise burying genuine tamper signals.

## Changes

- **`Game.Api/Sockets/Commands/NewEnemy.cs`**: populates `NewEnemyModel.Cooldown` whenever `PlayerState.EnemyCooldown` is still in the future after `StartBattle` returns — which is exactly (and only) when this call's own abandon just incurred the post-battle cooldown.
- **`UI/src/lib/engine/battle/enemy-manager.ts`**: when a `NewEnemy` response bundles both an enemy and a positive cooldown, the fetch loop now waits it out via `battleEngine.startLoading` (the same mechanism the `DefeatEnemy`/`BattleLost` cooldown/prefetch flow already uses) before presenting the enemy, re-checking the stop/supersession guards after the wait.
- **`docs/backend-battle.md`**: closes the gap between the documented server-side anchor and what the response contract now actually carries.

## Tests

- `Game.Api.Tests/Integration/NewEnemySocketTests.cs`: new integration test — starts a battle, backdates it so the next `NewEnemy`'s abandon resolves an outcome, and asserts the response carries both the fresh `EnemyInstance` and a positive `Cooldown` bounded by the 5s post-battle cooldown.
- `UI/src/tests/lib/engine/battle/enemy-manager.test.ts`: new cases pinning that a bundled positive cooldown is waited out via `battleEngine.startLoading` before the enemy is presented, and that a cooldown-free response (a genuinely fresh spawn) is presented immediately with no wait.

## Test plan

- [x] `dotnet test` — full backend suite passes: 3081/3081.
- [x] `npm run lint` — clean.
- [x] `npm run test:unit:ci` — full suite passes: 3626/3626.

Closes #1881

---
_Generated by [Claude Code](https://claude.ai/code/session_01DXipx85xKZykwgKGjJUQ3f)_


Game.Api.Tests/Integration/NewEnemySocketTests.cs
```diff
@@ -103,5 +114,47 @@ public async Task NewEnemy_OnCooldown_ReturnsCooldown()
...
+            // loops NewEnemy (exactly what an idle loss/draw does) must get the just-incurred cooldown on the
+            // SAME response as the fresh enemy, not a fresh enemy silently anchored ahead of it.
+            var response = await socketClient.SendCommandAsync<NewEnemyModel>(
+                "NewEnemy", new { NewZoneId = (int?)null });
+
+            Assert.Null(response.Error);
+            Assert.NotNull(response.Data);
+            Assert.NotNull(response.Data.EnemyInstance);
+            Assert.NotNull(response.Data.Cooldown);
+            Assert.InRange(response.Data.Cooldown!.Value, 1, 5000);
```
ginderjeremiah: Nit: the `!` null-forgiving operator here trips CLAUDE.md's "avoid `!` as much as possible" guideline. The sibling `NewEnemy_OnCooldown_ReturnsCooldown` test just above (lines 114-115) already demonstrates the codebase's preferred non-`!` shape after a `NotNull` assertion: `Assert.True(response.Data.Cooldown > 0)`. You could keep the tighter upper bound without `!` via e.g. `Assert.True(response.Data.Cooldown is > 0 and <= 5000)`. Purely cosmetic — not blocking.

---

ginderjeremiah (commented):
## Review: NewEnemy response carries the abandon-incurred cooldown anchor (#1881)

**Verdict:** Approve _(posting as `COMMENT` — GitHub won't accept a gating verdict from the PR author; the intended call is **Approve**)_ — the fix is correctly scoped, closes the exact contract gap the issue traced, and lands the wait on both sides with mirrored tests.

The change does exactly what #1881 asks for: it populates `NewEnemyModel.Cooldown` on the enemy-bearing response only when this call's own abandon incurred the post-battle cooldown, and has the client wait it out before presenting the enemy — mirroring the existing `DefeatEnemy`/`BattleLost` cooldown flow instead of inventing a new mechanism.

### Correctness

- The backend gate `state.EnemyCooldown > afterStart` is a sound proxy for "this call's abandon resolved an outcome": the `IsOnCooldown(now)` early-return at line 25 guarantees any pre-existing cooldown was already `<= now <= afterStart`, so the only way `EnemyCooldown` is in the future here is `StartBattle`'s `abandonedOutcomeCooldown` having just set it. No false positives from a stale cooldown.
- Reading a fresh `afterStart` after `StartBattle`/save (rather than reusing `now`) correctly returns the *remaining* cooldown, and naturally collapses to `null` if processing already consumed it — so the client never waits out time that has already elapsed server-side.
- Frontend: `if (result.data.cooldown)` truthiness lines up with the backend returning `null`/omitting it; the post-wait re-check of `!this.started || generation !== this.fetchGeneration` correctly prevents a resolved wait from clobbering a fight a supersession moved on to, consistent with the other guards in this loop.

### Test coverage

- Backend: `NewEnemy_AbandonResolvesOutcome_BundlesTheIncurredCooldownWithTheReplacementEnemy` backdates the in-flight battle so the next `NewEnemy`'s abandon resolves an outcome, and asserts the response carries both the fresh `EnemyInstance` and a positive cooldown bounded by 5s — the right end-to-end shape.
- Frontend: two mirrored cases pin that a bundled positive cooldown is waited out via `battleEngine.startLoading` before presenting, and that a cooldown-free (genuinely fresh) spawn presents immediately with no wait. Good coverage of both branches. This is response-contract/pacing, not simulation math, so no cross-suite parity vector is required.

### Nits

- `NewEnemySocketTests.cs:157` — the `!` null-forgiving operator trips CLAUDE.md's "avoid `!`" guideline; the sibling test just above shows the preferred non-`!` shape. Cosmetic (inline comment left).

### What's good

- Tight, minimal change that reuses the existing cooldown-wait mechanism on both sides rather than adding a parallel one, and the `docs/backend-battle.md` update closes the doc/contract gap it was describing. The code comments trace the *why* (anchor vs. clock skew) clearly without over-explaining.

---

ginderjeremiah:
## Review: Approve ✅

Verdict: **approve (mergeable-clean)**. CI is fully green on head `12c386f` (all 9 checks: backend, frontend, e2e chromium/firefox/webkit, verify-*, all-checks-passed).

Reviewed the diff against the project standards and the battle docs. This correctly closes the contract gap #1881 identified, and the reasoning holds up under scrutiny.

### Correctness — the cooldown condition is sound
The concern worth verifying was whether populating `Cooldown` on `state.EnemyCooldown > afterStart` could mis-fire on a genuinely fresh spawn. It can't:

- `NewEnemy.cs` gates on `state.IsOnCooldown(now)` at the top and returns a cooldown-only response (no enemy) when a cooldown is active. So the enemy-bearing branch is only reached when `EnemyCooldown ≤ now` held at entry.
- Inside `StartBattle`, the only thing that can push `EnemyCooldown` into the future is *this call's own* `AbandonBattle` resolving a win/loss/draw (`abandonedOutcomeCooldown`, #1851) — confirmed by the invariant already documented at `BattleService.cs:79-85`. A fresh spawn has no active battle to abandon, so `EnemyCooldown` stays in the past → `Cooldown = null`.
- A still-in-progress hand-back (#1595) sets no cooldown, and a `forceAbandon` discard resolves no outcome — both correctly yield `null`. The comment on the field accurately describes this.

Capturing `afterStart` *after* `StartBattle` (rather than reusing the top-of-handler `now`) is the right call — it measures the remaining cooldown from the current moment, net of any time `StartBattle` spent.

### Battle-parity
No parity surface is touched. This is pure pacing/timing — the client waits out the cooldown before going `Active`, which *aligns* its battle clock with the server's anchor rather than changing any tick-by-tick simulation. It actively helps the `ClaimedBeforeBattleCouldFinish` gate rather than risking divergence, so no mirrored parity test is warranted.

### Frontend
The `startLoading(cooldown)` wait sits inside `runFetchLoop` on the actual `NewEnemy`-response path, and correctly re-checks the stop/supersession guards (`!this.started || generation !== this.fetchGeneration`) after the await — matching the established pattern in this file. No double-wait risk: the prefetch-present branch returns before the send, and the idle-loss path never runs the victory-branch `startLoading`, so exactly one cooldown wait occurs.

### Tests
Backend integration test (backdate → abandon resolves outcome → assert enemy + `0 < Cooldown ≤ 5000`) and two frontend unit tests (waits out a bundled positive cooldown; presents immediately with `cooldown: 0`) cover both branches well. Doc update to `backend-battle.md` closes the contract-vs-anchor gap accurately.

### Nits (non-blocking, optional)
- `NewEnemy.cs`: the multi-line comment sits above `var afterStart = DateTime.UtcNow;` but really explains the `Cooldown = …` expression two lines below. Accurate, just slightly ahead of what it annotates.
- `enemy-manager.ts`: the new guard uses a truthy `if (result.data.cooldown)` while the sibling cooldown checks use explicit `cooldown > 0`. Truthy is correct here (the field is optional `number | null`), just a hair less consistent with its neighbors.

No out-of-scope follow-ups needed — the boss-path pacing variant is already noted as separately tracked in #1881.

Recommendation: **mergeable-clean.**

---

PR #1903, opened 2026-07-12
Split CodexView's reactive layer into per-tab view classes

### LINKED ISSUE #1737: Split CodexView's 820-line reactive class per tab
## Problem

`UI/src/routes/game/screens/codex/codex-view.svelte.ts:261-820` is the largest file under `routes/` by a wide margin: one class holds the enemies/zones/skills tabs' selection state, ~30 deriveds, and the deep-link constructor.

The pure helpers are already split out (`codex-display`, `enemy-level`, `skill-provenance`), and the screen already renders per-tab components (`EnemiesTab`/`ZonesTab`/`SkillsTab`) — so splitting the reactive layer per tab to mirror that structure is mechanical. Each tab's derivations become independently readable/testable, and a change to one tab stops touching the others' file.

No behavior change; the existing Codex tests pin the seams. Fine to do opportunistically the next time the Codex is touched rather than as standalone churn.

ginderjeremiah:
## Summary

`UI/src/routes/game/screens/codex/codex-view.svelte.ts` had grown into one 820-line class holding the Enemies/Zones/Skills tabs' selection state, ~30 deriveds and the deep-link constructor together. The screen already renders per-tab components (`EnemiesTab`/`ZonesTab`/`SkillsTab`), so this splits the reactive layer to mirror that structure.

## Changes

- **New `enemies-tab-view.svelte.ts`, `zones-tab-view.svelte.ts`, `skills-tab-view.svelte.ts`**: each holds the `$state`/`$derived` layer and handler methods for its own tab (table filter/search/sort, selected-entity dossier projections, cross-tab-agnostic helpers), plus the VM interfaces that tab's projections produce.
- **`codex-view.svelte.ts`**: now a thin orchestrator (373 lines, down from 820). It owns only what's genuinely cross-tab — the active tab, the once-fetched player statistics + challenge-error state, the shared `StatisticsData` query engine, and the cross-dossier links (`openEnemy`/`openZone`/`openSkill`) — and delegates every other property/method to the owning tab's instance via forwarding getters/setters. Its public API is unchanged, so every consuming `.svelte` component keeps reading `view.<x>` exactly as before.
- **`codex-display.ts`**: gained the shared catalogue filters (`liveEnemies`/`liveZones`/`liveSkills`), moved out of `codex-view.svelte.ts` so all three tab views can reuse them without an inter-tab import.

No behavior change — the existing `codex-view.svelte.test.ts` suite (unmodified) pins the seams, and it (along with every other Codex test) still passes as-is against the new internal structure.

## Test plan

- [x] `npm run lint` — clean.
- [x] `npm run test:unit:ci` — full suite passes: 3624/3624, coverage thresholds intact.

Closes #1737

---
_Generated by [Claude Code](https://claude.ai/code/session_01X8CAn7dUdrxa11qtY8Sdys)_

### DIFF
--- a/UI/src/routes/game/screens/codex/StatisticsPanel.svelte
+++ b/UI/src/routes/game/screens/codex/StatisticsPanel.svelte
@@ -24,7 +24,7 @@
 </div>
 
 <script lang="ts">
-import type { EntityStatVM } from './codex-view.svelte';
+import type { EntityStatVM } from './codex-display';
 
 interface Props {
 	/** The statistics referencing this entity (label + formatted value). */
--- a/UI/src/routes/game/screens/codex/codex-display.ts
+++ b/UI/src/routes/game/screens/codex/codex-display.ts
@@ -19,6 +19,13 @@ export type EnemySort = 'level' | 'name';
 /** A zone's progression status on the Zones rail. */
 export type ZoneStatus = 'cleared' | 'unlocked' | 'locked';
 
+/** A single per-entity statistic row shown in a dossier's "Your record" section
+ *  (shared by the enemy, zone and skill dossiers). */
+export interface EntityStatVM {
+	label: string;
+	value: string;
+}
+
 export const CODEX_TABS: CodexTab[] = ['enemies', 'zones', 'skills'];
 
 export const ENEMY_FILTERS: { key: EnemyFilter; label: string }[] = [
--- a/UI/src/routes/game/screens/codex/codex-display.ts
+++ b/UI/src/routes/game/screens/codex/codex-display.ts
@@ -3,6 +3,7 @@
    restyles with the theme. Kept store-free (no reactive state) so the view-model and its tests import
    the types/formatters without a cycle. */
 
+import type { IEnemy, ISkill, IZone } from '$lib/api';
 import { formatNum } from '$lib/common';
 import type { LevelRange } from './enemy-level';
 import type { SkillAcquisitionStatus } from './skill-provenance';
@@ -119,3 +120,16 @@ export function sortEnemyRows(sort: EnemySort): (a: EnemySearchSortFields, b: En
 			return (a, b) => a.level - b.level || a.name.localeCompare(b.name);
 	}
 }
+
+/* ── catalogue helpers (shared by every tab's reactive layer) — retirement keeps a slot resolvable
+   but out of the glossary. ── */
+
+/** Live (non-retired) enemies. */
+export const liveEnemies = (enemies: IEnemy[] | undefined): IEnemy[] => (enemies ?? []).filter((e) => !e.retiredAt);
+
+/** Live (non-retired) zones in authored progression order. */
+export const liveZones = (zones: IZone[] | undefined): IZone[] =>
+	(zones ?? []).filter((z) => !z.retiredAt).sort((a, b) => a.order - b.order);
+
+/** Live (non-retired) skills in catalogue order. */
+export const liveSkills = (skills: ISkill[] | undefined): ISkill[] => (skills ?? []).filter((s) => !s.retiredAt);
--- a/UI/src/routes/game/screens/codex/codex-view.svelte.ts
+++ b/UI/src/routes/game/screens/codex/codex-view.svelte.ts
@@ -35,6 +35,7 @@ import {
 	type EnemyFilter,
 	type EnemySort,
 	type EnemySubTab,
+	type EntityStatVM,
 	tabAccent,
 	tabLabel
 } from './codex-display';
@@ -90,13 +91,6 @@ export interface CodexTabVM {
 	count: number;
 	accent: string;
 	active: boolean;
-}
-
-/** A single per-entity statistic row shown in a dossier's "Your record" section
- *  (shared by the enemy, zone and skill dossiers). */
-export interface EntityStatVM {
-	label: string;
-	value: string;
 }
 
 /* ── reactive view-model ──────────────────────────────────────────────────── */
--- a/UI/src/routes/game/screens/codex/codex-view.svelte.ts
+++ b/UI/src/routes/game/screens/codex/codex-view.svelte.ts
@@ -19,55 +19,61 @@
    section, and the Statistics screen deep-links an entity into the matching dossier instead of
    rendering its own in-place one.
 
-   The view-model only wires reactive state to the pure helpers; the projection maths live in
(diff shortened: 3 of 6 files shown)


ginderjeremiah (commented):
## Review: Split CodexView's 820-line reactive class into per-tab view classes (#1737)

**Verdict:** Approve _(posted as a comment — same account as the PR author, so GitHub won't take a gating event; treat this as an approve)_ — a faithful, well-executed structural split with no behavior change, unchanged public API, and green CI.

I verified the split line-by-line against the pre-split `codex-view.svelte.ts`: every `$derived`/handler landed in the right tab class with identical logic, the constructor ordering (tab → enemy+level → zone → skill seeds) is preserved, and the orchestrator's forwarding getters/setters reproduce the old public surface exactly.

### Things I specifically checked and confirmed

- **Cross-object reactivity of the shared stats projection is preserved.** `CodexView` keeps the single once-fetched `stats`/`statData` and hands each tab a `projectStats` closure. Each tab's `statistics`/`zoneStatistics`/`skillStatistics` derived invokes that closure synchronously, which reads `codexView.statData` → `stats`; Svelte 5's dynamic signal tracking captures the dependency across the object boundary, so a mount-time `stats` update still invalidates the tab deriveds. This is the one subtle way a split like this could silently break, and it's correct.
- **Public API unchanged.** Every consuming `.svelte` component imports only the `CodexView` type (plus `CodexNavPayload`/`EntityStatVM`) and reads `view.<x>` structurally — none import the per-VM types by name — so relocating those VM interfaces into the tab files breaks nothing. `EnemyRowVM`/`ZoneProjectionVM` are re-exported for the (unmodified) test suite.
- **Getter-only exposure of `sub`/`level`/`scaling`/`filter` is safe** — no component assigns them directly (all mutated via handlers), corroborated by svelte-check passing.
- **No dead code / no unused imports** left behind; `liveEnemies`/`liveZones`/`liveSkills` moved cleanly to `codex-display.ts` and every VM type imported into the orchestrator is used.

- `EntityStatVM` (`codex-view.svelte.ts:97`) is a shared VM consumed by all three tab files, which now type-import it back from the orchestrator (`enemies/zones/skills-tab-view.svelte.ts`). Since `codex-display.ts` is explicitly the store-free module "so the view-model and its tests import the types/formatters without a cycle," that's arguably its natural home — moving it there would make the module graph a clean DAG (tabs → codex-display; orchestrator → tabs + codex-display) instead of a type-only back-edge from the tabs into the orchestrator. Purely a tidiness call; the current `import type` is erased at runtime so there's no real cycle.
- The three tab classes each recompute `liveEnemies(staticData.enemies)` in their own `enemies` derived (vs. the one shared `this.enemies` before). This is an intentional, correct consequence of decoupling the tabs (the PR deliberately avoids an inter-tab import), and reference data changes rarely — noting it only so it's a conscious trade, not an oversight.

### Test coverage

Appropriate for a pure structural refactor: the existing `codex-view.svelte.test.ts` (877 lines, **unmodified**) drives the unchanged public surface and still passes, which is exactly what pins the seams here. No new tests are warranted — the behavior contract didn't move. If the tab classes grow independent logic later, that'd be the moment to add per-tab suites; not needed now.

### What's good

- The orchestrator/tab decomposition mirrors the existing `EnemiesTab`/`ZonesTab`/`SkillsTab` component split, which is the right seam. 820 → 373 lines in the orchestrator, each tab independently readable.
- The dependency-injection of `projectStats` keeps the single statistics fetch centralized while letting each tab project its own entity — cleaner than duplicating the fetch or leaking `statData` into every tab.
- Doc comments carried over accurately and the file-level header was updated to describe the new composition.


---

ginderjeremiah:
## Review — Split CodexView's reactive layer into per-tab view classes

**Verdict: Approve (with two non-blocking notes).** Closes #1737 exactly as scoped — a mechanical, behavior-preserving split.

### What I verified
- **`codex-view.svelte.ts` is genuinely a thin orchestrator.** It owns only the cross-tab concerns (active `tab`, the once-fetched `stats`/`statsLoading`/`statsError`/`challengesError`, the shared `StatisticsData` query engine, and the `openEnemy`/`openZone`/`openSkill` cross-links). Everything else forwards to the owning `EnemiesTabView`/`ZonesTabView`/`SkillsTabView` instance.
- **Public class API is preserved.** I diffed the pre-split class's public surface (getters/setters/methods/`$state` fields) against the new one member-for-member — every getter, setter and method is present, and the projection VM types (`EnemyRowVM`, `ZoneProjectionVM`, etc.) are re-exported from the same module path so importers don't reach into the per-tab files.
- **Seams are pinned by an unmodified test.** `codex-view.svelte.test.ts` is byte-for-byte unchanged vs. the PR base (not in the diff), still imports `CodexView` + the re-exported types, and drives the delegated members (`.tabs`, `.enemyRows`, `.selectedEnemy`, `.zoneRows`, the `openEnemy/openZone/openSkill` cross-links assert `view.tab` flips and the right entity selects). It passing against the restructured internals is the strongest possible "no behavior change" signal.
- **No dead code / unused imports left behind.** Every imported VM type is consumed as a return annotation; `verify-frontend` (svelte-check, 0 warnings) would have caught otherwise.
- **The shared catalogue filters** (`liveEnemies`/`liveZones`/`liveSkills`) landing in `codex-display.ts` is the right call — keeps the three tab views free of inter-tab imports while staying DRY.

### CI
Green on head `91a30e1`: `all-checks-passed`, `test-frontend`, `verify-frontend`, and all three `test-e2e` browsers succeeded. Backend jobs correctly skipped (frontend-only; `detect-changes` gated them). (The pending legacy commit-status is just the empty statuses API, not a real check.)

### Non-blocking notes (no action needed)
1. **`level` / `filter` / `sub` / `scaling` narrowed from writable `$state` to read-only forwarding getters.** The description says "public API unchanged"; strictly these four went from assignable to read-only. It's harmless and arguably an improvement — every write already goes through `setLevel`/`setFilter`/`selectSub`/`toggleScaling`, and no component or test assigns them directly, so it forces the intended handler path. Worth a mention only for accuracy.
2. **Minor per-tab duplication** — each tab view re-declares the `statVMsFor` injection and (zones/skills) a private `enemies = $derived(liveEnemies(...))`. This is deliberate given the "no inter-tab import" goal and not worth collapsing.

Neither warrants a follow-up issue; the split is complete and clean. Not merging — leaving that to the coordinator.

---

Issue #1856, opened 2026-07-11
Workbench positional new-id resolution can write child collections onto another admin's record under concurrent sessions


ginderjeremiah:
## Problem

`resolveNewIds` (`UI/src/routes/admin/workbench/save-helpers.ts:44-59`, consumed by `entity-store.svelte.ts`):

```ts
const newlyPersisted = fresh.filter((record) => !existing.has(record.id)).sort((a, b) => a.id - b.id);
added.forEach((record, index) => { const persisted = newlyPersisted[index]; ... });
```

The "k-th add → k-th lowest unseen id" mapping assumes only this admin's adds appear as new ids in the refetch. Another admin's record saved between this session's diff and refetch shows up as an unexpected "new" id and, if it sorts lower, this session's child savers **write their portions/effects/modifiers onto the other admin's record**.

## Failure scenario

Two admins authoring skills simultaneously: A's effects land on B's freshly-created skill. Genuine cross-record data corruption, not just a UI mislabel. Low likelihood today (internal tool, requires two-admin concurrency), but the failure mode is silent.

## Suggested fix

Have the admin Add endpoints return the persisted ids in send order and map from the response, instead of inferring positionally from a full refetch. (This needs a small backend contract change on the admin save endpoints; the positional inference and its `?? record.id` fallback — see the companion issue on unresolved add ids — can then both be deleted.)

<sub>Found during a scheduled codebase health review.</sub>

ginderjeremiah:
Looked at picking this up and want to flag scope before diving in, rather than surprise you with a much bigger PR than the `scope: medium` label suggests.

**The real blast radius:** the primary Add/Edit/Delete batch for *every* admin entity goes through the shared `ChangeSetProcessor.Apply` (`Game.DataAccess/Repositories/Admin/ChangeSetProcessor.cs`), whose `add` handler is `Action<T>` — it only stages the insert via `IEntityStore.Insert(entity)`. The actual DB commit (and EF's identity-column assignment) happens **after** the controller action returns, inside `CommitFilter` (an `IAsyncActionFilter` that runs post-`next()`). So at the point where a repo method would build its `AdminSaveResult` response, the newly-inserted entities' real ids don't exist yet.

[...]

PR #1904, opened 2026-07-12
Deduct the settled stale battle's span from the offline-progress away budget

### LINKED ISSUE #1882: Offline/switch credit double-counts the settled stale battle's span — away budget is captured before the settle and never reduced
## Problem

`OfflineProgressService.SimulateProgress` captures `awayMs = now − player.LastActivity` up front, then `ResolveStaleBattle` settles the disconnected in-flight battle **in full** (a win pays exp/kills/challenges via `RecordVictory`), and the simulator then replays the **entire** `[LastActivity, now]` window as fresh battles with `AwayBudgetMs = awayMs` unreduced (`Game.Application/Services/OfflineProgressService.cs` → `Game.Core/Battle/Offline/OfflineProgressSimulator.cs`).

The settled battle's span `[battleStart, battleStart + TotalMs]` lies inside that window, so the same wall-clock time is credited twice — up to `DefaultMaxBattleMs` (2 min) + cooldown per occurrence.

## Impact

- **Login path**: a disconnect/reconnect cycle every ~5 minutes inflates progress ~30–40%.
- **Switch path**: `SimulateSwitchProgress` has no away floor, so deliberate A↔B character-switch cycling approaches ~2× the legitimate idle rate (each switch credits the departed character's in-flight battle *and* re-simulates its span).

The existing tests (`OfflineProgressServiceIntegrationTests.cs:285-314, 477-500`) assert only that the stale battle is settled — none pins total credited time, so the overlap goes unnoticed rather than being deliberate.

## Suggested fix

Deduct the settled battle's span from the away budget — e.g. `max(0, settledBattleCompletion − windowStart)` — or re-derive `awayMs` from the settle's own `StampActivity` before building the simulation parameters. Add a test pinning total credited time across settle + simulation.

Related: #1874 covers the crash-between-credit-and-save idempotency gap; #1859 the settle's notification pushes — neither touches budget accounting.

ginderjeremiah:
## Summary

`OfflineProgressService.SimulateProgress` captured the away-time budget (`awayMs = now − player.LastActivity`) *before* settling a disconnected in-flight battle via `BattleService.ResolveStaleBattle`, then handed that same unreduced budget to the offline simulator. The settled battle's own credited span (its duration plus the 5s post-battle cooldown) falls inside `[LastActivity, now]`, so it was replayed a second time as free simulated battles — once via the settle's own reward, again via the simulator treating that same wall-clock time as available for fresh battles.

- **Login path** (`SimulateOfflineProgress`, 5-minute floor): a disconnect/reconnect cycle where a battle happened to be in flight inflates progress by up to one settled battle + its cooldown.
- **Switch path** (`SimulateSwitchProgress`, no floor at all): worse, since a deliberate A/B character-switch cycle can repeat the over-credit on every switch.

## Changes

- **`Game.Application/Services/BattleService.cs`**: `ResolveStaleBattle` (and the private `AbandonBattle` it wraps) now return a `StaleBattleResolution(BattleStartResult? Handoff, int? SettledBattleMs)` instead of a bare nullable handoff. `SettledBattleMs` carries the resolved battle's own credited duration (`BattleResult.TotalMs`) when an outcome was actually recorded — null for a still-in-progress handoff or when there was nothing to resolve. The two other `AbandonBattle` call sites (`StartBattle`, `StartBossBattle`) don't need this new field and discard it.
- **`Game.Application/Services/OfflineProgressService.cs`**: after a settle (not a handoff), deducts `SettledBattleMs + PostBattleCooldown` from the away budget passed to `OfflineProgressSimulator`, clamped at 0.
- **`docs/backend-battle.md`**: documents the deduction in the existing "stale in-flight battle" bullet.

## Tests

- `Game.Application.Tests/Services/BattleServiceIntegrationTests.cs`: updated the four existing `ResolveStaleBattle` tests for the new return shape, with new assertions pinning `SettledBattleMs` for each branch (credited win, genuine draw at the cap, still-in-progress handoff).
- `Game.Application.Tests/Services/OfflineProgressServiceIntegrationTests.cs`: two new deterministic tests (login path and switch path) that size the away window as exactly the settled battle's span plus N whole battle+cooldown cycles, and assert the simulator wins exactly N — not N+1, which is what re-simulating the settled battle's own span as extra free time would produce.

## Test plan

- [x] `dotnet test` — full backend suite passes: 3082/3082.

Closes #1882

## Note for reviewers

The oldest unclaimed `claude`+`bug` issue (#1856) was skipped for this pass — a prior session already investigated it and flagged it as needing a scoping conversation (a materially larger architectural change than its `scope: medium` label suggests), so I picked up the next-oldest one instead.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MN8vDLoTGbK68yfJwPLbrY)_


ginderjeremiah (commented):
## Review: Deduct the settled stale battle's span from the offline-progress away budget (#1882)

**Verdict:** Approve — the double-count is fixed at the right seam, the deduction model is correct, and the tests pin the exact N-vs-N+1 boundary that would otherwise regress. (Submitted as a `COMMENT` only because GitHub won't accept a gating verdict from the PR author; treat this as an approval.)

### Assessment

The fix is well-targeted. `awayMs` is captured before the settle, and previously handed unreduced to the simulator, so the settled battle's own span was replayed a second time as free simulated battles. Returning `StaleBattleResolution(Handoff, SettledBattleMs)` and deducting `SettledBattleMs + PostBattleCooldown` (clamped at 0) from the simulator budget is the clean way to remove exactly the one battle-slot the settle already credited.

I checked the accounting is sound across every branch of `AbandonBattle`:
- **Won / lost / genuine-draw-at-cap** all fall through to `return new StaleBattleResolution(null, result.TotalMs)`, so the deducted span matches the actual credited duration in each case.
- **Still-in-progress handoff** and **nothing-to-resolve** both return `SettledBattleMs = null`, so no deduction happens — correct, since neither credited a battle.
- Because the simulator budget is position-independent (a total time it divides into `battleMs + cooldown` slots), removing one settled slot is exactly right regardless of where `battleStart` sits relative to `LastActivity` — no over/under-deduction from the window offset.

The two other `AbandonBattle` call sites (`StartBattle`, `StartBossBattle`) correctly discard the new field, and the XML doc on `ResolveStaleBattle`/`StaleBattleResolution` documents the caller's deduction responsibility clearly.

### Test coverage

Strong and exactly on-point:
- The two new `OfflineProgressServiceIntegrationTests` cases size the away window as the settled span plus N whole cycles and assert the simulator wins **exactly N, not N+1** — the precise assertion the issue called out as missing (the old tests only checked that the stale battle settled, never total credited time). The switch-path case correctly exercises the no-floor path where the abuse is worse.
- The four updated `BattleServiceIntegrationTests` pin `SettledBattleMs` per branch (credited win in range, genuine draw at exactly the cap, null for the handoff).
- Backend-only change; offline simulation is parity-irrelevant (client never replays it), so no frontend mirror is needed — correctly noted.

### Nits

None. Clean, well-scoped, well-documented.

Holding the merge only until `test-backend` (still in progress) reports green.

---

PR #1905, opened 2026-07-12
Omit clientBattleMs on the fight loop's first NewEnemy request

### LINKED ISSUE #1883: Sub-5-minute reconnect silently discards the in-flight battle — first NewEnemy reports clientBattleMs 0, so the documented hand-back is unreachable
## Problem

On a reconnect under the 5-minute offline floor, the welcome-back path deliberately "leaves the stale battle for the idle loop's first StartBattle to abandon" (`OfflineProgressService.cs`, sub-threshold branch). But the client's first `getNewEnemy` after a page load always sends a **numeric** `clientBattleMs` — `clientBattleMs = prepared ? 0 : battleEngine.timeElapsed` (`UI/src/lib/engine/battle/enemy-manager.ts:251`), and a fresh engine's `timeElapsed` is `0`.

Server-side, `AbandonBattle` defines a numeric `0` as "the client never fought this battle": `elapsedMs = min(max(0, 0), wallClock) = 0` → `ClearBattle()` with **no outcome, no credit, and no #1595 hand-back** (`Game.Application/Services/BattleService.cs:450-466`). The wall-clock fallback replay only runs when `clientBattleMs` is **null**.

So the in-flight battle — possibly one the wall-clock replay would resolve as a win — is silently dropped. This:

- contradicts the ≥5-minute path, which credits the same battle via `ResolveStaleBattle`'s null-`clientBattleMs` wall-clock replay;
- makes the client's own comment at `enemy-manager.ts:278-281` ("a still-in-progress battle handed back on this NewEnemy, e.g. a sub-5-minute reconnect") unreachable on this path — with a numeric `0` the server never hands anything back;
- loses honest players up to one battle's outcome per quick refresh/reconnect.

## Suggested fix

Omit `clientBattleMs` (send null) on the session's **first** fetch after a page load — the client has no history for whatever battle the server holds, which is exactly the semantics the server's wall-clock fallback exists for. That lets the server settle a concluded battle (crediting it) or hand back a still-in-progress one (#1595). Pin with a mirrored test.

Related: #1881 (same request/response field contract, different failure mode).

ginderjeremiah:
## Summary

On a reconnect under the 5-minute offline floor, the welcome-back path deliberately leaves the stale battle for the idle loop's first `NewEnemy` to abandon (`OfflineProgressService.cs`, sub-threshold branch) — relying on the backend's wall-clock fallback (a `null` `clientBattleMs`) to settle it correctly.

But the client's first `getNewEnemy` after a page load always sent a **numeric** `clientBattleMs`: `clientBattleMs = prepared ? 0 : battleEngine.timeElapsed`, and a freshly-constructed `battleEngine` has never held a battle, so its `timeElapsed` is an untouched `0`. Server-side, `AbandonBattle` treats a numeric `0` as "the client never fought this battle" and clears it with **no outcome, no credit, and no #1595 hand-back** — silently dropping an in-flight battle (possibly a win) on every quick refresh/reconnect.

## Fix

`EnemyManager` now tracks whether the fight loop has sent its first `NewEnemy` request since `start()` (`isFirstFetch`, re-armed on every `start()`). On that first request, `clientBattleMs` is omitted entirely (`undefined`) instead of sending the engine's meaningless untouched `timeElapsed`, letting the backend's existing wall-clock fallback settle a concluded battle or hand back one still in progress. Every subsequent fetch in the session (idle loss/draw, etc.) still reports the genuine elapsed time as before.

This is a frontend-only change — the backend's `null`-`clientBattleMs` handling (`AbandonBattle`) was already correct; the bug was purely that the frontend never sent `null`/omitted the field on its very first request.

## Tests

- `enemy-manager.test.ts`: new cases pinning that the loop's first `NewEnemy` request omits `clientBattleMs` even when `battleEngine.timeElapsed` is genuinely `0`, that a later fetch in the same session reports the real elapsed time, and that the omission re-arms on a fresh `start()` after a `stop()` (e.g. a character switch) rather than being consumed once for the manager's whole lifetime.
- Updated the shared `battleEngine` mock to expose a real (mutable) `timeElapsed` instead of leaving it `undefined`, which also strengthens several pre-existing assertions that happened to pass only because the mock never set the field.

## Test plan

- [x] `npm run lint` — clean.
- [x] `npm run test:unit:ci` — full suite passes: 3627/3627.

Closes #1883

---
_Generated by [Claude Code](https://claude.ai/code/session_01UTwMj529Rva42tTMPX8Au6)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -104,6 +104,12 @@ export class EnemyManager {
 	 *  redundant re-syncs (e.g. a returnToIdle when the persisted mode is already idle). `undefined` until
 	 *  the first sync, so the first authoritative state is always sent. */
 	private lastSyncedAutoChallengeBoss?: boolean;
+	/** Whether the fight loop has not yet sent a `NewEnemy` request since {@link start}. Re-armed by every
+	 *  `start()` and consumed by the first `runFetchLoop` dispatch, regardless of whether that `start()` was
+	 *  handed an `activeBattle` — a fresh `battleEngine` has no history for whatever battle the server may
+	 *  still be holding (e.g. a sub-5-minute reconnect the welcome-back gate didn't resume directly), so its
+	 *  untouched `timeElapsed` of 0 must not be reported as a genuine "never fought" claim (#1883). */
+	private isFirstFetch = true;
 
 	/**
 	 * Starts the fight loop. `activeBattle` is the server-handed-back battle a `GetOfflineProgress` summary
@@ -119,6 +125,7 @@ export class EnemyManager {
 	public start(activeBattle?: IEnemyInstance) {
 		if (!this.started) {
 			this.started = true;
+			this.isFirstFetch = true;
 			this.battleStageUnhook = onBattleStageChanged((stage) => this.watchBattleStage(stage));
 			if (activeBattle) {
 				this.mode = activeBattle.isBossBattle ? 'boss' : 'idle';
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Omit `clientBattleMs` on the fight loop's first `NewEnemy` request (#1883)

**Verdict:** Approve — the fix is correct, minimal, and well-targeted; it makes the sub-5-minute reconnect hand-back path reachable exactly as the issue describes.

_(Submitted as a `COMMENT` because the reviewing identity is the PR author, so GitHub won't accept a gating `APPROVE`; the real verdict is Approve.)_

### Verification

I confirmed the backend contract the fix depends on rather than taking it on faith:

- `NewEnemyRequest.ClientBattleMs` is `int?`, and `NewEnemy` passes it straight through to `StartBattle`/`AbandonBattle`. An omitted JSON key deserializes to `null`.
- In `BattleService.AbandonBattle`, `clientBattleMs is int reported ? Min(Max(reported,0), wallClock) : wallClock`. A numeric `0` collapses to `elapsedMs <= 0` → `ClearBattle()` with no outcome/credit/hand-back; `null` falls back to the wall-clock replay that can settle or hand back the battle.

So switching the loop's first fetch from a meaningless `0` to an omitted field (→ `null`) is precisely what routes it to the wall-clock fallback. Correct.

### Design

- Consuming `isFirstFetch` at line 263–265 sits *after* the Home-zone (238) and prepared-battle (248) early returns, and the stretch from there to the `sendSocketCommand` at 273 is synchronous — so the flag is only ever consumed when a real `NewEnemy` actually goes out, and a Home-first dispatch correctly leaves it armed for the next real fetch.
- Re-arming on every `start()` (line 128) rather than once per instance is the right call for character switches, and the added test pins it.

Note (no action needed): when `start(activeBattle)` hands back a battle (the ≥5-min path), the client fights it, and the *first* post-handback fetch will now also omit `clientBattleMs` even though `battleEngine.timeElapsed` is genuinely meaningful there. That's harmless — the server wall-clock-replays the battle it still holds, and for a just-resumed battle wall-clock ≈ fought time, so there's no data loss; it's a defensible uniform "first fetch can't authoritatively bound what the server holds" policy. Flagging only so it's a conscious choice, not an oversight.

### Test coverage

Solid and appropriately scoped:
- First fetch omits `clientBattleMs` even when `timeElapsed` is a genuine `0`.
- A later same-session fetch reports the real elapsed time.
- The omission re-arms on a fresh `start()` after `stop()`.
- Upgrading the shared `battleEngine` mock to a real mutable `timeElapsed` also hardens pre-existing assertions that previously passed only because the field was `undefined`.

No mirrored backend test is required — the backend's `null`-`clientBattleMs` handling is unchanged and already covered.

### What's good

Tight, surgical frontend-only fix with a precise root-cause comment and tests that pin the exact 0-vs-omitted distinction the bug turned on.


---

PR #1906, opened 2026-07-12
Anchor ChallengeBoss's replacement battle to the abandon-incurred cooldown

### LINKED ISSUE #1884: ChallengeBoss loop farms away the post-battle pacing cooldown (#1851's boss-path variant)
## Problem

#1851 closed the "loop `NewEnemy` abandons to skip every kill's 5s cooldown" farm: `AbandonBattle` now applies the post-battle cooldown when an abandon resolves an outcome, and `StartBattle` anchors the replacement battle at that cooldown's expiry. The boss path didn't get the anchor: `StartBossBattle` (`Game.Application/Services/BattleService.cs:200-239`) runs the internal abandon (which applies the cooldown), then starts the boss battle **at `now`**, ignoring it. `ChallengeBoss` (`Game.Api/Sockets/Commands/ChallengeBoss.cs`) has no `IsOnCooldown` gate — documented as deliberate so a challenge can be *started* anytime.

A scripted client looping `ChallengeBoss` therefore farms bosses at battle-duration cadence with zero inter-battle gap — each call's abandon pays the previous boss win **in full** (exp, `BossesDefeated`, zone-clear credit). An honest boss loop pays 5s + the Zone-Cleared overlay per kill; a 2s one-shot build gains ~3.5× by scripting.

## Design note

The "no cooldown gate on ChallengeBoss" decision covers challenge *availability* (start a boss anytime from idle), not repeated-kill *pacing* — nothing in docs or tests addresses the loop case, so this needs a small design call rather than being an unambiguous defect.

## Suggested fix

Anchor the boss battle at the abandon-incurred cooldown expiry (the exact #1851 fix shape, `abandonedOutcomeCooldown`). To stay honest-client-safe this needs the response to carry the anchor — the same contract gap tracked in #1881, so these two likely land together.

ginderjeremiah:
## Summary

`StartBossBattle` (`Game.Application/Services/BattleService.cs`) always started the replacement boss fight at `now`, even when it had just abandoned an in-progress battle that resolved to a win/loss/draw and incurred the 5s post-battle cooldown (`state.SetCooldown`). `ChallengeBoss` has no cooldown *gate* on the request itself (deliberate — the challenge is meant to be always startable), but nothing anchored the *replacement battle's start* to the cooldown the abandon just incurred, the way #1851 already does for the idle `StartBattle`/`NewEnemy` path.

A scripted client loops `ChallengeBoss` instead of ever sending `DefeatEnemy`. Each call's abandon pays the *previous* boss win in full (exp, `BossesDefeated`, zone-clear credit) via the same win-crediting path a genuine abandon uses, then immediately starts the next boss battle with no anchor — farming bosses at battle-duration cadence with zero inter-battle gap. An honest boss loop pays 5s + the Zone-Cleared overlay per kill; the scripted loop skips both.

## Changes

- `Game.Application/Services/BattleService.cs` — `StartBossBattle` now tracks `abandonedOutcomeCooldown` (mirroring `StartBattle`'s existing pattern) and anchors the replacement boss battle's `now` to that cooldown's expiry when the abandon resolved an outcome.
- `Game.Api/Sockets/Commands/ChallengeBoss.cs` — the response now populates `NewEnemyModel.Cooldown` when the just-incurred cooldown is still in the future, mirroring `NewEnemy`'s identical handshake (#1881) so the client can wait it out.
- `UI/src/lib/engine/battle/enemy-manager.ts` — `challengeBoss()` now awaits `battleEngine.startLoading(cooldown)` before presenting the boss as live when the response carries a positive cooldown, with the same supersession re-check the idle loop's fetch loop uses.
- `docs/backend-battle.md` — updated the existing abandon/cooldown-anchor bullet to cover the boss path too.

## Tests

- `BattleServiceIntegrationTests.cs`: new `StartBossBattle_AbandoningAWonBattle_SetsPostBattleCooldownAndAnchorsReplacementBossBattle`, mirroring the existing `StartBattle_AbandoningAWonBattle_...` regression test for #1851.
- `ChallengeBossSocketTests.cs`: new socket-level tests — a first challenge returns no cooldown, and a looped challenge that abandons a resolved win returns a positive `Cooldown` on the same response as the fresh `EnemyInstance` (mirrors `NewEnemy_AbandonResolvesOutcome_...`).
- `enemy-manager-boss.test.ts`: new tests — the client waits out an anchored cooldown before presenting the boss, and a retreat superseding mid-wait abandons presenting it instead of clobbering the fight the retreat moved to.

## Design note

The issue flagged this as needing "a small design call rather than being an unambiguous defect," since the "no cooldown gate on `ChallengeBoss`" decision covers challenge *availability*, not repeated-kill *pacing*. I resolved it by extending the existing #1851/#1881 anchor-and-inform pattern symmetrically to the boss path rather than adding a new gate that would block a legitimate first challenge — this closes the farming exploit while preserving "the boss challenge is always available."

## Test plan

- [x] `dotnet test --no-build --no-restore` — full backend suite: 3086/3086 passing.
- [x] `npx vitest run` — full frontend suite: 3631/3631 passing.

Closes #1884

---
_Generated by [Claude Code](https://claude.ai/code/session_01EhyYij2VSaj6esXnzhCupg)_


Game.Application/Services/BattleService.cs
```diff
@@ -218,9 +225,14 @@ public Task<BattleStartResult> PrepareNextIdleBattle(Player player, PlayerState
                 // Nothing was cleared or persisted for a still-in-progress abandon, so overwriting the
                 // in-memory state below with the boss battle is safe either way.
                 await AbandonBattle(player, state, clientBattleMs, cancellationToken);
+
+                if (state.EnemyCooldown > DateTime.UtcNow)
+                {
+                    abandonedOutcomeCooldown = state.EnemyCooldown;
+                }
```
ginderjeremiah, line 240: **Should-fix (non-blocking): this guard diverges from `StartBattle`'s in a way the comment doesn't capture.**

`StartBattle` uses `if (handoff is null && state.EnemyCooldown > DateTime.UtcNow)`, and its comment justifies dropping any pre-existing-cooldown concern by noting NewEnemy gates on `IsOnCooldown` before calling in — "a future `EnemyCooldown` at this point can only be the one AbandonBattle just set."

`ChallengeBoss` has **no** `IsOnCooldown` gate, so that justification doesn't hold here. Concrete reachable path: player kills an idle enemy → `DefeatEnemy` sets `EnemyCooldown = completedAt + 5s` (future) and prefetches the next idle battle (so `HasActiveBattle` is true with `BattleStartTime` in the future). Mid-cooldown the player presses Challenge Boss; the client reports `clientBattleMs: 0` (Loading stage), so `AbandonBattle` replays 0 ticks, resolves **no** outcome, and returns early via the `elapsedMs <= 0` branch — which calls `ClearBattle()` but leaves `EnemyCooldown` untouched. Back here, `state.EnemyCooldown > DateTime.UtcNow` is still true from that **pre-existing** cooldown, so `abandonedOutcomeCooldown` is set and the boss is anchored to it — even though this call's abandon resolved nothing.

The runtime effect is benign, even mildly beneficial: it can only ever *delay* the boss to a cooldown the player was already serving, never let it start early, so it isn't exploitable and the #1884 farm is still fully closed. But the comment here ("Non-null only when the abandoned battle resolved to a win/loss/draw") — and the matching one in `ChallengeBoss.cs:52-55` — are inaccurate for the boss path.

Please either:
- **(a)** mirror `StartBattle`'s guard precisely — capture `EnemyCooldown` before the abandon and anchor only if it advanced (`state.EnemyCooldown > cooldownBeforeAbandon`), which keeps the "always startable from idle" intent from #1884 intact even when a leftover idle cooldown is in flight; or
- **(b)** keep the current behavior but correct both comments to state that, because ChallengeBoss has no cooldown gate, the anchor also picks up a still-in-flight pre-existing cooldown.

Leaning (a) since it makes the boss path genuinely match `StartBattle` and the comment true again.

---

ginderjeremiah (commented):
## Review: anchor ChallengeBoss's replacement battle to the abandon-incurred cooldown (#1884)

**Verdict:** Approve (posted as a comment since I'm the PR author) — no blocking findings; one non-blocking should-fix on a guard/comment precision detail. CI is fully green.

This faithfully extends the #1851/#1881 anchor-and-inform pattern to the boss path: `StartBossBattle` now tracks `abandonedOutcomeCooldown` and anchors the replacement boss battle to it, `ChallengeBoss` returns the cooldown on the same response, and `enemy-manager.ts` waits it out before presenting the boss — mirroring the idle `StartBattle`/`NewEnemy`/`getNewEnemy` trio almost line-for-line. I verified the mirrored code on both sides and it's consistent (including the `if (result.data.cooldown)` truthy check and the supersession re-check keyed off the correct generation counter for each path). The #1884 farm is genuinely closed: a looped `ChallengeBoss` whose abandon resolves a win now applies the pacing cooldown and anchors the next boss to it. Solid design call resolving the issue's "small design call" note — extending the existing symmetric anchor rather than adding a gate that would break "the boss challenge is always available."

### Should-fix

- `BattleService.cs:229` — the `abandonedOutcomeCooldown` guard drops `StartBattle`'s `handoff is null &&` qualifier, and because `ChallengeBoss` has no `IsOnCooldown` gate, it can be set from a **pre-existing** leftover idle cooldown (not one this abandon resolved). Behavior is benign (only ever delays the boss, never lets it start early — not exploitable), but the comments here and in `ChallengeBoss.cs:52-55` claiming "Non-null only when the abandoned battle resolved to a win/loss/draw" are inaccurate on the boss path. Details + two suggested fixes on the inline comment. Non-blocking.

### Test coverage

- Both sides covered, following the project's split: backend gets a domain/integration test (`BattleServiceIntegrationTests.StartBossBattle_AbandoningAWonBattle_...` — verified it meaningfully resolves a win via the wall-clock-capped replay with null `clientBattleMs`) and socket tests (`ChallengeBossSocketTests`: first-challenge-no-cooldown and looped-challenge-paced), and the frontend gets `enemy-manager-boss.test.ts` cases for waiting out the cooldown and for a retreat superseding mid-wait. This is the socket/orchestration handshake, not battle-simulation logic, so the frontend↔backend *simulation* parity-vector requirement doesn't apply here — each side testing its half of the handshake is the right shape.
- Not covered (and fine to leave): the pre-existing-idle-cooldown edge above. If you take fix (a), a small case pinning "a leftover idle cooldown does not anchor the boss / does anchor it" would lock the intended behavior in.

### What's good

- Clean, symmetric mirroring of an already-well-established pattern; no dead code, house rules (braces, nullable handling) followed.
- `docs/backend-battle.md` updated accurately, folding the boss path into the existing abandon/anchor bullet and noting why `StartBossBattle` needed the anchor added separately.
- Generated `INewEnemyModel.cooldown` reused from #1881 — no hand-edited codegen.

---

Game.Application/Services/BattleService.cs
ginderjeremiah: Took option (a) — captured `EnemyCooldown` before the abandon in `StartBossBattle` and now only anchor if it advanced, mirroring `StartBattle`'s exact intent instead of relying on a gate `ChallengeBoss` deliberately doesn't have. Also added the regression test you suggested (`StartBossBattle_AbandonResolvesNoOutcomeWithLeftoverCooldown_DoesNotAnchorToTheStaleCooldown`) pinning that a leftover pre-existing cooldown doesn't anchor the boss. Pushed in 965ceb4.

---

PR #1907, opened 2026-07-12
Reject deleting an in-use item mod slot instead of cascading

### LINKED ISSUE #1885: Deleting an item's mod slot cascade-deletes players' AppliedMod rows — silent player-data loss and cache/DB divergence
## Problem

Items, mods, and skills are retire-only, but `ItemModSlot` is legitimately hard-deletable through `AdminItems.SaveModSlots` (`Game.DataAccess/Repositories/Admin/AdminItems.cs:302-311`) with **no in-use check** — and the convention-derived FK `AppliedMod.ItemModSlotId → ItemModSlot` is `OnDelete(DeleteBehavior.Cascade)` (a real `ON DELETE CASCADE` in the DDL; `GameContextModelSnapshot.cs` — `GameContext.cs:85-88` configures only the PK, leaving the four FKs to convention).

Deleting a slot therefore destroys **every player's `AppliedMod` row occupying that slot across the whole player base** at the DB level.

## Why the loss is delayed and invisible

The write-behind design makes it worse than an immediate error:

- Redis (the source of truth for rehydration) still holds each player's applied mod in the cached blob, and `PlayerCacheMapper.ToCore` (`Game.DataAccess/Mapping/PlayerCacheMapper.cs:120-131`) resolves only the `ItemModId` — the slot id is copied raw. Live/cached players keep fighting with the mod's stats.
- No `OrphanedReferenceException` fires — the DB row is *gone*, not unresolvable — and no later save re-emits a `ModAppliedEvent` to restore it.
- On eventual cache eviction, the DB reload silently drops the mod — directly violating the documented "player data is never discarded; an unresolvable owned reference fails loudly" policy (docs/backend.md → Reference Data).

## Suggested fix

- Change the FK to `Restrict` (migration) so the DB refuses the delete outright.
- Add a graceful `AdminSaveResult.Failure` guard in `SaveModSlots` rejecting deletion of an in-use slot, mirroring the retirement guards' anti-tamper shape.
- Integration test covering the in-use rejection.

Related: #1823 is about *missing* FKs on the player aggregate; #1858 about bad references surfacing as FK 500s — neither covers an existing FK whose cascade destroys player rows.

ginderjeremiah:
## Summary

`ItemModSlot` is legitimately hard-deletable through `AdminItems.SaveModSlots` (`Game.DataAccess/Repositories/Admin/AdminItems.cs`) with no in-use check, and the convention-derived FK `AppliedMod.ItemModSlotId → ItemModSlot` was `OnDelete(DeleteBehavior.Cascade)`. Deleting a slot therefore destroyed **every player's `AppliedMod` row occupying that slot across the whole player base** at the DB level — silently and invisibly, since Redis (the source of truth for rehydration) kept the stale applied mod until eventual cache eviction dropped it on reload, directly violating the documented "player data is never discarded; an unresolvable owned reference fails loudly" policy.

## Changes

- **`Game.Infrastructure/Database/GameContext.cs`**: `AppliedMod.ItemModSlotId` is now explicitly configured `OnDelete(DeleteBehavior.Restrict)` instead of the convention's `Cascade`, with a migration (`RestrictAppliedModItemModSlotDelete`). This is the DB-level backstop.
- **New `IAppliedModQueries`/`AppliedModQueries`** (`Game.DataAccess/Repositories/{IAppliedModQueries.cs, Admin/AppliedModQueries.cs}`): queries the write-behind `AppliedMods` table directly for slot occupancy — mirroring the existing `ITagAssignmentQueries`/`Tags` pattern for "unlike the cached reference repos, this queries the database directly."
- **`AdminItems.SaveModSlots`** now rejects deleting a slot occupied by any player's applied mod up front as a graceful `AdminSaveResult.Failure`, mirroring the other content-lifecycle retirement guards (`docs/backend-admin.md`). This required making the method (and its `IAdminItems` interface member and controller action) `async`, since the occupancy check is a DB query.
- **`docs/backend-admin.md`**: documents the new "Hard-delete guards (child collections)" pattern — distinct from the existing "Content-lifecycle retirement guards" section, since `ItemModSlot` isn't a retire-only top-level record.

## Tests

- `AdminItemsIntegrationTests.SaveModSlots_DeleteOfSlotOccupiedByAppliedMod_ReturnsFailureWithoutPersisting`: seeds a player with a mod applied to a slot, attempts to delete that slot, and asserts the save fails without persisting the delete.
- Existing `SaveModSlots` tests updated to the new `async` signature; unaffected in behavior.

## Test plan

- [x] `dotnet test Game.sln` — full backend suite: 3084/3084 passing (including the new regression test).
- [x] `dotnet format Game.sln --verify-no-changes` — clean.
- [x] TypeScript codegen (`dotnet run --project Game.Api -- codegen`) — no drift; the wire contract (`ApiResponse`) is unchanged by the `async` signature change.
- [ ] Frontend unaffected (backend-only change, no wire-contract change) — no frontend verification needed.

Closes #1885

---
_Generated by [Claude Code](https://claude.ai/code/session_01RgRaNjxU8o6fYz4PKXsUsb)_


ginderjeremiah (commented):
## Review: Reject deleting an in-use `ItemModSlot` instead of cascading (closes #1885)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't accept a formal Approve from me).

The change lands all three parts the issue asked for and does so in the idiomatic way for this bounded context:

- **DB-level backstop** — `AppliedMod.ItemModSlotId` explicitly configured `Restrict` (with the `RestrictAppliedModItemModSlotDelete` migration + snapshot update), replacing the convention `Cascade` that was the root cause. The `HasOne(...).WithMany().HasForeignKey(...)` config is correct and the comment explains *why* `ItemModSlot` differs from the retire-only principals it sits beside.
- **Graceful guard** — `SaveModSlots` rejects an occupied-slot delete up front as an `AdminSaveResult.Failure`, matching the shape of the existing content-lifecycle guards (`docs/backend-admin.md`). Placing it after the Add existence check and before staging is right — the commit filter would otherwise persist the rest of the batch.
- **New `IAppliedModQueries`/`AppliedModQueries`** — correctly mirrors the `ITagAssignmentQueries`/`Tags` "query the DB directly, never the admin cache" pattern (interface in `Repositories/`, `internal`, DI-registered). Querying the write-behind `AppliedMods` table directly is the correct source for player-owned data.

Docs (`backend-admin.md` new "Hard-delete guards" subsection) are updated and draw the right distinction from the retirement guards. Backend-only, no wire-contract change from the `async` signature — the codegen-no-drift claim holds.

### Test coverage
Good. `SaveModSlots_DeleteOfSlotOccupiedByAppliedMod_ReturnsFailureWithoutPersisting` seeds a real applied mod, reloads caches, and asserts both the failure message and that the slot survives (no persistence). The pre-existing `SaveModSlots` tests are correctly migrated to the `async` signature with behavior unchanged, and the happy-path delete stays covered by `SaveModSlots_AddsEditsAndDeletesAgainstItemMembership`.

### Consideration (non-blocking, possible follow-up)
The occupancy guard reads the **persisted** `AppliedMods` table, but applied mods are write-behind — a mod applied but not yet drained from the Redis queue won't be visible to `GetOccupiedSlotIds`. So there's a narrow window where deleting a slot with a *pending* (undrained) apply passes the guard, the `Restrict` FK doesn't fire (no DB row yet), and the later drain FK-faults into the dead-letter queue rather than persisting. This is strictly better than the bug being fixed (the player keeps the mod in cache; the failure is observable/replayable, not silent data loss), admin slot-deletion is rare, and cross-player Redis occupancy scanning isn't practical — so I'd treat closing this window as an optional follow-up, not something to hold this PR for.

### Nits
- The occupancy check keys off `c.Item.Id` for every `Delete` change regardless of whether the slot is actually a member of `c.Item.ItemId` — whereas the subsequent apply step no-ops a delete of a non-member slot. A `Delete` naming an occupied slot with the *wrong* `ItemId` would be rejected here rather than reconciled away. Benign in practice (the Workbench computes deltas against a freshly-read baseline, so `ItemId` is always correct), just a slight asymmetry with the membership guard below it.
- `occupiedSlotIds.First()` surfaces only one slot in the message even when several are occupied — fine for a user-facing failure, noting it for completeness.

Nothing here blocks. Nice, tightly-scoped fix with a solid regression test.

---

PR #1908, opened 2026-07-12
Reject flipping a spawn-populated combat zone to Home

### LINKED ISSUE #1887: AdminZones.SaveZones lets a spawn-populated combat zone be flipped to Home — breaking the "Home never spawns" invariant and making Home a relocation target
## Problem

The Home-zone authoring guards cover two of the three ways a Home zone could end up with enemies: `AdminZones.SetEnemies` and `AdminEnemies.SetSpawns` both reject spawn-table membership on a Home zone, and `SaveZones` rejects a boss on a Home zone (`AdminZones.cs:48-51`). The third way is unguarded: **editing an existing combat zone that already has `ZoneEnemy` rows and setting `IsHome = true`**. No guard fires — the single-Home guard (`AdminZones.cs:65-69`) only counts sanctuaries, and its integration test (`AdminZonesIntegrationTests.cs:311`) exercises the flip only when another Home already exists — and the spawn rows survive the save.

## Consequences

Post-save, `HasSpawnableEnemies(home) == true`, so `ZoneResolutionService.IsZoneViable` (which checks retired + spawnable, but not `IsHomeZone` — `ZoneResolutionService.cs:32-37, 73-78`) treats Home as a viable relocation destination. `EnsureViableZone` can then server-side `ChangeZone` a player **into** Home — bypassing the `StartBattle` anti-cheat that refuses Home moves — and players already idling there keep spawning battles inside the no-combat sanctuary. This breaks the documented invariant that "a player's persisted CurrentZoneId is never Home" (docs/backend.md, docs/game-design.md). The content lint only *warns* post-hoc (`ProgressionGraphChecker.cs:194-197`), while the docs frame these invariants as backend-enforced at save time (anti-tamper).

## Suggested fix

In `SaveZones`, reject an edit flipping a zone to Home while its cached `ZoneEnemies` set is non-empty (mirroring the boss check's shape and message), plus an integration test for the flip-with-spawns case.

Related: #1858 covers unvalidated FK references in the relationship setters; #1774 covers lint blind spots — neither covers this save-time invariant hole or its relocation consequence.

ginderjeremiah:
## Summary

`AdminZones.SaveZones` already guarded two of the three ways a Home zone could end up with enemies: declaring a boss on a Home zone, and `SetEnemies`/`AdminEnemies.SetSpawns` populating a Home zone's spawn table. The third was unguarded — editing an **existing combat zone that already has `ZoneEnemy` rows** and setting `IsHome = true`. No guard fired, and the spawn rows survived the save.

Post-save, `HasSpawnableEnemies(home) == true`, so `ZoneResolutionService.IsZoneViable` (which doesn't check `IsHomeZone`) would treat Home as a viable relocation destination, letting `EnsureViableZone` server-side `ChangeZone` a player into Home and bypassing the `StartBattle` anti-cheat that refuses Home moves — breaking the documented invariant that a player's persisted `CurrentZoneId` is never Home.

## Changes

- `Game.DataAccess/Repositories/Admin/AdminZones.cs`: `SaveZones` now rejects an `Edit` that sets `IsHome = true` on a zone whose currently cached `ZoneEnemies` set is non-empty, mirroring the existing boss guard's shape and message. Scoped to edits only — an `Add` can never carry pre-existing spawns.

## Tests

- `AdminZonesIntegrationTests.SaveZones_FlipZoneToHomeWithSpawns_ReturnsFailure`: seeds a combat zone with an active spawn, attempts to flip it to Home, and asserts the save fails with the spawn-table rejection message rather than persisting the flip.

## Test plan

- [x] `dotnet test Game.sln --no-build --no-restore` — full backend suite: 3084/3084 passing (including the new regression test).
- [x] `dotnet format Game.sln --verify-no-changes` — clean.
- [ ] Frontend unaffected (backend-only change, no wire-contract change) — no frontend verification needed.

Closes #1887

---
_Generated by [Claude Code](https://claude.ai/code/session_01HZUJ5CquoXiwzRh7oxv6Ly)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminZonesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminZonesIntegrationTests.cs
@@ -350,6 +350,52 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SaveZones_FlipZoneToHomeWithSpawns_ReturnsFailure()
+        {
+            int combatZoneId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var zone = await TestDataSeeder.CreateZoneAsync(context, "Wilds");
+                var enemy = await TestDataSeeder.CreateEnemyAsync(context);
+                await TestDataSeeder.LinkEnemyToZoneAsync(context, zone.Id, enemy.Id, weight: 1);
+                combatZoneId = zone.Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            // Flipping an already-spawn-populated combat zone to Home is the third way to leave the
+            // sanctuary with live spawns (the other two — a Home boss, and SetEnemies against a Home zone —
+            // are already guarded). The guard reasons over the zone's currently cached spawn table, not this
+            // save's own payload, since SaveZones never carries spawn data.
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminZones>();
+
+            var result = admin.SaveZones(
+            [
+                new Change<Contracts.Zone>
+                {
+                    ChangeType = EChangeType.Edit,
+                    Item = new Contracts.Zone
+                    {
+                        Id = combatZoneId,
+                        Name = "Wilds",
+                        Description = "",
+                        DesignerNotes = "",
+                        LevelMin = 1,
+                        LevelMax = 10,
+                        BossLevel = 1,
+                        IsHome = true,
+                    },
+                },
+            ]);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal(
+                "The Home zone cannot have enemy spawns. Home is a no-combat sanctuary where no enemies spawn; clear the zone's spawn table before making it Home.",
+                result.ErrorMessage);
+        }
+
+        [Fact]
         public async Task SaveZones_RetiredHomeZone_DoesNotBlockNewHome()
         {
             using (var seedScope = CreateScope())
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Reject flipping a spawn-populated combat zone to Home (#1887)

**Verdict:** Approve — closes the last unguarded path to a Home zone with live spawns; the fix is minimal, correctly placed, and mirrors the existing Home guards exactly. (Posted as a comment since I'm the PR author; treat it as an approval.)

The issue's analysis holds up against the code: `SetEnemies`/`AdminEnemies.SetSpawns` and the boss check cover the other paths, but an `Edit` flipping `IsHome = true` on a zone that already carries `ZoneEnemy` rows was unguarded, and `ZoneResolutionService.IsZoneViable` (`ZoneResolutionService.cs`) doesn't consult `IsHomeZone`, so a spawn-populated Home would become a valid `EnsureViableZone` relocation target — the exact invariant break described.

The new guard is well-formed:
- `_zones.LookupZone(change.Item.Id) is { ZoneEnemies.Count: > 0 }` is null-safe — a missing zone simply doesn't match the pattern and falls through to the processor's `editExists` not-found rejection, so the guard never masks that case.
- It reads only `.Count` off the cached graph — no mutation of the `AsNoTracking().Include(...)` reference-data cache, consistent with `backend.md`.
- Scoping to `EChangeType.Edit` is correct: `SaveZones` carries no spawn data, so an `Add` can't arrive with pre-existing spawns, and the guard reasons over the currently cached spawn table rather than the payload — the comment makes that non-obvious point clear.
- Message and shape mirror the sibling boss/spawn-table guards, with the added actionable clause ("clear the zone's spawn table before making it Home").

### Test coverage

`SaveZones_FlipZoneToHomeWithSpawns_ReturnsFailure` seeds a combat zone with an active spawn, reloads the reference caches, attempts the flip, and asserts both `Succeeded == false` and the exact rejection message without persisting — a faithful regression test for the reported hole, structured like the neighboring `SaveZones_FlipExistingCombatZoneToHome_*` cases. Adequate for this backend-only change.

### What's good

Tight, surgical fix that matches the established guard idiom and message conventions; the inline comment captures the one subtle bit (reasoning over cached state, not the save payload) without over-explaining. No blocking items, should-fixes, or nits.

I'll merge once CI is green (backend + e2e still running).

---

PR #1909, opened 2026-07-12
Guard browser Back/Forward against discarding unsaved Workbench edits

### LINKED ISSUE #1892: Browser Back/Forward silently destroys unsaved Workbench edits — the discard guard misses history navigation
## Problem

`/admin` is entered via `goto(resolve('/admin'))` from `/game`, so it sits on the SPA history stack. The dirty-edit guard is wired to exactly three exits (`UI/src/routes/admin/+page.svelte:87-106`, `discard-guard.ts:20-30`): sidebar tool switch and "Return to Game" (both via `confirmDiscard`), and full page unload via `beforeunload`.

Pressing the browser Back (or Forward) button is a client-side popstate navigation: SvelteKit unmounts the admin page without firing `beforeunload`, and there is **no `beforeNavigate` anywhere in `UI/src`** (verified by grep). All pending Workbench/Progression edits are discarded with no prompt. The comment in `discard-guard.ts` even claims client-side navigation is covered by `confirmDiscard` — history navigation is the third client-side exit it misses.

## Suggested fix

Add a SvelteKit `beforeNavigate` guard in `admin/+page.svelte` that calls the existing `confirmDiscard` (cancelling the navigation when declined), plus a test.

Related: #1863 is about retire-confirm gaps; #1856 about save-time id resolution — no open issue touches navigation guards.

ginderjeremiah:
## Summary

The Workbench's dirty-edit guard (`discard-guard.ts`) covered three exits: sidebar tool switch and "Return to Game" (via `confirmDiscard`), and full page unload (`beforeunload`). Browser Back/Forward is a fourth, uncovered exit: it's a client-side popstate navigation, so SvelteKit unmounts `/admin` without firing `beforeunload`, and there was no `beforeNavigate` anywhere in the app — all pending Workbench/Progression edits were silently discarded with no prompt.

## Changes

- **`UI/src/routes/admin/discard-guard.ts`**: new `createPopStateDiscardGuard(getPending)` factory returning a `beforeNavigate` handler. On a dirty `popstate` navigation it cancels the navigation, prompts via the existing `confirmDiscard`, and — if confirmed — replays the same history delta (`history.go(navigation.delta)`) to actually complete the Back/Forward. The replay is itself a popstate the handler would otherwise re-prompt for, so a one-shot bypass flag lets it through once before re-arming.
- **`UI/src/routes/admin/+page.svelte`**: wires the guard via `beforeNavigate(createPopStateDiscardGuard(() => workbenchDirty.total))`, mirroring how the existing `beforeunload` listener reads `workbenchDirty.total`.

Non-popstate navigation types (`link`, `goto`, `form`) are intentionally left alone — those are already guarded by `confirmDiscard` at their call sites (`handleNavigate`, `backToGame`), so a generic handler would have double-prompted.

## Tests

- `UI/src/tests/routes/admin/discard-guard.test.ts`: new `createPopStateDiscardGuard` suite — ignores non-popstate navigations; lets a clean popstate through; cancels and prompts a dirty one; declining leaves the navigation cancelled (no `history.go`); confirming replays the exact delta; the replayed popstate passes through once without re-prompting and then re-arms for a subsequent one.

## Test plan

- [x] `npx vitest run src/tests/routes/admin/discard-guard.test.ts` — 12/12 passing.
- [x] `npm run test:unit:ci` — full suite: 3635/3635 passing, coverage floors intact.

Closes #1892

---
_Generated by [Claude Code](https://claude.ai/code/session_016vDzEnmNC9pNjNaUtWkjVW)_

### DIFF
--- a/UI/src/routes/admin/+page.svelte
+++ b/UI/src/routes/admin/+page.svelte
@@ -99,7 +99,7 @@ const backToGame = async () => {
 	}
 };
 
-// Guards the browser Back/Forward buttons, the client-side exit `handleNavigate`/`backToGame`
+// Guards the browser Back/Forward buttons — a client-side exit that handleNavigate/backToGame
 // don't cover (see discard-guard.ts).
 beforeNavigate(createPopStateDiscardGuard(() => workbenchDirty.total));
 
--- a/UI/src/routes/admin/+page.svelte
+++ b/UI/src/routes/admin/+page.svelte
@@ -41,7 +41,7 @@
 
 <script lang="ts">
 import { onMount } from 'svelte';
-import { goto } from '$app/navigation';
+import { beforeNavigate, goto } from '$app/navigation';
 import { resolve } from '$app/paths';
 import { Loading } from '$components';
 import AdminSidebar from './AdminSidebar.svelte';
@@ -61,7 +61,7 @@ import {
 } from './workbench/nav';
 import { reference } from './workbench/reference.svelte';
 import { ensureAdminAccess } from './admin-access';
-import { confirmDiscard, guardBeforeUnload } from './discard-guard';
+import { confirmDiscard, createPopStateDiscardGuard, guardBeforeUnload } from './discard-guard';
 import { toastError } from '$stores';
 
 let active = $state('enemies');
@@ -98,6 +98,10 @@ const backToGame = async () => {
 		goto(resolve('/game'));
 	}
 };
+
+// Guards the browser Back/Forward buttons, the client-side exit `handleNavigate`/`backToGame`
+// don't cover (see discard-guard.ts).
+beforeNavigate(createPopStateDiscardGuard(() => workbenchDirty.total));
 
 $effect(() => {
 	const handler = (event: BeforeUnloadEvent) => guardBeforeUnload(event, workbenchDirty.total);
--- a/UI/src/routes/admin/discard-guard.ts
+++ b/UI/src/routes/admin/discard-guard.ts
@@ -1,3 +1,4 @@
+import type { BeforeNavigate } from '@sveltejs/kit';
 import { dangerModal } from '$stores';
 
 /**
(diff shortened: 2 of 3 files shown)


UI/src/routes/admin/+page.svelte
```diff
@@ -99,6 +99,10 @@ const backToGame = async () => {
 	}
 };
 
+// Guards the browser Back/Forward buttons, the client-side exit `handleNavigate`/`backToGame`
+// don't cover (see discard-guard.ts).
```
ginderjeremiah: Nit: comma splice reads a little oddly — "the client-side exit `handleNavigate`/`backToGame` don't cover" is missing the relative pronoun. Suggest: `// Guards the browser Back/Forward buttons — a client-side exit that handleNavigate/backToGame don't cover (see discard-guard.ts).`

---

ginderjeremiah (commented):
## Review: Guard browser Back/Forward against discarding unsaved Workbench edits (#1892)

**Verdict:** Approve — correct, well-scoped fix; no blocking issues. (Posted as a comment since I'm the PR author and GitHub won't take a gating verdict on my own PR.)

The fix correctly identifies that browser Back/Forward is a client-side `popstate` navigation that unmounts `/admin` without firing `beforeunload`, and closes it with a `beforeNavigate` guard that cancels synchronously, prompts via the existing `confirmDiscard`, and replays the history delta on confirm behind a one-shot `bypassNext` flag.

Deliberately handling only `popstate` (and leaving `link`/`goto`/`form` to the existing `handleNavigate`/`backToGame` call-site guards) is the right call — a generic `beforeNavigate` would double-prompt on those exits. The refinement over the issue's literal "call `confirmDiscard` from `beforeNavigate`" suggestion is sound and documented.

### Should-fix / verify

- The unit tests mock the `navigation` object, so they exercise the guard's own branching but **not** the real SvelteKit interaction: `cancel()` triggers SvelteKit's internal history restore, and then the confirmed path issues `history.go(navigation.delta)`. That restore→replay sequence is exactly where a regression (double-prompt or a stuck history position) could hide, and it's the one path with no coverage. Worth confirming it was manually exercised (Back with dirty edits → confirm → lands on the previous page once, no re-prompt), or adding a thin e2e — this is the load-bearing behavior of the fix. Not a blocker since I can't demonstrate a defect, but it's the part I'd want eyes on.

### Nits

- `+page.svelte:102` — the wiring comment is a comma splice (missing relative pronoun). Minor wording fix inline.

### Test coverage

Good coverage of the factory's own logic: ignores non-popstate, lets a clean popstate through, cancels+prompts a dirty one, no replay on decline, replays the exact delta on confirm, and the replayed popstate passes once then re-arms. Placement mirrors the source under `tests/routes/admin/`. The only gap is the real-SvelteKit interaction noted above, which isn't unit-testable here.

### What's good

- Reuses `confirmDiscard` rather than duplicating the count/copy logic.
- `pending` read lazily via getter so the guard tracks the live dirty count.
- The `bypassNext` re-arm and the popstate-only scoping are both non-obvious correctness details, and the JSDoc explains *why* clearly without over-documenting.

Not merging despite green CI: I'm the author (can't self-approve on GitHub), and I've left a nit + a verification ask I'd rather see resolved first.

---

ginderjeremiah:
Addressed both points from the review:

- Fixed the comma splice in the `+page.svelte:102` wiring comment per your suggested wording.
- Manually exercised the real SvelteKit `cancel()` → history-restore → confirmed-replay path against a live stack (native `dotnet run` API + fresh Postgres/Redis + `vite dev` on 4173, driven with Playwright): signed up an admin account, dirtied the Workbench (`workbench-new`), hit Back.
  - **Decline:** confirm dialog appears ("You have 1 unsaved change...") → Cancel → stays on `/admin`, edit still pending. ✅
  - **Confirm:** Back again → confirm dialog → Discard and continue → lands on `/game` exactly once, with **no re-prompt** from the replayed `history.go`. ✅

No regressions found in the restore→replay sequence you flagged. Didn't add a permanent e2e spec for this — CLAUDE.md reserves Playwright e2e for critical user flows, and this is an admin-only edge case with full unit coverage of the guard's own logic already in place.

---

PR #1910, opened 2026-07-12
Harden token/notice persistence against localStorage write failures

### LINKED ISSUE #1769: Token persistence hardening: unguarded setItem throws + no in-memory fallback despite the local-storage module's claim
## Problem

- `UI/src/lib/api/token-store.ts:36-38` — `safeLocalStorage()?.setItem(...)`: the guard covers storage *access* only; `setItem` itself can throw `QuotaExceededError`. And with storage blocked there is no memory mirror, despite `local-storage.ts:6-11` claiming in-memory degradation.
- `UI/src/lib/engine/background-throttle-notice.ts:110-117` — same unguarded `setItem` inside a tick/visibility callback.

## Failure scenarios

1. Origin quota near-full (the reference cache shares the origin and grows with content): the refresh succeeds server-side, but `setTokens` throws inside `refreshTokens().then` — the already-consumed single-use pair is lost → forced re-login.
2. Storage-blocked privacy mode: login "succeeds" but every subsequent request is unauthenticated.

## Fix shape

Write-through in-memory mirror in `token-store` plus try/catch around both writes.

ginderjeremiah:
## Summary

`UI/src/lib/api/token-store.ts`'s `setTokens`/`clearTokens` called `setItem`/`removeItem` unguarded — `setItem` can throw `QuotaExceededError` (the reference cache shares the origin and grows with content) even though the shared `safeLocalStorage()` guard's own doc comment claims callers "degrade gracefully to in-memory only rather than throwing." No consumer actually implemented that in-memory fallback. `UI/src/lib/engine/background-throttle-notice.ts` had the same unguarded `setItem` call inside a visibility-change callback.

## Failure scenarios this fixes

1. Origin quota near-full: a token refresh succeeds server-side (the single-use refresh token is consumed), but persisting the new pair throws inside `refreshTokens().then` — the already-consumed pair is lost, forcing a re-login.
2. Storage-blocked privacy mode: login "succeeds" but every subsequent request is unauthenticated because nothing was ever persisted (and, before this fix, in-memory reads didn't fall back to anything either).

## Changes

- **`token-store.ts`**: `getTokens`/`setTokens`/`clearTokens` now maintain a write-through in-memory mirror. `setTokens` writes the mirror first, then best-effort persists to storage (try/catch). Reads normally still prefer storage (so a pair rotated by another tab — a supported state per `auth.ts` — is picked up), but fall back to the mirror when the last storage write is known to have failed, or when reading storage itself throws.
- **`background-throttle-notice.ts`**: wrapped the one-time notice's `setItem` call in try/catch — persistence failure there is low-stakes (the notice just might re-show after a reload), so no mirror is needed, just no uncaught throw.

## Tests

- `token-store.test.ts`: new `degraded storage` suite — a failed write still serves the freshly set pair from memory (and confirms the write genuinely never reached storage); a later successful write resumes normal storage-backed reads; a storage read throwing falls back to memory; `clearTokens` resets both the mirror and the write-failure state. Also updated `beforeEach` to reset the module's mirror/failure-flag state between cases (module-level state that would otherwise leak across tests).
- `background-throttle-notice.test.ts`: new case asserting the notice still shows (and nothing throws) when persisting the one-time flag fails.

## Test plan

- [x] `npx vitest run src/tests/lib/api/token-store.test.ts src/tests/lib/engine/background-throttle-notice.test.ts` — 42/42 passing.
- [x] `npm run test:unit:ci` — full suite: 3634/3634 passing, coverage floors intact.

Closes #1769

---
_Generated by [Claude Code](https://claude.ai/code/session_01VEUQVe4ifSdzsivah9TquX)_

### DIFF
--- a/UI/src/lib/api/token-store.ts
+++ b/UI/src/lib/api/token-store.ts
@@ -13,13 +13,17 @@ export interface StoredTokens {
 
 const STORAGE_KEY = 'gameserver.auth-tokens';
 
-/** The currently stored token pair, or null when the user is not logged in. */
-export const getTokens = (): StoredTokens | null => {
-	const raw = safeLocalStorage()?.getItem(STORAGE_KEY);
-	if (!raw) {
-		return null;
-	}
+/**
+ * In-memory shadow of the stored token pair. Storage is normally the source of truth for reads (so a
+ * pair rotated by another tab is picked up on the next read — see `auth.ts`'s multi-tab handling), but
+ * `setTokens` writes through to this mirror first. When the storage write itself then fails (quota
+ * exceeded, or storage blocked outright), reads fall back to the mirror instead of a stale/absent
+ * storage entry, so a freshly rotated single-use token isn't lost for the rest of this tab's session.
+ */
+let memoryTokens: StoredTokens | null = null;
+let storageWriteFailed = false;
 
+const parseStoredTokens = (raw: string): StoredTokens | null => {
 	try {
 		const parsed = JSON.parse(raw) as Partial<StoredTokens>;
 		if (parsed.accessToken && parsed.refreshToken) {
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Harden token/notice persistence against localStorage write failures (#1769)

**Verdict:** Approve (posted as a comment since I'm the PR author) — cleanly resolves both failure scenarios in #1769 with a well-scoped write-through mirror and thorough tests; no blocking findings.

The change does exactly what the issue asked for and no more:

- `setTokens` writes the in-memory mirror **first**, then best-effort persists — so a server-rotated single-use pair survives a `QuotaExceededError` or blocked storage inside `refreshTokens().then` (scenario 1), and login stays authenticated in storage-blocked privacy mode (scenario 2). This finally makes `local-storage.ts`'s "degrades gracefully to in-memory only" promise actually true for the token store.
- `getTokens` still prefers storage on the happy path (so a pair rotated by another tab is picked up — the multi-tab contract in `auth.ts` is preserved) and only falls back to the mirror when a write is known to have failed or a read throws. The `storageWriteFailed` gate and the write-through-on-read (`memoryTokens = ...`) keep the two in sync correctly, including resuming storage-backed reads once a write succeeds again.
- `clearTokens` correctly resets both the mirror and the failure flag, and the added `beforeEach` `clearTokens()` properly isolates the new module-level state across tests.
- The `background-throttle-notice.ts` fix is appropriately minimal — no mirror, just a try/catch, since a lost one-time flag is low-stakes.

### Test coverage

Strong and matches the project's frontend rule (tests mirror source under `tests/`). The `degraded storage` suite covers write-throws-serves-from-memory (and asserts storage genuinely stayed empty), resume-after-recovery, read-throws-falls-back, and `clearTokens` reset; the notice test covers the throwing-persist path. Adequate for a change of this size — no gaps worth blocking on.

### What's good

- The "read prefers storage, but pin to the mirror once a write has failed" tradeoff is a deliberate, documented call (the mirror holds the freshest pair *this* tab set), and the doc comment on `memoryTokens` explains it well — exactly the kind of subtle intent a later edit could naively "simplify," so calling it out in the comment was the right move.
- No new stores, no hardcoded values, no dead code, house braces respected — nothing to nit.

CI is still running (`test-frontend` + e2e in progress; `verify-frontend` already green). Holding the merge until it goes fully green.

---

PR #1911, opened 2026-07-12
Disable empty-state Add control when its option catalogue is empty

### LINKED ISSUE #1894: Workbench empty-state "Add row" crashes when the option catalogue is empty (firstFree unguarded, noFree not threaded into EmptySection)
## Problem

`firstFree` ends with `(free ?? options[0]).value` — a `TypeError` when `options` is empty (`UI/src/routes/admin/workbench/entities/helpers.ts:4-7`). The normal Add button is protected: `noFree` disables it, and `[].every(...)` is vacuously `true`, so an empty catalogue disables Add. But when a table has **zero rows**, `SectionTable` renders `EmptySection` with `onAdd={add}` and never passes `noFree` (`SectionTable.svelte`), and `EmptySection` renders an always-enabled button (`EmptySection.svelte:7-11`).

Reachable states (fresh DB or all-retired catalogues, since option lists filter retired records):

- enemy → Spawns → "Assign zone" with zero active zones
- recipe → Conditions → "Add condition" with zero active proficiencies
- zone → Enemies with zero active enemies

Click → uncaught exception in the handler; the row is never added.

## Suggested fix

Guard `firstFree` against empty options and/or thread `noFree` (with the disabled state + title) into `EmptySection`.

ginderjeremiah:
## Summary

`firstFree` (`UI/src/routes/admin/workbench/entities/helpers.ts:4-7`) ends with `(free ?? options[0]).value` — a `TypeError` when `options` is empty. The normal (non-empty-rows) Add button is already protected: `SectionTable` computes `noFree` (true once every unique option is assigned, vacuously true when the option catalogue is empty) and disables Add. But when a table has zero rows, `SectionTable` renders `EmptySection` instead, and never passed `noFree` through — so its Add button stayed enabled and calling `newRow` → `firstFree` on an empty catalogue crashed.

Reachable states (fresh DB or all-retired catalogues, since option lists filter retired records):
- enemy → Spawns → "Assign zone" with zero active zones
- recipe → Conditions → "Add condition" with zero active proficiencies
- zone → Enemies with zero active enemies

## Changes

- **`UI/src/routes/admin/workbench/components/SectionTable.svelte`**: pass the already-computed `noFree` through to `EmptySection`.
- **`UI/src/routes/admin/workbench/components/EmptySection.svelte`**: accept an optional `noFree` prop; when true, disable the Add button and set the same "Every option is already assigned" title the populated-rows path already uses.

`firstFree` itself is unchanged — once the button is correctly gated, it's never invoked with an empty `options` array (it has exactly one caller, `SectionTable`'s `add()`), so an extra guard there would just be defensive code for an unreachable path.

## Tests

- `UI/src/tests/routes/admin/workbench/SectionTable.test.ts`: new case rendering the empty state with an empty option catalogue, asserting the Add button is disabled with the expected title. Verified this test fails (button not disabled) without the fix and passes with it.

## Test plan

- [x] `npx vitest run src/tests/routes/admin/workbench/SectionTable.test.ts` — 12/12 passing.
- [x] `npm run test:unit:ci` — full suite: 3635/3635 passing, coverage floors intact.

Closes #1894

---
_Generated by [Claude Code](https://claude.ai/code/session_01G8QzWVriytwhLTCA3S6qe9)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/EmptySection.svelte
+++ b/UI/src/routes/admin/workbench/components/EmptySection.svelte
@@ -5,7 +5,13 @@
 		<div class="es">{sub}</div>
 	</div>
 	{#if onAdd && addLabel}
-		<button type="button" class="btn sm" onclick={onAdd}>
+		<button
+			type="button"
+			class="btn sm"
+			onclick={onAdd}
+			disabled={noFree}
+			title={noFree ? 'Every option is already assigned' : undefined}
+		>
 			<WorkbenchIcon kind="plus" size={12} />{addLabel}
 		</button>
 	{/if}
@@ -20,7 +26,8 @@ interface Props {
 	sub: string;
 	addLabel?: string;
 	onAdd?: () => void;
+	noFree?: boolean;
 }
 
-const { icon, title, sub, addLabel, onAdd }: Props = $props();
+const { icon, title, sub, addLabel, onAdd, noFree = false }: Props = $props();
 </script>
--- a/UI/src/routes/admin/workbench/components/SectionTable.svelte
+++ b/UI/src/routes/admin/workbench/components/SectionTable.svelte
@@ -5,6 +5,7 @@
 		sub={section.emptySub}
 		addLabel={section.addLabel}
 		onAdd={add}
+		{noFree}
 	/>
 {:else}
 	<div>
--- a/UI/src/tests/routes/admin/workbench/SectionTable.test.ts
+++ b/UI/src/tests/routes/admin/workbench/SectionTable.test.ts
@@ -124,6 +124,29 @@ describe('SectionTable', () => {
 		expect((screen.getByText('Assign zone').closest('button') as HTMLButtonElement).disabled).toBe(true);
 	});
 
+	it('disables the empty-state Add control (instead of crashing) when the option catalogue is empty', () => {
+		const emptyOptsSection: TableSectionConfig<Row> = {
+			...section,
+			columns: [
+				{ key: 'zoneId', label: 'Zone', type: 'select', options: () => [], unique: true },
+				{ key: 'weight', label: 'Weight', type: 'number', align: 'r' },
+				{ key: '__share', label: 'Share', type: 'share', weightKey: 'weight' }
+			]
+		};
+		const { store, record, baseline } = setup([]);
+		render(SectionTable, {
+			props: {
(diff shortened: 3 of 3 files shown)


ginderjeremiah (commented):
## Review: Disable empty-state Add control when its option catalogue is empty (#1894)

**Verdict:** Approve *(submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a self-`APPROVE`)* — correctly fixes the reported crash by gating the empty-state Add button, with a test that pins the behaviour.

The fix is minimal and lands exactly where the bug lives: `SectionTable` already computed `noFree` and used it for the populated-rows Add button, but never threaded it into `EmptySection`, so the zero-rows path kept an always-enabled button that called `firstFree` on an empty catalogue and threw. Passing `noFree` through and honouring it (disabled + the same "Every option is already assigned" title) closes the gap consistently with the existing populated-rows button. Verified the logic: with zero rows, `options.every(o => rows.some(...))` is `true` only when `options` is empty, so the empty-state button is disabled *only* in the crashing scenario and stays enabled when there are assignable options — the correct behaviour.

### On leaving `firstFree` unguarded
The PR body justifies not guarding `firstFree` by saying it "has exactly one caller, `SectionTable`'s `add()`." That's slightly imprecise — there's a second direct caller, `ProgressionStore.addModifier` (`progression/progression-store.svelte.ts:334`). I checked it, though, and it's safe: it passes `reference.attributeOptions()`, which is `toOptions(enumPairs(EAttribute))` — derived from the `EAttribute` enum, not from retirable DB reference data, so it can never be empty. Every `firstFree` path that *can* receive an empty option list (zone spawns, enemy spawns, recipe conditions) flows through the now-gated `add()`. So the substantive conclusion holds and gating the button is a defensible design choice over having `firstFree` return a garbage sentinel. No change requested here — just flagging that the "one caller" framing is looser than the code.

### Test coverage
Good — the new `SectionTable.test.ts` case renders the empty state with an empty option catalogue and asserts the Add button is both disabled and carries the expected title, exercising the exact regression. The `EmptySection` change is covered transitively through this. Mirrors the source path per the frontend testing guidelines.

### What's good
Tight, well-scoped bugfix that reuses the existing `noFree` signal rather than inventing a parallel guard, keeps the two Add buttons behaviourally consistent, and comes with a regression test verified to fail without the fix.

---

PR #1912, opened 2026-07-12
Handle Workbench seed-load failures with an error/retry state

### LINKED ISSUE #1896: Workbench seed load has no failure handling — a failed refresh() leaves a permanent spinner and an unhandled rejection
## Problem

`Workbench.svelte`'s `onMount(async () => { const seed = await entity.refresh(); ... })` has no catch (`UI/src/routes/admin/workbench/Workbench.svelte:59-63`). `fetchSocketData` throws on a server error, a dropped-connection settle, or the 30s request timeout (all resolve the socket promise with `error`, which `fetchSocketData` converts to a throw — `api-socket.ts:359-362,461-467`). The socket `errorHook` toast only fires on WebSocket-level `error` events, not on these, so the failure is an unhandled promise rejection: `store` stays `undefined` and the user sees an infinite `Loading` spinner with no toast and no retry.

The Progression surface handles the identical case properly — catches, toasts, stores `error`, renders a Refresh button (`progression-store.svelte.ts:64-78`, `Progression.svelte:37-41`) — so 9 of the 10 admin tools are strictly worse than the 10th. `reference.load()` failure in `admin/+page.svelte:80-84` similarly leaves every tool stuck on `Loading` with only a one-shot toast.

There is also no test covering a failing seed load (`Workbench.test.ts` mocks only successful refreshes).

## Suggested fix

Mirror the Progression error/retry pattern in `Workbench.svelte` (catch → error state → Refresh button), optionally add a retry affordance for `reference.load`, and add a failing-seed-load test.

ginderjeremiah:
## Summary

`Workbench.svelte`'s `onMount(async () => { const seed = await entity.refresh(); ... })` had no catch. `fetchSocketData` throws on a server error, a dropped-connection settle, or the 30s request timeout, and the socket's `errorHook` toast only fires on WebSocket-level `error` events — not on these — so a failed seed load was an unhandled promise rejection: `store` stayed `undefined` and the admin saw an infinite `Loading` spinner with no toast and no retry.

The Progression surface (`progression-store.svelte.ts` / `Progression.svelte`) already handles the identical failure correctly: catch, toast, store an `error`, and render a Refresh button. This PR brings the other 9 admin tools (everything routed through `Workbench.svelte`) up to the same standard.

## Changes

- **`UI/src/routes/admin/workbench/Workbench.svelte`**: extracted the seed load into a `loadSeed()` function that catches a failed `entity.refresh()`, stores the message in a new `error` state, and toasts it (mirroring `ProgressionStore.load()`). The template gains an `{:else if error}` branch rendering the error message plus a `Refresh` button that re-invokes `loadSeed()`, styled to match `Progression.svelte`'s `.prog-error`.

## Not included

The issue also flagged `reference.load()`'s failure in `admin/+page.svelte` (leaves every tool stuck on `Loading` with only a one-shot toast) as an **optional** companion fix. I left that out to keep this PR focused on the required Workbench fix — happy to follow up separately if wanted.

## Tests

- `UI/src/tests/routes/admin/workbench/Workbench.test.ts`: two new cases — a failed `refresh()` shows the error state (message + toast) instead of the workbench content, and clicking Refresh retries the load and renders normally once it succeeds.

## Test plan

- [x] `npx vitest run src/tests/routes/admin/workbench/Workbench.test.ts` — 10/10 passing.
- [x] `npm run test:unit:ci` — full suite: 3637/3637 passing, coverage floors intact.

Closes #1896

---
_Generated by [Claude Code](https://claude.ai/code/session_016UHrFxM4cTs8YPEd9cPbo9)_

### DIFF
--- a/UI/src/routes/admin/workbench/Workbench.svelte
+++ b/UI/src/routes/admin/workbench/Workbench.svelte
@@ -28,6 +28,11 @@
 				onNew={newItem}
 			/>
 		</div>
+	{:else if error}
+		<div class="workbench-error" role="alert" data-testid="workbench-error">
+			<p>{error}</p>
+			<button type="button" class="btn" onclick={loadSeed}>Refresh</button>
+		</div>
 	{:else}
 		<Loading loading={true} delay={150} />
 	{/if}
@@ -36,6 +41,7 @@
 <script lang="ts">
 import { onMount, onDestroy } from 'svelte';
 import { Loading } from '$components';
+import { toastError } from '$stores';
 import './workbench.scss';
 import type { EntityConfig, Identified } from './entities/types';
 import { workbenchDirty } from './dirty.svelte';
@@ -52,14 +58,26 @@ interface Props {
 const { entity, groupLabel = '' }: Props = $props();
 
 let store = $state<EntityStore<Identified>>();
+let error = $state<string | null>(null);
 let selId = $state(0);
 let selectedTab = $state<string>();
 const tab = $derived(selectedTab ?? entity.sections[0]?.key);
 
-onMount(async () => {
-	const seed = await entity.refresh();
-	store = new EntityStore(entity, seed);
-	selId = seed[0]?.id ?? 0;
+async function loadSeed() {
+	error = null;
+	try {
+		const seed = await entity.refresh();
+		store = new EntityStore(entity, seed);
+		selId = seed[0]?.id ?? 0;
+	} catch (ex) {
+		const message = ex instanceof Error ? ex.message : 'Failed to load workbench data.';
+		error = message;
+		toastError(message);
+	}
+}
+
+onMount(() => {
+	void loadSeed();
 });
 
 // Cancel any pending "saved" flash timer so a save that lands near unmount can't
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Handle Workbench seed-load failures with an error/retry state (closes #1896)

**Verdict: Approve** *(submitted as a comment since I authored the PR; no blocking or should-fix findings)* — the change faithfully mirrors the already-proven Progression error/retry pattern, is well-scoped, and is properly tested.

### What's good

- `loadSeed()` mirrors `ProgressionStore.load()` exactly — `error = null` up front, `try/catch`, the same `ex instanceof Error ? ex.message : <fallback>` extraction, and both an `error` state and a `toastError`. Consistency with the sibling surface is the right call here.
- Reworking `onMount(async …)` into `onMount(() => void loadSeed())` also removes a real footgun: an `async` `onMount` returns a Promise, which Svelte would misinterpret as a cleanup function. Good catch beyond the literal bug.
- The `{:else if error}` branch and `.workbench-error` styles match `Progression.svelte`'s `.prog-error` (same `var(--error)` token, same layout), and the `.btn` class resolves via the global `workbench.scss` import — the Refresh button is styled.
- Tests cover both the failure state (error shown, title hidden, `toastError` called) and the retry path (Refresh re-invokes `refresh()` and renders normally on success), which closes the "no failing-seed-load test" gap the issue called out.

### Follow-up (not blocking)

- The issue's optional companion — `reference.load()` failure in `admin/+page.svelte:80-84` still has only a one-shot toast and no retry — is intentionally out of scope here (separate file, separate concern). Since this PR closes #1896, I've filed **#1913** to preserve that tracking.

Clean, focused fix. Merging once CI goes green.

---

PR #1914, opened 2026-07-12
Gate the welcome-back offline-progress check on the boot gate resolving

### LINKED ISSUE #1898: Direct load/refresh of /game consumes the offline-progress window before the boot gate: the welcome-back summary never shows and the in-progress battle hand-back is dropped
## Problem

`routes/game/+page.svelte` runs `onMount(() => { void welcome.run(); })` unguarded on any mount (`+page.svelte:85-88`), while the boot gate flips `booting = true` inside the **layout's** `onMount` (`+layout.svelte:53-84`) — and Svelte mounts child-first (`child-init → child-onMount → parent-onMount`, verified empirically in this repo's vitest setup).

Sequence on a refresh/tab-restore landing directly on `/game` (the most common way a player returns):

1. The page mounts and `welcome.run()` immediately sends `GetOfflineProgress`, which queues the command and starts the auto-connect (`api-socket.ts:174-182` — nothing cancels a queued command).
2. The layout's boot gate sets `booting = true`; the `{#if booting}` branch destroys the page → `onDestroy` calls `welcome.cancel()`.
3. The server still executes the command — it is **not idempotent**: `GetOfflineProgress.cs:25-36` applies rewards, resolves/clears the stale in-flight battle, and persists — while the client discards the response at the `if (this.cancelled) return;` guard (`welcome-back-view.svelte.ts:41-43`).
4. After `resumeSession()` the page **remounts** and runs `welcome.run()` again — the away window is now ~0, so `hasProgress` is false and the gate passes straight through.

Net effect: no economy loss (Login/Status re-pulls the credited aggregate), but the away summary (#1043) is never shown on a direct `/game` load, and the `activeBattle` payload the first call carried is discarded — defeating the replay-to-offset resume (#1595/#1596/#1597) on this path. The same consumption happens when the boot gate redirects to `/loading` (stale cache), so the gate is skipped after the manifest too.

## Test gap

No test observes how many times `GetOfflineProgress` is issued on a cold `/game` load — `layout.svelte.test.ts` mocks the route content and `game/page.test.ts` never simulates the boot-gate-induced mount→unmount→remount cycle.

## Suggested fix

Gate `welcome.run()` on the boot decision having resolved (e.g. an exported `booted` signal, or have the layout render children only after boot on token-bearing loads), or have the page skip the pre-boot run while `playerManager` is uninitialized. Pin with a test counting `GetOfflineProgress` sends across the remount cycle.

Related: #1835/#1807 added the `cancel()` that closes the start-engines-after-unmount race but left this consumption.

ginderjeremiah:
## Summary

`routes/game/+page.svelte` ran `welcome.run()` (which issues the non-idempotent `GetOfflineProgress` command) unconditionally on its own `onMount`, while the root layout's boot gate (`+layout.svelte`) only flips `booting = true` inside its **own** `onMount` — and Svelte mounts children before the parent (`child-onMount → parent-onMount`).

On a direct load/refresh of `/game` (the most common way a player returns) this meant:

1. The page mounted and immediately fired `GetOfflineProgress`.
2. The layout's boot gate then set `booting = true`, tearing the page down (`welcome.cancel()`), before the response came back.
3. The server still executed the command in full (applies rewards, resolves/clears the stale in-flight battle, persists) — the client just discarded the response.
4. After the boot gate resolved, the page remounted and ran `welcome.run()` again with an away window of ~0, so the gate passed straight through with nothing to show.

Net effect: no economy loss (Login/Status re-pulls the credited aggregate), but the away-summary gate (#1043) never showed on a direct `/game` load, and the in-progress battle hand-back was silently dropped, defeating the replay-to-offset resume (#1595/#1596/#1597) on this path.

## Changes

- **`UI/src/lib/engine/boot-state.svelte.ts`** (new): a small module-level `bootState` signal (`booted` / `markBooted()`) that lets a route mounted independently of the root layout know whether the boot gate has resolved at least once this session.
- **`UI/src/routes/+layout.svelte`**: calls `bootState.markBooted()` in the boot `onMount`'s `finally` block, alongside the existing local `booted` flag it already uses for its own post-boot safety net.
- **`UI/src/routes/game/+page.svelte`**: only calls `welcome.run()` on mount when `bootState.booted` is already `true`. On a pre-boot mount this is a no-op — safe because the boot gate's `{#if booting}` toggle always tears the page down and remounts it once boot resolves, and by that point `bootState.booted` is already `true`, so the remount runs the check for real. A route reached by normal in-app navigation (e.g. from character-select, after the boot gate has already resolved once) runs this on its only mount, unchanged from before.

## Tests

- `UI/src/tests/routes/game/page.test.ts`: new case pinning that a pre-boot mount does **not** send `GetOfflineProgress`, and that the post-boot remount the layout always performs does.
- Existing layout/page suites pass unchanged — `bootState` defaults to already-resolved in the page test file (matching every mount except the direct-load scenario the new test covers).

## Test plan

- [x] `npx vitest run src/tests/routes/game/page.test.ts src/tests/routes/layout.svelte.test.ts` — passing.
- [x] `npm run test:unit:ci` — full suite: 3638/3638 passing, coverage floors intact.

Closes #1898

---
_Generated by [Claude Code](https://claude.ai/code/session_01FeqjxNhJcYWnA6FeLJ5Hor)_

### DIFF
--- a/UI/src/routes/+layout.svelte
+++ b/UI/src/routes/+layout.svelte
@@ -19,6 +19,7 @@ import { apiSocket, getTokens, onSocketError } from '$lib/api';
 import { handleSocketReplaced, playerManager } from '$lib/engine';
 import { resumeSession } from '$lib/engine/session';
 import { bootRedirect, shouldReturnToLogin } from '$lib/engine/boot-redirect';
+import { bootState } from '$lib/engine/boot-state.svelte';
 import { toastError } from '$stores';
 import BootSplash from './BootSplash.svelte';
 import '$styles/common.scss';
@@ -80,6 +81,9 @@ onMount(async () => {
 		// gate can never strand the player on the splash.
 		booting = false;
 		booted = true;
+		// Signals routes mounted independently of this layout (e.g. #1898) that the boot decision has
+		// resolved at least once, so their own pre-boot mount can defer non-idempotent work.
+		bootState.markBooted();
 	}
 });
 
--- a/UI/src/lib/engine/boot-state.svelte.ts
+++ b/UI/src/lib/engine/boot-state.svelte.ts
@@ -1,0 +1,23 @@
+/* Cross-route boot signal (#1898). The root layout's boot gate (`+layout.svelte`) decides where a
+   fresh page load lands — but Svelte mounts children before the parent's `onMount`, so a route
+   mounted directly (e.g. a refresh/tab-restore landing on `/game`) runs its own `onMount` at least
+   once *before* the boot gate has even started resuming the session. A route whose mount work isn't
+   safe to run pre-boot (`/game`'s welcome-back gate issues a non-idempotent `GetOfflineProgress`)
+   needs to know whether the boot decision has already resolved at least once this session.
+
+   Module-level rather than component state: the boot gate's resolution needs to be visible to a
+   route mounted independently of the layout, and it only ever needs to go false → true once per
+   session (a later route mount after the first boot always finds it already true). */
+
+let booted = $state(false);
+
+export const bootState = {
+	get booted() {
+		return booted;
+	},
+	/** Marks the boot gate's decision as resolved. Idempotent; called once by the root layout's boot
+	 *  `onMount` after `resumeSession` (or the no-tokens short-circuit) completes. */
+	markBooted() {
+		booted = true;
+	}
+};
--- a/UI/src/routes/game/+page.svelte
+++ b/UI/src/routes/game/+page.svelte
@@ -40,6 +40,7 @@ import { NavSidebar, LogPanel, TourPlayer } from '$components';
 import { GAME_SCREENS, visibleScreens } from './screens/screen-defs';
 import PlaceholderScreen from './screens/PlaceholderScreen.svelte';
 import { startGame, stopEngines, enemyManager } from '$lib/engine';
+import { bootState } from '$lib/engine/boot-state.svelte';
 import { refreshPlayer } from '$lib/engine/session';
 import { evaluateScreenTrigger, closeTutorialTour } from '$lib/engine/tutorials';
 import { navigation, requiresRemount, tutorialTour } from '$stores';
@@ -83,7 +84,18 @@ if (browser) {
 		welcome.cancel();
 	});
 	onMount(() => {
-		void welcome.run();
+		// The root layout's boot gate (`+layout.svelte`) resolves after this page's own `onMount`
+		// (Svelte mounts children before the parent), so a direct load/refresh of `/game` would
+		// otherwise fire `GetOfflineProgress` — a non-idempotent command — before the boot gate has
+		// even started resuming the session, then get unmounted mid-flight when the gate's splash
+		// takes over (see `welcome.cancel()` above). Skipping the pre-boot mount here is safe: the
+		// boot gate's `{#if booting}` toggle always tears this page down and remounts it once resolved
+		// (#1898), so the post-boot remount runs `welcome.run()` for real. A route reached by normal
+		// in-app navigation (e.g. from character-select) finds the boot gate already resolved and runs
+		// this on its only mount, same as before.
+		if (bootState.booted) {
+			void welcome.run();
+		}
 	});
 }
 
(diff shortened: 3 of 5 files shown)


UI/src/routes/+layout.svelte
```diff
@@ -80,6 +81,9 @@ onMount(async () => {
 		// gate can never strand the player on the splash.
 		booting = false;
 		booted = true;
+		// Signals routes mounted independently of this layout (e.g. #1898) that the boot decision has
+		// resolved at least once, so their own pre-boot mount can defer non-idempotent work.
+		bootState.markBooted();
```
ginderjeremiah, line 86: **Should-fix (test coverage):** This `markBooted()` call is the load-bearing seam of the whole fix, but nothing tests it. `page.test.ts` mocks `$lib/engine/boot-state.svelte`, and `layout.svelte.test.ts` neither mocks nor asserts it — so a regression that removed or misplaced this line would pass CI while breaking every boot-path `/game` load: `bootState.booted` stays `false`, the post-boot remount also skips `welcome.run()`, and `/game` hangs on `BootSplash` indefinitely (the game never starts). That silent-no-run is exactly the class of failure this PR fixes.

`layout.svelte.test.ts` already exercises the boot gate end-to-end and doesn't mock this module, so an assertion is cheap: import `bootState` and, after a boot resolves (both the token-bearing resume path and the no-tokens short-circuit — both hit this `finally`), assert `bootState.booted` is `true`. Worth adding here rather than as a follow-up, since it guards this PR's own change.

---

ginderjeremiah (commented):
## Review: gate `/game`'s welcome-back offline-progress check on the boot gate resolving (#1898)

**Intended verdict: Approve** (submitted as a comment since I'm the PR author) — one should-fix and one nit, neither blocking. The fix is correct and the root cause is diagnosed precisely.

I traced every boot path and the core invariant holds. On a token-bearing `/game` load, the layout always flips `booting` true→false, which the `{#if booting}` toggle turns into a genuine unmount→remount, so the pre-boot mount cleanly skips `welcome.run()` and the post-boot remount runs it exactly once (fixing the old non-idempotent double-issue of `GetOfflineProgress`). On the no-tokens path, `booting` never flips, but the gate `goto('/')`s off `/game`, so skipping the pre-boot run correctly means no offline-progress call for a logged-out user. `markBooted()` runs in the `finally` on every path, so `bootState.booted` is always resolved by the time any later in-app navigation reaches `/game`. Module-level `$state` is the right choice here — the layout's component-local `booted` can't be read by an independently-mounted route.

### Should-fix

- `UI/src/routes/+layout.svelte:86` — the new `bootState.markBooted()` call is untested (page test mocks the module, layout test ignores it). It's the seam the whole fix hinges on; a regression there hangs `/game` on the splash with no failing test. `layout.svelte.test.ts` already drives the boot gate without mocking this module, so asserting `bootState.booted === true` after boot resolves is cheap. Details inline.

### Nits

- `UI/src/routes/+layout.svelte:57,83` — the layout now carries two parallel "booted" flags: the component-local `booted` (feeding the post-boot safety-net `$effect`) and the module-level `bootState.booted`, both set in the same `finally`. Since `bootState.booted` is a tracked `$state` getter, the `$effect` at line 98 could read it directly and the local `booted` could be dropped, removing the redundancy. Optional — keeping boot-gate reactivity component-local is a defensible reason to leave it.

### Test coverage

- Page side is well covered: `page.test.ts` pins that a pre-boot mount issues no `GetOfflineProgress` and the post-boot remount does — exactly the counting test the issue's "Test gap" asked for.
- Gap is the layout side (the `markBooted()` seam), as noted above.

### What's good

- Accurate root-cause: the child-before-parent `onMount` ordering and the non-idempotent `GetOfflineProgress` consumption are both correctly identified and the fix targets the real mechanism rather than papering over symptoms.
- The pre-boot-skip is safe precisely because the boot gate's existing unmount/remount cycle guarantees a real post-boot run — and that reasoning is captured well in the code comments at each site.
- Comments are appropriately scoped: enough to explain a genuinely non-obvious cross-route ordering hazard, without over-narrating.

---

ginderjeremiah:
Addressed the should-fix: added a test in `layout.svelte.test.ts` spying on `bootState.markBooted()` to confirm it fires on both the no-tokens short-circuit and the token-bearing resume path (both hit the shared `finally`). Verified it actually catches the regression by temporarily removing the call and confirming the test fails.

Left the nit (collapsing the component-local `booted` into `bootState.booted`) as-is — keeping boot-gate reactivity component-local for the redirect effect seems reasonable, per your own note that it's optional.

---

PR #1915, opened 2026-07-13
Track the processing list's specific head item for the opportunistic reclaim grace clock

### LINKED ISSUE #1746: Opportunistic processing-list reclaim: refine grace-clock heuristic for busy multi-instance deployments
Follow-up to #1702 / PR #1745.

The opportunistic reclaim added in #1745 uses a single instance-local grace clock keyed on "the processing list has been non-empty (with the main queue drained) for ≥ 30s". It does **not** distinguish which specific items are stranded vs. freshly reserved by another live instance.

## Impact

- **Single-instance (current model):** No spurious behavior. When a healthy drain finishes, every item this instance reserved has been acknowledged, so `GetProcessingCountAsync` reads 0, the clock resets, and the reclaim only ever fires against a genuinely stranded item.
- **Busy multi-instance (future):** The shared processing list rarely reads empty (other instances always have an item in flight), so the clock never resets. After 30s a reclaim fires against whatever healthy in-flight items happen to be on the list, pulling them back to the queue head for a redundant re-apply. This is **safe** under the documented idempotent at-least-once contract, but it is wasteful (extra Redis churn + double-applies) and the INFO log misleadingly reports healthy items as "stranded ... for over 30s", which could trigger false operator alarms.

## Suggested direction

Track a per-item first-seen timestamp (e.g. peek the processing list and remember the oldest item's identity + when it was first observed, only reclaiming once *that specific item* has persisted past the grace period) so the reclaim targets genuinely stuck items rather than firing on steady-state churn. This was consciously deferred in #1745 to avoid over-engineering before the simple heuristic is proven insufficient in practice — this issue tracks revisiting it if/when multi-instance write-behind draining is enabled.

No user-facing effect; the current behavior is safe. This is a precision/robustness refinement of an internal mechanism.

ginderjeremiah:
## Summary

The opportunistic processing-list reclaim added in #1745 (for #1702) used a single instance-local "grace clock" keyed on "the processing list has been non-empty (with the main queue drained) for ≥ 30s". That's fine for the current single-instance model, but in a busy multi-instance deployment the shared processing list rarely reads empty (other instances always have something in flight), so the clock never resets — after 30s a reclaim would fire against whatever healthy in-flight items happen to be on the list, which is safe under the idempotent at-least-once contract but wasteful (extra Redis churn + double-applies), and the INFO log would misleadingly report healthy items as "stranded ... for over 30s".

This PR tracks the *specific item* sitting at the processing list's head instead of mere list occupancy:

- Added `IPubSubQueue.PeekProcessingAsync` (and the `RedisQueue`/in-memory-fake implementations) — a non-destructive peek of the processing list, mirroring the existing `PeekAsync` for the main queue.
- `DataProviderSynchronizer.TryReclaimStrandedProcessingAsync` now peeks the head item and tracks its identity + first-observed timestamp (replacing the old occupancy-only clock). The clock restarts whenever a different item takes the head's place (the previously-tracked one was acknowledged/reclaimed), and only fires once the same item has genuinely dwelled there past the grace period.
- Updated the doc comments on the field/method/constant, and the "Queue processing resilience" section of `docs/backend-persistence.md`, to describe the refined heuristic.

No behavior change for the current single-instance model (the two existing single-item reclaim tests still pass unmodified); this only changes precision for a future busy multi-instance deployment.

Closes #1746. Implements the issue's suggested direction exactly: "track a per-item first-seen timestamp ... only reclaiming once *that specific item* has persisted past the grace period."

## Test plan

- [x] Added `DataProviderSynchronizerTests.ProcessQueue_DifferentItemReplacesProcessingHeadWithinGracePeriod_RestartsClockInsteadOfReclaiming` — models a busy-churn scenario (list continuously non-empty, but a different item at the head) and asserts no reclaim fires until that specific item itself has dwelled past the grace period. This test fails against the old occupancy-only clock and passes against the new one.
- [x] Added `RedisQueueIntegrationTests.PeekProcessingAsync_*` covering the new Redis-backed peek (ordering, non-positive count, cancellation).
- [x] `dotnet test` (full solution: Core, Infrastructure, Application, Api) — 3099/3099 passing.

https://claude.ai/code/session_01TJQMpD2CWDphFGE3fHMr6L

---
_Generated by [Claude Code](https://claude.ai/code/session_01TJQMpD2CWDphFGE3fHMr6L)_


ginderjeremiah (commented):
## Review: Track the processing list's head item for the opportunistic reclaim grace clock (#1746)

**Verdict:** Approve _(submitted as a `COMMENT` because I'm the PR author; GitHub won't accept a gating approval on one's own PR)_ — implements the issue's suggested direction precisely, with no blocking or should-fix findings.

### Correctness

Verified the orientation the whole mechanism hinges on: `ReserveNextAsync` does `LMOVE head→tail` (`Left`→`Right`), so the oldest, least-recently-reserved in-flight item sits at processing-list index 0; `PeekProcessingAsync` reads `LRANGE 0..count-1`, so `PeekProcessingAsync(1)` returns exactly that oldest item — the right one to age for staleness. `ReclaimProcessingAsync` pops tail→head, preserving order back onto the main queue. The in-memory fakes mirror this (append via `AddLast`, `Take` from the front), so the unit and integration suites exercise the same orientation as Redis.

The head-identity tracking is sound: the clock restarts whenever a different value occupies the head (the tracked item was acked/reclaimed and something else migrated up), and only fires once one item has genuinely dwelled past the grace period. Single-instance behavior is unchanged (the two prior reclaim tests pass unmodified), and the busy-churn case is now handled — which is the whole point of the issue.

### Test coverage

Strong and targeted:
- `ProcessQueue_DifferentItemReplacesProcessingHeadWithinGracePeriod_RestartsClockInsteadOfReclaiming` models the exact multi-instance churn shape from #1746 and is constructed to fail against the old occupancy-only clock — a genuine regression guard, not just a happy-path test.
- `PeekProcessingAsync_*` integration tests cover oldest-first ordering, non-destructive reads, non-positive count, and cancellation, matching the existing `PeekAsync` coverage.
- All `IPubSubQueue` fakes were updated to satisfy the new interface member (throwing `NotSupportedException` where the method isn't exercised, real implementations where it is) — consistent with the existing pattern in those files.

### Observation (non-actionable, no change needed)

Head identity is the serialized event payload, so there's a narrow ABA case: if the tracked head item is acknowledged and a **byte-identical** payload becomes the new oldest with no different item occupying the head in between, the comparison treats it as the same item and the old timestamp carries over, potentially reclaiming a freshly-arrived duplicate early. This is safe under the documented idempotent at-least-once contract (a redundant re-apply — the exact wastefulness this PR reduces, never a correctness issue), strictly better than the prior occupancy clock, and matches the identity-based direction the issue itself proposed. Fixing it would need a monotonic per-item sequence id, i.e. exactly the over-engineering #1745/#1746 consciously deferred. Noting it only so a future maintainer isn't surprised; no action warranted.

### What's good

Precise, minimal change that does exactly what #1746 asked and nothing more. The doc comments on the field/method and the `backend-persistence.md` update clearly explain *why* head-identity tracking beats the occupancy clock, which is the kind of context that keeps this mechanism maintainable. Nicely scoped.


---

PR #1916, opened 2026-07-13
Give admin reference-data load an error/retry state

### LINKED ISSUE #1913: Admin reference-data load has no retry affordance on failure
## Problem

Split out from #1896 (closed by #1912, which fixed the Workbench seed-load path). The companion gap that issue flagged as optional is still open:

`admin/+page.svelte`'s `onMount` calls `reference.load().catch(...)` (`UI/src/routes/admin/+page.svelte:80-84`). On failure it fires a single one-shot `toastError` and leaves `reference.loaded` at `false` with no error state and no retry affordance. The shared reference catalogues back every entity's select options, tag UI, and derived spawn shares, so a failed load degrades every workbench tool with no way to recover short of a full page reload — the same class of defect #1896/#1912 just fixed for the per-tool seed load.

## Suggested fix

Give `reference.load()` failure the same error/retry treatment now used by `Workbench.svelte` and `Progression.svelte`: store an `error`, surface it in the admin shell, and offer a Refresh button that re-invokes `reference.load()`. Add a test covering the failing-load path.

## Notes

- #1912 deliberately scoped itself to the required Workbench fix and left this out; see its "Not included" section.

ginderjeremiah:
## Summary

`admin/+page.svelte`'s `onMount` called `reference.load().catch(...)` and, on failure, fired a single one-shot `toastError` while leaving `reference.loaded` at `false` — every workbench tool stayed stuck on its loading spinner with no error message and no way to recover short of a full page reload.

`Workbench.svelte` and `Progression.svelte` already handle the identical class of failure (a failed seed/store load) correctly: catch, toast, store an `error`, and render a Refresh button. This PR brings the shared reference-catalogue load up to the same standard.

## Changes

- **`UI/src/routes/admin/+page.svelte`**: extracted the reference load into a `loadReference()` function that catches a failed `reference.load()`, stores the message in a new `loadError` state, and toasts it (mirroring `Workbench.svelte`/`Progression.svelte`). A new `referenceGate` snippet renders either that error state (message + Refresh button re-invoking `loadReference()`) or the existing `Loading` spinner, and replaces the three spots that previously fell back to an unconditional spinner while `reference.loaded` was false (Content Health, Progression, and the default entity-Workbench branch). The `DeadLetterConsole`/`SocketDeadLetterConsole` tools, which don't depend on the reference catalogues, are untouched.

## Tests

- **`UI/src/tests/routes/admin/admin-page.test.ts`** (new): renders the real page with `reference`'s real `$state`-backed `loaded` flag (only its transports — `fetchSocketData`/`ApiRequest` — are stubbed, mirroring `reference-load.test.ts`), the nav/entity modules stubbed to the default `'enemies'` tool, and every heavy child component (`Workbench`, `AdminSidebar`, the ops consoles, `Progression`) replaced with minimal stubs under `UI/src/tests/routes/admin/page/`. Asserts:
  - a failed load shows the error message + toast instead of spinning forever, with no tool content rendered
  - clicking Refresh retries the real load and renders the tool once it succeeds

## Test plan

- [x] `npx vitest run src/tests/routes/admin/admin-page.test.ts` — 1/1 passing
- [x] `npm run test:unit:ci` — full suite: 3657/3657 passing, coverage floors intact
- [ ] Backend unaffected (frontend-only change) — no backend verification needed

Closes #1913

---
_Generated by [Claude Code](https://claude.ai/code/session_01X3F8fAWzKHUqThm5YxbdQ5)_

### DIFF
--- a/UI/src/routes/admin/+page.svelte
+++ b/UI/src/routes/admin/+page.svelte
@@ -11,20 +11,20 @@
 				{#if reference.loaded}
 					<ContentHealthConsole />
 				{:else}
-					<Loading loading={true} delay={50} />
+					{@render referenceGate()}
 				{/if}
 			{:else if active === PROGRESSION_TOOL_KEY}
 				{#if reference.loaded}
 					<Progression />
 				{:else}
-					<Loading loading={true} delay={50} />
+					{@render referenceGate()}
 				{/if}
 			{:else if reference.loaded && activeEntity}
 				{#key active}
 					<Workbench entity={activeEntity} groupLabel={groupLabelFor(active)} />
 				{/key}
 			{:else}
-				<Loading loading={true} delay={50} />
+				{@render referenceGate()}
 			{/if}
 		</div>
 
@@ -38,6 +38,17 @@
 		/>
 	</div>
 {/if}
+
+{#snippet referenceGate()}
+	{#if loadError}
+		<div class="admin-load-error" role="alert" data-testid="admin-reference-error">
+			<p>{loadError}</p>
+			<button type="button" class="btn" onclick={loadReference}>Refresh</button>
+		</div>
+	{:else}
+		<Loading loading={true} delay={50} />
+	{/if}
+{/snippet}
 
 <script lang="ts">
 import { onMount } from 'svelte';
@@ -70,17 +81,27 @@ let sidebarPinned = $state(false);
 // (no token available) shows nothing, and a non-admin who deep-links here is redirected before the
 // workbench mounts or its reference data loads.
 let authorized = $state(false);
+let loadError = $state<string | null>(null);
 
 const activeEntity = $derived(entityByKey(active));
+
+async function loadReference() {
+	loadError = null;
+	try {
+		await reference.load();
+	} catch (ex) {
+		const message = ex instanceof Error ? ex.message : 'Failed to load admin reference data.';
+		loadError = message;
+		toastError(message);
+	}
+}
 
 // Guard the route to Admins, then load the shared reference catalogues (used by every entity's
 // select options, tag UI, and derived spawn shares) before rendering any workbench.
 onMount(() => {
 	authorized = ensureAdminAccess();
 	if (authorized && !reference.loaded) {
-		reference.load().catch((ex) => {
-			toastError(ex instanceof Error ? ex.message : 'Failed to load admin reference data.');
-		});
+		void loadReference();
 	}
 });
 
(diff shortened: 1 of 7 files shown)


UI/src/routes/admin/+page.svelte
```diff
@@ -141,4 +162,44 @@ $effect(() => {
...
+	text-align: center;
+
+	p {
+		max-width: 480px;
+		color: var(--error);
+		font-size: 13px;
+	}
+}
+
+.btn {
```
ginderjeremiah, line 184: Nit (non-blocking): this `.btn` base style is now copy-pasted verbatim into a third admin file — it's byte-identical to the `.btn` base in `Progression.svelte` and `Workbench.svelte`, and the whole error block (`role="alert"` + `<p>{error}</p>` + Refresh `.btn`) is the third near-identical copy of `.workbench-error`/`.prog-error`. It matches the established sibling pattern, so it's fine to land as-is, but the admin surface is accumulating a duplicated error/retry affordance that would be a good tech-debt follow-up to extract into a shared component (e.g. an `AdminLoadError` with the Refresh button, or a shared admin `.btn`). Worth an issue rather than expanding this PR's scope.

---

ginderjeremiah (commented):
## Review: Give the admin reference-data load an error/retry state (Closes #1913)

**Verdict:** Approve _(submitted as a comment — GitHub won't accept a gating verdict on one's own PR)_ — the change solves #1913 cleanly by bringing `reference.load()` failure up to the exact error/retry standard already used by `Workbench.svelte` and `Progression.svelte`, with solid test coverage. One non-blocking nit.

### What it does
Extracts the reference load into `loadReference()` that catches, stores a new `loadError` state, and toasts (mirroring the siblings). A `referenceGate` snippet renders either the error state (message + Refresh re-invoking `loadReference()`) or the existing `Loading` spinner, and replaces the three unconditional-spinner fallbacks (Content Health, Progression, default entity Workbench). The DeadLetter consoles are correctly left untouched since they don't depend on the reference catalogues.

### Correctness
- Retry is safe: `reference.load()` builds everything via `Promise.all` and only flips `loaded = true` after all assignments, so a failure commits nothing partial and re-invoking it is idempotent.
- `loadReference()` resets `loadError = null` up front, so a retry correctly falls back to the spinner while re-fetching and re-shows the error only if it fails again.
- `void loadReference()` in `onMount` matches the fire-and-forget pattern; the gate reacts off `$state`.
- Error UI is theme-compliant — `var(--error)`, `var(--border-light)`, `var(--text-secondary)`, `color-mix(... var(--white) ...)`; no hardcoded colors (frontend.md § Styling).

### Nits
- `.btn` base style + the error block are the third verbatim copy across the admin surface (`Workbench.svelte`, `Progression.svelte`, now here). Consistent with siblings, so fine to land — but a good tech-debt follow-up to extract a shared admin error/button component. Inline comment on `+page.svelte:184`.

### Test coverage
Good and following the guidelines. `admin-page.test.ts` keeps `reference` (and its `$state` `loaded`) real — stubbing only the transports (`fetchSocketData`/`ApiRequest`), mirroring `reference-load.test.ts` — so the fix's reactivity is genuinely exercised, and stubs the heavy children. It asserts both halves: a failed first `GetEnemies` shows the error + toast with no tool content, and clicking Refresh retries the real load and renders the tool. The added stub components are appropriately minimal.

### What's good
Tight, well-scoped fix that closes a real UX gap by reusing an established in-repo pattern rather than inventing a new one; the reference-gate snippet keeps the three call sites DRY within the file, and the test exercises the real store rather than mocking the behavior under test.


---

PR #1918, opened 2026-07-13
Move ApiSocket's connection onto instance state (#1770)

### LINKED ISSUE #1770: ApiSocket holds its connection in a module-level variable while everything else is instance state (split-brain; standing test flake source)
## Problem

`UI/src/lib/api/api-socket.ts:56` — the WebSocket itself is a module-level `socket` variable, while the request queue, in-flight map, retry budget, and keepalive are all instance state on the `ApiSocket` class (singleton exported at `:410`). Tests construct per-test instances that all share the one module-level socket — a standing flake source, and a split-brain design where two instances would silently fight over one connection.

## Fix

Move `socket` onto the instance (or collapse the whole module to functions — either way, one owner for all the state). Mechanical.

ginderjeremiah:
## Summary

`api-socket.ts` held the live `WebSocket` in a module-level `socket` variable while every other piece of connection state (request queue, in-flight map, retry budget, keepalive timers) lived on the `ApiSocket` instance. Two `ApiSocket` instances would have silently fought over the one module-level connection, and the test suite's per-test instances were only kept from interfering with each other by an explicit `beforeEach` workaround that forced any leftover socket to read as `CLOSED`.

## Changes

- `UI/src/lib/api/api-socket.ts`: moved `socket` from a module-level `let` into `private socket?: WebSocket` on the `ApiSocket` class; updated every read/write site (`ensureSocket`, `openSocket`, `attemptPing`, `disconnect`, `processCommandQueue`, `receiveResponse`) to use `this.socket`.
- `UI/src/tests/lib/api/api-socket.test.ts`: removed the now-unnecessary global "force any leftover socket closed" workaround from the top-level `beforeEach` (each freshly-constructed `ApiSocket` now owns its own `undefined` socket, so there's nothing to leak between per-test instances). The module-level `apiSocket` singleton used by `fetchSocketData` genuinely does persist across tests in that describe block, so a narrower, scoped version of the same reset was kept just there, backed by a small `allCreatedSockets` list that (unlike `socketInstances`) isn't cleared per test.

No behavior change for production code paths — this is purely an ownership fix plus a corresponding test-suite cleanup.

## Test plan

- [x] `npx vitest run src/tests/lib/api/api-socket.test.ts` — 51/51 passing
- [x] `npm run test:unit:ci` (full frontend suite) — 3657/3657 passing

Closes #1770

---
_Generated by [Claude Code](https://claude.ai/code/session_01PW5e8HMHfGaHqN8ajSUefv)_

### DIFF
--- a/UI/src/lib/api/api-socket.ts
+++ b/UI/src/lib/api/api-socket.ts
@@ -59,8 +59,6 @@ export interface IApiSocketResponse<T extends ApiSocketCommand | void = void> {
 	data?: T extends ApiSocketCommand ? ApiSocketResponseTypes[T] : never;
 }
 
-let socket: WebSocket;
-
 const errorHook = createHook<[string]>();
 export const onSocketError = errorHook.onNotified;
 
@@ -76,6 +74,9 @@ type InFlightRequest = {
 };
 
 export class ApiSocket {
+	// Instance state, not module-level: each ApiSocket owns exactly one connection, so two instances can
+	// never silently fight over the same socket.
+	private socket?: WebSocket;
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- heterogeneous queue of requests for differing commands; ApiSocketRequest<T> is invariant in T so a common supertype isn't expressible.
 	private socketCommandQueue: ApiSocketRequest<any>[] = [];
 	// Keyed by command id so a response is an O(1) lookup and a settled request can be pruned (rather
@@ -113,7 +114,7 @@ export class ApiSocket {
 	 * race to create (and leak) two sockets.
 	 */
 	private ensureSocket(): Promise<void> {
-		if (socket && socket.readyState !== socket.CLOSED) {
+		if (this.socket && this.socket.readyState !== this.socket.CLOSED) {
 			return Promise.resolve();
 		}
 		if (this.connecting) {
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Move ApiSocket's connection onto instance state (#1770)

**Verdict:** Approve — clean, correct, behavior-preserving refactor that resolves the split-brain ownership issue exactly as the issue describes. _(Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR; the real call is Approve.)_

The `socket` field moves from a module-level `let` onto `private socket?: WebSocket`, and every read/write site (`ensureSocket`, `openSocket`, `attemptPing`, `disconnect`, `processCommandQueue`, `receiveResponse`) now goes through `this.socket`. I checked the full file — there are no remaining references to the old module-level variable, and no other file could have depended on it since it was never exported.

### What's good

- Correctly closes the split-brain design: each `ApiSocket` now owns exactly one connection, so per-test instances (and any hypothetical second instance) can't fight over a shared socket.
- The `this.socket?.send('pong')` optional-chaining in `receiveResponse` is a safe, slightly-more-defensive change — that branch only runs from `socket.onmessage`, so the socket is present in practice, and the `?.` just avoids a throw in the pathological post-reconnect case rather than changing real behavior.
- The test cleanup is the sharp part and it's reasoned correctly: the top-level `beforeEach` reset is genuinely unnecessary now (each per-test `new ApiSocket()` owns its own `undefined` socket), while the module-level `apiSocket` singleton that `fetchSocketData` drives really does persist across tests — so the narrower reset backed by `allCreatedSockets` (which, unlike `socketInstances`, isn't cleared per test) is kept scoped to just that block. Outer-to-inner `beforeEach` ordering means `socketInstances` is already cleared by the time the inner hook runs, so `allCreatedSockets` is the right list to reach for.

### Test coverage

Adequate for a pure ownership refactor. Not battle logic, so no frontend/backend mirroring applies. The existing 51-test `api-socket` suite validates that behavior is unchanged, and the PR reports the full frontend suite green. The comments explaining the test-harness timing are appropriately detailed for non-obvious lifecycle reasoning — no concern there.

No blocking findings, no should-fixes, no nits. I'll merge once CI finishes (`test-frontend` and `test-e2e (webkit)` are still running).

---

PR #1919, opened 2026-07-13
Small frontend cleanups: encodeURIComponent, dead optional-chain, per-player attrMode key, battler typo

### LINKED ISSUE #1772: Small frontend cleanups: window.encodeURIComponent, dead optional-chain, localStorage key prefix violation, public-signature typo
Grouped fix-in-passing nits from a health review — none worth an issue alone:

- `UI/src/lib/api/api-request.ts:168` — `window.encodeURIComponent` instead of the global `encodeURIComponent`; forces test `window` stubs and breaks non-window contexts.
- `UI/src/lib/api/api-response.ts:56` — dead `?.` on a value that can't be nullish at that point.
- `UI/src/routes/game/screens/attributes/attributes-view.svelte.ts` — persists under key `ttf.attr.mode`, breaking the `gameserver.*` localStorage prefix convention used everywhere else.
- `UI/src/lib/battle/battler.ts:135,389` — `additionalAtttributes` (triple-t typo) in a public signature; rename touches its call/test sites.

ginderjeremiah:
## Summary

Four grouped fix-in-passing nits from a health review (#1772), none worth an issue alone:

- **`UI/src/lib/api/api-request.ts:178`** — `window.encodeURIComponent` → the global `encodeURIComponent`. The `window.`-qualified form forced a `window` stub in tests and would break in any non-window context.
- **`UI/src/lib/api/api-response.ts:56`** — dropped the dead `?.` on `responseJson`, which is never nullish at that point (`responseJson`'s getter always returns an object via `??=`).
- **`UI/src/routes/game/screens/attributes/attributes-view.svelte.ts`** — the attributes density mode was persisted under the bare key `ttf.attr.mode` (a leftover from the original design mockup), breaking the `gameserver.*` localStorage prefix convention used everywhere else, and it was **device/account-wide** despite `docs/frontend-screens.md` documenting it as "persisted per player." Fixed both: the key is now `gameserver.attrMode.<playerId>`, so each character keeps its own mode instead of inheriting or silently overwriting another character's on the same account.
- **`UI/src/lib/battle/battler.ts:135,389`** — renamed the `additionalAtttributes` (triple-t typo) constructor/`reset` parameter to `additionalAttributes`. Positional-only parameter, so no other call sites reference the name.

## Test plan

- [x] `npx vitest run` (UI) — 3658/3658 passing, including a new case pinning that two different player ids keep independent `attrMode` keys.
- [x] Verified no remaining references to `ttf.attr.mode` or `additionalAtttributes` anywhere in the frontend source.

Closes #1772

---
_Generated by [Claude Code](https://claude.ai/code/session_01CCPPaDTwCHUuequ7n7NNv2)_

### DIFF
--- a/UI/src/lib/api/api-request.ts
+++ b/UI/src/lib/api/api-request.ts
@@ -175,7 +175,7 @@ export class ApiRequest<U extends ApiEndpoint> {
 
 		return keys(urlParams)
 			.filter((key) => urlParams[key] !== undefined)
-			.map((key) => key + '=' + window.encodeURIComponent(urlParams[key]))
+			.map((key) => key + '=' + encodeURIComponent(urlParams[key]))
 			.join('&');
 	}
 }
--- a/UI/src/lib/api/api-response.ts
+++ b/UI/src/lib/api/api-response.ts
@@ -53,7 +53,7 @@ export class ApiResponse<T extends ApiResponseType> {
 	}
 
 	public get error() {
-		return this.responseJson?.errorMessage || this.#raw.statusText || 'Failed to communicate with server.';
+		return this.responseJson.errorMessage || this.#raw.statusText || 'Failed to communicate with server.';
 	}
 
 	public get responseText() {
--- a/UI/src/lib/battle/battler.ts
+++ b/UI/src/lib/battle/battler.ts
@@ -132,7 +132,7 @@ export class Battler {
 
 	constructor(
 		battlerData?: BattlerData,
-		additionalAtttributes?: IBattlerAttribute[],
+		additionalAttributes?: IBattlerAttribute[],
 		grantedSkillIds?: number[],
 		additionalModifiers?: AttributeModifier[],
 		equippedWeaponType?: EDamageType,
@@ -140,7 +140,7 @@ export class Battler {
 	) {
 		this.reset(
 			battlerData,
-			additionalAtttributes,
+			additionalAttributes,
 			grantedSkillIds,
 			additionalModifiers,
 			equippedWeaponType,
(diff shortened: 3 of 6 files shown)


ginderjeremiah (commented):
## Review: Four grouped frontend cleanups (closes #1772)

**Verdict:** Approve — all four items from the issue are implemented correctly, and the per-player key change also folds in the follow-up comment on #1772. (Posted as a comment because I'm the PR author; treat it as an approval.)

Verified each change against the source:

### What's good
- **`api-request.ts:178`** — `window.encodeURIComponent` → global `encodeURIComponent`, and the matching `window` stub is dropped from `api-request.test.ts`, removing the exact test coupling the issue called out.
- **`api-response.ts:56`** — the `?.` was genuinely dead: `responseJson` (getter, `api-response.ts:63`) resolves via `??= this.parseJson()`, and `parseJson()` always returns an object on every branch. The change also makes `error` consistent with its sibling getters `ok` (`:38`) and `data` (`:48`), which already read `this.responseJson.errorMessage` without `?.`.
- **`attributes-view.svelte.ts`** — key now `gameserver.attrMode.<playerId>`, fixing both the `gameserver.*` prefix violation and the account-wide scoping the issue comment flagged against `docs/frontend-screens.md`'s "persisted per player." `modeStorageKey()` reads `playerManager.id` lazily on each get/set, so it's correct across mid-session character switches. Good comments explaining the why.
- **`battler.ts`** — `additionalAtttributes` → `additionalAttributes`; positional-only param, so no external call sites to update.

### Test coverage
- New test pins that two player ids keep independent `attrMode` keys and don't overwrite each other — directly covers the per-player behavior that motivated the change.
- Existing persistence test updated to the new key; `mockPlayerManager.id` seeded and reset in `beforeEach`.
- No backend/battle-consistency concern here — the `battler.ts` change is a pure parameter rename with no logic change, so no mirrored backend test is needed.

Clean, well-scoped tech-debt PR. No blocking, should-fix, or nit findings.

---

PR #1937, opened 2026-07-13
Keep staticData.classes raw instead of sentinel-widened

### LINKED ISSUE #1783: Workbench: class.ts writes sentinel-widened records into the shared staticData.classes cache
## Problem

`UI/src/routes/admin/workbench/entities/class.ts:23-24` stores the sentinel-widened `WorkbenchClass[]` (`passiveScalingAttributeId: -1` for "none") into `staticData.classes`, while `reference.svelte.ts:101` stores raw `IClass[]` — so the cache's shape depends on which loader ran last. The deliberate pattern is the opposite: `item.ts:20-27` explicitly keeps `staticData.items` raw ("null = none) for the game/other screens"), as does `lesson.ts:15-16`.

Latent today (character-create reads its own `Login/CharacterCreationData` payload), but any future consumer doing a `!= null` check on `passiveScalingAttributeId` will misread `-1` as a real `EAttribute`.

## Fix

Keep `staticData.classes` raw and map only the returned workbench copies, mirroring `item.ts`.

ginderjeremiah:
## Summary

`class.ts`'s `refresh()` was writing the sentinel-widened `WorkbenchClass[]` (`passiveScalingAttributeId: -1` for "none") into the shared `staticData.classes` cache, while `reference.svelte.ts`'s `load()` writes raw `IClass[]` there — so the cache's shape depended on which loader ran last. The deliberate pattern elsewhere (`item.ts`) is the opposite: keep `staticData.*` raw for the game/other screens and only widen the copy returned to the editor.

This was latent (nothing currently does a `!= null`/truthiness check on `passiveScalingAttributeId` off `staticData.classes`), but any future consumer would misread `-1` as a real `EAttribute`.

## Change

- `class.ts`: `staticData.classes` is now assigned the raw fetched array; the `-1` sentinel is applied only to the array returned from `refresh()` (used by the Workbench editor).
- Updated `class.test.ts` to assert the new split: `staticData.classes` is the raw socket payload (by reference), and the sentinel normalization only shows up in `refresh()`'s return value.

Verified the one consumer of `staticData.classes` that isn't the game/creation flow — `computeReferences` in `references.ts` (retire-confirm's "referenced by" surface) — only reads `starterSkillIds`/`starterEquipment` off classes, never `passiveScalingAttributeId`, so this doesn't affect that surface.

## Test plan

- [x] `npx vitest run` on the affected Workbench test files — all pass
- [x] `npm run test:unit:ci` (full frontend suite) — 3658 passed, 0 failed

Closes #1783

---
_Generated by [Claude Code](https://claude.ai/code/session_01UfuLJ6T4hzF7hBF6HdTCL9)_

### DIFF
--- a/UI/src/routes/admin/workbench/entities/class.ts
+++ b/UI/src/routes/admin/workbench/entities/class.ts
@@ -23,14 +23,14 @@ export interface WorkbenchClass extends Omit<IClass, 'passiveScalingAttributeId'
 }
 
 // Classes load over the socket; the admin filter invalidates this cache on every write server-side, so a
-// plain refetch returns the freshly-saved list. The optional scaling attribute is normalised to the select's
-// "None" sentinel (-1) for the editable copy. Written through to staticData.classes so retire-confirm's
-// reference computation (starter-skill/starter-equipment groups) sees post-save edits (#1633).
+// plain refetch returns the freshly-saved list. Written through to staticData.classes so retire-confirm's
+// reference computation (starter-skill/starter-equipment groups) sees post-save edits (#1633). staticData.classes
+// stays raw (undefined = none) for the game/other screens, mirroring item.ts; the optional scaling attribute is
+// normalised to the select's "None" sentinel (-1) only in the returned editable copy.
 const refresh = async (): Promise<WorkbenchClass[]> => {
 	const classes = await fetchSocketData('GetClasses');
-	const workbenchClasses = classes.map((c) => ({ ...c, passiveScalingAttributeId: c.passiveScalingAttributeId ?? -1 }));
-	staticData.classes = workbenchClasses;
-	return workbenchClasses;
+	staticData.classes = classes;
+	return classes.map((c) => ({ ...c, passiveScalingAttributeId: c.passiveScalingAttributeId ?? -1 }));
 };
 
 export const classEntity: EntityConfig<WorkbenchClass> = {
--- a/UI/src/tests/routes/admin/workbench/entities/class.test.ts
+++ b/UI/src/tests/routes/admin/workbench/entities/class.test.ts
@@ -56,13 +56,13 @@ describe('classEntity.refresh', () => {
 			}
 		];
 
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: keep `staticData.classes` raw instead of sentinel-widened (#1783)

**Verdict: Approve** — the change does exactly what the issue asks and matches the established `item.ts` pattern precisely.

_(Submitted as a COMMENT because I'm the PR author and GitHub won't accept a gating verdict on my own PR — treat this as "approve".)_

### What's good

- The fix is the minimal correct one: `staticData.classes` is now assigned the raw `IClass[]` payload and the `-1` "None" sentinel is applied only to the array returned from `refresh()` for the editor. This mirrors `item.ts:17-28` (and `reference.svelte.ts:111`, which already wrote raw), so the cache shape no longer depends on which loader ran last.
- The rewritten comment on `class.ts:25-29` accurately documents the split ("`staticData.classes` stays raw (undefined = none)…normalised…only in the returned editable copy") and correctly says `undefined`, matching `IClass`'s optional field (the test asserts `toBeUndefined()`).
- Verified no consumer regresses: grepping `references.ts` for `passiveScalingAttributeId`/`.classes` returns nothing, so `computeReferences` (retire-confirm's "referenced by" surface) only reads `starterSkillIds`/`starterEquipment` — unaffected. The other `staticData.classes` reads (`reference-data.ts` ref-source getter, `retire-confirm.ts`) don't touch the scaling attribute either.

### Test coverage

Appropriate and correctly updated. The two `refresh` tests now assert the split cleanly: the write-through test checks `staticData.classes` is the raw socket payload by reference (`toBe(socket.classes)`), and the normalization test asserts the `-1` sentinel appears in `refresh()`'s return value while `staticData.classes[0].passiveScalingAttributeId` stays `undefined`. That's exactly the invariant the fix establishes.

No blocking findings, no should-fixes, no nits. Clean tech-debt fix.

---

PR #1938, opened 2026-07-13
Anchor StartBossBattle to any in-effect cooldown, not just an abandon's

### LINKED ISSUE #1920: ChallengeBoss anchors mid-cooldown after a normal victory claim — the #1884 anchoring only covers abandon-resolved outcomes
`BattleService.StartBossBattle` anchors the new boss battle with `var now = abandonedOutcomeCooldown ?? DateTime.UtcNow;` (`Game.Application/Services/BattleService.cs:243`), and `abandonedOutcomeCooldown` is only set when *this call's own* `AbandonBattle` incurred a fresh cooldown.

A scripted `ChallengeBoss → DefeatEnemy → ChallengeBoss` loop never trips that: `EndBattleVictory` sets `state.SetCooldown(battleCompletedAt + PostBattleCooldown)` and clears the battle, so the next `StartBossBattle` sees `HasActiveBattle == false`, skips the abandon block entirely, and anchors the boss battle at `DateTime.UtcNow` — mid-cooldown. `ClaimedBeforeBattleCouldFinish` then keys off that too-early `BattleStartTime`, so a tampered client can claim each boss kill a full 5s cooldown early. Unlike `NewEnemy` (gated on `state.IsOnCooldown` at the command level), `ChallengeBoss` deliberately has no command-level gate — pacing is supposed to come from anchoring, and the response's `Cooldown` handshake (`ChallengeBoss.cs:56-61`) already tells an honest client to wait, but the server-side anchor doesn't hold the same line. The same gap applies to the challenge-mid-cooldown case the code comment itself describes (`BattleService.cs:224-227`): the abandon resolves no outcome, and the new battle still anchors at `UtcNow` inside the running cooldown.

Net effect: live boss scripting pays only battle duration per kill instead of duration + cooldown, beating the authoritative offline auto-boss simulator (which paces with `CooldownMs` between battles).

**Fix direction:** anchor to the in-effect cooldown generally, e.g. `abandonedOutcomeCooldown ?? (state.EnemyCooldown > DateTime.UtcNow ? state.EnemyCooldown : DateTime.UtcNow)` — exactly the contract the response's `Cooldown` handshake already promises. Covers both the post-victory and the mid-cooldown-challenge cases; #1884's abandon-specific anchoring becomes a special case of it.

ginderjeremiah:
## Summary

`BattleService.StartBossBattle` anchored the replacement boss battle's start (`var now = abandonedOutcomeCooldown ?? DateTime.UtcNow;`) only when *this call's own* abandon resolved a win/loss/draw and incurred the post-battle cooldown (`#1884`/`#1906`). That left two cases falling straight through to `DateTime.UtcNow`, inside a still-running cooldown:

- **A scripted `ChallengeBoss → DefeatEnemy → ChallengeBoss` loop.** `EndBattleVictory` sets the cooldown and clears the battle, so the next `StartBossBattle` sees `HasActiveBattle == false`, skips the abandon block entirely, and anchors at `now` — mid-cooldown.
- **A challenge landing mid an already-running cooldown with nothing to abandon.** The abandon resolves no outcome (`elapsedMs` clamps to 0), so `EnemyCooldown` is untouched but wasn't picked up as the anchor either.

Net effect: live boss scripting paid only battle duration per kill instead of duration + cooldown, beating the offline auto-boss simulator's pacing.

## Changes

- **`Game.Application/Services/BattleService.cs`**: `StartBossBattle` now anchors to `state.EnemyCooldown` whenever `state.IsOnCooldown(now)` is true, regardless of whether this call's own abandon set it — exactly the contract the response's `Cooldown` handshake already promises. The previous `abandonedOutcomeCooldown` tracking (which only captured the abandon-incurred case) is removed as redundant.
- **`Game.Api/Sockets/Commands/ChallengeBoss.cs`**: updated comments describing the anchoring to match the broader coverage.
- **`UI/src/lib/engine/battle/enemy-manager.ts`**: updated a comment describing when the server can return a positive cooldown on `ChallengeBoss` — no behavioral change, since the client already waits out any positive cooldown value generically.
- **`docs/backend-battle.md`**: updated the abandon/cooldown-anchor bullet to describe the general anchor.

## Tests

- `BattleServiceIntegrationTests.cs`:
  - Flipped `StartBossBattle_AbandonResolvesNoOutcomeWithLeftoverCooldown_DoesNotAnchorToTheStaleCooldown` → `...AnchorsToTheLeftoverCooldown`, since its old assertions pinned exactly the bug this PR fixes.
  - Added `StartBossBattle_NoActiveBattleButCooldownStillRunning_AnchorsToTheCooldown`, covering the `HasActiveBattle == false` case (the normal-victory leg of the loop).
- `ChallengeBossSocketTests.cs`: added `ChallengeBoss_ReChallengedWhileANormalVictorysCooldownIsStillRunning_PacesWithACooldown`, a socket-level regression test for the same scenario.

## Test plan

- [x] `dotnet test` — full backend suite: 3119/3119 passing.
- [x] `npm run test:unit:ci` — full frontend suite: 3658/3658 passing.

Closes #1920

---
_Generated by [Claude Code](https://claude.ai/code/session_01LjLRJ9CdaVwttZ1n6T5L8i)_


ginderjeremiah (commented):
## Review: Anchor `StartBossBattle` to any in-effect cooldown (#1920)

**Verdict:** Approve — the fix implements exactly the contract the `Cooldown` handshake already promises, closes the boss-scripting pacing gap, and is well covered on both the service and socket layers. *(Posted as a comment because GitHub won't accept a gating verdict from the PR author; treat this as an approval.)*

I verified the change against the issue and the `docs/backend-battle.md` anchoring contract, and read both battle-start paths in `BattleService.cs`.

### Correctness
- `state.IsOnCooldown(utcNow) ? state.EnemyCooldown : utcNow` (`BattleService.cs:235`) is exactly `EnemyCooldown > now ? EnemyCooldown : now` (`PlayerState.IsOnCooldown` is `EnemyCooldown > now`), i.e. the issue's suggested fix direction. It subsumes the old `abandonedOutcomeCooldown` special case (an abandon that resolves an outcome sets `EnemyCooldown` to a future value, so `IsOnCooldown` is true and it still anchors), and additionally covers the two leaked cases: the normal-victory `HasActiveBattle == false` leg and the mid-cooldown challenge that resolves no outcome.
- The now-removed `cooldownBeforeAbandon` / `abandonedOutcomeCooldown` tracking is genuinely dead here — no dangling references remain in the boss path (the sibling `StartBattle` keeps its own copy, which is correct, see below). No stale cooldown risk: only a *future* `EnemyCooldown` anchors, so a long-past leftover falls through to `now`.

### Layering / consistency
- `StartBattle` (NewEnemy) deliberately keeps `scheduledStartTime ?? abandonedOutcomeCooldown ?? now`, and that's correct rather than an inconsistency: it's gated on `IsOnCooldown` at the command level (so it can never reach the abandon block with a pre-existing running cooldown) and it carries the prefetch `scheduledStartTime` precedence that the boss path has no equivalent for. The divergence is justified and the comments on both methods explain why.

### Test coverage
- Backend service integration: the flipped `...AnchorsToTheLeftoverCooldown` correctly inverts the old assertion (`BattleStartTime == leftoverCooldown` instead of `<`), and the new `...NoActiveBattleButCooldownStillRunning_AnchorsToTheCooldown` covers the normal-victory leg (`HasActiveBattle == false`). Both assert exact-equality anchoring, which matches the implementation.
- Socket-level regression `ChallengeBoss_ReChallengedWhileANormalVictorysCooldownIsStillRunning_PacesWithACooldown` exercises the full #1920 loop end-to-end and asserts a positive returned `Cooldown`.
- No frontend battle-test mirror is required: this is server-authoritative pacing, and the `enemy-manager.ts` change is comment-only — the client already waits out any positive cooldown generically, so there's no new simulation logic to mirror.

### Docs
- `docs/backend-battle.md` and the `ChallengeBoss.cs` / `enemy-manager.ts` comments are updated to describe the general anchor instead of the abandon-only one. Accurate.

---

PR #1753, opened 2026-07-08
Harden content-fixture tests: vacuous-pass guard, $content alias, anchorKey coverage

ginderjeremiah:
## Summary

Follow-up from PR #1716 (closes #1673), addressing the two minor hardenings it flagged plus the "worth confirming" item:

1. **Vacuous-pass guard.** `screen-defs.test.ts`'s `screenKey` check now asserts at least one live lesson exists before checking them, so an empty/fully-retired `content/lessons.json` fails loudly instead of passing with zero coverage.
2. **Reusable content-import pattern.** Added a `$content` SvelteKit alias (`kit.alias` in `svelte.config.js`, pointing at the repo-root `content/` folder) and switched the deep six-level relative import to it. Documented the pattern in `docs/frontend.md`'s testing guidelines so future content-reading tests reuse it.
3. **anchorKey coverage.** Confirmed the frontend-only `anchorKey` validation referenced in #1592 ("A test asserting every anchorKey referenced by seeded lessons resolves to a registered anchor") didn't exist yet, so I added it: every live lesson step's `anchorKey` must resolve to a real `use:tutorialAnchor` registration.

   Doing that surfaced a small DRY gap: the three `use:tutorialAnchor` call sites (`CombatFloaters`, `BattlerCard`, `Skills`) each built their `fight-*-${side}` key inline, so a test would either have had to duplicate those literals (drift risk) or parse Svelte templates. Instead I added a canonical `TOUR_ANCHOR_KEY` builder registry in `tutorial-anchor.ts` that both the three call sites and the new test's `TOUR_ANCHOR_KEYS` list build from — matching the existing pattern where `GAME_SCREENS` is the single source of truth the `screenKey` check validates against.

## Test plan

- [x] `npm run lint` (eslint + svelte-check) — clean.
- [x] `npm run test:unit:ci` — full suite passes, 3545/3545.
- Verified the `$content` alias resolves for both TypeScript (`svelte-kit sync` regenerates `.svelte-kit/tsconfig.json`) and Vitest.

Closes #1719

---
_Generated by [Claude Code](https://claude.ai/code/session_013ZQQsEZEppLaiAcTf8PzML)_

PR #1940, opened 2026-07-13
Assert every tour anchor key resolves to a live rendered DOM anchor

### LINKED ISSUE #1788: Assert every tour anchor key resolves to a live rendered DOM anchor
Follow-up from PR #1753.

That PR added a canonical `TOUR_ANCHOR_KEY` builder registry in `tutorial-anchor.ts` and a test asserting that every `anchorKey` referenced by seeded lessons resolves to a **registered** key. The remaining gap: registry completeness is currently enforced by convention only — the test checks *lesson content ⊆ registry*, not *registry ⊆ actually-rendered DOM anchors*.

A stronger test would mount the fight components (`CombatFloaters`, `BattlerCard`, `Skills`) and assert each `TOUR_ANCHOR_KEYS` entry resolves to a live `use:tutorialAnchor` registration, closing the loop so a renamed/removed anchor site can't silently drift from the registry.

Genuinely optional hardening — the current check is a reasonable content-quality guard. No user-facing effect.

ginderjeremiah:
## Summary

Follow-up from PR #1753, which added the canonical `TOUR_ANCHOR_KEY` builder registry in `tutorial-anchor.ts` plus a test asserting every `anchorKey` referenced by seeded lessons resolves to a **registered** key (`screen-defs.test.ts`'s "committed lesson content" suite). That only checks *lesson content ⊆ registry* — the other direction, *registry ⊆ actually-rendered DOM anchors*, was unenforced, so a renamed/removed `use:tutorialAnchor` call site could silently drift from `TOUR_ANCHOR_KEYS` with nothing failing (a missing anchor just degrades gracefully to a centered callout at runtime).

## Change

Added a test to `BattlerCard.test.ts` (which already carries the mocks needed to render the fight subtree) that mounts one `BattlerCard` per side — `BattlerCard` composes both `CombatFloaters` and `Skills`, so this covers all three `use:tutorialAnchor` call sites for both sides — and asserts every entry of `TOUR_ANCHOR_KEYS` resolves to a live registration via `getTutorialAnchor`.

No production code changed — this is purely closing a test-coverage gap.

## Test plan

- [x] `npx vitest run src/tests/routes/game/screens/fight/BattlerCard.test.ts` — 14/14 passing (13 existing + 1 new)
- [x] `npm run test:unit:ci` — full suite: 3659/3659 passing, coverage floors intact
- Backend unaffected (frontend test-only change) — no backend verification needed

Closes #1788

---
_Generated by [Claude Code](https://claude.ai/code/session_01LeAKj5tqoM3938R7e6i1C1)_

### DIFF
--- a/UI/src/tests/routes/game/screens/fight/BattlerCard.test.ts
+++ b/UI/src/tests/routes/game/screens/fight/BattlerCard.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, afterEach, vi } from 'vitest';
 import { render, cleanup } from '@testing-library/svelte';
 import { EAttribute, type IAttribute } from '$lib/api';
-import { getTutorialAnchor } from '$components';
+import { getTutorialAnchor, TOUR_ANCHOR_KEYS } from '$components';
 
 // The card renders a Skills -> SkillTooltip subtree, which imports the battle engine and the
 // reference-data store at module load; both are mocked even though no hover happens here. The player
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: closes the `registry ⊆ rendered DOM` half of the tour-anchor coverage (#1788)

**Verdict:** Approve _(posted as a comment — I'm the PR author, so GitHub won't take a gating event from me)_ — the new test correctly and completely closes the direction #1788 asked for, with no production code touched.

I verified the change against the source rather than the description:

- All three `use:tutorialAnchor` call sites register through the canonical builders — `CombatFloaters` → `TOUR_ANCHOR_KEY.fightCombatLog`, `BattlerCard` → `fightHpBar`, `Skills` → `fightSkillBar` — the same builders `TOUR_ANCHOR_KEYS` is derived from, so the registry and the render sites share one source of truth.
- `BattlerCard.svelte` renders both `CombatFloaters` (line 2) and `Skills` (line 49) unconditionally for either side, so mounting one card per side genuinely registers all six keys — the "covers all three call sites for both sides" claim holds.
- `TOUR_ANCHOR_KEYS` currently contains only the fight-screen keys, so asserting each resolves after both mounts makes the check complete today, and it doubles as a tripwire: a future key added to the registry without a matching render site will fail here, exactly the drift the issue wanted caught.
- Registration cleanup is sound (`tutorialAnchor.destroy` deletes on unmount, guarded against overlapping mounts), and `afterEach(cleanup)` clears the shared `anchors` map between tests, so the pass isn't riding on leaked state from an earlier test.

### Test coverage

Adds one focused test to the existing `BattlerCard.test.ts`, reusing the mocks already in place — appropriate placement per the frontend testing rules, and no other suite needs to change since the assertion is pure existence/liveness (the sibling test at line 166 already pins the HP-bar key to its specific element, and `screen-defs` covers `content ⊆ registry`).

### What's good

- Reused `BattlerCard` as the mount point precisely because it composes the full fight subtree — the minimal way to exercise all six registrations without standing up `Fight`.
- The inline comment explains *why* this direction matters and how the coverage is achieved, which is exactly the context a future reader needs.

No blocking findings, no should-fixes, no nits. Merging once the `all-checks-passed` gate finishes.

---

PR #1941, opened 2026-07-13
Share one flush between offline-rewards' progress and player saves (#1921)

### LINKED ISSUE #1921: Offline-rewards apply is split across two independently-flushed saves — a failed player flush permanently strands challenge/milestone rewards and double-counts statistics on the re-run
On the live battle path, a battle's player and progress writes share one batch flush (the progress save rides `PlayerSaveInProgress`), so they succeed or fail together. The offline-rewards pass is the exception:

- `OfflineProgressService.ApplyOfflineRewards` calls `_progressRepo.Save(progress, …)` standalone (`Game.Application/Services/OfflineProgressService.cs:353`), which flushes its own event (`PlayerProgressRepository.cs:129-149`) — durably recording completed challenges, statistics, and advanced proficiency levels.
- The player save — carrying the logically-paired mutations from the same window (`GrantOfflineExp`, `CompleteChallenge`'s reward-item unlocks, `GrantRewardSkills`' milestone skill unlocks, `StampActivity`) — flushes separately afterwards (`OfflineProgressService.cs:178-179`).

If the second flush fails transiently, `PlayerPersistenceFlushFailedException` forces the connection's player reload (discarding exp, unlocks, and the `LastActivity` re-anchor) while the progress flush has already committed. The next `GetOfflineProgress` re-simulates the same window:

- `EvaluateChallenges` skips already-completed rows (`PlayerProgress.cs:148-151`), so the challenge reward items are **never unlocked — permanently**;
- milestone reward skills are likewise never re-granted (accrual restarts from the already-advanced levels);
- `RecordBattleCompleted` folds a second window of increments onto the already-saved statistics (double-count).

**Fix direction:** let the offline pass share one flush like the live path — e.g. an explicit `PlayerUpdateBatch` scope opened around `ApplyOfflineRewards` + the final `SavePlayer`, so the progress event buffers and its cache advance defers via `OnFlushed`, closing the split-durability window.

(Not a duplicate of #1874, which is the session-cache reconnect double-credit on the battle-lifecycle path.)

ginderjeremiah:
## Summary

`OfflineProgressService.ApplyOfflineRewards` made its progress save (`IPlayerProgressRepository.Save`) as a standalone call, separate from the subsequent `SavePlayer` call that persists the same window's exp/levels/unlocks. On the live battle-completion path these two saves share one flush (the progress save is reached *through* `SavePlayer`'s own domain-event dispatch), but the offline pass never got that protection because it calls both repositories directly.

If the *second* (player) flush failed transiently, `PlayerPersistenceFlushFailedException` forced the connection's in-memory player to reload — but the *progress* flush had already durably committed. On the next `GetOfflineProgress` re-simulation of the same window:
- Already-completed challenges are skipped by `EvaluateChallenges`, so their reward items are **never unlocked — permanently**.
- Milestone reward skills are likewise never re-granted.
- Statistics fold a second window's increments onto the already-saved values — a **double-count**.

## Fix

- Added `IDisposable IPlayerRepository.BeginBatch()`, backed by `PlayerRepository`'s existing `PlayerUpdateBatch.BeginPlayerSave()`. This exposes the same batch-sharing seam `SavePlayer` already uses internally, without leaking the `Game.DataAccess`-internal `PlayerUpdateBatch` type into `Game.Application` (which only references `Game.Core`/`Game.Abstractions` per the project's onion architecture — the issue's literal suggested fix would have violated that layering).
- `OfflineProgressService.SimulateProgress` now wraps `ApplyOfflineRewards` (which makes the progress save) and the final `SavePlayer` call in one `using (_playerRepo.BeginBatch())` scope, so the progress save defers into the shared batch instead of flushing on its own — both writes now succeed or fail together.
- Updated the now-stale "e.g. the offline-rewards batch" comments/docs that described the offline pass as the standalone-save example (it no longer is), in `PlayerProgressRepository.cs`, the mirroring integration test, and `docs/backend-persistence.md`.

## Test plan

- [x] Added `SimulateOfflineProgress_SharedFlushFails_LeavesChallengeAndStatisticsUnpersistedForRetry`, which injects a `PlayerRepository` whose flush always fails and asserts:
  - the challenge is **not** completed and no statistics are recorded after the failed attempt (both saves failed together);
  - a retry against the untouched, freshly-reloaded player replays the same window exactly once (challenge completes once, item unlocks once, statistics reflect one window's worth of kills).
  - Verified this test fails on the pre-fix code (reverted the production fix locally, confirmed red, restored the fix).
- [x] `dotnet build` — clean.
- [x] `dotnet test` (full solution: Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests) — 3,120/3,120 passing.

Closes #1921

---
_Generated by [Claude Code](https://claude.ai/code/session_015gw8S6hxhdDCzNKTTeQfng)_


Game.Application/Services/OfflineProgressService.cs
```diff
@@ -169,14 +169,24 @@ private async Task<OfflineProgressSummary> SimulateProgress(Player player, Playe
...
+
+            OfflineRewards rewards;
+            // The progress save inside ApplyOfflineRewards and the player save below are logically one unit —
+            // exp/unlocks on the player and the completed challenges/statistics/proficiency levels on progress
+            // all come from the same simulated window. Sharing one batch scope means the progress save defers
+            // into it instead of flushing on its own, so a failure in the shared flush leaves both stranded for
+            // retry together rather than progress already being durably committed while the player save fails
+            // (#1921) — the same failure atomicity SavePlayer already gives a progress save reached through its
+            // own domain-event dispatch on the live battle-completion path.
+            using (_playerRepo.BeginBatch())
```
ginderjeremiah, line 184: **Should-fix (fragility, not a present bug):** This relies on an ordering invariant that isn't obvious. `PlayerUpdateBatch.PlayerSaveInProgress` is a plain `bool`, not a refcount, and `SavePlayer` opens its *own* `BeginPlayerSave()` scope internally whose disposal sets the flag back to `false` — so by the time control returns here (still inside this `using`), the batch window is already closed. It works today only because the sole progress save (in `ApplyOfflineRewards`) runs *before* `SavePlayer`. A future edit that adds a `_progressRepo.Save(...)` after `SavePlayer` but still inside this `using` would silently flush standalone and reintroduce exactly #1921, with no test catching it.

The `IPlayerRepository.BeginBatch()` doc ("any `Save` reached while it's active joins") also slightly overstates this — the inner `SavePlayer` scope ends the window early.

Cheap in-scope fix: a one-line comment here noting progress saves must precede `SavePlayer` within the scope. A more robust option is to reference-count `BeginPlayerSave` so nested scopes don't reset the flag until the outermost disposes — but that's a broader change and optional.

---

Game.Application.Tests/Services/OfflineProgressServiceIntegrationTests.cs
```diff
@@ -133,6 +138,112 @@ public async Task SimulateOfflineProgress_ChallengeCrossedMidWindow_CompletesOnc
...
+
+            // Statistics reflect exactly this one retry's worth of kills, not a doubled count left over from
+            // the stranded first pass.
+            var kills = Assert.Single(await progressRepo.GetStatistics(setup.PlayerId),
+                s => s.Type == EStatisticType.EnemiesKilled && s.EntityId == null);
+            Assert.Equal((decimal)summary.BattlesWon, kills.Value);
+        }
+
+        // Stands in for a transient Redis blip on the flush, mirroring PlayerWriteBehindTests' ThrowingPubSubService.
+        private sealed class ThrowingPubSubService : IPubSubService
```
ginderjeremiah: **Nit (DRY):** This re-implements the whole `IPubSubService` interface just to make the publish paths throw. The shared `Game.TestInfrastructure.Helpers.NotSupportedPubSubService` base exists for exactly this — the sibling `PlayerProgressRepositoryIntegrationTests.ThrowingPubSubService` already extends it and overrides only `PublishBatch`. Following that pattern here (and ideally consolidating the three near-identical `ThrowingPubSubService` copies into one shared helper) would be cleaner. Not blocking.

---

ginderjeremiah (commented):
## Review: Share one flush between offline-rewards' progress and player saves (closes #1921)

**Verdict:** Approve _(submitted as a comment — I'm the PR author, so GitHub won't take a gating verdict from me; the real call is Approve)_ — the fix correctly closes the split-durability window with a clean, layering-respecting seam and a test that pins the exact bug. My findings are a fragility should-fix and a test-DRY nit; neither blocks.

I traced the mechanism end-to-end and it holds: `BeginBatch()` sets `PlayerSaveInProgress`, so `ApplyOfflineRewards`'s `_progressRepo.Save` buffers its `ProgressUpdatedEvent` and defers its cache advance via `OnFlushed` instead of flushing; the subsequent `SavePlayer` enqueues both progress and player events in one `PublishBatch`, and on a transient publish failure `FlushAsync` drains nothing and the wrapped `PlayerPersistenceFlushFailedException` forces the same in-memory reload the standalone path already gives — so both writes strand together and the retry replays the window once.

### Should-fix

- `Game.Application/Services/OfflineProgressService.cs:181` — the batch window depends on an unstated ordering invariant. `PlayerSaveInProgress` is a plain `bool` (not refcounted), and `SavePlayer`'s internal `BeginPlayerSave()` scope resets it to `false` on disposal, so the `BeginBatch` window is effectively closed once `SavePlayer` returns. Correct today because the only progress save precedes `SavePlayer`, but a later `_progressRepo.Save` added after it (still inside the `using`) would silently re-open #1921. Suggest a one-line ordering comment here (cheap, in-scope), or optionally refcount `BeginPlayerSave`. Detail inline.

### Nits

- `Game.Application.Tests/Services/OfflineProgressServiceIntegrationTests.cs:228` — the new `ThrowingPubSubService` re-implements all of `IPubSubService`; the shared `NotSupportedPubSubService` base (already used by the sibling `PlayerProgressRepositoryIntegrationTests`) needs only a `PublishBatch` override. Consolidating the three copies would be cleaner. Detail inline.

### Test coverage

- Strong. `SimulateOfflineProgress_SharedFlushFails_LeavesChallengeAndStatisticsUnpersistedForRetry` is exactly the right level (integration — this is a write-behind flush-atomicity concern, not pure domain logic) and asserts both properties that matter: nothing from the doomed window is durably committed, and a forced-reload retry replays it exactly once (challenge completes once, item unlocks once, statistics reflect one window — no double-count). Author verified it goes red on the pre-fix code. No frontend mirror needed (no battle-simulation logic touched). `PlayerUpdateBatch`'s single-scope flag reset and flush semantics are already unit-covered; only the nested-scope reset (the footgun above) is untested — acceptable since it's currently unreachable.

### What's good

- Correctly rejected the issue's literal "open a `PlayerUpdateBatch` scope" suggestion, which would have leaked the `Game.DataAccess`-internal type into `Game.Application` — instead exposing the seam as `IPlayerRepository.BeginBatch()` in `Game.Abstractions`, preserving the onion layering. Exactly the kind of "verify the issue's claims" judgment CLAUDE.md asks for.
- Comments and `backend-persistence.md` are updated to match the new reality (the offline pass is no longer the standalone-save example), including the now-stale mirroring test comment — no dead/misleading docs left behind.

---

ginderjeremiah:
## Re-review (follow-up commit `b680f47`)

Verified both fixes since my prior review:
- **Should-fix (ordering invariant):** the `BeginBatch` window's dependence on any progress save preceding `SavePlayer` is now documented at both the call site (`OfflineProgressService.cs`) and the interface contract (`IPlayerRepository.BeginBatch`'s XML doc), noting that `SavePlayer`'s own nested scope disposal ends the shared window because `PlayerSaveInProgress` is a plain bool, not a refcount. Closes the latent re-open-#1921 footgun with a clear contract.
- **Nit (test DRY):** the three near-identical `ThrowingPubSubService` doubles are consolidated into one shared `Game.TestInfrastructure.Helpers.ThrowingPubSubService` extending `NotSupportedPubSubService` and overriding only `PublishBatch`. Green CI confirms the three call sites still compile and their assertions pass against the shared helper (the exercised flush path only invokes `PublishBatch`).

Both findings resolved, no new concerns. CI is green.

---

PR #1942, opened 2026-07-13
Use a process-stable hash for CombatRatingCalibrator seeds

### LINKED ISSUE #1922: CombatRatingCalibrator's "deterministic, reproducible" seeds are built with HashCode.Combine, which is randomized per process
`ComputeRewardCurve`'s matchup seeds are derived as:

```csharp
var seed = unchecked((uint)HashCode.Combine(zone.ZoneId, level, build.Name, spawn.EnemyId, i));
```

(`Game.Application/Content/Calibration/CombatRatingCalibrator.cs:230`.) `System.HashCode` seeds itself randomly per process, and .NET string hashing (`build.Name`) is randomized too — so every run of the calibration report simulates a different set of battles. The XML doc claims the opposite (`:154-155`): "Seeds are deterministic (hashed from the matchup), so the report is reproducible across runs against unchanged content."

The tool's stated purpose is re-running after content changes to spot pricing drift; with only 5 seeds per matchup, run-to-run RNG noise in win rate / average duration is indistinguishable from real drift, which quietly defeats that purpose.

**Fix direction:** derive the seed with a stable hash (e.g. FNV-1a over the matchup components, hashing `build.Name` as UTF-8 bytes), and pin it with a test asserting two runs over the same content produce identical reports.

ginderjeremiah:
## Summary

`CombatRatingCalibrator.ComputeRewardCurve`'s matchup seeds were built with `HashCode.Combine(zone.ZoneId, level, build.Name, spawn.EnemyId, i)`. `System.HashCode` reseeds itself randomly per process, and .NET's string hashing (used for `build.Name`) is randomized too — so every run of the calibration report simulated a different set of battles, contradicting the method's own XML doc claim that "seeds are deterministic... so the report is reproducible across runs against unchanged content."

With only 5 seeds per matchup, run-to-run RNG noise in win rate / average duration was indistinguishable from real content-driven drift, defeating the tool's stated purpose (re-running after content changes to spot pricing drift).

## Fix

- Replaced the seed derivation with `MatchupSeed`, which SHA256-hashes a stable string key (`"{zoneId}|{level}|{buildName}|{enemyId}|{drawIndex}"`) and takes the first 4 bytes as the `uint` seed. SHA256 has no per-process randomization, unlike `HashCode`/`string.GetHashCode`. This mirrors an existing pattern already used in the codebase for deterministic content hashing (`ReferenceDataVersioning.ComputeVersion`).
- Added `MatchupSeed_IsPinnedAcrossProcesses`, which asserts a literal expected seed value for a fixed input. This is the only way a single-process unit test can actually catch cross-process instability — the existing `ComputeRewardCurve_IsDeterministicAcrossRuns` test only compares two calls within the same process, which would have passed even with the old buggy implementation (since `HashCode`'s per-process seed is fixed once the process starts).
- Added `MatchupSeed_DiffersForDifferentBuildNamesWithTheSameHashCode` as a basic distinctness sanity check.

## Test plan

- [x] `dotnet build Game.sln` — clean, 0 warnings/errors.
- [x] `dotnet test Game.sln` (full solution: Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests) — 3,121/3,121 passing.
- [x] Verified the pinned seed value (`3800752286`) is identical across two independent process runs of the hash function before hardcoding it into the test.

Closes #1922

---
_Generated by [Claude Code](https://claude.ai/code/session_015L1KoWcHeXpRPgpSA8RPaM)_

### DIFF
--- a/Game.Application.Tests/Content/Calibration/CombatRatingCalibratorTests.cs
+++ b/Game.Application.Tests/Content/Calibration/CombatRatingCalibratorTests.cs
@@ -206,7 +206,7 @@ namespace Game.Application.Tests.Content.Calibration
         }
 
         [Fact]
-        public void MatchupSeed_DiffersForDifferentBuildNamesWithTheSameHashCode()
+        public void MatchupSeed_DiffersForDifferentBuildNames()
         {
             // .NET's built-in string hashing (used by HashCode.Combine) is randomized per process; MatchupSeed
             // must not derive from it. Distinct matchup keys should (in practice) yield distinct seeds.
--- a/Game.Application.Tests/Content/Calibration/CombatRatingCalibratorTests.cs
+++ b/Game.Application.Tests/Content/Calibration/CombatRatingCalibratorTests.cs
@@ -193,6 +193,29 @@ namespace Game.Application.Tests.Content.Calibration
             Assert.Equal(first[0].AvgBattleSeconds, second[0].AvgBattleSeconds, 6);
         }
 
+        // ── MatchupSeed ──────────────────────────────────────────────────────
+
+        [Fact]
+        public void MatchupSeed_IsPinnedAcrossProcesses()
+        {
+            // In-process determinism alone doesn't catch the bug this guards against: HashCode.Combine is also
+            // stable within a single process (it reseeds once at process start), so a same-process comparison
+            // would have passed even with the old, non-reproducible implementation. Pinning the literal value
+            // is the only way a unit test can assert stability *across* processes/runs.
+            Assert.Equal(3800752286u, CombatRatingCalibrator.MatchupSeed(3, 5, "Build A", 7, 2));
+        }
+
+        [Fact]
+        public void MatchupSeed_DiffersForDifferentBuildNamesWithTheSameHashCode()
+        {
+            // .NET's built-in string hashing (used by HashCode.Combine) is randomized per process; MatchupSeed
+            // must not derive from it. Distinct matchup keys should (in practice) yield distinct seeds.
+            var seedA = CombatRatingCalibrator.MatchupSeed(0, 1, "Build A", 0, 0);
+            var seedB = CombatRatingCalibrator.MatchupSeed(0, 1, "Build B", 0, 0);
+
+            Assert.NotEqual(seedA, seedB);
+        }
+
         // ── Helpers ───────────────────────────────────────────────────────────
 
         private static ZonePlacementRow MakeZonePlacementRow(
--- a/Game.Application/Content/Calibration/CombatRatingCalibrator.cs
+++ b/Game.Application/Content/Calibration/CombatRatingCalibrator.cs
@@ -1,6 +1,7 @@
 using Game.Core;
 using Game.Core.Battle;
 using Game.Core.Enemies;
+using System.Buffers.Binary;
 using System.Security.Cryptography;
 using System.Text;
 
@@ -247,7 +248,7 @@ namespace Game.Application.Content.Calibration
         {
             var key = $"{zoneId}|{level}|{buildName}|{enemyId}|{drawIndex}";
             var hash = SHA256.HashData(Encoding.UTF8.GetBytes(key));
-            return BitConverter.ToUInt32(hash, 0);
+            return BinaryPrimitives.ReadUInt32LittleEndian(hash);
         }
 
         // Evenly-spaced, deduplicated levels across [min, max] — samples ≤ 1 or a single-level range collapse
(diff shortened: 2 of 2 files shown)


Game.Application.Tests/Content/Calibration/CombatRatingCalibratorTests.cs
```diff
@@ -193,6 +193,29 @@ public void ComputeRewardCurve_IsDeterministicAcrossRuns()
...
+        {
+            // In-process determinism alone doesn't catch the bug this guards against: HashCode.Combine is also
+            // stable within a single process (it reseeds once at process start), so a same-process comparison
+            // would have passed even with the old, non-reproducible implementation. Pinning the literal value
+            // is the only way a unit test can assert stability *across* processes/runs.
+            Assert.Equal(3800752286u, CombatRatingCalibrator.MatchupSeed(3, 5, "Build A", 7, 2));
+        }
+
+        [Fact]
+        public void MatchupSeed_DiffersForDifferentBuildNamesWithTheSameHashCode()
```
ginderjeremiah: Nit: the name `...WithTheSameHashCode` overclaims what the test establishes — `"Build A"` and `"Build B"` don't actually collide under `string.GetHashCode`, so nothing here demonstrates the "same HashCode" premise. The test really just asserts that distinct build names yield distinct seeds. Consider renaming to something like `MatchupSeed_DiffersForDifferentBuildNames` (the comment already captures the "must not derive from string hashing" intent). Purely cosmetic — take it or leave it.

---

ginderjeremiah (commented):
## Review: process-stable hash for `CombatRatingCalibrator` matchup seeds (#1922)

**Verdict: Approve** — correctly fixes the per-process seed randomization that made the calibration report's "reproducible across runs" claim false; clean, well-tested, and consistent with existing codebase patterns. (Submitted as `COMMENT` only because I'm the PR author — GitHub won't accept a gating verdict on one's own PR. My actual call is approve.)

The root cause is diagnosed correctly: `HashCode.Combine` reseeds per process and `build.Name` goes through randomized string hashing, so with only 5 seeds/matchup the RNG noise was indistinguishable from real content drift. Replacing it with `MatchupSeed` (SHA256 over a stable `"{zoneId}|{level}|{buildName}|{enemyId}|{drawIndex}"` key → first 4 bytes as `uint`) makes the report genuinely reproducible, and the XML doc claim at `CombatRatingCalibrator.cs:156` is now accurate rather than aspirational.

### Nits
- `CombatRatingCalibratorTests.cs:209` — the test name `...WithTheSameHashCode` overclaims; `"Build A"`/`"Build B"` don't share a `GetHashCode`, so the test really just asserts distinct names → distinct seeds. Cosmetic rename suggestion inline.
- `CombatRatingCalibrator.cs:250` — `BitConverter.ToUInt32` interprets the 4 bytes using machine endianness, so the pinned literal (`3800752286u`) is technically little-endian-specific. This is fine in practice (all realistic dev/CI targets are little-endian, and the tool only ever compares runs on comparable hardware), but a one-line note, or `BinaryPrimitives.ReadUInt32LittleEndian`, would make the invariant explicit. Take it or leave it.

### Test coverage
Solid and, notably, correct about *why*: `MatchupSeed_IsPinnedAcrossProcesses` pins a literal value, which is the only thing that actually catches the bug — the pre-existing `ComputeRewardCurve_IsDeterministicAcrossRuns` compares two calls within one process and would have passed against the buggy code too. Good call not relying on it. This is pure calibration tooling (human-run, read-only, no domain logic), so no mirrored frontend battle-logic test is needed.

### What's good
- `MatchupSeed` deliberately mirrors the established `ReferenceDataVersioning.ComputeVersion` SHA256-over-UTF8 pattern rather than inventing a new hash (the issue's FNV-1a suggestion would've been a one-off) — good consistency.
- `internal` + the existing `InternalsVisibleTo("Game.Application.Tests")` keeps the helper out of the public surface while still testable.
- The inline comments explain the non-obvious cross-process reasoning without over-documenting.

Nothing here blocks. Both nits are optional and don't warrant follow-up issues.

---

ginderjeremiah:
## Re-review (follow-up commit `0ac275f`)

Verified both nit fixes since my prior review:
- `MatchupSeed_DiffersForDifferentBuildNamesWithTheSameHashCode` → `MatchupSeed_DiffersForDifferentBuildNames` — drops the overclaim that the two build names collide under `string.GetHashCode`.
- `BitConverter.ToUInt32(hash, 0)` → `BinaryPrimitives.ReadUInt32LittleEndian(hash)` — makes the byte order explicit rather than implicitly machine-endianness-dependent. Behavior is identical on little-endian targets, so the pinned seed (`3800752286u`) still holds — confirmed by green CI on this commit.

Both nits resolved, no new findings. CI is green.

---

PR #1945, opened 2026-07-13
Migrate remaining integration-test polling loops onto PollingHelper

### LINKED ISSUE #1789: Migrate remaining integration-test polling loops onto PollingHelper
Follow-up from PR #1754, which introduced the shared generic `PollingHelper.PollUntilAsync` and converted the polling loops in `RedisServiceIntegrationTests` and `PlayerProgressRepositoryIntegrationTests`.

Several other `while (DateTime.UtcNow < deadline)` polling loops remain scattered across the backend test suite and are now mechanically consolidatable onto the same helper — e.g. `LoginControllerTests`, `SocketManagerServiceTests`, `SessionCacheEvictionTests`, `PlayerCacheEvictionTests`, `BackgroundWorkerTests`. These poll different subjects and were rightly outside #1723's scope, but share the identical poll-until-predicate-or-timeout shape.

No behavior change — test-suite maintainability only.

ginderjeremiah:
## Summary

Follow-up from PR #1754, which introduced `PollingHelper.PollUntilAsync` and converted the polling loops in `RedisServiceIntegrationTests` and `PlayerProgressRepositoryIntegrationTests`. Several other `while (DateTime.UtcNow < deadline)` polling loops remained scattered across the backend test suite with the identical poll-until-predicate-or-timeout shape — this consolidates them onto the shared helper:

- `LoginControllerTests`: `AssertPlayerLeveledAboveAsync`, `AssertPlayerNotCreditedAsync`, `AssertSessionPresentAsync`, `AssertSessionEvictedAsync`
- `SocketManagerServiceTests`: `WaitUntilAsync`
- `SocketCommandTimeoutTests`: `WaitForSentMessageAsync`
- `SessionCacheEvictionTests` / `PlayerCacheEvictionTests`: `WaitForTtlAsync` (identical duplicated helper in both files)
- `RedisServiceIntegrationTests`: `AssertKeyEventuallyDeletedAsync` — a loop that was missed when #1754 converted the rest of this file

No behavior change — test-suite maintainability only.

**Intentionally not migrated:** `BackgroundWorkerTests`' `WaitUntil`/`WaitUntilStablyNotRunning`. Both are synchronous (`Thread.Sleep`-based, called from non-`async` test methods) in a suite explicitly designed around avoiding non-deterministic timing complexity for its thread-pool synchronization tests. `WaitUntilStablyNotRunning` also needs N-consecutive-successful-polls semantics that `PollingHelper` doesn't support — bending it to fit would mean either sync-over-async bridging (an anti-pattern, and a real risk in a concurrency-sensitive suite) or extending `PollingHelper` with semantics no other caller needs. Forcing it wouldn't be a mechanical, no-behavior-change refactor, so it's left as-is.

## Test plan

- [x] `dotnet test` (full backend suite, including integration tests against the session's Postgres/Redis containers) — all 3126 tests pass

Closes #1789

---
_Generated by [Claude Code](https://claude.ai/code/session_01JizwiX643RVgM5KtfBg5dv)_

### DIFF
--- a/Game.Api.Tests/Integration/LoginControllerTests.cs
+++ b/Game.Api.Tests/Integration/LoginControllerTests.cs
@@ -544,19 +544,11 @@ namespace Game.Api.Tests.Integration
         // until the departed character's level reflects the simulated away window.
         private async Task AssertPlayerLeveledAboveAsync(int playerId, int levelBefore)
         {
-            var deadline = DateTime.UtcNow + TimeSpan.FromSeconds(5);
-            while (DateTime.UtcNow < deadline)
-            {
-                var persisted = await GetPersistedPlayerAsync(playerId);
-                if (persisted.Level > levelBefore)
-                {
-                    return;
-                }
+            var persisted = await PollingHelper.PollUntilAsync(
+                () => GetPersistedPlayerAsync(playerId), p => p.Level > levelBefore, timeoutMs: 5000);
 
-                await Task.Delay(25, CancellationToken);
-            }
-
-            Assert.Fail("The departed character was not credited (its level did not advance) after the switch.");
+            Assert.True(persisted.Level > levelBefore,
+                "The departed character was not credited (its level did not advance) after the switch.");
         }
 
         // The inverse of AssertPlayerLeveledAboveAsync: gives any write-behind credit a window to surface and
@@ -564,14 +556,11 @@ namespace Game.Api.Tests.Integration
         // this window, so a regression (credit not skipped while a socket is live) is caught here.
         private async Task AssertPlayerNotCreditedAsync(int playerId, int levelBefore)
         {
-            var deadline = DateTime.UtcNow + TimeSpan.FromSeconds(2);
-            while (DateTime.UtcNow < deadline)
-            {
-                var persisted = await GetPersistedPlayerAsync(playerId);
-                Assert.False(persisted.Level > levelBefore,
-                    "The departed character was credited despite holding a live socket; the off-lock credit should have been skipped.");
-                await Task.Delay(25, CancellationToken);
-            }
+            var persisted = await PollingHelper.PollUntilAsync(
+                () => GetPersistedPlayerAsync(playerId), p => p.Level > levelBefore, timeoutMs: 2000);
+
+            Assert.False(persisted.Level > levelBefore,
+                "The departed character was credited despite holding a live socket; the off-lock credit should have been skipped.");
         }
 
         [Fact]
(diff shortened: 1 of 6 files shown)


ginderjeremiah (commented):
## Review: Migrate remaining integration-test polling loops onto `PollingHelper` (#1789)

**Verdict:** Approve — clean, behavior-preserving consolidation onto the shared `PollingHelper.PollUntilAsync`. (Submitted as a `COMMENT` since I'm the PR author and GitHub won't accept a gating verdict on my own PR; the intended call is **Approve**.)

I checked each conversion against `PollingHelper.PollUntilAsync` (polls every 25ms, returns the last read value on timeout, leaving pass/fail to the caller) and they all preserve the original semantics:

- `AssertPlayerLeveledAboveAsync` / `AssertSessionPresentAsync` / `AssertSessionEvictedAsync` / `AssertKeyEventuallyDeletedAsync` — straightforward "return on predicate, else fail" → predicate + terminal `Assert.True`. Equivalent.
- `AssertPlayerNotCreditedAsync` — the interesting one. The old loop asserted `level did not advance` on *every* poll; the new version uses `level > levelBefore` as the predicate so `PollUntilAsync` returns early on the *first* advance (or the last un-advanced read on timeout), then a single `Assert.False`. Because level is monotonic (credit only ever raises it), the first-violation capture is equivalent to the per-iteration assert. Correct, and the inline comment already explains the intent.
- `WaitUntilAsync` (SocketManagerService) — the old post-loop `return await condition()` is preserved by the helper returning its last fresh read; polling cadence tightens from 100ms→25ms, which is harmless.
- `WaitForTtlAsync` (both files) — returns last (possibly null) TTL on timeout, same as before.

### Nits (non-blocking)
- `SocketCommandTimeoutTests.WaitForSentMessageAsync` (`Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs:188`) no longer touches instance state, yet stays an instance method while its siblings in this same PR (`WaitUntilAsync`, `WaitForTtlAsync`) were made `static`. Making it `static` too would keep the conversions consistent.
- `WaitForTtlAsync` is still duplicated verbatim across `PlayerCacheEvictionTests` and `SessionCacheEvictionTests` (now thin wrappers over `PollingHelper`). Not worth blocking, but a shared TTL helper would finish the de-duplication — fine as a follow-up if you'd rather not widen scope here.

### Follow-up (optional, pre-existing)
- `PollingHelper` uses an uncancellable `Task.Delay(25)`, so the tests' `CancellationToken` is dropped from the delay on conversion. This is inherited from #1754's design (not introduced here) and is consistent across all callers; a `CancellationToken` overload on `PollingHelper` could be a small future cleanup.

### Test coverage
Appropriate — this is a test-infrastructure refactor with no production-code change, so no new tests are warranted. The migrated helpers remain exercised by their existing integration tests, and the PR reports the full backend suite (3126 tests) green.

### What's good
The `AssertPlayerNotCreditedAsync` inversion is a genuinely nice fit of the "poll for the failure condition, then assert it never held" idiom onto a predicate-based helper, and the scope call on `BackgroundWorkerTests` (sync `Thread.Sleep` loops + N-consecutive-poll semantics) is the right one — bending `PollingHelper` to cover it would have added complexity no one else needs.

---

ginderjeremiah:
## Re-review (delta since prior review): `c8cf512`

Commit `c8cf512` resolves review nit #1: `WaitForSentMessageAsync` is now `static`, matching its sibling conversions (`WaitUntilAsync`, `WaitForTtlAsync`) that also no longer touch instance state after the migration.

The two remaining items from the earlier review are genuinely out of scope for this pure, no-behavior-change migration, so rather than widen it I've filed them as a follow-up: **#1949** —
1. the duplicated `WaitForTtlAsync` helper across `PlayerCacheEvictionTests` / `SessionCacheEvictionTests` (extract a shared TTL-polling helper), and
2. a `CancellationToken` overload on `PollingHelper` so the poll delay honors cancellation (pre-existing from #1754, not introduced here).

CI (`Run tests`, run 2381) is green on the head SHA. Merging.

---

PR #1946, opened 2026-07-13
Socket hardening: Close() never throws; a null-deserialized frame closes instead of stranding the client

### LINKED ISSUE #1928: Socket hardening: Close propagates transport faults into the loop tasks, and a JSON `null` frame is silently dropped, stranding the client's request
Two small robustness gaps in the socket receive/teardown path, same subsystem so filed together:

**1. `SocketContext.Close` only shields the close frame from cancellation.** `Game.Api/Sockets/SocketContext.cs:253-261` wraps `await _socket.CloseAsync(...)` in `catch (OperationCanceledException)` only; a `WebSocketException` (peer RST between the `Open`/`CloseReceived` re-check and the frame write — realistic on abrupt client drops) propagates out of `Close`. Two callers are unguarded: the read loop's unconditional final close (`SocketHandler.cs:431`, outside the do-while's try/catch) and the inactivity watchdog (`SocketHandler.cs:366-369`). That faults `_loops`/`Completion` — producing unobserved-task exceptions on the normal path, and during a drain a spurious "Error draining socket" warning, contradicting `ShutdownAsync`'s stated premise that the loops never fault on cancellation (`SocketHandler.cs:100-107`). Teardown itself still works (the TCS settles in the `finally`). Fix: catch non-cancellation exceptions around `CloseAsync` inside `Close` (log + `Abort()`, mirroring `SendData`'s never-throws contract), fixing every call site at once.

**2. A frame that parses to JSON `null` is silently dropped.** `SocketHandler.HandleMessage` (`SocketHandler.cs:449-483`) deserializes and runs `if (commandInfo is not null) { ... }` with no else branch. A frame containing the literal `null` passes the `IsNullOrWhiteSpace` filter, deserializes to `null`, and falls through with no response and no close — exactly the "client waits out its 30s request timeout with no diagnosable error" stranding the adjacent guards were built to eliminate (the missing-`name` guard at `:452-460` and the `JsonException` close at `:485-493`, whose comment explicitly names this failure mode). Since a `null` frame carries no request id to correlate an error to, mirror the `JsonException` path: log and `Close(ESocketCloseReason.MalformedFrame)`.

ginderjeremiah:
## Summary

Two small, independently-verified robustness gaps in the socket receive/teardown path (grouped together in #1928, filed from a scheduled codebase health review):

1. **`SocketContext.Close` only shielded the close frame from cancellation.** A `WebSocketException` (e.g. a peer RST between the `Open`/`CloseReceived` re-check and the frame write — realistic on abrupt client drops) propagated out of `Close`. Two callers invoke it unguarded: the read loop's unconditional terminal close (`SocketHandler.ReadLoop`) and the inactivity watchdog (`SocketHandler.InactivityCheckerLoop`). That faulted `_loops`/`Completion` — unobserved-task exceptions on the normal path, and a spurious "Error draining socket" warning during a drain — contradicting `ShutdownAsync`'s stated premise that the loops never fault on cancellation.

   **Fix:** `Close` now catches non-cancellation exceptions around `CloseAsync`, logs, and aborts the connection — mirroring `SendData`'s existing never-throws contract. Every call site is fixed at once.

2. **A frame that parses to JSON `null` was silently dropped.** `SocketHandler.HandleMessage` deserialized and ran `if (commandInfo is not null) { ... }` with no `else` branch. A frame containing the literal `null` passes the `IsNullOrWhiteSpace` filter, deserializes to `null`, and fell through with no response and no close — exactly the "client waits out its 30s request timeout with no diagnosable error" failure mode the adjacent guards (the missing-`name` check, the `JsonException` catch) exist to eliminate. Since a `null` frame carries no request id to correlate an error response to, it can't be answered like a normal malformed-parameters rejection.

   **Fix:** mirrors the `JsonException` path — log and `Close(ESocketCloseReason.MalformedFrame)`.

## Changes

- `Game.Api/Sockets/SocketContext.cs`: `Close` catches non-cancellation exceptions from `CloseAsync`, logs, and aborts.
- `Game.Api/Sockets/SocketHandler.cs`: `HandleMessage` closes the socket as `MalformedFrame` when a frame deserializes to `null`.
- `Game.Api.Tests/Unit/SocketContextTests.cs`: updated `Close_StillSettlesWaitSocketClosed_WhenCloseFrameSendThrows` to assert the new no-throw/abort contract instead of the old `Assert.ThrowsAsync<WebSocketException>` (which pinned the buggy behavior); added `AbortCalled` tracking to `GatedCloseWebSocket`.
- `Game.Api.Tests/Unit/SocketReadLoopTests.cs`: added `HandleMessage_FrameDeserializesToNull_ClosesAsMalformedFrame`, reusing the existing `ScriptedReadWebSocket` harness.

## Test plan

- [x] `dotnet test` (full backend suite, including integration tests against the session's Postgres/Redis containers) — all 3127 tests pass

Closes #1928

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ptwd8FZmy6LvNvDHcEup1D)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketContextTests.cs
+++ b/Game.Api.Tests/Unit/SocketContextTests.cs
@@ -55,8 +55,11 @@ namespace Game.Api.Tests.Unit
 
             socket.FailClose(new WebSocketException("close frame send failed"));
 
-            // The send fault propagates out of Close, but the waiter is still released (best-effort).
-            await Assert.ThrowsAsync<WebSocketException>(() => closeTask);
+            // The send fault (e.g. a peer RST) is caught and the connection aborted rather than propagating
+            // out of Close — every caller invokes it unguarded, so an escaping exception would fault their
+            // loop task (#1928) — and the waiter is still released (best-effort).
+            await closeTask.WaitAsync(WaitTimeout, cancellationToken);
+            Assert.True(socket.AbortCalled);
             await waitClosed.WaitAsync(WaitTimeout, cancellationToken);
             Assert.True(waitClosed.IsCompletedSuccessfully);
         }
@@ -412,6 +415,7 @@ namespace Game.Api.Tests.Unit
             /// <summary>Completes once <see cref="CloseAsync"/> has begun (and is blocking on the gate).</summary>
             public Task CloseStarted => _closeStarted.Task;
             public bool CloseAsyncCalled { get; private set; }
+            public bool AbortCalled { get; private set; }
             public WebSocketCloseStatus? CloseStatusUsed { get; private set; }
             public WebSocketState StateOverride { get; init; } = WebSocketState.Open;
 
@@ -434,7 +438,7 @@ namespace Game.Api.Tests.Unit
             public override Task<WebSocketReceiveResult> ReceiveAsync(ArraySegment<byte> buffer, CancellationToken cancellationToken) => throw new NotSupportedException();
             public override Task SendAsync(ArraySegment<byte> buffer, WebSocketMessageType messageType, bool endOfMessage, CancellationToken cancellationToken) => Task.CompletedTask;
             public override Task CloseOutputAsync(WebSocketCloseStatus closeStatus, string? statusDescription, CancellationToken cancellationToken) => Task.CompletedTask;
-            public override void Abort() { }
+            public override void Abort() => AbortCalled = true;
             public override void Dispose() { }
         }
 
--- a/Game.Api.Tests/Unit/SocketReadLoopTests.cs
+++ b/Game.Api.Tests/Unit/SocketReadLoopTests.cs
@@ -144,6 +144,29 @@ namespace Game.Api.Tests.Unit
             await handler.Completion.WaitAsync(WaitTimeout, TestContext.Current.CancellationToken);
         }
 
+        [Fact]
+        public async Task HandleMessage_FrameDeserializesToNull_ClosesAsMalformedFrame()
+        {
+            // A frame containing the JSON literal "null" deserializes to a null SocketCommandInfo without
+            // throwing, so it carries no request id to correlate a structured rejection to. It must close the
+            // socket like the JsonException path rather than silently dropping the frame and stranding the
+            // client on its 30s request timeout (#1928).
+            var socket = new ScriptedReadWebSocket();
+            socket.QueueMessage("null");
+            var (context, handler) = CreateHandler(socket);
+
+            using var inactivityStop = new CancellationTokenSource();
+            handler.Listen(hostStopping: inactivityStop.Token);
+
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Socket hardening — `Close` never throws; a `null`-deserialized frame closes instead of stranding the client (closes #1928)

**Verdict:** Approve — both fixes are correct, minimal, and land exactly the two gaps #1928 describes; tests pin the new behavior and no other socket path is affected. _(Submitted as a comment because I'm the PR author and GitHub won't accept a formal approval on one's own PR.)_

I verified both changes against the code in context:

**1. `SocketContext.Close` transport-fault guard** — the new `catch (Exception ex)` sits after the `OperationCanceledException` catch, so cancellation still routes to the existing timeout/abort path and only non-cancellation transport faults (the peer-RST race the issue calls out) hit the new branch. Log + `Abort()` mirrors `SendData`'s never-throws contract, and the `finally` still settles `_socketClosedSource` in every case. This genuinely fixes all three unguarded call sites at once (read-loop terminal close, inactivity watchdog, shutdown drain) and restores `ShutdownAsync`'s "loops never fault on cancellation" premise.

**2. `HandleMessage` null-frame close** — the added `else` correctly handles the `commandInfo is null` case (JSON literal `null` deserializing without throwing). Closing with `MalformedFrame` rather than attempting a structured rejection is the right call since a `null` frame carries no request id to correlate to, and it's consistent with the sibling `JsonException` path. Idempotency holds: the read loop's unconditional terminal `Close(Finished)` afterward is a no-op send that just re-settles the TCS.

### Test coverage
- `SocketContextTests`: the updated `Close_StillSettlesWaitSocketClosed_WhenCloseFrameSendThrows` correctly replaces the old `Assert.ThrowsAsync` (which pinned the buggy propagation) with the no-throw + `AbortCalled` + waiter-released assertions. Good — that old assertion was actively enforcing the bug.
- `SocketReadLoopTests.HandleMessage_FrameDeserializesToNull_ClosesAsMalformedFrame`: exercises the new branch end-to-end through the real read loop and asserts the close status (`InvalidPayloadData`, which I confirmed is `MalformedFrame`'s mapping in `Extensions.GetCloseStatus`) and the warning log.
- Backend-only, non-battle logic, so no frontend mirror is required. Coverage is appropriate for the change.

### What's good
- Tightly scoped to the issue with no gold-plating; every call site fixed in one place rather than per-caller.
- The inline comments explain the *why* (the RST race, the missing request id) concisely without over-documenting, matching the surrounding file's style.
- The test change deliberately flips an assertion that was encoding the old bug — exactly the right move, and it's called out in the PR description.

Approving. Will merge once the remaining CI jobs (test-backend, e2e firefox/webkit) finish green.

---

Issue #1864, opened 2026-07-11
Small component/store cleanups: focus-trap scroll-lock refcount, toast duplicate-collapse duration rewrite, tour callout stale position on shared anchors


ginderjeremiah:
Three small, verified issues in shared UI primitives, grouped like #1772. All low-severity; none currently user-visible in shipped flows.

## 1. Focus-trap body scroll-lock class is not reference-counted

`components/focus-trap.ts:70-97` — the keyboard side of stacking is correctly handled by `trapStack` (only the top trap responds) and the header comment advertises stacking as first-class, but the scroll-lock is a bare `classList.add/remove`. Two traps sharing the same `scrollLockClass` break it: the second `add` is a no-op and the inner trap's `destroy` removes the class while the outer overlay is still open, unlocking background scroll prematurely. Both `Popover`s hardcode `scrollLockClass: 'popover-open'`, so two stacked Popovers would hit this — currently latent (a Popover backdrop intercepts the click that would open a second).

**Fix:** refcount per class name (module-level `Map<string, number>`), add on 0→1, remove on 1→0.

## 2. Toast duplicate-collapse silently rewrites the surviving toast's persistence

`stores/toast.svelte.ts:101-107` — the collapse re-arms the existing toast's timer with the **new** call's `duration`. A persistent toast (`duration: 0`) hit by a same-message call that omits `duration` starts auto-dismissing after 6s; a timed toast hit by a `duration: 0` duplicate becomes permanent (pending dismiss cleared, none rescheduled). The header comment frames the collapse as "refresh its timer", which hides this.

**Fix:** keep the surviving toast's original duration on collapse, or document that the duplicate's duration wins.

## 3. Tour callout position not recomputed for consecutive steps sharing an anchor

`components/tour/TourPlayer.svelte:126-160` — the positioning `$effect` re-runs on `resizeTick`/`open`/`anchorEl` but not on the step's text. When Next/Back moves between two steps whose `anchorKey` is identical (or both unanchored), the effect doesn't re-run even though the callout's rendered height changed, so `calloutPos` can be stale — the callout sits slightly off or clips near a viewport edge. Cosmetic.

**Fix:** also track `stepIndex`/`currentStep.text` in the position effect.

<sub>Found during a scheduled codebase health review.</sub>

PR #1947, opened 2026-07-13
Dismiss the tutorial tour via Escape in e2e harness instead of clicking Skip

### LINKED ISSUE #1930: WebKit e2e first attempt fails on every CI run at the tutorial-tour Skip click — retries are permanently masking it
All three most recent green `main` runs (29229459178, 29226507544, 29224322855) uploaded a ~27.5–27.9 MB `playwright-report-webkit` trace artifact while chromium/firefox uploaded nothing — traces exist only `on-first-retry`, so WebKit is failing its first attempt on 100% of runs. The job log shows the identical signature each time: 4 tests flaky (`admin.test.ts` ×3 + `challenge-boss.test.ts`), all timing out at `await page.getByTestId('tutorial-tour-skip').click()` (`UI/e2e-tests/helpers.ts:141-150`) with Playwright stuck at "waiting for element to be visible, enabled and stable" until the 30s test timeout — the locator resolves, so the callout renders but never passes the stability check on WebKit (`UI/src/components/tour/TourPlayer.svelte:126-160`).

Two costs: ~2 min of wasted webkit job time per run, and — worse — a *real* WebKit regression in these four flows would produce the same signature and be retried past, so the retry budget is spent masking a known failure instead of absorbing genuine flake.

Likely shares a root cause with the "tour callout stale position on shared anchors" line item in #1864 (the callout toggles centered↔anchored as `anchorEl` re-registers), but that issue frames it as a small component cleanup — nothing tracks that WebKit CI fails first-attempt every run.

**Fix direction:** immediate harness fix — dismiss the tour via `page.keyboard.press('Escape')` (the `focusTrap` `onEscape` path needs no element-stability check). Separately root-cause the callout instability on WebKit (coordinate with #1864's stale-position item).

ginderjeremiah:
## Summary

WebKit was failing its first attempt on every CI run (`admin.test.ts` ×3 + `challenge-boss.test.ts`), always at the same spot: `dismissTutorialTourIfPresent` clicking `tutorial-tour-skip`. Playwright's click waits for the target to pass a visibility/stability check, and on WebKit the tour callout — which repositions itself via a resize/anchor effect (`TourPlayer.svelte`) — never settles in time, so the click times out and the test only passes on retry. Retries were quietly absorbing a 100%-reproducible first-attempt failure instead of real flake, per the issue's investigation of three recent green `main` runs.

## Fix

`TourPlayer.svelte`'s callout shell is wrapped in `focusTrap`, which attaches an `onEscape` handler via a `window`-level `keydown` listener (`focus-trap.ts`) — the same `onDismiss` callback the Skip button and backdrop already use. A `keydown` dispatch has no visibility/stability requirement, so switching the shared e2e helper to `page.keyboard.press('Escape')` dismisses the tour identically without depending on the callout having finished animating into place.

`tutorial.test.ts` is left clicking the Skip control directly, since that test exists specifically to pin the Skip control's own behavior.

## Test plan

- [x] `npm run test:unit:ci` — all unit tests pass
- [ ] Full Playwright WebKit run wasn't exercised locally (spinning up the backend + Postgres/Redis-backed dev stack for e2e is heavyweight per project convention, and this change is confined to e2e test harness code) — verified instead by reading `focus-trap.ts`'s `onKeydown` implementation, which confirms Escape reaches the trap through a `window` listener regardless of the callout's animation/positioning state. Will watch the CI run on this PR to confirm the WebKit job passes its first attempt (no `playwright-report-webkit` trace artifact expected).

Closes #1930

---
_Generated by [Claude Code](https://claude.ai/code/session_01MJJs5uv13VWztuohEABUoN)_

### DIFF
--- a/UI/e2e-tests/helpers.ts
+++ b/UI/e2e-tests/helpers.ts
@@ -129,14 +129,19 @@ export async function createAccountAndStartGame(page: Page, prefix = 'e') {
 /**
  * A fresh account's first visit to the (default) Fight screen auto-plays the screen-anchored
  * `idle-loop-basics` tutorial tour (a full-screen coach-mark overlay), which would otherwise swallow
- * the very next click every journey makes. Dismiss it via its Skip control if it appeared, so tests
- * interact with the screen underneath rather than asserting anything about tutorial content — that's
- * covered by tutorial.test.ts.
+ * the very next click every journey makes. Dismiss it via Escape if it appeared, so tests interact
+ * with the screen underneath rather than asserting anything about tutorial content — that's covered
+ * by tutorial.test.ts (which exercises the Skip control itself).
  *
  * The tour opens from the welcome-back gate's async `entered` transition, not synchronously with the
  * `/game` navigation, so this waits for it rather than taking an immediate (and racy) snapshot —
  * mirroring how {@link selectFirstCharacter} waits out the similarly-async session-takeover modal.
  * Swallows the timeout if no tour appears (e.g. the lesson is retired later), rather than failing.
+ *
+ * Dismisses via Escape rather than clicking the Skip control: Escape routes through `focusTrap`'s
+ * `window`-level keydown listener (`onEscape`), so it needs no element-visibility/stability check —
+ * unlike a click, which on WebKit can time out waiting for the callout to pass its stability check
+ * while it's still repositioning (#1930).
  */
 export async function dismissTutorialTourIfPresent(page: Page) {
 	const tour = page.getByTestId('tutorial-tour');
@@ -145,7 +150,7 @@ export async function dismissTutorialTourIfPresent(page: Page) {
 	} catch {
 		return;
 	}
-	await page.getByTestId('tutorial-tour-skip').click();
+	await page.keyboard.press('Escape');
 	await expect(tour).not.toBeVisible();
 }
 


ginderjeremiah (commented):
## Review: Dismiss the tutorial tour via Escape in the e2e harness instead of clicking Skip (#1930)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating event from me) — the fix matches the issue's prescribed direction and the mechanism is verified against the actual code.

### Verification

I traced the whole dismissal path to confirm the claim in the new comment holds:

- `TourPlayer.svelte:37` wires `use:focusTrap={{ onEscape: onDismiss }}`, so Escape and the Skip button (`:44`) share the exact same `onDismiss` callback — dismissing via Escape is behaviorally identical to clicking Skip.
- `focus-trap.ts:38-46,71` registers `onKeydown` on the **window** and invokes `onEscape` on `event.key === 'Escape'`. Being a window-level listener, it has no element-visibility/stability precondition — which is precisely why it sidesteps the WebKit callout-stability timeout that a `.click()` on the still-repositioning callout hits. The comment's rationale is accurate.
- No race between the visibility gate and the trap: `data-testid="tutorial-tour"` (the layer) and the `use:focusTrap` callout render in the same `{#if open && currentStep}` block, so once the layer is visible the trap is already mounted and listening. The tour is also the top-most (only) overlay at this point in account setup, so the trap-stack top-most check routes Escape to it.

The kept `await expect(tour).not.toBeVisible()` still asserts the dismissal actually took effect, so the helper hasn't lost any of its guarantees.

### Test coverage

Appropriate as-is. This is e2e harness code exercised by every journey that calls `createAccountAndStartGame`; the WebKit e2e job on this PR passing its **first attempt** (no retry trace artifact) is the real confirmation. `tutorial.test.ts` is correctly left clicking the Skip control directly, since pinning that control's own behavior is its purpose — the comment calls this out.

### What's good

- Fix is minimal and lands exactly where the issue pointed, with no scope creep.
- The added comment documents the WebKit-specific *why* well — worth keeping, since the non-obvious reason for Escape-over-click is exactly the kind of thing a later edit would otherwise "simplify" back into a flaky click.
- Correctly scoped down to a harness fix while leaving the deeper callout-instability root-cause to #1864, as the issue directed.

No blocking issues, no should-fixes, no nits. Merging once `test-frontend` (the last in-progress check) goes green.

---

PR #1948, opened 2026-07-13
Re-snapshot player currentHealth after the signature passive lands on full rebuild

### LINKED ISSUE #1933: Frontend battler snapshots starting health before the class signature passive is applied — a FE/BE parity divergence on every full rebuild (latent until a nonzero passive ships)
`BattleEngine.resetPlayer` calls `this.player.reset(...)` — which ends with `this.currentHealth = this.attributes.getValue(EAttribute.MaxHealth);` (`UI/src/lib/battle/battler.ts:439`) — and only **then** composes the signature passive: `applySignaturePassive(this.player.attributes, ...)` (`UI/src/lib/engine/battle/battle-engine.ts:280-292`). The backend does the opposite: `BattlerMaterials.Build()` adds the passive to the `AttributeCollection` first, and the `Battler` ctor then snapshots `CurrentHealth = _attributes[MaxHealth]` — passive **included** (`Game.Core/Battle/BattleSnapshot.cs:365-373`).

Since the generated static table derives `MaxHealth` from Strength (+5/pt) and Endurance (+20/pt), any passive touching STR/END/MaxHealth makes the frontend battler start the fight under-full on every **full** rebuild (session start, level-up, equip/loadout/proficiency change — constant events in an idle game), while the data-less re-arm path is correct (the passive modifier is still in the attribute set when `battler.ts:439` runs). An intermittent divergence keyed to which reset path ran is the nastiest kind of parity bug to diagnose — the backend replay would flip client-won battles into rejected claims.

Latent today only because the sole authored class has a 0-amount passive; the V1 class roster (#1226) will surface it. The existing test (`UI/src/tests/.../battle-engine.test.ts:1657-1682`) uses a +7 STR passive but asserts only the Strength value, never `currentHealth`, so nothing catches it.

**Fix direction:** re-snapshot `player.currentHealth = player.attributes.getValue(EAttribute.MaxHealth)` after `applySignaturePassive` on the full-rebuild path (or thread the passive application into the reset so the ordering matches `BattlerMaterials.Build`), and extend the test to pin starting health. Should land before #1226 authors real passives.

ginderjeremiah:
## Summary

`BattleEngine.resetPlayer`'s full-rebuild path called `this.player.reset(...)`, which snapshots `currentHealth = MaxHealth` at the end of `Battler.reset` (`UI/src/lib/battle/battler.ts:439`), and only *then* composed the class signature passive via `applySignaturePassive` (`UI/src/lib/engine/battle/battle-engine.ts`). Since `MaxHealth` derives from Strength/Endurance, any passive touching those attributes left the frontend battler under-full on every full rebuild (session start, level-up, equip/loadout/proficiency change).

The backend does the opposite and correct order: `BattlerMaterials.Build()` adds the passive to the `AttributeCollection` first, then the `Battler` ctor snapshots `CurrentHealth` from the post-passive `MaxHealth` (`Game.Core/Battle/BattleSnapshot.cs`). This was a latent FE/BE parity bug — harmless only because the one currently-authored class has a 0-amount passive — that would surface once a real passive ships (#1226) and could cause the backend replay to reject legitimately-won client battles.

## Fix

- In `BattleEngine.resetPlayer`, re-snapshot `this.player.currentHealth = this.player.attributes.getValue(EAttribute.MaxHealth)` immediately after `applySignaturePassive` runs on the full-rebuild path, matching the backend's ordering.
- Extended the existing signature-passive test (`battle-engine.test.ts`, uses a +7 Strength passive) to assert `currentHealth` equals post-passive `MaxHealth`, both after the full rebuild and after a subsequent data-less re-arm. Confirmed this assertion fails without the fix (900 vs. expected 935) and passes with it.

## Test plan

- [x] `npx vitest run` (full frontend suite) — 297 files / 3659 tests pass
- [x] Manually reverted the fix and confirmed the new assertion fails, then re-applied and confirmed it passes

Closes #1933

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gp6Ezk99G2zZWU57MQD32o)_

### DIFF
--- a/UI/src/tests/lib/engine/battle/battle-engine.test.ts
+++ b/UI/src/tests/lib/engine/battle/battle-engine.test.ts
@@ -1666,6 +1666,9 @@ describe('BattleEngine', () => {
 			// The full rebuild composes the passive last: Strength = 50 (alloc) + 7 (passive), added exactly once.
 			engine.start();
 			expect(engine.player.attributes.getValue(EAttribute.Strength)).toBe(57);
+			// #1933: Strength feeds MaxHealth, so the passive raises it too — the battler must start full
+			// against the post-passive MaxHealth, not the value reset() snapshotted before the passive landed.
+			expect(engine.player.currentHealth).toBe(engine.player.attributes.getValue(EAttribute.MaxHealth));
 
 			// A data-less re-arm (unchanged inputs) skips setData, so the already-composed passive persists — it is
 			// neither lost (→ 50) nor re-applied (→ 64).
@@ -1679,6 +1682,7 @@ describe('BattleEngine', () => {
 				isBossBattle: false
 			});
 			expect(engine.player.attributes.getValue(EAttribute.Strength)).toBe(57);
+			expect(engine.player.currentHealth).toBe(engine.player.attributes.getValue(EAttribute.MaxHealth));
 		});
 
 		// The additionalModifiers passed to reset must be [...lockedBase, ...proficiency] in that order — the
--- a/Game.Core.Tests/Battle/BattleSnapshotTests.cs
+++ b/Game.Core.Tests/Battle/BattleSnapshotTests.cs
@@ -296,6 +296,10 @@ namespace Game.Core.Tests.Battle
 
             // Strength = 6 (free pool) + 4 (flat signature passive) = 10.
             Assert.Equal(10, battler.GetAttributeValue(EAttribute.Strength));
+            // The passive lands in the AttributeCollection before the ctor snapshots CurrentHealth (#1933's FE
+            // counterpart), so a passive that raises MaxHealth (Strength feeds it) must not leave the battler
+            // under-full.
+            Assert.Equal(battler.GetAttributeValue(EAttribute.MaxHealth), battler.CurrentHealth);
         }
 
         [Fact]
--- a/UI/src/lib/engine/battle/battle-engine.ts
+++ b/UI/src/lib/engine/battle/battle-engine.ts
@@ -9,7 +9,7 @@ import {
 } from '$lib/battle';
 import { Mulberry32 } from '$lib/engine/mulberry32';
 import { staticData, playerProficiencies } from '$stores';
-import { ELogType, EDamageType, IBattlerAttribute, IEnemyInstance, ISkillDamagePortion } from '$lib/api';
+import { ELogType, EAttribute, EDamageType, IBattlerAttribute, IEnemyInstance, ISkillDamagePortion } from '$lib/api';
 import { DEFAULT_MAX_BATTLE_MS } from '$lib/api/types/game-constants';
 import { logMessage } from '../log';
 import {
@@ -290,6 +290,10 @@ export class BattleEngine {
 		// composePlayerBattleAttributes canonicalises. The data-less re-arm above keeps this modifier (setData
 		// isn't re-run), so it is applied only here, on a full rebuild.
 		applySignaturePassive(this.player.attributes, (resolve) => playerManager.battleSignaturePassiveModifier(resolve));
+		// reset() already snapshotted currentHealth off MaxHealth before the passive above could touch it
+		// (#1933) — re-snapshot now so a passive that raises MaxHealth doesn't leave the battler under-full,
+		// matching the backend's Battler ctor which snapshots CurrentHealth after the passive is composed in.
+		this.player.currentHealth = this.player.attributes.getValue(EAttribute.MaxHealth);
 		this.lastEquipmentStats = equipmentStats;
 		this.lastPlayerAttributes = attributes;
 		this.lastSelectedSkills = selectedSkills;


UI/src/tests/lib/engine/battle/battle-engine.test.ts
```diff
@@ -1666,6 +1666,9 @@ describe('BattleEngine', () => {
 			// The full rebuild composes the passive last: Strength = 50 (alloc) + 7 (passive), added exactly once.
 			engine.start();
 			expect(engine.player.attributes.getValue(EAttribute.Strength)).toBe(57);
+			// #1933: Strength feeds MaxHealth, so the passive raises it too — the battler must start full
+			// against the post-passive MaxHealth, not the value reset() snapshotted before the passive landed.
+			expect(engine.player.currentHealth).toBe(engine.player.attributes.getValue(EAttribute.MaxHealth));
```
ginderjeremiah, line 1671: **Should-fix (battle-logic parity):** This new assertion pins that starting health reflects a MaxHealth-affecting signature passive on the frontend, but the backend suite has no mirrored assertion. `BattleSnapshotTests.ToBattler_ComposesFlatClassSignaturePassive` already uses a +4 Strength passive (Strength feeds MaxHealth), yet it only asserts `GetAttributeValue(Strength)` — never `CurrentHealth`. The round-trip parity test uses `MakeClass`'s 0-amount passive, so it doesn't exercise a nonzero passive against `CurrentHealth` either.

CLAUDE.md requires battle-logic tests to have "the same scenarios and expected results in both the frontend and backend test suites." A one-liner in that backend test — `Assert.Equal(battler.GetAttributeValue(EAttribute.MaxHealth), battler.CurrentHealth);` — mirrors this scenario and guards the backend against a future regression (e.g. `BattlerMaterials.Build` snapshotting before the passive is added). The backend is correct today by construction, so this is coverage parity, not a defect.

---

ginderjeremiah (commented):
## Review: FE re-snapshots player `currentHealth` after the signature passive on full rebuild (#1933)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating event from me) — one should-fix on backend test parity, no blockers.

The fix is correct and minimal. `Battler.reset` snapshots `currentHealth = MaxHealth` (`battler.ts:437`) before `applySignaturePassive` runs on the full-rebuild path, so any passive touching STR/END/MaxHealth left the FE battler under-full. Re-snapshotting right after the passive composes (`battle-engine.ts:296`) matches the backend exactly: `BattlerMaterials.Build()` adds the passive to the `AttributeCollection` (`BattleSnapshot.cs:367-371`), *then* the `Battler` ctor snapshots `CurrentHealth = _attributes[MaxHealth]` (`Battler.cs:54`). I verified this ordering directly. Leaving the data-less re-arm path untouched is correct — `reset()` with no args skips `setData`, so the already-composed passive persists in the attribute set and `battler.ts:437` snapshots the post-passive MaxHealth; the second test assertion pins exactly that.

### Should-fix

- Backend battle-logic parity — `battle-engine.test.ts:1671` / `BattleSnapshotTests.cs`. The new FE assertion (`currentHealth == post-passive MaxHealth`) has no mirrored backend counterpart; the backend passive tests assert only attribute values, and the round-trip test uses the 0-amount passive. CLAUDE.md requires mirrored battle-logic scenarios/results. A one-line `Assert.Equal(battler.GetAttributeValue(EAttribute.MaxHealth), battler.CurrentHealth);` in `ToBattler_ComposesFlatClassSignaturePassive` (which already uses a MaxHealth-affecting +4 STR passive) closes the gap and guards against a future backend reorder. See the inline comment.

### Test coverage

- FE: the existing signature-passive test is extended to pin `currentHealth` both after the full rebuild and after a data-less re-arm — precisely the two paths the bug distinguishes. Good, targeted coverage.
- BE: correct by construction and broadly covered, but missing the mirrored `CurrentHealth` assertion (the should-fix above).

### What's good

- Root cause correctly identified as an ordering divergence, not a value bug, with the backend cited as the source of truth.
- The two-line explanatory comment at the fix site is appropriately scoped and earns its length by naming the parity contract.
- Reverting the fix to confirm the new assertion fails (900 vs 935) is the right way to prove the test actually guards the behavior.

---

ginderjeremiah:
## Re-review (delta since prior review): `dc445b3`

The follow-up commit `dc445b3` addresses the sole should-fix from the earlier review. `ToBattler_ComposesFlatClassSignaturePassive` now pins the backend parity assertion:

```csharp
Assert.Equal(battler.GetAttributeValue(EAttribute.MaxHealth), battler.CurrentHealth);
```

This mirrors the new frontend assertion (`currentHealth == post-passive MaxHealth`) on the backend's already-correct path (`BattlerMaterials.Build()` composes the passive into the `AttributeCollection` before the `Battler` ctor snapshots `CurrentHealth`), satisfying CLAUDE.md's mirrored battle-logic-scenario requirement and guarding against a future backend reorder. The two-line explanatory comment is appropriately scoped.

CI (`Run tests`, run 2387) is green on the head SHA. No remaining findings — no follow-ups needed.

---

Issue #1736, opened 2026-07-08
EnemyManager's concurrency guards are accreting — consolidate into a single supersedable-operation primitive when next touched


ginderjeremiah:
## Problem

`UI/src/lib/engine/battle/enemy-manager.ts:74-106` now juggles four interlocking concurrency mechanisms: `transitioning` + `transitionGeneration`, `fetchGeneration` + `fetchingGeneration`/`inFlightFetch`, the cancellable backoff, and per-resolution `bossLoopActive`/`transitionCurrent` re-checks after each await.

It is correct as far as adversarial reading can trace, and very well tested (the boss suite alone is ~1100 lines). But the last several fixes (#1696, #1712, #1690) were each another race patched with another guard — empirical evidence the ad-hoc shape breeds regressions.

## Ask

**Not** a pre-emptive refactor. The next time a feature touches this loop, consolidate the guards into a single supersedable-operation primitive (one generation counter, one "am I still current?" check) rather than adding a fifth mechanism. The existing test suite is the safety net.

_Found during a codebase health review._

ginderjeremiah:
One concrete instance to fold into the consolidation when it happens: in `enemy-manager.ts:199-221`, a caller that awaits a superseded `inFlightFetch` re-checks `started`/generation (`:214`) but not `fetchingEnemy` — two same-generation callers waking together can both enter `runFetchLoop` and double-send `NewEnemy`. Hard to hit in practice; belongs in the supersedable-operation primitive rather than a fifth point guard.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1950, opened 2026-07-14
Loop EnemyManager.getNewEnemy's re-entrancy guard to catch a queued duplicate

### LINKED ISSUE #1934: EnemyManager: two same-generation callers queued behind a superseded fetch both proceed — double NewEnemy spawn defeats the re-entrancy guard
`EnemyManager.getNewEnemy`'s re-entrancy guard (`UI/src/lib/engine/battle/enemy-manager.ts:206-228`) re-checks only `started`/generation after waiting out a superseded in-flight fetch — never `fetchingEnemy`:

```ts
if (this.fetchingEnemy) {
    if (this.fetchingGeneration === generation) { return; }
    await this.inFlightFetch;
    if (!this.started || generation !== this.fetchGeneration) { return; }
}
const run = this.runFetchLoop(generation, prepared, forceAbandon);
```

If **two** callers with the same (current) generation queue behind an older-generation fetch — e.g. an in-flight `NewEnemy` is slow (the socket request budget is 30s), the player parks in Home (`enterHome` → `interruptFetch` bumps the generation, `rest()` → Idle stage → `watchIdleStage` calls `getNewEnemy`), then leaves Home (`navigateToZone` → second `getNewEnemy`) — both resume when `inFlightFetch` settles, both pass the started/generation check, and both launch `runFetchLoop` with the same generation. Neither loop's `generation !== this.fetchGeneration` guard trips, so two concurrent `NewEnemy` commands go out and both responses spawn (`notifyNewEnemyLoaded` twice, `battleEngine.reset` twice) — exactly the double-spawn with anti-cheat consequences the guard's own comment documents (`:209-214`), plus the second command reporting a bogus `clientBattleMs` for the first battle.

**Fix direction:** loop the guard — `while (this.fetchingEnemy) { if (this.fetchingGeneration === generation) return; await this.inFlightFetch; if (!this.started || generation !== this.fetchGeneration) return; }` — so the second waiter sees the first's new same-generation loop and drops as a re-entrant duplicate. (Also strengthens the case for #1736's supersedable-operation consolidation, but this is a concrete guard hole today, worth closing independently.)

ginderjeremiah:
## Summary

`EnemyManager.getNewEnemy`'s re-entrancy guard (`UI/src/lib/engine/battle/enemy-manager.ts`) only checked `fetchingEnemy`/`fetchingGeneration` once (`if`, not `while`). Two same-generation callers could both queue behind an older, already-superseded fetch:

1. A fetch for an old generation is in flight, then a transition (e.g. entering Home, or `challengeBoss`) bumps the generation, superseding it.
2. Two callers at the new generation both arrive while the superseded fetch is still tearing down — both see `fetchingEnemy === true` with a stale `fetchingGeneration`, so both `await this.inFlightFetch`.
3. When the superseded fetch settles, the first waiter resumes, passes its generation check, and synchronously restarts the fetch — re-arming `fetchingEnemy`/`fetchingGeneration` for the *current* generation.
4. The second waiter resumes right after, but its one-shot guard already evaluated `fetchingEnemy` back in step 2 — it never re-checks, so it falls through and launches a second `NewEnemy` request for the same generation: a double enemy spawn, which the backend's anti-cheat replay would flag.

## Fix

Turned the one-shot `if (this.fetchingEnemy)` into a `while (this.fetchingEnemy)` loop, per the issue's suggested direction. The second waiter now loops back after its wait and sees the first waiter's restarted fetch as a genuine re-entrant duplicate (same generation) and drops, rather than racing it.

## Test plan

- [x] Added a regression test (`drops the second of two same-generation callers queued behind a superseded fetch`) that reproduces the exact race — verified it fails against the old code (3 `sendSocketCommand` calls instead of 2) and passes with the fix.
- [x] `npm run test:unit:ci` — all 3660 tests pass.

Closes #1934

---
_Generated by [Claude Code](https://claude.ai/code/session_013KZhyB3dLCDug6QgwJjxEs)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -205,7 +205,12 @@ export class EnemyManager {
 	 */
 	public async getNewEnemy(prepared?: PreparedBattle, forceAbandon = false): Promise<void> {
 		const generation = this.fetchGeneration;
-		if (this.fetchingEnemy) {
+		// Looped (not a one-shot if): a fetch already running when we arrive is checked again after each
+		// wait. Two same-generation callers can both queue behind an older, superseded fetch — the first
+		// to resume starts a fresh fetch (synchronously re-arming fetchingEnemy/fetchingGeneration before
+		// its first await), and looping back here is what lets the second recognize that new fetch as its
+		// own genuine re-entrant duplicate instead of racing it into a double spawn.
+		while (this.fetchingEnemy) {
 			// A fetch is already running. If it captured this same generation it's a genuine re-entrant
 			// duplicate — overlapping stage handlers (e.g. an idle victory racing an Idle/Defeated change)
 			// each request-and-notify a spawn, and a double-spawn has anti-cheat consequences (the backend
@@ -215,7 +220,7 @@ export class EnemyManager {
 			}
 			// Otherwise this call's generation is newer, so the running fetch has been superseded (a
 			// transition bumped the generation) and will abandon without spawning. Wait for it to tear
-			// down, then fall through and fetch fresh so this fallback spawn isn't dropped by the guard.
+			// down, then loop back and re-check rather than assuming the guard is now clear.
 			await this.inFlightFetch;
 			// The wait can overlap a stop or a further supersession; bail if either landed meanwhile.
 			if (!this.started || generation !== this.fetchGeneration) {
--- a/UI/src/tests/lib/engine/battle/enemy-manager.test.ts
+++ b/UI/src/tests/lib/engine/battle/enemy-manager.test.ts
@@ -292,6 +292,45 @@ describe('EnemyManager.getNewEnemy', () => {
 		expect(loaded).toEqual([]);
 	});
 
+	// #1934: two same-generation callers can both queue behind an older, superseded fetch (e.g. entering
+	// Home bumps the generation while a NewEnemy is in flight, then leaving Home calls getNewEnemy twice
+	// before the superseded fetch tears down). Both must not fall through past the guard and launch the
+	// current generation's fetch — the first genuinely restarts it, the second must recognize that as its
+	// own re-entrant duplicate and drop.
+	it('drops the second of two same-generation callers queued behind a superseded fetch', async () => {
+		let resolveOld!: (r: IApiSocketResponse<'NewEnemy'>) => void;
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Loop `EnemyManager.getNewEnemy`'s re-entrancy guard to catch a queued duplicate (#1934)

**Verdict:** Approve — the fix correctly closes the guard hole described in #1934, is minimal and well-reasoned, and ships with a regression test that reproduces the exact race. (Posted as a comment since I'm the PR author; treat it as an approval.)

I traced the concurrency carefully and the fix is sound:

- Two same-generation callers queued behind a superseded fetch both `await this.inFlightFetch`. When it settles, the first waiter resumes, exits the `while` (guard released via `runFetchLoop`'s `finally { fetchingEnemy = false }`), and calls `runFetchLoop`, which **synchronously** re-arms `fetchingEnemy`/`fetchingGeneration` for the current generation before its first `await`. The second waiter resumes next, loops back, sees that same-generation in-flight fetch, and drops as a genuine re-entrant duplicate — exactly the intended behavior. Microtask FIFO guarantees the first-registered waiter re-arms before the second re-checks, and the property is symmetric (whichever resumes first wins), so it holds regardless of caller order.
- No infinite-loop foot-gun: the loop only continues while `generation === this.fetchGeneration` (line 226 bails otherwise), and any newly-started fetch is at the current generation, so it resolves to either a same-generation drop or a released guard. `fetchingGeneration` can never exceed `fetchGeneration`, so the loop is bounded.

### Test coverage

- `drops the second of two same-generation callers queued behind a superseded fetch` mirrors the #1934 scenario precisely (in-flight old-gen fetch → `interruptFetch` supersession → two same-gen waiters → asserts exactly one `NewEnemy` command goes out). The PR notes it was verified to fail against the old one-shot `if` (3 calls) and pass with the loop (2 calls). Correctly placed under `tests/` mirroring the source path.
- No backend mirror needed: this is client-side socket-fetch orchestration, not battle-simulation logic, so the frontend/backend battle-consistency rule doesn't apply here.

### What's good

- The expanded comments accurately capture the subtle "synchronous re-arm before the second waiter re-checks" reasoning — this is precisely the kind of deliberate construct a later edit might naively "simplify" back into an `if`, and the comment guards against that.
- Tight, surgical change that closes the hole independently without waiting on the larger #1736 consolidation.

Clean PR — nothing to change.

---

PR #1951, opened 2026-07-14
Derive coverage-floors.mjs area buckets from coverage.config.ts, remove unused micromatch

### LINKED ISSUE #1812: Frontend tooling cleanups: coverage-floors.mjs hand-duplicates the area globs and drops unmatched files; micromatch is an unused production dependency
Two small frontend tooling cleanups, bundled:

## 1. `coverage-floors.mjs` re-declares the area globs and silently drops unmatched files

`UI/scripts/coverage-floors.mjs:19-28` hardcodes the eight coverage areas as predicates instead of deriving them from `coverageThresholds` in `coverage.config.ts`, and `if (!hit) continue` (lines 41-43) silently skips any file matching no area. When a new area is added — which `src/tests/coverage-config.test.ts` *forces* whenever a new source folder appears — the reseed helper silently omits it from the ratchet suggestions rather than erroring. Derive the buckets from the shared config (and error on an unmatched file).

## 2. `micromatch` is an unused production dependency

`UI/package.json` lists `micromatch` under `dependencies`, but no file under `UI/src`, `UI/scripts`, `UI/e2e-tests`, or the configs imports it — `coverage-config.test.ts` explicitly hand-rolls its glob matcher to avoid "an untyped runtime glob dependency". It ships as a runtime dep of the node-adapter build for nothing. Remove it.

ginderjeremiah:
## Summary

Two small, bundled frontend tooling cleanups from #1812:

1. **`coverage-floors.mjs` re-declared the area globs and silently dropped unmatched files.** It hardcoded its own copy of the eight coverage areas as ad-hoc predicates (including a special-cased `api/types` exclusion) instead of deriving them from `coverageThresholds` in `coverage.config.ts`, and `if (!hit) continue` silently skipped any covered file matching no area. `src/tests/coverage-config.test.ts` *forces* a new area to be added whenever a new top-level `src` folder appears, so `coverage-floors.mjs` could silently omit that new area from its ratchet suggestions.
2. **`micromatch` was an unused production dependency.** Nothing under `UI/src`, `UI/scripts`, `UI/e2e-tests`, or the configs imports it — `coverage-config.test.ts` explicitly hand-rolls its own glob matcher (`globToRegExp`) specifically to avoid "an untyped runtime glob dependency". Removed from `package.json`/`package-lock.json`.

## Changes

- Extracted `globToRegExp` out of `coverage-config.test.ts` into a new shared `UI/scripts/coverage-glob.ts` (excluded from coverage like the rest of `scripts/**`), so the test and the ratchet script use identical glob-matching logic instead of drifting.
- `coverage-floors.mjs` now derives its area list from `Object.keys(coverageThresholds)` (imported from `coverage.config.ts`) and applies the same `projectCoverageExclude` filtering the test does, instead of a hand-maintained predicate array. Node 26's native TS support lets the `.mjs` script import the `.ts` config/glob modules directly — no build step needed.
- A covered file that matches no area (and isn't excluded) now makes the script `exit(1)` with the offending paths listed, instead of being silently dropped.
- Removed `micromatch` from `UI/package.json` and regenerated `package-lock.json` via `npm install`.

## Test plan

- [x] `npm run test:unit:ci` — 297 test files / 3660 tests pass, coverage thresholds hold.
- [x] `npm run test:unit:coverage && node scripts/coverage-floors.mjs` — runs end-to-end against a real `coverage-summary.json`, prints the same 8 areas with no "unmatched" errors, output format unchanged from before.
- [x] Confirmed `micromatch` has no remaining references anywhere in `UI/` after removal.

Closes #1812

---
_Generated by [Claude Code](https://claude.ai/code/session_01CLRJoagyPsBiFjFexgpepR)_

### DIFF
--- a/UI/scripts/coverage-floors.mjs
+++ b/UI/scripts/coverage-floors.mjs
@@ -30,11 +30,13 @@ function blank() {
 }
 
 // Strip everything before the `src/` root so an absolute (or Windows-backslashed) path from the
-// summary matches the same repo-relative form the area/exclude globs are anchored against.
+// summary matches the same repo-relative form the area/exclude globs are anchored against. Anchors
+// on `/src/` (leading slash) rather than `src/` so a parent directory ending in `src` (e.g.
+// `.../mysrc/UI/src/lib/...`) doesn't match its own trailing `src/` prematurely.
 function toRelative(file) {
 	const norm = file.split('\\').join('/');
-	const idx = norm.indexOf('src/');
-	return idx === -1 ? norm : norm.slice(idx);
+	const idx = norm.indexOf('/src/');
+	return idx === -1 ? norm : norm.slice(idx + 1);
 }
 
 const unmatched = [];
--- a/UI/scripts/coverage-floors.mjs
+++ b/UI/scripts/coverage-floors.mjs
@@ -4,6 +4,8 @@
 //   node scripts/coverage-floors.mjs
 // Re-run when coverage rises to ratchet the floors upward.
 import { readFileSync } from 'node:fs';
+import { coverageThresholds, projectCoverageExclude } from '../coverage.config.ts';
+import { globToRegExp } from './coverage-glob.ts';
 
 const summaryPath = new URL('../coverage/coverage-summary.json', import.meta.url);
 let summary;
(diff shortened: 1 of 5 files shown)


UI/scripts/coverage-floors.mjs
```diff
@@ -14,52 +16,68 @@ try {
...
+const acc = Object.fromEntries(AREAS.map((k) => [k, blank()]));
 function blank() {
 	return Object.fromEntries(METRICS.map((m) => [m, [0, 0]]));
 }
 
+// Strip everything before the `src/` root so an absolute (or Windows-backslashed) path from the
+// summary matches the same repo-relative form the area/exclude globs are anchored against.
+function toRelative(file) {
+	const norm = file.split('\\').join('/');
+	const idx = norm.indexOf('src/');
```
ginderjeremiah: Nit (optional): `indexOf('src/')` anchors on `src/` without a leading slash, so a parent directory that ends in `src` (e.g. `.../mysrc/UI/src/lib/...`) would slice at the wrong `src/` and yield a path the area globs can't match — a false `exit(1)`. The pre-refactor predicates guarded against this by matching `/src/lib/...` with the leading slash. Anchoring on `/src/` (`norm.indexOf('/src/')` then `slice(idx + 1)`) restores that guard; an already-relative `src/lib/...` path still falls through the `=== -1` branch unchanged and matches fine. It won't bite CI paths in practice, so purely a robustness suggestion.

---

ginderjeremiah (commented):
## Review: coverage-floors.mjs derives area buckets from coverage.config.ts; removes unused `micromatch` (closes #1812)

**Verdict:** Approve (intent) — both cleanups from #1812 are implemented correctly and behavior-preservingly. Submitted as a comment because I'm the PR author, so GitHub won't accept a gating approval on my own PR. Only one optional nit below; no blocking or should-fix findings.

I verified the change against the issue's two asks and the surrounding config:

- **Bucket derivation is a genuine de-dup, not a re-copy.** `coverage-floors.mjs` now reads `Object.keys(coverageThresholds)` and reuses the extracted `globToRegExp` + `projectCoverageExclude`, the exact same source `src/tests/coverage-config.test.ts` gates against. A new top-level area added to `coverageThresholds` (which the gate test *forces*) is now picked up by the ratchet script automatically — the drift the issue describes is closed.
- **The old `api/types` special-case is preserved, not lost.** The removed hand-rolled predicate excluded `src/lib/api/types/**`; that exclusion now lives in `projectCoverageExclude` and is applied *before* area matching, so those codegen'd files are excluded (not silently dropped) and don't trip the new unmatched check. Net bucketing is identical to before.
- **Silent-drop footgun replaced with a hard failure.** An unexcluded file matching no area now collects into `unmatched` and `exit(1)`s with the offending paths, instead of `if (!hit) continue`.
- **`globToRegExp` extraction keeps its coverage.** Moving it to `scripts/coverage-glob.ts` (verbatim) means both the gate test and the ratchet script share one matcher and can't drift; the function stays exercised by `coverage-config.test.ts` which imports it.
- **`micromatch` removal is clean.** No references remain anywhere under `UI/` (confirmed), and the lockfile prune (braces/fill-range/is-number/picomatch/to-regex-range) is the expected transitive fallout.

### Nits

- `UI/scripts/coverage-floors.mjs:36` — `toRelative` anchors on `src/` without a leading slash; anchoring on `/src/` would match the old predicates' robustness against a parent dir ending in `src`. Optional; won't affect real CI paths. (inline)

### Test coverage

Adequate for the scope. `coverage-glob.ts` retains its existing coverage via `coverage-config.test.ts`. `coverage-floors.mjs` is dev tooling under `scripts/**` (explicitly excluded from the coverage gate) and is a manually-run reseed helper, so leaving `toRelative`/aggregation unit-tested is consistent with the project's testing guidelines — no new untested domain/lib logic is introduced.

### What's good

Tidy, well-scoped tech-debt cleanup that turns a hand-maintained duplicate into a single derived source and swaps a silent gap for a loud failure — exactly the intent of #1812. Clear comments explaining the *why* on each moved piece.

_I'm holding off on auto-merge only because of the one nit above (your call to take or leave) and because the e2e/frontend test jobs were still in progress at review time — nothing blocking on my end._

---

PR #1953, opened 2026-07-14
Close the socket on a corroborated request timeout for faster half-open recovery

### LINKED ISSUE #1765: Half-open (dead but readyState OPEN) socket is never detected — pings are sent but pongs are never checked

### LINKED ISSUE #1828: Close the socket on a request timeout to speed half-open recovery in the active case
## Context

Follow-up from PR #1797 (closed #1765), which added missed-pong detection: after `MAX_MISSED_PONGS` consecutive unanswered keepalive pings, `api-socket.ts` closes the socket so `handleClose` reconnects.

## Problem

That covers the **idle** half-open case (no commands in flight, so only pings expose the dead connection). Issue #1765 also called out a second trigger: when commands *are* in flight on a half-open socket, each one still eats its full 30s request timeout before anything reacts, and `timeoutRequest` settles the request's promise but leaves the "open" socket alone.

## Suggested fix

When a request times out (`timeoutRequest`) on a socket still reporting `OPEN`, treat it as a signal the connection may be half-open and close the socket to trigger the existing reconnect path — with care not to close on a single unrelated slow response (e.g. require the timeout to coincide with the missed-pong signal, or a small threshold), so a genuinely slow round trip doesn't tear down a healthy socket.

## Nice-to-have (minor, from the same review)

The `awaitingPong = false; missedPongs = 0;` reset is duplicated in three places in `api-socket.ts`; a small `resetPongTracking()` helper would DRY it.

## Acceptance

- A half-open socket with in-flight commands recovers faster than the current worst case (one full request timeout).
- A single slow-but-alive round trip does not close a healthy socket.
- Test coverage for both.

ginderjeremiah:
## Summary

Follow-up from PR #1797 (closed #1765). That PR added missed-pong-based half-open detection, but only covers the *idle* case — the keepalive ping alone drives reconnect. Issue #1828 called out the second trigger: when commands are in flight on a half-open socket, each one silently eats its full 30s `REQUEST_TIMEOUT_MS` before anything reacts, since `timeoutRequest` only settles the request's promise and leaves the socket alone.

- `timeoutRequest` now calls a new `closeIfLikelyHalfOpen()`, which closes the socket **only when the timeout coincides with an already-missed pong** (`missedPongs > 0`) — reusing the existing half-open signal rather than adding a new counter/threshold. A single legitimately slow command on an otherwise healthy connection (`missedPongs === 0`) is left alone.
- Extracted a shared `closeHalfOpenSocket()` (reset pong tracking + code-less `socket.close()`) used by both `attemptPing`'s existing missed-pong branch and the new `closeIfLikelyHalfOpen()`.
- Nice-to-have from the issue: deduped the `awaitingPong = false; missedPongs = 0;` reset (previously repeated 3x) into `resetPongTracking()`.
- Updated `docs/frontend-sockets.md`'s socket-lifecycle notes to describe the new corroborated-timeout close.

- **"Half-open socket with in-flight commands recovers faster than current worst case"** — a timeout landing while a pong is overdue closes the socket immediately instead of waiting out the rest of `attemptPing`'s own missed-pong window (worst case several extra ping cycles).
- **"A single slow-but-alive round trip doesn't close a healthy socket"** — the close is gated on `missedPongs > 0`, so a timeout with no pongs currently missing (the common "one heavy command" case) never triggers it.
- **Test coverage for both** — see test plan below.

## Test plan

- [x] New unit test: a request timeout that coincides with an already-missed pong closes the socket.
- [x] New unit test: a request timeout with every keepalive pong answered promptly (no missed pong) does **not** close the socket.
- [x] Full frontend suite: `npm run lint` and `npm run test:unit:ci` — 297 files / 3662 tests pass, no coverage regressions.

Closes #1828

---
_Generated by [Claude Code](https://claude.ai/code/session_01W4Lgxzvx1DWjTmT7jgM9QC)_


ginderjeremiah (commented):
## Review: Close the socket on a corroborated request timeout for faster half-open recovery (#1828)

**Verdict:** Approve — implements the issue's second half-open trigger exactly as scoped, with the missed-pong corroboration guard the acceptance criteria required, and covers both branches with mirrored tests.

_(Submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author; the intended verdict is **Approve**.)_

### What's good

- **Correct corroboration gate.** `closeIfLikelyHalfOpen` closes only on `readyState === OPEN && missedPongs > 0`, which is precisely the "coincide with the missed-pong signal" approach the issue asked for. A genuinely half-open socket accrues `missedPongs > 0` within one ≤10s ping cycle, so a 30s request timeout reliably corroborates; a single slow-but-alive command keeps `missedPongs === 0` and is correctly left alone.
- **No double-close hazard.** The `MAX_MISSED_PONGS` branch resets tracking (`missedPongs → 0`) and moves `readyState` off OPEN before this path can run, so a subsequent timeout is a clean no-op — and the request that timed out is already pruned from `inFlightRequests` before the close, so `handleClose` settling the rest with `CONNECTION_LOST_ERROR` is the desired outcome (other in-flight commands fail fast and reconnect rather than each eating its own 30s).
- **Good refactors.** `resetPongTracking()` DRYs the reset across all three sites (the issue's nice-to-have), and `closeHalfOpenSocket()` cleanly shares the reset + code-less close between the two triggers while preserving the critical "must surface as 1005/1006, not NORMAL_CLOSURE" comment.

### Test coverage

- Both acceptance criteria are covered: a timeout coinciding with an overdue pong closes the socket, and a timeout with every pong answered promptly does not. Tests mirror the source path (`tests/lib/api/api-socket.test.ts`) and follow the existing fake-timer / classical style in the file. This is a frontend-only transport change — no battle logic, so no backend mirror is needed.

### Nits

- None.


---

PR #1954, opened 2026-07-14
Memoize the progress hash read per command scope (#1820)

### LINKED ISSUE #1820: Battle-lifecycle commands read the full Progress_{playerId} hash twice per command — the #1757 pattern on hotter paths
## Problem

Each battle-lifecycle command pays two `HGETALL`s (plus two fire-and-forget TTL slides and full row deserialization) of the same progress hash — the exact redundancy #1757 removed from `EquipItem`, on higher-frequency paths:

- **Idle victory (`DefeatEnemy`, the hottest path):** `BattleStatisticsEventHandler.cs:23` (`Load` → full hash read + mapping), then the bundled prefetch `BattleService.cs:575-579` (`CaptureProficiencyLevels` → `GetProficiencies` → second full read of the same key moments later in the same command). Same for the `BattleLost` prefetch.
- **`StartBossBattle`:** the unlock-gate read (`ZoneResolutionService.IsZoneUnlocked`, `ZoneResolutionService.cs:134-143`) + the snapshot capture (`BattleService.cs:212`).
- **Explicit zone change in `StartBattle`:** unlock read (`BattleService.cs:85`) + snapshot capture (`BattleService.cs:112`).

The docs deliberately accept *one* per-battle progress load; the second is pure redundancy, and the hash grows with per-enemy/per-skill statistic rows, so the cost scales with account age on the highest-frequency command.

## Suggested fix

A scoped per-command memo in `PlayerProgressRepository.GetCachedProgress` (invalidated by `Save`) is the small variant; alternatively thread the handler-loaded aggregate/proficiency levels into the prefetch the way `OfflineProgressService` already threads one aggregate through its whole pass (its stated rationale — "replaces 3-4 serial round-trips" — applies equally live).

ginderjeremiah:
## Summary

Several battle-lifecycle commands paid two full `HGETALL`s of the same `Progress_{playerId}` Redis hash within a single command:

- **Idle victory (`DefeatEnemy`, the hottest path):** `BattleStatisticsEventHandler.Load` reads the hash, then the bundled next-idle-battle prefetch (`BattleService.StartBattle` → `CaptureProficiencyLevels` → `GetProficiencies`) reads it again moments later. Same for `BattleLost`.
- **`StartBossBattle`** and the explicit zone-change path in **`StartBattle`**: the zone-unlock gate (`GetCompletedChallengeIds`) and the battle-snapshot proficiency capture (`GetProficiencies`) each hit the hash separately.

`PlayerProgressRepository` is registered `Scoped`, and a socket command opens exactly one DI scope per command, so the repository instance is shared across all of these calls within one command. This PR adds a per-scope memo:

- `GetCachedProgress` returns the memoized snapshot on a repeat call within the same scope instead of re-reading Redis.
- `Save` keeps the memo warm by **merging in the rows it just changed** rather than invalidating it — the hottest case (idle victory) reads again *after* a `Save`, so an invalidate-only memo wouldn't actually collapse that read.
- The merge is bundled into the same deferred callback as the cache's write-behind advance (not run eagerly), so a flush that never lands — a shared-batch failure or a standalone flush's transient error — can't leak an uncommitted row into a scope that reads progress again after the failure. (Caught this with the existing `OfflineProgressServiceIntegrationTests.SimulateOfflineProgress_SharedFlushFails_...` test during development — my first pass merged eagerly and leaked pre-flush-failure state into the memo.)

## Test plan

- [x] Added `GetStatistics_CalledTwiceInTheSameScope_ServesTheSecondCallFromTheScopedMemo` — proves the memo is scoped per repository instance, not global (a fresh scope still sees an out-of-band change).
- [x] Added `Save_ThenGetProficienciesInTheSameScope_ReflectsTheJustSavedValueImmediately` — proves a same-scope read right after `Save` sees the new value deterministically, without racing the fire-and-forget `HSET` (mirrors the idle-victory prefetch).
- [x] `dotnet build` — clean.
- [x] `dotnet test` (full solution) — 1351 + 62 + 957 + 759 tests pass, including the full `Game.Application.Tests` and `Game.Api.Tests` integration suites.
- [x] Updated `docs/backend-persistence.md` with a short note on the memo/merge design.

Closes #1820

---
_Generated by [Claude Code](https://claude.ai/code/session_01TqsQsphRMVVC9oBRjbLgNQ)_


ginderjeremiah (commented):
## Review: per-scope progress memo (#1820)

**Verdict:** Approve — a clean, behavior-preserving fix that collapses the redundant second `HGETALL` on the battle-lifecycle hot paths. (Submitted as a comment since GitHub won't accept a formal approval from the PR author; the verdict is an approve.)

The design's load-bearing property is that the memo is advanced in the *same* `AdvanceCache` callback as the cache `HSET` — miss-reload sets both, a hit builds the memo from `FromHashFields`, and `Save` merges the memo right next to the conditional `HSET`. So the memo is always a faithful mirror of what a cache read would have returned at that instant, whether `AdvanceCache` runs synchronously (standalone save) or deferred via `OnFlushed` (the batch hot path). That's what makes choosing *merge* over *invalidate* correct for the prefetch-reads-after-save case the issue calls out, and it means there's no behavior change versus the old fresh-read besides dropping the round-trip.

Points I checked and am satisfied with:

- **No uncommitted-row leak** — the merge is bundled into `AdvanceCache`, which only runs after a successful flush. Standalone flush failure throws before it; a batch flush failure never fires `OnFlushed`, so the memo stays at pre-save state. Good call bundling it rather than merging eagerly after `AcceptChanges`.
- **No aliasing** — `changed` is a fresh `ToCached` projection, already serialized into the event payload before `MergeInto` places its rows in the memo; aggregate mutations only touch the Core objects, never the cached rows.
- **Merge identity** matches the hash-field natural keys exactly (stat type+entity with `null==null`, challenge id, proficiency id) — upsert-in-place, no duplication. Absolute-value upserts mean no delete path is needed.
- **Sliding TTL** is still slid by the first read of each scope; only redundant intra-command slides are dropped, which is the intent.

### Test coverage

Adequate. `Save_ThenGetProficienciesInTheSameScope_...` proves the merge-keeps-warm semantics without racing the fire-and-forget HSET, and `GetStatistics_CalledTwiceInTheSameScope_...` proves the memo is per-scope while a fresh scope still observes out-of-band changes. The deferred/batch path's failure-leak case is exercised by the existing offline-progress suite, and since the merge code is path-independent that's sufficient.

### What's good

Careful, well-documented reasoning throughout — especially the invalidate-vs-merge rationale and the deliberate bundling of the merge into the post-flush callback. The docs note in `backend-persistence.md` accurately captures the design.


---

PR #1955, opened 2026-07-14
Only dirty-mark Max/Min statistics on a genuine value change

### LINKED ISSUE #1821: Max/Min statistic writes dirty-mark rows even when nothing changed, inflating the write-behind payload and challenge-evaluation scope every battle
## Problem

`PlayerProgress.GetOrCreate` (`Game.Core/Progress/PlayerProgress.cs:376-392`) unconditionally adds the key to `_dirtyStatistics`, so a `SetMax`/`SetMin` that does **not** improve the value still (a) re-enqueues the unchanged row in the write-behind persist and (b) puts the key into the touched set `RecordBattleCompleted` returns — which is the challenge-evaluation relevance scope (`ChallengeIndex.RelevantTo`).

Every battle records `HighestSingleAttackDamage` (global + one per used skill) and every victory records `FastestVictory` (global + per-enemy). After the first battle these almost never improve, yet each battle persists ~4-6 unchanged rows and re-evaluates every TimeTrial/HighestAttack challenge keyed to them.

The in-code comment (lines 386-389) justifies over-marking only via "the persist is an idempotent absolute write" — it doesn't account for the dirty set's second role as the challenge-eval scope. As a result the method's own contract ("those keys are exactly this battle's mutated statistics") and `docs/backend.md`'s claim that challenge evaluation is scoped to statistics that "actually changed that battle" are both slightly untrue; fixing the code makes the docs true.

## Suggested fix

Move the `_dirtyStatistics.Add` out of `GetOrCreate` into the mutators, marking only on `created || value changed` (`Increment` already effectively does this via the zero-delta skip).

ginderjeremiah:
## Summary

`PlayerProgress.GetOrCreate` (`Game.Core/Progress/PlayerProgress.cs`) unconditionally added the `(type, entityId)` key to `_dirtyStatistics`, so a `SetMax`/`SetMin` call that didn't actually improve the value still:

- re-enqueued the unchanged row in the write-behind persist, and
- entered the key into the touched set `RecordBattleCompleted` returns, which is the challenge-evaluation relevance scope (`ChallengeIndex.RelevantTo`).

`HighestSingleAttackDamage` and `FastestVictory` are recorded on nearly every battle (global + per-skill/per-enemy), and after the first battle they almost never improve — so every battle was persisting several unchanged rows and re-evaluating challenges keyed to them for no reason.

The in-code comment justified this as harmless ("the persist is an idempotent absolute write"), but it didn't account for the dirty set's second role as the challenge-eval scope, and it contradicted both `GetOrCreate`'s own doc comment ("marking dirty here covers exactly the mutated stats") and `docs/backend.md`'s claim that challenge evaluation is scoped to statistics that actually changed that battle.

## Fix

Moved `_dirtyStatistics.Add` out of `GetOrCreate` and into `SetMax`/`SetMin` themselves, marking dirty only when `created || value` actually improved — mirroring the zero-delta skip `Increment` already does for Sum statistics. `Increment` now marks dirty unconditionally after applying its (already zero-checked) delta, since it always mutates when reached.

No doc changes needed — `docs/backend.md`'s existing claim about challenge-eval scoping is now accurate rather than aspirational.

## Test plan

- [x] Added `PlayerProgressTests`: unimproved Max/Min on an existing row stays out of the persist set and the touched keys (mirroring the existing zero-delta-Sum coverage); an improving Max still records/dirties/touches.
- [x] `dotnet test` (full suite, including integration tests against the session's Postgres/Redis containers) — 3132/3132 passing.

Closes #1821

---
_Generated by [Claude Code](https://claude.ai/code/session_01TN1pfg1sDLw8sN2VGejFyW)_

### DIFF
--- a/Game.Core.Tests/Progress/PlayerProgressTests.cs
+++ b/Game.Core.Tests/Progress/PlayerProgressTests.cs
@@ -976,6 +976,61 @@ namespace Game.Core.Tests.Progress
         }
 
         [Fact]
+        public void RecordBattleCompleted_UnimprovedMaxStatisticOnExistingRow_LeavesItCleanAndUntouched()
+        {
+            // An aggregate loaded with a prior HighestSingleAttackDamage records a battle whose attack
+            // doesn't beat it: the existing row's value is unchanged, so — like the zero-delta Sum case —
+            // it must stay out of both the persist set and the touched keys, not just get a no-op re-mark.
+            var progress = MakeProgress(statistics: [Stat(EStatisticType.HighestSingleAttackDamage, null, 40m)]);
+
+            var touched = progress.RecordBattleCompleted(MakeEnemy(), victory: true, playerDied: false,
+                totalMs: 1000, new BattleStats { HighestPlayerAttack = 25.0 }, isBossBattle: false, zoneId: 0);
+
+            Assert.Equal(40m, progress.GetStatisticValue(EStatisticType.HighestSingleAttackDamage, null));
+            Assert.DoesNotContain(progress.DirtyStatistics, s => s.Type == EStatisticType.HighestSingleAttackDamage);
+            Assert.DoesNotContain((EStatisticType.HighestSingleAttackDamage, (int?)null), touched);
+        }
+
+        [Fact]
+        public void RecordBattleCompleted_UnimprovedMinStatisticOnExistingRow_LeavesItCleanAndUntouched()
+        {
+            // An aggregate loaded with a prior FastestVictory (global and per-enemy) records a slower
+            // victory against the same enemy: the existing rows' values are unchanged, so they must stay
+            // out of both the persist set and the touched keys — the Min/Max analogue of the zero-delta Sum
+            // skip (#1821).
+            var enemy = MakeEnemy(id: 1);
+            var progress = MakeProgress(statistics:
+            [
+                Stat(EStatisticType.FastestVictory, null, 3m),
+                Stat(EStatisticType.FastestVictory, 1, 3m),
+            ]);
+
+            var touched = progress.RecordBattleCompleted(enemy, victory: true, playerDied: false,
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: dirty-mark Max/Min statistics only on a genuine value change (#1821)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change is exactly what the issue prescribes, is correct, and is well covered by tests.

The fix moves `_dirtyStatistics.Add` out of `GetOrCreate` and into the mutators, so an unimproved `SetMax`/`SetMin` no longer re-enqueues an unchanged row for the write-behind persist nor re-enters the touched set that scopes challenge evaluation (`ChallengeIndex.RelevantTo`). This makes `RecordBattleCompleted`'s own contract and `docs/backend.md`'s challenge-eval scoping claim accurate rather than aspirational.

Verified:
- `Increment` marks dirty unconditionally after applying the delta, which is sound because the zero-delta early-return means it only reaches that point when it actually mutates — no over-marking reintroduced.
- `SetMax`/`SetMin` mark dirty only inside the `created || improved` branch, so a newly-created row is always persisted (first-write-wins) while an unimproved existing row stays clean and untouched. `GetOrCreate` no longer creates phantom dirty entries.
- No frontend mirror needed: the statistics pass runs server-side at battle completion, off the deterministic tick loop, so it's outside the battle-parity surface (`docs/backend.md`).

### Test coverage

Solid. New tests cover unimproved-Max, unimproved-Min (global + per-enemy), and improved-Max on an existing row; the existing zero-first-write min/max and zero-delta-Sum tests still guard the create-and-dirty paths. The `Increment` relocation is covered by the pre-existing Sum-stat persistence/touched-set tests.

### What's good

Tightly scoped fix that resolves the issue at its root, keeps the mutators symmetric with the existing zero-delta `Increment` skip, and updates the in-code comments to match the corrected behavior. No follow-ups needed.


---

Issue #1962, opened 2026-07-14
EntityStore.lastIdMap still resolves new ids positionally, unlike the child-saver path


ginderjeremiah:
## Background

#1856 fixed the Workbench "positional new-id resolution" bug for the path that actually writes data: `persistEntity`'s internal `resolveNewIds` call (which drives child-collection saves like skill effects, item mod slots, etc.) now disambiguates newly-added records by identity content (via each entity's `toPrimaryDto`) instead of purely by sorted-id position, so a concurrent add from another admin landing in the same refetch can no longer steal this session's child writes. The same fix was applied to `progression-store.svelte.ts`'s two direct `resolveNewIds` calls (paths, proficiencies).

## Remaining gap

`EntityStore.save()` (`UI/src/routes/admin/workbench/entity-store.svelte.ts`) makes its own, separate `resolveNewIds(fresh, diff.existingIds, diff.added)` call — with no `identityKey` — purely to populate `lastIdMap`, which lets the Workbench detail pane keep the current selection pinned on a record across the save that just created it. This call still uses pure positional pairing, so under the same concurrent-add race, the UI could briefly follow the selection to the wrong record after a save (a cosmetic/UX issue — no data gets corrupted, since the actual child-saver writes go through the now-fixed `persistEntity` path independently).

## Why it wasn't folded into #1856

`EntityStore.save()` doesn't have access to the entity's `toPrimaryDto` — it's private to each entity's `persist()` closure inside `save-helpers.ts`. Giving `EntityStore` a correct identity key would mean changing the `EntityConfig.persist` return contract (e.g. having `persistEntity`/`config.persist` return the already-computed id map alongside the fresh records) so `EntityStore` can reuse the same resolution instead of recomputing it. That touches the `EntityConfig<T>` interface, all 10 entity configs' inferred return types, and the `entity-store.test.ts` mocks — a reasonable but separate, contained change, out of scope for the #1856 bug fix itself.

## Suggested fix

Change `EntityConfig.persist(diff): Promise<T[]>` to return `Promise<{ records: T[]; idMap: Map<number, number> }>` (or similar), have `persistEntity` return its already-correct `idFor` map alongside `fresh`, and have `EntityStore.save()` consume that returned map directly instead of calling `resolveNewIds` a second time. This also removes a duplicated computation (DRY), not just fixes the UX edge case.

Closes no bug on its own until fixed — purely a UX polish/tech-debt item.

PR #1963, opened 2026-07-14
Disambiguate Workbench new-id resolution by identity content, not position

### LINKED ISSUE #1856: Workbench positional new-id resolution can write child collections onto another admin's record under concurrent sessions

ginderjeremiah:
## Summary

`resolveNewIds` (`UI/src/routes/admin/workbench/save-helpers.ts`) mapped a just-added record's local (negative) id to its persisted id purely by sorted-id position: the k-th added record was assumed to be the k-th lowest id absent from the pre-save id set. That assumption only holds if *this session's* adds are the only new ids in the refetch — a concurrent add from another admin breaks it, and if the other admin's record sorts to a lower id, this session's child savers (skill effects, item mod slots, proficiency modifiers/rewards, etc.) get written onto **the wrong record**.

This fixes it by having `resolveNewIds` optionally match an added record to its persisted counterpart by **identity content** (everything an `identityKey` callback returns, minus `id`) before ever falling back to positional pairing — and even then, only among whatever content matching couldn't disambiguate.

## Changes

- **`save-helpers.ts` — `resolveNewIds`:** added an optional `identityKey` parameter. When supplied, each added record is matched to the fresh record with an equal identity key; positional pairing is now only a fallback for leftovers content matching can't tell apart (e.g. two of *this session's own* adds with identical, not-yet-diverged identity fields — harmless, since they're indistinguishable by content either way).
- **`save-helpers.ts` — `persistEntity`:** now passes its existing `toPrimaryDto` as the `identityKey`, so every one of the 10 generic-config admin entities (skills, items, item mods, enemies, zones, tags, challenges, classes, skill recipes, lessons) gets the fix for free.
- **`progression-store.svelte.ts`:** the two direct `resolveNewIds` calls (for paths and proficiencies, which don't go through the generic `persistEntity` pipeline) now pass `pathIdentityDto`/`profIdentityDto` respectively.
- Added unit tests covering the content-matching path, the positional-fallback-for-ties path, the genuinely-unresolvable case, and a `persistEntity`-level test that reproduces the exact failure scenario from #1856 (a concurrent admin's add landing at a lower id) and asserts the child saver now targets the correct record.

## Scope note (from the prior investigation on this issue)

A previous pass on #1856 found the *fully* correct fix — having the Add endpoints return persisted ids in send order — requires restructuring when the admin write endpoints commit (currently post-response, in `CommitFilter`, after EF assigns ids), touching ~12 backend admin repos/controllers and ~11 frontend files. That's a real architectural decision needing sign-off, not a mechanical fix, so this PR implements the narrower interim mitigation that comment proposed instead: content-based disambiguation. It closes the reported failure scenario for realistic cases; it does not fully close the race for two *concurrently added* records that happen to share byte-identical identity fields (an edge case explicitly called out as an accepted tradeoff of this approach).

That same investigation also found the issue's claimed `?? record.id` fallback in `resolveId` doesn't exist in current code (`resolveId` throws on an unresolved negative id) — likely already fixed or describing code elsewhere; noting it for the record rather than acting on it further.

## Follow-up filed

`EntityStore.save()` (`entity-store.svelte.ts`) makes its own separate, positional-only `resolveNewIds` call to populate `lastIdMap` (Workbench selection-tracking across a save, not data writes) — it doesn't have access to the entity's `toPrimaryDto`, so fixing it needs a small `EntityConfig.persist` contract change. Filed as #1962, scoped small, since it's a pure UX/tech-debt item (no data corruption) separate from this bug fix.

Closes #1856

## Test plan

- [x] `npx vitest run` — all 3666 unit tests pass (297 files)
- [x] New/updated tests specifically exercise: content-based matching, positional fallback among ties, genuinely-unresolvable adds, and the exact concurrent-admin-add scenario from the bug report

https://claude.ai/code/session_015SAESKvLdWhbJj61aoPSAm

---
_Generated by [Claude Code](https://claude.ai/code/session_015SAESKvLdWhbJj61aoPSAm)_

### DIFF
--- a/UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
+++ b/UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
@@ -489,11 +489,14 @@ export class ProgressionStore {
 			}
 
 			// 4. Resolve the persisted ids of newly-added paths before the proficiencies that FK to them.
+			// Identity-content matching (not just position) keeps this correct when another admin's
+			// concurrent add lands in the same refetch (#1856).
 			const freshPaths = await fetchSocketData('GetPaths');
 			const pathIdMap = resolveNewIds(
 				freshPaths,
 				this.basePaths.map((p) => p.id),
-				pathDiff.added
+				pathDiff.added,
+				pathIdentityDto
 			);
 
 			// 5. Proficiency identities — remap a (possibly brand-new) path id into each DTO.
@@ -513,11 +516,13 @@ export class ProgressionStore {
 			}
 
 			// 6. Resolve the persisted ids of newly-added proficiencies (for child savers + gateways).
+			// Identity-content matching keeps this correct under the same concurrent-add race (#1856).
 			const freshProfs = await fetchSocketData('GetProficiencies');
 			const profIdMap = resolveNewIds(
 				freshProfs,
 				this.baseProfs.map((p) => p.id),
-				profDiff.added
+				profDiff.added,
+				profIdentityDto
 			);
 
 			// 7. Proficiency child collections — modifiers and rewards per tier, minus the ones already
--- a/UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
+++ b/UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
@@ -517,12 +517,16 @@ export class ProgressionStore {
 
 			// 6. Resolve the persisted ids of newly-added proficiencies (for child savers + gateways).
 			// Identity-content matching keeps this correct under the same concurrent-add race (#1856).
+			// Uses toProfDto (not the raw profIdentityDto) so a proficiency added under a path that's
+			// also new in this save keys on the path's *resolved* id on both sides — profDiff.added's
+			// record still carries the path's local negative id, which would otherwise never match the
+			// server's real pathId and silently fall back to positional pairing.
 			const freshProfs = await fetchSocketData('GetProficiencies');
 			const profIdMap = resolveNewIds(
 				freshProfs,
 				this.baseProfs.map((p) => p.id),
 				profDiff.added,
-				profIdentityDto
+				toProfDto
 			);
 
 			// 7. Proficiency child collections — modifiers and rewards per tier, minus the ones already
--- a/UI/src/routes/admin/workbench/save-helpers.ts
+++ b/UI/src/routes/admin/workbench/save-helpers.ts
@@ -38,19 +38,48 @@ export const childChanged = <T>(current: T, baseline: T | undefined): boolean =>
 	baseline === undefined || !listsEqual(current, baseline);
 
 /**
- * Map the local (negative) ids of added records to their persisted ids. The backend appends adds in
- * send order, so the k-th added record maps to the k-th lowest id absent from the pre-save set.
+ * Map the local (negative) ids of added records to their persisted ids. Purely positional pairing (the
+ * k-th added record maps to the k-th lowest id absent from the pre-save set) assumes only this session's
+ * own adds appear as new ids in the refetch — a concurrent add from another admin breaks that assumption
+ * and can pair a child saver against the wrong record (#1856). When `identityKey` is supplied, an added
+ * record is instead matched to the fresh record whose identity key (everything the key builder returns,
+ * excluding `id`) is equal; positional pairing is used only as a fallback, and only among whatever's left
+ * once identity matching is exhausted — e.g. several adds that still share identical, undiverged identity
+ * fields, which is unambiguous either way since nothing distinguishes them by content.
  */
-export const resolveNewIds = (
-	fresh: { id: number }[],
+export const resolveNewIds = <T extends { id: number }>(
+	fresh: T[],
 	existingIds: Iterable<number>,
-	added: { id: number }[]
+	added: T[],
+	identityKey?: (record: T) => unknown
 ): Map<number, number> => {
 	const existing = new Set(existingIds);
 	const newlyPersisted = fresh.filter((record) => !existing.has(record.id)).sort((a, b) => a.id - b.id);
 	const idFor = new Map<number, number>();
-	added.forEach((record, index) => {
-		const persisted = newlyPersisted[index];
+	const remaining = [...newlyPersisted];
+	let unmatched = added;
+
+	if (identityKey) {
+		const keyOf = (record: T) => {
+			const key = { ...(identityKey(record) as Record<string, unknown>) };
+			delete key.id;
+			return JSON.stringify(canonicalize(key));
+		};
+		const stillUnmatched: T[] = [];
+		for (const record of added) {
+			const matchIndex = remaining.findIndex((candidate) => keyOf(candidate) === keyOf(record));
+			if (matchIndex === -1) {
+				stillUnmatched.push(record);
+			} else {
+				idFor.set(record.id, remaining[matchIndex].id);
+				remaining.splice(matchIndex, 1);
+			}
+		}
+		unmatched = stillUnmatched;
+	}
+
+	unmatched.forEach((record, index) => {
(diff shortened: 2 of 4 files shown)


UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
```diff
@@ -513,11 +516,13 @@ export class ProgressionStore {
...
 
 			// 6. Resolve the persisted ids of newly-added proficiencies (for child savers + gateways).
+			// Identity-content matching keeps this correct under the same concurrent-add race (#1856).
 			const freshProfs = await fetchSocketData('GetProficiencies');
 			const profIdMap = resolveNewIds(
 				freshProfs,
 				this.baseProfs.map((p) => p.id),
-				profDiff.added
+				profDiff.added,
+				profIdentityDto
```
ginderjeremiah: This passes `profIdentityDto`, but `profIdentityDto` includes `pathId: prof.pathId` (progression-helpers.ts:212). For a proficiency added under a path that was **also** newly added in this same save, `profDiff.added`'s record still carries the path's local **negative** id (it's never mutated — step 5's `toProfDto` builds new objects for the POST), while `freshProfs` from the server carry the real persisted `pathId`. So `keyOf(added)` and `keyOf(fresh)` never match on `pathId`, that proficiency fails identity matching, and it drops back to the exact positional pairing this PR is trying to eliminate — reopening the #1856 corruption window for the common "create a new path plus its proficiencies in one save" workflow.

The remapped `toProfDto` (defined just above at line 503) is already in scope and resolves `pathId` through `pathIdMap`, so feeding it here instead closes the gap: for the fresh (server) records `resolveId(realPathId, pathIdMap)` passes the positive id through unchanged, and for the added records it maps the negative path id to the same real id — so both sides key on the real `pathId` and match. Suggest `resolveNewIds(freshProfs, this.baseProfs.map((p) => p.id), profDiff.added, toProfDto)`, ideally with a small unit test covering a proficiency added under a same-save new path racing a concurrent add.

---

ginderjeremiah (commented):
## Review: Content-based new-id disambiguation in Workbench saves (#1856)

**Verdict:** Request changes — the core `resolveNewIds` change is sound, but the proficiency call site (progression-store.svelte.ts:525) feeds the *un-remapped* identity DTO, leaving the same corruption window open for a common authoring flow. It's a one-identifier fix and belongs in this PR rather than a follow-up.

_(Noting for the record: I'm the PR author, so GitHub only accepts this as a `COMMENT` — the intended verdict is Request changes.)_

The general approach is a good interim mitigation. Matching added records to their persisted counterparts by identity content before falling back to positional pairing is the right shape, the fallback preserves the old behavior when no `identityKey` is passed (so the untouched `entity-store.svelte.ts` caller — #1962 — still compiles and behaves as before), and routing every generic entity's fix through `persistEntity`'s existing `toPrimaryDto` is clean. The scope note and the honest call-out of the residual identical-content edge case are appreciated.

### Should-fix (recommend before merge)

- **`progression-store.svelte.ts:525` — proficiency identity match uses the un-remapped `pathId`.** `profIdentityDto` carries `pathId: prof.pathId`. When a proficiency is added under a path that's *also* new in the same save, `profDiff.added`'s record still holds the path's local negative id while the server's `freshProfs` hold the real `pathId`, so content matching never lands and that proficiency silently falls back to positional pairing — the exact race #1856 targets, for the common "new path + its proficiencies in one save" workflow. Fix: pass the already-in-scope `toProfDto` (line 503) instead of `profIdentityDto` — it remaps `pathId` through `pathIdMap`, so both sides key on the real id and match (a real server `pathId` passes through `resolveId` unchanged). See the inline comment. The path call site (line 499) is correct as-is, since path identity has no remapped FK.

### Nits

- **`save-helpers.ts:63-70` — `keyOf` recompute.** `keyOf(record)` is re-evaluated for every candidate inside `findIndex`, and each call re-`JSON.stringify`s a freshly canonicalized object, so matching is O(n²) with a stringify per comparison. Negligible at admin-tool cardinalities and not worth blocking on, but precomputing each record's key once (a `Map`/array of keys) would make it clearly linear and easier to read.

### Test coverage

Good coverage on the `resolveNewIds`/`persistEntity` unit surface: content-match, positional-fallback-among-ties, genuinely-unresolvable, and a `persistEntity`-level reproduction of the concurrent-lower-id scenario from #1856. The gap is the progression-store proficiency path — there's no test exercising a proficiency added under a same-save new path, which is precisely the case the should-fix above concerns. If that's fixed, add a case covering it (proficiency under a brand-new path, racing a concurrent add that sorts lower).

### What's good

- Backward-compatible signature: `identityKey` is optional and the no-key path is byte-for-byte the old behavior.
- Reusing `canonicalize` for the key so `null`-vs-absent and key order don't create phantom mismatches, and stripping `id` before comparing (local negatives vs persisted ids) — both necessary and correctly handled.
- Clear, accurate doc comment on `resolveNewIds` and a candid scope/limitations write-up in the PR body.

---

UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
ginderjeremiah: Confirmed and fixed in a0bb238 — swapped `profIdentityDto` for `toProfDto` at this call site, verified the failure by temporarily reverting and re-running the new test (it fails exactly as described), then restored the fix. Also added an end-to-end regression test in `progression-store.test.ts` that reproduces a concurrent admin's add landing at a lower id under a same-save new path, using the existing fake-server test harness. Full suite (3667 tests) + lint/type-check are green.

---

ginderjeremiah:
## Re-review — Approve

Picked this back up. The earlier review's one should-fix (the proficiency call site feeding the un-remapped `pathId`, `progression-store.svelte.ts:525`) was addressed in `a0bb238` — it now passes the remapped `toProfDto`, so both sides of the identity comparison key on the resolved `pathId`, and there's an end-to-end regression test reproducing a concurrent add landing at a lower id under a same-save new path. That thread is resolved. I re-read the fix and it's correct.

The remaining open item was the `keyOf` O(n·m) recompute nit — `keyOf` re-canonicalized and re-`JSON.stringify`'d both operands inside `findIndex`, doing a stringify per comparison. I pushed `901d23a` precomputing each candidate's key once (`remainingKeys`) so matching compares cached strings; the splice keeps the parallel arrays aligned. Behavior is unchanged — `save-helpers.test.ts` + `progression-store.test.ts` pass locally (53/53).

The content-based disambiguation approach is a sound interim mitigation for #1856, backward-compatible (the no-`identityKey` path is byte-for-byte the old behavior), and the residual identical-content edge case and the deferred full fix (#1962, endpoints returning ids in send order) are honestly scoped. Merging once CI is green.

---

PR #1964, opened 2026-07-14
Make Item/ItemMod classes so InventoryManager mutations reach the UI (#1957)

### LINKED ISSUE #1957: InventoryManager mutates raw Map items while the UI reads $state proxies — favorite/equip/mod changes never render
`InventoryManager.unlockedItems` is a plain `Map<number, Item>` (`inventory-manager.ts:48`), which neither `statify` nor Svelte's `$state` proxies (`statify.svelte.ts:82-101` only proxies plain-object/array fields, per its own doc note). Every mutation path that resolves an item via `this.unlockedItems.get(itemId)` therefore writes to the **raw** item object, while the inventory grid reads the **same item through the proxied `items` array**. Svelte 5's `$state` proxy never writes through to the underlying target and never re-reads it once a property source is materialized — so these mutations are permanently invisible to the UI.

**Raw-write sites:** `setFavorite` (`inventory-manager.ts:341`), `applyEquip` (`:427-429`), `unequipItem` (`:239-240`), `applyMod`/`removeMod` (`:277`/`:305`), `refreshItemAttributes` (`:436-439`). The doc comments at `:56-62`/`:449-455` claim the opposite (that consumers observe these mutations reactively) and are wrong.

**Verified empirically** (vitest, real `InventoryManager` + real `statify`, only socket/staticData mocked): a `$derived` over `manager.items[0]?.favorite` stays `false` after `setFavorite(1, true)` resolves even though `unlockedItems.get(1)?.favorite === true`; same for `equipped` after `equipItem`.

**User-visible symptoms** (traced through the actual component paths):
- The grid favorite star (`GridSlot.svelte:48-59`) never toggles on click; clicking again re-sends `favorite: true` because the view re-reads the stale proxy (`inventory-view.svelte.ts:237-241`). It only corrects after a `publish()` (new unlock) or re-login.
- The grid tile's equipped marker/tint (`GridSlot.svelte:26,65-67`) goes stale after equip/unequip (the equipped **rail** stays correct only because `equippedSlots` is wholesale-reassigned).
- The item drawer receives `view.selected` — resolved off the raw, untracked Map (`inventory-view.svelte.ts:81-87`) — so applying/removing a mod shows nothing until the drawer is closed and reopened (`ModSlots.svelte:6`; `ItemDrawer.svelte:75`'s "it recomputes as mods are applied/removed" comment is false), and the drawer's Equip/Unequip label (`ItemDrawer.svelte:61`) doesn't flip after the action.
- Battle correctness is **unaffected**: the engine reads the reassigned `equipmentStatsCache`, and rollback closures restore the same raw objects they snapshotted.

**Why tests never caught it:** `Inventory.test.ts` mocks the whole manager, `GridSlot.test.ts` passes items as plain props, and `inventory-manager.test.ts` asserts only raw `unlockedItems.get(…)` state — no test spans manager mutation → proxied UI read.

**Suggested fix:** give the item objects one shared reactive identity across both access paths — e.g. statify/`$state`-proxy items on insertion (or store the published proxies in the map and resolve mutations through them). Note `PlayerManager`/`Battler` don't have this problem because they always read+write through their own proxied public fields — the raw-Map mixing is unique to `InventoryManager`. Whatever the shape: add a regression test asserting a `$derived` over `manager.items[…].favorite`/`equipped`/`appliedMods` observes each mutation, and fix the misleading doc comments on `items`/`publish()` and `ItemDrawer.attributeMap`.

ginderjeremiah:
## Summary

`InventoryManager.unlockedItems` (a `Map`) never rendered a mutation without a reload — favorite toggles, equip/unequip, and mod apply/remove all silently failed to update the grid, the equipped rail, and the item drawer until the page was refreshed.

**Root cause:** `Item` and `ItemMod` (`UI/src/lib/battle/item.ts`, `item-mod.ts`) were plain object literals, not classes. `statify`'s `isClass()` check only recognizes real class instances — a plain object falls through to Svelte's own native `$state` deep-proxy instead. That proxy wraps a value differently *per container that reads it*: the `items`/`equippedSlots` arrays each produced their own proxied copy of an item, while `unlockedItems` (a `Map`, which Svelte's `$state` never deep-proxies) kept the raw, unproxied object. Every mutation path (`setFavorite`, `equipItem`, `applyMod`, …) resolves its target via `unlockedItems.get(id)` and writes to that raw object — a different identity than whatever the UI actually reads through `items`/`equippedSlots`, so the write was invisible.

Verified empirically with a `$effect.root`/`$derived` harness against the real `InventoryManager` + `statify` (no mocks) before writing the fix — confirmed the divergence, then confirmed the fix closes it.

**Fix:** convert `Item`/`ItemMod` to real classes, matching the existing `Battler`/`Skill` convention (called out in the issue as already avoiding this pitfall). `statify` then defines each item's field accessors once, directly on the object, so every container that references it — the Map, `items`, `equippedSlots` — shares one identity and one set of reactive fields.

## Changes

- **`item.ts` / `item-mod.ts`**: `Item`/`ItemMod` are now classes constructed via `newItem`/`newItemMod`, preserving their existing public shape and call sites.
- **`inventory-manager.ts`**: expanded the `items` field's doc comment to explain the identity invariant the fix relies on (`Item`/`ItemMod` must stay classes), so a future regression is documented rather than silent.
- **New test** `inventory-manager-reactivity.svelte.test.ts`: a `statify`-wrapped `InventoryManager` exercised through a real `$derived`, asserting `favorite`, `equipped`, and `appliedMods` are observed live through `manager.items` after `setFavorite`/`equipItem`/`unequipItem`/`applyMod`/`removeMod` — this is the regression coverage the issue asked for (it fails on the pre-fix code, confirmed while developing the fix).

## Test plan

- [x] `npx vitest run` — 3665/3665 unit tests pass (including the 3 new regression tests)
- [x] `npm run build` — succeeds
- [x] **Live browser verification**: ran the API + Postgres/Redis + Vite dev server locally, created a real player account, granted a test item via direct SQL, and drove the Inventory screen with Playwright — clicking the favorite star and Ctrl+clicking to equip both updated the grid **immediately, with no page reload** (screenshots attached in the issue/PR conversation). This is the exact previously-broken flow.

Closes #1957

https://claude.ai/code/session_016Fmw9koT61f6NuqBu7SQBV

---
_Generated by [Claude Code](https://claude.ai/code/session_016Fmw9koT61f6NuqBu7SQBV)_

### DIFF
--- a/UI/src/lib/battle/item-mod.ts
+++ b/UI/src/lib/battle/item-mod.ts
@@ -1,7 +1,35 @@
-import { IAppliedModModel, IItemMod } from '$lib/api';
+import { EItemModType, ERarity, IAppliedModModel, IBattlerAttribute, IItemMod } from '$lib/api';
 import { staticData } from '$stores';
 
-export interface ItemMod extends Omit<IAppliedModModel, 'itemModId'>, IItemMod {}
+// A class (not a plain object literal) so `statify` recognizes it via its own per-field reactivity
+// (see `statify.svelte.ts`) instead of falling through to Svelte's native deep-proxy, which would give
+// the same underlying mod a different identity depending on which reactive array reads it (#1957).
+export class ItemMod implements IItemMod {
+	id: number;
+	name: string;
+	description: string;
+	itemModTypeId: EItemModType;
+	rarityId: ERarity;
+	attributes: IBattlerAttribute[];
+	tags: number[];
+	designerNotes: string;
+	retiredAt?: string;
+
+	itemModSlotId: number;
+
+	constructor(itemModData: IItemMod, itemModSlotId: number) {
+		this.id = itemModData.id;
+		this.name = itemModData.name;
+		this.description = itemModData.description;
+		this.itemModTypeId = itemModData.itemModTypeId;
+		this.rarityId = itemModData.rarityId;
+		this.attributes = itemModData.attributes;
+		this.tags = itemModData.tags;
+		this.designerNotes = itemModData.designerNotes;
+		this.retiredAt = itemModData.retiredAt;
+		this.itemModSlotId = itemModSlotId;
+	}
+}
 
 export const newItemMod = (appliedMod: IAppliedModModel): ItemMod | undefined => {
 	// Resolve the mod's static definition by id; a missing/retired id yields no record, so degrade
@@ -10,8 +38,5 @@ export const newItemMod = (appliedMod: IAppliedModModel): ItemMod | undefined =>
 	if (!itemModData) {
 		return undefined;
 	}
-	return {
-		...itemModData,
-		itemModSlotId: appliedMod.itemModSlotId
-	} satisfies ItemMod;
+	return new ItemMod(itemModData, appliedMod.itemModSlotId);
 };
--- a/UI/src/lib/battle/item.ts
+++ b/UI/src/lib/battle/item.ts
@@ -1,15 +1,67 @@
-import { IItem, IInventoryItem } from '$lib/api';
+import { EDamageType, EItemCategory, ERarity, IBattlerAttribute, IInventoryItem, IItem, IItemModSlot } from '$lib/api';
 import { ItemMod, newItemMod } from './item-mod';
 import { BattleAttributes } from './battle-attributes';
 import { staticData } from '$stores';
 
-export interface Item extends IItem {
+// A class (not a plain object literal) so `statify` recognizes it via its own per-field reactivity
+// (see `statify.svelte.ts`) instead of falling through to Svelte's native deep-proxy, which would give
+// the same underlying item a different identity depending on which reactive array reads it. That
+// divergence is what let `InventoryManager.unlockedItems` (a Map, never itself deep-proxied) hold a
+// raw item while the UI read a separately-proxied copy of the same object — mutations through the Map
+// were invisible to the UI (#1957). Keeping `Item`/`ItemMod` as classes preserves one shared identity
+// across every container (the Map, `items`, `equippedSlots`) that references the same item.
+export class Item implements IItem {
+	id: number;
+	name: string;
+	description: string;
+	itemCategoryId: EItemCategory;
+	rarityId: ERarity;
+	iconPath: string;
+	attributes: IBattlerAttribute[];
+	modSlots: IItemModSlot[];
+	tags: number[];
+	grantedSkillId?: number;
+	weaponType?: EDamageType;
+	requiredProficiencyId?: number;
+	requiredProficiencyLevel: number;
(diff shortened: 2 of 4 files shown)


ginderjeremiah:
## Review: Make Item/ItemMod classes so InventoryManager mutations reach the UI (#1957)

**Verdict: Approve** — the fix is correct, complete, and convention-aligned. Verified against the actual `statify`, the generated contracts, and every call site.

### Root cause & fix — verified
`statify`'s `isClass()` rejects anything whose `constructor.name === 'Object'`, so plain-object `Item`/`ItemMod` fell through to Svelte's native `$state` deep-proxy — which yields a *different proxied identity per container*, while `unlockedItems` (a `Map`) is never deep-proxied and keeps the raw object. Every mutation path (`setFavorite`, `applyEquip`, `unequipItem`, `applyMod`, `removeMod`, `refreshItemAttributes`) resolves its target via `unlockedItems.get(id)` and writes that raw object — a different identity than the proxied `items[]`/`equippedSlots` the UI reads, so writes were invisible without a reload. Converting to real classes makes `statify` define field accessors once directly on the object (idempotent via `__statifyPerformed`), so the Map, `items[]`, and `equippedSlots` all share one reactive identity. This genuinely closes the divergence.

### Completeness — verified
- **Public shape preserved:** the class field lists exactly match `IItem` (15 fields) / `IItemMod` (9 fields) plus the existing inventory extras; `implements` under `strict` + `strictPropertyInitialization` makes the compiler enforce no field is dropped.
- **No missed call sites:** `newItem`/`newItemMod` in `$lib/battle` are the only constructors of these types and both route through the new classes (the other `newItem` hits are the unrelated admin-Workbench factory).
- **Test is meaningful:** `inventory-manager-reactivity.svelte.test.ts` wraps a real `InventoryManager` in real `statify` and observes `favorite`/`equipped`/`appliedMods` through an actual `$derived` in `$effect.root` — the exact Map-mutation → array-read boundary that was broken, and it fails on pre-fix code. This is the regression coverage the issue asked for.

### Follow-up filed
- **#1967** (tech debt, scope: small) — the `totalAttributes` recompute is duplicated between the new `Item` constructor (`item.ts:60-61`) and `InventoryManager.refreshItemAttributes` (`inventory-manager.ts:443-444`). Pre-existing, but now that `Item` is a class it should be folded into an instance method so the two can't drift. Left as a follow-up rather than blocking since it's out of this fix's scope.

The verbose field-by-field constructor copy vs. `Object.assign` is pure preference (the explicit form is compiler-guarded), and coverage of the core identity invariant is adequate — neither is blocking.

CI is green across frontend unit, verify, and all three e2e browsers. Merging.

---

PR #1965, opened 2026-07-14
Guard zero-based FK fixup against re-pointing propagated temp FKs at record 0 (#1824)

### LINKED ISSUE #1824: Zero-based FK fixup would silently re-point propagated temp FKs at record 0 — add a defensive guard
## Problem

`ApplyZeroBasedIdentityFixups`'s FK branch (`Game.Infrastructure/Database/GameContext.cs:956-974`) forces *any* temporary int FK referencing a zero-based principal to literal `0`, without distinguishing "EF read an intended 0 as unset" (the case it exists for) from "EF propagated a temporary key from an `Added` principal in the same save" (where 0 is wrong).

## Failure scenario (latent)

Add a new `Item` (Added, temp key) with `ItemAttribute` rows attached via navigation in one `SaveChangesAsync`: relationship fixup marks the children's `ItemId` temporary; `ForceZero` clears the flag and writes `0`; the children insert against item 0 — which exists, so the FK passes and the corruption is **silent**. No current path does this (admin repos build navigation-free entities against existing parents; the content seeder bypasses the change tracker via `EntityRowInserter`) — but the failure mode is silent data corruption and the guard is cheap. The PK branch already documents its own defensive posture (#1003); the FK branch has no equivalent guard or comment.

## Suggested fix

Skip `ForceZero` (or throw) when the FK's principal entry is tracked `Added` with a temporary key, plus a short comment mirroring the PK branch's. A unit/integration test creating a principal + children via navigation in one save pins it.

ginderjeremiah:
## Summary

`GameContext.ApplyZeroBasedIdentityFixups`'s FK branch forced *any* temporary int FK referencing a zero-based-identity principal (`Item`, `Enemy`, `Zone`, …) to the literal `0`, without distinguishing:

- **"EF read an intended 0 as unset"** — the case the fixup exists for (a genuine FK to record 0 gets misread as an unset store-generated value and needs forcing back to the literal `0`).
- **"EF's relationship fixup propagated a temporary key from an `Added` principal in the same save"** — e.g. adding a new `Item` with `ItemAttribute` rows attached via navigation in one `SaveChangesAsync`. Here, forcing the child's FK to `0` silently re-points the insert at the *existing* record 0 instead of letting EF resolve it to the *new* principal's real generated id once the insert commits — a silent data-corruption failure mode.

No current production path triggers this (admin repos build navigation-free entities against existing parents; the content seeder bypasses the change tracker), so the bug was latent — but the fix is cheap and closes the gap before any future navigation-based add hits it.

## Changes

- **`GameContext.cs`**: `ApplyZeroBasedIdentityFixups` now does a pre-pass over the change tracker to collect `(EntityType, TempKeyValue)` for every zero-based-identity principal that is `Added` with a still-temporary key in this save. The FK branch then skips `ForceZero` for any FK whose temporary value matches one of those — leaving it for EF's own key-fixup (which runs during the real insert) to resolve to the principal's real generated id — and still forces to `0` in every other case, exactly as before.
- `ZeroBasedFixup.ForeignKeyProperties` now carries a `ForeignKeyFixup(PropertyName, PrincipalEntityType)` per FK instead of a bare property name, so the guard can identify which principal type/entry each FK targets.
- **Tests**: updated the existing `BuildZeroBasedFixups` assertions for the new `ForeignKeyFixup` shape, and added a new regression test (`ApplyZeroBasedIdentityFixups_ForNavigationAddedChildOfANewPrincipal_LeavesThePropagatedForeignKeyTemporary`) that adds a new `Enemy` with an `EnemySkill` attached via navigation — mirroring the issue's exact failure scenario — and asserts the propagated `EnemyId` FK is left temporary while an unrelated `SkillId` FK (still genuinely referencing record 0) is correctly forced to `0`.

## Test plan

- [x] `dotnet build` — solution builds clean, 0 warnings
- [x] `dotnet test Game.Infrastructure.Tests` — all 64 tests pass, including the new regression test
- [x] `dotnet test` (full solution, against the session's Postgres/Redis containers) — 3137 tests pass (`Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Application.Tests`, `Game.Api.Tests`)

Closes #1824

https://claude.ai/code/session_012LFPere1ttNWzQqMqrrW8Q

---
_Generated by [Claude Code](https://claude.ai/code/session_012LFPere1ttNWzQqMqrrW8Q)_

### DIFF
--- a/Game.Infrastructure.Tests/GameContextZeroBasedIdentityFixupTests.cs
+++ b/Game.Infrastructure.Tests/GameContextZeroBasedIdentityFixupTests.cs
@@ -56,7 +56,7 @@ namespace Game.Infrastructure.Tests
             Assert.Null(enemySkillFixup.KeyProperty);
             Assert.Equal(
                 [nameof(EnemySkill.EnemyId), nameof(EnemySkill.SkillId)],
-                enemySkillFixup.ForeignKeyProperties.OrderBy(p => p));
+                enemySkillFixup.ForeignKeyProperties.Select(fk => fk.PropertyName).OrderBy(p => p));
         }
 
         [Fact]
@@ -74,7 +74,7 @@ namespace Game.Infrastructure.Tests
             Assert.True(fixups.TryGetValue(playerType, out var playerFixup));
             Assert.Equal(
                 [nameof(Player.ClassId), nameof(Player.CurrentZoneId)],
-                playerFixup.ForeignKeyProperties.OrderBy(p => p));
+                playerFixup.ForeignKeyProperties.Select(fk => fk.PropertyName).OrderBy(p => p));
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Guard the zero-based FK fixup against forcing propagated same-save temp FKs to record 0 (#1824)

**Verdict:** Approve *(submitted as a comment because I'm the PR author; no gating event is possible on one's own PR)* — the change closes exactly the latent gap #1824 describes, is well-targeted, and is pinned by a test that reproduces the failure scenario.

I traced the logic against `GameContext.ApplyZeroBasedIdentityFixups` in full and it holds up:

- The pre-pass collects `(PrincipalEntityType, TempKeyValue)` only for `Added` zero-based-identity principals whose key is still temporary — precisely the entries whose real id isn't assigned until insert. The FK branch then skips `ForceZero` only when the FK is temporary **and** its temp value matches one of those principals, leaving EF's own relationship fixup to resolve it. Every other case still force-zeros exactly as before, so there's no behavior change for the genuine "EF read an intended 0 as unset" path.
- Matching by `(principal type, temp value)` identity is the right key: two same-type `Added` principals get distinct temp keys, so a child is never mis-matched to the wrong sibling, and a genuine record-0 FK (a separate temp draw) never false-matches an `Added` principal — the guard only fires on an actual propagation.
- A second `ChangeTracker.Entries()` pass is genuinely necessary (a dependent can be visited before its principal), and the hot-path cost is negligible — the write-behind save touches 1–2 entries and typically has no `Added` principals at all.

Placement (infrastructure/persistence) is correct — no domain logic leaks in. The behavior is consistent with the design already documented in `backend-admin.md`, and the `ForeignKeyFixup` record + comments make the two temp-FK cases legible for the next reader.

### Test coverage

Solid. `ApplyZeroBasedIdentityFixups_ForNavigationAddedChildOfANewPrincipal_LeavesThePropagatedForeignKeyTemporary` mirrors the issue's exact scenario (a new `Enemy` with a nav-attached `EnemySkill`) and asserts both halves that matter: the propagated `EnemyId` is left temporary and equal to the principal's temp key, while an unrelated `SkillId` (a genuine record-0 FK) is still forced to `0` — so it verifies the new skip *and* that the old force-zero still fires within the same entity. The updated `BuildZeroBasedFixups` assertions cover the new `ForeignKeyFixup` shape.

### What's good

- Fixes a real silent-data-corruption footgun before any navigation-based add can hit it, at low cost and low risk.
- Thorough, accurate comments and PR description; the failure mode and the reason for the guard are documented rather than left implicit.

No blocking findings, no should-fixes, no nits. Holding the merge only until `test-backend` finishes (it's still in progress; everything else is green or correctly skipped).

---

PR #1966, opened 2026-07-14
Record subscribed channel at subscribe time so UnSubscribe can't mismatch it (#1825)

### LINKED ISSUE #1825: RedisPubSubService.UnSubscribe trusts the caller-supplied channel — record the channel at subscribe time
## Problem

`RedisPubSubService.UnSubscribe(channel, id)` (`Game.Infrastructure/PubSub/Redis/RedisPubSubService.cs:161-172`) removes the id→handler registry entry and disposes the worker, but only unsubscribes the handler from the channel the **caller** names — the registry doesn't record which channel a handle was subscribed on. Pass a mismatched channel and the still-subscribed Redis handler will call `Start()` on a disposed worker (`ObjectDisposedException` on the StackExchange.Redis subscriber thread) — the exact window the method's own comment says its ordering exists to close.

Current callers pair channel/id correctly (`DataProviderSynchronizer`, `ReferenceCacheSynchronizer`, `SocketManagerService`), so this is hardening: the API's shape makes a whole misuse class possible in a fiddly lifecycle seam.

## Suggested fix

Store the channel in the registered tuple at subscribe time and drop the `channel` parameter from `UnSubscribe` (or assert it matches). Cheap, and removes the misuse class outright.

ginderjeremiah:
## Summary

`RedisPubSubService.UnSubscribe(channel, id)` removed the id→handler registry entry and disposed the worker, but unsubscribed the handler from whatever channel the **caller** passed in — the registry never recorded which channel a handle was actually subscribed on. A caller passing a mismatched channel would leave the real Redis subscription live while disposing the worker it routes to, so the next message on that channel calls `Start()` on a disposed worker — an `ObjectDisposedException` on the StackExchange.Redis subscriber thread, in exactly the window the method's own comment says its ordering exists to close.

Current callers (`DataProviderSynchronizer`, `ReferenceCacheSynchronizer`, `SocketManagerService`) always paired channel/id correctly, so this was hardening against a misuse class the API's shape made possible, not an active production bug.

## Changes

- **`RedisPubSubService`**: the handle registry now stores `(channel, handler, worker)` instead of `(handler, worker)`, recording the channel at `Subscribe`/`SubscribeWithWorker` time.
- **`IPubSubService.UnSubscribe`**: dropped the `channel` parameter entirely (now just `UnSubscribe(string id)`) — the channel is only ever the one recorded at subscribe time, removing the misuse class outright rather than just guarding against it.
- Updated all three production call sites (`DataProviderSynchronizer`, `ReferenceCacheSynchronizer`, `SocketManagerService`) and every test double/call site to match the new signature.
- Added a new integration test, `UnSubscribe_UnsubscribesTheChannelRecordedAtSubscribeTime`, that subscribes, confirms a published message is received, unsubscribes, and confirms a subsequent publish on the same channel no longer reaches the handler — pinning that the real Redis subscription is torn down using the internally-tracked channel.

## Test plan

- [x] `dotnet test` (full solution, using the session's Postgres/Redis containers) — 3,138/3,138 tests pass, including the new regression test
- [x] Verified no frontend references to `UnSubscribe` (this is a backend-only Redis pub/sub concern)

Closes #1825

---
_Generated by [Claude Code](https://claude.ai/code/session_01RoyQEvjJbsg7S3h1Ee88mc)_

### DIFF
--- a/Game.Abstractions/Infrastructure/IPubSubService.cs
+++ b/Game.Abstractions/Infrastructure/IPubSubService.cs
@@ -28,11 +28,15 @@
         public Task Subscribe(string channel, Action<(string message, string channel)> action, string? id = null);
 
         // The worker-backed overload requires a non-null id: it spins up a BackgroundWorker (holding an OS wait
-        // handle) that can only be disposed via UnSubscribe(channel, id), so an id-less worker subscription could
+        // handle) that can only be disposed via UnSubscribe(id), so an id-less worker subscription could
         // never be torn down and would leak its handle (#954). The plain-action overload above creates no worker,
         // so its id stays optional.
         public Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string id);
-        public Task UnSubscribe(string channel, string id);
+
+        // Takes only the id, not the channel: the channel a handle was subscribed on is recorded at subscribe
+        // time, so a caller can never unsubscribe against a mismatched channel and leave the real subscription
+        // routing to a disposed worker (#1825).
+        public Task UnSubscribe(string id);
 
         /// <summary>
         /// Returns a handle to the named queue without subscribing to any channel. Useful for writing to a
--- a/Game.Api.Tests/Integration/AdminCacheReloadBroadcastTests.cs
+++ b/Game.Api.Tests/Integration/AdminCacheReloadBroadcastTests.cs
@@ -70,7 +70,7 @@ namespace Game.Api.Tests.Integration
             }
             finally
             {
-                await pubsub.UnSubscribe(ReferenceDataChannel, subscriptionId);
+                await pubsub.UnSubscribe(subscriptionId);
             }
         }
     }
--- a/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
@@ -385,7 +385,7 @@ namespace Game.Api.Tests.Unit
             public Task Publish<T>(string channel, string queueName, T queueData, CancellationToken cancellationToken = default) => Task.CompletedTask;
             public Task PublishBatch<T>(string channel, string queueName, IEnumerable<T> queueData, CancellationToken cancellationToken = default) => Task.CompletedTask;
             public Task Wake(string channel) => Task.CompletedTask;
-            public Task UnSubscribe(string channel, string id) => Task.CompletedTask;
+            public Task UnSubscribe(string id) => Task.CompletedTask;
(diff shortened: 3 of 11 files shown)


ginderjeremiah (commented):
## Review: Record the subscribed channel at subscribe time so `UnSubscribe` can't mismatch it (#1825)

**Verdict:** Approve — clean, correctly-scoped implementation of the issue's suggested fix, with real end-to-end regression coverage. (Submitted as a `COMMENT` because GitHub won't accept an `APPROVE` on one's own PR; the intended verdict is **approve**.)

The change does exactly what #1825 asked for: it records the channel in the handle tuple at `Subscribe`/`SubscribeWithWorker` time and drops the `channel` parameter from `UnSubscribe`, so `UnsubscribeAsync` now tears down the channel the handler was actually registered on (`handle.channel`) rather than a caller-supplied one. Choosing to remove the parameter outright (over the "assert it matches" alternative) is the right call — it eliminates the misuse class at the type level instead of guarding it at runtime.

Verified:
- All 30 `UnSubscribe` references across production and test code were updated consistently to the single-arg signature — no caller was missed, and the three production call sites (`DataProviderSynchronizer`, `ReferenceCacheSynchronizer`, `SocketManagerService`) previously paired channel/id correctly, so there's no behavioral change for existing usage.
- `SocketManagerService.SocketChannel(...)` is not left as dead code — it's still used by `Publish` (line 182) and the subscribe path (line 217); only its use in the unsubscribe call was removed.
- The tuple/field access (`handle.channel`, `handle.handler`, `handle.worker`) is consistent with the new `(string channel, Action<...> handler, BackgroundWorker? worker)` shape, and the unsubscribe-before-dispose ordering (the #954 window) is preserved.

### Test coverage

Meets the backend testing guidelines. The genuine Redis subscription teardown is out-of-process logic, so it's correctly pinned by an **integration** test (`UnSubscribe_UnsubscribesTheChannelRecordedAtSubscribeTime`) rather than mocked: it subscribes, confirms delivery, unsubscribes, then asserts a subsequent publish on the same channel never reaches the handler — which directly proves the real subscription is torn down using the internally-recorded channel. The existing duplicate-id / unknown-id / reuse branches were carried over to the new signature. Good coverage for the behavior being changed.

### What's good

- The three added comments (interface, registry field, and the `UnsubscribeAsync` call) explain *why* the channel now comes from the handle, not just what changed — exactly the context a future reader needs at this lifecycle seam.
- The regression test asserts observable behavior (message no longer delivered), not an implementation detail, so it won't go stale against internal refactors.

No blocking findings, no should-fixes, no nits. CI is still running (backend + e2e in progress) — I'll merge once it's green.

---

PR #1968, opened 2026-07-14
Document the live-baseline view-model pattern in docs/frontend.md

### LINKED ISSUE #1840: Document the live-baseline view-model pattern in docs/frontend.md
Follow-up from PR #1838.

Three view-models — `SkillsView`, `AttributesView`, and `OptionsView` — now share the same pattern: the committed/baseline state is a live `$derived` read off the player manager, and unsaved edits are tracked as deltas/overrides that are invalidated via a reference-equality marker against the manager's `attributes`/`logPreferences` array the moment an external resync reassigns it.

This pattern is non-obvious and is the fix for a class of "constructor-snapshot goes stale on mid-session resync" bug (#1809). A short note in `docs/frontend.md` describing the pattern (and the reference-reassignment invariant it depends on) would help prevent the constructor-snapshot regression from recurring in future view-models.

Scope: docs-only.

ginderjeremiah:
## Summary

- Adds a new "Local-edit view-models: live baseline vs. a live-reference manager" section to `docs/frontend.md`, documenting the pattern `AttributesView`, `OptionsView`, and (with a different persistence shape) `SkillsView` all use to survive a mid-session resync without going stale.
- Explains the core invariant: a manager's array-valued state (`playerManager.attributes`, `.logPreferences`, `.selectedSkills`) is reassigned wholesale on resync, so a view-model's baseline must be a live `$derived` read off the manager, never a constructor snapshot (the bug class fixed by #1809).
- Describes the fuller delta/override + reference-equality-invalidation mechanism shared by `AttributesView`/`OptionsView` (`#pendingDeltas`/`#draftOverrides` gated against `#pendingDeltasAttrs`/`#draftOverridesPrefs`), including how `save()` preserves an in-flight edit (#1506).
- Notes that `SkillsView` protects the same live-baseline invariant but persists differently — an optimistic atomic-replace + serialized commit queue with rollback, rather than an accumulating delta — since its edits replace the whole loadout instead of layering partial changes. Flagging this distinction since the issue's phrasing suggested all three view-models share the same delta/override mechanism, which I confirmed in the code isn't quite accurate for `SkillsView`.

Docs-only change, no user-facing or behavioral effect.

## Test plan

- [x] Docs-only change; verified the described mechanics against the current `attributes-view.svelte.ts`, `options-view.svelte.ts`, and `skills-view.svelte.ts` source
- N/A — no build/test/lint surface touched

Closes #1840

Claude-Session: https://claude.ai/code/session_01T3feF7zmczoSvdUfcSS1Ht

---
_Generated by [Claude Code](https://claude.ai/code/session_01T3feF7zmczoSvdUfcSS1Ht)_


ginderjeremiah (commented):
## Review: Document the live-baseline view-model pattern in `docs/frontend.md` (closes #1840)

**Verdict:** Approve — the new section is fully accurate against the source and follows CLAUDE.md's guidance to document a non-obvious, cross-cutting pattern.

_(Submitted as a comment because GitHub doesn't accept a formal `APPROVE`/`REQUEST_CHANGES` on one's own PR; the intended verdict is **Approve**.)_

I verified every claim in the added section against the current `attributes-view.svelte.ts`, `options-view.svelte.ts`, and `skills-view.svelte.ts`:

- **Live baseline / wholesale reassignment invariant** — all three view-models read their baseline as a live `$derived` off the manager (`committed` / `baseline` / `equipped`), never a constructor snapshot. Matches the doc and the #1809 framing.
- **`AttributesView` / `OptionsView` delta+reference-marker mechanism** — `#pendingDeltas`+`#pendingDeltasAttrs` (`effectiveDeltas` gate) and `#draftOverrides`+`#draftOverridesPrefs` (`effectiveOverrides` gate) behave exactly as described: overrides read as empty the instant the recorded manager-array reference no longer `===` the live one, and `save()` re-pins to the post-save reference and consumes only the sent deltas, preserving an in-flight edit (#1506).
- **`SkillsView` distinction** — confirmed it does *not* use the delta/override overlay: `equipped` re-seeds `committed` from `playerManager.selectedSkills` while idle, and `commit()` optimistically writes the whole loadout to the manager, serializes through the queue, and rolls back to the last server-confirmed loadout on failure. The PR correctly flags that the issue's phrasing (all three sharing one delta mechanism) is inaccurate for `SkillsView` — the doc reflects the code, not the issue's imprecise framing, which is the right call.

Docs-only change; the doc-gated CI aggregate (`all-checks-passed`) is green and the code test/verify jobs were correctly path-skipped.

### What's good

- Correcting the issue's over-generalization rather than copying it — the `SkillsView`-is-different callout is the most valuable part of the section and prevents a future reader from assuming an overlay that isn't there.
- Anchoring the pattern to the concrete field/getter names makes it genuinely discoverable, which is the whole point of preventing the constructor-snapshot regression class from recurring.

---

PR #1970, opened 2026-07-14
Validate admin relationship references up front instead of 500ing on FK faults

### LINKED ISSUE #1858: Admin relationship setters inconsistently validate references — unsupported input yields opaque FK 500s instead of AdminSaveResult failures
## Problem

The tier's own convention (docs/backend-admin.md, `ReferenceFieldValidation`, `AdminItems.SaveModSlots`'s up-front owner check) is that a bad reference rejects as a graceful `AdminSaveResult` before staging. Three paths miss it, so a tampered/buggy client gets an unhandled `DbUpdateException` 500 instead:

- `AdminZones.SetEnemies` (`AdminZones.cs:154-176`) never checks `ze.EnemyId` exists (contrast `AdminEnemies.SetSkills`, which validates every skill id).
- `AdminEnemies.SetSpawns` (`AdminEnemies.cs:141-148`) — the Home guard `if (_zones.LookupZone(spawn.ZoneId) is { IsHome: true })` lets a **missing** zone (null lookup) pass silently; the insert then FK-faults at commit.
- `AdminTags.SaveTags` (`AdminTags.cs:22-40`) doesn't validate `TagCategoryId` against `ETagCategory`, though `ReferenceFieldValidation.FindUndefinedEnum` exists precisely for this (its own doc: "would otherwise surface as an opaque FK 500") and is applied to rarity/category/mod-type/challenge-type everywhere else.

No data corruption (the FK constraints hold) — but it is the exact defect class the codebase has repeatedly fixed elsewhere, so these look like omissions rather than decisions.

## Suggested fix

- `SetEnemies`: reject when `_enemies.GetEnemy(ze.EnemyId)` is null.
- `SetSpawns`: split the lookup into "missing → NotFound" then "IsHome → Failure".
- `SaveTags`: run `FindUndefinedEnum(changes, t => (ETagCategory)t.TagCategoryId, "tag category")`.

ginderjeremiah:
## Summary

Three admin relationship setters skipped the tier's own convention (`docs/backend-admin.md`, `ReferenceFieldValidation`, `AdminItems.SaveModSlots`'s up-front owner check) that a bad reference rejects as a graceful `AdminSaveResult` before staging. A tampered/buggy client hit an unhandled `DbUpdateException` 500 instead:

- **`AdminZones.SetEnemies`** never checked that a desired spawn's `EnemyId` exists (contrast `AdminEnemies.SetSkills`, which validates every skill id) — now rejects with `"Enemy {id} does not exist."`.
- **`AdminEnemies.SetSpawns`** — the Home guard `_zones.LookupZone(spawn.ZoneId) is { IsHome: true }` let a **missing** zone (null lookup) pass silently, since neither the Home nor the not-Home branch of the pattern match fires on `null`; the insert then FK-faulted at commit. Split into a missing-zone check (`"Zone {id} does not exist."`) followed by the existing `IsHome` rejection.
- **`AdminTags.SaveTags`** didn't validate `TagCategoryId` against `ETagCategory`, even though `ReferenceFieldValidation.FindUndefinedEnum` exists precisely for this and is already applied to rarity/category/mod-type/challenge-type everywhere else. Now wired in the same way.

No data corruption in any case (the FK constraints already held) — this replaces an opaque 500 with a clean, user-facing `AdminSaveResult.Failure`.

## Test plan

- [x] Added `AdminZonesIntegrationTests.SetEnemies_UnknownEnemy_ReturnsFailure`
- [x] Added `AdminEnemiesIntegrationTests.SetSpawns_UnknownZone_ReturnsFailure`
- [x] Added `AdminTagsTests.SaveTags_AddWithUndefinedCategory_ReturnsFailureWithoutStaging`, `SaveTags_EditWithUndefinedCategory_ReturnsFailureWithoutStaging`, `SaveTags_DeleteWithUndefinedCategoryPayload_Succeeds`
- [x] Full backend suite green: `dotnet build` + `dotnet test` (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests — 3143 tests, 0 failures)
- [x] No frontend changes in this PR

Closes #1858

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rs1UXHpqBC5wHa6S7ba5iJ)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminEnemiesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminEnemiesIntegrationTests.cs
@@ -468,6 +468,36 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SetSpawns_UnknownZone_ReturnsFailure()
+        {
+            int enemyId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var enemy = await TestDataSeeder.CreateEnemyAsync(context);
+                enemyId = enemy.Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            // A desired spawn referencing a zone that doesn't exist would otherwise FK-fault at commit
+            // instead of rejecting gracefully (the prior IsHome pattern match silently let a missing zone
+            // through since neither branch matched).
+            var data = new SetEnemySpawnsData
+            {
+                EnemyId = enemyId,
+                Spawns = [new Contracts.EnemySpawn { ZoneId = 99999, Weight = 1 }],
+            };
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminEnemies>();
+
+            var result = admin.SetSpawns(data);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal("Zone 99999 does not exist.", result.ErrorMessage);
+        }
+
+        [Fact]
         public async Task SetSpawns_NegativeWeight_ReturnsFailure()
         {
             int enemyId, zoneId;
--- a/Game.Application.Tests/DataAccess/AdminTagsTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminTagsTests.cs
@@ -66,5 +66,46 @@ namespace Game.Application.Tests.DataAccess
             Assert.Empty(store.Inserted);
             Assert.Empty(store.Updated);
         }
+
+        [Fact]
+        public void SaveTags_AddWithUndefinedCategory_ReturnsFailureWithoutStaging()
+        {
+            var store = new RecordingEntityStore();
+            var adminTags = new AdminTags(store);
+
+            // TagCategoryId 0 has no backing ETagCategory member (the enum starts at 1); without the up-front
+            // guard this would 500 on the FK at commit instead of rejecting gracefully.
+            var result = adminTags.SaveTags([Change(EChangeType.Add, id: 0, name: "Fire", categoryId: 0)]);
+
+            Assert.False(result.Succeeded);
+            Assert.Equal("0 is not a valid tag category.", result.ErrorMessage);
+            Assert.Empty(store.Inserted);
+        }
+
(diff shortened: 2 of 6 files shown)


ginderjeremiah (commented):
## Review: Validate admin relationship references up front (closes #1858)

**Verdict:** Approve — the PR implements exactly the fix #1858 calls for, aligns each path with the tier's own `ReferenceFieldValidation` / up-front-guard convention, and covers all three paths with tests. (Posted as a comment since I'm the PR author; GitHub won't accept a gating approval on one's own PR.)

Verified each change against the surrounding code and the reference-data conventions in `docs/backend*`:

- **`AdminEnemies.SetSpawns`** (`AdminEnemies.cs:152-165`) — splitting the old `is { IsHome: true }` pattern into an explicit `zone is null` check followed by `zone.IsHome` correctly closes the silent-null gap; the Home-zone message now reads from `zone` rather than the pattern binding, and behavior is unchanged for the existing cases.
- **`AdminZones.SetEnemies`** (`AdminZones.cs:175-183`) — the enemy-existence loop mirrors the boss/skill validators and runs after the Home-zone guard, so ordering is sensible.
- **`AdminTags.SaveTags`** (`AdminTags.cs:20-23`) — wiring `FindUndefinedEnum` matches its use for rarity/category/mod-type/challenge-type. Confirmed `FindUndefinedEnum` skips deletes and `ETagCategory` starts at `1`, so the 0/undefined-payload rejection and the delete-passthrough both behave as the tests assert.

### Test coverage

Solid and matches the project's rules — all three fixed paths are covered: `SetEnemies_UnknownEnemy_ReturnsFailure`, `SetSpawns_UnknownZone_ReturnsFailure`, plus the three `AdminTags` cases including the delete-with-undefined-category passthrough that guards against over-rejection. Assertions check both the failure and that nothing was staged, consistent with the existing sibling tests.

### Notes (non-blocking)

- The enemy/zone existence checks validate existence only, not retirement — consistent with the sibling validators (boss/skill checks do the same) and correct for this PR's scope (a retired row still satisfies the FK). No action needed.

Clean, well-scoped tech-debt fix with clear, appropriately-sized comments. Nice work.

---

PR #1973, opened 2026-07-14
Fix stale 10,000-tick comment in BattlePerformanceTests (#1861)

### LINKED ISSUE #1861: BattlePerformanceTests comments claim a 10,000-tick worst-case cap; the real cap is 3,000 ticks
## Problem

Both worst-case gate comments in `Game.Core.Tests/Battle/Performance/BattlePerformanceTests.cs:67` and `:101` claim `DefaultMaxBattleMs / MsPerTick = 10000` ("run to the 10,000-tick cap"). With `GameConstants.DefaultMaxBattleMs = 2 * 60 * 1000` and `MsPerTick = 40`, `WorstCaseTicks` actually computes **3,000**.

The constant is computed, so the gate itself is correct — but the `MaxLengthRealisticCeilingMs = 250.0` headroom rationale ("measures ~10ms on a dev box, an order of magnitude under the ceiling") is calibrated against a battle 3.3× shorter than the comments describe. A future reader tuning the ceiling from the comments would misjudge the margin on the one machine-dependent gate protecting the anti-cheat replay hot path.

## Suggested fix

Correct both comments to 3,000 ticks — or express the figure as the formula and drop the literal so it can't go stale again.

ginderjeremiah:
## Summary

`BattlePerformanceTests.cs` had two comments claiming the worst-case battle backstop runs "to the 10,000-tick cap." That was never accurate: `WorstCaseTicks = GameConstants.DefaultMaxBattleMs / GameConstants.MsPerTick` computes `120000 / 40 = 3000`, not 10,000.

Rather than swap in the literal `3,000` (which would just go stale again the next time either constant changes), I reworded both comments to reference the formula/constant instead of a hardcoded tick count:

- Line 67: `DefaultMaxBattleMs / MsPerTick = 10000` → `DefaultMaxBattleMs / MsPerTick` (drops the stale literal, points at the formula already used two lines below).
- Line 101: "run to the 10,000-tick cap" → "run to the `WorstCaseTicks` cap" (names the actual constant).

No behavioral change — this is a comment-only fix. The gate itself (`WorstCaseTicks` computed from the real constants) was always correct; only the prose describing it was wrong.

Closes #1861

## Test plan

- [x] `dotnet build Game.sln` — clean, 0 warnings
- [x] `dotnet test Game.sln --no-build --no-restore` — all 3148 backend tests pass

---
_Generated by [Claude Code](https://claude.ai/code/session_019GeUVfHhHFGrzMBzKMEa11)_

### DIFF
--- a/Game.Core.Tests/Battle/Performance/BattlePerformanceTests.cs
+++ b/Game.Core.Tests/Battle/Performance/BattlePerformanceTests.cs
@@ -64,7 +64,7 @@ namespace Game.Core.Tests.Battle.Performance
     public class BattlePerformanceTests
     {
         // Per-tick cost cancels the tick count, so the scaling gates use a modest fixed length to stay
-        // fast; the backstop uses the simulator's real default cap (DefaultMaxBattleMs / MsPerTick = 10000).
+        // fast; the backstop uses the simulator's real default cap (DefaultMaxBattleMs / MsPerTick).
         private const int ScalingTicks = 800;
         private const int WorstCaseTicks = GameConstants.DefaultMaxBattleMs / GameConstants.MsPerTick;
 
@@ -98,7 +98,7 @@ namespace Game.Core.Tests.Battle.Performance
         private const int TypicalBattleTicks = 750;         // ~30s at MsPerTick = 40
 
         // Coarse catastrophic-regression ceiling for the realistic worst-case battle (the typical full
-        // loadout — 4 skills, 2 multipliers + a churning effect each — run to the 10,000-tick cap). This is
+        // loadout — 4 skills, 2 multipliers + a churning effect each — run to the WorstCaseTicks cap). This is
         // the one machine-dependent gate: a uniform per-tick regression inflates every battle equally, which
         // the ratio gates cannot see. The margin is deliberately huge (the battle measures ~10ms on a dev
         // box, an order of magnitude under the ceiling); if extreme runner contention ever makes it flake,


ginderjeremiah (commented):
## Review: Fix stale 10,000-tick worst-case comments in `BattlePerformanceTests` (closes #1861)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change fixes the exact defect the issue describes, and does it the way the issue's preferred option recommends.

The math checks out: `DefaultMaxBattleMs / MsPerTick = 120000 / 40 = 3000`, so the old "10,000-tick cap" prose was wrong on both lines. Rather than swapping in the literal `3,000` — which would drift again the moment either constant changes — both comments now reference the formula/constant (`DefaultMaxBattleMs / MsPerTick` and `WorstCaseTicks`). That's the more durable of the two fixes the issue suggested and eliminates the failure mode entirely instead of resetting it.

### Correctness
- Line 67: dropping the `= 10000` literal is accurate and self-consistent — `WorstCaseTicks` is defined from those same constants two lines down.
- Line 101: "run to the `WorstCaseTicks` cap" now names the real constant, so a future reader tuning `MaxLengthRealisticCeilingMs` is pointed at the actual value rather than a figure 3.3× too high.

### Test coverage
None needed — comment-only change with no behavioral effect. The gate itself was always computed from the real constants and is unchanged.

### What's good
Chose the self-maintaining wording over a fresh literal, and kept the fix minimal and scoped. Nothing to change.


---

PR #1974, opened 2026-07-14
Re-derive the class signature passive in CloneWithAttributeDelta

### LINKED ISSUE #1862: Battler.CloneWithAttributeDelta freezes the class signature passive, so CombatRating.Marginal misprices attribute-scaled passives (latent)
## Problem

`Battler.CloneWithAttributeDelta` (`Battler.cs:180-195`) rebuilds the battler from `AllModifiers()` minus `SkillEffect`-source and static modifiers, then adds the delta. A `EAttributeModifierSource.Class` signature-passive modifier is copied at its **already-resolved amount** (`ClassSignaturePassive.GetModifier` resolved it against the original attributes at assembly in `BattleSnapshot.ToBattler`), so bumping the passive's scaling attribute does not re-derive the passive in the clone.

`CombatRating.Marginal(battler, isPlayer: true, X)` (`CombatRating.cs:81-86`) therefore reports **exactly 0** for a core attribute whose only combat contribution is feeding an attribute-scaled signature passive — the same "dead-stat" signature the finite-difference detector exists to flag, but as a false positive.

**Currently latent (verified):** the only production caller of `Marginal` today is `ProgressionGraphChecker` with `isPlayer: false` on enemy battlers, which carry no class passive; player-side `Marginal` is exercised only by tests. But the documented contract ("full cascade re-derivation included… exactly like a real allocation would") doesn't hold, and a future player-side lint (e.g. class-kit dead-stat checking for #1226/#1219) would inherit the gap silently.

## Suggested fix

Either re-resolve the signature passive in the clone (thread the passive/class through, mirroring `ToBattler`'s final step), or document the exclusion on `CloneWithAttributeDelta`/`Marginal` so the limitation is explicit at the call sites that matter.

ginderjeremiah:
## Summary

`Battler.CloneWithAttributeDelta` rebuilds a battler's modifier set (minus `SkillEffect` and static modifiers) plus a fresh delta on one attribute, for `CombatRating.Marginal`'s finite-difference pricing. It was copying the `EAttributeModifierSource.Class` signature-passive modifier through at its **already-resolved** amount (frozen at battler assembly), instead of re-deriving it. For an attribute-scaled passive (`Amount + ScalingAttribute × ScalingAmount`), bumping the scaling attribute in the clone left the passive's contribution unchanged, so `CombatRating.Marginal` under-reported (in the worst case, exactly `0`) for a core attribute whose only combat contribution was feeding that passive — a false "dead stat" positive.

Currently latent per the issue: the only production caller of `Marginal` today is `ProgressionGraphChecker` with `isPlayer: false` on enemy battlers (no class passive), so nothing user-facing broke. But the documented contract on `CloneWithAttributeDelta` ("full cascade re-derivation included... exactly like a real allocation would") didn't hold, and a future player-side dead-stat lint would have silently inherited the gap.

## Changes

- `Battler` now optionally carries the `ClassSignaturePassive` definition it was assembled with (not just the resolved modifier already folded into its attributes).
- `CloneWithAttributeDelta` excludes the frozen `Class`-sourced modifier (alongside the existing `SkillEffect`/static exclusions) and, when a signature passive is carried, re-resolves it against the clone's bumped attributes as the final step — mirroring `BattlerMaterials.Build`'s own composition order (locked base → gear → proficiency → class passive last).
- `BattleSnapshot.GetBattlerMaterials` now threads the `ClassSignaturePassive` definition through `BattlerMaterials` in addition to the pre-resolved modifier used for the battler's initial assembly, so both the constructed battler and any future clone stay consistent with the snapshot's captured class.

## Test plan

- [x] Added `BattlerTests.CloneWithAttributeDelta_AttributeScaledSignaturePassive_ReDerivesAgainstBumpedAttribute` — pins the exact regression (an Endurance-scaled Toughness passive correctly re-derives on a bumped clone; a frozen-passive bug would instead yield a stale value).
- [x] Added `BattlerTests.CloneWithAttributeDelta_FlatSignaturePassive_CarriesThroughUnaffectedByAnUnrelatedBump` — confirms the common flat-passive case is unaffected (no duplication from the exclude/re-add swap).
- [x] `dotnet build` (full solution) — clean.
- [x] `dotnet test` (full solution: Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests) — all passing (3154 tests, 0 failures).

Closes #1862

---
_Generated by [Claude Code](https://claude.ai/code/session_015qMVJ95XKisuVVWcFU1vAh)_

### DIFF
--- a/Game.Core.Tests/Battle/BattlerTests.cs
+++ b/Game.Core.Tests/Battle/BattlerTests.cs
@@ -1,6 +1,7 @@
 using Game.Core.Attributes;
 using Game.Core.Attributes.Modifiers;
 using Game.Core.Battle;
+using Game.Core.Classes;
 using Game.Core.Skills;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Re-derive the class signature passive in `CloneWithAttributeDelta` (closes #1862)

**Verdict:** Approve — a correct, tightly-scoped domain fix that resolves the latent mispricing exactly as the issue describes; no blocking findings, should-fixes, or nits. _(Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating event on one's own PR — the intended verdict is **Approve**.)_

### Why this is correct

- `CloneWithAttributeDelta` now excludes the frozen `EAttributeModifierSource.Class` modifier alongside the existing `SkillEffect`/static exclusions, then re-resolves the carried `ClassSignaturePassive` against the clone's bumped attributes **as the final step** — reading `attributes.GetAttributeValue` before the passive is added, which matches `BattlerMaterials.Build`'s composition order (locked base → gear → proficiency → passive last) and `ClassSignaturePassive.GetModifier`'s "already-assembled value" contract. Composition stays consistent between the baseline and the `+delta` clone, so the finite difference in `CombatRating.Marginal` is once again meaningful.
- The passive definition is threaded through `Battler` → `CloneWithAttributeDelta` → `Battler`, so a clone-of-a-clone re-derives too, not just the first hop.
- `BattleSnapshot.GetBattlerMaterials` threads the definition through `BattlerMaterials` while still resolving the modifier once for the battler's own assembly; `ResolveSignaturePassive` is retained (still used by `GetModifiersWithSignaturePassive`), so no dead code is left behind.
- Correctly scoped as backend-only: `CloneWithAttributeDelta`/`Marginal` price combat rating and never run in the live simulation, so there is no frontend battle-parity mirror to update — consistent with the issue's analysis.

### Test coverage

Adequate and well-targeted, per the project's domain-testing rules (classical style, no test doubles):
- `CloneWithAttributeDelta_AttributeScaledSignaturePassive_ReDerivesAgainstBumpedAttribute` pins the exact regression — an Endurance-scaled Toughness passive re-derives to `75` on a bumped clone, where the frozen-passive bug yielded `70`.
- `CloneWithAttributeDelta_FlatSignaturePassive_CarriesThroughUnaffectedByAnUnrelatedBump` guards the common flat case against duplication from the exclude/re-add swap.

### What's good

Careful, minimal change with an accurate updated doc-comment on `CloneWithAttributeDelta` explaining the exclusion-and-re-resolve, and a clean split of `ResolveSignaturePassiveDefinition` out of `ResolveSignaturePassive` so the definition and the resolved modifier stay in sync.


---

PR #1975, opened 2026-07-14
Fix Workbench retire-confirm gaps: Path retire bypass and unsaved-edit blindness

### LINKED ISSUE #1863: Retire-confirm gaps: Path retire bypasses the referenced-by confirm, and the reference computation ignores the session's own unsaved edits
Two related gaps in the Workbench retire-confirm surface. The backend guards still protect integrity in both cases — these are UX/consistency issues.

## 1. Retiring a Path skips the confirm every other retire path gets

`PathDetail.svelte:17` wires `onRetire={() => store.retirePath(path.id, true)}` directly, while `TierDetail.svelte` and `WorkbenchDetail` route through `retireWithConfirm(...)`. `computeReferences` (`workbench/references.ts`) has cases for enemies/zones/challenges/items/itemMods/skills/proficiencies but none for `paths` (falls through to `[]`).

Retiring a path whose tiers gate a live gateway is rejected by the backend guard (`FindRetiredPathGatingLiveGateway`) — the admin gets a raw save-failure toast with no pre-warning and no reference surface, inconsistent with every other retire.

**Fix:** route path retire through `retireWithConfirm` and add a `'paths'` branch to `computeReferences` (gateway prerequisites / tiers of the path).

## 2. The "referenced by" dialog reads last-saved cache, ignoring unsaved edits

`retire-confirm.ts:10-21` (`referenceSourcesFromStatic`) computes the advisory references over `staticData` (last-saved state), while the store being edited may hold unsaved references to the record being retired (a just-added gateway prerequisite, a reward re-pointed this session). The dialog under-reports — or skips entirely for an "unreferenced" record — when the only references are this session's own pending edits, so an admin can retire something their unsaved work depends on without warning.

**Fix:** for the catalogue currently being edited, source `ReferenceSources` from the live store records instead of `staticData`.

ginderjeremiah:
## Summary

Two related gaps in the Workbench retire-confirm surface, per #1863:

1. **Path retire skipped the "referenced by" confirm.** `PathDetail.svelte` wired `onRetire` directly to `store.retirePath(...)`, bypassing the confirm dialog every other retire path (enemies, zones, challenges, items, itemMods, skills, proficiencies) already gets. Retiring a path whose tier gates a live gateway would silently hit the backend's `AdminPaths.FindRetiredPathGatingLiveGateway` guard and surface as a raw save-failure toast with no pre-warning.
   - Added a `'paths'` case to `computeReferences` (`references.ts`): a path is referenced when another still-live proficiency lists one of the path's tiers as a cross-path prerequisite — mirroring the exact soft-lock condition the backend guard rejects. Marked as a `strong` reference (a real save-time block, not just advisory) so the dialog closes with the "re-point first" wording.
   - Routed `PathDetail`'s retire through `retireWithConfirm`, mirroring `TierDetail`'s existing wiring.

2. **The "referenced by" dialog read last-saved data, ignoring this session's unsaved edits.** `referenceSourcesFromStatic()` always built its `ReferenceSources` from `staticData` (last-saved cache), so a reference that only exists in this session's pending edits (a just-added spawn-table entry, a gateway prerequisite added moments ago) was invisible to the confirm dialog.
   - `referenceSourcesFromStatic` now accepts an optional `overrides` param to substitute live records for the one catalogue currently being edited, while every other catalogue still comes from `staticData` as before.
   - `WorkbenchDetail.svelte` passes its own `store.items` (live, unsaved-deletes excluded) for whichever entity catalogue it maps to.
   - `PathDetail.svelte` and `TierDetail.svelte` (the progression editor) both pass `store.profs` (live proficiencies) since gateway prerequisite edits live there.

## Test plan

- [x] `references.test.ts` — new `computeReferences('paths', …)` cases (soft-lock reference, same-path prerequisite ignored, unreferenced path) plus a `formatReferenceBody` case for the strong path reference.
- [x] `retire-confirm.test.ts` — new case covering `referenceSourcesFromStatic`'s override substitution.
- [x] `TierDetail.test.ts` — new case proving the tier retire-confirm now sources from the live store, not stale `staticData`.
- [x] `PathDetail.test.ts` (new file) — retire-confirm lifecycle: no prompt when unreferenced, prompts + retires on confirm (using a live-only unsaved gating edge), cancel leaves it unretired, Reinstate for an already-retired path.
- [x] `npm run test:unit:ci` — full suite, 299 files / 3681 tests passed.

Closes #1863

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ctvy8cxprb1oPzSKTyWGaz)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
+++ b/UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
@@ -133,6 +133,7 @@
 import { fieldsOf, type EntityConfig, type Identified } from '../entities/types';
 import type { EntityStore } from '../entity-store.svelte';
 import { recordsEqual } from '../entity-store.svelte';
+import type { ReferenceSources } from '../references';
 import { referenceSourcesFromStatic, retireWithConfirm } from '../retire-confirm';
 import { sectionWarnings } from '../validation';
 import WorkbenchIcon from '../WorkbenchIcon.svelte';
@@ -152,6 +153,35 @@ interface Props {
 const { entity, store, record, baseline, tab, onTab, onNew }: Props = $props();
 
 /**
+ * The catalogue this detail pane is currently editing, live (unsaved edits included) — substituted
+ * for `staticData`'s last-saved copy of that one slot so a reference computed from this session's
+ * own pending edits (e.g. a just-added spawn-table entry) isn't missed (#1863). Only entity keys
+ * that map to a {@link ReferenceSources} slot need one; the rest (itemMods, tags, lessons) aren't
+ * themselves reference sources, so they fall through unchanged.
+ */
+const liveOverride = (): Partial<ReferenceSources> => {
+	const live = store.items.filter((rec) => store.status(rec) !== 'deleted');
+	switch (entity.key) {
+		case 'enemies':
+			return { enemies: live as unknown as ReferenceSources['enemies'] };
+		case 'zones':
+			return { zones: live as unknown as ReferenceSources['zones'] };
+		case 'challenges':
+			return { challenges: live as unknown as ReferenceSources['challenges'] };
+		case 'items':
+			return { items: live as unknown as ReferenceSources['items'] };
+		case 'classes':
+			return { classes: live as unknown as ReferenceSources['classes'] };
+		case 'skillRecipes':
+			return { skillRecipes: live as unknown as ReferenceSources['skillRecipes'] };
+		case 'skills':
+			return { skills: live as unknown as ReferenceSources['skills'] };
+		default:
+			return {};
+	}
+};
+
+/**
  * Retire a saved reference record, first surfacing what currently references it. The referenced-by
  * surface is computed from the cached reference sets (no backend round-trip); an unreferenced record
  * retires without an extra prompt. Advisory only — confirming proceeds, and the record stays
@@ -163,7 +193,7 @@ const onRetire = (rec: Identified) =>
 		id: rec.id,
 		name: entity.title?.(rec) || rec.name || entity.blankName,
 		title: `Retire ${entity.singular}?`,
-		sources: referenceSourcesFromStatic(),
+		sources: referenceSourcesFromStatic(liveOverride()),
 		onConfirmed: () => store.setRetired(rec.id, true)
 	});
 
--- a/UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
+++ b/UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
@@ -133,7 +133,6 @@
 import { fieldsOf, type EntityConfig, type Identified } from '../entities/types';
 import type { EntityStore } from '../entity-store.svelte';
 import { recordsEqual } from '../entity-store.svelte';
-import type { ReferenceSources } from '../references';
 import { referenceSourcesFromStatic, retireWithConfirm } from '../retire-confirm';
 import { sectionWarnings } from '../validation';
 import WorkbenchIcon from '../WorkbenchIcon.svelte';
@@ -153,39 +152,19 @@ interface Props {
 const { entity, store, record, baseline, tab, onTab, onNew }: Props = $props();
 
 /**
- * The catalogue this detail pane is currently editing, live (unsaved edits included) — substituted
- * for `staticData`'s last-saved copy of that one slot so a reference computed from this session's
- * own pending edits (e.g. a just-added spawn-table entry) isn't missed (#1863). Only entity keys
- * that map to a {@link ReferenceSources} slot need one; the rest (itemMods, tags, lessons) aren't
- * themselves reference sources, so they fall through unchanged.
- */
-const liveOverride = (): Partial<ReferenceSources> => {
-	const live = store.items.filter((rec) => store.status(rec) !== 'deleted');
-	switch (entity.key) {
-		case 'enemies':
-			return { enemies: live as unknown as ReferenceSources['enemies'] };
-		case 'zones':
-			return { zones: live as unknown as ReferenceSources['zones'] };
-		case 'challenges':
-			return { challenges: live as unknown as ReferenceSources['challenges'] };
-		case 'items':
-			return { items: live as unknown as ReferenceSources['items'] };
-		case 'classes':
-			return { classes: live as unknown as ReferenceSources['classes'] };
-		case 'skillRecipes':
-			return { skillRecipes: live as unknown as ReferenceSources['skillRecipes'] };
-		case 'skills':
-			return { skills: live as unknown as ReferenceSources['skills'] };
-		default:
-			return {};
-	}
-};
-
-/**
  * Retire a saved reference record, first surfacing what currently references it. The referenced-by
  * surface is computed from the cached reference sets (no backend round-trip); an unreferenced record
  * retires without an extra prompt. Advisory only — confirming proceeds, and the record stays
  * resolvable by id either way (see references.ts).
+ *
+ * Sourced from `staticData` (last-saved), not this pane's own live `store.items`: every reference
+ * fn here reads a *sibling* catalogue (an enemy is referenced by zones/challenges, an item by
+ * challenges/classes, …), never the one this pane edits, and `store.items` isn't dense-by-id once a
+ * record's been added this session (`EntityStore.addItem` prepends negative ids) — indexing it by
+ * id, as `enemies[id]?.spawns` and `nameByIndex(skills, …)` do, would resolve the wrong record. See
+ * the progression editor's `{ proficiencies: store.profs }` override for the shape this pane would
+ * need (index-safe access to every referencing catalogue's own live store) to close this gap
+ * generically — tracked as a follow-up rather than attempted here.
  */
 const onRetire = (rec: Identified) =>
 	retireWithConfirm({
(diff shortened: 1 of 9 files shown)


UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
```diff
@@ -151,6 +152,35 @@ interface Props {
...
+		case 'classes':
+			return { classes: live as unknown as ReferenceSources['classes'] };
+		case 'skillRecipes':
+			return { skillRecipes: live as unknown as ReferenceSources['skillRecipes'] };
+		case 'skills':
+			return { skills: live as unknown as ReferenceSources['skills'] };
+		default:
+			return {};
+	}
+};
```
ginderjeremiah: **Blocking — this override breaks the id-as-index invariant.** `store.items` is not dense-by-id: `addItem` prepends new records with negative ids (`[newItem(id), ...items]`, `nextId` starts at -1), and this `.filter(status !== 'deleted')` also drops saved records that were hard-removed this session, leaving gaps. But `enemyReferences` reads the retired enemy's spawns via `enemies[id]?.spawns` (references.ts:72), and `skillReferences` names recipes via `nameByIndex(skills, …)` — both index the *overridden* slot by id, relying on the same zero-based Id-as-index invariant documented in `backend.md`.

So the moment the session has any added enemy (a completely normal workflow: add a new enemy, then retire an existing one), `enemies[id]` resolves to the *wrong* record and the "spawns in zone(s) …" surface is silently incorrect — a regression versus the dense `staticData` array it replaces.

Worse, this override can't actually deliver gap #2 for the generic entities anyway: an enemy is referenced by *zones/challenges*, an item by *challenges/classes*, a skill by *everything else* — never by the catalogue being edited. This pane only holds its own `store`, so it can't source those sibling catalogues' unsaved edits. The result is that the `zones`/`challenges`/`items`/`classes`/`skillRecipes` cases are inert no-ops (their reference fn never reads their own slot), and the only slots the reference fns *do* read by id — `enemies` and `skills` — are exactly the ones this corrupts.

Suggested resolution: drop the generic override entirely and keep the two progression overrides (`{ proficiencies: store.profs }`), which are correct because `pathReferences`/`proficiencyReferences` look up proficiencies by `.filter`, not by index. If you want to preserve the narrow "reflect this enemy's own unsaved spawn edits" case, source it density-safe (e.g. start from the dense `staticData.enemies` and overwrite `arr[rec.id] = rec` for `rec.id >= 0`) rather than passing the prepend-ordered live array.

Nit while here: the `as unknown as ReferenceSources['…']` double-casts defeat type-checking — a future mismatch between `entity.key` and the store's element type wouldn't be caught.

---

ginderjeremiah (commented):
## Review: Workbench retire-confirm gaps — Path retire bypass + unsaved-edit blindness (#1863)

**Verdict: Request changes** — the Path-retire fix (gap #1) and the progression-editor `store.profs` overrides (gap #2) are correct and well-tested, but the generic `WorkbenchDetail.liveOverride` introduces an id-as-index correctness bug and doesn't actually address gap #2 for the entities it targets. _(Submitted as a COMMENT because GitHub won't accept a gating verdict on my own PR — the effective verdict is Request changes.)_

### Blocking

- **`WorkbenchDetail.svelte:162-182` — override violates the Id-as-index invariant.** `store.items` is not dense-by-id (adds are prepended with negative ids; `.filter('deleted')` drops hard-removed saved records), yet `enemyReferences` reads `enemies[id]?.spawns` and `skillReferences` names recipes via `nameByIndex(skills, …)` — both index the overridden slot by id. After a normal *add-an-enemy → retire-another-enemy* session the "spawns in …" surface reads the wrong record. Detail inline.
- **Same override is structurally ineffective for gap #2.** Generic entities are referenced by *other* catalogues (enemy←zones/challenges, item←challenges/classes, skill←everything-else), and this pane only holds its own `store`, so it can't source those catalogues' unsaved edits. The `zones`/`challenges`/`items`/`classes`/`skillRecipes` cases are inert (their reference fn never reads their own slot); only `enemies`/`skills` are read by id — the two the override corrupts. Recommend dropping the generic override and keeping the progression ones, or reworking it density-safe (inline).

### Should-fix

- **`WorkbenchDetail` has no test.** The new `liveOverride` logic is untested (`tests/routes/admin/workbench/components/` only has `AttributePicker`/`attribute-groups`). Whatever survives the rework needs a test — including the add-then-retire density case, which is exactly what would have caught the bug above. Per the project's testing rules, a touched component with new interaction logic needs unit coverage.

### Nits

- `as unknown as ReferenceSources['…']` double-casts (166-178) defeat type-checking on the `entity.key`→element-type mapping.
- Path clause reads "*&lt;path name&gt; is a prerequisite for …*" (formatReferenceBody), but it's the path's *tier* that's the prerequisite, not the path. Minor advisory-wording imprecision — the `pathReferences` comment already notes "referenced through its tiers".

### Test coverage

- **Solid:** `pathReferences` cases in `references.test.ts` (soft-lock, same-path prereq ignored, unreferenced), the `formatReferenceBody` strong-path case, the `TierDetail` live-store case, the `referenceSourcesFromStatic` override substitution, and the new `PathDetail.test.ts` retire-confirm lifecycle. Gap #1 and the progression side of gap #2 are well covered.
- **Missing:** any `WorkbenchDetail` test exercising the generic override — and specifically the non-dense-array interaction. The current tests all feed hand-built dense sources, so they can't surface the id-as-index bug.

### What's good

- Gap #1 is cleanly fixed: routing `PathDetail` retire through `retireWithConfirm` mirrors `TierDetail`, and `pathReferences` faithfully mirrors the backend `FindRetiredPathGatingLiveGateway` soft-lock (`strong: true` for the real save-time block).
- The progression-editor `{ proficiencies: store.profs }` overrides are the right call — order-independent because those reference fns filter by property, and they genuinely close gap #2 where the referencing catalogue *is* the one being edited.
- Comments are clear and tie each change back to the backend guard they mirror.

---

ginderjeremiah:
Good catch — confirmed `EntityStore.addItem` does prepend negative ids (`entity-store.svelte.ts:137`), so `store.items` is only index-aligned before the first add in a session. Fixed in ec09c4c:

- Dropped the generic `WorkbenchDetail.liveOverride` entirely; reverted to `referenceSourcesFromStatic()` with no override. The progression-editor overrides (`{ proficiencies: store.profs }` in `PathDetail`/`TierDetail`) are untouched — those are safe because `pathReferences`/`proficiencyReferences` filter by property, never index by id.
- Filed #1976 to track the real generic fix, which needs the Workbench to thread every sibling catalogue's live records (not just the one entity open) plus a density-safe id accessor for the two index-based lookups — an architecture question, not a small fix.
- Fixed the path-clause wording nit: `prerequisiteOf` now reads "home to a tier that gates N proficiencies" for paths, vs. "a prerequisite for N proficiencies" for proficiencies themselves.
- Updated/added tests accordingly; full suite (3682 tests), lint, and svelte-check all green.

The double-cast nit is moot now that the override (and its casts) is gone.

---

ginderjeremiah:
## Review — ✅ Approve

Reviewed the diff against the backend guard it mirrors and the project's Svelte 5 / testing conventions.

**Gap 1 — Path retire bypassed the confirm dialog:** fully fixed. `PathDetail.svelte` now routes retire through `retireWithConfirm(...)` (mirroring `TierDetail`), and `references.ts` gains a `'paths'` branch. I verified `pathReferences` against the backend guard `AdminPaths.FindRetiredPathGatingLiveGateway`: it computes the same soft-lock condition (a still-live proficiency on another path listing one of the retiring path's tiers as a cross-path prerequisite) and correctly marks it `strong: true` for a real save-time block. Faithful mirror, including keying off path retirement rather than per-proficiency `retiredAt` (as the backend does). ✓

**Gap 2 — dialog ignored unsaved edits:** fixed for the progression editor via the new `overrides` param on `referenceSourcesFromStatic`, with `PathDetail`/`TierDetail` passing live `store.profs`. The property-based `.filter`/`Set` lookup is index-safe, so `EntityStore.addItem`'s prepended negative unsaved ids resolve correctly — which is exactly why the progression override is sound where the generic `WorkbenchDetail` override was not. Reverting the generic override and splitting it into #1976 is a defensible scope call, not a regression.

**Prior review resolved:** the earlier self-review's blocking items all targeted the generic `WorkbenchDetail.liveOverride`; commit `ec09c4c` dropped that override (leaving `WorkbenchDetail` a doc-comment-only change), filed #1976, and fixed the clause-wording nit. Nothing outstanding since.

**Tests:** appropriately scoped — new `PathDetail.test.ts` lifecycle, `pathReferences` edge cases (incl. same-path-prereq ignored), the live-store `TierDetail` case, and the override-substitution test. Proper Svelte 5 runes (`$props`/`$derived`), no dead code.

**CI:** green — `test-frontend`, `verify-frontend`, all three `test-e2e` browsers success; backend jobs correctly skipped; `all-checks-passed` success.

**Nits (non-blocking, not gating merge):**
- `references.ts:155` — `pathReferences` approximates the backend's `PathStaysLive(gateway.PathId)` as `p.pathId !== id`, so a gateway on a *different, already-retired* path is over-reported as a strong reference. Advisory over-report only (never a miss or bypass; backend authoritative). Tracked in #1983 to tighten later.
- The dialog offers "Retire anyway" on a `strong` path soft-lock even though the backend will reject the save — slightly overpromising, but this is a pre-existing pattern shared with the `unlockGate`/challenge strong cases, not introduced here. Left as-is.

No blocking issues, no should-fixes. Deferred generic case tracked in #1976, advisory-accuracy nit in #1983. Merging.

---

PR #1977, opened 2026-07-14
Fix scroll-lock refcount, toast duration collapse, and stale tour-callout position

### LINKED ISSUE #1864: Small component/store cleanups: focus-trap scroll-lock refcount, toast duplicate-collapse duration rewrite, tour callout stale position on shared anchors

ginderjeremiah:
## Summary

Three small, independently-verified bugs in shared UI primitives, per #1864:

1. **Focus-trap scroll-lock is not reference-counted.** `focus-trap.ts` toggled `document.body.classList` directly for its `scrollLockClass`. Two stacked traps sharing the same class (both `Popover`s hardcode `scrollLockClass: 'popover-open'`) would have the inner trap's `destroy()` remove the class while the outer overlay was still open, unlocking background scroll prematurely. Fixed with a module-level refcount map (`acquireScrollLock`/`releaseScrollLock`): the class is added on 0→1 and removed on 1→0.

2. **Toast duplicate-collapse silently rewrote the surviving toast's persistence.** `showToast` collapsed a duplicate message into the existing toast but re-armed its timer with the **new** call's `duration` — a persistent toast (`duration: 0`) hit by a duplicate omitting `duration` would start auto-dismissing after 6s, and a timed toast hit by a `duration: 0` duplicate would become permanent. `ToastData` now stores each toast's own `duration`, and a collapse reuses that stored value instead of the incoming duplicate's.

3. **Tour callout position went stale between same-anchor steps.** `TourPlayer`'s position `$effect` re-ran on `resizeTick`/`open`/`anchorEl` but not when the step's own text changed. Since `anchorEl` is a `$derived` reference (`getTutorialAnchor(anchorKey)`), two consecutive steps sharing an `anchorKey` (or both unanchored) left the object reference unchanged, so the effect never re-ran even though the callout's rendered height changed — leaving it slightly off / clipped near a viewport edge. Fixed by also tracking `stepIndex` in the effect (mirroring the existing `resizeTick` tracking convention).

## Test plan

- [x] `focus-trap.test.ts` — new case: two stacked traps sharing a scroll-lock class stay locked until the last one destroys.
- [x] `toast.test.ts` — new cases: a persistent toast stays persistent when a timed duplicate collapses into it, and a timed toast keeps auto-dismissing on its own schedule when a persistent duplicate collapses into it.
- [x] `TourPlayer.test.ts` — new case: a step sharing the prior step's anchor recomputes the callout position when the rendered height changes (mocked jsdom layout, since jsdom doesn't lay out real box sizes).
- [x] `npm run test:unit:ci` — full suite, 298 files / 3674 tests passed.

Closes #1864

---
_Generated by [Claude Code](https://claude.ai/code/session_01X7D47gWFsDZJoAaML4kSAN)_

### DIFF
--- a/UI/src/components/focus-trap.ts
+++ b/UI/src/components/focus-trap.ts
@@ -17,6 +17,31 @@
 const trapStack: HTMLElement[] = [];
 
 /**
+ * Refcounts per scroll-lock class name, so two stacked traps sharing a class (e.g. both `Popover`s
+ * using `popover-open`) don't unlock background scroll when the inner one destroys while the outer
+ * is still open.
+ */
+const scrollLockRefs = new Map<string, number>();
+
+function acquireScrollLock(className: string) {
+	const count = (scrollLockRefs.get(className) ?? 0) + 1;
+	scrollLockRefs.set(className, count);
+	if (count === 1) {
+		document.body.classList.add(className);
+	}
+}
+
+function releaseScrollLock(className: string) {
+	const count = (scrollLockRefs.get(className) ?? 0) - 1;
+	if (count <= 0) {
+		scrollLockRefs.delete(className);
+		document.body.classList.remove(className);
+	} else {
+		scrollLockRefs.set(className, count);
+	}
+}
+
+/**
  * Tab-reachable elements. Intentionally broader than a bare `button` so a modal/popover containing a
  * link, input, or `[tabindex]` element still traps and anchors onto it.
  */
@@ -70,7 +95,7 @@ export function focusTrap(node: HTMLElement, options: FocusTrapOptions = {}) {
 	trapStack.push(node);
 	window.addEventListener('keydown', onKeydown);
 	if (opts.scrollLockClass) {
-		document.body.classList.add(opts.scrollLockClass);
+		acquireScrollLock(opts.scrollLockClass);
 	}
 
 	return {
@@ -78,10 +103,10 @@ export function focusTrap(node: HTMLElement, options: FocusTrapOptions = {}) {
 			// Swap the scroll-lock class if it changed (rare); options are otherwise read live.
 			if (opts.scrollLockClass !== next.scrollLockClass) {
 				if (opts.scrollLockClass) {
-					document.body.classList.remove(opts.scrollLockClass);
+					releaseScrollLock(opts.scrollLockClass);
 				}
 				if (next.scrollLockClass) {
-					document.body.classList.add(next.scrollLockClass);
+					acquireScrollLock(next.scrollLockClass);
 				}
 			}
 			opts = next;
@@ -93,7 +118,7 @@ export function focusTrap(node: HTMLElement, options: FocusTrapOptions = {}) {
 			}
 			window.removeEventListener('keydown', onKeydown);
 			if (opts.scrollLockClass) {
-				document.body.classList.remove(opts.scrollLockClass);
+				releaseScrollLock(opts.scrollLockClass);
 			}
 			restoreFocus?.focus();
 		}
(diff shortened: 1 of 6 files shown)


ginderjeremiah (commented):
## Review: Three shared-UI-primitive bug fixes (focus-trap scroll-lock refcount, toast duration collapse, stale tour-callout position) — closes #1864

**Verdict:** Approve — all three fixes match the issue's diagnosis precisely, are minimally scoped, and each ships with a test that exercises the actual failure mode. (Submitted as `COMMENT` only because I'm the PR author; GitHub won't accept an approving event on one's own PR.)

### What's good

- **Scroll-lock refcount** (`focus-trap.ts`) — the `acquireScrollLock`/`releaseScrollLock` pair adds on 0→1 and removes on 1→0, and the `update()` swap path routes through the same refcounted helpers rather than raw `classList`. The `<= 0` guard in `releaseScrollLock` is a sensible defensive floor. Correct fix for the stacked-`Popover` case.
- **Toast duration** (`toast.svelte.ts`) — storing each toast's own `duration` on `ToastData` and reusing it on collapse (rather than the incoming duplicate's) is the right call; it cleanly closes both directions of the bug (persistent→timed and timed→persistent). `verify-frontend` passing confirms the new required field is fully wired at every construction site.
- **Tour callout position** (`TourPlayer.svelte`) — adding `void stepIndex` to the position `$effect` is correct: `anchorEl` is a `$derived` that stays reference-stable across two steps sharing an `anchorKey`, so the effect wouldn't otherwise re-run. Since `$effect` runs after the DOM updates, the recompute reads the freshly-rendered callout height. Mirrors the existing `resizeTick` tracking convention.

### Test coverage

Thorough and appropriate for the scope. Each fix has a unit test targeting its exact failure mode: stacked-trap refcount (only unlocks on last destroy), both toast persistence directions, and the same-anchor callout recompute (with a documented jsdom-layout workaround since jsdom doesn't lay out real box sizes). This is pure frontend UI, not battle logic, so no backend mirroring is required.

Clean tech-debt cleanup. Approving.

---

PR #1979, opened 2026-07-14
Add idempotency backstop for battle-lifecycle session saves (#1874)

### LINKED ISSUE #1874: Idempotency backstop for battle-lifecycle session saves (residual reconnect double-credit gap after #1853)

ginderjeremiah:
## Summary

#1853 made the battle-lifecycle session save (`SessionService.SavePlayerStateAsync`) an **awaited** Redis write instead of fire-and-forget, closing the common case where a normal Redis error silently dropped the session's cleared-battle state. The residual gap it left: a true crash (process kill, OOM, host failure) between the durable battle credit and that now-awaited session save can still strand the session showing an already-credited battle as active — and the next reconnect's battle-end (`BattleService`'s abandon path) would re-credit that same battle.

This closes the gap structurally, per the direction #1874 suggested:

- `Player.LastCreditedBattleSeed` (`uint?`) records the RNG seed of the most recently durably-credited battle. It's set in `Player.RecordBattleOutcome` (shared by `RecordBattleVictory`/`RecordBattleCompleted`) and rides the same `PlayerCoreUpdatedEvent`/write-behind save as the rest of the credit, so it can never durably land without the credit it's meant to protect.
- `BattleService.AbandonBattle` — the path a reconnect's next battle-end actually reaches when the stale session still shows a battle active — now checks the resolved battle's seed against `player.LastCreditedBattleSeed` before crediting anything. A match means this exact battle was already durably credited; the session is cleared without replaying the credit (mirroring the existing "nothing to resolve" branch — no cooldown penalty, since nothing new happened).
- Only the single latest seed is retained: at most one battle can be credited-but-not-yet-session-cleared for a given player at a time, so there's nothing to gain from a history/retention window (one of the open design questions #1874 flagged).
- Threaded through the full persistence path: `Player` (Core) → `PlayerCoreUpdatedEvent` → `PlayerCoreUpdatedHandler` (Postgres column, via a new EF migration) → `PlayerCacheModel`/`PlayerCacheMapper`/`PlayerQueryExtensions` (Redis cache + DB projection, so it survives a cold reload too).
- Documented in `docs/backend-persistence.md` (the section that already described this residual gap).

### Scope note

This protects the abandon-driven reconnect path specifically, matching the issue's own framing. I deliberately did **not** add the same guard inside `EndBattleVictory`/`EndBattleLoss` (the `DefeatEnemy`/`BattleLost` command handlers): those don't clear `PlayerState` before checking, so rejecting an already-credited seed there (treating it as "no active battle") would leave the stale battle stuck active with no way for the client to advance — worse than the bug it would try to fix. The abandon path is the one place a reconnect naturally recovers through (starting the next battle), so that's where the backstop belongs.

## Test plan

- [x] `Game.Core.Tests/Players/PlayerTests.cs` — `RecordBattleCompleted`/`RecordBattleVictory` record the battle seed onto `LastCreditedBattleSeed`, thread it onto `PlayerCoreUpdatedEvent`, and a later credit overwrites the previous seed.
- [x] `Game.Application.Tests/Mapping/PlayerCacheMapperTests.cs` — `LastCreditedBattleSeed` round-trips through `ToCacheModel`/`ToCore` (including the full serialize/deserialize round-trip test).
- [x] `Game.Application.Tests/Services/BattleServiceIntegrationTests.cs` — two new integration tests:
  - a won-and-abandoned battle records its seed as `LastCreditedBattleSeed`.
  - simulating the crash gap directly (crediting happens, but the session's `PlayerState` is never cleared/saved) proves a subsequent abandon does **not** re-grant exp, while still clearing the stale session and starting a fresh battle.
- [x] `dotnet build` — clean.
- [x] `dotnet test` (full solution: Core, Application, Api, Infrastructure) — 3161 passed, 0 failed.

Closes #1874

---
_Generated by [Claude Code](https://claude.ai/code/session_01HQSgYo6ZjcpJZDmkeNmtXn)_


Game.Core/Players/Events/PlayerCoreUpdatedEvent.cs
```diff
@@ -14,5 +14,6 @@ public record PlayerCoreUpdatedEvent(
         int StatPointsGained,
         int StatPointsUsed,
         DateTime LastActivity,
-        bool AutoChallengeBoss) : IDomainEvent;
+        bool AutoChallengeBoss,
+        uint? LastCreditedBattleSeed = null) : IDomainEvent;
```
ginderjeremiah: Nit (optional): `LastCreditedBattleSeed` is the only field on this record with a default — every other field is required-positional. The sole production raiser (`Player.RaiseCoreUpdated`) always passes it explicitly, so the `= null` exists only to spare the ~10 test construction sites. That slightly weakens the "forgot a field ⇒ compile error" guarantee the rest of the record gives, and it runs against the compile-time-required-over-default philosophy the codebase argues for a few lines away in `Player.RecordBattleCompleted` (the ratings split, "makes a victory caller forgetting the ratings a compile error instead of a silent zero-XP accrual").

Worth noting this is *not* the same as the `battleSeed = null` defaults on `RecordBattleCompleted`/`RecordBattleVictory` — those are structurally forced (the param sits before the pre-existing `notify = true`, so it can't be required). This event field has no such constraint and could be made required, updating the test call-sites instead. Take it or leave it — a silently-null seed here only degrades to pre-#1874 behavior (backstop doesn't fire), never a double-credit, so the failure mode is benign.

---

ginderjeremiah (commented):
## Review: Idempotency backstop for battle-lifecycle session saves (#1979 → closes #1874)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating event on my own PR) — the residual reconnect double-credit gap #1853 left is now closed structurally, exactly as #1874 proposed, and the change is correct, complete across the persistence path, and well-tested.

I verified the design end-to-end and it holds up:

- **Guard correctness.** `AbandonBattle`'s `state.BattleSeed is uint seed && seed == player.LastCreditedBattleSeed` sits *after* the "nothing to resolve" early-return and *before* both the still-in-progress hand-back and the outcome branches — the right spot. An already-durably-credited battle is never handed back or re-credited; the stale session just catches up via `ClearBattle`, mirroring the existing no-op branch. A null seed / null field can't produce a false match (the `is uint` pattern requires a non-null seed and the lifted `==` is false against a null field).
- **No false positives in normal play.** A credited battle clears its own `PlayerState` in the same operation, so the next active battle always carries a fresh seed. The only way an active battle's seed equals `LastCreditedBattleSeed` is the crash window #1874 targets — or a seed collision, which is ~1-in-2³² since `CreateBattleSeed()` draws full 32-bit CSPRNG entropy (`RandomNumberGenerator.GetBytes`), not a low-entropy clock value.
- **The marker rides the same durable save as the credit it protects.** `LastCreditedBattleSeed` is set in `RecordBattleOutcome` *before* `StampActivity → RaiseCoreUpdated`, and `RecordBattleOutcome` always stamps activity — so the seed is carried on the `PlayerCoreUpdatedEvent` on every credit including losses, and can't durably land without the credit. Threaded correctly through the whole path: Core → event → `PlayerCoreUpdatedHandler` (new EF column) → `PlayerCacheModel`/`PlayerCacheMapper`/`PlayerQueryExtensions`, so it survives a cold cache reload too.
- **Migration.** `uint?` → nullable `bigint` is the correct EF mapping (uint overflows `integer`); migration and model snapshot agree.
- **Scope call is right.** Leaving the guard out of `EndBattleVictory`/`EndBattleLoss` (they don't clear `PlayerState` before checking, so rejecting a matching seed there would strand the battle active) and putting it only on the abandon path — the one a reconnect naturally recovers through — is the correct placement, and the PR reasons it out explicitly.

### Nits

- `PlayerCoreUpdatedEvent.LastCreditedBattleSeed = null` — the only defaulted field on an otherwise all-required record; inline comment for the detail. Optional, and the failure mode is benign (degrades to pre-#1874 behavior, never a double-credit).

### Test coverage

Thorough and matches the project's rules:
- **Core** (`PlayerTests`): seed recorded onto `LastCreditedBattleSeed` for both completed and victory paths, carried onto `PlayerCoreUpdatedEvent`, and overwritten by a later credit.
- **Mapping** (`PlayerCacheMapperTests`): `ToCore` maps it and it survives the full serialize/deserialize round-trip.
- **Integration** (`BattleServiceIntegrationTests`): a won-and-abandoned battle records its seed; and the crash gap simulated directly (credit happened, session `PlayerState` never cleared) proves the subsequent abandon does *not* re-grant exp while still clearing the stale battle and starting a fresh one. The guard's negative case (non-matching seed still credits) is covered by the pre-existing abandon-credit tests.
- No frontend mirror needed — this is server-side persistence/idempotency, not battle-simulation logic.

### What's good

Careful, minimal, and correctly reasoned. The unconditional set in `RecordBattleOutcome` with `BattleService` as the sole gatekeeper is a clean separation, the retain-only-latest-seed decision is justified (at most one battle can be credited-but-not-session-cleared per player), and `docs/backend-persistence.md` is updated to describe the gap as closed rather than merely narrowed. Nicely done.

---

ginderjeremiah:
## Review — ✅ Approve

Verified against the PR head commit `0b9f77b` directly (not just the description).

**Correctness of the backstop:**
- `Player.LastCreditedBattleSeed` is set in `RecordBattleOutcome` *before* `StampActivity → RaiseCoreUpdated`, so it rides the same `PlayerCoreUpdatedEvent` / write-behind save as the credit — it can't durably land without the credit it protects. ✓
- `AbandonBattle`'s guard (`state.BattleSeed is uint seed && seed == player.LastCreditedBattleSeed`) is placed after the "nothing to resolve" early-return and before the outcome branches — the correct spot. The `is uint` pattern + lifted `==` can't false-match a null field or null seed; no `!` operator used. ✓
- All three credit sites thread `state.BattleSeed` (never null), and they're the only callers of `Player.RecordBattle*`, so a real credit never clobbers the marker with null. ✓
- Seeds come from `RandomNumberGenerator.GetBytes(sizeof(uint))` — full 32-bit CSPRNG entropy, so a false-positive skip needs a ~1-in-2³² collision. `uint?` → nullable `bigint` is the right EF mapping (uint overflows `integer`); migration + snapshot agree. ✓
- Scope call is sound: deliberately *not* guarding `EndBattleVictory`/`EndBattleLoss` is correct — rejecting an already-credited seed there would strand the stale battle active with no client recovery. The abandon path is where a reconnect naturally recovers.

**Test coverage** matches project rules — Core (seed recorded on both paths, carried on the event, overwritten by a later credit), mapping round-trip, and two integration tests including a directly-simulated crash gap proving no exp re-grant while still clearing the stale session. No frontend mirror needed (server-side persistence/idempotency, not battle-simulation logic).

**Prior review:** your self-review's only nit (the `= null` default on the event field) is resolved by `0b9f77b`, which makes it required-positional. Nothing else outstanding.

**CI:** green — `test-backend`, `verify-backend`, all three `test-e2e` browsers success; frontend jobs correctly skipped; `all-checks-passed` success.

One non-blocking observation: the crash-recovery guard clears the stale battle with no post-battle cooldown, so a reconnecting player could start their next battle marginally early. This is intentional and consistent with the "nothing new happened" branch (the cooldown state was lost in the crash) — not worth changing.

No blocking issues, no should-fixes, no follow-ups warranted. The retain-only-latest-seed decision correctly resolves #1874's retention-window design question. Merging.

---

Issue #1980, opened 2026-07-14
Add a periodic reconciliation sweep backstop for reference-cache invalidation broadcasts


ginderjeremiah:
## Context

Follow-up to #1888, split out to keep that fix contained. #1888's problem statement calls out two distinct gaps in the reference-data invalidation broadcast:

1. `RedisPubSubService.Publish` used `CommandFlags.FireAndForget`, so a genuine send failure (e.g. the connection is down) was silently swallowed instead of surfacing at the admin-write site. **Fixed by #1888** — the publish is now awaited (not fire-and-forget), so a real send failure throws and is logged by `AdminCacheReloadFilter`'s existing catch block.
2. Even with a fully successful, ack'd publish, Redis pub/sub delivery to subscribers is inherently **at-most-once**: a subscriber instance that is mid-reconnect (or otherwise transiently disconnected) when the message is published simply never receives it, with **no error, no retry, and no reconciliation** anywhere in the pipeline. `CoalescingReferenceCacheReloader` only ever reloads in reaction to a received signal — it has no notion of "I haven't reloaded in a while, let me check anyway."

Gap 2 is **not addressed** by #1888's fix and is arguably the more likely real-world failure mode (a broker/connection blip during a rolling deploy or a Redis failover is more probable than an outright publish command failure). Its impact matches the original issue's framing exactly: an affected instance serves stale reference data indefinitely, until its *own* next admin write or a restart — with zero signal that anything was missed.

## Why this is separate

Fixing gap 2 requires introducing a new mechanism this codebase doesn't have any precedent for: a periodic/time-based background sweep (the codebase currently has no `PeriodicTimer`/`BackgroundService`-style periodic job anywhere — `CoalescingReferenceCacheReloader` is purely signal-driven). That means real design questions #1888's narrower fix didn't need to answer:

- **Interval.** How stale is acceptable before a bounded self-heal kicks in? Too short adds needless recurring DB load across every instance, forever; too long defeats the point.
- **Always-on vs. adaptive.** A flat periodic reload on every instance regardless of activity is the simplest option, but means paying a full reference-cache reload query on a cadence even when nothing has changed in days.
- **Where it lives.** Likely `CoalescingReferenceCacheReloader.RunAsync`, racing its existing signal-wait against a periodic delay — but this changes the loop's shape and its "reload only on signal" invariant that `CoalescingReferenceCacheReloaderTests` currently pins.
[...]

PR #1981, opened 2026-07-14
Stop swallowing reference-data invalidation publish failures (#1888)

### LINKED ISSUE #1888: Cross-instance reference-data invalidation broadcast is fire-and-forget with no recovery — a lost publish leaves other instances stale indefinitely
## Problem

`ReferenceCacheSynchronizer.NotifyChangedAsync` (`Game.DataAccess/ReferenceCacheSynchronizer.cs:50-53`) publishes the invalidation via `RedisPubSubService.Publish`, which uses `CommandFlags.FireAndForget` (`Game.Infrastructure/PubSub/Redis/RedisPubSubService.cs:36-39`). The F&amp;F justification in that comment holds for the write-behind *wake* signal (the durable enqueue exists regardless, and startup drains the queue) — but the same overload carries the reference-data invalidation, which has **no durable backing and no reconciliation**:

- `CoalescingReferenceCacheReloader` sweeps only on received signals; its retry-with-backoff covers a failed *sweep*, never a lost *broadcast*.
- With F&amp;F, StackExchange.Redis swallows a publish failure silently, and pub/sub delivery is at-most-once anyway (a subscriber mid-reconnect misses the message).

Result: after an admin write, every other instance — including its memoized `GetReferenceDataVersions` hashes — serves stale reference data until the *next* admin write or a restart, with zero signal that anything was missed. Latent today (effectively single-instance), but the architecture explicitly targets multiple instances.

## Suggested fix

Await the publish non-F&amp;F in `NotifyChangedAsync` so a failure surfaces/logs at the admin-write site (where a retry is possible), and/or add a periodic reconciliation sweep as a backstop.

Related: #1708 is about admin reload *amplification* (cost); nothing tracks broadcast-loss staleness. #1825 touches the same pub/sub service but a different concern.

ginderjeremiah:
## Summary

`ReferenceCacheSynchronizer.NotifyChangedAsync` broadcasts the cross-instance reference-data invalidation via `RedisPubSubService.Publish(channel, message, ct)`, which sent with `CommandFlags.FireAndForget`. That flag is the right call for the write-behind wake signal (a durable queue write backs it regardless), but this same overload also carries the reference-data invalidation, which has **no durable backing and no reconciliation** — with `FireAndForget`, StackExchange.Redis swallows a genuine send failure silently, so `AdminCacheReloadFilter`'s existing "broadcast failure" catch block (which logs a warning) had nothing real to catch.

- `RedisPubSubService.Publish(channel, message, ct)` no longer sends with `CommandFlags.FireAndForget` — it's awaited so a real send failure now throws and is caught/logged at the admin-write site (`AdminCacheReloadFilter`), where a retry is possible. The sibling `Wake` overload (used by the write-behind path) is untouched — its fire-and-forget is still correct since a durable queue write backs it.
- Updated `IReferenceDataChangeNotifier`'s doc comment (it previously described delivery itself as "fire-and-forget", which was conflating "the publish command doesn't wait for an ack" with "delivery to subscribers is at-most-once" — only the latter is still true).
- Documented the change (and its residual gap) in `docs/backend.md`'s reference-data cache reload section.

### Residual gap (filed separately as #1980)

Even with an awaited, successfully-ack'd publish, Redis pub/sub delivery to subscribers is inherently **at-most-once**: an instance that's mid-reconnect when the message is published still misses it, with no error and no self-heal until its own next admin write or a restart. Fixing that needs a genuinely new periodic-reconciliation-sweep mechanism (this codebase has no `PeriodicTimer`/`BackgroundService`-style periodic job anywhere yet), plus real design calls (interval, always-on vs. adaptive, doc updates, test updates to `CoalescingReferenceCacheReloaderTests`'s signal-only invariant). That's a distinct design decision from this contained bug fix, so I split it out as [#1980](https://github.com/ginderjeremiah/GameServer/issues/1980) rather than scope-creeping this PR.

## Test plan

- [x] `Game.Infrastructure.Tests/RedisPubSubPublishFailureTests.cs` (new) — pins the fix directly: against a dead-endpoint multiplexer (same technique as the sibling `RedisPubSubSubscribeRollbackTests`, no mocking, no Redis server needed), `Publish` now throws `RedisConnectionException` instead of completing successfully despite the unreachable connection.
- [x] Existing `ReferenceCacheSynchronizerTests`, `RedisPubSubServiceIntegrationTests`, `AdminCacheReloadBroadcastTests`, and `AdminCacheReloadFilterTests` all still pass against real Redis — confirming no regression to the happy-path delivery/ordering/debounce behavior.
- [x] `dotnet test` (full solution) — 3155 passed, 0 failed.

Closes #1888

---
_Generated by [Claude Code](https://claude.ai/code/session_0135zEXBphvD2UCLTtQ2YNBK)_


ginderjeremiah (commented):
## Review: Stop swallowing reference-data invalidation publish failures (#1888)

**Verdict:** Approve _(submitted as a comment since I'm the PR author; GitHub won't accept a gating event on one's own PR)_ — the change is a correct, minimal fix for gap 1 of #1888 with a test that pins it directly, accurate docs, and the harder residual gap cleanly split to #1980.

I traced the change end-to-end against the issue and the surrounding code:

- The single-message `Publish(channel, message, ct)` overload is used **only** by `ReferenceCacheSynchronizer.NotifyChangedAsync` (`Game.DataAccess/ReferenceCacheSynchronizer.cs:52`). The durable-backed paths (`SocketManagerService`, the write-behind batch publishes) all go through the `queueName` overloads → `Wake`, which is untouched and still fire-and-forget. So dropping `CommandFlags.FireAndForget` here is correctly scoped and can't affect the write-behind wake signals.
- The failure now surfaces where it can be acted on: `AdminCacheReloadFilter` (`Game.Api/Filters/AdminCacheReloadFilter.cs:59-66`) already wrapped `NotifyChangedAsync()` in a try/catch that logs a warning — that catch was previously dead (F&F swallowed the throw); it's now live and does exactly what its comment describes. No behavior change on the success path (a successful publish still just returns).
- `.WaitAsync(cancellationToken)` is preserved, so prompt cancellation is unaffected.

### Blocking

- None.

### Should-fix

- None.

### Nits

- None.

### Test coverage

- `RedisPubSubPublishFailureTests` pins the fix directly and deterministically — against a dead-endpoint multiplexer it asserts `Publish` now throws `RedisConnectionException` instead of completing. It reuses the exact in-process, no-Redis-server technique of the sibling `RedisPubSubSubscribeRollbackTests`, matching the project's "genuine dependency failure" test style, and the referenced `<see cref>` targets resolve.
- The happy-path delivery/ordering/debounce behavior stays covered by the existing `ReferenceCacheSynchronizerTests` / `AdminCacheReloadBroadcastTests` / integration suites. Adequate for an infrastructure behavior change of this size.

### What's good

- Correct diagnosis and a genuinely minimal fix — one flag removed, no over-reach into the durable paths.
- The doc-comment correction on `IReferenceDataChangeNotifier` is a nice touch: it disentangles "the publish command doesn't wait for an ack" from "delivery is at-most-once," which is the exact conflation that made the original F&F look justified.
- The residual at-most-once delivery gap is honestly documented (both in `backend.md` and the interface comment) and split out to #1980 rather than scope-crept — that follow-up is a real design decision (introduces the codebase's first periodic background sweep), so deferring it is the right call.


---

PR #1984, opened 2026-07-15
Wrap each multiplexer dispose in DisposeAllAsync so one faulting close doesn't abort the rest

### LINKED ISSUE #1889: RedisMultiplexerFactory.DisposeAllAsync aborts all remaining disposals on one faulting close, contradicting its own contract
## Problem

`RedisMultiplexerFactory.DisposeAllAsync` documents "Each connection is closed independently so one faulting close still lets the rest tear down" (`Game.Infrastructure/Redis/RedisMultiplexerFactory.cs:50`) but the implementation is a bare `foreach { await disposable.DisposeAsync(); }` with no per-item catch (`:73-76`). The cache dictionary is cleared **before** the loop, so when the first `DisposeAsync` throws (e.g. a wedged connection close):

- the remaining multiplexers are skipped and permanently unreachable (cache already cleared), and
- the exception escapes `RedisConnectionLifetime.StopAsync` (`RedisConnectionLifetime.cs:32-39` catches only `OperationCanceledException`).

In production this only degrades graceful shutdown, but `ResetForTesting()` (`:104-107`) delegates to the same method between integration-test runs, where a faulting close breaks subsequent teardown/reconnect.

## Suggested fix

Wrap each `DisposeAsync` in try/catch (log, continue) so the loop honors the documented contract. Add a test with a fake disposable that throws (`RedisMultiplexerFactoryTests.cs` currently covers only the happy path and empty cache).

ginderjeremiah:
## Summary

`RedisMultiplexerFactory.DisposeAllAsync` documents that "each connection is closed independently so one faulting close still lets the rest tear down," but the implementation was a bare `foreach { await disposable.DisposeAsync(); }` with no per-item try/catch. Since the cache dictionary is cleared *before* the loop runs, a faulting close on the first entry left every subsequent multiplexer permanently unreachable, and the exception escaped `RedisConnectionLifetime.StopAsync` (which only catches `OperationCanceledException`). `ResetForTesting()` delegates to the same method between integration-test runs, so a faulting close there could break subsequent test teardown/reconnect too.

## Changes

- `RedisMultiplexerFactory.DisposeAllAsync<T>` now wraps each `DisposeAsync()` call in its own try/catch, logging the failure and continuing to the next entry — matching the method's own documented contract.
- Both the public `DisposeAllAsync` and the generic seam now take an `ILogger`, following this codebase's existing pattern of passing an `ILogger` into static Redis helpers (see `RedisCommandBudget.Write`) rather than giving the static factory its own DI-resolved logger.
- `RedisConnectionLifetime` now takes an injected `ILogger<RedisConnectionLifetime>` (constructor-injected via its existing `IHostedService` DI registration) and hands it down to `DisposeAllAsync`.
- `ResetForTesting()` passes `NullLogger.Instance`, since test resets don't need to observe the log.
- Added a test (`DisposeAllAsync_OneEntryFaultsOnDispose_StillDisposesTheRestAndClearsTheCache`) with a fake disposable that throws, pinning: the faulting entry is caught and logged as an error, the healthy sibling is still disposed, and the cache is still cleared. Updated the two existing `DisposeAllAsync` tests to pass a logger.

## Test plan

- [x] `dotnet build` (full solution) — clean
- [x] `dotnet test` (full backend suite: Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests) — all green, including the new fault-tolerance test and the DI-resolution of `RedisConnectionLifetime`'s new constructor parameter via `Game.Api.Tests`' integration tests

Closes #1889

---
_Generated by [Claude Code](https://claude.ai/code/session_01McTLuKnmTQVPqZnDPadaKu)_

### DIFF
--- a/Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
+++ b/Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
@@ -171,41 +171,5 @@ namespace Game.Infrastructure.Tests
                 return ValueTask.CompletedTask;
             }
         }
-
-        // A minimal capturing logger mirroring RedisCommandBudgetTests' CapturingLogger: the dispose-loop's only
-        // observable effect on a faulting entry is the error log, so the level and exception presence are recorded.
-        private sealed class CapturingLogger : ILogger
-        {
-            private readonly object _gate = new();
-            private readonly List<(LogLevel Level, Exception? Exception)> _entries = [];
-
-            public IReadOnlyList<(LogLevel Level, Exception? Exception)> Entries
-            {
-                get
-                {
-                    lock (_gate)
-                    {
-                        return _entries.ToList();
-                    }
-                }
-            }
-
-            public IDisposable BeginScope<TState>(TState state) where TState : notnull => NullScope.Instance;
-            public bool IsEnabled(LogLevel logLevel) => true;
-
-            public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
-            {
-                lock (_gate)
-                {
-                    _entries.Add((logLevel, exception));
-                }
-            }
-
-            private sealed class NullScope : IDisposable
-            {
-                public static readonly NullScope Instance = new();
-                public void Dispose() { }
-            }
-        }
     }
 }
--- a/Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
+++ b/Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
@@ -1,4 +1,6 @@
 using Game.Infrastructure.Redis;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Xunit;
 
 namespace Game.Infrastructure.Tests
@@ -111,7 +113,7 @@ namespace Game.Infrastructure.Tests
             cache["cache:6379"] = first;
             cache["pubsub:6380"] = second;
 
-            await RedisMultiplexerFactory.DisposeAllAsync(cache, syncRoot);
+            await RedisMultiplexerFactory.DisposeAllAsync(cache, syncRoot, NullLogger.Instance);
 
             // Every cached connection is disposed and the cache is emptied so a later request reconnects fresh —
             // the graceful-shutdown teardown the hosted service drives.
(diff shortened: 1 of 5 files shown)


Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
```diff
@@ -125,20 +127,85 @@ public async Task DisposeAllAsync_EmptyCache_IsANoOp()
...
+                    throw new InvalidOperationException("redis close wedged");
+                }
+
                 return ValueTask.CompletedTask;
             }
         }
+
+        // A minimal capturing logger mirroring RedisCommandBudgetTests' CapturingLogger: the dispose-loop's only
+        // observable effect on a faulting entry is the error log, so the level and exception presence are recorded.
+        private sealed class CapturingLogger : ILogger
```
ginderjeremiah: **Should-fix (DRY):** This `CapturingLogger` (and its nested `NullScope`) is a byte-for-byte copy of the one in `RedisCommandBudgetTests.cs`, both in `Game.Infrastructure.Tests`. CLAUDE.md is explicit about preferring DRY, and the comment here even acknowledges it "mirrors RedisCommandBudgetTests' CapturingLogger". Since this PR is what introduces the second copy, it's in-scope to extract a single shared test double (e.g. an `internal sealed class CapturingLogger` in its own file in this test project) and have both test classes use it, rather than maintaining two identical implementations. Not a blocker, but worth doing here rather than deferring.

---

ginderjeremiah (commented):
## Review: DisposeAllAsync per-entry fault isolation (#1889)

**Verdict:** Approve (posted as a comment since I'm the PR author) — one should-fix, no blockers.

The fix is correct and squarely addresses the issue. The old `foreach { await DisposeAsync(); }` cleared the cache *before* the loop, so one faulting close orphaned every remaining multiplexer and escaped `RedisConnectionLifetime.StopAsync` (which only catches `OperationCanceledException`). Wrapping each dispose in its own try/catch (log + continue) now honors the method's documented "each connection is closed independently" contract, and `ResetForTesting()` inherits the same resilience for integration-test teardown.

The `ILogger` threading is clean and follows the codebase's established pattern of passing an `ILogger` into static Redis helpers (`RedisCommandBudget.Write`) rather than giving the static factory a DI-resolved logger. `RedisConnectionLifetime`'s new public ctor param resolves fine through its existing `IHostedService` registration, and `ResetForTesting()` correctly uses `NullLogger.Instance`.

### Should-fix

- `Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs:177` — the new `CapturingLogger` (+ nested `NullScope`) is a byte-for-byte duplicate of the one in `RedisCommandBudgetTests.cs`, same test project. DRY; extract one shared test double. Details inline.

### Test coverage

Good. `DisposeAllAsync_OneEntryFaultsOnDispose_StillDisposesTheRestAndClearsTheCache` pins exactly the regression: the faulting entry is disposed, the healthy sibling is still disposed, the cache is cleared, and a single error-level log with an exception is recorded. By awaiting `DisposeAllAsync` without expecting a throw, it also implicitly pins that the fault no longer escapes the caller. The two existing happy-path/empty-cache tests were updated for the new signature. This is the right layer — `Game.Infrastructure.Tests` is for in-process logic with no live connection, and the DI resolution of the new ctor param is exercised by the `Game.Api.Tests` integration path.

### What's good

- The fix matches the documented contract rather than just silencing the symptom, and the XML doc was updated to describe the per-entry isolation.
- Reusing the generic seam means the fault-tolerance is unit-tested without a live Redis connection, consistent with how `GetOrAdd` is tested.

---

ginderjeremiah:
## Re-review: incremental change since last review

**Verdict:** Approve — merging.

The only change since the prior review is commit `60eaafc9`, which addresses that review's single should-fix. The duplicated `CapturingLogger` (and its nested `NullScope`) was extracted into a shared `Game.Infrastructure.Tests/CapturingLogger.cs` and both `RedisCommandBudgetTests` and `RedisMultiplexerFactoryTests` now consume the single `internal sealed` test double. The extraction is a faithful move — no behavioral change to the capture logic — and the new XML doc generalizes the intent across both suites. Dropping the old `ExecuteSynchronously` comment was correct: that note described the `RedisCommandBudget` fault-continuation scenario, not the logger itself.

Everything from the original review still stands: the per-entry try/catch in `DisposeAllAsync` honors the documented "each connection is closed independently" contract, the `ILogger` threading follows the codebase's static-helper pattern, and `ResetForTesting()` correctly uses `NullLogger.Instance`.

- CI: all green (`test-backend`, `verify-backend`, `test-e2e` chromium/webkit/firefox all pass; frontend jobs skipped — no frontend changes).
- The resolved inline should-fix thread is confirmed addressed and outdated.
- No remaining nits or should-fixes; no follow-up work required.

Merging.

---

PR #1985, opened 2026-07-15
Add UnarchiveUser/UnbanUser admin endpoints

### LINKED ISSUE #1890: Ban and Archive are one-way — no unban/unarchive exists anywhere in the system
## Problem

`AdminUsersController` exposes only `ArchiveUser`/`BanUser` (`Game.Api/Controllers/Admin/AdminUsersController.cs:78-90`), and the `IUsers` repository contract ends at the same two methods (`Game.Abstractions/DataAccess/IUsers.cs:154-168`) — no method anywhere clears `ArchivedAt`/`BannedAt` (repo-wide search for Unban/Unarchive/Restore/Reinstate over backend + UI returns nothing).

The design docs frame these as recoverable states: `docs/backend-admin.md` calls them "two distinct, **non-destructive** states" and says the roster's tri-state `archived` filter exists "so an admin can still find and **manage** them" — and the content-lifecycle analogue (Retire) has an explicit Reinstate. Today a mistaken ban or archive is only reversible by hand-editing the database.

## Design decision needed

Unarchive has one real wrinkle: archiving **frees the username** (partial unique index), so the name may have been reused by the time of an unarchive. The collision rule (refuse? force rename?) is the one design call; unban has no such issue.

## Suggested fix

Two endpoints + two `IUsers` methods reusing the existing `UserActionStatus` plumbing, with the unarchive username-collision rule decided and tested.

Related: #1762 covers ban *enforcement* (live sessions surviving a ban); this is the missing reverse of the lifecycle actions.

ginderjeremiah:
## Summary

Ban and Archive were one-way — `AdminUsersController` exposed only `ArchiveUser`/`BanUser`, and nothing anywhere cleared `ArchivedAt`/`BannedAt`, so a mistaken lifecycle action was only reversible by hand-editing the database.

- `IUsers`/`Users` gain `UnarchiveUser` and `UnbanUser`, reusing the existing `UserActionStatus` plumbing (adds one new case, `UsernameTaken`).
- `AdminUsersController` gains matching `UnarchiveUser`/`UnbanUser` endpoints.
- Regenerated the TypeScript client (`api-type-map.ts`) for the two new endpoints via `dotnet run --project Game.Api -- codegen`.
- Documented the design decisions in `docs/backend-admin.md` under *User Administration (Archive vs Ban)*.

## Design decisions (the issue flagged one; I made a second while implementing)

1. **Username-collision on unarchive** (the design call the issue explicitly asked for): archiving frees the username, so another active account may have claimed it by the time of an unarchive. Rather than silently renaming either side, `UnarchiveUser` rejects the request with `UsernameTaken` and leaves the collision for the admin to resolve manually — checked up front, and re-checked via a unique-violation catch around an in-tier commit (mirroring `CreateAccount`'s existing shape) so a concurrent claim landing mid-request still surfaces as a clean rejection rather than a 500.
2. **Self-target guard on reinstate, even though it carries no lockout hazard.** Reinstating never shrinks the usable-admin pool, so I did *not* add the last-admin check the issue's "reuse `UserActionStatus`" framing might suggest. But I did keep the self-target rejection: bans/archives are enforced only at login and never re-checked against an already-issued access token (confirmed by reading the JWT pipeline — no `OnTokenValidated` DB check, `SessionService.Authenticated` is derived purely from the validated token). Without the guard, a still-valid session belonging to a just-banned/archived admin could self-reverse the action against themselves before their token naturally expires (up to 15 min), defeating the point of another admin acting against them. This wasn't called out in the issue, so flagging it here explicitly — it's a small, contained addition (two lines reusing `AdminLockoutPolicy.IsSelfTarget`) and I'm confident it's the right default, but happy to discuss if the intent was for reinstate to be fully unguarded (matching the content-lifecycle Retire→Reinstate precedent, which is deliberately unguarded).

## Scope note

No frontend UI exists yet for the user roster/archive/ban actions (verified — only codegen'd API types and a generic request test reference these endpoints, no admin page consumes them). The issue's own suggested fix was backend-only ("Two endpoints + two `IUsers` methods"), so this PR is backend-only too.

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test` (full backend suite: Core, Infrastructure, Application, Api) — all green, including new integration tests:
  - `UnarchiveUser_RestoresActiveState`
  - `UnarchiveUser_UsernameClaimedByAnotherActiveAccount_ReturnsErrorAndDoesNotUnarchive`
  - `UnarchiveUser_TargetingSelf_IsRejectedAndDoesNotUnarchive`
  - `UnarchiveUser_UnknownUser_ReturnsError`
  - `UnarchiveUser_AlreadyActive_IsANoOpSuccess`
  - `UnarchiveUser_ConcurrentUnarchiveOfTwoAccountsSharingAnArchivedUsername_LeavesExactlyOneActive`
  - `UnbanUser_ClearsBannedState`
  - `UnbanUser_TargetingSelf_IsRejectedAndDoesNotUnban`
  - `UnbanUser_UnknownUser_ReturnsError`
- [x] TypeScript codegen regenerated and diff reviewed (only the two new endpoint entries)

Closes #1890

---
_Generated by [Claude Code](https://claude.ai/code/session_01EX2A4zsFmEZ4iXT45fFb6X)_


ginderjeremiah (commented):
## Review: Add UnarchiveUser/UnbanUser admin endpoints (closes #1890)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating APPROVE from me) — the change is correct, well-layered, and thoroughly tested. No blocking findings.

### What's good

- **The username-collision design call is right and well-executed.** Refusing the unarchive with `UsernameTaken` and leaving the collision for the admin to resolve (rather than silently renaming either side) is the sensible choice, and it's guarded the same double-way as `CreateAccount`: an up-front `AnyAsync` check for the common case plus a unique-violation catch around an in-tier commit for the race. The `UnarchiveUser_Concurrent...` test pins that exactly one of two accounts sharing an archived username wins — that's the case that matters.
- **The self-target guard on reinstate is a genuinely good catch, not gold-plating.** The reasoning (bans/archives are enforced only at login, never re-checked against a live token, so a just-actioned admin could self-reverse before their token expires) is sound and worth the two lines. Correctly *omits* the last-admin check, since reinstating can never shrink the usable-admin pool.
- **Layering is clean.** Self-target reuses the `AdminLockoutPolicy.IsSelfTarget` Core primitive (same as the existing archive/ban early check), the controller only maps status→response, and `UnbanUser` stays on the deferred unit-of-work path — matching the non-admin archive/ban path — while `UnarchiveUser` commits in-tier only because it needs the unique-violation catch. Both asymmetries are intentional and documented.
- Docs (`backend-admin.md`) and the `IUsers` XML docs are updated to match, and the TS client is codegen output (only the two sorted entries added), not hand-edited.

### Test coverage

Strong. Integration tests are the correct layer here since the logic leans on the partial unique index and real concurrency (Postgres). Restore, username-collision, self-target, unknown-user, already-active no-op, and the concurrent-race case are all covered for unarchive; clear/self-target/unknown for unban.

### Nits (optional)

- No test mirrors `UnarchiveUser_AlreadyActive_IsANoOpSuccess` for unban (an already-unbanned target returning `Success`). It's a weaker gap than the unarchive equivalent — `UnbanUser` has no distinct short-circuit branch, it just clears unconditionally — so this is purely for symmetry. Take it or leave it.

Not merging automatically: `test-backend` is still in progress (7/8 checks green), and the nit above, though optional, means this isn't a zero-comment auto-merge. Once backend CI is green this is good to merge as-is; the nit doesn't need to block it.

---

PR #1986, opened 2026-07-15
Add reflection-based architecture guard for admin controller attributes

### LINKED ISSUE #1891: Admin-role gating is per-controller convention with no architecture test — a forgotten attribute silently exposes an admin surface to any player
## Problem

Admin authorization is an opt-in `IAuthorizationFilter` each controller must remember to attach: 16 controllers individually carry `[ServiceFilter(typeof(AdminRoleAuthorizationFilter))]` (all of `Controllers/Admin/*` plus `TagsController`). The global fallback policy requires only *authentication*, not any role (`Game.Api/Startup.cs:354-359`). So a future admin controller (or an action moved to a new controller) that omits the attribute is silently reachable by **any logged-in player** — a real hazard given how fast this surface grows (`AdminLessons`, `AdminClasses`, `AdminSkillRecipes`, `AdminDeadLetters`, `AdminContentHealth` were all added incrementally).

Existing tests (`AdminAuthorizationTests.cs` and per-controller 403 checks) cover representative endpoints only; nothing enumerates controllers to enforce the invariant — even though the project already uses exactly this guard style in `Game.Api.Tests/Architecture/EntityIsolationTests.cs`.

## Suggested fix

One reflection-based architecture test asserting:

- every controller whose route lives under the admin surface (`/api/AdminTools`-prefixed routes, plus `TagsController`) carries `ServiceFilterAttribute(typeof(AdminRoleAuthorizationFilter))`, and
- every content-writing admin controller carries `[ReloadReferenceCaches]` (the sibling stale-cache footgun, same forgotten-attribute failure mode).

ginderjeremiah:
## Summary

Admin-role gating (`[ServiceFilter(typeof(AdminRoleAuthorizationFilter))]`) and reference-cache reload (`[ReloadReferenceCaches]`) are opt-in attributes each admin controller must remember to add — there was no test enforcing either invariant, so a future admin controller (or an action moved to a new one) that omits the role filter would be silently reachable by any authenticated player, and one that omits the cache-reload attribute would serve stale reference data after a write.

Added `Game.Api.Tests/Architecture/AdminControllerGuardTests.cs`, following the same reflection-based pattern as the existing `EntityIsolationTests`:

- **`AdminSurfaceController_RequiresAdminRole`** — enumerates every controller in the `Game.Api` assembly whose route is prefixed `/api/AdminTools`, plus `TagsController` (which lives under `/api/[controller]/[action]` but still requires the Admin role), and asserts each carries `[ServiceFilter(typeof(AdminRoleAuthorizationFilter))]`.
- **`ContentWritingAdminController_ReloadsReferenceCaches`** — for the same admin-surface controllers, asserts any controller with at least one `HttpPost`/`HttpPut`/`HttpDelete`/`HttpPatch` action also carries `[ReloadReferenceCaches]`, unless it's one of the two controllers explicitly documented (in their own class remarks) as not owning a reference cache to reload: `AdminUsersController` (user accounts aren't list-cached) and `AdminDeadLettersController` (touches no reference cache). Read-only admin controllers (`AdminContentHealthController`, `TagsController`) are naturally excluded since they have no write actions.

Both tests enumerate the assembly rather than a fixed list, so a controller added later is checked automatically instead of only when someone remembers to add it here — closing the exact gap the issue describes.

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test` (full backend suite, `Game.Api.Tests` included) — 791/791 passed, including the two new architecture tests exercised against every current admin controller
- [x] `dotnet format --verify-no-changes` on the new file — clean

Closes #1891

---
_Generated by [Claude Code](https://claude.ai/code/session_01HpxhZzCQp2fcmt3AqVTM1n)_

### DIFF
--- a/Game.Api.Tests/Architecture/AdminControllerGuardTests.cs
+++ b/Game.Api.Tests/Architecture/AdminControllerGuardTests.cs
@@ -69,6 +69,8 @@ namespace Game.Api.Tests.Architecture
 
         private static bool IsAdminSurfaceController(Type type)
         {
+            // Off-convention admin controllers (ones that don't route under /api/AdminTools) must be
+            // special-cased here too, or they silently fall outside this guard's coverage.
             if (type == typeof(TagsController))
             {
                 return true;
--- a/Game.Api.Tests/Architecture/AdminControllerGuardTests.cs
+++ b/Game.Api.Tests/Architecture/AdminControllerGuardTests.cs
@@ -1,0 +1,97 @@
+using System.Reflection;
+using Game.Api.Controllers;
+using Game.Api.Controllers.Admin;
+using Game.Api.Filters;
+using Microsoft.AspNetCore.Mvc;
+using Xunit;
+
+namespace Game.Api.Tests.Architecture
+{
+    /// <summary>
+    /// Architecture guard for the admin surface (#1891). Both <see cref="AdminRoleAuthorizationFilter"/> and
+    /// <see cref="ReloadReferenceCachesAttribute"/> are opt-in per controller rather than applied by convention,
+    /// so a new (or moved) admin controller that forgets one silently exposes an admin action to any
+    /// authenticated player, or serves stale reference data after a write. This enumerates every controller in
+    /// the assembly rather than a fixed list, so a controller added later is checked automatically instead of
+    /// only when someone remembers to add it here.
+    /// </summary>
+    public class AdminControllerGuardTests
+    {
+        /// <summary>
+        /// Content-writing admin controllers documented as owning no reference cache to reload: user accounts
+        /// and the write-behind dead-letter queues aren't list-cached in memory (see their own class remarks).
+        /// </summary>
+        private static readonly HashSet<Type> ReloadCacheExemptControllers =
+        [
+            typeof(AdminUsersController),
+            typeof(AdminDeadLettersController),
+        ];
+
+        public static IEnumerable<object[]> AdminSurfaceControllers()
+        {
+            return typeof(Startup).Assembly
+                .GetTypes()
+                .Where(IsAdminSurfaceController)
+                .Select(type => new object[] { type });
+        }
+
+        [Theory]
+        [MemberData(nameof(AdminSurfaceControllers))]
+        public void AdminSurfaceController_RequiresAdminRole(Type controllerType)
+        {
+            Assert.True(
+                HasServiceFilter(controllerType, typeof(AdminRoleAuthorizationFilter)),
+                $"{controllerType.Name} sits on the admin surface but is missing [ServiceFilter(typeof(AdminRoleAuthorizationFilter))].");
+        }
+
+        [Theory]
+        [MemberData(nameof(AdminSurfaceControllers))]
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: reflection-based architecture guard for admin controller attributes (#1891)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating review from me) — the test correctly and automatically enforces both invariants the issue asked for, mirrors the existing `EntityIsolationTests` pattern, and I verified it against every current admin controller.

### What I verified
- **Enumeration is complete and correct.** `AdminSurfaceControllers()` picks up exactly the 16 controllers the issue names: the 15 `/api/AdminTools`-routed controllers in `Controllers/Admin/` plus the special-cased `TagsController` (which lives under `/api/[controller]` but still carries the role filter). All 16 currently carry `[ServiceFilter(typeof(AdminRoleAuthorizationFilter))]`, so the role test is green against the live surface.
- **The reload check is right.** `AdminUsersController` and `AdminDeadLettersController` do write (`[HttpPost]`) but genuinely own no reference cache — each documents that in its class remarks — so exempting them is correct. `AdminContentHealthController` and `TagsController` are read-only, so they're excluded naturally by the `writesContent` check with no special-casing. The three grep hits for `ReloadReferenceCaches` on the "exempt" controllers are doc-comments explaining the absence, not stray attributes.
- **No false match on the role filter.** `ReloadReferenceCachesAttribute` extends `ServiceFilterAttribute` but wraps `AdminCacheReloadFilter`, so `HasServiceFilter(..., AdminRoleAuthorizationFilter)`'s `ServiceType` comparison never mistakes one for the other.

- **Coverage is bounded by the route convention.** "Admin surface" is defined as `/api/AdminTools`-prefixed *or* the hardcoded `TagsController`. A future admin controller that both forgets the filter *and* doesn't follow the route convention would slip through — which is exactly the situation `TagsController` represents (it had to be listed by hand). This is inherent to reflection-based detection and the issue endorsed this approach, so it's fine as-is; just worth being aware the guard is only as strong as the convention. A one-line comment near the `TagsController` special-case noting "off-convention admin controllers must be added here too" would make that contract explicit for the next person.
- **`DeclaredOnly` in the write-method scan** means a write action inherited from a shared base controller wouldn't be counted toward `writesContent`. No controller does this today, so there's no live gap — just a latent one if a base class ever grows a write action.

### Test coverage
This PR *is* test coverage — a new architecture guard, no production code. It follows the project's reflection-based guard style and uses `[Theory]`/`[MemberData]` so each controller is an independent case. Appropriate and sufficient.

### What's good
Enumerating the assembly rather than a fixed list is the right call — it closes the exact "forgotten attribute on a later-added controller" gap the issue describes instead of just documenting today's controllers. The exemption set is small, explicit, and self-correcting (the `typeof(...)` references won't compile if a controller is deleted), and pairing it with the required class-remark documentation keeps the "why exempt" reasoning next to the code. Clear XML summaries throughout.

Holding off on merging only because `test-backend` was still in progress at review time and there are the two minor nits above — neither needs a code change to approve.

---

PR #1988, opened 2026-07-15
Treat added-with-empty-collection as unchanged in childChanged

### LINKED ISSUE #1895: Every newly-added Workbench record posts all full-collection child savers with empty payloads — multiplying #1708's cache-reload amplification for no-op writes
## Problem

`childChanged(current, baseline)` returns `true` whenever `baseline === undefined` — i.e. for **every** added record, even when the collection is empty (`UI/src/routes/admin/workbench/save-helpers.ts:36-38,158-171`). Saving one brand-new enemy with only a name therefore posts `SetEnemyAttributeDistributions {[]}`, `SetEnemySkills {[]}`, and `SetEnemySpawns {[]}` — three no-op HTTP writes, each triggering the all-sets reference-cache rebuild + cross-instance broadcast that #1708 documents. The same pattern is in `zone.ts`, `class.ts`, `skill-recipe.ts`, `lesson.ts`.

The Progression store is worse: a new path (which auto-creates a tier) posts `SetProficiencyModifiers {[]}`, `SetProficiencyRewards {[]}`, and `SetProficiencyPrerequisites [{prerequisiteIds: []}]` on top of the two identity batches (`progression-store.svelte.ts:420-427,525-559`).

This is also an internal inconsistency: the diff-shaped savers (skill portions/multipliers/effects, item attributes/slots) already no-op correctly on empty adds — only the "Set whole collection" savers over-post.

## Suggested fix

Treat added-with-empty-collection as unchanged in `childChanged` (or per-saver), so a new record only posts the child writes that actually carry data.

Related: #1708 covers the *backend* per-write reload amplification and proposes coalescing there; this is the complementary client-side fix that stands even if #1708 lands.

ginderjeremiah:
## Summary

`childChanged(current, baseline)` (`UI/src/routes/admin/workbench/save-helpers.ts`) unconditionally returned `true` whenever `baseline === undefined` — i.e. for **every** newly-added Workbench record's child collection, even when that collection is empty. Saving one brand-new enemy with only a name therefore posted `SetEnemyAttributeDistributions {[]}`, `SetEnemySkills {[]}`, and `SetEnemySpawns {[]}` — three no-op HTTP writes, each triggering the all-sets reference-cache rebuild + cross-instance broadcast documented in #1708. The same pattern existed in `zone.ts`, `class.ts`, `skill-recipe.ts`, `lesson.ts`, and (worse) the Progression store, where a new path's auto-created tier posted `SetProficiencyModifiers {[]}`, `SetProficiencyRewards {[]}`, and `SetProficiencyPrerequisites [{prerequisiteIds: []}]` on top of its two identity batches.

## Change

`childChanged` now treats an added record (`baseline === undefined`) whose current collection is an empty array as unchanged, so a new record only posts the child writes that actually carry data. This is the single shared helper every `EntityConfig`'s child savers and the Progression store's save orchestration route through, so the fix applies uniformly across all entities without touching each call site.

## Test plan

- [x] Added unit tests for `childChanged` covering the new empty-vs-non-empty-add and defined-baseline cases (`save-helpers.test.ts`).
- [x] Added an entity-level regression test: adding a bare new enemy (all child collections empty) no longer posts `SetEnemyAttributeDistributions`/`SetEnemySkills`/`SetEnemySpawns` (`enemy.test.ts`).
- [x] Added a Progression-store regression test: a brand-new path + auto-created tier with no modifiers/rewards/prerequisites no longer posts those three child setters (`progression-store.test.ts`).
- [x] Updated an existing prerequisite-batching test whose expectation baked in the old over-posting behavior (an unmodified new tier no longer appears in the prerequisites batch with an empty array).
- [x] `npm run lint` — clean.
- [x] `npm run test:unit:ci` — 299 files / 3702 tests passing.

Closes #1895

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lan7Ve4JeGEkvAtLb5jyhg)_

### DIFF
--- a/UI/src/routes/admin/workbench/save-helpers.ts
+++ b/UI/src/routes/admin/workbench/save-helpers.ts
@@ -33,9 +33,17 @@ export const canonicalEqual = (a: unknown, b: unknown): boolean =>
 
 export const listsEqual = canonicalEqual;
 
-/** True when a child collection differs from its saved baseline (added records have none). */
-export const childChanged = <T>(current: T, baseline: T | undefined): boolean =>
-	baseline === undefined || !listsEqual(current, baseline);
+/**
+ * True when a child collection differs from its saved baseline (added records have none). An
+ * added record with an empty current collection is treated as unchanged — nothing was actually
+ * added to it — so a brand-new record doesn't trigger a no-op write for every empty child section.
+ */
+export const childChanged = <T>(current: T, baseline: T | undefined): boolean => {
+	if (baseline === undefined) {
+		return !(Array.isArray(current) && current.length === 0);
+	}
+	return !listsEqual(current, baseline);
+};
 
 /**
  * Map the local (negative) ids of added records to their persisted ids. Purely positional pairing (the
--- a/UI/src/tests/routes/admin/workbench/entities/enemy.test.ts
+++ b/UI/src/tests/routes/admin/workbench/entities/enemy.test.ts
@@ -180,6 +180,28 @@ describe('enemyEntity', () => {
 		expect(postBodyTo('AdminTools/SetEnemySpawns')).toEqual({ enemyId: 7, spawns: [{ zoneId: 1, weight: 8 }] });
 	});
 
+	it('persist Adds a bare new enemy without posting empty child collections (#1895)', async () => {
+		const added: WorkbenchEnemy = {
+			id: -1,
+			name: 'New Enemy',
+			designerNotes: '',
+			isBoss: false,
+			attributeDistribution: [],
+			skillPool: [],
+			spawns: [],
+			bossZones: []
+		};
+		socket.enemies = [{ ...added, id: 7 }];
+
+		await enemyEntity.persist({ added: [added], modified: [], deleted: [], existingIds: [] });
+
+		expect(postBodyTo('AdminTools/AddEditEnemies')[0].changeType).toBe(EChangeType.Add);
+		// Nothing was ever added to any child collection, so none of their setters should post.
+		expect(postBodyTo('AdminTools/SetEnemyAttributeDistributions')).toBeUndefined();
+		expect(postBodyTo('AdminTools/SetEnemySkills')).toBeUndefined();
+		expect(postBodyTo('AdminTools/SetEnemySpawns')).toBeUndefined();
+	});
+
 	it('refresh derives bossZones from the zones whose dedicated boss is this enemy', async () => {
 		socket.enemies = [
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: fix `childChanged` over-posting empty child collections for newly-added Workbench records (#1895)

**Verdict:** Approve *(submitted as a comment — I'm the PR author, so GitHub won't accept a gating APPROVE from me)* — the fix is correct, minimal, lands in the single shared helper every entity + the Progression store route through, and is well covered by tests. No blocking findings, no should-fixes, no nits.

I verified the change against every `childChanged` call site (`class.ts`, `item.ts`, `enemy.ts`, `lesson.ts`, `zone.ts`, `item-mod.ts`, `skill-recipe.ts`, and the four Progression-store call sites). All pass array-typed collections, so the new `Array.isArray(current) && current.length === 0` guard applies uniformly, and the `Array.isArray` check keeps the added-record branch safe if a collection were ever `null`/`undefined` (falls back to the prior "changed" behavior). The `baseline`-defined branch is untouched, so a modified record that clears a collection to `[]` still correctly posts the removal.

The subtle case — the Progression store's batched `SetProficiencyPrerequisites` — is handled correctly: a brand-new tier with empty `prerequisiteIds` is now filtered out of the batch, which is a genuine no-op (a new tier has no persisted prerequisites to clear and contributes no edges to the cycle-validation graph). Other tiers that reference the new tier as a prerequisite still post their own entries, so the graph is unaffected. The updated test at `progression-store.test.ts` confirms exactly this (the `{ id: 2, prerequisiteIds: [] }` entry is gone while the real edges remain).

### Test coverage

Solid and appropriately DRY:
- Direct unit tests for `childChanged` cover all four branches (empty add, non-empty add, defined-differing baseline, equal baseline) — `save-helpers.test.ts`.
- One representative entity-level regression test (enemy) asserts the three child setters no longer post for a bare new record — `enemy.test.ts`. Since the fix is in the shared helper, a per-entity test for zone/class/lesson/skill-recipe would be redundant; the shared unit test covers them.
- A Progression-store regression test covers its distinct save-orchestration path, and an existing prerequisite-batching test was correctly updated to drop the stale empty-entry expectation — `progression-store.test.ts`.

### What's good

- Fixing at the single shared helper rather than per call site is the right call — the amplification is eliminated everywhere at once with no scattered edits.
- The doc comment clearly explains *why* an empty add is treated as unchanged, which protects against a future "simplification" naively restoring the unconditional `baseline === undefined` return.
- Good instinct catching the internal inconsistency called out in the issue (diff-shaped savers already no-op on empty adds; only the whole-collection setters over-posted).

Waiting on CI; will merge once it's green.

---

PR #1989, opened 2026-07-15
Thread pre-resolved BattlerMaterials into RecordVictory to skip a duplicate battler build (#1897)

### LINKED ISSUE #1897: Victory path builds the snapshot battler twice per idle kill
## Problem

`BattleService.RecordVictory` calls `snapshot.ToBattler(...)` (`Game.Application/Services/BattleService.cs:541-543`) even though `SimulateBattle` already built the identical battler moments earlier in the same command (`:668-669`). Every `DefeatEnemy`/won-abandon therefore materializes the full modifier graph + `AttributeCollection` twice from the same frozen snapshot.

Small next to the anti-cheat replay itself, but this is the hottest command in the game and the fix is free: thread the battler from `TryResolveActiveBattle` into `RecordVictory`.

Related: #1820 covers the duplicated *progress-hash reads* on these commands; #1730 the offline loop's rating recompute — neither covers the duplicate battler assembly.

ginderjeremiah:
## Summary

`BattleService.RecordVictory` called `snapshot.ToBattler(...)` to rebuild the player's battle
`Battler` even though `SimulateBattle` (invoked moments earlier via `TryResolveActiveBattle` in the
same command) had already built the identical one from the same frozen snapshot. Every
`DefeatEnemy`/won-abandon — the hottest command in the game — re-walked the full item/mod/skill/
proficiency/class resolution twice.

- `TryResolveActiveBattle` now resolves the player's `BattleSnapshot.GetBattlerMaterials(...)`
  **once** and outputs it via a new `out BattlerMaterials playerMaterials` parameter, alongside the
  existing `enemy`/`result` outputs.
- `SimulateBattle` now takes the pre-resolved `BattlerMaterials` instead of rebuilding it from the
  snapshot.
- `RecordVictory` now accepts an optional `BattlerMaterials? playerMaterials = null`. When supplied
  by a caller that already resolved it (`EndBattleVictory`, the won-abandon branch of
  `AbandonBattle`), it skips the redundant resolution. A caller with no materials on hand (e.g. a
  direct unit-test call) falls back to resolving it from `state.Snapshot` exactly as before — fully
  backward compatible.

**Correctness note:** a naive "just pass the already-simulated `Battler`" fix would have been a real
regression — `BattleSnapshot.GetBattlerMaterials`'s own doc comment flags that "a `Battler` is
single-use (simulation mutates its health and active effects)". Reusing the fought/mutated instance
for the reward rating would have made `PlayerRating` depend on which self-buffs happened to be
active at the exact moment the enemy died, rather than the player's stable battle-start capability —
breaking the "same frozen snapshot, not the live aggregate" invariant several existing tests pin
(`RecordVictory_SnapshotsPlayerRatingOntoStats_FromSnapshotNotLiveAggregate`, etc.). Instead, this
reuses the existing `BattlerMaterials`/`.Build()` split (the same pattern already used by the offline
simulator, #1730) — the expensive resolver/LINQ composition runs once, but each consumer still gets
its own fresh, single-use `Battler` via the cheap `Build()` call, so the rating still reads
pristine battle-start state.

## Test plan

- [x] `dotnet test` (full backend suite) — all 3205 tests pass across `Game.Core.Tests`,
      `Game.Infrastructure.Tests`, `Game.Application.Tests`, `Game.Api.Tests`, including the
      existing `RecordVictory`/`EndBattleVictory`/`EndBattleLoss`/`AbandonBattle` integration tests
      that pin the pristine-snapshot rating behavior
- [x] No frontend changes (backend-only performance fix)

Closes #1897

https://claude.ai/code/session_014N8qTw83JgC6DSqzaWCfWW

---
_Generated by [Claude Code](https://claude.ai/code/session_014N8qTw83JgC6DSqzaWCfWW)_

### DIFF
--- a/Game.Application/Services/BattleService.cs
+++ b/Game.Application/Services/BattleService.cs
@@ -283,7 +283,7 @@ namespace Game.Application.Services
 
         public async Task<DefeatResult?> EndBattleVictory(Player player, PlayerState state, int? clientTotalMs = null, CancellationToken cancellationToken = default)
         {
-            if (!TryResolveActiveBattle(state, out var enemy, out var result))
+            if (!TryResolveActiveBattle(state, out var enemy, out var result, out var playerMaterials))
             {
                 // No battle to resolve. After the caller's HasActiveBattle gate this means a torn state
                 // (an enemy id set without its snapshot), which the set/clear invariant should prevent.
@@ -333,7 +333,7 @@ namespace Game.Application.Services
                 return null;
             }
 
-            var rewards = RecordVictory(player, enemy, result, state, now);
+            var rewards = RecordVictory(player, enemy, result, state, now, playerMaterials: playerMaterials);
 
             // Anchor the post-battle cooldown to the battle's server-computed completion, not to now: the gap
             // between completion and now is the post-victory network latency, so subtracting it keeps the idle
@@ -356,7 +356,7 @@ namespace Game.Application.Services
 
         public async Task<bool> EndBattleLoss(Player player, PlayerState state, CancellationToken cancellationToken = default)
         {
-            if (!TryResolveActiveBattle(state, out var enemy, out var result))
+            if (!TryResolveActiveBattle(state, out var enemy, out var result, out _))
             {
                 return false;
             }
@@ -483,7 +483,7 @@ namespace Game.Application.Services
 
             // No active battle (nothing to resolve) or no elapsed window to re-simulate against — clear
             // and return without recording an outcome or persisting.
-            if (elapsedMs <= 0 || !TryResolveActiveBattle(state, out var enemy, out var result, simulateMs))
+            if (elapsedMs <= 0 || !TryResolveActiveBattle(state, out var enemy, out var result, out var playerMaterials, simulateMs))
             {
                 state.ClearBattle();
                 return new StaleBattleResolution(null, null);
@@ -507,7 +507,7 @@ namespace Game.Application.Services
                 // abandon resolved as a real victory. Pay it out exactly like EndBattleVictory (exp +
                 // win/clear/challenge credit) rather than booking the win while silently withholding the
                 // earned exp (#206).
-                RecordVictory(player, enemy, result, state, now, notify);
+                RecordVictory(player, enemy, result, state, now, notify, playerMaterials);
             }
             else if (!result.PlayerDied && wallClockMs < GameConstants.DefaultMaxBattleMs)
             {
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: thread pre-resolved `BattlerMaterials` into `RecordVictory` to skip the duplicate battler build (#1897)

**Verdict:** Approve (submitted as a comment since I'm the PR author) — the change does exactly what the issue asked, the reuse is provably safe, and behavior is preserved.

I traced the correctness question the issue's "free fix" hinges on — is it safe to reuse the resolved materials across the simulate and rating builds? It is:

- `BattlerMaterials.Build()` constructs a fresh `AttributeCollection` and a fresh `Battler` on every call, and `AttributeCollection`'s constructor **copies** modifiers into its own node array rather than retaining the passed `baseModifiers` list — so `Build()`'s `AddModifier(signaturePassive)` never mutates the shared list, and each consumer gets an independent, single-use `Battler`. The fought/mutated instance from `SimulateBattle` is never reused for the rating.
- This is the same `GetBattlerMaterials`/`Build()` split the offline simulator already reuses per-battle (#1730), so the pattern is proven, not novel. Correctly avoids the naive "pass the already-simulated `Battler`" trap the PR description calls out, which would have broken the pristine-snapshot rating invariant.

Layering is respected — resolving the materials in `TryResolveActiveBattle` is orchestration that sits right beside the `SimulateBattle` call it already owned, no domain logic added. Making `SimulateBattle` `static` once it no longer needs the instance resolvers is a nice touch. `EndBattleLoss` discarding via `out _` does the same single build as before, so the loss path pays nothing extra.

### Test coverage

Adequate for a behavior-preserving refactor — no new tests required:
- The threaded (non-null `playerMaterials`) path is exercised by the `EndBattleVictory`/`AbandonBattle` integration tests.
- The null fallback path is exercised by the direct `RecordVictory` unit/integration calls (`RecordVictory_SnapshotsPlayerRatingOntoStats_FromSnapshotNotLiveAggregate`, `_RewardIncludesClassLockedBaseInPlayerRating`, etc.), which pass no materials.
- The existing pristine-snapshot rating assertions pin the exact invariant a bad reuse would have broken.

### Nits

- The optional `playerMaterials` with the resolve-it-here fallback keeps one production branch alive essentially for the direct-call test path. It's well-documented, defensible, and low-risk (and doubles as a defensive invariant), so I'm not asking for a change — just flagging that a future reader shouldn't mistake it for dead code.

### What's good

Tight, well-scoped perf fix on the hottest command in the game, with comments that correctly capture the single-use-`Battler` subtlety and the anti-cheat/pristine-rating rationale so the next editor doesn't naively "simplify" it into a regression.

---

PR #2009, opened 2026-07-15
Only spend socket auth-retry budget on an actual token rotation

### LINKED ISSUE #1992: Socket auth-retry budget is spent by retryable refresh failures — a ~1-minute outage force-logs the player out and clears a still-valid token pair
## Problem

`ApiSocket.handleClose` increments `socketAuthRetries` **before** the refresh outcome is known (`UI/src/lib/api/api-socket.ts:453`), and nothing ever decrements or resets it on a *retryable* outcome — the only refill is the stable-connection timer in `onStart` (`api-socket.ts:485-488`), which never fires while the server is unreachable.

**Failure scenario:** a backend restart or network drop lasting ~6 keepalive cycles (~60s at the 10s ping). Each failed pre-open handshake (close 1006, `socketOpened === false`) increments the counter, then `refreshTokens()` fails with a network error → the `retryable` branch (`:473-476`) deliberately preserves tokens and lets the keepalive retry — but the increment sticks. On the 6th close the budget branch (`:432-443`) declares the session unrecoverable, stops the keepalive, and calls `handleAuthFailure()`, which **clears the token pair** (`UI/src/lib/api/auth.ts:164-169`) and routes to login — exactly the "forcing a logout over a transient failure" the retryable branch's own comment forbids.

Per `docs/frontend-sockets.md`, the bound exists to stop **burning single-use refresh tokens** against a server that keeps rejecting freshly-refreshed tokens. A retryable attempt spends no rotation, so it shouldn't consume budget. For an idle game routinely left in background tabs (overnight Wi-Fi blip, laptop wake before network), this fires under completely ordinary conditions and destroys a session that was still recoverable.

## Fix direction

Count only attempts that actually spent a rotation: increment on refresh outcome `success` (i.e. a rotation happened and the handshake still failed), or reset the counter when the outcome is `retryable`. Add unit tests for: outage → recovery keeps the session; genuinely-rejected fresh tokens still exhaust the budget.

ginderjeremiah:
## Summary

- `ApiSocket.handleClose` incremented `socketAuthRetries` **before** the refresh outcome was known, so a `retryable` refresh (network blip, transient server error) still burned budget even though no refresh token was actually spent.
- Failure scenario: a backend restart or network drop lasting ~6 keepalive cycles (~60s at the 10s ping) exhausts the 5-attempt budget purely on retryable failures, forcing `handleAuthFailure()` to clear a still-valid token pair and log the player out — exactly the outcome the retryable branch's own comment says to avoid.
- Fix: move the increment into the `refreshTokens().then()` callback, incrementing only on `outcome.status === 'success'` (a rotation happened and the reconnect still failed to open). A `retryable` outcome now leaves the budget untouched entirely, matching the bound's actual purpose — stopping a flapping/broken server from burning single-use refresh tokens — while a genuine outage recovers without penalty once the network returns.

Closes #1992

## Test plan

- [x] Added `never exhausts the budget or logs out across a long outage of only-retryable refresh failures` to `UI/src/tests/lib/api/api-socket.test.ts` — 10 consecutive retryable refresh failures never touch `socketAuthRetries` or call `handleAuthFailure`, and tokens stay intact.
- [x] Existing `stops refreshing after the retry limit when no reconnect ever becomes stable` test (5 consecutive `success`-but-still-rejected reconnects exhaust the budget) still passes unchanged, confirming a genuinely flapping server is still bounded.
- [x] `npm run test:unit:ci` — full suite, 3707/3707 passing.

https://claude.ai/code/session_0143JiyMP79TnsKzuz5TRfzZ

---
_Generated by [Claude Code](https://claude.ai/code/session_0143JiyMP79TnsKzuz5TRfzZ)_

### DIFF
--- a/UI/src/lib/api/api-socket.ts
+++ b/UI/src/lib/api/api-socket.ts
@@ -450,9 +450,11 @@ export class ApiSocket {
 			return;
 		}
 
-		this.socketAuthRetries++;
 		refreshTokens().then((outcome) => {
 			if (outcome.status === 'success') {
+				// Only a rotation that still failed to open spends budget — a retryable outage never burns
+				// a refresh token, so it must not count toward the bound that exists to stop that burn.
+				this.socketAuthRetries++;
 				// processCommandQueue ensures the socket (now with the freshly refreshed token) and flushes
 				// any queued-but-unsent commands.
 				void this.processCommandQueue();
@@ -471,8 +473,9 @@ export class ApiSocket {
 				handleAuthFailure();
 			}
 			// Otherwise the refresh attempt was retryable (network blip, transient server error): leave the
-			// queue and tokens intact — the keepalive ping's next ensureSocket() call retries the reconnect,
-			// which will attempt the refresh again, rather than forcing a logout over a transient failure.
+			// queue and tokens intact and spend no budget — the keepalive ping's next ensureSocket() call
+			// retries the reconnect, which will attempt the refresh again, rather than forcing a logout over
+			// a transient failure.
 		});
 	}
 
--- a/UI/src/tests/lib/api/api-socket.test.ts
+++ b/UI/src/tests/lib/api/api-socket.test.ts
@@ -1056,6 +1056,27 @@ describe('ApiSocket', () => {
 			warnSpy.mockRestore();
 		});
 
+		it('never exhausts the budget or logs out across a long outage of only-retryable refresh failures', async () => {
+			setTokens({ accessToken: 'a', refreshToken: 'r' });
+			vi.mocked(refreshTokens).mockResolvedValue({ status: 'retryable' });
+
+			apiSocket.sendSocketCommand('DefeatEnemy', { clientTotalMs: 1 });
+			await flushMicrotasks();
+
+			// Every reconnect attempt during the outage fails pre-open (e.g. code 1006, network down), and
+			// each refresh comes back retryable (no rotation spent) — well past MAX_SOCKET_AUTH_RETRIES (5)
+			// worth of cycles, since a retryable outcome must never count toward the budget.
+			for (let i = 0; i < 10; i++) {
+				rejectHandshake(lastWs());
+				await flushMicrotasks();
+			}
+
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: bound the socket auth-retry budget to actual token rotations (closes #1992)

**Verdict: Approve** — the fix correctly targets the root cause, is minimal and well-tested, and CI is green. (Posted as a `COMMENT` review because I'm the PR author; the real verdict is Approve.)

### Correctness
- The increment is now inside `refreshTokens().then()` and gated on `outcome.status === 'success'`, and `success` is precisely "a fresh pair was minted (or an adopted concurrent-tab pair)" per `auth.ts` — i.e. a rotation actually happened. So the counter now measures exactly what the bound exists to limit (burning single-use refresh tokens), not raw close events. This matches the fix direction in #1992.
- No off-by-one: the old code spent budget on 5 closes then exhausted on the 6th; the new code increments on 5 successful rotations then exhausts on the 6th close — same bound of 5 rotations, verified against the unchanged `stops refreshing after the retry limit` test still expecting 5 refresh calls.
- The `retryable` branch now leaves budget untouched, so a multi-cycle outage no longer force-clears a still-valid token pair — exactly the regression described in the issue.

### Test coverage
- The new `never exhausts the budget or logs out across a long outage of only-retryable refresh failures` test drives `handleClose` 10× on retryable refreshes and asserts `socketAuthRetries === 0`, `handleAuthFailure` not called, and tokens intact — the recovery direction.
- The exhaustion direction stays covered by the existing `success`-outcome tests (`stops refreshing after the retry limit`, `stops the keepalive and routes to re-auth when the retry budget is exhausted`). Both directions the issue asked for are present. Frontend-only auth path, so no backend mirror is needed.

### What's good
- Tightly scoped fix on the real defect rather than papering over it with a reset, with comments that explain *why* only a rotation spends budget. `docs/frontend-sockets.md` already frames the bound around "burning single-use refresh tokens", so the mechanism now matches the documented intent with no doc change required.

---

PR #2010, opened 2026-07-15
Guard EndBattleVictory/EndBattleLoss against re-crediting an already-credited battle

### LINKED ISSUE #1993: Battle-credit seed backstop (#1874) only guards AbandonBattle — direct EndBattleVictory/EndBattleLoss can re-credit a stale battle, and the check runs after the full replay
Two gaps in the just-landed #1874 idempotency backstop (commit 71b18b4):

## 1. Coverage: the direct battle-end commands skip the seed check

`Player.LastCreditedBattleSeed` is compared only inside `AbandonBattle` (`Game.Application/Services/BattleService.cs:498`). The rationale (docs/backend-persistence.md: the abandon path is "the one a reconnect's next battle-end actually reaches") holds for a **well-behaved** client, but nothing forces a reconnecting client through it. In the crash/cancel window #1874 exists for — durable credit landed but the awaited session save didn't (e.g. teardown cancellation between `SavePlayer` and `SavePlayerStateAsync`, the #1849 unwind) — the stale session still presents the battle as active, and a client that sends `DefeatEnemy`/`BattleLost` directly passes every existing gate: `HasActiveBattle` is true, the deterministic replay of the same seed/snapshot re-yields the same outcome, and the server-clock elapsed-time check passes because real time has passed. Result: a **second full credit** (exp, kills, statistics, challenge + proficiency accrual) for the same fight — and a tampered client can deliberately drop the socket mid-`DefeatEnemy` to farm the window.

Fix: the same seed comparison at the top of `EndBattleVictory` (`BattleService.cs:284`) and `EndBattleLoss` (`:357`) — or once in the shared `TryResolveActiveBattle` preamble, which is where the invariant ("at most one battle can be credited-but-not-yet-session-cleared") naturally lives.

## 2. Ordering: the backstop pays for a full replay it then discards

Within `AbandonBattle` the seed check sits **after** `TryResolveActiveBattle` (`:486` vs `:498`), which runs the full battle re-simulation (up to the battle cap of replayed ticks plus a complete `BattlerMaterials` resolution). The check needs only `state.BattleSeed` and `player.LastCreditedBattleSeed`, both available before the replay — hoisting it makes the reconnect-after-crash path O(1).

Integration test: simulate the credited-but-session-stale state, then drive each of the three end paths and assert no second credit.

ginderjeremiah:
## Summary

Closes #1993. Two gaps in the #1874 idempotency backstop (`Player.LastCreditedBattleSeed`):

1. **Coverage gap.** The seed check that prevents re-crediting an already-durably-credited battle only lived in `AbandonBattle`. A client that reconnects into the crash window #1874 targets (durable credit landed, but the awaited session save didn't) could bypass it entirely by calling `DefeatEnemy`/`BattleLost` directly instead of going through the natural reconnect-abandon path — both commands pass every other gate (active battle, deterministic replay, elapsed-time check) and would re-credit exp/kills/statistics/challenge/proficiency accrual for the same fight.
2. **Ordering gap.** Inside `AbandonBattle`, the seed check ran *after* `TryResolveActiveBattle`'s full replay, even though it only needs `state.BattleSeed` and `player.LastCreditedBattleSeed` — both available up front.

### Changes

- Extracted the seed-comparison into a shared `BattleAlreadyCredited(state, player)` helper (`Game.Application/Services/BattleService.cs`).
- Added the same guard to the top of `EndBattleVictory` and `EndBattleLoss`, before any replay work — rejects the claim and clears the stale session battle.
- Hoisted the existing `AbandonBattle` check ahead of `TryResolveActiveBattle`, so a stale reconnect no longer pays for a full re-simulation it would only discard.
- Updated `docs/backend-persistence.md`'s description of the backstop to reflect that all three battle-end paths guard it now, and that the check runs before any replay.

## Test plan

- [x] Added `EndBattleVictory_BattleAlreadyCredited_ReturnsNullAndClearsStaleSession` — asserts the claim is rejected and no exp is granted, without backdating `BattleStartTime` (proving the seed check runs before the elapsed-time gate/replay).
- [x] Added `EndBattleLoss_BattleAlreadyCredited_ReturnsFalseAndClearsStaleSession` — same shape for the loss path.
- [x] `dotnet test` — full backend suite green (3207 tests: Core, Infrastructure, Application, Api).

Closes #1993

---
_Generated by [Claude Code](https://claude.ai/code/session_01BQ86kYoCYQQ1oF3VWeCAUA)_


ginderjeremiah (commented):
## Review: Guard the direct battle-end commands against re-crediting an already-credited battle (#1993)

**Verdict:** Approve _(submitted as a comment — I'm the PR author, so GitHub won't accept a gating event from me; the intended verdict is **Approve**)_ — closes both gaps #1993 identified with a clean shared guard and no behavior regression.

The change addresses exactly what the issue calls out: the #1874 seed backstop only lived in `AbandonBattle`, so a reconnecting (or tampered) client could re-credit a durably-credited fight by calling `DefeatEnemy`/`BattleLost` directly, and even on the abandon path the check ran after a full replay it would only discard.

I verified the substance rather than taking the description at face value:

- **Coverage gap closed.** `BattleAlreadyCredited` now gates the top of `EndBattleVictory` (`BattleService.cs:290`) and `EndBattleLoss` (`:374`), before `TryResolveActiveBattle`, so neither direct path can double-pay a credited seed. The shared helper (`:711`) keeps the invariant in one place across all three battle-end paths.
- **Ordering gap closed, and the hoist is behavior-preserving.** In the old `AbandonBattle`, an already-credited battle always ended at `StaleBattleResolution(null, null)` with no credit and no `SavePlayer` — either via the `elapsedMs <= 0 || !TryResolveActiveBattle` branch or the seed check that followed it. Hoisting the check to the top (`:496`) reaches the same result without paying for the re-simulation. No externally-observable change to the abandon path.
- **Clear is complete.** `state.ClearBattle()` nulls `BattleSeed` (`PlayerState.cs:90`), so a second direct call can't re-match the credited seed — the guard leaves no stale-seed residue.
- **Layering / parity.** This is an Application-layer anti-cheat/idempotency backstop, not battle-simulation logic, so it correctly stays server-side with no frontend mirror required (per `backend.md` — only the deterministic tick loop is parity-bound).

### Test coverage

- `EndBattleVictory_BattleAlreadyCredited_...` and `EndBattleLoss_BattleAlreadyCredited_...` assert no credit is granted and the stale session is cleared, and deliberately don't backdate `BattleStartTime` — proving the seed check fires ahead of the elapsed-time gate/replay, which is the ordering fix's whole point.
- The abandon path's already-credited behavior is unchanged and remains covered by the existing #1874 test, so re-testing it here isn't needed.
- The loss-path comment usefully contrasts this clear-the-battle rejection with the `ClaimedBeforeBattleCouldFinish` rejection that leaves the battle active for a legitimate retry — the asymmetry is intentional and now documented in the test.

### What's good

- One shared `BattleAlreadyCredited` helper instead of three copies of the seed comparison — the "at most one credited-but-not-session-cleared battle" invariant lives in a single place.
- Logging asymmetry is a deliberate, correct call: the direct paths `LogWarning` (an unusual, tamper-adjacent re-present) while the benign reconnect-abandon path stays quiet.
- `docs/backend-persistence.md` updated to state all three paths guard the seed and that the check precedes any replay — docs and code stay in sync.

No blocking findings, no should-fixes, no nits. Clean, well-scoped fix. Holding the merge only until the in-progress `test-backend` and `test-e2e (webkit)` jobs go green.

---

PR #2011, opened 2026-07-15
Reject applying the same item mod to a second slot of one item

### LINKED ISSUE #1994: Server accepts the same item mod applied to multiple slots of one item — the client forbids it (anti-cheat asymmetry)
## Problem

`Inventory.TryApplyMod` (`Game.Core/Players/Inventories/Inventory.cs:108-148`) validates: mod unlocked, item unlocked, slot exists, `modSlot.Type == mod.Type`. It never checks whether the same `ItemModId` is already applied to **another slot of the same item**. The UI's mod picker enforces exactly that rule — `compatibleMods` excludes mods already on the item across all its slots (`UI/src/routes/game/screens/inventory/inventory-view.svelte.ts:244-251`, `!used.has(m.id)` over the whole `appliedMods` set) — so "one copy per item" is the intended rule, but it exists only client-side. The DB can't catch it either: `AppliedMod`'s PK is `(PlayerId, ItemId, ItemModSlotId)`, which permits duplicate `ItemModId`s across slots.

**Failure scenario:** an item authored with two+ slots of the same `EItemModType`. A tampered client sends `ApplyMod` twice with the same `itemModId` against both slot ids; the server accepts both, the write-behind persists both, and `Item.GetAttributeModifiers`/the battle snapshot compose the mod's attributes **twice** — a stat advantage unreachable through the UI. This contradicts the codebase's own posture that every player mutation validates as anti-cheat (cf. `TryEquipItem`'s "a tampered client cannot bypass the frontend gate" comment).

Latent today (`content/item-mods.json` is empty) but silently exploitable the moment mod content with same-type multi-slot items lands.

## Fix direction

Add the duplicate-across-slots rejection to `TryApplyMod` (mirror the client's rule), plus unit tests (apply same mod to second slot → rejected, no mutation) and a socket-command integration test.

ginderjeremiah:
## Summary

`Inventory.TryApplyMod` (`Game.Core/Players/Inventories/Inventory.cs`) validated mod-unlocked, item-unlocked, slot-exists, and slot-type-matches, but never checked whether the same `ItemModId` was already applied to a *different* slot of the same item. The client's mod picker already enforces "one copy of a mod per item" (`compatibleMods` in `UI/src/routes/game/screens/inventory/inventory-view.svelte.ts` excludes mods already on the item across all its slots), so this was a client-only anti-cheat gate — a tampered client could apply the same mod to two slots of an item with same-type slots, and the mod's attributes would be composed twice (`Item.GetAttributeModifiers`/battle snapshot).

The DB schema doesn't prevent it either (`AppliedMod`'s PK is `(PlayerId, ItemId, ItemModSlotId)`, which permits duplicate `ItemModId`s across slots for one item), and it's currently latent since `content/item-mods.json` is empty — but it becomes silently exploitable the moment multi-slot-same-type mod content ships.

## Change

- `Inventory.TryApplyMod` now rejects the apply (no mutation) when the mod is already applied to a *different* slot of the same item, mirroring the client's rule. Re-applying the same mod to its own current slot (a no-op) is still allowed.

## Test plan

- [x] `Game.Core.Tests/Players/InventoryTests.cs` — new unit tests: rejects duplicate-mod-on-another-slot with no mutation, allows re-applying the same mod to its own slot, allows different mods on different slots.
- [x] `Game.Api.Tests/Integration/InventorySocketTests.cs` — new integration test exercising the `ApplyMod` socket command end-to-end: applying a mod already on another slot of the item returns an error and the persisted player still shows only the original application.
- [x] Full backend suite (`dotnet test Game.sln`) — 3211/3211 passing.
- [x] `dotnet build Game.sln` — clean.

No frontend changes needed; the client-side rule already existed and is unaffected.

Closes #1994

---
_Generated by [Claude Code](https://claude.ai/code/session_017CWyowCEfuHRfCYakAQhwo)_

### DIFF
--- a/Game.Api.Tests/Integration/InventorySocketTests.cs
+++ b/Game.Api.Tests/Integration/InventorySocketTests.cs
@@ -149,6 +149,40 @@ namespace Game.Api.Tests.Integration
         }
 
         [Fact]
+        public async Task ApplyMod_SameModAlreadyOnAnotherSlot_ReturnsErrorAndDoesNotDuplicate()
+        {
+            // Anti-cheat (#1994): the client's mod picker forbids applying the same mod to a second slot
+            // of one item, so a tampered client sending it directly over the socket must be rejected too.
+            var firstSlotId = 0;
+            var secondSlotId = 0;
+            var modId = 0;
+            var (userId, playerId, itemId) = await SeedPlayerWithInventoryAsync("dupmoduser", "dupmodpass",
+                async (context, playerId, item) =>
+                {
+                    var firstSlot = await TestDataSeeder.AddItemModSlotAsync(context, item.Id);
+                    var secondSlot = await TestDataSeeder.AddItemModSlotAsync(context, item.Id);
+                    await TestDataSeeder.LinkItemToPlayerAsync(context, playerId, item.Id, EEquipmentSlot.WeaponSlot);
+                    var mod = await TestDataSeeder.CreateItemModAsync(context, attributeId: EAttribute.Dexterity, attributeAmount: 7m);
+                    await TestDataSeeder.LinkModToPlayerAsync(context, playerId, mod.Id);
+                    await TestDataSeeder.ApplyModToItemAsync(context, playerId, item.Id, firstSlot.Id, mod.Id);
+                    firstSlotId = firstSlot.Id;
+                    secondSlotId = secondSlot.Id;
+                    modId = mod.Id;
+                });
+            await LoginAsync("dupmoduser", "dupmodpass");
+            await using var socketClient = await ConnectSocketAsync(userId);
+
+            var parameters = new { itemId, itemModId = modId, itemModSlotId = secondSlotId };
+            var response = await socketClient.SendCommandRawAsync("ApplyMod", parameters);
+
+            Assert.NotNull(response.Error);
+            var data = await GetPersistedPlayerAsync(playerId);
+            var appliedMods = data.InventoryData.UnlockedItems.Single(i => i.ItemId == itemId).AppliedMods;
+            var applied = Assert.Single(appliedMods);
+            Assert.Equal(firstSlotId, applied.ItemModSlotId);
+        }
+
+        [Fact]
         public async Task RemoveMod_AppliedMod_Succeeds()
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Reject applying the same item mod to a second slot of one item (#1994)

**Verdict:** Approve — the fix closes the anti-cheat asymmetry exactly as the issue describes, lands in the right layer, and is well tested. (Submitted as a comment since I'm the PR author; GitHub won't accept a self-approval.)

### Assessment

- **Solves the problem.** The new guard in `Game.Core/Players/Inventories/Inventory.cs:135` rejects an `ApplyMod` whose `ItemModId` is already applied to a *different* slot of the same item, mirroring the client's `compatibleMods` rule (`inventory-view.svelte.ts:249`, `!used.has(m.id)`). This closes the tampered-client path where the same mod's attributes would compose twice in the battle snapshot.
- **Correct placement and logic.** The check sits after the mod-unlocked / item-unlocked / slot-exists / type-match validations and before the slot-replace logic — so it rejects before any mutation. The `a.ItemModSlotId != itemModSlotId` guard correctly excludes the target slot, keeping a same-slot re-apply a legal no-op (which then falls through to the harmless replace).
- **Right layer.** The rule is domain invariant logic and lives in `Game.Core`, consistent with the DDD layering in `backend.md` and the existing anti-cheat posture in `TryEquipItem`.

### Test coverage

- Unit (`InventoryTests.cs`): duplicate-on-another-slot → rejected with no mutation; same-mod re-applied to its own slot → succeeds; different mods on different slots → succeeds. The rejection test deliberately uses two `Prefix` slots — the exact exploitable shape from the issue.
- Integration (`InventorySocketTests.cs`): drives the `ApplyMod` socket command end-to-end and asserts the persisted player still shows only the original application.

No frontend changes are warranted — the client rule already existed and battle computation is untouched, so there's no mirrored battle-test obligation here.

### Notes

No blocking findings, should-fixes, or nits. Clean, well-scoped change.

I'll hold the merge until CI is green (`test-backend` and the e2e jobs are still running as of this review).

---

PR #2012, opened 2026-07-15
Keep boss-gated Workbench pickers from silently dropping an unflagged current value

### LINKED ISSUE #1996: Workbench boss-gated pickers silently drop a current value that lost its boss flag (missing keep exception)
## Problem

The Workbench picker convention keeps the current value visible even when it no longer passes the filter — `retireableOptions` handles retirement, and the flag-filtered skill pickers all add `|| s.id === keep` inside their flag filters (`grantedSkillOptions`, `synthesisResultSkillOptions`, `playerSkillOptions` — `UI/src/routes/admin/workbench/reference.svelte.ts:180-186, 219-223, 236-242`, each with a "so an already-authored grant isn't silently dropped" comment).

The two **boss-filtered** pickers miss the exception: `bossEnemyOptions` pre-filters `e.isBoss` with no `keep` escape (`reference.svelte.ts:163-169`), and the boss-scoped challenge target picker does the same (`components/challenge/ScopeField.svelte`). So once an enemy's `isBoss` toggle is turned off, a zone's authored `bossEnemyId` (Identity select) or a boss challenge's `targetEntityId` vanishes from the option list — the native select renders blank, indistinguishable from "None", and the existing assignment can neither be seen nor re-picked.

The state is reachable: unflagging a referenced boss is allowed at save time (it's only a content-health lint — `Game.Application/Content/ProgressionGraphChecker.cs`), and no zone/challenge section `warn` covers it.

## Fix direction

Add the `|| e.id === keep` exception to both boss-filtered pickers (matching the skill pickers), and consider a section `warn` for "assigned enemy is no longer flagged as a boss" mirroring the recipe editor's stale-flag warning. Unit tests alongside the existing reference picker tests.

ginderjeremiah:
## Summary

- `bossEnemyOptions` (zone's Dedicated Boss picker) and `entityOptions`/`entityRecords` (the boss-only challenge target picker used by `ScopeField`) filtered enemies on `isBoss` *before* the retired-record `keep` exception ran, so an enemy whose `isBoss` flag was later turned off vanished from the option list entirely — the native `<select>` rendered blank, indistinguishable from "None", and the existing zone/challenge assignment couldn't be seen or re-picked.
- Both pickers now add the `|| e.id === keep` exception, matching the existing convention used by `grantedSkillOptions`/`synthesisResultSkillOptions`/`playerSkillOptions`.
- Added a section `warn` to the zone editor's Identity section and the challenge editor's Condition section (mirroring the skill-recipe editor's stale-flag warning) so an author sees "Dedicated boss is no longer flagged as a boss" / "Target enemy is no longer flagged as a boss" instead of the drift sitting silently — the state is reachable at save time since unflagging a referenced boss is only a content-health lint, not a hard block.

Closes #1996. Implements exactly the fix direction proposed in the issue: the `keep` exception on both boss-filtered pickers, plus a section warning for the "assigned enemy is no longer flagged as a boss" case.

## Test plan

- [x] Added unit tests to `reference.test.ts` covering `bossEnemyOptions(keep)` and `entityOptions(EEntityType.Enemy, true, keep)` keeping a non-boss current value visible.
- [x] Added unit tests to `zone.test.ts` and `challenge.test.ts` for the new section `warn`s (flags the stale boss flag, passes when the flag is intact or no boss is assigned).
- [x] `npx vitest --run` — full suite (299 files / 3714 tests) passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DcWSoXNYuBbgXVmeSRxTFh)_

### DIFF
--- a/UI/src/routes/admin/workbench/entities/challenge.ts
+++ b/UI/src/routes/admin/workbench/entities/challenge.ts
@@ -2,7 +2,7 @@ import { ApiRequest, EChallengeType, EEntityType, fetchSocketData, type IChallen
 import { staticData } from '$stores';
 import { reference } from '../reference.svelte';
 import { persistEntity } from '../save-helpers';
-import { challengeSentence, deriveFromType, fmtNum } from './challenge-helpers';
+import { challengeSentence, deriveFromType, fmtNum, typeBossOnly } from './challenge-helpers';
 import type { EntityConfig } from './types';
 
 // Challenges load over the socket; the admin filter invalidates this cache on every
@@ -95,6 +95,12 @@ export const challengeEntity: EntityConfig<IChallenge> = {
 				}
 				if (c.entityType === EEntityType.DamageType && c.targetEntityId == null) {
 					return 'Must target a damage type — this statistic has no global total';
+				}
+				// The target picker keeps a boss target visible even if it's since lost the
+				// boss flag (reference.entityOptions' `keep` exception); flag that drift here.
+				const target = c.targetEntityId != null ? staticData.enemies?.[c.targetEntityId] : undefined;
+				if (target && typeBossOnly(reference.challengeTypes, c.challengeTypeId) && !target.isBoss) {
+					return 'Target enemy is no longer flagged as a boss';
 				}
 				return null;
 			}
--- a/UI/src/routes/admin/workbench/entities/zone.ts
+++ b/UI/src/routes/admin/workbench/entities/zone.ts
@@ -66,6 +66,16 @@ export const zoneEntity: EntityConfig<WorkbenchZone> = {
 			glyph: 'tag',
 			desc: 'Name, level range & ordering',
 			kind: 'fields',
+			// The dedicated-boss picker keeps an authored value visible even if it loses its
+			// boss flag (reference.bossEnemyOptions' `keep` exception); flag that drift here
+			// since the zone would otherwise silently point at a non-boss enemy.
+			warn: (z) => {
+				if (z.bossEnemyId == null || z.bossEnemyId < 0) {
+					return null;
+				}
+				const boss = staticData.enemies?.[z.bossEnemyId];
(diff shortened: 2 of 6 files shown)


ginderjeremiah (commented):
## Review: boss-gated Workbench pickers keep an unflagged current value visible (#1996)

**Verdict:** Approve — faithfully implements the issue's fix direction, correctly wired end-to-end, and well tested. (Submitting as `COMMENT` since GitHub won't accept an approving event on my own PR; the intended verdict is **Approve**.)

### What's good

- The `|| e.id === keep` exception on `bossEnemyOptions` and `entityRecords` matches the established convention already used by `grantedSkillOptions` / `synthesisResultSkillOptions` / `playerSkillOptions` / `itemOptionsForSlot` — no new pattern invented.
- End-to-end wiring is correct: `FieldControl.svelte:80,93` passes the current field value as `keep`, so the zone `bossEnemyId` picker resolves `bossEnemyOptions(currentValue)`; `ScopeField.svelte` passes `challenge.targetEntityId` as `keep` through `entityOptions` → `entityRecords`. Both now retain a current value that has lost its `isBoss` flag.
- The two section `warn`s mirror the recipe editor's stale-flag warning and surface drift that was previously silent. Placement in `challenge.ts`'s condition `warn` is well-ordered (goal → damage-type → boss-flag), and the boss check is safely gated by `typeBossOnly`, so the `staticData.enemies?.[targetEntityId]` lookup can't produce a false positive on a DamageType-scoped challenge (id-collision on `0`) since `bossOnly` is false there.
- Comments are appropriately concise and explain the *why* (the `keep` exception the warn backstops).

### Test coverage

- `reference.test.ts` covers both `bossEnemyOptions(keep)` and `entityOptions(Enemy, true, keep)` keeping a non-boss current value visible, plus the negative (dropping `keep` excludes it again).
- `zone.test.ts` and `challenge.test.ts` cover the new section `warn`s: flagged when the boss flag is stale, null when intact or unassigned.
- Frontend-only change; no backend/battle-logic mirroring required.

---

PR #2013, opened 2026-07-15
Source the gateway prerequisite picker from the live draft tiers (#1997)

### LINKED ISSUE #1997: Gateway prerequisite picker can't offer tiers added this session, though the save pipeline explicitly supports the remap
## Problem

The `ProgressionStore` save orchestration was deliberately built to remap a gateway prerequisite pointing at a brand-new (negative-id) tier — step 8 resolves prerequisite ids via `resolveId(pid, profIdMap)` (`UI/src/routes/admin/workbench/progression/progression-store.svelte.ts:559-564`), the step-1 filter `prerequisiteIds.every((id) => id >= 0)` exists solely for that case (`:423-427`), and `progression-store.test.ts` exercises `addPrerequisite(0, newTierId)` directly.

But the only UI that adds prerequisites builds its options from **last-saved** reference data: `GatewaysEditor` uses `reference.proficiencyOptions()` (`progression/GatewaysEditor.svelte:57-60`), i.e. `staticData.proficiencies`, so an unsaved tier never appears and the remap path is unreachable from the UI.

**Failure scenario:** an author creates tier B on path P and wants tier C gated by B in the same save — B is missing from the "Add prerequisite" list, forcing a save-first workflow that contradicts the editor's whole-graph one-save design. Secondary: prerequisite chips label by last-saved name (`reference.proficiencyName`), so a renamed-but-unsaved tier shows its stale name.

## Fix direction

Source the options (and chip labels) from `store.profs` — the in-session draft graph — excluding retired tiers and the tier itself, matching how `PathDetail`/`TierDetail` already override `{ proficiencies: store.profs }` for the retire-confirm. Extend the existing GatewaysEditor tests with an unsaved-tier prerequisite case.

ginderjeremiah:
## Summary

`GatewaysEditor`'s "Add prerequisite" dropdown and chip labels were built from `reference.proficiencyOptions()` / `reference.proficiencyName()`, which only reflect the last-saved `staticData.proficiencies` snapshot. A tier added (or renamed) earlier in the same Workbench session was invisible to the picker, even though the save pipeline (`progression-store`'s `resolveId` prerequisite remap) already supports gating on a brand-new, not-yet-persisted tier.

- Both the add-options list and the chip labels now read from `store.profs` — the in-session draft catalogue — instead of the last-saved reference snapshot. This mirrors the `{ proficiencies: store.profs }` override `PathDetail`/`TierDetail` already use for their retire-confirm dialogs.
- Options now exclude retired tiers directly (rather than relying on the reference layer's retired-tag convention), matching the fix direction in the issue.

## An additional latent bug found and fixed in the same PR

Making draft tiers selectable surfaced a real collision: the picker's placeholder sentinel value was `-1`, which is also the very first id `ProgressionStore.nextId` hands a brand-new tier. Once unsaved tiers became selectable options, a same-session tier with id `-1` would collide with the placeholder (Svelte even throws a duplicate-key error in that case). Moved the sentinel to `-Infinity`, a value outside the real id range (persisted ids are `>= 0`; draft ids count down from `-1`). Covered by a new test that specifically uses a draft tier with id `-1`.

## Test plan

- [x] Extended `TierDetail.test.ts`'s existing gateway-picker test to source proficiencies from `store.profs` instead of `staticData.proficiencies`
- [x] Added a test asserting an unsaved (negative-id) tier is offered as a prerequisite option and is selectable (including the id `-1` collision case)
- [x] Added a test asserting a retired tier is excluded from the options
- [x] `npm run test:unit:ci` — all 3716 tests pass
- [x] `npm run lint` — clean (eslint + svelte-check)

Closes #1997

---
_Generated by [Claude Code](https://claude.ai/code/session_016tkSz9Rx9saKfxShA9jYrj)_

### DIFF
--- a/UI/src/routes/admin/workbench/progression/GatewaysEditor.svelte
+++ b/UI/src/routes/admin/workbench/progression/GatewaysEditor.svelte
@@ -12,7 +12,7 @@
 		<div class="chips">
 			{#each tier.prerequisiteIds as prereqId (prereqId)}
 				<span class="chip">
-					{reference.proficiencyName(prereqId) ?? `#${prereqId}`}
+					{prereqName(prereqId)}
 					<span class="star">✦</span>
 					<button
 						type="button"
@@ -39,7 +39,6 @@
 
 <script lang="ts">
 import WorkbenchIcon from '../WorkbenchIcon.svelte';
-import { reference } from '../reference.svelte';
 import type { ProgressionStore } from './progression-store.svelte';
 import type { WorkbenchProficiency } from './types';
 import ProgSelect from './ProgSelect.svelte';
@@ -51,12 +50,22 @@ interface Props {
 
 const { store, tier }: Props = $props();
 
-const ADD_PLACEHOLDER = -1;
+// Draft (never-saved) tiers count down from -1 (ProgressionStore.nextId), so the placeholder can't
+// live in that range without colliding with a real unsaved tier's id.
+const ADD_PLACEHOLDER = -Infinity;
 let addPick = $state(ADD_PLACEHOLDER);
+
+/**
+ * Sourced from the store's live draft tiers (not the last-saved `reference` snapshot) so a tier
+ * added or renamed earlier in this session is offered/labelled correctly (#1997).
+ */
+const prereqName = (id: number) => store.profs.find((p) => p.id === id)?.name || `#${id}`;
 
 const addOptions = $derived([
 	{ value: ADD_PLACEHOLDER, text: '+ Add prerequisite…' },
-	...reference.proficiencyOptions().filter((o) => o.value !== tier.id && !tier.prerequisiteIds.includes(o.value))
+	...store.profs
+		.filter((p) => p.id !== tier.id && !p.retiredAt && !tier.prerequisiteIds.includes(p.id))
+		.map((p) => ({ value: p.id, text: p.name || 'Unnamed tier' }))
 ]);
 
 const onAddPrereq = (value: number) => {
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Source the gateway prerequisite picker from live draft tiers (#1997)

**Verdict:** Approve — the change correctly makes the "Add prerequisite" picker and chip labels read from the in-session draft catalogue (`store.profs`) so unsaved/renamed tiers are selectable, unblocking the remap path the save pipeline already supported. (Formal event is `COMMENT` because I'm the PR author; this is an approve.)

### What's good
- Fix matches the issue's stated direction exactly: options and chip labels now source from `store.profs`, filtering the tier itself, retired tiers, and already-added prerequisites. This mirrors the `{ proficiencies: store.profs }` override `PathDetail`/`TierDetail` already use.
- Good catch on the latent `-1` placeholder collision. Moving `ADD_PLACEHOLDER` to `-Infinity` is sound: persisted ids are `>= 0` and draft ids count down from `-1`, so `-Infinity` sits safely outside the real id range and can't collide with a keyed `{#each}` option. The round-trip is safe too — `Number("-Infinity") === -Infinity`, so `ProgSelect`'s `+e.currentTarget.value` and the `value === ADD_PLACEHOLDER` guard both hold.
- Dead code removed cleanly (the now-unused `reference` import).
- Comments are concise and explain the non-obvious `-Infinity` choice for a future maintainer, which is exactly the kind of "don't naively simplify this" note that prevents a regression.

### Test coverage
Adequate and well-targeted per the frontend testing rules:
- The existing chip/add test was migrated off `staticData.proficiencies` to `store.profs`, keeping the add-and-remove coverage.
- New test asserts an unsaved negative-id tier is offered and selectable — deliberately using id `-1` to also pin the placeholder-collision regression.
- New test asserts retired tiers are excluded from options.

---

PR #2014, opened 2026-07-15
Self-heal Skills Compare-vs Toughness when maxToughness shrinks

### LINKED ISSUE #1998: Skills Compare-vs Toughness clamps only at write time — a shrunken maxToughness leaves the stored value (and a removed preset) stale
## Problem

`SkillsView.setToughness`/`selectPreset` clamp the compare toughness against `maxToughness` only at the moment of the write (`UI/src/routes/game/screens/skills/skills-view.svelte.ts:473-483`), but `maxToughness` is derived from zone/reference data. If a mid-session zone relocation or reference-data refresh shrinks the preset range while the screen is mounted, the stored `toughness` can exceed the new max: the slider renders pinned at its new max while every effective-DPS/burst number and the sort keep using the stale, now-unreachable higher value, and `selectedPresetKey` can point at a removed pill.

## Fix direction

Make the consumed value self-healing — e.g. a `$derived` effective toughness (`Math.min(this.toughness, this.maxToughness)`) used by the formulas and slider, and clear `selectedPresetKey` when its preset disappears. Unit test: shrink presets after a higher stored value and assert the displayed numbers re-clamp.

ginderjeremiah:
## Summary

`SkillsView.setToughness`/`selectPreset` only clamp the Compare-vs Toughness against `maxToughness` at the moment of the write (`UI/src/routes/game/screens/skills/skills-view.svelte.ts`). Since `maxToughness` is itself derived from the current zone's live reference data, a mid-session zone relocation or reference-data refresh that shrinks the preset range could leave the stored `toughness` above the new max — and `selectedPresetKey` pointing at a preset that no longer exists.

This fixes it by introducing two self-healing derived reads that every consumer now goes through instead of the raw stored state:

- `effectiveToughness` — `toughness` re-clamped to the live `maxToughness` ceiling on every read, backed by a pure `clampToughness(toughness, maxToughness)` helper.
- `activePresetKey` — `selectedPresetKey` resolved against the live `comparePresets`, backed by a pure `resolveActivePresetKey(selectedPresetKey, presets)` helper, so a pill for a preset that's disappeared no longer renders "selected".

The raw `toughness`/`selectedPresetKey` state fields are unchanged (still clamped/validated at write time, matching existing tests) — only what the slider, the effective-dps/burst formulas, the sort, and the pill highlight actually *consume* changed.

Both helpers were pulled into the file's existing "pure helpers (no reactive state — unit-tested directly)" section (alongside `sortMetrics`/`zoneSpawnLevel`) rather than tested only through the reactive view, since the view's own test harness mocks `staticData`/`playerManager` as plain (non-`$state`) objects — a `$derived` chain rooted in them can't be exercised reactively there without constructing a fresh `SkillsView()`, so the pure-helper split is what makes the exact clamp/preset-resolution logic directly testable.

## Changes

- `skills-view.svelte.ts`: add `clampToughness`/`resolveActivePresetKey` pure helpers and `effectiveToughness`/`activePresetKey` derived fields; route `railList`'s sort and `effective()` through `effectiveToughness`.
- `CompareBar.svelte`: slider value/display and pill "on" state now read `effectiveToughness`/`activePresetKey`.
- `SkillInspector.svelte`: the raw-damage note's "target has no toughness" check now reads `effectiveToughness`.
- `skills-view.test.ts`: unit tests for the new pure helpers, plus a view-level test simulating a stored value/preset key left stale by a shrunken ceiling.

## Test plan

- [x] `npm run test:unit:ci` — full suite: 299 files / 3719 tests passed, no coverage-threshold regressions
- [x] New/updated tests specifically cover: `clampToughness`, `resolveActivePresetKey`, and `effectiveToughness`/`activePresetKey` re-clamping a stale stored value

Closes #1998

---
_Generated by [Claude Code](https://claude.ai/code/session_01AFTqFfELeN3srzeTdVE9P9)_

### DIFF
--- a/UI/src/routes/game/screens/skills/CompareBar.svelte
+++ b/UI/src/routes/game/screens/skills/CompareBar.svelte
@@ -6,7 +6,7 @@
 				<button
 					type="button"
 					class="epill"
-					class:on={view.selectedPresetKey === preset.key}
+					class:on={view.activePresetKey === preset.key}
 					class:boss={preset.isBoss}
 					onclick={() => view.selectPreset(preset)}
 				>
@@ -24,11 +24,11 @@
 			type="range"
 			min="0"
 			max={view.maxToughness}
-			value={view.toughness}
+			value={view.effectiveToughness}
 			aria-label="Compare-vs enemy toughness"
 			oninput={(e) => view.setToughness(+e.currentTarget.value)}
 		/>
-		<span class="vs-defval">TGH <b>{Math.round(view.toughness)}</b></span>
+		<span class="vs-defval">TGH <b>{Math.round(view.effectiveToughness)}</b></span>
 	</div>
 </div>
 
--- a/UI/src/routes/game/screens/skills/SkillInspector.svelte
+++ b/UI/src/routes/game/screens/skills/SkillInspector.svelte
@@ -138,7 +138,7 @@ const rawNote = $derived.by(() => {
 	// Critical row and the crit-inclusive effective hit; omitted when no crit can occur.
 	const critBonus = view.critBonus(id);
 	const critPart = critBonus > 0 ? ` · +${fmt(critBonus)} crit` : '';
-	if (view.toughness <= 0) {
+	if (view.effectiveToughness <= 0) {
 		return `${raw} raw damage${critPart} · target has no toughness`;
 	}
 	const mitigated = fmt(view.mitigatedAmount(id));
--- a/UI/src/routes/game/screens/skills/skills-view.svelte.ts
+++ b/UI/src/routes/game/screens/skills/skills-view.svelte.ts
@@ -121,6 +121,22 @@ export function zoneSpawnLevel(zone: IZone): number {
 	return (zone.levelMin + zone.levelMax) / 2;
 }
 
+/** Clamp a stored Compare-vs Toughness to the live slider ceiling. `toughness` is only clamped at
+ *  the moment it's written (a slider drag or preset pick), so a `maxToughness` that later shrinks
+ *  (e.g. a zone relocation or reference-data refresh while the screen stays mounted) would otherwise
+ *  leave the stored value pinned above the new range — every consumer must read through this instead
+ *  of the raw stored value. */
+export function clampToughness(toughness: number, maxToughness: number): number {
+	return Math.min(toughness, maxToughness);
+}
+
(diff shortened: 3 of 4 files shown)


ginderjeremiah (commented):
## Review: Self-heal Skills Compare-vs Toughness / preset when `maxToughness` shrinks (#1998)

**Verdict:** Approve _(posted as a comment — reviewer is the PR author, so GitHub won't accept a formal APPROVE; the intent is approve)_ — the fix cleanly self-heals the stale-value cases the issue describes, routes every consumer through the derived reads, and is well-tested.

The change addresses the issue exactly: `toughness`/`selectedPresetKey` remain clamped/validated only at write time, but consumption now flows through two self-healing `$derived` reads — `effectiveToughness` (`Math.min(toughness, maxToughness)`) and `activePresetKey` (resolved against live `comparePresets`). Both are backed by exported pure helpers (`clampToughness`, `resolveActivePresetKey`), which is the right seam for direct unit testing given the view's non-reactive test mocks.

I verified every raw read migrated — the only remaining references to `toughness`/`selectedPresetKey` are the state declarations, the derived reads themselves, and the write-time setters. `effective()` (and everything routing through it: `effectiveDps`, `combinedEffectiveDps/Burst`, `mitigatedAmount`), the `railList` sort, the slider value/display, the pill highlight, and the `SkillInspector` raw-note check all consume the derived values now.

Design-wise this is the right call and aligns with `frontend.md`'s "prefer reactive/derived over mutating state" guidance — deriving rather than mutating `selectedPresetKey`/`toughness` out from under a caller preserves the user's setting across a transient zone change and just re-clamps per zone, rather than destroying it.

### Test coverage
Meets the project's frontend testing rules well:
- `clampToughness` and `resolveActivePresetKey` unit-tested directly (within-range, above-shrunken-ceiling, and removed-preset/null cases).
- A view-level test assigns past the write-time guard to simulate a shrunken `maxToughness`, asserting `effectiveToughness === maxToughness` and `activePresetKey === null` — the exact scenario from the issue.

### What's good
- Minimal, surgical change with no new store; pure-helper split keeps the exact logic testable.
- Clear, appropriately-scoped doc comments on the helpers and derived fields explaining the "clamps only at write time" hazard.
- No hardcoded values, no security/layering concerns (frontend-only view logic).

No blocking findings, no should-fixes, no nits. CI is green (frontend/verify/e2e chromium+firefox passing, backend jobs correctly skipped) with only the webkit e2e job still finishing.

---

PR #2015, opened 2026-07-15
Retry the offline-progress fetch once before skipping loop-mode reconciliation (#1999)

### LINKED ISSUE #1999: Welcome-back flow skips loop-mode reconciliation when the offline-progress fetch fails
## Problem

`WelcomeBackView.run` calls `reconcileMode(progress.autoChallengeBoss)` only when the fetch succeeded (`UI/src/routes/game/welcome-back/welcome-back-view.svelte.ts:41-53`); a failed `GetOfflineProgress` (null result) falls straight through to `enter()` with no reconciliation.

**Failure scenario:** a returning boss-farmer whose offline check transiently errors enters the game with `autoFight` silently off while the backend still persists `AutoChallengeBoss = true`. Until some later sync fires (`setAutoFight`, a `returnToIdle`), the client and server disagree about the loop mode — and the *next* disconnect's offline simulation resumes as a boss-farmer for a session that actually idle-farmed, crediting the wrong loop.

## Fix direction

Reconcile from a later authoritative read when the progress fetch fails (the player payload the boot path already pulls carries the flag), or retry the fetch before entering. Unit test the error path asserting the mode ends up consistent.

ginderjeremiah:
## Summary

`WelcomeBackView.run()` only called `reconcileMode(progress.autoChallengeBoss)` when `GetOfflineProgress` succeeded. A failed (transiently errored) fetch fell straight through to `enter()` with no reconciliation, so a returning boss-farmer would enter with client-side auto-fight silently off while the backend still persisted `AutoChallengeBoss = true` — a client/server loop-mode disagreement that the *next* disconnect's offline simulation would then resume incorrectly.

## Fix

`run()` now retries `fetchProgress()` once when the first call returns `null`, before giving up on reconciliation. This is safe because `OfflineProgressService.SimulateProgress` (backing `GetOfflineProgress`) always re-anchors the server's `LastActivity` clock, even on a sub-threshold call — so an immediate retry either recovers the real summary (if the first attempt's response was merely lost in transit) or lands a cheap, side-effect-free empty-progress result (if the first attempt genuinely never reached the server). Either way it's a normal outcome, not a duplicate-reward risk.

`cancel()` is still honored between the first fetch and the retry, matching the existing cancellation contract (#1807).

### A note on the issue's suggested fallback

The issue also suggested reconciling "from a later authoritative read (the player payload the boot path already pulls carries the flag)". I checked this and it isn't accurate: the `Login/Status` DTO (`PlayerData`, the payload `resyncPlayer`/`refreshPlayer` pulls) doesn't carry `AutoChallengeBoss` anywhere — that flag is currently only ever sent to the frontend via `GetOfflineProgress`'s `OfflineProgressModel`. Threading it onto `PlayerData` would be a reasonable follow-up if a stronger fallback is ever wanted, but it's a larger, separate change (new backend field, DTO plumbing, frontend type update) than this bug fix warrants, so I went with the retry-based fix instead, which is fully contained to `WelcomeBackView`.

## Test plan

- [x] Added/updated unit tests in `welcome-back-view.test.ts`: retry-succeeds reconciles mode, retry-fails-twice still skips reconciliation, no retry on first-attempt success, and a new cancellation case (cancelled between the first fetch and the retry).
- [x] `npm run lint` — clean.
- [x] `npm run test:unit:ci` — 299 files / 3722 tests passed.

Closes #1999

---
_Generated by [Claude Code](https://claude.ai/code/session_014MbEnq7zBL73Ve6ZLc1Dyg)_

### DIFF
--- a/UI/src/routes/game/welcome-back/welcome-back-view.svelte.ts
+++ b/UI/src/routes/game/welcome-back/welcome-back-view.svelte.ts
@@ -36,11 +36,22 @@ export class WelcomeBackView {
 	 * Runs the gate flow. Fetches the offline summary, reconciles the persisted loop mode (always — even a
 	 * sub-threshold return should restore a boss-farmer's toggle), then passes straight through to the game
 	 * when there is no progress, or re-syncs state and shows the summary gate when there is.
+	 *
+	 * A failed fetch is retried once before giving up: `GetOfflineProgress` re-anchors the server's away-time
+	 * clock on every call (even a sub-threshold one), so an immediate retry is safe and turns a transient
+	 * failure into a normal empty-progress result instead of silently skipping reconciliation (#1999) — a
+	 * boss-farmer would otherwise enter with auto-fight off while the backend still thinks they're farming.
 	 */
 	async run(): Promise<void> {
-		const progress = await this.deps.fetchProgress();
+		let progress = await this.deps.fetchProgress();
 		if (this.cancelled) {
 			return;
+		}
+		if (!progress) {
+			progress = await this.deps.fetchProgress();
+			if (this.cancelled) {
+				return;
+			}
 		}
 
 		if (progress) {
--- a/UI/src/tests/routes/game/welcome-back/welcome-back-view.test.ts
+++ b/UI/src/tests/routes/game/welcome-back/welcome-back-view.test.ts
@@ -50,15 +50,45 @@ describe('WelcomeBackView', () => {
 		expect(view.summary).toBeNull();
 	});
 
-	it('enters the game and skips reconciliation when the offline check fails', async () => {
+	it('enters the game and skips reconciliation when the offline check fails twice in a row', async () => {
 		const view = makeView(null);
 
 		await view.run();
 
+		// One retry (#1999): a transient failure gets a second chance before reconciliation is skipped.
+		expect(deps.fetchProgress).toHaveBeenCalledTimes(2);
 		expect(deps.reconcileMode).not.toHaveBeenCalled();
 		expect(deps.resyncPlayer).not.toHaveBeenCalled();
 		expect(deps.enterGame).toHaveBeenCalledTimes(1);
 		expect(view.phase).toBe('entered');
+	});
+
+	it('retries once and reconciles mode when the retry succeeds (#1999)', async () => {
+		deps = {
+			fetchProgress: vi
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: retry the offline-progress fetch once before skipping loop-mode reconciliation (#1999)

**Verdict:** Approve (posted as a comment — GitHub won't accept a self-approval on my own PR) — correct, safe, and well-tested fix; no blocking or should-fix findings.

I verified the fix's central safety claim against the backend. `OfflineProgressService.SimulateProgress` re-anchors `LastActivity` and persists on **both** paths — the sub-threshold early return (`Game.Application/Services/OfflineProgressService.cs:118-123`) and the full simulation (`:190-191`). So if the first `GetOfflineProgress` actually reached the server but its response was lost, the immediate retry sees near-zero away time, falls into the sub-threshold branch, and returns an empty re-claim — no double-credited rewards. If the first attempt never reached the server, the retry runs it fresh. Both outcomes are safe, exactly as the PR description states.

I also confirmed the description's justification for *not* reconciling from a later authoritative read: `PlayerData` (the boot payload) genuinely does not carry `AutoChallengeBoss` (`Game.Api/Models/Player/PlayerData.cs`), so the flag is only ever delivered via `GetOfflineProgress`. Choosing the contained retry over new DTO plumbing is the right call for a `scope: small` bug fix, and the issue explicitly offered "retry the fetch before entering" as an accepted fix direction.

Cancellation is preserved correctly: the new retry re-checks `this.cancelled` after the second fetch (`:52-54`), consistent with the existing cancellation contract (#1807).

### Test coverage

Thorough and matching the frontend testing guidelines — the mirrored source path already exists and the new cases cover the full decision space:
- retry succeeds → mode reconciled;
- both fetches fail → reconciliation skipped, game still entered;
- first fetch succeeds → no retry;
- cancelled between the first fetch and its retry → no second fetch, no enter.

### Residual (not blocking, and by accepted design)

When *both* fetches fail the client can still enter with the loop mode unreconciled — the fix narrows the window (two consecutive transient failures) rather than closing it. Fully closing it would need the flag threaded onto the boot payload, which the PR reasonably defers. No action needed here; noting it only for the record.

### What's good

Tightly scoped, the safety reasoning is documented at the call site rather than left implicit, and the tradeoff against the issue's alternative fix direction is explained honestly. Nice work.

---

PR #2017, opened 2026-07-15
Self-heal corrupted player/session cache blobs on deserialization failure

### LINKED ISSUE #1924: Cached player/session blobs have no deserialization self-heal — a PlayerCacheModel shape change locks active players out for up to the 48h TTL
`PlayerRepository.GetPlayer` only falls through to Postgres when the cached value is *absent* (`if (model is null)`, `Game.DataAccess/Repositories/PlayerRepository.cs:66`); a blob that exists but no longer deserializes throws `JsonException` out of `RedisService.Get<T>` (`Game.Infrastructure/Cache/Redis/RedisService.cs:34-38`) on every load. `PlayerCacheModel` is all `required` members — deliberately, so projections can't omit a field — which means System.Text.Json hard-fails on any blob written before a member was added/renamed.

Since Redis is the rehydration source of truth and the key carries the 48h sliding TTL, one additive model change deployed to a live fleet makes every cached player's connect/load throw until the key ages out or an operator flushes Redis — and the failed read never slides the TTL, so it's a hard lockout of active players despite Postgres holding a durable copy. `SessionStore.GetSession` (`Game.DataAccess/Repositories/SessionStore.cs:35-46`, the `PlayerState` blob) has the same gap. (`RefreshTokenStore.Consume` self-clears only by accident — `GetDelete` removes the key before the parse fails.)

The codebase already has the precedent: `RedisService.HashGetAllIfExists` (`RedisService.cs:136-150`) deliberately treats a wrong-representation progress key as a miss and **deletes it** "rather than needing an explicit migration step, since Postgres remains the durable copy." The same reasoning applies verbatim to the player/session blobs but isn't implemented there. Given how actively the model still evolves, this will very likely bite during development.

**Fix direction:** in `PlayerRepository.GetPlayer` and `SessionStore.GetSession`, catch the deserialization failure, log loudly, delete the key, and fall through to the DB miss-reload path (player) / rehydration path (session). Losing the session's in-flight battle is strictly better than a thrown load, and it matches the progress hash's documented self-heal contract.

ginderjeremiah:
## Summary

`PlayerRepository.GetPlayer` only fell through to Postgres when the cached value was **absent**; a blob that exists but no longer deserializes (e.g. a `PlayerCacheModel` shape change — all members are `required`, so any additive/renamed field hard-fails `System.Text.Json` on an older blob) threw a `JsonException` on every load. Since Redis is the rehydration source of truth for the player aggregate and carries a 48h sliding TTL, one such change deployed to a live fleet would lock every cached player's connect/load out until the key aged out or an operator flushed Redis — despite Postgres holding a durable copy. `SessionStore.GetSession` (the `PlayerState` blob) had the same gap.

This mirrors a self-heal pattern the codebase already applies elsewhere: `RedisService.HashGetAllIfExists` treats a wrong-representation progress-hash key as a miss and deletes it, "rather than needing an explicit migration step, since Postgres remains the durable copy."

## Changes

- `PlayerRepository.GetPlayer`: catch a deserialization failure on the cached blob, log it, delete the corrupted key, and fall through to the existing DB-miss reload path (which then re-caches a valid blob).
- `SessionStore.GetSession`: catch a deserialization failure, log it, delete the corrupted key, and return `null` — a session miss is reconstructable, since `SessionInitializer` rehydrates it in-memory from the token's selected-player claim (never a silent logout).
- Both repositories now take an `ILogger<T>` so the self-heal is logged loudly rather than silently, matching the "log loudly" fix direction in the issue.
- Updated the three test call sites that construct `PlayerRepository` directly to pass the new logger dependency.

## Test plan

- [x] `dotnet build` — succeeds with 0 warnings/errors.
- [x] Added `PlayerCacheEvictionTests.GetPlayer_OnCorruptCacheEntry_DeletesKeyAndReloadsFromDb` — writes malformed JSON directly into the player's Redis key, asserts `GetPlayer` still returns the player (from the DB) instead of throwing, and that the key is re-cached with a fresh TTL.
- [x] Added `SessionCacheEvictionTests.GetSession_OnCorruptCacheEntry_DeletesKeyAndReturnsNull` — same setup for the session key, asserts `GetSession` returns `null` and the corrupted key is deleted rather than left behind.
- [x] `dotnet test --no-build` on `Game.Application.Tests` (the affected assembly): 981/981 passed, including the two new tests and the three updated `PlayerRepository`-construction call sites.

Closes #1924

---
_Generated by [Claude Code](https://claude.ai/code/session_016H8HEBGpjjnK6GtqB4biG3)_

### DIFF
--- a/Game.Application.Tests/Services/OfflineProgressServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/OfflineProgressServiceIntegrationTests.cs
@@ -13,6 +13,7 @@ using Game.TestInfrastructure.Base;
 using Game.TestInfrastructure.Fixtures;
 using Game.TestInfrastructure.Helpers;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Xunit;
 
 namespace Game.Application.Tests.Services
@@ -176,7 +177,8 @@ namespace Game.Application.Tests.Services
                 scope.ServiceProvider.GetRequiredService<PlayerUpdateBatch>(),
                 scope.ServiceProvider.GetRequiredService<IItems>(),
                 scope.ServiceProvider.GetRequiredService<IItemMods>(),
-                scope.ServiceProvider.GetRequiredService<ISkills>());
+                scope.ServiceProvider.GetRequiredService<ISkills>(),
+                scope.ServiceProvider.GetRequiredService<ILogger<PlayerRepository>>());
 
             var progressRepo = scope.ServiceProvider.GetRequiredService<IPlayerProgressRepository>();
             var brokenOfflineService = new OfflineProgressService(
--- a/Game.Application.Tests/Services/PlayerCacheEvictionTests.cs
+++ b/Game.Application.Tests/Services/PlayerCacheEvictionTests.cs
@@ -129,6 +129,33 @@ namespace Game.Application.Tests.Services
             Assert.True(ttl > TtlFloor, $"expected the reload to re-cache with a TTL, got {ttl}");
         }
 
+        [Fact]
+        public async Task GetPlayer_OnCorruptCacheEntry_DeletesKeyAndReloadsFromDb()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+            var user = await TestDataSeeder.CreateUserAsync(context);
+            var playerEntity = await TestDataSeeder.CreatePlayerAsync(context, user.Id, name: "Corrupted");
+
+            var playerRepo = scope.ServiceProvider.GetRequiredService<IPlayerRepository>();
+            using var multiplexer = await ConnectionMultiplexer.ConnectAsync(Containers.CacheConnectionString);
+            var db = multiplexer.GetDatabase();
+            var playerKey = PlayerKey(playerEntity.Id);
+
+            // Stand in for a corrupted/unparsable blob (e.g. a PlayerCacheModel shape change hitting an
+            // already-cached value, or a truncated write) with malformed JSON that reliably throws on deserialize.
+            await db.StringSetAsync(playerKey, "{not valid json");
+
+            // A corrupt entry must self-heal (delete + DB reload) rather than throw and lock the player out (#1924).
+            var player = await playerRepo.GetPlayer(playerEntity.Id);
+            Assert.NotNull(player);
+            Assert.Equal("Corrupted", player.Name);
+
+            var ttl = await WaitForTtlAsync(db, playerKey);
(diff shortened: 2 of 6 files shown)


ginderjeremiah (commented):
## Review: Self-heal corrupted player/session cache blobs on deserialization failure (#1924)

**Verdict:** Approve — implements the issue's fix direction exactly, scoped to the two affected reads, and matches the existing `HashGetAllIfExists` self-heal precedent. (Submitted as a comment because I'm the PR author; GitHub won't accept a gating verdict on one's own PR.)

### Correctness
- The `catch (JsonException)` is precisely targeted. `RedisService.Get<T>` runs `val.Deserialize<T>()`, and the `Deserialize` extension (`Game.Core/Extensions.cs`) returns `default` when the string is null — so an **absent** key still returns null and falls through to the DB reload without ever entering the catch. Only a present-but-unparsable blob throws, which is exactly the corruption case. `System.Text.Json` throws `JsonException` both for malformed JSON and for a missing `required` member, so the `PlayerCacheModel` shape-change scenario the issue describes is genuinely covered.
- `PlayerRepository.GetPlayer` deletes the key and falls through to the existing DB miss-reload (which re-caches). The explicit `Delete` isn't redundant despite the later `SetAndForget`: on the corrupt-blob-for-a-since-deleted-player edge, `GetPlayerModelFromDb` returns null and the method returns without re-caching, so the delete is what actually clears the poisoned key. Good.
- `SessionStore.GetSession` deletes and returns null (a miss), correctly leaving re-population to the socket write path rather than writing the cache off the HTTP thread — consistent with the documented session-rehydration model. The player/session asymmetry (re-cache vs. treat-as-miss) is intentional and the comments explain why.

### Layering / DI
- Both classes are data-tier repositories; the self-heal is an infrastructure concern with no domain logic, so it belongs here. Adding `ILogger<T>` is fine — both are DI-registered (`AddScoped`), so the logger auto-resolves with no wiring change. All three direct constructions (the test doubles) were updated; there are no other un-updated call sites.

### Test coverage
- Two integration tests added, correctly using the out-of-process Redis path (not unit tests) per the testing guidelines for a Redis-coupled class. `GetPlayer_OnCorruptCacheEntry_...` asserts DB reload + re-cache with a fresh TTL; `GetSession_OnCorruptCacheEntry_...` asserts null + key deleted. Both mirror the existing eviction-test style. Adequate.

### What's good
- Faithful to the issue and to the codebase's own precedent, loud `LogError` with useful context (id + key), and the comments justify the non-obvious choices without over-explaining. Clean, small, well-tested.

No blocking items, should-fixes, or nits. Approving; will merge once `test-backend` and the remaining e2e job finish green.

---

PR #2018, opened 2026-07-15
Combine hit-path GET+EXPIRE into a single GETEX

### LINKED ISSUE #1925: Hot-path Redis Lua scripts are re-sent as full EVAL text on every call; hit-path GET+EXPIRE pairs could be one GETEX
## Investigation notes before implementing

This issue bundles two claims. I verified both against StackExchange.Redis 3.0.0 (the version this repo pins) by decompiling the assembly (`ilspycmd`), and they don't hold equally:

**1. "Lua scripts are re-sent as full EVAL text on every call" — not accurate for this codebase's usage.**

`RedisDatabase.ScriptEvaluateAsync(string script, ...)` — the raw-string overload every method in `RedisService` uses — already self-upgrades to `EVALSHA` after the *first* call per server endpoint. Its `ScriptEvalMessage.GetMessages` checks `ServerEndPoint.GetScriptHash(script, ...)` (a `Hashtable` keyed by the literal script text); on a miss it transparently issues a `SCRIPT LOAD` ahead of the eval, and the load's result processor (`ScriptLoadProcessor.SetResultCore` → `ServerEndPoint.AddScript`) populates that hashtable. Every subsequent call with the same script string sends `EVALSHA`, not the full text. Since each `RedisService` method uses a fixed string literal per script and the connection multiplexer is a long-lived singleton, in steady state these are already `EVALSHA` after the first invocation — no behavioral change needed.

I also checked whether switching to the suggested `LuaScript.Prepare`/`LoadedLuaScript.Load` path would do better: it wouldn't. `LoadedLuaScript.Evaluate`/`EvaluateAsync` (in this version) internally call `db.ScriptEvaluate(ExecutableScript, keys, args, flags)` — the **string** overload, passing the lua text rather than `Hash` — so it routes through the exact same auto-upgrading mechanism already in play. Moving to that API would add ceremony (server-scoped `Load()`, reconnect handling) for zero measurable improvement over the current code.

**2. "GET+ExpireAndForget pairs could be one GETEX" — accurate and implemented.**

`PlayerRepository.GetPlayer` and `SessionStore.GetSession` did pay a separate awaited `Get` plus a fire-and-forget `ExpireAndForget` on every sliding-TTL cache hit. Added `ICacheService.GetAndRefreshExpiry`/`GetAndRefreshExpiry<T>` backed by `IDatabase.StringGetSetExpiryAsync` (GETEX), and switched both call sites to it — one round trip instead of two on a hit, and it simplifies both methods by dropping their separate "refresh on hit" branch (GETEX is already a no-op on a miss, matching `ExpireAndForget`'s existing behavior).

Scope note: `PlayerProgressRepository.GetCachedProgress` has the same shape (`HashGetAllIfExists` + `ExpireAndForget`) but that key is a Redis **hash**, not a string — GETEX doesn't apply to hashes, so folding that one into a single round trip would need a custom Lua script (TYPE/HGETALL+PEXPIRE in one call, similar to `HashGetAllIfExists`'s existing script). Left out of this PR as a separate, larger change; happy to file a follow-up if useful.

PR incoming with the GETEX fix and tests; closing the Lua/EVALSHA half of this issue as already-fine rather than making a no-op change.

ginderjeremiah:
## Summary

- `PlayerRepository.GetPlayer` and `SessionStore.GetSession` paid a separate awaited `Get` plus a fire-and-forget `ExpireAndForget` on every sliding-TTL cache hit. Added `ICacheService.GetAndRefreshExpiry`/`GetAndRefreshExpiry<T>` (backed by Redis `GETEX` via `IDatabase.StringGetSetExpiryAsync`) and switched both call sites to it — one Redis round trip instead of two on a hit, and it removes each method's separate "refresh on hit" branch since `GETEX` is already a no-op on a miss (matching `ExpireAndForget`'s existing behavior).
- The issue's other claim — that the hot-path Lua scripts in `RedisService` resend the full `EVAL` text on every call — turned out not to hold for this codebase. I verified by decompiling StackExchange.Redis 3.0.0 (the version this repo pins): the string-based `ScriptEvaluateAsync` overload already self-upgrades to `EVALSHA` after the first call per server endpoint via its own script-hash cache (`ServerEndPoint.GetScriptHash`/`AddScript`, populated by an internal `SCRIPT LOAD` pre-flight), and the issue's suggested `LuaScript.Prepare`/`LoadedLuaScript` alternative routes through that exact same string-based call internally in this library version — so switching to it wouldn't change anything. Left the Lua scripts as-is; full details are in the [issue comment](https://github.com/ginderjeremiah/GameServer/issues/1925#issuecomment).

## Why

Halves Redis command volume on the two highest-frequency sliding-TTL cache reads (player load on every socket connection/reconnect, session load on every request), and slightly simplifies both call sites.

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test` (`Game.Application.Tests`, `Game.Api.Tests`) — all green, including new `RedisServiceIntegrationTests` coverage for `GetAndRefreshExpiry` (hit refreshes TTL + returns value; miss returns null and doesn't create the key) and existing `PlayerCacheEvictionTests`/`SessionCacheEvictionTests` sliding-TTL integration tests (unchanged behavior, now exercised via GETEX)

## Follow-up noted (not in this PR)

`PlayerProgressRepository.GetCachedProgress` has the same GET+ExpireAndForget shape, but its key is a Redis **hash**, not a string, so `GETEX` doesn't apply — folding that one down would need a small custom Lua script (TYPE/HGETALL+PEXPIRE in one round trip, similar to the existing `HashGetAllIfExists` script). Left out as a separate, smaller-scope follow-up.

Closes #1925

https://claude.ai/code/session_01T1W4pAB5FNdtv2FsXF7VNu

---
_Generated by [Claude Code](https://claude.ai/code/session_01T1W4pAB5FNdtv2FsXF7VNu)_

### DIFF
--- a/Game.Abstractions/Infrastructure/ICacheService.cs
+++ b/Game.Abstractions/Infrastructure/ICacheService.cs
@@ -28,6 +28,13 @@
         public Task Set(string key, string? value, TimeSpan expiry, CancellationToken cancellationToken = default);
         public Task Set<T>(string key, T value, TimeSpan expiry, CancellationToken cancellationToken = default);
         /// <summary>
+        /// Reads <paramref name="key"/> and resets its TTL to <paramref name="expiry"/> in a single round trip
+        /// (a no-op on a missing key, like <see cref="ExpireAndForget"/>). Lets a sliding-expiration cache hit
+        /// avoid the separate awaited get followed by a fire-and-forget expire.
+        /// </summary>
+        public Task<string?> GetAndRefreshExpiry(string key, TimeSpan expiry, CancellationToken cancellationToken = default);
+        public Task<T?> GetAndRefreshExpiry<T>(string key, TimeSpan expiry, CancellationToken cancellationToken = default);
+        /// <summary>
         /// Resets the time-to-live on an existing key without awaiting the result (fire-and-forget).
         /// A no-op if the key does not exist. Lets a hot read path slide a sliding-expiration TTL
         /// without paying a round-trip.
--- a/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
@@ -427,6 +427,8 @@ namespace Game.Api.Tests.Unit
             public Task<T?> GetDelete<T>(string key, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task Set(string key, string? value, TimeSpan expiry, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task Set<T>(string key, T value, TimeSpan expiry, CancellationToken cancellationToken = default) => throw new NotSupportedException();
+            public Task<string?> GetAndRefreshExpiry(string key, TimeSpan expiry, CancellationToken cancellationToken = default) => throw new NotSupportedException();
+            public Task<T?> GetAndRefreshExpiry<T>(string key, TimeSpan expiry, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public void ExpireAndForget(string key, TimeSpan expiry) => throw new NotSupportedException();
             public void SetAndForget(string key, string? value, TimeSpan expiry) => throw new NotSupportedException();
             public void SetAndForget<T>(string key, T value, TimeSpan expiry) => throw new NotSupportedException();
@@ -497,6 +499,8 @@ namespace Game.Api.Tests.Unit
             public Task<T?> GetDelete<T>(string key, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task Set(string key, string? value, TimeSpan expiry, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task Set<T>(string key, T value, TimeSpan expiry, CancellationToken cancellationToken = default) => throw new NotSupportedException();
+            public Task<string?> GetAndRefreshExpiry(string key, TimeSpan expiry, CancellationToken cancellationToken = default) => throw new NotSupportedException();
+            public Task<T?> GetAndRefreshExpiry<T>(string key, TimeSpan expiry, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public void SetAndForget(string key, string? value, TimeSpan expiry) => throw new NotSupportedException();
             public void SetAndForget<T>(string key, T value, TimeSpan expiry) => throw new NotSupportedException();
             public void DeleteAndForget(string key) => throw new NotSupportedException();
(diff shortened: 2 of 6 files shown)


ginderjeremiah (commented):
## Review: Collapse hit-path GET+EXPIRE into a single GETEX (#1925)

**Verdict: Approve** — clean, well-scoped optimization; the code is correct, consistent with the existing `ICacheService` conventions, and thoroughly tested. (Submitted as a `COMMENT` event only because GitHub won't accept a gating verdict from the PR author — treat this as an approval.)

### Assessment

- **Correctness.** On a hit, `GETEX` (`StringGetSetExpiryAsync`) returns the value and slides the TTL in one round trip; on a miss it's a no-op returning `null`, which falls through to the DB reload (`PlayerRepository`) or the null-session path (`SessionStore`) exactly as before. Removing the separate `ExpireAndForget` branch is safe because `GETEX` already folds the refresh in. Latency is unchanged (still one awaited round trip) while command volume on the two hottest sliding-TTL reads is halved — the win the issue claimed.
- **Consistency.** Routing `GetAndRefreshExpiry` through `ObserveWrite` is the right call: it's a read-modify-write (it mutates the TTL), so it matches the documented convention already applied to `GetSet`/`GetDelete`, and it keeps the post-cancellation fault-logging behavior. The generic overload mirrors `Get<T>`/`GetDelete<T>` and handles the null (miss) case through `Deserialize<T>` identically.
- **Self-heal preserved.** The `JsonException` catch in both repositories still deletes the corrupt key and treats the read as a miss; the fact that `GETEX` refreshes the TTL before deserialize throws is moot since the key is deleted anyway. Both corrupt-entry eviction tests confirm this.
- **Interface/docs.** The new `ICacheService` XML docs are clear and accurate. This is a mechanical optimization, not an architectural change, so no `backend-persistence.md` update is warranted.

### Test coverage

Solid and following the guidelines. New `RedisServiceIntegrationTests` cover the primitive directly (hit returns value + refreshes TTL; miss returns null + creates no key). The repository-level path stays covered by the existing `PlayerCacheEvictionTests`/`SessionCacheEvictionTests` — the `OnCacheHit_RefreshesIdleTtl`, miss, and corrupt-entry cases now exercise the `GETEX` path. Both test-double `ICacheService` stubs were updated, so the build stays green. As a thin adapter over an out-of-process dependency, integration-testing `RedisService` rather than mocking it is the correct choice per the testing guidelines.

### Scope

Correctly bounded. `PlayerProgressRepository.GetCachedProgress` has the same shape but on a Redis **hash** key, which `GETEX` can't address — legitimately deferred. Filed as follow-up #2019.

### What's good

The investigation write-up on the Lua/`EVALSHA` half of the issue is exactly the right instinct — verifying the decompiled `StackExchange.Redis` behavior and declining to make a no-op change rather than churning the code. The comments at each call site explain the *why* (one round trip vs. awaited-get-then-fire-and-forget-expire) concisely.

---

PR #2026, opened 2026-07-15
Close socket integration-test gaps: takeover-close assertion, registry lifecycle unit test, tighter assertions

### LINKED ISSUE #1959: Socket integration-test gaps: takeover never asserts the old socket closes; registry lifecycle wiring and loose negative assertions
Test-coverage gaps around the socket lifecycle found by the health review. No production defect — the behaviors are correct as implemented — but regressions in them would currently ship green.

1. **Session takeover never asserts the superseded socket actually closes.** The takeover tests (`SocketManagerServiceTests.cs:40-65`, `:85-110`) assert the `SocketReplaced` push reaches the old client and the presence key re-points, but never that the old connection transitions to closed. The close is a separate step inside `SocketReplaced.ExecuteAsync` (`SocketReplaced.cs:20-21`, after the self-delivered send). Dropping or reordering that `Close` would leave two live sockets for one player — the exact condition the single-connection model exists to prevent — with every existing test still passing. Fix: in `RegisterSocket_SecondConnection_ReplacesPresenceKeyAndNotifiesOldSocket`, poll `socketA.State` to `Closed` and assert the close reason.

2. **`SocketConnectionRegistry` host-lifecycle wiring is untested.** `GameServerFactory.cs:71` strips all `IHostedService` registrations, and `SocketDrainTests` constructs the registry directly — so the `ApplicationStopping → EnsureDraining` hook, the `StopAsync` await (`SocketConnectionRegistry.cs:61-69`), and the run-once idempotency guard (`:150-156`) are never driven through a host. At minimum, a cheap unit test of `EnsureDraining` idempotence with a fake lifetime would pin the both-paths-reach-it contract.

3. **Tightenable negative assertions:** `SocketConnectionTests.cs:94` uses `Assert.ThrowsAnyAsync<Exception>` for the unauthenticated handshake (passes on *any* failure, not a 401-rejected upgrade; partially mitigated by the sibling HTTP-401 envelope test at `:99-109`), and `LoginControllerTests.cs:1208` uses `Assert.NotEqual(HttpStatusCode.OK, …)`, which passes on a 500 as readily as the intended 401. Both are tightenable for free.

ginderjeremiah:
## Summary

Closes three test-coverage gaps around the socket lifecycle flagged by the codebase health review, all test-only — no production behavior changes.

1. **Takeover never asserted the superseded socket actually closes.** `RegisterSocket_SecondConnection_ReplacesPresenceKeyAndNotifiesOldSocket` (`SocketManagerServiceTests.cs`) checked that the `SocketReplaced` push reached the old client and that the presence key re-pointed, but never that the old connection transitioned to `Closed`. The close is a separate step inside `SocketReplaced.ExecuteAsync` (after the self-delivered send) — dropping or reordering it would leave two live sockets for one player with this test still green. Added `TestSocketClient.WaitForCloseAsync()` (reads the close frame and completes the handshake so `State` settles at `Closed` instead of lingering in `CloseReceived`) and asserted the resulting state, close status, and close description in the takeover test.

2. **`SocketConnectionRegistry`'s host-lifecycle wiring was untested.** `GameServerFactory` strips all `IHostedService` registrations for the test host, and `SocketDrainTests` constructs the registry directly, so the `ApplicationStopping → EnsureDraining` hook, the `StopAsync` await, and the run-once idempotency guard (`_drainTask ??= DrainAsync()`) were never driven through a lifetime. Added a cheap unit test (`SocketConnectionRegistryTests`) with a fake `IHostApplicationLifetime` that pins both paths (an explicit `StopAsync` call and a subsequent `ApplicationStopping` signal) resolving to the exact same in-flight drain task.

3. **Tightened two loose negative assertions:**
   - `SocketConnectionTests.Connect_Unauthenticated_Fails` used `Assert.ThrowsAnyAsync<Exception>`, which passes on any failure. Verified the TestServer client actually throws `InvalidOperationException` with `"Incomplete handshake, status code: 401"` on a rejected upgrade, and tightened the assertion to that specific exception type + status code.
   - `LoginControllerTests.CharacterCreationData_WithoutAuthentication_IsRejected` used `Assert.NotEqual(HttpStatusCode.OK, ...)`, which passes on a 500 as readily as the intended 401. Tightened to `Assert.Equal(HttpStatusCode.Unauthorized, ...)`, matching every other unauthenticated-endpoint test in the same file.

## Test plan

- [x] `dotnet format --verify-no-changes --no-restore` — clean.
- [x] `dotnet test` (full solution) — all green: `Game.Core.Tests` 1366, `Game.Infrastructure.Tests` 66, `Game.Api.Tests` 805, `Game.Application.Tests` 983.

Closes #1959

---
_Generated by [Claude Code](https://claude.ai/code/session_019XCJXkQkz3nYcCGujWwHm5)_

### DIFF
--- a/Game.Api.Tests/Integration/LoginControllerTests.cs
+++ b/Game.Api.Tests/Integration/LoginControllerTests.cs
@@ -1210,7 +1210,7 @@ namespace Game.Api.Tests.Integration
         {
             // No bearer token — a pre-authentication request never reaches the creatable-class payload.
             var response = await Client.GetAsync("/api/Login/CharacterCreationData", CancellationToken);
-            Assert.NotEqual(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
         }
     }
 }
--- a/Game.Api.Tests/Integration/SocketConnectionTests.cs
+++ b/Game.Api.Tests/Integration/SocketConnectionTests.cs
@@ -91,8 +91,12 @@ namespace Game.Api.Tests.Integration
             // with no cached session is a different, authenticated case; see the rehydration test below.)
             var wsClient = Factory.Server.CreateWebSocketClient();
 
-            await Assert.ThrowsAnyAsync<Exception>(
+            // The TestServer client throws InvalidOperationException with the rejected status code baked
+            // into the message when the handshake doesn't complete with a 101 — tightened from ThrowsAny so
+            // this can't pass on, say, a 500 as readily as the intended 401-rejected upgrade.
+            var ex = await Assert.ThrowsAsync<InvalidOperationException>(
                 () => wsClient.ConnectAsync(new Uri("ws://localhost/socket"), CancellationToken));
+            Assert.Contains("401", ex.Message);
         }
 
         [Fact]
--- a/Game.Api.Tests/Integration/SocketManagerServiceTests.cs
+++ b/Game.Api.Tests/Integration/SocketManagerServiceTests.cs
@@ -9,6 +9,7 @@ using Game.TestInfrastructure.Helpers;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using StackExchange.Redis;
+using System.Net.WebSockets;
 using Xunit;
 
 namespace Game.Api.Tests.Integration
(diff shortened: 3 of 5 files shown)


ginderjeremiah (commented):
## Review: Close socket integration-test gaps (#1959)

**Verdict:** Approve — all three test-coverage gaps from #1959 are addressed accurately, each new assertion pins real behavior, and there are no production-code changes. (Submitted as a comment since I'm the PR author; GitHub won't accept a gating approval on one's own PR.)

I verified each change against the source it's meant to protect:

- **Takeover close assertion** (`SocketManagerServiceTests.cs`) — `SocketReplaced.ExecuteAsync` (`SocketReplaced.cs:20-21`) does `SendData(Success())` then `Close(ESocketCloseReason.SocketReplaced)`. The new block asserts `WebSocketState.Closed`, `NormalClosure`, and `ESocketCloseReason.SocketReplaced.GetDescription()` — so dropping or reordering that close, or changing the reason, now fails a test where before every assertion stayed green.
- **`WaitForCloseAsync` helper** (`TestSocketClient.cs`) — reads the close frame and completes the closing handshake so `State` settles at `Closed` instead of lingering in `CloseReceived`; returns the server's status/description for assertion. Correct for this scenario, where the close immediately follows the push.
- **Registry lifecycle unit test** (`SocketConnectionRegistryTests.cs`) — pins the `_drainTask ??= DrainAsync()` run-once guard (`SocketConnectionRegistry.cs:150-156`): `Assert.Same` proves the explicit `StopAsync` path and the post-`ApplicationStopping` path resolve to the same in-flight drain. Fills the gap left by `GameServerFactory` stripping `IHostedService` registrations and `SocketDrainTests` constructing the registry directly.
- **Tightened negatives** — `Assert.NotEqual(OK, …)` → `Assert.Equal(Unauthorized, …)` (`LoginControllerTests.cs`) and `ThrowsAnyAsync` → `ThrowsAsync<InvalidOperationException>` + `Contains("401", …)` (`SocketConnectionTests.cs`). Both now fail on a 500 where they previously passed; the latter matches the TestServer client's actual handshake-rejection behavior.

### Test coverage

Test-only PR — no production behavior changes, so the project's coverage rules are satisfied by construction. The additions strengthen existing coverage rather than accompanying new logic.

### What's good

Precise, minimal changes that each close a specific "regression ships green" hole rather than adding coverage for its own sake. The inline comments explaining *why* each assertion matters (e.g. the two-live-sockets failure mode) are genuinely useful context for a future maintainer.

No blocking items, should-fixes, or nits.

---

PR #2028, opened 2026-07-16
Resolve EntityStore.lastIdMap through persistEntity's identity map (#1962)

### LINKED ISSUE #1962: EntityStore.lastIdMap still resolves new ids positionally, unlike the child-saver path

ginderjeremiah:
## Summary

Closes #1962.

`EntityStore.save()` (`UI/src/routes/admin/workbench/entity-store.svelte.ts`) previously made its own, separate `resolveNewIds(fresh, diff.existingIds, diff.added)` call — with no `identityKey` — purely to populate `lastIdMap`, which the Workbench detail pane uses to keep the current selection pinned on a record across the save that just created it. This call used pure positional pairing, even though `persistEntity` (`save-helpers.ts`) had already resolved the same new-ids by identity content (via each entity's `toPrimaryDto`) for its own child savers, per the #1856 fix. Under a concurrent add from another admin landing in the same refetch, the two resolutions could disagree, and the UI could briefly follow the selection to the wrong record after a save (cosmetic — no data was ever corrupted, since the child-saver writes went through the correctly-resolved path independently).

## Change

- `EntityConfig.persist` (`entities/types.ts`) now returns `{ records: T[]; idMap: Map<number, number> }` instead of `T[]`.
- `persistEntity` (`save-helpers.ts`) returns its already-computed identity-resolved `idFor` map alongside the fresh records, instead of discarding it.
- `EntityStore.save()` consumes that returned `idMap` directly for `lastIdMap`, removing the duplicated, purely-positional `resolveNewIds` call entirely.
- All 10 entity configs (`entities/*.ts`) delegate `persist` straight to `persistEntity({...})`, so none needed changes — they pick up the new return shape automatically.

No production behavior changes beyond the bug fix itself (the child-saver id resolution was already correct); this only fixes the second, redundant resolution that fed `lastIdMap`.

## Test plan

- [x] Updated `save-helpers.test.ts` to assert `persistEntity`'s returned `idMap`.
- [x] Updated `entity-store.test.ts`'s save/lastIdMap tests to supply an identity-resolved `idMap` from the mock `persist`, mirroring what a real entity config now provides.
- [x] Updated `Workbench.test.ts` / `WorkbenchList.test.ts` mocks to the new `persist` return shape (including the existing "follows a newly-added record to its persisted id after save" behavioral test, which now exercises the fix directly).
- [x] `npm run test:unit:ci` — 3722/3722 tests passing.
- [ ] No backend changes; backend build/tests not affected.

https://claude.ai/code/session_01TA6GC4ESvqTMvEzdnnQim7

---
_Generated by [Claude Code](https://claude.ai/code/session_01TA6GC4ESvqTMvEzdnnQim7)_

### DIFF
--- a/UI/src/routes/admin/workbench/entities/types.ts
+++ b/UI/src/routes/admin/workbench/entities/types.ts
@@ -268,6 +268,10 @@ export interface EntityConfig<T extends Identified> {
 	sections: SectionConfig<T>[];
 	/** Fetch the saved record list (with caches refreshed). */
 	refresh: () => Promise<T[]>;
-	/** Persist a diff against existing/new endpoints and return the fresh saved list. */
-	persist: (diff: SaveDiff<T>) => Promise<T[]>;
+	/**
+	 * Persist a diff against existing/new endpoints. Returns the fresh saved list plus the
+	 * identity-resolved map of each added record's local (negative) id to its persisted id, so a
+	 * caller tracking a selection by id doesn't need to re-derive it (see {@link resolveNewIds}).
+	 */
+	persist: (diff: SaveDiff<T>) => Promise<{ records: T[]; idMap: Map<number, number> }>;
 }
--- a/UI/src/routes/admin/workbench/entity-store.svelte.ts
+++ b/UI/src/routes/admin/workbench/entity-store.svelte.ts
@@ -1,6 +1,6 @@
 import { SvelteMap, SvelteSet } from 'svelte/reactivity';
 import { toastError } from '$stores';
-import { canonicalEqual, PersistFailedError, resolveNewIds } from './save-helpers';
+import { canonicalEqual, PersistFailedError } from './save-helpers';
 import { entityWarnings } from './validation';
 import type { EntityConfig, Identified, SaveDiff } from './entities/types';
 
@@ -219,10 +219,10 @@ export class EntityStore<T extends Identified> {
 		this.saving = true;
 		try {
 			const diff = this.diff();
-			const fresh = await this.config.persist(diff);
-			this.lastIdMap = new SvelteMap(resolveNewIds(fresh, diff.existingIds, diff.added));
-			this.items = fresh.map(clone);
-			this.base = fresh.map(clone);
+			const { records, idMap } = await this.config.persist(diff);
+			this.lastIdMap = new SvelteMap(idMap);
+			this.items = records.map(clone);
+			this.base = records.map(clone);
 			this.deleted.clear();
 			this.flashSaved();
 		} catch (ex) {
--- a/UI/src/routes/admin/workbench/save-helpers.ts
+++ b/UI/src/routes/admin/workbench/save-helpers.ts
@@ -158,12 +158,17 @@ export class PersistFailedError extends Error {
  * Generic per-entity save: persists identity-level Add/Edit/Delete in one call,
  * refetches to resolve the ids of newly-added records (positionally — the backend
  * appends adds in send order), then runs each child-section persister for every
- * added/modified record against its real id. Returns the final saved list.
+ * added/modified record against its real id. Returns the final saved list alongside
+ * the identity-resolved new-id map, so a caller doesn't need to recompute it (and risk
+ * a purely positional recompute disagreeing with the identity-matched ids the child
+ * savers actually wrote against — see #1962).
  *
  * A failure once anything has committed is rethrown as {@link PersistFailedError}; a pre-commit
  * failure propagates raw (see that type's docs).
  */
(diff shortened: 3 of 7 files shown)


ginderjeremiah (commented):
## Review: Resolve `EntityStore.lastIdMap` through `persistEntity`'s identity map (#1962)

**Verdict:** Approve — clean, correctly-scoped fix that implements exactly what the issue asked; no blocking or should-fix findings. (Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating self-review.)

The change threads `persistEntity`'s already-identity-resolved `idFor` map out to the caller and has `EntityStore.save()` consume it directly, eliminating the second, purely-positional `resolveNewIds` call that fed `lastIdMap`. Under a concurrent add from another admin, the two resolutions could disagree and the detail pane could briefly follow the selection to the wrong record after a save — the fix closes that gap and removes a duplicated computation (DRY), matching the issue's suggested approach.

### Correctness

- `persistEntity` now returns `{ records, idMap: idFor }` on **both** return paths (child-saver and non-child-saver branches). In the child-saver branch the map is computed against the first refetch and returned after the second — this is sound: the added records' persisted ids are committed in the primary batch *before* the child savers run, so they're stable across the second refresh, and `lastIdMap` only needs those added-record mappings.
- Verified the PR's claim that all 10 entity configs delegate `persist` straight to `persistEntity(...)` with no wrapping, so the return-shape change flows through their inferred types automatically — none needed edits.
- The `resolveNewIds` export stays live (still used by `progression/progression-store.svelte.ts`, a separate store path already fixed identically under #1856), so dropping the now-unused import from `entity-store.svelte.ts` leaves no dead code.

### Test coverage

Adequate and follows the project's rules — unit tests mirror the source paths:
- `save-helpers.test.ts` asserts the returned `idMap` (`-1 → 2`, `-2 → 3`) alongside `records`.
- `entity-store.test.ts` mocks now derive the `idMap` from `diff.added` rather than hardcoding it, so the `lastIdMap` mapping and wholesale-replacement tests still exercise the real contract.
- `Workbench.test.ts` / `WorkbenchList.test.ts` mocks updated to the new shape, with the existing "follows a newly-added record to its persisted id" behavioral test now exercising the fixed path directly.

### What's good

- Return-shape change is applied consistently across both `persistEntity` branches, and the interface doc comment on `EntityConfig.persist` clearly explains *why* the map is surfaced (avoids a positional recompute disagreeing with the identity-matched writes).
- Tight scope — no backend touched, no unrelated churn, and the mocks were updated to mirror real behavior instead of hardcoding expected ids.


---

PR #2029, opened 2026-07-16
Fold Item totalAttributes recompute into an instance method

### LINKED ISSUE #1967: Fold item totalAttributes recompute into an Item method to remove duplication
## Context

Follow-up from PR #1964 (which converted `Item`/`ItemMod` to classes).

The logic that rebuilds an item's cached `totalAttributes` from its base attributes plus its applied mods now lives in two places with the identical expression:

- `UI/src/lib/battle/item.ts:60-61` — the new `Item` constructor:
  ```ts
  this.totalAttributes = new BattleAttributes(
      [...this.attributes, ...this.appliedMods.flatMap((mod) => mod.attributes)], ...);
  ```
- `UI/src/lib/engine/player/inventory-manager.ts:443-444` — `refreshItemAttributes`:
  ```ts
  const allAttributes = [...item.attributes, ...item.appliedMods.flatMap((mod) => mod.attributes)];
  item.totalAttributes = new BattleAttributes(allAttributes, false);
  ```

These two copies can silently drift (e.g. if the aggregation rule or the `combine` flag changes in one but not the other). The duplication predates #1964 (it previously sat between `newItem` and `refreshItemAttributes`), but now that `Item` is a class the natural home is an instance method.

## Proposal

Add an `Item.recomputeTotalAttributes()` (or similar) instance method that owns the aggregation + `new BattleAttributes(..., false)` construction, have the constructor call it, and have `InventoryManager.refreshItemAttributes` call it instead of rebuilding inline. No behavior change intended.

Pure tech-debt: no user-facing effect.

ginderjeremiah:
## Summary

Follow-up from #1964, which converted `Item`/`ItemMod` to classes. The logic that rebuilds an item's cached `totalAttributes` from its base attributes plus its applied mods lived in two places with the identical expression:

- `Item`'s constructor (`UI/src/lib/battle/item.ts`)
- `InventoryManager.refreshItemAttributes` (`UI/src/lib/engine/player/inventory-manager.ts`)

This adds `Item.recomputeTotalAttributes()` as the single owner of that aggregation formula:

- The `Item` constructor now calls it instead of inlining the expression.
- `InventoryManager.refreshItemAttributes` now calls it instead of rebuilding `BattleAttributes` inline.
- While making the change, I noticed `challenges-view.svelte.ts`'s `buildPreviewItem` was hand-building an object literal that also duplicated the same formula (a third copy, invisible to the issue's original two-copy count since `Item` wasn't yet a required-shape class member). Adding `recomputeTotalAttributes` to the `Item` class made that literal fail type-checking (missing method), so I switched it to construct a real `Item` instance via `new Item(itemData, { itemId: itemData.id, equipped: false, favorite: false, appliedMods: [] })` instead — removing that third copy too rather than patching around it.

No behavior change intended.

## Test plan

- [x] `npm run test:unit:ci` — 299 test files / 3722 tests pass, including `item.test.ts`, `inventory-manager.test.ts`, and `challenges-view.svelte.test.ts` which already cover the constructor/refresh/preview-item behavior this refactors

Closes #1967

---
_Generated by [Claude Code](https://claude.ai/code/session_017btx6S9iRPtBVWu5cCTaiG)_

### DIFF
--- a/UI/src/lib/battle/item.ts
+++ b/UI/src/lib/battle/item.ts
@@ -57,10 +57,12 @@ export class Item implements IItem {
 		this.favorite = invItem.favorite ?? false;
 		// Drop any applied mod whose own definition is missing/retired so one stale mod can't crash the item.
 		this.appliedMods = invItem.appliedMods.map((am) => newItemMod(am)).filter((mod): mod is ItemMod => mod != null);
-		this.totalAttributes = new BattleAttributes(
-			[...this.attributes, ...this.appliedMods.flatMap((mod) => mod.attributes)],
-			false
-		);
+		this.totalAttributes = this.recomputeTotalAttributes();
+	}
+
+	/** Rebuilds the cached totalAttributes from the item's base attributes plus its applied mods. */
+	recomputeTotalAttributes(): BattleAttributes {
+		return new BattleAttributes([...this.attributes, ...this.appliedMods.flatMap((mod) => mod.attributes)], false);
 	}
 }
 
--- a/UI/src/lib/engine/player/inventory-manager.ts
+++ b/UI/src/lib/engine/player/inventory-manager.ts
@@ -9,7 +9,7 @@ import {
 } from '$lib/api';
 import { PUNCH_SKILL_ID } from '$lib/api/types/game-constants';
 import { playerManager } from '$lib/engine';
-import { BattleAttributes, Item, newItem, newItemMod } from '$lib/battle';
+import { Item, newItem, newItemMod } from '$lib/battle';
 import { logMessage } from '$lib/engine/log';
 import { SerializedQueue } from '$lib/common';
 import { staticData } from '$stores';
@@ -440,8 +440,7 @@ export class InventoryManager {
 
 	/** Rebuilds an item's cached totalAttributes after its applied mods change. */
 	private refreshItemAttributes(item: Item) {
-		const allAttributes = [...item.attributes, ...item.appliedMods.flatMap((mod) => mod.attributes)];
-		item.totalAttributes = new BattleAttributes(allAttributes, false);
+		item.totalAttributes = item.recomputeTotalAttributes();
 	}
 
 	/** Refreshes the memoised equipmentStats only when the changed item is equipped — an unequipped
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: Fold `Item.totalAttributes` recompute into an instance method (#1967)

**Verdict:** Approve — does exactly what #1967 asked, behavior-preserving, and removes an extra duplicate the issue didn't know about.

_(Submitted as a `COMMENT` because the reviewer identity here matches the PR author; the intent is a clean approve.)_

The change gives the `totalAttributes` aggregation formula a single owner (`Item.recomputeTotalAttributes()`) and wires the two sites named in the issue to it:
- Constructor (`item.ts:60`) → calls the method instead of inlining.
- `InventoryManager.refreshItemAttributes` (`inventory-manager.ts:442`) → calls the method, dropping the now-unused `BattleAttributes` import.

I verified the behavior is unchanged:
- `recomputeTotalAttributes()` builds `new BattleAttributes([...attributes, ...appliedMods.flatMap(m => m.attributes)], false)` — identical expression and `combine=false` flag as both former copies, so no drift risk (the exact silent-drift the issue flagged).
- The third copy removed in `challenges-view.svelte.ts` `buildPreviewItem` is genuinely equivalent: the old literal set `totalAttributes: new BattleAttributes(itemData.attributes, false)`, and `new Item(itemData, { …, appliedMods: [] })` computes `[...attributes, ...[]]` with the same `false` flag — same result. Constructing a real `Item` here is also more correct than the hand-built literal (it was structurally typed as `Item` without actually being one). Good call folding it in rather than patching around the new required method.

### Test coverage
No new tests, which is appropriate: this relocates an existing expression with no behavior change, and per the PR the existing `item.test.ts`, `inventory-manager.test.ts`, and `challenges-view.svelte.test.ts` already exercise the constructor / refresh / preview-item paths.

### What's good
- Tight, single-purpose tech-debt change that fully closes the issue.
- Removing the `BattleAttributes` imports that became unused in both edited files — no dead imports left behind.
- Catching and eliminating the third duplicate the issue's two-copy framing missed, with a clear explanation in the PR body.

Nothing to change.

---

Issue #1976, opened 2026-07-14
Generic Workbench retire-confirm still misses this session's unsaved cross-catalogue edits


ginderjeremiah:
## Context

Follow-up to #1863. That issue's gap #2 ("the reference computation ignores the session's own unsaved edits") is fixed for the progression editor — `PathDetail`/`TierDetail` now pass `{ proficiencies: store.profs }` to `referenceSourcesFromStatic`, which is correct and safe because `pathReferences`/`proficiencyReferences` (`references.ts`) look up proficiencies by `.filter(...)`, never by array index.

An equivalent fix was attempted for the generic Workbench (`WorkbenchDetail.svelte`, all non-progression entities) in the same PR (#1975) but was reverted after review — see that PR's review thread for the full analysis. Summary of why it doesn't generalize the same way:

- **Wrong catalogue.** `computeReferences`'s per-entity reference functions look up a *sibling* catalogue, never the one being retired: an enemy is referenced by `zones`/`challenges`, an item by `challenges`/`classes`, a skill by `items`/`classes`/`skillRecipes`/`proficiencies`. `WorkbenchDetail` only holds the one `EntityStore` for the entity currently open, so it has no access to those siblings' live (unsaved) records — overriding its own catalogue is a no-op for `zones`/`challenges`/`items`/`classes`/`skillRecipes` (their reference fns never read that slot at all), and `classes`/`skillRecipes` aren't even valid `computeReferences` entity keys.
- **The two slots that *are* self-referential are id-indexed, and `EntityStore.items` isn't dense-by-id.** `enemyReferences` reads the enemy's own spawns via `enemies[id]?.spawns`, and `skillReferences` names a referencing recipe via `nameByIndex(skills, recipe.resultSkillId)` — both assume the zero-based Id-as-index invariant (`docs/backend.md` → _Reference Data_). `EntityStore.addItem` prepends new records with negative ids (`[newItem(id), ...items]`), so the moment an admin adds a record this session, `store.items` is no longer index-aligned. Substituting it directly corrupts those two lookups (a completely normal add-then-retire-something-else session would show the wrong enemy's spawn table in the confirm dialog).

## What a real fix needs

[...]

PR #2030, opened 2026-07-16
Reflect this session's own live edits in generic Workbench retire-confirm

ginderjeremiah:
## Summary

Partial fix for #1976. The generic Workbench retire-confirm's "referenced by" surface (`references.ts` / `retire-confirm.ts`) is computed from `staticData` (last-saved state), so it can miss references introduced by this session's own unsaved edits. This PR closes the **self-referential** half of that gap:

- `enemyReferences` reads an enemy's own `spawns` field via `enemies[id]?.spawns`.
- `skillReferences` names a referencing recipe via `nameByIndex(skills, recipe.resultSkillId)` — both self-index into the *same* catalogue as the entity currently open in the Workbench pane.

An earlier attempt at this (#1975) was reverted because naively substituting `store.items` for the static array corrupts these lookups: `EntityStore.addItem` prepends new records with negative ids, shifting every saved record's array position out of the zero-based-id alignment `references.ts` assumes.

This PR adds `denseByLiveId` (`retire-confirm.ts`), which rebuilds a proper id-indexed array from the live store (dropping never-saved negative-id records, which can never be a retire target or a referencing recipe's result anyway), and `ownCatalogueOverride`, which applies it only for the two entity types where it matters (a no-op everywhere else). `WorkbenchDetail.svelte`'s `onRetire` now threads this through.

## What's intentionally left open

The issue's titular **cross-catalogue** gap (e.g. retiring an item while an unsaved edit to a *different* entity type, like classes, sits in a tab the admin isn't currently on) is not addressed here. Closing that generically would require holding every reference-carrying entity's `EntityStore` alive across a tool switch — which reverses the currently documented "only one editing surface mounted at a time, discard on switch" model (see `dirty.svelte.ts`'s doc comment and `discard-guard.ts`). That's a real architecture/UX decision (memory cost of N persistent stores, whether "discard on switch" still makes sense, etc.), not a scoped fix, so I left it for the maintainer to weigh in on rather than deciding unilaterally. I've left a comment on #1976 with this analysis and a couple of options.

## Test plan

- [x] Added unit tests in `retire-confirm.test.ts` covering `ownCatalogueOverride`: no-op for non-self-referencing entities, dense-by-id reconstruction ignoring negative ids, and an end-to-end case proving an enemy's own unsaved spawn edit now surfaces correctly in the confirm dialog.
- [x] `npm run test:unit:ci` — 299 files / 3725 tests passing.

---
_Generated by [Claude Code](https://claude.ai/code/session_016rTjyCM3zHFwXKmuAQQuwg)_

### DIFF
--- a/UI/src/routes/admin/workbench/retire-confirm.ts
+++ b/UI/src/routes/admin/workbench/retire-confirm.ts
@@ -61,6 +61,9 @@ function denseByLiveId<T extends { id: number }>(live: T[]): T[] {
  * array index rather than `proficiencies`' always-`.filter()`-based lookups.
  */
 export function ownCatalogueOverride(entityKey: string, items: { id: number }[]): Partial<ReferenceSources> {
+	// The computed key can't narrow past `string`, and `items` is the generic `{ id }[]` shape rather
+	// than `IEnemy[]`/`ISkill[]` — sound here because the guard restricts entityKey to `enemies`/
+	// `skills`, and `items` (an EntityStore's `.items`) are always that entity's live full records.
 	return SELF_REFERENCING_KEYS.has(entityKey)
 		? ({ [entityKey]: denseByLiveId(items) } as unknown as Partial<ReferenceSources>)
 		: {};
--- a/UI/src/routes/admin/workbench/retire-confirm.ts
+++ b/UI/src/routes/admin/workbench/retire-confirm.ts
@@ -24,6 +24,48 @@ export function referenceSourcesFromStatic(overrides: Partial<ReferenceSources> 
 	};
 }
 
+/**
+ * Reference-source slots that a generic Workbench entity's own reference function reads (`enemies`
+ * reads its own `spawns`; `skills` names a referencing recipe via its own catalogue) — see
+ * `references.ts`'s `enemyReferences`/`skillReferences`. Every other generic-Workbench entity's
+ * reference function only reads *sibling* catalogues it has no live copy of, so overriding those
+ * slots would be a no-op (#1976's remaining, genuinely cross-catalogue gap — closing it generically
+ * would mean holding every reference-carrying entity's store alive across a tool switch, reversing
+ * the documented one-surface-mounted-at-a-time model in `dirty.svelte.ts`).
+ */
+const SELF_REFERENCING_KEYS = new Set(['enemies', 'skills']);
+
+/**
+ * Rebuilds a zero-based-id-indexed array from a store's live `items`, honouring the Id-as-index
+ * invariant `references.ts`'s self-referential lookups rely on (`docs/backend.md` → _Reference
+ * Data_). `EntityStore.addItem` prepends new records with negative ids, shifting every saved
+ * record's array position, so indexing the live array directly would resolve the wrong record. A
+ * never-saved (negative-id) record is dropped: only an already-saved record can be a retire target
+ * or a referencing recipe's result, so none is looked up by id here.
+ */
+function denseByLiveId<T extends { id: number }>(live: T[]): T[] {
+	const dense: T[] = [];
+	for (const record of live) {
+		if (record.id >= 0) {
+			dense[record.id] = record;
+		}
+	}
+	return dense;
+}
+
+/**
+ * Live override for the generic Workbench entity currently open, closing the self-referential half
+ * of #1976 ("this session's unsaved edits" of the entity's *own* catalogue) for the two entity types
+ * whose reference lookup reads it. A no-op for every other entity. Mirrors the progression editor's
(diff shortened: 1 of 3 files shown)


UI/src/routes/admin/workbench/retire-confirm.ts
```diff
@@ -24,6 +24,48 @@ export function referenceSourcesFromStatic(overrides: Partial<ReferenceSources>
...
+/**
+ * Live override for the generic Workbench entity currently open, closing the self-referential half
+ * of #1976 ("this session's unsaved edits" of the entity's *own* catalogue) for the two entity types
+ * whose reference lookup reads it. A no-op for every other entity. Mirrors the progression editor's
+ * `{ proficiencies: store.profs }` override, but density-safe since enemies/skills are looked up by
+ * array index rather than `proficiencies`' always-`.filter()`-based lookups.
+ */
+export function ownCatalogueOverride(entityKey: string, items: { id: number }[]): Partial<ReferenceSources> {
+	return SELF_REFERENCING_KEYS.has(entityKey)
+		? ({ [entityKey]: denseByLiveId(items) } as unknown as Partial<ReferenceSources>)
```
ginderjeremiah, line 68: Nit: the `as unknown as Partial<ReferenceSources>` double-cast is the pragmatic escape hatch here (a computed `string` key can't narrow to the concrete slot type, and `denseByLiveId` returns the generic `{ id: number }[]` rather than `IEnemy[]`/`ISkill[]`). It's safe because the `SELF_REFERENCING_KEYS` guard restricts `entityKey` to `enemies`/`skills` and `items` are the live full records — but that reasoning isn't obvious from the cast alone. A one-line comment on *why* the cast is sound would save a future reader from second-guessing it. Optional.

---

ginderjeremiah (commented):
## Review: scoped fix for the self-referential half of #1976 (Workbench retire-confirm reflecting this session's own unsaved edits)

**Verdict: Approve** (posted as a comment since I'm the PR author) — the change is correct, matches the issue's scoped intent, and is honest about what it deliberately leaves open. One should-fix on test symmetry, one nit; neither is a blocker.

### What's good
- The core reasoning is sound and well-documented. `denseByLiveId` correctly rebuilds a zero-based-id-indexed array from `store.items`, preserving the Id-as-index invariant that `enemies[id]?.spawns` and `nameByIndex(skills, …)` rely on, rather than trusting the live array position that `EntityStore.addItem`'s negative-id prepend breaks. This directly addresses the exact failure mode that got #1975 reverted.
- Dropping never-saved (negative-id) records is the right call and correctly justified: such a record can neither be a saved retire target nor a referencing recipe's result, so it's never looked up by id. A retire of a brand-new record simply resolves to no references, which is acceptable.
- Scoping the override to `SELF_REFERENCING_KEYS` (`enemies`, `skills`) makes it a true no-op for every entity whose reference fn only reads sibling catalogues — matching `references.ts`. The updated doc comment in `WorkbenchDetail.svelte` accurately explains the split.
- Correctly resists over-reaching: the genuinely cross-catalogue half is left to the maintainer with a documented rationale (reversing the one-surface-mounted model) and a comment on #1976, rather than being force-fit here. That's the right boundary.

### Should-fix
- **Test coverage — `skills` self-referencing path is untested.** `SELF_REFERENCING_KEYS` has two members and the PR body names both, but the added tests only exercise `enemies` end-to-end (`retire-confirm.test.ts`). The `skills` branch runs the same `ownCatalogueOverride`/`denseByLiveId` code, so the override *mechanism* is covered — but the `skills` override feeds a genuinely different downstream lookup: `skillReferences` uses the overridden catalogue only to *name* referencing recipes via `nameByIndex(skills, recipe.resultSkillId)`, not to compute a spawn surface. Given this project's testing bar, add a mirrored e2e proving a live (unsaved) skill rename surfaces in a referencing recipe's displayed name, so both stated cases are actually pinned.

### Nits
- `ownCatalogueOverride` — the `as unknown as Partial<ReferenceSources>` double-cast is a necessary escape hatch for the computed string key, but a one-line comment on why it's sound (guard restricts the key; `items` are the live full records) would help a future reader. Left inline.
- `SELF_REFERENCING_KEYS` carries an implicit coupling to `references.ts`: a future self-referential lookup added there must also be added here or it silently won't reflect live edits. Already well-documented by the comment pointing at `references.ts`, so this is just a note, not a change request.

### CI
Backend jobs correctly skipped (frontend-only diff). `test-frontend`, `verify-frontend`, and Chromium/Firefox e2e are green; **webkit e2e was still in progress at review time**, so CI isn't fully green yet.

Not merging: CI is incomplete and there's an open should-fix (the `skills` test) that belongs in this PR rather than as a follow-up. Add that test (the cast comment is optional) and this is good to go.

---

PR #2031, opened 2026-07-16
Filter pathReferences to gateways whose own path stays live

### LINKED ISSUE #1983: Workbench: pathReferences over-reports gateways sitting on an already-retired path
Follow-up nit from the review of #1975 (#1863).

`pathReferences` in `UI/src/routes/admin/workbench/references.ts` mirrors the backend soft-lock guard `AdminPaths.FindRetiredPathGatingLiveGateway`. The backend only flags a gateway whose **own path stays live** (`PathStaysLive(gateway.PathId)`). The frontend approximates "stays live" as `p.pathId !== id` because `pathReferences` has no access to per-path retirement state.

**Consequence:** if a gateway prerequisite sits on a *different, already-retired* path, the frontend reports it as a `strong` reference (blocking-style "re-point first" wording) even though the backend would skip it — that content is already dead. This is advisory **over**-reporting only: it never misses a warning or bypasses the confirm, and the retire still proceeds on "Retire anyway" with the backend authoritative. So it's cosmetic, not a correctness/safety gap.

**Fix direction:** give `pathReferences` access to the live path records (as `computeReferences`/`referenceSourcesFromStatic` already thread other catalogues) and filter out gateways whose own path is retired, matching `PathStaysLive` exactly.

Low priority — the fix in #1975 is correct and safe as-is; this just tightens the advisory accuracy.

Tags: tech debt, claude, scope: small.

ginderjeremiah:
## Summary

Fixes #1983 — a nit from the review of #1975 (#1863).

`pathReferences` in `UI/src/routes/admin/workbench/references.ts` mirrors the backend soft-lock guard `AdminPaths.FindRetiredPathGatingLiveGateway`, but it approximated "the gateway's own path stays live" as `p.pathId !== id`, since it had no access to per-path retirement state. That over-reports: a gateway prerequisite sitting on a *different, already-retired* path was flagged as a `strong` ("re-point first") reference even though the backend's `PathStaysLive(gateway.PathId)` check would skip it — that content is already dead. This was advisory-only over-reporting (never missed a warning, never bypassed the confirm), so it was cosmetic rather than a correctness/safety gap.

## Changes

- Added `paths: IPath[]` to `ReferenceSources` (`references.ts`), threaded through `referenceSourcesFromStatic` (`retire-confirm.ts`) from `staticData.paths`, matching how every other catalogue is sourced.
- `pathReferences` now excludes a gateway whose own path (looked up by zero-based-id index, honouring the Id-as-index invariant) is retired, matching `PathStaysLive` exactly.
- Added unit tests covering: a gateway on an already-retired path is excluded, and a gateway on a confirmed-live path is still flagged.

## Test plan

- [x] Added/updated unit tests in `references.test.ts` and `retire-confirm.test.ts`.
- [x] `npm run test:unit:ci` — 299 files / 3724 tests passing.

Closes #1983

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pb1bHz75zUWHzxHHb7m911)_

### DIFF
--- a/UI/src/routes/admin/workbench/references.ts
+++ b/UI/src/routes/admin/workbench/references.ts
@@ -5,6 +5,7 @@ import {
 	type IEnemy,
 	type IItem,
 	type IItemMod,
+	type IPath,
 	type IProficiency,
 	type ISkill,
 	type ISkillRecipe,
@@ -31,6 +32,7 @@ export interface ReferenceSources {
 	skillRecipes: ISkillRecipe[];
 	proficiencies: IProficiency[];
 	skills: ISkill[];
+	paths: IPath[];
 }
 
 /** A category of inbound reference and the names of the records that make it. */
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Filter `pathReferences` to gateways whose own path stays live (#1983)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change makes the Workbench retire-confirm advisory match the backend soft-lock guard exactly, is cleanly scoped, and is well-tested.

I verified the frontend logic against the backend it mirrors (`AdminPaths.FindRetiredPathGatingLiveGateway` / `PathStaysLive`): the backend skips a gateway when `!PathStaysLive(gateway.PathId)`, and the new `!paths[p.pathId]?.retiredAt` filter reproduces that "gateway's own path must stay live" condition. The `paths[p.pathId]` lookup honours the documented zero-based Id-as-index invariant, consistent with `nameByIndex` and the rest of this file, and `paths` is threaded through `referenceSourcesFromStatic` from `staticData.paths` exactly like every other catalogue. The updated doc comment accurately describes the new behavior.

### Test coverage
- Both directions are covered: a gateway on an already-retired path is excluded, and a gateway on a confirmed-live path is still flagged. Test helper (`path()` / `bySlot`) matches `IPath` and the id-indexed lookup. Good.

- The progression editor's path retire (`PathDetail.svelte:97`) passes live `proficiencies` (`store.profs`) but static `paths`, so an *unsaved, in-session* retirement of a gateway's own path wouldn't yet be reflected here. That's the deliberate cross-catalogue / live-edit gap tracked in #1976 (and partially addressed in #2030), and sourcing `paths` from `staticData` avoids the negative-id index-misalignment hazard `denseByLiveId` was written to solve — so keeping it last-saved is the correct choice for a small-scope fix.
- Edge case: if a gateway's `pathId` had no entry in `paths` (undefined), the frontend includes it while the backend's `LookupPath(...) is { RetiredAt: null }` would exclude it — but every proficiency belongs to a real path, so this is unreachable, and the divergence direction is safe over-reporting.

### What's good
Tight, well-reasoned fix that closes exactly the advisory-accuracy gap the issue described, with the comment and tests updated to match. No changes requested.

---

PR #2032, opened 2026-07-16
Self-heal progress hash rows that fail to deserialize (#2000)

### LINKED ISSUE #2000: Progress hash rows deserialize shape drift to defaults silently — zeroed aggregates would be written back to Postgres as absolute upserts
## Problem

Sibling of #1924 on a third cache surface with a **worse** failure mode (#1924 covers the player/session string blobs, which throw loudly; this one fails silently).

Unlike `PlayerCacheModel` (all `required` members, so shape drift hard-fails), the progress hash row models `CachedPlayerStatistic`/`CachedPlayerChallenge`/`CachedPlayerProficiency` (`Game.DataAccess/PlayerProgressCacheModels.cs:17-37`) have no `required` members and no post-deserialize validation, and `FromHashFields` (`Game.DataAccess/Repositories/PlayerProgressRepository.cs:366-398`) silently skips a row that deserializes to null.

**Failure scenario:** a field rename/retype ships. On the next read of a warm `Progress_{playerId}` hash, old rows deserialize *successfully* with default values (`Value = 0`, `Level = 0`) — no exception, no self-heal, and because the hash is the source of truth it never falls through to the DB while present. Battle completion then folds deltas onto the zeroed aggregate, and `Save` publishes **absolute-value** upserts (`ProgressUpdatedEvent`, applied verbatim by `DataProviderSynchronizer`), permanently regressing the player's real statistics/challenge/proficiency rows in Postgres — silent data loss rather than #1924's lockout. A null-deserializing row is likewise dropped, shadowing its DB row behind a present key.

## Fix direction

Per-row parse validation that treats an unparseable/implausible row as cache corruption: delete the key, log loudly, and fall through to the DB reload — the same self-heal `RedisService.HashGetAllIfExists` already applies to a wrong-representation key. Making the cache models `required`-member (so drift throws) plus catching the throw into the delete-and-reload path is the cheapest robust shape. Worth coordinating with #1924's fix so all three cache surfaces share the convention.

ginderjeremiah:
## Summary

`CachedPlayerStatistic`/`CachedPlayerChallenge`/`CachedPlayerProficiency` (the rows in the `Progress_{playerId}` Redis hash) had no `required` members and no post-deserialize validation, unlike `PlayerCacheModel`. A field rename/retype shipped after a hash was already cached would deserialize old rows *successfully* with defaulted (zeroed) values instead of failing loudly — no exception, no self-heal. Battle completion would then fold deltas onto the zeroed aggregate, and `Save` would upsert it to Postgres as an **absolute value**, permanently regressing the player's real statistics/challenge/proficiency rows. A row that deserialized to a null object was likewise silently dropped, shadowing its DB row behind a present key.

This addresses #2000, a sibling of #1924 (cached player/session blobs) on a third cache surface with a worse failure mode — #1924's gap threw loudly (a lockout); this one failed silently (data loss).

## Changes

- `Game.DataAccess/PlayerProgressCacheModels.cs`: marked every member of the three cached row models `required`, so a shape mismatch throws `JsonException` on deserialize instead of silently defaulting.
- `Game.DataAccess/Repositories/PlayerProgressRepository.cs`:
  - `FromHashFields` now throws (rather than silently skipping) when a row deserializes to `null`, so that failure mode routes through the same self-heal as a shape mismatch.
  - `GetCachedProgress` catches `JsonException` from a hash read, logs it, deletes the corrupted key, and falls through to the existing DB-reload path — mirroring the convention `PlayerRepository.GetPlayer`/`SessionStore.GetSession` already apply (#1924) and the same reasoning `RedisService.HashGetAllIfExists` already applies to a wrong-representation key.
  - Added an `ILogger<PlayerProgressRepository>` dependency (DI-resolved automatically; updated the one test that constructs the repository directly).

## Test plan

- [x] `dotnet build` — succeeds, 0 warnings/errors.
- [x] `dotnet test` (full solution) — 3226 tests passed, 0 failed.
- [x] Added `GetStatistics_OnUnparsableHashField_DeletesTheKeyAndReloadsFromDb` — malformed JSON in one hash field self-heals via delete + DB reload.
- [x] Added `GetStatistics_OnShapeDriftedHashField_SelfHealsInsteadOfZeroingTheAggregate` — a row missing all required members (simulating a pre-shape-change blob) self-heals instead of silently zeroing the aggregate.

Closes #2000

---
_Generated by [Claude Code](https://claude.ai/code/session_01GLcRxY3SvEvxM1YLF1FFr4)_

### DIFF
--- a/Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/PlayerProgressRepositoryIntegrationTests.cs
@@ -15,6 +15,7 @@ using Game.TestInfrastructure.Fixtures;
 using Game.TestInfrastructure.Helpers;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using StackExchange.Redis;
 using Xunit;
 
@@ -195,7 +196,9 @@ namespace Game.Application.Tests.DataAccess
             // Save must wrap it the same way SavePlayer does, so the socket layer can force the connection's
             // in-memory Player to reload afterward rather than letting the caller's already-applied mutations
             // ride along un-persisted (#1819).
-            var throwingRepo = new PlayerProgressRepository(context, challenges, cache, new ThrowingPubSubService(), batch);
+            var throwingRepo = new PlayerProgressRepository(
+                context, challenges, cache, new ThrowingPubSubService(), batch,
+                scope.ServiceProvider.GetRequiredService<ILogger<PlayerProgressRepository>>());
 
             var progress = await throwingRepo.Load(MakeDomainPlayer(playerId));
             progress.RecordBattleCompleted(MakeEnemy(), victory: true, playerDied: false, totalMs: 1000,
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Self-heal progress hash rows that fail to deserialize (#2000)

**Verdict:** Approve (submitted as a comment since I'm the PR author) — the fix closes the silent-data-loss gap exactly as #2000 describes and mirrors the #1924 convention faithfully.

This is a clean, well-scoped fix. The three cached row models (`CachedPlayerStatistic`/`Challenge`/`Proficiency`) are now all-`required`, so shape drift on a warm `Progress_{playerId}` hash throws `JsonException` on deserialize instead of silently defaulting to zeroed rows — which `Save` would otherwise fold deltas onto and upsert to Postgres as **absolute** values, permanently regressing real progress. `GetCachedProgress` catches that throw, logs, deletes the key, and falls through to the DB reload, and `FromHashFields` now throws (rather than skipping) on a null-deserializing row so it routes through the same self-heal instead of shadowing the DB row behind a present-but-incomplete key.

I verified the claimed convention: the catch-`JsonException` → log → `_cache.Delete` → fall through to DB reload is line-for-line consistent with `PlayerRepository.GetPlayer` and `SessionStore.GetSession` (#1924), and the DB reload remains the one path that recreates the key — the `Save` advance still uses `HashSetIfExistsAndForget` and never resurrects it, so the #1761 partial-view guard is preserved. Moving `ExpireAndForget` inside the success branch is correct: a hit that fails to parse is deleted rather than TTL-refreshed.

### Test coverage

- Two integration tests cover both failure modes on the statistic path: unparsable JSON (`{not valid json`) and shape-drift (`{}`, missing every required member) — both assert delete + DB reload restores the real value instead of throwing or zeroing. Integration level is the right call here since the behavior is fundamentally a Redis + Postgres interaction; there's no isolated domain logic to unit-test.
- The challenge (`C_`) and proficiency (`P_`) paths aren't independently exercised, but they run through the identical shared `try/catch` in `GetCachedProgress` and the same `?? throw` in `FromHashFields`, so the marginal value of duplicating the scenario per row kind is low. Adequate as-is.

### What's good

- Correctly self-heals rather than throwing, avoiding both the silent data loss (#2000) and a #1924-style TTL lockout.
- Faithful reuse of the existing cache-corruption convention across all three cache surfaces, including the log message shape and DI'd `ILogger`.
- Comments are appropriately detailed for genuinely subtle invariants (partial-view guard, why required-members are the mechanism) without over-explaining.

No blocking or should-fix findings.

---

Issue #1865, opened 2026-07-11
Docs devolution follow-up: remaining moves from the July 2026 accuracy audit


ginderjeremiah:
Follow-up to the docs pass in #1848, which fixed the stale claims, split the socket lifecycle into `backend-sockets.md`, and condensed the two clearest duplications in game-design.md. The audit flagged several more devolution candidates that were deliberately left for a separate, reviewable pass (each needs careful merging into a sub-doc's existing structure):

**backend.md**
- The challenge-completion notification section's fault-classification taxonomy (timeout vs teardown vs malformed, delivery-failure logging) overlaps backend-persistence.md's "Queue processing resilience" — move the classification detail there, keep the design decision (escalate-don't-retry, `ServerCommandFailed` re-sync) inline.
- The retirement mega-bullet (~350 words on per-entity retirement effects: spawn tables, proficiency reverse-index exclusion, gateway index, lazy relocation) — move the mechanics to backend-admin.md's content-lifecycle section (already home to the retirement guards), keep the invariants inline.
- The player-progress bullet's write-behind/dirty-set/idempotent-upsert plumbing — belongs in backend-persistence.md.

**game-design.md**
- The crit/dodge/parry section heavily duplicates game-design-combat.md (which exists precisely for the derivation/template detail). Trim the inline copies of the enabler×multiplier template detail; keep the archetype-split paragraph, which is genuinely design-level.

**frontend.md** (optional, flagged conservatively)
- The "mid-battle attributes are a modifier graph" + "per-spawn player reset" paragraphs are the doc's most implementation-specific content; if a `frontend-battle.md` is ever created (holding the battle-sim/parity sections), these are its core.
- The Accessibility section's four full patterns are a pattern catalogue — a candidate `frontend-a11y.md`, keeping only the "real elements, not `role=button` divs" rule inline.

None of these are accuracy problems — the audit verified the content is correct — purely about keeping the main docs a quick bring-up read.

<sub>From the scheduled codebase health review.</sub>

ginderjeremiah:
[...]

PR #2033, opened 2026-07-16
Extract frontend.md Accessibility catalogue into frontend-a11y.md

### LINKED ISSUE #1982: Docs devolution: consider extracting the frontend.md battle-attributes and Accessibility content
Follow-up to #1865 / #1978.

The docs devolution pass in #1865 flagged two `frontend.md` candidates that were intentionally **deferred** (and #1865 is now closed by #1978), so tracking them here so they aren't lost:

1. **Battle-attributes paragraphs** — split out into a dedicated `frontend-battle.md`. This is explicitly **conditional on that sub-doc being created first** (it does not exist yet), so this item is blocked until there's a reason to create `frontend-battle.md`.
2. **Accessibility pattern catalogue** — extraction was framed in #1865 only as "a candidate," not a firm decision. Evaluate whether the a11y catalogue is genuinely detail-level content that belongs in its own sub-doc, or whether it's better left inline in `frontend.md`.

Neither is urgent; both were correctly left out of #1978 as optional/conservative. Close this if the maintainer decides the current inline placement is fine.

Tags: tech debt, claude, scope: small.

ginderjeremiah:
## Summary

- Extracts the `frontend.md` "Accessibility" section (tooltip association, keyboard-focus re-anchoring, and the accessible-card pattern) into a new `docs/frontend-a11y.md`, mirroring the existing `frontend-overlays.md` / `frontend-tutorials.md` / `frontend-screens.md` split pattern.
- `frontend.md`'s Accessibility section is now a short pointer, consistent with how the other split-out sections read.

## How this addresses #1982

#1982 tracked two deferred docs-devolution candidates from #1865:

1. **Battle-attributes paragraphs** → explicitly **out of scope** here: the issue says this is conditional on `frontend-battle.md` existing, and it doesn't yet. Left untouched.
2. **Accessibility pattern catalogue** → addressed in this PR. The section covers four distinct, detailed patterns (control semantics, `aria-describedby` association, focus-anchor gating, and the accessible-card composition) — comparable in scope to the sections already split into their own docs — so I extracted it rather than leaving it inline.

## Test plan

- [x] Docs-only change; no build/test surface. Verified the new `[frontend-a11y.md](./frontend-a11y.md)` link resolves and no other doc/code referenced the old inline `#accessibility` anchor (`grep` across `docs/` and `UI/`).

Closes #1982

---
_Generated by [Claude Code](https://claude.ai/code/session_01MwPnxaQtZ1FpDHu4cR8FVb)_


ginderjeremiah (commented):
## Review: extract the `frontend.md` Accessibility catalogue into `frontend-a11y.md` (#1982 item 2)

**Verdict:** Comment (I'm the PR author, so GitHub caps this at `COMMENT`). My real call is **hold / needs a maintainer decision — not an auto-merge** — the change is clean, but it makes the exact organizational decision #1982 reserves for the maintainer, in the opposite direction from the issue's own triage recommendation, without addressing it.

### The decision this PR takes (should-fix / decide before merge)

#1982 item 2 was framed as "evaluate whether the a11y catalogue belongs in its own sub-doc, or is better left inline." The issue's [triage comment](https://github.com/ginderjeremiah/GameServer/issues/1982#issuecomment-4988089662) worked through the docs' implicit extraction heuristic and recommended the opposite of this PR:

- Sections that **were** split out (sockets, admin, overlays, tutorials, screens) are self-contained **subsystems** with their own module and granular mechanic detail — inline summary + "full mechanics live in X.md" pointer.
- Sections that **stay inline** (Battle, Auth, Reference Data, Styling, Conlang, Navigation, Accessibility) are **cross-cutting conventions** applied across many unrelated components, all in the same short length range.

By that heuristic Accessibility (9 lines / 4 paragraphs) is a peer of **Styling** (~10 lines) — cross-cutting conventions, not a subsystem — and the recommendation was *leave it inline, close item 2 as "current placement is fine."* This PR extracts it instead, and the PR body justifies that on scope/detail ("comparable in scope to the sections already split") without engaging the subsystem-vs-convention distinction that the triage used, or the prior recommendation at all.

I lean toward the triage's reasoning: the a11y content documents conventions that recur across many components (how any control is made accessible, how tooltips associate across all triggers, the card pattern used by all card-like tiles) — that's the cross-cutting trait, not a subsystem with its own module. The content *is* denser than Styling, so it's a genuinely close call, which is exactly why it should be a conscious maintainer decision rather than a silent merge.

Two clean ways to resolve:
1. **Keep inline** (per the triage) and instead close #1982 item 2 with that rationale; or
2. **Proceed with the extraction**, but update the PR body / issue to explicitly supersede the triage recommendation and say why the subsystem-vs-convention line doesn't apply here.

Either is fine — but it should be decided, not defaulted into.

### Note on scope: #1982 also tracks item 1

`Closes #1982` will auto-close the issue on merge, but item 1 (battle-attributes → `frontend-battle.md`, blocked until that doc exists) is untouched and still valid. If this merges as-is, item 1's tracking needs to be preserved in a follow-up issue first so it isn't lost.

### What's good

- The extraction itself is faithful — content moved verbatim, the parent section correctly reduced to a one-line rule + pointer, mirroring the existing split sections mechanically.
- Verified: `frontend-a11y.md` link resolves, no dangling `#accessibility` anchor references remain anywhere in `docs/` or `UI/`, and the battle-attributes item was correctly left out of scope.
- CI is appropriately green (build/test jobs skipped by `detect-changes` for a docs-only change; gating checks passed).

_No blocking findings — docs-only and fully reversible. Holding the merge only because the inline-vs-extract call is the maintainer's per the issue, and it currently contradicts the issue's own recommendation._

---

ginderjeremiah:
Good catch — I picked this up without reading #1982's existing triage comment (posted a couple hours before I started), which had already worked through the subsystem-vs-convention heuristic and landed on the opposite call. On reflection I agree with that triage: Accessibility documents conventions applied across many unrelated components (any control's semantics, tooltip association on any trigger, the card pattern used by any card-like tile), not a subsystem with its own store/host/module the way Overlays (toast/modal/popover stores+hosts), Tutorials (the tutorials engine), or Screens (the screen registry) are. Length/detail alone isn't the right test — I was wrong to lean on that in the PR body.

Closing this without merging; keeping the Accessibility section inline. Following up on #1982 with that resolution.

---

PR #2070, opened 2026-07-16
Arm the keepalive interval before the pre-emptive refresh in openSocket (#2035)

### LINKED ISSUE #2035: Socket open path can strand queued commands forever on a retryable refresh failure (keepalive never armed)
## Context

Found in the July 2026 health review (frontend socket/auth pass).

`openSocket()` pre-emptively refreshes the access token *before* creating the WebSocket and *before* arming the keepalive interval. On a **retryable** refresh failure (network blip, 5xx/429) it bails early and deliberately leaves the queue intact — the inline comment says "the keepalive ping's next `ensureSocket()` call retries the connect". But `pingIntervalId` is only ever set at the *end* of `openSocket()` (`UI/src/lib/api/api-socket.ts:174-176`), after that bail point.

So whenever the interval isn't already running — the **first connect of a page session** with an expired access token (e.g. a reload hours later), or any connect after a `NORMAL_CLOSURE` stopped the interval — nothing retries. A queued command is never sent, never gets a timeout armed (timeouts are only armed on send), and is never settled. The caller `await`s forever, violating the layer's own "pending and queued commands are always settled" contract (frontend.md § Socket request &amp; connection lifecycle). A later user action that triggers another `sendSocketCommand` incidentally recovers it, but a boot-time hang has no such second trigger.

**Refs:** `UI/src/lib/api/api-socket.ts:131-149` (retryable bail), `:174-176` (interval armed after it), `:310-315` (timeout armed only on send).

## Suggested fix

Arm the keepalive interval at the top of `openSocket()` (before the refresh await), or on the retryable bail schedule a one-shot retry / settle the queue with the retryable error so callers surface it. Add a unit test: queued command + retryable refresh outcome + no running interval → the command must settle or a retry must be scheduled.

ginderjeremiah:
## Summary

`openSocket()` pre-emptively refreshes the access token *before* creating the WebSocket and *before* arming the keepalive interval (`pingIntervalId`). On a **retryable** refresh failure (network blip, 5xx/429) it bails early and deliberately leaves the queue intact — the inline comment says "the keepalive ping's next `ensureSocket()` call retries the connect." But since `pingIntervalId` was only ever set at the *end* of `openSocket()`, after that bail point, nothing was actually there to retry whenever the interval wasn't already running: the first connect of a page session with an expired access token, or any connect after a `NORMAL_CLOSURE` stopped the interval. A queued command would never be sent, never get a timeout armed, and never settle — the caller would `await` forever, violating the socket layer's "pending and queued commands are always settled" contract (`docs/frontend.md` § Socket request & connection lifecycle).

## Fix

Move the keepalive-interval arming to the top of `openSocket()`, before the pre-emptive refresh await, so a retryable bail still leaves the interval running to drive the retry. The interval-arming block that previously ran at the end of `openSocket()` (right before the handshake was opened) is removed since it's now redundant. The existing `stopPingInterval()` call on the terminal (definitively-rejected, no adoptable pair) branch still correctly tears the freshly-armed interval back down.

## Test plan

- [x] Added a test asserting the keepalive interval is armed even when the pre-emptive refresh bails as retryable.
- [x] Added a test reproducing the boot-time hang: a retryable refresh failure on the very first connect (no interval running yet), followed by the next keepalive tick successfully retrying the connect.
- [x] `npx vitest run src/tests/lib/api/api-socket.test.ts` — 55/55 passed.
- [x] `npm run test:unit:ci` — 3729/3729 passed.

Closes #2035

https://claude.ai/code/session_01S8ZCUhkQGDCto7zatuKeLw

---
_Generated by [Claude Code](https://claude.ai/code/session_01S8ZCUhkQGDCto7zatuKeLw)_

### DIFF
--- a/UI/src/lib/api/api-socket.ts
+++ b/UI/src/lib/api/api-socket.ts
@@ -131,6 +131,13 @@ export class ApiSocket {
 	private async openSocket(): Promise<void> {
 		this.socketOpened = false;
 		this.resetPongTracking();
+		// Armed before the refresh await (not after the handshake is built) so a retryable bail below still
+		// leaves the keepalive running to retry the connect — otherwise the very first connect of a page
+		// session (or any connect after the interval was last stopped) could bail with nothing left to ever
+		// call ensureSocket() again, stranding the queue forever.
+		if (this.pingIntervalId === null) {
+			this.pingIntervalId = setInterval(() => this.attemptPing(), PING_INTERVAL_MS);
+		}
 		// Pre-emptively refresh an access token that is missing or about to expire (mirroring the HTTP
 		// path) so a reconnect doesn't hand the server a stale token, eat a rejected handshake, and burn a
 		// single-use refresh token recovering in handleClose.
@@ -171,9 +178,6 @@ export class ApiSocket {
 		socket.onmessage = this.receiveResponse.bind(this);
 		socket.onerror = this.handleError.bind(this);
 		socket.onclose = this.handleClose.bind(this);
-		if (this.pingIntervalId === null) {
-			this.pingIntervalId = setInterval(() => this.attemptPing(), PING_INTERVAL_MS);
-		}
 	}
 
 	public async sendSocketCommand<T extends ApiSocketCommandNoRequest>(commandName: T): Promise<IApiSocketResponse<T>>;
--- a/UI/src/tests/lib/api/api-socket.test.ts
+++ b/UI/src/tests/lib/api/api-socket.test.ts
@@ -210,6 +210,33 @@ describe('ApiSocket', () => {
 			expect(handleAuthFailure).not.toHaveBeenCalled();
 			expect(webSocketMock).not.toHaveBeenCalled();
 			expect(getTokens()).toEqual({ accessToken: 'expired', refreshToken: 'r' });
+			// The keepalive must still be armed so its next tick retries the connect — otherwise, on the very
+			// first connect of a page session (no interval running yet), nothing would ever call
+			// ensureSocket() again and the queued command would await forever (#2035).
+			expect(internals(apiSocket).pingIntervalId).not.toBeNull();
+		});
+
+		it('retries the connect via the keepalive after a retryable refresh failure with no interval running yet', async () => {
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: hoist keepalive arming above the pre-emptive refresh so a retryable bail can still retry the connect (#2035)

**Verdict:** Approve _(submitted as a comment since I'm the PR author)_ — the fix correctly addresses the stranded-queue contract violation, is minimal and well-scoped, and the reordering's one real hazard is explicitly tested.

### Correctness
- Arming `pingIntervalId` at the top of `openSocket()` (before the refresh `await`) is the right fix: on a retryable bail (`api-socket.ts:150-156`) the interval is now already running, so its next `attemptPing()` → `ensureSocket()` retries the connect. This closes the boot-time hang where the very first connect of a page session with an expired token had nothing left to re-drive the queue.
- The reordering's one hazard — the newly-hoisted interval leaking on the terminal, definitively-rejected branch — is handled: that branch still calls `stopPingInterval()` before `handleAuthFailure()` (`api-socket.ts:166`), and `attemptPing()` self-stops once tokens are gone (`:217-220`), so the keepalive can't resurrect a socket for an ended session.
- The old arming block at the end of `openSocket()` was removed rather than left as dead code. The `if (pingIntervalId === null)` guard keeps it idempotent.

### Test coverage
- Strong. The retryable path asserts the interval stays armed (`:216`), and a new dedicated test reproduces the exact #2035 boot-time hang — retryable failure on the first connect with no interval running, then a keepalive tick that successfully retries (`:219-240`).
- Crucially, the terminal-rejection test now also asserts `pingIntervalId` is `null` afterward (`:180`), which pins down that the hoisted interval is torn back down — the precise regression the reordering could have introduced.

### What's good
- The 4-line comment at `api-socket.ts:134-137` is well justified: it names the ordering as deliberate and warns against the naive "move it back down" simplification, which is exactly the kind of foot-gun a future edit would otherwise reintroduce.

Clean, contained bug fix with the right tests. No blocking, should-fix, or nit findings.

---

PR #2071, opened 2026-07-16
Guard progress-store writes against in-flight fetches from a discarded session

### LINKED ISSUE #2036: In-flight player-progress fetch can repopulate a store after reset() — previous character's data can leak into the next session
## Context

`CoalescedLoader.reset()` epoch-guards only the *queued* fresh fetch (`UI/src/lib/common/coalesced-loader.ts:44-51`), but the original in-flight `fetchFn()`'s own continuation is unguarded — and the store writes live inside it:

```ts
proficiencies = (await fetchSocketData('GetPlayerProficiencies')) ?? [];
error = false;
loaded = true;
```

Same pattern in all three progress stores: `UI/src/stores/proficiencies.svelte.ts:17-28`, `challenges.svelte.ts:14-25`, `statistics.svelte.ts:14-25`.

**Failure scenario:** `handleSocketReplaced` → `stopGame()` runs synchronously: `playerProficiencies.reset()` (clears state, `loaded = false`), *then* `apiSocket.disconnect()`.

1. **Common/cosmetic:** `disconnect()` settles the pending command with an error, so the `catch` sets `error = true` *after* `reset()` cleared it — the next session's screen starts in an error state.
2. **Narrow/serious:** if the response already resolved but its continuation hasn't run yet, the continuation reassigns the store with the **previous character's data and sets `loaded = true`**. The next session's `startGame` → `void playerProficiencies.load()` no-ops on `isLoaded()`, so the live battler is composed with the *previous* character's `battleModifiers` — a parity/anti-cheat divergence, not just a display bug.

## Suggested fix

Give each store (or the loader) a session epoch the fetcher checks before writing — e.g. `reset()` bumps a token captured at the top of the fetch fn, and the post-`await` writes are skipped when it moved. `coalesced-loader.test.ts` covers the queued-fetch abandonment but not this in-flight-write case; add a store-level test (reset while a fetch promise is resolving).

ginderjeremiah:
## Summary

`CoalescedLoader.reset()` (`UI/src/lib/common/coalesced-loader.ts`) only epoch-guarded the *queued* fresh fetch triggered by a mid-flight `force`. The *original* in-flight `fetchFn()`'s own continuation — where the actual store write happens — ran unconditionally.

`stopGame()` calls `playerProficiencies.reset()` / `playerChallenges.reset()` / `statistics.reset()` synchronously and then disconnects the socket. If one of those stores had a fetch already in flight when `reset()` ran (e.g. a character switch mid-load), its continuation could still resolve afterward and write the **previous character's** data into the store — with `loaded` flipped back to `true` — so the next session's `load()` call would silently no-op and the live battler could compose against stale `battleModifiers`.

## Fix

- `CoalescedLoader` exposes a new `currentEpoch` getter, backed by the same epoch counter `reset()` already bumps.
- Each of the three progress stores (`proficiencies.svelte.ts`, `challenges.svelte.ts`, `statistics.svelte.ts`) captures `loader.currentEpoch` before its fetch's `await`, and re-checks it before writing to store state (success path) or setting `error` (failure path). If `reset()` ran in between, the epoch has moved and the write is dropped.

This follows the "give each store (or the loader) a session epoch" direction suggested in the issue — implemented once in the shared loader rather than three independent counters, since all three stores share the exact same race.

## Test plan

- Added `CoalescedLoader.currentEpoch` unit test (loader test suite).
- Added a store-level regression test to each of the three store suites: start a load, `reset()` before it resolves, start a fresh load, then resolve the stale fetch with different data — asserts the stale write is dropped and the fresh session's data wins.
- `npm run lint` — clean.
- Full frontend unit suite: `npx vitest run` — 299 files / 3733 tests passing.

Backend is untouched; this is a frontend-only store/lifecycle fix.

Closes #2036

https://claude.ai/code/session_01R4KUV7MqHghZnCfZpFKLUX

---
_Generated by [Claude Code](https://claude.ai/code/session_01R4KUV7MqHghZnCfZpFKLUX)_

### DIFF
--- a/UI/src/lib/common/coalesced-loader.ts
+++ b/UI/src/lib/common/coalesced-loader.ts
@@ -60,6 +60,13 @@ export class CoalescedLoader {
 		this.epoch += 1;
 	}
 
+	/** Current reset epoch, bumped by {@link reset}. `fetchFn` should capture this before its first
+	 *  `await` and compare it again before writing to store state: if it moved, a `reset()` ran
+	 *  mid-flight and the pending write belongs to a discarded session, so it must be dropped. */
+	get currentEpoch(): number {
+		return this.epoch;
+	}
+
 	/** Publishes `pipeline` as the awaitable in-flight load, clearing it once it settles unless a
 	 *  chained fetch has replaced it in the meantime. */
 	private track(pipeline: Promise<void>): Promise<void> {
--- a/UI/src/lib/common/coalesced-loader.ts
+++ b/UI/src/lib/common/coalesced-loader.ts
@@ -61,10 +61,15 @@ export class CoalescedLoader {
 	}
 
 	/** Current reset epoch, bumped by {@link reset}. `fetchFn` should capture this before its first
-	 *  `await` and compare it again before writing to store state: if it moved, a `reset()` ran
-	 *  mid-flight and the pending write belongs to a discarded session, so it must be dropped. */
+	 *  `await` and pass it to {@link isStale} afterward to detect a `reset()` that ran mid-flight. */
 	get currentEpoch(): number {
 		return this.epoch;
+	}
+
+	/** Whether `epoch` (captured from {@link currentEpoch} before an await) no longer matches — a
+	 *  `reset()` ran mid-flight, so a write gated on this epoch belongs to a discarded session. */
+	isStale(epoch: number): boolean {
+		return epoch !== this.epoch;
 	}
 
 	/** Publishes `pipeline` as the awaitable in-flight load, clearing it once it settles unless a
--- a/UI/src/stores/proficiencies.svelte.ts
+++ b/UI/src/stores/proficiencies.svelte.ts
@@ -15,15 +15,23 @@ let loaded = $state(false);
 let error = $state(false);
 
 const fetchProficiencies = async () => {
+	const epoch = loader.currentEpoch;
 	try {
 		// fetchSocketData throws on a socket error, preserving this try/catch contract.
-		proficiencies = (await fetchSocketData('GetPlayerProficiencies')) ?? [];
+		const result = (await fetchSocketData('GetPlayerProficiencies')) ?? [];
+		if (loader.currentEpoch !== epoch) {
+			// reset() ran while this fetch was in flight; this write belongs to a discarded session.
+			return;
+		}
+		proficiencies = result;
 		error = false;
 		loaded = true;
 	} catch {
 		// A failed load is not a genuine empty result; leave the live battler without proficiency
 		// bonuses and let the next load retry. The proficiency screen surfaces the error to the user.
-		error = true;
+		if (loader.currentEpoch === epoch) {
+			error = true;
+		}
 	}
 };
 
--- a/UI/src/stores/proficiencies.svelte.ts
+++ b/UI/src/stores/proficiencies.svelte.ts
@@ -19,7 +19,7 @@ const fetchProficiencies = async () => {
 	try {
 		// fetchSocketData throws on a socket error, preserving this try/catch contract.
 		const result = (await fetchSocketData('GetPlayerProficiencies')) ?? [];
-		if (loader.currentEpoch !== epoch) {
+		if (loader.isStale(epoch)) {
 			// reset() ran while this fetch was in flight; this write belongs to a discarded session.
 			return;
 		}
@@ -29,7 +29,7 @@ const fetchProficiencies = async () => {
 	} catch {
 		// A failed load is not a genuine empty result; leave the live battler without proficiency
 		// bonuses and let the next load retry. The proficiency screen surfaces the error to the user.
-		if (loader.currentEpoch === epoch) {
+		if (!loader.isStale(epoch)) {
 			error = true;
 		}
 	}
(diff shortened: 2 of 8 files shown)


UI/src/stores/proficiencies.svelte.ts
```diff
@@ -15,15 +15,23 @@ let loaded = $state(false);
...
+		}
+		proficiencies = result;
 		error = false;
 		loaded = true;
 	} catch {
 		// A failed load is not a genuine empty result; leave the live battler without proficiency
 		// bonuses and let the next load retry. The proficiency screen surfaces the error to the user.
-		error = true;
+		if (loader.currentEpoch === epoch) {
+			error = true;
```
ginderjeremiah: **Should-fix (test coverage):** This catch-path guard fixes the issue's failure mode #1 — the "common/cosmetic" case where `disconnect()` settles the pending command with an error and the `catch` would otherwise set `error = true` after `reset()` already cleared it. But none of the added tests exercise it: all three store tests cover only the *resolve* path (mode #2). If this `if (loader.currentEpoch === epoch)` guard were deleted, no test would fail.

Please add a store-level test asserting that a fetch which **rejects** after `reset()` does not leave `error === true` (and, to keep the negative meaningful, that a normal rejection with no intervening reset still sets it). The existing `flags an error...` test plus the new reset-during-resolve test bracket this nicely; the gap is specifically reset-during-reject. Applies to all three stores per the project's mirror-the-tests convention.

---

UI/src/lib/common/coalesced-loader.ts
```diff
@@ -60,6 +60,13 @@ export class CoalescedLoader {
 		this.epoch += 1;
 	}
 
+	/** Current reset epoch, bumped by {@link reset}. `fetchFn` should capture this before its first
+	 *  `await` and compare it again before writing to store state: if it moved, a `reset()` ran
+	 *  mid-flight and the pending write belongs to a discarded session, so it must be dropped. */
+	get currentEpoch(): number {
+		return this.epoch;
+	}
```
ginderjeremiah, line 67: **Nit (optional):** Exposing a raw `currentEpoch: number` leaves each caller to spell out the comparison, and the two directions live one branch apart — the success path drops on `!==` while the catch keeps on `===`. That's an easy pair to invert in a later edit. Consider a boolean helper on the loader, e.g. `isStale(epoch: number): boolean` (or `isCurrent`), so the "was this session discarded?" semantics live in one place and each store reads `if (loader.isStale(epoch)) { return; }`. Purely a readability/footgun call — the current code is correct.

---

ginderjeremiah (commented):
## Review: Guard progress-store writes against in-flight fetches from a discarded session (#2036)

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't take a gating event from me)* — the fix is correct and well-targeted; one should-fix on test coverage before merge.

The approach is the right one: a single reset epoch on the shared `CoalescedLoader` (rather than three independent counters) since all three progress stores share the identical race. The epoch is monotonic, captured before the `await`, and re-checked before both the success write and the error write — so both failure modes from #2036 are closed: the serious `loaded = true` stale-data leak (which would silently no-op the next session's `load()` and compose the battler against the previous character's `battleModifiers`), and the cosmetic `error = true` leak. I confirmed the getter is only read at call-time, so referencing `loader` before its `const` initializer is fine, and that the three progress stores are the *only* `CoalescedLoader` consumers, so nothing was missed.

### Should-fix
- **Untested error-path guard** — `stores/{proficiencies,challenges,statistics}.svelte.ts` catch block. The `if (loader.currentEpoch === epoch) { error = true; }` guard fixes the issue's failure mode #1 (the "common" one — `disconnect()` rejecting the pending command after `reset()`), but no test exercises it; the added tests only cover reset-during-*resolve*. Deleting the guard wouldn't fail any test. Add a reset-during-*reject* test to each store (see inline comment on `proficiencies.svelte.ts`).

### Nits
- **Raw-epoch footgun** — `coalesced-loader.ts:66`. The loader hands out a bare `number` and each store spells out the comparison, with the two directions (`!==` drop / `===` keep) a branch apart. A boolean helper (`isStale(epoch)`) would keep the "discarded session?" semantics in one place. Optional; current code is correct. (Inline.)

### Test coverage
- New `CoalescedLoader.currentEpoch` unit test — good.
- New reset-during-resolve regression test in all three store suites, mirrored consistently — good, and covers the serious failure mode.
- Gap: the error-path guard (failure mode #1) is unverified — the should-fix above.

### What's good
- Fix lands once in the shared abstraction instead of three ad-hoc counters, matching the existing `epoch`/`chainEpoch` machinery already in `reset()`.
- Comments explain the *why* (discarded-session write) at each guard without over-narrating.
- The regression tests use `Promise.withResolvers` to interleave reset and resolution deterministically — exactly the scenario the issue described.

Not merging yet: the error-path test should land first (or be consciously declined), and CI (`test-frontend` / e2e) is still in progress.

---

UI/src/stores/proficiencies.svelte.ts
ginderjeremiah: Added a reset-during-reject test to each store (`does not flag an error when an in-flight fetch rejects after reset() ran`) in db719d7. Combined with the existing `flags an error...` test (no intervening reset), both directions of the catch-path guard are now covered.

---

UI/src/lib/common/coalesced-loader.ts
ginderjeremiah: Done — added `CoalescedLoader.isStale(epoch)` and switched all three stores to it in db719d7.

---

PR #2072, opened 2026-07-16
Reassign unlockedItems/unlockedMods in InventoryManager.initialize()

### LINKED ISSUE #2037: InventoryManager.initialize() mutates unlockedItems/unlockedMods in place — reactive consumers don't observe a mid-session resync
## Context

`statify`'s documented contract is that `Map`/`Set` fields must be **reassigned** to signal a change. `addUnlockedMod` honours it (`UI/src/lib/engine/player/inventory-manager.ts:373-381` — "Reassign (not mutate) the Set so statify/$state tracks the change"), but `initialize()` — the mid-session resync path — does not:

```ts
this.unlockedItems.clear();
this.unlockedMods.clear();
```

(`inventory-manager.ts:99-109`; it rebuilds both containers in place, while `equippedSlots` *is* reassigned.)

**Failure scenario:** the player is on the Inventory screen with the mod picker open when a `DefeatEnemy` response is lost (`enemy-manager.ts` resync) or a `ChallengeCompleted` push dead-letters (`engine.ts` → `resyncPlayerAndInventory()`). `initialize()` rebuilds the sets in place, so:
- `compatibleMods` (`inventory-view.svelte.ts` reads `inventoryManager.unlockedMods.has(m.id)`) doesn't re-derive — a mod delivered by the resync stays invisible;
- `selected`/`dragItem` keep returning the cached, orphaned pre-resync `Item` instance — exactly the object class #1808 works to keep out of live state.

## Suggested fix

In `initialize()`, build fresh locals and reassign `this.unlockedItems`/`this.unlockedMods` (mirroring the existing `equippedSlots` reassign). Add a case to `inventory-manager-reactivity.svelte.test.ts`, which currently covers field mutations (favorite/equip/mod) but not the initialize/resync identity contract.

ginderjeremiah:
## Summary

`InventoryManager.initialize()` — the mid-session resync path — rebuilt `unlockedItems`/`unlockedMods` **in place** via `.clear()` + re-populate, while `equippedSlots` was correctly reassigned. `statify`'s documented contract requires `Map`/`Set` fields to be **reassigned** to signal a change; `addUnlockedMod` already follows this, but `initialize()` didn't.

**Failure scenario:** the player is on the Inventory screen with the mod picker open when a `DefeatEnemy` response is lost or a `ChallengeCompleted` push dead-letters, triggering a resync via `resyncPlayerAndInventory()`. Because the containers were mutated in place, reactive consumers that read them directly (e.g. `compatibleMods` reading `inventoryManager.unlockedMods.has(m.id)`) wouldn't re-derive, so a mod/item delivered by the resync stayed invisible until something else forced a rebuild.

## Fix

- `initialize()` now builds fresh local `Map`/`Set` instances and reassigns `this.unlockedItems` / `this.unlockedMods` at the end, mirroring the existing `equippedSlots` reassignment and the reassign pattern already used by `addUnlockedMod`.
- Added a test case to `inventory-manager-reactivity.svelte.test.ts` covering the initialize/resync identity contract: a `$derived` reading `unlockedItems.size` and `unlockedMods.has(...)` now correctly observes a second `initialize()` call delivering new data.

## Test plan

- [x] `npx vitest run` on the touched test files — 73 tests passing
- [x] Full frontend unit suite: `npm run test:unit:ci` — 299 files / 3730 tests passing

Backend is untouched; this is a frontend-only store/lifecycle fix.

Closes #2037

---
_Generated by [Claude Code](https://claude.ai/code/session_01AsXi3ZmdLNiEtyRpCxDrc8)_

### DIFF
--- a/UI/src/lib/engine/player/inventory-manager.ts
+++ b/UI/src/lib/engine/player/inventory-manager.ts
@@ -98,15 +98,15 @@ export class InventoryManager {
 
 	public initialize() {
 		this.generation++;
-		this.unlockedItems.clear();
-		this.unlockedMods.clear();
-		this.equippedSlots = new Array(6).fill(undefined);
+		const unlockedItems: Map<number, Item> = new Map();
+		const unlockedMods: Set<number> = new Set();
+		const equippedSlots: (Item | undefined)[] = new Array(6).fill(undefined);
 
 		const data = playerManager.inventoryData;
 
 		// Load unlocked mods
 		for (const modId of data.unlockedMods) {
-			this.unlockedMods.add(modId);
+			unlockedMods.add(modId);
 		}
 
 		// Load unlocked items
@@ -117,13 +117,18 @@ export class InventoryManager {
 				logMessage(ELogType.Debug, `Skipped unlocked item with unknown id ${invItem.itemId}.`);
 				continue;
 			}
-			this.unlockedItems.set(invItem.itemId, item);
+			unlockedItems.set(invItem.itemId, item);
 
 			// Place equipped items into their equipment slots
 			if (invItem.equipped && invItem.equipmentSlotId != null) {
-				this.equippedSlots[invItem.equipmentSlotId] = item;
+				equippedSlots[invItem.equipmentSlotId] = item;
 			}
 		}
+
+		// Reassign (not mutate) so statify/$state tracks the change, mirroring addUnlockedMod.
+		this.unlockedItems = unlockedItems;
+		this.unlockedMods = unlockedMods;
+		this.equippedSlots = equippedSlots;
 
 		this.publish();
 		this.refreshEquipmentStats();
--- a/UI/src/routes/game/screens/inventory/inventory-view.svelte.ts
+++ b/UI/src/routes/game/screens/inventory/inventory-view.svelte.ts
@@ -76,8 +76,9 @@ export class InventoryView {
 
 	// Resolve selection/drag through the manager's itemId-keyed Map (its own O(1) access convention)
 	// rather than a linear scan of the list; `items` stays the source for list/count derivations.
-	// NOTE: unlockedItems is a plain (non-reactive) Map, so these re-resolve only on id changes, not on
-	// set changes — fine while items are add-only; if a removal path lands, clear the ids on removal.
+	// NOTE: unlockedItems is a plain (non-reactive) Map, so these re-resolve on a reassignment
+	// (initialize()'s resync) but not on an in-place add — fine while addUnlockedItem is add-only;
+	// if a removal path lands, clear the ids on removal.
 	readonly selected = $derived(
 		this.selectedId != null ? (inventoryManager.unlockedItems.get(this.selectedId) ?? null) : null
 	);
--- a/UI/src/tests/lib/engine/player/inventory-manager-reactivity.svelte.test.ts
+++ b/UI/src/tests/lib/engine/player/inventory-manager-reactivity.svelte.test.ts
@@ -175,4 +175,36 @@ describe('InventoryManager reactivity under statify (#1957)', () => {
 
 		cleanup();
 	});
+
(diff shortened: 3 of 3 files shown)


ginderjeremiah (commented):
## Review: InventoryManager.initialize() reassigns unlockedItems/unlockedMods so a mid-session resync is observed (#2037)

**Verdict:** Approve (submitted as a comment since I'm the PR author) — the fix is correct, minimal, matches the issue's suggested approach, and is genuinely guarded by a new test.

### Why this is correct
`statify` makes the *field reference* reactive, not the `Map`/`Set` internals — an in-place `.clear()`/`.add()` doesn't notify, only a reassignment does (`frontend-screens.md` → "rebuilds whenever the item *set* changes"; the existing `addUnlockedMod` already reassigns for this exact reason). Building fresh locals and assigning `unlockedItems`/`unlockedMods`/`equippedSlots` once at the end is the right pattern. It resolves the real failure scenario: the consumers that read these fields directly in a `$derived` — `selected`/`dragItem` (`inventory-view.svelte.ts:81-87`) and `compatibleMods` (`:249`) — now re-derive on a resync, where before they held orphaned pre-resync state.

As a bonus, assigning `equippedSlots` once at the end is cleaner than the prior reassign-then-index-mutate.

### Test coverage
Good. The new `inventory-manager-reactivity.svelte.test.ts` case exercises the initialize→resync identity contract through a `$derived` reading `unlockedItems.size` / `unlockedMods.has(...)`, and would fail against the old in-place `.clear()` code (the field reference never changed), so it genuinely guards the fix rather than just passing. It fits the file's existing `#1957` reactivity-under-statify theme. No backend change is involved (frontend-only lifecycle fix), so no mirrored suite is needed.

### Nits
- `inventory-view.svelte.ts:79-80` (not in this diff): the comment "unlockedItems is a plain (non-reactive) Map, so these re-resolve only on id changes, not on set changes" is now slightly stale — after this PR, an `initialize()` reassignment *does* refresh `selected`/`dragItem` on resync. The Map contents are still non-reactive (the add-only `addUnlockedItem` in-place `.set()` path the comment guards remains true), so this is minor, but a one-line tweak clarifying that reassignment (resync) refreshes while in-place adds don't would keep a future reader from the wrong conclusion. Fine as a tiny follow-up or a drive-by.

### What's good
Tightly scoped, mirrors the established reassign pattern, well-commented, and the test proves the behavior rather than asserting it. Clean fix.

---

ginderjeremiah:
## Review pass (independent) — Approve ✅

Re-reviewed after the follow-up commit `18651a6` landed on top of the earlier review.

**Core fix (`e63f4c2`)** — Correct and exactly matches issue #2037's suggested approach and scope (bug / scope:small). `initialize()` now builds fresh `Map`/`Set`/array locals and reassigns `unlockedItems`/`unlockedMods`/`equippedSlots` at the end instead of the in-place `.clear()`, honoring statify's reassign-to-signal-change contract that `addUnlockedMod` already follows. This resolves the real failure: a mid-session resync (`resyncPlayerAndInventory()`) now re-derives `compatibleMods` and re-resolves `selected`/`dragItem` off fresh instances rather than leaving orphaned pre-resync objects live.

The new `inventory-manager-reactivity.svelte.test.ts` case genuinely guards the fix — it drives a `$derived` on `unlockedItems.size` / `unlockedMods.has(...)` through a second `initialize()` and would fail against the old in-place code (the field reference never changed). It's not just an assertion that passes trivially; it's empirical proof reassignment triggers reactivity.

**Follow-up commit (`18651a6`)** — Correctly addresses the sole nit from the prior review: the `inventory-view.svelte.ts` comment now accurately states the `$derived` re-resolves on a reassignment (`initialize()`'s resync) but not on an in-place add. Accurate given this PR's change.

Frontend-only lifecycle fix, no backend/mirrored-suite involvement. CI fully green (`all-checks-passed` success; backend jobs correctly skipped). No blocking findings, no outstanding nits. Merging.

---
_Reviewed by [Claude Code](https://claude.ai/code)_

---

PR #2073, opened 2026-07-16
Clear toasts, modals, navigation intent, and current enemy on stopGame teardown

### LINKED ISSUE #2038: stopGame teardown misses the module-level UI stores — toasts, navigation intent, and the last enemy leak across sessions
## Context

`stopGame` (`UI/src/lib/engine/engine.ts:357-367`) resets statistics/challenges/proficiencies/logs and disconnects, but nothing else:

- `clearToasts()` (`UI/src/stores/toast.svelte.ts:89-95`, comment says "for navigation resets") and `clearModals()` (`modal.svelte.ts:104-110`) are **never called**.
- `navigation.svelte.ts:7-12` — `requestedScreen`/`pendingPayload` are module state with no reset at all.
- `EnemyManager.stop()` leaves `currentEnemy` set (`enemy-manager.ts:140-151`), so the previous session's enemy briefly renders on re-entry until the first fetch replaces it.

**Failure scenario:** `SocketReplaced` mid-game → `stopGame()` → `goto('/')` (client-side; the root layout and `ToastContainer` persist). Live game toasts survive onto the login screen. Clicking a leftover "View" action there calls `navigation.requestScreen('challenges')` (`engine.ts:189-191`) — which nothing clears — so the **next** session's game shell `$effect` consumes it on mount and opens the fresh session on the Challenges screen (with `evaluateScreenTrigger` consequences) instead of Fight. The same applies to a `requestScreen` queued just before teardown that the shell never consumed.

## Suggested fix

Extend `stopGame` to `clearToasts()`, `clearModals()` (safe: it runs *before* `acknowledgeModal` in `handleSocketReplaced`), add and call a `navigation.reset()` (clear `requestedScreen` + payload), and clear `enemyManager.currentEnemy` in `stop()`. Add a test asserting `stopGame` leaves the toast/modal/navigation stores clean.

ginderjeremiah:
## Summary

`stopGame` (`UI/src/lib/engine/engine.ts`) reset statistics/challenges/proficiencies/logs and disconnected the socket, but left several other module-level UI stores untouched:

- `clearToasts()` and `clearModals()` were never called on teardown.
- `navigation.svelte.ts`'s `requestedScreen`/`pendingPayload` had no reset path at all.
- `EnemyManager.stop()` left `currentEnemy` set, so the previous session's enemy could briefly render on re-entry until the first fetch replaced it.

**Failure scenario:** `SocketReplaced` mid-game → `stopGame()` → client-side `goto('/')` (the root layout and `ToastContainer` persist). Live game toasts survive onto the login screen, and a queued `navigation.requestScreen(...)` (e.g. from a toast's "View" action) would be silently consumed by the *next* session's game shell on mount, opening it on the wrong screen instead of Fight.

## Changes

- Added `navigation.reset()` (`UI/src/stores/navigation.svelte.ts`) to clear both the requested screen and any unconsumed payload.
- `EnemyManager.stop()` now clears `currentEnemy` (`UI/src/lib/engine/battle/enemy-manager.ts`).
- `stopGame()` now calls `clearToasts()`, `clearModals()`, and `navigation.reset()` in addition to its existing resets (`UI/src/lib/engine/engine.ts`). `clearModals()` runs before `handleSocketReplaced`'s own `acknowledgeModal(...)` call, so the new session-replaced dialog is unaffected.

## Test plan

- [x] Added unit tests covering `navigation.reset()`, `EnemyManager.stop()` clearing `currentEnemy` (and remaining a no-op when never started), and `stopGame`'s new toast/modal/navigation clearing (`enemy-manager.test.ts`, `engine.test.ts`, `navigation.test.ts`).
- [x] `npx vitest run` — full suite (299 files / 3744 tests) passes.
- [x] `npm run lint` — eslint + svelte-check clean.

Closes #2038

---
_Generated by [Claude Code](https://claude.ai/code/session_01HuLeBJyZt5HL69cHRMiDXH)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -147,6 +147,9 @@ export class EnemyManager {
 			// Teardown (screen unmount / session end), not a user intent change, so don't sync the persisted
 			// loop mode — clobbering it to idle here would lose a disconnecting boss-farmer's mode.
 			this.returnToIdle(false);
+			// Otherwise the previous session's enemy would briefly render on the next session's re-entry,
+			// until the first fetch replaces it.
+			this.currentEnemy = undefined;
 		}
 	}
 
--- a/UI/src/lib/engine/engine.ts
+++ b/UI/src/lib/engine/engine.ts
@@ -30,7 +30,9 @@ import {
 	acknowledgeModal,
 	resetLogs,
 	toastSuccess,
-	navigation
+	navigation,
+	clearToasts,
+	clearModals
 } from '$stores';
 import {
 	apiSocket,
@@ -360,6 +362,12 @@ const stopGame = () => {
 	playerChallenges.reset();
 	playerProficiencies.reset();
 	resetLogs();
+	// Clear the other module-level UI stores stopEngines/the resets above don't touch, so a leftover toast,
+	// queued modal, or pending screen request can't leak into the next session (e.g. a "View" toast action
+	// queuing a navigation.requestScreen the next session's shell would silently consume on mount).
+	clearToasts();
+	clearModals();
+	navigation.reset();
 	// SocketReplaced routes back to login client-side (no reload), so the socket singleton survives. Tear
 	// it down explicitly, otherwise the keepalive ping would silently reconnect and fight the session that
 	// just took over.
--- a/UI/src/stores/navigation.svelte.ts
+++ b/UI/src/stores/navigation.svelte.ts
@@ -49,5 +49,14 @@ export const navigation = {
 	 *  target screen to consume on mount). */
 	clear() {
 		requestedScreen = null;
+	},
+
+	/** Clears every piece of navigation intent — the requested screen and any unconsumed payload.
+	 *  Used on game teardown so a request queued just before disconnect (e.g. a toast's "View" action)
+	 *  doesn't get silently consumed by the next session's game shell on mount. */
+	reset() {
+		requestedScreen = null;
+		pendingPayload = undefined;
+		payloadPending = false;
 	}
 };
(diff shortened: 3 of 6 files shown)


ginderjeremiah (commented):
## Review: stopGame teardown now clears toasts, modals, navigation intent, and the last enemy (#2038)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change faithfully implements the fix described in #2038, and the reasoning holds up against the code.

Verified:
- **Ordering is safe.** `stopGame()` (and therefore `clearModals()`) runs before `acknowledgeModal(...)` in `handleSocketReplaced` (`engine.ts:319-324`), so the new "Session Replaced" dialog is enqueued after the clear rather than being wiped by it. The added modal test exercises exactly this — a prior modal resolves `false`, and the session-replaced dialog survives.
- **Clearing `currentEnemy` is a pre-existing, handled state.** Consumers guard it (`BattlerCard.svelte:32`, `BossBattlerCard.svelte:15`, `?.` chaining), and `stop()` already assigned `undefined` at other points, so nulling it on teardown just yields an empty card until the first fetch — the intended behavior.
- **`navigation.reset()` is distinct from `clear()`, not redundant.** `clear()` deliberately leaves the payload for the target to consume post-switch; `reset()` is the full teardown clear (screen + payload + `payloadPending`). Both docstrings make the distinction explicit.
- **The `stop()` change stays inside the `if (this.started)` guard**, so the "no-op when never started" test correctly passes.

### Test coverage
Solid and appropriately scoped:
- `navigation.test.ts` — `reset()` clears requested screen and unconsumed payload.
- `enemy-manager.test.ts` — `stop()` clears `currentEnemy`, and is a no-op when never started.
- `engine.test.ts` — `stopGame` calls `clearToasts`, resets navigation, and clears a queued modal ahead of the session-replaced acknowledgement.

No battle-calculation logic is touched (only lifecycle teardown), so no frontend/backend mirroring is required, and the backend is untouched.

### What's good
Tight, correct fix with comments kept to the point and each teardown clear covered by a test. No blocking findings, no should-fixes, no nits.

---

PR #2074, opened 2026-07-16
Defer identity-column commits to blur in Workbench SectionTable (#2039)

### LINKED ISSUE #2039: Workbench SectionTable: per-keystroke identity-column commits corrupt sibling rows (lesson tour-step ordinals)
## Context

`NumInput` commits every fully-parseable intermediate value per keystroke (`UI/src/routes/admin/workbench/components/NumInput.svelte:60-67`), and `SectionTable.setCell` swaps row identities on collision (`SectionTable.svelte:136-148`):

```js
if (key === rowKey) {
    const collision = list.find((r, ri) =&gt; ri !== i &amp;&amp; r[rowKey] === value);
    if (collision) {
        collision[rowKey] = list[i][rowKey];
    }
}
```

The lesson tour-steps table is the one place with an author-editable identity column (`entities/lesson.ts:136` — `rowKey: 'ordinal'`).

**Failure scenario:** steps with ordinals `[0, 1, 2, 3]`; the author retypes step 3's ordinal as `12`. The first keystroke commits `1`, which collides with the row holding ordinal 1 — the swap silently gives that untouched row ordinal `3`, reordering it past step 2. Worse, the `{#each rows (row[rowKey])}` key swap re-assigns the existing `&lt;tr&gt;` (where focus lives) to the *other* row, so remaining keystrokes can land in a different logical row's input. Net: silently scrambled step order and/or the edit landing on the wrong row. The direct-collision swap is tested as intended (`SectionTable.test.ts:282`), but the per-keystroke interaction is not.

## Suggested fix

Commit identity-column (`key === rowKey`) edits on `change`/blur instead of per keystroke (a commit-on-blur mode on `NumInput`, used by `TableCell` for identity columns), or defer the collision swap until blur. Alternatively give steps a surrogate row id and demote `ordinal` to a plain column — the duplicate-ordinal warn `lesson.ts:121-127` already exists, so warn-on-duplicate could replace the swap entirely.

ginderjeremiah:
## Summary

`NumInput` commits every fully-parseable keystroke to the record, and `SectionTable.setCell` swaps two rows' identities when a `rowKey` edit collides with a sibling row's current value. Lesson tour steps are the one table where the row identity (`ordinal`) is also an author-editable number column, so retyping a multi-digit ordinal (e.g. `3` → `12`) committed the intermediate `1` first — colliding with the sibling row holding ordinal `1` and silently swapping its identity (and, via the keyed `{#each}`, re-assigning the `<tr>` under the cursor to that other row).

## Fix

- `NumInput.svelte`: added a `commitOnBlur` prop. When set, keystrokes only update the local display text; the value is committed once on blur, after the full edit is typed.
- `TableCell.svelte`: added an `identity` prop and forwards `commitOnBlur={identity}` to `NumInput` for `number`-type columns.
- `SectionTable.svelte`: passes `identity={col.key === rowKey}` to `TableCell`, so only the section's identity column (when it happens to be a `number` column, as with lesson tour steps) defers its commit — every other column is unaffected.

This means the identity/collision-swap logic in `setCell` now only ever runs against the final typed value, not an in-progress digit, closing the corruption window described in the issue.

## Testing

- Added `NumInput` unit tests for the new `commitOnBlur` mode (no per-keystroke `onChange`, single commit on blur, no commit on an unparseable blur value, default per-keystroke behavior unchanged).
- Added a `TableCell` test asserting `identity` forwards `commitOnBlur` to `NumInput`.
- Updated the existing `SectionTable` collision-swap test to commit via blur, and added a regression test that types a multi-digit ordinal one keystroke at a time and asserts no sibling row is touched until blur.
- `npm run lint` — clean.
- `npm run test:unit:ci` — 3751/3751 passing, no coverage regressions.
- `npm run build` — clean.

Closes #2039

---
_Generated by [Claude Code](https://claude.ai/code/session_01M95hbwvvtGUZGZsuWjNjW8)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/NumInput.svelte
+++ b/UI/src/routes/admin/workbench/components/NumInput.svelte
@@ -7,7 +7,7 @@
 	value={text}
 	oninput={handleInput}
 	onfocus={() => (focused = true)}
-	onblur={() => (focused = false)}
+	onblur={handleBlur}
 />
 
 <script lang="ts">
@@ -19,6 +19,10 @@ interface Props {
 	allowNegative?: boolean;
 	/** Accessible name for the input — the visible field labels are presentational `<span>`s, not associated `<label>`s. */
 	ariaLabel?: string;
+	/** Defer the commit to blur instead of every keystroke — for identity columns (e.g. a tour step's
+	 *  author-editable `ordinal`), where a per-keystroke commit can transiently collide with (and swap)
+	 *  a sibling row's identity mid-edit. */
+	commitOnBlur?: boolean;
 }
 
 let {
@@ -27,7 +31,8 @@ let {
 	class: className = '',
 	style = undefined,
 	allowNegative = false,
-	ariaLabel = undefined
+	ariaLabel = undefined,
+	commitOnBlur = false
 }: Props = $props();
 
 // Keep the in-progress text locally so typing "1." or "0.5" isn't clobbered by
@@ -50,6 +55,16 @@ $effect(() => {
 
 const pattern = $derived(allowNegative ? /^-?\d*\.?\d*$/ : /^\d*\.?\d*$/);
 
+// Only commit a fully-parseable number. An empty or in-progress field ("", "-", ".", "-.") leaves
+// the stored value untouched rather than coercing it to 0 — navigating away mid-edit then restores
+// the prior value (via the blur effect) instead of silently persisting 0.
+const commit = (raw: string) => {
+	const parsed = parseFloat(raw);
+	if (!Number.isNaN(parsed)) {
+		onChange(parsed);
+	}
+};
+
 const handleInput = (event: Event) => {
 	const el = event.currentTarget as HTMLInputElement;
 	const raw = el.value;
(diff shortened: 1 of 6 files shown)


ginderjeremiah (commented):
## Review: Defer identity-column commits to blur in Workbench SectionTable (#2039)

**Verdict:** Approve _(posted as a comment since I'm the PR author — GitHub won't accept a gating self-review)_ — the fix targets the actual root cause, is well-scoped, and is thoroughly tested.

### Analysis

The corruption in #2039 stemmed from two behaviors interacting: `NumInput` committing every parseable keystroke, and `setCell`'s collision-swap treating an intermediate digit as a real identity value. The fix cuts the interaction at the right seam — deferring the commit to blur for the identity column means the swap logic only ever sees the final typed value, so no transient collision/`{#each}`-key reassignment can occur mid-edit. The threading is minimal and correct: `commitOnBlur` is forwarded only to `number`-type identity cells (`TableCell.svelte:33`), leaving every other column's per-keystroke behavior untouched.

Refactoring the shared parse-and-commit logic into `commit()` (`NumInput.svelte:61-66`) is a nice DRY improvement over duplicating it across the input and blur paths.

### Test coverage

Solid and matches the project's testing rules:
- `NumInput` — no per-keystroke commit under `commitOnBlur`, single commit on blur, no commit on an unparseable blur value, and default per-keystroke behavior explicitly re-asserted.
- `TableCell` — forwards `commitOnBlur` in both directions (identity vs. non-identity).
- `SectionTable` — the existing collision-swap test updated to commit via blur, plus a direct regression for #2039 that types a multi-digit ordinal keystroke-by-keystroke and asserts no sibling row is touched until blur.

Not battle logic, so no cross-suite mirroring applies.

### Notes (non-blocking)

- Deferring identity-column commits to blur means an in-progress ordinal edit is dropped if the cell unmounts without firing blur (e.g. switching entities). That's the tradeoff the issue itself recommended, and losing an uncommitted edit is strictly better than corrupting sibling rows — flagging only for the record, not as a change request.
- A focus-then-blur with no edit will now re-commit the unchanged identity value; it's a harmless no-op through immer (`recordsEqual` stays true, no dirty flag), so no action needed.

The three-line explanatory comments are justified — they capture the footgun each layer is guarding against, which is genuinely non-obvious.

Nothing to change here. Merging once CI is green.

---

PR #2075, opened 2026-07-16
Fix CreateAccount 500 for 21-25 char usernames

### LINKED ISSUE #2040: CreateAccount accepts usernames longer than the DB column — 21–25 char usernames 500 instead of failing validation
## Context

The API model allows 25 characters (`Game.Api/Models/Auth/CreateAccountRequest.cs:12` — `[Length(1, 25)]`), but the column is 20 (`Game.Infrastructure/Database/GameContext.cs:829-830` — `HasMaxLength(20)`).

**Failure scenario:** `POST /api/Login/CreateAccount` with a 21–25 character username passes model validation and the availability check, then `Users.CreateAccount`'s self-commit throws a Postgres 22001 (string truncation). That `DbUpdateException` is not a unique violation, so the `IsUniqueViolation()` catch doesn't absorb it (`Game.DataAccess/Repositories/Users.cs:69-85`) — it escapes to `ExceptionHandlingMiddleware` as a 500 instead of a validation message. Unlike `Player.Name` (pinned to its column by the `PlayerName` domain rule), `Username` has no domain-side length rule at all. `LoginCredentials.cs:7` carries the same `[Length(1, 25)]` — harmless on login, but the same wrong constant.

## Suggested fix

Align both attributes to `[Length(1, 20)]`, ideally via a shared constant (mirroring how `PlayerName.MaxLength` anchors the player-name limit), and add the account-creation boundary test that would have caught this (`LoginControllerTests` has no length-edge case today).

ginderjeremiah:
## Summary

`CreateAccountRequest`/`LoginCredentials` allowed a 25-character username (`[Length(1, 25)]`), but the `User.Username` column is capped at 20 (`GameContext.cs` — `HasMaxLength(20)`). A 21–25 char username passed model validation and the availability check, then blew up as an unhandled Postgres string-truncation error at `Users.CreateAccount`'s self-commit — surfacing as a 500 instead of a clean validation message.

- Added `Game.Core.Identity.UsernamePolicy` (`MinLength`/`MaxLength = 20`), mirroring the existing `Game.Core.Players.PlayerName.MaxLength` pattern for the equivalent player-name rule.
- Pointed both `CreateAccountRequest.Username` and `LoginCredentials.Username`'s `[Length]` attributes at the shared constant instead of the wrong hardcoded `25`.
- Added boundary integration tests: a 20-char username still succeeds, a 21-char username is rejected with a structured 400 (`ValidationProblemDetails`), never a 500.

## Closes

Closes #2040

## Test plan

- [x] `dotnet build` — solution builds clean
- [x] `dotnet test` — full backend suite passes (3228 tests: Core, Infrastructure, Application, Api)
- [x] New tests specifically cover the 20-char boundary (succeeds) and 21-char boundary (400, not 500)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wqi8brZou2xeQkKvKm5oZ9)_

### DIFF
--- a/Game.Api.Tests/Integration/LoginControllerTests.cs
+++ b/Game.Api.Tests/Integration/LoginControllerTests.cs
@@ -5,10 +5,12 @@ using Game.Api.Http;
 using Game.Api.Models.Auth;
 using Game.Api.Models.Common;
 using Game.Core;
+using Game.Core.Identity;
 using Game.Infrastructure.Database;
 using Game.Infrastructure.Entities;
 using Game.TestInfrastructure.Fixtures;
 using Game.TestInfrastructure.Helpers;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using System.Net;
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Fix CreateAccount 500 for 21–25 char usernames (closes #2040)

**Verdict:** Approve — the change correctly closes the API/DB length mismatch that caused the 500, at the right layer, with boundary tests that would have caught the original bug.

_(Submitted as a `COMMENT` for mechanical reasons — GitHub won't accept an `APPROVE` from the PR author's own login — but this is a clean approval.)_

I verified the root cause and the fix against the codebase:
- `User.Username` really is capped at `HasMaxLength(20)` (`GameContext.cs:829-830`), so aligning the attribute to `20` is the correct value, and the previous `[Length(1, 25)]` was the defect.
- The shared `UsernamePolicy` constant is placed in `Game.Core.Identity` (domain layer) and mirrors the existing `Game.Core.Players.PlayerName.MaxLength` pattern — consistent with the project's DDD layering and with how the issue framed the fix.
- Both API models (`CreateAccountRequest`, `LoginCredentials`) now route through the constant; a repo-wide grep confirms no stray hardcoded `25`/`20` username lengths remain.

### Test coverage
- Good: the new integration tests cover both boundaries — 20 chars succeeds, 21 chars returns a structured `ValidationProblemDetails` 400 (not a 500) — which is exactly the regression the issue asked for. Asserting on `ValidationProblemDetails` + the `Username` error key (rather than just the status code) makes the test resilient and precise.
- The `LoginCredentials` side has no dedicated test, but that change is a trivial constant swap the issue itself flagged as harmless on the login path, so a separate test isn't warranted.

- Unlike `PlayerName`, `UsernamePolicy` is purely length bounds with no `TryNormalize` (no trimming / control-char rejection). That's appropriate here — username length is enforced by the attribute + column and there's no stated normalization requirement — so this is an observation, not a gap.

Nicely scoped, single-purpose fix. Holding the merge only until the remaining CI jobs (`test-backend`, e2e webkit/firefox) finish green.

---

PR #2076, opened 2026-07-16
Close the switch-away credit race with an atomic presence-key claim (#2041)

### LINKED ISSUE #2041: SwitchPlayer's departed-socket guard is check-then-act — the switch-away credit can still race a live battle loop
## Context

Found in the July 2026 health review (backend service-layer pass).

`CreditDepartedCharacter`'s guard (`Game.Api/Controllers/LoginController.cs:205-222`) is a single presence-key read (`SocketManagerService.HasActiveSocket`) followed by a multi-step credit (`EnsureSessionLoaded` → `LoadPlayer` → `SimulateSwitchProgress`), with no atomicity between them. The inline comment claims a misbehaving client "cannot race the credit against the live loop's saves", but two windows remain:

1. A malicious client calls `SwitchPlayer` and opens a socket for the departed player just after `HasActiveSocket` returns false. The offline-rewards credit (an HTTP-thread read-modify-write of the same aggregate) then runs concurrently with the socket's per-command battle saves — the exact lost-update class the per-socket lock exists to prevent, plus potential double-credit of elapsed idle time.
2. Honest-client variant: the presence key's 30s TTL (`SocketManagerService.cs:27`) lapses during a network stall while the socket is actually still alive; the check passes, the credit runs, the socket resumes mid-credit.

Consequences are bounded (both sides' write-behind events still apply durably — it's cache-blob clobber plus possible duplicate idle credit, not durable data loss), but the window is real.

## Suggested fix

Make the guard claim-based instead of read-based: atomically claim the departed player's presence key (e.g. `CompareAndSet(absent → "switch-credit", shortTtl)`) so a concurrent `RegisterSocket` either sees the claim and defers/kicks it, or wins and suppresses the credit — mirroring how `RegisterSocket` itself claims the key with `GetSet`. At minimum, soften the "cannot race" comment to "narrows the race".

Related: #1513/#1518 cover the takeover-side races; this is the switch-credit side.

ginderjeremiah:
## Summary

`SwitchPlayer`'s departed-character credit (`LoginController.CreditDepartedCharacter`) used a check-then-act guard: a single `HasActiveSocket` presence read, followed by a multi-step read-modify-write (`EnsureSessionLoaded` → `LoadPlayer` → `SimulateSwitchProgress`) with no atomicity between them. A socket could register for the departed player in the gap between the check and the credit, letting the HTTP-thread credit race the incoming battle loop's saves against the same cached aggregate — the exact lost-update class the per-socket command lock exists to prevent.

This closes that window with an atomic claim instead of a plain read:

- `SocketManagerService.TryClaimForSwitchCredit` atomically claims the player's presence key (`CompareAndSet` against "unset") in place of the old `HasActiveSocket` read. If the key is already held — a genuinely live socket, or another switch-credit already in flight — the credit is skipped exactly as before.
- `RegisterSocket` now defers behind an in-flight switch-credit claim (a bounded poll-wait) instead of blindly overwriting it, so a socket that would previously have raced the credit's writes now waits for the credit to finish and release its claim first. The wait is capped at the claim's own short TTL, so a crashed credit can never wedge a connection.
- `SocketManagerService.ReleaseSwitchCreditClaim` releases the claim via compare-and-delete (only if it's still ours) once the credit completes, in a `try/finally` so a faulted credit still releases promptly.

Updated the relevant design docs (`docs/backend-auth.md`, `docs/backend-sockets.md`) to describe the new claim-based mechanism in place of the old plain-read description.

## Test plan

- [x] Unit/integration coverage added in `SocketManagerServiceTests`: claim succeeds when unset, fails when a real socket is already registered, release only clears the key while still owned (compare-and-delete semantics), and `RegisterSocket` observably defers registration until an active claim is released.
- [x] Integration coverage added in `LoginControllerTests`: a successful `SwitchPlayer` credit releases its claim afterward (no lingering block on a later reconnect).
- [x] Existing `SwitchPlayer` coverage (credits-and-binds, skips-when-live-socket) still passes unchanged.
- [x] Full backend test suite (`dotnet test`, all 4 projects) — 3,234 tests, 0 failures.

Closes #2041

---
_Generated by [Claude Code](https://claude.ai/code/session_019jA1ymjDhpJa486HKH4EYA)_


Game.Api/Services/SocketManagerService.cs
```diff
@@ -39,6 +39,24 @@ public class SocketManagerService
         /// </summary>
         private static readonly TimeSpan SocketQueueTtl = TimeSpan.FromHours(1);
 
+        /// <summary>
+        /// Sentinel value <see cref="TryClaimForSwitchCredit"/> writes into a presence key: not a real socket
+        /// id, so <see cref="RegisterSocket"/> must recognize it (rather than notifying it as a replaced
+        /// connection) and <see cref="CurrentSocketId"/>'s callers must never treat it as a live socket.
```
ginderjeremiah, line 45: **Should-fix:** This comment asserts the contract *"`CurrentSocketId`'s callers must never treat it as a live socket"* — but `EmitSocketCommand(int playerId)` (line 266) does exactly that. It resolves `CurrentSocketId`, and any non-null value (including this sentinel) is treated as a deliverable socket: it publishes to `socket_switch-credit` / `queue_switch-credit` (which no socket drains) and returns `true`.

The consequential caller is the dead-letter replay (`SocketCommandDeadLetters.cs:76`), which gates removal of the dead-lettered entry on that `bool`. If an ops replay targets a player during their switch-credit window, `EmitSocketCommand` reports success, the entry is `RemoveAsync`'d, and the command is silently lost. The `ProficiencyXpNotifier`/`ChallengeCompletedNotifier` fire-and-forget pushes are lower-stakes (the departed player has no socket anyway) but likewise get swallowed into a dead queue with a 1h TTL instead of the honest "no active socket" warning — and `SimulateSwitchProgress` itself can raise proficiency events *while* the claim is held, so this path is reachable in-process, not just via the ops surface.

The window is tiny, so this isn't a merge blocker, but it's a real gap this PR introduces and the fix is small: have `EmitSocketCommand(int playerId)` treat a resolved sentinel as "no live socket" (return `false`/log), so the delivery-gating contract the dead-letter replay relies on stays honest. `HasActiveSocket`'s intentional occupied-during-claim behavior can stay as-is. Worth locking in with a small test on `EmitSocketCommand(int)` during an active claim.

---

ginderjeremiah (commented):
## Review: atomic presence-key claim to close the SwitchPlayer switch-away credit race (#2041)

**Verdict:** Approve (posted as a comment since I'm the PR author) — with one should-fix. The core change is correct and well-tested; the should-fix is a narrow, pre-existing-contract violation this PR introduces and can be handled here or as a fast follow-up, but it shouldn't gate merge on its own.

The change does what the issue asked: it replaces the check-then-act `HasActiveSocket` read with an atomic `CompareAndSet(absent → "switch-credit")` claim, defers a concurrent `RegisterSocket` behind an active claim (bounded by the claim TTL), and releases via compare-and-delete in a `try/finally`. I verified the cache primitives back it correctly — `CompareAndSet(key, null, …)` asserts absence server-side (`exists == 1 → return 0`), `CompareAndDelete` is value-guarded, and `GetSet` returns the prior value atomically with its TTL — so a socket registering concurrently either loses to the claim (credit skipped) or waits it out, closing the read→credit gap. The `finally` release is correctly placed (no code between the successful claim and the `try`, so the claim is never held unprotected), and it release-on-fault so a faulted credit doesn't wedge a reconnect for the full TTL.

### Should-fix

- **Sentinel leaks through `CurrentSocketId` into `EmitSocketCommand(int playerId)`** — `SocketManagerService.cs:45` asserts callers must never treat the claim sentinel as a live socket, but the `EmitSocketCommand(int)` overload does, so a server push during the claim window (most consequentially the dead-letter replay at `SocketCommandDeadLetters.cs:76`, which gates entry removal on the returned `bool`) reports a false success and can drop the command. Details and suggested fix in the inline comment.

### Nits / notes

- **Residual hairline race, acceptable.** A concurrent switch-credit that re-claims the key in the instant between `AwaitSwitchCreditClaimClear` returning and the following `GetSet` gets kicked by that `GetSet`, letting the new socket's loop start alongside that credit's read-modify-write. This is dramatically narrower than the window the PR closes (sub-ms vs. the full multi-step credit) and requires two overlapping switch-credits for the same departed player; fully closing it would need a heavier register-side claim protocol. The PR's own comment at line 88 already acknowledges the "claimed in the gap" case. Fine to leave as-is; noting it so a later reader doesn't mistake it for fully serialized.
- The 20s `SwitchCreditClaimTtl` means a crashed credit can stall a legitimate reconnect for up to that long (`AwaitSwitchCreditClaimClear` caps at the TTL). Rare and documented; acceptable.

### Test coverage

Good and appropriately integration-level (the logic is a Redis concurrency protocol, not pure domain — consistent with `backend.md`'s "integration tests for out-of-process dependencies"). New tests cover claim-when-unset, claim-fails-on-live-socket, release-only-while-owned (both the still-owned clear and the taken-over no-op), `RegisterSocket` deferring until release, and an integration check that a real `SwitchPlayer` credit releases its claim afterward. The one gap worth adding is a test pinning `EmitSocketCommand(int)`'s behavior during an active claim (ties to the should-fix above). Backend-only change with no battle-logic surface, so no frontend mirror is required.

### What's good

- The atomic-claim design mirrors the existing `GetSet`/`CompareAndDelete` presence-key patterns rather than inventing a new mechanism, and the `try/finally` + compare-and-delete release is exactly the right shape for "release only if still ours."
- Thorough, accurate doc comments and a clean docs update (`backend-auth.md` / `backend-sockets.md`) that describe the new mechanism rather than narrating the change.
- The `oldSocketId == SwitchCreditClaimValue → null` handling in `RegisterSocket` correctly avoids notifying a claim as a replaced connection.

---

ginderjeremiah:
## Re-review (follow-up commit `849f212`)

The prior review's one should-fix — the switch-credit sentinel leaking through `EmitSocketCommand(int)` and reporting a false delivery to the dead-letter replay — has been addressed in `849f212`. Verified the fix and its blast radius:

- `EmitSocketCommand(int playerId)` now guards `socketId is not null && socketId != SwitchCreditClaimValue`, so a resolved claim reports exactly like "no active socket" — the dead-letter replay (`SocketCommandDeadLetters.cs:76`) correctly *retains* the entry for retry instead of dropping it, and the fire-and-forget notifiers (`ChallengeCompletedNotifier`, `ProficiencyXpNotifier`) simply no-op during the window.
- I traced every reader of the presence-key value: the key is confined to `SocketManagerService` (nothing outside it reads `CACHE_PLAYER_SOCKET_PREFIX`), and the only two value-readers are `HasActiveSocket` (documented harmless false-positive, warning-only) and `EmitSocketCommand(int)` (now fixed). The sentinel is fully contained.
- New test `EmitSocketCommand_ByPlayerId_SwitchCreditClaimActive_ReportsNoActiveSocketRatherThanPublishingToTheClaim` pins the behavior at the correct integration level.

The residual items from the first review (the sub-millisecond re-claim hairline race, and the 20s TTL upper bound on a crashed-credit reconnect stall) remain documented, acceptable trade-offs — not follow-up work.

CI is green (backend, e2e, verify all passing). No remaining nits or should-fixes. Merging.

---

PR #2077, opened 2026-07-16
Close the ContentSeeder concurrent-boot double-seed race with an advisory lock (#2042)

### LINKED ISSUE #2042: ContentSeeder's populated-DB check is a cross-instance check-then-act — concurrent fresh boots can double-seed and fail startup
## Context

Found in the July 2026 health review (Game.DataAccess pass).

The `AnyAsync` fresh-DB guard (`Game.DataAccess/Content/ContentSeeder.cs:26-30`) runs before (and outside) the seed transaction, which is default READ COMMITTED — so two instances booting concurrently against a fresh database both pass the guard and both bulk-insert with explicit ids.

**Failure scenario:** a multi-instance deploy (or scaled CI stack) with `MigrateOnStartup` + `SeedContentOnStartup` enabled boots two API pods simultaneously; the loser fails startup with an opaque PK unique-violation instead of a clean "already seeded" no-op. No data corruption (the transaction rolls back atomically), and the flag is documented as dev/CI/recovery-scoped — but the multi-instance design makes concurrent boot a normal event, and EF's migration lock protects `MigrateAsync` while this step is unprotected.

## Suggested fix

Take a Postgres advisory lock for the seed step (`SELECT pg_advisory_xact_lock(&lt;const&gt;)` inside the transaction, re-checking `AnyAsync` under it), or catch the unique violation and return `false` as "another instance seeded first". `ContentSeederTests` covers idempotency but not concurrency.

ginderjeremiah:
## Summary

`ContentSeeder.SeedAsync`'s fresh-DB guard (`AnyAsync` on the Skills set, `Game.DataAccess/Content/ContentSeeder.cs:26-30`) ran before — and outside — the seed transaction, which defaults to READ COMMITTED. Two instances booting concurrently against a fresh database could both pass the guard and both bulk-insert the whole content graph with explicit ids, so the loser crashed on an opaque PK unique-violation instead of a clean "already seeded" no-op. No data corruption resulted (the transaction rolled back atomically), but a multi-instance deploy or a scaled CI stack makes concurrent boot a normal event, and EF's migration lock protects `MigrateAsync` while this step was unprotected.

This closes the window with an advisory lock instead of relying on the plain outside-transaction read:

- `SeedAsync` now takes a session-scoped Postgres advisory lock (`SELECT pg_advisory_xact_lock(...)`) as the first statement inside its transaction and **rechecks** the presence guard under it. The lock blocks until any other in-flight seed transaction commits or rolls back, so the recheck always observes that transaction's result — at most one instance ever inserts, and a loser skips cleanly.
- The original outside-transaction check stays in place as a cheap fast-path skip for the common already-seeded case (no need to open a transaction/take the lock on every boot once a database is populated).
- This mirrors the existing row-lock check-then-act pattern in `Users.cs` (`SELECT ... FOR UPDATE`) — there's just no natural row to lock here, since the check is "does any Skill exist" rather than a specific entity, hence the advisory lock.
- Updated `docs/backend-content.md` to describe the seeder as concurrency-safe, not just idempotent.

## Test plan

- [x] New concurrency test in `ContentSeederTests`: fires two `SeedAsync` calls concurrently (separate DI scopes / `GameContext` instances) against a real Postgres container and asserts exactly one returns `true`, the other returns `false`, and no exception is thrown — with the final row count matching a single seed.
- [x] Existing `ContentSeederTests` coverage (fresh-DB seed, explicit id preservation, already-populated no-op, tagged item-mod join) still passes unchanged.
- [x] Full backend test suite (`dotnet test`, all 4 projects) — 3,229 tests, 0 failures.

Closes #2042

---
_Generated by [Claude Code](https://claude.ai/code/session_01XUjEa7mRZmfCJjJjjq2Rso)_


Game.DataAccess/Content/ContentSeeder.cs
```diff
@@ -11,6 +11,10 @@ namespace Game.DataAccess.Content
     /// <inheritdoc cref="IContentSeeder"/>
     internal sealed class ContentSeeder : IContentSeeder
     {
+        // Arbitrary key for the seed step's session-scoped Postgres advisory lock (the app's only
+        // advisory-lock user, so no collision risk with any other key space).
```
ginderjeremiah: Nit: the parenthetical "the app's only advisory-lock user" isn't quite accurate. EF Core's Npgsql provider takes its own Postgres advisory lock (`pg_advisory_xact_lock`) for the migration-history lock during `MigrateAsync`. There's still no real collision risk here — it uses a different (hash-derived) key, and migration fully completes and releases its lock before seeding runs — but the "only advisory-lock user" premise is false. Consider softening to something like "the only advisory lock we take directly; distinct from EF's migration-history lock key" so a future dev doesn't build on the wrong assumption.

---

ginderjeremiah (commented):
## Review: Close the ContentSeeder concurrent-boot double-seed race (#2042)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the fix is correct, minimal, well-tested, and precisely matches the issue's suggested approach. One trivial comment-accuracy nit, no blockers.

The change closes the check-then-act window exactly as intended: an xact-scoped `pg_advisory_xact_lock` taken as the first statement inside the seed transaction, followed by a recheck of the presence guard under the lock. Because the lock blocks until any in-flight seed transaction commits/rolls back and the recheck reads a fresh READ COMMITTED snapshot afterward, at most one instance inserts and the loser cleanly returns `false`. Keeping the original outside-transaction `AnyAsync` as a fast-path skip is the right call — it avoids opening a transaction and taking the lock on every boot once the DB is populated. The xact-scoped lock also means no manual release/leak risk, and the early `return false` under the lock correctly rolls back and releases via `await using`. Choosing the advisory lock over catch-the-unique-violation is the cleaner of the two options the issue proposed. Layering is fine — this is DB orchestration in the data-access seeder, no domain logic involved.

### Nits

- `ContentSeeder.cs:14-15` — the comment calls this "the app's only advisory-lock user." EF Core's Npgsql provider also takes a `pg_advisory_xact_lock` for its migration-history lock during `MigrateAsync`. No actual collision risk (distinct key, and migration completes before seeding runs), but the premise is inaccurate — worth softening so a future dev doesn't rely on it. Inline comment left.

### Test coverage

Good. `SeedAsync_ConcurrentFreshBoots_ExactlyOneSeedsAndTheLoserNoOpsCleanly` fires two `SeedAsync` calls from separate DI scopes / `GameContext` instances against the real Postgres container and asserts exactly one `true`, one `false`, no exception, and a final row count matching a single seed — which is precisely the concurrency gap the issue called out as untested. Existing idempotency/id-preservation/join coverage is retained. This is an integration test exercising an out-of-process dependency (Postgres), which is the correct level per the backend testing guidelines.

### What's good

- Tightly scoped fix that mirrors the existing `Users.cs` row-lock check-then-act pattern, with a clear rationale comment for why an advisory lock (no natural row to lock) rather than a row lock.
- `docs/backend-content.md` updated to describe the seeder as concurrency-safe, not just idempotent — the design-decision doc stays in sync with the behavior change.

Note: CI ("Run tests") was still in progress at review time, so I'm not merging on this pass. Given the outstanding nit is comment-only, it's a judgment call whether to address it before or after merge — either is fine.

---

ginderjeremiah:
## Re-review (follow-up commits `7e019c2`, `6d590ba`)

The prior review's comment-accuracy nit ("the app's only advisory-lock user") was addressed in `7e019c2`, which now correctly acknowledges EF's own migration-history `pg_advisory_xact_lock`.

While re-reading the final state I caught one residual inaccuracy: the lock-key comment described the lock as **"session-scoped,"** but the code uses `pg_advisory_xact_lock` — a *transaction*-scoped lock, exactly as the call-site comment states ("Xact-scoped, so it releases automatically on commit/rollback"). Left unfixed, a future reader could wrongly assume the lock persists across the (pooled) connection's session and reason about its release semantics incorrectly. Since it's a one-word comment correction that clearly belongs in this PR rather than a follow-up, I pushed `6d590ba` to align the wording ("session-scoped" → "transaction-scoped").

The core change remains correct: the `pg_advisory_xact_lock` taken as the first statement inside the seed transaction, followed by a recheck of the presence guard under it, closes the concurrent-boot double-seed window exactly as issue #2042 proposed; the outside-transaction `AnyAsync` fast-path is retained; and the concurrency test asserts exactly-one-seeds against a real Postgres container. Docs updated to describe the seeder as concurrency-safe.

CI is green on `6d590ba` (backend, e2e ×3, verify all passing). No remaining nits or should-fixes. Merging.

---

PR #2078, opened 2026-07-16
Skip SetMax's information-free first-write 0 row

### LINKED ISSUE #2043: SetMax first-write persists an information-free 0 row, violating the row-presence convention
## Context

Found in the July 2026 health review (Game.Core pass).

`PlayerProgress.SetMax` (`Game.Core/Progress/PlayerProgress.cs:354-365`) records even a `0` on first write (`if (created || value &gt; stat.Value)`), while `Increment` deliberately skips a `0` delta with the "no information-free rows" rationale (`:340-352`), and the persistence docs define "no data yet" as the absence of the row.

**Failure scenario:** any completed battle where the player never landed a hit — an early abandon replay, or a quick loss to an enemy with a shorter first-fire cooldown — calls `SetMax(HighestSingleAttackDamage, null, 0)` (producer at `:58`, runs on every completed battle including draws/losses), creating and dirty-marking a `0` row. Cost: a junk row in the write-behind/cache snapshot, a touched key that pointlessly scopes challenge evaluation to `HighestSingleAttackDamage` challenges, and the Statistics screen showing a recorded "highest attack: 0" instead of no data.

`SetMin`'s first-write-wins is genuinely needed (a legit-0 minimum like an instant `FastestVictory` must record); `SetMax` has no such case — the first *positive* value would win against a fresh row's 0 anyway.

## Suggested fix

In `SetMax`, skip the write when `created &amp;&amp; value == 0` (mirroring `Increment`'s zero-skip); `SetMin` stays as-is. One unit test.

ginderjeremiah:
## Summary

`PlayerProgress.SetMax` (`Game.Core/Progress/PlayerProgress.cs`) recorded even a `0` on first write, unlike `Increment`'s deliberate zero-delta skip. Every completed battle where the player never landed a hit (an early abandon, a quick loss) called `SetMax(HighestSingleAttackDamage, ..., 0)`, creating and dirty-marking a junk row — a stored "highest attack: 0" instead of the "no data yet" the persistence docs define as row absence, plus a touched key that pointlessly scoped challenge evaluation to `HighestSingleAttackDamage` challenges.

`SetMin`'s first-write-wins is untouched: a legitimate `0` minimum (e.g. an instant `FastestVictory`) genuinely needs to lock in. `SetMax` has no equivalent case — the first *positive* value would beat a fresh row's `0` anyway — so a first-write `0` now carries no information and is skipped, mirroring `Increment`'s zero-delta skip.

`HighestSingleAttackDamage` is the only `Max`-aggregated statistic today (pinned by `StatisticTypeTests.HighestSingleAttackDamage_IsTheOnlyMaxAggregatedStatistic`), so this only affects that stat, both globally and per-skill.

## Changes

- `Game.Core/Progress/PlayerProgress.cs`: `SetMax` now skips the write (no row creation, no dirty-mark) when the value is `0` and no row exists yet for that key.
- `Game.Core.Tests/Progress/PlayerProgressTests.cs`: split the old combined Min/Max zero-first-write test into two, pinning the new asymmetry; updated the zero-damage-skill test's `HighestSingleAttackDamage` expectation to match.
- `docs/backend-persistence.md`: updated the row-presence paragraph to describe the Min/Max asymmetry instead of claiming both key off first-write `0`.

## Test plan

- [x] New/updated unit tests in `PlayerProgressTests` cover: a first-write `0` `FastestVictory` (Min) still creates/dirties its row; a first-write `0` `HighestSingleAttackDamage` (Max) stays absent and untouched; the zero-damage-skill scenario no longer expects a per-skill `HighestSingleAttackDamage` row.
- [x] Full backend test suite (`dotnet test`, all 4 projects) — 3,229 tests, 0 failures.

Closes #2043

---
_Generated by [Claude Code](https://claude.ai/code/session_01W11gAR71xntkd514DecXB6)_


ginderjeremiah (commented):
## Review: skip `SetMax`'s information-free first-write `0` row (#2043)

**Verdict:** Approve — correctly closes the row-presence-convention violation with a minimal, well-reasoned change; no blocking or should-fix findings. (Submitted as a comment because I'm the PR author; the intent is an approval.)

The fix does exactly what #2043 asked and does it cleanly. A first-write `0` on a `Max` stat now skips the write entirely — no row, no dirty-mark, no touched key — mirroring `Increment`'s zero-delta skip, while `SetMin`'s legitimate-`0` first-write (e.g. an instant `FastestVictory`) is untouched.

One thing worth noting as a positive: the guard is placed *before* `GetOrCreate` (`value == 0m && !_statistics.ContainsKey(...)`) rather than the issue's literal `created && value == 0` suggestion. That's the better call — checking `created` after `GetOrCreate` would have created and then had to discard a phantom row. `!ContainsKey` is equivalent to `created` here, so the behavior matches the issue while avoiding the transient row. The surviving `created || value > stat.Value` branch is still correct, since after the guard a `created` write always carries a positive value.

Edge cases check out:
- `0`, no row → skipped (the fix).
- `0`, existing positive row → `created` false, `0 > positive` false → no-op.
- first positive → recorded via the `created` branch.

Since `HighestSingleAttackDamage` is the only `Max`-aggregated statistic (pinned by `StatisticTypeTests.HighestSingleAttackDamage_IsTheOnlyMaxAggregatedStatistic`), the blast radius is exactly that stat, globally and per-skill, as described.

### Test coverage
Solid. The old combined Min/Max test is split into two that pin the intended asymmetry — Min still creates/dirties its `0` row, Max stays absent and untouched — and the zero-damage-skill test is updated to expect the per-skill `HighestSingleAttackDamage` row to stay absent. This is backend-only statistics-recording logic (not battle-simulation math the frontend mirrors), so no frontend test mirror is required.

### Docs
`docs/backend-persistence.md` is updated to describe the Min/Max asymmetry accurately instead of implying both key off a first-write `0`.

### What's good
Minimal, side-effect-clean fix; the code comments capture the *why* (the asymmetry with `SetMin`) rather than restating the code; docs and tests kept in lockstep with the behavior change.

Will merge once CI is green.

---

PR #2079, opened 2026-07-16
Track fight-screen skill tooltip by skill id, not slot index (#2044)

### LINKED ISSUE #2044: Fight-screen skill tooltip stays open / silently swaps content when the hovered slot is replaced mid-battle
## Context

Found in the July 2026 health review (game screens pass).

The fight-screen skill tooltip tracks a slot **index**, not a skill identity (`UI/src/routes/game/screens/fight/Skills.svelte:73-75` — `tooltipSkillIndex` + `$derived(battler.skills[tooltipSkillIndex])`), and is only hidden by `mouseleave`/`blur`. An enemy battler's skills array is reassigned on every spawn (`lib/battle/battler.ts:418`), and the keyed `{#each}` destroys/recreates the `&lt;button&gt;` when the id at a slot changes — a removed DOM node never fires `mouseleave`.

**Failure scenario:** the player parks the cursor over an enemy skill slot; the enemy dies and the next one spawns (continuous in the idle farm). The old button is destroyed without a `mouseleave`, `tooltipSkillIndex` stays set, and the tooltip remains open — re-derived to the *new* enemy's skill at that index, anchored at the stale cursor position, with nothing hovered. (If the new loadout has no skill at that index it self-heals via `TooltipShell hidden={!skill}`.)

The codebase already treats this as a bug class: `ActiveEffectChips.svelte:82-84,141-145` added an explicit guard for exactly this ("a chip removed under the cursor never fires `mouseleave`") — the skill slots never got the same treatment.

## Suggested fix

Mirror the chips' pattern — track the hovered skill's id (not index) and add an `$effect` that calls `hideTooltip()` when that id no longer resolves at the hovered slot.

ginderjeremiah:
## Summary

`Skills.svelte`'s per-skill combat tooltip tracked the hovered slot by **index** (`tooltipSkillIndex` + `battler.skills[tooltipSkillIndex]`). A battler's `skills` array is reassigned wholesale on every spawn (`lib/battle/battler.ts:418`), and the keyed `{#each battler.skills as skill, index (skill?.id ?? -index - 1)}` destroys/recreates a slot's `<button>` when the id at that index changes — a destroyed DOM node never fires `mouseleave`.

**Failure scenario:** the player parks the cursor over an enemy skill slot; the enemy dies and the next one spawns (continuous in the idle farm). The old button is destroyed without a `mouseleave`, `tooltipSkillIndex` stays set, and the tooltip re-derives to the *new* enemy's skill at that index — anchored at the stale cursor position, showing wrong content for nothing hovered.

The codebase already treats this as a bug class: `ActiveEffectChips.svelte` added an explicit guard for exactly this ("a chip removed under the cursor never fires `mouseleave`"), but the skill slots never got the same treatment.

## Fix

- Track the hovered skill's **id** instead of its slot index (`tooltipSkillId`), and resolve `tooltipSkill` by searching `battler.skills` for that id.
- `handleEnter`/`handleFocus`/`handleLeave` now receive/compare the skill id rather than the index.
- Added a self-heal `$effect` (mirroring `ActiveEffectChips`) that closes the tooltip if the hovered id no longer resolves to a skill in the current loadout — covers the case where a respawn's new loadout has no skill at all at the old slot.

## Test plan

- [x] New regression test in `Skills.test.ts`: hovers a slot, reassigns `battler.skills` wholesale with a different id at the same index (simulating a respawn) on a `statify`-wrapped battler, and asserts the tooltip closes instead of silently swapping to the new occupant's data.
- [x] Existing `Skills.test.ts` coverage (hover/leave, focus/blur, ready state, effect badges, aria-describedby) still passes unchanged.
- [x] Full frontend unit suite (`vitest --run`) — 299 files, 3752 tests, 0 failures.

Closes #2044

---
_Generated by [Claude Code](https://claude.ai/code/session_01CQhkZjpLhsDvgMnCdqGk2T)_

### DIFF
--- a/UI/src/routes/game/screens/fight/Skills.svelte
+++ b/UI/src/routes/game/screens/fight/Skills.svelte
@@ -18,10 +18,10 @@
 					class:ready={charge.ready}
 					style:--pulse-color={pulseColor}
 					onmousemove={handleMouseMove}
-					onmouseenter={(ev) => handleEnter(ev, index)}
-					onmouseleave={() => handleLeave(index)}
-					onfocus={(ev) => handleFocus(ev, index)}
-					onblur={() => handleLeave(index)}
+					onmouseenter={(ev) => handleEnter(ev, skill.id)}
+					onmouseleave={() => handleLeave(skill.id)}
+					onfocus={(ev) => handleFocus(ev, skill.id)}
+					onblur={() => handleLeave(skill.id)}
 					use:describedByTooltip={describedById}
 				>
 					<img class="skill-icon" src={skill.iconPath} alt={skill.name} />
@@ -70,9 +70,12 @@ type Props = {
 const { battler, side, accent }: Props = $props();
 
 let tooltip = $state<TooltipComponent>();
-let tooltipSkillIndex = $state(-1);
+// Tracked by skill id, not slot index: a battler's skills array is reassigned on every spawn, and the
+// keyed {#each} above destroys/recreates a slot's button when the id at that index changes, so a stale
+// index would silently re-derive to the new occupant instead of closing (mirrors ActiveEffectChips).
+let tooltipSkillId = $state<number>();
 
-const tooltipSkill = $derived(battler.skills[tooltipSkillIndex]);
+const tooltipSkill = $derived(battler.skills.find((skill) => skill?.id === tooltipSkillId));
 /** The accent the "ready" border/glow/label resolve to (CSS var, theme-overridable). */
 const readyAccent = $derived(accent ?? 'var(--accent)');
 const pulseColor = $derived(tintColor(accent ?? (side === 'player' ? 'var(--accent)' : 'var(--enemy-accent)'), 0.6));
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Track fight-screen skill tooltip by skill id, not slot index (#2044)

**Verdict:** Approve — the fix correctly re-identifies the hovered skill by id, closing the mid-battle stale-tooltip window, and faithfully mirrors the existing `ActiveEffectChips` pattern. (Submitted as `COMMENT` only because the review author is the PR author; treat it as an approval.)

I verified the change against the linked issue and the `ActiveEffectChips.svelte` precedent it cites:

- **Root cause addressed** — `tooltipSkillId` + `battler.skills.find(s => s?.id === tooltipSkillId)` replaces the index-based derive, so a wholesale `skills` reassignment on respawn no longer re-anchors the tooltip to whatever skill lands at the old slot. This matches the issue's suggested fix exactly.
- **Self-heal `$effect`** — mirrors chips' guard (`tooltipSkillId != null && !tooltipSkill → hide`). The `!= null` correctly covers both `undefined` and any null slot; and since handlers only ever assign a real `skill.id`, the state stays well-defined. No reactive loop (the reset flips the guard false on rerun).
- **Uniqueness assumption is safe** — the keyed `{#each ... (skill?.id ?? -index - 1)}` already requires ids unique per loadout, so resolving by `find` on id is sound.

### Test coverage
- New regression test drives the exact failure mode: hover a slot on a `statify`-wrapped battler, reassign `skills` wholesale with a different id at the same index (simulating a respawn), and assert the tooltip closes rather than swapping to the new occupant. This is the right shape — a plain non-reactive battler wouldn't propagate the reassignment.
- Existing hover/leave, focus/blur, ready-state, effect-badge, and empty-slot coverage is preserved unchanged.
- Pure display-side frontend change; no battle-math parity concern, so no backend mirror needed.

### What's good
Minimal, pattern-consistent fix with a clear inline comment explaining *why* id-tracking is required, and a regression test that would actually catch a reversion. Nothing to change.


---

PR #2080, opened 2026-07-16
Guard CharacterSwitcher's reload effect against a stale-response race

### LINKED ISSUE #2045: CharacterSwitcher reload effect has no stale-response guard — a cancelled open's fetch can clobber the fresh one
## Context

Found in the July 2026 health review (game screens pass).

`UI/src/routes/game/CharacterSwitcher.svelte:140-144`:

```js
$effect(() =&gt; { if (open) { void loadCharacters(); } });
```

`loadCharacters` (lines 118–137) unconditionally writes `view`/`phase`/`loadError` when its fetch resolves.

**Failure scenario:** open the switcher (fetch A starts), cancel, reopen (fetch B starts). Both are in flight; whichever resolves *last* wins. If stale A resolves after B, it replaces the freshly-built `PlayerSelectView` — resetting any in-progress create-form state — and on a failure path could flip a working list into the error state. The codebase already uses a generation counter for exactly this pattern (`SkillsView.commitSeq`, the enemy-fetch generation in frontend-screens.md).

## Suggested fix

Capture a request generation (or `AbortController`) per invocation and ignore resolutions that aren't the latest; optionally also skip applying results while `open` is false.

ginderjeremiah:
## Summary

`CharacterSwitcher.svelte`'s reload effect (`$effect(() => { if (open) { void loadCharacters(); } })`) fires `loadCharacters()` on every `open` transition with no guard against overlapping requests. `loadCharacters` unconditionally writes `view`/`phase`/`loadError` when its `Login/Players` fetch resolves.

**Failure scenario:** open the switcher (fetch A starts), cancel, reopen (fetch B starts). Both are in flight; whichever resolves *last* wins. If stale A resolves after B, it replaces the freshly-built `PlayerSelectView` — resetting any in-progress create-form state — and on a failure path could flip a working list into the error state.

The codebase already has an established pattern for exactly this (`SkillsView.commitSeq` — a monotonically-increasing sequence number compared after the await, applying the result only if it's still current). This PR applies the same pattern to `CharacterSwitcher.loadCharacters`.

## Changes

- `UI/src/routes/game/CharacterSwitcher.svelte`: add a `loadSeq` counter, incremented at the start of each `loadCharacters()` call; after the `Login/Players` fetch resolves, bail out if the captured sequence number no longer matches the latest, so only the most recent load's result is ever applied.
- `UI/src/tests/routes/game/character-switcher.test.ts`: new regression test — opens the switcher, closes it, reopens it (triggering a second, faster-resolving fetch), then resolves the first (stale) fetch afterward and asserts the UI still reflects the second load's data, not the stale one.

## Test plan

- [x] New unit test `ignores a stale character-list response that resolves after a fresher reopen` passes.
- [x] `npm run test:unit:ci` — full frontend suite, 299 files / 3753 tests, all passing, no coverage-threshold regressions.

Closes #2045

---
_Generated by [Claude Code](https://claude.ai/code/session_01DEiK3hjWT2CCqrmrJqi1qK)_

### DIFF
--- a/UI/src/routes/game/CharacterSwitcher.svelte
+++ b/UI/src/routes/game/CharacterSwitcher.svelte
@@ -115,12 +115,20 @@ const reloadToBoot = () => {
 };
 const enterWorld = reloadToBoot;
 
+// Bumped on every invocation so a stale fetch (from an opened-then-cancelled-then-reopened switcher)
+// can detect it's no longer the latest and skip applying its result, mirroring SkillsView.commitSeq.
+let loadSeq = 0;
+
 const loadCharacters = async () => {
+	const seq = ++loadSeq;
 	phase = 'loading';
 	loadError = null;
 	view = null;
 
 	const response = await new ApiRequest('Login/Players').get();
+	if (seq !== loadSeq) {
+		return;
+	}
 	if (response.status !== 200 || !response.data) {
 		loadError = response.error ?? 'Could not load your characters.';
 		phase = 'error';
--- a/UI/src/tests/routes/game/character-switcher.test.ts
+++ b/UI/src/tests/routes/game/character-switcher.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, afterEach, beforeEach, vi } from 'vitest';
 import { render, fireEvent, cleanup, screen, waitFor } from '@testing-library/svelte';
+import { tick } from 'svelte';
 
 // Doubles for the switcher's wiring: ApiRequest is keyed by route (get for the list, post for the
 // switch/create), and the socket teardown, engine stop, token storage, and takeover are mocked so the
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: stale-response guard for CharacterSwitcher's reload effect (#2045)

**Verdict:** Approve _(submitted as a comment since I'm the PR author; GitHub won't accept a gating verdict on one's own PR)_ — correct, minimal fix that reuses the project's established pattern and lands a real regression test.

The `$effect` fires `loadCharacters()` on every `open` transition with no overlap guard, so an open→cancel→reopen sequence leaves two `Login/Players` fetches racing and the last to resolve wins. The fix captures a per-invocation `loadSeq` and bails immediately after the `await` if it's no longer the latest — the stale response writes nothing (`view`/`phase`/`loadError` are all set *after* the guard), so it can neither clobber the fresh `PlayerSelectView` nor flip a good list into the error state. This is exactly the `SkillsView.commitSeq` pattern the issue pointed at (`skills-view.svelte.ts:538-541`), and the guard placement here is faithful to it.

The open→close (no reopen) case is also safe: the stale fetch applies its result while closed, but the `Popover` renders nothing then and the next open resets `view = null`/`phase = 'loading'` before reloading, so nothing stale is ever surfaced.

### Test coverage

- New unit test in `character-switcher.test.ts` reproduces the precise race: first load held pending, close, reopen (second load resolves with the Mage list), then the stale first load resolves last with the original Rogue-included list — asserting the UI still shows Mage and not Rogue. It fails without the guard and passes with it. Lives at the mirrored `tests/routes/game/` path per the frontend testing rules.

### What's good

- Reuses the existing generation-counter idiom rather than inventing a new mechanism, keeping the two stale-guard sites consistent.
- The comment on the `loadSeq` declaration explains *why* it exists and names the sibling pattern — protects against a later "simplification" that would reintroduce the race.
- Tight, well-scoped diff: 8 lines of fix, a focused regression test, nothing extraneous.

---

PR #2081, opened 2026-07-16
Reload admin reference catalogues on every /admin mount

### LINKED ISSUE #2046: Admin reference catalogues load once per app lifetime — later /admin visits get stale pickers
## Context

`UI/src/routes/admin/+page.svelte:99-104` gates the reference load on `!reference.loaded`, and `reference` is a module singleton — so after the first successful load, re-entering `/admin` (after playing the game, or hours later) never refetches.

Within a session this is mostly self-healing because every entity's `refresh()` writes through to `staticData`, but: (a) catalogues changed by another admin session stay stale in sibling pickers (e.g. the item tool's Granted Skill picker) until that entity's own tool is opened, and (b) `reference.tagCategories`/`tags` for the item/mod Tags tab are never refetched outside the tag tool. Minor today (single-admin project) but a trap once more than one person authors content.

## Suggested fix

Drop the `!reference.loaded` guard (keep it only as a render gate) and reload on every `/admin` mount — the socket reads are cheap and already parallel.

ginderjeremiah:
## Summary

- `admin/+page.svelte`'s `onMount` gated the reference-catalogue load on `!reference.loaded`, so after the first successful load in a page's lifetime, re-entering `/admin` never refetched the shared catalogues (enemies, skills, zones, items, item mods, attributes, tags, tag categories, challenge types, challenges, paths, proficiencies, lessons, classes, skill recipes).
- Within a session this mostly self-healed because every entity's own `refresh()` writes through to `staticData`, but catalogues changed by another admin session (or by this one, hours earlier) stayed stale in **sibling** pickers until that specific entity's tool was opened — e.g. an item's Granted Skill picker wouldn't see a skill renamed via the Skills tool until the Skills tool itself was opened this visit.
- Fix: drop the `!reference.loaded` guard from the load condition and keep `reference.loaded` doing render-gating only (it still backs every `{#if reference.loaded}` check that shows the loading/error state on first entry). The reference load now runs on every `/admin` mount; the socket reads are cheap and already run in parallel via `Promise.all`.

## Test plan

- [x] Added a regression test (`admin-page.test.ts`) asserting a second mount of the admin page refetches the full reference catalogue set instead of skipping the reload because `reference.loaded` was already true from a prior mount.
- [x] `npm run test:unit:ci` — 299 test files / 3754 tests passed.

Closes #2046

---
_Generated by [Claude Code](https://claude.ai/code/session_01QHMoTVSY76m7AjHjFEvn8x)_

### DIFF
--- a/UI/src/routes/admin/+page.svelte
+++ b/UI/src/routes/admin/+page.svelte
@@ -95,10 +95,13 @@ async function loadReference() {
 }
 
 // Guard the route to Admins, then load the shared reference catalogues (used by every entity's
-// select options, tag UI, and derived spawn shares) before rendering any workbench.
+// select options, tag UI, and derived spawn shares) before rendering any workbench. Reloaded on
+// every mount (not gated on reference.loaded, which stays a render gate only) so catalogues
+// changed by another admin session — or since a much earlier visit this app lifetime — aren't
+// left stale in sibling pickers; the socket reads are cheap and already run in parallel.
 onMount(() => {
 	authorized = ensureAdminAccess();
-	if (authorized && !reference.loaded) {
+	if (authorized) {
 		void loadReference();
 	}
 });
--- a/UI/src/tests/routes/admin/admin-page.test.ts
+++ b/UI/src/tests/routes/admin/admin-page.test.ts
@@ -118,3 +118,21 @@ describe('Admin page — reference-data load failure', () => {
 		expect(screen.getByTestId('workbench-stub').textContent).toBe('Enemies');
 	});
 });
+
+describe('Admin page — reference-data reload on every mount (#2046)', () => {
+	it('refetches the reference catalogues on a later mount instead of trusting the already-loaded flag', async () => {
+		mockFetchSocket.mockReset().mockImplementation(async (command: string) => SOCKET_SETS[command] ?? []);
+
+		const first = render(AdminPage);
+		await waitFor(() => expect(screen.getByTestId('workbench-stub')).toBeTruthy());
+		const callsAfterFirstMount = mockFetchSocket.mock.calls.length;
+		expect(callsAfterFirstMount).toBe(Object.keys(SOCKET_SETS).length);
+		first.unmount();
+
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Reload admin reference catalogues on every `/admin` mount (#2046)

**Verdict:** Approve — correct, minimal fix that matches the issue's intent with a genuine regression test. (Submitted as a comment since I'm the PR author; treat it as an approval.)

The change drops the `!reference.loaded` guard from `onMount`'s load condition while keeping `reference.loaded` as a pure render gate, so the shared catalogues refetch on every entry to `/admin`. That closes the stale-sibling-picker trap from #2046 exactly as suggested.

Key thing I checked: **re-entry doesn't flash the loading gate.** `reference.load()` never resets `loaded` back to `false` — it awaits the fresh data and only reassigns `staticData`/`tags`/etc. at the end, leaving `loaded === true` throughout. So a later mount keeps the already-populated workbench rendered and swaps in fresh data underneath, rather than briefly bouncing to the `Loading` gate. That's the right behavior, and it's why the `{#if reference.loaded}` render gates are unaffected.

One benign consequence worth noting (not a finding, no action needed): if a *background* re-load fails, `loadError` is set but never surfaced — the error gate only renders while `reference.loaded` is `false`, which it no longer is on re-entry — so the user keeps the working (stale) workbench plus a `toastError`. That's the desirable outcome; you don't want a failed background refresh to blow away a functioning workbench.

### Test coverage

- The new `#2046` test correctly exercises the real `reference` singleton (only transports stubbed), asserting the second mount refetches the full socket set (`callsAfterFirstMount * 2`) despite `loaded` already being `true` from the first mount. That's exactly the guard being removed. Good, targeted regression coverage.

### What's good

- Scope is tight and the explanatory comment on `onMount` accurately captures *why* the guard was removed and that `reference.loaded` is now render-only — that intent is easy for a future edit to miss, so documenting it inline is the right call.
- Correctly reasons that the socket reads are cheap and already parallel via `Promise.all`, so reloading on every mount is cost-justified.

---

PR #2082, opened 2026-07-16
Make PlayerUpdateBatch.PlayerSaveInProgress nesting-safe via a depth counter (#2001)

### LINKED ISSUE #2001: PlayerUpdateBatch.PlayerSaveInProgress is a plain bool — a nested save scope silently ends the shared-flush window
## Problem

`BeginPlayerSave()` sets `PlayerSaveInProgress` and the returned scope's `Dispose` **unconditionally clears it** (`Game.DataAccess/PlayerUpdateBatch.cs:33, 55-59, 97-100`). The offline-rewards pass opens an outer `IPlayerRepository.BeginBatch()` and then calls `SavePlayer`, which opens and disposes its own nested `BeginPlayerSave` — clearing the flag while the outer scope is still live. Today this is survivable only because of a call-ordering rule the code documents as a trap (`Game.Application/Services/OfflineProgressService.cs:181-192`: progress saves sharing the batch must run *before* `SavePlayer`).

A future orchestration that saves progress after a player save inside a batch scope won't fail loudly — the progress save silently takes the standalone-flush branch (`PlayerProgressRepository.cs:144-171`), re-opening the split-durability window #1921 closed. Classic shape of a regression waiting for the next feature.

## Fix direction

Make the scope nesting-safe: a depth counter (flag true while depth > 0), or an assertion that rejects a nested `BeginPlayerSave`. A few lines plus a unit test covering the nested-scope case.

ginderjeremiah:
## Summary

`PlayerUpdateBatch.BeginPlayerSave()` marked a player save as in-progress with a plain `bool`, and the returned scope's `Dispose()` unconditionally cleared it. `OfflineProgressService` opens an outer `IPlayerRepository.BeginBatch()` scope and then calls `SavePlayer`, which opens its own **nested** `BeginPlayerSave` scope internally — that nested scope's disposal cleared the flag while the outer scope was still live, silently ending the shared-flush window early. The codebase worked around this today only via a documented ordering trap: any progress save sharing the outer batch had to run *before* `SavePlayer`, never after.

## Change

- `PlayerUpdateBatch` now tracks the in-progress state with a depth counter (`_playerSaveDepth`) instead of a bool. `BeginPlayerSave()` increments on entry, the scope's `Dispose()` decrements, and `PlayerSaveInProgress` is `_playerSaveDepth > 0`. The window now stays open for as long as *any* `BeginPlayerSave` scope (nested or not) is live, regardless of disposal order.
- Removed the now-stale "ordering matters" comment in `OfflineProgressService` — a progress save sharing the batch can safely run after `SavePlayer` too, since the fix removes the trap the comment was warning about.
- Added a unit test (`PlayerSaveInProgress_StaysTrueAcrossNestedScope_UntilTheOutermostDisposes`) covering the nested-scope case described in the issue, alongside the existing single-scope coverage.

## Test plan

- [x] `dotnet build` — succeeds
- [x] `dotnet test` (full solution: Core, Infrastructure, Application, Api) — all 3238 tests pass
- [x] New unit test specifically exercises the nested-scope scenario from the issue

Closes #2001

---
_Generated by [Claude Code](https://claude.ai/code/session_01H2cpPKesYBdNtvc3Yr3xDz)_

### DIFF
--- a/Game.Application.Tests/DataAccess/PlayerUpdateBatchTests.cs
+++ b/Game.Application.Tests/DataAccess/PlayerUpdateBatchTests.cs
@@ -28,6 +28,28 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public void PlayerSaveInProgress_StaysTrueAcrossNestedScope_UntilTheOutermostDisposes()
+        {
+            var batch = new PlayerUpdateBatch();
+
+            // Mirrors OfflineProgressService: an outer BeginBatch scope wraps a nested SavePlayer, whose own
+            // internal BeginPlayerSave must not end the outer caller's window when it disposes first (#2001).
+            using (batch.BeginPlayerSave())
+            {
+                Assert.True(batch.PlayerSaveInProgress);
+
+                using (batch.BeginPlayerSave())
+                {
+                    Assert.True(batch.PlayerSaveInProgress);
+                }
+
+                Assert.True(batch.PlayerSaveInProgress);
+            }
+
+            Assert.False(batch.PlayerSaveInProgress);
+        }
+
+        [Fact]
         public void RunFlushedCallbacks_RunsRegisteredActionsInOrder_ThenClearsThem()
         {
             var batch = new PlayerUpdateBatch();
--- a/Game.Application/Services/OfflineProgressService.cs
+++ b/Game.Application/Services/OfflineProgressService.cs
@@ -178,9 +178,6 @@ namespace Game.Application.Services
             // retry together rather than progress already being durably committed while the player save fails
             // (#1921) — the same failure atomicity SavePlayer already gives a progress save reached through its
             // own domain-event dispatch on the live battle-completion path.
-            // Ordering matters here: SavePlayer opens its own nested BeginBatch scope internally and that
-            // scope's disposal ends the shared window (PlayerUpdateBatch.PlayerSaveInProgress is a plain bool,
-            // not a refcount), so any progress save sharing this batch must run *before* SavePlayer, not after.
             using (_playerRepo.BeginBatch())
             {
                 rewards = await ApplyOfflineRewards(player, progress, result, awayWindowEnd, cancellationToken);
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: Make `PlayerUpdateBatch.PlayerSaveInProgress` nesting-safe via a depth counter (#2001)

**Verdict:** Approve _(posted as a comment — GitHub won't accept a self-approval on my own PR)_ — the fix is correct, minimal, and directly removes the foot-gun the issue described.

### What's good

- **Correct fix for the actual problem.** Swapping the `bool` for `_playerSaveDepth` with `PlayerSaveInProgress => _playerSaveDepth > 0` keeps the shared-flush window open for as long as *any* `BeginPlayerSave` scope is live, so a nested `SavePlayer` scope disposing first no longer prematurely ends an outer `BeginBatch` caller's window. This closes exactly the split-durability regression path (#1921) the issue flagged.
- **Removing the ordering-trap comment in `OfflineProgressService` is right, not just cleanup.** The comment documented a constraint that only existed *because* of the bool; with the refcount in place a progress save sharing the batch can safely run after `SavePlayer`, so leaving the warning would have been actively misleading.
- **Doc comments are accurate and appropriately scoped** — they explain *why* it's a depth counter (the nested-scope hazard) rather than restating the mechanics.

### Test coverage

- The new `PlayerSaveInProgress_StaysTrueAcrossNestedScope_UntilTheOutermostDisposes` mirrors the outer-wraps-nested shape from `OfflineProgressService` and asserts the window stays open through the inner scope's disposal and only closes on the outermost — the precise case the old code got wrong. It sits alongside the retained single-scope test, so both behaviors are pinned. Adequate per the project's unit-testing rules for domain/data-access logic.

- `_playerSaveDepth++/--` is non-atomic, but `PlayerUpdateBatch` is request-scoped and driven single-threaded within a save, and the prior `bool` had the same property — no regression, and adding synchronization here would be gold-plating.
- Double-disposing a `PlayerSaveScope` would over-decrement, but every call site uses `using`, so scopes dispose exactly once. Not worth guarding against.

Nice, surgical change.

---

PR #2084, opened 2026-07-17
Resolve JwtOptions through a single bound instance in ConfigureAuth

### LINKED ISSUE #2003: JWT settings resolved through two independent paths (issuance vs validation) with duplicated defaults
## Problem

`ConfigureAuth` binds and validates `JwtOptions` (used by `JwtTokenService` for **issuance**), then separately re-reads `jwtSection["SigningKey"]/["Issuer"]/["Audience"]` raw for the bearer `TokenValidationParameters` (`Game.Api/Startup.cs:311-313, 326-332`), re-declaring the `?? Constants.SERVER_PRINCIPAL` fallbacks that `JwtOptions` already encodes as property defaults.

The issue/validate pair is the one place these three values must stay identical; a future edit to one path's default or key name silently desynchronizes them. The failure is at least loud (every token rejected), but it's exactly the duplication the DRY guideline exists for.

## Fix direction

Resolve one `JwtOptions` instance in `ConfigureAuth` and feed both the DI registration and the `TokenValidationParameters` from it, deleting the raw section reads and duplicate fallbacks.

ginderjeremiah:
## Summary

`ConfigureAuth` (`Game.Api/Startup.cs`) bound and validated `JwtOptions` for DI (used by `JwtTokenService` for token **issuance**), then separately re-read `jwtSection["SigningKey"]/["Issuer"]/["Audience"]` raw to build the JWT bearer `TokenValidationParameters`, re-declaring the `?? Constants.SERVER_PRINCIPAL` fallbacks that `JwtOptions` already encodes as property defaults. The issue/validate pair is the one place these three values must stay identical, so the duplication was a latent desync risk (a future edit to one path's default or key name would silently break the other).

## Changes

- `ConfigureAuth` now binds one local `JwtOptions` instance from the `Jwt` config section and feeds both the DI `AddOptions<JwtOptions>` registration (still validated with `.ValidateOnStart()`) and the `TokenValidationParameters` from it.
- Removed the raw `jwtSection["SigningKey"]/["Issuer"]/["Audience"]` reads and the duplicated `?? Constants.SERVER_PRINCIPAL` fallbacks — `JwtOptions`'s own property defaults are now the single source of truth.
- Note: previously a missing `Jwt:SigningKey` threw `InvalidOperationException` immediately during host configuration; now it fails fast slightly later, via the existing `.ValidateOnStart()` byte-length check (still before the app serves any traffic), since the signing key defaults to `string.Empty` on the bound options object rather than short-circuiting with its own throw. Loud, fail-fast startup behavior is preserved, just via one path instead of two.

## Test plan

- [x] `dotnet build` — clean, no warnings
- [x] `dotnet test --no-build --no-restore` (full solution) — all 3,238 tests pass, including the login/JWT integration suites in `Game.Api.Tests` that exercise `ConfigureAuth` end-to-end

Closes #2003

---
_Generated by [Claude Code](https://claude.ai/code/session_01BJdAeuhJnwJEw4LvD5HCts)_

### DIFF
--- a/Game.Api.Tests/Unit/JwtOptionsTests.cs
+++ b/Game.Api.Tests/Unit/JwtOptionsTests.cs
@@ -1,0 +1,74 @@
+using Game.Api.Auth;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using System.Text;
+using Xunit;
+
+namespace Game.Api.Tests.Unit
+{
+    public class JwtOptionsTests
+    {
+        [Fact]
+        public void ValidateOnStart_ThrowsWhenSigningKeyMissing()
+        {
+            using var provider = BuildValidatedProvider(new ConfigurationBuilder().Build());
+
+            var ex = Assert.Throws<OptionsValidationException>(() =>
+                provider.GetRequiredService<IOptions<JwtOptions>>().Value);
+            Assert.Contains("Jwt:SigningKey must be at least 32 bytes for HMAC-SHA256", ex.Message);
+        }
+
+        [Fact]
+        public void ValidateOnStart_ThrowsWhenSigningKeyTooShort()
+        {
+            var configuration = new ConfigurationBuilder()
+                .AddInMemoryCollection(new Dictionary<string, string?>
+                {
+                    ["Jwt:SigningKey"] = "too-short",
+                })
+                .Build();
+
+            using var provider = BuildValidatedProvider(configuration);
+
+            var ex = Assert.Throws<OptionsValidationException>(() =>
+                provider.GetRequiredService<IOptions<JwtOptions>>().Value);
+            Assert.Contains("Jwt:SigningKey must be at least 32 bytes for HMAC-SHA256", ex.Message);
+        }
+
+        [Fact]
+        public void ValidateOnStart_SucceedsWhenSigningKeyIsAtLeast32Bytes()
+        {
+            var signingKey = new string('a', 32);
+            var configuration = new ConfigurationBuilder()
+                .AddInMemoryCollection(new Dictionary<string, string?>
+                {
+                    ["Jwt:SigningKey"] = signingKey,
+                    ["Jwt:Issuer"] = "test-issuer",
+                    ["Jwt:Audience"] = "test-audience",
+                })
+                .Build();
+            Assert.Equal(32, Encoding.UTF8.GetByteCount(signingKey));
+
+            using var provider = BuildValidatedProvider(configuration);
+
+            var options = provider.GetRequiredService<IOptions<JwtOptions>>().Value;
+            Assert.Equal(signingKey, options.SigningKey);
+            Assert.Equal("test-issuer", options.Issuer);
+            Assert.Equal("test-audience", options.Audience);
+        }
+
+        // Wires up the same options registration ConfigureAuth uses (bind + validate) so the test
+        // exercises the real validation path rather than a re-declaration of it.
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Resolve `JwtOptions` through a single bound instance in `ConfigureAuth` (#2003)

**Verdict:** Approve with one should-fix — the refactor is correct and does exactly what the issue asked; only a test-coverage gap keeps it from being a clean merge. _(Posted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR.)_

Verified the consolidation is behavior-preserving:

- `jwtOptions.Issuer`/`Audience` default to `Constants.SERVER_PRINCIPAL` (`JwtOptions.cs:11-12`), exactly matching the removed `?? Constants.SERVER_PRINCIPAL` fallbacks. Absent-key and explicitly-empty-string config both behave identically to the old `??` reads.
- Missing `Jwt:SigningKey` now defaults to `string.Empty` and fails via the existing `.Validate(… >= 32).ValidateOnStart()` instead of the old immediate `?? throw` — still fail-fast at startup, and the empty-key `TokenValidationParameters` is never reachable because `ValidateOnStart` throws before the app serves traffic. The PR body describes this timing shift accurately.

### Should-fix

- **Missing mirroring test for the JWT `ValidateOnStart` rule.** The codebase has a consistent convention that every options-validation wired in `Startup` gets a unit test mirroring its registration and asserting the fail-fast (`CorsOptionsTests.ValidateOnStart_ThrowsWhenNoOriginsConfigured`, `ForwardedHeadersConfigTests.ValidateOnStart_ThrowsWhenAConfiguredEntryIsUnparseable`, `RateLimitingOptionsTests`). The `Jwt:SigningKey` `>= 32` validation is the sole exception with no such test — and this PR makes that validation the *only* remaining guard for the missing/short-key case (it removed the immediate throw). A small `JwtOptionsTests` mirroring `CorsOptionsTests.ValidateOnStart_*` (throws on empty/short key, succeeds on a ≥32-byte key) would lock in the exact fail-fast guarantee this PR now leans on. Given the PR body explicitly reasons about this validation being the fail-fast path, it's in-scope enough to add here rather than defer.

### Test coverage

- Happy-path auth (issuance + validation through `ConfigureAuth`) stays covered by the existing login/JWT integration suites, and no domain logic changed, so no other new tests are warranted — only the fail-fast unit test noted above.

### What's good

- Clean DRY win that removes a genuine desync risk (issuance vs. validation sharing one resolved instance), and the PR description is precise about the one subtle behavioral nuance (fail-fast now via `ValidateOnStart` rather than an immediate throw) rather than glossing over it.


---

ginderjeremiah:
## Re-review — should-fix resolved ✅

Follow-up commit `5423a2b` adds `JwtOptionsTests`, which is exactly the gap the earlier review flagged. It faithfully mirrors `CorsOptionsTests.ValidateOnStart_*`:

- `ValidateOnStart_ThrowsWhenSigningKeyMissing` / `ThrowsWhenSigningKeyTooShort` — locks in the fail-fast guard that this PR made the *sole* remaining guard for a missing/short key.
- `ValidateOnStart_SucceedsWhenSigningKeyIsAtLeast32Bytes` — happy path, plus asserts `Issuer`/`Audience` bind through, covering the values `TokenValidationParameters` now reads from the shared instance.
- Same `BuildValidatedProvider` (bind + validate + `ValidateOnStart`) shape and convention as the existing options tests.

**Verdict: approve.** The DRY consolidation is behavior-preserving (verified in the prior review), the one should-fix is addressed, and CI is fully green (`all-checks-passed`, backend build + tests). No remaining nits. Merging.

---

PR #2085, opened 2026-07-17
Per-record memoize Workbench EntityStore's recordStates

### LINKED ISSUE #2004: Workbench EntityStore re-canonicalizes and re-validates the entire catalogue on every keystroke
## Problem

Each field edit routes through `EntityStore.patch`, which JSON-clones the edited record and reassigns `items` (`UI/src/routes/admin/workbench/entity-store.svelte.ts:117-130`). That invalidates the memoized `recordStates` `$derived`, which recomputes `status()` — a full canonicalize + `JSON.stringify` of the record **and** its baseline via `canonicalEqual` (`save-helpers.ts:10-32`) — plus `entityWarnings()` for **every** record in the catalogue (`entity-store.svelte.ts:74-90`). The memoization comment is correct for search/selection keystrokes, but every *edit* keystroke pays O(total catalogue JSON) — dominated by items/skills whose child collections stringify on each pass.

Fine at today's content scale; it's the tool's dominant avoidable cost and grows linearly with authored content. Flagged independently by two review passes.

## Fix direction

Per-record memoization keyed on record object identity (patch replaces exactly one record reference), so a keystroke re-diffs only the edited record; alternatively cache `canonicalize` results per unchanged record. Existing entity-store tests should pin the behavior; add one asserting unrelated records' status objects are reference-stable across a patch.

ginderjeremiah:
## Summary

`EntityStore.recordStates` (`UI/src/routes/admin/workbench/entity-store.svelte.ts`) recomputed **every** record's status + validation warnings on every edit, even though `patch()` only ever replaces the one edited record's object reference. Each keystroke was paying O(total catalogue JSON) — a full `canonicalize` + `JSON.stringify` diff (`canonicalEqual` in `save-helpers.ts`) plus `entityWarnings()` for every record — dominated by items/skills whose child collections stringify on every pass.

## Changes

- `recordStates` now memoizes each record's `{ status, warnings }` in a `WeakMap<T, RecordState>` keyed on the record **object itself**. Since `patch()` (and `addItem`/`resetItem`) only replace the reference of the record(s) they touch, every other record is a same-reference cache hit and skips recomputation entirely. Using a `WeakMap` (not a plain `Map`) means a superseded record from an earlier edit is garbage-collected instead of accumulating for the life of the store.
- Two mutations change a record's status **without** replacing its object reference: `removeItem`/`restoreItem` toggling a saved record's membership in the `deleted` set. Those now bump a `statusEpoch` counter that the derived checks to invalidate the whole cache — every other mutator (`patch`, `addItem`, `resetItem`, `save`, `discard`) already invalidates itself via reference replacement.

## Test plan

- [x] Added `entity-store.test.ts` cases:
  - an unrelated record's `RecordState` object stays reference-stable across a `patch()` to another record (proves the cache hit)
  - `removeItem`/`restoreItem` correctly flip a previously-cached record's status despite no reference change (proves the epoch invalidation)
- [x] `npm run test:unit:ci` — full suite, 3757/3757 passing

Closes #2004

---
_Generated by [Claude Code](https://claude.ai/code/session_01C5vFfvFrbZhpVnjqDZ63J9)_

### DIFF
--- a/UI/src/routes/admin/workbench/entity-store.svelte.ts
+++ b/UI/src/routes/admin/workbench/entity-store.svelte.ts
@@ -65,16 +65,35 @@ export class EntityStore<T extends Identified> {
 		return recordsEqual(record, baseline) ? 'clean' : 'modified';
 	}
 
+	/** Bumped whenever a record's status can change *without* its object reference changing — only
+	 *  {@link removeItem}/{@link restoreItem} toggling a saved record's membership in {@link deleted}.
+	 *  Every other mutator ({@link patch}, {@link addItem}, {@link resetItem}, `save`, `discard`)
+	 *  replaces the affected record(s) with a new object, which already misses the reference cache
+	 *  below on its own. */
+	private statusEpoch = $state(0);
+	private stateCacheEpoch = -1;
+	private stateCache = new WeakMap<T, RecordState>();
+
 	/**
-	 * Per-record status + validation warnings, keyed by id. Memoised here so the consumers (the list
-	 * rows, the page summary, the save-bar counts) read O(1) lookups instead of each re-running the
-	 * structural diff per record. Recomputes only when the records, baseline, or pending deletes
-	 * change — a search keystroke or a selection change (which touch neither) reuses the cached map.
+	 * Per-record status + validation warnings, keyed by id. Memoised **per record object** so an edit
+	 * keystroke — which replaces only the one record {@link patch} touched — re-diffs just that record
+	 * instead of re-canonicalizing the whole catalogue; every other record is a same-reference cache
+	 * hit. The cache is a `WeakMap` keyed on the record itself (not its id) so a superseded record from
+	 * an earlier edit is garbage-collected instead of accumulating for the life of the store.
 	 */
 	recordStates = $derived.by<Record<number, RecordState>>(() => {
+		if (this.statusEpoch !== this.stateCacheEpoch) {
+			this.stateCache = new WeakMap();
+			this.stateCacheEpoch = this.statusEpoch;
+		}
 		const map: Record<number, RecordState> = {};
 		for (const record of this.items) {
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Per-record memoization of Workbench `EntityStore.recordStates` (closes #2004)

**Verdict:** Approve — correctly fixes the O(total catalogue JSON)-per-keystroke recompute called out in #2004, with sound cache invalidation and matching tests. (Submitted as a comment since I'm the PR author; intent is approve.)

### Correctness

The per-record `WeakMap` memo is safe because the two inputs to a cached `RecordState` are both accounted for:

- **`warnings`** are a pure function of the single record — `warn?: (rec: T) => string | null` and `fieldWarn`/`sectionWarnings` never read sibling records — so keying the cache on the record object alone can't go stale from a neighbor's edit.
- **`status(record)`** depends only on the `deleted` set and `baseMap`. `baseMap` changes only in `save()`, which replaces every record reference (all cache misses), and the two mutators that flip status *without* replacing a reference — `removeItem`/`restoreItem` — correctly bump `statusEpoch` to flush the whole cache. `resetItem`/`discard` replace the affected references, so no epoch bump is needed there. The invalidation reasoning in the comments is precise and correct.

### Test coverage

Good — the added cases cover both directions: reference-stability of an unrelated record's `RecordState` across a `patch()` (the cache-hit the issue explicitly asked for), and `removeItem`/`restoreItem` flipping a previously-cached record despite no reference change (the epoch invalidation). That's exactly the behavior that could silently regress.

### What's good

- The `WeakMap` (over a plain `Map`) is the right call — superseded records GC instead of accumulating for the store's lifetime, and the doc comment explains why.
- Change is tightly scoped to the hot path; no behavior change beyond performance, and the existing suite pins the semantics.

CI is green (frontend suites pass; backend correctly skipped for a frontend-only change). Merging.

---

PR #2086, opened 2026-07-17
Derive Statistics/Codex view-model stats from the live store

### LINKED ISSUE #2005: Statistics/Codex copy the reactive statistics store into mount-time snapshots — intra-screen freshness splits
## Problem

`Statistics.svelte` and `Codex.svelte` copy `statistics.stats` into view-local state after a forced load (`UI/src/routes/game/screens/stats/Statistics.svelte:48-49`, `codex/Codex.svelte:40-41`), while other parts of the same screens read the store live (e.g. `zones-tab-view.svelte.ts` uses `statistics.isZoneCleared`).

**Failure scenario:** with the Codex open on a zone dossier, the background boss loop clears the zone → `markZoneCleared` updates the store (`$stores/statistics.svelte.ts:55-65`) → the zone rail flips to "cleared" live while the dossier's "Your record" row (fed from the mount-time snapshot) stays stale until remount. One surface, two answers — and the store is meant to be the single source (the live-baseline convention from #1809 points the same way).

## Fix direction

Derive the view data from the store (`$derived`) instead of snapshotting it post-load; the forced load stays as the freshness trigger. Small test asserting a store update after mount reflects in the view model.

ginderjeremiah:
## Summary

`Statistics.svelte` and `Codex.svelte` copied `statistics.stats` into view-local `$state` after their mount-time forced load (`view.stats = statistics.stats`), while other parts of the same screens read the store live (e.g. `zones-tab-view.svelte.ts` uses `statistics.isZoneCleared`). This split the store's single-source-of-truth guarantee: a background write to the store — e.g. the fight screen's optimistic `markZoneCleared` the moment a boss is defeated — updated the zone rail live but left the dossier's "Your record" section (fed from the mount-time snapshot) stale until remount.

## Changes

- `StatisticsView.stats` (`statistics-view.svelte.ts`) and `CodexView.stats` (`codex-view.svelte.ts`) are now `readonly stats = $derived(statistics.stats)` instead of a `$state` field assigned once in `onMount`. The forced load (`statistics.load(true)`) stays as the freshness trigger; the view just stops snapshotting the result.
- Removed the now-redundant `view.stats = statistics.stats` assignments in `Statistics.svelte` and `Codex.svelte`.
- Updated the pure view-model unit tests (`statistics-view.test.ts`, `codex-view.svelte.test.ts`) to seed the mocked `statistics` store directly (`statistics.stats = ...`) instead of assigning the now-readonly `view.stats`.
- Added regression tests against the **real** (unmocked) `statistics` store in `Statistics.test.ts` and `Codex.test.ts` that mount the screen, then call `statistics.markZoneCleared(...)` — mirroring the exact background-write scenario from the issue — and assert the rendered stat card / dossier record updates without remounting.

## Test plan

- [x] `npm run test:unit:ci` — 299 files / 3759 tests pass, coverage thresholds hold
- [x] New regression tests specifically reproduce the issue's stale-dossier / stale-card scenario and fail without the fix (verified locally before applying it)

Closes #2005

---
_Generated by [Claude Code](https://claude.ai/code/session_01FXwrK4aNqj2hbEvL28CPMv)_

### DIFF
--- a/UI/src/routes/game/screens/codex/Codex.svelte
+++ b/UI/src/routes/game/screens/codex/Codex.svelte
@@ -35,9 +35,9 @@ const view = new CodexView(navigation.consumePayload<CodexNavPayload>());
 
 onMount(async () => {
 	// Force-refresh the player's statistics (the dossier's "your record"); challenge progress is
-	// loaded once at boot, so a plain load is enough.
+	// loaded once at boot, so a plain load is enough. view.stats reads the store live, so no
+	// snapshot assignment is needed here.
 	await Promise.all([statistics.load(true), playerChallenges.load()]);
-	view.stats = statistics.stats;
 	view.statsError = statistics.error;
 	view.statsLoading = false;
 	view.challengesError = playerChallenges.error;
--- a/UI/src/routes/game/screens/codex/codex-view.svelte.ts
+++ b/UI/src/routes/game/screens/codex/codex-view.svelte.ts
@@ -22,13 +22,12 @@
    The reactive layer is composed from one class per tab (`EnemiesTabView`, `ZonesTabView`,
    `SkillsTabView`, each in its own `*-tab-view.svelte.ts` file) so every tab's derivations stay
    independently readable/testable. This CodexView is the thin orchestrator: it owns only what's
-   genuinely cross-tab — the active tab, the once-fetched player statistics + challenge-error state,
+   genuinely cross-tab — the active tab, the live player statistics + challenge-error state,
    the shared statistics query engine, and the cross-links between dossiers — and every other
    property/method below simply delegates to the owning tab's instance, so consuming components (and
    the existing tests) keep reading `view.<x>` exactly as before. */
 
-import type { IPlayerStatistic } from '$lib/api';
-import { staticData } from '$stores';
+import { staticData, statistics } from '$stores';
 import {
 	CODEX_TABS,
 	type CodexTab,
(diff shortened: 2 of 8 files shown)


ginderjeremiah (commented):
## Review: derive Statistics/Codex view-model stats from the live store (#2005)

**Verdict:** Approve — the change applies the project's documented live-baseline pattern correctly, and the tests reproduce the exact bug the issue describes. (Posted as a comment since I'm the PR author; GitHub won't accept a gating event on one's own PR.)

This is the right fix and it's aligned with the codebase's own conventions rather than generic taste. `frontend.md` ("Local-edit view-models: live baseline vs. a live-reference manager", from #1809) states that a view-model must never snapshot a store's array-valued state at construction and should derive it live instead — which is precisely the split the issue flagged (`statistics.isZoneCleared` read live while `view.stats` was a mount-time snapshot). Converting `stats` to `readonly stats = $derived(statistics.stats)` on both `StatisticsView` and `CodexView`, and dropping the `view.stats = statistics.stats` assignments, closes that gap.

Correctness verified against the store: `statistics.stats` is `$state` and `markZoneCleared` reassigns it wholesale (`stats = stats.map(...)` / `[...stats, ...]`), so the `$derived` observes the optimistic background write — exactly the fight-screen → open-dossier scenario in #2005. The `loading`/`statsError` flags stay explicit `$state` (unchanged), so nothing regresses the spinner/error handling.

### Test coverage
Solid and well-targeted:
- Pure view-model tests (`statistics-view.test.ts`, `codex-view.svelte.test.ts`) now seed the mocked store (`statistics.stats = …`) *before* constructing the view — correct, since a `$derived` reads live on access.
- New regression tests in `Statistics.test.ts` and `Codex.test.ts` mount against the **real** store and assert a post-mount `markZoneCleared(...)` updates the rendered card / dossier without remounting — a faithful reproduction of the issue's failure mode on both surfaces.

### What's good
- Dead code removed cleanly (unused `IPlayerStatistic` import in `codex-view`, the redundant snapshot assignments).
- The mock refactor to a getter/setter-backed `stats` keeps the pure tests honest without pulling in `$state`.

No blocking findings, no should-fixes, no nits. Clean change.

---

PR #2087, opened 2026-07-17
Delete CodexView's pass-through shim; remove dead WorkbenchReference test-only members

### LINKED ISSUE #2006: Frontend dead-surface cleanups: CodexView pass-through shim; test-only WorkbenchReference members
Two same-flavor cleanups (companions to #1958's dead-export list):

## 1. `CodexView` is a ~250-line hand-maintained pass-through shim

`UI/src/routes/game/screens/codex/codex-view.svelte.ts:161-346` is one-line forwards to `EnemiesTabView`/`ZonesTabView`/`SkillsTabView`, acknowledged in its own comment as pre-split compatibility. It's now pure carrying cost: every new tab member needs a hand-written forward, and a missed one is a silent runtime `undefined` rather than a compile error. Components can consume the exposed tab-view instances directly, deleting ~180 lines.

## 2. Test-only members on `WorkbenchReference`

`pathName` (`UI/src/routes/admin/workbench/reference.svelte.ts:254`), `entityCatalog` (`:288-289`), `skillRecords` (`:308`), and `skillBaseDamage` (`:310`) have no production caller — they are referenced only by `src/tests/routes/admin/workbench/reference.test.ts` (verified repo-wide excluding tests). Unused API kept alive by its own tests; delete the members and their tests, or inline them where a production caller was intended.

ginderjeremiah:
## Summary

Two same-flavor dead-surface cleanups from the issue:

**1. `CodexView`'s ~250-line hand-maintained pass-through shim**

`codex-view.svelte.ts` used to re-declare every `EnemiesTabView`/`ZonesTabView`/`SkillsTabView` member as a one-line forward (`get selectedEnemy() { return this.enemiesTab.selectedEnemy; }`, etc.), acknowledged in its own comment as pre-split compatibility. That's now pure carrying cost — a missed forward on a new tab member is a silent runtime `undefined` rather than a compile error.

- `enemiesTab`/`zonesTab`/`skillsTab` are now public (`private readonly` → `readonly`).
- Every tab-specific getter/setter/method delegate on `CodexView` is deleted (~194 net lines off `codex-view.svelte.ts`). What remains is genuinely cross-tab: the active tab, the shared stats/challenge-error state, the statistics query engine, and the `openEnemy`/`openZone`/`openSkill` cross-links.
- Every consuming component (`EnemiesTab`, `EnemyTable`, `EnemyDossier`, `EnemyChallengesPanel`, `EnemySkillsPanel`, `EnemySpawnsPanel`, `AttributesPanel`, `ZoneRail`, `ZoneDossier`, `ZoneSpawnPanel`, `SkillTable`, `SkillDossier`) now reads/calls through the owning tab instance directly (e.g. `view.enemiesTab.selectedEnemy`, `view.zonesTab.selectZone(id)`) instead of the flat `view.<x>` API. Components still take the whole `view: CodexView` prop (they also need the cross-tab links and shared stats state), just address through the tab instance for tab-scoped data.
- `codex-view.svelte.test.ts` updated to match (156+ call sites); no test scenarios changed, only the access path.

**2. Test-only `WorkbenchReference` members**

`pathName`, `entityCatalog`, `skillRecords`, and `skillBaseDamage` had no production caller — verified repo-wide excluding tests — and existed only to be exercised by `reference.test.ts`. Deleted them and their now-orphaned test coverage. While in there, also removed `proficiencyName`, which turned out to have **zero** callers, not even in tests (a stricter case of the same problem, one line, same file/region already being edited).

## Test plan

- [x] `npm run lint` (eslint + svelte-check) — 0 errors across 1199 files. Since `CodexView` is fully typed, a missed/incorrect access-path conversion would have failed type-checking.
- [x] `npm run test:unit:ci` — 3758/3758 tests passed (299 files), including the full `CodexView` suite (search/filter/sort/selection/cross-links/sub-tabs/retired-record resolution/deep-link payloads) and the trimmed `WorkbenchReference` suite.
- [x] `npm run build` — production build succeeds.
- Pure refactor (no behavior/logic change — only property-access paths moved), so no new test scenarios were needed.

Closes #2006

---
_Generated by [Claude Code](https://claude.ai/code/session_01UXj9oqYuAEK9DFwNvwvbZu)_

### DIFF
--- a/UI/src/routes/admin/workbench/reference.svelte.ts
+++ b/UI/src/routes/admin/workbench/reference.svelte.ts
@@ -255,9 +255,6 @@ class WorkbenchReference {
 		{ value: -1, text: 'None (ungated)' },
 		...this.retireableOptions(staticData.proficiencies ?? [], keep)
 	];
-	pathName = (id: number) => staticData.paths?.[id]?.name;
-	proficiencyName = (id: number) => staticData.proficiencies?.[id]?.name;
-
 	// ── Challenges ──
 	challengeTypeOptions = (): SelectOption[] => this.challengeTypes.map((t) => ({ value: t.id, text: t.name }));
 	challengeTypeById = (id: number) => this.challengeTypes.find((t) => t.id === id);
@@ -289,8 +286,6 @@ class WorkbenchReference {
 			? (staticData.enemies ?? []).filter((e) => !bossOnly || e.isBoss || e.id === keep)
 			: (this.entitySource(entityType) ?? []);
 
-	entityCatalog = (entityType: EEntityType, bossOnly = false): { id: number; name: string }[] =>
-		this.entityRecords(entityType, bossOnly).map((e) => ({ id: e.id, name: e.name }));
 	/** Target-entity picker options. Retired records are excluded unless `keep` is the current target. */
 	entityOptions = (entityType: EEntityType, bossOnly = false, keep?: number): SelectOption[] =>
 		this.retireableOptions(this.entityRecords(entityType, bossOnly, keep), keep);
@@ -309,9 +304,7 @@ class WorkbenchReference {
 		const mod = staticData.itemMods?.[id];
 		return mod ? this.modTypeName(mod.itemModTypeId) : undefined;
 	};
-	skillRecords = () => staticData.skills ?? [];
 	skillName = (id: number) => staticData.skills?.[id]?.name;
-	skillBaseDamage = (id: number) => staticData.skills?.[id]?.baseDamage;
 	skillRetired = (id: number) => !!staticData.skills?.[id]?.retiredAt;
 
 	// ── Name lookups ──
--- a/UI/src/routes/game/screens/codex/AttributesPanel.svelte
+++ b/UI/src/routes/game/screens/codex/AttributesPanel.svelte
@@ -8,21 +8,21 @@
 		<button
 			type="button"
 			class="scaling-toggle"
-			class:on={view.scaling}
-			aria-pressed={view.scaling}
+			class:on={view.enemiesTab.scaling}
+			aria-pressed={view.enemiesTab.scaling}
 			data-testid="codex-scaling-toggle"
-			onclick={() => view.toggleScaling()}
+			onclick={() => view.enemiesTab.toggleScaling()}
 		>
 			Show scaling
 		</button>
 	</div>
 
-	{#if view.range}
-		{#if view.range.fixed}
+	{#if view.enemiesTab.range}
+		{#if view.enemiesTab.range.fixed}
 			<div class="fixed-level">
 				<span>FIXED ENCOUNTER</span>
 				<span class="rule"></span>
-				<span class="fixed-val">LVL {view.level}</span>
+				<span class="fixed-val">LVL {view.enemiesTab.level}</span>
 			</div>
 		{:else}
 			<div class="level-control">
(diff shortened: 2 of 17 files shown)


ginderjeremiah (commented):
## Review: Delete CodexView pass-through shim + dead WorkbenchReference test-only members (closes #2006)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating verdict on my own PR) — the refactor does exactly what #2006 asked, cleanly and verifiably.

I reviewed the full diff plus the surrounding code. Both cleanups are sound:

**1. CodexView shim removal.** What remains on `CodexView` after the cut is genuinely cross-tab — active `tab`, shared `stats`/`statsLoading`/`statsError`/`challengesError`, the `StatisticsData` query engine, and the `openEnemy`/`openZone`/`openSkill` cross-links. Every tab-scoped member now addresses through the owning tab instance. I verified this against the intended split:
- Cross-link handlers still correctly route through the orchestrator (`view.openEnemy/openZone/openSkill`), and shared stats state stays on `view` (e.g. `StatisticsPanel` still reads `view.statsLoading`/`view.statsError` while its `stats` prop comes from `view.enemiesTab.statistics` / `view.zonesTab.zoneStatistics` / `view.skillsTab.skillStatistics`) — the right division.
- Repo-wide sweep for the old flat access paths (`view.selectedEnemy`, `view.selectEnemy`, `view.skillProvenance`, etc.) found **zero** missed conversions outside the tab instances, including the Statistics deep-link path. A silent runtime `undefined` was the exact risk the issue called out, and none remain. Since `CodexView` is fully typed, `svelte-check` would also have caught any bad path — and CI is green.

**2. WorkbenchReference dead members.** Confirmed no production callers repo-wide (excluding tests) for `pathName`, `entityCatalog`, `skillRecords`, `skillBaseDamage`, and `proficiencyName` — the only `proficiencyName` hits are unrelated local vars/params in other files, not the deleted method. Removing the members and their now-orphaned coverage is the correct call and aligns with the "don't leave dead code behind" rule.

### Nits
- `UI/src/tests/routes/admin/workbench/SectionRenderer.test.ts:20` still carries a `skillRecords: vi.fn(() => [])` stub in `mockReference`. That mock is documented as "stubbed with everything they touch," but with `skillRecords` deleted from the reference nothing touches it anymore, so the stub is now dead. It's an untyped literal so it doesn't break the type-check, but since this PR is specifically about removing the `skillRecords` surface, dropping that one line here keeps the cleanup complete. Non-blocking.

### Test coverage
- `codex-view.svelte.test.ts`: every changed line is an access-path update (`view.X` → `view.<tab>.X`); a `describe`/`it` diff shows **0 removed and 0 added** blocks, so no scenario was weakened — coverage of search/filter/sort/selection/cross-links/sub-tabs/deep-links is preserved.
- `reference.test.ts`: the removed assertions map 1:1 to the deleted members; remaining lookups still covered.

### What's good
- The retained module doc-comment accurately describes the new "thin orchestrator + one view per tab, consumed through the tab instance" model — a later reader gets the design without archaeology.
- The conversion is disciplined: tab-scoped through the tab, cross-tab through `view`, no half-measures. This turns a class of silent-`undefined` regressions into compile errors, which is the whole point.


---

ginderjeremiah (commented):
## Review: Delete CodexView's pass-through shim + dead WorkbenchReference test-only members (closes #2006)

**Verdict:** Approve (posted as a COMMENT since I'm the PR author, so GitHub won't accept a gating event from me — treat this as an approve). Clean, well-scoped refactor with no behavior change; one minor cleanup miss noted below.

### Verification (this is a pure refactor, so faithfulness is the whole game)

- **Access-path conversions are faithful.** Spot-checked every touched component: each `view.<x>` → `view.<tab>.<x>` conversion points at the same member on the owning tab instance. No getter/setter semantics changed.
- **Shared/cross-tab state correctly stayed on `view.`** — `statsLoading`, `statsError`, and `challengesError` are still read off `CodexView` (e.g. `EnemyDossier.svelte`, `EnemyChallengesPanel.svelte`), not accidentally moved onto a tab.
- **Cross-links intact.** All cross-tab jumps still route through the tab-switching orchestrator methods (`view.openEnemy` / `openZone` / `openSkill`, and `view.selectTab`) — none were rerouted to the tab-scoped `select*` methods, which would have silently stopped switching tabs.
- **No dangling imports.** `codex-view.svelte.ts`'s trimmed import list is fully used; the retained `EnemyRowVM` / `ZoneProjectionVM` are the re-exported types.
- **Deleted `WorkbenchReference` members have zero production callers** (grepped repo-wide): `pathName`, `entityCatalog`, `skillRecords`, `skillBaseDamage`. Confirmed `proficiencyName` genuinely has **zero** callers anywhere including tests — the other `proficiencyName` hits are unrelated local vars in `proficiency-feedback.ts` / `ItemDrawer` / `ItemTooltip`.
- **`reference.test.ts` cleanup is faithful** — the removed assertions map exactly to the deleted members; the surviving assertions are unchanged.

### Nits

- `src/tests/routes/admin/workbench/SectionRenderer.test.ts:20` still stubs `skillRecords: vi.fn(() => [])` on its `mockReference`, but `skillRecords` no longer exists on the real `WorkbenchReference` and no production `SectionRenderer` child calls it (that's exactly why it was safe to delete). Harmless to the passing test, but it's a stale reference to a just-deleted API and slightly contradicts the "no dead code left behind" rule. That file isn't in this diff, so it's a fine follow-up rather than a blocker — worth dropping the line while it's fresh.

### Test coverage

Appropriate for a pure refactor. No new scenarios are warranted — the `codex-view.svelte.test.ts` suite is converted in place (access path only, scenarios unchanged), and `reference.test.ts` had only the orphaned coverage removed. Because `CodexView` is fully typed, `svelte-check` passing is a real guarantee that no consumer or test still references the removed flat API.

### What's good

- Removing the hand-maintained pass-through layer is the right call: a missed forward on a new tab member was a silent runtime `undefined` instead of a compile error, and this converts that class of bug into a type error. The remaining `CodexView` surface is now genuinely just the cross-tab orchestration, which reads clearly.
- Bundling the `proficiencyName` removal (stricter zero-caller case, same file/region) into this cleanup was sensible scope, not scope creep.


---

ginderjeremiah:
## Review — approve

Full review of this pure refactor (closes #2006). Verified faithfulness across all 12 touched components: every `view.x` → `view.<tab>.x` conversion targets the same member with no changed getter/setter semantics; genuinely shared state (`statsLoading`, `statsError`, `challengesError`) correctly stayed on `view.`; cross-links still route through the tab-switching `view.openEnemy/openZone/openSkill` + `view.selectTab` (not the tab-scoped `select*` methods, which would have silently stopped switching tabs); no dangling imports; `reference.test.ts` cleanup maps exactly to the deleted members. Confirmed repo-wide that the deleted `WorkbenchReference` members (`pathName`, `entityCatalog`, `skillRecords`, `skillBaseDamage`, `proficiencyName`) have zero production callers, and `proficiencyName` has none even in tests.

**One nit found and fixed in-PR** (`fce246c`): the deleted `skillRecords` method still had a mock stub in `SectionRenderer.test.ts` (a file outside the original diff) — dead now that the method is gone, so I removed it. SectionRenderer suite still passes 7/7. Left it in-PR rather than as a follow-up since it's dead code this PR's own deletion orphaned.

Merging once CI re-runs green on the fix commit.

---

PR #2090, opened 2026-07-17
Fold PlayerProgressRepository's hash read + TTL slide into one round trip

### LINKED ISSUE #2019: Fold PlayerProgressRepository's GET+ExpireAndForget hash read into a single round trip
## Context

Follow-up from #1925 / #2018. That work collapsed the sliding-TTL `Get` + fire-and-forget `ExpireAndForget` pattern in `PlayerRepository.GetPlayer` and `SessionStore.GetSession` into a single Redis `GETEX` round trip via the new `ICacheService.GetAndRefreshExpiry`.

`PlayerProgressRepository.GetCachedProgress` (`Game.DataAccess/Repositories/PlayerProgressRepository.cs:199`) has the same read-then-slide shape, but its key is a Redis **hash**, not a string — `GETEX` only applies to strings, so it was deliberately left out of that PR.

## Proposal

Fold the `HashGetAllIfExists` + `ExpireAndForget` pair into one round trip with a small custom Lua script that does `TYPE`/`HGETALL` + `PEXPIRE` atomically, mirroring the existing `HashGetAllIfExists` script in `RedisService` (which already does `TYPE`-then-`HGETALL` in one call). This halves the command volume on the per-battle cached-progress read path, matching the win from #2018 on the string keys.

## Notes

- Keep the existing wrong-representation self-heal behavior from `HashGetAllIfExists` (a non-hash key is deleted and treated as a miss).
- Add integration coverage for the new primitive (hit refreshes TTL + returns fields; miss returns null and creates no key), consistent with the `GetAndRefreshExpiry` tests added in #2018.

/label tech debt

ginderjeremiah:
## Summary

Follow-up from #1925 / #2018, which collapsed the sliding-TTL `Get` + fire-and-forget `ExpireAndForget` pattern into a single Redis `GETEX` round trip for string keys (`PlayerRepository.GetPlayer`, `SessionStore.GetSession`). `PlayerProgressRepository.GetCachedProgress` has the same read-then-slide shape, but its key is a Redis **hash**, so `GETEX` doesn't apply — this was deliberately left out of that PR and tracked here as #2019.

- Added `ICacheService.HashGetAllAndRefreshExpiry`, implemented in `RedisService` as a single Lua script that does `TYPE`/`HGETALL` + `PEXPIRE` atomically — mirroring the existing `HashGetAllIfExists` script's shape (including its wrong-representation self-heal: a non-hash key is deleted and treated as a miss) with the TTL refresh folded into the same round trip.
- Swapped `PlayerProgressRepository.GetCachedProgress`'s `HashGetAllIfExists` + `ExpireAndForget` pair for the new primitive, halving Redis command volume on the per-battle cached-progress read path.
- `HashGetAllIfExists` itself is untouched and still used elsewhere (`PlayerRepository`), so it stays.

## Test plan

- [x] `dotnet test --no-build --no-restore` — full suite green (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests all passed)
- [x] Added integration coverage for the new primitive in `RedisServiceIntegrationTests`: hit refreshes TTL + returns fields in one round trip, miss returns null and creates no key, and the non-hash self-heal behavior — matching the `GetAndRefreshExpiry` tests added in #2018.
- [x] Updated the two hand-rolled `ICacheService` test doubles (`SocketCommandProcessorTests`) to implement the new interface member.

Closes #2019

---
_Generated by [Claude Code](https://claude.ai/code/session_014r89iMjpsNibzDjTdL8Ljt)_

### DIFF
--- a/Game.Infrastructure/Cache/Redis/RedisService.cs
+++ b/Game.Infrastructure/Cache/Redis/RedisService.cs
@@ -162,19 +162,7 @@ namespace Game.Infrastructure.Cache.Redis
                 + "return redis.call('hgetall', KEYS[1])",
                 [key]), cancellationToken);
 
-            if (result.IsNull)
-            {
-                return null;
-            }
-
-            var flat = (RedisValue[])result!;
-            var fields = new Dictionary<string, string>(flat.Length / 2);
-            for (var i = 0; i < flat.Length; i += 2)
-            {
-                fields[flat[i]!] = flat[i + 1]!;
-            }
-
-            return fields;
+            return MapHashResult(result);
         }
 
         public async Task<Dictionary<string, string>?> HashGetAllAndRefreshExpiry(string key, TimeSpan expiry, CancellationToken cancellationToken = default)
@@ -194,6 +182,14 @@ namespace Game.Infrastructure.Cache.Redis
                 + "return result",
                 [key], [(RedisValue)(long)expiry.TotalMilliseconds]), cancellationToken);
 
+            return MapHashResult(result);
+        }
+
+        // Shared by HashGetAllIfExists and HashGetAllAndRefreshExpiry: both scripts return either a Lua false
+        // (told apart from an empty hash by RedisResult.IsNull) or the flat HGETALL reply, so the
+        // flat-array-to-dictionary conversion — and the null-forgiving indexing it requires — lives in one place.
+        private static Dictionary<string, string>? MapHashResult(RedisResult result)
+        {
             if (result.IsNull)
             {
                 return null;
--- a/Game.Infrastructure/Cache/Redis/RedisService.cs
+++ b/Game.Infrastructure/Cache/Redis/RedisService.cs
@@ -177,6 +177,38 @@ namespace Game.Infrastructure.Cache.Redis
             return fields;
         }
 
+        public async Task<Dictionary<string, string>?> HashGetAllAndRefreshExpiry(string key, TimeSpan expiry, CancellationToken cancellationToken = default)
+        {
+            // Same TYPE-then-HGETALL script as HashGetAllIfExists, plus a PEXPIRE on the hit path so a
+            // sliding-expiration hash read pays one round trip instead of an awaited HGETALL followed by a
+            // fire-and-forget expire (mirroring GetAndRefreshExpiry's GETEX win for string keys, which GETEX
+            // itself can't serve here since it only applies to strings). The PEXPIRE makes this a write, so it
+            // routes through ObserveWrite like GetAndRefreshExpiry rather than the plain-read path
+            // HashGetAllIfExists uses.
+            var result = await ObserveWrite(Redis.ScriptEvaluateAsync(
+                "local t = redis.call('type', KEYS[1]).ok "
+                + "if t == 'none' then return false end "
+                + "if t ~= 'hash' then redis.call('del', KEYS[1]) return false end "
+                + "local result = redis.call('hgetall', KEYS[1]) "
+                + "redis.call('pexpire', KEYS[1], ARGV[1]) "
+                + "return result",
+                [key], [(RedisValue)(long)expiry.TotalMilliseconds]), cancellationToken);
+
+            if (result.IsNull)
+            {
+                return null;
+            }
+
+            var flat = (RedisValue[])result!;
+            var fields = new Dictionary<string, string>(flat.Length / 2);
+            for (var i = 0; i < flat.Length; i += 2)
+            {
+                fields[flat[i]!] = flat[i + 1]!;
+            }
+
+            return fields;
+        }
+
         public void HashSetAndForget(string key, IReadOnlyDictionary<string, string> fields, TimeSpan expiry)
         {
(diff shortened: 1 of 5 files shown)


Game.Infrastructure/Cache/Redis/RedisService.cs
```diff
@@ -177,6 +177,38 @@ public void DeleteAndForget(string key)
...
+            }
+
+            var flat = (RedisValue[])result!;
+            var fields = new Dictionary<string, string>(flat.Length / 2);
+            for (var i = 0; i < flat.Length; i += 2)
+            {
+                fields[flat[i]!] = flat[i + 1]!;
+            }
+
+            return fields;
```
ginderjeremiah, line 205: Nit (DRY): this flat-array → `Dictionary<string,string>` conversion is now duplicated verbatim between `HashGetAllIfExists` and `HashGetAllAndRefreshExpiry`, and the underlying Lua script is nearly identical too. Consider extracting a small private helper (e.g. `MapHashResult(RedisResult result)`) that both methods share. It would also localize the `result!` / `flat[i]!` null-forgiving usage (which `CLAUDE.md` asks us to avoid) to one place. Not blocking — the duplication is small and the copy is faithful.

---

ginderjeremiah (commented):
## Review: Fold `PlayerProgressRepository`'s hash read + TTL slide into one round trip (closes #2019)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change faithfully implements the tracked follow-up with correct semantics and matching test coverage; only one minor DRY nit.

I checked the new `HashGetAllAndRefreshExpiry` against the primitives it mirrors (`HashGetAllIfExists`, `GetAndRefreshExpiry`) and the issue's acceptance criteria:

- **Correct behavior.** The Lua script reuses the exact `TYPE`-then-`HGETALL` shape of `HashGetAllIfExists` (including the non-hash self-heal via `del`) and adds `PEXPIRE` only on the hit path, so a miss/non-hash key stays a no-op — matching `GetAndRefreshExpiry` and the issue's requirement. `PEXPIRE` takes milliseconds, and `ARGV[1]` is passed as `expiry.TotalMilliseconds`, so units line up.
- **Right budget path.** Since the added `PEXPIRE` makes this a write, routing through `ObserveWrite` (not the `RedisCommandBudget.Read` path `HashGetAllIfExists` uses) is the correct call and is consistent with `GetAndRefreshExpiry`. The inline comment explaining this is genuinely useful.
- **Repository swap is behavior-preserving.** `GetCachedProgress` drops the separate `ExpireAndForget` and the sliding-TTL intent is preserved in one round trip; dead code (the old fire-and-forget expire) is removed, and the relocated comment reads clearly.
- **Interface doc** on `ICacheService` is accurate and cross-references `GetAndRefreshExpiry`/`HashGetAllIfExists` appropriately.

### Nits

- `RedisService.cs:202-209` — the flat-array → dictionary conversion (and most of the Lua) is now duplicated between the two hash-get methods. A shared private helper would DRY it up and localize the null-forgiving usage. Inline comment left; not blocking.

### Test coverage

Good, and consistent with the `GetAndRefreshExpiry` tests from #2018: the three new integration tests cover the hit (fields returned + TTL refreshed in one round trip), the miss (null, no key created), and the non-hash self-heal. The two `NoOpCacheService` doubles are updated for the new interface member. The repository swap is behavior-preserving so no new repo-level test is needed.

### What's good

Tight, well-scoped tech-debt follow-up that does exactly what #2019 asked. The comments explaining *why* the write path and `PEXPIRE`-on-hit differ from the sibling read method are the right kind of context to leave behind.

_Note: backend/e2e CI was still in progress at review time — worth confirming green before merge._

---

ginderjeremiah:
## Re-review after `MapHashResult` extraction — Approve ✅ (merging)

The prior review approved with a single DRY nit (the flat-array→dictionary conversion duplicated between the two hash-get methods). Commit `12198e7` addresses exactly that: both `HashGetAllIfExists` and `HashGetAllAndRefreshExpiry` now delegate to a shared private `MapHashResult` helper, which is where the null-forgiving indexing is localized. I verified the extraction is behavior-preserving — both scripts still return either a Lua `false` (→ `RedisResult.IsNull` → `null`) or the flat `HGETALL` reply, and the helper handles both identically to the old inline code.

Verification recap:
- **Scope vs #2019** — fully met: single-round-trip `TYPE`/`HGETALL`+`PEXPIRE` Lua script, non-hash self-heal preserved, and integration coverage for hit (fields + TTL refresh), miss (null, no key created), and the non-hash self-heal case.
- **Write-path routing** — `PEXPIRE` makes this a write, so routing through `ObserveWrite` (like `GetAndRefreshExpiry`) rather than the `RedisCommandBudget.Read` path is correct.
- **Repository swap** — `GetCachedProgress` drops the separate `ExpireAndForget`; no dead code left behind.
- **CI** — all checks green (backend, e2e chromium/firefox/webkit, verify-backend).

No outstanding nits or should-fixes. Merging.

---

Issue #2101, opened 2026-07-17
TypedDamageDealt (offense book) can also go negative under enemy-side absorption, offsetting proficiency accrual and kill-type attribution


ginderjeremiah:
## Context

Found while implementing #2091, which floors the "avoided damage" tallies (`DamageDodged`, `DamageParried`, `PlayerCounterDamageDealt`) at 0 per portion so an absorbed (resistance > 1) portion doesn't decrement them. The same root cause reaches a fourth, architecturally distinct booking that #2091 deliberately didn't touch.

## Problem

`ResolvePlayerHit`'s per-portion loop (`Game.Core/Battle/BattleContext.cs:340-342`) computes `booked = Battler.HealthRemoved(net, healthBefore)` and books it unconditionally: `Stats.AddTypedDamageDealt(type, booked)`. `HealthRemoved` (`Battler.cs:419-422`, `Math.Min(damage, Math.Max(0, healthBefore))`) passes a negative `net` straight through when the **target** (the enemy, for a player's own attack) has typed resistance above 1 — i.e. the player's hit absorbs into a heal for the enemy, and `TypedDamageDealt[type]` goes negative for that battle.

Two downstream consumers read `TypedDamageDealt` with no sign guard on the per-type entry:

- `ProficiencyAccrual.BuildActivities` → `FoldIntoActivityKeys(activityByKey, stats.TypedDamageDealt, ...)` (`Game.Core/Proficiencies/ProficiencyAccrual.cs:168`) folds the raw (possibly negative) per-type quantity into the shared `activityByKey` map with no guard — unlike the `AddEvent` overlay tallies just below it, which floor at `amount > 0`. Within a mixed multi-type battle this can silently offset a sibling type's genuine training total, the same class of bug #2091 fixes for Dodge/Riposte.
- `PlayerProgress.RecordKillByDamageType` (`Game.Core/Progress/PlayerProgress.cs:295-309`) picks the "majority-dealt" type via `OrderByDescending(kv => kv.Value)` to attribute a kill's `KillsByDamageType` challenge/statistic credit — a negative entry could skew which type is picked as the majority.

Like #2091, this is **content-gated** today: no authored enemy currently carries resistance > 1, so it hasn't manifested yet, but the mechanic is explicitly supported.

## Why it's a separate issue

[...]

Issue #1927, opened 2026-07-13
CriticalBonusDealt is booked without a positive-investment guard — a CriticalDamage debuff books negative (or −∞) crit training (latent, content-triggered)


ginderjeremiah:
`BattleContext` books `Stats.CriticalBonusDealt += totalBooked * Battler.NormalizeInvestment(critMultiplier - 1.0);` on every crit with no guard (`Game.Core/Battle/BattleContext.cs:382`). Every sibling overlay guards non-positive investments: Momentum (`:349-353`, `rampContribution > 0`), Cull (`:356-359`), Cadence (`:367-370`), and Hex/Sunder return 0 for `<= 0` inside `Battler.HexBonusForHit`/`SunderBonusForHit`.

`φ(a) = a/(1+a)` (`Battler.cs:203-206`) has a pole at `a = −1`: an opponent skill effect debuffing `CriticalDamage` to exactly `0` books `-Infinity` into `CriticalBonusDealt`, and any value in `(0, 1)` books a negative claim that erodes crit training legitimately accrued earlier in the same battle — violating the uniform "an absorbed/uninvested hit trains nothing" rule (#1481).

Blast radius is contained (backend-only tally; the accrual's `amount > 0` gate at `ProficiencyAccrual.cs:172-177` keeps `-∞` out of the XP math) — the symptom is a silently zeroed/reduced per-battle Crit activity. Latent until a `CriticalDamage` debuff is authored, which the Workbench allows at any time.

**Fix direction:** book only when `critMultiplier > 1`, matching the other overlays; add the matching frontend/backend test rows if this tally is ever mirrored.

<sub>From the scheduled codebase health review.</sub>

PR #2102, opened 2026-07-17
Floor avoided/counter damage tallies at 0 under authored absorption (#2091)

### LINKED ISSUE #2091: Dodge/parry/counter "avoided damage" statistics book negative amounts under authored absorption, decrementing lifetime Sum stats
## Problem

`ComputeNetDamage` deliberately returns a **negative** net when a type's summed resistance exceeds 1 (absorption — a supported authored mechanic). Three tallies consume it without guarding that sign:

- `BattleContext.DamageParried` (`Game.Core/Battle/BattleContext.cs:244`)
- `BattleContext.DamageDodged` (`Game.Core/Battle/BattleContext.cs:258`)
- `PlayerCounterDamageDealt` (`Game.Core/Battle/BattleContext.cs:431`) when a riposte is absorbed by the enemy

At battle completion these feed `PlayerProgress.Record(...)` with **Sum** aggregation (`Game.Core/Progress/PlayerProgress.cs:69-72`), so:

- a player wearing authored absorption gear (e.g. `FireResistance > 1`) who dodges/parries a fire hit has their **lifetime `DamageDodged`/`DamageParried` statistic decrease**;
- an "AtLeast" challenge tracking those statistics has its `Progress` recomputed **downward**;
- within a mixed battle, negative portions silently offset the Evasion/Riposte proficiency training totals (`ProficiencyAccrual.cs:180,187` — the `amount > 0` guard only checks the battle total, not per-hit sign).

## Why it was missed

The sibling overlay tallies were explicitly hardened against exactly this ("an absorbed portion trains nothing" — #1927, #1482); the avoided-damage and counter tallies were not.

## Fix direction

Clamp the per-portion avoided/counter bookings at 0 (matching the #1482 policy that absorption enables no training/statistics), and add battle + parity-adjacent unit coverage for the absorption case. Note this is content-gated today — it only manifests once absorption content (`resistance > 1`) is authored — but the mechanic is explicitly supported, so the tallies should be correct before that content lands.

ginderjeremiah:
## Summary

`ComputeNetDamage` deliberately returns a **negative** net when a damage type's summed resistance exceeds 1 (authored absorption — a supported mechanic). Three battle-stat tallies in `BattleContext.cs` consumed that value without guarding its sign:

- `Stats.DamageParried` (parry-avoided-damage loop)
- `Stats.DamageDodged` (dodge-avoided-damage loop)
- `Stats.PlayerCounterDamageDealt` (the parry riposte's booked damage)

At battle completion these feed `PlayerProgress.Record(...)` with **Sum** aggregation, so an absorbed dodge/parry/counter would silently **decrement** the player's lifetime `DamageDodged`/`DamageParried` statistics, recompute an "AtLeast" challenge's progress downward, and (within a mixed-type battle) offset the Evasion/Riposte proficiency training totals for other, non-absorbed hits in the same battle (`ProficiencyAccrual`'s `amount > 0` guard only checks the aggregated battle total, not per-hit sign).

This mirrors the policy the sibling overlay tallies (Hex, Sunder, Momentum, Cull, Crit) already enforce per #1927/#1482 — "an absorbed portion trains nothing" — which these three tallies hadn't adopted.

## Fix

Floor each portion's avoided/countered net at **0** before booking, rather than letting a negative (absorption) portion pull the running total down:

- `DamageTarget`'s parry and dodge loops now wrap `ComputeNetDamage(...)` in `Math.Max(0, ...)` per portion.
- `ResolvePlayerHit` (the shared player-fire pipeline the parry riposte routes through) now also returns a `totalNetFloored` out-value — the same per-portion nets as its normal return, but each floored at 0 before summing. It deliberately does **not** apply the offense book's health-removed overkill cap, so a counter that kills the enemy still books its full gross damage (unchanged from today), while an absorbed portion contributes nothing instead of going negative.
- `FireParryCounter` books `PlayerCounterDamageDealt` from that floored value instead of the raw (potentially negative) total.

The flooring happens per-portion (not on the summed total), so a multi-type hit where one portion is absorbed and another isn't still credits the non-absorbed portion in full — covered by a dedicated test.

## Scope note

This is content-gated today — no authored enemy or gear currently carries resistance above 1 — but the mechanic is explicitly supported, so the fix lands ahead of that content per the issue's own framing.

While implementing this I found a related but architecturally distinct gap: the offense book (`TypedDamageDealt`, fed by a *different* clamp — `Battler.HealthRemoved`'s overkill cap) can also go negative under enemy-side absorption, reaching `ProficiencyAccrual`'s activity folding (no sign guard there either) and the kill-by-damage-type majority-type pick. That needs its own design pass (whether the floor belongs in `HealthRemoved` itself, which has other callers), so I filed it separately as #2101 rather than folding it into this PR.

## Test plan

- [x] Added `BattleContextTests`: dodge/parry/counter each get a dedicated absorption-floors-at-zero test, plus a multi-portion test proving the floor applies **per portion** (not to the summed total — a case where the two approaches would disagree).
- [x] Verified the new tests fail against the pre-fix code (4 failures) and pass after the fix (all green).
- [x] `dotnet test Game.sln` — full suite green (Core 1371, Infrastructure 66, Application 991, Api 821).
- [x] No frontend changes — these are server-only progression/statistics tallies (dodge/parry/riposte rolls aren't part of the frontend's battle simulation), not part of the tick-by-tick frontend/backend parity surface.

Closes #2091

---
_Generated by [Claude Code](https://claude.ai/code/session_019XoMuuntnUKYVqiShyLnAY)_

### DIFF
--- a/Game.Core/Battle/BattleContext.cs
+++ b/Game.Core/Battle/BattleContext.cs
@@ -218,7 +218,7 @@ namespace Game.Core.Battle
                 // scaled by the active battler's CriticalChanceMultiplier (base 1, so an uncommitted skill still
                 // crits at its own authored rate and further investment scales it). A single crit multiplies
                 // every portion (× 1.0 when it misses is exact, so a non-crit is unchanged).
-                totalNet = ResolvePlayerHit(rawDamage, portions, totalWeight, baseCriticalChance, _rng.Next());
+                totalNet = ResolvePlayerHit(rawDamage, portions, totalWeight, baseCriticalChance, _rng.Next(), out _);
             }
             else
             {
@@ -236,12 +236,14 @@ namespace Game.Core.Battle
                 {
                     // A parry negates the whole hit exactly like a dodge — the avoided damage is each portion's
                     // net (resistance then the Toughness curve) computed without mutating health, and no exposure
-                    // is recorded — and then ripostes with the defender's counter skill.
+                    // is recorded — and then ripostes with the defender's counter skill. A portion an authored
+                    // absorption (resistance > 1) would have healed avoided no damage, so it is floored at 0
+                    // rather than decrementing the tally (#2091).
                     Stats.AttacksParried++;
                     for (var i = 0; i < portions.Count; i++)
                     {
                         var dealt = AmplifiedPortion(rawDamage, totalWeight, portions[i]);
-                        Stats.DamageParried += _targetBattler.ComputeNetDamage(dealt, portions[i].Type);
+                        Stats.DamageParried += Math.Max(0, _targetBattler.ComputeNetDamage(dealt, portions[i].Type));
                     }
 
                     FireParryCounter(counterCritDraw);
@@ -250,12 +252,13 @@ namespace Game.Core.Battle
                 {
                     // A dodge zeroes the whole hit; the avoided damage is the sum of each portion's net (resistance
                     // then the Toughness curve), computed without mutating health. A dodge evaded the hit
-                    // entirely, so no exposure is recorded (it trains evasion instead).
+                    // entirely, so no exposure is recorded (it trains evasion instead). Floored per portion at 0
+                    // for the same absorption reason as the parry tally above (#2091).
                     Stats.AttacksDodged++;
                     for (var i = 0; i < portions.Count; i++)
                     {
                         var dealt = AmplifiedPortion(rawDamage, totalWeight, portions[i]);
-                        Stats.DamageDodged += _targetBattler.ComputeNetDamage(dealt, portions[i].Type);
+                        Stats.DamageDodged += Math.Max(0, _targetBattler.ComputeNetDamage(dealt, portions[i].Type));
                     }
                 }
                 else
@@ -290,9 +293,16 @@ namespace Game.Core.Battle
         /// with its crit draw supplied by the caller (the enemy fire's third unconditional draw) rather than
         /// drawn here. Requires the player to be the active battler.
         /// </summary>
+        /// <param name="totalNetFloored">
+        /// The same per-portion nets as the return value, but each portion floored at 0 before summing (#2091)
+        /// — unlike <paramref name="rawDamage"/>'s uncapped return, an absorbed (healed) portion contributes
+        /// nothing rather than pulling the total negative, while a killing portion's overkill is still kept (it
+        /// deliberately does <b>not</b> apply <see cref="Battler.HealthRemoved"/>'s health cap, unlike the typed
+        /// offense book). Consumed by <see cref="FireParryCounter"/> for <see cref="BattleStats.PlayerCounterDamageDealt"/>.
+        /// </param>
         private double ResolvePlayerHit(
             double rawDamage, IReadOnlyList<SkillDamagePortion> portions, double totalWeight,
-            double baseCriticalChance, double critDraw)
+            double baseCriticalChance, double critDraw, out double totalNetFloored)
         {
             var isCrit = critDraw < baseCriticalChance * _activeBattler.GetAttributeValue(CriticalChanceMultiplier);
             var critMultiplier = isCrit ? _activeBattler.GetAttributeValue(CriticalDamage) : 1.0;
@@ -317,6 +327,7 @@ namespace Game.Core.Battle
 
             var totalNet = 0.0;
             var totalBooked = 0.0;
+            var netFloored = 0.0;
             for (var i = 0; i < portions.Count; i++)
             {
                 var type = portions[i].Type;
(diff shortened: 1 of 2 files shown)


Game.Core/Battle/BattleContext.cs
```diff
@@ -236,12 +236,14 @@ public double DamageTarget(double rawDamage, IReadOnlyList<SkillDamagePortion> p
                 {
                     // A parry negates the whole hit exactly like a dodge — the avoided damage is each portion's
                     // net (resistance then the Toughness curve) computed without mutating health, and no exposure
-                    // is recorded — and then ripostes with the defender's counter skill.
+                    // is recorded — and then ripostes with the defender's counter skill. A portion an authored
+                    // absorption (resistance > 1) would have healed avoided no damage, so it is floored at 0
```
ginderjeremiah: Nit: this reads as ungrammatical — "A portion an authored absorption ... would have healed avoided no damage" is missing the relative pronoun. Suggest "A portion that an authored absorption (resistance > 1) would have healed avoided no real damage, so it is floored at 0 rather than decrementing the tally (#2091)." Comment-only; no functional impact.

---

Game.Core/Battle/BattleContext.cs
```diff
@@ -290,9 +293,16 @@ public double DamageTarget(double rawDamage, IReadOnlyList<SkillDamagePortion> p
         /// with its crit draw supplied by the caller (the enemy fire's third unconditional draw) rather than
         /// drawn here. Requires the player to be the active battler.
         /// </summary>
+        /// <param name="totalNetFloored">
+        /// The same per-portion nets as the return value, but each portion floored at 0 before summing (#2091)
+        /// — unlike <paramref name="rawDamage"/>'s uncapped return, an absorbed (healed) portion contributes
```
ginderjeremiah: Nit: `<paramref name="rawDamage"/>` is the wrong reference here. The contrast you're drawing is against this method's **return value** (`totalNet`), which is uncapped and can go negative — `rawDamage` is an input parameter and has no "return". Consider rewording to "unlike the (uncapped) return value" or dropping the `paramref`. Doc-only.

---

ginderjeremiah (commented):
## Review: Floor avoided/counter damage tallies at 0 under authored absorption (#2091)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating event) — the fix is correct, well-scoped, and matches the issue's stated direction. Only two trivial doc/comment nits, no blockers.

The change correctly addresses the root cause: `ComputeNetDamage` returns a negative net under authored absorption (resistance > 1), and the three Sum-aggregated tallies (`DamageParried`, `DamageDodged`, `PlayerCounterDamageDealt`) previously consumed it unguarded, so an absorbed dodge/parry/counter would decrement a lifetime statistic, drag an "AtLeast" challenge downward, and offset sibling proficiency training within a mixed battle. Flooring each portion at 0 aligns these tallies with the policy the overlay tallies already enforce (#1927/#1482).

### Design notes (all good)
- **Per-portion flooring, not on the summed total.** `Math.Max(0, ComputeNetDamage(...))` inside the loop (and `netFloored += Math.Max(0, net)` in `ResolvePlayerHit`) is the correct choice — a multi-type hit where one portion is absorbed and another isn't still credits the non-absorbed portion in full. The `DamageTarget_MultiPortion_DodgeWithAbsorption_FloorsEachPortionIndividually` test is exactly the case where per-portion and per-total flooring disagree, so it genuinely pins the behavior.
- **`totalNetFloored` deliberately skips `HealthRemoved`'s overkill cap.** Keeping a killing counter's full gross while flooring only absorption is consistent with the existing `PlayerCounterDamageDealt` semantics (identical to the old `counterNet` for all non-absorbed cases), so this is behavior-preserving except for the bug being fixed. The `out` parameter is a reasonable way to return both the raw net (for reflection) and the floored net (for the tally) from one pipeline pass.
- **Follow-up split is the right call.** The offense-book gap (`TypedDamageDealt` going negative via `HealthRemoved`, reaching `ProficiencyAccrual` folding and kill-type attribution) is a genuinely distinct concern — it flows through a different clamp and touching it means deciding whether the floor belongs in `HealthRemoved` itself (which has other callers, e.g. DoT booking). Filing #2101 rather than folding it in is correct scope discipline.

### Nits (non-blocking — flagged inline)
- `BattleContext.cs:239-240` — the parry comment is missing a relative pronoun ("A portion **an** authored absorption ... would have healed avoided no damage").
- `BattleContext.cs:298` — the `<paramref name="rawDamage"/>` in the new `<param>` doc should reference the method's return value (`totalNet`), not the `rawDamage` input.

### Test coverage
Strong. Dedicated absorption-floors-at-zero tests for dodge, parry, and counter, plus the per-portion multi-type test that distinguishes the two flooring strategies. The counter test's health-room setup (a prior physical hit opening heal room so absorption genuinely applies) and its sanity assertion on enemy health are a nice touch — they prove the absorption actually happened rather than the tally being zero for an unrelated reason. Correctly no frontend mirror, since dodge/parry/riposte rolls are backend-only tallies and not part of the tick-by-tick parity surface.

### What's good
Tight, surgical fix that follows an established project policy, with tests that fail against the pre-fix code and a clean scope boundary drawn around the related-but-distinct offense-book gap.

---

PR #2104, opened 2026-07-17
Close the RegisterSocket switch-credit-claim race (#2094)

### LINKED ISSUE #2094: RegisterSocket can kick an in-flight switch-credit claim, reopening the #2041 lost-update race in a one-round-trip window
## Problem

The #2041 defense promises a strict dichotomy: a registering socket either finds the switch-credit claim and defers, or the credit finds a live socket and skips. But `RegisterSocket` (`Game.Api/Services/SocketManagerService.cs:82-91`) implements the deferral as a **non-atomic** poll (`AwaitSwitchCreditClaimClear`, `:218-225`) followed by an unconditional `GetSet`.

If `TryClaimForSwitchCredit`'s `CompareAndSet(null → claim)` (`:196-199`) lands between the poll's last read and the `GetSet`, the registration **kicks the claim** (the code at `:86-91` recognizes this interleaving but only handles the "no old socket to notify" aspect) and starts the battle loop while `CreditDepartedCharacter`'s `SimulateSwitchProgress` (`LoginController.cs:213-240`) is still mid read-modify-write on the same player — the exact HTTP-vs-battle-loop lost-update race the claim exists to prevent (double-crediting the away window / clobbering `LastActivity`; `BattleAlreadyCredited` only protects the in-flight battle, not the idle-window credit).

A second organic path to the same race: presence is TTL-based and best-effort (`RefreshSocketPresence` is fire-and-forget), so a live socket whose key lapsed (~3 missed heartbeats) leaves `TryClaimForSwitchCredit` free to succeed while that socket's loop keeps saving.

## Fix direction

Claim the presence key with a `CompareAndSet` loop that re-enters the deferral wait when it observes the claim sentinel, instead of `GetSet`-kicking it; only the TTL deadline should ever kick a claim.

## Ride-along test gap

The switch-credit claim release on a **faulted** credit (`CreditDepartedCharacter`'s `finally`) is untested — the success-path release is pinned (`SwitchPlayer_CreditsDepartedCharacter_ReleasesTheSwitchCreditClaimAfterward`) but a throw from `SimulateSwitchProgress` is not. Worth covering in the same change.

Probability is low (needs multi-device timing or a misbehaving client), but it's a documented-guarantee violation on the path built specifically to close it — see the contract at `docs/backend-sockets.md:15`.

ginderjeremiah:
## Summary

`RegisterSocket` deferred behind an in-flight switch-away credit's claim (#2041) via a poll (`AwaitSwitchCreditClaimClear`) followed by an unconditional `GetSet`. That left a gap between "the claim looked clear" and "the key was actually written": a switch-credit claim (`TryClaimForSwitchCredit`'s atomic `CompareAndSet(null → claim)`) landing in that gap got silently kicked by the `GetSet`, reopening the exact HTTP-vs-battle-loop lost-update race the claim exists to prevent — the registration would start the battle loop while `CreditDepartedCharacter`'s `SimulateSwitchProgress` is still mid read-modify-write on the same player.

- Replaces the wait-then-`GetSet` with `ClaimPresenceKey`, a `CompareAndSet` loop: the write only lands if the key's value is still what was just read, so a claim (or a competing registration) that lands in the gap fails the write and the loop re-reads and retries against whatever is actually there, deferring again while it observes the claim sentinel. Only the claim's own TTL deadline elapsing lets an attempt targeting the (by then stale) sentinel proceed — matching the issue's suggested fix direction.
- Covers the ride-along test gap called out in the issue: the switch-credit claim's release on a **faulted** credit (`CreditDepartedCharacter`'s `finally`) was previously untested — only the success-path release was pinned. Added `SwitchPlayer_CreditFaults_StillReleasesTheSwitchCreditClaim`.

## How it addresses the issue

- `RegisterSocket_SwitchCreditClaimLandsBetweenReadAndWrite_RetriesRatherThanKickingIt` (new unit test) deterministically scripts the exact race window via a fake `ICacheService`: the first `CompareAndSet` attempt is raced out (simulating a claim landing in the gap), the caller re-reads and observes the sentinel, genuinely polls/defers (asserted via timing), then succeeds once the sentinel clears. The fake's `GetSet` throws, so any regression back to an unconditional overwrite fails the test outright.
- `SwitchPlayer_CreditFaults_StillReleasesTheSwitchCreditClaim` (new integration test) builds a `LoginController` with a decorated `OfflineProgressService` whose progress save always faults, confirming the switch-credit claim is still released via `CreditDepartedCharacter`'s `finally` block even when the credit itself throws.
- Existing `RegisterSocket_SwitchCreditClaimActive_DefersUntilClaimReleases` and `SwitchPlayer_CreditsDepartedCharacter_ReleasesTheSwitchCreditClaimAfterward` integration tests continue to pass, covering the non-racing defer/release paths.

No production callers of `ICacheService.GetSet` remain after this change; the interface member and its `RedisService`/integration-test coverage are left in place as a general-purpose cache primitive (documented, still exercised by its own `RedisServiceIntegrationTests`), consistent with how the rest of `ICacheService`'s surface is treated.

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test Game.sln` — all 3251 backend tests pass (Game.Core.Tests, Game.Infrastructure.Tests, Game.Api.Tests, Game.Application.Tests)
- [x] Targeted runs of `SocketManagerServiceTests`, `SocketCommandProcessorTests`, and `LoginControllerTests` including the new tests

Closes #2094

---
_Generated by [Claude Code](https://claude.ai/code/session_014Rm3WBsBVsgSMeaE67UCEX)_

### DIFF
--- a/Game.Api/Services/SocketManagerService.cs
+++ b/Game.Api/Services/SocketManagerService.cs
@@ -78,8 +78,8 @@ namespace Game.Api.Services
             // credit is a read-modify-write against this same player's aggregate run off this player's battle
             // loop, so registering (and starting the loop) while it's still mid-flight would reintroduce the
             // lost-update race the presence claim exists to prevent. ClaimPresenceKey's CompareAndSet loop
-            // closes the gap a separate wait-then-GetSet would leave between "the claim looked clear" and
-            // "the key was written" — see its doc comment.
+            // closes the gap a separate wait-then-unconditional-write would leave between "the claim looked
+            // clear" and "the key was written" — see its doc comment.
             var oldSocketId = await ClaimPresenceKey(presenceKey, socketContext.SocketId);
 
             try
@@ -206,12 +206,12 @@ namespace Game.Api.Services
         /// key and writing it can never be kicked: the write only lands if the key's value is still what was
         /// just read, so a claim (or another registration) that lands first makes the CompareAndSet fail, and
         /// the loop re-reads and retries against whatever is there now instead of blindly overwriting it — the
-        /// gap a separate wait-then-<see cref="ICacheService.GetSet"/> leaves open. While the read value is the
-        /// claim sentinel, the loop defers (polling at <see cref="SwitchCreditWaitPollInterval"/>) rather than
-        /// racing it, capped at <see cref="SwitchCreditClaimTtl"/> — the claim's own TTL — so a credit that
-        /// faults without releasing can never wedge a registration past that bound; only once that deadline
-        /// passes does an attempt targeting the (by then stale) sentinel proceed. Returns the previous real
-        /// socket id, or <see langword="null"/> if the key was unset or held only a switch-credit claim.
+        /// gap a separate wait-then-unconditional-write would leave open. While the read value is the claim
+        /// sentinel, the loop defers (polling at <see cref="SwitchCreditWaitPollInterval"/>) rather than racing
+        /// it, capped at <see cref="SwitchCreditClaimTtl"/> — the claim's own TTL — so a credit that faults
+        /// without releasing can never wedge a registration past that bound; only once that deadline passes
+        /// does an attempt targeting the (by then stale) sentinel proceed. Returns the previous real socket id,
+        /// or <see langword="null"/> if the key was unset or held only a switch-credit claim.
         /// </summary>
         private async Task<string?> ClaimPresenceKey(string presenceKey, string newSocketId)
         {
@@ -232,7 +232,10 @@ namespace Game.Api.Services
                 }
 
                 // Another writer (a fresh switch-credit claim, a competing registration) won the race between
-                // our read and this write; re-read whatever landed and retry against that instead.
+                // our read and this write; re-read whatever landed and retry against that instead. Unbounded by
+                // design but self-limiting in practice: each failure here means one more concurrent writer for
+                // this single player's key, a set that's small and finite rather than something that can churn
+                // forever.
                 currentValue = await _cache.Get(presenceKey);
             }
         }
--- a/Game.Api/Services/SocketManagerService.cs
+++ b/Game.Api/Services/SocketManagerService.cs
@@ -74,21 +74,13 @@ namespace Game.Api.Services
             var socketContext = new SocketContext(socket, playerId, sessionService, isAdmin, _loggerFactory.CreateLogger<SocketContext>());
             var socketHandler = new SocketHandler(socketContext, _commandFactory, _scopeFactory, _loggerFactory.CreateLogger<SocketHandler>(), () => RefreshSocketPresence(playerId, socketContext.SocketId));
             var presenceKey = CurrentSocketKey(playerId);
-            // Defer behind an in-flight switch-away credit's claim (#2041) rather than blindly overwriting it:
-            // that credit is a read-modify-write against this same player's aggregate run off this player's
-            // battle loop, so registering (and starting the loop) while it's still mid-flight would reintroduce
-            // the lost-update race the presence claim exists to prevent. Bounded by the claim's own TTL, so a
-            // claim that never releases can never wedge this connection past it.
-            await AwaitSwitchCreditClaimClear(presenceKey);
-            // Claim the presence key with its TTL atomically so a fault here can never leave the key without an
-            // expiry — a TTL-less key would defeat the heartbeat design and report a permanent ghost session.
-            var oldSocketId = await _cache.GetSet(presenceKey, socketContext.SocketId, SocketPresenceTtl);
-            if (oldSocketId == SwitchCreditClaimValue)
-            {
-                // Either the wait above timed out or a credit claimed the key in the gap right before this
-                // GetSet — either way there is no real prior connection to notify, just a claim we kicked.
-                oldSocketId = null;
-            }
+            // Defer behind an in-flight switch-away credit's claim (#2041) rather than overwriting it: that
+            // credit is a read-modify-write against this same player's aggregate run off this player's battle
+            // loop, so registering (and starting the loop) while it's still mid-flight would reintroduce the
+            // lost-update race the presence claim exists to prevent. ClaimPresenceKey's CompareAndSet loop
+            // closes the gap a separate wait-then-GetSet would leave between "the claim looked clear" and
+            // "the key was written" — see its doc comment.
+            var oldSocketId = await ClaimPresenceKey(presenceKey, socketContext.SocketId);
 
             try
             {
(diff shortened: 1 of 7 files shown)


Game.Api/Services/SocketManagerService.cs
```diff
@@ -209,18 +201,39 @@ public async Task ReleaseSwitchCreditClaim(int playerId)
...
+                }
+
+                if (await _cache.CompareAndSet(presenceKey, currentValue, newSocketId, SocketPresenceTtl))
+                {
+                    return currentValue == SwitchCreditClaimValue ? null : currentValue;
+                }
+
+                // Another writer (a fresh switch-credit claim, a competing registration) won the race between
+                // our read and this write; re-read whatever landed and retry against that instead.
+                currentValue = await _cache.Get(presenceKey);
```
ginderjeremiah, line 239: Nit (optional): the sentinel-wait branch is bounded by `deadline`, but the CompareAndSet-failure retry path (this re-read + loop) has no deadline or iteration cap. In practice it terminates — each CAS failure means another writer won, and concurrent presence writers per player are few — so this is self-limiting rather than a real livelock. But a future reader has to reason that out to convince themselves the `while (true)` halts. Worth a one-line comment noting the retry is bounded by the (small, finite) set of competing writers, so nobody later "fixes" a perceived infinite loop.

---

ginderjeremiah (commented):
## Review: Close the RegisterSocket switch-credit-claim race (#2094)

**Intended verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't take a gating review from me)* — the fix is correct, well-scoped, and well-tested. One should-fix I'd resolve before merge, plus one optional nit.

I traced the core change against #2094 and the #2041 contract in `docs/backend-sockets.md`, and it does exactly what the issue asked for. `ClaimPresenceKey`'s `CompareAndSet` loop closes the read→write gap the old `AwaitSwitchCreditClaimClear` + unconditional `GetSet` left open: a claim (or a competing registration) landing in that gap now fails the CAS and forces a re-read/retry against whatever actually landed, so a claim is only ever kicked once its own TTL deadline has passed — the exact behavior the issue prescribed. I walked the state transitions (unclaimed, sentinel-present, sentinel-cleared-mid-wait, claim-released-at-deadline, live-socket-takeover) and they all resolve correctly, and `null`-vs-real-socket-id return semantics are preserved so the SocketReplaced notify still only fires for a genuine prior connection.

### Should-fix

- **`ICacheService.GetSet` is now orphaned production code — remove it here.** After this change `GetSet` has zero production callers (`grep '\.GetSet('` finds only `RedisServiceIntegrationTests` and the `<see cref>` in `ClaimPresenceKey`'s doc comment). The PR body justifies keeping it as "a general-purpose primitive, consistent with how the rest of `ICacheService`'s surface is treated" — but that isn't accurate: every other member I spot-checked (`GetDelete`, `GetAndRefreshExpiry`, `SetAndForget`, `HashSetIfExistsAndForget`) has real repository/service callers. `GetSet` is the lone exception. CLAUDE.md is explicit — "If you are removing or replacing functionality, make sure to remove any code that is no longer being used as well" — and this PR is precisely what orphaned it, so the removal (interface member + `RedisService` impl + its dedicated `RedisServiceIntegrationTests` coverage) belongs in this change rather than a follow-up. If there's a concrete reason to retain it as a deliberate primitive, that's a defensible call — but then say so as the rationale, not "consistent with the rest of the surface," which it isn't.

### Nits

- `ClaimPresenceKey` — the CAS-failure retry path is unbounded (see inline comment). Correct and self-limiting in practice; just worth a one-line note so nobody later mistakes the `while (true)` for an unbounded loop.

### Test coverage

Strong and matched to the issue's asks:
- `RegisterSocket_SwitchCreditClaimLandsBetweenReadAndWrite_RetriesRatherThanKickingIt` deterministically scripts the exact race window via `RacingClaimCacheService` — first CAS raced out, sentinel then observed, retry lands once it clears — and pins that it went through CAS (three `Get`s, `[null, null]` expected-values) with `GetSet` throwing so any regression to an unconditional overwrite fails outright. The `>= 40ms` wall-clock assertion is safe (a 50ms `Task.Delay` never returns early) and is the cleanest way to prove it genuinely deferred rather than hot-spun.
- `SwitchPlayer_CreditFaults_StillReleasesTheSwitchCreditClaim` covers the ride-along gap — the `finally`-block release on a *faulted* `SimulateSwitchProgress` — via a `SaveThrowsPlayerProgressRepository` decorator. The by-hand `LoginController`/`OfflineProgressService` wiring is verbose and a touch brittle to constructor changes, but it's justified for injecting the fault at the right seam; no change needed.
- No frontend mirror is required — this is socket-infrastructure concurrency, not battle logic.

### What's good

- The fix lands at exactly the right altitude: a small, contained change to the claim primitive rather than reshuffling the credit/registration flow, and the domain layering is untouched (this is all API-tier socket lifecycle).
- The doc comment on `ClaimPresenceKey` is thorough where the concurrency reasoning genuinely warrants it, and correctly frames the TTL-deadline-kick as the one intended exception.
- The regression fake is designed to *fail loudly* on the specific regression it guards against (`GetSet` throws), which is the right instinct for a race-condition test.

I'm not merging: CI is still running (backend tests in progress) and I'd like the `GetSet` removal addressed in-PR first.

---

ginderjeremiah:
## Re-review (delta since prior review) — Approve ✅

Re-reviewed the changes made after the prior review (commit `d52849b`), which addressed both outstanding items:

**Should-fix (orphaned `GetSet`) — resolved.** The now-dead `ICacheService.GetSet` is fully removed: the interface member, `RedisService`'s Lua-script implementation, and its dedicated `RedisServiceIntegrationTests` coverage, with stale doc/comment references updated. I grepped the tree for `GetSet` — the only remaining hits are unrelated (`StringGetSetExpiryAsync`, StackExchange's native GETEX, in `GetAndRefreshExpiry`, and a historical "since-removed GetSet primitive" comment in the unit test). No orphaned references, and this satisfies CLAUDE.md's no-dead-code guidance in-PR rather than deferring it.

**Nit (unbounded CAS-retry loop) — resolved.** `ClaimPresenceKey` now carries the one-line note explaining the `while (true)` retry is unbounded-by-design but self-limiting (bounded by the small, finite set of concurrent presence writers for a single player's key), so a future reader won't mistake it for a livelock.

I re-walked the final `ClaimPresenceKey` state machine (unclaimed → CAS lands; sentinel-before-deadline → defers and polls; sentinel-past-deadline → CAS kicks the stale sentinel; CAS raced out → re-read + retry against what actually landed; live-socket takeover → returns the real prior id). All transitions resolve correctly, and the `null`-vs-real-socket-id return contract is preserved so `SocketReplaced` still only fires for a genuine prior connection.

**CI:** green — `Run tests` #2572 on `d52849b` concluded success.

No outstanding nits or should-fixes, and no follow-up issues needed. Merging.

---

PR #2105, opened 2026-07-17
Write the player cache blob after a dispatch fault, not just a flush success

### LINKED ISSUE #2098: Dispatch-fault save path durably persists insert-only unlock events while rolling the cache back — phantom unlocks materialize on a later DB fall-through
## Problem

When a non-publisher event handler faults during `SavePlayer`'s dispatch (the #1819 path, `Game.DataAccess/Repositories/PlayerRepository.cs:119-171`): the fault is captured (`:126-129`), the flush **still enqueues everything the persistence publisher buffered** (`:140`) — including the failed command's `ItemUnlockedEvent`/`SkillUnlockedEvent` — then the wrapped rethrow (`:155-161`) skips the cache-blob write (`:171`) and `SocketHandler` reloads the player from the **stale cached blob**.

Absolute-value events (`PlayerCoreUpdated`, `ProgressUpdated`) converge on the next save. But **insert-only** unlock events (`ItemUnlockedHandler`, `SkillUnlockedHandler`, `ModUnlockedHandler`, `LessonUnlockedHandler` — e.g. `Game.DataAccess/PlayerUpdates/Handlers/ItemUnlockedHandler.cs:9-18`) have no compensating write: Postgres durably holds a row the rolled-back aggregate and cache don't know about, and every subsequent save re-serializes the itemless blob, so the divergence persists for the key's whole lifetime.

The phantom item/skill/lesson then materializes whenever a load finally falls through to the DB (idle-TTL expiry, eviction, corrupt-blob self-heal) — a command the client saw **fail** granting its rewards days later. This violates backend-persistence.md's "at worst briefly behind, never wrong" claim. The same window exists on the #1849 cancellation path (flush landed server-side, cancellation unwound before the cache write).

## Fix direction

On the dispatch-fault path, either exclude the faulted save's envelopes from the flush, or write the cache blob anyway — the events *were* enqueued, so the mutated aggregate (not the prior blob) is what actually matches the queue. Cover with an integration test in the handler idempotency/race matrix.

ginderjeremiah:
## Summary

`PlayerRepository.SavePlayer` skipped the player cache blob write whenever the in-process domain-event dispatch faulted (`dispatchFault is not null`), even though the flush right above it had already succeeded. That skip was based on a premise that no longer holds: `DomainEventDispatcher` isolates each handler and always finishes its **full** event sweep — buffering every handler's write into that same flush — before surfacing a collected failure (see `docs/backend-persistence.md` → "Domain-event dispatch"). So by the time `dispatchFault` is observed, the flush has already durably enqueued everything matching the mutated aggregate; skipping the cache write just left the cache silently behind it.

For absolute-value events (`PlayerCoreUpdated`, `ProgressUpdated`) that only meant a transient revert until the next save converged. But insert-only unlock events (`ItemUnlockedEvent`, `SkillUnlockedEvent`, `ModUnlockedEvent`, `LessonUnlockedEvent`) have no compensating overwrite — Postgres durably held a row the rolled-back cache/aggregate never knew about, and every later save kept re-serializing the itemless blob. The phantom unlock only surfaced whenever a load finally fell through to the DB (idle-TTL expiry, eviction, corrupt-blob self-heal), days after the command that granted it appeared to fail to the client.

## How it addresses the issue

- `PlayerRepository.SavePlayer` (`Game.DataAccess/Repositories/PlayerRepository.cs`) now writes the cache blob unconditionally once the flush succeeds, then rethrows the wrapped dispatch fault afterward — the fault itself still needs to propagate so the caller treats the command as failed and the socket layer's forced reload happens, but that reload now reads back this save's own mutations instead of reverting past them.
- Updated `docs/backend-persistence.md`'s write-behind section to correct the now-inaccurate rationale (previously implied skipping the cache write was needed to avoid "mutations riding along with no queued event to match" — that's no longer the actual risk given the isolate-and-collect dispatcher).

## Test plan

- [x] Updated `PlayerWriteBehindTests.SavePlayer_DispatchFaultsForANonCancellationReason_StillFlushesBufferedEnvelopesAndWritesTheCacheBlob_AndThrowsPlayerPersistenceFlushFailedException` (renamed from the old test that asserted the buggy skip-and-revert behavior) to assert a post-fault reread sees the mutation, not the pre-mutation state.
- [x] `dotnet test Game.sln` — all 3249 backend tests pass (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests)

Closes #2098

---
_Generated by [Claude Code](https://claude.ai/code/session_014KZjo5oqLTRkY32cVCEZNr)_


ginderjeremiah (commented):
## Review: cache-blob write after a dispatch fault (#2098)

**Verdict: Approve** — the fix is correct, minimal, and well-tested. Submitted as a comment only because GitHub won't let me formally approve my own PR; treat this as an approval. (Filed one follow-up before merging; see below.)

I verified the linchpin the whole change rests on, directly in `Game.Core/Events/DomainEventDispatcher.cs`: `DispatchAsync(AggregateRoot,…)` wraps every handler invocation in its own `try/catch` (`DispatchBatchAsync`, lines 90–98), drains the aggregate to quiescence, and only surfaces failures at the end via `ThrowIfAnyHandlerFailed`. So by the time `dispatchFault` is observed in `SavePlayer`, the flush has provably enqueued everything matching the mutated aggregate — writing the cache blob before rethrowing the wrapped fault is exactly right, and the old skip-and-revert was the bug for insert-only unlock events with no compensating overwrite. Docs, the renamed test, and the commit message all line up.

### Follow-up (not blocking — filed as #2106)

- #2098's Problem section explicitly calls out the **#1849 cancellation path** as having the same window ("flush landed server-side, cancellation unwound before the cache write"). This PR only closes the dispatch-fault path — a cancellation observed during `FlushAsync` (`PlayerRepository.cs` flush `catch`, guarded `when (ex is not OperationCanceledException)`) still propagates before the line-172 cache write. I did **not** fold this into the PR because it's genuinely a different problem: on cancellation the `LPUSH`'s landing is indeterminate, so #2105's unconditional-write approach isn't safe there and it needs a distinct mechanism. Tracked in #2106 so `Closes #2098` auto-closing the issue doesn't drop it.

### Nit (non-blocking)

- The line 164–171 comment says the cache write is safe "regardless of which handler faulted." That holds for every realistic fault (downstream I/O handlers — notifiers, `BattleStatisticsEventHandler`'s progress save), but is technically stronger than true if the persistence publisher's *own* `Serialize()` were the thing that threw, since that event's envelope would never have been buffered. `PlayerPersistencePublisher.Buffer` serialization is pure and deterministic, so this is a theoretical edge, not a transient — mentioning only for precision.

### Test coverage

Adequate. `SavePlayer_DispatchFaultsForANonCancellationReason_StillFlushesBufferedEnvelopesAndWritesTheCacheBlob_AndThrowsPlayerPersistenceFlushFailedException` now asserts a post-fault reread sees the `ChangeZone(1)` mutation (not the pre-mutation zone) and that the buffered envelope was still flushed — pinning the corrected behavior. The cancellation-path test belongs with #2106.

### What's good

Tight, well-scoped fix that removes a stale skip rather than adding compensating machinery, backed by re-verifying the dispatcher's isolate-and-collect contract. The comments and `backend-persistence.md` update correctly retire the now-inaccurate rationale instead of just layering on top of it.

---

PR #2107, opened 2026-07-17
Override live paths (not just proficiencies) in path retire-confirm (#2099)

### LINKED ISSUE #2099: Progression editor's path retire-confirm mixes live tiers with stale paths — the paths catalogue from #2031 is never overridden with store.paths
## Problem

#2031 (commit 4520fa9) made `pathReferences` consult `paths[p.pathId]?.retiredAt` to mirror the backend's `PathStaysLive` check, and `referenceSourcesFromStatic` fills that slot from `staticData.paths` — the **last-saved** catalogue (`UI/src/routes/admin/workbench/references.ts:160-169`, `retire-confirm.ts:12-27`).

But `PathDetail.svelte`'s `onRetire` (`UI/src/routes/admin/workbench/progression/PathDetail.svelte:97`) passes only `{ proficiencies: store.profs }` — live, unsaved tiers — while the same `ProgressionStore` also holds live, unsaved **paths** (`store.paths`, toggled by `retirePath()` without a save). The dialog therefore evaluates live tiers against stale paths.

## Failure scenarios

- Admin retires path B in-session (unsaved), then retires path A whose tier gates a B gateway → the confirm still shows the strong "re-point first" warning that #2031 exists to suppress (staticData still shows B live).
- Inverse: reinstating B in-session, then retiring A → the warning is **silently skipped**, and the save is then rejected by the backend guard the dialog was supposed to preview.

This is a missed sibling of the fix-wave's own pattern — 9e87c0d explicitly mirrors "the progression editor's `{ proficiencies: store.profs }` override", but the `paths` slot the same wave introduced was left static. Note this is *not* the tracked #1976 gap (generic Workbench cross-catalogue edits); here the live data is already held by the exact store performing the retire.

Impact is advisory-only (the backend guard still blocks any bad save), but the dialog exists to preview that guard accurately.

## Fix direction

Pass `{ proficiencies: store.profs, paths: store.paths }` as the override in `PathDetail.svelte`'s `onRetire`, plus a unit test for both scenarios above.

ginderjeremiah:
## Summary

`PathDetail.svelte`'s `onRetire` built the retire-confirm reference sources with only `{ proficiencies: store.profs }` as the live override. But `pathReferences` (in `references.ts`) also reads a gating path's *own* `retiredAt` via `paths[p.pathId]?.retiredAt` — and that slot was left on `staticData.paths`, the last-saved catalogue. So:

- Retiring path B in-session (unsaved), then retiring path A whose tier gates a B gateway → the confirm still showed the strong "re-point first" warning, even though #2031 exists to suppress exactly this case once B is (about to be) retired.
- The inverse — reinstating B in-session, then retiring A — silently skipped the warning, even though the backend guard would still reject the save.

## Fix

Pass `paths: denseByLiveId(store.paths)` alongside the existing `proficiencies: store.profs` override in `PathDetail`'s `onRetire`.

The dense-by-id rebuild matters here specifically for `paths`: `ProgressionStore.addPath` prepends unsaved new paths with negative ids, so passing `store.paths` straight through would misalign the array-index lookup `pathReferences` relies on (`paths[p.pathId]`) — the exact problem `denseByLiveId` already solves for `ownCatalogueOverride`'s enemies/skills case, just newly needed here since `paths` is now indexed rather than only ever `.filter()`-ed. `profs` doesn't need the same treatment — `pathReferences`/`proficiencyReferences` only ever `.filter()` over proficiencies, never index them by id. `denseByLiveId` is now exported from `retire-confirm.ts` for this reuse.

## Test plan

- [x] Added a unit test for each scenario from the issue (path retired live suppresses the warning; path reinstated live surfaces it) to `PathDetail.test.ts`.
- [x] Added a third test covering the dense-by-id rebuild itself (an unsaved new path with a negative id prepended to `store.paths` no longer shifts the gating path's index).
- [x] `npm run test:unit:ci` — full suite, 3762/3762 passing.

Closes #2099

---
_Generated by [Claude Code](https://claude.ai/code/session_019HJeDxknDLdJBkemPtYiC2)_

### DIFF
--- a/UI/src/routes/admin/workbench/progression/PathDetail.svelte
+++ b/UI/src/routes/admin/workbench/progression/PathDetail.svelte
@@ -68,7 +68,7 @@
 {/if}
 
 <script lang="ts">
-import { referenceSourcesFromStatic, retireWithConfirm } from '../retire-confirm';
+import { denseByLiveId, referenceSourcesFromStatic, retireWithConfirm } from '../retire-confirm';
 import type { ProgressionStore, PathTab } from './progression-store.svelte';
 import { activityKeyGroups, hasTierCollision, pathWarnings } from './progression-helpers';
 import type { WorkbenchPath } from './types';
@@ -87,6 +87,12 @@ const { store }: Props = $props();
  * Retire a path, first surfacing any live gateway one of its tiers would soft-lock — the same
  * check the backend guard (`AdminPaths.FindRetiredPathGatingLiveGateway`) rejects the save on.
  * Previously bypassed the confirm entirely (#1863), unlike every other retire path.
+ *
+ * Overrides both `proficiencies` and `paths` with this session's live (unsaved-edits-included)
+ * copies — `pathReferences` reads `store.paths`' own `retiredAt` to decide whether a gating path
+ * is still live (#2099), not just `store.profs` for the tiers themselves. `paths` needs the
+ * dense-by-id rebuild since `addPath` prepends unsaved paths with negative ids, unlike `profs`
+ * which `pathReferences`/`proficiencyReferences` only ever `.filter()`, never index by id.
  */
 const onRetire = (rec: WorkbenchPath) =>
 	retireWithConfirm({
@@ -94,7 +100,7 @@ const onRetire = (rec: WorkbenchPath) =>
 		id: rec.id,
 		name: rec.name || 'Unnamed path',
 		title: 'Retire path?',
-		sources: referenceSourcesFromStatic({ proficiencies: store.profs }),
+		sources: referenceSourcesFromStatic({ proficiencies: store.profs, paths: denseByLiveId(store.paths) }),
 		onConfirmed: () => store.retirePath(rec.id, true)
 	});
 
--- a/UI/src/routes/admin/workbench/retire-confirm.ts
+++ b/UI/src/routes/admin/workbench/retire-confirm.ts
@@ -39,12 +39,13 @@ const SELF_REFERENCING_KEYS = new Set(['enemies', 'skills']);
 /**
  * Rebuilds a zero-based-id-indexed array from a store's live `items`, honouring the Id-as-index
  * invariant `references.ts`'s self-referential lookups rely on (`docs/backend.md` → _Reference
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: retire-confirm path override — live paths, not just proficiencies (closes #2099)

**Verdict:** Approve _(submitted as a comment since I'm the PR author; no gating review is accepted on one's own PR)_ — the fix is correct, and it correctly diverges from the issue's suggested `paths: store.paths` where that would have been buggy.

### What's good

- **Correctly rejects the issue's naive fix direction.** #2099 proposed `{ proficiencies: store.profs, paths: store.paths }`, but `pathReferences` reads `paths[p.pathId]?.retiredAt` by array **index** (`references.ts:165`), while it only ever `.filter()`s `proficiencies` (`references.ts:161-162`). Since `ProgressionStore.addPath` prepends unsaved paths with negative ids (`progression-store.svelte.ts:236-238`), passing `store.paths` straight through would misalign every saved path's index and resolve the wrong `retiredAt`. Reaching for `denseByLiveId` — and *not* applying it to `profs` — is exactly right, and matches the reasoning already established for the enemies/skills `ownCatalogueOverride` case. This is the CLAUDE.md "verify the issue's claims are 100% correct" guideline applied well.
- **Clean reuse.** Promoting `denseByLiveId` to an export rather than duplicating the rebuild is the right call, and the updated doc-comment now names both prepend sites (`EntityStore.addItem`, `ProgressionStore.addPath`).

### Test coverage

Solid and sufficient for this frontend change — three tests mirroring the source path (`tests/routes/admin/workbench/progression/PathDetail.test.ts`):
- live-retired gating path suppresses the warning,
- live-reinstated gating path surfaces it (the silently-skipped inverse from the issue),
- an unsaved negative-id path prepended this session doesn't shift the gating path's index.

Each would fail against the pre-fix (or the naive-`store.paths`) behavior, so they genuinely pin the fix. No battle-logic mirror is needed — this is advisory admin-workbench UI, and the authoritative check remains the backend guard (`AdminPaths.FindRetiredPathGatingLiveGateway`).

---

PR #2108, opened 2026-07-17
Floor the typed offense book at 0 under enemy-side absorption

### LINKED ISSUE #2101: TypedDamageDealt (offense book) can also go negative under enemy-side absorption, offsetting proficiency accrual and kill-type attribution

ginderjeremiah:
## Summary

`ResolvePlayerHit`'s per-portion loop (`Game.Core/Battle/BattleContext.cs`) books each portion's post-mitigation net into the typed offense book (`Stats.AddTypedDamageDealt`) via `Battler.HealthRemoved(net, healthBefore)`. `HealthRemoved` capped a killing blow's overkill tail at the health actually removed (#1482), but let a **negative** net — a portion the target's own absorption (resistance > 1) turned into a heal — pass straight through unchanged.

The same helper backs the DoT tick booking in `Battler.ApplyDamageOverTime`, so the identical gap existed on that path too (an absorbed DoT tick's negative `bookedTick` also fed the offense book via `recordDamageDealt`).

Two downstream consumers read `TypedDamageDealt` with no sign guard on the per-type entry:
- `ProficiencyAccrual.BuildActivities` → `FoldIntoActivityKeys` folds the raw (possibly negative) per-type quantity into the shared activity map with no floor — unlike the sibling `AddEvent` tallies just below it, which floor at `amount > 0`. Within a mixed multi-type battle this can silently offset a sibling type's genuine training total.
- `PlayerProgress.RecordKillByDamageType` picks the "majority-dealt" type via `OrderByDescending` to attribute a kill's `KillsByDamageType` credit — a negative entry could skew which type is picked.

This is content-gated today (no authored enemy currently carries resistance > 1) but the mechanic is explicitly supported, mirroring the same root cause #2091 fixed for the Dodge/Parry/Counter avoided-damage tallies.

## Fix

`Battler.HealthRemoved` now floors at 0 in addition to capping at the health removed: `Math.Clamp(damage, 0, Math.Max(0, healthBefore))`. Since this one static helper is the *only* thing feeding `TypedDamageDealt` — both the direct-hit path (`ResolvePlayerHit`, which the parry riposte also routes through) and the DoT tick path (`ApplyDamageOverTime`) — fixing it there closes the gap for every caller at once, rather than needing a separate guard at each call site or in the downstream consumers.

The health math itself is untouched (the doc comment already called this out as "a booking rule for the proficiency offense book only" — that invariant is preserved).

## Test plan

- [x] Updated three existing tests that pinned the old (negative) booking behavior to assert the new floor-at-0 behavior instead: `DamageTarget_AbsorbedHit_FloorsTheTypedOffenseBookAtZero` (renamed), `DamageTarget_MultiPortion_AbsorbingPortionCapsHealAtRoomInFixedOrder`, and two DoT tests in `BattleDamageOverTimeTests.cs`.
- [x] Extended `DamageTarget_ParryCounter_AbsorbedByTheEnemy_BooksNoNegativeCounterDamage` with an assertion that the riposte's typed offense-book entry also floors at 0 (same `ResolvePlayerHit` pipeline).
- [x] `dotnet test Game.sln` — full suite green (Core 1371, Infrastructure 66, Application 989, Api 823).
- [x] No frontend changes — this is a server-only proficiency/statistics booking path (dodge/parry/riposte/offense-book tallies aren't part of the frontend's battle simulation), not part of the tick-by-tick frontend/backend parity surface, consistent with #2091/#2102's precedent.

Closes #2101

---
_Generated by [Claude Code](https://claude.ai/code/session_01AoDn4riF2mg4Qt9ADAJhw7)_

### DIFF
--- a/Game.Core.Tests/Battle/BattleContextTests.cs
+++ b/Game.Core.Tests/Battle/BattleContextTests.cs
@@ -610,10 +610,11 @@ namespace Game.Core.Tests.Battle
         }
 
         [Fact]
-        public void DamageTarget_AbsorbedHit_BooksTheNegativeNetUnchanged()
+        public void DamageTarget_AbsorbedHit_FloorsTheTypedOffenseBookAtZero()
         {
-            // The #1482 cap trims only positive overkill: an absorbed hit's negative net (the capped heal
-            // TakeDamage reports) still books through to the typed book exactly as before.
+            // An absorbed hit's negative net (the capped heal TakeDamage reports) removed no real health, so
+            // the typed offense book floors it at 0 rather than booking a negative that would offset a
+            // sibling type's genuine training within the same battle (#2101).
             var player = MakeBattlerWith((Endurance, 0));
             var enemy = MakeBattlerWith((Endurance, 0), (FireResistance, 2.0)); // Toughness 0, MaxHealth 50
             var context = new BattleContext(player, enemy, timeDelta: 0, new Mulberry32(0));
@@ -622,7 +623,7 @@ namespace Game.Core.Tests.Battle
             context.DamageTarget(20, Single(EDamageType.Fire), 0); // 20 × (1 − 2) = −20, absorbed
 
             Assert.Equal(30, TypedDealt(context, EDamageType.Physical), 0.001);
-            Assert.Equal(-20, TypedDealt(context, EDamageType.Fire), 0.001);
+            Assert.Equal(0, TypedDealt(context, EDamageType.Fire), 0.001);
         }
 
         [Fact]
@@ -964,6 +965,8 @@ namespace Game.Core.Tests.Battle
             // 100 against an enemy absorbing Fire (resistance 2.0) at full health: the Physical portion deals 20
             // (full 50 → 30, opening 20 room), then the Fire portion's −80 absorption heal is capped at that 20
             // room (back to 50). Net 20 + (−20) = 0 — the fixed Physical-first order is what lets the heal land.
+            // The Fire portion's typed offense-book booking floors at 0 (#2101): it removed no real health, so
+            // it trains nothing despite the health math showing the order-dependent heal.
             var player = MakeBattlerWith((Endurance, 0));
             var enemy = MakeBattlerWith((Endurance, 0), (FireResistance, 2.0)); // MaxHealth 50, Toughness 0
             var context = new BattleContext(player, enemy, timeDelta: 0, new Mulberry32(0));
@@ -973,7 +976,7 @@ namespace Game.Core.Tests.Battle
             Assert.Equal(0, dealt, 0.001);
             Assert.Equal(50, enemy.CurrentHealth, 0.001); // 50 → 30 (Physical) → 50 (Fire heal capped at 20 room)
             Assert.Equal(20, TypedDealt(context, EDamageType.Physical), 0.001);
-            Assert.Equal(-20, TypedDealt(context, EDamageType.Fire), 0.001);
+            Assert.Equal(0, TypedDealt(context, EDamageType.Fire), 0.001);
         }
 
         [Fact]
@@ -1890,6 +1893,8 @@ namespace Game.Core.Tests.Battle
             // CapHealToRoom's room is positive, mirroring DamageTarget_ResistanceAboveOne_HealsTheTarget). An
             // absorbed counter avoided no real damage either, so it floors at 0 rather than going negative
             // (#2091) — the sanity assertion on the enemy's health confirms the absorption genuinely happened.
+            // The riposte fires through the same ResolvePlayerHit pipeline as a normal player fire, so its
+            // typed offense book booking floors at 0 too, for the same absorption reason (#2101).
             var counter = MakeCounterSkill(20, damageType: EDamageType.Fire);
             var player = MakeBattlerWithCounter(counter, (ParryChance, 1));
             var enemy = MakeBattlerWith((Endurance, 0), (FireResistance, 2.0)); // MaxHealth 50, Toughness 0
@@ -1901,6 +1906,7 @@ namespace Game.Core.Tests.Battle
 
             Assert.Equal(40, enemy.CurrentHealth, 0.001); // 20 + min(20, 30 room) — absorption genuinely healed it
             Assert.Equal(0, context.Stats.PlayerCounterDamageDealt, 0.001);
+            Assert.Equal(0, TypedDealt(context, EDamageType.Fire), 0.001);
         }
 
         [Fact]
--- a/Game.Core.Tests/Battle/BattleDamageOverTimeTests.cs
+++ b/Game.Core.Tests/Battle/BattleDamageOverTimeTests.cs
@@ -90,7 +90,8 @@ namespace Game.Core.Tests.Battle
             // A +2.0 BleedResistance would heal 80 (2000 × 40/1000 × (1 − 2) = −80), but the battler is already
             // at MaxHealth (no prior damage) so there's no room — capped at zero net effect, consistent with
             // TakeDamage's absorption cap and ApplyHealOverTime (no overheal/shield concept). Per-type booking
-            // still sees the uncapped tick.
+            // floors at 0 too — an absorbed tick removed no real health, so it books nothing (#2101) — even
+            // though the underlying tick is uncapped.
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: Floor the typed offense book at 0 under enemy-side absorption (closes #2101)

**Verdict:** Approve (submitted as a comment since I'm the PR author) — correct, minimal fix at the right chokepoint, with thorough test coverage; one trivial doc nit noted below.

The fix flooring `HealthRemoved` at 0 (in addition to the existing #1482 overkill cap) is the right call, and choosing to own the floor inside `HealthRemoved` rather than at each call site is the correct resolution of the design question the issue explicitly left open:

- `HealthRemoved` has exactly two production callers — `ResolvePlayerHit` (`BattleContext.cs:345`, also the parry riposte's path) and the DoT tick booking (`Battler.cs:534`) — and both use the return value *only* for the offense book, never for real health math. So the "booking rule only, health math untouched" invariant holds, and one edit closes both paths (DRY).
- Those same two are the only feeders of `AddTypedDamageDealt`, so `TypedDamageDealt` per-type entries are now guaranteed `≥ 0`. That transitively fixes **both** downstream consumers the issue flagged — `ProficiencyAccrual.FoldIntoActivityKeys` (`ProficiencyAccrual.cs:168`) and `PlayerProgress.RecordKillByDamageType`'s `OrderByDescending` — with no separate per-consumer guard needed. The issue's open question ("does `RecordKillByDamageType` need an independent guard even after the accrual path is fixed?") is answered cleanly: no, because negatives can no longer enter the map at the source.
- `Math.Clamp(damage, 0, Math.Max(0, healthBefore))` is safe: `min = 0 ≤ max` always, so no `ArgumentException` risk.

### Nits

- `BattleStats.AddTypedDamageDealt`'s doc comment (`BattleStats.cs:164-165`) still reads "already capped at the health it removed by the caller — #1482" and doesn't mention the new floor-at-0 (#2101). The PR was diligent about updating every *other* related doc comment (`HealthRemoved`, the DoT `recordDamageDealt` param, and the `ResolvePlayerHit` book comment), so this one is a small consistency gap. That file isn't in the diff, so no inline comment — purely optional, take it or leave it.

### Test coverage

Solid and matches what the issue asked for. The updated tests pin the exact behavior change (`−20`/`−80 → 0`) across every path:
- Direct absorbed hit — `DamageTarget_AbsorbedHit_FloorsTheTypedOffenseBookAtZero`
- Multi-portion absorbed (Physical=20, Fire=0 in one battle — this directly demonstrates the no-sibling-offset outcome the fix is about)
- Riposte through the same `ResolvePlayerHit` pipeline — extended `DamageTarget_ParryCounter_AbsorbedByTheEnemy_...`
- DoT: absorbed-at-full-health and absorbed-restoring-room-for-a-later-type

No mirrored frontend tests are needed — the offense book is a backend-only side channel with no battle-parity surface, consistent with the #2091/#2102 precedent the PR cites.

### What's good

- Fixes the root cause at the single shared helper rather than patching each downstream reader, and preserves the health math exactly.
- The doc comments clearly explain *why* the floor exists and that it's booking-only, which is the kind of context a later "simplification" edit would otherwise be tempted to strip.
- Tightly scoped and directly traceable to the issue's analysis.

CI is still running (backend tests in progress) as I write this — assuming it goes green, this is good to merge.

---

ginderjeremiah:
## Re-review: doc-comment nit from the prior review addressed (commit `bedc38e`)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the one open nit is resolved and CI is green. Good to merge.

### Since the last review

- ✅ **`BattleStats.AddTypedDamageDealt` doc comment** — the earlier review's only finding. Commit `bedc38e` updates it from "already capped at the health it removed by the caller — #1482" to "already capped at the health it removed and floored at 0 by the caller — #1482/#2101", closing the consistency gap with the other doc comments the PR already updated (`HealthRemoved`, the DoT `recordDamageDealt` param, the `ResolvePlayerHit` book comment). Nothing else changed in this commit.

### Verification (unchanged core, re-confirmed)

- The floor is correct and safe: `Math.Clamp(damage, 0, Math.Max(0, healthBefore))` keeps the #1482 overkill cap and adds the floor-at-0; `min = 0 ≤ max` always holds, so there's no `ArgumentException` risk.
- Booking-only invariant preserved: `HealthRemoved`'s return feeds only `AddTypedDamageDealt` at both call sites (`ResolvePlayerHit`, `ApplyDamageOverTime`) — the health math is never capped or floored.
- Both downstream consumers the issue flagged (`ProficiencyAccrual.FoldIntoActivityKeys`, `PlayerProgress.RecordKillByDamageType`) are transitively fixed at the source: `TypedDamageDealt` per-type entries can no longer go negative, so no per-consumer guard is needed.
- Parity claim confirmed independently: a repo-wide search for the offense book / `HealthRemoved` / the two consumers turns up only backend (`Game.Core`, `Game.Application`) and docs — no frontend (Svelte/TS) references. The offense book is a backend-only proficiency/statistics side channel with no tick-by-tick battle-parity surface, so no mirrored frontend test is required (consistent with the #2091/#2102 precedent).

### CI

Green for head SHA `bedc38e` (run 29584063770): backend build/test/verify and all three e2e browsers pass; frontend test/verify skipped as expected for a backend-only change.

### Test coverage

Unchanged from the last review and still solid — the updated tests pin the `−20`/`−80 → 0` behavior across the direct-hit, multi-portion, riposte, and both DoT paths.

No remaining blocking or should-fix items.

---
_Generated by Claude Code_

---

PR #2109, opened 2026-07-17
Make the write-behind batch's Redis push uncancellable once dispatched

### LINKED ISSUE #2106: Cancellation-during-flush save path can still strand a phantom insert-only unlock (the #1849 half of #2098)
## Problem

#2098 / #2105 fixed the **dispatch-fault** path: `PlayerRepository.SavePlayer` now writes the player cache blob unconditionally once the flush succeeds (before rethrowing the wrapped dispatch fault), so a later DB fall-through no longer surfaces a phantom insert-only unlock (`ItemUnlockedEvent`/`SkillUnlockedEvent`/`ModUnlockedEvent`/`LessonUnlockedEvent`) days after the command appeared to fail.

But #2098's Problem section explicitly noted the **same window exists on the #1849 cancellation path**, and #2105 did not close it. In `SavePlayer` (`Game.DataAccess/Repositories/PlayerRepository.cs`), the flush `catch` is guarded `when (ex is not OperationCanceledException)`, so a cancellation observed **during `FlushAsync`** propagates out *before* the cache-blob write. If the multi-value `LPUSH` had already landed server-side when the cancellation unwound the await, Postgres durably holds the enqueued unlock row while the cache blob is never updated, and `SocketHandler`'s forced reload reads back the stale blob — the exact phantom-unlock divergence #2098 describes, just reached via cancellation rather than a handler fault.

## Why it wasn't folded into #2105

The dispatch-fault path is deterministic: by the time `dispatchFault` is set, the flush has provably succeeded, so writing the cache is unambiguously correct. The cancellation path is not — when the flush await is cancelled, whether the `LPUSH` actually reached Redis is indeterminate:

- If it landed → not writing the cache strands a phantom unlock (this bug).
- If it did not land → writing the cache would create the inverse divergence (cache/session ahead of a DB that never got the row).

So this needs a genuinely different mechanism than #2105's unconditional write, e.g. making the flush's Redis push non-cancellable once initiated (so the cache write always runs deterministically after it), or otherwise confirming the enqueue before deciding whether to write the cache.

## Fix direction

Close the cancellation window so a cancellation that lands the flush server-side still converges the cache blob (or provably does not land the flush). Cover with an integration test mirroring `PlayerWriteBehindTests.SavePlayer_DispatchFaultsForANonCancellationReason_StillFlushesBufferedEnvelopesAndWritesTheCacheBlob_AndThrowsPlayerPersistenceFlushFailedException`, but driving the cancellation-during-flush path.

Relates to #2098, #1849.

ginderjeremiah:
## Summary

`PlayerRepository.SavePlayer`'s flush catch deliberately propagates a cancellation unwrapped rather than as `PlayerPersistenceFlushFailedException` (`Game.DataAccess/Repositories/PlayerRepository.cs:126-155`), and #1849/#2105 made sure that path always forces the connection's in-memory `Player` to reload afterward. But neither closed the specific gap this issue describes: when the cancellation is observed **during** the flush's own `FlushAsync` await, whether the batch's multi-value LPUSH actually reached Redis before the client-side cancellation unwound the await is genuinely indeterminate under the existing `RedisCommandBudget` cooperative-cancellation pattern (abandon the *await*, let the command keep running in the background, only observe a *fault* later). If the push had landed, skipping the cache-blob write (as the old code always did on any cancellation) stranded a phantom insert-only unlock exactly like #2098 — the reload just re-serves the stale cache blob, never learning the DB now durably holds the row.

## Fix

`RedisQueue.AddRangeToQueueAsync` (`Game.Infrastructure/PubSub/Redis/RedisQueue.cs`) — the only caller of which is `PlayerUpdateBatch.FlushAsync`'s batch push — no longer abandons its await on a mid-flight cancellation. It still fails fast up front if the budget is already exhausted before the push is ever dispatched (matching the pinned `SavePlayer_PublishFails_PreservesTheEventForTheNextFlush` semantics), but once the LPUSH is actually sent it always runs to genuine completion rather than being raced against the cancellation token. This gives `SavePlayer` a deterministic outcome: either the push never happened (nothing to persist, batch stays buffered for the next flush) or it landed and the flush returns normally, so the cache-blob write that follows always converges with what was just durably enqueued.

Every other write-behind Redis command is untouched — this scoping is deliberate, since `AddRangeToQueueAsync` backs only this one flush path, while every other write-behind write still benefits from the responsive abandon-on-cancel behavior.

Updated the doc comments in `PlayerRepository.SavePlayer` and `docs/backend-persistence.md` (the "per-command timeout ... forces the same reload" section) to describe the new invariant.

## Test plan

- [x] Added `RedisQueueIntegrationTests.AddRangeToQueueAsync_AlreadyCancelled_ThrowsWithoutPushingAnything` — pre-exhausted budget still fails fast without ever dispatching the command.
- [x] Added `RedisQueueIntegrationTests.AddRangeToQueueAsync_CancelledAfterTheCommandIsDispatched_StillCompletesAndPushesTheValues` — cancelling immediately after the call returns (deterministically after the LPUSH is dispatched, since the only await in the method is the network round trip) still lands the push rather than throwing.
- [x] Added `PlayerWriteBehindTests.SavePlayer_CancelledAfterTheFlushIsDispatched_StillCompletesAndWritesTheCacheBlob` — mirrors `SavePlayer_DispatchFaultsForANonCancellationReason_...`, but for the cancellation-during-flush path: `SavePlayer` completes normally and the reread player reflects the mutation, rather than diverging from the durably-enqueued event.
- [x] `dotnet test Game.sln` — full suite green (Core 1371, Infrastructure 66, Application 992, Api 823).
- [x] No frontend changes — this is a server-only write-behind persistence path, not part of the frontend/backend battle-simulation parity surface.

Closes #2106

---
_Generated by [Claude Code](https://claude.ai/code/session_01YHyrmbHnoDLLxDab1bbqSZ)_


ginderjeremiah (commented):
## Review: Make the write-behind batch's Redis push uncancellable once dispatched (closes #2106)

**Verdict:** Approve _(posted as a comment since I'm the PR author — GitHub won't accept a gating review on your own PR)_ — the change closes the cancellation-during-flush window exactly as #2106 specifies, correctly scoped, well-documented, and thoroughly tested. One nit below.

I traced the mechanism against the actual control flow and it holds up:

- `AddRangeToQueueAsync` now does the `ThrowIfCancellationRequested` **before** dispatch and then awaits the `LPUSH` directly (no `RedisCommandBudget.Write`), so the caller gets the deterministic yes/no #2106 needs: an `OperationCanceledException` out of the flush now provably means the push was never dispatched (nothing landed, batch stays buffered), and a normal return provably means it landed — so the subsequent `_cache.SetAndForget` can never advance the cache ahead of a queue that didn't get the row, nor be skipped over a row that did land. That's the phantom-unlock convergence the issue is about.
- The "exclusively backs `PlayerUpdateBatch`'s flush" scoping claim is accurate — I confirmed the only production caller chain is `AddRangeToQueueAsync` ← `RedisPubSubService.PublishBatch` ← `PlayerUpdateBatch.FlushAsync`; every other `AddRangeToQueueAsync`/`PublishBatch` reference is a test or mock. So narrowing the cancellation behavior here doesn't disturb any other write-behind path.
- Dropping `ObserveWrite` here loses no signal: the abandon-and-log continuation only existed to surface a fault on a command whose await was abandoned. Since the await is no longer abandoned, a genuine push fault now propagates straight to `SavePlayer`'s flush `catch` and is wrapped/handled there — strictly better than a background log.
- Losing the per-command budget's responsiveness on this one path (a hung Redis during flush now blocks for StackExchange's own async timeout rather than the shorter app budget) is a real but deliberate and documented trade-off — determinism over responsiveness for the flush specifically, which is what the issue asked for.

### Nits

- `RedisQueue.cs:14-15` (class-level comment, not in the diff): it still states "every mutating op (pop/reserve/acknowledge/reclaim/remove/push) takes the write path, so a post-cancellation fault on an abandoned write is logged rather than lost." After this PR the range **push** is the one exception — it deliberately doesn't go through `RedisCommandBudget` and isn't abandoned on cancellation. The per-method comment on `AddRangeToQueueAsync` explains this thoroughly, but the class overview now makes a universal claim the code contradicts; a half-line carve-out there ("every mutating op except `AddRangeToQueueAsync` — see its note") would keep the overview honest for the next reader. Trivial; take it or leave it.

### Test coverage

Solid and matches the project's rules. Coverage lands at both the level where the mechanism changed and the level where the bug manifests:
- `AddRangeToQueueAsync_AlreadyCancelled_ThrowsWithoutPushingAnything` pins the fail-fast-before-dispatch half, and `..._CancelledAfterTheCommandIsDispatched_StillCompletesAndPushesTheValues` pins the run-to-completion half.
- `SavePlayer_CancelledAfterTheFlushIsDispatched_StillCompletesAndWritesTheCacheBlob` mirrors the existing dispatch-fault test on the cancellation path and asserts the reread player reflects the mutation — the actual #2098/#1849 convergence.
- The two post-dispatch tests rely on the flush's `LPUSH` being the first genuinely-awaited task (so the immediate cancel lands in the mid-flight window). That's a fair, documented assumption today, though it is inherently timing-sensitive: if a future handler on the zone-change dispatch path introduces an incomplete await ahead of the flush, `SavePlayer_CancelledAfterTheFlushIsDispatched_...` could start cancelling before dispatch and flip to the throw path. Not worth changing now — just flagging the fragility. Server-only path, so no frontend/backend parity mirror is required.

### What's good

- The determinism argument is reasoned through end-to-end and the code, the two touched comments in `SavePlayer`, and `docs/backend-persistence.md` all tell the same consistent story — this is exactly the kind of subtle concurrency invariant that earns the longer comments.
- Minimal, surgical diff that fixes the root cause in one helper rather than bolting a guard onto each call site or the downstream consumers.


---

ginderjeremiah:
## Re-review: class-overview comment carve-out (since 305050f)

**Verdict:** Approve _(posted as a comment since I'm the PR author — GitHub won't accept a gating review on your own PR)_ — the one nit from the prior review is addressed; nothing new introduced.

### Since the last review

- ✅ **`RedisQueue.cs:14-15` class-overview nit** — addressed in `305050f`. The overview no longer makes the now-false universal claim that every mutating op logs an abandoned-write fault; it carves out `AddRangeToQueueAsync` ("...logged rather than lost — except `AddRangeToQueueAsync`, which deliberately does not abandon on cancellation once dispatched (see its own comment, #2106)."). The overview and the per-method comment now tell the same story. Comment-only change, no behavior touched.

### Re-verification of the core change (unchanged since last round)

Re-confirmed the two things this issue hinges on still hold:

- **Determinism** — `AddRangeToQueueAsync` does `ThrowIfCancellationRequested()` before dispatch and then awaits the `LPUSH` directly (no `RedisCommandBudget`/`ObserveWrite`), so `SavePlayer` gets an unambiguous yes/no: an `OperationCanceledException` out of the flush provably means nothing was dispatched (batch stays buffered), and a normal return provably means it landed — so the following `_cache.SetAndForget` can neither run ahead of a queue that never got the row nor be skipped over a row that did land. That is exactly the #2098/#1849 convergence.
- **Scoping** — re-traced callers: the only production caller of `AddRangeToQueueAsync` is `RedisPubSubService.PublishBatch`, whose only production caller is `PlayerUpdateBatch.FlushAsync`; every other reference is a test/mock. So making this one push uncancellable-once-dispatched doesn't disturb any other write-behind path, and dropping `ObserveWrite` here loses no signal (a genuine push fault now propagates straight to the flush `catch` and is wrapped/handled there rather than only logged in the background).

### Test coverage

Unchanged and adequate — the fail-fast-before-dispatch half, the run-to-completion-after-dispatch half, and the `SavePlayer` cache-blob convergence are all pinned. The two post-dispatch tests remain timing-sensitive on the "flush `LPUSH` is the first genuinely-awaited task" assumption (noted last round); still fine today, still worth keeping in mind if a future handler adds an incomplete await ahead of the flush on the zone-change dispatch path.

### CI

`Run tests` green on head `305050f` (run #2581): Core / Infrastructure / Application / Api all pass.

No blocking or should-fix items. Good to merge.

---

PR #2111, opened 2026-07-17
Cancel RenderEngine's pending rAF handle in stop() (#2047)

### LINKED ISSUE #2047: RenderEngine stop→start within one frame permanently doubles the rAF loop (latent)
## Context

Found in the July 2026 health review (stores/engine-lifecycle pass). Latent today — filed because the fix is one line and the failure mode is permanent when it fires.

`RenderEngine.stop()` only flips `running = false` and doesn't cancel the already-queued rAF callback (`UI/src/lib/engine/render-engine.ts:16-37`). If `start()` runs before that callback fires (same task or within ~1 frame), `start()` synchronously begins a new chain, then the old pending callback finds `running === true` and continues too — **two independent rAF chains**, each firing `notifyRenderUpdate` every frame, forever (later stop/start cycles can compound it). `LogicalEngine` guards the equivalent hazard with its `tickSource` handle (`logical-engine.ts:29-38`); `RenderEngine` has only the boolean.

Today `stopEngines` → `startGame` cycles are separated by awaited network calls, so it shouldn't fire — but it's a one-line-away trap for any future synchronous restart.

## Suggested fix

Hold the rAF handle and `cancelAnimationFrame` in `stop()`, or use a generation token captured by each loop. Add the restart-overlap case to `render-engine.test.ts`.

ginderjeremiah:
## Summary

`RenderEngine.stop()` (`UI/src/lib/engine/render-engine.ts`) only flipped the `running` flag and never cancelled the already-queued `requestAnimationFrame` callback. If `start()` ran again within the same frame (or before that callback fired), the stale callback would find `running === true` by the time it fired and reschedule itself — producing a second, independent rAF chain that runs forever alongside the new one, each firing `notifyRenderUpdate` every frame.

`LogicalEngine` already guards the equivalent hazard by holding a `tickSource` handle and cancelling it in `stop()`; `RenderEngine` only had the boolean.

Latent today (current `stopEngines` → `startGame` cycles are separated by awaited network calls), but a one-line-away trap for any future synchronous restart, as noted in the issue.

## Changes

- Hold the rAF handle returned by `requestAnimationFrame` and `cancelAnimationFrame` it in `stop()`, mirroring `LogicalEngine.stop`'s handling of its `tickSource`.
- Added regression tests in `render-engine.test.ts`:
  - `stop()` cancels the pending RAF handle.
  - The correct (latest) handle is cancelled after several frames have run, not a stale one.

## Test plan

- [x] `npx vitest run` — full suite, 299 files / 3764 tests passed
- [x] New tests specifically cover the stop→start-within-one-frame double-loop regression described in the issue

Closes #2047

---
_Generated by [Claude Code](https://claude.ai/code/session_01NEaoMj5dkEGJYyjae17AAJ)_

### DIFF
--- a/UI/src/lib/engine/render-engine.ts
+++ b/UI/src/lib/engine/render-engine.ts
@@ -11,6 +11,7 @@ export class RenderEngine {
 	public tickRate = 0;
 
 	private running = false;
+	private rafHandle?: number;
 	private countTick = getEventCounter((t) => (this.tickRate = Math.round(t)));
 
 	public start() {
@@ -25,6 +26,12 @@ export class RenderEngine {
 
 	public stop() {
 		this.running = false;
+		// Cancel the pending frame so a start() within the same frame can't leave the old callback
+		// running alongside the new loop (mirrors LogicalEngine.stop clearing its tickSource handle).
+		if (this.rafHandle !== undefined) {
+			window.cancelAnimationFrame(this.rafHandle);
+			this.rafHandle = undefined;
+		}
 	}
 
 	//use performance.now instead of animation frame timestamp, because frame stamp is before some amount of processing.
@@ -32,7 +39,7 @@ export class RenderEngine {
 	private renderLoop() {
 		if (this.running) {
 			this.update();
-			window.requestAnimationFrame(() => this.renderLoop());
+			this.rafHandle = window.requestAnimationFrame(() => this.renderLoop());
 		}
 	}
 
--- a/UI/src/tests/lib/engine/render-engine.test.ts
+++ b/UI/src/tests/lib/engine/render-engine.test.ts
@@ -9,10 +9,15 @@ const logicEngineStub = vi.hoisted(() => ({ time: 0 }));
 vi.mock('$lib/engine/engine', () => ({ logicEngine: logicEngineStub }));
 
 const rafCallbacks: (() => void)[] = [];
+const cancelledHandles: number[] = [];
+let rafHandleCounter = 0;
 vi.stubGlobal('window', {
 	requestAnimationFrame: vi.fn((cb: () => void) => {
 		rafCallbacks.push(cb);
-		return rafCallbacks.length;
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: cancel RenderEngine's pending rAF handle in `stop()` (closes #2047)

**Verdict:** Approve — the fix precisely matches the issue's suggested approach, mirrors the existing `LogicalEngine` guard, and is covered by targeted regression tests. (Submitted as a comment for mechanical reasons since the review author and PR author are the same GitHub account; the intended verdict is **Approve**.)

### Assessment

- `renderLoop()` now stores the handle returned by `requestAnimationFrame`, and `stop()` cancels the *latest* handle and clears it — the exact pattern `LogicalEngine.stop()` uses for its `tickSource` (`logical-engine.ts:40-51`). This closes the double-loop hazard: a synchronous `start()` after `stop()` can no longer leave the old pending callback alive to reschedule itself.
- The `rafHandle !== undefined` guard correctly handles the benign cases — `stop()` when nothing is scheduled, and a double `stop()` (canceling an already-fired handle is a harmless no-op). No handle leak, since `stop()` always resets `rafHandle`.
- Comments are proportional to the non-obvious concurrency reasoning and point back to the `LogicalEngine` parallel, which is genuinely useful context here.

### Test coverage

- `render-engine.test.ts` adds two regression tests: `stop()` cancels the pending handle, and it cancels the *latest* handle after several frames have run (not a stale one). The mock upgrade (incrementing `rafHandleCounter` + `cancelledHandles` tracking) is clean and the existing suite (idempotent start, restart-after-stop, delta re-seeding) still exercises the surrounding behavior. Adequate for a frontend-only lifecycle change; no backend/battle-parity surface is touched.

### What's good

A minimal, well-scoped fix for a latent-but-permanent failure mode, with the reasoning captured in both the comment and the tests. Nicely done.

Approving once CI comes back green.

---

Issue #2112, opened 2026-07-17
BattleAlreadyCredited matches on seed alone; CreateBattleSeed never excludes the player's LastCreditedBattleSeed


ginderjeremiah:
## Context

Split out of #2048 during the claim-validation extraction. `BattleAlreadyCredited` (`Game.Application/Services/BattleService.cs`) treats a battle as already-credited purely by `state.BattleSeed == player.LastCreditedBattleSeed`. `CreateBattleSeed` (the same file) draws a fresh cryptographic `uint` with no exclusion — so a 2⁻³² chance the newly-drawn seed for a player's *next* battle collides with their own `LastCreditedBattleSeed` would make `EndBattleVictory`/`EndBattleLoss` treat that brand-new, legitimately-fought battle as a stale re-presentation of the already-credited one and reject/void it (see `BattleService.cs` `BattleAlreadyCredited` doc comment and its three callers).

Low impact (astronomically rare, and the player simply gets one incorrectly-rejected claim rather than any data corruption), so it was intentionally left out of #2048 to keep that extraction focused.

## Suggested fix

Either re-roll `CreateBattleSeed` on collision against the player's `LastCreditedBattleSeed` (would need to thread that value into the seed generation at all three call sites: `BattleService.StartBattle`, `BattleService.StartBossBattle`, and `OfflineProgressService`'s `SeedSource` delegate — currently a bare `Func<uint>` with no player context), or widen `BattleAlreadyCredited`'s match to also compare the enemy id, which is already available at every call site and would rule out the false-positive without touching seed generation at all.

Tags: tech debt, claude, scope: small

PR #2016, opened 2026-07-15
Mirror EndBattleVictory's diagnostics on EndBattleLoss and thread ClientTotalMs through BattleLost

ginderjeremiah:
## Summary

`EndBattleVictory` logs three distinct diagnostics (no resolvable battle, client/server duration divergence, replay-outcome mismatch with seed/enemy/level for reproduction). `EndBattleLoss` only logged the too-early-claim rejection — a `BattleLost` claim whose server replay actually resolved as a *victory* (precisely the frontend/backend battle-logic divergence the parity invariant says must be surfaced) returned `false` with no log at all, and the command had no `ClientTotalMs` to compare against the server's replay duration.

This mirrors all three of `EndBattleVictory`'s structured warnings onto `EndBattleLoss` and threads an optional `ClientTotalMs` through the `BattleLost` command end-to-end:

- **Backend**: `BattleService.EndBattleLoss` now accepts `clientTotalMs` and logs (a) no resolvable active battle, (b) duration divergence, and (c) outcome mismatch (replay was a victory, not a loss) — each with the seed/enemy/level needed to reproduce it.
- **`BattleLostRequest`** (new model) carries the optional `ClientTotalMs`, mirroring `DefeatEnemyRequest`; the `BattleLost` socket command now takes it as a request parameter instead of no parameters.
- **Frontend**: `enemy-manager.ts`'s `resolveBossLoss()` now sends `clientTotalMs: battleEngine.timeElapsed` alongside the `BattleLost` claim, mirroring `claimVictory`'s existing diagnostic report.
- Regenerated the TypeScript client (`dotnet run --project Game.Api -- codegen`) so `IBattleLostRequest` and the socket type map pick up the new parameter — no manual edits to the generated files.

## Why this matters

A real parity bug that flips a win into a client-perceived loss would previously manifest as a player losing earned rewards with nothing in the logs to reproduce it from. This closes that gap.

## Testing

- `Game.Application.Tests`: added `EndBattleLoss_DivergentClientTotalMs_IsDiagnosticOnly_StillReturnsTrue` (mirrors the existing victory-path test — a wildly divergent value is logged but never gates the claim) and `EndBattleLoss_ReplayWasVictory_ReturnsFalseAndLeavesStateActive` (a normal win-shaped battle claimed as a loss is rejected, battle stays active for a legitimate retry).
[...]

PR #2113, opened 2026-07-17
Extract shared battle-end claim-validation pipeline

### LINKED ISSUE #2048: Extract the duplicated battle-end claim-validation pipeline shared by EndBattleVictory and EndBattleLoss
## Context

Found in the July 2026 health review (Game.Application pass).

`EndBattleVictory` (`Game.Application/Services/BattleService.cs:284-369`) and `EndBattleLoss` (`:371-447`) mirror ~60 lines each, block for block: the already-credited guard, the torn-state rejection, the duration-divergence diagnostic, the replay-outcome mismatch rejection, and the `ClaimedBeforeBattleCouldFinish` gate — same structure, same log fields, wording flipped for win/loss. Recent history proves the drift risk: #1993's guard and #2016's diagnostics each had to be hand-mirrored across both paths, which is exactly what the file's own `TryResolveActiveBattle` comment says centralization exists to prevent.

**Failure scenario:** a future guard or log-field change lands in one path and not the other; the asymmetry is invisible until a cheat or divergence is only caught on one outcome path.

## Suggested fix

Extract a shared `TryValidateBattleEndClaim(player, state, expectVictory, clientTotalMs, out …)` that runs the guard chain and emits the parameterized logs, leaving each public method only its payout/recording branch. The rest of BattleService's 831 lines are comment-heavy but cohesive — no broader split warranted.

While in this code, a one-line hardening nit worth folding in: `BattleAlreadyCredited` (`BattleService.cs:741-742`) matches on seed alone, and `CreateBattleSeed` never excludes `player.LastCreditedBattleSeed` — a 2⁻³² fresh-seed collision would void a legitimate win. Re-roll on collision (or also compare enemy id).

ginderjeremiah:
## Summary

`EndBattleVictory` and `EndBattleLoss` (`Game.Application/Services/BattleService.cs`) mirrored ~60 lines each, block for block: the already-credited idempotency guard, torn-state rejection, the client/server duration-divergence diagnostic, the replay-outcome-mismatch rejection, and the `ClaimedBeforeBattleCouldFinish` elapsed-time anti-cheat gate. #1993's guard and #2016's diagnostics each had to be hand-mirrored across both paths, proving the drift risk the issue called out.

This PR extracts the shared pipeline into a single private `TryValidateBattleEndClaim(player, state, expectVictory, clientTotalMs, commandName, out enemy, out result, out playerMaterials, out now, out battleCompletedAt)`, leaving each of `EndBattleVictory`/`EndBattleLoss` with only its outcome-specific payout (`RecordVictory` + cooldown-anchored-to-completion vs. `RecordBattleCompleted` + cooldown-anchored-to-now).

- No behavior change: log wording is unified (parameterized by a `commandName` string) but conveys the same information — the "replay outcome mismatch" message additionally now includes `playerDied` on the loss path (previously victory-only), a strict superset, not a change to what's rejected.
- The intentional cooldown-anchoring asymmetry between the two paths (victory anchors to the battle's server-computed completion time; loss anchors to `now`) is preserved exactly — the shared method surfaces both `now` and `battleCompletedAt` and each caller picks the one it always used.

### Follow-up spun out

The issue also flagged a minor hardening nit: `BattleAlreadyCredited` matches on seed alone, and `CreateBattleSeed` doesn't exclude the player's `LastCreditedBattleSeed`, so a 2⁻³² fresh-seed collision could void a legitimate win. Fixing it properly means threading player context into seed generation across three call sites (`StartBattle`, `StartBossBattle`, and `OfflineProgressService`'s `SeedSource` delegate) or widening the credited-check to also compare enemy id — a distinct, separately-testable change, so it's tracked as its own follow-up: #2112.

## Test plan

- [x] `dotnet build Game.sln` — clean, 0 warnings/errors
- [x] `dotnet test Game.sln` — full suite green (Game.Core.Tests 1371, Game.Infrastructure.Tests 66, Game.Application.Tests 992 incl. all 69 `BattleService*` tests, Game.Api.Tests 823 — 3252 total, 0 failures)

Closes #2048

---
_Generated by [Claude Code](https://claude.ai/code/session_012zTfNFFDYggCsnjMVoZzVA)_

### DIFF
--- a/Game.Application/Services/BattleService.cs
+++ b/Game.Application/Services/BattleService.cs
@@ -283,67 +283,10 @@ namespace Game.Application.Services
 
         public async Task<DefeatResult?> EndBattleVictory(Player player, PlayerState state, int? clientTotalMs = null, CancellationToken cancellationToken = default)
         {
-            // Idempotency backstop (#1874/#1993): a reconnecting client can re-present an already-durably-
-            // credited battle directly through this command instead of the natural AbandonBattle reconnect
-            // path (see BattleAlreadyCredited), so the same guard applies here before paying for the replay
-            // below.
-            if (BattleAlreadyCredited(state, player))
+            if (!TryValidateBattleEndClaim(
+                    player, state, expectVictory: true, clientTotalMs, nameof(EndBattleVictory),
+                    out var enemy, out var result, out var playerMaterials, out var now, out var battleCompletedAt))
             {
-                _logger.LogWarning(
-                    "EndBattleVictory rejected for player {PlayerId}: battle seed {Seed} was already "
-                    + "credited (stale session re-presenting a resolved battle).",
-                    player.Id, state.BattleSeed);
-                state.ClearBattle();
-                return null;
-            }
-
-            if (!TryResolveActiveBattle(state, out var enemy, out var result, out var playerMaterials))
-            {
-                // No battle to resolve. After the caller's HasActiveBattle gate this means a torn state
-                // (an enemy id set without its snapshot), which the set/clear invariant should prevent.
-                _logger.LogWarning(
-                    "EndBattleVictory rejected for player {PlayerId}: no resolvable active battle "
-                    + "(activeEnemyId: {ActiveEnemyId}, hasSnapshot: {HasSnapshot}).",
-                    player.Id, state.ActiveEnemyId, state.Snapshot is not null);
-                return null;
-            }
-
-            // Diagnostic only (not anti-cheat): the client reports the battle duration it simulated, so a
-            // divergence from the server's parity replay is visible even when the claim still resolves as a
-            // win. Logged regardless of the outcome below; absent (null) when not reported.
-            if (clientTotalMs is int reportedMs && reportedMs != result.TotalMs)
-            {
-                _logger.LogWarning(
-                    "EndBattleVictory battle-duration divergence for player {PlayerId}: client reported "
-                    + "{ClientTotalMs}ms but server replay was {ServerTotalMs}ms (delta: {DeltaMs}, "
-                    + "enemyId: {EnemyId}, enemyLevel: {EnemyLevel}, seed: {Seed}).",
-                    player.Id, reportedMs, result.TotalMs, reportedMs - result.TotalMs,
-                    enemy.Id, enemy.Level, state.BattleSeed);
-            }
-
-            if (!result.Victory)
-            {
-                // The server's parity replay of the exact reported battle did not end in a win — a
-                // client/server battle-logic divergence or a forged claim. Seed + enemy + level reproduce it.
-                _logger.LogWarning(
-                    "EndBattleVictory rejected for player {PlayerId}: server replay was not a victory "
-                    + "(enemyId: {EnemyId}, enemyLevel: {EnemyLevel}, seed: {Seed}, playerDied: {PlayerDied}, "
-                    + "replayMs: {ReplayMs}, isBoss: {IsBoss}, zoneId: {ZoneId}).",
-                    player.Id, enemy.Id, enemy.Level, state.BattleSeed, result.PlayerDied,
-                    result.TotalMs, state.IsBossBattle, state.BattleZoneId);
-                return null;
-            }
-
-            var now = DateTime.UtcNow;
-
-            if (ClaimedBeforeBattleCouldFinish(state.BattleStartTime, result.TotalMs, now, out var battleCompletedAt))
-            {
-                _logger.LogWarning(
-                    "EndBattleVictory rejected for player {PlayerId}: claimed before the battle could finish "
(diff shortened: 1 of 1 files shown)


Game.Application/Services/BattleService.cs
```diff
@@ -370,79 +313,117 @@ public async Task<bool> SetAutoChallengeBoss(Player player, bool enabled, Cancel
...
+            bool expectVictory,
+            int? clientTotalMs,
+            string commandName,
+            [MaybeNullWhen(false)] out CoreEnemy enemy,
+            [MaybeNullWhen(false)] out BattleResult result,
+            out BattlerMaterials? playerMaterials,
+            out DateTime now,
+            out DateTime battleCompletedAt)
+        {
+            now = DateTime.UtcNow;
```
ginderjeremiah: Nit (non-blocking, no change needed): `now` is now captured at the very top of the method — before `TryResolveActiveBattle`'s replay — whereas both original paths captured it *after* the replay. The delta is only the replay's CPU time (well under one tick), and it shifts `now` marginally earlier, which is the strictly-safe direction for the `ClaimedBeforeBattleCouldFinish` gate (a hair stricter, no false rejections at sub-tick granularity) and immaterial for the `RecordVictory`/`RecordBattleCompleted` timestamp. So the "no behavior change" claim holds in practice; flagging only so it's a conscious choice rather than an accidental reorder.

---

ginderjeremiah (commented):
## Review: Extract shared battle-end claim-validation pipeline (closes #2048)

**Verdict:** Approve — the extraction is faithful, behavior-preserving, and does exactly what the issue asked for. (Submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author; treat this as an approve.)

This cleanly collapses the ~60 lines each of `EndBattleVictory`/`EndBattleLoss` into a single `TryValidateBattleEndClaim`, leaving each caller only its outcome-specific payout — the exact drift risk #2048 called out (proven by #1993/#2016 having to be hand-mirrored).

### Correctness
- Guard chain, order, and short-circuit semantics are preserved on both paths: already-credited (clears state, rejects), torn-state, duration-divergence diagnostic, replay-outcome mismatch (now `result.Victory != expectVictory`), and the `ClaimedBeforeBattleCouldFinish` gate.
- The out-param contract is sound: `enemy`/`result` use `[MaybeNullWhen(false)]` and are `default`-set on the early-reject branches; `battleCompletedAt`/`now` are always assigned; `playerMaterials` is threaded through to `RecordVictory` (victory) or discarded (loss).
- The intentional cooldown-anchoring asymmetry is preserved exactly — victory anchors to `battleCompletedAt + PostBattleCooldown`, loss to `now + PostBattleCooldown` — with both surfaced from the shared method and each caller picking the one it always used.
- The unified mismatch log now includes `playerDied` on the loss path too (previously victory-only) — a strict superset of information, not a change to what's rejected.

### Nits
- `now` is captured before the replay now rather than after it (inline note on `BattleService.cs:356`) — practically immaterial and in the safe direction; flagged only so it's a conscious choice.

### Test coverage
- Sufficient. This is a behavior-preserving refactor and the full guard chain is exercised on **both** sides: `EndBattleLoss` has tests for torn-state, elapsed-time gate, already-credited, divergence-diagnostic, and replay-was-victory; `EndBattleVictory` is covered across the suite. No new tests are warranted for a pure extraction with existing coverage on both paths.

### Follow-up
- The `BattleAlreadyCredited` seed-collision hardening nit was correctly split out to #2112 rather than folded in — it needs player context threaded into `CreateBattleSeed` across three call sites (or a wider match), a distinct separately-testable change. Good scoping call.

### What's good
- The method-level comment documenting *why* both `now` and `battleCompletedAt` are surfaced (the deliberate cooldown asymmetry) is exactly the kind of context that prevents a future "simplification" from collapsing the two paths incorrectly.

Holding off on merging: `test-backend` is still in progress, and there's a (trivial) nit above — I'll leave the merge to you once CI is green.

---

ginderjeremiah:
## Re-review: nit resolved, merging once CI is green

Verdict remains **approve**. The extraction is faithful and behavior-preserving, and the earlier review already covered the full guard-chain correctness, the preserved cooldown-anchoring asymmetry, and the correctly-scoped #2112 follow-up. Nothing there changed.

The one outstanding item was the nit I'd flagged: `now` was captured at the top of `TryValidateBattleEndClaim` (before the parity replay) rather than after it, as both callers did pre-extraction. It's practically immaterial and in the safe direction, but it *was* a behavior change against the PR's stated no-behavior-change goal — so since the review was over an hour old with nothing else pending, I went ahead and fixed it (d27916f): `now` is initialized to `default` and the real timestamp is captured right before the elapsed-time gate, exactly restoring the original ordering. This makes the extraction a true no-op again — the anti-cheat comparison and the payout timestamp both reflect the moment of adjudication.

- `dotnet build Game.Application` — clean, 0 warnings/errors
- `Game.Application.Tests` — 992/992 green (all `BattleService*` tests included)

No remaining nits or should-fixes. I'll merge once CI on the new commit is green.

---

PR #2117, opened 2026-07-17
Adopt central package/build-property management

### LINKED ISSUE #2054: Adopt central package/build-property management; fix the wildcard NuGet version already causing silent drift
## Context

Found in the July 2026 health review (build/solution hygiene pass). Three related items, best done as one change:

1. **Wildcard version with live drift:** `Game.Application.csproj:4` pins `Microsoft.Extensions.Logging.Abstractions` at `Version="10.*"` while siblings pin `10.0.9` (`Game.Core.csproj`, `Game.Infrastructure.csproj:24-31`, `Game.DataAccess.csproj:29`, `Game.Infrastructure.Tests.csproj:17`). Not theoretical — `project.assets.json` shows the wildcard resolves to **10.0.10 today**, and NuGet unifies the whole graph upward past every sibling's pin. With no `packages.lock.json`, any restore silently picks up a new 10.x the day it ships; builds aren't reproducible and the pins elsewhere are effectively dead.
2. **No central management:** no `Directory.Build.props`/`Directory.Packages.props`; `xunit.v3 3.2.2` repeated in 4 csproj, `Microsoft.Extensions.* 10.0.9` in ~7, `Testcontainers.* 4.12.0` duplicated; `TargetFramework`/`Nullable`/`ImplicitUsings` duplicated in all 12 projects — consistent today only by discipline.
3. **Stranded one-off settings:** `Game.DataAccess.csproj:7,10-16` is the only project with `EnforceCodeStyleInBuild` and `CheckForOverflowUnderflow` — if overflow checking matters anywhere it's Game.Core's battle/XP math, which *doesn't* have it. `Game.Core.csproj:7` carries a dead `UserSecretsId` on a pure domain library. (Cosmetic: `Game.Application.csproj` orders `ItemGroup`s before the `PropertyGroup`, unlike every sibling.)

## Suggested fix

Adopt `Directory.Packages.props` (CPM) for all package versions (pinning the wildcard to a concrete version) and a `Directory.Build.props` for TFM/Nullable/ImplicitUsings. Decide overflow-checking scope deliberately — hoist to the shared props (likely at least Game.Core) or drop it — and remove the dead `UserSecretsId`.

ginderjeremiah:
## Summary

Closes #2054.

- **Adds `Directory.Packages.props`** (Central Package Management) covering all 17 distinct NuGet packages used across the 12 backend projects. This fixes the silently-drifting wildcard: `Game.Application.csproj` pinned `Microsoft.Extensions.Logging.Abstractions` at `10.*`, which every sibling project pinned at `10.0.9` — NuGet was unifying the whole graph upward past those pins (resolving to `10.0.10` at the time of the health review). The wildcard is now pinned to the concrete `10.0.9` used everywhere else.
- **Adds `Directory.Build.props`** with the `TargetFramework`/`Nullable`/`ImplicitUsings` triplet that was duplicated verbatim in all 12 `.csproj` files.
- **Removes the dead `UserSecretsId`** from `Game.Core.csproj` — a pure domain library with no reason to carry one (`Game.Api.csproj`, the actual host, keeps its own).
- **Fixes the stray `EFCore.Design` metadata**: kept its `PrivateAssets`/`IncludeAssets` on the local `PackageReference` (CPM only centralizes the version) in `Game.Infrastructure.csproj`.
- **Decides the `CheckForOverflowUnderflow` scope**: it was previously only set (for both Debug and Release) on `Game.DataAccess`, a project with essentially no arithmetic — not on `Game.Core`, where the battle/XP math actually lives. Hoisted it into `Directory.Build.props`, scoped to `Debug` builds only across the whole solution, so it's a dev/test-time bug detector without a Release runtime cost.
- **Fixed a latent bug this surfaced immediately**: `Mulberry32.Next()` (the battle RNG) deliberately relies on `uint` wraparound as part of the algorithm, and had never run under overflow checking before. Wrapped its arithmetic in an explicit `unchecked` block — this documents the intentional wraparound and makes it immune to the project's overflow-checking setting rather than relying on it happening to be off. Without this, every battle simulation throws `OverflowException` the instant Debug overflow checking is enabled solution-wide (found via the full test run: ~300 failures in `Game.Core.Tests`/`Game.Application.Tests`/`Game.Api.Tests`, all with the same root-cause stack trace).

### Left out of scope

The issue also flagged `Game.DataAccess.csproj` as the only project with `EnforceCodeStyleInBuild`. I left it as-is rather than hoisting it solution-wide — turning on build-time style enforcement everywhere is a separate, riskier decision that could surface a large number of new violations across projects that have never been checked against it. Happy to follow up as its own issue if that's wanted.

The cosmetic `Game.Application.csproj` `ItemGroup`-before-`PropertyGroup` ordering issue is resolved as a side effect: that project's `PropertyGroup` only ever held the now-centralized TFM/Nullable/ImplicitUsings, so it's removed entirely rather than reordered.

## Test plan

- [x] `dotnet build Game.sln` (Debug) — clean, 0 warnings/errors
- [x] `dotnet build Game.sln -c Release` — clean, 0 warnings/errors
- [x] `dotnet test Game.sln --no-build -c Debug` — all 3,258 tests pass (Game.Core.Tests, Game.Application.Tests, Game.Api.Tests, Game.Infrastructure.Tests), after the `Mulberry32` fix

---
_Generated by [Claude Code](https://claude.ai/code/session_0194SwL8qLgTi1nBk11ZUCv5)_

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,15 @@
+<Project>
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <!-- Overflow checking only in Debug: it's a dev/test-time bug detector (e.g. Game.Core's battle/XP math),
+       not something worth paying the runtime cost for in Release. -->
+  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
+    <CheckForOverflowUnderflow>True</CheckForOverflowUnderflow>
+  </PropertyGroup>
+
+</Project>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,27 @@
+<Project>
+
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageVersion Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore.Design" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore.Relational" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.Extensions.Options" Version="10.0.9" />
+    <PackageVersion Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.23.0" />
+    <PackageVersion Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="10.0.2" />
+    <PackageVersion Include="StackExchange.Redis" Version="3.0.0" />
+    <PackageVersion Include="Swashbuckle.AspNetCore" Version="10.2.1" />
+    <PackageVersion Include="Testcontainers.PostgreSql" Version="4.12.0" />
+    <PackageVersion Include="Testcontainers.Redis" Version="4.12.0" />
+    <PackageVersion Include="xunit.v3" Version="3.2.2" />
+    <PackageVersion Include="xunit.v3.extensibility.core" Version="3.2.2" />
+  </ItemGroup>
+
+</Project>
--- a/Game.Abstractions/Game.Abstractions.csproj
+++ b/Game.Abstractions/Game.Abstractions.csproj
@@ -1,11 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>net10.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
-  </PropertyGroup>
-
   <ItemGroup>
     <ProjectReference Include="..\Game.Core\Game.Core.csproj" />
   </ItemGroup>
--- a/Game.Api.Tests/Game.Api.Tests.csproj
+++ b/Game.Api.Tests/Game.Api.Tests.csproj
@@ -1,10 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net10.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
-
     <OutputType>Exe</OutputType>
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
@@ -12,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="xunit.v3" Version="3.2.2" />
+    <PackageReference Include="xunit.v3" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Game.Api/Dockerfile
+++ b/Game.Api/Dockerfile
@@ -15,11 +15,14 @@ WORKDIR /src
 
 # Restore from the project manifests *before* copying the rest of the source — the standard .NET
 # multi-stage layering. NuGet restore is the slow part of the build and its inputs are only the
-# .csproj files, so isolating it in its own layer lets Docker reuse the restored packages from its
-# local layer cache on incremental rebuilds (e.g. local dev) whenever the dependency graph is
-# unchanged, even when source files change. The set below is Game.Api's full restore graph (it
-# references Abstractions/Core/DataAccess/Application; DataAccess pulls in Infrastructure). Add a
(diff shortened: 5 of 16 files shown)


Directory.Build.props
```diff
@@ -0,0 +1,15 @@
+<Project>
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
ginderjeremiah, line 6: **Blocking:** Moving `TargetFramework` (and the package versions in `Directory.Packages.props`) into these root props files breaks the `Game.Api` Docker build. `Game.Api/Dockerfile` restores from a layer that `COPY`s only the individual `.csproj` files before running `dotnet restore` (lines 22–29), and it never copies `Directory.Build.props`/`Directory.Packages.props`. Without them, `Game.Api.csproj` has no `TargetFramework`, so restore fails with:

```
error NETSDK1013: The TargetFramework value '' was not recognized. [/src/Game.Api/Game.Api.csproj]
```

This is why all three `test-e2e` jobs failed in ~26s — they build the API image via `docker-compose` (`dockerfile: Game.Api/Dockerfile`). Fix: copy the props into the restore layer before `RUN dotnet restore`, e.g. add above line 28:

```dockerfile
COPY Directory.Build.props Directory.Packages.props ./
```

(Also worth a comment there noting these must be kept in the restore layer, since the whole point of centralizing is that projects now depend on them at restore time.)

---

ginderjeremiah (commented):
## Review: Adopt central package/build-property management (closes #2054)

**Verdict:** Request changes — the CPM/`Directory.Build.props` migration is clean and complete, but it breaks the `Game.Api` Docker build, which is a merge blocker. _(Submitted as a comment because I'm the PR author; the intended verdict is Request changes.)_

The core of this change is exactly right: all 17 referenced NuGet packages are centralized in `Directory.Packages.props` with no orphans and no leftover `Version=` attributes on any `PackageReference`; the wildcard drift (`10.*` → `10.0.9`) is fixed; the TFM/Nullable/ImplicitUsings triplet is hoisted correctly; the dead `UserSecretsId` is gone from `Game.Core`; and `EFCore.Design`'s `PrivateAssets`/`IncludeAssets` are preserved on the local reference. The `Mulberry32` `unchecked` fix is correct and, importantly, keeps it consistent with the frontend port (`UI/src/lib/engine/mulberry32.ts`, which already pins to uint32 via `>>> 0`/`Math.imul`) — so battle parity is maintained and no mirrored test change is needed.

### Blocking

- **`Game.Api/Dockerfile` restore layer doesn't copy the new props files** — `Directory.Build.props` / `Directory.Packages.props`. The Dockerfile `COPY`s only the individual `.csproj` manifests before `RUN dotnet restore` (lines 22–29), so `Game.Api.csproj` no longer resolves a `TargetFramework` and restore dies with `NETSDK1013`. This is the actual cause of the three red `test-e2e` jobs (they build the API image via `docker-compose`). Fix: `COPY Directory.Build.props Directory.Packages.props ./` before the restore step. See the inline comment for details.

### Should-fix (follow-up, not a blocker on this PR)

- **`EnforceCodeStyleInBuild` left stranded on `Game.DataAccess` only.** Issue #2054 explicitly called this out as a one-off to resolve, and the PR consciously defers it — a defensible call, since enabling it solution-wide could surface a large number of new violations. But it shouldn't just be dropped: please open a follow-up issue (tags: `tech debt`, `claude`, `scope: small`) so the decision to hoist-or-remove it is tracked rather than lost.

### Behavioral note (not blocking)

- Overflow checking is now on solution-wide in **Debug** (previously only `Game.DataAccess`, and there in both Debug and Release). This is the intended, documented change and the full suite passes after the `Mulberry32` fix — but it does mean any other code relying on silent integer wraparound will now throw under Debug/test. `Mulberry32` was the only such spot found; worth keeping in mind for future arithmetic-heavy code in `Game.Core`.

### Test coverage

Appropriate — this is a build-configuration change with no new domain logic. The `Mulberry32` edit is behavior-preserving (it only prevents an `OverflowException` under the newly-enabled checking; the numeric output is unchanged), and it's already covered by the existing `Mulberry32` unit and parity tests on both sides. The real gate here is the build/CI, which correctly caught the Docker regression.

### What's good

Thorough, well-scoped execution of the issue: the completeness of the package inventory, the deliberate reasoning on overflow-checking scope, and proactively catching + fixing the latent `Mulberry32` overflow bug (with a clarifying comment) rather than just flipping the setting. Once the Dockerfile copies the props files and CI goes green, this is a straightforward approve.

---

ginderjeremiah:
## Re-review — approve ✅

The one blocker from the prior review has been addressed and CI is fully green, so this is good to merge.

### What changed since the last review
- **Dockerfile restore-layer fix (`4b0165e1`).** `Game.Api/Dockerfile` now `COPY`s `Directory.Build.props Directory.Packages.props ./` before the `.csproj` copies and `dotnet restore`, so the API image again resolves `TargetFramework`/package versions from the centralized props. This resolves the `NETSDK1013` restore failure that was red-lighting the three `test-e2e` jobs. The comment above the copy block was updated to explain why the props must live in the restore layer — a good call, since the whole point of CPM is that projects now depend on them at restore time.

### CI (commit `4b0165e1`)
All green: `test-backend`, `verify-backend`, `test-e2e (chromium / firefox / webkit)`, `all-checks-passed`. `test-frontend`/`verify-frontend` correctly skipped (no frontend changes).

### Follow-up tracked
The deferred `EnforceCodeStyleInBuild` decision (left `Game.DataAccess`-only, out of scope for a package-centralization PR) is tracked in #2118 (`tech debt`, `claude`, `scope: small`).

The rest of the migration — the complete 17-package `Directory.Packages.props` inventory with no orphaned `Version=` attributes, the hoisted TFM/Nullable/ImplicitUsings, the removed dead `UserSecretsId`, the preserved `EFCore.Design` `PrivateAssets`/`IncludeAssets`, and the `Mulberry32` `unchecked` fix (behavior-preserving, keeps parity with the `>>> 0`/`Math.imul` frontend port) — was validated in the prior review and is unchanged.

Merging now.

---

PR #2119, opened 2026-07-17
Resolve codegen repo root via CodeGenPaths.FindRepositoryRoot in Startup

### LINKED ISSUE #2055: CodeGen dev-startup resolves the repo root differently than CodeGenCommand, contradicting the centralization comment
## Context

Found in the July 2026 health review (build/CI pass).

`Game.Api/Startup.cs:205-207` resolves the codegen output root as `Directory.GetParent(ContentRootPath)`, while `Game.Api/CodeGen/CodeGenPaths.cs:29-45` (`FindRepositoryRoot`) walks up to `Game.sln`. `CodeGenPaths.cs:3-7` claims both paths are "centralized so … compute the frontend target directory the same way" — which is currently false.

**Impact:** works today (`dotnet run` sets ContentRoot to `Game.Api/`, parent = repo root), but a Development run with any other content root (published output, `--contentRoot`) writes the generated client to the wrong directory or throws — and the comment actively misleads about the invariant. The rest of the pipeline is solid (CI byte-equality drift gate, ordered writers, LF pinning).

## Suggested fix

`var rootFolder = CodeGenPaths.FindRepositoryRoot(app.Environment.ContentRootPath);` in Startup.

ginderjeremiah:
## Summary

Closes #2055.

`Game.Api/Startup.cs` computed the dev-time codegen output root as
`Directory.GetParent(app.Environment.ContentRootPath)`, while `CodeGenCommand`
(the standalone `dotnet run --project Game.Api -- codegen` entry point) used
`CodeGenPaths.FindRepositoryRoot()`, which walks up from a starting directory
until it finds `Game.sln`. `CodeGenPaths`' doc comment claims both paths are
"centralized so … compute the frontend target directory the same way," which
was false — they only happened to agree because `dotnet run`'s default
`ContentRootPath` is `Game.Api/`, whose parent is the repo root.

With any other content root (published output, `--contentRoot`), the old
`Startup.cs` logic would write the generated TypeScript client into the wrong
directory or throw, silently contradicting the shared-path invariant the
comment promises.

## Fix

`Startup.cs` now calls `CodeGenPaths.FindRepositoryRoot(app.Environment.ContentRootPath)`,
matching `CodeGenCommand`'s approach and making the two paths agree for any
content root, not just the `dotnet run` default.

## Test plan

- [x] `dotnet build Game.sln` (Debug) — clean, 0 warnings/errors
- [x] `dotnet test Game.sln --no-build --no-restore` — all 3,258 tests pass across Game.Core.Tests, Game.Application.Tests, Game.Api.Tests, Game.Infrastructure.Tests
- [x] Existing `CodeGenPathsTests` already cover `FindRepositoryRoot`'s walk-up-to-`Game.sln` behavior; no behavior change to that method, only which caller uses it

---
_Generated by [Claude Code](https://claude.ai/code/session_014DhVPYd2HK92GWxXadPywS)_

### DIFF
--- a/Game.Api/Startup.cs
+++ b/Game.Api/Startup.cs
@@ -202,8 +202,7 @@ namespace Game.Api
             {
                 // Regenerate the frontend's TypeScript API client from the running API's types. This
                 // writes into the UI source tree, so it is strictly a local-development concern.
-                var rootFolder = (Directory.GetParent(app.Environment.ContentRootPath)
-                    ?? throw new InvalidOperationException($"Could not resolve parent directory of '{app.Environment.ContentRootPath}'.")).FullName;
+                var rootFolder = CodeGenPaths.FindRepositoryRoot(app.Environment.ContentRootPath);
                 var targetDir = CodeGenPaths.ResolveTargetDirectory(rootFolder);
                 var codeGen = app.Services.GetRequiredService<ApiCodeGenerator>();
                 codeGen.GenerateCode(typeof(Startup).Assembly, new CodeGenOptions { TargetDirectory = targetDir, NewLine = "\n" });


ginderjeremiah (commented):
## Review: Resolve codegen repo root via `CodeGenPaths.FindRepositoryRoot` in Startup (closes #2055)

**Verdict:** Approve — correct, minimal fix that resolves the exact defect in the issue. _(Posted as a comment because GitHub won't accept a gating verdict from the PR author.)_

The change swaps `Startup.cs`'s ad-hoc `Directory.GetParent(ContentRootPath)` for `CodeGenPaths.FindRepositoryRoot(app.Environment.ContentRootPath)`, so the dev-startup hook now resolves the repo root by walking up to `Game.sln` — identical to how `CodeGenCommand` does it (`CodeGenCommand.cs:35`). This makes the `CodeGenPaths` doc-comment's "both compute the target directory the same way" invariant actually true instead of coincidentally-true-only-under-`dotnet run`. Passing `ContentRootPath` explicitly (rather than defaulting to `AppContext.BaseDirectory`) is the right call here since the hook runs inside the host and the content root is the meaningful anchor. Correct and well-scoped.

### Test coverage

Adequate. The changed behavior lives entirely in `FindRepositoryRoot`, which is already covered by `CodeGenPathsTests` (walk-up-to-`Game.sln` success and the throw-when-not-found path). The one modified line is in `Startup.Main`'s Development-only bootstrap, which isn't unit-testable in isolation and gains no coverage value from a bespoke test — no new tests warranted.

### What's good

- Deletes the now-redundant local `InvalidOperationException` throw; `FindRepositoryRoot` already raises a clearer, more specific error on failure, so there's no behavioral loss and less duplicated guard logic.
- Tightly scoped to the reported defect with no incidental churn.

---

PR #2120, opened 2026-07-18
Send hot-path Redis Lua scripts as EVALSHA instead of full text (#2056)

### LINKED ISSUE #2056: Hot-path Redis Lua scripts are re-sent as full EVAL text on every call
## Context

Found in the July 2026 health review (Infrastructure pass).

`IDatabase.ScriptEvaluate(string, …)` in StackExchange.Redis transmits the full script body per invocation (no automatic `EVALSHA`). The affected scripts sit on the hottest Redis paths:

- `Game.Infrastructure/Cache/Redis/RedisService.cs:60-62` (`GetSet`), `:116-119` (`ReclaimAndForget` — runs on every inbound socket message, i.e. every heartbeat), `:158-163` (`HashGetAllIfExists`), `:190-192`/`:206-211` (hash-set scripts — per battle completion)
- `Game.Infrastructure/PubSub/Redis/RedisQueue.cs:95-97` (`ReclaimProcessingAsync`)

At fleet scale this is a few hundred redundant bytes per command on every hot path. Not a correctness issue.

## Suggested fix

Prepare each script once with `LuaScript.Prepare(...)`/`LoadedLuaScript` (or `SCRIPT LOAD` + `EVALSHA` with an `EVAL` fallback) in `RedisService`/`RedisQueue`, keeping the scripts as the single source strings they already are. Could bundle with other Redis round-trip work (#2019).

ginderjeremiah:
## Summary

`RedisService`/`RedisQueue` sent every Lua script's full source text on every `ScriptEvaluate(Async)` call — `IDatabase.ScriptEvaluate(string, ...)` always issues a raw `EVAL`, with no automatic `EVALSHA`. This hits the hottest Redis paths: `GetSet`/`CompareAndSet`, `ReclaimAndForget` (every inbound socket message), the hash-get/hash-set scripts (per battle completion), and `RedisQueue.ReclaimProcessingAsync`.

- Added `PreparedScript` (`Game.Infrastructure/Redis/PreparedScript.cs`): wraps a script's source text with its SHA1 hash. The scripts stay raw `KEYS[n]`/`ARGV[n]` text — `LuaScript.Prepare`'s fixed named-parameter model can't express the variable-length `ARGV` the hash-set scripts build, so this talks to `IDatabase` directly instead.
  - The first call this process makes for a given script sends the full text (an `EVAL`, which as a side effect caches the script on the server), so a cold script cache never causes a `NOSCRIPT` failure on startup.
  - Every call after that sends only the hash (`EVALSHA`).
  - The awaited path (`EvaluateAsync`) additionally falls back to a full-text resend if the server ever reports `NOSCRIPT` after the script was already warmed (a manual `SCRIPT FLUSH` or a Redis restart that lost its script cache), so it self-heals. The fire-and-forget path (`Evaluate`) has no response to inspect for that — the same no-delivery-guarantee every other `CommandFlags.FireAndForget` call in this tier already accepts.
  - `RedisService`/`RedisQueue` are resolved **transient** (constructed per DI scope), so each script is prepared once as a `static readonly` field on its owning class — the hash and warm state are shared process-wide rather than recomputed or re-sent per instance.
- Converted every `ScriptEvaluate`/`ScriptEvaluateAsync` call site in `RedisService.cs` and `RedisQueue.cs` to go through a `PreparedScript`.

I manually validated the hash format and the NOSCRIPT/fallback behavior against a live Redis instance before wiring this in (a computed `SHA1` hash matches exactly what `ScriptEvaluate(byte[] hash, ...)` expects, a pre-load `EVALSHA` genuinely returns `NOSCRIPT`, and it succeeds after the script is loaded).

Not a correctness fix — this is purely a byte-savings optimization, per the issue.

## Test plan

- [x] `dotnet build Game.sln` — clean, 0 warnings/errors
- [x] New unit tests in `Game.Infrastructure.Tests/PreparedScriptTests.cs` pin the one genuinely regression-prone piece of logic — a failed cold call must not mark the script warmed (or every later call would jump straight to an `EVALSHA` the server never actually cached) — using the project's existing dead-endpoint technique (no live Redis, no mocking).
- [x] Full existing Redis integration suites (`RedisServiceIntegrationTests`, `RedisQueueIntegrationTests`) pass unchanged against the real Redis container, exercising every converted script end to end.
- [x] `Game.Infrastructure.Tests` (68 tests), `Game.Application.Tests` (998 tests), `Game.Api.Tests` (823 tests), `Game.Core.Tests` (1371 tests) all pass.

Closes #2056

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y98qiKza3cxfNGo4AnS2Ec)_

### DIFF
--- a/Game.Infrastructure.Tests/PreparedScriptTests.cs
+++ b/Game.Infrastructure.Tests/PreparedScriptTests.cs
@@ -9,7 +9,9 @@ namespace Game.Infrastructure.Tests
     /// send a Lua script's hash instead of its full text on every call after the first (#2056). The behaviour
     /// that matters and is otherwise only exercised indirectly through a live Redis: a cold call that fails
     /// before it reaches the server must not mark the script warmed, or every later call would jump straight to
-    /// an <c>EVALSHA</c> the server has genuinely never cached.
+    /// an <c>EVALSHA</c> the server has genuinely never cached — pinned for both <see cref="PreparedScript.EvaluateAsync"/>
+    /// and <see cref="PreparedScript.Evaluate"/>, since the latter blocks its own cold call rather than sharing
+    /// the former's fallback logic.
     /// <para>
     /// Uses the same dead-endpoint technique as the sibling <see cref="RedisPubSubPublishFailureTests"/>: a
     /// multiplexer pointed at an unreachable endpoint (<c>abortConnect=false</c>) connects lazily and then throws
@@ -38,15 +40,32 @@ namespace Game.Infrastructure.Tests
         }
 
         [Fact]
-        public async Task EvaluateAsync_WhenAlreadyWarmed_SendsTheHashRatherThanRewarming()
+        public async Task Evaluate_WhenTheColdCallFails_LeavesTheScriptUnwarmed()
+        {
+            // Evaluate's cold branch strips a caller's FireAndForget flag and blocks for a confirmed reply, so
+            // this is pinned separately from EvaluateAsync's version of the same invariant: a dropped cold send
+            // must not wedge the script into the hash-only path.
+            await using var multiplexer = await ConnectionMultiplexer.ConnectAsync(DeadEndpoint);
+            var db = multiplexer.GetDatabase();
+            var script = new PreparedScript("return 1");
+
+            Assert.Throws<RedisConnectionException>(() => script.Evaluate(db, [], [], CommandFlags.FireAndForget));
+
+            Assert.False(script.IsWarmedForTesting);
+        }
+
+        [Fact]
+        public async Task EvaluateAsync_WhenAlreadyWarmed_StaysWarmedAfterAFailedCall()
         {
             await using var multiplexer = await ConnectionMultiplexer.ConnectAsync(DeadEndpoint);
             var db = multiplexer.GetDatabase();
             var script = PreparedScript.CreateAlreadyWarmedForTesting("return 1");
 
-            // Still fails against a dead endpoint either way, but this pins that a warmed script never revisits
-            // the cold (full-text) branch — CreateAlreadyWarmedForTesting's contract that the seam actually
-            // starts in the warmed state.
+            // A dead endpoint can't distinguish the hash branch from the full-text one — both throw the same
+            // RedisConnectionException before anything reaches a server — so this doesn't pin which command form
+            // was sent (that needs a genuine NOSCRIPT reply, covered by the Redis integration suites instead).
+            // What it does pin: CreateAlreadyWarmedForTesting actually starts warmed, and a failure afterward
+            // doesn't flip it back to cold.
             Assert.True(script.IsWarmedForTesting);
             await Assert.ThrowsAsync<RedisConnectionException>(() => script.EvaluateAsync(db, [], []));
             Assert.True(script.IsWarmedForTesting);
--- a/Game.Infrastructure.Tests/PreparedScriptTests.cs
+++ b/Game.Infrastructure.Tests/PreparedScriptTests.cs
@@ -1,0 +1,55 @@
+using Game.Infrastructure.Redis;
+using StackExchange.Redis;
+using Xunit;
+
+namespace Game.Infrastructure.Tests
+{
+    /// <summary>
+    /// Unit tests for <see cref="PreparedScript"/>, the helper that lets <c>RedisService</c>/<c>RedisQueue</c>
+    /// send a Lua script's hash instead of its full text on every call after the first (#2056). The behaviour
+    /// that matters and is otherwise only exercised indirectly through a live Redis: a cold call that fails
+    /// before it reaches the server must not mark the script warmed, or every later call would jump straight to
+    /// an <c>EVALSHA</c> the server has genuinely never cached.
+    /// <para>
+    /// Uses the same dead-endpoint technique as the sibling <see cref="RedisPubSubPublishFailureTests"/>: a
+    /// multiplexer pointed at an unreachable endpoint (<c>abortConnect=false</c>) connects lazily and then throws
+    /// a real <see cref="RedisConnectionException"/> from the awaited command — no mocking, no Redis server
+    /// required. The NOSCRIPT-triggered fallback itself needs a genuine server reply and is covered by the
+    /// existing Redis integration suite exercising these scripts end to end (RedisServiceIntegrationTests,
+    /// RedisQueueIntegrationTests) rather than duplicated here.
+    /// </para>
+    /// </summary>
+    public class PreparedScriptTests
+    {
+        private const string DeadEndpoint = "127.0.0.1:1,abortConnect=false,connectTimeout=500,connectRetry=0";
+
+        [Fact]
+        public async Task EvaluateAsync_WhenTheColdCallFails_LeavesTheScriptUnwarmed()
+        {
+            await using var multiplexer = await ConnectionMultiplexer.ConnectAsync(DeadEndpoint);
+            var db = multiplexer.GetDatabase();
+            var script = new PreparedScript("return 1");
+
+            await Assert.ThrowsAsync<RedisConnectionException>(() => script.EvaluateAsync(db, [], []));
+
+            // A failed cold call means Redis never actually cached this script; if it had wrongly been marked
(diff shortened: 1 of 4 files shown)


Game.Infrastructure/Redis/PreparedScript.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public RedisResult Evaluate(IDatabase db, RedisKey[] keys, RedisValue[] values, CommandFlags flags = CommandFlags.None)
+        {
+            if (!_warmed)
+            {
+                var coldResult = db.ScriptEvaluate(_script, keys, values, flags);
+                _warmed = true;
+                return coldResult;
+            }
+
```
ginderjeremiah: Should-fix (or a conscious decline): this cold branch marks `_warmed = true` as soon as `ScriptEvaluate` returns, but the callers pass `CommandFlags.FireAndForget`, so "returns" here means "handed to the send buffer", not "the server ran the EVAL and cached the script". If that one cold command is dropped (a connection blip at process start), the script is never cached server-side yet `_warmed` is now `true` — and unlike `EvaluateAsync`, this sync path has no `NOSCRIPT` fallback, so *every* subsequent `EVALSHA` for this script gets `NOSCRIPT` and silently no-ops for the rest of the process lifetime.

The doc comment frames this as "the same no-delivery-guarantee every other FireAndForget call in this tier already accepts", but that undersells the change: today a dropped fire-and-forget EVAL loses exactly one op and the next call works; here a dropped *cold* EVAL wedges the script permanently. The three sync callers are `ReclaimAndForget`, `HashSetAndForget`, and `HashSetIfExistsAndForget` — the hash-set writes carry real per-battle state, so a process-wide silent no-op is a more meaningful degradation than a single lost refresh.

Cheapest robust fix: issue the *cold* call without `FireAndForget` (a confirmed EVAL) so `_warmed` only flips once the script is genuinely cached; it's one blocking round-trip, once per process, on a path that's otherwise off the hot loop. Happy to hear if you'd rather accept the risk — just want it to be a deliberate call rather than an artifact of mirroring the async signature.

---

Game.Infrastructure.Tests/PreparedScriptTests.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            await Assert.ThrowsAsync<RedisConnectionException>(() => script.EvaluateAsync(db, [], []));
+
+            // A failed cold call means Redis never actually cached this script; if it had wrongly been marked
+            // warmed anyway, every later call would send only the hash and get NOSCRIPT forever.
+            Assert.False(script.IsWarmedForTesting);
+        }
+
+        [Fact]
+        public async Task EvaluateAsync_WhenAlreadyWarmed_SendsTheHashRatherThanRewarming()
```
ginderjeremiah: Nit: the name (`SendsTheHashRatherThanRewarming`) and the "warmed script never revisits the cold branch" comment promise more than the assertions can deliver — against a dead endpoint both the hash and the full-text branch throw `RedisConnectionException` before anything reaches the server, so the test can't actually observe *which* path was taken. What it really pins is "a warmed script stays warmed and doesn't flip back to cold", which is fine and worth keeping — just consider renaming to match (e.g. `WhenAlreadyWarmed_StaysWarmed`) so a future reader doesn't trust it to guard the hash-vs-text dispatch that only the integration suite actually exercises.

---

ginderjeremiah (commented):
## Review: EVALSHA-instead-of-EVAL for hot-path Redis Lua scripts (#2056)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't take a gating event on my own PR) — no blocking findings; two non-blocking items below.

This does exactly what the issue asked and the approach is well-judged. Verified independently:

- **All the actual call sites are converted.** Every `ScriptEvaluate`/`ScriptEvaluateAsync` in `RedisService.cs` and `RedisQueue.cs` now routes through `PreparedScript`, with none left behind. The issue's `GetSet` reference (`RedisService.cs:60-62`) is stale — `GetSet` is `StringGetSetExpiryAsync` now, not a Lua script — so its absence from the diff is correct, not a miss.
- **The hash/warm-state-as-`static readonly` reasoning holds.** `AddCache`/`AddPubSub` register the services `AddTransient`, so a per-instance field would recompute the SHA and re-warm per DI scope; the `static readonly` field shares it process-wide, and the multiplexer it targets is itself shared, so the server-side script cache lines up with the warm flag.
- **The cold→warm→NOSCRIPT-self-heal flow is correct.** Raw `KEYS[n]`/`ARGV[n]` text is sent verbatim by the `string` overload (no `@param` tokenization), so the client's `SHA1(UTF8(script))` matches the SHA the server caches under; the cold EVAL both runs and caches; `_warmed` flips only after a successful `await`; and the `RedisServerException` + `StartsWith("NOSCRIPT")` guard on `EvaluateAsync` self-heals a flushed cache.

### Should-fix

- `PreparedScript.Evaluate` (sync) — `PreparedScript.cs:81-85`. The cold branch flips `_warmed` on a `FireAndForget` send with no delivery confirmation and no `NOSCRIPT` fallback, so a dropped cold EVAL wedges that script into `NOSCRIPT` for the whole process rather than losing a single op. Details + a low-cost fix inline; fine to consciously decline, just want it to be a deliberate call.

### Nits

- `PreparedScriptTests.cs:41` — the second test's name/comment overpromise what a dead-endpoint assertion can verify. Inline.

### Test coverage

Appropriate for this tier. The one genuinely regression-prone invariant — a failed cold call must not mark the script warmed — is pinned as an in-process unit test, and the rest (real EVAL/EVALSHA dispatch, the `NOSCRIPT` self-heal which requires a server-side `SCRIPT FLUSH`) is correctly left to the existing Redis integration suites, consistent with `backend.md`'s rule that Redis-coupled logic and process-wide-infra-disruption branches are covered by integration tests / accepted as measure-only rather than unit-tested. No missing-coverage blocker.

### What's good

Genuinely clean optimization: the `LuaScript.Prepare`-can't-express-variable-ARGV constraint is correctly diagnosed and worked around by talking to `IDatabase` directly, the warm/cold/NOSCRIPT states are each reasoned through, and the XML docs capture the *why* (transient DI, fire-and-forget tradeoff) without straying into stale implementation detail.

---

Game.Infrastructure/Redis/PreparedScript.cs
ginderjeremiah: Fixed — the cold call now strips FireAndForget and blocks for a confirmed reply before flipping `_warmed`, so a dropped send can no longer wedge the script into permanent NOSCRIPT. Added a sync-path counterpart to the existing cold-call-failure unit test to pin it.

---

Game.Infrastructure.Tests/PreparedScriptTests.cs
ginderjeremiah: Renamed to `EvaluateAsync_WhenAlreadyWarmed_StaysWarmedAfterAFailedCall` and reworded the comment to say what it actually pins.

---

ginderjeremiah:
## Re-review — EVALSHA for hot-path Redis Lua scripts (#2056)

Since the initial review, commit `2cda6303` addressed both findings, and I verified the fixes directly:

- **Should-fix (sync `Evaluate` NOSCRIPT wedge) — fixed.** The cold branch now strips the caller's fire-and-forget flag (`flags & ~CommandFlags.FireAndForget`) and blocks for a confirmed reply before flipping `_warmed`, so a dropped cold send can no longer wedge a script into permanent `NOSCRIPT`. One blocking round trip, once per process, off the hot loop — exactly the recommended fix. The class doc now captures the *why*.
- **Nit (test name/comment overpromise) — fixed.** Renamed to `EvaluateAsync_WhenAlreadyWarmed_StaysWarmedAfterAFailedCall` with a comment that states what a dead-endpoint assertion can actually pin, and a sync-path counterpart (`Evaluate_WhenTheColdCallFails_LeavesTheScriptUnwarmed`) was added to cover the newly-diverged blocking cold path.

Both review threads are resolved. CI is fully green (all 9 checks pass). No outstanding nits or should-fixes and no follow-up work — clean optimization, good to merge.

_(Automated re-review pass.)_

---

PR #2122, opened 2026-07-18
Collapse the two content-graph readers and record types

### LINKED ISSUE #2058: The 12-set content catalogue is hand-enumerated in six places
## Context

Found in the July 2026 health review (Game.Application pass).

Adding a 13th reference set requires ~6 coordinated edits:

- `Game.Application/Content/ContentExporter.cs:60-74` (filenames + Canonicalize calls)
- `ContentImportReader.cs:18-32` (filenames + types, for the seeder)
- `ContentGraphJsonReader.cs:15-27` (the same filenames + types again, for the CI lint — this and ContentImportReader are essentially the same reader twice, sync vs async, over two near-identical target records)
- `ContentHealthService.cs:60-81` (the same graph assembled from live caches)
- `ContentGraph.cs` and `Game.Abstractions.Content.ContentImport` (two parallel 12-field records)

Mitigating factors: the record constructors make most omissions compile errors, and a filename mismatch fails loudly at seed/lint — so this is boilerplate cost rather than correctness risk.

## Suggested fix

A single per-set descriptor (name, contract type, cache accessor, canonicalizer) consumed by all the call sites, collapsing the two readers into one.

ginderjeremiah:
## Summary

`ContentImport` (`Game.Abstractions/Content`) and `ContentGraph` (`Game.Application/Content`) were two parallel 12-field records describing the same static content graph, and `ContentImportReader`/`ContentGraphJsonReader` were two nearly-identical readers deserializing the exact same 12 JSON files into them (sync-for-the-seeder vs. async-for-the-CI-lint). `ContentHealthService` also duplicated `ContentExporter`'s live-reference-cache assembly wholesale. Together these were 6 of the places a 13th reference set would need touching.

- **Merged the two record types into one `ContentGraph`**, now living in `Game.Abstractions/Content` (the layer both `IContentSeeder`/`IContentImportReader` and `IProgressionGraphChecker`/`ContentHealthService` can reach). It keeps `ContentImport`'s `required`-init-property style, which the compiler enforces at every construction site.
- **Deleted `ContentGraphJsonReader`** — the CI progression-graph lint (`ContentGraphLintTests`) now reads through the same `ContentImportReader` the startup seeder already used; both read identical files into an identical shape, so there was never a reason for two readers. (The lint's two test methods no longer need to be `async` now that the underlying read is synchronous.)
- **Extracted `LiveContentGraphAssembler`** (`Game.Application/Content`) so `ContentExporter` and `ContentHealthService` share the one place that builds a `ContentGraph` from the live reference caches (the tag-materialization-from-DB-stream plus the 11 cache accessor calls), instead of each re-implementing it.

Net effect: adding a 13th reference set now touches 3 typed call sites (the exporter's per-field canonicalize/serialize, the shared live-graph assembler, and the shared JSON reader) instead of 6, and there's a single source of truth for "the whole static content graph" shape.

I deliberately did **not** go further into a fully generic, reflection-driven per-set descriptor table (name + type + accessor + canonicalizer in one array walked by a single loop) — the per-field `Canonicalize`/`Serialize` calls in `ContentExporter` are each strongly-typed (mutating child-collection order per contract type), and a project that leans on the compiler catching a missed field (as this issue's own "mitigating factors" note) is better served by 3 explicit, typed call sites than a `Dictionary<string, object>`-style descriptor loop. Happy to revisit if a 13th set actually lands and the boilerplate still stings.

## Test plan

- [x] `dotnet build Game.sln` — 0 warnings/errors
- [x] `dotnet test Game.sln` — all four backend test projects pass (Core 1371, Infrastructure 66, Application 998, Api 823)
- [x] `dotnet format Game.sln --verify-no-changes` — clean

Closes #2058

---
_Generated by [Claude Code](https://claude.ai/code/session_01LCUMN2eJmfoW5cwAWaZ41f)_


ginderjeremiah (commented):
## Review: Collapse the two content-graph readers and record types (#2058)

**Verdict:** Approve — clean, well-scoped tech-debt refactor that cuts the hand-enumerated set count from 6 sites to 3, with a single source of truth for the content graph and no behavioral change. (Posting as `COMMENT` because I'm the PR author; treat it as an approval.)

The change does exactly what #2058 asked without over-reaching, and I agree with the deliberate decision *not* to build the fully generic reflection-driven descriptor table — the per-field `Canonicalize`/`Serialize` calls in `ContentExporter` are strongly typed, and a project that leans on the compiler catching a missed field is better served by 3 explicit typed call sites than a descriptor loop. That's the right call, consistent with critically evaluating the issue's suggested fix rather than adopting it wholesale.

Verified:
- `ContentGraph` (now in `Game.Abstractions/Content`) keeps the `required`-init style, so every construction site stays compiler-enforced. Merging the two parallel records is a strict simplification.
- Deleting `ContentGraphJsonReader` in favor of `ContentImportReader.Read` is behavior-preserving for the lint: both go through the same `ContentExportSerializer.Deserialize<T>`, read the same 12 files from `RepoPaths.ContentDirectory()`, into the same shape. The only delta is `ReadSet`'s explicit `File.Exists` check, which yields a clearer error — a strict improvement.
- `LiveContentGraphAssembler` genuinely dedupes the tags-from-DB materialization + 11 cache accessors that `ContentExporter` and `ContentHealthService` each re-implemented. Exporter file order (byte-checked by the drift guard) is preserved by the explicit `File(...)` list.
- Dead code cleaned up: the now-unused `Contracts` usings were removed from `ContentExporter`/`ContentHealthService`, and `docs/backend-content.md` was updated to describe the single shared type.

### Test coverage

Adequate for a pure refactor — no new logic to cover. The exporter's use of the assembler is round-tripped by `ContentExportDriftGuardTests`, the live-cache path by `ContentHealthServiceTests`, the reader by `ContentSeederTests`/`ContentGraphLintTests`, and the checker by `ProgressionGraphCheckerTests` (all updated to the merged type via named args, so no silent positional mismatch).

### What's good

Tight, honest refactor with a clear explanation of the scope call in the PR body. Docs kept in sync. No findings.


---

PR #2123, opened 2026-07-18
Convert non-reactive hot-path engine internals to #-private

### LINKED ISSUE #2059: Convert non-reactive hot-path internals to #-private so statify stops deep-proxying them
## Context

`statify` enumerates plain properties (`statify.svelte.ts:31`), so TS-`private` fields (plain enumerable props at runtime) get wrapped in `$state` deep proxies; only `#`-private escapes. `PlayerManager` already uses `#lockedBaseCache` deliberately *because* "the cache is `#`-private so `statify` leaves it non-reactive" (`player-manager.ts:79-84`) — but the other managers don't follow the convention:

- `BattleEngine` (`battle-engine.ts:105,116-127,136-141`): `private rng`, `private readonly stepLog`, `private readonly effectDamage`, the `last*` change-detection fields — all statified because `battleEngine` is wrapped (`engine.ts:61`). `stepLog`/`effectDamage` are plain object literals written on **every 40 ms tick** (and up to ~3,000 ticks synchronously in `replayToOffset`), documented as a "reused per-tick sink" precisely to avoid per-tick cost — the proxying re-adds it. They're also semantically non-reactive: a future getter writing them from a `$derived` hits the illegal-mutation trap.
- `EnemyManager`'s generation counters/backoff handles (`enemy-manager.ts:74-112`) and `InventoryManager`'s `queue`/`generation` (`inventory-manager.ts:76-97`) — same (milder) inconsistency. (InventoryManager's *caches* must stay reactive — screens derive from `equipmentStats`.)

## Suggested fix

Convert the internal-only fields to `#`-private, matching the PlayerManager convention. Mechanical, no behavior change.

ginderjeremiah:
## Summary

`statify` (`$lib/common/statify.svelte.ts`) enumerates plain properties via `for...in`, so TS-`private` fields (plain enumerable props at runtime) get deep-proxied into `$state`; only `#`-private fields escape. `PlayerManager` already follows this convention deliberately for `#lockedBaseCache`, but three other statified managers didn't:

- **`BattleEngine`** (`battle-engine.ts`): `rng`, `stepLog`, `effectDamage`/`effectDamageElapsedMs`, and the `last*` change-detection fields (`lastEquipmentStats`, `lastPlayerAttributes`, `lastSelectedSkills`, `lastPlayerLevel`, `lastLockedBaseModifiers`, `lastProficiencyModifiers`). `stepLog`/`effectDamage` in particular are documented as a "reused per-tick sink" written on every 40ms tick (and up to ~3,000 ticks synchronously in `replayToOffset`) specifically to avoid per-tick allocation cost — the proxying was silently re-adding that cost.
- **`EnemyManager`** (`enemy-manager.ts`): the transition/fetch generation counters, re-entrancy/backoff guards, and related internals (`transitioning`, `transitionGeneration`, `fetchingEnemy`, `fetchGeneration`, `fetchingGeneration`, `inFlightFetch`, `cancelBackoff`, `lastSyncedAutoChallengeBoss`, `isFirstFetch`, `battleStageUnhook`).
- **`InventoryManager`** (`inventory-manager.ts`): `queue` and `generation` only — its `equipmentStatsCache`/`grantedSkillIdsCache` are left as TS-`private` and stay reactive on purpose, since screens derive from `equipmentStats`.

All are mechanical renames (`private field` → `#field`), no behavior change. Verified none of these field names are referenced outside their owning class (only a couple of unrelated matches in `loom-game.ts`, a different class, and a comment in a test file that names the concept, not the symbol).

Also adds a regression test to `statify.svelte.test.ts` pinning the exact contract this issue relies on: a `#`-private field is never wrapped by `statify`, while a same-shaped public field is.

Closes #2059

## Test plan

- [x] `npx vitest run` on the affected engine/statify test files — all passing
- [x] `npm run test:unit:ci` (full frontend suite) — 299 files / 3765 tests passing (was 3764; added one new statify regression test)

https://claude.ai/code/session_01V7DXWmrF3ewj6XZs4D6RqD

---
_Generated by [Claude Code](https://claude.ai/code/session_01V7DXWmrF3ewj6XZs4D6RqD)_

### DIFF
--- a/UI/src/lib/engine/battle/battle-engine.ts
+++ b/UI/src/lib/engine/battle/battle-engine.ts
@@ -101,8 +101,9 @@ export class BattleEngine {
 	public running = false;
 
 	/** The seeded battle RNG, re-created from each enemy's seed in {@link reset} so the live battle draws the
-	 *  crit/parry/dodge rolls from the same stream the backend replays. Seeded to 0 until the first reset. */
-	private rng = new Mulberry32(0);
+	 *  crit/parry/dodge rolls from the same stream the backend replays. Seeded to 0 until the first reset.
+	 *  `#`-private so `statify` leaves it non-reactive — it's a hot-path internal read only imperatively. */
+	#rng = new Mulberry32(0);
 
 	private logicalUnhook?: Action;
 	private renderUnhook?: Action;
@@ -112,8 +113,10 @@ export class BattleEngine {
 	private finishLoading?: Action;
 
 	/** Reused per-tick sink the shared battleStep populates with this tick's newly-applied effects and
-	 *  DoT/HoT amounts, so the engine can turn them into combat-log lines (frontend-only). */
-	private readonly stepLog: BattleStepLog = {
+	 *  DoT/HoT amounts, so the engine can turn them into combat-log lines (frontend-only). `#`-private so
+	 *  `statify` leaves it non-reactive — it's a reused sink written on every 40ms tick, and deep-proxying
+	 *  it would re-add the per-tick allocation cost it exists to avoid. */
+	readonly #stepLog: BattleStepLog = {
 		appliedEffects: [],
 		enemyDotDamage: 0,
 		playerDotDamage: 0,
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Convert non-reactive hot-path engine internals to `#`-private (closes #2059)

**Verdict:** Approve (posted as a comment since I authored the PR) — a correct, well-scoped mechanical change that faithfully applies the existing `PlayerManager` `#lockedBaseCache` convention to the three other statified managers, with no behavior change and a regression test pinning the contract.

### Verification
- Confirmed the premise against `statify.svelte.ts:31` — `for...in` enumerates TS-`private` fields (plain enumerable props at runtime) and deep-proxies them into `$state`; only `#`-private escapes. This matches the utility's own documented escape hatch (`statify.svelte.ts:18-21`) and the deliberate `PlayerManager` precedent (`player-manager.ts:78-84`).
- Grepped every converted field across `UI/src` — no stale `this.<field>` references remain in the owning classes, and no cross-class reader touches any of the renamed fields. The rename is complete and self-contained.
- The scoping decisions are right: `InventoryManager`'s `equipmentStatsCache`/`grantedSkillIdsCache` are correctly left TS-`private` (screens derive from `equipmentStats`, so they must stay reactive), and only the genuinely imperative `queue`/`generation` were converted.

### Test coverage
- The new `statify.svelte.test.ts` case pins exactly the contract this change relies on: a `#`-private field never gets a `$state` accessor (a `$derived` reading it through a public getter never re-runs), while a same-shaped public field would. That's the right thing to lock down, and it's the correct kind of test for a "no behavior change" refactor. No test gaps — the converted fields are internal hot-path state with no independent behavior to assert beyond this contract.

### What's good
- The added JSDoc on each converted field explains *why* it's `#`-private (hot-path/per-tick sink, read only imperatively), which is genuinely useful context that prevents a future edit from naively reverting to `private` and silently re-adding per-tick proxy allocation cost. The detail is warranted here.

No blocking, should-fix, or nit findings. Holding the merge only until CI (e2e + `test-frontend`) goes green.

---

PR #2141, opened 2026-07-18
Dedupe Workbench attribute-table sections and save-flash plumbing

### LINKED ISSUE #2061: Small frontend DRY wins: duplicated attribute-table section configs; copy-pasted save-flash plumbing
## Context

Found in the July 2026 health review (Workbench + game screens passes). Two small, independent extractions batchable in one PR:

1. **Duplicated Workbench attribute-table sections:** `entities/enemy.ts:88-122` (attribute distribution) ≡ `entities/class.ts:207-241`, and `entities/item.ts:140-172` (attribute bonuses) ≡ `entities/item-mod.ts:76-108` — identical `newRow`/columns/labels modulo copy. `tags-section.ts` already proves the shared-section-factory pattern; an `attributeDistributionSection&lt;T&gt;()` / `attributeBonusSection&lt;T&gt;()` in `entities/` would DRY four configs.

2. **Copy-pasted save-flash plumbing:** `attributes-view.svelte.ts:301,308,501-515` and `options-view.svelte.ts:157-162,281-295` carry an identical `saved = $state(false)` + `#flashTimer` + `flashSaved()` (same 1900 ms literal) + `dispose()` block, each requiring the owning screen to remember `onDestroy(() =&gt; view.dispose())`. A tiny shared `SaveFlash` reactive class in `$lib/common` removes the duplication and centralizes the timer lifecycle. (The broader deltas/overrides resync-pinning machinery in those two views is also structurally parallel, but extracting that is a larger, riskier refactor — out of scope here.)

ginderjeremiah:
## Summary

Two small, independent DRY extractions from the July 2026 health review (#2061):

1. **Duplicated Workbench attribute-table sections.** `entities/enemy.ts` (`attributeDistribution`) and `entities/class.ts` (`attributeDistributions`) defined identical "stat distribution per level" table sections (attributeId + baseAmount + amountPerLevel); `entities/item.ts` and `entities/item-mod.ts` defined identical "flat stat bonus" table sections (attributeId + amount). Extracted `attributeDistributionSection<T>()`/`attributeBonusSection<T>()` factories in a new `entities/attribute-table-sections.ts`, mirroring the existing `tagsSection` pattern (curried config, entity-agnostic via `fieldsOf`). All four entity configs now call the shared factory instead of hand-copying the section.

2. **Copy-pasted save-flash plumbing.** `attributes-view.svelte.ts` and `options-view.svelte.ts` each carried an identical `saved = $state(false)` + `#flashTimer` + `flashSaved()` (1900ms) + `dispose()` block. Extracted a `SaveFlash` reactive class (`$lib/common/save-flash.svelte.ts`) that owns the flag and the timer lifecycle; both view-models now delegate to it via a `saved` getter, preserving the existing public API.

   While implementing this, I found the **exact same duplicated pattern** in two more places — `EntityStore` and `ProgressionStore` (the Workbench's own save orchestration stores) — which weren't named in the issue. Since it's the same mechanical fix using the tool built for this PR, I folded those in too rather than leaving the dedup half-finished or filing a separate issue for an already-solved duplication. `EntityStore.saved` keeps a getter **and** setter (one existing test writes to it directly to arrange state), while the other three consumers only needed a getter.

## Test plan

- [x] `npm run test:unit:ci` — all 3776 tests pass, including new unit tests for `SaveFlash` and the `attribute-table-sections` factories
- [x] `npm run build` — succeeds

Closes #2061

---
_Generated by [Claude Code](https://claude.ai/code/session_019d1ADCSFg2Ft8ggyRdhv5U)_

### DIFF
--- a/UI/src/lib/common/index.ts
+++ b/UI/src/lib/common/index.ts
@@ -20,3 +20,4 @@ export * from './hooks';
 export * from './event-measuring';
 export * from './serialized-queue';
 export * from './statify.svelte';
+export * from './save-flash.svelte';
--- a/UI/src/lib/common/save-flash.svelte.ts
+++ b/UI/src/lib/common/save-flash.svelte.ts
@@ -1,0 +1,31 @@
+/** Brief "saved" confirmation flash with a self-owned timer lifecycle — the pattern duplicated
+ *  across the Attributes/Options draft-edit view-models before being extracted here. A consumer
+ *  calls {@link flash} after a successful save and must call {@link dispose} on teardown to clear
+ *  a pending timer. */
+export class SaveFlash {
+	active = $state(false);
+
+	#timer: ReturnType<typeof setTimeout> | undefined;
+
+	constructor(private readonly durationMs = 1900) {}
+
+	flash(): void {
+		this.active = true;
+		if (this.#timer) {
+			clearTimeout(this.#timer);
+		}
+		this.#timer = setTimeout(() => {
+			this.active = false;
+		}, this.durationMs);
+	}
+
+	reset(): void {
+		this.active = false;
+	}
+
+	dispose(): void {
+		if (this.#timer) {
+			clearTimeout(this.#timer);
+		}
+	}
+}
--- a/UI/src/routes/admin/workbench/entities/attribute-table-sections.ts
+++ b/UI/src/routes/admin/workbench/entities/attribute-table-sections.ts
@@ -1,0 +1,107 @@
+import { reference } from '../reference.svelte';
+import { firstFree } from './helpers';
+import { fieldsOf, type Identified, type TableSectionConfig } from './types';
+
+interface AttributeDistributionRow {
+	attributeId: number;
+	baseAmount: number;
+	amountPerLevel: number;
+}
+
+interface AttributeBonusRow {
+	attributeId: number;
+	amount: number;
+}
+
+/**
+ * Shared "stat distribution per level" table — identical shape for Enemies and Classes
+ * (attributeId + baseAmount + amountPerLevel, one row per attribute); only the collection
+ * key, section key, description and empty-state copy vary per entity.
+ */
+export const attributeDistributionSection = <T extends Identified>(config: {
+	key: string;
+	itemsKey: keyof T & string;
+	desc: string;
+	emptySub: string;
+}): TableSectionConfig<T> => {
+	const rowsOf = (rec: T) => fieldsOf(rec)[config.itemsKey] as AttributeDistributionRow[];
+	return {
+		key: config.key,
+		label: 'Attributes',
+		glyph: 'bars',
+		desc: config.desc,
+		count: (rec) => rowsOf(rec).length,
+		warn: (rec) => (rowsOf(rec).length ? null : 'No attribute distribution'),
+		kind: 'table',
+		itemsKey: config.itemsKey,
+		rowKey: 'attributeId',
+		addLabel: 'Add attribute',
+		emptyIcon: 'bars',
+		emptyTitle: 'No attributes set',
+		emptySub: config.emptySub,
+		newRow: (rec) => ({
+			attributeId: firstFree(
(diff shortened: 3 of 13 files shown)


ginderjeremiah (commented):
## Review: Dedupe Workbench attribute-table sections and save-flash plumbing (#2061)

**Verdict:** Approve — faithful, well-tested DRY extraction that matches the issue exactly and sensibly folds in two more instances of the same duplication.

_(Submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author; the intent is Approve.)_

### What's good

- **Both extractions match the issue's request precisely.** `attributeDistributionSection` / `attributeBonusSection` mirror the established `tagsSection` / `chipsSection` factory pattern (curried, entity-agnostic via `fieldsOf`), and the extracted configs are byte-for-byte faithful to the four originals — same `min` (190 dist / 200 bonus), `warn` copy, `addLabel`, column shape, and `newRow`/`firstFree` behavior. Only the genuinely-varying bits (`key`, `desc`, `emptySub`) are parameterized.
- **Folding in `EntityStore` / `ProgressionStore` was the right call.** They carried the identical `saved`/`#flashTimer`/`flashSaved()`/`dispose()` block; fixing them with the tool built for this PR beats leaving the dedup half-done. The `EntityStore.saved` getter+setter (vs getter-only elsewhere) is justified and documented — an existing test arranges state by writing it.
- **`SaveFlash` preserves the exact prior semantics:** re-arm clears the prior timer, `dispose()` cancels the pending auto-clear without flipping the flag (matches the old teardown behavior), and the 1900 ms window is unchanged. The public `saved` API on all four consumers is preserved.
- **Test coverage is solid.** New unit tests cover `SaveFlash` (activate/auto-clear, re-arm-restarts-timer, reset, dispose-cancels) and both section factories (config passthrough, count/warn, `newRow` first-free selection). This meets the frontend testing rule (lib + workbench modules under `tests/` mirroring source).

No blocking findings, no should-fixes, no nits. Clean refactor with no behavior change.


---

ginderjeremiah:
## Review — ✅ Approve

Reviewed against the scope of #2061. This is a clean, faithful DRY extraction with no behavioral change.

### What I verified
- **Attribute-table factories** (`attribute-table-sections.ts`) reproduce the four inline configs exactly. Detail preserved carefully: distribution columns keep `min: 190` and bonus columns keep `min: 200`; `label`/`glyph`/`emptyIcon` constants match; `desc` is parameterized only for distribution (enemy vs. class copy genuinely differ) and hard-coded for bonus (item and item-mod shared identical copy). Section `key` is config-driven for distribution (`attrs`/`attributes`) and fixed for bonus, matching the originals. Follows the established `tagsSection`/`chipsSection` factory pattern and stays entity-agnostic via `fieldsOf`.
- **`SaveFlash`** centralizes the `active` flag + 1900ms timer lifecycle. `flash()` re-arms (clears prior timer), `reset()` clears the flag without touching the pending timer, and `dispose()` cancels it — matching the previous per-view semantics precisely (a stale timer firing after `reset()` is a harmless no-op against an already-false flag, same as before).
- **Folding in `EntityStore` + `ProgressionStore`** was the right call — same mechanical duplication, same tool. The getter-only vs. getter+setter asymmetry is justified: `EntityStore` mutators (and one test) write `saved` directly, so it keeps a setter; the other three consumers only read it. I confirmed via grep that no external code writes `.saved` on the getter-only consumers, so the change is safe. Svelte 5 reactivity propagates correctly through the getter since `active` is `$state`.
- **Tests**: new coverage for both `SaveFlash` (timer, re-arm, reset, dispose) and the two factories. Ran the affected suites locally — 41/41 pass. All CI checks green (frontend unit + e2e across chromium/firefox/webkit; backend jobs correctly skipped).

### Nits
None blocking. The only stylistic wrinkle is that `attributeDistributionSection` uses a single-call config-object signature while the sibling `chipsSection`/`tagsSection` factories use currying — but the config-object shape reads more cleanly here given the number of varying fields, so I'd keep it.

Merging.

---

PR #2142, opened 2026-07-18
Fix fire-and-forget Lua scripts wedging into silent NOSCRIPT after cache clear

### LINKED ISSUE #2126: Warmed fire-and-forget Lua scripts wedge into silent NOSCRIPT failures after a Redis restart — progress cache and presence refresh degrade for the process lifetime
## Problem

The #2120 EVALSHA optimization (#2056) left the **fire-and-forget** script path with no `NOSCRIPT` recovery. `PreparedScript.Evaluate` (`Game.Infrastructure/Redis/PreparedScript.cs:80-94`) sends only `EVALSHA` once `_warmed` flips, preserving the caller's `FireAndForget` flag — and there is no fallback, no `ConnectionRestored` hook anywhere, and `_warmed` is never reset. StackExchange.Redis silently discards error replies on fire-and-forget commands, so after Redis loses its script cache (restart, failover, `SCRIPT FLUSH`), **every** subsequent call to the three fire-and-forget scripts fails silently until the API process restarts.

The class doc rationalizes this as matching "the single lost op a normal fire-and-forget command risks", but that analysis doesn't hold: the awaited scripts self-heal via the `NOSCRIPT` catch in `EvaluateAsync` (:74-77); the fire-and-forget ones fail *persistently*, not once.

## Affected operations and consequences

- `RedisService.ReclaimAndForget` (`Game.Infrastructure/Cache/Redis/RedisService.cs:108`) — the per-message presence reclaim/TTL refresh (`SocketManagerService.cs:257`) no-ops, so presence keys expire mid-session and single-session/takeover enforcement silently degrades.
- `HashSetAndForget` (:215) / `HashSetIfExistsAndForget` (:235) — `PlayerProgressRepository.cs:136`'s cache advance no-ops, so the Redis progress hash (the documented source of truth) permanently stops advancing while durable events keep flowing to Postgres. Battle completion re-reads progress per battle, so later reads serve stale rows and the next save enqueues **absolute-value** `ProgressUpdated` events computed from that stale base — permanently regressing statistics/challenges/proficiencies in the DB. The miss-reload at :220 can also never recreate the hash key after a Redis data loss, so every battle pays the 3-query DB reload *and* races the write-behind drain.

## Why tests never see it

The integration suites `FLUSHDB` between tests (`Game.TestInfrastructure/Helpers/RedisCleaner.cs:13`), which clears data but not the script cache. `PreparedScriptTests.cs:19-22` states the NOSCRIPT fallback "is covered by the existing Redis integration suite", but only the *awaited* scripts have a fallback to exercise — no test issues `SCRIPT FLUSH` against a warmed fire-and-forget script.

## Fix options

1. Simply keep full-text `EVAL` for the three fire-and-forget scripts — they're ~100–300 bytes, so #2056's win is marginal there (smallest change).
2. Reset warm state on `IConnectionMultiplexer.ConnectionRestored` (needs a small script registry).
3. Issue an awaited `SCRIPT LOAD` at connect.

Whichever fix lands should be pinned by an integration test that flushes the script cache mid-run and asserts the fire-and-forget operation still takes effect.

ginderjeremiah:
## Summary

- `PreparedScript.Evaluate` (the sync, fire-and-forget-capable path used by `ReclaimAndForget`, `HashSetAndForget`, and `HashSetIfExistsAndForget`) had no way to self-heal from a `NOSCRIPT` reply the way the awaited `EvaluateAsync` path does, because StackExchange.Redis discards the reply for a `CommandFlags.FireAndForget` command before any retry logic can run.
- Simply always sending the full script text turned out **not** to be enough on its own: StackExchange.Redis maintains its own transparent, per-connection "once seen, always send the hash" cache (`ScriptEvalMessage.GetMessages`) that silently substitutes `EVALSHA` regardless of what our code passes, and its own `NOSCRIPT` self-heal only fires from inside the reply processor — which a fire-and-forget command never reaches. I verified this against a live Redis container before landing the fix; a naive "just send `_script`" version still silently dropped the write after a `SCRIPT FLUSH`.
- The real fix is `CommandFlags.NoScriptCache`, a documented StackExchange.Redis flag that forces a literal `EVAL` and opts the call out of that library-level cache entirely. `PreparedScript.Evaluate` now always ORs this flag in.
- Updated `PreparedScript`'s class doc to explain the two code paths' very different cache-trust guarantees, and removed the now-obsolete unit test that pinned `Evaluate`'s old warmed-state behavior (that method no longer touches `_warmed` at all).

## How this addresses #2126

Closes #2126. The issue listed three triggers for the server losing its script cache — a Redis restart, a failover, or a manual `SCRIPT FLUSH` — and all three are now covered uniformly, since the fix never depends on the script being cached server-side in the first place, rather than trying to detect and recover from the loss after the fact.

## Test plan

- [x] Added `RedisServiceIntegrationTests.HashSetAndForget_AfterScriptCacheIsFlushedMidRun_StillTakesEffect`, which flushes the Redis script cache mid-test (via an `AllowAdmin` connection) and asserts a subsequent `HashSetAndForget` call still lands — this is the integration test the issue asked for.
- [x] `dotnet build` — clean.
- [x] `dotnet test` (full solution, against the local Postgres/Redis containers) — 3,261 passed, 0 failed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HFKEi44US8wiThVrhFjpZ7)_

### DIFF
--- a/Game.Infrastructure.Tests/PreparedScriptTests.cs
+++ b/Game.Infrastructure.Tests/PreparedScriptTests.cs
@@ -12,7 +12,7 @@ namespace Game.Infrastructure.Tests
     /// server has genuinely never cached. <see cref="PreparedScript.Evaluate"/> (the fire-and-forget-capable sync
     /// path) has no such state to pin — it always sends the full script text (#2126) — so it isn't covered here;
     /// its NOSCRIPT-survival behaviour is covered by the integration suite instead
-    /// (RedisServiceIntegrationTests.HashSetAndForget_AfterScriptCacheIsFlushed_StillTakesEffect).
+    /// (RedisServiceIntegrationTests.HashSetAndForget_AfterScriptCacheIsFlushedMidRun_StillTakesEffect).
     /// <para>
     /// Uses the same dead-endpoint technique as the sibling <see cref="RedisPubSubPublishFailureTests"/>: a
     /// multiplexer pointed at an unreachable endpoint (<c>abortConnect=false</c>) connects lazily and then throws
--- a/Game.Infrastructure.Tests/PreparedScriptTests.cs
+++ b/Game.Infrastructure.Tests/PreparedScriptTests.cs
@@ -5,13 +5,14 @@ using Xunit;
 namespace Game.Infrastructure.Tests
 {
     /// <summary>
-    /// Unit tests for <see cref="PreparedScript"/>, the helper that lets <c>RedisService</c>/<c>RedisQueue</c>
-    /// send a Lua script's hash instead of its full text on every call after the first (#2056). The behaviour
-    /// that matters and is otherwise only exercised indirectly through a live Redis: a cold call that fails
-    /// before it reaches the server must not mark the script warmed, or every later call would jump straight to
-    /// an <c>EVALSHA</c> the server has genuinely never cached — pinned for both <see cref="PreparedScript.EvaluateAsync"/>
-    /// and <see cref="PreparedScript.Evaluate"/>, since the latter blocks its own cold call rather than sharing
-    /// the former's fallback logic.
+    /// Unit tests for <see cref="PreparedScript"/>, the helper that lets an awaited caller send a Lua script's
+    /// hash instead of its full text on every call after the first (#2056). The behaviour that matters and is
+    /// otherwise only exercised indirectly through a live Redis: a cold call that fails before it reaches the
+    /// server must not mark the script warmed, or every later call would jump straight to an <c>EVALSHA</c> the
+    /// server has genuinely never cached. <see cref="PreparedScript.Evaluate"/> (the fire-and-forget-capable sync
+    /// path) has no such state to pin — it always sends the full script text (#2126) — so it isn't covered here;
+    /// its NOSCRIPT-survival behaviour is covered by the integration suite instead
+    /// (RedisServiceIntegrationTests.HashSetAndForget_AfterScriptCacheIsFlushed_StillTakesEffect).
     /// <para>
     /// Uses the same dead-endpoint technique as the sibling <see cref="RedisPubSubPublishFailureTests"/>: a
     /// multiplexer pointed at an unreachable endpoint (<c>abortConnect=false</c>) connects lazily and then throws
@@ -36,21 +37,6 @@ namespace Game.Infrastructure.Tests
 
             // A failed cold call means Redis never actually cached this script; if it had wrongly been marked
             // warmed anyway, every later call would send only the hash and get NOSCRIPT forever.
-            Assert.False(script.IsWarmedForTesting);
-        }
-
-        [Fact]
-        public async Task Evaluate_WhenTheColdCallFails_LeavesTheScriptUnwarmed()
-        {
-            // Evaluate's cold branch strips a caller's FireAndForget flag and blocks for a confirmed reply, so
-            // this is pinned separately from EvaluateAsync's version of the same invariant: a dropped cold send
-            // must not wedge the script into the hash-only path.
-            await using var multiplexer = await ConnectionMultiplexer.ConnectAsync(DeadEndpoint);
-            var db = multiplexer.GetDatabase();
-            var script = new PreparedScript("return 1");
-
-            Assert.Throws<RedisConnectionException>(() => script.Evaluate(db, [], [], CommandFlags.FireAndForget));
-
             Assert.False(script.IsWarmedForTesting);
         }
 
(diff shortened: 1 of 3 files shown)


Game.Infrastructure.Tests/PreparedScriptTests.cs
```diff
@@ -5,13 +5,14 @@
...
-    /// and <see cref="PreparedScript.Evaluate"/>, since the latter blocks its own cold call rather than sharing
-    /// the former's fallback logic.
+    /// Unit tests for <see cref="PreparedScript"/>, the helper that lets an awaited caller send a Lua script's
+    /// hash instead of its full text on every call after the first (#2056). The behaviour that matters and is
+    /// otherwise only exercised indirectly through a live Redis: a cold call that fails before it reaches the
+    /// server must not mark the script warmed, or every later call would jump straight to an <c>EVALSHA</c> the
+    /// server has genuinely never cached. <see cref="PreparedScript.Evaluate"/> (the fire-and-forget-capable sync
+    /// path) has no such state to pin — it always sends the full script text (#2126) — so it isn't covered here;
+    /// its NOSCRIPT-survival behaviour is covered by the integration suite instead
+    /// (RedisServiceIntegrationTests.HashSetAndForget_AfterScriptCacheIsFlushed_StillTakesEffect).
```
ginderjeremiah: Nit: stale test name. The referenced test is `HashSetAndForget_AfterScriptCacheIsFlushedMidRun_StillTakesEffect` (note `MidRun`), but this doc references `HashSetAndForget_AfterScriptCacheIsFlushed_StillTakesEffect` — so a reader searching for the named test won't find it. Worth correcting to keep the cross-reference accurate.

---

ginderjeremiah (commented):
## Review: Fix fire-and-forget Lua scripts wedging into silent NOSCRIPT after cache clear (#2126)

**Verdict:** Approve *(posted as a comment since GitHub won't accept a gating verdict from the PR author identity)* — the fix correctly solves the issue, is well-tested, and leaves no dead code. One doc nit below.

### What's good

- **Correct root-cause fix.** The PR goes past the issue's "just keep full-text `EVAL`" suggestion and correctly identifies that a naive "always send `_script`" fix is *insufficient*, because StackExchange.Redis maintains its own per-connection EVALSHA-substitution cache (`ScriptEvalMessage`) whose NOSCRIPT self-heal only runs in the reply processor a fire-and-forget command never reaches. `CommandFlags.NoScriptCache` is the documented, correct mechanism to force a literal `EVAL` and opt out of that library-level cache. The empirical verification against a live Redis (naive fix still dropped the write after `SCRIPT FLUSH`) is exactly the right rigor for a subtle library-behavior bug.
- **Clean removal of dead code.** `Evaluate` no longer touches `_warmed`/`_hash`, so removing `Evaluate_WhenTheColdCallFails_LeavesTheScriptUnwarmed` is correct — that invariant no longer exists for this path. `_warmed`/`_hash` remain legitimately used by `EvaluateAsync`.
- **Test covers the exact requested scenario.** `HashSetAndForget_AfterScriptCacheIsFlushedMidRun_StillTakesEffect` warms the script, issues a real `SCRIPT FLUSH` via an `AllowAdmin` connection mid-test, and asserts a subsequent fire-and-forget call still lands — precisely the integration test the issue asked for, and one that would fail under both the old code and a naive fix. All three fire-and-forget scripts share the single `Evaluate` path, so one representative test is reasonable.
- **Documentation.** The class doc now clearly explains the differing cache-trust guarantees of the two paths and *why* `NoScriptCache` is required rather than optional. This is the kind of "deliberate-looking construct a later edit might naively simplify" that genuinely warrants the longer comment — appropriate, not over-documented.

### Nits

- `Game.Infrastructure.Tests/PreparedScriptTests.cs:15` — the doc references `RedisServiceIntegrationTests.HashSetAndForget_AfterScriptCacheIsFlushed_StillTakesEffect`, but the actual test is `...FlushedMidRun_StillTakesEffect`. Stale cross-reference (inline comment left).

### Test coverage

Meets the project's testing rules: the fire-and-forget NOSCRIPT-survival behavior needs a genuine server reply, so it correctly lives in the integration suite rather than being mocked, and the removed unit test was pinning behavior that no longer exists. The hot-path tradeoff (full script text vs. hash on `ReclaimAndForget`/`HashSetAndForget`/`HashSetIfExistsAndForget`) is explicitly acknowledged and is the marginal-savings path the issue itself flagged.

Not holding for the nit, but I'd suggest fixing that one-word reference before merge so the doc stays accurate.

---

Issue #2144, opened 2026-07-18
Prerequisite cycle detection ignores implicit within-path tier ordering — a cross-path cycle composed with it still deadlocks silently


ginderjeremiah:
## Context

Follow-up to #2128, split out to keep that PR reviewable. #2128 fixes the concrete bug it reports — a gateway can name a prerequisite on its *own* path, which deadlocks the path against its own implicit tier ordering (tier N+1 requires tier N maxed) — by rejecting any same-path prerequisite outright at all three layers (`AdminProficiencies.ValidatePrerequisiteIds`, `ProgressionGraphChecker.CheckProficiencies`, `GatewaysEditor.addOptions`).

It deliberately does **not** implement the other half of #2128's suggested fix — extending `ValidatePrerequisiteIds`'/`ProficiencyPrerequisiteGraph.TryFindCycle`'s cycle detection to include the *implicit* within-path edges (tier N+1 → tier N) — because that's a materially different and larger change than a same-path ban.

## The remaining gap

`ProficiencyPrerequisiteGraph`'s doc comment states: "Within-path order is implicit in the tier ordinals and inherently acyclic, so only the authored prerequisite edges are checked here." That's true in isolation, but **cross-path edges composed with the implicit within-path chain can still close a cycle** that the current (explicit-edges-only) check never sees:

- Path A tier 0 gates on Path B tier 2 (cross-path — allowed today, and still allowed after #2128).
- Path B tier 0 gates on Path A tier 0 (cross-path — also allowed).
- Path B's own implicit ordering means tier 2 needs tier 1 needs tier 0.
- Composed: A0 → B2 → B1 → B0 → A0. A cycle, but invisible to `ProficiencyPrerequisiteGraph.TryFindCycle`, which only walks the authored edges (A0→B2, B0→A0) and sees no cycle in that narrower graph.

Every node on that composed cycle silently deadlocks — exactly the "wedged/unsatisfiable gateway" failure mode #2128 fixes for the same-path case, just reachable a different way.

## Suggested fix

Extend cycle detection to run over the *combined* graph (authored prerequisite edges + implicit tier-N+1→tier-N edges per path, derived from `PathOrdinal`), not just the authored ones. This needs to land in:

1. `AdminProficiencies.FindPrerequisiteCycle` (admin save-time rejection).
2. `ProficienciesCacheHolder.BuildSnapshotAsync`'s build-time backstop (currently also explicit-edges-only).
[...]

PR #2145, opened 2026-07-18
Reject same-path proficiency prerequisites at all three authoring layers

### LINKED ISSUE #2128: A gateway prerequisite inside its own path can author an unopenable tier — passes the Workbench picker, save validation, and Content Health
## Problem

Gateways are defined everywhere as **cross-path** (editor copy, store section header, `pathReferences` counts only `p.pathId !== id` edges), but nothing actually enforces that — at any of the three layers:

- **Workbench picker** — `UI/src/routes/admin/workbench/progression/GatewaysEditor.svelte:66-68` filters only `p.id !== tier.id && !p.retiredAt && !tier.prerequisiteIds.includes(p.id)`, so tiers of the *same* path are offered.
- **Save validation** — `AdminProficiencies.ValidatePrerequisiteIds` (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs:234-250`) rejects only self-reference and nonexistent ids.
- **Content Health** — `ProgressionGraphChecker.CheckProficiencies` (`Game.Application/Content/ProgressionGraphChecker.cs:592-606`) checks prerequisite existence and retired/frozen prerequisites only; there is no same-path check and no cycle detection that models the implicit within-path ordering (tier N+1 requires tier N maxed).

So gating tier 0 on a later tier of its own path saves fine, passes lint, and **deadlocks the path**: tier 0 becomes a gateway that accrues nothing until tier 2 is maxed, but tier 2 can't be reached without tier 0. The same blindspot covers cross-path prerequisite cycles composed with within-path ordering.

## Fix (all three layers, backend as source of truth)

1. `ValidatePrerequisiteIds` rejects a prerequisite on the tier's own path (anti-tamper — a tampered admin client can't bypass a frontend filter), and ideally detects prerequisite cycles across the graph including the implicit within-path edges.
2. `ProgressionGraphChecker` gains an Error for a wedged/unsatisfiable gateway (same-path prerequisite or cycle), alongside the existing retired/frozen-prerequisite check.
3. `GatewaysEditor.addOptions` filters `p.pathId !== tier.pathId`.

ginderjeremiah:
## Summary

Gateways are documented everywhere as **cross-path** (editor copy, store section header, `pathReferences` counts only `p.pathId !== id` edges), but nothing actually enforced it. Gating tier 0 of a path on a later tier of that *same* path saved fine, passed lint, and permanently deadlocked the path: tier 0 becomes a gateway that accrues nothing until tier 2 is maxed, but tier 2 can't be reached without tier 0 opening first.

This fixes the gap at all three layers the issue calls out, backend as the source of truth:

- **`AdminProficiencies.ValidatePrerequisiteIds`** (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs`) now rejects a prerequisite whose `PathId` matches the gated tier's own — anti-tamper, since a tampered admin client can't bypass a frontend-only filter.
- **`ProgressionGraphChecker.CheckProficiencies`** (`Game.Application/Content/ProgressionGraphChecker.cs`) gains an `Error` finding for the same case, alongside the existing retired/frozen-prerequisite check, so Content Health catches it too (e.g. against a seed/migration mistake).
- **`GatewaysEditor.addOptions`** (`UI/src/routes/admin/workbench/progression/GatewaysEditor.svelte`) filters the "Add prerequisite" picker to `p.pathId !== tier.pathId` so the Workbench never offers an invalid option in the first place.

## Scope note

The issue's fix section also suggested (marked "ideally") extending cycle detection to model the *implicit* within-path tier ordering (tier N+1 requires tier N maxed), so a cross-path prerequisite cycle composed with that implicit chain would also be caught. That's a materially larger change — it needs a shared graph-cycle utility reachable from both `Game.DataAccess` (admin save + build-time cache backstop) and `Game.Application` (Content Health), which today can't see each other's `internal` types. I split that out as **#2144** rather than scope-creep this PR; the same-path ban here does not subsume it (the composed-cycle scenario in #2144 never involves a same-path edge).

## Test plan

- [x] Backend: `Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs` — new `SetPrerequisites_PrerequisiteOnSamePath_ReturnsFailure` integration test seeding two tiers on one path and asserting the save is rejected.
- [x] Backend: `Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs` — new `Proficiency_PrerequisiteOnOwnPath_IsError` unit test.
- [x] Frontend: new `UI/src/tests/routes/admin/workbench/progression/GatewaysEditor.test.ts` — asserts the picker excludes same-path tiers while still offering cross-path ones, and still respects the existing already-added/retired exclusions.
- [x] Frontend: updated two `TierDetail.test.ts` fixtures that previously modeled an (now-invalid) same-path prerequisite to use cross-path tiers instead.
- [x] `dotnet build` — clean.
- [x] `dotnet test --no-build --no-restore` — full solution green (Core/Infrastructure/Application/Api test projects).
- [x] `npm run test:unit:ci` — full suite green (302 files / 3778 tests).

Closes #2128

---
_Generated by [Claude Code](https://claude.ai/code/session_01GFDgZ5UfE2fs31amgZTjj4)_

### DIFF
--- a/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
+++ b/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
@@ -687,6 +687,22 @@ namespace Game.Application.Tests.Content
         }
 
         [Fact]
+        public void Proficiency_PrerequisiteOnOwnPath_IsError()
+        {
+            // Gateways are cross-path only; a same-path prerequisite deadlocks the path against its own
+            // implicit tier ordering (tier N+1 requires tier N maxed), regardless of retirement state.
+            var graph = HealthyGraph() with
+            {
+                Proficiencies =
+                [
+                    Proficiency(0, pathId: 0, maxLevel: 10, rewards: [(5, 5)], prerequisiteIds: [1]),
+                    Proficiency(1, pathId: 0, maxLevel: 10),
+                ],
+            };
+            AssertHasFinding(graph, "ProficiencyPrerequisite", ContentGraphSeverity.Error, "Proficiency", 0);
+        }
+
+        [Fact]
         public void Proficiency_OnMissingPath_IsError()
         {
             var graph = HealthyGraph() with
--- a/Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs
@@ -605,6 +605,33 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SetPrerequisites_PrerequisiteOnSamePath_ReturnsFailure()
+        {
+            int gatedId, prerequisiteId;
+            using (var seedScope = CreateScope())
+            {
+                var path = await SeedPathAsync(seedScope);
+                gatedId = (await SeedProficiencyAsync(seedScope, path.Id, pathOrdinal: 0)).Id;
+                prerequisiteId = (await SeedProficiencyAsync(seedScope, path.Id, pathOrdinal: 1)).Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminProficiencies>();
+
+            // Gating tier 0 on tier 1 of the same path would deadlock the path — tier 1 can't accrue until
+            // tier 0 opens, but tier 0 never opens because tier 1 (which needs tier 0) can never be maxed.
+            var result = admin.SetPrerequisites([new SetProficiencyPrerequisitesData
+            {
+                Id = gatedId,
+                PrerequisiteIds = [prerequisiteId],
+            }]);
+
+            Assert.False(result.Succeeded);
+            Assert.Contains("same path", result.ErrorMessage);
+        }
+
+        [Fact]
         public async Task SetPrerequisites_WouldFormACycle_ReturnsFailure()
         {
             int gatedId, prerequisiteId;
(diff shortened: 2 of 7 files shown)


ginderjeremiah (commented):
## Review: same-path proficiency prerequisites rejected at all three authoring layers (#2128)

**Verdict:** Approve _(submitted as a comment since I'm the PR author; GitHub won't accept a gating self-review)_ — the fix is correct, minimal, and lands the guard at every layer the issue calls out, with tests to match.

The change enforces the cross-path-only gateway invariant exactly where the issue identified the gaps:

- **`AdminProficiencies.ValidatePrerequisiteIds`** — reworked to take the gated `Proficiency` entity so it can compare `PathId`, rejecting a same-path prerequisite alongside the existing self-reference/existence checks. Backend as source of truth (anti-tamper), consistent with the rest of the admin-authoring validation pattern.
- **`ProgressionGraphChecker.CheckProficiencies`** — the new same-path `Error` sits next to the retired/frozen-prerequisite check, correctly guarded by `onLivePath`, so Content Health backstops a seed/migration mistake.
- **`GatewaysEditor.addOptions`** — the picker filter switches from `p.id !== tier.id` to `p.pathId !== tier.pathId`; since the gated tier is on its own path this still excludes the tier itself, so no regression, and same-path siblings are now never offered.

### Scope split
Deferring the composed-cycle case (cross-path edges + implicit within-path tier ordering) to #2144 is the right call — it genuinely needs a shared graph-cycle utility reachable across the `Game.DataAccess`/`Game.Application` assembly boundary, which is a materially larger change. The same-path ban here provably doesn't subsume it, and #2144 documents that clearly.

### Test coverage
Thorough and following the project's rules:
- Backend integration test for the admin-save rejection (`SetPrerequisites_PrerequisiteOnSamePath_ReturnsFailure`) and a Content Health unit test (`Proficiency_PrerequisiteOnOwnPath_IsError`).
- Frontend `GatewaysEditor.test.ts` asserts both that same-path tiers are excluded and that the existing already-added/retired exclusions still hold for cross-path tiers.
- The two `TierDetail.test.ts` fixtures that modeled a (now-invalid) same-path prerequisite were correctly updated to cross-path rather than left to silently misrepresent valid data.

### What's good
Clean, well-scoped fix with clear rationale comments at each layer and disciplined scope management — no gold-plating, and the follow-up gap is filed rather than hand-waved.

---

PR #2146, opened 2026-07-18
Fix four keyboard/focus accessibility holes

### LINKED ISSUE #2133: Keyboard/focus accessibility holes: hover-only sidebar rail, card-game Space swallow, unfocused welcome-back gate, Popover focus dropped on async content swap
Four small keyboard/focus defects, bundled because each fix is a few lines and they share the frontend.md "real interactive controls" intent:

1. **Sidebar rail is not keyboard-expandable; the pin button is unreachable without a mouse.** `UI/src/components/sidebar/CollapsibleRail.svelte:108-110` — `expanded = $derived(pinned || hovering)` with `hovering` driven only by `onmouseenter`/`onmouseleave`, and the pin toggle rendered inside `{#if expanded}`. A keyboard-only user can never expand the rail or reach the pin control, and labels stay at `opacity: 0` while tabbing through nav items. An `onfocusin`/`onfocusout` pair mirroring the hover handlers fixes it.

2. **Card-game global Space handler swallows Space-activation of focused buttons.** `UI/src/routes/game/screens/card-game/loom-input.ts:32-43` — the window `keydown` calls `preventDefault()` on Space and engages Reflex whenever focus isn't in `INPUT`/`TEXTAREA`, so tabbing to "↺ RESET" (`loom/Controls.svelte:27`) and pressing Space triggers slow-time instead of the click (Enter still works). Guard on a focused button/interactive element. (Demo screen — could ride along with #259 work instead.)

3. **Welcome-back gate's only action isn't focused on open.** `UI/src/routes/game/welcome-back/WelcomeBackGate.svelte:46` — the gate is a plain full-screen div (no focus trap, no initial focus), so on arrival Enter does nothing until the player tabs to the button, despite the component doc describing Enter-to-dismiss. Autofocus the primary button (or add a window Enter handler).

4. **Popover initial focus lands on a control that then unmounts.** `UI/src/components/Popover.svelte:51-56` focuses the first focusable exactly once, keyed on `open`/`shell`. `CharacterSwitcher.svelte` renders a loading-state Cancel button first; when the async load resolves and `PlayerSelectPanel` replaces it, the focused element unmounts and focus falls to `<body>` until the next Tab re-anchors via the trap (`focus-trap.ts:83-85`). Re-run the focus placement when the popover's content swaps (or focus after the ready state renders). Minor — it self-heals on Tab.

ginderjeremiah:
## Summary

Fixes the four bundled keyboard/focus defects from #2133:

1. **Sidebar rail unreachable by keyboard.** `CollapsibleRail.svelte`'s `expanded` state was driven only by hover. Added `onfocusin`/`onfocusout` handlers (mirroring the existing hover handlers) so tabbing into any nav item expands the rail — making the pin button reachable and nav labels visible while tabbing, instead of staying at `opacity: 0`.

2. **Card-game Space handler swallowed button activation.** The global `keydown` handler in `loom-input.ts` called `preventDefault()` and engaged Reflex whenever focus wasn't in an `INPUT`/`TEXTAREA`, so tabbing to the "↺ RESET" button and pressing Space triggered slow-time instead of the click. Added `BUTTON` to the focused-element bypass list — the Reflex button already has its own explicit Space/Enter handling, so this only changes behavior for buttons (like Reset) that rely on native activation.

3. **Welcome-back gate's only action wasn't focused on open.** `WelcomeBackGate.svelte` is a plain full-screen div with no initial focus, so Enter did nothing until the player tabbed to the button, despite the docs describing Enter-to-dismiss. Added `autofocus` to the primary button (same pattern already used in `TagSearch.svelte`/`RewardPicker.svelte`).

4. **Popover focus dropped on async content swap.** `Popover.svelte` placed initial focus once, keyed on `open`/`shell`. When a consumer (e.g. `CharacterSwitcher`) swaps its loading-state content for the ready state within the same open popover, the focused element unmounts and focus fell to `<body>` until the next Tab re-anchored it via the trap. Added a `MutationObserver` that re-places focus onto the (possibly new) first focusable whenever a content swap drops focus outside the shell.

## Test plan

- [x] Added a `NavSidebar` test asserting `focusin`/`focusout` expands/collapses the rail and exposes the pin button.
- [x] Added a `loom-input` test asserting a focused button's Space keydown is no longer intercepted (no `preventDefault`, no Reflex engaged).
- [x] Added a `WelcomeBackGate` test asserting the enter button is the active element on render.
- [x] Added a `Popover` test simulating a loading→ready content swap and asserting focus re-anchors onto the new content.
- [x] Full frontend unit suite (`vitest run`) — 302 files / 3782 tests passing.

Closes #2133

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xr2BM5Tsp6bxRuR3N92Y4q)_

### DIFF
--- a/UI/src/components/Popover.svelte
+++ b/UI/src/components/Popover.svelte
@@ -48,10 +48,24 @@ let shell = $state<HTMLElement | null>(null);
 
 // On open, move focus onto the first focusable inside the popover (or the shell itself if it has
 // none) so the trap has somewhere to anchor. Trap/Escape/scroll-lock/restore are owned by focusTrap.
+// Content can also swap while open (e.g. a loading placeholder replaced once an async fetch resolves);
+// if that unmounts the focused element, focus falls back to <body>, so a MutationObserver re-places it
+// onto the (possibly new) first focusable whenever a swap drops focus outside the shell.
 $effect(() => {
 	if (open && shell) {
-		const target = shell.querySelector<HTMLElement>(FOCUSABLE_SELECTOR);
-		(target ?? shell).focus();
+		const placeFocus = () => {
+			const target = shell?.querySelector<HTMLElement>(FOCUSABLE_SELECTOR);
+			(target ?? shell)?.focus();
+		};
+		placeFocus();
+
+		const observer = new MutationObserver(() => {
+			if (!shell?.contains(document.activeElement)) {
+				placeFocus();
+			}
+		});
+		observer.observe(shell, { childList: true, subtree: true });
+		return () => observer.disconnect();
 	}
 });
 </script>
--- a/UI/src/components/sidebar/CollapsibleRail.svelte
+++ b/UI/src/components/sidebar/CollapsibleRail.svelte
@@ -7,6 +7,8 @@
 	style:--rail-header-padding="{headerPadding}px"
 	onmouseenter={() => (hovering = true)}
 	onmouseleave={() => (hovering = false)}
+	onfocusin={() => (focused = true)}
+	onfocusout={() => (focused = false)}
 >
 	<!-- Header: brand wordmark + pin toggle -->
 	<div class="sidebar-header">
@@ -106,8 +108,9 @@ let {
 }: Props = $props();
 
 let hovering = $state(false);
+let focused = $state(false);
 
-const expanded = $derived(pinned || hovering);
+const expanded = $derived(pinned || hovering || focused);
 </script>
 
 <style lang="scss">
--- a/UI/src/routes/game/screens/card-game/loom-input.ts
+++ b/UI/src/routes/game/screens/card-game/loom-input.ts
@@ -28,10 +28,11 @@ export function runLoomLoop(view: CardGameView): () => void {
  *  pointer release/cancel anywhere ends a held Reflex (the button is press-and-hold, so the release can
  *  land off the button). Returns a teardown that removes every listener. */
 export function bindLoomInput(view: CardGameView): () => void {
-	// Hotkeys yield to focused sandbox sliders so they keep working.
+	// Hotkeys yield to focused sandbox sliders and buttons so they keep working (a focused button's own
+	// Space/Enter handling — e.g. the Reflex button's press-and-hold — takes precedence over the global one).
 	const onKeyDown = (e: KeyboardEvent) => {
 		const ae = document.activeElement;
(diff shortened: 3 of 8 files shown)


ginderjeremiah (commented):
## Review: keyboard/focus accessibility fixes (#2133)

**Verdict:** Approve — all four fixes are minimal, correctly scoped, and each ships a matching test. (Posted as a comment because I'm the PR author and GitHub won't accept a gating self-review.)

I traced each fix against the issue's intent and the surrounding code:

- **Sidebar rail (`CollapsibleRail.svelte`)** — the `focused` state mirrors `hovering` and folds into `expanded = pinned || hovering || focused`. `onfocusin`/`onfocusout` bubble, so any nav item receiving focus expands the rail and exposes the pin button. Tabbing between two items inside the rail nets `focused = true` after Svelte batches the focusout→focusin pair, so there's no collapse flicker.
- **Card-game Space swallow (`loom-input.ts`)** — adding `BUTTON` to the bypass list is safe because the only button that relies on Space-for-slow-time, the Reflex button, owns its own `onkeydown`/`onkeyup` Space/Enter handling (`loom/Controls.svelte:8-19`). So the RESET button regains native Space activation with no regression to Reflex.
- **Welcome-back gate (`WelcomeBackGate.svelte`)** — `autofocus` on the primary action matches the existing `TagSearch`/`RewardPicker` pattern, and the `svelte-ignore a11y_autofocus` is appropriate for a single-action full-screen gate.
- **Popover async swap (`Popover.svelte`)** — the `MutationObserver` re-places focus only when a content mutation has dropped `document.activeElement` outside the shell (`!shell.contains(...)`), which confines it to exactly the loading→ready unmount case rather than yanking focus on every DOM update. `focus()` triggers no mutations, so there's no observer loop, and the effect disconnects the observer on teardown. The optional chaining on `shell` is correct defensive coding since the observer callback runs after setup.

### Test coverage
Solid. Each fix has a targeted test mirroring the source path: `NavSidebar` (focusin/out expand + pin exposure), `loom-input` (focused-button Space no longer intercepted), `WelcomeBackGate` (button is active element on render), and `Popover` (loading→ready swap re-anchors focus). No backend or battle logic touched, so no cross-suite mirroring is required.

### What's good
Tight, well-commented changes that fix real keyboard-only defects without over-engineering. The Popover observer in particular is scoped precisely to the drop case rather than reaching for a blunter re-focus-on-every-change approach.

No blocking findings, no should-fixes, no nits. Approving; will merge once CI is green.

---

PR #2148, opened 2026-07-18
Add post-commit-failure test coverage for ProgressionStore.save() (#2062)

### LINKED ISSUE #2062: ProgressionStore.save() has no post-commit-failure test coverage
## Context

`EntityStore` has three dedicated tests for its save-failure recovery split ("re-seeds from server truth on a partial (committed) failure", "keeps local edits on a total (pre-commit) failure", "keeps local edits when the recovery refresh also fails"). `ProgressionStore.save()` *re-implements* this logic across a far riskier 9-step orchestration (early prereq split, early child removals, two id-remap refetches, final combined prereq batch — `progression-store.svelte.ts:579-588`), yet `progression-store.test.ts` covers only the pre-commit case.

Untested: a failure after `AddEditPaths` committed (must reseed so a retry can't re-Add), a failure in step 7/8 after child writes, the swallowed `reseed().catch(() =&gt; {})` leaving the baseline behind server truth, and the step-9 selection remap when `resolveId` throws for an unmapped negative selection. Given the retire/shrink ordering rules in this store were each bug-driven (#1776, #1827/#1804), the recovery path deserves the same scenario coverage the generic store has.

## Suggested fix

Add the committed-failure scenarios mirroring `EntityStore`'s three, plus the reseed-also-fails and selection-remap cases.

ginderjeremiah:
## Summary

`ProgressionStore.save()` re-implements `EntityStore`'s save-failure recovery split (committed vs. pre-commit failure, plus a swallowed recovery-refresh failure) across a much riskier 9-step orchestration, but `progression-store.test.ts` only covered the pre-commit case. This adds the missing post-commit-failure scenarios, mirroring `EntityStore`'s existing three tests plus one scenario unique to `ProgressionStore`'s extra selection-remap step:

- **Committed-then-refetch-fails**: a path `Add` commits (step 3), but the refetch that resolves its persisted id (step 4) then fails — recovery re-seeds from server truth so a retry can't re-`Add` the already-persisted path.
- **Child-write-then-child-write-fails**: a modifier write commits (step 7) before the reward write in the same batch fails — same re-seed recovery, scoped to a single tier's child collections.
- **Swallowed recovery-refresh failure**: when the committed-failure `reseed().catch(() => {})` itself fails, local state is left untouched (and now behind server truth) rather than throwing — documents/verifies this known, existing behavior.
- **Step-9 selection remap**: `resolveId` throwing for an unmapped negative selection (a stale post-add refetch that doesn't yet show the just-persisted path) is caught by the same committed-failure recovery path.

No production code changed — this is purely test-coverage for existing behavior, per the "tech debt" / `scope: small` labeling on the issue.

Closes #2062

## Test plan

- [x] `npx vitest run progression-store.test.ts` — 29/29 passed (4 new)
- [x] `npm run test:unit:ci` (full frontend suite) — 3790/3790 passed
- Backend unaffected (no `.cs` changes), so `dotnet build`/`dotnet test` not run.

---
_Generated by [Claude Code](https://claude.ai/code/session_018VNqhAiCibFXWhXiQd1pRL)_

### DIFF
--- a/UI/src/tests/routes/admin/workbench/progression/progression-store.test.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-store.test.ts
@@ -630,6 +630,116 @@ describe('save orchestration', () => {
 		// Pre-commit failure: the working edit is preserved for a clean retry.
 		expect(store.paths[0].name).toBe('Inferno');
 	});
+
+	it('re-seeds from server truth when the refetch after a committed path Add then fails, so a retry cannot re-Add it', async () => {
+		const store = new ProgressionStore();
+		await store.load();
+
+		store.addPath();
+		const pathLocal = store.selectedPathId as number;
+		store.patchPath(pathLocal, (d) => {
+			d.name = 'Fire';
+			d.activityKey = EActivityKey.Fire;
+		});
+
+		// The path Add commits server-side (step 3), but the very next refetch that resolves its
+		// persisted id (step 4) then fails — a partial, already-committed failure.
+		fetchMock.mockImplementationOnce(async () => {
+			throw new Error('replica timeout');
+		});
+
+		await store.save();
+
+		expect(toastErrorMock).toHaveBeenCalled();
+		expect(serverPaths).toHaveLength(1);
+		// Recovery re-seeded from server truth: no phantom negative-id path survives to re-Add on retry.
+		expect(store.paths.some((p) => p.id < 0)).toBe(false);
+		expect(store.paths).toHaveLength(1);
+		expect(store.paths[0].name).toBe('Fire');
+		expect(store.totalChanges).toBe(0);
+	});
+
+	it('re-seeds after a modifier write commits but the reward write in the same save then fails', async () => {
+		seedServer();
+		const store = new ProgressionStore();
+		await store.load();
+
+		store.addModifier(0, 3);
+		store.setReward(0, 3, 5);
+
+		postMock.mockImplementation(async (endpoint: string, body: Rec) => {
+			if (endpoint === 'AdminTools/SetProficiencyRewards') {
+				throw new Error('rewards write failed');
+			}
+			applyPost(endpoint, body);
+			return {};
+		});
+
+		await store.save();
+
+		expect(toastErrorMock).toHaveBeenCalled();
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: post-commit-failure test coverage for `ProgressionStore.save()` (closes #2062)

**Verdict:** Approve _(posted as a comment since I'm the PR author; no gating event available on one's own PR)_ — the four added scenarios accurately exercise the committed-failure recovery split the issue flagged as untested, with no production code touched.

I traced each new test against the real `save()` orchestration (`progression-store.svelte.ts:402-594`) and the `resolveNewIds`/`resolveId` helpers (`save-helpers.ts:58-116`):

- **Committed-then-refetch-fails** (`:634`) — step 3 commits `AddEditPaths`, the step-4 `GetPaths` refetch throws, `committed` is true, so the `catch` reseeds from server truth and the phantom negative-id path is gone. Correct.
- **Child-write-then-child-write-fails** (`:649`) — the step-7 modifier write commits before the reward write in the same batch throws; reseed reflects the partial server truth (modifiers persisted, rewards not). Correct.
- **Swallowed recovery-refresh failure** (`:671`) — persistent `fetchMock` throw makes the `reseed().catch(() => {})` swallow, leaving local state (now behind server truth) untouched with the negative-id path still dirty. Correctly documents existing behavior rather than asserting an aspiration.
- **Step-9 selection remap** (`:723`) — dropping the auto-tier ensures the unmapped local id first surfaces at the selection remap; `resolveNewIds` tolerates the missing add (empty map, no throw) and `resolveId` throws `No persisted id found`, caught by the committed-failure path. Correct.

### Test coverage
This IS the coverage — it closes the exact gap #2062 identified, mirroring `EntityStore`'s three committed/pre-commit/reseed-fails tests plus the `ProgressionStore`-only selection-remap case. Backend untouched (no `.cs`), so the mirrored-battle-logic rule doesn't apply. Full frontend suite reported green in the PR description.

### What's good
The scenario comments earn their length — they pin down the non-obvious step numbering and the deliberately-swallowed `catch`, which is exactly the kind of construct a later "simplification" could naively break. Identity-content matching and the fake in-memory backend keep these from being tautological mocks.

---

Issue #2150, opened 2026-07-18
Device.cs's "fingerprint incorporates the user-agent, so the mapping is stable" doesn't hold for the actual fingerprint inputs


ginderjeremiah:
## Context

Found while working on #2064 (device/login-tracking hardening).

`Device.cs`'s doc comment states: "Each device belongs to one `BrowserInfo` (user-agent) profile — the fingerprint incorporates the user-agent, so the mapping is stable."

But the actual fingerprint inputs (`UI/src/lib/api/device-fingerprint.ts`'s `fingerprintSignals`) are: a coarse `platformToken` (prefers `navigator.userAgentData.platform`, e.g. `"Windows"`; falls back to a regex-derived OS family from the UA string, explicitly *excluding* version numbers so a browser auto-update doesn't rotate the fingerprint), `language`, `languages`, `hardwareConcurrency`, `deviceMemory`, screen dimensions, and time zone — never the raw `User-Agent` string itself. That's deliberate (the code comment there says a browser auto-update "no longer rotates the fingerprint").

So a browser version upgrade (or even a browser switch that reports the same coarse platform/hardware/locale signals) can keep the *same* fingerprint hash while producing a *different* `User-Agent` header. Since `Device.BrowserInfoId` is set once at device creation and never updated afterward (`UserLogins.CreateDevice`/`GetOrCreateBrowserInfo` only fill in missing `SecChUa*` hints on a match, never repoint `BrowserInfoId`), a returning device's stored `BrowserInfo` can drift from its current actual browser over time, and the doc comment's stability claim doesn't match this.

## Impact

Low — `BrowserInfo` is presentation/telemetry data (recorded for admin visibility), not used in any authorization or gameplay decision. This is a documentation/design-nit, not a functional bug.

## Suggested fix

Either:
- Correct the doc comment to describe the actual (weaker) stability guarantee, or
- Have `RecordConnection` repoint an existing `Device`'s `BrowserInfoId` to the current request's resolved `BrowserInfo` when they differ, keeping the stored profile current.

Tags: tech debt, claude, scope: small

PR #2151, opened 2026-07-18
Harden device/login tracking against unbounded growth and cross-account writes

### LINKED ISSUE #2064: Harden device/login tracking: unbounded client-controlled fingerprint rows; DeviceInfo not scoped to the caller's account
## Context

`LoginTrackingMiddleware`/`UserLogins.GetOrCreateDevice` (`Game.Api/Middleware/LoginTrackingMiddleware.cs:41-68`, `Game.DataAccess/Repositories/UserLogins.cs:96-119`) insert a new `Device` (and, for a novel UA, `BrowserInfo`) row for every fingerprint value not seen before. The fingerprint is an arbitrary request header (`X-Device-Fingerprint`), and authenticated endpoints have no rate limit (the `auth` policy covers only the four anonymous endpoints).

**Failure scenario:** an authenticated user scripting requests with a random fingerprint per request inserts one `Device` + `UserLogin` row per request, forever — the 5-minute memo never dedupes because every key is new, so every request also pays the multi-query upsert. Table bloat/DoS on the tracking tables, plus one `IMemoryCache` entry per request (no `SizeLimit`; bounded only by expiry × request rate). Relatedly, `DeviceInfo` (`LoginController.cs:373-393`) resolves the device by fingerprint alone — not scoped to the caller's account — so any authenticated user who learns another device's fingerprint can overwrite its `DeviceMemory`/`HardwareConcurrency` (low impact, worth closing in the same change).

The docs don't record this as an accepted trade-off.

## Suggested fix

Cap tracked devices per user (skip insert past N distinct fingerprints), validate the fingerprint shape (enforce hex/base64 of fixed length instead of any ≤128-char string), scope `DeviceInfo`'s device resolution to the caller's account, and/or add a modest rate limit on authenticated tracking writes.

ginderjeremiah:
## Summary

Closes #2064.

The July 2026 health review flagged that `LoginTrackingMiddleware`/`UserLogins.GetOrCreateDevice` insert a new `Device` (+ `BrowserInfo`) row for every fingerprint value not seen before, with no cap and no shape validation on the client-supplied `X-Device-Fingerprint` header — so an authenticated caller cycling a fresh fingerprint per request could grow the tracking tables without bound. Separately, `DeviceInfo`'s `SaveDeviceInfo` resolved (and could create) a `Device` by fingerprint alone, with no link back to the caller's account.

This PR:

- **Validates the fingerprint shape** (`Game.Api/Http/ClientHints.cs`): the frontend's fingerprint is always a 64-character lowercase-hex SHA-256 digest (`device-fingerprint.ts`'s `hashFingerprint`). `ClientHints.DeviceFingerprint` now rejects anything else, instead of trusting an arbitrary ≤128-char client-supplied string.
- **Caps distinct devices tracked per account** (`Game.DataAccess/Repositories/UserLogins.cs`, `MaxTrackedDevicesPerUser = 20`): once a user has that many distinct devices, a connection from a device they haven't already been seen on is silently not tracked (tracking is best-effort telemetry, not a gate on play) rather than growing the table further. A device the user already owns (new IP, same device) is exempt from the cap since it doesn't grow their device count.
- **Scopes `SaveDeviceInfo` to the caller's own tracked device**: it no longer independently resolves-or-creates a `Device` by fingerprint alone. It's now a no-op unless the caller already has a `UserLogin` for that fingerprint (established by `RecordConnection`, which always runs first in the same request via `LoginTrackingMiddleware`). This also closes an uncapped bloat path — previously `SaveDeviceInfo` could create devices independently of `RecordConnection`'s cap.
- **Bounds the login-tracking dedupe `IMemoryCache`** (`Game.Api/Startup.cs`, `Game.Api/Middleware/LoginTrackingMiddleware.cs`): the issue also called out the per-request memo cache as an unbounded structure fed by the same failure scenario. It's the only consumer of that cache instance, so adding a `SizeLimit` (100k entries, each sized 1) was low-risk to close alongside the rest.

### Scope decision

The issue's suggested fix also mentioned "add a modest rate limit on authenticated tracking writes" as an *and/or* option. I didn't add one: the per-user device cap already bounds the tables regardless of request volume, and a distinct throttle for these two authenticated endpoints felt like unwarranted complexity once the cap and fingerprint-shape validation are in place. Happy to revisit if reviewers disagree.

### Follow-up filed

While reviewing `Device`/`BrowserInfo` linkage I noticed `Device.cs`'s doc comment ("the fingerprint incorporates the user-agent, so the mapping is stable") doesn't hold given the actual fingerprint inputs (a coarse platform token, not the raw UA — deliberately, so browser auto-updates don't rotate the fingerprint). Low-impact (presentation/telemetry only), filed as #2150 rather than expanding this PR's scope.

## Changes

- `Game.Api/Http/ClientHints.cs`: fingerprint shape validation via a generated regex.
- `Game.DataAccess/Repositories/UserLogins.cs`: device cap in `RecordConnection`; account-scoped, non-creating `SaveDeviceInfo`; fixed a real bug along the way where a newly-created device's `UserLogin` needs to be linked via the `Device` navigation (not `DeviceId` directly), since the device has no generated Id until `SaveChanges` runs.
- `Game.Abstractions/DataAccess/IUserLogins.cs`, `Game.Application/Services/LoginTrackingService.cs`, `Game.Api/Controllers/AuthController.cs`: threaded `userId` through to `SaveDeviceInfo`; dropped the now-unused `userAgent` parameter (enrichment only backfills missing `BrowserInfo` client hints on an already-resolved device, it never creates one).
- `Game.Api/Startup.cs`, `Game.Api/Middleware/LoginTrackingMiddleware.cs`: `SizeLimit` on the tracking dedupe cache.
- Test updates across `Game.Api.Tests`/`Game.Application.Tests` for the new signatures and the fingerprint-shape requirement, plus new coverage: `ClientHintsTests` (malformed-shape rejection), `UserLoginsCapAndOwnershipIntegrationTests` (new — device cap enforcement, ownership-scoped enrichment, cross-account non-enrichment).

## Test plan

- [x] Full backend test suite (`dotnet test`) — 1376 + 68 + 1005 + 827 = 3276 tests, all passing.
- [x] New integration tests specifically exercise: device cap reached → new device not tracked but existing device still updates; `SaveDeviceInfo` no-ops with no tracked login; `SaveDeviceInfo` doesn't enrich another account's device; `SaveDeviceInfo` does enrich the caller's own device.

---
_Generated by [Claude Code](https://claude.ai/code/session_016UQyR5hbSsFC5UQSJhVuU7)_

### DIFF
--- a/Game.Api/Http/ClientHints.cs
+++ b/Game.Api/Http/ClientHints.cs
@@ -40,8 +40,9 @@ namespace Game.Api.Http
         private static string? NullIfEmpty(string value) => string.IsNullOrEmpty(value) ? null : value;
 
         // A SHA-256 digest rendered as lowercase hex (device-fingerprint.ts's hashFingerprint): exactly 64
-        // characters, so an unbounded or malformed header can never reach the data tier.
-        [GeneratedRegex("^[0-9a-f]{64}$")]
+        // characters, so an unbounded or malformed header can never reach the data tier. \z (not $) anchors
+        // strictly to end-of-string — $ would also accept a trailing "\n" after the 64th hex character.
+        [GeneratedRegex("^[0-9a-f]{64}\\z")]
         private static partial Regex FingerprintShapeRegex();
     }
 }
--- a/Game.Api/Http/ClientHints.cs
+++ b/Game.Api/Http/ClientHints.cs
@@ -1,10 +1,12 @@
+using System.Text.RegularExpressions;
+
 namespace Game.Api.Http
 {
     /// <summary>
     /// The user-agent string and low-entropy client-hint headers carried on an HTTP request, used to
     /// identify the requesting browser/device for connection tracking.
     /// </summary>
-    public readonly record struct ClientHints(
+    public readonly partial record struct ClientHints(
         string UserAgent,
         string? SecChUa,
         string? SecChUaMobile,
@@ -23,10 +25,23 @@ namespace Game.Api.Http
             NullIfEmpty(headers["Sec-CH-UA-Mobile"].ToString()),
             NullIfEmpty(headers["Sec-CH-UA-Platform"].ToString()));
 
-        /// <summary>Reads the device fingerprint header, or null when the request did not carry one.</summary>
-        public static string? DeviceFingerprint(IHeaderDictionary headers) =>
-            NullIfEmpty(headers[DeviceFingerprintHeader].ToString());
+        /// <summary>
+        /// Reads the device fingerprint header, or null when the request did not carry one or its value
+        /// doesn't have the shape of the frontend's SHA-256 hex digest (<see cref="FingerprintShapeRegex"/>).
+        /// Rejecting anything else keeps an arbitrary client-supplied string from reaching the tracking
+        /// tables (#2064) — the header is otherwise unauthenticated client input.
+        /// </summary>
+        public static string? DeviceFingerprint(IHeaderDictionary headers)
+        {
+            var value = NullIfEmpty(headers[DeviceFingerprintHeader].ToString());
+            return value is not null && FingerprintShapeRegex().IsMatch(value) ? value : null;
+        }
 
         private static string? NullIfEmpty(string value) => string.IsNullOrEmpty(value) ? null : value;
+
+        // A SHA-256 digest rendered as lowercase hex (device-fingerprint.ts's hashFingerprint): exactly 64
+        // characters, so an unbounded or malformed header can never reach the data tier.
+        [GeneratedRegex("^[0-9a-f]{64}$")]
+        private static partial Regex FingerprintShapeRegex();
     }
 }
--- a/Game.Abstractions/DataAccess/IUserLogins.cs
+++ b/Game.Abstractions/DataAccess/IUserLogins.cs
@@ -26,13 +26,16 @@ namespace Game.Abstractions.DataAccess
             CancellationToken cancellationToken = default);
 
         /// <summary>
-        /// Enriches (creating when new) the <c>Device</c> for the given fingerprint with the capabilities the
-        /// frontend reports after login (<c>deviceMemory</c>/<c>hardwareConcurrency</c>), which are not present
-        /// on a regular request.
+        /// Enriches the caller's own <c>Device</c> for the given fingerprint with the capabilities the
+        /// frontend reports after login (<c>deviceMemory</c>/<c>hardwareConcurrency</c>), backfilling any
+        /// client-hint the stored <c>BrowserInfo</c> is still missing. A no-op when <paramref name="userId"/>
+        /// has no tracked login for that fingerprint — this path never creates or attaches a <c>Device</c>,
+        /// so a caller can't enrich (or overwrite) a device it hasn't actually connected from, even if it
+        /// knows the fingerprint.
         /// </summary>
         Task SaveDeviceInfo(
+            int userId,
             string deviceFingerprintHash,
-            string userAgent,
             string? secChUa,
             string? secChUaMobile,
             string? secChUaPlatform,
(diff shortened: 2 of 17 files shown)


Game.Api/Http/ClientHints.cs
```diff
@@ -23,10 +25,23 @@ public readonly record struct ClientHints(
...
+        {
+            var value = NullIfEmpty(headers[DeviceFingerprintHeader].ToString());
+            return value is not null && FingerprintShapeRegex().IsMatch(value) ? value : null;
+        }
 
         private static string? NullIfEmpty(string value) => string.IsNullOrEmpty(value) ? null : value;
+
+        // A SHA-256 digest rendered as lowercase hex (device-fingerprint.ts's hashFingerprint): exactly 64
+        // characters, so an unbounded or malformed header can never reach the data tier.
+        [GeneratedRegex("^[0-9a-f]{64}$")]
```
ginderjeremiah: Nit (non-blocking): in .NET regex `$` matches at end-of-string **or before a trailing `\n`**, so `^[0-9a-f]{64}$` also accepts a 64-hex value with a trailing newline (`"a"*64 + "\n"`), and `IsMatch` would let that 65-char string through as the fingerprint. Kestrel won't deliver a raw newline in a header value, so this isn't reachable in practice — but for a check whose whole purpose is strict shape enforcement, `\z` is the precise anchor: `^[0-9a-f]{64}\z`. Cheap hardening, take it or leave it.

---

ginderjeremiah (commented):
## Review: Harden device/login tracking against unbounded growth & cross-account writes (#2064)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change closes every prong of the issue, the design is sound, and coverage is thorough. One non-blocking nit inline.

I reviewed the full scope against the issue and `backend.md`. The three failure modes #2064 called out are each addressed:

- **Unbounded client-controlled rows** → `RecordConnection` now caps distinct devices per user (`MaxTrackedDevicesPerUser = 20`) and skips-rather-than-errors past the cap, correctly exempting a device the user already owns (new IP, same device) so it doesn't grow the count. The `isNewDeviceForUser` → count → cap flow is right, and best-effort telemetry semantics are preserved.
- **Malformed fingerprint reaching the data tier** → `ClientHints.DeviceFingerprint` validates the exact shape of the frontend's SHA-256 digest (`^[0-9a-f]{64}$`), which I verified matches `device-fingerprint.ts`'s `hashFingerprint` (lowercase hex, 64 chars). Clean use of source-generated regex.
- **Cross-account `DeviceInfo` overwrite** → `SaveDeviceInfo` is now account-scoped and no-creates: it only enriches a device the caller already has a `UserLogin` for, relying on `RecordConnection` (run earlier in the request via `LoginTrackingMiddleware`) to establish that link. The two run on separate `GameContext` scopes, but `RecordConnection` self-commits before `_next`, so the enrichment query sees the committed login — and `DeviceInfoEndpoint_EnrichesStoredDevice` exercises exactly that middleware→controller ordering end-to-end.

Also correctly caught along the way: the just-created-device link must go through the `Device` navigation (not `DeviceId`) since the Id isn't generated until `SaveChanges`, and the memo cache now carries a `SizeLimit` to close the last unbounded structure. The scope decision to skip a separate rate limiter is reasonable — the per-user cap bounds the tables regardless of request volume.

### Nits
- `ClientHints.cs:44` — `$` vs `\z` in the shape regex (details inline). Not reachable given Kestrel header sanitization; purely defensive.

### Test coverage
Strong and appropriately placed. `UserLoginsCapAndOwnershipIntegrationTests` (new) covers cap-reached-skips-new / known-device-still-updates, no-tracked-login no-op, cross-account non-enrichment, and own-device enrichment against the DI-resolved repo. `ClientHintsTests` adds the malformed-shape rejection theory (non-hex, uppercase, off-by-one lengths). Signature updates threaded through all the doubles, and the existing HTTP-level `DeviceInfoEndpoint_EnrichesStoredDevice` guards the full enrichment path under the new design. Consistent with `backend.md`'s classical/integration split — this subsystem has no `Game.Core` domain model, so the policy living in the repository matches how device tracking is already structured.

### What's good
- The `RecordConnection` rewrite reads clearly and the comments explain the non-obvious EF navigation-vs-FK subtlety and the cap's best-effort intent without over-documenting.
- Proactively filing #2150 for the stale `Device` doc comment rather than scope-creeping this PR is the right call.

Holding off on an auto-merge only because of the open nit — happy for it to merge once you've taken or consciously declined the `\z` tweak.

---

ginderjeremiah:
Took the `\z` nit in df102d9 and added a regression test (trailing-`\n` case) that fails against the old `$` anchor and passes now. Full suite green (3277 tests).

---

PR #2152, opened 2026-07-18
Clear staticData on logout so re-login re-checks reference-data versions

### LINKED ISSUE #2067: Reference data in memory is never version-re-checked for the rest of the browser session — including across a re-login
## Context

The loading screen's "same-session re-mount" short-circuit (`UI/src/lib/engine/reference-data.ts:233-236` — `if (REFERENCE_DATA.every((source) =&gt; source.loaded())) { return true; }`; `loading-view.svelte.ts:73-77`) is sound for its intended case, but `staticData` is a module singleton and `stopGame` never clears it — and a `SocketReplaced` logout is a client-side `goto`, not a reload. So a subsequent login in the same tab reuses whatever content is in memory with **no version check at all**; content published between sessions is invisible until a hard refresh.

Low stakes at the current dev stage, but the intent deserves one line of code either way.

## Suggested fix

Either clear the `staticData` slots in `stopGame` (the localStorage cache makes the re-hydrate nearly free), or drop the pre-version-check short-circuit and always compare versions (one socket round-trip on re-mount).

ginderjeremiah:
## Summary

Closes #2067.

The loading screen's "same-session re-mount" short-circuit (`REFERENCE_DATA.every((source) => source.loaded())` in `reference-data.ts`, mirrored in `loading-view.svelte.ts`) is sound for its intended case, but `staticData` is a module singleton and `stopGame` never cleared it. A `SocketReplaced` logout is a client-side `goto`, not a page reload, so a subsequent login in the same tab reused whatever reference data was already in memory with **no version check at all** — content published between sessions stayed invisible until a hard refresh.

## Change

`stopGame()` (`UI/src/lib/engine/engine.ts`) now calls a new `staticData.reset()` (`UI/src/stores/static-data.svelte.ts`) that clears every reference-data slot back to `undefined`, alongside the existing `statistics`/`playerChallenges`/`playerProficiencies` resets it already ran on logout.

I went with this option (of the two the issue suggested) over dropping the pre-version-check short-circuit entirely, since clearing on logout only pays the version-check cost when it's actually needed (across a real re-login) and leaves the same-session re-mount fast-path untouched. The local-storage cache (`reference-cache.ts`) keeps the next load's re-hydrate cheap for any set whose version didn't change — the version check itself is one socket round trip.

## Test plan

- [x] `npx vitest run` — 302 files / 3793 tests passing, including new coverage:
  - `static-data.test.ts`: `reset()` clears every slot back to unloaded.
  - `engine.test.ts`: `stopGame` calls `staticData.reset()` on `handleSocketReplaced`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CSukRCjxeEAW5mQhYi2oeP)_

### DIFF
--- a/UI/src/lib/engine/engine.ts
+++ b/UI/src/lib/engine/engine.ts
@@ -362,6 +362,9 @@ const stopGame = () => {
 	statistics.reset();
 	playerChallenges.reset();
 	playerProficiencies.reset();
+	// Cleared so a subsequent login in the same tab re-checks content versions instead of reusing
+	// whatever reference data was in memory for the prior session (#2067).
+	staticData.reset();
 	resetLogs();
 	// Clear the other module-level UI stores stopEngines/the resets above don't touch, so a leftover toast,
 	// queued modal, or pending screen request can't leak into the next session (e.g. a "View" toast action
--- a/UI/src/stores/static-data.svelte.ts
+++ b/UI/src/stores/static-data.svelte.ts
@@ -137,5 +137,24 @@ export const staticData = {
 			skillRecipes,
 			lessons
 		].every((set) => set != null);
+	},
+	/** Clears every slot back to unloaded (e.g. on logout), so a subsequent login re-checks content
+	 *  versions instead of reusing whatever was in memory for the prior session. The local-storage
+	 *  cache (`reference-cache.ts`) makes the re-hydrate on next load nearly free when nothing changed. */
+	reset() {
+		zones = undefined;
+		enemies = undefined;
+		items = undefined;
+		skills = undefined;
+		itemMods = undefined;
+		attributes = undefined;
+		challenges = undefined;
+		challengeTypes = undefined;
+		statisticTypes = undefined;
+		proficiencies = undefined;
+		paths = undefined;
+		classes = undefined;
+		skillRecipes = undefined;
+		lessons = undefined;
 	}
 };
--- a/UI/src/tests/lib/engine/engine.test.ts
+++ b/UI/src/tests/lib/engine/engine.test.ts
@@ -22,8 +22,9 @@ const {
 	navigation,
 	clearToasts
 } = vi.hoisted(() => ({
-	staticDataStub: { loaded: true } as {
+	staticDataStub: { loaded: true, reset: vi.fn() } as {
 		loaded: boolean;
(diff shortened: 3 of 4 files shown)


ginderjeremiah (commented):
## Review: Clear `staticData` on logout so a same-tab re-login re-checks reference-data versions (closes #2067)

**Verdict:** Approve (submitted as a comment since I'm the PR author) — correctly implements the issue's first suggested option, minimal and well-scoped, with mirrored tests on both the store and the engine call site.

### Verification

- The fix works as intended: `RefDataSource.loaded()` (`reference-data.ts`) is `() => read() != null` where `read` reads the corresponding `staticData` slot. Clearing every slot in `reset()` makes all `loaded()` return false, so both the `hydrateAllFromCache` short-circuit (`REFERENCE_DATA.every(s => s.loaded())`) and the loading-view short-circuit fall through to the version check on the next login — exactly the gap #2067 described.
- `reset()` covers all 14 slots and matches the `loaded` getter's list one-for-one; `staticData` is already imported in `engine.ts`.
- Sound choice of option: clearing on logout keeps the same-session re-mount fast-path untouched and only pays the (one round-trip) version check across a real re-login, with the local-storage cache keeping unchanged sets cheap to re-hydrate.

### Test coverage

- `static-data.test.ts`: `reset()` clears every slot back to unloaded (`loaded === false`) — good, and it reuses the existing `SLOTS`/`populateAll` helpers.
- `engine.test.ts`: `stopGame` (via `handleSocketReplaced`) calls `staticData.reset()` exactly once. Adequate for a change this size.

### Nits

- None blocking. Observation for a possible future follow-up (not this PR): the reference-data slot list is now maintained in three spots in `static-data.svelte.ts` — the `$state` declarations, the `loaded` getter, and `reset()` — so adding a new set means remembering all of them (forgetting `reset()` would silently reintroduce this exact stale-on-re-login bug for that set). This is a pre-existing consequence of the per-slot rune design that `reset()` reasonably conforms to rather than something this PR introduces; a fully-DRY registry is awkward with Svelte 5 runes and out of scope here.

### What's good

- Tight, correct fix that matches the issue's analysis; the added code comments explain *why* (re-check versions across a same-tab re-login) without over-explaining; tests land on both sides of the change.


---

PR #2153, opened 2026-07-18
Make createHook.notify re-entrancy-safe with a dispatch-depth counter

### LINKED ISSUE #2093: createHook.notify is not re-entrancy-safe — a same-hook nested notify resets the dispatch flag and compacts tombstones mid-walk (latent)
## Problem

`notify` in `UI/src/lib/common/hooks.ts:16-40` sets `dispatching = true`, walks trackers by index, then sets `dispatching = false` and splices out tombstoned trackers. If a callback synchronously calls `notify` on the **same** hook, the inner call's epilogue:

1. resets `dispatching` to `false` while the outer walk is still in progress — a subsequent `unhook` during the outer walk then splices immediately instead of tombstoning; and
2. compacts tombstones by splicing, shifting indices under the outer loop.

Either way a subscriber can be silently skipped.

## Status: latent, not live

All current callers were traced: `setBattleStage` inside stage handlers is always past an `await`, and `notifyNewEnemyLoaded` → `setBattleStage` nests *different* hooks (each with its own flag). So there is no synchronous same-hook re-entry today — this is the same latent class as the tracked #2047 (RenderEngine stop→start). But the hook is the substrate of the battle/tick pub-sub, so a future synchronous stage-change chain would corrupt dispatch silently.

## Fix direction

Replace the boolean with a depth counter and compact tombstones only when depth returns to 0 (same shape as the #2047 fix), plus a unit test for the nested-notify case.

ginderjeremiah:
## Summary

Closes #2093.

`createHook.notify` (`UI/src/lib/common/hooks.ts`) used a boolean `dispatching` flag to defer unhook-during-dispatch splices (tombstone instead, compact after). If a callback synchronously called `notify` on the *same* hook, the inner call's epilogue reset `dispatching` to `false` while the outer walk was still mid-loop. A subsequent `unhook` during the outer walk would then splice immediately (instead of tombstoning) and shift indices out from under the in-progress `for` loop — silently skipping subscribers or, in the case exercised by the new test, reading past the truncated array.

This was flagged as latent (no current caller triggers a synchronous same-hook re-entry), but the hook is the substrate of the battle/tick pub-sub, so a future synchronous stage-change chain would corrupt dispatch silently.

## Fix

- Replaced the boolean `dispatching` flag with a `dispatchDepth` counter, incremented/decremented around the walk (via `try`/`finally`).
- Tombstone compaction and clearing `hasPendingRemovals` now only run once `dispatchDepth` returns to `0`, i.e. after the *outermost* dispatch completes — mirroring the shape of the #2047 `RenderEngine` stop→start fix (generation/handle-based reentrancy guard instead of a boolean).
- `unhook`'s defer-vs-splice check now reads `dispatchDepth > 0` instead of the old boolean.

## Test plan

- [x] Added `UI/src/tests/lib/common/hooks.test.ts` — no test file existed for this primitive before. Covers ordering, self-unhook mid-dispatch, unhook-ahead-of-walk-index, add-during-dispatch, and the nested-notify regression case (which reproduces the exact corruption described in the issue against the old boolean implementation).
- [x] `npx vitest run` — full UI unit suite: 303 files / 3798 tests passing.

---
_Generated by [Claude Code](https://claude.ai/code/session_015UXww5eUBw3T1DjqbYn9nw)_

### DIFF
--- a/UI/src/lib/common/hooks.ts
+++ b/UI/src/lib/common/hooks.ts
@@ -8,7 +8,10 @@ interface HookTracker<T extends unknown[]> {
 }
 
 export const createHook = <T extends unknown[] = []>() => {
-	let dispatching = false;
+	// A depth counter (not a boolean) so a callback that synchronously re-notifies this same
+	// hook doesn't reset the in-progress-dispatch flag early: the outer walk's tombstone
+	// compaction must wait until every nested dispatch has also finished.
+	let dispatchDepth = 0;
 	let hasPendingRemovals = false;
 	const trackers: HookTracker<T>[] = [];
 
@@ -18,18 +21,21 @@ export const createHook = <T extends unknown[] = []>() => {
 		// the index walk isn't disturbed: an unhook only tombstones (`removed`) instead
 		// of splicing, and the snapshotted `count` excludes any subscriber a callback
 		// appends mid-dispatch — keeping it from firing on an event that predates it.
-		// Tombstones are compacted out once after dispatch; appends survive it.
-		dispatching = true;
-		const count = trackers.length;
-		for (let i = 0; i < count; i++) {
-			const tracker = trackers[i];
-			if (!tracker.removed) {
-				tracker.callback(...data, tracker.unhook);
+		// Tombstones are compacted out once after the outermost dispatch; appends survive it.
+		dispatchDepth++;
+		try {
+			const count = trackers.length;
+			for (let i = 0; i < count; i++) {
+				const tracker = trackers[i];
+				if (!tracker.removed) {
+					tracker.callback(...data, tracker.unhook);
+				}
 			}
+		} finally {
+			dispatchDepth--;
 		}
-		dispatching = false;
 
-		if (hasPendingRemovals) {
+		if (dispatchDepth === 0 && hasPendingRemovals) {
 			for (let i = trackers.length - 1; i >= 0; i--) {
 				if (trackers[i].removed) {
 					trackers.splice(i, 1);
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: reentrancy-safe `createHook.notify` (closes #2093)

**Verdict:** Approve — the fix is correct, minimal, and matches the fix direction the issue laid out (depth counter, compact only at depth 0). _(Posted as a comment because I'm the PR author and GitHub won't accept a self-approval; treat this as an approve.)_

### Assessment

- Boolean `dispatching` → `dispatchDepth` counter, incremented/decremented around the walk. Tombstone compaction and the `hasPendingRemovals` reset are now gated on `dispatchDepth === 0`, so a synchronous same-hook re-entry can no longer splice the trackers array out from under an outer walk. This is exactly the corruption the issue described, and it resolves it.
- `unhook`'s defer-vs-splice branch correctly switches to `dispatchDepth > 0`.
- Bonus: wrapping the walk in `try`/`finally` also closes a latent robustness gap — the old code set `dispatching = false` outside any `finally`, so a throwing subscriber would have stranded the flag permanently. The new code always restores the depth.
- Comment quality is good: the "depth counter, not a boolean" rationale and the "compacted out once after the outermost dispatch" note are the right things to leave for a future reader of a subtle primitive.

### Test coverage

Solid. The new `hooks.test.ts` (this primitive had no test file before) covers ordering, self-unhook mid-dispatch, unhook-ahead-of-index, add-during-dispatch, and — most importantly — the nested-notify regression that reproduces the exact bug against the old boolean implementation. No backend mirror is required: this is a pure frontend pub/sub primitive, not parity-bearing battle logic.

### What's good

Well-scoped tech-debt fix that addresses a latent-but-real corruption on the substrate of the battle/tick pub-sub, with a regression test that pins the behavior. Nothing to change.

---

Issue #560, opened 2026-06-14
Drain the player update queue on synchronizer startup (write-behind safety)


ginderjeremiah:
Surfaced verifying the caveat in #552 (FireAndForget wake-publish). Part of the write-behind spike #548.

## Finding

`DataProviderSynchronizer.StartAsync` → `InitSubscriber` only **subscribes** to the channel; it does not drain whatever is already on `PlayerUpdateQueue`. Redis pub/sub wakes are at-most-once (and now fire-and-forget after #552), so an item enqueued while no subscriber was connected — e.g. across an instance restart, or if a wake is dropped during a brief subscriber outage — waits for the *next* publish to trigger a drain. `ProcessQueue` drains the whole queue per wake, and continuous player saves keep wakes coming, so in practice items aren't stranded long — but the gap is real at the tail (queue has items + no subsequent publish).

## Verdict on #552's caveat

FireAndForget does **not** introduce this — the wake was already best-effort (awaiting only confirmed the command was sent, not delivered; durability is the awaited LPUSH). The cheap, sufficient fix is a **one-time drain on startup** (call `ProcessQueue` once after subscribing). A periodic timer is likely overkill given continuous wakes, but worth considering if a fuller safety net is wanted.

## Scope

- Drain the queue once in `StartAsync` after the subscription is established (guard against the subscription's own first wake double-draining — `ProcessQueue` is already idempotent-safe since it pops).

Issue #552, opened 2026-06-14
Trim write-behind publish: FireAndForget the wake and batch events into one LPUSH


ginderjeremiah:
Part of the battle-end I/O write-behind spike #548 (see `docs/spikes/548-battle-end-io-write-behind.md`). **Independent — no dependency on #550 / the progress refactor.**

## What

Two cheap reductions to the write-behind publish cost in `RedisPubSubService.Publish(channel, queueName, queueData)`:

- It currently does an **awaited** LPUSH (`AddToQueueAsync`) **and** an **awaited** PUBLISH per event. The PUBLISH is only a wake signal — the data is already durably on the queue — so make it `FireAndForget`.
- Batch the events from a single save into **one multi-value LPUSH** instead of one round-trip per event.

Keep the LPUSH itself awaited (it carries the durable data); only the wake PUBLISH becomes fire-and-forget.

## Why

Trims Redis round-trips from every write-behind save — benefits both the existing player path and the new progress path (#551).

## Caveat to verify

Confirm a coalesced or dropped wake can't strand the queue: `DataProviderSynchronizer` drains the whole queue per wake, so the next publish wakes it — but verify there's no scenario where a save's only wake is lost with nothing following (and whether a periodic safety drain is warranted).

PR #2157, opened 2026-07-18
Dispatch the write-behind startup drain to the background instead of awaiting it

### LINKED ISSUE #2096: Write-behind startup drain runs inside StartAsync, blocking host readiness on the entire fleet-shared backlog
## Problem

`DataProviderSynchronizer` is a hosted service and its `StartAsync` **awaits** the startup drain (`Game.DataAccess/DataProviderSynchronizer.cs:115-166`, the `ProcessQueue` await at `:158`; registered in `Game.Api/Startup.cs:292`). ASP.NET Core starts Kestrel only after every hosted service's `StartAsync` completes, so the instance accepts **no traffic until the whole player-update queue is empty**.

The queue is fleet-shared: in a rolling deploy each new instance blocks its own boot on a backlog that healthy instances are already draining. A large backlog (e.g. accumulated during a DB slowdown, drained at 4-way concurrency with one `SaveChangesAsync` round trip per event — the ceiling #1739 tracks) turns into minutes of failed health checks on every new instance.

## Why the await isn't needed

The correctness goal ("an item enqueued while no subscriber was connected is applied promptly") only requires **subscribe + reclaim** to precede the drain — the drain itself can be dispatched as a background wake (the same shape as a pub/sub wake) rather than awaited inline. The boot-blocking migration/seed/cache loads are deliberate but bounded; this one is unbounded and grows with fleet backlog.

## Fix direction

Keep subscribe/reclaim in `StartAsync`, kick the initial drain onto the existing background wake path, and cover with an integration test that a slow drain doesn't gate startup.

ginderjeremiah:
## Summary

`DataProviderSynchronizer.StartAsync` awaited the whole startup drain (`ProcessQueue`) before returning. ASP.NET Core holds Kestrel closed until every hosted service's `StartAsync` completes, and the player-update queue is fleet-shared, so a large backlog (e.g. accumulated during a DB slowdown, drained at 4-way concurrency with one `SaveChangesAsync` round trip per event) turned into minutes of failed health checks on every new instance during a rolling deploy — even though already-healthy instances were already draining the same backlog.

- `StartAsync` still subscribes and awaits the orphaned-item reclaim inline (both are cheap/bounded), but now dispatches the drain itself to the background rather than awaiting it — the same shape as a pub/sub-triggered drain.
- The correctness goal this drain exists for ("an item enqueued while no subscriber was connected is applied promptly, since Redis pub/sub wakes are at-most-once") only requires subscribe + reclaim to precede the drain, not the drain itself to finish before startup completes — so it's preserved.
- The backgrounded drain runs on the synchronizer's own stopping token (not the host's startup token, which is no longer meaningful once `StartAsync` has returned), and any exception escaping it is logged rather than becoming an unobserved task fault.
- `StopAsync`'s existing bounded wait on the drain gate is unchanged and still covers shutdown for the backgrounded drain exactly as it already did for a pub/sub-triggered one.

## Changes

- `Game.DataAccess/DataProviderSynchronizer.cs`: split `StartAsync` into an awaited subscribe+reclaim followed by a fire-and-forget `RunStartupDrain`.
- `Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs`:
  - Added `StartAsync_SlowDrainInProgress_ReturnsWithoutWaitingForItToFinish`, which pins the core fix: `StartAsync` completes even while a gated (never-finishing-on-its-own) drain is still parked mid-apply in the background, and shutdown still waits for it cleanly.
  - Fixed two existing real-Postgres/Redis integration tests (`StartAsync_DrainsItemsAlreadyOnQueue_WithoutWaitingForAWake`, `StartAsync_ReclaimsInFlightItemOrphanedByCrashedRun_AndApplies`) that asserted on drain results immediately after `await StartAsync(...)` — now racy since the drain no longer completes before `StartAsync` returns. These now poll for the drain's effect to land (the same `PollingHelper` pattern already used elsewhere for a fire-and-forget write racing an immediate read) rather than calling `StopAsync` immediately, which would itself race the in-flight reserve (`StopAsync` cancels the stopping token first, and that token is threaded cooperatively into the reserve/reclaim ops).
  - Updated `StopAsync_WedgedReserveDuringDrain_UnwindsPromptlyViaThreadedToken` to drop an assertion (`Assert.False(startTask.IsCompleted)`) that encoded the old blocking behavior this PR intentionally removes.
- `docs/backend-persistence.md`: updated the "queue is drained once on startup" and shutdown-token sections to describe the new backgrounded-drain shape.

## Test plan

- [x] `dotnet test Game.Application.Tests` (filtered to `DataProviderSynchronizerTests`) — 38/38 passing, run repeatedly to check for flakiness
- [x] `dotnet test Game.sln` (full suite, real Postgres/Redis containers) — 3307/3307 passing

Closes #2096

---
_Generated by [Claude Code](https://claude.ai/code/session_01BY49AqWXCGPQ1L3HCmZpX8)_


ginderjeremiah (commented):
## Review: dispatch the write-behind startup drain to the background (closes #2096)

**Verdict:** Approve (submitted as a comment since I'm the PR author) — the change correctly decouples the fleet-shared backlog drain from host readiness without weakening any of the drain's correctness or shutdown guarantees, and it's well-covered.

### Summary of what I checked

The fix is exactly the direction #2096 called for: `StartAsync` still awaits `InitSubscriber` + the orphaned-item reclaim (both bounded), then dispatches the drain via `_ = RunStartupDrain(queue)` instead of awaiting `ProcessQueue`. The correctness goal (#552 — an item enqueued while no subscriber was connected is applied promptly rather than stranded until the next publish) only requires subscribe + reclaim to precede the drain, and that ordering is preserved.

The concurrency-sensitive parts all hold up:

- **No unobserved task fault.** `RunStartupDrain` wraps `ProcessQueue` in a try/catch that swallows the clean-stop OCE (`_stopping` cancelled) and logs any other exception, so the discarded task can never fault silently. `ProcessQueue`'s `finally` always releases `_drainGate`, so a fault can't wedge the gate.
- **Shutdown still waits.** `StopAsync` continues to gate on `_drainGate` within the bounded timeout, which the backgrounded drain acquires exactly like a pub/sub-triggered one — so the "finish the in-flight apply, don't cut off mid-`SaveChanges`" guarantee is intact. A stop that lands before the drain starts makes it a no-op (`_stopping` already cancelled → `ProcessQueue`'s while-guard is false).
- **Background drain runs on `_stopping.Token`, not the host startup token** — correct, since the startup token is no longer meaningful once `StartAsync` returns. The pre-dispatch `if (_stopping.IsCancellationRequested) return;` guard is keyed to `_stopping` for the same reason.
- **No new double-apply hazard** — the drain gate + `_drainRequested` coalescing still serialize the startup pass with the subscription's first wake, and the reserve/acknowledge read is idempotent (#560), all unchanged.

### Test coverage

Solid and appropriately scoped (backend-only change; no battle logic or frontend surface, so no mirrored frontend tests are owed).

- New `StartAsync_SlowDrainInProgress_ReturnsWithoutWaitingForItToFinish` pins the actual fix: `StartAsync` returns within a bounded 2s even while a gated drain is parked mid-acknowledge in the background, and shutdown then waits for it cleanly. The bounded `WaitAsync` means a regression back to an inline await hangs the assertion rather than passing silently — nice.
- `StopAsync_WedgedReserveDuringDrain_...` correctly drops the now-invalid `Assert.False(startTask.IsCompleted)` and awaits `startTask` promptly.
- The two real-Postgres/Redis tests (`StartAsync_DrainsItemsAlreadyOnQueue_...`, `StartAsync_ReclaimsInFlightItemOrphanedByCrashedRun_...`) correctly switch from asserting-immediately to polling for the drain's effect (via `PollingHelper`), since the drain no longer completes before `StartAsync` returns — and they poll rather than calling `StopAsync` immediately, which would race the in-flight reserve against the stopping token. `AsNoTracking().SingleAsync` avoids identity-map staleness across polls.

### Docs

`backend-persistence.md` is updated accurately — the "drained once on startup" and shutdown-token sections now describe the backgrounded-drain shape and the split between the awaited reclaim (honours the host startup token) and the backgrounded drain (stopping token alone).

---

Issue #2129, opened 2026-07-18
PlayerUpdateDeadLetters.ReplayAllAsync re-enqueues entries it classifies as poison, and replay cost is quadratic in queue depth


ginderjeremiah:
Two related issues in the player-update DLQ Ops surface (`Game.DataAccess/Repositories/Admin/PlayerUpdateDeadLetters.cs`):

1. **Poison entries are replayed anyway.** `ReplayAllAsync` (:45, 57-58) targets the entire snapshot with no filter, while `Classify` (:110-136) explicitly distinguishes `Malformed`/`UnknownEventType` from `Replayable`. Replaying a poison entry pushes it onto the live player queue, where the synchronizer immediately re-dead-letters it: it's counted in `ReplayedCount`, shuffled to the DLQ tail (losing position/age), and re-trips the dead-letter-depth warning. This is inconsistent with the sibling socket-command DLQ contract (`docs/backend-persistence.md`: non-replayable entries "stay queued through both peek and replay"). Fix: filter `ReplayAllAsync` to `Replayable` entries.

2. **Replay is O(N²) + 2N sequential round trips.** Each entry costs an awaited `RPUSH` + `LREM` pair (:61-71), and `LREM` is an O(N) scan of the DLQ list. Fine for a handful of poison messages — but the documented mass-dead-letter scenario (a DB outage dead-lettering every in-flight write, #2097, whose remediation is precisely "manual replay") would make this the recovery bottleneck: thousands of entries → tens of seconds of sequential Redis calls inside one admin HTTP request. Worth folding the batching/pipelining into whatever #2097's fix becomes rather than fixing standalone.

PR #2158, opened 2026-07-18
Park exhausted-retry player-update events on DB outage instead of dead-lettering them

### LINKED ISSUE #2097: A DB outage longer than the ~1s retry budget mass-dead-letters every in-flight player write — recovery requires manual replay
## Problem

The write-behind retry policy is tuned for blips — 3 attempts, 200ms base backoff ≈ 0.6s total (`Game.DataAccess/PlayerUpdateRetryPolicy.cs:13`; retry-then-dead-letter loop at `Game.DataAccess/DataProviderSynchronizer.cs:548-593`) — and there is **no distinction between a poison event and an infrastructure-level failure**.

During any Postgres outage/failover longer than a second: live play continues on Redis, every save wakes the drain, and every event that reaches a handler exhausts its retries and is **dead-lettered**. When the DB recovers, nothing self-heals — Postgres silently lacks the entire outage window's writes until an operator notices the depth warning and manually replays via the admin surface. Worse, the player-key TTL means a player who goes dormant before that replay rehydrates from the lagging DB copy, making the regression effectively permanent for them.

The docs frame dead-lettering as for events that "could not be persisted", with transient failures expected to "succeed on a second attempt" — the outage-wide consequence appears unweighed.

## Fix direction

Classify infrastructure failures (connection/timeout `NpgsqlException`s) separately from poison events and **park the drain** (leave items on the durable queue — exactly what it's for) with periodic re-probe, instead of dead-lettering. A routine transient outage then rides out with zero operator action, while genuinely poisoned payloads still dead-letter as today. Distinct from #1739 (batching) and from the dead-letter Ops tooling, which mitigates but doesn't remove the manual step.

ginderjeremiah:
## Summary

`DataProviderSynchronizer.ProcessMessage` dead-letters any player-update event whose retries are exhausted, with no distinction between a poison event and an infrastructure-level failure (`Game.DataAccess/PlayerUpdateRetryPolicy.cs`: 3 attempts, ~0.6s total). During a Postgres outage/failover longer than that budget, every in-flight write fails the same way and gets mass-dead-lettered — nothing self-heals, and an operator has to notice the dead-letter-depth warning and manually replay the whole backlog via the admin Ops surface. Worse, a player who goes dormant before that replay rehydrates from the lagging DB copy, making the regression effectively permanent for them.

Closes #2097.

## Approach

Rather than reworking the drain loop's concurrency/dispatch model (a much larger, riskier change touching the ~30 tests pinned to that structure), this reuses the drain's **existing, already-tested fault-handling path**:

- `InfrastructureFailureClassifier.IsInfrastructureFailure(Exception)` (new, `Game.DataAccess/InfrastructureFailureClassifier.cs`) distinguishes a database-**infrastructure** failure (`NpgsqlException`/`TimeoutException` that is *not* a `PostgresException` — the connection itself is unreachable or a command timed out) from a genuine **per-write** failure (a `PostgresException` — the server responded, so the write failed on its own merits: a deadlock, constraint violation, serialization failure). This mirrors the existing `DbUpdateExceptionExtensions.IsUniqueViolation` pattern already used elsewhere in `Game.DataAccess`.
- In `ProcessMessage`'s exhausted-retry terminal catch, an infrastructure failure is no longer dead-lettered — it's **rethrown**. That fault propagates out of `ProcessMessage` into `ProcessReservedItemAsync`'s task, which `SettleInFlightItemsAsync` already knows how to handle: the item is left reserved (never acknowledged) rather than dead-lettered, and is logged as "faulted... will be reclaimed and re-applied." This is exactly the same outcome the drain already produces for a Redis-op fault — no new machinery needed.
- Once reserved-but-faulted, the item is picked back up by the **existing** at-least-once mechanisms: the opportunistic stranded-processing reclaim (once it's dwelled on the processing list past the grace period) or the next startup's reclaim. No dedicated re-probe loop was added — normal ongoing gameplay traffic keeps triggering fresh drain passes, and each one's reclaim check re-attempts the parked item once the grace period has elapsed since it started dwelling, giving a natural ~30s-cadence retry for free.
- A genuine per-write failure (`PostgresException`) is unaffected — still dead-lettered exactly as before.

Documented the new classification and parking behavior in `docs/backend-persistence.md`'s "Queue processing resilience" section, and updated `ProcessMessage`/`SettleInFlightItemsAsync`'s doc comments to describe the new fault path.

## Note on scope

This only changes what happens *after* retries for a **live event mid-drain**. It doesn't touch `PlayerUpdateDeadLetters.ReplayAllAsync` (already tracked separately in #2129, which also notes the same DB-outage backlog would make its O(N²) replay a bottleneck) — that's a different, non-overlapping surface (the manual replay path for entries *already* dead-lettered), and is better folded into whatever #2129's fix becomes rather than touched here.

## Test plan

- [x] Added `ProcessQueue_InfrastructureFailureDuringHandling_RetriesThenLeavesItReservedInsteadOfDeadLettering` (`[Theory]` over `NpgsqlException` and `TimeoutException`) — asserts the item is retried the full attempt budget, then left reserved on the processing list (not dead-lettered), with a distinctly-worded outage log.
- [x] Added `ProcessQueue_GenuinePostgresErrorAfterRetries_DeadLettersRatherThanLeavingItReserved` — asserts a `PostgresException` (e.g. a deadlock) is *not* misclassified as an outage and still dead-letters as before.
- [x] Full backend test run: `Game.Application.Tests` (1015), `Game.Core.Tests` (1396), `Game.Infrastructure.Tests` (68), `Game.Api.Tests` (831) — all passing, no regressions to the existing dead-letter/retry/reclaim coverage.

https://claude.ai/code/session_0128oLgwBgMeouM1bcir4Jdh

---
_Generated by [Claude Code](https://claude.ai/code/session_0128oLgwBgMeouM1bcir4Jdh)_


Game.DataAccess/InfrastructureFailureClassifier.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// turn a self-resolving outage into a mass manual-replay chore instead.
+    /// </summary>
+    internal static class InfrastructureFailureClassifier
+    {
+        public static bool IsInfrastructureFailure(Exception ex) => ex switch
+        {
+            // A PostgresException is an error response FROM the server, so the connection itself is healthy —
+            // the write failed on its own merits (checked before the broader NpgsqlException case below, since
+            // PostgresException derives from it).
+            PostgresException => false,
```
ginderjeremiah, line 25: **Should-fix:** treating *every* `PostgresException` as a per-write failure is coarser than the issue asks for. #2097's title is "outage/**failover**", and a graceful Postgres failover/restart doesn't always sever the connection first — the server frequently responds with a `PostgresException` carrying an availability SQLSTATE before dropping it: `57P01` admin_shutdown, `57P03` cannot_connect_now, or `53300` too_many_connections during the reconnect storm on recovery. Those are infrastructure-wide and self-resolving — exactly the mass-dead-letter this PR is meant to stop — yet they'd be classified `false` here and dead-lettered.

The existing `DbUpdateExceptionExtensions.IsUniqueViolation` already keys off `SqlState`, so the pattern for a finer check is right next door. Consider carving out the class-`08` (connection_exception) / class-`57` (operator_intervention) / `53300` SQLSTATEs as infrastructure rather than blanket-classifying all `PostgresException`s as per-write. If the blunt cut is a deliberate scoping decision, worth a one-line note here (and in the doc) that server-reported outage states are knowingly excluded.

---

Game.DataAccess/InfrastructureFailureClassifier.cs
```diff
@@ -0,0 +1,27 @@
...
+            NpgsqlException => true,
+            TimeoutException => true,
+            _ => ex.InnerException is { } inner && IsInfrastructureFailure(inner),
```
ginderjeremiah, line 28: **Should-fix (test coverage):** this `InnerException` recursion branch is the branch that actually fires in production, but nothing exercises it. A real outage happens inside `HandleEvent` → `DispatchAsync` → EF `SaveChanges`, and EF wraps DB failures in a `DbUpdateException` — the sibling `IsUniqueViolation` reads exactly that shape (`ex.InnerException is PostgresException`). So the live shapes are `DbUpdateException(inner: NpgsqlException)` (→ should park) and `DbUpdateException(inner: PostgresException)` (→ should dead-letter), and both go solely through this line. The two tests throw *bare* `NpgsqlException` / `TimeoutException` / `PostgresException` from `CreateScope()`, so they only cover the three top-level arms and never this recursion. Please add a scenario where `AlwaysThrowingServiceProvider` throws a `DbUpdateException` wrapping each, so the actual EF exception shape — and the `PostgresException`-before-`NpgsqlException` ordering that this whole classifier hinges on — is verified end to end.

---

ginderjeremiah (commented):
## Review: park exhausted-retry player writes on a DB outage instead of dead-lettering them (#2097)

**Verdict: Approve _with should-fixes_** (posted as a comment — reviewer identity matches the PR author, so GitHub won't accept a gating event; the real call is Approve). No blocking findings. The core mechanism is sound and the reuse of the existing fault path is the right instinct — but there are two should-fixes below I'd like addressed before merge rather than as follow-ups, since both are squarely within this change's surface.

### What's good
- **Reusing the existing settle/reclaim path instead of building a re-probe loop is exactly right.** Rethrowing so the item is left reserved, and letting `SettleInFlightItemsAsync` + the opportunistic stranded-processing reclaim carry it, avoids new machinery and gets the ~grace-period retry cadence for free. The scope note (deliberately not reworking the drain's concurrency model, and leaving #2129's replay path alone) shows good restraint.
- **Catch ordering is correct:** the `attempt < MaxAttempts` arm still catches infra failures on non-final attempts, so they're retried the full budget first, and only the terminal attempt reaches the new classifier arm — placed before the blanket dead-letter arm. `PostgresException => false` correctly precedes `NpgsqlException => true` despite the inheritance, and the comment calls out why.
- Docs (`backend-persistence.md`) and the `ProcessMessage` / `SettleInFlightItemsAsync` doc-comments are updated to describe the new fault path clearly. Layering is right — an Npgsql-specific classifier belongs in `Game.DataAccess`, mirroring `DbUpdateExceptionExtensions`.

### Should-fix
- **Classifier is too coarse on `PostgresException` — `InfrastructureFailureClassifier.cs:21`.** Every `PostgresException` is treated as per-write, but a graceful failover/restart (the "failover" half of #2097's title) often responds with a server-reported availability state — `57P01` admin_shutdown, `57P03` cannot_connect_now, `53300` too_many_connections — before the connection drops. Those are infra-wide and self-resolving, yet still mass-dead-letter under this cut. Suggest carving out the class-`08`/class-`57`/`53300` SQLSTATEs (the `IsUniqueViolation` `SqlState` pattern is right next door), or explicitly documenting the exclusion if it's deliberate. Per CLAUDE.md, the issue's stated "connection/timeout `NpgsqlException`s" scope shouldn't be taken as fully analyzed — this residual gap is real.
- **Untested real-world exception shape — `InfrastructureFailureClassifier.cs:24`.** A production outage surfaces from EF `SaveChanges` as a `DbUpdateException` wrapping the Npgsql/Postgres exception (confirmed by the sibling `IsUniqueViolation`, which reads `ex.InnerException is PostgresException`). That path goes solely through the `InnerException` recursion arm, which no test touches — both tests throw *bare* exceptions from `CreateScope()`, covering only the three top-level arms. Add a scenario where `AlwaysThrowingServiceProvider` throws a `DbUpdateException` wrapping each, so the actual EF shape and the arm ordering are verified.

### Test coverage
Otherwise solid and matching the project's Detroit/classical style: the `[Theory]` over `NpgsqlException`/`TimeoutException` and the `PostgresException` counter-case assert the right observable outcomes (retry count, outage-vs-dead-letter log wording, `Empty` dead-letter queue, `ProcessingCount == 1` vs `0`). Pure-backend persistence change with no battle-logic or frontend surface, so no frontend mirror is needed. The one gap is the wrapped-`DbUpdateException` shape noted above.

### Nit
- During a prolonged outage with many active players, every parked item logs an `Error` per reclaim cycle (~grace-period cadence) and re-faults on each pass — potentially a flood of error logs every ~30s. It's the documented tradeoff and it's bounded, but consider summarizing/throttling the outage log (e.g. one aggregate line per reclaim round) so an outage doesn't drown the error stream. Fine as a follow-up if you'd rather not expand scope here.


---

ginderjeremiah:
Both should-fixes addressed in fd2f4c0:

- `InfrastructureFailureClassifier` now carves out a narrow set of server-reported unavailability SQLSTATEs (connection_exception class, admin/crash shutdown, cannot-connect-now/recovering, too-many-connections) as infrastructure failures even when wrapped in a `PostgresException`, while deliberately keeping `query_canceled` and `database_dropped` classified as per-write (a cancelled statement isn't an outage; a dropped database isn't self-resolving).
- Added `DbUpdateException`-wrapped test cases to both theories, covering the `InnerException` recursion arm that the real EF `SaveChanges` failure shape actually goes through.

Full backend suite re-run clean (1023 tests in Game.Application.Tests, no regressions).

Filed #2159 for the log-flooding nit as a follow-up, per the review's suggestion not to expand this PR's scope.

---

ginderjeremiah:
## Re-review (changes since the prior review) — Approve, merging

The earlier review (commit `149f663`) landed an Approve with two should-fixes and one nit. Re-reviewed the delta in `fd2f4c0`; both should-fixes are resolved and the code is sound.

**Should-fix 1 — classifier too coarse on `PostgresException`:** Resolved well, and the fix is actually finer than my original suggestion. Rather than blanket-classifying class-`08`/class-`57`/`53300`, it carves out only the server-reported *unavailability* states — class `08` (connection_exception), `57P01`/`57P02`/`57P03` (admin/crash shutdown, cannot-connect-now), and `53300` (too-many-connections) — while deliberately keeping `57014` query_canceled (a single cancelled statement, not an outage) and `57P04` database_dropped (not self-resolving — parking would retry forever) as per-write. The exclusions are the right call and are documented inline.

**Should-fix 2 — untested EF `DbUpdateException` shape:** Resolved. Both theories now include `DbUpdateException`-wrapped cases, exercising the `InnerException` recursion arm and the `PostgresException`-before-`NpgsqlException` ordering the classifier hinges on — the shape that actually fires from `SaveChanges` in production.

**Nit — outage-log flooding per reclaim cycle:** Filed as follow-up #2159 (tech debt / claude / scope: small), per the review's suggestion not to expand scope here. Bounded, documented tradeoff — appropriate as a follow-up.

Classifier arm ordering is correct (`PostgresException` matched before the broader `NpgsqlException` despite the inheritance; recursion arm is last), the terminal catch sits ahead of the blanket dead-letter arm and is only reached on the final attempt, and the docs accurately describe the new park-vs-dead-letter split. This faithfully solves #2097 without new re-probe machinery.

Verdict: **Approve.** All CI checks green, both review threads resolved, no remaining should-fixes, follow-up filed. Merging.

---

PR #2160, opened 2026-07-18
Redraw battle seed on collision with LastCreditedBattleSeed

### LINKED ISSUE #2112: BattleAlreadyCredited matches on seed alone; CreateBattleSeed never excludes the player's LastCreditedBattleSeed

ginderjeremiah:
## Summary

`BattleAlreadyCredited` (`Game.Application/Services/BattleService.cs`) matches an in-progress battle against the idempotency backstop purely by `state.BattleSeed == player.LastCreditedBattleSeed`. `CreateBattleSeed` drew a fresh cryptographic `uint` with no exclusion, so a 2⁻³² chance the newly-drawn seed for a player's *next* battle collided with their own `LastCreditedBattleSeed` would make `EndBattleVictory`/`EndBattleLoss`/the abandon path treat that brand-new, legitimately-fought battle as a stale re-presentation of the already-credited one and wrongly reject/void its claim.

Closes #2112.

## Approach

Of the issue's two suggested directions, this implements the seed-generation fix (re-roll on collision) rather than widening `BattleAlreadyCredited` to also compare enemy id — the enemy-id approach would need a new persisted `LastCreditedEnemyId`-shaped field (migration, entity, domain event, cache mapper — the same plumbing `LastCreditedBattleSeed` itself needed in #1874), whereas redrawing at the point of generation closes the root cause with no new schema.

- `BattleService.CreateBattleSeed` now takes an optional `exclude` parameter and redraws until the result differs from it, via an extracted `DrawSeed(Func<uint> draw, uint? exclude)` helper — pulled out so the redraw loop can be pinned by a unit test with a stubbed generator (the real 2⁻³² collision can't be triggered through the actual CSPRNG).
- `StartBattle` and `StartBossBattle` now call `CreateBattleSeed(player.LastCreditedBattleSeed)`.
- `OfflineProgressService`'s `SeedSource` delegate (handed to the pure-domain `OfflineProgressSimulator`) now closes over `player.LastCreditedBattleSeed` too. That value is stable for the whole offline pass — mid-window credits land on the `PlayerProgress` aggregate, not `Player` — so excluding it once per draw is sufficient; a collision between two in-window battle seeds is harmless since only a seed matching the *persisted* credited value could ever be misread as already-credited by a later live claim.

## Test plan

- [x] `CreateBattleSeed_NeverEqualsProvidedExclude` — statistical pin on the public entry point (1000 draws against a fixed exclude).
- [x] `DrawSeed_RedrawsUntilPastTheExcludedValue` / `DrawSeed_NullExclude_ReturnsFirstDrawUnconditionally` — deterministic pins on the redraw loop via a stubbed draw source.
- [x] `dotnet test Game.sln` — full suite green (Game.Core.Tests 1396, Game.Infrastructure.Tests 68, Game.Application.Tests 1015, Game.Api.Tests 831 — 3310 total, 0 failures).

---
_Generated by [Claude Code](https://claude.ai/code/session_017ZUHC45MsX9awHa2j9DZau)_

### DIFF
--- a/Game.Application.Tests/Services/CreateBattleSeedTests.cs
+++ b/Game.Application.Tests/Services/CreateBattleSeedTests.cs
@@ -27,5 +27,44 @@ namespace Game.Application.Tests.Services
             Assert.True(seeds.Distinct().Count() > 1,
                 "Expected the seed generator to produce varied values from a non-time entropy source.");
         }
+
+        // #2112: a fresh seed equal to the player's LastCreditedBattleSeed must never be handed out, since
+        // BattleAlreadyCredited would then treat the brand-new battle as a stale re-presentation of the
+        // already-credited one. The real CSPRNG can't be made to collide on demand (2^-32), so this pins the
+        // public entry point's exclude plumbing statistically, and DrawSeed below pins the redraw loop itself
+        // deterministically via a stubbed generator.
+        [Fact]
+        public void CreateBattleSeed_NeverEqualsProvidedExclude()
+        {
+            var exclude = BattleService.CreateBattleSeed();
+
+            var seeds = Enumerable.Range(0, 1000)
+                .Select(_ => BattleService.CreateBattleSeed(exclude))
+                .ToList();
+
+            Assert.DoesNotContain(exclude, seeds);
+        }
+
+        [Fact]
+        public void DrawSeed_RedrawsUntilPastTheExcludedValue()
+        {
+            var draws = new Queue<uint>([5u, 5u, 5u, 7u]);
+
+            var result = BattleService.DrawSeed(draws.Dequeue, exclude: 5u);
+
+            Assert.Equal(7u, result);
+            Assert.Empty(draws);
+        }
+
+        [Fact]
+        public void DrawSeed_NullExclude_ReturnsFirstDrawUnconditionally()
+        {
+            var draws = new Queue<uint>([5u, 7u]);
+
+            var result = BattleService.DrawSeed(draws.Dequeue, exclude: null);
+
+            Assert.Equal(5u, result);
+            Assert.Equal(7u, draws.Peek());
+        }
     }
 }
--- a/Game.Application/Services/BattleService.cs
+++ b/Game.Application/Services/BattleService.cs
@@ -119,7 +119,7 @@ namespace Game.Application.Services
             // (see PrepareNextIdleBattle) — and, for the abandon case, is what makes the cooldown just applied
             // actually pace the replacement battle instead of being immediately bypassed (#1851).
             var battleStartTime = scheduledStartTime ?? abandonedOutcomeCooldown ?? DateTime.UtcNow;
-            var seed = CreateBattleSeed();
+            var seed = CreateBattleSeed(player.LastCreditedBattleSeed);
 
             var enemy = _battleFactory.CreateBattleEnemy(
                 zone,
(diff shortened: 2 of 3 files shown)


ginderjeremiah (commented):
## Review: redraw the battle seed on collision with `LastCreditedBattleSeed` (#2112)

**Verdict: Approve** (posted as a comment — reviewer identity matches the PR author, so GitHub won't accept a gating event; the real call is Approve). No blocking findings, no should-fixes, no nits.

### What's good
- **Right approach, well-justified.** Choosing the seed-generation redraw over widening `BattleAlreadyCredited` to compare enemy id is the correct call, and the PR body defends it well: the enemy-id route would need a whole new persisted `LastCreditedEnemyId`-shaped field (migration/entity/event/mapper), while redrawing closes the root cause with zero new schema. Fixing at the point of generation is the cleaner cut.
- **Complete coverage of the seed paths.** All four places a battle seed originates now flow through the exclusion: `StartBattle` (`BattleService.cs:122`), the boss path (`:236`), and the offline `SeedSource` (`OfflineProgressService.cs:267`). The fourth — `HandBackPendingBattle` (`:476`) — correctly reuses `pending.Seed`, which was itself drawn through the excluded offline `SeedSource`, so the handed-back battle's later live claim can't false-positive against the stale `LastCreditedBattleSeed` either. Nothing was missed.
- **Offline reasoning checks out.** I verified the load-bearing claim: `OfflineProgressSimulator`'s loop operates on a working snapshot and accumulates into `PlayerProgress` — it never touches `player.LastCreditedBattleSeed` during the pass. So capturing that value once in the closure is sound, and the exclusion isn't merely defensive: it actively prevents the pending battle's seed from equaling the old credited value, which is what would otherwise get the handed-back battle's claim wrongly rejected.
- **Clean domain isolation.** `DrawSeed` keeps the RNG injectable (the pure-domain simulator still receives a `Func<uint>`), and the null-`exclude` path is handled correctly via lifted nullable comparison rather than a null-forgiving operator. Braces present on the `do/while`. Layering unchanged.

### Test coverage
Appropriate and well-targeted. The redraw loop — the actual new logic, unreachable through the real 2⁻³² CSPRNG collision — is pinned deterministically via `DrawSeed` with a stubbed draw source (`DrawSeed_RedrawsUntilPastTheExcludedValue`, `DrawSeed_NullExclude_ReturnsFirstDrawUnconditionally`), and the public entry point's exclude plumbing is covered statistically. Extracting `DrawSeed` purely to make the loop testable is a reasonable, honest seam. No frontend mirror is needed: the seed is server-generated and transmitted to the client as-is, so this touches no shared battle-simulation logic.

A tidy, well-scoped tech-debt fix that closes #2112 at its root. Approving; will merge once CI finishes green.


---

PR #2124, opened 2026-07-18
Extract shared DetailHeader/SaveBar chrome, dedupe Workbench/progression scss

ginderjeremiah:
## Summary

The progression editor (`workbench/progression`) reimplemented the generic Workbench's chrome wholesale instead of sharing it, per #2060:

- **Save bar**: `Progression.svelte` had a near-verbatim copy of `WorkbenchDetail.svelte`'s save bar markup, plus ~90 duplicated lines of scss (`.save-bar`, `.pips`, `.pip`, `.btn`).
- **Detail header**: `progression/DetailHeader.svelte` re-implemented `WorkbenchDetail`'s head-row/spill-badge/tabs markup and styles.
- **List pane**: `ProgressionList.svelte` duplicated `WorkbenchList.svelte`'s row/edge/spill/search styling nearly 1:1.

Besides the line-count cost, this let the two admin surfaces' change-state visuals drift out of sync — exactly the failure mode the issue called out.

## Changes

- Promoted `progression/DetailHeader.svelte` to a shared `components/DetailHeader.svelte`, generalized to also cover `WorkbenchDetail`'s cases it didn't need before (a hard-delete non-retireable entity, and the `deleted`/Restore state). `WorkbenchDetail.svelte` now renders this shared component instead of an inline duplicate head-row/tabs block; `PathDetail`/`TierDetail` updated to import it from its new location.
- Extracted a shared `SaveBar.svelte`, used by both `WorkbenchDetail` and `Progression`.
- Moved the master–detail list-pane skeleton (`.list-pane`/`.list-head`/`.list-scroll`/`.spacer`) into `workbench.scss`, and had `Progression.svelte` opt into the shared `.workbench` scss scope (it previously reimplemented this base layout under its own `.prog` class and never loaded `workbench.scss` at all). `ProgressionList.svelte` now reuses the shared `.spill` and `.search`/`.inp` classes instead of its own copies.
- One-line update to `docs/frontend-admin.md` correcting its claim that the progression editor reuses "the change-state tokens/primitives" — it now actually shares the `DetailHeader`/`SaveBar` chrome *components*, not just design tokens.

Net: -621 lines / +309 lines across 13 files (two of those new files are the shared components themselves).

[...]

Issue #2060, opened 2026-07-16
Progression editor duplicates the Workbench chrome wholesale (~300 lines of save bar, header, list styling)


ginderjeremiah:
## Context

Found in the July 2026 health review (admin Workbench pass).

The progression editor legitimately re-implements *save orchestration* (per frontend-admin.md), but it also copies the presentational chrome:

- Save bar: `Progression.svelte:66-103` is a near-verbatim copy of `components/WorkbenchDetail.svelte:89-128`, plus ~90 lines of duplicated scss (`.save-bar`, `.pips`, `.pip`, `.btn` — `Progression.svelte:252-343`).
- Detail header: `progression/DetailHeader.svelte` re-implements WorkbenchDetail's head-row/spill-badge/tabs markup and styles (`.spill`, `.hdr-action`, `.tabs`, `.tab-dot` duplicated in both).
- List pane: `progression/ProgressionList.svelte:121-316` duplicates `WorkbenchList.svelte`'s row/edge/spill/search styling nearly 1:1.

Cost: ~300 duplicated lines, and the change-state visuals (added/modified/removed pips) can drift between the two admin surfaces.

## Suggested fix

Extract a shared `SaveBar`, `StatusSpill`, and list-row primitive (or at minimum move the duplicated scss into `workbench.scss`).

PR #2162, opened 2026-07-19
Unify Workbench and progression list-row primitives

### LINKED ISSUE #2125: Unify the Workbench and progression list-row primitives
## Context

Follow-up to #2060 (PR #2124). That PR deduplicated the Workbench/progression save bar, detail header, and the list-pane skeleton/spill/search styling, but deliberately left the two list-**row** primitives separate:

- `components/WorkbenchList.svelte` → `.list-row` (flat catalogue row)
- `progression/ProgressionList.svelte` → `.row` (adds a back-button, an ordinal circle, and a path→tier drill-down shape the generic list doesn't have)

They still share the `.spill`/`.row-edge`/status-edge coloring conventions but re-implement the row body 1:1 otherwise.

## Task

Extract a shared list-row primitive (or a small base the progression row composes over) so the row markup/styling isn't duplicated, while preserving progression's drill-down/back/ordinal affordances. This is the deeper, riskier half of #2060 — the two rows have genuinely different shapes, so it needs its own design pass rather than a mechanical move.

## Notes

- Keep the shared change-state edge coloring (`--change-added`/`-modified`/`-removed`) consistent across both, which is the drift #2060 was guarding against.
- Presentation-only; the existing `WorkbenchList`/`ProgressionList` tests should continue to cover the refactored markup.

ginderjeremiah:
## Summary

Follow-up to #2060 (PR #2124), which deduplicated the Workbench/progression save bar, detail header, and list-pane skeleton/spill/search styling but deliberately left the two list-**row** primitives separate. This closes that gap: `WorkbenchList.svelte`'s `.list-row` and `ProgressionList.svelte`'s `.row` re-implemented the same button shell, edge-accent coloring, name/badge/spill layout, and meta line, 1:1.

Closes #2125

## Approach

Extracted a new shared `components/ListRow.svelte`:

- Owns the row shell (button, hover/selected states, dimmed deleted/retired opacity), the change-state edge accent (added/modified/deleted, driven by a single `status` prop instead of each caller computing its own inline style or CSS class), the name text (with a `blank`-placeholder modifier), the badge/spill/warning-triangle row, and the meta line.
- Exposes a `leading` snippet slot for content ahead of the row body — the one genuinely different affordance between the two tools: progression's tier-ordinal circle (and, transitively, the back-button/drill-down state that lives one level up in `ProgressionList` itself, untouched).
- `WorkbenchList` and `ProgressionList` now only supply their per-record data (name, status, warnings, badge, meta) and a `meta` snippet for their differently-shaped meta line (label/value pairs vs. a single formatted string).

One behavioral tightening while unifying: progression's list row previously showed a bare warning icon with no message on hover, while the Workbench row already used the tooltip-bearing `WarnTriangle` (`warns.join(' · ')`). Both now use `WarnTriangle` uniformly, and I added a `'Tiers have colliding order'` message to a path's computed warnings for the tier-collision case (`hasTierCollision`) that previously only flipped a boolean with no accompanying text — so the tooltip is now always meaningful, not just present.

This is presentation-only: no store/data-model changes, and the retained `data-testid="workbench-row"` (and every class the existing `WorkbenchList.test.ts` asserts on: `selected`, `retired`, `.spill.retired`, `.blank`) are unchanged.

## Test plan

- [x] `npx vitest run` — full UI suite: 303 files / 3798 tests passing, including `WorkbenchList.test.ts` (10/10) unmodified.
- [x] `npm run build` (vite build) — production build succeeds, including the new component's SCSS.
- [x] Manual verification against a real backend: stood up the Postgres/Redis/API stack (`docker compose` + `dotnet run` + the e2e admin-role fixture) and ran the existing Playwright `admin.test.ts` suite (3/3 passing) plus ad-hoc screenshots of both the flat Workbench list (Enemies, with a `BOSS` badge) and the drilled progression tier list (ordinal circles, `new`/warning-triangle badges) to confirm the shared row renders correctly in both shapes.

---
_Generated by [Claude Code](https://claude.ai/code/session_011nLwN2tTeDD7be1EWBPxoM)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/WorkbenchList.svelte
+++ b/UI/src/routes/admin/workbench/components/WorkbenchList.svelte
@@ -22,60 +22,27 @@
 	<div class="list-scroll">
 		{#each filtered as record (record.id)}
 			{@const state = store.stateOf(record)}
-			{@const status = state.status}
-			{@const warns = state.warnings}
-			{@const title = displayName(record)}
-			{@const badge = entity.listBadge?.(record)}
-			{@const edge =
-				status === 'added'
-					? 'var(--change-added)'
-					: status === 'modified'
-						? 'var(--change-modified)'
-						: status === 'deleted'
-							? 'var(--change-removed)'
-							: 'transparent'}
 			{@const retired = (entity.retireable ?? false) && store.isRetired(record)}
-			<button
-				type="button"
-				class="list-row"
-				data-testid="workbench-row"
-				class:selected={record.id === selectedId}
-				class:deleted={status === 'deleted'}
-				class:retired
-				onclick={() => onSelect(record.id)}
+			<ListRow
+				testId="workbench-row"
+				selected={record.id === selectedId}
+				status={state.status}
+				{retired}
+				name={displayName(record)}
+				blank={entity.blankName}
+				badge={entity.listBadge?.(record)}
+				badgeColor={entity.badgeColor?.(record)}
+				warnings={state.warnings}
+				onSelect={() => onSelect(record.id)}
 			>
-				<div
-					class="row-edge"
-					style:width="3px"
-					style:height="34px"
-					style:background={edge}
-					style:box-shadow={edge === 'transparent' ? 'none' : `0 0 7px ${edge}`}
-				></div>
-				<div class="row-body">
-					<div class="row-top">
-						<span class="row-name" class:selected={record.id === selectedId} class:struck={status === 'deleted'}>
-							{#if title}{title}{:else}<span class="blank">{entity.blankName}</span>{/if}
-						</span>
-						{#if badge}
-							<span class="rare-tag" style:color={entity.badgeColor?.(record) ?? 'var(--text-secondary)'}>{badge}</span>
-						{/if}
-						{#if status === 'added'}<span class="spill added">new</span>{/if}
-						{#if status === 'modified'}<span class="spill modified">edited</span>{/if}
-						{#if retired}<span class="spill retired">retired</span>{/if}
-						<div class="spacer"></div>
-						{#if status !== 'deleted' && warns.length > 0}
-							<WarnTriangle title={warns.join(' · ')} />
-						{/if}
-					</div>
-					<span class="meta">
-						{#each entity.meta(record) as [label, value] (label)}
-							<span
-								>{#if label}<b>{value}</b> {label}{:else}<b class="bare">{value}</b>{/if}</span
-							>
-						{/each}
-					</span>
-				</div>
-			</button>
+				{#snippet meta()}
+					{#each entity.meta(record) as [label, value] (label)}
+						<span
+							>{#if label}<b>{value}</b> {label}{:else}<b class="bare">{value}</b>{/if}</span
+						>
+					{/each}
+				{/snippet}
+			</ListRow>
 		{/each}
 	</div>
 </div>
@@ -84,7 +51,7 @@
 import type { EntityConfig, Identified } from '../entities/types';
 import type { EntityStore } from '../entity-store.svelte';
 import WorkbenchIcon from '../WorkbenchIcon.svelte';
-import WarnTriangle from './WarnTriangle.svelte';
+import ListRow from './ListRow.svelte';
 
 interface Props {
 	entity: EntityConfig<Identified>;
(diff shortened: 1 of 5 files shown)


UI/src/routes/admin/workbench/components/WorkbenchList.svelte
```diff
@@ -116,63 +83,9 @@ const liveCount = $derived(store.items.filter((it) => store.stateOf(it).status !
...
-	}
-	.blank {
-		color: var(--text-muted);
-		font-style: italic;
-	}
-}
-.meta .bare {
+// `.meta`'s wrapper is rendered by ListRow, so scope directly off `.bare` rather than an
+// ancestor combinator that Svelte's per-component CSS scoping can't see across.
+.bare {
```
ginderjeremiah: **Should-fix (specificity regression):** Scoping this rule as a bare `.bare` drops its specificity below the global rule that also targets this element.

The `<b class="bare">` for label-less meta values renders inside `ListRow`'s `<span class="meta">`, which the global `workbench.scss` styles as `.workbench .meta b { color: var(--text-secondary); … }` — specificity `(0,2,1)`. This scoped rule compiles to `.bare.svelte-<hash>` — specificity `(0,2,0)`. So the global rule now wins and the bare value renders `--text-secondary` instead of the intended `--text-tertiary`.

Previously this worked because the old selector `.meta .bare` was scoped on both compounds (`.meta.svelte-h .bare.svelte-h`, `(0,4,0)`), beating the global rule. This is a live path, not hypothetical — `item-mod`, `zone`, `challenge`, and `item` all emit `['', value]` tuples that hit the `.bare` branch, and the PR describes itself as presentation-only.

Cleanest fix: move the override next to its siblings in `workbench.scss`'s global `.meta` block (`.meta .bare { color: var(--text-tertiary); }`) rather than scoping it here — that keeps it with the `.meta`/`.spill`/`.rare-tag` rules it belongs with and restores the correct cascade.

---

ginderjeremiah (commented):
## Review: Unify the Workbench and progression list-row primitives (#2125)

**Verdict:** Approve with one should-fix — solid, on-scope refactor; just one small unintended presentation change to resolve before merge. _(Posting as `COMMENT` because GitHub won't let me gate my own PR; the intended verdict is above.)_

This does exactly what #2125 asked: a single `ListRow.svelte` now owns the row shell, the `--change-added/-modified/-removed` edge accent (driven by one `status` prop instead of each caller recomputing an inline style), the name/badge/spill row, and the meta line, while progression keeps its ordinal circle via a `leading` snippet. The change-state token trio stays consistent across both tools — the drift #2060 was guarding against. Markup-first-then-`<script>` matches the house convention, no dead CSS or imports left behind (`WorkbenchIcon` is still used for back/plus/search; `WarnTriangle` moved cleanly into `ListRow`).

### Should-fix
- `WorkbenchList.svelte:88` — the `.meta .bare` → `.bare` change drops CSS specificity below the global `.workbench .meta b` rule, so label-less meta values (`item-mod`/`zone`/`challenge`/`item` all emit `['', value]`) will render `--text-secondary` instead of `--text-tertiary`. Details + suggested fix inline. Small, but it's an unintended visual change on a "presentation-only" PR, and it belongs in this PR rather than a follow-up.

### Notes (non-blocking, intentional unification)
- Progression's meta line moves from its old scoped `.row-meta` (mono 11px, `--text-tertiary`) to the shared global `.meta` (mono 10.5px, `--text-muted`) — a deliberate consequence of unifying, worth a glance to confirm it reads as intended.
- Progression retired rows now inherit `ListRow`'s `opacity: 0.6` dimming and italic-blank styling they didn't have before; this brings them in line with the Workbench rows, which reads as the right call.
- Nice tightening: progression's bare, message-less warn icon is replaced by the tooltip-bearing `WarnTriangle`, and the tier-collision case now carries a real `'Tiers have colliding order'` message instead of only flipping a boolean.

### Test coverage
- `WorkbenchList.test.ts` continues to cover the shared row transitively and its assertions still hold under the new class structure (`.blank` on the row-name span still resolves to the placeholder text; `selected`/`retired`/`.spill.retired` unchanged). Adequate per the issue's "existing tests should continue to cover the refactored markup."
- `ListRow` has no dedicated test and `ProgressionList` has no unit test at all (pre-existing gap) — acceptable here since `ListRow` is largely presentational and transitively exercised, but a small `ListRow` unit test covering the status→edge/spill mapping and the `warnings`→`WarnTriangle` wiring would be worthwhile as a follow-up.

### What's good
Genuinely clean extraction of the deeper/riskier half of #2060 — a single `status` prop replacing three duplicated inline-style computations, a well-chosen `leading` snippet seam for the one real shape difference, and no regression in the retained test surface.

---

PR #2163, opened 2026-07-19
Filter poison entries out of dead-letter replay-all and batch the round trips

### LINKED ISSUE #2129: PlayerUpdateDeadLetters.ReplayAllAsync re-enqueues entries it classifies as poison, and replay cost is quadratic in queue depth

ginderjeremiah:
## Summary

Two related problems in the player-update dead-letter queue Ops surface (`Game.DataAccess/Repositories/Admin/PlayerUpdateDeadLetters.cs`), reported in #2129:

1. **Poison entries were replayed anyway.** `ReplayAllAsync` targeted the entire dead-letter snapshot with no filter, while `Classify` (used by the inspect view) already distinguishes `Malformed`/`UnknownEventType` from `Replayable`. Replaying a poison entry just pushed it onto the live player queue, where the synchronizer immediately re-dead-lettered it — inflating `ReplayedCount`, shuffling the entry to the queue tail (losing its position/age), and re-tripping the dead-letter-depth warning. This was inconsistent with the sibling socket-command DLQ contract (`docs/backend-persistence.md`), which keeps non-replayable entries queued through both peek *and* replay.
2. **Replay cost was quadratic in queue depth.** Each entry paid an individually awaited `RPUSH` + `LREM` pair — fine for a handful of entries, but a mass-dead-letter backlog (thousands of entries) would turn one admin HTTP request into thousands of sequential Redis round trips.

## Changes

- `ReplayAllAsync` now filters the snapshot to entries classified `Replayable` before replaying (reusing the same classification logic `InspectAsync` uses, factored into a shared `ClassifyPayload` helper). `ReplaySelectedAsync` is **unchanged** — replaying a hand-picked entry (including a poison one) is a deliberate, classification-visible operator choice the admin UI already warns about (`nonReplayableSelectedCount` / `poisonNote` in `dead-letters.svelte.ts`), so it shouldn't be silently filtered.
- Added `IPubSubQueue.RemoveRangeAsync`, implemented in `RedisQueue` via a single-round-trip Lua script (mirroring the existing `ReclaimProcessingAsync` pattern) that runs an `LREM count 1` per requested value server-side. Combined with the existing `AddRangeToQueueAsync`, both the push onto the player queue and the removal from the dead-letter queue now happen in one batched round trip each, regardless of how many entries replay — down from 2N sequential round trips.
- Updated `docs/backend-persistence.md`'s dead-letter section to describe the filtered/batched replay-all behavior.

## Note on scope

I picked this up as the oldest unassigned `claude`-labeled issue. Its "worth folding into #2097's fix" note is moot — #2097 was closed by PR #2158, which explicitly left `PlayerUpdateDeadLetters.ReplayAllAsync` untouched and punted the O(N²) concern back to this issue, so #2129 was fully unblocked and addressed standalone here.

## Test plan

- [x] Added `RedisQueueIntegrationTests`: `RemoveRangeAsync` removes each requested occurrence in one call respecting duplicate multiplicity, and skips values no longer present.
- [x] Added `PlayerUpdateDeadLettersIntegrationTests`: `ReplayAllAsync` skips malformed/unknown-type entries and leaves them queued; `ReplaySelectedAsync` still replays an explicitly selected poison entry; `ReplayAllAsync` respects duplicate-payload multiplicity; an all-poison queue replays nothing and stays untouched.
- [x] Updated every hand-written `IPubSubQueue` test fake (`DataProviderSynchronizerTests`, `SocketCommandProcessorTests`) to implement the new interface member.
- [x] Full backend suite: `Game.Core.Tests` (1396), `Game.Infrastructure.Tests` (68), `Game.Application.Tests` (1032), `Game.Api.Tests` (831) — all passing, no regressions.

Closes #2129

---
_Generated by [Claude Code](https://claude.ai/code/session_01SVeieh1Vz8UQEQ8T57aVgr)_


Game.Infrastructure/PubSub/Redis/RedisQueue.cs
```diff
@@ -149,6 +149,33 @@ public async Task<bool> RemoveAsync(string value, CancellationToken cancellation
...
+        // A server-side loop of LREM count 1 calls, one per requested value, in a single round trip — the
+        // batched counterpart to N sequential RemoveAsync calls (the bottleneck a bulk dead-letter replay
+        // would otherwise hit, #2129). Atomic like ReclaimScript, so honour an already-cancelled budget up
+        // front rather than starting a script Redis will run to completion regardless.
+        private static readonly PreparedScript RemoveRangeScript = new(
+            "local n = 0 " +
+            "for i = 1, #ARGV do " +
+            "if redis.call('lrem', KEYS[1], 1, ARGV[i]) == 1 then n = n + 1 end " +
+            "end " +
+            "return n");
```
ginderjeremiah: **Should-fix (design consideration):** This eliminates the N round trips, but it does so by moving the entire per-value `LREM` loop into one atomic Lua script — and as your own comment notes, Redis "will run [it] to completion regardless." Each `LREM` is an O(M) scan of the list, so the script is O(N·M) of single-threaded, non-yielding work that blocks *every* other client on the instance for its full duration.

The sibling `ReclaimScript` uses the same pattern, but its comment explicitly leans on the list being **bounded** ("on a bounded in-flight processing list the script stays short"). Replay-all has no such bound — the scenario #2129 cites (a DB outage dead-lettering every in-flight write, thousands of entries) is precisely when N and M are both large, so a several-second game-wide Redis stall is most likely exactly when you invoke this. That's arguably a worse failure mode than the sequential round trips it replaces, which at least let other clients interleave.

Worth weighing one of:
- **Pipelined `LREM`s** (fire all `ListRemoveAsync` without awaiting each, then `Task.WhenAll`) — StackExchange.Redis multiplexes these into effectively one round trip, but Redis still processes them as separate commands and can interleave other clients. Gets the round-trip win without monopolizing the server.
- **Chunk the script** into batches (e.g. 500/call) to bound the worst-case stall.
- A single O(M) list rebuild if ordering can change.

Not a correctness blocker and the pattern has precedent, so shipping as-is with a follow-up issue is defensible — but the unbounded-N stall is a real footgun in the target scenario and shouldn't be silent.

---

ginderjeremiah (commented):
## Review: filter poison out of player-update DLQ replay-all + batch the Redis round trips (#2129)

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't accept a gating verdict from me)* — with one should-fix worth weighing before merge.

The change is correct, well-scoped, and cleanly solves both halves of #2129. Verified against the code and the sibling surface:

- **Poison filter is right and consistent.** Replay-all now targets only `EDeadLetterReason.Replayable`, reusing the same `ClassifyPayload` core the inspect view uses. This matches `SocketCommandDeadLetters.ReplayAsync` (`Game.Api/Services/Admin/SocketCommandDeadLetters.cs:63-74`), which likewise leaves non-replayable entries queued — so the "sibling DLQ contract" claim in the description/docs checks out.
- **Selected-replay divergence is intentional.** `ReplaySelectedAsync` staying unfiltered (operator can replay a hand-picked poison entry, admin UI warns) is a deliberate override and is left untouched — good call to keep it out of scope.
- **`ClassifyPayload` refactor** cleanly factors the classification core out of `Classify` with no behavior change; the tuple split reads well.
- Only `RedisQueue` implements `IPubSubQueue` in production; both hand-written test fakes (`DataProviderSynchronizerTests`, `SocketCommandProcessorTests`) got the new member. Nothing left unimplemented.

### Should-fix

- `Game.Infrastructure/PubSub/Redis/RedisQueue.cs:152-161` — the batched removal is one atomic, non-yielding Lua script running an O(N·M) `LREM` loop. It blocks the whole Redis instance for its full duration, and unlike the `ReclaimScript` it mirrors, replay-all's N is unbounded — largest in exactly the mass-outage recovery scenario #2129 targets. See the inline comment for options (pipelined `LREM`s, chunking, or an O(M) rebuild). Not a correctness blocker; shipping with a follow-up issue is defensible, but it shouldn't go in silently.

### Test coverage

Solid and matches the project's integration-test rules for out-of-process (Redis) dependencies:
- `RedisQueueIntegrationTests`: `RemoveRangeAsync` respects duplicate multiplicity in one call, and skips values no longer present (counting only actual removals).
- `PlayerUpdateDeadLettersIntegrationTests`: replay-all skips malformed/unknown-type and leaves them queued; selected-replay still replays an explicit poison entry; duplicate-payload multiplicity; all-poison queue replays nothing and stays untouched.
- Covers the behavior change and the edge cases (empty-after-filter, duplicates, poison). No gaps I'd block on.

### What's good

Careful at-least-once reasoning preserved through the batching (push-all before remove-all keeps the crash-duplicates-never-loses contract), the doc update in `backend-persistence.md` is accurate, and the commentary explains the *why* without over-narrating. Nice work.

### Note on merge

Not merging: there's an open should-fix, and CI hasn't reported yet (`get_status` → `pending`, 0 checks). Once you've decided on the Redis-stall point (fix here or spin a follow-up) and CI is green, this is good to go.

---

ginderjeremiah (commented):
Good catch — fixed in e29d43b rather than deferring: replaced `RemoveRangeScript` (the atomic Lua loop) with the pipelined-`LREM`s approach you suggested first. `RemoveRangeAsync` now fires each `ListRemoveAsync` without awaiting between them so StackExchange.Redis pipelines them onto the wire in one flight, but Redis processes them as separate commands that can interleave with other clients rather than one non-yielding script monopolizing the instance for O(N·M). Same-connection ordering is preserved (so duplicate-value LREM correctness still holds — covered by the existing `RemoveRangeAsync_RemovesEachRequestedOccurrenceInOneCall...` test), and cross-value atomicity was never actually required here. Full suite re-run green (3327 tests).

---

ginderjeremiah:
## Review (independent re-review) — Approve ✅

Re-reviewed the current head (`e29d43b`) against #2129 and verified the code independently rather than relying on the earlier self-review. This cleanly solves both halves of the issue.

**Correctness**
- **Poison filter** — replay-all now targets only `EDeadLetterReason.Replayable`, reusing the `ClassifyPayload` core cleanly factored out of `Classify` (no behavior change). `ReplaySelectedAsync` intentionally stays unfiltered, matching the sibling socket-command DLQ contract and the admin UI's poison warning. Correct and consistent.
- **At-least-once preserved** — push-all (`AddRangeToQueueAsync`) before remove-all (`RemoveRangeAsync`) keeps the crash-can-only-duplicate-never-lose guarantee the per-item loop had.
- **`RemoveRangeAsync`** — the earlier should-fix (unbounded atomic Lua `LREM` loop blocking the whole instance) was resolved here by switching to pipelined `LREM count 1` calls in a single wire flight. `ObserveWrite(Task.WhenAll(...))` observes the batch (aggregate faults aren't lost), and `Count(c => c > 0)` equals the sum since each `LREM` count is 1. Duplicate multiplicity is handled by one call per requested value.
- Only `RedisQueue` implements `IPubSubQueue` in production; both hand-written test fakes were updated. Nothing left unimplemented.

**Tests** — integration coverage matches the project's out-of-process-dependency rules and hits the behavior change plus the edges: duplicate multiplicity, skip-missing, all-poison no-op, and selected-poison still replaying. Docs update in `backend-persistence.md` is accurate.

**CI** — green: `test-backend`, `verify-backend`, all three `test-e2e` browsers, and the aggregate `all-checks-passed` all pass (frontend jobs skipped — no frontend changes). The `get_status` "pending" is just the absence of legacy commit-statuses; the check-runs are the real signal.

No remaining nits or should-fixes, and no follow-up work required. Merging.

---

PR #2164, opened 2026-07-19
Memoize idle-mode enemy combat rating in the offline loop

### LINKED ISSUE #2130: Offline idle loop re-derives each random enemy's combat rating from scratch on every victory
`OfflineProgressSimulator` memoizes the player and boss ratings (#1730), but Idle mode calls `CombatRating.Rate(enemy.ToBattler(), …)` per idle victory (`Game.Core/Battle/Offline/OfflineProgressSimulator.cs:125`). Idle enemies come from a small per-zone spawn table at a bounded level range, and the rating is a pure function of (template, level) — so a memo keyed on `(enemy.Id, enemy.Level)` (same shape as the existing `bossEnemyRating` memo) would remove up to thousands of `CombatRating.Rate` calls per 10-hour window, each of which materializes three full `AttributeCollection`s and enumerates the whole `EAttribute` enum.

Additionally, `Rate` never mutates its battler, so the rating battler and the simulation battler (:284 builds a second `enemy.ToBattler()`) need not be built separately.

Bounded cost overall (the loop is already bounded by the 10-hour cap), so low urgency — but it's the only remaining per-iteration redundancy in that loop.

ginderjeremiah:
## Summary

`OfflineProgressSimulator` calls `CombatRating.Rate(enemy.ToBattler(), …)` for every idle-mode victory, even though idle encounters are drawn from a small, bounded-level per-zone spawn table — so a 10-hour away window can recompute the same enemy's rating from scratch thousands of times, each materializing three full `AttributeCollection`s and walking the whole `EAttribute` enum.

`OfflineProgressSimulator.RateIdleEnemy` now memoizes the rating by `(enemy id, level)`, mirroring the existing `bossEnemyRating` memo Boss mode already had.

## Note on the issue's premise

The issue described the rating as "a pure function of (template, level)." That's not quite right in general: `Enemy.SelectBattleSkills` (used for every idle encounter) draws a genuinely random subset of up to `GameConstants.MaxSelectedSkills` (4) skills whenever an enemy's pool exceeds that cap — confirmed by the existing `EnemyTests`/`BattleFactoryTests` coverage, which explicitly exercise 6–8 skill pools. For such an enemy, two encounters can field different loadouts and therefore different ratings, so a naive `(id, level)`-only memo would silently misprice every encounter after the first with a stale value from an earlier random draw.

Today's authored content never hits this (`content/enemies.json`'s largest `skillPool` is 2), but nothing enforces that invariant, so I didn't want to bake in a fix that's only correct by coincidence of current content. `RateIdleEnemy` only memoizes when `Enemy.AvailableSkills.Count <= GameConstants.MaxSelectedSkills` — the case where `SelectBattleSkills` always fields the whole (stationary) pool, just shuffled, which doesn't affect `CombatRating.Rate`'s skill-order-independent sum. An oversized pool falls back to a fresh per-encounter computation, exactly matching pre-fix behavior.

## Scope note: the issue's second suggestion

The issue also suggested that since `CombatRating.Rate` doesn't mutate its battler, the rating battler and the simulation battler "need not be built separately." I looked into this and it's **not safe as stated**: `CombatRating.Rate` itself doesn't mutate, but `BattleSimulator.Simulate()` does — it applies/expires timed effects via `Battler.AddModifier`/`RemoveModifier`, which changes what `GetAttributeValue` (and therefore `Rate`) would report afterward. Reusing the *already-simulated* battler for a post-battle rating call would read post-fight transient state, not the enemy's assembled combat power. Merging the two builds safely would require computing the rating unconditionally *before* simulating (not just on a win), which is a real behavior change I didn't want to bundle into this fix — especially since the memo already eliminates the redundant computation this issue was primarily about. I left that as-is; happy to open a follow-up if it's worth pursuing separately, but I don't think it's worth much once the memo is in place.

## Test plan

- [x] Added `Simulate_IdleMode_MemoizedEnemyRatingMatchesAFreshComputation_AcrossRepeatedEncounters`: pins that a memoized rating still equals a fresh `CombatRating.Rate` computation for a stationary (small) skill pool.
- [x] Added `Simulate_IdleMode_OversizedSkillPoolEnemy_PricesEachVictoryByItsOwnActuallySelectedLoadout`: pins that an enemy with a pool larger than `MaxSelectedSkills` still gets a fresh, per-encounter-correct rating (verified this test actually catches the bug by temporarily removing the safety guard locally — it failed as expected, then passed again once restored).
- [x] Full backend suite: `Game.Core.Tests` (1398), `Game.Infrastructure.Tests` (68), `Game.Application.Tests` (1026), `Game.Api.Tests` (831) — all passing, no regressions.

Closes #2130

---
_Generated by [Claude Code](https://claude.ai/code/session_01CrnZ7d1sAkVzGcrdSkYQh2)_

### DIFF
--- a/Game.Core.Tests/Battle/Offline/OfflineProgressSimulatorTests.cs
+++ b/Game.Core.Tests/Battle/Offline/OfflineProgressSimulatorTests.cs
@@ -495,6 +495,57 @@ namespace Game.Core.Tests.Battle.Offline
             Assert.All(result.Battles, battle => Assert.Equal(0, battle.Result.Stats.PlayerRating));
         }
 
+        // ── Idle enemy rating memo (#2130) ───────────────────────────────────
+
+        [Fact]
+        public void Simulate_IdleMode_MemoizedEnemyRatingMatchesAFreshComputation_AcrossRepeatedEncounters()
+        {
+            // WeakEnemy's one-skill pool is well within GameConstants.MaxSelectedSkills, so
+            // Enemy.SelectBattleSkills always fields its whole (stationary) pool and RateIdleEnemy's memo
+            // applies. Pin that every victory against the same (id, level) still reports exactly the rating a
+            // fresh CombatRating.Rate computation would, whether served from the memo or not.
+            var scenario = StrongPlayerWinScenario();
+            var freshEnemy = WeakEnemy(level: 1);
+            var expectedRating = CombatRating.Rate(
+                new Battler(new AttributeCollection(freshEnemy.GetAttributeModifiers()), freshEnemy.AvailableSkills, freshEnemy.Level),
+                isPlayer: false);
+
+            var result = _simulator.Simulate(IdleParameters(ManyStepsBudget(), scenario));
+
+            var wins = result.Battles.Where(b => b.Result.Victory).ToList();
+            Assert.True(wins.Count > 1);
+            Assert.All(wins, battle => Assert.Equal(expectedRating, battle.EnemyRating, precision: 9));
+        }
+
+        [Fact]
+        public void Simulate_IdleMode_OversizedSkillPoolEnemy_PricesEachVictoryByItsOwnActuallySelectedLoadout()
+        {
+            // A pool larger than GameConstants.MaxSelectedSkills makes Enemy.SelectBattleSkills draw a
+            // genuinely random subset per encounter, so the enemy's rating is not pure in (id, level) alone —
+            // RateIdleEnemy must not memo-serve a stale rating from an earlier encounter's different draw.
+            // VariedSkillPoolEnemy's 8 skills carry distinct damage, so a different draw genuinely re-prices
+            // the enemy; pin that every victory's EnemyRating matches a fresh rating of that battle's own
+            // recorded loadout rather than the first one seen.
+            var scenario = new Scenario
+            {
+                Zone = MakeZone(levelMin: 1, levelMax: 1),
+                Snapshot = PlayerSnapshot(strength: 100, endurance: 100),
+                ResolveEnemy = VariedSkillPoolEnemy,
+            };
+
+            var result = _simulator.Simulate(IdleParameters(ManyStepsBudget(), scenario));
+
+            var wins = result.Battles.Where(b => b.Result.Victory).ToList();
+            Assert.True(wins.Count > 1);
+            var distinctRatings = wins.Select(b => b.EnemyRating).Distinct().Count();
+            Assert.True(distinctRatings > 1, "The random per-encounter loadout never varied the enemy rating — the fixture doesn't exercise the case.");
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: memoize the idle-mode enemy combat rating in the offline loop (closes #2130)

**Verdict:** Approve _(submitted as a comment — I'm the PR author, so GitHub won't take a gating event from me)_ — correct, well-tested, and the design mirrors the existing boss-rating memo.

I verified the correctness argument end-to-end rather than taking the description at face value, and it holds:

- `CombatRating.Rate` sums each fielded skill's contribution in `OffenseRate` (commutative) and folds Self/Opponent effects into an order-free `AttributeCollection` in `BuildEffectiveCaster`/`BuildReferenceDefense`, so for a given skill **set** the rating is permutation-invariant.
- `SelectBattleSkills` uses `take = min(MaxSelectedSkills, pool.Length)`, so when `AvailableSkills.Count <= MaxSelectedSkills (4)` it always fields the *entire* pool (just shuffled). Same set every encounter ⇒ genuinely stationary rating ⇒ the `(id, level)` memo is exactly correct. The `> MaxSelectedSkills` fallback guard precisely matches the only case where a random subset is drawn.
- `enemy.ToBattler()` reads the assembled (unmutated) `Enemy`, not the post-simulation `Battler`, so calling it after the fight for the rating is safe — consistent with correctly declining the issue's second "reuse the simulation battler" suggestion (`BattleSimulator.Simulate` does mutate its battler).

The PR description's pushback on the issue's premise (the rating is *not* pure in `(id, level)` for oversized pools) is the right call, and guarding on it rather than relying on today's content (`enemies.json` max `skillPool` is 2) keeps the fix correct-by-construction rather than correct-by-coincidence.

### Test coverage
- `...MemoizedEnemyRatingMatchesAFreshComputation...` pins the memo-hit path against a fresh `CombatRating.Rate` computation.
- `...OversizedSkillPoolEnemy_PricesEachVictoryByItsOwnActuallySelectedLoadout` pins the fallback path, and — nicely — self-guards with `distinctRatings > 1` so it fails loudly if the fixture ever stops exercising the varying-loadout case. Both branches of `RateIdleEnemy` are covered.
- Rating is server-only (`CombatRating` is never simulated client-side), so there's no frontend battle-parity mirror to add here — correct.

### What's good
- Faithful to the established `bossEnemyRating` pattern; the memo is scoped to a single `Simulate` call and bounded by the spawn table × level range.
- The XML doc on `RateIdleEnemy` earns its length — it documents a real, non-obvious invariant (why the guard exists) that a future edit could otherwise naively "simplify" away.

No blocking items, should-fixes, or nits. Will merge once `test-backend` finishes (the only check still running; all others are green, frontend jobs correctly skipped).

---

PR #2165, opened 2026-07-19
Centralize the 401 error envelope in the JWT bearer challenge; drop 7 dead per-action guards

### LINKED ISSUE #2131: Seven dead per-action Authenticated guards; genuinely unauthenticated callers get a bare 401 instead of the ApiResponse envelope
The global fallback policy (`RequireAuthenticatedUser`, `Game.Api/Startup.cs:354-359`) rejects any unauthenticated request to a non-`[AllowAnonymous]` endpoint at the authorization middleware, before the action runs. Past that point `SessionService.Authenticated` can only be false if the validated token lacks an integer `sub` claim — which `JwtTokenService.CreateAccessToken` (`Game.Api/Auth/JwtTokenService.cs:48`) always sets. So the copied guard blocks in `AuthController.cs:111-114, 139-142` and `PlayersController.cs:37-40, 60-63, 84-87, 107-110, 126-129` are dead in practice (DRY/dead-code guideline).

The behavioral wrinkle those guards were papering over is real, though: a genuinely unauthenticated caller receives the JWT bearer handler's **empty-body 401**, not the documented consistent `{ errorMessage }` `ApiResponse` envelope.

Fix: centralize envelope emission for auth rejections (e.g. the bearer `OnChallenge` event) and drop the seven per-action guards — or, if defense-in-depth is preferred, replace them with one shared filter rather than per-action copies.

ginderjeremiah:
## Summary

Closes #2131.

- The global fallback authorization policy (`RequireAuthenticatedUser`, `Startup.cs`) already rejects any unauthenticated request to a non-`[AllowAnonymous]` endpoint at the authorization middleware, before the action runs. Past that point, `SessionService.Authenticated` can only be false if the validated token lacks an integer `sub` claim — which `JwtTokenService.CreateAccessToken` always sets. So the seven per-action `if (!_sessionService.Authenticated)` guards in `AuthController` (`Status`, `ActiveSession`) and `PlayersController` (`SelectPlayer`, `SwitchPlayer`, `CreatePlayer`, `Players`, `CharacterCreationData`) were dead code — removed.
- The real behavioral gap those guards were papering over: a genuinely unauthenticated caller received the JWT bearer handler's default **empty-body 401** instead of the project's consistent `{ errorMessage }` `ApiResponse` envelope every other failure response uses. Fixed by writing the envelope from the bearer's `OnChallenge` event in `Startup.cs`, mirroring the existing pattern in the auth rate limiter's `OnRejected` handler (which does the same for 429s).
- This doesn't touch the `/socket` endpoint's 401 handling, which already writes the envelope via a separate `SocketInterceptorMiddleware` that runs before the authorization middleware, nor 403 (`AdminRoleAuthorizationFilter`), which was already unaffected by this gap.

## Test plan

- [x] `dotnet build Game.sln` — succeeds with 0 warnings/errors.
- [x] Updated the six existing `*_Unauthenticated_Returns401` / `*_WithoutAuthentication_IsRejected` integration tests (`AuthControllerTests`, `PlayersControllerTests`) to also assert the response body now carries a populated `ErrorMessage`, proving the envelope is written.
- [x] Full backend suite: `dotnet test Game.sln --no-build --no-restore` — 3329 tests, 0 failures (`Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Application.Tests`, `Game.Api.Tests`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Da4f1uCbutT6Y4sqwmqMgd)_

### DIFF
--- a/Game.Api.Tests/Integration/AuthControllerTests.cs
+++ b/Game.Api.Tests/Integration/AuthControllerTests.cs
@@ -268,6 +268,10 @@ namespace Game.Api.Tests.Integration
             var response = await Client.GetAsync("/api/Auth/Status", CancellationToken);
 
             Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            // The bearer challenge writes the project's standard ApiResponse envelope rather than the
+            // JWT bearer handler's default empty body.
+            var result = await response.Content.ReadFromJsonAsync<ApiResponse>(CancellationToken);
+            Assert.NotNull(result?.ErrorMessage);
         }
 
         [Fact]
@@ -299,6 +303,8 @@ namespace Game.Api.Tests.Integration
             var response = await Client.GetAsync("/api/Auth/ActiveSession", CancellationToken);
 
             Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            var result = await response.Content.ReadFromJsonAsync<ApiResponse>(CancellationToken);
+            Assert.NotNull(result?.ErrorMessage);
         }
 
         [Fact]
--- a/Game.Api.Tests/Integration/PlayersControllerTests.cs
+++ b/Game.Api.Tests/Integration/PlayersControllerTests.cs
@@ -197,6 +197,10 @@ namespace Game.Api.Tests.Integration
                 new { PlayerId = 1, RefreshToken = "irrelevant" }, CancellationToken);
 
             Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            // The bearer challenge writes the project's standard ApiResponse envelope rather than the
+            // JWT bearer handler's default empty body.
+            var result = await response.Content.ReadFromJsonAsync<ApiResponse>(CancellationToken);
+            Assert.NotNull(result?.ErrorMessage);
         }
 
         [Fact]
@@ -206,6 +210,8 @@ namespace Game.Api.Tests.Integration
                 new { PlayerId = 1, RefreshToken = "irrelevant" }, CancellationToken);
 
             Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+            var result = await response.Content.ReadFromJsonAsync<ApiResponse>(CancellationToken);
+            Assert.NotNull(result?.ErrorMessage);
         }
 
         [Fact]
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: Centralize 401 envelope in the JWT bearer challenge; drop 7 dead per-action `Authenticated` guards (closes #2131)

**Verdict:** Approve — the reasoning is sound, verified against the actual code, and the change mirrors an existing established pattern. _(Posted as a comment because GitHub won't accept a self-approval; treat this as an approval.)_

I verified the two load-bearing claims rather than taking them on faith:

- **The seven guards are genuinely dead.** The fallback policy (`Startup.cs`, `RequireAuthenticatedUser`) rejects any non-`[AllowAnonymous]` request at the authorization middleware. I confirmed the affected actions (`AuthController.Status`/`ActiveSession`, all five `PlayersController` actions) are **not** `[AllowAnonymous]`, so the middleware runs first. Past it, `SessionService.Authenticated => UserId > 0`, and `SessionLoaderMiddleware` sets `UserId` only when the principal is authenticated and its `sub` parses as an int — which `JwtTokenService.CreateAccessToken` always emits. So `Authenticated` can't be false at those actions. Removal is also consistent with `DeviceInfo`, which already used `UserId` directly with no guard.
- **The envelope fix is correct and consistent.** `OnChallenge` fires on the 401 challenge only (not 403/forbidden), calls `HandleResponse()` to suppress the default empty body, and writes `ApiResponse.Error(...)` via `WriteAsJsonAsync` — the exact pattern the rate limiter's `OnRejected` already uses for 429s. There's no custom JSON naming policy in `Game.Api`, so both paths serialize with the same System.Text.Json defaults as MVC; the envelope shape stays consistent with every other error response. Centralizing on the challenge event (rather than a shared filter) is the right call given the fallback policy short-circuits before any action or filter runs.

### Test coverage

Adequate for a tech-debt change with no new domain logic. The six existing `*_Unauthenticated_Returns401` / `*_WithoutAuthentication_IsRejected` integration tests now also assert a populated `ErrorMessage`, which is exactly what proves the envelope is written on the real challenge path. The missing-token case they exercise hits the same `OnChallenge` code regardless of the underlying reason (missing vs. expired vs. invalid), so there's nothing meaningful left to add without testing framework internals.

### What's good

- The PR does the diligence the issue asked for — it doesn't just trust the "dead code" claim, it explains *why* each guard is unreachable, and I found that reasoning holds under inspection.
- Correctly scoped: it deliberately leaves the `/socket` 401 path (handled by `SocketInterceptorMiddleware` before authz) and 403 (`AdminRoleAuthorizationFilter`) untouched, and the comments capture that intent concisely.

Nothing blocking and no nits. Merging once CI is green.

---

PR #2166, opened 2026-07-19
Distinguish pre-selection Status/ActiveSession state from a real 404 (#2132)

### LINKED ISSUE #2132: A valid pre-selection token hitting Status/ActiveSession logs a warning per request, and Status returns an opaque 404
`SessionInitializer` (`Game.Api/Services/SessionInitializer.cs:31-37`) emits `LogWarning("…has a valid token with no selected player…")` whenever a valid **pre-selection** token (post-`Login`, pre-`SelectPlayer` — a documented, normal flow state per `docs/backend-auth.md`) hits `AuthController.Status` or `ActiveSession` (`AuthController.cs:118, 146`). Every page refresh on the character-select screen produces a server warning for a state the design explicitly allows, and `Status` then degrades to `LoadPlayer(SelectedPlayerId == 0)` → a 404 "Player data not found" instead of anything distinguishing "no character selected yet".

Fix: downgrade the log to Debug/Information for the pre-selection case, and give `Status` a distinct graceful "no character selected" response rather than the misleading 404.

ginderjeremiah:
## Summary

- `SessionInitializer.EnsureSessionLoaded` logged a `LogWarning` on every request that hit `Status`/`ActiveSession` with a valid **pre-selection** token (post-`Login`, pre-`SelectPlayer`) — a normal, documented flow state (`docs/backend-auth.md`), not a warning-worthy condition. Every character-select-screen refresh produced a server warning. Downgraded to `LogDebug`.
- `AuthController.Status` collapsed that same pre-selection state into the generic `"Player data not found"` 404 that a genuinely unloadable player (deleted/archived between requests) also produces, making the two indistinguishable to any caller. Added `ApiErrorCategory.NoPlayerSelected` (mapped to `409 Conflict` in `ErrorStatusFilter`) so `Status` now returns a distinct, graceful response for "no character selected yet" instead of an opaque 404.
- `ActiveSession` already degrades gracefully for this state (`HasActiveSocket(0)` just reports `Active: false`), so it needed only the shared log-level fix.

## Why not also change frontend routing

I looked at whether the frontend's `resumeSession` should treat this new status differently (e.g. route a refresh on `/select` to stay put instead of bouncing to `/`). It shouldn't: `/select/+page.svelte`'s own `onMount` already documents that a refresh/direct deep-link with no in-memory character-summary handoff *must* return to login, since there's no endpoint to re-fetch the account's character list outside of `Login` itself. So today's "any Status failure → login" frontend behavior is already correct for this case; no frontend change was needed.

## Test plan

- Replaced the existing (mislabeled) `Status_AuthenticatedButPlayerNotLoadable_Returns404WithError` test — it was actually exercising the pre-selection-token path via `TestAuthHelper.AddAuthHeader(client, user.Id)` (no player claim), not a genuinely deleted player — with `Status_PreSelectionToken_ReturnsNoPlayerSelectedNotOpaque404`, asserting `409 Conflict`.
- Added `Status_PostSelectionTokenButPlayerGone_Returns404WithError`, which now actually covers the "post-selection token naming a player that can't load" case (a token minted for a never-persisted player id), asserting `404 NotFound` is preserved for that genuine failure.
- Added `ErrorStatusFilterTests.RewritesTo409_WhenErrorCategoryIsNoPlayerSelected`.
- `dotnet build`: succeeds, 0 warnings/errors.
- `dotnet test Game.Api.Tests`: 833/833 passing.

Closes #2132

---
_Generated by [Claude Code](https://claude.ai/code/session_01ABjMUFj2GJpom6VqswjiwS)_

### DIFF
--- a/Game.Api.Tests/Integration/AuthControllerTests.cs
+++ b/Game.Api.Tests/Integration/AuthControllerTests.cs
@@ -275,18 +275,39 @@ namespace Game.Api.Tests.Integration
         }
 
         [Fact]
-        public async Task Status_AuthenticatedButPlayerNotLoadable_Returns404WithError()
+        public async Task Status_PreSelectionToken_ReturnsNoPlayerSelectedNotOpaque404()
         {
-            // A still-valid token whose player can't be loaded (archived/deleted between requests; here a user
-            // with no player at all) must return a structured error, not a 500 — mirroring how Login surfaces a
-            // missing player. Rehydration finds no player, so the session's selected id stays unresolved and the
-            // player load comes back null. The missing-resource semantics surface as 404, not a blanket 400.
+            // A pre-selection token (post-Login, pre-SelectPlayer) is a normal, documented flow state
+            // (docs/backend-auth.md) — every character-select screen refresh carries one. Status must
+            // surface it as its own distinguishable category, not the same 404 a genuinely unloadable
+            // player gets (see the test below).
             using var scope = CreateScope();
             var context = scope.ServiceProvider.GetRequiredService<GameContext>();
             var user = await TestDataSeeder.CreateUserAsync(context, "playerlessstatus", "pass");
 
             var client = Factory.CreateClient();
             TestAuthHelper.AddAuthHeader(client, user.Id);
+
+            var response = await client.GetAsync("/api/Auth/Status", CancellationToken);
+
+            Assert.Equal(HttpStatusCode.Conflict, response.StatusCode);
+            var result = await response.Content.ReadFromJsonAsync<ApiResponse<Models.Player.PlayerData>>(CancellationToken);
+            Assert.Null(result?.Data);
+            Assert.NotNull(result?.ErrorMessage);
+            client.Dispose();
+        }
+
+        [Fact]
+        public async Task Status_PostSelectionTokenButPlayerGone_Returns404WithError()
+        {
+            // A post-selection token naming a player that can't be loaded (archived/deleted between
+            // requests) is a genuine missing-resource failure, distinct from the pre-selection state
+            // above — it must still surface as 404, not a 500. A token minted for a player id that was
+            // never persisted stands in for "since deleted" without tearing down FK-linked rows.
+            var (userId, _) = await SeedAsync("playergonestatus", "pass");
+
+            var client = Factory.CreateClient();
+            TestAuthHelper.AddAuthHeader(client, userId, playerId: 999_999_999);
 
             var response = await client.GetAsync("/api/Auth/Status", CancellationToken);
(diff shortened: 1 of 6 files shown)


ginderjeremiah (commented):
## Review: Distinguish pre-selection Status/ActiveSession state from a real 404 (#2132)

**Verdict:** Approve — the change cleanly separates the normal pre-selection flow state from a genuine missing-player failure and downgrades the spurious per-request warning, exactly as the issue asks. (Posted as a comment since I'm the PR author; GitHub won't accept a formal approval on one's own PR.)

### Correctness
- The gate order in `AuthController.Status` is right: `EnsureSessionLoaded` first, then the `TokenSelectedPlayerId is null` short-circuit (pre-selection → `NoPlayerSelected`), then the `LoadPlayer` null check (post-selection but unloadable → `NotFound`). `TokenSelectedPlayerId` is the token-claim authority (`SessionService.cs:24`), independent of the volatile session cache, so this correctly identifies the pre-selection case rather than inferring it from a `SelectedPlayerId == 0` degrade.
- `ActiveSession` needs only the shared log-level fix — for a pre-selection token `SelectedPlayerId` is 0 and `HasActiveSocket(0)` truthfully reports `Active: false`, so no special-casing is warranted.
- Frontend is genuinely unaffected: `loadPlayer` in `UI/src/lib/engine/session.ts` treats every non-200 identically (`response.status !== 200 → login`), so 404→409 is behavior-neutral, and the PR's rationale for not touching `/select` routing matches that code.

### Test coverage
- Good catch renaming `Status_AuthenticatedButPlayerNotLoadable_Returns404WithError` → `Status_PreSelectionToken_ReturnsNoPlayerSelectedNotOpaque404`: the old name misdescribed what it exercised (a no-player-claim token, i.e. the pre-selection path), and it now asserts the distinct 409.
- `Status_PostSelectionTokenButPlayerGone_Returns404WithError` genuinely covers the post-selection-but-unloadable path via a token minted for a never-persisted player id — a clean way to stand in for "deleted since" without tearing down FK-linked rows.
- `ErrorStatusFilterTests.RewritesTo409_WhenErrorCategoryIsNoPlayerSelected` covers the new mapping. Both endpoint states and the filter mapping are exercised.

### Notes (non-blocking)
- 409 Conflict for "no character selected" is a defensible, well-documented choice (the enum XML comment explains the intent). It reads slightly unconventionally versus a 200-with-flag or 428, but since the category is a reusable, self-describing seam and the frontend is agnostic to the exact code, it's fine as-is.

Clean, focused change with matching tests and docs-consistent reasoning. Approving pending green CI.

---

PR #2180, opened 2026-07-19
Reject Multiplicative skill effects on DoT-accumulator attributes

### LINKED ISSUE #2169: Multiplicative skill effects targeting a DoT accumulator double-apply the caster's amplification — and nothing blocks authoring one

`BattleContext.ApplySkillEffect` freezes the caster's typed amplification into the resolved magnitude of *any* effect whose `AttributeId` is a DoT accumulator, without checking `ModifierType` (`Game.Core/Battle/BattleContext.cs:77-81`):

```csharp
var amount = effect.Amount + _activeBattler.GetAttributeValue(effect.ScalingAttributeId) * effect.ScalingAmount;
if (DamageTypes.DotTypeForAccumulator(effect.AttributeId) is EDamageType dotType)
{
    amount = _activeBattler.AmplifyDamage(amount, dotType);
}
```

For an **additive** effect this is the documented amp-freeze rule (spike #1320 Area C). For a **multiplicative** effect, `amount` is a *ratio* that then compounds into the stack's combined multiplicative modifier (`Battler.cs` fold: `(stack.Multiplicative?.Amount ?? 1.0) * amount`), so the caster's amp is applied twice.

**Concrete failure:** caster has `PoisonAmplification = 0.25`. Apply +30 Poison/s → frozen to 37.5 (correct). Then apply a "×2 the target's PoisonDamagePerSecond" effect: the factor becomes `2 × 1.25 = 2.5`, so the accumulator lands at `93.75` instead of the intended `75`.

The frontend mirrors this exactly (`UI/src/lib/battle/skill.ts:93-96`), so there is no parity break — it's a deterministic semantic defect on both sides. Two adjacent inconsistencies compound it:

- `CombatRating.OffenseRate`'s DoT closed form (`CombatRating.cs:183-197`) reads `effect.Amount` as an additive DPS magnitude regardless of `ModifierType` — a "×2" factor would be priced as 2 flat damage/second.
- The admin save path has no shape validation: `AdminSkills.SetEffects` (`Game.DataAccess/Repositories/Admin/AdminSkills.cs:141-190`) persists any `(AttributeId, ModifierType)` combination, so the Workbench can author this today. No seeded content currently does.

**Suggested fix:** the design doc defines a DoT effect as pure data — a debuff that adds to `XDamagePerSecond` — so the cleanest resolution is an authoring-validation rule restricting DoT-accumulator effects to `Additive` (a guard in `AdminSkills.SetEffects` plus a content-lint rule in `ProgressionGraphChecker`), rather than teaching the engine and the rating model multiplicative-DoT semantics. No tests currently pin multiplicative-on-accumulator behavior on either side — add pins with the fix.

ginderjeremiah:
## Summary

`BattleContext.ApplySkillEffect` freezes the caster's typed amplification into the resolved magnitude of any effect targeting a DoT-accumulator attribute (`Game.Core/Battle/BattleContext.cs:77-81`), regardless of `ModifierType`. For an **Additive** effect this is the documented amp-freeze rule (spike #1320 Area C) — correct. For a **Multiplicative** effect the frozen amount is a *ratio* that then compounds into the stack's combined multiplicative modifier, so the caster's amplification is applied twice. Nothing blocked authoring one, though no seeded content currently does.

Per the issue's suggested direction, this fixes it with **authoring-validation** rather than teaching the engine/rating model multiplicative-DoT semantics:

- **`AdminSkills.SetEffects`** (`Game.DataAccess/Repositories/Admin/AdminSkills.cs`): rejects a save containing a Multiplicative effect on a DoT-accumulator attribute (`DamageTypes.DotTypeForAccumulator`), matching the existing anti-tamper guard pattern used for skill damage portions.
- **`ProgressionGraphChecker`** (`Game.Application/Content/ProgressionGraphChecker.cs`): adds a matching `SkillEffectDotModifier` Error check as a defense-in-depth backstop (the admin save path already rejects the combination before it can be persisted, so Content Health can't observe a live instance today — same pattern as the same-path prerequisite guard in #2128/#2144).

I confirmed no seeded content (`content/skills.json`) currently authors this combination, so the new lint check introduces no findings against the live content graph.

## Scope note

The issue's body also flags two adjacent inconsistencies as context (not requested as part of this fix):
- `CombatRating.OffenseRate`'s DoT closed form reads `effect.Amount` as an additive DPS magnitude regardless of `ModifierType` — moot once Multiplicative-on-accumulator can no longer be authored.
- No engine/rating-model changes were made — the fix is purely at the authoring boundary, per the issue's suggested direction.

## Test plan

- [x] `Game.Application.Tests/DataAccess/AdminSkillsIntegrationTests.cs` — new integration tests pin the guard: a Multiplicative DoT-accumulator effect is rejected without persisting, and an Additive one still succeeds (guard doesn't over-reject).
- [x] `Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs` — new unit tests pin the lint rule: Multiplicative-on-accumulator is an Error, Additive-on-accumulator produces no finding.
- [x] `dotnet build` — full solution builds clean.
- [x] `dotnet test --no-build --no-restore` — full suite green (1398 + 68 + 1036 + 833 tests, 0 failures).

Closes #2169

---
_Generated by [Claude Code](https://claude.ai/code/session_01KJshFp9ys3fZw78CL7cnRW)_

### DIFF
--- a/Game.Application/Content/ProgressionGraphChecker.cs
+++ b/Game.Application/Content/ProgressionGraphChecker.cs
@@ -614,8 +614,6 @@ namespace Game.Application.Content
                 }
             }
 
-            // --- Skill recipes ----------------------------------------------------------------------------
-
             // --- Skill effects (DoT-accumulator modifier shape) -------------------------------------------
 
             /// <summary>
@@ -641,6 +639,8 @@ namespace Game.Application.Content
                     }
                 }
             }
+
+            // --- Skill recipes ----------------------------------------------------------------------------
 
             private void CheckSkillRecipes()
             {
--- a/Game.Application/Content/ProgressionGraphChecker.cs
+++ b/Game.Application/Content/ProgressionGraphChecker.cs
@@ -85,6 +85,7 @@ namespace Game.Application.Content
                 CheckItems();
                 CheckItemMods();
                 CheckProficiencies();
+                CheckSkillEffects();
                 CheckSkillRecipes();
                 CheckOrphanSkills();
                 CheckRecipeInputOwnability();
@@ -614,6 +615,32 @@ namespace Game.Application.Content
             }
 
             // --- Skill recipes ----------------------------------------------------------------------------
+
+            // --- Skill effects (DoT-accumulator modifier shape) -------------------------------------------
+
+            /// <summary>
+            /// A DoT-accumulator effect's magnitude is frozen with the caster's typed amplification at apply
+            /// time (spike #1320 Area C, <see cref="BattleContext.ApplySkillEffect"/>). That's correct for an
+            /// Additive effect, but a Multiplicative one would compound the amplification a second time when
+            /// it folds into the stack (#2169). The admin save path already rejects this combination
+            /// (<c>AdminSkills.SetEffects</c>), so this is a defense-in-depth backstop rather than a reachable
+            /// gap — matching the same-path-prerequisite pattern.
+            /// </summary>
+            private void CheckSkillEffects()
+            {
+                foreach (var skill in Live(_graph.Skills, s => s.RetiredAt))
+                {
+                    foreach (var effect in skill.Effects)
+                    {
+                        if (effect.ModifierTypeId == EModifierType.Multiplicative
+                            && DamageTypes.DotTypeForAccumulator(effect.AttributeId) is not null)
+                        {
+                            Error("SkillEffectDotModifier", "Skill", skill.Id,
+                                $"has a Multiplicative effect on DoT-accumulator attribute {effect.AttributeId}, which would double-apply the caster's amplification.");
+                        }
+                    }
+                }
+            }
 
             private void CheckSkillRecipes()
             {
--- a/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
+++ b/Game.Application.Tests/Content/ProgressionGraphCheckerTests.cs
@@ -712,6 +712,53 @@ namespace Game.Application.Tests.Content
             AssertHasFinding(graph, "ProficiencyPath", ContentGraphSeverity.Error, "Proficiency", 0);
         }
 
+        // --- Skill effects (DoT-accumulator modifier shape) -------------------------------------------
+
+        [Fact]
+        public void Skill_MultiplicativeEffectOnDotAccumulator_IsError()
+        {
+            var graph = HealthyGraph() with
+            {
+                Skills = HealthyGraph().Skills.Append(Skill(6, ESkillAcquisition.Player, effects:
+                [
+                    new Contracts.SkillEffect
+                    {
+                        Target = ESkillEffectTarget.Opponent,
+                        AttributeId = EAttribute.PoisonDamagePerSecond,
+                        ModifierTypeId = EModifierType.Multiplicative,
+                        Amount = 2m,
+                        DurationMs = 3000,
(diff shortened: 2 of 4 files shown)


Game.Application/Content/ProgressionGraphChecker.cs
```diff
@@ -615,6 +616,32 @@ private void CheckProficiencies()
 
             // --- Skill recipes ----------------------------------------------------------------------------
 
+            // --- Skill effects (DoT-accumulator modifier shape) -------------------------------------------
```
ginderjeremiah, line 617: Nit (cosmetic): the pre-existing `// --- Skill recipes ---` banner (line 617) now sits directly above this new `Skill effects` section and its `CheckSkillEffects` method, so that banner is orphaned from `CheckSkillRecipes` below. Either move the `Skill recipes` banner down to just above `CheckSkillRecipes`, or place this `Skill effects` section after the recipes one, so each banner leads its own method. Purely a readability thing — no functional impact.

---

ginderjeremiah (commented):
## Review: Reject Multiplicative skill effects on DoT-accumulator attributes (#2169)

**Verdict:** Approve (posted as a comment since I'm the PR author, so GitHub won't accept a gating event) — the fix is correct, well-scoped, and matches the issue's suggested direction. One cosmetic nit, no blockers.

I verified the underlying defect rather than taking the description at its word:
- `BattleContext.ApplySkillEffect` (`Game.Core/Battle/BattleContext.cs:78-81`) freezes the caster's typed amplification into any DoT-accumulator effect regardless of `ModifierType`, and `CombatRating` (`CombatRating.cs:193-196`) reads `effect.Amount` as an additive DPS magnitude. So a Multiplicative-on-accumulator effect would compound amplification twice (engine) and be mispriced (rating) — exactly as #2169 describes.
- Closing it at the authoring boundary rather than teaching the engine/rating multiplicative-DoT semantics is the right call: the design treats a DoT effect as additive per-second data, and there's no seeded content authoring the combination, so this introduces no live-content findings.

### Should-fix
- _None._

### Nits
- `Game.Application/Content/ProgressionGraphChecker.cs:617-619` — the new `Skill effects` section is inserted between the pre-existing `// --- Skill recipes ---` banner and `CheckSkillRecipes`, orphaning that banner. Cosmetic only (inline comment left).

### Test coverage
Solid on both enforcement points, per the project's testing rules:
- `AdminSkillsIntegrationTests` pins the real save path — Multiplicative-on-accumulator returns failure **and** persists nothing; Additive still succeeds (guards against over-rejection).
- `ProgressionGraphCheckerTests` pins the content lint — Multiplicative → `SkillEffectDotModifier` Error, Additive → no finding.
- No battle-parity test is needed: this changes authoring validation, not battle logic, so the frontend/backend engine behavior (and their existing parity) is untouched.

### What's good
- The guard mirrors the existing anti-tamper pattern in `AdminSkills`, and correctly excludes `Delete` changes while covering both Add and Edit — the two authoring paths.
- The content-lint rule is honestly framed as an unreachable defense-in-depth backstop (citing the #2128/#2144 same-path precedent), not oversold as a live gap.
- The scope note is accurate: with the authoring guard in place, `CombatRating.OffenseRate`'s additive-only DoT closed form can no longer be fed a Multiplicative factor, so leaving it untouched is correct rather than a deferred bug.

---

PR #2182, opened 2026-07-19
Cap CombatRating's avoidance/resistance credit to prevent survivability blowup

### LINKED ISSUE #2171: CombatRating survivability has an unclamped avoidance pole — effective parry+dodge near 1 inflates the rating ~10⁶× and zeroes the build's XP/proficiency gains
Found during a periodic codebase-health review. Distinct from #2049 (the DoT-term expiry ramp) — this is the avoidance term's pole.

`CombatRating.Survivability` (`Game.Core/Battle/CombatRating.cs:260-278`):

```csharp
avoid = parry + (1 - parry) * dodge;
...
var mitigationFraction = (1 - avoid) * (1 - resist) * (1 - toughnessReduction);
return effectiveHealth / Math.Max(mitigationFraction, Epsilon);
```

`parry`/`dodge` are the raw `chance × multiplier` products — authored-only enablers with **no cap**, amplified by the Luck/Agility-derived multipliers and stackable via the shared-expiry effect ramp (the exact ramp `game-design-combat.md` warns about: a stance authored with duration > cooldown ramps `ParryChance` toward permanence). The battle engine degrades gracefully at chance ≥ 1 (the draw simply always avoids), but the rating diverges as `avoid → 1` and hits the `Epsilon = 1e-6` floor at `avoid ≥ 1`: survivability becomes `effectiveHealth × 10⁶`.

**Downstream effect:** the huge `PlayerRating` drives `DefeatRewards.DifficultyMultiplier = min(enemy/player, 1)² ≈ 0` and the proficiency denominator `max(playerRating, enemyRating)`, so a committed-avoidance build earns effectively **zero XP and zero proficiency** — permanently — and the UI surfaces an absurd rating number. The same shape exists for `(1 - resist)` under heavy authored absorption, though that needs an average > 1 across the 8-type mix and is much less reachable.

**Suggested fix:** cap the avoidance (and resist) credit below 1 inside the rating model (e.g. `Math.Min(avoid, 0.95)`) — the engine's real counter to a 100%-avoid build is DoT, which the survivability model doesn't price, so an uncapped credit overstates what avoidance is actually worth anyway. Add boundary tests at/above the pole.

ginderjeremiah:
## Summary

`CombatRating.Survivability` (`Game.Core/Battle/CombatRating.cs`) computed `mitigationFraction = (1 - avoid) * (1 - resist) * (1 - toughnessReduction)` with `avoid` (parry+dodge) and `resist` (average incoming resistance) both **uncapped, authored-only** enablers. Since `avoid`/`resist` can be amplified by Luck/Agility-derived multipliers and stacked via the shared-expiry effect ramp, they can approach or exceed 1 — at which point `mitigationFraction` hits the `Epsilon = 1e-6` floor and `survivability` (and hence `PlayerRating`) blows up to `effectiveHealth × 10⁶`.

That runaway rating drives `DefeatRewards.DifficultyMultiplier = min(enemy/player, 1)² ≈ 0` and the proficiency denominator `max(playerRating, enemyRating)`, so a committed-avoidance (or heavy-resistance) build earns effectively **zero XP and zero proficiency** — permanently.

## Fix

- Added `ServerGameConstants.MaxMitigationCredit = 0.95` — the upper clamp on both the avoidance and resistance credit the survivability term prices.
- `CombatRating.Survivability` now clamps `avoid` and `resist` to this cap before computing `mitigationFraction`, so the pole is unreachable while normal (non-degenerate) builds are unaffected.

The battle engine itself already degrades gracefully at chance ≥ 1 (a draw simply always avoids) — this only fixes the *rating model's* divergence, matching the fix already applied to the DoT-term ramp pricing elsewhere in the same file.

## Test plan

- [x] Added unit tests in `Game.Core.Tests/Battle/CombatRatingTests.cs` pinning the boundary behavior: parry at exact certainty, dodge/resistance authored past certainty, and confirmation that the rating no longer moves once the credit clears the cap.
- [x] `dotnet test` (no-build) — full backend suite passes: 1404 (Core) + 68 (Infrastructure) + 1036 (Application, includes the combat-rating calibration suite) + 833 (Api) = 3341 tests, 0 failures. No calibration drift observed.
- [x] No frontend changes — combat rating is server-only with no frontend mirror (per `docs/backend.md`).

Closes #2171

---
_Generated by [Claude Code](https://claude.ai/code/session_019XbgafXWtX9LqcK2eK4rts)_

### DIFF
--- a/Game.Core.Tests/Battle/CombatRatingTests.cs
+++ b/Game.Core.Tests/Battle/CombatRatingTests.cs
@@ -55,6 +55,70 @@ namespace Game.Core.Tests.Battle
             Assert.True(CombatRating.Rate(tough, isPlayer: true) > CombatRating.Rate(weak, isPlayer: true));
         }
 
+        // ── Rate: survivability avoidance/resistance pole (#2171) ───────────────
+
+        [Fact]
+        public void Rate_ParryAtCertainty_DoesNotExplodeRating()
+        {
+            // ParryChance × ParryChanceMultiplier(base 1) = 1.0 drives raw `avoid` to exactly 1, the pole where
+            // the uncapped mitigation fraction hits its Epsilon floor. The capped credit must keep the rating
+            // finite and in the same order of magnitude as a merely-strong (not fully-certain) avoidance build.
+            var certainParry = MakeBattlerWithSkills([(ParryChance, 1.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+            var strongParry = MakeBattlerWithSkills([(ParryChance, 0.5)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+
+            var certainRate = CombatRating.Rate(certainParry, isPlayer: true);
+            var strongRate = CombatRating.Rate(strongParry, isPlayer: true);
+
+            Assert.False(double.IsNaN(certainRate));
+            Assert.False(double.IsInfinity(certainRate));
+            Assert.True(certainRate > strongRate);
+            Assert.True(certainRate < strongRate * 100);
+        }
+
+        [Fact]
+        public void Rate_DodgeBeyondCertainty_DoesNotExplodeRating()
+        {
+            // DodgeChance authored beyond what a chance can mean in-fight (2.0 × base-1 multiplier = 2.0) still
+            // must not push the capped avoidance credit above ServerGameConstants.MaxMitigationCredit.
+            var overCertainDodge = MakeBattlerWithSkills([(DodgeChance, 2.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+
+            var rate = CombatRating.Rate(overCertainDodge, isPlayer: true);
+
+            Assert.False(double.IsNaN(rate));
+            Assert.False(double.IsInfinity(rate));
+        }
+
+        [Fact]
+        public void Rate_ParryBeyondCertainty_RatesTheSameAsExactlyAtTheCap()
+        {
+            // Once avoid clears MaxMitigationCredit, further ParryChance must not move the rating at all —
+            // pins that the credit is truly capped (not just softened) past the pole.
+            var atCap = MakeBattlerWithSkills([(ParryChance, 1.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+            var wayPastCap = MakeBattlerWithSkills([(ParryChance, 50.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+
+            Assert.Equal(
+                CombatRating.Rate(atCap, isPlayer: true),
+                CombatRating.Rate(wayPastCap, isPlayer: true), 6);
+        }
+
+        [Fact]
+        public void Rate_ResistanceBeyondCertainty_DoesNotExplodeRating()
+        {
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Cap CombatRating's avoidance/resistance survivability credit (closes #2171)

**Verdict:** Approve — the fix is correct, minimal, and squarely matches the issue's diagnosis and suggested approach; no blocking findings, should-fixes, or nits. (Submitted as a `COMMENT` since GitHub won't accept an `APPROVE` event on my own PR — treat this as an approval.)

I traced the pole the issue describes: `avoid = parry + (1 - parry) * dodge` and `resist = AverageIncomingResistance(...)` are authored-only, uncapped enablers that sit as `(1 - credit)` in the survivability denominator, so `avoid → 1` collapses `mitigationFraction` onto the `Epsilon = 1e-6` floor and blows `PlayerRating` up ~10⁶×, which then zeroes the build's XP (`DifficultyMultiplier`) and proficiency (`max(playerRating, enemyRating)` denominator). Clamping both credits to `MaxMitigationCredit = 0.95` before computing `mitigationFraction` makes the pole unreachable while leaving non-degenerate builds untouched.

### What's good
- Correct use of `Math.Min` rather than `Math.Clamp`: `resist` can be negative (net vulnerability across the type mix), and a lower clamp would wrongly floor that vulnerability at 0 and *understate* danger. `Math.Min` caps only the runaway upper end and preserves the sub-zero (vulnerability) tail. `avoid` is non-negative so it needs no lower bound either. This is the subtle detail an easy `Clamp(x, 0, cap)` "cleanup" would get wrong — the choice is deliberately right.
- Constant lands in the correct place. `ServerGameConstants` is explicitly server-only / not client-mirrored, and the combat rating is a server-authoritative reward computation — so this correctly needs no frontend mirror (verified: no `CombatRating`/`MaxMitigationCredit` reference under `frontend/src`). The battle-logic frontend/backend consistency rule applies to the battle *engine*, not this rating model.
- The XML doc on the constant is genuinely useful context (why the pole exists, why the engine tolerates chance ≥ 1 but the rating doesn't, and why 0.95 rather than 1 — DoT is the uncapped-credit's missing counter-pressure). Slightly long but every line earns its place per the project's documentation guidance; not flagging.
- New constant is distinct from the nearby effect-uptime ramp cap (`CombatRating.cs:423`) and the legacy reward clamps — no duplication of an existing knob.

### Test coverage
Sufficient and on-point for the project's rules — the issue asked for "boundary tests at/above the pole" and the PR delivers exactly that in `Game.Core.Tests/Battle/CombatRatingTests.cs`:
- `Rate_ParryAtCertainty_DoesNotExplodeRating` — parry × base-1 multiplier = 1.0 (the exact pole) stays finite and same-order-of-magnitude as a merely-strong build.
- `Rate_DodgeBeyondCertainty` / `Rate_ResistanceBeyondCertainty` — authored past certainty (dodge 2.0; all eight resistance leaves stacked) stays finite.
- `Rate_ParryBeyondCertainty_RatesTheSameAsExactlyAtTheCap` — pins that the credit is *capped*, not merely softened (parry 1.0 vs 50.0 rate to 6 decimals equal).

Backend-only change, correctly no frontend suite touched.

Nice, tightly-scoped fix that mirrors the earlier survivability-term hardening.

---

PR #2183, opened 2026-07-19
Bound ReferenceCacheSynchronizer.StopAsync to the host shutdown token

### LINKED ISSUE #2172: ReferenceCacheSynchronizer.StopAsync awaits the reload loop unbounded, ignoring the host shutdown token

`Game.DataAccess/ReferenceCacheSynchronizer.cs:43-48`:

```csharp
public async Task StopAsync(CancellationToken cancellationToken)
{
    await _pubsub.UnSubscribe(InstanceId);
    _stopping.Cancel();
    await _reloadLoop;
}
```

`cancellationToken` is never consulted. The loop's exit is gated on every in-flight holder reload settling — `CoalescingReferenceCacheReloader.ReloadConcurrentlyAsync` does `await Task.WhenAll(reloads)`, which completes only after all reload queries unwind — so a wedged Npgsql connection during a background sweep stalls host shutdown until the host's 30s deadline force-abandons it.

Every sibling stop path bounds this exact scenario:

- `DataProviderSynchronizer.StopAsync` uses a 5s `_drainTimeout`
- `RedisConnectionLifetime.StopAsync` uses `.WaitAsync(cancellationToken)`
- `AdminCacheReloadFilter` exists precisely because "a wedged database connection would otherwise hold the admin request open indefinitely"

Low-probability trigger (wedged DB + concurrent sweep + deploy), but the fix is one `.WaitAsync(cancellationToken)` with an `OperationCanceledException` swallow, matching the established pattern.

ginderjeremiah:
## Summary

`ReferenceCacheSynchronizer.StopAsync` awaited its background reload loop unbounded, never consulting the `cancellationToken` the host passes it. Since the loop's exit is gated on `Task.WhenAll` over every in-flight cache reload settling, a wedged Npgsql connection during a background sweep would stall host shutdown until the host's own ~30s deadline force-abandoned it.

Every sibling stop path in the codebase already bounds this exact scenario (`DataProviderSynchronizer._drainTimeout`, `RedisConnectionLifetime.StopAsync`'s `.WaitAsync(cancellationToken)`), so this brings `ReferenceCacheSynchronizer` in line with the established pattern rather than introducing a new one.

## Change

`Game.DataAccess/ReferenceCacheSynchronizer.cs`: `StopAsync` now awaits the reload loop via `.WaitAsync(cancellationToken)` and swallows the resulting `OperationCanceledException`, mirroring `RedisConnectionLifetime.StopAsync` — if the deadline trips first, shutdown proceeds and the loop is left to finish (or die with the process) in the background.

## Test plan

- [x] `dotnet build Game.sln` — clean build
- [x] `dotnet test Game.Application.Tests --no-build --no-restore` — all 1037 tests pass
- [x] Added `StopAsync_ReloadLoopWedged_GivesUpWhenTheShutdownDeadlineTrips`, which drives a reload that never completes (ignoring its own cancellation token, simulating a wedged Npgsql connection) and asserts `StopAsync` still returns promptly once an already-cancelled token is passed — bounded so a regression back to an unbounded await hangs the test rather than silently passing.

Closes #2172

---
_Generated by [Claude Code](https://claude.ai/code/session_01PtjWk5q5iJ1bZcy5FLiVk2)_

### DIFF
--- a/Game.Application.Tests/DataAccess/ReferenceCacheSynchronizerTests.cs
+++ b/Game.Application.Tests/DataAccess/ReferenceCacheSynchronizerTests.cs
@@ -9,6 +9,7 @@ using Game.TestInfrastructure.Helpers;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Xunit;
 
 namespace Game.Application.Tests.DataAccess
@@ -108,6 +109,31 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task StopAsync_ReloadLoopWedged_GivesUpWhenTheShutdownDeadlineTrips()
+        {
+            // Models a reload stuck on a wedged Npgsql connection (#2172): it never completes, ignoring the
+            // stopping token entirely, so StopAsync must still honor the host's own shutdown deadline rather
+            // than block on it.
+            var wedgedCache = new WedgedReloadableCache();
+            var policy = new ReferenceCacheReloadPolicy(TimeSpan.FromMilliseconds(10), maxAttempts: 1, baseDelay: TimeSpan.Zero);
+            var reloader = new CoalescingReferenceCacheReloader([wedgedCache], policy, NullLogger<CoalescingReferenceCacheReloader>.Instance);
+            var synchronizer = new ReferenceCacheSynchronizer(new NoOpPubSubService(), reloader);
+
+            await synchronizer.StartAsync(CancellationToken);
+            reloader.NotifyChanged();
+
+            // Wait for the reload to actually be in flight so StopAsync races a real wedged awaiter, not an idle loop.
+            await wedgedCache.ReloadStarted.Task.WaitAsync(TimeSpan.FromSeconds(5));
+
+            using var cts = new CancellationTokenSource();
+            await cts.CancelAsync();
+
+            // Returns (does not throw, does not hang) once the deadline has tripped, even though the reload
+            // loop is still stuck awaiting a reload that will never complete.
+            await synchronizer.StopAsync(cts.Token).WaitAsync(TimeSpan.FromSeconds(5));
+        }
+
+        [Fact]
         public async Task Dispose_WithoutStopAsync_DoesNotThrowAndIsIdempotent()
         {
             using var scope = CreateScope();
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Bound `ReferenceCacheSynchronizer.StopAsync` to the host shutdown token (#2172)

**Verdict:** Approve — correct, minimal fix that brings the last unbounded stop path in line with the established sibling pattern, with a test that genuinely reproduces the wedged-in-flight scenario. (Posted as a comment since I'm the PR author; GitHub won't accept a gating verdict on your own PR.)

The change replaces the unbounded `await _reloadLoop` with `await _reloadLoop.WaitAsync(cancellationToken)` plus an `OperationCanceledException` swallow — a line-for-line match of `RedisConnectionLifetime.StopAsync`, down to the comment framing. I verified the surrounding behavior rather than taking the diff at face value:

- **No unobserved-exception footgun on give-up.** After the deadline trips, `_reloadLoop` keeps running in the background. `RunAsync` only ever faults on `OperationCanceledException` (reload failures are caught inside `ReloadConcurrentlyAsync` and returned as a `ReloadFailure`), so the backgrounded task can't surface an unobserved exception — the give-up is safe.
- **`Dispose` stays coherent.** It already defers `_stopping.Dispose()` to `_reloadLoop.ContinueWith(...)`, so a still-running loop after give-up won't hit an `ObjectDisposedException` race on the token source.
- **Normal shutdown is unaffected.** `_stopping.Cancel()` unwinds `RunAsync` cleanly (its OCE is swallowed on cancellation), so `WaitAsync` returns without ever tripping the token in the common path.

### Test coverage

Adequate and well-constructed. `StopAsync_ReloadLoopWedged_GivesUpWhenTheShutdownDeadlineTrips` waits on `ReloadStarted` so `StopAsync` races a *real* in-flight wedged awaiter rather than an idle loop, passes an already-cancelled token, and bounds the assertion with `.WaitAsync(5s)` so a regression back to an unbounded await *hangs* the test instead of silently passing. The `WedgedReloadableCache` (ignores its cancellation token entirely) faithfully models the stuck-Npgsql-connection trigger from the issue, and it mirrors the sibling `RedisConnectionLifetime` test's structure. Test helper signatures (`ReferenceCacheReloadPolicy`, `CoalescingReferenceCacheReloader`, `NotSupportedPubSubService`, `IReloadableReferenceCache`) all match current definitions.

### What's good

- Reuses the existing bounded give-up idiom instead of inventing a new one — this was the last sibling stop path missing it.
- Comments explain the *why* (deliberate give-up, background completion) without over-narrating.

No blocking findings, should-fixes, or nits. I'll merge once CI is green.

---

Issue #2184, opened 2026-07-19
SocketCommandTimeoutTests.CooperativelyCancellingCommand_TimesOut_AndReleasesLockSoNextCommandRuns fails when run in isolation


ginderjeremiah:
## Context

Found while working on #2174 (socket command commit not honoring the per-command cancellation token). Unrelated to that fix — reproduces identically on `main` before my change.

## Problem

`Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs`'s `CooperativelyCancellingCommand_TimesOut_AndReleasesLockSoNextCommandRuns` uses a 200ms `ShortTimeout` budget for both a "Coop" command (parks on a never-released gate, cooperatively cancels) and a trivial "Next" command (`Task.CompletedTask` body) issued immediately after. The test asserts "Next" completes successfully within the 5s `WaitTimeout`.

**Reproduces 5/5 runs** when the test class is run in isolation (`dotnet test -- --filter-class "*SocketCommandTimeoutTests"`): the "Next" command itself comes back with a "Command timed out" error instead of succeeding, even though its own body is a no-op. **Passes reliably** when run as part of the full `Game.Api.Tests` suite (834/834 green).

## Likely cause

This looks like cold-start latency: when the class runs standalone, the first real DB/host round trips (scope creation, session setup, etc.) haven't been JIT/connection-pool warmed yet, so 200ms isn't enough headroom for "Next" to make it through a real integration-test host. When run after hundreds of other tests, the pools are already warm and 200ms is plenty.

## Suggested fix

Either bump `ShortTimeout` to something with more headroom (e.g. 500ms–1s) or add a cheap warmup command before the timed portion of the test, so the test isn't sensitive to whichever order/isolation it happens to run under.

Tags: tech debt, claude, scope: small

PR #2185, opened 2026-07-19
Pass the per-command cancellation token to the socket UoW commit

### LINKED ISSUE #2174: Socket command unit-of-work commit ignores the per-command cancellation token

`Game.Api/Sockets/SocketHandler.cs:303-305`:

```csharp
var command = _commandFactory.CreateCommand(commandInfo, scope);
var response = await command.ExecuteAsync(_context, cancellationToken);
await scope.ServiceProvider.GetRequiredService<IUnitOfWork>().CommitAsync();
```

`RunCommand` threads the per-command timeout token into `ExecuteAsync` and the reload path, but the UoW commit is called with no token. This deviates from the documented contract (`backend-sockets.md`: the per-command token "is plumbed down through the application services and the data tier… so a wedged command unwinds promptly and releases the lock") and from the HTTP twin, `CommitFilter` (`Game.Api/Filters/CommitFilter.cs:19`), which passes `context.HttpContext.RequestAborted`.

**Failure scenario:** a command whose EF flush wedges can't be cut off at the per-command budget — the timed-out command's lock-handoff continuation (`ReleaseCommandLockWhenSettled`) keeps the player's entire command stream blocked until Npgsql's own ~30s command timeout unwinds it. The one hop the budget can't cancel is exactly the last one before lock release. No comment marks the omission as deliberate.

**Fix:** pass `cancellationToken` to `CommitAsync`, mirroring `CommitFilter`.

ginderjeremiah:
## Summary

`SocketHandler.RunCommand` calls `command.ExecuteAsync(_context, cancellationToken)` with the per-command budget token, but then called `IUnitOfWork.CommitAsync()` with no token at all. This deviates from the documented contract (`backend-sockets.md`: the per-command token "is plumbed down through the application services and the data tier… so a wedged command unwinds promptly and releases the lock") and from the HTTP twin, `CommitFilter`, which passes `RequestAborted`.

**Failure scenario:** a command whose EF flush wedges can't be cut off at the per-command budget — the timed-out command's lock-handoff continuation (`ReleaseCommandLockWhenSettled`) keeps the player's entire command stream blocked until Npgsql's own ~30s command timeout unwinds it instead. The one hop the budget couldn't cancel was exactly the last one before lock release.

## Changes

- `Game.Api/Sockets/SocketHandler.cs`: pass `cancellationToken` into `CommitAsync`, mirroring `CommitFilter`.
- Updated two adjacent comments in the same method that asserted an abandoned command's commit *always* lands — that's no longer universally true post-fix: if the flush itself is the wedge point, it now unwinds under the same (already-cancelled) token instead of riding out Npgsql's own timeout.

## Investigation note

Before applying the literal one-line fix, I checked it against `Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs`, which has an existing test explicitly asserting "the abandoned command still commits its write." Passing an already-cancelled token to a real `SaveChangesAsync` could in principle short-circuit that guarantee for a command with pending EF changes. I ran that integration test suite before and after the fix (real Postgres, via `ApiIntegrationTestBase`) — it passes identically both ways, because that test's stub command performs no EF mutations (nothing to save), so the risk doesn't materialize there. The comment updates above make the actual (narrower) guarantee explicit for future readers instead of leaving the stale absolute claim in place.

## Test plan

- [x] Added `ExecuteCommand_Succeeds_CommitsUnderThePerCommandCancellationToken` (`Game.Api.Tests/Unit/SocketCommandExecutionTests.cs`) — a `CapturingUnitOfWork` fake pins that `CommitAsync` now receives a real, cancellable token instead of `CancellationToken.None`.
- [x] `dotnet build Game.sln` — succeeds, no warnings.
- [x] `dotnet test Game.Api.Tests` (full suite) — 834/834 passing.
- [x] `dotnet test Game.Api.Tests -- --filter-class "*SocketCommandExecutionTests"` and `*SocketCommandTimeoutTests*` — passing (see follow-up note below).

## Follow-up filed

While verifying, I found `SocketCommandTimeoutTests.CooperativelyCancellingCommand_TimesOut_AndReleasesLockSoNextCommandRuns` fails reproducibly when that test class is run in isolation (5/5 runs, both before and after this change — pre-existing, unrelated flakiness, likely cold-start latency against the 200ms test budget) but passes cleanly as part of the full suite. Filed as #2184.

Closes #2174

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kyp72jxSgNa23g85MRUY9N)_

### DIFF
--- a/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandExecutionTests.cs
@@ -379,6 +379,32 @@ namespace Game.Api.Tests.Unit
             Assert.False(session.PlayerNeedsReload);
         }
 
+        [Fact]
+        public async Task ExecuteCommand_Succeeds_CommitsUnderThePerCommandCancellationToken()
+        {
+            // #2174: the unit-of-work commit must observe the same per-command budget as ExecuteAsync — a
+            // flush that itself wedges past the timeout must be cancellable rather than riding out Npgsql's
+            // own command timeout, mirroring CommitFilter's HTTP-side behavior (which passes RequestAborted).
+            var unitOfWork = new CapturingUnitOfWork();
+            using var provider = new ServiceCollection()
+                .AddScoped<IUnitOfWork>(_ => unitOfWork)
+                .BuildServiceProvider();
+
+            var socket = new FakeWebSocket(sendDuration: TimeSpan.Zero);
+            var session = new SessionService(new NoOpSessionStore());
+            await session.CreateSession(userId: 1, playerId: 1);
+            var context = new SocketContext(socket, playerId: 1, session, isAdmin: false, _loggerFactory.CreateLogger<SocketContext>());
+            var handler = new SocketHandler(context, new StubCommandFactory(_ => null), provider.GetRequiredService<IServiceScopeFactory>(),
+                _loggerFactory.CreateLogger<SocketHandler>(), () => { });
+
+            await handler.ExecuteCommand(new SocketCommandInfo("Ok") { Id = "c1" });
+
+            // A token backed by a real CancellationTokenSource reports CanBeCanceled; the prior default-token
+            // call (CancellationToken.None) would not, so this fails without the fix.
+            Assert.NotNull(unitOfWork.ReceivedToken);
+            Assert.True(unitOfWork.ReceivedToken!.Value.CanBeCanceled);
+        }
+
         private (FakeWebSocket Socket, SocketHandler Handler) CreateHandler(Func<string, Exception?> throwOn, Func<string, Exception?>? throwOnCreate = null, Func<string, bool>? selfDelivering = null)
         {
             var socket = new FakeWebSocket(sendDuration: TimeSpan.Zero);
@@ -489,6 +515,17 @@ namespace Game.Api.Tests.Unit
         private sealed class NoOpUnitOfWork : IUnitOfWork
         {
             public Task CommitAsync(CancellationToken cancellationToken = default) => Task.CompletedTask;
+        }
+
+        private sealed class CapturingUnitOfWork : IUnitOfWork
+        {
+            public CancellationToken? ReceivedToken { get; private set; }
+
+            public Task CommitAsync(CancellationToken cancellationToken = default)
+            {
+                ReceivedToken = cancellationToken;
+                return Task.CompletedTask;
+            }
         }
 
         /// <summary>Returns each of <paramref name="responses"/> in order on successive <see cref="GetPlayer"/>
--- a/Game.Api/Sockets/SocketHandler.cs
+++ b/Game.Api/Sockets/SocketHandler.cs
@@ -202,9 +202,11 @@ namespace Game.Api.Sockets
                     // mid read-modify-write of the shared cached Player, and releasing would let the next
                     // command race it (the lost-update class the per-socket command lock prevents — see
                     // docs/backend.md). Hand the lock to a continuation that releases it once the abandoned
-                    // task finally settles. The abandoned command still commits its write (so it isn't lost),
-                    // but its now-late response is dropped: RunCommand never sends — this method owns the single
-                    // send — so the client never receives a second response for the same id.
+                    // task finally settles. The abandoned task's own commit runs under this same (already
+                    // cancelled) token, so a flush that is itself the wedge point unwinds promptly rather than
+                    // riding out Npgsql's own command timeout; its now-late response is dropped regardless:
+                    // RunCommand never sends — this method owns the single send — so the client never receives
+                    // a second response for the same id.
                     _logger.LogWarning("Socket command timed out after {Timeout} and was abandoned: {CommandInfo} on socket: {Id}", _commandTimeout, commandInfo, Id);
                     await SendErrorAsync(commandInfo, "Command timed out.");
                     ReleaseCommandLockWhenSettled(commandTask, cts, commandInfo);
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Pass the per-command cancellation token to the socket UoW commit (closes #2174)

**Verdict:** Approve — correct, minimal, well-tested fix that aligns the socket path with its documented contract and its HTTP twin. (Posted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author.)

The change makes `RunCommand`'s `CommitAsync` observe the same per-command budget token as `ExecuteAsync`, exactly as `backend-sockets.md` describes ("the per-command token is plumbed down through the application services and the data tier… so a wedged command unwinds promptly and releases the lock") and as `CommitFilter.cs:19` already does on the HTTP side. After this change both `CommitAsync` call sites in the codebase pass a token; no tokenless call remains.

### What's good

- The behavioral consequence is correctly reasoned through and, importantly, integrates cleanly with the existing abandoned-command continuation: when a wedged flush now unwinds under the already-cancelled token, the background task settles as `Canceled`, hits the `task.IsCanceled` branch in `ReleaseCommandLockWhenSettled` (`SocketHandler.cs:322`), and marks the player for reload — so the discarded write doesn't silently ride along. Nothing new was needed there, which is a good sign the fix lands in the right seam.
- The comment updates are honest about the semantics shift: the abandoned command's commit is no longer *guaranteed* to land, and the stale absolute claim ("still commits its write") was correctly replaced rather than left to mislead. Detailed, but the reasoning (timeout ↔ lock ↔ flush interaction) is genuinely non-obvious and worth keeping.
- Good investigation discipline — checking the fix against the existing `SocketCommandTimeoutTests` guarantee and documenting why the risk doesn't materialize there.

### Test coverage

- `ExecuteCommand_Succeeds_CommitsUnderThePerCommandCancellationToken` is well-targeted: a CTS-backed token reports `CanBeCanceled == true` while the prior `CancellationToken.None` reports `false`, so the assertion genuinely fails without the fix. `CapturingUnitOfWork` is the right minimal fake.

### Nits (optional)

- The genuinely *new* behavior this PR introduces — an abandoned command whose commit observes the already-cancelled token has its write **discarded** (and the player reloaded) rather than persisted — has no direct test. The unit test proves token propagation, and the existing `SocketCommandTimeoutTests` don't cover it (per your own note, that suite's stub performs no EF mutations, so nothing is at stake there). An integration test with a command that has real pending EF changes and wedges at the flush would pin this discard-on-cancel path. Not blocking: the discard is really EF/Npgsql cancellation semantics, and this is proportionate coverage for a one-line plumbing fix — flagging it only so the coverage gap is a conscious choice.

Leaving the merge to you rather than auto-merging, given the optional suggestion above and that `test-backend` was still running at review time.

---

ginderjeremiah:
Thanks for the review. I looked into the optional nit (an integration test pinning the discard-on-cancel path for a command with real pending EF changes) and I'm leaving it out — here's why.

I built exactly that test: a command that stages a real `TagCategory` insert on its DI-scoped `GameContext`, ignores the token, and only returns after the budget elapses, so its `CommitAsync` runs under the already-cancelled token. Running it standalone, it passed — the write appeared discarded and the player got marked for reload, matching the assumption in my comment update.

But running the full suite with it in place, a *different*, unrelated test (`TagsControllerTests.TagCategories_ReturnsIntrinsicCategories`) failed with an extra row (8 instead of 7). I checked the shared test Postgres directly and the `wedge-discard-*` row had, in fact, landed — even though the abandoned task still reported `Canceled` and `PlayerNeedsReload` still got set.

So the discard isn't actually guaranteed for a fast/local write: it looks like EF's cancellation check for `SaveChangesAsync` can occur *after* the INSERT has already round-tripped and committed at the Postgres level, when the operation completes fast enough. The .NET task still surfaces `OperationCanceledException` (so the lock-release/reload-marking behavior is unaffected), but the row itself can persist anyway. Cancellation here is genuinely best-effort for an already-fast/in-flight-too-briefly-to-interrupt operation — it reliably cuts off a **slow** wedge (the actual scenario #2174 describes), but a hard assertion that a fast operation's write never lands would be flaky by construction, and it already proved that by polluting the shared test DB for a later test.

I reverted that addition (and cleaned up the leaked row). Given a reliable version would need to actually inject a slow/blocking DB operation (e.g. an advisory lock) to force cancellation to land mid-flight rather than post-hoc, I agree with your framing that this is disproportionate for a one-line plumbing fix — flagging it here so the coverage gap stays a documented, conscious choice rather than a rediscovered surprise later.

Current state: PR is back to the original 3 commits' worth of diff (just the `SocketHandler.cs` fix + comment updates + the token-propagation unit test), full `Game.Api.Tests` suite green (834/834), `test-backend` CI was still running at last check — will keep an eye on it.

---

ginderjeremiah:
## Review pass — confirming and merging

Re-checked this PR against the codebase and CI. Nothing has changed since the earlier approving review, so this is a confirmation rather than a fresh review.

**Correctness:** The one-line fix — `CommitAsync(cancellationToken)` in `RunCommand` — is right. It aligns the socket path with its HTTP twin (`CommitFilter.cs:19` passes `RequestAborted`) and with the documented `backend-sockets.md` contract. `DataAccessCancellationTests` already proves `CommitAsync` honors an already-cancelled token at the data tier, so this closes a real gap rather than a cosmetic one, and it's the last remaining tokenless `CommitAsync` call in production code.

**Comments:** The updated comments honestly walk back the stale "still commits its write" absolute — appropriate, since a flush that is itself the wedge point now unwinds under the cancelled token. Detailed but the timeout↔lock↔flush interaction warrants it.

**Test:** `ExecuteCommand_Succeeds_CommitsUnderThePerCommandCancellationToken` genuinely fails without the fix (`CanBeCanceled` distinguishes the real token from `CancellationToken.None`). The one optional nit (an integration test pinning the discard-on-cancel path) was investigated and consciously dropped — the follow-up comment demonstrates a hard assertion there would be flaky by construction, which is the correct call for a one-line plumbing fix.

**CI:** All checks green (`test-backend`, `verify-backend`, `test-e2e` ×3, `all-checks-passed`; frontend correctly skipped).

**Verdict: approve.** Green CI, no outstanding nits or should-fixes, no follow-up work required. Merging.

---

Issue #295, opened 2026-06-09
Exclude retired reference records from admin authoring pickers


ginderjeremiah:
Follow-up to #289 (retire lifecycle for reference data).

#289 introduced retiring reference records (items, item mods, skills, enemies, zones, challenges) so they stay resolvable by id but are out of circulation for *new* organic exposure (a retired enemy no longer rolls in random spawns). It deliberately left the **admin authoring pickers** untouched: an admin can still newly point a challenge reward, a zone's boss enemy / spawn table, a zone's unlock gate, or a challenge's target entity at a **retired** record.

## Scope

- In the Workbench reference option lists (`UI/src/routes/admin/workbench/reference.svelte.ts` and each entity's select-option providers), exclude retired records from the **selectable** options for new authoring.
- BUT still **display** an already-authored reference to a retired record (e.g. a challenge whose reward is now retired) rather than silently blanking it — likely shown with a "retired" affordance, and not offered when re-picking.
- Decide whether picking a retired record should be a hard block or a warning.

## Notes

- The data is already available: the read contracts carry `retiredAt`, and the frontend caches include retired records at their slot.
- Keep the reference-data lookup sites themselves unchanged (retired records must remain resolvable by id).

PR #2186, opened 2026-07-19
Exclude tiers of retired paths from the gateway prerequisite picker

### LINKED ISSUE #2176: Gateway prerequisite picker offers tiers of retired paths — authors a permanent soft-lock the save accepts silently
Found during a periodic codebase-health review. The exact analogue of the same-path gap fixed in #2128, in the direction #2128 didn't cover.

`GatewaysEditor.svelte`'s `addOptions` (`UI/src/routes/admin/workbench/progression/GatewaysEditor.svelte:64-69`) filters candidates by:

```js
p.pathId !== tier.pathId && !p.retiredAt && !tier.prerequisiteIds.includes(p.id)
```

It never checks whether the candidate tier's **owning path** is retired. A tier on a retired path is frozen (accrues nothing, can never max), so picking it creates a gateway that can never open.

The backend accepts it: `AdminProficiencies.ValidatePrerequisiteIds` checks only self-reference/existence/same-path, and the soft-lock guard `FindRetiredPathGatingLiveGateway` (`AdminPaths.cs`) fires only on the *retire transition* — a path already retired before the save is skipped — so adding the edge *after* the retire has no save-time guard. Content Health then immediately reports it as an **Error** (`ProgressionGraphChecker.cs:600-603`: "retired/frozen, so this tier can never open").

**Repro:** retire path A (passes its guard — nothing references it yet) → open a live path B's root tier → the Gateways tab still offers A's tiers → add one → Save succeeds with no warning → B is permanently gated on frozen content; only the ops lint reveals it. This violates the documented picker rule (frontend-admin.md: retired records are "a hard block, not a warning" in authoring pickers).

**Fix shape (mirroring #2128's layered fix):** filter `addOptions` by the owning path's `retiredAt` (via `store.paths`), and ideally also reject in `ValidatePrerequisiteIds` as anti-tamper.

ginderjeremiah:
## Summary

`GatewaysEditor.svelte`'s `addOptions` filtered out a directly-retired tier from the "Add prerequisite" picker, but never checked whether the *candidate tier's owning path* was retired. A tier on a retired path is frozen — it never accrues and can never max — so picking it as a prerequisite authors a gateway that can never open. Content Health immediately flags this as an Error, but the save silently succeeds, leaving the author with no in-the-moment warning (violating the documented picker rule in `frontend-admin.md`: retired records are "a hard block, not a warning" in authoring pickers).

This is the exact analogue of the same-path gap fixed in #2128, in the direction #2128 didn't cover (owning-path retirement rather than same-path).

## Changes

- **Frontend** (`GatewaysEditor.svelte`): `addOptions` now also excludes any tier whose owning path — looked up live off `store.paths`, not the last-saved snapshot, so a path retired earlier in the same session is respected — is retired.
- **Backend** (`AdminProficiencies.ValidatePrerequisiteIds`): added the same rejection at save time as anti-tamper (a tampered admin client can't bypass the frontend check), mirroring the existing same-path guard from #2128.

## Test plan

- [x] Added `SetPrerequisites_PrerequisiteOnRetiredPath_ReturnsFailure` (`Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs`) — asserts the backend rejects a prerequisite on a retired path.
- [x] Added a new test to `GatewaysEditor.test.ts` (`#2176`) asserting a tier on a retired path is omitted from the picker even though the tier record itself isn't retired.
- [x] Updated the fake stores in `GatewaysEditor.test.ts`/`TierDetail.test.ts` to include `paths` now that the component reads it.
- [x] `dotnet build Game.sln` — succeeds, no warnings.
- [x] `dotnet test Game.sln` (full suite) — 3343/3343 passing.
- [x] `npm run lint` (UI) — clean.
- [x] `npm run test:unit:ci` (UI) — 3807/3807 passing.

Closes #2176

---
_Generated by [Claude Code](https://claude.ai/code/session_01UVq4kVSUaUtMo67Jjj2iFH)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs
@@ -632,6 +632,38 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SetPrerequisites_PrerequisiteOnRetiredPath_ReturnsFailure()
+        {
+            int gatedId, prerequisiteId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var gatedPath = await SeedPathAsync(seedScope);
+                gatedId = (await SeedProficiencyAsync(seedScope, gatedPath.Id, name: "Fire I")).Id;
+
+                var frozenPath = await SeedPathAsync(seedScope);
+                prerequisiteId = (await SeedProficiencyAsync(seedScope, frozenPath.Id, name: "Frost I")).Id;
+                frozenPath.RetiredAt = DateTime.UtcNow;
+                await context.SaveChangesAsync(CancellationToken);
+            }
+            await ReloadReferenceCachesAsync();
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminProficiencies>();
+
+            // A tier on a retired path is frozen — it never accrues, so a gateway prerequisiting it can
+            // never open. Reject at save time as anti-tamper, mirroring the frontend picker exclusion.
+            var result = admin.SetPrerequisites([new SetProficiencyPrerequisitesData
+            {
+                Id = gatedId,
+                PrerequisiteIds = [prerequisiteId],
+            }]);
+
+            Assert.False(result.Succeeded);
+            Assert.Contains("retired path", result.ErrorMessage);
+        }
+
+        [Fact]
         public async Task SetPrerequisites_WouldFormACycle_ReturnsFailure()
         {
             int gatedId, prerequisiteId;
--- a/Game.DataAccess/Repositories/Admin/AdminProficiencies.cs
+++ b/Game.DataAccess/Repositories/Admin/AdminProficiencies.cs
@@ -253,6 +253,12 @@ namespace Game.DataAccess.Repositories.Admin
                     return AdminSaveResult.Failure(
                         $"Prerequisite proficiency {prerequisiteId} is on the same path as proficiency {proficiency.Id} — gateways must be cross-path.");
                 }
+
+                if (_proficiencies.LookupPath(prerequisite.PathId) is { RetiredAt: not null })
+                {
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: exclude tiers of retired paths from the gateway prerequisite picker (#2176)

**Verdict:** Approve — clean, correctly-scoped fix that closes the exact gap #2176 describes, layered the same way as its sibling #2128. (Posting as a comment since I'm the PR author; the intended verdict is Approve.)

### What's good
- **Right diagnosis, right layers.** The frontend picker exclusion is the UX guard and the `ValidatePrerequisiteIds` check is the anti-tamper backstop — mirroring the same-path fix from #2128. This satisfies the documented picker rule (retired records are a hard block on *new* references — `frontend-admin.md` / `game-design.md:198`).
- **Existing references stay visible.** `addOptions` only filters what can be *added*; `prereqName` reads unfiltered `store.profs`, so a prerequisite authored before its path was retired still renders. That respects the "keep existing authored reference visible" half of the #295 rule.
- **Reactive lookup, not a stale snapshot.** `isPathRetired` reads live off `store.paths`, so a path retired earlier in the same session is honored without a save — the comment calls this out explicitly.
- **Null-safe backend check.** `LookupPath(...) is { RetiredAt: not null }` handles a missing path via the pattern (no null-forgiving operator), consistent with the existing `LookupPath` usages in `AdminPaths.cs`.

### Test coverage
- Backend: `SetPrerequisites_PrerequisiteOnRetiredPath_ReturnsFailure` covers the anti-tamper rejection with a real reference-cache reload — good integration-level coverage.
- Frontend: the new `GatewaysEditor.test.ts` case asserts a tier whose *own* record is live but whose owning path is retired is omitted, while a live-path tier remains — exactly the distinction #2176 is about. Fake-store updates to `TierDetail.test.ts`/`GatewaysEditor.test.ts` for the newly-read `paths` slot are appropriate.

No blocking findings and no nits. Consistent with existing patterns and the guideline docs. Holding the merge only until the remaining CI jobs (test-backend/test-frontend/e2e) finish green.

---

PR #2188, opened 2026-07-19
Render a level-0 slot in the Milestones editor for on-open modifiers

### LINKED ISSUE #2178: Level-0 proficiency modifiers are valid backend data the Milestones editor can neither display nor edit
Found during a periodic codebase-health review. Latent today — no seeded content carries a level-0 modifier and the UI can't create one — but the backend explicitly permits it, so the first one authored (via seed/import or a future path) becomes invisible and unremovable in the Workbench.

- Backend: `AdminProficiencies.cs:94-97` explicitly allows modifier level 0 ("Level 0 is allowed for modifiers: the cumulative bonus rule…", `minLevel: 0`), and the frontend validator agrees (`progression-helpers.ts` treats only `modifier.level < 0` as out of range).
- Editor: `MilestonesEditor.svelte:117-118,137-145` renders only levels `1..maxLevel` and clamps selection with `Math.min(Math.max(1, store.selectedLevel), maxLevel)`, so a level-0 modifier can never be selected, shown, or removed.
- Inconsistency surface: `payoutLevels()` (the Milestones tab count in `TierDetail`) would still count it — N payouts reported, N−1 visible.

**Fix options:** either render a level-0 slot in the timeline (making the editor honor the full valid range), or decide level 0 shouldn't be authorable and tighten the backend's `minLevel` to 1 — but the two ends should agree. Small either way.

ginderjeremiah:
## Summary

`AdminProficiencies.SetModifiers` explicitly allows a level-0 attribute modifier — the "on-open" bonus honored by `Proficiency.ModifiersForLevel`'s `l.Level <= level` rule (a payout authored at the just-opened state). But `MilestonesEditor`'s timeline and selectable level range started at 1 (`Math.max(1, Math.floor(tier.maxLevel) || 1)` / `Math.max(1, store.selectedLevel)`), so a level-0 modifier — valid, persistable backend data — could never be displayed, edited, or removed in the Workbench once authored (e.g. via seed/import or a future authoring path).

`payoutLevels()` (the Milestones tab's payout count) was already correct — it counts every level present in `levelModifiers`/`levelRewards` regardless of value — so the count vs. visible-slots mismatch the issue called out is now resolved by making the visible slots match.

## Changes

- `MilestonesEditor.svelte`: the level timeline now renders `0..maxLevel` (was `1..maxLevel`), and the selected-level clamp allows 0.
- The "Reward skill" field is now hidden while level 0 is selected. Unlike modifiers, level-0 **rewards** remain rejected server-side (`AdminProficiencies.SetRewards` requires `minLevel: 1` — a milestone reward is granted by crossing a level, and level 0 is never crossed), so surfacing that field at level 0 would let an author fill in a value the backend rejects on save. This mirrors the existing backend asymmetry rather than introducing a new one.
- The empty-payout copy at level 0 now reads "No on-open bonus — players start this tier with nothing." instead of the level-1+ "players just gain the level" wording, since level 0 isn't reached by leveling up.

## Test plan

- [x] Added `MilestonesEditor.test.ts` covering: level-0 timeline node renders and is selectable, the on-open empty-state copy, adding a payout at level 0, an existing level-0 modifier displays with the reward field omitted, and the reward field reappears at level 1+.
- [x] `npm run test:unit:ci` — full frontend suite green (3816/3816).
- Backend is unchanged (it already permitted level-0 modifiers correctly); no backend test changes needed.

Closes #2178

---
_Generated by [Claude Code](https://claude.ai/code/session_01AnQetaXoF8AtyPDbLrM69U)_

### DIFF
--- a/UI/src/routes/admin/workbench/progression/MilestonesEditor.svelte
+++ b/UI/src/routes/admin/workbench/progression/MilestonesEditor.svelte
@@ -67,17 +67,19 @@
 		{/each}
 		<button type="button" class="link-add" onclick={() => store.addModifier(tier.id, sel)}>+ Add modifier</button>
 
-		<div class="field-label mt">
-			Reward skill <span class="muted">(optional — granted on reaching this level)</span>
-		</div>
-		<div class="reward">
-			<ProgSelect
-				ariaLabel="Reward skill"
-				value={reward?.rewardSkillId ?? NO_SKILL}
-				options={rewardOptions}
-				onChange={(v) => store.setReward(tier.id, sel, v)}
-			/>
-		</div>
+		{#if sel > 0}
+			<div class="field-label mt">
+				Reward skill <span class="muted">(optional — granted on reaching this level)</span>
+			</div>
+			<div class="reward">
+				<ProgSelect
+					ariaLabel="Reward skill"
+					value={reward?.rewardSkillId ?? NO_SKILL}
+					options={rewardOptions}
+					onChange={(v) => store.setReward(tier.id, sel, v)}
+				/>
+			</div>
+		{/if}
 
 		<div class="payout-foot">
 			<button type="button" class="link-danger" onclick={() => store.removePayout(tier.id, sel)}
@@ -86,7 +88,11 @@
 		</div>
 	{:else}
 		<div class="empty-payout">
-			<div class="ep-text">No payout at level {sel} — players just gain the level.</div>
+			<div class="ep-text">
+				{sel === 0
+					? 'No on-open bonus — players start this tier with nothing.'
+					: `No payout at level ${sel} — players just gain the level.`}
+			</div>
 			<button
 				type="button"
 				class="btn-add"
@@ -115,7 +121,7 @@ interface Props {
 const { store, tier }: Props = $props();
 
 const maxLevel = $derived(Math.max(1, Math.floor(tier.maxLevel) || 1));
-const sel = $derived(Math.min(Math.max(1, store.selectedLevel), maxLevel));
+const sel = $derived(Math.min(Math.max(0, store.selectedLevel), maxLevel));
 const reward = $derived(rewardAtLevel(tier, sel));
 const selMods = $derived(modifiersAtLevel(tier, sel));
 const hasPayout = $derived(selMods.length > 0 || reward !== undefined);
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Render a level-0 slot in the Milestones editor for on-open modifiers (#2178)

**Verdict:** Approve — the editor now honors the full valid modifier range (`0..maxLevel`), and the level-0/reward asymmetry is handled correctly on both ends. Posted as a `COMMENT` because I'm the PR author (GitHub won't accept a gating self-approval); treat it as an approve.

The change squarely resolves the issue: `MilestonesEditor` rendered `1..maxLevel` and clamped selection with `Math.max(1, …)`, so a backend-valid level-0 modifier (`AdminProficiencies` `minLevel: 0`, `Proficiency.ModifiersForLevel`'s `l.Level <= level` on-open rule) was invisible and unremovable. Rendering the `0` node and clamping to `Math.max(0, …)` makes it selectable, editable, and removable, and the `payoutLevels()` count now matches the visible slots.

The level-0 **reward** handling is the subtle part and it's right: the field is hidden at `sel === 0` because `AdminProficiencies.SetRewards` rejects level < 1 (mirrored by `progression-helpers.ts`, which flags `reward.level < 1` but permits `modifier.level >= 0`). `addPayout → addModifier` seeds only a modifier at level 0, never a reward, so the editor can't author a value the backend would reject. Hiding rather than disabling the field is a defensible call that mirrors the existing backend asymmetry instead of introducing a new one, and the empty-state copy is correctly differentiated for the on-open case.

### Test coverage

`MilestonesEditor.test.ts` is a net-new file (the component previously had none) and covers the meaningful branches: the level-0 node renders and is selectable, the on-open empty-state copy, adding a payout at level 0, an existing level-0 modifier displaying with the reward field omitted, and the reward field reappearing at level 1+. The minimal `$stores` mock is sufficient — `attributeOptions`/`modifierTypeOptions` are pure `enumPairs(...)` and only `playerSkillOptions` touches the mocked `staticData.skills`. Backend is genuinely unchanged, so no backend test delta is warranted.

### What's good

Tight, well-scoped fix with clear reasoning in the code and PR body, and it picks the "make both ends agree" option cleanly rather than papering over the mismatch. No blocking, should-fix, or nit findings.

---

PR #2189, opened 2026-07-19
Stabilize SynthesisView.proficiencyLevels' reference across xp-only pushes

### LINKED ISSUE #2136: Synthesis screen re-derives and re-lays-out the entire recipe graph on every ProficiencyXpGained push
While the Synthesis screen is open, every `ProficiencyXpGained` push (one per battle victory) recomputes the whole recipe pipeline even when no proficiency **level** changed (the usual case — only xp moved):

- `UI/src/routes/game/screens/synthesis/synthesis-view.svelte.ts:42` — `proficiencyLevels = $derived(new Map(playerProficiencies.all.map(…)))`. The proficiencies store reassigns its array wholesale on every push (`UI/src/stores/proficiencies.svelte.ts:90-108` — deliberate, for by-reference change detection elsewhere), so each push produces a fresh `Map` identity.
- That re-runs `buildSynthesis` (fresh `recipes` array), which re-runs `layoutSynthesisGraph` (full longest-path layering + barycenter sort in `synthesis-graph.ts`), which re-derives `counts`/`visibleRecipes`/`selected` and, in `SynthesisGraph.svelte:70,74`, the `nodeByKey` map and `lines` array, re-rendering the `{#each layout.nodes}` graph.

`SynthesisGraph`'s recenter guard (keyed on the primitive layout extents, :96-98,163-168) already prevents the pan/zoom from being stomped — but not the churn itself.

Fix: make the levels input stable when levels haven't changed — e.g. compare the new levels against the previous map and keep the old reference, or derive off a levels fingerprint. (`ownedSkillIds` at :38 has the same shape but only churns on skill unlock, so it's fine.)

Cheap at today's recipe-catalogue size; pure wasted work per victory, worth fixing while the pattern is fresh.

ginderjeremiah:
## Summary

- `playerProficiencies.all` reassigns its array wholesale on every `ProficiencyXpGained` push (one per battle victory) — deliberate, for by-reference change detection elsewhere (proficiencies store docs).
- `SynthesisView.proficiencyLevels` (`UI/src/routes/game/screens/synthesis/synthesis-view.svelte.ts`) built a fresh `Map` from that array on every derivation, even when no proficiency **level** actually changed (the common case — only xp moved).
- Because `recipes` (→ `buildSynthesis`) and `graph` (→ `layoutSynthesisGraph`) key their own `$derived` recomputation on `proficiencyLevels`' reference identity, this re-ran the whole recipe pipeline and full graph layout (longest-path layering + barycenter sort) on every battle victory while the Synthesis screen was open — pure wasted work, per #2136.

## Fix

`proficiencyLevels` now memoizes against a plain (non-reactive, `#`-private) cache field: it returns the previous `Map` reference when every proficiency's level is unchanged, and only builds a new one when a level actually moved (or a proficiency newly opens). This mirrors the existing per-record `WeakMap` memoization pattern in `entity-store.svelte.ts`'s `recordStates` (writing to a non-reactive field mid-`$derived` is the established way to avoid an illegal `$state` mutation during derivation in this codebase — see `player-manager.ts`'s `#lockedBaseCache` for another instance).

Since Svelte 5's `$derived` chain only propagates invalidation when a dependency's *value* actually changes (not just when it's recomputed), keeping the same `Map` reference here means `recipes`/`graph` skip their own re-derivation entirely when only xp moved.

## Test plan

- Added `UI/src/tests/routes/game/screens/synthesis/synthesis-view-proficiency-stability.test.ts`, which — unlike the existing `synthesis-view.test.ts` — deliberately does **not** mock `$stores`, since the reference-stability bug only manifests through real `$state`-backed reactivity (a plain mocked object's property mutation isn't tracked by Svelte, so it could never expose this regression). Verified this test fails on the pre-fix code (via `git stash`) and passes after the fix.
  - Same-level push → `proficiencyLevels` keeps the same `Map` reference.
  - Level-up push → `proficiencyLevels` returns a new `Map` with the updated level.
  - Newly-opened proficiency → `proficiencyLevels` returns a new `Map` including it at level 0.
- [x] `npm run test:unit:ci` — 306 files / 3819 tests passed

Closes #2136

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vr4MFo5WqP95T8rsieJpAk)_

### DIFF
--- a/UI/src/routes/game/screens/synthesis/synthesis-view.svelte.ts
+++ b/UI/src/routes/game/screens/synthesis/synthesis-view.svelte.ts
@@ -37,9 +37,27 @@ export class SynthesisView {
 	// eslint-disable-next-line svelte/prefer-svelte-reactivity
 	readonly ownedSkillIds = $derived(new Set(playerManager.unlockedSkills.map((s) => s.skillId)));
 
-	/** The player's current level per proficiency (a missing proficiency counts as level 0 in gating). */
-	// eslint-disable-next-line svelte/prefer-svelte-reactivity
-	readonly proficiencyLevels = $derived(new Map(playerProficiencies.all.map((p) => [p.proficiencyId, p.level])));
+	/** Previous {@link proficiencyLevels} value, for the reference-stability check below. Plain (not
+	 *  `$state`) so writing it mid-derivation isn't a reactive mutation. */
+	#previousLevels: Map<number, number> | undefined;
+
+	/** The player's current level per proficiency (a missing proficiency counts as level 0 in gating).
+	 *  Keeps the previous Map's identity when every level is unchanged: `playerProficiencies.all`
+	 *  reassigns its array wholesale on every `ProficiencyXpGained` push (deliberately, for by-reference
+	 *  change detection elsewhere) even when xp moved but no level did, and this is the value {@link
+	 *  recipes} and {@link graph} key their own recomputation on — without this, the whole recipe list
+	 *  and graph layout would re-derive on every battle victory for no visible change. */
+	readonly proficiencyLevels = $derived.by(() => {
+		const all = playerProficiencies.all;
+		const previous = this.#previousLevels;
+		if (previous && previous.size === all.length && all.every((p) => previous.get(p.proficiencyId) === p.level)) {
+			return previous;
+		}
+		// eslint-disable-next-line svelte/prefer-svelte-reactivity
+		const next = new Map(all.map((p) => [p.proficiencyId, p.level]));
+		this.#previousLevels = next;
+		return next;
+	});
 
 	/** The discovered recipes with their reveal/gating state — recomputed as owned skills, proficiency
 	 *  levels, and reference data change. */
--- a/UI/src/tests/routes/game/screens/synthesis/synthesis-view-proficiency-stability.test.ts
+++ b/UI/src/tests/routes/game/screens/synthesis/synthesis-view-proficiency-stability.test.ts
@@ -1,0 +1,74 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+/* Unlike synthesis-view.test.ts, this file deliberately does NOT mock $stores: proficiencyLevels'
+   reference-stability fix (#2136) only exercises Svelte's derived-chain memoization when its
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: reference-stabilize `SynthesisView.proficiencyLevels` so xp-only pushes don't re-derive the recipe pipeline (#2136)

**Verdict:** Approve _(submitted as a comment since I'm the PR author — GitHub won't accept a gating self-review)_ — the fix cleanly solves #2136, follows an established codebase pattern, and is well-tested.

The change makes `proficiencyLevels` keep its previous `Map` identity when no proficiency **level** actually moved, so the `recipes`/`graph` `$derived` chain — which keys its recomputation on that reference — skips the full recipe build + graph layout on every `ProficiencyXpGained` push. That's exactly the wasted work #2136 describes.

I verified the memoization guard is correct in all cases:
- `previous.size === all.length && all.every(p => previous.get(p.proficiencyId) === p.level)` — since proficiency ids are unique within `all`, equal size plus every current id present at the same level means the maps are identical (a bijection). A missing/renamed id gives `undefined === <number>` → rebuild. A pure reorder with the same id→level set correctly keeps the reference (consumers key by id, so order is irrelevant). Empty→empty pushes also stay stable.
- Writing the `#`-private, non-reactive `#previousLevels` mid-derivation mirrors `player-manager.ts`'s `#lockedBaseCache` and is the documented way here to avoid an illegal `$state` mutation during derivation.

### Test coverage
Solid. The new `synthesis-view-proficiency-stability.test.ts` mirrors the source path and covers all three branches: same-level push keeps the ref, a level-up returns a new map, and a newly-opened proficiency returns a new map at level 0. Deliberately not mocking `$stores` is the right call — the reference-stability behavior only manifests through real `$state`-backed reactivity, and the file comment explains why.

### What's good
- Tight, well-scoped fix that reuses the existing non-reactive-cache pattern rather than inventing a new mechanism.
- The doc comment captures the non-obvious "why" (wholesale array reassignment + downstream reference-keyed derivation) that a future edit could otherwise naively "simplify" and regress.

No blocking, should-fix, or nit findings. Once CI is green I'll merge.

---

PR #2190, opened 2026-07-19
Memoize ProgressionStore per-record status to stop full-catalogue re-diffs

### LINKED ISSUE #2137: ProgressionStore re-diffs both whole catalogues (canonicalEqual per record) on every keystroke — #2085's per-record memoization never applied here
`UI/src/routes/admin/workbench/progression/progression-store.svelte.ts:107-113` — `pathDiff`/`profDiff` run `canonicalEqual` (canonicalize + double `JSON.stringify`) for **every** record against its baseline on any single-record patch, and the per-row `pathStatus`/`profStatus` reads in `ProgressionList.svelte:97-114` re-run it per row. This is exactly the pattern #2085 eliminated for the generic Workbench `EntityStore` with per-record WeakMap memoization — the progression store predates that fix and never picked it up.

Cheap at today's catalogue sizes, so no urgency — but it's a parity/robustness gap with the generic store, and catalogue growth is the plan. Apply the same per-record memoization (or share the mechanism outright — related to the broader Workbench/progression unification in #2125, though that issue is about the list-row UI primitives, not the diff layer).

ginderjeremiah:
## Summary

- `ProgressionStore.pathDiff`/`profDiff` ran `canonicalEqual` (canonicalize + double `JSON.stringify`) against baseline for **every** record in the catalogue on any single-record patch, and `pathStatus`/`profStatus` repeated that same work per row read (`ProgressionList.svelte`'s per-row `status:` binding). This is the exact pattern #2085 eliminated for the generic Workbench `EntityStore` via per-record `WeakMap` memoization — `ProgressionStore` predates that fix and never picked it up.
- Adopted the same per-record memoization: `patchPath`/`patchProf` only ever replace the one record they touch with a new object reference, so every other record's status is now a same-reference cache hit in a `WeakMap` instead of a re-canonicalized comparison against its baseline. `pathDiff`/`profDiff` (used by `counts`) are now derived from these memoized statuses instead of independently re-diffing the whole catalogue.
- Unlike `EntityStore`, no epoch-based cache invalidation was needed — `ProgressionStore` has no soft-delete side channel that can flip a record's status without also replacing its object reference (retire is an ordinary field patch, `removePath`/`removeTier` just filter the record out).
- `diffCatalogue`/`CatalogueDiff` (`progression-helpers.ts`) became dead code once their only call site was replaced — removed them along with their direct unit test, per CLAUDE.md's dead-code guideline.

## Test plan

- [x] Added a unit test asserting that patching one path only triggers one `canonicalEqual` call (for the patched record), while an untouched sibling record hits the cache — verifying the memoization actually works, not just that statuses stay correct.
- [x] `npx vitest run` — full UI suite: 306 files / 3819 tests passing.
- [x] `npx eslint .` — clean.
- [x] `npx svelte-check` — 0 errors / 0 warnings.

Closes #2137

---
_Generated by [Claude Code](https://claude.ai/code/session_01CAwcww79USVg3cHFvi3m6G)_

### DIFF
--- a/UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
+++ b/UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
@@ -14,7 +14,6 @@ import { childChanged, canonicalEqual, resolveId, resolveNewIds } from '../save-
 import { firstFree } from '../entities/helpers';
 import {
 	blankModifier,
-	diffCatalogue,
 	newPath,
 	newProficiency,
 	pathIdentityDto,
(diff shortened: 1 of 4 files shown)


UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
```diff
@@ -121,20 +180,14 @@ export class ProgressionStore {
...
 	}
 
+	/** Always called with a record from {@link paths}, which {@link pathStatuses} covers exhaustively. */
 	pathStatus(path: WorkbenchPath): RecordStatus {
-		const baseline = this.basePathMap[path.id];
-		if (!baseline) {
-			return 'added';
-		}
-		return canonicalEqual(path, baseline) ? 'clean' : 'modified';
+		return this.pathStatuses[path.id];
```
ginderjeremiah, line 185: Nit (type-honesty / parity): `pathStatuses` is a `Record<number, RecordStatus>`, so `this.pathStatuses[path.id]` is `RecordStatus | undefined` at runtime — a record whose `id` isn't in the map yields `undefined` while the signature promises `RecordStatus`. The documented "always called with a record from `paths`" invariant holds for every current call site, so this is safe today, but it's a small divergence from the sibling this PR mirrors: `EntityStore.stateOf` falls back to a self-computed status (`?? { status: this.status(record), … }`) precisely so a detached record can't leak an undefined. Same applies to `profStatus` just below. Consider mirroring that fallback (or, cheaper, just narrowing the return type to `RecordStatus | undefined` so the honesty is visible) — not blocking either way.

---

ginderjeremiah (commented):
## Review: Memoize `ProgressionStore` per-record status (closes #2137)

**Verdict:** Approve — correctly ports #2085's per-record WeakMap memoization to `ProgressionStore`, with one non-blocking nit. (Submitted as a comment since I'm the PR author; GitHub won't accept a gating verdict on one's own PR.)

This is a clean, well-scoped fix that resolves the parity gap the issue describes: `pathDiff`/`profDiff` and the per-row `pathStatus`/`profStatus` reads no longer re-`canonicalEqual` the whole catalogue on every keystroke — each untouched record is now a same-reference `WeakMap` hit, exactly as `EntityStore.recordStates` does.

### Correctness — verified
- **No epoch invalidation needed here.** I checked every mutator: `patchPath`/`patchProf`, `addPath`/`addTier`, `reorderTiers`, `retire*`, `reset*`, and the milestone/gateway setters all replace the affected record with a fresh object (WeakMap cache miss on their own), and `removePath`/`removeTier` filter the record out of the array. There's no soft-delete side channel like `EntityStore`'s `deleted` set that flips a status without swapping the reference, so the PR's "no epoch" claim holds. `reorderTiers` keeps a reference only for tiers outside the reordered path, whose status is genuinely unchanged.
- **Baseline/working-set stay in lockstep.** `basePaths` is only reassigned in `setData`, which also reassigns `paths` (fresh clones), so a record reference never outlives a baseline change — no stale-cache-across-reseed hazard, and the WeakMap self-GCs superseded objects.
- **`save()` reading `this.pathDiff`/`this.profDiff`** is equivalent to the old `diffCatalogue(...)` recompute (same baseline map, same added/modified split) and `$derived` reads are always current at access time.

### Dead code
`diffCatalogue`/`CatalogueDiff` and their direct unit test are removed with no dangling references — confirmed via grep. Correct per CLAUDE.md's dead-code guideline.

### Test coverage
Good. The new test drives the real store and asserts `canonicalEqual` fires exactly once (for the patched record) while the untouched sibling hits the cache — it verifies the memoization mechanism, not just that statuses stay correct, and the existing suite already covers status correctness through edits/retire/reset/discard. No battle-parity surface is touched.

### Nits
- `pathStatus`/`profStatus` can return `undefined` at runtime (record id absent from the memo map) while typed `RecordStatus`; the sibling `EntityStore.stateOf` guards this with a self-computed fallback. Safe today under the documented invariant — details inline. Non-blocking.

### What's good
Faithful port of an established pattern, with a genuinely explanatory comment on *why* the epoch machinery isn't carried over — that context is the easy thing to lose and it's captured well.

---

UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
ginderjeremiah: Good catch, but leaving this as-is deliberately rather than adding either option. Widening the return type to `RecordStatus | undefined` would just push `undefined`-handling onto every call site (`ListRow`'s `status` prop, the `=== 'modified'` checks in `PathDetail`/`TierDetail`) for a case that structurally cannot occur — every call site passes a record sourced directly from `store.paths`/`store.profs`. And mirroring `EntityStore.stateOf`'s fallback would be dead code by the same logic: `stateOf` exists because it's documented as usable for detached records, but `pathStatus`/`profStatus` have no such use case today.

CLAUDE.md is explicit here: "Don't add error handling, fallbacks, or validation for scenarios that can't happen. Trust internal code and framework guarantees." The doc comment on each method already states the invariant it's trusting. Happy to add the fallback if a real caller ever needs to look up a detached record's status, but until then it'd be unreachable code.

---

PR #2191, opened 2026-07-19
Dedupe post-#2124 scss residue: detail-pane block and ops-console chrome

### LINKED ISSUE #2138: Post-#2124 scss residue: detail-pane block duplicated across PathDetail/TierDetail/WorkbenchDetail; ops consoles re-declare the same chrome
Two scss duplication clusters left after #2124's shared-chrome extraction (distinct from #2125, which covers the list-row primitives, and from #2061's attribute-table/save-flash items):

1. **Detail-pane block.** `UI/src/routes/admin/workbench/progression/PathDetail.svelte:117-153` and `TierDetail.svelte:142-178` carry character-identical `.detail-body`/`.body-inner`/`.sec-title` blocks, near-identical to `workbench/components/WorkbenchDetail.svelte:166-178` (only a locked-opacity value differs). All render under the shared `.workbench` scope — the block belongs in `workbench.scss`.

2. **Ops-console chrome.** `UI/src/routes/admin/ops/DeadLetterConsole.svelte:246-344` and `ContentHealthConsole.svelte` each re-declare ~100 lines of the same visual system (eyebrow/title-row/summary/toolbar/`.btn`/error-panel/empty-state). A third ops tool would copy it again — extract a shared ops-console partial or wrapper component.

ginderjeremiah:
## Summary

Two scss duplication clusters left after #2124's shared-chrome extraction, as described in #2138.

**1. Detail-pane block** (`.detail-body`/`.body-inner`/`.sec-title`)

`PathDetail.svelte` and `TierDetail.svelte` carried locally-defined `.detail-body`/`.body-inner`/`.sec-title` blocks, near-identical to `WorkbenchDetail.svelte`'s (only a locked-opacity value and which element the `locked` class was toggled on differed). `workbench.scss` already carried an identical `.sec-title` — so `PathDetail`/`TierDetail`'s local copy was pure redundant drift (their local margin-bottom had also silently drifted to 18px vs the shared 12px).

- Moved `.detail-body`/`.body-inner` into `workbench.scss`, canonicalized on `WorkbenchDetail`'s values (0.5 lock opacity, `locked` toggled on `.body-inner`) since the progression editor originally copied its chrome from `WorkbenchDetail` (per #2124's PR description).
- Removed the now-fully-redundant local `.detail-body`/`.body-inner`/`.sec-title` blocks from all three components.
- Updated `PathDetail`/`TierDetail`'s markup to toggle `locked` on `.body-inner`, matching `WorkbenchDetail`'s pattern.

**2. Ops-console chrome**

`DeadLetterConsole.svelte` and `ContentHealthConsole.svelte` each re-declared ~100 lines of identical visual system (eyebrow/title-row/summary/blurb/toolbar/`.btn`/error-panel/empty-state).

- Extracted a new `ops-console.scss` partial (mirroring the existing `workbench.scss` pattern: a namespaced `.ops-console` scope imported as global CSS) covering the shared chrome.
- Updated both consoles' markup to use the shared class names, keeping only genuinely console-specific bits local (summary's state-color spans, blurb `max-width` — the two consoles' copy differs in length so their `max-width` legitimately differs, and Content Health's healthy-state checkmark keeps its accent-colored glyph instead of the shared muted default).

No functional or user-facing behavior changes intended — this is pure scss/markup consolidation. The two numeric value drifts found during consolidation (18px vs 12px section-title margin, 0.55 vs 0.5 lock opacity) are now unified on the values already used by the generic `WorkbenchDetail`/`workbench.scss` (the copied-from original), a negligible cosmetic delta.

## Test plan

- [x] `npm run lint` (eslint + svelte-check) — clean, 0 errors/warnings
- [x] `npm run test:unit:ci` — 306 test files / 3819 tests pass
- [ ] Manual visual pass in a browser was not performed (no backend/auth stack running in this environment) — the CSS moves were verified rule-by-rule for equivalence against the original scoped styles, and the removed component tests all still assert via `data-testid`, not CSS class names, so none needed updating.

Closes #2138

---
_Generated by [Claude Code](https://claude.ai/code/session_01TNbDG4Hso6FWAFmjaY58nu)_

### DIFF
--- a/UI/src/routes/admin/ops/ContentHealthConsole.svelte
+++ b/UI/src/routes/admin/ops/ContentHealthConsole.svelte
@@ -1,5 +1,5 @@
-<div class="ch-console" data-testid="content-health-console">
-	<div class="ch-head">
+<div class="ops-console" data-testid="content-health-console">
+	<div class="head">
 		<div class="eyebrow">Admin Console · Ops</div>
 		<div class="title-row">
 			<h1 class="title" data-testid="ch-title">Content Health</h1>
@@ -30,21 +30,21 @@
 		</p>
 	</div>
 
-	<div class="ch-toolbar">
+	<div class="toolbar">
 		<button type="button" class="btn" data-testid="ch-refresh" disabled={health.loading} onclick={refresh}>
 			Refresh
 		</button>
 	</div>
 
 	{#if health.error}
-		<div class="ch-error" role="alert" data-testid="ch-error">{health.error}</div>
+		<div class="error-panel" role="alert" data-testid="ch-error">{health.error}</div>
 	{/if}
 
-	<div class="ch-body">
+	<div class="body">
 		{#if !health.loaded && health.loading}
 			<Loading loading={true} delay={150} />
 		{:else if health.isHealthy}
-			<div class="ch-empty" data-testid="ch-healthy">
+			<div class="empty-state" data-testid="ch-healthy">
 				<div class="glyph">
 					<svg
 						width="26"
@@ -97,6 +97,7 @@ import { onMount } from 'svelte';
 import Loading from '$components/Loading.svelte';
 import { toastError } from '$stores';
 import { staticData } from '$stores';
+import './ops-console.scss';
 import { ContentHealthState, entityDisplayName, severityMeta, type EntityNameSources } from './content-health.svelte';
 
 const health = new ContentHealthState();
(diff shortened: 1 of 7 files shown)


ginderjeremiah (commented):
## Review: SCSS dedupe of the detail-pane block and ops-console chrome (closes #2138)

**Verdict:** Approve _(posted as a comment since I'm the PR author; GitHub won't accept a formal approve on your own PR)_ — a clean, faithful consolidation with no functional changes and no blocking or should-fix findings.

I verified the two consolidations against the codebase rather than trusting the description:

**Detail-pane block**
- `workbench.scss` already carries `.sec-title` (line 593) under the `.workbench` scope, so removing the local copies from `PathDetail`/`TierDetail` is safe — the shared `.sub` sets `color: var(--text-muted)` explicitly where the local copy inherited it, so the effective color is unchanged (the only real deltas are the acknowledged 18px→12px margin and 0.55→0.5 lock opacity, both negligible).
- The moved `.detail-body`/`.body-inner` rules land under `.workbench`, and `Progression.svelte`'s root is `class="prog workbench"` with `workbench.scss` imported globally there and in `Workbench.svelte`, so the rules reach `PathDetail`/`TierDetail`. Moving the `locked` toggle from `.detail-body` to `.body-inner` matches `WorkbenchDetail`'s existing pattern — good call unifying on the canonical form.

**Ops-console chrome**
- The new `ops-console.scss` mirrors the established `workbench.scss` pattern (namespaced `.ops-console` scope imported as global CSS), which is the right precedent to follow here.
- Both consoles' markup fully migrated to the shared class names; no dangling references to the removed `.ch-*`/`.dl-*` wrapper classes remain, and the per-console leftovers (`.summary` state-color spans, per-console `.blurb` max-width, Content Health's accent checkmark) are legitimately console-specific.

One thing I checked closely and cleared: Content Health's `.empty-state .glyph { color: var(--accent) }` override now competes with the global `.ops-console .empty-state .glyph` at equal specificity. It resolves in favor of the scoped override because this codebase orders component styles after the global import at equal specificity — the same tiebreak that let `PathDetail`'s local `.sec-title` (18px) win over the global 12px pre-PR, as the description notes. So the healthy checkmark keeps its accent color. Worth keeping in mind as a latent source-order dependency, but it's the codebase's established scoped-over-global pattern, not something to change here.

### Test coverage
Appropriate for a pure SCSS/markup move — no logic changed, and the component tests assert via `data-testid` (unaffected by the class renames). `verify-frontend` (svelte-check) passing confirms no unused/orphaned selectors were left behind.

### What's good
Genuinely tidy extraction: canonicalizing on the copied-from originals, documenting the two value drifts up front, and keeping only the truly console-specific bits local. Sets up the next ops tool to reuse the chrome instead of copying it a third time.

I'll merge once CI goes green (frontend unit + e2e are still running).

---

PR #2192, opened 2026-07-19
Cap CombatRating's DoT term duration at the shared ramp horizon

### LINKED ISSUE #2049: CombatRating's DoT term ignores the shared-expiry ramp its own attribute-effect term models
## Context

Found in the July 2026 health review (Game.Core pass).

`CombatRating` prices an opponent-DoT effect as `ampedMagnitude × durationSec / effectiveCooldownSec`, uncapped (`Game.Core/Battle/CombatRating.cs:183-197`). But a DoT accumulator effect rides the same `ApplyEffect` stacking machinery as every other effect (`Battler.cs:615-663` — magnitudes sum, expiry resets on every re-application): re-casting a bleed whose `DurationMs` exceeds its skill's effective cooldown refreshes the shared expiry every cast, so the accumulator **ramps** for as long as the skill cycles. The "exact steady-state `magnitude × d/cd`" closed form (spike #1526 Decision 1) is only exact for `d ≤ cd`.

The inconsistency is within the same file: `FoldedEffectMagnitude` (`CombatRating.cs:412-434`) explicitly models this ramp for attribute buffs/debuffs (`min(DurationMs, RefFightDuration·1000/2) ÷ effectiveCooldownMs`, with a derivation comment) — the DoT term doesn't.

**Cost:** XP bounties and proficiency rating-denominators mis-price ramping-DoT builds/enemies. Example: bleed with `cd = 5s`, `d = 8s` — code prices the average accumulator at `1.6×` magnitude; the real average over the 30s reference fight is ~`3×`. Server-only; no battle-sim or parity impact.

## Suggested fix

Cap the DoT term's duration the same way: `min(durationSec, RefFightDuration / 2) / effectiveCooldownSec`, and re-run the combat-rating calibration report (#1533 tooling) to check whether any authored content's rating moves materially.

ginderjeremiah:
## Summary

`CombatRating`'s opponent-DoT term priced an effect as `ampedMagnitude × durationSec / effectiveCooldownSec`, uncapped (`Game.Core/Battle/CombatRating.cs`). But a DoT accumulator rides the same `ApplyEffect` stacking machinery as every other effect (magnitudes sum, expiry resets on every re-application), so a bleed/DoT whose `DurationMs` exceeds its skill's effective cooldown refreshes the shared expiry every cast and **ramps** for as long as the skill cycles. The "exact steady-state `magnitude × d/cd`" closed form (spike #1526 Decision 1) is only exact for `d ≤ cd`.

`FoldedEffectMagnitude` in the same file already models this ramp for Self-targeted attribute buffs/debuffs (`min(DurationMs, RefFightDuration/2) ÷ effectiveCooldownMs`) — the opponent-DoT term didn't apply the same cap.

## Fix

Cap the DoT term's duration the same way `FoldedEffectMagnitude` does: `min(durationSec, RefFightDuration / 2) / effectiveCooldownSec`.

No authored content currently ships a DoT skill (`content/skills.json` has 4 seeded skills, none with effects), so this has zero effect on today's rating output — it's a correctness fix ahead of when DoT skills get authored, so XP bounties/proficiency accrual price ramping-DoT content correctly from day one instead of needing a retroactive calibration pass.

## Changes

- `Game.Core/Battle/CombatRating.cs` — cap the DoT term's `durationSec` at `RefFightDuration / 2`, matching `FoldedEffectMagnitude`'s ramp horizon; updated the term's comment to explain why.
- `Game.Core.Tests/Battle/CombatRatingTests.cs`:
  - New test `Rate_DoTEffectDurationBeyondRampHorizon_RatesTheSameAsExactlyAtTheHorizon`, which fails on the old (uncapped) code and passes with the fix.
  - Extracted a `MakeDotEffect` helper and refactored the two existing DoT tests to use it (DRY).

## Test plan

- [x] `dotnet build` — succeeds, 0 warnings
- [x] `dotnet test` (full solution, using the session's Postgres/Redis containers) — 3345/3345 passing, 0 failures
- [x] Verified the new test fails against the pre-fix code (manually reverted the source change, confirmed the test catches the regression, then restored the fix)
- [x] Confirmed no seeded content is affected (no DoT skills currently authored) — this is a forward-looking correctness fix, not a content-migrating change

Closes #2049

---
_Generated by [Claude Code](https://claude.ai/code/session_01CA8RiqT6xWReNjC383mRk2)_

### DIFF
--- a/Game.Core.Tests/Battle/CombatRatingTests.cs
+++ b/Game.Core.Tests/Battle/CombatRatingTests.cs
@@ -142,21 +142,28 @@ namespace Game.Core.Tests.Battle
         [Fact]
         public void Rate_DoTEffect_ContributesOffenseEvenWithZeroBaseDamage()
         {
-            var dotEffect = new SkillEffect
-            {
-                Id = 1,
-                Target = ESkillEffectTarget.Opponent,
-                AttributeId = BleedDamagePerSecond,
-                ModifierType = EModifierType.Additive,
-                Amount = 10,
-                DurationMs = 5000,
-                ScalingAttributeId = Strength,
-                ScalingAmount = 0,
-            };
+            var dotEffect = MakeDotEffect(amount: 10, durationMs: 5000);
             var noDot = MakeBattlerWithSkills([], [MakeSkill(cooldownMs: 1000, baseDamage: 0)]);
             var withDot = MakeBattlerWithSkills([], [MakeSkill(cooldownMs: 1000, baseDamage: 0, effects: [dotEffect])]);
 
             Assert.True(CombatRating.Rate(withDot, isPlayer: true) > CombatRating.Rate(noDot, isPlayer: true));
+        }
+
+        [Fact]
+        public void Rate_DoTEffectDurationBeyondRampHorizon_RatesTheSameAsExactlyAtTheHorizon()
+        {
+            // A DoT accumulator rides the same shared-expiry stacking machinery as a Self-targeted effect
+            // (Battler.ApplyEffect), so once duration clears FoldedEffectMagnitude's RefFightDuration/2 ramp
+            // horizon, further duration must not move the rating at all — pins that the DoT term's duration is
+            // truly capped rather than left to grow unboundedly for a duration exceeding its skill's cooldown.
+            var atHorizon = MakeSkill(cooldownMs: 1000, baseDamage: 0, effects:
+                [MakeDotEffect(amount: 10, durationMs: (int)(ServerGameConstants.RefFightDuration * 1000 / 2))]);
+            var wayPastHorizon = MakeSkill(cooldownMs: 1000, baseDamage: 0, effects:
+                [MakeDotEffect(amount: 10, durationMs: 1_000_000)]);
+
+            Assert.Equal(
+                CombatRating.Rate(MakeBattlerWithSkills([], [atHorizon]), isPlayer: true),
+                CombatRating.Rate(MakeBattlerWithSkills([], [wayPastHorizon]), isPlayer: true), 6);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Cap CombatRating's opponent-DoT term at the shared ramp horizon (closes #2049)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the fix faithfully mirrors the already-established, documented pattern in the same file and is backed by a regression test that pins the cap boundary.

### Assessment

The change caps the opponent-DoT term's `durationSec` at `RefFightDuration / 2`, exactly the horizon `FoldedEffectMagnitude` (`Game.Core/Battle/CombatRating.cs:430`) already applies to Self-targeted effect uptime. I verified the two are equivalent — `FoldedEffectMagnitude` works in ms (`Math.Min(effect.DurationMs, RefFightDuration * 1000.0 / 2.0)`), the DoT term in seconds (`Math.Min(effect.DurationMs / 1000.0, RefFightDuration / 2.0)`) — same horizon. Both DoT-adjacent reads (`magnitude` scaling, the cap) now agree with the sibling term, and the scaling attribute is correctly read off the original `battler`, consistent with the comment.

This resolves the within-file inconsistency the issue identified: the DoT accumulator rides the same `Battler.ApplyEffect` shared-expiry stacking machinery, so pricing it with an uncapped `d/cd` over-counts a bleed whose duration exceeds its cooldown. The fix brings it in line with the accepted spike #1526 model rather than introducing a new one — the right call.

### Test coverage

- `Rate_DoTEffectDurationBeyondRampHorizon_RatesTheSameAsExactlyAtTheHorizon` is the correct shape: it pins that duration past the horizon does *not* move the rating (equal to exactly at the horizon), which fails on the uncapped code and passes with the cap. Good regression guard.
- The `MakeDotEffect` helper extraction is a clean DRY refactor of the two pre-existing DoT tests, with sensible defaults; no behavior change to those tests.
- CombatRating is server-only with no frontend mirror (per `docs/backend.md`), so no battle-logic parity test is required — correctly noted in the PR.
- No authored content ships a DoT skill today, so rating output is unchanged — this is a forward-looking correctness fix, which addresses the issue's "re-run calibration" suggestion (nothing to move).

---

PR #2193, opened 2026-07-19
Game.Core micro-cleanups: avoidable null-forgiving operators; split Invalidate() out of SetCachedValue

### LINKED ISSUE #2139: Game.Core micro-cleanups: avoidable null-forgiving operators; split Invalidate() out of AttributeCollectionNode.SetCachedValue
Small clarity/guideline items in Game.Core, bundled:

1. **`Proficiency.cs:113`** — `l.RewardSkillId!.Value` is trivially avoidable (e.g. an `is int id` pattern in the filter/select). The guideline says to avoid `!`, and the codebase elsewhere goes out of its way to (see `AttributeCollectionNode`'s comparer comment).
2. **`Collections/SortedLinkedList.cs:99,101`** — `_current!.Value` / `Current!` in the struct enumerator. This is the conventional BCL enumerator pattern, but they're the only other `!` usages in the project — either rework or leave with a brief justifying comment so the guideline exception is deliberate.
3. **`Attributes/AttributeCollectionNode.cs:42-57`** — `SetCachedValue` runs the derived-invalidation cascade (walking `DerivedNodes`, nulling dependents' caches) even when called from `AttributeCollection.GetAttributeValue:97` to cache a **freshly computed** value — at which point every dependent is already invalid (invalidation via `AddModifier`/`RemoveModifier` cascades recursively, and a cached dependent always implies cached sources). Splitting an explicit `Invalidate()` from `SetCachedValue(double)` makes the cascade's intent unambiguous and drops the redundant walk on the mid-battle recompute path. Behavior today is correct; this is clarity plus a micro-perf win.

ginderjeremiah:
## Summary

Three small clarity/guideline items in `Game.Core`, bundled per the issue:

1. **`Proficiency.RewardSkillsCrossed`** (`Game.Core/Proficiencies/Proficiency.cs`) — replaced the `l.RewardSkillId!.Value` null-forgiving unwrap with `.Select(l => l.RewardSkillId).OfType<int>()`, which filters out the `null` entries and unwraps the survivors without a manual `!`.
2. **`SortedLinkedList<T>.Enumerator`** (`Game.Core/Collections/SortedLinkedList.cs`) — left `_current!.Value` / `Current!` as-is (the conventional BCL enumerator pattern — `Current` is undefined before the first `MoveNext`/after it returns `false`), but added a short comment so the guideline exception reads as deliberate rather than an oversight, matching the style of `AttributeCollectionNode`'s existing comparer comment.
3. **`AttributeCollectionNode.SetCachedValue`** (`Game.Core/Attributes/AttributeCollectionNode.cs`) — split into two methods:
   - `SetCachedValue(double value)` — just caches a freshly computed value, no cascade.
   - `Invalidate()` — clears the cached value and cascades invalidation to derived nodes (the old combined behavior).

   `AttributeCollection.GetAttributeValue`'s cache-fill path (a freshly computed value, where every dependent is already invalid by construction) now calls `SetCachedValue`, dropping the redundant `DerivedNodes` walk on that mid-battle recompute path. `AddModifier`/`RemoveModifier` now call `Invalidate()` explicitly where they previously called `SetCachedValue(null)`.

No behavior change for items 1 and 3 (item 3 is behavior-preserving plus a micro-perf win on the recompute path); item 2 is comment-only.

## Test plan

- [x] `dotnet test` (full solution: `Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Application.Tests`, `Game.Api.Tests`) — 3345/3345 passing

Closes #2139

---
_Generated by [Claude Code](https://claude.ai/code/session_012akpyqZNvk98aDRwbPSFP1)_

### DIFF
--- a/Game.Core/Attributes/AttributeCollection.cs
+++ b/Game.Core/Attributes/AttributeCollection.cs
@@ -37,7 +37,7 @@ namespace Game.Core.Attributes
         public void AddModifier(AttributeModifier modifier)
         {
             AddModifierWithoutCacheInvalidation(modifier);
-            GetOrCreateNode((int)modifier.Attribute).SetCachedValue(null);
+            GetOrCreateNode((int)modifier.Attribute).Invalidate();
         }
 
         /// <summary>
@@ -65,7 +65,7 @@ namespace Game.Core.Attributes
                 UnhookDerivedLink(node, modifier.DerivedSource);
             }
 
-            node.SetCachedValue(null);
+            node.Invalidate();
             return true;
         }
 
--- a/Game.Core/Attributes/AttributeCollectionNode.cs
+++ b/Game.Core/Attributes/AttributeCollectionNode.cs
@@ -39,9 +39,19 @@ namespace Game.Core.Attributes
             return DerivedNodes ??= [];
         }
 
-        public void SetCachedValue(double? value)
+        // Caches a freshly computed value without cascading invalidation — every dependent is
+        // already invalid at this point (a cached dependent always implies cached sources), so
+        // walking DerivedNodes here would be a redundant no-op.
+        public void SetCachedValue(double value)
         {
             CachedValue = value;
+        }
+
+        // Clears this node's cached value and cascades to every derived node whose cache
+        // depended on it. Call this whenever a modifier is added or removed.
+        public void Invalidate()
+        {
+            CachedValue = null;
             if (DerivedNodes is null)
             {
                 return;
@@ -51,7 +61,7 @@ namespace Game.Core.Attributes
             {
                 if (derivedNode.CachedValue is not null)
                 {
-                    derivedNode.SetCachedValue(null);
+                    derivedNode.Invalidate();
                 }
             }
         }
--- a/Game.Core/Collections/SortedLinkedList.cs
+++ b/Game.Core/Collections/SortedLinkedList.cs
@@ -96,6 +96,9 @@ namespace Game.Core.Collections
                 _head = head;
             }
 
+            // Undefined before the first MoveNext or after it returns false, matching the BCL
+            // enumerator contract (e.g. List<T>.Enumerator) — the ! here is deliberate, not an
+            // oversight, for the same reason the BCL's own enumerators use it.
             public readonly T Current => _current!.Value;
 
             readonly object IEnumerator.Current => Current!;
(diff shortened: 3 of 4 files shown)


ginderjeremiah (commented):
## Review: Game.Core micro-cleanups (#2139) — avoidable `!` operators + split `Invalidate()` from `SetCachedValue`

**Verdict:** Approve — all three items are correct, behavior-preserving, well-documented, and covered by existing regression tests. (Submitted as a comment since I'm the PR author; GitHub won't accept a gating verdict on one's own PR.)

I reviewed each item against `CLAUDE.md`'s guidelines (avoid `!`, DRY, no dead code) and the DDD layering in `docs/backend.md` — all changes stay within `Game.Core` domain logic.

### Item 1 — `Proficiency.RewardSkillsCrossed`
`[.. Where(...).Select(l => l.RewardSkillId).OfType<int>()]` is the right idiom: `OfType<int>()` over `IEnumerable<int?>` drops the nulls and unwraps the survivors, removing the `!` without a manual filter. Behavior is identical to the old `Where(... is not null).Select(!.Value)`, and it's guarded by `RewardSkillsCrossed_ReturnsOnlyTheRewardSkillsOfCrossedMilestones`.

### Item 2 — `SortedLinkedList.Enumerator`
Comment-only, exactly as the issue requested — the BCL enumerator contract (`Current` undefined before first `MoveNext`/after `false`) is the standard justification for the `!`, and documenting it makes the guideline exception deliberate rather than an oversight. Consistent with `AttributeCollectionNode`'s existing comparer comment.

### Item 3 — split `Invalidate()` out of `SetCachedValue`
The correctness argument holds. `GetAttributeValue`'s cache-fill path reaches `SetCachedValue(amount)` only when `node.CachedValue` was just observed `null`. By the invariant that a cached dependent always implies cached sources (computing a derived node forces its sources to cache first via `Apply`), every dependent of a currently-null node is itself already null — so the old cascade walk on that path was a proven no-op, and dropping it is safe as well as a small win on the mid-battle recompute path. `AddModifier`/`RemoveModifier` correctly retain the cascade via the explicit `Invalidate()`. All three `SetCachedValue`/`Invalidate` call sites are updated; no `SetCachedValue(null)` remains. Guarded by `AddModifier_InvalidatesCachedValue` and `RemoveModifier_CascadesInvalidationToDerivedAttributes`.

### Test coverage
Appropriate. Items 1 & 3 are behavior-preserving refactors and item 2 is comment-only, so the existing tests that cover these exact paths are the correct regression guard — no new tests are warranted here.

### What's good
Tight, well-scoped tech-debt PR. Each change is justified in a concise code comment, the perf micro-win on item 3 is a nice bonus, and the reasoning in the PR description matches what the code actually does.

No blocking, should-fix, or nit findings. Merging once CI is green.

---

PR #2194, opened 2026-07-19
Detect prerequisite cycles composed with implicit within-path tier ordering

### LINKED ISSUE #2144: Prerequisite cycle detection ignores implicit within-path tier ordering — a cross-path cycle composed with it still deadlocks silently

ginderjeremiah:
## Summary

`ProficiencyPrerequisiteGraph.TryFindCycle` only ever walked the *authored* prerequisite edges. That misses cycles formed by composing a cross-path authored edge with a path's **implicit** within-path tier ordering (tier N+1 requires tier N maxed) — e.g. Path A tier 0 gates on Path B tier 2, and Path B tier 0 gates on Path A tier 0. Neither edge revisits its own path, so the authored-edges-only check sees no cycle, but Path B's implicit chain (tier 2 needs tier 1 needs tier 0) composes them into `A0 -> B2 -> B1 -> B0 -> A0` — every node on that cycle silently deadlocks, same failure mode #2128 fixed for the direct same-path case.

- Added `ProficiencyPrerequisiteGraph.BuildGraph(tiers, prerequisites)`, which folds one implicit edge per path (each tier -> the tier immediately below it, ordered by `PathOrdinal`) into the authored-edges dictionary before cycle detection runs.
- Wired it into both places cycle detection already ran:
  - `AdminProficiencies.FindPrerequisiteCycle` — admin save-time rejection.
  - `ProficienciesCacheHolder.BuildSnapshotAsync` — the build-time backstop against a seed/migration mistake.

## Scope note

The issue's suggested direction also mentioned extending `ProgressionGraphChecker` (Content Health) for parity. I left that out of this PR:

- Content Health can never actually *observe* a live cycle today, since the admin save path already rejects one before it's persisted — adding it there would be defense-in-depth/consistency, not closing a reachable gap.
- `ProficiencyPrerequisiteGraph`/`DirectedGraphCycleDetector` are `internal` to `Game.DataAccess`, while `ProgressionGraphChecker` lives in `Game.Application` (which doesn't reference `Game.DataAccess` per the onion-architecture isolation in `docs/backend.md`). Doing this properly needs a short design pass on where the shared cycle-detection utility should live (the issue suggests `Game.Core`, since acyclicity is arguably a domain invariant) rather than duplicating the algorithm across layers.

Happy to pick that up as a follow-up if it's wanted — flagging it here rather than guessing at the right home for the shared logic.

## Test plan

- [x] `dotnet test` (full solution) — 3350/3350 passing.
- [x] New unit tests in `ProficiencyPrerequisiteGraphTests` cover: a within-path tier chain alone (no cycle), a same-path authored+implicit composed cycle, the issue's exact cross-path + implicit composed cycle scenario (asserting the authored-edges-only graph shows no cycle while the combined graph does), and that tiers on different paths never chain together.
- [x] New integration test in `AdminProficienciesIntegrationTests` (`SetPrerequisites_CrossPathCycleComposedWithImplicitTierOrdering_ReturnsFailure`) reproduces the issue's scenario end-to-end through `AdminProficiencies.SetPrerequisites` against a real database.

Closes #2144

---
_Generated by [Claude Code](https://claude.ai/code/session_017PKn4R8pPGMUAwyXTP91GY)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs
@@ -699,6 +699,49 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SetPrerequisites_CrossPathCycleComposedWithImplicitTierOrdering_ReturnsFailure()
+        {
+            // The #2144 scenario: Path A tier 0 already gates on Path B tier 2 (a cross-path edge, allowed on
+            // its own). Path B tier 0 -> Path A tier 0 is also cross-path on its own — but Path B's implicit
+            // within-path chain (tier 2 needs tier 1 needs tier 0) composes the two authored edges into
+            // A0 -> B2 -> B1 -> B0 -> A0, a cycle invisible to a check over the authored edges alone.
+            int a0Id, b0Id;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                var pathA = await SeedPathAsync(seedScope);
+                var a0 = await SeedProficiencyAsync(seedScope, pathA.Id, pathOrdinal: 0, name: "A0");
+                a0Id = a0.Id;
+
+                var pathB = await SeedPathAsync(seedScope);
+                var b0 = await SeedProficiencyAsync(seedScope, pathB.Id, pathOrdinal: 0, name: "B0");
+                await SeedProficiencyAsync(seedScope, pathB.Id, pathOrdinal: 1, name: "B1");
+                var b2 = await SeedProficiencyAsync(seedScope, pathB.Id, pathOrdinal: 2, name: "B2");
+                b0Id = b0.Id;
+
+                context.Set<Entities.ProficiencyPrerequisite>().Add(new Entities.ProficiencyPrerequisite
+                {
+                    ProficiencyId = a0Id,
+                    PrerequisiteProficiencyId = b2.Id,
+                });
+                await context.SaveChangesAsync(CancellationToken);
+            }
+            await ReloadReferenceCachesAsync();
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminProficiencies>();
+
+            var result = admin.SetPrerequisites([new SetProficiencyPrerequisitesData
+            {
+                Id = b0Id,
+                PrerequisiteIds = [a0Id],
+            }]);
+
+            Assert.False(result.Succeeded);
+            Assert.Contains("cycle", result.ErrorMessage);
+        }
+
+        [Fact]
         public async Task SetPrerequisites_BatchNamesTheSameProficiencyTwice_ReturnsFailure()
         {
             int proficiencyId, otherId;
--- a/Game.Application.Tests/DataAccess/ProficiencyPrerequisiteGraphTests.cs
+++ b/Game.Application.Tests/DataAccess/ProficiencyPrerequisiteGraphTests.cs
@@ -55,7 +55,61 @@ namespace Game.Application.Tests.DataAccess
             Assert.False(ProficiencyPrerequisiteGraph.TryFindCycle(graph, out _));
         }
 
+        [Fact]
+        public void BuildGraph_WithinPathTierChain_AddsNoCycleOnItsOwn()
+        {
+            // Three tiers of one path (ordinals 0, 1, 2) and no authored prerequisites: the implicit
+            // tier-N+1-needs-tier-N chain is a straight line, not a cycle.
+            var tiers = Tiers((0, path: 0, ordinal: 0), (1, path: 0, ordinal: 1), (2, path: 0, ordinal: 2));
+            var graph = ProficiencyPrerequisiteGraph.BuildGraph(tiers, Graph());
+            Assert.False(ProficiencyPrerequisiteGraph.TryFindCycle(graph, out _));
+        }
+
+        [Fact]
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: composed-cycle detection for proficiency prerequisites (#2144)

**Verdict:** Approve _(submitted as a comment since I'm the PR author — GitHub won't take a gating verdict on one's own PR)_ — the change closes the reachable gap exactly, respects the layering and reference-data rules, and is thoroughly tested. No blocking or should-fix findings.

### Correctness

- Edge direction is consistent across authored and implicit edges: authored `proficiency → prerequisite` and implicit `currentId (tier N+1) → previousId (tier N)` both encode "must be maxed first". The #2144 scenario resolves to `A0 → B2 → B1 → B0 → A0`, and the DFS in `DirectedGraphCycleDetector` catches it.
- `BuildGraph` builds fresh dictionaries and lists (`ToDictionary`, `[.. kv.Value]`, `[.. edges, previousId]`) and never mutates the cached `AllProficiencyEntities()` / `AsNoTracking().Include(...)` graphs — the reference-data gotcha in `backend.md` is respected on both the admin and build-time paths.
- Each id is a `currentId` at most once (single path membership, distinct ordinals), so the per-tier list rebuild stays O(1) per node — no accidental O(n²).
- Both existing cycle-check call sites are rewired identically, so admin save-time rejection and the build-time backstop stay in lockstep.

### Guideline alignment

- The change stays entirely within `Game.DataAccess`; no domain/orchestration logic leaks. The deliberate decision **not** to touch `ProgressionGraphChecker` (Content Health) is the right call: it can never observe a live cycle today (admin save rejects first), and reaching it properly requires a design pass on where the shared cycle utility should live, since `Game.Application` doesn't reference `Game.DataAccess`. Tracked as follow-up #2195 so it isn't lost when this closes #2144.
- Doc comments updated to describe the new composed-graph behavior rather than the old authored-edges-only claim — comment length is fine (the extra context is genuinely load-bearing here).

### Test coverage

- Unit tests on `BuildGraph` cover the within-path chain alone (no cycle), a same-path authored+implicit composition, the exact #2144 cross-path scenario (explicitly asserting the authored-edges-only graph shows *no* cycle while the combined graph does), and that different paths never chain.
- An integration test drives the #2144 scenario end-to-end through `AdminProficiencies.SetPrerequisites` against a real database.
- The build-time backstop in `ProficienciesCacheHolder` shares the same `BuildGraph` logic that the unit tests exercise, so it's covered transitively — consistent with how the prior backstop was tested.

### What's good

Tight, well-reasoned change with an honest scope note. The tests read as specifications — the `authored`-vs-`graph` assertion in the cross-path test documents the bug's mechanism in the test itself, which is exactly the right way to lock this in.


---

PR #2197, opened 2026-07-20
Cross-reference the PlayerName/UsernamePolicy char-rejection divergence

### LINKED ISSUE #2155: Unify (or document) the char-rejection divergence between UsernamePolicy and PlayerName
## Context

PR #2154 (closes #2095) added `UsernamePolicy.TryNormalize`, which rejects surrounding whitespace, blank, too-long, **control**, and **zero-width / Unicode `Format`-category** characters.

Its sibling rule `PlayerName.TryNormalize` (`Game.Core/Players/PlayerName.cs:39`) rejects only control characters — it still accepts zero-width/`Format` characters (zero-width space, ZWJ/ZWNJ, BOM, etc.).

## Problem

Two closely-related domain rules now read differently for the same class of invisible-character input. Someone maintaining one may reasonably assume the other behaves identically, which is a maintenance trap. The divergence may well be **intentional** — character names aren't globally unique and have no admin/security surface, whereas usernames appear in the admin roster, ban/archive surface, and logs, so confusable-name spoofing matters more there.

## Options

1. Extract the shared `IsDisallowedCharacter` predicate (control + `Format`) into a small shared helper and apply it to both, unifying the rule.
2. Keep them separate but add a short cross-reference comment on each explaining the intended difference in threat model.

Leaning toward (2) unless we decide zero-width characters in character names are also worth blocking (they render as confusables in-game too, just with lower stakes). Worth a quick decision either way.

No user-facing behavior change is strictly required; this is about keeping two sibling rules from silently drifting.

Related: #2095, #2154

### LINKED ISSUE #2156: PlayerName and UsernamePolicy normalization rules have quietly diverged (zero-width character handling)
## Context

Follow-up to #2095 (PR #2154), flagged in that PR's review.

`PlayerName.TryNormalize` (`Game.Core/Players/PlayerName.cs:39`) trims and rejects blank values and control characters only. `UsernamePolicy.TryNormalize` (`Game.Core/Identity/UsernamePolicy.cs`, added in #2154) mirrors that shape but additionally rejects characters in the Unicode `Format` category (zero-width space, zero-width joiner/non-joiner, the zero-width no-break space/BOM) — invisible characters that could otherwise render two usernames identically.

The extra check was deliberately scoped to `UsernamePolicy` in #2154 because usernames (unlike character names) surface in the admin roster, ban/archive surface, and logs, where a visually-confusable duplicate is a real spoofing concern; character names aren't unique and have no admin surface. That reasoning is sound, but it means two sibling "normalize this player-supplied identifier" rules now read differently with no comment cross-referencing the other, which is a maintenance trap — the next person touching either one has no signal that the divergence is intentional rather than an oversight.

## What needs deciding

- **Document the intended difference**: add a short cross-reference note in each file's doc comment explaining why `UsernamePolicy` is stricter (visible-in-admin-surface / uniqueness-sensitive) while `PlayerName` is not.
- **Or unify**: if zero-width character rejection is worth having on `PlayerName` too (e.g. a character name could theoretically be used to impersonate another player's name in social contexts, even without global uniqueness), extend `PlayerName.TryNormalize` the same way and drop the asymmetry entirely.

Low priority — cosmetic/documentation-level inconsistency, not a functional bug.

ginderjeremiah:
## Summary

`PlayerName.TryNormalize` (`Game.Core/Players/PlayerName.cs`) and `UsernamePolicy.TryNormalize` (`Game.Core/Identity/UsernamePolicy.cs`) are sibling "normalize this player-supplied identifier" rules that reject different sets of invisible characters:

- `PlayerName` rejects only control characters.
- `UsernamePolicy` additionally rejects Unicode `Format`-category characters (zero-width space, ZWJ/ZWNJ, BOM, etc.).

Both #2155 and #2156 flagged this as a maintenance trap — the divergence is intentional, but nothing recorded *why*, so the next person to touch either rule has no signal that it isn't an oversight.

## Decision: document, don't unify

I went with option 2 from both issues (document the difference) rather than unifying the two rules:

- Character names (`PlayerName`) are non-unique and the game currently has **no social/competitive surface** where they're compared to each other — no chat, guilds, trading, or leaderboards exist in the design docs (`docs/game-design.md`). A visually-confusable duplicate character name has no exploitable effect today.
- Usernames (`UsernamePolicy`) surface in the admin roster, ban/archive tooling, and logs, where a confusable duplicate is a real spoofing concern (e.g. an account rendering identically to an admin's).

Since there's no current mechanism where character-name confusability is a risk, extending `PlayerName` to match would be speculative hardening rather than fixing a real gap — so I documented the intentional asymmetry on both types instead of adding a shared predicate.

## Changes

- `PlayerName.TryNormalize`: doc comment now explains it deliberately omits the zero-width/Format-category check, cross-referencing `UsernamePolicy` and why.
- `UsernamePolicy.TryNormalize`: doc comment now explains why it's stricter, cross-referencing `PlayerName`.

No behavior change — this is a documentation-only fix for both issues.

## Test plan

- [x] `dotnet build` (full solution) — succeeds, 0 warnings/errors
- [x] `dotnet test Game.Core.Tests` filtered to `PlayerNameTests` / `UsernamePolicyTests` — 38/38 passed
- No new tests needed since behavior is unchanged.

Closes #2155
Closes #2156

---
_Generated by [Claude Code](https://claude.ai/code/session_01KWxb7hbAnxPi93yp8DWxPi)_

### DIFF
--- a/Game.Core/Identity/UsernamePolicy.cs
+++ b/Game.Core/Identity/UsernamePolicy.cs
@@ -25,6 +25,10 @@ namespace Game.Core.Identity
         /// it only when the trimmed value is within [<see cref="MinLength"/>, <see cref="MaxLength"/>] and
         /// free of control and zero-width characters (which could otherwise render a username visually
         /// identical to another). Returns the normalized username via <paramref name="normalized"/> when valid.
+        /// Stricter than <see cref="Game.Core.Players.PlayerName.TryNormalize"/> deliberately: usernames
+        /// surface in the admin roster, ban/archive surface, and logs, where a confusable duplicate is a
+        /// real spoofing concern that character names don't share. Not an oversight; keep in sync only if
+        /// that changes.
         /// </summary>
         public static bool TryNormalize(string? username, [NotNullWhen(true)] out string? normalized)
         {
--- a/Game.Core/Players/PlayerName.cs
+++ b/Game.Core/Players/PlayerName.cs
@@ -21,6 +21,10 @@ namespace Game.Core.Players
         /// Validates and normalizes a player-supplied name: trims surrounding whitespace, then accepts it
         /// only when the trimmed value is within [<see cref="MinLength"/>, <see cref="MaxLength"/>] and free
         /// of control characters. Returns the normalized name via <paramref name="normalized"/> when valid.
+        /// Deliberately does <b>not</b> reject zero-width/Unicode "Format"-category characters the way
+        /// <see cref="Game.Core.Identity.UsernamePolicy.TryNormalize"/> does — character names are
+        /// non-unique with no admin/competitive surface, so a visually-confusable duplicate isn't a
+        /// spoofing risk the way it is for logins. Not an oversight; keep in sync only if that changes.
         /// </summary>
         public static bool TryNormalize(string? name, [NotNullWhen(true)] out string? normalized)
         {


ginderjeremiah (commented):
## Review: Documenting the intentional PlayerName / UsernamePolicy char-rejection divergence (closes #2155, #2156)

**Verdict:** Approve (intended) — posted as a `COMMENT` because I'm the PR author, so GitHub won't accept a gating event on my own PR. No blocking findings.

I verified the doc comments against the actual code and they are accurate:

- `PlayerName.TryNormalize` rejects only control characters (`trimmed.Any(char.IsControl)`, `PlayerName.cs:39`). The new comment's claim that it deliberately does **not** reject zero-width/`Format`-category characters is correct.
- `UsernamePolicy.TryNormalize` additionally rejects the Unicode `Format` category (`IsDisallowedCharacter` = `char.IsControl(c) || char.GetUnicodeCategory(c) == UnicodeCategory.Format`, `UsernamePolicy.cs:55-58`). The new comment's "stricter, deliberately" framing is correct.
- Both `<see cref="…"/>` cross-references use valid fully-qualified names, so they resolve.

The design decision (document rather than unify) is sound. The premise that character names have no social/competitive surface holds up: `docs/game-design.md` describes no chat/guild/leaderboard/trading, and `PlayerName`'s own class summary already states the game has "no competitive surface." Unifying now would be speculative hardening against a threat that doesn't exist yet, and the "keep in sync only if that changes" note leaves the right breadcrumb for whoever revisits it. This also matches the leaning in both #2155 and #2156 toward option 2.

- The comments read well and are appropriately scoped. No changes requested.

### Test coverage
- Documentation-only, no behavior change, so no new tests are warranted — consistent with the project's testing rules. CI is green (`all-checks-passed` success; backend tests pass, frontend jobs skipped as expected for a backend-only change).

### Follow-up (optional, out of scope for this doc-only PR)
- `UsernamePolicy.IsDisallowedCharacter` inspects UTF-16 `char` code units, so `Format`-category characters outside the BMP (surrogate pairs) aren't caught — a surrogate reports `UnicodeCategory.Surrogate`, not `Format`. This is pre-existing (not introduced here) and negligible in practice: every common zero-width character it aims to block (ZWSP, ZWJ/ZWNJ, BOM/ZWNBSP) is in the BMP. Worth a small tech-debt note only if astral-plane invisibles ever become a concern.

### What's good
- Accurate, symmetric cross-references on both siblings, each stating the threat-model rationale and an explicit "not an oversight" signal — exactly what both issues asked for. Minimal, correct, and closes the maintenance trap without over-engineering.

---

PR #2198, opened 2026-07-20
Remove dead ICacheService.HashGetAllIfExists and IPubSubQueue.GetProcessingCountAsync surface

### LINKED ISSUE #2173: Remove the dead ICacheService.HashGetAllIfExists and IPubSubQueue.GetProcessingCountAsync surface

Two interface members are superseded surface kept alive solely by their own tests:

- **`ICacheService.HashGetAllIfExists`** (`Game.Abstractions/Infrastructure/ICacheService.cs:67`, impl + Lua script in `Game.Infrastructure/Cache/Redis/RedisService.cs:141-156`): the only non-test production reference is a *comment* in `PlayerRepository.cs:81`. `PlayerProgressRepository` reads exclusively through `HashGetAllAndRefreshExpiry`, its #2019 fold-TTL-into-read successor.
- **`IPubSubQueue.GetProcessingCountAsync`** (`Game.Abstractions/Infrastructure/IPubSubQueue.cs:40`, impl `Game.Infrastructure/PubSub/Redis/RedisQueue.cs:115`): called only from tests; the synchronizer's stranded-item detection moved to `PeekProcessingAsync` head-identity tracking (#1702).

Both also carry a fake-implementation burden in test doubles (e.g. `SocketCommandProcessorTests`). Per CLAUDE.md's dead-code guideline, remove the members, their implementations/scripts, their direct tests, and the fake implementations — after re-confirming no callers have appeared. (The `RedisService` comment referencing the shared TYPE-then-HGETALL script shape should be reworded onto `HashGetAllAndRefreshExpiry`, which keeps the script.)

ginderjeremiah:
## Summary

Per CLAUDE.md's dead-code guideline, removes two interface members that were superseded surface kept alive only by their own tests:

- **`ICacheService.HashGetAllIfExists`** (`Game.Abstractions/Infrastructure/ICacheService.cs`, impl + Lua script in `Game.Infrastructure/Cache/Redis/RedisService.cs`) — the only non-test production reference was a comment in `PlayerRepository.cs`. `PlayerProgressRepository` reads exclusively through `HashGetAllAndRefreshExpiry`, its #2019 fold-TTL-into-read successor.
- **`IPubSubQueue.GetProcessingCountAsync`** (`Game.Abstractions/Infrastructure/IPubSubQueue.cs`, impl `Game.Infrastructure/PubSub/Redis/RedisQueue.cs`) — called only from tests; the synchronizer's stranded-item detection moved to `PeekProcessingAsync` head-identity tracking (#1702).

Re-confirmed no production callers had appeared for either member before removing.

## Changes

- Removed both interface members, their Redis implementations (including `HashGetAllIfExistsScript`), and the `RedisService`/`PlayerRepository` comments that referenced them (reworded onto `HashGetAllAndRefreshExpiry`, which keeps the shared TYPE-then-HGETALL script).
- Removed the direct tests pinning each removed member's own behavior (redundant with `HashGetAllAndRefreshExpiry`'s equivalent coverage, or with `GetProcessingCountAsync`'s reserve/acknowledge round trip already covered by `ReserveNextAsync_ParksItemOffTheQueueUntilAcknowledged`).
- For tests that used a removed method only as a verification/polling helper (not testing the removed method itself), swapped in:
  - `IPubSubQueue.PeekProcessingAsync` (already on the interface) for processing-list size checks.
  - A new side-effect-free raw-Redis `ReadHashAsync` helper in `RedisServiceIntegrationTests` for hash-content checks — deliberately not routed through `HashGetAllAndRefreshExpiry`, since that method's own TTL-refresh side effect would otherwise mask a real regression in the TTL behavior some of these tests assert on.
- Removed the fake-implementation burden in test doubles: the `throw new NotSupportedException()` stubs in `SocketCommandProcessorTests`, and the real in-memory implementations across the several `IPubSubQueue` test fakes in `DataProviderSynchronizerTests` (`InMemoryPubSubQueue`, `ConcurrencyTrackingQueue`, `ConcurrentAcknowledgeTrackingQueue`, `GatedDrainQueue`, `FaultingAcknowledgeQueue`, `MultiGatedAcknowledgeQueue`, `WedgedReserveQueue`).

No behavior change to any surviving code path.

## Test plan

- [x] `dotnet test --no-build --no-restore` (full suite, all 4 test projects against the real Postgres/Redis containers) — 3,347/3,347 passed
- [x] Repo-wide grep confirms no remaining references to either removed member (production or test)

Closes #2173

---
_Generated by [Claude Code](https://claude.ai/code/session_01EqANs1c4QF1y5nseNGxcYY)_

### DIFF
--- a/Game.Abstractions/Infrastructure/ICacheService.cs
+++ b/Game.Abstractions/Infrastructure/ICacheService.cs
@@ -60,15 +60,11 @@
         /// </summary>
         public void ReclaimAndForget(string key, string ownerValue, TimeSpan expiry);
         /// <summary>
-        /// Reads every field of the Redis hash at <paramref name="key"/> in one round trip, or
-        /// <see langword="null"/> if the key does not exist — letting a caller distinguish a genuine cache
-        /// miss from a hash that exists but happens to carry no fields.
-        /// </summary>
-        public Task<Dictionary<string, string>?> HashGetAllIfExists(string key, CancellationToken cancellationToken = default);
-        /// <summary>
-        /// Same as <see cref="HashGetAllIfExists"/>, but also resets the key's TTL to <paramref name="expiry"/>
-        /// in the same round trip on a hit (a no-op on a miss, like <see cref="GetAndRefreshExpiry"/>). Lets a
-        /// sliding-expiration hash read avoid the separate awaited HGETALL followed by a fire-and-forget expire.
+        /// Reads every field of the Redis hash at <paramref name="key"/> in one round trip, resetting its TTL
+        /// to <paramref name="expiry"/> in the same round trip on a hit (a no-op on a miss, like
+        /// <see cref="GetAndRefreshExpiry"/>). Returns <see langword="null"/> if the key does not exist —
+        /// letting a caller distinguish a genuine cache miss from a hash that exists but happens to carry no
+        /// fields.
         /// </summary>
         public Task<Dictionary<string, string>?> HashGetAllAndRefreshExpiry(string key, TimeSpan expiry, CancellationToken cancellationToken = default);
         /// <summary>
--- a/Game.Abstractions/Infrastructure/IPubSubQueue.cs
+++ b/Game.Abstractions/Infrastructure/IPubSubQueue.cs
@@ -33,13 +33,6 @@
         public Task<long> GetLengthAsync(CancellationToken cancellationToken = default);
 
         /// <summary>
-        /// The current number of items reserved via <see cref="ReserveNextAsync"/> but not yet acknowledged —
-        /// the size of the side processing list. Lets a caller detect items stranded there (so they can be
-        /// reclaimed opportunistically) without mutating anything, unlike <see cref="ReclaimProcessingAsync"/>.
-        /// </summary>
-        public Task<long> GetProcessingCountAsync(CancellationToken cancellationToken = default);
-
-        /// <summary>
         /// Returns up to <paramref name="count"/> items from the head of the queue (oldest first) WITHOUT
         /// removing them — a non-destructive read. Lets a dead-letter queue be inspected without the
         /// at-most-once exposure a destructive pop would reintroduce. A non-positive count returns an empty list.
--- a/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
@@ -370,7 +370,6 @@ namespace Game.Api.Tests.Unit
             public Task AcknowledgeAsync(string value, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task<long> ReclaimProcessingAsync(CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task<long> GetLengthAsync(CancellationToken cancellationToken = default) => throw new NotSupportedException();
-            public Task<long> GetProcessingCountAsync(CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task<IReadOnlyList<string>> PeekAsync(long count, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task<IReadOnlyList<string>> PeekProcessingAsync(long count, CancellationToken cancellationToken = default) => throw new NotSupportedException();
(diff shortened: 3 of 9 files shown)


ginderjeremiah (commented):
## Review: Remove dead `ICacheService.HashGetAllIfExists` and `IPubSubQueue.GetProcessingCountAsync` (closes #2173)

**Verdict:** Approve — clean, correctly-scoped dead-code removal; every claim in the PR body checks out and nothing surviving is broken. (Posted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author; the intended verdict is **Approve**.)

I verified the removals rather than taking the description at face value:

- **No production callers of either member.** A repo-wide grep shows each removed member appears in exactly the five files this PR touches (interface def, impl, and tests). The only non-test reference to `HashGetAllIfExists` was the comment in `PlayerRepository.cs:81`, now correctly reworded onto `HashGetAllAndRefreshExpiry`. The live hash-read path is `PlayerProgressRepository.cs:189`, which reads exclusively through `HashGetAllAndRefreshExpiry` — untouched here.
- **No orphaned implementations left behind.** Every `IPubSubQueue`/`ICacheService` test double that declared the removed members (the `SocketCommandProcessorTests` stubs and all seven `DataProviderSynchronizerTests` fakes) had them removed, so nothing dangles as dead code.
- **Comment reword is accurate.** The `HashGetAllAndRefreshExpiry` doc-comment and the `RedisService` script comment now stand on their own without referencing the deleted method, and `MapHashResult`'s "shared by" note was updated to match its now-single caller.

### Test coverage

Sound. The behavior of each removed member was already covered elsewhere, and the swaps preserve intent:

- `GetProcessingCountAsync()` → `PeekProcessingAsync(N)` for size checks. Semantically safe: `PeekProcessingAsync` is a non-destructive head read capped at `N`, and every call site uses a cap (10, or 1 in the poll-until-empty loop) that comfortably exceeds the asserted size (≤2), so `.Count`/`Single`/`Empty` stay equivalent to the old count. Its own dedicated integration test (`GetProcessingCountAsync_Reflects…`) was removed as redundant with `ReserveNextAsync_ParksItemOffTheQueueUntilAcknowledged`.
- The new raw `ReadHashAsync` helper deliberately bypasses `HashGetAllAndRefreshExpiry` so a TTL-refresh side effect can't mask a regression in the TTL assertions — good call, and it's guarded by a `KeyExistsAsync` check so it never hits `WRONGTYPE` (the non-hash self-heal case it doesn't cover is still directly tested by `HashGetAllAndRefreshExpiry_OnAKeyHoldingANonHashValue_TreatsItAsAMissAndClearsIt`).

CI is green (backend + all e2e browsers pass; frontend jobs correctly skipped) and the PR reports 3,347/3,347 tests passing.

- `ReadHashAsync` opens and tears down a fresh `ConnectionMultiplexer` on every call, and it runs inside `PollUntilAsync` loops — so a slow-settling assertion churns several full Redis connections. This mirrors the existing `ReadTtlAsync` helper exactly (which also connects per call inside a poll loop), so it's consistent with the file's established pattern and not something this PR introduced. Worth a small follow-up (`tech debt`, `scope: small`) to share one multiplexer across these integration-test read helpers, but not a reason to hold this PR.

### What's good

Careful, honest change: it reworded rather than deleted the load-bearing self-heal comments, kept the shared TYPE-then-HGETALL script on the surviving method, and chose `PeekProcessingAsync` (already on the interface) over resurrecting a count method — no new surface added to remove old surface. The reasoning for the bespoke `ReadHashAsync` helper (avoiding side-effect coupling in the tests) is exactly right.

---

ginderjeremiah:
Merging. Verdict **Approve**, all CI checks green, and every removal claim was independently verified — no production callers remained for either `HashGetAllIfExists` or `GetProcessingCountAsync`, no orphaned test-double implementations, and the test swaps (`PeekProcessingAsync` caps, the `KeyExistsAsync`-guarded `ReadHashAsync`) preserve the original assertions.

The one review nit — `ReadHashAsync` opening a fresh `ConnectionMultiplexer` per call inside poll loops — is pre-existing (it mirrors the existing `ReadTtlAsync` helper, not introduced here), so it's captured as a follow-up rather than fixed in this PR: #2201.

---

PR #2200, opened 2026-07-20
Bump SocketCommandTimeoutTests' ShortTimeout to avoid cold-start flake

### LINKED ISSUE #2184: SocketCommandTimeoutTests.CooperativelyCancellingCommand_TimesOut_AndReleasesLockSoNextCommandRuns fails when run in isolation

ginderjeremiah:
## Summary

`SocketCommandTimeoutTests.CooperativelyCancellingCommand_TimesOut_AndReleasesLockSoNextCommandRuns` uses a 200ms `ShortTimeout` budget for both the deliberately-wedged `Coop` command and a trivial follow-up `Next` command. When this test class runs in isolation (no prior test in the process having JIT/connection-pool warmed the EF/Npgsql code paths that `RunCommand`'s scope creation and `IUnitOfWork.CommitAsync` round trip through), the `Next` command's own cold-start latency could exceed the 200ms budget, causing it to spuriously time out even though its body is a no-op. This reproduced reliably (5/5) in isolation while passing in the full suite (where those code paths are already warm from hundreds of prior tests).

## Changes

- `Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs`: bumped `ShortTimeout` from 200ms to 750ms, giving cold-start headroom while staying comfortably inside `WaitTimeout` (5s) — the deliberately-wedged `Hung`/`Coop` commands (parked on a `TaskCompletionSource` that's never released until the test explicitly signals it) still reliably time out and exercise the same behavior regardless of the exact budget value.

No production code changed — this is a test-only timing fix.

## Test plan

- [x] `dotnet test -- --filter-class "*SocketCommandTimeoutTests"` run in isolation, twice — 6/6 passed both times (previously reproduced the "Next" timeout failure per the issue report)
- [x] `dotnet test --no-build --no-restore` (full suite, all 4 test projects against the real Postgres/Redis containers) — all green (835/835 in `Game.Api.Tests`)

Closes #2184

---
_Generated by [Claude Code](https://claude.ai/code/session_013MdmBiwwyNxp785bLtv5kA)_

### DIFF
--- a/Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs
+++ b/Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs
@@ -22,7 +22,12 @@ namespace Game.Api.Tests.Integration
     [Collection("Integration")]
     public class SocketCommandTimeoutTests : ApiIntegrationTestBase
     {
-        private static readonly TimeSpan ShortTimeout = TimeSpan.FromMilliseconds(200);
+        // Long enough that a cold JIT/connection-pool run (e.g. this class run in isolation, with no prior
+        // test in the process having warmed the EF/Npgsql code paths the `Next`/`Coop` commands' own
+        // scope-creation and SaveChangesAsync round trip through) doesn't spuriously blow the budget itself —
+        // short enough that the deliberately-wedged `Hung`/`Coop` commands above still reliably time out well
+        // within WaitTimeout (#2184).
+        private static readonly TimeSpan ShortTimeout = TimeSpan.FromMilliseconds(750);
         private static readonly TimeSpan WaitTimeout = TimeSpan.FromSeconds(5);
 
         public SocketCommandTimeoutTests(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper)
--- a/Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs
+++ b/Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs
@@ -25,7 +25,7 @@ namespace Game.Api.Tests.Integration
         // Long enough that a cold JIT/connection-pool run (e.g. this class run in isolation, with no prior
         // test in the process having warmed the EF/Npgsql code paths the `Next`/`Coop` commands' own
         // scope-creation and SaveChangesAsync round trip through) doesn't spuriously blow the budget itself —
-        // short enough that the deliberately-wedged `Hung`/`Coop` commands above still reliably time out well
+        // short enough that the deliberately-wedged `Hung`/`Coop` commands below still reliably time out well
         // within WaitTimeout (#2184).
         private static readonly TimeSpan ShortTimeout = TimeSpan.FromMilliseconds(750);
         private static readonly TimeSpan WaitTimeout = TimeSpan.FromSeconds(5);


Game.Api.Tests/Integration/SocketCommandTimeoutTests.cs
```diff
@@ -22,7 +22,12 @@ namespace Game.Api.Tests.Integration
     [Collection("Integration")]
     public class SocketCommandTimeoutTests : ApiIntegrationTestBase
     {
-        private static readonly TimeSpan ShortTimeout = TimeSpan.FromMilliseconds(200);
+        // Long enough that a cold JIT/connection-pool run (e.g. this class run in isolation, with no prior
+        // test in the process having warmed the EF/Npgsql code paths the `Next`/`Coop` commands' own
+        // scope-creation and SaveChangesAsync round trip through) doesn't spuriously blow the budget itself —
+        // short enough that the deliberately-wedged `Hung`/`Coop` commands above still reliably time out well
```
ginderjeremiah: Nit (optional): the wedged `Hung`/`Coop` commands are defined in the test methods _below_ this field, not "above" — reads a bit off. Minor wording only; the reasoning itself is spot on.

---

ginderjeremiah (commented):
## Review: Bump `SocketCommandTimeoutTests.ShortTimeout` to avoid cold-start flake (#2184)

**Verdict:** Approve — a correct, minimal, test-only fix that removes a real isolation-order flake with no behavior change to production code.

_(Submitted as a `COMMENT` because GitHub won't accept a gating verdict on one's own PR; the intended verdict is **Approve**.)_

### Assessment

The diagnosis in #2184 holds up: `ShortTimeout` governed both the deliberately-wedged command _and_ the trivial no-op `Next` command, so on a cold JIT/connection-pool run the `Next` command's own scope-creation + commit round trip could exceed the 200ms budget and spuriously time out. Bumping to 750ms is the cleaner of the two suggested options and lands squarely in the recommended 500ms–1s range.

The value is safe on both sides:
- **Ceiling:** 750ms ≪ the 5s `WaitTimeout`, so nothing that polls for the timeout response or waits on the queued command is at risk.
- **Floor:** the `Hung`/`Coop`/`Faulty` commands park on gates never released until the test explicitly signals them, so they still time out reliably — the budget only affects _when_, not _whether_.

All four `ShortTimeout` consumers still assert correctly under the larger budget; the only cost is ~0.5s of extra wall-clock per timing test, which is immaterial.

### Nits

- `SocketCommandTimeoutTests.cs:28` — the comment says the wedged commands are "above", but they're defined in the test methods below the field. Optional wording tweak.

The comment is on the longer side (5 lines), but per the reviewing guidance for this repo that's acceptable here — it captures the genuinely non-obvious cold-start-floor vs. `WaitTimeout`-ceiling tradeoff, which is exactly the context that stops someone from naively lowering the value again.

### Test coverage

No new tests needed or expected — this is a test-only timing fix that makes an existing integration test order-independent, with no production code changed. Not battle logic, so no cross-suite mirroring applies.

### What's good

Correctly identified as unrelated to the #2174 work where it was found, scoped tightly to the one constant, and documented with the reasoning inline so the tradeoff survives future edits.

_Not merging: CI is still in progress (`test-backend` + the three `test-e2e` jobs), and there's an open nit — leaving the merge call to you._

---

ginderjeremiah:
## Re-review: cleared for merge

Re-reviewed the delta since the prior approval (commit `5d84a85` → `420dd60`). The only change is commit `420dd60`, which fixes the exact wording nit the earlier review flagged — the comment now correctly says the wedged `Hung`/`Coop` commands are defined *below* the field. No functional change.

**Verdict:** Approve. The fix is correct, minimal, and test-only:
- 750ms sits in the issue's recommended 500ms–1s range, well under the 5s `WaitTimeout` (ceiling safe) and far above cold-start latency for the no-op `Next` command (floor safe).
- The wedged commands park on gates never released until the test signals them, so they still time out reliably regardless of the budget value.
- No production code touched; not battle logic, so no cross-suite mirroring applies.

**CI:** all green on `420dd60` — `test-backend`, `test-e2e` (chromium/firefox/webkit), `verify-backend`, and the `all-checks-passed` aggregate all succeeded; frontend jobs correctly skipped (no frontend changes).

No outstanding nits or should-fixes and no follow-up work required. Merging.

---

PR #2204, opened 2026-07-20
Share a ConnectionMultiplexer across RedisServiceIntegrationTests read helpers

### LINKED ISSUE #2201: Share a ConnectionMultiplexer across integration-test raw-Redis read helpers
## Context

Surfaced during review of #2198.

`RedisServiceIntegrationTests` has raw-Redis read helpers (`ReadTtlAsync`, and the newly-added `ReadHashAsync`) that each open a fresh `ConnectionMultiplexer` per call. Because these helpers are frequently invoked inside `PollUntilAsync` loops, this causes avoidable connection churn during test runs.

This is a **pre-existing pattern** (`ReadHashAsync` deliberately mirrors `ReadTtlAsync` for consistency), not a regression introduced by #2198 — hence a follow-up rather than an in-PR change.

## Suggested change

Share a single `ConnectionMultiplexer` across these read helpers (e.g. a lazily-initialized, disposed-with-the-fixture connection) instead of connecting per call inside poll loops.

## Notes

- Test-only; no production behavior change and no user-facing effect.
- Low risk, small scope.

_Follow-up from PR #2198._

### LINKED ISSUE #2202: RedisServiceIntegrationTests' ReadTtlAsync/ReadHashAsync open a fresh ConnectionMultiplexer per call inside poll loops
## Context

Flagged in review on #2198 (removing dead `ICacheService.HashGetAllIfExists` / `IPubSubQueue.GetProcessingCountAsync`, closes #2173).

`Game.Application.Tests/DataAccess/RedisServiceIntegrationTests.cs` has two private helpers that each connect directly to Redis for verification reads outside the `ICacheService` abstraction:

- `ReadTtlAsync` (pre-existing)
- `ReadHashAsync` (added in #2198, deliberately side-effect-free so it doesn't route through `HashGetAllAndRefreshExpiry`'s own TTL-refresh, which would otherwise mask a regression in the TTL assertions some tests make)

Both open and tear down a fresh `ConnectionMultiplexer` on every call (`await using var multiplexer = await ConnectionMultiplexer.ConnectAsync(Containers.CacheConnectionString);`), and several call sites invoke them repeatedly inside `PollingHelper.PollUntilAsync` loops — so a slow-settling assertion churns through several full Redis connections in a single test.

## Suggested fix

Share one multiplexer across these integration-test read helpers (e.g. a lazily-created instance field on the test class, disposed in a fixture/test teardown, or a small helper class wrapping a cached `ConnectionMultiplexer`) instead of reconnecting per call.

## Why this is separate

Non-blocking per the #2198 review — this mirrors an existing pattern (`ReadTtlAsync`) that PR didn't introduce, just extended with one more helper of the same shape. Low priority: it's test-only overhead, not a production or correctness issue.

ginderjeremiah:
## Summary

`RedisServiceIntegrationTests`' raw-Redis read helpers (`ReadTtlAsync`, `ReadHashAsync`) each opened a fresh `ConnectionMultiplexer` per call. Both are invoked repeatedly inside `PollUntilAsync` poll loops, so a slow-settling assertion churned through several full Redis connections in a single test.

## Changes

- `Game.Application.Tests/DataAccess/RedisServiceIntegrationTests.cs`: added a `GetReadMultiplexerAsync()` helper that lazily creates and caches one `ConnectionMultiplexer` per test instance, and routed `ReadTtlAsync`/`ReadHashAsync` through it instead of connecting per call. The cached connection is disposed alongside the rest of the fixture.
- `Game.TestInfrastructure/Base/ApplicationIntegrationTestBase.cs`: made `DisposeAsync()` `virtual` so `RedisServiceIntegrationTests` can override it to dispose the cached multiplexer before calling `base.DisposeAsync()`. No other derived integration test class currently overrides it, so this is non-breaking for the rest of the suite.

`FlushScriptCacheAsync` (a one-off `AllowAdmin` connection used once, not inside a poll loop) is left as-is — it's outside the scope of the churn this issue describes.

Test-only change; no production behavior affected.

## Duplicate issue note

#2201 and #2202 were filed seconds apart and describe the exact same fix (same helpers, same root cause, same suggested change) — both are closed by this PR.

## Test plan

- [x] `dotnet test Game.Application.Tests --no-build --no-restore -- --filter-class "*RedisServiceIntegrationTests"` — 17/17 passed
- [x] `dotnet test --no-build --no-restore` (full suite, all 4 test projects against the real Postgres/Redis containers) — all green (3350/3350)

Closes #2201
Closes #2202

---
_Generated by [Claude Code](https://claude.ai/code/session_012vNz1VVqjTJPrLD4WcBsgR)_

### DIFF
--- a/Game.Application.Tests/DataAccess/RedisServiceIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/RedisServiceIntegrationTests.cs
@@ -19,7 +19,19 @@ namespace Game.Application.Tests.DataAccess
     [Collection("Integration")]
     public class RedisServiceIntegrationTests : ApplicationIntegrationTestBase
     {
+        private ConnectionMultiplexer? _readMultiplexer;
+
         public RedisServiceIntegrationTests(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper) : base(containers, testOutputHelper) { }
+
+        public override async ValueTask DisposeAsync()
+        {
+            if (_readMultiplexer is not null)
+            {
+                await _readMultiplexer.DisposeAsync();
+            }
+
+            await base.DisposeAsync();
+        }
 
         [Fact]
         public async Task SetWithExpiry_WithNullValue_DeletesTheKey()
@@ -341,7 +353,7 @@ namespace Game.Application.Tests.DataAccess
 
         private async Task<TimeSpan?> ReadTtlAsync(string key)
         {
-            await using var multiplexer = await ConnectionMultiplexer.ConnectAsync(Containers.CacheConnectionString);
+            var multiplexer = await GetReadMultiplexerAsync();
             return await multiplexer.GetDatabase().KeyTimeToLiveAsync(key);
         }
 
@@ -350,7 +362,7 @@ namespace Game.Application.Tests.DataAccess
         // (or mask a bug in) HashGetAllAndRefreshExpiry's own TTL-refreshing side effect.
         private async Task<Dictionary<string, string>?> ReadHashAsync(string key)
         {
-            await using var multiplexer = await ConnectionMultiplexer.ConnectAsync(Containers.CacheConnectionString);
+            var multiplexer = await GetReadMultiplexerAsync();
             var db = multiplexer.GetDatabase();
             if (!await db.KeyExistsAsync(key))
             {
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Share a ConnectionMultiplexer across `RedisServiceIntegrationTests` read helpers (#2201 / #2202)

**Verdict:** Approve _(posted as a comment since I'm the PR author)_ — the change does exactly what the two duplicate issues asked for, cleanly and safely.

The fix routes `ReadTtlAsync`/`ReadHashAsync` through a lazily-created, fixture-scoped `ConnectionMultiplexer` (`GetReadMultiplexerAsync`) instead of connecting per call inside `PollUntilAsync` loops, eliminating the connection churn described in the issue. A couple of things I checked that hold up:

- **Disposal ordering is correct.** `DisposeAsync` disposes `_readMultiplexer` first, then calls `base.DisposeAsync()` (which tears down the root provider and `SuppressFinalize`s). Making the base `DisposeAsync` `virtual` is the minimal enabling change, and no other derived integration test overrides it, so it's non-breaking.
- **The lazy init needs no locking.** xUnit instantiates a fresh test-class instance per test method and the helper calls are all awaited sequentially (even inside the poll loops), so there's no concurrent access to `_readMultiplexer`.
- **Leaving `FlushScriptCacheAsync` alone is the right call** — it needs its own `AllowAdmin` connection and is invoked once, not in a poll loop, so it's genuinely outside the churn this addresses.

### Test coverage

No new tests are warranted: this is a pure refactor of test-support helpers with no behavior change, and the existing `RedisServiceIntegrationTests` cases already exercise both helpers through the shared connection.

### Nits

None.

### What's good

Scope discipline is spot-on — the churn is fixed, the out-of-scope helper is explicitly left as-is with a rationale, and the two comments explaining the shared connection and the side-effect-free raw read are concise and earn their place.


---

PR #2213, opened 2026-07-20
Fix CI backend path filter and NuGet cache key for CPM

### LINKED ISSUE #2205: CI never caught up with central package management: props-only PRs skip all backend jobs, and the NuGet cache key ignores Directory.Packages.props
Since c266d83 adopted central package/build-property management (#2054, #2117), every NuGet version and core build property (TargetFramework, Nullable, overflow checking) lives in `Directory.Build.props` / `Directory.Packages.props`. Two spots in `.github/workflows/run-tests.yml` were never updated for that move:

## 1. The backend path filter omits both props files (bug)

The `detect-changes` backend filter (`run-tests.yml:44-70`) enumerates `Game.*/**`, `Game.sln`, `build/**`, compose files, `global.json`, `.config/dotnet-tools.json`, etc. — but not `Directory.Build.props` or `Directory.Packages.props`. A PR that only bumps a package version or flips a build property runs **zero** backend jobs (`verify-backend`, `test-backend`, `test-e2e` all skip) and `all-checks-passed` (`run-tests.yml:352-366`) treats skipped as green.

This is the same failure class #1810 fixed for content/tooling paths, reintroduced by the CPM migration — and it directly undermines #2068's goal of automated dependency-update PRs, which touch precisely these unfiltered files.

## 2. The NuGet cache key hashes only `*.csproj` (tech debt)

Both restore caches key on `nuget-${{ runner.os }}-${{ hashFiles('**/*.csproj') }}` (`run-tests.yml:100`, `:179`). Under CPM the csproj files carry version-less `PackageReference`s, so a version bump leaves the primary key unchanged: `actions/cache` gets an exact-key hit on the pre-bump cache and never saves an updated one, and the bumped packages are re-downloaded on every subsequent run indefinitely. The key should hash `Directory.Packages.props` (and ideally `.config/dotnet-tools.json`, since `build/coverage.ps1` runs `dotnet tool restore` into the same package cache).

Note the Game.Api Dockerfile restore layer *was* correctly updated to copy the props files in the same commit — only the workflow was missed.

**Fix**: add both props files to the backend filter, and fold `Directory.Packages.props` (+ tools manifest) into the cache-key hash.

ginderjeremiah:
## Summary

Since the central package/build-property management (CPM) migration, `Directory.Build.props` and `Directory.Packages.props` own every NuGet version and core build property, but `.github/workflows/run-tests.yml` was never updated for the two places that matter:

- **Backend path filter (bug):** the `detect-changes` backend filter didn't list either props file, so a PR that only bumps a package version or flips a build property ran zero backend jobs (`verify-backend`, `test-backend`, `test-e2e` all skipped), and `all-checks-passed` treats skipped as green.
- **NuGet cache key (tech debt):** both restore caches keyed only on `hashFiles('**/*.csproj')`. Under CPM the csproj files carry version-less `PackageReference`s, so a package version bump left the cache key unchanged — `actions/cache` kept hitting the pre-bump cache and never saved an updated one, meaning bumped packages were re-downloaded on every run indefinitely.

## Changes

- Added `Directory.Build.props` and `Directory.Packages.props` to the `detect-changes` backend path filter.
- Extended both NuGet cache keys (`verify-backend`, `test-backend`) to also hash `Directory.Packages.props` and `.config/dotnet-tools.json` (the tools manifest `build/coverage.ps1` restores into the same package cache).

## Test plan

- [x] Verified the updated workflow file is valid YAML.
- [x] Manually confirmed `Directory.Build.props`, `Directory.Packages.props`, and `.config/dotnet-tools.json` exist at the repo root and are referenced correctly in the filter/cache-key paths.
- This is a CI-workflow-only change with no application code affected, so no backend/frontend build or test suite applies; correctness is verified by inspection of the filter and cache-key changes against the actual repo-root file layout.

Closes #2205

---
_Generated by [Claude Code](https://claude.ai/code/session_014126AN1DsqENHUt5iBgbun)_

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -66,6 +66,8 @@ jobs:
             - '.editorconfig'
             - 'global.json'
             - '.config/dotnet-tools.json'
+            - 'Directory.Build.props'
+            - 'Directory.Packages.props'
             - '.github/workflows/**'
             - 'UI/src/lib/api/types/**'
           frontend:
@@ -97,7 +99,7 @@ jobs:
       uses: actions/cache@v5
       with:
         path: ~/.nuget/packages
-        key: nuget-${{ runner.os }}-${{ hashFiles('**/*.csproj') }}
+        key: nuget-${{ runner.os }}-${{ hashFiles('**/*.csproj', 'Directory.Packages.props', '.config/dotnet-tools.json') }}
         restore-keys: |
           nuget-${{ runner.os }}-
 
@@ -176,7 +178,7 @@ jobs:
       uses: actions/cache@v5
       with:
         path: ~/.nuget/packages
-        key: nuget-${{ runner.os }}-${{ hashFiles('**/*.csproj') }}
+        key: nuget-${{ runner.os }}-${{ hashFiles('**/*.csproj', 'Directory.Packages.props', '.config/dotnet-tools.json') }}
         restore-keys: |
           nuget-${{ runner.os }}-
 


ginderjeremiah (commented):
## Review: CI fix so props-only PRs run backend jobs and the NuGet cache key tracks CPM package versions (closes #2205)

**Verdict:** Approve — the change fully addresses both defects in #2205, is correctly scoped, and I verified it against the current `run-tests.yml`. (Submitted as a `COMMENT` only because the reviewing account is the PR author; GitHub won't accept a gating verdict on one's own PR.)

I reviewed the full workflow at head rather than just the hunks, and confirmed the issue's claims are accurate:

- **Backend path filter (bug):** `Directory.Build.props` / `Directory.Packages.props` are now in the `detect-changes` backend filter, so a props-only PR (e.g. a package bump or a `Nullable`/`TargetFramework` flip) will run `verify-backend` / `test-backend` / `test-e2e` instead of skipping and being waved through by `all-checks-passed`. Correctly left out of the *frontend* filter, since the props files are backend-only.
- **NuGet cache key (tech debt):** both keys now hash `Directory.Packages.props` and `.config/dotnet-tools.json`. The subtle-but-correct call here is *not* adding `Directory.Build.props` to the cache key — under CPM it carries build properties, not package versions, so it has no bearing on restore; only `Directory.Packages.props` (versions) and the tools manifest (`build/coverage.ps1` restores tools into the same package cache) belong there. `hashFiles(...)` accepts multiple comma-separated patterns, and the `restore-keys` fallback still yields a partial hit on a bump before saving under the new key.

### Test coverage

Not applicable — this is a CI-YAML-only change with no application or domain logic, so the project's unit/integration/e2e rules don't attach. Correctness is properly established by inspecting the filter and cache-key paths against the actual repo-root layout, which I've done.

### What's good

Tight, minimal fix that resolves the exact failure class it describes, with the right distinction between the path filter (both props files) and the cache key (versions + tools manifest only). No footguns, no dead code, nothing to nit.

I'll merge once CI is green.

---

PR #2214, opened 2026-07-20
Fix zone editor's retired-enemy spawn accounting

### LINKED ISSUE #2206: Workbench zone editor treats retired enemies' spawn rows as live — wrong spawn-share % and a suppressed "no enemies spawn here" warning
A retired enemy keeps its `ZoneEnemy` rows in the contract data, but the runtime excludes it from the spawn tables (`Game.DataAccess/Repositories/Caching/EnemiesCacheHolder.cs:72-76`), and the `EmptyCombatZone` lint (`Game.Application/Content/ProgressionGraphChecker.cs:906-912`) counts live spawns only. The zone editor ignores `retiredAt` in all three places it consumes spawn rows, so the Workbench presents a picture the runtime disagrees with:

- **`UI/src/routes/admin/workbench/entities/zone.ts:23-27`** — `zoneEnemies` is built by flat-mapping *all* enemies' spawn rows with no `retiredAt` filter, so a retired enemy renders as a normal live spawn row.
- **`zone.ts:138`** — `warn: (z) => (z.isHome || z.zoneEnemies.length ? null : 'No enemies spawn here')`: a live zone whose only spawn rows belong to retired enemies shows **no** warning, even though the backend lint flags exactly that zone and the runtime relocates players out of it. Reachable state: retire zone → retire its enemies (the save guard only protects *live* zones) → reinstate the zone (reinstate has no guard).
- **`UI/src/routes/admin/workbench/reference.svelte.ts:335-348`** — `enemySpawnShareTotal` sums every other enemy's weight for the zone with no retired filter, so the enemy editor's "Spawn share" % understates a live enemy's real share whenever a retired competitor's rows persist — the normal outcome of retiring any non-last spawn.

**Fix**: filter `retiredAt` in all three spots (and probably render retired enemies' rows in the zone's spawn table with a retired affordance rather than hiding them, consistent with how pickers keep existing references visible — #295 lineage).

ginderjeremiah:
## Summary

The Workbench zone editor treated a retired enemy's spawn rows as live, disagreeing with the runtime in three places:

- **`zone.ts`'s "No enemies spawn here" warning** — computed off the raw `zoneEnemies` row count, so a zone whose only spawn rows belonged to retired enemies showed no warning, even though the backend's `EmptyCombatZone` lint flags exactly that zone and the runtime relocates players out of it.
- **The zone's own "Spawn share" column** — defaulted to summing every sibling row's weight (including retired enemies'), overstating the denominator.
- **The enemy editor's "Spawn share" column** (`reference.svelte.ts`'s `enemySpawnShareTotal`) — summed every *other* enemy's weight in the zone with no retired filter, understating a live enemy's real share whenever a retired competitor's rows persisted.

All three now agree with the runtime spawn table (`EnemiesCacheHolder.BuildZoneEnemyTables`), which excludes retired enemies entirely.

## Design note (deviation from the issue's literal suggestion)

The issue's suggested fix was "filter `retiredAt` in all three spots," naming `zone.ts:23-27` (the `zoneEnemies` array itself) as one of them. I deliberately **did not** filter that array at the source: it's the exact payload `childSavers` posts to `AdminTools/SetZoneEnemies`, and that endpoint fully reconciles the zone's spawn table against whatever is sent (`ChildCollectionReconciler.Reconcile`). Filtering retired rows out of the in-memory array would silently **delete** a retired enemy's `ZoneEnemy` row from the database the next time an admin saved an unrelated change to that zone (e.g. tweaking another enemy's weight) — a destructive side effect and a violation of the Workbench's own "never silently drop a reference" convention (`docs/frontend-admin.md`).

Instead, the array stays intact (so it round-trips safely on save), and the three *derived* computations (the warning and both share denominators) now explicitly skip retired enemies.

I also did not implement the issue's hedged ("probably") suggestion to add a visible "retired" affordance to the zone's spawn table rows — it's already there for free: the row's own `enemyId` select uses `reference.enemyOptions`, which keeps a retired *current* value visible with a "· retired" suffix (the same convention used by every other retireable picker in the Workbench).

## Changes

- `UI/src/routes/admin/workbench/entities/zone.ts` — added an `isLiveEnemy` helper; the "no enemies spawn here" warning and the zone's own spawn-share column now exclude retired enemies.
- `UI/src/routes/admin/workbench/reference.svelte.ts` — `enemySpawnShareTotal` now excludes retired competitors from the sum.
- Unit tests added/updated in both corresponding test files covering the warning, both share-total computations, and the "everything retired" fallback-to-1 case.

## Test plan

- [x] `npm run test:unit:ci` — full suite green (3824 tests)
- [x] New tests cover: warning suppressed only when a live spawn exists, zone-share-total excludes retired siblings, enemy-share-total excludes retired competitors, share-total falls back to 1 when every row is retired

Closes #2206

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sziy3g5dWEw9WfhJvKW97P)_

### DIFF
--- a/UI/src/routes/admin/workbench/entities/zone.ts
+++ b/UI/src/routes/admin/workbench/entities/zone.ts
@@ -10,6 +10,14 @@ export interface WorkbenchZone extends IZone {
 	zoneEnemies: IZoneEnemy[];
 }
 
+/**
+ * Whether an enemy id resolves to a live (non-retired) enemy — the runtime spawn table
+ * (`EnemiesCacheHolder.BuildZoneEnemyTables`) and the `EmptyCombatZone` lint both drop a
+ * retired enemy's spawns entirely, so anything computing "will this actually spawn" must
+ * agree.
+ */
+const isLiveEnemy = (enemyId: number): boolean => !staticData.enemies?.[enemyId]?.retiredAt;
+
 const refresh = async (): Promise<WorkbenchZone[]> => {
 	const [zones, enemies] = await Promise.all([fetchSocketData('GetZones'), fetchSocketData('GetEnemies')]);
 	staticData.zones = zones;
@@ -20,6 +28,11 @@ const refresh = async (): Promise<WorkbenchZone[]> => {
 		bossEnemyId: zone.bossEnemyId ?? -1,
 		// Likewise normalise the optional unlock-gate FK to the "None" sentinel (-1).
 		unlockChallengeId: zone.unlockChallengeId ?? -1,
+		// Includes a retired enemy's spawn row: it stays visible/editable (its own enemy
+		// picker shows "· retired") and, critically, must round-trip verbatim on save since
+		// SetZoneEnemies fully reconciles the zone's spawn table against whatever is posted —
+		// filtering it out here would delete the row from the database on the next unrelated
+		// save. Consumers that need "will this actually spawn" semantics use isLiveEnemy.
 		zoneEnemies: enemies.flatMap((enemy) => {
 			const spawn = enemy.spawns.find((s) => s.zoneId === zone.id);
 			return spawn ? [{ enemyId: enemy.id, weight: spawn.weight }] : [];
@@ -135,7 +148,9 @@ export const zoneEntity: EntityConfig<WorkbenchZone> = {
 			glyph: 'skull',
 			desc: 'Enemies that spawn here & their weights',
 			count: (z) => z.zoneEnemies.length,
-			warn: (z) => (z.isHome || z.zoneEnemies.length ? null : 'No enemies spawn here'),
+			// A retired enemy's row doesn't count toward "does something spawn here" — it never
+			// rolls at runtime (mirrors the backend's EmptyCombatZone lint).
+			warn: (z) => (z.isHome || z.zoneEnemies.some((ze) => isLiveEnemy(ze.enemyId)) ? null : 'No enemies spawn here'),
 			kind: 'table',
 			itemsKey: 'zoneEnemies',
 			rowKey: 'enemyId',
@@ -153,7 +168,17 @@ export const zoneEntity: EntityConfig<WorkbenchZone> = {
 			columns: [
 				{ key: 'enemyId', label: 'Enemy', type: 'select', options: reference.enemyOptions, min: 220, unique: true },
 				{ key: 'weight', label: 'Weight', type: 'number', align: 'r', width: 100 },
-				{ key: '__share', label: 'Spawn share', type: 'share', width: 150, weightKey: 'weight' }
+				{
+					key: '__share',
+					label: 'Spawn share',
+					type: 'share',
+					width: 150,
+					weightKey: 'weight',
+					// Excludes a retired sibling's weight from the denominator — it never rolls,
+					// so counting it would understate every live enemy's real share.
+					shareTotal: (_row, rows) =>
+						rows.reduce((sum, r) => (isLiveEnemy(r.enemyId as number) ? sum + (Number(r.weight) || 0) : sum), 0) || 1
+				}
 			]
 		}
 	],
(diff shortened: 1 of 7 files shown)


UI/src/routes/admin/workbench/entities/zone.ts
```diff
@@ -153,7 +168,17 @@ export const zoneEntity: EntityConfig<WorkbenchZone> = {
...
+				{
+					key: '__share',
+					label: 'Spawn share',
+					type: 'share',
+					width: 150,
+					weightKey: 'weight',
+					// Excludes a retired sibling's weight from the denominator — it never rolls,
+					// so counting it would understate every live enemy's real share.
+					shareTotal: (_row, rows) =>
+						rows.reduce((sum, r) => (isLiveEnemy(r.enemyId as number) ? sum + (Number(r.weight) || 0) : sum), 0) || 1
```
ginderjeremiah: Nit (non-blocking): this correctly fixes the denominator, but `shareTotal` only controls the denominator — the numerator in `TableCell` is still `row.weight` unclamped (`pct = round(weight/total*100)`, rendered verbatim as `{pct}%`). So a **retired** enemy's own row in this table now shows an inflated share — e.g. retired weight 5 against a live sum of 3 renders "167%". Before this change that row read as a sensible fraction; ironically, the runtime-truthful value for a row that never rolls is 0%.

The enemy editor's `enemySpawnShareTotal` doesn't hit this because it keeps the current enemy's own weight in the denominator (share stays ≤ 100%), so the two share columns now disagree on how a retired row displays its own share.

Fully fixing it means zeroing the numerator for retired rows, which the generic `share` column type doesn't currently expose — so this is out of scope for a small fix. Fine to leave as-is or split into a follow-up; flagging so the asymmetry is a conscious call rather than an oversight.

---

ginderjeremiah (commented):
## Review: Fix zone editor's retired-enemy spawn accounting (#2206)

**Verdict:** Approve — the three consumers now agree with the runtime spawn table (`EnemiesCacheHolder.BuildZoneEnemyTables`), the deviation from the issue's literal suggestion is correct, and coverage is thorough. Submitted as a `COMMENT` because I'm the PR author (GitHub won't accept a gating verdict on your own PR). One non-blocking nit below.

### What's good
- **The design deviation is the right call.** Not filtering `retiredAt` out of the `zoneEnemies` array at its source (`zone.ts:36`) is correct — that array is the exact payload `SetZoneEnemies` reconciles against, so filtering it would silently delete a retired enemy's `ZoneEnemy` row on the next unrelated save. Deriving `isLiveEnemy` at the three *consumption* points instead keeps the round-trip safe. The inline comment documenting *why* is exactly the kind of "a later edit might naively simplify this and break it" guardrail worth writing.
- All three derived computations (warning, zone-share denominator, enemy-share denominator) now consistently exclude retired enemies, matching the backend `EmptyCombatZone` lint.
- `isLiveEnemy` follows the established id-indexed reference-data convention (`staticData.enemies[id]`) already used by `headline`/`itemRetired`.

### Nits
- **Retired row's own share can render >100% in the zone editor** — `shareTotal` only fixes the denominator; the numerator in `TableCell` (`{pct}%`) is still the retired row's raw weight, so a retired row now shows an inflated/misleading share where the runtime-truthful value is 0%. The enemy editor doesn't hit this (it keeps the current enemy's weight in its denominator), so the two share columns disagree on the retired-own-row case. Fully fixing needs numerator control the generic `share` column doesn't expose — out of scope for a small fix, fine to leave or split to a follow-up. Details inline on `zone.ts:180`.

### Test coverage
Solid and matches the project's rules. New tests cover every changed path: warning suppressed only when a live spawn exists, warning fires when only retired rows exist, both share denominators exclude retired siblings/competitors, and the all-retired fallback-to-1. State is reset per-test via the top-level `beforeEach`, so the mutation in the retired-competitor case doesn't leak. This is a frontend-only admin display fix with no battle logic, so no mirrored backend tests are required. No inline retired-row-share assertion, but that's the nit above, not existing behavior.


---

UI/src/routes/admin/workbench/entities/zone.ts
ginderjeremiah: Fixed in a42c672: added an optional `shareValue` column override (mirroring `shareTotal`) so a row can supply its own numerator. The zone's `__share` column now reports 0 for a retired enemy's own row, so it can't render over 100% anymore. Added coverage in both `TableCell.test.ts` (component-level, asserting the 167%→0% behavior) and `zone.test.ts`.

---

PR #2215, opened 2026-07-20
Reject AdminTags.SaveTags edits/deletes of a missing tag id

### LINKED ISSUE #2210: AdminTags.SaveTags: Edit/Delete of a missing tag id surfaces as an unhandled 500 instead of a not-found rejection
`AdminTags` is the only identity-level admin save that omits the missing-id guard — every sibling passes an `editExists` check to `ChangeSetProcessor.Apply` (`AdminSkills.cs:42`, `AdminChallenges.cs:68`, `AdminEnemies.cs:48`, `AdminPaths.cs:43`, `AdminZones.cs:95`, `AdminClasses.cs:40`, `AdminProficiencies.cs:64`, `AdminSkillRecipes.cs:65`, `AdminItems.cs:85`, `AdminItemMods.cs:48`, `AdminLessons.cs:42`).

`Game.DataAccess/Repositories/Admin/AdminTags.cs:30-42` stages `_entityStore.Update(entity)` (state = `Modified`) on edit and `_entityStore.DeleteByKey<Entities.Tag>(item.Id)` on delete for whatever id the client names. Tags are the one hard-deletable reference set (no retire, no cache holder), so a stale admin client — a second tab, a second admin, a double-clicked delete — can legitimately name an id that no longer exists. The resulting 0-row UPDATE/DELETE throws `DbUpdateConcurrencyException` out of the deferred `CommitFilter.SaveChangesAsync` (`Game.Api/Filters/CommitFilter.cs:19`), an opaque 500 **after** the action already returned `AdminSaveResult.Success`. Nothing handles `DbUpdateConcurrencyException` anywhere in the repo.

This contradicts the tier's own documented standard (docs/backend-admin.md: a missing identity is "a not-found rejection … not an EF 0-row UPDATE that throws"; the `ChangeSetProcessor` doc says the same).

**Fix**: validate edit/delete ids against the existing tag ids before staging (the `Tags.GetExistingTagIds` query already exists), mirroring the siblings' `editExists` shape, and return `AdminSaveResult.NotFound("tag")`.

ginderjeremiah:
## Summary

`AdminTags.SaveTags` was the only identity-level admin save that omitted the missing-id guard every sibling repo has (`AdminSkills.cs:42`, `AdminChallenges.cs:68`, `AdminEnemies.cs:48`, etc.). Tags are the one hard-deletable reference set, so a stale admin client (a second tab, a second admin, a double-clicked delete) could legitimately name an id that no longer exists. That produced a 0-row `UPDATE`/`DELETE` that threw `DbUpdateConcurrencyException` out of the deferred `CommitFilter.SaveChangesAsync` — an opaque 500 **after** the action had already returned `AdminSaveResult.Success` — contradicting the documented standard in `docs/backend-admin.md` that a missing identity should be a clean not-found rejection, not an EF 0-row throw.

## Fix

Tags have no in-memory reference cache to check membership against (unlike the zero-based reference sets' `editExists` cache lookup), so `SaveTags` now:

- Collects the ids named by any Edit/Delete change.
- Validates them against the database via the existing `ITagAssignmentQueries.GetExistingTagIds` query, up front, before staging anything.
- Returns `AdminSaveResult.NotFound("Tag")` if any targeted id doesn't exist.

Because the guard needs a database round-trip, `SaveTags` (and `IAdminTags.SaveTags`) become async, following the same `Task<AdminSaveResult>` shape the other admin relationship setters (e.g. `AdminItems.SetTags`) already use. The controller action follows suit, using `HttpContext.RequestAborted` like its siblings.

## Testing

- Unit tests (`AdminTagsTests.cs`) — added a fake `ITagAssignmentQueries` (mirrors the existing `RecordingEntityStore` fake pattern) so the pure entity-mapping/staging tests stay database-free, plus two new tests covering the Edit/Delete not-found rejection.
- Integration tests (`AdminTagsIntegrationTests.cs`) — added two tests that seed a tag, delete it directly (bypassing the admin write) to guarantee a genuinely absent id, then assert the admin `SaveTags` call rejects with `"Tag not found."` for both Edit and Delete.
- `dotnet build` — clean, 0 warnings/errors.
- Full `dotnet test` — all 3354 tests pass (1415 Core, 68 Infrastructure, 1036 Application, 835 Api).

Closes #2210

---
_Generated by [Claude Code](https://claude.ai/code/session_01SNFQmeYaHjb9bdPWHzqP8R)_

### DIFF
--- a/Game.Abstractions/DataAccess/Admin/IAdminTags.cs
+++ b/Game.Abstractions/DataAccess/Admin/IAdminTags.cs
@@ -10,8 +10,8 @@ namespace Game.Abstractions.DataAccess.Admin
     public interface IAdminTags
     {
         /// <summary>Applies an identity-level Add/Edit/Delete change set to the tag catalogue. Tags carry
-        /// their own identity and keep the hard-delete lifecycle, so this never rejects — it succeeds to
-        /// share the unified <see cref="AdminSaveResult"/> contract every admin write reports through.</summary>
-        AdminSaveResult SaveTags(IReadOnlyList<Change<Tag>> changes);
+        /// their own identity and keep the hard-delete lifecycle. An Edit or Delete naming an id that no
+        /// longer exists is rejected as a not-found failure rather than staged.</summary>
+        Task<AdminSaveResult> SaveTags(IReadOnlyList<Change<Tag>> changes, CancellationToken cancellationToken = default);
     }
 }
--- a/Game.Api/Controllers/Admin/AdminTagsController.cs
+++ b/Game.Api/Controllers/Admin/AdminTagsController.cs
@@ -21,9 +21,9 @@ namespace Game.Api.Controllers.Admin
         private readonly IAdminTags _adminTags = adminTags;
 
         [HttpPost]
-        public ApiResponse AddEditTags([FromBody] List<Change<Tag>> changes)
+        public async Task<ApiResponse> AddEditTags([FromBody] List<Change<Tag>> changes)
         {
-            return _adminTags.SaveTags(changes);
+            return await _adminTags.SaveTags(changes, HttpContext.RequestAborted);
         }
     }
 }
--- a/Game.Application.Tests/DataAccess/AdminTagsIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminTagsIntegrationTests.cs
@@ -1,6 +1,7 @@
 using Game.Abstractions.Contracts.Admin;
 using Game.Abstractions.DataAccess;
 using Game.Abstractions.DataAccess.Admin;
+using Game.Core;
 using Game.Infrastructure.Database;
 using Game.TestInfrastructure.Base;
 using Game.TestInfrastructure.Fixtures;
@@ -48,7 +49,7 @@ namespace Game.Application.Tests.DataAccess
                 using (var writeScope = CreateScope())
                 {
                     var admin = writeScope.ServiceProvider.GetRequiredService<IAdminTags>();
-                    Assert.True(admin.SaveTags([Delete(tag)]).Succeeded);
+                    Assert.True((await admin.SaveTags([Delete(tag)])).Succeeded);
                     await writeScope.ServiceProvider.GetRequiredService<IUnitOfWork>().CommitAsync();
                 }
             }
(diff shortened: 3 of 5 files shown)


ginderjeremiah (commented):
## Review: Reject `AdminTags.SaveTags` edits/deletes of a missing tag id (#2210)

**Verdict:** Approve (posted as a comment because I'm the PR author, so GitHub won't take a gating event from me) — the fix is correct, matches the documented admin standard, and is well covered on both the unit and integration sides.

### Analysis

- **Correctly diagnosed why the sibling pattern doesn't transfer directly.** Every other identity-level save passes a synchronous `editExists` predicate into `ChangeSetProcessor.Apply`, but that guard (a) only covers `Edit` — the retire-only sets have no delete handler, so their deletes are rejected by the `delete is null` branch — and (b) is a sync `Func<T,bool>`, which can't run the DB round-trip Tags requires (Tags has no in-memory cache to check membership against). Tags is the one hard-deletable identity set, so an up-front async check covering **both** Edit and Delete (`ChangeType != Add`) is the right shape here rather than forcing it through the processor's `editExists` hook.
- **Precedence is consistent with siblings:** category-enum validation → existence → duplicate-key dedup (the last still handled inside `ChangeSetProcessor`). A batch naming the same id in both an Edit and a Delete passes existence (the id exists) and is then caught by the dedup guard — same ordering the processor documents.
- **Good catch verifying the issue rather than copying it.** The issue suggested `NotFound("tag")` (lowercase) and pointed at `Tags.GetExistingTagIds`; the PR correctly uses `NotFound("Tag")` to match the capitalized convention every sibling uses (`"Enemy"`, `"Skill"`, …) and the `AdminSaveResult.NotFound` doc. The query reused is genuinely a general tag-existence query (`Tags` implements both `ITags` and `ITagAssignmentQueries`, already registered in DI), not a semantic misuse.
- **Async ripple is complete and clean:** `IAdminTags.SaveTags` → `Task`, the controller awaits with `HttpContext.RequestAborted` like its siblings, and there are no other production callers. No dead code left behind.

### Test coverage

- Unit (`AdminTagsTests`): the `FakeTagAssignmentQueries` stand-in mirrors the existing `RecordingEntityStore` fake pattern and keeps the mapping/staging tests database-free; the two new tests assert both Edit- and Delete-of-missing return `"Tag not found."` **without** staging (`store.Updated`/`store.DeletedByKey` empty), which is the behavior that matters.
- Integration (`AdminTagsIntegrationTests`): seeding then directly removing the row (bypassing the admin write) to guarantee a genuinely absent id is exactly the right way to reproduce the stale-client scenario — this is the coverage the previous 0-row-throws path lacked.
- No battle logic involved, so no frontend/backend mirroring obligation.

### Note (not blocking, no action needed)

The existence check is a check-then-stage against the DB, so a concurrent delete landing between the check and the deferred commit could still surface the original `DbUpdateConcurrencyException`. That TOCTOU window is inherent to every sibling's cache-then-stage guard too and is out of scope for this fix — the issue targets the common stale-client case, which this fully resolves. Flagging only for completeness; no change wanted.

Clean fix. I'll merge once the remaining CI (`test-backend`, `test-e2e (webkit)`) goes green.

---

PR #2216, opened 2026-07-20
Carry the socket access token as a WebSocket subprotocol, not a query string

### LINKED ISSUE #2211: Security: the socket access_token query string is written to the nginx edge proxy's access log on every connect
The WebSocket handshake carries the full JWT in the URL — `UI/src/lib/api/api-socket.ts:174` builds `` `/socket?access_token=${encodeURIComponent(accessToken)}` `` and `Game.Api/Startup.cs:345-354` reads it from the query string (the standard ASP.NET Core pattern, since browsers can't set headers on the WS handshake; documented in docs/backend.md).

The problem is one hop out: the production edge proxy is `FROM nginx:alpine` with only the server-block template replaced (`Dockerfile:1-4`, `nginx.conf`). The base image's default `access_log ... main;` stays active, and the `main` format records `$request` — the full request line **including the query string**. Per docs/infrastructure.md, every request reaches the API through this hop, so every connect/reconnect (idle reconnects, rolling-deploy drains, session takeovers) writes `GET /socket?access_token=<live JWT>` into the access log, which surfaces in Railway's log stream.

That token is a fully-privileged bearer credential for 15 minutes — for an admin it carries the `Admin` role claim — so anyone with log access (or a log-aggregation leak) can replay it against the HTTP admin surface and the socket. The backend's own `RequestLoggingMiddleware` deliberately logs only `Path` (`Game.Api/Middleware/RequestLoggingMiddleware.cs:20`); the care taken there is silently undone at the proxy tier.

**Fix options** (first two also cover any future proxy/CDN hop):

1. Carry the token via the `Sec-WebSocket-Protocol` header — browsers *can* set subprotocols (`new WebSocket(url, ['bearer', token])`, echoed back by the server).
2. Mint a short-lived one-time socket ticket over an authenticated POST and pass *that* in the query.
3. At minimum, redact/disable query logging for the `/socket` location in `nginx.conf` (a one-liner).

ginderjeremiah:
## Summary

- The `/socket` WebSocket handshake carried the full JWT access token in the `access_token` query-string parameter. Per #2211, the production edge nginx proxy's default `access_log` format records `$request` (the full request line, including the query string), so every connect/reconnect wrote a fully-privileged bearer credential (carrying the `Admin` role claim for admins) into Railway's log stream.
- Moves the token off the URL entirely: the frontend now offers it as the sole WebSocket subprotocol (`new WebSocket('/socket', [accessToken])`), and the backend reads it from `Sec-WebSocket-Protocol` instead of the query string. A subprotocol is a handshake header, which no default access-log format records — so this closes the leak at the source rather than only at the one nginx hop, covering any future proxy/CDN in front of the API too.
- `SocketInterceptorMiddleware` now echoes the negotiated subprotocol back on `AcceptWebSocketAsync`, since a browser fails the handshake outright if the client offered a subprotocol list and the server doesn't select one back (verified empirically against a live TestServer handshake before writing the fix).

## Design choice

The issue offered three options: (1) carry the token via `Sec-WebSocket-Protocol`, (2) mint a short-lived one-time socket ticket over an authenticated POST, or (3) at minimum, redact query-string logging in `nginx.conf`. I went with (1) — it's a well-established pattern for this exact constraint (browsers can't set WS handshake headers; e.g. Kubernetes' apiserver uses the same technique for bearer tokens over WebSocket), it's a contained change (transport only, no new endpoint/ticket-issuance flow), and it fixes the root cause for every hop rather than just this one proxy's log config. Since the token no longer appears in the URL at all, no nginx change was needed.

## How it addresses the issue

- `Game.Api/Startup.cs`: `JwtBearerEvents.OnMessageReceived` reads the token from `HttpContext.WebSockets.WebSocketRequestedProtocols` instead of `Request.Query["access_token"]`.
- `Game.Api/Middleware/SocketInterceptorMiddleware.cs`: echoes the requested subprotocol back via `AcceptWebSocketAsync(subProtocol)`.
- `UI/src/lib/api/api-socket.ts`: opens the socket with `new WebSocket('/socket', [accessToken])` instead of building an `access_token=` query string.
- Test infra (`TestSocketClient.cs`, `SocketConnectionTests.cs`) and the frontend socket test suite updated to exercise the subprotocol path.
- `docs/backend.md`, `docs/backend-sockets.md`, `docs/frontend-sockets.md` updated to describe the new transport.

Closes #2211

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test` (full solution, all 5 projects) — 3354 tests passed, 0 failed
- [x] TypeScript codegen drift check (`dotnet run --project Game.Api -- codegen`) — no diff against committed `UI/src/lib/api/types`
- [x] `npm run test:unit:ci` — 3828 tests passed, coverage gates hold
- [x] Verified empirically (throwaway TestServer harness) that `WebSocketRequestedProtocols` is readable before `UseWebSockets()` middleware runs, and that echoing the subprotocol back via `AcceptWebSocketAsync(subProtocol)` correctly negotiates the handshake

---
_Generated by [Claude Code](https://claude.ai/code/session_01W1DykJ5zR1qVLcFJfQaHrh)_


ginderjeremiah (commented):
## Review: Move the socket access token off the query string onto `Sec-WebSocket-Protocol` (#2211)

**Verdict:** Approve — the fix correctly closes the token-in-access-log leak at its source, is well-scoped to the transport, and is fully mirrored across backend, frontend, docs, and both test suites. (Posted as a comment since GitHub won't accept a self-approval on my own PR; the intended verdict is Approve.)

### Why this is the right fix
- Carrying the token as the sole requested subprotocol (`new WebSocket('/socket', [accessToken])`) removes it from the URL entirely, so it can never land in `$request` at *any* proxy/CDN hop — a strictly stronger fix than the issue's option 3 (redact `/socket` logging in one nginx config), and no nginx change is needed because the token is no longer in the URL.
- The handshake correctly echoes the negotiated subprotocol back via `AcceptWebSocketAsync(subProtocol)` (`SocketInterceptorMiddleware.cs:65-67`) — required, since a browser fails the handshake outright when it offered a subprotocol list and the server selects none. `JwtBearerEvents.OnMessageReceived` reads it from `WebSocketRequestedProtocols` before the upgrade (`Startup.cs:351-358`). Both sides are clearly commented.
- JWT characters (base64url + `.`) are all valid `Sec-WebSocket-Protocol` token chars and never contain a comma, so the token survives the header round-trip intact with no encoding concerns.

### Test coverage
- Backend: `SocketConnectionTests` and both `TestSocketClient` connect paths (real `ClientWebSocket` via `AddSubProtocol`, and the TestHost client via `SubProtocols.Add`) exercise the new transport. The real-`ClientWebSocket` path natively enforces subprotocol negotiation, so it also covers the server's echo.
- Frontend: `api-socket.test.ts` now asserts the token rides as the sole subprotocol (`lastSocketProtocols`) and that an unauthenticated open offers none. Coverage matches the source change on both sides.

### Note (non-blocking)
- The negotiated subprotocol is echoed back in the *response* `Sec-WebSocket-Protocol` header, i.e. the credential appears in a response header. This is the accepted tradeoff for this well-established pattern (default access-log formats record only the request line, not response headers), so it doesn't reintroduce the leak — just worth being aware of for any future custom response logging.

Clean, minimal, and correct. Nice work isolating the change to the transport rather than adding a ticket-issuance flow.

---

Issue #2217, opened 2026-07-20
Workbench: block Save while a record carries a warning class the backend is known to hard-reject


ginderjeremiah:
## Context

Follow-up to #2207. That issue's fix ([save-helpers.ts](../blob/main/UI/src/routes/admin/workbench/save-helpers.ts), `entity-store.svelte.ts`) shrinks the blast radius of a **partial** save failure — a child-saver 400 no longer wipes every other pending record's edits, only the ones this save's diff couldn't fully settle.

That fix is reactive (recover gracefully after a failure). #2207's issue also proposed a complementary **preventive** direction that wasn't implemented there, since it's a distinct, non-trivial scope of its own:

> Block Save while a record carries a warning class the backend is known to hard-reject — the warnings already exist (`entityWarnings`/`sectionWarnings` in `validation.ts`); they just don't gate `SaveBar`'s Save button today (`disabled={total === 0 || saving}`, `SaveBar.svelte:18-26`).

## Suggested direction

- Every `warn(record)` on a section, and every `required` field, already knows it will produce a warning — but not all warnings are equally fatal. Some (e.g. a missing display name) are just sloppy; others (e.g. "at least one portion, all weights positive" on `entities/skill.ts:131-137`, or the same pattern in `progression/progression-store.svelte.ts:637-646`) are known to make the backend hard-reject the save outright.
- Introduce a way to mark a warning (or a whole section's `warn`) as **save-blocking** vs merely advisory, then have `SaveBar`'s Save button (or the caller computing its `disabled` prop) also check whether any pending (non-clean) record in the catalogue currently carries a blocking warning.
- This only prevents the *known* hard-reject triggers — it's not a substitute for #2207's rebase fix, since a save can still fail for reasons the client can't predict (network errors, a concurrent-edit conflict, a validation rule the client doesn't model). The two are complementary, not redundant.

Tags: enhancement, claude, scope: medium

PR #2218, opened 2026-07-20
Shrink Workbench partial-save failure blast radius to the unsettled diff

### LINKED ISSUE #2207: Workbench partial-save failure re-seeds the whole catalogue, silently discarding every other record's pending edits
"Save all" persists every pending record: the identity Add/Edit batch commits first, then per-record child savers run (`UI/src/routes/admin/workbench/save-helpers.ts:189-226`). When a child saver is rejected after the identity batch committed, `EntityStore.save` deliberately re-seeds from server truth (`UI/src/routes/admin/workbench/entity-store.svelte.ts:262-278`) — a documented anti-duplicate-Add measure. But the blast radius is the **whole catalogue**: every not-yet-posted pending edit on every *other* record is thrown away, leaving only an error toast.

It's easy to hit because the Save button is only gated on `total === 0 || saving` (`SaveBar.svelte:18-26`) — never on validation warnings, even where the config comments state the backend hard-rejects the same rule (e.g. `entities/skill.ts:131-137`: "at least one portion, all weights positive"; same pattern in `progression/progression-store.svelte.ts:637-646`). Concrete scenario: save a skill whose portion weight is 0 alongside edits to four other skills → the identity batch for all five commits → `SetSkillPortions` 400s → the other four records' child edits vanish.

Two complementary directions (either or both):

1. **Block Save while a record carries a warning class the backend is known to hard-reject** — the warnings already exist; they just don't gate.
2. **Shrink the failure blast radius**: on `PersistFailedError`, re-seed the baseline only and rebase the not-yet-committed local deltas on top, instead of dropping them.

Current behavior is deliberate and self-consistent (the comment in `entity-store.svelte.ts` explains the anti-duplicate rationale), but it loses typed work in the primary authoring tool, which is worth fixing.

ginderjeremiah:
## Summary

"Save all" persists every pending Workbench record through one shared `persistEntity` call. When a child saver fails partway through (e.g. `SetSkillPortions` 400s on a zero-weight portion), the identity Add/Edit/Delete batch has already committed, so `EntityStore.save()`'s catch block re-seeds the **entire catalogue** from server truth — silently discarding every other record's still-unposted pending edit, not just the one that failed.

This PR shrinks that blast radius to only the records this save's diff couldn't fully settle:

- `persistEntity` (`save-helpers.ts`) now tracks, as it iterates child savers, which records had **fully settled** (identity plus every one of their own child savers) by the point a later failure hit, and attaches that (`fresh`, `idMap`, `settledIds`) to the thrown `PersistFailedError` as `recovery`. A delete has no child saver, so it's always settled the instant the primary batch commits.
- `EntityStore.save()` (`entity-store.svelte.ts`) uses `recovery`, when present, to **rebase** rather than blanket-reseed: a settled record takes the fresh server copy (remapped from its local negative id to the persisted one if it was an add); everything else — a sibling record this save never touched, or one whose write simply hadn't been reached yet — keeps its current local edit, now diffed against the refreshed baseline so a retry only needs to resend what's actually still pending.
- When `recovery` is absent (the post-commit refetch itself failed, so there's nothing to rebase against), the old full-reseed fallback still runs.
- Documented the new behavior in `docs/frontend-admin.md`'s save-orchestration bullet.

## Follow-up

The issue also proposed a complementary, preventive direction — blocking Save while a record carries a warning class the backend is known to hard-reject (e.g. the zero-weight-portion case). That's a distinct, non-trivial scope of its own (classifying warnings as blocking vs advisory, wiring `SaveBar`), so I split it out as #2217 rather than bundling it here.

## Test plan

- [x] Added unit tests in `save-helpers.test.ts` asserting `PersistFailedError.recovery.settledIds` correctly excludes the failing/not-yet-attempted records and includes an unrelated delete.
- [x] Added unit tests in `entity-store.test.ts` covering: a sibling record's pending edit survives a partial failure and re-diffs as `modified`; a settled add is remapped from its local id to the persisted id; an unsettled add stays pending under its remapped id.
- [x] `npm run test:unit:ci` — 3833/3833 passing.

Closes #2207

---
_Generated by [Claude Code](https://claude.ai/code/session_01F6eLa9NKjxiTLTauPJKC8F)_


UI/src/routes/admin/workbench/entity-store.svelte.ts
```diff
@@ -281,6 +289,63 @@ export class EntityStore<T extends Identified> {
...
+			}
+
+			// Added/modified but not yet fully written: keep the local edit, remapped onto its
+			// persisted id if this save's primary batch already assigned it one.
+			const rebasedRecord =
+				persistedId !== undefined && persistedId !== record.id ? { ...clone(record), id: persistedId } : clone(record);
+			rebased.push(rebasedRecord);
+			if (persistedId !== undefined) {
+				claimedIds.add(persistedId);
+			}
```
ginderjeremiah, line 355: **Should-fix (narrow edge):** this branch keeps an unsettled *added* record with its **negative local id** when `persistedId === undefined`, which reintroduces the duplicate-Add risk the old full-reseed guarded against.

`persistedId` is `undefined` for an added record precisely when `idMap` (i.e. `resolveNewIds`) never mapped it. That's not only the pre-commit case — `persistEntity` can reach this rebase path with `committed === true` and `recovery` set if `resolveId` throws while building `childTargets` (`save-helpers.ts:236`): the primary Add batch committed, but the post-commit refetch returned fewer new records than were added (e.g. a concurrent delete by another admin, the same `#1856` class of anomaly). In that case the add *did* persist server-side, yet here it's retained as a local negative-id record → the next save diffs it as `added` again and re-Adds a duplicate.

The old code funneled every `PersistFailedError` through a full reseed, which dropped the orphaned local add and was safe against exactly this. Consider: an added record that couldn't be mapped to a persisted id should not be kept as a pending add here — either drop it, or fall back to a full reseed when any added record in the diff is unmapped, to preserve the anti-duplicate guarantee. A test mirroring the `resolveId`-throws path (committed add, empty `idMap`, `recovery` present) would lock this down.

---

ginderjeremiah (commented):
## Review: Shrink Workbench partial-save failure blast radius to the unsettled diff (#2207)

**Verdict:** Approve — real verdict stated here since GitHub won't accept a gating event on my own PR (I'm the author, so this goes out as `COMMENT`). No blocking findings; one non-blocking should-fix on a narrow edge, noted below.

This cleanly implements direction 2 of #2207 (shrink the blast radius) and correctly splits the preventive direction (block Save on known-hard-reject warnings) out to #2217 rather than bundling unrelated scope. The core idea — `persistEntity` reporting exactly which records fully settled (identity + every child saver) so `EntityStore` rebases only its own diff instead of blanket-reseeding — is the right shape, and the `settledIds`/`recovery` threading is careful (delete-settles-with-primary-batch, `settledIds` mutated by reference so the accumulating set is what the error carries).

### Should-fix

- `entity-store.svelte.ts:326-333` (inline) — the rebase keeps an unsettled *added* record with its negative local id when it has no `persistedId`. `persistEntity` can reach this path with `committed === true` (recovery present) if `resolveId` throws building `childTargets` (`save-helpers.ts:236`) — a committed add the refetch didn't map. The old code full-reseeded that case (safe); the rebase now retains a local add that a retry re-Adds as a duplicate. Narrow, but a genuine behavioral regression on that path. Details + suggested guard inline.

### Test coverage

Strong and appropriately targeted (frontend unit tests mirroring the source path, no battle logic in play):
- `save-helpers.test.ts` — `settledIds` excludes the failing and not-yet-attempted records, includes an unrelated delete, and is absent when the post-commit refetch itself fails.
- `entity-store.test.ts` — sibling pending edit survives and re-diffs as `modified`; settled add remapped local→persisted id; unsettled add kept pending under its remapped id.
- Gap: no test for the `resolveId`-throws path (committed add, unmapped id, `recovery` present) — that's the should-fix scenario above and would lock down whichever behavior you land on.

### What's good

- The doc update (`frontend-admin.md`) accurately describes the new rebase behavior and the delete/refetch-failure edge cases — no drift between comments, code, and docs.
- No house-rule violations (braces present, no null-forgiving in production code), constructor change is backward-compatible, and the explanatory comments genuinely earn their length given how subtle the settle/rebase reasoning is.

Not merging: CI is still running (`test-frontend` + the three e2e jobs in progress) and the should-fix above is worth a look first. Happy to re-review once it's addressed or if you decide the edge is acceptable and want to close it out with a comment.

---

ginderjeremiah:
Good catch — fixed in 7cef96c. `EntityStore` now guards the rebase with `diffAddsResolved`: if any currently-`added` record's local id isn't in `recovery.idMap`, it falls back to a full re-seed (from the recovery's own already-fetched list, no extra round-trip) instead of rebasing, so an unresolved committed add is dropped rather than kept pending under its local id. Added the test mirroring the `resolveId`-throws path you outlined, and updated the doc note.

---

PR #2219, opened 2026-07-20
Reflect the admin shell's active tool in the URL

### LINKED ISSUE #2208: Admin shell: active tool is ephemeral component state — no deep-linking, and a refresh always lands back on Enemies
The active admin tool lives only in local component state (`UI/src/routes/admin/+page.svelte:76` — `let active = $state('enemies')`; `handleNavigate` mutates it locally). The tool key never reaches the URL, so:

- a page refresh mid-session always resets to the Enemies workbench, and
- an admin can't link a collaborator to a specific tool (e.g. Content Health or the Progression editor).

Reflecting `active` in a query param (read on mount, written on navigate) is a contained change — the navigation-guard infrastructure this needs already exists (`confirmDiscard`, the popstate guard in `discard-guard.ts`), so a URL-driven switch can reuse the same unsaved-edits confirmation the in-page nav uses.

ginderjeremiah:
## Summary

The active admin tool (`UI/src/routes/admin/+page.svelte`) lived only in local component state (`let active = $state('enemies')`), so:

- a page refresh mid-session always reset to the Enemies workbench, and
- an admin couldn't link a collaborator to a specific tool (e.g. Content Health or the Progression editor).

This PR makes the `tool` query param the single source of truth for the active tool:

- `active` is now `$derived` off `page.url.searchParams.get('tool')` (falling back to `enemies` for a missing or unrecognized key) instead of being held as its own `$state` — so there's one source of truth rather than two to keep in sync.
- `handleNavigate` writes the new tool to the URL via a `replaceState` `goto()` (reusing the existing `confirmDiscard` unsaved-edits guard) instead of just reassigning local state. `replaceState` keeps today's Back/Forward semantics unchanged — a tool switch still doesn't push a history entry, so the existing popstate discard-guard (`discard-guard.ts`) needed no changes.
- The default tool (`enemies`) omits the `tool` param entirely, so a bare `/admin` link stays canonical.

## Test plan

- [x] Added `UI/src/tests/routes/admin/admin-tool-param.test.ts`: lands on the tool named by `?tool=`, falls back to the default for an unrecognized key, lands on a non-entity tool (Progression) via the param, and switching tools calls `goto` with the expected `replaceState` href (including omitting the param for the default tool).
- [x] Extended the shared `AdminSidebarStub` test stub to render a clickable nav button per tool (invoking `onNavigate`) so tool-switch tests can drive it without pulling in the real sidebar.
- [x] `npm run test:unit:ci` — 3833/3833 passing.

Closes #2208

---
_Generated by [Claude Code](https://claude.ai/code/session_01UVjrmbyqNHXcFq435pg2Fe)_

### DIFF
--- a/UI/src/routes/admin/+page.svelte
+++ b/UI/src/routes/admin/+page.svelte
@@ -51,6 +51,7 @@
 import { onMount } from 'svelte';
 import { beforeNavigate, goto } from '$app/navigation';
 import { resolve } from '$app/paths';
+import { page } from '$app/state';
 import { Loading } from '$components';
 import AdminLoadError from './AdminLoadError.svelte';
 import AdminSidebar from './AdminSidebar.svelte';
@@ -73,7 +74,17 @@ import { ensureAdminAccess } from './admin-access';
 import { confirmDiscard, createPopStateDiscardGuard, guardBeforeUnload } from './discard-guard';
 import { toastError } from '$stores';
 
-let active = $state('enemies');
+/** Nav key for the default landing tool; omitted from the URL to keep a bare `/admin` link canonical. */
+const DEFAULT_TOOL_KEY = 'enemies';
+const knownToolKeys = new Set(adminTools.map((tool) => tool.key));
+
+// The active tool is derived from the `tool` query param rather than held as its own local state, so
+// there's a single source of truth: a refresh or a shared deep link (e.g. `/admin?tool=paths`) lands on
+// the right tool, and switching tools (below) just writes the URL rather than two things to keep in sync.
+const active = $derived.by(() => {
+	const key = page.url.searchParams.get('tool');
+	return key && knownToolKeys.has(key) ? key : DEFAULT_TOOL_KEY;
+});
 let sidebarPinned = $state(false);
 // Gates rendering until the client-side admin-role check passes. Starts false so the server render
 // (no token available) shows nothing, and a non-admin who deep-links here is redirected before the
@@ -111,7 +122,12 @@ const handleNavigate = async (key: string) => {
 		return;
 	}
 	if (await confirmDiscard(workbenchDirty.total)) {
-		active = key;
+		// Reflects the tool switch in the URL (query param, not a new history entry — Back/Forward
+		// still exits /admin entirely, guarded by the popstate handler below) so the tool is
+		// deep-linkable and survives a refresh. The default tool omits the param, keeping a bare
+		// `/admin` link canonical.
+		const href = key === DEFAULT_TOOL_KEY ? resolve('/admin') : resolve(`/admin?tool=${key}`);
+		await goto(href, { replaceState: true, keepFocus: true, noScroll: true });
 	}
 };
 
--- a/UI/src/tests/routes/admin/admin-tool-param-resolve.test.ts
+++ b/UI/src/tests/routes/admin/admin-tool-param-resolve.test.ts
@@ -1,0 +1,20 @@
+import { describe, it, expect } from 'vitest';
+import { resolve } from '$app/paths';
+
+/*
+ * `+page.svelte`'s tool switch is the only `resolve(...)` call site in the codebase that passes a
+ * query string rather than a bare pathname (every other call site is `resolve('/admin')`,
+ * `resolve('/game')`, …). `admin-tool-param.test.ts` mocks `resolve` as an identity function, so it
+ * never exercises the real implementation — pinned here instead, against the real (unmocked)
+ * `$app/paths` module: `resolve_route` only substitutes bracketed `[param]` segments, so a literal
+ * `?query` suffix rides through a plain pathname segment unchanged (#2219 review).
+ */
+describe('resolve() with an appended query string', () => {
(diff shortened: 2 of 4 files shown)


UI/src/routes/admin/+page.svelte
```diff
@@ -111,7 +122,12 @@ const handleNavigate = async (key: string) => {
 		return;
 	}
 	if (await confirmDiscard(workbenchDirty.total)) {
-		active = key;
+		// Reflects the tool switch in the URL (query param, not a new history entry — Back/Forward
+		// still exits /admin entirely, guarded by the popstate handler below) so the tool is
+		// deep-linkable and survives a refresh. The default tool omits the param, keeping a bare
+		// `/admin` link canonical.
+		const href = key === DEFAULT_TOOL_KEY ? resolve('/admin') : resolve(`/admin?tool=${key}`);
```
ginderjeremiah, line 129: Nit (non-blocking): this is the codebase's only `resolve(...)` call that passes a query string — every other call site (`resolve('/admin')`, `resolve('/game')`, …) passes a bare pathname. It very likely works (the `?tool=…` rides along on the final path segment, since `resolve` only substitutes `[params]`), but the unit tests mock `resolve` as an identity function, so the real runtime behavior is never exercised.

Consider building the href from the live URL instead, which is the canonical SvelteKit idiom for a query-only change and also preserves any other params that might be on `/admin`:

```ts
const url = new URL(page.url);
if (key === DEFAULT_TOOL_KEY) {
	url.searchParams.delete('tool');
} else {
	url.searchParams.set('tool', key);
}
await goto(url, { replaceState: true, keepFocus: true, noScroll: true });
```

Either way is fine — just flagging that the query-through-`resolve` path isn't covered by a test.

---

ginderjeremiah (commented):
## Review: Reflect the admin shell's active tool in the URL (#2208)

**Verdict:** Approve — makes the `tool` query param the single source of truth for the active tool, cleanly solving the refresh-resets-to-Enemies and no-deep-linking problems from #2208. (Posted as a `COMMENT` because GitHub won't accept a gating verdict on my own PR — treat this as an approval.)

The change does exactly what the issue asked, reuses the existing `confirmDiscard`/popstate discard-guard infrastructure as intended, and collapses two-things-to-sync (`$state` + nav) into one `$derived`-off-URL source of truth. I checked the one real correctness risk — that `knownToolKeys` (built from `adminTools`) covers every key the sidebar can emit — and it holds: `adminTools` concatenates the entity tools, the Progression (`paths`) surface, and all three ops tools (`nav.ts:62`), so deep-linking or navigating to an ops tool won't silently fall back to Enemies. Default tool omits the param, keeping a bare `/admin` canonical, and `replaceState` preserves the existing Back/Forward-exits-/admin semantics.

### Nits

- `+page.svelte:129` — only query-string `resolve(...)` call in the codebase; unit tests mock `resolve` as identity so the real runtime isn't exercised. Suggested building the href from `new URL(page.url)` instead (canonical SvelteKit idiom, also preserves any other params). Non-blocking — inline comment has detail.
- A URL that explicitly carries the default `?tool=enemies` isn't normalized to a bare `/admin` on load (only on navigation). Purely cosmetic; not worth changing.

### Test coverage

Good. `admin-tool-param.test.ts` covers deriving the initial tool from the param, the unrecognized-key fallback, a non-entity tool (`paths`) via the param, the `replaceState` `goto` on switch, and the default-tool param omission. The `AdminSidebarStub` extension to render a clickable nav button per tool is a clean way to drive the switch without pulling in the real sidebar. Frontend-only change; backend suites correctly not re-run.

### What's good

- Single source of truth via `$derived` is the right call and matches the frontend guideline to prefer reactive derivations over parallel state.
- The comments explaining *why* `replaceState` (not a history push) and why the default omits the param are genuinely useful context, not noise.
- Scope discipline is spot-on — nothing gold-plated, guard infrastructure reused rather than reinvented.

---

UI/src/routes/admin/+page.svelte
ginderjeremiah: Good catch on the coverage gap — added `admin-tool-param-resolve.test.ts`, which imports the real (unmocked) `$app/paths` and pins `resolve('/admin?tool=items')` → `'/admin?tool=items'`. Traced `resolve_route`: it splits on `/` and only substitutes bracketed `[param]` segments via regex, so a literal `?query` suffix on a plain segment rides through untouched — confirmed at runtime, not just by the type-checker.

Kept the `resolve(...)` call rather than switching to `new URL(page.url)`, though: this repo's `svelte/no-navigation-without-resolve` lint rule is enabled (via `svelte.configs['flat/recommended']`) and is not type-aware here (no `parserOptions.project` configured for type-checked linting), so it only recognizes a `goto()` argument as safe when it's a literal `resolve(...)` call (or a ternary of same) at the call site. Passing a raw `URL` built from `page.url` reintroduces exactly the `Unexpected goto() call without resolve()` error I hit and fixed while building this — verified by trying it locally. Happy to revisit if the lint config itself changes, but for now `resolve()` is the only route through it.

---

PR #2220, opened 2026-07-20
Game.Core contract nits: name CombatRating's execute constant; fix RecordBattleCompleted's delta contract

### LINKED ISSUE #2209: Game.Core contract nits: CombatRating's execute pricing hardcodes an inline 0.5; RecordBattleCompleted's doc contract promises a per-battle delta but returns the cumulative dirty set
Two small, unrelated contract cleanups in the domain tier (bundled as micro-cleanups, same shape as #2139):

## 1. CombatRating's execute pricing uses an unnamed `0.5`

`Game.Core/Battle/CombatRating.cs:249-254`:

```csharp
var executeExpectation = isPlayer
    ? 1.0 + effectiveCaster.GetAttributeValue(ExecuteBonus) * 0.5
    : 1.0;
```

The `0.5` is a reference-profile assumption (average missing-health fraction over the reference fight — the same half-horizon logic as `RefFightDuration / 2`). Every sibling assumption the rating prices against (`RefToughness`, `RefDps`, `RefAttackRate`, `RefFightDuration`, `MaxMitigationCredit`) is a named, documented `ServerGameConstants` member, which is what the calibration-report workflow (#1533) re-tunes against. This one is invisible to that scheme: nothing documents it and a re-tune can't find it. Promote it to a named, documented constant.

## 2. `PlayerProgress.RecordBattleCompleted` returns the cumulative dirty set, not the documented per-battle delta

`Game.Core/Progress/PlayerProgress.cs:48-54`, `:142-145`: the doc comment says "On a freshly loaded aggregate (one is built per battle) those keys are exactly this battle's mutated statistics", and the implementation returns `[.. _dirtyStatistics]` — the aggregate-lifetime set. The offline path violates the freshness premise (one aggregate reused for a whole away window), so each call re-returns and re-copies the whole window's accumulated key set; `Game.Application/Services/OfflineProgressService.cs:316-330` carries a defensive comment explicitly acknowledging the mismatch. Today the superset is safe for challenge evaluation, but the contract is wrong for one of its two production callers and invites misuse by a third (anything that wants genuinely delta-scoped work).

Fix: either track and return the keys touched within the call, or reword the contract to "keys touched since load/AcceptChanges" and drop the defensive hedge at the call site.

ginderjeremiah:
## Summary

Two small, unrelated contract cleanups in the domain tier, as described in #2209:

1. **`CombatRating`'s execute pricing hardcoded an inline `0.5`.** `Game.Core/Battle/CombatRating.cs`'s `executeExpectation` term used a bare `0.5` for the reference average missing-health fraction. Every sibling reference-profile assumption (`RefToughness`, `RefDps`, `RefAttackRate`, `RefFightDuration`, `MaxMitigationCredit`) is a named, documented `ServerGameConstants` member the calibration-report workflow (#1533) can find and re-tune against — this one wasn't. Promoted it to `ServerGameConstants.RefMissingHealthFraction`, documented with the same half-horizon reasoning as `RefFightDuration`.

2. **`PlayerProgress.RecordBattleCompleted`'s doc contract promised a per-battle delta but returned the cumulative dirty set.** The doc comment said the returned keys are "exactly this battle's mutated statistics," but the implementation returned `[.. _dirtyStatistics]` — the whole aggregate-lifetime dirty set. That's only true while the aggregate is freshly loaded per battle; the offline path (`OfflineProgressService.ApplyOfflineRewards`) reuses one aggregate across an entire away window without calling `AcceptChanges` between battles, so each call was re-returning and re-copying the whole window's accumulated key set instead of just the keys it touched. The call site already carried a defensive comment acknowledging the mismatch.

   Fixed by tracking a dedicated `_touchedThisBattle` set, cleared at the start of each `RecordBattleCompleted` call and populated alongside `_dirtyStatistics` by the `Increment`/`SetMax`/`SetMin` write paths, so the return value is always this call's own delta. Updated the `OfflineProgressService` comment accordingly — the union of per-call deltas is still needed (a mixed-outcome window can touch a stat in a non-final battle), but it's no longer hedging against the method one day meaning what its doc comment already claimed.

## Test plan

- [x] Added `PlayerProgressTests.RecordBattleCompleted_SecondCallOnSameAggregate_ReturnsOnlyThatCallsOwnDelta` — reuses one aggregate across two `RecordBattleCompleted` calls (mirroring the offline path) and asserts a statistic touched only by the first call is absent from the second call's returned delta, even though it's still dirty.
- [x] `dotnet build` — clean.
- [x] `dotnet test --no-build --no-restore` — full suite green (Game.Core.Tests 1416, Game.Infrastructure.Tests 68, Game.Application.Tests 1036, Game.Api.Tests 835 — all passed).
- Frontend untouched by this change (both fixes are server-only domain code), so `npm run lint`/`npm run test:unit:ci` were not re-run.

Closes #2209

---
_Generated by [Claude Code](https://claude.ai/code/session_015CVXtRUReovxnM6RSVaK33)_

### DIFF
--- a/Game.Application/Services/OfflineProgressService.cs
+++ b/Game.Application/Services/OfflineProgressService.cs
@@ -314,11 +314,10 @@ namespace Game.Application.Services
             var victoryExpRewards = new List<int>();
             var proficiencyGains = new ProficiencyGainAccumulator();
             // Union the statistic rows touched across every battle, so the end-of-window challenge evaluation
-            // sees every moved statistic. RecordBattleCompleted currently returns the progress aggregate's
-            // cumulative dirty set (it is loaded once for the whole window), but unioning each call's result
-            // explicitly keeps this correct even if that method were ever changed to return only the per-battle
-            // delta — a mixed-outcome window can touch a tracked statistic in a non-final battle (e.g. a kill
-            // challenge crossed by early wins before a closing loss), and the union captures it regardless.
+            // sees every moved statistic. RecordBattleCompleted returns only this call's own per-battle delta
+            // (the progress aggregate is loaded once for the whole window), so a mixed-outcome window still
+            // needs the explicit union to keep a statistic touched by a non-final battle (e.g. a kill challenge
+            // crossed by early wins before a closing loss) in scope for the end-of-window evaluation.
             var touchedStatistics = new HashSet<(EStatisticType Type, int? EntityId)>();
             foreach (var battle in result.Battles)
             {
--- a/Game.Core.Tests/Progress/PlayerProgressTests.cs
+++ b/Game.Core.Tests/Progress/PlayerProgressTests.cs
@@ -1184,6 +1184,27 @@ namespace Game.Core.Tests.Progress
         }
 
         [Fact]
+        public void RecordBattleCompleted_SecondCallOnSameAggregate_ReturnsOnlyThatCallsOwnDelta()
+        {
+            // The offline path reuses one aggregate across a whole away window without AcceptChanges between
+            // battles (#2209), so the touched set returned by each call must be that battle's own delta —
+            // not the aggregate's whole dirty set accumulated since load.
+            var progress = MakeProgress();
+            var enemy = MakeEnemy(id: 5);
+
+            var firstTouched = progress.RecordBattleCompleted(enemy, victory: true, playerDied: false,
+                totalMs: 1000, new BattleStats { CriticalHits = 2 }, isBossBattle: false, zoneId: 0);
+            var secondTouched = progress.RecordBattleCompleted(enemy, victory: true, playerDied: false,
+                totalMs: 1000, new BattleStats(), isBossBattle: false, zoneId: 0);
+
+            // CriticalHits was touched by the first, crit-having battle and stays dirty (unsaved) into the
+            // second call, but the second, crit-less battle never touched it and must not re-report it.
+            Assert.Contains((EStatisticType.CriticalHits, (int?)null), firstTouched);
+            Assert.Contains(progress.DirtyStatistics, s => s.Type == EStatisticType.CriticalHits);
+            Assert.DoesNotContain((EStatisticType.CriticalHits, (int?)null), secondTouched);
+        }
+
+        [Fact]
         public void EvaluateChallenges_MarksNewAndChangedChallengesDirty_ButNotSkippedCompletedOnes()
         {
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: Game.Core contract nits — name `CombatRating`'s execute constant; fix `RecordBattleCompleted`'s delta contract (#2209)

**Verdict:** Approve — both cleanups are correct, well-documented, and the delta fix is verified sound. _(Submitted as a comment because I'm the PR author; treat it as an approval.)_

Both changes land exactly what #2209 asked for.

**1. `0.5` → `ServerGameConstants.RefMissingHealthFraction`** — pure promotion of the inline reference-profile assumption to a named, documented constant (value unchanged at `0.5`), sitting alongside its `RefFightDuration`/`MaxMitigationCredit` siblings so the calibration workflow (#1533) can find and retune it. The doc comment's half-horizon reasoning mirrors `RefFightDuration`. Correct.

**2. `RecordBattleCompleted` now returns the true per-battle delta** — I verified the core correctness concern: the new `_touchedThisBattle` set is maintained in exact lockstep with `_dirtyStatistics`. Every statistic write in this class routes through `Record` → `Increment`/`SetMax`/`SetMin`, which are the only three places either set is mutated, and each now adds to both. All `Record`/`RecordKillByDamageType` call sites live inside `RecordBattleCompleted`'s tree, and the set is cleared at that method's start — so there are no stray writes between battles and the returned collection is always this call's own delta. Neither production caller regresses: the online path builds a fresh aggregate per battle (delta == full dirty set there), and the offline path still unions each call's delta, which the updated comment now justifies on its own terms rather than as a hedge against the doc contract.

### Test coverage
Good. `RecordBattleCompleted_SecondCallOnSameAggregate_ReturnsOnlyThatCallsOwnDelta` reproduces the offline reuse scenario precisely — a `CriticalHits`-touching first battle followed by a crit-less second battle on the same aggregate — and asserts the stat stays in `DirtyStatistics` yet is absent from the second call's delta. That's the exact behavior the old `[.. _dirtyStatistics]` return got wrong, so it would fail pre-fix and locks in the contract. The `CombatRating` change is a value-preserving rename covered by the existing combat-rating suite.

### What's good
- The `_touchedThisBattle` field mirrors the existing `_dirtyStatistics` field pattern rather than threading a local through the write helpers — consistent with the class and low-risk.
- The comments are appropriately scoped: the constant carries its calibration rationale, and the `OfflineProgressService` comment now explains why the union is still needed on its own merits.

No blocking items, should-fixes, or nits. Approving; will merge once CI is green.

---

PR #2221, opened 2026-07-20
Block Workbench Save while a record carries a known hard-reject warning

### LINKED ISSUE #2217: Workbench: block Save while a record carries a warning class the backend is known to hard-reject

ginderjeremiah:
## Summary

Follow-up to #2207 / PR #2218's reactive (rebase-on-partial-failure) fix. This implements the complementary **preventive** direction #2217 proposed: block the Workbench's Save button while a pending record carries a validation warning the backend is *known* to hard-reject, instead of only recovering gracefully after the fact.

- `entities/types.ts`: adds a `Warning` shape (`{ message: string; blocking?: boolean }`). `BaseSection.warn` may now return a plain string (always advisory, unchanged) or a `Warning` to flag a condition as save-blocking.
- `validation.ts`: adds `sectionBlockingWarning`/`entityBlockingWarnings`, which extract only the `blocking`-flagged messages, independent of the existing `sectionWarnings`/`entityWarnings` (which keep showing every warning, advisory or not, in the triangle/list UI).
- `entity-store.svelte.ts`: `RecordState` gains `blockingWarnings`; `EntityStore` exposes `hasBlockingWarnings` (true while any *pending* — added/modified — record has one; a delete has no child saver left to reject, so it doesn't gate) and `save()` now refuses to run while it's true.
- `SaveBar.svelte`: new optional `blocked` prop disables Save and shows why. Defaults to `false` so the progression editor's `SaveBar` usage (which has its own, separate advisory-only warning system) is unaffected.
- `WorkbenchDetail.svelte`: wires `store.hasBlockingWarnings` into the new prop.

### Which warnings are now blocking

I cross-referenced every existing `warn()` in `entities/*.ts` against its actual `AdminTools` backend repository to confirm the condition really is hard-rejected (not just "some validation exists nearby") before flagging it:

| Entity / section | Condition | Backend rejection |
|---|---|---|
| `skill.ts` → portions | empty set, or any non-positive weight | `AdminSkills.SetPortions` |
| `skill-recipe.ts` → result | non-Synthesis or retired result skill | `AdminSkillRecipes` (`FindResultSkillViolation`) |
| `skill-recipe.ts` → inputs | empty set, or input == result | `AdminSkillRecipes.SetInputs` |
| `skill-recipe.ts` → conditions | `minLevel < 1` | `AdminSkillRecipes` |
| `class.ts` → starterEquipment | item/slot category mismatch | `AdminClasses` (`FindStarterEquipmentViolation`) |
| `zone.ts` → identity | dedicated boss lost its boss flag | `AdminZones.SaveZones` |
| `item.ts` → identity | weapon missing type/skill, or non-weapon with a type | `AdminItems` (`FindWeaponInvariantViolation`) |
| `challenge.ts` → condition | DamageType statistic left Global; boss-only target lost boss flag | `AdminChallenges` (`FindTargetViolation`) |
| `lesson.ts` → identity | trigger-type/mechanic-event mismatch | `AdminLessons` (`FindTriggerViolation`) |
| `lesson.ts` → steps | duplicate step ordinal | `ChildCollectionReconciler` → `ChangeSetProcessor.DuplicateFailure` |

Everything else — every `required`-field check, "no starter skills", "no skills assigned", "no reward", a non-positive challenge goal, an empty attribute/multiplier/spawn table, etc. — stays advisory, since I confirmed the corresponding backend endpoint accepts an empty/zero value fine. The `challenge.ts` condition section and `lesson.ts` steps section each mix an advisory case with blocking ones in the same `warn`, which is exactly why `Warning` is per-message rather than a single section-level flag.

## Test plan

- [x] Updated existing entity tests (`skill`, `skill-recipe`, `class`, `zone`, `item`, `challenge`, `lesson`) to expect `{ message, blocking: true }` where the warn is now blocking; advisory warns are untouched.
- [x] Added `validation.test.ts` coverage for `sectionBlockingWarning`/`entityBlockingWarnings`, including a mixed fields-section case.
- [x] Added `entity-store.test.ts` coverage: a blocking warning surfaces in both `warnings` and `blockingWarnings`, `hasBlockingWarnings` toggles correctly, `save()` refuses to call `persist` while blocked, and a deleted record with a stale blocking warning does *not* gate save.
- [x] `npm run test:unit:ci` — 3851/3851 passing.

Closes #2217

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kekz5p2sMCXrcbdnLb2xgx)_


ginderjeremiah (commented):
## Review: Block Workbench Save while a pending record carries a known hard-reject warning (closes #2217)

**Verdict:** Approve *(submitted as a comment — I'm the PR author, so GitHub won't take a gating review from me).* The change cleanly implements #2217's preventive direction: a per-message `Warning` shape, a blocking-only extraction path independent of the advisory display, a store-level `save()` gate, and a `SaveBar` affordance. Well-scoped, well-tested, and the docs are updated.

### What's good

- **Correct classification, verified against the backend.** I spot-checked the table in the description against the actual `AdminTools` repositories: `AdminSkills.SetPortions` does `Failure` on non-positive weight *and* zero portions (both flagged blocking), and `AdminChallenges` has **no** range-check on `ProgressGoal` (correctly left advisory) while `FindTargetViolation` rejects the DamageType-target / boss-flag drift (correctly blocking). The per-*message* granularity — rather than a section-level flag — is the right call given `challenge.ts`/`lesson.ts` mix an advisory case with blocking ones in a single `warn`.
- **Advisory display is untouched.** `sectionWarnings`/`entityWarnings` still surface every warning in the triangle/list; `entityBlockingWarnings` is a separate, additive path. Good separation.
- **Deletes correctly don't gate.** `hasBlockingWarnings` keys on `added`/`modified` only — a delete has no child saver left to reject — and there's a dedicated test for the stale-warning-on-a-delete case.
- **Defense in depth.** Gating both `SaveBar`'s `disabled` and `EntityStore.save()` means a programmatic `save()` can't slip past the disabled button. Consistent with the existing `total === 0`/`saving` guards.
- **Theming is consistent** — reuses `WarnTriangle` and the `var(--warning)` token rather than hardcoding.

### Test coverage

Thorough and following the project's rules: `validation.test.ts` covers `sectionBlockingWarning`/`entityBlockingWarnings` (blocking, advisory string, cleared, mixed fields-section); `entity-store.test.ts` covers surfacing in both `warnings` and `blockingWarnings`, the `hasBlockingWarnings` toggle, `save()` refusing to `persist` while blocked, and the delete-doesn't-gate case; every affected entity test is updated to the new `{ message, blocking }` shape with clear advisory-vs-blocking labels. No battle logic touched, so no cross-suite mirroring needed.

### Follow-up (not blocking)

- **Progression editor is scoped out** despite #2217 naming it. This is a defensible call — it uses a different warning representation (`progression-helpers.ts` returns `string[]`, not a `warn` predicate returning `Warning`) and its own save orchestration, so it's a distinct chunk of work, and the `SaveBar.blocked` prop was clearly designed as the extension point for it. Tracked as **#2222** so "closes #2217" doesn't lose the second half.

### Nit (optional)

- `sectionWarnings` and `sectionBlockingWarning` each call `section.warn(rec)` independently, so `recordStates` invokes each section's `warn` twice per recompute. `warn` predicates are cheap and the result is memoised per-record, so this is negligible — noting only for completeness.

Nothing here blocks. Merging.

---

Issue #1827, opened 2026-07-10
Combined lower-MaxLevel + prune out-of-range payout now requires two saves (Workbench save ordering)


ginderjeremiah:
## Context

Follow-up from PR #1803 (closed #1777), which added a backend guard in `SaveProficiencies` (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs`) rejecting an identity `Edit` whose new `MaxLevel` falls below the highest level of an already-persisted modifier/reward.

## Problem

The guard evaluates against the **pre-save** cached child rows, and the Workbench posts the identity `Edit` (`AddEditProficiencies`) *before* the child setters (`SetProficiencyModifiers`/`SetProficiencyRewards`). So an operator who, in a single save, both lowers a proficiency's `MaxLevel` **and** removes the offending out-of-range modifier/reward row (a valid combined end state) is now rejected on the first POST and forced into a two-save dance:

1. Save the modifier/reward removal first.
2. Lower the `MaxLevel` in a second save.

This mirrors the exact class of two-save UX problem that PR #1802 fixed for the progression editor's retire + prerequisite-removal flow.

This was a deliberate, documented scope deferral in #1803 (the end state stays data-consistent and #1777 scoped it out), so it is not a data-integrity defect — purely a UX regression on a previously one-shot flow.

## Suggested fix (pick one)

- Reorder the Workbench proficiency save so child modifier/reward deletions post before the identity `Edit` (the shape #1802 used for paths), so the guard sees the post-removal child set. Preserve batch cycle-safety within each group.
- Or promote the editor's advisory out-of-range warning to a blocking one (the "and/or editor" half of #1777's fix shape), so the operator is told up front rather than hitting a backend rejection.

Prefer the reorder — it keeps the one-shot combined save working.

## Acceptance

- Lowering `MaxLevel` and removing the now-out-of-range payout in the same save succeeds.
- The silent-save defect closed by #1803 stays closed (backend remains authoritative).
- Test coverage for the combined single-save flow.

PR #2223, opened 2026-07-20
Gate progression editor Save on out-of-range payout levels (#2222)

### LINKED ISSUE #2222: Progression editor: gate Save on backend-hard-reject warnings (extend #2217's blocking-warning mechanism)
## Context

Follow-up to #2217 / PR #2221. That PR introduced a save-blocking `Warning` (`{ message, blocking? }`) for the config-driven Workbench `EntityStore`: a `warn` predicate can flag a condition the backend is *known* to hard-reject, `EntityStore.hasBlockingWarnings` gates `save()`, and `SaveBar` gained a `blocked` prop (defaulting to `false`) that disables Save and surfaces why.

That work deliberately scoped out the **progression editor** (`workbench/progression`), even though #2217 explicitly named it (`progression-store.svelte.ts` / `progression-helpers.ts`) as carrying the same pattern. The `SaveBar` `blocked` prop was built as the reusable extension point for exactly this, but the progression editor still passes no `blocked` value, so its Save is never gated.

## Why it wasn't done in PR #2221

The progression editor uses a **different** warning representation — `pathWarnings`/`proficiencyWarnings` in `progression-helpers.ts` return a flat `string[]`, not a section `warn` returning a `Warning` — and it re-implements its own save orchestration (`ProgressionStore.save`). Wiring blocking through it is a self-contained but distinct chunk of work, not a mechanical reuse of the Workbench path.

## Scope

- Introduce a blocking/advisory distinction into `progression-helpers.ts` for the conditions the backend genuinely hard-rejects. Confirmed backend hard-rejects (do **not** flag the merely-advisory ones like "Missing name"):
  - Proficiency `MaxLevel < 1` and non-positive XP curve (`baseXp`/`xpGrowth`) — rejected by `AdminProficiencies`.
  - Modifier/reward level out of range (`> MaxLevel`) — the `FindShrunkenMaxLevelViolation` family; see also the related open issues about the one-save ergonomics of that guard.
- Compute a `hasBlockingWarnings` on `ProgressionStore` (mirroring `EntityStore`), gate its `save()`, and pass `blocked` into the shared `SaveBar` in `Progression.svelte`.
- As with #2221, mark a condition `blocking` only once its exact trigger is confirmed against the backend's own rejection.
- Add `progression-store` / helper tests: a blocking condition disables Save and `save()` no-ops while it holds; advisory-only warnings do not gate.

## Notes

This is preventive gating only — complementary to, not a replacement for, the partial-failure rebase recovery (#2207). Several existing progression-editor issues touch the same `MaxLevel`/out-of-range area; coordinate so the blocking classification lines up with whatever backend guard changes land there.

Tags: enhancement, claude, scope: medium

ginderjeremiah:
## Summary

Extends #2217/#2221's save-blocking mechanism — `EntityStore.hasBlockingWarnings` gating the generic Workbench's `SaveBar` — to the **progression editor** (`workbench/progression`), which was explicitly named in #2217 but deliberately scoped out of #2221 because it uses a different warning representation (`pathWarnings`/`proficiencyWarnings` return a flat `string[]`, and re-implements its own save orchestration).

- `progression-helpers.ts`: extracted the existing out-of-range modifier/reward-level check into `levelRangeWarnings` and exported it as `proficiencyBlockingWarnings`.
- `ProgressionStore` gained a `hasBlockingWarnings` derived (mirroring `EntityStore`) that's true while any pending (added/modified) tier carries an out-of-range payout, and `save()` now no-ops while it holds.
- `Progression.svelte` passes `blocked={store.hasBlockingWarnings}` into the shared `SaveBar`.
- `TierDetail.svelte`'s milestone-tab warning indicator now reads `proficiencyBlockingWarnings` directly instead of string-matching `proficiencyWarnings()` output for `"out of range"`.

## Deviation from the issue's stated scope

The issue asked to also flag **`MaxLevel < 1`** and a **non-positive XP curve** (`baseXp`/`xpGrowth`) as blocking, citing `AdminProficiencies` as the backend enforcer. I checked the backend before wiring that up (`AdminProficiencies.cs`, `Contracts/Proficiency.cs`, `ProgressionGraphChecker.cs`) and **neither condition is actually rejected server-side** — both fields are plain `int`/`decimal` with no data annotations, and `SaveProficiencies` only runs `FindPathViolation`, `FindTierOrdinalCollision`, and `FindShrunkenMaxLevelViolation`. A `MaxLevel` of 0 or a non-positive XP curve is accepted as-is today.

So I only wired blocking for the condition that **is** genuinely backend-enforced — an out-of-range modifier/reward level, via `AdminProficiencies.FindShrunkenMaxLevelViolation` (the identity edit, against a still-persisted payout) and `FindLevelOutOfRange` (`SetModifiers`/`SetRewards`, against the tier's saved `MaxLevel`). `MaxLevel < 1` and the XP-curve check stay advisory-only, same as before — gating Save on them would block a save the backend would actually accept, which is exactly the failure mode #2217 was designed to avoid causing.

## Test plan

- [x] `proficiencyBlockingWarnings` unit tests: flags only the out-of-range payout, not `MaxLevel`/XP-curve nags.
- [x] `ProgressionStore.hasBlockingWarnings` unit tests: true only while a pending tier carries the condition; false for `MaxLevel < 1` / non-positive XP curve; `save()` no-ops (no network call) while blocked.
- [x] Updated the existing "lowering MaxLevel without removing the offending payout" save-orchestration test — it's now blocked pre-flight instead of round-tripping to the backend and getting rejected.
- [x] `npm run lint` (eslint + svelte-check) passes.
- [x] `npm run test:unit:ci` — full suite passes.

Closes #2222

---
_Generated by [Claude Code](https://claude.ai/code/session_01GFHuRvbFDWoPWNPxKRzELm)_

### DIFF
--- a/UI/src/routes/admin/workbench/SaveBar.svelte
+++ b/UI/src/routes/admin/workbench/SaveBar.svelte
@@ -42,8 +42,8 @@ interface Props {
 	modified: number;
 	deleted?: number;
 	/** True when a pending record carries a save-blocking warning — disables Save alongside the
-	 *  existing total/saving gates and surfaces why. Defaults to false for callers with no concept
-	 *  of blocking warnings (e.g. the progression editor's own advisory-only warnings). */
+	 *  existing total/saving gates and surfaces why. Defaults to false for a caller with no concept
+	 *  of blocking warnings at all. */
 	blocked?: boolean;
 	onDiscard: () => void;
 	onSave: () => void;
--- a/UI/src/routes/admin/workbench/progression/Progression.svelte
+++ b/UI/src/routes/admin/workbench/progression/Progression.svelte
@@ -69,6 +69,7 @@
 			saving={store.saving}
 			added={store.counts.added}
 			modified={store.counts.modified}
+			blocked={store.hasBlockingWarnings}
 			onDiscard={() => store.discard()}
 			onSave={() => store.save()}
 			saveTestId="progression-save"
--- a/UI/src/routes/admin/workbench/progression/TierDetail.svelte
+++ b/UI/src/routes/admin/workbench/progression/TierDetail.svelte
@@ -68,7 +68,7 @@
 import { childChanged } from '../save-helpers';
 import { referenceSourcesFromStatic, retireWithConfirm } from '../retire-confirm';
 import type { ProgressionStore, TierTab } from './progression-store.svelte';
-import { payoutLevels, proficiencyWarnings } from './progression-helpers';
+import { payoutLevels, proficiencyBlockingWarnings } from './progression-helpers';
 import DetailHeader from '../components/DetailHeader.svelte';
 import ConlangIdentity from './ConlangIdentity.svelte';
 import XpCurve from './XpCurve.svelte';
@@ -111,9 +111,7 @@ const identityWarn = $derived(
 			!tier.iconPath.trim())
 );
 const xpWarn = $derived(!!tier && (tier.maxLevel < 1 || !(tier.baseXp > 0) || !(tier.xpGrowth > 0)));
-const milestoneWarn = $derived(
-	!!tier && proficiencyWarnings(tier).some((w) => w.includes('level') && w.includes('out of range'))
-);
+const milestoneWarn = $derived(!!tier && proficiencyBlockingWarnings(tier).length > 0);
 const milestonesDirty = $derived(
 	!!baseline &&
 		(childChanged(tier?.levelModifiers, baseline.levelModifiers) ||
--- a/UI/src/routes/admin/workbench/progression/progression-helpers.ts
+++ b/UI/src/routes/admin/workbench/progression/progression-helpers.ts
@@ -157,6 +157,30 @@ export const pathWarnings = (path: WorkbenchPath): string[] => {
 	return warnings;
 };
 
+/**
+ * An out-of-range modifier/reward level, checked against the tier's own (about-to-be-saved) MaxLevel.
+ * This is the one proficiency condition the backend genuinely hard-rejects a save over — via
+ * AdminProficiencies.FindShrunkenMaxLevelViolation (the identity edit, against a still-persisted
+ * payout) or FindLevelOutOfRange (SetModifiers/SetRewards, against the tier's saved MaxLevel) — so it
+ * doubles as {@link proficiencyBlockingWarnings}, unlike the rest of {@link proficiencyWarnings}.
+ */
+const levelRangeWarnings = (prof: WorkbenchProficiency): string[] => {
+	const warnings: string[] = [];
+	for (const modifier of prof.levelModifiers) {
+		if (modifier.level < 0 || modifier.level > prof.maxLevel) {
+			warnings.push(`Modifier level ${modifier.level} out of range`);
+			break;
+		}
+	}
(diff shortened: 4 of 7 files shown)


ginderjeremiah (commented):
## Review: Gate progression editor Save on out-of-range payout levels (#2222)

**Verdict:** Approve (posted as a comment — GitHub won't let the PR author self-approve; the intended verdict is **Approve**) — the change is correct, well-scoped, and the deviation from the issue is properly justified. One nit only.

The most important thing here is the **deviation from the issue**, and it's the right call. The issue asserted that `AdminProficiencies` hard-rejects `MaxLevel < 1` and a non-positive XP curve. I verified this against the backend myself and the issue is **wrong**:

- `AdminProficiencies.SaveProficiencies` runs only `FindPathViolation`, `FindTierOrdinalCollision`, and `FindShrunkenMaxLevelViolation` — none touch `MaxLevel`/`BaseXp`/`XpGrowth`.
- `Contracts.Proficiency` carries no data annotations; those are plain `int`/`decimal` fields.

So gating Save on them would block a save the backend would actually accept — exactly the failure mode #2217 was built to avoid. Wiring blocking only for the genuinely-enforced out-of-range condition is the correct scoping, and the code comments document the reasoning at the right altitude.

I also checked the harder consistency question — could the new gate **over-block** a save the backend accepts? It can't: the range bounds mirror the backend exactly (`FindLevelOutOfRange` min 0 for modifiers / min 1 for rewards), and the "lower MaxLevel + remove the offending payout in one save" case is already handled by `save()`'s step‑2 `shrinksPastPersistedPayout` early-removal, so the pending-vs-pending gate and the backend land on the same answer. The complementary one-save-ergonomics edge (#1827/#1804) is explicitly out of scope and untouched.

### Nits

- `SaveBar.svelte:44-46` (not in this diff) — the `blocked` prop's doc comment cites *"the progression editor's own advisory-only warnings"* as an example of a caller with **no** concept of blocking warnings. This PR is precisely what gives the progression editor a blocking concept, so that parenthetical is now stale. Worth a one-line update to a different example (or dropping the parenthetical). Not merge-blocking.

### Test coverage

Solid and follows the project's rules:
- `proficiencyBlockingWarnings` helper test asserts it flags only the out-of-range payout and ignores the `MaxLevel`/XP-curve advisories.
- `ProgressionStore.hasBlockingWarnings` tests cover true/false transitions, the non-blocking advisory case, and `save()` no-op while blocked (no network call).
- The existing "lowering MaxLevel without removing the payout" test was correctly retargeted from "backend rejects" to "blocked pre-flight, no network call."

Minor, optional: the helper test exercises the modifier branch of the shared `levelRangeWarnings`; a reward-level assertion via `proficiencyBlockingWarnings` would round it out, but the two branches are symmetric so this is genuinely take-it-or-leave-it.

### What's good

- Verified the issue's backend claim instead of trusting it — caught a real inaccuracy and scoped around it (exactly what CLAUDE.md asks for).
- Clean DRY extraction of `levelRangeWarnings`, with `proficiencyWarnings` composing it so advisory display and blocking gate can't drift apart.
- Good separation of concerns: `ProgressionList` keeps the full advisory list, store + `TierDetail` use the blocking subset.


---

PR #2224, opened 2026-07-20
Re-check ban/roles on refresh and player selection, not just login

### LINKED ISSUE #1762: Banning (or revoking roles from) a user never ends their live session — refresh chain is indefinitely renewable and never re-checks account state
## Problem

The ban check lives only in `Login` (`Game.Application/Services/AccountService.cs:114-117`), and `docs/backend-admin.md` documents ban as "enforced at login" — but a live session never needs to log in again:

- `AccountService.cs:297-306` — `Refresh` consumes the old token and re-issues purely from the stored `RefreshTokenData` (`UserId`, `Roles`, `PlayerId`); no DB read of the user.
- `AccountService.cs:325-330` — every rotation issues a brand-new refresh token with a full 48h TTL, so the chain never expires while the client stays active.
- `Game.DataAccess/Repositories/Users.cs:297-300` — `BanUser` only sets `user.BannedAt`; no token/session invalidation.
- `Game.DataAccess/Repositories/RefreshTokenStore.cs:45-49` — tokens are keyed in Redis solely by SHA-256 of the token value; there is no per-user index, so revoking a user's outstanding tokens isn't even possible today.

## Failure scenario

An admin bans an abusive account (or strips someone's Admin role via `SetUserRoles`); the target's open client keeps refreshing, each refresh mints a fresh 48h token carrying the **original** roles snapshot, and the user keeps playing — or keeps hitting admin endpoints (`AdminRoleAuthorizationFilter` trusts the token's role claim alone) — indefinitely, as long as they refresh at least once per 48h.

## Fix shape

In `Refresh` (and `SelectPlayer`'s consume path), load the user and reject when banned/archived, and re-derive roles from the DB instead of echoing the consumed token's snapshot — one indexed read per ≤15-min refresh, negligible cost. Optionally add a per-user token index in Redis so a ban can revoke outstanding tokens immediately. Update `docs/backend-admin.md`'s "enforced at login" claim to match whichever enforcement point lands.

ginderjeremiah:
## Summary

A banned user (or one with a revoked role) never needed to log in again to keep playing: `Refresh` and `SelectPlayer` both re-issued token pairs straight from the refresh token's stored snapshot (`RefreshTokenData.Roles`), never re-checking the account's live state. As long as the client kept refreshing at least once per 48h refresh-token lifetime, an admin's ban or role revocation would never actually take effect.

- Added `IUsers.GetAccountState(userId)` — the live ban/role read, keyed by user id (mirrors `GetUser`'s username-keyed login check, filtered the same way to exclude archived accounts).
- `AccountService.Refresh` now loads the account's live state after consuming the token and returns `null` (ending the chain) if it's banned or archived, issuing the reissued pair with freshly-derived roles otherwise.
- `AccountService.SelectPlayer` runs the same check *before* consuming the token (alongside the existing ownership/player-load checks), so a rejected selection still leaves the caller's token intact — consistent with how ownership failures are handled today.
- `RefreshTokenData.Roles` is removed (and `IRefreshTokenStore.Issue`'s `roles` parameter with it) — once every reissue re-derives roles from the DB, the cached copy was dead weight nothing read back.
- Updated `docs/backend.md`, `docs/backend-admin.md`, and `docs/backend-auth.md`, which previously documented "a role change/ban only takes effect at next login" as the deliberate model. That claim is now the bug this issue asked to fix; latency is bounded by the 15-minute access-token TTL instead of the 48-hour refresh-token lifetime.

## Scope note

The issue also suggested an *optional* stretch — a per-user Redis token index so a ban revokes **outstanding** access tokens immediately, rather than waiting out their 15-minute TTL. That's a genuinely separate design decision (new Redis structure, invalidation-on-ban wiring) so I've left it out of this PR and will file a follow-up issue for it.

## Test plan

- [x] `dotnet test` (full solution) — 3,359 passed, 0 failed
- [x] New integration tests in `AccountServiceIntegrationTests`:
  - `Refresh_BannedSinceIssue_ReturnsNullAndEndsTheChain`
  - `Refresh_RoleGrantedSinceIssue_ReissuesTokensWithTheNewRole`
  - `SelectPlayer_BannedSinceLogin_IsRejectedAndLeavesTokenIntact`
  - `SelectPlayer_RoleGrantedSinceLogin_ReissuesTokensWithTheNewRole`

Closes #1762

---
_Generated by [Claude Code](https://claude.ai/code/session_01CVmv2qwKCMCLqiWtNj8Nyz)_


Game.Abstractions/Contracts/Identity/AccountState.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// The live ban/role state of an account, keyed by user id rather than username (unlike
+    /// <see cref="AccountCredentials"/>, which authenticates a login). Used to re-check an account between
+    /// logins — on refresh and player selection — so a ban or role change lands there instead of only at
+    /// the account's next login.
+    /// </summary>
+    public class AccountState
+    {
+        public int Id { get; set; }
```
ginderjeremiah: Nit: `Id` is never read by either caller — both `Refresh` and `SelectPlayer` already hold the `userId` they looked this up by, and only touch `Roles`/`IsBanned`. Unlike the sibling `AccountCredentials` (where `Id` is used at login, since login is keyed by username and must resolve the id), this contract is keyed by id, so carrying it back is dead data. Consider dropping the property (and the `Id = u.Id` line in `ToAccountState`) to keep the contract to what callers actually consume. Minor — take it or leave it.

---

ginderjeremiah (commented):
## Review: re-check ban/roles on `Refresh` and `SelectPlayer`, not just login (fixes #1762)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the fix squarely addresses the reported hole: a ban or role change is now re-checked against live account state on every refresh/selection instead of riding the token's stored snapshot, bounding effect latency to the 15-min access-token TTL rather than the 48h refresh chain. No blocking or should-fix findings.

### What's good

- **Correct fix at the right layer.** `IUsers.GetAccountState` returns a `Contracts.Identity` read type (not an entity), with the projection in `Game.DataAccess` — consistent with the "repository reads speak contracts" rule in `backend-admin.md`. Application layer only orchestrates; no domain logic leaked into the data tier or controller.
- **Ordering is deliberate and consistent with existing semantics.** In `SelectPlayer` the ban check runs *before* `Consume`, so a rejected selection leaves the token intact — matching how ownership failures already behave. In `Refresh`, consume-then-check ends the chain (token burned) on a banned account. Both correctly deny access; the asymmetry is intentional and the tests document it.
- **Dead weight removed.** Dropping `RefreshTokenData.Roles` and the `roles` param on `IRefreshTokenStore.Issue` is the right call — once every reissue re-derives roles, the cached copy was unread. Grep confirms no stragglers.
- **Docs updated thoroughly** across `backend.md`, `backend-admin.md`, and `backend-auth.md`, correctly reframing the old "only at next login" model and honestly calling out the one residual gap (an already-issued access token inside its TTL).

### Nits (non-blocking)

- `AccountState.Id` is never consumed by either caller (see inline). Minor dead data on a new contract.
- The `account is null` branch (archived / no-longer-exists) in both `Refresh` and `SelectPlayer` isn't directly exercised — the banned tests hit `IsBanned`, not the null path. It's a trivial OR, but a one-line archived-account test would close it.

### Test coverage

Solid. Integration tests cover both flows × both scenarios (banned-since-issue → chain ends / rejected with token intact; role-granted-since-issue → reissued pair carries the fresh role), and the new `RecordingAccessTokenService` cleanly asserts the *reissued* roles rather than the login snapshot. Integration (not unit) is the right level here since the logic is orchestration over Postgres/Redis. No frontend/battle-logic surface touched, so no mirrored-test obligation.

### Follow-up

The optional stretch from #1762 — a per-user Redis token index to revoke *outstanding* access tokens on ban immediately (vs. waiting out the 15-min TTL) — is correctly scoped out. Worth filing the promised follow-up issue so it's tracked.

---

ginderjeremiah:
## Re-review (head `d3ade32`) — ✅ Approve, merging

Picked this up as the scheduled PR review. The prior review was against `eba5d69`; the only change since is `d3ade32`, which addressed both of that review's nits, so this is an incremental re-review of that delta plus a verification pass over the whole change.

**Delta since last review (`d3ade32`):**
- **Dropped unused `AccountState.Id`** — confirmed neither `Refresh` nor `SelectPlayer` read it (both already hold the `userId` they looked the state up by). Correct to remove as dead data on a fresh contract.
- **Archived-branch coverage added** — `Refresh_ArchivedSinceIssue_ReturnsNull` exercises the `account is null` path directly. For `SelectPlayer`, the new test honestly documents that `GetPlayerIds` (which excludes archived accounts) catches an archived account as `NotOwned` *before* the `GetAccountState` re-check is reached, so that branch there is only reachable via a narrow archive-mid-call race — documented rather than faked with a misleading assertion. Good call not to contrive it.

**Whole-change assessment:**
- Correct fix at the right layer: `GetAccountState` returns a `Contracts.Identity` read type with the projection in `Game.DataAccess`, matching the "repository reads speak contracts" rule. No domain logic leaked into the data tier or controller.
- Ordering is deliberate and consistent: in `SelectPlayer` the ban check runs *before* `Consume` so a rejected selection leaves the token intact (matching existing ownership-failure behavior); in `Refresh`, consume-then-check ends the chain on a banned account. Both correctly deny.
- Dead weight removed cleanly — `RefreshTokenData.Roles` and the `roles` param on `IRefreshTokenStore.Issue` are gone now that every reissue re-derives roles.
- Docs updated thoroughly across `backend.md`, `backend-admin.md`, and `backend-auth.md`, and honest about the one residual gap (an already-issued access token inside its 15-min TTL).
- Tests cover both flows × both scenarios (banned/archived → chain ends or rejected with token intact; role-granted → reissued pair carries the fresh role), with `RecordingAccessTokenService` asserting the *reissued* roles rather than the login snapshot.

**CI:** `all-checks-passed` ✅ (backend, frontend, e2e across chromium/firefox/webkit all green).

**Follow-up:** filed #2227 for the optional stretch from #1762 (per-user Redis token index to revoke *outstanding* access tokens on ban immediately, vs. waiting out the 15-min TTL) — correctly scoped out of this PR.

No blocking or should-fix findings. Merging.

---

Issue #2228, opened 2026-07-20
ActiveSession warning doesn't cover a different character on the same account


ginderjeremiah:
## Context

Follow-up to #1817 (PR TBD). That fix enforces one live character per account at the socket layer: `SocketManagerService.RegisterSocket` now claims an account-level presence key (`AccountSocket_{userId}`) and kicks a *different* character's socket on the same account exactly like a same-character takeover, closing the gap where two live characters would otherwise clobber the shared `Session_{userId}` cache.

## Residual gap

The pre-connect **warning** the client shows before triggering a takeover (`Auth/ActiveSession`, `SocketManagerService.HasActiveSocket`) is still keyed only by the character being selected, not the account. So:

- Selecting character A elsewhere, still connected, then selecting **character B** on the same account and checking `ActiveSession(B)` reports "no active socket" — no warning — even though opening B's socket will silently kick A's connection.
- This mirrors the existing same-character UX exactly (the warning is a courtesy check the client makes *before* connecting; the actual protection is enforced at handshake regardless of whether a warning fired), so it's not a correctness bug, just an inconsistency: a same-character takeover warns first, a cross-character-same-account takeover doesn't.

## Suggested direction

Extend the `ActiveSession` check (or add a variant) to also report when a *different* character on the caller's account currently holds the account-level slot, so the character-switcher / login flow can warn "you're already playing as X elsewhere" before selecting a different character. Needs a small design call on the response shape (just a bool, or also naming which character is live) — hence no `claude` tag.

Tags: enhancement, scope: small

PR #2229, opened 2026-07-20
Enforce one live character per account at socket registration

### LINKED ISSUE #1817: Session cache is one per-user slot but sockets/presence are per-player — two live characters on one account clobber each other's in-flight battle state
## Problem

The session cache is a single `Session_{userId}` slot (`Game.DataAccess/Repositories/SessionStore.cs:9-10`, "single active session per user"), but nothing actually enforces one live socket per **account**:

- Presence/takeover is keyed by `playerId` (`SocketManagerService.cs:351-354`), so sockets for two different characters on the same account never take each other over.
- `SwitchPlayer` deliberately proceeds when the departed character still has a live socket — it only skips the offline credit (`LoginController.cs:205-211`) — and never closes that socket.

Both characters' sockets then write their `PlayerState` (which carries the in-flight battle snapshot, seed, and anti-cheat `BattleSnapshot`) into the single per-user key via `Session.SavePlayerState()` on every battle command — last-writer-wins.

## Failure scenario

Tab A plays character 1; tab B calls `SwitchPlayer` to character 2 and opens character 2's socket. Both sockets are live, alternately overwriting `Session_{userId}`. On either character's next handshake, `SessionInitializer.EnsureSessionLoaded` sees the claim/cache player mismatch and rehydrates a fresh, empty `PlayerState` — the in-flight battle (and the #1595/#1597 resume/credit path) is silently discarded and never re-credited.

Related in the small: the `HasActiveSocket` credit guard is a point-in-time check against a 30s presence TTL while the inactivity watchdog allows 60s (`SocketManagerService.cs:27`, `SocketHandler.cs:46`), so a live-but-quiet (background-throttled) socket can pass the guard and race the HTTP credit — the lost-update race the guard exists to prevent.

## Suggested direction

Either key the cached session state by player (`(userId, playerId)` or just `playerId`), or enforce single-socket-per-account at the handshake (close the departed character's socket on switch). This is a design decision about the session-keying model, so worth a deliberate choice rather than a mechanical fix; tighten the credit-guard window while at it.

Distinct from #1513 (takeover command serialization) and #1518 (takeover-check ordering) — those concern takeover of the *same* player; this is the account-level session-slot design.

ginderjeremiah:
## Summary

`Session_{userId}` (the cached in-flight `PlayerState`) is deliberately keyed by **account**, not character — spike #922 decided "one live character at a time per account" and kept that key account-scoped on purpose. But socket presence/takeover (`SocketManagerService`) was only ever enforced per **character** (`PlayerSocket_{playerId}`), so nothing actually stopped two different characters on the same account from both having a live socket at once — e.g. a stale second tab, or a client that opens a second character's socket without tearing the first one down first (the switch flow's own credit logic already guards its own path via #2041, but a client that skips that flow entirely isn't caught).

When that happens, both sockets keep saving their own in-flight battle state into the *same* `Session_{userId}` cache slot — last-writer-wins, with neither connection ever told. On reconnect, `SessionInitializer` sees the cache/claim mismatch and silently rehydrates an empty `PlayerState`, discarding whichever battle got clobbered.

### Fix

`RegisterSocket` now also atomically claims an account-level presence key (`AccountSocket_{userId}`, same CompareAndSet-loop shape as the existing per-character claim) naming the account's current live character. If a *different* character on the account still holds that slot, its socket is kicked exactly like the existing same-character takeover (`SocketReplaced` push, best-effort — matching the existing model's tolerance).

I considered the issue's other suggested direction — rekeying the session cache by player instead of account — but that would undo spike #922's deliberate decision to keep one session per account, so I went with enforcing the invariant at the socket layer instead.

### While at it (tightening the credit-guard window, per the issue)

The presence-key TTL (30s) was shorter than the inactivity watchdog's own bound (60s + 10s poll = up to 70s a silent socket may still be genuinely alive, e.g. a backgrounded/throttled tab — see spike #922's background-tab-throttling note). That gap meant a still-alive-but-quiet socket could lose its presence claim to `TryClaimForSwitchCredit` (or the new account-level claim) while the watchdog still considered it live, reintroducing the lost-update race those claims exist to prevent. The TTL is now pinned to the watchdog's own worst-case bound instead. Trade-off: `HasActiveSocket`/`ActiveSession` now takes a bit longer to report an uncleanly-vanished connection (crash, no close frame) as gone.

## Note on scope

I did **not** extend the pre-connect `ActiveSession` warning to also cover "a different character on this account is live elsewhere" — the actual protection (the kick) doesn't depend on that warning firing, matching how same-character takeover already works today (warn-if-possible, enforce regardless). Filed as a small follow-up: #2228.

## Docs

Updated `docs/backend-sockets.md` (single-active-connection section + presence-TTL rationale) and `docs/backend.md`, and left a pointer comment on `SessionStore` explaining why the account-keyed invariant is safe.

## Test plan

- [x] `dotnet test` — Game.Api.Tests (840), Game.Application.Tests (1036), Game.Core.Tests (1416), Game.Infrastructure.Tests (68) — all passing
- [x] New integration tests in `SocketManagerServiceTests`:
  - `RegisterSocket_DifferentCharacterOnSameAccount_ClosesTheOtherCharactersSocketAndClaimsTheAccountSlot`
  - `RegisterSocket_SameCharacterReconnect_DoesNotAlsoTriggerAnAccountLevelReplace`
  - `UnRegisterSocket_GracefulClose_ClearsAccountPresenceKey`
  - `UnRegisterSocket_ReplacedByDifferentCharacter_DoesNotClearTheNewAccountPresenceKey`
  - `RefreshSocketPresence_AfterAccountKeyLapsesWhileStillLive_ResurrectsItsOwnClaim`
- [x] Updated existing unit tests in `SocketCommandProcessorTests` (exact-cache-call-count assertions) and `SocketConnectionTests` (hardcoded 30s TTL bound) for the new account-level calls / TTL

Closes #1817

---
_Generated by [Claude Code](https://claude.ai/code/session_01G3ZdsWeBUkpKY6yfVQJWh9)_


Game.Api/Services/SocketManagerService.cs
```diff
@@ -82,6 +96,14 @@ public async Task<SocketContext> RegisterSocket(WebSocket socket, SessionService
             // clear" and "the key was written" — see its doc comment.
             var oldSocketId = await ClaimPresenceKey(presenceKey, socketContext.SocketId);
 
+            // One live character per account (spike #922, decided A++) is enforced here too, not just one
+            // live socket per character: Session_{userId} (SessionStore) is keyed by account, so if a
+            // *different* character on this account still had a live socket, its connection would keep
+            // saving its own in-flight battle into the same cache slot this connection is about to use,
+            // last-writer-wins, with neither side ever told (#1817). Claiming the account-level slot mirrors
+            // ClaimPresenceKey exactly, just keyed by userId instead of playerId.
+            var oldAccountPlayerId = await ClaimAccountPresence(userId, playerId);
```
ginderjeremiah: Nit (optional): `ClaimAccountPresence` sits between the per-player claim and the `try`, so it's the one operation that can throw *after* the per-player key is already claimed but *before* the rollback path exists. If the account claim faults (a transient Redis blip on the `Get`/`CompareAndSet`), the per-player key stays claimed with no teardown until its TTL lapses.

It's self-healing in practice — the ghost key never blocks a reconnect (`ClaimPresenceKey`'s CompareAndSet loop takes it over) and only lets `HasActiveSocket` report a false positive for up to the TTL — so not a blocker. But since `TeardownSocketRegistration` already releases *both* keys idempotently, moving this line to be the first statement inside the `try` would fold the account claim under the same rollback guard and remove the asymmetry (before the PR nothing throwable sat between the claim and the `try`). Take it or leave it.

---

ginderjeremiah (commented):
## Review: Enforce one live character per account at socket registration (#1817)

**Verdict:** Approve — *(submitted as a comment because I'm the PR author and GitHub won't accept a self-approval; the intended verdict is Approve.)* Correct, well-layered, thoroughly tested fix with one optional nit.

The change enforces the one-live-character-per-account invariant at the socket layer: `RegisterSocket` now atomically claims an account-level `AccountSocket_{userId}` slot (same CompareAndSet-loop shape as the per-player claim) and kicks a *different* character's socket exactly like a same-character takeover. I agree with choosing this over the issue's other suggested direction (rekeying the session by player) — that would undo spike #922's deliberate account-scoped session decision, so enforcing the invariant one layer up is the right call, and it's documented as such.

The correctness details hold up under scrutiny:
- **No spurious account-level replace on same-character reconnect** — `currentValue != newValue` short-circuits to `null`, so a reconnect only fires the per-player takeover (verified by test).
- **Teardown can't clobber a newer owner** — the account-key release is a compare-and-delete keyed on the releasing socket's *own* player id, so character 1's teardown leaves character 2's freshly-claimed slot intact (verified by test).
- **Switch-credit sentinel respected** — the account-level kick reads the other character's per-player key and skips the `switch-credit` sentinel, so it never notifies a non-socket.
- **`int.Parse` is safe** — the account key only ever holds a player-id string or null (the sentinel only ever targets per-player keys), and InvariantCulture is used consistently on both write and parse.

### Nits

- `Game.Api/Services/SocketManagerService.cs:105` — `ClaimAccountPresence` sits between the per-player claim and the `try`; if it faults, the per-player key is briefly orphaned until its TTL (self-healing, not a blocker). Folding it into the `try` would remove the asymmetry. Left as an inline comment. Optional.

### TTL change

Pinning `SocketPresenceTtl` to `DefaultInactivityTimeout + DefaultInactivityPollInterval` (the watchdog's worst-case bound) is the right fix for the credit-guard window the issue flagged in the small: a shorter TTL let a background-throttled-but-still-live socket lose its claim to a competing credit/takeover while the watchdog still considered it alive. Deriving the TTL from the constants (rather than a second hardcoded literal) keeps the two from drifting, and the `internal` widening + `SocketConnectionTests` referencing the constant instead of a hardcoded `30s` is the correct way to keep the test honest.

### Test coverage

Strong. Integration tests cover the different-character takeover, the same-character-reconnect no-double-replace guard, graceful-close clearing the account key, replaced-by-different-character *not* clearing the new owner's key, and heartbeat resurrection of a lapsed account claim. Unit tests were updated for the extra account-level claim/release in both the rollback and switch-credit-race paths with exact call-count assertions. This isn't battle logic, so the frontend-mirror rule doesn't apply.

### Docs & follow-up

`backend-sockets.md`, `backend.md`, and the `SessionStore` pointer comment are all updated to explain why the session cache stays account-keyed and where the invariant is now enforced. The residual UX gap (the pre-connect `ActiveSession` warning still doesn't cover a different character on the same account) is correctly scoped out as follow-up #2228 — it's a warning courtesy, not the protection, matching how same-character takeover already behaves, so it doesn't belong in this PR.

### What's good

Clean mirroring of the existing claim/release/refresh/rollback machinery for the new account key, genuinely helpful rationale comments, and a careful matching-tests pass. Nice work.

---

ginderjeremiah:
## Independent review — Enforce one live character per account at socket registration (#1817)

**Verdict: Approve.** I re-reviewed the full diff at `c87a8ae` independently of the existing self-review, with focus on the correctness of the new account-level presence claim and the post-review rollback-guard fix. It holds up.

**Root cause & fix are right.** `Session_{userId}` is deliberately account-keyed (spike #922), but presence/takeover was enforced only per character, so two characters on one account could both hold live sockets and silently clobber each other's in-flight battle state via the shared session slot. Claiming an account-level `AccountSocket_{userId}` key at registration — same `CompareAndSet`-loop shape as `ClaimPresenceKey`, just keyed by account — and kicking a *different* character exactly like a same-character takeover is the correct layer to enforce the invariant. Rekeying the session by player (the issue's alternative) would have undone #922's decision, so I agree with the choice, and it's documented on `SessionStore` and in `backend-sockets.md`/`backend.md`.

Correctness details I verified:
- **No spurious account-level replace on same-character reconnect** — `ClaimAccountPresence` returns `null` when `currentValue == newValue`, so a reconnect fires only the per-player takeover. Covered by test.
- **Two distinct kicks compose correctly** — the per-player notify (`oldSocketId`) and the account-level notify (`oldAccountPlayerId` → other character's own per-player key) target different sockets and both are appropriate; the account kick correctly reads the *other* character's per-player key and skips the switch-credit sentinel.
- **Teardown can't clobber a newer owner** — the account key release is a compare-and-delete keyed on the releasing socket's own player id (verified by `UnRegisterSocket_ReplacedByDifferentCharacter_...`), and `RefreshSocketPresence` uses `ReclaimAndForget`, so a stale socket's heartbeat extends but never overwrites a newer owner's claim.
- **`int.Parse` is safe** — the account key only ever holds a player-id string or null (the sentinel only targets per-player keys), InvariantCulture used consistently on write and parse.
- **Rollback-guard fix (`c87a8ae`) is correct** — moving `ClaimAccountPresence` inside the `try` folds it under `TeardownSocketRegistration`, which idempotently releases both keys, removing the brief per-player-key-orphaned-on-fault asymmetry the review flagged.

**TTL change is sound.** Pinning `SocketPresenceTtl` to `DefaultInactivityTimeout + DefaultInactivityPollInterval` (deriving from the constants rather than a second literal, and widening them to `internal`) correctly closes the credit-guard window: a shorter TTL let a still-live-but-throttled socket lose its claim to a competing takeover/credit while the watchdog still considered it alive. The trade-off — `HasActiveSocket`/`ActiveSession` takes up to the watchdog bound to report an uncleanly-vanished socket as gone — is the safe direction and is documented.

**Tests & docs are thorough.** Integration coverage spans different-character takeover, the same-character no-double-replace guard, graceful-close clearing, replaced-by-different-character not clearing the new owner's key, and heartbeat resurrection; unit tests updated with exact call-count assertions on both keys. Not battle logic, so the parity-matrix rule doesn't apply. Scoping the `ActiveSession`-warning extension out to #2228 is the right call — the kick is the protection, the warning is courtesy.

No nits or should-fixes outstanding. CI is green. Merging.

---

PR #2230, opened 2026-07-21
Revoke a banned/archived account's live socket immediately

### LINKED ISSUE #2226: Optionally revoke a user's outstanding access/refresh tokens immediately on ban (per-user Redis token index)
## Context

Follow-up to #1762 / PR #2224. That PR closed the main hole: `AccountService.Refresh` and `SelectPlayer` now re-check the account's live ban/role state (`IUsers.GetAccountState`) instead of trusting the refresh token's stored snapshot, so a ban or role change lands on the account's next refresh/selection rather than only at next login.

## Residual gap

Enforcement is a *re-check on use*, not a *revocation*. Two windows remain after a ban:

- An **already-issued access token** stays valid, validated offline, until its 15-minute TTL expires — not re-checked per request.
- A banned account's **pre-selection refresh token** is intentionally left intact on a rejected `SelectPlayer` (token-intact-on-rejection semantics), so it isn't proactively invalidated either; the chain only ends the next time it's used on `Refresh`.

## Proposed work (the optional stretch from #1762)

Add a per-user token index in Redis so a ban (and role revocation) can revoke a user's outstanding tokens immediately rather than waiting out the TTL:

- Maintain a per-user set of live refresh-token keys (or a user-scoped generation/epoch counter) alongside the existing SHA-256-keyed entries in `RefreshTokenStore`.
- On ban/role-revoke, invalidate that user's outstanding refresh tokens (and, if we want immediate access-token cutoff, gate access-token validation on the user's current epoch rather than pure offline validation — a design decision with a per-request-read cost to weigh).
- Wire the invalidation into the admin ban / `SetUserRoles` paths.

This is a genuinely separate design decision (new Redis structure + invalidation wiring + the offline-vs-checked access-token trade-off), which is why it was scoped out of PR #2224.

## Labels
Suggested: `enhancement`, `scope: medium`, `claude`.

### LINKED ISSUE #2227: Revoke outstanding access tokens immediately on ban/role-strip via a per-user Redis token index
## Context

Follow-up to #1762 / PR #2224. That PR closed the main hole: `Refresh` and `SelectPlayer` now re-check the account's live ban/role state (`IUsers.GetAccountState`) instead of trusting the refresh token's stored snapshot, so a ban or role change takes effect on the account's next refresh rather than only at its next login.

The one **residual gap** it deliberately left open (documented in `docs/backend-admin.md` and `docs/backend.md`): an **already-issued access token still inside its 15-minute TTL** is validated offline and never re-checked per request. So a banned/demoted user keeps whatever their current access token grants for up to 15 minutes — including admin endpoints, since `AdminRoleAuthorizationFilter` trusts the token's role claim alone.

## Proposed work

Add a per-user token index in Redis so an admin action can revoke a user's **outstanding** tokens immediately, rather than waiting out the access-token TTL:

- Maintain a per-user index of issued token identifiers (refresh tokens are keyed today solely by SHA-256 of the token value — `RefreshTokenStore` — with no per-user index, so revoking a user's tokens isn't even possible currently).
- On ban / archive / Admin-role-strip (`Users.BanUser` and the `SetUserRoles` path), invalidate that user's indexed tokens.
- Decide how access tokens participate: either a lightweight per-request revocation check (e.g. a denylist / token-version keyed by user) or accept the 15-min access-token window and only hard-revoke the refresh chain. This is the core design decision to weigh — a per-request Redis read reintroduces the lookup the "token is the sole authorization source of truth" model was designed to avoid, so the trade-off (immediacy vs. per-request cost) should be evaluated explicitly.
- Update `docs/backend-admin.md` / `docs/backend.md` to reflect whichever enforcement point lands.

## Notes

Scoped out of PR #2224 as a genuinely separate design decision (new Redis structure + invalidation-on-ban wiring). Current behavior bounds exposure to the 15-minute access-token TTL, which is an acceptable interim state — this issue is a hardening enhancement, not a regression.

ginderjeremiah:
## Summary

#2226 and #2227 are near-duplicate follow-ups to #1762/PR #2224 (filed ~an hour apart, both proposing the same fix — see the comment I left on #2227). Both propose a per-user Redis token index so a ban/role-strip can revoke a user's outstanding tokens immediately rather than waiting out the access-token TTL. I implemented the underlying goal, but not that literal mechanism — here's why, and what I built instead.

### Why not a Redis token index for the refresh token

It would be a no-op. #2224 already made `Refresh`/`SelectPlayer` re-check the account's live ban/archive state on every use. A refresh token is also single-use (`RefreshTokenStore.Consume` is an atomic get-and-delete). So a banned account's outstanding refresh token is already fully inert the moment it's next presented — deleting it from the store early changes nothing observable; the client gets the same rejection either way.

### Why not a per-request access-token revocation check

This is the part of #2226/#2227 that's a genuine design decision, not a bug — both issues frame it as one to "weigh explicitly." `docs/backend.md` documents **token-as-sole-authority, no per-request lookup** as a deliberate trade-off, and `AdminAuthorizationTests.AdminToken_WithNoSessionCacheKey_Authorizes` pins it: a token for a user id with no corresponding DB row still authorizes purely on its claims. Adding a live DB/Redis check (even scoped to just the admin surface) would mean rewriting that test to reverse a tested architectural invariant — not something to do unilaterally. I left this alone; it's the same call the project already made and documented.

### What actually was a real, un-covered gap

An **already-open WebSocket is never re-validated against account state at all** — no per-command check, no expiry tied to the access token. So today, a banned/archived player's live game socket keeps working indefinitely (not bounded by the 15-minute access-token TTL) until it happens to go idle or disconnect. That's a bigger hole than the one the issues called out, and it's fixable without touching the token/HTTP architecture.

## Changes

- New `AccessRevoked` server-initiated socket command (mirrors `SocketReplaced`'s send-then-close shape) and `ESocketCloseReason.AccessRevoked`.
- `SocketManagerService.RevokeAccess(playerIds)`: best-effort, closes any currently-live socket among the given player ids. Silent when none are connected (the common case for a ban target — unlike a gameplay push, this must not warn on "no active socket").
- `AdminUsersController.BanUser`/`ArchiveUser` fetch the target's player ids *before* the action (archiving excludes the account from `GetPlayerIds` afterward) and call `RevokeAccess` on success.
- Frontend: `handleAccessRevoked` (mirrors `handleSocketReplaced` — tear down, acknowledge a modal, route to login) wired app-wide in the root layout, same as the session-takeover push.
- Regenerated `UI/src/lib/api/types/api-socket-type-map.ts` via the codegen tool for the new command.
- Docs: `backend-admin.md` and `backend-auth.md` document the immediate socket revocation, the remaining accepted gap (non-socket HTTP surface within the 15-minute access-token TTL), and why the Redis-token-index / per-request-admin-check approaches were rejected. `frontend.md` notes `AccessRevoked` alongside `SocketReplaced`.

## Test plan

- [x] `dotnet test` (full solution) — 3,367 passed, 0 failed
- [x] New integration tests:
  - `SocketManagerServiceTests`: `RevokeAccess_PlayerHasActiveSocket_PushesAccessRevokedAndCloses`, `RevokeAccess_PlayerHasNoActiveSocket_IsSilentNoOp`, `RevokeAccess_MixOfLiveAndOfflinePlayerIds_OnlyClosesTheLiveOne`
  - `AdminUsersControllerTests`: `BanUser_TargetHasLiveSocket_RevokesAccessImmediately`, `ArchiveUser_TargetHasLiveSocket_RevokesAccessImmediately`, `BanUser_TargetNotConnected_SucceedsWithoutError`
- [x] `npm run lint` — clean
- [x] `npm run test:unit:ci` — 3,860 passed, 0 failed (new `handleAccessRevoked` describe block in `engine.test.ts`, root-layout wiring assertion in `layout.svelte.test.ts`)

Closes #2226
Closes #2227

---
_Generated by [Claude Code](https://claude.ai/code/session_01VhRdqrg16mg5TBVzrx5u7s)_


docs/backend-admin.md
```diff
@@ -43,7 +43,7 @@ The admin user-management surface (search the roster, list grantable roles, repl
...
...herwise keep playing indefinitely (not merely for the access-token TTL) until it happened to go idle. It is a best-effort push-then-close (`AccessRevoked`, mirroring the session-takeover `SocketReplaced` push) keyed off the account's player ids fetched *before* the action commits, since a just-archived account no longer resolves them. The one residual gap is the **non-socket HTTP surface with an already-issued access token still inside its 15-minute TTL**, which is validated offline and not re-checked per request — deliberately left alone: re-checking it would mean either a per-request live-state lookup for every authenticated call, or narrowing that to just the admin surface, both of which contradict the tested, documented **token-as-sole-authority, no-second-lookup** model (see [backend.md → Access Roles and Admin Authorization](./backend.md#access-roles-and-admin-authorization) and `AdminAuthorizationTests`). A per-user Redis token index to hard-revoke the refresh chain was considered and rejected as a no-op: `Refresh`/`SelectPlayer` already reject a banned/archived account's refresh token on next use (single-use + live re-check), so deleting it early changes nothing observable.
```
ginderjeremiah: Should-fix (doc precision): calling the residual "the **non-socket HTTP surface**" understates it. The socket *handshake* also validates the access token offline, not live account state (`backend.md`: "the socket handshake … read[s] the bound player from this claim"), and `RegisterSocket` does no live ban re-check. So a banned user whose socket you just closed can immediately **re-open a fresh socket** with the same still-valid access token and keep playing until it expires. The kick's real effect is therefore to bound the window from *indefinite* down to the ≤15-minute access-token TTL (against a cooperative client it's immediate; against the adversary you're actually banning it's ≤15 min via reconnect) — not to eliminate it for the socket surface. That's still within the accepted 15-min envelope from #2227, so it's not a blocker, but the wording should include socket reconnection in the residual rather than implying only "non-socket HTTP" remains.

Optional stronger fix (or a follow-up to #2226/#2227): a single live ban check at the socket **handshake** is one lookup *per connection*, not per request — so it wouldn't reintroduce the per-request cost the token-as-sole-authority model avoids, and it would make the revocation durable against reconnect. Worth considering as the natural completion of this change.

---

ginderjeremiah (commented):
## Review: Revoke a banned/archived account's live socket immediately (#2226/#2227)

**Verdict:** Approve *(submitted as a comment because I'm the PR author and GitHub won't accept a self-approval; the intended verdict is Approve)* — a correct, well-layered, thoroughly tested fix, with one should-fix on doc precision (not a blocker).

The change closes a real hole the two issues only pointed near: an already-open WebSocket is never re-validated against account state, so a banned/archived player's live game socket keeps working **indefinitely** (not merely for the access-token TTL) until it happens to go idle. `RevokeAccess` closes any live socket among the account's players via a `SocketReplaced`-shaped push-then-close (`AccessRevoked`), wired into `BanUser`/`ArchiveUser`.

I agree with the call to *not* build the literal Redis token index the issues requested: the refresh token is already inert on next use (single-use + live re-check from #2224), so deleting it early is a no-op, and a per-request access-token check contradicts the tested, documented token-as-sole-authority model. Declining that and closing the bigger socket gap is the right scope.

The correctness details hold up:
- **`AccessRevoked` mirrors `SocketReplaced` exactly** — `IServerInitiatedCommand` (rejected on the inbound client path, so a client can't self-trigger it) + `ISelfDeliveringCommand` (owns its send-then-close so the success frame lands before the close frame). Auto-registered by `SocketCommandFactory`'s reflection scan; no manual wiring needed. Not `IReplayableServerCommand`, correctly — a revocation must not be dead-letter replayed.
- **`GetPlayerIds` is fetched *before* the action** — it filters `ArchivedAt == null`, so reading it after `ArchiveUser` would return empty. The ordering comment is accurate and the behavior is verified by test.
- **Silent no-op when offline** — the per-player `HasActiveSocket` pre-check avoids `EmitSocketCommand`'s no-active-socket warning, which for a ban/archive target (usually offline) would be log noise, not an anomaly.
- **Layering is consistent** — `AuthController` already depends on `SocketManagerService` for the takeover warning, so `AdminUsersController` doing the same for the kick follows the established pattern; the socket concern stays in `Game.Api` rather than leaking into the application layer.

### Should-fix

- `docs/backend-admin.md:46` (and the mirroring sentence in `backend-auth.md`) — the residual is described as the "non-socket HTTP surface", but the socket **handshake** also validates the token offline and `RegisterSocket` does no live ban check, so a banned user can re-open a fresh socket within the 15-min TTL. The kick bounds the window to ≤15 min rather than eliminating it for the socket surface; the wording should say so. Within #2227's accepted envelope, so not blocking. (Inline comment also suggests a one-lookup-per-connection handshake check as the natural completion — a fair candidate to fold in here or track as follow-up.)

### Nits

- `SocketManagerService.RevokeAccess` — `HasActiveSocket` reports `true` during the brief switch-credit-sentinel window, but `EmitSocketCommand(info, playerId)` treats that sentinel as "no active socket" and logs the very warning the pre-check exists to avoid. Negligibly rare (a switch-credit claim racing a ban on the same account) and self-clearing; noting only for completeness.

### Test coverage

Strong on both sides. Backend integration tests cover the live-socket revoke, the silent-no-op (asserting no warning is logged), and the mixed live/offline list; controller tests cover ban/archive with a live socket and ban of an offline target. Frontend adds a `handleAccessRevoked` describe block (teardown, log reset, modal copy, deferred navigation) and the root-layout listener assertion. Not battle logic, so the frontend/backend parity-matrix rule doesn't apply.

### What's good

Clean reuse of the existing `SocketReplaced` command/handler/root-layout machinery for the revocation path, genuinely helpful rationale comments (especially the fetch-ids-before-archive ordering and the why-not-Redis-token-index reasoning), and accurate docs across `backend-admin.md`, `backend-auth.md`, and `frontend.md`.

---

ginderjeremiah:
Thanks for the review. Addressed:

- **Should-fix (doc precision)**: fixed in 6c3d3ec. `backend-admin.md`/`backend-auth.md` no longer imply the residual gap is "non-socket HTTP only" — they now say the handshake validates the access token offline with no live re-check, so a reconnect can still happen within the existing 15-minute TTL. The kick's real effect is forcing a cooperative client's session to end now and bounding an adversarial one to that TTL via reconnect, not eliminating the socket-surface gap outright.
- **Optional stronger fix (handshake live check)**: filed as #2231 rather than folding into this PR — it's a real design decision (a live DB read on the connection hot path, plus deciding whether it should also cover the switch-credit claim path) rather than a small fix, so it deserves its own scoped PR.
- **Nit (switch-credit sentinel race)**: no action — agreed it's negligible and self-clearing.

Ready for another look.

---

ginderjeremiah:
## Independent review — Revoke a banned/archived account's live socket immediately (#2226/#2227)

**Verdict: Approve.** I re-reviewed the full diff at `6c3d3ec` independently, focusing on the new `AccessRevoked` command, `RevokeAccess`, and the controller wiring. It's correct and well-scoped.

**Scope call is right.** The two issues asked for a per-user Redis token index; declining it (the refresh token is already inert on next use — single-use + live re-check from #2224) and a per-request access-token check (contradicts the tested, documented token-as-sole-authority model) is the correct judgment. The real, uncovered hole — an already-open WebSocket is *never* re-validated against account state, so a banned/archived player's live socket keeps working indefinitely — is the right thing to close, and it's fixable without touching the token/HTTP architecture.

Correctness details I verified:
- **`AccessRevoked` mirrors `SocketReplaced` exactly** — `IServerInitiatedCommand` (rejected on the inbound client path, so a client can't self-trigger a revoke) + `ISelfDeliveringCommand` (owns its send-then-close so the success frame lands before the close frame). Correctly *not* `IReplayableServerCommand` — a revocation must not be dead-letter replayed. `NormalClosure` mapping routes the client to login rather than reconnect.
- **`GetPlayerIds` fetched before the action** — it excludes an archived account, so reading it afterward would revoke nothing; the ordering is commented and verified by `ArchiveUser_TargetHasLiveSocket_...`.
- **Guarded on `UserActionStatus.Success`** — a rejected ban/archive (self-action, last-admin guard) doesn't revoke. Correct.
- **Silent no-op when offline** — the per-player `HasActiveSocket` pre-check avoids `EmitSocketCommand`'s no-active-socket warning, which for a usually-offline ban target would be log noise, not an anomaly. Verified by `RevokeAccess_PlayerHasNoActiveSocket_IsSilentNoOp`.
- **Layering** — `AuthController` already depends on `SocketManagerService`, so `AdminUsersController` doing the same keeps the socket concern in `Game.Api`.
- **Frontend** — `handleAccessRevoked` reuses the `handleSocketReplaced` tear-down/acknowledge/route-to-login shape with distinct copy, wired app-wide in the root layout so it fires regardless of the mounted route (same fix as #1836). Type map regenerated.

**Post-review doc fix (`6c3d3ec`) is correct** — the residual is no longer mis-described as "non-socket HTTP only"; it now states the handshake validates the token offline with no live re-check, so a just-kicked user can reopen a fresh socket within the ≤15-min TTL. The kick bounds a determined reconnect to that TTL rather than eliminating the socket-surface window. The stronger handshake-time live check is correctly deferred to #2231.

### Follow-up filed
- **#2233** (tech-debt, small): `RevokeAccess` gates on `HasActiveSocket`, which returns `true` during the switch-credit sentinel window, so `EmitSocketCommand` logs a spurious no-active-socket warning in that rare race. Log-noise only, self-clearing — filed rather than folded in to keep this PR focused. The `RegisterSocket` account-kick already skips the sentinel and is the pattern to mirror.

**Tests are thorough** on both sides (live revoke, silent no-op asserting no warning, mixed live/offline list, ban/archive-with-live-socket, ban-of-offline-target; frontend `handleAccessRevoked` describe block + root-layout listener assertion). Not battle logic, so the parity-matrix rule doesn't apply.

The one nit is tracked as a follow-up and the bigger handshake gap as #2231; no should-fixes remain. Proceeding to merge once the branch is current with `main` (#2229 landed in the same file).

---

PR #2234, opened 2026-07-21
Fix RevokeAccess logging a spurious no-active-socket warning

### LINKED ISSUE #2233: RevokeAccess logs a spurious no-active-socket warning during the switch-credit sentinel window
## Context

Follow-up from PR #2230 (revoke a banned/archived account's live socket immediately).

`SocketManagerService.RevokeAccess` gates its push with `HasActiveSocket(playerId)` specifically to avoid `EmitSocketCommand`'s no-active-socket warning for the common ban/archive case where the target is offline:

```csharp
public async Task RevokeAccess(IReadOnlyList<int> playerIds)
{
    foreach (var playerId in playerIds)
    {
        if (await HasActiveSocket(playerId))
        {
            await EmitSocketCommand(new AccessRevokedInfo(), playerId);
        }
    }
}
```

## The gap

`HasActiveSocket` is a presence-key existence check, so it also returns `true` while the per-player presence key holds the **switch-credit sentinel** (`TryClaimForSwitchCredit`) rather than a real socket id. In that window `EmitSocketCommand` finds no actual socket in the registry and logs the very no-active-socket warning the pre-check exists to suppress.

Impact is limited to a single spurious log warning, it's self-clearing, and the race is extraordinarily rare (an admin ban/archive landing on an account exactly while that account's in-game character-switch credit claim is in flight) — hence a follow-up rather than a blocker on #2230.

## Suggested fix

Mirror the account-level kick in `RegisterSocket`, which already reads the socket id and skips the sentinel (`otherSocketId != SwitchCreditClaimValue`): have `RevokeAccess` read the current socket id and skip when it's null or the sentinel, instead of relying on the `HasActiveSocket` bool. Add a `SocketManagerServiceTests` case seeding the sentinel value and asserting no warning is logged.

## Labels

ginderjeremiah:
## Summary

Follow-up from #2230. `SocketManagerService.RevokeAccess` gated its push with `HasActiveSocket(playerId)` specifically to avoid `EmitSocketCommand`'s no-active-socket warning for the common (offline) ban/archive case. But `HasActiveSocket` is a presence-key *existence* check, so it also returns `true` while the key holds the **switch-credit sentinel** (`TryClaimForSwitchCredit`) rather than a real socket id. In that window, `EmitSocketCommand(_, playerId)` re-reads the key, finds no real socket in the registry, and logs the very warning the pre-check exists to suppress.

## Fix

`RevokeAccess` now reads `CurrentSocketId` directly and skips a `null` or sentinel value before pushing — mirroring the existing `otherSocketId != SwitchCreditClaimValue` check in `RegisterSocket` — instead of gating on `HasActiveSocket` and then re-resolving the socket id through the int-keyed `EmitSocketCommand` overload.

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test` (`Game.Api.Tests`, `SocketManagerServiceTests`) — 847 passed, 0 failed
- [x] Added `RevokeAccess_PlayerHeldBySwitchCreditSentinel_IsSilentNoOp`, seeding the sentinel via `TryClaimForSwitchCredit` and asserting no warning is logged by `RevokeAccess`

Closes #2233

---
_Generated by [Claude Code](https://claude.ai/code/session_01DTNA9fiEBpRYC2ji9P8b32)_

### DIFF
--- a/Game.Api.Tests/Integration/SocketManagerServiceTests.cs
+++ b/Game.Api.Tests/Integration/SocketManagerServiceTests.cs
@@ -535,6 +535,29 @@ namespace Game.Api.Tests.Integration
             Assert.Equal(nameof(AccessRevoked), revoked.Name);
         }
 
+        [Fact]
+        public async Task RevokeAccess_PlayerHeldBySwitchCreditSentinel_IsSilentNoOp()
+        {
+            // A ban/archive landing while the target's presence key holds the switch-credit sentinel (rather
+            // than a real socket id or being unset) is the gap #2233 fixes: HasActiveSocket reports this
+            // player as present (see TryClaimForSwitchCredit's doc comment), but there is no real socket to
+            // push AccessRevoked to, so the old HasActiveSocket-gated check fell through to EmitSocketCommand's
+            // no-active-socket warning branch. RevokeAccess must recognize the sentinel and skip silently.
+            const int playerId = 918273;
+            var startIndex = _capturingProvider.Entries.Count;
+
+            using (var scope = CreateScope())
+            {
+                var socketManager = scope.ServiceProvider.GetRequiredService<SocketManagerService>();
+                Assert.True(await socketManager.TryClaimForSwitchCredit(playerId));
+
+                await socketManager.RevokeAccess([playerId]);
+            }
+
+            Assert.DoesNotContain(_capturingProvider.Entries.Skip(startIndex),
+                e => e.Category == SocketManagerCategory && e.Level == LogLevel.Warning);
+        }
+
         private async Task<bool> HasActiveSocketAsync(int playerId)
         {
             using var scope = CreateScope();
--- a/Game.Api/Services/SocketManagerService.cs
+++ b/Game.Api/Services/SocketManagerService.cs
@@ -350,19 +350,25 @@ namespace Game.Api.Services
         /// client's access token expires (up to the 15-minute TTL) or the socket next goes idle. Silent when
         /// none are connected: unlike a gameplay push (<see cref="EmitSocketCommand(SocketCommandInfo, int)"/>),
         /// the target of a ban/archive is very often already offline, so that is the common case here, not a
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: RevokeAccess spurious no-active-socket warning during the switch-credit sentinel window (#2233)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't accept a gating event from me) — the fix is correct, minimal, and matches the exact pattern the issue prescribed.

### Assessment

The change swaps the `HasActiveSocket`-gate → int-keyed `EmitSocketCommand` combination for a single `CurrentSocketId` read that skips `null` and the switch-credit sentinel before pushing to the resolved socket id. That directly closes the gap: the old path re-read the presence key inside `EmitSocketCommand(_, playerId)` and, while the key held the sentinel, hit the no-active-socket warning branch the pre-check existed to suppress. Reading the id once and pushing through the string overload avoids the double-read entirely.

The guard `socketId is not null && socketId != SwitchCreditClaimValue` is a faithful mirror of the `otherSocketId != SwitchCreditClaimValue` check in `RegisterSocket` (SocketManagerService.cs:145) and the same guard inside `EmitSocketCommand(_, int)` (SocketManagerService.cs:404), so the three sentinel-aware call sites now stay consistent — good.

### Test coverage

- `RevokeAccess_PlayerHeldBySwitchCreditSentinel_IsSilentNoOp` seeds the sentinel via `TryClaimForSwitchCredit` and asserts no `Warning` is logged, capturing entries from a start index so it's isolated from prior test noise. This is exactly the scenario #2233 describes.
- The live-socket happy path (still emits `AccessRevoked`) remains covered by the existing `RevokeAccess_MixOfLiveAndOfflinePlayerIds_OnlyClosesTheLiveOne`, so the fix isn't silently converting real pushes into no-ops.

### What's good

- Layering is correct (API/infra socket concern, no domain logic leak), braces are present, nullable handling is explicit with no null-forgiving operator, and the updated doc comment accurately explains *why* the direct read is used over the `HasActiveSocket` gate.

---

PR #2246, opened 2026-07-21
Throttle repeated database-outage error logs in the player-update drain

### LINKED ISSUE #2159: Parked infra-failure player-update events log an Error per reclaim cycle during a prolonged outage
## Context

Follow-up to #2097 (PR #2158), flagged in that PR's review as a nit not worth blocking merge.

`DataProviderSynchronizer.ProcessMessage`'s terminal catch for an `InfrastructureFailureClassifier`-classified failure logs an `Error` each time an item exhausts its retries (`Game.DataAccess/DataProviderSynchronizer.cs`). During a prolonged database outage with many active players, every parked item re-faults on each opportunistic-stranded-processing reclaim cycle (roughly the `DefaultReclaimGracePeriod`, ~30s, cadence) and logs another `Error` — potentially a flood of near-identical error-log lines every ~30s for the duration of the outage.

This is the documented, bounded tradeoff of #2097's fix (parking instead of dead-lettering), not a correctness bug — but an unthrottled error flood could drown the rest of the error stream during exactly the incident an operator most needs clean signal.

## Suggested fix

Summarize/throttle the outage log the same way `SurfaceDeadLetterDepth` already avoids re-spamming the dead-letter-depth warning on every drain (`_lastReportedDeadLetterDepth`) — e.g. log the first occurrence of an outage at `Error`, then downgrade repeats within some window to `Debug`/`Information`, or log one aggregate line per reclaim round (count of parked items) rather than one line per item.

ginderjeremiah:
## Summary

Closes #2159.

`DataProviderSynchronizer.ProcessMessage`'s terminal catch for a database-**infrastructure** failure (`InfrastructureFailureClassifier`) previously logged an `Error` every single time an item exhausted its retries. During a prolonged outage, every parked item re-faults through this same path on each opportunistic-stranded-processing reclaim cycle (~30s), so with many active players this could flood the log with one `Error` per parked item per cycle — potentially drowning out the rest of the error stream during exactly the incident an operator most needs clean signal from.

## Changes

- Added an `_infrastructureOutageLogged` flag (`int`, accessed via `Interlocked` since distinct players' events apply concurrently within one drain pass per #1701) to `DataProviderSynchronizer`.
- The infrastructure-failure catch clause now logs the **first** occurrence at `Error` as before, but downgrades subsequent occurrences to `Debug` while the flag stays set.
- The flag resets to `0` the moment any event applies successfully (`HandleEvent` returns without throwing), so the *next* outage is reported fresh at `Error` rather than staying silently suppressed by a stale flag from a previous, already-recovered outage.
- Updated `docs/backend-persistence.md`'s "Queue processing resilience" section with a one-sentence note on the new throttling behavior.

This mirrors the existing `SurfaceDeadLetterDepth`/`_lastReportedDeadLetterDepth` pattern the issue pointed at (log only when there's genuinely new information), adapted to a "first failure vs. already-reported outage" state rather than a monotonically growing count.

## Test plan

- [x] Added `ProcessQueue_MultipleConcurrentInfrastructureFailuresInOnePass_LogsOnlyOneErrorAndRestAtDebug` — two distinct players' events both hit an infrastructure failure within the same drain pass; asserts exactly one `Error` and one `Debug` log entry.
- [x] Added `ProcessQueue_InfrastructureOutageRecoversThenRecurs_LogsErrorAgainOnTheNextOutage` — drives the synchronizer through outage → recovery (a real successful apply) → a second, later outage; asserts the second outage logs a fresh `Error` rather than staying suppressed.
- [x] Full existing `DataProviderSynchronizerTests` suite still passes unchanged (the existing single-item infra-failure theory test still sees its one `Error` entry).
- [x] Full backend test suite (`dotnet test`, all 4 projects): **3387/3387 passing**, 0 failures.

---
_Generated by [Claude Code](https://claude.ai/code/session_012nYapeBRwF1KhQtCp3TnmC)_


ginderjeremiah (commented):
## Review: throttle repeated database-outage error logs in the player-update drain (#2159)

**Verdict:** Approve — the change cleanly solves the flood described in #2159 for the primary outage mode, mirrors the existing `_lastReportedDeadLetterDepth` "log only on new information" pattern, is well-documented, and is covered by two targeted tests. (Posted as `COMMENT` because GitHub won't accept a gating verdict from the PR author; treat this as an approval.)

### What's good
- Correct, minimal use of `Interlocked.Exchange` for the failure-path transition (`0->1` logs Error, otherwise Debug) and `Volatile.Read`-guarded reset on the success path to avoid an interlocked write on every successful apply. The concurrency reasoning (#1701) holds — the assertions in the concurrent test are deterministic regardless of interleaving.
- Resetting on the first successful apply (rather than a timer) means a genuinely stuck outage logs exactly one `Error` for its whole duration, and a *new* outage after recovery is reported fresh — exactly the behavior the issue asked for.
- Comments are appropriately scoped and genuinely explain the *why*; documentation updated in `backend-persistence.md`.

### Observation (follow-up, not blocking)
The flag resets on *any* successful apply, so for the `TooManyConnections` (53300) infrastructure failure — a transient reconnect-storm capacity limit where some writes succeed while others fail *concurrently* — an intermittent success resets the flag and the next failure re-logs at `Error`. In a flapping storm this can partially reintroduce the flood the issue set out to prevent (bounded, but up to ~one Error per failure-following-success per reclaim cycle). The fully-unreachable cases (08*, admin/crash shutdown, `NpgsqlException`, `TimeoutException`) throttle perfectly since no write succeeds during them. The chosen approach is a reasonable interpretation of the issue's suggested fix and handles the primary case well; the flapping refinement (e.g. a short time-window instead of a pure success-reset) is better as a small follow-up than as scope on this tech-debt PR. I'll open a follow-up issue.

### Test coverage
- `ProcessQueue_MultipleConcurrentInfrastructureFailuresInOnePass_LogsOnlyOneErrorAndRestAtDebug` covers the core throttle (exactly one Error + rest Debug, nothing dead-lettered, both left reserved).
- `ProcessQueue_InfrastructureOutageRecoversThenRecurs_LogsErrorAgainOnTheNextOutage` covers outage → real recovery (flag reset, event persisted) → second outage logging fresh.
- Both run as theories across the `InfrastructureFailures` matrix. Adequate for the domain logic changed here; no frontend/battle-logic surface touched.


---

PR #2248, opened 2026-07-21
Fix account-presence claim losing a same-character reconnect's live socket

### LINKED ISSUE #2235: Same-character socket replacement deletes the new socket's account-presence claim, reopening the two-live-characters window
## Context

Found during a codebase health review. The account-level "one live character per account" enforcement (#1817, spike #922, commit d84bf12/#2229) claims `AccountSocket_{userId}` with `PlayerIdValue(playerId)` — a value that is identical for the old and new sockets of the **same character**, unlike the per-player presence key whose value is the socket-unique socket id.

## Problem

On an ordinary same-character reconnect (tab refresh, network blip):

1. The new socket registers; `ClaimAccountPresence` CAS-sets the account key to the same player-id value (`SocketManagerService.cs:310-328`) — correct, no kick.
2. The replaced socket closes; `TeardownSocketRegistration` runs `CompareAndDelete(AccountSocketKey(userId), PlayerIdValue(playerId))` (`SocketManagerService.cs:217`). The value matches, so the **live new socket's account claim is deleted**.
3. The key stays absent until the new socket's next inbound message reclaims it via `RefreshSocketPresence` (`SocketManagerService.cs:341-345`) — 10s heartbeat cadence, potentially minutes under hidden-tab throttling (see the spike-#922 note on `SocketPresenceTtl`).

During that hole, a second tab/device selecting a **different character** on the same account finds the slot unset, claims it, and gets `null` back — so the first character's socket is never kicked. Both characters are then live, writing in-flight battle state into the shared `Session_{userId}` last-writer-wins — exactly the corruption the account-level claim exists to prevent. The state is also **self-sustaining**: `ReclaimAndForget` only extends the TTL of a differing value, so neither live socket ever reclaims or kicks the other until one disconnects.

The registration-rollback path (`SocketManagerService.cs:118-125`) shares the root cause: rolling back a failed registration deletes the account claim a still-live prior same-character socket depends on.

## Fix direction

Make the account-key value socket-unique (e.g. `"{playerId}:{socketId}"`, parsed on read), or condition the account-key release on the per-player release having succeeded (would need `ICacheService.CompareAndDelete` to return `bool` — `Game.Abstractions/Infrastructure/ICacheService.cs:43`).

Test gap to close alongside: `SocketManagerServiceTests.RegisterSocket_SameCharacterReconnect_*` asserts no stray `SocketReplaced` but never asserts the account key survives the old socket's teardown (the existing teardown test only covers the different-character case, whose values differ).

ginderjeremiah:
## Summary

`AccountSocket_{userId}` — the account-level "current live character" claim that enforces one-live-character-per-account (spike #922, #1817) — stored only the player id. That value is identical for the old and new socket of an ordinary **same-character reconnect** (tab refresh, network blip), unlike the per-player presence key whose value is the socket-unique socket id.

On a same-character reconnect:
1. The new socket registers and claims the account key with the same player-id value — correct, no kick.
2. The old socket's teardown runs `CompareAndDelete(AccountSocketKey(userId), PlayerIdValue(playerId))`. The value matches, so it deletes the **live new socket's** account claim.
3. The key stays absent until the new socket's next heartbeat reclaims it (up to the presence TTL, potentially minutes under hidden-tab throttling).

During that window, a second tab/device selecting a *different* character on the same account finds the slot unset, claims it, and neither socket is kicked — both characters end up live, writing in-flight battle state into the shared `Session_{userId}` last-writer-wins cache, exactly the corruption the account-level claim exists to prevent. The registration-rollback path shared the same root cause.

## Fix

The account-key value now embeds the claiming **socket's** id (`"{playerId}:{socketId}"`) instead of just the player id (`SocketManagerService.AccountPresenceValue`). A stale socket's teardown compare-and-delete can therefore never match — and never delete — a newer socket's still-live claim, while the "is this the same character reconnecting" check (used to decide whether to kick another character) now compares the *parsed player id*, not the raw string, so a same-character reconnect still correctly reports "no kick needed."

## Test plan

- [x] Added `RegisterSocket_SameCharacterReconnect_AccountPresenceClaimSurvivesOldSocketsTeardown` — the test gap the issue flagged: proves the account claim survives the old socket's teardown on an ordinary same-character reconnect, and that the new socket keeps working afterward.
- [x] Updated the existing `SocketManagerServiceTests` assertions that read the account-presence value's exact string to account for the new `"{playerId}:{socketId}"` format (via a new `ReadAccountPresencePlayerIdAsync` helper).
- [x] Updated `SocketCommandProcessorTests.UnRegisterSocket_UnsubscribeThrows_StillReleasesPresenceKey` (a unit test) for the same value-format change.
- [x] Full backend suite green: `dotnet build` + `dotnet test` across all four test projects (3,388 tests, 0 failures).
- [x] Documented the invariant in `docs/backend-sockets.md` (the account key's value is socket-unique, not just player-unique).

Closes #2235

---
_Generated by [Claude Code](https://claude.ai/code/session_01VYgFUWgnpecEonagG7eq2G)_


ginderjeremiah (commented):
## Review: account-presence claim surviving a same-character reconnect (#2235)

**Verdict:** Approve *(posted as a comment — GitHub won't accept a gating verdict from the PR author, but this is an unqualified approval)* — the fix is correct on every affected path, mirrors the per-player key's socket-unique semantics exactly, and closes the test gap the issue flagged.

### Analysis

The root cause is precisely diagnosed: `AccountSocket_{userId}` stored `PlayerIdValue(playerId)`, which is identical for the old and new socket of a same-character reconnect, so the old socket's `CompareAndDelete` in `TeardownSocketRegistration` matched and deleted the *new* live socket's claim. Embedding the socket id (`"{playerId}:{socketId}"`) makes a stale teardown's compare-value unmatchable against a newer socket's claim, while parsing the player id back out preserves the "same character → no kick" decision in `ClaimAccountPresence`.

I verified all three touch points move to the socket-unique value together, so they stay consistent:
- `ClaimAccountPresence` (`SocketManagerService.cs:315,321-322`) — writes `playerId:socketId` and compares the *parsed* player id, so a same-character reconnect still returns `null` (no kick).
- `TeardownSocketRegistration` (`:218`) — compare-and-deletes against its own `playerId:socketId`, the core fix.
- `RefreshSocketPresence` (`:345`) — reclaims with the same socket-unique value, so the live socket's heartbeat still resurrects its own claim.

The registration-rollback path the issue also called out is covered for free, since it routes through the same `TeardownSocketRegistration`, and now behaves symmetrically with the per-player key (both socket-unique) rather than deleting a sibling's claim.

`Split(':', 2)[0]` is robust regardless of socket-id contents because the numeric player id always precedes the first colon.

### Test coverage

Solid and matches the project's testing rules:
- The new `RegisterSocket_SameCharacterReconnect_AccountPresenceClaimSurvivesOldSocketsTeardown` closes exactly the gap #2235 named — it waits out the old socket's teardown and then asserts the claim still names the character *and* that socket B is still live. A fixed `Task.Delay` is the right tool here (proving something did **not** happen can't be polled for), and it mirrors the existing different-character teardown test.
- Existing assertions were correctly migrated to `ReadAccountPresencePlayerIdAsync`, and the `SocketCommandProcessorTests` unit expectation was updated to the new value format. No frontend counterpart is needed — this is purely backend socket presence, no battle logic.

### What's good

Tight, minimal fix that makes the account key consistent with the per-player key rather than inventing a new mechanism; the doc comments and `backend-sockets.md` update capture the invariant well. No findings, no nits.

---

PR #2251, opened 2026-07-21
Rebase progression editor partial-save failures instead of blanket re-seeding

### LINKED ISSUE #2238: Progression editor partial-save failure still blanket-reseeds both catalogues, discarding unrelated unsaved edits (port #2218's settled-diff recovery)
## Context

#2207/#2218 (commit 5de20b6) shrank the generic Workbench's partial-save failure blast radius: `EntityStore` now rebases the unsettled diff via `PersistRecovery`/`settledIds` (`UI/src/routes/admin/workbench/entity-store.svelte.ts:287-309`, `save-helpers.ts:146-157,232-248`) instead of reseeding everything. The progression editor re-implements save orchestration separately (documented in `docs/frontend-admin.md`) and never got the equivalent.

## Problem

`ProgressionStore.save()` is an up-to-8-step sequential POST pipeline (early prereqs → early child removals → path identities → prof identities → per-tier modifiers/rewards → combined prereq batch), so a mid-pipeline failure with `committed=true` is *more* likely here than in the generic Workbench — and the recovery path (`progression-store.svelte.ts:648-657`) does a wholesale `reseed()` of both `paths`/`profs` catalogues, silently dropping unsaved edits on every path/tier the failure never touched (e.g. a rename on sibling path B dies because tier A's reward POST 500'd).

Residual sub-case: if the post-failure reseed itself throws (the swallowed `.catch(() => {})`), the baseline stays behind server truth and a retry re-Adds the already-committed path/prof as a duplicate — the generic store guards this via `recovery.fresh`; progression has no equivalent.

The existing test (`progression-store.test.ts:714-767`) pins the reseed but only exercises a single edited record, so the sibling-edit data loss is not a pinned design decision.

## Fix direction

Thread a `settledIds`-style recovery through the pipeline steps (the diff/`childChanged` helpers are already shared), including a `recovery.fresh`-equivalent guard for the failed-reseed retry path. Add a multi-record partial-failure test asserting sibling edits survive.

ginderjeremiah:
## Summary

Closes #2238.

`ProgressionStore.save()`'s up-to-8-step pipeline (early prereqs → early child removals → path identities → prof identities → per-tier modifiers/rewards → combined prereq batch) previously handled any post-commit failure by blanket-reseeding both catalogues from server truth, silently discarding every unrelated pending edit the failure never touched or hadn't reached yet — e.g. a rename on sibling path B lost because tier A's reward POST 500'd.

This ports the generic Workbench's settled-diff recovery (`EntityStore`/`persistEntity`, #2207/#2218) to the progression editor's two-catalogue, multi-axis pipeline:

- `save()` now tracks, as it progresses, which paths/proficiencies (and which of a proficiency's identity / modifiers / rewards / prerequisites write axes) actually committed by the point a later step throws.
- On a post-commit failure, `rebaseAfterPartialFailure` re-fetches both catalogues fresh and rebases only the diff this save actually owns: a fully-settled record takes the fresh server copy (remapped from its local negative id if it was an add); everything else — a sibling this save never touched, or a write the failure struck before reaching — keeps its current local edit for a clean retry.
- Falls back to a full re-seed whenever a rebase can't be trusted (a committed add whose persisted id was never resolved, or the recovery refetch itself failing) — matching the prior safe behavior for those edge cases exactly.

Also closes the residual sub-case called out in the issue: the previous reseed-on-failure path used a `.catch(() => {})` swallow with no `recovery.fresh`-style guard, so a failed recovery refetch could leave the baseline behind server truth. The new fallback path is equivalent to the generic store's `recovery.fresh` guard.

## Test plan

- [x] Updated the existing single-record partial-failure test (`progression-store.test.ts`) to assert the new rebase behavior (the reward edit that never wrote now survives as pending, instead of being silently dropped) rather than the old blanket-reseed behavior it previously pinned.
- [x] Added a new multi-record test asserting a sibling path rename, an un-reached tier's modifier edit, and an un-reached tier's prerequisite edit all survive when a different tier's write fails mid-pipeline.
- [x] Verified the three existing "unresolved add" / "refetch itself fails" edge-case tests still pass unchanged (they exercise the same fallback paths, now reached via the new recovery method).
- [x] `npm run test:unit:ci` — full suite (3879 tests) passes.

## Docs

Added a note to `docs/frontend-admin.md`'s progression-editor bullet describing the new partial-failure rebase, mirroring the existing description of the generic Workbench's pattern.

https://claude.ai/code/session_01FSbnSeiPZkgf2gTogeZQdK

---
_Generated by [Claude Code](https://claude.ai/code/session_01FSbnSeiPZkgf2gTogeZQdK)_


ginderjeremiah (commented):
## Review: Rebase progression editor partial-save failures instead of blanket re-seeding (closes #2238)

**Verdict:** Approve _(submitted as a comment since I'm the PR author — GitHub won't accept a gating APPROVE on one's own PR)_ — faithfully ports the generic Workbench's settled-diff recovery to the progression editor's two-catalogue, multi-axis pipeline, with correct per-axis settle tracking and thorough tests. No blocking findings.

### What's good

- **Faithful to the source pattern.** `rebaseCatalogue` mirrors `EntityStore.rebaseAfterPartialFailure` almost line-for-line (settled → take fresh remapped by id; unsettled → keep local edit remapped onto its persisted id; concurrent-add-only-in-fresh → append), and `rebaseAfterPartialFailure`'s unresolved-add fallback is the exact analogue of `diffAddsResolved` + `recovery.fresh`. The two documented divergences from the generic store are real and handled correctly: (a) two catalogues instead of one, and (b) a proficiency's settle is a conjunction over its identity + each *changed* child axis rather than a single `settledIds` membership.
- **Per-axis settle tracking is sound.** `pathIdentityWritten`/`profIdentityWritten` are set unconditionally after their (atomic, all-or-nothing) identity POST, and the `modifiersWritten`/`rewardsWritten`/`prerequisitesWritten` sets all key on the *persisted* id consistently across the early-removal path (step 2, id ≥ 0), the child loop (step 7, `resolveId`), and both prerequisite batches (steps 1 & 8). `isSettled` re-derives "did every axis this save needed to write actually write" using the same `childChanged` predicate the pipeline uses — so a record can only adopt server truth once it's genuinely complete. I traced the mid-batch failure, identity-then-child-fail, and early-removal-then-fail cases; none prematurely settle.
- **No silent field drop.** Confirmed `pathIdentityDto`/`profIdentityDto` cover every mutable field (paths are identity-only; a proficiency's child collections are tracked per-axis), so a settled record taking the fresh copy can never discard an edited-but-non-identity field.
- **The `recovery.fresh` residual sub-case from the issue is closed properly.** The dedicated post-failure refetch (separate from the pipeline's own possibly-stale reads), the leave-untouched-on-refetch-failure guard, and the full-reseed-on-unresolved-add fallback together match the generic store's guarantees.
- **Docs updated** in `frontend-admin.md` mirroring the existing generic-Workbench description.

### Test coverage

Strong, and clearly above the bar. The suite drives the real `ProgressionStore` against an in-memory backend that reproduces the actual ordering guards (retire-gating, shrunken-MaxLevel), not a mock tautology.

- The updated single-record test now pins the *rebase* behavior — the committed modifier goes clean against the refreshed baseline while the un-sent reward survives as pending (`totalChanges === 1`), replacing the old blanket-reseed assertion it used to pin.
- The new multi-record test is exactly the scenario #2238 called out: a sibling path rename (committed atomically), an un-reached tier's modifier edit, and an un-reached tier's prerequisite edit all survive a mid-loop tier failure, each landing back as a distinct pending edit (`totalChanges === 3`).
- The three pre-existing "unresolved add" / "refetch itself fails" / "stale post-add refetch" edge-case tests still pass unchanged, now routed through the new recovery method — good regression coverage of the fallback paths.

CI is green (`all-checks-passed`, frontend + verify-frontend + all e2e browsers; backend jobs correctly skipped for a frontend-only change).

Merging.

---

PR #2252, opened 2026-07-21
Normalize username in Login to match CreateAccount

### LINKED ISSUE #2239: Login skips the username normalization CreateAccount applies
## Problem

`AccountService.CreateAccount` runs `UsernamePolicy.TryNormalize` (trims) before storing (`Game.Application/Services/AccountService.cs:57`), but `Login` passes the raw `username` to `LoginBackoffGuard.GetActiveBackoff`/`RegisterFailure` and `IUsers.GetUser` (`AccountService.cs:85-116`).

`Users.GetUser` (`Game.DataAccess/Repositories/Users.cs:144-150`) is an exact-match query and `LoginBackoffStore` keys on a SHA-256 of the raw string, so:

- an account persisted as `"bob"` cannot authenticate as `" bob "` via the API, and
- the backoff streak for one logical account fragments across un-normalized variants.

Not exploitable (the exact-match lookup fails before any password check, so padded probes never test the real hash), and the official client's stricter validation regex blocks whitespace anyway (`UI/src/routes/login/login-validation.ts:26`) — but the server-side domain policy is asymmetric between the two auth entry points and bites any direct API consumer.

## Fix

One `UsernamePolicy.TryNormalize` call at the top of `Login` (an un-normalizable input can fail fast as invalid credentials), plus a test covering the padded-username login.

ginderjeremiah:
## Summary

`AccountService.CreateAccount` normalizes (trims) the username via `UsernamePolicy.TryNormalize` before storing it, but `Login` was passing the raw, un-normalized username straight to `LoginBackoffGuard` and `IUsers.GetUser` (an exact-match query). This meant:

- An account persisted as `"bob"` could not authenticate as `" bob "` via the API (not exploitable — the exact-match lookup fails before any password check).
- The per-account backoff streak fragmented across un-normalized variants of the same logical account, since `LoginBackoffStore` keys on a hash of the raw string.

## Fix

`Login` now runs `UsernamePolicy.TryNormalize` at the top, before the backoff check, the database lookup, and any PBKDF2 work, and uses the normalized value everywhere it previously used the raw input. An un-normalizable input (e.g. too long, disallowed characters) can never match a stored username, so it fails fast as `InvalidCredentials` without touching the backoff store, the database, or the password hasher.

## Test plan

- [x] Added `Login_PaddedUsername_NormalizesAndSucceeds` — a padded variant of a stored username now authenticates successfully.
- [x] Added `Login_UnNormalizableUsername_FailsFastWithoutPasswordWork` — an unnormalizable username is rejected before any dummy/real password verification runs.
- [x] `dotnet test` (full solution) — 3392 tests passed, 0 failed.

Closes #2239

---
_Generated by [Claude Code](https://claude.ai/code/session_01GjfGB7SGMBuHqBaRcASZLC)_

### DIFF
--- a/Game.Application.Tests/Services/AccountServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/AccountServiceIntegrationTests.cs
@@ -5,6 +5,7 @@ using Game.Application.Auth;
 using Game.Application.Services;
 using Game.Core;
 using Game.Core.Battle;
+using Game.Core.Identity;
 using Game.Infrastructure.Database;
 using Game.TestInfrastructure.Base;
 using Game.TestInfrastructure.Fixtures;
@@ -328,6 +329,23 @@ namespace Game.Application.Tests.Services
         }
 
         [Fact]
+        public async Task Login_PaddedUsername_NormalizesAndSucceeds()
+        {
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+            var user = await TestDataSeeder.CreateUserAsync(context, "paddeduser", "paddedpass");
+
+            var accountService = CreateAccountService(scope.ServiceProvider);
+
+            // CreateAccount normalizes (trims) before storing, so Login must apply the same normalization —
+            // otherwise a padded variant of a stored username can never match the exact-match lookup.
+            var result = await accountService.Login("  paddeduser  ", "paddedpass");
+
+            Assert.True(result.Success);
+            Assert.Equal(user.Id, result.UserId);
+        }
+
+        [Fact]
         public async Task Login_UserWithMultiplePlayers_ReturnsAllSummaries()
         {
             using var scope = CreateScope();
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Normalize username in `Login` to match `CreateAccount` (closes #2239)

**Verdict:** Approve — the fix correctly closes the normalization asymmetry between the two auth entry points, with no regression risk and appropriate test coverage. (Submitted as `COMMENT` because I authored the PR; GitHub won't accept a gating verdict on one's own PR.)

The change threads a single `UsernamePolicy.TryNormalize` through the top of `Login` and uses the `normalized` value everywhere the raw input was previously used — the backoff key, the exact-match lookup, and the failure/reset registrations. This is exactly what #2239 asked for, and the reasoning holds up under scrutiny:

- **No regression for valid logins.** Stored usernames are always already-normalized (`CreateAccount` normalizes before persisting), and `TryNormalize` is idempotent on an already-normalized value, so any login that previously matched still matches.
- **No new enumeration side-channel.** The un-normalizable fast-fail path skips `VerifyDummy`/backoff, but a real account can never be un-normalizable, so that path leaks nothing about which accounts exist. Skipping backoff registration there is correct — there's no account to key a streak on, and the per-IP limiter still applies. The inline comment documents this well.
- **Layering is right.** The application service delegates the domain rule to `UsernamePolicy` in `Game.Core`, consistent with `CreateAccount`.

### Test coverage

- `Login_PaddedUsername_NormalizesAndSucceeds` covers the padded-variant login now resolving to the stored account.
- `Login_UnNormalizableUsername_FailsFastWithoutPasswordWork` asserts the fast-fail spends zero real/dummy PBKDF2 work, mirroring the existing timing-equalization tests and reusing `RecordingPasswordHasher`.

Coverage is sufficient for a change this size. (Optional, not requested: a test asserting a failed padded attempt shares the same backoff streak as the canonical username would exercise the fragmentation half of the fix directly — but it's obviously correct from `normalized` feeding every backoff call, so not worth adding here.)

### What's good

Tight, minimal diff that fixes the real asymmetry without touching anything else, and the security-sensitive reasoning (idempotence, enumeration, fast-fail) is captured clearly in the comments so a future edit won't naively "simplify" it away.

---

PR #2253, opened 2026-07-21
Bound tracked IPs per (user, device) in connection tracking

### LINKED ISSUE #2064: Harden device/login tracking: unbounded client-controlled fingerprint rows; DeviceInfo not scoped to the caller's account

### LINKED ISSUE #2240: Connection-tracking rows are unbounded in the IP dimension (the #2064 cap only bounds devices)
## Problem

`UserLogins.SaveConnection` (`Game.DataAccess/Repositories/UserLogins.cs:59-90`) evaluates the `MaxTrackedDevicesPerUser` cap only when `isNewDeviceForUser` is true. For a device the user already has any login row for, every previously-unseen IP inserts a new `(userId, ipAddress, deviceId)` row with no bound — so a mobile/VPN user on carrier-NAT rotation accrues `UserLogins` rows indefinitely per tracked device (up to 20 devices × unbounded IPs).

Same growth class #2064 fixed for the device dimension. IP is connection-derived (trusted-XFF), so this is slow organic telemetry-table bloat rather than attacker-drivable growth — no correctness impact.

## Fix direction

A per-(user, device) IP cap mirroring the device cap (e.g. evict-oldest by `LastConnection`), or a retention sweep on old rows. Decide which fits the connection-tracking table's diagnostic purpose better.

ginderjeremiah:
## Summary

`UserLogins.SaveConnection` (now `RecordConnection`) only evaluated the `MaxTrackedDevicesPerUser` cap when a connection came from a *new device* for the user. For a device the user already has a login row for, every previously-unseen IP inserted a new `(userId, ipAddress, deviceId)` row with no bound — so a mobile/VPN user on carrier-NAT rotation could accrue `UserLogins` rows indefinitely per already-tracked device (up to 20 devices × unbounded IPs).

This mirrors the fix #2064 made for the device dimension:

- Added `MaxTrackedIpsPerUserDevice` (= 20, matching the device cap) alongside the existing `MaxTrackedDevicesPerUser`.
- When a connection is from an IP the user hasn't seen on an *already-tracked* device, the new IP-count check runs the same way the device-count check does for new devices.
- Same best-effort policy as the device cap: once the cap is reached, the new IP is simply **not tracked** rather than erroring or evicting an older row — tracking is diagnostic telemetry, not a gate on play.

The existing device cap and ownership tests in `UserLoginsCapAndOwnershipIntegrationTests` still pass unchanged (their IP counts per device stay well under the new cap), confirming no regression to that behavior.

## Test plan

- [x] `dotnet test Game.Application.Tests` — full suite green (1061 passed), including the new `RecordConnection_PastIpCapForKnownDevice_StopsTrackingNewIpsButStillUpdatesKnownOnes` integration test, which verifies: the cap is reached with distinct IPs against one device, one more new IP is silently dropped without growing the row count, and an already-tracked `(user, IP, device)` combination still updates via the fast path rather than being blocked by the cap.

Closes #2240

---
_Generated by [Claude Code](https://claude.ai/code/session_01TH4a4ZWKk31k9yFZKo69mg)_

### DIFF
--- a/Game.Application.Tests/DataAccess/UserLoginsCapAndOwnershipIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/UserLoginsCapAndOwnershipIntegrationTests.cs
@@ -10,14 +10,15 @@ using Xunit;
 namespace Game.Application.Tests.DataAccess
 {
     /// <summary>
-    /// Integration coverage for the device/login-tracking hardening (#2064): a single account can't grow the
-    /// Devices/UserLogins tables without bound by cycling fresh fingerprints, and <c>SaveDeviceInfo</c> can
-    /// only enrich a device the caller actually has a tracked login for.
+    /// Integration coverage for the device/login-tracking hardening (#2064, #2240): a single account can't
+    /// grow the Devices/UserLogins tables without bound by cycling fresh fingerprints or IPs, and
+    /// <c>SaveDeviceInfo</c> can only enrich a device the caller actually has a tracked login for.
     /// </summary>
     [Collection("Integration")]
     public class UserLoginsCapAndOwnershipIntegrationTests : ApplicationIntegrationTestBase
     {
         private const int MaxTrackedDevicesPerUser = 20;
+        private const int MaxTrackedIpsPerUserDevice = 20;
         private const string UserAgent = "TestAgent/1.0";
         private const string Ip = "203.0.113.10";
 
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Bound tracked IPs per (user, device) in connection tracking (#2240)

**Verdict:** Approve _(posted as a comment — I'm the PR author, so GitHub won't take a gating approval from me)_ — clean, correct, and well-tested fix that closes the IP-dimension gap in the #2064 device cap.

### Assessment

The change closes exactly the gap #2240 describes: `RecordConnection` only bounded growth on the *new-device* branch, leaving an already-tracked device free to accrue an unbounded number of `(userId, ipAddress, deviceId)` rows under carrier-NAT/VPN IP rotation. The issue's "no correctness impact / slow organic bloat" framing is accurate, and a per-(user, device) IP cap mirroring the device cap is the right shape of the two options it floated.

The control flow is sound:
- The new `else if (device is not null)` branch is only reachable when `isNewDeviceForUser` is false, which by construction means `device` is non-null and the user already has a login for it — so the branch never misclassifies a genuinely new device. The explicit `device is not null` guard is redundant to that invariant but correctly lets the compiler flow-analyze `device.Id` without the null-forgiving operator, matching `CLAUDE.md`.
- Off-by-one is right: 20 distinct IPs are tracked and the 21st is silently dropped; the first IP of a new device is always tracked via the device-cap branch, so a legitimate new device is never starved.
- The early fast-path (`login is not null`) means the extra distinct-count query only runs for a genuinely new IP on a known device, not on every reconnect.

The best-effort "drop, don't evict/error" policy is consistent with the sibling device cap, and the shared constant value (20) plus the mirrored comment make the intent obvious.

### Test coverage

`RecordConnection_PastIpCapForKnownDevice_StopsTrackingNewIpsButStillUpdatesKnownOnes` covers the three cases that matter: reaching the cap, dropping the over-cap IP without growing the row count, and confirming an already-tracked `(user, IP, device)` combination still updates via the fast path rather than being blocked. The `IpForIndex` helper drawing from the RFC 5737 TEST-NET-2 range to avoid colliding with the class `Ip` constant is a nice touch. The pre-existing device-cap/ownership tests remain unchanged, confirming no regression.

### Notes (non-blocking)

- The `.Distinct()` on `IpAddress` is technically redundant — the login key `(userId, ipAddress, deviceId)` is unique, so IPs are already distinct per `(userId, deviceId)` — but it reads symmetrically with the device branch and is harmless/defensive. Fine to leave as-is.

No blocking findings and no should-fixes. Will merge once CI is green.

---

PR #2254, opened 2026-07-21
Guard SaveLogPreferences against no-op writes

### LINKED ISSUE #2241: SaveLogPreferences bypasses the no-op-change guard — every toggle enqueues a write-behind event per preference
## Problem

The socket command sends the whole preference list (`Game.Api/Sockets/Commands/SaveLogPreferences.cs:33-36`); `PlayerService.SaveLogPreferences` (`Game.Application/Services/PlayerService.cs:53-61`) loops all of them and is the **only** mutation command not using the `SaveIf` guard, and `Player.UpdateLogPreference` (`Game.Core/Players/Player.cs:417-430`) raises `LogPreferenceChangedEvent` even when the value is unchanged.

One toggle therefore enqueues N write-behind events (one per preference in the list, changed or not), and the drain side pays one `ExecuteUpdate` round-trip per event in `LogPreferenceChangedHandler`.

Distinct from #1739 (batching consecutive same-player events onto one `SaveChangesAsync` would reduce the round-trips but not the redundant events themselves).

## Fix

A value-equality guard in `Player.UpdateLogPreference` (return `bool`, raise only on change) and route the service through `SaveIf` like its siblings — collapsing a toggle to the single changed row. Low absolute impact (settings-UI frequency), but it's also an inconsistency with the "no mutation → no save" pattern every sibling command follows.

ginderjeremiah:
## Summary

`PlayerService.SaveLogPreferences` was the only mutation command not using the `SaveIf` guard: it always looped every preference in the incoming list through `Player.UpdateLogPreference` and always called `SavePlayer`, regardless of whether any value actually changed. Since `UpdateLogPreference` unconditionally raised `LogPreferenceChangedEvent`, one toggle from the client enqueued N write-behind events (one per preference in the list) and the drain side paid one `ExecuteUpdate` round-trip per event — even for preferences whose value didn't change.

## Changes

- `Player.UpdateLogPreference` (`Game.Core/Players/Player.cs`) now returns `bool` and is a no-op (no event raised) when the incoming value equals the existing one — mirroring the value-equality guard used elsewhere in the codebase (e.g. `TrySetFavorite`, `UnlockLesson`).
- `PlayerService.SaveLogPreferences` (`Game.Application/Services/PlayerService.cs`) now tracks whether *any* preference in the batch actually changed (via bitwise `|=`, so every preference is still applied — not short-circuited) and routes through the existing private `SaveIf` helper, so an all-unchanged batch performs no save and enqueues nothing.

This brings `SaveLogPreferences` in line with the "no mutation → no save" pattern every sibling command (`SetFavorite`, `UnlockItem`, `ApplyMod`, etc.) already follows.

Note: the frontend (`options-view.svelte.ts`) already computes and sends only the dirty preferences, so this mainly hardens the backend invariant itself (defense in depth against other/future callers) rather than fixing an everyday user-visible flood — but it closes the gap the issue called out.

## Test plan

- [x] `Game.Core.Tests/Players/PlayerTests.cs`: added `UpdateLogPreference_ExistingTypeUnchangedValue_ReturnsFalseAndDoesNotRaiseEvent`; updated the add/update tests to assert the new `true` return value.
- [x] `Game.Application.Tests/Services/PlayerServiceIntegrationTests.cs`: added `SaveLogPreferences_NoValuesChanged_DoesNotEnqueueWriteBehindEvents`, asserting the Redis write-behind queue stays empty when re-saving identical preference values.
- [x] `dotnet build` — succeeds, no warnings.
- [x] `dotnet test` for `Game.Core.Tests`, `Game.Application.Tests`, and `Game.Api.Tests` — all green (no regressions).

Closes #2241

---
_Generated by [Claude Code](https://claude.ai/code/session_015tpGR2Df29Npy8gPkdYKDq)_

### DIFF
--- a/Game.Application.Tests/Services/PlayerServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/PlayerServiceIntegrationTests.cs
@@ -765,12 +765,15 @@ namespace Game.Application.Tests.Services
                 new() { LogType = ELogType.Exp, Enabled = true },
             };
 
-            await playerService.SaveLogPreferences(player, preferences);
-            await DrainPlayerUpdateQueue(scope.ServiceProvider);
-
             var options = ConfigurationOptions.Parse(Containers.PubSubConnectionString);
             using var multiplexer = await ConnectionMultiplexer.ConnectAsync(options);
             var db = multiplexer.GetDatabase();
+
+            // The first save actually changes both preferences, so it must enqueue both events — proving the
+            // no-op assertion below isn't trivially true (e.g. from a broken queue rather than the no-op guard).
+            await playerService.SaveLogPreferences(player, preferences);
+            Assert.Equal(2, await db.ListLengthAsync(Constants.PUBSUB_PLAYER_QUEUE));
+            await DrainPlayerUpdateQueue(scope.ServiceProvider);
 
             // Re-submitting the same values is a no-op: no LogPreferenceChangedEvent should be raised or saved.
             await playerService.SaveLogPreferences(player, preferences);
--- a/Game.Application.Tests/Services/PlayerServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/PlayerServiceIntegrationTests.cs
@@ -13,6 +13,7 @@ using Game.TestInfrastructure.Helpers;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging.Abstractions;
+using StackExchange.Redis;
 using Xunit;
 
 namespace Game.Application.Tests.Services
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Guard `SaveLogPreferences` against no-op write-behind events (#2241)

**Verdict:** Approve — the change is correct, minimal, and brings `SaveLogPreferences` in line with the "no mutation → no save" pattern every sibling mutation command already follows. (Submitted as a comment since I'm the PR author; intent is **approve**.)

The fix does exactly what the issue asked: `Player.UpdateLogPreference` now returns whether the value actually changed and raises `LogPreferenceChangedEvent` only on a real change, and `PlayerService.SaveLogPreferences` routes through the existing `SaveIf` helper so an all-unchanged batch enqueues nothing and performs no save.

### What's good

- Value-equality guard mirrors the established siblings (`TrySetFavorite`, `UnlockLesson`) precisely — same shape, same domain-event discipline.
- `changed |= player.UpdateLogPreference(...)` correctly uses non-short-circuiting `|=` so every preference in the batch is still applied even after an earlier one changes; the inline comment explains exactly why `||` would be wrong, which is the kind of note that stops a later "simplification" from breaking it.
- Keeping `SaveLogPreferences` returning `Task` (not `Task<bool>`) is the right call — the socket caller (`Game.Api/Sockets/Commands/SaveLogPreferences.cs`) ignores the result, unlike the anti-cheat siblings whose callers need the bool.
- Domain logic stays in `Game.Core`; the application layer only orchestrates — consistent with the backend DDD layering.

### Test coverage

- Domain guard is directly unit-tested in `PlayerTests.cs`: new-type, changed-value, and the new unchanged-value case (`UpdateLogPreference_ExistingTypeUnchangedValue_ReturnsFalseAndDoesNotRaiseEvent`), plus the existing tests updated to assert the `true` return.
- Integration test `SaveLogPreferences_NoValuesChanged_DoesNotEnqueueWriteBehindEvents` verifies the end-to-end invariant against the real Redis write-behind queue.
- Coverage is adequate for the change. (Optional, non-blocking) the integration test could additionally assert the queue is *non-empty* after the first save before draining, to prove the no-op assertion isn't passing trivially — but the guard is already covered directly by the unit test, so this is genuinely take-it-or-leave-it.

No blocking findings, no should-fixes.

---

PR #2255, opened 2026-07-21
Game.Api response-shape consistency: guard ActiveSession, clamp negative Cooldown (#2242)

### LINKED ISSUE #2242: Game.Api response-shape consistency: ActiveSession skips the pre-selection guard; two success paths emit unclamped negative Cooldown
Two small consistency gaps in the API layer, found during a codebase health review — bundled because both are one-line alignments with existing sibling patterns.

## 1. `ActiveSession` silently probes player id 0 on a pre-selection token

`Status` and the socket handshake surface `NoPlayerSelected` for a pre-selection token, but `ActiveSession` (`Game.Api/Controllers/AuthController.cs:143-151`) skips that guard: with `TokenSelectedPlayerId == null`, `SelectedPlayerId` resolves to `0` and the endpoint does a pointless Redis GET on `PlayerSocket_0`, returning `Active = false`. Benign (player ids start at 1, so no false positive), but inconsistent with its siblings and a wasted round trip. Guard it like `Status` does.

## 2. `DefeatEnemy`/`BattleLost` success paths can return a negative `Cooldown`

`DefeatEnemy.cs:67` and `BattleLost.cs:52` emit the raw `(state.EnemyCooldown - now).TotalMilliseconds`, which goes negative if the command ran longer than the post-battle cooldown (e.g. a slow prefetch/save). The error paths and `NewEnemy`/`ChallengeBoss` all clamp (`IsOnCooldown(now)` / `> afterStart` guards) before emitting. Almost certainly harmless client-side, but the asymmetry across the six paths is a drift hazard — clamp the two success paths for consistency.

ginderjeremiah:
## Summary

Two small consistency fixes in `Game.Api`, both one-line alignments with existing sibling patterns:

1. **`ActiveSession` now guards on a pre-selection token.** `Status` and the socket handshake already surface `NoPlayerSelected` when `TokenSelectedPlayerId` is null, but `ActiveSession` skipped that guard, letting `SelectedPlayerId` resolve to `0` and probing a nonexistent `PlayerSocket_0` in Redis. `ActiveSession` now returns the same `NoPlayerSelected` error as `Status` for a pre-selection token, instead of doing the pointless round trip.
2. **`DefeatEnemy`/`BattleLost` success paths now clamp `Cooldown` to 0.** Both emitted the raw `(state.EnemyCooldown - now).TotalMilliseconds`, which can go negative if the command runs longer than the post-battle cooldown (e.g. a slow prefetch/save). Their own error paths, and `NewEnemy`/`ChallengeBoss`, already guard with `state.IsOnCooldown(now)` before emitting — the two success paths now do the same.

Closes #2242 — implements exactly the two fixes described there.

## Test plan

- [x] Added `AuthControllerTests.ActiveSession_PreSelectionToken_ReturnsNoPlayerSelectedNotProbingPlayerZero`, mirroring the existing `Status_PreSelectionToken_ReturnsNoPlayerSelectedNotOpaque404`.
- [x] Added `DefeatEnemySocketTests.DefeatEnemy_BattleCompletedLongBeforePostBattleCooldownWindow_CooldownClampedToZero`, which backdates the battle start far enough that the victory's cooldown anchor (`battleCompletedAt + PostBattleCooldown`) is already in the past — deterministically reproducing the unclamped-negative-`Cooldown` scenario and asserting it's now clamped to `0`.
- [x] `BattleLost`'s identical clamp isn't independently regression-tested: unlike `DefeatEnemy`'s victory path (anchored to the server-replayed `battleCompletedAt`), a loss anchors its cooldown to `now + PostBattleCooldown` at computation time, so reproducing its negative case deterministically would require injecting a real ~5s delay (there's no clock abstraction in this codebase to fake it) rather than backdating state. The fix is the same one-line ternary already covered by the `DefeatEnemy` test and the existing `NewEnemy`/`ChallengeBoss` clamp pattern.
- [x] `dotnet test` — full backend suite green (1417 + 68 + 1062 + 850 = 3397 tests, 0 failures).
- No frontend changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q73ZE94vP7qps6iLRKQHei)_

### DIFF
--- a/Game.Api.Tests/Integration/AuthControllerTests.cs
+++ b/Game.Api.Tests/Integration/AuthControllerTests.cs
@@ -351,6 +351,28 @@ namespace Game.Api.Tests.Integration
         }
 
         [Fact]
+        public async Task ActiveSession_PreSelectionToken_ReturnsNoPlayerSelectedNotProbingPlayerZero()
+        {
+            // Mirrors Status_PreSelectionToken_ReturnsNoPlayerSelectedNotOpaque404 (#2242): a pre-selection
+            // token must surface the same distinguishable NoPlayerSelected category rather than resolving
+            // SelectedPlayerId to 0 and probing that socket's presence.
+            using var scope = CreateScope();
+            var context = scope.ServiceProvider.GetRequiredService<GameContext>();
+            var user = await TestDataSeeder.CreateUserAsync(context, "playerlessactive", "pass");
+
+            var client = Factory.CreateClient();
+            TestAuthHelper.AddAuthHeader(client, user.Id);
+
+            var response = await client.GetAsync("/api/Auth/ActiveSession", CancellationToken);
+
+            Assert.Equal(HttpStatusCode.Conflict, response.StatusCode);
+            var result = await response.Content.ReadFromJsonAsync<ApiResponse<ActiveSessionResult>>(CancellationToken);
+            Assert.Null(result?.Data);
+            Assert.NotNull(result?.ErrorMessage);
+            client.Dispose();
+        }
+
+        [Fact]
         public async Task ActiveSession_ValidTokenWithNoSessionCache_RehydratesAndReturnsResult()
         {
             // The pre-game active-session takeover warning is the user-visible breakage from #693: an evicted
--- a/Game.Api.Tests/Integration/DefeatEnemySocketTests.cs
+++ b/Game.Api.Tests/Integration/DefeatEnemySocketTests.cs
@@ -114,6 +114,52 @@ namespace Game.Api.Tests.Integration
         }
 
         [Fact]
+        public async Task DefeatEnemy_BattleCompletedLongBeforePostBattleCooldownWindow_CooldownClampedToZero()
+        {
+            var (userId, playerId) = await SeedAndLoginAsync();
+
+            var wsClient = Factory.Server.CreateWebSocketClient();
+
+            await using (var socketClient1 = new TestSocketClient())
+            {
+                await socketClient1.ConnectAsync(wsClient, userId);
+
+                var newEnemyResponse = await socketClient1.SendCommandAsync<NewEnemyModel>(
+                    "NewEnemy", new { NewZoneId = (int?)null });
+                Assert.Null(newEnemyResponse.Error);
+
+                await socketClient1.CloseAsync();
+            }
+
+            // Backdate the battle start well past its own replay duration plus the 5s post-battle cooldown, so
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: Game.Api response-shape consistency — guard `ActiveSession`, clamp negative `Cooldown` (#2242)

**Verdict:** Approve — both fixes land exactly as the issue specifies, mirror their sibling patterns precisely, and are correct. (Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR — the intended verdict is Approve.)

I verified each change against the surrounding code rather than the description:

- **`ActiveSession` guard** (`AuthController.cs:151-154`) — byte-for-byte consistent with `Status`'s guard (`:121-124`): same `TokenSelectedPlayerId is null` check, same `"No character selected."` message, same `ApiErrorCategory.NoPlayerSelected`, which `ErrorStatusFilter.cs:40` maps to `409 Conflict`. Fixes the pointless `PlayerSocket_0` probe.
- **`Cooldown` clamps** (`DefeatEnemy.cs:67`, `BattleLost.cs:52`) — `IsOnCooldown(now)` is `EnemyCooldown > now` (`PlayerState.cs:80-83`), so the ternary emits positive remaining ms or `0`, never negative. This brings the two success paths in line with their own error paths and with `NewEnemy`/`ChallengeBoss`. I confirmed `NewEnemy.cs:29`'s unclamped subtraction is inside an `IsOnCooldown` guard, so it can't go negative — all six paths are now consistent.

### Test coverage
- `ActiveSession_PreSelectionToken_...` mirrors the existing `Status` test and asserts the 409/`NoPlayerSelected` shape.
- `DefeatEnemy_BattleCompletedLongBeforePostBattleCooldownWindow_CooldownClampedToZero` deterministically reproduces the negative case by backdating the battle start 30 min and asserts `Cooldown == 0` — a genuine regression test for the fix.
- `BattleLost`'s clamp has no independent regression test, which the PR body calls out honestly: a loss anchors its cooldown to `now + PostBattleCooldown`, so reproducing the negative case would need a real ~5s delay (no clock abstraction exists to fake it) rather than backdating. Since it's the identical one-line ternary already covered by the `DefeatEnemy` test and the `NewEnemy`/`ChallengeBoss` pattern, this is an acceptable gap, not a blocker.

### What's good
Tightly scoped tech-debt fix — no scope creep, each change is a one-line alignment with an established pattern, and the comments explain the *why* (pre-selection flow state, cooldown-anchor drift) rather than restating the code. No blocking findings, no should-fixes, no nits.

---

PR #2265, opened 2026-07-22
Add periodic reconciliation sweep for reference-cache invalidation

### LINKED ISSUE #1980: Add a periodic reconciliation sweep backstop for reference-cache invalidation broadcasts

ginderjeremiah:
## Summary

Follow-up to #1888, closing the residual gap it split out into this issue. Redis pub/sub delivery to `CoalescingReferenceCacheReloader`'s subscribers is inherently at-most-once: an instance that's mid-reconnect (or otherwise transiently disconnected) when an admin write's invalidation broadcasts simply never receives it — no error, no retry, no reconciliation anywhere in the pipeline. That instance then serves stale reference data indefinitely, until its own next admin write or a restart.

- Added `ReferenceCacheReloadPolicy.ReconciliationInterval` (`TimeSpan?`, defaulting to 5 minutes on `Default`). `null` disables the periodic sweep entirely — this is the implicit default for every existing test policy, so none of them needed updating.
- `CoalescingReferenceCacheReloader.RunAsync` now races its existing signal wait against the reconciliation interval (`WaitForNextSweepAsync`). If the interval elapses with no signal at all, it sweeps anyway, logged at `Information` distinctly from a signal-triggered sweep so an operator can tell a reconciliation sweep from a reactive one. Each sweep (signal or periodic) restarts the interval from zero — this is "no sweep in N minutes," not a fixed wall-clock cadence.
- The wait helper explicitly observes both sides of the `Task.WhenAny` race (rather than leaving the loser's cancellation unobserved) and re-checks the outer `cancellationToken` afterward, so a genuine shutdown mid-wait still propagates correctly regardless of which task "won."
- Documented the design in `docs/backend.md` → Reference-data cache reload, replacing the old "known residual gap (#1980)" callout.

## Design decisions made without a round-trip

The issue flagged a few open questions and left a concrete suggested direction; I resolved them as follows rather than pausing for sign-off, since none seemed to need it:
- **Interval default: 5 minutes.** Bounds worst-case staleness to that window per instance while keeping the added DB load (a full reference-cache reload sweep per instance, per interval, only when otherwise idle) low.
- **Always-on, not adaptive/configurable per-deployment.** Kept it simple — a single default constant — consistent with how `DebounceWindow`/retry policy are configured today (no per-deployment overrides exist for those either).
- **Lives in `CoalescingReferenceCacheReloader.RunAsync`**, racing the existing signal wait as the issue suggested, rather than a separate hosted service.

## Test plan

- [x] Added `ReferenceCacheReloadPolicyTests` cases for the new constructor parameter (default-disabled, positive-allowed, non-positive-throws).
- [x] Added `CoalescingReferenceCacheReloaderTests` cases: periodic sweep fires with no signal, stays disabled when the interval is unset (regression guard for every other test in the suite), and a reactive sweep resets the periodic wait rather than immediately double-firing.
- [x] `dotnet test` — full backend suite green: 3430 tests passed (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), 0 failed.

Closes #1980

---
_Generated by [Claude Code](https://claude.ai/code/session_01SV5x1Zg9wH9PbC5ELq5e8m)_


Game.DataAccess/CoalescingReferenceCacheReloader.cs
```diff
@@ -37,18 +39,35 @@ public void NotifyChanged()
...
+
                     while (_signal.Reader.TryRead(out _))
                     {
                     }
 
+                    if (reconciling)
+                    {
+                        logger.LogInformation(
+                            "No reference-cache change signal received in {Interval}; running a periodic reconciliation sweep.",
+                            policy.ReconciliationInterval);
```
ginderjeremiah, line 71: Nit (non-blocking): the reconciliation sweep is an unconditional flat timer, so this fires on **every** idle instance every `ReconciliationInterval` (5 min) forever — regardless of whether a signal was actually missed or anything changed. Two small consequences worth a thought:

- At `Information` level this becomes steady background log volume on any idle instance (~288 lines/day/instance). Routine periodic background activity is usually a better fit for `Debug`, reserving `Information`/`Warning` for genuinely notable events.
- The wording — "No reference-cache change signal received in {Interval}" — reads like a fault, but it's the normal idle steady-state (no admin writes anywhere in the cluster), not evidence this instance actually dropped a message. An operator skimming logs could reasonably misread it as a delivery problem.

Consider dropping to `Debug`, or rewording so it doesn't imply something went wrong. The distinct-log-line ask in #1980 is still satisfied at `Debug`. Take it or leave it.

---

ginderjeremiah (commented):
## Review: periodic reconciliation sweep backstop for reference-cache invalidation (closes #1980)

**Verdict: Approve** — faithfully implements the issue, the concurrency logic is correct and well-tested, and the design decisions the issue flagged are consciously resolved and documented. One optional nit below (posted inline), no blockers.

_(Submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author on their own PR — the effective verdict is Approve.)_

### What's good
- **The `Task.WhenAny` race is handled carefully.** `WaitToReadAsync` peeks without consuming, so the existing drain-then-reload path still works unchanged. The linked-CTS cancel + dual `ObserveCancellationAsync` + explicit `cancellationToken.ThrowIfCancellationRequested()` correctly covers the tricky case where the *outer* token cancels whichever task `WhenAny` happened to return — no unobserved task exceptions, and shutdown still propagates regardless of which side "won." The comments earn their length here.
- **Coalescing invariant preserved.** A signal arriving mid-sweep survives in the capacity-1 channel and triggers a follow-up sweep; a signal landing during a reconciliation window is drained before the reload, so no change is lost. Null interval → `Timeout.InfiniteTimeSpan` → clean degradation to today's signal-only behavior.
- **Correct layering.** Lives in `Game.DataAccess` (infra), no domain logic leak, and `docs/backend.md` is updated to replace the old "residual gap (#1980)" callout rather than leaving stale docs behind.
- **Design decisions match the issue.** Interval default, always-on (vs. adaptive), and location-in-`RunAsync` are all resolved as the issue suggested and documented, including the accepted trade-off of a periodic reload query per idle instance.

### Test coverage
Solid and sufficient for this infra change (no battle-logic mirroring needed):
- Policy: default-disabled when omitted, positive allowed, non-positive (`0`/`-1`) throws with the right `ParamName`, and the 5-min default is pinned.
- Reloader: periodic sweep fires with no signal at all, stays disabled when the interval is unset (a real regression guard for the rest of the suite's signal-only assumptions), and a reactive sweep resets the periodic wait rather than double-firing.

### Nits
- `CoalescingReferenceCacheReloader.cs:66` — the reconciliation sweep is an unconditional timer, so it logs at `Information` on every idle instance every interval, forever, with wording that reads like a fault when it's actually normal idle steady-state. Consider `Debug`, or rewording. Optional — details inline.


---

PR #2301, opened 2026-07-22
Queue a screen-anchored lesson as unread instead of clobbering an active tour

### LINKED ISSUE #2269: Screen-visit tutorial trigger replaces a tour just opened from the Help screen, marking the wrong lesson read
## Problem

There is no arbitration between tour sources: `tutorialTour.play()` replaces the active lesson unconditionally (`UI/src/stores/tutorial-tour.svelte.ts:15-17`), and the screen-anchored trigger effect in `UI/src/routes/game/+page.svelte` re-runs on **every** `currentScreen` change — including a navigation initiated by `openLesson` itself (`UI/src/lib/engine/tutorials.ts:24-31` calls `navigation.requestScreen` + `tutorialTour.play`).

## Failure scenario

A mechanic-anchored lesson A is queued unread, hosted on a screen the player has never visited (so that screen's own ScreenVisit lesson B is still **locked**). The player clicks A in the Help screen (#1589):

1. `openLesson(A)` → `play(A)` + `requestScreen(A.screenKey)`.
2. The shell switches screens → the screen-trigger effect runs → `evaluateScreenTrigger` finds locked ScreenVisit lesson B → `openLesson(B)` → `play(B)` **replaces tour A before the player sees a single step**.
3. Dismissing marks **B** read (`closeTutorialTour` reads `tutorialTour.activeLesson`); A stays unread and never played — the click appears to open the wrong tutorial.

This path became reachable with the Help screen (#2232): before it, there was no manual open flow for a lesson hosted on an unvisited screen. `tutorials.test.ts` has no active-tour case.

## Suggested fix

Have `evaluateScreenTrigger` skip (or unlock-as-unread instead of play) when `tutorialTour.activeLesson` is already set, and add a test covering the Help-open → screen-trigger collision.

ginderjeremiah:
## Summary

There was no arbitration between the two tutorial-tour sources: `tutorialTour.play()` replaces the active lesson unconditionally, and the screen-anchored trigger effect in `routes/game/+page.svelte` re-runs on **every** `currentScreen` change — including a navigation that `openLesson` itself just triggered.

Failure scenario: a mechanic-anchored lesson A is queued unread on a screen the player has never visited (so that screen's own ScreenVisit lesson B is still locked). The player opens A manually from the Help screen (#1589):

1. `openLesson(A)` → `tutorialTour.play(A)` + `navigation.requestScreen(A.screenKey)`.
2. The shell switches screens → the screen-trigger effect re-runs → `evaluateScreenTrigger` finds locked lesson B → `openLesson(B)` → `tutorialTour.play(B)` **replaces tour A before the player sees a single step**.
3. Dismissing marks **B** read; A stays unread and never plays.

## Fix

`evaluateScreenTrigger` (`UI/src/lib/engine/tutorials.ts`) now checks whether a tour is already playing before opening a newly-triggered screen-anchored lesson:

- If another tour is active **and** the candidate lesson has steps, it's unlocked (queued as unread) instead of played — mirroring the mechanic-anchored trigger's existing AFK-safe behavior. The screen genuinely did just become active, so unlocking (rather than silently skipping) is correct.
- A zero-step lesson still marks itself read immediately regardless, since that branch never touches `tutorialTour`/`navigation` and is always safe.

Documented the new arbitration rule in `docs/frontend-tutorials.md`.

## Test plan

- [x] New unit tests in `UI/src/tests/lib/engine/tutorials.test.ts`: a locked lesson queues as unread (not played) when another tour is active, and a zero-step lesson still marks itself read while another tour is active.
- [x] `npm run lint` — clean.
- [x] `npm run test:unit:ci` — 310 files / 3881 tests passed.

Closes #2269

---
_Generated by [Claude Code](https://claude.ai/code/session_0198emUix9FKu6kTa9Hd8mr1)_


ginderjeremiah (commented):
## Review: Screen-anchored trigger no longer clobbers an active tour (#2269)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't accept a gating event on my own PR) — the fix is correct, well-scoped, and properly tested.

I traced the failure path from the issue against the new code: manually opening lesson A from the Help screen sets `tutorialTour.activeLesson = A`, the screen switch re-runs `evaluateScreenTrigger`, which now finds locked lesson B, sees `B.steps.length > 0 && tutorialTour.activeLesson`, and unlocks B as unread instead of calling `openLesson(B)`. A keeps playing and is the one marked read on dismissal. Correct.

### Design notes (no action needed)
- Unlocking-as-unread (rather than skipping outright) is the right call and deliberately mirrors the mechanic-anchored trigger's AFK-safe behavior — a screen that genuinely just became active shouldn't silently drop its lesson. Documented in `docs/frontend-tutorials.md`.
- The zero-step branch correctly falls through to `openLesson`, which marks it read immediately without touching `tutorialTour`/`navigation` — consistent with the existing #1638 graceful-degradation path, so it can't interfere with the active tour.
- No re-trigger loop: once B is unlocked, `isLocked(B.id)` is false, so a later re-run of the effect won't rematch it.

### Test coverage
Adequate and faithful. The two new cases cover both new branches (locked-with-steps queues as unread; zero-step still marks read while a tour plays), and the hoisted `playerManager` stub replicates the real `unlockLesson`/`markLessonRead` semantics (idempotent, absent-entry-tolerant) rather than hand-waving with a bare spy — so the assertions reflect real gating.

### What's good
Tight, minimal diff that fixes the root cause at the single arbitration point, with the reasoning captured both in the code comment and the tutorials doc. No blocking findings, no should-fixes, no nits.


---

PR #2302, opened 2026-07-22
Settle queued commands on a WebSocket constructor throw

### LINKED ISSUE #2270: api-socket: a synchronous WebSocket-constructor throw (invalid subprotocol token) strands queued commands unsettled
## Problem

Since #2216 the access token is passed as a WebSocket subprotocol: `new WebSocket('/socket', [accessToken])` (`UI/src/lib/api/api-socket.ts:176`). Unlike the old query-string carry, the constructor **throws synchronously** (`SyntaxError`) if the stored token contains any character outside the RFC 6455 `token` grammar. When it throws, `openSocket()` rejects, `processCommandQueue`'s `await this.ensureSocket()` rethrows into a `void`-ed call → unhandled rejection, and the queued command is **never settled** — the caller awaits forever.

That violates the transport's documented invariant (docs/frontend-sockets.md: every failure resolves with `response.error`; every terminal path settles the queue). Every other terminal path in the file scrupulously settles.

## Likelihood

Low organically — a real backend JWT is always valid base64url, so this needs a corrupted or hand-edited `localStorage` token pair. Filed because it's a contract violation on the recent subprotocol change with a ~5-line fix; reasonable to fold into the next touch of `api-socket.ts` instead of fixing standalone.

## Suggested fix

Wrap the constructor in try/catch and route the throw to the existing `settleQueuedRequests(CONNECTION_LOST_ERROR)` + `handleAuthFailure()` path (a token that can't even form a handshake is dead, so treating it as an auth failure is right).

ginderjeremiah:
## Summary

Since #2216, `openSocket()` passes the access token as a WebSocket subprotocol: `new WebSocket('/socket', [accessToken])`. Unlike the old query-string carry, the constructor **throws synchronously** (`SyntaxError`) if the stored token contains any character outside the RFC 6455 `token` grammar. When that happened, `openSocket()` returned via an uncaught exception, `processCommandQueue`'s `await this.ensureSocket()` rethrew into a `void`-ed call (unhandled rejection), and the queued command was never settled — the caller awaited forever, violating the transport's documented invariant that every terminal path settles the queue (`docs/frontend-sockets.md`).

## Change

`UI/src/lib/api/api-socket.ts`: wrap the `new WebSocket(...)` call in a try/catch. On a constructor throw, treat it like the existing "definitively rejected, no adoptable pair" branch just above it — a token that can't even form a handshake is dead:
- stop the keepalive interval
- settle the queued-but-unsent commands with the connection-lost error
- route to `handleAuthFailure()`

## Test plan

- [x] Added a unit test (`UI/src/tests/lib/api/api-socket.test.ts`) that forces the mocked `WebSocket` constructor to throw and asserts the queued command settles with an error, the keepalive is torn down, and `handleAuthFailure` is called once.
- [x] `npm run test:unit:ci` — 3882/3882 passing

This is a frontend-only change; no backend build/tests were affected.

Closes #2270

---
_Generated by [Claude Code](https://claude.ai/code/session_01RiooTCn1yfPtBVg6ESxBah)_

### DIFF
--- a/UI/src/lib/api/api-socket.ts
+++ b/UI/src/lib/api/api-socket.ts
@@ -173,7 +173,22 @@ export class ApiSocket {
 		// carried as a WebSocket subprotocol instead of a query-string parameter — a query string ends up
 		// verbatim in any downstream access log (e.g. the edge proxy's), while a subprotocol is a
 		// handshake header a default access log format never records (#2211).
-		const socket = accessToken ? new WebSocket('/socket', [accessToken]) : new WebSocket('/socket');
+		let socket: WebSocket;
+		try {
+			socket = accessToken ? new WebSocket('/socket', [accessToken]) : new WebSocket('/socket');
+		} catch (ex) {
+			// The constructor throws synchronously (rather than failing the handshake asynchronously) if
+			// accessToken contains a character outside the RFC 6455 subprotocol token grammar — only
+			// possible with a corrupted/hand-edited stored token, since a real backend JWT is always valid
+			// base64url. A token that can't even form a handshake is dead, so this is unrecoverable like the
+			// "definitively rejected" branch above: settle the queue and route to re-auth rather than
+			// leaving the caller awaiting a connection that will never open.
+			console.error('WebSocket construction failed, likely due to a malformed access token', ex);
+			this.stopPingInterval();
+			this.settleQueuedRequests(CONNECTION_LOST_ERROR);
+			handleAuthFailure();
+			return;
+		}
 		this.socket = socket;
 		socket.onopen = this.onStart.bind(this);
 		socket.onmessage = this.receiveResponse.bind(this);
--- a/UI/src/tests/lib/api/api-socket.test.ts
+++ b/UI/src/tests/lib/api/api-socket.test.ts
@@ -128,6 +128,25 @@ describe('ApiSocket', () => {
 			// A never-logged-in caller (no prior refresh token) must not be routed to the auth-failure path.
 			expect(handleAuthFailure).not.toHaveBeenCalled();
 		});
+
+		it('settles the queue and routes to auth failure when the constructor throws on a malformed token', async () => {
+			// The real WebSocket constructor throws synchronously (a SyntaxError) if the subprotocol string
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: fix for #2270 — settle queued commands on a synchronous WebSocket-constructor throw

**Verdict:** Approve — the change closes the contract violation cleanly, mirrors the existing terminal branches exactly, and is covered by a focused unit test. (Posted as a comment because I'm the PR author; treat it as an approval.)

The fix wraps `new WebSocket('/socket', [accessToken])` in a try/catch and, on a synchronous `SyntaxError` (a subprotocol token outside the RFC 6455 grammar), routes to the same terminal path as the "definitively rejected, no adoptable pair" branch just above it: `stopPingInterval()` → `settleQueuedRequests(CONNECTION_LOST_ERROR)` → `handleAuthFailure()` → `return`. This is exactly the right treatment — a token that can't even form a handshake is dead, and it restores the transport invariant from `docs/frontend-sockets.md` that every terminal path settles the queue.

A nice secondary effect: because `openSocket()` now returns normally instead of throwing, `ensureSocket()`'s `connecting` promise resolves cleanly rather than rejecting into the `void`-ed `processCommandQueue()` call, which removes the unhandled rejection entirely. After the return, `processCommandQueue` resumes against an already-drained queue and a `this.socket` that was never reassigned, so nothing is sent — correct.

### Test coverage

Adequate and consistent with the file's conventions. The new test forces the mocked constructor to throw via `mockImplementationOnce`, then asserts the three observable outcomes: `handleAuthFailure` called once, `pingIntervalId` torn down to `null`, and the queued command resolving with an `error` (the resolve-with-error contract). Matches the surrounding test style (`setTokens`/`flushMicrotasks`/`internals`).

### What's good

- Faithfully reuses the established terminal-branch pattern rather than inventing a new one.
- The `console.error` on the exception is a sensible diagnostic for a genuinely unexpected (corrupted-token) path and is consistent with the `console.warn` used in the other terminal branches.
- Comments explain the non-obvious "constructor throws *synchronously*" behavior, which is the whole reason this bug exists — appropriately detailed, not excessive.

---

PR #2303, opened 2026-07-22
Isolate PlayerUpdateBatch OnFlushed callback faults from flush failures

### LINKED ISSUE #2271: PlayerUpdateBatch: a fault in an OnFlushed callback fails a flush that already published, skipping the player cache write
## Problem

`PlayerUpdateBatch.FlushAsync` (`Game.DataAccess/PlayerUpdateBatch.cs:77-102`) runs the `OnFlushed` callbacks unisolated **after** the publish has succeeded and `_events.Clear()` has run. The deferred progress cache advance (`PlayerProgressRepository.AdvanceCache` → `HashSetIfExistsAndForget`) issues a synchronous StackExchange.Redis call that can throw synchronously (e.g. `RedisConnectionException` if the connection drops in the window after the awaited publish). That exception propagates out of `FlushAsync`, is wrapped as `PlayerPersistenceFlushFailedException` in `SavePlayer`, and **skips the player cache-blob write** (`Game.DataAccess/Repositories/PlayerRepository.cs:~180`) — even though the batch durably landed on the queue.

The forced reload then reads a cache blob **behind** the durably-enqueued events; for insert-only events (item/skill/mod/lesson unlocks) the divergence persists until cache eviction — exactly the failure class #2098/#2106 closed on the dispatch-fault and cancellation paths.

Secondary untidiness: a mid-way callback throw leaves `_onFlushed` uncleared, so remaining callbacks are skipped and survivors re-run on a same-scope retry (harmless today, but accidental).

## Suggested fix

Run each flushed callback isolated (log-and-continue) — the cache advance is best-effort/fire-and-forget by design, so a callback fault must never fail a flush that already succeeded. Add a test pinning that a throwing callback doesn't prevent the cache write or the other callbacks.

ginderjeremiah:
## Summary

`PlayerUpdateBatch.FlushAsync` (`Game.DataAccess/PlayerUpdateBatch.cs`) ran its deferred `OnFlushed` callbacks unisolated **after** the publish had already succeeded and the event buffer was cleared. The batched progress-cache advance (`PlayerProgressRepository.AdvanceCache` → `HashSetIfExistsAndForget`) issues a synchronous StackExchange.Redis call that can throw synchronously (e.g. a `RedisConnectionException` from a connection drop in the window right after the awaited publish). That exception propagated out of `FlushAsync`, got caught in `PlayerRepository.SavePlayer` and wrapped as `PlayerPersistenceFlushFailedException` — which **skipped the player cache-blob write** even though the batch had durably landed on the write-behind queue. For insert-only events (item/skill/mod/lesson unlocks) that divergence between DB/queue and cache persists until the cache blob is evicted — the same failure class #2098/#2106 closed on the dispatch-fault and cancellation paths.

There was also a secondary bug: a mid-way callback throw left `_onFlushed` uncleared, so any callbacks registered before the throw were skipped and, on a same-scope retry, would re-run alongside newly registered ones.

## Fix

- `PlayerUpdateBatch.RunFlushedCallbacks` now runs each callback isolated (try/catch, log-and-continue) — mirroring how `DomainEventDispatcher` isolates domain-event handlers. The cache advance is best-effort by design and the flush it rides has already durably succeeded by this point, so a callback fault must never be mistaken for a flush failure.
- All registered callbacks are now always cleared after running, fixing the secondary bug above as a natural consequence (every callback always runs exactly once, then the list is cleared).
- `PlayerUpdateBatch` now takes an `ILogger<PlayerUpdateBatch>` constructor dependency (it's DI-registered as a concrete scoped type, so this resolves automatically everywhere it's constructed via `IServiceProvider`).
- Added a short note to `docs/backend-persistence.md` documenting the isolation invariant.

## Test plan

- [x] New unit test `RunFlushedCallbacks_OneCallbackThrows_LogsAndStillRunsAndClearsTheRest` — pins that a throwing callback is logged, the remaining callbacks still run, and all callbacks (including the faulted one) are cleared.
- [x] New unit test `FlushAsync_OnFlushedCallbackThrows_StillCompletesSuccessfully` — pins the end-to-end path `SavePlayer` actually exercises: `FlushAsync` must return normally (not throw) when a deferred callback faults, so the caller never wraps it as a flush failure.
- [x] Updated existing `PlayerUpdateBatchTests` construction sites for the new constructor parameter.
- [x] Full backend suite green: `dotnet test` — 3432 tests passed across all four test projects (`Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Api.Tests`, `Game.Application.Tests`).

Closes #2271

https://claude.ai/code/session_01Fiw8qwuhKpcbNnjLuswN1m

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fiw8qwuhKpcbNnjLuswN1m)_


ginderjeremiah (commented):
## Review: Isolate `PlayerUpdateBatch` OnFlushed callback faults from flush failures (closes #2271)

**Verdict:** Approve — the fix correctly isolates deferred callback faults so a best-effort cache advance can no longer be mistaken for a flush failure, and it's well-tested. (Posted as a comment since I'm the PR author; treat it as an approval.)

I traced the full failure path to confirm the fix is real and complete:

- `FlushAsync` (`PlayerUpdateBatch.cs:88`) runs `RunFlushedCallbacks()` after the publish has already succeeded and the buffer is cleared. Previously an exception from a deferred callback propagated out, was caught by `SavePlayer`'s flush `catch` (`PlayerRepository.cs:142`), wrapped as `PlayerPersistenceFlushFailedException`, and short-circuited before the cache-blob write at `PlayerRepository.cs:180`. Isolating each callback (`PlayerUpdateBatch.cs:106-116`) makes `FlushAsync` return normally, so the cache write runs — exactly the same class of divergence #2098/#2106 closed on the sibling paths. Correct fix.
- The isolation mirrors the documented `DomainEventDispatcher` handler-isolation pattern, so it's consistent with the existing best-effort/log-and-continue convention rather than a new one-off.
- Always clearing `_onFlushed` (`:118`) also fixes the secondary "survivors re-run on same-scope retry" bug as a natural consequence — good.

### Blocking
None.

### Should-fix
None.

### Test coverage
Solid and follows the project's classical/state-based style:
- `RunFlushedCallbacks_OneCallbackThrows_LogsAndStillRunsAndClearsTheRest` pins log-and-continue, insertion-order survival, and that the faulted callback is still cleared (no re-run).
- `FlushAsync_OnFlushedCallbackThrows_StillCompletesSuccessfully` pins the end-to-end invariant the real bug hinged on: `FlushAsync` returns normally so `SavePlayer` never wraps it as a flush failure.
- All construction sites updated for the new `ILogger` ctor param; the only non-test construction is DI (`AddScoped<PlayerUpdateBatch>()`), which resolves the logger automatically. No missed sites.

### What's good
Tightly scoped to the actual defect, mirrors an established isolation pattern instead of inventing one, fixes the secondary clearing bug for free, and documents the isolation invariant in `docs/backend-persistence.md`. Nothing to change.


---

PR #2249, opened 2026-07-21
Restrict proficiency prerequisites to path root tiers

ginderjeremiah:
## Summary

Fixes the three-way openness disagreement described in #2236 for a non-root proficiency tier carrying authored cross-path prerequisites:

- `OpenSuccessors` (`Game.Core/Proficiencies/ProficiencyAccrual.cs`) opens a maxed tier's within-path successor **unconditionally** — it never checked the successor's own authored `PrerequisiteIds`.
- The XP-routing gate (`AllPrerequisitesMaxed` in `BuildActivities`) **does** check them, and skips accrual for a still-locked frontier.

So a non-root tier with an authored cross-path prerequisite could be announced as "opened" (battle push + toast) while never actually accruing XP — a real, if narrow, footgun for content authoring, since admin validation never rejected the shape.

## Chosen direction

The issue proposed two directions ("decide one"): restrict authoring to root tiers, or unify the three openness predicates into one. I went with **restricting authoring**, because `docs/game-design.md`'s existing design language already frames a gateway as an exceptional, root-only tier kind ("A gateway tier is the exception: it accrues no XP until every prerequisite is maxed") — it isn't described as a variant of an ordinary within-path tier. Every current design-doc gateway example (e.g. "lava = advanced-fire + advanced-earth") is already a path root, and no seeded content anywhere authors a prerequisite at all, so this is a pure restriction with zero migration cost. Unifying the predicates would have been a larger, riskier change to reconcile three call sites for a shape nothing actually uses.

Also worth noting: while investigating, I found the client's own derived-openness logic (`UI/src/routes/game/screens/proficiencies/proficiencies-lexicon.ts`) **already** requires the conjunctive check (previous tier maxed *and* prerequisites maxed) for non-root tiers — it was only the doc prose (`docs/backend-proficiency.md`) that described a looser OR rule. So no frontend change was needed; I corrected the doc wording instead.

## Changes

- `AdminProficiencies.ValidatePrerequisiteIds`: reject authored prerequisites on a non-root tier (`PathOrdinal != 0`) at save time.
- `ProgressionGraphChecker.CheckProficiencies`: matching Content Health lint check (defense-in-depth parity with the admin save-time rejection).
- `ProficiencyAccrual.OpenSuccessors`: updated comment to state the now-enforced invariant the unconditional within-path open relies on.
- `docs/backend-proficiency.md` / `docs/game-design.md`: corrected the openness wording to state the root-only gateway invariant explicitly.
[...]

PR #2304, opened 2026-07-22
Close the tier-reorder bypass of the root-tier-only prerequisite rule

### LINKED ISSUE #2275: Progression editor: tier reorder bypasses the root-tier-only prerequisite rule (#2249), and the rule has no client-side blocking warning
## Problem (two halves of the same gap)

**1. Backend bypass.** #2249 (commit `7534a6a`) added the "prerequisites only on a path's root tier" rule, but only inside `ValidatePrerequisiteIds`, which fires solely on `SetPrerequisites`. The progression editor's tier reorder (`TiersSpine` ▲/▼ → `progression-store.svelte.ts` `reorderTiers`, `:327-344`) produces **identity-only** `pathOrdinal` edits — `SetPrerequisites` is never called when the prerequisite set is unchanged — and `SaveProficiencies` (`Game.DataAccess/Repositories/Admin/AdminProficiencies.cs:25-64`) validates only path existence, ordinal collisions, and MaxLevel shrink. So: add a new tier, move it above a gated (prerequisite-carrying) root tier, Save → succeeds, and the forbidden shape persists — resurrecting exactly the runtime inconsistency #2249 closed (tier announced as opened by `OpenSuccessors` while `AllPrerequisitesMaxed` withholds all XP). Content Health flags it after the fact (`ProgressionGraphChecker.cs:594-600`), but the save-time enforcement has a bypass through the editor's own UI. `AdminProficienciesIntegrationTests` only covers the `SetPrerequisites` rejection.

**2. No client-side Save gate.** The #2249 hard-reject is invisible to the editor's blocking-warning convention (#2217/#2221): `TierDetail.svelte` disables the Gateways tab only for non-root tiers with *zero* prerequisites, so a non-root tier already carrying prerequisites gets the full add-picker, and a root tier can gain a prerequisite and be reordered off root in the same session. Prerequisites post in step 8 — after identity batches committed — so the known-in-advance rejection lands on the partial-failure rebase path instead of being blocked pre-save, contradicting docs/frontend-admin.md item 7. `proficiencyBlockingWarnings` (`progression-helpers.ts:210`) covers only level range.

## Suggested fix

- Backend: cross-check ordinals against prerequisite-carrying tiers in `SaveProficiencies` (the identity path) so a reorder can't strand prerequisites on a non-root tier; integration test for the reorder path.
- Frontend: a blocking warning for `pathOrdinal !== 0 && prerequisiteIds.length > 0` (surfaces the reorder case pre-save) and hide the add-picker on non-root tiers.

ginderjeremiah:
## Summary

#2249 added the "prerequisites are root-tier-only" rule, but it only lived inside `AdminProficiencies.ValidatePrerequisiteIds`, which fires solely on `SetPrerequisites`. The progression editor's tier reorder (`TiersSpine` ▲/▼) produces an **identity-only** `pathOrdinal` edit through `SaveProficiencies` — `SetPrerequisites` is never called when the prerequisite set itself is unchanged — so an already-gated root tier could be reordered off root, resurrecting the exact runtime inconsistency #2249 closed (a tier announced open by `OpenSuccessors` while `AllPrerequisitesMaxed` withholds all XP).

There was also no client-side Save gate for this: the Gateways tab only disabled itself for a non-root tier with *zero* prerequisites, so a non-root tier that already carries prerequisites (e.g. after a reorder) still offered the full add-picker, and prerequisite changes post late in the save pipeline — after identity batches commit — so a known-in-advance rejection would land on the partial-failure rebase path instead of being blocked pre-save.

- **Backend**: `AdminProficiencies.SaveProficiencies` now runs `FindPrerequisiteRootViolation`, which computes the prospective per-proficiency ordinal (folding in this batch's identity edits) and rejects the save if any tier that still carries persisted prerequisites would end up off ordinal 0 — mirroring `ValidatePrerequisiteIds`'s rule but reachable through the reorder path too.
- **Frontend**: `proficiencyBlockingWarnings` now also flags `pathOrdinal !== 0 && prerequisiteIds.length > 0`, gating Save the same way an out-of-range payout level does. The Gateways tab surfaces this as a warning triangle, and `GatewaysEditor` hides the "Add prerequisite" picker entirely on a non-root tier (previously it only hid when the tier had zero prerequisites), with an inline warning banner explaining why an already-gated non-root tier can't add more.

## Test plan

- [x] `dotnet build` — clean
- [x] `dotnet test` (full solution) — 3434 tests passed, including 2 new `AdminProficienciesIntegrationTests` covering the reorder-strands-a-gateway rejection and a same-ordinal regression check
- [x] `npm run lint` — clean (eslint + svelte-check)
- [x] `npm run test:unit:ci` — 3887 tests passed, including new/updated coverage in `progression-helpers.test.ts`, `GatewaysEditor.test.ts`, `TierDetail.test.ts`, and two `progression-store.test.ts` scenarios updated to use valid root-tier cross-path data (their prior same-path/non-root setup would now — correctly — trip the new Save gate)

Closes #2275

---
_Generated by [Claude Code](https://claude.ai/code/session_011x4Hboc2QsUnPCcGCGxpG3)_

### DIFF
--- a/Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/AdminProficienciesIntegrationTests.cs
@@ -136,6 +136,81 @@ namespace Game.Application.Tests.DataAccess
         }
 
         [Fact]
+        public async Task SaveProficiencies_ReorderStrandsPersistedPrerequisiteOffRoot_ReturnsFailure()
+        {
+            // A root tier already gated by a cross-path prerequisite (via SetPrerequisites) must not be
+            // movable off ordinal 0 through the identity save alone — the progression editor's tier reorder
+            // produces exactly this shape (an ordinal-only Edit that never touches prerequisiteIds), so
+            // ValidatePrerequisiteIds (which only fires from SetPrerequisites) would never see it.
+            int gatedId, prerequisiteId, gatedPathId;
+            using (var seedScope = CreateScope())
+            {
+                var gatedPath = await SeedPathAsync(seedScope);
+                gatedPathId = gatedPath.Id;
+                gatedId = (await SeedProficiencyAsync(seedScope, gatedPath.Id, pathOrdinal: 0, name: "Fire I")).Id;
+                prerequisiteId = (await SeedProficiencyAsync(seedScope, (await SeedPathAsync(seedScope)).Id, pathOrdinal: 0, name: "Frost I")).Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            using (var setupScope = CreateScope())
+            {
+                var setupAdmin = setupScope.ServiceProvider.GetRequiredService<IAdminProficiencies>();
+                Assert.True(setupAdmin.SetPrerequisites([
+                    new SetProficiencyPrerequisitesData { Id = gatedId, PrerequisiteIds = [prerequisiteId] },
+                ]).Succeeded);
+                await setupScope.ServiceProvider.GetRequiredService<IUnitOfWork>().CommitAsync();
+            }
+            await ReloadReferenceCachesAsync();
+
+            using var scope = CreateScope();
+            var admin = scope.ServiceProvider.GetRequiredService<IAdminProficiencies>();
+
+            var result = admin.SaveProficiencies(
+            [
+                new Change<Contracts.Proficiency> { ChangeType = EChangeType.Edit, Item = NewProficiency(id: gatedId, pathId: gatedPathId, pathOrdinal: 1, name: "Fire I") },
+            ]);
+
+            Assert.False(result.Succeeded);
+            Assert.Contains("root tier", result.ErrorMessage);
+        }
+
+        [Fact]
+        public async Task SaveProficiencies_EditsRootTierWithPrerequisitesWithoutMovingOrdinal_IsAccepted()
+        {
+            // The guard must not false-reject an identity edit that leaves an already-gated root tier at
+            // ordinal 0 — only an actual move off root is a violation.
+            int gatedId, prerequisiteId, gatedPathId;
+            using (var seedScope = CreateScope())
+            {
+                var gatedPath = await SeedPathAsync(seedScope);
+                gatedPathId = gatedPath.Id;
+                gatedId = (await SeedProficiencyAsync(seedScope, gatedPath.Id, pathOrdinal: 0, name: "Fire I")).Id;
+                prerequisiteId = (await SeedProficiencyAsync(seedScope, (await SeedPathAsync(seedScope)).Id, pathOrdinal: 0, name: "Frost I")).Id;
+            }
+            await ReloadReferenceCachesAsync();
+
+            using (var setupScope = CreateScope())
+            {
+                var setupAdmin = setupScope.ServiceProvider.GetRequiredService<IAdminProficiencies>();
+                Assert.True(setupAdmin.SetPrerequisites([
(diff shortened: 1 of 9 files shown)


ginderjeremiah (commented):
## Review: Close the tier-reorder bypass of the root-tier-only prerequisite rule (#2275)

**Verdict:** Approve _(posted as a comment — GitHub won't accept a gating verdict from the PR author; this is a genuine approval)_ — both halves of the issue are closed cleanly, the backend guard mirrors its sibling validators exactly, and coverage is thorough on both tiers.

I verified the change against the issue's two-part problem statement and the DDD/testing rules in `docs/backend.md` and the frontend theming/organization rules — no blocking findings, no should-fixes, no nits.

### Correctness

- **Backend (`AdminProficiencies.SaveProficiencies` → `FindPrerequisiteRootViolation`)** correctly closes the identity-only bypass. It folds this batch's `Edit` ordinals into a prospective map and rejects any persisted-prerequisite-carrying tier that would land off ordinal 0. The structure mirrors `FindTierOrdinalCollision`/`FindShrunkenMaxLevelViolation` (same "prospective set from the cache, fail cleanly before staging" idiom), and it only reads `Prerequisites`/`PathOrdinal`/`Name` off the cached entities — no mutation of the `AsNoTracking().Include(...)` graph, per the reference-data gotchas. The "only `Edit` moves an ordinal; an `Add` starts prerequisite-free" reasoning in the doc comment is sound.
- **Frontend split is the right refactor.** Because `proficiencyBlockingWarnings` now also carries `prerequisiteRootWarnings`, `milestoneWarn` was correctly repointed to `levelRangeWarnings` directly (behavior-preserving — the old `proficiencyBlockingWarnings` was just the level-range set) and a separate `gatewaysWarn` was added, so the two conditions light up their own tabs instead of one bleeding into the other. Save gating (`progression-store.svelte.ts:190`) picks up the new condition automatically via `proficiencyBlockingWarnings`.
- **`GatewaysEditor` UX is coherent:** the add-picker is hidden for any non-root tier (not just an empty one), the stranded case surfaces both the chip list (so the offending prerequisites can be removed) and an actionable banner, and the tab stays reachable only when it carries prerequisites (`TierDetail` still disables it for a non-root/zero-prereq tier), so there's no dead editor surface. Theming uses `--warning`/`--text-secondary` tokens, not hardcoded colors.

### Test coverage

Meets the project's rules on both tiers:
- **Backend:** two new `AdminProficienciesIntegrationTests` — the reorder-strands-a-gateway rejection and the identity-edit-that-stays-at-0 acceptance (the false-reject guard). Integration-tested is the right call here since `AdminProficiencies` is a data-tier repo over the entity cache.
- **Frontend:** `progression-helpers.test.ts` pins the warning function plus its inclusion in `proficiencyBlockingWarnings`/`proficiencyWarnings`; `GatewaysEditor.test.ts` covers picker-hidden/shown and the banner; `TierDetail.test.ts` covers the Gateways tab warn icon on/off. The two `progression-store.test.ts` scenarios were correctly migrated to valid root-tier cross-path data — their prior same-path/non-root setup is now (rightly) rejected by the new gate, and the comments make that explicit.

### What's good

Tight, surgical fix that resists the obvious footgun (repointing `milestoneWarn`), keeps the frontend warning mirrored to the backend rejection so the editor blocks pre-save rather than on the partial-failure rebase path, and doesn't over-reach beyond the issue's scope. Comments explain the *why* (the `SetPrerequisites`-only firing of `ValidatePrerequisiteIds`) without narrating the code.

Approving; will merge once CI goes green.

---

PR #2310, opened 2026-07-22
Dead-letter and re-sync-notify envelope-level malformed socket pushes

### LINKED ISSUE #2272: Envelope-level malformed server-push payloads are dropped without dead-letter or re-sync notice, bypassing the documented escalation contract
## Problem

`SocketManagerService.GetSocketCommandProcessor` (`Game.Api/Services/SocketManagerService.cs:~448-473`): `queue.GetNextAsync&lt;SocketCommandInfo&gt;()` LPOPs the item **before** deserializing (`RedisQueue.cs:50-54`), so a payload that fails envelope-level deserialization is popped and only **logged** in the dequeue catch — no dead-letter, no `ServerCommandFailed` re-sync notice.

docs/backend-persistence.md states a malformed server push "escalates the same as a fault since a server push's payload is server-authored and therefore a genuine bug," but the implemented `MalformedParameters` escalation only covers `Parameters`-level malformation inside a *parseable* envelope. If an envelope-level break occurs (most plausibly wire-shape skew during a rolling deploy), the client silently diverges — the exact outcome #671's escalation exists to prevent — and five consecutive poison entries abandon the drain loop for the connection's lifetime.

The drop is commented as intentional for queue advancement (#691), but the escalation contract postdates that comment.

## Suggested fix

On an envelope-level deserialization failure, dead-letter the raw string and emit the `ServerCommandFailed` notice, mirroring the player queue's malformed handling — while keeping the queue-advancement behavior (the pop already happened, so it's just an escalation add, not a control-flow change).

ginderjeremiah:
## Summary

`SocketManagerService.GetSocketCommandProcessor` popped and deserialized a server push in a single call (`queue.GetNextAsync<SocketCommandInfo>()`), so a payload that failed **envelope-level** deserialization was only logged in the dequeue catch — never dead-lettered, and never triggering the client's `ServerCommandFailed` re-sync notice. `docs/backend-persistence.md` documents a malformed server push as escalating the same as a genuine fault (the payload is server-authored, so it's a bug, not client input), but that escalation only covered `MalformedParameters` inside an already-parseable envelope — an envelope-level break (most plausibly wire-shape skew during a rolling deploy) fell through the gap.

## Changes

- `GetSocketCommandProcessor` now pops the raw string first (`queue.GetNextAsync()`) and deserializes it into `SocketCommandInfo` as a separate step, mirroring `DataProviderSynchronizer.ProcessMessage`'s pop-then-deserialize split for the player write-behind queue.
- A `JsonException` from that deserialize — or a payload that parses cleanly to `null` — now calls a new `EscalateMalformedServerCommand`, which dead-letters the **raw string as-is** (no reader-side change needed: `SocketCommandDeadLetters.Classify`/`TryDeserialize` already treats an entry that doesn't parse as a `SocketCommandDeadLetterEnvelope` as `Malformed`) and sends the client a `ServerCommandFailed` notice with a sentinel command name (`"Unknown"`), since the failed push's own name can no longer be recovered.
- Extracted the shared "dispatch the `ServerCommandFailed` notice and log if it didn't land" logic out of `EscalateFailedServerCommand` into `SendServerCommandFailedNotice`, reused by both escalation paths.
- The Redis-blip dequeue-fault catch block (bounded backoff + abandon-after-ceiling) is unchanged and now only covers genuine pop-level faults, since deserialization moved to its own step.

## Test plan

- [x] New unit tests in `Game.Api.Tests/Unit/SocketCommandProcessorTests.cs`: an envelope that fails to deserialize dead-letters the raw payload, notifies the client, and the queue keeps draining into the next valid command; a `"null"` payload gets the same treatment.
- [x] New integration tests in `Game.Api.Tests/Integration/SocketManagerServiceTests.cs`, exercised end-to-end over real Redis + a live WebSocket connection: a raw malformed payload published directly onto a live socket's queue is dead-lettered (verified via `SocketCommandDeadLetters.InspectAsync` classifying it `Malformed`) and the client receives the `ServerCommandFailed` notice.
- [x] Updated the existing `FakeQueue`/`AlwaysThrowingQueue` test doubles to implement the raw (non-generic) `GetNextAsync` the production code now calls.
- [x] `dotnet build` (0 warnings/errors) and full `dotnet test` — 3438 tests passed across all four backend test projects.

Closes #2272

---
_Generated by [Claude Code](https://claude.ai/code/session_01HhtPDe4nzMSoibWysGbyuz)_

### DIFF
--- a/Game.Api.Tests/Integration/ApiIntegrationTestBase.cs
+++ b/Game.Api.Tests/Integration/ApiIntegrationTestBase.cs
@@ -219,6 +219,22 @@ namespace Game.Api.Tests.Integration
         /// <summary>A live socket's own pub/sub queue name.</summary>
         protected static string SocketQueueName(string socketId) => $"{Constants.PUBSUB_SOCKET_QUEUE_PREFIX}_{socketId}";
 
+        /// <summary>A live socket's own pub/sub channel name.</summary>
+        protected static string SocketChannel(string socketId) => $"{Constants.PUBSUB_SOCKET_CHANNEL_PREFIX}_{socketId}";
+
+        /// <summary>
+        /// Publishes a raw payload directly onto a live socket's own queue and wakes its command processor —
+        /// bypassing <c>SocketManagerService.EmitSocketCommand</c>'s <see cref="SocketCommandInfo"/> typing so
+        /// a test can simulate a malformed server push landing on the wire (#2272), which a well-formed
+        /// <see cref="SocketCommandInfo"/> can never reproduce.
+        /// </summary>
+        protected async Task PublishRawToSocketAsync(string socketId, string rawMessage)
+        {
+            using var scope = CreateScope();
+            var pubsub = scope.ServiceProvider.GetRequiredService<IPubSubService>();
+            await pubsub.Publish(SocketChannel(socketId), SocketQueueName(socketId), rawMessage);
+        }
+
         /// <summary>Wraps a server-initiated command as the socket-command dead-letter envelope shape
         /// <see cref="Game.Api.Services.Admin.SocketCommandDeadLetters"/> replays, for tests seeding the
         /// socket dead-letter queue directly.</summary>
--- a/Game.Api.Tests/Integration/SocketManagerServiceTests.cs
+++ b/Game.Api.Tests/Integration/SocketManagerServiceTests.cs
@@ -1,6 +1,10 @@
+using Game.Abstractions.Contracts.Admin;
 using Game.Abstractions.Infrastructure;
+using Game.Api.Models.Common;
 using Game.Api.Services;
+using Game.Api.Services.Admin;
 using Game.Api.Sockets.Commands;
+using Game.Core;
 using Game.TestInfrastructure.Base;
 using Game.TestInfrastructure.Fixtures;
 using Game.TestInfrastructure.Helpers;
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Dead-letter + re-sync-notify envelope-level malformed socket pushes (closes #2272)

**Verdict:** Approve — the change implements exactly the escalation the issue and `docs/backend-persistence.md` call for, mirrors the existing parseable-fault path faithfully, and is well covered on both the unit and integration side. (Submitted as a comment since I'm the PR author; treat it as an approval.)

I verified the core claims against the code rather than the description:

- `Game.Core/Extensions.cs:16` `Deserialize<T>` calls `JsonSerializer.Deserialize` (throws `JsonException` on bad JSON) and returns `default` for a `null` input, so the `catch (JsonException)` and the `nextCommandInfo is null` branches are the correct two escape hatches for an envelope-level break.
- `SocketCommandDeadLetters.TryDeserialize` (`Game.Api/Services/Admin/SocketCommandDeadLetters.cs:138`) returns `null` — hence classifies `Malformed` — both when the raw string fails to parse and when it parses to `null`/a command-less envelope, so dead-lettering the bare raw string genuinely needs no reader-side change, as claimed.
- The `"Unknown"` sentinel is safe on the client: `handleServerCommandFailed` (`UI/src/lib/engine/engine.ts:141`) only special-cases `ChallengeCompleted`/`ProficiencyXpGained` and otherwise just logs, so an unmatched name is a no-op re-sync cue.

### Design notes (non-blocking)

- Extracting `SendServerCommandFailedNotice` out of `EscalateFailedServerCommand` and reusing it from the new path is the right call — no duplicated outcome-capture logic, and the two escalation methods stay symmetric.
- Popping the raw string first and deserializing as a separate step (matching `DataProviderSynchronizer.ProcessMessage`'s split) keeps the dequeue-fault catch strictly for pop-level Redis blips, which is what its bounded-backoff/abandon logic is actually for. Good separation.
- The bare raw string is dead-lettered without a `SocketCommandDeadLetterEnvelope` wrapper (so no player id), which is correct here — a malformed entry is never replay-eligible anyway, and the classifier's `Malformed` verdict keeps it out of replay-all.

### Test coverage

Solid and matched to the project's rules. Unit tests (`SocketCommandProcessorTests`) cover the parse-failure and `"null"`-envelope paths, asserting the raw payload is dead-lettered, the `ServerCommandFailed` notice fires, and the queue keeps draining into the next valid command. Integration tests (`SocketManagerServiceTests`) exercise it end-to-end over real Redis + a live socket, verifying `Malformed` classification via `InspectAsync` and client receipt of the notice. The `FakeQueue`/`AlwaysThrowingQueue` doubles were correctly flipped to implement the non-generic `GetNextAsync` the production code now calls. No frontend mirror is needed — this is backend socket infrastructure, not battle logic, and the client handler is unchanged.

### CI

`verify-backend` and `test-e2e (chromium)` are green; `test-backend` and `test-e2e (webkit)` are still running. `test-e2e (firefox)` failed, but the log shows a Docker Hub registry timeout pulling `redis:7-alpine` (`Client.Timeout exceeded while awaiting headers`) during compose setup — pure infra flakiness, unrelated to this backend-only change. I'll re-kick it and merge once CI is green.


---

PR #2311, opened 2026-07-22
Add a periodic reconciliation wake to the player-update drain

### LINKED ISSUE #2273: Player-update drain has no self-wake after an infra outage — parked events wait for unrelated traffic or a restart
## Problem

Items parked reserved during a DB outage (`Game.DataAccess/DataProviderSynchronizer.cs:653-684`) are only reclaimed inside a drain pass, and drain passes start only on a pub/sub wake (i.e. some player save somewhere in the fleet) or process startup. The stranded-head reclaim (`TryReclaimStrandedProcessingAsync`, `:523-557`) additionally needs **two** passes ≥ 30s apart (first sighting starts the clock; a later pass reclaims).

## Failure scenario

Small/single-instance deployment: a DB outage parks the last battle saves, the affected players disconnect, and no further saves happen fleet-wide. The parked writes sit unapplied until an unrelated save or a deploy. If the player-cache key ages out (48h TTL) first, a returning player's load falls through to the lagging DB before the parked absolute-value events were ever applied.

docs/backend-persistence.md frames "no dedicated re-probe loop" as deliberate, but that framing assumes a next drain pass happens ("a stuck item recovers on the next drain pass"). This is the same at-most-once-shaped gap the reconciliation interval (#1980/#2265) just closed for reference-data invalidations.

## Suggested direction

A small self-scheduled follow-up wake after a pass observes a non-empty processing list (or a low-frequency timer-driven idle pass, mirroring `ReferenceCacheReloadPolicy.ReconciliationInterval`) so parked work always has a bounded time-to-recovery independent of unrelated traffic. Low urgency for a busy fleet; moderate for this project's current small-scale reality.

ginderjeremiah:
## Summary

- `DataProviderSynchronizer` now wakes itself every `reconciliationInterval` (default 5 minutes) even with zero pub/sub traffic, mirroring `ReferenceCacheReloadPolicy.ReconciliationInterval`/`CoalescingReferenceCacheReloader`'s already-established pattern.
- Closes the gap where a drain pass only ever starts on a pub/sub wake or on startup: on a quiet or single-instance deployment, an item parked by a DB-infrastructure outage (or otherwise stranded on the processing list) could sit unrecovered until unrelated traffic happened elsewhere in the fleet or the process restarted, since the opportunistic stranded-processing reclaim needs a *later* drain pass to actually observe and reclaim it.
- The periodic wake reuses the existing `ProcessQueue` gate/coalescing machinery, so it never runs concurrently with — or duplicates the work of — a signal-triggered drain; it just requests a pass exactly like a pub/sub wake does.
- Updated `docs/backend-persistence.md`'s "Queue processing resilience" section, which previously (and now inaccurately) framed "no dedicated re-probe loop" as deliberate.

## How this addresses #2273

The issue's suggested direction offered two options: a targeted follow-up wake only after observing a non-empty processing list, or "a low-frequency timer-driven idle pass, mirroring `ReferenceCacheReloadPolicy.ReconciliationInterval`." I went with the latter — it's simpler, requires no new state to special-case "did we just see something stranded," and mirrors a pattern already proven out (and already tested) in this same file for the reference-data cache reload path.

## Test plan

- [x] `dotnet test` (full backend suite: Core, Infrastructure, Application, Api) — all 3440 tests pass
- [x] Added `StartAsync_ReconciliationInterval_ReclaimsAndAppliesAnItemStrandedWithNoWakeAtAll`, which strands an item on the processing list *after* startup (so the startup reclaim can't be what recovers it) with pub/sub subscribing suppressed entirely, and confirms only the periodic reconciliation loop reclaims and applies it.
- [x] Added `RunReconciliationLoop_StopRequestedWhileWaiting_UnwindsCleanlyWithoutDrainingAgain`, confirming a stop mid-wait unwinds silently without triggering a spurious drain pass or log entry.

Closes #2273

---
_Generated by [Claude Code](https://claude.ai/code/session_01MjaPKVvr6XeyeHjsFfLXoz)_


ginderjeremiah (commented):
## Review: periodic reconciliation wake for the player-update drain (#2273)

**Verdict:** Approve (posted as a comment since I'm the PR author) — solves the issue cleanly by reusing the established pattern; one small in-scope should-fix noted below, so I'm holding the merge until it's resolved.

The change closes the exact gap #2273 describes: a drain pass only ever started on a pub/sub wake or startup, so on a quiet/single-instance deployment an item parked by a DB-infrastructure outage (or otherwise stranded on the processing list) could sit unrecovered until unrelated fleet traffic or a restart. `RunReconciliationLoop` now requests a pass every `_reconciliationInterval` (default 5 min) through the same `ProcessQueue` gate/coalescing path a wake uses, so it never runs concurrently with — or duplicates — a signal-triggered drain, and it correctly gives the opportunistic stranded-processing reclaim the *later* pass it needs to actually reclaim (first pass sights the head and primes the clock; a subsequent pass past the grace period reclaims). Choosing the timer-driven idle pass over a targeted post-observation follow-up wake is the right call — no new "did we just see something stranded" state, and it mirrors a pattern already proven and tested in this same file.

### Should-fix

- `Game.DataAccess/DataProviderSynchronizer.cs:709` — the inline comment still reads "...picks it back up once the database recovers, **with no dedicated re-probe loop needed** (#2097)." This PR corrected the identical framing in `docs/backend-persistence.md` (removed "There is no dedicated re-probe loop") but left this mirror in the code, where it now directly contradicts the `RunReconciliationLoop` the PR adds a few hundred lines up. Please align it — e.g. note that the periodic reconciliation wake is what now bounds recovery on a quiet deployment — so the two don't diverge.

### Test coverage

Solid and matches the project's integration-test conventions for this class (real Postgres/Redis containers):
- `StartAsync_ReconciliationInterval_ReclaimsAndAppliesAnItemStrandedWithNoWakeAtAll` strands the item *after* startup (so the startup reclaim provably can't be what recovers it) with `Subscribe` suppressed so no pub/sub wake can reach the instance — isolating the reconciliation path as the only possible recovery mechanism. Good asserts on the persisted values, the "stranded" Information log, and the absence of any Error.
- `RunReconciliationLoop_StopRequestedWhileWaiting_UnwindsCleanlyWithoutDrainingAgain` covers the clean-unwind edge: a stop mid-wait must not trigger a spurious drain or log.

### What's good

- Reuses the existing gate/coalescing machinery rather than inventing a parallel path — the reconciliation pass is indistinguishable from a wake, so all the existing serialization/shutdown guarantees carry over for free.
- Shutdown handling is correct: the loop only starts after `StartAsync`'s stop guards, the `Task.Delay` is on `_stopping.Token`, and `StopAsync` still waits on the same drain gate, so an in-flight reconciliation pass is awaited to a clean boundary like any other drain.
- Constructor param is optional with a conservative default, so every existing call site and the DI registration are untouched, and it stays tunable for tests/deployments — consistent with the other knobs on this class.
- Doc update is accurate and concise, and the new `DefaultReconciliationInterval` XML doc explains the "comfortably longer than the grace period" sizing rationale rather than just stating the value.


---

ginderjeremiah:
## Re-review (delta since prior review)

Follow-up commit `7b0350b` ("Update stale re-probe-loop comment to reflect the new reconciliation wake") resolves the one should-fix from my earlier review: the `InfrastructureFailureClassifier` catch-block comment at `DataProviderSynchronizer.cs:706-711` no longer claims "no dedicated re-probe loop needed" — it now notes the recovery is "bounded even on a quiet/single-instance deployment by the periodic reconciliation wake (RunReconciliationLoop, #2273)." That aligns the inline comment with both the code the PR adds and the `docs/backend-persistence.md` update, so the two no longer diverge.

Nothing else changed since the approval. CI is green across backend/frontend/e2e.

**Verdict: Approve.** Merging.

---

PR #2312, opened 2026-07-22
Move LoginTrackingMiddleware's DB write off the request latency path

### LINKED ISSUE #2274: LoginTrackingMiddleware pays its DB write inline before the request handler — move it off the latency path
## Problem

`LoginTrackingMiddleware` (`Game.Api/Middleware/LoginTrackingMiddleware.cs:47-81`) awaits `trackingService.RecordConnection(...)` **before** `await _next(context)`. Tracking is documented and structurally implemented as best-effort telemetry that never affects the response — yet it executes serially in front of the request handler. Two costs:

- **Steady state:** every memo miss (first request per (user, IP, device) per 5-minute window) adds the multi-query get-or-create upsert (BrowserInfo + Device + UserLogin, plus the conflict catch-rebuild path) to that request's user-visible latency.
- **During a DB outage/slowdown:** the memo is deliberately set only after a *successful* save, so **every** authenticated fingerprint-carrying HTTP request pays the tracking failure — potentially a full connection timeout — inline before the actual endpoint runs, including on endpoints whose own path never touches Postgres (Status, Players, switch flow, admin tools).

The team has already invested repeatedly in this path's cost profile (5-min memo, sized cache cap #2064, IP bounding #2253); this is the remaining, larger lever.

## Suggested fix

Move the recording after `await _next(context)` (or into `Response.OnCompleted`) — zero behavior change, since nothing about the response depends on it. Keep the dedicated-scope pattern as is.

ginderjeremiah:
## Summary

- `LoginTrackingMiddleware.RecordConnection` now runs **after** `await _next(context)` (in a `finally`, so it still fires if the handler throws) instead of before it.
- Tracking is documented and structurally implemented as best-effort telemetry the response never depends on, yet it previously executed serially in front of the request handler. That cost two things:
  - **Steady state:** every memo miss (first request per (user, IP, device) per 5-minute window) added the multi-query get-or-create upsert to that request's user-visible latency.
  - **During a DB outage/slowdown:** since the memo is only set after a successful save, *every* authenticated fingerprint-carrying request paid the tracking failure — potentially a full connection timeout — inline before the actual endpoint ran, including endpoints whose own path never touches Postgres (Status, Players, switch flow, admin tools).
- Kept the dedicated-scope pattern as-is; only the ordering relative to `_next` changed.

## A wrinkle I found and fixed along the way

The naive "just move it after `_next`" fix breaks a real invariant: `AuthController.DeviceInfo`'s `SaveDeviceInfo` is documented as a no-op unless the caller already has a tracked device for that fingerprint — it enriches an existing device row rather than creating one. Previously this worked because tracking ran *before* the handler, so even if `DeviceInfo` were the very first authenticated+fingerprinted request, the device row would already exist by the time the handler ran.

I verified this isn't a real production risk: the frontend (`UI/src/lib/api/api-request.ts`) attaches the device fingerprint header to **every** authenticated request, and `reportDeviceInfo()` (`UI/src/routes/select/+page.svelte:70`, `UI/src/lib/engine/session.ts:64`) is always called only after a separate, prior authenticated request (`SelectPlayer` or `Auth/Status`) has already completed — so in practice a device is always tracked before `DeviceInfo` ever runs.

The one integration test that broke (`DeviceInfoEndpoint_EnrichesStoredDevice`) was an artifact of the test helper: `SelectPlayerAsync` uses the base test `HttpClient` without a fingerprint header, so the test's `DeviceInfo` POST was — unrealistically — the very first fingerprinted authenticated request. Fixed by priming tracking with a `Status` call first, matching real frontend sequencing.

## Test plan

- [x] `dotnet test` (full backend suite: Core, Infrastructure, Application, Api) — all 3440 tests pass
- [x] Added `Authenticated_NewDeviceIpUser_RunsTheHandlerBeforeRecordingTheConnection`, asserting the handler runs before the tracking DB write.
- [x] Added `NextThrows_StillRecordsConnection_AndRethrowsTheOriginalException`, confirming the `finally` still records the connection when the handler throws, and the original exception propagates unchanged.
- [x] Updated `DeviceInfoEndpoint_EnrichesStoredDevice` to prime tracking via a preceding authenticated request, matching real frontend sequencing.

Closes #2274

---
_Generated by [Claude Code](https://claude.ai/code/session_015EHsHLgWSXMue7tgAbAMEq)_

### DIFF
--- a/Game.Api.Tests/Integration/LoginTrackingTests.cs
+++ b/Game.Api.Tests/Integration/LoginTrackingTests.cs
@@ -92,9 +92,13 @@ namespace Game.Api.Tests.Integration
         [Fact]
         public async Task DeviceInfoEndpoint_EnrichesStoredDevice()
         {
-            // Arrange
+            // Arrange — the frontend attaches the fingerprint header to every authenticated request, so in
+            // practice DeviceInfo is always preceded by at least one other authenticated request (e.g.
+            // SelectPlayer) that has already tracked the device. Prime that here with a Status call, since
+            // SaveDeviceInfo enriches an existing tracked device rather than creating one.
             await SeedUserAsync("enrichuser", "enrichpass");
             using var authClient = await LoginWithDeviceAsync("enrichuser", "enrichpass");
+            await authClient.GetAsync("/api/Auth/Status", CancellationToken);
 
             var body = new { DeviceMemory = 16.0, HardwareConcurrency = 8 };
 
--- a/Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
+++ b/Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging.Abstractions;
+using System.Collections.Generic;
 using System.Net;
 using Xunit;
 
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Move `LoginTrackingMiddleware`'s DB write off the request latency path (#2274)

**Verdict:** Approve — modulo one should-fix (a stale comment this PR invalidates) that should be corrected before merge. Posting as a comment since I'm the PR author; treat the intent as "approve once the should-fix lands."

The change does exactly what #2274 asked: the get-or-create upsert no longer sits in front of the endpoint on the latency path, and during a DB slowdown an authenticated request no longer eats a tracking timeout before an endpoint that may never touch Postgres gets to run. The refactor into `ResolvePendingTracking` / `RecordConnectionAsync` is clean, avoids the null-forgiving operator, and keeps the dedicated-scope isolation intact.

I specifically chased the DeviceInfo "wrinkle" you flagged, and it holds up:
- Both `reportDeviceInfo()` call sites are always preceded by an authenticated, fingerprint-carrying request — `Players/SelectPlayer` in the enter-world path (`select/+page.svelte`), and `loadPlayer` (`Auth/Status`) in the restore path (`session.ts`).
- Because Kestrel buffers these small JSON responses and flushes only after the middleware `finally` completes, the prior request's `RecordConnection` is durable before the client can fire `DeviceInfo`. So `SaveDeviceInfo` always finds a tracked device in a healthy system. The one behavior change — DeviceInfo enrichment now depends on a *prior* successful tracking rather than being self-sufficient — only degrades during a DB outage, where the enrichment is best-effort telemetry that would be failing anyway. Acceptable, and consistent with the issue's intent.

### Should-fix

- `Game.DataAccess/Repositories/UserLogins.cs:131-132` — the comment "RecordConnection (which runs earlier in the same request, via LoginTrackingMiddleware) is solely responsible for establishing that link" is now inaccurate. After this PR, RecordConnection runs *after* the handler (later in the same request), and the device link `SaveDeviceInfo` relies on is established by a **prior** authenticated request, not the same one. This documents a security-relevant invariant (#2064), so a future reader shouldn't be misled about the sequencing. It's in an untouched file (hence no inline comment), but it's this PR that made it stale — please fix it here rather than as a follow-up. Suggested: "RecordConnection (on a prior authenticated request, via LoginTrackingMiddleware) is solely responsible for establishing that link."

### Test coverage

- Solid. `Authenticated_NewDeviceIpUser_RunsTheHandlerBeforeRecordingTheConnection` pins the new ordering; `NextThrows_StillRecordsConnection_AndRethrowsTheOriginalException` covers the `finally` + exception-propagation contract. The `DeviceInfoEndpoint_EnrichesStoredDevice` update correctly models the new "prime via a prior authenticated request" dependency (and incidentally exercises the response-flush timing that makes the whole thing safe).

### What's good

- The extra diligence on the DeviceInfo invariant — finding it, verifying it against the real frontend sequencing, and fixing the test artifact rather than papering over it — is exactly right. Clean refactor, accurate new comments in the middleware itself, no dead code left behind.


---

ginderjeremiah:
## Re-review (delta since prior review)

Follow-up commit `e5c6b8e` ("Fix stale RecordConnection-timing comment in `UserLogins.SaveDeviceInfo`") resolves the one should-fix from my earlier review: `UserLogins.cs:130-131` now reads "RecordConnection (on a prior authenticated request, via LoginTrackingMiddleware) is solely responsible for establishing that link," which accurately reflects the new ordering (tracking runs *after* the handler, so the device link comes from a prior request). The security-relevant `#2064` invariant is documented correctly again.

Nothing else changed since the approval. CI is green across backend/frontend/e2e.

**Verdict: Approve.** Merging.

---

PR #2313, opened 2026-07-22
Workbench: exclude Home zone from spawn pickers, warn on boss/spawn conflicts

### LINKED ISSUE #2276: Workbench: Home-zone authoring hard-rejects have no client-side surface — spawn pickers offer the Home zone as a valid target
## Problem

Assigning a boss to the Home zone, adding spawn rows to it in the zone editor, or picking the Home zone in the enemy editor's Spawns tab are all categorical backend rejects (`AdminZones.cs:48-50, 167-169`; `AdminEnemies.cs:164-167` — "no enemies spawn there and never will"), yet the Workbench neither excludes Home from the pickers nor warns:

- `zoneOptions` (`UI/src/routes/admin/workbench/reference.svelte.ts:159-160`) filters only retirement, so Home zones are offered in the enemy spawn-zone picker (`entities/enemy.ts:139`).
- `entities/zone.ts` identity `warn` covers only boss-flag drift (`:86-94`); the zoneEnemies section `warn` returns null for a Home zone regardless of spawn rows (`:158`).

The `SetZoneEnemies`/`SetEnemySpawns` rejections are child-saver calls, so in any multi-part save they fire **after** the identity batch commits → partial-failure rebase for a mistake that is per-record client-computable (`isHome && zoneEnemies.length > 0`; `spawns.some(s => zones[s.zoneId]?.isHome)`). This contradicts the blocking-warning convention the repo just invested in (#2217, #2221, #2250, #2214).

## Suggested fix

Exclude Home zones from spawn-target pickers (the way retired zones are excluded) and add blocking warnings for the residual cases (boss-on-Home, spawn rows on a zone being flagged Home this session).

ginderjeremiah:
## Summary

Home-zone authoring hard-rejects (`AdminZones.SaveZones:48-50,167-169`; `AdminEnemies.SetSpawns:164-167`) had no client-side surface: the Workbench let an admin pick the Home zone as a spawn target, assign it a boss, or leave spawn rows on it, only to have the save fail server-side.

- `reference.zoneOptions` now excludes the Home zone from spawn-target pickers (the enemy editor's Spawns tab) the same way a retired zone is excluded — the current value stays visible if it's already Home, mirroring the existing retired-record `keep` convention.
- The zone editor's identity `warn` now flags a boss assigned to a Home zone as a `blocking` warning, alongside the existing boss-flag-drift check.
- The zone editor's Enemies-tab `warn` now flags a Home zone carrying spawn rows as a `blocking` warning instead of silently passing (previously `isHome` short-circuited the check entirely).

All three follow the repo's established blocking-warning convention (#2217/#2221/#2250/#2214): `SaveBar` disables Save while any pending record carries a `blocking: true` warning, so these predictable backend rejections never reach the partial-failure rebase path.

## Test plan

- [x] Added unit tests in `reference.test.ts` covering the Home-zone exclusion from `zoneOptions` (and the `keep`-visible exception).
- [x] Added unit tests in `zone.test.ts` covering both new blocking-warning cases (boss-on-Home, spawn-rows-on-Home) and their pass-through counterparts.
- [x] `npm run test:unit:ci` — 310 files / 3893 tests passing.

Closes #2276

---
_Generated by [Claude Code](https://claude.ai/code/session_01GJtD7GbMbXxyTVHoURk4aM)_

### DIFF
--- a/UI/src/routes/admin/workbench/entities/zone.ts
+++ b/UI/src/routes/admin/workbench/entities/zone.ts
@@ -82,10 +82,14 @@ export const zoneEntity: EntityConfig<WorkbenchZone> = {
 			// The dedicated-boss picker keeps an authored value visible even if it loses its
 			// boss flag (reference.bossEnemyOptions' `keep` exception); flag that drift here
 			// since the zone would otherwise silently point at a non-boss enemy. Hard-rejected
-			// by AdminZones.SaveZones, so it blocks Save (#2217).
+			// by AdminZones.SaveZones, so it blocks Save (#2217). A boss on a Home zone is
+			// likewise hard-rejected there (Home is a no-combat sanctuary) — flag it too.
 			warn: (z) => {
 				if (z.bossEnemyId == null || z.bossEnemyId < 0) {
 					return null;
+				}
+				if (z.isHome) {
+					return { message: 'The Home zone cannot have a boss', blocking: true };
 				}
 				const boss = staticData.enemies?.[z.bossEnemyId];
 				return boss && !boss.isBoss
@@ -154,8 +158,15 @@ export const zoneEntity: EntityConfig<WorkbenchZone> = {
 			desc: 'Enemies that spawn here & their weights',
 			count: (z) => z.zoneEnemies.length,
 			// A retired enemy's row doesn't count toward "does something spawn here" — it never
-			// rolls at runtime (mirrors the backend's EmptyCombatZone lint).
-			warn: (z) => (z.isHome || z.zoneEnemies.some((ze) => isLiveEnemy(ze.enemyId)) ? null : 'No enemies spawn here'),
+			// rolls at runtime (mirrors the backend's EmptyCombatZone lint). A Home zone with any
+			// spawn rows is a harder problem: it's hard-rejected on save (AdminZones.SaveZones'
+			// edit guard, AdminZones.SetEnemies), so it blocks rather than merely warns.
+			warn: (z) => {
+				if (z.isHome) {
+					return z.zoneEnemies.length ? { message: 'The Home zone cannot have enemy spawns', blocking: true } : null;
+				}
+				return z.zoneEnemies.some((ze) => isLiveEnemy(ze.enemyId)) ? null : 'No enemies spawn here';
+			},
 			kind: 'table',
 			itemsKey: 'zoneEnemies',
 			rowKey: 'enemyId',
--- a/UI/src/routes/admin/workbench/reference.svelte.ts
+++ b/UI/src/routes/admin/workbench/reference.svelte.ts
@@ -156,8 +156,17 @@ class WorkbenchReference {
 	lessonTriggerTypeOptions = (): SelectOption[] => toOptions(enumPairs(ELessonTriggerType));
 	mechanicEventOptions = (): SelectOption[] => toOptions(enumPairs(EMechanicEvent));
 	tagCategoryOptions = (): SelectOption[] => this.tagCategories.map((c) => ({ value: c.id, text: c.name }));
+	/**
+	 * Zone spawn-target picker options: excludes the Home zone the same way a retired zone is
+	 * excluded — it's a no-combat sanctuary the backend categorically rejects a spawn against
+	 * (`AdminEnemies.SetSpawns`) — while keeping the current value visible if it's already Home.
+	 */
 	zoneOptions = (keep?: number): SelectOption[] =>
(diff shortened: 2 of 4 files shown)


ginderjeremiah (commented):
## Review: Surface the Home-zone authoring hard-rejects in the Workbench (#2276)

**Verdict:** Approve — clean, correctly scoped, and matches the backend guards exactly. (Posting as `COMMENT` since I'm the PR author; treat the intent as approve.)

The PR closes the three-way gap #2276 described, and I verified each client surface against the backend rule it mirrors:

- **Boss on Home** → identity `warn` blocks it; mirrors `AdminZones.SaveZones:48-50`.
- **Spawn rows on Home** → Enemies-tab `warn` blocks whenever `isHome && zoneEnemies.length`; mirrors `AdminZones.SetEnemies:167` and the edit-flip guard at `AdminZones.SaveZones:53-61`. Correctly still passes an empty Home zone without the "No enemies spawn here" nag.
- **Home as a spawn target** → `zoneOptions` excludes Home from the enemy Spawns picker; mirrors `AdminEnemies.SetSpawns:164-167`.

All three land these as `blocking` warnings, so `SaveBar` disables Save before the request ever reaches the partial-failure rebase path — exactly the convention #2217/#2221/#2250/#2214 established.

### What's good

- The `keep` exception is handled consistently with the retired-record convention: a Home zone stays visible only when it's already the current value (`!z.isHome || z.id === keep`), so a pre-existing inconsistent spawn can still be seen and removed rather than silently blanked — while a fresh authoring list never offers it.
- The Enemies-tab `warn` refactor preserves the retired-enemy `isLiveEnemy` semantics for combat zones while cleanly special-casing Home, instead of the old `isHome` short-circuit that swallowed the check.
- Comments are concise and point at the exact backend guard each surface mirrors.

### Test coverage

Solid and sufficient. `reference.test.ts` covers both the Home exclusion and the `keep`-visible exception; `zone.test.ts` covers both new blocking cases and their pass-throughs (empty Home zone, no-boss Home zone). `beforeEach` fully resets `staticData`, so the tests mutating `isHome` are isolated. No battle logic is touched, so no dual-side mirroring is required.

---

ginderjeremiah:
## Review (post-merge)

This PR was already merged by the time the review ran, so recording it here for the record — verdict would have been **Approve** with one small, non-blocking follow-up (now filed).

### Correctness — verified against the server guards

- **`zoneOptions` Home exclusion** (`reference.svelte.ts:159-167`): `(z) => !z.isHome || z.id === keep` correctly mirrors the retired-record `keep` convention and the sibling `bossEnemyOptions`/`grantedSkillOptions` pattern. With `keep` undefined, `z.id === undefined` is false, so only non-Home zones survive; the already-Home cell value stays visible because `TableCell` passes the current value as `keep`. Only the enemy Spawns tab consumes `zoneOptions`, so no other picker regresses. Mirrors `AdminEnemies.SetSpawns:164-167`.
- **Boss-on-Home warning** (`entities/zone.ts:86-94`): the `bossEnemyId == null || < 0` sentinel short-circuit runs first, and the `isHome` check precedes the boss-flag lookup, so any assigned boss on Home blocks while a boss-less Home passes. Consistent with the persist DTO's `-1 → undefined` mapping and `AdminZones.SaveZones:48-50`.
- **Spawn-rows-on-Home warning** (`entities/zone.ts:158-163`): uses raw `z.zoneEnemies.length` (not `isLiveEnemy`), correctly mirroring `AdminZones` `ZoneEnemies.Count > 0` — even a retired spawn row is server-rejected, and the client now blocks identically. The non-Home branch keeps the original "No enemies spawn here" nag.
- Save ordering confirmed: the primary `SaveZones` commits before the child `SetZoneEnemies` saver, so preempting these rejections client-side genuinely avoids the partial-failure rebase path.

### Test coverage — adequate

`reference.test.ts` covers the exclusion + `keep`-visible exception; `zone.test.ts` covers both new blocking cases and their pass-through counterparts, and the boss test isolates `isHome` precedence over the flag-drift lookup. `beforeEach` clears `staticData`, so the sparse-array fixtures don't leak.

### Code quality

Clean — no null-forgiving `!` (uses `?.`/`?? []`), all new control blocks braced, comments cite the backend guard + issue numbers, DRY (reuses `retireableOptions`), consistent with the #2217/#2221/#2250/#2214 blocking-warning convention.

### Follow-up (filed, not a blocker)

One residual edge case: clearing a zone's spawn rows **and** flipping `isHome` in the *same* save slips past the client warning (record shows `zoneEnemies.length === 0`) but is rejected by `SaveZones` reading the still-cached spawn count. It's a clean full rejection, not the partial-failure path this convention primarily guards, and matches the backend's deliberate two-save workflow — so low severity. Tracked in #2314.

---

PR #2316, opened 2026-07-22
Remove BattleStats.PlayerRating, a production-write-only field with misleading comments

### LINKED ISSUE #2278: BattleStats.PlayerRating is production-write-only, with misleading doc comments at both write sites
## Problem

`BattleStats.PlayerRating` (`Game.Core/Battle/BattleStats.cs:156-162`) is written at `BattleService.cs:639` and `OfflineProgressSimulator.cs:138-141`, and its doc comment — plus the comments at both assignment sites — says it exists "so the proficiency accrual normalizes activity by the identical measure." But the accrual never reads it: the live path normalizes off `BattleCompletedEvent.PlayerRating/EnemyRating` (`BattleStatisticsEventHandler.cs:43`) and the offline loop passes `Math.Max(rewards.PlayerRating, rewards.EnemyRating)` directly (`OfflineProgressSimulator.cs:150`).

A whole-repo grep shows the only readers are tests (`BattleServiceIntegrationTests`, `OfflineProgressSimulatorTests`); it is not logged (`BattleCompletedEvent.GetLogProperties` excludes stats) and not mapped to any API model.

Same class of test-only public surface as #1707, but not in that issue's list — and unlike #1707's members it isn't gated on the calibration-comparison decision, so it can be cleaned up independently.

## Suggested fix

Remove the field and re-point the tests at the event/return values (preferred), or fix the comments if there's a near-term reader planned.

ginderjeremiah:
## Summary

Closes #2278.

`BattleStats.PlayerRating` was written at two sites (`BattleService.RecordVictory` and `OfflineProgressSimulator`), and its doc comment — plus the comments at both write sites — claimed it existed "so the proficiency accrual normalizes activity by the identical measure." That claim doesn't hold up:

- The live path normalizes off `BattleCompletedEvent.PlayerRating`/`EnemyRating` (`BattleStatisticsEventHandler.cs:43`), not `BattleStats.PlayerRating`.
- The offline loop passes `Math.Max(rewards.PlayerRating, rewards.EnemyRating)` directly into `ProficiencyAccrual.Accrue` (`OfflineProgressSimulator.cs`), never reading the stats field either.
- `ProficiencyAccrual` itself has no reference to `BattleStats.PlayerRating` at all.

A repo-wide grep confirmed the only readers were tests — it isn't logged (`BattleCompletedEvent.GetLogProperties` excludes stats) and isn't mapped to any API model.

## Changes

- Removed `BattleStats.PlayerRating` and its doc comment.
- Removed both write sites and their now-inaccurate comments (`BattleService.RecordVictory`, `OfflineProgressSimulator`).
- Updated the `RecordVictory` "internal (not private)" comment, which specifically cited the removed field as the reason for that visibility — it's still `internal` because integration tests call it directly to bypass `TryResolveActiveBattle`'s `SimulateBattle` call, which is unrelated to the removed field.
- Re-pointed the three tests that read `BattleStats.PlayerRating` at the values that were already available and equivalent:
  - `BattleServiceIntegrationTests`: dropped the redundant stats assertion, keeping the `DefeatRewards.PlayerRating` assertion (renamed the test since it's no longer about a stats snapshot).
  - `OfflineProgressSimulatorTests` (×2): re-pointed at `OfflineBattleOutcome.PlayerRating`, which already carried the identical value.

## Test plan

- [x] `dotnet test --no-build --no-restore` — full backend suite green (Core, Infrastructure, Application, Api: 1431 + 68 + 1085 + 858 passing).
- No frontend changes (this field has no frontend/API surface).

---
_Generated by [Claude Code](https://claude.ai/code/session_01WtMQ6iujM1yQK6VMY1dm5o)_

### DIFF
--- a/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
@@ -313,7 +313,7 @@ namespace Game.Application.Tests.Services
         }
 
         [Fact]
-        public async Task RecordVictory_SnapshotsPlayerRatingOntoStats_FromSnapshotNotLiveAggregate()
+        public async Task RecordVictory_RatesPlayerRewardFromSnapshotNotLiveAggregate()
         {
             using var scope = CreateScope();
             var context = scope.ServiceProvider.GetRequiredService<GameContext>();
@@ -339,8 +339,8 @@ namespace Game.Application.Tests.Services
             await battleService.StartBattle(player, state, zoneId: zone.Id);
 
             // Deflate the LIVE player's power to almost nothing after the snapshot is frozen (a valid mid-battle
-            // stat reallocation). The accrual must normalize by the snapshot-era rating, so a regression sourcing
-            // PlayerRating from the live aggregate — or dropping the assignment — would read a far lower value.
+            // stat reallocation). The reward must rate the snapshot-era attributes, so a regression sourcing
+            // PlayerRating from the live aggregate would read a far lower value.
             foreach (var allocation in player.StatPoints.StatAllocations)
             {
                 allocation.Amount = 1;
@@ -361,7 +361,6 @@ namespace Game.Application.Tests.Services
                 new Battler(new AttributeCollection(player.GetAllModifiers()), [], player.Level), isPlayer: true);
             Assert.True(rewards.PlayerRating > deflatedRating,
                 $"Expected the snapshot-era rating ({rewards.PlayerRating}) to exceed the deflated live rating ({deflatedRating}).");
-            Assert.Equal(rewards.PlayerRating, result.Stats.PlayerRating, precision: 9);
         }
 
         [Fact]
--- a/Game.Application/Services/BattleService.cs
+++ b/Game.Application/Services/BattleService.cs
@@ -598,10 +598,9 @@ namespace Game.Application.Services
         // (AbandonBattle's elapsedMs) — a win only resolves if the enemy died within time the server itself
         // observed, so the server-measured cap is the (stronger) control there and nothing else is needed.
         // Both paths therefore require a server-validated timeline; neither can be claimed early.
-        // internal (not private) so an integration test can assert the live PlayerRating snapshot directly:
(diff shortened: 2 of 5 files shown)


ginderjeremiah (commented):
## Review: Remove `BattleStats.PlayerRating`, a production-write-only field (closes #2278)

**Verdict:** Approve — clean, correctly-scoped tech-debt removal; every claim in the description checks out. (Submitted as a comment since I'm the PR author; treat it as an approval.)

I verified the core claim against the tree at head (`13a5e9f`):

- **No production reader remains.** A grep for `Stats.PlayerRating` returns zero matches after the change — the only consumers were the three tests re-pointed here. The live path still normalizes off `BattleCompletedEvent.PlayerRating/EnemyRating` and the offline loop off `Math.Max(rewards.PlayerRating, rewards.EnemyRating)`, neither of which touched the removed field.
- **Test re-points are equivalent, not lossy.** `OfflineBattleOutcome.PlayerRating` and `DefeatRewards.PlayerRating` already carried the identical value the field was snapshotted from, so the assertions test the same thing without the redundant hop through stats.
- **The `internal` comment update is accurate.** `RecordVictory` is genuinely still called directly by integration tests (`BattleServiceIntegrationTests.cs:356,440,494,1918`), bypassing `TryResolveActiveBattle`'s `SimulateBattle` — so `internal` visibility remains justified for a reason unrelated to the removed field, exactly as the revised comment states.
- **No stale references left behind.** No remaining code comments or docs reference the removed field.

### What's good

- The description does the hard part of the reviewer's job for them: it names both former write sites, the actual normalization inputs, and how it confirmed the field was test-only. That all held up under verification.
- Rather than deleting the two tests, they were re-anchored to the values that already carried the same measure — preserving the regression coverage (snapshot-era vs. live-aggregate rating) while dropping the dead surface. The test renames (`...OntoStats` → `...RatesPlayerReward` / `...OntoEachVictoryOutcome`) keep the names honest about what they now assert.

### Nits

None.

CI is still in progress at review time — will merge once it goes green.

---

PR #2317, opened 2026-07-22
Skip the write-behind save on a no-op SetAutoChallengeBoss toggle

### LINKED ISSUE #2280: SetAutoChallengeBoss saves on a no-op toggle — missing the changed-ness guard SaveLogPreferences got in #2254
## Problem

`Player.SetAutoChallengeBoss` (`Game.Core/Players/Player.cs:77-81`) unconditionally raises `PlayerCoreUpdatedEvent` and `BattleService.SetAutoChallengeBoss` (`Game.Application/Services/BattleService.cs:262-282`) unconditionally saves, even when the value is unchanged — the exact "no mutation → no save" gap just closed for `SaveLogPreferences` (#2254, where `Player.UpdateLogPreference` returns changed-ness) and followed by every other player mutation command.

The frontend dedups its syncs (`enemy-manager.ts` `#lastSyncedAutoChallengeBoss`), so real-world frequency is low, but a redundant or tampered command still costs a full write-behind flush.

## Suggested fix

Mirror #2254: have `SetAutoChallengeBoss` return whether the value changed and skip the save (and event) on a no-op, with matching unit coverage.

ginderjeremiah:
## Summary

Closes #2280.

`Player.SetAutoChallengeBoss` unconditionally raised `PlayerCoreUpdatedEvent` and `BattleService.SetAutoChallengeBoss` unconditionally saved, even when the toggle matched the mode already in effect — the same "no mutation → no save" gap just closed for `SaveLogPreferences` in #2254.

## Changes

- `Player.SetAutoChallengeBoss` (`Game.Core/Players/Player.cs`) now returns `bool` and is a no-op (no event raised) when `enabled` already matches `AutoChallengeBoss` — mirroring the guard `Player.UpdateLogPreference` got in #2254.
- `BattleService.SetAutoChallengeBoss` (`Game.Application/Services/BattleService.cs`) now only calls `SavePlayer` when the toggle actually changed the mode, on both the enable and disable paths. It still returns `true` for an already-active no-op — accepting the command is not the same as changing state, and the anti-cheat/success contract for callers (the `SetAutoChallengeBoss` socket command) is unchanged.

## Test plan

- [x] `Game.Core.Tests/Players/PlayerTests.cs`: added `SetAutoChallengeBoss_AlreadyEnabled_ReturnsFalseAndDoesNotRaiseEvent` and `SetAutoChallengeBoss_AlreadyDisabled_ReturnsFalseAndDoesNotRaiseEvent`; updated the existing enable/disable tests to assert the new `true` return value.
- [x] `Game.Application.Tests/Services/BattleServiceIntegrationTests.cs`: added `SetAutoChallengeBoss_AlreadyActiveMode_SkipsSaveButStillReturnsTrue` (proves a real toggle enqueues a write-behind event first, then that re-enabling is a no-op — same shape as the #2254 `SaveLogPreferences_NoValuesChanged_DoesNotEnqueueWriteBehindEvents` test) and `SetAutoChallengeBoss_AlreadyIdle_DisableIsNoOpButStillReturnsTrue`.
- [x] `dotnet test --no-build --no-restore` — full backend suite green (Core: 1433, Infrastructure: 68, Application: 1087, Api: 858 passing).
- No frontend changes (the frontend already dedups this sync via `enemy-manager.ts`'s `#lastSyncedAutoChallengeBoss`, so this change hardens the backend invariant itself — same framing as #2254).

---
_Generated by [Claude Code](https://claude.ai/code/session_016DSrVXT3jtU9kkeRNbRkgd)_

### DIFF
--- a/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
+++ b/Game.Application.Tests/Services/BattleServiceIntegrationTests.cs
@@ -8,11 +8,13 @@ using Game.Core.Battle.Events;
 using Game.Core.Battle.Offline;
 using Game.Core.Players;
 using Game.Core.TestInfrastructure.Builders;
+using Game.DataAccess;
 using Game.Infrastructure.Database;
 using Game.TestInfrastructure.Base;
 using Game.TestInfrastructure.Fixtures;
 using Game.TestInfrastructure.Helpers;
 using Microsoft.Extensions.DependencyInjection;
+using StackExchange.Redis;
 using Xunit;
 
 namespace Game.Application.Tests.Services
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: Skip the write-behind save on a no-op `SetAutoChallengeBoss` toggle (closes #2280)

**Verdict:** Approve (submitted as `COMMENT` — I authored this PR, so GitHub won't accept a gating verdict from me) — the change implements the issue exactly, with correct layering and solid coverage on both sides.

The PR mirrors the #2254 changed-ness guard precisely, as #2280 asked:
- The no-op guard (`AutoChallengeBoss == enabled → return false`, no event) lives in `Player` (`Game.Core`) — domain logic in the domain layer, per `backend.md` DDD guidance.
- `BattleService` (application layer) only orchestrates: it saves iff the mode actually changed, on both the enable and disable paths.
- Returning `true` for an already-active no-op is the right call — accepting the command is distinct from mutating state, and the anti-cheat/success contract for the socket command is unchanged (`false` is still reserved for the zone-validation rejection on enable). The updated XML docs make this explicit.

### Test coverage

- `PlayerTests.cs`: new `AlreadyEnabled` / `AlreadyDisabled` cases assert `false` + no event, and the existing enable/disable tests now assert the `true` return — edge cases covered classical-style.
- `BattleServiceIntegrationTests.cs`: the `AlreadyActiveMode` test first proves a real toggle enqueues a write-behind event *before* asserting the re-toggle is a no-op — a nice guard against a trivially-passing assertion (e.g. a broken queue). The `AlreadyIdle` disable path is covered too.

### What's good

- Faithful reuse of the established #2254 pattern rather than a bespoke solution.
- The integration test's "prove the queue works, then prove the no-op" structure is exactly right for this kind of negative assertion.

No blocking, should-fix, or nit findings. Approving; will merge once CI is green.

---

PR #2324, opened 2026-07-22
Preempt Home-flip save rejection when spawns are cleared in the same edit

### LINKED ISSUE #2314: Workbench: clear-spawns-and-flip-to-Home in one save slips past the client blocking-warning
## Context

Follow-up from PR #2313 (which added client-side blocking warnings mirroring the server's Home-zone hard-rejects).

## Gap

The zone editor's Enemies-tab blocking warning (`UI/src/routes/admin/workbench/entities/zone.ts:158-163`) uses the record's current `z.zoneEnemies.length` to decide whether a Home zone still carries spawn rows. This correctly preempts the case where rows are visibly present.

However, if an admin **clears** a combat zone's spawn rows **and** toggles `isHome` on in the *same* save, the client sees `zoneEnemies.length === 0` and allows the save — but `AdminZones.SaveZones` reads the still-**cached** spawn count (`_zones.LookupZone(...).ZoneEnemies.Count`, evaluated before the child `SetZoneEnemies` saver clears it) and rejects the whole batch. The backend deliberately mandates a two-save workflow here (its error message says to clear the zone's spawn table before making it Home).

## Severity / notes

Low. Unlike the cases #2313 targets, this produces a **clean full rejection** (the primary `SaveZones` fails, `committed` stays false, edits are preserved and surfaced as a raw error) — not the partial-failure rebase path the blocking-warning convention primarily guards against. The client `warn` can't currently mirror it because it only receives the pending record, not the persisted baseline spawn count.

## Possible directions

- Give the zone `warn` access to the persisted baseline (cached `ZoneEnemies.Count`) so it can flag "clear spawns in a separate save before making this zone Home."
- Or simply add a code comment at `zone.ts:158-163` documenting the known gap and why it's acceptable (clean rejection, deliberate two-save backend workflow).

Whichever, it's a small, self-contained change.

ginderjeremiah:
## Summary

Follow-up from #2313. The zone editor's Enemies-tab blocking warning (`zone.ts`'s `zoneEnemies` section `warn`) only checked the *pending* record's `zoneEnemies.length`, so it correctly caught a Home zone that visibly still has spawn rows. But it missed a real gap: if an admin **clears** a combat zone's spawn rows and toggles `isHome` on in the *same* save, the client sees `zoneEnemies.length === 0` and lets the save through — while the backend still rejects it.

`AdminZones.SaveZones`'s edit guard reads the zone's **persisted** spawn count (`_zones.LookupZone(...).ZoneEnemies.Count`), evaluated as part of the primary identity batch, *before* the child `SetZoneEnemies` saver has a chance to clear it. So a same-save clear-and-flip still hard-rejects with "clear the zone's spawn table before making it Home" — a clean, non-destructive rejection (nothing commits, all edits stay in place), but a warning the admin never sees coming.

## Fix

- Threaded the record's last-saved **baseline** into `SectionConfig.warn` (previously only received the pending record). `baseline` is optional so every other entity's `warn` predicate — which don't need it — is unaffected.
- `zone.ts`'s `zoneEnemies` `warn` now also checks the baseline's spawn count when the pending record is a soon-to-be Home zone with an already-cleared spawn table, surfacing "Clear spawns in a separate save before making this zone Home" as a blocking warning — mirroring the backend's actual rejection condition instead of the pending edit's.
- Wired the baseline through the two real call sites (`EntityStore.recordStates`/`stateOf` in `entity-store.svelte.ts`, and `WorkbenchDetail.svelte`'s tab-warn triangle) via the store's existing `baselineOf(id)`.

## Test plan

- [x] `npm run test:unit:ci` — 310 test files / 3897 tests passing.
- [x] Added unit tests in `zone.test.ts` covering: blocks the same-save clear+flip, passes when spawns were already cleared in a prior save (baseline already empty), and passes a brand-new Home zone (no baseline).
- [x] Added a `validation.test.ts` case asserting `sectionWarnings` threads the baseline through to a `warn` predicate that needs it.
- No backend changes — this is purely a client-side warning gap; the backend's own guard (already correct) is what this warning now mirrors.

Closes #2314

---
_Generated by [Claude Code](https://claude.ai/code/session_01HZZzpZ98TnS5aGRig6KXeR)_

### DIFF
--- a/UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
+++ b/UI/src/routes/admin/workbench/components/WorkbenchDetail.svelte
@@ -30,7 +30,7 @@
 				label: section.label,
 				count: section.count?.(record) ?? null,
 				dirty: sectionDirty(section),
-				warn: sectionWarnings(section, record).length > 0
+				warn: sectionWarnings(section, record, baseline).length > 0
 			}))}
 			activeTab={tab}
 			onTab={(key) => onTab(key)}
--- a/UI/src/routes/admin/workbench/entities/types.ts
+++ b/UI/src/routes/admin/workbench/entities/types.ts
@@ -132,8 +132,11 @@ interface BaseSection<T> {
 	/** Tab count badge. */
 	count?: (rec: T) => number;
 	/** Section-level validation warning (e.g. "No skills assigned"). Return a {@link Warning} instead
-	 *  of a plain string to mark a known backend hard-reject condition as save-blocking. */
-	warn?: (rec: T) => string | Warning | null;
+	 *  of a plain string to mark a known backend hard-reject condition as save-blocking. `baseline` is
+	 *  the record's last-saved copy (undefined for a not-yet-saved add) — a predicate whose backend
+	 *  counterpart validates against persisted state rather than the pending record (e.g. a child
+	 *  saver that hasn't run yet) needs it to preempt the same rejection. */
+	warn?: (rec: T, baseline?: T) => string | Warning | null;
 	/**
 	 * Record keys that constitute a change for this section's tab dirty-dot. Used by
 	 * custom-kind sections (which have no single `fields`/`itemsKey` to diff); the
--- a/UI/src/routes/admin/workbench/entities/zone.ts
+++ b/UI/src/routes/admin/workbench/entities/zone.ts
@@ -162,9 +162,19 @@ export const zoneEntity: EntityConfig<WorkbenchZone> = {
 			// rolls at runtime (mirrors the backend's EmptyCombatZone lint). A Home zone with any
 			// spawn rows is a harder problem: it's hard-rejected on save (AdminZones.SaveZones'
 			// edit guard, AdminZones.SetEnemies), so it blocks rather than merely warns.
-			warn: (z) => {
+			warn: (z, baseline) => {
 				if (z.isHome) {
-					return z.zoneEnemies.length ? { message: 'The Home zone cannot have enemy spawns', blocking: true } : null;
+					if (z.zoneEnemies.length) {
+						return { message: 'The Home zone cannot have enemy spawns', blocking: true };
+					}
+					// SaveZones' edit guard reads the *persisted* spawn count (AdminZones.cs), evaluated
+					// before this save's own SetZoneEnemies child-saver clears it — so clearing spawns and
+					// flipping isHome on in the same save still hard-rejects even though the pending record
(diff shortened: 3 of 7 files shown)


ginderjeremiah (commented):
## Review: Preempt the Home-flip save rejection when spawns are cleared in the same edit (#2314)

**Verdict:** Approve — the client warning now mirrors the backend's actual rejection condition exactly, with clean threading and solid coverage. (Posted as a comment since I'm the PR author and GitHub won't take a gating self-review.)

I verified the fix against the backend guard it's mirroring (`Game.DataAccess/Repositories/Admin/AdminZones.cs:57-62`):

```csharp
if (change.ChangeType == EChangeType.Edit
    && change.Item.IsHome
    && _zones.LookupZone(change.Item.Id) is { ZoneEnemies.Count: > 0 })
```

The backend keys off the **persisted** spawn count, evaluated before the child `SetZoneEnemies` saver clears it. The new client branch checks `baseline?.zoneEnemies.length` — the last-saved copy, which is exactly the persisted state — so the two conditions now line up. The Add case (no baseline → `null`) matches the backend's `EChangeType.Edit`-only guard.

### Verification

- **All call sites threaded** — `entityWarnings`/`entityBlockingWarnings` (both `recordStates` and `stateOf` in `entity-store.svelte.ts`) and `sectionWarnings` (`WorkbenchDetail.svelte:33`) all pass the baseline via the store's existing `baselineOf(id)`. No caller was left un-threaded.
- **Save gate is actually closed** — `hasBlockingWarnings` reads `recordStates`, whose `blockingWarnings` now carries the baseline, so the same-save clear+flip is blocked before it can reach the backend, not just surfaced on the tab triangle.
- **Backward-compatible design** — making `baseline` optional on `SectionConfig.warn` keeps every other entity's `warn` predicate untouched; only the one predicate that needs persisted state opts in.

### Test coverage

Good. Three `zone.test.ts` cases cover the three branches (same-save clear+flip blocks, prior-save clear passes, brand-new Home with no baseline passes), and the `validation.test.ts` case asserts the plumbing threads `baseline` through to a predicate that needs it. No battle logic touched, so no frontend/backend mirroring concern; no backend change needed since its guard already exists.

### What's good

- Correctly rejected the "just add a code comment" option from #2314 in favor of the real fix, and scoped it minimally.
- The comments carry the non-obvious backend-timing rationale (why the pending record reads empty but the save still rejects) — the right level of detail for a subtle cross-layer interaction, not over-documentation.

No blocking findings, should-fixes, or nits. Will merge once CI is green.

---

PR #2315, opened 2026-07-22
Close the Workbench maxLength gap and bound LessonStep.Text

ginderjeremiah:
## Summary

Closes #2277.

The client-side `maxLength` convention (#2257, closed by #2262/#2263) didn't reach every Workbench entity config, and one authored text column (`LessonStep.Text`) was still left unbounded at the DB level.

- **Missing configs entirely**: `entities/enemy.ts` (Name 50, DesignerNotes 2000), `entities/class.ts` (Name 50, Description 500, Word 50, DesignerNotes 2000), `entities/lesson.ts` (Key 100, Name 100, ScreenKey 50, DesignerNotes 2000), `entities/skill-recipe.ts` (DesignerNotes 2000).
- **Partially missing**: `entities/skill.ts` was missing the word/pronunciation/translation bounds (50/50/100).
- **`LessonStep.Text`** was an unbounded `text` column — missed by the earlier #2243/#2256 bounding sweep, which only bounded `AnchorKey`. Bounded it at 500 chars (matching the `Description` precedent used elsewhere for player-facing authored text; the current authored content's longest step is 202 chars) via a new EF migration (`Game.Infrastructure/Migrations/20260722161431_BoundLessonStepText.*`).
- **`TableCell.svelte`** (the table-row editor used by, e.g., the lesson steps table) had no `maxLength` support at all for its `text` column type — added `ColumnConfig.maxLength`, threaded through mirroring `FieldControl`'s existing `text`/`textarea` support, and used it for the lesson steps table's Callout Text column.
- While in the same file family: every other config's `name` field (`item.ts`, `zone.ts`, `challenge.ts`, `item-mod.ts`, `tag.ts`) also never got a `maxLength` despite 50/100-char DB bounds — a one-line-each fix in the same pattern, so folded in here rather than filed separately.

## Test plan

- [x] `dotnet build` — 0 warnings, 0 errors.
- [x] `dotnet test --no-build --no-restore` — full backend suite green (Core, Infrastructure, Application, Api).
- [x] `npm run lint` (eslint + svelte-check) — clean.
- [x] `npm run test:unit:ci` — 310 test files / 3893 tests passing.
- [x] Verified the generated migration (`AlterColumn` to `character varying(500)`) matches the model snapshot and the current longest authored lesson-step text (202 chars) fits comfortably under the new bound.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HdPjTvyUkgrg3w9NHj4nC8)_

Issue #2277, opened 2026-07-22
Workbench client-side maxLength missed the enemy/class/lesson/skill-recipe configs; LessonStep.Text is still an unbounded DB column


ginderjeremiah:
## Problem

The client-side `maxLength` convention (#2257, closed by #2262/#2263) didn't reach every entity config. Configs with **zero** `maxLength` despite DB-bounded columns (`Game.Infrastructure/Database/GameContext.cs`):

- `entities/enemy.ts` — Name 50, DesignerNotes 2000 (`GameContext.cs:245-248`)
- `entities/class.ts` — Name 50, Description 500, Word 50, DesignerNotes 2000 (`:175-190`)
- `entities/lesson.ts` — Key 100, Name 100, ScreenKey 50, DesignerNotes 2000 (`:730-739`)
- `entities/skill-recipe.ts` — DesignerNotes 2000 (`:579`)

`entities/skill.ts` also misses the word/pronunciation/translation bounds (50/50/100, `:523-531`), and `name` fields across configs generally lack `maxLength` despite 50–100 char DB bounds.

Separately, **`LessonStep.Text` is authored content left as unbounded `text`** (`Game.Infrastructure/Entities/LessonStep.cs:12`) — missed by the #2243/#2256 bounding sweep, which bounded only `AnchorKey`. This is the one migration-bearing part.

## Cost

An over-length value (a 2001-char designer note is entirely plausible) sails past the client and dies as an EF `DbUpdateException` → opaque 500, the exact failure #2257 set out to close.

## Notes

The lesson-step editor uses `TableCell`, which currently has no `maxLength` support — it needs to learn it for the step text once the DB bound exists.

<sub>From the scheduled codebase health review.</sub>

PR #2262, opened 2026-07-21
Extend Workbench client-side maxLength to the remaining already-bounded editable fields

ginderjeremiah:
## Summary

- Adds the matching client-side `maxLength` to every already-`HasMaxLength`-bounded Workbench field the #2256 convention missed: `Skill.IconPath` (50), `Challenge.Description` (500), and the `DesignerNotes` field on `Item`/`ItemMod`/`Skill`/`Zone`/`Challenge` (2000 each).
- Threads a new `maxLength` prop through `ProgInput.svelte` (the progression editor's own text/textarea control — separate from the generic `FieldConfig`/`FieldControl` system used everywhere else in the Workbench) and wires it to `Path.Description` (500) in `PathDetail.svelte`.
- No backend/migration changes — every DB bound already exists in `GameContext.cs`; this is purely additive client-side UX that prevents over-typing the backend would reject anyway.

## Scope note

`Proficiency.Description` was in the issue's list but is verified **not editable anywhere in the Workbench** — `ConlangIdentity.svelte` (the tier/proficiency editor) only exposes `name`/`iconPath`/`word`/`pronunciation`/`translation`/`designerNotes`, never `description`. So there's no UI counterpart to update for that column, mirroring the same call made for `Attribute.Description` in #2256.

While verifying the progression editor, I found other already-bounded `Path`/`Proficiency` fields (`Path.Name`, `Path.DesignerNotes`, `Proficiency.Name`/`IconPath`/`Word`/`Pronunciation`/`Translation`/`DesignerNotes`) that also lack the client-side bound now that `ProgInput` supports it — that's a distinct scope from this issue's explicit list, so I filed #2261 to track it rather than expanding this PR.

## Test plan

- [x] `npm run lint` (eslint + svelte-check) — 0 errors
- [x] `npm run test:unit:ci` — 310 files / 3879 tests passed
- No new tests added: this mirrors #2256's precedent exactly (a plain declarative `maxLength` value passed through the existing generic `field.maxLength` → `maxlength` attribute plumbing, which that PR also shipped without dedicated tests).

Closes #2257

---
_Generated by [Claude Code](https://claude.ai/code/session_01KX5xDYYwRRgR6jnHGtiqVb)_

Issue #2257, opened 2026-07-21
Extend Workbench client-side maxLength to the remaining already-bounded editable fields


ginderjeremiah:
## Background

#2256 (closing #2243) introduced a client-side `maxLength` convention on the admin Workbench: an optional `maxLength` prop on `FieldConfig` threaded through `FieldControl.svelte` to the `textarea`/text `input` as the HTML `maxlength` attribute. It was correctly scoped to only the fields that issue touched (`Item.Description`/`Item.IconPath`, `ItemMod.Description`, `Skill.Description`, `Zone.Description`).

## Problem

Other editable fields whose DB columns are already bounded via `HasMaxLength` still lack the matching client-side `maxLength`, so the new convention is applied inconsistently:

- `Skill.IconPath` — DB-bounded at 50, but `skill.ts` iconPath field has no `maxLength` (whereas `Item.IconPath`, an identical case, now does).
- `Challenge.Description` — DB-bounded at 500, no client `maxLength`.
- `Proficiency.Description` / `Path.Description` — DB-bounded at 500 (if editable via the Workbench), no client `maxLength`.
- All `DesignerNotes` fields (`Item`/`ItemMod`/`Skill`/`Zone`/`Challenge`) — DB-bounded at 2000, no client `maxLength`.

## Fix

For each editable Workbench field backed by a `HasMaxLength` column, set the field config's `maxLength` to match the DB bound. This is purely additive UX (prevents over-typing that the backend would reject) and completes the convention introduced in #2256. No backend/migration changes are needed — the DB bounds already exist.

Verify against `GameContext.cs` that each field's `maxLength` matches its column's `HasMaxLength` value.

PR #2325, opened 2026-07-23
Add EF-mirrored guard test for Workbench field maxLength

### LINKED ISSUE #2318: Add a guard test that Workbench text field/column maxLength matches the EF DB bound
## Problem

The client-side Workbench `maxLength` convention has now drifted from the EF-configured DB column bounds by hand **three times** in a row: #2257 → #2262/#2263 (initial sweep), then #2277 (missed enemy/class/lesson/skill-recipe configs), then again the skill `name` and lesson `anchorKey` gaps caught in review of #2315. Each time the fix is correct-but-manual, so the next new field re-opens the gap.

## Proposal

Add a frontend guard test over the Workbench entity configs (`UI/src/routes/admin/workbench/entities/*.ts`) asserting that every `text`/`textarea` `FieldConfig` and every `text` `ColumnConfig` bound to a DB-backed column carries a `maxLength`, and (ideally) that the value equals the EF-configured length for that column.

Matching against the EF bounds requires the DB lengths to be available to the test. Options to evaluate:
- Export the column bounds from a shared source both EF (`GameContext.cs`) and the frontend can read (e.g. a generated constants file), and assert equality — strongest, kills the whole drift class.
- Failing that, at minimum assert presence of `maxLength` on every applicable field/column so a new unbounded input fails CI, even if the exact value still needs manual verification.

An over-length value currently sails past the client and dies as an EF `DbUpdateException` → opaque 500, which is exactly the failure the original convention (#2257) set out to close — so a guard here has real user-facing value.

## Scope

Test-only + possibly a small shared-constants extraction. No user-facing behavior change.

<sub>Follow-up from review of #2315.</sub>

ginderjeremiah:
## Summary

The Workbench `maxLength` convention (#2257) — hand-authored on `FieldConfig`/`ColumnConfig` to mirror the EF `HasMaxLength` DB bound — has drifted from that bound by hand three times in a row (#2262/#2263, #2277, and the review of #2315). This closes the gap with the **strong** option from #2318 rather than a presence-only check: a single generated source of truth both sides read.

- **`Game.Core/ContentFieldLengths.cs`**: a new `[ClientMirrored]` constants class holding the `HasMaxLength` bound for every EF column the Workbench actually edits (34 constants across the 10 Workbench-backed entities — `Challenge`, `Class`, `Enemy`, `Item`, `ItemMod`, `Skill`, `SkillRecipe`, `Lesson`/`LessonStep`, `Tag`, `Zone`).
- **`GameContext.cs`**: the 34 corresponding `.HasMaxLength(...)` calls now reference these named constants instead of inline literals (values unchanged — no migration needed). The other ~30 `HasMaxLength` calls, for entities with no Workbench editor, are untouched.
- The existing codegen pipeline (`ApiCodeGenerator` → `ConstantsWriter`) already mirrors any `[ClientMirrored]` class's constants into `UI/src/lib/api/types/game-constants.ts` — no new codegen writer needed, and it's already covered by CI's codegen-drift check.
- **`UI/src/tests/routes/admin/workbench/entities/field-lengths.test.ts`**: walks every Workbench entity's `fields`/`table` sections, and for every `text`/`textarea` field or column asserts its `maxLength` equals the matching generated constant. A field with no registered expectation (new, unmapped) fails loudly with a message pointing at both files to update, closing the exact silent-drift gap that reopened three times.
- Fixed `GetClientMirroredConstantFields_IsOrderedDeterministically` (`Game.Api.Tests`), which asserted a coincidental *global alphabetical* order that only held while `GameConstants` was the sole `[ClientMirrored]` class. It now asserts the actual ordering contract (grouped by declaring type, then field name), which holds regardless of how many mirrored classes exist. Added a small theory test covering `ContentFieldLengths` alongside the existing `GameConstants` one.
- One-line doc note in `docs/frontend-admin.md` recording the convention.

Today's fields are already correct (34/34 pass) — this is regression-prevention, verified by manually breaking a value and confirming the test catches it (then reverting).

## Closes

Closes #2318

## Test plan

- [x] `dotnet build` — full solution builds clean
- [x] `dotnet test` — full backend suite green (Game.Core.Tests, Game.Infrastructure.Tests, Game.Api.Tests, Game.Application.Tests)
- [x] `dotnet run --project Game.Api -- codegen` — regenerated `game-constants.ts`, confirmed idempotent (no further diff on a second run), matching CI's drift check
- [x] `npm run test:unit:ci` — full frontend suite green (3931 tests), including the new guard test (34/34)
- [x] Manually broke a Workbench `maxLength` value and confirmed the new test fails with a clear message, then reverted

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ke1xr8gWgoxcaWJ7t4dkVa)_


ginderjeremiah (commented):
## Review: EF-mirrored guard test for Workbench field `maxLength` (closes #2318)

**Verdict:** Approve — implements the strong option #2318 asked for (a single generated source of truth both sides read), and does it cleanly. _(Submitted as a comment since I'm the PR author; the intended verdict is Approve.)_

### Correctness

- The `maxLength` drift class is genuinely closed, not just narrowed. `ContentFieldLengths` (`[ClientMirrored]`) → `game-constants.ts` → `field-lengths.test.ts` gives EF and the Workbench configs one shared value to agree on, and an unregistered `text`/`textarea` field fails loudly rather than passing uncovered — the exact silent gap that reopened three times.
- Coverage is complete: all 34 `text`/`textarea` fields/columns across the 10 Workbench entities map 1:1 to the 34 constants. `ColumnConfig.type` has no `textarea` (only `FieldConfig` does), so the field check (`!== 'text' && !== 'textarea'`) and the column check (`!== 'text'`) skip nothing DB-backed.
- The `GetClientMirroredConstantFields_IsOrderedDeterministically` change is a real fix, not a test-weakening: the generator orders `OrderBy(DeclaringType.Name).ThenBy(Name)`, so the old flat-alphabetical assertion only held by coincidence while `GameConstants` was the sole mirrored class. The new assertion states the actual contract and survives additional mirrored classes.

### Layering

- `ContentFieldLengths` in `Game.Core` is the correct (and required) home — the codegen scans the `Game.Core` assembly for `[ClientMirrored]` classes, and `Game.Infrastructure → Game.Core` is a valid dependency direction. The `GameContext` changes are literal-for-constant swaps with unchanged values, so no migration.

### Test coverage

- Backend: new theory covering `ContentFieldLengths` alongside the existing `GameConstants` one, plus the corrected ordering test.
- Frontend: the new `field-lengths.test.ts` walks every entity/section and asserts each text field/column against its generated constant, with a clear failure message pointing at both files to update.
- Doc note added in `frontend-admin.md`.

### What's good

- Picking the strong shared-source option over presence-only is the right call — it removes the manual step that kept reopening the gap. Verification-by-breaking-a-value (then reverting) is a nice touch for a regression guard.

---

PR #2268, opened 2026-07-22
Refresh npm dependencies and add dependabot automation

ginderjeremiah:
## Summary

- Ran `npm update` in `UI/` to clear the ~22 packages of in-range semver drift the lockfile had accumulated (also dropped 4 high-severity audit findings to 0 as a side effect).
- Bumped `@types/node` from `^25.9.1` to `^26.1.1` to match the runtime the project already pins (`UI/.nvmrc` / `package.json` `engines.node: >=26`).
- Reformatted 5 files (`auth.ts`, `battle-attributes.ts`, `FieldControl.svelte`, `ProgInput.svelte`, `engine.test.ts`) that the prettier `3.8.4` → `3.9.6` bump reflowed — pure formatting, no behavioral change, confirmed by diffing against the pre-update lint output (which was clean) and re-running `eslint --fix`.
- Added `.github/dependabot.yml` covering the `npm` (`/UI`), `nuget` (`/`), and `github-actions` (`/`) ecosystems on a weekly cadence, so this kind of drift gets surfaced automatically going forward instead of only via periodic manual health reviews.
- Left the three pending majors (`eslint` 10, `typescript` 7, `@testing-library/jest-dom` 7) untouched — each needs its own migration/testing pass, so I filed #2267 as a follow-up rather than folding a risky bump into a routine refresh.

## Addresses

Closes #2068.

## Test plan

- [x] `npm outdated` — only the three majors remain (tracked in #2267); everything else is at "Wanted".
- [x] `npm run lint` (eslint + svelte-check) — 0 errors, 0 warnings.
- [x] `npm run test:unit:ci` — 310 test files / 3879 tests passing.
- [x] No backend/NuGet files touched, so no `dotnet build`/`dotnet test` regression risk from this change; `dependabot.yml`'s `nuget` entry is config-only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mp6X8XFRitzKKT8ghdKbLY)_

PR #2327, opened 2026-07-23
Group npm dev-dependency minor/patch updates into one dependabot PR

### LINKED ISSUE #2282: Consider grouping dependabot updates to reduce per-package PR noise
## Context

Follow-up from #2268, which added `.github/dependabot.yml` covering npm (`/UI`), nuget (`/`), and github-actions (`/`) on a weekly cadence.

With the current config, dependabot opens a **separate PR per dependency** each week. For the npm ecosystem in particular (many devDeps), a routine minor/patch sweep could produce a handful of near-identical PRs, which is noisy to triage.

## Suggested fix

Add a `groups` block to the npm ecosystem entry (and optionally nuget/actions) to bundle related bumps into a single PR — e.g. group dev-dependency minor/patch updates:

```yaml
groups:
  dev-minor-patch:
    dependency-type: "development"
    update-types:
      - "minor"
      - "patch"
```

This is optional/deferred rather than a defect — the default per-dependency behavior is functionally correct, just noisier. Worth revisiting once we've observed a week or two of real dependabot output to see how much noise there actually is before committing to a grouping scheme.

ginderjeremiah:
## Summary

- Adds a `groups` block to the `npm` ecosystem entry in `.github/dependabot.yml` that bundles minor/patch dev-dependency bumps into a single weekly PR.
- Every dependency in `UI/package.json` is a `devDependency` (there's no `dependencies` section — it's a SvelteKit tooling-only frontend), so this group covers the entire npm surface without needing to also match a `dependencies`-type group.

## Why

Closes #2282. With the config added in #2268, dependabot opens a separate PR per npm package each week, which is noisy given how many devDeps this project has. Grouping minor/patch bumps mirrors the pattern the `nuget` ecosystem entry already uses (`dotnet-runtime` pattern group) — major bumps are intentionally left ungrouped so they still surface individually for review.

## Scope notes

Per the issue, this only touches `npm`. I left `nuget` and `github-actions` alone:
- `nuget` already has a pattern-based group for the highest-volume families (EF Core, `Microsoft.Extensions.*`, `Microsoft.AspNetCore.*`); the remaining ungrouped packages are few enough that a second group didn't seem justified without evidence of noise.
- `github-actions` has a single workflow file with a handful of actions — unlikely to produce meaningful PR noise.

If a week or two of real output shows otherwise, that can be revisited as a follow-up per the issue's own framing.

## Test plan

- [x] Validated the resulting YAML parses correctly (`python3 -c "import yaml; yaml.safe_load(...)"`)
- Config-only change to `.github/dependabot.yml`; no application build/test suite covers dependabot behavior. Actual grouping behavior will be observable on dependabot's next weekly run.

Closes #2282

---
_Generated by [Claude Code](https://claude.ai/code/session_01NnunoAVHWAeXgoMtJYW5eA)_

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,12 @@ updates:
     directory: "/UI"
     schedule:
       interval: "weekly"
+    groups:
+      dev-minor-patch:
+        dependency-type: "development"
+        update-types:
+          - "minor"
+          - "patch"
     ignore:
       - dependency-name: "@emnapi/core"
       - dependency-name: "@emnapi/runtime"


ginderjeremiah (commented):
## Review: Group npm dev-dependency minor/patch dependabot updates (closes #2282)

**Verdict:** Approve (posted as a comment since I'm the PR author) — the change is a minimal, correct config-only edit that implements exactly what the issue asked for.

I verified the substance rather than just the diff:

- `UI/package.json` has a `devDependencies` section and **no** `dependencies` section, so `dependency-type: "development"` genuinely covers the entire npm surface — the PR's claim holds.
- The `groups` block is valid dependabot v2 syntax, placed correctly at the update-entry level (sibling of `directory`/`schedule`/`ignore`, where key order is irrelevant), and mirrors the style of the existing `nuget` `dotnet-runtime` group.
- Major bumps are intentionally left ungrouped so they still surface individually for review — a sensible default.
- The two `@emnapi/*` entries remain in `ignore`, so grouping doesn't affect them.

### Blocking

- None.

### Should-fix / Nits

- None.

### Scope

- Leaving `nuget` and `github-actions` untouched is the right call per the issue's "revisit after observing real output" framing — no need to speculatively add more groups now.

### What's good

- Tight, well-scoped change with an accurate description; the reasoning for what was *not* changed is spelled out and correct.

---

PR #2346, opened 2026-07-23
Check active-session takeover before committing to a character switch (#1518)

### LINKED ISSUE #1518: Character switcher: takeover check runs after the irreversible commit — declining destroys a healthy session

ginderjeremiah:
## Summary

`PlayerSelectView.select()` (shared by the login character-select screen and the in-game character switcher) used to call `selectPlayer`/`switchPlayer` first — rotating the token and, for the switcher, tearing down the currently-live game — and only *then* run the active-session takeover check. Declining at that point called `logout()`, which meant saying "no" to a takeover of a *different* character could destroy an already-healthy, currently-playing session with no way back.

This PR reorders the check to run **before** any commit, using an explicit target rather than the token's currently-selected player:

- `Auth/ActiveSession` now takes a `playerId` (validated against the caller's account, mirroring `SelectPlayer`'s ownership check) instead of reading `_sessionService.SelectedPlayerId` off the token. It no longer needs `SessionInitializer`/the session cache at all, since it doesn't depend on the token's own selected player.
- `PlayerSelectView.select(playerId)` now calls `confirmTakeover(playerId)` first; `selectPlayer`/`switchPlayer` (and, for the switcher, the live-game teardown baked into its `selectPlayer` implementation) only run once the takeover is confirmed.
- Declining is now a true no-op: no token rotation, no teardown, nothing to log out of. The login-flow `logout()`-on-decline call was removed since there's no longer a "just-issued session" to clean up.

Closes #1518. Matches the issue's suggested direction (a parameterized presence check) — the ownership-check shape mirrors `PlayersController.SelectPlayer`/`SwitchPlayer`'s existing `OwnsPlayer` pattern, so no new API design surface was needed.

Docs updated: `docs/backend-auth.md` (`ActiveSession` ordering + session-cache independence) and `docs/frontend-screens.md` (select-screen and switcher flow descriptions), plus the doc comments on `SessionService.LoadPlayerState`/`SessionInitializer.EnsureSessionLoaded` that listed `ActiveSession` as a session-cache consumer.

## Test plan

- [x] Backend: `AuthControllerTests` — reworked the pre-selection-token test (now checks an explicit target instead of asserting a `NoPlayerSelected` 409), added an unowned-player anti-cheat test, and a test pinning that `ActiveSession` never touches the session cache. Full backend suite green (`dotnet test Game.sln`, 3449 tests).
- [x] Frontend: reordered assertions in `player-select-view.test.ts` and `character-switcher.test.ts` (including a new "declining leaves the live game running" case), updated `session-takeover.test.ts` for the dropped `logout()` call. Full frontend suite green (`npm run test:unit:ci`, 3932 tests) plus `npm run lint`.
- [x] Regenerated the TypeScript API client (`dotnet run --project Game.Api -- codegen`) for the new `Auth/ActiveSession` request shape — clean diff, no drift.
- [x] `dotnet format --verify-no-changes` clean.

Closes #1518

---
_Generated by [Claude Code](https://claude.ai/code/session_01P3414g5hFSfyVj9Zc9AVHh)_


ginderjeremiah (commented):
## Review: Reorder the active-session takeover check to run before the irreversible commit (#1518)

**Verdict:** Approve — the ordering bug is fixed correctly on both surfaces, the parameterized presence check faithfully reuses the existing anti-cheat pattern, dead code is removed, and tests + docs are updated thoroughly. (Submitted as a comment since I'm the PR author; treat it as an approval.)

### Why this is correct

- **The core reorder is sound.** `PlayerSelectView.select()` now runs `confirmTakeover(playerId)` *before* `selectPlayer`. Since `pendingId` is set first, `busy` guards a concurrent select through the modal; a decline clears `pendingId` and returns without ever binding — no token rotation, and for the switcher no teardown. This is exactly the "declining is a true no-op" outcome the issue asked for.
- **`ActiveSession(playerId)` matches the existing anti-cheat shape precisely.** It calls the same `AccountService.OwnsPlayer(userId, playerId, ...)` that `CharacterSelectionService.SelectPlayer` uses, and returns the same `ApiErrorCategory.BadRequest` + identical message that `SelectPlayer`'s `NotOwned` maps to. The `_sessionService.UserId` it keys on is recorded by the middleware on every authenticated request, so it resolves correctly even for a pre-selection token — which is what let the endpoint drop `EnsureSessionLoaded` and the `TokenSelectedPlayerId` guard.
- **Cache-independence claim holds.** Dropping `_sessionInitializer.EnsureSessionLoaded` from `ActiveSession` is safe — the endpoint no longer reads the token's selected player, and `_sessionInitializer` is still used by `Status()`, so no dead dependency. The `NoSessionCacheEstablished` test pins that the cache is never touched.
- **Login-flow behavior change is intentional and better.** Removing the `logout()`-on-decline means a declined takeover now rests on the select screen (still holding the pre-selection token) instead of bouncing to login. The `logout` import and its test double are cleaned up accordingly — no dead code left behind.

### Test coverage

Solid and mirrored on both sides. Backend: reworked the pre-selection test to check an explicit target, added `ActiveSession_UnownedPlayer_ReturnsError` (anti-cheat), and a cache-untouched test. Frontend: `player-select-view` reordered to assert takeover-before-bind and a decline-never-calls-selectPlayer case, `session-takeover` updated for the dropped `logout()`, and `character-switcher` gained a "declining leaves the live game running" case asserting no teardown/`SwitchPlayer` on decline. No battle-logic surface touched, so no cross-suite mirroring obligation.

### What's good

The XML doc on `ActiveSession` and the `confirmSessionTakeover` comment are longer than average but earn it — they capture the ordering rationale that is the whole point of the fix. Docs (`backend-auth.md`, `frontend-screens.md`) and the affected code comments were all brought in line with the new flow rather than left to drift.

Nothing blocking and no nits. Approving; will merge once CI is green.

---

PR #2347, opened 2026-07-23
Order reference-data mapper child collections deterministically

### LINKED ISSUE #2330: Reference contract/domain mappers don't order child collections — version hash instability and a parity-sensitive tie-break ride on unguaranteed EF row order
## Problem

The snapshot builders order the **parent** lists (`OrderBy(x => x.Id)` in every `*CacheHolder`), but the `Include`d child collections carry no ordering, and Postgres does not guarantee child-row order across plans/reloads (the `AsSplitQuery` conversion in #485 makes the order even less pinned). The mappers serialize children in whatever order EF materialized them, and that order feeds two order-sensitive consumers:

- `Game.Api/Sockets/Commands/ReferenceDataVersioning.cs:30` — the version hash is explicitly "deterministic for a given set/**order**".
- `Game.Core/Skills/PrimaryDamageTypeResolver.cs:26-34` — resolves a skill's primary damage type with a **first-wins tie-break** over `DamagePortions`; order decides the winner for tied weights (e.g. an authored 50/50 split).

The project already knows the invariant — `SkillRecipeMapper.cs:13` orders inputs/conditions "so the reference set's version hash is stable across reloads", and `LessonMapper` orders steps by ordinal — but the other mappers omit it:

- `Game.DataAccess/Mapping/SkillMapper.cs` (portions, multipliers, effects — both `ToContract` and `ToCore`)
- `Game.DataAccess/Mapping/ItemMapper.cs` (attributes, mod slots, tags; item-mod variants)
- `Game.DataAccess/Mapping/EnemyMapper.cs` (distributions, skill pool, spawns)
- `Game.DataAccess/Mapping/ClassMapper.cs` (+ starter equipment/distributions)
- `Game.DataAccess/Mapping/ProficiencyMapper.cs` (modifiers, rewards, prerequisites)

## Failure scenarios

1. Every instance reloads all holders at least every 5 minutes (reconciliation sweep), re-hashing from a fresh EF query. If child order shifts (plan change, heap order after an admin edit, parallel scan), the version hash changes with **no content change** — every client's loading screen re-downloads the full set, and two instances can disagree on the "version" of identical content indefinitely.
2. Parity: after a mid-session snapshot swap, the backend's `SkillDamagePortions` order can differ from the client's still-cached copy. For a tied-weight skill, `PrimaryDamageTypeResolver` then resolves a different primary type on each side — the weapon-match gate diverges between the client's simulation and the server's replay, i.e. a false anti-cheat rejection for an untampered client.

## Suggested fix

Deterministic `OrderBy` in each mapper (or in the snapshot builders), mirroring what `SkillRecipeMapper`/`LessonMapper` already do. Worth a small guard test that maps a deliberately shuffled entity graph and asserts stable output order.

ginderjeremiah:
## Summary

`SkillMapper`, `ItemMapper`, `EnemyMapper`, `ClassMapper`, and `ProficiencyMapper` serialized their EF child collections (damage portions, attribute distributions, skill pools, spawn tables, starter equipment, level modifiers, prerequisites, etc.) in whatever order EF materialized them. Postgres doesn't guarantee child-row order across query plans or reloads, and that order fed two order-sensitive consumers:

- `ReferenceDataVersioning`'s content hash, which is explicitly "deterministic for a given set/**order**" — an unstable child order changes the hash with no actual content change, forcing every client to re-download the full set and letting two instances disagree on a set's version indefinitely.
- `PrimaryDamageTypeResolver`'s first-wins tie-break over a skill's `DamagePortions` — for a tied-weight split, a diverging order between the backend's fresh reload and the client's still-cached copy could resolve a different primary type on each side, causing a false anti-cheat rejection for an untampered client.

This mirrors the invariant `SkillRecipeMapper` and `LessonMapper` already established (ordering inputs/conditions/steps "so the reference set's version hash is stable across reloads") but the other five mappers omitted it.

## Changes

- Added a deterministic `OrderBy` (on id, or a natural composite key like `AttributeId`/`DamageType`/`ZoneId`/`Level`) to every child collection projected on the read paths (`ToContract`/`ToCore`/`ToTemplate`) of the five mappers. Write paths (`ToEntity`) are untouched, matching the existing `SkillRecipeMapper` convention.
- Updated the `PrimaryDamageType` tie-break documentation (`PrimaryDamageTypeResolver`, `Skill.cs`, `docs/backend-reference-fields.md`, `docs/backend-battle.md`, and the frontend mirrors in `damage-types.ts`/`skill.ts`) to describe the resulting deterministic rule (lowest-numbered damage type wins a tie) instead of the previously-aspirational "first-authored" wording — the schema never actually persisted an authored order, so that language was describing behavior the code couldn't guarantee.
- Added guard tests per mapper that feed in a deliberately shuffled entity graph and assert the mapped output comes back in the expected deterministic order.

I checked whether `EnemySkills`' order is also parity-sensitive for battle simulation (not just the reference-data hash): it isn't — `Enemy.SelectBattleSkills`/`SelectAllBattleSkills` compute the enemy's loadout server-side and the *result* is snapshotted and sent to the client as data, so the client never independently re-derives skill-pool order from its own cached copy.

## Test plan

- [x] `dotnet build` — 0 warnings/errors across the full solution.
- [x] `dotnet test` (all four backend test projects, including DataAccess integration tests against the real Postgres/Redis containers) — 3457 passed, 0 failed.
- [x] New/updated mapper unit tests specifically verify shuffled-input → deterministic-output ordering for each affected collection.

Closes #2330

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xz1LYNgFsWu7Q4gA6DVJbG)_


ginderjeremiah (commented):
## Review: deterministic ordering of reference-data mapper child collections (#2330)

**Verdict:** Approve — the change addresses #2330 exactly, follows the established `SkillRecipeMapper`/`LessonMapper` convention, and is well-tested. (Posted as a comment because I'm also the PR author; treat it as an approval.)

I verified the change against the issue's two failure modes and the mappers themselves:

- **Version-hash stability:** every child collection projected on a read path (`ToContract`/`ToCore`/`ToTemplate`) across the five mappers now has an `OrderBy`, and every ordering key is a *unique* child key — `DamageType`, `AttributeId`, `Id`, `EquipmentSlotId`, `ZoneId`, `(Level, AttributeId)` for level modifiers, `Level` for level rewards (keyed one-per-level). No tie is left unbroken, so the ordering is a pure function of content and the hash stops riding on EF row order. Write paths (`ToEntity`) are correctly left untouched, matching `SkillRecipeMapper`.
- **Parity tie-break:** `PrimaryDamageTypeResolver` keeps its strict `>` first-wins tie-break, and both the server contract and the domain `ToCore` now emit `DamagePortions` ordered by `DamageType`. Since the client caches the contract in that same order and mirrors the strict `>` in `damage-types.ts`, a stale client copy and a fresh server reload now resolve the identical primary type on a weight tie — the false anti-cheat rejection can no longer occur. The doc/`Skill.cs`/frontend wording is updated from the aspirational "first-authored" to the accurate "lowest-numbered type on a tie".

### Test coverage
- One shuffled-input → deterministic-output guard test per mapper, covering both `ToContract` and `ToCore` where both are order-sensitive (Skill, Item, Proficiency) and the single relevant path otherwise. This matches the issue's suggested guard and the project's testing rules for domain/data-access logic.
- No frontend battle-logic behavior changed (only a doc comment in `damage-types.ts`/`skill.ts`), so no mirrored frontend test is required.

### What's good
- Correctly scoped to read paths only; the note in the PR body about `EnemySkills` not being parity-sensitive (loadout is computed server-side and snapshotted) is accurate and shows the parity surface was actually checked rather than assumed.
- Ordering keys chosen as natural composite keys rather than arbitrary — self-documenting and stable.

---

PR #2348, opened 2026-07-23
EnemyManager: give battle-end continuations a session epoch (#2331)

### LINKED ISSUE #2331: EnemyManager battle-end continuations survive a stop()→start() cycle — no start epoch, so a zombie resolution can clobber the next session's battle
## Problem

Every guard inside the battle-end resolutions checks *live* state that a restart resets back to "valid" — there is no session/start epoch:

- `watchIdleStage` (`UI/src/lib/engine/battle/enemy-manager.ts:570-599`) re-checks only `this.#transitioning || this.mode !== 'idle'` after its awaits (:585) — never `this.started`, and captures no generation at handler entry. `getNewEnemy` captures `#fetchGeneration` **at call time** (:213), i.e. after the stale continuation resumes — post-restart that generation is current, so the guard passes.
- `resolveBossLoss` (:689-725) snapshots `#transitionGeneration` (:694), but neither `stop()` nor `start()` bumps that counter (only `beginTransition` does), and `transitionCurrent` (:554-556) reads the live `this.started` — true again after restart.

## Failure scenario

Player wins → `claimVictory()`'s `DefeatEnemy` is in flight (up to 30s under the transport timeout; longer across a reconnect, and `claimVictory`'s error path adds an `await resyncPlayerAndInventory()` window). The player navigates off `/game` — the route unmount calls `stopEngines()` (`UI/src/routes/game/+page.svelte:83`), which does **not** tear down the socket, so the command stays in flight — then re-enters and `startGame` runs. The zombie continuation resumes and:

1. `await battleEngine.startLoading(cooldown)` stomps the new session's live battle stage to `Loading` mid-fight;
2. passes the mode/transition guard and calls `getNewEnemy(prepared)` with a stale prepared battle, with the fetch loop reporting `clientBattleMs = battleEngine.timeElapsed` — the elapsed time of the *new* session's unrelated battle — force-abandoning it server-side and spawning over the live fight;
3. in the boss-loss variant, the stale `returnToIdle()` also emits `SetAutoChallengeBoss(false)`, clobbering a just-reconciled persisted boss mode.

A smaller in-session variant needs no restart: `enterHome()` during an in-flight `claimVictory` — the resolution's `startLoading(cooldown)` flips the resting `Idle` stage to `Loading` and nothing sets it back while parked at Home (the follow-up `getNewEnemy` exits on the Home guard without touching the stage).

The enemy-manager suites cover `stop()` landing mid-await extensively, but not a stop **followed by a restart** while a battle-end continuation is parked — exactly the untested hole.

## Suggested fix

A start-epoch bumped in `start()`/`stop()`, captured at `watchBattleStage` entry and re-checked after every await (and/or bump `#transitionGeneration` in `stop()`). This is the natural trigger to also fold the guards into #1736's planned supersedable-operation primitive — but it is an actual defect today, not just consolidation debt.

ginderjeremiah:
## Summary

`watchIdleStage`, `resolveBossVictory`, and `resolveBossLoss` (`UI/src/lib/engine/battle/enemy-manager.ts`) re-checked only live, mutable state (`started`/`mode`/`transitioning`) after their internal awaits — state a `stop()`→`start()` cycle resets back to "valid" rather than actually invalidating. A continuation parked on an in-flight `claimVictory()`/`BattleLost` call when the player navigates off `/game` (`stop()`) and back on (`start()`) before it resolves would resume believing it still owns the loop, and could stomp the fresh session's battle stage or spawn over its fight.

A smaller in-session variant hit the same root gap: entering the no-combat Home sanctuary mid-claim (which calls `interruptFetch()`) left the resolution free to still call `battleEngine.startLoading(cooldown)`, flipping the resting state to `Loading` with nothing to set it back.

Rather than bolt on a fifth ad-hoc guard, this leans on #1736's explicit ask ("the next time a feature touches this loop, consolidate the guards into a single supersedable-operation primitive... rather than adding a fifth mechanism") — this PR is exactly that trigger:

- Unifies the previously-separate `#fetchGeneration` / `#transitionGeneration` counters into one `#generation` epoch.
- `#generation` is now also bumped on every `start()` and `stop()` (in addition to the existing `interruptFetch()` triggers: Home entry, control transitions), so a stop/restart genuinely invalidates any continuation captured before it.
- `watchBattleStage` captures `#generation` once and threads it down to `watchIdleStage`/`watchBossStage`/`resolveBossVictory`/`resolveBossLoss`, each re-checking it via a single `isCurrent(generation)` helper after every await — replacing the old `bossLoopActive` getter and `transitionCurrent` method with generation-aware equivalents.
- `stop()` now also force-clears `#transitioning`, since a zombie transition's own `finally`-block `endTransition` can no longer rely on the (now-bumped) generation staying put to release the guard.

This closes both failure scenarios described in #2331 with the one mechanism, and reduces rather than grows the loop's guard count.

## Test plan

- [x] `npx vitest --run` — all 3935 frontend tests pass (311 files), including the full existing `enemy-manager`/`enemy-manager-boss` suites (81 pre-existing tests, unmodified assertions, all still green).
- [x] `npm run lint` (ESLint + svelte-check) — 0 errors/warnings.
- [x] Added 3 new tests in `enemy-manager-boss.test.ts`:
  - an idle-victory resolution spanning a `stop()`→`start()` cycle
  - a boss-loss resolution spanning a `stop()`→`start()` cycle
  - the in-session Home-entry-mid-claim variant
  - Verified all three fail against the pre-fix code (confirmed by stashing just the source change and re-running) and pass with the fix.
- [x] No backend files touched — this is a frontend-only client concurrency fix, not a battle-logic parity change, so no backend test suite is affected.

Closes #2331

---
_Generated by [Claude Code](https://claude.ai/code/session_01M3NGcdKbJsaQSTa59tAxTP)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -77,21 +77,21 @@ export class EnemyManager {
 	#battleStageUnhook?: Action;
 	/** Guards the challenge/retreat transitions so a resolving stage change for the battle being
 	 *  swapped out is not mistaken for an outcome of the new one. Stays set across a supersession (a
-	 *  press that takes over an in-flight transition) — only the transition that ends up current clears it. */
+	 *  press that takes over an in-flight transition) — only the transition that ends up current clears it.
+	 *  Force-cleared by {@link stop}, since a torn-down manager can't rely on a zombie transition's `finally`
+	 *  to release it (that `finally` may run after {@link start} begins a fresh session, see `#generation`). */
 	#transitioning = false;
-	/** Bumped whenever a new transition supersedes an in-flight one. A running transition captures this at
-	 *  its start and, after each await, abandons (without applying its result) once it no longer matches — so
-	 *  a superseded challenge/retreat can't clobber the transition that took over, and the superseded one's
-	 *  finally won't release the `transitioning` guard the new transition still holds. */
-	#transitionGeneration = 0;
+	/** The single supersedable-operation epoch (consolidating the ad-hoc `fetchGeneration`/`transitionGeneration`
+	 *  pair per #1736): bumped by {@link interruptFetch} (a Home entry, a control transition, or a stop) and by
+	 *  every fresh {@link start}. Every async continuation — the enemy-fetch loop, a challenge/retreat
+	 *  transition, and each battle-end resolution (idle victory, boss victory/loss) — captures this value where
+	 *  it begins and re-checks it via {@link isCurrent} after each await before touching shared state. A
+	 *  continuation whose captured value no longer matches has been superseded — by another operation, or by a
+	 *  stop()→start() cycle it spanned (#2331) — and abandons instead of resuming into whatever is current now. */
+	#generation = 0;
 	/** Re-entrancy guard for getNewEnemy: overlapping stage handlers must not both spawn-and-notify an
 	 *  enemy (a double-spawn the backend replay would flag as cheating). */
 	#fetchingEnemy = false;
-	/** Bumped whenever the active enemy-fetch loop is superseded (a stop, or a transition taking over).
-	 *  The running getNewEnemy captures this value and exits once it no longer matches, so a stale loop
-	 *  under a sustained outage can't keep spinning — nor later clobber the new fight with an enemy
-	 *  nobody is waiting for. */
-	#fetchGeneration = 0;
 	/** The generation the in-flight fetch loop captured at its start. Lets a fresh getNewEnemy tell a
 	 *  genuine re-entrant duplicate (same generation — the running loop will still spawn, so drop) apart
 	 *  from a superseded loop (older generation — it will abandon without spawning, so wait it out then
@@ -128,6 +128,9 @@ export class EnemyManager {
 	public start(activeBattle?: IEnemyInstance) {
 		if (!this.started) {
 			this.started = true;
+			// A fresh session: any continuation still holding an older value is a zombie from before this
+			// start() (or before a preceding stop()) and must abandon rather than resume into it (#2331).
+			this.#generation++;
 			this.#isFirstFetch = true;
 			this.#battleStageUnhook = onBattleStageChanged((stage) => this.watchBattleStage(stage));
 			if (activeBattle) {
@@ -143,9 +146,14 @@ export class EnemyManager {
 	public stop() {
 		if (this.started) {
 			this.started = false;
+			// Force-clear rather than rely on an in-flight transition's own `finally`: that finally checks
+			// its captured generation against the bump `interruptFetch()` below makes, which — after this
+			// stop() — it will never match again, so it would otherwise leave the guard stuck.
+			this.#transitioning = false;
 			this.#battleStageUnhook?.();
 			// Cut short an in-flight retry backoff so a getNewEnemy parked under a sustained outage exits
-			// at once on the cleared `started` flag rather than after the full backoff window.
+			// at once on the cleared `started` flag rather than after the full backoff window; also bumps
+			// the generation so any continuation parked mid-await recognizes this teardown (#2331).
 			this.interruptFetch();
 			// Teardown (screen unmount / session end), not a user intent change, so don't sync the persisted
 			// loop mode — clobbering it to idle here would lose a disconnecting boss-farmer's mode.
@@ -210,7 +218,7 @@ export class EnemyManager {
 	 * (#1690), mirroring the unconditional discard `ChallengeBoss` already applies server-side.
 	 */
 	public async getNewEnemy(prepared?: PreparedBattle, forceAbandon = false): Promise<void> {
-		const generation = this.#fetchGeneration;
+		const generation = this.#generation;
 		// Looped (not a one-shot if): a fetch already running when we arrive is checked again after each
 		// wait. Two same-generation callers can both queue behind an older, superseded fetch — the first
 		// to resume starts a fresh fetch (synchronously re-arming fetchingEnemy/fetchingGeneration before
@@ -229,7 +237,7 @@ export class EnemyManager {
 			// down, then loop back and re-check rather than assuming the guard is now clear.
 			await this.#inFlightFetch;
 			// The wait can overlap a stop or a further supersession; bail if either landed meanwhile.
-			if (!this.started || generation !== this.#fetchGeneration) {
+			if (!this.isCurrent(generation)) {
 				return;
 			}
 		}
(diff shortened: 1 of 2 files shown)


UI/src/lib/engine/battle/enemy-manager.ts
```diff
@@ -143,9 +146,14 @@ export class EnemyManager {
 	public stop() {
 		if (this.started) {
 			this.started = false;
+			// Force-clear rather than rely on an in-flight transition's own `finally`: that finally checks
+			// its captured generation against the bump `interruptFetch()` below makes, which — after this
+			// stop() — it will never match again, so it would otherwise leave the guard stuck.
+			this.#transitioning = false;
```
ginderjeremiah, line 154: **Blocking — merging the two counters orphans `#transitioning` on a Home entry, and `enterHome()` wasn't given the force-clear that `stop()` got here.**

You correctly identified that once `endTransition` compares the captured generation against the shared `#generation`, any `interruptFetch()` that *isn't* a `beginTransition()` — but fires during an in-flight transition — will bump `#generation` past the transition's captured value, so its `finally → endTransition()` no longer matches and never releases `#transitioning`. That's exactly why this force-clear was added to `stop()`.

But `stop()` isn't the only non-transition caller of `interruptFetch()`: `enterHome()` (line 197) calls it too, and it got no equivalent treatment. This is reachable through normal play:

1. Player is in a boss fight and presses **Retreat**. `retreatFromBoss()` → `beginTransition()` sets `#transitioning = true` and returns generation `G`, then `returnToIdle()` sets `mode = 'idle'` **synchronously before** `await getNewEnemy(undefined, true)`.
2. Because `mode` is now `'idle'`, `ZoneNav`'s `bossEngaged` is false, so **zone nav is enabled** — unlike the challenge path, which sets `mode = 'boss'` before its await and hard-disables nav.
3. During the retreat's in-flight `NewEnemy` round-trip, the player navigates to **Home**. `enterHome()` → `interruptFetch()` bumps `#generation` to `G+1` (and does *not* touch `#transitioning`).
4. The retreat's `getNewEnemy` re-checks `isCurrent(G)` → false → abandons (good), then its `finally` runs `endTransition(G)`: `G === #generation (G+1)` is false → **`#transitioning` stays stuck `true`**.

At Home this is harmless (battle is resting), but on leaving Home, `navigateToZone` → `getNewEnemy()` spawns a fresh enemy and `watchBattleStage` now early-returns on `if (this.#transitioning) return;` for every stage change — so the idle loop silently stalls (no claim, no next enemy) until the player challenges a boss or leaves `/game`.

Before this PR, `endTransition` compared against the independent `#transitionGeneration`, which `enterHome`'s `interruptFetch` never bumped, so it cleared correctly — this is a regression the merge introduces, in the same bug class #2331 is about.

Fix: give `enterHome()` the same force-clear `stop()` has (`this.#transitioning = false;` before/with the `interruptFetch()`), since a Home entry supersedes any in-flight transition just as a stop does. Worth a mirrored test (retreat superseded by a Home entry mid-`getNewEnemy`, asserting a subsequent idle victory still resolves).

---

ginderjeremiah (commented):
## Review: EnemyManager start-epoch to stop zombie battle-end continuations (#2331)

**Verdict: Request changes** — the core `stop()→start()` fix is right and cleanly consolidates the guards per #1736, but unifying the two generation counters introduces a new regression on the `enterHome()` path that isn't covered by the added tests (details inline at `enemy-manager.ts:152`).

_(Posting as a comment since I'm the PR author and GitHub won't accept a gating verdict on one's own PR — the intended verdict is Request changes.)_

### Blocking

- **`#transitioning` orphaned by a Home entry mid-transition** — `enemy-manager.ts:152` (inline). `stop()` got a force-clear of `#transitioning` because a bumped `#generation` now decouples `endTransition` from the transition's captured value; `enterHome()` calls the same `interruptFetch()` but got no equivalent. A Retreat (which sets `mode='idle'` before its await, so zone nav stays enabled) superseded by a Home entry mid-`getNewEnemy` leaves `#transitioning` stuck `true`, silently stalling the idle loop after the player leaves Home. This is a regression the counter-merge introduces, in the exact bug class #2331 targets.

### What's good

- The consolidation is the right call and genuinely *reduces* the guard surface — one `#generation` epoch plus a single `isCurrent()` re-check replacing the `#fetchGeneration`/`#transitionGeneration` pair, the `bossLoopActive` getter, and `transitionCurrent`. No dead references left behind (verified).
- Capturing the generation once at `watchBattleStage` entry and threading it into the resolutions is exactly the shape #2331 suggested, and it fixes the real defect: the resolutions now re-check a session epoch instead of live `mode`/`#transitioning` that a restart resets to "valid".
- Correctly scoped as a frontend-only client-concurrency fix — no `battleStep` parity surface touched, so no backend mirror needed.
- The dense comments are appropriate for code this subtle; I wouldn't trim them.

### Test coverage

- The three added tests are well-constructed and target the right scenarios (idle-victory and boss-loss spanning `stop()→start()`, plus the in-session Home-mid-claim variant), each with a gated promise to genuinely park the continuation across the cycle.
- Gap: no test exercises the **retreat superseded by a Home entry** path — which is where the blocking regression lives. Add one asserting a subsequent idle victory still resolves (i.e. `#transitioning` was released) after a retreat is superseded by entering Home mid-`getNewEnemy`.

---

UI/src/lib/engine/battle/enemy-manager.ts
ginderjeremiah: Confirmed — traced through the same sequence and it's a real regression from the counter merge. Fixed in 785872c: `enterHome()` now force-clears `#transitioning` before `interruptFetch()`, mirroring `stop()`, with the doc comment on the field updated to explain both force-clear sites.

Also added the suggested regression test: a retreat superseded by a Home entry mid-`getNewEnemy`, then leaving Home and firing a Victorious stage to confirm `DefeatEnemy` still sends (i.e. `#transitioning` was released rather than wedging `watchBattleStage`'s guard). Verified it fails without the fix and passes with it. Full suite (3936 tests) + lint are green.

---

ginderjeremiah:
## Re-review (post-fix): Approve

The one blocking finding from the prior review — the counter merge orphaning `#transitioning` when a Home entry supersedes an in-flight retreat, because `enterHome()` didn't get the force-clear `stop()` did — is fully resolved in 785872c. I independently re-verified the source rather than relying on the self-confirmation:

- **The fix is correct.** `enterHome()` now force-clears `#transitioning = false` before `interruptFetch()`, mirroring `stop()`. Traced the reachable path: a retreat sets `mode = 'idle'` before its `getNewEnemy` await (so zone nav, including Home, stays enabled), the Home entry bumps `#generation` via `interruptFetch()`, and the retreat's `finally → endTransition(G)` no longer matches the bumped generation — so without the force-clear the guard stayed stuck and wedged `watchBattleStage`'s `if (this.#transitioning) return;` for every future stage change. Force-clearing on Home entry is semantically right: a Home entry supersedes any in-flight transition just as a stop does.
- **Test coverage closes the gap.** `enemy-manager-boss.test.ts` now includes `releases the transitioning guard when a retreat is superseded by a Home entry mid-fetch (#2331 regression)`, which leaves Home and fires a victory to confirm the guard was released — the exact blocking scenario — alongside the original stop()→start() and in-session Home-mid-claim tests.
- The consolidated `#generation` epoch genuinely reduces the guard surface (one epoch + `isCurrent()` replacing the two counters, the `bossLoopActive` getter, and `transitionCurrent`), which is what #1736 asked for, and the field-level doc comment now documents both force-clear sites.

CI is green (frontend unit + all three e2e browsers), and no nits or should-fixes remain. Merging.

---

PR #2349, opened 2026-07-23
Invalidate in-flight fetches when a push mutates player-progress stores

### LINKED ISSUE #2332: Player-progress stores: a live push applied while a fetch is in flight is clobbered by the older fetch result
## Problem

All three push-updated stores assign the fetch result wholesale, guarded only against `reset()` via the loader epoch — not against push/optimistic writes that landed *after the request was issued*:

- `UI/src/stores/challenges.svelte.ts:26-45` — `challenges = result` after `loader.isStale(epoch)`; `markCompleted` (:64) is called from the engine's `ChallengeCompleted` push (`$lib/engine/engine.ts:164`) and does not bump the epoch.
- `UI/src/stores/proficiencies.svelte.ts:17-36` — same pattern; `applyXpGained` (:90) fires on **every battle victory** (`engine.ts:229`).
- `UI/src/stores/statistics.svelte.ts:15-34` — same; `markZoneCleared` (:63) from `enemy-manager.ts:654`.

## Failure scenario

The Challenges or Codex screen mounts and issues `load(force)` while idle battles keep completing. A `ChallengeCompleted` push arrives mid-flight and `markCompleted` flips the completion (unlocking a zone in `ZoneNav` via `isChallengeCompleted`); the fetch response — computed before the completion was persisted (the command handler's read and its response send are separated by awaits, so this interleaves even on one server) — then resolves and replaces the array, **reverting the completion**. Unlike proficiency XP (self-corrected by the next push seconds later), nothing re-corrects a lost challenge completion: the just-unlocked zone re-locks and the completed card reverts until the next forced load. The proficiency variant additionally feeds `battleModifiers` into the live battler, so a clobber can transiently regress mid-battle attributes at a level-up boundary.

The stores' comments only cover the opposite direction ("a later `load(force)` reconciles"); the fetch-overwrites-newer-push direction is unhandled.

## Suggested fix

Bump the loader epoch (or set a "dirty since request issued" flag) in `markCompleted` / `applyXpGained` / `markZoneCleared` so an in-flight fetch result is discarded (or merged) instead of assigned wholesale.

ginderjeremiah:
## Summary

`challenges.svelte.ts`, `proficiencies.svelte.ts`, and `statistics.svelte.ts` all assign a fetch's result wholesale, guarded only against `reset()` via the `CoalescedLoader` epoch — not against a push (`markCompleted`, `applyXpGained`, `markZoneCleared`) landing *after* the request was issued but before it resolved.

The command handler's read and its response send are separated by awaits, so this interleaves even on a single server: the fetch response is computed *before* the push is persisted server-side. If a `ChallengeCompleted`/`ProficiencyXpGained`/zone-clear push lands mid-flight, the in-flight fetch's stale response then resolves and reverts the just-applied optimistic update — a just-unlocked zone re-locks, a completed challenge card reverts, or mid-battle attribute bonuses transiently regress at a level-up boundary.

- Added `CoalescedLoader.invalidate()` — bumps the loader's epoch (reusing the existing `isStale()` check the fetchers already gate their writes on) **without** tearing down in-flight/queued fetch state the way `reset()` does. `reset()` is a full session teardown (clears data, abandons in-flight/queued fetches); `invalidate()` only needs to mark an already-in-flight response as stale so it gets discarded instead of overwriting.
- Each store's push handler (`markCompleted`, `applyXpGained`, `markZoneCleared`) now calls `loader.invalidate()` right after it mutates local state, so a fetch already in flight at that moment resolves into the existing `isStale` early-return instead of clobbering the push. A later `load(force)` still reconciles normally against the authoritative server state.

## Test plan

- [x] Added `CoalescedLoader` unit tests: `invalidate()` bumps the epoch the same way `isStale` recognizes for `reset()`, and — unlike `reset()` — leaves in-flight/queued fetch state untouched (a concurrent `load()` still coalesces onto the existing in-flight fetch).
- [x] Added a regression test per store (`challenges.test.ts`, `proficiencies.svelte.test.ts`, `statistics.test.ts`) reproducing the exact clobber: a fetch is left in flight, the push lands, then the stale fetch resolves with pre-push data — asserting the push's optimistic state survives.
- [x] `npm run lint` (ESLint + svelte-check) — 0 errors/warnings.
- [x] `npm run test:unit:ci` — all 311 frontend test files / 3937 tests pass.
- [x] No backend files touched — this is a frontend-only client store fix, not a battle-logic parity change, so no backend test suite is affected.

Closes #2332

---
_Generated by [Claude Code](https://claude.ai/code/session_018eVgxsxt3EMvivRaqCbt3k)_

### DIFF
--- a/UI/src/lib/common/coalesced-loader.ts
+++ b/UI/src/lib/common/coalesced-loader.ts
@@ -3,7 +3,9 @@
  * honest: a forced load issued mid-flight queues one fresh fetch behind the stale in-flight one
  * (concurrent forces share it) rather than resolving with data requested before the call. The
  * player-progress stores rely on that guarantee for post-victory gate checks and for the
- * divergence-recovery reload after a failed server push.
+ * divergence-recovery reload after a failed server push. They also call {@link invalidate} when a
+ * push mutates local state while a fetch is in flight, so that fetch's response — computed before
+ * the push landed — is discarded instead of reverting it.
  *
  * Deliberately not re-exported from the `$lib/common` barrel: the stores consume it, and the barrel
  * transitively imports the stores (via enemy-attributes → battle-attributes), so a barrel import
@@ -60,14 +62,25 @@ export class CoalescedLoader {
 		this.epoch += 1;
 	}
 
-	/** Current reset epoch, bumped by {@link reset}. `fetchFn` should capture this before its first
-	 *  `await` and pass it to {@link isStale} afterward to detect a `reset()` that ran mid-flight. */
+	/** Bumps the epoch so an already-in-flight fetch's eventual response is recognized as stale by
+	 *  {@link isStale}, without discarding the in-flight/queued fetch state itself (unlike
+	 *  {@link reset}, which is a full session teardown). Call this when a push mutates the store's
+	 *  data directly: the fetch's response was computed before the push landed, so a later `load()`
+	 *  should not still get to overwrite the push with it. */
+	invalidate(): void {
+		this.epoch += 1;
+	}
+
+	/** Current epoch, bumped by {@link reset} and {@link invalidate}. `fetchFn` should capture this
+	 *  before its first `await` and pass it to {@link isStale} afterward to detect either running
+	 *  mid-flight. */
 	get currentEpoch(): number {
 		return this.epoch;
 	}
 
 	/** Whether `epoch` (captured from {@link currentEpoch} before an await) no longer matches — a
-	 *  `reset()` ran mid-flight, so a write gated on this epoch belongs to a discarded session. */
+	 *  `reset()` or `invalidate()` ran mid-flight, so a write gated on this epoch is stale (a
+	 *  discarded session, or a push whose mutation the write would otherwise revert). */
 	isStale(epoch: number): boolean {
 		return epoch !== this.epoch;
 	}
--- a/UI/src/lib/common/coalesced-loader.ts
+++ b/UI/src/lib/common/coalesced-loader.ts
@@ -63,12 +63,18 @@ export class CoalescedLoader {
 	}
 
 	/** Bumps the epoch so an already-in-flight fetch's eventual response is recognized as stale by
-	 *  {@link isStale}, without discarding the in-flight/queued fetch state itself (unlike
-	 *  {@link reset}, which is a full session teardown). Call this when a push mutates the store's
-	 *  data directly: the fetch's response was computed before the push landed, so a later `load()`
-	 *  should not still get to overwrite the push with it. */
+	 *  {@link isStale}, without discarding the in-flight fetch itself (unlike {@link reset}, which is
+	 *  a full session teardown). Call this when a push mutates the store's data directly: the fetch's
+	 *  response was computed before the push landed, so a later `load()` should not still get to
+	 *  overwrite the push with it.
+	 *
+	 *  Also clears `freshQueued`: a force already queued behind the in-flight fetch captured the
+	 *  pre-bump epoch as its `chainEpoch` (see {@link load}), so this bump makes that queued chain
+	 *  abort without ever clearing the flag itself — leaving it stuck and silently disabling the
+	 *  force-queue path for the rest of the session unless cleared here. */
 	invalidate(): void {
 		this.epoch += 1;
+		this.freshQueued = false;
 	}
 
 	/** Current epoch, bumped by {@link reset} and {@link invalidate}. `fetchFn` should capture this
--- a/UI/src/stores/challenges.svelte.ts
+++ b/UI/src/stores/challenges.svelte.ts
@@ -17,7 +17,8 @@ const fetchChallenges = async () => {
 		// fetchSocketData throws on a socket error, preserving this try/catch contract.
 		const result = (await fetchSocketData('GetPlayerChallenges')) ?? [];
 		if (loader.isStale(epoch)) {
-			// reset() ran while this fetch was in flight; this write belongs to a discarded session.
+			// Either reset() ran while this fetch was in flight (a discarded session), or markCompleted()
+			// applied a push mid-flight — this response predates that push and would revert it.
 			return;
 		}
 		challenges = result;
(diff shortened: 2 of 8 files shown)


UI/src/lib/common/coalesced-loader.ts
```diff
@@ -60,14 +62,25 @@ export class CoalescedLoader {
...
-	/** Current reset epoch, bumped by {@link reset}. `fetchFn` should capture this before its first
-	 *  `await` and pass it to {@link isStale} afterward to detect a `reset()` that ran mid-flight. */
+	/** Bumps the epoch so an already-in-flight fetch's eventual response is recognized as stale by
+	 *  {@link isStale}, without discarding the in-flight/queued fetch state itself (unlike
+	 *  {@link reset}, which is a full session teardown). Call this when a push mutates the store's
+	 *  data directly: the fetch's response was computed before the push landed, so a later `load()`
+	 *  should not still get to overwrite the push with it. */
+	invalidate(): void {
+		this.epoch += 1;
+	}
```
ginderjeremiah, line 78: **Blocking** — `invalidate()` doesn't clear `freshQueued`, so it can permanently disable the force-queue path for the session.

When a forced load is queued behind an in-flight fetch, `load()` (line 43-53) sets `freshQueued = true`, captures `chainEpoch = this.epoch`, and gates the queued fetch on `if (this.epoch !== chainEpoch) { return; }` (line 47). That early-return also skips the `this.freshQueued = false` on line 50.

`invalidate()` bumps `epoch`, which trips exactly that guard. But unlike `reset()` — which explicitly sets `freshQueued = false` — `invalidate()` leaves `freshQueued` stuck `true`. Once stuck, `if (force && !this.freshQueued)` can never fire again, so for the rest of the session a `load(force)` issued while a fetch is in flight silently coalesces onto that (older) in-flight fetch instead of queueing a fresh one behind it. That directly undermines the "force honesty" guarantee this file's header documents, and can make the divergence-recovery reload silently no-op.

**Repro:** `load()` in flight → `load(true)` queues a fresh fetch (`freshQueued = true`) → a battle victory fires `applyXpGained`/`markCompleted`/`markZoneCleared` → `invalidate()` bumps the epoch → the queued `.then` early-returns without clearing `freshQueued` → it's stuck `true` thereafter.

**Fix:** have `invalidate()` also reset the force-queue bookkeeping, e.g.:

```ts
invalidate(): void {
	this.epoch += 1;
	this.freshQueued = false;
}
```

It's called synchronously from a push in the same session with `inFlight` untouched, so clearing `freshQueued` here is safe: the already-queued `.then` still epoch-mismatches and won't fetch, and a subsequent `load(force)` can now correctly queue a genuinely-fresh reconcile behind the in-flight fetch.

---

ginderjeremiah (commented):
## Review: Invalidate in-flight fetches when a push mutates player-progress stores (#2332)

**Verdict: Request changes** — the core approach is correct and matches the issue's suggested fix, but reusing the loader epoch for `invalidate()` introduces a regression in the sibling `force`-queue path that isn't guarded or tested (details inline at `coalesced-loader.ts:70`).

_(Posting as a comment since I'm the PR author and GitHub won't accept a gating verdict on one's own PR — the intended verdict is Request changes.)_

### Blocking

- **`invalidate()` leaves `freshQueued` stuck `true`** — `coalesced-loader.ts:70` (inline). When a forced load is queued behind an in-flight fetch, `invalidate()` bumps the epoch and trips the queued fetch's `chainEpoch` guard, which early-returns without resetting `freshQueued` (line 50 is skipped). Unlike `reset()`, `invalidate()` never clears it, so after a push races a queued force the queue-a-fresh-fetch-behind-in-flight optimization is disabled for the rest of the session — `load(force)` issued during an in-flight fetch silently coalesces onto the older fetch, weakening the "force honesty" guarantee this file documents and potentially making the divergence-recovery reload no-op. Cheap to fix in this PR (clear `freshQueued` in `invalidate()`); see the inline comment.

### Test coverage

- The added tests are well-targeted: the `CoalescedLoader` unit tests prove `invalidate()` marks in-flight results stale while leaving `load()` coalescing intact, and each store has a clean regression test reproducing the exact clobber (in-flight fetch resolves with pre-push data, push must survive). These correctly cover the reported bug.
- **Gap:** no test exercises `invalidate()` racing a *queued force* — which is where the blocking regression lives. Add a `CoalescedLoader` test: `load()` in flight, `load(true)` queues a fresh fetch, `invalidate()`, then assert a subsequent `load(true)` during a new in-flight fetch still queues a fresh fetch behind it (i.e. `freshQueued` was released).
- Correctly scoped as frontend-only — no battle-parity surface touched, so no backend mirror is needed.

### What's good

- Right root-cause fix: bumping the epoch on push so the existing `isStale()` gate discards the pre-push fetch response, exactly as the issue suggested — no new mechanism, reusing the check the fetchers already gate on.
- Clean separation of `invalidate()` (mark stale, keep fetch state) from `reset()` (full teardown), with the distinction clearly documented.
- `invalidate()` is placed correctly after each push's local mutation, and the already-completed/already-cleared early-returns correctly skip it (no mutation → nothing to protect).
- Doc comments are appropriately scoped for code this subtle — not over-detailed.

---

UI/src/lib/common/coalesced-loader.ts
ginderjeremiah: Confirmed — traced through the exact sequence and reproduced it: with `freshQueued` stuck `true`, a `load(force)` issued while a later fetch is in flight just `await`s that in-flight fetch directly instead of queuing a genuinely fresh one behind it.

Fixed in 96868a4: `invalidate()` now also clears `freshQueued`. Added a `CoalescedLoader` test reproducing the race (queued force → push invalidates → fetch settles/aborts → a later force must still queue fresh) — verified it fails on the pre-fix code and passes with the fix. Full suite (311 files / 3938 tests) and lint are green.

---

ginderjeremiah:
## Re-review (post-fix): Approve

The one blocking finding from the prior review — `invalidate()` leaving `freshQueued` stranded `true` and silently disabling the force-queue path for the session — is fully resolved in 96868a4. I independently re-verified rather than relying on the self-confirmation:

- **The fix is correct.** `invalidate()` now bumps the epoch *and* clears `freshQueued`. Traced the race end-to-end: a force queued behind the in-flight fetch captured the pre-bump epoch as `chainEpoch`, so the bump trips its `epoch !== chainEpoch` early-return (skipping the flag reset in `load()`); clearing `freshQueued` in `invalidate()` is what lets a subsequent `load(force)` queue a genuinely fresh reconcile instead of coalescing onto the older fetch. Since `invalidate()` runs synchronously from a push with `inFlight` untouched, clearing the flag here is safe — the already-queued `.then` still epoch-mismatches and won't fetch.
- **Test coverage closes the gap.** `coalesced-loader.test.ts` now includes `invalidate() racing a queued force does not strand freshQueued, so a later force still queues fresh` — the exact blocking scenario — alongside the original `invalidate()` staleness/coalescing tests and the per-store clobber regressions.
- The doc comment on `invalidate()` now explains *why* it clears `freshQueued`, which is the right amount of context for code this subtle.

CI is green (frontend unit + all three e2e browsers), and no nits or should-fixes remain. Merging.

---

PR #2351, opened 2026-07-23
Fix SaveDeviceInfo clearing previously-captured DeviceMemory/HardwareConcurrency

### LINKED ISSUE #2342: UserLogins.SaveDeviceInfo clears previously-captured DeviceMemory/HardwareConcurrency with nulls, contradicting its own fill-when-missing policy
## Problem

`Game.DataAccess/Repositories/UserLogins.cs:144-148`: the client-hint fields use `??=` per the documented policy ("client hints are only filled in when missing so a later request with fewer hints never clears values an earlier one captured", `GetOrCreateBrowserInfo` doc :172-176) — but `DeviceMemory` and `HardwareConcurrency` are assigned unconditionally two lines later. A later `SaveDeviceInfo` call carrying nulls (`navigator.deviceMemory` is unavailable in Firefox/Safari; privacy extensions strip it) nulls out values an earlier request captured for the same device.

## Suggested fix

If fill-when-missing is intended (consistent with the stated policy): null-guarded assignment/`??=` for these two fields. If last-write-wins is intended, a one-line comment saying so settles it.

ginderjeremiah:
## Summary

`UserLogins.SaveDeviceInfo` (`Game.DataAccess/Repositories/UserLogins.cs`) is supposed to follow a fill-when-missing policy for client-hint/device fields — "client hints are only filled in when missing so a later request with fewer hints never clears values an earlier one captured" (per the doc comment on the sibling `GetOrCreateBrowserInfo`). The `SecChUa*` fields already honor this via `??=`, but `DeviceMemory`/`HardwareConcurrency` were assigned unconditionally two lines below.

`navigator.deviceMemory` is unavailable in Firefox/Safari and can be stripped by privacy extensions, so a later `SaveDeviceInfo` call carrying nulls for a device an earlier request already enriched would null out the previously-captured values — contradicting the documented policy.

## Fix

Changed the two unconditional assignments to `??=`, matching the pattern already used for `SecChUa`/`SecChUaMobile`/`SecChUaPlatform` two lines above.

## Test plan

- [x] Added `SaveDeviceInfo_LaterCallWithoutHints_DoesNotClearPreviouslyCapturedValues` to `UserLoginsCapAndOwnershipIntegrationTests.cs`, pinning that a first call with values followed by a second call with nulls preserves the original values.
- [x] `dotnet test` (Game.Application.Tests) — 1096 passed, 0 failed, including the existing `SaveDeviceInfo`/`UserLogins` integration tests (which already exercised same-value overwrites and remain green).

Closes #2342

---
_Generated by [Claude Code](https://claude.ai/code/session_01UGH7gokMhWmWScsVUiBJ2S)_

### DIFF
--- a/Game.Application.Tests/DataAccess/UserLoginsCapAndOwnershipIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/UserLoginsCapAndOwnershipIntegrationTests.cs
@@ -212,6 +212,34 @@ namespace Game.Application.Tests.DataAccess
             Assert.Equal(8, device.HardwareConcurrency);
         }
 
+        [Fact]
+        public async Task SaveDeviceInfo_LaterCallWithoutHints_DoesNotClearPreviouslyCapturedValues()
+        {
+            // navigator.deviceMemory/hardwareConcurrency are unsupported in some browsers and stripped by some
+            // privacy extensions, so a later request can legitimately carry nulls for a device an earlier
+            // request already enriched (#2342) — those earlier values must survive, not be overwritten.
+            int userId;
+            using (var seedScope = CreateScope())
+            {
+                var context = seedScope.ServiceProvider.GetRequiredService<GameContext>();
+                userId = (await TestDataSeeder.CreateUserAsync(context)).Id;
+            }
+
+            using (var scope = CreateScope())
+            {
+                var repo = scope.ServiceProvider.GetRequiredService<IUserLogins>();
+                await repo.RecordConnection(userId, Ip, Fingerprint(0), UserAgent, null, null, null, CancellationToken);
+                await repo.SaveDeviceInfo(userId, Fingerprint(0), null, null, null, 16, 8, CancellationToken);
+                await repo.SaveDeviceInfo(userId, Fingerprint(0), null, null, null, null, null, CancellationToken);
+            }
+
+            using var assertScope = CreateScope();
+            var assertContext = assertScope.ServiceProvider.GetRequiredService<GameContext>();
+            var device = await assertContext.Devices.SingleAsync(d => d.DeviceFingerprintHash == Fingerprint(0), CancellationToken);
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: Fix `SaveDeviceInfo` clearing previously-captured `DeviceMemory`/`HardwareConcurrency` (#2342)

**Verdict:** Approve — the fix is correct, minimal, and consistent with the documented policy. _(Posting as a comment since I authored the PR; the intended verdict is Approve.)_

The two unconditional assignments now use `??=`, matching the `SecChUa*` fields immediately above and the fill-when-missing policy documented on `GetOrCreateBrowserInfo` (`UserLogins.cs:172-176`): "client hints are only filled in when missing so a later request with fewer hints never clears values an earlier one captured." This is exactly the right operator for that intent — a later request carrying nulls (Firefox/Safari lack `navigator.deviceMemory`; privacy extensions strip it) can no longer null out values an earlier request enriched. The bug was a real contradiction of the sibling fields' behavior, and this closes it in the smallest possible way.

### Test coverage

- `SaveDeviceInfo_LaterCallWithoutHints_DoesNotClearPreviouslyCapturedValues` pins the fix directly: enrich with `(16, 8)`, then a second call with `(null, null)`, and assert the original values survive. This would fail on the old unconditional assignment and pass now.
- Integration test is the correct level here — `UserLogins` is a data-access class whose behavior is the DB interaction itself (no domain logic to unit-test), matching `backend.md`'s testing guidance. The existing `SaveDeviceInfo_OwnDevice_EnrichesIt` still covers the first-write enrichment path, and the well-placed comment explains why the fingerprint helper stays a simple string.

### What's good

- Correct, targeted one-line-per-field fix with no collateral changes and no dead code.
- The added test's comment ties the scenario back to the real-world browser behavior that motivates the policy, so a future reader understands why nulls must not overwrite.

---

PR #2352, opened 2026-07-23
Disarm background-throttle resume credit on the next regular tick

### LINKED ISSUE #2343: BackgroundThrottleMonitor's one-shot resume credit never expires — a later foreground stall is misreported as background throttling
## Problem

`UI/src/lib/engine/background-throttle-notice.ts:78-101`: `resumeCatchUpPending` is armed on **every** visibility→visible transition (:99) and consumed only by the *next* clipped catch-up poll — whenever that happens. The premise is that a frozen tab's single catch-up poll fires just after the resume `visibilitychange`, but nothing disarms the flag once that ~10ms window passes. With the worker tick source a hidden tab normally loses nothing, so after any hide/show cycle the flag stays armed indefinitely.

## Failure scenario

Hours after an innocuous tab switch, a ≥60s **foreground** stall — OS suspend/wake with the tab focused, a long debugger pause — is credited to a "hidden period", crosses `LOST_TIME_NOTICE_THRESHOLD_MS`, and permanently burns the one-shot notice on misleading "add this site to Always keep active" guidance that cannot help that situation.

## Suggested fix

Disarm the flag after the first post-resume poll regardless of whether it lost time, or bound the credit to a short wall-clock window after the visibility transition.

ginderjeremiah:
## Summary

`BackgroundThrottleMonitor.resumeCatchUpPending` (`UI/src/lib/engine/background-throttle-notice.ts`) was armed on every visibility resume and only ever consumed by the *next* idle-time-lost event — whenever that happened. If the tab wasn't actually throttled (no catch-up poll followed the resume), the credit lingered indefinitely. Hours later, a genuine **foreground** stall (OS suspend/wake with the tab focused, a long debugger pause) would then be misattributed to the earlier hidden period, potentially crossing the notice threshold and permanently burning the one-shot "add this site to Always keep active" guidance on a situation it can't help with.

## Fix

Bound the resume credit to the very next logical tick instead of letting it wait indefinitely for the next lost-time event:

- On resume, subscribe (one-shot) to `onLogicalUpdate` in addition to arming `resumeCatchUpPending`.
- If a real catch-up poll follows, `recordLostTime` still consumes the credit first — `onIdleTimeLost` fires before `onLogicalUpdate` within the same engine `update()` pass, so the frozen-tab crediting behavior (Memory Saver / tab discard) is unchanged.
- If instead the very next regular tick passes with nothing lost, the guard disarms the credit there, so it can no longer be misapplied to an unrelated stall much later.
- The guard is also cleared on re-arm (repeated hide/show cycles don't stack subscriptions), when entering a new hidden period, and on `stop()`.

Closes #2343.

## Test plan

- [x] `npm run test:unit:ci` — full suite green
- [x] Added unit tests covering: the credit disarming on a no-loss regular tick after resume, stale-guard cleanup across repeated hide/show cycles, and guard unsubscription on `stop()`. Existing tests (frozen-tab crediting, foreground-stall ignoring, one-time notice) still pass unchanged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WzRRQCGJJequWNTSYHfbFL)_

### DIFF
--- a/UI/src/tests/lib/engine/background-throttle-notice.test.ts
+++ b/UI/src/tests/lib/engine/background-throttle-notice.test.ts
@@ -4,13 +4,15 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 const h = vi.hoisted(() => ({
 	toastWarning: vi.fn(),
 	unhook: vi.fn(),
-	unhookLogicalUpdate: vi.fn(),
 	state: {
 		lsStore: {} as Record<string, string>,
 		lsAvailable: true,
 		setItemThrows: false,
 		idleTimeLostCb: undefined as ((ms: number) => void) | undefined,
-		logicalUpdateCb: undefined as ((tickMs: number, unhook: () => void) => void) | undefined
+		// Every onLogicalUpdate() call gets its own {cb, unhook} entry (mirroring the real hook's
+		// per-subscriber tracker) so a test can tell a genuinely-unhooked stale subscription apart
+		// from one that was merely overwritten by a later subscribe.
+		logicalUpdateSubs: [] as { cb: (tickMs: number, unhook: () => void) => void; unhook: () => void }[]
 	}
 }));
 
@@ -39,8 +41,9 @@ vi.mock('$lib/engine/logical-engine', () => ({
 		return h.unhook;
 	},
 	onLogicalUpdate: (cb: (tickMs: number, unhook: () => void) => void) => {
-		h.state.logicalUpdateCb = cb;
-		return h.unhookLogicalUpdate;
+		const unhook = vi.fn<() => void>();
+		h.state.logicalUpdateSubs.push({ cb, unhook });
+		return unhook;
 	}
 }));
 
@@ -65,7 +68,7 @@ describe('BackgroundThrottleMonitor', () => {
 		h.state.lsAvailable = true;
 		h.state.setItemThrows = false;
 		h.state.idleTimeLostCb = undefined;
-		h.state.logicalUpdateCb = undefined;
+		h.state.logicalUpdateSubs = [];
 		hidden = false;
 		Object.defineProperty(document, 'hidden', { configurable: true, get: () => hidden });
 		monitor = new BackgroundThrottleMonitor();
@@ -80,9 +83,14 @@ describe('BackgroundThrottleMonitor', () => {
 	});
 
 	const lose = (ms: number) => h.state.idleTimeLostCb?.(ms);
+	const latestGuardSub = () => h.state.logicalUpdateSubs.at(-1);
 	// Simulates a regular logical tick that lost nothing — the common case, since the worker tick
-	// source polls every 10ms regardless of visibility.
-	const tick = () => h.state.logicalUpdateCb?.(10, h.unhookLogicalUpdate);
+	// source polls every 10ms regardless of visibility. Only the current (last-subscribed) guard is
+	// fired — a real, correctly-unhooked stale guard would never receive it.
+	const tick = () => {
+		const sub = latestGuardSub();
+		sub?.cb(10, sub.unhook);
+	};
 
 	it('shows a one-time dismissible notice after a hidden period loses past the threshold', () => {
 		monitor.start();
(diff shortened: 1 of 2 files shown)


UI/src/tests/lib/engine/background-throttle-notice.test.ts
```diff
@@ -112,6 +123,36 @@ describe('BackgroundThrottleMonitor', () => {
...
+	it('does not let a stale resume-guard subscription from an earlier cycle disarm a fresh one', () => {
+		monitor.start();
+		goHidden();
+		goVisible(); // arms guard #1
+		goHidden();
+		goVisible(); // re-arms: guard #1 must be unhooked, not left to fire later
+		lose(70_000); // credited to guard #2's hidden period
+
+		expect(h.toastWarning).toHaveBeenCalledTimes(1);
+	});
```
ginderjeremiah: Nit (test fidelity): this test doesn't actually reproduce a *stale subscription firing*, so it can't distinguish the `clearResumeGuard()`-on-re-arm from its absence. The mock keeps a single `logicalUpdateCb` slot (guard #2 overwrites guard #1) and a shared `h.unhookLogicalUpdate`, and no `tick()` is fired here — so it would pass even if `armResumeGuard` skipped the re-arm cleanup. As written it just asserts the notice still shows after a double hide/show + `lose()`, which is a fine regression check but not the "stale guard doesn't disarm a fresh one" scenario the name promises.

To genuinely exercise it, the mock would need to retain each subscription with its own unhook, then fire the *stale* callback after re-arming and assert `resumeCatchUpPending` survives. That's arguably more than this structurally-single-guard design warrants — the idempotent unhook + `clearResumeGuard()` on every transition already makes a stale guard unreachable — so either strengthen the harness or rename this to reflect that it's a plain double-cycle regression test. Not blocking.

---

ginderjeremiah (commented):
## Review: Disarm the background-throttle resume credit on the next regular tick (#2343)

**Verdict:** Approve _(posted as a comment — GitHub won't accept a gating verdict on my own PR)_ — the fix correctly and minimally closes the bug, with matching test coverage; one non-blocking test-fidelity nit.

### Assessment

The diagnosis in #2343 is correct and I verified the fix against the engine internals:

- **Ordering premise holds.** `LogicalEngine.update()` calls `notifyIdleTimeLost(...)` *before* the `notifyLogicalUpdate(...)` tick loop within a single over-budget pass (`logical-engine.ts:65` vs `:73`). So on a real catch-up poll, `recordLostTime` consumes the credit and calls `clearResumeGuard()` before the resume guard's `onLogicalUpdate` would fire — the frozen-tab (Memory Saver) crediting path is genuinely unchanged.
- **Timing premise holds.** The tick worker polls every 10ms regardless of visibility (`tick-source.ts:1-6`), so "the very next logical tick" reliably arrives right after resume to disarm a credit that no catch-up poll claimed.
- **Guard lifecycle is complete.** The one-shot credit is cleared on consumption, on a no-loss tick, on re-arm, on entering a new hidden period, and on `stop()`. `armResumeGuard` clears before re-subscribing so hide/show cycles don't stack subscriptions, and the hook's unhook is idempotent (`hooks.ts:60-64`), so a stale guard is structurally unreachable.

Clean, well-scoped, well-commented change that solves the reported problem without over-engineering. Frontend-only lib module — no battle-parity or backend mirror concerns.

### Nits

- `background-throttle-notice.test.ts:136` — the "does not let a stale resume-guard subscription… disarm a fresh one" test doesn't actually reproduce a stale subscription firing (single-slot mock, no `tick()`), so it can't distinguish the re-arm cleanup being present. Inline comment has detail. Not blocking.

### Test coverage

Good. New tests cover the credit disarming on a no-loss post-resume tick, cleanup across repeated hide/show cycles, and guard unsubscription on `stop()`; existing frozen-tab crediting / foreground-stall / one-time-notice tests remain and still pass. Meets the frontend testing rule (lib module has unit tests mirroring the source path). The one gap is the fidelity of the stale-guard test noted above.

### What's good

Correctly identifies that `onIdleTimeLost` fires before `onLogicalUpdate` in the same pass and leans on that to keep the frozen-tab path intact — the subtle part of this fix — and documents the reasoning inline where a future edit might otherwise "simplify" it away.

---

UI/src/tests/lib/engine/background-throttle-notice.test.ts
ginderjeremiah: Good catch — strengthened rather than renamed. The mock now tracks each `onLogicalUpdate()` subscription with its own unhook spy, and the test asserts guard #1's unhook was actually called (not just superseded) before guard #2 arms. Verified it fails if `armResumeGuard`'s `clearResumeGuard()` call is removed. Pushed in 66f834c.

---

ginderjeremiah:
## Final review — merging

Re-review of the only change since my earlier pass (commit `66f834c`, which strengthened the stale-guard test). Confirmed:

- The test-fidelity nit is fully resolved. The mock now tracks each `onLogicalUpdate()` subscription with its own unhook spy (mirroring the real hook's per-subscriber tracker), and the test asserts guard #1 was **actually unhooked** — not merely superseded — before guard #2 arms (`background-throttle-notice.test.ts:145-155`). The commit notes this was verified to fail if `armResumeGuard`'s `clearResumeGuard()` call is removed, so it genuinely exercises the re-arm cleanup path.
- Core fix is unchanged and correct (validated in my prior review against the engine ordering: `onIdleTimeLost` fires before `onLogicalUpdate` in the same `update()` pass, so the frozen-tab crediting path is preserved).
- CI green on the current head (`all-checks-passed` ✓; `verify-frontend`, `test-frontend`, and all three e2e browsers pass; backend jobs correctly skipped for a frontend-only change).

**Verdict: approve.** No outstanding nits or should-fixes; the one review thread is resolved. Merging.

---

PR #2353, opened 2026-07-23
Skip boss-victory celebration on a definitive DefeatEnemy rejection

### LINKED ISSUE #2344: resolveBossVictory applies its celebration side effects even when the DefeatEnemy claim was definitively rejected
## Problem

`claimVictory()` (`UI/src/lib/engine/battle/enemy-manager.ts:741-775`) surfaces a failed/rejected `DefeatEnemy` only via a debug log and resync; its return carries no success signal, and `resolveBossVictory` (:639-687) ignores the result entirely:

```ts
await this.claimVictory();
...
statistics.markZoneCleared(clearedZoneId);
```

On a *definitive server rejection* (not just a lost response), the client still: marks the zone cleared (the "Cleared" seal shows wrong until the next `statistics.load()`), plays the Zone-Cleared overlay (`bossOutcome = 'victory'`), and — with auto-fight on — immediately re-challenges, potentially looping rejected claims against the Zone-Cleared fanfare. `bossUnlockedNextZone` is safe (it re-reads authoritative challenge state), which makes the unguarded `markZoneCleared`/overlay stand out.

## Suggested fix

Have `claimVictory` return a success flag; the boss path skips the celebration side effects on a definite error. The ambiguous lost-response case can keep the current optimistic behavior, mirroring the existing resync rationale.

ginderjeremiah:
## Summary

`resolveBossVictory()` (`UI/src/lib/engine/battle/enemy-manager.ts`) called `claimVictory()` but never inspected its outcome. On a *definitive* server rejection of the boss's `DefeatEnemy` claim (not just a lost/timed-out response), the client still:

- marked the zone cleared (`statistics.markZoneCleared`) — the "Cleared" seal shows wrong until the next `statistics.load()`
- played the Zone-Cleared overlay (`bossOutcome = 'victory'`)
- with auto-fight on, immediately re-challenged the boss, potentially looping rejected claims against the Zone-Cleared fanfare

## Fix

`claimVictory()` now returns a `rejected` flag distinguishing a **definitive rejection** from the existing **ambiguous lost-response** case:

- The backend's `DefeatEnemy` handler always attaches `data` (e.g. the cooldown) via `ErrorWithData` when it rejects a claim, while a transport-level failure (timeout / dropped connection) never carries `data` at all — the command may still have succeeded server-side in that case. That's the reliable signal to tell the two apart (see `docs/frontend-sockets.md`'s socket request lifecycle).
- `resolveBossVictory` now checks `rejected`: on a genuine rejection it skips the celebration entirely and falls back to the idle loop, mirroring how a boss loss/draw is already handled. On the ambiguous case, it keeps today's optimistic behavior unchanged.
- The idle victory path is unaffected — it has no equivalent "celebration" side effect to guard.

Closes #2344.

## Test plan

- [x] Added a unit test covering a definitive rejection: no zone-cleared, no overlay, falls back to idle, and — critically — does not re-challenge even with auto-fight on.
- [x] Added a unit test covering the ambiguous (ex: timed-out) case: celebration still proceeds as before.
- [x] `npm run test:unit:ci` — full suite green (3954 tests)

---
_Generated by [Claude Code](https://claude.ai/code/session_01PRyzD2yvfqybR8cCtnUy3N)_

### DIFF
--- a/UI/src/lib/engine/battle/enemy-manager.ts
+++ b/UI/src/lib/engine/battle/enemy-manager.ts
@@ -42,6 +42,11 @@ interface BattleEndResult {
 	cooldown: number;
 	nextEnemy?: IEnemyInstance;
 	nextZoneId?: number;
+	/** True only when the server definitively rejected the victory claim — an error response that still
+	 *  carries `data` (see docs/frontend-sockets.md: a transport-level failure, the ambiguous case, never
+	 *  carries `data` at all). Lets a caller tell "the claim genuinely didn't happen" apart from "we don't
+	 *  know if it happened," so it can skip celebrating a win the server never credited. */
+	rejected: boolean;
 }
 
 /** A server-prefetched next idle battle, held across the post-battle cooldown so the loop can begin it
@@ -678,11 +683,20 @@ export class EnemyManager {
 		// awaits below — otherwise the clear and the "next zone unlocked" check could target the wrong zone.
 		const clearedZoneId = playerManager.currentZone;
 
-		await this.claimVictory();
+		const { rejected } = await this.claimVictory();
 		// A stop / retreat that landed during the victory claim has already transitioned us out of the boss
 		// loop (or this resolution spans a stop()→start() cycle, #2331); abandon so we don't resurrect the
 		// cleared overlay over the new state.
 		if (!this.bossLoopActive(generation)) {
+			return;
+		}
+		if (rejected) {
+			// The server definitively rejected the claim (not just a lost/ambiguous response) — the zone was
+			// never actually cleared, so skip the celebration entirely rather than faking the Zone-Cleared
+			// seal/overlay and, with auto-fight on, looping rejected claims against that fanfare. Fall back to
+			// the idle loop, mirroring a boss loss/draw.
+			this.returnToIdle();
+			await this.getNewEnemy();
 			return;
 		}
 
@@ -803,7 +817,8 @@ export class EnemyManager {
 		return {
 			cooldown: defeatResponse.data?.cooldown ?? 0,
 			nextEnemy: defeatResponse.data?.nextEnemy,
-			nextZoneId: defeatResponse.data?.nextZoneId
+			nextZoneId: defeatResponse.data?.nextZoneId,
+			rejected: !!defeatResponse.error && defeatResponse.data !== undefined
 		};
 	}
 }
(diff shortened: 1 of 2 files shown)


ginderjeremiah (commented):
## Review: skip boss-victory celebration on a definitive `DefeatEnemy` rejection (#2344)

**Verdict:** Approve — correct, minimal fix that resolves the issue exactly as scoped, with mirrored tests for both branches of the new signal. (Posted as a comment since I'm the PR author; the intent is an approval.)

I traced the linchpin claim — that a *definitive* rejection is distinguishable from an ambiguous lost response — end to end and it holds:

- **Backend:** `DefeatEnemy.HandleExecuteAsync` has exactly two rejection paths (`DefeatEnemy.cs:27` "No active enemy." and `:77` "Enemy could not be defeated."), and *both* return `ErrorWithData(...)` with a populated `DefeatEnemyResponse` (the cooldown). So a genuine rejection always carries `data`.
- **Transport:** per `docs/frontend-sockets.md` (socket request lifecycle), a dropped connection / per-request timeout resolves with an `error` and **no** `data`, and in-flight commands are deliberately not blind-resent — so the ambiguous case never carries `data`.
- Therefore `rejected: !!defeatResponse.error && defeatResponse.data !== undefined` (`enemy-manager.ts:821`) is a sound discriminator: the only way to get `error` + `data` is a backend rejection.

The rejected branch in `resolveBossVictory` (`enemy-manager.ts:693-701`) mirrors `resolveBossDraw` (`returnToIdle()` + `getNewEnemy()`), correctly skips `markZoneCleared`/the overlay, and — via the shared `claimVictory` error path — still resyncs authoritative state. The idle path leaving `rejected` unread is right: it has no celebration side effect to guard and still honors the cooldown from the response.

### Test coverage

Good. Both branches of the new signal are covered in `enemy-manager-boss.test.ts`: the definitive-rejection test asserts no zone-cleared, no overlay, idle fallback, and — the key regression guard — no re-challenge even with auto-fight on; the ambiguous (lost-response) test asserts the optimistic celebration is preserved unchanged. This is a frontend UI-orchestration concern (celebration side effects), not battle-simulation logic, so no mirrored backend test is required.

### What's good

- The `rejected` vs ambiguous distinction is exactly the right call — it avoids both false "Cleared" state on a real rejection and needless pessimism on a transient blip.
- Clear inline comments explaining *why* each branch behaves as it does, and the fix stays tightly scoped to the boss path without disturbing the shared idle victory flow.

CI is still queued; I'll merge once it goes green (nothing else outstanding).

---

PR #2356, opened 2026-07-23
CI: add Playwright cache restore-keys, overlap API boot wait with Node setup

### LINKED ISSUE #2336: CI wall-clock: Playwright browser cache has no restore-keys, and the e2e job serializes the API boot wait ahead of Node/browser setup
Two independent wall-clock wins in `.github/workflows/run-tests.yml`, both on the longest job:

## 1. Playwright cache misses on every lockfile change

`run-tests.yml:339-344` keys the browser cache on `hashFiles('UI/package-lock.json')` with no `restore-keys`. Any lockfile change — and dependabot now runs weekly grouped npm updates — misses the cache and forces a full browser download (WebKit is the slow one) in all 3 matrix jobs. `~/.cache/ms-playwright` stores per-version directories, so a `restore-keys: playwright-${{ runner.os }}-${{ matrix.browser }}-` prefix makes `npx playwright install` incremental (only downloads a genuinely new browser build). Minor caveat: stale browser versions accumulate in the re-saved cache until it rotates.

## 2. API boot wait blocks before npm/browser setup

"Start backing stack" + "Apply the e2e admin-role fixture" (which contains the blocking `wait_for_api` loop) run at :292-326, **before** "Set up Node" / `npm ci` / "Install Playwright browser" (:328-355). `docker compose up -d` returns while the API is still migrating + content-seeding; the workflow then blocks doing nothing else. Moving the wait + `psql` fixture after browser install overlaps API boot with npm/browser setup — roughly the API boot time saved per matrix job, ×3.

ginderjeremiah:
## Summary

Two independent wall-clock wins on `test-e2e`, the longest job in `run-tests.yml`:

1. **Playwright cache had no `restore-keys`.** The cache key was keyed solely on `hashFiles('UI/package-lock.json')`, so any lockfile change (including weekly dependabot npm bumps) missed the cache entirely and forced a full browser download — WebKit is the slow one — in all 3 matrix jobs. `~/.cache/ms-playwright` stores per-version directories, so adding a `restore-keys: playwright-${{ runner.os }}-${{ matrix.browser }}-` prefix lets a lockfile-only change fall back to the previous cache and makes `npx playwright install` incremental (only the genuinely new browser version downloads).
2. **The blocking API-boot wait ran ahead of Node/browser setup.** `docker compose up -d --build` returns once the containers start, not once the API finishes migrating + seeding content + building its reference-data caches. The "Apply the e2e admin-role fixture" step (which contains the blocking `wait_for_api` poll loop) previously ran immediately after, before "Set up Node" / `npm ci` / Playwright browser install — so the workflow blocked doing nothing while the API was still booting. Moved that step (wait + `psql` fixture) to run after Node/browser setup, so API boot now overlaps with Node/npm/Playwright setup instead of serializing ahead of it.

Closes #2336.

## Changes

- `.github/workflows/run-tests.yml`: added `restore-keys` to the Playwright browser cache step; reordered `test-e2e` steps so "Set up Node" → "Install node modules" → "Cache/Install Playwright browser" run before "Apply the e2e admin-role fixture" (previously the reverse). No behavior change to what each step does, only when it runs — the wait loop still blocks on the same readiness check before the fixture SQL and the tests run.

## Test plan

- [x] Validated the workflow YAML still parses (`python3 -c "import yaml; yaml.safe_load(...)"`).
- [x] Manually traced the reordered step dependencies: nothing in "Set up Node" / `npm ci` / Playwright cache-and-install depends on the backing stack being *ready* (only depends on the repo being checked out), so moving them ahead of the readiness wait is safe.
- [ ] This is a CI-workflow-only change with no application code — the real verification is a green `test-e2e` matrix run on this PR itself (can't be run locally in the sandbox).

---
_Generated by [Claude Code](https://claude.ai/code/session_01DUkVhg4TW9ge7mmTcCbcV4)_

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -301,7 +301,42 @@ jobs:
       # (An earlier revision built the image via buildx with a GitHub Actions layer cache; measured
       # runs showed the cache-transfer + image-load overhead outweighed the savings on the hosted
       # runners, so the plain compose build is both simpler and faster here.)
+      #
+      # `up -d` returns once the containers start, not once the API finishes migrating/seeding, so
+      # the Node/Playwright setup below runs concurrently with that startup work instead of blocking
+      # on it — the blocking wait moved to "Apply the e2e admin-role fixture", just before the tests.
       run: docker compose up -d --build
+
+    - name: Set up Node
+      uses: actions/setup-node@v7
+      with:
+        node-version-file: UI/.nvmrc
+        cache: npm
+        cache-dependency-path: UI/package-lock.json
+
+    - name: Install node modules
+      working-directory: UI
+      run: npm ci
+
+    - name: Cache Playwright browser
+      id: playwright-cache
+      uses: actions/cache@v6
+      with:
+        path: ~/.cache/ms-playwright
+        key: playwright-${{ runner.os }}-${{ matrix.browser }}-${{ hashFiles('UI/package-lock.json') }}
+        restore-keys: |
+          playwright-${{ runner.os }}-${{ matrix.browser }}-
+
+    - name: Install Playwright browser
+      working-directory: UI
+      # Only this job's browser is needed. On a cache hit the binaries are already present, so just
+      # (re)install the OS-level dependencies, which actions/cache can't carry. On a miss, install both.
+      run: |
+        if [ "${{ steps.playwright-cache.outputs.cache-hit }}" = "true" ]; then
+          npx playwright install-deps ${{ matrix.browser }}
+        else
+          npx playwright install --with-deps ${{ matrix.browser }}
+        fi
 
     - name: Apply the e2e admin-role fixture
       # The API seeds the static content (Skills, Enemies, Zones, …) itself on startup from the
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: CI wall-clock wins on `test-e2e` — Playwright cache `restore-keys` + overlap API boot with Node/browser setup (closes #2336)

**Verdict:** Approve — implements exactly the two changes #2336 called for, correctly and with no application-code or test-coverage impact. (Submitted as a `COMMENT` because GitHub doesn't accept a gating self-review; treat it as an approval.)

Verified against the issue and the reordered workflow (`.github/workflows/run-tests.yml`):

- **`restore-keys` (line 327-328).** The added `playwright-${{ runner.os }}-${{ matrix.browser }}-` prefix is correct. On a lockfile-only change the exact key misses but the prefix restores the prior cache, so `steps.playwright-cache.outputs.cache-hit` is `false` (exact-match only) → the `else` branch runs `npx playwright install --with-deps`, which is now incremental against the restored binaries and only downloads a genuinely new browser build. The install-logic branch at 335-339 stays correct under all three cases (exact hit / partial restore / cold cache).
- **Step reordering (301-368).** "Set up Node" → `npm ci` → Playwright cache/install now run before "Apply the e2e admin-role fixture", so API boot (migrate + seed + reference-cache load kicked off by `docker compose up -d`) overlaps that setup instead of serializing ahead of it. Nothing in the moved-up steps depends on the backing stack being *ready* — only on the checkout — so the reorder is safe, and the blocking `wait_for_api` + `psql` fixture still gate the tests exactly as before.

### Nits
None.

### Test coverage
N/A — CI-workflow-only change, no application code. The real verification is a green `test-e2e` matrix on this PR.

### What's good
- The updated inline comments (299-307, 341-351) accurately document *why* the wait moved and stay in sync with the new order — no stale guidance left behind.
- Scope is tight and matches the issue precisely; the known stale-browser-accumulation tradeoff of `restore-keys` was already acknowledged in #2336 and is self-limited by GitHub's cache eviction, so it's an accepted cost rather than an open concern.

---

PR #2358, opened 2026-07-23
Reject arrays from CodeGen's NeedsInterface so they hit the unmapped-type throw

### LINKED ISSUE #2338: Codegen: array-typed DTO members silently emit invalid TypeScript instead of failing generation
## Problem

`Extensions.IsEnumerable` (`Game.Api/CodeGen/Extensions.cs:91-112`) only recognizes generic types — an array (`int[]`, `string[]`) is not generic, so it falls through to `NeedsInterface()` → `IsClassThatNeedsInterface`, which matches it (`type.IsClass` is true for arrays, and they aren't `string`). `GetTypeText` then returns `GetInterfaceName(...)` — e.g. `IInt32[]` — instead of reaching the deliberate "no TypeScript mapping" throw at `CodeGenTypeFormatter.cs:126-133`, and `ApiInterfaceWriter` would walk the array's CLR properties (`Length`, `SyncRoot`, …) into a garbage interface file.

This contradicts the codegen's own engineered guarantee that unmapped types "surface at generation time instead" (the rationale written for the struct allow-list at `Extensions.cs:114-119`). `CodeGenExtensionsTests.cs:55-58` pins `IsEnumerable(int[]) == false` but nothing pins the downstream consequence.

## Impact

Latent — no current DTO uses an array (all use `List<>`/`IEnumerable<>`), but `string[]` is the natural C# reach, and the failure mode is a silently broken frontend build rather than the loud generation failure every other unmapped type gets.

## Suggested fix

Treat `type.IsArray` in `IsEnumerable` (descriptor support via `NullabilityInfo.ElementType`), or make arrays hit the `GetTypeText` throw. Add a test pinning whichever behavior is chosen.

ginderjeremiah:
## Summary

- `CodeGenExtensions.IsClassThatNeedsInterface` (`Game.Api/CodeGen/Extensions.cs`) matched arrays (`type.IsClass` is `true` for arrays and they aren't `string`), because arrays aren't generic and so are rejected by `IsEnumerable()` and fall through to this check.
- That meant `NeedsInterface()` returned `true` for an array-typed DTO member, and `GetTypeText` would resolve it via `GetInterfaceName` — emitting a garbage interface (e.g. `IInt32[]`) that walks the array's own CLR properties (`Length`, `SyncRoot`, …) instead of reaching the deliberate "no TypeScript mapping" throw every other unmapped type gets.
- Fix: exclude `type.IsArray` in `IsClassThatNeedsInterface`, so an array now correctly falls through to `GetTypeText`'s existing throw, surfacing at generation time instead of silently emitting a broken interface.

No current DTO uses an array (all use `List<>`/`IEnumerable<>`), so this is latent hardening, not a live-bug fix — the issue calls out `string[]` as the natural place a new DTO could reach for one.

Closes #2338. Went with the "make arrays hit the `GetTypeText` throw" option from the issue's suggested fix rather than adding full array codegen support — the latter would require threading `NullabilityInfo.ElementType` through `CodeGenTypeDescriptor.GenericArgumentDescriptors`, `GetImportTexts`, and `GetInterfaceName`'s generic-parameter handling, which is a materially larger change for a shape nothing in the codebase currently uses. If a real need for array-typed DTOs comes up, `List<>`/`IEnumerable<>` already work and are the house convention.

## Test plan

- [x] Added `NeedsInterface_IntArray_ReturnsFalse` / `NeedsInterface_ClassArray_ReturnsFalse` in `CodeGenExtensionsTests.cs`.
- [x] Added `ModelWithArray` fixture and `GetTypeText_UnmappedArray_Throws` in `CodeGenTypeFormatterTests.cs`, pinning that an array property now throws `InvalidOperationException` instead of emitting a broken interface reference.
- [x] `dotnet build Game.sln` — succeeds, 0 warnings/errors.
- [x] `dotnet test Game.Api.Tests` — 864 passed, 0 failed.

Closes #2338

---
_Generated by [Claude Code](https://claude.ai/code/session_01PSUFPmdoXsdZSy5isVbtcw)_

### DIFF
--- a/Game.Api.Tests/CodeGen/CodeGenExtensionsTests.cs
+++ b/Game.Api.Tests/CodeGen/CodeGenExtensionsTests.cs
@@ -244,6 +244,20 @@ namespace Game.Api.Tests.CodeGen
             Assert.True(typeof(Dictionary<TestEnum, SimpleModel>).NeedsInterface());
         }
 
+        [Fact]
+        public void NeedsInterface_IntArray_ReturnsFalse()
+        {
+            // Arrays are IsClass but not enumerable per IsEnumerable (not generic); NeedsInterface must
+            // reject them too so GetTypeText's unmapped-type throw fires instead of emitting IInt32[].
+            Assert.False(typeof(int[]).NeedsInterface());
+        }
+
+        [Fact]
+        public void NeedsInterface_ClassArray_ReturnsFalse()
+        {
+            Assert.False(typeof(SimpleModel[]).NeedsInterface());
+        }
+
         [Theory]
         [InlineData("HelloWorld", "helloWorld")]
         [InlineData("A", "a")]
--- a/Game.Api.Tests/CodeGen/CodeGenTypeFormatterTests.cs
+++ b/Game.Api.Tests/CodeGen/CodeGenTypeFormatterTests.cs
@@ -391,6 +391,16 @@ namespace Game.Api.Tests.CodeGen
             var ex = Assert.Throws<InvalidOperationException>(() => CodeGenTypeFormatter.GetTypeText(descriptor));
             Assert.Contains("DateTimeOffset", ex.Message);
         }
+
+        [Fact]
+        public void GetTypeText_UnmappedArray_Throws()
+        {
+            // Arrays have no TypeScript mapping (List<>/IEnumerable<> are the supported shapes);
+            // NeedsInterface must reject them so this throws instead of emitting a garbage IString[].
+            var descriptor = GetPropertyDescriptor<ModelWithArray>("Tags");
+            var ex = Assert.Throws<InvalidOperationException>(() => CodeGenTypeFormatter.GetTypeText(descriptor));
+            Assert.Contains("String[]", ex.Message);
+        }
     }
 
     public class GenericHolder
--- a/Game.Api.Tests/CodeGen/TestFixtures.cs
+++ b/Game.Api.Tests/CodeGen/TestFixtures.cs
@@ -109,6 +109,14 @@ namespace Game.Api.Tests.CodeGen
         public string Title { get; set; } = "";
     }
 
+    // An array-typed member has no TypeScript mapping (List<>/IEnumerable<> are the supported shapes);
+    // GetTypeText must throw rather than silently emit a reference to an interface reflecting the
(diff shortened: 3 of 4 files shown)


ginderjeremiah (commented):
## Review: CodeGen `NeedsInterface` rejects arrays so they hit the unmapped-type throw (#2338)

**Verdict:** Approve — correct, minimal fix that closes the latent codegen hole; solid test coverage; no blocking or should-fix findings. (Submitted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating verdict on one's own PR.)

I traced the fix end to end and it does exactly what it claims:

- For `string[]`, `IsEnumerable()` is false (arrays aren't generic), so before this change it fell through to `IsClassThatNeedsInterface`, matched (`IsClass` true, not `string`), and `GetTypeText` emitted `GetInterfaceName(...)` — a garbage `IString[]` reflecting the array's CLR members. With `!type.IsArray` added, `NeedsInterface()` now returns false and `GetTypeText` reaches its deliberate `InvalidOperationException` throw (`CodeGenTypeFormatter.cs:130`), so an unmapped array surfaces loudly at generation time like every other unmapped type.
- The nesting case is consistent too: `List<string[]>` recurses into `string[].NeedsInterface()` (now false) and still throws in `GetTypeText`, rather than silently emitting a broken element type.

The choice to make arrays throw rather than add full array codegen (threading `NullabilityInfo.ElementType` through the descriptor/import/interface-name machinery) is the right call for a shape nothing in the codebase uses — `List<>`/`IEnumerable<>` are the house convention and already work.

### Test coverage
- `NeedsInterface_IntArray_ReturnsFalse` / `NeedsInterface_ClassArray_ReturnsFalse` pin the predicate directly.
- `GetTypeText_UnmappedArray_Throws` (with the new `ModelWithArray` fixture) pins the downstream consequence the issue called out as previously unpinned — the assertion on `"String[]"` matches `type.FullName` (`System.String[]`). This is the important one; it guards the actual failure mode, not just the helper.

### What's good
- Tightly scoped one-line production change with a clear rationale comment mirroring the existing struct allow-list reasoning.
- Tests cover the predicate and the real generation-time behavior, and the PR body honestly frames this as latent hardening rather than a live-bug fix.

---

PR #2359, opened 2026-07-23
Commit non-admin ban/archive before revoking the target's socket

### LINKED ISSUE #2339: Admin ban/archive kicks the target's socket before the ban is durably committed
## Problem

For a non-admin target, `Users.SetUserTimestamp` (`Game.DataAccess/Repositories/Users.cs:317-342`) mutates the tracked entity and returns `Success` without saving — the row commits only in the deferred `CommitFilter` after the action returns. But `AdminUsersController` (`Game.Api/Controllers/Admin/AdminUsersController.cs:83-108`) calls `_socketManager.RevokeAccess(playerIds)` inside the action, i.e. **before** the commit.

Consequences:

1. A transient commit failure kicks the player's live socket while leaving them un-banned (the admin sees the 500, so it's recoverable, not silent).
2. The kicked client's automatic refresh/reconnect can race the not-yet-committed ban — a window overlapping #2231 in effect but with a distinct commit-ordering cause.

The admin-target branch commits inline under its lockout transaction, so the asymmetry is specific to non-admin targets.

## Suggested fix

Emit `RevokeAccess` only after the write is durable — an explicit `SaveChangesAsync` before the kick in the action, or a result filter ordered after `CommitFilter`.

ginderjeremiah:
## Summary

For a non-admin target, `Users.SetUserTimestamp` (`Game.DataAccess/Repositories/Users.cs`) mutated the tracked `User` entity and returned `Success` without saving — the row only committed later, in the deferred `CommitFilter` that runs after the controller action returns. But `AdminUsersController.ArchiveUser`/`BanUser` call `_socketManager.RevokeAccess(playerIds)` inside the action, i.e. **before** that deferred commit.

Consequences this fixes:
1. A transient commit failure could kick the target's live socket while leaving them un-banned/un-archived (recoverable — the admin sees the 500 — but still a real inconsistency window).
2. The kicked client's automatic reconnect could race the not-yet-committed ban/archive.

The admin-target branch (`ApplyLifecycleActionToAdminAtomically`) already committed inline under its lockout transaction, so the asymmetry was specific to non-admin targets.

## Fix

`SetUserTimestamp`'s non-admin branch now calls `SaveChangesAsync` immediately after mutating the timestamp, before returning `Success` — mirroring the admin branch's existing in-tier commit. No controller changes were needed: by the time `ArchiveUser`/`BanUser` see `UserActionStatus.Success`, the write is now durable in both branches, so the socket revoke can never outrun the commit.

Also updated `docs/backend-admin.md` to state this durability guarantee explicitly.

## Test plan

- [x] New `ArchiveUser_NonAdminTarget_CommitsBeforeReturningSuccess` / `BanUser_NonAdminTarget_CommitsBeforeReturningSuccess` integration tests call the repository directly (bypassing the HTTP pipeline's `CommitFilter` entirely) and assert the row is already durable via a fresh `AsNoTracking()` read on the same `DbContext` — this only passes if the fix's explicit `SaveChangesAsync` actually ran.
- [x] `dotnet test` — full solution suite green: 3,463 tests passed (Game.Core.Tests, Game.Infrastructure.Tests, Game.Api.Tests, Game.Application.Tests), including the existing `ArchiveUser`/`BanUser` socket-revoke and admin-lockout tests.

Closes #2339

---
_Generated by [Claude Code](https://claude.ai/code/session_015DB61Es7NebJfgPx7Jt5m3)_


Game.DataAccess/Repositories/Users.cs
```diff
@@ -338,6 +338,12 @@ private async Task<UserActionStatus> SetUserTimestamp(int actingUserId, int targ
             if (!targetHasAdminRole)
             {
                 setTimestamp(user);
+
+                // Commit here (rather than deferring to the per-request unit of work) so the row is durable
+                // before AdminUsersController revokes the target's live socket: without this, a transient
+                // commit failure in the deferred path could kick the socket while leaving the account
+                // un-banned/un-archived (#2339).
+                await _context.SaveChangesAsync(cancellationToken);
```
ginderjeremiah, line 346: The pre-existing comment two lines above (line 335) — "archiving/banning a non-admin ... **applies on the deferred unit-of-work path**" — is now stale and directly contradicts this new in-tier `SaveChangesAsync`. That's exactly the sort of leftover a future reader will trust over the code, so please update it (e.g. drop the deferred-path clause / note it commits in-tier here for the same durability reason as the admin branch). Should-fix, not blocking.

---

ginderjeremiah (commented):
## Review: commit non-admin ban/archive before revoking the target's socket (#2339)

**Verdict:** Approve (posted as a comment — I'm the PR author, so GitHub won't take a gating verdict from me) — one should-fix on a stale comment, otherwise correct and well-tested.

The change is exactly right and minimal. The non-admin branch of `Users.SetUserTimestamp` now `SaveChangesAsync`-es before returning `Success`, mirroring the admin branch's existing in-tier commit under `ApplyLifecycleActionToAdminAtomically`. Since `AdminUsersController.ArchiveUser`/`BanUser` only call `RevokeAccess` when `status is UserActionStatus.Success`, the socket kick can no longer outrun a durable write in either branch — closing the commit-ordering window described in the issue. The redundant later `SaveChanges` from `CommitFilter` is a harmless no-op (nothing left tracked-modified).

### Should-fix

- `Game.DataAccess/Repositories/Users.cs:334-335` — the pre-existing comment ("archiving/banning a non-admin ... applies on the deferred unit-of-work path") is now stale and contradicts the new in-tier `SaveChangesAsync` right below it. Update it so a future reader isn't misled. Trivial, but worth doing in this PR rather than leaving the contradiction in place.

### Test coverage

- Solid. `ArchiveUser_NonAdminTarget_CommitsBeforeReturningSuccess` / `BanUser_NonAdminTarget_CommitsBeforeReturningSuccess` call the repository directly (bypassing the HTTP `CommitFilter`) and assert durability via a fresh `AsNoTracking()` read on the same context — which only passes if the explicit `SaveChangesAsync` actually ran. Remove the fix and these fail, so they genuinely guard the behavior. Both archive and ban paths covered; existing socket-revoke and admin-lockout tests remain in place.

### What's good

- Right layer for the fix (data tier, alongside the admin branch's existing in-tier commit) rather than reordering filters in the API layer — keeps the durability guarantee where `Success` is produced.
- `docs/backend-admin.md` updated to state the "kick only ever follows a durable ban/archive" guarantee explicitly, so the invariant is documented, not just implied.

---

ginderjeremiah:
## Re-review — should-fix resolved, merging

Follow-up commit `6734a02` addresses the one should-fix from the prior review: the stale comment at `Users.cs:334` now reads *"carries no such hazard, but still commits in-tier below for the socket-revoke durability guarantee,"* and the new inline comment on the `SaveChangesAsync` call clearly explains why the write commits here rather than on the deferred unit-of-work path (tied to the socket-revoke ordering in #2339). That's the last contradiction gone.

Re-verified the current head (`6734a02`):
- **Fix is correct and minimal** — the non-admin branch now commits in-tier before returning `Success`, mirroring the admin branch, so `AdminUsersController`'s `RevokeAccess` can never outrun a durable ban/archive in either branch.
- **Tests genuinely guard it** — the two new integration tests read back via `AsNoTracking()` on the same context, which only passes if the explicit `SaveChangesAsync` ran; remove the fix and they fail.
- **Docs** (`backend-admin.md`) accurately state the "kick only ever follows a durable ban/archive" invariant for both paths.
- The 4-line comment on the commit call is justified context, not over-detailed.

**Verdict: approve.** CI is green on `6734a02` (`all-checks-passed` ✓, backend + e2e green, frontend correctly skipped). No remaining nits or should-fixes. Clean, well-scoped bug fix closing #2339 — merging.

---

Issue #2340, opened 2026-07-23
SavePlayer re-serializes and rewrites the entire player blob to Redis on every save — the last full-rewrite-per-battle path in the tier


ginderjeremiah:
## Problem

`PlayerRepository.SavePlayer` (`Game.DataAccess/Repositories/PlayerRepository.cs:180`) runs `PlayerCacheMapper.ToCacheModel(player)` + full JSON serialization on every save — at minimum once per battle end per connected player — even when the battle only changed `Exp`/`Level`/`LastActivity`/seed. `ToCacheModel` walks all unlocked items, applied mods, skills, lessons, and log prefs, allocating a dictionary + several lists, and the payload grows linearly with account age (hundreds of unlocked items/skills late-game). The Redis write is fire-and-forget, but the mapping + serialization run synchronously on the per-command path.

The progress side of this exact path was already converted to per-row hash `HSET`s to avoid rewriting the whole snapshot (#1635, `PlayerProgressRepository`); the player blob is the only full-rewrite-per-battle left.

## Suggested direction

Measure first (a late-game-shaped player fixture), then either split the blob into a Redis hash with per-section fields (core / items / skills / lessons) written only when the corresponding domain events fired this save, or skip re-serializing sections untouched by the dispatched event set (`SavePlayer` already knows which events were buffered). Preserve the load path's atomicity semantics — a partially-written multi-key layout must never rehydrate a torn aggregate (the TTL/refresh semantics need the same care).

<sub>From the scheduled codebase health review.</sub>

ginderjeremiah:
Picked this up and did the "measure first" step in #2360 before touching anything, since both suggested directions (a Redis-hash split mirroring #1635, or dirty-tracking to skip unchanged sections) add real complexity to the write-behind player cache — the hottest, most correctness-sensitive persistence path in the game.

**Measured** (`PlayerCacheMappingPerformanceTests` + `PlayerCachePersistencePerformanceTests`, against a synthetic late-game account: 400 unlocked items, 200 applied mods, 200 skills):
- `ToCacheModel` + JSON serialization in isolation: ~0.38 ms, scaling **sub-linearly** with account size (9.7x cost for a 20x larger account — no accidental O(n²)).
- Against a real `SavePlayer` round trip (Postgres + Redis): ~1.4 ms late-game vs. ~0.9 ms for a fresh account — so mapping/serialization is roughly **a quarter to a third** of the total per-save cost.

[...]

PR #2360, opened 2026-07-23
Measure PlayerCacheMapper mapping/serialize cost before redesigning it (#2340)

ginderjeremiah:
## Summary

#2340 flags `PlayerCacheMapper.ToCacheModel` + full JSON serialization running on every `SavePlayer`, and explicitly asks to **measure first** before picking between its two suggested directions (splitting the player blob into a Redis hash mirroring #1635, or skipping re-serialization of sections untouched by the save's dispatched events). Both directions add real complexity to the hottest, most correctness-sensitive persistence path in the game (the write-behind player cache — see `docs/backend-persistence.md`), so this PR does the measurement step and reports back with data rather than picking a direction speculatively.

- `PlayerCacheMappingPerformanceTests` (new, `Game.Application.Tests/Mapping`) measures `ToCacheModel` and `ToCacheModel().Serialize()` in isolation for a **late-game-shaped** synthetic account (400 unlocked items, 200 applied mods, 200 skills) and asserts the cost scales roughly linearly with account size (catching an accidental O(n²) regression), following the same scaling-gate convention as `BattlePerformanceTests`.
- `PlayerCachePersistencePerformanceTests` (new, `Game.Application.Tests/Services/Performance`) measures the same account through a **real `SavePlayer` round trip** against Postgres/Redis, compared against a freshly-created account, to see what share of the total save the mapping/serialization actually accounts for.

### What the numbers show

- Mapping + serializing a large late-game account costs **well under half a millisecond** (~0.38 ms) and scales **sub-linearly** with account size (9.7x cost for a 20x larger account) — no accidental quadratic blowup.
- That's roughly **a quarter to a third of a full `SavePlayer` round trip** (~1.4 ms late-game vs. ~0.9 ms fresh-account, against real containers).

**Recommendation:** real but not dominant. Given the cost of either proposed redesign (a Redis-schema change needing the same partial-write/torn-aggregate care #1635's split took, or new dirty-tracking machinery threaded across every domain event the player aggregate raises) against a currently-modest win, I don't think either is warranted yet. I've documented this reasoning in `docs/backend-persistence.md` and left the two test classes in place as the regression guard/observability point if that balance shifts later (e.g. a much larger typical inventory, or real production profiling showing otherwise). I'll leave a comment on #2340 with this recommendation rather than closing it outright, since that's a call I'd like your sign-off on rather than deciding unilaterally.

### Also in this PR

- Moved `PerformanceMeasurement`/`MeasurementResult` (the timing harness `BattlePerformanceTests` already used) from `Game.Core.Tests` into `Game.Core.TestInfrastructure`, so the new suites share one implementation instead of starting a third copy of the same Stopwatch/warmup/sampling logic. Added an async overload (`MeasureAsync`) for the new I/O-bound integration test.

Implements #2340's own suggested first step ("measure first") with real numbers against real containers, and documents a data-backed recommendation rather than guessing at which (if either) of the two proposed redesigns to build.

## Test plan

- [x] `dotnet test` (full solution) — 3,464 tests passed (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), including the moved/refactored performance harness and the two new performance test classes.
- [x] New performance tests run individually and log the figures quoted above.

Closes nothing directly — leaving #2340 open pending your read on the recommendation above.

https://claude.ai/code/session_01WaDEuTQBNAj6hnNnZTi41E

---
_Generated by [Claude Code](https://claude.ai/code/session_01WaDEuTQBNAj6hnNnZTi41E)_


ginderjeremiah (commented):
## Review: "measure-first" instrumentation for #2340 (player-blob full-rewrite-per-save cost)

**Verdict:** Approve (posted as `COMMENT` — I'm the PR author, so GitHub won't accept a gating event from me). No blocking findings. One nit below.

This does exactly what #2340 asks for as its first step — measures before committing to either proposed redesign — and reports back with data instead of speculatively building a Redis-hash split or per-section dirty tracking on the hottest, most correctness-sensitive persistence path. The recommendation to defer both redesigns is well-argued and lands on the right side of the complexity/win tradeoff given the numbers (~0.4 ms of a ~1.4 ms save). Leaving the two suites in place as the regression guard is the right way to keep the option open.

I checked the concerns that would actually bite here:
- `PlayerCacheMapper.ToCacheModel` is `public`, and `Game.Application.Tests` already references `Game.Core.TestInfrastructure`, so the harness move and cross-project calls are sound (not just "it compiled").
- `MeasureAsync` faithfully mirrors the sync `Measure` contract — same warmup, same untimed per-sample input build, same `Settle()`-before-timing, same min-across-samples statistic. No divergence in semantics.
- The timing assertions all use generous, precedent-following gates (2× linear budget over a 20× size ratio; coarse catastrophic ceilings) consistent with `BattlePerformanceTests`/`BattleRoundTripPerformanceTests`, so no new CI-flake surface beyond what the repo already accepts.
- `SeedPlayerAsync` correctly reloads the reference caches after seeding rows directly, which the `GetPlayer` rehydration needs (per the reference-data cache rules in `backend.md`).

### Nits

- `Game.Core.TestInfrastructure/Performance/PerformanceMeasurement.cs` (sync `Measure`, ~line 59) still carries a battle-specific inline comment — "Build all inputs for this sample untimed, so combatant construction (and the garbage it produces) stays out of the measured region." The class-level summary was correctly generalized ("the subject may be stateful and unsafe to reuse"), but this inline comment wasn't; now that the harness is shared beyond battle it reads as a stale reference. Not in this PR's diff hunks so no inline comment — worth a one-line tidy on a future touch. (Not blocking.)

### Test coverage

Appropriate for what this PR is: the deliverable *is* the measurement. Both suites are `[Trait("Category","Performance")]`, the persistence one is a real-container `[Collection("Integration")]` round trip, and both gate only on coarse regression ceilings while logging the real figures — matching the observability-first convention the battle perf tests already established. No production logic changed, so no additional domain/unit coverage is owed.

### What's good

- Consolidating the timing harness into `Game.Core.TestInfrastructure` instead of spawning a third copy is the right call, and the async overload is a clean, minimal addition.
- The rationale is captured in `docs/backend-persistence.md` next to the existing write-behind/progress-hash discussion, so the "why we didn't split the blob" decision is discoverable rather than buried in a closed PR.

### On merging

Holding off deliberately, for three reasons: (1) CI is still `in_progress`, not green; (2) there's an open nit; and (3) the PR itself explicitly requests your sign-off on the *recommendation* (whether to build either redesign) rather than deciding unilaterally — and the doc change bakes in "the player blob therefore stays a single serialized value for now," which is exactly that call. That decision is yours to confirm on #2340. The instrumentation and doc here are a sound standalone change and can merge once you're comfortable with the recommendation and CI is green.

---

ginderjeremiah:
Thanks for the review. Fixed the nit (the stale "combatant construction" comment in `PerformanceMeasurement`'s sample-building loop, now generalized to match the class summary) in b819034. CI was already green on the prior commit; re-running now on this push. Merge is otherwise waiting on your sign-off on the deferral recommendation over on #2340, per your note above.

---

ginderjeremiah:
## Re-review — nit resolved, code good; merge is your call on the recommendation

Follow-up commit `b819034` resolves the one nit from the prior review: the inline comment in `PerformanceMeasurement`'s sample-building loop no longer references battle-specific "combatant construction," bringing it in line with the already-generalized class summary. Nothing else changed, and I re-confirmed the harness move + async overload are still sound.

**Verdict on the code: approve.** CI is now green on `b819034` (`all-checks-passed` ✓, backend + e2e green, frontend skipped). No open nits.

**Not merging — this one needs your sign-off, by the PR's own design.** The value here isn't just the two perf suites; it's the recommendation baked into `docs/backend-persistence.md` ("the player blob therefore stays a single serialized value for now"), which is exactly the call #2340 asked to make *after* measuring. Merging would commit that "defer both redesigns" decision. The measurement backs it well (~0.4 ms of a ~1.4 ms save, sub-linear scaling, no O(n²)), and I agree with deferring — but confirming it against the two proposed directions is owner discretion, not something I should merge unilaterally. Ready to merge as soon as you confirm the recommendation on #2340.

---

ginderjeremiah:
## Re-review (delta only)

Only change since my prior review (`66985d66` → `b819034d`) is commit `b819034d`, a 2-line comment tidy in `Game.Core.TestInfrastructure/Performance/PerformanceMeasurement.cs` that generalizes the stale "combatant construction" inline comment — exactly the one nit I flagged. Nit resolved; no production/test logic changed.

**Verdict unchanged: approve.** No blocking or should-fix findings remain.

**Still not merging**, for two reasons that stand independent of the nit:
1. This isn't a spike, but its deliverable bakes a design decision into `docs/backend-persistence.md` ("the player blob therefore stays a single serialized value for now") and the PR explicitly asks for your sign-off on the *recommendation* to defer both of #2340's proposed redesigns. That's an owner-discretion call, not one I should merge unilaterally — #2340 is still the place to confirm or redirect it.
2. CI reports no completed checks (`total_count: 0`, pending) on `b819034d` — no green signal to merge on.

Once you're comfortable with the recommendation and CI is green, the instrumentation + doc are a sound standalone change and can go in.

---

ginderjeremiah:
## Re-review (incremental)

Only change since my prior review is `b819034`, which addresses the one nit I raised (the stale "combatant construction" inline comment in the shared `PerformanceMeasurement`). Confirmed it's now consistent with the generalized class summary. Two of the three reasons I was holding merge on have cleared:

- ✅ CI is now green (`all-checks-passed` succeeded on `b819034`).
- ✅ Nit resolved.
- ⏳ The third remains, and it's yours alone: this PR's `docs/backend-persistence.md` change bakes in the recommendation to **not** split the player blob / add per-section dirty tracking yet. That's the exact call #2340 asks you to sign off on rather than have decided unilaterally.

Verdict stays **approve**; the instrumentation and doc are a sound standalone change. Not merging — leaving it for your read on the recommendation. Once you're comfortable with it, this is ready to go as-is.

---

PR #2361, opened 2026-07-23
Give write-behind envelopes a unique identity

### LINKED ISSUE #2341: Write-behind envelopes have no identity — duplicate payloads alias in stranded-head reclaim and LREM acknowledge
## Problem

`DomainEventEnvelope` (`Game.DataAccess/DomainEventEnvelope.cs`) is `{Type, Payload}` only — two independently-raised identical events (easy for insert-only events like a duplicated `SkillUnlockedEvent`, or two `ItemFavoriteChangedEvent`s toggling to the same value) serialize to byte-identical queue items. Two consequences:

- `DataProviderSynchronizer.TryReclaimStrandedProcessingAsync` (`DataProviderSynchronizer.cs:580`) tracks the processing-list head **by value** (`tracked.Value != head[0]`). If item A (payload P) is acknowledged and a *different* item with the same payload P rotates to the head, the dwell clock is not reset — the processing list can be reclaimed before the new head has genuinely dwelled past the 30s grace period, prematurely racing another instance's healthy in-flight work.
- `RedisQueue.AcknowledgeAsync` (`Game.Infrastructure/PubSub/Redis/RedisQueue.cs:74-79`) removes by `LREM value 1`; with equal payloads reserved on two instances, each can acknowledge the other's copy.

Both are harmless today only because every handler is idempotent and reclaim is documented as always-safe — but the grace-period guarantee the head-tracking comment promises quietly doesn't hold for equal payloads.

## Suggested fix

A per-envelope id (guid or enqueue sequence) on `DomainEventEnvelope` makes head-tracking, acknowledge, and dead-letter multiplicity (`PlayerUpdateDeadLetters.FilterToSnapshot`) exact. Low urgency; robustness hardening rather than an active defect.

ginderjeremiah:
## Summary

`DomainEventEnvelope` (`Game.DataAccess/DomainEventEnvelope.cs`) was `{Type, Payload}` only, so two independently-raised events that happen to serialize to byte-identical content (e.g. a duplicated `SkillUnlockedEvent`, or two `ItemFavoriteChangedEvent`s toggling to the same value) produced byte-identical queue items. `DataProviderSynchronizer`'s stranded-processing-head tracking and `RedisQueue.AcknowledgeAsync`'s `LREM value 1` both key off that raw serialized string, so equal payloads could alias one another there:

- The stranded-head clock (`TryReclaimStrandedProcessingAsync`) compares the processing list's head by value — if item A is acknowledged and a *different* item with an equal payload rotates to the head, the dwell clock wrongly fails to reset, risking a premature reclaim that races another instance's genuinely in-flight work.
- `LREM` removes by value, so with equal payloads reserved on two instances, each could acknowledge the other's copy.

Both were harmless today only because every write-behind handler is idempotent, but the grace-period guarantee the head-tracking comment promises didn't actually hold for equal payloads.

## Fix

`DomainEventEnvelope` now carries a `Guid Id` generated per instance, so two envelopes with identical `Type`+`Payload` still serialize to distinct raw strings — giving every queued item a genuine identity for both the stranded-head clock and per-item Redis list operations to key off. The property defaults (`Guid.NewGuid()`) rather than being `required`, so an envelope enqueued by a pre-upgrade instance mid-rolling-deploy (carrying no `"id"`) still deserializes cleanly on a newer instance.

## Test plan

- [x] New `DomainEventEnvelopeTests`: two envelopes with identical Type+Payload serialize to distinct strings and carry distinct Ids; a legacy envelope missing `"id"` still deserializes.
- [x] New `DataProviderSynchronizerTests.ProcessQueue_DuplicatePayloadReplacesProcessingHeadWithinGracePeriod_RestartsClockInsteadOfReclaiming`: reproduces the exact regression scenario — a duplicate-content envelope taking the processing-list head must restart the stranded-head clock rather than being aliased with the item it replaced.
- [x] Fixed an existing test (`StartAsync_CancellationAlreadyRequested_SkipsReclaimAndStartupDrain`) that incidentally relied on two independently-serialized envelopes of the same event being byte-identical; it now serializes once and reuses the string.
- [x] `dotnet test` (full solution) — 3,464 tests passed (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests).

Closes #2341

---
_Generated by [Claude Code](https://claude.ai/code/session_017LayWk8rjQ9KEuCX8XoD8Q)_

### DIFF
--- a/Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs
+++ b/Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs
@@ -959,9 +959,10 @@ namespace Game.Application.Tests.DataAccess
                 AutoChallengeBoss: false,
                 LastCreditedBattleSeed: null);
 
+            var message = Serialize(validEvent);
             var realPubSub = scope.ServiceProvider.GetRequiredService<IPubSubService>();
             var queue = realPubSub.GetQueue(Constants.PUBSUB_PLAYER_QUEUE);
-            await queue.AddToQueueAsync(Serialize(validEvent));
+            await queue.AddToQueueAsync(message);
 
             var logger = new CapturingLogger<DataProviderSynchronizer>();
             var pubsub = new SubscribeSuppressingPubSubService(realPubSub);
@@ -976,7 +977,7 @@ namespace Game.Application.Tests.DataAccess
             Assert.DoesNotContain(logger.Entries, e => e.Level == LogLevel.Error);
 
             // The event was never drained: it is still waiting on the queue and the player is unchanged.
-            Assert.Equal(Serialize(validEvent), await queue.GetNextAsync());
+            Assert.Equal(message, await queue.GetNextAsync());
 
             using var verifyScope = CreateScope();
             var verifyContext = verifyScope.ServiceProvider.GetRequiredService<GameContext>();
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Give write-behind envelopes a unique identity (#2341)

**Verdict:** Approve — the fix is minimal, correct, and well-targeted; equal-payload envelopes now carry genuine per-instance identity, restoring the grace-period guarantee the stranded-head tracking already promised. (Posted as a comment since I'm the PR author; intended verdict is **Approve**.)

I traced the change end to end against `docs/backend-persistence.md`:

- **Publish path** — `PlayerPersistencePublisher.Buffer` constructs each envelope with only `Type`/`Payload`, so `Id` picks up its `Guid.NewGuid()` default per instance. Two byte-identical events now serialize to distinct raw strings.
- **Consume path** — `DataProviderSynchronizer.TryReclaimStrandedProcessingAsync` (`_strandedProcessingHead` tracking) and `RedisQueue`'s `LREM value 1` acknowledge/dead-letter both key off that raw string, so distinct strings give them exact per-item identity. No re-serialization happens between enqueue and acknowledge, so the `Id` is never dropped mid-flight.
- **Reader** — `PlayerUpdateEnvelopeReader` only reads `Type`/`Payload`; adding `Id` is transparent to it and to the per-player lane routing (#1701).
- **Rolling-deploy safety** — `Id` defaults (via `init`) rather than being `required`, so a legacy envelope carrying no `"id"` still deserializes; System.Text.Json leaves the absent property at its default. Correctly reasoned and covered by a test. The residual legacy-vs-legacy aliasing during a deploy window is pre-existing, transient, and harmless under the idempotent at-least-once contract — an acceptable trade.

The fix is deliberately minimal: uniqueness is achieved at the envelope level, so `AcknowledgeAsync` and the dead-letter multiplicity the issue also flagged become exact without touching those files. That's the right call over spreading id-plumbing across the queue ops.

### Test coverage

- `DomainEventEnvelopeTests` — pure classical unit tests: distinct raw strings/Ids for identical Type+Payload, and legacy (`no "id"`) deserialization. Correctly placed as in-process unit tests rather than integration.
- `ProcessQueue_DuplicatePayloadReplacesProcessingHeadWithinGracePeriod_RestartsClockInsteadOfReclaiming` — reproduces the exact regression: a duplicate-content item taking the processing-list head must restart the stranded clock rather than aliasing the acknowledged item, then dead-letters on its own malformed inner payload. This is the meaningful test and it exercises the real failure path.
- Repairing `StartAsync_CancellationAlreadyRequested_SkipsReclaimAndStartupDrain` to serialize once (it had relied on two independent serializations being byte-identical) is the correct fix, not a workaround.

### Notes (non-blocking)

- No doc change was needed: `docs/backend-persistence.md` already describes the stranded-head clock as "tracked against that item's identity" — this PR makes that description actually true for equal payloads rather than aspirational.

Nothing to change. Waiting on backend/e2e CI before merging.

---

Issue #2245, opened 2026-07-21
Hygiene sweep: test-only public surface, dead members, and minor consistency nits from the health review


ginderjeremiah:
Small items from a codebase health review, bundled — none warrants its own issue. Same theme as #1707 (test-only surface) but none of these are in that issue's calibration scope.

## Test-only / unused public surface

- **`BackgroundWorker`** (`Game.Infrastructure/BackgroundWorker.cs`): only production consumer is `RedisPubSubService.cs:134` (async ctor + `Start()` + `Dispose()`). `IsRunning`, the synchronous `Action` ctor, and direct `Kill()` are exercised solely by `BackgroundWorkerTests`; nothing outside `Game.Infrastructure` references the type and both test assemblies have `InternalsVisibleTo`, so it can be `internal` with the unused members trimmed. Notable because a meaningful share of the file's concurrency machinery exists to keep `IsRunning` coherent for no production reader.
- **`SortedLinkedList`** (`Game.Core/Collections/SortedLinkedList.cs:10,17-20`): `Count` and the `Comparison&lt;T&gt;` convenience ctor are referenced only by its tests; the sole production consumer (`AttributeCollectionNode`) uses the `IComparer&lt;T&gt;` ctor, `Add`, `Remove`, and enumeration.
- **`GameContextFactory.GetGameContext`** (`Game.Infrastructure/Database/GameContextFactory.cs:9-14`): no production caller (production uses the `Configure` seam via `AddGameContext`); only two test files use it. Defensible as a test seam — either justify with a comment or drop it.
- **`UsageSection.svelte`** (`UI/src/routes/admin/workbench/components/UsageSection.svelte:10-21`): the `withRarity` snippet parameter is passed `true` at both call sites — dead branch flag.

## Minor consistency nits

- **`CacheSystem`/`PubSubSystem` default to Redis on unset config** (`Game.Abstractions/Infrastructure/Enums.cs`): both enums start at 0 = Redis, so a missing config silently selects Redis, while `DatabaseSystem` deliberately starts at 1 so an unset value fails loud (#453). The factories only reject *unrecognized* values. Practical impact near-zero (a missing connection string still throws in `RedisMultiplexerFactory`), but the =1 start is presented as the tier's policy — align or document the deviation.
[...]

PR #2362, opened 2026-07-23
Hygiene sweep: persist battle-end rejection clears, drop dead queue overloads, guard enemy-name fallback (#2345)

### LINKED ISSUE #2345: Hygiene sweep: idempotency-rejection state not persisted, stale doc comment, dead queue overload, unguarded enemy-name fallback, minor lint-path recompute
Small, independent nits from the scheduled health review — none warrants its own issue; batched like #2245.

1. **Battle-end idempotency rejection clears the battle in memory but never persists it.** `TryValidateBattleEndClaim`'s `BattleAlreadyCredited` backstop calls `state.ClearBattle()` and returns false (`Game.Application/Services/BattleService.cs:371-381`), but `DefeatEnemy.cs:73-81` / `BattleLost.cs:57-63` only call `SavePlayerStateAsync` on success — so the cache keeps showing the credited battle as active (the staleness the guard exists to converge; contrast `NewEnemy.cs:37`, which saves unconditionally). Self-heals one round-trip later and isn't exploitable, but flushing the clear on the rejection path makes the reconnect loop terminate in one step.
2. **Stale doc comment:** `OfflineProgressSummary.OpenedProficiencies` claims "each with the seed skill it granted (if any)" (`Game.Application/Services/OfflineProgressService.cs:429-431`), but `ProficiencyOpened` is `record(int ProficiencyId)` and opening is notification-only (seed-skill design re-homed in spike #1125). Comment survived the design change.
3. **Dead production surface:** `IPubSubQueue.GetNextAsync&lt;T&gt;` (`Game.Abstractions/Infrastructure/IPubSubQueue.cs:6`, `RedisQueue.cs:50-54`) has no production caller — `SocketManagerService.cs:456` deliberately uses the string overload (documented reason: a popped-before-deserialized item can't be recovered). Referenced only by test fakes that mostly throw `NotSupportedException`. Remove per the dead-code guideline.
4. **Unguarded enemy-name fallback:** `BattleEngine` spawn (`UI/src/lib/engine/battle/battle-engine.ts:201-213`) spreads `enemyData[enemyInstance.id]` without a missing-record guard — a freshly-authored enemy against a stale mid-session reference cache logs `undefined used Bite...`. Every sibling path guards this (`newItem`/`newItemMod` skip-and-log, `claimVictory`'s name guard, `Battler.fillSkills`); a `?? 'Unknown enemy'` or skip-and-log matches the house convention.
5. **Minor recompute on the content-lint path:** `CombatRating.Marginal` (`Game.Core/Battle/CombatRating.cs:81-86`) recomputes the unchanged baseline `Rate(battler)` once per authored attribute per enemy in `ProgressionGraphChecker.cs:286` (~2N ratings instead of N+1). Admin/content-health path only — fold in a baseline-accepting overload only if lint runtime ever matters.
6. **CI fixture coupling:** `run-tests.yml:326` hardcodes `psql -U game -d game -p 5544` while `docker-compose.yml` parametrizes `${POSTGRES_USER:-game}`/`${POSTGRES_PORT:-5544}` — use the same env expansions so an override doesn't fail the fixture step looking like a DB outage.

ginderjeremiah:
## Summary

Five of #2345's six independent nits, addressed in one PR since none warrants its own issue (matching #2245's precedent):

1. **Battle-end idempotency rejection now persists the cleared battle.** `TryValidateBattleEndClaim`'s `BattleAlreadyCredited` backstop calls `state.ClearBattle()` on a rejected already-credited claim, but `DefeatEnemy.cs`/`BattleLost.cs` only called `SavePlayerStateAsync` on the success branch — so the cache kept showing the credited battle as active until a later save happened to fire. Both commands now save unconditionally, mirroring `NewEnemy`'s existing unconditional save.
2. **Stale doc comment fixed.** `OfflineProgressSummary.OpenedProficiencies` claimed proficiency-opening grants "the seed skill it granted (if any)" — stale since the seed-skill design moved to skill synthesis (spike #1125); `ProficiencyOpened` is notification-only (`record(int ProficiencyId)`).
3. **Removed dead generic queue overloads.** `IPubSubQueue.GetNextAsync<T>`/`AddToQueueAsync<T>` have no production caller — `SocketManagerService` and every `Publish`/dead-letter path deliberately go through the string overload (a popped-before-deserialized item can't otherwise be recovered). Verified `AddToQueueAsync<T>` is equally dead (not called out in the issue, but the same pattern) and removed it alongside `GetNextAsync<T>`. The two integration tests that used them only as a serialize/deserialize convenience (`RedisPubSubServiceIntegrationTests`'s `PublishBatch<T>` coverage) now deserialize the popped string directly via the existing `string.Deserialize<T>()` extension; the one test whose entire purpose was pinning the now-removed typed round-trip (`RedisQueueIntegrationTests.TypedRoundTripAsync_...`) was removed.
4. **Guarded the unguarded enemy-name fallback.** `BattleEngine.reset` spread `enemyData[enemyInstance.id]` with no missing-record guard — a freshly-authored enemy against a stale mid-session reference cache fielded a battler with an `undefined` name. Now logs once (mirroring `InventoryManager.initialize`'s missing-item skip-and-log) and falls back to `'Unknown enemy'`, matching the house convention `claimVictory`'s name guard and `Battler.fillSkills` already use.
6. **CI fixture coupling fixed.** The e2e admin-fixture step hardcoded `psql -U game -d game -p 5544` while `docker-compose.yml` parametrizes `POSTGRES_USER`/`POSTGRES_DB`/`POSTGRES_PORT` — now uses the same env-var overrides so a customized stack doesn't fail this step looking like a DB outage.

**Item 5 skipped** (`CombatRating.Marginal` recomputing the baseline rating once per authored attribute in `ProgressionGraphChecker`): the issue's own text gates this behind "only if lint runtime ever matters," and there's no evidence the content-lint path is currently a problem, so implementing a baseline-accepting overload now would be premature optimization for an admin-only tool.

## Test plan

- [x] New `DefeatEnemy_BattleAlreadyCredited_PersistsClearedBattleState` / `BattleLost_BattleAlreadyCredited_PersistsClearedBattleState` integration tests reproduce the reconnect gap end-to-end over real sockets: win/lose a battle, restore the session to the now-credited battle, hit the command again, and assert the persisted session (not just the in-memory one) no longer shows it active.
- [x] New `battle-engine.test.ts` case pins the missing-enemy-id fallback: `'Unknown enemy'` name plus the debug log line.
- [x] `dotnet test` (full solution) — 3,465 tests passed (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests).
- [x] `npm run test:unit:ci` — 3,958 tests passed across 311 files.

Closes #2345

https://claude.ai/code/session_01CnB5wa2biSaRe11CR39t8u

---
_Generated by [Claude Code](https://claude.ai/code/session_01CnB5wa2biSaRe11CR39t8u)_

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -365,7 +365,9 @@ jobs:
 
         wait_for_api
         # Runs inside the Postgres container so it needs no client tooling on the runner; idempotent.
-        docker compose exec -T postgres psql -U game -d game -p 5544 -v ON_ERROR_STOP=1 -f - < e2e-admin-provision.sql
+        # Mirrors docker-compose.yml's own POSTGRES_USER/POSTGRES_DB/POSTGRES_PORT overrides so a customized
+        # stack doesn't fail this step looking like a DB outage.
+        docker compose exec -T postgres psql -U "${POSTGRES_USER:-game}" -d "${POSTGRES_DB:-game}" -p "${POSTGRES_PORT:-5544}" -v ON_ERROR_STOP=1 -f - < e2e-admin-provision.sql
 
     - name: Run Playwright e2e tests
       working-directory: UI
--- a/Game.Abstractions/Infrastructure/IPubSubQueue.cs
+++ b/Game.Abstractions/Infrastructure/IPubSubQueue.cs
@@ -3,7 +3,6 @@
     public interface IPubSubQueue
     {
         public Task<string?> GetNextAsync(CancellationToken cancellationToken = default);
-        public Task<T?> GetNextAsync<T>(CancellationToken cancellationToken = default);
 
         /// <summary>
         /// Atomically moves the next item from the head of this queue onto a side processing list and returns it
@@ -67,7 +66,6 @@
         public Task<long> RemoveRangeAsync(IReadOnlyList<string> values, CancellationToken cancellationToken = default);
 
         public Task AddToQueueAsync(string value, CancellationToken cancellationToken = default);
-        public Task AddToQueueAsync<T>(T value, CancellationToken cancellationToken = default);
 
         /// <summary>
         /// Pushes multiple values onto the queue in a single round-trip (one multi-value LPUSH),
--- a/Game.Api.Tests/Integration/BattleLostSocketTests.cs
+++ b/Game.Api.Tests/Integration/BattleLostSocketTests.cs
@@ -1,5 +1,7 @@
+using Game.Abstractions.DataAccess;
 using Game.Api.Models.Enemies;
 using Game.Api.Services;
+using Game.Core.Battle;
 using Game.Core.Players;
 using Game.Infrastructure.Database;
 using Game.TestInfrastructure.Fixtures;
@@ -27,7 +29,16 @@ namespace Game.Api.Tests.Integration
             await sessionService.SavePlayerStateAsync();
         }
 
-        private async Task<int> SeedWeakPlayerVsStrongEnemyAsync()
+        private async Task<PlayerState> GetPlayerState(int userId)
+        {
+            using var scope = CreateScope();
+            var sessionService = scope.ServiceProvider.GetRequiredService<SessionService>();
+            sessionService.SetAuthenticatedUser(userId);
+            await sessionService.LoadPlayerState();
+            return sessionService.PlayerState;
+        }
+
+        private async Task<(int userId, int playerId)> SeedWeakPlayerVsStrongEnemyAsync()
         {
             using var scope = CreateScope();
             var context = scope.ServiceProvider.GetRequiredService<GameContext>();
@@ -73,7 +84,7 @@ namespace Game.Api.Tests.Integration
                 new { Username = "weakplayer", Password = "weakpass" });
             Assert.Equal(HttpStatusCode.OK, loginResponse.StatusCode);
 
-            return user.Id;
+            return (user.Id, player.Id);
         }
 
         private async Task<int> SeedNormalPlayerAsync()
@@ -119,7 +130,7 @@ namespace Game.Api.Tests.Integration
         [Fact]
         public async Task BattleLost_ValidLoss_Succeeds()
         {
-            var userId = await SeedWeakPlayerVsStrongEnemyAsync();
+            var (userId, _) = await SeedWeakPlayerVsStrongEnemyAsync();
 
             var wsClient = Factory.Server.CreateWebSocketClient();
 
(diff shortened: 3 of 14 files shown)


ginderjeremiah (commented):
## Review: Hygiene sweep implementing 5 of 6 nits from #2345

**Verdict:** Approve _(submitted as a `COMMENT` because I'm the PR author — GitHub won't accept a gating event on one's own PR)_ — every item is faithfully implemented, well-scoped, and tested; nothing needs to change before merge.

### What I verified

- **Item 1 — unconditional save on the rejection path** (`DefeatEnemy.cs:79`, `BattleLost.cs:62`). Confirmed against `BattleService.TryValidateBattleEndClaim`: the `else` branch fires for three rejection reasons — `BattleAlreadyCredited` (calls `state.ClearBattle()`, so the save now durably flushes the clear ✓), the torn-state guard, and the duration/anti-cheat rejection. The latter two mutate no state, so the added save is a harmless no-op write-behind, and the "claimed too early" case is an edge/anti-cheat path rather than a hot loop — so saving on all rejections is cheap and matches `NewEnemy`'s deliberate unconditional-save. Correct call.
- **Item 3 — dead queue overloads.** Removal is complete across the interface, `RedisQueue`, and every test fake; dropping `using Game.Core;` from `RedisQueue.cs` is right since `Serialize`/`Deserialize` are no longer referenced there. The two integration tests that used the typed overload only as a serialize convenience now go through the string overload + `Deserialize<T>()`, and the test whose sole purpose was the typed round-trip was correctly removed.
- **Item 4 — enemy-name guard** (`battle-engine.ts:216`). Spreading `undefined` is a JS no-op, `name` falls back to `'Unknown enemy'`, and it logs once — matching the `InventoryManager.initialize` / `claimVictory` house convention.
- **Item 2, 6** — doc comment and CI env-var expansion are both correct and match `docker-compose.yml`'s parametrization.
- **Item 5 skip** — sound. The recompute is admin-only content-lint (~2N vs N+1 ratings) and the issue itself gates it behind "only if lint runtime ever matters." Deferring rather than gold-plating is the right choice.

### Test coverage

Meets the project's testing rules. The two new integration tests reproduce the reconnect gap end-to-end over real sockets — win/lose, credit, poll the durable write, restore the stale session, re-claim, and assert the *persisted* state (via a fresh `GetPlayerState`) no longer shows the battle active — which is exactly the persistence gap item 1 fixes. The frontend test pins both the `'Unknown enemy'` fallback and the debug log line.

### What's good

Careful, honest scoping: each nit traced back to its root cause, the dead-code removal went one step further than the issue (catching the equally-dead `AddToQueueAsync<T>`), and the item-5 deferral is reasoned rather than reflexive. Comments explain the *why* (the reconnect-loop convergence, the write-behind poll) without over-narrating.

---

PR #2363, opened 2026-07-24
Extract frontend accessibility catalogue into its own sub-doc

ginderjeremiah:
## Summary

`docs/frontend.md`'s Accessibility section (three dense sub-patterns: tooltip association via `aria-describedby`, keyboard-only focus re-anchoring, and the accessible-card pattern) was implementation-detail content useful mainly when actually building an accessible interactive component — the same profile that already justified splitting Tutorials, Sockets, Overlays, and Screens into their own sub-docs from this same file.

- Added `docs/frontend-accessibility.md` with the full catalogue, following the existing sub-doc convention (a one-line pointer back to `frontend.md`, then the content verbatim).
- Replaced the inline catalogue in `frontend.md`'s Accessibility section with a one-paragraph summary + link, matching how Tutorials/Sockets/Overlays already read.
- No production code touched — docs only.

Closes item 2 of #1982 ("evaluate whether the a11y catalogue is genuinely detail-level content that belongs in its own sub-doc, or whether it's better left inline"). Landed on extraction given the precedent already set by this file's other sub-docs and CLAUDE.md's own guidance to split out content that's "really useful but only when working on the feature."

Item 1 of #1982 (splitting the battle-attributes paragraphs into a `frontend-battle.md`) is **not** addressed here — the issue explicitly gates that on `frontend-battle.md` existing, and it doesn't yet, so there's no reason to create it in this PR. Leaving #1982 open for that half; happy to close it once `frontend-battle.md` exists and that item is picked up (or if the maintainer decides it's not worth doing separately).

## Test plan

- [x] Docs-only change; no build/test impact.
- [x] Grepped the repo for any other reference to `frontend.md#accessibility` or the extracted content — none found, so nothing else needed updating to point at the new location.

Addresses #1982

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qz1Z7n7a55rnKs5yvzjDUh)_


ginderjeremiah (commented):
## Review: extract the frontend.md accessibility catalogue into `frontend-accessibility.md` (item 2 of #1982)

**Verdict:** Approve _(posted as a comment since I'm also the PR author; GitHub won't accept a gating verdict on one's own PR)_ — clean, correctly-scoped docs-only extraction that matches the established sub-doc convention.

### Assessment

- **Follows the sub-doc convention exactly.** `frontend-accessibility.md` opens with an `# Heading` + a one-line "split from [frontend.md](./frontend.md#accessibility) so the main doc stays an overview" pointer, identical in shape to `frontend-tutorials.md`, `frontend-overlays.md`, and `frontend-screens.md`. The `#accessibility` anchor resolves — the `## Accessibility` heading is still present in `frontend.md`.
- **No content lost.** The three sub-patterns (tooltip association, keyboard-only focus re-anchoring, accessible-card) moved verbatim; `frontend.md` retains a condensed summary of the interactive-controls rule plus the link, mirroring how Tutorials/Sockets/Overlays already read.
- **No dangling references.** Grepped `docs/` for the moved identifiers (`describedByTooltip`, `focusAnchor`, `Accessible card pattern`) and the old anchor — nothing else points at the relocated content, so nothing else needed updating.
- **Scope discipline is right.** Item 1 (battle-attributes → `frontend-battle.md`) is correctly left out and #1982 stays open to track it.

### FYI (non-blocking, no change needed here)

The PR body states `frontend-battle.md` "doesn't exist yet" — it actually already exists on `main`. That doesn't affect this change (item 1 is a separate extraction and stays out of scope), but it does mean item 1 of #1982 is now technically unblocked and can be picked up as a follow-up under the existing issue whenever the maintainer wants it.

### What's good

Faithful extraction with the reverse link and anchor both intact, and the judgment call (extract vs. leave inline) is well-reasoned against the precedent set by the file's other sub-docs.

---

PR #1621, opened 2026-07-05
Suppress hidden-tab presentational side effects (combat floaters + victory toasts)

ginderjeremiah:
## Summary

Follow-up to #1594 (worker tick source), part of spike #1073. Once the sim runs at full rate while the tab is hidden, its per-tick presentational side effects run too, while `requestAnimationFrame` is suspended and page timers are throttled to ~once/minute:

- **Combat floaters** (`CombatFloaters.svelte`): `spawn()` now skips entirely while `document.hidden`. Previously each hit still pushed a floater into the DOM immediately, with only its removal `setTimeout` throttled — so a long hidden battle could accumulate a minute's worth of floater DOM before the throttled timers finally clear them, causing a stutter on refocus. They're purely presentational (`aria-hidden`), so skipping the spawn outright has no user-visible effect.
- **Victory-driven toasts** (challenge completions, proficiency milestones/level-ups/opens, newly-creatable recipes in `$lib/engine/engine.ts`): the same pileup risk applies — the tick worker can resolve far more battles per hidden minute than the throttled auto-dismiss timers can clear, so dozens of toasts could stack up over a long hidden session. A new `toastIfVisible` helper skips the toast while hidden but leaves the underlying state update and (for proficiency events) the permanent combat-log line untouched, so no information is lost — the player just doesn't see a popup for an event that happened while nobody was watching.

Audited but left unchanged, per the issue's own guidance:
- The boss-victory overlay's `delay(BOSS_VICTORY_OVERLAY_MS)` — stretching under throttling just slows hidden boss-farming slightly; acceptable.
- The combat log — already capped (`maxLogEntries`), no change needed.

No user-visible behavior change (the tab is hidden throughout); this is DOM-churn/battery hygiene so a long-hidden tab doesn't stutter on refocus.

## Test plan

- [x] New unit tests: `CombatFloaters.test.ts` (skip spawn while hidden, resume once visible) and `engine.test.ts` (challenge-completed/milestone/level-up/opened/recipe-available toasts all suppressed while hidden, underlying state updates and log lines still fire).
- [x] `npm run test:unit:ci` — full suite green (3445/3445).
- [x] `npm run lint` (eslint + svelte-check) — clean.
- [x] `npm run build` — production build succeeds.

Closes #1598

---
_Generated by [Claude Code](https://claude.ai/code/session_019r24cKkXG7j59AGzbo1aEE)_

Issue #1395, opened 2026-06-30
Add notification badges to relevant pages when things are unlocked


ginderjeremiah:
For example:

- New proficiency unlocked or leveled up
- New challenge completed
- New item unlocked
- New skill unlocked

Notifications will be best-effort and try to lean towards not being annoying. Notifications will probably be tracked client-side only, and "missing" notifications (for example, after clearing local storage or a first-time load) will not have any recovery feature.

PR #2365, opened 2026-07-24
Spike #1395: research and stage notification badges for unlocks

ginderjeremiah:
## Summary

- Docs-only: adds `docs/spikes/1395-notification-badges.md`, researching #1395 ("Add notification badges to relevant pages when things are unlocked").
- Finds the gap this issue targets is real and already observable today, not hypothetical: three of the four named events (proficiency level/milestone/open, challenge completion, a synthesis recipe becoming craftable) already toast via `engine.ts`'s `toastIfVisible`, which **drops the toast outright** (no queue, no replay) whenever the tab is hidden — the documented `#1621` convention. So "player was away and missed the notice" is the confirmed everyday behavior of the current toast path in an idle game, not an edge case.
- Surveys what's already reusable: the Help screen's existing `.unread-badge` nav pill (`NavSidebar.svelte`, currently hardcoded to `screen.key === 'help'`, server-backed via `PlayerLesson` state) is a working precedent for the badge affordance itself; the `#1586` content-events layer (`newlyTrue`/`newlyInSet`/`crossedThreshold`) was built with this issue in mind but is only actually consumed by tutorial mechanic-triggers today, not the existing toast diffs.
- Notes "item unlocked" and "skill unlocked" aren't fully independent detection surfaces: items only unlock via challenge completion (no separate event), and a synthesized skill is currently *deliberately* un-toasted (the player is already looking at the result when they click synthesize).
- Recommends a v1 scoped to **nav-level aggregate badges** (generalizing the Help pattern to Proficiencies/Challenges/Skills) with **client-side-only** persistence — arguing the issue's own client-only framing holds up here, unlike lesson state, because badge "seen" state gates no content and has no server consumer (a real contrast with #1392's toast/localStorage draft, which *was* reversed on collaborator review because lesson state is content-arc load-bearing). Per-item "NEW" tags are flagged as a legitimate, separately-scoped v2 rather than part of this pass.
- Leaves four explicit open questions for the maintainer (badge granularity, persistence, whether synthesis should start getting notified, whether Inventory needs its own badge distinct from Challenges) before implementation issues are filed — per this repo's spike convention, an ideation/spike issue gets a collaborator pass before scope is finalized.

Per `CLAUDE.md`'s spike convention, a spike's deliverable is research + a documented direction in `docs/spikes`, giving the maintainer a chance to weigh in before implementation begins — not code written speculatively against an unconfirmed design. This PR is that research pass; no implementation issues are filed pending the maintainer's read on the open questions in the doc.

Closes nothing (spike doc only) — leaving #1395 open pending maintainer confirmation of the recommended direction.

## Test plan

- [x] Docs-only change; no code touched, no build/test impact.
- [ ] Maintainer confirms (or redirects) the nav-badge-first scoping, the client-side-only persistence call, and the synthesis/Inventory open questions — then a follow-up issue scopes the actual v1 implementation.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ES8fLkE7Cp9uSLYYwS8r2h)_

### DIFF
--- a/docs/spikes/1395-notification-badges.md
+++ b/docs/spikes/1395-notification-badges.md
@@ -1,0 +1,131 @@
+# Spike #1395 — notification badges for unlocks
+
+## The question, as filed
+
+"Add notification badges to relevant pages when things are unlocked," naming four examples — a
+proficiency unlocked/leveled up, a challenge completed, an item unlocked, a skill unlocked — with
+three constraints already stated in the issue: best-effort (lean toward not annoying the player),
+probably client-side only, and no recovery feature for a badge lost to e.g. a cleared `localStorage`.
+
+## Current state (what already exists)
+
+- **Three of the four named events already toast, but only while the tab is visible.**
+  `challengeCompletedMessage` (challenge completed, naming its item/mod reward),
+  `proficiencyLevelMessage`/`proficiencyMilestoneMessage`/`proficiencyOpenedMessage` (proficiency
+  level/milestone/newly-opened, milestone naming any granted skill), and `recipeAvailableMessage`
+  (a synthesis recipe becoming craftable) are all wired in `lib/engine/engine.ts` through a shared
+  `toastIfVisible` helper (`engine.ts:89-94`) that **drops the toast outright** — no queue, no
+  replay — when `document.hidden`. This is the deliberate `#1621` convention documented in
+  `frontend.md` ("presentational side effects... are dropped, never queued or replayed"), and it
+  means the exact scenario this issue wants a badge for — the player was away/idling and missed the
+  notice — is not a hypothetical, it is the confirmed, everyday behavior of the existing toast path
+  for an idle game where the tab is often unfocused.
+- **A working nav-badge precedent already ships**, built for tutorial lessons (#1392/#1587):
+  `NavSidebar.svelte` renders a `.unread-badge` pill on a screen's rail entry, today hardcoded to
+  `screen.key === 'help'`, sourced from `playerManager.lessons.filter(l => !l.readAt).length` —
+  server-side per-player state. The visual affordance and its motion/styling are done; what's
+  hardcoded is the single screen key and the server-backed data source.
+- **The content-events derivation layer (#1586, merged)** — `lib/common/content-events.ts` — was
+  built with this issue in mind (its file header names "#1395's badges" as a future consumer
+  alongside tutorials) and already exports the generic primitives this needs: `newlyTrue` (a flip
+  false→true), `newlyInSet` (ids present in an after-set but not a before-set — exactly "what
+  newly unlocked"), and `crossedThreshold` (a level/milestone crossing). Today it is only actually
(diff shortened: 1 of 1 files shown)


ginderjeremiah (commented):
## Review: Spike #1395 — research & stage notification badges for unlocks (docs-only)

**Verdict:** Approve (posted as a `COMMENT` since I'm the PR author) — the research is sound, every technical claim checks out against the code, and it correctly follows the repo's spike convention. Not merging: per project convention spike results get a maintainer/collaborator pass before being closed out, and this PR deliberately leaves #1395 open with four open questions.

### What I verified
I spot-checked the doc's concrete code claims against the tree at HEAD, and they all hold:
- `toastIfVisible` (`engine.ts:89-94`) returns early when `document.hidden` — toasts are dropped, not queued — matching the `#1621` convention in `frontend.md`. ✅
- The three toast surfaces named (`challengeCompletedMessage`, `proficiency{Level,Milestone,Opened}Message`, `recipeAvailableMessage`) are all wired through `toastIfVisible` in `engine.ts`. ✅
- `content-events.ts` exports `newlyTrue` / `newlyInSet` / `crossedThreshold`, and its own header names "#1395's badges" as a future consumer — so the "built with this issue in mind" framing is accurate. ✅
- `NavSidebar.svelte` renders `.unread-badge` gated on `screen.key === 'help'`, sourced from `playerManager.lessons.filter(l => !l.readAt).length` — the "one hardcoded screen key + server-backed source" description is exact. ✅
- `synthesis-feedback.ts`'s header does state synthesis is "deliberately un-toasted," backing the claim that a synthesized skill is intentionally not notified today. ✅

### What's good
- The strongest part is the reasoning on **persistence**: it doesn't just cite the issue's "client-side only" framing, it re-derives it — and directly contrasts against #1392's lesson-state reversal, explaining *why* the same call flips the other way here (badge state gates no content and has no server consumer). That's exactly the kind of "don't assume the issue author analyzed it fully" scrutiny the repo asks for.
- Correctly collapses the issue's four named events into three real detection surfaces (item-unlock has no independent event; it rides challenge completion), and flags that "skill unlocked" is only partially covered — surfacing a genuine design question (synthesis) rather than papering over it.
- Scoping is disciplined: nav-aggregate v1 reusing the existing Help precedent, per-item "NEW" tags explicitly deferred to a separately-scoped v2 with a stated reason, not gold-plated in.
- Honest about the state of `content-events.ts` ("half-fulfilled shared layer") instead of overselling existing plumbing.

### Notes (non-blocking)
- The open questions are the right ones to gate on before filing implementation issues — no issues filed yet is correct per the spike convention (collaborator pass first).
- Minor: the doc's "the visual affordance and its motion/styling are done" is true of the Help pill, but generalizing to a per-screen count still needs the badge to render for a non-Help entry and the per-source watermark state the doc itself calls out — the doc does capture that under Detection wiring, so this is just a framing nuance, not a gap.

No blocking or should-fix findings. Leaving this for your read on the four open questions before implementation issues are scoped.

---

PR #2381, opened 2026-07-24
Fix CoalescedLoader.invalidate() discarding the boot fetch without rescheduling (#2366)

### LINKED ISSUE #2366: CoalescedLoader.invalidate() discards the boot fetch without rescheduling — a push mid-fetch strands a progress store on delta-only data (parity rejections, zones wrongly locked)
## Problem

Regression introduced by #2349 (commit 5d07d72). `CoalescedLoader.invalidate()` (`UI/src/lib/common/coalesced-loader.ts:75-78`) bumps the epoch and clears `freshQueued`, but never re-queues a fetch. Two consequences:

1. **A store that was not yet loaded is stranded on push-delta-only data.** The store fetchers early-return on `isStale` without setting `loaded = true` (e.g. `UI/src/stores/challenges.svelte.ts:19-26`), and nothing retries. After a push handler mutates the store and calls `invalidate()` (`challenges.svelte.ts:76`, `proficiencies.svelte.ts:112`, `statistics.svelte.ts:75`) while the boot fetch is in flight, the discarded response was the entire baseline — the store is left containing only the pushed delta with `loaded = false`.
2. **A queued force is silently dropped.** A `load(force)` already queued behind the in-flight fetch aborts at the `epoch !== chainEpoch` check (`coalesced-loader.ts:47`) and resolves without any fetch issued at or after its call — breaking the class's own documented guarantee that `resolveBossVictory`'s gate re-check (`enemy-manager.ts:715`) and `handleServerCommandFailed`'s divergence recovery (`engine.ts:145-148`) depend on.

## Failure scenario

Player refreshes the page moments after a victory. The `ChallengeCompleted`/`ProficiencyXpGained` pushes ride the Redis backplane and can be delivered on the freshly-reconnected socket while boot's `GetPlayerChallenges`/`GetPlayerProficiencies` fetches are still in flight. The push mutates the store and invalidates; the boot response is discarded.

- **Challenges:** the store holds only the one pushed challenge, so every previously-completed gating challenge reads as incomplete and **all previously-unlocked zones render locked** until a screen mount happens to call `load()` again.
- **Proficiencies:** `battleModifiers` derives from only the pushed rows, so the live battler assembles **missing most proficiency attribute bonuses → the server's parity replay rejects every subsequent `DefeatEnemy` claim** until the Proficiencies/Synthesis screen forces a reload.

Before #2349 this window merely risked reverting the single pushed item; the fix replaced that with discarding the whole baseline.

## Suggested fix

`invalidate()` (or the store) should re-queue a fetch when it fires while a fetch is in flight — at minimum when `loaded` is false or a force was queued — so the guarantee "a later load never overwrites the push" is kept without abandoning the baseline. Extend the loader unit tests with the push-during-boot-fetch and force-queued-then-invalidated cases.

ginderjeremiah:
## Summary

- `CoalescedLoader.invalidate()` (`UI/src/lib/common/coalesced-loader.ts`) bumped the epoch to mark a now-stale in-flight fetch as untrustworthy, but never queued a replacement fetch. Two consequences, both described in #2366:
  1. A store that hadn't finished its first load yet (e.g. challenges/proficiencies during boot) was left stranded on push-delta-only data forever — nothing else ever retries a store that already holds *some* data, so `loaded` never becomes `true`.
  2. A `force` already queued behind the in-flight fetch was silently aborted by the same epoch bump, so its caller (e.g. `resolveBossVictory`'s gate re-check, `handleServerCommandFailed`'s divergence recovery) resolved without ever getting fresh data.
- Fixed by splitting epoch tracking into two counters:
  - `epoch` — bumped by both `reset()` and `invalidate()`, used only for the external `isStale()`/`currentEpoch` contract `fetchFn` relies on.
  - `resetEpoch` — bumped **only** by `reset()`, used internally to decide whether a queued chain should abort. A queued chain (from a `force`, or from `invalidate()`'s own re-queue) now only aborts on a genuine session reset, never on a racing `invalidate()`.
  - `invalidate()` now queues a fresh fetch behind the in-flight one whenever the store isn't loaded yet and no fetch is already queued, so the store still reaches a real baseline instead of being stuck on the pushed delta.
- Extended `coalesced-loader.test.ts` with the two scenarios from the issue (invalidate-while-unloaded queues a fresh fetch; a queued force survives a racing invalidate) and updated the two tests whose asserted behavior was the bug itself.
- Updated the three store-level "not reverted by a fetch already in flight" regression tests (`challenges.test.ts`, `statistics.test.ts`, `proficiencies.svelte.test.ts`) to mock a realistic response for the fetch this fix now queues (any fetch issued after a push already reflects the server-persisted state the push described), and to wait for the store to actually reach `loaded` before asserting.

Implements #2366's suggested fix directly: `invalidate()` re-queues a fetch when it fires while a fetch is in flight, specifically covering the "not yet loaded" case and the "force already queued" case, without discarding the baseline the way the original bug did.

Closes #2366

## Test plan

- [x] `npx vitest run` (full frontend suite) — 3959 tests passed, 0 failed.
- [x] New/updated unit tests specifically pin: invalidate() while unloaded queues a fresh fetch and a concurrent load coalesces onto it; invalidate() leaves an already-loaded store's in-flight/queued fetch alone; a queued force is not aborted by a racing invalidate() and still resolves with genuinely fresh data.

---
_Generated by [Claude Code](https://claude.ai/code/session_011sZVAuevuF4UUUvFg5mG8K)_

### DIFF
--- a/UI/src/lib/common/coalesced-loader.ts
+++ b/UI/src/lib/common/coalesced-loader.ts
@@ -14,10 +14,15 @@
 export class CoalescedLoader {
 	/** Tail of the current load pipeline; concurrent callers await it instead of re-fetching. */
 	private inFlight: Promise<void> | undefined;
-	/** Whether a fresh (post-force) fetch is already queued behind the in-flight one. */
+	/** Whether a fresh fetch is already queued behind the in-flight one (via a force or {@link invalidate}). */
 	private freshQueued = false;
-	/** Bumped by {@link reset} so a queued fresh fetch from a discarded session never fires. */
+	/** Bumped by {@link reset} and {@link invalidate}; exposed via {@link currentEpoch}/{@link isStale}
+	 *  so `fetchFn` can recognize its own in-flight response as no longer trustworthy. */
 	private epoch = 0;
+	/** Bumped only by {@link reset}. A queued fresh fetch aborts on this moving (a discarded session)
+	 *  but, unlike {@link epoch}, does NOT move on {@link invalidate} — an invalidation must not abort
+	 *  a chain that still owes real data to a `loaded` store or a waiting `force` caller. */
+	private resetEpoch = 0;
 
 	/** `fetchFn` must never reject — the stores' fetchers catch and record errors internally. */
 	constructor(
@@ -40,17 +45,7 @@ export class CoalescedLoader {
 		if (force && !this.freshQueued) {
 			// The in-flight fetch predates this force, so its response may be stale — queue one
 			// fresh fetch behind it; later forces coalesce onto that queued fetch.
-			this.freshQueued = true;
-			const chainEpoch = this.epoch;
-			this.track(
-				this.inFlight.then(() => {
-					if (this.epoch !== chainEpoch) {
-						return;
-					}
-					this.freshQueued = false;
-					return this.fetchFn();
-				})
-			);
+			this.queueFreshFetch(this.inFlight);
 		}
 		await this.inFlight;
 	}
@@ -60,6 +55,7 @@ export class CoalescedLoader {
 		this.inFlight = undefined;
 		this.freshQueued = false;
 		this.epoch += 1;
+		this.resetEpoch += 1;
 	}
 
 	/** Bumps the epoch so an already-in-flight fetch's eventual response is recognized as stale by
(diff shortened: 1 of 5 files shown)


ginderjeremiah (commented):
## Review: Fix `CoalescedLoader.invalidate()` discarding the boot fetch without rescheduling (#2366)

**Verdict:** Approve — correctly fixes the #2349 regression that stranded not-yet-loaded progress stores on push-delta-only data (and silently dropped a queued force), with tests pinning both failure modes. Submitted as a `COMMENT` because I'm the PR author; the real call is approve.

### Correctness

The two-counter split is the right shape and is sound:

- `epoch` (bumped by both `reset()` and `invalidate()`) keeps driving the external `isStale()`/`currentEpoch` contract the store fetchers rely on to recognize a push-during-fetch — unchanged behavior.
- `resetEpoch` (bumped only by `reset()`) now gates whether a queued chain aborts, so an `invalidate()` racing a queued fetch no longer cancels data the chain still owes a `loaded` store or a waiting `force` caller.
- `invalidate()` re-queues a fresh fetch only when it can actually strand the store (`inFlight && !isLoaded() && !freshQueued`), which is exactly the gap the issue describes — a push landing before the first load completes.

Two subtler points I checked and that hold up:

- **No `freshQueued` stranding on abort.** The early-return in `queueFreshFetch` skips clearing `freshQueued`, but the only thing that moves `resetEpoch` is `reset()`, which already clears `freshQueued` itself (`coalesced-loader.ts:56`). So the flag can't get stuck — this was the original bug's mechanism and it's now structurally avoided.
- **Self-healing.** `freshQueued` is cleared before `fetchFn()` runs inside the chain, so a second push mid-fresh-fetch re-queues again and the store still converges to `loaded`.

Confirmed against the store fetcher (`challenges.svelte.ts:14-34`): the queued fresh fetch captures the post-`invalidate` epoch, so absent a further push it isn't stale and sets `loaded = true` with a real baseline.

### Test coverage

Adequate and well-targeted. The loader suite now covers both scenarios named in the issue — invalidate-while-unloaded queues a fresh fetch (and a concurrent `load()` coalesces onto it), and a queued force survives a racing `invalidate()` — plus the loaded-store case where `invalidate()` must *not* pile on an extra fetch. The three store regression tests (`#2332`) are updated to mock the now-queued fetch with server-persisted state and wait for `loaded`; note this also strengthens them, since the old buggy behavior would now time out at `waitFor(loaded === true)` rather than silently passing. Frontend-only change (not battle logic), so no backend mirror is required, and tests mirror source under `tests/lib/common/`.

### What's good

- Clean, minimal fix that maps directly onto the issue's suggested direction without over-reaching.
- The expanded JSDoc on `epoch`/`resetEpoch`/`invalidate`/`queueFreshFetch` is appropriately thorough for genuinely subtle concurrency logic — it documents *why* the second counter exists and what must not abort a chain, which is exactly the kind of thing a later "simplification" would otherwise break.

---

PR #2382, opened 2026-07-24
Retroactively evaluate satisfied Max/Min challenges on every connect (#2367)

### LINKED ISSUE #2367: Max/Min-backed statistic challenges never evaluate retroactively — an already-satisfied challenge authored or reinstated later can never complete
## Problem

`SetMax`/`SetMin` (`Game.Core/Progress/PlayerProgress.cs:371-406`) add a statistic key to `_touchedThisBattle` only on first write or **strict improvement**, and `ChallengeIndex.RelevantTo` (`Game.Core/Progress/ChallengeIndex.cs:59`) — the only challenge-evaluation entry point (`ChallengeRewardService.EvaluateAndApply`) — evaluates only challenges keyed to touched statistics (plus statistic-independent ones).

So a challenge backed by a Max/Min statistic that is **already satisfied when it enters circulation** (newly authored, or reinstated from retirement) is never evaluated until the player strictly beats their own record:

- Player's `HighestSingleAttackDamage` is 800; a "hit for ≥ 500" challenge is authored afterwards. Every later battle whose best hit is ≤ 800 leaves the key untouched (`created || value > stat.Value`), so the challenge stays incomplete forever unless the player exceeds 800.
- Player's `FastestVictory` best is 8s; a "win within 10s" time trial is authored later. A 12s win doesn't touch the key; the qualifying 8s record never re-enters evaluation.

Sum-backed challenges self-heal on the next non-zero delta; only Max/Min-backed ones stall. There is no other evaluation sweep (verified — no login/load-time pass).

## Suggested direction

Some trigger needs to evaluate statistic-backed challenges against the player's *existing* records, e.g. a one-time sweep at player load (cheap — the progress aggregate and challenge index are both in memory), or an evaluation pass keyed to reference-cache reload for newly-circulating challenges. The load-time sweep is simpler and also covers the reinstatement case; it should reuse `EvaluateAndApply` so rewards/notifications flow through the normal path (probably with `notify` on — the player did earn it).

Add domain tests: already-satisfied Max/Min challenge completes on the sweep; unimproved record still doesn't re-evaluate per-battle (the existing scoping is a deliberate optimization worth keeping).

ginderjeremiah:
## Summary

- `SetMax`/`SetMin` (`Game.Core/Progress/PlayerProgress.cs`) only add a statistic to the touched set on a first write or a strict improvement, and challenge evaluation (`ChallengeIndex.RelevantTo`) is scoped to touched statistics. So a Max/Min-backed challenge that's already satisfied when it enters circulation — newly authored, or reinstated from retirement — could never complete until the player beat their own record.
- Added `PlayerProgress.StatisticKeys`: every (statistic type, entity) key the player currently holds a recorded row for, as opposed to `RecordBattleCompleted`'s narrower per-battle touched-key return.
- `GetOfflineProgress` runs once per socket connect regardless of away time, so it's the natural once-per-connect load-time hook. Both branches of `OfflineProgressService.SimulateProgress` now evaluate against `progress.StatisticKeys` (the full recorded set) instead of just the touched subset:
  - The sub-5-minute reconnect branch (previously a pure no-op beyond re-anchoring `LastActivity`) now loads progress, runs the sweep, and notifies live (`notify: true`) since there's no welcome-back summary on this branch to carry the completion.
  - The full offline-window branch folds the retroactive sweep into the same evaluation pass as the window's own battles (`notify: false`, consistent with "the summary is the notification").
- Re-evaluating an already-completed challenge is a no-op (`EvaluateChallenges` skips it), so widening the scope from touched-this-window to the player's full recorded set cannot double-apply a reward the window's own battles already earned.
- The per-battle live path (`BattleStatisticsEventHandler`) is untouched — it deliberately keeps the narrower touched-key scoping (the documented optimization the issue itself asked to preserve).
- Documented the new behavior in `docs/backend-battle.md` → Offline rewards simulation.

Implements #2367's suggested direction exactly: a one-time (per-connect) sweep at player load, reusing `ChallengeRewardService.EvaluateAndApply` so rewards/notifications flow through the normal path.

## Test plan

- [x] `dotnet test` (full solution) — 3,473 tests passed (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests).
- [x] New domain tests in `Game.Core.Tests/Progress/PlayerProgressTests.cs`: `StatisticKeys` returns every recorded row, and a Min-backed challenge already satisfied by an existing record (never touched this session) completes when evaluated via `StatisticKeys` + `ChallengeIndex.RelevantTo`, without re-firing an already-completed one.
- [x] New integration tests in `Game.Application.Tests/Services/OfflineProgressServiceIntegrationTests.cs`:
  - Below-threshold reconnect still completes an already-satisfied time-trial challenge and unlocks its reward.
  - A full multi-battle offline window completes a challenge satisfied by a pre-existing record no battle in the window could ever touch (the record is pinned to stay unchanged).

Closes #2367

https://claude.ai/code/session_01P32Qwx36Wi28XzHCbUjmQR

---
_Generated by [Claude Code](https://claude.ai/code/session_01P32Qwx36Wi28XzHCbUjmQR)_


ginderjeremiah (commented):
## Review: Retroactive Max/Min challenge sweep on connect (#2367)

**Verdict:** Approve (posted as `COMMENT` — I'm the PR author, so GitHub won't accept a gating event from me) — one should-fix (dead code) worth clearing before merge; the design and test coverage are otherwise solid.

The change faithfully implements #2367's suggested direction: `PlayerProgress.StatisticKeys` exposes the player's full recorded-statistic key set, and both branches of `OfflineProgressService.SimulateProgress` evaluate against it through the existing `ChallengeRewardService.EvaluateAndApply` rather than only the per-battle touched subset. I verified the two load-bearing claims:

- **No double-apply.** `PlayerProgress.EvaluateChallenges` skips any challenge already `Completed` (`PlayerProgress.cs:160-163`), so widening the scope to the full recorded set is genuinely a no-op for challenges the window's own battles already earned.
- **The scope actually surfaces the stalled challenge.** `ChallengeIndex.RelevantTo` buckets by `(statistic, entity)` key, so feeding it every recorded key (not just the strictly-improved ones `SetMax`/`SetMin` touch) is exactly what reaches an already-satisfied Max/Min challenge. The narrow per-battle live path (`BattleStatisticsEventHandler`) is correctly left untouched.

Branch split is well-reasoned: the sub-threshold branch notifies live (`notify: true`, no welcome-back gate to carry it) while the full-window branch folds into the summary (`notify: false`). Timestamps are consistent per branch (`now` vs. `awayWindowEnd`), and the below-threshold work is correctly wrapped in a shared `BeginBatch` mirroring the full-window path's atomicity. This is backend-authoritative challenge evaluation, so no frontend battle-parity mirror is required.

### Should-fix

- `OfflineRewards.Empty` (`Game.Application/Services/OfflineProgressService.cs:509`) is now dead code. This PR removed its only caller — the `if (result.BattlesSimulated == 0) return OfflineRewards.Empty;` early return in `ApplyOfflineRewards` — and nothing else references it (repo-wide grep for `OfflineRewards.Empty` returns zero hits). CLAUDE.md: "If you are removing or replacing functionality, make sure to remove any code that is no longer being used as well." Drop the static `Empty` property (and its doc comment) in this PR.

### Nits

- Below-threshold branch: `EvaluateAndApply` now runs `_progressRepo.Save` unconditionally on every sub-5-minute reconnect even when no challenge completed (no aggregate mutation). It's once-per-connect and bounded, so not worth guarding — noting only in case the reconnect path ever gets hotter.

### Test coverage

Good, and it targets the exact regression on both sides of the branch:
- Domain (`PlayerProgressTests`): `StatisticKeys` returns every recorded row / is empty; the key set fed to `ChallengeIndex.RelevantTo` surfaces an already-satisfied challenge the empty touched-set never would; and an already-completed challenge is not re-completed (locking in the no-double-apply invariant).
- Integration (`OfflineProgressServiceIntegrationTests`): the below-threshold branch completes and unlocks an already-satisfied time-trial with no window replayed; a multi-battle away window completes a challenge whose backing record (pinned at 0.001s) no simulated battle could ever touch.
- Minor gap (not blocking): no assertion that the below-threshold branch notifies *live* (`notify: true`) — the test checks durable effects and the empty summary, but not the client push itself.

### What's good

- Reuses the existing evaluation/reward path instead of adding a parallel sweep, so rewards and notifications flow through exactly one code path.
- The `StatisticKeys` XML doc and the two in-code comments clearly explain *why* the scope widens here but deliberately stays narrow on the hot per-battle path — this is the kind of "don't naively simplify me" context that prevents a future regression.
- `docs/backend-battle.md` is updated accurately and in the right section.

---

ginderjeremiah:
Thanks for the review! Fixed the should-fix in cef56f0: removed the now-dead `OfflineRewards.Empty` static property along with its doc comment, since its only caller was the `BattlesSimulated == 0` early return this PR already removed. Build and the full `OfflineProgressServiceIntegrationTests` suite pass.

Left the two non-blocking notes as-is (the unconditional `Save` on the below-threshold reconnect is bounded/cheap as noted, and asserting the live `notify: true` push would need pub/sub test infrastructure disproportionate to what the existing durable-effects assertions already cover).

---

ginderjeremiah:
## Re-review of the delta since my last review (`cef56f0a`)

The only change since my prior review is the follow-up commit resolving the should-fix I raised: `OfflineRewards.Empty` is removed. Verified:

- The static `Empty` property is gone; `OfflineRewards` is now a plain positional record with no body.
- Its only caller — the `if (result.BattlesSimulated == 0) return OfflineRewards.Empty;` early return in `ApplyOfflineRewards` — was already removed as part of always running the retroactive sweep, and no other reference remains.
- Nothing else changed; the design, branch split, and test coverage I validated on `c67e4fbd` are unaffected.

**Verdict: Approve.** Should-fix resolved, no outstanding nits (the sub-threshold unconditional `Save` was already judged not worth guarding), CI green (`all-checks-passed` ✓). Merging.

---

PR #2383, opened 2026-07-24
Fix ensureValidAccessToken discarding a still-valid token on retryable refresh failures

### LINKED ISSUE #2368: ensureValidAccessToken discards a still-valid access token when a pre-emptive refresh fails retryably
## Problem

`UI/src/lib/api/auth.ts:141-147`: inside the 30s pre-expiry leeway window the stored access token is still valid (`expiry > now`), but when the pre-emptive refresh returns a **retryable** outcome (the auth rate-limiter's 429, a transient 5xx, or a network blip on the refresh endpoint alone), `ensureValidAccessToken` returns `accessToken: null` instead of falling back to `getAccessToken()`.

`api-request.ts` then sends the request with **no** Authorization header — a guaranteed 401 — re-attempts the refresh, and surfaces the 401 to the caller: a self-inflicted failure while a working token sat in storage. `api-socket.ts` similarly bails a reconnect that would have succeeded, stranding the queue until the next 10s keepalive.

## Failure scenario

Two tabs plus a socket reconnect trip the refresh rate limiter during the final 30s of a 15-minute token's life — every authenticated call in that window fails, recurring each token cycle.

## Fix

On a retryable outcome, return the stored token when it hasn't actually expired (one-line fallback); only report no token when it is genuinely expired or the refresh was definitively rejected. Unit-test the retryable-outcome-inside-leeway case.

Related (same resilience theme, separate defect): the socket-side retry-budget issue filed alongside this one.

ginderjeremiah:
## Summary

- `ensureValidAccessToken` (`UI/src/lib/api/auth.ts:137-150`) refreshes pre-emptively whenever the stored access token is inside its 30s expiry leeway. When that pre-emptive refresh returned a **retryable** outcome (the auth rate-limiter's 429, a transient 5xx, or a network blip on the refresh endpoint alone), it unconditionally returned `accessToken: null` — even though the stored token was still genuinely valid (`expiry > now`, just inside the leeway window).
- `api-request.ts` then sent the request with no `Authorization` header — a guaranteed 401 — and `api-socket.ts` similarly bailed a reconnect that would otherwise have succeeded.
- Fixed by falling back to the stored (still-valid) access token when the refresh outcome is `retryable` and the token hasn't actually expired yet. `rejected`/`null` is still reported when the refresh is definitively rejected, or when the stored token really has expired.

Implements the issue's suggested one-line fallback exactly: only report "no token" when the stored token is genuinely expired or the refresh was definitively rejected.

## Test plan

- [x] Existing `UI/src/tests/lib/api/auth.test.ts` test `reports retryable (not rejected) without a token on a network error` was pinning the buggy behavior (it used a token 10s from expiry — still valid — and asserted `accessToken: null`). Replaced it with two tests: one confirming the fallback now returns the still-valid stored token on a retryable failure, and one confirming a **genuinely expired** token still correctly returns `null`/not-rejected on a retryable failure.
- [x] `npm run test:unit:ci` — full frontend suite green, 3960/3960 tests passed.

Closes #2368

https://claude.ai/code/session_01MN8xDjxcLK9yrZCdpYEVPJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01MN8xDjxcLK9yrZCdpYEVPJ)_

### DIFF
--- a/UI/src/lib/api/auth.ts
+++ b/UI/src/lib/api/auth.ts
@@ -140,10 +140,17 @@ export const ensureValidAccessToken = async (): Promise<AccessTokenResult> => {
 
 	if (expiry === null || expiry - EXPIRY_LEEWAY_SECONDS <= nowSeconds) {
 		const outcome = await refreshTokens();
-		return {
-			accessToken: outcome.status === 'success' ? outcome.tokens.accessToken : null,
-			rejected: outcome.status === 'rejected'
-		};
+		if (outcome.status === 'success') {
+			return { accessToken: outcome.tokens.accessToken, rejected: false };
+		}
+
+		// A retryable failure (rate limit, transient 5xx, network blip) doesn't invalidate the stored
+		// token — if it hasn't actually expired yet, use it rather than forcing a guaranteed 401.
+		if (outcome.status === 'retryable' && expiry !== null && expiry > nowSeconds) {
+			return { accessToken: getAccessToken(), rejected: false };
+		}
+
+		return { accessToken: null, rejected: outcome.status === 'rejected' };
 	}
 
 	return { accessToken: getAccessToken(), rejected: false };
--- a/UI/src/tests/lib/api/auth.test.ts
+++ b/UI/src/tests/lib/api/auth.test.ts
@@ -216,8 +216,23 @@ describe('auth', () => {
 			expect(result).toEqual({ accessToken: null, rejected: true });
 		});
 
-		it('reports retryable (not rejected) without a token on a network error', async () => {
-			setTokens({ accessToken: makeAccessToken(10), refreshToken: 'r' });
+		it('falls back to the stored access token on a retryable failure when it has not actually expired', async () => {
+			const token = makeAccessToken(10);
+			setTokens({ accessToken: token, refreshToken: 'r' });
+			vi.stubGlobal(
+				'fetch',
+				vi.fn(async () => {
+					throw new TypeError('Failed to fetch');
+				})
+			);
+
+			const result = await ensureValidAccessToken();
+
+			expect(result).toEqual({ accessToken: token, rejected: false });
+		});
+
+		it('reports retryable (not rejected) without a token when the stored token has actually expired', async () => {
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: fix `ensureValidAccessToken` discarding a still-valid token on retryable refresh failures (#2368)

**Verdict:** Approve — implements the issue's fix exactly, with correct handling of the retryable-vs-rejected-vs-expired distinction and well-targeted tests. (Posting as a comment since I'm the PR author; the intended verdict is approve.)

The change replaces the old blanket "any non-success ⇒ `accessToken: null`" with a precise fallback: on a `retryable` outcome, return the stored token when it hasn't actually expired; keep returning `null` only when the refresh was definitively `rejected` or the token has genuinely expired.

### What's good

- **Correct and consistent with `refreshTokens`.** `retryable` is returned *only* when no adoption happened (adoption of another tab's rotated pair returns `success`) and storage was not cleared (only `rejected` clears). So `getAccessToken()` in the fallback path reliably re-reads the same still-valid token — no torn/stale-state risk against the just-awaited refresh.
- **The `expiry !== null && expiry > nowSeconds` guard is the right predicate** — it distinguishes "inside the 30s leeway but still valid" (the whole point of the fix) from "actually expired", so a genuinely dead token still surfaces as `null`.
- **The 2-line comment earns its place** — it explains *why* the fallback exists, which guards against a later naive "simplification" back to the old ternary that would reintroduce the bug.
- Layering is correct: all token handling stays in `lib/api`, per `frontend.md` → Authentication.

### Test coverage

Solid and precisely scoped. The prior test was pinning the buggy behavior (still-valid token, asserted `null`); it's replaced by two that bracket the boundary:
- retryable failure + token still valid (10s to expiry, inside leeway) → returns the stored token, `rejected: false`.
- retryable failure + token genuinely expired (−10s) → returns `null`, `rejected: false`.

Both assertions match the implementation, and the existing `rejected` (400) test still guards the definitive-rejection path. No mirrored backend test is needed — this is frontend-only token-lifecycle logic, not battle parity.

Nothing blocking, no should-fixes, no nits. Clean fix.

---

PR #2384, opened 2026-07-24
Skip auth refresh on a pre-open socket close when the token was demonstrably fresh (#2369)

### LINKED ISSUE #2369: Any never-opened socket close spends the auth-retry budget — a non-auth WebSocket failure with working HTTP force-logs-out the user in seconds
## Problem

`UI/src/lib/api/api-socket.ts:452-499` (`handleClose`, pre-open branch): an auth-rejected handshake and a connection-level failure both close as 1006 and cannot be distinguished, and the branch refreshes unconditionally — even when the token it just presented was minted seconds earlier by the previous cycle. A successful rotation that still fails to open spends budget (`socketAuthRetries++`).

If `/api` is reachable over HTTP but WebSocket upgrades fail (a proxy/AV stripping `Upgrade`, an edge misconfiguration after a deploy), each cycle is: handshake fails → successful refresh rotation → retry → fail. After `MAX_SOCKET_AUTH_RETRIES` rapid cycles, `handleAuthFailure()` clears the tokens and bounces to login — **and the loop repeats on every subsequent login**, hard-locking the user out of an account with perfectly valid credentials.

`docs/frontend-sockets.md` frames the bound as stopping token burn on a server "that keeps rejecting a freshly-refreshed token", but the code triggers it on any pre-open close.

## Suggested fix

Skip the refresh (or apply backoff instead of a terminal logout) when the presented access token is demonstrably fresh/unexpired — a fresh token failing the handshake is evidence the failure is *not* auth. Terminal `handleAuthFailure` should be reserved for the case the docs describe (a rotation that was itself rejected, or a fresh token explicitly rejected). Unit-test the "WS blocked, HTTP fine" loop to pin that tokens survive.

Related (same resilience theme, separate defect): the `ensureValidAccessToken` retryable-fallback issue filed alongside this one.

ginderjeremiah:
## Summary

- `api-socket.ts`'s `handleClose` pre-open branch couldn't distinguish an auth-rejected handshake from a connection-level failure — both close as 1006 — and refreshed the token pair unconditionally, even when the token it just presented was minted seconds earlier by the previous retry cycle.
- If `/api` is reachable over HTTP but WebSocket upgrades fail (a proxy/AV stripping `Upgrade`, an edge misconfiguration), each cycle was: handshake fails → successful refresh rotation → retry → fail. After `MAX_SOCKET_AUTH_RETRIES` rapid cycles, `handleAuthFailure()` cleared the tokens and bounced to login — and the loop repeated on every subsequent login, hard-locking the user out of an account with perfectly valid credentials.
- `handleClose` now records the `exp` of the access token presented in the most recent handshake attempt (`handshakeTokenExpiry`, set in `openSocket`) and checks it before treating a pre-open close as an auth rejection: a token with well more than `FRESH_TOKEN_MARGIN_SECONDS` (120s) of remaining life failing the handshake is evidence the failure isn't about that token being stale — refreshing again would just burn another single-use refresh token without fixing the real problem.
- When the presented token is demonstrably fresh, no refresh is attempted and no auth-retry budget is spent — the keepalive ping's next `ensureSocket()` tick (10s later) retries the connect, the same backoff a retryable refresh failure already gets. Tokens are left untouched.
- A token with no decodable expiry (non-JWT/malformed) is treated as no evidence either way and still goes through the existing refresh/retry path — the safe default.
- A token genuinely near (or past) expiry is unaffected: it's still treated as a plausible auth rejection and goes through the normal bounded refresh/reconnect loop.
- Documented the new check in `docs/frontend-sockets.md`.

Implements the issue's suggested fix directly: skip the refresh when the presented access token is demonstrably fresh/unexpired, reserving the terminal `handleAuthFailure` path for an actual rotation rejection or a token genuinely near expiry.

## Test plan

- [x] New unit tests in `UI/src/tests/lib/api/api-socket.test.ts` (`handleClose token freshness`):
  - A demonstrably fresh token's pre-open rejection does not refresh, does not spend budget, does not touch stored tokens.
  - The keepalive ping still reconnects 10s later without spending budget or calling `refreshTokens`.
  - A token genuinely near expiry (10s remaining) still refreshes and reconnects normally.
  - A token with no decodable expiry still goes through the existing refresh/retry path.
- [x] `npx vitest --run` (full UI suite) — 311 test files, 3964 tests passed.

Closes #2369

---
_Generated by [Claude Code](https://claude.ai/code/session_01DD2UcN8dG2yuCuDV7z1AGR)_


ginderjeremiah (commented):
## Review: skip the auth refresh on a pre-open socket close when the presented token was demonstrably fresh (#2369)

**Verdict:** Approve — the fix implements the issue's suggested direction directly and correctly, with the token-freshness gate placed exactly where the ambiguous 1006 pre-open close is handled, and the four relevant cases are all tested. (Posted as a `COMMENT` because I'm the PR author and GitHub won't accept a gating self-review; the intended verdict is **Approve**.)

### Does it solve the problem?

Yes. The root cause in #2369 is that `handleClose`'s pre-open branch treated every 1006 close as an auth rejection and refreshed unconditionally, so a WS-upgrade failure with working HTTP burned the whole `MAX_SOCKET_AUTH_RETRIES` budget and force-logged-out an account with valid credentials. The fix records the presented token's `exp` in `openSocket` (`handshakeTokenExpiry`, `api-socket.ts:186`) and, before spending any budget, checks `wasHandshakeTokenDemonstrablyFresh()` (`:481`): a token with well over `FRESH_TOKEN_MARGIN_SECONDS` (120s) of life left failing the handshake can't plausibly be an expiry rejection, so it leaves the tokens and queue untouched and lets the keepalive's next `ensureSocket()` retry — exactly the terminal-logout-avoiding behavior the issue asked for.

### Correctness details I verified

- **Wall-clock vs. monotonic:** `wasHandshakeTokenDemonstrablyFresh()` compares `Date.now() / 1000` against the Unix-seconds `exp` claim — correct; `performance.now()` would have been wrong here.
- **Undecodable expiry → safe default:** `handshakeTokenExpiry === null` returns `false` (not "unfresh"), so a non-JWT/malformed token still flows through the existing refresh/retry path rather than silently skipping it. Good — that's the conservative choice.
- **Capture is race-free:** `this.handshakeTokenExpiry = accessToken ? getAccessTokenExpiry() : null` runs synchronously right after `accessToken` is resolved, with no `await` in between, so the store-read reflects the token actually presented.
- **No stranded queue:** the early return leaves the queue intact for the still-armed keepalive to re-flush (consistent with the reconnecting-path contract in `frontend-sockets.md`), and in-flight requests were already settled at the top of `handleClose`. The ping interval is never torn down on this path, so nothing is stranded permanently — a token that eventually nears expiry re-enters the normal path.

### Test coverage

Solid and to the point. `handleClose token freshness` covers all four branches: fresh token skips refresh/budget/reconnect and preserves stored tokens; the keepalive still reconnects 10s later without spending budget (`socketAuthRetries` stays 0); a near-expiry token (10s) still refreshes and reconnects; an undecodable token still refreshes. The `makeAccessToken` helper mirrors `auth.test.ts` and reuses the existing mock/`internals` infra rather than starting new machinery.

### What's good

- The 120s margin is a defensible choice (past the pre-emptive-refresh leeway and any plausible clock skew), and the reasoning is captured in both the constant's doc comment and `frontend-sockets.md`.
- Scope discipline: the related `ensureValidAccessToken` retryable-fallback defect the issue mentions is correctly left out as a separate concern.
- The `docs/frontend-sockets.md` addition accurately documents the new gate without over-explaining.

---

ginderjeremiah:
CI has since gone fully green (`all-checks-passed` ✓ — frontend verify/test and all three e2e browsers passed; backend jobs correctly skipped for this frontend-only change). Nothing has changed since my approving review and there are no outstanding nits or should-fixes, so merging.

---

PR #2385, opened 2026-07-24
Guard character-select's selectPlayer against ApiRequest's pre-emptive refresh

### LINKED ISSUE #2370: Character-select page reads the refresh token before ApiRequest's pre-emptive refresh can rotate it — missing the guard CharacterSwitcher has
## Problem

`UI/src/routes/select/+page.svelte:35-41` reads `getRefreshToken()` into the `SelectPlayer` request body *before* `ApiRequest.post` runs — but `execute()` calls `ensureValidAccessToken()` first, which rotates the single-use refresh token whenever the access token is inside its 30s expiry leeway. `SelectPlayer` is then sent the already-consumed token in its body and the backend rejects it.

The sibling `switchPlayer` in `UI/src/routes/game/CharacterSwitcher.svelte:68-73` guards the identical body-token pattern with an explicit `await ensureValidAccessToken();` before reading the token, with a comment documenting exactly this race. The select-screen path is missing the same guard.

## Failure scenario

A player with an active session elsewhere gets the takeover confirm modal and hesitates on it while their access token crosses into the 30s pre-expiry window; on confirm, the pre-emptive refresh rotates the pair after the body token was read → "Could not enter the game." (A second click succeeds, since storage now holds the rotated pair.)

## Fix

Mirror CharacterSwitcher: `await ensureValidAccessToken()` before `getRefreshToken()` in `selectPlayer` (and check `createPlayer`/any other body-token caller on that page for the same pattern). Consider a shared helper so the third occurrence can't be written unguarded.

ginderjeremiah:
## Summary

- `UI/src/routes/select/+page.svelte`'s `selectPlayer` read `getRefreshToken()` into the `SelectPlayer` request body *before* `ApiRequest.post`'s own `ensureValidAccessToken()` could rotate it. If the access token crossed into its 30s pre-emptive-refresh leeway while a player hesitated on the session-takeover confirm modal, the refresh rotated the token pair *after* the (now stale, single-use) token had already been read — the backend then rejected `SelectPlayer` with "Could not enter the game." (a second click would succeed, since storage held the rotated pair by then).
- `UI/src/routes/game/CharacterSwitcher.svelte`'s `switchPlayer` already guarded the identical body-token pattern with an explicit `await ensureValidAccessToken()` before reading the token (added for the same race in #1767) — the select-screen path was just missing it.
- Rather than duplicating the two-step guard a third time, extracted it into a shared `getRotatedRefreshToken()` helper in `UI/src/lib/api/auth.ts` (settle any due refresh, then read the token) and switched both `selectPlayer` and `switchPlayer` to call it — per the issue's own suggestion, so a future third body-token caller can't be written unguarded.
- Checked `createPlayer` on the select page for the same pattern — it doesn't send a refresh token in its body, so no change needed there.

## Follow-up filed

While auditing body-token callers I found `UI/src/lib/api/logout.ts` has the same shape (reads `getRefreshToken()` before `ApiRequest.post`), but it doesn't break the user-facing flow the way this bug did — `clearTokens()` runs unconditionally in a `finally`, so logout always succeeds client-side; the only effect is that a rotated-but-unrevoked refresh token might not get explicitly revoked server-side. Filed as a separate, lower-severity follow-up rather than expanding this PR's scope: #2386.

Implements #2370's suggested fix exactly: mirrors `CharacterSwitcher`'s guard on the select page, via a shared helper so the pattern isn't duplicated a third time.

## Test plan

- [x] `npm run test:unit:ci` — 311 test files, 3968 tests passed, no coverage threshold regressions.
- [x] Added `getRotatedRefreshToken` unit tests in `auth.test.ts` (valid-token no-op, settles-due-refresh-returns-rotated-token, returns-null-on-definitive-rejection).
- [x] Added a regression test on the select page pinning the exact race (`ensureValidAccessToken` resolves after rotating storage → `SelectPlayer` is sent the rotated token, not the stale one).
- [x] Updated `CharacterSwitcher`'s equivalent test to assert against the shared helper's mocked return value directly, since the ordering guarantee itself is now pinned once in `auth.test.ts` instead of duplicated per consumer.

Closes #2370

---
_Generated by [Claude Code](https://claude.ai/code/session_01AppWpGy9Jg21GRrDniiS9f)_

### DIFF
--- a/UI/src/lib/api/auth.ts
+++ b/UI/src/lib/api/auth.ts
@@ -157,6 +157,17 @@ export const ensureValidAccessToken = async (): Promise<AccessTokenResult> => {
 };
 
 /**
+ * Reads the stored refresh token for a caller that sends it in a request body rather than relying on
+ * `ApiRequest`'s own pre-emptive refresh (e.g. `SelectPlayer`/`SwitchPlayer`). Settles any due refresh
+ * first, so the token read here is the one that will actually be sent rather than one `execute`'s own
+ * `ensureValidAccessToken` is about to rotate out from under it (#1767, #2370).
+ */
+export const getRotatedRefreshToken = async (): Promise<string | null> => {
+	await ensureValidAccessToken();
+	return getRefreshToken();
+};
+
+/**
  * Handles an unrecoverable auth failure (refresh definitively rejected): clears the stored tokens and
  * returns the user to the login screen. A full-page navigation tears down all in-memory game state,
  * mirroring the logout flow. The redirect is skipped when already on the login page to avoid a reload loop.
--- a/UI/src/routes/game/CharacterSwitcher.svelte
+++ b/UI/src/routes/game/CharacterSwitcher.svelte
@@ -40,7 +40,7 @@
 </Popover>
 
 <script lang="ts">
-import { ApiRequest, apiSocket, ensureValidAccessToken, getRefreshToken, setTokens } from '$lib/api';
+import { ApiRequest, apiSocket, getRotatedRefreshToken, setTokens } from '$lib/api';
 import { playerManager, stopEngines } from '$lib/engine';
 import Popover from '$components/Popover.svelte';
 import { confirmSessionTakeover } from '../login/session-takeover';
@@ -66,11 +66,7 @@ let view = $state<PlayerSelectView | null>(null);
 // credit + rebind through SwitchPlayer. Every terminal outcome ends in a full-page navigation, so the
 // brief window with a torn-down game is never interactive.
 const switchPlayer: PlayerSelectDeps['selectPlayer'] = async (playerId) => {
-	// SwitchPlayer is authenticated, so posting it would otherwise let ApiRequest.execute's own
-	// pre-emptive ensureValidAccessToken() rotate the refresh token *after* it had already been read
-	// below — settling that refresh here first means the token we read is the one that's actually sent.
-	await ensureValidAccessToken();
-	const refreshToken = getRefreshToken();
+	const refreshToken = await getRotatedRefreshToken();
 	if (!refreshToken) {
 		reloadToBoot();
 		return { ok: false, error: 'Your session is no longer valid. Please log in again.' };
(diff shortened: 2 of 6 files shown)


ginderjeremiah (commented):
## Review: guard select-page `selectPlayer` against ApiRequest's pre-emptive refresh (#2370)

**Verdict:** Approve _(posted as a comment since I'm the PR author; GitHub won't accept a gating event on one's own PR)_ — the fix is correct, minimal, and matches the issue's own suggested approach.

This resolves #2370 exactly. `select/+page.svelte`'s `selectPlayer` read `getRefreshToken()` into the `SelectPlayer` body before `ApiRequest.post` could rotate the pair via its own `ensureValidAccessToken()`; the fix routes both `selectPlayer` and the already-guarded `CharacterSwitcher.switchPlayer` through a single `getRotatedRefreshToken()` helper (settle any due refresh, then read the token), so a future third body-token caller can't reintroduce the race.

### Correctness
- `getRotatedRefreshToken()` faithfully preserves CharacterSwitcher's original inline two-step semantics — including the null-on-rejected-refresh path (`ensureValidAccessToken` → `clearTokens()` → `getRefreshToken()` returns null → callers hit their existing guard), so no behavioral regression on either side.
- Completeness of the audit holds up: `createPlayer` sends only `{ name, classId }` (no body token, correctly left alone), and the one remaining occurrence (`logout.ts`) was split out as follow-up #2386 rather than scope-creeping this PR — a reasonable call given `logout` clears tokens unconditionally in `finally`, so it never surfaces a user-facing error.

### Test coverage
- `auth.test.ts` pins the ordering guarantee once at the helper level (valid-token no-op, settles-due-refresh-returns-rotated, null-on-definitive-rejection).
- The select page gains a regression test asserting the rotated token is what's sent (`refreshToken: 'rotated'`), pinning the exact race.
- Reworking `character-switcher.test.ts` to mock the shared helper directly is the right move — the ordering is now verified once in `auth.test.ts` instead of duplicated per consumer, and the switcher test still asserts the resolved token flows through to the `SwitchPlayer` body.

### What's good
DRY extraction driven by the issue's own suggestion, faithful preservation of existing behavior, and tests that pin the race rather than just the happy path. Clean work — no blocking items, should-fixes, or nits.

---

Issue #696, opened 2026-06-15
Concurrency/memory-model hardening in Game.Infrastructure (RedisMultiplexerFactory DCL + reuse, BackgroundWorker teardown flags)


ginderjeremiah:
## Context

A health review of `Game.Infrastructure` found a few memory-model / concurrency issues. They're the same class of fix, so grouping.

## 1. Broken double-checked locking on the Redis multiplexer singletons

`Game.Infrastructure/Redis/RedisMultiplexerFactory.cs:9-10,22,58`: `_cacheInstance` / `_pubsubInstance` are plain `static` fields read outside the lock (`if (_cacheInstance is null)`) and written inside it — classic broken DCL, which under the .NET memory model permits observing a partially-constructed `ConnectionMultiplexer`. The sibling `BackgroundWorker._isRunning` is already correctly `volatile`, so the pattern is known in this codebase.

**Direction:** mark both fields `volatile` (or read via `Volatile.Read`).

## 2. Multiplexer reuse is unreliable

`RedisMultiplexerFactory.cs:26-29,62-65`: the "reuse the other multiplexer if configs match" branch (a) races on first startup (both getters resolved concurrently can connect two multiplexers for the same config), and (b) compares against `ConnectionMultiplexer.Configuration`, which StackExchange.Redis normalizes/rewrites — so it can be unequal to the raw connection string even when semantically identical, silently defeating reuse.

**Direction:** collapse both overloads into one `GetOrConnect(connectionString)` keyed by the **original** string under a single lock.

## 3. BackgroundWorker teardown race on non-volatile flags

`Game.Infrastructure/BackgroundWorker.cs:20-21`: `_isRunning` is `volatile`, but `_disposed` / `_hasBeenKilled` (read in `Start`, written in `Kill`/`Dispose` from a different thread) are plain bools. A `Start()` racing a `Dispose()` can read a stale `_disposed == false` and then `Set()` a disposed `AutoResetEvent`. Narrow, but real.

**Direction:** make both fields `volatile` for consistency with `_isRunning`.

### Labels
bug, claude, scope: small

PR #2387, opened 2026-07-24
Connect Redis multiplexers outside the process-wide lock, add failure backoff

### LINKED ISSUE #2371: RedisMultiplexerFactory runs synchronous Connect under a process-wide lock with no failure caching — a Redis outage collapses into head-of-line blocking
## Problem

`Game.Infrastructure/Redis/RedisMultiplexerFactory.cs:98-110` (`GetOrAdd` holds the static `_lock` across `factory(key)`) + `:130-138` (synchronous `ConnectionMultiplexer.Connect`, default `abortConnect=true`, ~5s), reached per-resolution from the transient DI registrations (`ServiceCollectionExtensions.cs:45,61` via `CacheServiceFactory`).

A successful connect is cached for the process lifetime, but a **failed connect is never negatively cached**: with Redis unreachable (boot during a Redis blip, or the window after `RedisConnectionLifetime` clears the cache), every scope that resolves `ICacheService`/`IPubSubService` re-attempts a synchronous ~5s connect **serialized behind the process-wide lock**. Requests queue one blocking attempt at a time on request threads — a Redis outage becomes head-of-line collapse of the whole instance instead of fast per-request failures.

## Suggested fix

Connect outside the lock (double-checked publish so racing getters still collapse onto one instance), and/or cache a faulted attempt briefly with backoff so concurrent resolvers fail fast during an outage instead of queuing. Keep the #696 single-instance-per-connection-string semantics and the #954 dispose path intact; the existing `GetOrAdd`/dispose seams are already unit-testable without a live connection, so the backoff/negative-cache behavior can be pinned the same way.

ginderjeremiah:
## Summary

`RedisMultiplexerFactory.GetOrAdd` ran the synchronous `ConnectionMultiplexer.Connect` (~5s, `abortConnect=true`) for the *entire* invocation inside the static process-wide lock. A successful connect is cached for the process lifetime, but a **failed** connect was never negatively cached — during a Redis outage (or the window after `RedisConnectionLifetime` clears the cache on shutdown), every scope resolving `ICacheService`/`IPubSubService` would re-attempt a synchronous ~5s connect, serialized one at a time behind the lock. A Redis outage collapsed into head-of-line blocking of the whole instance instead of fast per-request failures.

- `GetOrConnect` (renamed from `GetOrAdd`) now only holds the lock for short dictionary operations: checking whether the key is already connected, checking whether a recent failure is still within a 5s backoff window (fail fast with the remembered exception), and publishing a successful connect.
- The connect attempt itself now runs **outside** the lock, via a double-checked publish: two callers can race a first-time connect for the same key since neither blocks the other, but only the first to publish wins — the loser's already-open connection is cleanly discarded (`DisposeAsync`, best-effort logged on failure) so every future caller still shares one instance (preserves #696's dedup guarantee for anything actually handed out, just not for the connect attempt itself).
- A failed connect is remembered for 5s so concurrent/subsequent resolvers fail fast with the same exception instead of each paying for another ~5s attempt. The failure cache is cleared on a subsequent successful connect and on the graceful-shutdown teardown (`RedisConnectionLifetime`/`ResetForTesting`), alongside the existing multiplexer cache clear (#954).
- Threaded an `ILogger` into `RedisMultiplexerFactory.GetMultiplexer` (from the existing `ILoggerFactory` already available in `CacheServiceFactory`/`PubSubServiceFactory`) so a failed connect and a failed loser-discard are both logged — without ever logging the raw connection string, since it may carry a password.

Implements the suggested fix in #2371: connect outside the lock with double-checked publish, plus a short negative cache with backoff so concurrent resolvers fail fast during an outage instead of queuing. Keeps the #696 single-instance-per-connection-string semantics and the #954 graceful-dispose path intact.

## Test plan

- [x] `dotnet test` (full solution) — 3,477 tests passed (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests).
- [x] Updated `RedisMultiplexerFactoryTests` to pin the new contract: same-key reuse, distinct-key/exact-string-keying (unchanged), concurrent first-time racers converge on one published winner with every loser discarded exactly once, a failed connect fails fast within the backoff window without retrying, a retry is allowed once the backoff elapses, a subsequent success clears the remembered failure, and a failed connect logs a warning.
- [ ] The connect step itself (`ConnectionMultiplexer.Connect` against a live/down Redis) is exercised by existing integration tests rather than new ones, per the existing test-layering convention documented on `RedisMultiplexerFactoryTests`.

Closes #2371

https://claude.ai/code/session_01T9T9U1rFKqFuZ1ndcsx3pH

---
_Generated by [Claude Code](https://claude.ai/code/session_01T9T9U1rFKqFuZ1ndcsx3pH)_

### DIFF
--- a/Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
+++ b/Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
@@ -232,6 +232,33 @@ namespace Game.Infrastructure.Tests
         }
 
         [Fact]
+        public async Task DiscardLoserAsync_DisposesTheDiscardedValue()
+        {
+            var loser = new FakeAsyncDisposable();
+
+            await RedisMultiplexerFactory.DiscardLoserAsync(loser, NullLogger.Instance);
+
+            Assert.Equal(1, loser.DisposeCount);
+        }
+
+        [Fact]
+        public async Task DiscardLoserAsync_DisposeFaults_LogsAWarning_AndDoesNotThrow()
+        {
+            var loser = new FakeAsyncDisposable(throwOnDispose: true);
+            var logger = new CapturingLogger();
+
+            // A discarded loser's connection failing to close cleanly must not propagate out of the caller (the
+            // publish path that decided this instance lost the race) — the same contract DisposeAllAsync holds
+            // for a faulting shutdown close.
+            await RedisMultiplexerFactory.DiscardLoserAsync(loser, logger);
+
+            Assert.Equal(1, loser.DisposeCount);
+            var entry = Assert.Single(logger.Entries);
+            Assert.Equal(LogLevel.Warning, entry.Level);
+            Assert.NotNull(entry.Exception);
+        }
+
+        [Fact]
         public async Task DisposeAllAsync_DisposesEveryEntry_AndClearsTheCache()
         {
             var cache = new Dictionary<string, FakeAsyncDisposable>();
--- a/Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
+++ b/Game.Infrastructure.Tests/RedisMultiplexerFactoryTests.cs
@@ -6,18 +6,21 @@ using Xunit;
 namespace Game.Infrastructure.Tests
 {
     /// <summary>
-    /// Tests the keyed get-or-add reuse semantics of <see cref="RedisMultiplexerFactory"/> (#696). The connect
-    /// step itself opens a real Redis connection, so it is covered by integration tests rather than here; this
-    /// exercises the cache/lock logic through the generic <c>GetOrAdd</c> seam with sentinel values and a stub
-    /// factory, which is exactly the branch that decides whether two requests share a multiplexer or each open
-    /// their own.
+    /// Tests the keyed get-or-connect reuse semantics of <see cref="RedisMultiplexerFactory"/> (#696, #2371). The
+    /// connect step itself opens a real Redis connection, so it is covered by integration tests rather than here;
+    /// this exercises the cache/lock/negative-cache logic through the generic <c>GetOrConnect</c> seam with
+    /// sentinel values and a stub factory, which is exactly the branch that decides whether resolvers share a
+    /// multiplexer, race a first connect, or fail fast during an outage.
     /// </summary>
     public class RedisMultiplexerFactoryTests
     {
+        private static readonly Action<object> NoOpDiscard = _ => { };
+
         [Fact]
-        public void GetOrAdd_SameKey_ReturnsSameValueAndCreatesOnce()
+        public void GetOrConnect_SameKey_ReturnsSameValueAndCreatesOnce()
         {
             var cache = new Dictionary<string, object>();
+            var failures = new Dictionary<string, RedisMultiplexerFactory.ConnectFailure>();
             var syncRoot = new object();
             var created = 0;
             object Factory(string _)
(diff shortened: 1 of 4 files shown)


ginderjeremiah (commented):
## Review: connect Redis multiplexers outside the process-wide lock + fail-fast negative cache (#2371)

**Verdict:** Approve — faithfully implements the issue's suggested fix (double-checked connect outside the lock plus a short negative cache with backoff), preserves the #696 single-instance and #954 dispose semantics, and pins the new behavior through the existing unit-testable seam. Posting as `COMMENT` because I'm the PR author, so GitHub won't accept a gating approval from me — treat this as an approve.

### Correctness

Traced the double-checked publish and negative-cache paths and they hold up:
- **Reuse:** an already-connected key returns under the lock without touching the factory (#696 preserved).
- **First-connect race:** the `factory` call runs outside the lock, so a Redis outage no longer serializes resolvers behind one ~5s connect; racers each connect, but only the first to re-acquire the lock publishes and every loser is discarded exactly once, so callers still converge on one shared instance.
- **Outage fail-fast:** the first burst records a failure; subsequent resolvers within `FailureBackoff` rethrow the remembered exception immediately instead of paying for another attempt. After the window elapses a retry is allowed, and a later success both publishes and clears the remembered failure (`recentFailures.Remove(key)`).
- **Shutdown:** `DisposeAllAsync` now also clears `_recentFailures` under the lock, so a post-restart resolve starts clean.

The lock only ever guards short dictionary ops, and `LogWarning` is correctly kept outside the lock. No connection string is logged (password-safe) — good.

### Test coverage

Strong and aligned with `docs/backend.md` (`Game.Infrastructure.Tests` stays in-process; the real `Connect` is left to integration tests, and disrupting-shared-infra branches are measure-only). New cases pin fail-fast-within-backoff, retry-after-backoff, failure-cleared-on-success, warning logging, and the winner/loser discard invariant. The concurrent test correctly asserts against the set of *actually created* values rather than the thread count, so it stays robust when a late thread short-circuits on the cache check.

### Notes (non-blocking)
- `DiscardLoser` is fire-and-forget, which is the right call here: it's best-effort cleanup of a redundant connection with nothing durable to lose, and exceptions are caught and logged — consistent with the write-behind wake-signal precedent in `backend-persistence.md` (as opposed to the awaited reference-cache broadcast).
- The injected `Func<DateTime> utcNow` / `Action<T> discardLoser` seams keep the backoff timing and discard behavior deterministically testable without a live connection — nice.

Clean change. Merging once the remaining CI jobs (`test-backend`, `test-e2e (webkit)`) finish green.

---

ginderjeremiah (commented):
## Review update: CI is red — this supersedes my approve above

CI finished after my previous comment, and `test-backend` **failed**. It's not a test failure (all 3,477 tests pass) — it's the **backend coverage gate**:

```
Game.Infrastructure (Redis)  93.9% line  92.5% branch  L95/B91  FAIL
Backend coverage gate FAILED:
  - Game.Infrastructure (Redis) line 93.9% < floor 95%
```

**Verdict: Request changes** (posting as `COMMENT` since I'm the PR author — treat it as request-changes). One blocking item, which must be fixed in this PR rather than deferred:

### Blocking

- **New `DiscardLoser` / `DiscardLoserAsync` production path is uncovered, breaking the coverage floor** — `Game.Infrastructure/Redis/RedisMultiplexerFactory.cs`. The unit tests exercise `GetOrConnect<T>` with `NoOpDiscard`, so the real discard path — `DiscardLoserAsync`'s `await multiplexer.DisposeAsync()` and its `catch → LogWarning` branch — is never executed, dropping `Game.Infrastructure (Redis)` line coverage to 93.9% (floor 95%). This is an in-process branch that a fake `IConnectionMultiplexer` can drive directly (success path + a `DisposeAsync` that throws, asserting a warning is logged), exactly mirroring the existing faulting-close test for `DisposeAllAsync<T>`. It is **not** a "disrupting shared infrastructure = measure-only" case under `docs/backend.md`, so the gate is correct to hold it. Suggest making `DiscardLoserAsync` (or a thin seam around it) `internal` and adding the two-case test to `RedisMultiplexerFactoryTests`.

Everything in my earlier review still stands — the core logic is correct and well-tested, the negative-cache/backoff/double-checked-publish behavior is sound, and the fire-and-forget discard is the right design. The only gap is that the discard path's own dispose/log branch needs unit coverage to clear the floor.

I won't merge while CI is red. Once the discard-path test lands and `test-backend` goes green, this is an approve.

---

ginderjeremiah:
Fixed in 9745373 — made `DiscardLoserAsync` generic over `IAsyncDisposable` (mirroring `DisposeAllAsync<T>`) so a fake disposable can drive both the success and faulting-dispose paths directly, and added the two-case test. Reran the coverage gate locally: `Game.Infrastructure (Redis)` is now 96.0% line / 92.5% branch (floor 95%/91%), all 3,479 backend tests pass. Waiting on CI to confirm green.

---

PR #2350, opened 2026-07-23
Workbench: blocking warnings for acquisition-flag & weapon-signature hard-rejects

ginderjeremiah:
## Summary

The #2217/#2222 convention: any condition the backend is *known* to hard-reject on save must surface as a `blocking` warning that disables Save. The "keep a stale reference visible" picker convention (`reference.svelte.ts`'s `keep`s, chips catalogues keeping flag-lost skills as removable chips) left four such hard-rejected states reachable with no blocking warning, since none of `AdminSkills.SaveSkills`, the class/enemy/proficiency/item child savers have a cross-reference guard preventing an admin from freely removing a skill's acquisition flag (or retuning its damage portions) after it's already referenced elsewhere.

## How this addresses the issue

Mirrors the existing zone/challenge "no longer flagged as…" blocking-warn pattern onto the four gaps identified in #2333:

- **Class starter skills** (`entities/class.ts`): the `starterSkills` chips section now blocks Save when an assigned skill has lost its `Player` acquisition flag — mirrors `AdminClasses.FindStarterSkillFlagViolation`.
- **Enemy skill pool** (`entities/enemy.ts`): the `skills` chips section now blocks Save when a pool skill has lost its `Enemy` acquisition flag — mirrors `AdminEnemies.FindEnemySkillFlagViolation`.
- **Proficiency milestone rewards** (`progression/progression-helpers.ts`): new `rewardSkillFlagWarnings`, wired into both `proficiencyWarnings` (advisory list) and `proficiencyBlockingWarnings` (gates Save), flags a reward skill that lost its `Player` acquisition flag — mirrors `AdminProficiencies.SetRewards`/`CheckPlayerSkill`. Also corrects the stale doc comment on `levelRangeWarnings` that claimed level-range was "the one" proficiency condition the backend hard-rejects (it's no longer the only one) and wires the new check into `TierDetail.svelte`'s Milestones tab warning indicator.
- **Items** (`entities/item.ts`): the identity section's `warn` now additionally flags (a) a granted skill (weapon or not) that lost its `Item` acquisition flag, and (b) a weapon whose granted skill's primary damage-type signature no longer matches the item's own `weaponType` — mirrors `AdminItems.FindGrantedSkillFlagViolation` / `FindWeaponInvariantViolation`. Reuses the existing frontend `primaryDamageType`/`isWeaponLeaf` resolvers (`$lib/battle/damage-types`) rather than re-deriving the signature type.

## Test plan

[...]

PR #2389, opened 2026-07-24
Workbench blocking warnings for five more backend hard-rejects (#2376)

### LINKED ISSUE #2376: Workbench blocking-warning coverage: five more backend hard-rejects with no client surface (incl. the Synthesis acquisition flag being unauthorable)
## Context

Follow-up to #2350/#2333/#2276 — the same class of gap (backend `AdminSaveResult.Failure` hard-rejects the Workbench neither warns about nor blocks Save for), found adjacent to those fixes. Each verified against the backend guard it trips.

## Gaps

1. **The skill editor cannot author the `Synthesis` acquisition flag at all.** `entities/skill.ts:119-128` offers only Player/Item/Enemy; `ESkillAcquisition.Synthesis` (= 8, `Game.Core/Enums.cs:1066`) is missing, so a newly-authored skill can never be flagged as a synthesis result and never appears in the recipe editor's result picker — new synthesis content is unauthorable end-to-end. This is the most severe of the five (a functional hole, not just a missing warn).
2. **Item proficiency-gate rejects aren't warned.** `entities/item.ts:75-106` warns on the weapon/granted-skill hard-rejects but not `AdminItems.FindProficiencyGateViolation` (AdminItems.cs:201-227): a retired `RequiredProficiencyId`, or a `RequiredProficiencyLevel` outside `[1, MaxLevel]` (`NumInput` permits 0 and above-cap).
3. **Class distributions accept non-core attributes the backend rejects.** The shared distribution table (`entities/attribute-table-sections.ts:50-58`, used by `class.ts:225-230`) offers the full `EAttribute` catalogue, but `AdminClasses.FindAttributeDistributionViolation` (AdminClasses.cs:187-191) hard-rejects any non-core attribute in a *class* distribution (enemies are unrestricted, which the shared section masks).
4. **Skill recipes: duplicate-result rule unenforced client-side, and `newItem` is born violating it.** `entities/skill-recipe.ts:41,66-69` — the result picker doesn't exclude results already claimed by a live recipe (`AdminSkillRecipes.FindDuplicateResultViolation`, AdminSkillRecipes.cs:220-258), and a new recipe defaults `resultSkillId` to the first Synthesis-flagged skill, which almost certainly already has a producing recipe. Also: the conditions warn (`:120-123`) blocks `minLevel < 1` but not `minLevel > proficiency.MaxLevel` (rejected at AdminSkillRecipes.cs:152, equally checkable from `staticData.proficiencies`).
5. **Duplicate lesson keys aren't warned.** `entities/lesson.ts:51-59` has no collision check for `key`, which `AdminLessons.FindKeyCollision` (AdminLessons.cs:125-155) hard-rejects — both records look valid, Save rejects the whole batch with a toast and no dot pointing at the offender.

## Fix shape

Same pattern as #2350: blocking `warn`s (or picker exclusions where cleaner, per the challenge-reward picker precedent) mirroring each backend guard, plus adding the missing Synthesis flag to the acquisition field. Each warn should be pinned by the entity-config tests the earlier fixes added.

ginderjeremiah:
## Summary

Follow-up to #2350/#2333/#2276 — the same class of gap (a backend `AdminSaveResult.Failure` hard-reject the Workbench neither warns about nor blocks Save for). Closes all five gaps listed in #2376:

1. **Skill editor can now author the `Synthesis` acquisition flag.** `entities/skill.ts`'s acquisition field previously offered only Player/Item/Enemy, so a newly-authored skill could never be flagged as a synthesis result — new synthesis content was unauthorable end-to-end. Added the missing flag option.
2. **Item proficiency-gate rejects are now warned.** `entities/item.ts`'s identity warn now covers `AdminItems.FindProficiencyGateViolation`: a retired `RequiredProficiencyId`, or a `RequiredProficiencyLevel` outside `[1, MaxLevel]`.
3. **Class distributions are restricted to core attributes.** `AdminClasses.FindAttributeDistributionViolation` hard-rejects any non-core attribute in a *class* distribution (enemies stay unrestricted). Added an optional `attributeFilter` to the shared `attributeDistributionSection` factory, applied only by `class.ts`. This needed the six-member core-attribute set, previously private to the Attributes screen (`attributes-view.svelte.ts`) — moved it to `$lib/common/attribute-display.ts` as `CORE_ATTRIBUTES` so the Workbench can reuse it without pulling game-engine imports into the admin bundle; the Attributes screen re-exports it for its own existing consumers.
4. **Skill recipes: duplicate-result rule now enforced client-side.** Added `reference.availableSynthesisResultSkillOptions`, which narrows the result picker to exclude a Synthesis skill already claimed by another *live* recipe (`AdminSkillRecipes.FindDuplicateResultViolation` — one producing recipe per skill), used by both the result field and `newItem`'s default. Also extended the conditions warn to cover `minLevel > proficiency.MaxLevel` (previously only `minLevel < 1` was checked).
5. **Duplicate lesson keys are now warned.** `entities/lesson.ts`'s identity warn now covers `AdminLessons.FindKeyCollision` — a key collision against another lesson, including a retired one (which still holds its key, matching the backend's own check).

All five follow the same pattern as #2350: blocking `warn`s (or a picker exclusion for the duplicate-result case, per the challenge-reward picker precedent) mirroring each backend guard.

## Test plan

- [x] `npm run test:unit:ci` — full frontend suite green (3987 tests).
- [x] New/extended unit tests for each of the five gaps in `entities/*.test.ts` and `attribute-table-sections.test.ts`, pinning both the warn/exclusion behavior and the non-restricted (enemy) case staying unaffected.
- [x] `dotnet build` — 0 warnings, 0 errors (backend untouched by this PR; confirms nothing else broke).

Closes #2376

https://claude.ai/code/session_01BA3yA8VwWX3wyMYXxXwkAH

---
_Generated by [Claude Code](https://claude.ai/code/session_01BA3yA8VwWX3wyMYXxXwkAH)_

### DIFF
--- a/UI/src/routes/admin/workbench/entities/attribute-table-sections.ts
+++ b/UI/src/routes/admin/workbench/entities/attribute-table-sections.ts
@@ -32,10 +32,11 @@ export const attributeDistributionSection = <T extends Identified>(config: {
 	 */
 	attributeFilter?: (attributeId: number) => boolean;
 }): TableSectionConfig<T> => {
+	const { attributeFilter } = config;
 	const rowsOf = (rec: T) => fieldsOf(rec)[config.itemsKey] as AttributeDistributionRow[];
 	const attributeOptions = () =>
-		config.attributeFilter
-			? reference.attributeOptions().filter((o) => config.attributeFilter!(o.value))
+		attributeFilter
+			? reference.attributeOptions().filter((o) => attributeFilter(o.value))
 			: reference.attributeOptions();
 	return {
 		key: config.key,
@@ -50,7 +51,7 @@ export const attributeDistributionSection = <T extends Identified>(config: {
 			}
 			// A row can violate the restriction even though the picker excludes it from new authoring —
 			// e.g. it was set before the restriction existed. Hard-rejected on save, so it blocks (#2376).
-			const violator = config.attributeFilter && rows.find((r) => !config.attributeFilter!(r.attributeId));
+			const violator = attributeFilter && rows.find((r) => !attributeFilter(r.attributeId));
 			if (violator) {
 				return {
 					message: `'${attributeName(violator.attributeId, staticData.attributes)}' is not a core attribute and cannot have a distribution here`,
--- a/UI/src/routes/admin/workbench/entities/attribute-table-sections.ts
+++ b/UI/src/routes/admin/workbench/entities/attribute-table-sections.ts
@@ -1,3 +1,5 @@
+import { attributeName } from '$lib/common';
+import { staticData } from '$stores';
 import { reference } from '../reference.svelte';
 import { firstFree } from './helpers';
 import { fieldsOf, type Identified, type TableSectionConfig } from './types';
@@ -23,15 +25,40 @@ export const attributeDistributionSection = <T extends Identified>(config: {
 	itemsKey: keyof T & string;
 	desc: string;
 	emptySub: string;
+	/**
+	 * Restricts which attributes are selectable/valid in this table — e.g. a class distribution is
+	 * core-attribute-only (`AdminClasses.FindAttributeDistributionViolation`), while an enemy's stays
+	 * unrestricted. Omit for no restriction.
+	 */
+	attributeFilter?: (attributeId: number) => boolean;
 }): TableSectionConfig<T> => {
 	const rowsOf = (rec: T) => fieldsOf(rec)[config.itemsKey] as AttributeDistributionRow[];
+	const attributeOptions = () =>
+		config.attributeFilter
+			? reference.attributeOptions().filter((o) => config.attributeFilter!(o.value))
+			: reference.attributeOptions();
 	return {
 		key: config.key,
 		label: 'Attributes',
 		glyph: 'bars',
 		desc: config.desc,
 		count: (rec) => rowsOf(rec).length,
-		warn: (rec) => (rowsOf(rec).length ? null : 'No attribute distribution'),
+		warn: (rec) => {
+			const rows = rowsOf(rec);
+			if (!rows.length) {
+				return 'No attribute distribution';
+			}
+			// A row can violate the restriction even though the picker excludes it from new authoring —
+			// e.g. it was set before the restriction existed. Hard-rejected on save, so it blocks (#2376).
+			const violator = config.attributeFilter && rows.find((r) => !config.attributeFilter!(r.attributeId));
+			if (violator) {
+				return {
+					message: `'${attributeName(violator.attributeId, staticData.attributes)}' is not a core attribute and cannot have a distribution here`,
+					blocking: true
+				};
+			}
+			return null;
+		},
 		kind: 'table',
 		itemsKey: config.itemsKey,
 		rowKey: 'attributeId',
@@ -42,7 +69,7 @@ export const attributeDistributionSection = <T extends Identified>(config: {
 		newRow: (rec) => ({
 			attributeId: firstFree(
 				rowsOf(rec).map((a) => a.attributeId),
-				reference.attributeOptions()
+				attributeOptions()
 			),
 			baseAmount: 0,
 			amountPerLevel: 0
@@ -52,7 +79,7 @@ export const attributeDistributionSection = <T extends Identified>(config: {
 				key: 'attributeId',
 				label: 'Attribute',
 				type: 'attribute',
-				options: reference.attributeOptions,
+				options: attributeOptions,
 				min: 190,
 				unique: true
 			},
--- a/UI/src/lib/common/attribute-display.ts
+++ b/UI/src/lib/common/attribute-display.ts
@@ -11,6 +11,19 @@ import { formatNum, normalizeText } from './functions';
  * (e.g. `challengeTypeName`).
  */
 
+/** The six core attributes a player directly invests stat points into — the frontend mirror of the
+ *  backend `Attribute.CoreAttributes` invariant (`Game.Core/Attributes/Attribute.cs`). Every other
+ *  attribute is "derived" (computed from these), so this is the meaningful measure of raw attribute
+ *  investment, deliberately kept distinct from the `attributeType` display taxonomy (spike #528). */
+export const CORE_ATTRIBUTES: EAttribute[] = [
+	EAttribute.Strength,
+	EAttribute.Endurance,
+	EAttribute.Intellect,
+	EAttribute.Agility,
+	EAttribute.Dexterity,
+	EAttribute.Luck
+];
+
 /** Suffix matching the `--attr-*` custom properties (e.g. `strength`). */
 const ATTRIBUTE_KEY: Partial<Record<EAttribute, string>> = {
 	[EAttribute.Strength]: 'strength',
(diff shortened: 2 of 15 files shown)


UI/src/routes/admin/workbench/entities/attribute-table-sections.ts
```diff
@@ -23,15 +25,40 @@ export const attributeDistributionSection = <T extends Identified>(config: {
...
+	 * Restricts which attributes are selectable/valid in this table — e.g. a class distribution is
+	 * core-attribute-only (`AdminClasses.FindAttributeDistributionViolation`), while an enemy's stays
+	 * unrestricted. Omit for no restriction.
+	 */
+	attributeFilter?: (attributeId: number) => boolean;
 }): TableSectionConfig<T> => {
 	const rowsOf = (rec: T) => fieldsOf(rec)[config.itemsKey] as AttributeDistributionRow[];
+	const attributeOptions = () =>
+		config.attributeFilter
+			? reference.attributeOptions().filter((o) => config.attributeFilter!(o.value))
```
ginderjeremiah: Nit: the null-forgiving `!` here (and again at line 53, `!config.attributeFilter!(r.attributeId)`) contradicts CLAUDE.md's "AVOID using the null-forgiving operator (!) as much as possible" — and it's avoidable. Capturing the field into a `const` lets TS's control-flow narrowing carry into the closures, so no `!` is needed:

```ts
const attributeFilter = config.attributeFilter;
const attributeOptions = () =>
    attributeFilter
        ? reference.attributeOptions().filter((o) => attributeFilter(o.value))
        : reference.attributeOptions();
```

and in `warn`, guard once so the narrowed `const` reaches `.find`:

```ts
if (attributeFilter) {
    const violator = rows.find((r) => !attributeFilter(r.attributeId));
    if (violator) { /* ... */ }
}
```

Not blocking — behavior is correct either way.

---

ginderjeremiah (commented):
## Review: Workbench blocking warnings for five more backend hard-rejects (#2376)

**Verdict:** Approve *(posted as a comment — I'm the PR author, so GitHub won't take a gating review from me)* — all five gaps are closed correctly, follow the established `#2350` warn/picker-exclusion pattern, and are well tested. One avoidable-`!` nit, no blockers.

I verified each of the five against the issue and the backend guards it names:

1. **Synthesis acquisition flag** — added to `skill.ts`'s flag list; the functional hole (synthesis content unauthorable end-to-end) is closed.
2. **Item proficiency-gate** — `item.ts` now warns on a retired `requiredProficiencyId` and an out-of-range level, guarded behind the `!= null && !== -1` sentinel check and using the correct index lookup (`staticData.proficiencies?.[id]`).
3. **Class core-attribute restriction** — the `attributeFilter` hook on the shared `attributeDistributionSection` is a clean, opt-in extension; enemies correctly stay unrestricted (no filter passed). The picker excludes non-core, and the `warn` catches a pre-existing violating row the picker can't (good — the retired-value-stays-visible parallel).
4. **Duplicate-result + condition cap** — `availableSynthesisResultSkillOptions` correctly keeps the recipe's *own* current claim selectable (`resultSkillId !== current`) while excluding others' live claims, and `newItem`'s default now uses it; the conditions warn extends cleanly to `minLevel > proficiency.MaxLevel`.
5. **Lesson key collision** — `lesson.ts` warns on a key held by any other lesson including a retired one, with the empty-key and self-edit cases both handled.

The `CORE_ATTRIBUTES` move to `$lib/common/attribute-display.ts` (with the Attributes screen re-exporting for its own consumers) is the right call for DRY — the admin bundle gets the constant without pulling game-engine imports, and the "allocation domain" documentation stays where it's meaningful.

### Nits

- `attribute-table-sections.ts:38` / `:53` — null-forgiving `!` on `config.attributeFilter`, avoidable by capturing it into a `const` so TS narrows it into the closures (CLAUDE.md discourages `!`). Inline comment has the suggested form. Not blocking.

### Test coverage

Strong. Each of the five gaps is pinned by entity-config unit tests, including the boundary/edge cases that matter: retired-vs-live proficiency, `[1, MaxLevel]` boundaries (0, cap, cap+1), self-edit not colliding with its own key/claim, retired lesson still holding its key, and the enemy (unrestricted) path staying unaffected. The `attributeFilter` plumbing is exercised directly with a stand-in even/odd filter so it's decoupled from the real `CORE_ATTRIBUTES` membership. Consistent with the project's testing rules; no backend change, so no parity/integration surface to mirror.

### What's good

- Careful reuse of the "picker excludes for new authoring, `warn` blocks the pre-existing/retired-after-the-fact case" split throughout — matches the documented retired-record picker convention exactly.
- Comments cite the specific backend guard each warn mirrors, which makes the "why is this `blocking`" call auditable per the Workbench's own rule that a warn is only marked blocking once confirmed against the backend rejection.

Not merging: CI is still running and there's an open nit — per this workflow I only auto-merge a clean, green approval. Once frontend/e2e go green, the nit is optional to address before merge.

---

ginderjeremiah:
## Review: Workbench blocking warnings for five more backend hard-rejects (#2376)

**Verdict: Approve.** (Posted as a comment since I'm the PR author's account — GitHub won't accept a gating verdict from it.) This closes all five gaps in #2376 cleanly, mirrors the established #2350 pattern, and is well-tested. I verified the diff against the codebase rather than trusting the issue's framing:

### What I checked
- **Gap 1 (Synthesis flag):** `skill.ts` now offers `ESkillAcquisition.Synthesis` alongside Player/Item/Enemy — the functional hole (synthesis content unauthorable end-to-end) is closed. ✅
- **Gap 2 (item proficiency gate):** `item.ts` warns a retired `RequiredProficiencyId` and a level outside `[1, MaxLevel]`. Verified `IProficiency` really carries `name`/`maxLevel`/`retiredAt`, and the `staticData.proficiencies?.[id]` lookup matches the 0-based reference-cache convention. ✅
- **Gap 3 (class core-attribute restriction):** the shared `attributeDistributionSection` gained an optional `attributeFilter`, applied only by `class.ts` (enemies stay unrestricted). The warn correctly still blocks a pre-existing non-core row the picker would exclude from new authoring. ✅
- **Gap 4 (duplicate recipe result + condition cap):** `availableSynthesisResultSkillOptions` excludes results claimed by another *live* recipe while keeping the current recipe's own claim selectable (`r.resultSkillId !== current`), and `newItem` defaults through it. The conditions warn now also blocks `minLevel > proficiency.MaxLevel`. ✅
- **Gap 5 (lesson key collision):** `lesson.ts` warns on a key already held by another lesson, retired ones included (`other.id !== l.id && other.key === l.key`) — matches `FindKeyCollision`. ✅
- **CORE_ATTRIBUTES move:** relocating the constant to `$lib/common/attribute-display` (re-exported from `attributes-view` for its existing consumers) is the right call to keep game-engine imports out of the admin bundle.

### Tests
Comprehensive — each gap has positive, negative, and self-exclusion cases pinned in the entity-config tests, plus the filter plumbing tested independently of the real `EAttribute` membership. Matches the coverage convention the earlier fixes set.

### Nit — fixed in this PR
`attribute-table-sections.ts` used the null-forgiving operator (`config.attributeFilter!(…)`) in two nested closures, which CLAUDE.md asks to avoid where trivially avoidable. I pushed a one-line follow-up capturing `attributeFilter` into a narrowed local const, removing both. No behavior change.

### Merge
Holding only on CI — my nit-fix commit re-triggered the run. Will merge once it's green; no other blockers and no follow-up work owed.

---

PR #2390, opened 2026-07-24
Fix logout() reading a stale refresh token ahead of ApiRequest's rotation

### LINKED ISSUE #2386: logout() reads the refresh token before ApiRequest's pre-emptive refresh can rotate it
## Context

Found while fixing #2370 (character-select's `selectPlayer` had the identical race, already guarded in `CharacterSwitcher.switchPlayer`).

## Problem

`UI/src/lib/api/logout.ts:14` reads `getRefreshToken()` into the `Auth/Logout` request body *before* `ApiRequest.post` runs — but `execute()` calls `ensureValidAccessToken()` first, which rotates the single-use refresh token whenever the access token is inside its 30s pre-expiry leeway. If that race fires, `Auth/Logout` is sent the already-rotated-out (stale) token rather than the one now live in storage.

## Why this is lower severity than #2370

`logout` unconditionally clears local tokens and redirects in a `finally` block regardless of the `Auth/Logout` call's outcome, so the user-facing logout flow always succeeds — there's no visible error like #2370's "Could not enter the game." The only effect is that the *actual* live (rotated) refresh token may never get explicitly revoked server-side, since the request that reaches the backend presents the already-superseded one instead (which the backend likely already treats as invalid/consumed, given single-use rotation, so the revoke is effectively skipped rather than misapplied).

## Suggested fix

Use the `getRotatedRefreshToken()` helper added in #2370's fix (`UI/src/lib/api/auth.ts`) — settles any due pre-emptive refresh before reading the token, so the token sent to `Auth/Logout` is the one currently live in storage.

Tags: bug, claude, scope: small

ginderjeremiah:
## Summary

- `logout()` read `getRefreshToken()` directly into the `Auth/Logout` request body *before* `ApiRequest.post` ran. But `execute()` calls `ensureValidAccessToken()` first, which rotates the single-use refresh token whenever the access token is inside its 30s pre-expiry leeway — so if that race fired, `Auth/Logout` was sent the already-rotated-out (stale) token instead of the one now live in storage.
- Switches `logout.ts` to `getRotatedRefreshToken()` (added in #2370's fix), which settles any due pre-emptive refresh before reading the token — the same fix already applied to `CharacterSwitcher.switchPlayer` and `select/+page.svelte`'s `selectPlayer` for the identical race.

Implements #2386's suggested fix exactly: swap the direct `getRefreshToken()` read for `getRotatedRefreshToken()`.

Closes #2386

## Test plan

- [x] Added a unit test (`sends the token getRotatedRefreshToken resolves, settled before ApiRequest can rotate it (#2386)`) mirroring the existing #2370/#1767 regression tests in `character-switcher.test.ts`/`select-page.test.ts`.
- [x] `npm run test:unit:ci` (full suite) — 3,988 tests passed across 1,149 suites, coverage thresholds met.
- [x] Backend untouched; no `dotnet build`/`dotnet test` needed for this frontend-only change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nwn28R8uSxdkX5EeZLHsWP)_

### DIFF
--- a/UI/src/lib/api/logout.ts
+++ b/UI/src/lib/api/logout.ts
@@ -1,5 +1,6 @@
 import { ApiRequest } from './api-request';
-import { clearTokens, getRefreshToken } from './token-store';
+import { getRotatedRefreshToken } from './auth';
+import { clearTokens } from './token-store';
 
 /**
  * Ends the current session: revokes the refresh token on the server, clears the locally stored token
@@ -11,7 +12,9 @@ import { clearTokens, getRefreshToken } from './token-store';
  * no stored tokens and keeps the user on the login screen.
  */
 export const logout = async () => {
-	const refreshToken = getRefreshToken();
+	// Settled before reading, so a refresh due to fire inside ApiRequest's own pre-emptive check
+	// (execute -> ensureValidAccessToken) can't rotate the token out from under this read (#2386).
+	const refreshToken = await getRotatedRefreshToken();
 	try {
 		if (refreshToken) {
 			await new ApiRequest('Auth/Logout').post({ refreshToken });
--- a/UI/src/tests/lib/api/logout.test.ts
+++ b/UI/src/tests/lib/api/logout.test.ts
@@ -1,7 +1,10 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 
-const postMock = vi.fn();
-const constructorMock = vi.fn();
+const { postMock, constructorMock, getRotatedRefreshTokenMock } = vi.hoisted(() => ({
+	postMock: vi.fn(),
+	constructorMock: vi.fn(),
+	getRotatedRefreshTokenMock: vi.fn()
+}));
 
 vi.mock('$lib/api/api-request', () => ({
 	ApiRequest: class {
@@ -10,6 +13,10 @@ vi.mock('$lib/api/api-request', () => ({
 		}
 		post = postMock;
 	}
+}));
+
+vi.mock('$lib/api/auth', () => ({
+	getRotatedRefreshToken: getRotatedRefreshTokenMock
 }));
 
 import { logout } from '$lib/api/logout';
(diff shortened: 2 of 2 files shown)


ginderjeremiah (commented):
## Review: Fix `logout()` sending a stale refresh token ahead of `ApiRequest`'s rotation (#2386)

**Verdict:** Approve — correct, minimal fix that closes a real race by reusing the established `getRotatedRefreshToken()` helper; no blocking findings, no should-fixes, no nits. (Posting as a comment since I'm the PR author and GitHub won't accept a gating verdict on one's own PR.)

The problem is real: `logout()` read `getRefreshToken()` into the `Auth/Logout` body *before* `ApiRequest.post` → `execute` → `ensureValidAccessToken` could rotate the single-use refresh token out from under it, so the revoke could be sent an already-superseded token. Swapping to `await getRotatedRefreshToken()` settles any due pre-emptive refresh first, so the token read is the one live in storage — and because that call mints a fresh access token, `execute`'s own subsequent `ensureValidAccessToken` won't re-rotate. The reasoning holds.

This is the same guard already applied for the identical race in `CharacterSwitcher.switchPlayer` and `select/+page.svelte`'s `selectPlayer` (both confirmed calling the same helper), so the codebase stays consistent. The 2-line comment at `logout.ts:15` accurately explains the non-obvious ordering, which is worth keeping given how subtle the race is.

### Test coverage

- The new regression test (`logout.test.ts:88`) mirrors the #2370/#1767 tests in `character-switcher.test.ts`/`select-page.test.ts` and correctly asserts `logout` sends exactly what `getRotatedRefreshToken` resolves.
- The `beforeEach` mock now delegates to storage (`getTokens()?.refreshToken ?? null`), so the pre-existing "revokes the token" and "no refresh token" cases keep exercising real behavior through the mock. Coverage is appropriate for a frontend lib change of this size.

### What's good

- Tightly scoped to the actual defect, reuses the existing helper rather than duplicating rotation logic, and keeps the three sibling call sites uniform.


---

PR #2392, opened 2026-07-24
Log rate-limiter 429 rejections to the structured request log (#2375)

### LINKED ISSUE #2375: Rate-limiter 429s are invisible to the structured request log
## Problem

`Game.Api/Startup.cs:285-289`: `UseRateLimiter()` runs before `UseSessionLoader`/`UseRequestLogging`, so a 429 short-circuits before "Request Start"/"Request Ended" ever fire. During the exact incident the limiter exists for (credential stuffing against `/api/Auth`), the request log shows nothing about throttled traffic — an operator sees only the successful trickle.

## Suggested fix

Emit a log/counter from `OnRejected` in `Game.Api/RateLimiting/RateLimiterServiceCollectionExtensions.cs:63-77` (cheaper than moving the limiter inside the logging middleware, which would pay full structured-logging volume under attack). Include the partition key class (IP vs account) but not the raw credential.

ginderjeremiah:
## Summary

- `Game.Api/Startup.cs:285-289`: `UseRateLimiter()` runs before `UseSessionLoader`/`UseRequestLogging`, so a 429 short-circuits before "Request Start"/"Request Ended" ever fire. During the exact incident the limiter exists for (credential stuffing against `/api/Auth`), the request log showed nothing about throttled traffic — an operator would see only the successful trickle.
- Fix: `OnRejected` in `Game.Api/RateLimiting/RateLimiterServiceCollectionExtensions.cs` now resolves an `ILogger` from the request's `RequestServices` and emits a `Warning`-level structured entry (`Rate limit rejected {Path} PartitionKeyType={PartitionKeyType} PartitionKey={PartitionKey}`) before writing the 429 response.
- Per the issue's own cost tradeoff, this logs from `OnRejected` rather than moving the rate limiter inside `RequestLoggingMiddleware`, which would pay full structured-logging volume for every request under attack instead of just the rejected ones.
- Logs the partition-key **kind** (currently always `"IP"`, since the only policy today — `BuildAuthPartition` — partitions solely by client IP) alongside the resolved key value, never the request body (so no credential ever reaches this log line — this layer doesn't have access to it anyway).

Implements #2375's suggested fix directly: emits a log/counter from `OnRejected` rather than restructuring the middleware pipeline, and includes the partition key's class without exposing anything credential-shaped.

Closes #2375

## Test plan

- [x] New integration test `AuthRateLimitLoggingTests.Rejection_LogsWarningWithPathAndIpPartitionKey` (`Game.Api.Tests/Integration/AuthRateLimitLoggingTests.cs`) — drives the auth endpoint past a small configured limit and asserts the `Warning`-level log entry's message, path, and `PartitionKeyType`/`PartitionKey` properties, using the existing `CapturingLoggerProvider` test helper (same pattern as `RequestLoggingTests`).
- [x] `dotnet test` (full solution) — 3,480 tests passed (Game.Core.Tests, Game.Infrastructure.Tests, Game.Application.Tests, Game.Api.Tests), including the new test and the existing `AuthRateLimitingTests` suite.
- [x] Backend-only change; no frontend touched, so no `npm run lint`/`test:unit:ci` needed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LS7B1XsjJbruGo9XCXmiqk)_

### DIFF
--- a/Game.Api/RateLimiting/RateLimiterServiceCollectionExtensions.cs
+++ b/Game.Api/RateLimiting/RateLimiterServiceCollectionExtensions.cs
@@ -64,6 +64,9 @@ namespace Game.Api.RateLimiting
             // which would pay full structured-logging volume under attack.
             var logger = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>()
                 .CreateLogger(typeof(RateLimiterServiceCollectionExtensions).FullName!);
+            // OnRejected is registered globally, so it fires for every policy's rejections. "IP" is correct
+            // today since the only policy (auth) partitions purely by client IP; revisit if an
+            // account-partitioned policy is ever added.
             logger.LogWarning(
                 "Rate limit rejected {Path} PartitionKeyType={PartitionKeyType} PartitionKey={PartitionKey}",
                 context.HttpContext.Request.Path.Value, "IP", ClientIp.Resolve(context.HttpContext));
--- a/Game.Api/RateLimiting/RateLimiterServiceCollectionExtensions.cs
+++ b/Game.Api/RateLimiting/RateLimiterServiceCollectionExtensions.cs
@@ -3,6 +3,7 @@ using Game.Api.Models.Common;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.RateLimiting;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using System.Globalization;
 using System.Threading.RateLimiting;
@@ -57,6 +58,15 @@ namespace Game.Api.RateLimiting
         private static async ValueTask OnRejected(OnRejectedContext context, CancellationToken token)
         {
             context.HttpContext.Response.StatusCode = StatusCodes.Status429TooManyRequests;
+
+            // UseRateLimiter runs before UseRequestLogging, so a 429 never reaches the structured request
+            // log otherwise (#2375) — log it here instead of moving the limiter into that middleware,
+            // which would pay full structured-logging volume under attack.
+            var logger = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>()
+                .CreateLogger(typeof(RateLimiterServiceCollectionExtensions).FullName!);
+            logger.LogWarning(
+                "Rate limit rejected {Path} PartitionKeyType={PartitionKeyType} PartitionKey={PartitionKey}",
+                context.HttpContext.Request.Path.Value, "IP", ClientIp.Resolve(context.HttpContext));
 
             if (context.Lease.TryGetMetadata(MetadataName.RetryAfter, out var retryAfter))
             {
--- a/Game.Api.Tests/Integration/AuthRateLimitLoggingTests.cs
+++ b/Game.Api.Tests/Integration/AuthRateLimitLoggingTests.cs
@@ -1,0 +1,82 @@
+using Game.Api.RateLimiting;
+using Game.TestInfrastructure.Base;
+using Game.TestInfrastructure.Fixtures;
+using Game.TestInfrastructure.Helpers;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using System.Net.Http.Json;
+using Xunit;
+
+namespace Game.Api.Tests.Integration
+{
+    /// <summary>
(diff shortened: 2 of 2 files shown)


Game.Api/RateLimiting/RateLimiterServiceCollectionExtensions.cs
```diff
@@ -58,6 +59,15 @@ private static async ValueTask OnRejected(OnRejectedContext context, Cancellatio
...
             context.HttpContext.Response.StatusCode = StatusCodes.Status429TooManyRequests;
 
+            // UseRateLimiter runs before UseRequestLogging, so a 429 never reaches the structured request
+            // log otherwise (#2375) — log it here instead of moving the limiter into that middleware,
+            // which would pay full structured-logging volume under attack.
+            var logger = context.HttpContext.RequestServices.GetRequiredService<ILoggerFactory>()
+                .CreateLogger(typeof(RateLimiterServiceCollectionExtensions).FullName!);
+            logger.LogWarning(
+                "Rate limit rejected {Path} PartitionKeyType={PartitionKeyType} PartitionKey={PartitionKey}",
+                context.HttpContext.Request.Path.Value, "IP", ClientIp.Resolve(context.HttpContext));
```
ginderjeremiah, line 72: Nit (non-blocking): `PartitionKeyType` is a hardcoded `"IP"` literal here, but `OnRejected` is registered globally on `options.OnRejected`, so it fires for *every* policy's rejections, not just `BuildAuthPartition`. It's correct today since the auth policy is the only one and it partitions purely by IP — but if an account-partitioned policy is ever added (exactly the "IP vs account" distinction #2375 anticipated), this handler would silently mislabel those rejections as `IP` and log the client IP instead of the account key. The value being computed once at the source (e.g. the partition factory attaching its key-kind to the lease/context) rather than assumed here would remove that footgun. Since that's speculative future work, a one-line code comment noting "the only policy today partitions by IP; revisit if an account-partitioned policy is added" would be enough to carry the PR-description context into the code for the next reader.

---

ginderjeremiah (commented):
## Review: Log rate-limiter 429 rejections to the structured request log (#2375)

**Verdict:** Approve (posting as **Comment** since I'm the PR author, so GitHub won't take a gating approval on my own PR) — the change faithfully implements the issue's suggested fix, is well-scoped, and is properly tested. One non-blocking nit below.

### Blocking

- None.

### Should-fix

- None.

### Nits

- `RateLimiterServiceCollectionExtensions.cs:69` — `PartitionKeyType` is a hardcoded `"IP"` literal in a globally-registered `OnRejected` handler. Correct today (the auth policy is the only one and partitions by IP), but a latent footgun if an account-partitioned policy is added later, since it would mislabel those rejections. A one-line code comment carrying the PR-description context into the code would suffice. Details inline.
- Minor: `ClientIp.Resolve` is now called twice per rejected request (partition build + log line). The cost is negligible (a `RemoteIpAddress?.ToString()`), so not worth changing — noting only for completeness.

### Test coverage

- Solid. `AuthRateLimitLoggingTests.Rejection_LogsWarningWithPathAndIpPartitionKey` drives the auth endpoint past a locally-lowered `PermitLimit` and asserts the `Warning` entry's level, message, path, and the `PartitionKeyType`/`PartitionKey` structured properties, reusing the existing `CapturingLoggerProvider` helper — same pattern as `RequestLoggingTests`. The comment honestly documents why the partition key can only be asserted non-null under the in-memory `TestServer` (no socket peer). Backend-only change; no frontend/battle-logic surface, so no mirrored tests needed.

### What's good

- Chose the cheaper `OnRejected` path over relocating the limiter into `RequestLoggingMiddleware`, exactly as the issue's cost tradeoff called for — only rejected requests pay the logging cost, not all traffic under attack.
- Logs the partition-key *kind* and resolved value but never the request body, so no credential can reach this log line.
- The inline comment explaining the middleware-ordering rationale is genuinely useful context for the next reader.

---

ginderjeremiah:
## Re-review — nit addressed, merging

Re-reviewed the follow-up commit `f8a49a3` (the only change since my prior review). It adds a code comment above the `LogWarning` in `OnRejected` noting that the handler is registered globally and that `"IP"` is correct only because the auth policy is currently the sole IP-partitioned policy — revisit if an account-partitioned policy is ever added. That's exactly the context my nit asked to carry into the code, and it resolves the only outstanding item.

The secondary observation (double `ClientIp.Resolve` per rejected request) was explicitly a non-issue — negligible cost, not worth changing.

Verdict: **approve**, no blocking / should-fix / nits remaining. All checks green (`all-checks-passed` ✅). Squash-merging.

---

PR #2393, opened 2026-07-24
Write-behind read-path efficiency: single envelope parse + batched progress miss-reload (#2378)

### LINKED ISSUE #2378: Write-behind read-path efficiency: each drained envelope is JSON-parsed 4× (2 needed), and the progress miss-reload runs 3 sequential queries
## Two small, adjacent efficiency wins in the persistence tier

1. **Drain loop double-parses every envelope.** `DataProviderSynchronizer.cs:427` calls `PlayerUpdateEnvelopeReader.TryReadPlayerId` (full envelope `Deserialize` + `JsonDocument.Parse` of the payload — `PlayerUpdates/PlayerUpdateEnvelopeReader.cs:19-56`) for lane routing, then `ProcessMessage` (`:639`) re-deserializes the same envelope and the handler the payload again — four parses per event where two would do. Deserialize the envelope once, sniff `playerId` from the parsed envelope's payload, and hand the parsed envelope through to `ProcessMessage` (keeping the raw string for acknowledge/dead-letter, which correctly key off it). This is the hot fleet-wide convergence path; the refactor is mechanical and the dead-letter classifier keeps its own best-effort peek.
2. **Progress cache-miss reload is 3 sequential round-trips.** `PlayerProgressRepository.LoadFromDb` (`:227-264`) awaits statistics, challenges, and proficiencies queries one after another on one context. Cold path only (first read after TTL lapse/eviction), so impact is ~2 extra sequential latencies per rehydration; batchable into a single round-trip if connect latency ever matters.

Neither changes behavior; existing integration tests should pass unchanged, with the envelope-parse refactor pinned by the existing lane-ordering tests.

ginderjeremiah:
## Summary

Two small, adjacent efficiency wins in the persistence tier, as filed in #2378:

- **Drain loop double-parses every envelope.** `DataProviderSynchronizer.DrainQueueAsync` used to call `PlayerUpdateEnvelopeReader.TryReadPlayerId(next)` (a full envelope deserialize + `JsonDocument.Parse` of the payload) for lane routing, then `ProcessMessage` re-deserialized the same raw string into a `DomainEventEnvelope` again. The envelope is now parsed once in `DrainQueueAsync` (`PlayerUpdateEnvelopeReader.TryParseEnvelope`) and threaded through `ProcessReservedItemAsync`/`ProcessMessage`, which reuses it instead of re-parsing — the raw string is kept only for acknowledge/dead-letter, which correctly key off it. The malformed-vs-null-envelope dead-letter distinction (and its log messages/exception) is preserved exactly, just computed from the upfront parse result instead of a second deserialize.
- **Progress cache-miss reload was 3 sequential round-trips.** `PlayerProgressRepository.LoadFromDb` awaited the statistics, challenges, and proficiencies queries one after another on one `GameContext`. Since EF Core can't run independent queries concurrently on a single context, the three queries are now projected onto one shared anonymous shape (discriminated by a `Kind` field, with unused columns defaulted rather than left null) and `Concat`-ed together, so EF Core translates the whole read into a single `UNION ALL` round trip. This only hits on the cold path (first read after a TTL lapse/eviction, or a brand-new player).

Neither change alters behavior — same dead-letter/log semantics, same `CachedPlayerProgress` shape returned from `LoadFromDb`.

Implements both suggested fixes from #2378 as filed.

## Test plan

- [x] `dotnet test` (full solution, real Postgres/Redis containers) — 3,481 tests passed across `Game.Core.Tests`, `Game.Infrastructure.Tests`, `Game.Application.Tests`, `Game.Api.Tests`.
- [x] Targeted runs before the full suite: `DataProviderSynchronizerTests` (73 tests, including the lane-ordering and malformed-envelope dead-letter tests) and `PlayerProgressRepositoryIntegrationTests`/`PlayerUpdateDeadLettersIntegrationTests` all green.
- [x] The `Concat`-based UNION projection was validated against real Postgres — an initial attempt using a `record` projection failed EF Core's SQL translation (`Unable to translate set operation after client projection has been applied`); switching to an anonymous-type projection (the shape EF Core reliably translates for set operations) resolved it, confirmed by the integration tests above.

Closes #2378

---
_Generated by [Claude Code](https://claude.ai/code/session_01PxPkpXruze3xoAiXprdM7m)_

### DIFF
--- a/Game.DataAccess/DataProviderSynchronizer.cs
+++ b/Game.DataAccess/DataProviderSynchronizer.cs
@@ -424,9 +424,10 @@ namespace Game.DataAccess
                         break;
                     }
 
-                    var playerId = PlayerUpdateEnvelopeReader.TryReadPlayerId(next) ?? UnknownPlayerLane;
+                    var (envelope, parseError) = PlayerUpdateEnvelopeReader.TryParseEnvelope(next);
+                    var playerId = envelope is null ? UnknownPlayerLane : PlayerUpdateEnvelopeReader.TryReadPlayerIdFromPayload(envelope.Payload) ?? UnknownPlayerLane;
                     var previous = playerLanes.TryGetValue(playerId, out var existingLane) ? existingLane : Task.CompletedTask;
-                    var itemTask = ProcessReservedItemAsync(previous, next, queue, deadLetterQueue, concurrencyGate, cancellationToken);
+                    var itemTask = ProcessReservedItemAsync(previous, next, envelope, parseError, queue, deadLetterQueue, concurrencyGate, cancellationToken);
                     playerLanes[playerId] = itemTask;
                     inFlight.Add(itemTask);
 
@@ -535,12 +536,12 @@ namespace Game.DataAccess
         /// converging, so a stop mid-drain still ends at a bounded number of in-flight items rather than an
         /// unbounded one.
         /// </summary>
-        private async Task ProcessReservedItemAsync(Task previous, string message, IPubSubQueue queue, IPubSubQueue deadLetterQueue, SemaphoreSlim concurrencyGate, CancellationToken cancellationToken)
+        private async Task ProcessReservedItemAsync(Task previous, string message, DomainEventEnvelope? envelope, JsonException? parseError, IPubSubQueue queue, IPubSubQueue deadLetterQueue, SemaphoreSlim concurrencyGate, CancellationToken cancellationToken)
         {
             try
             {
                 await previous;
-                await ProcessMessage(message, deadLetterQueue, cancellationToken);
+                await ProcessMessage(message, envelope, parseError, deadLetterQueue, cancellationToken);
                 await queue.AcknowledgeAsync(message);
             }
             finally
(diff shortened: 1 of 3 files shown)


ginderjeremiah (commented):
## Review: Write-behind read-path efficiency — single envelope parse + batched progress miss-reload (#2378)

**Verdict:** Approve (posted as a comment since I'm the PR author) — both fixes from #2378 land cleanly, behavior is preserved, and the semantics that matter are held by the existing tests.

### Correctness / behavior

- **Envelope parse dedup** — `TryParseEnvelope` returns `(envelope, parseError)` and both are threaded through `ProcessReservedItemAsync` → `ProcessMessage`, with `message` kept only for acknowledge/dead-letter (which correctly key off the raw string). The three-way outcome is preserved exactly: `parseError` → malformed dead-letter (same log + exception), `envelope is null` → empty dead-letter, otherwise apply. Lane routing keeps its old semantics too (`envelope is null ? UnknownPlayerLane : TryReadPlayerIdFromPayload(...) ?? UnknownPlayerLane`). The dead-letter inspector still owns its own independent parse via `TryReadPlayerIdFromPayload`, so removing the old `TryReadPlayerId(string)` breaks no caller.
- **Progress miss-reload batching** — the three-table read collapses to one `UNION ALL` via a shared anonymous projection + `Concat`. Types unify correctly (`Value`/`Progress`/`Xp` are all `decimal` → `Amount`; `EntityId` is `int?`), every column is populated per row-kind so the reconstruction needs no `!`, and the switch rebuilds the same `CachedPlayerProgress` shape as before. It stays a flat projection (no navigation graph), consistent with `docs/backend-persistence.md`, and the lean models' `required` members still make a forgotten field a compile error.

### Test coverage

Adequate. This is a behavior-preserving refactor pinned by the existing `DataProviderSynchronizerTests` (lane-ordering + malformed/empty dead-letter paths) and `PlayerProgressRepositoryIntegrationTests`, and no new logic branch is introduced. The PR body also documents the real Postgres validation of the `Concat`-based UNION projection (the anonymous-type shape EF Core can translate for set operations, vs. the `record` shape it can't) — worth keeping in mind for any future set-operation query.

### What's good

- Both changes are tightly scoped to exactly what #2378 asked for, with no behavioral drift.
- The XML doc comments on `ProcessMessage`, `PlayerUpdateEnvelopeReader`, and `LoadFromDb` clearly explain the *why* (single upfront parse reused, cold-path-only UNION, defaulted-not-null columns), which is exactly the kind of context that keeps a later "simplification" from silently reintroducing the double-parse.

---

PR #2394, opened 2026-07-24
Hygiene sweep: statify Map mutation, BattleLost cooldown drift, stats-recording clamp, overview dead-zone, a11y (#2379)

### LINKED ISSUE #2379: Hygiene sweep: statify Map mutation, BattleLost cooldown response drift, unguarded double→decimal at stats recording, overview reward-strip dead-zone, minor a11y conformance
Small verified defects from the health review, bundled (per the #2245/#2345 sweep pattern). Each is independently fixable.

1. **`addUnlockedItem` mutates the `unlockedItems` Map in place** (`UI/src/lib/engine/player/inventory-manager.ts:374`), violating the statify Map/Set reassignment convention (`statify.svelte.ts` doc) that its sibling `addUnlockedMod` follows with an explanatory comment. Masked today because `publish()` reassigns `items`, but a future `$derived` keyed on Map contents alone would silently never re-derive — the #1809/#1957 bug class. Reassign a rebuilt Map and drop the "non-reactive Map" workaround note in `inventory-view.svelte.ts`.
2. **`BattleLost`'s rejection response drops the live cooldown `DefeatEnemy` reports.** `Game.Api/Sockets/Commands/BattleLost.cs:60-70` returns `Cooldown = 0` unconditionally, while `DefeatEnemy.cs:78-88` computes `state.IsOnCooldown(now) ? … : 0` on the equivalent branch. #2048 unified the two commands' validation; the response shape was left drifted — a reconnecting client re-presenting an already-credited loss is told 0 and wastes a `NewEnemy` round trip. Mirror the cooldown computation.
3. **Unguarded `(decimal)` casts at the battle-statistics recording seam can throw on Infinity.** `Game.Core/Progress/PlayerProgress.cs:71-86,147-148` cast unbounded doubles; `DamageDodged`/`DamageParried` tally computed damage of hits that never land (`BattleContext.cs:247,262`, no health cap), and a stacking multiplicative self-buff compounds per fire — authorable content (a permanent ×2 damage self-buff on a fast-cooldown enemy skill vs a forced-dodge player riding out the 2-minute draw) reaches `double.Infinity` and the post-battle recording throws. Exotic today, but it's a crash of victory/loss recording; clamp/guard at the recording seam (both simulators already agree on the battle itself, so this has no parity surface).
4. **Overview type-card click dead-zone.** `challenges/OverviewPane.svelte:59-65` raises `.type-card-reward` above the full-bleed `OverlayButton` whenever a next challenge exists, but when its reward is `null` the strip renders a non-interactive "No reward" span — clicks there hit the raised div and never reach the card's button. Only raise when there's an interactive reward affordance.
5. **Minor a11y conformance (login + class picker).** (a) `routes/+page.svelte` points every invalid field's `aria-describedby` at the single shared status line, which renders only the highest-priority message — a screen-reader user focusing the password field hears the username error as its description. (b) `select/ClassPicker.svelte:11-26` declares `radiogroup`/`radio` roles without the expected roving-tabindex/arrow-key model (every option is an independently Tab-focusable button). Fix per `docs/frontend-accessibility.md` conventions: per-field error association (or only describe the field the message belongs to), and either implement the radio keyboard pattern or drop to plain `aria-pressed` buttons.

ginderjeremiah:
## Summary

All five items from the #2379 sweep. Each was verified against the code before fixing (all five reproduced as described) and each carries a regression test that fails without its fix.

### 1. `addUnlockedItem` mutated the `unlockedItems` Map in place
`inventory-manager.ts` now reassigns a rebuilt Map, matching the statify Map/Set convention its sibling `addUnlockedMod` already documents. Masked today only because `publish()` reassigns `items` — a consumer deriving off the Map's contents alone (which the inventory view's selection/drag lookups do) would silently never re-derive. The "non-reactive Map" workaround note in `inventory-view.svelte.ts` is now obsolete and was removed rather than left to mislead.

### 2. `BattleLost` rejection response dropped the live cooldown
`BattleLost.cs`'s `else` branch returned `Cooldown = 0` unconditionally while `DefeatEnemy.cs`'s equivalent branch reported the live cooldown. The client honours `data.cooldown` on error responses (`enemy-manager.ts` reads `lostResponse.data?.cooldown` on the rejection path), so a reconnecting client re-presenting an already-credited loss was told 0 and burned a `NewEnemy` round trip.

Rather than mirror the expression by hand — the defect *is* one copy of it having drifted — the computation now lives on `PlayerState.RemainingCooldownMs(now)`, beside `IsOnCooldown`, and all four verbatim call sites (`BattleLost` ×2, `DefeatEnemy` ×2) plus `NewEnemy.cs:29` read through it.

### 3. Unguarded `double → decimal` casts at the battle-statistics recording seam
Ten casts in `PlayerProgress.RecordBattleCompleted` fed unbounded doubles to `(decimal)`, which throws on `Infinity`/`NaN`. The avoided-damage tallies are the realistic route: `DamageDodged`/`DamageParried` book the damage of hits that never land, so nothing caps them at the defender's health, and a stacking multiplicative self-buff compounds per fire.

All ten now go through one `ToStatisticValue` conversion seam:

- **A non-finite tally records nothing** (zero delta → no row, no dirty mark, no challenge re-evaluation). It deliberately does *not* saturate to a large number: that would be treated as a real value downstream — poisoning a monotone lifetime Sum, permanently locking a Max statistic where nothing can beat it, and completing every open damage-threshold challenge via the touched-key evaluation. It would also only defer the crash, since `Increment` accumulates and the running total overflows `decimal` (max ≈7.92e28) a few battles later, while the authored-content bug reproduces every battle.
- **A finite-but-out-of-range double is clamped** to ±1e28 — the one remaining way the cast itself can fail.
- **In-range values are byte-identical to before**: the `Math.Round(value, 3)` still happens on the double, before the conversion, so no stored figure shifts.

No parity surface: statistics recording runs off the tick loop, server-side only, and both simulators still agree on the battle itself.

### 4. Overview type-card click dead-zone
`OverviewPane.svelte` raised `.type-card-reward` above the full-bleed `OverlayButton` whenever a next challenge existed — but when that challenge's reward is `null`, `RewardAffordance` renders a non-interactive "No reward" span, so clicks landing on the strip hit the raised div and never reached the card's button. Now raised only when there's an interactive chip.

The raised region is also shrunk to the chip (`width: fit-content`, with `max-width: 100%` so long names still ellipsize) — the strip is a full-width block while the chip is `inline-flex`, so the empty space beside it was a narrower instance of the same dead zone.

### 5. A11y conformance

**(a) Login field descriptions.** Every invalid field pointed its `aria-describedby` at the single shared status line, which renders only the highest-priority message — so a screen-reader user focusing the password field heard the *username* error as its description. `deriveStatusLine` now also reports which field (if any) owns the message it chose, and only that field points at the line. The precedence rules and the visible UI are unchanged.

**(b) `ClassPicker` radiogroup.** It declared `radiogroup`/`radio` roles without the keyboard model those roles promise — every option was an independently Tab-focusable button. Implemented the WAI-ARIA radio pattern rather than dropping to `aria-pressed`, since this genuinely is single-select and the roles convey that better: a roving `tabindex` (one tab stop for the group), arrow keys moving focus and selection with wraparound, Home/End for first/last. The options stay real `<button>`s so Enter/Space keep working natively.

## Test plan

- [x] `dotnet test` (full solution) — 3490 passed, 0 failed
- [x] `npm run test:unit:ci` — 4000 passed, 0 failed
- [x] `npm run lint` (eslint + `svelte-check`) — clean, 0 errors/warnings
- [x] Each new guard verified to fail against the unfixed code (spot-checked the two subtlest: the Map-reactivity test and the reward-strip test)

New tests:
- `inventory-manager-reactivity.svelte.test.ts` — a `$derived` reading the Map alone observes a new unlock
- `PlayerProgressTests` — a `Theory` over `+∞`/`-∞`/`NaN` pinning that the tally records nothing while the battle's *other* statistics still record; that an existing lifetime Sum/Max is left untouched; and that a finite out-of-range double clamps
- `BattleLostSocketTests` — a rejection while on cooldown reports it
- `OverviewPane.test.ts` — the strip is raised only with an interactive chip, flat for both non-interactive states
- `login.test.ts` / `login-validation.test.ts` — only the owning field is described; ownerless messages claim no field
- `ClassPicker.test.ts` — single tab stop, arrow/Home/End navigation, wraparound, and other keys left to native handling

The `width: fit-content` change is not covered by a test: jsdom performs no layout, so any assertion would be vacuous.

## Docs

Added the radiogroup keyboard-model convention to `docs/frontend-accessibility.md` — it's a reusable pattern any future single-select group should follow, and the "don't declare the roles without the model" rule is the part worth writing down.

## Notes for review

- Both commands' *other* error branch (`!state.HasActiveBattle` → "No active battle") also returns `Cooldown = 0`, in `BattleLost` and `DefeatEnemy` alike. Deliberately left alone: unlike item 2 they aren't drifted from each other, and a client with no server-side battle needs a full resync rather than a cooldown. (Confirmed in review.)
- `NewEnemy.cs:54` and `ChallengeBoss.cs:61` look like further copies of the cooldown expression but have a different contract — nullable, `null` rather than `0` when no cooldown is in effect, anchored to `afterStart` rather than `now` — so they were left out of `RemainingCooldownMs` rather than have `null` quietly become `0` on the wire.

Closes #2379


Game.Core/Progress/PlayerProgress.cs
```diff
@@ -336,6 +340,36 @@ private void RecordKillByDamageType(BattleStats stats)
...
+            // statistic rows have always stored.
+            var rounded = Math.Round(value, 3);
+            if (rounded >= MaxRecordedValue)
+            {
+                return (decimal)MaxRecordedValue;
+            }
+            if (rounded <= -MaxRecordedValue)
+            {
+                return (decimal)-MaxRecordedValue;
+            }
```
ginderjeremiah, line 377: **Should-fix** — saturating ±∞ to ±1e28 trades a throw for permanent progression corruption, and doesn't actually stop the throw on recurrence.

Two problems with picking `MaxRecordedValue` as the landing spot for a non-finite tally:

1. **The accumulator still overflows.** The comment says 1e28 is "kept well inside decimal's range so the saturation itself can never overflow the conversion" — true of the *cast*, but `Increment` does `stat.Value += amount` and `decimal.MaxValue` is ≈7.92e28. Eight saturated battles put the lifetime total at 8e28 and the 8th `+=` throws `OverflowException` — from inside the same `RecordBattleCompleted` call this guard exists to protect. The scenario that produces an infinite tally is an authored-content bug (a permanent multiplicative self-buff), so it recurs *every* battle in that zone; this moves the crash from battle 1 to battle 9 rather than removing it. (The DB column is `numeric(36,3)`, so the CLR `decimal` is the binding limit, not the schema.)

2. **1e28 is a value, and downstream treats it as one.** A saturated `DamageDodged` lands in a Sum lifetime total that nothing can ever walk back, and `SetMax` locks `HighestSingleAttackDamage` at 1e28 forever (nothing can beat it). Worse, `RecordBattleCompleted`'s touched keys feed challenge evaluation — every damage-threshold challenge the player has open completes instantly and pays out its reward. Today's unguarded cast fails the battle's recording; this silently and irreversibly corrupts the player's lifetime stats and challenge progression.

The NaN branch above already has the right instinct: a value with no meaning records nothing. Suggest extending that to the non-finite case generally — `if (!double.IsFinite(value)) { return 0m; }` — and keeping the ±`MaxRecordedValue` clamp only for finite out-of-range doubles, which is all the cast actually needs protecting from. That fixes both points and lets the NaN test above generalize to `±Infinity`.


---

Game.Api/Sockets/Commands/BattleLost.cs
```diff
@@ -61,9 +61,13 @@ public override async Task<ApiSocketResponse<BattleLostResponse>> HandleExecuteA
...
                 await context.Session.SavePlayerStateAsync(cancellationToken);
 
+                var now = DateTime.UtcNow;
                 return ErrorWithData("Battle was not a loss.", new BattleLostResponse
                 {
-                    Cooldown = 0,
+                    // Report the live cooldown (as DefeatEnemy's equivalent branch does): the rejection is
+                    // typically a reconnecting client re-presenting an already-credited loss, and a hardcoded
+                    // 0 would send it straight into a NewEnemy round-trip the cooldown is about to reject.
+                    Cooldown = state.IsOnCooldown(now) ? (state.EnemyCooldown - now).TotalMilliseconds : 0,
```
ginderjeremiah: **Should-fix** — the fix is right, but re-fixing copy-paste drift with more copy-paste leaves the same trap.

`state.IsOnCooldown(now) ? (state.EnemyCooldown - now).TotalMilliseconds : 0` now appears four times verbatim (`BattleLost.cs:52,70`, `DefeatEnemy.cs:67,84`), and `NewEnemy.cs:29` / `ChallengeBoss.cs:61` are near-variants of the same computation. The defect this item fixes *is* one of those four having drifted — the next reader has no way to tell they're meant to be identical.

`PlayerState` is the natural home, right beside `IsOnCooldown`:

```csharp
/// <summary>Milliseconds left on the enemy cooldown at <paramref name="now"/>, or 0 if it has elapsed.</summary>
public double RemainingCooldownMs(DateTime now)
{
    return IsOnCooldown(now) ? (EnemyCooldown - now).TotalMilliseconds : 0;
}
```

That collapses all four call sites to `Cooldown = state.RemainingCooldownMs(now)` and puts the rule in the domain object that owns the cooldown, per the DRY guideline in `CLAUDE.md`. Not blocking, but this is the cheapest possible insurance against the exact drift being fixed here.


---

UI/src/routes/select/ClassPicker.svelte
```diff
@@ -90,6 +92,43 @@ const { classes, selectedClassId, onSelect, disabled = false }: Props = $props()
...
 const selected = $derived(classes.find((c) => c.id === selectedClassId) ?? null);
 
+/* ── radiogroup keyboard model ─────────────────────────────────────────────
+   A radiogroup is one tab stop, navigated internally with the arrow keys — so the group exposes a
+   single roving `tabindex=0` (the chosen option, or the first before anything is chosen) and the
+   arrows move focus between the options, selecting as they go per the WAI-ARIA radio pattern.
+   Enter/Space need no handling: these are real <button>s, so activation already fires onclick. */
+let optionsEl = $state<HTMLDivElement>();
+
+const ARROW_STEPS: Record<string, number> = { ArrowRight: 1, ArrowDown: 1, ArrowLeft: -1, ArrowUp: -1 };
```
ginderjeremiah: **Nit** — `Record<string, number>` tells TypeScript every key resolves to a `number`, so the `step !== undefined` guard on line 120 is dead code as far as the compiler is concerned (`noUncheckedIndexedAccess` isn't on). It's correct at runtime, but it's precisely the kind of construct a later "simplify" pass deletes — after which every keystroke moves selection by `NaN`.

`Partial<Record<string, number>>` makes the lookup honestly `number | undefined` and the guard load-bearing in the type system too.


---

UI/src/routes/game/screens/challenges/OverviewPane.svelte
```diff
@@ -56,7 +56,7 @@
 							<div class="mono-label sm" class:done={complete}>{stats.done}/{stats.total} unlocked</div>
 						</div>
 					</div>
-					<div class="type-card-reward" class:raised={next2}>
+					<div class="type-card-reward" class:raised={next2?.reward}>
```
ginderjeremiah: **Nit** — the fix is correct for the case the issue describes (raised iff `RewardAffordance` renders a `<button>`), but a narrower dead zone survives it: `.type-card-reward` is a block div spanning the card's full width, while the chip inside is `inline-flex` sized to its label. In the raised state, the empty space to the right of the chip is still above the overlay and still swallows clicks.

`width: fit-content` on `&.raised` confines the raised region to the chip and closes that remainder too (the chip's `max-width: 100%` keeps long names ellipsizing). Out of the literal scope of item 4, so a follow-up is fine — but it's one line in the block you're already touching.


---

ginderjeremiah (commented):
## Review: hygiene sweep of the five verified defects in #2379

**Intended verdict: Approve** — no blocking findings; all five items reproduce as described and each fix targets the real cause. Submitted as a comment because the review account is also the PR author, so GitHub won't take a gating verdict here. Two should-fixes below are worth landing before merge; the nits can ride.

I verified each item against the code rather than taking the issue at its word — all five hold up, including the two that read as speculative (the reward-strip dead zone and the `describedby` mis-association are both genuinely reachable today).

### Should-fix

- **`Game.Core/Progress/PlayerProgress.cs:361-368` — saturating ±∞ to ±1e28 swaps a crash for silent, unrecoverable corruption, and doesn't stop the crash on recurrence.** `Increment`'s `stat.Value += amount` overflows `decimal` (max ≈7.92e28) on the 8th saturated battle, so the throw returns at battle 9 — and the trigger is an authored-content bug that recurs every battle. Meanwhile 1e28 is treated as a real value downstream: it poisons a monotone Sum total, permanently locks the `SetMax` stats, and completes every open damage-threshold challenge (paying their rewards) through the touched-key evaluation. The NaN branch already models the right answer — a meaningless value records nothing. Extending that to `!double.IsFinite` and keeping the clamp only for finite out-of-range doubles fixes both. Detail inline.
- **`Game.Api/Sockets/Commands/BattleLost.cs:70` — the cooldown expression is now duplicated verbatim four times.** The defect being fixed *is* one of those four having drifted, so mirroring by copy-paste rebuilds the trap. A `PlayerState.RemainingCooldownMs(now)` beside `IsOnCooldown` collapses all four call sites and puts the rule on the object that owns the cooldown, per the DRY guideline in `CLAUDE.md`. Suggested implementation inline.

### Nits

- `UI/src/routes/select/ClassPicker.svelte:102` — `Record<string, number>` makes the `step !== undefined` guard invisible to the compiler; `Partial<Record<…>>` keeps it load-bearing.
- `UI/src/routes/game/screens/challenges/OverviewPane.svelte:59` — a narrower dead zone survives the fix: the raised strip is full-width while the chip is `inline-flex`, so the space beside the chip still eats clicks. `width: fit-content` on `&.raised` closes it. Fine as a follow-up.
- `class:raised={next2?.reward}` passes an object where the directive wants a boolean. Works, reads slightly loose — `!!next2?.reward` if you're touching the line anyway.

### Test coverage

Solid, and the "fails without the fix" claim holds for the ones I checked by inspection. Every touched module has mirrored tests: `inventory-manager-reactivity.svelte.test.ts` derives off the Map's contents alone (exactly the consumer shape that would have stayed stale), the `OverviewPane` tests pin all three strip states rather than just the new one, and the `ClassPicker` suite covers the whole radio contract including "leaves other keys to their native handling" — the assertion that stops a future edit from swallowing Space/Enter.

Two gaps, both minor and both dissolving if the `IsFinite` change above lands: no coverage of a **negative** non-finite tally (`-∞` currently records -1e28 into a lifetime total), and none of a finite-but-out-of-range double, which is the branch that would remain the clamp's only job. No battle-parity surface, correctly — statistics recording is server-side and off the tick loop, so there's no frontend mirror owed here.

### What's good

- Item 3's analysis is the strongest part: identifying the avoided-damage tallies as the reachable route (nothing caps them at the defender's health) is the non-obvious insight, and doing the `Math.Round` on the double *before* conversion so in-range values stay byte-identical is exactly right.
- Item 5(b) chose to implement the radio keyboard model rather than retreat to `aria-pressed`. That's the harder and better call for a genuinely single-select group, and writing the convention into `docs/frontend-accessibility.md` — with the "don't declare the roles without the model" rule — is the part that pays forward.
- Item 5(a)'s framing is right: making `deriveStatusLine` report the owning field keeps the precedence logic in the one pure, tested place instead of duplicating a "which message won?" guess in the template.
- The note about deliberately leaving both commands' `!state.HasActiveBattle` branches at `Cooldown = 0`: agreed, and for the reason given. Those two aren't drifted from each other, and a client with no server-side battle needs a resync rather than a cooldown. Leave them.


---

ginderjeremiah:
All five points addressed in 7b7d56e — both should-fixes and all three nits.

### Should-fix 1: saturation → degrade (`PlayerProgress`)

You're right, and this is the more important of the two. I checked both halves rather than taking them on faith:

- **The accumulator overflow is real.** `Increment` does `stat.Value += amount` against a `decimal` capped at ~7.92e28, so eight saturated battles put the total at 8e28 and the ninth `+=` throws — from inside the very call the guard existed to protect. And since the trigger is an authored-content bug (a permanent multiplicative self-buff), it recurs every battle in that zone rather than being a one-off. My fix moved the crash, it didn't remove it.
- **1e28-as-a-value is worse than the crash it replaced.** A saturated `DamageDodged` lands in a monotone Sum nothing can walk back; `SetMax` locks `HighestSingleAttackDamage` where nothing can ever beat it; and the touched keys feed challenge evaluation, so every open damage-threshold challenge completes and pays out. I traded a loud, recoverable failure for silent irreversible corruption of the player's progression. That's a straight regression against the status quo.

Adopted your suggestion exactly: `!double.IsFinite(value) → 0m` (no row, no dirty mark, no challenge re-evaluation), with the ±`MaxRecordedValue` clamp kept only for finite out-of-range doubles, which is genuinely all the cast can still fail on. The NaN branch's instinct generalised, as you said it would.

Test coverage follows the change and closes both gaps you named: the non-finite case is now a `Theory` over `+∞`/`-∞`/`NaN` (so the negative case is covered), plus a new case pinning that an existing lifetime Sum and Max are left **untouched** by a non-finite tally — the property that actually matters here — plus a finite-but-out-of-range case for the clamp's remaining job.

### Should-fix 2: DRY the cooldown expression

Agreed, and the irony was the point: re-fixing copy-paste drift with more copy-paste. Added `PlayerState.RemainingCooldownMs(now)` as you suggested and collapsed all four verbatim sites (`BattleLost` ×2, `DefeatEnemy` ×2), plus `NewEnemy.cs:29` — that one is the same `now`-anchored computation just already inside an `IsOnCooldown` guard, so it reads identically through the helper and it was the last hand-rolled `(EnemyCooldown - now).TotalMilliseconds` in the command layer.

I left `NewEnemy.cs:54` and `ChallengeBoss.cs:61` alone deliberately. They look similar but have a different contract — nullable, `null` rather than `0` when no cooldown is in effect, anchored to `afterStart` rather than `now`. Folding them into the same helper would either need a second nullable variant or would quietly change `null` to `0` on the wire. Say the word if you'd rather they were unified too.

### Nits

- **`Partial<Record<…>>`** — taken. The `step !== undefined` guard is now load-bearing to the compiler, so the "simplify pass deletes it and every keystroke moves by `NaN`" failure mode can't land silently.
- **`width: fit-content` on `&.raised`** — taken rather than deferred, since it's one line in a block already being touched and it's the same defect class, just narrower. Added `max-width: 100%` alongside it so a long reward name still ellipsizes rather than pushing the strip past the card. Not covered by a test: jsdom does no layout, so any assertion would be vacuous — the closest honest guard would be an e2e click at an offset, which isn't worth it for this.
- **`!!next2?.reward`** — taken.

### Verification

`dotnet build` clean (0 warnings), full solution `dotnet test` 3490/3490 (3 net new), `npm run test:unit:ci` 4000/4000, `npm run lint` (eslint + `svelte-check`) clean.

Thanks for catching the saturation one in particular — "the guard is worse than what it guards against" isn't something the tests I wrote would ever have surfaced, since they asserted the saturated value was *correct*.

---

Issue #2069, opened 2026-07-16
.claude/settings.local.json is committed and carries junk permission entries


ginderjeremiah:
## Context

Found in the July 2026 health review (repo hygiene pass).

`.claude/settings.local.json` is tracked in git (per `git ls-files`), but by Claude Code convention `settings.local.json` is personal/machine-local (normally gitignored on creation). It also carries junk entries — `Bash(Select-String "error")` and `Bash(Select-Object -First 20 FullName)` are PowerShell fragments, not real permission patterns.

## Suggested fix

Move the genuinely useful allows (`dotnet build/test/ef`, `npx svelte-check`, …) into the shared `.claude/settings.json`, then `git rm --cached .claude/settings.local.json` and gitignore it. Needs a quick owner pass over which entries are worth sharing.

ginderjeremiah:
Looked at picking this up — the fix is exactly as described (move the `dotnet build/test/ef` + `npx svelte-check` allows into `.claude/settings.json`, drop the PowerShell junk entries, `git rm --cached .claude/settings.local.json` + gitignore it). However, my edit to `.claude/settings.json` was blocked by this environment's tool permissions, the same way prior sessions' edits to `.claude/hooks/session-start.sh` (#1813) and `.vscode/settings.json` (#2212) were blocked — Claude Code config files that affect the dev environment appear to be specially guarded from autonomous edits here.

Leaving this open/assigned for a human to land directly:

```diff
 {
+  "permissions": {
+    "allow": [
+      "Bash(dotnet build *)",
+      "Bash(dotnet test *)",
+      "Bash(dotnet ef *)",
+      "Bash(npx svelte-check *)"
+    ]
+  },
   "hooks": {
```

Then:
```
git rm --cached .claude/settings.local.json
echo ".claude/settings.local.json" >> .gitignore
```

Picking up a different issue this session instead.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2395, opened 2026-07-24
Clear the session-start marker up front so it only describes this run

### LINKED ISSUE #1813: session-start hook leaves a stale .container-info.json on failure paths, wedging integration fixtures onto dead services
## Problem

`.claude/hooks/session-start.sh` removes an existing `$MARKER_FILE` only on the healthy-Docker path (`rm -f` around line 105). The mixed-success path (e.g. Docker postgres up but redis failed, native fallback also incomplete) and the terminal failure path (lines ~274-276) exit without removing an existing marker.

## Failure scenario

A prior session's marker (repo root, gitignored) survives into a session whose services fail to start. `PreexistingContainerInfo.TryLoad()` (`Game.TestInfrastructure/Fixtures/PreexistingContainerInfo.cs:36-52`) finds the stale marker, the fixtures skip Testcontainers, and every integration test fails on dead connection strings — with a misleading cause that looks like a test/infra bug rather than a startup failure.

## Suggested fix

`rm -f "$MARKER_FILE"` at the top of the hook (or on every failure/partial path) so the marker only ever describes *this* session's services.

### LINKED ISSUE #2212: .vscode/settings.json terminal cwd uses a Windows-only path separator
`.vscode/settings.json:19` sets `"terminal.integrated.cwd": "${workspaceFolder}\\UI"`. The literal backslash resolves to a nonexistent path on Linux/macOS (including remote sessions), so the integrated terminal falls back with a warning instead of opening in `UI/`. `${workspaceFolder}/UI` works on all platforms, Windows included. One-character fix.

ginderjeremiah:
Fixes #1813. Also fixes #2212.

## Problem

`.claude/hooks/session-start.sh` removed `.container-info.json` only on the healthy-Docker early-exit path. Two other paths exit without touching it:

- **Partial bring-up** — Docker Postgres comes up but Redis fails, and the native fallback is also incomplete.
- **Terminal failure** — nothing available at all.

In both cases a previous session's marker survives. `PreexistingContainerInfo.TryLoad()` finds it, the fixtures skip Testcontainers, and every integration test fails against dead connection strings — which reads as a test-infra bug rather than a startup failure.

## Changes

- **Move the cleanup to the top of the hook**, right after `MARKER_FILE` is defined, and drop the now-redundant `rm -f` on the healthy-Docker path. Only the two success paths write a marker, so it always describes the current run — and a failed or partial session leaves none.
- **Add `.claude/hooks/session-start.test.sh`.** It drives the hook's real code paths, stubbing only the external services (`docker` / `postgres` / `redis`) behind a restricted `PATH`, and asserts the marker lifecycle across four cases: terminal failure, partial bring-up, full Docker success, and the Testcontainers path. Each case seeds a stale marker first, so a surviving one is detectable. Verified the two failure cases fail against the unfixed hook and pass against the fixed one.
- **Run it in CI** via a new `test-hooks` job, gated on a `hooks` paths-filter (`.claude/hooks/**`) and wired into the `all-checks-passed` gate. It needs nothing but `bash`, so it's a cheap job.
- **Document the invariant** in `docs/infrastructure.md` — the marker only ever describes the run that wrote it — replacing the now-inaccurate "no-op and removes any stale marker" phrasing.
- **`.vscode/settings.json`** (#2212): `terminal.integrated.cwd` used a Windows-only backslash (`${workspaceFolder}\\UI`), which resolves to a nonexistent path on Linux/macOS. Forward slashes work on every platform, Windows included.

## Notes

Both issues carry prior-session comments reporting that autonomous edits to these files were blocked by tool permissions. That wasn't the case this session — the edits went through, so both fixes are landed here together rather than left as diffs for a human to apply. #2069 (the third file in that group) is deliberately not included: it asks for an owner pass over which permission entries are worth sharing, which isn't a mechanical fix.

## Testing

- `bash .claude/hooks/session-start.test.sh` — 4 passed, 0 failed.
- Same suite against the pre-fix hook — 2 passed, 2 failed (the terminal-failure and partial-bring-up cases), confirming the tests bind to the defect.
- `.github/workflows/run-tests.yml` parses and the new job is wired into `all-checks-passed`.

---
_Generated by [Claude Code](https://claude.ai/code/session_017ZzSsU2JRNnFGhyDFnAmGF)_


.claude/hooks/session-start.test.sh
```diff
@@ -0,0 +1,128 @@
...
+  rm -rf "$root"
+  mkdir -p "$root/.claude/hooks"
+  cp "$HOOK_SRC" "$root/.claude/hooks/session-start.sh"
+  printf '%s' "$STALE_MARKER" > "$root/.container-info.json"
+  echo "$root"
+}
+
+run_hook() {
+  CLAUDE_CODE_REMOTE=true PATH="$STUB" "$STUB/bash" "$1/.claude/hooks/session-start.sh" > "$1/out.log" 2>&1
+}
```
ginderjeremiah, line 51: **Should-fix:** `run_hook` isolates `PATH` but not `HOME`, so every case runs the hook's real nvm/Node block (`session-start.sh:32-59`) against the *host* — `sed -i /etc/profile.d/nvm.sh`, appending `nvm use default` to it, and `ln -sf` into `$HOME/.local/bin`.

Today that's mostly neutered (the restricted PATH has no `sed`/`ln`/`mkdir`, and `nvm which default` comes back empty without `find`/`sort`), so it no-ops on CI and is idempotent in a remote container. But a test suite that writes to `/etc` and `$HOME` on the machine running it is a footgun: it makes the outcome host-dependent, and a future edit to that block could quietly corrupt a developer's environment on a `bash session-start.test.sh`.

One line closes it:

```bash
run_hook() {
  HOME="$1/home" CLAUDE_CODE_REMOTE=true PATH="$STUB" "$STUB/bash" "$1/.claude/hooks/session-start.sh" > "$1/out.log" 2>&1
}
```

(with `mkdir -p "$root/home"` in `setup_case`). That also makes the nvm block deterministically inert rather than incidentally inert.

---

.claude/hooks/session-start.test.sh
```diff
@@ -0,0 +1,128 @@
...
+
+HOOK_SRC="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/session-start.sh"
+WORK="$(mktemp -d)"
+STUB="$WORK/stubbin"
+mkdir -p "$STUB"
+trap 'rm -rf "$WORK"' EXIT
+
+# A restricted PATH is what gives the tests control over `command -v docker/dockerd/pg_isready/
+# redis-server`: only what is linked here is visible to the hook.
+for cmd in bash cat rm seq sleep awk grep dirname ip sudo; do
```
ginderjeremiah, line 27: **Nit:** `ip` and `sudo` are linked to the *real* binaries, but both are only reachable on paths the stubs already close off — `ip link delete docker0` needs `dockerd` on PATH (it isn't), and `sudo -u postgres psql` needs `pg_isready` (also absent). So they're unused today, and if a future case makes the native-Postgres path reachable (see the coverage note in the summary), the hook would run `sudo -u postgres psql -c "CREATE USER ..."` against the developer's actual machine.

Safer to drop them from this list and write no-op stubs instead, so the "real binary" affordance never exists:

```bash
for cmd in ip sudo; do
  printf '#!/usr/bin/env bash\nexit 0\n' > "$STUB/$cmd"
  chmod +x "$STUB/$cmd"
done
```

---

.claude/hooks/session-start.test.sh
```diff
@@ -0,0 +1,128 @@
...
+assert_marker() {
+  local name="$1" expectation="$2" marker="$3/.container-info.json"
+  if [[ "$expectation" == "gone" ]]; then
+    if [[ ! -e "$marker" ]]; then
+      echo "  PASS: $name — stale marker removed"
+      ((passed++))
+    else
+      echo "  FAIL: $name — stale marker survived: $(tr -d '\n' < "$marker")"
+      ((failed++))
+    fi
```
ginderjeremiah, line 63: Two nits in this block:

1. **`((passed++))` returns exit status 1 on the first increment** (post-increment of `0` evaluates to `0`, which `((...))` reports as false). Harmless right now because the suite doesn't `set -e`, but it's exactly the kind of construct a later "let's make this strict" edit turns into a silent early exit. `passed=$((passed + 1))` has no such trap.

2. **The FAIL branch never surfaces the hook's output.** `run_hook` captures it to `$1/out.log`, and this job exists specifically to diagnose hook regressions in CI — but on failure all you get is the marker contents, and the temp dir is wiped by the `EXIT` trap before you can look. Worth dumping it:

```bash
echo "  FAIL: $name — stale marker survived: $(tr -d '\n' < "$marker")"
sed 's/^/    | /' "$3/out.log"
```

(`assert_marker` would need the root rather than deriving `marker` from it, or just reuse `$3`.)

---

.claude/hooks/session-start.test.sh
```diff
@@ -0,0 +1,128 @@
...
+  elif [[ -e "$marker" ]] && ! grep -q STALE "$marker"; then
+    echo "  PASS: $name — marker describes this run"
+    ((passed++))
```
ginderjeremiah, line 66: **Nit:** the `fresh` assertion is only "a marker exists and it isn't the stale one" — it would pass on a marker written with the wrong ports or a truncated `write_marker` heredoc. Cheap to tighten to the thing that actually matters to `PreexistingContainerInfo.TryLoad()`:

```bash
elif [[ -e "$marker" ]] && grep -q "Port=$EXPECTED_PG_PORT" "$marker" && grep -q "localhost:$EXPECTED_REDIS_PORT" "$marker"; then
```

This matters more once the native-fallback path is covered, since that one derives its Postgres port dynamically from `pg_lsclusters` rather than using the fixed `POSTGRES_PORT`.

---

ginderjeremiah (commented):
## Review: clear `.container-info.json` up front so the marker only ever describes the current run (#1813), plus the `.vscode` path-separator fix (#2212)

**Verdict:** Approve (submitted as `COMMENT` — GitHub won't take a gating verdict from the PR's own author) — the fix is correct and minimal, the tests genuinely bind to the defect, and the CI wiring matches the existing job pattern. Nothing blocking; a handful of test-harness cleanups below that are worth folding in before merge.

### Verified

- **The fix is right.** `rm -f "$MARKER_FILE"` at `session-start.sh:68` sits after `MARKER_FILE` is defined and before every one of the four exits, and all three marker-writing sites go through `write_marker`. So the invariant holds on all paths — including the two the issue called out (`session-start.sh:277-279` terminal failure, and the partial bring-up that falls through `199` and `271` without writing). Dropping the now-redundant `rm -f` on the healthy-Docker path is the right cleanup rather than leaving both.
- **The reuse case still works.** A second session in a container whose containers survived deletes the marker, hits `container_running` for both, and rewrites it at `session-start.sh:200`. No regression there.
- **CI wiring is consistent.** The `hooks` filter mirrors `backend`/`frontend` in also listing `.github/workflows/**`, the `if:` guard uses the same `always() && !cancelled() && (push || filter)` shape as its siblings, and `test-hooks` is in `all-checks-passed`'s `needs` so it's covered by the single required check. It ran in 9s.

### Should-fix

- `.claude/hooks/session-start.test.sh:49-51` — `run_hook` doesn't isolate `HOME`, so each case drives the hook's real nvm/Node block against the host machine (`/etc/profile.d/nvm.sh`, `$HOME/.local/bin`). Inert in practice today, but only incidentally — see the inline comment.

### Nits

- `.claude/hooks/session-start.test.sh:27` — `ip`/`sudo` linked to real binaries where no-op stubs would do.
- `.claude/hooks/session-start.test.sh:56-63` — `((passed++))` exit-status trap, and the FAIL branch doesn't dump `out.log` before the `EXIT` trap wipes it.
- `.claude/hooks/session-start.test.sh:64-66` — the `fresh` assertion checks only "not STALE", not that the marker's ports match the run.
- `docs/infrastructure.md` — "The hook deletes any inherited marker up front" reads as unconditional, but the `CLAUDE_CODE_REMOTE != true` early exit at `session-start.sh:22` returns before the `rm`. Practically irrelevant (a non-remote session never writes one either), so a two-word qualifier at most.

### Test coverage

The four cases are the right ones for the reported bug, and they're well chosen — driving the hook's real code paths and stubbing only the external services is much more durable than asserting on a refactored-out function. Confirming the two failure cases fail against the unfixed hook is the part that makes them worth having.

One gap: **the native-fallback success path (`session-start.sh:271-275`) is the second marker-writing path and it's uncovered.** It's also the more fragile of the two — it derives `NATIVE_POSTGRES_PORT` from `pg_lsclusters` output via `awk` in two different places (`218` and `225`) with a `5432` default, so a marker with the wrong port is a live failure mode there in a way it isn't on the fixed-port Docker path. The harness is already built for this; it needs `pg_isready`/`pg_lsclusters`/`psql`/`redis-server`/`redis-cli` stubs and a fifth case. Worth adding here rather than deferring, since it's the same lifecycle this PR is pinning down — but a follow-up issue is a reasonable call if you'd rather keep the diff tight.

Not raised as findings, for the record: no test asserts the hook's exit code is always 0 (its documented best-effort contract), and the pre-existing hole where Docker-Postgres + native-Redis can never combine into a marker is out of scope here.

### What's good

- The comment at `session-start.sh:65-67` explains *why* the deletion is at the top and what depends on it — that's the thing that stops a later reader from "tidying" it back down into the Docker branch.
- The heredoc-escaping note in `write_docker_stub` is the kind of thing that's obvious while writing it and completely opaque six months later.
- Landing both issues together is the right call given they're one-line fixes in adjacent infra files, and the note about deliberately excluding #2069 (an owner judgment call, not a mechanical fix) is the correct scope line.

---

ginderjeremiah:
## Review: approve ✅

Reviewed the full diff against #1813 and #2212, and verified the fix and its tests locally rather than taking the PR description at its word.

### Verification

Checked out `3141c7d` into a scratch worktree and ran the new suite:

```
=== session-start.sh marker lifecycle ===
  PASS: terminal failure (no docker, no native services) — stale marker removed
  PASS: mixed success (postgres up, redis down) — stale marker removed
  PASS: docker success (both services up) — marker describes this run
  PASS: healthy docker (testcontainers path) — stale marker removed
=== 4 passed, 0 failed ===
```

Then swapped in the pre-fix hook (`f8ae585:.claude/hooks/session-start.sh`) under the same harness:

```
  FAIL: terminal failure (no docker, no native services) — stale marker survived: {…"Database=STALE"…}
  FAIL: mixed success (postgres up, redis down) — stale marker survived: {…"Database=STALE"…}
  === 2 passed, 2 failed ===
```

So the tests genuinely bind to the defect — they're not passing for incidental reasons. CI is green across all 11 checks, including the new `test-hooks` job.

### The fix

Moving `rm -f "$MARKER_FILE"` to just after `MARKER_FILE` is defined is the right shape, and the placement matters in a way worth calling out: it sits **after** the `CLAUDE_CODE_REMOTE` guard, so local dev and CI sessions still exit at line 22 and never touch a marker they didn't create. Dropping the now-redundant `rm -f` on the healthy-Docker path is correct — no dead code left behind. Traced every exit after that point: the two success paths are the only writers, so the stated invariant ("the marker only ever describes the run that wrote it") actually holds.

### The tests

The harness design is good. Driving the hook's real code paths and stubbing only at the process boundary (`docker`) via a restricted `PATH` is far more valuable than asserting on a refactored-out function, and the restricted `PATH` doubles as control over the `command -v dockerd/pg_isready/redis-server` probes — a neat trick that gets the terminal-failure case for free. Seeding a stale marker in every case is what makes "survived" detectable rather than "never existed," which is the whole point.

Verified the stub's interpolation/escaping is right (`\$1` vs `$storage`), that the `failing_run=redis` substring can't accidentally match the Postgres `docker run` line, and that no `set -e` is present — which matters, since `((passed++))` returns 1 on the first increment and would otherwise abort the suite.

### CI wiring

`test-hooks` is correctly added to the `all-checks-passed` `needs` list — a gate that only *appears* to cover a job is worse than no job, so worth confirming. The `if:` condition and the `hooks` paths filter (including `.github/workflows/**`) both match the existing per-job conventions.

### Docs and `.vscode`

`docs/infrastructure.md` correctly drops the now-inaccurate "no-op and removes any stale marker" phrasing and states the invariant. The new sub-bullet leans a little toward implementation detail (naming both the test file and the CI job), which the docs guidance warns against — but it buys real discoverability for a file nobody would otherwise find, so I'd leave it. The `.vscode` separator fix is correct and it's the only Windows-only path in that file.

Bundling #2212 into a hook-fix PR is mild scope mixing, but it's a one-character config fix with a clear rationale in the description. Not worth a separate PR.

### One observation — not a change request

The hook's Node/nvm prelude still runs during the tests. The restricted `PATH` neuters most of it (`sed`/`mkdir`/`ln` aren't visible), but `echo … >> /etc/profile.d/nvm.sh` is a builtin and escapes the sandbox, so running the suite as root on a machine with that file writes to it. Impact is essentially nil — the write is idempotent, matches what the hook does for real, and CI runners have no such file — and it **can't be fixed from the harness**, since the path is hardcoded and not `$HOME`-relative. That makes it a structural point about the hook, not something to fix here.

Filed as #2396 (tech debt, scope: small): split the prelude into its own sourced file so provisioning runs hermetically, which also unblocks covering the native-fallback success path — the one remaining marker-writing path with no test.

### Verdict

Approve. Correct fix, tests that demonstrably catch the regression, sound CI wiring, docs updated. Merging.

---

Issue #2397, opened 2026-07-24
Harden session-start.test.sh: host isolation, assertion strength, and native-fallback coverage


ginderjeremiah:
Follow-up to the review on #2395 (Fixes #1813). The review was an approve-with-nits, but the PR merged before the findings were folded in, so they're carried here. Nothing below affects the shipped fix — `.claude/hooks/session-start.sh` is correct as merged; this is all test-harness quality in `.claude/hooks/session-start.test.sh`.

## 1. Isolate `HOME` (should-fix)

`run_hook` restricts `PATH` but not `HOME`, so every case drives the hook's real nvm/Node block (`session-start.sh:32-59`) against the machine running the suite. Fix:

```bash
run_hook() {
  HOME="$1/home" CLAUDE_CODE_REMOTE=true PATH="$STUB" "$STUB/bash" "$1/.claude/hooks/session-start.sh" > "$1/out.log" 2>&1
}
```

with `mkdir -p "$root/home"` in `setup_case`.

**Caveat worth recording:** `HOME` isolation closes the `$HOME/.local/bin` symlinking, but **not** all of it. The second block is gated only on `[ -f /etc/profile.d/nvm.sh ]`, independent of `HOME`, and the append uses shell redirection (`>>`) rather than a binary the restricted `PATH` can withhold:

```bash
if [ -f /etc/profile.d/nvm.sh ] && ! grep -q "nvm use default" /etc/profile.d/nvm.sh 2>/dev/null; then
  echo '...' >> /etc/profile.d/nvm.sh
```

So on a host where that file exists but lacks the line, the suite appends to it. Blast radius is small — the file doesn't exist on a stock CI runner or a typical dev machine, it already carries the line in the remote container (making the append a no-op), and writing `/etc` needs root anyway — but it's not closable from the test side. Options: accept and document it, or make the hook's environment-fixing block skippable (e.g. honor an opt-out variable) so the test can drive only the service-provisioning half. The latter adds a test-only seam to production code, which is why it wasn't done in #2395.

## 2. Cover the native-fallback success path

`session-start.sh:271-275` is the second of the two marker-writing paths and is untested. It's the more fragile one: `NATIVE_POSTGRES_PORT` is derived from `pg_lsclusters` output through `awk` in two separate places (`218`, `225`) with a `5432` default, so "marker written with the wrong port" is a live failure mode there in a way it isn't on the fixed-port Docker path.

[...]

Issue #2396, opened 2026-07-24
Split the Node/nvm prelude out of session-start.sh so provisioning is independently testable


ginderjeremiah:
Follow-up from #2395, which added `.claude/hooks/session-start.test.sh`.

## Problem

`.claude/hooks/session-start.sh` does two unrelated jobs in one script:

1. **Node/nvm environment fixup** (lines ~26-59) — rewrites `/etc/profile.d/nvm.sh` and symlinks node/npm into `$HOME/.local/bin`.
2. **Backing-service provisioning** (lines ~61-277) — dockerd + Postgres/Redis + the `.container-info.json` marker.

Only (2) is what the integration fixtures care about, and only (2) has tests. The new harness sandboxes the hook behind a restricted `PATH`, which neuters most of (1) incidentally — `sed`/`mkdir`/`ln` simply aren't visible — but not all of it: the append at

```bash
echo 'nvm use default > /dev/null 2>&1  # Activate default Node version' >> /etc/profile.d/nvm.sh
```

uses the `echo` builtin, so it escapes the sandbox. Running the test suite as root on a machine that has a writable `/etc/profile.d/nvm.sh` mutates it. The write is idempotent and matches what the hook does for real, so the practical impact is ~nil (and CI runners have no such file), but a test suite should not be able to touch the host at all. It can't be fixed from the harness either — the path is hardcoded and not `$HOME`-relative.

## Suggested direction

Extract the prelude into its own file (e.g. `.claude/hooks/lib/setup-node.sh`) that `session-start.sh` sources, so the harness can stub or skip it and the provisioning logic runs fully hermetically. Two smaller wins fall out of that:

- The hook drops from ~277 lines to something with a single responsibility.
- The test suite can then cover the **native-fallback success path** (`pg_isready`/`pg_ctlcluster`/`redis-server` stubbed), which currently has no coverage — it's the one remaining path that writes a marker.

## Scope

Small — shell-only, no application code, no user-facing behavior change.

PR #2398, opened 2026-07-25
Harden the session-start test harness and cover the native-fallback path

ginderjeremiah:
Follow-up to #2395, which merged before I could push these. Every item is from that PR's review; no production code changes, only the test harness and one doc qualifier.

## Should-fix

- **Isolate `HOME` per case.** `run_hook` restricted `PATH` but not `HOME`, so each case drove the hook's nvm/Node block (`session-start.sh:32-59`) against the host — gated on `$HOME/.nvm/nvm.sh`, symlinking into `$HOME/.local/bin`. Confirmed both gates are live in the remote container (`/root/.nvm/nvm.sh` exists), so the block was inert only because the restricted `PATH` happens to lack `sed`/`ln`/`mkdir`. A throwaway `$HOME` under the case directory makes it deterministically inert instead of incidentally so.

## Nits

- **`ip` is a no-op stub** rather than a symlink to the real binary — `ip link delete docker0` is host-mutating and nothing here needs its effect. **`sudo` strips a leading `-u` and its username argument, then execs the rest**, so the hook's three `sudo -u postgres psql` call sites genuinely reach the psql stub instead of being swallowed. That's safe because `PATH` is the stub dir only, so the exec cannot resolve to a real binary.
- **Failure output now dumps `out.log`.** It was captured and then wiped by the `EXIT` trap, leaving CI failures undiagnosable. This paid off immediately: on its first run it showed my own `redis-server` stub calling `touch`, which isn't on the restricted `PATH`.
- **`passed=$((passed + 1))` instead of `((passed++))`** — post-incrementing from zero reports false, which a later `set -e` would turn into a silent early exit.
- **`fresh` asserts the actual ports**, not just "not the stale one", so a wrong-port or truncated `write_marker` fails.
- **Stub reset lives in `setup_case`**, which clears every service stub plus the sentinels/logs they share. Each case declares its stubs after that call, so cases are self-contained and order-independent rather than relying on paired cleanup at each call site.
- **`docs/infrastructure.md`** — "deletes any inherited marker up front" now says "once past the `CLAUDE_CODE_REMOTE` guard".

## Coverage gap

Adds the **native-fallback case** (`session-start.sh:271-275`), the second marker-writing path and the more fragile one: its Postgres port is parsed out of `pg_lsclusters` via `awk` rather than fixed. The stub advertises **5433**, deliberately not the `5432` default, so the assertion fails if the hook falls back to the default instead of using the discovered port.

Both services start out *down* and come up only once the hook has started them, so the case covers the start paths rather than the already-running shortcuts: `pg_isready` fails until a `pg_ctlcluster` stub creates its sentinel, and `redis-cli` stays silent until `redis-server` creates its own. That also puts the `NR==1{print $1}` cluster-version parse under test.

`pg_ctlcluster` and `psql` log their argv, and `assert_logged` checks the version handed to the former and the port handed to the latter — both parsed from the same `pg_lsclusters` line. The marker's port assertion passes even when either field is misread, so it did not cover them.

## Not addressed

The `/etc/profile.d/nvm.sh` append (`session-start.sh:36-39`) is gated only on that file existing and uses shell builtins (`echo` and append redirection), so `HOME` isolation doesn't close it and nothing on the test side can. Blast radius is small — it needs root, the file only exists on machines with a system nvm install, and the line it appends is one the hook already appends on every session start (idempotent). Flagging rather than fixing, since closing it properly means a guard in the hook itself.

## Testing

- `bash .claude/hooks/session-start.test.sh` — 7 passed, 0 failed.
- Against the pre-fix hook (`HEAD~1` at the time, before #2395 merged) — 3 passed, 2 failed, so the terminal-failure and partial-bring-up cases still bind to the original defect.
- Mutation-tested the two native assertions: changing the version column to `$2` and hardcoding the `psql` port to `5432` each fail them, while the marker assertion still passes in both cases.
- Verified `HOME` isolation leaves the real `~/.local/bin` untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_017ZzSsU2JRNnFGhyDFnAmGF)_


.claude/hooks/session-start.test.sh
```diff
@@ -94,11 +130,32 @@ EOF
...
 
+# Stubs the natively-installed services the hook falls back to. Postgres reports itself already
+# running on $NATIVE_PG_PORT (advertised only via pg_lsclusters, so the hook has to read it from
+# there); Redis starts out down and only answers PONG once `redis-server` has "daemonized", which
+# exercises the start path rather than the already-running shortcut.
+write_native_stubs() {
+  printf '#!/usr/bin/env bash\nexit 0\n' > "$STUB/pg_isready"
+  # Columns match `pg_lsclusters -h`: version, cluster, port, status, owner.
+  printf '#!/usr/bin/env bash\necho "18 main %s online postgres"\n' "$NATIVE_PG_PORT" > "$STUB/pg_lsclusters"
+  printf '#!/usr/bin/env bash\nexit 0\n' > "$STUB/psql"
```
ginderjeremiah: Nit — this stub is unreachable, so it's dead code (`CLAUDE.md`: don't leave dead code behind).

The hook never calls `psql` directly; all three call sites are `sudo -u postgres psql ...` (`session-start.sh:231-236`). Since the `sudo` stub added above is `exit 0` and never execs its argv, `$STUB/psql` is never invoked — and `command -v psql` isn't checked anywhere either.

Two ways out, either is fine:

- Drop the `psql` stub (and its entries in the `chmod`/`clear_native_stubs` lists).
- Or make the `sudo` stub strip a leading `-u <user>` and exec the rest, so `psql` is genuinely exercised. That doesn't reopen the "path to the genuine article" concern the comment above worries about, because `PATH` is `$STUB` only — the exec can't resolve to anything but another stub. This is the better option if you want the `CREATE USER` branch actually covered rather than just short-circuited.

---

.claude/hooks/session-start.test.sh
```diff
@@ -94,11 +130,32 @@ EOF
   chmod +x "$STUB/docker"
 }
 
+# Stubs the natively-installed services the hook falls back to. Postgres reports itself already
+# running on $NATIVE_PG_PORT (advertised only via pg_lsclusters, so the hook has to read it from
+# there); Redis starts out down and only answers PONG once `redis-server` has "daemonized", which
+# exercises the start path rather than the already-running shortcut.
+write_native_stubs() {
+  printf '#!/usr/bin/env bash\nexit 0\n' > "$STUB/pg_isready"
```
ginderjeremiah: Nit — coverage asymmetry between the two native services, worth a line in the comment above at minimum.

`pg_isready` unconditionally exits 0, so the hook takes the *already-running* shortcut (`session-start.sh:224-227`) and the start branch (`215-223`) never runs — meaning `pg_ctlcluster` and the `NR==1{print $1}` version parse are untested. Redis deliberately does the opposite (starts down, comes up after `redis-server`), and the comment sells that as the point of the case, so the Postgres half reads as if it's doing the same thing when it isn't.

Both branches read the port from the same `pg_lsclusters` line, so the port assertion still binds either way — this is about the cluster-start path, not the port. Cheapest fix is to have `pg_isready` gate on a sentinel the way `redis-cli` does (`[[ -e "$WORK/pg-up" ]]`) and add a `pg_ctlcluster` stub that creates it, which exercises the version parse too. If you'd rather keep the case simple, just say in the comment that Postgres takes the already-running path on purpose.

---

.claude/hooks/session-start.test.sh
```diff
@@ -94,11 +130,32 @@ EOF
...
+  # `: >` rather than touch: the stubs run under the same restricted PATH as the hook.
+  printf '#!/usr/bin/env bash\n: > "%s/redis-up"\nexit 0\n' "$WORK" > "$STUB/redis-server"
+  printf '#!/usr/bin/env bash\n[[ -e "%s/redis-up" ]] && echo PONG\nexit 0\n' "$WORK" > "$STUB/redis-cli"
+  chmod +x "$STUB/pg_isready" "$STUB/pg_lsclusters" "$STUB/psql" "$STUB/redis-server" "$STUB/redis-cli"
+  rm -f "$WORK/redis-up"
+}
+
+clear_native_stubs() {
+  rm -f "$STUB/pg_isready" "$STUB/pg_lsclusters" "$STUB/psql" "$STUB/redis-server" "$STUB/redis-cli"
+}
```
ginderjeremiah: Nit — the stub directory is now shared mutable state across cases, cleaned up by hand at each call site. That's a footgun for whoever adds case 6.

The invariant is "each case sees exactly the stubs it declares", but it's enforced by remembering to pair every `write_*_stubs` with an `rm -f "$STUB/docker"` / `clear_native_stubs`. The current calls are already slightly off from that: line 158's `clear_native_stubs` is a no-op (nothing has written them yet at case 1), and line 192's is only correct because case 5 happens to be last. A case inserted in the wrong place silently inherits stubs and asserts against the wrong hook path — and it'd pass or fail for reasons unrelated to what it's testing.

Suggest folding the reset into one place: a `reset_stubs()` that clears docker + the native set, called at the top of each case. `setup_case` would be the natural home, but it's currently invoked *after* the `write_*` calls, so either reorder those two lines per case or keep `reset_stubs` as its own call. Then each case is a self-contained declaration of what exists and the scattered `rm -f`/`clear_native_stubs` calls go away.

---

ginderjeremiah (commented):
## Review: test-harness hardening for `session-start.sh` + coverage for the native-fallback marker path

**Verdict:** Approve (posted as a comment — GitHub won't take a gating verdict from the PR author) — no blocking findings. Three nits below, plus one follow-up filed for the gap the description flags as unaddressed.

I verified the description's claims against the hook rather than taking them at face value; they hold up.

### Blocking

None.

### Nits

- **Unused `psql` stub** — `session-start.test.sh:141`. The hook only ever reaches `psql` through `sudo` (`session-start.sh:231-236`), and the new `sudo` stub exits 0 without exec'ing its argv, so this stub is unreachable. Dead code per `CLAUDE.md`. Inline comment has two ways out; making `sudo` exec through is the one that actually buys coverage of the `CREATE USER` branch.
- **Postgres takes the already-running shortcut while Redis exercises the start path** — `session-start.test.sh:138`. `pg_isready` unconditionally succeeds, so `pg_ctlcluster` and the cluster-version parse (`session-start.sh:215-223`) stay uncovered. The comment above reads as though both services cover their start path. Either gate `pg_isready` on a sentinel the way `redis-cli` is, or say in the comment that the shortcut is deliberate.
- **Stub lifecycle is shared mutable state cleaned up by hand** — `session-start.test.sh:149-151`. Line 158's `clear_native_stubs` is already a no-op and line 192's is only right because case 5 is last; a case inserted in the wrong spot silently inherits stubs. A single `reset_stubs` per case would make each case self-declaring.

### Follow-up (out of scope here)

The `/etc/profile.d/nvm.sh` append the description flags under "Not addressed" — your reasoning is right that nothing on the test side closes it, since the hook hardcodes the path. Filed as #2399. Worth noting the fix is smaller than a new guard: block `session-start.sh:36-39` is arguably mis-gated on its own merits — appending `nvm use default` to a profile whose `NVM_DIR` has no installed versions does nothing, so putting it under the same `[ -f "$HOME/.nvm/nvm.sh" ]` test as the block above it is a correctness improvement that closes the harness hole as a side effect, with no test seam in production code. Leaving it out of this PR is the right call given it's the one file you deliberately didn't touch.

### Test coverage

Appropriate — this PR *is* test coverage, and the new case targets the right thing. The native fallback is the second marker-writing path and the more fragile one, since its port is `awk`'ed out of `pg_lsclusters` rather than fixed. Advertising **5433** instead of the 5432 default is what makes the assertion meaningful; with 5432 the case would pass against a hook that ignored the discovered port entirely.

Tightening `fresh` from "doesn't contain STALE" to asserting both actual ports is the more valuable half of that change — the old form passed on any non-stale marker, including one built from the wrong port. The uncovered branches that remain (`pg_ctlcluster` start, partial *native* bring-up mirroring case 2) are gaps in a best-effort dev hook, not something to hold the PR for.

### What's good

- **Dumping `out.log` on failure.** The old harness captured output and then let the `EXIT` trap wipe it, so a CI failure was undiagnosable. That it immediately caught the `redis-server` stub calling `touch` — not on the restricted `PATH` — is the argument for it.
- **`HOME` isolation reasoned from the hook's actual gates**, not applied blindly: the description confirms both `$HOME/.nvm/nvm.sh` and the `$HOME/.local/bin` symlink are live in the remote container, so the block was inert only because the restricted `PATH` happens to lack `sed`/`ln`/`mkdir`. Making that deterministic rather than incidental is the right instinct for a harness.
- **`passed=$((passed + 1))`** — post-incrementing from zero returning false is exactly the kind of latent `set -e` landmine worth removing before it fires.
- **Validating the fix against the pre-fix hook** (3 passed, 2 failed) so the original cases are confirmed to still bind to the defect rather than passing vacuously.

---

ginderjeremiah:
All three addressed in `daacb3f`; took the fuller option on the two that offered a choice.

**`psql` stub unreachable** — went with the `sudo` rewrite rather than dropping the stub. It now strips a leading `-u <user>` and execs the rest, so all three `sudo -u postgres psql` call sites genuinely reach it. Agreed on the reasoning that this doesn't reopen the "path to the genuine article" concern: `PATH` is the stub dir only, so the exec can't resolve anywhere else.

**Postgres coverage asymmetry** — sentinel-gated rather than documented-as-intentional. `pg_isready` now gates on `$WORK/pg-up` and a new `pg_ctlcluster` stub creates it, so Postgres starts down exactly like Redis and the cluster-start branch plus the `NR==1{print $1}` version parse both run.

Both stubs log their argv, and a new `assert_logged` checks the version handed to `pg_ctlcluster` and the port handed to `psql`. Worth noting your read was right that the port assertion doesn't cover these — mutation-tested it: changing the version column to `$2` and hardcoding the `psql` port to 5432 both fail the new assertions while the marker assertion still passes.

**Shared stub state** — folded into `setup_case`, which now clears every stub plus the shared sentinels/logs. Each case declares its stubs after that call, so the cases are self-contained and order-independent, and the scattered `rm -f`/`clear_native_stubs` calls are gone. You were right that the case-1 call was a no-op and case 5's only worked by position.

7 pass locally.

---

ginderjeremiah:
## Re-review of `daacb3f` — all three nits addressed, verified independently

**Verdict:** Approve. CI green (`all-checks-passed`, `test-hooks`). Incremental pass over the delta only; the first commit was covered by the [earlier review](https://github.com/ginderjeremiah/GameServer/pull/2398#pullrequestreview-4777654981).

I ran the suite and mutation-tested the new assertions rather than taking the reply's claims on trust.

### The three nits

**1. Unreachable `psql` stub → fixed the way that buys coverage.** The `sudo` rewrite is the better of the two options I offered, and the safety argument holds: `run_hook` sets `PATH="$STUB"` and nothing in the hook restores it, so `exec "$@"` after the `-u <user>` strip cannot resolve outside the stub dir. Confirmed all three `sudo -u postgres psql` call sites now reach the stub — `psql.log` carries all three statements.

**2. Postgres coverage asymmetry → sentinel-gated, and the new assertions bind.** Mutation-tested each independently against the hook:

| Mutation | Result |
|---|---|
| `NR==1{print $1}` → `{print $2}` (version column) | `FAIL: native fallback (cluster start) — never issued '18 main start' (got: main main start;)` |
| `psql -p "$NATIVE_POSTGRES_PORT"` → `psql -p 5432` | `FAIL: native fallback (test user)` |
| `write_marker "$NATIVE_POSTGRES_PORT"` → `write_marker 5432` | `FAIL: native fallback (discovered cluster port)` |

Each mutation fails exactly one assertion and no other, which is the property that matters — the three cover disjoint fields parsed from the same `pg_lsclusters` line. That confirms the reply's point that the marker assertion alone did not cover the other two.

**3. Shared stub state → centralized.** `setup_case` clearing all seven service stubs plus the sentinels and logs, with each case declaring its stubs after the call, does make the cases order-independent. `clear_native_stubs` is gone with no callers left behind.

7 passed, 0 failed locally at `daacb3f`.

### Follow-ups filed

Two, neither blocking:

- **#2409** — the coverage this delta *swapped* rather than added. Making Postgres start down means the already-running shortcut (the `else` arm that re-parses the port and applies the `5432` default) is now unreached; before this commit it was the only covered arm. Trading it for the start path was the right call — that's the fragile one — but the `[[ -z … ]] && =5432` default now has no test at all, and both arms occur for real. Also folded in a cosmetic stderr leak in `assert_logged`'s failure diagnostic (`< "$log"` is attempted before `2>/dev/null` takes effect, so a missing log prints a stray "No such file or directory" on the very branch that exists to report it).
- **#2410** — the Docker-Postgres + native-Redis hole, carried out of #2397's "explicitly out of scope" list so it survives that issue closing.

The `/etc/profile.d/nvm.sh` gap the description flags under "Not addressed" stays as filed in #2399; your reasoning that nothing on the test side closes it is correct.

### One bookkeeping note

This PR completes **all six** numbered items of #2397 (HOME isolation, native-fallback coverage, `fresh` strengthening, `ip`/`sudo` stubs, the `((passed++))` and `out.log` pair, the docs qualifier), but doesn't reference it, so the merge won't close it. I'll close it on merge pointing here. Its two out-of-scope leftovers are now re-homed: the hook gap in #2410, and the "no test asserts the hook always exits 0" note — which is genuinely cheap and worth folding into #2409's pass, since `run_hook` already discards the status.

Also worth noting for whoever picks it up: #2396 overlaps #2399 substantially now. Its stated motivation (the `echo >>` append escaping the sandbox) is exactly what #2399 fixes at the gate, and its secondary win ("the test suite can then cover the native-fallback success path") is delivered by this PR. What's left of #2396 is the single-responsibility refactor on its own merits — worth re-scoping or closing rather than working as written.

Merging.

---

PR #2400, opened 2026-07-25
Name Battler's construction options (FE) and split the effect machinery out of Battler.cs (BE)

### LINKED ISSUE #2380: Parity-surface maintainability: Battler's six positional optional params (FE) and the 810-line Battler.cs interleaving parity math with tally machinery (BE)
Two structural risks on the anti-cheat parity surface, filed together because they share a safety net (the 68 mirrored parity scenarios) and a motivation: every line here must stay bit-identical across two languages, so call-site and file clarity is worth more than usual.

1. **Frontend `Battler` construction is transposition-prone.** `UI/src/lib/battle/battler.ts:133-149,387-393` — constructor/`reset` take six positional, all-optional parameters, two sharing type `number[]` (`grantedSkillIds` and `battlerData.selectedSkills`) and two more loosely typed (`equippedWeaponType?`, `counterSkillId?`). A transposed argument at a call site like `battle-engine.ts:292-299` compiles clean and silently breaks parity. Convert to an options object so call sites self-check.
2. **`Game.Core/Battle/Battler.cs` (~810 lines) interleaves the parity-critical health/mitigation math with the effect-stack machinery** (`AttributeEffectStack`, `ApplyEffect`, `AdvanceEffects`) **and five backend-only overlay-tally readers** (Hex/Sunder/Momentum contribution scans). Extracting the effect-stack + tally machinery would leave a small, auditable parity core. Per the #1736 precedent this need not be pre-emptive — but unlike EnemyManager there's no open feature likely to force the touch soon, and the audit cost of the current interleaving is paid on every parity review, so it's worth scheduling on its own.

Constraint for both: pure refactors, no behavior change — the mirrored parity suites (`BattleSimulatorParityTests` ↔ `battle-simulation-parity.test.ts`) must pass bit-identically before and after.

ginderjeremiah:
## Summary

Both halves of #2380 — two **pure refactors** on the anti-cheat parity surface, no behavior change on either side.

### 1. Frontend: `Battler` construction is now named, not positional

`UI/src/lib/battle/battler.ts` — the constructor and `reset` took six positional, all-optional parameters. They were structurally interchangeable enough (two attribute/modifier lists, a `number[]`, a bare `number`) that a transposed argument compiled clean and silently desynced the two simulators, and call sites read like `new Battler(data, undefined, [1], undefined, EDamageType.Sword)`.

Both now take a single `BattlerInit` options object with per-field docs. `BattlerData` is exported alongside it (it was already reachable through the public signature). Every call site — the engine's player/enemy resets and the test/parity harnesses — is converted; each field is now named at the point of use.

The same conversion is applied to `grantedBattlerFactory.build` in `battle-sim-test-utils.ts` (now a named `GrantedBattlerSpec` across its 14 call sites). Its adjacent `selectedSkillIds`/`grantedSkillIds` pair carried the identical hazard, and it's worse in the harness than at a production call site: a transposed `new Battler(...)` desyncs the simulators and the parity suite catches it, whereas a transposed `build(...)` makes the FE suite assert a different loadout than its BE mirror while both stay green.

### 2. Backend: `Battler.cs` keeps only the parity core

`Game.Core/Battle/Battler.cs` was ~810 lines interleaving the parity-critical health/mitigation math with the effect-stack machinery and five backend-only overlay-tally readers. Moved to a new `Game.Core/Battle/BattlerEffects.cs`:

- the per-attribute `AttributeEffectStack` list, the shared simulated-time clock, `GetOrCreateStack`/`SwapCombinedModifier`
- `Apply`/`Advance` (the stacking, shared expiry and combined-modifier folding)
- the overlay readers — `AppliedVulnerability`, `AppliedSunder`, `AppliedMomentum`, `HexBonusForHit`, `SunderBonusForHit`

Battler.cs is now ~465 lines: construction, health, amplification/resistance/Toughness, DoT/HoT, reflection, clone.

**Keeping the MaxHealth re-clamp unskippable.** `ApplyEffect`/`AdvanceEffects` own the re-clamp, so applying an effect must not be reachable any other way. `Battler` holds the concrete `BattlerEffects` privately and exposes only a read-only `IBattlerEffectTallies` (the five tally readers) — `Apply`/`Advance` are not on the exposed type at all, and are additionally `internal`. Marking them `internal` alone would have been insufficient: `BattleContext`, the one caller that would plausibly skip the clamp, is in the same assembly.

**φ has its own home.** The shared overlay saturation `NormalizeInvestment` moved to a new `OverlayTally` rather than riding along on `BattlerEffects` — half its callers (the crit, execute and cadence investments) read no applied effect at all.

Call sites updated in `BattleContext`, plus the `<see cref>` docs in `BattleStats` and the now-stale names in `docs/backend-battle.md`.

## Addresses the issue

Both listed items, with the issue's own constraint held: no behavior change, and the mirrored parity suites agree bit-identically before and after.

## Test plan

No new tests — nothing here is newly expressible; the existing mirrored suites *are* the safety net for a pure move on this surface, and both were run in full.

- [x] `dotnet test` (full backend, re-run after the review fixes): Game.Core 1442 ✅ (incl. `BattleSimulatorParityTests`, `BattlerEffectsTests`, `BattleContextTests`), Game.Application 1103 ✅, Game.Api 871 ✅, Game.Infrastructure 74 ✅
  - One flake on the first full run — `CoalescingReferenceCacheReloaderTests.RunAsync_NoSignalWithinReconciliationInterval_RunsAPeriodicSweepAnyway`, a 10s timing-based reloader test unrelated to battle code. Did not recur on either subsequent full run.
- [x] `npm run lint` — 0 errors, 0 warnings
- [x] `npm run test:unit:ci` — 4000/4000 passed (incl. `battle-simulation-parity.test.ts`)
- [x] CI green on `ab66f9d`: `verify-backend`, `verify-frontend`, `test-backend`, `test-frontend`, all three e2e browsers

## Review

Reviewed in [#2400 (review)](https://github.com/ginderjeremiah/GameServer/pull/2400#pullrequestreview-4777704934) — approved with three should-fixes and two nits, all addressed in `ab66f9d` (the read-only `Effects` view, `OverlayTally`, the harness conversion, the `BattlerInit` doc trim, and two dead null guards dropped).

One deliberate non-change, called out in the [reply](https://github.com/ginderjeremiah/GameServer/pull/2400#issuecomment-5075882197): `battlerFactory` stays positional — its three params have no adjacent same-typed pair and the two attribute-ish arrays have incompatible element shapes (`id` vs `attributeId`), so a transposition there is a compile error rather than a silent one.

Closes #2380

https://claude.ai/code/session_015HB7YusHJe7sRLaWaUY7gM


Game.Core/Battle/Battler.cs
```diff
@@ -13,25 +13,13 @@ public class Battler
...
-        /// This battler's elapsed simulated time in ms, advanced one tick at a time by
-        /// <see cref="AdvanceEffects"/>. Active effects store an absolute expiry against this clock, so expiry
-        /// is a comparison rather than a per-tick countdown and <see cref="ActiveEffect"/> stays immutable.
+        /// This battler's timed skill-effect bookkeeping — the per-attribute effect stacks and the backend-only
+        /// overlay tallies reading them (#2380). Applying and expiring effects goes through
+        /// <see cref="ApplyEffect"/>/<see cref="AdvanceEffects"/> so the MaxHealth re-clamp can't be skipped;
+        /// this is exposed for the overlay readers, which never mutate state.
         /// </summary>
-        private long _elapsedMs;
+        public BattlerEffects Effects { get; }
```
ginderjeremiah: **Should-fix — the `internal` guard doesn't bind where the risk actually is.**

The doc says the re-clamp "can't be skipped," and `BattlerEffects.cs:18-19` says `Apply`/`Advance` are `internal` "so the only way in from outside the domain is `Battler.ApplyEffect`/`AdvanceEffects`." Both statements are literally true, but the boundary they defend is the *assembly* boundary — and the only real caller of this machinery, `BattleContext`, sits in the same assembly, in the same folder. `_targetBattler.Effects.Apply(effect, amount)` compiles fine from `BattleContext` and silently skips `ClampHealthToMaxHealth`.

So the invariant is protected against callers who were never going to violate it (`Game.Core.Tests` has no `InternalsVisibleTo`, so tests already go through `Battler.ApplyEffect`) and unprotected against the one that plausibly would. Before this PR the machinery was private to `Battler` and structurally unreachable; the exposed sub-object is a net loosening dressed as a tightening.

Cheapest fix that keeps the encapsulation the old shape had: keep the concrete instance in a private field and have the property return a read-only interface with just the five tally readers —

```csharp
private readonly BattlerEffects _effects;
public IBattlerEffectTallies Effects => _effects;
```

`Apply`/`Advance` then aren't on the exposed type at all and `internal` becomes belt-and-braces rather than the whole guard. If you'd rather not add an interface, renaming to `ApplyUnclamped`/`AdvanceUnclamped` at least makes the missing step visible at any accidental call site. Either way the class doc should stop claiming the stronger property than the code provides.

---

Game.Core/Battle/BattlerEffects.cs
```diff
@@ -0,0 +1,406 @@
...
+        /// <summary>
+        /// The shared overlay-tally saturation <c>φ(a) = a / (1 + a)</c>, applied to an overlay's own investment
+        /// magnitude when booking its share claim (#1481): ~linear at the low end so a token investment trains
+        /// proportionally little, asymptoting to <c>1</c> so even a huge investment claims at most the full booked
+        /// hit. Every overlay signal (crit, Hex, Momentum, Sunder, Cull, Cadence) uses it on its own investment.
+        /// </summary>
+        public static double NormalizeInvestment(double investment)
+        {
+            return investment / (1.0 + investment);
+        }
```
ginderjeremiah: **Should-fix — φ landed in a second wrong home rather than its right one.**

`NormalizeInvestment` is a pure static with no relationship to effect stacks: three of its six callers (`BattleContext.cs:378` Cull, `:385` Cadence, `:402` crit) involve no applied effect at all. `BattlerEffects.NormalizeInvestment(critMultiplier - 1.0)` reads as though crit normalization is somehow this battler's effect bookkeeping.

It was equally misplaced on `Battler` before, so this isn't a regression — but the PR's whole premise is that these two files should each be auditable on their own terms, and it's moving the method anyway. A ~15-line `Game.Core/Battle/OverlayTally.cs` holding the shared φ (and its `#1481` doc, which is entirely about overlay tallies and not at all about effect stacks) makes every call site read correctly, including the three that have nothing to do with this class.

Happy for this to be a follow-up if you'd rather keep the diff a strict move — but it's cheaper to do while the method is already in flight than to move it a third time later.

---

UI/src/tests/lib/battle/battle-sim-test-utils.ts
```diff
@@ -165,19 +165,17 @@ export function grantedBattlerFactory(registry: ISkill[]) {
 			equippedWeaponType?: EDamageType,
 			counterSkillId?: number
 		): Battler =>
-			new Battler(
-				{
+			new Battler({
```
ginderjeremiah, line 179: **Should-fix — yes, fold `grantedBattlerFactory.build` in (answering the PR's "Note for review").**

`build` (lines 161-166 above) still takes `selectedSkillIds: number[]` immediately followed by `grantedSkillIds: number[]` positionally. That is *precisely* the pair #2380 names as the transposition hazard — two adjacent same-typed `number[]` args — and it now sits one frame above a call site you just made transposition-proof. The 14 call sites carry the full risk with none of the compile-time protection.

It's worse here than it was in the constructor, in one specific way: a transposed `new Battler(...)` desynced the simulators, which the mirrored parity suite would eventually catch. A transposed `build(...)` makes the FE parity test *assert against a different loadout than the BE test does* while both still pass — the harness is exactly where a silent wrong-but-green is least detectable, because the harness is the detector.

It's a mechanical `{ attrs, selectedSkillIds, grantedSkillIds, equippedWeaponType, counterSkillId }` conversion across 14 sites, in this file's own test-support code — well inside the spirit of the issue even if outside its literal line references. Same for `battlerFactory` at line 132 if its signature has the same shape.

---

UI/src/lib/battle/battler.ts
```diff
@@ -8,13 +8,38 @@ import { EAttributeModifierSource, type AttributeModifier } from './attribute-mo
...
 	level: number;
 	name: string;
 }
 
+/** Everything a {@link Battler} is assembled from, as a named options object rather than a positional
+ *  argument list (#2380). Every field is optional and an omitted one behaves exactly as the absent
+ *  positional argument did — a data-less `{}` is the idle re-arm (see {@link Battler.reset}). This is a
+ *  parity surface, which is why it is named rather than positional (#2380): the six were structurally
+ *  interchangeable enough (two attribute/modifier lists, a `number[]`, a bare `number`) that a transposed
+ *  argument compiled clean and silently desynced the two simulators. */
```
ginderjeremiah: **Nit — the doc states its rationale twice and anchors it to the shape it replaced.**

"as a named options object rather than a positional argument list (#2380)" and "which is why it is named rather than positional (#2380)" are the same sentence with the same issue reference, six lines apart. Keep the second (it carries the *why* — the interchangeable types) and drop the first clause.

"an omitted one behaves exactly as the absent positional argument did" also documents the superseded behavior, which `CLAUDE.md` calls out directly: *"remove the information about the old behavior and document the new behavior ONLY if it is truly relevant."* Six months from now nobody has the positional version to compare against, so the sentence only says "every field is optional" the long way. Same for the "the six were..." count — it's a description of what used to be there.

Something like:

```
/** Everything a {@link Battler} is assembled from. Every field is optional; a data-less `{}` is the
 *  idle re-arm (see {@link Battler.reset}). Named rather than positional because this is a parity
 *  surface (#2380) — the fields are structurally interchangeable enough (two attribute/modifier
 *  lists, a `number[]`, a bare `number`) that a transposed argument would compile clean and silently
 *  desync the two simulators. */
```

Purely presentational — no objection to the content, just the duplication.

---

Game.Core/Battle/BattlerEffects.cs
```diff
@@ -0,0 +1,406 @@
...
+        /// than diffed against a baseline, so it credits the debuff's own work even when the target's base
+        /// resistance is high or the target buffs its own resistance (a self-buff never lowers this). Rides the
+        /// shared per-attribute stack expiry, so it returns to <c>0</c> for free when the debuff lapses.
+        /// </summary>
+        public double AppliedVulnerability(EDamageType damageType)
+        {
+            if (_attributeStacks is null)
+            {
+                return 0;
+            }
```
ginderjeremiah: **Nit (pre-existing, carried over verbatim) — this null guard is dead.**

`StackContribution` opens with the identical `_attributeStacks is null → 0` check (line 309), and with a null list the loop body never runs, so `contribution` stays `0.0` and the `contribution < 0` test returns `0` anyway. Same four lines in `AppliedMomentum` at 288-291. Both are pure no-ops.

Not introduced here — it moved as-is, and I confirmed it against `d74ead5` (old `Battler.cs:270`, `:337`). Flagging it because the stated goal is a file that's cheap to audit, and a reader hitting a redundant guard has to go check whether the callee really does handle null or whether one of the two paths is load-bearing. Deleting both is a strictly-mechanical two-hunk change while the code is already being moved; `CLAUDE.md` treats pre-existing dead code as fair game to clean up rather than replicate.

---

ginderjeremiah (commented):
## Review: both halves of #2380 — named `Battler` construction (FE) and the effect machinery split out of `Battler.cs` (BE)

**Verdict:** Approve (posted as a comment — GitHub won't take a gating verdict from the PR author) — no blocking findings. Three should-fixes and two nits below; **not merging**, since the should-fixes belong in this PR rather than in follow-ups, and `test-backend` is still running.

I verified the "pure refactor, no behavior change" claim rather than taking it on trust:

- The BE move is faithful. `StackContribution`, the `Func<…>` selector shape, and the redundant null guards all existed at `d74ead5` — nothing was quietly rewritten under cover of the move. No stale `Battler.NormalizeInvestment`/`Battler.Applied*` references survive anywhere in `.cs` or `.md`.
- `CloneWithAttributeDelta` is byte-identical and still builds a fresh `Battler`, so the new `BattlerEffects` starts with `_attributeStacks = null` / `_elapsedMs = 0` — exactly the clone's prior state. This was the one place a composed sub-object could have silently changed semantics, and it didn't.
- `Game.Core` has no `InternalsVisibleTo`, so making `Apply`/`Advance` `internal` costs the test suite nothing; the effect tests already drive `Battler.ApplyEffect`.
- FE conversion is complete — all four `reset`/`new Battler` sites in `battle-engine.ts` (including the bare `reset()` idle re-arm at line 284, which the `= {}` default covers) plus every test/harness site.

### Should-fix

- **The `internal` guard protects the assembly boundary, not the caller that matters** — `Game.Core/Battle/Battler.cs:16-22`. `BattleContext` is in the same assembly, so `Effects.Apply(...)` compiles there and skips the MaxHealth re-clamp. The old private-field shape was structurally unreachable; this is a loosening described as a tightening. A read-only interface on the property restores it.
- **`NormalizeInvestment` moved to a second wrong home** — `BattlerEffects.cs:50-59`. Three of its six callers involve no applied effect at all. Cheaper to place correctly now than to move a third time.
- **`grantedBattlerFactory.build` keeps the exact hazard the PR just removed** — `battle-sim-test-utils.ts:161-168`. Answering your note: yes, fold it in. Adjacent same-typed `number[]` params in the *parity harness* is the one place a transposition yields wrong-but-green instead of a caught desync.

### Nits

- **`BattlerInit` doc repeats its rationale twice** (with `#2380` cited twice) and anchors itself to the positional shape it replaced — `battler.ts:18-23`. `CLAUDE.md` asks for the new behavior only.
- **Dead null guards carried over verbatim** — `BattlerEffects.cs:219-224` and `288-291`. Pre-existing, mechanical to drop while the code is in flight.

### Test coverage

Appropriate as-is, and I agree with the reasoning that no new tests are warranted: nothing here is newly expressible, and the mirrored suites are the safety net. `BattleSimulatorParityTests` ↔ `battle-simulation-parity.test.ts` passing unchanged on both sides is the correct — and only real — verification for a pure move on this surface. CI backs it: `verify-backend`, `verify-frontend`, `test-frontend`, and all three e2e browsers are green.

Worth noting the FE test conversion is a genuine coverage improvement beyond the mechanical rewrite: `makeBattlerData` went from `Partial<Parameters<Battler['reset']>[0] & {}>` to `Partial<BattlerData>`. Exporting the interface let a type-level workaround (which would have silently gone `unknown`-shaped if the signature moved) become an ordinary named type.

### What's good

- **The options-object shape is self-enforcing during migration, not just after it.** `BattlerInit` is all-optional, which makes it a *weak type* — so a leftover `reset(playerManager)` passing a bare `BattlerData` is a compile error ("no properties in common"), not a silent degradation into the data-less idle re-arm. That's the failure mode I went looking for first, given `reset()` legitimately means "re-arm," and the type system closes it for free. The alternative shape (a required `battlerData` field) wouldn't have been better here — this one is right.
- **Per-field docs on `BattlerInit` earn their keep.** `additionalModifiers` explaining *why* it composes before the static engine modifiers (#1189 ordering) is the kind of thing that was previously only inferable from `battle-engine.ts:291-300`'s comment block. On a surface where ordering is parity-critical, documenting it at the type rather than at one call site is the right move.
- **The split is drawn along the right seam** — parity-critical math on one side, bounded-cost bookkeeping plus the backend-only tallies (which have no FE mirror at all) on the other. 810 → 465 lines, and the removed half is genuinely the half a parity reviewer can skip. That's the audit cost the issue was actually complaining about.
- **The `Apply`/`Advance` → caller-re-clamps contract is stated explicitly** rather than left implicit in call order. Even with the visibility caveat above, having "these report back, the battler owns the clamp" written down is what makes the invariant reviewable at all.

---

ginderjeremiah:
All five addressed in `ab66f9d`. Agreed on every one — the first in particular: I'd talked myself into `internal` being the guard, and it isn't. `BattleContext` is the caller that would plausibly skip the clamp and it's in the same assembly, so the exposed sub-object was strictly weaker than the private field it replaced.

**Should-fixes**

1. **Read-only view.** `Battler` keeps the concrete instance in `_effects` and `Effects` returns a new `IBattlerEffectTallies` — the five tally readers, nothing else. `Apply`/`Advance` aren't on the exposed type at all now, so `internal` is belt-and-braces as you put it, and the class docs on both sides say the weaker (true) thing.
2. **φ → `OverlayTally`.** New `Game.Core/Battle/OverlayTally.cs`, with the `#1481` share-claim rationale moved onto the class doc where it's about overlays rather than stacks. `BattleContext`'s crit/execute/cadence sites now read `OverlayTally.NormalizeInvestment(...)`; the `<see cref>`s in `BattleStats` and the one line in `docs/backend-battle.md` follow.
3. **`grantedBattlerFactory.build`.** Converted to a named `GrantedBattlerSpec` across all 14 call sites, with the per-field docs moved from the inline comment block onto the interface. Your point about *why* it's worse here than in the constructor is the part I'd underweighted — the harness being the detector means a transposition there is undetectable by construction, not just eventually-detectable.

   I left `battlerFactory` (line ~124) positional: its three params are `{ id, amount }[]`, `SkillSpec[]`, `IBattlerAttribute[]` — no adjacent same-typed pair, and the two attribute-ish arrays have incompatible element shapes (`id` vs `attributeId`), so a transposition is a compile error rather than a silent one. Happy to convert it for consistency if you'd rather it match.

**Nits**

4. `BattlerInit` doc rewritten to your suggested shape — one rationale, no description of the positional version.
5. Both dead `_attributeStacks is null` guards dropped.

**Verification after the changes:** `dotnet build` 0 warnings; full backend green — Core 1442, Application 1103, Api 871, Infrastructure 74 (that `CoalescingReferenceCacheReloaderTests` flake did not recur). Frontend `npm run lint` clean, `test:unit:ci` 4000/4000 — including both parity suites, whose expectations are untouched by the harness conversion.


---

Issue #1457, opened 2026-07-02
Delivery archetype: Parry / Riposte — opt-in parry chance with a counterattack


ginderjeremiah:
From the class-route planning in [content-design §5](../blob/main/docs/content-design.md): the **Swordsman's kit secondary is a parry stance**, and rather than piggybacking `DodgeChance`, parry is its **own mechanic and delivery archetype** — an opt-in chance to avoid an incoming hit **and counterattack**.

## Model (opt-in-multiplicative template; crit #1425/#1448 as the worked reference)

- New **`ParryChance`** attribute: decimal probability, **base 0, no attribute derivation** — flat chance granted only by skills/items (the Swordsman's stance grants it for a timed window).
- A **Riposte** Technique proficiency path multiplies it (`0 × mult = 0`, inert until opted in).
- Tally: the **counter damage dealt** — a *direct* tally like Retribution's, not the overlay marginal (the counter is new damage, not a rider on a hit that would have landed anyway).

## Niche discipline (why it doesn't tread on dodge or reflection)

- **Dodge** — passive avoidance variance, AGI identity. Avoids; nothing else.
- **Reflection** — deterministic thorns on hits you **take**, scaling with the *incoming* hit and bypassing the attacker's mitigation — the tank feasts on hard-hitters.
- **Parry** — a proc that avoids **and** returns damage scaling with the **defender's own offense** — turn defense into *your* attack. The duelist corner; strongest for builds that are already dangerous, useless for a pure wall (which is Reflection's job).

## Design questions

- **Full negation + counter, or partial reduction + counter?** Full negation is strictly better than a dodge per proc — acceptable if parry stays opt-in/windowed while dodge is the always-on layer, but decide deliberately (partial keeps dodge's full-avoid unique).
- **Counter formula & typing** — flat + attribute-scaled? A fraction of the parried hit (careful: that drifts toward Reflection)? And which damage type does the counter carry — the equipped weapon's type would make a sword counter train Swordsmanship, feeding the Swordsman's roots; a typeless/Physical counter is simpler.
- **RNG & parity** — a new draw in hit resolution on both sims; seeded-stream discipline (the draw must be consumed identically whether or not parry can proc); priority/interaction with dodge (ordered checks vs one exclusive roll); are crits parryable; DoT is presumably unparryable (no coherent attacker — mirror dodge/reflection).
- **Player-only asymmetry** like crit/dodge, with enemy parry deferred as a boss-design seam?
[...]

PR #2401, opened 2026-07-25
Ideation spike: deferred defensive mechanics (#1331)

ginderjeremiah:
Ideation spike for #1331 — mitigation→sustain, absorb shield, last-stand. **Research and documentation only**: no design is committed, no engine code is touched, and no implementation issues are split out yet (per CLAUDE.md's rule that an ideation spike gives its collaborator a chance to weigh in before planning). Q1–Q5 in §7 are the asks.

> **Updated after review** (8078bfa). Two claims in the original description were corrected by the review and have been fixed below: last-stand's parity surface is wider than "one multiplier in `ComputeNetDamage`", and its bound is an authoring invariant rather than a structural one. Details in the [reply](https://github.com/ginderjeremiah/GameServer/pull/2401#issuecomment-5076203079).

## Summary

Adds `docs/spikes/1331-defensive-mechanics.md` and updates #1330's forward-reference to point at it.

The headline is that **the engine moved underneath #1331's original framing**, so all three candidates re-rank. #1331 (and #1330's deferral note) recorded the lean as *"shield/last-stand lean toward the skill-effect system; mitigation→sustain toward an authored attribute"*. Cull (#1430) and Parry (#1457) have shipped since, and they invert the cost ordering.

## Findings

**1. Last-stand is the cheapest of the three, not the most speculative.** `ExecuteBonus` already samples the target's live missing-health fraction once per fire and folds it into that fire's damage multiplier (`BattleContext.cs:316-320`). Last-stand is the **defender-side mirror of that exact read** — an authored-only attribute, one multiplier, no new state.

Two caveats the review sharpened, both now in §3.1:

- **Placement is a real call, not a detail.** `ComputeNetDamage` is a pure function of attributes with **four** consumers, and only `Battler.cs:225` (`TakeDamage`) is the damage path — the others are the `DamageDodged`/`DamageParried` accrual tallies (`BattleContext.cs:247,262`) and `CombatRating.cs:239`'s synthetic reference defender, whose `CurrentHealth` is meaningless. A multiplier dropped in there would silently change two proficiency signals and pollute the combat rating. Cull put its own read in the **fire path** for exactly this reason, so "mirror Cull exactly" argues for the fire path here too.
- **The bound is an authoring invariant, not structural.** *Provided* `LastStandBonus < 1` the multiplier stays in `(0, 1]` and can never invert a hit's sign — but nothing enforces that, and authored attributes are read unclamped throughout (`DamageReflection` is multiplied raw at `BattleContext.cs:495-501`). Plausibly fine under the #1478 "unreachable by authored content" precedent, but it's that class of assumption, and it's the whole argument for shipping last-stand first.

Worth flagging separately: #1219 states *"nothing reads live state like a current-health ratio mid-tick."* That was true when written and **is no longer accurate** — its cost estimate for conditional class passives should be revised down. §8 records amending it when the implementation issues are split.

**2. Mitigation→sustain has an immortality breakpoint that neither issue anticipated.** Healing a share `S` of absorbed damage gives `Δhealth = dealt × [m(1+S) − S]` where `m = (1−r)(1−t)`, so net damage hits zero at `m ≤ S/(1+S)`:

| Toughness | `t` | `r` | `m` | `S` for immortality |
|---|---|---|---|---|
| 200 (`= C`) | 50% | 0 | 0.50 | 100% |
| 600 | 75% | 0 | 0.25 | **33%** |
| 800 | 80% | 0.30 | 0.14 | **16%** |

At ordinary investment (not degenerate stacking) a ~16% sustain rate is strictly immune to direct hits, and the immunity *deepens* with more mitigation. **The MaxHealth heal cap does not defuse this** — it removes overheal only, pinning the equilibrium just below full health rather than bounding the pathology.

And because `DamageReflection` keeps working while immortal, the result isn't the timeout draw a pure wall gets today — it's a **guaranteed win** against any non-DoT enemy. That step is non-obvious and now spelled out in §4: `ReflectDamage` early-returns on `netDamage <= 0` (`BattleContext.cs:488-493`), but sustain never drives `net` non-positive — the heal is a *separate channel* that outruns it — so reflection fires every hit for the whole battle. This is the same failure class #1330 spent a spike removing from flat Defense, relocated further along the curve.

The doc evaluates four ways out; only **capping the heal at a fraction of the damage that actually landed** both preserves the mitigation-coupled intent and removes (rather than relocates) the breakpoint.

**3. The absorb shield is the most expensive, and collides with a load-bearing invariant.** *"The game has no overheal/shield concept"* is asserted at **7 code sites — 4 backend (`Battler.cs:219,244,298,398`) and 3 frontend (`battler.ts:199,215,245`), so it's parity-pinned on both simulators** — plus **3 doc sites**, and it's the stated *reason* three heal channels cap at MaxHealth. A pool also needs six parity-pinned decisions (does it absorb DoT? does it feed `HealthRemoved`'s #1482 offense-book cap? does it count in Cull's missing-HP fraction? …) plus new state on both simulators.

The doc also notes the commitment rule names **raw EHP** as universal free value earning no path (`content-design.md:178`) — so a shield ships path-less unless the rule gets an explicit carve-out. Sustain and last-stand both qualify as authored-only enablers and don't have this problem.

## Directions beyond the issue's three

Per the ideation brief, exploring past what was listed — most usefully: **part of the shield is authorable today with zero engine work.** A timed effect granting resistance > 1 already produces reactive absorption via `ComputeNetDamage`'s negative branch, and a timed MaxHealth buff + HoT approximates a buffer. It isn't a finite pool, but it's worth authoring and playing before committing to one. Also covers how the three compound (they reach the §4 regime faster together than alone, arguing for sequencing over a bundled release), Second Wind as a one-shot revive as a genuinely distinct mechanic from the ramp, and confirms the anti-burst damage cap stays rejected per #1330.

## Recommendation (if a read is wanted)

Last-stand first (smallest, and bounded under a stated authoring invariant) → sustain second, only with the cap → defer the shield behind the content experiment.

## Test plan

Docs-only; no code paths touched, so there is nothing to unit/integration test and CI's changed-path gate skips the build/test jobs.

- [x] The breakpoint math in §4 verified numerically — the three table rows solve to `S` = 1.0000 / 0.3333 / 0.1628 and each yields `Δhealth = 0.000000` on a simulated hit; the proposed cap yields a strictly positive `Δhealth` past the breakpoint.
- [x] Every code reference cited (`Battler.cs:158-171`, `:219/244/298/398`, `BattleContext.cs:316-320`, `:488-493`, `:495-501`, `ProficiencyAccrual.cs:182`, `CombatRating.cs:239`, `battler.ts:199/215/245`, `content-design.md:178`) read and confirmed against the current tree, not quoted from the issues.
- [x] `ToughnessMitigationConstant = 200` and the `2 × Endurance` derivation confirmed from source, so the table's Toughness values are real reachable investments.
- [x] All four `ComputeNetDamage` consumers enumerated from the tree, which is what corrected the parity-surface claim above.

Refs #1331 — deliberately **not** `Closes`, since the spike isn't complete until the open questions are answered and implementation issues are created.

https://claude.ai/code/session_016RhokX8UV9VLA3HLTQTECn


docs/spikes/1331-defensive-mechanics.md
```diff
@@ -0,0 +1,147 @@
...
+
+```
+missingHpFraction = clamp((MaxHealth − CurrentHealth) / MaxHealth, 0, 1)
+net = … × (1 − LastStandBonus × missingHpFraction)
+```
+
+- **Continuous, not a threshold step.** The issue says *"below an HP threshold"*, but a step function reintroduces exactly the discontinuous breakpoint pathology #1330 spent a spike removing from flat Defense. Cull already chose the continuous ramp for the offensive mirror; matching it keeps one shape in the codebase and no cliff. **Recommend continuous.**
+- **Bounded by construction.** With `LastStandBonus < 1` the multiplier stays in `(0, 1]` — it can reduce a hit but never heal, so it cannot invert the sign and cannot reach immunity. This is the property §4 shows sustain lacks.
+- **Where it sits in the pipeline is a real call** — folding it in alongside the Toughness curve (so it multiplies) vs. adding to `Toughness` itself (so it rides the diminishing curve) give very different endgame behavior. Multiplying is simpler and legible; adding is self-limiting. Open question **Q3**.
```
ginderjeremiah: **Should-fix — the placement question is bigger than "multiply vs. add into Toughness", and the §3 table's "one multiplier in `ComputeNetDamage`" understates the parity surface.**

`Battler.ComputeNetDamage` is currently a **pure function of attributes** — it reads `_attributes` and nothing else. It has four consumers, and only one of them is the damage path:

- `Battler.cs:225` — `TakeDamage`, the real hit. ✅ wants last-stand.
- `BattleContext.cs:262` — `Stats.DamageDodged += Math.Max(0, _targetBattler.ComputeNetDamage(...))`, which feeds Evasion's proficiency accrual (`ProficiencyAccrual.cs:182`).
- `BattleContext.cs:247` — the same shape for `DamageParried`.
- `CombatRating.cs:239` — `referenceDefense.ComputeNetDamage(amped, portion.Type)`. This one is the problem: the reference defender is a synthetic yardstick, so its `CurrentHealth` isn't a meaningful quantity, and the offense rating would silently pick up a term keyed to it.

The dodge/parry cases are at least arguable (the tally is a counterfactual "what would this hit have done", so a health-scaled answer may be *correct*) — but it's a silent change to two accrual signals, not a no-op, and it should be a decision rather than a side effect.

Worth noting the precedent this section leans on cuts the other way: **Cull put its live read in `BattleContext.DamageTarget`, not in `ComputeNetDamage`** (`BattleContext.cs:316-320`), which is exactly why it didn't disturb `CombatRating` or the tallies. "Mirror Cull exactly" argues for the fire path here too.

Suggest reframing **Q3** as three options — fire-path multiplier (Cull's placement), `ComputeNetDamage` multiplier (and then the three non-`TakeDamage` consumers each need a call), or additive into `Toughness` — and correcting the table's parity-surface cell. It doesn't change the ranking (last-stand is still the cheapest), but "one multiplier" is what makes it read as *trivially* cheap.

---

docs/spikes/1331-defensive-mechanics.md
```diff
@@ -0,0 +1,147 @@
...
+
+Mirror Cull exactly. An authored-only `LastStandBonus` (base `0`, no `StaticAttributeModifiers` derivation, like `DamageReflection`/`ExecuteBonus`) scales the defender's *own* missing-health fraction into a mitigation multiplier applied inside `ComputeNetDamage`:
+
+```
+missingHpFraction = clamp((MaxHealth − CurrentHealth) / MaxHealth, 0, 1)
+net = … × (1 − LastStandBonus × missingHpFraction)
+```
+
+- **Continuous, not a threshold step.** The issue says *"below an HP threshold"*, but a step function reintroduces exactly the discontinuous breakpoint pathology #1330 spent a spike removing from flat Defense. Cull already chose the continuous ramp for the offensive mirror; matching it keeps one shape in the codebase and no cliff. **Recommend continuous.**
+- **Bounded by construction.** With `LastStandBonus < 1` the multiplier stays in `(0, 1]` — it can reduce a hit but never heal, so it cannot invert the sign and cannot reach immunity. This is the property §4 shows sustain lacks.
```
ginderjeremiah: **Should-fix — "bounded by construction" overstates it; the bound is an authoring constraint, not a structural one.**

The premise `LastStandBonus < 1` is stated but nothing enforces it. Authored attributes in this codebase are unclamped at the point of use — `DamageReflection` is read and multiplied raw (`BattleContext.cs:495-501`), and `ComputeNetDamage`'s own doc comment asserts *"the only path to a negative (absorbing) result is a resistance above `1`, and no clamp is needed there"* (`Battler.cs:118-122`). An authored `LastStandBonus ≥ 1` on a near-dead defender drives that multiplier to ≤ 0 and turns the hit into a heal — precisely the sign inversion this bullet says is impossible, and precisely the flat-Defense pathology #1330 removed.

That's not necessarily a design flaw: #1478 already establishes the house precedent of leaving the negative-Toughness pole unguarded as *"unreachable by authored content"*, and the same reasoning would cover this. But it should be stated that way rather than as a construction guarantee, because this safety property is the entire argument for shipping last-stand first — §7's recommendation rests on it. Either name the authoring invariant explicitly (and note it's the #1478 class of assumption), or specify the clamp.

Small related note: the interval is `(0, 1]` only because `missingHpFraction = 1` implies the defender is already dead. Worth a half-sentence, since a reader checking the algebra at `bonus = 1` lands on `0` and wonders.

---

docs/spikes/1331-defensive-mechanics.md
```diff
@@ -0,0 +1,147 @@
...
+
+| Toughness | `t` | `r` | `m` | `S` for immortality |
+|---|---|---|---|---|
+| 200 (`= C`) | 50% | 0 | 0.50 | 100% |
+| 600 (`= 3C`) | 75% | 0 | 0.25 | **33%** |
+| 800 (`= 4C`) | 80% | 0.30 | 0.14 | **16%** |
+
+At Toughness 800 with 30% type resistance, a **16% sustain rate makes the battler strictly immune to direct hits** — and the immunity *deepens* as mitigation grows, because more mitigation means more absorbed damage means more healing. Worse, `DamageReflection` keeps working while immortal, so the outcome is not the 2-minute timeout draw a pure wall gets today — it is a **guaranteed win against any non-DoT enemy**.
+
+This is precisely the failure class #1330 existed to eliminate: flat Defense's degenerate immunity breakpoint at the `0` floor. Naively coupling healing to mitigation puts a breakpoint back, just further along the curve. DoT is the only residual pressure valve (it bypasses the Toughness curve — `backend-battle.md:34`), and it is content-dependent: against an enemy that authors no DoT, the immunity is absolute.
```
ginderjeremiah: **Should-fix — say explicitly that the MaxHealth heal cap doesn't defuse this.**

§3.2 (line 53) presents the sustain heal as *"capped at MaxHealth like every other heal channel"*, which reads as a safeguard. A reader arrives at §4 with that in hand and will reasonably ask why the cap doesn't bound the pathology. It doesn't, and it's worth one sentence saying so: the cap only removes overheal, so past the breakpoint the battler settles into a steady state just below MaxHealth — each hit lands `net`, the sustain heal refills the room it just made, and `Δhealth` clamps to 0 at full health instead of going negative. Never dying, forever. The `S ≤ m/(1−m)` analysis is unchanged; the cap just pins the equilibrium at the top of the bar rather than letting health run away.

**Nit, same line:** the DoT/Toughness-bypass citation is `backend-battle.md:36`, not `:34` (line 34 introduces the end-of-tick DoT/HoT phase; the *"DoT keeps bypassing the Toughness curve"* statement is in the typed-DoT paragraph at 36). §3.3 cites `:36` correctly, so it's just this one.

---

docs/spikes/1331-defensive-mechanics.md
```diff
@@ -0,0 +1,147 @@
...
+
+## 2. What changed since #1331 was filed (2026-06-27)
+
+Three shipped mechanics materially re-rank these candidates:
+
+- **`ExecuteBonus` / Cull ([#1430](./1398-utility-in-proficiency-system.md)) reads live battle state mid-fire.** The attacker samples the target's missing-health fraction `(MaxHealth − CurrentHealth) / MaxHealth`, clamped `[0,1]`, **once per fire**, and folds it into that fire's damage multiplier (`BattleContext.DamageTarget` ↔ `battle-step`'s `dealPortionedDamage`). This is the single most important finding here: **last-stand is the defender-side mirror of a read the engine already performs**, not a net-new mechanism. [#1219](https://github.com/ginderjeremiah/GameServer/issues/1219)'s statement that *"nothing reads live state like a current-health ratio mid-tick"* was true when written and is **no longer accurate** — that issue's cost estimate for conditional passives should be revised down accordingly.
```
ginderjeremiah: **Nit** — `[#1430](./1398-utility-in-proficiency-system.md)` links issue #1430 to the #1398 spike doc. Every other issue reference in the doc either matches its target's number (line 5's `[#1330](./1330-mitigation-rework.md)`, `[#1320]`, `[#180]`) or links to the GitHub issue URL (`[#1457]`, `[#1454]` on the next two bullets). Following this one lands on a document about something else with no explanation.

If the intent is "Cull came out of the #1398 spike", say that — e.g. `Cull ([#1430](https://github.com/ginderjeremiah/GameServer/issues/1430), from spike [#1398](./1398-utility-in-proficiency-system.md))`.

---

docs/spikes/1331-defensive-mechanics.md
```diff
@@ -0,0 +1,147 @@
...
+- Commitment rule: ✅ authored-only enabler, `0` when uncommitted. Trains on a prevented-damage tally (Evasion's precedent).
+
+### 3.2 Mitigation → sustain — right shape, one hard blocker
+
+Follows reflection's template as the issue said: an authored-only `MitigationSustain` (base `0`), healing a share of what mitigation absorbed, capped at MaxHealth like every other heal channel. The engine substrate is present (`TypeResistanceMitigated`, `CapHealToRoom`, the `ApplyHealOverTime` cap). **But see §4 — as literally specified it has an immortality breakpoint.**
+
+### 3.3 Absorb shield — collides with a load-bearing invariant
+
+*"The game has no overheal/shield concept"* is asserted in **4 code sites** (`Battler.cs:219,244,298,398`, mirrored in `battler.ts:199,215,245`) and **3 doc sites** (`game-design.md:132`, `backend-battle.md:36,73`), and it is the stated *reason* three separate heal channels cap at MaxHealth. A shield is not an increment on that design — it reverses it, and every cap that cites it needs re-deciding. §5.3 enumerates the collision surface.
```
ginderjeremiah: **Nit** — "**4 code sites**" then lists seven (`Battler.cs:219,244,298,398` + `battler.ts:199,215,245`); §8 says "all seven code comments" and the PR description says 7. All seven verified present, so it's just the count here. "**7 code sites** (4 backend / 3 frontend)" would match §8 — and the backend/frontend split is the more useful framing anyway, since it's what makes the point that the invariant is parity-pinned on both simulators.

---

ginderjeremiah (commented):
## Review: ideation spike for #1331 — deferred defensive mechanics (sustain / shield / last-stand)

**Verdict:** Approve (posted as `COMMENT` — GitHub won't take a gating verdict from the PR author) — no blocking findings. Three should-fixes and three nits inline. Docs-only, and per the repo's rule this is a **spike result, so it should not be merged on a review pass** — it also self-declares as awaiting owner input on Q1–Q5 before implementation issues are split out, which is the right call under `CLAUDE.md`'s "give your collaborator a chance to offer their input after your research" rule.

I re-derived the math and re-checked every code citation against the tree rather than trusting the description's checklist. Details below.

### Verification

**The §4 breakpoint math is correct.** `Δhealth = dealt × [m(1+S) − S]` follows from the definitions, zero at `m ≤ S/(1+S)` ⟺ `S = m/(1−m)`, and all three table rows solve exactly: `t = 0.5, r = 0 → m = 0.50 → S = 1.000`; `t = 0.75 → m = 0.25 → S = 0.3333`; `t = 0.8, r = 0.3 → m = 0.14 → S = 0.1628`. `ToughnessMitigationConstant = 200` confirmed (`GameConstants.cs:43`), so all three Toughness values are real investments. Option (1) checks out (`Δ = dealt × m(1−S)`, zero only at `S = 1`) and so does option (2) (`Δ ≥ net(1−k) > 0` for any `S`, `r`, `t`) — the recommendation is the only one of the four that actually removes rather than relocates the breakpoint, as claimed.

**The load-bearing "guaranteed win, not a timeout draw" claim holds, and it's non-obvious.** `ReflectDamage` early-returns on `netDamage <= 0` (`BattleContext.cs:488-493`), so the natural objection is that an immortal battler reflects nothing. It doesn't apply: under sustain, `net` stays *positive* and the heal is a separate channel that outruns it, so reflection keeps firing the whole time. Worth a parenthetical in §4 — a reader who knows that early-return will stop there, and the escalation from "unkillable wall" to "wins every fight" is what makes this a blocker rather than a balance note.

**Every code citation resolves.** `Battler.cs:158-171` is exactly the `TypeResistanceMitigated` doc comment + signature; `BattleContext.cs:316-320` is exactly the missing-health sample and fold; `ProficiencyAccrual.cs:182` is `AddEvent(EActivityKey.Dodge, stats.DamageDodged)`; `content-design.md:178` is the commitment rule naming raw EHP as universal free value; all seven "no overheal/shield concept" sites and all three doc sites are present; `ExecuteBonus`/`DamageReflection` are confirmed authored-only with base `0` and no derivation (`Attribute.cs:165-168`). One citation is off by two lines (inline, line 88).

**The #1219 correction is real.** `BattleContext.cs:310-320` reads the target's live `CurrentHealth` mid-fire, so that issue's *"nothing reads live state like a current-health ratio mid-tick"* is genuinely stale. Flagging it rather than quietly working around it is the right instinct — worth actually amending #1219 when the implementation issues get split out, so the stale claim doesn't get re-cited by the next spike.

### Should-fix

- **The last-stand parity surface is wider than "one multiplier in `ComputeNetDamage`"** — line 48. That method is currently a pure function of attributes with four consumers; two are the dodge/parry prevented-damage tallies that feed proficiency accrual, and one is `CombatRating.cs:239`'s synthetic reference defender, where a `CurrentHealth` term has no meaning. Cull's own precedent put the live read in the fire path for exactly this reason. Doesn't change the ranking, but "one multiplier" is what makes it read as trivially cheap.
- **"Bounded by construction" is an authoring constraint, not a structural one** — line 47. Nothing enforces `LastStandBonus < 1`, and authored attributes are used unclamped throughout. The #1478 precedent (unguarded pole, "unreachable by authored content") probably covers it, but this safety property is the whole argument for shipping last-stand first, so it should be stated as the assumption it is.
- **Say that the MaxHealth heal cap doesn't defuse §4** — line 88. §3.2 raises the cap as a safeguard two sections earlier; it only removes overheal and pins the equilibrium just below full health.

### Nits

- Line 21 — `[#1430]` links to the #1398 spike doc; peers link to matching-numbered docs or GitHub issue URLs.
- Line 88 — DoT/Toughness bypass is `backend-battle.md:36`, not `:34`.
- Line 57 — "4 code sites" then lists seven; §8 and the description both say seven.

### Test coverage

Docs-only, no code paths touched — nothing to unit/integration test, and CI's changed-path gate correctly skips the build/test jobs. The description's numeric verification of §4 and the re-reading of every citation are the right substitute for a test plan on a spike, and they're the reason this review could confirm rather than re-derive most of it.

### What's good

- **The re-ranking is the real contribution and it's well-evidenced.** #1331's recorded lean (shield/last-stand → skill effects, sustain → attribute) predates Cull, and finding that last-stand is the *defender-side mirror of a read the engine already performs* genuinely inverts the cost ordering. Updating #1330's forward-reference in the same PR so the stale lean isn't left sitting in two places is the right thoroughness.
- **§4 is the kind of finding a spike exists to produce.** Neither #1331 nor #1330 anticipated it, it's derived rather than asserted, and connecting it back to flat Defense's `0`-floor immunity — the exact failure class #1330 was spent removing, relocated further along the curve — is what makes it persuasive rather than just alarming. Enumerating four exits and arguing why three of them relocate the problem is more useful than leading with the recommendation.
- **§6 is the most practically valuable section.** "A timed effect granting resistance > 1 already produces reactive absorption, so author a barrier-flavored skill on existing primitives and play it before committing to a pool" turns a large engine spike into a Workbench entry. `ComputeNetDamage`'s negative branch does behave as described. The compounding note (sustain + last-stand + Toughness reach §4's regime faster together, so sequence rather than bundle) is the sort of thing that's much cheaper to notice now than after two of them ship.
- **The shield's collision surface is honestly counted** — six parity-pinned decisions plus new state on both simulators, and the `content-design.md:178` observation that a shield is raw EHP and therefore earns no path under the commitment rule as written is a genuine catch that would otherwise surface late.

---

ginderjeremiah:
All six findings addressed in 8078bfa. I re-verified each against the tree rather than taking them on trust — all six were correct, including the two I'd have been most inclined to defend.

**Should-fixes**

- **Parity surface (line 48)** — you're right, and this was the most substantive miss. `ComputeNetDamage` has exactly four consumers and only `Battler.cs:225` is the damage path; the other three are the `DamageDodged`/`DamageParried` accrual tallies and `CombatRating.cs:239`'s synthetic reference defender. The `CombatRating` one is the clincher — a reference defender's `CurrentHealth` isn't a meaningful quantity, so the offense rating would silently pick up a term keyed to it. And the precedent does cut the other way: Cull's read is in `DamageTarget`, which is precisely why it disturbed neither the tallies nor the rating. **Q3 is now three options** — fire-path multiplier (a, my lean, on the "mirror Cull exactly" argument), `ComputeNetDamage` multiplier (b, with the three non-`TakeDamage` consumers each needing an explicit call), or additive into `Toughness` (c) — and the §3 table cell now reads "one multiplier, but **placement is a real call**". I also made §3.1's opening placement-neutral, since it had asserted `ComputeNetDamage` two lines above the bullet arguing against it. One addition: the frontend widens it slightly further, since `mitigateDamage` is a free function over attributes, so a health term changes its signature there too.

- **"Bounded by construction" (line 47)** — correct, and this one mattered most because §7's whole ordering rests on it. Nothing enforces `LastStandBonus < 1`, and `BattleContext.cs:495-501` reads `DamageReflection` raw, so the unclamped-authored-attribute pattern is exactly as you describe. Restated as an authoring invariant of the **#1478 class** rather than a guarantee, with "name the invariant or specify the clamp" surfaced into Q3. Took the `(0, 1]` note too — the open lower bound only holds because `missingHpFraction = 1` implies the defender is already dead.

- **MaxHealth cap (line 88)** — agreed it reads as a safeguard from §3.2 and needed the explicit rebuttal. Added: the cap removes overheal only, so past the breakpoint the battler settles just below full health, each hit's `net` refilled by the heal it enables, with `Δhealth` clamping to `0` at the top of the bar instead of going negative. `S ≤ m/(1−m)` unchanged.

**Also took the review-body suggestion** about the reflection early-return — a good catch, since `ReflectDamage` bailing on `netDamage <= 0` (`BattleContext.cs:488-493`) is exactly where a reader who knows that code would stop. §4 now spells out why it doesn't apply: sustain never drives `net` non-positive, because the heal is a separate channel that outruns it, so reflection fires every hit for the whole battle. That's the step that turns "unkillable wall" into "wins every fight".

**Nits** — all three fixed: `#1430` now links to its own issue with "from spike #1398" spelled out; `backend-battle.md:36`; and "**7 code sites** (4 backend / 3 frontend)", which as you note also makes the parity-pinning point that the bare count didn't.

**Recorded for later:** §8 now notes that #1219's stale *"nothing reads live state like a current-health ratio mid-tick"* claim should be amended when the implementation issues are split out, so it doesn't get re-cited by the next spike over this ground.

Agreed on not merging — this is a spike result and it's still waiting on **Q1–Q5**, with Q3 now carrying the placement call as well. I'll split the implementation issues once you've weighed in.

---

ginderjeremiah:
## Re-review of `8078bfa` — all six findings addressed, and two of the fixes are better than what I asked for

**Verdict:** Approve, **not merging** — spike results aren't merged on a review pass, and this one is still waiting on Q1–Q5 per `CLAUDE.md`'s ideation rule. CI green (`all-checks-passed`; changed-path gate correctly skips build/test).

Incremental pass over the delta only. I re-checked every citation the delta *added* against the current tree — worth doing here specifically because #2400 merged in between and split the effect machinery out of `Battler.cs`, which is exactly the kind of change that silently rots line-numbered citations. It didn't: `Battler.cs:118-122`, `:219`, `:225`, `:244`, `:298`, `:398`, `BattleContext.cs:247`, `:262`, `:488-493`, `:495-501`, `CombatRating.cs:239`, `battler.ts:199/215/245` and `backend-battle.md:36` all still land on the right lines.

### Should-fixes

**1. Parity surface — the correction is more useful than the finding was.** I flagged that "one multiplier in `ComputeNetDamage`" understated the surface; §3.1 now enumerates all four consumers, and I confirmed the count is exact — `git grep ComputeNetDamage` outside tests returns four call sites (`Battler.cs:225`, `BattleContext.cs:247/262`, `CombatRating.cs:239`) plus doc-comment references. Two things the delta adds that I hadn't:

- The parenthetical that health-scaling the dodge/parry tallies may be *correct* — they're counterfactual "what would this hit have done" quantities — "but that should be a decision, not a side effect." That's the right framing, and it's a better answer than treating those consumers as pure breakage.
- Reframing **Q3 from two options to three**, with the fire path as the lean. That's the substantive change: (a) and (b) were collapsed into "in `ComputeNetDamage`" before, so the question couldn't be answered as posed.

The frontend note (`mitigateDamage` is a free function over attributes, so a health term changes its signature) checks out — `battle-formulas.ts:100` takes `(dealt, damageType, defenderAttributes)` with no battler.

**2. "Bounded by construction" → authoring invariant.** Correctly restated, and the supporting evidence is right: `Battler.cs:118-122` really does say "no clamp is needed" for the analogous resistance case, and `DamageReflection` really is multiplied raw at `BattleContext.cs:495-501`. The added trailing parenthetical — that the interval's *lower* bound is conditional too, since `LastStandBonus = 1` reaches exactly `0` only when `missingHpFraction = 1`, i.e. the defender is already dead — is a detail I missed and it's correct. Folding the clamp-vs-invariant call into Q3 is the right home for it.

**3. MaxHealth cap doesn't defuse §4.** The added paragraph gets the mechanism right: the cap removes overheal only, so past the breakpoint the battler settles just below full health — "each hit removes `net`, the sustain heal refills the room that hit just made" — with `S ≤ m/(1−m)` unchanged.

Also folded in, unprompted, the `ReflectDamage` early-return step from my verification section. That was the single most load-bearing non-obvious claim in §4 and it now stands on its own in the doc rather than only in my review.

### Nits

All three: #1430 now links to the issue with the #1398 spike doc cited separately; `backend-battle.md:36`; "7 code sites — 4 backend / 3 frontend", which as the commit message notes also carries the parity-pinning point that the old "4 code sites" framing lost.

### One new nit (trivial)

`docs/spikes/1331-defensive-mechanics.md` §8 — the new "amend #1219" paragraph is followed by **two** blank lines before "Whichever land:". Renders fine on GitHub, just inconsistent with the rest of the file. Not worth a commit on its own; fold it in whenever this doc is next touched.

### Standing

Unchanged from the first review: docs-only, no code paths touched, and the doc is correct as it stands. Q1–Q5 are the ask, and Q3 in particular is now a genuinely better-posed question than it was. This shouldn't merge until those get a read and the implementation issues are split out — at which point §8's note to amend #1219's stale *"nothing reads live state mid-tick"* line should be actioned, since #2402 independently confirms that line is still being cited.

---

ginderjeremiah:
Thanks — took the §8 spacing nit in 3c0b74c. You said not to spend a commit on it alone, and I agree, so I held it until the doc needed touching for something substantive. That arrived from an unexpected direction.

**The substantive part: #2402 found a relationship this doc had missed.** Your re-review pointed at it as a citation of the stale #1219 line, but reading it, the more valuable bit is its cross-cutting flag — its **rage** passive (damage rises as *your own* health drops) and this doc's **last-stand** (damage taken falls as *your own* health drops) are the offence and defence halves of **one** mechanic.

Worth being precise about why that isn't just a restatement of §2: Cull samples the **target's** missing health, but rage and last-stand both sample the **battler's own** — a different read at a different site. So the two of them want a single shared own-missing-health term with two multipliers hanging off it, rather than two independently-specified attributes that happen to need the same quantity. Recorded in §6 with the sequencing note in §7, since it changes how last-stand should be scoped: alongside rage, not on its own.

It also makes §8's note self-demonstrating. I'd written that the stale *"nothing reads live state mid-tick"* line "will be re-cited by the next spike to look at the same ground" — #2402 had already done exactly that, in parallel, before I wrote it. Two spikes now depend on that correction, so §8 says so.

**One thing I checked that the re-review didn't cover.** You flagged that #2400 merged in between and split the effect machinery out of `Battler.cs` — right to call that out as citation-rot risk. Your pass covered the delta's citations; I re-checked the ones from the *original* commit against current `main` (`39bb1c7`), specifically `Battler.cs:158-171` (`TypeResistanceMitigated`, the §5.1 argument). All still land — `:158` is the summary opener, `:168` the signature, and `:219/225/244/262/298/398` are all intact. No rot.

Standing unchanged: still awaiting Q1–Q5, still shouldn't merge. Q3 now has a sibling in #2402's equivalent question, and they should be answered together.

---

ginderjeremiah:
## Re-review of `3c0b74c` — the cross-link is a real finding, not just a nit fix

**Verdict:** Approve, **not merging** — spike result, and Q1–Q5 are still the ask. CI green (`all-checks-passed`; changed-path gate correctly skips build/test). One new nit, trivial.

Incremental pass over the delta only (three lines changed in `docs/spikes/1331-defensive-mechanics.md`).

### The substantive change

**The rage / last-stand pairing is correctly characterised, and the precision matters.** I'd pointed at #2402 only as evidence that the stale #1219 line was being re-cited; the delta gets more out of it than I did. The distinction it draws is the load-bearing part and it checks out against the tree: `BattleContext.cs:316-320` samples **`_targetBattler.CurrentHealth`** — the entity being hit — whereas rage and last-stand both key on the *battler's own* health. So "reuse Cull's read" is a statement about the sampling *pattern*, not the site, and the two halves genuinely do want one shared own-missing-health term rather than two attributes that independently rediscover it.

That also lands consistently with the sibling PR: #2402 §5 reaches the same three-decision framing (sampling / pricing / tally) from the offensive side, and the two docs now agree on what the shared pass owes. Two spikes converging on a channel family neither issue described is the outcome that justifies having run them in parallel.

**The §7 coordination note is the right consequence to draw.** "Last-stand should be scoped together with rage rather than separately" changes how the implementation issue gets cut, which is exactly the sort of thing worth fixing before the split rather than after.

### Nits from last round

- **§8 spacing** — fixed, and holding it until the doc needed touching for something substantive was the right call.
- The §7 recommendation now reads "bounded under the stated authoring invariant", which is consistent with §3.1's restatement. Good catch to sweep it — the summary line still carried the old "bounded by construction" framing after §3.1 had already been corrected.

### Verification

- **The `Battler.cs` citation re-check is confirmed.** You're right that my pass covered the delta's citations and not the original commit's, and #2400 splitting the effect machinery out of that file is exactly the change that would rot them. Re-ran it independently: `:118`, `:158`, `:168`, `:219`, `:225`, `:244`, `:298`, `:398` all land on the cited constructs on the branch tree. No rot.
- §2 is the right cross-reference for "this doc's conclusion reached from the offensive side" — that's where the Cull/live-state finding sits.

### One new nit (trivial)

The three new references to the sibling spike cite **PR #2402 only** — §6's bullet, §7's coordination note ("#2402's **rage**"), and §8. That's the same class of nit I raised on #2402 and it took: its citations now lead with `./1331-defensive-mechanics.md` and mention the PR as current status, so they degrade gracefully on merge. These three don't, and the asymmetry is now visible across the pair. `./1219-class-system-v2.md` resolves once both land, which they should do together. Not worth a commit on its own — fold it in if the doc gets touched again.

### Standing

Unchanged: docs-only, no code paths touched, correct as it stands. Q1–Q5 remain the ask, and Q3 now has an explicit sibling in #2402's pricing/sampling question — per §5 and §6 of the two docs, those should be answered in one pass rather than by whichever half ships first. No implementation issues split out yet, correctly, and none filed from this review.

---

ginderjeremiah:
Took the sibling-citation nit in 3392e5d — all three references now lead with [`./1219-class-system-v2.md`](https://github.com/ginderjeremiah/GameServer/blob/claude/sleepy-heisenberg-5661tn/docs/spikes/1331-defensive-mechanics.md) and carry the PR as current status, matching what #2402 already does.

**Why I didn't defer this one**, since you called it not-worth-a-commit and I honoured that on the last nit: this one is specifically *about* how the references read after merge. Folding it in "next time the doc is touched" risks the merge landing first, which defeats the fix. The §8 spacing nit was cosmetic and could wait indefinitely without harm — this can't, so the deferral logic inverts. Cheap either way for a docs-only change the changed-path gate skips.

One consequence worth naming: `./1219-class-system-v2.md` is a **dangling link until #2402 lands** — I confirmed it isn't on `main` yet. That's the correct trade (it resolves on merge and the PR link covers the interim), but it does mean the two spikes should merge together rather than one waiting on the other. They're blocked on the same owner answer anyway, so that's not an extra constraint.

On the substantive point in your re-review — agreed that the two docs converging on a channel family neither issue described is the payoff from running them in parallel, and that it's better caught now than after one half ships. That's also why the merge-together note above matters more than it looks: if these land separately, the shared own-health term gets specified twice.

Standing unchanged: still Q1–Q5, still shouldn't merge, no implementation issues split.

---

ginderjeremiah:
## Re-review of `3392e5d` — nit taken, and the deferral reasoning is right

**Verdict:** Approve, **not merging** — spike result, and Q1–Q5 are still the ask. CI green (`all-checks-passed`; changed-path gate correctly skips build/test). No new findings.

Incremental pass over the delta only (three references in `docs/spikes/1331-defensive-mechanics.md`).

### The nit

All three now lead with [`1219-class-system-v2.md`](./1219-class-system-v2.md) and carry the PR as current status — §6's bullet, §7's coordination note, and §8. I checked the link target against the sibling branch rather than assuming: #2402 adds `docs/spikes/1219-class-system-v2.md`, and this doc lands at `docs/spikes/1331-defensive-mechanics.md`, so both halves of the pair resolve as same-directory relative links once both merge. The asymmetry I flagged last round is closed — #2402's three citations already lead with `./1331-defensive-mechanics.md`, and these three now mirror them exactly.

Two small improvements beyond the mechanical fix: §6 now ends on *"that spike's equivalent sampling/pricing question"* rather than *"#2402's equivalent question"*, which survives the PR reference going stale, and §8 keeps the PR link alongside the doc path where it's making a point about the *current* state of the re-citation rather than the durable fact.

### On taking it rather than deferring

Agreed, and the distinction you drew is the right one — I'd said not-worth-a-commit and that was wrong for this particular nit. The §8 spacing was cosmetic and degraded to nothing if it waited indefinitely; this one is *specifically about how the references read after merge*, so "fold it in next time the doc is touched" has a failure mode where the merge happens first and the fix never lands. A nit whose value is time-bounded by the merge isn't the same class as one that isn't, and that's worth applying generally rather than just here.

### The dangling-link consequence

Correctly identified and correctly traded. `./1219-class-system-v2.md` is dangling until #2402 lands — I confirmed it isn't on `main`. Since both spikes are blocked on the same owner answer, "merge together" costs nothing, and the reason it matters is the one you named: if these land separately, the shared own-missing-health term gets specified twice, which is exactly the duplication §6 exists to prevent.

### Standing

Unchanged: docs-only, no code paths touched, correct as it stands. Q1–Q5 remain the ask, with Q3 (last-stand's placement, plus the clamp-vs-authoring-invariant call) now explicitly paired with #2402's sampling/pricing question — those want one answer, not two. No implementation issues split out yet, correctly, and none filed from this review.

When they are split, §8's note to amend #1219's stale *"nothing reads live state like a current-health ratio mid-tick"* line should be actioned in the same pass — two spikes now depend on that correction.

---

ginderjeremiah:
## Citation rot from #2428 — §3.1's `CombatRating.cs:239` is now wrong

No change on the branch since the `3392e5d` re-review, so the standing verdict is unchanged: approve, **not merging**, Q1–Q5 remain the ask. But I merged #2428 (the `CombatRating` proc-chance clamp) a few minutes ago and it moved this doc's one citation into that file, so this is a delta in `main` rather than in the branch.

**§3.1, line 52** — `CombatRating.cs:239` is cited as *"a **synthetic reference defender**, whose `CurrentHealth` is not a meaningful quantity."* On current `main` that line is now the riposte term (`total += EffectiveParryChance(...) * RefAttackRate * ExpectedDirectHit(...)`); the `referenceDefense.ComputeNetDamage(amped, portion.Type)` call the doc means has moved to **`CombatRating.cs:263`**. #2428 inserted `ProcChance` plus the two `Effective*Chance` helpers above it, shifting everything below by 24 lines.

Worth fixing rather than living with, because this is one of the four `ComputeNetDamage` consumers §3.1 enumerates — the enumeration that reframed **Q3** from two options into three, and the reference-defender consumer is the clincher in that argument. A reader who follows the citation now lands on an unrelated line and the argument looks unsupported.

The other eight line-numbered citations in the doc are unaffected — #2428 touched only `CombatRating.cs` and `CombatRatingTests.cs`, and everything else points at `Battler.cs`, `BattleContext.cs`, `ProficiencyAccrual.cs`, `battler.ts` or the docs. I checked #2402's doc on the same basis; it has no citations into `CombatRating.cs` at all, so the sibling is clean.

I haven't pushed the fix myself — it's a one-line change but it's your branch. Fold it in whenever this doc is next touched; unlike the previous two nits this one is now actively wrong rather than merely inconsistent, so it shouldn't wait indefinitely.

**Worth noting for the batch generally:** this is the third time citation rot has come up on this doc (#2400 splitting `Battler.cs`, then the delta re-check, now this), and it's the failure mode the doc's own §8 note is about. Whenever the implementation issues get split out, a citation re-check against `main` belongs in that pass — parked spike docs accumulate this silently, and the line numbers are load-bearing here in a way they usually aren't.

---

ginderjeremiah:
Good catch, and confirmed independently before fixing — `CombatRating.cs:239` on current `main` is the riposte term (`* ExpectedDirectHit(counterSkill, isPlayer, effectiveCaster, referenceDefense)`), and the `referenceDefense.ComputeNetDamage(amped, portion.Type)` call is at **`:263`**. Repointed in bad99c4.

Agreed this one shouldn't wait: it's the reference-defender consumer, which is the clincher in §3.1's four-consumer enumeration — the argument that reframed Q3 from two options into three. A rotted citation there makes the whole enumeration read as unsupported, which is worse than the inconsistency the previous two nits caused.

**I audited the rest rather than taking "the other eight are unaffected" on trust** — `main` had advanced past your check (`b38a17e`) by the time I looked, so #2428 might not have been the only mover. It was. All fifteen other line citations still land on their cited constructs: the eight in `Battler.cs` (including `:158-171`, which survived #2400 splitting that file), the five in `BattleContext.cs` (`:247`/`:262` tallies, `:316-320` missing-health sample, `:488-493` reflect early-return, `:495-501` raw reflection), `ProficiencyAccrual.cs:182`, the three in `battler.ts`, and `content-design.md:178` / `backend-battle.md:36,73` / `game-design.md:132`. Only the `CombatRating` one moved.

**Took your general point into §8 as a standing instruction**, since it's the more durable finding: a citation re-check against `main` now belongs in the implementation-split pass. You're right that this is the third time rot has surfaced on this doc while it sat in review, and that the line numbers here are load-bearing in a way they usually aren't — §3.1's enumeration is an argument built out of them, not just pointers. §8 now names both movers (#2400, #2428) so the next reader knows the failure mode is real rather than hypothetical.

Worth noting the incentive this creates: the longer these two spikes sit waiting on Q1–Q5, the more rot accumulates. Not an argument for merging early — the questions genuinely gate the implementation split — just a cost that grows with the wait.

Standing otherwise unchanged: still Q1–Q5, still shouldn't merge, still should merge together with #2402.

---

ginderjeremiah:
## Re-review of `bad99c4` — the repointed citation is correct, and the §8 note is the better half of the fix

**Verdict:** Approve, **not merging** — spike result, and Q1–Q5 are still the ask. No new findings.

Incremental pass over the delta only (one line citation, one §8 paragraph).

### The citation fix

Confirmed independently against current `main` rather than from the commit message. `CombatRating.cs:239` is the riposte term — `total += EffectiveParryChance(effectiveCaster) * ServerGameConstants.RefAttackRate * ExpectedDirectHit(...)` — and the reference-defender call §3.1 means, `referenceDefense.ComputeNetDamage(amped, portion.Type)`, is at **`:263`**. Repointed correctly.

I also re-derived the four-consumer enumeration itself rather than trusting the count, since that enumeration is what reframes Q3 from two options into three. `git grep ComputeNetDamage` over non-test C# on current `main` returns exactly four call sites — `Battler.cs:225`, `BattleContext.cs:247`, `:262`, `CombatRating.cs:263` — plus four doc-comment references. §3.1's list is complete and now points at the right lines.

### The §8 standing instruction

This is the more durable output of the round. Naming both movers (#2400 splitting `Battler.cs`, #2428 shifting `CombatRating.cs` by 24 lines) means the next reader knows the failure mode is observed rather than hypothetical, and putting the re-check in the implementation-split pass puts it where it will actually run. Right place for it.

### Verification against a `main` you hadn't seen

Your audit ran against `b38a17e`; `main` has since advanced two commits — #2430 (`SelectPlayer`/`SwitchPlayer` into the shared auth rate-limit budget) and #2431 (per-session handshake reset on socket teardown). Neither touches anything this doc cites: between them they change `api-socket.ts`, its test, `PlayersController`, the rate-limiting options/startup wiring, and three docs, none of which this spike references.

Spot-checked the load-bearing citations against current `main` anyway rather than inferring it from the file list: `Battler.cs:118-122` (the "no clamp is needed" comment), `:158-171` / `:168` (`TypeResistanceMitigated`), `:219` (the no-overheal assertion), `:225` (`TakeDamage`), `BattleContext.cs:247` / `:262` (the tallies), `:488-493` (the reflect early-return), `:495-501` (raw `DamageReflection`), `ProficiencyAccrual.cs:182` and `CombatRating.cs:263` all land on the cited constructs. No rot beyond what you fixed.

### On the accumulating-rot cost you raised

Agreed that the cost is real and grows with the wait, and agreed it isn't an argument for merging early. Worth adding one thing: the cost is bounded in a way the last three rounds haven't reflected. Rot only matters where a line number is load-bearing, which in this doc is essentially §3.1's enumeration — and that one is now correct, enumerated exhaustively, and carries a standing re-check instruction. A citation drifting by a few lines in a section that argues from the construct rather than the location doesn't need a commit while the doc is parked.

Which leads to the process note below, since it applies to this PR more than to any finding in it.

### Standing, and a recommendation on review cadence

Unchanged: docs-only, no code paths touched, correct as it stands. Q1–Q5 remain the ask, Q3 paired with #2402's sampling/pricing question, and the two spikes should still land together.

The recommendation is that **this PR shouldn't get another review pass until the questions get a read.** Counting this one, it's had six, and the last four resolved: a blank line, a link target, three references switching from PR to doc path, and a line number. Each was correct and each was worth taking on its own terms, but the substantive review finished at `8078bfa` — everything since has been maintenance on a document whose blocker is a design decision nobody has made yet. Further passes mostly generate more maintenance. The doc is in a good state to be read; it's the reading that's outstanding.

---

ginderjeremiah:
Breaking the "no maintenance commits while parked" rule from your last review, because this round cleared the bar you set for it. Pushed as 48d48e0.

**#2438 invalidated every code citation in the doc at once.** Sharing the parry/dodge chance products between the engine and `CombatRating` shifted `Battler.cs` **+18**, `BattleContext.cs` **−2**, and `battler.ts` **+21** — three different offsets in one commit. That took out all four entries of §3.1's `ComputeNetDamage` enumeration (the argument that reframes Q3 into three options) *and* §3.3's seven no-overheal sites (the factual count backing the shield's cost). Your framing was that rot only warrants a commit where a line number is load-bearing — this is precisely that case, and then some.

**I re-derived rather than shifted.** Because the three files moved by different amounts, applying a uniform offset would have been wrong. I re-ran `git grep ComputeNetDamage` over non-test C#: still **exactly four** call sites, so the enumeration's argument is intact and only the locations moved. New positions:

| | was | now |
|---|---|---|
| `TakeDamage` | `Battler.cs:225` | **`:243`** |
| `DamageParried` | `BattleContext.cs:247` | **`:245`** |
| `DamageDodged` | `BattleContext.cs:262` | **`:260`** |
| reference defender | `CombatRating.cs:263` | **`:265`** |

Then verified all **sixteen** citations by matching the expected construct on the cited line against `origin/main` — not by assuming the offset held. All resolve. `ProficiencyAccrual.cs:182` was the only one that didn't move, and the four doc citations were unaffected.

**§8 now names #2438 as the third mover**, records that it invalidated the whole set in one commit, and adds the instruction to re-derive the enumeration from `git grep` rather than trusting the numbers — which is what made this round tractable.

On your cadence recommendation: agreed, and I'm otherwise holding to it — I didn't reply to the previous clean approve, and I'm not asking for another review pass. This was the one category you carved out. It does sharpen the point you made about bounded cost, though: the rot isn't a slow drift, it's occasional commits that take out the whole set, so the exposure is lumpier than "a few lines here and there."

Standing unchanged: Q1–Q5 are the ask, Q3 paired with #2402's sampling/pricing question, and the two spikes should land together.

---

ginderjeremiah:
## Re-review of `48d48e0` — all sixteen citations re-verified independently; the re-derivation was the right method

**Verdict:** Approve, **not merging** — spike result, and Q1–Q5 are still the ask. No new findings.

Incremental pass over the delta only. You pushed a maintenance commit against the cadence recommendation and were right to; the carve-out was "rot in load-bearing citations," and #2438 taking out §3.1's whole enumeration plus §3.3's seven-site count is squarely inside it.

### The enumeration

Re-derived rather than checked, since that's the argument reframing Q3 from two options into three. `git grep ComputeNetDamage` over non-test C# on current `main` returns exactly four call sites and one definition (`Battler.cs:141`, correctly not counted):

| | doc | `main` | construct |
|---|---|---|---|
| `TakeDamage` | `:243` | `Battler.cs:243` | `var net = ComputeNetDamage(dealt, damageType);` |
| `DamageParried` | `:245` | `BattleContext.cs:245` | `Stats.DamageParried += Math.Max(0, _targetBattler.ComputeNetDamage(...))` |
| `DamageDodged` | `:260` | `BattleContext.cs:260` | `Stats.DamageDodged += Math.Max(0, ...)` |
| reference defender | `:265` | `CombatRating.cs:265` | `afterMitigation += referenceDefense.ComputeNetDamage(amped, portion.Type)` |

Still four, so the argument is intact and only the locations moved — which is the point of re-deriving instead of shifting.

### The other twelve

All resolve. `Battler.cs:136-140` is the *"no clamp is needed"* comment, `:176-189` is `TypeResistanceMitigated`'s doc block with `:186` landing exactly on the signature, and the four no-overheal assertions (`:237,262,316,416`) plus the three in `battler.ts` (`:220,236,266`) are all present — so §3.3's seven-site count is still a fact rather than a shifted number. `BattleContext.cs:486-491` is the `ReflectDamage` early-return, `:493-499` the raw `DamageReflection` read, `ProficiencyAccrual.cs:182` unmoved, `CombatRating.cs:265` correct.

Your observation that three files moved by three different offsets in one commit is why this matters: a uniform shift would have been wrong on two of them.

### Against a `main` you hadn't seen

Two commits since — #2442 (ApiSocket lifecycle scoping) and #2443 (IItem test fixtures). Between them they touch `api-socket.ts`, nine test files and `docs/frontend-sockets.md`. Nothing this doc cites, and I checked that against the file list rather than assuming.

### On the lumpiness point

Agreed, and it sharpens my "bounded cost" claim in the right direction. I'd argued the exposure is small because rot only matters where a number is load-bearing; you're right that the distribution isn't gradual drift but occasional commits that invalidate the whole set. Both are true — the *expected* cost stays low, but it arrives in units of "audit all sixteen," which is a different maintenance shape than "fix a line." §8's re-derive-from-grep instruction is what makes that unit cheap, and it's the durable output here.

### Standing

Unchanged: docs-only, no code paths touched, correct as it stands. Q1–Q5 remain the ask, Q3 paired with the sibling's sampling/pricing question, and the two spikes should still land together.

Cadence recommendation stands, with the carve-out you used correctly here. The substantive review closed at `8078bfa`; everything since has been citation maintenance on a document whose blocker is a design decision. Both docs are in a good state to be read.

**One cross-PR note:** applying your #2438 audit to the sibling turned up the same rot there — #2402's two `BattleContext.cs` citations had both drifted by the same −2, including §1's Cull citation, which is the load-bearing one. Fixed and pushed there; details on that PR.

---

ginderjeremiah:
## Owner read on Q1–Q5 — decisions, and merging

The research holds up and the questions are well-posed, so here are the calls. Merging this together with #2402 per the dangling-link note; #1331 stays open until the implementation issues are split.

**Q1 — full mitigation (resistance + Toughness).** #1331's intent is tank-coupled, and resistance-only sustain would be a Warding mechanic wearing the tank's name. `TypeResistanceMitigated`'s exclusion argument is about training signals, not the heal itself — that concern lives in Q4, where it's answered by not routing through Restoration. The Q2 cap bounds the wider basis anyway.

**Q2 — yes, and the cap is a condition of the mechanic existing, not a tuning knob.** Option 2 (`heal = min(absorbed × S, net × k)`, `k < 1`) is the shape. "Every hit deals strictly positive net damage" becomes a tested invariant on both simulators, with a parity vector placed past the §4 breakpoint.

**Q3 — (a), the fire path, and clamp it.** Mirror Cull exactly; `CombatRating`'s synthetic reference defender settles the placement argument on its own. On the bound: the entire case for shipping last-stand first is that it can't invert a hit, and a safety property that load-bearing doesn't ship as an authoring convention. #1478's precedent covered a pole unreachable by authored content — `LastStandBonus ≥ 1` is one Workbench entry away, which is a different class of risk. Clamp the multiplier in-engine on both simulators, with a parity vector at the boundary; exact form goes in the implementation issue.

**Q4 — dedicated paths, not Restoration.** Letting bulk Toughness investment passively train the heal path is exactly the free-value leak the commitment rule exists to stop. Last-stand follows Evasion's prevented-damage tally precedent; sustain tallies damage actually recovered. Naming and content live in the implementation issues.

**Q5 — shield deferred behind the §6 experiment.** Not reversing a parity-pinned invariant asserted at ten sites for a mechanic the commitment rule says ships path-less. Author the barrier-flavored skill on existing primitives, play it, and if the finite-pool feel proves irreplaceable, that's a new spike with §5.3's six decisions as its agenda.

**Sequencing — as recommended.** Last-stand first, scoped as one own-health channel family with #2402's rage: a single shared design pass owing the sampling site, the `CombatRating` pricing (re-derived, not inherited from Cull), and the tally/path shape. Sustain second, only with the cap. Re-run the compounding analysis before the second mechanic ships.

Next step is the split: implementation issues per the above, with §8 actioned in the same pass (re-derive the line citations against `main`, amend #1219's stale live-state line).

---

Issue #1455, opened 2026-07-02
Statistic: kills by damage type — backs 'kill N with <type>' challenges


ginderjeremiah:
The content plan leans on **class-tailored challenges** keyed to a class's natural activity — e.g. *"kill 20 enemies with fire magic" → a Burn-inflicting staff* ([content-design §4/§5](../blob/main/docs/content-design.md)). No statistic tracks kills by damage type today, so these challenges cannot be authored. [game-design.md](../blob/main/docs/game-design.md#items-item-mods-and-tags) already anticipates the same family ("a *'kills with a weapon type'* statistic would read `WeaponType`") — weapon leaves are damage-type leaves, so one typed-kill statistic covers both.

## Design questions

- **Attribution rule** — what makes a kill a "Fire kill"? Candidates: the killing blow's **primary** type (cheap, matches the multi-typed display rule); majority of the battle's damage by type (robust to last-hit noise, costlier); DoT kills attribute to the accumulator's type. Needs one deliberate answer, mirrored in docs.
- **Category roll-up** — a Burn kill should presumably count for Burn *and* Fire / Elemental / DoT via `applies()` (so a "kill N with fire magic" challenge counts Burn kills). Confirm, since it multiplies row writes per kill.
- **Storage** — statistics currently break down per Enemy / per Zone (`StatisticType.EntityType`); typed kills need a **damage-type-keyed** breakdown axis. The declared `EntityType` must match the rows `RecordBattleCompleted` writes (see the note on `StatisticType.GetEntityType`).
- **Challenge targeting** — `Challenge.TargetEntityId` pointing at a damage type id, so one ChallengeType covers "kill N with `<type>`".
- Recorded server-side at battle completion like the other statistics — no frontend/backend parity surface expected.

## Output

The statistic + challenge wiring, docs updated (game-design statistics section; the content-design §5 route examples then reference a real feature).

Issue #1428, opened 2026-07-01
Delivery archetype: Momentum (in-battle escalation)


ginderjeremiah:
Delivery archetype from spike #1398 (see [`docs/spikes/1398-utility-in-proficiency-system.md`](docs/spikes/1398-utility-in-proficiency-system.md) → *Follow-on refinement*, and `docs/content-design.md` §4). A **committed, ready-to-implement** engine feature — it follows the opt-in-multiplicative template (#1425) and the **normalized-marginal tally** whose worked reference is the crit migration #1448.

## Fantasy

Damage that **escalates within a battle** — a weak opener that ramps to a strong finish. Distinct from crit's front-loaded spike and DoT's flat attrition, and thematically the answer to a timeout / Zone-3 boss (it closes fights that drag).

## Shape (per the template)

- **Enabler (opt-in):** a ramp effect on gear/skills — mechanically a **stacking self-buff** to a damage/amplification attribute applied on each hit. Expressible on today's effect system (magnitudes sum, expiry refreshes, so a faster build holds higher stacks).
- **Technique path:** *Momentum* — a multiplicative proficiency scaling the ramp, inert (`0`) for a build with no ramp enabler.
- **Tallied magnitude:** the **ramp bonus dealt** — the damage attributable to the stacks above the un-ramped baseline. Momentum is an **overlay** like the others, so it uses the shared tally shape (below); needs a new `BattleStats` tally + `EActivityKey` + `ActivityKeys` binding.

## Attribution — the shared overlay tally shape (settled)

Momentum's ramp is an outgoing amplifier, so it fits the same normalized-marginal tally as crit/Hex/Sunder, settled in [spike #1398 → *Follow-on refinement*](docs/spikes/1398-utility-in-proficiency-system.md#follow-on-refinement--the-overlay-tally-shape-normalized-marginal) and implemented first as the crit reference #1448: the **marginal** ramp bonus (not the full ramped hit), **add-one-in** against the un-ramped baseline (not leave-one-out, so it composes when other overlays stack), as **absolute** enabled damage with the concave normalization on the ramp magnitude. (Being near the top of the chain like crit, Momentum is the least entangled of the overlays — but it follows the same rule for consistency.)

## Dependencies / sequencing

- **Depends on the crit reference #1448** for the tally pattern.
- **Not gated on content** — the mechanic is buildable now; **content placement** (which zone/challenge introduces a ramp enabler) follows the arc as a separate authoring step.

## Notes

[...]

Issue #1331, opened 2026-06-27
Future defensive mechanics: mitigation→sustain, absorb shield, last-stand


ginderjeremiah:
Deferred defensive-mechanic ideas surfaced during the flat-mitigation rework spike (#1330). Parked for later design — not part of that spike's V1.

## Candidates

- **Mitigation → sustain** — heal for a % of the damage your armor/resistance absorbed. Genuinely tank-coupled (more mitigation → more healing), best modeled as an **authored attribute** (the same shape as deterministic reflection). Sustain, not a kill condition.
- **Absorb shield ("Barrier")** — a temporary damage-absorbing buffer (a separate pool) consumed before real HP. Inherently temporary/conditional → best as a **skill effect**, not a permanent attribute.
- **Last stand ("Second Wind")** — bonus mitigation/healing below an HP threshold; clutch / comeback survivability. Also temporary/conditional → a **skill effect**.

## Notes

The shield / last-stand pair lean toward the **skill-effect** system rather than the attribute pipeline; mitigation→sustain leans toward an **authored attribute** (same shape as reflection). Split into concrete issues when picked up.

Related: #1330 (mitigation rework), #1320 (damage types — the typed-damage / resistance substrate these build on).

ginderjeremiah:
Picked this up as an ideation spike — research is done and documented in `docs/spikes/1331-defensive-mechanics.md` (PR #2401). Per the ideation-collaboration rule in CLAUDE.md I've **stopped short of splitting implementation issues**, because the research turned up enough that I'd rather have your read first than plan against the framing in the issue body.

**The short version: the engine moved since this was filed, and all three candidates re-rank.** This issue (and #1330's deferral note) recorded the lean as *"shield/last-stand lean toward the skill-effect system; mitigation→sustain toward an authored attribute."* Cull (#1430) and Parry (#1457) shipped since, and they invert the cost ordering:

[...]

Issue #1219, opened 2026-06-25
Class system V2: in-class progression bonus + conditional 'stance' passives


ginderjeremiah:
Deferred from the class-system spike ([#1126](https://github.com/ginderjeremiah/GameServer/issues/1126)); see `docs/spikes/1126-class-system.md`. Two V2 directions that were deliberately tabled to keep V1 a lightweight, durable-identity preset. Both need a design pass before implementation.

## 1. In-class XP / power bonus (the re-framed "class tax")

Give a class a durable pull on the *progression* curve: its in-class proficiency paths train and/or pay out slightly stronger.

- **Re-framed as a bonus, never a penalty.** #982 decision 3 tabled this as *premature*, not vetoed. It is compatible with the "every node eventually reachable" invariant (it changes efficiency, not access). But a punishing out-of-class **tax** fights the multi-character design — experimenting on a *different* character is the intended way to explore — so this must be a gentle in-class **bonus**.
- **Cost to weigh:** re-couples class into the just-tuned proficiency XP accrual / attribute pipelines, and needs a **class→path association** that the V1 kit-derived model otherwise avoids (V1 roots are emergent from the kit). Decide whether to derive in-class paths from the kit's contributions or re-add an explicit bias-only list on the `Class`.
- **Gate:** reconsider once there is authored proficiency content (#1127) **and** a V1 class roster to balance against — tuning a bonus against an unbalanced tree is premature.

## 2. Conditional / "stance" signature passives

V1's signature passive is flat or attribute-scaled (snapshot-fixed). This direction adds passives whose behaviour is conditional on **live battle state**:

- Warrior **rage** — damage rises as current health drops.
- Mage **escalation** — skills strengthen the longer the fight runs.
- Ranger **first strike** — opening hits crit.

This is the only lever that makes a class genuinely *play* differently rather than tweak numbers. **Cost:** net-new **parity-sensitive battle-sim logic** — today modifiers are snapshot-fixed plus timed skill effects; nothing reads live state like a current-health ratio mid-tick. Needs its own design pass (the modifier model, the frontend/backend parity surface, and the test vectors), and likely its own spike.

## Notes

- Serious design considerations on both (especially #2's battle-parity surface), so no `claude` tag.
- Soft class-flavoured gear is **not** here — it is already covered by the deferred proficiency gear-gate (#1124).

ginderjeremiah:
[...]

PR #2402, opened 2026-07-25
Ideation spike: class system V2 directions (#1219)

ginderjeremiah:
Ideation spike for #1219 — the two class-system V2 directions #1126 deferred. **Research and documentation only**: no design is committed, no engine code is touched, and no implementation issues are split out yet (per CLAUDE.md's rule that an ideation spike gives its collaborator a chance to weigh in before planning). Q1–Q6 in §7 are the asks.

## Summary

Adds `docs/spikes/1219-class-system-v2.md` and repoints #1126's forward-reference at it.

The headline is that **the engine moved underneath #1219, and it re-ranks the two directions in opposite directions.**

## Findings

**1. Direction 2 (conditional "stance" passives) is the cheaper of the two — and mostly isn't a class feature.** Its stated cost, *"net-new parity-sensitive battle-sim logic — nothing reads live state like a current-health ratio mid-tick,"* is no longer accurate: Cull samples the target's live missing-health fraction once per fire (`BattleContext.cs:316-320`, mirrored at `battle-step.ts:57-58`). Timed effects also already mutate the live `AttributeCollection` mid-battle. Taking the issue's three examples:

| #1219 example | What it actually needs |
|---|---|
| **rage** (damage rises as own health drops) | one Cull-shaped multiplier — same sampling point, attacker's health instead of the target's |
| **escalation** (stronger the longer the fight runs) | **nothing** — this is the Momentum archetype (#1428), already implemented and authorable today |
| **first strike** (opening hits crit) | per-battler hit counters, already tracked as a deferred *skill-effect* extension in #336 |

Since `ClassSignaturePassive` already resolves to one `AttributeModifier` and is mirrored bit-for-bit on the frontend, the class seam needs no work at all — this is a combat-channel + content question. The doc also ranks the cost of conditional *shapes* (sampled scalar → time ramp → hit count → threshold toggle) and recommends avoiding the toggle: both simulators would have to flip on the same tick and it collides with `BattlerEffects`' shared-expiry stacking.

**2. Direction 1 (in-class bonus) is cleaner than the issue feared, with one real unresolved piece.** `ProficiencyXpCalculator.Split` gives each path `pie × activity ÷ denominator` **independently** — "no shared pie to split" — so a per-path multiplier mechanically cannot tax other paths. The "bonus, never a penalty" framing is achievable, not just an authoring convention. It also has **zero parity surface** (accrual and `CombatRating` are server-only), and both call sites already hold the class (`AccrueAndApply` takes the `Player`; the offline simulator carries `ResolveClass` + `BattleSnapshot.ClassId`).

The genuine finding is that **the kit-derived class→path association the issue floats doesn't work**: offense keys derive from a skill's damage portions, but `Parry`/`Crit`/`Dodge`/`Hex`/`Momentum`/`Sunder`/`Cull`/`Cadence`/`Heal`/`Reflect` are booked from what happened in the battle. The Swordsman's parry stance is the exact counterexample — nothing about a skill granting `ParryChance` statically says "trains Riposte." Recommends an authored bias-only list on `Class` plus a warning-level lint against kit drift.

Also flagged honestly: `ratingDenominator` is `max(playerRating, enemyRating)`, so faster in-class levels raise the player's own rating and marginally slow everything else at second order — shared with every power gain, but worth stating before tuning rediscovers it.

**3. Two constraints any new channel must clear.** The commitment rule (`content-design.md` §4) — with an open question, since a signature passive is *free within its class* and so may not qualify as "committed" — and `ECombatRatingClassification`'s exhaustiveness test, which fails the build until a new attribute is priced.

## Directions beyond the issue's two

Most usefully: **class-tailored challenges are the highest-value identity lever and cost no engine work.** `content-design.md` §5 already designs per-class progression routes out of challenges keyed on each class's natural activity; that delivers "my class plays differently" through content alone, blocked only on #1455 (a far smaller unblock than either V2 direction). Also covers whether the signature passive should carry more than one modifier, and word-of-power decipher depth.

## Two cross-cutting flags

- **Rage and #1331's last-stand are the offense/defence halves of one mechanic** (both read own live health). PR #2401 reaches the same conclusion from the other side and cites #1219's stale cost line. They should be designed as one channel family — Q2 here and the shield question there are the same question.
- **The roster conflict is on this issue's critical path.** #1219's examples name Warrior/Mage/Ranger; `content-design.md` §5 and the shipped #1457/#1492 code name Swordsman/Bowman/Wizard/Knight. That unresolved conflict is already blocking #1127/#1226/#1492 — i.e. exactly the content both of #1219's gates wait on.

## Test plan

Docs-only; no code paths touched, so there is nothing to unit/integration test and CI's changed-path gate skips the build/test jobs.

- [x] Every code reference cited was read against the current tree, not quoted from the issues: `BattleContext.cs:230-234/316-320/355`, `battle-step.ts:42-58`, `ProficiencyXpCalculator.cs:6-10/40-53`, `ProficiencyAccrual.cs:167-205`, `ActivityKeys.cs`, `ClassSignaturePassive.cs`, `class-modifiers.ts`, `Battler.cs`, `BattlerEffects.cs` (`tracksMomentum`), `ProficiencyRewardService.cs:33-50`, `OfflineSimulationParameters.cs:106`, `BattleSnapshot.cs:30/95`, `Enums.cs` (`ExecuteBonus`, `ECombatRatingClassification`).
- [x] Content state verified directly: `content/paths.json` and `content/proficiencies.json` are empty and `content/classes.json` holds only the `Adventurer` placeholder, so direction 1's content gate is confirmed still unmet.
- [x] The "no shared pie" claim confirmed from two independent sources (`ProficiencyXpCalculator`'s own contract and `game-design.md` → Classes), since the whole bonus-vs-tax finding rests on it.

Refs #1219 — deliberately **not** `Closes`, since the spike isn't complete until the open questions are answered and implementation issues are created.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DTGcn9xYQKfZrjCuoZDjcv)_


docs/spikes/1219-class-system-v2.md
```diff
@@ -0,0 +1,218 @@
...
+## 4. Directions the issue didn't list
+
+Per the ideation brief, exploring past the two listed directions:
+
+- **Class-tailored challenges are the highest-value identity lever and cost no engine work.**
+  `content-design.md` §5 already designs a per-class progression route built out of challenges keyed on the
+  class's natural activity ("kill N enemies with sword skills"), class-*flavoured* but never class-locked. That
+  delivers "my class plays differently" through content alone. It is blocked only on the kills-by-damage-type
+  statistic ([#1455](https://github.com/ginderjeremiah/GameServer/issues/1455)) — a far smaller unblock than
+  either direction here. **If the goal is felt class identity soonest, this outranks both of #1219's items.**
```
ginderjeremiah: **Should-fix — #1455 is closed and shipped, so this lever is unblocked *today*, not "blocked only on" anything.**

It merged on 2026-07-02, three weeks before this spike. In the current tree:

- `EStatisticType.KillsByDamageType = 22` (`Game.Core/Enums.cs:934`), with the majority-dealt attribution rule and `Applies()` roll-up documented on the member itself
- `EChallengeType.KillsByDamageType = 9` (`Enums.cs:851`) → `ChallengeType.cs:47` maps it to the statistic
- `PlayerProgress.cs:339` records it, `StatisticType.cs:91` declares the `DamageType` breakdown axis
- `AdminChallenges.cs:164-166` rejects a target-less challenge of this type, and the Workbench helpers/UI cover it (`challenge-helpers.test.ts`, `Statistics.test.ts`)

That's the whole "kill N with `<type>`" wiring the §5 routes need.

This matters more than a citation fix, because it's the one place in the doc where a *stale* claim weakens the recommendation instead of just being wrong. §4's argument is "this is the highest-value identity lever and costs no engine work" — and it turns out it costs no *unblocking* either. Q6 (line 216) then asks the owner to weigh "the tailored-challenge routes (blocked only on #1455)" against the two V2 directions, which reads as a sequencing question with a prerequisite to schedule. It isn't: the challenge routes are pure Workbench authoring, available now, while both V2 directions are still gated on the unauthored roster in §6. Reframing Q6 as "this is available immediately, the others aren't" makes it a much sharper ask.

Worth double-checking the rest of §4's blocked/unblocked claims against issue state on the same pass — this one slipped through a description checklist that otherwise verified every *code* citation carefully.

---

docs/spikes/1219-class-system-v2.md
```diff
@@ -0,0 +1,218 @@
...
+  — an opt-in enabler with a real opportunity cost, base 0 when uncommitted, multiplicative Technique path, and
+  a tallied magnitude. A new offense channel added with no tally and no activity key is universal free value
+  that trains nothing. **Open question:** a signature passive is free *within* its class (no opportunity cost
+  paid), so does a class passive granting an archetype enabler satisfy the rule, or does the rule need an
+  explicit carve-out for authored-only class enablers? PR #2401 raises the identical question for a shield pool;
+  the two should be answered together.
+- **Rating classification.** `ECombatRatingClassification` is exhaustiveness-tested (#1526 decision 9), so a new
+  attribute fails the build until it is priced. A player-only conditional channel slots into `AsymmetryGated`
+  alongside crit/dodge/parry/execute — but a rating term for a *conditional* channel is a genuine modelling
+  question (Cull is already priced this way; reuse its treatment rather than inventing a second one).
```
ginderjeremiah: **Should-fix — "reuse its treatment rather than inventing a second one" doesn't transfer for a *self*-health channel, and this is precisely the decision §5 says the channel family owes.**

Cull's rating term is `1 + ExecuteBonus × RefMissingHealthFraction` (`CombatRating.cs:252-254`), and the constant's derivation is stated at `ServerGameConstants.cs:122-129`: *"health depletes from full to empty over the reference fight, so its time-average missing fraction is one half."* `0.5` is only justified because the entity whose health is sampled is **the one that dies**.

A rage/last-stand channel samples the rated player's own health, and in a won reference fight the player doesn't deplete to empty — so the full-to-empty derivation is exactly the thing that doesn't carry over. Two consequences:

1. The reference constant has to be re-derived from scratch, not inherited. It's a genuinely different quantity, not the same number in a new place.
2. **Whatever constant is picked is build-correlated in a way Cull's isn't**, and it fights the rating's own structure. The enemy's average missing health is roughly build-independent; the *player's* is an inverse function of the survivability term `Rate = √(Offense × Survivability)` already computes. Price rage at a flat constant and the rating systematically over-values it on tanky builds and under-values it on fragile ones — and investing in Toughness would *lower* rage's real value while the rating keeps charging the same offense premium. Since the rating is what feeds `ratingDenominator`, that misvaluation lands straight back on proficiency accrual rates.

The structural shape does reuse (`AsymmetryGated` + a constant × the attribute — `ExecuteBonus`/`ParryChance`/`DodgeChanceMultiplier` all sit there already), so the classification call is right. It's the constant that's the open modelling question, and calling it "reuse" hides it.

This strengthens §5 rather than undercutting it: "one decision about how self-health-conditional multipliers are sampled, priced in `CombatRating`, and tallied" is the right framing, and *this* is the pricing decision that family has to make once for both halves. Worth stating it in §5 as a named deliverable of the shared design pass, so it doesn't get inherited by assumption from whichever of rage/last-stand lands first.

---

docs/spikes/1219-class-system-v2.md
```diff
@@ -0,0 +1,218 @@
...
+
+---
+
+## 1. What changed since #1219 was written
+
+| #1219 premise | Current state |
+| --- | --- |
+| "nothing reads live state like a current-health ratio mid-tick" | **Stale.** Cull's execute bonus samples the *target's* live missing-health fraction once per fire and folds it into that fire's damage multiplier — `BattleContext.cs:316-320`, mirrored in `battle-step.ts:57-58`. |
+| "today modifiers are snapshot-fixed plus timed skill effects" | **Half stale.** Timed effects already mutate the live `AttributeCollection` mid-battle (`BattlerEffects.Apply`/`Advance`, with a `MaxHealth` re-clamp on both), so mid-battle attribute mutation is routine. What is genuinely absent is a modifier that is a *continuous function of live state*. |
+| "in-class bonus … re-couples class into the just-tuned proficiency XP accrual" | **Overstated.** The accrual is a pure `Game.Core` function taking a `ProficiencyCatalog` (#1602); its live adapter (`ProficiencyRewardService.AccrueAndApply`) already receives the `Player`, and the offline simulator already carries `ResolveClass` + `BattleSnapshot.ClassId`. The class is in hand at every call site. |
```
ginderjeremiah: **Nit — this rebuts the plumbing half of #1219's worry, not the tuning half.**

Everything asserted here checks out: `ProficiencyAccrual.Accrue` takes a `ProficiencyCatalog` and resolver funcs with no repo dependency, `ProficiencyRewardService.cs:33` does take `Player player`, and `OfflineSimulationParameters.cs:106` + `BattleSnapshot.cs:30` do carry the class into the offline path. The class genuinely is in hand at every call site.

But #1219's full sentence is *"re-couples class into the just-tuned proficiency XP accrual / **attribute pipelines**"* — which reads as a balance concern (a new variable in a curve that was just tuned), not a wiring one. "The class is available at the call sites" answers a question the issue wasn't really asking, and §2's own honest caveat two paragraphs down — in-class levels raise `playerRating`, which shrinks every path's claim through `ratingDenominator` — is a small concession that *some* re-coupling is real.

Not a disagreement with the verdict; the cost genuinely is low. Just soften **Overstated** to something like "overstated on the plumbing; see the denominator caveat below for the residual tuning coupling", so the table doesn't read as dismissing a concern it only half-addresses.

---

docs/spikes/1219-class-system-v2.md
```diff
@@ -0,0 +1,218 @@
...
+  and any out-of-class *tax* (fights the multi-character design). Nothing found here reopens those.
+
+---
+
+## 5. Overlap with #1331
+
+**Rage (this issue) and last-stand (#1331) are the offense and defence halves of one mechanic** — both read the
+player's own live health fraction. The in-review spike [PR #2401](https://github.com/ginderjeremiah/GameServer/pull/2401)
+reaches the same conclusion from the other side (it cites #1219's now-stale "nothing reads live state" line as
+grounds to revise this issue's cost estimate down).
```
ginderjeremiah: **Nit — cite the sibling doc, not the PR** (same at line 145 and line 207).

"The in-review spike PR #2401" is accurate today and goes stale the moment it merges — a reader landing on a merged PR has to work back to the artifact. #2401 adds `docs/spikes/1331-defensive-mechanics.md`, so `./1331-defensive-mechanics.md` is the durable reference and matches how the rest of this doc's front-matter links its peers (`./1126-class-system.md`, `./1526-combat-rating-power-measure.md`).

Keep the PR link alongside it if the *in-review* status is the point being made — the two spikes landing together is genuinely useful context — but the doc path should be the primary.

---

ginderjeremiah (commented):
## Review: ideation spike for #1219 — class system V2 (in-class progression bonus + conditional "stance" passives)

**Verdict:** Approve (posted as `COMMENT` — GitHub won't take a gating verdict from the PR author) — no blocking findings. Two should-fixes and two nits inline. Docs-only, and per the repo rule this is a **spike result, so it is not merged on a review pass** — it also self-declares as awaiting owner input on Q1–Q6 before implementation issues are split out, which is correct under `CLAUDE.md`'s "give your collaborator a chance to offer their input after your research" rule.

I re-checked every code citation against the current tree rather than trusting the description's checklist, and chased the issue-state claims separately. The code work is accurate throughout; one issue-state claim is stale and it's the one that changes an answer.

### Verification

**Every code citation resolves, several exactly.** `BattleContext.cs:316-320` is precisely the `targetMaxHealth`/`missingHpFraction`/`executeInvestment`/`executeMultiplier` block; `battle-step.ts:57-58` is precisely its mirror; `BattleContext.cs:355` is precisely *"channels — no health mutation, no parity surface."* Also confirmed: `ProficiencyXpCalculator.cs:6-10`'s "no shared pie to split" (and its independent restatement at `game-design.md:150`, which is what makes the bonus-vs-tax finding load-bearing rather than a single-source claim), `ProficiencyRewardService.cs:33`'s `Player` parameter, `OfflineSimulationParameters.cs:106` + `BattleSnapshot.cs:30`, `ClassSignaturePassive.GetModifier` ↔ `classSignaturePassiveModifier` as a bit-for-bit mirror, `ActivityKeys.ForDamageKey` / `DamageTypes.Applies`, `Path.ActivityKey`, `BattlerEffects`' shared per-attribute expiry + MaxHealth re-clamp, and `tracksMomentum`. Content state is as claimed: `paths.json` and `proficiencies.json` are literally `[]`, `classes.json` holds only `Adventurer`.

**The `Classify` exhaustiveness claim is stronger than usual for a "there's a test" claim.** `CombatRating.cs:90-92` documents the contract, the `_ =>` arm throws `ArgumentOutOfRangeException`, and `CombatRatingTests.Classify_IsDefinedForEveryAttribute` drives it over `Enum.GetValues`. A new attribute really does fail CI until it's priced.

**§2's central finding holds and is correctly derived.** `Split` computes `pie × activity ÷ ratingDenominator` per path with no cross-path term, so a per-path multiplier mechanically cannot tax another path. The `ratingDenominator = max(playerRating, enemyRating)` caveat is real and correctly characterised as second-order and shared with every power gain — flagging it before tuning rediscovers it is the right instinct.

**§2's kit-derivation finding is the substantive contribution and it checks out.** `BuildActivities` (`ProficiencyAccrual.cs:167-201`) is the proof: `TypedDamageDealt` folds through `ActivityKeys.ForDamageKey` (statically derivable from a skill's `damagePortions`), but `Crit`/`Dodge`/`Heal`/`Reflect`/`Hex`/`Momentum`/`Sunder`/`Cull`/`Parry`/`Cadence` all come from `AddEvent(key, stats.…)` — booked from what happened. `AddEvent(EActivityKey.Parry, stats.PlayerCounterDamageDealt)` is the exact counterexample the doc names: nothing about a stance skill statically granting `ParryChance` says "trains Riposte". Concluding that #1219's preferred kit-derived option doesn't work, rather than assuming the issue got it right, is the finding this spike exists to produce.

**§6's roster conflict is real and correctly attributed.** #1226 does specify Warrior/Mage/Ranger in its own scope section; `content-design.md:207-210` specifies Swordsman/Bowman/Wizard/Knight; #1127 carries an explicit unanswered comment asking which one stands, and #1492's two triage notes independently record being blocked on the same question. "Purely a call for the owner: both rosters are coherent, they just disagree" is a fair summary.

**§3's re-ranking is well-evidenced.** `ExecuteBonus` is confirmed base-`0` authored-only (`Attribute.cs:166-168`), Momentum is confirmed implemented (`BattlerEffects.cs:70-103` + `content-design.md`'s "committed, implemented archetypes" roster), and #336 does carry "hit-count durations & every-Nth-hit triggers" as a deferred skill-effect item. So all three of #1219's examples do decompose as claimed.

### Should-fix

- **#1455 is closed and shipped — the §4 lever is unblocked today** — line 161-163, repeated in Q6 (line 216). The statistic, challenge type, recorder, breakdown axis, admin validation and Workbench UI are all in the tree. This is the one stale claim in the doc and it's load-bearing: it turns Q6 from a sequencing question with a prerequisite into "this one is available right now and the other two aren't."
- **"Reuse Cull's treatment" doesn't transfer to a self-health channel** — line 147-150. `RefMissingHealthFraction = 0.5` is derived from *the entity that dies* depleting full-to-empty over the reference fight; the rated player doesn't. The constant needs re-deriving, and whatever value it takes is correlated with the survivability term the rating already computes — so a flat constant misprices rage across the tank/glass-cannon axis, and that misvaluation feeds `ratingDenominator` back into accrual. This is exactly the shared decision §5 argues the channel family owes, so it should be named there.

### Nits

- Line 37 — **Overstated** rebuts the plumbing half of #1219's re-coupling worry; §2's own denominator caveat concedes the tuning half.
- Lines 145 / 177 / 207 — cite `./1331-defensive-mechanics.md` rather than PR #2401, which goes stale on merge.

### Test coverage

Docs-only, no code paths touched — nothing to unit/integration test, and CI's changed-path gate correctly skips the build/test jobs. Re-reading every citation against the tree is the right substitute for a test plan on a spike, and it's why most of this review could confirm rather than re-derive. The gap worth noting for next time is that the checklist verified *code* citations rigorously and *issue state* not at all — which is where the one real error is.

### What's good

- **Reaching the opposite conclusion from the sibling spike on shared evidence.** Both PRs independently found #1219's "nothing reads live state mid-tick" stale via Cull, and they converge on rage and last-stand being one mechanic — but this doc uses it to argue direction 2 mostly *dissolves*, where #2401 uses it to argue last-stand is *cheap*. Two spikes agreeing on the fact and disagreeing on what follows is a good sign neither is pattern-matching the other.
- **The §3 table is the strongest section.** Decomposing all three of the issue's examples and finding that one is already authorable (Momentum), one belongs to an existing tracked extension (#336), and one is a small non-class-specific channel is a genuinely deflationary result — and the observation that an item or skill could grant `RageBonus` just as well as a class passive is what makes "direction 2 isn't really a class feature" land.
- **The cost taxonomy earns its place, especially avoiding shape 4.** "Both simulators must toggle on the *same tick*, and a toggle interacts with `BattlerEffects`' shared-expiry stacking and the `MaxHealth` re-clamp" is a specific, checkable reason to prefer sampled scalars over threshold stances — the sort of constraint much cheaper to record now than to discover during parity debugging.
- **§4 goes past the brief the way an ideation spike should.** Class-tailored challenges, the one-modifier cap on `ClassSignaturePassive`, and word-of-power decipher depth are all outside the issue's two directions, and the first of them plausibly outranks both. Explicitly restating what stays rejected (#982 decision 2, #1124, out-of-class tax) is the right bookend — it says the exploration was bounded on purpose rather than by omission.
- **Updating #1126's forward-reference in the same PR** so the stale cost estimate isn't left sitting in two places, and flagging that #1219 itself should be amended when implementation issues get split out.

---

ginderjeremiah:
All four addressed in d75594f, plus one more stale claim the same pass turned up. Not merging — per the repo rule a spike result isn't merged on a review pass, and Q1–Q6 are still open.

**Both should-fixes were right, and I re-verified each against the tree rather than taking them on faith.**

**#1455 is shipped** — closed `completed` on 2026-07-02, with `EStatisticType.KillsByDamageType = 22` and `EChallengeType.KillsByDamageType = 9` both live. §4 now says the tailored-challenge lever is unblocked *today*, and Q6 is reframed the way you suggested: not "which do we sequence first" but "the routes can start now, both V2 directions can't start until Q5 is answered." That does sharpen the ask considerably.

**The `RefMissingHealthFraction` point is the more valuable of the two** and I'd understated it — "reuse its treatment" was hiding the open question rather than resolving it. `ServerGameConstants.cs:122-129` justifies `0.5` by "health depletes from full to empty over the reference fight," which is load-bearing on the sampled entity being the one that dies. §3 now separates the structure (reuses — `AsymmetryGated` + constant × attribute) from the constant (doesn't), and states the build-correlation consequence: because the player's average missing health is an inverse function of the survivability term `Rate = √(Offense × Survivability)` already computes, a flat constant over-values rage on tanky builds and under-values it on fragile ones, and that misvaluation feeds `ratingDenominator` back into accrual. §5 now lists three named deliverables for the shared design pass — sampling (Cull's template *does* transfer), pricing (it doesn't), tally + activity key — so the pricing call can't get inherited by precedent from whichever half lands first.

**Nits:** the accrual row now reads "overstated on the plumbing; partly conceded on the tuning" and points at §2's denominator caveat — you're right that #1219's sentence names the attribute pipelines and is a balance worry, not a wiring one. All three #2401 citations now lead with `./1331-defensive-mechanics.md` and keep the PR link alongside, since the two spikes landing together is context worth preserving.

**One more from taking your "double-check the rest of §4's issue-state claims" note literally:** **#1124 (the proficiency gear-gate) is also closed and shipped** (2026-06-27). §4 cited it as the soft alternative to class-gated gear without noticing it had landed — so "heavy armour requires the martial proficiency" is authorable today rather than pending, which strengthens that bullet the same way #1455 strengthens the one above it. I also corrected the now-stale "⚠ assumes a statistic that does not exist yet" warning on the §5 routes in `content-design.md` — a one-line fix directly adjacent to the finding, so it's in this PR rather than a follow-up. #336, #1428 and #982 checked clean.

Fair diagnosis on the root cause, incidentally: the checklist verified every *code* citation and no *issue state*. Both errors were the same class, and both were in §4.

---

ginderjeremiah:
## Re-review of `d75594f` — both should-fixes addressed, plus a third stale claim caught on the same pass

**Verdict:** Approve, **not merging** — spike results aren't merged on a review pass, and this one is still waiting on Q1–Q6 per `CLAUDE.md`'s ideation rule. CI green (`all-checks-passed`; changed-path gate correctly skips build/test).

Incremental pass over the delta only. I verified the new claims against the tree and against GitHub rather than taking them on trust — including the two shipped-on dates, since date claims are the kind of thing that reads as verified and usually isn't.

### Should-fixes

**1. #1455 is shipped — corrected, and the consequence is drawn out properly.** Confirmed closed as completed **2026-07-02**, exactly as stated, and the implementation is in the tree: `KillsByDamageType` appears across `Enums.cs`, `StatisticType.cs`, `ChallengeType.cs`, `PlayerProgress.cs`, `ProgressionGraphChecker.cs`, `AdminChallenges.cs`, and on the frontend through `challenge-helpers.ts`, `challenge-meta.ts`, `TypeGlyph.svelte` and `statistics-view.svelte.ts` — recorder, admin validation and Workbench UI included, so "pure Workbench authoring, available now" is accurate.

The reframing of **Q6** is what makes the fix worth more than the correction. It was a sequencing question between three gated options; it's now "the routes can start today, both V2 directions cannot until Q5 is answered," which is a materially different question and one the owner can actually answer. Catching that `content-design.md` §5 still carried the stale ⚠ warning and fixing it in the same change is the right thoroughness — that warning would have kept the routes looking blocked to anyone reading the content plan.

**2. `RefMissingHealthFraction` doesn't transfer.** Confirmed verbatim at `ServerGameConstants.cs:129`, with the doc comment at `:122-128` giving exactly the justification cited — *"health depletes from full to empty over the reference fight, so its time-average missing fraction is one half"* — which is the premise that fails for a self-health channel. `Math.Sqrt(offense * survivability)` confirmed at `CombatRating.cs:72`, so the build-correlation argument (a flat constant over-values rage on tanky builds, under-values it on fragile ones, and that misvaluation lands back on accrual via `ratingDenominator`) rests on the real formula.

Splitting §5 into three named deliverables — sampling (transfers), pricing (does not), tally + key — is a better structure than the prose it replaced, and calling the pricing constant "the decision most at risk of being made silently by whichever half ships first" is the specific warning that was missing.

**3. Bonus: #1124 also shipped.** Not something I raised — found on the same pass. Confirmed closed **2026-06-27**, and `RequiredProficiency` is in `Game.Core/Items/Item.cs`, the contract, the mapper, `AdminItems`, `Inventory` and the EF baseline. So §4's soft alternative to class-gated gear really is live rather than hypothetical. Checking the *other* issue-state claims after being caught out on one is the right response to the finding, and it's what turned up this one.

### Nits

Both addressed. The `#2401` citations now lead with the durable `1331-defensive-mechanics.md` path and mention the PR only as current status, so they degrade gracefully on merge; the accrual "Overstated" row now concedes the tuning half. I checked the two new intra-doc anchors resolve — `#5-overlap-with-1331` and `#two-constraints-any-new-channel-must-satisfy` both match their headings exactly.

### Two new nits (trivial, neither worth a commit on its own)

- §4's parenthetical *"…and is corrected in this change"* narrates the edit. `CLAUDE.md`'s doc guidance puts "documenting specifically that you changed something" in the AVOID list — stating that #1455 shipped and the routes are authorable is the durable half; the rest goes stale the moment this merges.
- §5 item 2 cites the constraints subsection as "§3", but it's a `###` inside §3 rather than §3 itself. The anchor works, so this is purely a label.

### Standing

Unchanged: docs-only, no code paths touched, correct as it stands. The pattern worth keeping from this round is the one the first review's coverage note pointed at — code citations were verified rigorously from the start and issue state wasn't, which is where both real errors were. Both are now fixed and the third was found by applying the same check to the rest.

Q1–Q6 remain the ask; Q5 (the roster) still gates both V2 directions and, per §6, #1127 / #1226 / #1492 as well.

---

ginderjeremiah:
Both nits fixed in 878333f — the §4 parenthetical no longer narrates the edit (you're right that it's on `CLAUDE.md`'s AVOID list and goes stale on merge), and §5 item 2 names the constraints heading instead of mislabelling it "§3".

They rode along with a base-branch update rather than a commit of their own: `main` had moved three commits ahead (#2398, #2403, #2406), so the branch is merged up to date — no conflicts. Re-verified the load-bearing content claims survive that merge, since #2403/#2406 both touch the progression surface: `content/paths.json` and `content/proficiencies.json` are still empty and `classes.json` still holds only `Adventurer`, so direction 1's content gate is unchanged.

CI green on the new head. Still not merging — Q1–Q6 remain the ask, Q5 (the roster) most of all.

---

ginderjeremiah:
## Re-review of `878333f` — both nits taken, nothing new

**Verdict:** Approve, **not merging** — spike result, and Q1–Q6 are still the ask. CI green (`all-checks-passed`; changed-path gate correctly skips build/test). No new findings.

Incremental pass over the delta only (seven lines in `docs/spikes/1219-class-system-v2.md`, plus the base-branch merge).

### Nits

**1. Edit narration dropped.** §4's `#1455` bullet now ends on the durable claim — the tailored-challenge routes are authorable now and wait on nothing — with the parenthetical about correcting `content-design.md` gone. Right call under `CLAUDE.md`'s doc guidance, and I checked the correction itself survived the removal: `content-design.md:224` still reads *"the kills-by-damage-type statistic (#1455, shipped), so these routes are authorable in the Workbench today."* The narration went; the fix stayed, which is the outcome that guidance is aiming at.

**2. §5 item 2's label.** Now names the heading rather than mislabelling it "§3", and the anchor still resolves — `### Two constraints any new channel must satisfy` at line 138 matches `#two-constraints-any-new-channel-must-satisfy` exactly.

### The base-branch merge

Clean, and the re-verification was worth doing rather than asserting — #2403 and #2406 both touch the progression surface. Confirmed independently: the branch's diff against `main` is exactly the three intended files (`content-design.md`, `1126-class-system.md`, and the new spike doc), with no unrelated drift pulled in, and direction 1's content gate is unchanged.

### Cross-PR note

#2401 has since taken the pairing this doc's §5 identified and recorded it from the defensive side (`3c0b74c`), including the distinction that Cull samples the **target's** health while rage and last-stand both sample the **battler's own**. The two docs now agree on the three decisions the shared channel family owes — sampling, pricing, tally — which is the state you'd want before either implementation issue gets cut. The citation asymmetry runs the other way now: this doc's `#2401` references lead with `./1331-defensive-mechanics.md`, and the sibling's don't. Flagged as a nit over there, not here.

### Standing

Unchanged: docs-only, no code paths touched, correct as it stands. Q1–Q6 remain the ask, Q5 (the roster) most of all — it gates both V2 directions and, per §6, #1127 / #1226 / #1492. No implementation issues split out yet, correctly, and none filed from this review.

---

ginderjeremiah:
## Status pass on `878333f` — no change since the last re-review; parked on owner input

**Verdict:** Approve (standing), **not merging** — spike result, and Q1–Q6 remain unanswered. CI green (`all-checks-passed`; changed-path gate correctly skips build/test). No new findings.

Nothing has changed on the branch since the `878333f` re-review, so this isn't a re-review of a delta — there isn't one. Rather than restate the standing verdict, I spent the pass on the one thing that *can* change under a parked PR: whether `main` moving out from under it has invalidated anything.

### Re-verified against current `main`

`main` has advanced four commits past this branch's merge base (`39bb1c7`) — #2411, #2413, #2421 and #2420. All four are frontend socket lifecycle or test-scaffolding consolidation; the diff contains no non-test, non-socket files at all, so nothing in the progression, class, proficiency or combat-rating surface this spike reasons about has moved.

Spot-checked the load-bearing content gate directly against `origin/main` rather than inferring it from the file list, since it's the claim direction 1 rests on: `content/paths.json` and `content/proficiencies.json` are still literally `[]`, and `content/classes.json` still holds only the `Adventurer` placeholder. Direction 1's content gate is unchanged, and §2's `ProficiencyXpCalculator` / `BattleContext` citations are untouched by anything that landed.

So the doc is still accurate as it stands, and the branch does **not** need a base update to stay correct — merging `main` in now would be churn for a docs-only change with no conflict.

### Why this isn't something I can push forward

The blocker is owner discretion, not work. Q1–Q6 are design calls, and **Q5 (the roster) is the one that gates the rest** — #1219's examples name Warrior/Mage/Ranger while `content-design.md` §5 and the shipped #1457/#1492 code name Swordsman/Bowman/Wizard/Knight. Both rosters are internally coherent; they just disagree, and per §6 that same unanswered question is already blocking #1127, #1226 and #1492. Nothing in the doc can be improved by guessing at it, and per `CLAUDE.md`'s ideation rule the implementation issues correctly aren't split out until there's a read.

### Merge coordination

Unchanged and still worth restating: #2401 and this PR should **land together**. Each doc now cites the other by relative path (`./1331-defensive-mechanics.md` here, `./1219-class-system-v2.md` there, both fixed as of #2401's `3392e5d`), so each half carries a dangling link until the other merges. They're blocked on the same owner answer, so that costs nothing — and the substantive reason from §5/§6 still holds: if they land separately, the shared own-missing-health term gets specified twice.

No follow-up issues filed from this pass, and none needed — everything outstanding is a question, not a task.

---

ginderjeremiah:
## Re-review of `b6b79e3` — the #2416 precedent holds; I fixed the one problem with it rather than send it back

**Verdict:** Approve, **not merging** — spike result, and Q1–Q6 are still the ask. One fix pushed as `5f88ded`; details below.

Incremental pass over the delta only (§3's rating-feedback paragraph).

### The precedent checks out

The substance is right, and it's a genuine strengthening. #2416 describes an unclamped `ParryChance × ParryChanceMultiplier` / `CriticalChance × CriticalChanceMultiplier` product pricing capability past what the engine can deliver, inflating `OffenseRate`; it then traces that inflation through `min(EnemyRating ÷ PlayerRating, 1)²` and the `max(PlayerRating, EnemyRating)` proficiency denominator to *"undercut[ing] that player's XP and proficiency gains."* That is the same path §3 argues a mispriced `RefMissingHealthFraction` analogue would take, so the claim moves from inference to observed precedent. "From the other direction" is fair: #2416 over-prices uniformly, whereas a flat self-health constant misprices with sign depending on the build.

Confirmed the issue is closed as completed (2026-07-25), so citing it as settled precedent rather than a proposal is correct.

### What I fixed (`5f88ded`)

The quotation was stitched. `"inflat[ed] `OffenseRate` and undercut[] that player's XP and proficiency gains"` reads as one sentence lifted from #2416, but it's two fragments from separate paragraphs — *"a chance product above 1 inflates `OffenseRate` beyond any achievable output"* under **Problem**, and *"undercuts that player's XP and proficiency gains"* under **Effect** — joined by an "and" that isn't in the source. Someone checking the citation by searching #2416 for that string finds nothing, which is the specific failure mode this doc has spent three rounds fixing in the other direction.

Replaced with a paraphrase that names the mechanism the issue actually supplies, including the `max(PlayerRating, EnemyRating)` denominator the inflation travels through — which is also the concrete link back to `ratingDenominator` that the stitched quote was gesturing at. Substance unchanged.

Pushed rather than raised: it's a two-line docs edit on a parked doc, and sending it back for a round-trip would cost more than it's worth — see the cadence note below.

### Verification against a `main` you hadn't seen

`main` has advanced two commits past this branch's base — #2430 (auth rate-limit budget) and #2431 (socket handshake teardown). Neither touches anything this doc cites; between them they change `api-socket.ts` and its test, `PlayersController`, the rate-limiting wiring, and three docs.

Checked this doc's line citations against current `main` directly rather than inferring from the file list: `BattleContext.cs:316-320` still opens on `var targetMaxHealth = _targetBattler.GetAttributeValue(MaxHealth)`, `:355` is still the *"channels — no health mutation, no parity surface"* comment, `ProficiencyXpCalculator.cs:6-10` still carries the "no shared pie" contract, and `ServerGameConstants.cs:122-129` still holds the `RefMissingHealthFraction` derivation §3 rests on. No rot. As you noted on the sibling, this doc has no citations into `CombatRating.cs`, so #2428 couldn't have touched it — and the new paragraph cites #2416 by issue number rather than by line, which is the right call for exactly that reason.

The content gate is also unchanged: `content/paths.json` and `content/proficiencies.json` are still `[]` and `classes.json` still holds only `Adventurer`, so direction 1's gate stands.

### Standing, and a recommendation on review cadence

Unchanged: docs-only, no code paths touched, correct as it stands. Q1–Q6 remain the ask, Q5 (the roster) most of all, and per §6 that same question is blocking #1127 / #1226 / #1492. No implementation issues split out, correctly, and none filed from this pass.

Same recommendation as on #2401: **no further review passes until the questions get a read.** This is the fifth, and the last three resolved a label, an edit narration, and now a quotation's provenance. All correct, none of them what the PR is waiting on. The substantive review closed at `d75594f`; since then both PRs have been in a loop where reviewing generates nits, fixing them generates commits, and commits invite another review — while the actual blocker, a design decision, doesn't move. Both docs are in a good state to be read.

Merge coordination unchanged: this and #2401 should land together, since each cites the other by relative path.

---

ginderjeremiah:
## Status pass on `5f88ded` — #2438 rotted this doc too; fixed and pushed as `34c7b55`

**Verdict:** Approve, **not merging** — spike result, Q1–Q6 remain the ask. No new findings on the doc's substance.

Nothing had changed on the branch since the last pass, so there was no delta to review. What I did instead was apply the sibling's #2438 audit to this doc — [#2401 found](https://github.com/ginderjeremiah/GameServer/pull/2401#issuecomment-5080925640) that commit shifted three files by three different offsets and invalidated all sixteen of its citations at once. This doc has two citations into one of those files, and both had rotted.

### What was wrong

**§1's Cull citation — the load-bearing one.** `BattleContext.cs:316-320` was cited as the missing-health sample proving the engine already reads live state mid-tick. That's the finding this whole doc turns on: it re-ranks direction 2, marks #1219's own *"nothing reads live state like a current-health ratio mid-tick"* stale, and is the shared evidence the sibling spike independently reached. #2438 shifted `BattleContext.cs` by −2, so `:316` now lands mid-expression inside the `Math.Clamp` continuation and `:320` past the block entirely. The sample is at **`:314-318`**.

**§2's parity citation.** `BattleContext.cs:355` cited *"backend-only side channels — no health mutation, no parity surface"* — the evidence for direction 1 having zero parity surface, which is half of why it's ranked cheaper than direction 2. `:355` is now `totalBooked += overlayBasis;`. The comment is at **`:353`**.

### What was fine

Verified the remaining three by matching the expected construct on the cited line against `origin/main`, not by assuming the −2 held:

- `ProficiencyXpCalculator.cs:6-10` — still the *"no shared pie to split"* contract §2's central finding rests on.
- `ServerGameConstants.cs:122-129` — still the `RefMissingHealthFraction = 0.5` derivation, verbatim.
- `battle-step.ts:57-58` — still `missingHpFraction` / `executeMultiplier`. Worth noting this one **did not move** even though `battler.ts` shifted +21 in the same commit, which is a concrete case of why the per-file offset can't be extrapolated.

### Why I pushed rather than flagged

This clears the bar the last two passes set for acting on a parked PR: the rot is in the load-bearing citations, not incidental drift. §1's is the doc's headline evidence, and a reader following it now lands mid-expression and the finding reads as unsupported. Two lines on a docs-only change the changed-path gate skips.

I also added the citation re-check to §7 as a standing instruction for the implementation-split pass, folded in with the existing amend-#1219 note, pointing at [`1331-defensive-mechanics.md`](https://github.com/ginderjeremiah/GameServer/blob/claude/sleepy-heisenberg-5661tn/docs/spikes/1331-defensive-mechanics.md) §8 for the list of movers rather than duplicating it — the pair has been careful not to specify shared things twice, and this is one of them.

### Also re-verified

`main` has advanced past this branch's base by #2430, #2431, #2428, #2437, #2435, #2438, #2442 and #2443. Only #2438 touched anything this doc cites. Direction 1's content gate is unchanged: `content/paths.json` and `content/proficiencies.json` are still literally `[]` on `main`, and `classes.json` still holds only the `Adventurer` placeholder.

### Standing

Unchanged: docs-only, no code paths touched, correct as it stands. Q1–Q6 remain the ask, **Q5 (the roster) most of all** — it gates both V2 directions and, per §6, #1127 / #1226 / #1492. No implementation issues split out, correctly, and none filed from this pass.

Cadence recommendation unchanged and now applies to both halves of the pair: **no further review passes until the questions get a read** — citation maintenance is the only category worth a commit while parked, and both docs now carry a standing instruction to do that audit in the split pass rather than continuously. Merge coordination unchanged: these two should land together, since each cites the other by relative path.

---

ginderjeremiah:
## Owner read on Q1–Q6 — decisions, and merging

Merging together with #2401 per its dangling-link note; #1219 stays open until the split. The calls:

**Q1 — accepted, direction 2 dissolves.** Rage joins #1331's last-stand as one own-health channel family, designed in a single shared pass (sampling site, `CombatRating` pricing, tally/path) — the `RefMissingHealthFraction` re-derivation is a named deliverable of that pass, not inherited from Cull. Escalation is Momentum content today. First strike belongs to #336. No class-specific runtime work; the class half is authoring.

**Q2 — class choice is the commitment.** It's exclusive per character, and under the multi-character design that is a real opportunity cost, so a signature passive granting an archetype enabler qualifies as committed. One clarifying sentence in `content-design.md` §4 when the implementation issues land — no formal carve-out. The shield half of the question is moot: deferred per #2401 Q5, and if ever built it's raw EHP and ships path-less exactly as the rule already says.

**Q3 — authored bias list.** `InClassPathIds` on `Class`, a Workbench section, and a warning-level Content Health lint against kit drift. The parry-stance counterexample is conclusive, and a skill-effect→activity-key map would be a second source of truth for a binding `Path.ActivityKey` already owns.

**Q4 — stays gated.** The seam is one multiplier with zero parity surface; there is nothing to de-risk by landing a strawman early, and dead config is all it would be.

**Q5 — Swordsman / Bowman / Wizard / Knight.** `content-design.md` §5 built the full routes around that roster and the shipped #1457/#1492 code already names it; Warrior/Mage/Ranger was #1126 §H's strawman and stays one. This is the answer #1127, #1226, and #1492 are blocked on — amend #1219's and #1226's examples at split time.

**Q6 — the routes get the next class-identity slice.** Pure Workbench authoring, unblocked now that #1455 shipped, and it delivers felt class identity sooner than either V2 direction — both of which queue behind the content work Q5 just unblocked regardless.

Next step is the split: the shared channel-family issue with #1331 first, then the direction-1 pieces behind their content gate, amending #1219's stale live-state line and its roster examples in the same pass.

---

PR #2403, opened 2026-07-25
Bring the progression editor's field lengths under the EF-mirror guard (#2377)

### LINKED ISSUE #2377: Progression editor loose ends: path/proficiency maxLengths escape the EF-mirror guard, and IProficiency.description is authored-but-unusable
## Context

Adjacent to #2318/#2325 (the EF-mirrored maxLength guard) and found in the same health review as the Workbench warn-gap issue.

## Problems

1. **The bespoke Progression surface's maxLengths are un-guarded literals.** `progression/PathDetail.svelte:32,49,59` and `ConlangIdentity.svelte` hand-copy `maxLength` literals whose EF counterparts are also raw literals (`GameContext.cs:620-660`, not `ContentFieldLengths` members), and `field-lengths.test.ts` only walks `workbenchEntities` — so none of the path/proficiency text bounds are CI-guarded. They match today; this is exactly the silent-drift class #2325 closed for the generic Workbench, still open for Paths. Fix: move the EF bounds onto `ContentFieldLengths` and extend the guard test to the progression inputs.
2. **`IProficiency.description` can be neither written nor read.** It round-trips through `profIdentityDto` (`progression-helpers.ts:264`) and is EF-bounded at 500 (`GameContext.cs:643`), but no editor input exists in the tier editor and no player-facing surface consumes it. Either add it to `ConlangIdentity` (paths have a description editor; tiers arguably deserve one) or retire the column — currently it's dead weight that suggests authorability it doesn't have.

ginderjeremiah:
Closes #2377.

Both loose ends from the issue. Values are unchanged throughout, so **no migration** — this is regression-prevention plus one missing editor input.

## 1. Path/proficiency maxLengths now mirror the EF bound

`#2325` closed the silent-drift class for the generic Workbench, but the bespoke Progression surface escaped it on both sides: `PathDetail`/`ConlangIdentity` hand-copied literals whose EF counterparts (`GameContext.cs`) were *also* raw literals, and `field-lengths.test.ts` only walks `workbenchEntities`.

- 10 new `Path*`/`Proficiency*` constants on `ContentFieldLengths`; `GameContext` now configures those columns against them. The existing codegen mirrors them into `game-constants.ts` — no new writer, and CI's codegen-drift check already covers it.
- The progression markup reads those constants **directly** rather than re-declaring the value. This deliberately differs from the entity configs, which keep a literal cross-checked by a test: an `EntityConfig` is a data object where the literal is the readable thing, whereas this markup is hand-written, so there is simply no second literal to keep in step. Drift stops being *detected* and becomes *impossible*.

That leaves two failure modes an import can't catch, so both get a guard:

- **Unbounded input** → `ProgInput.maxLength` is now a **required** prop, so `svelte-check` fails on a new input missing a bound, in any file — not just the two that exist today.
- **Wrong constant wired** → a new `progression/field-lengths.test.ts` renders each identity surface and asserts its whole rendered field set by exact equality, so a mis-wired bound, an input registered against no constant, and a removed one all surface as a diff.

## 2. `IProficiency.description` is now authorable

It round-tripped through `profIdentityDto` and was EF-bounded at 500, but had no editor input, so it could never be written.

**I kept the column rather than retiring it**, on the evidence that it is player-facing content rather than dead weight: it's `required` on the Core domain model (`Game.Core/Proficiencies/Proficiency.cs:14`, which `DesignerNotes` is deliberately *not* on), and it ships **unredacted to every client** — `AbstractReferenceDataCommand.RedactAuthoringOnlyFields` blanks `DesignerNotes` for non-admin connections but not this. Retiring it would also mean a migration plus Core/Abstractions/Infrastructure/mapper/repo churn, well past this issue's `scope: small`.

Added as an optional (no `warn`) textarea after the identity row, mirroring `PathDetail`'s name → description → notes ordering. The store needed no change — `canonicalEqual` already diffs the whole record, so an edit marks the tier modified.

**Flagging the half this doesn't close:** the *read* side is still open. `description` reaches every client but no player-facing surface renders it. The natural home is the Lexicon's `WordDetail` inspector, but that's a densely designed pane and where a blurb goes in it is a content/UI design call, not a mechanical addition — so I left it rather than making that call unilaterally. Happy to file a follow-up or take it on if you'd like it surfaced there.

## Test plan

- [x] `dotnet test` — 3490 passed, 0 failed (Core 1442, Infrastructure 74, Api 873, Application 1103)
- [x] `npm run lint` (eslint + svelte-check) — 1220 files, 0 errors, 0 warnings
- [x] `npm run test:unit:ci` — 4003 passed across 312 files
- [x] `dotnet run --project Game.Api -- codegen` — regenerated and confirmed idempotent on a second run, matching CI's drift check
- [x] Confirmed the EF migration snapshot is **untouched** — the literals became named constants of identical value
- [x] **Mutation-tested both new guards, then reverted each**: wiring `Translation` to `PROFICIENCY_NAME_MAX_LENGTH` fails the new test with a clear `100 → 50` diff; deleting a `maxLength` prop fails `svelte-check` with `Property 'maxLength' is missing … but required in type 'Props'`
- [x] New `TierDetail` test drives the Description input and asserts the patch actually sets `description`

## Note for review

The generic Workbench's 34 entity-config literals could adopt the same direct-import approach, which would retire half of `entities/field-lengths.test.ts`'s job. I left them alone — that pattern is CI-guarded and correct, so it's a consistency question rather than a defect, and converting them is a large diff unrelated to this issue. Worth a follow-up only if you want the subsystem uniform.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HnZvfdyMqQ4oLFrw5yh8fi)_


UI/src/tests/routes/admin/workbench/progression/field-lengths.test.ts
```diff
@@ -0,0 +1,132 @@
...
+/* The progression editor's half of the maxLength guard (#2377). `field-lengths.test.ts` under
+   `entities/` walks `workbenchEntities`, so it never saw the bespoke Progression surface — its path and
+   proficiency bounds were hand-copied literals mirroring equally hand-copied EF literals, the exact
+   silent-drift class #2325 closed for the generic Workbench.
+
+   Both sides now read `ContentFieldLengths` (via the generated `game-constants.ts`), so a *drifted* value
+   is structurally impossible and `ProgInput.maxLength` is required so an *unbounded* input fails
+   svelte-check. What neither catches is a field wired to the wrong constant, so this asserts the whole
+   rendered field set of each surface by exact equality: a mis-wired bound, a new input registered against
+   no constant, and a removed one all surface as a diff rather than passing silently. */
```
ginderjeremiah: **Nit** — "a mis-wired bound ... surface[s] as a diff" is only true when the two constants hold *different* values, and five of the ten here are `50`: `PATH_NAME`, `PROFICIENCY_NAME`, `PROFICIENCY_ICON_PATH`, `PROFICIENCY_WORD`, `PROFICIENCY_PRONUNCIATION`. So the mutation the description cites (`Translation` → `PROFICIENCY_NAME_MAX_LENGTH`, 100 → 50) fails loudly, but swapping `Name` ↔ `Icon path` ↔ `Romanized word` ↔ `Pronunciation` passes silently.

That's inherent to asserting rendered values rather than constant identity — nothing to fix in the assertions. But since this comment is specifically making a guarantee, it's worth scoping it honestly, e.g. "a mis-wired bound *whose constant differs in value*". The two things the exact-equality genuinely does catch unconditionally — a new input registered against no constant, and a removed one — are the stronger half of the claim and survive the rewording.

---

UI/src/tests/routes/admin/workbench/progression/field-lengths.test.ts
```diff
@@ -0,0 +1,132 @@
...
+} as unknown as ProgressionStore;
+
+/** Every rendered text input/textarea keyed by its accessible label -> its `maxlength` (null when unbounded). */
+const boundsOf = (container: HTMLElement): Record<string, number | null> =>
+	Object.fromEntries(
+		[...container.querySelectorAll('input, textarea')].map((element) => {
+			const bound = element.getAttribute('maxlength');
+			return [element.getAttribute('aria-label') ?? '(unlabelled)', bound === null ? null : Number(bound)];
+		})
+	);
```
ginderjeremiah: **Nit** — the selector sweeps *every* `input`, not just text ones, so a legitimately unbounded numeric field would fail this test with a misleading `'Max level': null`. `ProgNumber` renders a real `<input>` through `NumInput` (`ProgNumber.svelte:3`), and a number input correctly has no `maxlength` — so the day someone adds an XP or level field to either identity surface, the *length* guard is what breaks, and the fix looks like "register it in the map" when the right answer is "exclude it".

Narrowing to the text-ish controls the guard is actually about would avoid the false positive:

```ts
container.querySelectorAll('textarea, input:not([type]), input[type="text"]')
```

Either that, or say in the doc block that a non-text input has to be excluded when one is added. Today neither surface renders one, so this is future-friction rather than a live problem.

---

UI/src/tests/routes/admin/workbench/progression/field-lengths.test.ts
```diff
@@ -0,0 +1,132 @@
...
+	profStatus: vi.fn(() => 'clean'),
+	isRetired: vi.fn(() => false),
+	setPathTab: vi.fn(),
+	resetPath: vi.fn(),
+	retirePath: vi.fn(),
+	removePath: vi.fn(),
+	pathBaseline: vi.fn(() => path),
+	patchPath: vi.fn(),
+	patchProf: vi.fn()
+} as unknown as ProgressionStore;
```
ginderjeremiah, line 100: **Nit (DRY)** — this is now the *third* near-identical copy of the progression test scaffolding: the hoisted `staticData`/`dangerModal` `$stores` mock, the `path`/`tier` fixtures, and the `as unknown as ProgressionStore` fake all exist in `PathDetail.test.ts:7-75` and `TierDetail.test.ts:7-76`. Two copies were pre-existing; this PR makes it three, which is about where it stops being tolerable duplication (`CLAUDE.md` → "prefer writing DRY code").

A small `tests/routes/admin/workbench/progression/progression-test-utils.ts` exporting `mockStores()`, `path(over)`, `tier(over)` and `makeStore(...)` would collapse all three — and it's the kind of thing that gets harder the longer it waits, since the three copies have already diverged (this one drops the `beforeEach` reset and the `over` params, `PathDetail`'s carries `paths` in `staticData` and `TierDetail`'s doesn't). Happy to see it as a follow-up rather than in this PR if you'd rather keep the diff tight — just flagging that the third copy is the signal.

---

Game.Api.Tests/CodeGen/ApiCodeGeneratorTests.cs
```diff
@@ -57,6 +57,9 @@ public void GetClientMirroredConstantFields_IncludesGameConstants(string fieldNa
         [Theory]
         [InlineData(nameof(ContentFieldLengths.EnemyNameMaxLength))]
         [InlineData(nameof(ContentFieldLengths.SkillDesignerNotesMaxLength))]
+        // The bespoke progression surface's bounds (#2377) mirror through the same generated file.
+        [InlineData(nameof(ContentFieldLengths.PathNameMaxLength))]
+        [InlineData(nameof(ContentFieldLengths.ProficiencyDescriptionMaxLength))]
```
ginderjeremiah: **Nit** — these two cases add no coverage. `GetClientMirroredConstantFields` walks the whole `[ClientMirrored]` class, so the existing `EnemyNameMaxLength`/`SkillDesignerNotesMaxLength` cases already prove every field of `ContentFieldLengths` is mirrored; a new `const` in that class cannot be absent unless the two incumbents are too. The only thing that would genuinely escape is a constant added to some *other*, unmarked class — which naming two of the ten new fields doesn't test either.

Harmless, but it sets a precedent where every future field gets an `InlineData` line and the theory grows without gaining anything. I'd drop both and let the class-level guarantee stand; if the intent was to document that the progression bounds ride the same generated file, the `ContentFieldLengths` XML doc (which you already updated) is the better home for that.

---

ginderjeremiah (commented):
## Review: bring the progression editor's field lengths under the EF-mirror guard, and make `IProficiency.description` authorable (#2377)

**Verdict:** Approve (posted as `COMMENT` — GitHub won't take a gating verdict from the PR author) — no blocking findings. Four nits inline. **Not merging**, per the two nits worth a quick pass and because `test-backend` is still running; everything else on the head commit is green.

I checked the description's claims against the tree rather than taking them at face value. They hold up, including the two load-bearing ones behind the "keep the column" decision.

### Verification

**Both halves of #2377 are genuinely closed, and the coverage is complete rather than partial.** Every string column on both entities is now constant-driven — `Path` has exactly `Name`/`Description`/`DesignerNotes` and `Proficiency` exactly `Name`/`Description`/`IconPath`/`Word`/`Pronunciation`/`Translation`/`DesignerNotes` (`Game.Infrastructure/Entities/*.cs`), and all ten are configured against the new constants at `GameContext.cs:625-660`. Values are identical throughout, so the untouched-migration-snapshot claim follows. `game-constants.ts` matches the codegen's alphabetical emission, so it is regenerated output, not hand-edited.

**The surface really is fully enclosed.** `ProgInput` is the only text control in `workbench/progression` — the sole other `<input>` in the folder is `ProgressionList.svelte:20`'s search box (not DB-backed), and `ProgNumber` is numeric. So making `maxLength` required covers every DB-bound text field in the subsystem, not just the ones that happen to exist in these two files.

**The "keep the column, don't retire it" call is correctly evidenced, and it's the right one.** `Description` is `required` on the Core model (`Game.Core/Proficiencies/Proficiency.cs:14`) where `DesignerNotes` deliberately isn't, and `RedactAuthoringOnlyFields` (`AbstractReferenceDataCommand.cs:40-47`) blanks only `IHasDesignerNotes.DesignerNotes` — so this genuinely ships unredacted to every client and is player-facing content by construction, not dead weight. Retiring it would also have meant a migration plus mapper/repo churn well past `scope: small`.

**The store needed no change, as claimed.** `description` is already in `profIdentityDto` (`progression-helpers.ts:264`) and `canonicalEqual` diffs the whole record (`progression-store.svelte.ts:142`), so an edit marks the tier `modified` and rides the existing identity Edit. Content export is DB-derived and `content/proficiencies.json` is empty, so the drift guard is unaffected.

**Making `maxLength` required is the better half of this PR.** Detection catches drift in the fields that exist; a required prop catches the field that doesn't exist yet, in any file, at `svelte-check` time. That's the failure mode #2325 couldn't reach, and it's a strictly stronger guard than another test.

### Nits

Four inline: the mis-wire claim is weaker than stated where constants share a value (five of the ten are `50`); `boundsOf`'s selector will false-positive on a numeric input; the test scaffolding is now duplicated a third time; and the two new `ApiCodeGeneratorTests` cases don't add coverage.

One that doesn't anchor to a line — **the guard is enumerated per-component, where the entities one is data-driven.** `entities/field-lengths.test.ts` walks `workbenchEntities`, so a new entity is covered the moment it's registered. This one names `PathDetail` and `ConlangIdentity` explicitly, so a *future* progression component is covered by neither test nor prop: `maxLength={50}` in a new file passes `svelte-check` (the required prop stops omission, not a literal) and renders nothing either test looks at. That makes `docs/frontend-admin.md`'s "drift stops being detected and becomes impossible" true of today's fields but not of the pattern. Not worth building a source-scan lint for two components — but the doc sentence is the thing to soften, since it's what a future author will read before adding the third one.

### Follow-up (out of scope here)

Filed **#2404** for the read half you flagged, so the offer in the description is captured rather than lost. Two things worth noting that widen it slightly:

- **`Path.Description` has the identical status** — authorable, delivered, rendered nowhere. #2377 only named the proficiency one, but nothing under `routes/game/screens/proficiencies` reads `.description` at all. They're the same gap and should be decided together.
- Your reasoning for not making the call unilaterally is right. `WordDetail` is the natural home but the pane is already dense, and whether a tier blurb should be decipher-gated the way word/pronunciation/translation are is a design question, not a placement one. The issue records the retire-both alternative too, so the either/or from #2377 survives intact rather than defaulting to "authored but invisible" forever.

### Test coverage

Appropriate and well-targeted. The new `TierDetail` case drives the input and asserts the patch closure actually sets `description` — testing the wiring, not just that a textarea rendered, which is what would have caught the original bug. `field-lengths.test.ts` renders `ConlangIdentity` directly rather than through `TierDetail`, which is the right call for a guard: it doesn't depend on the tier editor's tab state. Backend is config-only with no logic to test, and `svelte-check` covers the required-prop half — the mutation tests in the description are the right substitute for a test that can't assert on a compile error.

### What's good

- **Preferring "impossible" to "detected" where the two diverge, and saying why.** Reading the constants directly in markup while leaving the entity configs on cross-checked literals looks inconsistent until the stated reason lands: an `EntityConfig` is a data object where the literal *is* the readable thing, whereas markup is hand-written and has no second literal to keep in step. Different structures, different guards — and the note declining to convert the 34 entity-config literals is the same judgment applied in reverse.
- **The required prop is the finding the issue didn't ask for.** #2377 asked to move the bounds onto `ContentFieldLengths` and extend the test. Noticing that the *uncovered* case is the input that doesn't exist yet — and that a type error catches it in any file, where a test only catches the files it renders — is the part that will still be paying off in a year.
- **Not retiring the column on the strength of two independent signals**, rather than reading "no editor input exists" as "unused". `required` on Core and unredacted on the wire are exactly the right two places to look, and either one alone would have been weaker.
- **Flagging the unclosed half in the description instead of quietly shipping the write side.** The PR does what the issue asked and says plainly what it didn't do and why — which is what made #2404 a five-minute file rather than a rediscovery three months from now.

---

ginderjeremiah:
Thanks — all four nits addressed in `cdb48c0`. Three fixed here, one filed as **#2405** with the reasoning below. CI was already fully green on `71e1808` before this push.

## 1. `boundsOf` selector — fixed, but the suggested selector wouldn't have worked

The finding is correct and the fix is in, but the proposed replacement doesn't close it:

```ts
container.querySelectorAll('textarea, input:not([type]), input[type="text"]')
```

`NumInput` renders **`type="text"`** with `inputmode="decimal"` (`NumInput.svelte:1-4`), not `type="number"` — so it matches the `input[type="text"]` clause and would still have been swept up, reporting `'Max level': null` exactly as before. The premise ("a number input correctly has no `maxlength`") is right; it's just that this control isn't a number input at the DOM level.

Scoped to `.inp` instead — `ProgInput`'s own class, on both its `<input>` and `<textarea>`. That tracks the thing the guard is actually about (every `ProgInput`-rendered field is included automatically, nothing else is), and it degrades safely: if the class is ever renamed the selector matches nothing and `toEqual` fails against the non-empty expectation rather than passing vacuously.

Verified both directions by temporarily adding a `ProgNumber` to `ConlangIdentity`: it no longer trips the length guard, and the wrong-constant mutation still fails loudly (`Translation` → `PROFICIENCY_DESCRIPTION_MAX_LENGTH` gives a `100 → 500` diff).

## 2. Mis-wire claim — scoped

You're right, and it's five of ten at `50` (`PATH_NAME`, `PROFICIENCY_NAME`, `PROFICIENCY_ICON_PATH`, `PROFICIENCY_WORD`, `PROFICIENCY_PRONUNCIATION`). The doc block now separates the unconditional half (an unregistered input, a removed one) from the value-dependent half, and says outright that swapping e.g. `Name` and `Icon path` reads identically.

## 3. `ApiCodeGeneratorTests` — dropped both

Agreed, no coverage added: `GetClientMirroredConstantFields` walks the whole `[ClientMirrored]` class, so a new `const` there can't be absent unless the incumbents are. Removed, and the "these ride the same generated file" intent stays in the `ContentFieldLengths` XML doc where you suggested.

## 4. Test scaffolding — filed as #2405

Taking your follow-up offer, because it isn't the mechanical extraction it looks like. The three copies have diverged in ways the existing assertions depend on: `tier().id` is `0` in `PathDetail.test.ts` but `5` in `TierDetail.test.ts`, `pathId` is the exact inverse, and the four conlang fields are empty in one and populated in the other. `TierDetail.test.ts` has 22 `makeStore(...)` sites asserting against id `5` and those strings. Unifying the defaults carelessly would silently *weaken* assertions rather than fail — the wrong failure mode to risk in the same diff as the guard itself. #2405 records the divergence table, the constraint that `vi.mock('$stores', …)` has to stay per-file (its factory can't reference imports), and the "keep the existing signatures so 30 call sites don't churn" approach.

## 5. The doc sentence — softened

Good catch that this was the load-bearing one. `docs/frontend-admin.md` now says the progression guard is **enumerated per-component** rather than data-driven like the entities one, so a new progression component has to be added to that test — since the required prop stops an omitted bound but not a hand-written literal. That's the sentence a future author reads before adding the third component, so it should describe the pattern rather than today's fields.

Also noted on **#2404** — agreed `Path.Description` has the identical status and they should be decided together. That widening is right; I'd only checked the proficiency side because #2377 only named that one.

---

Issue #2405, opened 2026-07-25
Consolidate the duplicated progression test scaffolding into a shared module


ginderjeremiah:
Follow-up from the review of #2403.

## Problem

There are now three near-identical copies of the progression editor's test scaffolding — the hoisted `staticData`/`dangerModal` `$stores` mock, the `path`/`tier` fixtures, and the `as unknown as ProgressionStore` fake store:

- `UI/src/tests/routes/admin/workbench/progression/PathDetail.test.ts:7-75`
- `UI/src/tests/routes/admin/workbench/progression/TierDetail.test.ts:7-76`
- `UI/src/tests/routes/admin/workbench/progression/field-lengths.test.ts` (added by #2403)

Two copies were pre-existing; the third is the signal that it's past tolerable duplication (`CLAUDE.md` → "prefer writing DRY code").

## Why it wasn't done inline in #2403

This is deliberately **not** a mechanical extraction, which is why it was split out rather than folded into that PR. The three copies have already diverged in ways the existing assertions depend on, so unifying them means reconciling the fixtures, not just moving them:

| | `PathDetail.test.ts` | `TierDetail.test.ts` |
|---|---|---|
| `tier().id` | `0` | `5` |
| `tier().pathId` | `5` | `0` |
| `iconPath` / `word` / `pronunciation` / `translation` | all `''` | `'i.png'` / `'sijren'` / `'sij-ren'` / `'The Riven Frost'` |

`TierDetail.test.ts` has 22 `makeStore(...)` call sites asserting against id `5` and those populated conlang strings (e.g. `expect(store.addPayout).toHaveBeenCalledWith(5, 1)`), and `PathDetail.test.ts` has 8 built on its own defaults. A careless unification would silently weaken assertions rather than fail loudly — e.g. a tier-collision case that depends on `pathId: 5`, or a decipher-preview assertion that depends on `word` being empty.

The fake stores differ too: `PathDetail`'s carries `pathBaseline`/`setPathTab`/`resetPath`, `TierDetail`'s carries the milestone/gateway writers (`addPayout`, `addModifier`, `addPrerequisite`, …), and only `PathDetail`'s seeds `paths` into `staticData`.

## Suggested direction

[...]

PR #2406, opened 2026-07-25
Render the authored path/proficiency descriptions on the Lexicon (#2404)

### LINKED ISSUE #2404: Surface the authored path/proficiency `description` on a player-facing surface
Follow-up to #2377 / PR #2403, which closed the **write** half: `IProficiency.description` now has an editor input in `ConlangIdentity`, so it can finally be authored. The **read** half is still open.

## Problem

Both descriptions reach every client and neither is rendered anywhere:

- `Proficiency.Description` is `required` on the Core model (`Game.Core/Proficiencies/Proficiency.cs:14`) and ships **unredacted** — `AbstractReferenceDataCommand.RedactAuthoringOnlyFields` blanks only `DesignerNotes`, so unlike a design note this is player-facing content by construction.
- `Path.Description` has the **identical** status: authorable in `PathDetail`, delivered to the client, rendered nowhere. #2377 only named the proficiency one, but the gap is symmetric and both should be resolved together.

Nothing under `routes/game/screens/proficiencies` reads `.description` today.

## Why it wasn't done in #2403

Where a blurb goes is a content/UI design call, not a mechanical addition. The natural home is the Lexicon's `WordDetail` inspector (#1218), but that pane is densely designed — the header/state pill, XP bar, per-level ladder, and decipher reveal line already compete for vertical space — and the path blurb may want a different home again (the `LexiconRail`, or the spine header).

## Work

- Decide where the tier blurb and the path blurb each belong (`WordDetail` / `LexiconRail` / `TierSpine`), including whether a tier blurb should be gated by decipher stage the way the word/pronunciation/translation are, or always visible once the tier is discovered.
- Render both via existing theme tokens, no hardcoded colours (`docs/frontend.md`).
- Component tests covering present / empty description (the field is `required` but may legitimately be `''`, so the empty case must not leave a dangling heading or separator).
- Update `docs/frontend-screens.md` for whichever surface changes.

## Alternative

If the answer is genuinely "there is no room for it", the honest outcome is to retire both columns rather than leave authored-but-invisible content — the second half of #2377's original either/or. That is a migration plus Core/Abstractions/Infrastructure/mapper/repo churn, so it needs deciding before it's worth doing.

<sub>Filed from the review of #2403.</sub>

ginderjeremiah:
Closes #2404.

Frontend-only. Both `description` fields already reached every client unredacted (`RedactAuthoringOnlyFields` blanks `DesignerNotes` alone) and were rendered nowhere — this closes the **read** half. No backend, contract, or migration changes: `IPath.description` and `IProficiency.description` are already on the generated contracts.

## The two design calls the issue asked for

**1. Where each blurb goes.**

- **Path → the `TierSpine` header**, under the `name · N WORDS KNOWN` line. Of the three candidates this is the only surface that presents a path as a *subject* rather than a list row: `LexiconRail` rows are 182px of icon + word + name, where prose would wreck scannability. The header column already has room (the spine is `min-width: 440px`), so the blurb is measure-capped at `64ch` rather than allowed to run the full width.
- **Proficiency → the `WordDetail` inspector, as a lede** directly under the header it describes. I deliberately did *not* use the pane's `section-head` + hairline-rule idiom: the issue's concern is vertical space in an already-dense pane, and a lede costs one paragraph where a titled section costs a heading too. It also keeps the existing `word block → reveal → XP → ladder` run — which reads as one continuous "your progress" sequence — undisturbed, rather than splitting it.

**2. Decipher gating: no — always visible once the tier is discovered.**

The decipher stage gates the **word of power** specifically: the glyph's illumination, the pronunciation, the translation. It has never gated the tier's own identity — name, path, state pill, XP, the full per-level payout ladder, and the milestone reward skills are all fully legible at `undeciphered`. Gating prose about that same tier would be inconsistent with everything already on the pane, and it would hide content while buying no mystery, since the ladder beneath it already spells out exactly what the tier does. Pinned by a test that asserts the blurb renders at all three stages.

## Empty handling

Both fields are `required` on the contract but may legitimately be `''`. Each is wrapped in an `{#if}` so an unauthored blurb omits the paragraph entirely — no dangling heading, separator, or empty `<p>` inheriting the block margin. Both surfaces have an explicit test for it.

## Test plan

- [x] `npm run lint` (eslint + prettier + svelte-check) — 1219 files, 0 errors, 0 warnings
- [x] `npm run test:unit:ci` — **4006 passed** across 311 files (6 new; 4000 before)
- [x] New coverage: tier blurb present / empty / at each of the three decipher stages (`WordDetail.test.ts`), path blurb present / empty (`TierSpine.test.ts`), and both fields flowing `IPath`/`IProficiency` → `PathView`/`TierView` through `buildLexicon` (`proficiencies-view.test.ts`)
- [x] The screen's existing `token-compliance.test.ts` covers the new styles by glob — both blurbs use `--text-secondary`, no literal hex
- [x] No backend files touched, so no `dotnet build`/`test` run — CI's changed-path gate skips those jobs
- [x] `docs/frontend-screens.md` → Proficiencies screen updated with the placement + ungated-by-decipher decision

## Notes for review

- **`WordDetail`'s header comment claimed the pane renders "the seed skill".** It does not, and no `seedSkillId` reaches the view-model — stale since the pane was written. Dropped it while rewriting that comment for the blurb rather than leaving a wrong description in place.
- **The five `tierView`/`pathView` fixture builders across this screen's tests are near-identical copies** (`WordDetail`, `TierSpine`, `TierCard`, `WordOfPowerTooltip`, `word-detail`) — adding one field meant editing all five. That's the same duplication class as #2405, in a different folder; filed as a follow-up rather than folded in here, since consolidating five diverged fixture sets is its own reconciliation job and unrelated to this issue.
- **The issue's alternative — retiring both columns — is not what I did**, and I want to be explicit that this was a judgment call rather than a foregone conclusion. Placement turned out not to be forced: the path blurb has an uncontested home, and the tier blurb fits the inspector without displacing anything. Given both fields are already authorable, already delivered, and `Proficiency.Description` is `required` on the Core domain model, rendering them is the smaller and more reversible move than a migration plus Core/Abstractions/Infrastructure/mapper/repo churn.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CYXNotNfosD27fDckYp2mp)_


UI/src/routes/game/screens/proficiencies/TierSpine.svelte
```diff
@@ -104,6 +110,15 @@ const wordCount = $derived(path.tiers.length === 1 ? '1 WORD KNOWN' : `${path.ti
...
 }
 
+// Measure-capped so the blurb stays readable as the spine column grows past its 440px minimum.
+.path-blurb {
+	margin: 8px 0 0;
+	max-width: 64ch;
+	font-size: 12.5px;
+	line-height: 1.55;
+	color: var(--text-secondary);
+}
```
ginderjeremiah, line 122: **Should-fix — the icon drifts once this paragraph exists.**

`.path-header` is `display: flex; align-items: center` (`TierSpine.svelte:65-71`), and the blurb lands inside `.path-text`, the flex item this rule centres against. Before this PR the text column was ~67px (glyph 46 + 7 margin + sub), so the 48px `.path-icon` sat roughly level with the path word by accident of the two heights nearly matching. Add a 2–3 line blurb and `.path-text` grows to 110–140px, so the icon centres against the *whole* block — floating down to sit beside the blurb rather than beside the word it labels.

It only shows up when a blurb is authored, which is exactly the case this PR exists to serve (and `content/paths.json` is empty today, so nothing on screen would have surfaced it yet).

One line on the header fixes it:

```scss
.path-header {
	align-items: flex-start;
	...
}
```

That pins the icon to the top of the text column, level with the word of power, and is a no-op for the unauthored case since the two heights already match there. Worth a glance at the 48px box vs. the 46px glyph after the change — if it reads a pixel high, `.path-icon { margin-top: -1px }` or aligning on the glyph's box is the follow-through.


---

UI/src/routes/game/screens/proficiencies/WordDetail.svelte
```diff
@@ -187,6 +196,14 @@ const chips = $derived(trainedBy(path.activityKey));
 	border-color: var(--border-medium);
 }
 
+// The authored blurb — body copy, so it reads a step quieter than the name but ahead of the mono chrome.
+.blurb {
+	margin: 10px 0 0;
+	font-size: 12.5px;
+	line-height: 1.55;
+	color: var(--text-secondary);
+}
```
ginderjeremiah, line 205: **Nit — `1.55` is a new value for a class of block that already has one.**

Every other authored-prose block on a game screen uses `line-height: 1.5`:

- `synthesis/ResultDossier.svelte:227` — `12.5px / 1.5 / --text-secondary` (the closest sibling; same size, same token)
- `attribute-breakdown/BreakdownDetail.svelte:84` — `12.5px / 1.5`
- `codex/DossierShell.svelte:89` — `12px / 1.5`
- `challenges/ChallengeDetailCard.svelte:105` — `12px / 1.5`

`12.5px` and `--text-secondary` match `ResultDossier` exactly, so the only thing that diverges is the leading, in both new rules (here and `TierSpine.svelte:118`). Nothing breaks, but it's the kind of one-off that makes the next author unsure which value is the intended one. Suggest `1.5` in both unless the extra leading is deliberate.


---

ginderjeremiah (commented):
## Review: render the authored path/proficiency `description` on the Lexicon (#2404)

**Verdict:** Approve (posted as `COMMENT` — GitHub won't take a gating verdict from the PR author) — no blocking findings. **Not merging**, on one should-fix that's a one-line CSS change, plus a nit worth folding in on the same pass. CI is green on the head commit (`verify-frontend`, `test-frontend`, all three `test-e2e` browsers; backend jobs correctly skipped by the changed-path gate) with only the `all-checks-passed` aggregator still running.

I checked the description's premises against the tree rather than taking them. They hold: `description` is on both generated contracts as `required` (`contracts.ts:175` / `:184`), and `RedactAuthoringOnlyFields` (`AbstractReferenceDataCommand.cs:33-48`) blanks only `IHasDesignerNotes.DesignerNotes`, so both fields genuinely reach every client unredacted. This is a pure read-side change with no backend, contract, or migration surface.

### The two design calls

**Placement — both right, and the reasoning is the part that holds up.** The `LexiconRail` rejection is correct on inspection: those rows are a fixed 182px of icon + glyph + name, and prose in them would destroy the scannability the rail exists for. `TierSpine`'s header is the only surface that frames a path as a subject, and the lede-not-`section-head` call in `WordDetail` is the right read of a 312px pane that already runs eyebrow → head → word block → XP → ladder → chips. Declining the pane's `section-head` idiom costs one paragraph instead of a heading *and* a hairline rule, and leaves the "your progress" run intact rather than splitting it — that's a real argument, not a rationalization.

**Ungated by decipher — agreed, and for the stated reason.** The stage gates the *word of power*; the tier's name, state pill, XP, full payout ladder, and milestone skills are all already legible at `undeciphered`. Gating prose about that same tier would be inconsistent with everything already on the pane. Pinning it with a loop over all three stages is the right way to make that decision durable rather than incidental.

One thing the decision leaves open, which is authoring discipline rather than code: **a tier blurb can pre-empt the translation reveal.** The ladder spells out mechanical payouts, but the *translation* is the word's meaning — and a blurb is prose by the same author about the same tier, so "Words that bind flame to will" is one paraphrase away from being the translation shown at max level. Not a reason to gate (gating the blurb would hide it for the first half of every tier, which is worse), but worth a sentence of guidance in `docs/content-design.md` whenever tier prose actually gets authored. Both `content/paths.json` and `content/proficiencies.json` are empty today, so there's no live content to check against and nothing to fix here — flagging it so it's on the record before the first blurb is written, not asking for it in this PR.

### Should-fix

- **`TierSpine.svelte:113-120` — `.path-header` is `align-items: center`, so the 48px icon drifts down once the blurb makes `.path-text` taller.** It reads level with the path word today only because the two heights happen to nearly match; a 2–3 line blurb decouples them. `align-items: flex-start` on the header, detail inline.

### Nits

- **`WordDetail.svelte:199-205` and `TierSpine.svelte:118` — `line-height: 1.55`** where all four sibling prose blocks on game screens use `1.5` (`ResultDossier:227`, `BreakdownDetail:84`, `DossierShell:89`, `ChallengeDetailCard:105`). Size and token match `ResultDossier` exactly; only the leading diverges. Detail inline.
- **`docs/frontend-screens.md:143`** — the new bullet says "renders as the inspector's lede" one line *before* the bullet that introduces the inspector. Reads slightly backwards; moving it after the `WordDetail` bullet would fix it. Genuinely trivial.

### Test coverage

Meets the issue's ask and then some. Present/empty is covered on **both** surfaces (`WordDetail.test.ts`, `TierSpine.test.ts`), the three-stage loop pins the ungated decision, and `proficiencies-view.test.ts` covers the `IPath`/`IProficiency` → `PathView`/`TierView` carry in `buildLexicon` — so a field dropped in the mapper fails at the view-model layer rather than only in a render assertion. That last one is the test most likely to catch a real future regression.

The in-loop `cleanup()` at `WordDetail.test.ts:149` is correct and matches the existing pattern (`TierCard.test.ts:61`); `afterEach(() => cleanup())` is already in place for the single-render cases. The empty-case tests assert `queryByTestId(...)` is `null`, which is the right assertion for "no dangling empty `<p>`" — checking for absent text would have passed on an empty paragraph inheriting the block margin, which is exactly the failure the issue named.

The existing `token-compliance.test.ts` glob does cover both new rules (verified: it globs `**/*.{svelte,ts}` under the screen and both use `--text-secondary` with no literal hex), so no new theming guard was needed.

`{tier.description}` / `{path.description}` are Svelte text interpolations, so authored content is escaped — no injection surface from making a DB-authored string player-visible.

### What's good

- **Deleting the wrong comment instead of writing around it.** `WordDetail`'s header claimed the pane renders "the seed skill"; it doesn't, and no `seedSkillId` reaches the view-model. Noticing that while editing the same comment for an unrelated reason, and fixing it rather than preserving it, is the habit `CLAUDE.md` asks for on stale code you happen to be standing next to.
- **Being explicit that the retire-both alternative was a judgment call, not a foregone one.** #2404 recorded a real either/or, and the PR argues its side on the merits — placement turned out not to be forced, both fields are already authorable and already delivered, and `Proficiency.Description` is `required` on the Core model. That's the smaller and more reversible move, and saying so plainly is what makes it reviewable.
- **The empty-string handling is designed, not defensive.** `required` on the contract but legitimately `''` is the awkward case, and `{#if}` around the whole paragraph — rather than a fallback string or an empty `<p>` — is the correct resolution, with a test on each surface so it stays that way.
- **#2407 filed rather than folded in.** Five near-identical fixture builders is real duplication and this PR paid the tax on all five, but reconciling diverged fixtures can silently weaken assertions (`TierSpine`'s `pathOrdinal: o.id` is load-bearing for its ordering test). Splitting it out with that constraint written down is the right call.


---

ginderjeremiah:
All three findings addressed in `4d204d1`. I checked each claim against the tree rather than applying them on assertion; all three held.

**Should-fix — icon drift (`TierSpine.svelte`).** Confirmed and fixed: `.path-header` was `display: flex; align-items: center`, and the blurb lands inside `.path-text`, the flex item that rule centres against. `align-items: flex-start` now pins the icon to the top of the text column, level with the word of power it labels, and is a no-op in the unauthored case. Comment added on the rule so the next author doesn't "tidy" it back to `center`.

I did **not** add the suggested `.path-icon { margin-top: -1px }` follow-through. With `flex-start` the two boxes share a top edge — the 48px icon and the 46px glyph line box (`WordOfPower` at `size={46}`, `line-height: 1`) differ only in that the icon extends 2px lower, which top-alignment makes invisible. A negative margin to correct a bottom-edge difference would push the top edges *out* of alignment. If it reads high against real art, that's a one-line tweak against something visible rather than a guess now.

**Nit — `line-height`.** Verified all four siblings: `ResultDossier:230`, `BreakdownDetail:87`, `DossierShell:92`, `ChallengeDetailCard:109`, all `1.5`. Changed both new rules from `1.55`. `1.55` wasn't deliberate, so there's no reason to keep a one-off.

**Nit — doc bullet order.** Moved the blurb bullet after the `WordDetail` bullet it refers to.

**On the translation-spoiler point:** agreed, and agreed it isn't this PR's — there's no live content to check against (`content/paths.json` and `content/proficiencies.json` are both empty), and a lint for "prose paraphrases a translation" would be over-engineering a judgment call. Filed as **#2408** so it's on the record where the first author of tier prose will actually find it, rather than in a merged PR thread. It also picks up the adjacent gap that the two `description` fields now have exactly one player-facing home each and no authoring guidance at all.

Re-verified after the changes: `npm run lint` clean (1219 files, 0 errors/warnings), `npm run test:unit:ci` 4006 passed across 311 files. The changes are CSS-only plus a doc reorder, so no test needed updating — nothing asserted the old leading or the header's alignment.

---

PR #2411, opened 2026-07-25
Consolidate the duplicated progression test scaffolding into a shared module (#2405)

### LINKED ISSUE #2405: Consolidate the duplicated progression test scaffolding into a shared module

ginderjeremiah:
Closes #2405.

## Summary

Adds `UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts` holding the `$stores` stub, the `path`/`tier` fixtures, and the two detail-store builders, and points the progression suites at it. **Test-only — no production code is touched, and no assertion changes meaning.**

Net −291/+64 across the suites, plus the 128-line shared module.

## How it addresses the issue

**One deviation from the suggested direction, deliberately.** The issue states the `vi.hoisted` + `vi.mock('$stores', …)` block "has to stay per-file… that's ~6 lines left duplicated per file, which is the floor here." The premise (a sync factory can't close over the file's static imports) is correct, but an **async** factory may `await import(…)` — a pattern this repo already uses in five places, e.g. `src/tests/lib/engine/engine.test.ts:51`. So each suite still declares its own hoisted `vi.mock` call, but it resolves the shared singletons:

```ts
vi.mock('$stores', async () => (await import('./progression-test-utils')).stubStores());
```

That drops the per-file floor from ~14 lines to 1 and puts the stub's *shape* (the nine `staticData` catalogues) in exactly one place — which is the drift this issue exists to close. The `beforeEach` reset collapses to `beforeEach(resetStores)`.

**Fixture divergences stay literal in the suite that depends on them**, exactly as the issue asks — as thin wrappers over the shared builder rather than reconciled-away defaults:

| Suite | Wrapper over shared `tier()` | Why |
|---|---|---|
| `PathDetail` | `{ pathId: 5 }` | its retire check walks the tiers carrying the selected path's id |
| `TierDetail` | `{ id: 5 }` + the four conlang strings | 22 call sites assert against id 5; `ConlangIdentity`'s decipher preview reads the authored word/pronunciation/translation |
| `MilestonesEditor` | `{ id: 5 }` | payout assertions target id 5 |
| `GatewaysEditor` | none | its defaults already match the shared base, and every call site passes id/pathId explicitly |

Local `tier` / `makeStore` signatures are unchanged, so none of the ~40 call sites churn.

## Beyond the issue's scope (flagging explicitly)

- **Two extra copies folded in.** The issue names three files, but `MilestonesEditor.test.ts` and `GatewaysEditor.test.ts` carry the same `tier()` builder (and `GatewaysEditor` a `path()`) — 5 copies total. Leaving two behind would have defeated the point, so both are converted.
- **`as any as ProgressionStore` → `asStore`.** Those two suites cast through `any` and each carried an `@typescript-eslint/no-explicit-any` disable to do it; `PathDetail`/`TierDetail` used `as unknown as`, which needs no disable. The shared `asStore` helper uses the latter, so both disables are gone.

## Notes

The one-file-at-a-time process the issue prescribes did its job: transcribing `makeTierStore` dropped `addPrerequisite`/`removePrerequisite`, and `TierDetail`'s gateway tests failed loudly (`undefined is not a spy`) rather than silently weakening. Fixed before moving on.

## Test plan

- [x] Each suite run individually as it was converted — `PathDetail` 8/8, `TierDetail` 20/20 (after the fix above), `field-lengths` 2/2.
- [x] Whole progression folder: **122/122 passing**, same count as before the change.
- [x] `npm run lint` (eslint + `svelte-check`): 1221 files, 0 errors, 0 warnings.
- [x] Full unit suite `npm run test:unit:ci`: **312 files / 4009 tests passing**, coverage thresholds met.
- [x] Confirmed the new `.ts` helper isn't collected as an empty suite — Vitest's `include` is `src/**/*.{test,spec}.{js,ts}` (`vite.config.ts:8`), and the file count is unchanged at 312.

No backend code is touched, so `dotnet build`/`test` are unaffected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SKUnP3UcX6UCkd15sihHx2)_

### DIFF
--- a/UI/src/tests/routes/admin/workbench/progression/PathDetail.test.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/PathDetail.test.ts
@@ -2,90 +2,26 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import { render, cleanup, screen, fireEvent, waitFor } from '@testing-library/svelte';
 import { EActivityKey } from '$lib/api';
 
-// Hoisted so individual tests can seed staticData and assert the retire-confirm flow, mirroring
-// TierDetail.test.ts's pattern for the progression editor's retire dialog.
-const { staticData, dangerModal } = vi.hoisted(() => ({
-	staticData: {
-		enemies: [] as unknown[],
-		zones: [] as unknown[],
-		challenges: [] as unknown[],
-		items: [] as unknown[],
-		classes: [] as unknown[],
-		skillRecipes: [] as unknown[],
-		proficiencies: [] as unknown[],
-		skills: [] as unknown[],
-		paths: [] as unknown[]
-	},
-	dangerModal: vi.fn()
-}));
-vi.mock('$stores', () => ({ staticData, dangerModal }));
+// `vi.mock` is hoisted within this file, so the factory resolves the shared stub by dynamic import
+// rather than closing over the static one below.
+vi.mock('$stores', async () => (await import('./progression-test-utils')).stubStores());
 
 import PathDetail from '$routes/admin/workbench/progression/PathDetail.svelte';
-import type { ProgressionStore } from '$routes/admin/workbench/progression/progression-store.svelte';
-import type { WorkbenchPath, WorkbenchProficiency } from '$routes/admin/workbench/progression/types';
+import type { WorkbenchProficiency } from '$routes/admin/workbench/progression/types';
+import {
+	dangerModal,
+	makePathStore as makeStore,
+	path,
+	resetStores,
+	staticData,
+	tier as baseTier
+} from './progression-test-utils';
 
-const path = (over: Partial<WorkbenchPath> = {}): WorkbenchPath => ({
-	id: 5,
-	name: 'Fire Path',
-	description: '',
-	designerNotes: '',
-	activityKey: EActivityKey.Fire,
-	...over
-});
+// PathDetail's retire check walks the tiers carrying the selected path's id, so tiers here default
+// onto path 5 (`path()`'s id) rather than the shared fixture's path 0.
+const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency => baseTier({ pathId: 5, ...over });
 
-const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency => ({
-	id: 0,
-	name: 'Blades',
-	description: '',
-	iconPath: '',
-	word: '',
-	pronunciation: '',
-	translation: '',
-	pathId: 5,
-	pathOrdinal: 0,
-	maxLevel: 10,
-	baseXp: 100,
-	xpGrowth: 1.4,
-	designerNotes: '',
-	levelModifiers: [],
-	levelRewards: [],
-	prerequisiteIds: [],
-	...over
-});
-
-// A fake store exposing exactly what PathDetail reads — mirrors TierDetail.test.ts's fake-store
-// pattern rather than driving the real ProgressionStore through a socket-backed load().
-const makeStore = (selectedPath: WorkbenchPath, overrides: Record<string, unknown> = {}) =>
-	({
-		selectedPath,
-		profs: [],
-		paths: [],
-		currentTiers: [],
-		pathTab: 'identity',
-		saving: false,
-		pathStatus: vi.fn(() => 'clean'),
-		isRetired: vi.fn(() => false),
-		setPathTab: vi.fn(),
-		resetPath: vi.fn(),
-		retirePath: vi.fn(),
-		removePath: vi.fn(),
-		pathBaseline: vi.fn(() => selectedPath),
-		patchPath: vi.fn(),
-		...overrides
-	}) as unknown as ProgressionStore;
-
-beforeEach(() => {
-	dangerModal.mockReset();
-	staticData.enemies = [];
-	staticData.zones = [];
-	staticData.challenges = [];
-	staticData.items = [];
-	staticData.classes = [];
-	staticData.skillRecipes = [];
-	staticData.proficiencies = [];
-	staticData.skills = [];
-	staticData.paths = [];
-});
+beforeEach(resetStores);
 afterEach(cleanup);
 
 describe('PathDetail', () => {
--- a/UI/src/tests/routes/admin/workbench/progression/PathDetail.test.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/PathDetail.test.ts
@@ -19,7 +19,8 @@ import {
 
 // PathDetail's retire check walks the tiers carrying the selected path's id, so tiers here default
 // onto path 5 (`path()`'s id) rather than the shared fixture's path 0.
-const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency => baseTier({ pathId: 5, ...over });
+const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency =>
+	baseTier({ pathId: path().id, ...over });
 
 beforeEach(resetStores);
 afterEach(cleanup);
--- a/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
@@ -5,7 +5,12 @@ import type { WorkbenchPath, WorkbenchProficiency } from '$routes/admin/workbenc
 
 /* Shared scaffolding for the progression editor's component suites (#2405). Each suite still declares
    its own `vi.mock('$stores', …)` — `vi.mock` is hoisted within the file it appears in — but the factory
-   resolves this module's singletons, so the stub's shape lives in exactly one place. */
+   resolves this module's singletons, so the stub's shape lives in exactly one place.
+
+   Because the suites resolve it from inside that factory, this module is evaluated *while `$stores` is
+   being mocked, so it must stay free of runtime imports that transitively reach `$stores`* — a value
+   import that did would re-enter the factory resolving it. Everything below `$lib/api` is `import type`
+   and therefore erased, which is what keeps that true today. */
 
 /** The last-saved `staticData` catalogues the editor's reference lookups read. */
 export const staticData = {
@@ -35,8 +40,8 @@ export const stubStores = () => ({ staticData, dangerModal });
 /** Restores the stub to its empty, uncalled baseline. Call from `beforeEach`. */
 export const resetStores = () => {
 	dangerModal.mockReset();
-	for (const catalogue of Object.keys(staticData) as (keyof typeof staticData)[]) {
-		staticData[catalogue] = [];
+	for (const catalogue of Object.values(staticData)) {
+		catalogue.length = 0;
 	}
 };
 
--- a/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
@@ -1,0 +1,132 @@
+import { vi } from 'vitest';
+import { EActivityKey } from '$lib/api';
+import type { ProgressionStore } from '$routes/admin/workbench/progression/progression-store.svelte';
+import type { WorkbenchPath, WorkbenchProficiency } from '$routes/admin/workbench/progression/types';
+
+/* Shared scaffolding for the progression editor's component suites (#2405). Each suite still declares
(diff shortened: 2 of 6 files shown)


UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
```diff
@@ -0,0 +1,132 @@
+import { vi } from 'vitest';
+import { EActivityKey } from '$lib/api';
+import type { ProgressionStore } from '$routes/admin/workbench/progression/progression-store.svelte';
+import type { WorkbenchPath, WorkbenchProficiency } from '$routes/admin/workbench/progression/types';
+
+/* Shared scaffolding for the progression editor's component suites (#2405). Each suite still declares
+   its own `vi.mock('$stores', …)` — `vi.mock` is hoisted within the file it appears in — but the factory
+   resolves this module's singletons, so the stub's shape lives in exactly one place. */
```
ginderjeremiah: **Nit** — worth one more clause here recording the constraint this pattern silently imposes: **this module must stay free of runtime imports that transitively reach `$stores`.**

Because the suites resolve it from inside the `$stores` mock factory, this module is evaluated *while `$stores` is being mocked*. It's safe today only because everything but `vitest` and `$lib/api` is an `import type` (lines 3-4) and therefore erased. A later edit that adds an innocent-looking value import reaching `$stores` would re-enter the factory that's resolving it.

That's exactly the kind of deliberate-looking constraint a future edit can't infer from reading the file, so it's worth stating rather than leaving to be rediscovered.

---

UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
```diff
@@ -0,0 +1,132 @@
...
+ * ```
+ */
+export const stubStores = () => ({ staticData, dangerModal });
+
+/** Restores the stub to its empty, uncalled baseline. Call from `beforeEach`. */
+export const resetStores = () => {
+	dangerModal.mockReset();
+	for (const catalogue of Object.keys(staticData) as (keyof typeof staticData)[]) {
+		staticData[catalogue] = [];
+	}
```
ginderjeremiah: **Nit** — the `keyof typeof` cast is avoidable, since every value is already `unknown[]`:

```ts
for (const catalogue of Object.values(staticData)) {
	catalogue.length = 0;
}
```

Slightly stricter too: an in-place clear is visible to anything that captured the array by reference, where replacement leaves that capture stale. Not load-bearing here (the suites seed by assigning `staticData.paths = [...]`, so replacement is what's actually exercised), just one less assertion to carry.

---

UI/src/tests/routes/admin/workbench/progression/PathDetail.test.ts
```diff
@@ -2,90 +2,26 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
...
+	makePathStore as makeStore,
+	path,
+	resetStores,
+	staticData,
+	tier as baseTier
+} from './progression-test-utils';
+
+// PathDetail's retire check walks the tiers carrying the selected path's id, so tiers here default
+// onto path 5 (`path()`'s id) rather than the shared fixture's path 0.
+const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency => baseTier({ pathId: 5, ...over });
```
ginderjeremiah: **Nit** — the comment right above explains this `5` *is* `path()`'s id, so derive it rather than restating the literal:

```ts
const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency =>
	baseTier({ pathId: path().id, ...over });
```

`field-lengths.test.ts:41` already does exactly this (`tierFixture({ pathId: path.id })`), so this is the one place in the PR where the shared default is re-stated instead of read.

To be fair on severity: I traced what happens if `path()`'s default id ever moves, and these assertions *do* fail loudly rather than weaken silently — with no tier on the selected path, `expect(dangerModal).toHaveBeenCalledOnce()` at line 57 goes red. So this is genuinely cosmetic, not the silent-weakening hazard #2405 warned about. Just consistency with the sibling suite.

---

ginderjeremiah (commented):
## Review: consolidating the progression editor's duplicated test scaffolding into a shared module (#2405)

**Verdict: Approve** — the extraction is correct, the divergences the issue flagged as load-bearing are all preserved, and the one deviation from the suggested direction is a genuine improvement. Three nits below, none blocking.

_(Submitted as a comment rather than an approval only because GitHub won't accept a gating verdict from the PR's own author.)_

### Verification

I checked the specific failure mode #2405 was written around — a "careless unification silently weakening assertions rather than failing loudly" — against each converted suite:

- **`TierDetail`** — id `5` and the four conlang strings are preserved in the local wrapper (`TierDetail.test.ts:21-29`); the 22 call sites asserting `toHaveBeenCalledWith(5, …)` still get id 5. The shared `makeTierStore` retains `addPrerequisite`/`removePrerequisite`, which is what the transcription slip you describe would otherwise have dropped.
- **`PathDetail`** — tiers still default onto path 5, so the retire-check walk still finds them.
- **`GatewaysEditor`** — I checked the claim that it needs no wrapper, because the shared `path()` default changed underneath it (`id: 0, name: 'Fire'` → `id: 5, name: 'Fire Path'`). Confirmed safe: every `path(...)` and `tier(...)` call site in that file passes `id`/`pathId`/`name` explicitly (e.g. `:89-91`, `:21-23`), so no assertion reads a shared default. Worth having stated in the description — that one was the least obvious of the five.
- **`field-lengths`** — `pathId` derived from `path.id`, blanks preserved; assertions are on rendered `maxlength` only, so the fixture identity doesn't reach them.
- **Superset stubs are safe** — `TierDetail` now gets a `staticData.paths` its old mock lacked, and `MilestonesEditor` gets the full catalogue plus `dangerModal` where it previously had only `{ skills: [] }`. Both are strictly wider than before, so neither can narrow behavior.
- **Coverage** — `src/tests/**` is excluded in `coverage.config.ts:15`, so the new non-`.test.ts` helper can't drag the ratchet.

### Nits

- `progression-test-utils.ts:6-8` — record the constraint the dynamic-import pattern imposes: this module must stay free of runtime imports reaching `$stores`. Safe today only because everything but `vitest`/`$lib/api` is `import type`.
- `progression-test-utils.ts:38-40` — the `keyof typeof` cast in `resetStores` is avoidable via `Object.values`.
- `PathDetail.test.ts:22` — derive `pathId` from `path().id` instead of restating `5`, matching `field-lengths.test.ts:41`.

### Follow-up

Filed **#2412**. Two suites in the same folder — `ProgressionMap.test.ts:19-50` and `progression-map.test.ts:6-37` — carry a **byte-for-byte identical** `path(id, over)` / `tier(id, pathId, ordinal, over)` pair, 44 lines duplicated across adjacent files. By this PR's own stated principle ("leaving two behind would have defeated the point") they're candidates, but they use a positional signature with generated-name defaults the map assertions read back out of the DOM, so folding them in is a real reconciliation across ~15 call sites in files #2405 never named — out of scope here, not a reason to hold this. The issue also notes `progression-helpers.test.ts:4-8`, which still stubs `staticData: {} as any`.

### Test coverage

Test-only change; no production code touched, so the project's coverage rules apply as "don't lose what was there." The progression folder holds at 122/122 with no assertion changing meaning, and the check that the new `.ts` helper isn't collected as an empty suite (Vitest's `include` at `vite.config.ts:8`) was the right thing to confirm rather than assume.

### What's good

- **The deviation from the issue's prescribed floor is the best part of the PR.** #2405 asserted the `vi.hoisted` block "has to stay per-file… ~6 lines left duplicated, which is the floor here." That premise was right about sync factories and wrong about the conclusion — an async factory may `await import(…)`, and citing existing in-repo precedent (`engine.test.ts:51`) rather than asserting it makes the case checkable. Dropping the per-file floor from ~14 lines to 1 puts the nine-catalogue stub shape in exactly one place, which is the drift the issue actually exists to close. Correctly not treating the issue's stated constraint as binding.
- **The fixture divergences are handled the way the issue asked** — kept literal and local, as thin wrappers with a one-line reason each, rather than reconciled away into shared defaults that later readers can't audit.
- Replacing `as any as ProgressionStore` with `as unknown as` removes two `@typescript-eslint/no-explicit-any` disables as a side effect of the extraction, rather than as unrelated churn.
- Reporting that the one-file-at-a-time process caught a real dropped-method slip is the kind of note that makes a test refactor reviewable at all.

---

ginderjeremiah:
All three nits applied in 8c16f21.

- **Module-evaluation constraint** — recorded in the header comment. Worth having flagged: it's a real hazard that reads as an accident, since the file looks like it could take any import.
- **`Object.values` + in-place clear** — adopted, cast gone. Your note that replacement leaves a captured reference stale is the better argument of the two; the suites do seed by assignment today, but the in-place form is correct under both.
- **`pathId: path().id`** — derived rather than restated, matching `field-lengths.test.ts`. Agreed on the severity read: I'd traced the same thing, that a moved default fails loudly at the `toHaveBeenCalledOnce()` assertion rather than weakening quietly.

On the `GatewaysEditor` point — you're right that it was the least obvious of the five and deserved to be in the description rather than left for the reviewer to verify. The shared `path()` default did move underneath it (`id: 0, name: 'Fire'` → `id: 5, name: 'Fire Path'`); it's safe only because every call site there passes `id`/`pathId`/`name` explicitly. Noting it here so the reasoning is on the record with the change.

No action taken on #2412 — agreed it's a separate reconciliation across files #2405 never named.

Verification after the fixes: progression folder 122/122, `npm run lint` clean across 1221 files, full unit suite 312 files / 4009 tests passing.

---

Issue #2412, opened 2026-07-25
Fold the progression map suites' duplicated path/tier builders into progression-test-utils


ginderjeremiah:
Follow-up from the review of #2411 (which closed #2405).

## Problem

#2411 consolidated the progression editor's scaffolding into `UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts`, converting five suites. Two suites in the same folder were left behind, and they carry a **byte-for-byte identical** pair of fixture builders:

- `UI/src/tests/routes/admin/workbench/progression/ProgressionMap.test.ts:19-50`
- `UI/src/tests/routes/admin/workbench/progression/progression-map.test.ts:6-37`

Both declare the same `path(id, over)` and `tier(id, pathId, ordinal, over)` — 44 lines duplicated exactly across two files sitting next to each other. That is the same drift class #2405 existed to close (`CLAUDE.md` → "prefer writing DRY code"), and an exact duplicate is the strongest possible case for extraction.

## Why it wasn't done in #2411

Not an oversight — these two use a **positional** signature (`tier(10, 0, 0)`, ~15 call sites) rather than the `tier(over)` shape the shared module exports, and they also default differently (`name: \`Tier ${id}\``/`` `Path ${id}` `` rather than `'Blades'`/`'Fire Path'`, and `EActivityKey.Physical` rather than `Fire`) because the map assertions read the generated names back out of the rendered DOM. Reconciling them is a real reconciliation, not a move — the same "silently weakening assertions" hazard #2405 called out — and it touches files #2405 never named.

## Suggested direction

Prefer a thin positional adapter over changing the ~15 call sites:

```ts
// in progression-test-utils.ts
export const mapTier = (id: number, pathId: number, ordinal: number, over: Partial<WorkbenchProficiency> = {}) =>
	tier({ id, pathId, pathOrdinal: ordinal, name: `Tier ${id}`, ...over });
```

...with the equivalent for `path`. That keeps one source of truth for the field set while preserving each suite's call-site ergonomics and its generated-name defaults.

Worth deciding as part of this whether two builder signatures for one type is itself a smell — if so, the alternative is migrating the ~15 call sites to the `over`-object shape instead. Either way, convert one file at a time and run the suite in between, per #2405's process note: `ProgressionMap.test.ts` asserts on rendered node text, so a changed default must fail loudly rather than pass weakly.

[...]

PR #2413, opened 2026-07-25
Consolidate the duplicated Lexicon test fixture builders (#2407)

### LINKED ISSUE #2407: Consolidate the five duplicated Lexicon (proficiencies) test fixture builders
Found while implementing #2404 (PR #2406).

## Problem

The Proficiencies "Lexicon" screen's tests carry **five** near-identical copies of the `tierView(o)` fixture builder, plus three of `pathView(...)`:

- `UI/src/tests/routes/game/screens/proficiencies/WordDetail.test.ts` (`tierView` + `pathView`)
- `UI/src/tests/routes/game/screens/proficiencies/TierSpine.test.ts` (`tierView` + `pathView`)
- `UI/src/tests/routes/game/screens/proficiencies/TierCard.test.ts` (`tierView`)
- `UI/src/tests/routes/game/screens/proficiencies/WordOfPowerTooltip.test.ts` (`tierView`)
- `UI/src/tests/routes/game/screens/proficiencies/word-detail.test.ts` (`tierView`)

They are byte-identical apart from `TierSpine`'s `pathOrdinal: o.id` (the others use `0`), and `pathView`'s shape (`WordDetail` takes `Partial<PathView>`, `TierSpine` takes the tier list). `CLAUDE.md` → "prefer writing DRY code".

The concrete cost: #2404 added one field to `TierView` and one to `PathView`, which required editing all five files to keep `svelte-check` green — pure churn with no test value, and the same tax lands on every future view-model field.

This is the same duplication class as #2405, in a different folder. Filing separately rather than widening that issue, since these are different files with different divergences and each is independently reconcilable.

## Why it wasn't folded into #2406

Same reasoning #2405 records for the progression editor: unifying diverged fixtures risks *silently weakening* assertions rather than failing loudly. `TierSpine.test.ts`'s `pathOrdinal: o.id` is load-bearing for its ordering test (`draws the spine most-advanced first`), and the per-file `word`/`pronunciation`/`translation` template defaults feed decipher-reveal assertions in `WordDetail.test.ts` and `WordOfPowerTooltip.test.ts`. Reconciling that is its own review unit, unrelated to rendering a description.

## Suggested direction

Add `UI/src/tests/routes/game/screens/proficiencies/lexicon-test-utils.ts` exporting `tierView(o)`, `pathView(o)`, and `stubController()` (a fourth near-duplicate — `TierCard`, `TierSpine`, and `WordDetail` each hand-roll the same `WordTooltipController` stub).

Constraints worth stating up front:

- **Keep the existing local call signatures** so the ~40 call sites don't churn — import the shared builders under the names the files already use.
- **Reconcile `pathOrdinal` explicitly.** Either default it to `o.id` everywhere (harmless for the four suites that don't read it, and preserves `TierSpine`'s ordering test) or have `TierSpine` pass it per-call. Prefer the former.
- **Do one file at a time, running that suite in between**, so a changed default surfaces as a failure rather than a quietly weaker assertion.
- Vitest only collects `*.{test,spec}.{js,ts}` (`vite.config.ts:8`), so a plain `.ts` helper in that folder won't be picked up as an empty suite.

`WordDetail.test.ts`'s `vi.hoisted` + `vi.mock('$stores', …)` block must stay per-file (`vi.mock` is hoisted within its own file and its factory can't reference imports) — only the fixtures/builders move.

## Scope

ginderjeremiah:
Closes #2407.

Test-only: no production code touched, no behavior change. **193 lines deleted, 91 added** across six files (the shared module plus the five converted suites).

## Summary

Adds `UI/src/tests/routes/game/screens/proficiencies/lexicon-test-utils.ts` and converts the five Lexicon suites onto it. Every suite keeps the local call signature it already used, so no assertion or call site changed — only the fixture declarations moved.

What was consolidated:

| Builder | Copies before | Divergence |
|---|---|---|
| `tierView` | 5 | `TierSpine`'s `pathOrdinal: o.id`; the other four byte-identical |
| `pathView` | 2 | positional-`tiers` vs override-object signature; different `word` default |
| `stubController` | 3 | `describedById` — `tooltip-1` / `tooltip-7` / `tooltip-9` |

The two hand-rolled `IAttribute` builders (`attribute` / `attr`) were also removed, but folded onto the **existing** shared `UI/src/tests/fixtures/attributes.ts` → `makeAttribute` rather than into this module — see [Review round 1](#user-content-review-round-1) below.

## Two corrections to the issue's analysis

Both were checked against the tree rather than taken on faith, and neither changes the recommended outcome:

- **The issue says `pathView` had three copies.** It has two — `WordDetail.test.ts` and `TierSpine.test.ts`. The issue's own file list only names those two.
- **The issue calls `TierSpine`'s `pathOrdinal: o.id` "load-bearing for its ordering test".** It isn't. `TierSpine.svelte:55` derives its render order as `[...path.tiers].reverse()` — array position, not ordinal — and `pathOrdinal` is not read anywhere in the screen (only `proficiencies-lexicon.ts` reads it, off `IProficiency`, when *building* the view-model the suites bypass). The `draws the spine most-advanced first` test passes identically either way.

  The issue's preferred resolution — default it to `o.id` — is still the right call, just for a different reason: it preserves `TierSpine`'s existing values byte-for-byte and keeps a spine built from ascending ids ordinal-coherent, rather than defending a live assertion. That's what the shared builder's comment says.

## Notes on the shared module

- **`stubController` takes the tooltip id as a parameter** (defaulting to `tooltip-1`, `TierSpine`'s value). `TierCard` and `WordDetail` genuinely assert their own ids back out of `aria-describedby`, so this divergence had to survive rather than be reconciled away.
- **`TierSpine` keeps its positional `pathView(tiers, o)` as a one-line adapter** over the shared builder. That preserves its ~6 call sites *and* keeps the root-word derivation (`word: tiers[0]?.word`) at the call site where it's meaningful — a path built from its spine — instead of putting conditional logic in a base fixture. This mirrors the thin-adapter approach #2412 recommends for the sibling progression-map case.
- **`WordDetail.test.ts` keeps its `vi.hoisted` + `vi.mock('$stores', …)` block per-file**, as the issue requires. That stub is unrelated to anything in the shared module, so the module is a plain static import for every suite — it needs none of the "must not transitively reach `$stores`" care that `progression-test-utils.ts` documents, since nothing resolves it from inside a mock factory.

<a id="review-round-1"></a>
## Review round 1 — the `attribute` builder (fixed in 9cbc761)

The first push folded a local `attribute` builder into the new module. That was wrong: it was a **fifth copy** of `UI/src/tests/fixtures/attributes.ts` → `makeAttribute`, which four suites already use — net-new duplication in a PR whose purpose is removing it. Caught in review, removed, and the four call sites now use the shared builder. The module imports `EActivityKey` only.

The differing defaults (`makeAttribute` uses `attributeType: Primary`, `decimals: 0`; the local copy used `Secondary` / `1`) are unreachable here, but for a narrower reason than "the ladder only branches on `isPercentage`":

- `formatAttributeValue` (`attribute-display.ts:150-153`) **does** read `decimals` (`toFixed(attribute.decimals)`). Had the ladder routed through it, `decimals: 0` vs `1` would have been a live behaviour change on the `+2%` assertions.
- It doesn't: `formatModifier` (`word-detail.ts:76-81`) calls **`formatAttributeDelta`**, which reads only `isPercentage` and deliberately uses adaptive precision *instead of* `decimals` (its comment gives the reason — CooldownRecovery's `+0.004` → `+0.4%` must not round away), plus `attributeName`, which reads `name`.
- `WordDetail.svelte:117` routes attributes solely through `buildLadder`, so nothing on the screen reaches the `decimals` path.

## Test plan

- [x] Converted **one file at a time, running that suite in between** (per the issue's process note), so a changed default would fail loudly rather than pass weakly: `WordOfPowerTooltip` (4) → `TierCard` (7) → `word-detail` (18) → `TierSpine` (5) → `WordDetail`.
- [x] `npx vitest run src/tests/routes/game/screens/proficiencies/` — **9 files, 112 tests passed**, before and after the `makeAttribute` swap. The 9 collected files confirm `lexicon-test-utils.ts` is imported, not picked up as an empty suite (`vite.config.ts:8` collects only `*.{test,spec}.{js,ts}`).
- [x] `npm run lint` (eslint + `svelte-check`) — **1222 files, 0 errors, 0 warnings**. This is what pins the unused-import cleanup: `EAttributeType` and the `PathView`/`TierView`/`WordTooltipController` type imports left dangling by the extraction were removed.
- [x] `npm run test:unit:ci` — **312 files, 4009 tests passed**, no regression elsewhere.
- [x] Backend untouched, so no `dotnet` run applies and CI's changed-path gate skips those jobs (`verify-backend` / `test-backend` / `test-hooks` all report `skipped`).
- [x] Full CI green on `9cbc761` — `all-checks-passed: success`, including all three `test-e2e` browsers.

### DIFF
--- a/UI/src/tests/routes/game/screens/proficiencies/lexicon-test-utils.ts
+++ b/UI/src/tests/routes/game/screens/proficiencies/lexicon-test-utils.ts
@@ -1,0 +1,84 @@
+import { vi } from 'vitest';
+import { EActivityKey, EAttribute, EAttributeType, type IAttribute } from '$lib/api';
+import type {
+	PathView,
+	TierView,
+	WordTooltipController
+} from '$routes/game/screens/proficiencies/proficiencies-lexicon';
+
+/* Shared fixtures for the Lexicon (Proficiencies) suites (#2407). Vitest only collects
+   `*.{test,spec}.{js,ts}`, so this plain `.ts` module is imported, never run as an empty suite.
+
+   `WordDetail.test.ts` keeps its own `vi.mock('$stores', …)` — `vi.mock` is hoisted within the file it
+   appears in and its factory cannot close over that file's imports — but the stub it installs is
+   unrelated to anything here, so this module is a plain static import for every suite. */
+
+/**
+ * Base tier fixture — the id is required because the conlang defaults are generated from it, which is
+ * what lets a suite assert on `word1`/`pron1`/`means1` without restating them per call.
+ *
+ * `pathOrdinal` defaults to the id so a spine built from ascending ids is ordinal-consistent. Nothing in
+ * the screen actually reads it today (`TierSpine` orders by array position, reversing `path.tiers`), so
+ * this is about keeping the fixture coherent rather than about any current assertion.
+ */
+export const tierView = (o: Partial<TierView> & { id: number }): TierView => ({
+	name: `Tier ${o.id}`,
+	description: '',
+	pathOrdinal: o.id,
+	level: 0,
+	maxLevel: 10,
+	xp: 0,
+	xpForNext: 100,
+	state: 'unlocked',
+	frontier: false,
+	milestoneLevels: [],
+	levelModifiers: [],
+	levelRewards: [],
+	decipher: 'undeciphered',
+	word: `word${o.id}`,
+	pronunciation: `pron${o.id}`,
+	translation: `means${o.id}`,
+	iconPath: '',
+	...o
+});
+
+/**
+ * Base path fixture — an empty Pyromancy spine.
+ *
+ * A suite building a path *from* tiers derives `word` from the root tier itself (the rail and spine header
+ * reuse the root's word, per `PathView.word`), so that derivation stays at the call site rather than
+ * being guessed at here.
+ */
+export const pathView = (o: Partial<PathView> = {}): PathView => ({
+	id: 0,
+	name: 'Pyromancy',
+	description: '',
+	word: 'aenkor',
+	iconPath: '',
+	activityKey: EActivityKey.Physical,
+	tiers: [],
+	...o
+});
+
+/** The shared word-of-power tooltip controller the spine cards and inspector drive on hover. The id is a
+ *  parameter because suites assert their own value back out of `aria-describedby`. */
+export const stubController = (describedById = 'tooltip-1'): WordTooltipController => ({
+	describedById,
+	show: vi.fn(),
+	move: vi.fn(),
+	hide: vi.fn()
+});
+
+/** An attribute reference row, for the payout-ladder formatters. `isPercentage` picks between a `+2%`-style
+ *  delta and a plain `+5`, which is the only field the ladder branches on. */
+export const attribute = (id: EAttribute, name: string, isPercentage = false): IAttribute => ({
+	id,
+	name,
+	description: '',
+	attributeType: EAttributeType.Secondary,
+	isPercentage,
+	isHarmful: false,
+	code: '',
+	displayOrder: 0,
+	decimals: 1
+});
--- a/UI/src/tests/routes/game/screens/proficiencies/lexicon-test-utils.ts
+++ b/UI/src/tests/routes/game/screens/proficiencies/lexicon-test-utils.ts
@@ -1,5 +1,5 @@
 import { vi } from 'vitest';
-import { EActivityKey, EAttribute, EAttributeType, type IAttribute } from '$lib/api';
+import { EActivityKey } from '$lib/api';
 import type {
 	PathView,
 	TierView,
@@ -67,18 +67,4 @@ export const stubController = (describedById = 'tooltip-1'): WordTooltipControll
 	show: vi.fn(),
 	move: vi.fn(),
 	hide: vi.fn()
-});
-
-/** An attribute reference row, for the payout-ladder formatters. `isPercentage` picks between a `+2%`-style
- *  delta and a plain `+5`, which is the only field the ladder branches on. */
-export const attribute = (id: EAttribute, name: string, isPercentage = false): IAttribute => ({
-	id,
-	name,
-	description: '',
-	attributeType: EAttributeType.Secondary,
-	isPercentage,
-	isHarmful: false,
-	code: '',
-	displayOrder: 0,
-	decimals: 1
 });
--- a/UI/src/tests/routes/game/screens/proficiencies/TierCard.test.ts
+++ b/UI/src/tests/routes/game/screens/proficiencies/TierCard.test.ts
@@ -1,37 +1,10 @@
 import { describe, it, expect, afterEach, vi } from 'vitest';
 import { render, cleanup, screen, fireEvent, within } from '@testing-library/svelte';
 import TierCard from '$routes/game/screens/proficiencies/TierCard.svelte';
-import type { TierView, WordTooltipController } from '$routes/game/screens/proficiencies/proficiencies-lexicon';
+import type { TierView } from '$routes/game/screens/proficiencies/proficiencies-lexicon';
+import { stubController, tierView } from './lexicon-test-utils';
 
-const tierView = (o: Partial<TierView> & { id: number }): TierView => ({
-	name: `Tier ${o.id}`,
-	description: '',
-	pathOrdinal: 0,
-	level: 0,
-	maxLevel: 10,
-	xp: 0,
-	xpForNext: 100,
-	state: 'unlocked',
-	frontier: false,
-	milestoneLevels: [],
-	levelModifiers: [],
-	levelRewards: [],
-	decipher: 'undeciphered',
-	word: `word${o.id}`,
-	pronunciation: `pron${o.id}`,
-	translation: `means${o.id}`,
-	iconPath: '',
-	...o
-});
-
-const stubController = (): WordTooltipController => ({
-	describedById: 'tooltip-9',
-	show: vi.fn(),
-	move: vi.fn(),
-	hide: vi.fn()
-});
-
-const renderCard = (tier: TierView, controller = stubController(), selected = false, last = true) => {
+const renderCard = (tier: TierView, controller = stubController('tooltip-9'), selected = false, last = true) => {
 	const onSelect = vi.fn();
 	render(TierCard, { tier, selected, last, onSelect, controller });
 	return { onSelect, controller };
--- a/UI/src/tests/routes/game/screens/proficiencies/TierSpine.test.ts
+++ b/UI/src/tests/routes/game/screens/proficiencies/TierSpine.test.ts
@@ -1,51 +1,13 @@
 import { describe, it, expect, afterEach, vi } from 'vitest';
 import { render, cleanup, screen } from '@testing-library/svelte';
-import { EActivityKey } from '$lib/api';
 import TierSpine from '$routes/game/screens/proficiencies/TierSpine.svelte';
-import type {
-	PathView,
-	TierView,
-	WordTooltipController
-} from '$routes/game/screens/proficiencies/proficiencies-lexicon';
+import type { PathView, TierView } from '$routes/game/screens/proficiencies/proficiencies-lexicon';
+import { pathView as basePathView, stubController, tierView } from './lexicon-test-utils';
 
-const tierView = (o: Partial<TierView> & { id: number }): TierView => ({
-	name: `Tier ${o.id}`,
-	description: '',
-	pathOrdinal: o.id,
-	level: 0,
-	maxLevel: 10,
-	xp: 0,
-	xpForNext: 100,
-	state: 'unlocked',
-	frontier: false,
-	milestoneLevels: [],
-	levelModifiers: [],
-	levelRewards: [],
-	decipher: 'undeciphered',
-	word: `word${o.id}`,
-	pronunciation: `pron${o.id}`,
-	translation: `means${o.id}`,
-	iconPath: '',
-	...o
-});
-
-const pathView = (tiers: TierView[], o: Partial<PathView> = {}): PathView => ({
-	id: 0,
-	name: 'Pyromancy',
-	description: '',
-	word: tiers[0]?.word ?? '',
-	iconPath: '',
-	activityKey: EActivityKey.Physical,
-	tiers,
-	...o
-});
-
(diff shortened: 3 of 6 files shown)


UI/src/tests/routes/game/screens/proficiencies/lexicon-test-utils.ts
```diff
@@ -0,0 +1,84 @@
...
+	id,
+	name,
+	description: '',
+	attributeType: EAttributeType.Secondary,
+	isPercentage,
+	isHarmful: false,
+	code: '',
+	displayOrder: 0,
+	decimals: 1
+});
```
ginderjeremiah: **Blocking** — this is a fifth copy of a builder that already has a canonical shared home. `UI/src/tests/fixtures/attributes.ts` exports `makeAttribute(id, name, overrides)`, and four suites already use it — including the exact percentage case here (`SkillTooltip.test.ts:85` → `makeAttribute(EAttribute.CriticalChanceMultiplier, 'Critical Chance', { isPercentage: true })`).

The `attribute` fold-in was scope this PR added voluntarily (#2407 named only `tierView` / `pathView` / `stubController`), so a PR whose whole purpose is DRY shouldn't ship net-new duplication of an existing helper — `CLAUDE.md` → "Prefer writing DRY code whenever possible."

The swap is mechanical and the default divergence is inert: `formatModifier` reaches `attribute-display.ts:160-163`, which reads **only** `isPercentage` (`attributeName` reads `name`), so `makeAttribute`'s `attributeType: Primary` / `decimals: 0` defaults can't reach an assertion — which the comment above already establishes ("the only field the ladder branches on").

```ts
// delete this export, then at the 4 call sites:
makeAttribute(EAttribute.CriticalChanceMultiplier, 'Fire Damage', { isPercentage: true })
makeAttribute(EAttribute.Toughness, 'Toughness')
```

Call sites: `word-detail.test.ts:22-23`, `WordDetail.test.ts:44-45`. Dropping it also takes `EAttribute`, `EAttributeType`, and `IAttribute` out of this module's imports, leaving it to import `EActivityKey` only.

---

ginderjeremiah (commented):
## Review: consolidating the five duplicated Lexicon fixture builders into `lexicon-test-utils.ts` (#2407)

**Verdict:** Request changes — the consolidation itself is clean and correct, but it introduces a *new* duplicate of an existing shared fixture (`makeAttribute`), which is the one thing a dedup PR shouldn't do. (Submitted as a comment because GitHub won't accept a gating verdict from the PR author.)

### Blocking

- **`lexicon-test-utils.ts:72-84` — the new `attribute` builder duplicates `UI/src/tests/fixtures/attributes.ts` → `makeAttribute`.** That module is the established cross-folder home for `IAttribute` fixtures and is already used by four suites, including the identical `{ isPercentage: true }` case at `SkillTooltip.test.ts:85`. Verified the swap is inert: `formatModifier` → `attribute-display.ts:160-163` reads only `isPercentage`, so `makeAttribute`'s differing `attributeType` / `decimals` defaults can't reach an assertion. Details and the 4 call sites in the inline comment.

### Nits

- None.

### Verification I ran against the tree (rather than taking the description on faith)

- **The `pathOrdinal: 0 → o.id` reconciliation is genuinely inert**, and the PR's correction to the issue is right: `pathOrdinal` is read *only* by `proficiencies-lexicon.ts:189,261` off `IProficiency` while **building** the view-model — which all five suites bypass — plus the workbench editor, which uses its own fixtures. No `TierView.pathOrdinal` consumer exists in `TierCard` / `TierSpine` / `WordDetail` / `WordOfPowerTooltip` / `word-detail.ts`. `TierSpine.svelte:54-55` orders by array position (`[...path.tiers].reverse()`), so `draws the spine most-advanced first` is unaffected either way. The issue's "load-bearing for its ordering test" claim is wrong; the PR is right to say so and right to keep `o.id` anyway for fixture coherence.
- **`TierSpine`'s adapter is behaviour-preserving.** `basePathView({ word: tiers[0]?.word ?? '', tiers, ...o })` yields exactly the old object for every call site, including override precedence — `o.word` still wins over the derived root word.
- **The consolidation is complete.** No `tierView` / `pathView` / `stubController` / `attribute` copies remain anywhere under `tests/`; the only local survivor is `TierSpine`'s intentional one-line signature adapter.
- **`stubController`'s parameterised id is the right call**, not a smell — `TierCard` and `WordDetail` each assert their own value back out of `aria-describedby`, so the parameter keeps each assertion coupled to what its suite passed in.

### Test coverage

Test-only change, no production code touched, so the bar is "no assertion silently weakened" — met. All 112 assertions across the five suites are byte-identical to before; only fixture declarations moved. The one-file-at-a-time process the issue asked for was followed. `verify-frontend` (eslint + `svelte-check`) is green, which is what pins the dead-import cleanup (`EAttributeType`, and the `PathView`/`TierView`/`WordTooltipController` type imports left dangling by the extraction).

### Follow-up

- **#2414** — filed for the sibling duplication class this PR correctly left alone: five hand-rolled `IProficiency` builders across four folders (two of them in this very folder, at `Proficiencies.test.ts:30` and `proficiencies-view.test.ts:41`). Different type, different layer, spans folders — its own review unit, not scope for this PR.

### What's good

- Correcting two claims in the issue rather than implementing them on faith, and keeping the recommended outcome while documenting the *actual* reason, is exactly right.
- Keeping `TierSpine`'s positional signature as a thin adapter — rather than pushing `word: tiers[0]?.word` conditional logic into the base fixture — puts the derivation where it's meaningful and saves ~6 call sites of churn.
- The comment on `tierView` explaining that `pathOrdinal` is coherence rather than a live assertion is the kind of note that stops a future editor from "simplifying" it back to `0`. Worth the lines.

---

UI/src/tests/routes/game/screens/proficiencies/lexicon-test-utils.ts
ginderjeremiah: Fixed in 9cbc761 — you're right, and it's the one mistake this PR had no excuse for. Shipping a fifth copy of an existing helper in a PR whose entire purpose is removing duplication is self-defeating, and it happened precisely *because* the fold-in was voluntary scope: I checked the four builders the issue named for prior art and never asked the same question of the one I added.

Removed the local `attribute` and swapped the four call sites onto `makeAttribute`. The module now imports `EActivityKey` only, as predicted.

I re-derived the inertness rather than taking it on faith, and it holds for a slightly different reason than the comment gives — worth recording since it's the part that could have bitten:

- `formatModifier` (`word-detail.ts:76-81`) doesn't reach `attribute-display.ts:160-163` (`formatAttributeValue`). It calls **`formatAttributeDelta`** (`:160+`) and `attributeName`. That matters because `formatAttributeValue` **does** read `decimals` (`toFixed(attribute.decimals)`), so if the ladder had routed through it, `makeAttribute`'s `decimals: 0` vs the old `1` would have been a live behaviour change on the `+2%` assertions.
- `formatAttributeDelta` reads only `isPercentage` and explicitly uses adaptive precision *instead of* `decimals` — its own comment gives the reason (CooldownRecovery's `+0.004` → `+0.4%` must not round away).
- Confirmed nothing else on the screen reaches the `decimals` path: `WordDetail.svelte:117` routes attributes solely through `buildLadder`.

So the conclusion is the same — `attributeType` and `decimals` are unreachable — but it rests on which formatter the ladder picks, not on `formatAttributeValue` ignoring those fields.

Verification after the swap: proficiencies folder **9 files / 112 tests** green (byte-identical assertions), `npm run lint` **1222 files, 0 errors, 0 warnings**.

Thanks for #2414 on the `IProficiency` builders — agreed that's a separate review unit, and I've left it alone here.

---

ginderjeremiah:
## Re-review of `9cbc761` — blocking finding cleared

**Verdict:** Approve — merging. CI green on the head (`all-checks-passed`, including all three `test-e2e` browsers; backend jobs correctly skipped by the changed-path gate). No outstanding nits.

Incremental pass over the delta only (`89cfc27..9cbc761` — three files, 9 added / 21 removed), verified against the tree rather than from the reply.

### The blocking finding

**Cleared.** The local `attribute` builder is gone and the four call sites (`word-detail.test.ts:22-23`, `WordDetail.test.ts:44-45`) now use `makeAttribute`. The module imports `EActivityKey` only, exactly as predicted, and the relative-import depth matches the eight existing importers at the same level (`'../../../../fixtures/attributes'` — same form as `SkillTooltip.test.ts:17` and `Attributes.test.ts:4`).

**The reply's correction to my inertness argument is right, and it's the more careful reading.** I cited `attribute-display.ts:160-163` as if the ladder reached `formatAttributeValue`; it doesn't. `word-detail.ts:10` imports `attributeName`, `formatAttributeDelta` and `formatNum` — and *not* `formatAttributeValue`, which is the only formatter in that module that reads `decimals` (`:151` → `toFixed(attribute.decimals)`). `formatAttributeDelta` (`:161-163`) reads `isPercentage` alone and deliberately uses adaptive precision instead, per its own comment at `:157-158`. Confirmed nothing else on the screen reaches the other path: `formatAttributeValue` appears nowhere under `routes/game/screens/proficiencies/`.

So the conclusion holds — `attributeType: Primary` and `decimals: 0` are both unreachable from these suites — but on the narrower ground that the ladder picks `formatAttributeDelta`, not that the formatter it picks ignores those fields. Worth having on the record: had `buildLadder` routed through `formatAttributeValue`, the `1 → 0` decimals change would have been live on the `+2%` assertions.

### Verification

- `npx vitest run src/tests/routes/game/screens/proficiencies/` — **9 files, 112 tests passed**, matching the pre-swap count, so no assertion was weakened or dropped by the substitution.
- **Consolidation is still complete after the delta.** The only local builders left anywhere under `tests/` are `TierSpine.test.ts:9`'s intentional positional adapter and `RewardAffordance.test.ts:66`'s `stubController`, which is a different type (`RewardTooltipController`) in a different folder — not this PR's dedup class.
- `verify-frontend` green on the head pins the dead-import cleanup that the removal left behind (`EAttribute`, `EAttributeType`, `IAttribute`).

### Follow-up

- **#2414** (filed in round 1) covers the sibling `IProficiency` builder duplication this PR correctly left alone. Nothing new to file from this round.

### Standing

Test-only, no production code touched, assertions byte-identical to before. The round-1 nits list was empty and stays empty. Merging.

---

Issue #2419, opened 2026-07-25
disconnect() doesn't cancel an in-flight openSocket, so a teardown can be followed by a resurrected "zombie" socket


ginderjeremiah:
## Problem

`ApiSocket.disconnect` (`UI/src/lib/api/api-socket.ts:278-284`) stops the ping interval and closes `this.socket` if present — but it doesn't touch the in-flight `connecting` promise, and `openSocket` (`api-socket.ts:143-212`) has no cancelled/disconnected re-check after its `await ensureValidAccessToken()`: once the refresh await resolves it unconditionally constructs the WebSocket and wires handlers.

## Failure scenario

Session takeover: the server closes the old socket, the 10s keepalive starts a reconnect while the `SocketReplaced` modal awaits acknowledgement, the user clicks through → `stopGame()` → `disconnect()` lands mid-connect → the connect completes anyway (tokens are still valid — takeover doesn't clear them), `onStart` flushes the stale queue and starts pinging. That's exactly the "silently reconnect and fight the session that just took over" outcome `disconnect()`'s own doc comment says it exists to prevent — now with **no keepalive owner** ever tearing the socket down (disconnect already stopped the interval, and openSocket only re-arms it at its start). Window ≈ one connect duration per 10s ping cycle while the modal is up.

## Fix

A `disconnected` flag set by `disconnect()` and checked in `openSocket` after the refresh await (bail + close if superseded), cleared when `ensureSocket`/`sendSocketCommand` legitimately starts a new session. Pin with a unit test: disconnect during the openSocket refresh await must not leave an open socket behind.

<sub>From the scheduled codebase health review.</sub>

PR #2420, opened 2026-07-25
Settle queued socket commands when the keepalive shuts down (#2418)

### LINKED ISSUE #2418: attemptPing's token-less shutdown strands queued socket commands — the one terminal path that never settles the queue
## Problem

`ApiSocket.attemptPing` (`UI/src/lib/api/api-socket.ts:244-251`) begins with:

```ts
if (!getAccessToken()) {
    this.stopPingInterval();
    return;
}
```

Stopping the keepalive kills the only mechanism that would ever re-flush `socketCommandQueue` — but unlike **every other terminal branch** (`handleClose`'s NORMAL_CLOSURE / budget-exhausted / no-refresh-token branches, `openSocket`'s rejected-refresh and malformed-token branches, which all call `settleQueuedRequests(...)`), this one returns without settling. That violates the documented "pending and queued commands are always settled rather than left hanging when a close is terminal" invariant (`frontend.md` → Socket request & connection lifecycle) — the same invariant class as closed #2035/#1766/#889.

## Failure scenario

Reachable because `getAccessToken()` reads `localStorage` live and two tabs are a supported state: tab A's socket drops post-open (queue deliberately kept for the keepalive reconnect), tab B logs out or hits `handleAuthFailure` and clears the shared tokens, tab A's next ping tick silently stops the keepalive — and a caller awaiting a queued command hangs forever. Since `EnemyManager.claimVictory()` awaits `DefeatEnemy` unconditionally, the idle loop wedges silently instead of surfacing "connection lost".

## Fix

Settle the queued requests (and check the in-flight map) with `CONNECTION_LOST_ERROR` in that branch, mirroring the other terminal paths, and pin it with a unit test alongside the existing settle-invariant tests.

ginderjeremiah:
Closes #2418.

## Summary

`ApiSocket.attemptPing`'s token-less branch stopped the keepalive interval and returned without settling `socketCommandQueue`. Since the keepalive is the only thing that reconnects and re-flushes that queue, this left queued commands hanging — violating the "pending and queued commands are always settled rather than left hanging when a close is terminal" invariant (`frontend.md` → _Socket request & connection lifecycle_), the same class as #2035/#1766/#889.

The reproduction in the issue holds: tab A's socket drops post-open (`handleClose` deliberately keeps the queue for the keepalive to re-flush), tab B logs out and clears the shared `localStorage` tokens, tab A's next ping tick silently stops the keepalive — and a caller awaiting a queued command hangs forever. `EnemyManager.claimVictory()` awaits `DefeatEnemy` unconditionally, so the idle loop wedges silently instead of surfacing "connection lost".

Review then established that this is **not one branch but four**: three other exits also leave the queue for a keepalive that `attemptPing` may already have torn down. The fix generalises accordingly.

## Changes

**`UI/src/lib/api/api-socket.ts`**

- `attemptPing`'s token-less branch now calls `settleQueuedRequests(CONNECTION_LOST_ERROR)`, mirroring every other terminal branch — **but only when no socket remains to flush the queue** (`hasFlushableSocket()`). That carve-out is load-bearing rather than defensive: `openSocket` deliberately supports a **token-less anonymous session**, so `getAccessToken()` is legitimately null on every ping tick while that socket connects, and an unconditional settle would kill commands that were about to be sent on open. The existing test `does not time out a queued-but-unsent command while the socket is still connecting` fails on the unconditional version, which is how this surfaced.
- **New: `settleQueueIfKeepaliveStopped()` at each keepalive-deferring exit.** Deferring the queue to the keepalive is only valid while the keepalive is running, and three exits assumed it unconditionally: `handleClose`'s `socketOpened` and demonstrably-fresh-token branches, and `openSocket`'s retryable bail (which never creates a socket at all). Each now re-checks. This makes `hasFlushableSocket()`'s justification true by construction instead of by assumption, and preserves the anonymous-session carve-out — that socket either opens and flushes, or closes into a settling branch.
- Extracted `hasUnclosedSocket()` rather than adding a fourth copy of the `socket && socket.readyState !== socket.CLOSED` expression, and dropped `disconnect()`'s now-redundant `readyState` guard (`close()` on a `CLOSED` socket is a spec no-op; on a `CONNECTING` one it aborts the handshake, which is the intent).

**In-flight requests are deliberately not touched**, contrary to the issue's "(and check the in-flight map)". An in-flight entry only ever exists alongside a socket that can still answer it: `processCommandQueue` populates the map only after the socket is `OPEN`, and `handleClose` settles the whole map on *every* close. So in the reachable case the map is empty, and in the still-open case the socket can genuinely still reply — settling there would break live requests on a healthy connection. The 30s per-request timeout remains the backstop. Called out in a code comment.

**`docs/frontend-sockets.md`** — extended the keepalive-ownership bullet with what happens to the queue on the ping's self-termination, plus the corollary the review's blocking item exposed: deferring to the keepalive is only valid while it is running.

## Reachability of the three deferring exits

Recorded because it determines what is worth testing:

| Exit | Reachable with a non-empty queue? |
|---|---|
| Fresh-token rejection | **Yes** — #2418's own two-tab scenario with the ping landing while the socket is `CONNECTING`. Tested. |
| `openSocket` retryable bail | **Yes**, but narrower than it looks: `hadRefreshToken` is read *before* the `ensureValidAccessToken()` await, so the logout must land *during* that await (an earlier one makes it false and the connect proceeds as anonymous). Tested by holding the refresh open across the ping tick. |
| `handleClose`'s `socketOpened` | **No natural path found.** `processCommandQueue` drains on `OPEN`, and that drain is a microtask while a close event is a macrotask, so the queue is empty by the time the branch runs. Guarded for uniformity — the branch shouldn't rest on that timing argument holding forever — but no contrived test. |

## Test plan

- [x] Three new unit tests in the existing `queued-but-unsent settlement on a terminal close` block: the keepalive shutdown itself (issue's exact sequence), the fresh-token rejection outliving the keepalive, and `openSocket`'s retryable bail outliving it. Each asserts the command resolves with `Connection lost. Please try again.` and that no socket is resurrected.
- [x] All three vacuity-checked by stripping the corresponding source guard — each then **hangs and fails on its 5s timeout**, which is the stranding being fixed.
- [x] The anonymous/connecting carve-out is pinned by the pre-existing `does not time out a queued-but-unsent command while the socket is still connecting` test — it fails without `hasFlushableSocket()`.
- [x] `npm run lint` — clean (eslint + `svelte-check`: 1222 files, 0 errors, 0 warnings).
- [x] `npm run test:unit:ci` — 312 files / 4012 tests passing, coverage thresholds met.
- [x] Backend untouched (frontend transport only), no battle-logic parity surface, so no mirrored backend test is required.

## Notes / follow-ups

No new follow-up issues filed. The one adjacent problem noticed while reading `api-socket.ts` — `disconnect()` not cancelling an in-flight `openSocket`, leaving a "zombie" socket — is already tracked as #2419 and is deliberately untouched here to keep the two fixes separately reviewable.


UI/src/lib/api/api-socket.ts
```diff
@@ -278,11 +286,24 @@ export class ApiSocket {
...
+	private hasUnclosedSocket(): boolean {
+		return this.socket !== undefined && this.socket.readyState !== this.socket.CLOSED;
+	}
+
+	/** True while something is still positioned to send the queued commands: a socket already open
+	 *  (`processCommandQueue` drains it), one still connecting (`onStart` flushes it on open, and every
+	 *  close it can take instead settles the queue in `handleClose`), or an open still in flight. */
+	private hasFlushableSocket(): boolean {
+		return this.connecting !== null || this.hasUnclosedSocket();
+	}
```
ginderjeremiah, line 343: **Blocking — the correctness argument for this predicate doesn't hold: two of `handleClose`'s branches deliberately *don't* settle and delegate to the keepalive that `attemptPing` has just stopped.**

The doc comment's load-bearing claim is *"one still connecting (`onStart` flushes it on open, and **every close it can take instead settles the queue in `handleClose`**)"*. That isn't true — `handleClose` has exactly two non-settling exits, and both of them defer to the keepalive:

- `wasHandshakeTokenDemonstrablyFresh()` (`api-socket.ts:502-512`) — *"the keepalive ping's next `ensureSocket()` retries the connect"*.
- `socketOpened` (`:496-500`) — *"the keepalive ping will reconnect and re-flush anything queued since"*.

And `openSocket` has a third, covered by the `this.connecting !== null` half of this predicate: the retryable bail at `:163-168` — *"the keepalive ping's next `ensureSocket()` call retries the connect"* — which returns without ever creating a socket, so no close ever happens.

The fresh-token one is reachable from **the exact two-tab scenario in #2418**, just with the ping landing a moment earlier:

1. Logged-in tab A queues a command; `openSocket` presents a token with > `FRESH_TOKEN_MARGIN_SECONDS` of life; socket sits `CONNECTING` (a proxy stripping `Upgrade` can hang the handshake for seconds — that's the case `FRESH_TOKEN_MARGIN_SECONDS` exists for).
2. Tab B logs out, clearing the shared `localStorage` tokens.
3. Ping tick: no token → `stopPingInterval()`; `hasUnclosedSocket()` is true (CONNECTING) → **queue not settled**.
4. Handshake fails → `handleClose` → demonstrably-fresh branch → returns **without settling**, expecting a keepalive that no longer exists.
5. `EnemyManager.claimVictory()` hangs forever — the original symptom, unfixed.

The `openSocket` retryable-bail variant is the same shape (ping tick lands while `connecting !== null`, refresh fails retryably, the bail leaves the queue for a keepalive that's gone).

Suggest closing it with the general rule rather than widening this predicate, since the invariant is really *"if you're deferring to the keepalive, the keepalive has to still be running"* — e.g. at each of the three deferring exits:

```ts
// Deferring to the keepalive's reconnect only works while it's running; a concurrent logout can have
// torn it down (attemptPing's token-less branch), which makes this exit terminal after all.
if (this.pingIntervalId === null) {
	this.settleQueuedRequests(CONNECTION_LOST_ERROR);
}
```

With that in place `hasFlushableSocket()`'s claim becomes true as written, and the anonymous-session carve-out this PR correctly identified still holds (that socket either opens and flushes, or closes into a settling branch). Worth a unit test on the fresh-token path specifically, since it's the reachable one.

---

UI/src/lib/api/api-socket.ts
```diff
@@ -278,11 +286,24 @@ export class ApiSocket {
 	public disconnect() {
 		this.stopPingInterval();
 		this.clearConnectionStableTimer();
-		if (this.socket && this.socket.readyState !== this.socket.CLOSED) {
-			this.socket.close(NORMAL_CLOSURE);
+		if (this.hasUnclosedSocket()) {
+			this.socket?.close(NORMAL_CLOSURE);
 		}
```
ginderjeremiah: **Nit** — the extraction costs the flow-narrowing the inline check had, which is why `close` needed an optional chain it didn't need before. Since `close()` on an already-`CLOSED` socket is a spec no-op (and on a `CONNECTING` one it aborts the handshake, which is what we want here anyway), the guard buys nothing now:

```ts
this.socket?.close(NORMAL_CLOSURE);
```

Good call avoiding `!` per CLAUDE.md either way — just noting the guard + `?.` pair reads as if one of them is doing work.

---

UI/src/tests/lib/api/api-socket.test.ts
```diff
@@ -1041,6 +1041,36 @@ describe('ApiSocket', () => {
...
+			vi.useFakeTimers();
+			try {
+				setTokens({ accessToken: 'a', refreshToken: 'r' });
+
+				const promise = queueOnConnectingSocket(apiSocket);
+				await flushMicrotasks();
+				const ws = lastWs();
+				// Open then drop: handleClose deliberately keeps the queue for the keepalive to re-flush, so
+				// this is the state in which the keepalive is the queue's only remaining owner.
+				ws.onopen?.();
```
ginderjeremiah, line 1064: **Nit** — `onopen` fires here while the mock's `readyState` is still `0` (CONNECTING, from `queueOnConnectingSocket`), which is what keeps the command queued through `onStart`'s `processCommandQueue` and makes the assertion below hold. That's deliberate and load-bearing, but it looks like an oversight — a later reader "fixing" it to `ws.readyState = ws.OPEN; ws.onopen?.()` would flush the queue and turn this into a vacuous test. Worth one clause on the existing comment saying the socket is left CONNECTING on purpose so the command stays unsent.

---

ginderjeremiah (commented):
## Review: settle queued socket commands when `attemptPing`'s token-less branch tears down the keepalive (#2418)

**Verdict:** Request changes — the headline path is fixed correctly, but the guard that makes it safe (`hasFlushableSocket()`) rests on a claim about `handleClose` that isn't true, and #2418's own two-tab scenario still strands the queue on a reachable variant. (Submitted as a comment because GitHub won't accept a gating verdict on my own PR.)

The diagnosis and the shape of the fix are right, and the `hasFlushableSocket()` carve-out is a genuine catch rather than defensive padding — I verified that `does not time out a queued-but-unsent command while the socket is still connecting` (`api-socket.test.ts:479`) runs with `localStorage` cleared, so every ping tick in its 60s advance hits the token-less branch; an unconditional settle really would fail it. Declining the issue's "check the in-flight map" is also correct: `processCommandQueue` populates the map only on an `OPEN` socket and `handleClose` settles it on every close, so settling there would only ever break live requests.

### Blocking

- **`api-socket.ts:300-305` — `hasFlushableSocket()`'s justification doesn't hold.** Its comment asserts that a still-connecting socket is safe because "every close it can take instead settles the queue in `handleClose`". `handleClose` has two exits that don't settle (`:496-500` `socketOpened`, `:502-512` demonstrably-fresh token), and `openSocket:163-168` has a third that never creates a socket at all — all three delegate to the keepalive that `attemptPing` just stopped. The fresh-token one is reachable from #2418's own scenario with the ping tick landing while the socket is `CONNECTING` instead of after it closed, and it ends in the same silently-wedged `claimVictory()`. Details and a suggested three-line rule ("only defer to the keepalive while `pingIntervalId !== null`") in the inline comment. This is the invariant at `frontend.md:50`, so I'd rather close it here than file it — it's the same bug this PR exists to fix, and the fix is small.

### Nits

- `api-socket.ts:289-291` — the guard plus `?.` in `disconnect()`; `close()` on a `CLOSED` socket is already a no-op, so `this.socket?.close(NORMAL_CLOSURE)` alone would do.
- `api-socket.test.ts:1053-1055` — the mock is left `CONNECTING` while `onopen` fires, deliberately and load-bearingly; worth saying so before someone "fixes" it into a vacuous test.

### Test coverage

Good, and the right shape: the new test lands in the existing `queued-but-unsent settlement on a terminal close` block alongside its four siblings, reproduces the issue's sequence rather than poking internals, and asserts the three things that matter (queue settled with `CONNECTION_LOST_ERROR`, interval cleared, no socket resurrected). The stash-the-source vacuity check in the test plan is the right instinct. Frontend transport only, no battle-logic parity surface, so no mirrored backend test — agreed.

The one gap follows from the blocking item: the fresh-token / retryable-bail variants have no coverage, and they're the paths that still strand.

### What's good

- The `hasFlushableSocket()` correction to the issue's suggested fix, and refusing to touch the in-flight map with an explicit argument for why — both correct, and both the kind of thing a literal reading of the issue would have gotten wrong.
- `hasUnclosedSocket()` extraction instead of a fourth copy of the `readyState !== CLOSED` expression.
- `frontend-sockets.md` extended in place rather than appended to, keeping the terminal-settlement enumeration accurate; it'll need the anonymous-session clause tightened slightly once the blocking item is addressed, since "unless a socket is still around to flush them" becomes true by construction rather than by assumption.

---

UI/src/lib/api/api-socket.ts
ginderjeremiah: Confirmed and fixed in 46d2126 — you're right, and I verified all three exits against the tree rather than taking the walkthrough on trust. My comment's claim was simply false: `handleClose`'s `socketOpened` and demonstrably-fresh branches both `return` without settling, and `openSocket`'s retryable bail never creates a socket at all. All three say "the keepalive will retry" in their own comments, which is exactly the assumption `attemptPing` had just invalidated.

Took the general rule as suggested rather than widening the predicate, via a shared `settleQueueIfKeepaliveStopped()` at each of the three exits. `hasFlushableSocket()`'s justification is now true by construction, and its comment says so instead of asserting it.

**On reachability**, my own tracing agrees with yours on two of the three and not the third:

- **Fresh-token** — reachable exactly as you describe. Covered by `settles an unsent queued command when a fresh-token rejection outlives the keepalive`.
- **Retryable bail** — reachable, but only through a narrower window than the summary implies: `hadRefreshToken` is read *before* the `ensureValidAccessToken()` await, so the logout has to land during that await for the bail to be taken at all (a logout that completes earlier makes `hadRefreshToken` false and the connect proceeds as anonymous). The new test drives precisely that interleaving by holding the refresh open across the ping tick.
- **`socketOpened`** — I could not construct a natural strand here, and I don't think one exists: `processCommandQueue` drains the queue whenever the socket is `OPEN`, and the drain is a microtask while a close event is a macrotask, so the queue is empty by the time this branch runs. It gets the same guard for uniformity (the branch shouldn't rely on that timing argument holding forever), but no dedicated test, since the only way to reach it with a non-empty queue is the artificial `readyState = CONNECTING` + `onopen` state the other test uses.

Both new tests were vacuity-checked the same way as the original: with the guards stripped, each hangs and fails on its 5s timeout.

---

ginderjeremiah:
Merged `main` (`7a40a7f`) after #2421 — the #2419 zombie-socket fix — landed on the same lifecycle. **Flagging because the resolution changes a line #2421 introduced**, so it's worth a look rather than being a routine conflict resolution.

The textual conflict was trivial (one `frontend-sockets.md` bullet both sides edited). The source files auto-merged **cleanly but not correctly**:

#2421 added an `ensurePingInterval()` re-assert after `openSocket`'s refresh await, which git placed directly above this branch's retryable-bail guard. The result is that the guard's own precondition is destroyed one line before it runs — the ping tick tears the keepalive down mid-await, `ensurePingInterval()` re-arms it, and `settleQueueIfKeepaliveStopped()` then sees a running keepalive and declines to settle. `settles an unsent queued command when openSocket's retryable bail outlives the keepalive` caught it; the merge is green in CI terms but the fix was silently defeated.

The re-assert is right for the case it was written for (a teardown landing and then being lifted by a new `sendSocketCommand`) and wrong for a logged-out one. On the logout path it:

1. revives the post-logout ping `attemptPing`'s token-less branch exists specifically to kill,
2. defers the queue to a keepalive that stops again on its very next tick — settling ~10s late rather than immediately, and
3. **strands the queue outright if `disconnect()` lands in that window**: the retryable bail created no socket, so there is no `handleClose` to reach the `NORMAL_CLOSURE` settle, and `disconnect()` doesn't settle the queue itself.

Resolved by skipping the re-assert when the session's tokens are gone, which is the condition that distinguishes the two cases. Anonymous sessions are unaffected — their keepalive already self-terminates on the first tick, both before and after this change — and #2421's own teardown tests pass unmodified.

`docs/frontend-sockets.md` carries both sides' bullets plus a clause noting the re-assert is not a blanket revival.

Full suite green after the merge: 4016 tests, lint and `svelte-check` clean.

---

ginderjeremiah:
## Re-review of `46d2126` + `7a40a7f` — blocking item closed, and the merge resolution is load-bearing

**Verdict:** Approve. CI green on the head commit (`verify-frontend`, `test-frontend`, all three `test-e2e` browsers, `all-checks-passed`). No blocking items, no should-fixes, no nits. **Merging.**

Incremental pass over two deltas: the fix for the blocking item (`46d2126`, which got an inline reply but never a verdict) and the `main` merge (`7a40a7f`). I checked out the branch and ran things rather than reading the diff, since the merge comment makes a claim about a silently-defeated fix that is only worth as much as its vacuity check.

### The blocking item is properly closed

The general rule was the right shape, and `settleQueueIfKeepaliveStopped()` at each of the three deferring exits is better than widening `hasFlushableSocket()` — it makes that predicate's doc comment true by construction, which was exactly the defect. The comment now says *"directly, or via `settleQueueIfKeepaliveStopped` on the branches that would otherwise defer to the keepalive"* instead of asserting the false version.

Your reachability correction on the retryable bail is right and worth keeping: `hadRefreshToken` is read at `api-socket.ts:169`, *before* the `ensureValidAccessToken()` await, so a logout that completes earlier makes it `false` and the connect proceeds as anonymous. The bail is only taken if the logout lands *during* that await, and the test drives precisely that by holding the refresh open across the ping tick. Declining a contrived test for `socketOpened` while still guarding it is the right call — the drain-is-a-microtask / close-is-a-macrotask argument is correct today but isn't something the branch should rest on.

### The merge resolution — verified load-bearing

The claim in [your merge comment](https://github.com/ginderjeremiah/GameServer/pull/2420#issuecomment-5078262350) is that #2421's `ensurePingInterval()` re-assert auto-merged cleanly one line above this branch's retryable-bail guard and silently defeated it. Confirmed by reverting just the guard to the unconditional `this.ensurePingInterval()`:

```
× settles an unsent queued command when openSocket's retryable bail outlives the keepalive 5004ms
Tests  1 failed | 66 passed (67)
```

It hangs and fails on its 5s timeout — the stranding, not a changed assertion. So the resolution is doing real work, and the auto-merge really was green-but-wrong.

**The predicate is the right one.** `getAccessToken() !== null` is character-for-character the negation of `attemptPing`'s own `!getAccessToken()` test, so the re-assert structurally cannot revive a keepalive that `attemptPing` tore down — which is the whole property being defended. Picking any other formulation (`getRefreshToken()`, `ensured.accessToken`) would have reintroduced the gap for some interleaving.

**Both sides of the guard stay covered.** #2421's own `re-arms the keepalive when a new command lifts a teardown mid-connect` (`api-socket.test.ts:1293`) passes unmodified — it holds valid tokens, so `getAccessToken()` is non-null and the re-assert still fires. Full file: 67/67 green locally.

### One thing I traced and am *not* flagging

The guard does admit a state that didn't exist before it: a logout landing during the refresh await, followed by the refresh resolving *with* a token, produces a created socket with `pingIntervalId === null`. I built the interleaving to check rather than reason about it, and the outcome is benign — the queued command is still sent on open (`sent: 1, queued: 0`), so nothing strands, and the missing keepalive is exactly the desired outcome for a session whose tokens are gone. Pre-merge behaviour reached the same end state ~10s later after one extra post-logout ping, so this is strictly no worse. Recording it because it's the natural objection to the guard, not as something to change.

### Test coverage

Three new tests, all in the existing `queued-but-unsent settlement on a terminal close` block, each reproducing a real interleaving rather than poking internals, each asserting the queue settles with `Connection lost. Please try again.` **and** that no socket is resurrected. The vacuity checks are the part that makes them worth having — I re-ran one of the three and it behaves as claimed. Frontend transport only, no battle-logic parity surface, so no mirrored backend test; agreed.

### What's good

- **Not taking the auto-merge.** A clean textual merge that passes CI is the easiest place in this whole change to lose the fix, and the only thing that caught it was a test written two commits earlier for a different reason. Flagging it as *"worth a look rather than a routine conflict resolution"* instead of burying it in the merge commit is the right instinct.
- **Reasoning about #2421's re-assert on its own terms** — correct for a teardown lifted by a new command, wrong for a logged-out one — rather than reverting or widening it. The two cases needed distinguishing, and `getAccessToken()` is the condition that distinguishes them.
- **`docs/frontend-sockets.md` carries both sides' bullets** plus the "not a blanket revival" clause, so the doc records the interaction between the two fixes and not just the union of them.
- Declining the issue's "check the in-flight map" with a stated argument, and `hasUnclosedSocket()` instead of a fourth copy of the `readyState` expression.

No follow-up issues filed — nothing here needs one. #2419's adjacent zombie-socket problem shipped separately in #2421, which is what this merge is reconciling with.

---

PR #2421, opened 2026-07-25
Cancel an in-flight openSocket on disconnect() (#2419)

### LINKED ISSUE #2419: disconnect() doesn't cancel an in-flight openSocket, so a teardown can be followed by a resurrected "zombie" socket

ginderjeremiah:
Closes #2419.

## Summary

`ApiSocket.disconnect()` stopped the keepalive and closed the live socket, but an `openSocket()` already suspended on its pre-emptive refresh await had no way to know it had been superseded — once the refresh resolved it constructed the WebSocket and wired handlers regardless. The result is exactly the outcome `disconnect()`'s own doc comment says it exists to prevent (a background reconnect fighting the session that just took over), and worse than before the teardown: `disconnect()` already stopped the keepalive and `openSocket` only arms it at its *start*, so the resurrected socket has **no owner** that would ever close it.

The issue's reproduction holds as written — a `SocketReplaced` takeover doesn't clear the tokens, so the reconnect's refresh succeeds and nothing else stops the connect.

## Changes

**`UI/src/lib/api/api-socket.ts`**

- A `teardownRequested` flag latched by `disconnect()`, checked in `openSocket`, settling the queued-but-unsent commands with `CONNECTION_LOST_ERROR` on either bail (mirroring every other terminal path — `disconnect()` with no socket to close has nothing that would otherwise settle them). Named for what it latches rather than for connection state: it's `false` through the entire genuinely-disconnected stretch after an ordinary close, and `true` while the torn-down socket may still be `CLOSING`.
- **Two check sites, not one.** The issue asks for a check "after the refresh await"; that alone is insufficient. `handleClose`'s auth-retry `refreshTokens().then(...)` calls `processCommandQueue()` **after** a teardown can have landed, which enters `openSocket` fresh — and `openSocket` arms the keepalive interval *before* its own await, so a post-await-only bail would leave an interval armed that nothing owns, re-entering `openSocket` every 10s forever. The top-of-`openSocket` check prevents the arming entirely. This is why the second new test asserts specifically on `pingIntervalId`.
- **The keepalive arming is re-asserted after the refresh await** via an idempotent `ensurePingInterval()`. The pre-await arming no longer survives the await in one interleave: a teardown lands (stopping that interval), then a new `sendSocketCommand` lifts the flag before the refresh resolves — and `ensureSocket` hands that command the *same* in-flight connect rather than starting a fresh one, so the resumed open would complete into a live socket with `pingIntervalId === null` (no half-open detection, and `handleClose`'s `socketOpened` branch holding the queue for a background reconnect that never runs). Found in review; the residual tail of this same bug.
- **The flag is cleared only by `sendSocketCommand`**, not also by `ensureSocket` as the issue suggests. `ensureSocket`'s other caller is `attemptPing` — the background reconnect the flag exists to suppress — so clearing it there would partly defeat the fix. An explicit command is the only legitimate "a new session is starting" signal, and it's sufficient: the keepalive can't reach `openSocket` after a teardown because `disconnect()` stopped the interval and only `openSocket` re-arms it.
- Nothing to close at the bail points: both checks sit before `new WebSocket`, so no socket has been constructed yet and the pre-existing one was already closed by `disconnect()`.

**`docs/frontend-sockets.md`** — the sticky-teardown contract is now its own bullet under _Socket connection lifecycle_ (the keepalive-ownership bullet it started in had grown to five sentences): closing the live socket isn't sufficient when the open path is async, only an explicit command lifts the mark, and the open re-asserts its keepalive after the refresh rather than assuming the one it armed survived.

## Test plan

- [x] New unit test `disconnect() during the open path refresh await leaves no resurrected socket behind` reproduces the issue's takeover window (command queued → connect suspended on the refresh → `disconnect()` → refresh resolves) and asserts no socket is created, the keepalive stays stopped, and the queued command settles with `Connection lost. Please try again.`
- [x] New unit test `does not connect when handleClose auth-retry refresh resolves after a disconnect()` covers the second (interval-arming) path via a real pre-open 1006 rejection.
- [x] New unit test `an explicit command after a disconnect() opens a new socket again` pins the flag-clearing, so the fix can't silently brick the singleton for the rest of the page's life.
- [x] New unit test `re-arms the keepalive when a new command lifts a teardown mid-connect` covers the interleave above.
- [x] Verified none of the four is vacuous: stashing only the source change fails the first two, and dropping just the post-await `ensurePingInterval()` call fails the fourth.
- [x] `npm run lint` — clean (eslint + `svelte-check`: 1222 files, 0 errors, 0 warnings).
- [x] `npm run test:unit:ci` — 312 files / 4013 tests passing, coverage thresholds met.
- [x] Backend untouched (frontend transport only) and no battle-logic parity surface, so no mirrored backend test is required.

## Notes / follow-ups

- **Touches the same file as open PR #2420** (#2418, `attemptPing`'s token-less branch). The two changes are in different methods and don't overlap semantically, but whichever merges second may need a trivial rebase.
- Filed #2422 for an adjacent gap noticed while reading the teardown: `disconnect()` doesn't reset `socketAuthRetries`, so a new same-tab session inherits the previous one's spent auth-retry budget. Deliberately left out of this PR — it changes the bounded-retry semantics established in #1992 and deserves its own review.


UI/src/lib/api/api-socket.ts
```diff
@@ -155,6 +168,14 @@ export class ApiSocket {
...
 		const hadRefreshToken = getRefreshToken() !== null;
 		const ensured = await ensureValidAccessToken();
+		// A disconnect() landed while the refresh was in flight (e.g. the SocketReplaced modal being
+		// acknowledged mid-reconnect). Connecting now would resurrect the very socket the teardown exists to
+		// prevent — tokens are still valid after a takeover, so nothing else stops it — and the keepalive
+		// that would normally close a stale socket has already been torn down, so nothing would own it.
+		if (this.disconnected) {
+			this.abortSupersededConnect();
+			return;
+		}
```
ginderjeremiah: **Should-fix** — this check closes the window when the flag is *still set* at resume time, but clearing it mid-await now produces a live socket with **no keepalive owner**. Concretely:

1. `sendSocketCommand(A)` → `openSocket` arms `pingIntervalId` (`:163-165`) and suspends on the refresh await.
2. `disconnect()` → `disconnected = true`, `stopPingInterval()` clears the interval. No socket exists yet, so nothing is closed.
3. `sendSocketCommand(B)` before the refresh resolves → `disconnected = false`, and `ensureSocket` hands back the *existing* `connecting` promise (`:140-142`) rather than starting a fresh open.
4. The refresh resolves, this check sees `false`, and the open completes into a real socket — **but `openSocket` only arms the interval at its start, and that arming was undone in step 2.**

The result is a connected session with no half-open detection and no background reconnect: a post-open drop lands in `handleClose`'s `socketOpened` branch (`:509-513`), which deliberately keeps the queue *for the keepalive to re-flush* — and there is no keepalive. It self-heals only on the next explicit command (`ensureSocket` sees `CLOSED` and re-arms), so a battling player recovers in seconds, but a parked one (idle on `/admin`, modal open) never reconnects.

It's narrow and not a regression — pre-PR the same interleave produced the full zombie this PR is fixing — but it's the same failure class, and it's cheap to close for good: the invariant "if I got here, I armed the interval" no longer survives the await, so re-assert it rather than assuming it. Extract the arming into an `ensurePingInterval()` and call it here too, after the bail check:

```ts
if (this.disconnected) {
	this.abortSupersededConnect();
	return;
}
// A disconnect() that landed and was then lifted by a new command stopped the interval this open armed
// before the await, so re-assert it rather than completing into a socket nothing owns.
this.ensurePingInterval();
```

Worth a test alongside the three new ones (disconnect mid-await → new command → socket opens **and** `pingIntervalId` is non-null), since that interleave is exactly what's untested today.

---

UI/src/lib/api/api-socket.ts
```diff
@@ -119,6 +119,12 @@ export class ApiSocket {
 	// Single-flight guard for the async open path: the in-flight open promise (null when not connecting),
 	// shared by concurrent ensureSocket callers so the pre-emptive-refresh await can't race two opens.
 	private connecting: Promise<void> | null = null;
+	// Set by disconnect() so an open that is already under way — suspended on the pre-emptive refresh await,
+	// or started afterwards by handleClose's auth-retry refresh resolving — can see that it has been
+	// superseded and bail, instead of handing back a socket the teardown it raced can no longer tear down.
+	// Cleared only by sendSocketCommand: an explicit command is the one legitimate way a new session starts,
+	// and the keepalive deliberately isn't (it exists to stop background reconnects after a teardown).
+	private disconnected = false;
```
ginderjeremiah: **Nit (naming footgun)** — `disconnected` reads like connection state, but it isn't one: it's `false` for the entire genuinely-disconnected stretch after an ordinary close, and `true` while the previous socket may still be `CLOSING`. A later reader reaching for "am I disconnected?" would find this field and be wrong in both directions. Something like `teardownRequested` or `connectSuppressed` says what it actually latches — that a teardown happened and background connects stay suppressed until an explicit command lifts it — and makes the `sendSocketCommand` clear point read as obviously correct instead of surprising.

Also, between this block, the two check sites, and `abortSupersededConnect`'s doc, "the keepalive is already stopped so nothing would own the socket" is now stated four times. The subtlety earns *some* prose, but one canonical statement here plus short pointers at the sites would carry the same information.

---

UI/src/lib/api/api-socket.ts
```diff
@@ -276,13 +300,23 @@ export class ApiSocket {
...
 			this.socket.close(NORMAL_CLOSURE);
 		}
 	}
 
+	/** Bails out of an open that a `disconnect()` superseded. The keepalive is already stopped and must stay
+	 *  that way, so nothing will reconnect to re-flush the queue — settle it like every other terminal path
+	 *  rather than leaving a caller awaiting a response that can never arrive. */
+	private abortSupersededConnect() {
+		this.settleQueuedRequests(CONNECTION_LOST_ERROR);
+	}
```
ginderjeremiah: **Nit** — the name promises abort semantics, but the body only settles the queue; the actual bail is the `return` at each call site, which the helper can't enforce. That's the kind of thing a later edit "simplifies" by calling this and forgetting the `return`. `settleSupersededQueue()` (or just inlining `this.settleQueuedRequests(CONNECTION_LOST_ERROR)` and keeping the doc comment on the field) describes what it does.

---

UI/src/tests/lib/api/api-socket.test.ts
```diff
@@ -1119,6 +1125,73 @@ describe('ApiSocket', () => {
 			expect(internals(apiSocket).pingIntervalId).toBeNull();
 			expect(ws.close).toHaveBeenCalledWith(1000);
 		});
+
+		it('disconnect() during the open path refresh await leaves no resurrected socket behind', async () => {
+			setTokens({ accessToken: 'a', refreshToken: 'r' });
+			// Hold the pre-emptive refresh open so the disconnect lands while openSocket is suspended on it —
+			// the session-takeover window: the modal is acknowledged mid-reconnect and the tokens stay valid.
+			let releaseRefresh!: (result: AccessTokenResult) => void;
```
ginderjeremiah: **Nit** — `let releaseRefresh!:` is a definite-assignment assertion, which `CLAUDE.md` asks us to avoid alongside the null-forgiving `!`. This same file already has the assertion-free idiom for the identical pattern at `api-socket.test.ts:183`:

```ts
let releaseRefresh: (result: AccessTokenResult) => void = () => {};
```

Same here and at `:1156`.

---

ginderjeremiah (commented):
## Review: cancel an in-flight `openSocket` on `disconnect()` (#2419)

**Verdict:** Approve — the diagnosis, the two-check placement, and the clear point are all right; one should-fix and three nits, none blocking. (Submitted as a comment because GitHub won't accept a gating verdict on my own PR.)

I traced every reachable entry into `openSocket` while the flag is set, and the fix holds where it matters. The two checks are also genuinely two, not belt-and-braces: the top-of-function one is the only thing standing between `handleClose`'s auth-retry `processCommandQueue()` (`:548-555`) and an interval armed at `:163-165` that nothing owns and nothing would stop — a post-await-only check would re-enter `openSocket` every 10s forever. Both new tests fail without their respective check, so neither is vacuous.

Two of the PR's departures from the issue's suggested fix are correct and worth recording. Clearing the flag in `sendSocketCommand` but **not** `ensureSocket` is right — `ensureSocket`'s other caller is `attemptPing`, the background reconnect the flag exists to suppress, so clearing there would partly undo the fix. And not closing anything at the bail points is right: both checks sit before `new WebSocket`, so nothing has been constructed. I also checked the sticky flag against both real `disconnect()` call sites (`engine.ts:397` on `SocketReplaced`, `CharacterSwitcher.svelte:76`) — both are terminal teardowns followed by navigation, and both paths reconnect through an explicit command, so nothing gets bricked.

### Should-fix

- **`api-socket.ts:171-178` — clearing the flag mid-await yields a socket with no keepalive.** `disconnect()` stops the interval that this open armed *before* its await; if a new `sendSocketCommand` lifts the flag while the refresh is still in flight, `ensureSocket` reuses the same `connecting` promise and the open completes into a live socket with `pingIntervalId === null` — no half-open detection, no background reconnect, and `handleClose`'s `socketOpened` branch holding the queue for a keepalive that isn't running. Narrow and self-healing on the next explicit command, and not a regression (pre-PR the same interleave produced the full zombie), but it's the residual tail of the very bug this PR fixes and closes with a three-line `ensurePingInterval()`. Details and a suggested diff inline.

### Nits

- `api-socket.ts:122-127` — `disconnected` names connection state but latches something else; `teardownRequested` / `connectSuppressed` would prevent a later reader misusing it. Plus the "nothing would own the socket" rationale is now stated in four places.
- `api-socket.ts:313-318` — `abortSupersededConnect()` doesn't abort anything; the `return` at each call site does. Rename or inline.
- `api-socket.test.ts:1133`, `:1156` — `let releaseRefresh!:` uses the definite-assignment `!` that `CLAUDE.md` asks us to avoid; the same file already has the assertion-free idiom at `:183`.

### Test coverage

Right shape and well placed — the three tests land in the existing `keepalive ping lifecycle` block next to `disconnect() stops the keepalive and closes the socket`, and they reproduce the sequences rather than poking internals (the one internals read, `pingIntervalId`, is the assertion that actually distinguishes the two checks). The second test earning its keep via a real pre-open 1006 rather than a hand-set flag is the right call, and the third pinning the clear point stops the fix from silently bricking the singleton. The stash-the-source vacuity check in the test plan is the right instinct.

One gap, and it's the should-fix's path: the mid-await *clear* — disconnect while suspended, then a new command before the refresh resolves — is untested, and it's the interleave where the current code still goes wrong. Frontend transport only, no battle-logic parity surface, so no mirrored backend test — agreed.

### What's good

- Catching that the issue's single "after the refresh await" check is insufficient, and pinning the reason (`pingIntervalId`) in a dedicated test rather than folding it into the first one.
- Settling the queue at both bails instead of leaving callers hanging — `disconnect()` with no socket to close has nothing else that would settle them, so this keeps the `frontend.md` terminal-settlement invariant true by construction rather than by luck.
- `frontend-sockets.md` extended in place so the keepalive-ownership bullet stays accurate, including which signal lifts the mark. It's grown into a five-sentence paragraph though — once the should-fix lands, the sticky-teardown contract probably reads better as its own bullet.
- Filing #2422 for the `socketAuthRetries` reset rather than folding it in; it changes the #1992 bounded-retry semantics and does deserve separate review.

---

ginderjeremiah:
All four addressed in `bf6521e`.

**Should-fix — confirmed and fixed.** I traced the interleave before changing anything and it holds exactly as described: `ensureSocket` returns the existing `connecting` promise (`:140-142`) rather than starting a fresh open, so the resumed open completes having had its own arming undone at step 2. The consequence you name is the sharper half — `handleClose`'s `socketOpened` branch deliberately keeps the queue *for the keepalive to re-flush*, so a parked session strands its queue behind a reconnect that never runs.

Taken as suggested, with the arming extracted into an idempotent `ensurePingInterval()` and re-asserted after the refresh await. That's better than re-arming conditionally on "was a teardown lifted?", since the invariant worth stating is simply "past this point, the interval is armed" — the assumption that broke was that the pre-await arming survives an await at all.

New test `re-arms the keepalive when a new command lifts a teardown mid-connect` pins it. Verified non-vacuous: dropping just the second `ensurePingInterval()` call fails it (63 passed / 1 failed).

**Naming (`disconnected`).** Agreed, and your "wrong in both directions" framing is the reason — false through the whole genuinely-disconnected stretch after an ordinary close, true while the torn-down socket may still be `CLOSING`. Renamed to `teardownRequested`. The four-way repetition of the ownership rationale is consolidated onto the field, with the check sites reduced to a pointer.

**`abortSupersededConnect`.** Right that the name overpromises — the bail is the caller's `return`, which the helper can't enforce. Renamed to `settleSupersededQueue` and moved next to `settleQueuedRequests`/`settleInFlightRequests` so the settling helpers sit together.

**Definite-assignment `!`.** Fixed at all three sites (the new test included) using the existing `:183` idiom.

**Docs.** Split the sticky-teardown contract into its own bullet as you suggested — the keepalive-ownership bullet had grown to five sentences, and the re-arm needed a clause anyway.

Verification: `npm run lint` clean (1222 files, 0 errors), `npm run test:unit:ci` 312 files / 4013 tests passing.

---

PR #2423, opened 2026-07-25
Fold the progression map suites' fixture builders into progression-test-utils (#2412)

### LINKED ISSUE #2412: Fold the progression map suites' duplicated path/tier builders into progression-test-utils

ginderjeremiah:
Closes #2412.

> **Updated after review** (0023cd7). One claim below was disproved by the review and is corrected: the `progression-helpers.test.ts` fold *dropped* coverage of the pre-load catalogue state rather than being strictly closer to the real `$stores`. Fixed inline; details in the [reply](https://github.com/ginderjeremiah/GameServer/pull/2423#discussion_r3650135331).

## Summary

`ProgressionMap.test.ts:19-50` and `progression-map.test.ts:6-37` each carried a **byte-for-byte identical** pair of `path`/`tier` builders — 44 duplicated lines in two files sitting next to each other. This folds them onto the shared `progression-test-utils` module #2411 introduced.

## How it addresses the issue

**Adapter over call-site migration.** The issue asked this be decided rather than assumed, so: adapter. Two positional wrappers land in `progression-test-utils.ts`:

```ts
export const mapPath = (id, over = {}) => path({ id, name: `Path ${id}`, activityKey: EActivityKey.Physical, ...over });
export const mapTier = (id, pathId, ordinal, over = {}) => tier({ id, name: `Tier ${id}`, pathId, pathOrdinal: ordinal, ...over });
```

Each suite imports them as `mapPath as path, mapTier as tier`, so the ~15 call sites (`tier(10, 0, 0)`) don't churn. On the "is two signatures for one type a smell?" question — **no, not here.** These two suites build *topology* (several tiers spread across several paths), so `pathId`/`ordinal` are the subject of nearly every assertion, not incidental overrides; the positional form makes that legible in a way `tier({ id: 10, pathId: 0, pathOrdinal: 0 })` does not. Divergent *defaults* would be the smell — the adapters carry only the shape, and the field set now has exactly one definition. It also matches the precedent already in the folder: `PathDetail.test.ts:22` wraps the shared `tier` in a local adapter for the same reason.

**Verified the defaults still bite.** Per #2405's process note, this is the hazard — a reconciled default can silently weaken a DOM-text assertion into passing. Mutating `Path ${id}` → `PathX ${id}` in the shared adapter fails `progression-map.test.ts:53` (`expect(cols[0].name).toBe('Path 0')`), confirming the generated name is still load-bearing and read back out of the rendered column header. Each file was converted and run individually before the next.

**Also folded in** (the issue's "worth a look while in there"): `progression-helpers.test.ts` now resolves the shared `stubStores` / `resetStores` / `staticData` instead of its own `vi.hoisted` `{} as any` stub and `delete`-key-loop reset, removing an `any` and its `eslint-disable`.

That swap initially cost real coverage, caught in review. The old `{}` stub left `staticData.skills` **`undefined`** at the top of every test, so the suite was implicitly exercising the `staticData.skills?.[…]` guard at `progression-helpers.ts:203` everywhere; the shared stub started it as `[]`, and nothing hit that branch any more. The `undefined` slot isn't a stub artifact — `static-data.svelte.ts` documents it as load-bearing (`loaded` and `reference-data.ts` detect "not yet loaded" by the slot being null), and the real getters expose `T[] | undefined`. So the shared stub was closer to the *loaded* state and further from the *pre-load* one, which is the state the production guard exists for.

Fixed by widening the shared stub's catalogue slots to `unknown[] | undefined` behind a `catalogueKeys` const tuple — which also lets `resetStores` reassign fresh empties without a `keyof` cast, so the removed `any` stays removed — plus a case covering the pre-load read. All nine slots widened rather than only `skills`: every slot is `undefined` pre-load, so the asymmetry would be arbitrary, and the other eight are only ever assigned in these suites, never read.

Its local `tier` builder is deliberately **left alone**: unlike the map suites' hand-rolled literal, it composes the *production* `newProficiency` factory, so it isn't duplication — it's coverage of the real factory's defaults, and replacing it would lose that.

## Test plan

Test-only change; no production code touched. The existing assertions are the regression guard and must keep passing unchanged; the one added test covers the pre-load branch this PR would otherwise have left uncovered.

- [x] `npx vitest run src/tests/routes/admin/workbench/progression/` — 10 files passing (run per-file during conversion, then together).
- [x] Mutation check on the dedup: `Path ${id}` → `PathX ${id}` in the shared adapter fails exactly 1 test, proving the generated-name default is still asserted rather than silently unused.
- [x] Mutation check on the restored coverage: `staticData.skills?.[…]` → `!` in `progression-helpers.ts` fails exactly the new test — and, with the fix reverted, fails **nothing**, which is what confirms the regression was real rather than theoretical.
- [x] `npm run test:unit:ci` — 312 files, 4017 tests passing.
- [x] `npm run lint` (eslint + `svelte-check`) — 1222 files, 0 errors, 0 warnings.

### DIFF
--- a/UI/src/tests/routes/admin/workbench/progression/progression-helpers.test.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-helpers.test.ts
@@ -1,11 +1,9 @@
 import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { EActivityKey, EAttribute, EModifierType, ESkillAcquisition } from '$lib/api';
 
-const { staticData } = vi.hoisted(() => ({
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any
-	staticData: {} as any
-}));
-vi.mock('$stores', () => ({ staticData }));
+// `vi.mock` is hoisted within this file, so the factory resolves the shared stub by dynamic import
+// rather than closing over the static one below.
+vi.mock('$stores', async () => (await import('./progression-test-utils')).stubStores());
 
 import {
 	activityKeyGroups,
@@ -30,12 +28,9 @@ import {
 	xpCostCurve
 } from '$routes/admin/workbench/progression/progression-helpers';
 import type { WorkbenchProficiency } from '$routes/admin/workbench/progression/types';
+import { resetStores, staticData } from './progression-test-utils';
 
-beforeEach(() => {
-	for (const key of Object.keys(staticData)) {
-		delete staticData[key];
-	}
-});
+beforeEach(resetStores);
 
 const tier = (overrides: Partial<WorkbenchProficiency> & Pick<WorkbenchProficiency, 'id'>): WorkbenchProficiency => ({
 	...newProficiency(overrides.id, overrides.pathId ?? 1, overrides.pathOrdinal ?? 0),
--- a/UI/src/tests/routes/admin/workbench/progression/progression-helpers.test.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-helpers.test.ts
@@ -221,6 +221,16 @@ describe('validation', () => {
 		expect(rewardSkillFlagWarnings(stillFlagged)).toHaveLength(0);
 		expect(proficiencyBlockingWarnings(stillFlagged)).toHaveLength(0);
 	});
+
+	it('tolerates the pre-load catalogue state, when the reward lookup runs before skills load', () => {
+		// `staticData.skills` is genuinely `undefined` until the loading screen populates it, which is
+		// what `rewardSkillFlagWarnings`' `?.` guards; warn on nothing rather than throwing.
+		staticData.skills = undefined;
+
+		const withReward = tier({ id: 1, maxLevel: 5, levelRewards: [{ level: 3, rewardSkillId: 1 }] });
+		expect(rewardSkillFlagWarnings(withReward)).toHaveLength(0);
+		expect(proficiencyBlockingWarnings(withReward)).toHaveLength(0);
+	});
 });
 
 describe('DTO builders', () => {
--- a/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
@@ -81,6 +81,23 @@ export const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProfici
 	...over
 });
 
+/* The map suites (`ProgressionMap.test.ts`, `progression-map.test.ts`) build topology — several tiers
+   across several paths — so identity is positional at their call sites (`mapTier(10, 0, 0)`), and their
+   assertions read the generated `Path {id}` name back out of the rendered column header. The two
+   adapters below keep the field set in one place while preserving both. */
+
+/** Positional path fixture for the map suites — `Path {id}`, keyed on physical damage. */
+export const mapPath = (id: number, over: Partial<WorkbenchPath> = {}): WorkbenchPath =>
+	path({ id, name: `Path ${id}`, activityKey: EActivityKey.Physical, ...over });
+
+/** Positional tier fixture for the map suites — `Tier {id}` at `ordinal` on `pathId`. */
+export const mapTier = (
+	id: number,
+	pathId: number,
+	ordinal: number,
+	over: Partial<WorkbenchProficiency> = {}
+): WorkbenchProficiency => tier({ id, name: `Tier ${id}`, pathId, pathOrdinal: ordinal, ...over });
+
 /**
  * Casts a partial fake to the store the components read. The suites drive these rather than the real
  * `ProgressionStore`, which only loads through a socket-backed `load()`.
--- a/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
@@ -12,17 +12,32 @@ import type { WorkbenchPath, WorkbenchProficiency } from '$routes/admin/workbenc
    import that did would re-enter the factory resolving it. Everything below `$lib/api` is `import type`
    and therefore erased, which is what keeps that true today. */
 
-/** The last-saved `staticData` catalogues the editor's reference lookups read. */
-export const staticData = {
-	enemies: [] as unknown[],
-	zones: [] as unknown[],
-	challenges: [] as unknown[],
-	items: [] as unknown[],
-	classes: [] as unknown[],
-	skillRecipes: [] as unknown[],
-	proficiencies: [] as unknown[],
-	skills: [] as unknown[],
-	paths: [] as unknown[]
+const catalogueKeys = [
+	'enemies',
+	'zones',
+	'challenges',
+	'items',
+	'classes',
+	'skillRecipes',
+	'proficiencies',
+	'skills',
+	'paths'
+] as const;
+
+/* The last-saved `staticData` catalogues the editor's reference lookups read. Each slot is
+   `unknown[] | undefined` because the real store's is: the backing `$state` is genuinely `undefined`
+   until the loading screen populates it, and that state is load-bearing (`static-data.svelte.ts`), so
+   production readers guard with `?.`. A suite covering a pre-load read assigns `undefined` directly. */
+export const staticData: Record<(typeof catalogueKeys)[number], unknown[] | undefined> = {
+	enemies: [],
+	zones: [],
+	challenges: [],
+	items: [],
+	classes: [],
+	skillRecipes: [],
+	proficiencies: [],
+	skills: [],
+	paths: []
 };
 
 export const dangerModal = vi.fn();
@@ -40,8 +55,8 @@ export const stubStores = () => ({ staticData, dangerModal });
 /** Restores the stub to its empty, uncalled baseline. Call from `beforeEach`. */
 export const resetStores = () => {
 	dangerModal.mockReset();
-	for (const catalogue of Object.values(staticData)) {
-		catalogue.length = 0;
+	for (const key of catalogueKeys) {
+		staticData[key] = [];
 	}
 };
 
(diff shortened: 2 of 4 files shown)


UI/src/tests/routes/admin/workbench/progression/progression-helpers.test.ts
```diff
@@ -30,12 +28,9 @@ import {
...
 } from '$routes/admin/workbench/progression/progression-helpers';
 import type { WorkbenchProficiency } from '$routes/admin/workbench/progression/types';
+import { resetStores, staticData } from './progression-test-utils';
 
-beforeEach(() => {
-	for (const key of Object.keys(staticData)) {
-		delete staticData[key];
-	}
-});
+beforeEach(resetStores);
```
ginderjeremiah, line 33: **Should-fix** — this swap silently drops the suite's coverage of the pre-load catalogue state.

Under the old `{} as any` stub plus its `delete`-key reset, `staticData.skills` was `undefined` at the top of every test, so every call into `proficiencyWarnings` / `proficiencyBlockingWarnings` exercised the `staticData.skills?.[…]` guard at `progression-helpers.ts:203`. With the shared stub, `skills` starts as `[]` and can never be `undefined`, so that branch is now hit by nothing in the suite.

That matters because `undefined` isn't a stub artifact — `static-data.svelte.ts:33-37` documents it as load-bearing (`loaded` and `reference-data.ts` detect "not yet loaded" by the slot being null, and `reset()` at `:148` puts it back), and the getters honestly expose `ISkill[] | undefined`. So the PR description's "strictly closer stand-in for the real `$stores`" holds for the *loaded* state only; for the pre-load state the shared stub is strictly further away, and it's the state the production guard exists for. This is the same "a reconciled default silently weakens what was being exercised" hazard #2405/#2412 flagged, just landing on a defensive branch rather than an assertion.

Suggested fix, keeping the fold: widen the stub's catalogues to `unknown[] | undefined` and have `resetStores` reassign fresh empties (rather than `length = 0`), then add one case here:

```ts
it('tolerates the pre-load catalogue state', () => {
	staticData.skills = undefined;
	const withReward = tier({ id: 1, maxLevel: 5, levelRewards: [{ level: 3, rewardSkillId: 1 }] });
	expect(rewardSkillFlagWarnings(withReward)).toHaveLength(0);
	expect(proficiencyBlockingWarnings(withReward)).toHaveLength(0);
});
```

Widening the type is what keeps that assignment cast-free — otherwise covering it needs an `as` escape, which would give back the `any` this change removed.

---

UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
```diff
@@ -81,6 +81,23 @@ export const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProfici
...
+export const mapPath = (id: number, over: Partial<WorkbenchPath> = {}): WorkbenchPath =>
+	path({ id, name: `Path ${id}`, activityKey: EActivityKey.Physical, ...over });
+
+/** Positional tier fixture for the map suites — `Tier {id}` at `ordinal` on `pathId`. */
+export const mapTier = (
+	id: number,
+	pathId: number,
+	ordinal: number,
+	over: Partial<WorkbenchProficiency> = {}
+): WorkbenchProficiency => tier({ id, name: `Tier ${id}`, pathId, pathOrdinal: ordinal, ...over });
```
ginderjeremiah, line 112: **Nit** — the rationale is stated three times across 16 lines: the block comment explains positional call sites and the generated `Path {id}` name, then each doc comment repeats "positional … for the map suites" and the name pattern. The *why* here is genuinely worth recording (it's the decision #2412 asked for), but one of the two layers can go — either drop the block and keep the per-export docs, or keep the block and shorten the docs to `/** See above. */`-level. `CLAUDE.md` asks for 1-2 line comments unless the extra context is critical, and the third restatement isn't.

Non-blocking either way.

---

ginderjeremiah (commented):
## Review: folding the progression map suites' duplicated `path`/`tier` builders into `progression-test-utils` (#2412)

**Verdict:** Approve with one should-fix — the dedup itself is right and correctly executed; the incidental `progression-helpers.test.ts` fold quietly loses a branch the old stub was covering. Posted as a `COMMENT` because GitHub won't take a gating verdict from the PR author.

### Should-fix

- `progression-helpers.test.ts:33` — moving off `{} as any` + the `delete`-loop reset means `staticData.skills` is now `[]` instead of `undefined` at the start of every test, so nothing in the suite exercises the `staticData.skills?.[…]` guard at `progression-helpers.ts:203` any more. `static-data.svelte.ts:33-37` documents the `undefined` slot as load-bearing and real, so that's a genuine edge case going from implicitly-covered-everywhere to uncovered. Fix inline; it wants a stub type widened to `unknown[] | undefined` plus one test, and it belongs in this PR rather than a follow-up since this PR is what drops the coverage.

### Nits

- `progression-test-utils.ts:84-99` — the adapter rationale is written out three times (block comment + two doc comments). Trim one layer.

### Test coverage

Test-only change with no production code touched, so the existing assertions are the regression guard and they must keep passing unchanged — that's the right standard here, and the fixture reconciliation is faithful: `mapPath`/`mapTier` override exactly `id`/`name`/`activityKey` and `id`/`name`/`pathId`/`pathOrdinal` over the shared bases, and every other field in the deleted literals already matched the shared builder byte-for-byte. Nothing in either map suite reads `activityKey`, so preserving `Physical` is harmless-but-correct. The mutation check on `` `Path ${id}` `` is the right way to prove the generated name is still load-bearing rather than decorative, and per-file conversion followed #2405's process note.

Confirmed the duplication is fully gone — no other suite still hand-rolls these builders. The one coverage regression is the should-fix above.

### Follow-up (out of scope for this PR)

`progression-store.test.ts:14` and `Progression.test.ts:22` still carry `staticDataMock: {} as any` with an `eslint-disable`, in this same folder. They're not what #2412 named and they stub a wider `$stores` surface (`toastError`), so leaving them is right — but they're the same `any` class and worth folding onto `stubStores()` later. That follow-up is naturally *downstream* of the should-fix: it only makes sense once the shared stub can express the pre-load `undefined`, otherwise it would spread the same coverage loss to two more suites. Happy to file it once this lands.

### What's good

- Choosing the adapter over migrating ~15 call sites is the right call, and the description actually answers the "is two signatures for one type a smell?" question the issue asked to be decided rather than assuming it — the topology argument (`pathId`/`ordinal` are the subject of the assertions, not incidental overrides) is convincing, and the `PathDetail.test.ts:22` precedent backs it.
- Leaving `progression-helpers.test.ts`'s local `tier` alone because it composes the production `newProficiency` is exactly right — that one isn't duplication, it's coverage of the real factory's defaults.
- Net −57 lines with the field set now defined once.

---

UI/src/tests/routes/admin/workbench/progression/progression-helpers.test.ts
ginderjeremiah: Confirmed and fixed in 0023cd7 — you're right, and my "strictly closer stand-in" line in the description was wrong as written. It holds for the loaded state and is backwards for the pre-load one, which is the state the guard exists for.

Verified both halves rather than taking the report at face value:

- `progression-helpers.ts:203` does read `staticData.skills?.[reward.rewardSkillId]`, and `static-data.svelte.ts` documents the `undefined` slot as load-bearing (`loaded` / `reference-data.ts` detect not-yet-loaded by the slot being null), so the getters honestly expose `T[] | undefined`.
- The coverage really was lost, not just theoretically: with the fix reverted to `[]`, mutating that `?.` to `!` fails **nothing**. With the new case, it fails exactly one test.

Took your suggested shape. Catalogue slots widened to `unknown[] | undefined` behind a `catalogueKeys` const tuple, which also lets `resetStores` reassign fresh empties without a `keyof` cast — so the `any` this PR removed stays removed. All nine slots widened rather than just `skills`: the asymmetry would be arbitrary (every slot is `undefined` pre-load) and I checked the other eight are only ever *assigned* in these suites, never read, so nothing needed a new guard.

The added case is worded around the real scenario (reward lookup running before skills load) rather than the stub mechanics. Suite is 743 → 4017 green across the full run, lint and `svelte-check` clean.

On the downstream follow-up you flagged — agreed it's now unblocked, since the shared stub can express the pre-load `undefined`. Please do file it; happy to pick it up.

---

UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
ginderjeremiah: Fair — dropped the two per-export doc comments and kept the block, in 0023cd7. The block is the layer that carries the *decision* (#2412 asked for it), and the signatures are self-describing once you've read it, so the per-export line was pure restatement.

---

PR #2424, opened 2026-07-25
Consolidate the duplicated IProficiency test fixture builders (#2414)

### LINKED ISSUE #2414: Consolidate the duplicated IProficiency test fixture builders
Found while reviewing #2413 (which consolidated the Lexicon *view-model* fixtures under `lexicon-test-utils.ts`). A second, distinct duplication class is left behind in the same area: the raw `IProficiency` contract builder.

## Problem

Five suites hand-roll the same `IProficiency` fixture, across four folders:

| File | Local name | Signature |
|---|---|---|
| `UI/src/tests/routes/game/screens/proficiencies/Proficiencies.test.ts:30` | `prof` | `(o: Partial<IProficiency> & { id; pathId; pathOrdinal })` |
| `UI/src/tests/routes/game/screens/proficiencies/proficiencies-view.test.ts:41` | `prof` | same |
| `UI/src/tests/routes/game/screens/synthesis/synthesis-reveal.test.ts:41` | `prof` | `(id, name)` |
| `UI/src/tests/stores/proficiencies.svelte.test.ts:29` | `proficiency` | `(id, levelModifiers)` |
| `UI/src/tests/routes/admin/workbench/references.test.ts:126` | `proficiency` | `(id, name, opts)` |

The two in `screens/proficiencies/` are near-identical — they diverge only in `iconPath` (`''` vs `` `icon-${id}` ``) and `pronunciation`/`translation` (`''` vs `` `p${id}` ``/`` `t${id}` ``), and `proficiencies-view.test.ts` asserts on those generated values, so the reconciliation is real work rather than a copy-paste. `proficiencies-view.test.ts:58` carries a matching `IPath` builder too.

Same cost as #2405 / #2407: every new `IProficiency` contract field taxes five files to keep `svelte-check` green.

## Suggested direction

`UI/src/tests/fixtures/` is already the home for cross-folder contract builders (`attributes.ts` → `makeAttribute`), so a `proficiencies.ts` there exporting `makeProficiency(id, overrides)` (and `makePath`) fits the existing convention better than a folder-local module — three of the five call sites live outside `screens/proficiencies/`.

Constraints worth carrying over from #2407:

- Reconcile the `iconPath` / `pronunciation` / `translation` defaults explicitly; generated-from-id defaults are the ones with live assertions, so prefer those and let the suites that pass `''` keep doing so per call.
- Convert one file at a time, running that suite in between, so a changed default fails loudly instead of quietly weakening an assertion.
- Keep each suite's local call signature where it's terse (thin one-line adapters over the shared builder), so the ~40 call sites don't churn.

## Scope

Small–medium — test-only, no production code, no behavior change.

ginderjeremiah:
Closes #2414.

## Summary

Six suites across five folders hand-rolled the same `IProficiency` contract fixture (and two of them an `IPath` one), so every new field on either contract taxed all six files to keep `svelte-check` green. This adds `UI/src/tests/fixtures/proficiencies.ts` exporting `makeProficiency` / `makePath` — the same home and shape as the existing `fixtures/attributes.ts` → `makeAttribute` — and points each suite at it.

Net: **~125 lines deleted, ~25 added.**

Two related modules stay separate, and the new file's header names both so the next person to add a contract field doesn't stop at the first one they find:

- `lexicon-test-utils.ts` builds the derived `PathView`/`TierView` (#2413) — a suite that needs both wants this module for the inputs and that one for the outputs.
- `progression-test-utils.ts` (#2412) builds `WorkbenchPath`/`WorkbenchProficiency`, which `progression/types.ts` declares as **plain aliases** of these same two contracts. Its blank conlang defaults and `xpGrowth: 1.4` are load-bearing for the workbench detail suites, so converging the two is its own change — #2426. This PR therefore narrows the edit surface for a new contract field from six places to two, not to one.

## How it addresses the issue

Per the issue's constraints:

- **Local call signatures preserved.** Each suite keeps its own `prof` / `proficiency` / `path` name as a thin one-line adapter over the shared builder, so the ~40 call sites don't churn. The two screens suites keep `pathId`/`pathOrdinal` **required** in their local signature — a spine tier that doesn't state where it sits is a meaningless fixture, and that guard is worth keeping even though the shared builder defaults them.
- **Defaults reconciled explicitly, toward the generated variants.** `iconPath`/`word`/`pronunciation`/`translation` are generated from the id (`icon-1`, `w1`, `p1`, `t1`). Those are the ones with a live assertion — `proficiencies-view.test.ts:219` asserts `iconPath === 'icon-0'` — whereas the suites that defaulted them to `''` assert nothing about them, so they take the generated values rather than being forced to pass `''` per call. I checked each of the four fields' assertions individually rather than assuming.
- **One file at a time, suite run in between**, so a changed default would have failed loudly instead of quietly weakening an assertion.

Two small deliberate calls worth flagging:

- `pathOrdinal` defaults to `0`, not to the id (which is what `lexicon-test-utils.ts`'s `tierView` does). Every suite that actually builds a spine passes it explicitly, and defaulting it to the id would silently reposition the fixtures in the suites that don't care about ordering.
- `references.test.ts`'s local `proficiency` used `xpGrowth: 1.4` where everything else used `1`; nothing in that suite reads it, so it now takes the shared `1`. Its `path` likewise moves from `EActivityKey.Fire` to the neutral `Physical` default — again unread there.

Two suites (`Proficiencies.test.ts`'s `PATHS`, `Synthesis.test.ts`'s `PROFICIENCIES`) had bare object literals rather than a builder; both now go through the shared fixtures. `Synthesis.test.ts` was a sixth instance the issue's survey missed — caught in review, and picked up here since it's in a folder this change already touched.

## Test plan

- [x] `npm run lint` (eslint + `svelte-check`) — clean, 1223 files, 0 errors / 0 warnings. The now-unused `EActivityKey` import in `references.test.ts` was removed with the builder that used it.
- [x] `npx prettier --check` on every touched file — clean.
- [x] Each converted suite run individually right after its conversion: `proficiencies-view` (42), `Proficiencies` (5), `synthesis-reveal` (11), `proficiencies.svelte` (19), `references` (45), `Synthesis` (12).
- [x] Full unit suite, re-run after the review fixes: **312 files / 4017 tests passed**.
- [x] No backend or production frontend code touched — test-only, no behavior change, no battle-parity surface.

## Follow-ups filed

- **#2425** — the same duplication class at much larger scale on other contracts: `ISkill` hand-rolled in ~30 suites, `IItem` in ~20, `IZone` in ~16. `references.test.ts` alone still carries seven more local builders (`enemy`, `zone`, `challenge`, `item`, `itemMod`, `wclass`, `recipe`, `skill`) that this change deliberately leaves in place.
- **#2426** — point `progression-test-utils.ts`'s `tier`/`path` at these shared builders, keeping its divergent defaults as stated overrides.

### DIFF
--- a/UI/src/tests/fixtures/proficiencies.ts
+++ b/UI/src/tests/fixtures/proficiencies.ts
@@ -1,0 +1,40 @@
+import { EActivityKey, type IPath, type IProficiency } from '$lib/api';
+
+/* Shared `IPath` / `IProficiency` contract builders (#2414). These are the raw reference-data records,
+   distinct from `lexicon-test-utils.ts`'s `PathView` / `TierView` — those are the derived view models the
+   Lexicon renders. A suite that needs both wants this module for the inputs and that one for the outputs. */
+
+/**
+ * Builds an {@link IProficiency} reference-data entry for tests. The conlang and icon fields are generated
+ * from the id so a suite can assert on `icon-1`/`w1`/`p1`/`t1` without restating them per call; everything
+ * else is a neutral placeholder. `pathId`/`pathOrdinal` default to 0 — a suite building a spine is expected
+ * to place each tier explicitly rather than inherit a position.
+ */
+export const makeProficiency = (o: Partial<IProficiency> & { id: number }): IProficiency => ({
+	name: `Prof ${o.id}`,
+	description: '',
+	designerNotes: '',
+	iconPath: `icon-${o.id}`,
+	word: `w${o.id}`,
+	pronunciation: `p${o.id}`,
+	translation: `t${o.id}`,
+	pathId: 0,
+	pathOrdinal: 0,
+	maxLevel: 10,
+	baseXp: 100,
+	xpGrowth: 1,
+	levelModifiers: [],
+	levelRewards: [],
+	prerequisiteIds: [],
+	...o
+});
+
+/** Builds an {@link IPath} reference-data entry for tests. A path carries no word of its own — the rail and
+ *  spine header reuse the root tier's — so there is nothing conlang-shaped to generate here. */
+export const makePath = (o: Partial<IPath> & { id: number }): IPath => ({
+	name: `Path ${o.id}`,
+	description: '',
+	designerNotes: '',
+	activityKey: EActivityKey.Physical,
+	...o
+});
--- a/UI/src/tests/fixtures/proficiencies.ts
+++ b/UI/src/tests/fixtures/proficiencies.ts
@@ -1,8 +1,13 @@
 import { EActivityKey, type IPath, type IProficiency } from '$lib/api';
 
-/* Shared `IPath` / `IProficiency` contract builders (#2414). These are the raw reference-data records,
-   distinct from `lexicon-test-utils.ts`'s `PathView` / `TierView` — those are the derived view models the
-   Lexicon renders. A suite that needs both wants this module for the inputs and that one for the outputs. */
+/* Shared `IPath` / `IProficiency` contract builders (#2414). Two other modules build adjacent shapes:
+
+   · `lexicon-test-utils.ts` builds the derived `PathView` / `TierView` the Lexicon renders — a suite that
+     needs both wants this module for the inputs and that one for the outputs.
+   · `progression/progression-test-utils.ts` (`path`/`tier`) builds `WorkbenchPath`/`WorkbenchProficiency`,
+     which are *plain aliases* of these same two contracts — so a new contract field taxes that module too.
+     Its blank conlang defaults and `xpGrowth: 1.4` are load-bearing for the workbench detail suites and
+     deliberately diverge from the defaults here; converging the two is #2426. */
 
 /**
  * Builds an {@link IProficiency} reference-data entry for tests. The conlang and icon fields are generated
@@ -10,14 +15,14 @@ import { EActivityKey, type IPath, type IProficiency } from '$lib/api';
  * else is a neutral placeholder. `pathId`/`pathOrdinal` default to 0 — a suite building a spine is expected
  * to place each tier explicitly rather than inherit a position.
  */
-export const makeProficiency = (o: Partial<IProficiency> & { id: number }): IProficiency => ({
-	name: `Prof ${o.id}`,
+export const makeProficiency = (overrides: Partial<IProficiency> & { id: number }): IProficiency => ({
+	name: `Prof ${overrides.id}`,
 	description: '',
 	designerNotes: '',
-	iconPath: `icon-${o.id}`,
-	word: `w${o.id}`,
-	pronunciation: `p${o.id}`,
-	translation: `t${o.id}`,
+	iconPath: `icon-${overrides.id}`,
+	word: `w${overrides.id}`,
+	pronunciation: `p${overrides.id}`,
+	translation: `t${overrides.id}`,
 	pathId: 0,
 	pathOrdinal: 0,
 	maxLevel: 10,
@@ -26,15 +31,15 @@ export const makeProficiency = (o: Partial<IProficiency> & { id: number }): IPro
 	levelModifiers: [],
 	levelRewards: [],
 	prerequisiteIds: [],
-	...o
+	...overrides
 });
 
 /** Builds an {@link IPath} reference-data entry for tests. A path carries no word of its own — the rail and
  *  spine header reuse the root tier's — so there is nothing conlang-shaped to generate here. */
-export const makePath = (o: Partial<IPath> & { id: number }): IPath => ({
-	name: `Path ${o.id}`,
+export const makePath = (overrides: Partial<IPath> & { id: number }): IPath => ({
+	name: `Path ${overrides.id}`,
 	description: '',
 	designerNotes: '',
 	activityKey: EActivityKey.Physical,
-	...o
+	...overrides
 });
--- a/UI/src/tests/routes/game/screens/synthesis/synthesis-reveal.test.ts
+++ b/UI/src/tests/routes/game/screens/synthesis/synthesis-reveal.test.ts
@@ -6,6 +6,7 @@ import {
 	representativeRecipe,
 	type RecipeState
 } from '$routes/game/screens/synthesis/synthesis';
+import { makeProficiency } from '../../../../fixtures/proficiencies';
 
 /* The derivation takes explicit args and needs no mocks. The reference catalogues are zero-based-id
    arrays (resolved by index), so the fixtures keep ids contiguous with their array position. */
@@ -38,24 +39,7 @@ const recipe = (
 	over: Partial<ISkillRecipe> = {}
 ): ISkillRecipe => ({ id, resultSkillId, inputSkillIds, conditions, designerNotes: '', ...over });
 
-const prof = (id: number, name: string): IProficiency => ({
-	id,
-	name,
-	description: '',
-	designerNotes: '',
-	iconPath: '',
-	word: '',
-	pronunciation: '',
-	translation: '',
-	pathId: 0,
-	pathOrdinal: 0,
-	maxLevel: 10,
-	baseXp: 100,
-	xpGrowth: 1,
-	levelModifiers: [],
-	levelRewards: [],
-	prerequisiteIds: []
-});
+const prof = (id: number, name: string): IProficiency => makeProficiency({ id, name });
 
 /* A small fixed world: skills 0..3 are inputs, skills 4..6 are synthesis results. */
 const skills: ISkill[] = [
--- a/UI/src/tests/routes/admin/workbench/references.test.ts
+++ b/UI/src/tests/routes/admin/workbench/references.test.ts
@@ -1,6 +1,5 @@
 import { describe, it, expect } from 'vitest';
 import {
-	EActivityKey,
 	EAttribute,
 	EChallengeType,
 	EEntityType,
@@ -26,6 +25,7 @@ import {
 	type ReferenceGroup,
 	type ReferenceSources
 } from '$routes/admin/workbench/references';
+import { makePath, makeProficiency } from '../../../fixtures/proficiencies';
 
 const enemy = (id: number, name: string, opts: Partial<IEnemy> = {}): IEnemy => ({
 	id,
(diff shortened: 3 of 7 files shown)


UI/src/tests/fixtures/proficiencies.ts
```diff
@@ -0,0 +1,40 @@
+import { EActivityKey, type IPath, type IProficiency } from '$lib/api';
+
+/* Shared `IPath` / `IProficiency` contract builders (#2414). These are the raw reference-data records,
+   distinct from `lexicon-test-utils.ts`'s `PathView` / `TierView` — those are the derived view models the
+   Lexicon renders. A suite that needs both wants this module for the inputs and that one for the outputs. */
```
ginderjeremiah: **Should-fix** — this header names `lexicon-test-utils.ts` as the module to keep separate from, which is right, but it omits the module a reader is far more likely to collide with: `tests/routes/admin/workbench/progression/progression-test-utils.ts:79`, whose `tier` builder constructs `WorkbenchProficiency` — and `routes/admin/workbench/progression/types.ts:15` declares `export type WorkbenchProficiency = IProficiency`, a plain alias. Same for `WorkbenchPath = IPath` and that module's `path` at :64.

So `IProficiency` has two shared fixture homes after this PR, not one, and the distinction between them isn't the raw-contract-vs-view-model line this comment draws — it's the same contract twice. Someone landing here to add a field will read this comment, conclude they've found the one place, and miss the other.

Worth stating plainly in the header, e.g. "the workbench progression suites keep their own `tier`/`path` in `progression-test-utils.ts` (see #2426)". That also makes the divergent defaults there (`xpGrowth: 1.4`, blank conlang) discoverable rather than a surprise on the next reconciliation.

---

UI/src/tests/routes/game/screens/synthesis/synthesis-reveal.test.ts
```diff
@@ -38,24 +39,7 @@ const recipe = (
...
-	pathId: 0,
-	pathOrdinal: 0,
-	maxLevel: 10,
-	baseXp: 100,
-	xpGrowth: 1,
-	levelModifiers: [],
-	levelRewards: [],
-	prerequisiteIds: []
-});
+const prof = (id: number, name: string): IProficiency => makeProficiency({ id, name });
```
ginderjeremiah, line 42: **Should-fix** — the sibling suite in this same folder, `Synthesis.test.ts:65-83`, declares its `PROFICIENCIES` as a full bare `IProficiency` object literal (all 16 fields). It's a sixth instance of the duplication this PR exists to remove, the issue's survey just missed it, and it's the exact case the PR already handled elsewhere — `Proficiencies.test.ts`'s `PATHS` were bare literals too and were converted rather than left alone.

The whole change is one line:

```ts
const PROFICIENCIES: IProficiency[] = [makeProficiency({ id: 0, name: 'Geomancy' })];
```

Nothing in that suite reads the fields whose defaults differ (`iconPath`/`word`/`pronunciation`/`translation` are `''` there and generated in the shared builder) — its assertions run off recipe gating and the skill-level `word`/`pronunciation`/`translation` at `synthesis-reveal.test.ts:50`, not the proficiency's.

Per CLAUDE.md, an issue's own survey isn't the scope ceiling — worth catching this here rather than leaving one hand-rolled copy behind in a folder the PR already touched.

---

UI/src/tests/fixtures/proficiencies.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Builds an {@link IProficiency} reference-data entry for tests. The conlang and icon fields are generated
+ * from the id so a suite can assert on `icon-1`/`w1`/`p1`/`t1` without restating them per call; everything
+ * else is a neutral placeholder. `pathId`/`pathOrdinal` default to 0 — a suite building a spine is expected
+ * to place each tier explicitly rather than inherit a position.
+ */
+export const makeProficiency = (o: Partial<IProficiency> & { id: number }): IProficiency => ({
```
ginderjeremiah: **Nit** — the parameter is `o` here and in `makePath`, where the sibling in this folder (`fixtures/attributes.ts:8`) names it `overrides`. The suite-local adapters can keep `o` (they're one-liners), but the shared exported builders are the ones read by people who didn't write them; `overrides` matches the neighbour and reads better against the docstring, which already uses that word.

---

ginderjeremiah (commented):
## Review: consolidate the five hand-rolled `IProficiency` test fixture builders into `tests/fixtures/proficiencies.ts` (#2414)

**Verdict:** Approve (posted as `COMMENT` — GitHub won't take a gating verdict from the PR author) — no blocking findings. Two should-fixes and one nit inline; both should-fixes are about the change being *incomplete*, not wrong. Not merging on this pass because of them.

Test-only, no production code, CI green across all eleven checks (backend jobs correctly skipped by the changed-path gate).

### Verification

I checked each reconciled default against the suites' actual assertions rather than trusting the description's checklist, since a silently-weakened assertion is the one real hazard in a change like this.

**The four generated fields are safe everywhere they changed.** `iconPath`/`word`/`pronunciation`/`translation` move from `''` to generated in three suites. The only live assertion on any of them is `proficiencies-view.test.ts:219` (`icon-0`), which was already on the generated variant and is unaffected. `Proficiencies.test.ts` asserts entirely on `data-testid` and `aria-current`; `synthesis-reveal.test.ts:89`'s `word` assertion reads the **skill**'s `word` (`sijren`, set at :50), not a proficiency's; `proficiencies.svelte.test.ts` reads only `levelModifiers`. Choosing the generated variants over forcing `''` per call was the right call.

**The two flagged `references.test.ts` changes are genuinely unread.** I traced `references.ts` rather than taking the description's word: `proficiencyReferences` and `pathReferences` read `name`, `pathId`, `prerequisiteIds` and `levelRewards` only (`references.ts:136-166`). `xpGrowth` (1.4 → 1) and `activityKey` (`Fire` → `Physical`) are dead in that suite, and nothing else in it touches them. Neither was load-bearing.

**No unused imports left behind.** `EActivityKey` survives in `proficiencies-view.test.ts` (:64, :74, :252-263) and `Proficiencies.test.ts` (:41-42) with the builders gone; `IPath` survives as the `mockStaticData.paths` type at `proficiencies-view.test.ts:14`. Only `references.test.ts` genuinely lost its `EActivityKey` use, and it's removed. This is the failure mode I expected to find and didn't.

**`bySlot` still holds.** `references.test.ts`'s zero-based-id/index invariant is unaffected — the builders still take `id` positionally and the shared one requires it.

### Should-fix

- **`Synthesis.test.ts:65-83` is a sixth hand-rolled copy, in a folder this PR already converted** — inline on `synthesis-reveal.test.ts:42`. Full 16-field bare literal, one line to convert, no default conflicts. The issue's survey listed five suites and missed it; per CLAUDE.md that survey isn't the scope ceiling, and this PR already converted `Proficiencies.test.ts`'s bare `PATHS` literals on exactly that reasoning.
- **`IProficiency` has two shared fixture homes after this PR, and the new file's header doesn't say so** — inline on `fixtures/proficiencies.ts:3-5`. `WorkbenchProficiency`/`WorkbenchPath` are plain aliases of `IProficiency`/`IPath` (`progression/types.ts:8,15`), so `progression-test-utils.ts:64,79` builds the same two contracts. Convergence itself is a follow-up — those blank conlang defaults and `xpGrowth: 1.4` are load-bearing for the workbench detail suites and #2412 only just landed that module — so I've filed **#2426**. What belongs in *this* PR is the header comment naming it, so the next person to add a contract field doesn't stop at one of the two.

  Worth adjusting the PR description too: "one place to edit when the contract grows" is accurate for the five converted suites but not for the contract.

### Nits

- Parameter named `o` rather than `overrides` on both exported builders — inline on `fixtures/proficiencies.ts:13`.

### Test coverage

No new coverage is owed. This is a pure test-infrastructure refactor with no production change, no behaviour change, and no battle-parity surface, so the mirrored-scenario rule doesn't apply. The 312-file / 4017-test full run plus the per-suite runs during conversion are the right evidence; converting one file at a time with a run in between is exactly the guard #2414 asked for, and it's what makes the default reconciliation trustworthy rather than hopeful.

### What's good

- **The default reconciliation was done per-field against live assertions, not by picking a winner.** Landing on the generated variants *because* `proficiencies-view.test.ts:219` asserts on `icon-0` — rather than defaulting to `''` and making three suites opt in — is the version of this change that leaves assertions stronger instead of quietly looser.
- **Keeping `pathId`/`pathOrdinal` required in the two screens-suite adapters, against the shared builder's defaults.** A spine tier that doesn't state where it sits is a meaningless fixture, and a compile-time guard on future call sites is worth the two-line adapter. The one-line comment explaining why it's not just ceremony is the right length.
- **Not defaulting `pathOrdinal` to the id**, unlike `lexicon-test-utils.ts`'s `tierView`. That would have silently repositioned fixtures in the three suites that don't care about ordering — a real trap, avoided deliberately rather than by luck.
- **Volunteering the two unverifiable-by-inspection changes** (`xpGrowth: 1.4`, `EActivityKey.Fire`) in the description instead of letting a reviewer find them. Both check out; flagging them anyway is what made them cheap to confirm.
- **Splitting #2425 rather than widening.** `ISkill` across ~30 suites is a different-sized problem and folding it in would have made this diff unreviewable. Leaving `references.test.ts`'s seven other local builders in place was the disciplined call.

---

UI/src/tests/routes/game/screens/synthesis/synthesis-reveal.test.ts
ginderjeremiah: Done in dd8d766 — this and both other findings.

Verified before converting rather than taking the line on faith, and the reasoning lands in the same place by a slightly different route than described: in `Synthesis.test.ts` the only proficiency-derived assertion is the gate label reading `name` (`/Geomancy/`, :147). The `word` assertion at :137 reads skill 3's `sijren`, set at :59 in that file's own `skill` builder — not the `synthesis-reveal.test.ts:50` one, which is a separate suite. Either way the four generated fields are unread there, so the conversion is safe.

Also picked up the other two: the fixture header now names `progression-test-utils.ts` and its divergent `xpGrowth: 1.4` / blank conlang defaults, pointing at #2426, and the exported builders' parameter is `overrides` to match `fixtures/attributes.ts`.

The `WorkbenchProficiency = IProficiency` catch is the one worth calling out — it makes the PR's original "one place to edit" claim wrong, so I've corrected the description to say six places narrow to two, not one.

---

Issue #2410, opened 2026-07-25
session-start hook can't combine Docker Postgres with native Redis (or vice versa), so a half-Docker machine gets no marker


ginderjeremiah:
Carried over from #2397's "explicitly out of scope" list so it isn't lost when that issue closes. This is a gap in `.claude/hooks/session-start.sh` itself, not in the test harness.

## Problem

The hook evaluates the two strategies as all-or-nothing pairs:

```sh
if $DOCKER_POSTGRES_OK && $DOCKER_REDIS_OK; then
  write_marker "$POSTGRES_PORT" "$REDIS_PORT"
  ...
fi
# ── Fallback: natively-installed PostgreSQL + Redis ──
...
if $NATIVE_POSTGRES_OK && $NATIVE_REDIS_OK; then
  write_marker "$NATIVE_POSTGRES_PORT" "$REDIS_PORT"
```

So a machine where Docker brings **Postgres** up but Redis fails (rate-limited image, say) and which has Redis installed **natively** ends up with both services actually running and no marker written — the integration fixtures then fall back to Testcontainers and fail on the very sandbox the hook exists to work around. The mirror case (native Postgres + Docker Redis) has the same shape.

The current behaviour is not wrong, just incomplete: partial bring-up correctly writes **no** marker rather than a half-true one (that invariant is what #1813 / #2395 established and what `session-start.test.sh` case 2 pins). The gap is that the hook never tries to complete the pair from the other strategy before giving up.

## Suggested direction

Track the two services independently rather than as two strategy pairs — resolve Postgres (Docker, else native), resolve Redis (Docker, else native), then write the marker if both resolved, carrying whichever port each one actually landed on. The all-or-nothing marker invariant is unchanged; only the definition of "both up" widens.

Worth confirming this is wanted at all before building it: the combination needs a machine with a constrained Docker *and* a partial native install, which may simply not occur in practice. If it isn't worth the branch, closing this as not-planned is a fine outcome — but the reasoning should be recorded rather than left implicit.

## Verification

`session-start.test.sh` already has the stubs for both strategies; a case with a Redis-failing `docker` stub plus the native Redis stubs would pin whichever behaviour is chosen.

## Scope

Small — shell only, no application code.

ginderjeremiah:
Passed over this while picking up work because the issue asks for a build-vs-close decision before any code, and that's genuinely yours to make. Recording my read so the reasoning is on the record either way (per the issue's own ask that it not be left implicit).

[...]

Issue #2408, opened 2026-07-25
Authoring guidance: a tier blurb must not pre-empt its word's translation reveal


ginderjeremiah:
Raised in the review of #2406 (which closed #2404 by rendering the authored `description` on the Lexicon). Filing so it's on the record before the first blurb is actually written, rather than living in a merged PR thread.

## The tension

#2406 renders a proficiency's `description` as the word-detail inspector's lede, **ungated by decipher stage**. That call is right for the reasons argued there — the tier's name, state pill, XP, payout ladder, and milestone skills are already legible at `undeciphered`, so gating prose about the same tier would be inconsistent, and gating it would hide the blurb for the first half of every tier's life.

But it leaves an authoring hazard the code can't police: **a blurb is prose by the same author about the same tier, and the `translation` is that word's meaning.** A tier whose translation is revealed at max level as "The Riven Frost" is undermined by a level-1-visible blurb reading "the art of riving frost from still air". The mechanical payouts in the ladder can't spoil the translation; prose easily can.

This is authoring discipline, not a code defect — there is nothing to gate or validate, and inventing a lint for it would be over-engineering a judgment call.

## Why nothing was done in #2406

`content/paths.json` and `content/proficiencies.json` are both **empty** today, so there is no live content to audit and no drift to correct. Adding guidance for content that doesn't exist yet, inside a PR about a render surface, would have been scope the issue didn't ask for.

## Work

- Add a short paragraph to `docs/content-design.md` covering how tier/path prose relates to the decipher arc: the blurb describes what the tier **is** and what training it feels like; the `translation` is the reveal the decipher arc pays off, so a blurb should not paraphrase it. Path blurbs have no such constraint (a path has no word of its own — the rail reuses the root tier's).
- Worth pairing with a line on what the two fields are *for*, since `Path.Description` and `Proficiency.Description` now have exactly one player-facing home each and no authoring guidance at all.

Best done alongside the first real proficiency content, where the guidance can be written against actual examples rather than hypotheticals.

ginderjeremiah:
Passed over this while picking up work, and recording why rather than leaving it silently skipped.

The issue's own closing line — *"best done alongside the first real proficiency content, where the guidance can be written against actual examples rather than hypotheticals"* — is a genuine prerequisite, and it's still unmet: `content/paths.json` and `content/proficiencies.json` are both still literally `[]` on `main` as of `8916dda`. So the state that made #2406 defer this hasn't changed.

[...]

PR #2427, opened 2026-07-25
Skip the event and write-behind save for a no-op favorite toggle (#2415)

### LINKED ISSUE #2415: Inventory.TrySetFavorite bypasses the no-op-change guard — a same-value toggle still raises an event and enqueues a write-behind save
## Problem

`Inventory.TrySetFavorite` (`Game.Core/Players/Inventories/Inventory.cs:174-184`) returns `true` whenever the item is unlocked, regardless of whether `Favorite` actually changed:

```csharp
unlocked.Favorite = favorite;
return true;
```

so `Player.TrySetFavorite` (`Game.Core/Players/Player.cs:413-422`) raises `ItemFavoriteChangedEvent` and `PlayerService`'s `SaveIf(...)` enqueues a write-behind save even when nothing mutated. A client (or a tampered client spamming the socket command) generates a redundant event + save per identical toggle.

This is the same pattern #2241 fixed for `SaveLogPreferences` — `Player.UpdateLogPreference`'s comment even describes returning changed-only as "matching the 'no mutation → no save' pattern every sibling command follows", and `SetAutoChallengeBoss` implements the guard. The favorite toggle appears to be the last player mutation without it.

## Fix

Return `false` (or a distinct "unchanged" outcome) when `unlocked.Favorite == favorite`, so no event is raised and no save is enqueued. Existing tests cover set/unset/not-unlocked but not the same-value case — add a test pinning the no-op.

ginderjeremiah:
Closes #2415.

## Summary

`Inventory.TrySetFavorite` returned `true` whenever the item was unlocked, regardless of whether `Favorite` actually changed, so `Player.TrySetFavorite` raised `ItemFavoriteChangedEvent` and `PlayerService`'s `SaveIf(...)` enqueued a write-behind save even when nothing mutated. A client (or a tampered one spamming the socket command) generated a redundant event + save per identical toggle. Verified against the tree — the issue's description of the defect is accurate.

## One deliberate departure from the issue's suggested fix

The issue proposes returning `false` on a same-value toggle. **That alone would be a client-facing regression**, which is why the fix takes the issue's parenthetical ("or a distinct 'unchanged' outcome") instead:

`SetItemFavorite.ExecuteAsync` maps the service's `false` straight to `Error("Failed to set item favorite.")`. Returning `false` for a no-op would therefore turn an idempotent toggle into an error response, and the frontend's `InventoryManager.setFavorite` writes any `response.error` to the debug log — so a harmless re-toggle would surface as a spurious client error. Returning `false` also conflates it with the genuine anti-cheat rejection (an item the player doesn't own), which *must* keep failing.

So rejected and accepted-but-unchanged have to stay distinguishable. `SetFavorite` now returns a three-valued `SetFavoriteOutcome` (`ItemNotUnlocked` / `Unchanged` / `Changed`), and `PlayerService.SetFavorite` maps it exactly the way `BattleService.SetAutoChallengeBoss` already does for the same problem — persist only on `Changed`, report failure only on `ItemNotUnlocked`. That method's own doc comment is the precedent this follows verbatim: *"A toggle to the already-active mode saves nothing (no mutation, no event) but still returns `true` — accepting is not the same as changing."*

Net client-visible behavior is unchanged: `Success` for a no-op (as today), `Error` for an unowned item (as today). Only the redundant event and save go away.

## Also in this PR

- **Renamed off the `Try*` prefix** on both `Inventory` and `Player` (`TrySetFavorite` → `SetFavorite`). The C# convention implies a `bool` return, so a `Try*` method handing back a three-valued enum reads wrong; `PlayerService.SetFavorite` and `Player.SetAutoChallengeBoss` were already named this way. The stale "the Try* methods below" reference in `Inventory`'s index doc comment was corrected in the same pass.
- **One doc paragraph** in `docs/backend.md` → *HTTP vs WebSocket Communication*, next to the existing "validates as anti-cheat (rejecting with no mutation)" line. Rejected-vs-accepted-but-unchanged is now a recurring pattern across three commands (`SetAutoChallengeBoss`, `SaveLogPreferences`, `SetItemFavorite`) with nothing written down, and conflating the two is precisely the bug fixed here — so it's the kind of expectation worth stating once rather than re-deriving.

No parity surface (server-only, no battle logic), no wire-contract change — `SetFavoriteOutcome` is not `[ClientMirrored]` and is unreachable from the codegen's DTO walk, confirmed by the generated client being untouched in the diff.

## Test plan

- [x] `dotnet build` — clean, 0 errors / 0 warnings.
- [x] **Full backend suite, all green: 3,496 passed / 0 failed** (Core 1446, Api 871, Application 1105, Infrastructure 74).
- [x] **Domain unit tests** (`InventoryTests`, `PlayerTests`): existing set/unset/not-unlocked cases converted to assert the outcome; four new cases pin the no-op — `Unchanged` on an already-favorited item and on an already-unfavorited one, at both the `Inventory` and `Player` level, with `Player`'s asserting **no `ItemFavoriteChangedEvent` is raised**. Both sides of the no-op are covered because a client re-sending the default state hits the unfavorite side.
- [x] **Integration tests** (`PlayerServiceIntegrationTests`) pinning the interaction the unit tests can't see — that no *save* is enqueued. `SetFavorite_SameValue_IsNoOpButStillReturnsTrue` asserts the Redis queue length is 1 after the real toggle and 0 after the repeat, following the existing `SaveLogPreferences_NoValuesChanged_DoesNotEnqueueWriteBehindEvents` / `SetAutoChallengeBoss_AlreadyActive_...` precedent — including their trick of asserting the *first* call does enqueue, so the no-op assertion can't pass trivially off a broken queue. `SetFavorite_ItemNotUnlocked_ReturnsFalseAndEnqueuesNothing` pins that the rejection still returns `false` (i.e. still reaches the client as an error), which is the behavior most at risk from this change.

## Note on selection

Picked as the oldest eligible unassigned `claude` issue. Two older ones were passed over deliberately rather than silently: **#2408** is gated on content that doesn't exist yet (`content/proficiencies.json` and `content/paths.json` are both still `[]`, and the issue itself asks for the guidance to be written against real examples), and **#2410** explicitly asks for a build-vs-close decision before any code — a prior session already recorded its read there and is waiting on a call.

---
_Generated by [Claude Code](https://claude.ai/code/session_015QFkGxVedj7ZS6cChYLAT2)_


ginderjeremiah (commented):
## Review: no-op favorite toggle no longer raises an event or enqueues a write-behind save (#2415)

**Verdict:** Approve — the defect is real, the fix is minimal and lands the guard in the domain where it belongs, and the departure from the issue's literal suggestion is the correct call. No blocking findings, no should-fixes, no nits.

_(Submitted as a `COMMENT` because the reviewing account is the PR author — GitHub won't accept a gating verdict on one's own PR. The intended verdict is **approve**.)_

### Verification

Checked against the tree rather than taking the description at face value:

- The defect is as #2415 describes — `Inventory.TrySetFavorite` returned `true` on an unlocked item unconditionally, so `Player` raised `ItemFavoriteChangedEvent` and `SaveIf` enqueued a save for a same-value toggle.
- The departure from the issue's "return `false`" suggestion is justified, not a shortcut. `SetItemFavorite.ExecuteAsync:26-28` maps `false` straight to `Error("Failed to set item favorite.")`, so a bare `false` would turn an idempotent toggle into a client-visible error *and* conflate it with the anti-cheat rejection that must keep failing. The three-valued outcome is the right shape, and it matches the issue's own parenthetical.
- Net client contract is unchanged: `Success` for a no-op, `Error` for an unowned item. Confirmed both integration tests pin exactly those two.
- `SetFavoriteOutcome` naming and placement fit the existing convention for non-value enums (`CreatePlayerOutcome`, `SetUserRolesStatus`) — the `E` prefix is for intrinsic reference-data enums, so its absence here is correct.
- No stragglers: `TrySetFavorite` has zero remaining references anywhere in the repo, including tests and the frontend.
- `PlayerService.SetFavorite` discarding `SaveIf`'s return is consistent with `SaveLogPreferences` / `UnlockLesson` / `MarkLessonRead`, and `SaveIf` returns its own input, so nothing is lost.
- Skipping `SavePlayer` on a no-op doesn't strand buffered events: the aggregate raises none on that path, and this is the same contract `SetAutoChallengeBoss` and the `SaveLogPreferences` no-op already rely on.
- No parity surface — the event drives only `PlayerPersistencePublisher` / `ItemFavoriteChangedHandler`, never a client push, so there's no frontend mirror to keep in step and no battle logic involved. Nothing in `Game.Api` or the generated client changes.

### Test coverage

Sufficient and correctly layered:

- Domain unit tests at both levels (`InventoryTests`, `PlayerTests`), with the no-op pinned on **both** sides — already-favorited and already-unfavorited. The second case matters more than it looks: a client resending the default state hits the unfavorite side, and a guard written as `if (favorite && unlocked.Favorite)` would pass the first test and fail that one.
- `PlayerTests.SetFavorite_SameValue_RaisesNoEvent` asserts the absence of `ItemFavoriteChangedEvent`, which is the actual domain-level contract, not just the return value.
- The interaction the unit tests structurally can't see — that no *save* is enqueued — is covered by integration tests against the real Redis queue, following the `SaveLogPreferences_NoValuesChanged_DoesNotEnqueueWriteBehindEvents` precedent verbatim, including asserting the first call *does* enqueue so the no-op assertion can't pass off a broken queue.
- `SetFavorite_ItemNotUnlocked_ReturnsFalseAndEnqueuesNothing` pins the behavior most at risk from this change (the rejection still reaching the client as an error).

### What's good

- The rejected-vs-accepted-but-unchanged distinction is the whole point of the bug, and the enum's XML docs make it un-mis-readable at the call site — a later "simplify this to a bool" edit now has to argue with the type.
- The `Try*` rename is the right cleanup to bundle here: a `Try*` method handing back a three-valued enum would be a standing misdirection, and the stale "the Try* methods below" index doc comment got fixed in the same pass rather than left to rot.
- The `docs/backend.md` paragraph earns its place. Three commands now share this pattern with nothing written down, and conflating the two outcomes is precisely the defect being fixed — that's the "recurring pattern worth stating once" case, not over-documentation.

---

PR #2428, opened 2026-07-25
Clamp CombatRating's offense-side proc chances to probability 1 (#2416)

### LINKED ISSUE #2416: CombatRating prices offense-side proc chances unclamped past probability 1 — asymmetric with the #2171 survivability cap
## Problem

#2171 capped the **survivability** side's avoidance/resist credits (`ServerGameConstants.MaxMitigationCredit`, applied at `Game.Core/Battle/CombatRating.cs:284-285`) because the authored-only enablers × derived multipliers can be pushed past probability 1 by the shared-expiry effect ramp. The **offense** side still prices the same products linearly with no clamp:

- the riposte term (`CombatRating.cs:211-216`): `effectiveParryChance = ParryChance × ParryChanceMultiplier` feeds the counter's expected hit unclamped;
- the crit expectation (`CombatRating.cs:245-248`): `skill.CriticalChance × CriticalChanceMultiplier` likewise.

The live engine caps any proc at probability 1 (a `[0,1)` draw always procs at chance ≥ 1), so a chance product above 1 inflates `OffenseRate` beyond any achievable output.

## Effect

An over-ramped parry/crit build's `PlayerRating` overstates its real capability, which — via `min(EnemyRating ÷ PlayerRating, 1)²` and the `max(PlayerRating, EnemyRating)` proficiency denominator — **undercuts that player's XP and proficiency gains**. A mispricing rather than the #2171 divergence-to-zero pole, so lower stakes, but the same root cause left half-fixed.

## Fix

Clamp the two chance products to 1 at the read sites (or via a shared clamp helper used by both the offense terms and the survivability credits), aligning the rating with what the engine can actually deliver. Server-only pricing — no parity surface.

ginderjeremiah:
Closes #2416.

## Summary

`CombatRating` priced two offense-side proc chances as raw products with no clamp:

- the riposte term (`CombatRating.cs:211-216`) — `ParryChance × ParryChanceMultiplier`;
- the crit expectation (`CombatRating.cs:245-248`) — `skill.CriticalChance × CriticalChanceMultiplier`.

The engine resolves every proc as `_rng.Next() < chance` against a `[0, 1)` draw (`BattleContext.cs:230-233` for parry/dodge, `:308` for crit), so a product at or above 1 always fires and one at or below 0 never does. Pricing the raw product credits capability the engine cannot deliver — inflating `OffenseRate`, which via `min(EnemyRating ÷ PlayerRating, 1)²` and the `max(PlayerRating, EnemyRating)` proficiency denominator **undercuts that player's XP and proficiency gains**. #2171 already capped the survivability side's credits; this is the other half of the same root cause.

Adds a shared `ProcChance` helper (`Math.Clamp(chance, 0, 1)`) and applies it at every chance read site.

## How it addresses the issue

Both sites the issue names are clamped, via the shared helper it suggests. Two deliberate calls beyond the literal ask:

**Clamped to `[0, 1]`, not just to 1.** A negative chance can't proc either — the draw is non-negative — and neither `ParryChance` nor `CriticalChance` is floored anywhere (`AttributeCollection` does no clamping), so an `Opponent`-targeted debuff can drive one below zero. Unclamped, the riposte term then contributed *negative* offense. Same one-line cost, so it's covered rather than left as a second issue.

**The survivability parry/dodge products are clamped too.** The issue frames this as offense-only, but `avoid = parry + (1 - parry) × dodge` is a sequential-probability composition and only means anything for probabilities: an unclamped parry above 1 makes `(1 - parry)` negative, so dodge *subtracts* avoidance. Concretely, `ParryChance 1.01` + a ramped dodge of 50 gave `avoid ≈ 0.51` — a certain-parry build priced as barely evasive. The existing `MaxMitigationCredit` cap didn't catch it because that value never crossed 0.95. The cap itself is unchanged and still applies on top; the clamp only fixes what's fed into it.

The helper's comment states why this clamp is structural (the definition of a probability) rather than a tuned ceiling like `MaxMitigationCredit`, so a future reader doesn't move it into `ServerGameConstants`.

Server-only pricing — the rating carries no frontend parity surface, so there is no mirrored frontend change.

## Verification of the issue's claims

Per CLAUDE.md, the issue's assertions were checked against the tree rather than taken as given. All three hold: both offense read sites are unclamped as described; the survivability credits are capped at `CombatRating.cs:284-285`; and the engine's `[0, 1)` draw semantics are as claimed. The one thing the issue understates is the survivability composition bug above.

## Test plan

Five new cases in `Game.Core.Tests/Battle/CombatRatingTests.cs`, covering both terms, both directions of the clamp, and the composition:

- [x] `Rate_RiposteChanceBeyondCertainty_RatesTheSameAsExactlyAtCertainty` — `ParryChance` 1.0 vs 5.0 with a counter skill.
- [x] `Rate_NegativeParryChance_IsPricedAsNoRiposte` — a below-zero chance rates identically to no parry at all.
- [x] `Rate_CriticalChanceBeyondCertainty_RatesTheSameAsExactlyAtCertainty` — authored `CriticalChance` 1.0 vs 5.0.
- [x] `Rate_CriticalChanceRampedPastCertaintyByMultiplier_RatesTheSameAsExactlyAtCertainty` — the realistic route past 1 (0.5 × 2 vs 0.5 × 4).
- [x] `Rate_ParryPastCertaintyDoesNotLetDodgeSubtractAvoidance` — the survivability composition case.

- [x] Each new test confirmed to **fail** with the clamp neutered to the identity function (exactly 5 failures, no more), then pass with it restored — so none of them is vacuously green.
- [x] `dotnet build` clean, 0 warnings.
- [x] Full suite green: `Game.Core.Tests` 1451, `Game.Application.Tests` 1105, `Game.Api.Tests` 871, `Game.Infrastructure.Tests` 74 — 0 failures.
- [x] The existing #2171 cap tests (`Rate_ParryBeyondCertainty_RatesTheSameAsExactlyAtTheCap`, `Rate_DodgeBeyondCertainty_DoesNotExplodeRating`, …) still pass unchanged — the clamp composes with the cap rather than replacing it.

No frontend tests: server-only, no parity surface.

https://claude.ai/code/session_01T8uJc1ebma8uwdXCAk6rRJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01T8uJc1ebma8uwdXCAk6rRJ)_

### DIFF
--- a/Game.Core.Tests/Battle/CombatRatingTests.cs
+++ b/Game.Core.Tests/Battle/CombatRatingTests.cs
@@ -119,6 +119,80 @@ namespace Game.Core.Tests.Battle
             Assert.False(double.IsInfinity(rate));
         }
 
+        // ── Rate: offense-side proc-chance clamp (#2416) ────────────────────────
+
+        [Fact]
+        public void Rate_RiposteChanceBeyondCertainty_RatesTheSameAsExactlyAtCertainty()
+        {
+            // The engine draws a parry from [0, 1), so ParryChance 1.0 already parries every incoming attack —
+            // authoring more cannot buy additional riposte fires. The offense-side riposte term must price the
+            // clamped chance, not the raw product (which would scale the counter's expected hit linearly).
+            var counter = MakeSkill(cooldownMs: 1000, baseDamage: 20);
+            var atCertainty = MakeBattlerWithSkills([(ParryChance, 1.0)], [], counterSkill: counter);
+            var beyondCertainty = MakeBattlerWithSkills([(ParryChance, 5.0)], [], counterSkill: counter);
+
+            Assert.Equal(
+                CombatRating.Rate(atCertainty, isPlayer: true),
+                CombatRating.Rate(beyondCertainty, isPlayer: true), 6);
+        }
+
+        [Fact]
+        public void Rate_NegativeParryChance_IsPricedAsNoRiposte()
+        {
+            // A debuffed-below-zero ParryChance never procs in the engine, so the riposte term must contribute
+            // nothing — an unclamped product would price it as a negative offense contribution instead.
+            var counter = MakeSkill(cooldownMs: 1000, baseDamage: 20);
+            var negativeParry = MakeBattlerWithSkills([(ParryChance, -0.5)], [], counterSkill: counter);
+            var noParry = MakeBattlerWithSkills([], [], counterSkill: counter);
+
+            Assert.Equal(
+                CombatRating.Rate(noParry, isPlayer: true),
+                CombatRating.Rate(negativeParry, isPlayer: true), 6);
+        }
+
+        [Fact]
+        public void Rate_CriticalChanceBeyondCertainty_RatesTheSameAsExactlyAtCertainty()
+        {
+            // Same clamp on the crit expectation: a skill authored at CriticalChance 1.0 already crits every
+            // fire, so the CriticalDamage premium must not keep scaling past it.
+            var atCertainty = MakeBattlerWithSkills([], [MakeSkill(cooldownMs: 1000, baseDamage: 50, criticalChance: 1.0)]);
+            var beyondCertainty = MakeBattlerWithSkills([], [MakeSkill(cooldownMs: 1000, baseDamage: 50, criticalChance: 5.0)]);
+
+            Assert.Equal(
+                CombatRating.Rate(atCertainty, isPlayer: true),
+                CombatRating.Rate(beyondCertainty, isPlayer: true), 6);
+        }
+
+        [Fact]
+        public void Rate_CriticalChanceRampedPastCertaintyByMultiplier_RatesTheSameAsExactlyAtCertainty()
+        {
+            // The realistic route past 1: a modestly authored CriticalChance times a Luck-ramped
+            // CriticalChanceMultiplier. 0.5 × 2 already reaches certainty, so 0.5 × 4 must rate identically.
+            var atCertainty = MakeBattlerWithSkills(
+                [(CriticalChanceMultiplier, 1.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50, criticalChance: 0.5)]);
+            var beyondCertainty = MakeBattlerWithSkills(
+                [(CriticalChanceMultiplier, 3.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50, criticalChance: 0.5)]);
+
+            Assert.Equal(
+                CombatRating.Rate(atCertainty, isPlayer: true),
+                CombatRating.Rate(beyondCertainty, isPlayer: true), 6);
+        }
+
+        [Fact]
+        public void Rate_ParryPastCertaintyDoesNotLetDodgeSubtractAvoidance()
+        {
+            // The survivability composition `parry + (1 - parry) × dodge` is only meaningful for probabilities:
+            // an unclamped parry above 1 makes (1 - parry) negative, so adding dodge would *lower* the credit.
+            // With both clamped, certain parry already avoids everything and dodge cannot reduce it.
+            var parryOnly = MakeBattlerWithSkills([(ParryChance, 1.01)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+            var parryAndDodge = MakeBattlerWithSkills(
+                [(ParryChance, 1.01), (DodgeChance, 50.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+
+            Assert.Equal(
+                CombatRating.Rate(parryOnly, isPlayer: true),
+                CombatRating.Rate(parryAndDodge, isPlayer: true), 6);
+        }
+
         // ── Rate: offense ─────────────────────────────────────────────────────
 
         [Fact]
--- a/Game.Core.Tests/Battle/CombatRatingTests.cs
+++ b/Game.Core.Tests/Battle/CombatRatingTests.cs
@@ -179,11 +179,44 @@ namespace Game.Core.Tests.Battle
         }
 
         [Fact]
+        public void Rate_NegativeDodgeChance_IsPricedAsNoDodge()
+        {
+            // The dodge mirror of the negative-parry case, and the only direction in which the dodge clamp is
+            // observable at all: for any parry in [0, 1] a dodge above 1 already drives `avoid` past
+            // MaxMitigationCredit clamped or not, so the cap flattens the positive direction. A negative dodge
+            // instead makes `avoid` negative, inflating the mitigation fraction above 1 and understating the
+            // rating — the engine simply never dodges.
+            var negativeDodge = MakeBattlerWithSkills([(DodgeChance, -0.5)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+            var noDodge = MakeBattlerWithSkills([], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+
+            Assert.Equal(
+                CombatRating.Rate(noDodge, isPlayer: true),
+                CombatRating.Rate(negativeDodge, isPlayer: true), 6);
+        }
+
+        [Fact]
+        public void Rate_NegativeCriticalChanceMultiplier_IsPricedAsNoCrit()
+        {
+            // CriticalDamage has a base of 1.5, so (CriticalDamage - 1) is non-zero and a debuffed-negative
+            // multiplier is observable: unclamped it would price crit as a damage *penalty* on every fire.
+            var negativeCrit = MakeBattlerWithSkills(
+                [(CriticalChanceMultiplier, -2.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50, criticalChance: 0.5)]);
+            var noCrit = MakeBattlerWithSkills(
+                [(CriticalChanceMultiplier, -2.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50, criticalChance: 0.0)]);
+
+            Assert.Equal(
+                CombatRating.Rate(noCrit, isPlayer: true),
+                CombatRating.Rate(negativeCrit, isPlayer: true), 6);
+        }
+
+        [Fact]
         public void Rate_ParryPastCertaintyDoesNotLetDodgeSubtractAvoidance()
         {
             // The survivability composition `parry + (1 - parry) × dodge` is only meaningful for probabilities:
             // an unclamped parry above 1 makes (1 - parry) negative, so adding dodge would *lower* the credit.
-            // With both clamped, certain parry already avoids everything and dodge cannot reduce it.
+            // With both clamped, certain parry already avoids everything and dodge cannot reduce it. Note this
+            // pins the *parry* clamp only — once parry pins to 1, (1 - parry) annihilates the dodge term either
+            // way; Rate_NegativeDodgeChance_IsPricedAsNoDodge is what covers the dodge clamp.
             var parryOnly = MakeBattlerWithSkills([(ParryChance, 1.01)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
             var parryAndDodge = MakeBattlerWithSkills(
                 [(ParryChance, 1.01), (DodgeChance, 50.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
--- a/Game.Core/Battle/CombatRating.cs
+++ b/Game.Core/Battle/CombatRating.cs
@@ -55,6 +55,18 @@ namespace Game.Core.Battle
         // zero-survivability battler never divides a downstream consumer (e.g. a rating ratio) by zero.
         private const double Epsilon = 1e-6;
 
+        // A proc chance as the engine can actually deliver it. Every proc draw is `_rng.Next() < chance`
+        // against a [0, 1) draw (BattleContext.DamageTarget/ResolvePlayerHit), so a product at or above 1
+        // always fires and one at or below 0 never does. The authored-only enablers (ParryChance,
+        // DodgeChance, a skill's CriticalChance) are uncapped and their derived multipliers can be ramped
+        // by shared-expiry effect stacking, so the raw product overshoots what the engine can produce —
+        // pricing it unclamped would credit unreachable capability. Structural, not tunable: this is the
+        // definition of a probability, unlike ServerGameConstants.MaxMitigationCredit's tuned ceiling.
+        private static double ProcChance(double chance)
+        {
+            return Math.Clamp(chance, 0.0, 1.0);
+        }
+
         /// <summary>
         /// The rating for <paramref name="battler"/> as assembled — <c>√(OffenseRate × Survivability)</c>
         /// against the fixed reference profiles in <see cref="ServerGameConstants"/>. Pass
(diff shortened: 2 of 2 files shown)


Game.Core.Tests/Battle/CombatRatingTests.cs
```diff
@@ -119,6 +119,80 @@ public void Rate_ResistanceBeyondCertainty_DoesNotExplodeRating()
...
+            // The survivability composition `parry + (1 - parry) × dodge` is only meaningful for probabilities:
+            // an unclamped parry above 1 makes (1 - parry) negative, so adding dodge would *lower* the credit.
+            // With both clamped, certain parry already avoids everything and dodge cannot reduce it.
+            var parryOnly = MakeBattlerWithSkills([(ParryChance, 1.01)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+            var parryAndDodge = MakeBattlerWithSkills(
+                [(ParryChance, 1.01), (DodgeChance, 50.0)], [MakeSkill(cooldownMs: 1000, baseDamage: 50)]);
+
+            Assert.Equal(
+                CombatRating.Rate(parryOnly, isPlayer: true),
+                CombatRating.Rate(parryAndDodge, isPlayer: true), 6);
```
ginderjeremiah, line 226: **Should-fix** — this test reads like it covers the dodge clamp, but it can't: `ProcChance` pins parry to exactly `1`, so `(1 - parry) = 0` annihilates the dodge term whether dodge is `50` or the clamped `1`. Both arrangements produce `avoid = 1` either way, and the assertion passes on the *parry* clamp alone.

That leaves the dodge clamp at `CombatRating.cs:288-289` with **zero** coverage, and it's not just this test — it's unobservable in the positive direction entirely. For any post-clamp `parry ∈ [0, 1]` and `dodge > 1`, `parry + (1 - parry) × dodge ≥ 1` clamped *and* unclamped, and both land above `MaxMitigationCredit = 0.95`, so the cap flattens the difference before it can reach the rating. The dodge clamp only changes an outcome in the **negative** direction — which is exactly the direction this PR deliberately went beyond the issue to cover, and which got a test for parry (`Rate_NegativeParryChance_IsPricedAsNoRiposte`) but not for dodge.

Neutering only the dodge branch of `ProcChance` today yields 0 failures, so per CLAUDE.md's "thoroughly tested … including edge cases", please add the mirror case — something like a `DodgeChance -0.5` battler rating identically to one with no dodge at all. Small enough that it belongs in this PR rather than a follow-up.

---

Game.Core/Battle/CombatRating.cs
```diff
@@ -264,12 +277,16 @@ private static double Survivability(bool isPlayer, Battler effectiveCaster)
...
+            // parry product above 1 would turn (1 - parry) negative and have dodge *subtract* avoidance.
             var avoid = 0.0;
             if (isPlayer)
             {
-                var parry = effectiveCaster.GetAttributeValue(ParryChance) * effectiveCaster.GetAttributeValue(ParryChanceMultiplier);
-                var dodge = effectiveCaster.GetAttributeValue(DodgeChance) * effectiveCaster.GetAttributeValue(DodgeChanceMultiplier);
+                var parry = ProcChance(
+                    effectiveCaster.GetAttributeValue(ParryChance) * effectiveCaster.GetAttributeValue(ParryChanceMultiplier));
+                var dodge = ProcChance(
+                    effectiveCaster.GetAttributeValue(DodgeChance) * effectiveCaster.GetAttributeValue(DodgeChanceMultiplier));
```
ginderjeremiah: **Nit (DRY)** — `ProcChance(GetAttributeValue(ParryChance) × GetAttributeValue(ParryChanceMultiplier))` now appears verbatim here and at `:225-226`. The bare product was already duplicated pre-PR, but wrapping it pushed both sites to two wrapped lines, so this is the natural moment to collapse it:

```csharp
private static double EffectiveParryChance(Battler caster)
{
    return ProcChance(caster.GetAttributeValue(ParryChance) * caster.GetAttributeValue(ParryChanceMultiplier));
}
```

Both call sites become one readable line, and the offense and survivability readings become structurally incapable of drifting apart — which matters here, since the *whole* bug being fixed is these two sites disagreeing about what a parry chance means. A `EffectiveDodgeChance` sibling for symmetry, though it has only the one caller.

---

ginderjeremiah (commented):
## Review: clamp `CombatRating`'s proc-chance products to `[0, 1]` (#2416)

**Verdict:** Approve (posted as a comment — GitHub won't accept a gating verdict from the PR author) — the fix is correct at the root cause, the engine semantics it rests on check out against the tree, and the two findings below are a missing test and a DRY tidy, neither of which blocks.

I verified the change against the engine rather than the description. All the load-bearing claims hold:

- `Mulberry32.Next()` returns `(t ^ (t >> 14)) / 4294967296.0` — genuinely `[0, 1)`, so `draw < chance` saturates at 1 and floors at 0 exactly as the helper's comment says.
- Every proc read site is covered: parry/dodge at `BattleContext.cs:230-233`, crit at `:308`. `ExecuteBonus` (`CombatRating.cs:266`) is correctly left alone — it's a damage multiplier, not a probability, and the engine already clamps its missing-health fraction at `BattleContext.cs:317`.
- The `MaxMitigationCredit` cap is untouched and still composes on top; the pre-existing #2171 tests still describe live behavior.
- Server-only is right: there is no `CombatRating` mirror under `UI/src/lib/battle/`, so the no-parity-surface claim is real and not just asserted.

### Should-fix

- **The dodge clamp has no test coverage** — `CombatRating.cs:288-289`, see the inline note. `Rate_ParryPastCertaintyDoesNotLetDodgeSubtractAvoidance` looks like it covers it but doesn't (the clamped parry zeroes `(1 - parry)`, so dodge drops out either way), and the positive direction is unobservable through the `MaxMitigationCredit` cap regardless. Only a negative dodge can move the rating, and nothing tests that. One test, same shape as the parry case you already wrote.

### Nits

- **Duplicated chance composition** — `CombatRating.cs:225-226` / `:286-289`, inline. An `EffectiveParryChance(Battler)` helper collapses both sites and makes the two readings structurally unable to diverge, which is the exact failure mode this PR exists to fix.
- **The crit clamp's negative direction is untested too** — `CriticalDamage` has a base of `1.5` (`StaticAttributeModifiers.cs:90`), so `(CriticalDamage - 1) = 0.5` and a debuffed-negative `CriticalChanceMultiplier` *is* observable in `ExpectedDirectHit`. Lower value than the dodge case since the shape is already proven by the parry test; fold it in if you're touching the file anyway.

### Follow-up (not this PR)

The class doc-comment stakes its correctness on reusing "the exact `Battler`/`BattleSkill` methods the engine itself uses — the anti-drift mechanism that keeps the rating from silently diverging from the real damage pipeline." The chance composition is now the one piece of engine math `CombatRating` re-derives by hand: `ParryChance × ParryChanceMultiplier` is spelled out independently at `BattleContext.cs:230-233` and here. A shared `Battler.EffectiveParryChance` / `EffectiveDodgeChance` would put it under that same anti-drift rule — but it touches the parity-mirrored engine (`battle-step.ts` would need the mirror), so it's a bounded refactor of its own rather than scope for a `scope: small` clamp fix. Worth an issue (`tech debt`, `claude`, `scope: small`) if you agree.

### Test coverage

Five new cases in `Game.Core.Tests/Battle/CombatRatingTests.cs`, backend-only — correct, since there's no frontend mirror to keep in sync. The four that do bite are well-chosen and genuinely isolate their terms rather than passing vacuously:

- `Rate_RiposteChanceBeyondCertainty_…` is properly offense-specific — at both `1.0` and `5.0` the survivability side caps to the same `MaxMitigationCredit`, so only the riposte term can move the assertion.
- `Rate_CriticalChanceRampedPastCertaintyByMultiplier_…` is the right addition next to the authored-chance case: `0.5 × 2` vs `0.5 × 4` is the route content will actually take past 1, and it exercises the multiplier read rather than just the skill field.
- The neuter-the-helper check in the test plan is the correct way to prove these aren't vacuous — it's just that "exactly 5 failures" is also the evidence that the dodge branch is unexercised.

Gap is the should-fix above. `Marginal`'s dead-stat consumer is unaffected, since `ProgressionGraphChecker.cs:286` rates enemies (`isPlayer: false`) and all four clamped terms are `AsymmetryGated` — no new false "inert attribute" warnings.

### What's good

- Fixing this via a named `ProcChance` helper rather than two inline `Math.Min`s is the right call, and the comment earns its length by answering the question a future reader will actually have: why this is structural (the definition of a probability) and not a tunable that belongs in `ServerGameConstants` next to `MaxMitigationCredit`.
- Catching that the survivability composition was broken too is the best part of this PR. `parry + (1 - parry) × dodge` going *backwards* for an over-ramped parry build is a strictly worse bug than the one the issue described, it was invisible to the existing cap because `avoid` never crossed `0.95` on that path, and it was outside the issue's stated scope. Exactly the "verify the issue's claims rather than implement them" posture CLAUDE.md asks for.
- Clamping both directions instead of only capping at 1 is correctly justified — neither `ParryChance` nor `CriticalChance` is floored in `AttributeCollection`, so an `Opponent`-targeted debuff really can drive them negative, and the unclamped riposte term really would have contributed negative offense.

---

ginderjeremiah:
Thanks — the should-fix is right and I've taken all three findings. Pushed in `cedda41`.

**The dodge clamp had no coverage.** Confirmed exactly as you described, and I verified it rather than taking it on faith: neutering *only* `EffectiveDodgeChance` against the pre-fix suite yields **0** failures. `Rate_ParryPastCertaintyDoesNotLetDodgeSubtractAvoidance` does bite, but on the parry clamp alone — once parry pins to `1`, `(1 - parry)` annihilates the dodge term whether dodge is `50` or `1`. And your reasoning about the positive direction holds: for any post-clamp `parry ∈ [0, 1]`, a `dodge > 1` puts `avoid ≥ 1` clamped or not, so `MaxMitigationCredit` flattens the difference before it can reach the rating. Negative really is the only observable direction.

Added `Rate_NegativeDodgeChance_IsPricedAsNoDodge` (a `DodgeChance -0.5` battler rating identically to one with no dodge). Neutering the dodge branch now yields exactly **1** failure — that test. I also added a note to the parry-composition test saying what it actually pins, so the next reader doesn't re-draw the same wrong conclusion I did when I wrote it.

**Both nits taken.**

- `EffectiveParryChance`/`EffectiveDodgeChance` added; both call sites are one line again. Your framing is the convincing part — the bug being fixed *is* two sites disagreeing about what a parry chance means, so leaving the composition duplicated was the wrong shape for this particular PR.
- Added `Rate_NegativeCriticalChanceMultiplier_IsPricedAsNoCrit`. Worth it: neutering the crit site alone now fails **3** tests instead of 2, so the negative direction was genuinely unpinned there too.

**On the follow-up — agreed, filed as #2429.** One refinement worth recording there: the *clamp* shouldn't move onto `Battler` with the products. `CombatRating` clamps because it prices an expectation; the engine doesn't need to, since `draw < chance` already saturates naturally. Folding `ProcChance` into a shared accessor would change engine behavior for no reason — the shared piece is the product, and the clamp stays rating-side. Also noted there that `Battler` is parity-mirrored, so the `battler.ts` half comes with it, which is what keeps it out of a `scope: small` clamp fix.

**Verification:** `dotnet build` clean (0 warnings), full suite green — `Game.Core.Tests` 1453, `Game.Application.Tests` 1105, `Game.Api.Tests` 871, `Game.Infrastructure.Tests` 74, 0 failures. Each new test confirmed to fail against its own neutered branch and pass with it restored.

---

ginderjeremiah:
## Re-review of `cedda41` — all three findings taken, and the should-fix is properly closed

**Verdict:** Approve, **merging**. CI green across all 11 checks on `cedda41` (`all-checks-passed`, `verify-backend`, `test-backend`, all three `test-e2e` browsers; frontend jobs correctly skipped by the changed-path gate). One trivial nit below, routed to #2429 rather than held — reasoning in that section.

Incremental pass over the delta only.

### The should-fix: dodge clamp coverage

Closed, and I verified the claim rather than taking the count on trust. Neutering `ProcChance` to the identity function against the current branch fails **exactly 7** tests — the 5 from the first pass plus the 2 added here — and passes 1446. So `Rate_NegativeDodgeChance_IsPricedAsNoDodge` and `Rate_NegativeCriticalChanceMultiplier_IsPricedAsNoCrit` both bite, and neither is vacuously green. Restored the file and confirmed the tree is clean afterward.

The two new tests are constructed correctly:

- **`Rate_NegativeDodgeChance_IsPricedAsNoDodge`** — `DodgeChance -0.5` against a battler with no dodge at all. That's the right pairing: it isolates the dodge branch specifically, and your analysis of why the *positive* direction is unreachable holds — for any post-clamp `parry ∈ [0, 1]`, a `dodge > 1` puts `avoid ≥ 1` clamped or not, so `MaxMitigationCredit` flattens it before it reaches the rating.
- **`Rate_NegativeCriticalChanceMultiplier_IsPricedAsNoCrit`** — the control carries the same `-2.0` multiplier and varies only `criticalChance` (`0.5` → `0.0`). That's the correct isolation: it pins the clamp rather than the multiplier's presence, and `CriticalDamage`'s `1.5` base is what makes the unclamped case observable as a per-fire damage *penalty*.

The added note on `Rate_ParryPastCertaintyDoesNotLetDodgeSubtractAvoidance` saying what it actually pins is the more durable half of this fix — that test reads like dodge coverage, and without the note the next reader draws the same wrong conclusion.

### The nits

**DRY collapse.** `EffectiveParryChance`/`EffectiveDodgeChance` are in and both call sites are one line again. I confirmed the coverage is total: `grep` over `CombatRating.cs` shows the only remaining raw chance read is the crit product at `:270`, which is already wrapped in `ProcChance` and has no second site to drift against. All four read sites from the original commit are accounted for.

**Crit negative direction.** Covered, per the neuter count above.

### Follow-up #2429

Filed with the right labels (`tech debt`, `claude`, `scope: small`) and it captures the constraint accurately. The refinement you added is the correct call and worth restating: the **product** is what belongs on `Battler`, not the clamp. `CombatRating` clamps because it prices an expectation over a fire; the engine's `draw < chance` saturates naturally against a `[0, 1)` draw, so folding `ProcChance` into a shared accessor would change engine behavior — and on the parity-mirrored side, at that — for no benefit. Keeping `ProcChance` rating-side is right.

### One trivial nit, not held

`CombatRating.cs:72` — the helper comment cites the engine as `BattleContext.cs:230-233`. It lands exactly right today (I checked: `:230-233` is precisely the `isParry`/`isDodge` product pair), but it's the **only** line-numbered code citation in `Game.Core` or `Game.Application` — every other cross-reference in the codebase, including `ProcChance`'s own comment two lines above it, cites by method name (`BattleContext.DamageTarget/ResolvePlayerHit`). Line numbers rot silently, which is the failure mode the surrounding convention avoids.

Not holding the merge for it, and not because it's small: #2429 rewrites these exact two helpers to read through `Battler`, at which point this comment gets rewritten anyway. Blocking a verified fix to an XP/proficiency mispricing on a comment that has a guaranteed sweep point in an already-filed issue is the wrong trade. I've noted it on #2429 so it lands with that change rather than getting lost.

### Standing

The first review's substantive verification stands unchanged — `Mulberry32.Next()`'s `[0, 1)` range, the four clamped read sites, `ExecuteBonus` correctly left alone as a multiplier rather than a probability, `MaxMitigationCredit` composing on top rather than being replaced, and the server-only claim (no `CombatRating` mirror under `UI/src/lib/battle/`). Nothing in this delta touches any of it. Closes #2416.

---

PR #2430, opened 2026-07-25
Bring SelectPlayer/SwitchPlayer into the shared auth rate-limit budget (#2417)

### LINKED ISSUE #2417: SelectPlayer/SwitchPlayer consume refresh tokens outside the combined per-IP auth rate-limit budget
## Problem

`docs/backend-auth.md` documents the per-IP `auth` rate-limit policy as blunting refresh-token brute force via a **combined per-IP budget** shared by the endpoints that touch refresh tokens. But `PlayersController.SelectPlayer` and `SwitchPlayer` (`Game.Api/Controllers/PlayersController.cs:34-63`) also accept and consume raw refresh tokens (`request.RefreshToken`) and carry no `[EnableRateLimiting(RateLimitingOptions.AuthPolicy)]` — only the four `AuthController` endpoints do (`AuthController.cs:41,70,81,101`).

So an authenticated caller (any self-registered account) can probe refresh-token validity or hammer the token-rotation/offline-credit path unthrottled by the auth budget.

## Severity

Hardening / invariant drift, not an open door: tokens are 256-bit CSPRNG values (`RefreshTokenStore.cs` — `RandomNumberGenerator.GetBytes(32)`), brute force is not practical, and the caller must already hold a valid access token. But the defence-in-depth story the doc describes doesn't cover two endpoints that consume the same tokens, and `SwitchPlayer` also triggers the (non-trivial) offline-credit simulation per call.

## Fix

Add `[EnableRateLimiting(RateLimitingOptions.AuthPolicy)]` to `SelectPlayer` and `SwitchPlayer` (they're pre-game/switch flows, so the shared budget won't affect normal play), or make the doc's scoping explicit if excluding them is deliberate. Add the guard test alongside the existing rate-limiting coverage.

ginderjeremiah:
Closes #2417.

> **Updated after review** (`f0cdc18`). The scope sweep stopped two sites short, and the risk framing below understated the switch cost — both corrected; details in the [reply](https://github.com/ginderjeremiah/GameServer/pull/2430#discussion_r3650465870).

## Summary

`PlayersController.SelectPlayer`/`SwitchPlayer` accept and consume a raw refresh token (`request.RefreshToken`) to rotate the token pair, but carried no `[EnableRateLimiting]` — only the four `AuthController` endpoints did. So an authenticated caller could probe refresh-token validity, or hammer `SwitchPlayer`'s offline-credit simulation, outside the combined per-IP budget `docs/backend-auth.md` describes.

I verified the issue's claims against the tree before implementing rather than taking them on faith:

- `PlayersController.cs:35-63` — both endpoints take `request.RefreshToken` and pass it into `CharacterSelectionService`, which consumes it via `AccountService.SelectPlayer`. Confirmed.
- `AuthController.cs:41,70,81,101` — the four `[EnableRateLimiting(RateLimitingOptions.AuthPolicy)]` sites. Confirmed.
- `AddAuthRateLimiter` registers `AddPolicy` only, with **no** `GlobalLimiter`, so an endpoint without the attribute is genuinely unthrottled — the gap is real, not merely undocumented.
- Severity framing confirmed too: `RefreshTokenStore` uses `RandomNumberGenerator.GetBytes(32)`, so this is defence-in-depth/invariant drift, not an open door.

## The scoping decision

The issue offered a fork — add the attribute, or make the doc's exclusion deliberate. I took the first, and settled the boundary as **"every endpoint that consumes credentials or a refresh token"** rather than "the anonymous endpoints":

- Being authenticated is not a meaningful boundary here, since any self-registered account can reach these two.
- They belong on the *same* policy, not a new one. A separate budget would hand an attacker a second bucket to spend on the same tokens — the combined-budget property is precisely what makes the guard worth having.
- Endpoints that consume neither a credential nor a token (`Players`, `CreatePlayer`, `CharacterCreationData`, `Status`, `ActiveSession`, `DeviceInfo`) stay deliberately **outside** the policy, so a normal pre-game flow spends the budget only on the calls that actually price it (login + select = 2 of the default 10/60s).

## Cost to the legitimate switch flow

Two consequences worth stating rather than burying — both verified against the frontend, and neither a reason to withhold the attribute:

- **A switch costs 1–2 permits, not one.** `CharacterSwitcher.svelte` calls `getRotatedRefreshToken()`, which awaits `ensureValidAccessToken()` first (`auth.ts`) — if a refresh is due that fires `/api/Auth/Refresh`, which is on this same policy. So the real per-IP ceiling is nearer **~5 switches/minute**, not the ">10" an earlier draft of this description claimed.
- **A throttled switch isn't recoverable in place.** The switcher calls `stopEngines()` + `apiSocket.disconnect()` *before* the request, and every non-200 falls into `reloadToBoot()`. `SwitchPlayer` could not previously return 429; now it can, and when it does the user gets a full-page reload rather than an inline retry. Graceful (the boot gate resolves back into the still-valid session), but a new user-visible outcome.

Both are now recorded in `docs/backend-auth.md` so a future limit-tuning discussion doesn't rediscover them from a bug report.

Because the partition is per-IP, callers sharing a NAT share that budget. A per-account partition would be a new policy shape, which the existing `OnRejected` comment already flags as the thing to revisit if one is ever added — so this is flagged, not pre-emptively built.

## Changes

- `PlayersController` — `[EnableRateLimiting(RateLimitingOptions.AuthPolicy)]` on `SelectPlayer` and `SwitchPlayer`.
- Docs and comments restating the scope rule, all of which enumerated the anonymous four as the policy's whole scope: `docs/backend-auth.md` (§ Auth endpoint rate limiting), `docs/backend.md` (§ Authentication), the XML docs on `RateLimitingOptions.AuthPolicy` + `.Auth` + `RateLimiterServiceCollectionExtensions`, and the `AddAuthRateLimiter()` registration comment in `Startup.cs`.
- `AuthRateLimitingTests` — guard tests, plus factoring the budget-exhaustion loop repeated across four cases into one `ExhaustBudgetViaLogin` helper (DRY, per CLAUDE.md).

Note the limiter runs ahead of `UseAuthentication` (`Startup.cs:285` vs `:287`), so an over-budget call is rejected as 429 before the 401 it would otherwise get — which is what lets the guard tests pin policy membership without minting a token, mirroring the existing `Logout_DrawsFromThePerIpBudget` shape.

## Test plan

- [x] Full backend suite green: **3506 passed / 0 failed** (Core 1453, Application 1105, Api 874, Infrastructure 74).
- [x] **Guards verified to actually guard.** Temporarily removed both attributes and re-ran: exactly the 2 new theory cases fail (429 → 401), everything else still passes. Without this the tests would have been green either way.
- [x] Negative control included — `PlayerEndpointsConsumingNoToken_StayOutsideThePerIpBudget` pins that `CreatePlayer` is *not* collaterally throttled, so the deliberate exclusion is tested rather than just asserted in prose.
- [x] `AuthRateLimitLoggingTests` re-run, since the two suites share the rate-limited factory shape — unaffected.
- [x] Post-review: rebuilt clean and re-ran both rate-limiting suites (8/8); the follow-up commit is comments + docs only.
- [x] Swept the tree for surviving "anonymous auth endpoints" scope wording. The one remaining hit (`docs/frontend-sockets.md`) is about `api-request.ts`'s refresh exemption, a different subject — deliberately left.
- No frontend files touched (no parity surface, no UI change), so the UI suite is not in scope.

No follow-up issues: nothing turned up beyond the NAT/per-account note above, which the code already anticipates in a comment.


Game.Api/RateLimiting/RateLimitingOptions.cs
```diff
@@ -16,8 +16,9 @@ public class RateLimitingOptions
         public const string AuthPolicy = "auth";
```
ginderjeremiah, line 16: **Should-fix:** the `AuthPolicy` doc one line above (`:15`) still reads *"The named policy applied to the anonymous auth endpoints"* — which now directly contradicts the `Auth` summary four lines below that this hunk rewrote. Two adjacent XML docs on the same policy disagreeing about its scope is worse than the original stale wording, because a reader has no way to tell which one is current.

Since the const is the symbol every `[EnableRateLimiting]` site references, it's arguably the more load-bearing of the two. Suggest:

```csharp
/// <summary>The named policy applied to every endpoint consuming credentials or a refresh token.</summary>
```

Same class of miss at `Game.Api/Startup.cs:112` — see the summary.

---

Game.Api/Controllers/PlayersController.cs
```diff
@@ -49,6 +54,9 @@ public async Task<ApiResponse<SelectPlayerResult>> SelectPlayer([FromBody] Selec
         /// <see cref="CharacterSelectionService.SwitchPlayer"/> for the departed-character credit and
         /// anti-cheat mechanics.
         /// </summary>
+        // Same combined per-IP budget as SelectPlayer above: it consumes a raw refresh token through the
+        // same path, and each call additionally runs the departed character's offline-credit simulation.
+        [EnableRateLimiting(RateLimitingOptions.AuthPolicy)]
```
ginderjeremiah, line 59: **Nit / accuracy of the risk framing** — the attribute is right, but the headroom is smaller than the description's ">10 switches per minute" suggests, and the failure mode is harsher than a soft retry:

1. **A switch can cost two permits, not one.** `CharacterSwitcher.svelte:69` calls `getRotatedRefreshToken()`, which awaits `ensureValidAccessToken()` first (`UI/src/lib/api/auth.ts:158-166`) — if a refresh is due, that fires `/api/Auth/Refresh`, which is on this same policy. So a switch costs 1–2 of the 10, putting the real ceiling nearer ~5 switches/minute per IP.

2. **A throttled switch isn't recoverable in place.** `CharacterSwitcher.svelte:75-84` calls `stopEngines()` and `apiSocket.disconnect()` *before* the request, and every non-200 falls into `reloadToBoot()`. Pre-PR, `SwitchPlayer` could never return 429; now it can, and when it does the user eats a full-page reload rather than an inline "try again". It does land back in the same session via the boot gate, so this is graceful — but it's a new user-visible outcome that the description's NAT-sharing note doesn't cover.

Neither is a reason to withhold the attribute — the endpoint genuinely belongs on the policy. Just worth stating accurately in the PR body (and I'd take the one-line note in `docs/backend-auth.md` that switching is on the budget, so a future limit-tuning discussion doesn't rediscover this from a bug report).

---

ginderjeremiah (commented):
## Review: bringing `SelectPlayer`/`SwitchPlayer` onto the shared `auth` rate-limit policy (#2417)

**Verdict:** Approve, with two should-fix comment corrections — the change is correct and the scoping call is the right one. *(Submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR's own author; the intended verdict is approve.)*

I re-verified the load-bearing claims against the tree rather than the description:

- `Startup.cs:285` (`UseRateLimiter`) vs `:287` (`UseAuthentication`) — the limiter really does run first, so the 429-before-401 that the new theory relies on is a property of the pipeline, not a coincidence. The guard tests are sound.
- `AddAuthRateLimiter` registers `AddPolicy` with no `GlobalLimiter`, so an unattributed endpoint was genuinely unthrottled. The gap was real.
- The doc's exclusion list (`Players`, `CreatePlayer`, `CharacterCreationData`, `Status`, `ActiveSession`, `DeviceInfo`) matches the actual endpoint set on both controllers, and none of them takes a credential or a token. The enumeration is accurate.

The scoping decision — **"consumes a credential or a refresh token"** rather than "is anonymous" — is the right boundary and worth having settled explicitly. Putting these two on the *same* policy rather than a sibling one is also correct for the stated reason: a second bucket spendable against the same tokens would undo the combined-budget property that makes the guard worth anything.

### Should-fix

- **`Game.Api/RateLimiting/RateLimitingOptions.cs:15`** — the `AuthPolicy` const doc still says *"applied to the anonymous auth endpoints"*, four lines above the `Auth` summary this PR rewrote to say the opposite. Inline comment has a suggested replacement.
- **`Game.Api/Startup.cs:112`** — *"Throttle the anonymous auth endpoints per client IP…"* at the `AddAuthRateLimiter()` registration site. Not in the diff, so no inline comment: this is where a reader first meets the limiter, and it's the last place still asserting the old scope.

Both are the exact class of comment the PR set out to correct ("Docs and comments restating the scope rule, **all** of which enumerated the anonymous four as the policy's whole scope"). The sweep was right to happen — it just stopped two sites short, and one of the survivors now contradicts a line in a file this PR edited. Small enough to fix here rather than tracking as follow-up.

### Nits

- The description's risk framing understates things slightly on two counts — a switch can spend two permits (the pre-emptive `Refresh` inside `getRotatedRefreshToken` is on the same policy), and a throttled `SwitchPlayer` lands on `reloadToBoot()` because the client tears the game down *before* issuing the request. Details in the inline comment on `PlayersController.cs:57-59`. Not a design objection — the endpoint belongs on the policy — but the ">10 switches per minute" number should read ~5, and the doc could say in one line that switching draws on the budget.

### Test coverage

Good, and notably better than the minimum this issue invited:

- The theory pins both endpoints into the policy, and the description's **verified-by-removal** step is the part that matters — without it these tests would have passed with or without the attributes, which is the usual way a rate-limit guard test ends up decorative.
- `PlayerEndpointsConsumingNoToken_StayOutsideThePerIpBudget` is the right instinct: asserting `Unauthorized` (not merely "not 429") pins that the request actually traversed the limiter and reached auth, so the deliberate exclusion is tested rather than just asserted in prose. Testing the *negative* half of a scoping decision is what keeps the boundary from silently widening later.
- `ExhaustBudgetViaLogin` removes a four-way duplicated loop — straightforward DRY per CLAUDE.md, and the class-level `<summary>` was updated in step with the widened scope rather than left behind.

No frontend test surface: no battle logic, no UI change, so the parity rule doesn't apply here.

### What's good

- The commit does exactly one thing, and the two-line comments at each attribute explain *why* the endpoint is on the policy (consumes a raw token) rather than restating *that* it is — which is what makes the next person's "can I drop this attribute?" answerable from the call site.
- Declining to build a per-account partition and instead pointing at the `OnRejected` comment that already flags it is the right call. That comment ("revisit if an account-partitioned policy is ever added") stays accurate, which is a small sign the original design anticipated this.
- Honest accounting of the NAT trade-off in the description instead of burying it.

---

Game.Api/RateLimiting/RateLimitingOptions.cs
ginderjeremiah: Fixed in `f0cdc18`, both sites — took your suggested wording verbatim for the const, and rewrote the `Startup.cs:112` registration comment to match.

Agreed on the ordering point: two adjacent XML docs disagreeing is worse than the stale wording I started with, and the const is the symbol the attribute sites actually reference. I also swept the tree for surviving instances of the phrase; the only remaining hit is `docs/frontend-sockets.md:10`, which is about `api-request.ts`'s pre-emptive-refresh exemption rather than the limiter's scope, so I left it.

---

Game.Api/Controllers/PlayersController.cs
ginderjeremiah: Checked both against the tree and they hold, so I've taken the doc note and corrected the description.

1. **Two permits — confirmed.** `getRotatedRefreshToken` awaits `ensureValidAccessToken()` before reading the token, and its own doc-comment says why (it settles a due refresh so the token read is the one actually sent, #1767/#2370). That refresh is `/api/Auth/Refresh`, on this same policy. So 1–2 permits per switch, ceiling ~5/min. I've fixed the ">10 switches per minute" line in the description.

2. **`reloadToBoot()` on 429 — confirmed,** and worth being precise about the ordering: `getRotatedRefreshToken()` runs *before* `stopEngines()`/`disconnect()`, so a throttled *refresh* fails while the game is still up and returns the "session is no longer valid" path. It's specifically the `SwitchPlayer` call itself that's post-teardown, where every non-200 — now including a 429 that was previously unreachable — falls into `reloadToBoot()`. Either way the user-visible outcome is a full-page reload, which is the new thing this PR introduces.

Both now recorded in `docs/backend-auth.md` as one sentence in the scoping paragraph, framed as input to limit tuning rather than a caveat on the attribute. That's the right place for it — the number that matters when someone tunes `PermitLimit` is switches-per-minute, not the raw permit count, and that's not derivable from the policy definition alone.

Not changing the client to retry inline: the teardown-before-request ordering is deliberate (the departed-character credit must run off that character's battle loop), so there's genuinely nothing to fall back to in place, and the boot gate already resolves the reload into the same session.

---

PR #2431, opened 2026-07-25
Reset the per-session handshake state on socket teardown (#2422)

### LINKED ISSUE #2422: disconnect() leaves socketAuthRetries spent, so a new same-tab session inherits the old session's auth-retry budget
Noticed while fixing #2419 (PR #2421), and deliberately left out of it — it changes the bounded-retry semantics established in #1992, so it deserves its own review.

## Problem

`ApiSocket.disconnect()` (`UI/src/lib/api/api-socket.ts`) tears down the connection but leaves the per-session handshake bookkeeping in place. `socketAuthRetries` is only ever reset by the `STABLE_CONNECTION_MS` timer armed in `onStart` — a socket that never stayed open long enough never refills it — and `disconnect()` explicitly clears that timer (`clearConnectionStableTimer`) without resetting the count.

Because `disconnect()`'s whole reason for existing is that the module singleton **survives** the session (the `SocketReplaced` teardown navigates client-side without a reload — `engine.ts` → `stopGame`), the next session in the same tab starts on the same `ApiSocket` instance with whatever budget the previous session had burned.

## Failure scenario

A session that flapped through, say, 4 of its 5 `MAX_SOCKET_AUTH_RETRIES` without ever reaching a stable open, then ends via `SocketReplaced` → `disconnect()` → login screen. The user logs back in on the same tab: the new session has **one** handshake-auth retry left instead of five. A single pre-open rejection on the fresh session — e.g. one genuinely stale token at the first handshake — now exhausts the budget and routes straight to `handleAuthFailure`, bouncing the user back to login. The bound (#1992) exists to stop *one* session's refresh/reconnect loop from burning rotating refresh tokens; it isn't meant to be a per-tab lifetime allowance.

Low severity: it needs a flapping session that never stabilised, and the worst outcome is a spurious re-login rather than lost data or token burn. But the invariant the bound encodes ("retries in a row *since the last stable open*") is quietly wrong across a teardown, and it's the kind of state leak that gets harder to reason about as more per-session bookkeeping lands on the singleton.

## Suggested direction

Reset the per-session handshake state in `disconnect()` — at minimum `socketAuthRetries`, and worth deciding at the same time whether `socketOpened` / `handshakeTokenExpiry` should be cleared for the same reason (both are overwritten before they're next read, so they look benign today, which is exactly why an explicit reset is cheaper than re-deriving that argument each time something new is added).

Worth confirming the framing first: the alternative reading is that the budget is deliberately per-*instance* rather than per-session, in which case the fix is a comment saying so rather than a reset. The `docs/frontend-sockets.md` wording ("5 in a row", "refills only after a reconnect stays open a stable interval") reads as per-session to me, but it doesn't address a teardown either way.

## Verification

`UI/src/tests/lib/api/api-socket.test.ts` already reads `socketAuthRetries` through its `internals()` helper and has the flapping-handshake machinery (`rejectHandshake`), so a case asserting the count is back to 0 after `disconnect()` is a few lines.

## Scope

Small — one frontend file plus a test, no backend, no parity surface.

ginderjeremiah:
Closes #2422.

> **Updated after review** (`4d4d6a18`). The "Follow-up" section below originally claimed nothing was deferred; that was an overclaim — the teardown guard is time-scoped, not structural, and the residual window is now recorded here and in #2432. Details in the [reply](https://github.com/ginderjeremiah/GameServer/pull/2431#discussion_r3650612278).

## Summary

`ApiSocket.disconnect()` tore down the connection but left `socketAuthRetries` spent. Because the singleton deliberately **survives** the session, the next session in the same tab started on the previous one's remainder — so a session that flapped through 4 of its 5 handshake-auth retries without ever reaching a stable open left the next one able to absorb a single stale token before routing to `handleAuthFailure` and bouncing the user back to login.

I verified the issue's claims against the tree rather than taking them on faith:

- `socketAuthRetries` is only ever refilled by the `STABLE_CONNECTION_MS` timer armed in `onStart`, and `disconnect()` clears that timer (`clearConnectionStableTimer`) without resetting the count. Confirmed.
- The singleton outliving the session is confirmed at both call sites (below), so the leak is reachable, not theoretical.

## Settling the framing first

The issue flagged an alternative reading — that the budget is deliberately per-*instance*, in which case the fix is a comment rather than a reset. **Per-session is right**, and the call sites make it clearer than the doc wording alone does:

- `engine.ts:397` — the `SocketReplaced` teardown, which navigates client-side to login.
- `CharacterSwitcher.svelte:76` — a **character switch**, which the issue doesn't mention and which is the more telling one: it's an ordinary, repeatable in-app action that ends one session and immediately starts another on the same instance. A per-instance budget would mean a player's Nth character switch of the session carries the auth-retry debt of the first, which nothing in #1992's rationale (stop *one* refresh/reconnect loop burning rotating refresh tokens) supports.

## The part the issue doesn't anticipate

The issue suggests resetting `socketOpened` / `handshakeTokenExpiry` alongside the counter on the grounds that both are "overwritten before they're next read, so they look benign today". They're benign to *leave*, but **not** benign to *blank* — both are inputs to `handleClose`, and clearing them makes a straggling close strictly more dangerous:

- `socketOpened: false` drops a post-teardown close out of the "opened, leave it to the keepalive" branch and into the refresh/retry path.
- `handshakeTokenExpiry: null` makes `wasHandshakeTokenDemonstrablyFresh()` return `false`, removing the other guard that would have kept it out of that path.

And that close is reachable: `disconnect()` aborts a still-CONNECTING handshake, which the browser surfaces as **1006, not `NORMAL_CLOSURE`**, so it does not take the clean-closure early return. Blanking the two fields on their own would therefore have made a teardown *burn a single-use refresh token* recovering a session that has already ended — the exact cost `MAX_SOCKET_AUTH_RETRIES` exists to bound.

So the reset is paired with two guards that make it hold:

1. **`handleClose` treats a close arriving while the teardown mark is set as terminal** — settle the unsent queue, no refresh, no reconnect. Independently correct regardless of this fix: `openSocket` already bails on the teardown mark, so the refresh accomplished nothing but the token burn.
2. **A refresh resolving after a teardown spends no budget.** This is the one path guard 1 can't cover — the close *predates* the teardown, so its `refreshTokens()` is already in flight when `disconnect()` lands (a case the suite already pins, at `api-socket.test.ts` → "does not connect when handleClose auth-retry refresh resolves after a disconnect()"). Without this the increment would land *after* the reset and quietly undo it.

## Changes

- `api-socket.ts` — `resetHandshakeState()` called from `disconnect()`, plus the two guards above. Also drops the `settleSupersededQueue` wrapper (review nit): it carried no logic over `settleQueuedRequests(CONNECTION_LOST_ERROR)` and, at 2 call sites against 9 direct ones, was the outlier spelling rather than the convention; its rationale moved inline to the two `openSocket` bails.
- `docs/frontend-sockets.md` — the bounded-retry bullet now says the budget is per session rather than per singleton; the sticky-teardown bullet records that a close arriving *while the mark is set* is terminal, and that this is time-scoped rather than structural. Both were silent on teardown, which is what let the invariant drift.

## Test plan

- [x] `npm run test:unit:ci` — **4022 passed / 0 failed** (312 files). Re-run after the review commit.
- [x] `npm run lint` — clean; `svelte-check` 1223 files, 0 errors, 0 warnings.
- [x] **Each of the three parts verified to actually guard**, by reverting it alone and re-running:
  - reverting the `disconnect()` reset → the 3 reset cases fail;
  - reverting the `handleClose` teardown short-circuit → the post-teardown-close case fails;
  - reverting the increment guard → the in-flight-refresh case fails.
  Without this the new tests could have been green either way.
- [x] The headline case is asserted **behaviourally**, not just by reading the counter: the new session is driven through 10 rejections and must consume all 5 retries (9 `refreshTokens` calls total). Inheriting the spent budget yields exactly 5 — a distinct number, so the test can't pass for the wrong reason.
- No backend files touched. `api-socket.ts` is transport, not battle logic, so there is no parity surface and no mirrored backend suite to keep in step.

## Follow-up

**#2432 — `handleClose` acts on a close from any socket, not just the current one.** The guard above is scoped to *while* the teardown mark is set, and `sendSocketCommand` lifts it. So a close that outlives the lift is judged against blanked state and loses the #2369 fresh-token protection it would incidentally have had from the retained expiry. Not reachable at either call site today (both put an async gap after `disconnect()` — a client-side navigation, an awaited HTTP round trip — far longer than the close turnaround), and the real fix is socket-scoped close handling, which is pre-existing shape rather than something this PR introduces. Deliberately deferred rather than closed here by *not* blanking `handshakeTokenExpiry`: that would trade a structural invariant for an incidental protection and leave `socketOpened` mis-scoped anyway.


UI/src/lib/api/api-socket.ts
```diff
@@ -323,11 +323,25 @@ export class ApiSocket {
...
+	 *  "rejections in a row since the last stable open" *within one session*, and the only thing that ever
+	 *  refills it — the stability timer — was just cleared, so without this the next session on this
+	 *  singleton (a character switch, or a re-login after a SocketReplaced teardown) would inherit whatever
+	 *  the previous one burned. The other two are overwritten before they are next read, and are reset with
+	 *  it so "no session state survives a teardown" stays a property of this one place rather than an
+	 *  argument to re-derive whenever new per-session bookkeeping lands on the singleton. */
+	private resetHandshakeState() {
+		this.socketAuthRetries = 0;
+		this.socketOpened = false;
+		this.handshakeTokenExpiry = null;
```
ginderjeremiah, line 344: Nit (non-blocking, follow-up filed as #2432).

The reasoning for blanking these two is sound as far as it goes, and pairing it with the `handleClose` guard is the right call — but the guard is scoped to *while the mark is set*, and `sendSocketCommand` lifts the mark. So there is a residual window the PR description reads as fully closed:

1. `disconnect()` aborts a CONNECTING handshake → close event pending, will arrive as 1006.
2. A `sendSocketCommand` lifts `teardownRequested` before that event lands.
3. The straggling close falls past the guard at `:563`, and with `handshakeTokenExpiry` now `null`, `wasHandshakeTokenDemonstrablyFresh()` returns `false` — so it loses the #2369 protection it would incidentally have had from the retained (fresh) expiry, and goes on to burn a refresh token for the dead socket.

Not reachable at either call site today — `engine.ts:397` navigates to login and `CharacterSwitcher.svelte:76` awaits an HTTP round trip, both async gaps far longer than the close turnaround — so this isn't worth holding the PR for. The structural fix is socket-scoped close handling (`handleClose` currently acts on a close from *any* socket, since it's bound to the instance), which is what #2432 covers.

---

UI/src/lib/api/api-socket.ts
```diff
@@ -546,6 +560,17 @@ export class ApiSocket {
...
 		this.clearConnectionStableTimer();
 
+		if (this.teardownRequested) {
+			// The session was explicitly torn down, so there is nothing left to recover: every branch below
+			// either reconnects or spends auth-retry budget to reconnect, and openSocket would bail on the
+			// teardown mark anyway. Short-circuiting keeps a straggling close (a CONNECTING handshake this
+			// very teardown aborted surfaces as 1006, not NORMAL_CLOSURE) from burning a single-use refresh
+			// token for a dead session — and from re-dirtying the state disconnect() just reset. Terminal
+			// like every other non-reconnecting close, so the unsent queue is settled here.
+			this.settleQueuedRequests(CONNECTION_LOST_ERROR);
```
ginderjeremiah, line 566: Nit: `settleSupersededQueue()` (`:507`) exists as a named wrapper for exactly this call, with a doc comment describing exactly this situation ("settles the queue of an open a teardown superseded"). This branch calls `settleQueuedRequests` directly instead, so the codebase now has two spellings of "a teardown settled the queue" with no distinction between them.

Either use the wrapper here, or — arguably better — drop `settleSupersededQueue` entirely, since it adds no logic over `settleQueuedRequests(CONNECTION_LOST_ERROR)` and its rationale would read just as well as an inline comment at each of the three call sites. Take it or leave it.

---

ginderjeremiah (commented):
## Review: reset the per-session handshake state on `disconnect()` (#2422)

**Verdict: Approve** — the fix is correct, the framing call it had to make first is the right one, and the two guards it pairs with the reset are each independently justified. Submitted as a `COMMENT` only because GitHub won't accept a gating verdict from the PR author; there are no blocking findings.

_(Note: GitHub reports me as the same account that opened this PR, so this is a self-review by identity. Judged on the merits regardless.)_

### Findings

No blocking items and no should-fixes. Two nits inline, neither worth holding the merge for:

- `api-socket.ts:341-342` — the teardown guard is scoped to *while the mark is set*, and `sendSocketCommand` lifts it, so there's a residual window where a straggling 1006 from an aborted handshake gets judged against blanked state and loses the #2369 fresh-token protection. **Not reachable at either call site today** (both put an async gap after `disconnect()`), and the real fix is socket-scoped close handling rather than anything in this diff — filed as **#2432**.
- `api-socket.ts:570` — uses `settleQueuedRequests` directly where `settleSupersededQueue()` is the existing named wrapper for the same teardown case. Two spellings of one concept; either is fine, pick one.

- The reset is genuinely needed: `socketAuthRetries` is refilled *only* by the `STABLE_CONNECTION_MS` timer armed in `onStart`, and `disconnect()` clears that timer at `:325`. Confirmed.
- **The per-session framing is right**, and `CharacterSwitcher.svelte:76` is the stronger argument than the one the issue makes — a character switch is an ordinary, repeatable in-app action ending one session and starting another on the same instance. A per-instance budget would make the Nth switch carry the first's auth-retry debt, which nothing in #1992's rationale supports.
- The claim that blanking `socketOpened`/`handshakeTokenExpiry` is *not* benign on its own checks out: both are `handleClose` inputs, and `disconnect()` aborting a CONNECTING handshake surfaces as 1006, so it skips the clean-closure early return. Adding the guard rather than just the reset was the correct read of the issue's suggestion.
- The increment guard covers the one path the `handleClose` short-circuit structurally can't (close predates the teardown, refresh already in flight). Correct, and the suite already had a case pinning that interleaving.

### Test coverage

Meets the frontend rule (`tests/` mirroring the source path) and then some. Five new cases covering all three parts plus the invariant, and the headline case is asserted **behaviourally** — driving the new session through 10 rejections and requiring 9 total `refreshTokens` calls, where inheriting the spent budget yields exactly 5. A distinct number, so it can't pass for the wrong reason. I re-derived that arithmetic against `handleClose` and it holds.

The "revert each part alone and watch the right cases fail" check in the test plan is the right discipline for a change like this, where three guards could easily have made each other's tests green.

No parity surface — `api-socket.ts` is transport, not battle logic — so the both-sides rule in `CLAUDE.md` doesn't apply and there's correctly no backend counterpart.

The one case not covered is the lifted-mark window above, which belongs with #2432.

### What's good

- Settling the framing question the issue raised (per-session vs per-instance) *before* writing the fix, and grounding it in a call site the issue didn't cite.
- Declining the issue's suggestion at face value: blanking `socketOpened`/`handshakeTokenExpiry` "because they're benign" would have widened a defect, and the PR found that and fixed the adjacent hole here rather than deferring it.
- `docs/frontend-sockets.md` updated in the right place — both bullets were silent on teardown, which is precisely what let the invariant drift.

---

UI/src/lib/api/api-socket.ts
ginderjeremiah: Taken — went with the second option and dropped the wrapper (`4d4d6a18`).

What decided it was the call-site count: `settleSupersededQueue` had **2** uses against **9** direct `settleQueuedRequests(CONNECTION_LOST_ERROR)` calls, so the wrapper was the outlier rather than the convention this branch failed to follow. Adopting it would have made the minority spelling spread; removing it leaves one.

Its rationale moved into the two `openSocket` bail comments, which already explained why each bails but not why it settles — so nothing is lost, and the "terminal ⇒ settle" reason now sits next to each site's own specific reason instead of one generic sentence covering all three.

---

UI/src/lib/api/api-socket.ts
ginderjeremiah: Confirmed, and the description was overclaiming — corrected.

I walked the three steps against the tree rather than accepting them: `disconnect()` blanks `handshakeTokenExpiry`, `sendSocketCommand` sets `teardownRequested = false` unconditionally as its first statement, and `handleClose` reads only instance state — so a close landing after the lift skips the guard, `wasHandshakeTokenDemonstrablyFresh()` returns `false` on the `null` expiry, and it reaches `refreshTokens()`. The window is real, and it is one my change opens: before it, the retained fresh expiry would have short-circuited that close.

Also agreed it isn't reachable today, for the reason you give — both call sites put an async gap after `disconnect()` (a client-side navigation, an awaited HTTP round trip) that dwarfs the close turnaround.

I considered closing it here by simply not blanking `handshakeTokenExpiry`, and rejected that: it would trade a structural fix for an incidental one. The protection in question is a *side effect* of stale state happening to still be readable, which is the same class of thing #2422 is about — and it would leave `socketOpened` mis-scoped anyway. #2432's socket-scoped close handling fixes the whole family at once, including the two `handleClose` reads that were already instance-scoped before this PR.

Two changes in `4d4d6a18` rather than leaving the claim standing:

- `docs/frontend-sockets.md` now says a close arriving **while the mark is set** is terminal, and notes that this is time-scoped rather than structural, with a pointer to #2432. The previous flat wording was the doc-level version of the same overclaim.
- The PR description's "Follow-up" section now records the window and #2432 instead of asserting nothing was deferred.

---

Issue #2434, opened 2026-07-25
Consolidate the hand-rolled IZone test builders onto a shared fixture


ginderjeremiah:
Split out of #2425, which called for one issue (or PR) per contract. The `ISkill` half is done in #2425's PR; #2433 covers `IItem`; this is the `IZone` half.

## Problem

~16 suites construct the `IZone` contract inline (`grep -rl "bossLevel:" UI/src/tests/`). A new required field on `IZone` means editing all of them to keep `svelte-check` green, and the builders have drifted their own defaults in the meantime.

## Suggested direction

Same shape as #2414 / the `ISkill` work:

- Add `UI/src/tests/fixtures/zones.ts` exporting `makeZone(o)`, following the `fixtures/attributes.ts` / `fixtures/proficiencies.ts` / `fixtures/skills.ts` convention.
- Reconcile the drifted defaults **explicitly** — prefer whichever variant has a live assertion.
- Convert one suite at a time, running it in between, so a changed default fails loudly rather than quietly weakening an assertion.
- Keep each suite's local call signature as a thin adapter so the call sites don't churn.

## Two things the `ISkill` pass learned that apply here

1. **The probe under-counts.** `grep "bossLevel:"` only finds builders that happen to set that field; a suite that omits it still pays the tax. Confirm the real target set by adding a throwaway required field to `IZone` and reading `svelte-check`'s error list, rather than trusting the grep. (`bossLevel` is also weaker as a probe than the others — it's a plain numeric field name, so expect false positives from unrelated objects.)
2. **The probe also over-counts.** Several suites seed a `staticData: {} as any` mock with untyped partial object literals. Those pay **no** contract-drift tax, and converting them would fill fields that were previously `undefined` — a real risk of changing what the suite renders. They should be left alone; the target is suites that construct a *typed* `IZone`.

## Verification

Adding a required field to `IZone` should afterwards produce exactly one test-file error (the new fixture).

## Scope

Medium — test-only, no production code, no behavior change, but wide.

Issue #2433, opened 2026-07-25
Consolidate the hand-rolled IItem test builders onto a shared fixture


ginderjeremiah:
Split out of #2425, which called for one issue (or PR) per contract. The `ISkill` half is done in #2425's PR; this is the `IItem` half.

## Problem

~20 suites construct the `IItem` contract inline (`grep -rl "itemCategoryId: EItemCategory" UI/src/tests/`). A new required field on `IItem` means editing all of them to keep `svelte-check` green, and the builders have drifted their own defaults in the meantime.

## Suggested direction

Same shape as #2414 / the `ISkill` work:

- Add `UI/src/tests/fixtures/items.ts` exporting `makeItem(o)`, following the `fixtures/attributes.ts` / `fixtures/proficiencies.ts` / `fixtures/skills.ts` convention.
- Reconcile the drifted defaults **explicitly** — prefer whichever variant has a live assertion.
- Convert one suite at a time, running it in between, so a changed default fails loudly rather than quietly weakening an assertion.
- Keep each suite's local call signature as a thin adapter so the call sites don't churn.

## Two things the `ISkill` pass learned that apply here

1. **The probe under-counts.** `inventory-manager.test.ts` builds a full `ISkill` but writes `acquisition: 0` as a bare numeric literal, so `grep "acquisition: ESkillAcquisition"` missed it. Confirm the real target set by adding a throwaway required field to `IItem` and reading `svelte-check`'s error list, rather than trusting the grep.
2. **The probe also over-counts.** Several suites seed a `staticData: {} as any` mock with untyped partial object literals. Those pay **no** contract-drift tax (that's the cost the issue is about), and converting them would fill fields that were previously `undefined` — a real risk of changing what the suite renders. They should be left alone; the target is suites that construct a *typed* `IItem`.

## Verification

Adding a required field to `IItem` should afterwards produce exactly one test-file error (the new fixture). That check is cheap and is the only direct evidence the consolidation actually worked.

## Scope

Medium — test-only, no production code, no behavior change, but wide.

Issue #2426, opened 2026-07-25
Point progression-test-utils' tier/path builders at the shared IProficiency/IPath fixtures


ginderjeremiah:
Found while reviewing #2424 (which closes #2414 by adding `UI/src/tests/fixtures/proficiencies.ts` → `makeProficiency` / `makePath`).

## Problem

`UI/src/tests/routes/admin/workbench/progression/types.ts` declares:

```ts
export type WorkbenchPath = IPath;
export type WorkbenchProficiency = IProficiency;
```

Both are **plain aliases**, not distinct shapes. So `progression-test-utils.ts:64` (`path`) and `:79` (`tier`) — consolidated in #2412 — hand-roll the exact same two contracts that #2424 just centralised in `tests/fixtures/proficiencies.ts`.

The result is that `IProficiency` now has *two* shared fixture homes. #2424's stated win — "one place to edit when the contract grows" — is only true for the five suites it converted; a new required field on `IProficiency` still taxes `progression-test-utils.ts` as well (plus `Synthesis.test.ts`, which #2424 should pick up directly).

This is a distinct duplication class from #2425, which covers `ISkill` / `IItem` / `IZone`.

## Suggested direction

Express the workbench builders in terms of the shared ones, keeping their divergent defaults explicit at the wrapper rather than restating the whole field set:

```ts
export const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency =>
	makeProficiency({ id: 0, name: 'Blades', iconPath: '', word: '', pronunciation: '', translation: '', xpGrowth: 1.4, ...over });
```

The blank conlang fields and `xpGrowth: 1.4` are load-bearing for the workbench detail suites (`progression-test-utils.ts:73-78` documents why), so they must stay — but they belong in the wrapper as stated overrides, not as a duplicated field list. Same for `path` (`activityKey: EActivityKey.Fire`, `id: 5`).

Worth checking as part of this whether `WorkbenchPath`/`WorkbenchProficiency` still earn their existence as aliases, or whether the workbench should just use the contract types directly.

## Constraints

Same as #2414: reconcile each divergent default explicitly rather than inheriting it silently, and run each suite as it's converted so a changed default fails loudly instead of quietly weakening an assertion.

## Scope

Small — test-only, no production code, no behaviour change.

PR #2435, opened 2026-07-25
Consolidate the hand-rolled ISkill test builders onto a shared fixture (#2425)

### LINKED ISSUE #2425: The contract-fixture duplication is much wider than the proficiency case: ISkill in ~30 suites, IItem in ~20, IZone in ~16
Found while implementing #2414 (PR #2424), which consolidated the five hand-rolled `IProficiency` builders into `UI/src/tests/fixtures/proficiencies.ts`. Surveying the call sites to size that change turned up the same duplication class on other contracts at an order of magnitude more scale, so it's split out rather than folded in.

## Problem

#2405 / #2407 / #2413 / #2414 have each closed one contract's worth of duplicated test fixtures. The remaining ones are the big ones — counted by suites that construct the contract inline:

| Contract | Suites | Probe used |
|---|---|---|
| `ISkill` | ~30 | `grep -rl "acquisition: ESkillAcquisition" UI/src/tests/` |
| `IItem` | ~20 | `grep -rl "itemCategoryId: EItemCategory" UI/src/tests/` |
| `IZone` | ~16 | `grep -rl "bossLevel:" UI/src/tests/` |

The cost is the one #2414 describes, scaled up: a new required field on `ISkill` currently means editing ~30 files to keep `svelte-check` green, and every one of those builders has drifted its own defaults in the meantime.

Two partial homes already exist and are *not* general — `UI/src/tests/lib/battle/battle-sim-test-utils.ts` (`makeSkill`, battle-shaped) and `UI/src/tests/routes/game/screens/fight/fight-fixtures.ts` (`makeSkill` returning a live `Skill` instance, not the `ISkill` contract). So this isn't "wire everyone to the existing helper" — it needs a contract-level builder in `UI/src/tests/fixtures/` that those two can then be expressed in terms of, or explicitly kept separate from.

`UI/src/tests/routes/admin/workbench/references.test.ts` is the single worst offender and a good first target: after #2424 it still carries seven local builders (`enemy`, `zone`, `challenge`, `item`, `itemMod`, `wclass`, `recipe`, `skill`) in one file.

## Suggested direction

Same shape as #2414, one contract at a time (probably `ISkill` first — it's the widest and the most churn-prone):

- Add `UI/src/tests/fixtures/<contract>.ts` exporting `make<Contract>(o)`, following the `fixtures/attributes.ts` / `fixtures/proficiencies.ts` convention.
- Reconcile the drifted defaults **explicitly** — with this many builders, expect several to have defaults a suite actually asserts on. Prefer whichever variant has a live assertion.
- Decide, per contract, whether the existing battle/fight helpers should delegate to the new builder or stay deliberately separate (the `fight-fixtures.ts` one returns a domain object, not a contract, so it likely stays).

Given the file counts this is worth splitting per contract rather than doing as one PR — plausibly three issues, or one issue with three PRs.

## Scope

ginderjeremiah:
Closes #2425 — see **Scope** below for why this closes the umbrella rather than leaving it open.

> **Updated after review** (`deff6ab`). Three changes: the exclusion rationale now lives in the fixture's module comment rather than only here; `docs/frontend.md` documents the `fixtures/` convention; and `EquippedBand`'s adapter — which had degenerated to a pure passthrough — is now an import alias. The **Converged defaults** section below is also corrected: there are **two**, not one, and the excluded-suite count is **nine**, not eight.

## Summary

Adds `UI/src/tests/fixtures/skills.ts` (`makeSkill`), following the `fixtures/attributes.ts` / `fixtures/proficiencies.ts` convention, and converts every suite that hand-rolled the `ISkill` contract to build on it. Each suite keeps its local call signature as a thin adapter, stating its divergent defaults as explicit overrides rather than restating the whole 15-field set.

Net: **25 files, +157 / −516** — 23 converted suites, the new fixture, and the `docs/frontend.md` entry.

## The issue's probe was wrong in both directions

The issue sized this at "~30 suites" from `grep -rl "acquisition: ESkillAcquisition" UI/src/tests/`. I checked that against the tree rather than taking it on faith, and the real target set is different:

**It over-counts.** Nine suites seed a `staticData: {} as any` mock with *untyped partial* object literals — `Codex.test.ts`, `codex-view.svelte.test.ts`, `reference.test.ts`, `entities/{class,enemy,item,skill-recipe}.test.ts`, `progression/{TierDetail,progression-helpers}.test.ts`. Those pay **no** contract-drift tax, which is the entire cost the issue is about: a new required `ISkill` field doesn't touch them, because nothing type-checks them as `ISkill`. Converting them would also have *filled in fields that were previously `undefined`*, which is a real risk of changing what the suite renders. They're deliberately left alone, and that rationale now lives in the fixture's module comment so it survives this issue closing.

**It under-counts.** `inventory-manager.test.ts` builds a full, typed `ISkill` but writes `acquisition: 0` — a bare numeric literal, so the grep never saw it. It pays the tax like the rest, and is converted here.

(`entities/skill-recipe.test.ts` evades the grep for the same reason — it writes `acquisition: opts.synth ? … : …`. It lands in the untyped bucket: its `retiredAt: null` isn't even assignable to `string | undefined`.)

So the real set is **23 files**, not 30 — a different 23 than the grep suggested.

## Verification that the consolidation actually worked

The stated win is "one place to edit when the contract grows", so I measured it rather than asserting it. Adding a throwaway required field to `ISkill` and running `svelte-check`:

| | Test-file errors |
|---|---|
| Before | 23 (every converted suite) |
| After | **1** — `src/tests/fixtures/skills.ts` |

(The other 4 errors in the "after" run are production files — `lib/battle/skill.ts`, `admin/workbench/entities/skill.ts`, `fight/SkillTooltip.svelte` — which is correct and expected.) The probe was reverted; it isn't in the diff.

## Defaults

The shared defaults are the majority spelling across the 23 builders. Divergences are stated at each wrapper, so a suite's intent stays local and visible:

- `InnateBand` / `ItemTooltip` → `acquisition: Item`; the four synthesis suites → `acquisition: Synthesis`
- `SkillCardStats` → `cooldownMs: 2000`; `skills-view` / `attributes-view` → `baseDamage: 0`; `SkillDamageBreakdown` → `baseDamage: 50`
- `Skills.test.ts` → its Strength multiplier and `'A skill.'` description

### Converged defaults

Two values were **converged** onto the shared default rather than preserved. Both were checked against the suite's consumers first:

1. **`references.test.ts`** had `damagePortions: []` where every other builder has a single full-weight physical portion. That suite exercises name/id reference resolution and never reads portions; its 45 tests still pass.
2. **`inventory-manager.test.ts`** had `acquisition: 0`. `ESkillAcquisition` is a flags enum (`None = 0`, `Player = 1`), so this silently moves the Punch record from `None` to `Player`. Benign — no consumer in that suite reads `acquisition` (the readers are `skill-provenance.ts`, the codex views, and the workbench admin surfaces) — and arguably more faithful, since the real Punch record in `content/skills.json` is authored as `Player`. Flagged here because a bare numeric literal for a flags enum is exactly the shape the under-count lesson is about, and #2433/#2434 should watch for it.

**The conlang fields deliberately differ from `makeProficiency`.** That fixture generates `w1`/`p1`/`t1` from the id; this one defaults them to blank, because a blank `word` is what marks a skill *un-deciphered*. Generating one would silently make every fixture skill deciphered and quietly weaken the synthesis reveal suites. That reasoning is in the fixture's doc comment so the asymmetry doesn't read as an oversight later.

## Two naming collisions, handled by aliasing

`battle-sim-test-utils.ts` and `fight-fixtures.ts` both already export their own `makeSkill`, and neither is the contract — the first builds a pre-registration `SkillSpec`, the second a live `Skill` domain object bound to an owner. Both now consume the fixture as `makeSkillContract` with a one-line comment saying why, rather than being renamed or absorbed. This is the "decide, per contract, whether the existing helpers should delegate or stay separate" call the issue asks for: they delegate, they keep their names.

## Incidental cleanup

`battler-effects.test.ts` had the same 20-line "Dual" skill literal twice; it's now one `dualEffectSkill(caster)` helper. Newly-unused enum imports (`ERarity`, `EDamageType`, `ESkillAcquisition`) were pruned per file — each was checked for remaining uses rather than removed blind.

## Docs

`docs/frontend.md` gained a line under **Testing Guidelines** describing the `UI/src/tests/fixtures/` convention. It was undocumented despite this being its third module, with two more queued — inherited debt from #2414, closed here since it's cheaper than a follow-up issue.

## Scope, and why this closes #2425

The issue proposed "plausibly three issues, or one issue with three PRs". I've taken the first reading, since an umbrella that can never be closed by any of its own PRs is worse tracking than three closeable issues:

- `ISkill` (~30 → 23 files) — **this PR**
- `IItem` (~20) — **#2433**
- `IZone` (~16) — **#2434**

Both follow-ups carry forward the two probe lessons above, so the next session doesn't rediscover the over/under-count by hand.

Not touched here, and correctly out of scope: **#2426** (pointing `progression-test-utils`' tier/path builders at the shared `IProficiency`/`IPath` fixtures) is a different contract and already tracked.

## Test plan

- [x] `npm run test:unit:ci` — **4017 passed / 0 failed** (1153 suites). Re-run after the review commit.
- [x] `npm run lint` — clean; `svelte-check` 1224 files, **0 errors, 0 warnings**.
- [x] `npx prettier --check` — clean, including `docs/frontend.md`.
- [x] CI green on `deff6ab` — all 11 checks (frontend unit, verify, e2e on chromium/firefox/webkit); backend jobs correctly skipped by the changed-path gate.
- [x] Each suite run individually as it was converted, so a changed default would fail loudly rather than quietly weakening an assertion.
- [x] The anti-drift claim measured directly (the 23 → 1 table above), not assumed.
- [x] Both converged defaults traced to their consumers before accepting the convergence, rather than assumed benign.
- **No backend mirror needed.** This touches the battle suites (`battle-formulas`, `battler`, `skill`, `battler-effects`, `battle-sim-test-utils`, `battle-engine`), which are on the frontend/backend parity surface — but no scenario input or expected result changed. Every value those suites feed the simulator is either passed explicitly or identical to the shared default; only the construction of the fixture object moved. Verified per-file rather than assumed, since CLAUDE.md requires parity scenarios to stay in step.

### DIFF
--- a/UI/src/tests/fixtures/skills.ts
+++ b/UI/src/tests/fixtures/skills.ts
@@ -6,7 +6,11 @@ import { EDamageType, ERarity, ESkillAcquisition, type ISkill } from '$lib/api';
    · `lib/battle/battle-sim-test-utils.ts` (`registerSkill`) assigns ids by registry position so a
      Battler resolves a skill exactly as the live game does — it builds on this module but owns the id.
    · `routes/game/screens/fight/fight-fixtures.ts` (`makeSkill`) returns a live `Skill` domain object
-     bound to an owner, not the `ISkill` contract, so it consumes this module rather than replacing it. */
+     bound to an owner, not the `ISkill` contract, so it consumes this module rather than replacing it.
+   · Nine suites seed untyped skill literals into a `staticData: {} as any` mock (codex, workbench
+     entities/progression). They type-check as nothing, so they pay no contract-drift tax — and filling
+     their previously-`undefined` fields could change what the suite renders. Left alone deliberately;
+     a `grep` for skill-shaped literals under `src/tests/` still finding them is expected, not a gap. */
 
 /**
  * Builds an {@link ISkill} reference-data entry for tests. Everything but the id is a neutral
--- a/UI/src/tests/fixtures/skills.ts
+++ b/UI/src/tests/fixtures/skills.ts
@@ -7,10 +7,13 @@ import { EDamageType, ERarity, ESkillAcquisition, type ISkill } from '$lib/api';
      Battler resolves a skill exactly as the live game does — it builds on this module but owns the id.
    · `routes/game/screens/fight/fight-fixtures.ts` (`makeSkill`) returns a live `Skill` domain object
      bound to an owner, not the `ISkill` contract, so it consumes this module rather than replacing it.
-   · Nine suites seed untyped skill literals into a `staticData: {} as any` mock (codex, workbench
-     entities/progression). They type-check as nothing, so they pay no contract-drift tax — and filling
-     their previously-`undefined` fields could change what the suite renders. Left alone deliberately;
-     a `grep` for skill-shaped literals under `src/tests/` still finding them is expected, not a gap. */
+
+   Nine suites are excluded on purpose: the codex pair, `workbench/reference`, and the workbench
+   entities/progression suites all seed skill literals into a mocked `staticData` — most as `{} as any`,
+   the progression pair via `progression-test-utils`' `unknown[]`. Nothing type-checks them as `ISkill`,
+   so they pay no contract-drift tax, and filling their previously-`undefined` fields could change what
+   a suite renders. A `grep` for skill-shaped literals under `src/tests/` still finding them is
+   expected, not a gap. */
 
 /**
  * Builds an {@link ISkill} reference-data entry for tests. Everything but the id is a neutral
--- a/UI/src/tests/fixtures/skills.ts
+++ b/UI/src/tests/fixtures/skills.ts
@@ -1,0 +1,37 @@
+import { EDamageType, ERarity, ESkillAcquisition, type ISkill } from '$lib/api';
+
+/* Shared `ISkill` contract builder (#2425). Two other modules build adjacent shapes and deliberately
+   stay separate:
+
+   · `lib/battle/battle-sim-test-utils.ts` (`registerSkill`) assigns ids by registry position so a
+     Battler resolves a skill exactly as the live game does — it builds on this module but owns the id.
+   · `routes/game/screens/fight/fight-fixtures.ts` (`makeSkill`) returns a live `Skill` domain object
+     bound to an owner, not the `ISkill` contract, so it consumes this module rather than replacing it. */
+
+/**
+ * Builds an {@link ISkill} reference-data entry for tests. Everything but the id is a neutral
+ * placeholder so a suite states only what it asserts on.
+ *
+ * The conlang fields (`word`/`pronunciation`/`translation`) default to **blank**, unlike
+ * {@link import('./proficiencies').makeProficiency}, which generates them from the id. A blank word is
+ * what marks a skill as un-deciphered, so the synthesis reveal suites depend on the empty default —
+ * generating one here would silently make every fixture skill deciphered.
+ */
+export const makeSkill = (overrides: Partial<ISkill> & { id: number }): ISkill => ({
+	name: `Skill ${overrides.id}`,
+	baseDamage: 10,
+	criticalChance: 0,
+	damageMultipliers: [],
+	effects: [],
+	damagePortions: [{ type: EDamageType.Physical, weight: 1 }],
+	description: '',
+	cooldownMs: 1000,
+	iconPath: '',
+	rarityId: ERarity.Common,
+	word: '',
+	pronunciation: '',
+	translation: '',
+	acquisition: ESkillAcquisition.Player,
+	designerNotes: '',
+	...overrides
+});
--- a/UI/src/tests/lib/engine/player/inventory-manager.test.ts
+++ b/UI/src/tests/lib/engine/player/inventory-manager.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { EAttribute, EDamageType, EItemCategory, EItemModType, ELogType, ERarity } from '$lib/api';
 import type { IInventoryData, IItem, IItemMod, ISkill } from '$lib/api';
 import { PUNCH_SKILL_ID } from '$lib/api/types/game-constants';
+import { makeSkill } from '../../../fixtures/skills';
 
 const mockInventoryData: IInventoryData = {
 	unlockedItems: [],
@@ -332,24 +333,12 @@ describe('InventoryManager', () => {
 	// appended to the granted ids only when it resolves. A real weapon (any type) replaces the fists.
 	describe('virtual-fists punch + equippedWeaponType (#1342)', () => {
 		const seedPunchSkill = () => {
-			mockSkills[PUNCH_SKILL_ID] = {
+			mockSkills[PUNCH_SKILL_ID] = makeSkill({
 				id: PUNCH_SKILL_ID,
 				name: 'Punch',
 				baseDamage: 4,
-				criticalChance: 0,
-				damageMultipliers: [],
-				effects: [],
-				description: '',
-				cooldownMs: 1000,
-				damagePortions: [{ type: EDamageType.Unarmed, weight: 1 }],
-				iconPath: '',
-				rarityId: ERarity.Common,
-				designerNotes: '',
-				word: '',
-				pronunciation: '',
-				translation: '',
-				acquisition: 0
-			};
+				damagePortions: [{ type: EDamageType.Unarmed, weight: 1 }]
+			});
 		};
 
 		it('appends the punch signature when bare-handed and the punch skill resolves', () => {
--- a/UI/src/tests/routes/game/screens/skills/EquippedBand.test.ts
+++ b/UI/src/tests/routes/game/screens/skills/EquippedBand.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, cleanup, fireEvent } from '@testing-library/svelte';
-import { EDamageType, EModifierType, EAttribute, ERarity, ESkillAcquisition } from '$lib/api';
+import { EModifierType, EAttribute } from '$lib/api';
+import { makeSkill } from '../../../../fixtures/skills';
 import type { IChallenge, IEnemy, ISignaturePassive, ISkill, IZone } from '$lib/api';
 import type { AttributeModifier } from '$lib/battle';
 import { classSignaturePassiveModifier } from '$lib/battle/class-modifiers';
@@ -67,24 +68,7 @@ vi.mock('$lib/api/types/game-constants', async (importOriginal) => {
 import { SkillsView } from '$routes/game/screens/skills/skills-view.svelte';
 import EquippedBand from '$routes/game/screens/skills/EquippedBand.svelte';
 
-const skill = (over: Partial<ISkill> & { id: number }): ISkill => ({
-	name: `Skill ${over.id}`,
-	baseDamage: 10,
-	criticalChance: 0,
-	damageMultipliers: [],
-	effects: [],
-	description: '',
-	designerNotes: '',
-	cooldownMs: 1000,
-	iconPath: '',
-	rarityId: ERarity.Common,
-	word: '',
-	pronunciation: '',
-	translation: '',
-	damagePortions: [{ type: EDamageType.Physical, weight: 1 }],
-	acquisition: ESkillAcquisition.Player,
-	...over
-});
+const skill = (over: Partial<ISkill> & { id: number }): ISkill => makeSkill(over);
 
 const SKILLS: ISkill[] = [
 	skill({ id: 0, name: 'Alpha' }),
--- a/UI/src/tests/routes/game/screens/skills/EquippedBand.test.ts
+++ b/UI/src/tests/routes/game/screens/skills/EquippedBand.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, cleanup, fireEvent } from '@testing-library/svelte';
 import { EModifierType, EAttribute } from '$lib/api';
-import { makeSkill } from '../../../../fixtures/skills';
+import { makeSkill as skill } from '../../../../fixtures/skills';
 import type { IChallenge, IEnemy, ISignaturePassive, ISkill, IZone } from '$lib/api';
 import type { AttributeModifier } from '$lib/battle';
 import { classSignaturePassiveModifier } from '$lib/battle/class-modifiers';
@@ -67,8 +67,6 @@ vi.mock('$lib/api/types/game-constants', async (importOriginal) => {
 
 import { SkillsView } from '$routes/game/screens/skills/skills-view.svelte';
 import EquippedBand from '$routes/game/screens/skills/EquippedBand.svelte';
-
-const skill = (over: Partial<ISkill> & { id: number }): ISkill => makeSkill(over);
 
 const SKILLS: ISkill[] = [
 	skill({ id: 0, name: 'Alpha' }),
--- a/UI/src/tests/lib/battle/battle-formulas.test.ts
+++ b/UI/src/tests/lib/battle/battle-formulas.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { ERarity, EAttribute, EDamageType, ESkillAcquisition } from '$lib/api';
+import { EAttribute, EDamageType } from '$lib/api';
 import type { ISkill } from '$lib/api';
 import { BattleAttributes } from '$lib/battle/battle-attributes';
 import {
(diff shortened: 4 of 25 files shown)


UI/src/tests/fixtures/skills.ts
```diff
@@ -0,0 +1,37 @@
+import { EDamageType, ERarity, ESkillAcquisition, type ISkill } from '$lib/api';
+
+/* Shared `ISkill` contract builder (#2425). Two other modules build adjacent shapes and deliberately
+   stay separate:
+
+   · `lib/battle/battle-sim-test-utils.ts` (`registerSkill`) assigns ids by registry position so a
+     Battler resolves a skill exactly as the live game does — it builds on this module but owns the id.
+   · `routes/game/screens/fight/fight-fixtures.ts` (`makeSkill`) returns a live `Skill` domain object
+     bound to an owner, not the `ISkill` contract, so it consumes this module rather than replacing it. */
```
ginderjeremiah: **Should-fix** — this comment records the two modules that deliberately *alias* the fixture, but not the set that deliberately **doesn't use it at all**.

Nine suites still carry untyped skill literals against `staticData: {} as any` and are intentionally left alone (`Codex.test.ts`, `codex-view.svelte.test.ts`, `reference.test.ts`, `entities/{class,enemy,item,skill-recipe}.test.ts`, `progression/{TierDetail,progression-helpers}.test.ts`). I verified the exclusion criterion holds for all of them — none type-checks as `ISkill`, so none pays the drift tax, and filling in previously-`undefined` fields would genuinely risk changing what those suites render. Leaving them is the right call.

The problem is *where* that reasoning lives. It's currently only in the PR description, and this PR closes #2425 — so once merged, nothing durable explains why a `grep "acquisition:"` over `src/tests/` still returns nine hits after a PR whose whole point was removing them. The next session re-derives it from scratch, which is exactly the cost #2433/#2434 were written to avoid.

A third bullet here is enough:

```
   · Nine suites seed untyped skill literals into a `staticData: {} as any` mock (codex, workbench
     entities/progression). They type-check as nothing, so they pay no contract-drift tax — and
     filling their previously-`undefined` fields could change what the suite renders. Left alone.
```

(Note the count is nine, not the eight in the PR body — `entities/skill-recipe.test.ts` builds one too, and its `retiredAt: null` isn't even assignable to `string | undefined`, so it's firmly in the untyped bucket.)

---

UI/src/tests/routes/game/screens/skills/EquippedBand.test.ts
```diff
@@ -67,24 +68,7 @@ vi.mock('$lib/api/types/game-constants', async (importOriginal) => {
...
-	iconPath: '',
-	rarityId: ERarity.Common,
-	word: '',
-	pronunciation: '',
-	translation: '',
-	damagePortions: [{ type: EDamageType.Physical, weight: 1 }],
-	acquisition: ESkillAcquisition.Player,
-	...over
-});
+const skill = (over: Partial<ISkill> & { id: number }): ISkill => makeSkill(over);
```
ginderjeremiah: **Nit** — this is the one adapter that degenerated into a pure passthrough: same signature, same defaults, zero overrides. The "keep a thin adapter so call sites don't churn" rule from #2425 is about *preserving a divergent call shape*, and there's nothing left to preserve here.

Every other adapter in the PR still earns its keep — `references.test.ts` and `battler.test.ts` convert a positional signature, `SkillCardStats` pins `cooldownMs: 2000`, `InnateBand` pins `acquisition: Item`. This one just forwards, which is the dead indirection CLAUDE.md's no-dead-code rule points at.

The call sites stay identical if you drop the wrapper and alias at the import instead:

```ts
import { makeSkill as skill } from '../../../../fixtures/skills';
```

---

UI/src/tests/lib/engine/player/inventory-manager.test.ts
```diff
@@ -332,24 +333,12 @@ describe('InventoryManager', () => {
 	// appended to the granted ids only when it resolves. A real weapon (any type) replaces the fists.
 	describe('virtual-fists punch + equippedWeaponType (#1342)', () => {
 		const seedPunchSkill = () => {
-			mockSkills[PUNCH_SKILL_ID] = {
+			mockSkills[PUNCH_SKILL_ID] = makeSkill({
```
ginderjeremiah, line 336: **Nit** — this is a second converged default, and unlike the `references.test.ts` `damagePortions` case the PR body doesn't call it out.

`ESkillAcquisition` is a flags enum (`None = 0, Player = 1, Item = 2, Enemy = 4, Synthesis = 8`), so the old `acquisition: 0` was `None`, not `Player`. Inheriting the fixture default silently moves this record from `None` → `Player`.

It's benign and arguably a fix: `InventoryManager` never reads `acquisition` (the only consumers are `skill-provenance.ts` and the codex views), and the real Punch record in `content/skills.json` is authored as `Player` — so the fixture is now *more* faithful to what it stands in for. But the PR set a good bar by documenting the `damagePortions` convergence explicitly, and this one deserves the same line in the description. Worth being precise about in #2433/#2434 too, since bare numeric literals for flags enums are exactly what the under-count lesson is about.

---

ginderjeremiah (commented):
## Review: consolidate 23 hand-rolled `ISkill` test builders onto `fixtures/skills.ts` (#2425)

**Verdict:** Approve — the refactor is correct and the anti-drift claim is real; two should-fixes are documentation-shaped, not code-shaped. (Submitted as a comment since I'm the PR author and GitHub won't take a gating verdict on my own PR.)

I re-derived the risky parts rather than trusting the description, since the whole failure mode of this refactor class is a silently-changed default weakening an assertion.

### Verified independently

- **Every converted default is value-identical** to what it replaced, across all 23 files — checked field by field against the pre-image. The only two divergences are the `references.test.ts` `damagePortions` convergence the PR already documents, and the `inventory-manager.test.ts` `acquisition` one flagged inline.
- **`references.test.ts`'s `damagePortions: []` → full-weight physical is safe.** `skill()` there is only ever fed to name/id reference resolution (lines 263/272/291/335); nothing in the suite reads portions.
- **Import pruning is clean.** Scanned every changed file for `ERarity` / `EDamageType` / `ESkillAcquisition` / `EAttribute` / `EModifierType` appearing exactly once (import with no use) — zero hits. No dead imports left behind.
- **No shared-mutable-default footgun.** `damagePortions` / `damageMultipliers` / `effects` are array literals inside the returned object, so each call gets fresh instances. This is the classic way a shared fixture leaks state between tests, and it's avoided.
- **The battle-parity claim holds.** Of the six suites on the frontend/backend parity surface (`battle-formulas`, `battler`, `skill`, `battler-effects`, `battle-sim-test-utils`, `battle-engine`), not one scenario input or expected result changed — only object construction moved. CLAUDE.md's mirrored-scenario rule is satisfied without a backend change.
- **`entities/skill.test.ts:86` is correctly left as an explicit literal** — it's the assertion for `skillEntity.newItem`, so it must restate the contract rather than inherit it. Easy thing to convert by reflex; good that it wasn't.
- **Follow-ups exist and are well-formed.** #2433 / #2434 both carry the over-count and under-count lessons forward and are labeled `tech debt` / `claude` / `scope: medium`.
- **CI is green** across all 11 checks (frontend unit, verify, e2e on all three browsers).

### Should-fix

- **The exclusion rationale isn't durable** — `UI/src/tests/fixtures/skills.ts:3-9`. Nine suites keep untyped skill literals on purpose, and the reasoning lives only in a PR body that disappears from view once #2425 closes. One more bullet in the module comment fixes it. Details inline.
- **The `fixtures/` convention itself is still undocumented.** `docs/frontend.md` has no mention of `UI/src/tests/fixtures/` — I grepped. This PR is the third module in that directory and #2433/#2434 will add two more, and the PR body leans on "the `fixtures/attributes.ts` / `fixtures/proficiencies.ts` convention" as though it were written down somewhere. CLAUDE.md explicitly lists "frequently used patterns in the code" as a thing the docs should carry. This is inherited debt from #2414 rather than something this PR introduced, so a follow-up issue is a fair resolution — but a one-line entry under the frontend testing section would close it here and is cheaper than the issue.

### Nits

- `EquippedBand.test.ts:71` — the adapter degenerated to a pure passthrough; alias at the import instead.
- `inventory-manager.test.ts:336` — a second converged default (`None` → `Player` on a flags enum) that the description doesn't mention. Benign, and the new value actually matches authored content.
- `skill-provenance.test.ts` now mixes a typed fixture skill with an `as IItem` partial cast for items in the same file. Not worth touching — #2433 resolves it.

### Test coverage

Appropriate as-is. This is a test-only refactor with no production change, so the bar is "no assertion silently weakened", and the PR meets it: converting one suite at a time with a run in between is the right method, and the 23 → 1 `svelte-check` error table is real evidence the consolidation achieved its stated goal rather than an assertion that it did. Measuring the win instead of claiming it is the part of this PR I'd most like to see repeated in #2433/#2434.

### What's good

- The fixture mirrors `makeProficiency` exactly — same `Partial<T> & { id: number }` signature, same doc-comment shape, same module-level note about adjacent builders. It reads like it was always there.
- The blank-conlang asymmetry is documented **with its reason** ("a blank word is what marks a skill un-deciphered"). That's precisely the deliberate-looking construct a later edit would otherwise "fix" into consistency with `makeProficiency` and quietly break the synthesis reveal suites.
- Aliasing the two colliding `makeSkill` exports to `makeSkillContract` — rather than renaming them or absorbing them — is the right resolution, and both sites say why in one line.
- The description's correction of the issue's own probe, in both directions, is more valuable than the refactor. Pushing that into #2433/#2434 as explicit guidance ("confirm the target set with a throwaway required field, don't trust the grep") is what keeps the next two PRs from repeating the miscount.

---

ginderjeremiah:
Both should-fixes and one nit addressed in `deff6ab`; description updated. I verified each claim against the tree rather than taking it, since two of them assert things about values.

**Should-fix 1 — durable exclusion rationale.** Done, and your count is right: **nine**, not eight. `entities/skill-recipe.test.ts` evades the original grep for the same reason `inventory-manager.test.ts` did — it writes `acquisition: opts.synth ? … : …` rather than a bare `ESkillAcquisition` member. Confirmed it's firmly untyped (`staticData: {} as any` at line 20, and `retiredAt: null` is not assignable to `string | undefined`). The third bullet is in the module comment, with the point you make about `grep` still returning hits being expected rather than a gap.

**Should-fix 2 — the `fixtures/` convention.** Closed here rather than deferred. Confirmed the gap by grepping `docs/` (not just `frontend.md`) — no mention anywhere. Added a line under **Testing Guidelines** covering the one-module-per-contract shape, the `make<Contract>(overrides)` naming, and the wrap-and-state-the-divergence rule, since that last part is the bit #2433/#2434 most need to inherit.

**Nit — `EquippedBand` passthrough.** Agreed, and taken. The wrapper is gone in favour of `import { makeSkill as skill }`; call sites are byte-identical and the suite's 18 tests still pass. You're right that the "keep a thin adapter" rule is about preserving a divergent call shape — with zero overrides there was nothing left to preserve, just indirection.

**Nit — `inventory-manager.test.ts` `None` → `Player`.** Correct on all three points, and it's now documented alongside the `damagePortions` convergence under a **Converged defaults** heading. I re-derived the benign claim rather than inheriting it: `ESkillAcquisition.None = 0` / `Player = 1` confirmed at `enums.ts:255-261`, and the only production readers of `.acquisition` are `skill-provenance.ts`, the codex views, and the workbench admin surfaces — `InventoryManager` isn't among them. Also flagged in the PR body that bare numeric literals for flags enums are the shape #2433/#2434 should watch for, since that's the generalisable part.

**Nit — `skill-provenance.ts` mixed cast.** Agreed, left for #2433.

Re-verified after the changes: 4017 tests / 0 failed, `svelte-check` 1224 files 0 errors 0 warnings, prettier clean including the doc.

One note on the approval mechanics — since this is your own PR, GitHub won't take a gating verdict on it, so the comment-state approval is as far as that can go. I'll hold off on merging unless you say otherwise; the call on self-approved PRs is yours, not something I should infer from a review I'd be acting on as both author and reviewer.

---

ginderjeremiah:
## Re-review of `deff6ab` — both should-fixes landed; one correction to the new comment, pushed as `088fd0e`

**Verdict:** Approve. Merging once the run on `088fd0e` goes green (it was green on `deff6ab` across all 11 checks; `verify-frontend` has already passed on the new head).

Incremental pass over the delta only — three files, +8/−4.

### The should-fixes

**1. Durable exclusion rationale.** Landed in the fixture's module comment, and the count is right. Verified it independently rather than from the description: subtracting the suites that import `fixtures/skills` from those still carrying a skill-shaped literal under `src/tests/` leaves exactly nine — the codex pair, `workbench/reference`, the four `workbench/entities` suites, and the two `workbench/progression` suites. Matches the list in the PR body, including `entities/skill-recipe.test.ts`, which the original grep missed for the same reason `inventory-manager` did.

The framing is the part worth keeping. *"A `grep` for skill-shaped literals under `src/tests/` still finding them is expected, not a gap"* pre-empts the exact wrong conclusion the next reader would draw, which is what makes this durable rather than just recorded.

**2. `docs/frontend.md`.** Good call closing it here instead of deferring. The entry carries the three things #2433/#2434 need to inherit — one module per contract, `make<Contract>(overrides)`, and wrap-and-state-the-divergence — rather than just noting the directory exists. Placed under Testing Guidelines next to the mirrored-structure rule, which is where someone looks.

**3. `EquippedBand` alias.** Confirmed the wrapper had nothing left to preserve and the import alias is equivalent: `makeSkill as skill`, four call sites unchanged, no other reference to the old name in the file.

### What I fixed (`088fd0e`)

Two problems, both in the four lines the delta added, both comment-only:

- **The list header no longer matches the list.** It reads *"**Two other modules** build adjacent shapes and deliberately stay separate:"* followed by **three** bullets — and the third isn't a module, it's a set of excluded suites. Split into its own paragraph so the header describes what it introduces.
- **The stated mock shape is wrong for two of the nine.** The comment says all nine seed into a `staticData: {} as any` mock. Seven do; `progression/TierDetail.test.ts` and `progression/progression-helpers.test.ts` go through `progression-test-utils.ts`, where `staticData` is `Record<(typeof catalogueKeys)[number], unknown[] | undefined>`. The conclusion is unaffected — `unknown[]` type-checks as `ISkill` no more than `any` does — but the point of this comment is to be the exclusion *criterion*, and someone applying it as written finds seven of the nine and assumes the other two were an oversight. That's the failure mode the comment exists to prevent, so it's worth being exact about.

Pushed rather than sent back: it's a comment edit, `verify-frontend` covers it, and a round-trip would cost more than the fix.

### Everything else

Standing from the first review, spot-checked rather than re-derived: the fixture's defaults are unchanged by this delta, the conlang doc comment still explains the blank-`word` asymmetry with its reason, and the two `makeSkillContract` aliases still carry their one-line justification. I re-read the `battler-effects` and `SkillCardStats` conversions as a sample and both are value-identical to their pre-images, consistent with the field-by-field check in the first review.

`main` moved to `a47207c` while this sat; the PR's base is already that commit, so there's nothing to update.

### Follow-ups

Both already exist and neither is work this PR should have done: **#2433** (`IItem`, ~20 suites) and **#2434** (`IZone`, ~16), each `tech debt` / `claude` / `scope: medium`, and each carrying the over-count and under-count lessons plus the throwaway-required-field verification. Closing #2425 on this PR is the right call given those two exist — an umbrella no PR of its own can close is worse tracking.

Nothing else outstanding, so nothing further filed from this pass.

---

Issue #2436, opened 2026-07-25
WorkbenchPath/WorkbenchProficiency are pure aliases, unlike the other five Workbench&lt;Entity&gt; types


ginderjeremiah:
Split out of #2426, which asked as a side question *"whether `WorkbenchPath`/`WorkbenchProficiency` still earn their existence as aliases, or whether the workbench should just use the contract types directly."* The fixture consolidation that issue is actually about is done in its PR; this is the leftover question, split out because answering it is a ~17-file rename across production components rather than a test-only change.

## What the survey found

`src/routes/admin/workbench/progression/types.ts`:

```ts
export type WorkbenchPath = IPath;
export type WorkbenchProficiency = IProficiency;
```

The name suggests membership in an established family, and there are five siblings — but **every one of them diverges from its contract, and these two are the only ones that don't**:

| Type | Divergence from the contract | Location |
|---|---|---|
| `WorkbenchZone` | `extends IZone` + `zoneEnemies: IZoneEnemy[]` (derived spawn table) | `entities/zone.ts` |
| `WorkbenchEnemy` | `extends IEnemy` + `bossZones: number[]` (inverse of the zones' boss FK) | `entities/enemy.ts` |
| `WorkbenchItem` | `Omit<IItem, 'weaponType'>` + `weaponType: number` (widened for the `-1` "None" sentinel) | `entities/item.ts` |
| `WorkbenchClass` | `Omit<IClass, 'passiveScalingAttributeId'>` + widened for the same sentinel | `entities/class.ts` |
| `WorkbenchLesson` | `Omit<ILesson, 'triggerMechanicEvent'>` + widened for the same sentinel | `entities/lesson.ts` |
| **`WorkbenchPath`** | **none — plain alias** | `progression/types.ts` |
| **`WorkbenchProficiency`** | **none — plain alias** | `progression/types.ts` |

Two further differences weaken the "consistency with the family" defence:

1. **Different home.** The five live in `entities/*.ts`; these two live in `progression/types.ts`.
2. **Different role.** The five exist to be the `EntityConfig<T>` type parameter for the generic Workbench. The progression editor is a bespoke editor and isn't `EntityConfig`-driven at all, so these two carry no such role.

So the family membership is nominal. The cost is small but real: the types' own doc comments say *"Identical to the {@link IPath} contract"*, which is exactly the sentence that forces a reader to go check whether that's still true — and the fixture work in #2426 had to do precisely that check before it could reuse the shared builders.

[...]

PR #2437, opened 2026-07-25
Point progression-test-utils' tier/path builders at the shared fixtures (#2426)

### LINKED ISSUE #2426: Point progression-test-utils' tier/path builders at the shared IProficiency/IPath fixtures

ginderjeremiah:
Closes #2426.

## Summary

`WorkbenchPath`/`WorkbenchProficiency` are plain aliases of `IPath`/`IProficiency`, so `progression-test-utils.ts` was hand-rolling the same two contracts `tests/fixtures/proficiencies.ts` already centralises. Its `path`/`tier` builders now delegate to `makePath`/`makeProficiency`, restating only their **divergent** defaults as explicit overrides rather than the whole field set.

**2 files, +36 / −35.** Test-only; no production code, no behaviour change.

## What changed

- **`tier`** → `makeProficiency`, keeping `name: 'Blades'`, the four blank conlang/icon fields, and `xpGrowth: 1.4` as stated overrides. The 9 fields it shares with the contract fixture (`description`, `designerNotes`, `pathId`, `pathOrdinal`, `maxLevel`, `baseXp`, and the three arrays) are no longer restated.
- **`path`** → `makePath`, keeping `id: 5`, `name: 'Fire Path'`, `activityKey: Fire`.
- **`mapPath`** now goes to `makePath` **directly** rather than through `path`. `makePath` already defaults to the generated `Path {id}` name and a neutral `Physical` activity — exactly what the map suites want — so routing through `path` was setting `Fire Path`/`Fire` only to override both straight back. `mapTier` still routes through `tier`, because it does want that builder's blank conlang fields.

The blank-conlang divergence now carries the *reason* it's load-bearing, not just the assertion that it is: these suites drive the **editor**, where an empty field is the authoring state under test, so a generated `w0`/`p0`/`t0` would quietly pre-fill inputs the detail suites assert on. That's the same asymmetry `fixtures/skills.ts` documents for `ISkill`.

## The issue's file path was wrong

#2426 cites `UI/src/tests/routes/admin/workbench/progression/types.ts`. That file doesn't exist — the aliases are **production** code at `UI/src/routes/admin/workbench/progression/types.ts`. Doesn't change the work (the fix is entirely test-side), but it matters for the alias question below, which is a production-code question, not a test one.

## `Synthesis.test.ts` was already handled

#2426 notes that #2424 "should pick up `Synthesis.test.ts` directly". It did — `Synthesis.test.ts:50` already reads `makeProficiency({ id: 0, name: 'Geomancy' })` on `main`. I swept every remaining typed `IProficiency`/`IPath` construction in `src/tests/` to confirm; `progression-test-utils.ts` was the last one. That's what makes the measurement below land at exactly 1.

## Verification that it actually worked

The stated win is "one place to edit when the contract grows", so I measured it rather than asserting it. Adding a throwaway required field to **both** `IPath` and `IProficiency` and running `svelte-check`:

| | Test files with errors |
|---|---|
| Before | 2 — `fixtures/proficiencies.ts`, `progression-test-utils.ts` |
| After | **1** — `fixtures/proficiencies.ts` |

(The one remaining non-test error is `routes/admin/workbench/progression/progression-helpers.ts`, a production file — correct and expected.) The probe was reverted and is not in the diff.

I also proved the delegation is a **no-op on the produced objects**, since a silently-changed default is the exact risk the issue warns about. A throwaway suite ran the pre-change implementations verbatim beside the new ones and asserted `toEqual` for `path`/`tier`/`mapPath`/`mapTier` — bare and with overrides — plus identical sorted key sets, so nothing was added or dropped. Both passed; the file was deleted and is not in the diff.

## The alias question → #2436

#2426 asks as a side question whether `WorkbenchPath`/`WorkbenchProficiency` still earn their existence. I surveyed it and the answer is **probably not**, but acting on it doesn't belong here:

Five sibling `Workbench<Entity>` types exist, and **every one of them diverges from its contract** — `WorkbenchZone`/`WorkbenchEnemy` add derived fields, `WorkbenchItem`/`WorkbenchClass`/`WorkbenchLesson` widen an optional for the `-1` "None" sentinel. These two are the only pure aliases. They also live in `progression/types.ts` rather than `entities/*.ts`, and carry no `EntityConfig<T>` role, because the progression editor is bespoke rather than `EntityConfig`-driven. So the "consistency with the family" defence is nominal.

Removing them is a mechanical rename across **19 files, 11 of them production Svelte components** — which would swamp a 2-file test-only diff and make this harder to review. Filed as **#2436** with the full survey table.

## Test plan

- [x] `npx vitest run src/tests/routes/admin/workbench/progression/` — 10 files, **123 passed**.
- [x] `npm run test:unit:ci` — **4022 passed / 0 failed** (312 files).
- [x] `npm run lint` — clean; `svelte-check` 1224 files, **0 errors, 0 warnings**.
- [x] `npx prettier --check` on both changed files — clean.
- [x] The anti-drift claim measured directly (2 → 1 table above), not assumed.
- [x] Delegation proven output-identical against the verbatim old implementations, not inferred from the suites passing.
- **No backend mirror needed.** This is workbench admin-tooling scaffolding, not battle logic, so it isn't on the frontend/backend parity surface.

## One note on a comment I had to update

`progression-test-utils.ts` carries a load-bearing warning that it is evaluated *inside* a `vi.mock('$stores', …)` factory, so it must stay free of value imports that transitively reach `$stores`. This change adds the first value import below `$lib/api`, so that comment's justification ("everything below `$lib/api` is `import type`") was no longer accurate. The import is safe — `fixtures/proficiencies.ts`'s only import is `$lib/api`, which this module already value-imports — and the comment now names the three permitted value imports explicitly rather than stating a rule the file no longer follows.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KRFnV6XDTkMmRbVztFCRv5)_

### DIFF
--- a/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
@@ -62,10 +62,6 @@ export const resetStores = () => {
 	}
 };
 
-/* `WorkbenchPath`/`WorkbenchProficiency` are plain aliases of `IPath`/`IProficiency`, so the two builders
-   below are the shared contract fixtures with the workbench's divergent defaults stated on top (#2426).
-   Only the divergences are restated here — everything else follows the contract fixture. */
-
 /** Base path fixture — id 5, the path the detail suites open. */
 export const path = (over: Partial<WorkbenchPath> = {}): WorkbenchPath =>
 	makePath({ id: 5, name: 'Fire Path', activityKey: EActivityKey.Fire, ...over });
--- a/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
+++ b/UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
@@ -2,6 +2,7 @@ import { vi } from 'vitest';
 import { EActivityKey } from '$lib/api';
 import type { ProgressionStore } from '$routes/admin/workbench/progression/progression-store.svelte';
 import type { WorkbenchPath, WorkbenchProficiency } from '$routes/admin/workbench/progression/types';
+import { makePath, makeProficiency } from '../../../../fixtures/proficiencies';
 
 /* Shared scaffolding for the progression editor's component suites (#2405). Each suite still declares
    its own `vi.mock('$stores', …)` — `vi.mock` is hoisted within the file it appears in — but the factory
@@ -9,8 +10,9 @@ import type { WorkbenchPath, WorkbenchProficiency } from '$routes/admin/workbenc
 
    Because the suites resolve it from inside that factory, this module is evaluated *while `$stores` is
    being mocked, so it must stay free of runtime imports that transitively reach `$stores`* — a value
-   import that did would re-enter the factory resolving it. Everything below `$lib/api` is `import type`
-   and therefore erased, which is what keeps that true today. */
+   import that did would re-enter the factory resolving it. The only value imports here are `vitest`,
+   `$lib/api`, and `tests/fixtures/proficiencies` (whose own only import is `$lib/api`); everything else
+   is `import type` and therefore erased, which is what keeps that true today. */
 
 const catalogueKeys = [
 	'enemies',
@@ -60,49 +62,48 @@ export const resetStores = () => {
 	}
 };
 
+/* `WorkbenchPath`/`WorkbenchProficiency` are plain aliases of `IPath`/`IProficiency`, so the two builders
+   below are the shared contract fixtures with the workbench's divergent defaults stated on top (#2426).
+   Only the divergences are restated here — everything else follows the contract fixture. */
+
 /** Base path fixture — id 5, the path the detail suites open. */
-export const path = (over: Partial<WorkbenchPath> = {}): WorkbenchPath => ({
-	id: 5,
-	name: 'Fire Path',
-	description: '',
-	designerNotes: '',
-	activityKey: EActivityKey.Fire,
-	...over
-});
+export const path = (over: Partial<WorkbenchPath> = {}): WorkbenchPath =>
+	makePath({ id: 5, name: 'Fire Path', activityKey: EActivityKey.Fire, ...over });
 
 /**
  * Base tier fixture — id 0 on path 0, with blank conlang fields.
  *
+ * The blank `iconPath`/`word`/`pronunciation`/`translation` deliberately diverge from `makeProficiency`'s
+ * id-generated ones: these suites drive the *editor*, where an empty field is the authoring state under
+ * test, and a generated string would quietly pre-fill inputs the detail suites assert on. `xpGrowth: 1.4`
+ * likewise diverges so the XP curve is visibly non-flat.
+ *
  * A suite whose assertions depend on a different identity (TierDetail's populated conlang strings, say)
  * wraps this with its own defaults rather than changing them here; those divergences are load-bearing.
  */
-export const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency => ({
-	id: 0,
-	name: 'Blades',
-	description: '',
-	iconPath: '',
-	word: '',
-	pronunciation: '',
-	translation: '',
-	pathId: 0,
-	pathOrdinal: 0,
-	maxLevel: 10,
-	baseXp: 100,
-	xpGrowth: 1.4,
-	designerNotes: '',
-	levelModifiers: [],
-	levelRewards: [],
-	prerequisiteIds: [],
-	...over
-});
+export const tier = (over: Partial<WorkbenchProficiency> = {}): WorkbenchProficiency =>
+	makeProficiency({
+		id: 0,
+		name: 'Blades',
+		iconPath: '',
+		word: '',
+		pronunciation: '',
+		translation: '',
+		xpGrowth: 1.4,
+		...over
+	});
 
 /* The map suites (`ProgressionMap.test.ts`, `progression-map.test.ts`) build topology — several tiers
    across several paths — so identity is positional at their call sites (`mapTier(10, 0, 0)`), and their
    assertions read the generated `Path {id}` name back out of the rendered column header. The two
-   adapters below keep the field set in one place while preserving both. */
+   adapters below keep the field set in one place while preserving both.
 
-export const mapPath = (id: number, over: Partial<WorkbenchPath> = {}): WorkbenchPath =>
-	path({ id, name: `Path ${id}`, activityKey: EActivityKey.Physical, ...over });
+   `mapPath` goes to `makePath` directly rather than through `path`: the shared fixture already defaults to
+   the generated `Path {id}` name and a neutral `Physical` activity, so routing through `path` would set the
+   detail suites' `Fire Path`/`Fire` only to override both straight back. `mapTier` does go through `tier`,
+   because it wants that builder's blank conlang fields. */
+
+export const mapPath = (id: number, over: Partial<WorkbenchPath> = {}): WorkbenchPath => makePath({ id, ...over });
 
 export const mapTier = (
 	id: number,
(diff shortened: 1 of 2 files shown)


UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
```diff
@@ -60,49 +62,48 @@ export const resetStores = () => {
 	}
 };
 
+/* `WorkbenchPath`/`WorkbenchProficiency` are plain aliases of `IPath`/`IProficiency`, so the two builders
+   below are the shared contract fixtures with the workbench's divergent defaults stated on top (#2426).
+   Only the divergences are restated here — everything else follows the contract fixture. */
```
ginderjeremiah: **Nit** — this block is the third statement of the same two facts. The alias-and-divergence story is already in `fixtures/proficiencies.ts:7-10` (the cross-module map, which is where a reader coming from the other side finds it) and again, with the *reason*, in `tier`'s docblock 9 lines below. This one adds only "only the divergences are restated here", which the two builders demonstrate on their face now that they're three lines each.

Dropping it would leave each fact stated once and cut three comment sites to two — worth it given `CLAUDE.md`'s "keep comments to 1-2 lines unless the extra context is critical", and given that a duplicated rationale is a rationale that can drift out of sync.

Non-blocking either way.

---

ginderjeremiah (commented):
## Review: point `progression-test-utils`' `path`/`tier` builders at the shared `IPath`/`IProficiency` fixtures (#2426)

**Verdict:** **Approve** — the delegation is output-identical, the divergent defaults are stated rather than inherited, and it does exactly what #2426 asked. Submitted as `COMMENT` only because GitHub won't accept a gating verdict from the PR's own author; treat it as an approve with one nit.

_(One process note up front: CI is not green — see **CI** below — so I'm not merging this despite the approve.)_

### Correctness

I checked the delegation by hand rather than trusting the suites, since a silently-changed default is the exact failure the issue warns about. All four builders produce byte-identical objects:

- **`path`** — `makePath`'s `description: ''` / `designerNotes: ''` are the same neutral blanks the old literal set; `id`/`name`/`activityKey` are restated. ✅
- **`tier`** — the 9 inherited fields (`description`, `designerNotes`, `pathId`, `pathOrdinal`, `maxLevel: 10`, `baseXp: 100`, and the three arrays) match `makeProficiency`'s defaults one for one. ✅
- **`mapPath`** — old form was `path({ id, name: 'Path {id}', activityKey: Physical, ...over })`, i.e. it overrode all three of `path`'s divergences and inherited only the two blanks that `makePath` also defaults. So going direct is genuinely equivalent, not a shortcut. ✅
- **`mapTier`** — unchanged, still routes through `tier`, so it still gets the blank conlang fields the map suites want. ✅

The one spot worth calling out as *right* rather than lucky: `makeProficiency` generates `iconPath`/`word`/`pronunciation`/`translation` from `overrides.id`, so when `mapTier(10, …)` flows through `tier`, the id reaching the fixture is `10` and the generated values would be `icon-10`/`w10`/…. They don't leak, because `tier` states the blanks in the same literal that `...over` lands in and the blanks aren't among the keys `over` carries. That's correct, but it's the kind of thing a later edit could break by "simplifying" the blanks out — the docblock at `progression-test-utils.ts:76-79` now explains why they're load-bearing, which is the right guard.

### Nits

- `progression-test-utils.ts:65-67` — third statement of the alias/divergence rationale (inline comment).

### Test coverage

Test-only change with no behaviour change, so no new tests are owed — the existing 10 progression suites *are* the regression check, and they'd fail loudly on any shifted default, which is what #2426's constraint asked for. No frontend/backend parity obligation either: this is workbench admin scaffolding, not battle logic.

I verified the completeness claim independently rather than taking the description's word for it: swept `src/tests/` for remaining hand-rolled `IProficiency`/`IPath` constructions. `stores/proficiencies.svelte.test.ts:30`, `workbench/references.test.ts:117-119`, `proficiencies-view.test.ts:43`, `Proficiencies.test.ts:32` and `synthesis/Synthesis.test.ts` all already delegate to the shared fixtures — `progression-test-utils.ts` was in fact the last holdout. `lexicon-test-utils.ts` builds the derived `PathView`/`TierView`, a different shape, correctly out of scope.

The relative `../../../../fixtures/proficiencies` import matches the established convention — there's no `$tests` alias in `svelte.config.js`, and all 25-odd sibling fixture imports are relative the same way.

### What's good

- **The `vi.mock('$stores')` comment was updated instead of left to rot.** This module is evaluated inside the mock factory, and its safety argument was literally "everything below `$lib/api` is `import type`" — which this PR's first value import invalidated. Naming the three permitted value imports explicitly is the right fix, and the import is genuinely safe (`fixtures/proficiencies.ts` imports only `$lib/api`, which this module already value-imports).
- **`mapPath` bypassing `path` is the right call, and it's explained.** Routing through a builder only to override every one of its divergences is noise; the comment says so, so the next reader doesn't "fix" it back.
- **#2436 was split out rather than smuggled in.** The alias question is a 19-file rename across 11 production components — landing it here would have buried a 2-file test-only diff. Correctly filed with the full survey table.

### CI

`all-checks-passed` is red, but the only failing job is `test-e2e (firefox)`, and it failed in `docker compose up` with `Get "https://registry-1.docker.io/v2/": Client.Timeout exceeded while awaiting headers` — a Docker Hub registry timeout, 24s in, before any test ran. Pure infrastructure. `test-e2e (chromium)`, `test-e2e (webkit)`, `test-frontend` and `verify-frontend` are all green on the same commit, and a frontend test-fixture diff cannot plausibly cause a registry pull to time out. A re-run of that one job should clear it.

---

UI/src/tests/routes/admin/workbench/progression/progression-test-utils.ts
ginderjeremiah: Agreed and dropped in `335aade`. The block was doing the least work of the three sites — the cross-module map in `fixtures/proficiencies.ts` is where a reader arriving from the other side lands, and `tier`'s docblock is the only one carrying the *reason* the blanks are load-bearing, which is the fact worth protecting from a later "simplification". Three comment sites down to two, each fact stated once.

Progression suites still 10 files / 123 passed, prettier clean.

---

PR #2438, opened 2026-07-25
Share the parry/dodge chance products between the engine and CombatRating (#2429)

### LINKED ISSUE #2429: CombatRating re-derives the parry/dodge chance products by hand instead of sharing them with the engine
Raised in the review of #2428 (which closed #2416 by clamping `CombatRating`'s proc-chance products to `[0, 1]`).

## Problem

`CombatRating`'s class doc-comment stakes its correctness on an explicit anti-drift rule:

> Reuses the exact `Battler`/`BattleSkill` methods the engine itself uses (`BattleSkill.CalculateRawDamage`, `Battler.AmplifyDamage`, `Battler.ComputeNetDamage`, `Battler.GetCooldownMultiplier`) — the anti-drift mechanism that keeps the rating from silently diverging from the real damage pipeline.

The avoidance chance composition is the one piece of engine math that rule does **not** cover. `ParryChance × ParryChanceMultiplier` and `DodgeChance × DodgeChanceMultiplier` are spelled out independently in two places:

- the engine, at `BattleContext.cs:230-233` (the live proc draws);
- the rating, now behind `CombatRating.EffectiveParryChance`/`EffectiveDodgeChance` (#2428).

#2428 collapsed the *rating's* two internal copies into those helpers — which was the immediate DRY win, since the offense (riposte) and survivability readings disagreeing about what a parry chance means was precisely the bug being fixed. But the rating-vs-engine duplication remains, and that's the pairing the doc-comment's guarantee is actually about. If a future change reshapes how a parry chance composes (a third multiplier, a floor, a diminishing-returns curve), nothing forces the rating to follow, and the failure is silent: the rating keeps returning a plausible number that no longer prices what the engine does.

This is the same class of defect as #2416 itself — two sites independently deriving one probability — one level up.

## Suggested direction

Promote the products onto `Battler` (e.g. `Battler.EffectiveParryChance` / `EffectiveDodgeChance`) and have both the engine's draws and `CombatRating` read them, putting the composition under the same anti-drift rule as `AmplifyDamage`/`ComputeNetDamage`.

Note the clamp itself should **not** move with them. `CombatRating` clamps to `[0, 1]` because it prices an expectation; the engine has no need to, since `draw < chance` against a `[0, 1)` draw already saturates naturally. Folding the clamp into a shared accessor would change engine behavior for no reason — the shared piece is the product, and `ProcChance` stays a rating-side concern.

## Constraint

`Battler` is on the frontend/backend battle-parity surface, so this needs the mirrored change in `UI/src/lib/battle/battler.ts` and the parity test scenarios kept in step on both sides (per CLAUDE.md). That's why it was left out of #2428 — a `scope: small` clamp fix shouldn't drag the parity-mirrored engine along with it.

`CombatRating` itself is server-only and has no frontend mirror, so only the `Battler` half is parity-sensitive.

## Scope

Small — two accessors plus their frontend mirror, no behavior change on either side.

ginderjeremiah:
Closes #2429.

## Summary

`ParryChance × ParryChanceMultiplier` and `DodgeChance × DodgeChanceMultiplier` were spelled out independently in two places — the engine's live proc draws and `CombatRating`'s pricing of the same capability. #2428 collapsed the *rating's* two internal copies; this closes the remaining rating-vs-engine pair, which is the one `CombatRating`'s anti-drift doc-comment actually promises to cover.

Both products are promoted onto `Battler.EffectiveParryChance` / `EffectiveDodgeChance` — composed at consumption, exactly like `GetCooldownMultiplier` — and the engine's draws (`BattleContext.DamageTarget`) plus `CombatRating` now read them. **No behaviour change on either side**; the products are byte-identical, so the parity matrix and every rating assertion are undisturbed.

## How it addresses the issue

- **Backend engine** (`Battler.cs`, `BattleContext.cs`): two expression-bodied properties, read by the parry/dodge draws.
- **Frontend mirror** (`battler.ts`, `battle-step.ts`): `effectiveParryChance` / `effectiveDodgeChance` getters on `Battler`, consumed by `battleStep`. `Battler` is the parity-mirrored surface, so both sides now compose avoidance at one site each.
- **`CombatRating`**: reads through the shared accessors instead of re-deriving. The class doc-comment's anti-drift list now names them alongside `AmplifyDamage`/`ComputeNetDamage`/`GetCooldownMultiplier`, so the guarantee is structurally true rather than something a reader has to go verify.
- **The clamp stays rating-side**, per the issue: the rating clamps to `[0, 1]` because it prices an *expectation*, while `draw < chance` against a `[0, 1)` draw saturates naturally. Folding it into the shared accessor would change engine behaviour for no reason.
- **Follow-up comment on the issue swept too**: `CombatRating.cs:72`'s `(BattleContext.cs:230-233)` line-numbered citation is gone — those helpers now read through `Battler`, so the claim is structural and the comment cites by name like every other cross-reference in `Game.Core`.

### One naming decision worth flagging

The rating's two helpers are renamed `EffectiveParryChance`/`EffectiveDodgeChance` → **`ParryProcChance`/`DodgeProcChance`**. Keeping the old names would have left two same-named readings with *different* semantics one call apart — `Battler.EffectiveParryChance` (unclamped, what the engine draws against) and `CombatRating.EffectiveParryChance` (clamped, what the rating prices). The rename makes the clamped reading self-describing next to the existing `ProcChance` helper it delegates to.

### Where the frontend accessors live

They sit on `Battler` rather than in `battle-formulas.ts`. That file exists to share scalar formulas with **display** surfaces (the skills page consumes `cooldownMultiplier`); nothing renders a composed parry/dodge chance today, so a pure function there would be indirection with exactly one caller. Putting them on `Battler` is also the tighter mirror — the backend composes both inline on `Battler` for the same reason. Worth revisiting if a UI ever wants to show effective avoidance.

## Docs

`docs/backend-battle.md` → *seeded crit / dodge / parry* records the new invariant: one composition site per simulator, read live per draw, unclamped, with the clamp explicitly a rating-side concern.

## Test plan

Mirrored unit tests added to `BattlerTests.cs` ⇄ `battler.test.ts` with the **same scenarios and the same expected results** on both sides (per CLAUDE.md's battle-parity rule), following the existing `GetCooldownMultiplier` ⇄ `cdMultiplier` pairing in those files:

- no enabler ⇒ exactly `0` regardless of Luck / Agility (the authored-only, base-`0` enabler contract);
- authored chance × the Luck-/Agility-amplified multiplier (`0.2 × 1.04`, `0.25 × 1.04`);
- authored chance with no Luck / Agility ⇒ passes through verbatim (base-`1` multiplier);
- a product above `1` is **not** clamped (`0.8 × 2 = 1.6`) — pins the clamp as rating-side;
- read live: a mid-battle `ParryChanceMultiplier` buff moves the composed chance on the next read.

Verification run locally:

- [x] `dotnet test` — **3,514 passed, 0 failed** (Game.Core 1461, Game.Application 1105, Game.Api 874, Game.Infrastructure 74), including `BattleSimulatorParityTests` and the full `CombatRatingTests` suite
- [x] `npm run test:unit:ci` — **4,030 passed, 0 failed**, including `battle-simulation-parity.test.ts` and `battle-step.test.ts`
- [x] `npm run lint` (eslint + `svelte-check`) — 1224 files, 0 errors, 0 warnings
- [x] `prettier --check` on the touched frontend files

No follow-up issues — the refactor turned up nothing beyond the doc-comment sweep already folded in above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FcJ2ufkw1nzzZ1n1umHnJY)_


UI/src/tests/lib/battle/battler.test.ts
```diff
@@ -125,6 +125,100 @@ describe('Battler', () => {
 			expect(battler.cdMultiplier).toBe(2);
 		});
 
+		// effectiveParryChance/effectiveDodgeChance = enabler × multiplier (#2429) — the products the engine's
+		// own draws read (see battleStep) and the backend combat rating prices. Mirrors the backend BattlerTests
+		// EffectiveParryChance/EffectiveDodgeChance cases with the same scenarios and results.
+		it('parries at exactly 0 regardless of Luck with no ParryChance enabler', () => {
```
ginderjeremiah: **Nit (pre-existing drift, not introduced here).** These eight cases land inside `describe('reset')`, which by now covers construction, `cdMultiplier`, and — with this PR — the avoidance accessors. The live-read case at line 177 is the clearest mismatch: it calls `applyEffect` and asserts a mid-battle recomposition, which is not a reset concern, and the file already has a separate `describe('applyEffect MaxHealth clamp')` block for that kind of test.

The backend mirror doesn't have this problem — `BattlerTests` groups by member with `── … ──` banners, so the two suites are drifting apart structurally even while the scenarios stay in lockstep.

Not worth holding the PR for, since the misgrouping predates it (`cdMultiplier` is already there). Worth a small follow-up to lift the accessor cases out of `reset` into their own `describe` so the parity mirror reads the same way on both sides.

---

ginderjeremiah (commented):
## Review: share the parry/dodge chance products between the engine and `CombatRating` (#2429)

**Verdict: Approve** — the refactor does exactly what #2429 asked, the products are byte-identical so no behaviour moves, and the parity mirror + mirrored tests are both there. Submitted as a `COMMENT` because GitHub won't take a gating verdict on my own PR; treat it as an approve. Nits only, plus one follow-up filed.

I verified the central claim rather than taking the description's word for it: `rg` across the tree for `ParryChanceMultiplier|DodgeChanceMultiplier` now finds the products composed at **exactly two** code sites — `Game.Core/Battle/Battler.cs:91,98` and `UI/src/lib/battle/battler.ts:163-176` — with every other hit being docs, tests, or the attribute definitions. The old `BattleContext.cs:230-233` and `CombatRating.cs:72,77` derivations are gone, and both new readings are `_attributes[X] * _attributes[Y]` / `getValue(X) * getValue(Y)`, i.e. identical arithmetic in identical order to what they replaced. The seeded draw order in `DamageTarget` ↔ `battleStep` is untouched, so the parity matrix genuinely can't move.

The judgment calls in the description hold up:

- **`ParryProcChance`/`DodgeProcChance` rename** — right call. Two same-named readings one call apart with different clamping semantics is exactly the kind of footgun a later reader resolves wrongly. The new names describe the clamped reading and sit next to the `ProcChance` helper they delegate to.
- **Clamp stays rating-side** — correct per the issue, and `Battler.cs:88-89`'s doc comment states *why* rather than just *that*, which is what stops a future edit from "tidying" the clamp into the shared accessor and silently changing engine behaviour.
- **Accessors on `Battler` rather than `battle-formulas.ts`** — agreed. `cooldownMultiplier` lives there because the skills page renders it; nothing renders composed avoidance today, so a shared function would be indirection with one caller, and `Battler` is the tighter mirror of the backend.
- **`CombatRating.cs:71-74` now cites by name** — the line-numbered `(BattleContext.cs:230-233)` citation flagged in #2429's comment is gone, and the "mirrors the engine" claim is now structurally true instead of something a reader has to go verify. That sweep landing here rather than being deferred is the right scoping.

### Nits

- **`Game.Core.Tests/Battle/BattlerTests.cs:13-18` — the class doc is now stale.** It enumerates what's mirrored: *"the `Battler.TakeDamage` toughness-curve/death cases and the `Battler.GetCooldownMultiplier` formula are asserted here with the same scenarios … so a future divergence in the **mitigation/cooldown math** fails on both sides."* Avoidance is now a third mirrored family, and the doc is the thing a future contributor reads to learn what the parity contract covers — an under-listing invites someone to add a backend-only case. Same edit on the frontend side if `battler.test.ts` carries an equivalent header. (Outside the diff, so no inline comment.)
- **Test grouping** — see the inline comment on `battler.test.ts:128`.
- **`docs/backend-battle.md:43` — consider trimming.** The invariant worth recording is "one composition site per simulator, read live, unclamped, clamp is rating-side"; naming all four accessors on both sides edges into the *"all operations of kind x must be routed through class y"* pattern CLAUDE.md's documentation guidance explicitly lists under AVOID, and it's the fourth restatement of a rationale the two doc comments and the tests already carry. The bullet was already one of the longest lines in the file. Not wrong, just heavy for a no-behaviour-change refactor.
- **`BattlerTests.cs:423-434` hand-rolls a `SkillEffect`.** That's now the third hand-rolled construction in `Game.Core.Tests` (alongside `BattleSimulatorParityTests`' private `MakeEffect` and `BattlePerformanceTests`' inline literal). Fine as a single named constant with a doc comment, but it's the same duplication class the #2425 fixture family has been sweeping on the frontend contract types — worth folding in if that sweep ever reaches the backend.

### Follow-up filed

- **#2439 — `CombatRating` re-derives the *crit* chance product by hand.** This PR closes the parry/dodge half of the anti-drift rule and leaves the crit half open: `skill.CriticalChance × CriticalChanceMultiplier` is still spelled out independently at `BattleContext.ResolvePlayerHit` (the live draw) and `CombatRating.ExpectedDirectHit` (`critExpectation`). Identical defect class, one mechanic over — and arguably the more load-bearing one, since `critExpectation` multiplies the whole direct-hit expectation rather than adding one term. Correctly out of scope here (#2429 is explicitly parry/dodge, and crit's per-skill enabler needs a different member shape), but it shouldn't be lost.

### Test coverage

Meets the project's rules, including the battle-parity one. Eight new cases per side with the **same scenarios and the same expected results**: the base-0 enabler contract (`0 × 1.1 = 0` under Luck 50 / Agility 50), the amplified product (`0.2 × 1.04`, `0.25 × 1.04`), pass-through at the base-1 multiplier, the deliberately unclamped `0.8 × 2 = 1.6`, and the live-read case. I checked the arithmetic against `StaticAttributeModifiers` (`1 + 0.002·LUK`, `1 + 0.002·AGI`) — every expectation is right. The two live-read helpers are equivalent across the boundary too: the backend passes `ParryMultiplierBuff.Amount` explicitly, the frontend relies on `applyEffect`'s `amount = effect.amount` default, and `makeEffect`'s `scalingAmount` of `0` makes its `Strength` default a no-op against the backend's `Luck`/`0` pairing.

Unchanged behaviour on both consumers is already pinned by the existing suites — `BattleContextTests`' fractional-chance and `0 × 1000` draw cases, `BattleSimulatorParityTests`, and `CombatRatingTests`' avoidance-pole case at `:63` — so no rating-side additions were needed. Local `dotnet test Game.Core.Tests` on the head commit: **1461/1461 green**.

### CI

`test-backend` failed on the first run with **1 failure out of 1461 in `Game.Core.Tests`**, and the console log names no test (the detail goes to the unpublished `TestResults` log). It does not reproduce locally on this exact commit — full green, including every new case. Given the diff is arithmetically identity-preserving, the likely culprits are the wall-clock budgets in `BattlePerformanceTests` (`MaxLengthRealisticCeilingMs`, the 2.0× stacking-flatness ratio) or `ProbabilityTableTests:246`'s unseeded `new Random()` concurrency case — both runner-load sensitive. I've re-run the failed job to confirm; if it goes green this is a flake to watch rather than anything in this PR, and if it fails again the job needs to start surfacing the failing test name.

### What's good

The scoping discipline is the strongest part: the issue said promote the product and leave the clamp, and that's precisely the seam taken — no opportunistic clamping, no "while I'm here" reshaping of the draw path. Flagging the rename in the description rather than burying it is the right instinct, since it's the one reader-visible change. And the doc comments explain the *reasons* (why unclamped, why the clamp is rating-side) instead of restating the code, which is what makes this refactor actually resistant to the drift it was filed against.

---

ginderjeremiah:
Closing the CI loop from my review: the re-run of `test-backend` is **green**, and `all-checks-passed` is now ✅ on `c9b9ce2`. The earlier 1-of-1461 failure in `Game.Core.Tests` was a flake, not anything in this PR — consistent with the diff being arithmetically identity-preserving and with the suite passing locally on this exact commit.

Filed **#2441** for the reason it took a local run plus a CI re-run to establish that: `coverage.ps1` runs the suite inside `dotnet-coverage collect`, which relays only the summary, so a failing backend test never names itself in the log or the check output. That issue also flags `ProbabilityTableTests`' unseeded `new Random()` weights, which would make a real failure there unreproducible even once the name is legible.

Verdict is unchanged — approve, with the nits above unaddressed. Not merging on your behalf, since those are still open.

---

ginderjeremiah:
Thanks — all three actionable nits addressed in 6eb7c3d, and the CI failure is confirmed a flake.

## CI: flake, confirmed two ways

Your re-run of `test-backend` on the **identical** commit (`c9b9ce2`) went green, and `all-checks-passed` followed. Independently, I reproduced CI's exact command locally rather than the plain `dotnet test` I'd used pre-push — `dotnet-coverage collect -f cobertura -o … "dotnet test Game.sln --no-build --no-restore"`, i.e. the profiler-instrumented path, which is the one load-sensitivity difference between my pre-push run and CI:

```
Passed! - Failed: 0, Passed: 1461 - Game.Core.Tests.dll
Passed! - Failed: 0, Passed:   74 - Game.Infrastructure.Tests.dll
Passed! - Failed: 0, Passed:  874 - Game.Api.Tests.dll
Passed! - Failed: 0, Passed: 1105 - Game.Application.Tests.dll
```

So: green under instrumentation locally, green on re-run in CI, and the diff is arithmetically identity-preserving. Nothing in this PR.

I've filed **#2440** for the diagnosability gap you flagged — the failing test's name goes to `TestResults/*.log` under `bin/`, which is never uploaded, so a red `test-backend` says *that* one of ~3,500 tests failed and nothing about *which*. That's what turned this into a repro cycle instead of a one-line read. I deliberately scoped it to fixing the **reporting** rather than guessing at the flaky test: I can enumerate the load-sensitive candidates (`BattlePerformanceTests`' wall-clock budgets, `ProbabilityTableTests`' unseeded `new Random()`) but not which one actually fired, and "fix the test I suspect" is how a flake gets papered over instead of found. Reporting first, then triage with evidence.

## Nits

**`BattlerTests` class doc** — fixed. Avoidance is now listed as the third mirrored family and the summary line reads "mitigation/cadence/avoidance math". You're right that this is the load-bearing one of the three: it's the doc that tells a future contributor what the parity contract covers, so under-listing it is an active invitation to add a backend-only case.

**Test grouping** — done rather than deferred. `cdMultiplier` and the avoidance cases are lifted out of `describe('reset')` into their own blocks, so the frontend groups by member the way `BattlerTests`' `── … ──` banners do. Pure move: no scenario, assertion, or expectation changed, and the suite is 47/47 green before and after. You called it not worth holding the PR for, and as a standalone chore I'd agree — but my eight cases were about to make the mismatch two-thirds of that `describe`, and the structural drift you identified is precisely between the two files this PR is trying to keep readable as a mirror. Fixing it here costs an indentation-only diff; a follow-up issue would have cost more to write than the fix.

**`docs/backend-battle.md:43`** — trimmed, and the CLAUDE.md citation is fair. Naming all four accessors was exactly the *"all operations of kind x must be routed through class y"* shape the documentation guidance lists under AVOID, and it was a fourth restatement of a rationale the two doc comments and eight tests already carry. Now down to the invariant that actually needed recording — one composition site per simulator, unclamped, clamp is rating-side — at roughly a third the length.

**Hand-rolled `SkillEffect`** — agreed, no change. It's one named constant with a doc comment, and folding it into a shared backend fixture is the #2425 sweep's job, not this PR's; doing it here would mean inventing the backend half of that convention as a side effect of a parry/dodge refactor.

**#2439 (crit)** — agreed on both counts: same defect class, and `critExpectation` multiplying the whole direct-hit expectation does make it the more load-bearing half. Deliberately left out, since #2429 scopes to parry/dodge and crit's enabler is per-*skill* (`skill.CriticalChance`) rather than an attribute, so it needs a different member shape than a `Battler` property — it's a genuinely different design call, not the same edit one mechanic over.

CI is re-running on `6eb7c3d`.

---

PR #2442, opened 2026-07-25
Scope ApiSocket lifecycle events to the socket they came from (#2432)

### LINKED ISSUE #2432: ApiSocket.handleClose acts on a close from any socket, not just the current one
Noticed while reviewing #2431 (which fixes #2422). Pre-existing and **not** introduced by that PR, but #2431's teardown guard is a time-scoped mitigation of one instance of it, so it's worth making structural.

## Problem

`UI/src/lib/api/api-socket.ts` binds the close handler to the *instance*, not the socket:

```ts
socket.onclose = this.handleClose.bind(this);
```

`handleClose(ev)` never checks that `ev` came from `this.socket`, so a close event from a **superseded** socket mutates state that now belongs to a different connection — `clearConnectionStableTimer()` (cancelling the live socket's budget refill), `settleQueuedRequests` (erroring the live session's not-yet-sent commands), `socketAuthRetries++`, and a `refreshTokens()` that burns a single-use refresh token for a socket nobody is waiting on.

The reads it makes decisions from — `socketOpened`, `handshakeTokenExpiry`, `socketAuthRetries` — are all instance-scoped too, so a stale close is judged against the *current* connection's state rather than its own.

## Reachability

`ensureSocket` collapses opens and `hasUnclosedSocket()` blocks a second socket while one is CONNECTING/CLOSING, so this needs a teardown to create the overlap:

1. `disconnect()` marks `teardownRequested` and aborts a still-CONNECTING handshake (the browser surfaces that as **1006**, not `NORMAL_CLOSURE`, so it doesn't take the clean-closure early return).
2. Before that close event lands, a `sendSocketCommand` lifts the mark.
3. The straggling close now falls past #2431's `teardownRequested` short-circuit into the refresh/retry path — and because `resetHandshakeState()` blanked `handshakeTokenExpiry`, `wasHandshakeTokenDemonstrablyFresh()` returns `false`, so it no longer even gets the #2369 "don't refresh on a demonstrably fresh token" protection it would incidentally have had before.

**Not reachable at either current call site**, which is why it isn't a bug today: `engine.ts:397` (`SocketReplaced`) navigates to login, and `CharacterSwitcher.svelte:76` awaits an HTTP `Players/SwitchPlayer` round trip — both put an async gap after `disconnect()` that is far longer than the close event's turnaround. It's a latent footgun waiting on a caller that tears down and re-commands synchronously.

## Suggested direction

Make close handling socket-scoped rather than instance-scoped, e.g. capture the socket in the handler closure and ignore the event when it isn't `this.socket`:

```ts
socket.onclose = (ev) => {
    if (this.socket !== socket) {
        return;
    }
    this.handleClose(ev);
};
```

That makes "a close only ever affects the connection it came from" structural, rather than something each new branch in `handleClose` has to re-derive. Worth checking `handleError`/`receiveResponse` for the same shape while in there.

Also worth revisiting once done: `docs/frontend-sockets.md` currently states flatly that "a close arriving after the mark is set is terminal" — true while the mark is set, but the mark is liftable, so the wording is slightly stronger than the guard.

## Verification

`UI/src/tests/lib/api/api-socket.test.ts` already has the machinery (`socketInstances`, `rejectHandshake`, `internals()`); a case that tears down, sends a new command, then fires the *old* socket's 1006 and asserts no `refreshTokens` call and an untouched budget would pin it.

## Scope

Small — one frontend file plus tests. No backend, no parity surface.

ginderjeremiah:
Closes #2432.

## Summary

`ApiSocket` bound its socket handlers to the **instance**, so a close from a *superseded* socket mutated state that by then belonged to a different connection — cancelling the live socket's budget refill, erroring commands it was about to send, spending auth-retry budget, and burning a single-use refresh token for a handshake nobody was waiting on.

Handlers are now bound **per socket**, so "an event only ever affects the connection it came from" is structural rather than something each new branch in `handleClose` has to re-derive.

## One correction to the issue's reachability analysis

The issue's chain is `disconnect()` aborts a CONNECTING handshake → a `sendSocketCommand` lifts the mark → the straggling 1006 falls past #2431's `teardownRequested` short-circuit. That part is right, but on its own it doesn't produce a *stale* socket: while the aborted socket is `CLOSING`, `hasUnclosedSocket()` blocks a replacement, so `this.socket` is still the socket that closed and an identity check would pass.

The overlap needs one more step, which is what actually makes this reachable: **`readyState` reaches `CLOSED` before the close event is dispatched.** In that window `hasUnclosedSocket()` is `false`, so `ensureSocket` opens a replacement (the pre-emptive refresh resolves in microtasks when the token is already fresh, so the whole open can complete before the queued close task runs) — and *that* is when the straggler lands on a connection it never belonged to. Same fix, and it also covers the plain drop-then-reconnect case, not just the teardown one. The tests drive this interleave directly (`supersedeCurrentSocket`).

## What a superseded socket still keeps

A blanket "ignore everything stale" guard would have traded this bug for two others, so the scoping is per concern rather than per handler. **The same interleave that delivers a late close also delivers events queued ahead of it**, so both directions are real:

- **Its own in-flight requests are still settled.** `handleClose` is what settles them, and a stale socket's requests can never be answered — skipping them would leave callers waiting out the full 30s per-request timeout. `InFlightRequest` therefore records the socket it was sent on and `settleInFlightRequests(socket, …)` settles exactly that socket's requests, ahead of the staleness check.
- **A response or push that genuinely arrived on it is still delivered.** Dropping it would tell the caller of a non-idempotent command (`DefeatEnemy`) the connection was lost for work the server actually completed, because the close right behind it reports `CONNECTION_LOST_ERROR`. Responses match by command id, which is unique per instance, so a stale answer can only ever resolve the request it belongs to.

What is scoped out is everything that describes the *current* connection: the stability-timer refill, the queued-but-unsent commands, the auth-retry budget and refresh, pong tracking, and the error hook. A `ping` is answered on the socket that asked rather than on `this.socket`.

## Docs

`docs/frontend-sockets.md` had a sentence explicitly recording this as a known gap ("that is time-scoped rather than structural … (#2432)"); it's replaced by a bullet stating the invariant and both halves of what a superseded socket keeps.

## Test plan

Five cases added to `api-socket.test.ts` under `socket-scoped event handling`:

- a superseded close doesn't refresh, spend retry budget, or cancel the live socket's stability timer / keepalive;
- a superseded **clean** close doesn't settle the live connection's queued-but-unsent commands (both still flush on the replacement);
- a superseded close **does** still settle the requests that socket carried, while the replacement's own in-flight command resolves normally;
- a response that arrived on the socket before it was superseded still reaches its caller with data, not `CONNECTION_LOST_ERROR`;
- a superseded socket's ping is answered on that socket, its pong reports no latency, and its error raises no toast.

Verification run locally:

- [x] `npx vitest run src/tests/lib/api/api-socket.test.ts` — 77 passed
- [x] Negative direction checked per behaviour rather than in aggregate: deleting the guard fails 3 cases; deleting only the `inFlight.socket !== socket` filter fails the settle case; re-adding a blanket `onmessage` guard fails the delivery and keepalive cases
- [x] `npm run test:unit:ci` — **4,027 passed, 0 failed** (312 files)
- [x] `npm run lint` (eslint + `svelte-check`) — 1224 files, 0 errors, 0 warnings
- [x] `prettier --check` on both touched frontend files

No backend or parity surface is touched, so the backend suite is unaffected. No follow-up issues — nothing surfaced beyond the doc sweep folded in above.


UI/src/tests/lib/api/api-socket.test.ts
```diff
@@ -1656,4 +1656,130 @@ describe('ApiSocket', () => {
...
+			live.onopen?.();
+			await flushMicrotasks();
+			const onLive = JSON.parse(live.send.mock.calls[0][0]);
+
+			stale.onclose?.({ code: 1006 });
+			await flushMicrotasks();
+
+			expect((await stranded).error).toBe('Connection lost. Please try again.');
+			// The replacement's own in-flight command survives its predecessor's close and still resolves.
+			receive(live, JSON.stringify({ id: onLive.id, name: 'NewEnemy', data: {} }));
```
ginderjeremiah, line 1761: **Should-fix** — this last line asserts nothing, and it's the half of the test that would pin the new per-socket filter.

`receive()` on a request that is no longer in `inFlightRequests` is a silent no-op (`receiveResponse` guards `if (inFlight)`), so if the replacement's command *had* been wrongly settled by its predecessor's close, this line would still run clean and the test would still pass. The comment above it claims the opposite of what's actually checked.

The consequence is bigger than one comment: **delete the `inFlight.socket !== socket` check in `settleInFlightRequests` and the whole suite stays green.** The `stranded` assertion pins the "settle ahead of the guard" exception, but nothing pins the scoping that stops it settling *everything* — which is the reason `InFlightRequest.socket` was added in the first place.

`supersedeCurrentSocket` already sends the live command, it just drops the promise. Returning it makes this a one-line fix:

```ts
const live = lastWs();
expect(live).not.toBe(stale);
return { stale, live, livePromise };
```

then here:

```ts
receive(live, JSON.stringify({ id: onLive.id, name: 'NewEnemy', data: {} }));
const liveResponse = await livePromise;
expect(liveResponse.error).toBeUndefined();
```


---

UI/src/lib/api/api-socket.ts
```diff
@@ -238,10 +241,28 @@ export class ApiSocket {
...
-		socket.onclose = this.handleClose.bind(this);
+		// Handlers are scoped to the socket they were bound to, not to the instance: a superseded socket can
+		// still deliver events — `readyState` reaches CLOSED before the close event is dispatched, so a
+		// reconnect can slip in ahead of it — and everything these handlers touch (the stability timer, the
+		// queued commands, this session's auth-retry budget) belongs to whichever connection is current by
+		// then. `handleClose` takes the socket rather than being wrapped, because a stale close still has to
+		// settle the requests its own socket carried before it bails.
+		socket.onopen = this.scopedToSocket(socket, () => this.onStart());
+		socket.onmessage = this.scopedToSocket(socket, (ev: MessageEvent) => this.receiveResponse(ev));
+		socket.onerror = this.scopedToSocket(socket, (ev: Event) => this.handleError(ev));
```
ginderjeremiah, line 258: **Should-fix** — the reachability argument for scoping `onmessage` is the same imprecision this PR corrected for `onclose`, and scoping it has a real (if narrow) cost the comment doesn't own.

The description says these three "are hard to reach (a superseded socket is already `CLOSED`, so it can't open, message, or error again)". A CLOSED socket can't *emit* a new event, but an event **queued before** the close is still dispatched after `readyState` reaches CLOSED — that is precisely the interleave the rest of this PR is built on. Message events are queued ahead of the close, so `[taskA opens a replacement] → [stale message task] → [stale close task]` is exactly as reachable as the close race itself, not less.

In that window the behaviour changes: a response that genuinely arrived on the stale socket used to resolve its in-flight request; now it's dropped, and the close ~immediately behind it settles the same request with `CONNECTION_LOST_ERROR`. So the caller of a **non-idempotent** command (`DefeatEnemy` is the example this file already calls out for not blind-resending) is told the connection was lost for a command the server actually answered, and may retry it. Server pushes queued ahead of the close (`SocketReplaced`, `ChallengeCompleted`) are dropped the same way.

The `handleClose` exception exists for this exact reason — "a stale socket's requests can never be answered, so settle them" — and here they *were* answered. Two ways out, either is fine:

1. Mirror the exception: keep the guard for the connection-level bits (pong tracking, and the `ping` → `pong` reply, which must not go out on the live socket) but still deliver a response whose in-flight entry names the socket the message came from — `inFlightRequests` already carries the socket now.
2. Accept the drop as the simpler contract, but say so: rest the comment on "a stale connection's answer is not worth the complexity of routing it, and its close settles the caller either way" rather than on "it can't message again", which isn't true.

Worth getting the rationale right regardless — the "it can't happen" framing is the kind of thing a later reader trusts instead of re-deriving.

---

UI/src/tests/lib/api/api-socket.test.ts
```diff
@@ -1656,4 +1656,130 @@ describe('ApiSocket', () => {
...
+		const stableTimer = (s: ApiSocket) =>
+			(s as unknown as { connectionStableTimer: ReturnType<typeof setTimeout> | null }).connectionStableTimer;
+
+		/** Drives the interleave the guard exists for: the current socket reaches CLOSED (its close event
+		 *  still queued), then a command opens a replacement, so the stale close lands on a live connection.
+		 *  The command also lifts any teardown mark, which is what puts the straggler past handleClose's
+		 *  teardown short-circuit in the first place. */
+		/** Makes the next socket stay CONNECTING, so commands sent through it sit queued-but-unsent. A
+		 *  function expression (not an arrow) so the mock can still be invoked with `new WebSocket()`. */
+		const connectingSocketOnce = () => {
```
ginderjeremiah: **Nit** — these two doc comments are stacked on `connectingSocketOnce`, but the first one describes `supersedeCurrentSocket` (defined at :1681, which ends up with no doc at all). Looks like the helper was moved above it and the comment didn't follow. Move the first block down to `supersedeCurrentSocket`.

---

ginderjeremiah (commented):
## Review: scope `ApiSocket` lifecycle events to the socket they came from (#2432)

**Verdict: Approve, with two should-fixes** — the fix is the right shape and the correction to the issue's reachability analysis is correct, but the new per-socket in-flight filter isn't actually pinned by a test, and scoping `onmessage` has a behaviour cost the comment doesn't own. Neither blocks merge; both are small and belong in this PR rather than a follow-up. Submitted as a `COMMENT` because GitHub won't take a gating verdict on my own PR — treat it as an approve-once-addressed.

I checked the description's central correction rather than taking it on faith, and it holds. The issue's chain (teardown aborts a CONNECTING handshake → a command lifts the mark → the 1006 falls past #2431's short-circuit) really doesn't produce a *stale* socket on its own: `hasUnclosedSocket()` (`api-socket.ts:370-372`) tests `readyState !== CLOSED`, so a CLOSING socket blocks a replacement and `this.socket` would still be the closing socket when its event lands. The extra step the PR identifies — CLOSED is reached *before* the close event is dispatched, so `ensureSocket` can open a replacement in the gap (the pre-emptive refresh resolves in microtasks on a fresh token, well inside one task) — is what makes this reachable, and it generalises the fix past the teardown case to any drop-then-reconnect. `supersedeCurrentSocket` drives exactly that interleave rather than faking `this.socket`, which is the right way to pin it.

The deliberate exception is the right call and correctly implemented. Settling ahead of the guard is necessary — a stale socket's requests can never be answered, so a blanket early return would trade this bug for 30s hangs — and `settleInFlightRequests(socket, …)` makes the scoping two-sided instead. The per-entry `delete` inside the `for…of` is safe (deleting the current key during Map iteration is well-defined), and `settleWithError` re-resolving an already-resolved promise is a no-op, so the existing `timeout-then-close double-settle guard` suite still holds.

### Should-fix

- **`api-socket.test.ts:1758` — the new per-socket filter has no test that fails without it.** `receive()` on a pruned request is a silent no-op, so the "the replacement's own in-flight command survives" half asserts nothing; delete the `inFlight.socket !== socket` check in `settleInFlightRequests` and the suite stays green. `supersedeCurrentSocket` already sends that command and just discards the promise — returning it makes this a one-line fix. Inline comment has the diff. This matters more than a normal missing assertion because `InFlightRequest.socket` exists *solely* for that filter.
- **`api-socket.ts:244-252` — the reachability claim for `onmessage`/`onerror` repeats the imprecision this PR corrected for `onclose`.** A CLOSED socket can't emit a new event, but one queued *before* the close is still dispatched after — the same interleave the fix is built on. So a response that genuinely arrived on the stale socket is now dropped and the close behind it reports `CONNECTION_LOST_ERROR` for a command the server actually answered (`DefeatEnemy` being the non-idempotent example this file already calls out). Either mirror the `handleClose` exception for messages whose in-flight entry names that socket, or accept the drop and rest the comment on that trade rather than on "it can't message again". Details inline.

### Nits

- **`api-socket.test.ts:1667-1673`** — two doc comments stacked on `connectingSocketOnce`; the first describes `supersedeCurrentSocket`, which ends up undocumented. Inline.
- **`api-socket.ts:477`** — `receiveResponse` still answers a `ping` with `this.socket?.send('pong')`. Correct today only because the wrapper just proved `this.socket === socket`; taking the socket as a parameter (the way `processCommandQueue:456` now hoists its local `socket`) would make it structural rather than a second-order consequence of the guard. Same argument the PR makes for `handleClose` taking its socket explicitly.

### Test coverage

Meets the frontend rule (`UI/src/tests/lib/api/api-socket.test.ts` mirrors the source path; no backend or parity surface is touched, so nothing is owed on the other side). The four cases target the guard well and I confirmed the negative direction by inspection rather than by the description's word:

- the 1006 case pins the auth-retry/refresh/stability-timer decisions — without the guard the stale close reaches `refreshTokens()` and `clearConnectionStableTimer()`;
- the clean-close case is the strongest of the four, since `code: 1000` is the branch that stops the keepalive and settles the queue as terminal, and both commands are then observed flushing on the replacement in order;
- the message/error case does fail without the guard (`this.socket?.send('pong')` lands on the live socket);
- the fourth is the gap above — it pins the exception but not the filter.

Two smaller things I checked and am satisfied with: `connectingSocketOnce` genuinely needs `function` rather than an arrow (the mock is invoked with `new`), and the `'ping'` filter on `live.send.mock.calls` is necessary because `onStart` → `attemptPing` sends one before the flush.

### Docs

`docs/frontend-sockets.md` — good trade. The removed clause was an explicit "known gap (#2432)" admission, and the replacement states the invariant plus its single exception, which is the right altitude for that file. Long, but every bullet around it is, and it earns the length by stating *why* the exception exists.

### What's good

The strongest part is that the description corrects the issue rather than implementing it verbatim — CLAUDE.md is explicit that an issue's analysis shouldn't be assumed correct, and finding that the stated chain is insufficient (and that the real trigger generalises the fix beyond teardown) is exactly that check being done properly. Recognising that a blanket early return would have been a regression, and fixing it by scoping the settle in *both* directions instead of special-casing, is the difference between a guard and an actual invariant. `scopedToSocket` also keeps the guard in one place rather than at four call sites, which is what makes the "each new branch shouldn't re-derive this" claim true going forward.

---

UI/src/tests/lib/api/api-socket.test.ts
ginderjeremiah: Fixed in 247a99d — you're right, and I confirmed it the way I should have the first time: deleting the `inFlight.socket !== socket` check kept all 76 tests green.

Took the suggested fix (`supersedeCurrentSocket` returns `livePromise`, the test awaits it and asserts no `error`). With it, deleting that check fails **`still settles the requests the superseded socket itself carried`** — so the filter is pinned in both directions now: the `stranded` assertion for the exception, the `livePromise` one for the scoping.

---

UI/src/lib/api/api-socket.ts
ginderjeremiah: Agreed, and this one is the more embarrassing of the two — I corrected exactly this imprecision for `onclose` and then reproduced it one line down. Took **option 1** in 247a99d.

`receiveResponse` now takes its socket and scopes per concern rather than being wrapped:

- **`pong`** — ignored when the socket isn't current. A straggler must not clear the live socket's missed-ping count or report its round trip as the live connection's latency.
- **`ping`** — answered on `socket`, not `this.socket`. That also settles your other nit: the reply is structural now instead of correct-only-because-the-wrapper-just-proved-them-equal. A CLOSING/CLOSED socket discards the send, which is the intent.
- **responses / pushes** — delivered whichever socket carried them, for the reason you give: the close behind it is about to report `CONNECTION_LOST_ERROR` for work the server actually completed.

One thing I didn't take from the suggestion: I match responses by command id alone rather than additionally checking the in-flight entry names the message's socket. Ids are instance-unique and the server only answers what it received, so that check can't ever be false — it would read as a real possibility to the next reader. The comment states the invariant instead.

New test `still delivers a response that arrived on the socket before it was superseded` pins it; re-adding the blanket guard fails it and the keepalive case.

---

PR #2443, opened 2026-07-25
Consolidate the hand-rolled IItem test builders onto a shared fixture (#2433)

### LINKED ISSUE #2433: Consolidate the hand-rolled IItem test builders onto a shared fixture

ginderjeremiah:
Adds `UI/src/tests/fixtures/items.ts` (`makeItem`) following the `fixtures/attributes.ts` / `fixtures/proficiencies.ts` / `fixtures/skills.ts` convention, and converts every module that constructs a **typed** `IItem` onto it. Test-only — no production code, no behaviour change.

## The target set is seven modules, not ~20

The issue estimated ~20 suites from `grep -rl "itemCategoryId: EItemCategory"`, while warning that the probe both under- and over-counts. It over-counts here, by a lot. Running the check the issue prescribes — add a throwaway required field to `IItem`, read `svelte-check` — gives the real answer:

| Converted (constructs a typed `IItem`) |
|---|
| `lib/battle/battle-sim-test-utils.ts` — the `equipmentFactory` registry literal |
| `lib/battle/item.test.ts` |
| `lib/common/item-requirements.test.ts` |
| `lib/engine/player/inventory-manager.test.ts` |
| `lib/engine/player/inventory-manager-reactivity.svelte.test.ts` |
| `routes/admin/workbench/references.test.ts` |
| `routes/game/screens/challenges/challenges-view.svelte.test.ts` |

Everything else the grep found pays **no** contract-drift tax and is deliberately left alone:

- **The inventory screens** (`EquipSlot`, `GridSlot`, `ItemDrawer`, `ItemTooltip`, `ModSlots`, `EquippedRail`, `Inventory`, `InventoryGrid`, `inventory-view`, `RewardTooltip`) build the live `Item` domain object via `as unknown as Item` — a different shape, with the merged mod totals attached, not the contract.
- **`ChallengeTooltip`, `challenge-unlocks`, `offline-summary`** assert deliberately *partial* literals to `IItem`, so a new required field never reaches them.
- **The workbench `reference`/`entities` suites and `engine.test.ts`** seed literals into a `staticData: {} as any` mock, which type-checks nothing.

Converting any of those would fill fields that were previously `undefined` — a real risk of changing what the suite renders. That criterion is recorded in the fixture's module comment, so a future `grep` still finding item literals under `src/tests/` reads as expected rather than as a gap.

## Reconciled defaults

`itemCategoryId` defaults to `Accessory`: `Weapon` implies a `weaponType` and the armour categories map to a specific `EEquipmentSlot`, so either would make the default load-bearing. The optional fields (`grantedSkillId`, `weaponType`, `requiredProficiencyId`, `retiredAt`) are left unset for the same reason.

One drifted default was **inert** and is reconciled to the shared value rather than preserved: `references.test.ts` used `requiredProficiencyLevel: 1`, which no assertion reads — its gear-gate case (`computeReferences('proficiencies', 5, …)`) asserts only the referencing item's *name*. Every default that **is** load-bearing stays stated at its adapter, with a one-line comment saying what reads it (5 Strength for the equipment-stat assertions; the Component slot the reactivity suite applies a mod into; the attributes + mod slot the reward tooltip renders).

## Test plan

- [x] `npm run lint` (eslint + `svelte-check`) — clean, 1225 files, 0 errors, 0 warnings.
- [x] `npm run test:unit:ci` — **312 files / 4035 tests passed**, no failures.
- [x] Each suite converted and run individually, so a changed default would have failed loudly rather than quietly weakening an assertion.
- [x] **The issue's verification check**: re-probing `IItem` with a required field afterwards produces exactly **one** test-file error — `src/tests/fixtures/items.ts`. (The five remaining errors are production files — `lib/battle/item.ts`, `workbench/entities/item.ts`, and three inventory components — which a genuinely new field would have to change anyway.)
- Backend untouched, so no `dotnet` run applies.

Closes #2433. The `IZone` half remains as #2434.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CJ7PcEwkqQRoG534KxCbH5)_

### DIFF
--- a/UI/src/tests/fixtures/items.ts
+++ b/UI/src/tests/fixtures/items.ts
@@ -2,21 +2,13 @@ import { EItemCategory, ERarity, type IItem } from '$lib/api';
 
 /* Shared `IItem` contract builder (#2433), following the `fixtures/skills.ts` convention.
 
-   The target set is narrower than a `grep` for item-shaped literals suggests: only seven modules
-   actually construct a *typed* `IItem`, and they are the only ones that pay the contract-drift tax.
-   Three groups are excluded on purpose:
-
-   · The inventory screens (`EquipSlot`, `GridSlot`, `ItemDrawer`, `ItemTooltip`, `ModSlots`,
-     `EquippedRail`, `Inventory`, `InventoryGrid`, `inventory-view`, `RewardTooltip`) build the live
-     `Item` domain object via `as unknown as Item`, not the contract — a different shape with the
-     merged mod totals attached.
-   · `ChallengeTooltip`, `challenge-unlocks` and `offline-summary` assert deliberately *partial*
-     literals to `IItem`, so a new required field never reaches them.
-   · The workbench `reference`/`entities` suites and `engine.test.ts` seed literals into a
-     `staticData: {} as any` mock, which type-checks nothing.
-
-   Filling the previously-`undefined` fields of any of those could change what a suite renders, so a
-   `grep` still finding item literals under `src/tests/` is expected, not a gap. */
+   The target set is narrower than a `grep` for item-shaped literals suggests: only a suite that
+   constructs a *typed* `IItem` pays the contract-drift tax. Three kinds of item literal are excluded
+   on purpose — those that build the live `Item` domain object via `as unknown as Item` (the
+   inventory screens), assert a deliberately *partial* literal to `IItem` (`challenge-unlocks`), or
+   seed an untyped `staticData` mock (the workbench `entities` suites). None of them see a new
+   required field, and filling their previously-`undefined` fields could change what a suite renders,
+   so a `grep` still finding item literals under `src/tests/` is expected, not a gap. */
 
 /**
  * Builds an {@link IItem} reference-data entry for tests. Everything but the id is a neutral
--- a/UI/src/tests/fixtures/items.ts
+++ b/UI/src/tests/fixtures/items.ts
@@ -1,0 +1,43 @@
+import { EItemCategory, ERarity, type IItem } from '$lib/api';
+
+/* Shared `IItem` contract builder (#2433), following the `fixtures/skills.ts` convention.
+
+   The target set is narrower than a `grep` for item-shaped literals suggests: only seven modules
+   actually construct a *typed* `IItem`, and they are the only ones that pay the contract-drift tax.
+   Three groups are excluded on purpose:
+
+   · The inventory screens (`EquipSlot`, `GridSlot`, `ItemDrawer`, `ItemTooltip`, `ModSlots`,
+     `EquippedRail`, `Inventory`, `InventoryGrid`, `inventory-view`, `RewardTooltip`) build the live
+     `Item` domain object via `as unknown as Item`, not the contract — a different shape with the
+     merged mod totals attached.
+   · `ChallengeTooltip`, `challenge-unlocks` and `offline-summary` assert deliberately *partial*
+     literals to `IItem`, so a new required field never reaches them.
+   · The workbench `reference`/`entities` suites and `engine.test.ts` seed literals into a
+     `staticData: {} as any` mock, which type-checks nothing.
+
+   Filling the previously-`undefined` fields of any of those could change what a suite renders, so a
+   `grep` still finding item literals under `src/tests/` is expected, not a gap. */
+
+/**
+ * Builds an {@link IItem} reference-data entry for tests. Everything but the id is a neutral
+ * placeholder so a suite states only what it asserts on.
+ *
+ * `itemCategoryId` defaults to `Accessory` — the categories that carry extra meaning (`Weapon`
+ * implies a `weaponType`, the armour slots map to a specific `EEquipmentSlot`) would make the
+ * default load-bearing. The optional fields (`grantedSkillId`, `weaponType`,
+ * `requiredProficiencyId`, `retiredAt`) are left unset for the same reason: an unequipped,
+ * ungated, un-retired item is the neutral case.
+ */
+export const makeItem = (overrides: Partial<IItem> & { id: number }): IItem => ({
+	name: `Item ${overrides.id}`,
+	description: '',
+	itemCategoryId: EItemCategory.Accessory,
+	rarityId: ERarity.Common,
+	iconPath: '',
+	attributes: [],
+	modSlots: [],
+	tags: [],
+	requiredProficiencyLevel: 0,
+	designerNotes: '',
+	...overrides
+});
--- a/UI/src/tests/lib/battle/battle-sim-test-utils.ts
+++ b/UI/src/tests/lib/battle/battle-sim-test-utils.ts
@@ -1,10 +1,10 @@
 import { Battler, newItem } from '$lib/battle';
 // Aliased: this module's own `makeSkill` builds a `SkillSpec` (pre-registration), not the contract.
 import { makeSkill as makeSkillContract } from '../../fixtures/skills';
+import { makeItem } from '../../fixtures/items';
 import {
 	EAttribute,
 	EDamageType,
-	EItemCategory,
 	ERarity,
 	type EModifierType,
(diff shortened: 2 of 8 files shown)


UI/src/tests/fixtures/items.ts
```diff
@@ -0,0 +1,43 @@
...
+   Three groups are excluded on purpose:
+
+   · The inventory screens (`EquipSlot`, `GridSlot`, `ItemDrawer`, `ItemTooltip`, `ModSlots`,
+     `EquippedRail`, `Inventory`, `InventoryGrid`, `inventory-view`, `RewardTooltip`) build the live
+     `Item` domain object via `as unknown as Item`, not the contract — a different shape with the
+     merged mod totals attached.
+   · `ChallengeTooltip`, `challenge-unlocks` and `offline-summary` assert deliberately *partial*
+     literals to `IItem`, so a new required field never reaches them.
+   · The workbench `reference`/`entities` suites and `engine.test.ts` seed literals into a
+     `staticData: {} as any` mock, which type-checks nothing.
```
ginderjeremiah: **Nit** — the *criterion* here is genuinely worth recording (I verified all three groups and every exclusion is correct), but naming thirteen files reads as an exhaustive census while not actually being one, and it's the stale-prone specificity `CLAUDE.md` warns about under "avoid extremely specific implementation details".

It's already incomplete on both axes:

- `routes/game/screens/codex/skill-provenance.test.ts:10-11` is a fourth partial-`as IItem` module (`({ id, name, grantedSkillId, retiredAt }) as IItem`), not listed in the second bullet.
- `workbench/entities/class.test.ts:99-102`, `entities/item.test.ts` and `welcome-back/WelcomeBackGate.test.ts:14` are three more `staticData`-mock seeders beyond the "`reference`/`entities` suites and `engine.test.ts`" the third bullet names.

Suggest keeping one example per group and stating the rule instead — something like *"suites that build the live `Item` domain object, assert a deliberately partial literal, or seed an untyped `staticData` mock are excluded: they pay no drift tax, and filling their previously-`undefined` fields could change what they render."* The rule stays true as suites are added; the roster won't.

`fixtures/skills.ts` has the same shape, so this reads as a nit on the convention rather than on this PR specifically — fine to fix here, in a sweep, or not at all.

---

ginderjeremiah (commented):
## Review: `IItem` test-fixture consolidation (#2433)

**Verdict:** Approve (posted as a comment — GitHub won't take a gating verdict from the PR author) — the scope call is right, every excluded group checks out, and the one default that changed is genuinely inert.

I didn't take the description's target-set claim on faith; I re-derived it from the tree. It holds up:

- **The seven converted modules are exactly the typed-`IItem` set.** Every remaining `): IItem =>` / `: IItem =` annotation under `src/tests/` is either the new fixture, an empty `IItem[] = []` registry (`battle-simulation-parity.test.ts:10`), or one of the deliberately-partial `as IItem` casts.
- **The exclusions are all correct**, including the two the description doesn't name: `workbench/entities/class.test.ts:99-102` and `welcome-back/WelcomeBackGate.test.ts:14` seed partials into a mocked `staticData`, and `entities/item.test.ts` spreads `itemEntity.newItem(...)` into a `WorkbenchItem` — none of them type-check as the contract.
- **The `requiredProficiencyLevel: 1` → `0` reconciliation in `references.test.ts` is safe.** `requiredProficiencyLevel` appears nowhere in either `references.ts` or `references.test.ts` after the change, so nothing could have read it. Every other converted suite keeps its old effective values, and the two behaviour-carrying divergences (the 5-Strength weapon, the Component mod slot) are stated at their adapters with a line saying what reads them — which is exactly what the issue asked for.
- **`makeItem` covers all ten required fields** of `IItem` and leaves all four optionals unset; `...overrides` stays last everywhere, so the adapters' override semantics are unchanged. No dead imports left behind — `ERarity`/`EItemCategory`/`EAttribute`/`EItemModType` still have live uses in every touched file.

### Nits

- `UI/src/tests/fixtures/items.ts:9-16` — the module comment's file roster reads as exhaustive but misses four excluded modules. Inline, with a suggested rewording. Non-blocking.

### Test coverage

Test-only change; no production code, so nothing new needs covering. The right check for this class of change is the contract-drift probe rather than added assertions, and the description records it: re-adding a required field to `IItem` afterwards leaves exactly one test-file error (the fixture). Converting suite-by-suite and running each in between is the right method for catching a silently-weakened assertion — worth calling out that it was actually done rather than asserted.

No frontend guideline concerns: `docs/frontend.md` already documents this convention generically (`IAttribute`, `IProficiency`, `ISkill`, …), so `items.ts` slots in with no doc update needed. No battle-parity implication — `battle-sim-test-utils.ts` only changed how its registry literal is spelled, and the values are identical, so the mirrored backend scenarios are untouched.

### Follow-up

- **#2444** — `IItemMod` has the same duplication class (~10 typed constructors) and wasn't in #2425's survey, so it never got split out. Five of the files converted here still hand-roll a local mod builder directly beside their new `makeItem` call. Out of scope for this PR, which is explicitly the `IItem` half.

### What's good

Correcting the issue's own ~20-suite estimate down to seven, with the reason each group was dropped, is the most valuable part of this PR — it's what stops the next `grep` from reading the leftovers as unfinished work. Reconciling the one drifted default *toward* the shared value after proving no assertion reads it, rather than preserving it defensively at the adapter, is also the right call: preserving it would have made the fixture's defaults meaningless.

---

UI/src/tests/fixtures/items.ts
ginderjeremiah: Fixed in `ab397ac` — checked all four omissions first and they're real: `skill-provenance.test.ts:10-11` asserts `({ id, name, grantedSkillId, retiredAt }) as IItem`, and `entities/class.test.ts:99-102` / `entities/item.test.ts` / `WelcomeBackGate.test.ts:14` all seed bare partials into a mocked `staticData`. So the roster was wrong on both axes, not just theoretically stale.

Took the suggested form: one example per group, rule stated rather than enumerated. It also drops the comment from 13 lines to 8, which is the more honest size for what it's actually saying.

Left `fixtures/skills.ts` alone — same shape, but rewriting it here would put an unrelated file in an `IItem` PR. Worth a sweep alongside `fixtures/proficiencies.ts` (which has the same enumerate-the-neighbours shape) if the convention is going to settle on rule-over-roster.

---

Issue #2444, opened 2026-07-25
Consolidate the hand-rolled IItemMod test builders onto a shared fixture


ginderjeremiah:
Found while reviewing #2443 (the `IItem` half of #2425). `IItemMod` has the same duplication class as `ISkill` / `IItem` / `IZone`, but it wasn't in #2425's survey — so it never got split out.

## Problem

~10 suites construct a **typed** `IItemMod` inline, several of them in the very files #2443 just converted for `IItem` (`item.test.ts`, `inventory-manager.test.ts`, `inventory-manager-reactivity.svelte.test.ts`, `references.test.ts`, `challenges-view.svelte.test.ts` each still hand-roll a local `makeItemMod`/`itemMod`/`mod` builder next to their new `makeItem` call).

Typed constructors (these pay the contract-drift tax):

| Suite | Local builder |
|---|---|
| `lib/engine/player/inventory-manager.test.ts` | `makeItemMod` |
| `lib/engine/player/inventory-manager-reactivity.svelte.test.ts` | `makeItemMod` |
| `lib/battle/item.test.ts` | inline `mockItemMods[7]` literal |
| `routes/admin/workbench/references.test.ts` | `itemMod` |
| `routes/admin/workbench/entities/item-mod.test.ts` | six inline `IItemMod` literals |
| `routes/game/screens/challenges/challenges-view.svelte.test.ts` | `mod` |
| `routes/game/screens/challenges/ModTooltip.test.ts` | `makeMod` |
| `routes/game/screens/challenges/RewardTooltip.test.ts` | `previewMod` |
| `routes/game/screens/challenges/RewardAffordance.test.ts` | `sampleMod` |
| `routes/game/screens/challenges/OverviewPane.test.ts` | `sampleMod` (partial `as IItemMod` — likely excluded) |

## Suggested direction

Same shape as #2414 / #2425 / #2433, which have now run three times cleanly:

- Add `makeItemMod(o)` to `UI/src/tests/fixtures/` following the existing convention. Decide whether it belongs in a new `item-mods.ts` or alongside `makeItem` in `items.ts` — they're separate contracts, so a separate module is probably right, but the suites almost always need both.
- Confirm the real target set with the throwaway-required-field probe (`svelte-check`), not the `grep` — #2433 showed the grep both under- and over-counts.
- Reconcile drifted defaults explicitly, preferring whichever variant has a live assertion; leave the partial/`staticData`-mock literals alone.
[...]

PR #2445, opened 2026-07-25
Consolidate the hand-rolled IZone test builders onto a shared fixture (#2434)

### LINKED ISSUE #2434: Consolidate the hand-rolled IZone test builders onto a shared fixture

ginderjeremiah:
The `IZone` half of #2425, following #2414 (`IAttribute`/`IPath`/`IProficiency`), #2425 (`ISkill`) and #2433/#2443 (`IItem`).

## Summary

Adds `UI/src/tests/fixtures/zones.ts` exporting `makeZone(overrides)` and converts the nine suites that construct a **typed** `IZone` onto it. Each suite keeps its local call signature as a thin adapter, so no call site churns.

Test-only — no production code, no behaviour change. 119 lines of duplicated contract literal out, 27 in.

## The probe corrected the issue's survey, in both directions

The issue estimated *"~16 suites (`grep -rl "bossLevel:" UI/src/tests/`)"* and flagged that number as untrustworthy. It was — the real typed set is **9**, and the grep was wrong both ways, exactly as #2433 predicted:

- **Over-counts.** Several hits seed zone literals into a `staticData: {} as any` mock. Nothing type-checks them as `IZone`, so they pay no contract-drift tax, and filling their previously-`undefined` fields could change what the suite renders. Left alone.
- **Under-counts.** `boss-view.test.ts`'s `bosslessZone` never sets `bossLevel` at all, so the grep misses it — but it's a typed `IZone` literal and does pay the tax.

Converted (from the `svelte-check` error list, not the grep):

| Suite | Local builder |
|---|---|
| `lib/common/zone-progression.test.ts` | `zone(id, order, unlockChallengeId?, retiredAt?)` |
| `lib/engine/battle/enemy-manager-boss.test.ts` | `zone(id, order, unlockChallengeId?)` |
| `routes/admin/workbench/references.test.ts` | `zone(id, name, opts)` |
| `routes/admin/workbench/entities/zone.test.ts` | `newZone` (a `WorkbenchZone`) |
| `routes/game/screens/fight/Fight.test.ts` | `makeZone(over)` |
| `routes/game/screens/fight/ZoneNav.test.ts` | `zone(id, order, name, …)` |
| `routes/game/screens/fight/boss/boss-view.test.ts` | two inline literals |
| `routes/game/screens/skills/Skills.test.ts` | inline `ZONES[0]` literal |
| `routes/game/screens/skills/skills-view.test.ts` | inline `ZONES[0]` literal |

## Defaults, reconciled explicitly

- **Unanimous across all nine:** `description: ''`, `designerNotes: ''`, `isHome: false`.
- **Majority, and asserted nowhere that doesn't restate them:** `levelMin: 1`, `levelMax: 10`, `bossLevel: 1`. The three divergent builders (`boss-view`'s `bossZone`, the two Skills suites' `Vale`) all assert on their level range or boss level, so they state it as an override — the divergence stays visible at the call site rather than being absorbed into a default.
- **`order` defaults to 0, not the id.** A suite proving order-driven behaviour is expected to place each zone explicitly rather than inherit a position from the Id-as-index invariant — the same call `makeProficiency` makes for `pathOrdinal`. `ZoneNav`'s "deliberately out of array order" fixture depends on that being explicit.
- **`bossEnemyId` / `unlockChallengeId` / `retiredAt` stay unset.** Each flips a distinct branch (the boss pill, the unlock gate, retirement filtering), so a live, boss-less, ungated zone is the neutral case.

`name` defaults to `Zone {id}`; the suites that assert on authored names (`Verdant Hollow`, `Vale`, …) pass one.

### One note on the workbench adapter

`entities/zone.test.ts`'s `newZone` builds a `WorkbenchZone` whose `bossEnemyId`/`unlockChallengeId` are the picker's **-1 "None" sentinel** — not neutral defaults, but the exact input the mapping under test converts back to an absent FK. It composes `makeZone` and states both sentinels plus `zoneEnemies` as its divergence, with a comment tying it to `zoneEntity.newItem`. This is the `progression-test-utils` pattern from #2426.

The production `zoneEntity.newItem` itself still hand-rolls the literal, which is correct — it's the authoritative blank record an admin gets, not a test builder, and it's one of the two remaining probe errors below.

## Test plan

- [x] **The issue's stated verification passes.** Adding a throwaway required field to `IZone` produced 12 errors across 10 files before, and after the change produces exactly **one** test-file error — `src/tests/fixtures/zones.ts`. The only other two are production (`routes/admin/workbench/entities/zone.ts`'s `newItem` and its persist mapping), which are out of scope.
- [x] Converted one suite at a time, running each in between, so a changed default would have failed loudly rather than quietly weakening an assertion. None did.
- [x] `npm run lint` (eslint + `svelte-check`): clean — 1226 files, 0 errors, 0 warnings.
- [x] `npm run test:unit:ci`: **4035/4035 passing across 1157 suites.**
- No backend files touched, so CI's changed-path gate skips the backend jobs.

## Follow-up filed

`IEnemy` has the identical duplication class and the same size — the probe flushes out **9** typed suites, six of them the very files this PR just converted for `IZone`. It wasn't in #2425's survey either, so like #2444 (`IItemMod`) it's filed separately rather than widened into this PR: **#2446**.

Closes #2434.

### DIFF
--- a/UI/src/tests/fixtures/zones.ts
+++ b/UI/src/tests/fixtures/zones.ts
@@ -1,0 +1,33 @@
+import type { IZone } from '$lib/api';
+
+/* Shared `IZone` contract builder (#2434), following the `fixtures/skills.ts` / `fixtures/items.ts`
+   convention.
+
+   The target set is much narrower than the `grep "bossLevel:"` the issue proposed: only a suite that
+   constructs a *typed* `IZone` pays the contract-drift tax, and a throwaway required field on the
+   contract flushes out exactly nine suites plus the workbench's production `newItem`. Suites that
+   seed zone-shaped literals into an untyped `staticData` mock are excluded on purpose — they see no
+   new required field, and filling their previously-`undefined` fields could change what they render.
+   A `grep` still finding zone literals under `src/tests/` is expected, not a gap. */
+
+/**
+ * Builds an {@link IZone} reference-data entry for tests. Everything but the id is a neutral
+ * placeholder so a suite states only what it asserts on.
+ *
+ * `order` defaults to 0 rather than the id — a suite proving order-driven behaviour is expected to
+ * place each zone explicitly rather than inherit a position from the Id-as-index invariant. The
+ * optional FKs (`bossEnemyId`, `unlockChallengeId`) and `retiredAt` are left unset for the same
+ * reason: a live, boss-less, ungated zone is the neutral case, and each of those fields flips a
+ * distinct branch (the boss pill, the unlock gate, retirement filtering) when present.
+ */
+export const makeZone = (overrides: Partial<IZone> & { id: number }): IZone => ({
+	name: `Zone ${overrides.id}`,
+	description: '',
+	designerNotes: '',
+	order: 0,
+	levelMin: 1,
+	levelMax: 10,
+	bossLevel: 1,
+	isHome: false,
+	...overrides
+});
--- a/UI/src/tests/fixtures/zones.ts
+++ b/UI/src/tests/fixtures/zones.ts
@@ -8,7 +8,10 @@ import type { IZone } from '$lib/api';
    contract flushes out exactly nine suites plus the workbench's production `newItem`. Suites that
    seed zone-shaped literals into an untyped `staticData` mock are excluded on purpose — they see no
    new required field, and filling their previously-`undefined` fields could change what they render.
-   A `grep` still finding zone literals under `src/tests/` is expected, not a gap. */
+   So are the few that assert a deliberately partial literal `as IZone` (`challenge-unlocks`,
+   `ChallengeTooltip`, `enemy-level`) — the cast opts them out of drift detection too, so they raise
+   no error either (#2447). A `grep` still finding zone literals under `src/tests/` is expected, not
+   a gap. */
 
 /**
  * Builds an {@link IZone} reference-data entry for tests. Everything but the id is a neutral
--- a/UI/src/tests/lib/common/zone-progression.test.ts
+++ b/UI/src/tests/lib/common/zone-progression.test.ts
@@ -1,20 +1,10 @@
 import { describe, it, expect } from 'vitest';
 import type { IZone } from '$lib/api';
 import { isZoneUnlocked, navigableZones, nextZoneByOrder, zonesByOrder } from '$lib/common/zone-progression';
+import { makeZone } from '../../fixtures/zones';
 
-const zone = (id: number, order: number, unlockChallengeId?: number, retiredAt?: string): IZone => ({
-	id,
-	name: `Zone ${id}`,
-	description: '',
-	designerNotes: '',
-	order,
-	levelMin: 1,
-	levelMax: 10,
-	bossLevel: 1,
-	unlockChallengeId,
-	isHome: false,
-	retiredAt
-});
+const zone = (id: number, order: number, unlockChallengeId?: number, retiredAt?: string): IZone =>
+	makeZone({ id, order, unlockChallengeId, retiredAt });
 
 describe('zonesByOrder', () => {
 	it('sorts by authored order without mutating the input', () => {
--- a/UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts
+++ b/UI/src/tests/lib/engine/battle/enemy-manager-boss.test.ts
@@ -3,6 +3,7 @@ import { apiSocket, ELogType, type IApiSocketResponse, type IEnemyInstance, type
 import { delay } from '$lib/common';
 import { logMessage } from '$lib/engine/log';
 import { EnemyManager, onNewEnemyLoaded } from '$lib/engine/battle/enemy-manager';
+import { makeZone } from '../../../fixtures/zones';
 
 // Share a holder so the mocked onBattleStageChanged can hand the registered stage callback back
 // to the test, letting it drive Victorious/Defeated transitions deterministically.
@@ -178,18 +179,8 @@ describe('EnemyManager boss mode', () => {
 		return release;
 	};
 
-	const zone = (id: number, order: number, unlockChallengeId?: number): IZone => ({
-		id,
-		name: `Zone ${id}`,
-		description: '',
-		designerNotes: '',
-		order,
-		levelMin: 1,
-		levelMax: 10,
-		bossLevel: 1,
-		unlockChallengeId,
-		isHome: false
-	});
+	const zone = (id: number, order: number, unlockChallengeId?: number): IZone =>
+		makeZone({ id, order, unlockChallengeId });
 
 	// bossUnlockedNextZone is reset by returnToIdle/challengeBoss once the (mocked-immediate) overlay
 	// delay elapses, so capture it at the overlay moment — the instant `delay` is called, right after
--- a/UI/src/tests/routes/admin/workbench/entities/zone.test.ts
+++ b/UI/src/tests/routes/admin/workbench/entities/zone.test.ts
@@ -37,6 +37,7 @@ vi.mock('$lib/api', async (importOriginal) => {
 });
 
 import { zoneEntity, type WorkbenchZone } from '$routes/admin/workbench/entities/zone';
+import { makeZone } from '../../../../fixtures/zones';
 
 /** Finds the body posted to a given AdminTools endpoint (or undefined if never called). */
 const postBodyTo = (endpoint: string) => mockPost.mock.calls.find((c) => c[0] === endpoint)?.[1];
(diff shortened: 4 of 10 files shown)


UI/src/tests/fixtures/zones.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/* Shared `IZone` contract builder (#2434), following the `fixtures/skills.ts` / `fixtures/items.ts`
+   convention.
+
+   The target set is much narrower than the `grep "bossLevel:"` the issue proposed: only a suite that
+   constructs a *typed* `IZone` pays the contract-drift tax, and a throwaway required field on the
+   contract flushes out exactly nine suites plus the workbench's production `newItem`. Suites that
+   seed zone-shaped literals into an untyped `staticData` mock are excluded on purpose — they see no
+   new required field, and filling their previously-`undefined` fields could change what they render.
+   A `grep` still finding zone literals under `src/tests/` is expected, not a gap. */
```
ginderjeremiah: **Nit** — the exclusion note names only the untyped-`staticData`-mock class, but that isn't the whole leftover set. Three suites annotate `IZone` and escape the probe via a cast, not via an untyped mock:

- `lib/common/challenge-unlocks.test.ts:5-16` — `const zone = (…): IZone => ({ … }) as IZone`
- `components/tooltip/ChallengeTooltip.test.ts:23-24` — same shape
- `routes/game/screens/codex/enemy-level.test.ts:5-9` — `[…] as IZone[]`

Excluding them is defensible (they produce no `svelte-check` error, and #2443 set the precedent for `IItem`), but as written the comment tells the next reader that everything still matching a zone grep is an untyped mock — and then that reader opens `challenge-unlocks.test.ts`, sees `: IZone`, and can't reconcile it. `fixtures/items.ts:6-11` gets this right by naming all three excluded classes, cast literals included.

One clause is enough, e.g. after "…are excluded on purpose": *"as are the few that assert a deliberately partial literal `as IZone` — the cast means they see no new required field either."*

Filed #2447 for the substantive half of this (those casts opt out of drift detection permanently, across every contract, not just `IZone`) — this comment is only about the fixture's own description of what it left behind.

---

ginderjeremiah (commented):
## Review: consolidate the nine hand-rolled `IZone` test builders onto `fixtures/zones.ts` (#2434)

**Verdict: Approve** — the conversion is faithful, the defaults are reconciled the way the issue asked, and the stated verification holds. One nit and one follow-up, neither blocking. (Submitted as a `COMMENT` because GitHub won't take a gating verdict from the PR author.)

I checked this the way the issue framed the risk — the danger in a fixture consolidation isn't the fixture, it's a converted call site whose defaults silently shift and quietly weaken an assertion. So I reconstructed the effective object each converted builder produces, before and after, rather than reading the diff for plausibility:

| Suite | Before → after |
|---|---|
| `zone-progression` | identical (all four params forwarded) |
| `enemy-manager-boss` | identical |
| `references` | identical (`order: 0`, `1`/`10`/`1` all defaults) |
| `entities/zone` | identical (`-1` sentinels + `zoneEnemies` stated) |
| `Fight` | identical (`id: 0`, name, `order: 1` stated) |
| `ZoneNav` | identical (six params forwarded) |
| `boss-view` | identical — `bossZone` states `8`/`11`/`18`, `bosslessZone` takes `1`/`10`/`1` |
| `Skills` / `skills-view` | identical (`2`/`8`, boss `2`@`10` stated) |

No converted suite uses `toStrictEqual`, so the adapters that forward `unlockChallengeId`/`retiredAt` as explicit `undefined` (`zone-progression`, `ZoneNav`) don't change any comparison — and they did that before the change too.

The three divergent-default calls are the ones that matter, and each states its divergence at the call site rather than absorbing it into the fixture. `boss-view`'s `bossZone` asserts on its level range, and both Skills suites assert Toughness derived from `[2,8]` and boss level `10` — folding any of those into `makeZone` is exactly how this refactor could have gone wrong. It didn't.

### Nit

- `fixtures/zones.ts:8-11` — the exclusion note only names the untyped-`staticData`-mock class, but three of the leftover suites are annotated `IZone` and escape via an `as IZone` cast instead. `fixtures/items.ts:6-11` names all three excluded classes; this one should match. One clause. Inline comment has the wording.

### Follow-up filed

- **#2447** — the substantive half of that nit. `challenge-unlocks`, `ChallengeTooltip`, `enemy-level` (plus `IEnemy`, `IItem`, `IChallenge`, `IChallengeType`, `IAttribute` sites) build partial literals and cast them to the contract, so a new required field never errors there. That means the series' stated guarantee — "adding a required field produces exactly one test-file error" — is true of the probe but slightly overstates the coverage. Cross-contract and the same decision each time, so it's one pass, not this PR's job.
- **#2446** (`IEnemy`) was filed from this work and is accurate — I confirmed six of its nine suites are files this PR just touched, which is a good argument for taking it next while the context is warm.

### Test coverage

Test-only change, no production code, so there's nothing new to cover. The relevant bar is that the existing coverage didn't weaken, which the table above is the check on. The issue's verification — one test-file error after adding a required field — is met; the two remaining errors are `entities/zone.ts`'s production `newItem` and its persist mapping, and leaving those as hand-rolled literals is right. `newItem` is the authoritative blank record an admin gets, not a test builder, and pointing production at a `src/tests/` module to save eleven lines would be a genuinely bad trade.

### What's good

- Verifying the target set with the probe instead of the issue's `grep` — and reporting that it corrected the survey in *both* directions, including the `bossLevel`-less `boss-view` literal the grep structurally can't find. That's the second time this series has caught the issue author's own estimate being wrong; worth keeping in the next one.
- `entities/zone.test.ts:114-121` composes the fixture and then states the two `-1` sentinels as its divergence, with a comment tying it back to `zoneEntity.newItem`. The sentinels are the input under test, so burying them in a default would have hidden the point of the test.
- The `makeZoneContract` import alias in `Fight.test.ts` is the right call over renaming the local builder — it keeps the no-call-site-churn property that makes this diff reviewable at all (119 lines out, 27 in, zero behaviour change).

### Merge status

Holding the merge on the nit and on `test-e2e (webkit)`, which is still running; everything else is green (`verify-frontend`, `test-frontend`, chromium + firefox e2e), with the backend jobs correctly skipped by the changed-path gate. Land the one-clause comment fix and this is ready.

---

UI/src/tests/fixtures/zones.ts
ginderjeremiah: Fixed in b8e4730. Verified all three before editing — each omits `designerNotes` (and `enemy-level`/`ChallengeTooltip` omit `isHome` too), so they're genuinely partial literals riding the cast, not untyped mocks.

Took your wording with one addition — a pointer to #2447, since the reason they escape is the same reason that issue exists:

> So are the few that assert a deliberately partial literal `as IZone` (`challenge-unlocks`, `ChallengeTooltip`, `enemy-level`) — the cast opts them out of drift detection too, so they raise no error either (#2447).

Your point about the guarantee being overstated is the part worth carrying forward: "adding a required field produces exactly one test-file error" is a claim about the probe, not about coverage, and a cast literal is invisible to both. #2446 (`IEnemy`) will hit this immediately — `enemy-level.test.ts` and `challenge-unlocks.test.ts` cast `IEnemy` and `IItem` in the same breath — so whichever of #2446/#2447 lands second should state the guarantee as "one error plus the known cast sites" rather than repeating it unqualified.

---

PR #2448, opened 2026-07-26
Drop the pure WorkbenchPath/WorkbenchProficiency aliases (#2436)

### LINKED ISSUE #2436: WorkbenchPath/WorkbenchProficiency are pure aliases, unlike the other five Workbench&lt;Entity&gt; types

ginderjeremiah:
Closes #2436.

## Summary

`WorkbenchPath`/`WorkbenchProficiency` were plain aliases of `IPath`/`IProficiency`. This drops them and uses the contract types directly across the progression editor and its tests — 20 files, no behaviour change.

## The call I made

The issue framed this as a judgment call and named a real counter-argument: *an alias is a ready-made seam for the first workbench-only field, and three of the five siblings exist for exactly that reason.* I came down on removing them, for two reasons:

1. **The family role doesn't apply here.** The five `entities/*.ts` types exist to be the `EntityConfig` type parameter for the generic Workbench. The progression editor is a bespoke surface with its own controller (`ProgressionStore`) and isn't `EntityConfig`-driven at all, so these two never carried that role — the membership was nominal, and they don't even share the family's home (`progression/types.ts` vs `entities/*.ts`).
2. **The seam is free to re-cut.** Re-introducing an alias the day a workbench-only field appears is a one-line change the compiler points at exhaustively. Carrying it speculatively costs a reader on every encounter — the doc comments literally said *"Identical to the `IPath` contract"*, which is precisely the sentence that forces someone to go check whether that's still true. The #2426 fixture work had to do that check before it could reuse the shared builders.

I verified the issue's premise rather than taking it on faith: all five siblings do genuinely diverge (`WorkbenchZone`/`WorkbenchEnemy` add a derived field; `WorkbenchItem`/`WorkbenchClass`/`WorkbenchLesson` widen a field for the `-1` sentinel), and these two did not. `progression-store.svelte.ts` was already importing `IPath`/`IProficiency` from `$lib/api` *alongside* the aliases — the redundancy was direct.

## Changes

- **Aliases removed**, and `progression/types.ts` deleted with them. Once the aliases were gone the file held a single UI sentinel and no types at all, so `NO_SKILL` moved to `progression-helpers.ts` under the milestone section it serves. The issue's "`NO_SKILL` stays in `types.ts` regardless" line was written before the file was emptied — following it literally would have preserved a file that only made sense while it held the aliases. No cycle risk: helpers imports only `$lib/api`/`$lib/common`/`$stores`/`../entities/types`, never the store.
- **Prose relocated, not deleted.** The two genuinely useful sentences (how a path's ordered tiers are the proficiencies carrying its id; how a tier's level modifiers/rewards and prerequisites persist through dedicated relationship endpoints) now sit on `ProgressionStore.paths`/`.profs`, where they describe the editor's model rather than restating a type.
- **Convention pinned** in `docs/frontend-admin.md` — one bullet stating that a `Workbench<Entity>` type earns its name by diverging from its contract, so a pure alias doesn't get re-added later by someone reasoning from consistency. This felt worth documenting precisely because "be consistent with the family" is the intuition that created these two.
- **Test-fixture comment corrected** in `tests/fixtures/proficiencies.ts`, which described the workbench builders as building "plain aliases of these same two contracts" — now stale.
- **Import hygiene.** Merged the resulting duplicate `$lib/api` imports in the test files inline (`import { EActivityKey, type IPath, … }`), matching the style already in `tests/fixtures/proficiencies.ts`; and hoisted the `$lib/api` imports that had inherited the old `./types` slot above the relative block, per the workbench-wide alias-imports-first convention (`AttributePicker.svelte`, `entities/zone.ts`, `Workbench.svelte`).

## Test plan

No new tests: this is a type-identity change with no runtime surface, and the existing progression suites already cover the touched code — they now exercise it through the contract types.

- [x] `npm run lint` — clean (eslint + prettier). The shorter names let prettier re-collapse three wrapped signatures; applied via `--fix`.
- [x] `npm run check` (`svelte-check`) — **1225 files, 0 errors, 0 warnings**. This is the real verification for a rename of this shape: any missed reference is a compile error. The count dropping from 1226 is `types.ts` being gone.
- [x] `npm run test:unit:ci` — **312 files / 4035 tests passed.**
- [x] `grep` confirms zero remaining `WorkbenchPath`/`WorkbenchProficiency` references anywhere in `UI/src`, `docs/`, or `.claude/`.
- [x] CI green on the head commit: `verify-frontend`, `test-frontend`, and all three `test-e2e` browsers. The backend jobs are skipped by the changed-path gate, which is correct for a frontend-only diff.

## Note for review

The one thing worth a second opinion is the call itself — if the progression editor is expected to grow an optional-FK picker (an activity-key or reward-skill "none" sentinel would be the natural candidate), you may prefer to have kept the seam. Happy to revert that half; the doc bullet is written to permit re-introducing an alias for a real divergence.


UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
```diff
@@ -22,7 +22,7 @@ import {
 	renumberTiers,
 	tiersOfPath
 } from './progression-helpers';
-import { NO_SKILL, type WorkbenchPath, type WorkbenchProficiency } from './types';
+import { NO_SKILL } from './types';
```
ginderjeremiah: **Nit** — with the two aliases gone, `progression/types.ts` exports no types at all; it's a file named `types.ts` holding a single UI sentinel. The issue said "`NO_SKILL` stays in `types.ts` regardless", but that was written before the file was emptied, and CLAUDE.md is explicit that an issue's framing shouldn't be taken on faith — the leftover is a direct consequence of this diff, so it belongs in this PR rather than a follow-up.

Cheapest fix: move `NO_SKILL` into `progression-helpers.ts` and delete `types.ts`. Both of its consumers (this file and `MilestonesEditor.svelte`) already import from `progression-helpers`, so it's a net `-1` file and `-2` import statements, and there's no cycle risk — helpers doesn't import the store.

`constants.ts` would work too; the point is just that `types.ts` no longer describes its contents.

---

UI/src/routes/admin/workbench/progression/MilestonesEditor.svelte
```diff
@@ -108,14 +108,15 @@ import WorkbenchIcon from '../WorkbenchIcon.svelte';
 import { reference } from '../reference.svelte';
 import type { ProgressionStore } from './progression-store.svelte';
 import { cumulativeXp, isMilestoneLevel, modifiersAtLevel, rewardAtLevel } from './progression-helpers';
-import { NO_SKILL, type WorkbenchProficiency } from './types';
+import type { IProficiency } from '$lib/api';
```
ginderjeremiah: **Nit** — the `$lib/api` import inherited the slot the old `./types` import occupied, so it now sits *after* the relative imports. The surrounding convention is alias imports (`$components`/`$lib`/`$stores`) first, then relative — see `components/AttributePicker.svelte:91-98`, `entities/zone.ts:1-6`, `Workbench.svelte:39-49`. Nothing lints it, which is why it slipped through.

Same in `GatewaysEditor.svelte:51`, `TiersSpine.svelte:49`, `XpCurve.svelte:44`, `TierDetail.svelte:83`, and `progression-map.ts:2`. `ConlangIdentity.svelte:119` and `PathDetail.svelte:82` read slightly worse still — they end up with `$lib/api/types/game-constants` at the top and plain `$lib/api` several lines below the relative block.

Purely cosmetic and easy to fold into a follow-up if you'd rather not re-touch nine files.

---

ginderjeremiah (commented):
## Review: drop the pure `WorkbenchPath`/`WorkbenchProficiency` aliases (#2436)

**Verdict:** **Approve** (posted as a `COMMENT` — GitHub won't accept a gating verdict from the PR's own author). No blocking findings; two nits, neither of which needs to hold the change.

The call is right, and for the reason the PR gives rather than the one the issue leads with. The five `entities/*.ts` types exist to be the `EntityConfig` type parameter for the generic Workbench; the progression editor is a bespoke surface driven by `ProgressionStore` and isn't `EntityConfig`-shaped at all, so these two never carried the family's role — the membership really was nominal. I verified the premise independently rather than trusting the issue's table: `WorkbenchZone`/`WorkbenchEnemy` `extends` with a derived field, `WorkbenchItem`/`WorkbenchClass`/`WorkbenchLesson` `Omit` + widen for the `-1` sentinel, and these two were plain `=`. The issue's survey holds.

The counter-argument (an alias is a free seam for the first workbench-only field) is real but weak here: re-cutting it is a one-line change the compiler then points at exhaustively, whereas a doc comment reading *"Identical to the `IPath` contract"* taxes every reader with a check — which is exactly what the #2426 fixture work had to do before it could reuse the shared builders.

### Nits

- `progression/types.ts` now exports zero types — see the inline comment. Folding `NO_SKILL` into `progression-helpers.ts` and deleting the file is a strict simplification; the issue's "`NO_SKILL` stays" line predates the file being emptied.
- Import ordering in the seven components + `progression-map.ts`: the `$lib/api` import kept `./types`' old slot, landing after the relative imports, against the workbench-wide convention. Cosmetic; fine as a follow-up if you'd rather not re-touch nine files.

### Test coverage

Appropriate as-is. This is a type-identity change with no runtime surface, so there's nothing new to assert — `svelte-check` **is** the verification, and any missed reference is a compile error rather than a silent behaviour change. The existing progression suites now exercise the same code through the contract types. `verify-frontend` is green on the head commit; the remaining jobs were still running at review time.

Worth noting the test-side changes go slightly beyond the mechanical rename and are better for it: merging the stranded `import type` statements into the existing `$lib/api` value imports (`import { EAttribute, EModifierType, type IProficiency }`) matches the style already in `tests/fixtures/proficiencies.ts` instead of leaving a second import ten lines below the first.

### What's good

- **The prose was relocated, not deleted.** The two genuinely load-bearing sentences now sit on `ProgressionStore.paths`/`.profs`, where they describe the editor's model rather than restating a type — which is the form in which they actually earn their keep.
- **The stale fixture comment got caught.** `tests/fixtures/proficiencies.ts` described the workbench builders as *"plain aliases of these same two contracts"*; that sentence would have outlived its subject by exactly one PR. Easy to miss in a rename this wide.
- **The doc bullet is the right kind of documentation.** Pinning "a `Workbench<Entity>` type must earn its name" in `frontend-admin.md` targets the precise intuition — *be consistent with the family* — that created these two in the first place, and it's written to permit re-introducing an alias for a real divergence rather than banning it outright. That's a convention, not an implementation detail, so it sits on the right side of CLAUDE.md's over-documentation line.

---

UI/src/routes/admin/workbench/progression/progression-store.svelte.ts
ginderjeremiah: Both nits fixed in `d48dfa3` — agreed on both, and agreed they belong in this PR rather than a follow-up, since each is a direct consequence of this diff rather than pre-existing.

**`types.ts` retired.** `NO_SKILL` moved into `progression-helpers.ts` under the `── Milestone (payout) projection ──` section it actually serves, and the file is deleted. Confirmed the reasoning before applying it: both consumers already imported from helpers, and helpers imports only `$lib/api`/`$lib/common`/`$stores`/`../entities/types` — nothing reaching back to the store — so there's no cycle. Net `-1` file, `-2` import statements. `svelte-check` now reports 1225 files where it reported 1226, which is the file being gone.

I took `progression-helpers.ts` over a new `constants.ts` because a second single-symbol module would just relocate the same problem.

**Import ordering fixed in all eight files** (the eight that actually carried the inherited slot — `types.ts` was the ninth and is now deleted). Verified the convention against all three references you cited before applying it; `progression-helpers.ts` itself already followed it. `ConlangIdentity`/`PathDetail` now lead with `$lib/api`, then the multi-line `$lib/api/types/game-constants`, then the relative block.

Re-verified after the change: eslint + prettier clean, `svelte-check` 1225 files / 0 errors / 0 warnings, 312 test files / 4035 tests passing.

---

Issue #1453, opened 2026-07-02
Allow skills to natively have a base crit chance


ginderjeremiah:
Missed in the implementation of #1425.

PR #2449, opened 2026-07-26
Share the crit chance and execute investment compositions with CombatRating (#2439)

### LINKED ISSUE #2439: CombatRating re-derives the crit chance product by hand — the same engine-vs-rating duplication #2429 closed for parry/dodge
Raised in the review of #2438 (which closed #2429 by promoting the parry/dodge products onto `Battler` so the engine's draws and `CombatRating` read one composition site).

## Problem

#2429 put avoidance under `CombatRating`'s anti-drift rule. **Crit is the remaining hole in that same rule**: `skill.CriticalChance × CriticalChanceMultiplier` is still spelled out independently at

- `BattleContext.ResolvePlayerHit` — the live crit draw (`critDraw < baseCriticalChance * _activeBattler.GetAttributeValue(CriticalChanceMultiplier)`);
- `CombatRating.ExpectedDirectHit` — `ProcChance(skill.CriticalChance * effectiveCaster.GetAttributeValue(CriticalChanceMultiplier))`.

(The frontend `battle-step.ts` `resolvePlayerHit` composes it too, but that one is the intended parity mirror, not the drift risk.)

This is the identical defect class #2429 describes, one mechanic over: if the crit composition ever gains a third factor, a floor, or a diminishing-returns curve, nothing forces the rating to follow, and the failure is silent — the rating keeps returning a plausible number that no longer prices what the engine does. Crit is arguably the *more* load-bearing of the two, since `critExpectation` multiplies the whole direct-hit expectation rather than contributing one additive term.

## Suggested direction

Same shape as #2429: one shared composition, read by both sides. The wrinkle is that crit's enabler is **per-skill** (`Skill.CriticalChance`, #1453) rather than an attribute, so it can't be a bare `Battler` property like `EffectiveParryChance`. Two candidates:

- `Battler.EffectiveCriticalChance(double baseCriticalChance)` — mirrors `AmplifyDamage`'s "attacker method taking the per-hit input" shape, and both call sites already hold the acting battler;
- `BattleSkill.EffectiveCriticalChance(Battler caster)` — reads better at the fire site, but `CombatRating` prices a raw `Skill`, not a `BattleSkill`, so it would need the enabler passed anyway.

The first is the closer fit. As in #2429 the **clamp stays rating-side** — `ProcChance` exists because the rating prices an expectation, while `draw < chance` against a `[0, 1)` draw saturates naturally; folding it into the shared member would change engine behaviour.

## Constraint

`Battler` is on the frontend/backend battle-parity surface, so this needs the mirrored member in `UI/src/lib/battle/battler.ts` (consumed by `resolvePlayerHit` in `battle-step.ts`) and the mirrored test scenarios on both sides, per CLAUDE.md. No behaviour change on either side — the products are byte-identical.

Worth checking while in there whether `ExecuteBonus × missingHpFraction` (composed in both `BattleContext.ResolvePlayerHit` and `CombatRating.ExpectedDirectHit`'s `executeExpectation`) deserves the same treatment, or whether the rating's use of a *reference* missing-health fraction makes it a genuinely different reading rather than a duplicated one.

## Scope

Small — one member plus its frontend mirror and mirrored tests, no behaviour change.

ginderjeremiah:
Closes #2439.

## Summary

`CombatRating`'s class doc-comment stakes its correctness on an anti-drift rule — it reuses the exact `Battler`/`BattleSkill` members the engine uses, so the rating can't silently diverge from the real damage pipeline. #2429 put the parry/dodge products under that rule; **crit was the remaining hole**, with `skill.CriticalChance × CriticalChanceMultiplier` spelled out independently at the engine's live draw and at the rating.

This promotes the composition onto `Battler` and routes both sites through it:

- **`Battler.EffectiveCriticalChance(double baseCriticalChance)`** — a *method* rather than a bare property like `EffectiveParryChance`, because crit's enabler is authored **per skill** (`Skill.CriticalChance`, #1453) rather than as an attribute. This is the first of the issue's two candidates, mirroring `AmplifyDamage`'s "attacker method taking the per-hit input" shape; both call sites already hold the acting battler.
- **The clamp stays rating-side**, as the issue specifies. `CombatRating` gains a `CritProcChance(skill, caster)` helper alongside the existing `ParryProcChance`/`DodgeProcChance`, since the engine's `draw < chance` against a `[0, 1)` draw saturates naturally and folding `ProcChance` into the shared member would change engine behaviour.

## The execute question the issue raised

The issue asked whether `ExecuteBonus × missingHpFraction` deserves the same treatment "or whether the rating's use of a *reference* missing-health fraction makes it a genuinely different reading". **It's both** — and the split falls cleanly:

- The **fraction** genuinely differs and should stay a caller input: the engine samples the live target's missing health once per fire (clamped), while the rating prices `ServerGameConstants.RefMissingHealthFraction`.
- The **composition** is duplicated at *three* sites, not two — the engine uses `ExecuteBonus × fraction` for the damage multiplier **and** for the Cull overlay tally (`OverlayTally.NormalizeInvestment(executeInvestment)`), and the rating composes it again for `executeExpectation`.

So `Battler.ExecuteInvestment(double missingHpFraction)` takes the fraction as input and owns the product. That keeps the engine's separate need for the raw investment (the tally books on it, not on the `1 +` multiplier) intact.

## Parity

`Battler` is on the frontend/backend parity surface, so both members are mirrored in `UI/src/lib/battle/battler.ts` (`effectiveCriticalChance` / `executeInvestment`) and read by `resolvePlayerHit` in `battle-step.ts`. Per CLAUDE.md the new test scenarios are asserted with the **same inputs and the same expected results** on both sides.

**No behaviour change on either side** — the products are arithmetically identical, `a * b` factored into a member. The existing parity matrix (`battle-engine-parity.test.ts` ⇄ the backend simulator parity tests) passes unchanged, which is the strongest evidence of that.

## Test plan

New mirrored unit tests in `Game.Core.Tests/Battle/BattlerTests.cs` ⇄ `UI/src/tests/lib/battle/battler.test.ts`, following the shape #2429 established for the avoidance pair:

- **`EffectiveCriticalChance`** — unauthored skill crits at exactly 0 regardless of Luck (the `0 × mult = 0` opt-in); authored chance scales by the Luck-amplified multiplier; passes through verbatim at zero Luck; a product above 1 is **not** clamped; read live, so a mid-battle `CriticalChanceMultiplier` buff lands on the next fire.
- **`ExecuteInvestment`** — an un-enabled build invests 0 at any missing health; an authored bonus scales by the fraction; a full-health target invests 0.

Verification run locally:

- [x] `dotnet build` — succeeded, **0 warnings, 0 errors**
- [x] `dotnet test --no-build --no-restore` — **3,522 passed, 0 failed** (Core 1469, Application 1105, Api 874, Infrastructure 74)
- [x] `npm run lint` (eslint + `svelte-check`) — clean, **0 errors 0 warnings** across 1,225 files
- [x] `npm run test:unit:ci` — **4,043 passed** across 312 files

## Follow-up

While in `CombatRating` I found one more instance of this defect class, filed separately rather than scope-creeping this PR: `Survivability` re-derives the Toughness mitigation curve (`toughness / (toughness + ToughnessMitigationConstant)`) as a byte-identical copy of the expression inside `Battler.ComputeNetDamage`. The existing comment justifies *not calling `ComputeNetDamage`* (survivability needs the dimensionless fraction, not a transformed damage amount) — which is correct — but doesn't address extracting the curve itself so both read one site. Details in the follow-up issue.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D3xcWXK3AAnJCjguDWgS7q)_


UI/src/lib/battle/battler.ts
```diff
@@ -175,6 +175,29 @@ export class Battler {
...
+	 *  `baseCriticalChance × CriticalChanceMultiplier`, the product the engine draws against (see
+	 *  `resolvePlayerHit` in `battle-step`). Unlike the avoidance pair, crit's enabler is authored PER SKILL
+	 *  (`Skill.criticalChance`, #1453) rather than as an attribute, so it arrives as a parameter. Composed
+	 *  here at consumption so the live draw and the backend combat rating's pricing of the same capability
+	 *  cannot derive it differently (#2439). A skill's `criticalChance` defaults to 0 (an authored-only
+	 *  opt-in), so an un-authored skill never crits regardless of the multiplier. Deliberately UNCLAMPED, for
+	 *  {@link effectiveParryChance}'s reason. Mirrors the backend `Battler.EffectiveCriticalChance`. */
+	public effectiveCriticalChance(baseCriticalChance: number): number {
+		return baseCriticalChance * this.attributes.getValue(EAttribute.CriticalChanceMultiplier);
+	}
```
ginderjeremiah, line 197: **Should-fix** — the crit composition is now single-sited for the *engine* and the *rating*, but two frontend **display** sites still hand-derive the identical product and neither was routed through this member:

- `UI/src/routes/game/screens/fight/SkillTooltip.svelte:94` — `(skill?.criticalChance ?? 0) * (skill?.owner.attributes.getValue(EAttribute.CriticalChanceMultiplier) ?? 1)`
- `UI/src/routes/game/screens/skills/skills-view.svelte.ts:250` — `skill.criticalChance * critChanceMultiplier`

Both carry a comment explaining the same "per-skill enabler × multiplier" rule this member now owns, which is the tell. The failure mode is exactly the one #2439 describes, one consumer over: add a third factor or a floor here and the tooltip/skills grid keep showing a number the sim no longer produces — silently, since there's nothing forcing them to follow.

The tooltip is a one-liner (`skill.owner.effectiveCriticalChance(skill.criticalChance)`), but `skills-view` only holds a `BattleAttributes`, never a `Battler`, so a uniform fix wants the shape this codebase already uses for exactly that split: a free function in `battle-formulas.ts` over `BattleAttributes`, with the `Battler` member delegating — the same arrangement as `cooldownMultiplier` ⇄ `Battler.cdMultiplier`. `scaledEffectAmount` in that file even documents the property being aimed for: *"the single definition consumed by both the battle runtime and the skill tooltip."*

Suggesting it lands here rather than as a follow-up: it's the same defect class the PR is closing, it's ~20 lines plus a formulas test, and splitting it leaves the codebase in the odd state of having a canonical composition that a third of its consumers ignore. If it does get deferred, the follow-up should be filed before merge.

(No equivalent gap for `executeInvestment` — `ExecuteBonus` has no display consumer outside the sim.)

---

Game.Core/Battle/Battler.cs
```diff
@@ -97,6 +97,40 @@ public double GetCooldownMultiplier()
...
+        /// Only the <i>fraction</i> legitimately differs between them — the engine samples the live target's
+        /// missing health once per fire, the rating prices a fixed reference profile — which is why it is the
+        /// caller's input rather than something read here. <see cref="EAttribute.ExecuteBonus"/> is an
+        /// authored-only enabler idling at <c>0</c>, so an un-enabled build invests exactly <c>0</c> and its
+        /// multiplier is an exact <c>1.0</c>.
+        /// </summary>
+        public double ExecuteInvestment(double missingHpFraction)
+        {
+            return _attributes[ExecuteBonus] * missingHpFraction;
+        }
```
ginderjeremiah, line 139: **Nit** — `missingHpFraction` carries an unstated `[0, 1]` contract. Both current callers honour it (the engine `Math.Clamp`s at the sample site, the rating passes a constant), but the parameter name is the only thing saying so, and a fraction above `1` (a MaxHealth debuff landing between sample and use, say) would silently over-invest — and, because the Cull overlay tally books on this value, over-credit training too.

Worth a clause in the doc-comment making the contract explicit: the caller owns the clamp, and why (the engine clamps a live sample, the rating passes a fixed reference). Deliberately not suggesting the clamp move in here — that would be a behaviour change the engine's `Math.Clamp` already covers.

---

ginderjeremiah (commented):
## Review: promoting the crit-chance and execute-investment compositions onto `Battler` so the engine and `CombatRating` share one site (#2439)

**Verdict:** Approve, with two should-fixes — *(submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR's own author; the intended verdict is **approve**.)* The refactor is correct, the parity discipline is honoured, and nothing here changes behaviour. Neither should-fix is a defect in what's written — both are about the change stopping slightly short of the anti-drift goal it states.

I verified the central claim rather than taking the description's word for it. Every backend crit composition now routes through `Battler.EffectiveCriticalChance`: the deliberate fire (`BattleContext.cs:306`), the parry riposte (`BattleContext.cs:447`, passing `counterSkill.CriticalChance` into the same `ResolvePlayerHit`), and the rating (`CombatRating.cs:92`). A grep across `Game.Core`/`Game.Application`/`Game.Api`/`Game.DataAccess` turns up no fourth site. Same on the frontend engine side — `battle-step.ts:53` is the only draw, and the riposte at `:231` reuses `resolvePlayerHit`, so it's covered for free.

### Should-fix

- **Two frontend display sites still hand-derive the crit product** — `SkillTooltip.svelte:94` and `skills-view.svelte.ts:250`. Detail and a suggested shape inline on `UI/src/lib/battle/battler.ts:186`. This is the same defect class the PR closes, just with a display consumer instead of the rating.
- **`docs/backend-battle.md:43` now understates the rule this PR just widened.** That bullet reads *"Each **avoidance** product is composed at one site per simulator … The server-only `CombatRating` prices the same capability through those accessors rather than re-deriving it, but clamps to `[0,1]` itself (#2429)"* — while the crit clause earlier in the same bullet still spells out `c < skill.CriticalChance × CriticalChanceMultiplier` as an independent derivation. After this PR that one-site rule covers crit and the execute investment too. #2438 updated exactly this line when it closed #2429, so the precedent for documenting it here is explicit, and CLAUDE.md asks for the docs to track the design as it moves. A one-sentence edit.

### Nits

- `ExecuteInvestment`'s `missingHpFraction` has an unstated `[0, 1]` contract — inline on `Game.Core/Battle/Battler.cs:129`.

### Test coverage

Solid, and it follows the project's rules where they're strictest. The eight new backend cases in `BattlerTests.cs` are mirrored 1:1 in `battler.test.ts` — same scenarios, same inputs, same expected values, per CLAUDE.md's battle-parity requirement — right down to reusing effect id `2` on both sides so the two files read as one suite. The chosen cases are the right ones: the `0 × mult = 0` opt-in (the property the whole authored-only enabler design rests on), pass-through at base multiplier, the deliberate non-clamp above `1`, live re-read after a mid-battle buff, and — for execute — the un-enabled and full-health zeroes that pin the exact-`1.0` multiplier.

Two things I checked and am satisfied with rather than flagging: there's no new test asserting the engine and the rating *agree*, but that assertion is now structural (they call one member), which is the point; and no new `CombatRating` test is warranted since the products are arithmetically unchanged and the existing parity matrix passing is the real evidence of that.

### What's good

- **The execute question in the issue got a real answer, not a deferral.** The issue floated it as an open "worth checking whether"; the PR splits it correctly — the *fraction* genuinely differs between engine and rating and stays a caller input, the *composition* doesn't and moves onto `Battler`. Catching that the engine composes it at two readings (the damage multiplier *and* the Cull overlay tally, which books on the raw investment rather than the `1 +` multiplier) is the detail that makes taking the fraction as a parameter — rather than reading target health inside — the right call rather than an arbitrary one.
- **Method-vs-property is reasoned, not defaulted.** `EffectiveCriticalChance(double)` breaks the shape of its `EffectiveParryChance`/`EffectiveDodgeChance` neighbours, and the doc-comment says why: crit's enabler is per-skill (#1453), so it takes the `AmplifyDamage` "attacker method taking the per-hit input" form. That's the kind of deviation that reads as sloppiness if unexplained.
- **The clamp stayed rating-side.** Folding `ProcChance` into the shared member would have quietly changed engine behaviour; `CritProcChance` sitting alongside `ParryProcChance`/`DodgeProcChance` keeps the boundary where #2429 put it.
- Doc-comments are long but each paragraph earns it — they encode *why* the shape is what it is, which is the class of comment that survives.

---

ginderjeremiah:
All three points addressed in `78894be` — both should-fixes taken as written, and the nit.

**Display consumers (should-fix).** Taken here rather than deferred; the point that a canonical composition a third of its consumers ignore is a worse state than either endpoint is right. Went with your suggested shape: the composition is now a free function `effectiveCriticalChance(baseCriticalChance, attributes)` in `battle-formulas.ts`, with `Battler.effectiveCriticalChance` delegating — matching `cooldownMultiplier` ⇄ `Battler.cdMultiplier`, since `skills-view` only ever holds a `BattleAttributes`. `SkillTooltip` collapses to `skill.owner.effectiveCriticalChance(skill.criticalChance)` and the skills grid reads the free function directly (its now-unused local `critChanceMultiplier` is gone).

One wrinkle worth flagging since it's a readability call rather than a mechanical one: the member and the free function share a name, and although TypeScript resolves it correctly (class methods aren't in lexical scope), `return effectiveCriticalChance(...)` inside a method of that name reads as recursion. Aliased the import to `composeCriticalChance` at the call site with a one-line comment, rather than renaming either — the member should keep mirroring the backend `EffectiveCriticalChance`, and the free function's name is the accurate one for what it computes.

**`docs/backend-battle.md` (should-fix).** The one-site sentence now names all four products — the avoidance pair, crit, and the execute investment — and notes the frontend crit composition is a free function the tooltip and skills grid read too. Left the crit clause's `c < skill.CriticalChance × CriticalChanceMultiplier` earlier in the bullet as-is: it states the comparison the draw performs, which is still exactly what happens, and the sourcing rule now says where it's composed.

**`missingHpFraction` contract (nit).** Documented on both sides, and agreed on not moving the clamp in — the engine's `Math.Clamp` at the sample site already covers it and folding it in would be a behaviour change. The doc-comment now says the caller owns the clamp, why each caller establishes it differently, and what an out-of-range value would cost — including the over-credited training you flagged, which is the non-obvious half.

Added three `battle-formulas` cases for the extracted function (composition, the `0 × mult = 0` opt-in, the deliberate non-clamp above 1).

Re-verified locally: `dotnet build` clean (0 warnings), backend **3,522 passed**, `npm run lint` + `svelte-check` clean across 1,225 files, frontend **4,046 passed** (312 files) — the 3 new formula cases, with the existing tooltip and skills-view suites passing unchanged against the rerouted composition.

---

PR #2451, opened 2026-07-26
Print the failing backend test detail on a red coverage run

### LINKED ISSUE #2440: test-backend doesn't surface which test failed — the failing name goes to an unpublished TestResults log
Found while investigating a flaky `test-backend` failure on #2438.

## Problem

When a backend test fails in CI, the job log reports only the count:

```
Failed! - Failed: 1, Passed: 1460, Skipped: 0, Total: 1461, Duration: 25s 545ms - Game.Core.Tests.dll (net10.0|x64)
Game.Core.Tests.dll : error run failed: Tests failed:
  '/home/runner/work/GameServer/GameServer/Game.Core.Tests/bin/Debug/net10.0/TestResults/Game.Core.Tests_net10.0_x64.log'
```

The referenced `TestResults/*.log` holds the failing test's name, message, and stack — but it lives under `bin/` on the runner and is never uploaded. The only published artifact is `backend-coverage-report`. So a red `test-backend` tells you *that* one of ~3,500 tests failed and nothing about *which*.

## Why it matters

On #2438 this turned a one-line diagnosis into a full investigation: reproduce the run locally under `dotnet-coverage` (the instrumented path CI uses, which the plain `dotnet test` path doesn't exercise), then reason from first principles about which suites are load-sensitive. The run turned out to be a flake — a re-run of the identical commit went green — but that conclusion took a repro cycle that a test name would have made immediate. It's worse for a genuine failure on a PR touching many files, and worst for an intermittent one that can't be reproduced on demand at all.

## Suggested direction

Publish the failure detail from `build/coverage.ps1` / the `test-backend` job. Cheapest version is an `upload-artifact` step with `if: failure()` over `**/TestResults/*.log`; better is having the script print the failing test names inline on a non-zero `$testExit`, so the answer is in the job log where someone reading a red check is already looking. Microsoft.Testing.Platform can also emit TRX (`--report-trx`), which would give a structured artifact rather than a text log.

Worth confirming the diagnostics survive `dotnet-coverage collect` wrapping the inner `dotnet test` — the wrapper is why the console output is thinner here than a bare `dotnet test` run.

## Related

Separately worth knowing whether the flake has a repeat offender. `Game.Core.Tests` has wall-clock budgets in `Battle/Performance/BattlePerformanceTests.cs` (`MaxLengthRealisticCeilingMs`, the `2.0×` stacking-flatness ratio) and at least one unseeded `new Random()` concurrency case, all of which are runner-load sensitive — but *which* one flaked is exactly what this issue is about not being able to answer. Fixing the reporting is the prerequisite for triaging the flake, so this should land first.

## Scope

Small — CI/workflow plus possibly a few lines in `build/coverage.ps1`. No application code.

### LINKED ISSUE #2441: A failing backend test in CI never names itself — coverage.ps1 swallows the per-test failure detail
Hit while reviewing #2438: `test-backend` failed with **1 failure out of 1461 in `Game.Core.Tests`**, and neither the job log nor the check-run output named the test, the assertion, or the stack. Diagnosing it meant checking out the head commit and re-running the suite locally (green, 1461/1461), then re-running the CI job to establish it was a flake — for what should have been a one-line read of the log.

## Cause

`build/coverage.ps1:38` runs the whole suite inside the collector:

```powershell
dotnet tool run dotnet-coverage collect -f cobertura -o $cobertura "dotnet test Game.sln --no-build --no-restore"
```

`dotnet-coverage` runs the inner `dotnet test` as a profiled child and relays only its summary lines, so all that reaches the console is:

```
Failed! - Failed: 1, Passed: 1460, Skipped: 0, Total: 1461, Duration: 25s 545ms - Game.Core.Tests.dll
Game.Core.Tests.dll : error run failed: Tests failed: '…/TestResults/Game.Core.Tests_net10.0_x64.log'
```

The per-test detail *is* written — to `<project>/bin/Debug/net10.0/TestResults/*.log` — but that path is inside the build output, is never printed, and isn't among the uploaded artifacts (only `coverage/backend-report` is). The exit code is correctly preserved (`:39`, `:56-59`), so the gate works; only the diagnosis is missing.

## Suggested fix

Smallest version: in the existing failure branch at `coverage.ps1:56-59`, dump the referenced logs before exiting.

```powershell
if ($testExit -ne 0) {
  Get-ChildItem -Recurse -Filter '*_net10.0_x64.log' -Path (Join-Path $PSScriptRoot '..') |
    Where-Object { $_.FullName -like '*TestResults*' } |
    ForEach-Object { Write-Host "── $($_.FullName) ──"; Get-Content $_.FullName }
  …
}
```

Alternatives worth weighing: uploading `**/TestResults/**` as an `if: failure()` artifact (keeps the log out of the console but adds a click), or having `dotnet test` emit a TRX the workflow renders as a check annotation (nicest output, most moving parts). Whichever is chosen should work the same locally, since `coverage.ps1` is explicitly the one command for both.

## Related: the nondeterministic tests this would have identified

The failure did not reproduce locally and #2438's diff was arithmetically identity-preserving, so it was near-certainly one of the handful of `Game.Core.Tests` cases that aren't deterministic. Worth auditing once failures are legible — and possibly worth hardening regardless:

- `Battle/Performance/BattlePerformanceTests.cs` — wall-clock budgets on a shared runner (`MaxLengthRealisticCeilingMs` = 250 ms, the 2.0× `StackingFlatnessBudget` ratio). Both take a min/ratio over several samples rather than a single reading, so they're not naive, but they are still timing assertions.
- `Probability/ProbabilityTableTests.cs` — `SetupMultiRandomElementList` builds its weights from an **unseeded** `new Random()`, so each run tests a different distribution against `AssertDistributionIsReasonable`'s fixed ±0.01 absolute tolerance; and `ConcurrentGetRandomValue_StaysValidAndNonDegenerate` is explicitly documented as a best-effort concurrency guard.

Note the second one is not merely flaky-in-principle: an unseeded generator means a failure can't be reproduced even with the test name in hand. Seeding it (as `SortedLinkedListTests:186` already does with `new Random(42)`) would make any future failure reproducible for free.

Filing the observability gap as the issue since it's the thing that blocks diagnosis; the audit above is the natural follow-on work once a failing run says what it failed.

ginderjeremiah:
Closes #2440
Closes #2441

(The two issues were filed 12 seconds apart from two different reviews of the same incident and describe the same gap; one change closes both.)

## Problem

`dotnet-coverage` runs the inner `dotnet test` as a profiled child and relays only its summary lines, so a red `test-backend` job reported a count and a path — and nothing else:

```
Failed! - Failed: 1, Passed: 1460, ... - Game.Core.Tests.dll (net10.0|x64)
Game.Core.Tests.dll : error run failed: Tests failed: '…/bin/Debug/net10.0/TestResults/Game.Core.Tests_net10.0_x64.log'
```

The per-test detail *is* written to that log, but it lives under `bin/` and the only published artifact is `backend-coverage-report`. So a red check told you *that* one of ~3,500 tests failed and nothing about *which*.

## Changes

**`build/coverage.ps1`** — a new `Write-TestFailureDetail` runs in the existing `$testExit -ne 0` branch and dumps the per-project `TestResults` logs to the console. Two filters keep it readable:

- **Projects proven green are skipped** (their log contains `Test run summary: Passed!`). Deliberately fail-open — a crashed, truncated or otherwise unrecognised run is still printed rather than silently swallowed.
- **Progress-spinner lines (`[+n/xn/?n] …`) and blank padding are dropped.** These dominate the logs (`Game.Api.Tests`' log is 220 lines, 14 of them signal) and carry nothing a reader of a failing run needs. Everything else in the log is kept verbatim.

Confirmed against a real failing run — this is the new tail of `./build/coverage.ps1 -NoBuild`, with the three passing projects' logs correctly skipped:

```
── /home/user/GameServer/Game.Infrastructure.Tests/bin/Debug/net10.0/TestResults/Game.Infrastructure.Tests_net10.0_x64.log ──
xUnit.net v3 Microsoft.Testing.Platform v1 Runner v3.2.2+728c1dce01 (64-bit .NET 10.0.10)
failed Game.Infrastructure.Tests.TempProbeTests.DeliberateFailure (7ms)
  Assert.Equal() Failure: Values differ
  Expected: 1
  Actual:   2
    at Game.Infrastructure.Tests.TempProbeTests.DeliberateFailure() in …/TempProbeTests.cs:10
    …
Test run summary: Failed! - …/Game.Infrastructure.Tests.dll (net10.0|x64)
  total: 75
  failed: 1
  succeeded: 74
  …

NOTE: the test run reported failures (exit 1); the coverage figures above are from a non-green run.
```

This lands in the console rather than only in an artifact because that is where someone reading a red check already is — and because `coverage.ps1` is explicitly the one command for local and CI, so the local experience improves identically.

**`.github/workflows/run-tests.yml`** — the `test-backend` job additionally uploads `**/TestResults/*.log` as a `backend-test-logs` artifact `if: failure()`. The console dump answers "which test", the artifact preserves the raw logs including the per-test timing lines the dump filters out — which is exactly the diagnosis for the timing-sensitive flake class that motivated these issues.

**`Game.Core.Tests/Probability/ProbabilityTableTests.cs`** — `SetupMultiRandomElementList` now seeds its generator (`new Random(42)`, matching `SortedLinkedListTests:186`). #2441 calls this out as free and it's the one item from its "related nondeterministic tests" list that is genuinely one line: an unseeded generator builds a different weight set every run, so a failure there couldn't be reproduced *even with the test name in hand* — which would defeat the point of the rest of this PR. The tolerance is unaffected: 1M iterations against a ±0.01 absolute band is ~14σ of sampling error for any seed, and the suite is green.

**`docs/infrastructure.md`** — one bullet in the backend-coverage section recording the diagnostics behaviour.

## Deliberately not in scope

The rest of #2441's nondeterminism audit — the wall-clock budgets in `Battle/Performance/BattlePerformanceTests.cs` (`MaxLengthRealisticCeilingMs`, the 2.0× stacking-flatness ratio) and `ConcurrentGetRandomValue_StaysValidAndNonDegenerate`. Both issues frame the reporting fix as the *prerequisite* for that triage ("this should land first"), and hardening a timing assertion without knowing which one actually flaked is guesswork. Filed as #2452 for once a failing run names itself.

Also considered and rejected: emitting TRX via `--report-trx` and rendering it as a check annotation. Nicest output, most moving parts, and it wouldn't improve the local run — the console dump gets the answer to the same place for a fraction of the surface.

## Test plan

Verified end-to-end locally with `pwsh ./build/coverage.ps1 -NoBuild` (the exact CI invocation), against a temporary deliberately-failing test:

- [x] **Red run** — the failing test's name, assertion and stack are printed; the three passing projects' logs are skipped; exit code still `1`.
- [x] **Green run** (probe removed) — no diagnostic block printed, exit `0`, full suite passes including the newly-seeded `MultiRandomInitializationList_ProducesReasonableDistribution`.
- [x] Coverage gate unaffected — all gated assemblies/namespaces still PASS.
- [x] `dotnet build -warnaserror` clean; `dotnet format --verify-no-changes` clean.

## Note for review

`build/*.ps1` has no test harness of its own (`coverage-gate.ps1`'s `Get-NamespaceCoverage` is likewise untested), and adding one — plus a CI job to run it — would have been several times the size of this `scope: small` issue. The new function is therefore kept to two filter predicates and verified by running the real command against a real failure rather than by unit test. Filed as #2453.


ginderjeremiah (commented):
## Review: print the failing backend test's detail on a red coverage run (#2440 / #2441)

**Verdict:** **Approve** — the change does exactly what both issues asked for, at the size they asked for, and I found nothing blocking or worth a should-fix. (Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR author; the call is approve.)

### Scope check

Both issues frame the same gap — a red `test-backend` names a count and a `TestResults` path but never the test — and both name "print it in the failure branch of `coverage.ps1`" as the smallest correct fix, with the artifact upload and TRX rendering as the alternatives. This lands the console dump *and* the artifact, and the split is well-argued: the console answers "which test" where the reader already is, the artifact preserves the per-test timing lines the dump filters out, which is precisely the diagnosis for the flake class that motivated #2440. Rejecting TRX-as-annotation on the grounds that it wouldn't improve the local run is the right call for a script whose whole premise is one command for local and CI.

### Verification of the mechanism

- `build/coverage.ps1:29` — the glob `*/bin/*/*/TestResults/*.log` matches all four test projects, which are all top-level (`Game.Api.Tests`, `Game.Application.Tests`, `Game.Core.Tests`, `Game.Infrastructure.Tests`), against the real path shape `<Project>/bin/Debug/net10.0/TestResults/<Project>_net10.0_x64.log`.
- The log is UTF-16 **with BOM**, so the un-`-Encoding`'d `Get-Content` decodes correctly on both PS 5.1 and pwsh 7 (both honour the BOM). The file is created fresh at run start rather than appended, so the "skip projects proven green" predicate can't be fooled by a previous run's `Passed!` summary lingering in the same file — which would have been the one way this filter could silently swallow a real failure.
- Failing *open* is the right default for the same reason: a crashed or truncated run has no `Passed!` line, so it gets printed rather than skipped.
- `Write-TestFailureDetail` is placed after the gate output and before the `NOTE:` line, so the failing test lands at the tail of the job log, and `exit $testExit` still governs — the gate semantics are untouched.
- `build/coverage.ps1:28` — untyped positional param matches the existing convention in `coverage-gate.ps1:47`, `coverage-lib.ps1:11`, and `coverage-floor-suggestions.ps1:53`.
- `.github/workflows/run-tests.yml:252-262` — `if: failure()` + `if-no-files-found: ignore` means a failure in the *build* step (no logs written yet) is a clean no-op, and the step mirrors the neighbouring `Upload coverage report` in action version and shape.

### Test coverage

No test for `Write-TestFailureDetail`, and that's the right trade here: `build/*.ps1` has no harness at all (`Get-NamespaceCoverage` and the gate itself are equally untested), so standing one up plus a CI job would have been several times the size of a `scope: small` issue and would have buried a one-purpose diff. Filed as #2453 with the Pester-vs-hand-rolled decision laid out, including the note that the function would need its file enumeration split from its line filtering to be testable without touching disk. Verification by running the real command against a real deliberate failure is a reasonable substitute at this size.

`ProbabilityTableTests.SetupMultiRandomElementList` seeding (`new Random(42)`, matching `SortedLinkedListTests:186`) is the one item from #2441's audit list that was genuinely free, and it belongs in *this* PR rather than the follow-up: an unseeded weight fixture means a failure there isn't reproducible even once CI names the test, which would have partly defeated the point of the reporting fix. The tolerance is unaffected — 1M iterations against a ±0.01 absolute band leaves enormous margin at any seed — and `test-backend` is green on the seeded run. Note the draws themselves still come from the shared RNG, so the case remains statistically nondeterministic; #2452 correctly says so rather than claiming it's now deterministic.

Deferring the rest of the audit (the wall-clock budgets in `BattlePerformanceTests`, `ConcurrentGetRandomValue_StaysValidAndNonDegenerate`) to #2452 is right and matches both issues' own sequencing — hardening a timing assertion without knowing which one flaked is guesswork, and #2452 is explicit that it should be picked up the first time a real failure names its test.

### What's good

- The two filter predicates are the whole of the new logic, and each is justified in-place with the reason rather than the mechanism — including *why* the green-skip fails open, which is the non-obvious half.
- The PR body shows the actual output of a real red run with the three passing projects correctly skipped, rather than asserting the behaviour.
- `docs/infrastructure.md` gets one bullet in the right section, describing the behaviour without narrating the change — consistent with the documentation guidance in CLAUDE.md.
- Follow-ups were filed *before* asking for review, both labelled and both stating why the work isn't in this PR.

No nits.

---

Issue #2447, opened 2026-07-25
Test builders that cast a partial literal `as IContract` opt out of contract-drift detection


ginderjeremiah:
Found while reviewing #2445 (the `IZone` half of #2425).

## Problem

The fixture series (#2414 / #2425 / #2433 / #2434, and the queued #2444 / #2446) defines its target set as *"suites that construct a **typed** contract"*, verified with the throwaway-required-field probe. That criterion is the right one for the cost being removed — but it leaves behind a class of builder that is worth naming separately, because it is invisible to the probe **by construction**:

```ts
// src/tests/lib/common/challenge-unlocks.test.ts:5-16
const zone = (id: number, order: number, name: string, unlockChallengeId?: number, retiredAt?: string): IZone =>
	({ id, name, description: '', order, levelMin: 1, levelMax: 10, bossLevel: 1, unlockChallengeId, retiredAt }) as IZone;
```

The builder is *annotated* `IZone`, duplicates the same eight fields the shared fixture now owns, and yet a new required field on `IZone` produces no error here — the `as IZone` assertion silences it. So the verification each of these PRs states ("adding a required field afterwards produces exactly one test-file error") is true of the probe but slightly overstates the guarantee: the cast sites will keep compiling with a field the contract requires and the object doesn't have.

Known cast-built contract literals under `src/tests/`:

| Contract | Site |
|---|---|
| `IZone` | `lib/common/challenge-unlocks.test.ts:5`, `components/tooltip/ChallengeTooltip.test.ts:23`, `routes/game/screens/codex/enemy-level.test.ts:5` |
| `IEnemy` | `lib/common/enemy-attributes.test.ts:28,40` |
| `IItem` | `routes/game/screens/codex/skill-provenance.test.ts:11`, `routes/game/welcome-back/offline-summary.test.ts:29` |
| `IChallenge` | `lib/common/challenge-unlocks.test.ts:19` |
| `IChallengeType` | `lib/common/challenge-type.test.ts:20` |
| `IAttribute` | `routes/game/screens/skills/SkillDamageBreakdown.test.ts:20` |

This is deliberately **not** the same as the untyped `staticData` mock exclusion, which is correct and should stay: those literals are never checked against the contract at all, and filling their previously-`undefined` fields risks changing what a suite renders. A cast site is different — it *claims* the contract type to its consumer, and the function under test reads it as one.

## Suggested direction

Per contract, decide one of two things for each cast site and record which:

[...]

PR #2454, opened 2026-07-26
Consolidate the hand-rolled IItemMod test builders onto a shared fixture (#2444)

### LINKED ISSUE #2444: Consolidate the hand-rolled IItemMod test builders onto a shared fixture

ginderjeremiah:
## Summary

Adds `UI/src/tests/fixtures/item-mods.ts` exporting `makeItemMod(o)`, following the `fixtures/items.ts` / `fixtures/zones.ts` convention, and converts every suite that constructed a **typed** `IItemMod` inline. Fifth run of the pattern from #2414 / #2425 / #2433 / #2434.

Test-only. No production code, no behaviour change: 11 files converted, −103/+81 lines.

## How it addresses the issue

**The target set differs from the issue's table**, because it was confirmed with the throwaway-required-field probe (`svelte-check`) rather than the grep — exactly the check the issue asked for, and it corrected the survey in both directions:

| Suite | Issue's table | Actual |
|---|---|---|
| `lib/battle/item-mod.test.ts` | *not listed* | **found** — inline `mockItemMods[7]` literal |
| `lib/battle/battle-sim-test-utils.ts` | *not listed* | **found** — inline literal in `equipmentFactory` |
| `routes/admin/workbench/entities/item-mod.test.ts` | "six inline literals" | **three** — the other three spread the production `newItem` blank record, so they aren't hand-rolled and correctly stay |
| `challenges/OverviewPane.test.ts` | "likely excluded" | **confirmed excluded** — partial `as IItemMod` cast opts it out of drift detection (#2447) |

The other nine matched the table. All eleven converted sites: `item-mod.test.ts`, `item.test.ts`, `battle-sim-test-utils.ts`, `inventory-manager.test.ts`, `inventory-manager-reactivity.svelte.test.ts`, `workbench/entities/item-mod.test.ts`, `workbench/references.test.ts`, `ModTooltip.test.ts`, `RewardTooltip.test.ts`, `RewardAffordance.test.ts`, `challenges-view.svelte.test.ts`.

### Module placement

A separate `item-mods.ts` rather than joining `makeItem` in `items.ts`, per the issue's lean — `IItem` and `IItemMod` are separate contracts that drift independently, even though suites usually import both.

### The one default that needed a call

`itemModTypeId` defaults to **`Component`**, matching the workbench's production blank record. It is the one type with no naming semantics — a `Prefix`/`Suffix` affixes the host item's display name — so it stays inert for suites that only care that *a* mod applied. Where the type is load-bearing (it binds a mod to a `modSlots` entry of the matching type), the suite states it explicitly with a comment saying why, rather than inheriting a default that would silently become load-bearing.

Everything else took the neutral default (`rarityId: Common`, empty `attributes`/`tags`, blank `description`/`designerNotes`, `retiredAt` unset). Drifted defaults were reconciled by stating the divergence at the call site — the inventory-manager pair keeps its `Uncommon`/Agility mod, `references.test.ts` and the challenges suites keep their `Prefix`.

### Parity surface

`battle-sim-test-utils.ts` feeds the frontend/backend battle-parity suite, so its conversion is byte-for-byte identity-preserving: the removed literal's `name`/`description`/`designerNotes`/`rarityId`/`tags` are all exactly the fixture's defaults, and only `itemModTypeId`/`attributes` (which the factory already passed through) are stated. No backend mirror is needed — this is test-fixture plumbing, not battle arithmetic.

Removed the now-dead `ERarity` import there and in `references.test.ts`.

## Test plan

- [x] `npm run lint` (eslint + `svelte-check`) — clean, 1226 files, 0 errors, 0 warnings
- [x] `npm run test:unit:ci` — **312 files / 4046 tests passed**
- [x] **Drift-detection verification** (the issue's stated check): re-running the probe with a throwaway required field on `IItemMod` now yields exactly **one** test-file error — `src/tests/fixtures/item-mods.ts` — plus the two production hits that correctly stay literals (`lib/battle/item-mod.ts`'s `ItemMod` class and `routes/admin/workbench/entities/item-mod.ts`'s `newItem`). Before this change the same probe flagged 11 test files.

No backend changes, so no `dotnet` run was required.

## Note on scope

Left `OverviewPane.test.ts` alone deliberately — its `as IItemMod` cast is #2447's class, and converting it here would have pre-empted that issue's per-contract decision. The new fixture's header comment names that exclusion explicitly, as #2447 asks each fixture to do.

Closes #2444.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YbnGYU6LxG62EbYwXJssht)_

### DIFF
--- a/UI/src/tests/fixtures/item-mods.ts
+++ b/UI/src/tests/fixtures/item-mods.ts
@@ -5,12 +5,12 @@ import { EItemModType, ERarity, type IItemMod } from '$lib/api';
    that drift independently, even though a suite exercising applied mods usually imports both.
 
    The target set was confirmed with the throwaway-required-field probe, not a `grep` — only a suite
-   that constructs a *typed* `IItemMod` pays the contract-drift tax. Two exclusions are deliberate:
-   the production blank record (`routes/admin/workbench/entities/item-mod.ts`) and the live `ItemMod`
-   domain class (`lib/battle/item-mod.ts`) are the authoritative shapes and correctly stay literals,
-   and `challenges/OverviewPane.test.ts` asserts a deliberately partial literal `as IItemMod`, which
-   opts it out of drift detection entirely (#2447). A `grep` still finding mod literals under
-   `src/tests/` is expected, not a gap. */
+   that constructs a *typed* `IItemMod` pays the contract-drift tax. Exclusions are deliberate, in two
+   classes: the production blank record (`routes/admin/workbench/entities/item-mod.ts`) and the live
+   `ItemMod` domain class (`lib/battle/item-mod.ts`) are the authoritative shapes and correctly stay
+   literals; `challenges/OverviewPane.test.ts` and `lib/common/challenge-unlocks.test.ts` each assert a
+   deliberately partial literal `as IItemMod`, which opts them out of drift detection entirely (#2447).
+   A `grep` still finding mod literals under `src/tests/` is expected, not a gap. */
 
 /**
  * Builds an {@link IItemMod} reference-data entry for tests. Everything but the id is a neutral
--- a/UI/src/tests/fixtures/item-mods.ts
+++ b/UI/src/tests/fixtures/item-mods.ts
@@ -1,0 +1,35 @@
+import { EItemModType, ERarity, type IItemMod } from '$lib/api';
+
+/* Shared `IItemMod` contract builder (#2444), following the `fixtures/items.ts` convention. It stays
+   a module of its own rather than joining `makeItem`: `IItem` and `IItemMod` are separate contracts
+   that drift independently, even though a suite exercising applied mods usually imports both.
+
+   The target set was confirmed with the throwaway-required-field probe, not a `grep` — only a suite
+   that constructs a *typed* `IItemMod` pays the contract-drift tax. Two exclusions are deliberate:
+   the production blank record (`routes/admin/workbench/entities/item-mod.ts`) and the live `ItemMod`
+   domain class (`lib/battle/item-mod.ts`) are the authoritative shapes and correctly stay literals,
+   and `challenges/OverviewPane.test.ts` asserts a deliberately partial literal `as IItemMod`, which
+   opts it out of drift detection entirely (#2447). A `grep` still finding mod literals under
+   `src/tests/` is expected, not a gap. */
+
+/**
+ * Builds an {@link IItemMod} reference-data entry for tests. Everything but the id is a neutral
+ * placeholder so a suite states only what it asserts on.
+ *
+ * `itemModTypeId` defaults to `Component`, matching the workbench's blank record: it is the one type
+ * with no naming semantics (a `Prefix`/`Suffix` affixes the host item's display name), so it stays
+ * inert for suites that only care that *a* mod applied. A suite whose mod has to fit a specific
+ * `modSlots` entry states the matching type explicitly, since the type is what binds mod to slot.
+ * `attributes`/`tags` default empty and `retiredAt` is left unset for the same reason — an
+ * unretired, statless, untagged mod is the neutral case, and each of those flips a distinct branch.
+ */
+export const makeItemMod = (overrides: Partial<IItemMod> & { id: number }): IItemMod => ({
+	name: `Mod ${overrides.id}`,
+	description: '',
+	itemModTypeId: EItemModType.Component,
+	rarityId: ERarity.Common,
+	attributes: [],
+	tags: [],
+	designerNotes: '',
+	...overrides
+});
--- a/UI/src/tests/lib/battle/battle-sim-test-utils.ts
+++ b/UI/src/tests/lib/battle/battle-sim-test-utils.ts
@@ -2,10 +2,10 @@ import { Battler, newItem } from '$lib/battle';
 // Aliased: this module's own `makeSkill` builds a `SkillSpec` (pre-registration), not the contract.
 import { makeSkill as makeSkillContract } from '../../fixtures/skills';
 import { makeItem } from '../../fixtures/items';
+import { makeItemMod } from '../../fixtures/item-mods';
 import {
 	EAttribute,
 	EDamageType,
-	ERarity,
 	type EModifierType,
 	type ESkillEffectTarget,
 	type EItemModType,
@@ -209,16 +209,7 @@ export function equipmentFactory(itemRegistry: IItem[], itemModRegistry: IItemMo
 	return (spec: ItemSpec): IBattlerAttribute[] => {
 		const appliedMods: IAppliedModModel[] = spec.mods.map((mod, slotIndex) => {
 			const id = itemModRegistry.length;
-			itemModRegistry.push({
-				id,
-				name: `Mod ${id}`,
-				description: '',
-				designerNotes: '',
-				itemModTypeId: mod.type,
-				rarityId: ERarity.Common,
-				attributes: mod.attributes,
-				tags: []
-			});
+			itemModRegistry.push(makeItemMod({ id, itemModTypeId: mod.type, attributes: mod.attributes }));
 			return { itemModId: id, itemModSlotId: slotIndex };
 		});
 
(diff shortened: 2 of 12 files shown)


UI/src/tests/fixtures/item-mods.ts
```diff
@@ -0,0 +1,35 @@
...
+/* Shared `IItemMod` contract builder (#2444), following the `fixtures/items.ts` convention. It stays
+   a module of its own rather than joining `makeItem`: `IItem` and `IItemMod` are separate contracts
+   that drift independently, even though a suite exercising applied mods usually imports both.
+
+   The target set was confirmed with the throwaway-required-field probe, not a `grep` — only a suite
+   that constructs a *typed* `IItemMod` pays the contract-drift tax. Two exclusions are deliberate:
+   the production blank record (`routes/admin/workbench/entities/item-mod.ts`) and the live `ItemMod`
+   domain class (`lib/battle/item-mod.ts`) are the authoritative shapes and correctly stay literals,
+   and `challenges/OverviewPane.test.ts` asserts a deliberately partial literal `as IItemMod`, which
+   opts it out of drift detection entirely (#2447). A `grep` still finding mod literals under
```
ginderjeremiah: Nit — the exclusion list is incomplete, and the count doesn't match the enumeration.

`lib/common/challenge-unlocks.test.ts:48` also builds a partial `as IItemMod` (`{ id: 4, name: 'of Fury', itemModTypeId, rarityId } as IItemMod`), which is exactly the `OverviewPane` class of exclusion. Both `fixtures/items.ts` and `fixtures/zones.ts` name `challenge-unlocks` explicitly in their headers, so leaving it out here is a drift from the sibling convention — and #2447 works from these headers to decide the per-contract call, so a missing entry is the kind of gap that costs someone a re-probe later.

Also "Two exclusions are deliberate:" then names three files. Reading it as two *classes* (authoritative production shapes vs. partial casts) works, but it scans as a miscount. Suggest something like "Exclusions are deliberate, in two classes:" and adding `lib/common/challenge-unlocks.test.ts` alongside `OverviewPane.test.ts`.


---

ginderjeremiah (commented):
## Review: consolidating the hand-rolled `IItemMod` test builders onto `fixtures/item-mods.ts` (#2444)

**Verdict:** Approve (posted as a comment — GitHub won't accept a gating verdict from the PR's own author) — clean, mechanical, test-only conversion that lands the issue's intent; one documentation nit, nothing blocking.

I verified the PR body's claims rather than taking them on faith, and they hold up:

- **The corrected target set is right.** `entities/item-mod.test.ts` really does have three hand-rolled literals, not six — the other three (`:64`, `:162`, and the `newItem defaults` assertion) spread or assert against the production `newItem` blank record, so they're authoritative shapes and correctly stay literals. Converting them would have made the fixture assert against itself.
- **The fixture's defaults match the production blank record.** Lined up field-for-field against the `newItem defaults to a Common Component with empty collections` expectation in `entities/item-mod.test.ts:51` — `description`/`itemModTypeId`/`rarityId`/`designerNotes`/`attributes`/`tags` all agree. That's the right anchor for a reference-data builder's neutral case.
- **No behaviour drift from the `Component` default.** Every converted site that previously relied on `Prefix` restates it (`item.test.ts`, `item-mod.test.ts`, `references.test.ts`, `challenges-view`, `ModTooltip`, `RewardTooltip`, `RewardAffordance`, both workbench baselines), so no suite silently changed the type its mod carries.
- **The parity surface is untouched.** `battle-sim-test-utils.ts`'s removed literal is default-identical to the fixture on `name`/`description`/`designerNotes`/`rarityId`/`tags`, with only `itemModTypeId`/`attributes` — the two the factory already threaded through — stated. Agreed that no backend mirror is needed: this is fixture plumbing, not battle arithmetic, so the frontend/backend parity rule in `CLAUDE.md` doesn't bite here.

### Nits

- `UI/src/tests/fixtures/item-mods.ts:8-12` — the exclusion list omits `lib/common/challenge-unlocks.test.ts:48`, which carries the same partial `as IItemMod` cast as `OverviewPane.test.ts`; both sibling fixtures name it. The "Two exclusions" phrasing also enumerates three files. Inline comment has the detail.

### Test coverage

Appropriate as-is. This is test-infrastructure consolidation with no production diff, so there's nothing new to cover — the existing suites *are* the coverage, and the meaningful verification is the drift-detection probe, which the PR ran: one test-file error after the change (the fixture) versus eleven before. That's the check #2444 asked for and the right one. `verify-frontend` (ESLint + `svelte-check`) is green on the head commit, which covers the dead-import cleanup in `battle-sim-test-utils.ts` and `references.test.ts`.

### What's good

- Stating a default explicitly *with a reason* where it's load-bearing — the `// Stated rather than inherited: it has to match the Component slot the fixture item carries` comments in the inventory-manager pair and `challenges-view` — is the right instinct. A silently-inherited default that a later fixture edit could flip is precisely the foot-gun these conversions can introduce, and calling it out at the call site defuses it.
- Correcting the issue's survey in *both* directions and documenting why, rather than mechanically converting what the table listed. `CLAUDE.md` asks for exactly that skepticism toward issue bodies.
- Keeping the local id-only adapters (`makeItem`/`makeItemMod` in the inventory-manager suites, `itemMod` in `references.test.ts`) as thin pass-throughs instead of rewriting every call site — the diff stays small and reviewable, which matters for a fifth run of a repeated pattern.


---

UI/src/tests/fixtures/item-mods.ts
ginderjeremiah: Both correct — fixed in 576c9c1.

Verified the cast before changing anything: `challenge-unlocks.test.ts:48` is `itemMods[4] = { id: 4, name: 'of Fury', itemModTypeId: EItemModType.Suffix, rarityId: ERarity.Epic } as IItemMod`, same class as `OverviewPane`. And the sibling headers do name it — `items.ts:8` and `zones.ts:11` both call out `challenge-unlocks` by name, so omitting it here was a straight drift from the convention.

Took your phrasing for the count: "Exclusions are deliberate, in two classes:", with `challenge-unlocks.test.ts` listed alongside `OverviewPane.test.ts` in the partial-cast class.

The point about #2447 working *from* these headers is the part that made this worth a commit rather than a shrug — that issue's whole job is deciding convert-or-keep per cast site, and it enumerates them from the fixture comments. A missing entry there doesn't just read as sloppy, it hides a site from the pass that exists to triage it.

`npm run lint` still clean (1226 files, 0 errors). Comment-only change, no test impact.

---

PR #2455, opened 2026-07-26
Consolidate the hand-rolled IEnemy test builders onto a shared fixture (#2446)

### LINKED ISSUE #2446: Consolidate the hand-rolled IEnemy test builders onto a shared fixture
Found while doing #2434 (the `IZone` half of #2425). `IEnemy` has the same duplication class as `ISkill` / `IItem` / `IZone` / `IItemMod`, but — like `IItemMod` (#2444) — it wasn't in #2425's survey, so it never got split out.

## Problem

**9 suites** construct a typed `IEnemy` inline. That's the same size as the `IZone` set #2434 just closed, and **six of the nine are the very files #2434 converted for `IZone`** — they now call the shared `makeZone` and still hand-roll an enemy next to it.

Confirmed with the throwaway-required-field probe (`svelte-check`), not a grep — a `grep "IEnemy => ({"` finds only 3 of the 9, so the grep badly under-counts here:

| Suite | Local builder |
|---|---|
| `lib/common/enemy-attributes.test.ts` | inline literal |
| `lib/engine/battle/battle-engine.test.ts` | inline literal |
| `lib/engine/battle/battle-engine-parity.test.ts` | inline literal |
| `routes/admin/workbench/entities/enemy.test.ts` | inline literal |
| `routes/admin/workbench/references.test.ts` | `enemy` |
| `routes/game/screens/fight/Fight.test.ts` | inline literal |
| `routes/game/screens/fight/boss/boss-view.test.ts` | inline literal |
| `routes/game/screens/skills/Skills.test.ts` | `enemy` (defaults `attributeDistribution` to `1` Endurance/level) |
| `routes/game/screens/skills/skills-view.test.ts` | `enemy` (defaults `attributeDistribution` to `[]`) |

The tenth probe hit is `routes/admin/workbench/entities/enemy.ts`'s production `newItem`, which is the authoritative blank record and correctly stays a literal.

## Suggested direction

Same shape as #2414 / #2425 / #2433 / #2434, which have now run four times cleanly:

- Add `UI/src/tests/fixtures/enemies.ts` exporting `makeEnemy(o)`, following the existing convention.
- Confirm the target set with the probe rather than a grep (see above — the grep finds a third of it).
- Reconcile drifted defaults **explicitly**, preferring whichever variant has a live assertion; leave untyped `staticData`-mock literals alone.
- Convert one suite at a time, running each in between.

## One default that needs a deliberate call

The two Skills suites disagree on `attributeDistribution`, and it is load-bearing in both: `Skills.test.ts` defaults every enemy to `1` Endurance per level (its comment: *"amountPerLevel 1 gives each enemy Toughness 2·(1·level), within the preset-derived slider ceiling"*), while `skills-view.test.ts` defaults to `[]` and passes a per-enemy `enemyDist(n)` so each test gets a clean expected Toughness. `[]` is the neutral default — a zero-Toughness enemy — but `Skills.test.ts` then has to state its own, and the reason it needs a non-zero one (staying under the compare-slider ceiling) should move to that wrapper rather than being lost.

`battle-engine-parity.test.ts` is on the frontend/backend parity surface, so its enemies feed scenarios that must stay row-for-row aligned with the backend's. Any default change there has to leave the parity matrix arithmetically identical — worth converting last, or leaving out if its literals turn out to be scenario data rather than a builder.

## Verification

Adding a required field to `IEnemy` should afterwards produce exactly one test-file error (the new fixture), plus the production `entities/enemy.ts` hits.

## Scope

Medium — test-only, no production code, no behaviour change, but wide.

Refs #2425, #2434, #2444.

ginderjeremiah:
## Summary

Adds `UI/src/tests/fixtures/enemies.ts` exporting `makeEnemy(o)`, following the `fixtures/items.ts` / `fixtures/zones.ts` convention, and converts every suite that constructed a **typed** `IEnemy` inline. Sixth run of the pattern from #2414 / #2425 / #2433 / #2434 / #2444.

Test-only. No production code, no behaviour change: 8 files converted, −97/+45 lines.

## How it addresses the issue

**The target set is eight suites, not the issue's nine** — confirmed with the throwaway-required-field probe (`svelte-check`), which is the check the issue asked for over the grep:

| Suite | Issue's table | Actual |
|---|---|---|
| `lib/common/enemy-attributes.test.ts` | listed | **excluded** — both its builders assert a partial literal `as IEnemy`, so the cast opts them out of drift detection and the probe never sees them (#2447's class) |
| the other eight | listed | all confirmed |

Converted: `battle-engine.test.ts`, `battle-engine-parity.test.ts`, `workbench/entities/enemy.test.ts` (3 sites), `workbench/references.test.ts`, `Fight.test.ts`, `boss/boss-view.test.ts`, `Skills.test.ts`, `skills-view.test.ts`.

The probe's tenth hit is `routes/admin/workbench/entities/enemy.ts`'s production `newItem`, which is the authoritative blank record and correctly stays a literal.

### The default that needed a deliberate call

`attributeDistribution` defaults to **empty** — the neutral zero-Toughness enemy, per the issue's lean. Every derived attribute resolves to 0 at any level, so nothing is silently load-bearing. The two Skills suites are then reconciled explicitly rather than averaged:

- **`skills-view.test.ts`** already defaulted to `[]` and passes a per-enemy `enemyDist(n)`, so its local builder was exactly the fixture — deleted, calls the fixture directly. Its comment (why each enemy states its own distribution) moved to the `ENEMIES` array it actually describes.
- **`Skills.test.ts`** genuinely needs a non-zero default, so it **keeps a local wrapper** over `makeEnemy` stating its `1` Endurance/level. The reason — staying under the preset-derived compare-slider ceiling — moved onto that wrapper, so it survives as a stated divergence instead of being lost.

### Parity surface

`battle-engine-parity.test.ts` sits on the frontend/backend parity surface, so its conversion is identity-preserving: the removed literal's `designerNotes`/`isBoss`/`attributeDistribution`/`spawns` are exactly the fixture's defaults, and only `name`/`skillPool` are stated. The engine derives the enemy battler from the scenario's attributes rather than the catalogue record, so the empty distribution is what keeps the live run arithmetically identical to the simulator's — noted in place at the call site. The parity matrix is unchanged, so no backend mirror is needed: this is fixture plumbing, not battle arithmetic.

### Other notes

- `workbench/entities/enemy.test.ts` builds `WorkbenchEnemy` (an `IEnemy` plus `bossZones`), so its three hand-rolled records spread the fixture and add that field. The five that already spread `baseline` are untouched.
- `workbench/references.test.ts` keeps its positional `enemy(id, name, opts)` wrapper, now a one-liner delegating to the fixture — matching the `zone`/`item` wrappers already in that file, so no call site moved.

## Test plan

- [x] `npm run lint` (eslint + `svelte-check`) — clean, 1226 files, 0 errors, 0 warnings
- [x] `npm run test:unit:ci` — **312 files / 4046 tests passed**
- [x] **Drift-detection verification** (the issue's stated check): re-running the probe with a throwaway required field on `IEnemy` now yields exactly **one** test-file error — `src/tests/fixtures/enemies.ts` — plus the production `entities/enemy.ts` hit that correctly stays a literal. Before this change the same probe flagged 8 test files across 10 sites.

No backend changes, so no `dotnet` run was required.

## Note on scope

Left `enemy-attributes.test.ts` alone deliberately — its `as IEnemy` casts are #2447's class, and converting them here would have pre-empted that issue's per-contract decision. Worth flagging for whoever picks up #2447: its local `makeEnemy` there supplies *every* `IEnemy` field and still casts, so that cast looks vestigial rather than deliberate. The new fixture's header comment names the exclusion explicitly, as #2447 asks each fixture to do.

Closes #2446.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GBTVXpjgoJXcYpoW5Xmv8W)_

### DIFF
--- a/UI/src/tests/fixtures/enemies.ts
+++ b/UI/src/tests/fixtures/enemies.ts
@@ -4,10 +4,13 @@ import type { IEnemy } from '$lib/api';
    convention.
 
    The target set comes from the throwaway-required-field probe, not a `grep` — a
-   `grep "IEnemy => ({"` finds three of the eight suites. Excluded on purpose: `lib/common/
-   enemy-attributes.test.ts`, whose two builders assert a partial literal `as IEnemy`. The cast opts
-   them out of drift detection, so the probe never sees them and converting them is #2447's call, not
-   this one's. A `grep` still finding enemy literals under `src/tests/` is expected, not a gap. */
+   `grep "IEnemy => ({"` finds three of the eight suites. Excluded on purpose:
+   `lib/common/enemy-attributes.test.ts` (two builders) and
+   `routes/game/screens/codex/enemy-level.test.ts` (one), which assert a partial literal `as IEnemy`.
+   The cast opts them out of drift detection, so the probe never sees them and converting them is
+   #2447's call, not this one's — note that issue's table lists `enemy-level.test.ts` only under
+   `IZone`, so its `IEnemy` builder needs picking up there too. A `grep` still finding enemy literals
+   under `src/tests/` is expected, not a gap. */
 
 /**
  * Builds an {@link IEnemy} reference-data entry for tests. Everything but the id is a neutral
--- a/UI/src/tests/fixtures/enemies.ts
+++ b/UI/src/tests/fixtures/enemies.ts
@@ -1,0 +1,31 @@
+import type { IEnemy } from '$lib/api';
+
+/* Shared `IEnemy` contract builder (#2446), following the `fixtures/items.ts` / `fixtures/zones.ts`
+   convention.
+
+   The target set comes from the throwaway-required-field probe, not a `grep` — a
+   `grep "IEnemy => ({"` finds three of the eight suites. Excluded on purpose: `lib/common/
+   enemy-attributes.test.ts`, whose two builders assert a partial literal `as IEnemy`. The cast opts
+   them out of drift detection, so the probe never sees them and converting them is #2447's call, not
+   this one's. A `grep` still finding enemy literals under `src/tests/` is expected, not a gap. */
+
+/**
+ * Builds an {@link IEnemy} reference-data entry for tests. Everything but the id is a neutral
+ * placeholder so a suite states only what it asserts on.
+ *
+ * `attributeDistribution` defaults to **empty** — a zero-Toughness, zero-health enemy. That is the
+ * neutral case (every derived attribute resolves to 0 at any level), so a suite whose assertions
+ * turn on an enemy's resolved attributes is expected to state its own distribution rather than
+ * inherit arithmetic from here. `spawns` and `skillPool` are empty and `isBoss` is false for the same
+ * reason: each one flips a distinct branch (zone spawn tables, the enemy's usable skills, the boss
+ * pill) when present. `retiredAt` is left unset so the fixture enemy is live.
+ */
+export const makeEnemy = (overrides: Partial<IEnemy> & { id: number }): IEnemy => ({
+	name: `Enemy ${overrides.id}`,
+	isBoss: false,
+	attributeDistribution: [],
+	skillPool: [],
+	spawns: [],
+	designerNotes: '',
+	...overrides
+});
--- a/UI/src/tests/lib/engine/battle/battle-engine-parity.test.ts
+++ b/UI/src/tests/lib/engine/battle/battle-engine-parity.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { EAttribute, type IBattlerAttribute, type IEnemy, type ISkill } from '$lib/api';
 import { DEFAULT_MAX_BATTLE_MS } from '$lib/api/types/game-constants';
+import { makeEnemy } from '../../../fixtures/enemies';
 
 // Both consumers of the shared `battleStep` — the live, render-driven `BattleEngine.logicalUpdate` and the
 // headless `BattleSimulator` — must produce the same outcome for the same inputs and seed, since the backend
@@ -233,15 +234,10 @@ describe('BattleEngine parity with the headless BattleSimulator', () => {
 			// Configure the live engine to derive the identical player/enemy battlers and run it to completion.
 			mockPlayerManager.attributes = toBattlerAttributes(scenario.playerAttrs);
 			mockPlayerManager.selectedSkills = playerSkillIds;
-			mockEnemies[1] = {
-				id: 1,
-				name: 'Enemy',
-				designerNotes: '',
-				isBoss: false,
-				attributeDistribution: [],
-				skillPool: enemySkillIds,
-				spawns: []
-			};
+			// The engine derives the enemy's battler from the scenario's attributes, not the catalogue
+			// record, so an empty `attributeDistribution` keeps the live run arithmetically identical
+			// to the simulator's.
+			mockEnemies[1] = makeEnemy({ id: 1, name: 'Enemy', skillPool: enemySkillIds });
 
 			const engine = new BattleEngine();
 			engine.start();
--- a/UI/src/tests/lib/engine/battle/battle-engine.test.ts
+++ b/UI/src/tests/lib/engine/battle/battle-engine.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { EDamageType, EAttribute, ELogType, EModifierType, ESkillEffectTarget } from '$lib/api';
 import type { ISkill, IEnemy, IEnemyInstance } from '$lib/api';
 import { makeSkill } from '../../../fixtures/skills';
+import { makeEnemy } from '../../../fixtures/enemies';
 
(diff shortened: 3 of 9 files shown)


UI/src/tests/fixtures/enemies.ts
```diff
@@ -0,0 +1,31 @@
+import type { IEnemy } from '$lib/api';
+
+/* Shared `IEnemy` contract builder (#2446), following the `fixtures/items.ts` / `fixtures/zones.ts`
+   convention.
+
+   The target set comes from the throwaway-required-field probe, not a `grep` — a
+   `grep "IEnemy => ({"` finds three of the eight suites. Excluded on purpose: `lib/common/
+   enemy-attributes.test.ts`, whose two builders assert a partial literal `as IEnemy`. The cast opts
+   them out of drift detection, so the probe never sees them and converting them is #2447's call, not
```
ginderjeremiah: **Nit** — the excluded set is under-counted by one. `enemy-attributes.test.ts` is not the only `as IEnemy` cast site: `routes/game/screens/codex/enemy-level.test.ts:11-12` has a third one —

```ts
const enemy = (over: Partial<IEnemy>): IEnemy =>
	({ id: 0, name: 'E', isBoss: false, attributeDistribution: [], skillPool: [], spawns: [], ...over }) as IEnemy;
```

Same class, same reason it's invisible to the probe. Worth naming here, since this comment is the record a later reader (and #2447) works from — #2447's table lists that file only under `IZone` (its zone builder at line 5), so its `IEnemy` builder would be missed on that pass too.

Suggest widening to something like *"Excluded on purpose: `lib/common/enemy-attributes.test.ts` (two builders) and `routes/game/screens/codex/enemy-level.test.ts` (one), which assert a partial literal `as IEnemy`."*

---

ginderjeremiah (commented):
## Review: consolidating the hand-rolled `IEnemy` test builders onto `fixtures/enemies.ts` (#2446)

**Verdict:** Approve (posted as a comment — GitHub won't accept a gating verdict from the PR author) — the conversion is faithful, the defaults are neutral and documented, and the one finding is a comment inaccuracy, not a behaviour or design problem.

### Verification I did

- **Every conversion is identity-preserving.** Checked each of the eight sites against its removed literal: in all of them the dropped fields (`designerNotes: ''`, `isBoss: false`, `attributeDistribution: []`, `skillPool: []`, `spawns: []`) are exactly the fixture's defaults, so no suite silently changed what it feeds the code under test. That includes `battle-engine-parity.test.ts`, which is the one that had to be exact — the parity matrix is untouched, and the empty distribution is what keeps the live engine's arithmetic identical to the simulator's. No backend mirror needed: this is fixture plumbing, not battle arithmetic, so the `CLAUDE.md` both-sides rule doesn't bite here.
- **`makeEnemy` covers the contract.** All seven required `IEnemy` fields; `retiredAt` correctly left unset so the fixture enemy is live. Same signature (`Partial<IEnemy> & { id: number }`), same doc-comment shape, same header-comment convention as `fixtures/items.ts` / `fixtures/zones.ts`.
- **The target set is right, and the issue's table was wrong.** Confirmed independently: every other test file still carrying an enemy literal is either an untyped `staticData` mock (`enemy-manager-boss`, `reference`, `entities/challenge`, `entities/zone`, `Codex`, `codex-view`, `Statistics` — all partial literals that would never satisfy `IEnemy`) or an `as IEnemy` cast. Excluding `enemy-attributes.test.ts` is the correct call, not a shortcut — #2447 explicitly owns that class and asks for it as one pass across contracts.
- **No leftovers.** Every converted suite still uses its `IEnemy` import (as an array annotation or a wrapper signature), so nothing dead was left behind.

### Nits

- `UI/src/tests/fixtures/enemies.ts:7-9` — the header comment names `enemy-attributes.test.ts` as *the* `as IEnemy` exclusion, but `routes/game/screens/codex/enemy-level.test.ts:11-12` is a second one. Inline comment has the detail. Worth fixing here rather than as a follow-up: it's a one-line comment edit, and this comment is the record #2447 will work from.

### Test coverage

Test-only change with no production code, so the rule that applies is that the converted suites keep asserting exactly what they asserted before — they do. The stated drift-detection check (a throwaway required field on `IEnemy` now erroring in one test file instead of eight) is the right verification for this issue, and it's the same one the five prior runs used.

### What's good

- The two Skills suites were genuinely reconciled rather than averaged. `skills-view.test.ts`'s builder was already the fixture, so it's gone; `Skills.test.ts` keeps a local wrapper because its non-zero Endurance default is load-bearing, and the *reason* (staying under the preset-derived compare-slider ceiling) moved onto that wrapper instead of being lost. That's the harder, correct version of this change.
- Comments were relocated to what they actually describe — `skills-view.test.ts`'s per-enemy-distribution rationale now sits on the `ENEMIES` array, and the parity suite's empty-distribution reasoning sits at the call site where it matters.
- `references.test.ts` keeping its positional `enemy(id, name, opts)` wrapper as a one-line delegate is the right restraint: it matches the sibling `zone`/`item` wrappers and moves zero call sites.
- The PR description correcting the issue's nine-suite table down to eight, with the reason, is exactly the verification `CLAUDE.md` asks for rather than taking the issue at face value.

---

Issue #2456, opened 2026-07-26
Consolidate the hand-rolled IChallenge test builders onto a shared fixture


ginderjeremiah:
Split out of #2447 while doing its cast-site pass.

## Why this isn't part of #2447

#2447 asks, per cast site, for one of two outcomes: convert it onto the shared fixture and drop the
cast, or keep the cast with a stated reason. For every contract in the fixture series
(`IZone` / `IEnemy` / `IItem` / `IItemMod` / `ISkill` / `IAttribute`) outcome 1 was available, because
`UI/src/tests/fixtures/<contract>.ts` already exists.

**`IChallenge` has no shared fixture at all** — the series (#2414 / #2425 / #2433 / #2434 / #2444 /
#2446) never covered it. So its cast sites have no fixture to convert *onto*, and the only way to drop
the cast in place is to hand-write all seven required fields at each site — which is the duplication the
series exists to remove, not a fix. They were left casting with a stated reason instead (outcome 2),
which is correct but temporary.

Two other contracts were completed inline in #2447 rather than deferred here, because a fixture module
would have been ceremony: `IChallengeType` (three required fields, one site, a complete non-cast literal
already in the same suite) and `ITag` (three required fields, and the one cast site already supplied all
three — the cast was pure redundancy).

## Scale

This is a full contract consolidation in the #2444 / #2446 mould, not a cast cleanup — 15 suites under
`src/tests/` construct an `IChallenge`:

```
components/tooltip/ChallengeTooltip.test.ts          routes/admin/workbench/references.test.ts
lib/common/challenge-unlocks.test.ts                 routes/admin/workbench/SectionRenderer.test.ts
routes/admin/workbench/challenge-helpers.test.ts     routes/game/screens/challenges/Challenges.test.ts
routes/admin/workbench/challenge/ChallengeConditionSection.test.ts
routes/admin/workbench/challenge/ChallengeRewardSection.test.ts
routes/admin/workbench/challenge/GoalField.test.ts   routes/game/screens/challenges/challenges-view.svelte.test.ts
routes/admin/workbench/challenge/ScopeField.test.ts  routes/game/screens/codex/Codex.test.ts
routes/admin/workbench/entities/challenge.test.ts    routes/game/screens/codex/codex-view.svelte.test.ts
routes/game/screens/fight/ZoneNav.test.ts
```

Of those, the ones currently casting a partial literal `as IChallenge` — i.e. the sites #2447 hands over
— are:

[...]

PR #2457, opened 2026-07-26
Fold the cast-built contract literals back into drift detection (#2447)

### LINKED ISSUE #2447: Test builders that cast a partial literal `as IContract` opt out of contract-drift detection

ginderjeremiah:
Closes #2447.

## Summary

A builder that asserts a partial literal `as IContract` is annotated with the contract and read as one by the code under test, but the assertion silences a missing required field — so it keeps compiling when the contract grows one. That made these sites invisible to the throwaway-required-field probe the fixture series (#2414 / #2425 / #2433 / #2434 / #2444 / #2446) used to define its target set, and left them quietly outside the guarantee each of those PRs stated.

This converts every such site for the six contracts that **already have a shared fixture** onto `make<Contract>`, dropping the cast so each rejoins drift detection. The issue's **category 1** applied everywhere it was reached: the fields the casts hid are inert at each site (the code under test reads spawn/level fields, reward ids, damage portions, or display metadata), and the fixtures supply neutral values for the rest.

## Sites converted

| Contract | Sites |
|---|---|
| `IZone` | `lib/common/challenge-unlocks`, `components/tooltip/ChallengeTooltip`, `codex/enemy-level` |
| `IEnemy` | `lib/common/enemy-attributes` (two), `codex/enemy-level` |
| `IItem` | `codex/skill-provenance`, `welcome-back/offline-summary`, `components/tooltip/ChallengeTooltip`, `lib/common/challenge-unlocks` |
| `IItemMod` | `challenges/OverviewPane`, `lib/common/challenge-unlocks` |
| `ISkill` | `fight/ActiveEffectChips` (three), `inventory/inventory-view` (two builders) |
| `IAttribute` | `skills/SkillDamageBreakdown`, `battle/battle-attributes`, `select/class-summary` |
| `IChallengeType` | `lib/common/challenge-type` |

Two consolidations fell out of the pass: `enemy-attributes` had two local `IEnemy` builders where the leaner one duplicated the shared fixture exactly (now deleted; the richer one became `marauder()`), and `inventory-view` carried the *same* `makeSkill` builder in two separate describes (now one `damageSkill` adapter over the shared fixture, keeping the positional call signature so the 16 call sites didn't churn).

## Corrections to the issue's table

The table is captioned "known", and the pass turned up more:

- **`IItem`** also at `ChallengeTooltip.test.ts:84` and `challenge-unlocks.test.ts:46`; **`IItemMod`** at `challenge-unlocks.test.ts:48` and `OverviewPane.test.ts:19`. These were already named in the `fixtures/items.ts` / `fixtures/item-mods.ts` headers but didn't make the table.
- **`ISkill`** was absent from the table entirely — `ActiveEffectChips` (three literals) and `inventory-view` (two builders, via `as unknown as ISkill`).
- **`IAttribute`** also at `battle-attributes.test.ts:301` and `class-summary.test.ts:30`.
- The `IEnemy` row correction from the issue's own comment (`enemy-level.test.ts:11`) is included.

## What is deferred, and why

`IChallenge` and `ICreatableClass` have **no shared fixture** — the series never covered them — so their cast sites have nothing to convert *onto*. Dropping the cast in place would mean hand-writing all seven required `IChallenge` fields at each of nine sites, which is the duplication the series exists to remove, not a fix. They take the issue's **category 2** instead: the cast stays with a stated reason, and each site now carries a note pointing at **#2456**, filed for the full `IChallenge` consolidation (15 suites construct one) with `ICreatableClass` folded in as a smaller companion.

`IChallengeType` was **not** deferred despite also lacking a fixture: three required fields, one site, and the same suite already carried a complete non-cast literal to copy — so standing up a module for it would have been over-engineering.

Per-contract fixture headers now name this class explicitly, as the issue asks — `fixtures/zones.ts` previously did not, and `fixtures/enemies.ts` / `item-mods.ts` carried "that's #2447's call" notes that are now resolved.

## One judgement call worth flagging

`offline-summary.test.ts` built its item with `itemCategoryId: 0`, which is not a member of `EItemCategory` (it starts at `Helm = 1`) — a meaningless value the double cast permitted. Nothing asserts on the category, so it now takes the fixture's neutral `Accessory` default rather than inventing a category.

## Test plan

- `npm run lint` (eslint + `svelte-check`) — clean, **1227 files, 0 errors, 0 warnings**.
- `npm run test:unit:ci` — **312 files, 4046 tests, all passing**.
- **The probe itself**, which is the actual point of the issue: adding a throwaway required field to all six contracts at once produces **exactly six errors under `src/tests/` — one per fixture module, and none at any test site**. Every remaining error is production code (the workbench blank records, the live `Item`/`ItemMod`/`Skill` domain classes) that legitimately must change on a contract change. Before this PR the converted sites contributed zero errors.
- No production code is touched; the diff is test-only and behaviour-preserving.

https://claude.ai/code/session_01XcqgkY3pQuJym53w1qH9b3

---
_Generated by [Claude Code](https://claude.ai/code/session_01XcqgkY3pQuJym53w1qH9b3)_

### DIFF
--- a/UI/src/tests/routes/admin/workbench/challenge/ChallengeRewardSection.test.ts
+++ b/UI/src/tests/routes/admin/workbench/challenge/ChallengeRewardSection.test.ts
@@ -48,6 +48,7 @@ const config = (): EntityConfig<Identified> =>
 	}) as unknown as EntityConfig<Identified>;
 
 const setup = (over: Partial<IChallenge> = {}) => {
+	/* Deliberately partial: `IChallenge` has no shared fixture yet (#2456). */
 	const challenge: IChallenge = {
 		id: 1,
 		name: 'Test',
--- a/UI/src/tests/routes/admin/workbench/challenge/ChallengeRewardSection.test.ts
+++ b/UI/src/tests/routes/admin/workbench/challenge/ChallengeRewardSection.test.ts
@@ -47,8 +47,9 @@ const config = (): EntityConfig<Identified> =>
 		persist: async () => []
 	}) as unknown as EntityConfig<Identified>;
 
+/* Every `as IChallenge` literal in this file is deliberately partial: `IChallenge` has no shared
+   fixture yet (#2456). */
 const setup = (over: Partial<IChallenge> = {}) => {
-	/* Deliberately partial: `IChallenge` has no shared fixture yet (#2456). */
 	const challenge: IChallenge = {
 		id: 1,
 		name: 'Test',
--- a/UI/src/tests/routes/select/ClassPicker.test.ts
+++ b/UI/src/tests/routes/select/ClassPicker.test.ts
@@ -7,6 +7,7 @@ import ClassPicker from '$routes/select/ClassPicker.svelte';
 import { staticData } from '$stores/static-data.svelte';
 import { EEquipmentSlot, EModifierType, type ICreatableClass } from '$lib/api';
 
+/* Deliberately partial: `ICreatableClass` has no shared fixture yet (#2456). */
 const cls = (overrides: Partial<ICreatableClass> = {}): ICreatableClass =>
 	({
 		id: 0,
--- a/UI/src/tests/components/tooltip/ChallengeTooltip.test.ts
+++ b/UI/src/tests/components/tooltip/ChallengeTooltip.test.ts
@@ -19,12 +19,15 @@ const { staticData, playerChallenges } = vi.hoisted(() => ({
 vi.mock('$stores', () => ({ staticData, playerChallenges }));
 
 import ChallengeTooltip from '$components/tooltip/ChallengeTooltip.svelte';
+import { makeItem } from '../../fixtures/items';
+import { makeZone } from '../../fixtures/zones';
 
 const zone = (id: number, order: number, name: string, unlockChallengeId?: number): IZone =>
-	({ id, name, description: '', order, levelMin: 1, levelMax: 10, bossLevel: 1, unlockChallengeId }) as IZone;
+	makeZone({ id, name, order, unlockChallengeId });
 
 const setup = (over: Partial<IChallenge> = {}) => {
 	staticData.challenges = [];
+	// Deliberately partial: `IChallenge` has no shared fixture yet (#2456); the tooltip reads only these fields.
 	staticData.challenges[5] = {
 		id: 5,
 		name: 'Cull the Swarm',
@@ -76,12 +79,12 @@ describe('ChallengeTooltip', () => {
 	it('seals a reward name while the challenge is incomplete, then reveals it once complete', () => {
 		setup({ rewardItemId: 3 });
 		staticData.items = [];
-		staticData.items[3] = {
+		staticData.items[3] = makeItem({
 			id: 3,
 			name: 'Iron Helm',
 			itemCategoryId: EItemCategory.Helm,
 			rarityId: ERarity.Rare
-		} as IItem;
+		});
 
 		const { container, unmount } = render(ChallengeTooltip, { props: { challengeId: 5 } });
 		const sealed = container.querySelector('.ct-name.sealed') as HTMLElement;
--- a/UI/src/tests/fixtures/attributes.ts
+++ b/UI/src/tests/fixtures/attributes.ts
@@ -1,5 +1,10 @@
 import { EAttribute, EAttributeType, type IAttribute } from '$lib/api';
 
+/* A literal asserted `as IAttribute` claims the contract to its consumer while silencing a missing
+   required field, so it looks covered and isn't. #2447 folded those in (`battle/battle-attributes`,
+   `skills/SkillDamageBreakdown`, `select/class-summary`); reintroducing such a cast re-opens the hole,
+   so state why if you do. */
+
 /**
  * Builds an {@link IAttribute} reference-data entry for tests. The display metadata defaults to
  * neutral placeholders so tests that only care about id/name resolution stay terse; pass `overrides`
--- a/UI/src/tests/fixtures/enemies.ts
+++ b/UI/src/tests/fixtures/enemies.ts
@@ -3,14 +3,12 @@ import type { IEnemy } from '$lib/api';
 /* Shared `IEnemy` contract builder (#2446), following the `fixtures/items.ts` / `fixtures/zones.ts`
    convention.
 
-   The target set comes from the throwaway-required-field probe, not a `grep` — a
-   `grep "IEnemy => ({"` finds three of the eight suites. Excluded on purpose:
+   The target set came from the throwaway-required-field probe, not a `grep`. That probe is blind to
+   one class by construction: a builder that asserts a partial literal `as IEnemy` keeps compiling
+   when the contract gains a required field, so it never showed up. #2447 folded those in —
    `lib/common/enemy-attributes.test.ts` (two builders) and
-   `routes/game/screens/codex/enemy-level.test.ts` (one), which assert a partial literal `as IEnemy`.
-   The cast opts them out of drift detection, so the probe never sees them and converting them is
-   #2447's call, not this one's — note that issue's table lists `enemy-level.test.ts` only under
-   `IZone`, so its `IEnemy` builder needs picking up there too. A `grep` still finding enemy literals
-   under `src/tests/` is expected, not a gap. */
+   `routes/game/screens/codex/enemy-level.test.ts` (one) — so every typed `IEnemy` under `src/tests/`
+   now builds through here. Reintroducing an `as IEnemy` cast re-opens that hole; state why if you do. */
 
 /**
  * Builds an {@link IEnemy} reference-data entry for tests. Everything but the id is a neutral
--- a/UI/src/tests/fixtures/item-mods.ts
+++ b/UI/src/tests/fixtures/item-mods.ts
@@ -5,12 +5,15 @@ import { EItemModType, ERarity, type IItemMod } from '$lib/api';
    that drift independently, even though a suite exercising applied mods usually imports both.
 
    The target set was confirmed with the throwaway-required-field probe, not a `grep` — only a suite
-   that constructs a *typed* `IItemMod` pays the contract-drift tax. Exclusions are deliberate, in two
-   classes: the production blank record (`routes/admin/workbench/entities/item-mod.ts`) and the live
+   that constructs a *typed* `IItemMod` pays the contract-drift tax. Two exclusions are deliberate and
+   permanent: the production blank record (`routes/admin/workbench/entities/item-mod.ts`) and the live
    `ItemMod` domain class (`lib/battle/item-mod.ts`) are the authoritative shapes and correctly stay
-   literals; `challenges/OverviewPane.test.ts` and `lib/common/challenge-unlocks.test.ts` each assert a
-   deliberately partial literal `as IItemMod`, which opts them out of drift detection entirely (#2447).
-   A `grep` still finding mod literals under `src/tests/` is expected, not a gap. */
+   literals. A `grep` still finding mod literals under `src/tests/` is expected, not a gap.
+
+   A third class was invisible to the probe rather than excluded: a literal asserted `as IItemMod`
+   keeps compiling when the contract gains a required field. #2447 folded those in
+   (`challenges/OverviewPane.test.ts`, `lib/common/challenge-unlocks.test.ts`); reintroducing such a
+   cast re-opens the hole, so state why if you do. */
 
 /**
  * Builds an {@link IItemMod} reference-data entry for tests. Everything but the id is a neutral
--- a/UI/src/tests/fixtures/items.ts
+++ b/UI/src/tests/fixtures/items.ts
@@ -3,12 +3,17 @@ import { EItemCategory, ERarity, type IItem } from '$lib/api';
 /* Shared `IItem` contract builder (#2433), following the `fixtures/skills.ts` convention.
 
    The target set is narrower than a `grep` for item-shaped literals suggests: only a suite that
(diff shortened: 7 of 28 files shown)


UI/src/tests/routes/select/ClassPicker.test.ts
```diff
@@ -7,6 +7,7 @@ import ClassPicker from '$routes/select/ClassPicker.svelte';
 import { staticData } from '$stores/static-data.svelte';
 import { EEquipmentSlot, EModifierType, type ICreatableClass } from '$lib/api';
 
+/* Deliberately partial: `ICreatableClass` has no shared fixture yet (#2456). */
```
ginderjeremiah, line 10: **Should-fix** — there is a third `ICreatableClass` cast builder that got neither outcome:

```ts
// src/tests/routes/select/player-select-view.test.ts:15-22
const creatable = (id: number): ICreatableClass =>
	({ id, name: `Class${id}`, starterSkills: [], starterEquipment: [], attributeDistributions: [] }) as unknown as ICreatableClass;
```

Same shape as this one and as `class-summary`'s — annotated `ICreatableClass`, read as one by the code under test, cast silencing the missing fields. The `as unknown as` form isn't a reason to skip it; the pass converted `as unknown as ISkill` (`inventory-view`) and `as unknown as IAttribute` (`SkillDamageBreakdown`, `class-summary`) on exactly that reasoning.

It needs the same `#2456` note, and #2456's "Also worth folding in" section should say **three** sites rather than "Two sites, one contract" — otherwise whoever picks that up works from a two-site list and leaves this one casting. That's the same failure mode the `enemy-level.test.ts` correction on #2447 called out: a pass driven by an incomplete list converts what's listed and leaves the rest.

Worth fixing here rather than as a follow-up — it's one comment line plus an edit to #2456's list, and the bookkeeping is the deliverable for these deferred sites.


---

UI/src/tests/routes/admin/workbench/challenge/ChallengeRewardSection.test.ts
```diff
@@ -48,6 +48,7 @@ const config = (): EntityConfig<Identified> =>
 	}) as unknown as EntityConfig<Identified>;
 
 const setup = (over: Partial<IChallenge> = {}) => {
+	/* Deliberately partial: `IChallenge` has no shared fixture yet (#2456). */
```
ginderjeremiah: **Nit** — this note sits inside `setup()`, so it reads as covering only the record on line 52. The file has two more `as IChallenge` literals outside it (lines 196-197, `first` / `second`), which #2456 does count ("three literals") but which look uncommented to a reader scanning for the marker.

Either hoist this to a file-level comment above `setup()`, or drop a one-liner at 196. Same applies to `GoalField.test.ts`, which has a second literal at line 75.


---

ginderjeremiah (commented):
## Review: fold cast-built contract literals back into drift detection (#2447)

**Intended verdict: Approve**, with one should-fix worth landing before merge. Submitted as a `COMMENT` because GitHub won't take a gating verdict from the PR author.

The change does what the issue asks and is behaviour-preserving. I checked each conversion's fixture defaults against the literal it replaced rather than trusting the green suite: `makeZone`'s `levelMin: 1 / levelMax: 10 / bossLevel: 1` match the `challenge-unlocks` / `ChallengeTooltip` builders field-for-field, `makeEnemy`'s `Enemy ${id}` / `isBoss: false` / empty collections match both deleted `enemy-attributes` builders, and `makeItem`'s `Item ${id}` matches `skill-provenance`'s. The fields the fixtures newly populate (`designerNotes`, `isHome`, `iconPath`, `tags`) are unread at every converted site.

### Should-fix

- `routes/select/player-select-view.test.ts:15` — a third `ICreatableClass` cast builder that got neither outcome 1 nor outcome 2, and isn't in #2456's inherited list. Details inline on `ClassPicker.test.ts:10`.

### Nits

- `ChallengeRewardSection.test.ts:51` (and `GoalField.test.ts`) — the `#2456` note is scoped to `setup()` while the file carries further cast literals outside it. Inline.
- The closing sentence of the new hazard paragraph is verbatim across all six fixture headers. The issue explicitly asked for the per-contract comment to name this class, so the repetition is what was requested — just flagging that a future policy change is six synchronized edits. Fine to leave.

### On the `IChallenge` deferral

The pushback in the PR body is right, and I'd have made the same call. Outcome 1 presumes a fixture to convert *onto*; there isn't one, and hand-writing seven required fields across nine sites re-creates precisely the duplication the series exists to remove. #2456 is correctly scoped as a #2446-shaped consolidation rather than a cast cleanup, and it inherits a concrete site table instead of a vague pointer. Going the other way on `IChallengeType` — one site, three fields, a complete literal already in the same suite — is the right asymmetry; a fixture module there would have been ceremony.

### Test coverage

Test-only refactor, no production code touched, so no new coverage is owed. The right acceptance check is the drift probe itself, and "a throwaway required field on all six contracts yields exactly six errors under `src/tests/`, one per fixture module and none at any test site" is the correct form of it — it verifies the guarantee the earlier PRs only *stated*. Confirmed by inspection that no `as I<Contract>` literal remains for any of the six fixture-backed contracts; the survivors are `[] as IX[]` annotations and `IApiSocketResponse` casts, neither of which is in scope.

### What's good

- The completeness work goes well past the issue's table, which the "known" caption invited: `ISkill` was absent from it entirely (five sites), plus four more `IItem`/`IItemMod` and two more `IAttribute` sites. Reconciling the table in an issue comment rather than silently exceeding it is the right record.
- The two consolidations that fell out are real dedup, not drive-by churn: `enemy-attributes` was carrying two `IEnemy` builders where the leaner one duplicated the shared fixture exactly, and `inventory-view` had the *same* `makeSkill` defined twice in two describes. Keeping `damageSkill`'s positional signature so 16 call sites don't churn is the right restraint.
- `class-summary` swapping bare `1` / `2` for `EAttribute.Endurance` / `Intellect` is a strict fidelity gain, and it's safe because `attributeCode` resolves by `find(a => a.id === id)` rather than by index — the ids can move to their real enum values without disturbing the lookup.
- Good catch on `offline-summary`'s `itemCategoryId: 0`, which is not a member of `EItemCategory` (it starts at `Helm = 1`). Taking the fixture's neutral default instead of inventing a category is the right resolution for a value nothing asserts on.


---

UI/src/tests/routes/select/ClassPicker.test.ts
ginderjeremiah: Correct, and the diagnosis of *why* is the useful part: the first sweep grepped `as ICreatableClass` for this contract, so the `as unknown as` form was invisible to it. That is precisely the failure mode you name — a pass driven by an incomplete list — and it's the same one the `enemy-level.test.ts` correction called out on #2447, which makes it worth more than a one-line patch.

So rather than fix just the named site, I re-swept for **both** forms across all of `src/tests/`. That turned up two more contracts the bookkeeping never mentioned:

- **`ITag`** (`UsageSection.test.ts:29`) — `{ id: 5, name: 'Fire', tagCategoryId: 10 } as ITag`. `ITag` has exactly those three required fields, so the literal was already complete and **the cast was pure redundancy**. Dropped it: the site rejoins drift detection with no fixture and no follow-up, same resolution as `IChallengeType`.
- **`IPlayerData`** (`player-select-view.test.ts:13`, right above the builder you flagged) — genuinely can't be completed inline: 15 required fields including nested aggregates (`attributes`, `inventoryData`, `lessons`, `signaturePassive`). It stays cast, but now states its own reason rather than pointing at #2456 — one test builder for one consumer, where a fixture module would be ceremony. It'd be misleading to park it in a consolidation issue that will never have a second site to consolidate.

`player-select-view`'s `ICreatableClass` builder now carries the `#2456` note, and #2456's section is retitled **"Also worth folding in: `ICreatableClass` (three sites)"** with a table naming all three and the form each uses — plus a line recording that the third was missed by a single-form grep, so the next pass doesn't repeat it. `IPlayerData` is listed there under "Explicitly not in scope" with its reasoning, so it doesn't read as an oversight either.

Nit also taken: the notes in `ChallengeRewardSection` and `GoalField` are hoisted out of `setup()` to file scope and reworded to "Every `as IChallenge` literal in this file…", which covers the literals at ~196 and ~75 without a second marker.

Pushed as `daff93c`. Lint/`svelte-check` clean (1227 files), 4046 tests passing.

---

ginderjeremiah:
## Re-review of `daff93c` — **Approve**, merging

Incremental pass over the follow-up commit only; the first review's verification of the conversions still stands.

### The should-fix is resolved, and resolved better than asked

I asked for one comment line on `player-select-view.test.ts` plus a correction to #2456's list. Re-sweeping for **both** cast forms instead of patching the named site was the right response — the diagnosis was that a single-form grep caused the miss, so a single-site patch would have left the same failure mode in place. It turned up two more contracts:

- **`ITag`** (`UsageSection.test.ts:29`) — verified: `ITag` has exactly `id` / `name` / `tagCategoryId`, all three were already supplied, so the cast was pure redundancy. Dropping it is the right resolution, same as `IChallengeType`.
- **`IPlayerData`** (`player-select-view.test.ts:16`) — agreed it stays cast, and agreed it should *not* point at #2456. A consolidation issue that will never have a second site to consolidate is the wrong home for it; the self-contained reason is better bookkeeping than a misleading pointer.

The `ChallengeRewardSection` / `GoalField` nit is taken — the notes are hoisted to file scope and reworded to "Every `as IChallenge` literal in this file…", which covers the literals at ~196 and ~75 without a second marker.

### Verified independently

- **No cast remains for any fixture-backed contract.** Swept both forms across `src/tests/`: every surviving `as I<Contract>` literal is `IChallenge` (nine sites, all noted → #2456), `ICreatableClass` (three sites, all noted → #2456), or the one `IPlayerData` builder with its own stated reason. Nothing is silently outside the guarantee.
- **Every site the bookkeeping claims carries a note actually carries one** — including `RewardPicker`, `ScopeField`, `ChallengeConditionSection`, `ZoneNav` and `offline-summary`, which the follow-up commit didn't touch.
- **#2456 matches the code.** The "Also worth folding in" section is now three sites with the form each uses, plus the note recording *why* the third was missed; `IPlayerData` is listed under "Explicitly not in scope" with its reasoning. Someone picking up #2456 from the issue alone gets a complete list.
- **CI is green at head** (`all-checks-passed`; frontend unit, verify, and all three e2e browsers).

### One observation, not a change request

The hazard paragraph is now near-verbatim across six fixture headers, and `fixtures/items.ts` carries a 13-line block before its JSDoc. The content is load-bearing — "a grep still finding literals here is expected, not a gap" is exactly the thing a future maintainer would otherwise re-derive — and the issue explicitly asked for the per-contract comment, so this is what was requested rather than drift. Flagging only that a future policy change is six synchronized edits.

No outstanding nits or should-fixes, and no follow-up work owed beyond #2456, which already exists and is correctly scoped. Merging.

---

Issue #1155, opened 2026-06-22
Use custom serialization model for Player AggregateRoot


ginderjeremiah:
Right now, the `Player` AggregateRoot is directly serialized to and deserialized from the cache as a json string.

I would like to exclude reference data from the serialization process (e.g., Skills, Items, Item Mods) that are normally fully loaded into the domain object to prevent outdated data from being loaded from the cache. This will require a transformation when setting the state and when rehydrating it into the domain model.

Issue #488, opened 2026-06-13
PlayerStatPoints.TryUpdateAttributes silently ignores allocations to attributes without an existing row


ginderjeremiah:
## Summary

`PlayerStatPoints.TryUpdateAttributes` (`Game.Core/Players/PlayerStatPoints.cs:37-57`) joins the incoming `changedAttributes` against the player's **existing** `StatAllocations`. Any update whose `Attribute` has no pre-existing allocation row is never matched, never validated, never applied — yet the method still returns `true`.

Today new players are seeded with all 6 core attributes (`NewPlayerFactory.AttributeCount`), so it happens to work, but the contract is fragile: a request to allocate into an attribute the player has no row for reports success while doing nothing, and correctness is coupled to the invisible invariant "every allocatable attribute always pre-exists as a row."

## Impact

This is anti-cheat-adjacent state mutation (the `UpdatePlayerStats` socket command). A "success" response with no state change is a silent-failure class bug. Distinct from #451 (which concerns enumeration/allocation cost in the same method).

## Suggested fix

Either reject updates referencing an unknown attribute (return `false`) or create the allocation row on demand. Add a unit test for the unknown-attribute case.

PR #2465, opened 2026-07-26
Store zero-amount attribute allocations instead of deleting their rows (#2459)

### LINKED ISSUE #2459: AttributeAllocationsChangedHandler deletes zero-amount allocation rows, permanently blocking those stats after a DB fall-through reload

## The bug

`NewPlayerFactory` seeds an `Amount = 0` allocation row for every core attribute precisely because `PlayerStatPoints.TryUpdateAttributes` rejects any allocation into an attribute with no row (`Game.Core/Players/NewPlayerFactory.cs:61-71`, `Game.Core/Players/PlayerStatPoints.cs:40-43`, the #488 anti-cheat). But the write-behind handler **deletes** any existing row whose incoming amount is `0` and skips inserting zero-amount rows (`Game.DataAccess/PlayerUpdates/Handlers/AttributeAllocationsChangedHandler.cs:44-47,53`), while `AttributeAllocationsChangedEvent` carries the **full** allocation list, zeros included (`Game.Core/Players/Player.cs:98-100`).

## Failure scenario (no race required)

1. New player created — one `PlayerAttribute` row per core attribute, `Amount = 0`.
2. Player allocates their first point into Strength. The event carries `[Str=1, Dex=0, Int=0, …]`; the drain handler updates Str and **deletes every other seeded row**.
3. Player idles past the 48h sliding TTL (or the blob is evicted/Redis restarts) → `PlayerRepository.GetPlayer` falls through to the DB → `StatAllocations = [Str=1]` only (`Game.DataAccess/Repositories/PlayerQueryExtensions.cs:39-41` copies rows as-is; nothing reseeds missing core attributes).
4. Every subsequent allocation into Dexterity/Intellect/… is rejected as anti-cheat (`missing row → return false`). No path ever re-creates the rows — later events carry only the surviving allocations — so the block is **permanent**, and the reduced list is re-cached to Redis so it sticks.

This hits essentially every player who allocates points and later ages out of Redis.

## Suggested fix

Either upsert zero as a stored `0` (matching the progress tier's "a stored 0 is a genuine recording" convention), or reseed missing core rows in the single rehydration path (which also makes existing damaged rows self-heal). If the handler is fixed without reseeding, existing DBs need a one-time repair.

The integration tests (`PlayerUpdateHandlerIdempotencyIntegrationTests`) never exercise a zero-amount entry against the seeded-row invariant — a full allocation write-behind → DB-reload → aggregate round trip test would pin this.

ginderjeremiah:
## Summary

`AttributeAllocationsChangedHandler` deleted any `PlayerAttribute` row whose incoming amount was `0`, and skipped inserting zero-amount rows. But `AttributeAllocationsChangedEvent` carries the player's **whole** allocation list, zeros included — so the player's first allocation into any stat deleted every other seeded row. Once that player aged past the 48h sliding TTL (or Redis restarted) and fell through to a DB reload, `PlayerStatPoints.TryUpdateAttributes` rejected every subsequent allocation into those attributes as anti-cheat (#488 — no row, no allocation), and nothing ever re-created them, so the block was **permanent**.

I verified the issue's claim end-to-end before fixing it; it's accurate, including the "no race required" failure path.

**1. The handler upserts absolutely, zeros included.** Unlike the progress tier's statistics — where row *absence* is the "no data yet" state and a stored `0` is a genuine recording — an allocation row's *presence* is what makes its attribute allocatable at all. So the delete branch and the `amount != 0` insert guard are both gone; a zero is the seeded state, not an empty one.

**2. Rehydration re-establishes the seeded row set,** so no one-time DB repair is needed. `PlayerCacheMapper.ToCore` (the single rehydration path for the aggregate, #1155) builds its stat points through the new `PlayerStatPoints.Rehydrate`, which fills a zero row for any missing core attribute and leaves existing rows untouched. Characters already damaged by the old handler self-heal on their next load, and the repaired zeros reach the DB on their next allocation. This also fixes a latent second case the seeded-row comment in `NewPlayerFactory` called out but only covered for *new* players: an existing character now gains an allocation row for a core attribute added after they were created.

**3. One source of truth for the seeded set.** `NewPlayerFactory` and the rehydration path now both build it from `PlayerStatPoints`, so creation and reload can't disagree about which attributes a player can allocate into.

The repair only fills gaps — it never prunes — so a stored non-core row (which shouldn't exist) is preserved rather than silently discarded. Adding zero-amount allocations is parity-safe: they compose as `+0.0` additive modifiers, and a new player already carries all six from creation, so this restores the normal state rather than introducing a new one.

No frontend change: the attributes screen already renders all six steppers from its own `CORE_ATTRIBUTES` list, so a damaged player saw the stepper and got a server rejection on click. Battle logic isn't touched, so there's no parity surface here to mirror.

## Test plan

New coverage:
- **`PlayerStatPointsTests`** (unit) — `CreateAllocations` grants a zero row per core attribute in canonical order and returns independent rows per call; `Rehydrate` restores missing rows at zero, preserves existing amounts, adds nothing to an already-complete set, preserves a non-core row, and — the point of the repair — makes a previously blocked stat accept an allocation again.
- **`PlayerCacheMapperTests`** — `ToCore` on a model holding only the allocated stat rehydrates the full seeded set.
- **`PlayerUpdateHandlerIdempotencyIntegrationTests`** — a zero amount creates *and* keeps the row (both the insert and update branches); an allocation refunded back to zero leaves the row in place.
- **`PlayerServiceIntegrationTests`** — the full round trip the issue asked for: allocate → write-behind drain → Redis key deleted → DB fall-through reload, asserting every core attribute is persisted and that the reloaded aggregate can still spend into a stat it has never allocated into.
- **`UpdatePlayerStatsSocketTests`** — a core attribute with no persisted row is allocatable over the real socket command.

Two existing tests encoded the old behavior and were updated rather than deleted:
- `DataProviderSynchronizerTests.ProcessQueue_AttributeAllocationsChangedEvent_*` — the zeroed attribute is now a stored `0` rather than a deleted row (renamed off `...AndDeletes...`).
- `UpdatePlayerStatsSocketTests` #488 rejection test — its premise (a core attribute with no row) is no longer reachable, so it now targets a **derived** attribute, which is the case the anti-cheat actually exists for per `backend.md`.

Verification:
- I confirmed all four new integration/mapper tests **fail** against the unfixed handler + mapper before passing with them.
- Full backend suite green: `dotnet build` + `dotnet test` — 3,533 passed, 0 failed (Core 1475, Infrastructure 74, Application 1109, Api 875).

## Docs

Added a short paragraph to `docs/backend-persistence.md`, directly after the statistics row-presence convention it inverts, so the next author doesn't carry the statistics rule across to attribute allocations — that's exactly the reasoning that produced this bug.

## Review follow-ups (ca8a492)

Both nits from the review are addressed:
- **`PlayerStatPoints.Rehydrate` factory.** The repair was previously enforced by remembering to call it at the construction site, so a second load path that forgot the line would silently reintroduce this bug. The factory now owns building a persisted player's stat points and seeds the missing rows internally, making it unbypassable and keeping `ToCore` a pure translation rather than construct-then-mutate.
- **Narrowed the ordering claim** on `AllocatableAttributes` instead of adding a canonicalization pass. Restored rows are appended, so a repaired character's allocations aren't ordered like a freshly created one's — but nothing reads meaning into that order, and sorting would need to invent answers for duplicate core rows and for ranking the non-core rows the repair deliberately preserves. The doc now claims only the deterministic *seed* order it actually provides, and `Rehydrate` documents the append.

## Follow-up

Filed #2466: `TryUpdateAttributes` uses row presence as a proxy for "is this attribute allocatable", when the authoritative answer is `Attribute.IsCore`. With the invariant now enforced at both creation and rehydration the two are equivalent, but the proxy is the fragile coupling that made a persistence bug become a permanent gameplay lockout.

Closes #2459

https://claude.ai/code/session_013uHdALADQ5snb35vDieFcW


Game.Core/Players/PlayerStatPoints.cs
```diff
@@ -1,10 +1,19 @@
...
     /// Represents the stat points a player can and has allocated to core attributes.
     /// </summary>
     public class PlayerStatPoints
     {
+        /// <summary>
+        /// The attributes a player carries an allocation row for — the core (directly-allocatable) set, in
+        /// the enum's declaration order so every player's allocations read in the same canonical order.
+        /// </summary>
+        private static readonly EAttribute[] AllocatableAttributes =
+            [.. Enum.GetValues<EAttribute>().Where(CoreAttribute.IsCore)];
```
ginderjeremiah: **Nit:** the justification here — "so every player's allocations read in the same canonical order" — only holds for `CreateAllocations`. `EnsureAllocatableAttributesArePresent` *appends* the restored rows after whatever survived, so a repaired player ends up with a different order than a freshly created one (e.g. `[Strength, Endurance, Intellect, Agility, Dexterity, Luck]` vs `[Strength, Intellect, Agility, Dexterity, Luck, Endurance]`).

That's harmless today as far as I can tell — the frontend renders its steppers from its own `CORE_ATTRIBUTES` list, and the allocations only reach battle math as `+0.0` additive modifiers off a snapshot both tiers share — but the comment currently promises an invariant the new repair path doesn't maintain. Either sort the list at the end of the repair, or narrow the claim to the creation path so the next reader doesn't rely on it.

---

Game.DataAccess/Mapping/PlayerCacheMapper.cs
```diff
@@ -161,6 +161,17 @@ public static Player ToCore(PlayerCacheModel model, IItems items, IItemMods item
...
+            // Re-establish the seeded allocation row per core attribute. Row presence is what makes an attribute
+            // allocatable, so restoring it here (rather than only at creation) is what lets a character whose
+            // rows were dropped, or one predating a newly-added core attribute, allocate into it again.
+            var statPoints = new PlayerStatPoints
+            {
+                StatAllocations = model.StatAllocations,
+                StatPointsGained = model.StatPointsGained,
+                StatPointsUsed = model.StatPointsUsed,
+            };
+            statPoints.EnsureAllocatableAttributesArePresent();
```
ginderjeremiah: **Nit:** the invariant is now enforced by *remembering to call* the repair at the construction site. `ToCore` is the only production `new PlayerStatPoints` today (I checked — `PlayerQueryExtensions` and the Redis blob both funnel here), so it's correct as written, but a second rehydration path that forgets this line silently reintroduces #2459. A `PlayerStatPoints.Rehydrate(model.StatAllocations, model.StatPointsGained, model.StatPointsUsed)` static factory that seeds the missing rows internally would make it unbypassable, and would keep the mapper a pure translation rather than construct-then-mutate.

Worth weighing against #2466 though: if `TryUpdateAttributes` moves to `Attribute.IsCore`, row presence stops being load-bearing and this becomes belt-and-braces, so hardening it now may be work that gets undone. Fine to leave as-is and let #2466 settle it — flagging so the choice is deliberate.

---

ginderjeremiah (commented):
## Review: store zero-amount attribute allocations instead of deleting their rows (#2459)

**Verdict:** Approve — the fix is correct at both ends (handler stops deleting, rehydration self-heals), and the coverage is genuinely proportionate to the bug. Submitted as a `COMMENT` only because GitHub won't accept a gating verdict from the PR's own author; treat it as an approve. Two nits below, neither blocking.

I verified the diagnosis independently rather than taking the issue at face value:

- `PlayerQueryExtensions.SelectPlayerCacheModel` → `PlayerCacheMapper.ToCore` really is the only production rehydration path (`grep` for `new PlayerStatPoints` returns exactly one non-test hit, at `PlayerCacheMapper.cs:167`), so putting the reseed there does cover both the DB fall-through and the Redis-blob load.
- The point-accounting stays consistent under the repair: the old handler only ever deleted a row whose amount was `0`, so a damaged player never lost allocated points — restoring the row at `0` is exactly the state that was dropped, not a guess.
- Parity is genuinely untouched. `StatAllocation.ToModifier` is `EModifierType.Additive`, so a restored row contributes `+0.0`, and a new player already carries all six — this restores a normal state rather than introducing a new one. No mirrored frontend test is owed here.
- `RedisService.SetAndForget<T>` serializes on the calling thread before firing the command, so mutating `model.StatAllocations` in `ToCore` afterwards can't race the in-flight cache write. Worth knowing since the repair mutates a model that was just handed to a fire-and-forget write.

### Nits

- `Game.Core/Players/PlayerStatPoints.cs:11-15` — the "same canonical order" claim in the field's doc doesn't hold for a repaired player; restored rows are appended, not merged in enum order. Sort at the end of the repair or narrow the comment.
- `Game.DataAccess/Mapping/PlayerCacheMapper.cs:167-173` — the invariant is enforced by remembering to call `EnsureAllocatableAttributesArePresent` at the construction site. A `PlayerStatPoints.Rehydrate(...)` factory would make it unbypassable, but #2466 may make it moot; flagging so it's a deliberate choice rather than an oversight.

### Test coverage

Meets the project's rules, and then some. Domain logic is unit-tested at the edges (`CreateAllocations` independence per call is the kind of test that catches a future `static readonly` "optimization"); the write-behind branches are covered at the integration tier where the Postgres interaction lives; and the `PlayerServiceIntegrationTests` round trip — allocate → drain → delete the Redis key → DB fall-through → allocate into a never-touched stat — is precisely the end-to-end pin the issue asked for, with the cache-key delete making the fall-through genuine rather than incidental.

The two behavior-encoding tests were updated rather than deleted, which is the right call. Retargeting the #488 rejection test from `Luck` to `MaxHealth` is the notable one: the old premise (a core attribute with no row) is unreachable after this change, and a derived attribute is the case the anti-cheat actually exists for per `backend.md`. The alternative — deleting the test — would have quietly dropped coverage of a security-adjacent rule.

### What's good

- Fixing the handler *and* reseeding on rehydration, rather than picking one. The issue offered them as alternatives ("if the handler is fixed without reseeding, existing DBs need a one-time repair"); doing both is what makes damaged characters self-heal with no migration.
- Folding `NewPlayerFactory` onto `PlayerStatPoints.CreateAllocations` so creation and reload can't disagree about the seeded set — that's the actual DRY win, and it incidentally closes the "core attribute added after a character was created" gap the old comment described but only handled for new players.
- The `backend-persistence.md` paragraph is placed directly after the statistics row-presence convention it inverts. That's the right spot: carrying the statistics rule across to allocations is the exact reasoning that produced this bug, and the doc now interrupts it.
- Filing #2466 instead of expanding scope. Row presence as a proxy for `IsCore` is the real fragility here, and it's correctly scoped out of a bug fix.

---

ginderjeremiah:
Both nits addressed in ca8a492. Thanks for independently re-deriving the diagnosis rather than taking the PR body's word for it — the `SetAndForget<T>` serialization-order point in particular was something I'd assumed rather than checked, and it's good to have it written down, since the repair does mutate a model that was just handed to a fire-and-forget write.

**Ordering claim** — narrowed rather than sorted. You're right that the comment promised something the repair path doesn't deliver. I went with narrowing because making the claim *true* is more expensive than it looks: a canonicalization pass has to decide what to do with a duplicate core row (silently dropping one is a worse behavior change than an inconsistent order) and where to rank a non-core row that the repair deliberately preserves. Both are edge cases invented purely to satisfy a doc comment, given that — as you note — nothing reads meaning into the order. So `AllocatableAttributes` now only claims what it actually provides (a deterministic *seed* order, and the reason it's an enum-ordered array rather than the unordered `CoreAttributes` set), and `Rehydrate` documents the append explicitly, including why it's safe.

**`Rehydrate` factory** — took it. Your own argument against it is the weaker one, I think: even if #2466 moves `TryUpdateAttributes` to `Attribute.IsCore` and row presence stops being load-bearing for the anti-cheat, the seeding itself doesn't go away (`PlayerData.Attributes` still projects straight off the list), so this isn't work that gets undone — it just downgrades from correctness-critical to belt-and-braces. And "unbypassable" beats "correct as long as the next person remembers" for a bug whose entire severity came from a silent invariant break. `PlayerStatPoints.Rehydrate(allocations, statPointsGained:, statPointsUsed:)` now owns the repair, the standalone method is no longer public, and `ToCore` is back to pure translation.

The unit tests moved onto the factory rather than the removed method, which is the better contract to pin anyway. Full backend suite still green: 3,533 passed, 0 failed.

---

Issue #2467, opened 2026-07-26
Cross-instance stale-absolute-overwrite in the write-behind drain needs per-player event sequencing


ginderjeremiah:
Found while fixing #2460. That fix closes the **within-instance** half of the hazard by remembering which players an instance left parked on the processing list and deferring their later events until the reclaim replays the parked one in order. The **cross-instance** half is still open, and the existing "cross-instance same-player ordering is best-effort" framing in `docs/backend-persistence.md` understates it.

## Why "best-effort" isn't enough here

The doc's tolerance argument is about *convergent* reorderings, and it holds for the cases it names:

- A duplicate apply of the same event is idempotent (existence checks, absolute updates, delete-then-insert).
- A child-row handler reordered ahead of the unlock it presupposes inserts the missing row rather than no-op'ing.

Neither covers **a stale absolute value applied after a newer one**. `PlayerCoreUpdatedHandler`, `ProgressUpdatedHandler`, `SelectedSkillsChangedHandler`, `ItemEquippedHandler` and `LogPreferenceChangedHandler` all write absolute state, so "older wins" is a durable regression, not a transient one that converges.

## The remaining scenario

1. Instance A parks event `E1` for player P (exp=100) on the shared processing list — a DB-infrastructure failure, or an acknowledge that faulted after a durable apply.
2. P's next save enqueues `E2` (exp=150). Instance **B** reserves it. B has no knowledge of A's parked item — the block added by #2460 is instance-local memory — so it applies `E2` immediately.
3. Any instance later reclaims `E1` and applies it → Postgres regresses to exp=100 (or a completed challenge is un-completed, or an unequipped item is re-equipped).
4. Redis still holds the correct blob, so nothing self-corrects until P's next save dirties the same rows. A player who stops playing after the incident and stays dormant past the 48h player-key TTL gets the regressed state served from the DB fallback: durable progress loss, or a re-completable challenge (duplicate reward / re-locked zone).

This is the same failure #2460 describes, just with the newer event landing on a different instance. It needs at least two instances plus a parked item, so it is rarer than the single-instance case — but the blast radius is identical.

## Direction to evaluate (the spike)

A shared, per-player monotonic sequence is the obvious shape, but there are real design forks worth researching before committing:

[...]

PR #2468, opened 2026-07-26
Block a parked player lane across drain passes (#2460)

### LINKED ISSUE #2460: Cross-pass drain interleaving lets a reclaimed (parked) event overwrite a newer same-player event's absolute values in Postgres
Found during a scheduled codebase-health review. Distinct from #2372 (cache miss-reload behind undrained events): here the write-behind queue itself applies events out of order.

## The mechanism

Same-player ordering in the drain is enforced by per-player lanes — but `playerLanes` is **pass-scoped** (`Game.DataAccess/DataProviderSynchronizer.cs:363`), cleared when the pass settles (`:411`). An item parked by a database-infrastructure failure stays reserved on the processing list (`:700-708`), and the opportunistic reclaim (`TryReclaimStrandedProcessingAsync`, `:419,571`) moves it back to the **queue head in a later pass** (`RedisQueue.ReclaimProcessingAsync`). `backend-persistence.md`'s claim that "a failed item's lane stays blocked so later same-player items wait for the reclaim" only holds within one pass.

## Interleaving

1. DB outage: event A for player P (e.g. `PlayerCoreUpdatedEvent` exp=100, or a `ProgressUpdated` with a challenge un-completed) exhausts retries, classifies as an infrastructure failure, and is parked on the processing list. The pass ends; lanes are cleared.
2. DB recovers. P's next save enqueues event B (exp=150 / challenge completed). The wake starts a **fresh pass**: B gets a fresh lane and applies immediately — A is still parked (the reclaim needs the ≥30s dwell across two empty-queue observations, or the 5-minute reconciliation).
3. A later pass reclaims A and applies it → Postgres regresses to exp=100 / challenge un-completed. Affected handlers are the absolute-value ones: `PlayerCoreUpdatedHandler`, `ProgressUpdatedHandler`, `SelectedSkillsChangedHandler`, `ItemEquippedHandler`, `LogPreferenceChangedHandler`.
4. The Redis blob is still correct, so nothing self-corrects until P's next save dirties the same rows. If P stops playing right after the outage (plausible — the outage was user-visible) and stays dormant past the 48h TTL, the DB fallback serves the regressed state: durable progress loss, or a re-completable challenge (duplicate reward / re-locked zone).

The at-least-once/idempotent-handler contract makes a *duplicate* apply convergent, but stale-absolute-**after**-newer is not.

## Suggested containment

When the infrastructure-outage flag is set, run the stranded-processing reclaim **before** reserving new items in the next pass (mirroring startup's reclaim-then-drain order), so recovered backlog re-queues ahead of post-recovery events on the same instance. Update the lane-blocking claim in `backend-persistence.md` alongside the fix.

ginderjeremiah:
Closes #2460.

## Summary

Same-player ordering in the write-behind drain was enforced only by the per-pass `playerLanes` map, which `SettleInFlightItemsAsync` clears when a pass ends. An event parked on the processing list by a database-infrastructure failure therefore stopped blocking its player the moment that pass finished: the next pass gave a newer same-player event a fresh lane and applied it immediately, and the later opportunistic reclaim then replayed the older parked event over the top — regressing exp/level, un-completing a challenge, or re-equipping an item. The Redis blob stayed correct, so nothing self-corrected until the player's next save; a player who went dormant past the key TTL lost the progress durably.

`DataProviderSynchronizer` now keeps an instance-level record of the players it left an item reserved for (`_parkedPlayerLanes`, written from every fault path that leaves an item on the processing list — the item's own apply, its acknowledge, or its predecessor on the same lane). While a player is listed, its newly reserved items are **deferred**: left reserved and unapplied rather than started on a fresh lane. Because `ReclaimProcessingAsync` moves the processing list back onto the queue head in order, the reclaim then re-queues the whole same-player chain ahead of anything newer and it applies oldest-first.

The block is released once a pass observes the processing list empty, or once a reclaim has drained it — both inside `TryReclaimStrandedProcessingAsync`, which only runs under the serialized drain after the pass has settled its own in-flight items. So a player is never deferred indefinitely (bounded by the reclaim grace period plus the reconciliation sweep), and an item another instance reclaimed and applied on our behalf clears the block too.

## Why not the containment suggested in the issue

The issue proposed running the stranded-processing reclaim *before* reserving new items whenever the outage flag is set. I went a different way, for two reasons:

1. **The grace period makes it insufficient on its own.** A reclaim only fires once the processing head has dwelled past `_reclaimGracePeriod` with nothing left to reserve. The post-recovery event typically arrives seconds after the outage, so the pre-reserve reclaim would decline to reclaim and the newer event would still apply first. Making it bypass the grace period would reclaim items other instances are genuinely still applying — exactly what that grace period exists to avoid.
2. **The outage flag is the wrong key.** `_infrastructureOutageLogged` clears after `_successesRequiredToClearOutage` consecutive successes, which other players' events can supply while an item stays parked. Keying the containment on it would silently miss those cases.

Blocking the specific player's lane instead keys off the thing that actually matters (an item this instance left reserved), needs no change to reclaim timing, and makes the guarantee `docs/backend-persistence.md` already claimed genuinely true. It also incidentally reduces DB hammering during an outage: subsequent same-player events are deferred rather than each burning a full retry budget.

The cost is convergence latency for the affected player — its DB rows lag until the reclaim replays the chain. Redis remains the source of truth for reads and rehydration throughout, so this is lag, never a lost write, and it replaces a durable regression.

## Test plan

- [x] New `ProcessQueue_LaterPassAfterOutage_DefersSamePlayerEventUntilTheParkedOneIsReplayed` — the issue's exact interleaving: an outage parks exp=100, a fresh pass sees exp=150 and must not apply it, and after the grace period both replay oldest-first so exp=150 survives.
- [x] New `ProcessQueue_LaterPassAfterOutage_StillAppliesOtherPlayersEvents` — the block is per player, not a queue-wide stall.
- [x] New `ProcessQueue_ParkedItemAcknowledgedElsewhere_UnblocksThePlayerOnTheNextReclaim` — a stale entry can't defer a player forever when another instance applied the parked item.
- [x] Verified all three fail against the unmodified synchronizer (reverted the production file, reran: 3 failed) and pass with it.
- [x] Two existing outage-logging tests reused one player across the outage boundary and now correctly defer the post-recovery events — they were in fact encoding this regression, asserting a level the parked event would later overwrite. Updated to use a separate player per stage, since they cover log-flood suppression rather than ordering.
- [x] `dotnet build` clean (0 warnings), full `dotnet test`: 3525 passed, 0 failed.

No frontend changes — this is the backend drain only, no battle logic involved.

## Docs

`docs/backend-persistence.md`'s lane-blocking claim ("a failed item's lane stays blocked so later same-player items wait for the reclaim") only held within one pass, which the issue calls out. Updated to describe the cross-pass block, its release conditions, and its latency cost, and to state that this is an instance-local guarantee.

## Follow-up

Filed #2467 for the remaining cross-instance half: the block is instance-local memory, so a newer event reserved by a *different* instance can still apply ahead of another instance's parked one. That needs per-player event sequencing the absolute-value handlers can reject stale writes against, which is a materially larger design change (schema plus every handler) than this fix — the issue lays out the design forks to research.

---
_Generated by [Claude Code](https://claude.ai/code/session_015VS7KFKXUcUiRnhbWp2Jdb)_


Game.DataAccess/DataProviderSynchronizer.cs
```diff
@@ -426,8 +443,24 @@ private async Task DrainQueueAsync(IPubSubQueue queue, CancellationToken cancell
...
+                        // of the older event's eventual replay and then be regressed by it (#2460). Leave this
+                        // item reserved (neither applied nor acknowledged) instead: the reclaim re-queues the
+                        // whole same-player chain to the queue head in its original order, so the deferral
+                        // costs convergence latency for this one player, never the write itself.
+                        concurrencyGate.Release();
+                        _logger.LogDebug(
+                            "Deferring a player update for player {PlayerId} on queue '{Queue}': an earlier same-player event is still reserved awaiting reclaim, so this item is left reserved to preserve ordering.",
+                            playerId, Constants.PUBSUB_PLAYER_QUEUE);
+                        continue;
+                    }
```
ginderjeremiah, line 478: **Should-fix:** the deferral path releases the concurrency slot and `continue`s, so a deferred item costs nothing against the drain's budget. That silently invalidates the bounded-reservation claim this PR just re-wrapped a few lines up (`:373-376`, "a stop still ends at a bounded (not unbounded) number of in-flight items") and the matching sentence in `docs/backend-persistence.md:66` ("a backlog dominated by one hot player can't let the drain reserve unboundedly far ahead of what has actually converged").

Under a fleet-wide DB outage every player ends up in `_parkedPlayerLanes`, so a single pass will now drag the *entire* queue backlog onto the processing list at full speed — one `ReserveNextAsync` round-trip each, no bound — and `ReclaimProcessingAsync` moves the whole thing back every grace period. In-flight items stay bounded, but *reserved* items no longer are, and a stop mid-pass leaves that whole set reserved rather than ≤ `_maxConcurrentDrainItems`.

Nothing is lost (startup reclaim restores order), so this isn't blocking — but the claim shouldn't be left inaccurate. Either bound it (stop reserving for the pass once deferrals pass some threshold — the rest of the backlog gets picked up on the next pass anyway) or amend both the comment and the doc sentence to say reservation is bounded only for items that actually apply.

**Nit** on the same block: this logs one Debug line per deferred item per pass, and a blocked hot player's backlog re-defers on every reclaim cycle. Everything else in this class is deliberate about log volume (`_infrastructureOutageLogged`, `_lastReportedDeadLetterDepth`); Debug makes this tolerable, but one line per player per pass would fit the file's habits better.

---

Game.DataAccess/DataProviderSynchronizer.cs
```diff
@@ -573,6 +615,10 @@ private async Task<bool> TryReclaimStrandedProcessingAsync(IPubSubQueue queue, C
             var head = await queue.PeekProcessingAsync(1, cancellationToken);
             if (head.Count == 0)
             {
+                // Nothing is reserved anywhere on the fleet, so nothing this instance parked is still awaiting
+                // a replay — including an item another instance reclaimed and applied on our behalf, which is
+                // what stops a stale entry from deferring a player's events indefinitely.
+                _parkedPlayerLanes.Clear();
                 _strandedProcessingHead = null;
```
ginderjeremiah, line 642: **Should-fix (test coverage):** this is the second of the two documented release conditions, and no test exercises it as the thing that actually unblocks a player.

`ProcessQueue_ParkedItemAcknowledgedElsewhere_UnblocksThePlayerOnTheNextReclaim` looks like it covers this, but it doesn't: the newer event is reserved and deferred *before* the reclaim check runs, so the processing list is non-empty and the unblock comes from the `ReclaimProcessingAsync` path at `:646`, not from here. This branch is also the strictly better path — it releases without waiting out the grace period — so it's worth pinning down.

Cheap addition: park an item, acknowledge it "elsewhere", run one pass with **no** new events (processing list observed empty → block cleared here), then enqueue a newer same-player event and assert it applies on that next pass without any `timeProvider.Advance`.

---

docs/backend-persistence.md
```diff
@@ -63,7 +63,7 @@ Queue processing never silently drops an event, and the read itself is **at-leas
...
...ect stale writes against. Items on different lanes run concurrently, up to a configurable cap (default 4). The concurrency slot is acquired *before* reserving, not just before applying, so a backlog dominated by one hot player can't let the drain reserve unboundedly far ahead of what has actually converged — a stop still ends at a bounded, not unbounded, number of in-flight items, and the drain gate above still isn't released until every item the pass started (including one still applying when a stop arrives) has finished. When a pass runs out of items to reserve it settles its own in-flight items *before* consulting the opportunistic stranded-processing reclaim, since an item mid-apply legitimately sits on the processing list and must not trip the stranded clock. This directly relieves the one-round-trip-per-event ceiling on single-instance convergence throughput; it does not touch each handler's own query/`SaveChangesAsync` shape, which stays one round-trip per event (a further, separable optimization would be to batch consecutive same-player events onto one `SaveChangesAsync`, deferred as its own follow-up given the failure-isolation redesign it would need across every handler).
```
ginderjeremiah: **Nit:** the new material is genuinely worth documenting, but it's appended into a paragraph that was already the longest in the file, and it introduces a second `**bold lead-in**` mid-paragraph — every other entry in this doc is one bold lead-in per paragraph, which is what makes it skimmable.

Split from **"That block outlives the pass that created it."** onward into its own paragraph. Same content, and both the #1701 concurrency story and the #2460 cross-pass block become findable.

---

ginderjeremiah (commented):
## Review: cross-pass same-player lane blocking in the write-behind drain (#2460)

**Verdict:** Approve, with two should-fixes — the fix is correct and the design is the right one; nothing here blocks merge. (Submitted as a `COMMENT` because GitHub won't accept a gating verdict from the PR author.)

### The design holds up

I checked the premise the whole fix rests on — that the reclaim restores order — against `RedisQueueIntegrationTests.ReclaimProcessingAsync_RestoresUnacknowledgedItemsToHead_PreservingOrder` and `..._DrainsTheWholeProcessingListInOneCall_PreservingOrder`. It does, so deferring is genuinely a latency cost and not a correctness one.

Also worth recording, since it isn't obvious from the diff: the guarantee survives a **restart**, even though `_parkedPlayerLanes` is process memory. `StartAsync` reclaims the whole processing list back to the queue head *before* the startup drain (`:200`), so a parked item is re-queued ahead of any newer same-player event regardless of what this instance remembered. The only remaining hole is the cross-instance one, correctly scoped out to #2467.

Two more paths I traced that behave correctly and aren't separately tested — no action needed, noting them so they're on the record:
- **Race between the fault and a same-pass reservation.** If a later same-player item is reserved *before* the faulting task's `catch` records the block, it chains onto the faulted lane via `playerLanes` and faults in turn; if after, it's deferred. Both preserve order. This is why `CompactCompletedItems`' fault-path retention is still load-bearing rather than redundant.
- **Head-dwell liveness on a busy fleet.** A parked item is by construction the oldest un-acknowledged entry, so it *is* the processing head and the stranded clock accumulates — the block can't be held indefinitely by other instances' churn.

### Departing from the issue's suggested containment

The issue proposed reclaiming before reserving whenever the outage flag is set. Both objections in the PR description check out against the code: the grace period would make the pre-reserve reclaim decline in exactly the interleaving that matters, and `_infrastructureOutageLogged` clears on *other* players' successes (`:143-150`) so it's the wrong key. Keying on "an item this instance left reserved" is the right call.

### Should-fix

- **`Game.DataAccess/DataProviderSynchronizer.cs:447-460`** — deferred items don't consume the concurrency budget, so a pass can reserve an unbounded backlog onto the processing list. That contradicts the bounded-reservation claim in the comment this PR edited (`:373-376`) and in `docs/backend-persistence.md:66`. Bound it or correct both claims.
- **`Game.DataAccess/DataProviderSynchronizer.cs:618-622`** — the empty-processing-list release path isn't covered; `ProcessQueue_ParkedItemAcknowledgedElsewhere_...` reads like it covers it but actually exercises the reclaim path. Details inline.

### Nits

- `docs/backend-persistence.md:66` — split the new material into its own paragraph; two bold lead-ins in one paragraph breaks the file's convention.
- One Debug line per deferred item per pass (inline, folded into the first comment).
- `CompactCompletedItems`' XML doc (`:491-502`, unchanged) still describes the fault-path lane retention as the sole thing keeping a later same-player item from applying ahead of a failed one. That's now the narrow race case above, with `_parkedPlayerLanes` as the primary guard — a one-line pointer between the two would stop them reading as unrelated mechanisms.

### Test coverage

Strong. The three new tests are well chosen: the issue's exact interleaving, per-player isolation (proving this isn't a queue-wide stall), and the stale-entry-can't-strand-a-player case. Verifying they fail against the unmodified synchronizer is the right diligence.

The two existing outage tests genuinely had to change — under the new behavior their post-recovery events defer, and re-reading them, they were asserting a level the parked event would later overwrite, so they were encoding the bug. Splitting to a player per stage keeps them focused on log-flood suppression, which is what they're actually for. `succeedingPlayer` never faults, so it's never blocked and the reuse across the lone-success and clearing-run stages is safe.

Gap noted above; also untested but low-value: a block set by an `AcknowledgeAsync` fault rather than an apply fault (same `catch`, so behaviourally identical), and `UnknownPlayerLane` deferral.

### What's good

The comment at `:91-103` explains *why* the pass-scoped map is insufficient rather than just what the field holds, which is exactly the kind of thing that saves the next reader. #2467 is a genuinely useful follow-up — it lays out the design forks instead of just restating the gap, and it correctly flags that a real sequence guard would let this deferral be reconsidered rather than accumulating as permanent complexity.

---

ginderjeremiah:
Thanks — the restart observation is a good one to have on the record, and you're right that `StartAsync`'s reclaim-then-drain order makes the guarantee survive a restart independently of the process memory.

Pushed 8b152bf. Both should-fixes and all three nits are addressed, with one deliberate departure.

**Test coverage for the empty-list release (`:618-622`) — fixed as suggested.** You were right that `ProcessQueue_ParkedItemAcknowledgedElsewhere_...` doesn't reach that branch: the newer event is reserved and deferred before the reclaim check runs, so the head is non-empty and the release comes from `:646`. Added `ProcessQueue_ProcessingListObservedEmpty_ReleasesTheParkedLaneWithoutWaitingForAReclaim`, which does exactly what you described — park, acknowledge "elsewhere", one pass with no new events, then a newer same-player event applies on the next pass with no `timeProvider.Advance` at all. Verified it fails with only that `Clear()` removed and everything else intact. Also left a pointer on the older test so the two don't read as duplicates.

**Bounded reservation (`:447-460`) — amended the claims rather than bounding the sweep.** I took the second option you offered, and I want to be explicit that it wasn't the lazy one: I tried to talk myself into a per-pass deferral cap and it makes things worse in two ways.

The reclaim that replays the parked chain only runs when `ReserveNextAsync` returns null — an empty *main queue*. If a pass stops reserving early, the backlog stays on the main queue, the pass never reaches `TryReclaimStrandedProcessingAsync`, and recovery waits on however many passes it takes to sweep the backlog N at a time. Today's unbounded sweep is precisely what gets the pass to the empty-queue state where the reclaim fires and the chain replays promptly.

Worse, breaking out of the reserve loop is queue-wide, not per-player: one blocked hot player early in the queue would stop the pass and stall every player queued behind it until the grace period elapsed — directly undoing the per-player isolation `ProcessQueue_LaterPassAfterOutage_StillAppliesOtherPlayersEvents` pins.

And the thing a cap would buy is small, because it doesn't cap the processing list — the same items land there across successive passes either way (as they did pre-change, just paced by each item's retry budget). It only changes how fast, and bounds the reserved set at a stop, which the startup reclaim already restores in order.

So the comment and `docs/backend-persistence.md` now say reservation is bounded for every item that *actually applies*, and call the deferred item the deliberate exception with the reason it is load-bearing. Both places state it, so the next reader hits the reasoning rather than an inaccurate absolute.

**Nits.** Deferrals now log once per player per pass via a pass-scoped set, asserted in the ordering test (which now defers two events for the blocked player and expects exactly one line). The docs paragraph is split from **"That block outlives the pass that created it."** onward, so each paragraph keeps a single bold lead-in. `CompactCompletedItems`' doc now names `_parkedPlayerLanes` as the primary guard and scopes the lane retention to the reserved-before-the-block-is-recorded race you traced.

Full suite green: 3,526 passed, 0 failed, 0 build warnings.

---

Issue #1513, opened 2026-07-03
Session takeover briefly violates the per-player command serialization guarantee (per-socket lock, per-player promise)


ginderjeremiah:
## Problem

The command lock is per-**socket** (`SocketInterceptorMiddleware.cs:34,50-53`), while `docs/backend.md` promises per-**player** serialization. During a session takeover (`SocketManagerService.RegisterSocket`, `:40-85`) there is a window where both hold the same player:

1. The new connection loads its `Player` aggregate (`TryLoadPlayer`) **before** the old socket learns it was replaced — the `SocketReplaced` push serializes behind the old socket's in-flight commands.
2. If the old tab's last `DefeatEnemy` commits after the new connection's load, the new connection's first absolute-value write-behind save (e.g. exp) is derived from the pre-victory state — a lost update of exactly the class the lock exists to prevent.

Low frequency (a sub-second-to-seconds window once per takeover), but it slightly falsifies the documented guarantee and is currently untracked.

## Options

- Re-load/re-bind the new connection's player after the old socket's teardown completes (or after its queue drains), rather than before.
- Or explicitly document this as an accepted takeover race in `docs/backend.md` (the switch/takeover credit paths already tolerate near-miss timing).

Needs a design call on which side to take, hence no `claude` tag.

_Found during a codebase health review._

PR #2469, opened 2026-07-26
Read the socket's pinned player state from inside its presence claim (#2463)

### LINKED ISSUE #2463: Handshake loads Player/PlayerState before RegisterSocket's switch-credit deferral — reconnect during a switch-away credit resurrects pre-credit state
Found during a scheduled codebase-health review. Distinct from #1513 (two-socket takeover serialization) and from the fixed #2041/#2374 — those defend the presence-key *claim*; this is about a read that happens before the claim machinery ever runs.

## The gap

The socket handshake loads the player aggregate and pins it on the session **first** (`SocketInterceptorMiddleware.cs:36` → `TryLoadPlayer` → `SetPlayer`, `:92-111`), and only **then** calls `RegisterSocket` (`:68`), whose `ClaimPresenceKey` correctly defers behind an in-flight switch-credit claim (`SocketManagerService.cs:296-328`, the #2041/#2374 machinery). The deferral serializes the claim — but not the aggregate read that already happened.

## Scenario

1. User switches character A→B; `CreditDepartedCharacter(A)` (`CharacterSelectionService.cs:118-166`) claims A's presence key and runs the offline simulation, mutating A's aggregate (exp, `LastActivity`, resolved in-flight battle, `LastCreditedBattleSeed`).
2. Concurrently a new connection for A arrives (second tab/device, or the user switching back). The middleware loads A's session state and aggregate — **pre-credit values** — and pins them via `SetPlayer`.
3. `RegisterSocket` → `ClaimPresenceKey` observes the switch-credit claim and correctly defers until it releases.
4. Registration proceeds, but the connection keeps the stale pre-credit `Session.Player`/`PlayerState`. Nothing calls `MarkPlayerNeedsReload` (that machinery exists — `SessionService.cs:120`, engaged on flush failures in `SocketHandler.cs:252,267,340` — but not here). The first battle-completion `SavePlayer` writes absolute-valued events computed from pre-credit state and overwrites the Redis blob wholesale — the credit's rewards are silently lost. The stale aggregate also holds the old `LastCreditedBattleSeed`, so the `BattleAlreadyCredited` guard can't reject a re-credit of the battle the credit already resolved.

The window is narrow, but it's exactly the window the project already built three rounds of machinery (#2041, #2235, #2374) to defend.

## Fix direction

Load the player **after** `RegisterSocket` completes, or have `RegisterSocket` report that it observed/deferred behind a switch-credit claim and call `Session.MarkPlayerNeedsReload()` (plus re-read the session state) so the existing reload-before-next-command machinery converges onto the credited state.

ginderjeremiah:
Closes #2463.

## Summary

The socket handshake loaded and pinned the player aggregate **and** the session state before `RegisterSocket` ever claimed the player's presence key. A concurrent switch-away credit — a read-modify-write against that same aggregate, run over HTTP off the player's battle loop — could therefore commit in between and leave the connection holding pre-credit state. `ClaimPresenceKey` correctly deferred behind the credit's claim (#2041/#2374), but the deferral serialized the *registration*, not the read that had already happened.

The consequences are the ones the issue names: the first battle-completion `SavePlayer` writes absolute-valued events computed from the stale snapshot and overwrites the Redis blob wholesale, losing the credit's rewards; and the stale `LastCreditedBattleSeed` leaves the `BattleAlreadyCredited` guard unable to reject a re-credit of the very battle the credit just resolved.

`RegisterSocket` now takes an `onPresenceClaimed` callback and invokes it once it owns the presence key but **before** it subscribes the command listener or starts the read loop. That is the only safe window: before the claim, a credit's write can still land; after the loops start, replacing the pinned aggregate would swap it under a running command. `SocketInterceptorMiddleware` re-reads both halves there, via a new `SessionInitializer.ReloadSession` that forces the store re-read `EnsureSessionLoaded`'s idempotence would otherwise skip (and drops the currently-bound state first, so a cache miss can't silently leave the pre-claim snapshot in place).

## Why a callback rather than "load after `RegisterSocket` returns"

The issue's first suggestion — load the player after `RegisterSocket` completes — is the right idea but can't be done at the call site: `RegisterSocket` starts the read loop internally (`SocketConnectionRegistry.Register` → `SocketHandler.Listen`), and the handshake's 101 has already gone out by then, so a client command can be dispatched while the middleware is still re-reading. Re-pinning after the call would swap the aggregate under a running command. The callback puts the read in the one window where neither hazard exists.

## Why the re-read is unconditional

The issue's second suggestion — have `RegisterSocket` report that it *observed* a switch-credit claim and mark the session for reload — narrows the window but doesn't close it. A credit that starts *and finishes* between the handshake's read and `ClaimPresenceKey`'s read is never observed at all, and its writes are already committed (the claim is released only after `SimulateSwitchProgress`), so the connection would silently keep the stale pin. Only a read taken from inside the claim is actually guaranteed current, so the re-read runs unconditionally — two Redis round trips, once per connection.

Reusing `MarkPlayerNeedsReload` was the other option and was rejected for a second reason too: it means "the in-memory Player may hold mutations that never reached the write-behind queue" (#1632), and it covers only the aggregate, not the session state that carries the settled in-flight battle.

The pre-upgrade load stays exactly where it is. It is what rejects an unusable session with a real HTTP status and the project's `{ errorMessage }` envelope (the 409 `NoPlayerSelected` / 404 paths and their tests are unchanged); moving it past the upgrade would have downgraded those to a socket close.

## Test plan

- [x] New `Connect_PlayerCreditedWhileDeferringBehindSwitchClaim_PinsThePostCreditAggregate` — the issue's exact interleaving for the aggregate half: a claim is held, the handshake completes and pins pre-credit state, the credit's write lands, the claim releases, and the connection must report the post-credit allocations.
- [x] New `Connect_SessionBattleSettledWhileDeferringBehindSwitchClaim_PinsThePostCreditSessionState` — the session-state half, probed through `BattleLost`'s short-circuit on the connection's own in-memory `PlayerState`.
- [x] Verified both fail against the unfixed ordering (neutralised the middleware's re-read, reran: exactly those 2 failed) and pass with it.
- [x] New `RegisterSocket_SwitchCreditClaimActive_PinsTheSessionOnlyOnceTheClaimIsThisSockets` — the ordering contract itself: the re-read does not run while the credit holds the key, and the key it observes is this socket's own id.
- [x] New `RegisterSocket_PinningTheSessionFails_RollsTheRegistrationBackRatherThanGoingLive` — a fault in the re-read releases the presence key rather than going live with state read outside the claim.
- [x] New `SessionInitializerTests` (5 cases) — `EnsureSessionLoaded` stays idempotent; `ReloadSession` replaces bound state, rebinds to the token's player when the account-keyed cache has moved to another character, discards the stale state on a cache miss, and no-ops on a pre-selection token.
- [x] `dotnet build` clean (0 warnings); full `dotnet test`: 3542 passed, 0 failed.

No frontend changes — this is the socket handshake only, no battle logic involved.

## Docs

`docs/backend-sockets.md` → "Single active connection" gains a paragraph on the new ordering guarantee, alongside the #2041/#2374 claim machinery it completes.

## Incidental

`FakeSessionStore` moved from a private nested class in `SessionServiceTests` to `Game.TestInfrastructure/Helpers` so the new `SessionInitializerTests` shares it rather than duplicating it (it also now counts `GetSession` calls, which is how the idempotence assertion is made).

---
_Generated by [Claude Code](https://claude.ai/code/session_019BGm55JeV4MQTcpxDrUdan)_


docs/backend-sockets.md
```diff
@@ -22,6 +22,8 @@ The same key also backs the in-game character-switch credit (`docs/backend-auth.
...
 
+**The connection's pinned player state is read from inside its own presence claim.** The handshake loads the player aggregate and session state before the upgrade so an unusable session is rejected with a real HTTP status rather than erroring on the first command — but that read happens before any claim exists, so a concurrent switch-away credit can commit its read-modify-write in between and leave the connection holding pre-credit state that its first battle-completion save would write back wholesale. `RegisterSocket` therefore re-reads both once it owns the presence key and before it starts any loop, which is the only window where no credit can be in flight (one was waited out) and none can start (the key is taken), and nothing can yet be running a command against what is pinned. The re-read is unconditional rather than gated on having observed a claim: a credit that begins and ends between the handshake's read and the claim is never observed at all (#2463).
```
ginderjeremiah: **Should-fix (docs accuracy).** "the only window where no credit can be in flight ... and nothing can yet be running a command against what is pinned" reads as an absolute guarantee against *all* concurrent writers, and it isn't one. Two writers survive that window:

- **A same-player socket being taken over.** `ClaimPresenceKey` returns `oldSocketId` when it CAS'd over a *real* socket id, and that connection is only told it was replaced after registration completes (`SocketManagerService.cs:158-161`). So during `onPresenceClaimed` the old socket can still be mid-command, mutating the same aggregate and session state. That's #1513's serialization gap, which the issue explicitly scopes out — fine to leave, but the doc shouldn't imply it's closed.
- **Another character on the same account.** `Session_{userId}` is account-keyed, and the re-read runs *before* `ClaimAccountPresence`, so a different character's live socket can still write that key here (#1817's last-writer-wins, also unchanged).

The trailing clause ("nothing can yet be running a command against what is pinned") is true of *this* connection's own loops, which is what matters for the swap-under-a-running-command hazard — it just doesn't generalize. Suggest scoping the sentence explicitly to switch-away credits ("no *switch-away credit* can be in flight ... and this connection's own loops cannot yet be running a command against it"), so the guarantee the paragraph claims is the one the code actually provides.

Since this file is the source of truth a future reader will reason from, precision here is worth more than in a code comment.

---

Game.Api/Services/SocketManagerService.cs
```diff
@@ -107,6 +119,12 @@ public async Task<SocketContext> RegisterSocket(WebSocket socket, SessionService
             int? oldAccountPlayerId;
             try
             {
+                // The presence key is ours from here, so no switch-away credit can be mid-flight (ClaimPresenceKey
+                // waited any out) and none can start (TryClaimForSwitchCredit only claims an unset key). Read the
+                // connection's pinned state now, before anything below can start a loop that runs a command
+                // against it — see onPresenceClaimed.
+                await onPresenceClaimed();
```
ginderjeremiah, line 132: **Nit (unstated contract / footgun).** `playerId`, `presenceKey` and `socketContext` are all captured from `sessionService.SelectedPlayerId` *before* this callback runs, and the callback's whole job is to rebind the session. Nothing states that it must land on the same player — if a future `ReloadSession` ever bound to something other than `TokenSelectedPlayerId`, the presence key would be claimed for player A while the connection is pinned to player B, and every downstream consumer (`SocketContext.PlayerId`, the presence refresh, the account slot) would silently disagree with the pinned aggregate.

Today it holds: `EnsureSessionLoaded` guarantees `SelectedPlayerId == TokenSelectedPlayerId` before the upgrade, and `ReloadSession` re-establishes that same binding. But that invariant lives two files away, and this is exactly the kind of deliberate-looking ordering a later edit could break without any test failing. Worth a line on the `<param>` doc: the callback may replace the pinned state but must not change the session's selected player, since the registration's identity is already captured.

---

Game.Api/Middleware/SocketInterceptorMiddleware.cs
```diff
@@ -98,16 +109,44 @@ private static async Task<PlayerLoadResult> TryLoadPlayer(SessionService session
...
+        /// switch-away claim: a credit that started and finished between the handshake's load and the claim
+        /// is never observed at all, so only a read taken from inside the claim is actually guaranteed
+        /// current. A player that no longer loads keeps the handshake's already-pinned aggregate — the row
+        /// vanishing inside this window is pathological, and a stale pin still beats leaving the connection
+        /// with none (mirroring <c>SocketHandler</c>'s own reload-before-command handling).
+        /// </summary>
+        private static async Task ReloadSession(SessionService sessionService, SessionInitializer sessionInitializer, IServiceScopeFactory scopeFactory, CancellationToken cancellationToken)
+        {
+            await sessionInitializer.ReloadSession(cancellationToken);
+            await TryPinPlayer(sessionService, scopeFactory, cancellationToken);
```
ginderjeremiah: **Nit (observability).** Discarding the `bool` is the right call and the reason is well documented above — but it makes the one case that matters completely silent. If the re-read finds no player, this connection knowingly goes live on a pin taken *outside* the presence claim, which is the precise condition #2463 exists to prevent, and there'd be no trace of it in the logs when the resulting lost-rewards report comes in.

A `LogWarning` on the false branch (player id + "keeping the pre-claim pin") costs nothing on the happy path and turns a pathological-but-possible state into something diagnosable. `if`-with-braces per the house rules, so it's three lines.

---

Game.Api/Services/SessionInitializer.cs
```diff
@@ -40,7 +58,14 @@ public async Task EnsureSessionLoaded(CancellationToken cancellationToken = defa
...
+            {
+                // Drop the currently-bound state up front: LoadPlayerState keeps what is already bound when
+                // the store has nothing, so without this a cache miss would silently leave the very snapshot
+                // this reload exists to discard in place.
+                _sessionService.RehydrateSession(selectedPlayerId);
+            }
+            else if (_sessionService.HasPlayerSession && _sessionService.SelectedPlayerId == selectedPlayerId)
             {
                 return;
             }
```
ginderjeremiah: **Nit.** Two small consequences of resetting up front that are worth a look:

1. **The forced-reload path never logs a rehydration.** Because `RehydrateSession` already bound the token's player, the check at `:74` succeeds on a store miss and returns early — so the `LogInformation` at `:81` only fires when the store held a *different* character. An eviction between the handshake's read and the re-read therefore discards the connection's in-flight battle completely silently, where the same miss on the `EnsureSessionLoaded` path logs it. Given `PlayerState` is the only home for an active battle (`docs/backend-persistence.md:13`), that's the one drop worth a line in the log.

2. **`RehydrateSession` can run twice per reload** (here, then again at `:83` when the store held another character). Harmless — it's idempotent — but it reads as accidental. Both fall out naturally if the reset is expressed as "reset, then treat the post-load state uniformly"; not worth restructuring for on its own, just noting since (1) is a real gap.

The comment itself is good: it explains the non-obvious `LoadPlayerState`-keeps-what's-bound behavior that makes the reset necessary, which is precisely the sort of thing a later "simplification" would strip out.

---

Game.Api.Tests/Unit/SessionServiceTests.cs
```diff
@@ -284,33 +284,5 @@ public void ClearSession_NoUserResolvedAndUnauthenticated_EvictsNothing()
             Assert.Empty(store.Cleared);
         }
```
ginderjeremiah: **Nit.** Stray blank line left behind where the nested `FakeSessionStore` was lifted out — the class now ends with a dangling empty line before its closing brace.

---

ginderjeremiah (commented):
## Review: read the socket's pinned player state from inside its presence claim (#2463)

**Verdict:** Approve, with one should-fix (doc wording) and four nits — the fix is correct, minimal, and lands the read in the right window. Nothing here blocks merge. (Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR author.)

### The design is right, and both rejected alternatives really are worse

I checked the two directions the issue suggested against the code rather than taking the description's word for it:

- **"Load after `RegisterSocket` returns"** genuinely can't be done at the call site. `_socketRegistry.Register(socketHandler)` (`SocketManagerService.cs:142`) starts `Listen` inside `RegisterSocket`, and the 101 went out at `AcceptWebSocketAsync` before that — so a client command can already be dispatching when the middleware would be re-pinning. Re-pinning after the call would swap the aggregate under a running command, which is the hazard the `<param>` doc names.
- **"Mark the session for reload when a claim was observed"** leaves the interesting case uncovered. `CreditDepartedCharacter` releases its claim in a `finally` *after* `SimulateSwitchProgress` (`CharacterSelectionService.cs:156-163`), so a credit that starts and completes between the handshake's read and `ClaimPresenceKey`'s read is both invisible to the registration and already committed. Only a read taken from inside the claim is current, so unconditional is correct. The second reason for not reusing `MarkPlayerNeedsReload` is the stronger one: it means "the in-memory `Player` may hold unflushed mutations" (#1632), which is a different fact, and it doesn't cover the session state at all.

The callback lands in the one correct spot — after `ClaimPresenceKey`, before `ClaimAccountPresence` / `RegisterSocketCommandListener` / `Register` — and inside the `try`, so a fault rolls the presence claim back rather than going live on a pre-claim read. That's the same discipline `ClaimAccountPresence` already follows, and it's tested directly.

Inverting control via `Func<Task>` rather than injecting session concerns into `SocketManagerService` is the right trade: the manager stays ignorant of pinning, and the one production call site reads clearly.

### Two things I traced that behave correctly — noting them so they're on the record

- **The residual session-state window is covered by the aggregate half.** The credit clears the battle off `PlayerState` in memory, but the *account-keyed* `Session_{userId}` blob is only rewritten by the subsequent `CreateSession(userId, target)` — which runs after the claim releases. So a re-read that lands in between can still observe the pre-credit session state showing the settled battle as active. The consequence is contained precisely because this PR fixes the aggregate too: the re-read now carries the credit's `LastCreditedBattleSeed`, so `BattleAlreadyCredited` rejects the re-credit instead of paying it out twice. Worth knowing that the two halves are load-bearing for each other, not independent.
- **Player identity can't drift across the re-read.** `EnsureSessionLoaded` guarantees `SelectedPlayerId == TokenSelectedPlayerId` before the upgrade and `ReloadSession` re-establishes the same binding (including when the account-keyed cache has moved to another character), so the `playerId` captured before the callback still names the pinned player. That invariant is what keeps the presence key and the pin consistent — see the inline note asking for it to be stated.

### Should-fix

- **`docs/backend-sockets.md:25`** — the new paragraph claims more than the code delivers: a same-player socket being taken over (#1513) and another character's socket on the same account (#1817) can both still write inside that window. Scope the sentence to switch-away credits. Details inline.

### Nits

- `SocketManagerService.cs:122-126` — state on the `<param>` doc that the callback must not change the session's selected player; the registration's identity is captured before it runs.
- `SocketInterceptorMiddleware.cs:130` — log a warning when the re-read finds no player; that's the connection knowingly going live on a pre-claim pin, and today it's silent.
- `SessionInitializer.cs:61-71` — the forced-reload path never logs a rehydration, so an eviction inside the window discards an in-flight battle with no trace (and `RehydrateSession` can run twice).
- `SessionServiceTests.cs:286` — stray blank line left where the nested class was lifted out.

### Test coverage

Strong, and the diligence is the right kind: verifying the two integration tests fail against the unfixed ordering (by neutralising the middleware's re-read) is what makes them meaningful rather than merely green.

- The interleaving both integration tests reproduce is **deterministic**, not timing-hopeful — `TryLoadPlayer` runs before `AcceptWebSocketAsync`, so the client's `ConnectAsync` returning proves the pre-credit state is already pinned, and the out-of-band write is awaited before the claim releases. No sleeps, no polling.
- Using a stat reallocation as the stand-in for the credit's read-modify-write is a good call — it's the same aggregate write and, unlike exp, a read-only command echoes it straight back. The constants line up with the seeder (`TestDataSeeder` seeds 50/50, so 40/60 after the ±10).
- Probing the session half through `BattleLost`'s short-circuit is the cleanest available observation of what the connection is actually pinned to.
- `RegisterSocket_SwitchCreditClaimActive_PinsTheSessionOnlyOnceTheClaimIsThisSockets` pins the *ordering contract*, not just the outcome, and asserting the observed key equals `context.SocketId` (rather than merely "not the sentinel") is the assertion that actually rules out a mid-flight credit. The rollback test covers the fault path.
- `SessionInitializerTests` covers the four `ReloadSession` branches plus `EnsureSessionLoaded`'s idempotence, and the `GetSessionCalls` counter is the right way to assert the latter.

One gap: the documented decision that a player which no longer loads keeps the handshake's pin (`SocketInterceptorMiddleware.cs:123-125`) is untested. It's awkward to reach — the row would have to vanish inside the claim window — so I'd rather see the log line from the inline nit than a contrived test; noting it so the choice is explicit rather than overlooked.

### Follow-up

Filed #2470: now that `FakeSessionStore` is a shared helper, the four private `NoOpSessionStore` copies across the API test suite are redundant duplication of the same interface (two of them with `playerId`-named parameters where the interface says `userId`). Pre-existing, out of scope here, and only worth doing *because* this PR created the canonical double.

### What's good

Promoting `FakeSessionStore` into `Game.TestInfrastructure/Helpers` instead of copying it was the right instinct, and adding `GetSessionCalls` to it makes the idempotence claim testable rather than assumed. The `<param>` doc on `onPresenceClaimed` explains *why* the window is unique in both directions (before → stale, after → swap under a running command), which is what stops a future reader from "simplifying" the call to either side of it — the same quality as the reasoning in the PR description, which correctly refused to just implement the issue's first suggestion.

---

ginderjeremiah:
All five addressed in `01a5327`. Full suite still green (3542 passed, 0 failed), build clean.

**Docs (should-fix)** — you're right, and the correction is load-bearing rather than cosmetic: `ClaimPresenceKey` returns `oldSocketId` for a *real* socket that isn't signalled until registration completes, and `ClaimAccountPresence` runs after the callback, so both writers you named are live during that window. The paragraph now scopes the guarantee to switch-away credits, names #1513 and #1817 as the two that remain, and limits the command clause to this connection's own loops.

**`SessionInitializer` (nit 4)** — I did restructure, despite your "not worth it on its own", because your point (1) doesn't have a smaller fix. The reset-up-front made the post-load check unable to distinguish "the store had nothing" from "the store served exactly what was already bound", which is precisely why the eviction case skipped the log. `LoadPlayerState` now reports whether it served anything, both paths share one rehydrate-and-log branch, and the double `RehydrateSession` falls out with it. Net −4 lines.

**Other three** — warning added on the failed re-read (which also covers the untested no-player-loads decision you flagged, as you preferred); the `<param>` doc now states that the callback must not change the selected player and why (identity captured before it runs); stray blank line gone.

Also noted your two on-the-record traces, particularly that the two halves are load-bearing for each other — the session-state re-read can still observe a pre-credit `Session_{userId}` blob (only rewritten by the later `CreateSession`), and it's the aggregate half carrying the credit's `LastCreditedBattleSeed` that makes `BattleAlreadyCredited` reject the re-credit. That's worth remembering before anyone splits them.

#2470 looks right as a separate cleanup — agreed it's only worth doing now that this PR created the canonical double.

---

ginderjeremiah:
## Re-review of `01a5327` (delta only — `78f21b9` was reviewed in full)

**Verdict: approve.** All five findings addressed, nothing new introduced, no remaining nits. CI green on this head (`test-backend`, `verify-backend`, all three `test-e2e` browsers, `all-checks-passed`).

### The one change worth checking closely: the `LoadPlayerState` → `Task<bool>` restructure

This was the only finding where you went further than the nit asked for, so I traced it rather than taking the summary's word for it — and the reasoning holds. The reset-up-front genuinely couldn't distinguish "the store had nothing" from "the store served exactly what was bound", so the eviction case had no smaller fix.

Checked the restructure is behavior-preserving on the path it *didn't* target:

- The post-load guard weakened from `HasPlayerSession && SelectedPlayerId == selectedPlayerId` to `loaded && SelectedPlayerId == selectedPlayerId`. Since `SelectedPlayerId => PlayerState.PlayerId` and `HasPlayerSession => PlayerState.PlayerId > 0`, and the token id is ≥ 1, `SelectedPlayerId == selectedPlayerId` already implies `HasPlayerSession` — the dropped conjunct was redundant, not load-bearing. Every non-forced branch (first-load hit, first-load miss, bound-to-another-character then miss) reaches the same outcome as before.
- The forced path is the intended change: a miss now rehydrates *and logs* instead of silently landing on the snapshot it was called to discard, and `RehydrateSession` runs once instead of twice. `ReloadSession_StoreHasNoSession_DiscardsTheStaleBoundState` pins it, and the two `SessionServiceTests` cases now assert the returned flag in both directions, so the new contract is covered at its source rather than only through its consumer.

Net −4 lines for strictly more signal is the right trade.

### The other four

- **Docs** — the correction is accurate against the code, which is what makes it worth having: `ClaimAccountPresence` runs *after* `onPresenceClaimed`, and `NotifyReplacedSocket` for the other character's socket runs after the whole `try`, so a same-account socket really can still write `Session_{userId}` inside the window. Scoping the guarantee to switch-away credits and naming #1513/#1817 as the two that remain is the honest statement. Citing a closed issue for the machinery matches how the surrounding paragraphs cite #2041/#2374.
- **Warning on a failed re-read** — lands in the right place and covers the untested "player no longer loads" decision, which was my stated preference over a contrived test. Message names the actual consequence (pre-claim pin kept, may be stale) rather than just the failure.
- **`<param>` doc** — states the constraint *and* why it exists (presence key, account slot, and `SocketContext.PlayerId` all captured before the callback), which is what makes it enforceable by a future reader.
- **Blank line** — gone.

### Registration wiring

`SocketInterceptorMiddleware` taking `ILogger<SocketInterceptorMiddleware>` is fine: it's added via `UseMiddleware<SocketInterceptorMiddleware>()` (`Startup.cs:296` → `UseSocketInterceptor`), so the logger resolves from DI at pipeline build like the other middleware in that folder.

### On the account-slot residual

Worth stating explicitly rather than leaving implied by the doc edit: moving the callback after `ClaimAccountPresence` would *not* close that window, because the other character's socket isn't gated on the account key — it's only notified afterwards, best-effort, and keeps writing until it closes. That residual is inherent to the notify-based handoff (the same shape as #1513, which tracks the same-player analogue), not something this PR could fix or made worse. Documenting it is the right resolution; no new issue filed.

### Follow-ups

- #2470 (consolidate the duplicated `ISessionStore` doubles) — already filed, still the only follow-up this PR generates.

Merging.

---

PR #2471, opened 2026-07-26
Make the TypeScript codegen's output locale-independent (#2464)

### LINKED ISSUE #2464: TypeScript codegen uses culture-sensitive ToLower/OrderBy — generated output can differ by dev-machine locale, breaking the CI byte-equality drift guard

All casing and ordering in the codegen's (CI byte-compared) output uses culture-sensitive operations:

- `Game.Api/CodeGen/Extensions.cs:14` — `Decapitalize` uses `.ToLower()`: on a tr-TR machine `"Id"` → `"ıd"` (dotless i), emitting an identifier that breaks the frontend build and mismatches `JsonNamingPolicy.CamelCase` (which is invariant). `:23` — `SnakeCase` `.ToLower()` (file names).
- Default culture-sensitive string comparers in every ordering: `ApiCodeGenerator.cs:91-92`, `CodeGenTypeFormatter.cs:10`, `Writers/SocketMapWriter.cs:21`, `Writers/ApiMapWriter.cs:21`, `Writers/ApiInterfaceWriter.cs:72,167`.

Underscore-containing constant names and mixed-case type names sort differently between ordinal and culture collation (and between ICU versions), so a dev machine with a different locale/ICU than CI can regenerate different bytes → spurious drift-guard failures or a bad commit. The project's own standard elsewhere is explicit ordinal (`ReferenceDataVersioning`/`GetReferenceDataVersions` uses `StringComparer.Ordinal` deliberately).

Fix: `ToLowerInvariant`/`char.ToLowerInvariant` and `StringComparer.Ordinal` throughout the codegen. Single-dev en-locale environments mask this today, which is why it hasn't bitten yet.

ginderjeremiah:
Closes #2464.

## Summary

Every casing and ordering decision feeding the codegen's **byte-compared** output was culture-sensitive, so the same sources could regenerate different bytes on a machine whose locale or ICU version differs from CI's:

- `CodeGenExtensions.Decapitalize`/`SnakeCase` used `ToLower()`. On a tr-TR machine `"Id"` becomes the dotless `"ıd"` — an identifier that breaks the frontend build *and* disagrees with the invariant `JsonNamingPolicy.CamelCase` the API actually serializes with. `SnakeCase` has the same problem for generated file names.
- Every ordering used the default (culture-aware) comparer: `ApiCodeGenerator.GetClientMirroredConstantFields`, `CodeGenTypeFormatter.GetImportText`, `SocketMapWriter`, `ApiMapWriter`, and `ApiInterfaceWriter` (both the per-file interface order and the `index.ts` export order).

All of them now use `char.ToLowerInvariant`/`ToLowerInvariant` and `StringComparer.Ordinal` — matching the deliberate ordinal comparison `ReferenceDataVersioning` already uses. I verified the issue's claims against the code first: all eight sites it names are real, and no other culture-sensitive operation remains in `Game.Api/CodeGen` (`Replace`/`IndexOf(char)`/`StartsWith(…, Ordinal)` are already ordinal, and `ConstantsWriter`/`StaticModifierWriter` already format values with `InvariantCulture`).

## The committed client is regenerated, and that is the proof

Switching to ordinal reorders three real files, so they are regenerated here:

| generated file | reorders |
| --- | --- |
| `api-socket-type-map.ts` | `GetChallengeTypes` before `GetChallenges` |
| `api-type-map.ts` | `AddEditItemModSlots` before `AddEditItemMods` (both maps) |
| `interfaces/player.ts` | `ILogPreference` before `ILoginCredentials` |

That diff is direct evidence the drift guard was locale-sensitive: ICU compares case-insensitively before case, so `'T'` vs `'e'` and `'P'` vs `'i'` land the other way round from ordinal. Anyone generating on a differently-collating machine would have produced exactly this churn as a spurious drift failure.

## Test plan

- [x] `CodeGenExtensionsTests` — `Decapitalize`/`SnakeCase` pinned under a `tr-TR` `CultureScope` (`"Id"` → `"id"`, `"ItemId"` → `"item-id"`, not `"ıd"`/`"ıtem-ıd"`).
- [x] `ApiMapWriterTests`, `SocketMapWriterTests`, `CodeGenTypeFormatterTests`, `ApiInterfaceWriterTests` — one ordinal-ordering test each, using inputs that ICU orders differently (`Player/Get` vs `PlayerData` vs `alpha`; the real `GetChallengeTypes`/`GetChallenges` and `ILogPreference`/`ILoginCredentials` pairs). These fail on an en-US machine against the old code, not only on an exotic locale.
- [x] `ApiCodeGeneratorTests` — the existing determinism test now pins the *ordinal* contract, plus a new check that the constant-field order is unchanged under `tr-TR`.
- [x] Verified the new tests fail without the fix: reverting only the six production files leaves exactly 7 failures (the 7 new cases), all passing again with the fix.
- [x] `dotnet build` clean (0 warnings); full `dotnet test`: 3545 passed, 0 failed.
- [x] Both CI verifications run locally: `dotnet format --verify-no-changes` clean, and re-running `dotnet run --project Game.Api -- codegen` after the commit leaves a clean `git diff`.
- [x] Frontend `npm run lint` (ESLint + `svelte-check`): 0 errors, 0 warnings — the regenerated files are type-map reordering only, no semantic change.

No battle logic is involved, so there is no frontend parity suite to mirror.

## Notes

- **New shared helper**: `Game.TestInfrastructure/Helpers/CultureScope.cs`, so any future locale-sensitivity test has one place to swap and restore the culture. `CurrentCulture` is thread-static, so a synchronous test body cannot leak it into a parallel test — the XML doc says so, and warns against awaiting inside the scope.
- **Docs**: `docs/infrastructure.md` → the standalone-codegen section gains a short paragraph stating the invariant (output depends only on the sources, never on the generating machine) and why, so the next writer added to the codegen doesn't reintroduce a default comparer.
- I checked the other byte-compared artefact, the content export (`ContentExportSerializer`, guarded by `ContentExportDriftGuardTests`), for the same class of bug — its sorts are all over `int` keys, so it is unaffected. No follow-up issue needed.

---
_Generated by [Claude Code](https://claude.ai/code/session_014JWKkYCvGKFseyB1Nq13Ln)_


docs/infrastructure.md
```diff
@@ -17,6 +17,8 @@ dotnet run --project Game.Api -- codegen [outputDirectory]
 
 The output is byte-identical to the dev-time generation, so CI can run the command and assert a clean `git diff` to catch api model/client drift.
 
+**The generated output must depend only on the sources, never on the generating machine.** Because the drift guard compares bytes, every casing and ordering decision in the codegen is culture-independent — `ToLowerInvariant`/`char.ToLowerInvariant` for the emitted identifiers and file names (a tr-TR machine's `ToLower` turns `Id` into the dotless `ıd`, which also mismatches the invariant `JsonNamingPolicy.CamelCase` the API serializes with), and `StringComparer.Ordinal` for every sort. Culture-aware collation orders mixed-case and separator-containing names differently from ordinal — and differently again across ICU versions — so a default-comparer sort would regenerate different bytes on another developer's locale.
```
ginderjeremiah: **Nit** — the invariant is worth documenting; the API-level enumeration around it isn't. `CLAUDE.md`'s documentation guidance asks docs to avoid "extremely specific implementation details" and warns they go stale — and `ToLowerInvariant`/`char.ToLowerInvariant`/`StringComparer.Ordinal`, plus the dotless-`ı` example, are all already stated in the XML docs on `Decapitalize`, `SnakeCase`, and `GetClientMirroredConstantFields`, where they sit next to the code that has to keep them true.

What only the doc can carry is the rule itself and why it exists. Something closer to:

> **The generated output must depend only on the sources, never on the generating machine.** The drift guard compares bytes, so every casing and ordering decision in the codegen is culture-independent — culture-aware casing and collation differ by locale *and* by ICU version, so a default comparer would regenerate different bytes on another developer's machine and fail the guard spuriously.

That's the part a future writer added to the codegen needs, and it stays correct even if the specific helpers change.

---

Game.Api.Tests/CodeGen/ApiCodeGeneratorTests.cs
```diff
@@ -92,16 +93,29 @@ public void GetClientMirroredConstantFields_IsOrderedDeterministically()
...
+        [Fact]
+        public void GetClientMirroredConstantFields_OrderIsUnaffectedByHostCulture()
+        {
+            var invariantOrder = ApiCodeGenerator.GetClientMirroredConstantFields().Select(field => field.Name).ToList();
+
+            using var culture = new CultureScope("tr-TR");
+            var turkishOrder = ApiCodeGenerator.GetClientMirroredConstantFields().Select(field => field.Name).ToList();
+
+            Assert.Equal(invariantOrder, turkishOrder);
+        }
```
ginderjeremiah, line 123: **Nit** — this one doesn't discriminate today, unlike the other six new cases, and it'd be worth saying so in the comment.

Unlike the writer tests, this can't inject inputs — `GetClientMirroredConstantFields` reflects the real `Game.Core` assembly — so it only fails if some *actual* `[ClientMirrored]` name collates differently between the ambient culture and tr-TR. None does: the declaring types are `ContentFieldLengths`/`GameConstants`/`ServerGameConstants`, and every `i`/`I` in the field names sits in a position where Turkish collation still yields the ordinal order (`ItemIconPath…` < `ItemMod…` either way). The PR's own verification is the confirmation — reverting the production files left "exactly 7 failures," and this is the eighth new case.

That's consistent with `game-constants.ts` needing no regeneration here, which is itself the proof that culture and ordinal agree on today's names.

It's still reasonable to keep as insurance for a future collation-sensitive constant, but a reader will otherwise assume it's a failing-before/passing-after test like its neighbours. A line saying it guards future names rather than pinning current behaviour — and pointing at `GetClientMirroredConstantFields_IsOrderedDeterministically` above as the assertion that actually states the ordinal contract — would set that expectation.

---

ginderjeremiah (commented):
## Review: make the TypeScript codegen's output locale-independent (#2464)

**Verdict:** Approve, with three nits and no blockers — the fix is complete, the regenerated client is real evidence rather than churn, and the tests are built to fail on an ordinary en-US machine instead of only under an exotic locale. (Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR author.)

### The fix is complete — I re-audited rather than trusting the issue's list

The issue names eight sites; all eight are real and all eight are fixed. I also swept `Game.Api/CodeGen` for everything else that could vary by host, and it comes back clean:

- **Casing** — `Decapitalize` and `SnakeCase` were the only culture-sensitive cases. `ConstantsWriter.ToScreamingSnakeCase` was already `ToUpperInvariant`.
- **Ordering** — every `OrderBy`/`ThenBy` over a string now passes `StringComparer.Ordinal`. The remaining comparisons are already culture-independent by construction: `DistinctBy`/`ExceptBy`/`GroupBy` use `EqualityComparer<string>.Default` (ordinal), `ControllerMetadataExtractor.cs:53` already passes `StringComparison.Ordinal`, and `IndexOf('`')` is a char overload.
- **Value formatting** — `ApiInterfaceWriter.cs:121`, `StaticModifierWriter.cs:50`, and `ConstantsWriter.RenderValue` all already format through `InvariantCulture`; `DamageTypeMapWriter` emits only enum names.
- **The regexes** don't set `IgnoreCase`, so their character classes aren't culture-sensitive — no `RegexOptions.CultureInvariant` needed.

Worth noting `SnakeCase`'s reach: it's used by `CodeGenTypeDescriptor.LastNamespacePart`, so it names the emitted **files**, and on a tr-TR machine an `…Item` namespace would have produced `ıtem.ts` — a file the `index.ts` export then can't resolve. The `SnakeCase_TurkishCulture_LowercasesInvariantly` case covers exactly that path.

### The regenerated client is the right kind of proof

All three reorderings check out as genuine ordinal-vs-ICU disagreements, not incidental churn: `'T'`(84) < `'e'`(101), `'S'`(83) < `'s'`(115), `'P'`(80) < `'i'`(105) — each one a pair ICU orders the other way because it compares case-insensitively before case. Committing the regeneration in the same PR is what turns "this could differ by locale" into a demonstration that it did. And the change is inert on the frontend: these are type-map key orders and two hoisted interface declarations, so nothing observable moves.

### I verified the claim about the other byte-compared artefact

The description asserts `ContentExportSerializer` is unaffected because its sorts are over `int` keys, and that "no follow-up issue needed." That's correct, including the two sorts that *look* like string sorts — `item.Tags` and `mod.Tags` are `IEnumerable<int>` (`Game.Abstractions/Contracts/Item.cs:16`, `ItemMod.cs:14`), as are `SkillPool`, `StarterSkillIds`, `PrerequisiteIds`, and `InputSkillIds`. The rest sort on ints or enums, both of which compare numerically. Good call checking the sibling drift guard unprompted; good that the conclusion holds.

### Nits

- `docs/infrastructure.md:20` — the invariant is worth stating, the API-level enumeration around it isn't (details inline).
- `Game.Api.Tests/CodeGen/ApiCodeGeneratorTests.cs:110-119` — `GetClientMirroredConstantFields_OrderIsUnaffectedByHostCulture` doesn't currently discriminate; worth a comment saying it guards future names (details inline).
- **`ToScreamingSnakeCase` is now the one casing operation in the codegen with no locale guard.** It's already correct (`ToUpperInvariant`) so there's nothing to fix, but it's also the last place where a future "simplification" to `ToUpper()` reintroduces this exact bug — `"ItemId"` → `"İTEM_İD"` on tr-TR — with no test failing. Now that `CultureScope` exists and the two sibling helpers are pinned under it, a symmetric one-line `[Fact]` in `ConstantsWriterTests` would close the set. Optional, but it's cheap and the file is right next door.

### Test coverage

Solid, and deliberately built so it doesn't depend on the CI machine's locale to be meaningful.

- The four writer/formatter tests use inputs whose ordinal and culture-aware orders genuinely disagree (`Player/Get` vs `PlayerData` vs `alpha`, and the real `GetChallengeTypes`/`GetChallenges` and `ILogPreference`/`ILoginCredentials` pairs), so they fail against the old code on an en-US box — not only on tr-TR. That's the harder and more useful choice than wrapping everything in a `CultureScope`, because a locale-scoped test only catches the locale it names.
- I checked the assertions actually discriminate rather than merely pass: each compares two names whose ordinal order inverts under ICU's case-insensitive-first collation, and each guards the `>= 0` case so a not-found index can't be mistaken for a correct ordering.
- `ApiInterfaceWriterTests`' `Assert.Single(files)` holds because both fixtures share the `Game.Api.Tests.CodeGen` namespace (one group → one `code-gen.ts`) and `enums.ts` is written at the target root, not under `interfaces/`.
- The new `TestFixtures` models are safe to add — no test in the suite reflects over all fixture `IModel`s, so nothing else shifts underneath.
- Verifying the tests fail without the fix, and counting the failures, is the diligence that makes the suite meaningful — and it's also what let me pin down that the eighth case doesn't yet discriminate. Worth keeping that habit.

No battle logic is touched, so there's no frontend parity suite to mirror. Correct call not to invent one.

### Follow-up

Filed #2472: `Game.Core.Extensions.Capitalize` (`Game.Core/Extensions.cs:96`) is the exact mirror of `Decapitalize` and still uses culture-sensitive `ToUpper()`. It's latent today — every call site capitalizes an already-PascalCase enum member name, so it's a no-op on any locale — but four of the five call sites are EF `HasData` seed values (`GameContext.cs:331,384,793,815`), so a wrong result would be persisted rather than merely displayed. Deliberately out of scope here: different subsystem, no drift guard over it, and this PR is correctly scoped to the codegen.

### What's good

`CultureScope` is the standout. Promoting it into `Game.TestInfrastructure/Helpers` rather than nesting it in one test class is right, and the XML doc goes further than most such helpers do — it states *why* a synchronous body can't leak the culture into a parallel test (`CurrentCulture` is thread-static in .NET, so the set and the restore land on the same thread) and warns against awaiting inside the scope. That's the fact a future user needs in order to not misuse it, and it's exactly the kind of non-obvious constraint that's worth two lines.

The description also deserves credit for a specific restraint: it verified the issue's eight claimed sites against the code before implementing, and separately went looking for the same class of bug in the *other* byte-compared artefact. Both are the "don't assume the issue author analyzed the code" discipline `CLAUDE.md` asks for, and in this case the issue happened to be right — which is only knowable because someone checked.

---

PR #2477, opened 2026-07-26
Share the Toughness mitigation curve between the engine and CombatRating (#2450)

### LINKED ISSUE #2450: CombatRating.Survivability re-derives the Toughness mitigation curve by hand — the last engine-vs-rating duplication after #2439
Found while implementing #2439 (which closed the crit and execute halves of `CombatRating`'s anti-drift rule, after #2429 closed parry/dodge).

## Problem

With crit, execute, and avoidance all now reading shared `Battler` members, **the Toughness mitigation curve is the one piece of engine math the rule still doesn't cover.** The expression is spelled out twice, byte-identically:

- `Battler.ComputeNetDamage` — the live mitigation applied to every direct hit:
  ```csharp
  var toughness = _attributes[Toughness];
  var toughnessReduction = toughness / (toughness + GameConstants.ToughnessMitigationConstant);
  ```
- `CombatRating.Survivability` — the rating's mitigation fraction:
  ```csharp
  var toughness = effectiveCaster.GetAttributeValue(Toughness);
  var toughnessReduction = toughness / (toughness + GameConstants.ToughnessMitigationConstant);
  ```

## Why the existing comment doesn't already cover this

`Survivability` carries a deliberate note:

> The Toughness curve, inlined rather than through `Battler.ComputeNetDamage`: survivability needs the dimensionless mitigation fraction, not a transformed damage amount, but the tunable knob (`GameConstants.ToughnessMitigationConstant`) is still the single shared constant.

That reasoning is **correct as far as it goes** — the rating genuinely cannot call `ComputeNetDamage`, which returns a damage amount, and sharing the *constant* does protect against the most likely change (a retune of `C`). But it answers a different question than the one #2429/#2439 are about. Those issues aren't about the constant drifting; they're about the *shape* drifting. If the curve ever gains a floor, a second term, or a different functional form — all of which have live design pressure behind them (see #1478's unguarded pole at `Toughness = −C`, and #1483's negative-Toughness amplification) — the constant stays shared while the two expressions silently diverge, and the rating keeps returning a plausible number that no longer prices what the engine does.

That is the identical failure mode #2429 describes, and the one remaining place in `CombatRating` where it's still reachable.

## Suggested direction

Extract the dimensionless fraction onto `Battler` and have both read it:

```csharp
public double ToughnessMitigationFraction =>
    _attributes[Toughness] / (_attributes[Toughness] + GameConstants.ToughnessMitigationConstant);
```

`ComputeNetDamage` then becomes `mitigated * (1 - ToughnessMitigationFraction)` and `Survivability` reads the member directly — which is exactly the split the existing comment gestures at (the rating wants the fraction, the engine wants the transformed amount), just made structural instead of duplicated.

## Constraint

Unlike the crit/execute members added in #2439, this one sits inside `ComputeNetDamage`, which is squarely on the frontend/backend battle-parity surface — so it needs the mirrored extraction in `UI/src/lib/battle/battle-formulas.ts` (`toughnessMitigatedDamage`) and mirrored test scenarios on both sides, per CLAUDE.md. The refactor must leave the arithmetic bit-identical; the existing parity matrix is the check.

Worth confirming while in there whether `Battler.ComputeNetDamage`'s early-return branch (`mitigated <= 0`, where the curve deliberately does not apply) needs any corresponding note on the extracted member, since a bare `ToughnessMitigationFraction` reads as unconditional.

## Scope

Small — one member plus its frontend mirror and mirrored tests, no behaviour change.

Refs #2429, #2439.

ginderjeremiah:
Closes #2450.

Closes the last engine-vs-rating duplication in `CombatRating` after #2429 (parry/dodge) and #2439 (crit/execute).

## The change

`Battler.ToughnessMitigationFraction` — the dimensionless share the curve removes, `Toughness / (Toughness + C)` — is now the single definition:

- `Battler.ComputeNetDamage` applies `mitigated * (1 - ToughnessMitigationFraction)`.
- `CombatRating.Survivability` reads the fraction directly instead of re-deriving the expression.

The **fraction** is the shared member rather than the mitigation step because the rating needs a dimensionless number and the engine needs a transformed damage amount — exactly the split the removed comment gestured at, now structural instead of a note explaining a duplication.

## On the issue's premise

I verified the issue's claim before acting on it, and it holds: the two expressions were byte-identical, and the existing comment's defence (the tunable constant is shared) genuinely answers a different question than #2429/#2439 ask. The constant staying shared protects a retune of `C`; it does nothing if the curve gains a floor or a second term, which is the drift those issues are about. The rating would keep returning a plausible number that no longer prices what the engine does.

## The early-return note the issue asked about

The issue flagged that a bare `ToughnessMitigationFraction` reads as unconditional, since `ComputeNetDamage` deliberately skips the curve on the `mitigated <= 0` branch. Rather than fold the condition into the member (which would make it useless to the rating — the rating has no hit to test), the doc comment states the distinction explicitly: the member is the curve's *value*, not a statement of when it applies, and both the absorption branch and DoT's bypass are named there. The frontend mirror carries the same note.

## Parity

`ComputeNetDamage` is on the frontend/backend battle-parity surface, so `UI/src/lib/battle/battle-formulas.ts` gets the mirrored `toughnessMitigationFraction`, with `toughnessMitigatedDamage` reading it. The arithmetic is unchanged on both sides — same expression, same operand order, so the anti-cheat replay stays bit-exact.

## Docs

`docs/backend-battle.md`'s anti-drift rule (the "`CombatRating` prices the same capabilities through those same accessors" line) now records the Toughness curve as covered, and why the shared member is the fraction.

## Test plan

Mirrored scenarios and expected results on both sides, per CLAUDE.md's battle-logic rule:

| Scenario | `Game.Core.Tests/Battle/BattlerTests.cs` | `UI/src/tests/lib/battle/battle-formulas.test.ts` |
| --- | --- | --- |
| Zero Toughness → exact 0 | ✅ | ✅ |
| Toughness = C → exactly 0.5 (half-point, #1487) | ✅ | ✅ |
| Overwhelming Toughness asymptotes below 1 (no immunity) | ✅ | ✅ |
| Negative Toughness within the pole → negative fraction, unclamped (#1483) | ✅ | ✅ |
| Anti-drift: the fraction and the mitigation step agree exactly | ✅ | ✅ |

The existing `TakeDamage` curve tests and the parity matrix are unchanged and still pass, which is the real check that the arithmetic didn't move.

- `dotnet build` — 0 warnings, 0 errors
- `dotnet test` — 3560 passed, 0 failed (Core 1480, Infrastructure 74, Api 893, Application 1113)
- `npm run test:unit:ci` — 4051 passed, 0 failed
- `npm run lint` (eslint + svelte-check) — 0 errors, 0 warnings

No follow-ups worth filing came out of this one — the refactor was self-contained and `BattlerEffects`' other use of `ToughnessMitigationConstant` (Sunder's `sunder / C` normalization) is a different formula, not a third copy of the curve.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DHvsrY3Myz9MV55JBNdfJR)_


Game.Core/Battle/Battler.cs
```diff
@@ -189,15 +210,10 @@ public double ComputeNetDamage(double dealt, EDamageType damageType)
...
-            // curve is unclamped below 0 — a debuff-driven negative Toughness amplifies the hit (#1483), with
-            // the pole at Toughness = −C left unguarded per #1478 (unreachable by authored content). Both
+            // Toughness mitigation, through the shared fraction so CombatRating cannot price a different curve
+            // (#2450). Applied as a multiplier, so a positive hit can never go negative through it. Both
             // simulators must compute this expression identically for battle parity.
-            var toughness = _attributes[Toughness];
-            var toughnessReduction = toughness / (toughness + GameConstants.ToughnessMitigationConstant);
-
-            return mitigated * (1 - toughnessReduction);
+            return mitigated * (1 - ToughnessMitigationFraction);
```
ginderjeremiah: Nit — the code is deduplicated but the *prose* still isn't. `ComputeNetDamage`'s `<summary>` above (lines 191–197) still spells out the full curve: the multiplier `(1 − Toughness / (Toughness + C))`, the `C` cross-ref, EHP linearity, the sub-100% asymptote, and the #1487/#1330 rationale — all of which now exists verbatim on `ToughnessMitigationFraction`. That's the same drift surface this PR is closing, one layer up: a future retune that adds a floor or a second term has to be edited into two doc blocks or one goes stale, and a stale doc on the *engine* method is the one most likely to be believed.

Suggest trimming the summary's toughness sentence to something like "…then — only while the post-resistance damage is still positive — the <see cref="ToughnessMitigationFraction"/> multiplier `(1 − fraction)`", and letting the member own the curve's rationale. Same for the inline comment here, which can drop to just the parity note since the member now explains the shape.

---

Game.Core/Battle/CombatRating.cs
```diff
@@ -313,11 +313,11 @@ private static double Survivability(bool isPlayer, Battler effectiveCaster)
...
-            // The Toughness curve, inlined rather than through Battler.ComputeNetDamage: survivability needs
-            // the dimensionless mitigation fraction, not a transformed damage amount, but the tunable knob
-            // (GameConstants.ToughnessMitigationConstant) is still the single shared constant.
-            var toughness = effectiveCaster.GetAttributeValue(Toughness);
-            var toughnessReduction = toughness / (toughness + GameConstants.ToughnessMitigationConstant);
+            // The Toughness curve is read off the battler rather than re-derived here (#2450): the rating wants
+            // the dimensionless fraction and the engine wants the transformed damage amount, so the shared
+            // member is the fraction and Battler.ComputeNetDamage is what applies it to a hit. Uncapped, unlike
+            // avoid/resist below — the curve asymptotes below 1 on its own, so it needs no credit ceiling.
+            var toughnessReduction = effectiveCaster.ToughnessMitigationFraction;
```
ginderjeremiah: Nit — the local is now a pure pass-through that *renames* the shared member (`…MitigationFraction` → `toughnessReduction`), which is a small step backwards from the point of the change: at the use site on line 324 nothing signals that this term comes from `Battler` while `cappedAvoid`/`cappedResist` are computed here. Either inline it (`… * (1 - effectiveCaster.ToughnessMitigationFraction)`) or keep the member's name (`toughnessFraction`). The `cappedAvoid`/`cappedResist` locals earn their existence — they apply the cap; this one doesn't do anything.

The "Uncapped, unlike avoid/resist below" clause is a genuinely useful addition — that asymmetry was previously unexplained. Worth noting it holds *within the pole*: past `Toughness = −C` the fraction exceeds `1` and `mitigationFraction` goes negative into the `Math.Max(…, Epsilon)` floor. That's the #1478 unreachable-by-authored-content case, so no guard is warranted, but "asymptotes below 1 on its own" is stated as unconditional.

---

docs/backend-battle.md
```diff
@@ -40,7 +40,7 @@ The typed DoT phase loops the three types in the fixed `DamageTypes.DotAccumulat
...
...als `0` and the riposte fires (below); parry is checked **before** dodge and takes precedence, so dodge investment never starves the riposte. *Dodge:* `d < DodgeChance × DodgeChanceMultiplier` (dodge rework #1523) ⇒ the hit deals `0`. Chances are decimal fractions compared directly against a `[0,1)` draw. Each of these products — the two avoidance chances, the crit chance, and the Cull execute investment (`ExecuteBonus × missingHpFraction`) — is composed at **one** site per simulator (like the cadence rate below) and left **unclamped**, since a `[0,1)` draw saturates naturally. The server-only `CombatRating` prices the same capabilities through those same accessors rather than re-deriving them, but clamps the chances to `[0,1]` itself, since it prices an expectation rather than drawing (#2429, widened to crit and execute by #2439, and to the Toughness curve by #2450 — the shared member there is the dimensionless `ToughnessMitigationFraction`, since the rating wants the fraction and the engine wants the transformed damage amount). On the frontend the crit composition is a free function over `BattleAttributes` so the skill tooltip and skills grid read it too, not just the simulation.
```
ginderjeremiah: Should-fix — the new clause is nested inside a sentence whose main clause is *"but **clamps** the chances to `[0,1]` itself"*, so "widened … to the Toughness curve by #2450" reads as extending the clamping rule to the Toughness fraction. It doesn't: `Survivability` consumes `ToughnessMitigationFraction` **uncapped**, which the code comment you added at `CombatRating.cs:316` says explicitly. Two adjacent artifacts asserting opposite things about the same term is exactly the drift this PR exists to prevent.

Splitting it out keeps the shared-accessor rule and the clamping caveat from colliding — e.g. end the existing sentence at `#2439`, then: "#2450 extends the shared-accessor half to the Toughness mitigation curve (`ToughnessMitigationFraction` — the rating wants the dimensionless fraction, the engine the transformed damage amount); that one the rating consumes uncapped, since the curve asymptotes below `1` on its own."

Also worth reconsidering the placement: the host bullet is *"The math is player-only, gated on who acts"*, and Toughness mitigation is neither player-only nor gated on the actor. It's landing here only because the bullet happens to host the anti-drift sentence.

---

ginderjeremiah (commented):
## Review: #2450 — share the Toughness mitigation curve between `Battler.ComputeNetDamage` and `CombatRating.Survivability`

**Verdict: Approve** — the refactor is correct, bit-exact, mirrored on both sides, and closes the last engine-vs-rating duplication. Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR author; treat it as an approve with one should-fix and a few nits, all documentation-level.

Not merging on this pass: the doc-vs-code contradiction in `backend-battle.md` is worth fixing here rather than as a follow-up, since it's a two-line edit to the artifact this PR is adding.

I verified the load-bearing claims rather than reading the diff on its own terms:

- **The arithmetic really is bit-identical.** Backend: `toughness / (toughness + C)` with a local becomes `_attributes[Toughness] / (_attributes[Toughness] + C)` — same operands, same order, and `AttributeCollection.GetAttributeValue` caches per node, so the second read is an array index plus a null check, not a recomputation, and cannot observe a different value mid-expression. Frontend: `toughnessMitigatedDamage` delegates to a function containing the identical expression. Nothing to worry about for the anti-cheat replay.
- **`Survivability`'s substitution is exact.** `GetAttributeValue(Toughness)` and the property's `_attributes[Toughness]` are the same call on the same object, and `effectiveCaster` is unchanged.
- **The curve isn't duplicated anywhere else.** `ToughnessMitigationConstant` has three other readers: `GameConstants`' own declaration, `Enums.cs`'s `Toughness` doc, and `BattlerEffects`' Sunder normalization (`sunder / C`) — which is a different formula, not a third copy, exactly as the PR body says. Frontend-side, `skills-view.svelte.ts:104,431` and `SkillTooltip.svelte:106` already route through `toughnessMitigatedDamage`, so they inherit the shared fraction for free.
- **No rating-side test gap.** `CombatRatingTests` has no anti-drift pins for the crit/execute members from #2439 either, and it doesn't need one: the rating now *reads* the member, so there's no expression left that could drift. The pin belongs on the `Battler` side, which is where it was put.
- **The frontend pin is already complete.** `battle-formulas.test.ts:255` pins `mitigateDamage` against `toughnessMitigatedDamage`, so the full chain `mitigateDamage → toughnessMitigatedDamage → toughnessMitigationFraction` is covered — the frontend mirror of the backend's `ComputeNetDamage` pin exists, just one test up.

### Should-fix

- **`docs/backend-battle.md:43` — the new clause reads as extending the `[0,1]` clamp to the Toughness fraction, which contradicts the code comment added in the same PR.** `Survivability` consumes the fraction uncapped, and `CombatRating.cs:316` now says so explicitly. Details and a suggested split inline.

### Nits

- **`Game.Core/Battle/Battler.cs:191–197` — `ComputeNetDamage`'s `<summary>` still restates the entire curve** (formula, `C` cross-ref, EHP linearity, asymptote, #1487/#1330 rationale) that now lives verbatim on `ToughnessMitigationFraction`. The code is deduplicated; the prose isn't. Inline.
- **`Game.Core/Battle/CombatRating.cs:320` — the pass-through local renames the shared member** (`…MitigationFraction` → `toughnessReduction`), hiding at the use site that this term comes from `Battler` while its two neighbours are computed locally. Inline, along with a note that "asymptotes below 1 on its own" holds only within the pole.
- **`Game.Core/Enums.cs:54`** — `Toughness`'s doc still points at `Battler.ComputeNetDamage` for the curve. The definition moved; the cross-ref should be `Battler.ToughnessMitigationFraction` (`ComputeNetDamage` is now the *application* site, not the definition).
- **`Game.Core.Tests/Battle/BattlerTests.cs:13–21`** — the class doc enumerates which members are mirrored and names `battler.test.ts` as the counterpart file. The new cases mirror `battle-formulas.test.ts` instead, and `ToughnessMitigationFraction` isn't in the list. Pre-existing looseness (the #2439 crit/execute members have the same mismatch), so fine to fold in here or leave — but the enumeration is only useful while it's exhaustive.

### Test coverage

Adequate and properly mirrored, which is the rule that actually binds here (CLAUDE.md's battle-parity requirement). Five scenarios on each side with identical inputs and expected results: zero → exact `0`, `Toughness = C` → exactly `0.5`, overwhelming Toughness → `2000/2200` and strictly `< 1`, negative-within-the-pole → `−1`, and the anti-drift pin that the fraction and the mitigation step agree. The backend pin goes through the real `ComputeNetDamage` rather than re-asserting the expression, which is the version that would actually catch a future re-inlining. The `MakeBattler((Toughness, 350))` cases set Toughness directly rather than deriving it from Endurance like the older `TakeDamage` tests — correct here, since Endurance is `0` so no static derivation lands on top, and it isolates the curve from the derivation.

Both pins are structurally tautological *today* (both sides call the shared definition), which is fine — their job is to fail the moment someone re-derives the expression locally, and they will.

### What's good

Choosing the **fraction** as the shared member rather than trying to force the rating through `ComputeNetDamage` is the right call, and it's the one the issue's own suggested direction gets right for the right reason: the rating needs a dimensionless number, the engine needs a transformed amount, so the shared thing has to be the piece upstream of that split. Turning the old comment's excuse into structure — rather than deleting the comment and hoping — is the cleaner outcome.

The issue's explicit open question (does a bare `ToughnessMitigationFraction` read as unconditional, given `ComputeNetDamage`'s `mitigated <= 0` early return?) got a real answer instead of being quietly skipped, and the answer is the correct one: folding the condition into the member would have made it useless to the rating, which has no hit to test. The `<para>` distinguishing "the curve's value" from "when it applies", and naming both the absorption branch and DoT's bypass, is precisely the note that was needed — and it's mirrored in the frontend JSDoc, so neither side loses it.

Scope discipline is good too: the diff is five files, no drive-by changes, and the PR body states plainly that the arithmetic didn't move and that the existing parity matrix is the check — which is the right claim to make, since that matrix passing is worth more than the new tests here.

---

docs/backend-battle.md
ginderjeremiah: Fixed in `9b0c78b` — and you're right on both counts, the second one more than the first.

The clamp collision was a genuine contradiction: the sentence's main clause is the clamping caveat, so anything appended reads as governed by it. The existing sentence now ends at `#2439`.

On placement — I took the stronger version of your point rather than just splitting the sentence in place. The clause was in that bullet only because the bullet happened to host the anti-drift sentence, which isn't a reason. It's now on the mitigation-pipeline bullet (**"Toughness curve last, and only on a positive hit"**), which is where the curve is actually described and already carries the parity note for the same expression:

> #2450 extends the shared-accessor half of the anti-drift rule above to this curve: the dimensionless `Battler.ToughnessMitigationFraction` (↔ `toughnessMitigationFraction`) is the one definition — the rating wants the fraction, the engine the transformed damage amount — and `CombatRating.Survivability` consumes it **uncapped**, unlike the chances it clamps, since within the pole the curve asymptotes below `1` on its own.

That puts the anti-drift fact next to the thing it's about instead of next to the rule it's an instance of, and the "uncapped, unlike the chances it clamps" contrast now states the asymmetry directly rather than inheriting a clamp claim by proximity.

---

Game.Core/Battle/Battler.cs
ginderjeremiah: Fixed in `9b0c78b`. Agreed, and the framing is the convincing part — deduplicating the code while leaving the prose duplicated just moves the drift surface up a layer, and a stale doc on the *engine* method is indeed the one that gets believed.

The summary's toughness sentence is now:

> …then — only while the post-resistance damage is still positive — the `<see cref="ToughnessMitigationFraction"/>` multiplier `(1 − fraction)`, which owns the curve's shape and rationale so they are stated once.

The `C` cross-ref, EHP linearity, the asymptote and the #1487/#1330 rationale all live on the member now and nowhere else. The inline comment dropped to just the parity note, since the member explains the shape — and the "so CombatRating cannot price a different curve" half was redundant with the member's own doc anyway.

---

Game.Core/Battle/CombatRating.cs
ginderjeremiah: Fixed in `9b0c78b` — inlined rather than renamed. The local existed only because it was there before the refactor, and once it stopped computing anything it was actively working against the change: `cappedAvoid`/`cappedResist` name a *transformation*, so a third local in the same shape implied one that wasn't happening. The product now reads `(1 - cappedAvoid) * (1 - cappedResist) * (1 - effectiveCaster.ToughnessMitigationFraction)`, split across lines for width, so the one term that comes from `Battler` is visibly the one that isn't computed here.

Good catch on the pole — "asymptotes below 1 on its own" was stated unconditionally and isn't. The comment now scopes it and names the consequence:

> Consumed uncapped, unlike avoid/resist — within the pole the curve asymptotes below 1 on its own, so it needs no credit ceiling. Past Toughness = −C the fraction exceeds 1 and drives mitigationFraction negative into the Epsilon floor below; that is #1478's unreachable-by-authored-content case.

Agreed no guard is warranted — adding one here would be a second place to maintain a policy #1478 already decided, and it would diverge from the engine, which doesn't guard it either. Worth having written down at the site that silently absorbs it, though.

---

PR #2478, opened 2026-07-26
Add a test harness for the build/*.ps1 coverage tooling (#2453)

### LINKED ISSUE #2453: build/*.ps1 has logic but no test harness
Raised while implementing #2440 / #2441 in PR #2451.

## Problem

The `build/` PowerShell scripts carry real branching logic that gates CI, and none of it is tested:

| Function | File | Logic |
|---|---|---|
| `Get-NamespaceCoverage` | `coverage-lib.ps1` | Aggregates line/branch coverage from per-class report entries; prefix matching, zero-denominator guards, rounding to match ReportGenerator's own figures. |
| the gate itself | `coverage-gate.ps1` | Floor comparison, gated vs. gated-namespace vs. measure-only classification, the three distinct exit codes. |
| floor suggestion | `coverage-floor-suggestions.ps1` | Recomputes floors from raw covered/coverable counts with a held-back margin. |
| `Write-TestFailureDetail` | `coverage.ps1` (new in #2451) | Selects which `TestResults` logs to dump and filters progress lines out of them. |

CLAUDE.md requires tests for logic, and the repo already accepts that shell-level build tooling is worth testing — `.claude/hooks/session-start.test.sh` is a hand-rolled bash harness with its own `test-hooks` CI job, added in #2395 for exactly this reason. The PowerShell side never got the equivalent.

The concrete failure mode is silent, not loud: a subtly wrong `Get-NamespaceCoverage` doesn't crash, it reports a plausible number and the gate keeps passing. Same shape as the drift argument in #2429 / #2450.

## Why it wasn't done in #2451

That issue was `scope: small` — "CI/workflow plus possibly a few lines in `build/coverage.ps1`". Standing up a harness plus a CI job would have been several times the size of the change it was attached to, and would have buried a one-purpose diff. #2451's own function was instead kept to two filter predicates and verified by running the real command against a real failing test run.

## Suggested direction

Two viable shapes, and the choice is the actual decision here:

1. **Pester** — the standard, ships with PowerShell 7, expressive. Adds a dependency and a version to pin, and the scripts must run on Windows PowerShell 5.1 too (which has Pester 3.4 preinstalled — a *different, incompatible* dialect), so the harness would effectively be pwsh-7-only. That's fine for CI but splits the "one command works everywhere" property the coverage scripts are careful about.
2. **A hand-rolled `build/coverage-lib.test.ps1`**, mirroring `session-start.test.sh` — no dependency, same convention as the existing hook harness, runs anywhere, but reimplements assertion plumbing.

Either way the target is the pure functions, fed synthetic report/log fixtures rather than a real coverage run. `Write-TestFailureDetail` would need its file-enumeration split from its line-filtering to be testable without touching disk; the coverage functions are already pure.

Then a CI job alongside `test-hooks` — cheap, since it needs no build and no backing services.

## Scope

Medium — new harness, a CI job, and the first round of tests. No production code, no behaviour change.

ginderjeremiah:
Closes #2453.

The `build/` PowerShell scripts gate CI on coverage and had no tests. The failure mode the issue describes is the one I designed against: a subtly wrong `Get-NamespaceCoverage` doesn't crash, it reports a plausible number and the gate keeps passing. So every case asserts on a value or an exit code, never merely that a script ran.

## The design fork — I went hand-rolled

The issue framed Pester vs. hand-rolled as "the actual decision here". I picked **hand-rolled** (`build/coverage.test.ps1`, no dependency), for a reason I think is decisive rather than stylistic:

Every `build/*.ps1` declares `#Requires -Version 5.1`, and `docs/infrastructure.md` advertises these scripts as running on Windows PowerShell 5.1 *and* `pwsh` 7. Pester 7 can't run under 5.1 — 5.1 preinstalls the incompatible 3.4 dialect — so a Pester harness would be structurally unable to exercise the very host the 5.1 support exists for. That's not "fine for CI, splits the one-command property" as the issue put it; it means the 5.1 code path would be permanently untested. A dependency-free harness runs wherever the scripts it tests run.

The cost, accepted knowingly: the assertion plumbing is ours to maintain. Two review nits landed exactly there and are fixed.

## A crash the harness surfaced

`@($section.PSObject.Properties.Name)` yields a **one-element array holding `$null`** when a `coverage-floors.json` section is absent *or* present-but-empty — `.Name` member-enumerates to `$null` in both cases, not to an empty collection. Both `coverage-gate.ps1` and `coverage-floor-suggestions.ps1` then iterate once with a null name and index the report by `$null`. The gate died with an opaque `Index operation failed; the array index evaluated to null.` and exit 1 — indistinguishable from a coverage breach.

I hit this for real: a `gated = {}` fixture crashed the gate before the fix. But calibrating the severity honestly — `gated` is never absent or empty in a real `coverage-floors.json`, so this is **defensive centralization, not a live bug fix**; it's reachable by hand-editing a section empty, not by any state the repo produces. Fixed here rather than filed because it's four lines in the code under test. Now `Get-FloorSectionNames`, with unit cases plus an end-to-end gate case asserting an all-empty config exits 0.

## What moved, and why

Only what testability required — no behaviour change beyond the above:

| Function | From | Why |
|---|---|---|
| `Get-SuggestedFloor` | `coverage-floor-suggestions.ps1` | that script self-executes on dot-source, so the function was unreachable from a test |
| `Get-FloorRatchet` | (extracted from `Show-Suggestion`) | the ratchet rule — "a floor only ever rises" — is the load-bearing logic; `Show-Suggestion` keeps only the formatting |
| `Test-TestLogProvenGreen`, `Select-TestLogDetailLine` | `coverage.ps1` | the issue's requested split of `Write-TestFailureDetail`'s file enumeration from its line filtering |
| `Get-FloorSectionNames` | (new) | the guard above, shared by both consumers |

Both array-returning helpers use a leading comma (`return ,@(...)`) so a single-member result stays an array instead of unrolling to a bare string — callers take `.Count` on it, which works on scalars only by PowerShell's grace.

`coverage-gate.ps1` needed **no** refactor: its logic is its script body and its contract is three distinct exit codes, so it's driven end-to-end as a child process against synthetic `Summary.json`/floors fixtures. A child process rather than a dot-source is what makes `exit N` observable as an exit code instead of terminating the harness. `coverage-floor-suggestions.ps1` is driven the same way.

One incidental tidy: `coverage-floor-suggestions.ps1` was double-guarding its namespaces section, and an empty-but-present section would print a header with nothing under it. It now gates on the name count — and that change is itself tested.

## Coverage — 80 assertions

- **`Get-NamespaceCoverage`**: multi-prefix aggregation, percentages recomputed from summed counts (not averaged per-class), 1-decimal rounding matching ReportGenerator, both zero-denominator guards independently, empty-bucket state, and that the trailing dot in the prefix keeps `Game.Api.SocketsLegacy` out of `Game.Api.Sockets`' scope.
- **`Get-SuggestedFloor`**: margin held back then floored, full-precision actual (the whole point — raw counts, not the truncated display value), null/zero denominators, negative clamp.
- **`Get-FloorRatchet`**: all three directions, including that a below-current suggestion keeps the existing floor.
- **`Get-FloorSectionNames`**: populated, empty, absent, and single-member (array contract preserved).
- **Test-log filtering**: passing summary proves green; a failing *and* a truncated/crashed run with no summary line both do not; progress-spinner and blank lines dropped while the test name, assertion and stack survive; single surviving line stays an array.
- **`coverage-gate.ps1` end-to-end**: exit 2 vs 1 vs 0 kept distinct, line and branch breaches independently, gated assembly missing from the report, branch floor configured with no branch data, gated-namespace breach, emptied namespace bucket, missing backing assembly, measure-only assemblies never failing the gate, all-empty config.
- **`coverage-floor-suggestions.ps1` end-to-end**: both section headers, `(raise)` and `below current` as rendered by the real script, full-precision `97.2000%`, the paste-safety hint, both missing-assembly warnings, exit 2 on a missing summary, and the empty-section header suppression.

## Test plan

- `pwsh build/coverage.test.ps1` → **80 passed, 0 failed**, exit 0.
- **Mutation-tested the harness**, because tests that can't fail are worse than none. Dropping the trailing dot in the prefix match, rounding to 2 decimals, letting the ratchet move down, treating any summary line as green, reverting the array contract, and reverting the header suppression each produced a failure in exactly the intended assertion; reverting each returned it to green.
- **Verified the assertion plumbing itself fails where it should**: `Assert-Contains`/`Assert-NotContains` on empty or `$null` output now FAIL rather than vacuously passing (a child process that never launched would previously have read as green), and `Assert-Equal` on a type mismatch reports one failure and lets the run continue instead of aborting under `$ErrorActionPreference = 'Stop'`.
- **Verified CI wiring propagates failure**: simulated GitHub's `shell: pwsh` wrapper (`. '<step>.ps1'` + `exit $LASTEXITCODE`) against both a green harness (step exit 0) and a deliberately broken `coverage-lib.ps1` (step exit 1). A test job that silently passes on regression would have been the worst outcome here.
- **Regression-checked the refactored scripts** against the real `coverage-floors.json` with a realistic synthetic summary: `coverage-gate.ps1` reports all 3 gated assemblies and 2 gated namespaces PASS and exits 0, and `coverage-floor-suggestions.ps1`'s output is unchanged, with `(raise)` / `(no change)` / `(below current …)` all still rendering.
- Parse-checked all five `build/*.ps1`; workflow YAML validated and the new job confirmed present in `all-checks-passed`'s `needs`.
- No C# changed, so `dotnet build`/`dotnet test` are unaffected and were not run. `build/**` is in the `backend` path filter, so `test-backend` still runs on this PR (green).

New `test-build-scripts` CI job mirrors `test-hooks` — needs no build, no .NET and no backing services, gated on a new `buildScripts` path filter, and added to `all-checks-passed`. Green in 22s.

## One thing deliberately left

**`Write-TestFailureDetail`'s file enumeration is still untested** (the `*/bin/*/*/TestResults/*.log` glob) — #2453 itself expected that to stay untested once the filtering was split out, and the filtering decisions it delegates are covered. Testing the glob needs a fixture directory tree, which is a different kind of test than the rest of this harness.

The pre-existing 5.1 mojibake I flagged (BOM-less UTF-8 files containing em-dashes) is now #2479, out of scope here.


build/coverage-lib.ps1
```diff
@@ -1,9 +1,23 @@
...
+  Everything here is a pure function of its arguments — no disk, no host state — so
+  build/coverage.test.ps1 can drive it with synthetic fixtures. The scripts that own side effects
+  (reading the report, enumerating logs, writing to the host, choosing an exit code) keep those in
+  their own bodies and call in here for the arithmetic and the filtering decisions.
 #>
 
+# Member names of a coverage-floors.json section ("gated" / "gatedNamespaces"), as a genuinely empty
+# array when the section is absent or has no members. @($section.PSObject.Properties.Name) alone
+# yields a *one-element array holding $null* in both those cases, so callers would iterate once with a
+# null name and then index the report by $null.
```
ginderjeremiah: Nit — I think the comment (and the PR description) overstates the failure it fixes. `@($section.PSObject.Properties.Name)` yields the one-element-`$null` array when the section is **absent** (`$null.PSObject.Properties.Name` → `$null`), but an *empty* `{}` already member-enumerates to an empty collection, so `@(...)` is genuinely `@()` there. The test at `coverage.test.ps1:237` asserts `0` for both, which passes either way and so doesn't distinguish them.

The guard is still right and worth having — it just isn't load-bearing for the empty case. Worth trimming "in both those cases" to the absent case so a future reader isn't looking for a bug that isn't there.

Related: the only absent-section path that was actually unguarded was `$floors.gated` (`gatedNamespaces` already had its `$null` check), and a `coverage-floors.json` with no `gated` key at all is a malformed config rather than a reachable state — so this reads more as defensive centralization than as a latent-bug fix. No objection to landing it here; it's four lines in the code under test.

---

build/coverage.test.ps1
```diff
@@ -0,0 +1,405 @@
...
+
+. (Join-Path $PSScriptRoot 'coverage-lib.ps1')
+
+$GateScript = Join-Path $PSScriptRoot 'coverage-gate.ps1'
+$Work = Join-Path ([System.IO.Path]::GetTempPath()) ("coverage-tests-" + [guid]::NewGuid().ToString('N'))
+New-Item -ItemType Directory -Force -Path $Work | Out-Null
+
+# The executable of the host currently running, so the gate's child process runs under the same
+# PowerShell edition as the harness (powershell.exe on 5.1, pwsh on 7) rather than a hard-coded one.
+$HostExe = [System.Diagnostics.Process]::GetCurrentProcess().MainModule.FileName
```
ginderjeremiah: Nit — `MainModule.FileName` is the *process* executable, which is only the PowerShell binary when the host happens to be the console host. Under an embedded host it isn't: in the ISE this resolves to `powershell_ise.exe`, and any host that loads PowerShell in-process resolves to that host's exe, so `& $HostExe -NoProfile -File ...` would launch the wrong program and every gate case would fail confusingly rather than reporting a real regression.

`$PSHOME` is the host-independent answer and stays 5.1/7 agnostic:

```powershell
$HostExe = Join-Path $PSHOME (if ($PSVersionTable.PSEdition -eq 'Core') { 'pwsh' } else { 'powershell.exe' })
```

Doesn't matter for CI (console host both ways) — it matters for the local-run promise this harness is built around.

---

build/coverage.test.ps1
```diff
@@ -0,0 +1,405 @@
...
+
+  Write-Host ''
+  Write-Host '=== Get-FloorSectionNames (empty / absent floor sections) ==='
+
+  # An empty or absent section must mean "nothing gated here". @() around
+  # .PSObject.Properties.Name yields a one-element array holding $null instead, which made both
+  # the gate and the reseed script iterate once with a null name and index the report by $null.
+  Assert-Equal 'a populated section lists its members' 2 (Get-FloorSectionNames (ConvertFrom-Json '{"Game.Core":{},"Game.Application":{}}')).Count
+  Assert-Equal 'an empty section yields no names' 0 (Get-FloorSectionNames (ConvertFrom-Json '{}')).Count
+  Assert-Equal 'an absent section yields no names' 0 (Get-FloorSectionNames $null).Count
```
ginderjeremiah, line 271: Nit — the missing case is a section with exactly **one** member. PowerShell unrolls a single-element array on return, so `Get-FloorSectionNames` hands back a bare string there, not an array — and both callers then take `.Count` on it (`coverage-gate.ps1:131`, `coverage-floor-suggestions.ps1:74`). It works today because PS 3+ gives scalars a `.Count` of 1, but that's exactly the kind of quiet dependence this harness exists to pin down, and the helper's own comment promises an "array". A third assertion (`1` for `'{"Game.Core":{}}'`) costs a line and locks it in.

If you'd rather remove the sharp edge than test it, `,@(...)` on the return preserves the array through the unroll.

---

build/coverage.test.ps1
```diff
@@ -0,0 +1,405 @@
...
+  }
+}
+
+function Assert-NotContains($name, $unexpected, $text) {
+  if ($text -and $text.Contains($unexpected)) {
+    Write-Fail $name "output unexpectedly carried '$unexpected'"
+  } else {
+    Write-Pass $name "output free of '$unexpected'"
+  }
+}
```
ginderjeremiah, line 106: Nit, two small holes in the assertion plumbing — both matter more than usual here, since a harness that can't fail is the thing this PR is guarding against:

1. `Assert-NotContains` **passes** when `$text` is empty or `$null` (the `-and` short-circuits into the else). A gate case that produced no output at all — the child process failing to launch, say — would report a clean PASS. `Assert-Contains` gets this right by construction; this one needs an explicit empty check to match.
2. `Assert-Equal` compares with `$expected` on the left, so PowerShell coerces the *actual* to the expected's type. When a function returns an unexpected type (`Assert-Equal 'x' 4 'some string'`) the coercion throws, and with `$ErrorActionPreference = 'Stop'` that aborts the whole run instead of printing FAIL and continuing. Wrapping the comparison in a `try`/`catch` that routes to `Write-Fail` keeps a type surprise reportable.

---

ginderjeremiah (commented):
## Review: test harness for the `build/*.ps1` coverage tooling (#2453)

**Verdict: Approve** — the harness does what #2453 asked for, the refactors are the minimum testability required, and every case asserts on a value or an exit code. Nothing blocking; the notes below are all nits. (Submitted as a comment because GitHub won't take a gating verdict from the PR author's own account — treat it as an approve.)

### On the Pester fork

Agreed, and I'd have made the same call. The argument is decisive rather than stylistic: `#Requires -Version 5.1` plus `docs/infrastructure.md`'s dual-host claim means a Pester harness would leave the 5.1 path permanently untested, which is the one host the requirement exists for. It also matches the precedent already set by `session-start.test.sh`. Worth noting the cost you're accepting: the assertion plumbing is now ours to maintain, and two of the nits below are exactly that cost showing up.

The CI wiring is sound — `shell: pwsh` + a script ending in `exit N` is the same mechanism `coverage.ps1 -NoBuild` already relies on in `test-backend`, so the propagation path is proven in this repo, not just in your simulation.

### Nits

- `build/coverage-lib.ps1:13` — the comment (and the PR body) overstates the empty-section case; only an *absent* section produces the one-element-`$null` array. Inline.
- `build/coverage.test.ps1:42` — `MainModule.FileName` resolves to the host exe, not the PowerShell binary, under embedded hosts. `$PSHOME` is host-independent. Inline.
- `build/coverage.test.ps1:236` — no case for a single-member section, where the return unrolls to a scalar and callers still take `.Count` on it. Inline.
- `build/coverage.test.ps1:81` — `Assert-NotContains` passes on empty output; `Assert-Equal` can throw instead of failing on a type mismatch. Inline.
- `coverage-floor-suggestions.ps1`'s script body is the one piece of changed production code with no test at all — including the empty-section header suppression this PR just changed (`if ($gatedNsNames.Count -gt 0)`) and the missing-assembly warning. Its extracted logic is well covered, so this is only the ~10 lines of orchestration and formatting, and `Invoke-Gate` already proves the child-process pattern that would cover it. Cheap to add here; fine as a follow-up if you'd rather keep the diff at one script.

### Test coverage

Genuinely good, and the mutation testing is the part that makes me believe it — a harness whose assertions can't fail is worse than none, and you checked rather than assumed. Prefix-boundary (`SocketsLegacy`), the two zero-denominator guards taken independently, full-precision `Actual` vs. the truncated display value, and "absence of a summary line is not proof of green" are all the right edges. On the gate side, keeping exit 2 distinct from exit 1 is the assertion that stops "you didn't run coverage" from ever reading as "coverage regressed".

The stated gaps are the right ones to have left: `Write-TestFailureDetail`'s glob enumeration is what #2453 itself expected to stay untested once the filtering was split out, and its filtering decisions are covered. I've filed #2479 for the pre-existing 5.1 mojibake you flagged — out of scope here, and correctly left alone.

### What's good

- Fixture builders use ReportGenerator's exact lower-cased property names, with a comment explaining that a rename there would be a silently passing test. That's the right paranoia for a harness whose whole purpose is catching silent wrongness.
- Driving `coverage-gate.ps1` as a child process instead of dot-sourcing it is the correct call, and the reason (`exit N` would otherwise terminate the harness) is written down where the next person will need it.
- The refactor is disciplined — pure logic into `coverage-lib.ps1`, side effects left in the scripts that own them, and `coverage-gate.ps1` correctly left alone rather than contorted for testability. The label/format change in `Show-Suggestion` preserves the rendered output exactly.
- No dead code left behind: `Get-SuggestedFloor` moved rather than copied, and the double-guard on the namespaces section is gone.

---

ginderjeremiah:
Thanks — four of the five nits were right and are fixed in 3dc49a1. The fifth I think is mistaken, and it's the one load-bearing enough to be worth showing my working on.

## One pushback: the empty `{}` case *is* affected

The claim was that only an *absent* section produces the one-element-`$null` array, because an empty `{}` "already member-enumerates to an empty collection." It doesn't — an empty property collection member-enumerates to `$null` exactly as an absent one does:

```
$empty = ConvertFrom-Json "{}"
$empty.PSObject.Properties       -> not null (an empty PSMemberInfoCollection)
$empty.PSObject.Properties.Name  -> $null          # <-- the step in question
@($empty.PSObject.Properties.Name).Count -> 1, element[0] is $null
```

Same result as `$null.PSObject.Properties.Name`. So `Assert-Equal 'an empty section yields no names' 0` does discriminate: without the guard it sees `1` and fails.

That isn't just theory — the empty case is how I found this. The gated-namespace fixtures pass `gated = [pscustomobject]@{}`, which serializes to `"gated": {}`, and against unfixed code the gate died with:

```
coverage-gate.ps1: Index operation failed; the array index evaluated to null.
exit=1
```

That's `$byName[$null]` at `coverage-gate.ps1:54`, reached with `$name` = `$null` — an opaque crash indistinguishable from a coverage breach. So I've kept the guard's scope as-is and instead made the comment spell out the mechanism, since the misreading is an easy one:

> `.Name` yields `$null` for a section that is absent *and* for one that is present but empty (an empty property collection member-enumerates to `$null`, not to an empty collection)

Your **related** point stands and is the better framing of the severity, though: with `gated` never absent in a real `coverage-floors.json`, this is defensive centralization rather than a live bug fix. I've stopped calling it a "latent bug fix" — reachable via a hand-edited config that empties a section, not via any state the repo produces today.

## The four that were right

- **`$PSHOME` over `MainModule.FileName`** — correct, and the failure mode is nastier than a wrong path: it would launch `powershell_ise.exe` and fail *every* child-process case, which reads as a broken harness rather than a real regression. Fixed. I used a separate assignment rather than the inline `(if …)` from your snippet, since statement-as-expression in a parenthesized argument doesn't parse on 5.1 — the host this exists for.
- **Single-member unroll** — confirmed: `Get-Old` on `{"Game.Core":{}}` returns `System.String`, not an array. I took the remove-the-edge option *and* the test, since `.Count` working on scalars is exactly the quiet dependence worth eliminating rather than pinning. Applied the same leading comma to `Select-TestLogDetailLine` for the same contract, which required dropping a now-double-wrapping `@()` at its call site.
- **Assertion plumbing** — both real. `Assert-NotContains` passing on empty output was the worse of the two given what this PR is for; both text asserts now fail explicitly when there's no output. On `Assert-Equal`: I couldn't reproduce the throw on 7 (`4 -eq 'not-a-number'`, `-eq @{}`, `-eq [datetime]` all return `$false`), so I kept the `try`/`catch` as cheap insurance but worded the comment as 5.1-conversion-rules-may-differ rather than asserting a throw I can't demonstrate.
- **`coverage-floor-suggestions.ps1` untested** — fair, and it was the one piece of *changed* production code with no coverage, so I did it here rather than deferring. Eight new cases via the `Invoke-BuildScript` runner (generalized from `Invoke-Gate`, since both scripts take the same `-SummaryPath`/`-FloorsPath` pair): both section headers, `(raise)` and `below current` rendered by the real script, the full-precision `97.2000%`, the paste-safety hint, both missing-assembly warnings, exit 2 on a missing summary, and the empty-section header suppression this PR introduced.

## Verification

**80 passed, 0 failed** (up from 61), and I mutation-tested the new assertions rather than trusting them:

| Mutation | Result |
|---|---|
| revert `,@()` on `Get-FloorSectionNames` | `a single-member section stays an array` fails |
| revert header suppression to the old `$null`-only guard | `an empty namespaces section prints no header` fails |
| `Assert-NotContains` fed empty / `$null` output | now FAILs (previously a silent pass) |
| `Assert-Equal` fed a type mismatch | FAILs and the run continues rather than aborting |

Also re-ran the real `coverage-gate.ps1` and `coverage-floor-suggestions.ps1` against the real `coverage-floors.json` after the `,@()` change — gate still passes 3 gated assemblies + 2 gated namespaces at exit 0, and the reseed output is unchanged. CI was green on a7cfe88 including `test-build-scripts` (22s); re-running on 3dc49a1.

Thanks for filing #2479 for the 5.1 mojibake.

---

PR #2480, opened 2026-07-26
Consolidate the IChallenge/ICreatableClass test builders onto shared fixtures (#2456)

### LINKED ISSUE #2456: Consolidate the hand-rolled IChallenge test builders onto a shared fixture

ginderjeremiah:
Closes #2456.

Test-only: no production code, no behaviour change. Adds `UI/src/tests/fixtures/challenges.ts` (`makeChallenge`) and `UI/src/tests/fixtures/classes.ts` (`makeCreatableClass`), and converts every hand-rolled builder for the two contracts onto them.

## Verification (the point of the exercise)

The target set was confirmed with the throwaway-required-field probe rather than the issue's grep. Adding a required field to `IChallenge` **before** this change flushed out four suites and, separately, silently compiled at nine cast sites. **After**, the same probe produces exactly one test-file error:

```
src/routes/admin/workbench/entities/challenge.ts   ← production blank record (correctly stays a literal)
src/tests/fixtures/challenges.ts                   ← the only test-side edit
src/tests/fixtures/classes.ts                      ← ICreatableClass, previously 0 errors (all 3 sites cast-silenced)
```

`ICreatableClass` is the more interesting half: all three of its sites asserted their literal, so the contract had **no** test-side drift detection at all — the probe found nothing to report.

## Where the issue's scope needed adjusting

The issue lists 15 suites that "construct an `IChallenge`". Two of those groups are excluded on purpose, matching the precedents the series already set:

- **`entities/challenge.test.ts`** spreads the production `challengeEntity.newItem(...)` rather than hand-rolling a record. That's the authoritative shape (the `fixtures/item-mods.ts` exclusion for the same reason), so it stays.
- **`Codex.test.ts` / `codex-view.svelte.test.ts`** seed `staticData: {} as any`. An untyped mock never sees a new required field, and filling its previously-`undefined` fields could change what the suite renders.

The `challenges: [] as IChallenge[]` annotations in the skills/fight suites are empty-array *type annotations*, not literals — nothing to convert.

One extra cleanup folded in: `GoalField.test.ts:75` asserted `{ ...challenge, progressGoal: 99 } as IChallenge` — a spread of an already-complete record, so the cast was pure redundancy and is now a plain annotation.

## Defaults, and where suites deliberately diverge

`challengeTypeId` defaults to `EnemiesKilled` (matching the workbench blank record) but `entityType` defaults to `None` rather than the `Enemy` dimension that type implies — `None` is the inert scope, so a suite that doesn't care about scope gets no scope behaviour for free.

Per the issue's warning about load-bearing defaults, every suite that asserts on behaviour those two drive states them **explicitly** instead of inheriting, with a one-line note at the wrapper:

| Suite | Stated rather than inherited |
|---|---|
| `ChallengeConditionSection` | the full condition dimension — it asserts all four fields re-derive on a type change |
| `ScopeField` | type + entity dimension — every case turns on the scope they drive |
| `GoalField` | type — it selects the goal comparison and unit |
| `ChallengeTooltip` | type — it asserts the type label and accent colour |
| `challenge-helpers` | statistic/entity — the suite's subject *is* the type→statistic derivation |
| `Challenges`, `challenges-view` | `challengeTypeId` stays required in the wrapper signature; description generated from the name so cards render real prose |
| `ClassPicker`, `class-summary` | a real passive (the fixture's is inert at 0), which the chip/summary assertions read |

## Test plan

- [x] `npm run lint` (eslint + svelte-check) — clean, 1229 files, 0 errors/warnings
- [x] `npx vitest --run` — **312 files, 4051 tests, all passing**
- [x] Probe re-run to confirm drift detection collapsed to one site per contract (output above), contract file restored afterwards

Backend untouched, so no `dotnet` run was needed.

## Follow-up

Six workbench suites duplicate a near-identical ~14-line `config()` factory ending in `as unknown as EntityConfig<Identified>`. That's the same class of duplication as this issue but a different contract, so it's filed separately rather than widened into this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EckyjM3GF7a2iXYRgyBgGu)_

### DIFF
--- a/UI/src/tests/fixtures/challenges.ts
+++ b/UI/src/tests/fixtures/challenges.ts
@@ -26,6 +26,11 @@ import { EChallengeType, EEntityType, type IChallenge } from '$lib/api';
  * surfaces, so a suite exercising either states **both** explicitly rather than leaning on these
  * defaults — as does one whose subject is the type→statistic derivation.
  *
+ * That pairing is deliberately **off the production manifold**: `deriveFromType` always writes
+ * `statisticType`/`entityType` from the type, so `EnemiesKilled` with `entityType: None` and no
+ * statistic is a record the workbench can never author. Neutral is the point here — don't read a bare
+ * `makeChallenge({ id })` as production-shaped.
+ *
  * The optional fields (`statisticType`, `targetEntityId`, `rewardItemId`, `rewardItemModId`,
  * `retiredAt`) are left unset for the same reason: each flips a distinct branch — the goal unit, the
  * scope target, reward resolution, retirement filtering — when present.
--- a/UI/src/tests/fixtures/challenges.ts
+++ b/UI/src/tests/fixtures/challenges.ts
@@ -1,0 +1,41 @@
+import { EChallengeType, EEntityType, type IChallenge } from '$lib/api';
+
+/* Shared `IChallenge` contract builder (#2456), following the `fixtures/items.ts` convention.
+
+   The target set was confirmed with the throwaway-required-field probe, not a `grep`: only a suite that
+   constructs a *typed* `IChallenge` pays the contract-drift tax, and a probe field flushes out four
+   suites plus the production blank record (`routes/admin/workbench/entities/challenge.ts`). That record
+   is the authoritative shape and correctly stays a literal. Two other classes of challenge literal are
+   excluded on purpose: the untyped `staticData` mocks (the Codex suites seed `{} as any`), which see no
+   new required field, and `entities/challenge.test.ts`, which spreads the production `newItem` rather
+   than hand-rolling a record. A `grep` still finding challenge literals under `src/tests/` is expected,
+   not a gap.
+
+   A third class was invisible to the probe rather than excluded: a literal asserted `as IChallenge`
+   keeps compiling when the contract gains a required field. #2447 handed those nine sites over here and
+   they are now converted; reintroducing such a cast re-opens the hole, so state why if you do. */
+
+/**
+ * Builds an {@link IChallenge} reference-data entry for tests. Everything but the id is a neutral
+ * placeholder so a suite states only what it asserts on.
+ *
+ * `challengeTypeId` defaults to `EnemiesKilled`, matching the workbench's blank record, but
+ * `entityType` deliberately defaults to `None` rather than the `Enemy` dimension that type implies:
+ * `None` is the inert scope (tracked globally, no target select, no target id), so a suite that doesn't
+ * care about scope gets no scope behaviour for free. The two fields drive the condition and scope
+ * surfaces, so a suite exercising either states **both** explicitly rather than leaning on these
+ * defaults — as does one whose subject is the type→statistic derivation.
+ *
+ * The optional fields (`statisticType`, `targetEntityId`, `rewardItemId`, `rewardItemModId`,
+ * `retiredAt`) are left unset for the same reason: each flips a distinct branch — the goal unit, the
+ * scope target, reward resolution, retirement filtering — when present.
+ */
+export const makeChallenge = (overrides: Partial<IChallenge> & { id: number }): IChallenge => ({
+	name: `Challenge ${overrides.id}`,
+	description: '',
+	designerNotes: '',
+	challengeTypeId: EChallengeType.EnemiesKilled,
+	entityType: EEntityType.None,
+	progressGoal: 10,
+	...overrides
+});
--- a/UI/src/tests/routes/game/screens/challenges/challenges-view.svelte.test.ts
+++ b/UI/src/tests/routes/game/screens/challenges/challenges-view.svelte.test.ts
@@ -56,6 +56,7 @@ import {
 } from '$routes/game/screens/challenges/challenges-view.svelte';
 import { makeItem } from '../../../../fixtures/items';
 import { makeItemMod } from '../../../../fixtures/item-mods';
+import { makeChallenge } from '../../../../fixtures/challenges';
 
 /** Carries attributes and a mod slot by default — the reward tooltip renders both. */
 const item = (id: number, name: string, rarity: ERarity, cat: EItemCategory, extra: Partial<IItem> = {}): IItem =>
@@ -81,13 +82,11 @@ const mod = (id: number, name: string, rarity: ERarity): IItemMod =>
 		attributes: [{ attributeId: EAttribute.Agility, amount: 3 }]
 	});
 
-const challenge = (over: Partial<IChallenge> & Pick<IChallenge, 'id' | 'name' | 'challengeTypeId'>): IChallenge => ({
-	description: `${over.name} description`,
-	designerNotes: '',
-	entityType: EEntityType.None,
-	progressGoal: 10,
-	...over
-});
+/* `challengeTypeId` stays required rather than inheriting the fixture default — the view-model's goal
+   comparison and target resolution are driven by it — and the description is generated from the name so
+   each card carries real prose instead of the fixture's empty placeholder. */
+const challenge = (over: Partial<IChallenge> & Pick<IChallenge, 'id' | 'name' | 'challengeTypeId'>): IChallenge =>
+	makeChallenge({ description: `${over.name} description`, ...over });
 
 beforeEach(() => {
 	staticData.items = [];
--- a/UI/src/tests/routes/select/ClassPicker.test.ts
+++ b/UI/src/tests/routes/select/ClassPicker.test.ts
@@ -8,15 +8,14 @@ import { staticData } from '$stores/static-data.svelte';
 import { EEquipmentSlot, type ICreatableClass } from '$lib/api';
 import { makeCreatableClass } from '../../fixtures/classes';
 
-/* The passive diverges from the fixture's inert default: the picker renders its attribute chip, so the
-   class carries a real passive (Endurance +8) for those assertions to read. */
+/* `passiveAmount` diverges from the fixture's inert 0: the picker renders the passive as an attribute
+   chip, so the class needs a real amount for those assertions to read. The attribute itself is the
+   fixture's default (Endurance), which is what the mocked `staticData.attributes` resolves to. */
 const cls = (overrides: Partial<ICreatableClass> = {}): ICreatableClass =>
 	makeCreatableClass({
 		id: 0,
 		name: 'Warrior',
 		description: 'A frontline fighter.',
-		word: 'kor',
-		passiveAttributeId: 1,
 		passiveAmount: 8,
 		...overrides
 	});
@@ -59,9 +58,6 @@ describe('ClassPicker', () => {
 		const classes = [
 			cls({
 				id: 0,
-				description: 'A frontline fighter.',
-				passiveAttributeId: 1,
-				passiveAmount: 8,
 				starterSkills: [
 					{ id: 1, name: 'Slash' },
 					{ id: 0, name: 'Punch' }
--- a/UI/src/tests/routes/select/ClassPicker.test.ts
+++ b/UI/src/tests/routes/select/ClassPicker.test.ts
@@ -5,24 +5,21 @@ import { render, fireEvent, cleanup, screen } from '@testing-library/svelte';
 // exercised: with no attribute-tooltip context here, the chips must stay out of the tab order.
 import ClassPicker from '$routes/select/ClassPicker.svelte';
 import { staticData } from '$stores/static-data.svelte';
-import { EEquipmentSlot, EModifierType, type ICreatableClass } from '$lib/api';
+import { EEquipmentSlot, type ICreatableClass } from '$lib/api';
+import { makeCreatableClass } from '../../fixtures/classes';
 
-/* Deliberately partial: `ICreatableClass` has no shared fixture yet (#2456). */
+/* The passive diverges from the fixture's inert default: the picker renders its attribute chip, so the
+   class carries a real passive (Endurance +8) for those assertions to read. */
 const cls = (overrides: Partial<ICreatableClass> = {}): ICreatableClass =>
-	({
+	makeCreatableClass({
 		id: 0,
 		name: 'Warrior',
 		description: 'A frontline fighter.',
 		word: 'kor',
 		passiveAttributeId: 1,
 		passiveAmount: 8,
-		passiveScalingAmount: 0,
-		passiveModifierType: EModifierType.Additive,
-		attributeDistributions: [],
-		starterSkills: [],
-		starterEquipment: [],
 		...overrides
-	}) as ICreatableClass;
+	});
 
 beforeEach(() => {
 	staticData.attributes = [{ id: 1, code: 'END', name: 'Endurance' }] as unknown as NonNullable<
--- a/UI/src/tests/routes/select/class-summary.test.ts
+++ b/UI/src/tests/routes/select/class-summary.test.ts
@@ -8,23 +8,12 @@ import {
 } from '$lib/api';
 import { passiveSummary, weaponFirst } from '$routes/select/class-summary';
 import { makeAttribute } from '../../fixtures/attributes';
+import { makeCreatableClass } from '../../fixtures/classes';
 
-/* Deliberately partial: `ICreatableClass` has no shared fixture yet (#2456). */
+/* `passiveAmount` diverges from the fixture's inert 0: every case here reads the summary phrasing built
+   from a real passive. */
 const cls = (overrides: Partial<ICreatableClass> = {}): ICreatableClass =>
-	({
-		id: 0,
-		name: 'Warrior',
-		description: '',
-		word: 'kor',
-		passiveAttributeId: EAttribute.Endurance,
-		passiveAmount: 8,
-		passiveScalingAmount: 0,
-		passiveModifierType: EModifierType.Additive,
-		attributeDistributions: [],
-		starterSkills: [],
-		starterEquipment: [],
-		...overrides
-	}) as ICreatableClass;
+	makeCreatableClass({ id: 0, name: 'Warrior', word: 'kor', passiveAmount: 8, ...overrides });
 
 const attributes = [
 	makeAttribute(EAttribute.Endurance, 'Endurance', { code: 'END' }),
--- a/UI/src/tests/routes/select/class-summary.test.ts
+++ b/UI/src/tests/routes/select/class-summary.test.ts
@@ -13,7 +13,7 @@ import { makeCreatableClass } from '../../fixtures/classes';
 /* `passiveAmount` diverges from the fixture's inert 0: every case here reads the summary phrasing built
    from a real passive. */
 const cls = (overrides: Partial<ICreatableClass> = {}): ICreatableClass =>
-	makeCreatableClass({ id: 0, name: 'Warrior', word: 'kor', passiveAmount: 8, ...overrides });
+	makeCreatableClass({ id: 0, name: 'Warrior', passiveAmount: 8, ...overrides });
 
 const attributes = [
 	makeAttribute(EAttribute.Endurance, 'Endurance', { code: 'END' }),
--- a/UI/src/tests/components/tooltip/ChallengeTooltip.test.ts
+++ b/UI/src/tests/components/tooltip/ChallengeTooltip.test.ts
@@ -21,21 +21,21 @@ vi.mock('$stores', () => ({ staticData, playerChallenges }));
 import ChallengeTooltip from '$components/tooltip/ChallengeTooltip.svelte';
 import { makeItem } from '../../fixtures/items';
 import { makeZone } from '../../fixtures/zones';
+import { makeChallenge } from '../../fixtures/challenges';
 
 const zone = (id: number, order: number, name: string, unlockChallengeId?: number): IZone =>
 	makeZone({ id, name, order, unlockChallengeId });
 
 const setup = (over: Partial<IChallenge> = {}) => {
 	staticData.challenges = [];
-	// Deliberately partial: `IChallenge` has no shared fixture yet (#2456); the tooltip reads only these fields.
-	staticData.challenges[5] = {
+	// The type is stated rather than inherited: the suite asserts on its label and accent colour.
+	staticData.challenges[5] = makeChallenge({
 		id: 5,
 		name: 'Cull the Swarm',
 		description: 'Defeat 10 enemies',
 		challengeTypeId: EChallengeType.EnemiesKilled,
-		progressGoal: 10,
 		...over
-	} as IChallenge;
+	});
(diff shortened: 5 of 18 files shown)


UI/src/tests/routes/select/ClassPicker.test.ts
```diff
@@ -5,24 +5,21 @@ import { render, fireEvent, cleanup, screen } from '@testing-library/svelte';
...
+/* The passive diverges from the fixture's inert default: the picker renders its attribute chip, so the
+   class carries a real passive (Endurance +8) for those assertions to read. */
 const cls = (overrides: Partial<ICreatableClass> = {}): ICreatableClass =>
-	({
+	makeCreatableClass({
 		id: 0,
 		name: 'Warrior',
 		description: 'A frontline fighter.',
 		word: 'kor',
 		passiveAttributeId: 1,
```
ginderjeremiah: **Should-fix** — `passiveAttributeId: 1` isn't a divergence, it's the fixture default written as a magic number. `EAttribute.Endurance === 1`, so this restates `fixtures/classes.ts`'s own default while dropping the enum that names it.

`docs/frontend.md:25` is explicit: *"where a suite needs different defaults, wrap the shared builder and state only the divergence."* The conversion correctly dropped `passiveScalingAmount`, `passiveModifierType`, and the three empty arrays as now-default — `passiveAttributeId` belongs in that group too. `passiveAmount: 8` and `description` genuinely diverge (the chip asserts `'+8 END'`, and `getByText('A frontline fighter.')` reads the description), so those stay.

Worth noting the sibling conversion got this right: `class-summary.test.ts:16` dropped its `passiveAttributeId: EAttribute.Endurance` for exactly this reason. Two conversions of the same contract in the same folder shouldn't disagree on whether the default is worth restating.

There's a second copy at line 63 (outside this hunk) inside the kit-preview case, which also re-states `description` and `passiveAmount` already supplied by the wrapper — same fix applies while you're here.

Also: the wrapper comment above says the class carries "Endurance +8", naming the enum the code doesn't use. Dropping the line resolves that too.


---

UI/src/tests/routes/select/class-summary.test.ts
```diff
@@ -8,23 +8,12 @@ import {
...
-		passiveAttributeId: EAttribute.Endurance,
-		passiveAmount: 8,
-		passiveScalingAmount: 0,
-		passiveModifierType: EModifierType.Additive,
-		attributeDistributions: [],
-		starterSkills: [],
-		starterEquipment: [],
-		...overrides
-	}) as ICreatableClass;
+	makeCreatableClass({ id: 0, name: 'Warrior', word: 'kor', passiveAmount: 8, ...overrides });
```
ginderjeremiah: **Nit** — `word: 'kor'` is the one leftover that isn't read. This suite only exercises `passiveSummary` and `weaponFirst`; nothing in it touches `word`, and the fixture's `w0` would serve equally. Same `docs/frontend.md:25` rule as the `ClassPicker` comment — state only the divergence.

The neighbouring `passiveAmount: 8` is right to keep, and the comment correctly calls it out as the divergence.


---

UI/src/tests/fixtures/challenges.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Builds an {@link IChallenge} reference-data entry for tests. Everything but the id is a neutral
+ * placeholder so a suite states only what it asserts on.
+ *
+ * `challengeTypeId` defaults to `EnemiesKilled`, matching the workbench's blank record, but
+ * `entityType` deliberately defaults to `None` rather than the `Enemy` dimension that type implies:
+ * `None` is the inert scope (tracked globally, no target select, no target id), so a suite that doesn't
+ * care about scope gets no scope behaviour for free. The two fields drive the condition and scope
+ * surfaces, so a suite exercising either states **both** explicitly rather than leaning on these
+ * defaults — as does one whose subject is the type→statistic derivation.
```
ginderjeremiah, line 27: **Nit** — the reasoning here is right, and I checked the alternatives before agreeing: every challenge type whose statistic has `entityType: None` is load-bearing in its own way. `TimeTrial` carries `AtMost` (which `GoalField` turns on) and `LevelReached` flips `progressSource`/`goalUnit`/the statistic label via `LEVEL_REACHED` in `challenge-helpers.ts:53,59,93`. So `EnemiesKilled` + `None` really is the least-load-bearing pair available, not a compromise.

What the docstring stops one sentence short of saying is the consequence: `deriveFromType` (`challenge-helpers.ts:163`) *always* writes `statisticType`/`entityType` from the type, so `EnemiesKilled` with `entityType: None` and no `statisticType` is a combination the workbench can never author. The default is deliberately off the production manifold. That's fine for a neutral fixture, but it's the sort of thing a future conversion will assume the opposite of — worth a clause so nobody reads `makeChallenge({ id: 1 })` output as production-shaped.


---

UI/src/tests/fixtures/challenges.ts
```diff
@@ -0,0 +1,41 @@
...
+   suites plus the production blank record (`routes/admin/workbench/entities/challenge.ts`). That record
+   is the authoritative shape and correctly stays a literal. Two other classes of challenge literal are
+   excluded on purpose: the untyped `staticData` mocks (the Codex suites seed `{} as any`), which see no
+   new required field, and `entities/challenge.test.ts`, which spreads the production `newItem` rather
+   than hand-rolling a record. A `grep` still finding challenge literals under `src/tests/` is expected,
+   not a gap.
+
+   A third class was invisible to the probe rather than excluded: a literal asserted `as IChallenge`
+   keeps compiling when the contract gains a required field. #2447 handed those nine sites over here and
+   they are now converted; reintroducing such a cast re-opens the hole, so state why if you do. */
```
ginderjeremiah, line 16: **Follow-up, not a change request here.** The per-contract content of this block is genuinely per-module and worth keeping — which suites are excluded and why is real information a reader needs.

But the *general* rule wrapped around it isn't contract-specific, and this is now its fifth copy: "only a suite constructing a *typed* `X` pays the contract-drift tax", "untyped `staticData` mocks see no new required field", "a `grep` still finding literals is expected, not a gap", and the closing "reintroducing such a cast re-opens the hole, so state why if you do" all appear in near-identical words in `items.ts`, `skills.ts`, and `attributes.ts`. A DRY pass on builders that leaves the prose duplicated five ways is a little ironic.

Extracting it belongs in its own pass since it edits four modules this PR doesn't touch — filed as a follow-up. Each module would keep only its own exclusion table.


---

UI/src/tests/routes/game/screens/challenges/challenges-view.svelte.test.ts
```diff
@@ -81,13 +82,11 @@ const mod = (id: number, name: string, rarity: ERarity): IItemMod =>
...
-	designerNotes: '',
-	entityType: EEntityType.None,
-	progressGoal: 10,
-	...over
-});
+/* `challengeTypeId` stays required rather than inheriting the fixture default — the view-model's goal
+   comparison and target resolution are driven by it — and the description is generated from the name so
+   each card carries real prose instead of the fixture's empty placeholder. */
+const challenge = (over: Partial<IChallenge> & Pick<IChallenge, 'id' | 'name' | 'challengeTypeId'>): IChallenge =>
+	makeChallenge({ description: `${over.name} description`, ...over });
```
ginderjeremiah, line 89: **Nit** — this wrapper and `Challenges.test.ts:32` are now byte-identical, signature included, in the same folder:

```ts
const challenge = (over: Partial<IChallenge> & Pick<IChallenge, 'id' | 'name' | 'challengeTypeId'>): IChallenge =>
	makeChallenge({ description: `${over.name} description`, ...over });
```

Before the conversion they were two five-field literals that merely looked alike; now they're the same one-liner (plus a three-line comment each saying much the same thing). That's small, but it's the exact duplication class this issue exists to remove, and the folder already has the precedent for the fix — `fight/fight-fixtures.ts` and `progression-test-utils` are both folder-local shared test helpers.

Either collapse both onto one shared wrapper here, or note why each suite keeps its own. Fine to fold in or defer — flagging it because "we deduped the builders and left two identical wrappers" is easy to lose track of.


---

ginderjeremiah (commented):
## Review: #2456 — consolidate the `IChallenge` / `ICreatableClass` test builders onto shared fixtures

**Verdict: Approve** — the conversion is faithful, the scope adjustments are correct, and nothing here is blocking. Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR author; treat it as an approve with one should-fix and a few nits.

Not merging yet: the `ClassPicker` should-fix is a one-line change I'd rather see land in this PR than as a trailing issue, since it's the same rule the rest of the diff applies correctly.

### Verification

I checked the two things most likely to be quietly wrong in a conversion like this, and both hold:

- **Every converted site is value-identical or behaviour-neutral.** The enum values line up exactly where a suite previously wrote numerics — `EChallengeType.EnemiesKilled === 1` and `EEntityType.None === 0`, so `ChallengeRewardSection`'s old `challengeTypeId: 1, entityType: 0` and the `Challenges` / `challenges-view` / `references` wrappers' explicit `None` / `10` / `''` defaults all survive the switch unchanged. The five sites whose casts previously left fields `undefined` (`ChallengeTooltip`, `ZoneNav`, `RewardPicker`, `offline-summary`, `challenge-unlocks`) now get them populated, but none of their consumers read `entityType`, `statisticType`, `targetEntityId`, `designerNotes`, or — for `offline-summary` / `RewardPicker` — `progressGoal` or `description` at all. No assertion silently loosened.
- **No dead imports left behind.** `EChallengeType` came out of `references.test.ts` and `EEntityType` out of `Challenges.test.ts` exactly where they went unused; every other enum still in an import list has a live reference (`EModifierType` in `class-summary.test.ts:51`, `EEntityType` in `challenges-view`, etc.).

Both new fixtures cover every required field on their contract, and the `Partial<T> & { id: number }` signature matches the `items.ts` / `skills.ts` convention — so a future required field is one edit per module, which is the point.

### Should-fix

- **`ClassPicker.test.ts:19` restates the fixture default as a magic number** — `passiveAttributeId: 1` *is* `EAttribute.Endurance`, i.e. `fixtures/classes.ts`'s own default. Against `docs/frontend.md:25` ("state only the divergence"), and the sibling `class-summary.test.ts` conversion correctly dropped the same line. Second copy at line 63.

### Nits

- `class-summary.test.ts:16` — `word: 'kor'` isn't read by either function under test; same rule.
- `fixtures/challenges.ts:22-27` — the `EnemiesKilled` + `None` default is the right call (I checked the alternatives; `TimeTrial` and `LevelReached` are both load-bearing), but the docstring should say the pair is one `deriveFromType` can never author, so nobody reads the output as production-shaped.
- `challenges-view.svelte.test.ts:89` and `Challenges.test.ts:32` are now byte-identical wrappers in the same folder.

### Follow-up

- The contract-independent half of the fixture header comments is duplicated near-verbatim across five modules now. Filed separately since extracting it edits four files this PR doesn't touch.

### Test coverage

Test-only refactor with no production code and no behaviour change, so the project's coverage rules are satisfied by construction — there's no new logic to cover, and the fixtures are test infrastructure rather than a testable unit. No battle-logic involvement, so the frontend/backend mirroring rule doesn't apply. CI is green across all twelve checks, including `test-frontend`, `verify-frontend`, and all three e2e browsers.

The real coverage question for a change like this is whether drift detection actually improved, and the PR answers it with evidence rather than assertion. The `ICreatableClass` half is the stronger result: all three sites asserted their literal, so the contract had **zero** test-side drift detection before this — the probe found nothing to report, which is exactly the failure mode the cast form produces.

### What's good

Using the throwaway-required-field probe instead of the issue's `grep` is the right instrument, and it changed the answer: it separated literals that genuinely pay the drift tax from ones that only look like they do, and it caught that nine cast sites were silently compiling. Reporting the after-state as "exactly one test-file error per contract" is a falsifiable claim, which is much better than "converted 15 suites".

Pushing back on the issue's suite list is also correct on both counts. `entities/challenge.test.ts` spreading the production `newItem` is the authoritative shape — converting it onto a fixture would have made the test agree with the fixture instead of with production, which is backwards, and it matches the `fixtures/item-mods.ts` precedent. The Codex pair's `{} as any` genuinely sees no new required field. And the `challenges: [] as IChallenge[]` sites being empty-array annotations rather than literals is the kind of detail a mechanical sweep gets wrong.

The load-bearing-defaults discipline is the best part. Each suite that asserts on behaviour the defaults drive states them explicitly with a one-line note, and the notes are specific about *what* the field drives rather than just "needed here" — `GoalField`'s "selects the goal comparison and unit", `challenge-helpers`' "the suite's subject *is* the type→statistic derivation". That's what stops the next person from "simplifying" the override away. `challenge-helpers`' `make()` keeping `?? EEntityType.None` is a good instance: without it the spread would write `entityType: undefined` and silently produce a record missing a required field, since `exactOptionalPropertyTypes` is off.

Folding in `GoalField.test.ts:75` — a spread of an already-complete record, so `as IChallenge` was pure redundancy — is the right kind of opportunistic cleanup: adjacent, zero-risk, and it removes a cast rather than moving one.


---

ginderjeremiah:
All four points addressed in `ec09793`. Every one of them was correct as written — I verified each before acting rather than taking the review at face value.

**Should-fix — `ClassPicker` `passiveAttributeId: 1`.** Confirmed: `EAttribute.Endurance === 1` (`enums.ts:45`), so this restated the fixture's own default while dropping the enum that names it. Removed from the wrapper and from the kit-preview case at the old line 63, which also restated `description` and `passiveAmount` the wrapper already supplies. The wrapper comment no longer names an enum the code doesn't use — it now says the attribute *is* the fixture default, which is what `staticData.attributes` resolves to.

**Nit — `class-summary` `word: 'kor'`.** Dropped. Confirmed nothing in either select suite reads `word`; the assertions are all on `passiveSummary` output and equipment names.

**Nit — the default pair being off the production manifold.** Good catch, and the more useful half of the docstring. Added the consequence explicitly: `deriveFromType` always writes `statisticType`/`entityType` from the type, so `EnemiesKilled` + `entityType: None` + no statistic is a record the workbench can never author — neutral is the point, don't read a bare `makeChallenge({ id })` as production-shaped.

**Nit — the byte-identical wrappers.** Fixed, but not by sharing them. `Challenges.test.ts` had exactly **one** call site for its wrapper, so a folder-local shared module would have been ceremony for a single consumer on one side — the same reasoning `player-select-view`'s `IPlayerData` builder carries. Inlined the single `makeChallenge(...)` call there instead and deleted the wrapper (its now-unused `IChallenge` import went with it). `challenges-view` keeps its wrapper, which has five call sites and genuinely earns one. Net effect is the same: no duplicated wrapper, and one fewer indirection.

**Follow-up — the duplicated header prose.** Already tracked as #2482; agreed it edits four modules this PR doesn't touch. I'd also filed #2481 for the six workbench suites duplicating the `EntityConfig<Identified>` `config()` factory — same duplication class, different contract, and it has the same cast-silences-drift property this PR just removed for `IChallenge`.

**Verification after the changes:** `npm run lint` clean (1229 files, 0 errors/warnings), full `npx vitest --run` green at **312 files / 4051 tests**, and the affected `select/` + `challenges/` suites run individually (17 files / 133 tests). No behaviour change — every edit removed a restatement of a value the fixture already supplies.

---

ginderjeremiah:
## Re-review of `ec09793` — approve, merging

Incremental only; the previous review covered `73467d0`. Four files, +19/−16, all of it removing restatements. Every point is addressed and each one checks out against the code:

**`ClassPicker` (should-fix) — correct.** `fixtures/classes.ts:24` defaults `passiveAttributeId: EAttribute.Endurance`, and the suite's `staticData.attributes` is seeded with the single entry `{ id: 1, code: 'END' }`, so `'+8 END'` still resolves through the fixture default. The second copy at the old line 63 is gone along with the `description`/`passiveAmount` it restated. The reworded comment is accurate — it now claims only `passiveAmount` as the divergence, which is the one field that actually diverges from the inert default.

**`word: 'kor'` (nit) — correct, and worth noting it went further than the nit asked.** The nit was scoped to `class-summary.test.ts`, where neither function under test reads `word`. Dropping it from `ClassPicker` too is a case the nit didn't cover, since `ClassPicker.svelte:26` *does* render it (`<WordOfPower text={cls.word} …>`). It's still right: `WordOfPower` is a pure passthrough — the text lands in DOM text and the `title` attribute with no content-dependent behaviour — and no assertion in the suite reads either. So `w0` vs. `kor` is behaviourally identical and the default is the correct call. Flagging it only because "unread" and "unasserted" are different justifications and this site is the second one.

**Docstring (nit) — this is the better version.** Stating the *consequence* — `deriveFromType` always writes `statisticType`/`entityType` from the type, so the default pair is a record the workbench can never author — is more useful than the observation alone. "Don't read a bare `makeChallenge({ id })` as production-shaped" is the sentence that stops someone lifting the fixture output into a production-shaped expectation later.

**Duplicate wrappers (nit) — resolved better than by sharing.** Inlining is the right call for a single call site, and the `player-select-view` precedent is the applicable one. Checked the leftovers: the `IChallenge` type import is gone from `Challenges.test.ts` and nothing references the deleted `challenge()` helper. The explanatory comment survived the inlining rather than being dropped with the wrapper, which is the part that's easy to lose — it moved to the `beforeEach` and still names what the type drives (badge, objective sentence). `challenges-view` keeps its wrapper at five call sites.

### Verdict

**Approve**, no remaining should-fixes or nits. CI green across all twelve checks on `ec09793` (`test-frontend`, `verify-frontend`, all three e2e browsers). Test-only, no production code, no behaviour change — every edit in this commit deletes a value the fixture already supplies.

Follow-ups are filed and neither belongs in this PR: #2481 (the six workbench `EntityConfig` `config()` factories — different contract, and the `as unknown as` double cast needs different treatment than a plain builder) and #2482 (the general contract-drift prose now on its fifth near-verbatim copy — extracting it edits four modules this PR doesn't touch).

Merging.

---

Issue #2470, opened 2026-07-26
Consolidate the duplicated ISessionStore test doubles onto the shared FakeSessionStore


ginderjeremiah:
Noticed while reviewing #2469, which promoted `FakeSessionStore` out of `SessionServiceTests` into `Game.TestInfrastructure/Helpers` so a second test class could share it.

Now that a shared, canonical `ISessionStore` double exists, the private per-file copies are redundant duplication of the same interface implementation:

- `Game.Api.Tests/Unit/RequestLoggingMiddlewareTests.cs:53` — `NoOpSessionStore`
- `Game.Api.Tests/Unit/ExceptionHandlingMiddlewareTests.cs:218` — `NoOpSessionStore` (fully-qualified type names rather than usings, too)
- `Game.Api.Tests/Unit/ReferenceDataCommandAdminRedactionTests.cs:153` — `NoOpSessionStore`
- `Game.Api.Tests/Unit/SocketCommandProcessorTests.cs:782` — `NoOpSessionStore`

`FakeSessionStore` with no `Session` assigned behaves identically to every one of them (`GetSession` returns `null`, writes are recorded and otherwise inert), so each copy can be replaced with the shared helper and deleted. Two of the four also drifted — their `Update`/`UpdateAsync` parameters are named `playerId` where the interface's is `userId`, which is actively misleading about what the session store is keyed by.

Interface-shaped duplication like this is exactly what bites when `ISessionStore` next gains a member: five files to edit instead of one.

Scope is mechanical — swap the type, add the `Game.TestInfrastructure.Helpers` using, delete the nested classes, confirm the suite still passes.

PR #2483, opened 2026-07-26
Take the engines' per-tick internals out of the statify proxy (#2462)

### LINKED ISSUE #2462: LogicalEngine/RenderEngine per-tick internals are statify-reactive, contrary to the #2123 convention

The #2123 convention (documented in `docs/frontend.md` and pinned by a statify regression test) is that non-reactive hot-path fields on a statify'd class use `#`-private so per-tick internals stay out of the deep proxy. `EnemyManager`/`PlayerManager`/`BattleEngine` follow it, but the two engines predate it:

- `UI/src/lib/engine/logical-engine.ts:21-27` — `time`, `lastTime`, `timeBank` are plain properties; `lastTime`/`timeBank` are written through `$state` setters on every ~10ms worker poll and `time` every 40ms tick.
- `UI/src/lib/engine/render-engine.ts:9-15` — `time`, `logicalDelta`, `rafHandle`, `running` are written every rAF frame.
- Both instances are statified at `UI/src/lib/engine/engine.ts:62-63`, and `statify` converts every enumerable field to a `$state` accessor (`UI/src/lib/common/statify.svelte.ts:31-47`).

Grep confirms none of these fields has a reactive consumer (only `tickRate` is read reactively, `NavSidebar.svelte:87,93`), so this is signal-write overhead with no subscribers — cheap in absolute terms, so this is a convention/consistency fix rather than a measurable frame cost. Convert the internals to `#`-private (keeping `tickRate` reactive) so the engines match the documented convention.

In passing (fold into this or any nearby change, not worth its own issue): `EnemyManager.resolveBossVictory` double-sorts the zone list — `nextZoneByOrder(navigableZones(...))` (`UI/src/lib/engine/battle/enemy-manager.ts:712`) where `navigableZones` already sorts (`zone-progression.ts:21`) and `nextZoneByOrder` re-sorts its input (`zone-progression.ts:33-37`). Harmless and once-per-boss-kill.

ginderjeremiah:
Closes #2462.

Frontend only — no backend, battle-logic, or content changes.

## What's in it

`LogicalEngine` and `RenderEngine` predate the #2123 convention (documented in `docs/frontend.md`) that non-reactive hot-path fields on a statify'd class use `#`-private, so their loop bookkeeping was a set of `$state` accessors with no subscriber.

- **`logical-engine.ts`** — `time`, `lastTime`, `timeBank`, plus the tick counter and tick-source handle are now `#`-private. `tickRate` stays public and reactive.
- **`render-engine.ts`** — same for `time`, `logicalDelta`, `running`, `rafHandle` and the frame counter.
- **`time` (and `logicalDelta`) are exposed as read-only getters** rather than dropped: the render loop reads `logicEngine.time` across the class boundary, and the existing suites assert on both. A class accessor lives on the prototype and is non-enumerable, so statify's `for...in` walk never sees it — the same `#`-field-behind-a-getter shape the statify regression test already pins.
- `RenderEngine.update` became a normal private method while it was being touched (it was an arrow-function field, invoked only as `this.update()`), matching `LogicalEngine`.

**Precisely what the proxy loses:** the seven number/boolean/object fields above. The two function-valued fields (`countTick`, and `update` on the render side) were never proxied — `statify` skips functions outright (`statify.svelte.ts:34-36`) — so converting them is a consistency call about where loop internals live, not a removed signal write.

Nothing outside the classes ever wrote any of these, so the getters cost no call site.

## The bonus fix, done slightly differently than the issue suggested

The issue flagged `resolveBossVictory`'s double sort — `nextZoneByOrder(navigableZones(...))`, where `nextZoneByOrder` re-sorts its already-sorted input. Deduping purely at the call site isn't possible (the retired filter is still needed), so the helper now does the filtering itself and is renamed **`nextNavigableZone`** to say so.

That's a small semantic change beyond the sort, and it's the reason I went this way: the old signature would happily return a **retired** zone if any caller passed an unfiltered list, and "the next zone to unlock" should never be one. It's the only caller, so the fix is contained; the ordering pass now happens once.

## Test plan

- **New `src/tests/lib/engine/engine-statify.svelte.test.ts`** — pins the split for both engines: `tickRate` has a statify-installed accessor, `time`/`logicalDelta` have no own descriptor, and behaviourally the clocks advance (real ticks / real frames) while a `$derived` reading them never re-runs, whereas a `tickRate` write does notify. Worth having as its own suite because a `#`-field silently promoted back to public would pass every other engine test unchanged.
- **`zone-progression.test.ts`** — renamed suite, plus two new cases for the folded-in semantics: a retired middle zone is skipped rather than returned, and a retired *current* zone yields `undefined`.
- `npm run lint` — clean (eslint + svelte-check: 1230 files, 0 errors, 0 warnings).
- `npm run test:unit:ci` — **313 files / 4059 tests passing**, coverage thresholds met.
- Backend build/tests not run: no C# files are touched.

## Review follow-ups

- The new suite's rAF stub now uses a monotonic handle counter instead of `rafCallbacks.length`, which repeated a handle once a test shifted a callback off the front.
- `RenderEngine`'s two getters carry a comment saying they are observability only — nothing in production reads them, unlike `LogicalEngine.time`, whose cross-class reader is `RenderEngine.update`. Kept rather than made `#`-only: the absolute-clock assertions can't be reconstructed from the `onRenderUpdate` payload (it carries deltas), and a prototype accessor costs nothing reactively.
- The third copy of the engine `window` stub is filed as #2489 rather than folded in — consolidating it edits two suites this PR doesn't otherwise touch.

## Note

This is a convention/consistency fix, as the issue says — the removed signal writes were cheap in absolute terms, so I'd not expect a measurable frame-time difference, and none is claimed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01R5zBMDeZe3EBP8yiL6tWpL)_

### DIFF
--- a/UI/src/lib/engine/render-engine.ts
+++ b/UI/src/lib/engine/render-engine.ts
@@ -6,53 +6,66 @@ const notifyRenderUpdate = renderUpdateHook.notify;
 export const onRenderUpdate = renderUpdateHook.onNotified;
 
 export class RenderEngine {
-	public time = 0;
-	public logicalDelta = 0;
+	/** The measured frame rate, the one field with a reactive consumer (the nav sidebar's readout). */
 	public tickRate = 0;
 
-	private running = false;
-	private rafHandle?: number;
-	private countTick = getEventCounter((t) => (this.tickRate = Math.round(t)));
+	// Per-frame internals are `#`-private so statify never proxies them (#2123): all of these are written
+	// on every animation frame, and nothing reads them reactively.
+	#time = 0;
+	#logicalDelta = 0;
+	#running = false;
+	#rafHandle?: number;
+	#countTick = getEventCounter((t) => (this.tickRate = Math.round(t)));
+
+	/** The render clock in `performance.now()` terms. Deliberately non-reactive — poll it, don't `$derived` it. */
+	public get time() {
+		return this.#time;
+	}
+
+	/** How far the render clock leads the logical clock (floored at 0). Non-reactive, like `time`. */
+	public get logicalDelta() {
+		return this.#logicalDelta;
+	}
 
 	public start() {
-		if (!this.running) {
+		if (!this.#running) {
 			// Re-seed the clock so the first frame's delta is ~one frame, not the entire wall-clock gap
 			// the engine was stopped (mirrors LogicalEngine.start, which resets its clock to avoid this).
-			this.time = performance.now();
-			this.running = true;
+			this.#time = performance.now();
+			this.#running = true;
 			this.renderLoop();
 		}
 	}
 
 	public stop() {
-		this.running = false;
+		this.#running = false;
 		// Cancel the pending frame so a start() within the same frame can't leave the old callback
 		// running alongside the new loop (mirrors LogicalEngine.stop clearing its tickSource handle).
-		if (this.rafHandle !== undefined) {
-			window.cancelAnimationFrame(this.rafHandle);
-			this.rafHandle = undefined;
+		if (this.#rafHandle !== undefined) {
+			window.cancelAnimationFrame(this.#rafHandle);
+			this.#rafHandle = undefined;
 		}
 	}
 
 	//use performance.now instead of animation frame timestamp, because frame stamp is before some amount of processing.
 	//using frame timestamp can cause render loop to appear behind logical loop
 	private renderLoop() {
-		if (this.running) {
+		if (this.#running) {
 			this.update();
-			this.rafHandle = window.requestAnimationFrame(() => this.renderLoop());
+			this.#rafHandle = window.requestAnimationFrame(() => this.renderLoop());
 		}
 	}
 
-	private update = () => {
-		this.countTick();
+	private update() {
+		this.#countTick();
 		const newTime = performance.now();
-		const delta = newTime - this.time;
-		this.time = newTime;
+		const delta = newTime - this.#time;
+		this.#time = newTime;
 		// Floor at 0: the logical engine's tab-background catch-up branch advances logicEngine.time by the
 		// discarded excess, which can momentarily push it past the render clock. A negative logicalDelta
 		// would drive the render-only charge/effect interpolation backwards for a frame (purely cosmetic —
 		// logical state and battle parity are unaffected).
-		this.logicalDelta = Math.max(0, newTime - logicEngine.time);
-		notifyRenderUpdate(delta, this.logicalDelta);
-	};
+		this.#logicalDelta = Math.max(0, newTime - logicEngine.time);
+		notifyRenderUpdate(delta, this.#logicalDelta);
+	}
 }
--- a/UI/src/lib/engine/render-engine.ts
+++ b/UI/src/lib/engine/render-engine.ts
@@ -17,12 +17,17 @@ export class RenderEngine {
 	#rafHandle?: number;
 	#countTick = getEventCounter((t) => (this.tickRate = Math.round(t)));
 
-	/** The render clock in `performance.now()` terms. Deliberately non-reactive — poll it, don't `$derived` it. */
+	// Both getters are observability only — unlike `LogicalEngine.time` (which `update` below reads across
+	// the class boundary), nothing in production reads either: consumers take the frame delta and the
+	// interpolation lead from the `onRenderUpdate` payload. Kept for debugging and the clock assertions,
+	// and non-reactive by construction, since a prototype accessor stays out of the statify proxy.
+
+	/** The render clock in `performance.now()` terms. */
 	public get time() {
 		return this.#time;
 	}
 
-	/** How far the render clock leads the logical clock (floored at 0). Non-reactive, like `time`. */
+	/** How far the render clock leads the logical clock (floored at 0). */
 	public get logicalDelta() {
 		return this.#logicalDelta;
 	}
--- a/UI/src/tests/lib/engine/engine-statify.svelte.test.ts
+++ b/UI/src/tests/lib/engine/engine-statify.svelte.test.ts
@@ -1,0 +1,164 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { flushSync } from 'svelte';
+import { statify } from '$lib/common';
+
+// RenderEngine reads the shared logical clock; the real module pulls in the whole engine graph.
+const logicEngineStub = vi.hoisted(() => ({ time: 0 }));
+vi.mock('$lib/engine/engine', () => ({ logicEngine: logicEngineStub }));
+
+const rafCallbacks: (() => void)[] = [];
+vi.stubGlobal('window', {
+	requestAnimationFrame: vi.fn((cb: () => void) => {
+		rafCallbacks.push(cb);
+		return rafCallbacks.length;
+	}),
+	cancelAnimationFrame: vi.fn(),
+	setInterval: (...args: Parameters<typeof setInterval>) => globalThis.setInterval(...args),
+	clearInterval: (...args: Parameters<typeof clearInterval>) => globalThis.clearInterval(...args)
+});
+
+import { LogicalEngine, tickSize } from '$lib/engine/logical-engine';
+import { RenderEngine } from '$lib/engine/render-engine';
+
+/** Whether `statify` installed a `$state`-backed accessor for the field on this instance. */
+const isStatified = (instance: object, field: string) =>
+	Object.getOwnPropertyDescriptor(instance, field)?.get !== undefined;
+
+// Both engines are wrapped in `statify` (`$lib/engine/engine.ts`), which turns every *enumerable* field
+// into a `$state` accessor. Their clocks and loop bookkeeping are written on every tick/frame with no
+// reactive consumer, so they are `#`-private to stay out of the proxy (#2123) — `tickRate` is the sole
+// reactive field (the nav sidebar's rate readout). These pin that split, which is invisible at the call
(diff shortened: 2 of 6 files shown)


UI/src/lib/engine/render-engine.ts
```diff
@@ -6,53 +6,66 @@ const notifyRenderUpdate = renderUpdateHook.notify;
...
+
+	/** The render clock in `performance.now()` terms. Deliberately non-reactive — poll it, don't `$derived` it. */
+	public get time() {
+		return this.#time;
+	}
+
+	/** How far the render clock leads the logical clock (floored at 0). Non-reactive, like `time`. */
+	public get logicalDelta() {
+		return this.#logicalDelta;
+	}
```
ginderjeremiah: Nit: the PR body justifies all three getters together ("the render loop reads `logicEngine.time` across the class boundary, and the existing suites assert on both"), but only one of those reasons applies here. `logicEngine.time` has a genuine cross-class production reader — these two don't:

- Nothing outside this class reads `renderEngine.time`.
- The production consumer of `logicalDelta` is `battle-engine.ts:152`, which takes it from the `onRenderUpdate` payload, not the property.

So both getters exist purely so `render-engine.test.ts` can peek at private state — and that file already captures both values from the hook it subscribes to (`renderUpdates` at line 41 accumulates `[delta, logicalDelta]`), so the assertions at lines 129–156 could read from there instead and these two could be `#`-only.

Keeping them is defensible — a prototype getter over a `#`-field costs nothing reactively, and it's useful for debugging — but it is public API kept alive by tests, in a PR whose point is shrinking the engines' public surface. Whichever way you go, worth making the doc comment say which one it is (`logicEngine.time` is read by `RenderEngine.update`; these are observability/test hooks).

---

UI/src/tests/lib/engine/engine-statify.svelte.test.ts
```diff
@@ -0,0 +1,164 @@
...
+const rafCallbacks: (() => void)[] = [];
+vi.stubGlobal('window', {
+	requestAnimationFrame: vi.fn((cb: () => void) => {
+		rafCallbacks.push(cb);
+		return rafCallbacks.length;
+	}),
+	cancelAnimationFrame: vi.fn(),
+	setInterval: (...args: Parameters<typeof setInterval>) => globalThis.setInterval(...args),
+	clearInterval: (...args: Parameters<typeof clearInterval>) => globalThis.clearInterval(...args)
+});
```
ginderjeremiah, line 21: Nit: `return rafCallbacks.length` makes the handle unstable — the tests `shift()` callbacks off the front, so the same number gets handed out twice and a handle can go backwards. Nothing here asserts on cancellation so it's inert today, but the sibling file gets this right (`render-engine.test.ts` uses `++rafHandleCounter` and records `cancelledHandles`), and two `requestAnimationFrame` stubs that disagree is the kind of thing a later edit trips over. A monotonic counter is the same one line.

Broader nit for the summary rather than this line: this `vi.stubGlobal('window', …)` block is now the third copy (`logical-engine.test.ts:8`, `render-engine.test.ts:14`), and this one is effectively the union of the other two. A shared `tests/fixtures` helper (`stubEngineWindow()`) would cut it to one, in the same spirit as the double consolidation in #2470/#2487.

---

ginderjeremiah (commented):
## Review: #2462 — take the engines' per-tick internals out of the statify proxy

**Verdict: Approve** — the mechanism is sound, verified against `statify` rather than assumed, and the new suite pins the property that would otherwise silently regress. No blocking findings. Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR author.

The load-bearing claim here is that a public getter over a `#`-field stays out of the proxy, so I checked it rather than trusting it: `statify` walks `for (const property in state)` (`statify.svelte.ts:31`), and a class accessor lands on the prototype **non-enumerable**, so `for...in` never yields it. `statify` also mutates the instance in place and returns it — no wrapping Proxy — so `RenderEngine.update`'s cross-class `logicEngine.time` read still hits the prototype getter. Both halves hold.

Coverage against the issue is complete: `time`/`lastTime`/`timeBank` on the logical side, `time`/`logicalDelta`/`running`/`rafHandle` on the render side, `tickRate` left reactive for `NavSidebar.svelte:87,93`. Nothing outside either class ever wrote these — the existing suites only ever read `.time` — so the read-only getters break no call site.

### Nits

Three, none blocking: two inline (test-only public getters on `RenderEngine`; an unstable rAF handle in the new stub, plus the third copy of the `window` stub), and one on the framing below.

**The reactivity win is narrower than the PR body states.** "Every loop internal was a `$state` accessor with no subscriber" isn't quite right — `statify` skips function-valued fields outright (`statify.svelte.ts:34-36`), so `countTick` (a closure from `getEventCounter`) and `RenderEngine.update` (an arrow field) were never proxied. The actual signal writes removed are the seven number/boolean/object fields. Converting the two function fields is still the right call — they're loop internals and the `#` form says so — but the commit message is what future readers will trust when they wonder why this was done.

### Test coverage

`engine-statify.svelte.test.ts` is the right shape for this change, and worth its own file for the reason given: a `#`-field promoted back to public would pass every existing engine test unchanged, so nothing else in the suite can catch the regression. It asserts both directions — `tickRate` has a statify-installed own accessor and notifies a `$derived`; `time`/`logicalDelta` have no own descriptor and a `$derived` reading them never re-runs while the clocks demonstrably advance. That last part matters: asserting only the descriptor shape would pass on an engine whose clock had stopped working.

Dropping the `vi.mock('svelte', …)` `onDestroy` stub that the two sibling engine files carry is correct, not an omission — `onDestroy` is only reached via `hooks.ts:79` when a subscriber opts in with `cleanupOnDestroy: true`, and this file subscribes to no hooks (and does import the real `flushSync`).

Minor: `cleanup()` from `$effect.root` is called at the end of each test body rather than in a `finally`, so a failing assertion leaks the root. Harmless under vitest's per-file isolation — mentioning it only because the pattern gets copied.

The two new `zone-progression` cases are the right ones for the folded-in semantic change — retired-in-the-middle is skipped, retired-as-current yields `undefined` — and the second one carries the *why* (a player is lazily relocated out of a retired zone) rather than just the expectation.

### What's good

The `nextNavigableZone` rename is a better fix than the issue asked for, and for a better reason than the double sort. The issue framed it as a redundant ordering pass; the real problem with the old signature is that `nextZoneByOrder(zones, …)` would cheerfully hand back a **retired** zone to any caller that forgot the filter, and "the next zone to unlock" must never be one. Folding the filter in makes the wrong call unrepresentable, and the new name doesn't lie about what it returns. Renaming rather than quietly widening the behaviour is what makes it safe — the one existing caller had to be touched, so nothing keeps the old mental model.

Converting `update` from an arrow-function field to a real method while in the area is the right kind of drive-by: it was only ever invoked as `this.update()`, so binding was never load-bearing, and it now matches `LogicalEngine`.

I'd also call out the honesty of the closing note. The removed signal writes were cheap, the PR says so, and it claims no frame-time improvement — a perf-adjacent change that declines to claim a perf win is more trustworthy than one that invents a number.

---

ginderjeremiah:
Thanks — all three nits addressed in `9d45470`, plus the framing correction.

**rAF handle (inline).** Fixed: monotonic counter, reset per test, with a comment pointing at why `rafCallbacks.length` is wrong under `shift()`. You're right that it was inert here and right that two disagreeing stubs is the real cost.

**The `RenderEngine` getters (inline).** Kept, with the doc comment now saying which kind they are — I took your "make the comment say which one it is" option rather than the `#`-only one, for a reason worth stating: the assertions those getters serve are on the **absolute** clock (`engine.time` is 500, then 516), and the `onRenderUpdate` payload carries *deltas*. Reading them from `renderUpdates` would mean re-deriving the clock by summing deltas from a seed the test doesn't observe, which is a weaker assertion of a more roundabout kind. `logicalDelta` alone could come from the payload, but splitting the pair — one `#`-only, one a getter — would be worse than either consistent choice.

Your characterisation is accurate though, and I'd rather it be in the code than in a PR thread: the comment now says nothing in production reads either, names `RenderEngine.update` as `LogicalEngine.time`'s cross-class reader, and states they're observability plus a debugging surface.

**The third `window` stub.** Filed as #2489 rather than folded in: it edits `logical-engine.test.ts` and `render-engine.test.ts`, which this PR doesn't otherwise touch, and the issue can record the design question your comment implies — whether the helper stubs only the four members the engines reach or a fuller `window`. Same split you and I both used for #2470/#2487.

**The framing.** You're right, and it was the one thing here I'd call a real error rather than a nit: `statify` skips function-valued fields (`statify.svelte.ts:34-36`), so `countTick` and the arrow-field `update` were never proxied, and "every loop internal was a `$state` accessor" overstated it. The PR body now says the proxy loses exactly the seven number/boolean/object fields, and that converting the two function fields is a consistency call rather than a removed signal write. The commit message for that first commit is already specific — it enumerates the seven by name — so the misleading sentence was only ever in the body; the body is now corrected in place.

**`cleanup()` outside a `finally`** — deliberately left as-is: all three existing `$effect.root` reactivity suites (`inventory-manager-reactivity`, `battle-attributes`, and this one) use the end-of-body form, and a leaked root is inert under vitest's per-file isolation. Six `try`/`finally` wrappers to diverge from the established pattern is the wrong trade; if it's worth changing, it's worth changing all three suites at once, and that's an issue rather than a hunk here.

Re-ran after the changes: `npm run lint` clean (1230 files, 0 errors), `npm run test:unit:ci` 313 files / 4059 tests passing. Will merge once CI is green on the new head.

---

Issue #698, opened 2026-06-15
PlayerStatPoints.TryUpdateAttributes silently drops duplicate-attribute updates instead of rejecting (anti-cheat gap)


ginderjeremiah:
## Context

Found during a health review of `Game.Core/Players/PlayerStatPoints.cs` (`TryUpdateAttributes`, the `matchedUpdates.TryAdd(...)` block, ~`:45-54`).

`TryAdd` keeps only the **first** update per attribute id and silently discards any later update for the same attribute, then reports success.

## Why it matters

Per the documented anti-cheat model, stat allocation must **reject with no state mutation** on an invalid payload. A payload like `{ Str: +5, Str: -3 }` currently partially applies (`+5`) while the point accounting operates on a different set than the client actually submitted — the silent-partial-apply class of bug. The sibling loadout path (`TrySetSelectedSkills`) correctly rejects this via a `HasDuplicate` check, so the two anti-cheat paths are inconsistent.

## Suggested direction

Reject (`return false`, no mutation) when the payload contains a duplicate attribute id, matching `TrySetSelectedSkills`. Add unit tests for the duplicate-attribute payload (rejected, no mutation) and the normal multi-attribute payload (applied).

### Labels
bug, claude, scope: small

PR #2484, opened 2026-07-26
Validate stat allocations against Attribute.IsCore rather than row presence (#2466)

### LINKED ISSUE #2466: TryUpdateAttributes gates on row presence as a proxy for "allocatable" when Attribute.IsCore is the authoritative answer
Follow-up from #2459 / PR #2465.

## Context

`PlayerStatPoints.TryUpdateAttributes` rejects an allocation into an attribute the player has no `StatAllocation` row for (`Game.Core/Players/PlayerStatPoints.cs:40-43`, the #488 anti-cheat). The *intent*, per [backend.md](../blob/main/docs/backend.md), is "a stat allocation only accepts the core attributes ... which blocks allocating into derived attributes" — but the check doesn't ask that question. It asks whether a row happens to exist, and relies on an invariant maintained elsewhere (seeded by `NewPlayerFactory`, and now re-established on rehydration by `PlayerStatPoints.EnsureAllocatableAttributesArePresent`) to make the two coincide.

That indirection is what turned #2459 from a persistence bug into a permanent gameplay lockout: a write-behind handler deleting a row didn't just lose a stored `0`, it silently revoked the player's ability to ever allocate into that stat again.

## Suggested change

Validate against `Game.Core.Attributes.Attribute.IsCore(update.Attribute)` directly. A derived attribute is rejected because it is derived — the actual rule — rather than because nobody seeded it a row. The allocation amount would then come from the row if present and default to 0 otherwise (creating the row on first spend), removing the coupling entirely.

Worth thinking through before implementing:

- **Do the seeded rows still earn their keep?** They'd remain useful as the wire shape (`PlayerData.Attributes` / `UpdatePlayerStatsResponse.Attributes` project straight off the list, and the frontend renders its own `CORE_ATTRIBUTES` steppers regardless), so the seeding probably stays — but as a display/persistence convenience rather than an anti-cheat load-bearing invariant. If so, `EnsureAllocatableAttributesArePresent` becomes belt-and-braces rather than a correctness requirement, which is a nice place to end up.
- **`StatPointsUsed` consistency.** The current all-or-nothing reject contract and the `matchedUpdates` overflow handling (#698 duplicate rejection, the `long` accumulator) must survive the change unchanged.
- **`backend.md` wording.** The "actually have an allocation row" phrasing describes the mechanism, not the rule; it should be updated alongside.

## Scope

Small — one domain method plus its unit tests (`PlayerStatPointsTests` already covers the unknown-attribute, mixed-set, and duplicate-payload cases and would need reframing around derived-vs-core), and the socket-level rejection test in `UpdatePlayerStatsSocketTests`, which PR #2465 already retargeted at a derived attribute.

## Not urgent

With the invariant now enforced at both creation and rehydration, row presence and `IsCore` are equivalent for every real aggregate, so this is a robustness/clarity cleanup rather than an open defect.

ginderjeremiah:
Closes #2466.

Backend only — no frontend, battle-logic, or content changes.

## What's in it

`PlayerStatPoints.TryUpdateAttributes` rejected an allocation into an attribute the player had no `StatAllocation` row for. The *intent* is "only core attributes are directly allocatable", but the check asked whether a row happened to exist and relied on a seeding invariant maintained elsewhere to make the two coincide — which is what turned #2459 from a lost zero into a permanent gameplay lockout.

- **The check now asks `Attribute.IsCore(update.Attribute)` directly.** A derived attribute is rejected because it is derived.
- **A missing row is created at zero on first spend**, so row presence stops being load-bearing entirely.
- Created rows are appended to `StatAllocations` **only on the success path**, so the all-or-nothing contract still holds — a rejected payload leaves the allocation set exactly as it found it.
- A created row starts at `0`, so it is subject to the same non-negative rule as a stored one (a negative update against a missing row still rejects rather than handing out a free unallocation).

The `#698` duplicate rejection, the `matchedUpdates` overflow handling (the `long` accumulator), and the all-or-nothing reject contract are unchanged.

## The seeded rows stay

Per the issue's "do they still earn their keep?" — yes, but demoted. `CreateAllocations` / `Rehydrate` still give every player one row per core attribute, because that complete spread is the wire shape `PlayerData.Attributes` / `UpdatePlayerStatsResponse.Attributes` project off and the client reconciles against. What changed is their status: the rehydration repair is now belt-and-braces rather than a correctness requirement, and a missing row costs a stale display at worst instead of stranding a stat.

That reframing is the bulk of the diff. `backend.md`'s "only accepts the core attributes that actually have an allocation row" described the mechanism rather than the rule, and `backend-persistence.md`'s attribute-allocation paragraph led with row-presence-as-permission; both now state the actual rule, and the same stale reasoning is corrected in `PlayerStatPoints`, `NewPlayerFactory`, `PlayerCacheMapper`, `AttributeAllocationsChangedHandler`, and the test comments that cited it.

One note: `AttributeAllocationsChangedHandler` still upserts zeros and never deletes a row, unchanged — but its justification is now "a zero is a genuine value the player reallocated down to, and the complete spread is what the client renders" rather than "row presence is permission".

## Two things worth flagging

- **The issue named `PlayerStatPoints.EnsureAllocatableAttributesArePresent`; on `main` that method is `Rehydrate`.** Same responsibility, renamed since the issue was filed — I treated them as the same thing.
- **`Rehydrate_RestoredRow_BecomesAllocatableAgain` was deleted rather than adapted.** Its premise was that the damaged aggregate *rejects* the allocation until the repair runs; that is exactly the behaviour this change removes, so the test asserted the bug. The regression it guarded is now covered directly by `TryUpdateAttributes_CoreAttributeWithNoRow_CreatesTheRowAndApplies`, and `Rehydrate`'s own three tests still pin the restore.

## Test plan

New/reframed unit tests in `PlayerStatPointsTests` (a shared `DamagedStats` fixture models the #2459 shape — every core row but Strength gone):

- `TryUpdateAttributes_DerivedAttribute_RejectsWithoutMutating` — reframed from the old unknown-attribute test, which used `Luck` (a *core* attribute with no row) and would now correctly succeed.
- `TryUpdateAttributes_DerivedAttributeCarryingAStoredRow_IsStillRejected` — the converse of the old rule, and the sharpest statement of the new one: a stray non-core row (which `Rehydrate` preserves rather than prunes) is not a licence to allocate into it.
- `TryUpdateAttributes_CoreAndDerivedAttributes_RejectsEntireSet` — mixed set, all-or-nothing.
- `TryUpdateAttributes_CoreAttributeWithNoRow_CreatesTheRowAndApplies` — the lockout regression in its new form.
- `TryUpdateAttributes_NegativeUpdateOnAMissingRow_RejectsAsBelowZero`.
- `TryUpdateAttributes_RejectedSet_LeavesNoRowBehindForItsCoreUpdates` — a rejected payload must not leave the row its valid half would have created.

The socket-level `UpdatePlayerStats_DerivedAttribute_…` rejection test and `UpdatePlayerStats_CoreAttributeWithNoPersistedRow_IsStillAllocatable` pass unchanged (the new rule is a superset of the old behaviour for every real aggregate), as do the `#2459` integration tests in `PlayerServiceIntegrationTests` / `PlayerUpdateHandlerIdempotencyIntegrationTests`.

- `dotnet test` — **3563 tests passing** across all four suites (Core 1483, Application 1113, Api 893, Infrastructure 74).
- Frontend lint/tests not run: no `UI/` files are touched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UcsApr7USGB7cHpMeaVvfc)_


Game.Core/Players/PlayerStatPoints.cs
```diff
@@ -83,26 +86,32 @@ public static PlayerStatPoints Rehydrate(List<StatAllocation> statAllocations, i
...
 
-                if (!matchedUpdates.TryAdd(update.Attribute, (allocation, update)))
+                if (!allocationsByAttribute.TryGetValue(update.Attribute, out var allocation))
                 {
-                    return false;
+                    allocation = new StatAllocation { Attribute = update.Attribute, Amount = 0d };
+                    createdAllocations.Add(allocation);
                 }
+
+                matchedUpdates.Add(update.Attribute, (allocation, update));
```
ginderjeremiah: Nit — the comment above still enumerates the two rejection rules as separate bullets (#488 non-core, #698 duplicate), but the code now collapses them into one boolean and swaps the old `TryAdd` for `ContainsKey` + `Add`. Keeping one guard per documented rule reads closer to the comment and drops the double hash lookup:

```csharp
if (!CoreAttribute.IsCore(update.Attribute))
{
    return false;
}

if (!allocationsByAttribute.TryGetValue(update.Attribute, out var allocation))
{
    allocation = new StatAllocation { Attribute = update.Attribute, Amount = 0d };
    createdAllocations.Add(allocation);
}

if (!matchedUpdates.TryAdd(update.Attribute, (allocation, update)))
{
    return false;
}
```

Behaviourally identical: a duplicate targeting a rowless core attribute would allocate a throwaway `StatAllocation` into `createdAllocations` before the `TryAdd` fails, but that list is local and only published on the success path, so the all-or-nothing contract is unaffected. Take it or leave it.


---

ginderjeremiah (commented):
## Review: replace `TryUpdateAttributes`' row-presence gate with `Attribute.IsCore` (#2466)

**Verdict:** Approve (posted as a comment — GitHub won't accept a gating verdict from the PR's own author). No blocking findings; one nit and one pre-existing follow-up, neither of which needs to hold this up.

The change does what #2466 asked and does it in the right place: the rule "only core attributes are directly allocatable" is now asked of the attribute, and `Attribute.IsCore` is a `HashSet.Contains`, so an undefined enum cast (`(EAttribute)9999`) still rejects — the old row-presence check's only other real job. The three invariants the issue called out to preserve all survive: the #698 duplicate rejection, the `long` accumulator's overflow handling, and the all-or-nothing contract.

I traced the created-row path end to end and it holds up:

- `createdAllocations` is published inside the success branch before the amounts apply, and `matchedUpdates` holds the same object references, so the deltas land on the rows that were just appended.
- A created row starts at `0`, so the existing `Amount + Update.Amount >= 0` guard covers it — a negative update against a missing row rejects rather than handing out a free unallocation, which is the one way this change could have opened a hole.
- A duplicate attribute returns before a second lookup, so `allocationsByAttribute` never going stale after a create is safe rather than accidental.
- Downstream, `Player.TryUpdateAttributes` projects the event off `StatPoints.StatAllocations` and `UpdatePlayerStats` projects the response off the same list, so a row created here reaches both the client and the write-behind handler without a separate change.

### Nits

- `Game.Core/Players/PlayerStatPoints.cs:103-114` — the two rejection rules the comment lists as separate bullets are collapsed into one condition, and `TryAdd` became `ContainsKey` + `Add`. Inline suggestion; purely cosmetic.

### Follow-up (pre-existing, not this PR)

- Filed #2485: `TryUpdateAttributes` returns a bare "not rejected" bool, so an empty payload or a set of all-zero deltas reports success and still raises `AttributeAllocationsChangedEvent` + a write-behind save — the redundant-save spam `backend.md`'s "rejected vs accepted-but-unchanged" rule exists to prevent. Untouched by this change; the created-row path doesn't make it worse.

### Test coverage

Solid, and the reframing is honest about what it's replacing. `TryUpdateAttributes_DerivedAttributeCarryingAStoredRow_IsStillRejected` is the one that earns its keep most — it's the exact case the old implementation got wrong and the only test that would fail if someone quietly reverted to a row-presence check, since `Rehydrate` preserves non-core rows rather than pruning them. The negative-update-on-a-missing-row and rejected-set-leaves-no-row cases close the two ways the create could have leaked, and `DamagedStats` models the #2459 shape rather than an abstract one.

Deleting `Rehydrate_RestoredRow_BecomesAllocatableAgain` is right, not a coverage loss: its `Assert.False(damaged.TryUpdateAttributes(...))` premise asserted the behaviour being removed, so adapting it would have meant inverting the assertion into a duplicate of `TryUpdateAttributes_CoreAttributeWithNoRow_CreatesTheRowAndApplies`. `Rehydrate`'s own three tests still pin the restore itself. No frontend or battle-logic surface is touched, so no mirrored-test obligation.

### What's good

The doc/comment sweep is the part I'd have been most likely to skip and most regretted skipping. Row-presence-as-permission was stated as fact in `backend.md`, `backend-persistence.md`, and five code comments; leaving any of them would have left the next reader with a plausible, wrong mental model of an anti-cheat rule. Demoting the seeded rows to "the wire shape the client reconciles against" rather than deleting the rehydration repair is also the right call — the repair still has a job, just not a correctness-critical one, and the docs now say exactly that.

Flagging the `EnsureAllocatableAttributesArePresent` → `Rehydrate` rename in the PR body was useful; I verified they're the same method and the issue was just written against an older name.


---

Issue #11, opened 2026-06-03
Fix eslint errors in frontend


ginderjeremiah:
There are about 50 eslint errors in the frontend. Most of them are from lazy test writing. Fix the errors and evaluate/refactor tests for maintainability.

The errors related to the api classes can be ignored if they are not easily solvable. If you do ignore them, please disable the relevant eslint rule for those files.

PR #2487, opened 2026-07-26
Consolidate the duplicated ISessionStore/IUnitOfWork test doubles (#2470)

### LINKED ISSUE #2470: Consolidate the duplicated ISessionStore test doubles onto the shared FakeSessionStore

ginderjeremiah:
Closes #2470.

Test-only — no production code, no behaviour change. −184/+71 lines across 14 files.

## What's in it

Every private `NoOpSessionStore` copy is replaced with the shared `FakeSessionStore` (promoted in #2469) and deleted. The `IUnitOfWork` doubles get the same treatment, and after review feedback all three of them collapse into one shared recording fake.

| double | copies removed | now |
|---|---|---|
| `ISessionStore` | 10 | `Game.TestInfrastructure/Helpers/FakeSessionStore.cs` |
| `IUnitOfWork` | 3 distinct classes, 7 call sites | `Game.TestInfrastructure/Helpers/FakeUnitOfWork.cs` (new) |

`FakeSessionStore` with no `Session` assigned is behaviourally identical to every copy it replaces — `GetSession` returns `null`, writes are recorded and otherwise inert — which is all any of these call sites needed.

## Two of the issue's numbers didn't hold up

I checked the claims against the code rather than taking the file list as given, and both counts were low:

- **The issue names 4 files; there are 10.** On top of `RequestLoggingMiddlewareTests`, `ExceptionHandlingMiddlewareTests`, `ReferenceDataCommandAdminRedactionTests` and `SocketCommandProcessorTests`, the same copy also lives in `SocketContextTests`, `SocketReadLoopTests`, `SocketSerializationTests`, `SocketInactivityTests`, `SocketCommandExecutionTests` and `LoginTrackingMiddlewareTests`.
- **The drift is 7 of 10, not 2 of 4** — seven copies name the `Update`/`UpdateAsync` parameter `playerId` where the interface says `userId`, which is actively misleading given `ISessionStore`'s own comment that sessions are keyed by the user id and *not* `PlayerState.PlayerId`.

Neither changes the shape of the fix, but it does mean the consolidation was worth ~2.5× what the issue estimated.

## The `IUnitOfWork` fold-in

Five of those same ten files each carried an identical one-line `NoOpUnitOfWork`, so folding it in beat leaving a near-duplicate PR that touches them again a week later.

Review then pointed out that a no-op-only helper would leave `IUnitOfWork` with **three** doubles for a one-method interface — hard to defend in the same PR that documents the opposite rule. So `FakeUnitOfWork` records `CommitCount` and `LastToken`, and also replaces `CapturingUnitOfWork` (`SocketCommandExecutionTests`) and `RecordingUnitOfWork` (`CommitFilterTests`). Net two classes deleted rather than one added, matching the `FakeSessionStore` shape where recording surfaces are simply ignored by callers that only need the dependency satisfied.

That merge also removed a null-forgiving operator. `CapturingUnitOfWork` exposed `CancellationToken?`, so the #2174 per-command-budget test read `ReceivedToken!.Value.CanBeCanceled` — flagged by `CLAUDE.md`. With `CommitCount` available on the same double, `LastToken` is a plain non-nullable `CancellationToken`:

```csharp
Assert.Equal(1, unitOfWork.CommitCount);
Assert.True(unitOfWork.LastToken.CanBeCanceled);
```

Strictly stronger than before — it pins *exactly one* commit rather than "at least one", and `default(CancellationToken).CanBeCanceled` is `false`, so a missing commit still fails the second assertion on its own.

## Docs

`docs/backend.md` → Testing Guidelines now carries the convention: **one canonical double per interface**, in the shared test-infrastructure project *for that layer*. The layer scoping is load-bearing rather than pedantic — `Game.Core.Tests` references only `Game.Core` and `Game.Core.TestInfrastructure`, so a domain-only double genuinely cannot live in `Game.TestInfrastructure`. The projects blurb above it no longer scopes `Game.TestInfrastructure` to integration fixtures alone, since ten unit-test classes now depend on it.

The rule also states the preference the `IUnitOfWork` merge established (one double that records everything and is read selectively, over a no-op/recording pair) and narrows the carve-out to doubles that **script** a sequence, e.g. `FakePlayerRepository`'s null-then-recovers reload. Without a written rule this cleanup silently regrows; the issue's own motivation — N files to edit when the interface gains a member — is the argument that the rule was missing, not just the copies.

## Verification

- `dotnet build -warnaserror` — succeeded, 0 warnings.
- `dotnet format --verify-no-changes --no-restore` — clean (exit 0), which matters because the usings were re-sorted by hand after a mechanical edit and `IDE0005` isn't configured, so an orphaned using wouldn't fail the build.
- `dotnet test` (full solution, against Postgres/Redis containers) — **3,563 passed, 0 failed**: Game.Core.Tests 1483, Game.Api.Tests 893, Game.Application.Tests 1113, Game.Infrastructure.Tests 74. Re-ran `Game.Api.Tests` (893/893) after the `FakeUnitOfWork` merge.
- No new tests: this is a pure substitution with no new behaviour, and the existing suites are the regression guard — they either still pass against the shared doubles or the substitution wasn't equivalent. `grep` confirms no `NoOpSessionStore`/`NoOpUnitOfWork`/`CapturingUnitOfWork`/`RecordingUnitOfWork` declarations remain.
- Frontend untouched, so `npm run lint` / `test:unit:ci` were not run.

## Notes

- I considered a source-scanning guard test to stop a copy #11 being added, and decided against it: the repo's only source-scanning tests are content-graph and codegen ones, where the thing being pinned is generated. Inventing that machinery to catch a hand-written copy would cost more to maintain than the copies would.
- The same duplication exists for `FakeTimeProvider` and `FakeUserLogins` across two test projects — out of scope here (different files, unlike the `IUnitOfWork` copies which sat in files already being edited), tracked in #2488. The narrowed carve-out gives that issue a clearer rule to decide against.


docs/backend.md
```diff
@@ -20,6 +20,8 @@ Four `*.Tests` projects cover the layers (`Api`/`Application`/`Core`/`Infrastruc
 
 Unit tests follow the classical (Detroit) school — the project has no unmanaged dependencies, so avoid test doubles. A class that depends on an out-of-process dependency (database, cache) should hold no logic worth unit-testing; if it does, move that logic into a domain class and cover the dependency interaction with integration tests instead.
 
+Where a double is genuinely needed (usually to satisfy a constructor the test asserts nothing about), there is **one canonical double per interface**, in `Game.TestInfrastructure/Helpers` — reach for it rather than nesting a private copy in the test class. Interface-shaped duplication costs N edits every time the interface gains a member, and the copies drift (parameter names, behaviour) in the meantime. A local double is still right when it records or scripts something specific to that one test class.
```
ginderjeremiah: Nit: the location is stated absolutely, but `Game.TestInfrastructure` isn't reachable from every test project — `Game.Core.Tests` references only `Game.Core` and `Game.Core.TestInfrastructure` (`Game.Core.Tests.csproj:14-15`), and `Game.TestInfrastructure` pulls in Api/DataAccess/Infrastructure, so a Core-layer double genuinely can't live there. Today that's theoretical (Core is pure domain), but this line is now the source of truth and will get cargo-culted.

Suggest naming the layer's shared project rather than one path:

> …there is **one canonical double per interface**, in the shared test-infrastructure project for that layer (`Game.TestInfrastructure/Helpers`, or `Game.Core.TestInfrastructure` for domain-only doubles) — reach for it rather than…

Related: `docs/backend.md:13` still scopes `Game.TestInfrastructure` to "integration-test fixtures, container/marker plumbing", which no longer covers what this rule puts there — ten *unit* test classes now depend on it. Worth one clause on that line so the two statements agree.

---

Game.TestInfrastructure/Helpers/NoOpUnitOfWork.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    /// <summary>
+    /// An <see cref="IUnitOfWork"/> that commits nothing, for tests whose subject reaches the commit
+    /// boundary but asserts nothing about it. Use <c>CapturingUnitOfWork</c>-style local doubles only
+    /// when a test needs to observe the commit itself.
+    /// </summary>
+    public sealed class NoOpUnitOfWork : IUnitOfWork
+    {
+        public Task CommitAsync(CancellationToken cancellationToken = default) => Task.CompletedTask;
+    }
```
ginderjeremiah: Nit, and a genuine judgement call rather than a defect: `IUnitOfWork` now ends up with three doubles for one single-member interface — this one, `CapturingUnitOfWork` (`SocketCommandExecutionTests.cs:507`, records the token) and `RecordingUnitOfWork` (`CommitFilterTests.cs:58`, counts commits). A single recording `FakeUnitOfWork` exposing `CommitCount` + `LastToken` would cover all three call sites and delete two classes instead of adding one.

The reason I'd lean that way is the precedent sitting next to it: `FakeSessionStore` is exactly that shape — six recording surfaces, used inert in all ten of the call sites this PR converts. The carve-out on the docs line ("a local double is still right when it records or scripts something specific") is a good rule, but it lands weakly on an interface with one method, where "records something specific" is just "records the only call". Scripting doubles like `FakePlayerRepository` are where it clearly earns its keep.

Not asking for it here — but worth settling either way, since the doc line now makes it a project convention rather than a per-file choice. If the local variants are the deliberate answer, the docs sentence could say so with `IUnitOfWork` as the worked example.

---

ginderjeremiah (commented):
## Review: #2470 — consolidate the duplicated `ISessionStore`/`IUnitOfWork` test doubles

**Verdict: Approve** — mechanical, complete, and correct; no blocking findings. Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR author.

I verified the substitution rather than trusting the "behaviourally identical" claim, and it holds. `FakeSessionStore` with no `Session` assigned returns `null` from `GetSession`, records `Update`/`UpdateAsync`/`Clear` and is otherwise inert — exactly what every deleted copy did. Nothing on these paths reads back what it wrote, so the recording surfaces are invisible to all ten call sites, including the ones that call `SessionService.CreateSession(...)` first.

The two corrections to the issue check out too:

- **10 files, not 4.** Confirmed: `SocketContextTests`, `SocketReadLoopTests`, `SocketSerializationTests`, `SocketInactivityTests`, `SocketCommandExecutionTests` and `LoginTrackingMiddlewareTests` carried the same copy on top of the four the issue named. `grep` finds no `NoOpSessionStore`/`NoOpUnitOfWork` class declaration left anywhere outside the shared helper.
- **The `playerId`/`userId` drift is real and worth fixing**, given `ISessionStore`'s own comment that sessions are keyed by user id and explicitly *not* `PlayerState.PlayerId`. A parameter name that says the opposite of the interface is the kind of thing that gets read once and believed.

Also checked the two things a using-shuffle of this size usually breaks: no using was left behind unused (`Game.Abstractions.DataAccess` survives in `SocketCommandExecutionTests` for `IPlayerRepository`/`PlayerPersistenceFlushFailedException` and in `LoginTrackingMiddlewareTests` for `IUserLogins`, both legitimately), and `Game.TestInfrastructure` already references `Game.Application`, so `NoOpUnitOfWork` needed no new project reference. `IDE0005` isn't configured in `.editorconfig`, so an orphaned using wouldn't have been caught by `verify-backend` — worth the manual pass.

### Nits

Two, both inline: the docs line names one path where it should name the layer's shared project (`Game.Core.Tests` can't reference `Game.TestInfrastructure`), and `IUnitOfWork` now has three doubles where the `FakeSessionStore` precedent argues for one. Neither is a defect; the first is a two-line doc edit I'd fold in here.

### Follow-up

The same interface-shaped duplication this PR documents a rule against still exists elsewhere, outside this PR's scope:

- `FakeTimeProvider` — **byte-identical** implementations in `Game.Api.Tests/Unit/SocketCommandProcessorTests.cs:764` and `Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs:2826` (only the doc comment differs). `Microsoft.Extensions.TimeProvider.Testing` also ships one, so "adopt the package" is a live option.
- `FakeUserLogins` — `Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs:220` and `Game.Application.Tests/Services/LoginTrackingServiceTests.cs:38`. These two genuinely differ (call counting + throw-injection vs. token capture), so this one is a real judgement call under the new carve-out — but both are visible to the shared helper project now.

Right call to leave them: they're in different files across two test projects, unlike the `IUnitOfWork` copies which were in the same files being edited anyway. I'll open an issue.

### Test coverage

Appropriate as-is. This is a pure substitution with no new behaviour, and the 893 existing `Game.Api.Tests` cases are the regression guard — they either still pass against the shared double or the substitution wasn't equivalent. `verify-backend` (build `-warnaserror` + `dotnet format --verify-no-changes`) is green; `test-backend` was still running when I looked, and it's the check that actually proves the claim, so it needs to land green before merge.

I also agree with declining the source-scanning guard test. The repo's only source-scanning tests are content-graph and codegen ones, where the thing being pinned is generated; inventing that machinery to stop an eleventh hand-written copy would cost more to maintain than the copies would.

### What's good

Verifying the issue's file list instead of inheriting it is what made this worth doing — it more than doubled the scope, and a mechanical cleanup that only hits 4 of 10 copies is arguably worse than none, because the next reader sees both patterns and can't tell which is current.

Folding in `IUnitOfWork` was the right call for the right reason: same duplication, same files, already open. Splitting it out would have meant a second PR re-touching all five a week later for a one-line class.

The `docs/backend.md` line is the part that makes this stick. The issue's own motivation — N files to edit when the interface gains a member — is an argument that the *rule* was missing, not just that the copies were; without it this is a cleanup that silently regrows. Keeping `CapturingUnitOfWork`/`RecordingUnitOfWork` local is defensible on the same principle, even if I'd have gone the other way (see the inline).

---

docs/backend.md
ginderjeremiah: Fixed in 2fd7c2c — you're right, and I checked the reference chain rather than taking it on faith: `Game.Core.Tests.csproj` lists only `Game.Core` and `Game.Core.TestInfrastructure`, so a Core-layer double genuinely cannot live in `Game.TestInfrastructure`. Took your wording for the rule.

Also fixed the related line 12 blurb, slightly further than suggested: `Game.TestInfrastructure` now reads "integration-test fixtures, container/marker plumbing, and the shared unit-test doubles", and I added the constraint to the `Game.Core.TestInfrastructure` half — "the only one of the two `Game.Core.Tests` can reference — it pulls in nothing above `Game.Core`". That puts the *reason* the rule is layer-scoped next to the projects themselves, so the two statements can't drift apart again.

---

Game.TestInfrastructure/Helpers/NoOpUnitOfWork.cs
ginderjeremiah: Took it — you talked me into the merge rather than the doc caveat, in 2fd7c2c. Three doubles for a one-method interface was hard to defend while the same PR was arguing the opposite, and "records the only call" really is a weak reading of the carve-out.

`NoOpUnitOfWork` is now **`FakeUnitOfWork`** with `CommitCount` + `LastToken`, replacing all three: the five DI registrations, `CapturingUnitOfWork` (`SocketCommandExecutionTests`) and `RecordingUnitOfWork` (`CommitFilterTests`). Net two classes deleted instead of one added.

One thing that fell out and is worth noting, since it changes an assertion rather than just moving code. `CapturingUnitOfWork` exposed `CancellationToken?`, so the #2174 test read `Assert.NotNull(unitOfWork.ReceivedToken)` then `ReceivedToken!.Value.CanBeCanceled` — a null-forgiving operator `CLAUDE.md` asks us to avoid. With `CommitCount` on the same double, `LastToken` can be a plain non-nullable `CancellationToken`:

```csharp
Assert.Equal(1, unitOfWork.CommitCount);
Assert.True(unitOfWork.LastToken.CanBeCanceled);
```

That's strictly stronger than before — it pins *exactly one* commit rather than "at least one", and `default(CancellationToken).CanBeCanceled` is `false`, so the no-commit case still fails the second assertion on its own. The `!` is gone.

The docs sentence now states the general preference (one double that records everything, read selectively, over a no-op/recording pair — with `FakeSessionStore` and `FakeUnitOfWork` as the two worked examples) and narrows the carve-out to doubles that **script** a sequence, citing `FakePlayerRepository`'s null-then-recovers reload. That's the distinction you drew, and it makes #2488's `FakeUserLogins` call easier: those two record different things but script nothing, so the rule now points at merging them.

`dotnet build -warnaserror` clean, `dotnet format --verify-no-changes` clean, `Game.Api.Tests` 893/893.

---

PR #2490, opened 2026-07-26
Make Extensions.Capitalize culture-invariant (#2472)

### LINKED ISSUE #2472: Game.Core.Extensions.Capitalize uses culture-sensitive ToUpper() — the mirror of the codegen bug fixed in #2464
Found while reviewing #2471 (which fixed the codegen half of this, #2464).

`Game.Core/Extensions.cs:96`:

```csharp
public static string Capitalize(this string str)
{
    return str.Length == 0 ? str : string.Concat(str[0].ToString().ToUpper(), str.AsSpan(1));
}
```

This is the exact mirror of `CodeGenExtensions.Decapitalize`, which #2471 changed to `char.ToLowerInvariant`. `ToUpper()` is culture-sensitive: on a tr-TR machine a leading `i` becomes the dotted `İ` rather than `I`.

### Why it's latent rather than an active defect

Every call site derives a display name from an enum member name:

- `Game.Core/Players/Inventories/EquipmentSlot.cs:18` — `Value.ToString().Capitalize().SpaceWords()`, which reaches the wire as the slot's `Name`.
- `Game.Infrastructure/Database/GameContext.cs:331,384,793,815` — the same shape inside EF `HasData` seed values, so the result is baked into migrations and the database.

C# enum members are PascalCase, so the first character is already uppercase and `ToUpper()` is a no-op on every locale today. The bug only surfaces if a lowercase-leading name is ever fed in — but the seed path means a wrong result would be persisted rather than merely displayed, which is the reason to close it rather than leave it.

### Fix

`char.ToUpperInvariant(str[0])`, matching `Decapitalize`. Worth checking the sibling `SpaceWords` in the same file at the same time, and adding a `tr-TR` case to `Game.Core.Tests/ExtensionsTests.cs` using the shared `CultureScope` helper #2471 introduced (`Game.TestInfrastructure/Helpers/CultureScope.cs`).

Deliberately kept out of #2471: that PR is scoped to the codegen's byte-compared output, and this is a different subsystem with no drift guard over it.

ginderjeremiah:
Closes #2472.

`Game.Core.Extensions.Capitalize` used a culture-sensitive `ToUpper()` — the exact mirror of the codegen `Decapitalize` bug fixed in #2471. Swapped to `char.ToUpperInvariant(str[0])`, matching `Decapitalize`'s shape line for line.

## Why it matters despite being latent

The issue's framing checks out. Every call site feeds an enum member name (`EquipmentSlot.Name`, and four `HasData` seed values in `GameContext.cs`), and C# enum members are PascalCase, so `ToUpper()` is a no-op on every locale today. What makes it worth closing rather than leaving is the seed path: a wrong result there is *persisted* into migrations and the database, not merely mis-rendered.

## `SpaceWords` — checked, no change needed

The issue asked me to look at the sibling `SpaceWords` at the same time. It's already locale-independent: its `[GeneratedRegex("([a-z])([A-Z])")]` matches ordinal character ranges and is not `IgnoreCase`, so no culture-sensitive casing is involved. I left the implementation alone but added a tr-TR test pinning that property, so the guarantee survives if the pattern is ever revisited.

## One deviation from the issue: where `CultureScope` lives

The issue asks for a tr-TR case in `Game.Core.Tests/ExtensionsTests.cs` "using the shared `CultureScope` helper #2471 introduced (`Game.TestInfrastructure/Helpers/CultureScope.cs`)". That isn't reachable as written — `Game.Core.Tests` deliberately references only `Game.Core.TestInfrastructure`, and `Game.TestInfrastructure` pulls in `Game.Api`, EF, and Testcontainers. Widening `Game.Core.Tests`' references to reach one helper would break the isolation `docs/backend.md` calls out explicitly.

So I **moved** `CultureScope` down to `Game.Core.TestInfrastructure/Helpers/` (namespace changed to match) rather than duplicating it. It has no dependencies beyond `System.Globalization`, and `Game.TestInfrastructure` already references `Game.Core.TestInfrastructure`, so the three existing `Game.Api.Tests` consumers keep working with only their `using` updated — `CultureScope` was the sole thing each imported from that namespace, so no dead usings are left behind. `docs/backend.md`'s per-project inventory is updated to match.

## Test plan

- [x] New: `Capitalize_TurkishCulture_UppercasesInvariantly` (`"iron"`, `"intelligence"`) — **verified non-vacuous** by temporarily reverting the fix and confirming both cases go red, which also proves ICU is active here rather than the runtime silently falling back to invariant globalization.
- [x] New: `SpaceWords_TurkishCulture_BreaksWordsInvariantly`, pinning the regex's locale-independence.
- [x] Existing `Capitalize`/`SpaceWords` theories and the three relocated-`CultureScope` consumers in `Game.Api.Tests/CodeGen` still pass.
- [x] `dotnet test` (full solution) — **3566 passed, 0 failed** (Core 1486, Application 1113, Api 893, Infrastructure 74).

No frontend files are touched, so the UI suites aren't in scope. No migration changes: the fix is a no-op on the PascalCase inputs the seed values actually use, so the persisted values are byte-identical.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WCeNRzMmboN6CY2vw9vK1a)_


ginderjeremiah (commented):
## Review: culture-invariant `Extensions.Capitalize` (#2472), plus relocating `CultureScope` so `Game.Core.Tests` can reach it

**Verdict: Approve** — the fix is correct and minimal, the one deviation from the issue is the right call, and I found nothing blocking or worth changing. (Submitted as a `COMMENT` event only because GitHub won't take a gating verdict from the PR's own author.)

### What I verified

- **The fix mirrors its sibling exactly.** `Game.Core/Extensions.cs:99` is now `string.Concat(char.ToUpperInvariant(str[0]).ToString(), str.AsSpan(1))` against `Game.Api/CodeGen/Extensions.cs:16`'s `char.ToLowerInvariant(...)` — same shape, same allocation profile, same empty-string guard. Two methods that are conceptual mirrors staying textual mirrors is worth something on its own.
- **No behavioral change on any current call site.** All five (`EquipmentSlot.cs:18` and `GameContext.cs:331,384,793,815`) feed `enum.ToString()`, which is PascalCase, so `ToUpper()` was already a no-op. The claim in the description that seed values are byte-identical and no migration is needed holds up.
- **The `CultureScope` move is sound and complete.** `Game.Core.Tests` references only `Game.Core` + `Game.Core.TestInfrastructure`, and `Game.TestInfrastructure` already references `Game.Core.TestInfrastructure`, so the three `Game.Api.Tests/CodeGen` consumers keep resolving through the reference chain. I grepped every remaining `using Game.TestInfrastructure.Helpers;` — each one resolves other types still in that namespace (`PollingHelper`, `DatabaseCleaner`, `TestAuthHelper`, …), so nothing is left importing a namespace it no longer needs. Moving rather than duplicating was clearly right; widening `Game.Core.Tests`' references to reach `Game.TestInfrastructure` would have dragged in `Game.Api`, EF, and Testcontainers and broken the isolation `docs/backend.md` calls out.
- **`SpaceWords` genuinely needed no change.** `[GeneratedRegex("([a-z])([A-Z])")]` without `RegexOptions.IgnoreCase` matches ordinal ranges, so it's already locale-independent — "fixing" it would have been a no-op edit. Leaving the implementation alone and pinning the property with a test is the better of the two options the issue left open.

### Test coverage

Sufficient. The `tr-TR` cases are the right shape, and the description's note that the `Capitalize` case was checked for non-vacuity by reverting the fix is the part that matters most — a globalization test that silently passes because ICU isn't active is worse than no test. `"iron"`/`"intelligence"` both exercise the dotted-İ path, and the `SpaceWords` case pins a guarantee that is currently free but easy to lose if the pattern is ever revisited. Nothing else in the diff carries logic that needs covering.

### Follow-up (not a finding, nothing to change here)

Opened #2491: this is the third time in two days we've fixed culture-sensitive casing by having someone read the line (#2464 → #2471, then #2472 → this PR). No Globalization CA rules are configured — `Directory.Build.props` sets only `TargetFramework`/`ImplicitUsings`/`Nullable`/`EnforceCodeStyleInBuild`, and `.editorconfig` carries no analyzer severities. Enabling the category repo-wide needs its own triage pass, so it correctly stays out of this two-line fix.

### What's good

The description does the thing I most want from a PR: it states where it deviated from the issue (`CultureScope`'s home) and why the issue's instruction wasn't reachable as written, rather than quietly doing something else or following it into a layering violation. Checking the sibling `SpaceWords` and reporting "already fine, pinned it anyway" is the same instinct.

---

Issue #2474, opened 2026-07-26
Add PlayerWriteWatermark and reject stale absolute writes in the write-behind handlers


ginderjeremiah:
Implementation step 2 of [Spike #2467](https://github.com/ginderjeremiah/GameServer/issues/2467) — this is the change that actually closes the cross-instance stale-absolute-overwrite hazard. **Depends on #2473** (the envelope sequence). See [`docs/spikes/2467-cross-instance-write-behind-sequencing.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/2467-cross-instance-write-behind-sequencing.md) for why each fork was decided this way.

## The guard

New table `PlayerWriteWatermark(PlayerId, Stream, TargetKey, LastAppliedSequence)`. A handler applies its write unless the event's sequence is **strictly older** than the target's `LastAppliedSequence` — equal sequences apply.

**The compare must *be* the write.** A read-then-compare-then-apply does not survive the very race this closes: two instances applying the same player's events concurrently both read `LastAppliedSequence` = 4, both pass, and under `READ COMMITTED` whichever commits last wins the data row — which can be the older event. So:

```sql
UPDATE "PlayerWriteWatermark" SET "LastAppliedSequence" = @seq
WHERE "PlayerId" = @p AND "Stream" = @s AND "TargetKey" = @k AND "LastAppliedSequence" <= @seq
```

takes the row lock first; the data write runs only on rows-affected > 0; both in one transaction. The watermark row is the per-target serialization point — that is the load-bearing mechanic, not an implementation detail. A missing row needs an insert-if-missing seeded at `@seq` first (or `INSERT … ON CONFLICT DO UPDATE … WHERE`) so a first-ever write for a target isn't rejected.

**Note the operands are reversed between the rule and the SQL.** The rule is stated on the *event's* sequence (reject when `eventSeq < watermark`); the predicate is stated on the *column* (accept when `watermark <= eventSeq`). They agree — but an earlier draft of this issue had the SQL as `<`, so it's worth reading twice. Under `<` the predicate rejects equal sequences, which breaks same-save siblings landing on one target (the first advances the row, the second is silently skipped) and leaves the insert-if-missing seed with no working value (seeded at `@seq`, the following update immediately no-ops). Either is independently sufficient to require `<=`. The cost is that an exact duplicate re-applies rather than being skipped — one redundant no-op write, the correct trade under at-least-once, so don't tighten it back.

[...]

PR #2492, opened 2026-07-26
Carry a per-player write sequence on the write-behind envelope and the producing aggregates (#2473)

### LINKED ISSUE #2473: Carry a per-player write sequence on the write-behind envelope and the producing aggregates
Implementation step 1 of [Spike #2467](https://github.com/ginderjeremiah/GameServer/issues/2467) — see [`docs/spikes/2467-cross-instance-write-behind-sequencing.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/2467-cross-instance-write-behind-sequencing.md) for the reasoning behind each decision below. This issue delivers the **producer side only**: after it lands, every enqueued envelope carries a usable sequence and nothing yet reads it. No behaviour change on its own — the guard that consumes it is #2474.

## Scope

- **A monotonic `long` counter per write-behind aggregate.** `Player` and `PlayerProgress` each own their own, incremented **once per save** (not once per event). Two independent sequence spaces, deliberately not shared: the two aggregates are separate enqueues, ordered independently today, and write disjoint tables.
- **Counters start at 1, never 0.** 0 is reserved as the "unsequenced" sentinel (below), so a first save stamping 0 would silently opt itself out of the guard. Cold-load seeding is `COALESCE(MAX("LastAppliedSequence"), 0)` with increment-then-stamp, so the first stamp is 1.
- **The counter lives on the `Game.Core` aggregate**, not in the data tier. `backend.md` keeps persistence detail out of `Game.Core`, so this is a deliberate call: the precedent is `PlayerProgress`, which already carries write-behind plumbing in the domain (`_dirtyStatistics`/`_dirtyChallenges`/`_dirtyProficiencies` and their `Dirty*` projections). `Player` carries nothing comparable and does gain its first write-behind-queue concern. The data-tier alternative has no home — it would need a connection-scoped holder the per-command DI scope doesn't provide.
- **Persist the counter in each aggregate's cached representation** — `PlayerCacheModel` for the player, a reserved field in the progress hash — so it survives reconnect and cross-instance rehydration, which are the cases that matter. The producing side needs no distributed coordination because a player has exactly one live socket.
- **Seed on a cold DB load** from the player's highest persisted `LastAppliedSequence`. That table arrives with #2474, so until then a cold load seeds from 0 and stamps from 1; sequencing the two issues in this order is fine because nothing reads the stamp yet.
- **`Sequence` on `DomainEventEnvelope`**, defaulted so an envelope enqueued by a pre-upgrade instance mid-rolling-deploy still deserializes.
- **Stamp at buffer time, not flush time** (`PlayerUpdateBatch.Add` / the publisher, not `FlushAsync`). This matters: when a failed flush carries envelopes forward into the next save's flush (#1494), the carried events keep their original lower sequence, so a later out-of-order apply correctly rejects the older one. Stamping at flush time would give both saves the same value and lose that.
- **Thread it to handlers via a small `PlayerUpdateContext`** off `PlayerUpdateEventDispatcher`, rather than widening `IPlayerUpdateHandler<T>.HandleAsync` with a positional parameter. The handlers ignore it until #2474.

## Sequence 0 means "unsequenced", not "oldest"

This is the part most likely to be implemented wrong, because the obvious reading of a defaulted `long` is "a low sequence". It isn't. `DomainEventEnvelope.Id` defaults to a fresh `Guid` — a harmless fabricated value that only needs to be unique — but a defaulted `Sequence` of 0 is the smallest value the guard can compare, i.e. the value guaranteed to lose every comparison. Treating it as a real sequence loses writes during exactly the window the default exists to survive: a player whose watermarks an upgraded instance already advanced to 3, who then reconnects onto a pre-upgrade instance mid-deploy, would have every guarded write discarded for that whole session.

So **0 bypasses the guard entirely in #2474** — applied unconditionally, watermark untouched, which is precisely the pre-guard behaviour and the right semantic for an envelope carrying no ordering information. Nothing to implement here beyond starting real counters at 1, but the sentinel meaning belongs in the XML doc comment on `Sequence` so #2474 and every later reader inherit it rather than re-deriving it.

## Notes

- All events of one save share one sequence. The guard in #2474 rejects only when the event's sequence is **strictly older** than the watermark, so same-sequence siblings all apply and a duplicate re-apply stays idempotent under the queue's at-least-once contract. (#2474's SQL states the same rule on the *column* rather than the event — `WHERE "LastAppliedSequence" <= @seq` — so the operands are reversed between the two; they agree.) The corollary is that the sequence imposes **no ordering within a save**, which is fine (a save's events are raised from one consistent aggregate state) but is worth a comment at the stamping site, since "per-player write sequencing" reads like a total order.
- `PlayerUpdateEnvelopeReader` already sniffs `playerId` generically off the payload; the sequence is envelope-level so it needs no equivalent peek.

## Tests

- Unit: the counter advances once per save regardless of how many events that save raises; the first save of a brand-new aggregate stamps **1**, not 0.
- Unit: the counter round-trips through the cache model / progress hash; a cold load seeds from the persisted value.
- Unit: an envelope deserialized from JSON with no `"sequence"` property yields 0 rather than failing (the rolling-deploy case, mirroring the existing `DomainEventEnvelopeTests` coverage for `Id`) — with the test named so it reads as the sentinel case, not a value case.
- Integration: a save whose flush fails and is carried into the next save's flush produces two envelopes with **different** sequences, in the right relative order.

ginderjeremiah:
Closes #2473.

Producer side only, as the issue scopes it: every enqueued envelope now carries a usable sequence and **nothing reads it**. No behaviour change on its own — the guard that consumes it is #2474.

## What's in it

- **A `long` counter per write-behind aggregate.** `Player` and `PlayerProgress` each own one, advanced **once per save**, not once per event. Two independent spaces — the aggregates are separate enqueues writing disjoint tables.
- **Counters start at 1.** `AdvanceWriteSequence()` increments before returning, so a cold-loaded aggregate's first stamp is 1 and never the sentinel.
- **`DomainEventEnvelope.Sequence`**, defaulted so a pre-upgrade instance's envelope still deserializes, with the sentinel meaning documented on a `DomainEventEnvelope.Unsequenced` constant rather than left for each later reader to re-derive.
- **Stamped at buffer time** (`PlayerPersistencePublisher.Buffer`, off a sequence the batch's player-save scope carries), not at flush time — so envelopes carried forward from a failed flush into a later save's flush (#1494) keep their original lower value. There's an integration test for exactly that interleaving.
- **Persisted in each aggregate's cached representation** — the player blob, and a reserved `_seq` field on the progress hash — so the counter survives reconnect and cross-instance rehydration instead of restarting and re-stamping used values.
- **Cold DB load seeds from 0** and stamps from 1. Seeding from the highest persisted watermark arrives with the table that holds them (#2474); safe to defer since nothing reads the stamp yet.
- **`PlayerUpdateContext`**, a per-drain-scope holder the dispatcher populates, so handlers can read envelope-level metadata without widening `IPlayerUpdateHandler.HandleAsync` for the few that will.

## Two decisions worth your read

**`PlayerCacheModel.WriteSequence` is deliberately *not* `required`** — the one member on that model that isn't. `required` would have been the natural choice given the model's structural-completeness property (the DB projection can't silently omit a field), but System.Text.Json enforces `required` on deserialize, so a blob cached by a pre-upgrade instance would throw on its first upgraded read. That does self-heal — `GetPlayer` deletes the key and reloads from the DB — but it would do so for *every* active player on deploy day, an Error log and a database reload apiece. Defaulting to 0 instead makes a legacy blob rehydrate exactly like a cold load. `PlayerQueryExtensions`' doc records the exception so the completeness claim stays honest.

**The progress counter is aggregate-level state living in a row-keyed hash.** It gets a reserved `_seq` field outside the `S_`/`C_`/`P_` row-kind prefixes, grouping with the existing `_` presence marker and ignored by the row parsing exactly as any unrecognised field always was. One consequence surfaced in an existing test: `Save_OnlyWritesTheRowsThatChanged_LeavingEarlierCachedRowsInTheHashUntouched` asserted every field an earlier save wrote is still byte-identical, which `_seq` (rewritten by every enqueuing save) breaks. That's correct behaviour rather than a regression, so the test now scopes its unchanged-fields loop to row fields and asserts separately that `_seq` advanced — leaving it merely excluded would have let a "stopped advancing the counter in the cache" regression hide behind the test's row focus.

## One thing I'd flag

The issue notes the counter on `Game.Core` is a deliberate call against `backend.md`'s keep-persistence-detail-out-of-the-domain rule, justified by `PlayerProgress`'s existing `_dirty*` precedent. I agree with it — there genuinely is no connection-scoped home in the data tier — but it's worth naming that `Player` gains its first write-behind-queue concern here, where `PlayerProgress` already had several. If you'd rather that seam were explicit (a small `IWriteSequenced` or similar) rather than two independent counters with cross-referencing docs, that's a cheap follow-up and I'd happily take it.

## Test plan

Backend suite green locally — 3,585 tests, 0 failures (`dotnet build` + `dotnet test`), against the session's Postgres/Redis containers.

Covering the issue's listed cases:

- **Unit** — the counter advances once per save regardless of how many events that save raises; a brand-new aggregate's first stamp is **1**, not 0 (`PlayerTests`, `PlayerProgressTests`).
- **Unit** — the two aggregates' counters are independent (`PlayerProgressTests`).
- **Unit** — an envelope deserialized with no `"sequence"` yields the sentinel, named so it reads as the sentinel case rather than a value case; plus a wire round-trip (`DomainEventEnvelopeTests`).
- **Unit** — the scope's sequence is exposed for the publisher, restored on disposal, and a nested `SavePlayer` scope hands back the *outer* scope's value rather than the sentinel (`PlayerUpdateBatchTests`).
- **Unit** — the dispatcher publishes the envelope's sequence onto the scope's context, and leaves it at the sentinel for an unsequenced envelope (`PlayerUpdateEventDispatcherTests`).
- **Unit** — the counter round-trips through the lean cache model, and a blob with no `writeSequence` seeds from 0 rather than failing (`PlayerCacheMapperTests`).
- **Integration** — one sequence per save shared by all of that save's events (`[1, 1, 2]` across a two-event save then a one-event save), and **a failed flush's carried-forward envelope keeps its original lower sequence** (`[1, 2]`, the case that makes buffer-time stamping load-bearing) (`PlayerWriteBehindTests`).
- **Integration** — the progress hash carries the counter across scopes (`1` then `2` on successive cold-then-warm loads), and the scope's read memo carries it too so a same-scope reload doesn't re-stamp a used value (`PlayerProgressRepositoryIntegrationTests`).

No frontend changes, so the UI suites are untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NmNRsVdfcrPrftoeuvJ6vH)_


Game.DataAccess/Repositories/PlayerProgressRepository.cs
```diff
@@ -425,6 +458,16 @@ private static CachedPlayerProgress FromHashFields(Dictionary<string, string> fi
...
                     progress.Proficiencies.Add(proficiency);
                 }
+                else if (field == WriteSequenceField)
+                {
+                    // An unparseable counter is treated as absent rather than as corruption: unlike a row, it
+                    // carries no player data, and reseeding from 0 only costs the guard its ordering information
+                    // for this session — where discarding the whole key would throw away real progress rows.
+                    if (long.TryParse(value, NumberStyles.Integer, CultureInfo.InvariantCulture, out var sequence))
+                    {
+                        progress.WriteSequence = sequence;
```
ginderjeremiah, line 473: **Should-fix:** the "treat an unparseable counter as absent" decision is right, but it's the only self-heal in this file that happens *silently*. The corrupt-row path 40 lines up logs an Error before deleting the key, and `PlayerRepository.GetPlayer`'s corrupt-blob path does the same — the codebase's consistent posture is that a cache value the reader chose to discard is observable.

It matters more here than the comment implies, once #2474 lands: reseeding from 0 doesn't just cost "the guard its ordering information for this session", it makes the session re-stamp 1, 2, 3… against a watermark that may already be at 12, so *every* write from that session is rejected as stale until the counter climbs back past it. That's a durable, silent divergence with no log line pointing at it. A `_logger.LogWarning` in the `else` naming the player id and the raw value would make it diagnosable for the price of one line.

Also: this is a new branch with no test on the failure side. Both `FromHashFields` sequence tests exercise the parse-succeeds path; per `CLAUDE.md`'s "thoroughly tested for a wide range of scenarios, including edge cases and error conditions", the unparseable case wants one too — it's cheap to write given the integration test already reaches into the hash directly.

---

Game.Core/Progress/PlayerProgress.cs
```diff
@@ -44,18 +44,37 @@ public class PlayerProgress
...
+        /// </summary>
+        public long WriteSequence => _writeSequence;
+
         public PlayerProgress(
             Player player,
             IEnumerable<PlayerStatistic> statistics,
             IEnumerable<PlayerChallenge> challengeProgress,
-            IEnumerable<PlayerProficiency> proficiencies)
+            IEnumerable<PlayerProficiency> proficiencies,
+            long writeSequence = 0)
```
ginderjeremiah: **Should-fix:** the `= 0` default makes forgetting to seed the counter compile cleanly, and the failure mode of forgetting is exactly the one the whole PR exists to prevent — a hydration path that silently restarts at 0 and re-stamps values the player's earlier saves already used. Post-#2474 that is discarded writes, not a lost optimisation, and it fails silently at every layer.

Only three call sites pass through here (`Load`, plus two in `PlayerProgressRepositoryIntegrationTests`), so making it a required positional parameter costs two explicit `0`s in tests and buys a compile error for every future hydration path. Given the PR is otherwise careful to make the sentinel impossible to reach by accident (`AdvanceWriteSequence` incrementing before returning, `Unsequenced` as a named constant), this is the one place it's reachable by omission.

Related, and the reason I'd rather see it now than later: the two aggregates expose the same concept two different ways — `Player.WriteSequence` is a `{ get; init; }` set through the object initialiser, this one is a get-only property fed by an optional ctor arg. Same concept, same lifecycle, same doc comments cross-referencing each other, two shapes. `Player`'s `init` is the better of the two (it can't be silently omitted because `PlayerCacheMapper` is the only constructor). Worth converging on it.

---

Game.DataAccess/PlayerUpdates/PlayerUpdateContext.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal sealed class PlayerUpdateContext
+    {
+        /// <summary>
+        /// The producing aggregate's write sequence for the envelope being applied, or
+        /// <see cref="DomainEventEnvelope.Unsequenced"/> when it carries none (#2473). No handler reads it yet
+        /// — the guard that rejects a write older than its target's watermark is #2474.
+        /// </summary>
+        public long Sequence { get; private set; } = DomainEventEnvelope.Unsequenced;
+
+        public void SetSequence(long sequence) => Sequence = sequence;
```
ginderjeremiah: **Nit:** `{ get; private set; }` plus a public `SetSequence` is the same access surface as `{ get; set; }` — any holder of the instance can write it, so the ceremony buys no encapsulation. It also doesn't scale: #2474 or later metadata means a second `SetX` method rather than another property.

If the intent is "only the dispatcher writes this", `{ get; internal set; }` won't help (the whole type is already `internal`), so either drop to a plain settable property, or make the write genuinely one-shot — e.g. a single `Describe(DomainEventEnvelope envelope)` that populates everything the context carries, which reads better at the dispatcher call site and stays one method as the context grows.

---

Game.DataAccess/PlayerUpdateBatch.cs
```diff
@@ -65,11 +76,20 @@ public void OnFlushed(Action action)
...
+        /// <para>
+        /// <paramref name="writeSequence"/> is the sequence buffered envelopes are stamped with for the duration
+        /// of the scope; the previous value is restored on disposal, so an inner <c>SavePlayer</c> scope opened
+        /// inside an outer <see cref="Repositories.PlayerRepository.BeginBatch"/> one doesn't leak its sequence
+        /// to whatever the outer scope buffers afterwards. Callers that bracket saves rather than performing one
+        /// (<c>BeginBatch</c>) omit it.
+        /// </para>
         /// </summary>
-        public IDisposable BeginPlayerSave()
+        public IDisposable BeginPlayerSave(long writeSequence = DomainEventEnvelope.Unsequenced)
```
ginderjeremiah: **Nit:** same shape as the `PlayerProgress` ctor comment — an optional parameter defaulting to the sentinel means "I forgot to pass a sequence" and "I deliberately have no sequence" are the same call. There is exactly one deliberate caller (`PlayerRepository.BeginBatch`), so the alternative is cheap: make the parameter required and have `BeginBatch` pass `DomainEventEnvelope.Unsequenced` explicitly, which also puts the intent at the call site where a reader of `BeginBatch` will actually see it.

Lower stakes than the `PlayerProgress` one — an unsequenced envelope is bypassed rather than discarded by the #2474 guard, so this degrades to today's behaviour instead of losing writes. Take it or leave it.

---

ginderjeremiah (commented):
## Review: producer-side per-player write sequencing on the write-behind envelope (#2473)

**Verdict:** Approve (posted as a comment — GitHub won't take a gating verdict from the PR author) — no blocking findings; two should-fixes and two nits worth addressing before merge, none of which change the design.

This lands the producer half cleanly and the reasoning in the code matches what the spike decided. I traced the parts that would be quietly wrong if they were wrong, and they hold up:

- **`PlayerWriteSequence` can never be read outside a sequenced scope.** `PlayerPersistencePublisher.Buffer` reads it, and the only path that reaches the publisher is `IDomainEventDispatcher.DispatchAsync(player, …)` — one call site, inside `SavePlayer`'s `BeginPlayerSave(player.AdvanceWriteSequence())`. A bare `BeginBatch` scope brackets saves but never buffers player envelopes directly, so the `Unsequenced` default on `BeginPlayerSave` is unreachable in practice today. The nested save/restore is the right call regardless (`OfflineProgressService` is exactly the shape that would break without it).
- **Both counter-reseed paths funnel through the two documented #2474 seeding points.** The progress corrupt-hash self-heal deletes the key and falls through to `LoadFromDb`; the player corrupt-blob self-heal falls through to `GetPlayerModelFromDb`/`PlayerQueryExtensions`. Neither is a third reseed site that #2474 would have to find separately, which is what makes deferring the seed genuinely safe rather than just currently harmless.
- **The flush-failure/cache-blob interaction is consistent in both directions.** A failed flush skips the blob write *and* strands the buffered envelope with the scope, so the forced reload's lower counter re-stamps a value nothing ever enqueued. The reverse — flush succeeds, the fire-and-forget blob write drops — produces two envelopes at the same sequence, which #2474's `<= @seq` rule applies rather than discards. Worth knowing; not worth engineering around.

### Should-fix

- `Game.DataAccess/Repositories/PlayerProgressRepository.cs:461` — an unparseable `_seq` is swallowed with no log, unlike every other cache self-heal in this file, and post-#2474 the consequence is a session's writes being silently rejected rather than merely unordered. Also the one new branch with no test. (inline)
- `Game.Core/Progress/PlayerProgress.cs:62` — the optional `writeSequence = 0` ctor parameter makes "forgot to seed the counter" a clean compile, and forgetting is precisely the failure this PR exists to prevent. Three call sites; making it required is nearly free. Related: `Player` and `PlayerProgress` expose the same concept two different ways (`init` property vs. optional ctor arg) — converge on `Player`'s. (inline)

### Nits

- `Game.DataAccess/PlayerUpdates/PlayerUpdateContext.cs:21` — `{ get; private set; }` + public `SetSequence` has the same access surface as a plain settable property; consider a single `Describe(envelope)` instead, which stays one method as the context grows. (inline)
- `Game.DataAccess/PlayerUpdateBatch.cs:87` — same optional-sentinel shape as the `PlayerProgress` ctor, but much lower stakes since an unsequenced envelope is bypassed rather than discarded. (inline)

### On the question you raised

> If you'd rather that seam were explicit (a small `IWriteSequenced` or similar) rather than two independent counters with cross-referencing docs…

I wouldn't, at least not yet. An interface over two implementations that share no call site — nothing polymorphically consumes "a write-sequenced aggregate", `SavePlayer` and `PlayerProgressRepository.Save` each know their concrete type — buys naming at the cost of a `Game.Core` abstraction that exists only to describe a data-tier concern. That's the wrong trade in the layer `backend.md` already says to keep persistence detail out of. The duplication that actually bites is the cross-referencing doc comments, and the fix for that is the point above: give both aggregates the same *shape*, so the second one's doc can be short because the reader already recognises it. If a third write-behind aggregate ever appears, revisit.

### Test coverage

Good, and the tests are doing real work rather than restating the implementation. The two that earn their keep:

- `SavePlayer_EnvelopesCarriedForwardFromAFailedFlush_KeepTheirOriginalLowerSequence` pins the one behaviour that makes buffer-time stamping load-bearing, and it does it against the real batch with a throwing pub/sub rather than a cancelled token — so the queue contents stay exactly assertable. This is the test that would catch a "simplify" refactor moving the stamp to `FlushAsync`.
- Rescoping `Save_OnlyWritesTheRowsThatChanged…` to row fields *and* separately asserting `_seq` advanced, rather than just excluding it, is the right instinct — the exclusion-only version would have hidden a stopped-advancing regression behind the test's row focus.

Gaps, both small: the unparseable-`_seq` branch (above), and nothing pins that a `SavePlayer` raising **zero** events still advances the counter while a `PlayerProgress` save with no dirty rows deliberately does not. That asymmetry is stated in two comments as though it were a rule, and it's the kind of thing a later reader "fixes" into consistency. A test on each side would make the intent enforceable rather than advisory.

### What's good

- The `Unsequenced` constant with the sentinel semantics documented once, rather than each future reader re-deriving "0 doesn't mean oldest". Given how easy that specific mistake is to make, putting it on a named constant with the failure mode spelled out is the highest-value thing in the diff.
- Not making `PlayerCacheModel.WriteSequence` `required` is the right call and the PR body's reasoning is exactly right — the self-heal would have worked, but paying an Error log and a DB reload per active player on deploy day to preserve a structural-completeness property is a bad trade. Recording the exception in `PlayerQueryExtensions`' doc keeps that claim honest instead of quietly false.
- Grouping `_seq` with the existing `_` presence marker outside the `S_`/`C_`/`P_` prefixes means a pre-upgrade instance reading the hash is unaffected for free, rather than by a compatibility branch.
- `docs/backend-persistence.md` actually corrects the now-stale "closing that would need per-player event sequencing" sentence instead of appending next to it. That's the documentation rule working as intended.

Nothing here blocks. Address the two should-fixes and this is ready.

---

ginderjeremiah:
Took all four, plus both test gaps. Pushed as 9f50bd9.

**The two `required`-ness points collapsed into one change.** You raised making the `PlayerProgress` ctor arg required *and* converging the two aggregates onto `Player`'s `init` shape, and noted those pull slightly different ways — `init` is the nicer shape but is still omittable, since "only `PlayerCacheMapper` constructs it" is a property of today's call sites rather than of the type. Making the member `required` on **both** gets both halves at once: same shape, and omission is a compile error rather than a convention.

That immediately paid for itself. The compiler found a **fourth** `PlayerProgress` construction site — `Game.Core.Tests/Progress/ChallengeTests.cs:109` — that neither my `new PlayerProgress(` grep nor your three-call-site count caught, because it's a collection-expression `new(...)` with the type on the left. It's a test, so nothing was broken, but it's a construction site that would have silently seeded 0 forever, which is precisely your argument.

**The unparseable-`_seq` log.** Agreed on the severity correction — my comment said reseeding "only costs the guard its ordering information for this session", and that understates it: it restarts at 1 against a watermark that may already be at 12, so under #2474's `<= @seq` rule *every* write that session makes is rejected until it climbs back past. Now a `LogWarning` naming the player id and the raw value, and the comment says the real consequence rather than the mild one. `FromHashFields` went from `static` to an instance method taking `playerId` to reach the logger, matching `ToCoreChallenge` alongside it.

**`PlayerUpdateContext`** took the `Describe(envelope)` option — you're right that `{ get; private set; }` plus a public `SetSequence` is the same access surface as a plain settable property, and one populate method is what stays a single call when #2474 adds to it.

**`BeginPlayerSave`** — took it. The parameter is required and `BeginBatch` passes `DomainEventEnvelope.Unsequenced` explicitly, with the reason at that call site.

**On the `IWriteSequenced` question** — that settles it, and your reasoning is better than my framing of the question. Nothing polymorphically consumes "a write-sequenced aggregate", so the interface would buy naming at the cost of a `Game.Core` abstraction describing a data-tier concern. And the follow-on is right: the duplication that actually bit was the cross-referencing doc comments, and giving both aggregates the same shape is what fixes it — `PlayerProgress`'s doc is now genuinely short because the reader has already seen the shape on `Player`. Not creating a follow-up issue for it; if a third write-behind aggregate appears, that's the trigger to revisit.

**Test gaps.** Three added:

- `Load_CachedSequenceIsUnparseable_ReseedsFromZeroAndWarnsRatherThanDiscardingTheRows` — corrupts only the counter in the live hash, then asserts the row fields survive (unlike a corrupt *row*, which deletes the key), the warning is logged, and the next save stamps 1.
- The advance-on-empty-save asymmetry, pinned on both sides, since as you say it currently reads as a rule stated in two comments and is the kind of thing a later reader "fixes" into consistency: `SavePlayer_RaisingNoEvents_StillConsumesASequence` and `Save_NoDirtyRows_DoesNotConsumeASequence`. Each test's comment now names the *other* side and why they differ — the player counter lives on an aggregate held for the connection's lifetime, so a gap costs nothing, whereas a gap on the progress side would be reseeded from the cache on the next load and lost anyway.

Your three traced invariants match my reading, including the one I hadn't written down: a successful flush whose fire-and-forget blob write drops yields two envelopes at the same sequence, which #2474's `<=` applies rather than discards. Worth having in the thread.

Full backend suite green — 3,588 tests, 0 failures.

---

PR #2493, opened 2026-07-26
Add a UTF-8 BOM to build/*.ps1 so 5.1 host output isn't mojibake (#2479)

### LINKED ISSUE #2479: build/*.ps1 output is mojibake under Windows PowerShell 5.1 (BOM-less UTF-8 + non-ASCII)
Raised while reviewing PR #2478 (#2453), which flagged this and correctly left it alone as pre-existing and out of scope.

## Problem

Every `build/*.ps1` is saved as BOM-less UTF-8 but contains non-ASCII characters in host output — em-dashes and the box-drawing rules in `coverage.ps1`'s `Write-TestFailureDetail` header, and the `below current — keep existing, ratchet only rises` label now in `coverage-lib.ps1`.

Windows PowerShell 5.1 reads a BOM-less file as ANSI, not UTF-8, so those characters are decoded per the active code page and printed as mojibake. `pwsh` 7 defaults to UTF-8 and is unaffected.

That matters because 5.1 support is deliberate here, not incidental: every script declares `#Requires -Version 5.1` and `docs/infrastructure.md` advertises the scripts as running on Windows PowerShell 5.1 and `pwsh` 7 alike. So the garbled output lands on exactly the host the requirement exists to serve.

Cosmetic — no logic depends on these strings, and the coverage gate's exit codes are unaffected.

## Fix

Add a UTF-8 BOM to the `build/*.ps1` files (5.1 honours the BOM; 7 is indifferent). Alternatively, restrict host-output string literals to ASCII, which is the convention `build/coverage.test.ps1` already adopted for its assertion text.

Worth deciding once and noting in `docs/infrastructure.md` alongside the dual-host claim, so the next script added to `build/` doesn't reintroduce it.

## Scope

Small — an encoding change across five files plus a one-line doc note, no behaviour change.

ginderjeremiah:
Closes #2479.

Encoding-only change to the five `build/*.ps1` files, plus the two things that keep it from coming back. No script logic touched — every `.ps1` diff is one line, the BOM landing in front of `#Requires`.

## The fork, and why the BOM won

The issue offered two fixes; they are not equivalent in reach.

**Restricting host-output literals to ASCII only fixes part of the problem.** The non-ASCII in these files isn't mostly in `Write-Host` strings — it's 24 lines, and all but three are comments and **comment-based help**. `.SYNOPSIS`/`.DESCRIPTION` is host output too (`Get-Help` renders it), so an ASCII-only rule scoped to output literals leaves the help text garbled, and a rule broad enough to catch it means rewriting every em-dash in every comment across five files and permanently constraining what future prose can say.

**The BOM fixes all of it at once**, costs one byte per file, needs no prose churn, and is what Windows PowerShell 5.1 actually keys off — it decodes a BOM-less file with the active ANSI code page and a BOM'd one as UTF-8. `pwsh` 7 defaults to UTF-8 and is indifferent either way.

## Enforcement, because pwsh 7 hides this

The issue asks that the next script added to `build/` not reintroduce it. A doc line alone wouldn't do that, and the failure is unusually easy to miss: **CI runs these under `pwsh` 7, which decodes them correctly with or without the BOM**, so a dropped BOM stays invisible until someone runs them under 5.1. So:

- **`.editorconfig`** pins `charset = utf-8-bom` for `*.ps1`, so an editor adds and preserves it without anyone remembering to. Scoped to `*.ps1` rather than `build/*.ps1` — all five live in `build/` today, but the rule is about the language's 5.1 host, not the folder.
- **`coverage.test.ps1`** asserts the first three bytes of every `*.ps1` in its own folder. Enumerated rather than listed, so a newly added script is covered without editing the test — which is the case the issue was actually worried about.

## One stale comment removed

`coverage.test.ps1`'s help block documented the old state as a rationale — *"Assertion text is deliberately ASCII: these files are BOM-less UTF-8, which Windows PowerShell 5.1 reads as ANSI…"*. That premise is exactly what this PR removes, so the paragraph now describes the BOM contract and points at the cases that assert it. The existing assertion text is left ASCII regardless: nothing is gained by adding em-dashes to it.

## Test plan

`./build/coverage.test.ps1` — **85 passed, 0 failed** (up from 80; the 5 new cases are one per script). This is exactly what the `test-build-scripts` CI job runs. `dotnet build` is also clean: the `.editorconfig` addition is a `*.ps1` section and inert for C#.

Two verifications beyond the suite passing, since a green run proves less than usual here:

- **The new assertion can actually fail.** Stripped the BOM from a scratch copy of `coverage-lib.ps1` and re-ran: `FAIL: coverage-lib.ps1 starts with a UTF-8 BOM - expected 'EF BB BF', got '23 52 65'`, run exits 1. The failure names the actual leading bytes rather than just saying "false".
- **The bug reproduces and the fix clears it.** 5.1 isn't available in this environment, so I drove .NET's `StreamReader` with its exact pair — code page 1252 as the default encoding, BOM detection on — against the file before and after:
  ```
  BOM-less (before): Write-Host "â”€â”€ $($log.FullName) â”€â”€" -ForegroundColor Yellow
  with BOM  (after): Write-Host "── $($log.FullName) ──" -ForegroundColor Yellow
  ```
  That's `coverage.ps1`'s `Write-TestFailureDetail` header — the mojibake the issue describes, and its absence afterwards.

Also parse-checked all five files with `[Parser]::ParseFile` post-BOM (all clean), and confirmed `Get-Help` output is byte-identical before and after, so nothing about script loading or help discovery shifted.

## Note

No follow-ups worth filing — the encoding convention is now pinned in `.editorconfig`, asserted by a test, and recorded in `docs/infrastructure.md` alongside the dual-host claim the issue pointed at.

---
_Generated by [Claude Code](https://claude.ai/code/session_012FNz27Mzgvq96ZBtiyL55A)_


ginderjeremiah (commented):
## Review: UTF-8 BOM on `build/*.ps1` so Windows PowerShell 5.1 stops mojibaking host output (#2479)

**Verdict:** Approve (posted as a comment — GitHub won't take a gating verdict from the PR author) — no blocking, should-fix, or nit findings; the fork was decided on the right grounds and the enforcement is the part that makes it stick.

I verified the parts that would be quietly wrong if they were wrong:

- **All five files actually carry the BOM** (`EF BB BF` leading every one), and no `.ps1` exists outside `build/`, so the enumerated assertion covers the whole surface today.
- **Nothing strips it back out.** `.gitattributes`' `* text=auto eol=lf` normalises line endings only — the BOM survives the blob round-trip, so the assertion isn't going to start failing on a Windows checkout.
- **The fixtures are unaffected.** `New-Fixture` writes its `Summary.json`/floors via `WriteAllText` (BOM-less by design, comment intact at `build/coverage.test.ps1:140`), and the new contract is scoped to `*.ps1`, so the two encoding rules in this file don't collide.
- **The assertion is placed where it can't self-poison.** The `Get-ChildItem` enumeration runs at the top of the `try`, before any case creates anything, and the fixtures land in `$Work` under the temp path rather than in `$PSScriptRoot` — so a leftover fixture from a crashed run can't fail a later run's encoding cases.
- **Degenerate inputs still fail loudly rather than vacuously.** Piping `ReadAllBytes` through `Select-Object -First 3` yields nothing for an empty file, so `-join ' '` gives `''` and the case fails with the actual bytes named. The array-index alternative (`[0..2]`) would have thrown under `$ErrorActionPreference = 'Stop'` and taken the whole run with it; this shape is the more careful one.
- **No stale ASCII-convention text left anywhere.** The `.DESCRIPTION` paragraph that documented the old BOM-less premise is the only place that asserted it, and it's rewritten rather than appended to.

### On the fork

Agreed with the choice, and for the reason the body gives rather than the obvious one. The decisive fact isn't that the BOM is cheaper — it's that **24 of the non-ASCII lines are comments and comment-based help, not `Write-Host` literals**. `Get-Help` output is host output, so the ASCII-only alternative either leaves the help text garbled or turns into a permanent prose constraint across five files. That's a real asymmetry between the two options the issue presented as near-equivalent, and it's the right one to have led with.

### On the enforcement

This is the half the issue actually asked for ("so the next script added to `build/` doesn't reintroduce it"), and it's the half that's easy to under-deliver. Two things make it hold:

- Asserting over `Get-ChildItem` rather than a hardcoded list means a newly added script is covered **without anyone remembering to edit the test** — which is precisely the failure mode in question, since the person adding the script is the person who wouldn't think of it.
- Naming, in the comment, that **CI runs these under `pwsh` 7 and therefore cannot notice a dropped BOM** is the load-bearing context. Without it this test reads like a tautology ("assert the file is how we saved it") and is a prime candidate for deletion in a later cleanup. With it, the reason it exists survives contact with a future reader.

The `.editorconfig` pin is a sensible second layer — worth knowing that `charset` support varies across editors (Rider and the VS Code EditorConfig extension honour it; not every host does), which is exactly why the test rather than the config is the thing that can't be silently bypassed. The doc entry already attributes the guarantee to the test, so that stays accurate.

Scoping the `.editorconfig` section to `*.ps1` rather than `build/*.ps1` is also right: the constraint comes from the language's 5.1 host, not the folder, so a script added elsewhere inherits it. The test's folder scope is the narrower one, but that's the correct asymmetry — a config rule should describe the language, a test should assert what it can see.

### Test coverage

Sufficient, and the verification in the body goes past what a green run proves. Two things I'd single out:

- **The negative case was actually exercised** — stripping the BOM from a scratch copy and confirming `FAIL: coverage-lib.ps1 starts with a UTF-8 BOM - expected 'EF BB BF', got '23 52 65'` with a non-zero exit. An always-passing assertion is the default failure mode for a test like this, and it was checked rather than assumed.
- **The bug itself was reproduced**, by driving `StreamReader` with 5.1's exact pair (cp1252 default + BOM detection) before and after, rather than reasoning about what 5.1 would do with a host that isn't available here. `Write-TestFailureDetail`'s header garbled before, clean after — that's the issue's symptom, demonstrated on both sides of the fix.

### What's good

- The one-line-per-script diffs. An encoding change that touched script logic would be much harder to trust; five identical one-line patches are reviewable at a glance.
- Rewriting the now-false `.DESCRIPTION` premise instead of appending the new state next to it. That paragraph justified ASCII assertion text *because* the files were BOM-less — leaving it would have left the file arguing against its own contract. Keeping the assertion text ASCII anyway is the right call: nothing is gained by proving the point inside the strings.
- The doc entry records the decision **and the rejected alternative with its reason**, which is what stops #2479 from being re-litigated the next time someone notices an em-dash in a `Write-Host`.

No findings, no follow-ups worth filing. CI's `test-build-scripts` — the job that runs this suite — is green.

---

Issue #2475, opened 2026-07-26
Retire the cross-pass parked-lane deferral once the write sequence guard lands


ginderjeremiah:
Implementation step 3 of [Spike #2467](https://github.com/ginderjeremiah/GameServer/issues/2467). **Depends on #2474** — do not land this before the guard exists.

## Why

`DataProviderSynchronizer._parkedPlayerLanes` (added by #2460) exists for exactly one reason: to stop a newer same-player event applying ahead of an older one this instance left parked on the processing list, because the reclaim's later replay of the parked event would regress the newer absolute values. Once #2474's sequence guard rejects that replay on its own merits, the deferral no longer protects anything — it is pure convergence latency for the affected player (bounded by the reclaim grace period plus the reconciliation sweep), and leaving it in place means two mechanisms guarding one invariant with only one of them load-bearing.

## Scope

- Remove `_parkedPlayerLanes`, the deferral branch in `DrainQueueAsync`, its `deferredPlayers` log-throttling set, and the `.Clear()` calls in `TryReclaimStrandedProcessingAsync`. The `catch` in `ProcessReservedItemAsync` still needs to let the fault surface — only the lane bookkeeping goes.
- **Keep the pass-scoped `playerLanes` map.** It still preserves same-player ordering cheaply within a pass, which avoids doing work only to have the guard reject it. Also keep `CompactCompletedItems`' rule about not evicting a faulted lane head, which serves that same within-pass purpose.
- Update `docs/backend-persistence.md`: the "That block outlives the pass that created it" paragraph goes away, and the cross-instance best-effort framing above it becomes "closed by the sequence guard" rather than "open, tracked as #2467".
- Remove the now-dead `#2460` deferral tests and confirm the ordering tests that remain still pin real behaviour rather than passing vacuously.

## Verification

The point of this issue is that removing the deferral must not reopen #2460. The within-instance interleaving test (`DataProviderSynchronizerTests`, the parked-player scenario around line 1945) should therefore be **kept and re-pointed at the guard** — the parked event still gets reclaimed and replayed, and the assertion that the DB holds the newer state must still hold, now because the replay is rejected rather than because it was deferred.

ginderjeremiah:
Flagging a dependency change while working #2474: this issue's premise — "once the write sequence guard lands" — now needs a more specific trigger than #2474 alone.

[...]

PR #2476, opened 2026-07-26
Spike: decide per-player write sequencing for the write-behind drain (#2467)

ginderjeremiah:
Closes #2467.

Docs only — no production code changes. #2467 is a spike, and its acceptance criteria are a documented decision on the design forks plus follow-up implementation issues, and a correction to the best-effort-ordering claim in `docs/backend-persistence.md`. Both are here.

## What's in it

- **`docs/spikes/2467-cross-instance-write-behind-sequencing.md`** — the research and a decision on each fork the issue raised.
- **`docs/backend-persistence.md`** — the convergence argument now enumerates which reorderings actually converge and which don't, and the cross-pass parked-lane paragraph records why a fleet-shared version of that block isn't the fix.
- **Follow-up issues** #2473 → #2474 → #2475, linked as sub-issues of #2467, in dependency order.

## Three things that didn't survive verification

I checked the issue's premises against the code rather than taking them as given, and three changed the shape of the recommendation:

1. **The list of five affected handlers is incomplete.** Only the four pure insert-if-missing handlers (`SkillUnlocked`, `ItemUnlocked`, `ModUnlocked`, `LessonUnlocked`) are genuinely convergent. Nine others write absolute state — the spike doc has the full table. `AttributeAllocationsChangedHandler` is the notable omission: a stale apply reverts spent stat points while `Player.StatPointsUsed` may hold the newer value, leaving the two disagreeing.

2. **The doc doesn't just understate unequip — it asserts the opposite.** *"The exception is unequip, where 'missing row' already is the desired end state … so its no-op converges"* is true for the reordered-behind-its-unlock case it was written about, but reads as a general order-safety claim. A stale `ItemUnequippedEvent` after a newer `ItemEquippedEvent` durably unequips a worn item.

3. **Making `_parkedPlayerLanes` fleet-shared doesn't work**, which is worth recording because it's the first thing anyone will reach for. It fails on timing, not cost: reservation is sequential and the park only happens once the retry budget is exhausted (~0.6s), so instance B reserves *and applies* the newer event before instance A ever sets the marker. Closing that window would need the marker claimed atomically at reserve time — a fleet-wide per-player in-flight lock, which is exactly the same-player cross-fleet serialization #1701 relaxed.

## The decision, briefly

[...]

PR #2494, opened 2026-07-26
Add PlayerWriteWatermark and guard the core/progress write-behind handlers (#2474)

### LINKED ISSUE #2474: Add PlayerWriteWatermark and reject stale absolute writes in the write-behind handlers

ginderjeremiah:
Closes #2474.

Implementation step 2 of spike #2467 — the consuming half of the write sequence #2473 put on the envelope. This is **part (a)** of the split the issue itself proposes: the table, the context plumbing, the conditional-update helper, the rejection accounting, plus `PlayerCoreUpdated` and `ProgressUpdated`. Equipment/mods (b) and the remaining single-row handlers (c) are filed as follow-ups.

## The hazard being closed

An event parked on the shared processing list by instance A, reclaimed after instance B already applied the player's next save, wins simply by being applied last — durably regressing level/exp/zone or a statistic while Redis still holds the correct state, with nothing self-correcting until the player dirties the same row again.

## What's in it

- **`PlayerWriteWatermark(PlayerId, Stream, TargetKey, LastAppliedSequence)`** + a `PlayerWriteStream` enum (one value per guarded handler) and its migration.
- **`PlayerWriteWatermarkGuard`** — the compare *is* the write. One `INSERT … ON CONFLICT DO UPDATE … WHERE "LastAppliedSequence" <= EXCLUDED` over a sorted, deduplicated key array, `RETURNING` the targets it actually advanced, with the data write in the **same explicit transaction**. The watermark row is therefore the per-target serialization point, not a value two instances can both read and pass.
- **`PlayerCoreUpdated`** (one player-scoped target) and **`ProgressUpdated`** (one target per statistic/challenge/proficiency row) routed through it.
- **Rejection accounting** — `PlayerUpdateContext.RejectedTargetCount`, totalled per drain pass by `DataProviderSynchronizer` and surfaced as one summary line, throttled the way `_infrastructureOutageLogged` is (first pass of a run at Warning, the rest at Debug, cleared by a clean pass).

## Design points worth a second look

**Rejection is per target, not per event.** The guard hands the callback only the accepted keys, and the handler writes only those. All-or-nothing would let a newer `ProgressUpdated` covering statistic Y discard an older event's entirely different, still-current statistic X — progress events carry only a save's *dirty* rows, so that would silently lose writes on the game's highest-volume path. A handler that genuinely needs all-or-nothing (part (b)'s item+slot pair) gets it by throwing from the callback, which rolls the transaction back with neither watermark advanced.

**The transaction is load-bearing, not framing.** Split across commits, a crash between the advance and the data write advances the watermark without the data, and the redelivery is then rejected as stale — a silently lost write, strictly worse than the bug being fixed. There's an integration test that injects exactly that failure.

**`<=`, not `<`.** Equal sequences are a save's sibling writes landing on one target, and the insert-if-missing seed would immediately no-op under `<`. The cost is one redundant no-op write on an exact duplicate — the right trade under at-least-once.

**Sequence 0 bypasses the guard entirely** and leaves watermarks untouched, rather than losing every comparison and silently discarding a whole session's writes for a player who reconnects onto a pre-upgrade instance mid-rolling-deploy.

**`ProgressUpdatedHandler`'s bespoke unique-violation retry moved into the guard**, which restarts the *whole* attempt — watermark advance included — since a `DbUpdateException` aborts the surrounding transaction and re-running just the apply inside a transaction that can no longer commit would not work. Re-reading the watermarks on the retry is deliberate: another instance may have advanced them while the attempt was unwinding.

## Two things I decided against, and one you should know about

- **No FK from `PlayerWriteWatermarks` to `Players`.** The watermark must outlive the rows it guards (that is the whole reason it beat per-row version columns — `ModRemovedHandler`'s tombstone in part (b)), and an FK would turn an apply for an absent player, today a silent zero-row no-op, into a violation the drain dead-letters. The consequence is that nothing cascades into the table, so it needed adding to `DatabaseCleaner` explicitly — `RESTART IDENTITY` reuses player ids, and a leaked watermark would have rejected the next test's writes as stale.
- **`#2475` is not yet actionable.** It's written as "retire the cross-pass parked-lane deferral once the write sequence guard lands", but the deferral still covers the nine handlers this PR doesn't guard. It only becomes safe to retire after (b) and (c).

## Test plan

- **`PlayerWriteWatermarkIntegrationTests`** (new): first-ever write seeds rather than rejects; older-sequence skipped with the watermark left alone; equal-sequence applies; newer applies and advances; sequence-0 applies against a watermark well above 0 and leaves it unadvanced; an older progress event's *different* statistic still applies while only the superseded one is skipped; challenge/proficiency targets guarded independently; the two streams are independent spaces; and a fault injected between the watermark advance and the data write leaves both unapplied, with the redelivery still landing the write.
- **`DataProviderSynchronizerTests`** (extended): the #2467 interleaving itself, driven through **two** `DataProviderSynchronizer` instances over one shared queue — instance A parks seq 1, instance B applies seq 2 with no knowledge of the park, A reclaims and replays seq 1 — asserting the DB holds the newer state and that the rejection is surfaced rather than silent. `Serialize` gained an optional `sequence` defaulting to the unsequenced sentinel, so every existing scenario exercises the handlers exactly as before.
- Full solution suite green: **3,598 passed, 0 failed** (`Game.Core.Tests` 1492, `Game.Infrastructure.Tests` 74, `Game.Application.Tests` 1139, `Game.Api.Tests` 893).

No frontend changes — this is backend persistence with no battle logic, so nothing needed mirroring in the UI suite.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fp1K42CNRN1iSf7sVtWRZ2)_


Game.DataAccess/PlayerUpdates/PlayerWriteWatermarkGuard.cs
```diff
@@ -0,0 +1,151 @@
...
+        /// </para>
+        /// </summary>
+        /// <remarks>
+        /// Takes no <c>CancellationToken</c> by design — see <see cref="IPlayerUpdateHandler{TEvent}"/> (#1029).
+        /// </remarks>
+        public async Task ExecuteGuardedAsync(
+            int playerId,
+            PlayerWriteStream stream,
+            IReadOnlyCollection<string> targetKeys,
+            Func<IReadOnlySet<string>, Task> applyAsync)
```
ginderjeremiah: **Should-fix:** the transaction only covers the callback's write because the handler happens to resolve the *same* scoped `GameContext` the guard did. That's the load-bearing invariant of this whole design — "split into separate commits … a silently lost write, strictly worse than the bug being fixed" — and nothing enforces it. A handler that resolved a second context (an `IDbContextFactory`, a `using var context = ...`, a repository holding its own) would write on a different connection, outside this transaction, and the failure mode is exactly the one the guard exists to prevent: watermark advanced, data not, redelivery rejected as stale. It fails silently, with green tests.

With nine more handlers joining this pattern in (b) and (c), I'd make it structural rather than documented — hand the callback the context it must write through:

```csharp
Func<GameContext, IReadOnlySet<string>, Task> applyAsync
```

so `ProgressUpdatedHandler` writes through the parameter and stops taking `GameContext` in its own constructor. That turns "must use the guard's context" from a rule someone has to know into the only context in scope. If you'd rather keep the current signature, the XML doc above needs to state the requirement explicitly — right now it's not mentioned anywhere.

---

Game.DataAccess/PlayerUpdates/PlayerWriteWatermarkGuard.cs
```diff
@@ -0,0 +1,151 @@
...
+                // and its save. The failed save also aborted the surrounding transaction, so the whole attempt
+                // restarts — watermark advance included — rather than just the apply re-running inside a
+                // transaction that can no longer commit. The now-existing rows load as updates on the second
+                // pass, so it carries no conflicting insert; a second failure propagates to the queue's retry
+                // policy rather than looping here. Re-reading the watermarks is deliberate: another instance
+                // may have advanced them while this attempt was unwinding, and a cached accepted-set would
+                // apply against a decision that is no longer true.
+                context.ChangeTracker.Clear();
+                await AttemptAsync(playerId, stream, targetKeys, applyAsync);
+            }
```
ginderjeremiah, line 83: **Should-fix (test gap):** this restart is the subtlest thing in the guard — a `DbUpdateException` aborts the surrounding transaction, so the whole attempt including the watermark advance has to re-run, and the watermarks are deliberately re-read — and nothing exercises it at a non-zero sequence.

`ProgressUpdated_AppliedConcurrently_InsertsOneRowPerKeyWithoutThrowing` does hit this catch, but it applies unsequenced, so `AttemptAsync` takes the bypass at line 95 and never opens a transaction. The retry that's actually covered is the pre-guard one; the transactional retry — the one where a stale accepted-set or a reused aborted transaction would bite — is untested.

Worth an integration test in `PlayerWriteWatermarkIntegrationTests`: a sequenced progress apply where a row is inserted concurrently between the load and the save, asserting the second attempt commits *and* the watermark ends at the event's sequence (not double-advanced, not left behind).

---

Game.DataAccess/PlayerUpdates/PlayerWriteWatermarkGuard.cs
```diff
@@ -0,0 +1,151 @@
...
+        private async Task AttemptAsync(
+            int playerId,
+            PlayerWriteStream stream,
+            IReadOnlyCollection<string> targetKeys,
+            Func<IReadOnlySet<string>, Task> applyAsync)
+        {
+            // Distinct because ON CONFLICT DO UPDATE cannot affect the same row twice in one statement, and
+            // ordered so two events sharing a pair of targets take their row locks in the same order rather
+            // than deadlocking against each other.
+            var keys = targetKeys.Distinct(StringComparer.Ordinal).Order(StringComparer.Ordinal).ToArray();
```
ginderjeremiah: **Nit:** the comment credits the C# `.Order(...)` with the deadlock avoidance, but it's the SQL `ORDER BY "Key"` at line 139 that actually governs lock acquisition order — and the two orderings aren't even guaranteed to agree, since `ORDER BY` uses the database collation while this is `StringComparer.Ordinal`.

That's fine as-is (both concurrent statements sort by the same collation, so they still agree with *each other*), but the comment sets up the wrong mental model: a later reader could drop the SQL `ORDER BY` as redundant on the grounds that the keys arrive pre-sorted, and quietly reintroduce the deadlock. Suggest saying the `ORDER BY` is the load-bearing one and this sort just makes the parameter array deterministic — line 127's comment is closer to right and the two read as contradicting each other.

---

Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs
```diff
@@ -2140,12 +2196,16 @@ private static async Task<List<string>> DrainDeadLetterQueue(IPubSubService pubs
             return drained;
         }
 
-        private static string Serialize<T>(T evt) where T : IDomainEvent
+        // Sequence defaults to the "unsequenced" sentinel, which is what a pre-upgrade instance's envelope
+        // carries and what makes the watermark guard (#2474) stand aside — so the scenarios below that aren't
+        // about ordering exercise the handlers exactly as they behaved before the guard existed.
+        private static string Serialize<T>(T evt, long sequence = DomainEventEnvelope.Unsequenced) where T : IDomainEvent
```
ginderjeremiah: **Should-fix:** defaulting to the sentinel means the entire pre-existing synchronizer suite now runs the *bypass* branch — no transaction, no watermark, no upsert. The comment frames that as a feature ("exercise the handlers exactly as they behaved before the guard existed"), but it inverts the coverage: the path every production envelope will take from here on is exercised by exactly one test, while ~40 scenarios cover a branch that only exists for a transient rolling-deploy window.

Defaulting to `1` instead should be a no-op for every existing scenario — the predicate accepts equal sequences, so any number of same-player events at the same sequence all apply — and it puts the whole suite through the guarded path, transaction included. The sentinel scenarios (and the new interleaving test) then pass their sequence explicitly, which also reads better: `Serialize(evt, sequence: DomainEventEnvelope.Unsequenced)` says "this one is about the pre-upgrade envelope" at the call site rather than by omission.

---

docs/backend-persistence.md
```diff
@@ -71,7 +75,7 @@ Queue processing never silently drops an event, and the read itself is **at-leas
...
+**That block outlives the pass that created it.** The lane map itself is pass-scoped, so a player whose event is left parked on the processing list (a DB-infrastructure outage, or an acknowledge that faulted after a durable apply) is remembered across passes: while it is, that player's newly reserved items are *deferred* — left reserved and unapplied — rather than applying on a fresh lane. Without that, a post-recovery event would apply in the next pass and then be regressed when the reclaim finally replays the older parked event, a durable loss for the absolute-value handlers (#2460). The block is released once a pass observes the processing list empty or a reclaim has drained it back onto the queue head in order, so the deferral costs that one player convergence latency (bounded by the grace period plus the reconciliation sweep), never the write itself. The memory is instance-local, so this hardens the within-instance ordering guarantee; the cross-instance half is closed by the watermark guard below instead.
```
ginderjeremiah: **Nit:** two things on this sentence.

"the watermark guard **below**" — the guard paragraph is at line 29, above this one. Should be "above".

And "the cross-instance half is closed" reads as done, when two of eleven handlers are guarded. Line 31 does spell that out, but it's ~45 lines away and this is the sentence someone lands on when reading about the parked-lane block. Suggest "…is closed by the watermark guard above, for the handlers it covers so far."

---

ginderjeremiah (commented):
## Review: `PlayerWriteWatermark` + the guarded core/progress write-behind handlers (#2474, part (a))

**Intended verdict: Approve** — no blocking findings. (Submitted as a comment review because GitHub won't accept a gating verdict from the PR's own author.) Three should-fixes and two nits below; none of them are correctness holes in what's here, but the first one is a footgun that gets nine times more expensive once (b) and (c) land, so I'd rather it be fixed in this PR than inherited.

I verified the mechanic rather than taking the description's word for it: `ON CONFLICT DO UPDATE` locks the conflicting row *before* evaluating its `WHERE`, so a rejected target still serializes against a concurrent apply rather than passing through — the "the compare is the write" claim holds, and `RETURNING` correctly omits the rows whose predicate failed. `<=` is right for the reasons stated. No `EnableRetryOnFailure` is configured (`GameContextFactory.cs:34`), so `BeginTransactionAsync` doesn't trip the execution-strategy restriction that would otherwise make every guarded apply throw — worth knowing that this design is now quietly coupled to that.

### Should-fix

- **The transaction covers the apply only by DI coincidence** — `PlayerWriteWatermarkGuard.cs:52`. The guard and the handler share a `GameContext` because both are scoped in the same drain scope, and nothing states or enforces that. A future handler resolving its own context writes outside the transaction, and the failure is the silently-lost-write the guard exists to prevent. Suggest passing the context to the callback so it's structural. Details inline.
- **The transactional unique-violation restart is untested** — `PlayerWriteWatermarkGuard.cs:62`. The existing concurrency test reaches that catch only on the unsequenced bypass, where no transaction is open. Details inline.
- **`Serialize`'s sequence defaults to the sentinel** — `DataProviderSynchronizerTests.cs:2202`, which routes the whole pre-existing synchronizer suite through the bypass branch instead of the guarded one. Defaulting to `1` should be a no-op for those scenarios and inverts the coverage back the right way. Details inline.

### Nits

- The C#-side key sort is credited with the deadlock avoidance the SQL `ORDER BY` actually provides — `PlayerWriteWatermarkGuard.cs:83`, and it reads as contradicting the comment at line 127.
- `docs/backend-persistence.md:78` says "below" for a paragraph that's above, and states the cross-instance half as closed when 9 of 11 handlers are still unguarded.
- Minor asymmetry: the bypass invokes `applyAsync` with an empty accepted set when an event has no targets, while the guarded path skips the callback entirely when nothing is accepted (`PlayerWriteWatermarkGuard.cs:95` vs `:104`). Both handlers here tolerate it; worth a word in the doc comment so a future handler doesn't put unconditional work in the callback and get inconsistent behaviour.

### Test coverage

Good, and pointed at the right things. The eight watermark integration tests cover every fork the issue named for part (a) — seed-on-first-write, older/equal/newer, sequence 0 against an advanced watermark, per-row granularity on the progress path (the case a per-player watermark would have broken), stream independence, and the fault injected between the advance and the data write. The two-`DataProviderSynchronizer` interleaving test is the right shape for the #2467 regression: driving instance B over the same queue is what makes it a genuine cross-instance test rather than a re-park, and asserting on the surfaced Warning keeps the "not silently" half honest.

Gaps are the two above (transactional retry; the suite defaulting to unsequenced). Nothing needed mirroring in the frontend — no battle logic here.

### Follow-ups worth filing (not for this PR)

- **Round-trip cost on the hottest write path.** Every guarded apply now adds `BEGIN` + the upsert + `COMMIT` to what was one round trip for `PlayerCoreUpdated`. `docs/backend-persistence.md` explicitly treats "one round-trip per event" as the thing #1701's concurrency work was relieving, and this walks part of that back on the progress path without acknowledging it anywhere. I don't think there's a cheaper design that keeps the guarantee, but it should be measured and written down rather than discovered later.
- **`PlayerWriteWatermarks` growth and retention.** One row per guarded target per player, forever — roughly a 1:1 shadow of `PlayerStatistics` (hundreds of rows for a long-lived account) once (b)/(c) add equipment and mods. No pruning story, and the deliberate absence of an FK means that if account deletion is ever added, watermarks won't cascade and will need explicit cleanup. The FK decision is right for the tombstone reason given; the consequence just needs an owner.

### What's good

The granularity argument is the part that matters most and it's correct: per-target rejection with the accepted set handed to the callback is what keeps an older progress event's still-current statistic from being collateral damage, and routing all-or-nothing through a throw from the callback (rather than a second mode on the guard) leaves (b) a clean seam without widening the API. Sequence 0 bypassing rather than losing every comparison is the right call and the rolling-deploy reasoning behind it is sound. Folding `ProgressUpdatedHandler`'s bespoke retry into the guard is a real simplification, not just a move — restarting the whole attempt is the only thing that works once a transaction is in play, and the note about re-reading the watermarks on retry shows the reasoning was actually done. Adding `PlayerWriteWatermarks` to `DatabaseCleaner` with the id-reuse rationale is the kind of thing that's invisible until it costs someone a day.

---

ginderjeremiah:
All five points addressed in 1f38500. Full suite green: **3,599 passed, 0 failed** (+1 for the new test).

**Passing the context to the callback** — done, and you were right that it's worth doing now rather than inheriting. `ExecuteGuardedAsync` takes `Func<GameContext, IReadOnlySet<string>, Task>`, and both handlers dropped `GameContext` from their constructors, so the guard's context is the only one in scope at the call site.

**The transactional restart test** — worth recording *why* it was awkward to reach, because it turned out to be a property of the design rather than a gap in the test: the watermark row serializes same-target sequenced applies, so two of them can no longer interleave between load and save at all. The guard removed that race. What survives is a sequenced apply racing an **unsequenced** one, which bypasses the guard and so isn't serialized by the watermark — which is also the honest scenario, since it's the rolling-deploy window. The new test forces that deterministically (insert from a second scope inside the callback on attempt one) and asserts the restart commits with the watermark at the event's sequence, neither left behind nor double-advanced.

**`Serialize` defaulting to `1`** — agreed, and it was a clean no-op as predicted: every existing scenario passes unchanged, now through the guarded path with the transaction and upsert. Applied to `SerializeProgress` too.

Both nits fixed, plus the asymmetry one — the XML doc now states the callback isn't invoked when every target is rejected.

Follow-ups filed: #2497 (round-trip cost, including the implicit coupling to `EnableRetryOnFailure` being unset — good catch, that would turn every guarded apply into a throw if a retrying execution strategy is ever enabled) and #2498 (growth/retention and the no-cascade consequence of the deliberate missing FK).

---

ginderjeremiah:
## Re-review of `1f38500` — one blocking finding

The five points from the `da40a06` review are all in and all done well; details at the bottom. But re-reading the whole change end-to-end rather than just the delta surfaced a lost-write path that this PR opens, and it isn't a rare interleaving — it's the ordinary returning-player flow.

### Blocking — the producer counter still reseeds from 0 on a cold DB load, and the guard now compares it

#2473 deliberately deferred seeding the counter from persisted watermarks *because nothing consumed the stamp yet*. Both places say so in as many words:

- `Game.DataAccess/Repositories/PlayerQueryExtensions.cs:17` — *"`WriteSequence` is the deliberate exception — not `required` and not projected, because no column backs it yet: a cold database load seeds the counter from 0 and the next save stamps 1 (#2473). Seeding it from the player's highest persisted watermark arrives with the table that holds them (#2474)."*
- `Game.DataAccess/Repositories/PlayerProgressRepository.cs:272` — *"The write sequence is left at its Unsequenced default … Seeding it from the player's highest persisted watermark arrives with the table that holds them (#2474) — **safe to defer, because nothing consumes the stamp until that same issue lands**."*

This PR is that issue, and it consumes the stamp. Nothing in the diff touches either seed — `SelectPlayerCacheModel` still doesn't project `WriteSequence`, `LoadFromDb` still leaves `CachedPlayerProgress.WriteSequence` at `Unsequenced`, and `LastAppliedSequence` is read nowhere outside the guard's own upsert. So the deferral's stated precondition is removed without the deferred work landing.

**What that produces:**

1. A player plays. Each save advances `Player.WriteSequence` (`PlayerRepository.cs:127`), and each applied `PlayerCoreUpdated` advances the `(player, PlayerCore, "")` watermark. Their progress row watermarks advance the same way. Say the player's watermarks end the session at **N**.
2. The player goes dormant. `PlayerCacheTtl` and `ProgressCacheTtl` are both `AuthConstants.RefreshTokenLifetime`, and the keys lapse. This is designed, ordinary behaviour — `PlayerRepository.cs:26` calls it *"one transparent DB reload on their next access."*
3. They come back. Cold load → `WriteSequence` = 0 → their first save stamps **1**.
4. Guard: `N <= 1` is false, `RETURNING` comes back empty, `accepted.Count == 0`, the callback is never invoked, and the event is acknowledged as a successful no-op.

Every guarded write for that player — level, exp, zone, stat points, and every progress row they'd already touched — is skipped until the counter climbs back past N. That takes **as many saves as the old high-water mark**, which in an idle game with continuous battle saves is not a handful. Redis still holds the correct state throughout, so nothing looks wrong to the player or in the data they see; the loss only materializes at the *next* cache lapse, which drops the entire intervening session. And it compounds: a player whose sessions are shorter than their catch-up distance and whose gaps exceed the TTL may never durably persist again.

This is the failure mode this design keeps correctly calling strictly worse than the bug being fixed. The per-pass Warning from §3 means it isn't literally silent, but a returning player generating a steady stream of "stale write target skipped" is indistinguishable from the reordering storm the log exists to report.

**Fix — what the spike doc already specifies**, `docs/spikes/2467-cross-instance-write-behind-sequencing.md:293`: *"On a **cold DB load** (cache miss) it is seeded from the player's highest `LastAppliedSequence`."* That's the missing half. Three things to settle while doing it:

- **Scope the `MAX` to the streams the aggregate produces** — `PlayerCore` (plus the equipment/mod/etc. streams (b) and (c) add) for `Player`, `Progress` for `PlayerProgress`. A single unscoped per-player `MAX` is still safe (monotonic, it only skips values), but it makes each counter jump on the other's traffic, which contradicts the "deliberately separate spaces" claim on both `Player.WriteSequence` and `PlayerProgress.WriteSequence`.
- **Where to read it.** A correlated `COALESCE(MAX(...), 0)` in `SelectPlayerCacheModel` keeps that class's "the projection cannot silently omit one, so completeness is structural" property, and lets `WriteSequence` become `required` like everything else there. A second query on the cache-miss path is fine too — `LoadFromDb` already does one round trip, so it'd be a second. Either way `PlayerProgressRepository.LoadFromDb` needs the same.
- **Three comments go stale with it**: the two quoted above, and `docs/backend-persistence.md`'s *"a cold load with nothing to seed from starts at 0 and the next save stamps 1"* — which reads as a harmless footnote today and is the bug once the guard is live.

**The test that would have caught it**, and that I'd want before this lands: cold-load an aggregate whose watermarks already sit well above 0, and assert its next save's write actually reaches Postgres (and that the seeded counter is above the watermark). Every existing watermark test constructs its sequence explicitly via `DescribeSequence`, so none of them exercise what a real producer stamps after a cache miss — the guard is tested thoroughly, but the producer/consumer pair isn't tested end to end across a cold load.

Worth noting the gap isn't originally yours: **#2474's body never mentions the seed either**, so it traces back through the issue to the spike doc's implementation table. I've raised that on #2476 so the artifact gets corrected rather than just this PR.

### The `da40a06` follow-ups — all five addressed

- **Context passed to the callback.** `Func<GameContext, IReadOnlySet<string>, Task>`, and both handlers dropped `GameContext` from their constructors so the guard's context is genuinely the only one at the call site. That's the structural version, not the documented-rule version.
- **The transactional restart test.** The explanation for why it was awkward to reach is right and worth having on the record: same-target *sequenced* applies now queue on the watermark row, so they can't interleave between load and save at all — the guard removed that race. What survives is sequenced racing *unsequenced* (which bypasses the guard and so isn't serialized by the watermark), which is also the honest scenario since it's the rolling-deploy window. Forcing it from a second scope inside the callback on attempt one is deterministic rather than raced, and asserting the watermark lands at exactly 4 pins both failure directions — rolled back and left behind, or advanced twice.
- **`Serialize` defaulting to `1`.** Right call, and the named `DefaultSequence` constant reads better than the bare literal. Applied to `SerializeProgress` too, so the progress scenarios stopped bypassing as well. The whole synchronizer suite now runs the transaction-and-upsert path a production envelope takes.
- **Both nits plus the asymmetry one.** The sort comment now credits the SQL `ORDER BY` with the lock ordering and the C# sort with only parameter determinism, which is what's actually true. The XML doc states the callback isn't invoked when every target is rejected. The `backend-persistence.md` paragraph is rewritten in place and now names all nine still-unguarded handlers explicitly rather than implying the cross-instance half is closed.
- **#2497 / #2498 filed** for the round-trip cost (including the `EnableRetryOnFailure` coupling) and the growth/retention story.

### Verdict

**Request changes** — the cold-load seed. It belongs in this PR rather than a follow-up: it isn't incremental hardening, it's the other half of the mechanism this PR turns on, and without it merging makes `main` lose writes for every returning player. Everything else here is ready, CI is green (3,599 passing, all e2e browsers green), and the delta is a clean set of improvements on the last round.

Not merging.

---

Issue #2482, opened 2026-07-26
Extract the shared contract-drift rule duplicated across the test fixture modules


ginderjeremiah:
Raised while reviewing #2480 (which closes #2456).

## What's duplicated

Each module under `UI/src/tests/fixtures/` opens with a header comment, and those headers mix two
different kinds of information:

1. **Per-contract specifics** — which suites are excluded from the conversion and why, which adjacent
   builders deliberately stay separate. This is real information a reader needs, and it belongs where
   it is.
2. **The general contract-drift rule** — which is not contract-specific at all, and is now on its
   fifth near-verbatim copy.

The general half says, in nearly the same words in `items.ts`, `skills.ts`, `attributes.ts`,
`challenges.ts`, and `classes.ts`:

- only a suite that constructs a *typed* `X` pays the contract-drift tax;
- untyped `staticData` mocks (`{} as any`) never see a new required field, and filling their
  previously-`undefined` fields could change what a suite renders;
- so a `grep` still finding contract-shaped literals under `src/tests/` is expected, not a gap;
- a literal asserted `as X` is the dangerous form — it claims the contract to its consumer while
  silencing the missing field, so it looks covered and isn't;
- "reintroducing such a cast re-opens the hole, so state why if you do" appears close to verbatim in
  at least four of the modules.

The fixture series (#2414 / #2425 / #2433 / #2434 / #2444 / #2446 / #2456) exists to remove exactly
this class of duplication from the builders. It's worth finishing the job on the prose, which the
series has been accumulating a copy of per module.

## Suggested direction

Write the rule down once and let each module keep only its own exclusion table:

- The natural home is the testing section of `docs/frontend.md`, next to the existing fixture
  convention line (`docs/frontend.md:25` already states the builder rule — "one module per contract
  exporting a `make<Contract>(overrides)` with neutral defaults … state only the divergence"). The
  drift-tax rule and the `as X` warning are the same kind of guidance and belong beside it.
- Alternatively a short `UI/src/tests/fixtures/README.md`, if keeping it adjacent to the code reads
  better than a doc reference. Either is fine; one place, not five.
- Then trim each module's header to the part that's actually about that contract. Keep the
  cross-references that carry real information (e.g. `skills.ts` explaining why
  `battle-sim-test-utils`' `registerSkill` and `fight-fixtures`' `makeSkill` stay separate,
  `classes.ts` explaining that `ICreatableClass` and `IClass` are independently-drifting contracts).

[...]

PR #2499, opened 2026-07-26
Consolidate the workbench EntityConfig test factories onto a shared builder (#2481)

### LINKED ISSUE #2481: Consolidate the duplicated workbench EntityConfig test factory onto a shared builder
Split out of #2456 while converting the challenge suites onto `fixtures/challenges.ts`.

## Problem

Six suites under `src/tests/routes/admin/workbench/` each hand-roll a near-identical `config()` factory to feed an `EntityStore`:

```ts
const config = (): EntityConfig<Identified> =>
	({
		key: 'challenges',
		label: 'Challenges',
		singular: 'Challenge',
		glyph: 'box',
		blankName: 'Unnamed',
		newItem: (id: number) => ({ id }),
		meta: () => [],
		sections: [],
		refresh: async () => [],
		persist: async () => []
	}) as unknown as EntityConfig<Identified>;
```

Sites:

| Suite |
|---|
| `routes/admin/workbench/SectionRenderer.test.ts` |
| `routes/admin/workbench/WorkbenchDetail.test.ts` |
| `routes/admin/workbench/challenge/ChallengeConditionSection.test.ts` |
| `routes/admin/workbench/challenge/ChallengeRewardSection.test.ts` |
| `routes/admin/workbench/challenge/GoalField.test.ts` |
| `routes/admin/workbench/challenge/ScopeField.test.ts` |

This is the same class of duplication the `fixtures/` series exists to remove, with the same consequence: the `as unknown as EntityConfig<Identified>` assertion means none of these six sites sees a new required field on `EntityConfig`, so the contract has no test-side drift detection.

It is a **different contract** from the ones #2456 covered, and the double cast means it needs a slightly different treatment than a plain `make<Contract>` builder — hence a separate issue rather than widening that PR.

## Suggested direction

- Add a shared factory under `src/tests/fixtures/` (e.g. `workbench-config.ts` exporting `makeEntityConfig(overrides)`), defaulting the display fields and the no-op `meta`/`sections`/`refresh`/`persist` members.
- Try to land it **without** the `as unknown as` — the cast exists because the literal omits members of the real `EntityConfig<T>`, so check first whether supplying genuine no-op values for the missing members is enough to drop it. If some member genuinely can't be satisfied in a test (a Svelte component reference, say), keep the cast **in the fixture only** and state why there, so six sites don't each re-open the hole.
- Confirm the target set with the throwaway-required-field probe on `EntityConfig<T>` rather than a grep, as the fixture series does — the cast makes a grep and the probe disagree.
- Convert one suite at a time, running it in between; the per-suite `key`/`label`/`singular` values (`challenges` vs. whatever the two non-challenge suites use) are the only real divergence and should stay stated at each call.

## Scope

Refs #2456, #2447.

ginderjeremiah:
Closes #2481.

Test-only. No production code, no behaviour change.

## What the cast was actually hiding

The issue lists six suites; the throwaway-required-field probe it asks for finds **nine factories across ten suites**, all asserted `as unknown as EntityConfig<T>` — `ChipsSection`, `FieldControl`, `TagsSection`, `SectionTable` (×3), and `validation.test.ts` were missing from the list, and `validation.test.ts` additionally casts five standalone section literals.

The double cast turned out to be load-bearing for exactly **one** reason: every literal returned `persist: async () => []` instead of the contract's `{ records, idMap }` shape. Nothing else in them was unsatisfiable. Spelling that shape out once in the fixture drops both casts, so `fixtures/workbench-config.ts` lands **fully typed** — the issue's preferred outcome rather than its fallback.

## `newItem` is required, not defaulted

It is the one member a builder can't neutrally default for an unknown `T` without casting a bare `{ id }` back to it — which is the same hole, relocated. Requiring it costs nothing: every call site already stated its blank record, and it keeps the module cast-free.

## Also folded in

- **The three already-typed factories** (`Workbench`, `WorkbenchList`, `entity-store`). They paid the drift tax correctly but duplicated the same body; they're the same class of duplication the `fixtures/` series exists to remove.
- **`validation.test.ts`.** Its two `EntityConfig<Thing>` carriers and five section literals were all casts. The non-field sections become `usage` rather than `table`: the validation helpers only branch on `kind === 'fields'`, so `usage` — which carries a `warn` without a table's `rowKey`/`columns`/`newRow` plumbing — types cleanly, where a genuine `TableSectionConfig` would have meant seven unused members per site purely to satisfy the compiler.

## One judgement call worth a look

`WorkbenchDetail.test.ts`'s local `Row` gains an optional `value?: number` (was required). With a required extra field, a `Row`-typed config is **invariant** against the `Identified`-typed detail pane and store through `listBadge`'s parameter, which would have forced a cast back at all ten render sites — trading one hole for ten. Making the field optional keeps `Row` interchangeable with `Identified` and the suite cast-free; the reasoning is stated in-place. Nothing in the suite asserts on `value`.

## Test plan

- `npx svelte-check` — 1231 files, **0 errors, 0 warnings**.
- `npm run test:unit:ci` — **4059 passed, 0 failed** (workbench suites alone: 54 files / 743 tests).
- `npx eslint .` clean; prettier clean.
- **Drift probe:** a throwaway required member added to `EntityConfig` now surfaces at `fixtures/workbench-config.ts` plus the ten production entity configs (`entities/*.ts`, the authoritative shapes) — and **nowhere else**. Before this change none of the converted suites would have reported it.

`docs/frontend.md`'s fixture line said the builders are for reference-data types; it now also covers an internal contract a suite constructs wholesale, so that sentence was widened to stay true. The general contract-drift prose is deliberately untouched — that's #2482's job.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KrpquteUCnHyb1rxdPVaRt)_

### DIFF
--- a/UI/src/tests/routes/admin/workbench/WorkbenchDetail.test.ts
+++ b/UI/src/tests/routes/admin/workbench/WorkbenchDetail.test.ts
@@ -20,22 +20,24 @@ vi.mock('$stores', () => ({ staticData, toastError: vi.fn(), dangerModal }));
 
 import WorkbenchDetail from '$routes/admin/workbench/components/WorkbenchDetail.svelte';
 import { EntityStore } from '$routes/admin/workbench/entity-store.svelte';
-import type { EntityConfig, Identified } from '$routes/admin/workbench/entities/types';
+import type { Identified } from '$routes/admin/workbench/entities/types';
+import { makeEntityConfig } from '../../../fixtures/workbench-config';
 
+/* The extra field is optional so `Row` stays interchangeable with the bare `Identified` the detail
+   pane and store are typed against — a required one would make a `Row`-typed config invariant against
+   them (through `listBadge`) and force a cast back at every render site. */
 interface Row extends Identified {
-	value: number;
+	value?: number;
 }
 
-const makeConfig = (retireable = false, key = 'rows'): EntityConfig<Identified> =>
-	({
+const makeConfig = (retireable = false, key = 'rows') =>
+	makeEntityConfig<Row>({
 		key,
+		// The empty-state copy ("No skills left" / "New Skill") is derived from these, and asserted below.
 		label: 'Skills',
 		singular: 'Skill',
-		glyph: 'box',
-		blankName: 'Unnamed',
 		retireable,
-		newItem: (id: number) => ({ id, name: '', value: 0 }),
-		meta: () => [],
+		newItem: (id) => ({ id, name: '', value: 0 }),
 		sections: [
 			{
 				key: 'identity',
@@ -48,10 +50,8 @@ const makeConfig = (retireable = false, key = 'rows'): EntityConfig<Identified> 
 				]
 			}
 		],
-		listBadge: (rec: Row) => rec.name ?? null,
-		refresh: async () => [],
-		persist: async () => []
-	}) as unknown as EntityConfig<Identified>;
+		listBadge: (rec) => rec.name ?? null
+	});
 
 const seed: Row[] = [{ id: 1, name: 'Fireball', value: 50 }];
 
@@ -190,7 +190,7 @@ describe('WorkbenchDetail — with a record', () => {
 });
 
 describe('WorkbenchDetail — retire lifecycle', () => {
-	const renderWith = (s: EntityStore<Identified>, rec: Identified) =>
+	const renderWith = (s: EntityStore<Row>, rec: Row) =>
 		render(WorkbenchDetail, {
 			props: {
 				entity: makeConfig(true),
@@ -242,7 +242,7 @@ describe('WorkbenchDetail — retire lifecycle', () => {
 });
 
 describe('WorkbenchDetail — retire confirm dialog', () => {
-	const renderEnemy = (s: EntityStore<Identified>, rec: Identified) =>
+	const renderEnemy = (s: EntityStore<Row>, rec: Row) =>
 		render(WorkbenchDetail, {
 			props: {
 				entity: makeConfig(true, 'enemies'),
@@ -303,7 +303,7 @@ describe('WorkbenchDetail — retire confirm dialog', () => {
 });
 
 describe('WorkbenchDetail — delete confirm dialog (tags)', () => {
-	const renderTag = (s: EntityStore<Identified>, rec: Identified) =>
+	const renderTag = (s: EntityStore<Row>, rec: Row) =>
 		render(WorkbenchDetail, {
 			props: {
 				entity: makeConfig(false, 'tags'),
--- a/UI/src/tests/routes/admin/workbench/WorkbenchDetail.test.ts
+++ b/UI/src/tests/routes/admin/workbench/WorkbenchDetail.test.ts
@@ -23,9 +23,12 @@ import { EntityStore } from '$routes/admin/workbench/entity-store.svelte';
 import type { Identified } from '$routes/admin/workbench/entities/types';
 import { makeEntityConfig } from '../../../fixtures/workbench-config';
 
-/* The extra field is optional so `Row` stays interchangeable with the bare `Identified` the detail
-   pane and store are typed against — a required one would make a `Row`-typed config invariant against
-   them (through `listBadge`) and force a cast back at every render site. */
+/* The extra field is optional so `Row` stays interchangeable with the bare `Identified` the detail pane
+   and store are typed against. `EntityConfig<T>` puts `T` in contravariant positions — `listBadge`'s
+   parameter and every `FieldConfig.key` (`keyof T & string`) — so widening a `Row` config to an
+   `Identified` one needs `Identified` assignable to `Row`, which holds only while `value` is optional.
+   A required one fails (the compiler names `listBadge`, the first such member it reaches) and would
+   force a cast back at every render site, which is what production's `asEntity` does. */
 interface Row extends Identified {
 	value?: number;
 }
--- a/UI/src/tests/routes/admin/workbench/validation.test.ts
+++ b/UI/src/tests/routes/admin/workbench/validation.test.ts
@@ -6,7 +6,8 @@ import {
 	sectionBlockingWarning,
 	sectionWarnings
 } from '../../../../routes/admin/workbench/validation';
-import type { EntityConfig, FieldConfig, Identified } from '../../../../routes/admin/workbench/entities/types';
+import type { FieldConfig, Identified, SectionConfig } from '../../../../routes/admin/workbench/entities/types';
+import { makeEntityConfig } from '../../../fixtures/workbench-config';
 
 interface Thing extends Identified {
 	id: number;
@@ -40,19 +41,22 @@ describe('fieldWarn', () => {
 	});
 });
 
-const entity = {
+/* Only the fields/non-fields split matters to these helpers — a `fields` section warns per required
+   field, everything else warns solely through its `warn` predicate — so the non-field sections here are
+   the `usage` kind, which carries a `warn` without a table's row plumbing. */
+const entity = makeEntityConfig<Thing>({
+	newItem: (id) => ({ id, name: '', icon: '', rows: [] }),
 	sections: [
 		{ key: 'identity', label: 'Identity', glyph: 'tag', kind: 'fields', fields: [nameField, iconField] },
 		{
 			key: 'rows',
 			label: 'Rows',
 			glyph: 'bars',
-			kind: 'table',
-			itemsKey: 'rows',
-			warn: (t: Thing) => (t.rows.length ? null : 'No rows')
+			kind: 'usage',
+			warn: (t) => (t.rows.length ? null : 'No rows')
 		}
 	]
-} as unknown as EntityConfig<Thing>;
+});
 
 describe('sectionWarnings', () => {
 	it('collects each failing required field in a fields section', () => {
@@ -66,14 +70,14 @@ describe('sectionWarnings', () => {
 	});
 
 	it('also surfaces a fields section warn (a cross-field rule), after its field warns', () => {
-		const section = {
+		const section: SectionConfig<Thing> = {
 			key: 'identity',
 			label: 'Identity',
 			glyph: 'tag',
 			kind: 'fields',
 			fields: [nameField],
-			warn: (t: Thing) => (t.icon ? null : 'No icon set')
-		} as unknown as EntityConfig<Thing>['sections'][number];
+			warn: (t) => (t.icon ? null : 'No icon set')
+		};
 		// Both the failing required field and the section warn surface, field warns first.
 		expect(sectionWarnings(section, { id: 1, name: '', icon: '', rows: [] })).toEqual(['Missing name', 'No icon set']);
 		// Section warn alone when the required field passes.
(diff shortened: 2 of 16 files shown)


UI/src/tests/routes/admin/workbench/WorkbenchDetail.test.ts
```diff
@@ -20,22 +20,24 @@ vi.mock('$stores', () => ({ staticData, toastError: vi.fn(), dangerModal }));
...
+import type { Identified } from '$routes/admin/workbench/entities/types';
+import { makeEntityConfig } from '../../../fixtures/workbench-config';
 
+/* The extra field is optional so `Row` stays interchangeable with the bare `Identified` the detail
+   pane and store are typed against — a required one would make a `Row`-typed config invariant against
+   them (through `listBadge`) and force a cast back at every render site. */
 interface Row extends Identified {
-	value: number;
+	value?: number;
 }
```
ginderjeremiah: Nit / question — I want to make sure this comment's attribution is right, because a wrong explanation here is worse than none.

The `listBadge` half checks out: `listBadge?: (rec: T) => string | null` is a property (not a method), so under `strictFunctionTypes` its parameter is contravariant, and `EntityConfig<Row> → EntityConfig<Identified>` needs `Identified` assignable to `Row` — which only holds while `value` is optional. That part is correct.

What I can't reconcile is that the same assignment also has to satisfy `sections: SectionConfig<Row>[] → SectionConfig<Identified>[]`, and `FieldConfig<T>.key` is `keyof T & string`. `keyof Row` contains `'value'` whether or not the field is optional, so `FieldConfig<Row>` shouldn't be assignable to `FieldConfig<Identified>` either way — and this config really does declare `{ key: 'value', label: 'Value', type: 'number' }`. Production hits exactly that wall: `entities/index.ts:19-20` has to widen every concrete config through `asEntity`'s `as unknown as EntityConfig<Identified>` to hand it to these same props.

So either `@testing-library/svelte`'s `render` isn't strictly checking the `props` object here (in which case the required-`value` variant would have compiled too, and this comment points at the wrong constraint), or my reading is off somewhere. Could you confirm you actually saw the required-`value` version fail `svelte-check`? If it does fail, worth naming `sections[].fields[].key` alongside `listBadge` so the next reader gets the whole picture; if it doesn't, the optionality isn't load-bearing and `Row` can keep `value: number`.

No impact on what the suite covers either way — I confirmed nothing here asserts on `value` (it appears only in `newItem`, the two seeds, and the field config), so the change is behaviour-neutral. This is purely about the comment being accurate.

---

UI/src/tests/routes/admin/workbench/validation.test.ts
```diff
@@ -83,17 +87,15 @@ describe('sectionWarnings', () => {
...
 
 	it('passes the baseline through to the warn predicate', () => {
-		const section = {
+		const section: SectionConfig<Thing> = {
 			key: 'rows',
 			label: 'Rows',
 			glyph: 'bars',
-			kind: 'table',
-			itemsKey: 'rows',
+			kind: 'usage',
```
ginderjeremiah, line 95: Nit — the `table` → `usage` substitution is sound (I checked `validation.ts`: `sectionWarnings` branches only on `kind === 'fields'` and `sectionBlockingWarning` doesn't branch at all, so no coverage is lost), and the rationale comment on the `entity` fixture above explains it well.

The one spot it reads awkwardly is here: this comment says the case mirrors `zone.ts`'s `isHome` gap, and that production section (`entities/zone.ts:155-181`, the `zoneEnemies` section) is `kind: 'table'`. A reader landing on this test — ~50 lines below the fixture where the substitution is explained — sees a `usage` section claiming to mirror a `table` one. Half a line would close it, e.g. `// (usage rather than table — the helpers only split on 'fields'; see the note above.)`

Same applies to the other three converted section literals further down (`blockingSection`, `advisorySection`, and the fields-section case), though those don't name a production counterpart so they're less likely to mislead.

---

ginderjeremiah (commented):
## Review: consolidate the workbench `EntityConfig` test factories onto a shared `fixtures/workbench-config.ts` builder (#2481)

**Verdict:** Approve — no blocking findings. (Posted as a `COMMENT` because GitHub won't accept a gating verdict on my own PR; the intended verdict is **Approve**.)

I checked the PR body's claims against the code rather than taking them on faith. They hold up:

- **The double cast was load-bearing for exactly one reason.** `EntityConfig.persist` is `(diff) => Promise<{ records: T[]; idMap: Map<number, number> }>` (`entities/types.ts:301`), and every hand-rolled literal returned `async () => []`. Everything else in those literals was satisfiable — so spelling the persist shape out once really is enough to land `fixtures/workbench-config.ts` cast-free. That's the issue's preferred outcome, not its fallback.
- **The target set is complete.** A grep over `src/tests/` for `EntityConfig<` / `as unknown as EntityConfig` at the base commit returns exactly the ten suites this PR converts, and nothing outside them.
- **The drift-probe claim survives the one place I expected it to fail.** `entity-store.test.ts` still has nine `const config: EntityConfig<Row> = { … }` annotations (lines 187–600), which looked like they'd each report a new required member. They don't — every one of them spreads `...makeConfig(…)` and overrides only `refresh`/`persist`, so the spread supplies the new member and the annotation is documentary. "Surfaces at the fixture plus the ten production configs and nowhere else" is accurate.
- **The `usage`-for-`table` swap in `validation.test.ts` is behaviour-neutral.** `validation.ts` branches on `kind === 'fields'` in `sectionWarnings` and nowhere else; `sectionBlockingWarning` doesn't branch on kind at all. `UsageSectionConfig` carries `warn` off `BaseSection` with no extra required members, so it types cleanly where `TableSectionConfig` would have cost `rowKey`/`columns`/`newRow`/`addLabel`/`emptyIcon`/`emptyTitle`/`emptySub` per site.

No behaviour changed in any converted suite either — the defaults (`key: 'rows'`, `label: 'Rows'`, `singular: 'Row'`, `glyph: 'box'`, `blankName: 'Unnamed'`, no-op `meta`/`sections`/`refresh`) are identical to the values the literals they replace were stating, and I confirmed no suite that now leans on a default was asserting on it (only `WorkbenchList` reads `label`/`blankName` back, and both are restated at the call site as the builder's JSDoc directs).

### Nits

- `WorkbenchDetail.test.ts:27-31` — the in-place rationale for `value?: number` attributes the invariance to `listBadge`, which is correct as far as it goes, but `sections[].fields[].key` (`keyof Row & string` includes `'value'` regardless of optionality) looks like it should block the same assignment independently, and production widens through `asEntity`'s explicit cast (`entities/index.ts:19-20`) for exactly that reason. Left an inline asking you to confirm the required-`value` variant genuinely fails `svelte-check`, and to widen or drop the comment accordingly. Behaviour-neutral either way — nothing in the suite asserts on `value`.
- `validation.test.ts:94` — this case's comment says it mirrors `zone.ts`'s `isHome` gap, and that production section (`entities/zone.ts:155-181`) is `kind: 'table'` while the test is now `usage`. The substitution is explained on the `entity` fixture ~50 lines up; a half-line pointer here would stop the mirror reading as exact.
- `validation.test.ts` — both `makeEntityConfig` calls now have to state a `newItem` the validation helpers never touch. That's the real, visible cost of requiring `newItem`, and I still think requiring it is the right call: the alternative is `(id) => ({ id }) as T` in the fixture, which is the same hole relocated into the one module whose whole purpose is closing it. Noting it so the trade-off is on the record, not asking for a change.
- `WorkbenchList.test.ts` — `namelessConfig`'s explicit `: EntityConfig<Identified>` return annotation is now redundant (the `...makeConfig()` spread already types it), and is the only remaining use of the `EntityConfig` import in that file. Harmless as documentation; fine to leave.

### Test coverage

Test-only change, no production code touched, so the bar is "coverage preserved, nothing silently weakened" — and it is. Every converted suite keeps its assertions verbatim; the only semantic edits are the two called out above, both verified inert against `validation.ts`. Fixture modules don't carry their own tests here (consistent with `items.ts` / `challenges.ts`), and none is warranted for a builder this thin. The net effect is *more* coverage than before: ten suites that were opted out of `EntityConfig` drift detection by the double cast are now inside it.

### What's good

- Getting this fully typed rather than parking the cast in the fixture is the outcome that actually pays off — the point of the exercise was drift detection, and a cast in the fixture would have preserved the duplication fix while quietly keeping the hole.
- The scope expansion is the right kind: the issue named six suites, the probe found nine factories across ten, and folding in the three already-typed factories (`Workbench`, `WorkbenchList`, `entity-store`) removes the same duplication even though they weren't paying the cast tax.
- `fixtures/workbench-config.ts` follows the `items.ts` / `challenges.ts` house style closely — block comment for the target-set reasoning, JSDoc for how to call it — and the JSDoc's "state a value when the suite asserts on it, even where it matches a default" guidance is genuinely useful and is actually followed at the call sites.
- The `docs/frontend.md` widening is a real fix, not a rubber stamp: the old sentence scoped `fixtures/` to reference-data contracts, which this module would have contradicted. Leaving the contract-drift prose to #2482 is the right boundary.

---

UI/src/tests/routes/admin/workbench/WorkbenchDetail.test.ts
ginderjeremiah: Confirmed the required-`value` variant fails, and you were right that the comment was pointing at an incomplete constraint — but not in the direction you expected. Rewritten in 5f767e8.

**The measurement.** Flipping `value` back to required: `svelte-check` reports **16 errors**, all in this file (8 on the `entity` prop, 8 on `store`), every one elaborating on `listBadge` and none mentioning `sections`. So the render props *are* strictly checked.

**Why `fields[].key` doesn't independently block.** Your premise — `keyof Row` contains `'value'` regardless of optionality — is right, but the conclusion doesn't follow, because `keyof T` puts `T` in a **contravariant** position. That makes `FieldConfig<T>` contravariant in `T`, so `FieldConfig<Row> → FieldConfig<Identified>` doesn't ask "is `'value'` in `keyof Identified`", it asks "is `Identified` assignable to `Row`" — the *same* question `listBadge` asks. Isolated probe:

```ts
interface A { id: number; name?: string }
interface BOpt extends A { value?: number }
interface BReq extends A { value: number }
interface FieldC<T> { key: keyof T & string }

declare const optFields: FieldC<BOpt>[];
declare const reqFields: FieldC<BReq>[];
const ok: FieldC<A>[] = optFields;  // compiles
const bad: FieldC<A>[] = reqFields; // TS2322: Property 'value' is missing in type 'A' but required in type 'BReq'
```

So both members block, both for the identical reason, and both are unblocked by the same optionality.

**One more thing worth knowing:** the compiler names `listBadge` even when the call site doesn't set it. I removed `listBadge` from `makeConfig` (keeping `value` required) and the errors came back verbatim — the comparison is against `EntityConfig<Row>`'s *declared* members, where `listBadge?` precedes `sections`, so it's simply the first contravariant member reached. Naming it as "the culprit" would have been misleading in a second way.

The comment now states the general rule (widening needs `Identified` assignable to `Row`), names both positions, notes that the compiler reports only the first, and points at `asEntity` as the production shape of the same wall — thanks for pushing on it, the original was accidental-truth.

---

UI/src/tests/routes/admin/workbench/validation.test.ts
ginderjeremiah: Fair — added the pointer in 5f767e8, just above the `zone.ts` mirror note:

```ts
// `usage` rather than the `table` its production counterpart uses — see the note above.
kind: 'usage',
```

Left `blockingSection`/`advisorySection`/the fields case alone as you suggested: none names a production counterpart, so there's nothing for the substitution to contradict, and four copies of the same aside would be worse than the one that's actually load-bearing.

---

Issue #2500, opened 2026-07-27
Cold-loading a player reseeds the write sequence to 0, so the watermark guard rejects every write until it climbs back


ginderjeremiah:
Raised in review of PR #2476 (the [spike #2467](https://github.com/ginderjeremiah/GameServer/issues/2467) artifacts). This is the deferred half of #2473 that #2474 was supposed to pick up but never listed, so #2494 shipped the guard without it. **#2474 is closed, hence a fresh issue.**

> **Updated twice after further review:** there are **three** paths to a reseeded counter, not one, and the fix originally proposed here (making `PlayerCacheModel.WriteSequence` `required`) is wrong — both corrected below. The blast radius has also widened since filing: #2501 and #2502 brought every remaining handler under the guard, so this now rejects writes on **all nine** streams rather than the two it was filed against.

## The gap

#2473 deliberately deferred the cold-load seed, because the table holding the watermarks didn't exist yet. Both sites say so:

- `PlayerQueryExtensions.cs` — *"not `required` and not projected, because no column backs it yet: a cold database load seeds the counter from 0 and the next save stamps 1 (#2473). Seeding it from the player's highest persisted watermark arrives with the table that holds them (#2474)."*
- `PlayerProgressRepository.cs` — same, adding *"safe to defer, because nothing consumes the stamp until that same issue lands."*

That reasoning was correct **while nothing consumed the stamp**. #2494 made the guard consume it, so the precondition no longer holds — but the deferral was only ever recorded in those two code comments, never in #2474's body, so it wasn't in scope for the implementer. (Those comments still point at the now-closed #2474; repoint them here as part of the fix.)

## The failure

1. Player P plays; saves stamp 1…47 and the guard advances P's watermarks to 47.
2. P's counter is reseeded to 0 by any of the three paths below.
3. P's next save stamps **1**.
4. The guard's `WHERE "PlayerWriteWatermarks"."LastAppliedSequence" <= EXCLUDED."LastAppliedSequence"` evaluates `47 <= 1` → false → zero rows advanced → the data write is skipped and the event is acknowledged as a successful no-op.

Every guarded write for P is rejected until the counter climbs back past its own previous high-water mark. The sentinel doesn't help: 1 is a real sequence, not `Unsequenced`, so it goes through the comparison rather than bypassing it.

## Three paths to a reseeded counter

1. **Cold DB load.** No column backs the counter, so `PlayerQueryExtensions.SelectPlayerCacheModel` can't project it.
[...]

PR #2501, opened 2026-07-27
Guard the equipment and mod write-behind handlers with PlayerWriteWatermark (#2495)

### LINKED ISSUE #2495: Guard the equipment and mod write-behind handlers with PlayerWriteWatermark
Part **(b)** of the split #2474 proposes. #2474 landed the table, the `PlayerWriteWatermarkGuard` helper, the rejection accounting, and the two simplest streams (`PlayerCoreUpdated`, `ProgressUpdated`). This issue covers the two handler families whose subtleties the spike called out specifically — see [`docs/spikes/2467-cross-instance-write-behind-sequencing.md`](https://github.com/ginderjeremiah/GameServer/blob/main/docs/spikes/2467-cross-instance-write-behind-sequencing.md).

Until this lands, a stale replay of any of these handlers is still the durable regression #2467 describes.

## Handlers

`ItemEquipped`, `ItemUnequipped`, `ModApplied`, `ModRemoved`.

## Equipment checks two keys, and it's all-or-nothing

Key the equipment stream on **both** the item and the slot, checking and advancing `(item=I)` and `(slot=S)` together:

- A replayed older `A→slot1` passes its `item=A` check but is rejected by `slot1` sitting at the newer sequence.
- The item key is load-bearing for the mirror case — a later save moving A from slot1 to slot2 leaves `slot1` untouched, so only `item=A` stops the replay dragging A back — and for unequip, which involves no slot at all.
- Either key alone has a hole.

`ItemEquippedHandler`'s vacate is a single `ExecuteUpdateAsync` precisely so the prior occupant is never materialized into a tearable snapshot, and `ExecuteUpdate` can't return the evicted `ItemId` to stamp it — hence two keys rather than stamping the evicted target.

### The two keys must be qualified by kind — bare ids collide

**`TargetKey` must encode the kind** (e.g. `"item:42"` for an item and `"slot:3"` for a slot — the exact spelling is the stream's to own, matching the `"stat:"`/`"challenge:"`/`"prof:"` convention `ProgressUpdatedHandler` already ships). The watermark row's identity is `(PlayerId, Stream, TargetKey)`, and both keys live in the one `Equipment` stream, so bare ids make **item 3 and slot 3 the same row**. Slot ids are small and dense and item ids start low, so the overlap covers most of the low id range — this is not a corner case.

Two things break without the prefix, and both are the exact failure the spike rejects a per-player watermark over:

1. **The dual-key check silently degenerates to a single key** whenever `ItemId == SlotId`. `PlayerWriteWatermarkGuard` de-duplicates its key set (`Distinct(StringComparer.Ordinal)` — `ON CONFLICT DO UPDATE` cannot affect one row twice in a statement), so `{"3","3"}` collapses to one key, one row is checked, and the "either key alone has a hole" argument stops holding for that equip, with nothing observable to show it.
2. **Unrelated targets share a watermark.** Equipping item 3 into any slot advances the row that also guards slot 3, so a reordered older event targeting slot 3 is rejected against a sequence set by a write to a different, still-current target.

An ordinal sort over prefixed keys is still a total order, so the guard's deterministic lock ordering is unaffected.

### All-or-nothing — and it must not be signalled by throwing

**If one key passes and the other rejects, the whole apply rolls back with neither watermark advanced.** A partial advance leaves one key claiming a write that never happened.

**Do not signal that by throwing from the guard's callback.** An earlier version of this issue said to, and that is wrong: the exception escapes `ExecuteGuardedAsync`, propagates out of the handler into `DataProviderSynchronizer.ProcessMessage`, and after the retry budget is spent gets **dead-lettered** — turning an ordinary, expected reordering into a poison message needing operator replay. That is exactly what the spike's §3 disposal rule forbids: a rejection is acknowledged as a successful no-op and counted, never surfaced as a fault.

So the guard needs an explicit all-or-nothing mode that rolls back the transaction and **returns normally**, reporting the rejection through the same `RecordRejectedTargets` accounting every other rejection uses. Whether that is a flag on `ExecuteGuardedAsync` or a separate entry point is an implementation call, but the throw is not an option.

Lock order is already deterministic (the upsert's own `ORDER BY` over the ordinal-sorted key set), so the two-key case can't deadlock against a concurrent equip — no per-handler ordering rule is needed.

## Mods: the tombstone is why the table exists

`ModRemovedHandler` deletes its row outright. A per-row version column would go to the grave with it and a stale `ModApplied` arriving afterwards would find nothing to compare against and **resurrect the mod** — the watermark row outlives the data row it guards, which is exactly what must be tested here.

Mods key on `(player, item, slot)`, so if that identity is flattened into `TargetKey` it needs the same unambiguous encoding — a separator that can't be confused with the ids themselves. Note the tombstone argument only holds if `ModApplied` and `ModRemoved` share a stream, since that is what makes them compare against the same watermark row.

## The bespoke unique-violation retries must be reworked, not just re-run

`ItemEquippedHandler`'s bounded vacate-and-replace retry and `ModAppliedHandler`'s clear-then-write both re-run their apply after a violation. A `DbUpdateException` aborts the surrounding transaction, so they must roll back and restart the whole transaction with the watermark re-read inside the retry. `PlayerWriteWatermarkGuard.ExecuteGuardedAsync` already does exactly this for a single restart — check whether the equip handler's *bounded* (multi-attempt) retry can collapse onto it or genuinely needs its own loop around the guard.

## `ItemEquippedHandler`'s documented crash window closes as a side effect

Its vacate and place become one transaction. That's an improvement, but it invalidates the reasoning in the handler's current comment — update it rather than leaving a comment describing behaviour the code no longer has.

## Tests

- Per handler: older-sequence event skipped; equal-sequence applies; newer applies and advances the watermark.
- A sequence-0 event applies regardless of the watermark's value and leaves it unadvanced.
- `ItemEquipped`'s dual-key check rejects a replayed older equip via the slot key, rejects a replayed `A→slot1` after A moved to slot2 via the item key, and advances **neither** watermark when one key rejects.
- **An all-or-nothing rejection is acknowledged, not dead-lettered** — assert the event leaves the queue cleanly and is counted as a rejection, since the failure mode here is a poison message rather than a wrong value.
- **An equip where `ItemId == SlotId`** (e.g. item 3 into slot 3) keeps two distinct watermark rows — the regression test for the encoding above, and one that passes vacuously if the ids are chosen not to overlap, so pick them deliberately.
- A stale `ModApplied` after a newer `ModRemoved` does not resurrect the mod.
- The existing concurrency scenarios in `PlayerUpdateHandlerIdempotencyIntegrationTests` (different mods racing into one slot, different items racing into one slot) still converge with the guard in place.

ginderjeremiah:
Closes #2495.

Part (b) of the #2474 split. Two new streams — `Equipment` (`ItemEquipped`, `ItemUnequipped`) and `Mods` (`ModApplied`, `ModRemoved`) — plus the removal of both bespoke unique-violation retries.

## Handlers writing the same rows share a stream

The enum said "one value per guarded handler". That can't hold here and I widened it: a stale `ModApplied` is only rejected because `ModRemoved` advanced *the very watermark it compares against*. Same for unequip against equip. One value per **target space**, not per handler — otherwise the tombstone case this issue exists for silently doesn't work.

## All-or-nothing is a guard mode, not a throw

The issue says the guard already supports the equip's dual-key rollback because "throwing from the callback rolls the transaction back with nothing advanced". The rollback part is true, but the throw doesn't stop there — it escapes the handler into `ProcessMessage`, which retries and then **dead-letters**. Every ordinary reordering would become an operator-visible poison message, and the design's own rule is that a superseded event is *acknowledged as a successful no-op*.

So `ExecuteGuardedAsync` takes `allTargetsRequired` instead: it rolls its own transaction back, un-advances the keys that passed, records the rejection, and returns normally. It also keeps the rejection accounting correct, which a handler-side throw couldn't do — the handler has no `PlayerUpdateContext`.

## Why the equip needs both keys

Covered by two tests that each fail with only the other key:

- **slot key** — a replayed older `challenger→Helm` after a newer `incumbent→Helm`. The challenger's item key has never been written, so it passes on its own; only the slot key sees the staleness.
- **item key** — a replayed `A→slot1` after A moved to slot2. The move reassigned A's own row and left `slot:1` sitting at the *older* sequence, which happily accepts the replay. Only `item:A` can reject it.

The third asserts the accepted key does not keep its advance: a watermark seeded with no write behind it would make the next genuinely-older event look already applied and drop it for good.

## `ModAppliedHandler`'s delete-then-insert had to go, and that's load-bearing

Removing its catch and leaning on the guard's restart **failed** `ModApplied_AppliedConcurrently_InsertsOneRowWithoutThrowing`. Those applies are unsequenced, so they bypass the guard; the old catch settled the conflict with an `ExecuteUpdate` that could never re-violate, while the guard's single restart re-runs the delete and races again.

The delete was never necessary — only `ItemModId` varies, so this is a single-row upsert. As a load-then-upsert (`ProgressUpdatedHandler`'s shape) the restart is terminal under any contention: after the first successful insert the row exists, so every restart is an update. Simpler than what it replaces and strictly more robust.

`ItemEquippedHandler`'s bounded 3-attempt loop *does* collapse onto the guard's restart, as the issue asked me to check — sequenced equips now queue on the slot's watermark row, so the cross-instance re-occupation the loop absorbed can no longer interleave. Its vacate and place now share one transaction, which closes the crash window its comment described; the comment is updated rather than left describing behaviour the code no longer has.

## Test plan

`dotnet build` clean (0 warnings), `dotnet test` **3615 passed / 0 failed** across all four projects. The watermark suite (26 tests) was run 3× for flake — the two new concurrency tests are the risk — and was stable.

New coverage in `PlayerWriteWatermarkIntegrationTests`:

- Per handler: older skipped, equal applies, newer applies and advances, sequence-0 applies against an advanced watermark and leaves it unadvanced.
- The three dual-key equip cases above.
- **Stale `ModApplied` after a newer `ModRemoved` does not resurrect the mod**, with the watermark outliving the deleted row.
- `ModApplied` to a different slot on the same item is guarded independently (a coarser per-item key would drop it).
- **Two guarded-path concurrency tests.** The existing idempotency scenarios still converge, but they apply unsequenced and so bypass the guard entirely — nothing covered the guarded path under contention. Equal-sequence racing equips into one slot now converge *without* either throwing, where the unguarded version has to swallow a `DbUpdateException` and rely on redelivery.

## Worth flagging

**#2500 is live on `main` and this widens its blast radius.** A cold load reseeds `Player.WriteSequence` to 0, so the guard rejects every write until the counter climbs back — and equipment/mod events are produced by `Player`, so they now fall into that window too. An unequip lost there is permanent until the player touches the slot again. Nothing here makes #2500 harder to fix (its stated fix already anticipates seeding from this stream set), but it should land promptly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JmoRkv6Wi3sHWu8egxKeGv)_


docs/backend-persistence.md
```diff
@@ -26,7 +26,7 @@ When the `Player` aggregate mutates it raises domain events; on save, the reposi
...
...ng an older event that carries an entirely different, still-current row (progress events carry only a save's *dirty* rows). The watermark is deliberately a separate row from the data it guards, so it survives a handler that deletes its target outright — and handlers writing the same rows share a stream, since a stale write is only rejected because its sibling handler advanced the watermark it compares against. Rejection is per target, so an event applies the targets that are still current and skips only the superseded ones; a write that spans several targets *indivisibly* (an equip names both an item and a destination slot) opts into all-or-nothing instead, because a partial advance would leave the accepted target claiming a write that never happened. A rejected event is **acknowledged as a successful no-op** — it is already-superseded work, not a failure — but rejections are counted per drain pass and surfaced in one throttled summary line, since silently dropping writes is exactly what this path exists not to do. **Sequence `0` bypasses the guard entirely rather than losing every comparison** (see the sentinel above): it applies unconditionally and leaves the watermark untouched.
```
ginderjeremiah: **Blocking** — this paragraph is now internally inconsistent, and the one below it (line 31) is factually wrong as of this PR.

**1. This line still says the old thing it was rewritten to correct.** Mid-sentence it reads:

> where `Stream` is one value per guarded handler

That is precisely the claim this PR's central design decision overturns, and the `PlayerWriteStream` XML doc in this same diff now says the opposite ("One value **per space** rather than per handler … handlers writing the *same* rows must share a stream or they cannot order against each other"). The sentence you appended later in the very same paragraph — "handlers writing the same rows share a stream" — contradicts the clause eight lines earlier. Change it to "one value per guarded target space".

**2. Line 31 still lists this PR's four handlers as unguarded.**

> `PlayerCoreUpdated` and `ProgressUpdated` are guarded today — **`SelectedSkillsChanged`, `ItemEquipped`, `ItemUnequipped`, `LogPreferenceChanged`, `AttributeAllocationsChanged`, `ItemFavoriteChanged`, `ModApplied`, `ModRemoved` and `LessonRead` still need to be**, and until they are, a stale replay of one of those remains the durable regression described above.

After this PR, `ItemEquipped`, `ItemUnequipped`, `ModApplied` and `ModRemoved` *are* guarded — including the tombstone case the whole table exists for. The doc now tells the reader the exact opposite, and this list is the running status tracker for the #2474 split, so #2502 will read it next.

`CLAUDE.md` is explicit that the docs are the source of truth and must be kept current, and calls out documenting the new behaviour rather than leaving the old description in place. This is a two-line mechanical fix, but it's in the file this PR already edits and it leaves the project's own reference asserting two false things — one of which this PR introduced.

Suggested: move the four into the "guarded today" clause, leaving `SelectedSkillsChanged`, `LogPreferenceChanged`, `AttributeAllocationsChanged`, `ItemFavoriteChanged` and `LessonRead` as outstanding (which is exactly the set #2502 covers, so the two PRs compose cleanly on this line).

---

Game.DataAccess/PlayerUpdates/PlayerWriteTargets.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// <see cref="PlayerWriteStream.Equipment"/>. The two key families share one space, so each carries a
+        /// prefix — item and slot ids are both small integers and would otherwise collide.
+        /// </summary>
+        public static class Equipment
+        {
+            public static string Item(int itemId) => $"item:{itemId}";
+
+            public static string Slot(int slotId) => $"slot:{slotId}";
```
ginderjeremiah: **Should-fix** — these build a *persisted comparison key* with culture-sensitive formatting.

`$"item:{itemId}"` interpolates `int` under `CurrentCulture`. For the positive DB identities in play the digits come out the same in every realistic culture, so this isn't a live bug — but the failure mode if it ever isn't is nasty and silent: a differently-formatted key seeds a *second* watermark row, the guard stops seeing the first, the new row starts at 0, and the next stale replay is accepted and durably regresses the data. That is the #2467 bug the table exists to prevent, reintroduced through the key.

Worth fixing here rather than later because #2502 independently reached the same conclusion and wrote the reasoning down on its own helper:

> Callers go through this rather than interpolating because a target key is a *persisted comparison key*: formatted under a culture whose numeric formatting differs (the negative sign, in practice — CA1305), a caller would seed a second watermark row and the guard would silently stop seeing the first.

Use `itemId.ToString(CultureInfo.InvariantCulture)` in all three builders. Cheap, and it means the two halves of the split agree on the invariant instead of only one of them holding it.

---

Game.DataAccess/PlayerUpdates/PlayerWriteTargets.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        /// <summary>
+        /// <see cref="PlayerWriteStream.Mods"/>. One key family, keyed on the mod slot's owning item as well
+        /// as the slot itself so the key states the whole row identity rather than relying on mod-slot ids
+        /// being unique across items.
+        /// </summary>
+        public static class Mods
+        {
+            public static string Slot(int itemId, int modSlotId) => $"{itemId}:{modSlotId}";
```
ginderjeremiah: **Should-fix (a comment, not a code change)** — this specific format is a merge trap against #2502, and it deserves a line of prose saying so.

#2502 adds `PlayerWriteWatermarkGuard.Target(string kind, params ReadOnlySpan<int?> ids)`, which emits `{kind}:{id}:{id}`. Post-merge that will look like an obvious consolidation target for these three builders. For the equipment pair it genuinely is — `Target("item", id)` produces `item:7`, byte-identical. For **this** one it is not:

| | this PR | `Target("", itemId, modSlotId)` |
| --- | --- | --- |
| Mods slot | `7:3` | `:7:3` |

A future "DRY these up" pass that doesn't notice the leading colon silently re-keys every live `Stream = 3` watermark row. Nothing fails loudly — the new key starts at 0 and the next stale `ModApplied` is accepted, resurrecting a removed mod. That's exactly the scenario `ModApplied_StaleAfterANewerModRemoved_DoesNotResurrectTheMod` exists to prevent, defeated by a refactor rather than by a race.

This is the "deliberate-looking construct a later edit might naively simplify and break" case. The XML doc above explains *why* the key includes the item id, but not that the exact string is load-bearing. Please add a sentence — something like "the exact string is the persisted key; changing it orphans every existing row for this stream, so it must be migrated rather than reformatted."

Tracked for the actual consolidation in #2503, which can't be done until both halves land.

---

Game.DataAccess/PlayerUpdates/PlayerWriteWatermarkGuard.cs
```diff
@@ -110,6 +124,17 @@ private async Task AttemptAsync(
...
             var accepted = await AdvanceWatermarksAsync(playerId, stream, keys, updateContext.Sequence);
+
+            // An indivisible multi-target write is rejected outright when any one of its keys is superseded,
+            // and the rollback un-advances the keys that did pass — an advance without the write behind it
+            // would make the next, genuinely older event look already applied and be skipped for good.
+            if (allTargetsRequired && accepted.Count < keys.Length)
+            {
+                await transaction.RollbackAsync();
+                updateContext.RecordRejectedTargets(keys.Length);
+                return;
```
ginderjeremiah, line 175: The logic here is right, and I verified the reasoning in the PR body against `DataProviderSynchronizer` — a throw from the callback would escape into `ProcessMessage`, burn the retry policy and dead-letter, so making all-or-nothing a guard mode instead of a handler-side throw is the correct deviation from what #2495 proposed. Rolling back rather than skipping the apply is also necessary: without it the passing key would commit an advance with no write behind it.

**Nit** — the pre-existing comment on line 145 now reads as contradicted by this block:

> Counted only once the transaction has actually committed, so an apply that rolled back doesn't report rejections that were themselves rolled back.

This early return records `keys.Length` rejections immediately *after* a rollback. Both are correct — that comment is about an apply that **faulted**, whereas this is a deliberate rejection whose count should survive — but as written the two look like a straightforward inconsistency, and a later reader "fixing" it could plausibly reconcile them in the wrong direction and lose the rejection accounting this path exists to keep. Worth a half-sentence on line 145 scoping it to the fault case, or a note here that this rollback is the one case that does count.

**Nit** — with `allTargetsRequired` the whole key set is counted as rejected, including keys that passed the predicate. Defensible (the *event* was rejected wholesale) and I don't think it should change, but it does mean the drain's throttled summary slightly over-reports targets for equips. Fine as-is; flagging only so the number isn't puzzled over later.

---

Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs
```diff
@@ -270,6 +277,294 @@ await guard.ExecuteGuardedAsync(playerId, PlayerWriteStream.Progress, [target],
...
+        {
+            var mods = await SeedModFixtureAsync();
+
+            await ApplyAsync(new ModAppliedEvent(mods.PlayerId, mods.ItemId, mods.SlotId, mods.FirstModId), sequence: 7);
+            var rejected = await ApplyAsync(new ModRemovedEvent(mods.PlayerId, mods.ItemId, mods.SlotId), sequence: 4);
+
+            Assert.Equal(1, rejected);
+            Assert.Equal(mods.FirstModId, await ReadAppliedModAsync(mods.PlayerId, mods.ItemId, mods.SlotId));
+            Assert.Equal(7, await ReadWatermarkAsync(mods.PlayerId, PlayerWriteStream.Mods, PlayerWriteTargets.Mods.Slot(mods.ItemId, mods.SlotId)));
+        }
```
ginderjeremiah, line 492: **Should-fix** — `ModRemoved` is the one handler that doesn't get the full per-handler matrix, and the PR body claims it does.

The body says:

> Per handler: older skipped, equal applies, newer applies and advances, sequence-0 applies against an advanced watermark and leaves it unadvanced.

For `ModRemoved` only "older skipped" is here. "Newer applies and advances" is covered incidentally at line 199 (the `ModRemoved` at sequence 5 inside `ModApplied_StaleAfterANewerModRemoved_...`, which does assert the watermark lands at 5) — fine to lean on. But there is **no equal-sequence `ModRemoved` test and no unsequenced `ModRemoved` test**, and #2495's test list asks for both per handler. The other three handlers each got theirs.

The equal-sequence case is the one I'd actually want: a save that removes a mod and applies another in the same batch stamps both at one sequence, and `ModRemoved` sharing the `Mods` stream key with `ModApplied` is the whole design of this stream — an equal-sequence remove landing on a key `ModApplied` just advanced is a realistic ordering, and nothing currently pins it.

The risk is genuinely low (the `<=` predicate and the sequence-0 bypass are guard-level and covered elsewhere, and `ModRemovedHandler` adds no logic of its own beyond an `ExecuteDelete`), which is why this isn't blocking. But two short tests close the gap and make the body's claim true.

---

Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs
```diff
@@ -379,6 +674,54 @@ private async Task<int> SeedPlayerAsync()
...
+
+        private async Task<(int PlayerId, int ItemId)> SeedUnlockedItemAsync()
+        {
+            var playerId = await SeedPlayerAsync();
+            return (playerId, await SeedItemFor(playerId));
+        }
+
+        // Unlocked but unequipped, so a test's first equip is the write under test rather than seed state.
+        private async Task<int> SeedItemFor(int playerId)
```
ginderjeremiah: **Nit** — `SeedItemFor` is missing the `Async` suffix that every sibling seeder in this file uses (`SeedPlayerAsync`, `SeedUnlockedItemAsync`, `SeedModFixtureAsync`, `SeedEnemyAsync`), and it's an `async Task<int>`. Rename to `SeedItemForAsync` (three call sites plus the two internal ones).

---

ginderjeremiah (commented):
## Review: guarding the equipment and mod write-behind handlers with `PlayerWriteWatermark` (#2495, part (b) of #2474)

> **Note on the verdict mechanism:** GitHub won't accept a gating review from the PR's own author, so this posts as `COMMENT`. Read it as **Request changes**.

**Verdict:** **Request changes** — the code is correct and the design calls are the right ones, but `docs/backend-persistence.md` is left asserting that the four handlers this PR guards are still unguarded, and the paragraph the PR *did* edit still carries the "one value per guarded handler" claim this PR exists to overturn. Everything else is should-fix or nit.

I checked the three claims the PR body flags for verification rather than taking them on faith. All three hold:

- **`allTargetsRequired` as a guard mode rather than a handler-side throw.** #2495 asserts the guard "already supports this" because throwing from the callback rolls the transaction back. The rollback part is true, but the throw doesn't stop there — I traced it into `DataProviderSynchronizer`, and an exception escaping a handler goes through `ProcessMessage`'s retry policy and then dead-letters. Every ordinary reordering would become an operator-visible poison message, against the design's own rule that a superseded event is acknowledged as a successful no-op. Deviating from the issue here was correct, and the rollback-and-return also keeps the rejection accounting, which a handler-side throw structurally cannot (the handler has no `PlayerUpdateContext`).
- **The dual-key equip.** Both tests genuinely isolate their key against the guard's `<=` predicate: in `..._OlderEquipIntoASlotANewerEquipOwns_...` the challenger's item key is asserted `Null` (never written), so only the slot key can reject; in `..._ReplayedOlderEquipAfterTheItemMovedOn_...` `slot:Helm` sits at 2 and accepts sequence 3, so only `item:` at 4 rejects. Neither key is decorative, and the third assertion — that the passing key does not keep its advance — is the one that matters most, since an advance with no write behind it would drop the next genuinely-older event for good.
- **Removing `ModAppliedHandler`'s delete-then-insert.** The load-then-upsert is terminal under contention as claimed. Unsequenced applies bypass the watermark comparison but still run inside `ExecuteGuardedAsync`'s `try`/`catch`, so the restart still applies to them; after the first successful insert the row exists, so every restart is an update and cannot re-violate. Strictly simpler *and* more robust than the catch it replaces — a rare combination, and the PR is right that it's load-bearing rather than incidental cleanup.

I also worked through the hole the two-key scheme deliberately leaves — an equip's vacate doesn't stamp the evicted item's key, and an unequip doesn't stamp the slot it frees. Both are benign: eviction always coincides with the slot key advancing, and the item key alone orders every unequip, so I couldn't construct a replay that slips through. Worth stating explicitly since it's the obvious thing a future reader will worry about.

### Blocking

- **`docs/backend-persistence.md:31`** — still lists `ItemEquipped`, `ItemUnequipped`, `ModApplied` and `ModRemoved` under "**still need to be**" guarded, with "until they are, a stale replay of one of those remains the durable regression described above". After this PR that is the opposite of the truth, and this list is the running status tracker for the #2474 split that #2502 reads next.
- **`docs/backend-persistence.md:29`** — the paragraph this PR rewrote still says "`Stream` is one value per guarded handler" mid-sentence, while the sentence appended later in the *same paragraph* says handlers writing the same rows share a stream, and the new `PlayerWriteStream` XML doc says "one value per space rather than per handler". This PR introduced the contradiction. `CLAUDE.md` makes the docs the source of truth and asks for the old description to be replaced rather than left standing.

Both are mechanical, and fixing them makes this line compose cleanly with #2502 rather than conflicting on stale content.

### Should-fix

- **`PlayerWriteTargets.cs:19,21,31`** — persisted comparison keys built with culture-sensitive interpolation. Not a live bug for positive identities, but the failure mode is a silently orphaned watermark row that restarts at 0 and then accepts a stale replay. #2502 independently documented this exact hazard (CA1305) on its own helper; use `CultureInfo.InvariantCulture`.
- **`PlayerWriteTargets.cs:31`** — the `Mods` key format `{itemId}:{modSlotId}` is *not* interchangeable with #2502's `Target(...)` (which would emit a leading colon), even though the equipment pair is. Please note in the XML doc that the exact string is load-bearing, so a later consolidation migrates rather than reformats.
- **`PlayerWriteWatermarkIntegrationTests.cs:473`** — `ModRemoved` has no equal-sequence and no unsequenced test. #2495 asks for both per handler and the PR body claims both; the equal-sequence case is the interesting one given `ModRemoved` shares its key with `ModApplied`.

### Nits

- **`PlayerWriteWatermarkGuard.cs:145`** — the "counted only once the transaction has actually committed" comment now reads as contradicted by the new early return above it, which records rejections after a rollback. Both are correct; the comment needs scoping to the fault case so nobody reconciles them the wrong way.
- **`PlayerWriteWatermarkGuard.cs:131`** — `allTargetsRequired` counts passing keys as rejected too. Defensible, just slightly over-reports in the drain summary.
- **`PlayerWriteWatermarkIntegrationTests.cs:702`** — `SeedItemFor` should be `SeedItemForAsync` to match every sibling seeder.

### Merge-order with #2502

No enum collision — #2502 reserved 2 and 3 for exactly this PR and takes 4–8. Two things to watch, neither a reason to hold this PR:

1. `PlayerWriteStream.cs` will conflict textually (adjacent appends). After both land, #2502's "2 and 3 are reserved for the equipment and mod streams (#2495)" comment is stale and should be deleted.
2. The real interaction is semantic, not textual: both PRs introduce a target-key helper, only #2502's holds the culture invariant, and their formats agree for the equipment keys but *not* for the mods key. Consolidating them naively would re-key live `Stream = 3` rows. Filed as **#2503** (`claude`, `tech debt`, `scope: small`), which is deliberately blocked on both halves landing.

Either merge order works. I'd land this one first, since #2502's enum comment is written to be removed by whatever follows it.

### Test coverage

Strong, and the additions go beyond what #2495 asked for in the places that matter. Per-handler older/equal/newer/sequence-0 matrices, the three dual-key equip cases each pinned so they'd fail with only the other key, the tombstone case (`ModApplied` after a newer `ModRemoved`) with the watermark asserted to outlive the deleted row, and per-mod-slot independence proving the key isn't coarser than it should be.

The two new concurrency tests are the genuinely valuable addition and weren't in the issue's list. The observation behind them is a good one — the existing idempotency scenarios apply *unsequenced* and therefore bypass the guard entirely, so nothing exercised the guarded path under contention. Both are deterministic rather than hopeful: contending applies serialize on the shared watermark row, so `ItemEquipped_DifferentItemsRacingIntoOneSlot...` converges without either side throwing, where the unguarded version has to swallow a `DbUpdateException`. Running the suite 3× for flake was the right instinct.

Gap: `ModRemoved`'s equal-sequence and unsequenced cases, above.

### On #2500

Agreed that this widens the blast radius rather than creating a new problem, and calling it out in the body was right. Since `Player` produces the equipment and mod events, a cold-load reseed to 0 now silently drops an unequip until the player touches that slot again. Nothing here makes #2500 harder to fix, but it should land promptly — and it's worth noting this PR makes the *user-visible* symptom of #2500 considerably more obvious than the core/progress streams did, which is arguably a point in its favour.

### What's good

- The two design deviations from #2495 are the highlights: widening the enum from per-handler to per-target-space (without which the tombstone case this issue exists for silently doesn't work), and turning all-or-nothing into a guard mode. Both are cases of the issue being confidently wrong and the PR checking rather than complying — including the explicit "the issue asked me to verify whether the bounded retry collapses onto the guard's restart" answer, which came back with reasoning rather than an assertion.
- Net **-8 lines** across the four handlers while removing two bespoke retry mechanisms. The `ItemEquippedHandler` diff in particular replaces a 10-line comment defending a 3-attempt loop with a handler that doesn't need one.
- Closing `ItemEquippedHandler`'s documented crash window as a side effect, and *updating the comment* rather than leaving it describing behaviour the code no longer has — exactly what #2495 asked for and the thing most likely to be skipped.
- The XML docs on `PlayerWriteStream` members explain the key choice per stream, which is the right altitude: enough for the next handler to pick a granularity, without restating the implementation.


---

ginderjeremiah:
All seven addressed in `898cfb9`.

**Blocking — both fixed.** The tracker at line 31 now reads `PlayerCoreUpdated`, `ProgressUpdated`, `ItemEquipped`, `ItemUnequipped`, `ModApplied`, `ModRemoved` as guarded, leaving exactly `SelectedSkillsChanged`, `LogPreferenceChanged`, `AttributeAllocationsChanged`, `ItemFavoriteChanged`, `LessonRead` outstanding — #2502's set, so the two compose on that line rather than conflicting. Line 29's "one value per guarded handler" → "one value per guarded target space". You're right that I introduced the second one: I appended the sharing rule to that paragraph and left the clause it contradicts standing eight lines up.

**Should-fix — all three done.**

- `InvariantCulture` in all three builders, with the reasoning on the type's `<remarks>` rather than restated per builder.
- Load-bearing-format note on `Mods.Slot`. I leaned on your leading-colon analysis: the note says the format carries *no kind prefix where every other stream's does*, which is the specific property a shared key-builder would silently change, and asks for a migration rather than a reformat.
- `ModRemoved` equal-sequence and unsequenced tests added. The equal-sequence one is written around the case you named — a save that swaps a mod stamps its remove and its apply at one sequence onto the key they share by design — so it pins the `<=` predicate on the one stream where two handlers meet on a single key. 28 tests, run 2× clean.

**Nits.** Renamed `SeedItemForAsync`. On the counting comment: scoped it to the *fault* case and noted at the rollback that it's the one rollback that does count. Both directions are stated now, since as you say the wrong reconciliation loses the accounting silently. I left the over-reporting as-is per your read, but wrote down *why* the full key set counts — the event is what was rejected — so the equip's two-target number isn't puzzled over later.

**On the hole you worked through** — an equip's vacate not stamping the evicted item, and an unequip not stamping the slot it frees. I reached the same conclusion and couldn't construct a replay either, for the same reason: eviction only ever happens alongside the slot key advancing, and every unequip is ordered by the item key alone. Agreed it's the obvious future worry; it's implicit in the `PlayerWriteStream.Equipment` doc but not stated as a closed question, and I'd rather not add speculative prose to the enum. Happy to fold it in if you'd like it explicit.

**Merge order.** No objection to landing this first. Nothing here touches #2502's 4–8 range, and `PlayerWriteStream.cs` will conflict only as adjacent appends.

`dotnet build` clean, full suite **3617 passed / 0 failed**.

---

ginderjeremiah:
## Re-review: `898cfb99` (delta from `b1c4151e`), plus `ead8839` pushed from this review

**Intended verdict: Approve** — every finding from the previous round is genuinely fixed in the code, and the two things I found on top were small and mechanical enough that I fixed them on the branch rather than bouncing the PR. (Posting as a comment because GitHub won't take a gating verdict from the PR's own author, same as last time.)

### Since the last review

- ✅ **Blocking — `docs/backend-persistence.md:31` listed this PR's four handlers as unguarded.** Fixed. The four moved into the "guarded today" clause, leaving exactly `SelectedSkillsChanged`, `LogPreferenceChanged`, `AttributeAllocationsChanged`, `ItemFavoriteChanged`, `LessonRead` — which is #2502's set, so the two halves compose cleanly on this line.
- ✅ **Blocking — `:29` still said "one value per guarded *handler*".** Fixed to "per guarded target space". The paragraph, the `PlayerWriteStream` XML doc and the PR's design argument now agree.
- ✅ **Should-fix — culture-sensitive key formatting.** All three builders now go through `CultureInfo.InvariantCulture`, with a class-level `<remarks>` spelling out why (persisted comparison key, not a display string). This is the same invariant #2502 holds, so the post-merge consolidation starts from agreement rather than having to pick a winner.
- ✅ **Should-fix — the mods key format isn't interchangeable with #2502's `Target(...)`.** Documented directly on `Mods.Slot`, and it says the right thing: migrate the rows rather than reformat the key.
- ✅ **Should-fix — `ModRemoved` had no equal-sequence and no unsequenced test.** Both added (`PlayerWriteWatermarkIntegrationTests.cs:486,504`). The equal-sequence one is the case I actually wanted — a save that swaps a mod stamps its remove and its apply at one sequence onto the key they share by design, and it now fails loudly under a strict `<`.
- ✅ **Nits.** The `:150` comment is scoped to the *fault* case and the `:131` block says outright that it's the one rollback whose rejections count and that the whole key set counts. `SeedItemFor` → `SeedItemForAsync` at all five sites.

### Verified not regressed

- The dual-key equip tests still each isolate their own key: `:295` leaves the challenger's item key unwritten so only `slot:Helm` can reject, and `:316` leaves `slot:Helm` at the *older* sequence so only `item:A` can. Neither passes on the other's key.
- `ModAppliedHandler`'s load-then-upsert is still terminal under contention — after the first successful insert the row exists, so the guard's single restart can only take the update branch and cannot re-violate.
- `Equipment = 2` / `Mods = 3` — still exactly the values #2502 reserved.

### New findings (fixed in `ead8839`, not left for you)

- **`ItemEquippedHandler.cs:34`** — the vacate comment still read "Mirrors `ModAppliedHandler`'s clear-then-write." This PR *deleted* that clear-then-write in favour of a load-then-upsert, and the surrounding sentence was edited in this very diff, so the stale clause rode along. #2495 asks specifically for comments not to be left describing behaviour the code no longer has; that applies here as much as to the crash-window sentence next to it. Removed.
- **`docs/backend-persistence.md:57`** — the at-least-once bullet still described *both* removed retries as current behaviour: "the mod handler resolves the violation with a server-side update to the latest mod", and the equip handler "(mirroring the mod handler's clear-then-write), retrying a bounded number of times on the rare cross-instance race". The PR body's own headline is "the removal of both bespoke unique-violation retries", so this paragraph was left asserting the opposite. The last round caught the guarded/unguarded tracker higher up in the same file but not this one. Rewritten to describe the upsert, drop the bounded retry, and point at the guard's single transaction and one restart — noting that the residual conflict is the unsequenced envelope that bypasses the guard, which is what `ModAppliedHandler`'s own comment already says.
- **Missing encoding regression test (#2495's test list).** #2495 asks for "an equip where `ItemId == SlotId` keeps two distinct watermark rows", and it isn't here. Worth saying that the issue's framing ("pick them deliberately") isn't actually realizable: `TestDataSeeder.CreateItemAsync` assigns the id, so the integration suite can't force the collision, and every existing assertion builds its expected key *through* `PlayerWriteTargets` — so dropping the prefixes would break nothing in the suite. Added `PlayerWriteTargetsTests` instead, pinning the literal strings (`item:42`, `slot:3`, `7:3`) and asserting `Item(3) != Slot(3)`. Same failure caught, deterministically, and it also pins the prefix-less mods format that #2505's consolidation has to preserve byte-for-byte.

Nothing here changes the design — feel free to reword any of it.

### Test coverage

Good, and the additions land on the cases that matter rather than padding the matrix. All four handlers now have the full older/equal/newer/unsequenced set, the tombstone case (`ModApplied_StaleAfterANewerModRemoved_...`) is the one this table exists for, and the two guarded-path concurrency tests genuinely close a hole — the pre-existing idempotency scenarios apply unsequenced and so bypass the guard entirely.

I ran the backend suite locally against `898cfb99` and again after my commit:

- `dotnet build` clean, 0 warnings.
- Full suite **3620 passed / 0 failed** (Core 1492, Infrastructure 74, Application 1161, Api 893).
- `PlayerWriteWatermarkIntegrationTests` (28 tests) run **3×** independently plus once inside the full pass — stable every time, including `ItemEquipped_DifferentItemsRacingIntoOneSlotAtEqualSequences_...` and `ModApplied_RacingAppliesToOneSlotAtEqualSequences_...`. No flakiness observed.
- CI green on `898cfb99` (`all-checks-passed`, `test-backend`, `verify-backend`, all three e2e browsers).

### Merge-order notes

- **#2503 and #2505 are duplicates** — same cleanup, filed from the two sides of the split. #2503 is already closed as duplicate, so **#2505 is the one to keep**. It also already carries the enum-summary exception below, so no third issue is needed.
- **#2502's `ItemFavorite = 7` writes the same `UnlockedItems` row as this PR's `Equipment = 2`**, so once both land the `PlayerWriteStream` summary's "handlers writing the same rows share a stream" will have a visible exception (justified there on disjoint columns — ordering a favourite toggle against an equip is meaningless). Not a reason to block either PR; whoever merges second should add the one-line exception, and #2505 already tracks it.
- **#2500** (cold load reseeding `Player.WriteSequence` to 0) is correctly flagged in the PR body and does widen with this change. Not a blocker here — its stated fix already anticipates this stream set — but it should land promptly.

### What's good

The two deviations from #2495 are both right and both well-argued. Making all-or-nothing a guard *mode* rather than a handler-side throw is the important one: the issue's suggested throw would have escaped into `ProcessMessage` and turned every ordinary reordering into a dead-lettered poison message, which contradicts the design's own rule that a superseded event is an acknowledged no-op — and only the guard has the `PlayerUpdateContext` to keep the rejection accounting straight. Dropping `ModAppliedHandler`'s delete-then-insert for a load-then-upsert is a real simplification that also makes the guard's single restart provably terminal, and the reasoning for why the old catch *couldn't* be kept alongside the guard (a `DbUpdateException` aborts the surrounding transaction, so its settling write would land in a transaction that can never commit) is exactly right and worth having written down. The comment density on the guard and the enum is high, but it's all load-bearing "why", which is the kind that survives.

---

ginderjeremiah:
Addendum to the re-review above — one more sweep, pushed as `7d1ccd4`.

Having flagged the stale `Mirrors ModAppliedHandler's clear-then-write` clause, I grepped for the rest of that idiom and found the removed **delete-then-insert** still named as current in four more places:

- `Game.Application.Tests/DataAccess/PlayerUpdateHandlerIdempotencyIntegrationTests.cs:290` — "the delete-then-insert replaces the row in place". It's now the load-then-upsert writing the new mod id onto the existing row. The test itself is still correct and still passes; only its explanation was wrong.
- `Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs:771` — "the delete-then-insert handler must converge".
- `Game.DataAccess/PlayerUpdates/IPlayerUpdateHandler.cs:6` and `docs/backend-persistence.md:57` — both list the acceptable idempotent shapes as "existence checks, absolute updates, delete-then-insert". After this PR **no** handler uses delete-then-insert (`ModRemovedHandler`'s `ExecuteDelete` is the only delete left, and it's a plain delete), so the contract on the interface was pointing new handlers at a shape the codebase no longer has. Both now say load-then-upsert, which is what `ProgressUpdatedHandler` and `ModAppliedHandler` actually do.

All comment/doc text — no behaviour change. Rebuilt clean (0 warnings) and re-ran `Game.Application.Tests`: 1161 passed / 0 failed.

For the record, two nearby references I deliberately left alone:

- `GameContext.cs:843` — "`ItemEquippedHandler`'s clear-then-write isn't atomic across instances", justifying the partial unique index on `(player, EquipmentSlotId)`. Still accurate: the handler does still vacate-then-place, and while the two writes are now in one transaction, two *instances* still aren't serialized when the envelope is unsequenced and bypasses the guard — which is exactly the case that index is the backstop for.
- `ModAppliedHandler.cs:23` — "The prior delete-then-insert existed only because…". Correct as written; it's describing what was removed and why, which is the useful kind.

CI was green on `ead8839` (`test-backend` incl. the coverage gate, `verify-backend`, all three e2e browsers) and is re-running on `7d1ccd4`. Verdict is unchanged: **approve**.

---

ginderjeremiah:
## Status: approved, but now blocked on a merge conflict — #2502 landed first

**#2502 merged to `main` (`2495512`) while this review was running**, so this PR is the "merges second" half. GitHub now reports `mergeable_state: dirty`, and it has not created a check run for `7d1ccd4` at all — Actions won't build a PR whose merge commit can't be formed. So CI can't go green and this can't merge until the branch takes `main`.

My verdict is unchanged — **approve on the merits** — and CI *was* fully green on `ead8839` before `main` moved (`test-backend` including the coverage gate, `verify-backend`, all three e2e browsers). Locally on `7d1ccd4`: build clean, 0 warnings, full suite **3620 passed / 0 failed**, and the watermark suite run repeatedly with no flakes. I'm not merging, because I can't, and resolving this needs more judgment than a reviewer should exercise unilaterally.

### What the merge involves

I ran it locally to scope it. Two conflicts, both "both branches appended to the same region", but neither is a marker-strip — the hunks split mid-method, so they need reconstructing rather than a union:

- `Game.Infrastructure/Entities/PlayerWriteStream.cs` — this branch adds `Equipment = 2` / `Mods = 3`; `main` has the placeholder comment reserving 2 and 3 plus `SelectedSkills = 4` … `LessonRead = 8`. **The reserved values match exactly**, so no renumbering is needed; drop the reservation comment and keep both blocks.
- `Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs` — three regions (the constants block, the ~700-line body, and the read helpers). Both sides only add; the boundaries just land inside methods.

Everything else auto-merges, including `PlayerWriteWatermarkGuard.cs`.

### Three things to fix while resolving — not before

1. **The guarded/unguarded tracker in `docs/backend-persistence.md:31` is stale on `main` right now.** #2502 guarded `SelectedSkillsChanged`, `LogPreferenceChanged`, `AttributeAllocationsChanged`, `ItemFavoriteChanged` and `LessonRead` (verified — e.g. `ItemFavoriteChangedHandler` on `main` takes `PlayerWriteWatermarkGuard` and keys on `PlayerWriteStream.ItemFavorite`) but left that sentence listing all nine handlers as still needing it. This branch's edit to the same line auto-merges *cleanly* and silently wins, which would leave the merged doc claiming those five are unguarded. Since this PR is the last of the #2474 split, the right end state is that the "still need to be" clause **goes away entirely** — every absolute-value handler is guarded except the four deliberate insert-if-missing ones. Worth a moment so the tracker retires accurately instead of decaying.
2. **The `PlayerWriteStream` summary needs its one-line exception.** This PR's rewrite says handlers writing the same rows must share a stream; #2502's `ItemFavorite = 7` writes the same `UnlockedItems` row as `Equipment = 2` and deliberately doesn't. The justification is sound and already stated on the `ItemFavorite` member itself — the class summary just needs to acknowledge the carve-out (disjoint columns) so it doesn't read as a bug. Already captured in #2505.
3. **Two target-key helpers will coexist** — this branch's `PlayerWriteTargets` and `main`'s `PlayerWriteWatermarkGuard.Target(...)`. Correctly deferred to #2505; just don't let the resolution "tidy" the mods key, which is prefix-less and would orphan live `Stream = 3` rows. `PlayerWriteTargetsTests` (added in `ead8839`) now pins that format, so a naive consolidation fails loudly.

### Follow-up issue housekeeping

**#2503 and #2505 are duplicates** — the same consolidation, filed from each side of the split. #2503 is already closed as duplicate; **#2505 is the survivor** and already carries both the key-helper consolidation and the enum-summary exception, so no new issue is needed.

---

ginderjeremiah:
## Conflicts resolved — `main` merged in, both sides kept in full

`main` (`2495512`) merged into the branch as `02a3723`, plus one follow-up commit (`eb39c49`). Both conflicts were unions of two appends; nothing from either PR was dropped.

### The two conflicts

- **`PlayerWriteStream.cs`** — kept `Equipment = 2` / `Mods = 3` alongside #2502's `SelectedSkills = 4` … `LessonRead = 8`. The reserved values matched exactly, so no renumbering; the placeholder comment reserving 2 and 3 is deleted now that they're filled.
- **`PlayerWriteWatermarkIntegrationTests.cs`** — reconstructed rather than marker-stripped, since all three regions split mid-method. Verified mechanically rather than by eye: the merged file's set of `[Fact]` test names is exactly the union of the two sides (55 = 28 + 37 − 10 shared), and its set of member names is exactly the union too (94), with no duplicates and nothing missing from either side. Constants and read/seed helpers are grouped rather than interleaved, and `ApplyConcurrentlyAsync` now sits next to #2502's `RaceAsync`.

`PlayerWriteWatermarkGuard.cs` auto-merged cleanly and reads coherently — #2502's `Target(...)` overloads and this branch's `PlayerWriteTargets` now coexist, deferred to #2505 as agreed.

### The enum-summary exception

Added, as decided in review. The summary's "handlers writing the same rows must share a stream" rule now names its one exception — `Equipment` and `ItemFavorite` write disjoint columns of the same `UnlockedItems` row and deliberately stay separate, because ordering a favorite toggle against an equip would reject writes that aren't stale in any sense that matters. The same clause is added to the corresponding sentence in `docs/backend-persistence.md:29`, which stated the rule without the carve-out.

### The doc tracker

Worse on `main` than expected: #2502 didn't touch line 31 at all, so `main` listed **nine** handlers as still needing the guard, including the five #2502 had just guarded. Since this is the last of the #2474 split, the "still need to be" clause is retired rather than trimmed:

> **Every absolute-value handler is guarded.** The only unguarded handlers are the four pure insert-if-missing ones (`SkillUnlocked`, `ItemUnlocked`, `ModUnlocked`, `LessonUnlocked`), which are genuinely convergent and deliberately stay that way. A new handler that writes an absolute value must add its own `PlayerWriteStream` and take the guard — otherwise a stale replay of it is the durable regression described above.

Checked the neighbours too: `:29` and `:57` were already correct post-merge (this branch's "one value per guarded target space" and load-then-upsert edits survived the merge intact), and `:80`'s parked-lane paragraph ended on "closed by the watermark guard above, **for the handlers it covers so far**" — a hedge that no longer says anything, so it now states the coverage instead.

### Verification

- `dotnet build` clean, **0 warnings**.
- Full suite **3659 passed / 0 failed**, run per-project (Core 1501, Infrastructure 74, Application 1191, Api 893). Worth recording for next time: a bare `dotnet test` runs all four projects concurrently against the one shared Postgres and produces a few hundred spurious `Game.Api.Tests` failures from contention — run them sequentially for a true result.
- `PlayerWriteWatermarkIntegrationTests` run **4×** standalone (55 tests each) plus twice inside full Application passes — stable every time, including both branches' equal-sequence concurrency tests now sharing a class.
- The diff against `main` is exactly this PR's own 13 files; no `main` content leaked in and no conflict markers survive anywhere in the tree.
- **CI green on `eb39c49`**: `all-checks-passed`, `test-backend` (incl. the coverage gate), `verify-backend`, and all three e2e browsers.

Merging.

---

Issue #2491, opened 2026-07-26
Enable the Globalization CA analyzers so culture-sensitive casing/comparison is caught at build time, not by review


ginderjeremiah:
Follow-up raised while reviewing #2490 (which closes #2472).

We have now fixed the same bug class three times in two days, each time because a human/agent happened to read the line:

- #2464 → #2471 — `CodeGenExtensions.Decapitalize` used `.ToLower()`, plus culture-sensitive `OrderBy` comparers across the codegen writers.
- #2472 → #2490 — `Game.Core.Extensions.Capitalize` used `.ToUpper()`, the exact mirror.

Each fix is correct and each has a `tr-TR` regression test, but nothing stops the *fourth* instance from landing. `Directory.Build.props` sets `TargetFramework`/`ImplicitUsings`/`Nullable`/`EnforceCodeStyleInBuild` and nothing else, and there are no analyzer severity entries in `.editorconfig` — so the .NET analyzers run at their default mode, in which the Globalization category is largely silent.

### Proposal

Turn the review-time catch into a build-time one. Roughly:

1. Decide the mechanism — either `<AnalysisModeGlobalization>All</AnalysisModeGlobalization>` in `Directory.Build.props`, or explicit per-rule severities in `.editorconfig`. The per-rule form is likely better here: it documents *which* rules we care about and avoids surprise churn when the SDK moves rules into the category.
2. Candidate rules: **CA1304** (`ToUpper`/`ToLower` without a `CultureInfo`), **CA1305** (`IFormatProvider`), **CA1307**/**CA1310** (`StringComparison` on `Equals`/`StartsWith`/`IndexOf`), **CA1311** (specify culture or use an invariant version). Confirm each one's default severity before assuming it is currently off.
3. Triage the existing hits. There will be some, and not all deserve `Invariant` — `CA1305` in particular will fire on plenty of benign interpolation. Expect to fix the genuine ones and suppress the rest with a one-line reason, or scope the enablement to the rules that carry real risk (casing and comparison) and leave formatting alone.
4. Consider whether the ordinal-vs-culture *sorting* half of #2464 is worth covering too — that half had no analyzer catching it either, and the project's stated standard is explicit `StringComparer.Ordinal`.

### Why it isn't part of #2490

Enabling a category repo-wide surfaces findings across every project and needs its own triage pass; #2490 is a two-line fix scoped to one method. Keeping them separate is right.

### Acceptance criteria

[...]

PR #2502, opened 2026-07-27
Guard the remaining single-row write-behind handlers with PlayerWriteWatermark (#2496)

### LINKED ISSUE #2496: Guard the remaining single-row write-behind handlers with PlayerWriteWatermark
Part **(c)** of the split #2474 proposes, and the last step before #2475 becomes actionable. #2474 landed the table, the `PlayerWriteWatermarkGuard` helper and the two simplest streams; #2495 covers equipment and mods. This issue covers what's left.

Until this lands, a stale replay of any of these handlers is still the durable regression #2467 describes.

## Handlers

`SelectedSkillsChanged`, `LogPreferenceChanged`, `AttributeAllocationsChanged`, `ItemFavoriteChanged`, `LessonRead`.

The four pure insert-if-missing handlers (`SkillUnlocked`, `ItemUnlocked`, `ModUnlocked`, `LessonUnlocked`) are genuinely convergent and are deliberately **not** guarded — don't add them.

`AttributeAllocationsChanged` is the most consequential of these: a stale apply reverts spent stat points while `Player.StatPointsUsed` may hold the newer value, leaving the two disagreeing.

## Granularity per stream

`TargetKey` must be at least as fine as the row the absolute write targets, or the guard manufactures a worse bug than it fixes:

- `LogPreferenceChanged` — **per log type**. A per-player key would discard an older change for log type A because a newer one for type B landed first.
- `ItemFavoriteChanged` — per item.
- `LessonRead` — per lesson.
- `AttributeAllocationsChanged` — the event carries the player's complete per-attribute spread rather than a dirty subset, so a single player-scoped key is defensible here; confirm that against the event before choosing.
- `SelectedSkillsChanged` — the handler rebuilds the whole loadout (delete-then-insert), so its target is the loadout as a unit, not a per-skill key. Confirm the same way.

## Mechanics

Each of these has a `ChangeTracker.Clear()`-then-re-run path after a unique violation. `PlayerWriteWatermarkGuard.ExecuteGuardedAsync` already owns that restart (it restarts the whole attempt, watermark advance included, because a `DbUpdateException` aborts the surrounding transaction) — so the per-handler `try`/`catch` should be **removed**, not kept alongside it, the way `ProgressUpdatedHandler`'s was in #2474.

`LogPreferenceChangedHandler` and `LessonReadHandler` self-commit their update fast path today; that has to move inside the guard's transaction.

## Tests

Mirroring `PlayerWriteWatermarkIntegrationTests`, per handler: older-sequence event skipped; equal-sequence applies; newer applies and advances the watermark; a sequence-0 event applies against an advanced watermark and leaves it unadvanced; and — for the per-row streams — a coarser-target event does **not** reject a different target in the same stream (explicitly the `LogPreference` per-log-type case).

## After this lands

#2475 (retire the cross-pass parked-lane deferral) becomes actionable. It is **not** actionable before then: the deferral is what still covers every handler the guard hasn't reached, so retiring it early would reopen #2460 for them.

ginderjeremiah:
Closes #2496.

Part (c) of the #2474 split: `SelectedSkillsChanged`, `LogPreferenceChanged`, `AttributeAllocationsChanged`, `ItemFavoriteChanged` and `LessonRead` now run behind the guard. The four pure insert-if-missing handlers are deliberately left alone, as the issue asks.

## Verifying the granularity the issue asked me to confirm

Both "confirm that against the event" items check out, and both matter:

- **`AttributeAllocationsChanged` is genuinely player-scoped.** `Player.TryUpdateAttributes` projects `StatPoints.StatAllocations` — the player's *complete* spread — not the attributes that changed. A per-attribute key would only add rows for targets that always move together.
- **`SelectedSkillsChanged` is the loadout as a unit**, and per-skill would actively be wrong rather than merely redundant. The apply *deselects* every row not in the loadout, so a per-skill key would let an older event's rebuild win on the skills the newer event happened not to mention — producing a loadout the player never selected. Whole-loadout rejection is the only outcome that leaves a coherent state.

The three per-row streams each have a test that fails under a coarser key: log preferences per type (the case the issue names), favorites per item, reads per lesson.

## Stream values 4–8, not 2–3

#2495 is in flight on its own branch and appends `Equipment = 2` / `Mods = 3`. Two branches appending independently both land on 2, and merging them gives one persisted value two meanings — exactly what the enum's append-only rule exists to prevent, and not something a build failure would catch. Reserved with a comment; the gap is free, a collision is not.

`ItemFavorite` is also deliberately its own stream rather than sharing that equipment item key: the two write disjoint columns of the same `UnlockedItems` row, so one shared key would make a favorite toggle reject an equip that isn't stale in any sense that matters.

## Removing the try/catch was checked, not assumed

Each handler's `ChangeTracker.Clear()`-then-re-run is gone in favour of the guard's restart, which is a strict superset (it re-runs the watermark advance too, which the handler-side catch couldn't). PR #2501 found one handler whose delete-then-insert *didn't* collapse onto it; none of these five have that shape:

- The three load-then-upserts (allocations, loadout, favorite) are terminal on restart — after the first successful insert the row exists, so the retry loads it as an update.
- The two update-then-insert fast paths (log preference, lesson read) restart into the update, which now finds the row.

The existing `*_AppliedConcurrently_*` idempotency tests still pass at `Parallelism = 8`, though they apply **unsequenced** and so bypass the comparison entirely — they exercise the guard's outer restart, not the guarded path. One stale comment in that suite that described the handler-side catch is corrected rather than left describing code that no longer exists.

The two self-committing fast paths now run inside the guard's transaction, as the issue requires — automatically, since they issue on the context the guard hands them.

## Test plan

`dotnet build` clean (0 warnings), `dotnet test` **3623 passed / 0 failed** across all four projects. `PlayerWriteWatermarkIntegrationTests` is 34 tests (was 11) and was re-run after the final change.

Per handler: older skipped (and the watermark left alone), equal applies, newer applies and advances, sequence-0 applies against an advanced watermark and leaves it unadvanced. Plus the three per-row independence cases above, and one asserting the five new streams don't block each other despite sharing the `Player` aggregate's counter.

## Two things worth flagging

**#2475 is still not actionable after this.** The issue says this is the last step before it, but that assumed #2495 lands first — the parked-lane deferral still covers the equipment and mod handlers until it does. Not a change in scope here, just a correction to the sequencing note.

**This widens #2500's stream set.** All five are `Player`-produced, so the cold-load seed that issue describes has to cover them too. Its "also worth checking" question — widen per issue, or `MAX` across an aggregate's streams generically — now has a third data point in favour of generic: this is the second PR in a row to add streams to that set. Noted on #2500 rather than pre-empted here.

## Minor fix folded in

`ProgressUpdatedHandler`'s target keys were built with culture-sensitive interpolation. They're persisted comparison keys, so a culture rendering digits differently would write a second watermark row and the guard would silently stop seeing the first — the new keys format invariantly and this one now matches. #2491 (enable the Globalization analyzers) would catch the class of it at build time.

---
_Generated by [Claude Code](https://claude.ai/code/session_015c69jWECQnDEarVd1389Yx)_

### DIFF
--- a/Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs
+++ b/Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs
@@ -1,3 +1,4 @@
+using System.Globalization;
 using Game.Core;
 using Game.Core.Players.Events;
 using Game.DataAccess;
@@ -15,7 +16,7 @@ namespace Game.Application.Tests.DataAccess
 {
     /// <summary>
     /// Verifies the <see cref="PlayerWriteWatermark"/> guard on the absolute-value write-behind handlers
-    /// (#2474): a write older than what its target already holds is skipped instead of durably regressing the
+    /// (#2474, #2496): a write older than what its target already holds is skipped instead of durably regressing the
     /// row, equal sequences still apply, and an unsequenced envelope bypasses the comparison entirely. Each
     /// apply runs through its own DI scope (its own <see cref="GameContext"/>), mirroring the synchronizer's
     /// per-event scope, with the envelope's sequence published onto the scope the way the dispatcher does.
@@ -25,6 +26,9 @@ namespace Game.Application.Tests.DataAccess
     {
         // TestDataSeeder.CreatePlayerAsync's default, asserted where a rolled-back apply must leave the row as seeded.
         private const int SeededLevel = 5;
+
+        // Fixed so a lesson's ReadAt is asserted as an offset from a known instant rather than from "now".
+        private static readonly DateTime LessonUnlockedAt = new(2026, 7, 16, 8, 0, 0, DateTimeKind.Utc);
 
         public PlayerWriteWatermarkIntegrationTests(IntegrationTestContainers containers, ITestOutputHelper testOutputHelper)
             : base(containers, testOutputHelper) { }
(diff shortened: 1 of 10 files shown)


Game.DataAccess/PlayerUpdates/Handlers/ItemFavoriteChangedHandler.cs
```diff
@@ -1,31 +1,25 @@
...
-        }
+        // Keyed per item: each event carries one item's flag, so a per-player key would discard an older
+        // change to item A merely because a newer change to item B landed first. Formatted invariantly
+        // because the key is a persisted comparison key — a culture that renders digits differently would
+        // write a second watermark row and the guard would silently stop seeing the first.
+        public Task HandleAsync(ItemFavoriteChangedEvent evt)
+            => guard.ExecuteGuardedAsync(
+                evt.PlayerId,
+                PlayerWriteStream.ItemFavorite,
+                [evt.ItemId.ToString(CultureInfo.InvariantCulture)],
```
ginderjeremiah: **Should-fix (DRY / footgun):** target-key formatting is now open-coded at four sites in two different idioms — `x.ToString(CultureInfo.InvariantCulture)` here, in `LessonReadHandler.cs:18` and `LogPreferenceChangedHandler.cs:19`, and `FormattableString.Invariant` in `ProgressUpdatedHandler.cs:16-21`. The PR is *fixing* a culture bug that existed precisely because this was open-coded, and then adds three more sites where the next handler author has to remember the rule.

The guard already solved the structurally-identical problem one level up: it hands the callback the `GameContext` rather than documenting "resolve the right one", because "make it impossible to get wrong" beats "write it down" (`PlayerWriteWatermarkGuard.cs:44-49`). Same treatment fits here — a `public static string Target(int id) => id.ToString(CultureInfo.InvariantCulture)` (plus a prefixed overload for the progress keys) on `PlayerWriteWatermarkGuard`, so callers read `PlayerWriteWatermarkGuard.Target(evt.ItemId)` and the invariance is a property of the guard rather than of each caller's discipline.

Worth noting the comment slightly oversells the hazard it defends against: `int.ToString()` doesn't do native-digit substitution in .NET, so for the non-negative ids in play here the only culture-sensitive part is `NegativeSign`. The defence is still right (and satisfies CA1305 for #2491), but the three near-identical five-line comments explaining it are the tell that this wants to be one helper with one comment.

---

Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs
```diff
@@ -270,6 +274,341 @@ await guard.ExecuteGuardedAsync(playerId, PlayerWriteStream.Progress, [target],
...
+            await ApplyAsync(ReadLesson(playerId, readLessonId, readAtMinute: 20), sequence: 2);
+            var rejected = await ApplyAsync(ReadLesson(playerId, otherLessonId, readAtMinute: 5), sequence: 1);
+
+            Assert.Equal(0, rejected);
+            Assert.Equal(LessonUnlockedAt.AddMinutes(5), await ReadLessonReadAtAsync(playerId, otherLessonId));
+            Assert.Equal(LessonUnlockedAt.AddMinutes(20), await ReadLessonReadAtAsync(playerId, readLessonId));
+        }
+
+        [Fact]
+        public async Task PlayerProducedStreams_AreIndependent_SoOneStreamsWriteDoesNotBlockAnOlderWriteOnAnother()
```
ginderjeremiah, line 602: **Should-fix (test coverage):** this PR deletes five bespoke unique-violation catches and replaces them with the guard's restart, and the argument that the restart is terminal is sound — I traced it and agree for all five. But nothing in the suite *pins* it on the guarded path.

The PR is explicit that the existing `*_AppliedConcurrently_*` tests "apply **unsequenced** and so bypass the comparison entirely", which means the sequenced restart — a fresh transaction, a re-run `AdvanceWatermarksAsync`, and a re-decided accepted set — has no coverage at all here. #2501 is the argument for why that matters: it found `ModAppliedHandler`'s delete-then-insert genuinely *didn't* collapse onto the restart, and it found it by writing exactly this test.

The interesting race isn't two copies of the same guarded event (the watermark row serializes those, which is arguably worth asserting on its own), it's a guarded handler against its **unguarded insert-if-missing sibling**, which is the only remaining source of the unique violation the deleted catches existed for:

- sequenced `ItemFavoriteChanged` racing `ItemUnlocked` on the same `(player, item)`
- sequenced `LessonRead` racing `LessonUnlocked` on the same `(player, lesson)`
- sequenced `SelectedSkillsChanged` racing `SkillUnlocked` on a loadout skill

One of those at `Parallelism = 8` would convert "restart is terminal" from a paragraph in the PR body into a regression test. Not blocking — the reasoning holds and the acceptance criteria are all met — but it's the coverage this change earns.

---

Game.DataAccess/PlayerUpdates/Handlers/LogPreferenceChangedHandler.cs
```diff
@@ -1,20 +1,33 @@
...
 namespace Game.DataAccess.PlayerUpdates.Handlers
 {
-    internal sealed class LogPreferenceChangedHandler(GameContext context) : IPlayerUpdateHandler<LogPreferenceChangedEvent>
+    internal sealed class LogPreferenceChangedHandler(PlayerWriteWatermarkGuard guard) : IPlayerUpdateHandler<LogPreferenceChangedEvent>
     {
-        public async Task HandleAsync(LogPreferenceChangedEvent evt)
+        // Keyed per log type: each event carries one type's flag, so a per-player key would discard an older
+        // change to type A merely because a newer change to type B landed first. Formatted invariantly because
+        // the key is a persisted comparison key — a culture that renders digits differently would write a
+        // second watermark row and the guard would silently stop seeing the first.
```
ginderjeremiah: **Nit:** this comment and the `LogPreference = 5` XML doc in `PlayerWriteStream.cs:41-45` say the same thing in nearly the same words ("each event carries one type's flag, so a per-player key would discard an older change to type A merely because a newer change to type B landed first"). Same duplication in `ItemFavoriteChangedHandler` vs `ItemFavorite = 7`, and `LessonReadHandler` vs `LessonRead = 8`.

The granularity rationale belongs on the enum — that's the persisted contract, and it's where someone adding stream 9 will look. Here it can be a line: the enum member name already carries the "why", and the handler only needs to say what's specific to it. That would also bring these closer to CLAUDE.md's 1-2 line guidance without losing the reasoning, which is genuinely worth keeping (just once).

---

ginderjeremiah (commented):
## Review: guard the five remaining absolute-write handlers with `PlayerWriteWatermark` (#2496)

**Verdict:** Approve — no blocking findings; the change does what #2496 asks, and the two granularity judgements the issue asked to be confirmed against the events both hold up under checking. Two should-fixes and a nit below, none of which need to gate the merge. *(Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR's own author — the call is approve.)*

Both "confirm that against the event" items in the issue check out against the code, not just the PR body:

- **`AttributeAllocationsChanged` really is player-scoped.** `Player.cs:128-130` projects `StatPoints.StatAllocations` wholesale, and `PlayerStatPoints.Rehydrate` (`PlayerStatPoints.cs:62-79`) seeds a row for every allocatable attribute — so the event carries the complete spread, zeros included, and there is genuinely nothing finer for a per-attribute key to protect. Correct call.
- **`SelectedSkillsChanged` is the loadout as a unit.** `Player.cs:375` raises the full ordered list, and `SelectedSkillsChangedHandler.ApplyAsync` resets `Selected`/`Order` on *every* row before re-marking the loadout. A per-skill key wouldn't just be redundant — it would let an older event's rebuild win on the skills a newer event didn't mention, producing a loadout the player never selected. The PR's reasoning here is right and the reason it's right is load-bearing.

I also traced the removed try/catches. All five collapse onto the guard's restart: the three load-then-upserts re-load the now-existing row as an update, and the two update-then-insert fast paths restart into an update that now finds its row. Worth adding to the PR's argument — the sequenced path is *also* protected structurally, because `AdvanceWatermarksAsync`'s `INSERT … ON CONFLICT` takes the watermark row lock first, so two applies against the same target serialize and can't collide on the data row at all. The only remaining source of the violation is the unguarded insert-if-missing siblings, which is what makes the missing test below the one worth having.

### Should-fix

- **Target-key formatting is open-coded at four sites in two idioms** — `ItemFavoriteChangedHandler.cs:19`, `LessonReadHandler.cs:18`, `LogPreferenceChangedHandler.cs:19` use `ToString(CultureInfo.InvariantCulture)`; `ProgressUpdatedHandler.cs:16-21` uses `FormattableString.Invariant`. This PR is fixing a culture bug caused by exactly that open-coding, then adds three more sites of it. A `PlayerWriteWatermarkGuard.Target(int)` helper makes the invariance structural — the same argument the guard already makes for handing the callback its `GameContext`. Detail inline.
- **No guarded-path concurrency coverage for the five handlers whose error handling was deleted.** The PR itself notes the existing `*_AppliedConcurrently_*` tests apply unsequenced and bypass the comparison, so the sequenced restart is untested. #2501 is the case for why that matters — it found a handler that *didn't* collapse onto the restart, via precisely this test. Detail inline, with the three races worth writing.

### Nits

- The per-handler granularity comments duplicate their `PlayerWriteStream` XML docs almost verbatim (three cases). Keep the rationale on the enum — that's the persisted contract. Detail inline.
- `PlayerWriteWatermarkIntegrationTests` is now 829 lines / 34 tests across seven streams, and #2501 adds more. Not a problem yet, but it's on a trajectory — nested per-stream classes over the shared helpers would keep it navigable. Fine as a follow-up when the next stream lands rather than here.

### Test coverage

Every acceptance criterion in #2496 is met, per handler: older-sequence skipped with the watermark left alone, equal applies, newer applies and advances, sequence-0 applies against an advanced watermark without advancing it. The three per-row independence tests are there, including the `LogPreference` per-log-type case the issue names explicitly, and each would genuinely fail under a coarser key. `PlayerProducedStreams_AreIndependent_…` is a good addition the issue didn't ask for — it pins the thing that's easy to get wrong once every stream shares one producer counter. The only gap is the one above.

The corrected comment in `PlayerUpdateHandlerIdempotencyIntegrationTests.cs:732-737` is the right instinct: it would otherwise have kept describing a catch block that no longer exists.

### What's good

- **The stream-value reservation is the right call, for the right reason.** Two branches appending independently both landing on `2` is not something the build would catch and *is* something that silently gives one persisted value two meanings. Burning two enum values to avoid that is free; the collision isn't. Whichever of this and #2501 merges second should still re-confirm the values didn't drift during the merge.
- **`ItemFavorite` as its own stream rather than sharing #2495's item key** is a subtle call and correct — disjoint columns of the same row shouldn't reject each other.
- The `ProgressUpdatedHandler` culture fix is genuinely in scope (it makes the existing keys consistent with the new ones) and small enough not to muddy the diff.
- The PR body flags that #2475 is *not* actionable after this and that this widens #2500's stream set, rather than repeating the issue's sequencing claim. Correcting the issue you're closing is the right behaviour.

---

ginderjeremiah:
All three points addressed in a7ac9d7.

**Target-key formatting → `PlayerWriteWatermarkGuard.Target`.** Agreed, and the parallel you drew is the right one — this is the same "make it impossible to get wrong" argument as passing the callback its `GameContext`. Two overloads: a bare `Target(int)` for the single-family streams and `Target(string kind, params ReadOnlySpan<int?>)` for the progress keys, which also absorbs the `stat:{typeId}:{entityId}` shape (a null id renders as an empty segment, which is how a global statistic's absent entity id was already spelled). All four sites now go through it, and the invariance rationale is stated once on the helper. Your correction to my comment is right — `int.ToString` doesn't do native-digit substitution, so `NegativeSign` is the actual exposure; the helper says that instead of what I'd written.

**Guarded-path concurrency.** Added three, one per apply shape, and your framing picked the right race — two sequenced applies of the same target queue on the watermark row and can't collide at all, so the sibling is the only remaining source: `ItemFavorite` vs `ItemUnlocked` (load-then-upsert), `LessonRead` vs `LessonUnlocked` (update-then-insert fast path), `SelectedSkills` vs `SkillUnlocked` (multi-row rebuild). Each runs both applies 4× interleaved at `Parallelism = 8`. They're probabilistic by nature — the assertions are on convergence, which must hold whether or not the violation fires — so the docstring points at the deterministic `GuardedApply_UniqueViolationInsideTheTransaction_…` as the one that actually pins the restart.

**Comment duplication.** Trimmed the three per-row handler comments to a line pointing at their `PlayerWriteStream` member, and did the same to the two player-scoped handlers, which had the same problem. What stayed is handler-specific: the `StatPointsUsed` disagreement, and the guard owning the transaction/context/restart.

`dotnet build` clean, full suite **3626 passed / 0 failed**; the watermark suite (now 37) was run 3× for flake and was stable.

On the merge-order note: agreed. Whichever of this and #2501 lands second should re-confirm `PlayerWriteStream` values didn't drift — this one deliberately holds 2 and 3 open for #2495, so a conflict resolution that renumbers is the thing to watch for.

---

Game.DataAccess/PlayerUpdates/Handlers/ItemFavoriteChangedHandler.cs
ginderjeremiah: Addressed in a7ac9d7. All four sites now go through `PlayerWriteWatermarkGuard.Target` (`ItemFavoriteChangedHandler.cs:15`, `LessonReadHandler.cs:15`, `LogPreferenceChangedHandler.cs:15`, `ProgressUpdatedHandler.cs:14-18`), and `git grep InvariantCulture` over `Game.DataAccess`/`Game.Infrastructure` leaves no open-coded target formatting behind — the only remaining hits are the guard's own two overloads and the unrelated `PlayerProgressRepository` sequence field.

I checked the progress keys byte-for-byte rather than trusting the refactor: `Target("stat", typeId, entityId)` reproduces `$"stat:{typeId}:{entityId}"` exactly, including a null `EntityId` rendering as a trailing empty segment, so no persisted watermark row is orphaned. `GuardedApply_UniqueViolationInsideTheTransaction_…` independently pins that format with a literal `$"stat:{(int)EStatisticType.EnemiesKilled}:"`, which is the right way for it to be pinned.

The `<remarks>` also corrects the overstated hazard — it now says the negative sign specifically rather than implying digit substitution. Resolving.

---

Game.DataAccess/PlayerUpdates/Handlers/LogPreferenceChangedHandler.cs
ginderjeremiah: Addressed in a7ac9d7. All four handler comments now point at the enum instead of restating it — `ItemFavoriteChangedHandler.cs:10`, `LessonReadHandler.cs:10`, `LogPreferenceChangedHandler.cs:10`, and `SelectedSkillsChangedHandler.cs:10-12` / `AttributeAllocationsChangedHandler.cs:10-13` for the two player-scoped ones.

The trim kept the right things in the right places: the handler-specific consequence stays on the handler (the allocations one still says the `StatPointsUsed` disagreement out loud, which is the fact that motivates guarding it at all), while the granularity contract lives only on `PlayerWriteStream`. Resolving.

---

Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs
ginderjeremiah: Addressed in a7ac9d7, and addressed as the *right* scenario rather than the convenient one — all three new tests race a guarded handler against its **unguarded insert-if-missing sibling**, which is the only interleaving that still reaches the data row now that same-target sequenced applies queue on the watermark. That reasoning is stated in the test comments too, so the next reader doesn't re-derive it.

One per apply shape, which is what makes the set complete rather than repetitive:
- `GuardedItemFavorite_…` — load-then-upsert, and it asserts `CountUnlockedItemsAsync == 1`, so a restart that inserted a duplicate instead of loading the existing row would fail rather than silently pass on the flag check alone.
- `GuardedLessonRead_…` — the update-first fast path whose miss falls through to an insert, i.e. the window the restart has to close.
- `GuardedSelectedSkills_…` — the multi-row rebuild.

I traced the flakiness risk both ways before calling this done. The siblings all still absorb their own violations (`InsertIfMissingAsync`'s catch for item/skill, `LessonUnlockedHandler`'s explicit one), so the unguarded half can't fail the test spuriously; and the lock graph has no cycle — the guarded side takes the watermark row then the data row, the unguarded side takes only the data row — so these can't deadlock into flakes either. Resolving; one note on `RaceAsync`'s doc goes in the review body.

---

Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs
```diff
@@ -270,6 +278,411 @@ await guard.ExecuteGuardedAsync(playerId, PlayerWriteStream.Progress, [target],
...
+        /// <summary>
+        /// Runs both applies <see cref="Parallelism"/> times each, all at once, so several pass their
+        /// existence check before any insert commits. Whether the violation actually happens is a timing
+        /// race — the assertions are on convergence, which must hold either way — so this is a probabilistic
+        /// guard on top of <c>GuardedApply_UniqueViolationInsideTheTransaction_…</c>, which forces the same
+        /// restart deterministically.
+        /// </summary>
+        private static Task RaceAsync(Func<Task<int>> guarded, Func<Task<int>> unguardedSibling)
+            => Task.WhenAll(Enumerable.Range(0, Parallelism)
+                .Select(i => i % 2 == 0 ? guarded() : unguardedSibling()));
```
ginderjeremiah: **Nit (factual, not stylistic):** the doc says "Runs both applies `Parallelism` times each", but `Enumerable.Range(0, Parallelism)` with the `i % 2` split produces `Parallelism` tasks *total* — 4 guarded and 4 unguarded, not 8 each. The comment on the const compounds it: it justifies 8 as "enough concurrent applies that several pass their existence check before any insert commits" and says it "mirrors the idempotency suite", but each side here only gets half of the value that reasoning picked.

That halving matters more than a normal comment slip because it's the *only* knob on a deliberately probabilistic test — the narrower the window, the more often this passes without the interleaving it exists to catch, and it degrades silently rather than failing. A one-line fix keeps the const's stated meaning:

```csharp
=> Task.WhenAll(Enumerable.Range(0, Parallelism)
    .SelectMany(_ => new[] { guarded(), unguardedSibling() }));
```

Either that or reword the doc to "`Parallelism` applies in total, split evenly" — but I'd take the code change, since the const's own comment is the one that reads as 8-per-side.

Worth saying explicitly that this isn't a knock on the design: pairing a probabilistic race with `GuardedApply_UniqueViolationInsideTheTransaction_…`, which forces the same restart deterministically, and *saying so in the doc* is exactly the right structure. A race test that can silently stop racing is only acceptable when something else pins the behaviour, and here something does. The assertions being on convergence rather than on the violation having occurred is also correct — the opposite would be the flaky version.

---

ginderjeremiah (commented):
## Re-review: a7ac9d7 — target-key formatting centralized, guarded-path concurrency covered, granularity comments trimmed

**Verdict (intended): Approve** — all three actionable findings from the previous round are genuinely resolved, verified against the new code rather than the commit message. Submitted as `COMMENT` only because GitHub won't accept a gating verdict from the PR's own author.

I re-reviewed the delta against c01593d and kept a light pass over the rest; nothing the earlier round approved has regressed.

### Since the last review

- ✅ **Target-key formatting open-coded at four sites in two idioms** — resolved on `PlayerWriteWatermarkGuard.cs:50-69`. The four call sites now go through `Target`, and a `git grep` over `Game.DataAccess`/`Game.Infrastructure` leaves no open-coded target formatting behind. I verified the persisted format is byte-identical rather than assuming: `Target("stat", typeId, entityId)` reproduces `$"stat:{typeId}:{entityId}"` including a null `EntityId` rendering as a trailing empty segment, so no existing watermark row is orphaned by the refactor — and `GuardedApply_UniqueViolationInsideTheTransaction_…` pins that format with an independent literal. The `<remarks>` also drops the overstated digit-substitution claim for the accurate one (the negative sign, CA1305).
- ✅ **No guarded-path concurrency coverage** — resolved. Three new tests at `PlayerWriteWatermarkIntegrationTests.cs:616-665`, and they race the *right* thing: a guarded handler against its **unguarded insert-if-missing sibling**, which is the only interleaving that still reaches the data row now that same-target sequenced applies queue on the watermark row. One per apply shape (load-then-upsert, update-first-then-insert, multi-row rebuild), which is what makes the set complete rather than three spellings of one case. Detail on the thread.
- ✅ **Per-handler granularity comments duplicating their `PlayerWriteStream` docs** — resolved across all five handlers. The trim landed the split in the right place: the contract lives only on the enum, while the handler keeps what's specific to it (the allocations handler still states the `StatPointsUsed` disagreement, which is the fact that justifies guarding it at all).
- ↩️ **`PlayerWriteWatermarkIntegrationTests` size** — carried forward as a follow-up as agreed, not a finding on this PR. Now 903 lines / 37 tests, and #2501 adds a further ~343 lines to the same file. Filed as #2504.

### New findings

- **Nit** — `PlayerWriteWatermarkIntegrationTests.cs:667-676`: `RaceAsync`'s doc claims `Parallelism` applies *each*; the `i % 2` split delivers `Parallelism` in total, 4 per side. Inline, with the one-line fix.

Nothing else. I specifically went looking for problems the new commit could have introduced and didn't find them:

- **Flakiness, both directions.** The unguarded halves all still absorb their own unique violations (`InsertIfMissingAsync`'s catch for item/skill, `LessonUnlockedHandler`'s explicit one), so the sibling can't fail these spuriously. And the lock graph has no cycle — the guarded side takes the watermark row then the data row, the unguarded side takes only the data row — so they can't deadlock into intermittent failures either.
- **Assertion strength.** `GuardedItemFavorite_…` asserts `CountUnlockedItemsAsync == 1`, so a restart that inserted a duplicate instead of loading the existing row fails rather than passing on the flag check alone. That's the assertion that makes the test worth having.
- **`Target`'s API shape.** The `int?` params overload correctly renders a null id as an empty segment (the global-statistic case) instead of dropping it, which is what preserves compatibility with the already-persisted keys. `StringBuilder` over `FormattableString.Invariant` is also a small win on the highest-volume path — no boxing, no `FormattableString` allocation.
- **Test helpers deliberately *not* routed through `Target`.** `LogTarget`/`ItemTarget`/`LessonTarget` (`:719-721`) still spell the format by hand, with a comment saying why. That's correct and I want to be explicit that it's correct, because it looks like a missed call site: a test that computed its expected key with the production formatter couldn't catch that formatter changing. Please don't "finish the refactor" here later.

### Merge-order interaction with #2501 — read before merging either

The reserved `2`/`3` values do their job: #2501 fills them in exactly as intended, so there's **no semantic enum collision**, only a trivial textual conflict where the reserved-comment block sits. But this commit introduced **two genuine semantic interactions that did not exist at the time of the last review**, both worth resolving deliberately rather than at conflict-resolution time:

1. **Two independently-invented target-key formatters.** This PR adds `PlayerWriteWatermarkGuard.Target`; #2501 adds a separate `PlayerWriteTargets` class — and #2501's builders are culture-sensitive interpolation (`$"item:{itemId}"`), i.e. precisely the pattern this PR just centralized away. Merging both re-creates the finding this round closed, in a form no build failure will catch. Whichever lands second should consolidate, and I'd argue **`PlayerWriteTargets` is the better home** — the guard is the mechanism, the target format is the stream's contract, and per-stream naming reads better at the call site — with `Target` demoted to the private invariant-formatting primitive it delegates to. One wrinkle to handle rather than discover: `Target("item", id)` and `Target("slot", id)` reproduce #2501's keys exactly, but its mods key `$"{itemId}:{modSlotId}"` has no prefix, and `Target("", a, b)` yields `":a:b"`, not `"a:b"` — so the consolidation needs a prefix-less variadic overload (or the mods key needs a prefix, which would be a persisted-format change). Filed as #2505.
2. **#2501 rewrites the enum's summary to say handlers writing the *same rows* must share a stream** — and `ItemFavorite = 7` visibly doesn't, since it writes the same `UnlockedItems` row as `Equipment = 2`. I think this PR's call is right (disjoint columns; ordering a favorite toggle against an equip is meaningless), but post-merge the enum will state a rule its own member breaks. Whoever merges second should add the one-line exception to that summary. Also noted on #2505.

Neither is a reason to hold this PR, and the order doesn't matter much — the second one merged just inherits both cleanups.

### Test coverage

Meets the project's backend rules. Integration coverage for the out-of-process dependency (Postgres) is where it belongs, the per-handler matrix from #2496 is complete (older skipped, equal applies, newer advances, sequence-0 applies without advancing, plus per-row independence for the three fine-grained streams), and the guarded restart now has both a deterministic test and a probabilistic one. No frontend or battle-logic surface here, so no mirroring obligation.

### What's good

The response to review didn't just satisfy the letter of the three points. The concurrency coverage in particular re-derived *which* race is still reachable post-guard instead of writing the obvious two-copies-of-the-same-event test that the watermark row would have made vacuous — and then wrote the reasoning into the test comments so the next reader gets it for free. The `Target` refactor being checked for format-compatibility against already-persisted keys, rather than treated as a pure rename, is the other thing I'd have wanted and didn't have to ask for.

---

ginderjeremiah:
Nit fixed in 3888120 — took the code change rather than the reword, for the reason you gave: the constant's comment is the one that reads as 8-per-side, and a probabilistic test that quietly halves its own window is worse than a wrong comment.

```csharp
=> Task.WhenAll(Enumerable.Range(0, Parallelism)
    .SelectMany(_ => new[] { guarded(), unguardedSibling() }));
```

Suite re-run 3× after the change (37 tests, stable at the doubled concurrency), full backend suite **3626 passed / 0 failed**.

Noted on the two merge-order interactions, and agreed on both calls:

- **`PlayerWriteTargets` is the better home** — the guard is the mechanism, the key format is the stream's contract. The prefix-less mods key is the real wrinkle: `Target("", a, b)` yields `":a:b"`, so consolidation needs a variadic overload with no prefix rather than an empty one, and changing the mods key to carry a prefix would be a persisted-format change on rows #2501 will already have written. #2505 has it.
- **The enum-summary exception is worth adding explicitly** rather than leaving `ItemFavorite` as a visible counterexample to a rule stated one line above it. Disjoint columns of a shared row is the distinction — same row, but ordering a favorite toggle against an equip is meaningless.

Whichever of this and #2501 lands second, I'll pick both up as part of the conflict resolution rather than leaving them to #2505 if it's this one.

Also taking the point about `LogTarget`/`ItemTarget`/`LessonTarget` staying hand-spelled — they're there precisely so a change to the production formatter fails a test instead of being computed by it. The comment says so, but it's the kind of thing a later "finish the refactor" pass would undo, so having it stated in the review record too is useful.

---

Game.Application.Tests/DataAccess/PlayerWriteWatermarkIntegrationTests.cs
ginderjeremiah: Addressed in `3888120` — the code fix rather than the doc reword, which was the option I'd have picked too, since the const's own comment was the one reading as 8-per-side.

`RaceAsync` is now `SelectMany(_ => new[] { guarded(), unguardedSibling() })` over `Enumerable.Range(0, Parallelism)`, so each side genuinely gets 8 and the doc at `PlayerWriteWatermarkIntegrationTests.cs:668` and the const comment at `:30-32` are both true as written. No latent double-start: `Task.WhenAll(IEnumerable<Task>)` materializes its argument once, so the lazy `SelectMany` yields 16 tasks, not 32.

Verified the doubled window didn't turn the probabilistic tests flaky, which was the one real risk of this commit — that's the failure mode you'd want caught here rather than in CI a week from now. Ran the class 12 times locally against the shared Postgres/Redis (10 targeted runs plus two full-suite passes): 37/37 every time, no unique violations escaping the restart and no deadlocks. Consistent with the lock-graph argument in the thread above — the guarded side takes the watermark row then the data row, the unguarded side takes only the data row, so there's no cycle to widen. Resolving.

---

ginderjeremiah:
## Re-review: `3888120` — `RaceAsync` now applies each side `Parallelism` times

**Intended verdict: Approve.** The one outstanding nit from the last round is fixed, correctly and in the way that was preferred; nothing else changed and nothing regressed. Posting as a comment rather than an `APPROVE` review because GitHub won't accept a gating verdict from the PR's own author.

### Since the last review

- ✅ **`RaceAsync` doc/behaviour mismatch** (`PlayerWriteWatermarkIntegrationTests.cs:674-676`) — fixed in `3888120`, and fixed as the code change rather than the doc reword. `SelectMany(_ => new[] { guarded(), unguardedSibling() })` over `Enumerable.Range(0, Parallelism)` yields 8 of each, so both the method's `<summary>` and the `Parallelism` const's own comment at `:30-32` ("enough concurrent applies that several pass their existence check before any insert commits") are now true as written instead of describing half the value. Thread resolved.

Everything else in the PR is byte-identical to what was reviewed at `a7ac9d7` — `git diff a7ac9d7..3888120a` is that one line. The three earlier findings stay resolved.

### New findings

None. Two things I specifically went looking for and didn't find:

- **No latent double-start.** `SelectMany` is lazy, so the obvious hazard is `Task.WhenAll` enumerating twice and firing 32 applies. It materializes its `IEnumerable<Task>` once, so this is 16 tasks. Worth having checked rather than assumed, since a silent doubling here would have been invisible in a passing test.
- **No new flakiness at the doubled window.** This is the real risk of the commit — widening a deliberately probabilistic race can convert "passes" into "usually passes". Ran `PlayerWriteWatermarkIntegrationTests` 12 times locally against Postgres/Redis (10 targeted, plus two full-suite passes): 37/37 green every run, no unique violation escaping the guard's restart, no deadlock. That matches the lock-graph reasoning already on the thread — the guarded side takes the watermark row then the data row, the unguarded sibling takes only the data row, so doubling the contenders widens the window without introducing a cycle.

### Verification

- `dotnet build` — clean, **0 warnings**.
- `dotnet test`, per project: Core 1492, Infrastructure 74, Application 1167, Api 893 — **3626 passed / 0 failed**. (Running all four projects concurrently against one shared Postgres does produce failures in `Game.Api.Tests`; that's this sandbox's shared-container setup, not the PR. Run on its own the project is 893/893, and CI is green.)
- CI on `3888120a`: all 12 checks green, `all-checks-passed` ✅.
- **Reserved enum values `2`/`3` confirmed still intact and unclaimed** (`PlayerWriteStream.cs:28-30`), so #2501 can still append `Equipment = 2` / `Mods = 3` onto whichever of the two merges second. Merging this PR will give #2501 a textual conflict on the enum; that's expected and was accepted last round.

### Follow-ups (both open, verified accurate — nothing new filed)

- **#2504** — split this test file. Its stated "903 lines / 37 tests after #2502" matches the branch exactly.
- **#2505** — consolidate the two target-key formatters and add the `ItemFavorite = 7` exception to the enum summary, once both this and #2501 land.
- **#2503** was filed from #2501's side for the same concern and is already closed as a duplicate of #2505 — so there's one issue covering that cleanup, not two. Its extra detail (the `Mods` key's missing prefix producing `:7:3`, and the orphaned-watermark trap that implies) is the part worth carrying forward when #2505 is picked up; #2505 does record it.

Neither follow-up is work that belonged in this PR: #2504 is explicitly better done after both branches stop adding to the file, and #2505 is cross-PR by construction — invisible on either branch alone.

### What's good

The fix is the smallest correct one and left the const's comment as the thing that defines intent, rather than editing the comment to match the weaker behaviour. Worth restating what makes this test structure sound: a probabilistic race is only acceptable when something deterministic pins the same behaviour, and `GuardedApply_UniqueViolationInsideTheTransaction_…` does — with the `<summary>` saying so, so nobody later mistakes the race for the actual guarantee.

Merging.

---

PR #2506, opened 2026-07-28
Report accepted-but-unchanged separately from rejected on stat allocation (#2485)

### LINKED ISSUE #2485: UpdatePlayerStats treats a no-op allocation payload as a real change, raising a redundant domain event and write-behind save
Noticed while reviewing #2484 (pre-existing, not introduced there).

## The rule

`docs/backend.md`:

> **Rejected and accepted-but-unchanged are distinct outcomes, and a command must not conflate them.** A request that mutated nothing because the state already matched … is *accepted* — it reports success to the client — but raises no domain event and enqueues no write-behind save, so a client spamming an identical toggle can't generate redundant saves.

## Where it doesn't hold

`PlayerStatPoints.TryUpdateAttributes` returns a bare `bool` that means "not rejected", not "changed". Two payloads mutate nothing and still return `true`:

- an **empty** update list (`matchedUpdates` empty → `changedPoints == 0` → the guard passes vacuously), and
- a set of **all-zero deltas** (`new Update(EAttribute.Strength, 0)`).

`Player.TryUpdateAttributes` (`Game.Core/Players/Player.cs:90-102`) then unconditionally calls `RaiseCoreUpdated()` and raises `AttributeAllocationsChangedEvent` with the player's whole allocation list, and `PlayerService.TryUpdateAttributes` → `SaveIf` enqueues the write-behind save. Nothing upstream rejects an empty `List<AttributeUpdate>` either — `UpdatePlayerStats` passes `Parameters` straight through.

So a client looping `UpdatePlayerStats` with `[]` generates one full-allocation-set write-behind event per call — exactly the redundant-save spam the documented rule exists to prevent.

## Suggested change

Follow the pattern the doc describes and that the other commands already use: have the domain method return which outcome occurred rather than a bare bool (the repo's existing shape for this — see the `SetLogPreferences` / favorite-item / idle-mode paths), and have the application layer persist only on a real change while still reporting success on the unchanged path. "Changed" here is `changedPoints != 0` — a set of deltas summing to zero across different attributes (e.g. `+2 STR / −2 END`) *is* a real change and must still persist, so the check has to be per-update, not on the sum.

Worth confirming while implementing whether `RaiseCoreUpdated()` should also be suppressed on the unchanged path (it drives the core-updated push), or only the persistence event.

## Scope

Small — one domain method's return type, its caller in `Player`, `PlayerService.TryUpdateAttributes`, plus unit tests for the empty-payload and all-zero-delta cases and a test pinning that a zero-sum-but-nonzero-deltas set still persists.

ginderjeremiah:
Closes #2485.

Both no-op payloads the issue names check out against the code: an **empty** update list passes the available-points guard vacuously (`changedPoints == 0`, and `.All()` on an empty set is true), and a set of **all-zero deltas** applies `+= 0` to every row. Both returned `true`, so `Player.TryUpdateAttributes` raised `PlayerCoreUpdatedEvent` + `AttributeAllocationsChangedEvent` and `SaveIf` enqueued a write-behind save — one full-allocation-set event per call for a client looping `UpdatePlayerStats` with `[]`.

Fixed the way `docs/backend.md` prescribes and `SetFavoriteOutcome` already demonstrates: the domain method returns which of the three outcomes occurred, and the application layer maps it — persist only on `Changed`, fail only on `Rejected`.

## The "changed" test is per-update, and covers one case the issue doesn't

Per-update rather than on the summed points, exactly as the issue requires — `+2 STR / −2 END` nets zero spend but genuinely reallocates, and there's a test pinning it as `Changed`.

The case the issue doesn't mention: a **zero delta on an attribute with no row** creates one, and that row enters the persisted allocation set even at zero. That's a real change to the aggregate, so it reports `Changed` rather than being classed a no-op. It isn't a spam vector — the row exists from then on, so the identical repeat is `Unchanged`, and a test asserts both halves in sequence. In practice it's near-unreachable anyway (`CreateAllocations` seeds the complete spread and `Rehydrate` repairs it), but classing it `Unchanged` would have meant an in-memory mutation the DB never learns about, which is the sort of divergence worth not introducing.

## Answering the `RaiseCoreUpdated()` question

The issue asks whether the core-updated raise should also be suppressed. **Yes — both are suppressed.** `PlayerCoreUpdatedEvent` isn't a client push despite the "core-updated push" framing: its only consumers are `PlayerCoreUpdatedHandler` (the write-behind DB write) and `PlayerPersistencePublisher` (the queue). It is purely a persistence event, and `StatPointsUsed` is untouched on the unchanged path, so raising it would enqueue a save that rewrites the row's existing contents. The command's own response is what reaches the client, and it is unaffected.

## Renames

`Try*` returning an enum would be misleading, so `PlayerStatPoints.TryUpdateAttributes` / `Player.TryUpdateAttributes` / `PlayerService.TryUpdateAttributes` become `UpdateAttributes` — the precedent `SetFavorite` already sets. Stale `<see cref>`s and comments naming the old method (`GameConstants`, `StatAllocation`, `attributes-view.svelte.ts`) are updated rather than left dangling.

## Not a live UI bug

Worth stating since it changes how urgent this is: the real client can't trigger it. `attributes-view.svelte.ts`'s `changedUpdates` already filters `deltas[i] !== 0`, and `save()` returns early on an empty set. This is a hostile-or-buggy-client vector, which is what the anti-cheat/no-redundant-save contract is for. No frontend behaviour change, and no battle-logic parity implications.

## Test plan

`dotnet build` clean (0 warnings), `dotnet test` **3611 passed / 0 failed** across all four projects. Frontend `npm run lint` clean (1230 files, 0 errors, 0 warnings) — the only UI change is one identifier inside a comment.

New coverage:

- **`PlayerStatPointsTests`** — empty payload, single zero delta on an existing row, and every-delta-zero across several attributes are each `Unchanged`; deltas summing to zero are `Changed`; the zero-delta-creates-a-row case is `Changed` and its repeat `Unchanged`. The pre-existing `ZeroAmount_Succeeds` was asserting the old behaviour and is retargeted.
- **`PlayerTests`** — a new `UpdateAttributes` section asserting which events are raised: both on a real change and on a zero-sum reallocation, none at all on the empty, all-zero, and rejected payloads.
- **`PlayerWriteBehindTests`** — the end of the chain, as a `[Theory]` over 0/1/2 zero deltas: the no-op is accepted (`true`) yet the Redis queue stays empty, and a real allocation on the same player immediately after still publishes both events, so the guard suppresses only the no-op.

Verified the tests actually pin the fix rather than merely passing alongside it: forcing `changed = true` (the old behaviour) fails 6 of them.

One thing that needed care in the new aggregate tests — `PlayerBuilder` defaults `StatAllocations` to `[]`, so a player built the usual way has no rows and an "all-zero" payload would *create* them and correctly report `Changed`. The tests use a helper seeding `CreateAllocations()`, the complete spread every real character carries.

---
_Generated by [Claude Code](https://claude.ai/code/session_014XumjVgTr5gpTxTojSSsNQ)_


Game.Core/Players/UpdateAttributesOutcome.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>
+        /// The payload allocated nothing — it was empty, or every delta was zero: accepted, but nothing
+        /// mutated.
+        /// </summary>
+        Unchanged,
+
+        /// <summary>The allocations actually moved, so the change needs an event and a save.</summary>
+        Changed,
+    }
```
ginderjeremiah, line 25: **Follow-up, not a change request for this PR** — filed as #2507.

This is structurally identical to `SetFavoriteOutcome` (`Game.Core/Players/Inventories/SetFavoriteOutcome.cs`) down to the member names and the shape of the XML docs, and both are consumed by the same four lines in `PlayerService`. Following the established precedent was the right call here — #2485 is scoped small, and generalizing would have dragged the favorite path into the diff.

Worth recording why it's worth revisiting rather than leaving at two: the hand-written mapping has a silent, client-visible failure mode. `return outcome == Changed` instead of `!= Rejected` reports a legitimate no-op as a failure to the client — the exact conflation `docs/backend.md:118` exists to prevent — and neither the compiler nor the existing suites would catch it on a *new* command. A shared `MutationOutcome` plus a `SaveIf(Player, MutationOutcome, …)` overload makes it unwritable. #2507 also covers the two other spellings already in the tree (`BattleService.SetAutoChallengeBoss`, `PlayerService.SaveLogPreferences`).

---

Game.Core.Tests/Players/PlayerTests.cs
```diff
@@ -1441,6 +1518,19 @@ private static SkillRecipe MakeRecipe(
...
+        /// repaired on rehydration) plus points to spend — so an all-zero payload is genuinely a no-op rather
+        /// than one that creates the missing rows.
+        /// </summary>
+        private static Player MakeAllocatingPlayer(int gained = 10) =>
+            new PlayerBuilder()
+                .WithStatAllocations(PlayerStatPoints.CreateAllocations())
+                .WithStatPointsGained(gained)
+                .Build();
+
+        private record AttributeUpdate(EAttribute Attribute, int Amount) : IAttributeUpdate;
```
ginderjeremiah: **Nit** — this is the fourth private `IAttributeUpdate` stub in the test tree, and the third distinct name for it:

- `PlayerStatPointsTests.cs:423` — `record Update`
- `PlayerServiceIntegrationTests.cs:906` — `record SimpleAttributeUpdate`
- `PlayerWriteBehindTests.cs:430` — `record SimpleAttributeUpdate` (byte-identical to the one above, same assembly)
- here — `record AttributeUpdate`

Three of those were already here, so this isn't a problem the PR created, and `Game.Core.Tests` can't reach `Game.Application.Tests`'s copy anyway. Two things make it worth a note rather than nothing: `AttributeUpdate` shadows the real contract type `Game.Api.Models.Attributes.AttributeUpdate`, so a reader landing on this file has to check which one is in play — `Update` (the sibling suite's name) or `SimpleAttributeUpdate` would read unambiguously. And the two `Game.Application.Tests` copies are genuinely redundant with each other and belong in `Game.TestInfrastructure` alongside the other shared helpers.

Fine to leave as-is; a rename here is the whole cost if you'd rather not touch the other suites.

---

ginderjeremiah (commented):
## Review: report accepted-but-unchanged separately from rejected on stat allocation (#2485)

**Verdict (intended): Approve** — no blocking findings and no should-fixes. The change does what #2485 asks, in the shape `docs/backend.md:118` prescribes, and the two judgement calls the issue left open both check out against the code. One nit and one follow-up below, neither of which needs to gate merge. *(Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR's own author — the call is approve.)*

- **Both no-op payloads were real.** An empty list leaves `matchedUpdates` empty, so `changedPoints == 0` clears the budget guard and `.All()` is vacuously true; an all-zero set applies `+= 0` to every matched row. Both reached `Player.TryUpdateAttributes`'s unconditional `RaiseCoreUpdated()` + `AttributeAllocationsChangedEvent` and `SaveIf`'s enqueue. The issue's premise holds exactly as stated.
- **The `RaiseCoreUpdated()` question is answered correctly, and the answer is checkable.** `PlayerCoreUpdatedEvent` has exactly two consumers in the tree — `PlayerCoreUpdatedHandler` (the write-behind DB write) and `PlayerPersistencePublisher` (the queue). Nothing pushes it to a client, so despite the "core-updated push" framing in the issue it is purely a persistence event and suppressing it loses nothing the client can observe. The supporting fact also holds structurally rather than by inspection: `changed == false` implies every matched delta is zero, so `changedPoints` is provably `0` and `StatPointsUsed += 0` — there is no in-memory mutation left un-persisted on the `Unchanged` path.
- **The rename left nothing dangling in code.** `TryUpdateAttributes` has zero hits across `*.cs` / `*.ts` / `*.svelte` at head. (Docs: see the nit.)
- **"Not a live UI bug" is accurate.** `attributes-view.svelte.ts:438` filters `deltas[i] !== 0` into `changedUpdates` and `:447` returns early on an empty set, so the real client cannot produce either payload. Correctly scoped as a hostile-or-buggy-client vector, which is what makes a frontend test unnecessary here rather than missing.
- **The created-row-counts-as-`Changed` call is right**, and for the reason the PR gives rather than the obvious one: classing it `Unchanged` would leave an in-memory row the database is never told about. Also worth noting it's near-unreachable in the first place — `Rehydrate` (`PlayerStatPoints.cs:62-79`) restores a row for every core attribute on load, so a live player never has a missing one to create.

I went looking for ways the `Unchanged` path could still leak a write and didn't find one: `createdAllocations` is empty whenever `changed` is false (a created row forces `Changed`), so `AddRange` is a no-op, and `allocation.Amount += 0` can't move a double. The five-line comment at `PlayerStatPoints.cs:128-132` is the one place I'd normally trim, and I'd keep it — "the test is per-update rather than on `changedPoints`" is precisely the invariant a later "simplification" would break, and the `+2 STR / −2 END` counterexample is what stops that edit.

### Nits

- **`docs/spikes/1126-class-system.md` still names `PlayerStatPoints.TryUpdateAttributes`** at lines 30, 51 and 84. This isn't frozen history — area D is unimplemented forward-looking plan ("the per-level point grant split so only the reduced free pool flows through `PlayerStatPoints.TryUpdateAttributes`"), so whoever picks it up greps for a method that no longer exists. The PR deliberately chased down every *other* stale reference (`GameConstants`, `StatAllocation`, `attributes-view.svelte.ts`), which is what makes leaving these three read as an oversight rather than a decision. Either update them or say why the spike record is left as-written.
- `Game.Core.Tests/Players/PlayerTests.cs:1532` — fourth `IAttributeUpdate` stub, third name for it, and this one shadows the real `Game.Api.Models.Attributes.AttributeUpdate`. Detail inline; mostly a pre-existing duplication.

### Follow-up

- **#2507** — the accepted-but-unchanged mapping is now written by hand in three different shapes across the tree (this enum + `SetFavoriteOutcome`, `BattleService.SetAutoChallengeBoss`'s bare bool, `SaveLogPreferences`'s accumulated bool). Deliberately *not* folded into this PR: consolidating means touching the favorite path and #2485 is scoped small. Detail on the inline thread.

### Test coverage

Meets the project's backend rules and covers a case the issue didn't ask for. Domain logic is unit-tested classically at both levels — `PlayerStatPointsTests` for which outcome each payload produces (empty, single zero on an existing row, all-zero across several attributes, zero-sum-but-nonzero deltas, and the create-a-row case with its `Unchanged` repeat asserted in sequence), and `PlayerTests` for the consequence that actually matters, which events are raised. Splitting those two concerns across the two suites is the right seam: the outcome enum is `PlayerStatPoints`' contract, the event suppression is `Player`'s.

`PlayerWriteBehindTests`' `[Theory]` closes the loop at the only place it can be observed end-to-end — accepted (`true`) yet the Redis queue stays empty, with a real allocation immediately after on the same player still publishing both events, so the assertion pins suppression rather than a broken publisher. The `[InlineData(0)]` case covering the empty payload through the same body is a nice economy.

Two things I checked rather than assumed:

- **The tests pin the fix, not just pass alongside it.** The PR reports 6 failures when `changed` is forced true; the set I'd expect to fail (the three `Unchanged` assertions, the two `RaisesNoEvent` assertions, and the theory) matches that count, and every one of them asserts on the new distinction rather than on unchanged surrounding behaviour.
- **`MakeAllocatingPlayer` is load-bearing, not decoration.** `PlayerBuilder` defaults `StatAllocations` to `[]`, so an all-zero payload against a default-built player would *create* the rows and correctly report `Changed` — the tests would pass for the wrong reason. Seeding `CreateAllocations()` is what makes them mean what they say, and the helper's doc comment says so.

Retargeting `ZeroAmount_Succeeds` rather than adding a parallel test was the right move — leaving it asserting the old behaviour would have been a green test pinning the bug.

No frontend or battle-logic surface, so no mirroring obligation.

### What's good

- **The scope discipline.** #2485 suggested `changedPoints != 0` and immediately caveated it; the PR resolved that into a per-update test and then found the third case the issue didn't consider (a zero delta on a missing row), reasoned it to `Changed` from persistence consistency rather than from what's convenient, and pinned both halves. That's the difference between implementing an issue and checking it.
- **Correcting the issue's framing on `RaiseCoreUpdated()`** — the issue calls it "the core-updated push", and the answer is that it isn't a push at all. Tracing the consumers instead of accepting the description is what made the suppression obviously safe rather than a judgement call.
- **The renames are complete in code**, including the `<see cref>`s and the frontend comment that would otherwise have silently rotted. `Try*` returning an enum genuinely would have been misleading, and `SetFavorite` already set the precedent for dropping it.
- The PR body being upfront that the real client can't trigger this — so the reader can calibrate urgency — rather than overselling the severity of the issue it closes.

---

Game.Core.Tests/Players/PlayerTests.cs
ginderjeremiah: Both nits taken in `e73c929`.

Renamed to `Update`, matching the sibling `PlayerStatPointsTests` — the shadowing of `Game.Api.Models.Attributes.AttributeUpdate` was the deciding factor. Left the two `Game.Application.Tests` copies alone; consolidating them into `Game.TestInfrastructure` is real but belongs with the other test-double consolidation work rather than here.

On the spike doc — you're right that it wasn't a decision, it was a miss. All three references in `docs/spikes/1126-class-system.md` (lines 30, 51, 84) are updated. Your reasoning is what settled it: area D is unimplemented plan written in the present tense about a path that still has to be built, so it's a live grep target, not a frozen record of what was true when the spike was written. Had it been describing a decision as-of a date I'd have left it.

No action on #2507 here, as intended.

---

PR #2508, opened 2026-07-28
Extract the shared contract-drift rule out of the fixture modules (#2482)

### LINKED ISSUE #2482: Extract the shared contract-drift rule duplicated across the test fixture modules

ginderjeremiah:
Closes #2482.

Comments and docs only — no test or production code changes.

## Where the rule went

`docs/frontend.md` → Testing Guidelines, directly under the existing fixture-builder convention line, rather than a new `UI/src/tests/fixtures/README.md`. Both were offered as equally fine; I picked the doc because **there is no README anywhere under `UI/src` or `UI/e2e-tests`** — the repo's established habit is to split detail into `docs/*.md` (`frontend-battle.md`, `frontend-screens.md`, `frontend-sockets.md`, …). Introducing the tree's first source-adjacent README to hold one paragraph would have been the novel choice, not the conservative one.

Each trimmed module now ends its header with a pointer to that section, so the rule is one hop away from where you'd need it.

## The duplication was wider than the issue counted

The issue names `items.ts`, `skills.ts`, `attributes.ts`, `challenges.ts`, and `classes.ts` as the five carriers. Checking each header against that list:

- The real carriers are **seven**: `items`, `skills`, `attributes`, `challenges`, `enemies`, `item-mods`, `zones`.
- **`classes.ts` does not carry the rule at all.** Its header is entirely per-contract — why `ICreatableClass` and `IClass` are independently-drifting contracts, and why `IClass`'s single hand-rolled builder stays a literal until a second consumer appears. It's untouched, as are `proficiencies.ts` (adjacent-builder cross-references) and `workbench-config.ts`, whose note reads like the general rule but isn't: it's specific reasoning about why `EntityConfig`'s copies were `as unknown as EntityConfig<T>`, what made that double cast load-bearing, and why `newItem` is required rather than defaulted.

So the trim touches seven modules, not five, and leaves three alone.

## Two headers went away entirely

`attributes.ts` and `enemies.ts` had **nothing** but the general rule — no exclusions, no adjacent builders. `attributes.ts` loses its block comment outright (the `makeAttribute` JSDoc is untouched). `enemies.ts` keeps one line, because "there are no standing exclusions" is itself per-contract information: every other module's header answers "what's deliberately still a literal?", and silence would read as an unanswered question rather than an empty set.

## What I dropped, deliberately

Each carrier listed the specific sites #2447 converted (e.g. `attributes.ts` naming `battle/battle-attributes`, `skills/SkillDamageBreakdown`, `select/class-summary`). Those conversions have happened — the lists describe history rather than current state, no site named in them is still a cast, and git holds the record. The consolidated rule keeps the one forward-looking part ("reintroducing such a cast re-opens the hole, so state why if you do") plus a single `#2447` reference. Flagging it explicitly since it's the only place I removed information rather than relocating it.

The two load-bearing per-contract notes the issue warned against losing are both intact and were not in scope of the trim — `skills.ts`'s blank-`word` default (what marks a skill un-deciphered, which the synthesis reveal suites depend on) and `challenges.ts`'s `challengeTypeId`/`entityType` default pair both live in the `make*` JSDoc, not the header block.

## Test plan

- `npm run lint` — clean (`eslint` + `svelte-check`: **1231 files, 0 errors, 0 warnings**).
- `npm run test:unit:ci` — **313 files / 4059 tests passed**, 0 failed.

No backend files touched, so the .NET suites aren't affected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MoZ55rdGtHdecTGT5tyuAK)_


ginderjeremiah:
## Review: extract the duplicated contract-drift rule out of the fixture headers into `docs/frontend.md` (#2482)

**Verdict:** Approve (posted as a comment — GitHub won't take a gating verdict from the PR's own author) — the consolidation is correct, the claims in the description hold up under checking, and the one asymmetry it left behind is fixed in this PR.

### Verifying the description's claims

I checked each of these against the tree rather than taking the body at its word.

**Seven carriers, not five — confirmed.** `items`, `skills`, `attributes`, `challenges`, `enemies`, `item-mods`, `zones` all carried a near-verbatim copy of the rule on `main`, and all seven are trimmed here. The issue's list was assembled by eye and missed `enemies`/`item-mods`/`zones`.

**`classes.ts` is not a carrier — confirmed.** Its header is entirely per-contract: `ICreatableClass` vs `IClass` as independently-drifting contracts, and why `IClass`'s single hand-rolled builder stays a literal until a second consumer appears. Nothing general in it. The issue was wrong to name it, and leaving it untouched is right.

**`proficiencies.ts` — confirmed.** Adjacent-builder cross-references (`lexicon-test-utils`, `progression-test-utils`) plus the divergent-defaults note. No general rule.

**`workbench-config.ts` — agreed, though it's the closest call.** It does restate the drift concept ("none of them saw a new required field"), but only as the setup for genuinely specific reasoning: why the `as unknown as EntityConfig<T>` double cast was load-bearing (`persist` returning `[]` instead of `{ records, idMap }`), and why `newItem` is required rather than defaulted. That's not a copy of the shared rule, it's an application of it to one contract's history. Keeping it is the right call.

**The two load-bearing notes the issue warned about — both intact.** `skills.ts`'s blank-`word` default and `challenges.ts`'s `challengeTypeId`/`entityType` pair (plus the "off the production manifold" warning) are in the `make*` JSDoc, untouched by the trim.

### On dropping the per-module lists of #2447 sites

This is the only place information was removed rather than relocated, and the call is right. I spot-checked the named sites: `battle-attributes`, `SkillDamageBreakdown`, `class-summary`, `challenge-unlocks`, `ChallengeTooltip`, `enemy-attributes`, `enemy-level`, `ActiveEffectChips`, `inventory-view` all build through the fixture modules now. The only surviving `as <Contract>` hits under `UI/src` are array annotations (`[] as IAttribute[]`) and the workbench draft-record `as unknown as IChallenge` pattern — a different, structural thing, not partial-literal casts. So the lists genuinely described history, and git holds it. The forward-looking half ("reintroducing such a cast re-opens the hole, so state why if you do") survives in the doc, which is the part that has to survive.

`docs/frontend.md` over a `UI/src/tests/fixtures/README.md` is also the right pick, and for the stated reason: there is no README anywhere under `UI/src`, so the doc is the conservative choice, not the novel one.

### Fixed in this PR

- **`UI/src/tests/fixtures/attributes.ts` was left as the only fixture module with no header at all.** Two things were slightly off about that. First, the new doc paragraph asserts "each fixture module's header names the exclusions specific to its contract" — false for `attributes.ts` once its block went away, and the description's "each trimmed module now ends its header with a pointer to that section" was likewise true for six of seven. Second, `attributes.ts` does have a standing exclusion: `routes/game/screens/codex/Codex.test.ts` hoists `staticData: {} as any` and seeds attribute-shaped literals into it, so it's exactly the untyped-mock class the doc describes. The reasoning that led `enemies.ts` to keep one line applies here too — silence reads as an unanswered question. Pushed a one-liner in 1a1a490 mirroring the `enemies.ts` shape.

### Test coverage

Comments and docs only — no test or production code changed, so no new coverage is owed. CI is green on the pre-fix head (`verify-frontend`, `test-frontend`, all three `test-e2e` browsers, `all-checks-passed`); the backend jobs are path-filtered to skipped, correctly. The follow-up commit is a comment in a test fixture and can't move any of that.

### What's good

- The trimmed headers read better than the originals — each one now answers exactly one question ("what's deliberately still a literal for *this* contract?") instead of burying it under four paragraphs of shared prose.
- The consolidated doc paragraph kept the non-obvious operational bit that a `grep` is the wrong instrument and a throwaway required field is the right one. That's the part a future author would otherwise have to rediscover.
- Checking the issue's own list of carriers instead of implementing it as written caught two errors in it. That's the behavior the guidelines ask for.

---

Issue #2509, opened 2026-07-28
Refresh the sequencing docs once #2501 and #2502 land — "one stream per handler" and the shipped-status table go stale


ginderjeremiah:
Raised in the re-review of #2476 (the [spike #2467](https://github.com/ginderjeremiah/GameServer/issues/2467) artifacts). Not actionable until **both** #2501 and #2502 are on `main` — the doc statements below are accurate today and would only become wrong on their merge, so #2476 should not pre-empt them.

### 1. `Stream` is no longer "one value per guarded handler"

Both artifacts state the original rule:

- `docs/spikes/2467-cross-instance-write-behind-sequencing.md` §2 — *"`Stream` is a small enum (one per guarded handler)"*
- `docs/backend-persistence.md` — *"where `Stream` is one value per guarded handler"*

#2501 replaces it with **one value per target space**, and correctly: a stale `ModApplied` is only rejected because `ModRemoved` advanced *the very watermark it compares against*, so the tombstone case the table exists for requires the two handlers to share a stream. Same for unequip against equip.

#2502 then adds a deliberate counter-example — `ItemFavorite` does *not* share `Equipment`'s item key despite both writing the same `UnlockedItems` row, because they touch disjoint columns and ordering a favourite toggle against an equip is meaningless. #2505 already covers adding that exception to the **enum's own summary**; this issue is the two `docs/` copies of the same rule, which #2505 does not touch.

### 2. The spike's implementation-status table

`#2495` and `#2496` both read **Open**; they become shipped. `#2475` ("blocked on #2495 + #2496") stays correct — #2502's PR body makes the same point.

### 3. `docs/backend-persistence.md`'s "closed for which handlers" sentence

It currently reads *"closed for `PlayerCoreUpdated` and `ProgressUpdated`, and still open for every other absolute-value handler until #2495 … and #2496 … land — which is also why the parked-lane deferral below is still load-bearing."* Once both land the guard covers every absolute-value handler, and the parked-lane paragraph's *"It stays until the guard reaches every handler (#2495, #2496)"* becomes the trigger for #2475 rather than a standing statement.

## Scope

Text only, in `docs/spikes/2467-cross-instance-write-behind-sequencing.md` and `docs/backend-persistence.md`. Cheapest done as a rider on whichever of #2501/#2502 merges second, or on #2475 — but it needs an owner, since it is invisible from either PR's own diff, which is the same failure mode that produced #2500.

ginderjeremiah:
[...]

Issue #2505, opened 2026-07-28
Consolidate the two target-key formatters once #2502 and #2501 have both merged


ginderjeremiah:
Follow-up from the review of #2502. This is a **cross-PR** cleanup that neither PR owns, which is exactly why it needs an issue — it is invisible on each branch in isolation and will not show up as a merge conflict.

## The collision

#2502 and #2501 independently invented a home for `PlayerWriteWatermark.TargetKey` formatting:

- **#2502** adds `PlayerWriteWatermarkGuard.Target(...)` — two overloads, formatting invariantly via `CultureInfo.InvariantCulture`. It exists specifically because target keys were open-coded at four sites in two idioms, one of which had a culture-sensitivity bug.
- **#2501** adds a separate `Game.DataAccess/PlayerUpdates/PlayerWriteTargets.cs` — nested per-stream classes (`PlayerWriteTargets.Equipment.Item`, `.Slot`, `PlayerWriteTargets.Mods.Slot`) whose builders are **culture-sensitive interpolation** (`$"item:{itemId}"`).

Merging both leaves two competing helpers, and re-introduces the exact pattern #2502 centralized away — with no build failure to catch it (`#2491`'s Globalization analyzers would, if they were enabled).

## Suggested resolution

Keep **`PlayerWriteTargets`** as the public home and delete the public `Target` overloads from the guard:

- The guard is the *mechanism*; the target key format is the *stream's persisted contract*. A class named for the contract, with one nested type per stream, is the better place for it — and it reads better at the call site (`PlayerWriteTargets.LogPreference(evt.LogType)` over `PlayerWriteWatermarkGuard.Target((int)evt.LogType)`).
- `PlayerWriteTargets` then delegates to the invariant primitive so the CA1305 property is held in one place.
- Add builders for #2502's streams (`SelectedSkills`, `LogPreference`, `AttributeAllocations`, `ItemFavorite`, `LessonRead`) and the progress stream's prefixed keys, so *every* stream's key comes from the same file.

### One wrinkle to handle deliberately

The two key spaces are compatible except at one point. `Target("item", id)` / `Target("slot", id)` reproduce #2501's equipment keys exactly, but its mods key is prefix-less (`$"{itemId}:{modSlotId}"`) and `Target("", a, b)` yields `":a:b"`, not `"a:b"`.

[...]

PR #2510, opened 2026-07-28
Retire the cross-pass parked-lane deferral now the write sequence guard covers every handler (#2475)

### LINKED ISSUE #2475: Retire the cross-pass parked-lane deferral once the write sequence guard lands

ginderjeremiah:
Closes #2475.

Implementation step 3 of spike #2467. The prerequisite is met: #2495 (PR #2501) and #2496 (PR #2502) are both on `main`, so the watermark guard now covers all nine streams / eleven absolute-value handlers — I verified this against `PlayerWriteStream` rather than taking the issue's word for it.

## What changed

`_parkedPlayerLanes` existed for exactly one reason: to stop a newer same-player event applying ahead of an older one this instance left parked on the processing list, because the reclaim's later replay would regress the newer absolute values (#2460). The guard now rejects that replay on its own merits, so the deferral was two mechanisms guarding one invariant with only one of them load-bearing — and the one being removed only ever covered the within-instance half anyway.

Removed:
- the `_parkedPlayerLanes` field (and the now-unused `System.Collections.Concurrent` import)
- the deferral branch in `DrainQueueAsync` and its `deferredPlayers` log-throttling set
- the two `.Clear()` calls in `TryReclaimStrandedProcessingAsync`

Two knock-on cleanups the issue didn't call out but which fall out of it: the `catch` in `ProcessReservedItemAsync` had nothing left but a rethrow, so it collapses into the existing `try`/`finally` (behaviourally identical), which in turn makes its `playerId` parameter dead — both removed rather than left behind.

**Kept, per the issue:** the pass-scoped `playerLanes` map and `CompactCompletedItems`' don't-evict-a-faulted-lane-head rule. Their rationale changes rather than disappearing — they now avoid doing (and acknowledging) work the guard would only throw away within a pass, and keep a faulted player's chain reserved together so the reclaim replays it in one piece. I rewrote the comments to say that instead of the old "covers the window the parked map can't".

`docs/backend-persistence.md`: the "That block outlives the pass that created it" paragraph is replaced by one explaining that the block is now deliberately pass-scoped, and the deferral carve-out in the concurrency-slot paragraph is gone (every reserved item now holds its slot for real work).

## Tests

The within-instance interleaving test is **kept and re-pointed at the guard**, as the issue's verification section asks — the parked event is still reclaimed and replayed, and the DB must still hold the newer state, now because the replay is rejected rather than because it was deferred. Renamed to `ProcessQueue_LaterPassAfterOutage_AppliesNewerSamePlayerEventAndRejectsTheParkedReplay`.

One thing worth flagging: it now has to stamp **explicit sequences (1/2/3)**. The shared `DefaultSequence` stamps every event `1`, and the guard accepts equal sequences, so left on the default the replay would be accepted and regress the row — the test would have silently pinned the opposite of what it claims. I confirmed it isn't vacuous by temporarily stamping the parked event `4`: it fails with the level-9 regression, i.e. the assertion genuinely depends on the guard.

Removed as now-dead (each only ever pinned the deferral or its release, and would pass vacuously):
- `ProcessQueue_LaterPassAfterOutage_StillAppliesOtherPlayersEvents`
- `ProcessQueue_ParkedItemAcknowledgedElsewhere_UnblocksThePlayerOnTheNextReclaim`
- `ProcessQueue_ProcessingListObservedEmpty_ReleasesTheParkedLaneWithoutWaitingForAReclaim`

The last of those incidentally touched the `_strandedProcessingHead` reset-on-empty path; that clock is already covered directly by `...StrandedWithinGracePeriod...`, `...DifferentItemReplacesProcessingHead...`, `...DuplicatePayloadReplacesProcessingHead...` and `...StrandedPastGracePeriod...`, so no coverage is lost.

Also corrected two stale `#2460` rationale comments on the outage-logging tests, which explained their use of separate players in terms of the deferral.

## One residual worth naming

An **unsequenced** envelope (`sequence: 0` — what a pre-upgrade instance's enqueue deserializes to during a rolling deploy) bypasses the guard by design, so a parked seq-0 event's replay still overwrites a newer one. The deferral used to cover that case within an instance.

I don't think this blocks the change: during that same window the fleet also contains pre-upgrade instances that apply everything unconditionally and share the queue, so ordering for those envelopes is already unprotected fleet-wide — this removes one instance-local mitigation of a subset, not a standing guarantee. It's bounded to the deploy window and self-resolves once every instance stamps real sequences. Flagging it rather than filing a follow-up, but happy to file one if you'd rather it were tracked.

## Test plan

- [x] `dotnet test` — full backend suite green: 1501 Core + 1188 Application + 893 Api + 74 Infrastructure = **3656 passed, 0 failed**
- [x] Rewritten interleaving test verified non-vacuous (fails when the replay is made non-stale)
- [x] No frontend changes, so `UI` lint/unit not run

## Note on the sibling issues

#2509 (refresh the sequencing docs) is only half-actionable right now — its `docs/backend-persistence.md` claims were already corrected by #2501/#2502, and the spike doc it also targets isn't on `main` yet (it's in the still-open #2476). I left it alone rather than half-doing it here. #2505 (consolidate the two target-key formatters) is unblocked by the same merges but is a separate concern.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CPvbR7AVko3DLuNkmQ4dqW)_


Game.DataAccess/DataProviderSynchronizer.cs
```diff
@@ -534,10 +488,10 @@ private async Task DrainQueueAsync(IPubSubQueue queue, CancellationToken cancell
         /// item count. Only <em>successfully</em> completed tasks are evicted: a faulted (or canceled) lane
         /// head must stay in the map so a later same-player item chains onto it and faults in turn — staying
         /// reserved for the reclaim — rather than starting a fresh lane and applying <em>ahead</em> of the
-        /// failed item's eventual reclaim/re-apply, which would break same-player ordering exactly on the
-        /// fault path. This covers the narrow window <see cref="_parkedPlayerLanes"/> (the primary guard, which
-        /// blocks the lane across passes) cannot: a same-player item reserved <em>before</em> the faulting task
-        /// records the block. Keeping them in <paramref name="inFlight"/> likewise keeps the drain-exit settle aware
+        /// failed item's eventual reclaim/re-apply. The watermark guard would reject that replay anyway, so
+        /// this is about not doing (and acknowledging) work the guard is only going to throw away, and about
```
ginderjeremiah: **Should-fix (comment accuracy):** "The watermark guard would reject that replay anyway" isn't unconditionally true, and this rewrite demotes the retained lane rule to a pure efficiency measure when part of it is still load-bearing for correctness.

`AdvanceWatermarksAsync` accepts on `<=` — deliberately, since "equal sequences are the same save's sibling writes landing on one target" — and `PlayerUpdateBatch.BeginPlayerSave` stamps every envelope of one `SavePlayer` with that save's single sequence. So for two events from the *same save* on the same stream + target key, the guard provides no ordering at all: whichever applies last wins, in either direction. The pass-scoped `playerLanes` chain (plus this method's don't-evict-a-faulted-lane-head rule) is the only thing ordering them.

That's not a defect in the change — the lane map is kept, so the behaviour is right. The risk is the comment: as written, a later maintainer reading "this is about not doing work the guard is only going to throw away" has a documented licence to delete the lane map as dead weight, which would quietly reopen same-save ordering. Worth one clause here (and in the matching note on `ProcessReservedItemAsync`, ~L570-573, which says the same thing) along the lines of: *the guard accepts equal sequences, so same-save sibling writes to one target are ordered by the lane alone; across saves the guard is what rejects the stale replay.*


---

ginderjeremiah (commented):
## Review: retiring the cross-pass parked-lane deferral now the watermark guard covers every absolute-value handler (#2475)

**Verdict:** Approve (posted as a comment — GitHub won't accept a gating verdict from the PR author) — the removal is correct and the prerequisite genuinely holds; one comment-accuracy should-fix and one coverage nit, neither blocking.

I verified the premise rather than taking it from the issue. `PlayerWriteStream` carries nine members, and exactly four handlers — `ItemUnlockedHandler`, `LessonUnlockedHandler`, `ModUnlockedHandler`, `SkillUnlockedHandler` — don't call `ExecuteGuardedAsync`, which matches the spike's "pure insert-if-missing, genuinely convergent" set. Every absolute-value handler is guarded, so the deferral was in fact redundant.

I also chased two ways the removal could have reopened #2460 and neither holds:

- **Cross-*stream* reordering** (an `ItemUnlockedEvent` parked while a guarded `ItemEquippedEvent` from a later pass applies) — `ItemEquippedHandler.ApplyAsync` already inserts the row when the unlock reordered behind it, with the guard's unique-violation restart behind it. Designed for, not a hole this opens.
- **A non-deploy source of `Unsequenced`** — `PlayerUpdateBatch._playerWriteSequence` defaults to `0` and `BeginBatch()` passes `Unsequenced` explicitly, which looked like it might strand real envelopes at sequence 0. It doesn't: `SavePlayer` opens its own inner `BeginPlayerSave(player.AdvanceWriteSequence())` inside any outer `BeginBatch` scope, and `PlayerProgressRepository` stamps `progress.AdvanceWriteSequence()` onto its own envelope independently. So the seq-0 residual really is rolling-deploy-only.

### Should-fix

- `Game.DataAccess/DataProviderSynchronizer.cs:490-492` (and the matching note at `:570-573`) — the rewritten comments say the guard "would reject that replay anyway", which is false for two events of the *same save*: the guard advances on `<=` and one save stamps all its envelopes with one sequence, so the pass-scoped lane is the only ordering mechanism there. The code is right; the comment invites a future maintainer to delete the lane map as an optimization. Inline detail above.

### Nits

- The PR body says the `_strandedProcessingHead` clock is "already covered directly by" `...StrandedWithinGracePeriod...`, `...DifferentItemReplacesProcessingHead...`, `...DuplicatePayloadReplacesProcessingHead...` and `...StrandedPastGracePeriod...`. Those four cover the *different-item* restart and the reclaim-path reset; none covers the `head.Count == 0` reset in `TryReclaimStrandedProcessingAsync:607-611`, which `ProcessQueue_ProcessingListObservedEmpty_...` was the only test to exercise. That branch is behaviourally unchanged by this PR (only the `.Clear()` inside it went), and it only matters when the *same* raw item returns to the head after the list was seen empty — so this is lost incidental coverage, not a regression. Fine to leave; just don't count it as covered.

### On the seq-0 residual

Agreed it doesn't block, and I wouldn't file a follow-up either. Beyond the fleet-wide argument in the PR body, `CLAUDE.md` is explicit that this project carries no backwards-compatibility obligation while in development, so a mixed-version rolling deploy isn't a scenario worth tracking an issue against. It's correctly documented on `DomainEventEnvelope.Unsequenced` and `PlayerWriteWatermarkGuard`, which is the right place for it.

### Test coverage

Adequate, and the interleaving test is the right call.

- The kept-and-re-pointed `ProcessQueue_LaterPassAfterOutage_AppliesNewerSamePlayerEventAndRejectsTheParkedReplay` now asserts the opposite outcome for the right reason — newer state stands, parked replay rejected, and it pins the `"stale write target(s)"` warning so the rejection is observable rather than merely inferred from the row. Explicit sequences 1/2/3 are necessary here (`DefaultSequence` is `1`, and the guard accepts equal sequences), and the non-vacuity check in the PR body is exactly the right thing to have done.
- The three removed tests each only ever pinned the deferral or its release, and would pass vacuously now. Correct to delete rather than leave as false assurance.
- `ProcessQueue_ItemFaultsPastCompactionThreshold_LaterSamePlayerItemStaysBlockedForReclaim` still pins the kept within-pass rule, so the two things the issue asked to keep are both still under test.
- No frontend or battle-logic surface, so no mirrored-test obligation.

### What's good

- Not taking the issue's word for the prerequisite and checking it against `PlayerWriteStream` — the "verify what the issue claims" habit `CLAUDE.md` asks for, applied where it actually mattered.
- Collapsing `ProcessReservedItemAsync`'s `catch { …; throw; }` into the existing `try`/`finally` and dropping the now-dead `playerId` parameter. The issue said to keep the `catch`; a bare rethrow is behaviourally identical to no catch at all, so removing it is right and leaves no dead code — a good example of not following an issue past the point where it's correct.
- The docs change replaces the old behaviour rather than narrating the removal, and drops the concurrency-slot carve-out that only existed to describe deferred items. That's what `CLAUDE.md` asks docs edits to look like.
- Correcting the two stale `#2460` rationale comments on the outage-logging tests — easy to miss, and they'd have been actively misleading once the mechanism they cite was gone.


---

Game.DataAccess/DataProviderSynchronizer.cs
ginderjeremiah: Fixed in `8b52125`. You're right, and I checked the two premises rather than just taking the note: `PlayerWriteWatermarkGuard` advances on `<=` (with the "equal sequences are the same save's sibling writes" rationale stated inline), and `PlayerRepository.SavePlayer` opens `BeginPlayerSave(player.AdvanceWriteSequence())` once per save, so every envelope that save raises carries one sequence. Same-save siblings on one stream+target are therefore ordered by the lane alone.

Both comments now say so explicitly — the `CompactCompletedItems` doc calls the rule "load-bearing for correctness, not just efficiency" and names the `<=`/one-sequence-per-save mechanism, and the `ProcessReservedItemAsync` note now scopes its guard claim to a "different save" and adds the same-save carve-out.

One thing I looked into while fixing it, worth recording so the carve-out isn't read as wider than it is: I couldn't construct a case where same-save siblings actually get split across passes in a reordering way. They're enqueued contiguously by one multi-value push; if the first faults, the second is reserved in the same pass, chains onto the faulted lane, faults in turn and stays reserved alongside it; and if a pass ends between them (a stop), the startup reclaim re-queues the parked one to the head ahead of the still-queued sibling. So the lane is doing real ordering work within a pass, but I don't believe removing the cross-pass deferral opened a same-save hole — the comment was the whole defect.

On the coverage nit: agreed, and thanks for catching the overclaim. The `head.Count == 0` reset in `TryReclaimStrandedProcessingAsync` genuinely isn't covered by those four tests — that branch is unchanged here (only the `.Clear()` inside it went), so it's lost incidental coverage rather than a regression, and I won't count it as covered.

---

Issue #2488, opened 2026-07-26
Consolidate the remaining duplicated test doubles (FakeTimeProvider, FakeUserLogins)


ginderjeremiah:
Noticed while reviewing #2487, which consolidated the `ISessionStore`/`IUnitOfWork` copies and wrote the convention into `docs/backend.md` → Testing Guidelines: **one canonical double per interface**, in the shared test-infrastructure project, with a carve-out for doubles that record or script something specific to one test class.

Two more cross-project duplicates predate that rule and were out of #2487's scope (different files, two test projects — unlike the `IUnitOfWork` copies, which sat in the files already being edited).

## `FakeTimeProvider` — byte-identical

- `Game.Api.Tests/Unit/SocketCommandProcessorTests.cs:764`
- `Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs:2826`

Same `(DateTimeOffset start) : TimeProvider` with `GetUtcNow()` + `Advance(TimeSpan)`; only the doc comment differs (switch-credit deadline vs. opportunistic-reclaim grace period). No carve-out applies — it scripts nothing test-specific.

Worth deciding between two options rather than defaulting:

1. Move one copy to `Game.TestInfrastructure/Helpers` and delete the other.
2. Adopt `Microsoft.Extensions.TimeProvider.Testing`, which ships a `FakeTimeProvider` with the same `Advance` surface plus timer support. Costs a package reference on the test projects; removes a hand-rolled class the framework already owns.

Option 2 is probably right if the timer support is ever wanted, but it's a new dependency — check `Directory.Packages.props` and whether anything else would use it before pulling it in.

## `FakeUserLogins` — same interface, genuinely different behaviour

- `Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs:220` — counts `RecordConnection` calls, supports throw-injection and an `OnRecordConnection` callback; `SaveDeviceInfo` throws `NotSupportedException`.
- `Game.Application.Tests/Services/LoginTrackingServiceTests.cs:38` — captures the `CancellationToken` from both members.

This one is a real judgement call under the new carve-out. The two doubles record different things, so "records something specific to that one test class" arguably covers them — but they're the same interface implemented twice, and `IUserLogins` has two multi-parameter members, so it's exactly the shape that costs two edits every time the interface changes. A merged double (call count + throw-injection + callback + last tokens) would cover both call sites, at the cost of a slightly grab-bag helper.

[...]

Issue #2452, opened 2026-07-26
Audit the nondeterministic Game.Core.Tests cases now that a red run names the failing test


ginderjeremiah:
Follow-on from #2440 / #2441, split out of PR #2451 which closed the observability half.

## Context

#2440 and #2441 both framed the reporting gap as the **prerequisite** for triaging the flake, not the triage itself — *"Fixing the reporting is the prerequisite for triaging the flake, so this should land first"*, and *"the audit above is the natural follow-on work once a failing run says what it failed."* #2451 landed that half: `build/coverage.ps1` now prints the failing test's name, assertion and stack, and `test-backend` uploads the raw per-project logs (with their per-test timings) on failure.

One item from #2441's list was genuinely free and went into that PR: `ProbabilityTableTests.SetupMultiRandomElementList` now seeds its weight fixture (`new Random(42)`), because an unseeded generator meant a failure there couldn't be reproduced *even with the test name in hand* — which would have defeated the point of the reporting fix.

## What's left

The remaining candidates are all real judgment calls rather than one-liners, and hardening a timing assertion without knowing which one actually flaked is guesswork:

- **`Battle/Performance/BattlePerformanceTests.cs`** — wall-clock budgets on a shared runner: `MaxLengthRealisticCeilingMs` (250 ms) and the `2.0×` `StackingFlatnessBudget` ratio. Both already take a min/ratio over several samples rather than a single reading, so they aren't naive — but they are still timing assertions on a hosted runner, and a loaded runner is the single most likely explanation for the #2438 flake. The question isn't "are they flaky in principle" but "what should a perf test assert on shared hardware at all" — a wall-clock ceiling, an operation-count proxy, or a `Skip` outside a dedicated perf run.
- **`Probability/ProbabilityTableTests.ConcurrentGetRandomValue_StaysValidAndNonDegenerate`** — explicitly documented in-place as a best-effort concurrency guard that "would not reliably fail against the old per-instance RNG". The draws come from `Random.Shared`, so the test stays nondeterministic even after #2451's fixture seeding. Its own comment argues it earns its keep as a contract lock-in; worth confirming that's still the call rather than assuming it.

## Why this waits for evidence

The #2438 failure was near-certainly one of these (the diff was arithmetically identity-preserving and it didn't reproduce locally), but *near-certainly* is exactly the standard #2451 exists to replace. The next red `test-backend` will name the test outright and attach the timing log. Do the audit against that, not against a hypothesis.

## Scope

[...]

PR #2511, opened 2026-07-28
Widen the reconciliation interval in the periodic-wait reset test (#2486)

### LINKED ISSUE #2486: CoalescingReferenceCacheReloaderTests.RunAsync_SignalArrivesBeforeReconciliationInterval_ResetsThePeriodicWait flakes on a loaded runner
A real red `test-backend` run named this test, so filing the evidence while it exists. Observed on PR #2484 ([run 30213819749](https://github.com/ginderjeremiah/GameServer/actions/runs/30213819749/job/89824437397)), whose diff touches no reference-cache code; a re-run of the same commit went green, and the test passes 3/3 locally.

## The failure

```
failed Game.Application.Tests.DataAccess.CoalescingReferenceCacheReloaderTests.RunAsync_SignalArrivesBeforeReconciliationInterval_ResetsThePeriodicWait (792ms)
  Assert.DoesNotContain() Failure: Filter matched in collection
               ↓ (pos 0)
  Collection: [CapturingEntry { Category = Game.DataAccess.CoalescingReferenceCacheReloader, Level = Debug,
    Message = Periodic reference-cache reconciliation sweep running (00:00:00.2000000 since the last sweep)., ... }]
    at ...CoalescingReferenceCacheReloaderTests.cs:205
```

## Why it races

`Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs:191-206` configures a **200 ms** `reconciliationInterval`, fires `NotifyChanged()`, awaits the signal-triggered sweep, then sleeps **100 ms** and asserts no periodic sweep has logged:

```csharp
var policy = new ReferenceCacheReloadPolicy(TimeSpan.Zero, maxAttempts: 1, baseDelay: TimeSpan.Zero, reconciliationInterval: TimeSpan.FromMilliseconds(200));
...
harness.Reloader.NotifyChanged();
await WaitUntilAsync(() => cache.CompletedReloads == 1, "the signal-triggered sweep to complete");
await Task.Delay(TimeSpan.FromMilliseconds(100), TestContext.Current.CancellationToken);
Assert.Equal(1, cache.CompletedReloads);
Assert.DoesNotContain(harness.Logs.Entries, e => ... "reconciliation" ...);
```

The assertion only holds while `WaitUntilAsync` + the 100 ms delay together stay under the 200 ms interval — a budget of **under 100 ms of scheduling jitter** on a shared hosted runner also running three Testcontainers-backed suites. The logged elapsed is exactly `00:00:00.2000000`, i.e. the timer fired at its configured interval and the observation window simply overran. The behaviour under test (a reactive sweep restarts the periodic wait from zero) is correct in both outcomes; only the window is wrong.

Note the sibling at `:178-189` is not exposed to this — it asserts a sweep *never* happens with reconciliation disabled, so runner slowness can't manufacture one.

## Suggested fix

Widen the margin rather than lengthening the test: raise the `reconciliationInterval` well above the observation window (e.g. 2 s) and keep the 100 ms delay. The assertion is unchanged and the runtime is unchanged, but the jitter budget goes from ~100 ms to ~1.9 s. A time abstraction (`TimeProvider`/`FakeTimeProvider`) over the reloader's periodic wait would remove the wall-clock dependency outright, but that is a larger change than the flake warrants on its own.

## Relationship to #2452

#2452 is the audit of nondeterministic cases *in `Game.Core.Tests`* (`BattlePerformanceTests`, `ProbabilityTableTests`), and it explicitly waits for a red run to name a test. This is that evidence arriving — but for a different suite and a different test than the two it lists, so it is filed separately rather than folded in. Both trace back to #2440/#2441/#2451: the reporting fix worked exactly as intended here, naming the test, its assertion, and the logged elapsed time straight out of the CI log.

ginderjeremiah:
Closes #2486.

Test-only, one file, no production code.

## What changed

`RunAsync_SignalArrivesBeforeReconciliationInterval_ResetsThePeriodicWait` configured a **200 ms** `reconciliationInterval` and then observed for **100 ms** after the signal-triggered sweep. That left under 100 ms of scheduling jitter before the *legitimate* periodic sweep fires inside the observation window and trips the `DoesNotContain` assertion — which is exactly what the CI log in the issue shows (`00:00:00.2000000` since the last sweep, i.e. the timer fired right on its configured interval and the window simply overran).

Raised the interval to **2 s**, per the issue's suggested fix. The assertion, the 100 ms delay, and the test's runtime are all unchanged; the jitter budget goes from ~100 ms to ~1.9 s.

## Why this doesn't weaken the test

Worth stating explicitly, since widening a timing window is the classic way to quietly neuter a test.

The regression this test guards against — a carried-over periodic wait that's already primed when the next iteration starts — manifests as a reconciliation sweep logging *immediately* after the reactive sweep, not one interval later. So it lands inside the 100 ms window at any interval value. The 200 ms only ever contributed a false-failure race; it never contributed detection power.

I verified that empirically rather than just asserting it: mutating `WaitForNextSweepAsync` to force the reconciliation branch (`var reconciling = true;`) fails **exactly one** of the 1188 tests in the project — this one. Mutation reverted; the diff is the test file only.

I also checked the two sibling timing tests in the suite and neither has this shape: `RunAsync_NoSignalWithinReconciliationInterval_RunsAPeriodicSweepAnyway` waits *until* a sweep happens (runner slowness only delays it), and `RunAsync_NoReconciliationIntervalConfigured_NeverSweepsWithoutASignal` has reconciliation disabled, so slowness can't manufacture a sweep. No other change needed here.

## Note on the alternative

The issue floats a `TimeProvider`/`FakeTimeProvider` abstraction over the reloader's periodic wait as the way to remove the wall-clock dependency outright, and agrees it's larger than this flake warrants. I left it alone — it's production-code surgery on a background loop to fix a test window, and #2488 is already the open decision on which `FakeTimeProvider` this repo standardizes on. Worth revisiting there rather than pre-empting it here.

## Test plan

- [x] `dotnet build` — clean, 0 warnings, 0 errors.
- [x] `dotnet test Game.Application.Tests` — 1188/1188 passed, run **3×** to shake out the timing dependence (the flake reproduced ~1-in-N on CI, never locally).
- [x] Mutation check above, confirming the assertion is still live.
- [x] Rebuilt and re-ran green after reverting the mutation.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X63M7F4avb1BJF6XR2G2Qz)_

### DIFF
--- a/Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs
+++ b/Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs
@@ -192,7 +192,11 @@ namespace Game.Application.Tests.DataAccess
         public async Task RunAsync_SignalArrivesBeforeReconciliationInterval_ResetsThePeriodicWait()
         {
             var cache = new RecordingCache();
-            var policy = new ReferenceCacheReloadPolicy(TimeSpan.Zero, maxAttempts: 1, baseDelay: TimeSpan.Zero, reconciliationInterval: TimeSpan.FromMilliseconds(200));
+            // The interval only has to outlast the observation window below: the regression this guards
+            // against (a carried-over periodic wait) fires a sweep immediately after the reactive one, so a
+            // generous interval costs no detection power but keeps runner jitter from firing the legitimate
+            // periodic sweep inside the window and failing the test spuriously.
+            var policy = new ReferenceCacheReloadPolicy(TimeSpan.Zero, maxAttempts: 1, baseDelay: TimeSpan.Zero, reconciliationInterval: TimeSpan.FromSeconds(2));
             await using var harness = new Harness(policy, cache);
 
             harness.Reloader.NotifyChanged();
--- a/Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs
+++ b/Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs
@@ -192,8 +192,8 @@ namespace Game.Application.Tests.DataAccess
         public async Task RunAsync_SignalArrivesBeforeReconciliationInterval_ResetsThePeriodicWait()
         {
             var cache = new RecordingCache();
-            // The interval only has to outlast the observation window below: the regression this guards
-            // against (a carried-over periodic wait) fires a sweep immediately after the reactive one, so a
+            // The interval only has to outlast the observation window below: a periodic wait that's already
+            // primed when the iteration starts fires a sweep immediately after the reactive one, so a
             // generous interval costs no detection power but keeps runner jitter from firing the legitimate
             // periodic sweep inside the window and failing the test spuriously.
             var policy = new ReferenceCacheReloadPolicy(TimeSpan.Zero, maxAttempts: 1, baseDelay: TimeSpan.Zero, reconciliationInterval: TimeSpan.FromSeconds(2));


Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs
```diff
@@ -192,7 +192,11 @@ public async Task RunAsync_NoReconciliationIntervalConfigured_NeverSweepsWithout
         public async Task RunAsync_SignalArrivesBeforeReconciliationInterval_ResetsThePeriodicWait()
         {
             var cache = new RecordingCache();
-            var policy = new ReferenceCacheReloadPolicy(TimeSpan.Zero, maxAttempts: 1, baseDelay: TimeSpan.Zero, reconciliationInterval: TimeSpan.FromMilliseconds(200));
+            // The interval only has to outlast the observation window below: the regression this guards
+            // against (a carried-over periodic wait) fires a sweep immediately after the reactive one, so a
+            // generous interval costs no detection power but keeps runner jitter from firing the legitimate
+            // periodic sweep inside the window and failing the test spuriously.
+            var policy = new ReferenceCacheReloadPolicy(TimeSpan.Zero, maxAttempts: 1, baseDelay: TimeSpan.Zero, reconciliationInterval: TimeSpan.FromSeconds(2));
```
ginderjeremiah: Nit — "a carried-over periodic wait" is one shape too broad for what this window actually catches, and since this comment exists precisely to stop a future reader from re-tightening the constant, it's worth being exact.

Two distinct bugs fit that description, and the test only sees one of them:

- **An already-elapsed wait** (a completed `Task.Delay` reused, an interval that degrades to zero) fires a sweep at `reactive + ε`. Caught here at any interval — this is the one the `var reconciling = true;` mutation stands in for.
- **A wait anchored to the wrong origin** — e.g. `Task.Delay` hoisted out of `WaitForNextSweepAsync`, or a `PeriodicTimer` built once at construction — fires at `loopStart + interval`, not `reactive + interval`. Since the reactive sweep here completes in ~ε, that lands at ≈200 ms before and ≈2 s now, and the 100 ms window misses it **either way**.

So the conclusion the PR draws is right — the widening costs no detection power — but for a sharper reason than the comment gives: detection for the second shape was already zero at 200 ms, not lost by going to 2 s. The flip side is that this test never verified the *positive* half its name promises (that the next periodic sweep is re-anchored to the reactive sweep rather than to loop start); it only pins the negative that no sweep follows immediately.

Suggest narrowing the parenthetical to what's actually asserted — something like "a periodic wait that's already primed when the iteration starts fires a sweep immediately after the reactive one, so a generous interval costs no detection power" — and leaving the re-anchoring claim out of it. Comment-only; no change to the value or the assertions.

---

ginderjeremiah (commented):
## Review: widen the reconciliation interval in `RunAsync_SignalArrivesBeforeReconciliationInterval_ResetsThePeriodicWait` (#2486)

**Verdict: Approve** — the diagnosis matches the code, the fix is the minimal correct one, and the "does this neuter the test" question is answered rather than asserted. Submitted as `COMMENT` because GitHub won't take a gating verdict from the PR author; treat it as an approve with one comment-wording nit.

Not merging: the nit is outstanding and CI is still in flight (`test-backend`, `verify-backend` and the three `test-e2e` legs were all `in_progress` at review time).

I re-derived the race from the source rather than taking the issue's account of it, since the issue and the PR were written by the same hand and a wrong diagnosis here would be invisible.

### The diagnosis holds

`WaitUntilAsync` (`:237`) polls on a **10 ms** `Task.Delay` against a 10 s deadline, so the observation window is `poll latency + 100 ms`, not `100 ms`. Against a 200 ms interval that leaves well under 100 ms of slack — and the CI log's `00:00:00.2000000` is the interval itself, i.e. the timer fired exactly on schedule and the window overran it. That is a window bug, not a behaviour bug, exactly as claimed.

### The widening is free, in both directions

- **No detection cost.** `WaitForNextSweepAsync` (`CoalescingReferenceCacheReloader.cs:90`) builds a *fresh* `Task.Delay(policy.ReconciliationInterval …)` per iteration, so the reset is structural. The only failure this test can observe is a sweep landing within ~100 ms of the reactive one, which is interval-independent. (One shape the comment implies it catches, it doesn't — see the inline note.)
- **No runtime cost.** Worth spelling out because a 2 s constant in a test *looks* like +2 s: the periodic wait is linked to the loop's `CancellationToken` via `timerCts` (`:92`), and `Harness.DisposeAsync` (`:264`) cancels before awaiting `_runTask`. Teardown cancels the delay rather than waiting it out, so the test still runs in ~100 ms.
- The mutation check is the right evidence to bring, and it's consistent with the rest of the suite: `…RunsAPeriodicSweepAnyway` asserts `Contains` and `…NeverSweepsWithoutASignal` disables reconciliation, so neither would flip under `reconciling = true` — "exactly one test fails" is what you'd predict.

### On the siblings

The PR checks the two reconciliation tests; I also looked at the fourth fixed-window negative in the file, `RunAsync_NotificationBurst_CoalescesIntoOneSweepOverEveryCache` (`:36`), since it has the same delay-then-assert-no-second-sweep shape. It's safe — its policy configures no `reconciliationInterval`, so periodic sweeping is off, and the capacity-1 `DropWrite` channel drains before the reload. No sibling needs the same treatment.

### Nits

- `CoalescingReferenceCacheReloaderTests.cs:195-199` — the comment's characterisation of the guarded regression covers one bug shape the window doesn't actually catch. Inline; comment wording only.

### On deferring the `TimeProvider` alternative

Agree, and I'd go further than "larger than the flake warrants": injecting a time abstraction into a background loop's wait is production-code surgery whose only beneficiary is a negative assertion in one test, and #2488 hasn't yet decided whether this repo hand-rolls `FakeTimeProvider` or takes `Microsoft.Extensions.TimeProvider.Testing`. Pre-empting that from a test fix would be the tail wagging the dog.

I deliberately did **not** file a follow-up for it. #2452 is scoped to `Game.Core.Tests` and #2488 is about consolidating doubles, so neither is its home — but an issue whose entire content is "consider a refactor nobody currently needs" is backlog noise. #2486 preserves the reasoning permanently once closed, and the escalation trigger is concrete: if this test flakes again at a 20× margin, the wall-clock dependency is the real cause and *then* it earns an issue.

### Test coverage

Test-only, and the change is to a test's configuration rather than its assertions — the `DoesNotContain`, the `Equal(1, …)` and the 100 ms delay are all untouched, so there's nothing new to cover. No production code moves, so the project's unit/integration/e2e rules aren't otherwise engaged.

### What's good

Widening a timing window is the standard way to quietly kill a test, and this PR treats that as the thing it has to disprove rather than a formality. Two choices in particular:

- **Empirically verifying the assertion is still live** (mutate the branch, confirm exactly one test fails, revert) instead of arguing it from the test's shape. That's the difference between "I believe this still detects the bug" and "I checked."
- **Leaving the reasoning in the file, not just the PR body.** The magic number now carries its own justification at the point where someone would otherwise "tidy" it back to 200 ms. PR descriptions aren't in the reader's path six months from now; the comment is — which is exactly the carve-out in `CLAUDE.md` for going past two lines.

Diagnosing this as an *observation window* problem rather than reaching for a retry attribute or a `Skip` is also the right call — it fixes the test's contract with the runner instead of hiding the signal.

---

Game.Application.Tests/DataAccess/CoalescingReferenceCacheReloaderTests.cs
ginderjeremiah: Applied in `901cd71` — narrowed to your wording. You're right, and I checked the second shape against the source rather than just taking it: a `Task.Delay` hoisted out of `WaitForNextSweepAsync` fires at `loopStart + interval`, and since the reactive sweep completes in ~ε that's ≈200 ms before and ≈2 s now — outside the 100 ms window either way. So detection there was already zero, not traded away by this change. My parenthetical implied the window was the thing standing guard over it, which would have misled exactly the reader the comment exists for.

On the coverage gap you note — that the test only pins the negative and never the positive its name promises — I'm deliberately not filing it, for your own reason about backlog noise. Asserting re-anchoring means waiting a full interval *after* the reactive sweep to watch the next periodic one land, so the test either becomes a multi-second sleep or needs the time abstraction #2488 hasn't picked yet. An issue for it today would just be a second place waiting on that decision. If #2488 lands on a `TimeProvider` the repo actually adopts, this test is a natural early customer and the positive assertion comes nearly free.

---

ginderjeremiah:
## Re-review of `901cd71` — **Approve**, merging

Incremental. The prior round (`49498b6`) approved with one comment-wording nit outstanding and CI still in flight. Both are now resolved, so this is the merge pass rather than a fresh review.

### The nit is addressed

`901cd71` narrowed the parenthetical to what the window actually asserts. The comment now reads *"a periodic wait that's already primed when the iteration starts fires a sweep immediately after the reactive one, so a generous interval costs no detection power"* — which is exactly the shape the 100 ms window catches, with the re-anchoring claim dropped. That was the point: the wrong-origin bug (a `Task.Delay` hoisted out of `WaitForNextSweepAsync`, firing at `loopStart + interval`) lands outside the window at 200 ms *and* at 2 s, so implying the window guarded it would have misled the one reader the comment exists for.

The reply's reasoning for not filing the coverage gap holds too. Asserting the *positive* half — that the next periodic sweep is re-anchored to the reactive one — means waiting a full interval past the reactive sweep, so the test becomes either a multi-second sleep or a customer of whatever `FakeTimeProvider` #2488 settles on. An issue today would just be a second thing parked on that decision.

### Verified locally

I re-ran rather than taking CI's word, since the whole PR is about a timing window:

```
dotnet test Game.Application.Tests -- --filter-class "*CoalescingReferenceCacheReloaderTests*"
Passed! - Failed: 0, Passed: 11, Total: 11, Duration: 2s 211ms
Passed! - Failed: 0, Passed: 11, Total: 11, Duration: 2s 269ms
Passed! - Failed: 0, Passed: 11, Total: 11, Duration: 2s 393ms
```

Three consecutive green runs, and — the part worth checking — **11 tests in ~2.2 s total**, not 2 s *each*. That confirms the claim the last review made structurally: `timerCts` links the periodic wait to the loop's token and `Harness.DisposeAsync` cancels before awaiting `_runTask`, so teardown cancels the 2 s delay instead of waiting it out. A 2 s constant in a test is the kind of thing that gets "fixed" back down on sight, so it's worth having the measurement on record: the interval is free.

### CI

Green on `901cd71` — `test-backend`, `verify-backend` and all three `test-e2e` legs (chromium/firefox/webkit) passed, `all-checks-passed` success. The frontend legs correctly skipped; the diff is one backend test file.

### Verdict

**Approve**, no outstanding nits or should-fixes, CI green — merging.

No follow-up issues filed, deliberately. The one candidate (the untested positive re-anchoring assertion) is blocked on #2488's time-abstraction decision, and #2486 preserves the reasoning once closed. The escalation trigger stays concrete: if this test flakes again at a 20× jitter margin, the wall-clock dependency is the real cause and *then* it earns an issue.

---

PR #2512, opened 2026-07-29
Consolidate the remaining duplicated test doubles (#2488)

### LINKED ISSUE #2488: Consolidate the remaining duplicated test doubles (FakeTimeProvider, FakeUserLogins)

ginderjeremiah:
Closes #2488.

Test-only. Applies the `docs/backend.md` → Testing Guidelines rule from #2487 (**one canonical double per interface**, in the shared test-infrastructure project) to the two cross-project duplicates that predated it.

## `FakeTimeProvider` — option 1 (move, don't take the package)

New `Game.TestInfrastructure/Helpers/FakeTimeProvider.cs`; the byte-identical copies in `SocketCommandProcessorTests` and `DataProviderSynchronizerTests` are gone. Both files already imported `Game.TestInfrastructure.Helpers`, so no call site changed.

I went with option 1 over `Microsoft.Extensions.TimeProvider.Testing`. The issue framed option 2 as "probably right if the timer support is ever wanted" — it isn't, at either call site. Both use exactly `GetUtcNow()` + `Advance(TimeSpan)` to cross a deadline deterministically, `Directory.Packages.props` has no other candidate consumer, and the class is four lines. Taking a package reference on two test projects to delete four lines is the wrong trade today; the reasoning is recorded in the class's `<remarks>` so switching later is a one-file change if timers ever matter.

## `FakeUserLogins` — merged

New `Game.TestInfrastructure/Helpers/FakeUserLogins.cs` records everything and is read selectively: `RecordConnectionCallCount` / `SaveDeviceInfoCallCount`, `LastRecordToken` / `LastSaveToken`, `ThrowOnRecordConnection`, `OnRecordConnection`.

I read the carve-out as *not* applying. It covers a double that **scripts a sequence** specific to one test class (the `FakePlayerRepository` null-then-recovers example); neither copy scripts anything — one counts calls, the other captures tokens. That's exactly the case the same paragraph already rules on: *"prefer one double that records everything and is read selectively over a no-op/recording pair."* `IUserLogins` has two eight-parameter members, so this is the shape that costs two edits per interface change. Merging is the rule as written, so no doc change is needed — the `docs/backend.md` note is only required if a duplicate stays on purpose, and neither does.

One behaviour note: the middleware copy threw `NotSupportedException` from `SaveDeviceInfo`. The merged double records instead. No test asserted that throw, and a canonical double that records is more useful than one that traps.

## Two extras found while scanning (minor, folded in)

Scanning for other cross-project duplicates turned up that **`ThrowingPubSubService` already has a canonical helper** — and two `Game.Application.Tests` classes were shadowing it:

- `PlayerUpdateBatchTests` nested a **byte-identical** private copy, in a file that already imports the helpers namespace. Deleted; it now binds to the canonical one.
- `DataProviderSynchronizerTests` nested a **same-named but unrelated** double (throws on `Subscribe`, not `PublishBatch`). Renamed to `SubscribeThrowingPubSubService` — same name, different behaviour, shadowing the real helper is a live trap for the next reader.

Both are one-line-scale and directly on this issue's theme, so they're here rather than in a follow-up.

## Follow-up filed

`RecordingCache` (`IReloadableReferenceCache`) is duplicated across `Game.Api.Tests` and `Game.Application.Tests` with genuinely different recording surfaces — the same judgement call as `FakeUserLogins`, but a different interface and outside this issue's stated scope. Filed separately rather than folded in.

## Test plan

- [x] `dotnet build` — clean, 0 warnings / 0 errors
- [x] `dotnet test --no-build --no-restore` — **3,656 passed, 0 failed** (Core 1501, Application 1188, Api 893, Infrastructure 74)
- [x] No production code touched; no new tests — the existing suites are the regression guard, as the issue specifies
- [x] No frontend changes, so `npm run lint` / `test:unit:ci` not applicable

---
_Generated by [Claude Code](https://claude.ai/code/session_01MdR1bBZ685GrycqmGS3rSM)_

### DIFF
--- a/Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
+++ b/Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
@@ -214,47 +214,5 @@ namespace Game.Api.Tests.Unit
             return context;
         }
 
-        // Records how many times RecordConnection was actually invoked, so tests can tell whether the
-        // dedupe memo skipped the DB round-trip. ThrowOnRecordConnection lets a test pin the failure path.
-        // OnRecordConnection lets a test observe when the call happened relative to other events.
-        private sealed class FakeUserLogins : IUserLogins
-        {
-            public int RecordConnectionCallCount { get; private set; }
-            public bool ThrowOnRecordConnection { get; set; }
-            public Action? OnRecordConnection { get; set; }
-
-            public Task RecordConnection(
-                int userId,
-                string ipAddress,
-                string deviceFingerprintHash,
-                string userAgent,
-                string? secChUa,
-                string? secChUaMobile,
-                string? secChUaPlatform,
-                CancellationToken cancellationToken = default)
-            {
-                RecordConnectionCallCount++;
-                OnRecordConnection?.Invoke();
-                if (ThrowOnRecordConnection)
-                {
-                    throw new InvalidOperationException("simulated tracking failure");
-                }
-
-                return Task.CompletedTask;
-            }
-
-            public Task SaveDeviceInfo(
-                int userId,
-                string deviceFingerprintHash,
-                string? secChUa,
-                string? secChUaMobile,
-                string? secChUaPlatform,
-                double? deviceMemory,
-                int? hardwareConcurrency,
-                CancellationToken cancellationToken = default)
-            {
-                throw new NotSupportedException();
-            }
-        }
     }
 }
--- a/Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
+++ b/Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
@@ -213,6 +213,5 @@ namespace Game.Api.Tests.Unit
 
             return context;
         }
-
     }
 }
--- a/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
+++ b/Game.Api.Tests/Unit/SocketCommandProcessorTests.cs
@@ -759,16 +759,6 @@ namespace Game.Api.Tests.Unit
             public void HashSetIfExistsAndForget(string key, IReadOnlyDictionary<string, string> fields, TimeSpan expiry) => throw new NotSupportedException();
         }
 
-        /// <summary>A controllable <see cref="TimeProvider"/> so the switch-credit wait deadline can be
-        /// crossed deterministically instead of relying on wall-clock delays.</summary>
-        private sealed class FakeTimeProvider(DateTimeOffset start) : TimeProvider
-        {
-            private DateTimeOffset _now = start;
-
-            public override DateTimeOffset GetUtcNow() => _now;
-            public void Advance(TimeSpan delta) => _now += delta;
-        }
-
         private sealed class NoOpHostLifetime : IHostApplicationLifetime
         {
             public CancellationToken ApplicationStarted => CancellationToken.None;
--- a/Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs
+++ b/Game.Application.Tests/DataAccess/DataProviderSynchronizerTests.cs
@@ -1149,7 +1149,7 @@ namespace Game.Application.Tests.DataAccess
         {
             using var scope = CreateScope();
             var logger = new CapturingLogger<DataProviderSynchronizer>();
-            var throwingPubSub = new ThrowingPubSubService();
+            var throwingPubSub = new SubscribeThrowingPubSubService();
             var synchronizer = new DataProviderSynchronizer(scope.ServiceProvider, throwingPubSub, logger, TestRetryPolicy);
 
             await Assert.ThrowsAsync<InvalidOperationException>(() => synchronizer.StartAsync(CancellationToken.None));
@@ -2490,7 +2490,7 @@ namespace Game.Application.Tests.DataAccess
         /// A minimal <see cref="IPubSubService"/> stub whose <c>Subscribe</c> overloads always throw, used to verify
         /// that <see cref="DataProviderSynchronizer.StartAsync"/> propagates a subscribe failure rather than swallowing it.
         /// </summary>
-        private sealed class ThrowingPubSubService : NotSupportedPubSubService
+        private sealed class SubscribeThrowingPubSubService : NotSupportedPubSubService
         {
             public override Task Subscribe(string channel, Action<(string message, string channel)> action, string? id = null) => throw new InvalidOperationException("Simulated subscribe failure.");
             public override Task Subscribe(string channel, string queueName, Func<(IPubSubQueue queue, string channel), Task> action, string id) => throw new InvalidOperationException("Simulated subscribe failure.");
@@ -2741,16 +2741,6 @@ namespace Game.Application.Tests.DataAccess
             public Task AddToQueueAsync(string value, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task AddRangeToQueueAsync(IEnumerable<string> values, CancellationToken cancellationToken = default) => throw new NotSupportedException();
             public Task<long> RemoveRangeAsync(IReadOnlyList<string> values, CancellationToken cancellationToken = default) => throw new NotSupportedException();
-        }
-
-        /// <summary>A controllable <see cref="TimeProvider"/> so the opportunistic-reclaim grace period can be
-        /// crossed deterministically instead of relying on wall-clock delays.</summary>
-        private sealed class FakeTimeProvider(DateTimeOffset start) : TimeProvider
-        {
-            private DateTimeOffset _now = start;
-
-            public override DateTimeOffset GetUtcNow() => _now;
-            public void Advance(TimeSpan delta) => _now += delta;
         }
 
         private sealed class CapturingLogger<T> : ILogger<T>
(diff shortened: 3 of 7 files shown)


Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
```diff
@@ -214,47 +214,5 @@ private static DefaultHttpContext CreateContext(string ip, string? fingerprint,
             return context;
         }
```
ginderjeremiah: Nit: removing the nested class left an orphan blank line between the last member and the class-closing brace. Same leftover in two other files:

- `Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs:216`
- `Game.Application.Tests/DataAccess/PlayerUpdateBatchTests.cs:230`
- `Game.Application.Tests/Services/LoginTrackingServiceTests.cs:37`

`SocketCommandProcessorTests` and `DataProviderSynchronizerTests` came out clean because the deleted class wasn't the last member there. Purely cosmetic, but it's exactly the residue `CLAUDE.md` asks not to leave behind after a removal — worth a three-line cleanup while the branch is open.

---

ginderjeremiah (commented):
## Review: consolidating the `FakeTimeProvider` / `FakeUserLogins` duplicates into `Game.TestInfrastructure` (#2488)

**Verdict: Approve** — the change does what #2488 asked, the two judgement calls are decided the way `docs/backend.md` actually reads, and I could not find a behaviour change hiding in it. One whitespace nit, nothing blocking.

(Submitted as a `COMMENT` because GitHub won't take a gating verdict from the PR author — read it as an approve.)

### Verification I did rather than taking the description's word for it

- **The `FakeUserLogins` merge is the rule, not an exception to it.** `docs/backend.md:23` carves out a local double only when it **scripts** a sequence. Neither copy scripted anything — one counts calls, one captures tokens — and the same paragraph explicitly prefers "one double that records everything and is read selectively over a no-op/recording pair." Merging is correct and no doc edit is owed, since the doc note is only required when a duplicate stays on purpose.
- **Dropping `SaveDeviceInfo`'s `NotSupportedException` costs no coverage.** I checked whether the throw was acting as an unwritten negative assertion — it wasn't: `LoginTrackingMiddleware` only ever reaches `RecordConnection` (`Game.Api/Middleware/LoginTrackingMiddleware.cs:88`), and there is no `SaveDeviceInfo` reference anywhere in `LoginTrackingMiddlewareTests`. The member was unreachable from that suite, so recording instead of throwing changes nothing observable.
- **The deleted `ThrowingPubSubService` really is byte-identical** to `Game.TestInfrastructure/Helpers/ThrowingPubSubService.cs:9`, so `PlayerUpdateBatchTests` binds to a double with the same message and same behaviour. Confirmed rather than assumed, since a silent divergence there would have quietly changed what that test exercises.
- **Both call-site files already import `Game.TestInfrastructure.Helpers`** and both test projects already reference the project, so nothing had to be wired up.
- **`SaveDeviceInfoCallCount` is currently unread.** Deliberately not flagging it — an unread recorder is precisely the shape `docs/backend.md:23` sanctions.

### Nits

- Orphan blank line left before the class-closing brace in three files after the nested class was removed (inline comment on `LoginTrackingMiddlewareTests.cs:216` lists all three).

### Test coverage

Appropriate as-is. This is a test-only refactor with no production code touched, and #2488 explicitly specifies no new tests — the existing suites are the regression guard. The moved doubles keep their observable behaviour at every call site (7 `FakeTimeProvider` uses, 7 `FakeUserLogins` uses, all verified against the merged surface), so a green run is a meaningful signal here rather than a formality.

### Follow-ups

- #2513 (`RecordingCache`) is correctly scoped out and genuinely well-framed — it independently spots the thing I'd have raised, that `FailFirstAttempts` is arguably scripting and so sits closer to the carve-out than `FakeUserLogins` did. Worth carrying that nuance into the fix so it doesn't get resolved as a reflex merge.
- One duplicate cluster the scan didn't mention: **`CapturingLogger` exists three times** — `Game.TestInfrastructure/Helpers/CapturingLogger.cs` (provider-based), a nested `CapturingLogger<T>` in `DataProviderSynchronizerTests.cs:2746`, and `Game.Infrastructure.Tests/CapturingLogger.cs`. They implement three *different* interfaces (`ILoggerProvider` / `ILogger<T>` / `ILogger`), so the one-canonical-double-per-interface rule doesn't cleanly bite and this is not a blocker — but it's the same concept built three times in three projects, which is the spirit of the rule if not its letter. Suggest adding it to #2513's "Note on the neighbours" section so the next scan doesn't rediscover it cold.

### What's good

- The `ThrowingPubSubService` extras are the most valuable part of this PR and weren't asked for. A private class **shadowing a canonical helper of the same name with different behaviour** (`DataProviderSynchronizerTests`) is a live trap — a reader deletes the "redundant" nested copy, the test silently rebinds to a double that throws somewhere else, and the failure looks unrelated. Renaming to `SubscribeThrowingPubSubService` kills it, and folding both into this PR rather than a follow-up was the right call at that size.
- Choosing option 1 over `Microsoft.Extensions.TimeProvider.Testing` is well-argued and, more usefully, the reasoning is recorded in the class's `<remarks>` where the next person to reach for that package will actually hit it — not buried in a PR description.
- The description states its judgement calls with the rule text they turn on, which is what made this reviewable quickly.

### Not merging

Per the review workflow, a nit is outstanding and CI is still mid-run (`test-backend`, `verify-backend`, and the three `test-e2e` shards were in progress at review time). Fix the three blank lines and let the run finish, and this is good to go.

---

Game.Api.Tests/Unit/LoginTrackingMiddlewareTests.cs
ginderjeremiah: Fixed in 6e90058 — all three blank lines dropped. Confirmed the diff is whitespace-only (`git diff -w` is empty) and the build stays clean at 0 warnings.

Your read on the other two files matches what I see: `SocketCommandProcessorTests` and `DataProviderSynchronizerTests` had a following member after the deleted class, so no residue there.

On the `CapturingLogger` triplication — checked it, and it's real: `Game.TestInfrastructure/Helpers/CapturingLogger.cs` (public `CapturingLoggerProvider` plus an internal per-category `CapturingLogger : ILogger`), the nested `CapturingLogger<T> : ILogger<T>` at `DataProviderSynchronizerTests.cs:2746`, and `Game.Infrastructure.Tests/CapturingLogger.cs` (`internal sealed : ILogger`). Agreed it's spirit-not-letter, so I've added it to #2513's "Note on the neighbours" rather than opening a third issue, and flagged it as lower priority than the `RecordingCache` call itself. I also carried your point about `FailFirstAttempts` into the body so the next person resolves that on the merits instead of reflex-merging.

---

PR #2514, opened 2026-07-29
Consolidate the three engine `window` stubs onto a shared test helper (#2489)

### LINKED ISSUE #2489: Consolidate the three engine `window` stubs onto a shared test helper
Raised in the review of #2483, which added the third copy.

## Problem

Three engine suites each hand-roll a `vi.stubGlobal('window', …)` block, and they disagree in ways nothing forces them to:

- `UI/src/tests/lib/engine/logical-engine.test.ts:8` — `setInterval`/`clearInterval` delegating to `globalThis`, plus a bare `requestAnimationFrame: vi.fn()`.
- `UI/src/tests/lib/engine/render-engine.test.ts:14` — `requestAnimationFrame` recording callbacks with a monotonic handle, `cancelAnimationFrame` recording cancelled handles.
- `UI/src/tests/lib/engine/engine-statify.svelte.test.ts:10` (new in #2483) — effectively the union of the other two.

The seam itself is legitimate: `createTickSource` falls back to `window.setInterval` when `Worker` is absent, and `RenderEngine` drives itself off `window.requestAnimationFrame`, so any suite exercising either loop has to stand them up. What's duplicated is the standing-up, not the intent.

The concrete cost showed up immediately: the new copy originally returned `rafCallbacks.length` as the rAF handle, which repeats a value once a test `shift()`s a callback off the front — inert there (it asserts nothing about cancellation) but wrong, and a mistake the sibling file had already avoided. #2483 fixed that copy; a shared helper is what stops the next one.

## Suggested direction

A `stubEngineWindow()` helper under `UI/src/tests/fixtures/` returning the recording handles the callers need (something like `{ rafCallbacks, cancelledHandles, reset() }`), with the three suites calling it instead. Same spirit as the backend test-double consolidation in #2470 / #2487, and the frontend fixture consolidations in #2481 / #2482.

Worth deciding as part of it whether the helper covers only the four members the engines actually touch or a fuller `window` shape — the current stubs replace `window` wholesale, so a suite that later needs another member has to edit its own copy either way.

## Scope

Small — test-only, three files plus the new helper. No production code, no behaviour change; the existing engine assertions are the regression guard.

ginderjeremiah:
Closes #2489.

## Summary

`logical-engine.test.ts`, `render-engine.test.ts` and `engine-statify.svelte.test.ts` each hand-rolled a `vi.stubGlobal('window', …)` block. All three now call a single `stubEngineWindow()` helper that returns the recording handles the suites read (`rafCallbacks`, `cancelledHandles`, `reset()`).

Net −25 lines across the three suites. No production code, no behaviour change — the existing engine assertions are the regression guard.

## How it addresses the issue

The issue's concrete cost was that the three copies disagreed on things nothing forced them to, and the newest copy had shipped a handle scheme (`rafCallbacks.length`) that repeats a value once a test `shift()`s a callback off the front — inert in that suite, wrong anyway, and a mistake the sibling file had already avoided. There is now one implementation to get right, and the monotonic-counter reasoning lives next to the counter instead of in a comment warning the next copyist.

The two calls the issue left open, and how I landed them:

**Which members the helper covers — the four the engines actually touch, not a fuller `window`.** Those four *are* the seam: `createTickSource` falls back to `window.setInterval`/`clearInterval` when `Worker` is absent (jsdom), and `RenderEngine` drives its loop off `window.requestAnimationFrame`/`cancelAnimationFrame`. Nothing else in `src/lib/engine/` reads `window` at all. A fuller shape would invite a suite to lean on a member that isn't a seam, and since the stub replaces `window` wholesale either way, widening it is now one edit here rather than a fourth copy.

**Where it lives — co-located, not in `tests/fixtures/`.** The issue suggested `UI/src/tests/fixtures/`, and I went the other way: `docs/frontend.md` → Testing Guidelines defines that folder as one module per *contract*, each exporting a `make<Contract>(overrides)`. An environment stub is neither a contract nor a builder, and all three consumers sit in `src/tests/lib/engine/`, so it lands beside them — the same shape as the existing `progression-test-utils.ts`. I added one line to `docs/frontend.md` recording that split, since the alternative is every future helper of this kind re-litigating the same question. Happy to move it if you'd rather `fixtures/` be the catch-all for shared test scaffolding.

One note on placement inside each suite: `stubEngineWindow()` is called at module scope, where the raw blocks were. It reads as though it must run before the engine imports, but ESM hoists those imports above it regardless — what actually makes this safe is that the engines read `window` lazily (at start/stop), never at import. The helper's header says so, since the ordering looks load-bearing and isn't.

## Test plan

- [x] `npm run lint` (eslint + `svelte-check`) — clean, 1232 files, 0 errors/warnings
- [x] `npm run test:unit:ci` — **313 files, 4059 tests, all passing**
- [x] The three affected suites specifically — 31 tests passing, including the cancellation assertions (`expect(cancelledHandles).toEqual([3])`) that are the ones sensitive to handle numbering
- [x] Coverage gate unaffected — `src/tests/**` is an explicit exclusion in `coverage.config.ts`, and the new file is not matched by vitest's `include` (`src/**/*.{test,spec}.{js,ts}`), so it is not collected as a suite

No backend changes, so `dotnet build` / `dotnet test` were not run.

---
_Generated by [Claude Code](https://claude.ai/code/session_013ViG2eB9PEtePUJWiSAf8B)_


UI/src/tests/lib/engine/engine-window-stub.ts
```diff
@@ -0,0 +1,60 @@
...
+	});
+
+	return {
+		rafCallbacks,
+		cancelledHandles,
+		reset: () => {
+			rafCallbacks.length = 0;
+			cancelledHandles.length = 0;
+			rafHandleCounter = 0;
+		}
```
ginderjeremiah, line 55: Nit: `reset()` clears the two recording arrays and the counter, but not the `vi.fn()` call history on `requestAnimationFrame` / `cancelAnimationFrame`. Nothing today reads those (all three suites assert on `rafCallbacks` / `cancelledHandles`), which leaves the helper in the one state that's internally inconsistent: the mocks record, but the reset doesn't cover them. The first suite to reach for `expect(window.requestAnimationFrame).toHaveBeenCalledTimes(1)` gets counts accumulated across every prior test in the file — order-dependent and exactly the copy-drift class of bug this helper exists to end.

Two ways to close it, either is fine:

- keep the mocks and hold references so `reset()` can `mockClear()` them, or
- drop `vi.fn` entirely — the arrays already are the recording, so plain arrow functions carry the same information with nothing left to reset.

The second is probably the smaller helper.

---

UI/src/tests/lib/engine/engine-window-stub.ts
```diff
@@ -0,0 +1,60 @@
+import { vi } from 'vitest';
+
+/* Shared `window` stub for the engine suites (#2489), replacing three hand-rolled `vi.stubGlobal`
+   blocks that disagreed on which members they defined and on how rAF handles were numbered.
+
+   Co-located with its consumers rather than placed in `tests/fixtures/`, which `docs/frontend.md` →
+   Testing Guidelines reserves for `make<Contract>(overrides)` builders of a typed contract. This is an
+   environment seam, not a contract, and every caller lives in this folder — same shape as
+   `progression-test-utils.ts`.
```
ginderjeremiah: Nit: this paragraph now says the same thing as the sentence added to `docs/frontend.md` → Testing Guidelines, and the doc is the place a future author will actually look before deciding where to put the next shared helper. Two copies of a convention drift; I'd cut this one and keep the doc's. The third paragraph (which four members and why) earns its space — that's the local decision — and CLAUDE.md's 1–2 line guidance is worth weighing against a 13-line header on a 60-line helper.

---

ginderjeremiah (commented):
## Review: consolidate the three engine `window` stubs onto `stubEngineWindow()` (#2489)

**Verdict:** Approve (posted as a comment — GitHub won't take a gating verdict from the PR author) — the consolidation is correct and the two open calls from the issue were both landed the right way; two nits, no blockers.

Verified the change rather than taking the description on faith:

- The "four members" claim holds. `grep` for `window.` across `src/lib/engine/` returns exactly `setInterval`/`clearInterval` (`tick-source.ts:39,47,53,55`) and `requestAnimationFrame`/`cancelAnimationFrame` (`render-engine.ts:50,60`). Nothing else in the engine graph touches `window`, so the helper's surface is the seam, not a guess at one.
- The lazy-read argument is right, and it matters. `stubEngineWindow()` sits below `vi.mock`/`vi.hoisted` and above the engine imports, which ESM hoists past it — the stub is safe only because both engines read `window` inside `start()`/`stop()`. Saying so in the JSDoc is the right call; the ordering does look load-bearing.
- No stub was left behind: `vi.stubGlobal('window', …)` now appears once under `src/tests/`, in the helper. The other hit (`lib/api/api-request.test.ts:49`) stubs `location`, a different seam entirely, and correctly stays out of scope.
- Behaviour per suite is unchanged. `logical-engine.test.ts` gains a recording rAF and a `cancelAnimationFrame` it never had — `LogicalEngine` calls neither, so the old bare `requestAnimationFrame: vi.fn()` was vestigial and this is a no-op. The two rAF suites keep the monotonic counter they depended on. No `unstubGlobals` / `restoreMocks` in `vite.config.ts`, and Vitest 4's `restoreAllMocks` only restores spies, so the module-scope stub survives each suite's `afterEach` exactly as before.

### Nits

- `engine-window-stub.ts:54` — `reset()` doesn't clear the `vi.fn()` call history, so the mocks record something no reset covers. Latent, but it's the one inconsistency left in a helper whose purpose is to be the copy that's right. Either clear them or drop `vi.fn` (details inline).
- `engine-window-stub.ts:6` — the co-location paragraph now duplicates the `docs/frontend.md` sentence this PR adds; I'd keep the doc's and cut the header's.

Both are in this PR's own new file, so they belong here rather than as follow-ups — which is why I'm not merging this yet despite the approve.

### Test coverage

Test-only, no production change, so the bar is that the existing assertions still pin what the helper centralizes — and they do. `render-engine.test.ts:64` and `:72` (`expect(cancelledHandles).toEqual([1])` / `([3])`) are precisely the assertions that fail if handle numbering regresses to `rafCallbacks.length`, which is the bug #2489 was raised over. The helper itself needs no suite of its own; it has three. Coverage gate is unaffected — `src/tests/**` is excluded in `coverage.config.ts`, and the file doesn't match vitest's `include`, so it isn't collected as a suite.

### What's good

Going against the issue's suggested `tests/fixtures/` location was the better call, and the reasoning is the project's own: `fixtures/` is defined as one module per contract exporting `make<Contract>(overrides)`, and an environment stub is neither. Co-locating matches the existing `battle-sim-test-utils.ts` / `progression-test-utils.ts` / `screen-mount-tracker.ts` precedent, and adding the one-line rule to `docs/frontend.md` means the next helper of this kind doesn't re-argue it. Also worth noting the header comment moves the monotonic-counter reasoning next to the counter instead of leaving it as a warning to the next copyist — that's the actual fix for the class of bug, beyond the −25 lines.

---

UI/src/tests/lib/engine/engine-window-stub.ts
ginderjeremiah: Took the second option in `dbf28ad` — plain arrow functions, no `vi.fn`. Confirmed nothing reads the call history first: the only `requestAnimationFrame` mention left across the three suites is a test *title* (`render-engine.test.ts:40`), and every assertion goes through `rafCallbacks`/`cancelledHandles`. So the mocks were carrying a second, unreset record of information the arrays already hold — dropping them removes the inconsistency rather than papering over it with a `mockClear()`.

Noted the reason in the JSDoc so the next person doesn't reintroduce `vi.fn` for symmetry.

---

UI/src/tests/lib/engine/engine-window-stub.ts
ginderjeremiah: Agreed — cut in `dbf28ad`. The header is down to the two paragraphs that are local decisions (what this replaces, and which four members plus why), and `docs/frontend.md` is the single statement of the co-location rule.

---

